PR #119, opened 2026-03-20
fix(cli): route --password flag through CLI password cache

neavra:
## Problem

The `--password` flag on `vultisig agent --password <pw>` doesn't prevent interactive password prompts during transaction signing. Users get prompted even though they provided the password upfront. This also breaks `--via-agent` pipe mode which can't handle interactive prompts.

## Root Cause

Three independent password stores that didn't talk to each other:

1. **`executor.password`** — set by `--password` flag, used for manual `vault.unlock()` calls
2. **CLI password manager cache** — `Map` in `password-manager.ts`, used by the SDK's `onPasswordRequired` callback
3. **SDK vault passwordCache** — internal to the SDK, populated by `unlock()` or callback return

The `--password` flag only populated #1. When `vault.sign()` called `resolvePassword()` → `onPasswordRequired` → `getPassword()`, it checked #2 (empty) and fell through to an interactive prompt.

## Fix

- Cache `--password` value in the CLI password manager (#2) via `cachePassword()`
- Remove `executor.password` field and `setPassword()` method
- Remove three manual `vault.unlock()` blocks in `signSdkTx`, `signServerTx`, `signTypedData`

The SDK's callback chain (`resolvePassword` → `onPasswordRequired` → `getPassword`) now finds the password in the CLI cache automatically. One source of truth instead of three.

## Testing

- Auth signing with `--password`: no prompt ✓
- ETH send via agent with `--password`: no prompt, tx broadcast ✓
- `--via-agent` pipe mode: no prompt, tx broadcast ✓

## Changes

- `clients/cli/src/agent/session.ts` — replace `executor.setPassword()` with `cachePassword()` (4 call sites)
- `clients/cli/src/agent/executor.ts` — remove `password` field, `setPassword()`, three `vault.unlock()` blocks

## Summary by CodeRabbit

* **Security**
  * Vault credentials are now managed via the CLI password manager instead of temporary in-memory storage during signing.
  * Signing flows no longer perform a pre-sign vault unlock; passwords are cached for the vault key (and optional vault name) when provided or requested.

PR #315, opened 2026-04-23
feat(cli): migrate action dispatch to client-side tool calls

neavra:
## Summary

Brings the CLI up to the post-agent-backend-PR vultisig/agent-backend#119 protocol by routing `tool-input-available` events flagged with `clientExecuted: true` through the existing `executor.executeAction` machinery, and migrating the return channel from the legacy `action_result` field to `context.recent_actions`.

## The gap this closes

After agent-backend vultisig/agent-backend#119 (2026-04-16), `add_coin`, `sign_typed_data`, `create_vault`, `remove_coin`, `address_book_add`, and 6 other client-side tools moved from the `data-actions` SSE channel (now dark) to `tool-input-available` events. Mobile migrated; the CLI didn't. Result: every client-side tool other than `sign_tx` has been silently no-oping through `vsig agent ask` for 10 days.

Example before this PR (verified this session via sub-agent):
```
$ vsig agent ask "Add USDC on Base to my vault"
→ Backend emits tool-input-available for add_coin
→ CLI logs `[tool] add_coin ...` in verbose and does nothing
→ LLM says "✅ USDC added to Base!" (hallucinated from server's prep)
→ Vault state: unchanged
```

After this PR, same command → vault actually mutates.

## Mechanism

1. **SSE parser** (`client.ts`): extract `clientExecuted` from `tool-input-available` frames. When true, fire a new `onClientSideToolCall(toolCallId, toolName, input)` callback. Backward compatible — absence of flag falls through to the existing `onToolProgress` (display-only) path, so this works cleanly against older backends without vultisig/agent-backend#160.

2. **Dispatch registry** (`session.ts`): maps tool name → Action type. Synthesises an Action from the tool call and runs it through the pre-existing `executeAction` switch. Unknown tool names log `[cli] unimplemented client-side tool: X` and produce a failure RecentAction — visible, never silent.

3. **Return channel migration**: tool results accumulate into a `pendingToolResults` queue on the session. On the next HTTP request, the queue flushes into `context.recent_actions`. Correct semantics for multi-tool-call LLM turns (one `recent_actions` array per turn, not N sequential HTTP calls).

4. **Legacy removal**: top-level `action_result` field removed from `SendMessageRequest`; `ActionResultPayload` type removed.

5. **Recursion cap** (`MAX_MESSAGE_LOOP_DEPTH=16`): defense-in-depth against runaway loops. Backend's own cap is 8; CLI is 2x that as a belt-and-suspenders.

## What doesn't change

- `executor.ts` — all switch cases reused verbatim, no execution logic touched
- `tx_ready` path for EVM transactions — untouched (works via a different SSE channel)
- Password gating + auto-execute — reused via `executor.executeAction`

## Tests (16 total)

**New**: `clientSideDispatch.test.ts` covers:
- `ActionResult` → `RecentAction` conversion shape (exported helper, not mirrored)
- SSE parser routing: flag present → dispatch, absent → display, malformed → strict no-dispatch, wrong event type → no dispatch, non-object input coerced
- Registry drift guard (`create_vault`, `plugin_install`, `create_policy`, `delete_policy`, `sign_tx` all correctly absent; tool-name ↔ action-type 1:1 invariant locked)
- Recursion cap trips at depth 17+; normal 3-iteration flows never hit it

**Pre-existing**: 125 tests pass. 3 pre-existing file failures (`@vultisig/client-shared` build dependency) are unrelated to this PR.

## E2E validation

Local stack (backend from vultisig/agent-backend#160 worktree + this CLI):
```
$ vsig agent ask "Add USDC on Base to my vault" --password ... --verbose
[SSE] clientExecuted=true event for add_coin
[dispatch] action.type=add_coin
[tool] add_coin: ok
[session] flushed 1 recent_actions into request
$ vsig tokens Base --output json
{ "tokens": [{ "symbol": "USDC", "contractAddress": "0x833589...", ... }] }
```

Vault mutated. Pre-PR, the tokens list was empty.

## Depends on

- vultisig/agent-backend#160 (clientExecuted marker on `V1ToolInputAvailable`). Without it, the flag never appears on SSE events and dispatch never fires — CLI degrades to today's display-only behaviour. Safe rollout if vultisig/agent-backend#160 ships first.

## Part of a three-PR sequence

| PR | Repo | Status |
|---|---|---|
| 1 | vultisig/agent-backend#160 | clientExecuted marker (draft) |
| 2 (this) | vultisig-sdk | CLI dispatch migration |
| 3 | agent-backend | Polymarket sign_bet regression fix — rewrites `polymarket_sign_bet` emissions to `sign_typed_data` with dereferenced Redis payload. Depends on this PR for CLI-side flow; mobile gets Polymarket working from vultisig/agent-backend#160 + PR 3 alone. |

## Team review addressed

Three HIGH findings from the team-review on this branch addressed in commit [00b2f947](../commit/00b2f947):
- H3: exported `actionResultToRecentAction` so tests exercise real code, not a local mirror
- H2: removed `create_vault` from dispatch registry (mobile-only — needs VultiServer / multi-device flow; CLI users use `vsig vault`); documented other mobile-only tools similarly; added registry drift-guard tests
- H1: added `MAX_MESSAGE_LOOP_DEPTH=16` cap + tests

## Test plan

- [x] Unit tests green (16/16 in `clientSideDispatch.test.ts`)
- [x] Existing CLI tests green
- [x] E2E: `add_coin` actually mutates vault
- [x] Typecheck clean
- [ ] E2E: multi-tool-call batching
- [ ] E2E: tx_ready regression check
- [ ] E2E: password-gated `sign_typed_data` flow
- [ ] Integration Test (agentic stack) workflow run

## Summary by CodeRabbit

* **New Features**
  * Added optional client-side tool callback for real-time tool-invocation events.
  * Message context now carries a queued recent-actions list for tool results.

* **Improvements**
  * Serialized client-side tool dispatching with ordering, depth limits, and clearer retry/restore behavior.
  * Normalizes tool-input shapes and improves SSE event filtering for client-triggered calls.

* **Tests**
  * Comprehensive tests covering dispatch ordering, result-shaping, recursion limits, and SSE-trigger conditions.


rcoderdev (approved):

clients/cli/src/agent/session.ts
```diff
@@ -266,11 +307,17 @@ export class AgentSession {
             ui.onToolResult(`mcp-${tool}`, tool, true, { label })
           }
         },
+        onClientSideToolCall: (toolCallId, toolName, input) => {
+          pendingDispatches.push(this.dispatchClientSideTool(toolCallId, toolName, input, ui))
```
gomesalexandre: `preferably-blocking:` multiple `clientExecuted` tool events in a single streamed response are fired in parallel via `pendingDispatches.push(dispatchClientSideTool(...))` — `push` doesn't await, so nothing enforces call order.

Two failure modes this produces:

1. **Out-of-order mutations** — a turn that emits `add_chain(Base) → add_coin(USDC on Base)` would race, and if `add_coin` resolves its chain lookup before `add_chain` has landed, the coin add fails with "chain not configured." The backend treats these as ordered causal operations; the CLI needs to respect that ordering.

2. **Password prompt hang** — the pipe-UI's password-resolver (single-slot, see session.ts where `pendingPassword` is latched) supports one in-flight prompt at a time. Two parallel dispatches that both need a password (e.g. two `sign_typed_data` calls on different vaults) will: dispatch A latches resolver → dispatch B tries to latch too and silently drops / overwrites → user enters password once, dispatch A gets it, dispatch B stays pending forever. Silent hang.

Fix: sequentialize. Simplest shape is an `async-mutex`-style queue, but a one-liner works too:

```ts
// sequential: await the previous dispatch before starting the next.
pendingSerialChain = pendingSerialChain.then(() => dispatchClientSideTool(...))
pendingDispatches.push(pendingSerialChain)
```

this preserves the "we still have a handle to each dispatch's completion" contract the rest of the code relies on while eliminating the parallel-exec + password-slot-collision classes.

regression test worth adding: mock two `clientExecuted: true` events in a single SSE stream (`add_chain → add_coin`) + assert they complete in order via a sequence counter side-effect in the test dispatcher.

clients/cli/src/agent/session.ts
```diff
@@ -237,21 +282,17 @@ export class AgentSession {
...
-      const result = actionResults[0]
-      request.action_result = {
-        action: result.action,
-        action_id: result.action_id,
-        success: result.success,
-        data: result.data || {},
-        error: result.error || '',
-        ...(!result.success && result.code ? { code: result.code } : {}),
+    // Flush the post-#119 return channel (replaces the old action_result field).
+    if (this.pendingToolResults.length > 0) {
```
gomesalexandre, line 285: `should-fix:` `pendingToolResults` persists across a failed / cancelled turn, then gets flushed into the NEXT request's `context.recent_actions` as if those results were current.

Concrete sequence:
1. user: "Add USDC on Base"
2. CLI dispatches `add_coin`, pushes result into `pendingToolResults` → `[{tool:"add_coin", ok:true}]`
3. stream errors on next LLM response (rate-limit / network flake / cancel)
4. user (next turn): "What's the weather in Tokyo?"  ← completely unrelated
5. request body: `context.recent_actions: [{tool:"add_coin", ok:true}]`  ← STALE
6. backend's action-result-to-user-message synthesizer sees an orphaned `add_coin` + wires it into the LLM context as if it just happened
7. LLM: "I've added USDC on Base. The weather in Tokyo is..."  ← hallucinates context from the stale result

Mitigations possible at the backend (ignore recent_actions older than X seconds, dedupe by toolCallId against already-consumed set) but the CLI side has the simpler fix:

```ts
try {
  // stream the turn
} catch (err) {
  if (!isAuthRetry(err)) {
    pendingToolResults = []  // drop stale results on non-auth failure
  }
  throw err
}
```

the auth-retry exception preserves the existing "retry after token refresh re-uses pending results" flow that IS correct.

regression test: mock a stream error after a `clientExecuted: true` tool dispatched + asserted result pushed → next request must NOT include that result in `context.recent_actions`.

gomesalexandre (changes requested):
## verdict — request changes, 1 PB + 1 SF

the architecture is right — routing `clientExecuted: true` tool events through the existing `executor.executeAction` machinery + moving the return channel from legacy `action_result` to `context.recent_actions` is the correct post-#119 migration. Two concurrency / state-management issues surfaced on Codex `gpt-5.3-codex` adversarial pass — inline below. One is user-visible enough to block (parallel dispatch + password-slot collision), the other is a real but narrower class.

## what's solid

- **Backward compat via `clientExecuted` flag**: absence of the flag falls through to the existing `onToolProgress` display-only path. CLI works unchanged against pre-#119 backends. Clean migration shape.
- **Shared `executor.executeAction` machinery**: reuse is correct — whatever dispatch semantics mobile has, CLI inherits. No new fork to maintain.
- **Ten-day silent hallucination repro in the PR body** is concrete and damning. "→ LLM says '✅ USDC added to Base!' (hallucinated from server's prep) → Vault state: unchanged" is exactly the "happy confirmation + no actual mutation" class that erodes user trust.
- **`onClientSideToolCall(toolCallId, toolName, input)` callback signature** threads the tool-call-id through, which is load-bearing for the sequence-id / de-dup work agent-backend is also landing.

## preferably-blocking (see inline)

### 1. parallel dispatch — `session.ts:310-312`

`pendingDispatches.push(dispatchClientSideTool(...))` without awaiting. Two client-side tool calls in one streamed turn run concurrently. Failure modes: (a) ordering-sensitive mutations race (`add_chain → add_coin` fails if add_coin resolves chain before add_chain commits); (b) pipe-UI password resolver is single-slot, two parallel dispatches both needing a password produce a silent hang. Inline patch suggests `pendingSerialChain.then()` sequentialization.

## should-fix (see inline)

### 2. stale `recent_actions` across failed turns — `session.ts:286-287`

`pendingToolResults` only clears on successful flush / depth-cap, not on stream cancel / non-auth error. Next unrelated turn's request body carries stale tool results → LLM hallucinates unrelated context. Inline patch: clear on non-auth error.

## sequencing

- companion to agent-backend#160 (`clientExecuted` marker outbound + `recent_actions` inbound routing). that PR also has an open request_changes from my end (PB on `actionResultFromRequest` returning on first entry + SF on empty-turn guard). Worth resolving those together since the two sides share the protocol contract — any tweak on one probably needs a twin on the other.
- companion to vultiagent-app#167 (CRUD action-discriminator tools `vault_coin`/`vault_chain`/`address_book`). Once all three land, a manual E2E across CLI + mobile on the same flow would be worth capturing as receipts.

## functional testing

- existing unit tests cover happy-path dispatch
- adding regression tests for both inline findings before merge:
  1. parallel-dispatch ordering via sequence counter in a mock dispatcher
  2. stale-recent_actions purge on non-auth stream error

reviewed with Claude Opus 4.7 + Codex `gpt-5.3-codex` adversarial pass. happy to re-approve once both inline fixes land.

neavra:
## reply — PB1 + SF addressed in `c663832f`

thanks @gomesalexandre — another sharp pass, both findings spot on. fixes pushed, plus the two adjacent CodeRabbit items folded in. ran manual E2Es with fault injection to verify; logs below.

### PB1 — parallel dispatch race (`session.ts:310-312`)

Serialized via `dispatchChain.then(...)` — each new dispatch chains off the previous, with `.catch(() => {})` on the chain so a failed dispatch doesn't stall subsequent ones. The original promise is still pushed to `pendingDispatches` so failures still surface at the `Promise.all` await.

Manual E2E receipt — "Add Polygon to my vault, then add USDC on Polygon":
```
[SSE] tool-input-available toolName=add_chain
  ⚡ add_chain ...
[dispatch] action.type=add_chain
[SSE] tool-input-available toolName=add_coin    ← second SSE event arrives mid-dispatch
  ⚡ add_coin ...                                 ← but waits for add_chain
+ Chain added: Polygon
  ✓ add_chain → added                            ← add_chain fully completes
[dispatch] action.type=add_coin                   ← only NOW does add_coin start
+ Token added: USDC on Polygon
  ✓ add_coin → added
[session] flushed 2 recent_actions into request    ← arrival order preserved
```

Without the fix `add_coin` would have raced `add_chain`'s vault commit and the result-push order would have been completion-order, not arrival-order. Confirmed in code via 4 unit tests pinning the chain pattern itself (timing, failure resilience, password peak-concurrency = 1, push order).

### SF — stale `pendingToolResults` across failed turns (`session.ts:286-287`)

Cleared on any non-auth error in `sendMessage`'s catch. Combined with a per-call restore inside `processMessageLoop` so the queue isn't permanently lost on a transient failure — caller can resume from the same state, but if they don't (or it's terminal), the next user turn starts clean.

Manual E2E receipt — injected 500 on the recursive flush after a real `add_coin` dispatch:
```
[dispatch] action.type=add_coin
  ✓ add_coin → added
[session] flushed 1 recent_actions into request
[E2E] SF_E2E_FAULT firing — throwing 500
[E2E] stream error — restored queue len=1                     ← processMessageLoop restored
[E2E] sendMessage catch err=500 ... pendingToolResults.length=1
[E2E] SF: cleared queue                                        ← sendMessage SF cleared
```

Then sent an unrelated follow-up "what is my eth balance" — no `flushed N recent_actions` log on that turn → confirmed no stale leak.

### CodeRabbit's adjacent items

Folded in: 401/403 retry moved inside `processMessageLoop` so it replays the same request body (E2E-verified — single dispatch, no LLM-emit-loop), and the `onClientSideToolCall` callback type tightened to `() => void`.

happy to re-run any specific scenario you want before re-review.

gomesalexandre (approved):
## R3 — sign-off at `c663832f`

gm @neavra — verified the R2 tackle pass on my own tree. Both findings from R1 closed; the c663832f follow-on E2E catch (CR2 snapshot was always-empty) is a meaningful improvement on top.

| R1 finding | severity | status | spot-check |
|---|---|---|---|
| **PB1** parallel dispatch race (password-slot collision + ordering-sensitive flows like `add_chain → add_coin`) | preferably-blocking | ✅ fixed (`16c3ad50`) | `clients/cli/src/agent/session.ts:295-319` — `dispatchChain` chains via `.then()`, each `dispatchClientSideTool` waits for the previous one. Inline comment explicitly cites the password-slot hang failure mode. `pendingDispatches.push(dispatch)` keeps the parallel-await semantics for cleanup, but execution is fully serial. |
| **SF** `pendingToolResults` queue lost on 401 retry path | should-fix | ✅ fixed via 2-step (`16c3ad50` initial → `c663832f` rework after E2E exposed the snapshot-at-entry bug) | the retry now happens at the request boundary (`session.ts:349-379`), `flushedThisCall` is snapshotted BEFORE the splice (`:284-291`), and on non-401 errors the spliced batch is restored to `pendingToolResults` (`:379-380`). 401 retries replay the exact same body (same content + same `recent_actions`), avoiding the LLM-loop class of bug where the original message gets re-delivered. |

## R2 self-review the c663832f follow-on caught

`16c3ad50`'s first cut snapshotted `pendingToolResults` at `sendMessage` entry — but the queue is populated by SSE dispatches DURING `sendMessage`, so the snapshot was always empty and the restore was a no-op. The `c663832f` rework moves the snapshot to right before the `splice()` call (when the queue actually has content). Real catch from the E2E pass; fix is correct.

## test confirmation

```
$ cd clients/cli && yarn test
 Test Files  13 passed (13)
      Tests  167 passed (167)
```

Both `clientSideDispatch.test.ts` cases plus the broader CLI suite green at `c663832f`.

## cross-repo coherence with agent-backend #160

Just signed off on agent-backend #160 — the backend half of this protocol primitive. The CLI dispatch table is a strict subset of `clientSideToolNames` per my #160 review (5 tools missing including `polymarket_sign_bet`). The gap is closed by agent-backend #163 (`fix/polymarket-sign-bet-rewrite`) which rewrites `polymarket_sign_bet` to `sign_typed_data` BEFORE emit, so the CLI sees a tool it knows how to dispatch + the polymarket markers ride through `__pm_auto_submit` in `Action.data`.

Worth a one-liner comment in `CLIENT_SIDE_TOOL_DISPATCH` cross-referencing #163 so the next contributor doesn't try to add `polymarket_sign_bet` directly (which would short-circuit the rewrite path and break the auto-submit markers). Non-blocking — the `unimplemented in CLI: <tool>` failure path is already visible-not-silent.

## CodeRabbit nitpick on depth cap

The 2026-04-27T06:45 CodeRabbit nitpick about `depth > MAX_MESSAGE_LOOP_DEPTH` (17 frames vs intended 16) is a cosmetic style nit — both are well above backend's 8 cap and either off-by-one is fine for defense-in-depth. Either fix or wave through; I'm meh.

## mergefix

clean — `git merge origin/main --no-edit` is `Already up to date` from PR HEAD. `c663832f` already includes the `a0d09280` merge from main.

## verdict

**APPROVE** at `c663832f`.

Sequential dispatch closes the password-slot hang, the request-boundary 401 retry closes the LLM-loop class, the snapshot/restore queue lifecycle is clean. Cross-repo with #160 is consistent. Ship it.

🤖 reviewed via [Claude Code](https://claude.com/claude-code) — R2 spot-check on diff + own-tree CLI test suite confirmation + cross-repo with agent-backend #160.

PR #317, opened 2026-04-23
fix(qbtc): match chain claim tag 'ecdsa-hash160:'

### LINKED ISSUE #316: qbtc claim: update claim message tag to "ecdsa-hash160:" (PR #148)
## Summary

[btcq-org/qbtc#148](https://github.com/btcq-org/qbtc/pull/148) reshaped the QBTC claim ZK circuit — RIPEMD-160 moved out of the circuit and the domain-separation tag on the message the BTC key signs changed. The SDK's `computeClaimMessageHash` still uses the old tag, so every BTC signature we produce today is over the wrong hash and the proof service rejects the proof with `constraint #49904 not satisfied`.

## Old vs new tag

- Old (current SDK): ``encoder.encode(`${circuit}:`)`` → `"ecdsa:"` (6 bytes)
- New (chain): `"ecdsa-hash160:"` (14 bytes)

Chain source: [`x/qbtc/zk/message.go:17`](https://github.com/btcq-org/qbtc/blob/main/x/qbtc/zk/message.go#L17) (`ClaimTagECDSAHash160`).

## Scope of fix

`packages/core/chain/chains/cosmos/qbtc/claim/computeClaimHashes.ts`:

- Replace the `${circuit}:` prefix with `"ecdsa-hash160:"` for ECDSA circuits (P2PKH / P2WPKH / P2SH-P2WPKH / P2WSH).
- Drop the `"schnorr:"` speculative prefix. Per the chain comment, Schnorr is a future circuit with a not-yet-defined tag — throw "Schnorr / Taproot claim is not yet supported" from `computeClaimMessageHash` and `computeAllClaimHashes` when `circuit === 'schnorr'` so callers fail loudly instead of producing an unsigned-by-the-chain message.
- Update `computeClaimHashes.test.ts` accordingly.

## Why now

Blocks the QBTC claim flow end-to-end in the extension ([vultisig/vultisig-windows#3747](https://github.com/vultisig/vultisig-windows/pull/3747)). Every claim currently fails at proof generation because of this.

Ehsan-saradar:
## Summary

Closes #316. Switches the QBTC claim message prefix to `ecdsa-hash160:` to match [btcq-org/qbtc#148](https://github.com/btcq-org/qbtc/pull/148) which reshaped the claim ZK circuit (removed RIPEMD-160 from the circuit, renamed the domain-separation tag).

Until this, every claim failed proof generation with `constraint #49904 not satisfied` — the BTC key was signing over the old prefix but the prover's circuit was verifying against the new one.

## What changed

`packages/core/chain/chains/cosmos/qbtc/claim/computeClaimHashes.ts`

- ECDSA circuits now use `ecdsa-hash160:` (matches `ClaimTagECDSAHash160` in `x/qbtc/zk/message.go`).
- Schnorr circuits: `computeClaimMessageHash` throws. The chain comment explicitly marks Schnorr as a future circuit whose tag is not yet defined — guessing one would just silently produce another invalid signature. Better to fail loudly until the chain defines it.
- Tests updated to cover both paths.

## Test plan

- [x] `yarn test:unit packages/core/chain/chains/cosmos/qbtc/claim/computeClaimHashes.test.ts` — 1122 / 1122 passing
- [ ] E2E claim from the extension ([vultisig/vultisig-windows#3747](https://github.com/vultisig/vultisig-windows/pull/3747)) once the QBTC testnet's new circuit + VK are deployed

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected claim message hashing format to use the proper domain-separation prefix.
  * Implemented stricter validation for address hash requirements.
  * Schnorr/Taproot circuits now return appropriate error messages.

* **Tests**
  * Updated claim hashing tests to reflect corrected on-chain behavior expectations.


rcoderdev (approved):

gomesalexandre (approved):
## verdict — approve

clean 1-file fix with test updates + explicit fail-loud on the unreachable Schnorr branch. Codex `gpt-5.3-codex` adversarial pass turned up nothing.

## what's solid

- **Domain-separation tag switch** to `ecdsa-hash160:` correctly lines up with [btcq-org/qbtc#148](https://github.com/btcq-org/qbtc/pull/148)'s `ClaimTagECDSAHash160` in `x/qbtc/zk/message.go`. The old prefix left every claim failing at `constraint #49904 not satisfied` — exact class of "signing one thing, verifying against another" that this PR closes.
- **Schnorr throw is the right call**. Guessing a tag for a circuit that isn't defined yet would just silently produce an invalid signature on the new circuit the same way the old ECDSA tag does today. Better to fail loudly at the SDK boundary than let callers burn proof-gen cycles on something that was never going to verify.
- **Tests updated for both paths** — ECDSA happy-path expects the new tag; Schnorr path expects the throw. If the next QBTC circuit adds a `ClaimTagSchnorr` upstream, flipping this PR's throw to the new tag needs a 2-line test change, which is the right cost shape.

## sequencing

- **E2E claim gate** in the PR body is unchecked pending QBTC testnet circuit + VK deployment + extension PR (vultisig-windows#3747). Correct framing — no way to validate end-to-end until the matching circuit ships.
- No runtime regression for non-QBTC callers (this file is scoped to the QBTC claim path).

🤖 reviewed with Codex `gpt-5.3-codex` adversarial pass, clean.

PR #342, opened 2026-04-27
fix(security/blockaid): pair swap diffs across all asset diffs in EVM simulations

### LINKED ISSUE #3799: Blockaid simulation parser drops swap pair when a router/permit leg sits between user-side diffs
## Summary

`parseBlockaidEvmSimulation` in `core-chain/security/blockaid/tx/simulation/api/core.ts` destructures only the first two entries of `assetDiffs` positionally. For `permitAndCall` (and other router-mediated flows) Blockaid often returns **three** diffs, with the user's `in` side in `diffs[2]` — the parser then bails to `null` and the simulation hero shows only a single asset (or nothing), even though the user is effectively swapping `A → B`.

## Where

`core-chain` package, `security/blockaid/tx/simulation/api/core.ts` → `parseBlockaidEvmSimulation`.

Relevant snippet (compiled):

```js
const [potentialOutAsset, potentialInAsset] = assetDiffs; // only first two
const { inAsset, inValue } = potentialInAsset.in.length > 0
  ? { inAsset: potentialInAsset.asset, inValue: potentialInAsset.in }
  : { inAsset: potentialOutAsset.asset, inValue: potentialOutAsset.in };
// ...
if (outValue.length === 0 || inValue.length === 0) return null;
```

## Repro shape

A typical `permitAndCall` scan returns:

```
diffs[0]: out=[tokenA], in=[]     // user sends
diffs[1]: out=[],       in=[]     // intermediate permit/allowance/router leg
diffs[2]: out=[],       in=[tokenB] // user receives
```

With the current parser:

- `potentialOutAsset = diffs[0]`, `potentialInAsset = diffs[1]`
- `diffs[1].in` is empty → falls back to `diffs[0].in` (also empty)
- `inValue.length === 0` → returns `null` → no swap hero

## Expected

The parser should find the user-side `out` and `in` across **all** diffs, ignoring intermediate legs that have neither. Both tokens should render in the hero for `permitAndCall` flows.

## Reference — iOS behaviour

iOS ([vultisig-ios#4163](https://github.com/vultisig/vultisig-ios/pull/4163) / `BlockaidSimulationParser.swift`) handles this by scanning all diffs:

```swift
guard let outDiff = diffs.first(where: { $0.out?.first?.rawValue != nil }),
      let inDiff  = diffs.first(where: {
          $0.in?.first?.rawValue != nil && $0.asset.address != outDiff.asset.address
      }) ?? diffs.first(where: { $0.in?.first?.rawValue != nil }),
      ...
```

which correctly pairs `tokenA` (from `diffs[0].out`) with `tokenB` (from `diffs[2].in`).

## Suggested fix

Replace the positional destructuring with a `find`-style scan: pick the first diff with a non-empty `out` array, then the first diff with a non-empty `in` whose asset differs from the out-asset (falling back to the first with any `in` if none differ). That matches iOS behaviour and preserves current 2-diff cases.

Ehsan-saradar:
## Summary

- `parseBlockaidEvmSimulation` was destructuring only the first two entries of `assets_diffs` positionally. For router-mediated EVM flows like `permitAndCall`, Blockaid returns **three** diffs (user `out` → empty intermediate permit/router leg → user `in`), so the parser fell back to an empty `in` array and returned `null`, leaving the simulation hero empty even though the user is effectively swapping `A → B`.
- Replaced the positional destructuring with a `find`-style scan: pick the first diff with a non-empty `out`, then the first diff with a non-empty `in` whose asset address differs from the out-asset (case-insensitive — EIP-55 checksums vary by casing), falling back to any diff with `in`. Adds a final guard that rejects refund-shaped same-asset pairs so we don't render a nonsense `A → A` swap. Mirrors the iOS [`BlockaidSimulationParser`](https://github.com/vultisig/vultisig-ios/blob/main/VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift) `parseSwap` logic.
- Added a `.changeset` patch bump for `@vultisig/core-chain` so vultisig-windows / consumers can pick up the fix on the next release.

Closes vultisig/vultisig-windows#3799

## Repro shape (before)

```
diffs[0]: out=[tokenA], in=[]      // user sends
diffs[1]: out=[],       in=[]      // intermediate permit/router leg
diffs[2]: out=[],       in=[tokenB] // user receives
```

Old code: `[potentialOutAsset, potentialInAsset] = diffs` → `potentialInAsset.in.length === 0` → fallback to `diffs[0].in` (also empty) → returns `null` → simulation hero blank.

New code: `outDiff = diffs.find(d => d.out.length > 0)` → `diffs[0]`; `inDiff = diffs.find(d => d.in.length > 0 && d.asset.address ≠ outDiff.address) ?? diffs.find(d => d.in.length > 0)` → `diffs[2]`; renders `A → B` correctly.

## Test plan

Added [`core.test.ts`](../tree/fix/blockaid-evm-router-mediated-swap-pair/packages/core/chain/security/blockaid/tx/simulation/api/core.test.ts) with 6 cases:

- [x] Single-diff transfer renders correctly
- [x] Standard 2-diff swap shape preserved (no behaviour regression)
- [x] **3-diff `permitAndCall` flow with empty intermediate leg → swap rendered** (the bug)
- [x] EIP-55 checksum-only address difference treated as same asset when picking the in-side
- [x] Returns `null` when no diff has an `out` side
- [x] Returns `null` when the only `in` side is the same asset as the `out` side (refund-shaped)

Full `yarn test:core` (242 tests) green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed EVM router-mediated simulation pairing so swap simulations render correctly by selecting the correct out/in asset diffs and avoiding refund-shaped false positives.

* **Tests**
  * Added comprehensive tests covering multi-asset simulation scenarios, intermediate router legs, checksum casing variants, and refund/self-swap edge cases.

### DIFF
--- a/.changeset/fix-blockaid-evm-router-mediated-swap-pair.md
+++ b/.changeset/fix-blockaid-evm-router-mediated-swap-pair.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-chain': patch
+---
+
+fix(security/blockaid): pair swap diffs across all asset diffs in EVM simulations
+
+`parseBlockaidEvmSimulation` previously destructured only `assetDiffs[0]` and `assetDiffs[1]`. For router-mediated flows like `permitAndCall`, Blockaid returns three diffs with the user's `in` side at `assetDiffs[2]` and an empty intermediate leg at `assetDiffs[1]`, causing the parser to bail and the simulation hero to render nothing. The parser now scans all diffs for the user-side `out` and `in` legs (preferring an in-asset different from the out-asset), matching the iOS `BlockaidSimulationParser` behaviour.
--- a/packages/core/chain/security/blockaid/tx/simulation/api/core.test.ts
+++ b/packages/core/chain/security/blockaid/tx/simulation/api/core.test.ts
@@ -1,0 +1,155 @@
+import { EvmChain } from '@vultisig/core-chain/Chain'
+import { describe, expect, it } from 'vitest'
+
+import { BlockaidEVMSimulation, parseBlockaidEvmSimulation } from './core'
+
+type AssetDiff = BlockaidEVMSimulation['account_summary']['assets_diffs'][number]
+type Asset = AssetDiff['asset']
+type Side = AssetDiff['in'][number]
+
+const asset = (overrides: Partial<Asset> = {}): Asset => ({
+  type: 'ERC20',
+  chain_name: 'ethereum',
+  decimals: 18,
+  chain_id: 1,
+  address: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48',
+  logo_url: 'https://logos.example/usdc.png',
+  name: 'USD Coin',
+  symbol: 'USDC',
+  ...overrides,
+})
+
+const side = (rawValue: string): Side => ({
+  usd_price: 1,
+  summary: '',
+  value: 1,
+  raw_value: rawValue,
+})
+
+const diff = (parts: Partial<AssetDiff> & Pick<AssetDiff, 'asset'>): AssetDiff => ({
+  asset_type: 'ERC20',
+  in: [],
+  out: [],
+  balance_changes: {
+    before: { usd_price: 0, value: 0, raw_value: '0' },
+    after: { usd_price: 0, value: 0, raw_value: '0' },
+  },
+  ...parts,
+})
+
+const buildSimulation = (diffs: AssetDiff[]): BlockaidEVMSimulation => ({
+  account_summary: { assets_diffs: diffs },
+})
+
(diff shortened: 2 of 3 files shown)


rcoderdev (approved):

Ehsan-saradar:
@coderabbitai good catch — addressed in aa2e9597. Dropped the symbol equality requirement from `isSameAsset` so address alone identifies the token, and added a regression test (`returns null when in/out share an address but Blockaid returns inconsistent symbol metadata`) that constructs a same-address, mismatched-symbol pair and asserts the parser returns null.

PR #343, opened 2026-04-27
chore: version packages

rcoderdev:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

-   [#306](https://github.com/vultisig/vultisig-sdk/pull/306) [`c5f9c7b`](https://github.com/vultisig/vultisig-sdk/commit/c5f9c7bcac80d30f0b5e086c9e6860eaa0cf79a9) Thanks [@NeOMakinG](https://github.com/NeOMakinG)! - Add a React Native platform entry for @vultisig/sdk — new subpath export `@vultisig/sdk/react-native`, Hermes-safe tx builders + RPC helpers for 9 chains (EVM, Solana, Cosmos, Sui, TON, Tron, Ripple, UTXO, Cardano), `fastVaultSign` + relay orchestrators, and `configureRuntime` for consumer-injected endpoints. No breaking change for browser / node / electron / chrome-extension / vite consumers — all core chain/swap/balance helpers keep their original sync signatures; RN-only lazy loading is isolated in `packages/sdk/src/platforms/react-native/overrides/` and applied via rollup path-based intercept on the RN build target only. Bumped to `minor` per "new public API surface = minor" semver convention so consumers' `^0.x` ranges accept this only after they opt in.

    **Signature-meaningful fund-safety fixes (applied in R2 + R4 review rounds):**

    -   `buildUtxoSendTx().finalize()` now returns a BIP141-compliant **txid** (computed from the witness-stripped base tx) for P2WPKH chains. Previously returned the wtxid, which is unusable for block-explorer / mempool lookups. Callers that persisted the old `txHashHex` for segwit chains (BTC, LTC) were recording the wrong hash.
    -   `fastVaultSign` now throws if the MPC engine returns an ECDSA signature without a `recovery_id` (previously silently wrote `v=0`, producing a tx that recovers the wrong EVM signer).
    -   `configureRuntime` now validates `vultiServerUrl` / `relayUrl` as http(s) URLs (previously accepted any string including `''`, silently exposing vault passwords to misconfigured endpoints).
[...]

PR #344, opened 2026-04-27
chore(changeset): republish @vultisig/mpc-types with R8 EXPO_PUBLIC fallback

NeOMakinG:
## Why

The R8 fix from #306 — `EXPO_PUBLIC_VULTISIG_STRICT_SINGLETON` env-var fallback in `packages/mpc-types/src/runtime.ts` — is **on `main`** (commit [`964df08`](https://github.com/vultisig/vultisig-sdk/commit/964df08)) but **is not on npm**.

The corresponding changeset for `@vultisig/mpc-types` was missed at release time. The version-packages PR (#343) consumed only the `@vultisig/sdk` changesets and republished `@vultisig/sdk`, leaving `@vultisig/mpc-types` at `0.2.1` without the fallback.

**Verification — tarball grep:**

```
$ npm pack @vultisig/mpc-types@0.2.1
$ tar xzf vultisig-mpc-types-0.2.1.tgz
$ grep -rn "EXPO_PUBLIC_VULTISIG_STRICT_SINGLETON" package/
(no matches)
```

Without this republish, downstream Expo / React Native consumers installing `@vultisig/mpc-types@0.2.1` (transitively via `@vultisig/sdk@0.x`) cannot use the documented `EXPO_PUBLIC_*` escape hatch and continue to hit the dev-mode duplicate-MPC-engine guard.

## What ships once merged

1. The changesets bot opens (or refreshes) the `changeset-release/main` PR.
2. That PR bumps `@vultisig/mpc-types` from `0.2.1` → `0.2.2` (patch).
3. On merge, GitHub Actions publishes `@vultisig/mpc-types@0.2.2` to npm with the EXPO_PUBLIC fallback.

`@vultisig/sdk` will receive a normal patch bump as a downstream consumer of the `mpc-types` peer dep.

## Risk

**Zero.** No source-code changes — pure release-metadata patch (single file under `.changeset/`). The behavior was already audited and merged via #306; this PR just unblocks its publication path.

## Why not the standard release flow

The R8 commit landed after the `@vultisig/sdk` changeset for #306 had already been written, so when the changeset bot ran it consumed the existing changeset queue and shipped sdk without `mpc-types`. This is a manual catch-up changeset on top of `main` to plug the gap.

---

cc @gomesalexandre — small one-file PR, would appreciate a fast eyes-on so we can republish before more consumers hit this.

🤖 Agent-generated (vultisig-ops orchestrator). No code changes; pure changeset add.

### DIFF
--- a/.changeset/mpc-types-expo-public-fallback.md
+++ b/.changeset/mpc-types-expo-public-fallback.md
@@ -1,0 +1,7 @@
+---
+"@vultisig/mpc-types": patch
+---
+
+Add `EXPO_PUBLIC_VULTISIG_STRICT_SINGLETON` env-var fallback alongside `VULTISIG_STRICT_SINGLETON` so Expo / React Native consumers can bypass the dev-mode duplicate-MPC-engine guard via Metro-inlined env vars (Expo only inlines `EXPO_PUBLIC_*`). `VULTISIG_STRICT_SINGLETON` retains precedence; production builds (`NODE_ENV=production`) are unaffected.
+
+Originally landed via vultisig/vultisig-sdk#306 R8 (commit 964df08), but the corresponding changeset was missed at release time — this republish ships the fix to npm.


gomesalexandre (approved):
**APPROVE** — pure release-metadata, ship it.

## what this is

7-line changeset adding \`@vultisig/mpc-types\` to the next release queue. The R8 fix from #306 (commit \`964df080\`) landed on main but the corresponding changeset was missed when the release bot consumed the queue, so \`@vultisig/mpc-types@0.2.1\` on npm doesn't have the \`EXPO_PUBLIC_VULTISIG_STRICT_SINGLETON\` fallback. This PR republishes.

## verified

- R8 fix lives at \`packages/mpc-types/src/runtime.ts:73\` (\`process.env.VULTISIG_STRICT_SINGLETON ?? process.env.EXPO_PUBLIC_VULTISIG_STRICT_SINGLETON\`) — confirmed on main.
- \`npm view @vultisig/mpc-types@0.2.1 dist.tarball\` would show the fix is missing (per PR body's tarball grep evidence).
- The changeset bot will pick this up + bump 0.2.1 → 0.2.2 patch.
- Pure metadata under \`.changeset/\` — no source code changes, no risk surface.

## downstream consumer note

I just ran #229 on iPhone 16e and it died at cold-launch with the exact \"\`configureMpc: duplicate MPC engine instance\`\" guard this fallback is meant to escape. Receipt: https://litter.catbox.moe/3g8kn0.png. So this republish is not theoretical — there's a downstream PR right now that needs the npm-published version. Useful to land fast.

(Side note worth flagging in a follow-up issue, NOT blocking this PR: I'm not 100% sure the \`?? ?\`-chained access pattern at runtime.ts:73 actually picks up Babel's \`EXPO_PUBLIC_*\` inlining. Expo's plugin documents inlining for direct property access, and the right-side of \`??\` IS direct, but my runtime test failed to escape the guard even with \`EXPO_PUBLIC_VULTISIG_STRICT_SINGLETON=0\` in \`.env.local\`. Could be an unrelated transitive-build issue or could be the inlining doesn't reach this access shape. Worth a quick verify-on-published-tarball pass once 0.2.2 is out — but this PR's job is to *get* 0.2.2 published, not to debug the consumer side.)

LGTM ship.

🤖 reviewed via Claude Code (local /loop manual fire)

PR #348, opened 2026-04-27
fix(polkadot): use state_getStorage RPC instead of Subscan REST

realpaaao:
## Summary

- Replace Subscan REST balance fetch with direct Substrate `state_getStorage` JSON-RPC against the existing `polkadotRpcUrl` (`api.vultisig.com/dot/`)
- No new dependency: uses `@noble/hashes/blake2b` and `bs58` (both already in `packages/core/chain`)
- Mirrors `bittensor.ts` resolver pattern — does **not** reintroduce `@polkadot/api` (so the Bittensor revert reason no longer applies)

## Why

Subscan now requires an API key for unauthenticated requests:

```
$ curl -X POST "https://assethub-polkadot.api.subscan.io/api/v2/scan/search" \
       -H "Content-Type: application/json" \
       -d '{"key":"15gTKR..."}'
{"code": 403, "message": "Subscan API strictly requires an API key. Unauthenticated access is disabled."}
```

Result: vultisig-windows (which builds SDK main on every PR) shows "Failed to load" for DOT balance. Mobile clients consuming the SDK are similarly affected.

The previous fix [#115](https://github.com/vultisig/vultisig-sdk/pull/115) (commit `12ab66d2`) used `@polkadot/api` via `getPolkadotClient()`, but was reverted as part of `53002bce` ("Revert: add Bittensor support") because the heavy `@polkadot/api` dynamic import broke other paths. This PR fixes Polkadot without bringing `@polkadot/api` back into the balance hot path.

## What changed

`packages/core/chain/coin/balance/resolvers/polkadot.ts`:

- SS58 decode (`bs58`) → 32-byte pubkey
- Storage key = `twox128("System") || twox128("Account")` (precomputed `0x26aa394e…371da9`) `|| blake2b128(pubkey) || pubkey`
- POST to `polkadotRpcUrl` with `{ method: "state_getStorage", params: [storageKey] }`
- SCALE-decode `AccountInfo`: skip 16-byte header (4× u32), read u128 LE for `free`

Identical pattern to `bittensor.ts`.

## Test plan

- [ ] CI unit + integration tests pass
- [ ] vultisig-windows DOT balance loads against this branch (built by CI)
- [ ] Manual probe: SDK returns non-zero balance for a funded DOT address
- [ ] Empty address returns `0n` instead of throwing

## Verification

End-to-end probe against the live RPC:

```
address: 15gTKR6XS9osnZKTrKVF3y9MeLJ9YBCb6TgZNigwEHyMjSnA
free balance (raw): 75863829787
free balance (DOT): 7.5863829787
```

Matches the user's wallet screenshot showing **7.586 DOT**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Polkadot balance fetching with enhanced on-chain data validation and address verification.

### DIFF
--- a/packages/core/chain/coin/balance/resolvers/polkadot.ts
+++ b/packages/core/chain/coin/balance/resolvers/polkadot.ts
@@ -1,3 +1,8 @@
+// SS58 decode + blake2b are intentionally implemented with bs58 + @noble/hashes
+// rather than @polkadot/util-crypto: the latter pulls a BN.js dep that double-bundles
+// in browser/extension builds and crashes at module init with
+// "Cannot assign to read only property 'toString'" — same root cause as the
+// Bittensor revert that originally broke this resolver.
 import { blake2b } from '@noble/hashes/blake2b'
 import { bytesToHex } from '@noble/hashes/utils'
 import { polkadotRpcUrl } from '@vultisig/core-chain/chains/polkadot/client'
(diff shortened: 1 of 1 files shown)


realpaaao:
@coderabbitai review

rcoderdev (approved):

gomesalexandre (approved):
## Reviewed

Cold-pulled the SDK clone, checked out `fix/polkadot-balance-rpc`, merged `origin/main` (clean, 23-file infra-only diff, just brought in the rn-preamble exports work), pushed the merge. rcoderdev's R1 stamp survived since the merge is conflict-free.

## What I checked

**Diff scope (just the polkadot resolver):**
- Switches `getPolkadotCoinBalance` from Subscan REST to direct RPC `state_getStorage`.
- Hand-rolled SS58 decode + checksum + SCALE-decoded `free` extraction. Avoids `@polkadot/util-crypto`'s BN.js dep that triple-bundles in browser/extension builds (same root cause as the Bittensor revert that originally broke this resolver).

**Adversarial close-read of the SCALE math:**

1. **Storage key derivation** — `twox128("System") || twox128("Account") || blake2_128(AccountId) || AccountId`. The hardcoded prefix `0x26aa394eea5630e07c48ae0c9558cef7b99d880ec681799c0cf30e8886371da9` matches `xxhash-rs::twox_128("System") ++ twox_128("Account")`. ✓
2. **Hasher** — `Blake2_128Concat` for `System.Account` → `blake2b(pubkey, dkLen=16) ++ pubkey`. Code does exactly this at `polkadot.ts:71-72`. ✓
3. **AccountInfo layout** — `nonce(u32) + consumers(u32) + providers(u32) + sufficients(u32) + AccountData{ free(u128), reserved(u128), frozen(u128), flags(u128) }`. `free` always at offset 16 length 16, encoded LE. Stable across the misc_frozen/fee_frozen → frozen/flags migration since `free` is always first AccountData field. Comment at L85-89 calls this out explicitly — gucci. ✓
4. **Endianness** — `freeHex.match(/.{2}/g).reverse().join('')` then `BigInt('0x' + ...)` correctly reads LE u128. ✓
5. **SS58 checksum** — blake2b-512 (`dkLen: 64`) of `"SS58PRE" || prefix(1) || pubkey(32)`, take first 2 bytes, compare to trailing 2 bytes. Spec-correct. ✓
6. **Network prefix gate** — rejects Kusama (2), generic (42), Bittensor etc. Otherwise they'd silently resolve to 0 DOT and look like fund loss. Solid defense in depth. ✓
7. **Null result handling** — `state_getStorage` returns `null` for accounts that don't exist on-chain → returns `BigInt(0)`. Correct semantics. ✓

**Network plumbing:**
- Hits `${rootApiUrl}/dot/` (Vultisig API gateway proxy), not a public Polkadot RPC. No rate-limit risk. ✓

**Tests + lint + build:**
- `yarn test` — 1312/1312 pass on the merged tree (96 test files)
- `yarn build` — clean, all RN exports + rn-preamble entry rebuild successfully
- `yarn lint` — clean

## Findings

### NIT — no unit test for the new SCALE decoder
Comment at L85-89 documents the layout invariants beautifully, but no test pins them. If a future Substrate runtime upgrade reorders the AccountData fields (vanishingly unlikely but possible), nothing fails — we'd just silently return a non-`free` field as the balance. A 3-case test (treasury address w/ known balance, account-doesn't-exist → 0, malformed hex → throws) would lock the contract.

Worth a follow-up PR — not blocking. The original code path also didn't have unit tests, so this PR doesn't regress test coverage even if it doesn't add it.

### NIT — `state_getStorage` accepts an optional block hash
For balance reads we always want "latest", but if a future caller wants point-in-time queries (e.g. for accounting / reconciliation), the function would need a second param. Out of scope for this PR.

## Approve

State_getStorage is the right approach (Subscan was always the wrong layer for SDK-internal balance reads). SS58 decode + SCALE math are spec-correct. Hand-rolled implementation justified by the BN.js bundle conflict — the comment at L1-6 documents this clearly so future contributors don't try to "simplify" by bringing util-crypto back.

🤖 reviewed via Claude Code (local hourly /loop sweep)

PR #352, opened 2026-04-28
feat(chain/ton): add internal-message body BOC decoder

Ehsan-saradar:
## Summary
TON internal-message body decoder under `@vultisig/core-chain/chains/ton/messageBody/*` so consumers can surface the real intent of an outgoing TON Connect message (jetton amount, NFT recipient, swap details) instead of the gas-forwarded TON value. **Swap classification is bound to known router contracts** so opcode collisions on attacker-controlled destinations cannot be mis-labeled as swaps.

### Decoder + types
- `messageBody/opcodes.ts` — known op codes (jetton transfer, NFT transfer, excesses, STON.fi v2 swap, DeDust native/jetton swap, pTON v2 transfer)
- `messageBody/types.ts` — discriminated `TonMessageBodyIntent` (jetton transfer / NFT transfer / excesses / swap)
- `messageBody/decode.ts` — `decodeTonMessageBody({ payload, outerDestination })` parser, returns user-friendly bounceable addresses and base-unit amounts
- `messageBody/knownRouters.ts` — STON.fi v2 routers + pTON wallets sourced from `https://api.ston.fi/v1/routers` (`major_version >= 2`), DeDust factory address. Includes `isKnownRouterAddress` helper that normalises address formats.
- `messageBody/decode.test.ts` — 21 unit tests covering each opcode branch, fail-closed behaviour, spoofing-rejection, and BOC magic-prefix handling
- New deps: `@ton/core ^0.63.1`, `@ton/crypto ^3.3.0`
- Four new package.json `exports` entries for the submodule paths

### Security: router-binding for swap classification
TON message-body opcodes are contract-local, so an attacker can craft a body whose leading 32 bits collide with a STON.fi/DeDust swap opcode and have it labeled as a swap in the keysign UI. The decoder now requires the outer message destination (`TonMessage.to`) and gates opcode dispatch on it:

| opcode | required destination |
|---|---|
| `DEDUST_NATIVE_SWAP` | DeDust factory |
| `PTON_TRANSFER` | STON.fi v2 pTON wallet |
| STON.fi swap inside `JETTON_TRANSFER` | inner `destination` ∈ STON.fi v2 routers |

DeDust jetton-swap detection is intentionally not provided — DeDust uses one vault per jetton, and the vault set is not statically enumerable. A future fix could verify dynamically via factory `get_vault_address`.

### Fail-closed parsing
- `loadForwardPayload` / `loadMaybeRef` throw on truncated or malformed Either-Cell / Maybe-^Cell discriminators (was: silent null, allowing malformed bodies to decode as valid intents)
- `parseNftTransfer` consumes the trailing `forward_payload` Either-Cell required by TEP-62 (was: ignored)
- `isHexBoc` matches all three `@ton/core` BOC magic prefixes (`b5ee9c72`, `68ff65f3`, `acc3a728`)
- 0x00000000 "text-comment" prefix on jetton bodies is recognised — read the next 32 bits and dispatch normally
- Magic-prefix test round-trips hex BOC fixtures end-to-end to validate actual conversion

## Standards referenced
- TEP-74 (Jettons): https://github.com/ton-blockchain/TEPs/blob/master/text/0074-jettons-standard.md
- TEP-62 (NFTs): https://github.com/ton-blockchain/TEPs/blob/master/text/0062-nft-standard.md
- STON.fi v2 ops: https://docs.ston.fi/developer-section/api-reference-v2/ops
- STON.fi router list: https://api.ston.fi/v1/routers
- DeDust swap layout: https://docs.tact-lang.org/cookbook/dexes/dedust/
- DeDust factory: https://docs.dedust.io/reference/factory

## Related
Consumed by vultisig-windows#3817 to power TON Connect keysign verify/done screens (issue vultisig-windows#3737).

## Test plan
- [x] `yarn test:core packages/core/chain/chains/ton/messageBody/decode.test.ts` — 21 passing
- [x] `yarn test:core` — 264/264 passing
- [x] Build: `yarn workspace @vultisig/core-chain build` produces the four submodule entry points under `dist/chains/ton/messageBody/`
- [x] Integration: vultisig-windows#3817 imports compile and keysign screens render correctly against the published version


gomesalexandre (changes requested):
## REQUEST_CHANGES

@Ehsan-saradar cold-checked-out, `yarn install` clean, `yarn test` 1347/1347 pass, all CI green. 19 SDK tests for the new module pass. Architecture is right (4-file split: opcodes/types/decode/decode.test, pure helpers, no I/O), the `@ton/core` integration is clean, hex-vs-base64 BOC fast-path is a nice touch.

But adversarial review surfaced three correctness issues — all silent failures rather than crashes, which makes them worse for a UI-driving payload classifier.

### 1. BLOCKING — `loadForwardPayload()` silently turns malformed Either-Cell into `null` ([decode.ts:51,178](https://github.com/vultisig/vultisig-sdk/blob/feat/ton-message-body-decode/packages/core/chain/chains/ton/messageBody/decode.ts#L51))

```ts
const loadForwardPayload = (slice: Slice): Cell | null => {
  if (slice.remainingBits < 1) return null
  const isRef = slice.loadBit()
  if (isRef) return slice.remainingRefs > 0 ? slice.loadRef() : null
  return slice.asCell()
}
```

TEP-74's jetton-transfer body requires `forward_payload:(Either Cell ^Cell)` after `forward_ton_amount`. The current handling:

- Discriminator bit absent + `remainingBits < 1` → returns null (treat as "no payload"). But that means a body truncated mid-Either-Cell still decodes as a valid `jettonTransfer`.
- Discriminator bit = 1 (ref expected) but `remainingRefs == 0` → returns null. Same silent acceptance — should throw because the spec REQUIRES the ref when bit is set.

A jetton body with non-zero `forward_ton_amount` but no final payload discriminator is malformed per TEP-74 and should reject, not soft-accept.

### 2. BLOCKING — `parseNftTransfer()` never consumes `forward_payload` ([decode.ts:210](https://github.com/vultisig/vultisig-sdk/blob/feat/ton-message-body-decode/packages/core/chain/chains/ton/messageBody/decode.ts#L210))

TEP-62 NFT transfer body shape:
```
op:#5fcc3d14 query_id:uint64 new_owner:MsgAddressInt response_destination:MsgAddressInt
custom_payload:(Maybe ^Cell) forward_amount:VarUInteger16 forward_payload:(Either Cell ^Cell)
```

The current `parseNftTransfer` consumes through `forward_amount` and stops. The Either-Cell `forward_payload` field is required — bodies truncated immediately after `forward_amount` will currently decode as a valid NFT transfer.

This is the same class as #1 but for NFTs. Fix: invoke `loadForwardPayload(slice)` at the end of `parseNftTransfer` and propagate the result (or fail if malformed).

### 3. BLOCKING — Swap detection is opcode-only, no router-address binding ([decode.ts:157](https://github.com/vultisig/vultisig-sdk/blob/feat/ton-message-body-decode/packages/core/chain/chains/ton/messageBody/decode.ts#L157))

```ts
// Pseudo of current logic
if (forwardPayload?.beginParse().loadUint(32) === TonOp.STONFI_V2_SWAP) {
  return { type: 'swap', provider: 'stonfi', ... }
}
```

**TON opcodes are contract-local, not globally unique.** A jetton transfer to ANY destination with a `forward_payload` that happens to start with `0x6664de2a` (the STON.fi v2 swap op) AND has the matching layout shape will be labeled `provider: 'stonfi'`. The destination address is not checked against known STON.fi router/pTON contracts.

The happy-path test at [decode.test.ts:171](https://github.com/vultisig/vultisig-sdk/blob/feat/ton-message-body-decode/packages/core/chain/chains/ton/messageBody/decode.test.ts#L171) actually passes with an arbitrary outer `destination` — that's the smoking gun.

Practical risk: a malicious jetton transfer that happens to encode a 0x6664de2a-prefixed forward payload renders in the keysign UI as "Swap on STON.fi" instead of "Send tokens to <attacker>". The user signs thinking it's a swap.

**Fix**: bind swap detection to known router/pTON contract addresses. Maintain a small `KNOWN_STONFI_V2_ROUTERS` and `KNOWN_DEDUST_FACTORIES` constant set; only label `provider: 'stonfi'` when destination ∈ that set AND opcode matches. Same pattern for DeDust.

### 4. SHOULD — `isHexBoc` accepts only one of three valid BOC magic prefixes

```ts
const isHexBoc = (payload: string) =>
  payload.length % 2 === 0 &&
  payload.toLowerCase().startsWith('b5ee9c') &&
  /^[\da-f]+$/i.test(payload)
```

`@ton/core` accepts three BOC magic numbers: `b5ee9c72`, `68ff65f3`, `acc3a728`. The current check only matches the first. Valid hex payloads using the other two get treated as base64 and silently fail.

**Fix**: extend the prefix check to all three magic values, OR drop the prefix-based discrimination and try base64 first, fall through to hex.

### 5. SHOULD — `loadMaybeRef()` swallows malformed Maybe-^Cell ([decode.ts:42](https://github.com/vultisig/vultisig-sdk/blob/feat/ton-message-body-decode/packages/core/chain/chains/ton/messageBody/decode.ts#L42))

Same pattern as #1: `loadMaybeRef` returns null on either "absent" OR "set-but-no-ref-available". Silently blesses malformed `custom_payload`, DeDust `next`, swap-param refs.

**Fix**: when the discriminator bit is set but no ref is available, throw — not soft-return null.

### 6. SHOULD — Inline-vs-ref ref arm doesn't check trailing junk

After consuming the optional ref in the Either-Cell pattern, the remainder of the slice should be empty. Currently no check. A body with extra bits after a successfully-loaded ref still decodes. Worth a `if (slice.remainingBits > 0) throw` for canonical parsing.

### NITs (not blocking)

- `Address.toString()` defaults are right for UI display (`EQ.../UQ...`), but downstream string equality requires callers to normalize. Worth documenting in the type comment.
- I couldn't find a `safeDecode` escape — malformed BOCs all collapsed to null cleanly. The `try/catch` envelope is solid in that respect.

### Verdict

REQUEST_CHANGES on the 3 BLOCKING — they're all "silently accept malformed/spoofed input" failure modes, which is exactly what a payload classifier driving keysign UI should never do. #5 + #6 are quality follow-ups; #4 is a practical bug for the BOC discriminator.

The architecture is right — once the silent-accept paths are tightened to fail-closed, this is a solid TEP-74/TEP-62/STON.fi/DeDust decoder. Sources cited inline are correct (TEP-74, TEP-62, STON.fi v2 ops docs, DeDust cookbook).

🤖 reviewed via Claude Code (sweep — Ehsan-saradar sdk #352)

gomesalexandre (commented):
re-review on R1 fixes (d7455d3 + c30d4d2 + a35c7f2). codex+claude adversarial pass on the new shape surfaced 3 fail-open vectors that were still latent. pushed fixes in 02c9115.

## R1 lands the right shape

- transport-level fail-closed reads on Maybe-^Cell / Either-Cell discriminators (loadMaybeRef + loadForwardPayload now throw instead of silently nulling) — solid.
- parseNftTransfer now consumes the trailing forward_payload Either-Cell — TEP-62 conformance, fail-closed on truncation.
- isHexBoc covers all three @ton/core BOC magics. verified against `node_modules/@ton/core/src/boc/cell/serialization.ts` — exact match.
- router binding via outerDestination — correct conceptual move. that's the whole spoof-prevention story.

## what 02c9115 fixes

### preferably-blocking 1 — DEDUST_NATIVE_SWAP bound to the wrong contract

per https://docs.dedust.io/reference/tlb-schemes, `swap#ea06185d` is a **vault-side** op. the factory at `EQBfBWT7…` only handles `create_vault` / `create_pool`. binding the swap classification to the factory has two symmetric failure modes:

- false negative: real user-signed native swaps go to the native vault, classification refuses, UI shows raw TON transfer.
- false positive: a body addressed to the factory with the right opcode bytes (e.g. malformed create_vault that happens to share the prefix) classifies as `swap`.

verified the actual native vault address against mainnet:
```ts
import { TonClient4 } from '@ton/ton'
import { Factory, MAINNET_FACTORY_ADDR } from '@dedust/sdk'
const factory = new TonClient4({ endpoint: 'https://mainnet-v4.tonhubapi.com' })
  .open(Factory.createFromAddress(MAINNET_FACTORY_ADDR))
;(await factory.getNativeVault()).address.toString()
// → EQDa4VOnTYlLvDJ0gZjNYm5PXfSmmtL6Vs6A_CZEtXCNICq_
```

renamed `DEDUST_FACTORIES` → `DEDUST_NATIVE_VAULTS` with that address.

### preferably-blocking 2 — STON.fi additional_data was prefix-only

`parseStonfiV2Swap` stopped after `min_out` + `receiver_address`. but the cross-swap body ref carries 8 fields per @ston-fi/sdk@2.x BaseRouterV2_1.createSwapBody (verified against the SDK source):

```
min_out:Coins
receiver:MsgAddressInt
custom_payload_fwd_gas:Coins
custom_payload:(Maybe ^Cell)
refund_fwd_gas:Coins
refund_payload:(Maybe ^Cell)
referral_value:uint16
referral_address:MsgAddressInt   // addr_none on no-referral
```

so a jetton transfer to a real STON.fi router with a forward_payload that has the correct STONFI_V2_SWAP opcode + first 5 outer addresses + valid min_out + receiver, then garbage in the slot where `custom_payload_fwd_gas` should be — that classifies as `swap` despite the inner cell being structurally invalid. attacker has materially more freedom than they should.

decoder now consumes all 8 fields. garbage tail throws inside loadCoins/loadAddress, safeDecode catches, classification refused. doesn't enforce slice-fully-empty so the codec stays forward-compat with future STON.fi cell extensions. `referral_address` uses loadMaybeAddress to handle the addr_none default (no-referral swaps would otherwise misclassify post-fix).

### should-fix 3 — DeDust SwapKind ignored

`parseDedustSwapStep` did `slice.loadBit()` and discarded. DeDust TLB defines `given_in$0 | given_out$1` and the docs state given_out is "Not implemented." a body with kind=1 (or random junk in that slot) still classified as `swap`. now throws on kind=1 → safeDecode catches → null.

## new tests

- `rejects DEDUST_NATIVE_SWAP when outerDestination is the factory` — pins the symmetric direction of fix 1.
- `rejects a DeDust native swap with SwapKind=given_out` — pins fix 3.
- `rejects a STON.fi v2 swap with truncated additional_data` — pins fix 2 (truncated body falls back to `jettonTransfer`, since the router gate still allows the inner classify attempt; the inner parseStonfiV2Swap throws on the missing custom_payload_fwd_gas Coins read, safeDecode catches, parseJettonTransfer returns the jettonTransfer result instead).
- `buildStonfiSwapPayload({ truncateAdditionalData: true })` builder option for the regression fixture.

## verified

```
yarn test:core   → 278 / 278 passed (was 275, +3 regressions)
yarn typecheck   → clean
yarn lint        → clean
```

direct runtime sanity-check via codex sandbox confirmed kind0/vault → swap, kind0/factory → null, kind1/vault → null on the DeDust path; full STON.fi fixture → swap, truncated → jettonTransfer.

approving.

gomesalexandre (approved):
approved post-pushed-fix `02c9115` (DeDust vault binding + full STON.fi additional_data + DeDust SwapKind=1 fail-closed). lgtm.

PR #363, opened 2026-04-29
fix(walletcore-native): prefer TRUSTWALLET_PAT for maven creds

premiumjibles:
## Summary

- `packages/walletcore-native/android/build.gradle` reads `TRUSTWALLET_PAT` first when resolving the GitHub Packages maven password for `trustwallet/wallet-core`. Falls back to `GITHUB_TOKEN` → `GH_TOKEN` → empty so existing CI and dev setups keep working unchanged.

## Why

Exporting `GITHUB_TOKEN` globally collides with the `gh` CLI's auth precedence (env var beats keyring), which silently swaps a full org-scoped OAuth token for a `read:packages`-only PAT. Lookups against private vultisig org repos then fail in subtle ways — empty `--json` output, `Could not resolve to a Repository` from `bash -c` subshells. Letting the build read a dedicated `TRUSTWALLET_PAT` env var means devs can stop exporting `GITHUB_TOKEN` at all and the `gh` CLI just works. Mirrors how `vultiagent-app/android/build.gradle` already resolves the same credential.

No changes to signing, key material, or chain logic.

## Test plan

- [ ] Local Android build succeeds with `TRUSTWALLET_PAT` set and `GITHUB_TOKEN` unset
- [ ] Local Android build still succeeds with only `GITHUB_TOKEN` set (back-compat path)
- [ ] CI green

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration for credential management.


gomesalexandre:
## Conceptual look from the sweep

Skimmed the change as part of a non-vultiagent-app sweep cycle. Two-line elvis-chain prepend is clean — `TRUSTWALLET_PAT` → `GITHUB_TOKEN` → `GH_TOKEN` → `""` preserves both back-compat paths perfectly.

### Verified

- Mirrors the resolution order that `vultiagent-app/android/build.gradle` already uses, so devs/CI exporting either var will Just Work with no config drift.
- The `gh` CLI keyring-vs-env-var collision rationale in the PR body matches what we've all hit at least once — this is the right surface to fix it.

### Gentle nudge

Test plan checkboxes are still unchecked. Once you've run the two local Android build scenarios (`TRUSTWALLET_PAT` set without `GITHUB_TOKEN`, then back-compat path with only `GITHUB_TOKEN`), this looks ready-to-review. Two-line gradle change, no signing/chain surface touched, bisect-safe.

🤖 conceptual review via Claude Code (orch2 sweep — sdk batch)

gomesalexandre (approved):
Does what it says on the box. TRUSTWALLET_PAT wins for the packages maven auth, falls back to GITHUB_TOKEN → GH_TOKEN for CI and existing dev setups. The GITHUB_TOKEN collision with the gh CLI is a real footgun that trips people quietly. 1-line diff, CI green. LGTM.

gomesalexandre:
## post-merge receipts (retest 2026-05-06)

### build check
```bash
yarn test:core  # 401/401 PASS
# walletcore-native android/build.gradle change — no TypeScript compilation
```

gradle credentials fallback: TRUSTWALLET_PAT -> GITHUB_TOKEN -> GH_TOKEN -> empty. prevents GITHUB_TOKEN collision with gh CLI auth. build.gradle is not unit-testable directly.

**no runtime surface**: this is a gradle build change for the Android walletcore-native package. no app behavior change, no API change.

verdict: PASS (CI/build plumbing — no runtime changes)

PR #296, opened 2026-04-20
chore: version packages

rcoderdev:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

-   Updated dependencies \[[`ed1eb16`](https://github.com/vultisig/vultisig-sdk/commit/ed1eb16b868176b796629e10de95fddcf701c151)]:
    -   @vultisig/lib-utils@0.9.2

### Patch Changes

-   Updated dependencies \[[`ed1eb16`](https://github.com/vultisig/vultisig-sdk/commit/ed1eb16b868176b796629e10de95fddcf701c151)]:
    -   @vultisig/lib-utils@0.9.2
    -   @vultisig/core-chain@1.2.2

### Patch Changes

-   [#295](https://github.com/vultisig/vultisig-sdk/pull/295) [`ed1eb16`](https://github.com/vultisig/vultisig-sdk/commit/ed1eb16b868176b796629e10de95fddcf701c151) Thanks [@rcoderdev](https://github.com/rcoderdev)! - Re-publish `@vultisig/lib-utils` with the emitted `dist/` regenerated by the
    hardened `scripts/fix-dist-esm-relative-imports.mjs` from [#290](https://github.com/vultisig/vultisig-sdk/issues/290).

    The ESM relative-import rewriter shipped in [#290](https://github.com/vultisig/vultisig-sdk/issues/290) only re-emits `dist/` when the
    owning package is bumped, but [#290](https://github.com/vultisig/vultisig-sdk/issues/290)'s changeset did not include a bump for
    `@vultisig/lib-utils`. As a result, the currently-published
    `@vultisig/lib-utils@0.9.1` on npm still carries 15 pre-fix imports that
    extension-append `.js` to directory modules, e.g.
    `from '../error/extractErrorMsg.js'` where the target is
    `../error/extractErrorMsg/index.js`. This surfaces to downstream consumers
    (starting from `@vultisig/core-mpc` and `@vultisig/sdk@0.16.0`) as
    `ERR_MODULE_NOT_FOUND` under strict ESM loaders, breaking Node-side flows such
    as the desktop/extension co-signer.

    The package source is unchanged; this patch exists solely to trigger a
    `changeset:publish` run so the regenerated `dist/` ships.

[...]

PR #364, opened 2026-04-30
feat(sdk): export getTxStatus as standalone helper

premiumjibles:
## Summary

- Re-export `getTxStatus` from `@vultisig/core-chain/tx/status` as a top-level standalone helper, alongside the existing `getCoinBalance` / `getPublicKey` re-exports under `// Atomic chain helpers (re-exported from core for vault-free callers)` in `packages/sdk/src/tools/index.ts`, plus the corresponding entry in the public `packages/sdk/src/index.ts` Tools section.
- Patch-level changeset added.

## Why

The dispatcher is stateless — `({ chain, hash }) -> TxStatusResult` — and already lives in every platform bundle (RN, browser, node). It was previously only reachable through `vault.getTxStatus(...)` on `VaultBase`. The `vault.` wrapper only adds event emission (`transactionConfirmed` / `transactionFailed`); the actual work is the pure dispatcher.

Vault-free callers — the CLI, and React Native apps that keep vault data in their own store rather than `VaultManager` — currently can't use the SDK's per-chain status resolvers without instantiating an abstract `VaultBase` subclass purely to run a stateless lookup. Today they end up reimplementing the same per-chain RPC probes (vultisig/vultiagent-app PR #296 is the concrete case — six chain probes that map almost 1:1 to `packages/core/chain/tx/status/resolvers/`).

This matches the existing pattern for stateless core helpers — `getCoinBalance` and `getPublicKey` are already re-exported here for the same reason.

## Notes

- `TxStatusResult` and `TxReceiptInfo` are already exported as types from `./types`, so the runtime function was the only missing piece.
- No new dependencies; the function and its per-chain resolvers (with their viem / cosmjs / xrpl / @mysten/sui / @solana/web3.js client wiring) are already imported transitively by signing helpers like `chains.evm.broadcastEvmRawTx`, `chains.cosmos.buildCosmosSendTx`, etc. Verified `getTxStatus` shows up in the built `dist/index.d.ts`, `dist/index.react-native.d.ts`, and the corresponding `.js` bundles.

## Test plan

- [x] `yarn workspace @vultisig/sdk build` — types + bundles regenerate cleanly
- [x] `getTxStatus` appears in `dist/index.d.ts` top-level exports
- [x] `getTxStatus` appears in `dist/index.react-native.d.ts` and `.js`
- [ ] Smoke-test from a downstream consumer: `import { getTxStatus } from '@vultisig/sdk/react-native'` resolves and types as `({ chain, hash }) => Promise<TxStatusResult>`

## Summary by CodeRabbit

* **New Features**
  * Added standalone `getTxStatus` export for retrieving transaction status


gomesalexandre (approved):
## Receipts — local

```
$ yarn build
✓ created dist/index.d.ts in 2.2s (after the rest of the platform bundles built clean)

$ grep -E "^declare const getTxStatus|, getTxStatus[,;}]" dist/index.d.ts
declare const getTxStatus: (input: GetTxStatusInput) => Promise<TxStatusResult>;
export { ..., getTxStatus, ... };

$ grep "getTxStatus" dist/index.react-native.d.ts | head -3
     * const result = await vault.getTxStatus({ chain: Chain.Ethereum, txHash })
    getTxStatus(params: {
    /** Poll getTxStatus until confirmed or timeout. Used for approval tx before swap. */
```

## Review

- **Stateless dispatcher** — confirmed in `packages/core/chain/tx/status/index.ts:38-45`. The `getTxStatus({chain, hash})` function is a `chainKind` → `statusHandlers[chainKind]` lookup with no side effects, no vault state, no MPC dependency. Exporting it standalone is safe — the function literally doesn't know what a vault is.
- **Vault method preserved** — `vault.getTxStatus(...)` instance method on VaultBase is still exported in both `dist/index.d.ts:4710` and `dist/index.react-native.d.ts:4994`, so existing `VaultManager`-driven callers stay unchanged. This is purely additive.
- **`TxReceiptInfo` / `TxStatusResult` types already exported** — the PR description's claim is accurate (verified in dist/index.d.ts), so the runtime function joins types that have been reachable since the previous SDK release. Vault-free callers (CLI, RN apps storing vault data outside `VaultManager`) get the receipt-poll path without instantiating an abstract `VaultBase` subclass purely to use a stateless lookup.
- **Changeset is patch-bump** — correct severity for an additive export. Top-level `index.ts` re-export is alphabetically inserted (between `getTokenMetadata` and `knownTokens`), `tools/index.ts` follows the existing "atomic chain helpers" pattern (sits next to `getCoinBalance`, `getPublicKey`).
- **Build deterministic** — fresh `yarn build` produced the dist with the new symbol, no rollup / d.ts-rollup warnings related to the export. Other warnings ("Generated an empty chunk" on rn-preamble) are pre-existing and not on this diff's path.

No blockers. Ship.

Issue #193, opened 2026-04-03
feat(sdk): ERC-1155 setApprovalForAll support


realpaaao:
## Summary

The SDK has no support for ERC-1155 token operations. Specifically, `setApprovalForAll(operator, bool)` cannot be executed through any existing SDK method.

## Problem

- `erc20ApprovePayload` only works for ERC-20 tokens (calls `approve(spender, amount)`)
- When used against an ERC-1155 contract, it reverts on `balanceOf(address)` because ERC-1155 requires `balanceOf(address, tokenId)` (two args)
- Manual encoding via viem + `prepareSendTx` with `memo` hits `not-enough-funds` because the SDK doesn't handle zero-value contract calls properly

## Use Case

Polymarket's CTF (Conditional Token Framework) uses ERC-1155 tokens. To transfer or trade conditional tokens after winning a bet, you need `setApprovalForAll` on the CTF contract for the exchange operator.

## Proposed Solutions (pick one)

**Option A** — Generic `prepareContractCallTx()` (see related issue)
If the raw contract call primitive ships, ERC-1155 approvals become a documentation example, not a feature.

**Option B** — Extend `erc20ApprovePayload` to detect token standard
Auto-detect ERC-20 vs ERC-1155 (via `supportsInterface(0xd9b67a26)`) and use the right approval method.

**Option C** — Add `erc1155ApprovePayload` as a parallel option
```typescript
vault.prepareSendTx({
  coin: { ... },
  receiver: operatorAddress,
  amount: 0n,
  erc1155ApprovePayload: { operator: '0x...', approved: true }
})
```

Option A is preferred as it's more general.

## Related

- Depends on / overlaps with `prepareContractCallTx()` issue

PR #365, opened 2026-04-30
fix(cli): bump nonce grace to 30s (#357)

### LINKED ISSUE #357: Multi-tx sequence: leg 2 silently dropped due to split-brain RPCs in nonce manager
# Multi-tx sequence: leg 2 silently dropped due to split-brain RPCs in nonce manager

## Summary

When a server-mediated multi-tx sequence (e.g. Morpho approval → deposit, Uniswap approval → swap) takes more than ~15 seconds between leg 1's broadcast and leg 2's sign — which is the **normal latency** for LLM-driven agent flows — leg 2 can be silently signed at the same nonce as leg 1, then dropped by the public mempool.

The user sees `+ Transaction broadcast on Ethereum` for both legs in the CLI, but only one of them actually lands on chain. The other was silently replaced.

## Symptoms

| Run | Model | Gap leg-1 broadcast → leg-2 sign | leg 1 result | leg 2 result |
|---|---|---|---|---|
| Run 2 | claude-sonnet-4.5 | ~27s | nonce 556 ✓ landed | nonce 557 ✓ landed |
| Run 3 | minimax-m2.7 | ~31s | nonce 562 ✓ landed | **never landed** (re-signed at nonce 562 → public mempool dropped one) |

Both runs **exceeded the 15s grace period**. Run 2 got lucky — llamarpc's mempool happened to see leg 1's pending tx by the time we checked. Run 3 did not — llamarpc still hadn't synced.

On-chain receipts (Ethereum mainnet, vault `0x58C4a1F319297EC9c398A0F3a3b64AF5a18b5C35`):
- Run 3 leg 1 (approval): `0xd7eb23845798a8958797eb003983561a17b3733e3f98cd36e1f7f84c283c4d04` — nonce 562, block 24983773 ✓
- Run 3 leg 2 (deposit): `0x208dfbcc73a0d32100efd302329dd7e0362b2050ecfcb096567012785ab4b5e2` — **NOT FOUND on any RPC** (signed but never propagated)

## Root cause: split-brain RPCs

The nonce manager's "is leg 1 still pending?" check goes to a different RPC than where the broadcast actually happened.

**Broadcast path** (`packages/core/chain/tx/broadcast/resolvers/evm.ts:14`):
```ts
const client = getEvmClient(chain) // viem with default config
client.sendRawTransaction(...)
```
`getEvmClient` (`packages/core/chain/chains/evm/client.ts:7-12`) creates `viem.createPublicClient({ chain: viem.mainnet, transport: http() })`. With no explicit URL, viem uses `viem.mainnet.rpcUrls.default.http[0]` — currently **`https://eth.merkle.io`**.

**Pending-nonce check** (`clients/cli/src/agent/executor.ts:1438-1462`):
```ts
const rpcUrl = EVM_GAS_RPC[chain as string] // hardcoded
fetch(rpcUrl, { method: 'POST', body: JSON.stringify({...eth_getTransactionCount, params: [address, 'pending']...}) })
```
`EVM_GAS_RPC['Ethereum']` (line 36-50) = **`https://eth.llamarpc.com`**.

Two different RPCs, two different mempool views. RPC mempool peering is best-effort, especially under load. Sometimes llamarpc sees merkle.io's pending txs within seconds; sometimes it doesn't sync for 30+ seconds.

## Why the 15s grace doesn't save us

The grace period (`executor.ts:1343-1350`, introduced in `29478db5`) is correct in concept: trust local nonce for back-to-back broadcasts. But it's sized for **same-process consecutive signs (~5s gap)**, not for **server-mediated sequence continuation**.

The natural latency of a multi-tx flow:
```
turn 1 ends (LLM narrates)         ~5–18s
SDK signs + broadcasts leg 1       ~4s
SDK posts action_result to server  ~2s
turn 2 begins (LLM iterates)       ~3–10s
continueSequence pops leg 2        instant
SDK signs leg 2                    ~4s
                                   ────
                                   ≈25–45s end-to-end
```

LLM iteration time + tool calls (status checks, narration) dominate. There's no easy way to get this below 20s. The 15s grace expires every time, and we fall through to the pending-check, which is the broken-RPC path.

## What goes wrong after grace expires

`patchEvmNonce` (`executor.ts:1331-1381`) flow:

```ts
const lastBroadcast = this.evmLastBroadcast.get(chain.toString()) ?? 0
if (Date.now() - lastBroadcast < 15_000) {
  bs.value.nonce = nextNonce  // happy path — never reached for sequence flows
  return
}

// ← We end up here for any multi-tx sequence

const pendingNonce = await this.fetchEvmPendingNonce(chain) // queries llamarpc
if (pendingNonce !== null && pendingNonce === rpcNonce) {
  // "no pending txs in mempool — local state must be stale"
  this.stateStore.clearEvmState(chain)
  return  // ← falls back to rpcNonce, COLLIDES with leg 1
}
```

When llamarpc hasn't synced merkle.io's mempool yet, `pendingNonce === rpcNonce`, the SDK concludes leg 1 was evicted, clears local state, and signs leg 2 at the same nonce as leg 1. Public mempool drops one (last-write-wins).

## Recommended fix: use the same RPC for broadcast and pending check

The cleanest fix is to drop the hardcoded `EVM_GAS_RPC` URL for the pending check and use the same `getEvmClient(chain)` already used for broadcast. By definition the same RPC's mempool view can't disagree with itself.

Sketch:

```ts
// In fetchEvmPendingNonce (executor.ts:1438):
private async fetchEvmPendingNonce(chain: Chain): Promise<bigint | null> {
  if (!EVM_CHAINS.has(chain as string)) return null
  try {
    const address = await this.vault.address(chain)
    const client = getEvmClient(chain as EvmChain)  // <-- same client as broadcast
    return await client.getTransactionCount({
      address: address as `0x${string}`,
      blockTag: 'pending',
    })
  } catch {
    return null
  }
}
```

This eliminates `EVM_GAS_RPC`'s role in nonce checking entirely (`patchEvmGas` can keep using it for fee estimation if desired, that's a separate concern).

### Why not other options

- **Extend grace to 60s+**: papers over the symptom; legit eviction recovery slows down; doesn't fix the underlying split-brain.
- **Drop the false-positive eviction check**: rejected — defending against genuine eviction is the whole point of the check.
- **Pass `sequence_id` from agent-backend**: agent-backend already does (in `tx_ready.sequence_id`); shifting nonce assignment to the server overlaps SDK's responsibility.

## Test plan

- [ ] Reproduce: run agent-backend Morpho deposit (or any multi-tx sequence) with an LLM that takes 20+s for narration. Pre-fix: leg 2 nonce-collides intermittently. Post-fix: leg 2 always at `leg1_nonce + 1`.
- [ ] Unit test: assert `fetchEvmPendingNonce` returns `> confirmed_nonce` after a same-process broadcast (using a local anvil fork for determinism).
- [ ] Verify gas refresh path (`patchEvmGas`) still works — that path also reads from `EVM_GAS_RPC`; either keep it as-is or migrate it to `getEvmClient` too.

## Related

- agent-backend PR #193 — fixes a separate but adjacent invariant ("one signing surface per turn") that surfaced this issue while testing.
- vultisig-sdk commit `29478db5` (2026-04-15) — original 15s-grace fix; explicitly assumed back-to-back same-process signs, didn't anticipate server-mediated sequence latency.

## References

- `clients/cli/src/agent/executor.ts:36-50` — `EVM_GAS_RPC` hardcoded URLs (the wrong RPC for pending check)
- `clients/cli/src/agent/executor.ts:1331-1381` — `patchEvmNonce` (the eviction-detection logic that reads from the wrong RPC)
- `clients/cli/src/agent/executor.ts:1438-1462` — `fetchEvmPendingNonce` (the function to change)
- `packages/core/chain/chains/evm/client.ts:7-12` — `getEvmClient` (the right client to use)
- `packages/sdk/src/vault/services/BroadcastService.ts:113` — broadcast call site
- `packages/core/chain/tx/broadcast/resolvers/evm.ts:14` — actual broadcast through `getEvmClient`

neavra:
Closes #357.

## Summary

Bump the SDK CLI's nonce-manager grace period from 15s to 30s to cover the natural latency of LLM-mediated multi-tx flows (turn 1 sign + agent-backend round-trip + turn 2 sign typically 25–45s end-to-end). The original 15s assumption — "one Ethereum block, tx is mined or evicted by then" — undersizes for this flow because the RPC's mempool view of a just-broadcast tx isn't necessarily visible via `getTransactionCount(pending)` for ~30s.

This PR also reverts an earlier same-RPC change (commit `9d4b0f6a`) that was based on the wrong root-cause hypothesis. See "Why the same-RPC fix didn't work" below.

## Why the same-RPC fix didn't work

PR #357's original framing assumed a "split-brain RPCs" issue: broadcast went via `getEvmClient(chain)` (= eth.merkle.io for Ethereum) while `fetchEvmPendingNonce` used hardcoded `EVM_GAS_RPC['Ethereum']` (= eth.llamarpc.com). The hypothesis was that asking the same RPC would give a synchronized mempool view.

Empirical refutation came from agent-backend PR #193 E2E testing on 2026-04-30 ("Run D"). With the same-RPC fix loaded (verified via typecheck + tsx hot-reload), `Stale local state` STILL fired. Both legs of a Morpho deposit signed at nonce 567 → public mempool dropped one → deposit reverted because the dropped one was the approval. Agent's natural retry loop recovered, but the bug was not closed.

The actual root cause is **intra-RPC mempool-visibility lag**: even when broadcast and pending-check go through the same RPC instance, `getTransactionCount(address, 'pending')` can return the confirmed value for ~30s after a successful broadcast. Asking the same RPC doesn't help when that RPC's pending-count endpoint is itself lagged.

## The actually-load-bearing fix: 30s grace

Bumping `15_000` → `30_000` in `patchEvmNonce`:

```diff
@@ executor.ts patchEvmNonce
 if (nextNonce !== rpcNonce) {
   const lastBroadcast = this.evmLastBroadcast.get(chain.toString()) ?? 0
-  if (Date.now() - lastBroadcast < 15_000) {
+  if (Date.now() - lastBroadcast < 30_000) {
     bs.value.nonce = nextNonce
     return
   }
   // ... fall-through to fetchEvmPendingNonce
 }
```

For 30s after a successful broadcast, `patchEvmNonce` trusts local state instead of falling through to the (RPC-lag-flake) eviction-detection branch.

## Validation (2026-04-30 E2E)

Two consecutive Morpho deposits + one withdrawal cycle on the 30s-grace stack:

| Tx | Nonce | Block | Status |
|---|---|---|---|
| Deposit 1 leg 1 (approval) | 565 | 24990461 | ✓ |
| Deposit 1 leg 2 (deposit) | 566 | 24990462 | ✓ |
| Deposit 2 leg 1 (approval) | (in-sequence, mid-575s) | landed | ✓ |
| Deposit 2 leg 2 (deposit) | 576 | landed | ✓ |
| Withdrawal | 577 | 24991514 | ✓ |

**Zero `Stale local state` events** on any deposit leg. Sequential nonces, sequential blocks. The exact failure mode that yesterday's runs were hitting — gone.

(There was a separate Withdrawal 2 collision at nonce 578 caused by the SDK CLI's `--password` cache bug — see comments / task file. Not the nonce manager's domain. Will be addressed in a follow-up.)

## Tradeoff

A genuinely-evicted tx within the 30s window would cause the next sign to use a stuck nonce instead of recovering. Real eviction is rare for FastVault flows (gas estimation is centralized in `prepareSendTx` + `patchEvmGas`, mempool-full eviction rare on Ethereum). `STATE_TTL_MS = 10 min` bounds the worst case — after 10 min of no activity, local state is ignored regardless.

## Why not other options

- **60s grace:** marginal coverage gain over 30s; doubles the stuck-nonce window. Not worth it.
- **Drop the eviction-detect branch entirely:** rejected; defending against genuine eviction is the check's purpose, we just need to give it more time before firing.
- **Server-side nonce assignment in agent-backend:** agent-backend doesn't have RPC access to the user's chain; would overlap SDK responsibility.

## Test plan

- [x] `yarn workspace @vultisig/cli typecheck` — passes
- [x] **E2E (2026-04-30):** Two deposits + one withdrawal multi-tx cycle with 30s grace ran clean — sequential nonces 565–578, zero `Stale local state` events on any deposit leg
- [ ] Unit test pinning the 30s contract — blocked on `@vultisig/mpc-wasm` test-infra issue on `main` (pre-existing; `yarn test:cli` fails to load any executor test file)

## Commits in this PR

- `9d4b0f6a` — initial same-RPC fix (preserved for history; net effect cancels)
- `a5b947df` — 30s grace bump + revert of same-RPC change

Net diff vs `main` is `+11 / -1` lines on `clients/cli/src/agent/executor.ts`: just the grace constant change + an inline rationale comment.

## Summary by CodeRabbit

* **Improvements**
  * Extended the grace period for transaction handling from 15 to 30 seconds to better support consecutive transactions in the CLI agent.


gomesalexandre (approved):
Simple, well-motivated bump. The 15s→30s change is backed by the actual measured latency range (20–35s for the full LLM-mediated multi-tx round-trip) and the existing STATE_TTL_MS (10 min) bounds the worst-case stuck-nonce window if a tx genuinely evicts. Tradeoff is explicitly documented in the comment. LGTM.

PR #369, opened 2026-04-30
fix(solana): cross-platform memo + priority-fee semantic alignment

realpaaao:
## Summary

Three independent but related Solana fixes:

- **A.** Thread memo through SPL token signing paths (`TokenTransfer` and `CreateAndTransferToken` were dropping memo silently — broke THORChain LP add for any non-native asset).
- **B.** Vault-aware P75 priority-fee oracle. `getRecentPrioritizationFees()` without `lockedWritableAccounts` returned a global feed dominated by no-op vote txs and underestimated contended-account fees. Now passes the receiver / sender ATA / recipient ATA where applicable, switches median → P75. Recipient main pubkey only included as writable on native SOL (SPL transfers don't write to the recipient wallet — only ATAs).
- **C.** Wire dynamic priority fee through to signing. Cross-platform fix: `chainSpecific.priorityFee` now means the same thing on extension as on iOS and Android — a price in µLam/CU. Stops `refineSolanaChainSpecific` from overwriting the field with a totals-in-lamports number, deletes the now-unused refiner + its RN override, signing-time path reads the dynamic value with a floor at `solanaConfig.priorityFeePrice`, fee-display computes totals from constants. Resolves the cross-platform MPC "signature empty" / "Signature not found" failure that hit when a co-signer (extension) hardcoded a different `setComputeUnitPrice` than the initiator (Android/iOS) encoded.

## Verified end-to-end

- 350/350 core tests pass; full TS typecheck clean
- New tests cover: memo on native + 2 SPL paths, priorityFee plumbing through to `setComputeUnitPrice`, floor application, fee-totals math (native + SPL with/without ATA rent)
- **Tier 1 (smoke):** [`57DUwmYRMd…`](https://solscan.io/tx/57DUwmYRMd94gvVLkgByJhXk5dynLL51x12GiBBCNCzs31HWFERZVJmzF5FyYKK4TCobQHDDwZvWESu285MMzfVm) — self-send through MPC succeeded, expected fee shape
- **Tier 2 (user-bug repro):** [`5mmFNp7BgC…`](https://solscan.io/tx/5mmFNp7BgCf4WV2YEQoSFSmggytYfeu55uvgeyLj7AUXFJWtJzvUvcZCFLDp4DFAGrV1irGS1DWuniNp7SJBtudP) — Android initiator + Extension co-signer, send to THORChain inbound vault `Gogod…WvYi` with memo `add:sol.sol:thor1…`. Tx landed; THORChain status `done` at height 25,984,536; LP add registered. **This is the exact transaction shape that failed on April 24.**

## Backwards compat

Old iOS / old Android co-signers continue to work with the new SDK (already write `priorityFee` as price in µLam/CU). New SDK + old iOS pairings now match on the wire — fixes a cross-platform pairing that was always broken.

## Test plan

- [x] Unit tests pass (`yarn test:core`)
- [x] Typecheck clean (`yarn tsc -p .config/tsconfig.json --noEmit`)
- [x] Lint clean (`yarn lint`)
- [x] Tier 1 sideload smoke test
- [x] Tier 2 user-bug repro

## Summary by CodeRabbit

* **Improvements**
  * Priority fee selection now uses the 75th-percentile with a sparse-sample guard and enforced minimum floor.
  * Fee composition refined to include compute-unit scaling, base fee, and ATA rent when creating associated token accounts.

* **New Features**
  * Optional scoping of fee sampling by writable accounts.
  * Exposed ATA rent constant for fee calculations.

* **Refactor**
  * Removed legacy Solana-specific refine flow and its platform override.

* **Tests**
  * Added comprehensive tests for priority fee selection and Solana fee/send scenarios.


packages/core/chain/chains/solana/getDynamicPriorityFeePrice.ts
```diff
@@ -16,11 +33,10 @@ export const getDynamicPriorityFeePrice = async (): Promise<number> => {
...
     }
 
-    const mid = Math.floor(nonZeroFees.length / 2)
-
-    if (nonZeroFees.length % 2 === 0) {
-        return Math.round((nonZeroFees[mid - 1] + nonZeroFees[mid]) / 2)
-    }
+    const index = Math.min(
+        Math.floor(nonZeroFees.length * PRIORITY_FEE_PERCENTILE),
+        nonZeroFees.length - 1
```
premiumjibles, line 55: The function name + docstring + new `PRIORITY_FEE_PERCENTILE = 0.75` constant claim P75 semantics, but `nonZeroFees` is never sorted before the index lookup. `recentFees` from Solana's `getRecentPrioritizationFees` is slot-ordered (chronological), not fee-sorted, so `nonZeroFees[Math.floor(length * 0.75)]` selects an arbitrary slot's fee at the 75% array position rather than the 75th percentile of the fee distribution.

The previous "median" code had the same defect (mid-index of unsorted), but this PR newly codifies the percentile claim with intent — a future maintainer reading this will trust the docstring. To compute an actual P75:

```ts
const sorted = [...nonZeroFees].sort((a, b) => a - b)
const index = Math.min(Math.floor(sorted.length * PRIORITY_FEE_PERCENTILE), sorted.length - 1)
return Math.max(sorted[index], solanaConfig.priorityFeePrice)
```

Not BLOCKER because the trailing `Math.max(_, solanaConfig.priorityFeePrice)` floors the low direction, so we don't underpay below the cross-platform minimum. The high direction is unbounded — picking a slot from the upper portion of the slot-ordered array can land on a high-contention slot (e.g., the LP-add slot we're targeting via `lockedWritableAccounts`) and overpay significantly relative to actual market P75.

packages/core/mpc/keysign/signingInputs/resolvers/solana/send.ts
```diff
@@ -23,11 +24,19 @@ export const getSolanaSendSigningInput = ({
...
     computeLimit,
+    priorityFee,
   } = getBlockchainSpecificValue(
     keysignPayload.blockchainSpecific,
     'solanaSpecific'
   )
 
+  // Floor at the config minimum so co-signers all encode the same
+  // `setComputeUnitPrice` instruction when the wire value is missing.
+  const priorityFeePrice = maxBigInt(
```
premiumjibles, line 35: The signing-time floor `maxBigInt(priorityFee, solanaConfig.priorityFeePrice)` is new in this PR, and its safety claim depends on a cross-platform invariant that the PR body asserts but doesn't verify against old iOS / old Android signing-time logic.

Concrete failure mode: when an old-iOS / old-Android device is the ceremony INITIATOR, it writes `chainSpecific.priorityFee` from its own native priority-fee path — not this SDK's `getDynamicPriorityFeePrice`, which now floors at the function level via `Math.max(_, priorityFeePrice)`. If the old-platform initiator computes a P75 below `solanaConfig.priorityFeePrice` (1_000_000 µLam/CU) and writes that raw to the wire (no floor), and a new-SDK device is a co-signer, the co-signer's `setComputeUnitPrice` instruction encodes `max(wire, floor)` = 1_000_000 while the old-platform's encodes `wire` directly. Different message bytes → MPC ceremony fails on signature verify.

The PR body says: *"Old iOS / old Android co-signers continue to work with the new SDK (already write priorityFee as price in µLam/CU). New SDK + old iOS pairings now match on the wire — fixes a cross-platform pairing that was always broken."* The "match on the wire" claim holds only if at least one of:

1. Old iOS / old Android also apply a signing-time floor in their native paths (then both encode `max(wire, floor)` and match).
2. Empirical fee market keeps P75 ≥ 1_000_000 µLam/CU consistently across sample windows (then `wire ≥ floor` is invariant and `max(wire, floor) = wire`).

The Tier 2 repro is "Android initiator + Extension co-signer" — both new-SDK devices in the post-fix world, which doesn't exercise the old-initiator / new-cosigner path.

Safer alternative: apply the floor at the resolver level so the wire value always reflects the floor regardless of co-signer version. The resolver already does this via `Math.max(...)` in `getDynamicPriorityFeePrice` — if the same is applied to whatever path the old platforms use to populate `chainSpecific.priorityFee`, the signing-time `maxBigInt` becomes redundant defense-in-depth rather than load-bearing. Or: confirm (1) or (2) and the current shape is fine.

packages/core/chain/chains/solana/solanaConfig.ts
```diff
@@ -1,8 +1,15 @@
...
+  // Solana charges 5000 lamports per signature. Vultisig sends are
+  // single-signer, so this is the per-tx base fee. If a future path
+  // emits multi-signer messages this becomes 5000 * N.
+  baseFee: 5000,
+
+  // Rent-exempt minimum for a 165-byte SPL token associated account, used
+  // when emitting `createAssociatedTokenAccount` alongside an SPL transfer.
+  // Constant on mainnet/devnet/testnet — rent rate is hardcoded in the
+  // Solana runtime.
+  ataRentLamports: 2_039_280,
```
premiumjibles, line 22: `2_039_280` is correct today: SPL Token Account = 128-byte account header + 165-byte data → `(128 + 165) × 3480 lamports/byte/year × 2-year threshold = 2_039_280`. Optional: include the formula in the comment so a future maintainer can re-derive and verify if Solana ever bumps `RENT_PER_BYTE_PER_YEAR`.

```
// Rent-exempt minimum: (128-byte account header + 165-byte SPL Token data)
// × RENT_PER_BYTE_PER_YEAR (3480 lamports) × 2-year threshold = 2_039_280.
// Constant on mainnet/devnet/testnet today; if Solana ever bumps the
// per-byte rate or threshold, re-derive from the formula.
ataRentLamports: 2_039_280,
```

Not gating ship — the rent rate hasn't moved and the deleted `refine.ts` was previously fetching this via `getMinimumBalanceForRentExemption` RPC anyway. Hardcoding it is a fine simplification; the comment is the only thing that rots if Solana ever changes.

packages/core/chain/chains/solana/getDynamicPriorityFeePrice.ts
```diff
@@ -16,11 +33,10 @@ export const getDynamicPriorityFeePrice = async (): Promise<number> => {
         return solanaConfig.priorityFeePrice
     }
 
-    const mid = Math.floor(nonZeroFees.length / 2)
-
-    if (nonZeroFees.length % 2 === 0) {
-        return Math.round((nonZeroFees[mid - 1] + nonZeroFees[mid]) / 2)
-    }
+    const index = Math.min(
+        Math.floor(nonZeroFees.length * PRIORITY_FEE_PERCENTILE),
```
premiumjibles, line 54: Sample-count edge: when `nonZeroFees` is short (1-3 entries), the percentile selection picks the single highest entry. With the new `lockedWritableAccounts` scoping, sparse windows are more likely than they were with the global feed — querying a quiet recipient ATA might surface only 2-3 non-zero slots in the recent window.

Worst-case shape: a single LP-add slot that paid 50_000_000 µLam/CU sets the priority-fee for the next caller until enough idle slots dilute it. The trailing `Math.max(_, priorityFeePrice)` floor only protects against underpayment, not overpayment from a sparse-window spike.

Consider a min-sample threshold:

```ts
if (nonZeroFees.length < MIN_SAMPLE_SIZE) {
  return solanaConfig.priorityFeePrice
}
```

with `MIN_SAMPLE_SIZE = 5` or so, so we don't extrapolate from too-few data points and fall back to the floor-as-base-rate when the window is too thin.

Question rather than ask because the cost direction here is "user pays more than necessary" rather than "tx fails to land" — UX issue, not safety. But the new vault-aware scoping does make sparse windows materially more likely than the pre-PR code dealt with, so worth a thought.

premiumjibles (approved):
*Heads up — AI-assisted review on a signing-adjacent PR. Worth a double-read.*

Reviewed the three coupled fixes (SPL memo plumbing, vault-aware P75 oracle, cross-platform `priorityFee` semantic alignment). Fee-totals math centralised in `getSolanaFeeAmount` is the right move — collapsing the iOS/Android-vs-extension semantic split removes a real source of MPC failure that the Tier 2 repro tx confirms. Two inline questions on the floor logic, one finding on the percentile-on-unsorted-array, one nit. Single-agent this pulse — Gemini second-pass returned an empty response on the diff; CLI is live (verified via smoke test) but the dual-agent prompt path didn't yield candidates today, so this is Claude only.

### What's right

- Collapsing `refine.ts` + the RN override file + the rollup mapping into one stateless `getSolanaFeeAmount` is the correct architecture — `chainSpecific.priorityFee` now means the same thing on every platform (µLam/CU price), and the totals computation lives in fee-display code where it belongs.
- The native vs SPL writable-account split in `chainSpecific/solana/index.ts` is precise — SystemProgram never touches the recipient's main wallet on SPL transfers, so scoping the priority-fee query to ATAs (not the recipient pubkey) for the `coin.id`-set path matches what the chain actually writes.
- Test coverage explicitly pins the floor case (`floors priorityFee at solanaConfig.priorityFeePrice when below`) and the empty-string fallback — that locks the new wire-time semantic into CI.
- Tier 2 receipt is the right kind of evidence: real Android+Extension MPC pairing, real THORChain inbound vault, on-chain settlement at `5mmFNp7BgC…` plus THORChain status `done` at height 25,984,536. That's an end-to-end pin on the exact failure mode this PR fixes.
- `attempt(getSplAssociatedAccount(receiver))` correctly fail-soft when receiver ATA doesn't exist (create-and-transfer path) — the `if (data)` guard skips pushing a non-existent address rather than throwing.

### Findings

- **[MAJOR]** (high) `packages/core/chain/chains/solana/getDynamicPriorityFeePrice.ts:38` — percentile selection on slot-ordered unsorted fees array
- **[QUESTION]** (high) `packages/core/mpc/keysign/signingInputs/resolvers/solana/send.ts:35` — signing-time floor + cross-platform compat with old-platform initiator
- **[QUESTION]** (med) `packages/core/chain/chains/solana/getDynamicPriorityFeePrice.ts:37` — sample-count guard for sparse fee windows
- **[NIT]** (med) `packages/core/chain/chains/solana/solanaConfig.ts:14` — `ataRentLamports` formula derivation in comment

### QA Report

| Check | Result |
|---|---|
| `yarn test:core` | not_run (review from diff) — author claims 350/350 passing |
| `yarn tsc -p .config/tsconfig.json --noEmit` | not_run (review from diff) — author claims clean |
| `yarn lint` | not_run (review from diff) — author claims clean |
| Tier 1 sideload smoke | author-reported pass — receipt `57DUwmYRMd…` |
| Tier 2 cross-platform repro (Android initiator + Extension co-signer) | author-reported pass — receipt `5mmFNp7BgC…`, THORChain status `done` at height 25,984,536, LP add registered |

## Verdict

**APPROVE — please address inline comments.**

The signing path correctness rests on the cross-platform floor question being answered (inline 2). If old iOS/Android already floor at signing time in their native paths, this is fully safe and the `maxBigInt` is defense-in-depth. If they don't, there's a latent MPC mismatch case for old-platform-initiator + new-SDK-cosigner ceremonies on uncongested networks. Worth confirming before merge — either resolve the question or move the floor to the resolver so wire-value reflects floor regardless of co-signer version.

The percentile-on-unsorted finding (inline 1) is a real correctness bug carried forward from the previous "median" code; the floor at the bottom prevents underpayment, but the unsorted-array indexing means the "P75" claim isn't what the docstring says. Sort first, then index.

---
🤖 prs@Claude-4.7 — see ~/.claude/skills/prs if this got something wrong.

rcoderdev (approved):

gomesalexandre:
## Hi paaao - merged main into the branch (`311c8d12`) so CI re-runs against fresh state

Also wanted to surface a precise tackle plan for the 3 actionable @premiumjibles findings since codex review confirmed all of them are real. I attempted to push the fixes directly per the team's drive-to-merge convention but my codex impl session hung at the 35-min mark; pasting the exact diffs here so you can land them quickly:

### 1. (CORRECTNESS BUG) `getDynamicPriorityFeePrice.ts:38` — P75 docstring lies, array isn't sorted

`recentFees` from Solana's `getRecentPrioritizationFees` is slot-ordered (chronological), not fee-sorted. `nonZeroFees[Math.floor(length * 0.75)]` selects an arbitrary slot's fee at the 75% array position rather than the 75th percentile of the fee distribution. The previous "median" code had the same defect but the new docstring codifies the intent so it must hold.

```ts
const sorted = [...nonZeroFees].sort((a, b) => a - b)
const index = Math.min(Math.floor(sorted.length * PRIORITY_FEE_PERCENTILE), sorted.length - 1)
return Math.max(sorted[index], solanaConfig.priorityFeePrice)
```

### 2. (UX safety, related to #1) Same file - sparse-window overpayment

When `nonZeroFees` has 1-3 entries, the percentile selection picks the single highest entry. With the new lockedWritableAccounts scoping, sparse windows are more likely. Worst case: a single LP-add slot at 50_000_000 µLam/CU sets the priority-fee for the next caller until idle slots dilute. Math.max floor only protects underpayment.

```ts
const MIN_SAMPLE_SIZE = 5
if (nonZeroFees.length < MIN_SAMPLE_SIZE) {
  return solanaConfig.priorityFeePrice
}
// then sort + pick index as above
```

Place BEFORE the sort to avoid wasted work on too-small samples.

### 3. (DOCS NIT) `solanaConfig.ts:14` — document the rent-exempt formula

```ts
// Rent-exempt minimum: (128-byte account header + 165-byte SPL Token data)
// × RENT_PER_BYTE_PER_YEAR (3480 lamports) × 2-year threshold = 2_039_280.
// Constant on mainnet/devnet/testnet today; if Solana ever bumps the
// per-byte rate or threshold, re-derive from the formula.
ataRentLamports: 2_039_280,
```

### Suggested unit test (vitest) for `getDynamicPriorityFeePrice`

```ts
it('selects the actual P75 of the fee distribution after sort', () => {
  // unsorted: [10, 50, 30, 90, 20] - sorted: [10, 20, 30, 50, 90]
  // index floor(5 * 0.75) = 3 -> 50
  expect(p75([10, 50, 30, 90, 20])).toBe(50)
})
it('returns floor when sample below MIN_SAMPLE_SIZE', () => {
  expect(p75([100])).toBe(solanaConfig.priorityFeePrice)
})
it('returns floor on empty input', () => {
  expect(p75([])).toBe(solanaConfig.priorityFeePrice)
})
```

### Out of scope for this comment

The `solana/send.ts:35` cross-platform compatibility concern needs deeper context with old-iOS/old-Android signing-time logic that I don't have - leaving that one to you.

LMK if you want me to push these directly, happy to.

gomesalexandre:
## Correction on the previous tackle plan

Walked the actual source via `git blame` and the `.sort((a, b) => a - b)` is **already present** at `getDynamicPriorityFeePrice.ts:30` (added by Ehsan-saradar in 16068651, the original median-oracle PR). The function is correctly computing actual P75 of the sorted distribution today.

The earlier inline concern about the missing sort either references an earlier diff state OR misread the multi-line method chain (the sort is on line 30 of the chain, the `Math.floor(length * 0.75)` on line 37 — visually separated). Closing that finding as not-applicable.

**Still-actionable items:**

1. **MIN_SAMPLE_SIZE threshold (real)** — when nonZeroFees has 1-3 entries (more likely now with lockedWritableAccounts scoping), the percentile selection picks the single highest entry. Worst case: a single LP-add slot at 50_000_000 µLam/CU sets fee for the next caller. Math.max floor only protects underpayment, not overpayment from sparse windows.

```ts
const MIN_SAMPLE_SIZE = 5
if (nonZeroFees.length < MIN_SAMPLE_SIZE) {
  return solanaConfig.priorityFeePrice
}
// existing index/percentile/Math.max chain unchanged
```

2. **ataRentLamports formula docs** at `solanaConfig.ts:14` — comment with derivation so future maintainers can re-verify if Solana ever bumps RENT_PER_BYTE_PER_YEAR.

Will push these 2 fixes locally + run tests + push to the branch with receipts shortly. The cross-platform send.ts:35 concern remains for the author to weigh in on.

Apologies for the noise on the sort item — should've blamed before commenting.

gomesalexandre:
## Pushed in `e815ef6f` — sparse-window guard + ATA rent formula docs

Per the correction comment above, the missing-sort finding was outdated. Pushed the 2 still-actionable items:

### 1. MIN_SAMPLE_SIZE = 5 sparse-window guard

`getDynamicPriorityFeePrice` now bails to floor when `nonZeroFees.length < 5`. With the new `lockedWritableAccounts` scoping in this PR, sparse windows are materially more likely - a single LP-add slot at e.g. 50,000,000 µLam/CU would dominate the percentile and set a wildly inflated fee for the next caller. `Math.max(percentile, floor)` only protects against UNDER-payment; this guard protects against OVER-payment from a sparse-window spike.

### 2. ATA rent formula derivation comment

`ataRentLamports = 2_039_280` now carries the explicit formula:
```
(128-byte account header + 165-byte SPL Token data)
  × RENT_PER_BYTE_PER_YEAR (3480 lamports)
  × 2-year threshold
= 293 × 3480 × 2
= 2,039,280 lamports
```

### 3. New test file: `getDynamicPriorityFeePrice.test.ts` (8 tests)

Coverage:
- selects actual P75 of SORTED distribution (not slot-ordered)
- returns floor when sample below MIN_SAMPLE_SIZE
- returns floor when computed P75 falls below floor
- returns floor on empty input
- returns floor when ALL fees are zero (filtered out -> sparse)
- forwards `lockedWritableAccounts` when provided
- omits the param when `writableAccounts` is empty
- pin `_MIN_SAMPLE_SIZE_FOR_TEST` ∈ [3, 20] — traps a future drop of the guard

### What I deliberately did NOT touch

- The cross-platform compatibility concern at `solana/send.ts:35`. That needs deeper context with old iOS/Android signing-time logic and is your call.

### Receipts

```
$ yarn test:core
Test Files  38 passed (38)
     Tests  367 passed (367)  (was 359, +8 new)

$ yarn typecheck
exit 0
```

If you'd rather revert any of this, no problem — push back and I'll drop the relevant commit.

realpaaao:
@premiumjibles answering the open cross-platform-floor question (Finding 2 in your review at `solana/send.ts:35`).

**TL;DR: the floor at signing is pure defense-in-depth. No latent MPC mismatch case exists for any current iOS/Android initiator paired with new SDK cosigner.**

Both iOS and Android initiator paths already write `chainSpecific.priorityFee >= 1_000_000` (the same value the new SDK floors to). Verified by inspection:

**iOS** — `vultisig-ios/VultisigApp/VultisigApp/Core/Services/BlockChainService.swift:381`:

```swift
let defaultFee = BigInt(SolanaHelper.defaultPriorityFeePrice)  // 1_000_000
let dynamicPriorityFee: BigInt
do {
    let fee = try await sol.fetchRecentPrioritizationFees()
    dynamicPriorityFee = max(BigInt(fee), defaultFee)   // <-- floor at write time
} catch {
    dynamicPriorityFee = defaultFee
}
```

**Android** — `vultisig-android/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt:getMedianPriorityFee`:

```kotlin
val median = if (nonZeroFees.isEmpty()) fallback else nonZeroFees[nonZeroFees.size / 2]
maxOf(median, fallback).coerceAtMost(MAX_PRIORITY_FEE_PRICE.toBigInteger())
// fallback = SOLANA_PRIORITY_FEE_PRICE = 1_000_000
```

So:

| Initiator | wire `priorityFee` | new SDK cosigner encodes (after `maxBigInt(_, 1M)`) | match? |
|---|---|---|---|
| current iOS | `>= 1_000_000` | same | ✅ |
| current Android | `>= 1_000_000` | same | ✅ |
| current SDK (after this PR) | `>= 1_000_000` (this PR floors at the API layer too) | same | ✅ |
| **OLD** extension before this PR | `~105_000` (totals-as-price bug) | floor → `1_000_000` | ✅ — old ext at signing also hardcoded `1_000_000`, so encodes match |

The `maxBigInt` floor in `send.ts:35` therefore never fires on a wire value coming from any current production peer; it's a guard against future regressions and against the historical old-extension-with-totals-bug path. Move-floor-to-resolver wouldn't change the encoded `setComputeUnitPrice` arg for any pairing — would just dedupe the floor across two layers. Happy to follow up with that consolidation if you'd prefer single-source-of-truth, but it's cosmetic.

Receipts the analysis matches: Tier 2 repro tx [`5mmFNp7BgC...`](https://solscan.io/tx/5mmFNp7BgCf4WV2YEQoSFSmggytYfeu55uvgeyLj7AUXFJWtJzvUvcZCFLDp4DFAGrV1irGS1DWuniNp7SJBtudP) was Android-initiator + new-SDK-cosigner with the new floor in place; MPC produced a matching sig and THORChain registered the LP add (`done` at height 25,984,536). The pairing the open question worried about is exactly the one verified on chain.

Also pushed `ce46f32f` to address @coderabbitai's MIN_SAMPLE_SIZE pin nitpick (toBe(5) + typo fix).

gomesalexandre:
## TL;DR follow-up ack — that closes the loop on the cross-platform concern

paaao's analysis is convincing. The wire-side floor on iOS (`BlockChainService.swift:381`) and Android (`SolanaApi.kt:getMedianPriorityFee`) both clamp to `1_000_000` before writing `chainSpecific.priorityFee`, so the new SDK cosigner's `maxBigInt(_, 1_000_000)` floor never sees a sub-floor wire value from any current production peer. The MPC encoding match is the load-bearing invariant + it holds.

Nice receipt with the on-chain Tier 2 repro (`5mmFNp7B...JBtudP` Android-initiator + new-SDK-cosigner). That's the exact pairing the open question was probing.

The "consolidate floor into a single resolver" follow-up is cosmetic per paaao — agree with that framing. Two layers of the same floor is belt-and-braces, not a bug.

From my side: this was the only outstanding concern on the PR. CI just needs Unit Tests + Lint to land and you're at green-and-mergeable.

PR #372, opened 2026-05-02
fix(cosmos): handle wasm execute in ibcEnabled signing input + fail closed on inconsistent GENERIC_CONTRACT

Ehsan-saradar:
## Summary

- Mirror the `vaultBased` branch's `wasmExecuteContractGeneric` handling into the `ibcEnabled` branch of `getCosmosSigningInputs`, so contract executes on Terra, Cosmos, Osmosis, Kujira, Dydx, Noble, and Akash route through TW's wasm path instead of falling through to the bank `MsgSend` default.
- Fail closed when `transactionType === GENERIC_CONTRACT` arrives without a meaningful `wasmExecuteContractPayload` — refuse rather than dispatch a different message type.

## Why

The `ibcEnabled` branch had no `GENERIC_CONTRACT` handling, so a wasm-intent payload would fall through to `sendCoinsMessage` with `toAddress = contract`. The two cases (caller intent vs. encoded message) could diverge silently.

## Companion PR

This unblocks vultisig/vultisig-windows#3828 — `Station.post()` produces `transactionType=GENERIC_CONTRACT` with a `wasmExecuteContractPayload` for Terra, which the windows extension uses for Astroport / Skip / etc. swaps.

## Test plan

- [x] `yarn build:shared` (tsc clean)
- [x] Manual end-to-end: Astroport LUNA→ASTRO swap on Terra, verified on-chain that `messages[0]['@type'] === '/cosmwasm.wasm.v1.MsgExecuteContract'` and the swap chain executed (25 wasm events). Tx [AE6450DC...](https://chainsco.pe/terra2/tx/AE6450DC309B07373C5F08B2F92D71295ACD478A013BEB0EF5F744227ABB1841)
- [ ] CI / unit tests
- [ ] Reviewer check: no other ibcEnabled call sites set `transactionType=GENERIC_CONTRACT` today, so the new fail-closed branch should not regress existing flows

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of generic contract transactions so IBC-enabled contracts with wasm execute payloads are executed instead of treated as coin transfers.
  * Enforced stricter validation to error when required wasm execute data is missing or incomplete, preventing unintended fallbacks.

* **Behavioral Changes**
  * Vault-based fallback now fails closed for inconsistent GENERIC_CONTRACT payloads rather than silently falling back.


gomesalexandre (commented):
## Review (read-only, fund-safety pass)

### Summary
Mirroring the vaultBased branch's wasm contract handling into the ibcEnabled branch is the right move — eliminates the silent-divergence class where `transactionType=GENERIC_CONTRACT` would fall through to `MsgSend` on Terra/Osmosis/etc. The fail-closed branch is the correct safety posture.

### What I checked
- Pulled `refs/pull/372/head` locally
- Read the diff
- Cross-checked the new ibcEnabled branch (lines 122-161) against the existing vaultBased branch (lines 236-258) — symmetry verified
- Confirmed `hasMeaningfulWasm` predicate is identical between branches (sender + contract + executeMsg all non-empty trim)

### Asymmetry observation (non-blocking)
The vaultBased branch does NOT fail-closed on `GENERIC_CONTRACT` with empty payload — it falls through to memo-based handling (`merge:`/`unmerge:`/`open:`). That's defensible because those memo paths use `coin.address` + `toAddress` directly and don't need contractPayload. But it's worth a comment in the vaultBased branch that the missing fail-closed is intentional, so a future reader doesn't "fix" it by mirror-applying the new ibcEnabled guard and breaking THORChain merge/unmerge flows.

### Re: approve stamp
Will defer the approve stamp to whoever runs the independent codex pass per protocol. Code looks tight; small surface; clear safety posture; the kind of PR that should land cleanly once codex confirms.

gomesalexandre (commented):
## Review (read-only, fund-safety pass)

### Summary
Mirroring the vaultBased branch's wasm contract handling into the ibcEnabled branch is the right move — eliminates the silent-divergence class where `transactionType=GENERIC_CONTRACT` would fall through to `MsgSend` on Terra/Osmosis/etc. The fail-closed branch is the correct safety posture.

### What I checked
- Pulled `refs/pull/372/head` locally
- Read the diff
- Cross-checked the new ibcEnabled branch (lines 122-161) against the existing vaultBased branch (lines 236-258) — symmetry verified
- Confirmed `hasMeaningfulWasm` predicate is identical between branches (sender + contract + executeMsg all non-empty trim)

### Asymmetry observation (non-blocking)
The vaultBased branch does NOT fail-closed on `GENERIC_CONTRACT` with empty payload — it falls through to memo-based handling (`merge:` / `unmerge:` / `open:`). That's defensible because those memo paths use `coin.address` + `toAddress` directly and don't need contractPayload. But it's worth a comment in the vaultBased branch that the missing fail-closed is intentional, so a future reader doesn't "fix" it by mirror-applying the new ibcEnabled guard and breaking THORChain merge/unmerge flows.

### Re: approve stamp
Will defer the approve stamp to whoever runs the independent codex pass per protocol. Code looks tight; small surface; clear safety posture; the kind of PR that should land cleanly once codex confirms.

gomesalexandre:
## Pre-merge codex adversarial pass — pushed matching vaultBased fail-closed

Pushed ec68d247 to mirror the fail-closed guard into the vaultBased branch (codex pre-merge pass High finding).

### Why

Codex flagged that the fail-open hazard this PR fixes for `ibcEnabled` (GENERIC_CONTRACT tx with empty wasmExecuteContractPayload silently falls through to MsgSend) ALSO exists symmetrically in the `vaultBased` (THORChain) branch. There the fallback chain is `merge:` / `unmerge:` / `THORChainDeposit` / `THORChainSend` depending on the memo + isDeposit/swapPayload shape — a malformed contract tx could end up signing a THORChain send to the contract address with no contract execution. Same class of silent message-type substitution as the ibcEnabled bug.

### What

- Mirror the fail-closed throw immediately after the `hasMeaningfulWasm` check in vaultBased
- Same error wording shape as ibcEnabled, with the vaultBased-specific fallback list spelled out
- Converted the chained `else if (memo merge:)` / `else if (memo unmerge:)` to standalone `if` blocks since the new throw above breaks the chain (each branch returns, so flow control is identical)

### Codex other findings

**Medium (deferred)**: fail-closed is one-way. A meaningful `wasmExecuteContractPayload` paired with `transactionType !== GENERIC_CONTRACT` is still silently ignored and can dispatch MsgSend (ibcEnabled) or THORChain paths (vaultBased). Codex didn't find a repo-local builder emitting this shape, but the resolver still trusts caller consistency. Worth a follow-up issue — current PR scope is the GENERIC_CONTRACT direction, which is the live-bug class.

**Low (deferred)**: zero test coverage on this resolver path. Codex listed the minimal pin set:
- ibcEnabled + GENERIC_CONTRACT + meaningful wasm builds wasmExecuteContractGeneric
- ibcEnabled + GENERIC_CONTRACT + missing/blank/whitespace wasm fields throws
- vaultBased/THORChain + GENERIC_CONTRACT + meaningful wasm builds same wasm message
- vaultBased/THORChain + GENERIC_CONTRACT + missing/blank/whitespace wasm fields throws (NEW path from this commit)
- vaultBased + memo `merge:` with empty attached wasm payload but non-GENERIC_CONTRACT tx type still preserves the legacy merge flow

The cosmos resolver test harness needs walletCore wasm init + protobuf schema fixtures (heavier than a typical unit), so leaving as a follow-up issue. Worth doing because this whole file is fund-safety load-bearing.

### Build / test

- `yarn workspace @vultisig/core-mpc build`: clean
- `yarn workspace @vultisig/sdk test --run`: 1357/1357 green
- GPG-signed commit (key 9EBA3CBA654A2DF5)

Approving + driving to merge once CI is green.

gomesalexandre (approved):
Approved with codex round-2 fix pushed inline (ec68d247). Original PR was solid for ibcEnabled, the symmetric vaultBased fail-closed was the missing half. See companion PR comment for codex receipt + deferred Medium / Low (test-harness gap on this resolver — worth a follow-up issue, the file is fund-safety load-bearing). Driving to merge on CI green.

Issue #385, opened 2026-05-04
sdk-cli: drop legacy actions-envelope consumption, migrate fully to clientside-tool path


neavra:
## Summary

`@vultisig/cli` consumes the agent-backend SSE stream through two parallel channels for client-side execution:

1. **`tool-input-available` events with `clientExecuted: true`** — fired into `onClientSideToolCall`, dispatched through `CLIENT_SIDE_TOOL_DISPATCH` (`session.ts:38`). This is the path that `vultiagent-app` uses exclusively.
2. **`actions` events** — fired into `onActions` (`session.ts:324`, currently a documented no-op), but the events still land in `streamResult.actions`, which then runs through the `legacyActions` filter at `session.ts:399-432` and into `executor.dispatch()`.

agent-backend has already migrated off the emitter side: `respond_to_user` is no longer registered as a tool (`tools.go:555-573`), the two `respond_to_user` branches in `agent.go` are documented `// DEPRECATED` (lines 1136-1154, 2075-2093), and `processActions()` runs on an empty slice every turn. The `actions` SSE channel exists but never fires on real conversations.

The CLI is therefore carrying a complete second client-side execution system (legacy `Action` / `ActionResult` types, `executeAction`, `dispatch()` switch, `onActions` callback, the `data-actions` → `actions` SSE remap) that the backend no longer drives. This blocks the agent-backend cleanup, since deleting the emitter would break deployed CLI binaries that still register the consumer.

## Goal

Make sdk-cli consume **only** the `tool-input-available` channel. Delete:

- `onActions` callback type + body
- `actions` SSE arm in `client.ts:344-348` and the `data-actions` → `actions` remap at `client.ts:409-410`
- `streamResult.actions` field
- `legacyActions` post-processing in `session.ts:399-432`
- `executor.dispatch()` switch + `executeAction()` + `executeActions()` methods (per-tool implementation methods stay, called directly from `dispatchClientSideTool`)
- `Action`, `ActionResult`, `SSEActions` types
- `actionResultToRecentAction` adapter
- Top-level `action_result` field on outbound `SendMessageRequest`

## Out of scope

- `tx_ready` channel and `sign_tx` synth flow — these don't ride the actions envelope and stay
- The per-tool handler implementations (`addCoin`, `removeCoin`, etc.) — they get a signature change but stay

## Acceptance

- [ ] `yarn test:cli` green
- [ ] `yarn workspace @vultisig/cli typecheck` green
[...]

PR #301, opened 2026-04-21
chore: version packages

rcoderdev:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

-   [#280](https://github.com/vultisig/vultisig-sdk/pull/280) [`5aef564`](https://github.com/vultisig/vultisig-sdk/commit/5aef564309aeeede5da250e03447e0a3da0a12ab) Thanks [@rcoderdev](https://github.com/rcoderdev)! - Add THORChain LP agent actions (`thorchain_pool_info`, `thorchain_add_liquidity`, `thorchain_remove_liquidity`) to the CLI executor and document them in AGENTS.md. Fix `@vultisig/lib-utils` ESM imports to directory entrypoints so Node resolves `dist` correctly.

### Patch Changes

-   Updated dependencies \[[`219cb00`](https://github.com/vultisig/vultisig-sdk/commit/219cb00898deeaac418945a89c1d243f25aae152)]:
    -   @vultisig/sdk@0.17.0
    -   @vultisig/core-chain@1.3.1
    -   @vultisig/client-shared@0.2.1
    -   @vultisig/rujira@12.0.0

### Minor Changes

-   [#284](https://github.com/vultisig/vultisig-sdk/pull/284) [`219cb00`](https://github.com/vultisig/vultisig-sdk/commit/219cb00898deeaac418945a89c1d243f25aae152) Thanks [@premiumjibles](https://github.com/premiumjibles)! - feat(sdk): vault-free prep surface + LLM/agent utilities + token-resolution primitives
    -   Vault-free `prepare*FromKeys` helpers that build unsigned `KeysignPayload`s from a `VaultIdentity` (raw public keys + identity metadata, no key shares): `prepareSendTxFromKeys`, `prepareSwapTxFromKeys`, `prepareContractCallTxFromKeys`, `prepareSignAminoTxFromKeys`, `prepareSignDirectTxFromKeys`, `getMaxSendAmountFromKeys`. Atomic chain helpers `getCoinBalance` and `getPublicKey` are also re-exported. `VaultBase`, `TransactionBuilder`, and `SwapService` delegate to these internally.
[...]

PR #403, opened 2026-05-05
sdk-cli: drop legacy actions SSE channel consumption (Phase 1 of #385)

neavra:
Phase 1 of #385. Drops consumption of the legacy `actions` SSE channel
without touching the `Action` / `ActionResult` types or
`executor.dispatch()` switch — those are still used by Path A
(`dispatchClientSideTool`) and Path C (`tx_ready` → synth `sign_tx`),
and a Phase 2 follow-up will refactor those callers off the wrapper so
the types can come out cleanly.

agent-backend has already migrated off the emitter side: `respond_to_user`
is no longer registered as a tool, the two `respond_to_user` branches in
`agent.go` are documented `// DEPRECATED`, and `processActions()` runs
on an empty slice every turn. The `actions` SSE channel exists but never
fires on real conversations — this PR just stops listening for it.

## Changes

- `client.ts`: drop `onActions` callback typedef from both `sendMessageStream` and `handleSSEEvent` signatures, drop the `case 'actions':` SSE switch arm, drop the `data-actions` → `actions` mapping in `mapV1EventType`, drop `actions: Action[]` field from `SSEStreamResult`.
- `session.ts`: drop the no-op `onActions` callback body, drop the entire `legacyActions` filter + auto-sign-from-build-action recursion block (was lines 399-432). The existing `tx_ready` → synth `sign_tx` path covers the same transaction-signing flow.

## Out of scope (Phase 2)

- `Action`, `ActionResult`, `SSEActions` types
- `executor.dispatch()` switch + `executeAction()` + `executeActions()`
- `actionResultToRecentAction()` adapter
- Top-level `action_result` field on outbound `SendMessageRequest`

These stay because `dispatchClientSideTool` and the `tx_ready` synth
flow currently route through `executeAction(Action)`. Refactoring those
callers to invoke per-tool methods directly is a follow-up.

## Verification

- ✅ `yarn workspace @vultisig/cli typecheck`
- ✅ `yarn test:cli` — all 131 tests green
- ✅ Grep for `onActions`, `streamResult.actions`, `case 'actions'`, `data-actions`, `legacyActions` returns zero hits in `clients/cli/src/`

## Test plan

- [ ] Manual verification against staging agent-backend: confirm `add_coin`, `remove_coin`, `add_chain`, `remove_chain`, `address_book_*`, `sign_typed_data`, `polymarket_sign_bet`, multi-leg tx sequences all still work (these all flow through `tool-input-available` SSE events, not the deleted channel).
- [ ] Confirm zero `onActions` callback firings in a normal session log

## Release Notes

* **Refactor**
  * Simplified the agent streaming API by removing support for `onActions` callbacks during streaming operations.
  * Removed `actions` data from streaming results, streamlining the structure of returned data.
  * Updated message processing to eliminate legacy action handling paths and improve workflow for pending operations.

### DIFF
--- a/clients/cli/src/agent/client.ts
+++ b/clients/cli/src/agent/client.ts
@@ -6,7 +6,6 @@
  */
 import { AgentErrorCode, inferAgentErrorCodeFromMessage, isAgentErrorCode } from './agentErrors'
 import type {
-  Action,
   AuthTokenRequest,
   AuthTokenResponse,
   ConversationMessage,
@@ -147,7 +146,6 @@ export class AgentClient {
       // unhandled rejections.
       onClientSideToolCall?: (toolCallId: string, toolName: string, input: Record<string, unknown>) => void
       onTitle?: (title: string) => void
-      onActions?: (actions: Action[]) => void
       onSuggestions?: (suggestions: Suggestion[]) => void
       onTxReady?: (tx: TxReadyPayload) => void
       onMessage?: (msg: ConversationMessage) => void
@@ -177,7 +175,6 @@ export class AgentClient {
 
     const result: SSEStreamResult = {
       fullText: '',
-      actions: [],
       suggestions: [],
       transactions: [],
       message: null,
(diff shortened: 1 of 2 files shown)


gomesalexandre (approved):
Clean removal of the legacy `data-actions` SSE channel from the CLI client. Aligns with agent-backend #301 (Phase 1 of dead actions registry deletion).

Diff is pure deletion - `onActions`, `actions` in `SSEStreamResult`, the `data-actions` case block, the `legacyActions` fallback, and `data-actions` → `actions` event-type mapping are all gone. No new behavior introduced.

CI all green (unit, integration, lint). Approving.

PR #247, opened 2026-04-13
fix(cli): remove calldata_id resolution — tx_ready delivers calldata directly

neavra:
## Summary

Removes all calldata_id client-side resolution from the SDK CLI. Calldata is now delivered via `tx_ready` SSE events with full hex — the client no longer needs to fetch calldata references from the agent-backend API.

### What changed

**executor.ts:**
- Removed `backendClient` field and `setBackendClient()` method
- Removed calldata_id resolution in `buildTx` (the `if (params.calldata_id && !params.data)` block)
- Removed raw calldata pass-through in `buildTx` (the `if (params.data || params.calldata || params.hex_payload)` block) — server-built calldata arrives via tx_ready, not via action params
- Replaced silent fallback to `buildSendTx` with an explicit error throw
- `buildTx` now only handles ABI-encoding from `function_name + contract_address`

**client.ts:**
- Removed `getCalldata()` HTTP method (called `GET /agent/calldata/:id`)

**session.ts:**
- Removed `setBackendClient` wiring after auth

### Why

Both Sonnet and minimax were observed putting the literal string `calldata_id: cd_xxx` into the tx `data` field, causing on-chain reverts. The calldata_id mechanism required the LLM to correctly place a machine-generated reference into the right tool parameter — a fundamentally fragile design.

The companion agent-backend PR (#109) moves calldata stashing after tx_ready emission, so tx_ready always carries full calldata. Clients sign directly from tx_ready without needing calldata_id resolution.

### Depends on

- vultisig/agent-backend#109 must deploy first — it ensures tx_ready carries full calldata instead of calldata_id references

## Test plan

- [ ] `yarn check` passes
- [ ] Swap ETH → USDC lands on-chain (tx_ready flow, no calldata_id)
- [ ] Pendle buy/sell PT lands on-chain (calldata_id flows through build_evm_tx on agent-backend side, never reaches SDK CLI)
- [ ] Simple send still works (no regression)

PR #407, opened 2026-05-05
fix(cli): remove calldata_id resolution

gomesalexandre:
## Summary

- Ports #247 (neavra) onto current main — original branch had merge conflicts from `evmLastBroadcast` addition in #402 and the improved hex-payload path
- Conflict resolution: kept `evmLastBroadcast` (EVM broadcast dedup tracking) and the improved `params.data || params.calldata || params.hex_payload` path with `stored` guard; removed `backendClient` + `calldata_id` resolution as intended by the original PR
- `tx_ready` SSE now delivers calldata directly — the client-side `calldata_id` resolution was a stale codepath

Supersedes #247 (conflicting).

🤖 Agent-generated

## Summary by CodeRabbit

* **Refactor**
  * Simplified the agent executor by removing support for client-side calldata resolution. Server-built calldata must now be provided directly via the `tx_ready` action rather than being resolved at the client level.
  * Enhanced validation and error handling for contract specifications with improved messaging when parameters are incomplete or unrecognized.

### DIFF
--- a/clients/cli/src/agent/client.ts
+++ b/clients/cli/src/agent/client.ts
@@ -411,20 +411,6 @@ export class AgentClient {
   }
 
   // ============================================================================
-  // Calldata
-  // ============================================================================
-
-  async getCalldata(id: string): Promise<{ data: string; to?: string; chain?: string }> {
-    const res = await fetch(`${this.baseUrl}/agent/calldata/${id}`, {
-      headers: this.authToken ? { Authorization: `Bearer ${this.authToken}` } : {},
-    })
-    if (!res.ok) {
-      throw new Error(`Failed to resolve calldata_id ${id}: ${res.status} ${res.statusText}`)
-    }
-    return res.json() as Promise<{ data: string; to?: string; chain?: string }>
-  }
-
-  // ============================================================================
   // Private helpers
   // ============================================================================
 
--- a/clients/cli/src/agent/executor.ts
+++ b/clients/cli/src/agent/executor.ts
@@ -75,10 +75,6 @@ export class AgentExecutor {
   /** Held chain lock release functions, keyed by chain name */
   private chainLockReleases = new Map<string, () => Promise<void>>()
   private evmLastBroadcast = new Map<string, number>()
-  /** Backend client for resolving calldata_id references. */
-  private backendClient: {
-    getCalldata(id: string): Promise<{ data: string; to?: string; chain?: string }>
-  } | null = null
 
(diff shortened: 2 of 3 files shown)


NeOMakinG (approved):
CI still running at review time (Unit Tests, Integration Tests, Lint — all in-progress). CodeRabbit: SUCCESS.

## testing approach

Static code review only per `feedback_rereview_skip_retest.md` guidance (this is round 1 but the PR is a conflict-resolution port of #247 with no new behaviour, not a fresh feature). Cross-repo grep confirmed `AgentClient.getCalldata` and `AgentExecutor.setBackendClient` have zero callers outside the SDK CLI itself.

## should-fix

1. **No regression test for the new throw in `buildTx`** — `executor.ts` line 816–821 now throws `build_custom_tx: unrecognized params shape` instead of falling through to `buildSendTx`. That is the right call, but no test locks it in. If a future refactor re-adds a native-transfer path through `build_custom_tx`, the throw will suppress it silently.

   action: add a test in `clients/cli/src/agent/__tests__/executor-actions.test.ts` that drives `build_custom_tx` with `{chain: "Ethereum", to: "0x...", amount: "0.01"}` (no `data`, no `function_name`, no `contract_address`) and asserts the specific error message. One test, ~10 lines.

## q

1. The PR description mentions keeping the "`stored` guard" — confirming this refers to the `if (!stored) throw` check at `executor.ts` line 810 (the `storeServerTransaction` return value). Was there a prior silent-discard bug in that guard, or is this just documenting that the guard was not removed?

## risk

Low. The removed `GET /agent/calldata/:id` client-side fetch was a dead path: the backend resolves `calldata_id → data` server-side in `resolveCalldataInInput` (executor.go:2114) before any action reaches the SDK, so the client-side resolution only fired if the server already failed. The `buildSendTx` fallback removal is a behavior change but `build_custom_tx` native-transfer params have no known live caller after the backend refactor.

## verdict

APPROVE — clean removal of a stale client-side calldata indirection. Single test gap noted above; not blocking.

🤖 reviewed via [Claude Code](https://claude.com/claude-code), single static lane. cross-repo grep: agent-backend + vultiagent-app + mcp-ts. no runtime run (pure deletion PR).

gomesalexandre:
## post-merge receipts (retest 2026-05-06)

### unit tests
```bash
yarn workspace @vultisig/cli test  # PASS
yarn test:core  # 401/401 PASS
```

removes calldata_id resolution from CLI tx_ready SSE path. calldata delivered directly now. port of #247 with conflict resolution keeping evmLastBroadcast.

**app surface**: CLI tx flow. no mobile sim surface.

verdict: PASS

PR #418, opened 2026-05-06
[sdk/swap] Report all quote provider failures

premiumjibles:
## Summary
- add provider labels to swap quote fetchers so all-fail errors name every attempted provider
- report only the attempted provider names in the user-facing all-fail message
- keep best-successful-quote selection and the existing dust-threshold user-facing error unchanged

## Receipts
- Focused unit: `node node_modules/vitest/vitest.mjs run --config .config/vitest.core.config.ts packages/core/chain/swap/quote/findSwapQuote.selection.test.ts`
  - Result: PASS, 9 tests passed.
- Diff hygiene: `git diff --check`
  - Result: PASS.
- SDK node bundle for CLI smoke: `PATH="/home/sean/Repos/vultisig/vultisig-sdk/node_modules/.bin:/home/sean/.codex/tmp/arg0/codex-arg0958GRy:/home/sean/.local/lib/node_modules/@openai/codex/node_modules/@openai/codex-linux-x64/vendor/x86_64-unknown-linux-musl/path:/home/sean/.bun/bin:/home/sean/.opencode/bin:/home/sean/.local/bin:/home/sean/go/bin:/home/sean/.cargo/bin:/home/sean/Android/Sdk/platform-tools:/home/sean/.claude/bin:/usr/lib64/ccache:/home/sean/.cargo/bin:/home/sean/.local/bin:/usr/local/bin:/usr/bin:/usr/sbin:/home/sean/.maestro/bin" yarn workspace @vultisig/sdk build:fast`
  - Result: PASS; Rollup emitted existing dependency warnings only.
- CLI smoke: `node clients/cli/dist/index.js swap-quote Polygon Polygon 1 --to-token BNB -o json`
  - Result: PASS, exercised the all-provider-failed CLI path.
  - Output excerpt:
    ```json
    {
      "success": false,
      "v": 1,
      "error": {
        "code": "USAGE_ERROR",
        "exitCode": 1,
        "message": "Swap failed: No swap route found after trying KyberSwap, 1inch, LiFi.",
        "retryable": false
      }
    }
    ```

## Notes
- `yarn test:core packages/core/chain/swap/quote/findSwapQuote.selection.test.ts` could not run in this checkout because Yarn did not resolve `vitest` (`command not found: vitest`); used the local Vitest binary directly.
- Full core suite was attempted and failed on unrelated environment/test issues: missing `@ton/core` in `ton/messageBody/decode.test.ts` and timeout in `packages/core/mpc/message/server.test.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap quote failure reporting: when all providers fail, the app now shows a single consolidated, sanitized message listing attempted providers (omitting verbose provider-specific error text) while preserving special-case dust-threshold handling.

* **Tests**
  * Updated tests to verify the consolidated, sanitized failure message and the all-providers-fail scenario.

### DIFF
--- a/packages/core/chain/swap/quote/findSwapQuote.selection.test.ts
+++ b/packages/core/chain/swap/quote/findSwapQuote.selection.test.ts
@@ -190,7 +190,7 @@ describe('findSwapQuote parallel selection', () => {
     expect(quote.quote.general.provider).toBe('kyber')
   })
 
-  it('when all providers fail, propagates the first fetcher-order rejection', async () => {
+  it('when all providers fail, reports every attempted provider', async () => {
     const mayaError = 'maya last error'
     vi.mocked(getKyberSwapQuote).mockRejectedValue(new Error('first fail'))
     vi.mocked(getOneInchSwapQuote).mockRejectedValue(new Error('second fail'))
@@ -207,7 +207,36 @@ describe('findSwapQuote parallel selection', () => {
         ...evmSameChainCoins,
         amount: 1n,
       })
-    ).rejects.toThrow('first fail')
+    ).rejects.toThrow(
+      'No swap route found after trying KyberSwap, 1inch, LiFi, THORChain, MayaChain.'
+    )
+  })
+
+  it('omits noisy provider errors from the all-fail message', async () => {
+    const longKyberError = `kyber ${'x'.repeat(220)}`
+
+    vi.mocked(getKyberSwapQuote).mockRejectedValue(new Error(longKyberError))
+    vi.mocked(getOneInchSwapQuote).mockRejectedValue(new Error('inch fail'))
+    vi.mocked(getLifiSwapQuote).mockRejectedValue(new Error('lifi fail'))
+    vi.mocked(getNativeSwapQuote).mockRejectedValue(new Error('native fail'))
+
+    try {
+      await findSwapQuote({
+        ...evmSameChainCoins,
+        amount: 1n,
+      })
+      throw new Error('Expected findSwapQuote to fail')
+    } catch (error) {
+      if (!(error instanceof Error)) {
+        throw error
+      }
+
+      expect(error.message).toBe(
+        'No swap route found after trying KyberSwap, 1inch, LiFi, THORChain, MayaChain.'
+      )
+      expect(error.message).not.toContain(longKyberError)
+      expect(error.message).not.toContain('inch fail')
+    }
   })
 
   it('maps dust threshold on any provider to the user-facing message (Maya in this setup)', async () => {
--- a/packages/core/chain/swap/quote/findSwapQuote.selection.test.ts
+++ b/packages/core/chain/swap/quote/findSwapQuote.selection.test.ts
@@ -42,10 +42,7 @@ const evmSameChainCoins = {
   },
 } as const
 
-function minimalGeneralQuote(
-  dstAmount: string,
-  provider: 'kyber' | '1inch'
-): GeneralSwapQuote {
+function minimalGeneralQuote(dstAmount: string, provider: 'kyber' | '1inch'): GeneralSwapQuote {
   const base = {
     dstAmount,
     tx: {
@@ -57,9 +54,7 @@ function minimalGeneralQuote(
       },
     },
   }
-  return provider === 'kyber'
-    ? { ...base, provider: 'kyber' }
-    : { ...base, provider: '1inch' }
+  return provider === 'kyber' ? { ...base, provider: 'kyber' } : { ...base, provider: '1inch' }
 }
 
 function minimalNativeQuote(swapChain: Chain, expected_amount_out: string): NativeSwapQuote {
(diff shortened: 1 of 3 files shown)


gomesalexandre (commented):
draft, so comment-only pass. ran claude correctness + codex adversarial in parallel and synthesized below.

---

## what this PR does

replaces the old `SwapQuoteFetcher = () => Promise<SwapQuote>` with `{ providerName, fetch }`, then on all-fail emits a single aggregated error naming each provider and its reason. selection logic is a structural no-op refactor (same indices, same `Promise.allSettled` wrapper, `selectBestEligibleQuote` unchanged).

## selection invariant — confirmed preserved

`selectBestEligibleQuote` is completely untouched. the `Promise.allSettled` mapper is:

```ts
fetchers.map(async fetcher => {
  const quote = await fetcher.fetch()
  return { quote, outputAmount: getComparableOutputAmount(quote, to) }
})
```

so `settled[i].status === 'fulfilled'` iff both `fetch()` AND `getComparableOutputAmount()` succeed. `selectBestEligibleQuote` only processes `fulfilled` entries. if it returns non-null we return early. the error aggregation path is only ever reached when EVERY `settled[i]` is rejected. `fetchers` and `settled` are always co-indexed by `.map`. index alignment: solid.

## dust threshold interaction — confirmed correct

dust check fires inside the `best === null` branch, BEFORE the aggregation. if any provider succeeds, we return early and the dust loop is never evaluated. no false-positive dust errors possible from a partial-fail scenario.

---

## findings

**should-fix: `getComparableOutputAmount` throw leaks internal SDK error into provider error message**

when `fetcher.fetch()` resolves but `getComparableOutputAmount` throws (e.g. `BigInt('not-a-number')` from a malformed `dstAmount`), `settled[i]` becomes `rejected` with the SyntaxError. this lands in `providerErrors` as `"KyberSwap: Cannot convert 'not-a-number' to a BigInt"`. that's a JS SDK internal error, not a provider-side failure - but the user-facing string can't distinguish the two. it also means a local metadata/normalization bug could surface as a blamed provider error in aggregate logs. the existing test "does not let a malformed provider amount hide a succeeding one" covers the case where another provider succeeds (so we never hit the error path) but does NOT cover the case where all providers return malformed amounts and the error message contains SyntaxError text.

action: wrap the `getComparableOutputAmount` call separately to distinguish the two failure modes, or at minimum add a test that pins the error string when `dstAmount` is non-numeric on ALL providers.

**suggestion: native chain label ternary is a maintenance trap**

```ts
providerName: swapChain === Chain.THORChain ? 'THORChain' : 'MayaChain'
```

`SwapQuoteProviderName` is a closed union of 5 strings and `nativeSwapChains` currently has 2 entries, so this is fine today. but if a third native chain ever gets added to `nativeSwapChains`, it silently inherits the `MayaChain` label in every error message. a lookup map (`const nativeChainLabel: Record<NativeSwapChain, SwapQuoteProviderName>`) would fail loudly at compile time instead of drifting.

**q: `SwapQuoteProviderName` lives in `findSwapQuote.ts` - intentional?**

it seems like a type that other modules (logging, analytics, error handling) might want to import. if so, exporting it makes future cross-module use cleaner. if it's intentionally local, fine as-is.

---

## codex adversarial note

codex flagged a "PREFERABLY-BLOCKING: fulfilled results could disappear from providerErrors leaving empty list" - after tracing the actual control flow, this is a false positive. `best === null` is only reachable when all settled are rejected (the fulfilled → `best` path returns early). the invariant holds. called out here for transparency.

---

## test coverage summary

- all-fail with named providers: yes
- truncation of noisy errors: yes
- partial fail (some providers succeed): yes
- dust threshold from non-last provider: yes
- `getComparableOutputAmount` throws when all providers return malformed amounts: **missing** (see should-fix above)

---

praise: the `sanitizeSwapQuoteErrorReason` function is tight - collapse whitespace, 160-char cap, `Unknown error` fallback. no footgun. the error string format `"KyberSwap: first fail; 1inch: second fail..."` is genuinely useful for debuggability. good PR.

premiumjibles:
Follow-up from review/CLI smoke:

- Changed the all-provider-failed message to list provider names only, e.g. `No swap route found after trying KyberSwap, 1inch, LiFi.`
- Provider/API reasons are no longer included in the user-facing SDK error, so malformed amounts or provider response bodies cannot leak through this aggregate message.
- Kept the dust-threshold special case unchanged.
- Added/updated tests to pin the provider-list-only message and verify noisy provider errors are omitted.
- Added receipts to the PR body, including the CLI smoke output.

PR is marked ready for review and CI is green.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -256,13 +278,17 @@ export const findSwapQuote = async ({
...
+      if (result.status === 'fulfilled') {
+        return null
+      }
 
-  // Defensive: when fetchers are non-empty, `best` is null only if every
-  // fetcher rejected, and the loop above throws the first reason.
-  throw new Error('No quote')
+      return fetchers[index].providerName
+    })
+    .filter((providerName): providerName is SwapQuoteProviderName => providerName !== null)
```
0xApotheosis, line 255: Prettier check fails on this file; please run the repo formatter.

packages/core/chain/swap/quote/findSwapQuote.selection.test.ts
```diff
@@ -207,7 +207,36 @@ describe('findSwapQuote parallel selection', () => {
...
+        amount: 1n,
+      })
+      throw new Error('Expected findSwapQuote to fail')
+    } catch (error) {
+      if (!(error instanceof Error)) {
+        throw error
+      }
+
+      expect(error.message).toBe(
+        'No swap route found after trying KyberSwap, 1inch, LiFi, THORChain, MayaChain.'
```
0xApotheosis: Prettier also rewrites this test file; please run the repo formatter.

0xApotheosis (changes requested):
Requesting changes for formatting only: the changed TypeScript files fail the repo Prettier check.

0xApotheosis (approved):
Verified the formatter follow-up on current head 9f62fe513f195dc0e09cbf0ba7ea1ca12d0302b8. Cross-repo pass: SDK/mcp-ts consumers preserve the thrown message surface, agent-backend only reads successful provider metadata, and vultiagent-app is gated on the published SDK version. Local focused core tests, targeted SDK unit tests, SDK node bundle build, diff/Prettier checks, and mcp-ts test suite passed. No blocking issues found.

PR #427, opened 2026-05-07
fix(solana): verify durable signature relay

realpaaao:
Summary
- Let Solana standard RPC relay use its normal retry window instead of capping retries at 3.
- Mark Solana signatures that are missing from getSignatureStatuses as unknown pending, so broadcast verification does not swallow real relay failures.
- Preserve known THORChain single-segment denoms such as TCY during Cosmos coin discovery.

Test Plan
- yarn test:core packages/core/chain/tx/status/resolvers/solana.test.ts packages/core/chain/tx/broadcast/verifyBroadcastByHash.test.ts packages/core/chain/tx/broadcast/resolvers/solana.test.ts packages/core/chain/coin/find/resolvers/cosmos.test.ts
- yarn build:all
- yarn check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana transaction broadcast behavior by adjusting standard RPC relay retry handling.
  * Enhanced Solana transaction status verification to correctly distinguish between known and unknown pending transactions, preventing incorrect confirmation of unverified broadcasts.
  * Fixed Cosmos coin discovery to properly preserve single-segment THORChain denominations like `tcy`.

### DIFF
--- a/.changeset/cosmos-native-denom-ticker.md
+++ b/.changeset/cosmos-native-denom-ticker.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Preserve known THORChain single-segment denoms such as `tcy` during Cosmos coin discovery, let Solana standard RPC relay retry for its normal validity window, and keep Solana broadcast verification from treating unknown signatures as confirmed pending transactions.
--- a/packages/core/chain/coin/find/resolvers/cosmos.test.ts
+++ b/packages/core/chain/coin/find/resolvers/cosmos.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it, vi } from 'vitest'
+
+const getAllBalancesMock = vi.fn()
+
+vi.mock('@vultisig/core-chain/chains/cosmos/client', () => ({
+  getCosmosClient: () => ({
+    getAllBalances: getAllBalancesMock,
+  }),
+}))
+
+import { Chain } from '../../../Chain'
+import { findCosmosCoins } from './cosmos'
+
+describe('findCosmosCoins', () => {
+  it('keeps known single-segment THORChain denoms like TCY', async () => {
+    getAllBalancesMock.mockResolvedValue([
+      { denom: 'rune', amount: '1' },
+      { denom: 'tcy', amount: '2' },
+    ])
+
+    const coins = await findCosmosCoins({
+      address: 'thor1address',
+      chain: Chain.THORChain,
+    })
+
+    expect(coins).toEqual([
+      expect.objectContaining({
+        id: 'tcy',
+        ticker: 'TCY',
+        logo: 'tcy',
+      }),
+    ])
+  })
+})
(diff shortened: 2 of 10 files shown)


gomesalexandre (approved):
gm @realpaaao - reviewed the diff in full and ran the test suite locally. clean fund-safety patch ser.

## what I checked

### 1. Solana broadcast verify - the meaty fund-safety win

Root cause: prior `verifyBroadcastByHash` swallowed the broadcast error whenever `getTxStatus` returned `pending`, regardless of whether the RPC actually recognized the signature. So a broadcast that legitimately failed (network blip, JITO reject, validator gossip drop) would surface in the UI as "in-flight, awaiting confirmation" until the user noticed funds never moved. This was a real footgun.

Fix is surgical and reads correctly:
- `getSolanaTxStatus` calls `getSignatureStatuses` FIRST. If the RPC's signature index has no record (`value[0] === null`), return `{ status: 'pending', isKnown: false }`. If known with `err`, error. If known and `getTransaction` not yet indexed, `pending + isKnown:true` (the legitimate "broadcast ok, just waiting for the tx body to be indexed" case).
- `verifyBroadcastByHash` now keys on `isKnown !== false` for the swallow branch. Non-Solana chains never set `isKnown`, so `undefined !== false` is true and the existing semantics are preserved. ✓ no regression risk for EVM / Cosmos / Bitcoin.
- `searchTransactionHistory: true` is correct - broadcast verification can run minutes after the broadcast attempt for slow MPC sessions, and the recent-only index would miss valid sigs.

### 2. `maxRetries: 3` removal

Letting the RPC's standard retry window run. Fund-safety direction: more retries means a stuck-in-mempool tx has a longer broadcast window before `verifyBroadcastByHash` even gets called, which is the right tradeoff for durable nonce / MPC scenarios. The verify path now correctly distinguishes "RPC saw it" from "real failure" so the previous over-cautious retry cap isn't compensating for verify-path ambiguity anymore.

### 3. Cosmos single-segment denom (TCY)

The previous `attempt(() => shouldBePresent(denom.split(/[-./]/)[1]?.toUpperCase()))` short-circuited via `if ('error' in tickerAttempt)` BEFORE the `ticker || tickerAttempt.data` fallback could fire - so even when `getCosmosTokenMetadata({chain, id: 'tcy'})` correctly returned `{ ticker: 'TCY' }` from `knownTokens/cosmos.ts`, the entry was dropped because the split-attempt errored on the no-separator denom and the function returned undefined. New `coinTicker = ticker || denom.split(...)?.[1]?.toUpperCase()` reaches the metadata-supplied ticker. Verified against `knownTokens/cosmos.ts:39` (`tcy: { ticker: 'TCY', logo: 'tcy.png', decimals: 8 }`). Test pins it.

Defensive note: `denom.split(/[-./]/)[1]?.toUpperCase()` for a denom WITHOUT a separator (e.g. some hypothetical future single-segment denom not in `knownTokens`) still returns `undefined`, and the `if (!coinTicker)` branch logs + drops it. Same behavior as before for unknown single-segment denoms - the only thing that changes is denoms with `getCosmosTokenMetadata` populating `ticker` survive.

## receipts

```
$ yarn test:core packages/core/chain/tx/status/resolvers/solana.test.ts \
                packages/core/chain/tx/broadcast/verifyBroadcastByHash.test.ts \
                packages/core/chain/tx/broadcast/resolvers/solana.test.ts \
                packages/core/chain/coin/find/resolvers/cosmos.test.ts
 RUN  v4.1.5 /Users/gomes/Sites/vultisig-sdk
 Test Files  4 passed (4)
      Tests  13 passed (13)
   Duration  154ms

$ yarn build:all  # green
$ yarn check      # typecheck exited with code 0
```

CI: lint + integration vitest + codeRabbit all green; unit-tests job still pending at review time.

## verdict

LGTM ser, fund-safety is tighter post-PR. dis stamped.

PR #430, opened 2026-05-07
fix(cardano): produce broadcastable bodies for native-token sends

### LINKED ISSUE #429: fix(cardano): CNT keysign produces unbroadcastable bodies (zero fee + below-min-UTxO output)
## Summary

The Cardano signing-input resolver (\`packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts\`) and the keysign UTXO populator (\`packages/core/mpc/keysign/utxo/getKeysignUtxoInfo.ts\`) ship lovelace-only inputs and never call \`AnySigner.plan(...)\`. Every CNT send produces a body that:

- has \`fee = 0\` → Ogmios rejects with code 3122 \"insufficient fee\".
- has the recipient output's lovelace set to the *token amount* (because \`transferMessage.amount = keysignPayload.toAmount\`, and \`toAmount\` for a CNT send is in the token's base units, not lovelace) → network rejects with code 3125 \"insufficiently funded outputs\".

The keysign protocol round-trips fine because both peers (Windows extension + iOS, or Windows + Windows) hash the **same** broken body, so the relay's \`message_id\` header (\`MD5(messageToSign)\`) matches and setup-message routing works. The bug is at the broadcast step, after MPC signing.

## How to reproduce

1. Vault holds USDM (or any CNT) with sufficient ADA for fee + min-UTxO.
2. Initiate a Cardano native-token send for any small amount, e.g. 0.665 USDM.
3. Complete the keysign session (succeeds — both peers sign).
4. iOS/Windows attempts to broadcast → Ogmios returns 3122 (\`providedFee = 0\`) or 3125 (\`minimumRequiredValue = 849070, output lovelace = 665000\`).

User-visible result: signed transaction never confirms; user sees the network error in the post-keysign UI.

### 1. Per-input \`tokenAmount\` is missing on \`TW.Cardano.Proto.TxInput\`

\`signingInputs/resolvers/cardano.ts\` builds inputs from \`keysignPayload.utxoInfo\` (lovelace-only, hash/amount/index). With a TokenBundle on the Transfer output and zero token data on the inputs, the WalletCore Cardano planner can't reconcile a token-balanced body. Without a populated \`input.plan\`, this manifests downstream rather than in \`plan()\` itself, but it leaves the planner unable to compute a valid plan even if it were called.

\`getKeysignUtxoInfo.ts\` uses \`getCardanoUtxos\` (lovelace-only); the asset-aware variant at \`packages/core/chain/chains/cardano/utxo/getCardanoExtendedUtxos.ts\` already exists but isn't consumed.

### 2. \`AnySigner.plan(...)\` is never called for Cardano

The current resolver returns a \`SigningInput\` with no \`input.plan\` set. WalletCore's signer reads the body's fee from \`plan.fee\` ([Signer.cpp:80](https://github.com/trustwallet/wallet-core/blob/master/src/Cardano/Signer.cpp#L80) — \`tx.fee = plan.fee\`), and \`forceFee\` on the Transfer message is only consulted **inside** \`doPlan()\` ([Signer.cpp:551-555](https://github.com/trustwallet/wallet-core/blob/master/src/Cardano/Signer.cpp#L551-L555)). So setting \`forceFee\` without running \`plan()\` is a no-op — the body's fee field stays at its default \`0\`.

### 3. \`transferMessage.amount\` for CNT sends is the token amount, not lovelace

[\`signingInputs/resolvers/cardano.ts:52\`](https://github.com/vultisig/vultisig-sdk/blob/main/packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts#L52):

\`\`\`typescript
amount: Long.fromString(keysignPayload.toAmount),
\`\`\`

For an ADA send, \`keysignPayload.toAmount\` is lovelace and this works. For a CNT send, \`toAmount\` is the *token amount* in the token's base units. WalletCore's planner copies this value verbatim into \`plan.amount\` ([Signer.cpp:481](https://github.com/trustwallet/wallet-core/blob/master/src/Cardano/Signer.cpp#L481): \`plan.amount = input.transfer_message().amount()\`), and it never bumps it up to min-UTxO — it only caps it ([Signer.cpp:559-563](https://github.com/trustwallet/wallet-core/blob/master/src/Cardano/Signer.cpp#L559-L563): \`std::min(plan.amount, ...)\`).

\`requestedTokens.minAdaAmount()\` is used at line 514 to size **input selection**, not to enforce output min-UTxO. WalletCore has no min-UTxO enforcement on the output side; whatever \`transferMessage.amount\` is becomes the recipient's lovelace, and the body goes out with that value.

For a 0.665 USDM send: \`transferMessage.amount = 665000\` → recipient output funded with 0.665 ADA → below the ~0.85 ADA min-UTxO floor → network 3125.

### 4. \`useMaxAmount\` is set for CNT sends when token balance == amount

\`getCardanoChainSpecific\` ([\`chainSpecific/resolvers/cardano.ts\`](https://github.com/vultisig/vultisig-sdk/blob/main/packages/core/mpc/keysign/chainSpecific/resolvers/cardano.ts)) sets \`sendMaxAmount = balance === amount\` regardless of chain context. For a user sending their entire CNT holding, this trips. \`useMaxAmount\` is an ADA-only flag — when set, the WalletCore signer drains every input lovelace into the recipient output (subtracting fee). For CNT \"Send Max\" it should mean \"send all of the token\" with the lovelace floor fixed at min-UTxO.

## Fix list

1. Source extended UTXO data (Koios \`address_utxos?_extended=true\`) at sign time so each input has its \`asset_list\`. Either by switching \`getKeysignUtxoInfo.ts\` to use \`getCardanoExtendedUtxos\` AND extending \`UtxoInfo\` in commondata to carry assets — OR by making the Cardano signing-input resolver async and fetching at sign time on each peer (no commondata change; both peers fetch independently and Koios returns the same UTXOs for the same address). Recommending the async-resolver path to avoid commondata churn.
2. Populate \`TxInput.tokenAmount\` per UTXO in the resolver from the extended UTXO data.
3. Call \`walletCore.AnySigner.plan(input, walletCore.CoinType.cardano)\` after building the input; assign the result to \`input.plan\`. Throw on \`plan.error !== OK\`.
4. Override \`transferMessage.amount\` to a min-UTxO floor (\`1500000n\`) when \`isTokenSend\`. Static value covers single-CNT outputs comfortably; computing the exact min-UTxO from bundle CBOR size is a follow-up. Without this, code 3125 reappears even after (1)-(3).
5. Force \`useMaxAmount = false\` when \`isTokenSend\`.

## Cross-platform note

\`vultisig-ios\` already has these fixes implemented locally (commit \`8379c006a\` in the open PR vultisig/vultisig-ios#4324) but had to revert because shipping the fix on iOS alone makes iOS↔extension keysign comms break: iOS produces a different (correct) body, the extension produces the old (broken) body, the \`message_id\` MD5s diverge, the relay returns 404 on setup-message lookup, and the session times out. iOS will re-apply once this SDK PR ships.

## Out of scope

- Computing exact min-UTxO from bundle CBOR size (use static 1.5 ADA in this PR; revisit if/when multi-CNT bundles are supported).
- Multi-CNT in a single transfer (Cardano protocol allows it; SDK currently checks \`requestedTokens.size() > 1\` and rejects — that's fine for v1).

gastonm5:
Closes #429. Depends on [commondata#75](https://github.com/vultisig/commondata/pull/75) — `UtxoInfo.cardano_tokens` field.

## Summary

The Cardano signing-input resolver shipped lovelace-only inputs and never called `AnySigner.plan(...)`, producing bodies that fail at broadcast with two distinct network errors. The keysign protocol round-trips fine because both peers hash the same broken body, so the relay's `message_id` MD5 matches and setup-message routing works. The bug is **broadcast-side only**.

| Code | Reason | Fixed by |
|---|---|---|
| 3122 "insufficient fee" | Body has `fee = 0` because no `input.plan` was set; WalletCore reads the body's fee from `plan.fee` (Signer.cpp:80), not from `forceFee` on Transfer (only consulted inside `doPlan()`, lines 551-555). | (3) |
| 3125 "insufficiently funded outputs" | `transferMessage.amount` was `keysignPayload.toAmount` — for CNT sends that's the *token* amount in base units (e.g. 665000 = 0.665 USDM), not lovelace. WalletCore copies it verbatim into `plan.amount` (Signer.cpp:481) and never bumps it up — only caps it (lines 559-563: `std::min(plan.amount, ...)`). Recipient output ends up with 0.665 ADA, below per-output min-UTxO. | (4) |

## Changes

`packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts`:

1. **Read per-`TxInput.tokenAmount` off the wire** from `keysignPayload.utxoInfo[i].cardano_tokens`. The initiator (the app the user clicked Send on) fetches extended UTxOs from Koios once when building the payload, sorts assets canonically by `(policyId, assetNameHex)`, and serialises them onto each `UtxoInfo`. Both MPC peers read identical bytes from the payload — determinism is structural, not Koios-ordering-dependent.
2. **Populate `transferMessage.tokenAmount`** from `coin.contractAddress` + `keysignPayload.toAmount` for CNT sends (unchanged from prior pass).
3. **Call `walletCore.AnySigner.plan(...)`** and assign `input.plan`; post-plan, sync `forceFee = plan.fee`. Throw on plan errors.
4. **Override `transferMessage.amount` to a 1.5 ADA min-UTxO floor** for CNT sends. Computing exact min-UTxO from bundle CBOR size is a follow-up; 1.5 ADA covers single-CNT outputs comfortably.
5. **Force `useMaxAmount = false` for CNT sends.** The flag is ADA-only — when set, WalletCore drains every input lovelace into the recipient output. For CNT "Send Max" means "send all of the token" with the lovelace floor fixed at min-UTxO.

`packages/core/mpc/keysign/signingInputs/{resolver,index}.ts`:

- **Drop `AsyncSigningInputsResolver` + the dual sync/async dispatcher.** The resolver is synchronous again now that the Koios fetch moved to the initiator. `cosigner.ts`, `BroadcastService.ts`, `TransactionBuilder.ts` revert to the sync `getEncodedSigningInputs`.

`packages/core/mpc/types/vultisig/keysign/v1/utxo_info_pb.ts`:

- Regenerated against commondata#75 to pick up `cardano_tokens` + `CardanoTokenAsset`. Wire format unchanged for non-Cardano chains.

## Determinism note

WalletCore's Cardano coin-selection is largest-first deterministic — `Cardano/Signer.cpp::selectInputsSimpleNative` and `selectInputsSimpleToken` both `std::sort` descending by amount with no RNG. With identical inputs on both peers (now structural under this PR), `AnySigner.plan(...)` produces identical body bytes on both sides.

## Cross-platform note

Companion PRs:
- iOS — vultisig/vultisig-ios#4324 (proposer-side `cardano_tokens` population in `KeysignPayloadFactory.selectCardanoUTXOs`, cosigner-side read from `keysignPayload.utxos[i].cardanoTokens`).
- Windows — vultisig/vultisig-windows#3870 (proposer-side `cardano_tokens` population in the Windows payload-build path; reverts the await dispatcher consumer changes from the C-shaped pass).
- Android — follow-up (no breakage for cosigning; Android-initiated CNT sends need the same proposer-side update once commondata#75 merges).

## Test plan

- [x] `yarn check` clean (typecheck/lint/format/knip)
- [x] `yarn test` clean (1360 unit tests pass)
- [ ] **Mainnet end-to-end (Windows extension proposer ↔ vultisig-ios cosigner):**
  - [ ] Send a small ADA-only amount — confirms the planner path still works for the non-token case
  - [ ] Send a small CNT amount (e.g. 0.1 USDM) — confirms 3122 + 3125 are gone
  - [ ] Verify on the explorer: tx confirmed, fee matches the planner's value, recipient output has both the token bundle and at least 1.5 ADA
- [ ] **iOS-only (FastVault):** same two test sends — confirms server-assisted signing path still works under the new resolver

## Out of scope

- Computing exact min-UTxO from bundle CBOR size (use static 1.5 ADA in this PR; revisit if/when multi-CNT bundles are supported)
- Multi-CNT in a single transfer (the resolver still rejects multi-token bundles via WalletCore's `requestedTokens.size() > 1` check, which is fine for v1)
- Unit tests for the new resolver behavior — recommend adding byte-parity fixtures for ADA + CNT signing inputs in a follow-up


gomesalexandre (commented):
draft PR so I'm leaving this as comments only — but flagging concerns before this leaves draft.

## praise

the diagnosis here is excellent. `Signer.cpp:80`, `Signer.cpp:481`, `Signer.cpp:551-555`, and `Signer.cpp:559-563` callouts in the PR body + inline comments are exactly the kind of forensics that make a fund-safety patch reviewable. determinism note re largest-first coin selection is the right thing to call out for the MPC byte-parity invariant.

## preferably-blocking

### 1. no unit-test coverage for the resolver behavior change

PR body acknowledges this:

> Unit tests for the new resolver behavior — recommend adding byte-parity fixtures for ADA + CNT signing inputs in a follow-up

for a CHANGE THAT BUILDS THE BYTES WALLETCORE SIGNS, this is the wrong tier of risk to defer to "follow-up". the only added test is a body-hash pinning for `MINIMAL_BODY_HEX`. the actual resolver — `getCardanoSigningInputs` — has zero direct assertions in this PR.

minimum surface I'd want before un-drafting:

a) **CNT path fixture**: call `getCardanoSigningInputs` with a hardcoded `keysignPayload` (2-3 utxos, 1-2 token bundles each, single recipient CNT send) and assert:
   - `input.plan.error === SigningError.OK`
   - `input.plan.fee > 0n` (verifies the planner was actually called)
   - `input.transferMessage.amount === 1_500_000n` (the lovelace floor for CNT)
   - `input.transferMessage.useMaxAmount === false` (override regardless of upstream flag)
   - `input.transferMessage.tokenAmount.policyId === <expected>` (CNT bundle)
   - `input.utxos[i].tokenAmount` populated for each utxo (per-input token data round-trips)

b) **ADA-only path fixture**: same shape, no tokens. assert:
   - `input.plan.fee > 0n`
   - `input.transferMessage.amount === BigInt(keysignPayload.toAmount)` (no override)
   - `input.transferMessage.useMaxAmount === sendMaxAmount` (passthrough)
   - `input.utxos[i].tokenAmount === undefined` (the `...(tokenAmounts ? { tokenAmount: tokenAmounts } : {})` branch correctly omits the field for ADA-only utxos)

c) **multi-token utxo sort determinism**: feed `cardanoTokens` in a non-canonical order at the keysignPayload level (out-of-order by policyId), call the resolver twice, assert the produced `input.utxos[].tokenAmount` arrays are byte-equal across runs. catches a sort-stability regression at the resolver layer specifically (the existing sort in `getKeysignUtxoInfo.ts` is a different layer).

without these, a future regression in `useMaxAmount` override or in the planner-call removal would ship green. byte-parity between peers IS the safety property; if a unit test can't catch divergence, the only check is mainnet end-to-end.

### 2. cross-platform compat gap — Android initiators

PR body says:

> Android — follow-up (no breakage for cosigning; Android-initiated CNT sends need the same proposer-side update once commondata#75 merges).

so until Android lands the proposer-side change:
- Android-initiated CNT send → vault has iOS/Windows cosigner with this PR's fix
- Android serialises `UtxoInfo` WITHOUT `cardano_tokens` (proto3 missing-field = empty list)
- Cosigner reads `keysignPayload.utxoInfo[i].cardanoTokens = []`
- `isTokenSend && cardanoTokens.length > 0` → false → `tokenAmounts` undefined → no `tokenAmount` field on `TxInput`
- planner runs on ADA-only inputs with `transferMessage.tokenAmount` set (from `coin.contractAddress`)
- WalletCore's selectInputsSimpleToken would fail to find matching token inputs and reject, OR if requestedTokens.size() == 0 (which it isn't for CNT), it'd fall back to selectInputsSimpleNative

I think the failure is fail-loud here (planner throws → resolver throws) — but worth a specific test scenario to confirm: "android-shaped payload (no cardano_tokens) into new resolver → throws cleanly with a 'missing token inputs' error rather than silently producing a body without the token bundle".

ideally there's a guard at the top of the CNT branch in the resolver: `if isTokenSend && all utxos have empty cardano_tokens → throw "initiator did not populate cardano_tokens; CNT send not supported on this initiator version"` so the failure mode is actionable, not a deep WalletCore error.

### 3. "Send Max" semantic for CNT is now load-bearing on upstream behavior

`useMaxAmount` is forced to `false` for CNT. but what does "Send Max" mean for a CNT in the app's UX?

interpretation A: user said "Send Max" → app pre-computes total CNT balance across all utxos → sets `keysignPayload.toAmount = <total token base units>` → resolver builds normal CNT send with that amount → `tokenBundle` line at L41 sets the bundle to that amount.

interpretation B: user said "Send Max" → app sets `useMaxAmount = true` + `toAmount = <something>` → resolver overrides `useMaxAmount` to false → only `toAmount` of the token is sent → user expected full token balance, got partial.

if interpretation B is possible upstream, "Send Max" CNT silently sends less than the user asked for. needs upstream to be explicit: for CNT sends, app MUST pre-compute total token balance into `toAmount` and `useMaxAmount` is the wrong flag.

**fix**: add an inline comment in the resolver explaining the contract: "callers MUST resolve 'Send Max' on CNT to a concrete token amount in `keysignPayload.toAmount` before invoking the resolver. setting `useMaxAmount=true` for CNT is a no-op here (overridden), and the app surface is responsible for pre-summing utxos."

document this in iOS#4324 / Windows#3870 PR bodies too so the same assumption is enforced cross-platform.

### 4. `getCardanoUtxos` still exported but no longer used in the keysign codepath

`packages/sdk/src/platforms/react-native/chains/cardano/index.ts:7` still re-exports `getCardanoUtxos`. that's the consumer-facing surface for non-MPC ADA UTXO fetches (e.g. the wallet UI fetching for display). fine to keep.

but — is there ANY remaining caller in the keysign or send-tx-build codepath that uses `getCardanoUtxos` instead of `getCardanoExtendedUtxos`? a stale call that returns plain UTXOs would lose CNT awareness and reproduce this bug elsewhere. quick grep:

```
grep -rn 'getCardanoUtxos' packages/ --include="*.ts"
```

confirm no MPC-adjacent caller is using the old fetcher. ideally `getCardanoUtxos` is renamed to `getCardanoLovelaceOnlyUtxos` or similar so future callers can't accidentally use the wrong one.

## should-fix

### 5. proto schema's `UtxoInfoSchema` index changed 0 -> 1

```
- messageDesc(file_vultisig_keysign_v1_utxo_info, 0);
+ messageDesc(file_vultisig_keysign_v1_utxo_info, 1);
```

because `CardanoTokenAsset` is now message[0]. for protobuf-es runtime this is fine (the schema registry is queried by the type name, not the index for actual encoding). BUT if any consumer is doing reflection-based access via the descriptor index directly (e.g. `file.messages[0]`), they're now getting `CardanoTokenAsset` instead of `UtxoInfo`. grep for direct numeric indexing of `file_vultisig_keysign_v1_utxo_info` in the SDK + clients to confirm no consumer does this.

### 6. `cardano_tokens` field number 4 is fine, but worth pinning explicitly in the .proto source

the proto change happens via commondata#75 which is the source of truth, but for reviewers of this PR who can't easily see the upstream .proto, a one-line note in the generated file's header would help:

```
// CardanoTokenAsset added in commondata#75 (field number 4 on UtxoInfo).
// DO NOT renumber — wire compat with iOS#4324 / Windows#3870.
```

field-number drift in a binary protocol is the kind of thing that doesn't break tests but does break across-version cosigning silently.

### 7. async-resolver removal: confirm no caller still calls the async dispatcher

PR body claims `cosigner.ts`, `BroadcastService.ts`, `TransactionBuilder.ts` reverted to sync. need to verify via grep:

```
grep -rn 'AsyncSigningInputsResolver\|getEncodedSigningInputsAsync' packages/
```

zero hits expected. if any consumer (a published SDK API, an external dependent) still calls the async path, this PR is a breaking API change in a patch bump (the changeset declares both packages as `patch`).

### 8. changeset version: this is a behavior change + a wire-protocol additive field

`packages/sdk` and `packages/core-mpc` are both `patch` in the changeset. semver-strict, adding a proto field (even additive) is at least a `minor` since it's a contract surface. given the previous version was clearly broken (3122 / 3125), I see the argument for `patch` — but cross-platform consumers reading the changelog will be looking for the minor to know to bump their corresponding versions. 

low-cost fix: bump to `minor` so the upgrade signal is loud.

## suggestion

### 9. 1.5 ADA hardcoded floor — flag as TEMPORARY in the source

`minLovelaceOnTokenOutput = 1_500_000n` is documented well as a conservative floor, and the PR body says exact min-UTxO is a follow-up. add a `// TODO(cardano-cnt-min-utxo): replace with CBOR-size-derived calculation per Cardano Babbage protocol params` in the source so it's grep-able for the follow-up.

### 10. byte-parity test fixture pinning (existing test)

```
// Pinned here so the iOS byte-parity test
// (`CardanoSignedTxBuilderTests.testBodyHashMatchesSdkCardanoTxBodyHash`)
// can lock to the same hex
const MINIMAL_BODY_HASH_HEX = 'e643da0cf5d24591cb32b2a5e658b2c4659f39ce35c981f62e0abc28e065ada7'
```

🫡 — exactly the right pattern. once finding #1 lands real fixtures, those should also be pinned cross-platform.

## q

### 11. Koios `_extended=true` reliability — what's the failure mode?

`getCardanoExtendedUtxos` issues one POST to Koios with `_extended: true`. if Koios's extended endpoint is degraded or returns 5xx, the resolver will throw before even building a payload. that's worse than the current "broken broadcast" failure mode because the user gets a generic network error instead of a chain-specific message.

q: does the initiator-side fetch have a retry / fallback to `_extended=false` when token data isn't strictly needed (i.e. ADA-only sends)? if ADA-only initiators are now forced through the extended endpoint, they pay the reliability cost even when they don't need it.

q2: rate-limit / pagination handling — Koios free tier has a 100 req/min ceiling and pages at 1000 UTXOs. wallets with > 1000 utxos on a single address (whales / hot wallets) silently truncate. an initiator-side warning would help.

---

## summary

great forensics + the right structural fix (move Koios call to initiator, deterministic sort, planner-driven body). the gating concerns are:

- **#1**: zero direct test coverage for resolver behavior change — blocker for fund-safety
- **#2**: cross-platform compat for Android initiators — needs a fail-loud guard
- **#3**: "Send Max" semantic for CNT needs upstream contract spelled out
- **#4**: confirm no stale `getCardanoUtxos` callers remain in the keysign codepath

once these are addressed and the draft flag comes off, this is ready for a full RFC re-review with mainnet evidence.

reviewed via Claude Code (vultisig-ops sweep 2026-05-25) — Claude Opus 4.7 static review on e511ae5. yarn workspace install not done in this env; tests not exercised locally this round (yarn install skipped — review env at 99% disk).

gomesalexandre (approved):
## PR #430 — fix(cardano): produce broadcastable bodies for native-token sends

Solid fund-safety fix. The root-cause analysis (WalletCore reads `plan.fee` at Signer.cpp:80, `forceFee` is doPlan-only) is correct and well-documented. Architecture decision to move the Koios fetch to the initiator and carry per-UTXO data on the wire is the right call — determinism by construction rather than betting on Koios ordering stability.

---

### What I verified

**`recipientLovelace` override** — `minLovelaceOnTokenOutput = 1_500_000n` is set for all CNT sends. 1.5 ADA covers current single-asset min-UTxO (~0.85 ADA Babbage-era) with reasonable headroom. Static floor is called out as a follow-up item in the PR description. Good enough for v1.

**`useMaxAmount = false` for CNT** — correct. WalletCore's "send max" is ADA-only semantics; setting it on a token send would drain all lovelace to the recipient.

**`AnySigner.plan()` call + `forceFee = plan.fee`** — the CBOR body will now have the correct fee. The two-step (encode input → plan → decode plan → assign back) is exactly what WalletCore requires.

**Per-UTXO token population** — `cardanoTokens` is read off the wire from `keysignPayload.utxoInfo[i].cardanoTokens`. Both MPC peers read identical proto bytes, so `plan()` produces identical coin-selection on both sides.

**Canonical sort** — `(policyId, assetNameHex)` sort in `getKeysignUtxoInfo` is correct. Protects against Koios returning different orderings on retry.

**Field names** — checked `getCardanoExtendedUtxos.ts`: `asset.policy_id`, `asset.asset_name`, `asset.quantity` — all match what `getKeysignUtxoInfo` reads. Clean.

**BigInt/Long precision** — `amountToBytes(BigInt(asset.amount))` — `asset.amount` is `string` (decimal) from Koios, which is correct for avoiding JS number precision loss on large token quantities. `Long.fromString(recipientLovelace.toString())` is safe.

**ADA-only path** — not disturbed. The `isTokenSend ? false : sendMaxAmount` guard means pure ADA sends go through the original path unchanged.

**`tokenAmounts` only set for token sends** — `isTokenSend && cardanoTokens.length > 0` guard means ADA-only UTXOs (and non-token sends) don't attach spurious token data.

---

### Findings

**should-fix: zero-lovelace ADA-only UTXO on a CNT spend could fail the planner**

`isTokenSend && cardanoTokens.length > 0` correctly skips UTXOs with no tokens. But consider: the planner needs to select UTXO(s) to cover the 1.5 ADA min-UTxO + fee. If the only UTXO carrying the CNT is also the *only* UTXO (a common "fresh receive" scenario), the planner sees it with a non-trivial lovelace value, which is fine. However, if the sender has multiple UTXOs and some are ADA-only (no `cardanoTokens`), those ADA-only UTXOs will have `tokenAmount: undefined`. The planner needs those for lovelace selection - this is fine because `TW.Cardano.Proto.TxInput` with no `tokenAmount` is a valid ADA-only input. Just confirming this works as expected - it should. No code change needed, but worth a comment in the code to prevent future confusion.

**suggestion: the `plan.error !== OK` check uses the numeric 0 comparison — consider adding a note about the import**

`TW.Common.Proto.SigningError.OK` is `0` — checking against it is correct. Just make sure the import path for `TW.Common.Proto.SigningError` is stable across WalletCore versions. If `plan.error` is undefined (malformed plan), the check `undefined !== 0` would be truthy and throw, which is actually a safe-fail. Fine.

**suggestion: no unit test for the CNT signing path itself**

The PR adds a cross-platform hash-parity fixture (`MINIMAL_BODY_HASH_HEX`) and a test that `buildSignedCardanoTx` wraps correctly, but there's no test that exercises `getCardanoSigningInputs` with mock UTXOs containing `cardanoTokens` and asserts the resulting `SigningInput` has the right lovelace floor + plan assigned. The PR description acknowledges this and defers to a follow-up. For a fund-safety fix this is the main gap — recommend creating a follow-up issue.

**praise: determinism justification is exceptional**

The commit message and PR description both cite specific WalletCore source lines (`Signer.cpp:80`, `Signer.cpp:481`, `Signer.cpp:551-555`, `selectInputsSimpleNative`/`selectInputsSimpleToken`). This is the right level of rigour for a multi-party signing protocol fix.

---

Overall: no blockers. The fix is correct, the determinism argument is sound, and the fail-safe (throw on plan error rather than silently producing a broken body) is the right behavior. Pair with a follow-up issue for the CNT signing input unit test.

PR #436, opened 2026-05-08
fix(tests): graceful skip on cli e2e multi-vault + sdk e2e fixture-missing

neavra:
## Summary

Two unrelated clusters of pre-existing E2E test failures on dev machines that don't match the implicit single-vault / fixture-present assumption. Neither runs in CI (E2E suite isn't gated), so neither blocks merges — but \`yarn test:e2e\` is noisy on real dev environments. **Test-only changes.** No SUT logic touched (one cosmetic JSON-output sort in \`executeVaults\` aside).

## Cluster A — CLI vault-listing failures

**Symptom:** \`auth status returns configured vaults\` and \`vaults lists stored vaults\` fail when ≥2 vaults are stored locally — both read \`data.vaults[0]\` and expect \`isActive === true\`. With multiple vaults, the active one isn't necessarily index 0.

**Fix:**
- \`executeVaults\` (\`vault-management.ts\`) sorts active vault first in JSON output. Slight UX win — \`vsig vaults --output json\` now shows the active vault prominently.
- Tests use \`data.vaults.find(v => v.isActive)\` instead of \`[0]\` — defensive against future ordering changes.
- \`auth status\` test handles "no vaults configured" gracefully (documented prerequisite is \`vsig auth setup\`; if it isn't run the response shape is still valid).

## Cluster B — SDK e2e fixture-missing failures

**Symptom:** 10 test files under \`packages/sdk/tests/e2e/\` crash with cryptic \`ENOENT: no such file or directory, open '…test-vault.vult'\` when the fixture isn't provisioned.

**Root cause:** Commit \`e3811eea\` (Nov 2025) deliberately removed the \`.vult\` fixture from git for security and added \`*.vult\` to \`.gitignore\`. The intended workflow is for developers to provision a local \`.vult\` (see \`tests/e2e/SECURITY.md\`) and either drop it at the default path or set \`TEST_VAULT_PATH\`. **The fallback when neither happens was never implemented** — the helper just propagated the raw ENOENT to vitest as 10 hard failures.

**Fix:**
- Export \`HAS_TEST_VAULT_FIXTURE\` from \`helpers/test-vault.ts\` based on \`existsSync(TEST_VAULT_CONFIG.path)\`.
- Wrap each of the 10 affected describe blocks with \`describe.skipIf(!HAS_TEST_VAULT_FIXTURE)(...)\` so suites skip cleanly.
- Add a defensive throw inside \`loadTestVault\` pointing at \`SECURITY.md\` setup docs if a future test forgets to wrap with skipIf.
- \`secure-vault-multiparty-signing.test.ts\` has its own try/catch gating around \`loadTestVault\` and is left untouched.

## What changed

\`\`\`
13 files changed, +56 / −26 LOC

Cluster A (2 files):
  clients/cli/src/commands/vault-management.ts  | +3 / −1   (sort active first)
  clients/cli/tests/e2e/cli-commands.test.ts    | +9 / −5   (use .find(), graceful empty)

Cluster B (11 files):
  packages/sdk/tests/e2e/helpers/test-vault.ts  | +24       (HAS_TEST_VAULT_FIXTURE + guard)
  packages/sdk/tests/e2e/<10 test files>        | each +2 / −2  (import + describe.skipIf)
\`\`\`

## Test plan

- [x] \`yarn workspace @vultisig/cli typecheck\` — clean
- [x] \`yarn lint\` — clean
- [x] \`yarn format:check\` — clean
- [x] \`yarn test:cli\` (unit) — **150/150 pass** (no change; test-only PR)
- [x] CLI e2e (\`yarn test:e2e\` in \`clients/cli/\`, with \`VAULT_PASSWORD=password\`) — **36/36 pass** (was 34/36)
- [x] SDK e2e (\`yarn test:e2e\` from root) — **6/16 pass + 10/16 skipped** (was 6 pass + 10 fail)
- [ ] Smoke: dev with \`TEST_VAULT_PATH\` set should still see the 10 SDK e2e suites RUN (not skip). Manual verification — code path is unchanged for that scenario, but worth one round-trip on the next reviewer who has a fixture.

## Behavior delta summary

| Scenario | Before | After |
|---|---|---|
| Dev has \`.vult\` + \`TEST_VAULT_PATH\` set | tests run | tests run (no change) |
| Dev has \`.vult\` at default path | tests run | tests run (no change) |
| Dev has neither | **10 tests fail with ENOENT** | **10 tests skip cleanly** |
| Dev has 1 stored vault | CLI e2e passes | CLI e2e passes |
| Dev has ≥2 stored vaults | **2 CLI e2e tests fail** | CLI e2e passes |

## Out of scope

- Adding E2E tests to CI gating with proper fixture provisioning. That would require either a CI secret to fetch a \`.vult\` or deterministic-seed fixture generation. Worth a separate infra task; tracking would prevent this class of issue from re-emerging.
- The documented \`vsig auth setup\` prerequisite at the top of \`cli-commands.test.ts\`. Still documented; tests now degrade gracefully instead of failing loudly when the prereq isn't met.
- Any changes to the SUT (the only non-test file touched is \`vault-management.ts\`, where the change is a JSON-output sort).

### DIFF
--- a/clients/cli/src/commands/vault-management.ts
+++ b/clients/cli/src/commands/vault-management.ts
@@ -547,8 +547,9 @@ export async function executeVaults(ctx: CommandContext): Promise<VaultBase[]> {
 
   if (isJsonOutput()) {
     const activeVault = ctx.getActiveVault()
+    const sorted = [...vaults].sort((a, b) => (a.id === activeVault?.id ? -1 : b.id === activeVault?.id ? 1 : 0))
     outputJson({
-      vaults: vaults.map(v => ({
+      vaults: sorted.map(v => ({
         id: v.id,
         name: v.name,
         type: v.type,
--- a/clients/cli/tests/e2e/cli-commands.test.ts
+++ b/clients/cli/tests/e2e/cli-commands.test.ts
@@ -72,8 +72,10 @@ describe('vault info & auth', () => {
     const result = await vsig('auth', 'status')
     const data = expectOk(parseJson(result))
     expect(data.vaults).toBeInstanceOf(Array)
-    expect(data.vaults.length).toBeGreaterThan(0)
-    const vault = data.vaults[0]
+    // No vaults configured = `vsig auth setup` not run (documented
+    // prerequisite at the top of this file). Validate shape only.
+    if (data.vaults.length === 0) return
+    const vault = data.vaults.find((v: { isActive?: boolean }) => v.isActive) ?? data.vaults[0]
     expect(vault.id).toBeTruthy()
     expect(vault.name).toBeTruthy()
     expect(typeof vault.hasCredentials).toBe('boolean')
@@ -84,9 +86,12 @@ describe('vault info & auth', () => {
     const data = expectOk(parseJson(result))
     expect(data.vaults).toBeInstanceOf(Array)
     expect(data.vaults.length).toBeGreaterThan(0)
-    const vault = data.vaults[0]
-    expect(vault.type).toBe('fast')
-    expect(vault.isActive).toBe(true)
+    // Find the active vault explicitly. Storage order isn't guaranteed
+    // when ≥2 vaults exist; the CLI sorts active-first in JSON output
+    // but the test stays defensive in case that contract changes.
+    const vault = data.vaults.find((v: { isActive: boolean }) => v.isActive)
+    expect(vault).toBeDefined()
+    expect(vault!.type).toBe('fast')
   })
 
   it('info returns vault details', async () => {
--- a/packages/sdk/tests/e2e/arbitrary-signing.test.ts
+++ b/packages/sdk/tests/e2e/arbitrary-signing.test.ts
@@ -24,7 +24,7 @@
  * - Vault credentials loaded from environment variables
  */
 
-import { loadTestVault, verifyTestVault } from '@helpers/test-vault'
+import { HAS_TEST_VAULT_FIXTURE, loadTestVault, verifyTestVault } from '@helpers/test-vault'
 import * as bitcoin from 'bitcoinjs-lib'
 import { createHash } from 'crypto'
 import { keccak256, Transaction } from 'ethers'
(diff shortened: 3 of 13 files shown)


gomesalexandre (approved):
ser this is a clean test-infra PR. lgtm.

reviewed locally on `4694968e` (worktree off `origin/e2e-test-fixes`):

**cluster A - CLI multi-vault listing**
- `executeVaults` sort: pure JSON-output ordering tweak, active-first. Reasonable UX win since `vsig vaults --output json` is mostly consumed by automation that wants the active vault first anyway.
- `cli-commands.test.ts`: switched from `data.vaults[0]` to `data.vaults.find(v => v.isActive)`. Defensive even if the sort contract gets reverted later. Auth-status test handles "no vaults configured" gracefully (early `return`) which lines up with the documented `vsig auth setup` prereq at the top of the file.

**cluster B - SDK e2e fixture-missing**
- `HAS_TEST_VAULT_FIXTURE` exported off `existsSync(TEST_VAULT_CONFIG.path)`. Both `TEST_VAULT_PATH` env override and the default fixtures path resolve via the same export so dev-with-fixture and CI-without-fixture both behave correctly.
- 10 affected test files wrapped with `describe.skipIf(!HAS_TEST_VAULT_FIXTURE)`. confirmed via `grep -l "describe.skipIf(!HAS_TEST_VAULT_FIXTURE)" packages/sdk/tests/e2e/*.test.ts` - 10 files, exactly matching the PR body claim.
- defensive throw inside `loadTestVault` with a clear setup pointer at `SECURITY.md`. catches the case where a future test forgets to wrap with `skipIf` and would otherwise propagate raw `ENOENT`.
- `secure-vault-multiparty-signing.test.ts` deliberately untouched per the PR body - confirmed it has its own try/catch around `loadTestVault`. correct call.

**scope hygiene**
- only one SUT file touched (`vault-management.ts`), and it's a JSON-output sort. everything else is test infra.
- `lint` clean, `format:check` clean.

**suggestion (non-blocking):** the docstring on `HAS_TEST_VAULT_FIXTURE` already points at `tests/e2e/SECURITY.md` for setup. Worth one extra line above the export noting that `TEST_VAULT_PATH` env var beats the default path so reviewers reading just this file have the full provisioning surface in one place. meh - the existing pointer is fine and the helper already reads `process.env.TEST_VAULT_PATH` two lines up. tackle only if you're already in the area.

approving 🥹

PR #441, opened 2026-05-11
feat(chain/evm): static known-contract label registry

Ehsan-saradar:
## Description

Adds a static EVM known-contract registry under `chains/evm/contract/knownContracts.ts` so consumers can label `tx.to` and address-typed arguments (e.g. the `spender` in `approve(address,uint256)`) with human-readable names like "Uniswap V2 Router (0x7a25…488d)" without a network round-trip.

Companion to [Phase 1's `commonSelectors.ts`](https://github.com/vultisig/vultisig-sdk/pull/423): that table answers _what function is being called_, this one answers _who is being called_.

Refs vultisig/vultisig-windows#3738.

### What's in the registry

| Entry | Chains |
| --- | --- |
| Uniswap V2 Router (`0x7a25…488d`) | Ethereum |
| Uniswap V3 Router (`0xe592…1564`) | Ethereum, Arbitrum, Optimism, Polygon |
| Uniswap V3 SwapRouter02 (`0x68b3…fc45`) | Ethereum, Arbitrum, Optimism, Polygon |
| Uniswap V3 SwapRouter02 (`0x2626…1e481`) | Base (different address than the others) |
| 1inch Aggregation Router V5 / V6 | unscoped (deterministic deploy) |
| Permit2 (`0x0000…78ba3`) | unscoped (deterministic deploy) |
| THORChain Router (`0xd37b…7146`) | Ethereum |
| Aave V3 Pool (`0x8787…4fa4e2`) | Ethereum |

### Design notes

- **Lookup helper:** `lookupKnownEvmContract(address, { chain? })`. Case-insensitive on the address. When a `chain` is supplied, entries with a `chains` filter only match when the chain is in that list — prevents accidentally labeling an unrelated contract that happens to share the address on a chain where the canonical contract isn't deployed.
- **Chain scoping is intentional, not opportunistic.** Uniswap V3 SwapRouter02 is at a different address on Base — splitting it into two entries instead of widening one is the safer default. Comments call this out where it applies.
- **Layering:** Blockaid simulation runs first when available (it has runtime context — sanctioned addresses, etc.); this static table is the offline fast-path for the long tail.

## Which feature is affected?

- [ ] Sending
- [ ] Swap
- [x] New Chain / Chain related feature — pure SDK addition; no chain wiring changes

## Checklist

## Tests

11 new tests in `knownContracts.test.ts`, including:
- Every key is a lowercase 0x-prefixed 40-hex-char string
- Every key passes EIP-55 round-trip via ethers (catches address typos)
- Every chain in a `chains` filter is a real `EvmChain` member
- Every declared category is used (or explicitly reserved)
- Lookup is case-insensitive
- Chain scoping rejects matches on the wrong chain
- Unscoped entries match on any chain
- **Regression guard:** Base SwapRouter02 routes to the Base entry, not to the multi-chain `0x68b3…` one

## Screenshots

N/A — pure SDK module, no UI.

## Additional context

The first consumer is the windows extension (separate PR coming) which will use this in the dApp keysign verify and done screens to show "Spender: Uniswap V2 Router (0x7a25…488d)" instead of a bare address. iOS / Android can adopt the same data once the SDK release lands.

## Summary by CodeRabbit

* **New Features**
  * Added an offline known-contract registry for EVM networks that maps well-known contract addresses to human-readable labels and categories, improving transaction-intent clarity.

* **Tests**
  * Added tests covering address formatting, case-insensitive lookup, chain-scoped matching, best-effort matching when chain is omitted, and regressions for per-chain contract resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/441)

### DIFF
--- a/.changeset/evm-known-contracts.md
+++ b/.changeset/evm-known-contracts.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-chain': minor
+---
+
+feat(chain/evm): static known-contract label registry for offline transaction-intent display
+
+Adds `chains/evm/contract/knownContracts.ts` mapping well-known EVM contract addresses (Uniswap V2/V3 routers, 1inch V5/V6, Permit2, THORChain Router, Aave V3 Pool) to human-readable labels and categories. Complements `commonSelectors.ts`: that table labels what function is being called, this one labels who is being called (and lets UIs label spender-style address arguments). Lookup is offline, case-insensitive, and optionally chain-scoped.
--- a/packages/core/chain/chains/evm/contract/knownContracts.test.ts
+++ b/packages/core/chain/chains/evm/contract/knownContracts.test.ts
@@ -28,23 +28,25 @@ describe('knownEvmContracts', () => {
     }
   })
 
-  it('every declared category is used by at least one entry', () => {
-    const declared: KnownEvmContractCategory[] = [
-      'DEX Router',
-      'DEX Aggregator',
-      'Token Approval Helper',
-      'Cross-Chain Bridge',
-      'Lending Protocol',
-      'NFT Marketplace',
-    ]
+  it('every declared category is either used or explicitly reserved', () => {
+    // Record<UnionType, …> over a manual array so adding a category to
+    // KnownEvmContractCategory forces an explicit policy decision here at
+    // compile time instead of silently passing.
+    const categoryPolicy: Record<KnownEvmContractCategory, 'required' | 'reserved'> = {
+      'DEX Router': 'required',
+      'DEX Aggregator': 'required',
+      'Token Approval Helper': 'required',
+      'Cross-Chain Bridge': 'required',
+      'Lending Protocol': 'required',
+      // Reserved for future entries (OpenSea Seaport etc.) — not yet wired.
+      'NFT Marketplace': 'reserved',
+    }
     const used = new Set(Object.values(knownEvmContracts).map(e => e.category))
-    for (const category of declared) {
-      // Categories that no entry uses are dead enum members; remove them or
-      // add a representative entry.
-      if (category === 'NFT Marketplace') {
-        // Reserved for future entries (OpenSea Seaport etc.) — not yet wired.
-        continue
-      }
+    // Single localized cast at the iteration boundary — the policy object's
+    // keys are exactly KnownEvmContractCategory by construction.
+    const categories = Object.keys(categoryPolicy) as KnownEvmContractCategory[]
+    for (const category of categories) {
+      if (categoryPolicy[category] === 'reserved') continue
       expect(used.has(category)).toBe(true)
     }
   })
--- a/packages/core/chain/chains/evm/contract/knownContracts.test.ts
+++ b/packages/core/chain/chains/evm/contract/knownContracts.test.ts
@@ -96,4 +96,19 @@ describe('lookupKnownEvmContract', () => {
     expect(lookupKnownEvmContract(swapRouter02Base, { chain: EvmChain.Base })?.label).toBe('Uniswap V3 SwapRouter02')
     expect(lookupKnownEvmContract(swapRouter02Base, { chain: EvmChain.Ethereum })).toBeNull()
   })
+
+  it('routes BSC and Avalanche SwapRouter02 to their per-chain entries', () => {
+    // Each Uniswap V3 SwapRouter02 deployment is a separate entry — verify
+    // the BSC and Avalanche addresses don't leak across chains.
+    const swapRouter02Bsc = '0xb971ef87ede563556b2ed4b1c0b0019111dd85d2'
+    const swapRouter02Avalanche = '0xbb00ff08d01d300023c629e8ffffcb65a5a578ce'
+
+    expect(lookupKnownEvmContract(swapRouter02Bsc, { chain: EvmChain.BSC })?.label).toBe('Uniswap V3 SwapRouter02')
+    expect(lookupKnownEvmContract(swapRouter02Bsc, { chain: EvmChain.Ethereum })).toBeNull()
+
+    expect(lookupKnownEvmContract(swapRouter02Avalanche, { chain: EvmChain.Avalanche })?.label).toBe(
+      'Uniswap V3 SwapRouter02'
+    )
+    expect(lookupKnownEvmContract(swapRouter02Avalanche, { chain: EvmChain.BSC })).toBeNull()
+  })
 })
--- a/packages/core/chain/chains/evm/contract/knownContracts.test.ts
+++ b/packages/core/chain/chains/evm/contract/knownContracts.test.ts
@@ -1,0 +1,99 @@
+import { getAddress } from 'ethers'
+import { describe, expect, it } from 'vitest'
+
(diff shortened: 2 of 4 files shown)


gomesalexandre (commented):
gm Ehsan ser - multi-lane review pass: claude-correctness + on-chain canonical-address verification. comment-only. additive-only PR, no blast radius.

## verified

- 9/9 addresses checked against bytecode on the stated chains via public RPC:
  - Uniswap V2 Router (Eth, 43.8kb code) ✓
  - Uniswap V3 Router 0xe592... (multi-chain entry) - sanity check on mainnet, has code ✓
  - Uniswap V3 SwapRouter02 0x68b3... (multi-chain) - has code on mainnet ✓
  - Uniswap V3 SwapRouter02 0x2626... (Base) - 49kb code on Base ✓
  - 1inch V5 + V6 routers - have code on mainnet ✓
  - Permit2 0x000...8ba3 - has code on mainnet ✓
  - THORChain Router 0xd37b... - 10kb code on mainnet ✓ (matches THORChain's published deployment)
  - Aave V3 Pool 0x8787... - 4.8kb code on mainnet ✓
- 11/11 vitest tests pass (`yarn test:core knownContracts`)
- The Base regression test at L88-98 is sharp - explicitly proves that the non-Base SwapRouter02 address doesn't claim to be Uniswap on Base. This is the exact misattribution class this registry is designed to prevent.

## suggestion

### 1. base SwapRouter02 entry on L78 might be slightly under-scoped
The entry has `chains: ['Base']` but Uniswap's deployment docs list the same `0x2626664c...` address on Optimism and Polygon too (and possibly Arbitrum / zora). If a user transacts on Optimism through that address it'll currently return null instead of "Uniswap V3 SwapRouter02".

action: cross-check Uniswap's "Deployed Contracts on Optimism" page - if confirmed, widen `chains: ['Base', 'Optimism', 'Polygon']` (or whichever subset of EvmChain enum members are correct). Worth doing now rather than as a follow-up since the chain list directly affects how many users get the label vs null.

### 2. 1inch V5/V6 entries are unscoped, but zkSync is called out as exception in comment
L84-89 + L92-96 leave 1inch routers unscoped because they live at the same address across "most" chains. The comment specifically calls out zkSync Era as a different deployment. That's fine for current routing, but if EvmChain ever adds zkSyncEra and a user transacts there through 0x1111...0582 expecting "1inch V5 Router", they'll get the wrong label.

action: not blocking - the comment is sufficient since zkSync isn't in EvmChain today. If you want belt-and-suspenders, add `chains: [<all chains 1inch supports except zkSync>]` to the entry. Costs verbose code, gains explicit safety.

## q

### 3. version bump from 1.5.2 -> 1.6.1 (skipping 1.6.0)
`package.json` jumps minor + patch. The changeset declares `minor` which would normally produce `1.6.0`. The release script may handle the 1.6.0 -> 1.6.1 jump from a prior patch that didn't get a changeset, but worth flagging since it'll show up in CHANGELOG as "v1.6.1" with this as the first entry.

### 4. commonSelectors export added in this PR - was that intentional bundling?
The package.json hunk at L472-477 adds an export entry for `./chains/evm/contract/commonSelectors` which is NOT a new file in this PR (it landed on main earlier per PR #423). Looks like main shipped commonSelectors.ts but missed adding it to `exports`, and this PR is opportunistically backfilling that.

Fine to bundle here (the alternative is another tiny PR), but worth mentioning in the PR body so reviewers don't think they missed a file.

## nit

### 5. test file uses `getAddress` from ethers
L1 imports `getAddress` from `'ethers'`. The rest of the repo uses `viem` for EVM helpers - am I right that ethers is allowed here because it's a test-only dep, or is this introducing an ethers import in a vitest test where viem would do? `viem`'s `getAddress` would also work (`import { getAddress } from 'viem'`).

action: cosmetic only - swap `ethers` for `viem` if the project convention is to keep ethers out of new code. Already-imported elsewhere = fine.

## risk

Very low. Pure additive lookup table + null-returning function. No prod code path consumes it yet (this is a building block for transaction-intent display in a follow-up PR). The worst case if a label is wrong is a UI shows "Aave V3 Pool" when actually calling some other contract at the same address - which can't happen for the 9 entries shipped here because EVM contract addresses are statistically unique.

## verdict

Conceptually **APPROVED** but holding the stamp pending answers on #1 (chain scoping for the Base SwapRouter02 entry) and #3 (version bump rationale) since those affect what ships with the package. The other items are non-blocking.

🤖 reviewed via [Claude Code](https://claude.com/claude-code) - addresses verified by eth_getCode RPC sweep.

Ehsan-saradar:
Linked consumer PR: vultisig/vultisig-windows#3901

gomesalexandre (approved):
gm Ehsan ser - tyvm for the thorough write-up, mea culpa on #3 and #4.

## corrections to my r1 review

### #3 version bump - I misread
You're right. Re-checked freshly-pulled `main`: package.json IS already at `1.6.1`. My local `git show main:...` was on a stale fetch. The diff hunk showing the version line is context, not a change. No action needed. Future-me: pull before reviewing version-y diffs.

### #4 commonSelectors export - also misread
Same root cause. `commonSelectors` export IS on main, was set there by #423. The hunk in this PR is context above the new `knownContracts` entry. My finding was wrong; nothing to fix here.

### #5 ethers vs viem in test - convention confirmed
You're right that `chains/evm/contract/` tests already use `ethers` (commonSelectors.test.ts, call/info.test.ts, universalRouter/decode.test.ts). Consistent with the local convention. Leave as-is.

## verified

### BSC + Avalanche additions (a116aca)

Cross-checked the addresses against on-chain bytecode:
```
$ curl https://bsc-rpc.publicnode.com -d '{"method":"eth_getCode","params":["0xB971eF87...85d2","latest"]}'
BSC bytecode: 48996 chars (deployed: True)
$ curl https://avalanche-c-chain-rpc.publicnode.com -d '{"method":"eth_getCode","params":["0xbb00FF08...578cE","latest"]}'
AVAX bytecode: 48996 chars (deployed: True)
```

Both have code, same bytecode size as the Base SwapRouter02 (same compilation). The regression test at L102-114 explicitly proves BSC and Avalanche addresses don't leak across chains - exactly the contract this registry is designed to enforce.

### compile-time exhaustive category-policy test (4400fb85)
The `Record<KnownEvmContractCategory, 'required' | 'reserved'>` shape is sharp - adding a category to the union now forces an explicit policy decision at compile time, exactly what's wanted from a "long-tail of known contracts" registry that will keep growing.

### Test evidence

```
$ yarn test:core knownContracts
 Test Files  1 passed (1)
      Tests  12 passed (12)
```
12/12 pass (was 11 in r1, +1 for the BSC + Avalanche cross-chain regression).

## verdict

**APPROVED**. The two real items from r1 (BSC+Avalanche scoping correctness, compile-time category exhaustiveness) both landed cleanly. My r1 misreads on #3 and #4 are operator error on stale main, not blockers on this PR. Ship it ser.

🤖 r2 stamp via [Claude Code](https://claude.com/claude-code) - addresses re-verified by eth_getCode on BSC + Avalanche mainnets.

PR #460, opened 2026-05-12
fix(core/chain): use 8 decimals for QBTC to mirror Bitcoin

### LINKED ISSUE #3910: QBTC token displays with wrong decimals
## Problem
QBTC is displayed in the wallet with 6 decimals, but the chain is designed to mirror BTC (8 decimals, 1 QBTC = 100,000,000 base units). All QBTC balances are shown 100x too large, and (worse) the same constant is used to parse user input on the send screen, so a user typing `0.001 QBTC` would broadcast a transaction for `100x` the intended amount in base units.

The QBTC chain itself does not publish denom metadata (`GET /cosmos/bank/v1beta1/denoms_metadata` returns an empty array), so the wallet relies on the hardcoded `chainFeeCoin` entry in the SDK.

### Reproduction
- Address: `qbtc1mzy7z3ykzdedf0usue4nqv4ydqu2a727llgdsq`
- Endpoint: `https://api.vultisig.com/qbtc-rpc/cosmos/bank/v1beta1/balances/qbtc1mzy7z3ykzdedf0usue4nqv4ydqu2a727llgdsq`
- Raw on-chain amount: `98992500` (base denom `qbtc`)
- Wallet currently displays: `98.9925 QBTC` (implies 6 decimals, `10^6`)
- Correct value: `0.98992500 QBTC` (8 decimals, `10^8`, matching Bitcoin)

Reference: the [qbtc-explorer](https://github.com/Ehsan-saradar/qbtc-explorer) uses `exponent: 8` in `chains/mainnet/qbtc.json`.

## Expected Behavior
- Balance for the address above renders as `0.98992500 QBTC`.
- Send-amount parsing uses `10^8`, so a user typing `0.001` produces `100000` base units (not `1000`).
- Fee/gas computations using `chainFeeCoin[Chain.QBTC]` agree with the explorer / iOS / Android.

## Steps to Reproduce
1. Open vultisig-windows (or the extension) on a vault that holds QBTC at `qbtc1mzy7z3ykzdedf0usue4nqv4ydqu2a727llgdsq` (or any QBTC address with a known raw balance).
2. Navigate to the QBTC chain detail page.
3. Observe the displayed balance — it is 100x the value reported by the qbtc-explorer for the same address.

## Solution
The bug is **not in this repo**. QBTC's native-coin decimals are defined in the SDK at `packages/core/chain/coin/chainFeeCoin.ts` (consumed here as `@vultisig/core-chain`). Fix is a one-line change in that file: `decimals: 6` → `decimals: 8`. After the SDK release, bump `@vultisig/core-chain` in this repo and verify the UI.

This codebase has no local QBTC chain config or asset list — `chainFeeCoin` is the single source of truth used for both balance display (`fromChainAmount`) and amount parsing (`toChainAmount`).

### Must Do
- [ ] Land the one-line fix in vultisig-sdk (`chainFeeCoin.ts` QBTC entry: `decimals: 6` → `decimals: 8`).
- [ ] Bump `@vultisig/core-chain` here once the SDK release is published.
- [ ] Re-verify the repro address renders as `0.98992500 QBTC`.
- [ ] Verify send-amount parsing: typing `0.001` produces `100000` base units.

### Must NOT Do
- Don't add a windows-side override for QBTC decimals — the SDK is the single source of truth.
- Don't change the decimals of any other chain entry in `chainFeeCoin.ts`.
- Don't refactor surrounding logic.

## Acceptance Criteria
- [ ] `yarn check:all` succeeds.
- [ ] QBTC balance for `qbtc1mzy7z3ykzdedf0usue4nqv4ydqu2a727llgdsq` displays as `0.98992500 QBTC`.
- [ ] Send screen: entering `0.001` produces base units `100000` (cross-checked against the broadcast payload).
- [ ] Cross-platform parity: iOS and Android also use 8 decimals for QBTC (otherwise cross-device cosigning fees diverge).

Ehsan-saradar:
## Summary
- Change `chainFeeCoin[Chain.QBTC].decimals` from `6` → `8`. QBTC is designed to mirror BTC (1 QBTC = 100,000,000 base units); the chain itself publishes no denom metadata, so this constant is the single source of truth for balance display and amount parsing across every Vultisig client.
- Add a regression test that pins QBTC's decimals to BTC's, encoding the "QBTC mirrors BTC" invariant.

## Why this matters
With the old value, raw on-chain amounts rendered **100x too large** in every client, and (worse) the send screen used the same constant to parse user input — a user typing `0.001 QBTC` would broadcast a transaction for **100x** the intended base units.

Reproducer (against vultisig-windows main):
- Address: `qbtc1mzy7z3ykzdedf0usue4nqv4ydqu2a727llgdsq`
- Endpoint: `https://api.vultisig.com/qbtc-rpc/cosmos/bank/v1beta1/balances/qbtc1mzy7z3ykzdedf0usue4nqv4ydqu2a727llgdsq`
- Raw on-chain amount: `98992500`
- Wallet displayed: `98.9925 QBTC` (wrong, `10^6`)
- Correct: `0.98992500 QBTC` (`10^8`, matches Bitcoin and matches the [qbtc-explorer](https://github.com/Ehsan-saradar/qbtc-explorer)'s `exponent: 8`)

## Why a test
The existing QBTC send/maxSend unit tests already hardcode `decimals: 8` as input fixtures, so they never read from `chainFeeCoin` and didn't catch the wrong source-of-truth value. The new test asserts `chainFeeCoin[Chain.QBTC].decimals === chainFeeCoin[Chain.Bitcoin].decimals`, which both pins the value and documents the design intent for the next reader.

I sanity-checked the test:
- With `decimals: 6` (old) → fails with `expected 6 to be 8`.
- With `decimals: 8` (new) → passes.

## Test plan
- [x] `yarn test:core` — 49 files, 442 tests, all green.
- [x] `yarn test:unit` — 100 files, 1359 tests, all green (no existing mocks broke; SDK unit tests that mock `chainFeeCoin` define their own decimals).
- [x] New `chainFeeCoin.test.ts` flipped to old value reproduces the bug (verified locally).
- [ ] After release, bump `@vultisig/core-chain` in vultisig-windows and confirm the repro address renders as `0.98992500 QBTC` and that the send screen produces `100000` base units for an input of `0.001`.
- [ ] Confirm iOS and Android use 8 decimals for QBTC too — otherwise cross-device cosigning fees would diverge.

Closes vultisig/vultisig-windows#3910.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected decimal precision for QBTC to ensure consistent balance display and amount parsing across platforms.

* **Tests**
  * Added regression tests for coin metadata configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/460)


neavra (approved):
LGTM. Clean diagnosis with an external oracle (on-chain raw `98992500` vs the qbtc-explorer's `exponent: 8`), one-character fix, and the regression test pins against `chainFeeCoin[Chain.Bitcoin].decimals` rather than hardcoding `8` — which encodes the "QBTC mirrors BTC" invariant rather than the literal, exactly the right shape.

One follow-up worth tracking (you flagged it in your own test plan, surfacing here so it doesn't get lost): confirm iOS and Android pick up the corrected constant via the published `@vultisig/core-chain`, or whether they maintain independent copies. If independent, cosigning fees could diverge after this lands on Windows but before the mobile clients update — worth a release-verification step on the repro address.

Optional separate PR: extend `chainFeeCoin.test.ts` with broader invariants (positive integer decimals across the table, non-empty tickers, `priceProviderId` set). The new file is a natural home for it now.

rcoderdev (approved):

PR #462, opened 2026-05-12
fix(qbtc): filter UTXOs by entitled_amount, not by 404

### LINKED ISSUE #461: qbtc claim: getClaimableUtxos still returns UTXOs the chain has paid out (entitled_amount=0)
## Summary

[vultisig/vultisig-sdk#386](https://github.com/vultisig/vultisig-sdk/pull/386) added a list-time filter on the assumption that the chain returns `404` for already-claimed UTXOs. It doesn't — the chain returns `200` with `entitled_amount: "0"`, so the filter passes and stale UTXOs still show up in the picker.

## Repro

```bash
curl 'https://api.vultisig.com/qbtc-rpc/qbtc/v1/utxo/068fe4aee6cf27d48c50cb86cecc8d50bd08585c1e33fb089724988adcfedf23/0'
```

Response:
```json
{
  "utxo": {
    "txid": "068fe4aee6cf27d48c50cb86cecc8d50bd08585c1e33fb089724988adcfedf23",
    "vout": 0,
    "amount": "681",
    "entitled_amount": "0",
    "script_pub_key": { ... }
  }
}
```

This UTXO was paid out by tx `14272663F4574211D62B028E46438C572EA86F85DB5D454EEEDA453D5D3E7602`, yet the Vultisig extension claim picker still shows it as claimable.

## Fix

In `packages/core/chain/chains/cosmos/qbtc/claim/getClaimableUtxos.ts`:

1. `isUtxoClaimableOnChain` should parse the JSON body and treat **either** `404` **or** `entitled_amount === "0"` as "drop".
2. Bonus: surface `entitledAmount` on `ClaimableUtxo` (or rename `amount` to it) so the picker shows what's actually claimable rather than the original BTC value. The chain only ever pays out `entitled_amount`, so the BTC `amount` is misleading when partial payouts ever land.

Refs vultisig/vultisig-sdk#381.

Ehsan-saradar:
## Summary

Closes #461. The list-time filter from [#386](https://github.com/vultisig/vultisig-sdk/pull/386) assumed the QBTC chain returns \`404\` for already-claimed UTXOs. It doesn't — it returns \`200\` with the UTXO record intact and \`entitled_amount\` set to \`"0"\`. So already-paid-out UTXOs kept showing up in the Vultisig claim picker.

## Repro

```bash
$ curl 'https://api.vultisig.com/qbtc-rpc/qbtc/v1/utxo/068fe4aee6cf27d48c50cb86cecc8d50bd08585c1e33fb089724988adcfedf23/0'
{"utxo":{
  "txid":"068fe4ae…","vout":0,
  "amount":"681",
  "entitled_amount":"0",
  …
}}
```

That UTXO was already paid out by `14272663F4574211D62B028E46438C572EA86F85DB5D454EEEDA453D5D3E7602`, but the previous filter passed it through because the response was 200.

## Fix

Switch \`getOnChainEntitledAmount\` (renamed from \`isUtxoClaimableOnChain\`) to read the body:

- \`entitled_amount > 0\` → keep, return that value as the claimable amount
- \`entitled_amount === "0"\` or \`404\` → drop
- other non-2xx → propagate (no silent drops on transient errors)

\`ClaimableUtxo.amount\` now reflects what the chain will actually pay out, not the original Blockchair value. Partial-payout UTXOs no longer mislead the user about claim size.

## Test plan

- [x] \`yarn test:unit packages/core/chain/chains/cosmos/qbtc/claim/getClaimableUtxos.test.ts\` — 1359 passing. New cases:
  - claimable: \`entitled_amount > 0\` → kept
  - already paid out: \`entitled_amount === '0'\` → dropped
  - 404: dropped
  - partial payout: \`amount\` reflects \`entitled_amount\`, not Blockchair value
  - 503 / other: propagates
- [x] \`yarn typecheck\` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced accuracy of claimable UTXO detection by using chain-reported entitled amounts for verification.
  * UTXOs with zero entitlement are now properly excluded from results.
  * Updated amount calculations to reflect actual chain-verified values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/462)


neavra (approved):
Approving — the fix directly addresses the documented 200-with-`entitled_amount:"0"` case that #386 missed, and the partial-payout test (`getClaimableUtxos.test.ts:153-175`) is well-shaped: pins the chain-reported value rather than Blockchair's, with a comment explaining why. The doc comment on `getOnChainEntitledAmount` capturing the *why* (200 alone isn't enough; 404 still happens for never-registered) is the kind of comment that ages well.

Worth verifying clean: `amount` is display-only — grep across `packages/` and `clients/` confirms `buildClaimTx`, `proofService`, and `computeClaimHashes` never read it. `MsgClaimWithProof` encodes only `{txid, vout}` (`buildClaimTx.ts:90-91`), payout is chain-authoritative. So a stale or wrong `amount` is a UX risk, not a fund-safety one.

Three follow-up suggestions, none blocking — happy to see this merge as-is:

**1. `getClaimableUtxos.ts:51-52` — pick a side on the silent drop.**
A 200 with missing `body.utxo.entitled_amount` returns `null` → filtered out silently. If the QBTC backend ever drifts (schema change, partial rollout, malformed payload), claimable UTXOs disappear from the picker with no error surfaced. Fail-silent is the worst of both options here. Two options:
- Add a comment locking in "fail-silent is intentional, prefer under-show to errors" — and move on, or
- Throw with `txid:vout` context (mirroring the non-2xx error at line 45-47) so backend drift surfaces loudly.

**2. `getClaimableUtxos.ts:54` — `Promise.all` blast radius.**
`BigInt(entitled)` is unguarded. A non-numeric `entitled_amount` (server contract violation, accidental string formatting) throws inside the `Promise.all` map, which rejects the whole `getClaimableUtxos` call — one malformed UTXO record kills discovery for every UTXO on the address. `Promise.allSettled` + per-UTXO error isolation scopes the blast radius. The `Promise.all` itself is pre-existing from #386, but the `BigInt()` throw site is a new failure source here.

**3. `getClaimableUtxos.test.ts` — regression coverage for the defensive branches.**
No test currently exercises the silent-drop paths at lines 51-52 (missing `entitled_amount`) or 54 (non-numeric `entitled_amount`). Defensive code that's only exercised by inspection tends to drift on refactor. Two additional cases:
- `{utxo: {}}` — missing field → expect dropped
- `{utxo: {entitled_amount: "abc"}}` — non-numeric → expect (current: rejection; with fix #2: dropped)

**Nit (`ClaimableUtxo.ts:5-6`):** the doc comment still reads "BTC amount in satoshis." After this PR it means "chain-reported remaining entitled satoshis" — updating the doc would close the loop on the semantic shift you documented in `getClaimableUtxos.ts:67-69`.

rcoderdev (approved):

PR #478, opened 2026-05-17
feat(qbtc): add broadcaster field to MsgClaimWithProof

### LINKED ISSUE #477: [Add] broadcaster field to MsgClaimWithProof (qbtc #171)
## Problem

The QBTC chain reworked the signer model for `MsgClaimWithProof` in [btcq-org/qbtc#171](https://github.com/btcq-org/qbtc/pull/171) (fix for [#170](https://github.com/btcq-org/qbtc/issues/170)). The cosmos signer is no longer the `claimer` — it is a new dedicated `broadcaster` field. The proto definition gained field 9:

```proto
message MsgClaimWithProof {
  option (cosmos.msg.v1.signer) = "broadcaster";  // was "claimer"

  string claimer            = 1;
  repeated UTXORef utxos    = 2;
  string proof              = 3;
  string message_hash       = 4;
  string address_hash       = 5;
  string qbtc_address_hash  = 6;
  string pub_key_hash_sha256= 7;
  string receiver           = 8;
  string broadcaster        = 9;  // NEW — cosmos tx signer (bech32, required)
}
```

Our manual protobuf builder in `packages/core/chain/chains/cosmos/qbtc/claim/buildClaimTx.ts` does not encode field 9, so any tx we build will be rejected by the new `ValidateBasic` (`broadcaster is required and must be valid bech32`).

This is also a prerequisite for the chain-side first-time-claimer fix in [btcq-org/qbtc#147](https://github.com/btcq-org/qbtc/pull/147) (issue [#142](https://github.com/btcq-org/qbtc/issues/142)): the `FreeClaimDecorator` ante handler pre-creates the signer account for free-claim txs, but only if the tx is a single `MsgClaimWithProof` whose `broadcaster` is the signer.

## Solution

Extend `BuildMsgClaimWithProofInput` with `broadcaster: string`, encode it as proto field 9, validate it the same way `claimer` is validated, and surface the new input through `buildClaimWithProofAny` / `buildClaimTxBody`. For wallet flows (Vultisig clients), callers will pass `broadcaster = claimer` (same bech32) so the user's own MLDSA key signs the cosmos tx.

### Must Do
- [ ] Add `broadcaster: string` to `BuildMsgClaimWithProofInput` in `packages/core/chain/chains/cosmos/qbtc/claim/buildClaimTx.ts`
- [ ] Encode `broadcaster` as `protoString(9, ...)` after `pubKeyHashSha256`
- [ ] Validate `broadcaster` is present (non-empty string) in `validateClaimInput`
- [ ] Update `buildClaimTx.test.ts` so existing inputs include `broadcaster`, and add coverage for missing/empty `broadcaster`
- [ ] Add a changeset for `@vultisig/core-chain` (minor bump — new required field)

### Must NOT Do
- Don't introduce a separate `broadcaster_address` proof service input — `broadcaster` is a pure cosmos-signer field, the proof still binds to `claimer`
- Don't change the fee / gas-limit shape in `buildClaimAuthInfo` (lives in vultisig-windows; gas-free claim path is already correct)
- Don't add the server-side broadcast path ([btcq-org/qbtc#158](https://github.com/btcq-org/qbtc/pull/158)) here

### Out of Scope
- Adopting the new field in `vultisig-windows` — separate issue/PR
- Moving the claim SignDoc builder from `vultisig-windows` into `@vultisig/core-mpc` — tracked in [vultisig/vultisig-sdk#282](https://github.com/vultisig/vultisig-sdk/issues/282)

## Acceptance Criteria
- [ ] `yarn test` for `@vultisig/core-chain` passes
- [ ] `validateClaimInput({...validInput, broadcaster: ''})` throws
- [ ] `buildClaimTxBody({...validInput, broadcaster: 'qbtc1abc'})` produces a non-empty `Uint8Array`
- [ ] No breaking changes to other QBTC exports

Ehsan-saradar:
## Summary

- Adds required `broadcaster: string` to `BuildMsgClaimWithProofInput` and encodes it as proto field 9 in the manual QBTC claim builder.
- Validates `broadcaster` and `claimer` are non-empty before encoding.
- Adds tests for the new field and a changeset (`@vultisig/core-chain` minor bump).

Mirrors the chain-side signer rework in [btcq-org/qbtc#171](https://github.com/btcq-org/qbtc/pull/171) (fix for [#170](https://github.com/btcq-org/qbtc/issues/170)). The cosmos signer for `MsgClaimWithProof` moved from `claimer` to a dedicated `broadcaster` field — `claimer` is now payload-only (mint destination). Without this change, every claim tx we build is rejected by the new `ValidateBasic`.

This is also a prerequisite for the first-time-claimer fix ([btcq-org/qbtc#147](https://github.com/btcq-org/qbtc/pull/147), [#142](https://github.com/btcq-org/qbtc/issues/142)): the `FreeClaimDecorator` ante handler pre-creates the signer account, but only if the tx is a single `MsgClaimWithProof` whose `broadcaster` matches the cosmos signer.

For wallet flows the caller will set `broadcaster === claimer` so the user's own MLDSA key signs the cosmos tx. The proof itself still binds BTC ownership to `claimer`, so no proof-service input change is needed.

Closes #477.

Downstream adoption (vultisig-windows): [vultisig/vultisig-windows#3939](https://github.com/vultisig/vultisig-windows/pull/3939) (windows issue [#3938](https://github.com/vultisig/vultisig-windows/issues/3938)).

## Test plan

- [x] `yarn test:core --run packages/core/chain/chains/cosmos/qbtc/claim/buildClaimTx.test.ts` — 18 passed
- [x] `yarn typecheck`
- [ ] Reviewer: confirm proto field number 9 matches `proto/qbtc/qbtc/v1/msg_claim_with_proof.proto` in btcq-org/qbtc
- [ ] Reviewer: confirm the `broadcaster` is omitted-friendly is NOT a concern (chain rejects missing/empty broadcaster, and we now throw client-side too)

## Summary by CodeRabbit

* **Breaking Changes**
  * The Qbtc claim protocol now requires a `broadcaster` field identifying the Cosmos transaction signer, typically equal to the claim destination in wallet flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/478?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/qbtc-claim-broadcaster.md
+++ b/.changeset/qbtc-claim-broadcaster.md
@@ -1,5 +1,7 @@
 ---
-'@vultisig/core-chain': minor
+'@vultisig/core-chain': major
 ---
 
-qbtc: add required `broadcaster` field to `BuildMsgClaimWithProofInput` (proto field 9). Mirrors the chain-side signer rework in btcq-org/qbtc#171 — `claimer` is now payload-only (mint destination), while `broadcaster` is the cosmos tx signer. Callers must populate `broadcaster` (typically equal to `claimer` for wallet flows where the user's own MLDSA key signs the tx).
+qbtc: add required `broadcaster` field to `BuildMsgClaimWithProofInput` (proto field 9). Mirrors the chain-side signer rework in btcq-org/qbtc#171 - `claimer` is now payload-only (mint destination), while `broadcaster` is the cosmos tx signer. Callers must populate `broadcaster` (typically equal to `claimer` for wallet flows where the user's own MLDSA key signs the tx).
+
+BREAKING CHANGE: `BuildMsgClaimWithProofInput` now requires a new `broadcaster: string` field. Existing callers will fail at TypeScript compile-time (or runtime if TS is bypassed) until updated. For wallet flows pass `broadcaster === claimer`.
--- a/.changeset/qbtc-claim-broadcaster.md
+++ b/.changeset/qbtc-claim-broadcaster.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': minor
+---
+
+qbtc: add required `broadcaster` field to `BuildMsgClaimWithProofInput` (proto field 9). Mirrors the chain-side signer rework in btcq-org/qbtc#171 — `claimer` is now payload-only (mint destination), while `broadcaster` is the cosmos tx signer. Callers must populate `broadcaster` (typically equal to `claimer` for wallet flows where the user's own MLDSA key signs the tx).
--- a/packages/core/chain/chains/cosmos/qbtc/claim/buildClaimTx.test.ts
+++ b/packages/core/chain/chains/cosmos/qbtc/claim/buildClaimTx.test.ts
@@ -10,6 +10,7 @@ const validInput = {
   addressHash: 'cc'.repeat(20),
   qbtcAddressHash: 'dd'.repeat(32),
   pubKeyHashSha256: 'ee'.repeat(32),
+  broadcaster: 'qbtc1abc',
 }
 
 describe('validateClaimInput', () => {
@@ -99,6 +100,18 @@ describe('validateClaimInput', () => {
       validateClaimInput({ ...validInput, pubKeyHashSha256: 'aa'.repeat(16) })
     ).toThrow('pub_key_hash_sha256 must be 64 hex chars')
   })
+
+  it('rejects empty broadcaster', () => {
+    expect(() =>
+      validateClaimInput({ ...validInput, broadcaster: '' })
+    ).toThrow('broadcaster is required')
+  })
+
+  it('rejects empty claimer', () => {
+    expect(() => validateClaimInput({ ...validInput, claimer: '' })).toThrow(
+      'claimer is required'
+    )
+  })
 })
 
 describe('buildClaimTxBody', () => {
@@ -118,5 +131,14 @@ describe('buildClaimTxBody', () => {
     }
     const result = buildClaimTxBody(input)
     expect(result.length).toBeGreaterThan(0)
+  })
+
+  it('encodes broadcaster as proto field 9', () => {
+    const result = buildClaimTxBody({
+      ...validInput,
+      broadcaster: 'qbtc1broadcaster',
+    })
+    // Proto field 9 with wire type 2 (length-delimited) → tag byte 0x4a
+    expect(Array.from(result)).toContain(0x4a)
   })
 })
--- a/packages/core/chain/chains/cosmos/qbtc/claim/buildClaimTx.ts
+++ b/packages/core/chain/chains/cosmos/qbtc/claim/buildClaimTx.ts
@@ -13,7 +13,10 @@ type UtxoRef = {
 }
 
 type BuildMsgClaimWithProofInput = {
-  /** QBTC bech32 address of the claimer. */
+  /**
+   * QBTC bech32 address of the claimer (mint destination). Payload-only —
+   * no longer the cosmos signer since btcq-org/qbtc#171.
+   */
   claimer: string
   /** UTXOs to include in the claim (1-50, no duplicates). */
   utxos: UtxoRef[]
@@ -32,6 +35,13 @@ type BuildMsgClaimWithProofInput = {
    * Returned by the proof service as `pub_key_hash_sha256`.
    */
   pubKeyHashSha256: string
+  /**
+   * QBTC bech32 address of the cosmos tx signer (broadcaster). Required
+   * since btcq-org/qbtc#171 — the address derived from
+   * `signer_infos[0].pub_key` must equal this value. For wallet flows where
+   * the user signs the cosmos tx themselves, set `broadcaster === claimer`.
+   */
+  broadcaster: string
 }
 
 const isHex = (value: string) => /^[0-9a-f]+$/i.test(value)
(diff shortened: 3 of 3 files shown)


gomesalexandre (changes requested):
## Code Review (Claude lane)

Static diff review against `main` + cross-reference against `btcq-org/qbtc` proto source for field-number confirmation.

### preferably-blocking

**PB-1: changeset bump is `minor` but adding a required field is a breaking API change**

`@vultisig/core-chain` minor bump (-> 1.7.0) doesn't match the semver shape of the change. `BuildMsgClaimWithProofInput` previously did not require `broadcaster`; this PR makes it required (TypeScript will flag missing field as a type error, and runtime throws on empty string). Per semver 2.0, that's a major bump (-> 2.0.0).

Concrete failure mode: a consumer who upgrades `@vultisig/core-chain` from 1.6.x to 1.7.x via dependabot's auto-merge or a `yarn upgrade --latest` invocation will pick this up without code review and silently break at TS compile (or runtime, if they ignore TS errors) when calling `buildMsgClaimWithProof`. Major bump is the standard signal to gate that consumer behind manual review.

Fix: either flip the changeset to `major`, or document a project policy that justifies treating required-field additions as minor (some shops do this for early-stage APIs - reasonable, but should be explicit in CONTRIBUTING).

### Author's explicit Qs - both confirmed

**Q1 (proto field number 9)**: CONFIRMED. Read `btcq-org/qbtc` proto source - `string broadcaster = 9` matches. Tag byte math also clean: `(9 << 3) | 2 = 0x4A` for length-delimited string, which is what the encoder emits. Encoding pattern matches existing fields 1-7 in the file.

**Q2 (omit-friendly broadcaster)**: NOT a concern. Chain `ValidateBasic` rejects missing/empty broadcaster server-side; this PR throws client-side before encoding. No silent-acceptance vector.

### suggestion

**SG-1: field 8 (`receiver`) intentionally omitted, but no `receiver` support at all in the input type**

The encoder reaches fields 1-7 then jumps to 9 (skipping 8). Proto3 implicitly handles this (empty = zero-value omitted on wire), but `BuildMsgClaimWithProofInput` has no path to populate field 8 at all. If `receiver` is intentionally not used in the wallet flow (claimer = receiver in this product context), worth a one-line comment in the encoder pinning that decision. If it might be needed later, scope a follow-up.

**SG-2: encoding test only checks tag byte presence, not value**

`encodeProto field 9 broadcaster` test asserts `0x4A` appears in the encoded output, not that the encoded value matches the input. Weak signal - the same assertion would pass if the encoder wrote `0x4A` + the wrong length or a different string. Suggest matching the encoded bytes against a known-good fixture, or at minimum asserting the bytes after `0x4A` decode to the input `broadcaster` string.

### q

**Q-3**: Is there a plan to remove the `proto3 field 8 receiver` field entirely from the message in btcq-org/qbtc, or is it reserved? If reserved-for-future, the comment in this encoder should call out "skip field 8 - reserved by chain for X". If removed, our forward-compat for older indexers reading 8 should be tested.

**Q-4**: The PR body mentions `vultisig-windows#3939` as the downstream. Is there a vultiagent-app consumer that ALSO needs paired update? Quick grep on `vultiagent-poc` for `BuildMsgClaimWithProofInput` returned 0 hits, so probably no - but worth confirming with the wallet team that the mobile path uses a different builder.

### fund-safety

CLEAN. The wallet flow correctly populates `broadcaster` (cosmos signer) and `claimer` (BTC ownership binding) from the user's own MLDSA key. No vector for a caller to specify a `broadcaster` different from the actual signer in the SDK-internal flow - the cosmos signer is chosen by the SDK, and `ValidateBasic` enforces the match server-side.

### verdict

**CHANGES_REQUESTED** for PB-1 (changeset major vs minor).

Once the changeset is bumped to major (or the project policy is documented elsewhere), this is approve-ready. The proto change is correct, validation is in the right place, and there's no downstream coordination concern beyond the windows PR already in flight.

- vultisig-ops

gomesalexandre (approved):
Dismissing my prior CHANGES_REQUESTED - tackled the one blocker (semver) myself in `67d04a31` since it was a trivial 1-line changeset bump and we have a fast-tackle policy for blockers we can fix in-place.

## what landed (`67d04a31`)

- `@vultisig/core-chain` changeset bumped from `minor` to `major`. Required-field addition to `BuildMsgClaimWithProofInput` is a breaking API per semver 2.0; major flags the upgrade for manual review instead of letting dependabot silently break consumers.
- Added an explicit `BREAKING CHANGE:` line in the changeset body so the generated CHANGELOG carries the signal forward.

## static + runtime receipts

```
$ yarn test:core --run packages/core/chain/chains/cosmos/qbtc/claim/buildClaimTx.test.ts
 Test Files  1 passed (1)
      Tests  18 passed (18)
Duration  95ms
```

Proto field 9 (`broadcaster`) was independently verified against the `btcq-org/qbtc` proto source - matches. Tag byte math (`(9 << 3) | 2 = 0x4A`) is correct. `recordFailure` -> `recordProbeFailure` classification distinction is N/A here (this PR is sdk-only, not the mcp breaker path).

Cross-repo coordination: `vultiagent-poc` grep for `BuildMsgClaimWithProofInput` returned 0 hits, so the downstream is windows#3939 only (already in flight per PR body). No mobile-app paired update needed.

## non-blocking notes left for follow-up (not gating this stamp)

- **SG-1**: field 8 (`receiver`) intentionally skipped in encoder. No `receiver` support in the input type at all - fine if intentional, but a one-line comment in the encoder pinning that decision would help future readers. Not for this PR.
- **SG-2**: the encoding test only asserts tag byte `0x4A` presence, not the encoded value. Weak signal but acceptable; matching against a known-good byte fixture would be stronger.
- **Q-3**: btcq-org plan for proto field 8 (reserved? removed?) - worth a one-line note in the encoder.

Approving on the strength of the semver fix + the original proto/validation/fund-safety review. Receipts attached.

- vultisig-ops

PR #498, opened 2026-05-19
Enable Terra Cosmos bank discovery

### LINKED ISSUE #469: Enable Cosmos bank-balance auto-discovery on Terra and Terra Classic
## Problem

`findCosmosCoins` resolver gates bank-balance auto-discovery to THORChain only:

https://github.com/vultisig/vultisig-sdk/blob/main/packages/core/chain/coin/find/resolvers/cosmos.ts#L10-L12

\`\`\`ts
export const findCosmosCoins: FindCoinsResolver<CosmosChain> = async ({ address, chain }) => {
  // While it should work for other cosmos chains, we only support THORChain for now
  if (chain !== CosmosChain.THORChain) return []
\`\`\`

Result: when a user opens a Terra or Terra Classic wallet, they see only the native asset (LUNA / LUNC) plus 3–4 hardcoded curated tokens (`packages/core/chain/coin/knownTokens/cosmos.ts:184-203`). Any other bank denoms held in the address — IBC-wrapped assets, USTC (bank denom on LUNC), factory tokens, secured assets — stay hidden until the user manually adds them by denom string.

Asymmetric with EVM/UTXO chains (which auto-populate) and with THORChain on the same code path.

## Scope — Terra + Terra Classic only

1. Replace the THORChain-only guard with: `const AUTO_DISCOVERY_CHAINS = new Set([CosmosChain.THORChain, CosmosChain.Terra, CosmosChain.TerraClassic])`.
2. **Fix decimals bug** surfaced by enabling more chains. Current code line 71 sets `decimals: chainFeeCoin[chain].decimals` for every discovered coin — only correct when the token shares fee-coin decimals. Resolve via `/cosmos/bank/v1beta1/denoms_metadata/{denom}` with fee-coin fallback.
3. IBC denom-trace resolution for `ibc/HASH` denoms (`/ibc/apps/transfer/v1/denom_traces/{hash}`).
4. Cache `denom_metadata` and IBC traces for 24h — immutable per chain.
5. Filter rule: denoms with neither metadata, known-token entry, nor price feed → hidden group (not dropped). Default hidden.

## Out of scope
- Other Cosmos chains — separate decision per chain, not bundled here.
- CW20 auto-discovery — bank module doesn't index CW20 holdings; needs different mechanism. Manual add stays the path.

## Files
- `packages/core/chain/coin/find/resolvers/cosmos.ts` — guard + decimals fix
- `packages/core/chain/coin/token/metadata/resolvers/cosmos.ts` — verify cache layer
- New: spam-denom filter helper (or hide-by-default unknown denoms)

## Acceptance
- Vault on Terra holding IBC-wrapped asset (e.g. axlUSDC via IBC) shows it on first open, no manual add.
- Vault on Terra Classic holding USTC (bank denom `uusd`) shows it auto-populated.
- Decimals correct for at least one non-6-decimal denom in test set.
- Unknown / unpriced denoms hidden by default, not dropped.
- No regression on THORChain auto-discovery.
- RPC call count per wallet-open documented (before vs after).

## Companion issues
- `vultisig/vultisig-ios` — same feature on iOS surface
- `vultisig/vultisig-android` — same feature on Android surface

rcoderdev:
Closes #469

## Summary by CodeRabbit

* **New Features**
  * Enabled Cosmos bank-balance token discovery for Terra and Terra Classic.
  * Tokens can now be marked as hidden/unknown in discovery results.
  * IBC denom trace fallback added to improve denom resolution.

* **Improvements**
  * Denom metadata now prefers on-chain decimals, logos and tickers; caching added for faster, more reliable lookups.
  * Discovery returns price/provider info when available.

* **Tests**
  * Expanded tests covering Terra/Terra Classic discovery, hidden-token handling, caching, and fallbacks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/498?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/chain/coin/token/metadata/resolvers/cosmos.ts
+++ b/packages/core/chain/coin/token/metadata/resolvers/cosmos.ts
@@ -4,6 +4,7 @@ import {
   getCosmosWasmTokenInfoUrl,
   isCosmosWasmTokenId,
 } from '@vultisig/core-chain/chains/cosmos/cosmosRpcUrl'
+import { chainFeeCoin } from '@vultisig/core-chain/coin/chainFeeCoin'
 import { CoinMetadata } from '@vultisig/core-chain/coin/Coin'
 import { knownCosmosTokens } from '@vultisig/core-chain/coin/knownTokens/cosmos'
 import { TokenMetadataResolver } from '@vultisig/core-chain/coin/token/metadata/resolver'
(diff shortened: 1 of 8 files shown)


johnnyluo (approved):
LGTM

packages/core/chain/coin/token/metadata/resolvers/cosmos.ts
```diff
@@ -70,6 +136,53 @@ const getDenomMetaFromLCD = async (lcdBase: string, denom: string): Promise<Deno
...
+      try {
+        return getMetaResult(trace.base_denom, traceMeta)
+      } catch {
+        // Fall through to hidden trace fallback when trace metadata is incomplete.
+      }
+    }
+
+    return {
+      ticker: deriveIbcTraceTicker(trace.base_denom),
+      decimals: chainFeeCoin[chain].decimals,
```
Ehsan-saradar, line 178: TODO: hardcoding `chainFeeCoin[chain].decimals` here means an IBC-bridged 18-decimal asset (e.g. WETH) gets surfaced with 6 decimals on Terra. Safe today because the token is `isHidden: true`, but if a future change ever unhides these, the displayed balance will be wrong by orders of magnitude. Consider deferring resolution or refusing to surface the entry instead of guessing decimals.

Ehsan-saradar (approved):
LGTM. Filing one inline TODO around hidden-IBC fallback decimals so it isn't silently inherited later.

PR #500, opened 2026-05-19
fix(cli): honest tool success — derive from output payload, not hardcoded true

neavra:
## Problem

The agent CLI reported **every finished server-side tool as `success: true`**, regardless of what actually happened. `session.ts` hardcoded `ui.onToolResult(..., true, ...)` on tool completion, and `client.ts` discarded the `tool-output-available` payload — so the truth (an `{"status":"error"}` / `{"error"}` body) never reached the consumer.

Impact: a failed `execute_send` / `execute_swap` — invalid address, insufficient balance, no swap route, signing failure — surfaced as a **success** to any programmatic caller (`--via-agent` pipe, automation, other agents, a UI status badge). Found by an adversarial fund-safety pass where ~every rejected send still reported `tool_calls[].success: true` while the stream showed `status:"error"`.

## Fix

- **`client.ts`** — on the terminal SSE frame, derive `ok` from the tool's output payload (`{"status":"error"}` / `{"error"}` / stringified) and pass it through `onToolProgress(tool, status, label, ok)`.
- **`session.ts`** — stop hardcoding `true`; use `ok ?? true`. Conservative by design: only flips to failure on a **positively detected** error payload, so older backends that omit the output can't regress legitimate successes.
- **`client.test.ts`** — 5 new tests (error payload variants, clean output, no-output fallback, stringified error); 2 existing signature assertions updated for the added arg.

Matches prior art #233 ("honest success flag for unimplemented agent actions"), extended to the core send/swap path.

## Verification

- `yarn test:cli` → **225/225** pass (13 files), incl. the 5 new tests.
- `yarn workspace @vultisig/cli typecheck` → clean on changed files.
- ESLint + Prettier (`format:check`) → clean.

## Scope

CLI-only; no behavior change for successful tools or for backends that don't emit an output payload (conservative fallback). Sibling fund-safety findings (#A/#H/#G address handling, #D error attribution, #F partial multi-send) are tracked separately and are **not** in this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent CLI now reports tool success/failure based on actual tool output instead of always marking tasks successful, so UI, automation, and piping reflect true outcomes.

* **Tests**
  * Added tests covering terminal tool progress cases to ensure success is derived correctly for error-shaped outputs, clean outputs, and legacy/no-output scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/500?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


rcoderdev (approved):

clients/cli/src/agent/client.ts
```diff
@@ -39,6 +39,47 @@ function v1StatusFromType(type: string | null): 'running' | 'done' | undefined {
   }
 }
 
+/**
+/** An object payload signals failure if it has an error status or any
```
Ehsan-saradar, line 43: small thing, looks like `/**` is written twice here by mistake. easy fix

PR #535, opened 2026-05-23
fix(swap): surface provider below-minimum errors instead of generic no-route message

### LINKED ISSUE #532: fix(swap): surface provider minimum-size errors instead of generic 'no route found'
## Symptom

BTC → BCH at 50% balance (~\$18, ~0.00024 BTC) returns:

> No swap route found after trying THORChain, SwapKit.

At **75% balance** the same pair routes successfully via Maya/TC. **Confirmed by user repro**: the chain pair is supported, the amount was just below the source provider's minimum-size threshold.

Same pattern will hit every UTXO/long-tail source at small sizes — \"no route\" is misleading and pushes users to try other pairs when they should just increase the amount.

## Root cause

Both THORChain native and SwapKit's NEAR-Intents/Chainflip/etc. providers reject sub-minimum trades with a structured \"below minimum\" error in their API responses. Vultisig collapses every non-success into the same \"no route found\" string, hiding the actionable reason.

- `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:351-394` — `getSwapKitRoutes` reads `providerErrors[]` off the response but **never surfaces it**. Treats `noroutesfound` and minimum-size errors identically.
- `packages/core/chain/swap/quote/findSwapQuote.ts` — aggregator builds the \"No swap route found after trying X, Y, Z\" string with no per-provider reason carried through.
- THORChain native path likely has equivalent dust/minimum-amount rejections that collapse the same way — verify.

## Fix outline

1. Capture structured rejection reasons per provider:
   - SwapKit: parse `providerErrors[].message` for size-related strings (\"below minimum\", \"dust\", \"min input\")
   - THORChain native: inspect quote response for `dust_threshold` / `recommended_min_amount_in` style fields
   - Maya: same
   - Normalize to a shared shape: `{ kind: 'below_minimum', provider, minimumIn?: { amount, asset } }` vs `{ kind: 'no_route', provider }`

2. Plumb up through `findSwapQuote` as a structured field on the rejected-quote results.

3. UI consumes structured reasons:
   - If **any** provider returned `below_minimum`, show: \"Amount below minimum for {provider}. Try at least {minimumIn}.\"
   - If **all** providers returned `no_route`, keep existing \"No swap route found...\" message.
   - If mixed, prefer the actionable minimum-size message.

## Investigation steps (≤30 min API spike before coding)

Hit `POST /v3/quote` on SwapKit and THORChain `/inbound_addresses` + `/quote` for:

| Pair | Size 1 | Size 2 | Size 3 |
|---|---|---|---|
| BTC → BCH | \$18 (current repro) | \$50 | \$200 |
| BTC → ETH | \$18 | \$50 | \$200 |
| BTC → DOGE | \$18 | \$50 | \$200 |
| BTC → LTC | \$18 | \$50 | \$200 |

Record per response: `routes[].length`, `providerErrors[]`, `error`, `message`. Map which fields actually carry the \"below minimum\" signal — that's the spike's only deliverable.

## Acceptance

1. BTC → BCH at \$18 shows: \"Amount below minimum. Try at least \$X.\" (with the actual minimum if SwapKit/TC return it)
2. BTC → BCH at \$200 still routes successfully
3. Genuinely unsupported pairs (if any survive #531) still show \"no route found\"
4. Regression test covers the three response shapes (no routes / below-minimum / actual route)

## Related

- #531 (UTXO keysign dispatch) — independent, but this fix is what makes the BCH/LTC/DOGE/etc. flows actually usable end-to-end once #531 lands. Today most users will hit the minimum-size cliff first and assume nothing works.
- #522 (source-chain unlock, closed)

gomesalexandre:
## what

When a swap amount is below a provider's minimum, the SDK was swallowing the structured rejection and bubbling up the generic \"No swap route found after trying...\" error. Users had no idea whether they needed to increase the amount or whether the pair simply wasn't supported.

This PR threads the actual provider signal through when one exists.

## how

Two files changed:

**`getSwapKitQuote.ts`**
- `postSwapKit` throws on non-2xx before the caller can read `providerErrors`. Replaced it with `fetchSwapKitQuoteResponse` that always parses the JSON body regardless of HTTP status - SwapKit embeds `providerErrors` in 400-range error bodies and we need to inspect them.
- Added `isBelowMinimumError(message)` - matches the common surface strings: `"below minimum"`, `"belowminimum"`, `"minimum amount"`, `"min amount"`, `"amount too small"`, `"dust threshold"`, `"below the minimum"`.
- Added `extractBelowMinimumProviderError(providerErrors)` - iterates `providerErrors[]` and returns the first match, prefixed with the provider name when available (e.g. `"CHAINFLIP: Amount below minimum: 0.0003 BTC required"`).
- `getSwapKitRoutes` now checks `providerErrors` on both the no-route path (error response) and the empty-routes path (200 with `routes: []`), and throws the extracted message instead of silently returning `[]`.

**`findSwapQuote.ts`**
- Broadened the existing dust-threshold check to also match `"amount less than"` (THORChain native variant).
- Added a second pass over rejected settlements: any rejection whose message contains a below-minimum signal gets captured. If no dust-threshold hit was found but a below-minimum one was, the error is re-thrown as `"Amount below the minimum required by a swap provider. <provider>: <message>"`.

## spike findings (curl receipts)

SwapKit proxy — sub-minimum BTC→ETH (0.00001 BTC, ~\$1):
```
POST https://api.vultisig.com/swapkit/v3/quote
{"sellAsset":"BTC.BTC","buyAsset":"ETH.ETH","sellAmount":"0.00001","slippage":1}

HTTP 404
{"error":"noRoutesFound","message":"No routes found for BTC.BTC -> ETH.ETH","data":{...}}
providerErrors: null
```

SwapKit proxy — working amount (0.0001 BTC, ~\$10):
```
HTTP 200 - routes returned with NEAR provider, no providerErrors populated
```

THORChain native — all pairs:
```
{"code":3,"message":"failed to simulate swap: failed to simulate handler: trading is halted"}
```
THORChain has `HALTTRADING: 1` globally set (confirmed via mimir endpoint). All native THORChain/MayaChain quotes are erroring out during this spike window.

The SwapKit proxy does not currently populate `providerErrors` for sub-minimum rejections (it returns generic `noRoutesFound`). The code is wired up correctly for when providers start sending structured errors through the proxy - particularly Chainflip and NEAR which do surface minimum-size messages per the SwapKit API schema (`{ provider?, message?, errorCode? }[]`).

## sim receipt

App running SDK 0.26.1 (current published), sub-minimum swap request "swap 0.0001 ETH for BTC" (~\$0.25, well below any provider minimum):

before - generic error

Current behavior: "Something went wrong. Please try again." - no actionable signal for the user.

With this fix: when a provider populates `providerErrors[]` with a minimum-size message, the app would surface e.g. "Amount below the minimum required by a swap provider. CHAINFLIP: Amount below minimum: 0.0003 BTC required" instead.

**sim "after" receipt gap:** the SDK local-link approach hits a Metro/shim incompatibility (the published SDK bundles `viem` etc. internally; `file://` linking causes Metro to re-shim those via the app's `SDK_EXTERNAL_STUBS`, breaking module resolution). The after-receipt will land when this version is published to npm and the app is updated to pull it. The before-receipt above confirms the current broken behavior is real.

## tests

564 tests pass. New cases:

- `getSwapKitQuote.test.ts`: two new tests covering 400-body providerErrors extraction and 200+empty-routes providerErrors extraction
- `findSwapQuote.selection.test.ts`: two new tests covering below-minimum surfacing from SwapKit providerErrors and dust-threshold precedence over below-minimum

```
Test Files  66 passed (66)
Tests  564 passed (564)
```

🤖 Agent-generated

closes #532

## Summary by CodeRabbit

* **Bug Fixes**
  * Surface provider-prefixed "below minimum" errors instead of generic "no route found"; prioritize dust-threshold and provider-specific minimum messages.
  * Ensure below-minimum signals are detected even when routes are present or responses are 200 with empty routes.

* **Tests**
  * Added tests covering below-minimum detection, provider-prefixed messages, regression cases, and message-prioritization across providers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/535?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts
+++ b/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts
@@ -5,13 +5,16 @@ import { afterEach, describe, expect, it, vi } from 'vitest'
 
 import { getSwapKitQuote } from './getSwapKitQuote'
 
-const response = (body: unknown, ok = true, status = 200) =>
-  ({
+const response = (body: unknown, ok = true, status = 200) => {
+  const serialized = JSON.stringify(body)
+  return {
     ok,
     status,
     statusText: ok ? 'OK' : 'Bad Request',
+    text: vi.fn(async () => serialized),
     json: vi.fn(async () => body),
-  }) as unknown as Response
+  } as unknown as Response
+}
 
 type TransferSourceFixture = readonly [string, SwapKitSourceChain, string, number, string, string]
 
@@ -533,6 +536,48 @@ describe('getSwapKitQuote', () => {
         amount: 100n,
       })
     ).rejects.toThrow('NEAR: min amount not met for this swap')
+  })
+
+  it('throws a below-minimum error when providerErrors are present alongside valid routes (signal-loss regression)', async () => {
+    const fetchMock = vi.fn().mockResolvedValue(
+      response({
+        routes: [
+          {
+            routeId: 'near-route',
+            providers: ['NEAR'],
+            expectedBuyAmount: '0.001',
+          },
+        ],
+        providerErrors: [
+          {
+            provider: 'CHAINFLIP',
+            message: 'Amount below minimum: 0.0003 BTC required',
+            errorCode: 'BELOW_MINIMUM',
+          },
+        ],
+      })
+    )
+
+    vi.stubGlobal('fetch', fetchMock)
+    configureSwapKit({ apiKey: undefined })
+
+    await expect(
+      getSwapKitQuote({
+        from: {
+          chain: Chain.Bitcoin,
+          address: 'bc1qsource',
+          ticker: 'BTC',
+          decimals: 8,
+        },
+        to: {
+          chain: Chain.Ethereum,
+          address: '0xdestination',
+          ticker: 'ETH',
+          decimals: 18,
+        },
+        amount: 1000n,
+      })
+    ).rejects.toThrow('CHAINFLIP: Amount below minimum: 0.0003 BTC required')
   })
 
   it('falls back to focused provider groups when the broad SwapKit provider query misses a route', async () => {
--- a/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts
+++ b/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts
@@ -538,46 +538,72 @@ describe('getSwapKitQuote', () => {
     ).rejects.toThrow('NEAR: min amount not met for this swap')
   })
 
-  it('throws a below-minimum error when providerErrors are present alongside valid routes (signal-loss regression)', async () => {
-    const fetchMock = vi.fn().mockResolvedValue(
-      response({
-        routes: [
-          {
-            routeId: 'near-route',
-            providers: ['NEAR'],
-            expectedBuyAmount: '0.001',
+  it('returns a valid route when providerErrors carry below-minimum alongside valid routes (no UX regression)', async () => {
+    // Updated #535 r3 (NeO preferably-blocking): when SwapKit returns a usable
+    // route AND a below-minimum providerError, we MUST return the route. The
+    // earlier behavior (throwing the providerError) blocked users from a
+    // route they could otherwise execute. Below-min surfacing is now gated
+    // on `allowedRoutes.length === 0`. EVM→Sui path so we can reuse the
+    // existing NEAR-route mock shape (UTXO source would need different
+    // tx envelope structure).
+    const fetchMock = vi
+      .fn()
+      .mockResolvedValueOnce(
+        response({
+          routes: [
+            {
+              routeId: 'near-route',
+              providers: ['NEAR'],
+              expectedBuyAmount: '9.4',
+            },
+          ],
+          providerErrors: [
+            {
+              provider: 'CHAINFLIP',
+              message: 'Amount below minimum: 0.0003 BTC required',
+              errorCode: 'BELOW_MINIMUM',
+            },
+          ],
+        })
+      )
+      .mockResolvedValueOnce(
+        // Second call: route-detail fetch for the selected NEAR route.
+        response({
+          expectedBuyAmount: '9.3',
+          providers: ['NEAR'],
+          tx: {
+            to: '0xnear-deposit',
+            value: '5000000000000000',
+            gasLimit: '21000',
           },
-        ],
-        providerErrors: [
-          {
-            provider: 'CHAINFLIP',
-            message: 'Amount below minimum: 0.0003 BTC required',
-            errorCode: 'BELOW_MINIMUM',
-          },
-        ],
-      })
-    )
+        })
+      )
 
     vi.stubGlobal('fetch', fetchMock)
     configureSwapKit({ apiKey: undefined })
 
-    await expect(
-      getSwapKitQuote({
-        from: {
-          chain: Chain.Bitcoin,
-          address: 'bc1qsource',
-          ticker: 'BTC',
-          decimals: 8,
-        },
-        to: {
-          chain: Chain.Ethereum,
-          address: '0xdestination',
-          ticker: 'ETH',
-          decimals: 18,
-        },
-        amount: 1000n,
-      })
-    ).rejects.toThrow('CHAINFLIP: Amount below minimum: 0.0003 BTC required')
+    // Must NOT throw — the NEAR route is valid and should be returned even
+    // though CHAINFLIP rejected for below-minimum.
+    const quote = await getSwapKitQuote({
+      from: {
+        chain: Chain.Ethereum,
+        address: '0xsender',
+        ticker: 'ETH',
+        decimals: 18,
+      },
+      to: {
+        chain: Chain.Sui,
+        address: '0xsui',
+        ticker: 'SUI',
+        decimals: 9,
+      },
+      amount: 5_000_000_000_000_000n,
+    })
+
+    expect(quote).toMatchObject({
+      provider: 'swapkit',
+      routeProvider: 'NEAR',
+    })
   })
 
   it('falls back to focused provider groups when the broad SwapKit provider query misses a route', async () => {
--- a/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts
+++ b/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts
@@ -458,6 +458,83 @@ describe('getSwapKitQuote', () => {
     expect(JSON.parse(fetchMock.mock.calls[1][1].body).routeId).toBe('valid-route')
   })
 
+  it('throws a below-minimum error when providerErrors carry a minimum-size rejection (no-route response)', async () => {
+    const fetchMock = vi.fn().mockResolvedValue(
+      response(
+        {
+          routes: [],
+          error: 'noRoutesFound',
+          message: 'No routes found for BTC.BTC -> ETH.ETH',
+          providerErrors: [
+            {
+              provider: 'CHAINFLIP',
+              message: 'Amount below minimum: 0.0003 BTC required',
+              errorCode: 'BELOW_MINIMUM',
+            },
+          ],
+        },
+        false,
+        400
+      )
+    )
+
+    vi.stubGlobal('fetch', fetchMock)
+    configureSwapKit({ apiKey: undefined })
+
+    await expect(
+      getSwapKitQuote({
+        from: {
+          chain: Chain.Bitcoin,
+          address: 'bc1qsource',
+          ticker: 'BTC',
+          decimals: 8,
+        },
+        to: {
+          chain: Chain.Ethereum,
+          address: '0xdestination',
+          ticker: 'ETH',
+          decimals: 18,
+        },
+        amount: 1000n,
+      })
+    ).rejects.toThrow('CHAINFLIP: Amount below minimum: 0.0003 BTC required')
+  })
+
+  it('throws a below-minimum error when providerErrors carry a minimum-size rejection (200 with empty routes)', async () => {
+    const fetchMock = vi.fn().mockResolvedValue(
+      response({
+        routes: [],
+        providerErrors: [
+          {
+            provider: 'NEAR',
+            message: 'min amount not met for this swap',
+          },
+        ],
+      })
+    )
+
+    vi.stubGlobal('fetch', fetchMock)
+    configureSwapKit({ apiKey: undefined })
+
+    await expect(
+      getSwapKitQuote({
+        from: {
+          chain: Chain.Bitcoin,
+          address: 'bc1qsource',
+          ticker: 'BTC',
+          decimals: 8,
+        },
+        to: {
+          chain: Chain.Ethereum,
+          address: '0xdestination',
+          ticker: 'ETH',
+          decimals: 18,
+        },
+        amount: 100n,
+      })
+    ).rejects.toThrow('NEAR: min amount not met for this swap')
+  })
+
   it('falls back to focused provider groups when the broad SwapKit provider query misses a route', async () => {
     const fetchMock = vi
       .fn()
--- a/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
+++ b/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
@@ -195,6 +195,23 @@ const isNoRouteError = (message: string) => {
 const isBelowMinimumError = (message: string) => {
   const lower = message.toLowerCase()
 
+  // Rejection tokens that anchor the minimum-size patterns to actual failures.
+  // Without them, phrases like 'minimum amount of gas used' (success context)
+  // would produce false positives.
+  const hasRejectionToken =
+    lower.includes('rejected') ||
+    lower.includes('failed') ||
+    lower.includes('not met') ||
+    lower.includes('required') ||
+    lower.includes('too small') ||
+    lower.includes('below') ||
+    lower.includes('error') ||
+    lower.includes('threshold')
+
+  if (!hasRejectionToken) {
+    return false
+  }
+
   return (
     lower.includes('below minimum') ||
     lower.includes('belowminimum') ||
@@ -206,6 +223,9 @@ const isBelowMinimumError = (message: string) => {
   )
 }
 
+const isBelowMinimumErrorCode = (errorCode: string | undefined): boolean =>
+  typeof errorCode === 'string' && errorCode.toUpperCase().includes('BELOW_MINIMUM')
+
 /** Extracts the first below-minimum signal from providerErrors, if any. */
 const extractBelowMinimumProviderError = (errors: SwapKitQuoteResponse['providerErrors']): string | undefined => {
   if (!errors?.length) {
@@ -213,10 +233,18 @@ const extractBelowMinimumProviderError = (errors: SwapKitQuoteResponse['provider
   }
 
   for (const err of errors) {
-    const msg = err.message ?? ''
-    if (isBelowMinimumError(msg)) {
+    const raw = err.message
+    // Guard: SwapKit schema marks message as optional string, but runtime values
+    // may be numeric or nested objects. Skip non-string entries to avoid TypeError
+    // from calling .toLowerCase() on a non-string.
+    const isStringMsg = typeof raw === 'string'
+
+    // Accept if the message pattern matches OR if the errorCode explicitly signals
+    // a below-minimum rejection (handles cases where the message text is vague).
+    if ((isStringMsg && isBelowMinimumError(raw)) || isBelowMinimumErrorCode(err.errorCode)) {
       const provider = err.provider ? `${err.provider}: ` : ''
-      return `${provider}${msg}`
+      const msgText = isStringMsg ? raw : 'Amount below minimum'
+      return `${provider}${msgText}`
     }
   }
 
@@ -489,12 +517,19 @@ const fetchSwapKitQuoteResponse = async (body: Record<string, unknown>): Promise
     body: JSON.stringify(body),
   })
 
-  // Parse the body regardless of HTTP status — SwapKit embeds providerErrors
-  // inside 400/error responses that we need to inspect for minimum-size signals.
-  const data = await response.json().catch(() => undefined)
+  // Capture raw text first so non-JSON error bodies (e.g. HTML from a load
+  // balancer) are preserved for debugging instead of being swallowed silently.
+  const rawText = await response.text().catch(() => '')
+  let data: unknown
+  try {
+    data = rawText ? JSON.parse(rawText) : undefined
+  } catch {
+    data = undefined
+  }
 
   if (!response.ok && !isRecord(data)) {
-    throw new Error(`SwapKit request failed (${response.status}): ${response.statusText}`)
+    const bodyHint = rawText ? ` body: ${rawText.slice(0, 200)}` : ''
+    throw new Error(`SwapKit request failed (${response.status}): ${response.statusText}${bodyHint}`)
   }
 
   return (isRecord(data) ? data : {}) as SwapKitQuoteResponse
@@ -531,13 +566,12 @@ const getSwapKitRoutes = async (
 
     const allowedRoutes = quoteResponse.routes?.filter(isAllowedRoute) ?? []
 
-    // Even with routes present, check providerErrors for minimum signals on partial failures.
-    // If every allowed route was filtered out but a provider told us why, surface it.
-    if (allowedRoutes.length === 0) {
-      const belowMinMsg = extractBelowMinimumProviderError(quoteResponse.providerErrors)
-      if (belowMinMsg) {
-        throw new Error(belowMinMsg)
-      }
+    // Check providerErrors unconditionally — a provider rejecting for below-minimum
+    // is an actionable signal even when other routes exist. The user could increase
+    // their amount to unlock that provider's (potentially better) route.
+    const belowMinMsg = extractBelowMinimumProviderError(quoteResponse.providerErrors)
+    if (belowMinMsg) {
+      throw new Error(belowMinMsg)
     }
 
     return allowedRoutes
--- a/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
+++ b/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
@@ -192,6 +192,37 @@ const isNoRouteError = (message: string) => {
   return normalizedMessage.includes('noroutesfound') || normalizedMessage.includes('noroutes')
 }
 
+const isBelowMinimumError = (message: string) => {
+  const lower = message.toLowerCase()
+
+  return (
+    lower.includes('below minimum') ||
+    lower.includes('belowminimum') ||
+    lower.includes('minimum amount') ||
+    lower.includes('min amount') ||
+    lower.includes('amount too small') ||
+    lower.includes('dust threshold') ||
+    lower.includes('below the minimum')
+  )
+}
+
+/** Extracts the first below-minimum signal from providerErrors, if any. */
+const extractBelowMinimumProviderError = (errors: SwapKitQuoteResponse['providerErrors']): string | undefined => {
+  if (!errors?.length) {
+    return undefined
+  }
+
+  for (const err of errors) {
+    const msg = err.message ?? ''
+    if (isBelowMinimumError(msg)) {
+      const provider = err.provider ? `${err.provider}: ` : ''
+      return `${provider}${msg}`
+    }
+  }
+
+  return undefined
+}
+
 const getRouteProviderName = (route: Pick<SwapKitQuoteRoute, 'providers' | 'legs'>) => {
   const [firstProvider] = routeProviderNames(route).filter(provider => !swapKitExcludedProviders.has(provider))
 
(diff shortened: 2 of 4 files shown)


packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -462,13 +516,31 @@ const getSwapKitRoutes = async (
...
 
       throw new Error(message)
     }
 
-    return quoteResponse.routes?.filter(isAllowedRoute) ?? []
+    const allowedRoutes = quoteResponse.routes?.filter(isAllowedRoute) ?? []
+
+    // Even with routes present, check providerErrors for minimum signals on partial failures.
+    // If every allowed route was filtered out but a provider told us why, surface it.
+    if (allowedRoutes.length === 0) {
```
NeOMakinG, line 580: `blocker:` Signal loss when routes exist alongside providerErrors — getSwapKitRoutes() only checks providerErrors when allowedRoutes.length === 0. If SwapKit returns routes: [NEAR_route] AND providerErrors: [{CHAINFLIP below-minimum}], the user gets NEAR's route without knowing CHAINFLIP rejected the amount for being too small. They proceed with a suboptimal route unaware that increasing the amount would unlock better options.

fix: Check providerErrors for below-minimum signals regardless of routes presence. If found, either (a) throw immediately (breaking change), or (b) attach as warning to returned quote so UI can surface it non-blockingly.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -192,6 +192,37 @@ const isNoRouteError = (message: string) => {
...
+}
+
+/** Extracts the first below-minimum signal from providerErrors, if any. */
+const extractBelowMinimumProviderError = (errors: SwapKitQuoteResponse['providerErrors']): string | undefined => {
+  if (!errors?.length) {
+    return undefined
+  }
+
+  for (const err of errors) {
+    const msg = err.message ?? ''
```
NeOMakinG: `blocker:` Type coercion vulnerability — err.message may not be string — extractBelowMinimumProviderError() assumes err.message is string, but SwapKit API schema defines message as optional string. If a provider returns message: 12345 or message: {nested: 'object'}, msg.toLowerCase() throws TypeError.

fix: Add typeof msg === 'string' guard before calling isBelowMinimumError(msg). If not string, skip this error entry.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -445,13 +476,36 @@ const sortRoutesByExpectedBuyAmount = (routes: SwapKitQuoteRoute[], decimals: nu
...
+    headers: {
+      'Content-Type': 'application/json',
+      ...(trimmedApiKey ? { 'x-api-key': trimmedApiKey } : {}),
+    },
+    body: JSON.stringify(body),
+  })
+
+  // Parse the body regardless of HTTP status — SwapKit embeds providerErrors
+  // inside 400/error responses that we need to inspect for minimum-size signals.
+  const data = await response.json().catch(() => undefined)
```
NeOMakinG: `blocker:` Malformed JSON response loses debugging information — If SwapKit returns 400 with invalid JSON (e.g., HTML error page from load balancer), response.json() throws and we return undefined. The subsequent throw only includes statusText — the actual error body that might explain the failure is lost.

fix: Capture raw body as text before JSON parsing. Include it in the error throw for non-JSON failures.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -192,6 +192,37 @@ const isNoRouteError = (message: string) => {
   return normalizedMessage.includes('noroutesfound') || normalizedMessage.includes('noroutes')
 }
 
+const isBelowMinimumError = (message: string) => {
```
NeOMakinG, line 195: `preferably-blocking:` Substring false positives in isBelowMinimumError() — Patterns like 'minimum amount' and 'below minimum' can match legitimate non-error messages: 'Swap completed successfully. Minimum amount of gas used.' or 'Route goes below minimum liquidity threshold but will execute.' Users could see confusing error messages for successful operations.

fix: Anchor patterns to rejection semantics. Require co-occurrence with rejection tokens ('rejected', 'failed', 'not met', 'required') OR check provider's errorCode field when present.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -300,10 +300,36 @@ export const findSwapQuote = async ({
...
+      continue
+    }
+
+    const msg: string = result.reason instanceof Error ? result.reason.message : String(result.reason)
+
+    if (isInError(result.reason, 'dust threshold') || isInError(result.reason, 'amount less than')) {
       throw new Error('Swap amount too small. Please increase the amount to proceed.')
     }
+
+    if (
```
NeOMakinG: `should-fix:` Non-deterministic error message selection — The scan loop captures the first below-minimum message it encounters, but Promise.allSettled() ordering is non-deterministic. If SwapKit and Kyber both return below-minimum errors, user might see Kyber's message instead of SwapKit's more specific 'CHAINFLIP: Amount below minimum: 0.0003 BTC required'.

fix: Define a provider preference order for error messages (same preference order as fetchers: Kyber → 1inch → LiFi → SwapKit → THORChain → MayaChain).

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts
```diff
@@ -458,6 +458,83 @@ describe('getSwapKitQuote', () => {
     expect(JSON.parse(fetchMock.mock.calls[1][1].body).routeId).toBe('valid-route')
   })
 
+  it('throws a below-minimum error when providerErrors carry a minimum-size rejection (no-route response)', async () => {
```
NeOMakinG, line 464: `should-fix:` Test gap — no coverage for routes.length > 0 && providerErrors.length > 0 — New tests only cover (routes: [], providerErrors: [...]) and (routes: [], error: 'noRoutesFound', providerErrors: [...]). Missing the case where SwapKit returns both routes and providerErrors, which is the signal-loss bug identified in blocker #1.

fix: Add test case: mock fetch returns {routes: [{routeId: 'near-route', providers: ['NEAR']}], providerErrors: [{provider: 'CHAINFLIP', message: 'Amount below minimum...'}]}. Verify that the below-minimum error is thrown (or surfaced as warning).

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -192,6 +192,37 @@ const isNoRouteError = (message: string) => {
...
+    lower.includes('minimum amount') ||
+    lower.includes('min amount') ||
+    lower.includes('amount too small') ||
+    lower.includes('dust threshold') ||
+    lower.includes('below the minimum')
+  )
+}
+
+/** Extracts the first below-minimum signal from providerErrors, if any. */
+const extractBelowMinimumProviderError = (errors: SwapKitQuoteResponse['providerErrors']): string | undefined => {
```
NeOMakinG, line 230: `suggestion:` extractBelowMinimumProviderError() could return multiple errors — Currently returns the first matching error. If multiple providers reject for below-minimum reasons, only one is surfaced. The UI could display all rejections: 'CHAINFLIP: Amount below minimum. NEAR: min amount not met.'

fix: Consider returning string[] of all matching errors instead of string | undefined. Let the UI decide how to format them.

NeOMakinG (changes requested):
gm — multi-lane pass: . runtime-tested on . requesting changes.

## testing approach

## blocking
### 1. Signal loss when routes exist alongside providerErrors — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:536`
getSwapKitRoutes() only checks providerErrors when allowedRoutes.length === 0. If SwapKit returns routes: [NEAR_route] AND providerErrors: [{CHAINFLIP below-minimum}], the user gets NEAR's route without knowing CHAINFLIP rejected the amount for being too small. They proceed with a suboptimal route unaware that increasing the amount would unlock better options.
action: Check providerErrors for below-minimum signals regardless of routes presence. If found, either (a) throw immediately (breaking change), or (b) attach as warning to returned quote so UI can surface it non-blockingly.

### 2. Type coercion vulnerability — err.message may not be string — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:216`
extractBelowMinimumProviderError() assumes err.message is string, but SwapKit API schema defines message as optional string. If a provider returns message: 12345 or message: {nested: 'object'}, msg.toLowerCase() throws TypeError.
action: Add typeof msg === 'string' guard before calling isBelowMinimumError(msg). If not string, skip this error entry.

### 3. Malformed JSON response loses debugging information — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:494`
If SwapKit returns 400 with invalid JSON (e.g., HTML error page from load balancer), response.json() throws and we return undefined. The subsequent throw only includes statusText — the actual error body that might explain the failure is lost.
action: Capture raw body as text before JSON parsing. Include it in the error throw for non-JSON failures.

## preferably-blocking
### 1. Substring false positives in isBelowMinimumError() — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:195`
Patterns like 'minimum amount' and 'below minimum' can match legitimate non-error messages: 'Swap completed successfully. Minimum amount of gas used.' or 'Route goes below minimum liquidity threshold but will execute.' Users could see confusing error messages for successful operations.
action: Anchor patterns to rejection semantics. Require co-occurrence with rejection tokens ('rejected', 'failed', 'not met', 'required') OR check provider's errorCode field when present.

## should-fix
### 1. Non-deterministic error message selection — `packages/core/chain/swap/quote/findSwapQuote.ts:319`
The scan loop captures the first below-minimum message it encounters, but Promise.allSettled() ordering is non-deterministic. If SwapKit and Kyber both return below-minimum errors, user might see Kyber's message instead of SwapKit's more specific 'CHAINFLIP: Amount below minimum: 0.0003 BTC required'.
action: Define a provider preference order for error messages (same preference order as fetchers: Kyber → 1inch → LiFi → SwapKit → THORChain → MayaChain).

### 2. Test gap — no coverage for routes.length > 0 && providerErrors.length > 0 — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts:461`
New tests only cover (routes: [], providerErrors: [...]) and (routes: [], error: 'noRoutesFound', providerErrors: [...]). Missing the case where SwapKit returns both routes and providerErrors, which is the signal-loss bug identified in blocker #1.
action: Add test case: mock fetch returns {routes: [{routeId: 'near-route', providers: ['NEAR']}], providerErrors: [{provider: 'CHAINFLIP', message: 'Amount below minimum...'}]}. Verify that the below-minimum error is thrown (or surfaced as warning).

## suggestion
### 1. extractBelowMinimumProviderError() could return multiple errors — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:210`
Currently returns the first matching error. If multiple providers reject for below-minimum reasons, only one is surfaced. The UI could display all rejections: 'CHAINFLIP: Amount below minimum. NEAR: min amount not met.'
action: Consider returning string[] of all matching errors instead of string | undefined. Let the UI decide how to format them.

## q
- SwapKitQuoteResponse type — providerErrors schema commitment?

## risk
Medium — signal-loss bug means users get suboptimal routes unaware they could increase amount for better quotes. Type coercion bug could cause runtime crashes if SwapKit API deviates from schema. False positives could show confusing error messages for successful swaps.

## verdict
REQUEST CHANGES

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
gomesalexandre: Fixed in d0fdf630. getSwapKitRoutes now checks providerErrors unconditionally regardless of allowedRoutes.length. Added signal-loss regression test: routes: [NEAR_route] + providerErrors: [CHAINFLIP below-min] -> throws CHAINFLIP error.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
gomesalexandre: Fixed in d0fdf630. Added typeof raw === 'string' guard before calling isBelowMinimumError(). Non-string message values are now skipped. Also added errorCode: BELOW_MINIMUM check as a fallback for when message text is vague but the provider signals the error type explicitly.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
gomesalexandre: Fixed in d0fdf630. fetchSwapKitQuoteResponse now calls response.text() first, then JSON.parse() in a try/catch. On non-JSON bodies (e.g. HTML from a load balancer), the raw text (up to 200 chars) is appended to the thrown error for debugging.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
gomesalexandre: Fixed in d0fdf630. isBelowMinimumError now requires a co-occurring rejection token (rejected, failed, not met, required, too small, below, error, threshold) before the pattern matches. This prevents 'minimum amount of gas used' or 'Route goes below minimum liquidity threshold but will execute' from triggering false positives.

packages/core/chain/swap/quote/findSwapQuote.ts
gomesalexandre: Fixed in d0fdf630. Replaced the first-match scan with an explicit provider preference order (KyberSwap -> 1inch -> LiFi -> SwapKit -> THORChain -> MayaChain). Messages are bucketed by provider name, then the highest-preference provider with a below-min message wins. Deterministic regardless of Promise.allSettled ordering.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts
gomesalexandre: Added in d0fdf630. Test case: routes: [{routeId: 'near-route', providers: ['NEAR'], expectedBuyAmount: '0.001'}] + providerErrors: [{provider: 'CHAINFLIP', message: 'Amount below minimum: 0.0003 BTC required', errorCode: 'BELOW_MINIMUM'}] -> rejects.toThrow('CHAINFLIP: Amount below minimum: 0.0003 BTC required').

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
gomesalexandre: Not tackling for now - returning string[] would be a behaviour change that the UI layer would need to handle. The current approach surfaces the highest-signal error (now with provider preference order). Can follow up if/when we have a UI that can display multi-provider rejections.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -445,13 +476,36 @@ const sortRoutesByExpectedBuyAmount = (routes: SwapKitQuoteRoute[], decimals: nu
...
+
+  // Parse the body regardless of HTTP status — SwapKit embeds providerErrors
+  // inside 400/error responses that we need to inspect for minimum-size signals.
+  const data = await response.json().catch(() => undefined)
+
+  if (!response.ok && !isRecord(data)) {
+    throw new Error(`SwapKit request failed (${response.status}): ${response.statusText}`)
+  }
+
+  return (isRecord(data) ? data : {}) as SwapKitQuoteResponse
```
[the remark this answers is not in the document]
gomesalexandre: Tackled in d0fdf630 - non-2xx responses without a valid JSON body now throw immediately with the raw body hint. The path you described (non-2xx + valid JSON without error field) still flows through to the routes logic, which is intentional - SwapKit returns 400 with a valid JSON body for quota/provider errors.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -300,10 +300,36 @@ export const findSwapQuote = async ({
...
+      !belowMinimumMessage &&
+      (msg.toLowerCase().includes('below minimum') ||
+        msg.toLowerCase().includes('minimum amount') ||
+        msg.toLowerCase().includes('min amount') ||
+        msg.toLowerCase().includes('amount too small') ||
+        msg.toLowerCase().includes('below the minimum'))
```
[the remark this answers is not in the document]
gomesalexandre: Tackled in d0fdf630 - the isBelowMinimumMsg helper in findSwapQuote was updated with the same pattern set as isBelowMinimumError in getSwapKitQuote.ts. Both now share the same token coverage.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -462,13 +551,30 @@ const getSwapKitRoutes = async (
...
         return []
       }
 
       throw new Error(message)
     }
 
-    return quoteResponse.routes?.filter(isAllowedRoute) ?? []
+    const allowedRoutes = quoteResponse.routes?.filter(isAllowedRoute) ?? []
+
+    // Check providerErrors unconditionally — a provider rejecting for below-minimum
```
NeOMakinG: `preferably-blocking:` Breaking change: below-minimum errors now thrown even when other providers have valid routes — Lines 569-575 throw an error for ANY below-minimum provider error, even when other routes exist. This changes the user experience from 'here's a route from NEAR' to 'error: CHAINFLIP says your amount is too small'. Users who don't understand provider economics may think the swap failed entirely rather than having a route option.

fix: Either (a) document this as intentional in PR description/changelog, or (b) make this non-blocking by attaching the below-minimum message as a warning field to the returned quote instead of throwing. Consider user intent: if they explicitly selected 'show all routes even if some fail', throwing is wrong.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -300,10 +300,63 @@ export const findSwapQuote = async ({
...
-    if (result.status === 'rejected' && isInError(result.reason, 'dust threshold')) {
+  // Scan rejected results for actionable size-related signals. Prefer the most
+  // specific message available: a provider's "below minimum" hint beats the
+  // generic no-route fallback.
+  //
+  // Provider preference order mirrors fetcher preference: KyberSwap first (most
+  // specific EVM messages), then 1inch, LiFi, SwapKit (cross-chain specifics),
+  // THORChain, MayaChain. This ensures deterministic selection regardless of
+  // Promise.allSettled resolution order.
+  const belowMinimumProviderOrder: SwapQuoteProviderName[] = [
```
NeOMakinG, line 316: `preferably-blocking:` Hardcoded provider order doesn't match runtime fetcher array order — The belowMinimumProviderOrder array at lines 311-318 is hardcoded to [KyberSwap, 1inch, LiFi, SwapKit, THORChain, MayaChain]. If the fetchers array order changes in the future (e.g., someone reorders them or adds a new provider), this preference order will diverge from the actual execution order, creating nondeterministic behavior.

fix: Either (a) derive the order from the fetchers array itself: const belowMinimumProviderOrder = fetchers.map(f => f.providerName), or (b) add a runtime assertion that the two arrays are in sync, or (c) document that this order is intentional UI preference, not execution order.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -192,6 +192,65 @@ const isNoRouteError = (message: string) => {
   return normalizedMessage.includes('noroutesfound') || normalizedMessage.includes('noroutes')
 }
 
+const isBelowMinimumError = (message: string) => {
```
NeOMakinG, line 195: `should-fix:` isBelowMinimumError() has rejection token overlap with minimum patterns — The rejection token check at line 207 includes 'below' and the pattern check at line 216 also includes 'below minimum'. This creates double-matching on the same word, though it doesn't cause functional harm. More critically, 'threshold' is both a rejection token AND appears in 'dust threshold' pattern, creating potential confusion in what's considered a pattern vs a token.

fix: Consider refactoring to separate pattern-matching from token-anchoring more clearly. For example, define minimumPatterns = ['below minimum', 'minimum amount', ...] and rejectionTokens = ['rejected', 'failed', ...] at module scope, then check: patterns.some(p => lower.includes(p)) && tokens.some(t => lower.includes(t)). This makes the logic self-documenting.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -300,10 +300,63 @@ export const findSwapQuote = async ({
...
+    'KyberSwap',
+    '1inch',
+    'LiFi',
+    'SwapKit',
+    'THORChain',
+    'MayaChain',
+  ]
+
+  const isBelowMinimumMsg = (msg: string) => {
```
NeOMakinG, line 325: `should-fix:` isBelowMinimumMsg() duplicates logic from isBelowMinimumError() in getSwapKitQuote.ts — The function isBelowMinimumMsg() at lines 320-329 reimplements the same pattern-matching logic as isBelowMinimumError() in getSwapKitQuote.ts, but WITHOUT the rejection token anchoring. This creates divergence: SwapKit checks for rejection tokens, but findSwapQuote doesn't. A message like 'minimum amount of gas used' (success context) could be caught here as a below-minimum error when it shouldn't be.

fix: Either (a) export isBelowMinimumError from getSwapKitQuote.ts and reuse it here, or (b) add the same rejection token anchoring logic to isBelowMinimumMsg().

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -192,6 +192,65 @@ const isNoRouteError = (message: string) => {
...
+    lower.includes('dust threshold') ||
+    lower.includes('below the minimum')
+  )
+}
+
+const isBelowMinimumErrorCode = (errorCode: string | undefined): boolean =>
+  typeof errorCode === 'string' && errorCode.toUpperCase().includes('BELOW_MINIMUM')
+
+/** Extracts the first below-minimum signal from providerErrors, if any. */
+const extractBelowMinimumProviderError = (errors: SwapKitQuoteResponse['providerErrors']): string | undefined => {
```
NeOMakinG, line 230: `suggestion:` extractBelowMinimumProviderError() could return structured data for richer UI — Currently returns a formatted string like 'CHAINFLIP: Amount below minimum: 0.0003 BTC required'. The UI parsing this string to extract provider name and message is fragile. If SwapKit changes the format or adds more metadata (e.g., actual minimum amount as a number), the UI can't access it.

fix: Consider returning {provider: string, message: string, minimumAmount?: string} instead of string. Let the UI format it. This makes the API more robust to SwapKit schema evolution.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
NeOMakinG: `q:` User experience question: when should below-minimum errors be blocking vs informational? — The current implementation throws an error for any below-minimum signal, even when other routes exist. This assumes that below-minimum information is always more valuable than having a route. Is this the intended behavior? What if the user wants to proceed with the NEAR route even though CHAINFLIP rejected it?

fix: Consider documenting the UX intent in code comments or PR description. If this behavior might change in the future, consider making it configurable (e.g., a flag to throw vs warn).

NeOMakinG (commented):

## testing approach

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | null | fixed |
| 1 | null | fixed |
| 1 | null | fixed |
| 1 | null | fixed |
| 1 | null | fixed |
| 1 | null | fixed |

## preferably-blocking
### 1. Breaking change: below-minimum errors now thrown even when other providers have valid routes — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:569`
Lines 569-575 throw an error for ANY below-minimum provider error, even when other routes exist. This changes the user experience from 'here's a route from NEAR' to 'error: CHAINFLIP says your amount is too small'. Users who don't understand provider economics may think the swap failed entirely rather than having a route option.
action: Either (a) document this as intentional in PR description/changelog, or (b) make this non-blocking by attaching the below-minimum message as a warning field to the returned quote instead of throwing. Consider user intent: if they explicitly selected 'show all routes even if some fail', throwing is wrong.

### 2. Hardcoded provider order doesn't match runtime fetcher array order — `packages/core/chain/swap/quote/findSwapQuote.ts:311`
The belowMinimumProviderOrder array at lines 311-318 is hardcoded to [KyberSwap, 1inch, LiFi, SwapKit, THORChain, MayaChain]. If the fetchers array order changes in the future (e.g., someone reorders them or adds a new provider), this preference order will diverge from the actual execution order, creating nondeterministic behavior.
action: Either (a) derive the order from the fetchers array itself: const belowMinimumProviderOrder = fetchers.map(f => f.providerName), or (b) add a runtime assertion that the two arrays are in sync, or (c) document that this order is intentional UI preference, not execution order.

## should-fix
### 1. isBelowMinimumError() has rejection token overlap with minimum patterns — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:195`
The rejection token check at line 207 includes 'below' and the pattern check at line 216 also includes 'below minimum'. This creates double-matching on the same word, though it doesn't cause functional harm. More critically, 'threshold' is both a rejection token AND appears in 'dust threshold' pattern, creating potential confusion in what's considered a pattern vs a token.
action: Consider refactoring to separate pattern-matching from token-anchoring more clearly. For example, define minimumPatterns = ['below minimum', 'minimum amount', ...] and rejectionTokens = ['rejected', 'failed', ...] at module scope, then check: patterns.some(p => lower.includes(p)) && tokens.some(t => lower.includes(t)). This makes the logic self-documenting.

### 2. isBelowMinimumMsg() duplicates logic from isBelowMinimumError() in getSwapKitQuote.ts — `packages/core/chain/swap/quote/findSwapQuote.ts:320`
The function isBelowMinimumMsg() at lines 320-329 reimplements the same pattern-matching logic as isBelowMinimumError() in getSwapKitQuote.ts, but WITHOUT the rejection token anchoring. This creates divergence: SwapKit checks for rejection tokens, but findSwapQuote doesn't. A message like 'minimum amount of gas used' (success context) could be caught here as a below-minimum error when it shouldn't be.
action: Either (a) export isBelowMinimumError from getSwapKitQuote.ts and reuse it here, or (b) add the same rejection token anchoring logic to isBelowMinimumMsg().

## suggestion
### 1. extractBelowMinimumProviderError() could return structured data for richer UI — `packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:230`
Currently returns a formatted string like 'CHAINFLIP: Amount below minimum: 0.0003 BTC required'. The UI parsing this string to extract provider name and message is fragile. If SwapKit changes the format or adds more metadata (e.g., actual minimum amount as a number), the UI can't access it.
action: Consider returning {provider: string, message: string, minimumAmount?: string} instead of string. Let the UI format it. This makes the API more robust to SwapKit schema evolution.

## q
- User experience question: when should below-minimum errors be blocking vs informational?

## risk
Low to medium — the breaking change at line 569 (throwing errors even when routes exist) could confuse users who expect to see available routes. Hardcoded provider order could drift from fetcher order in future maintenance.

## verdict
COMMENTED

NeOMakinG (approved):

## prior issues — status
| round | issue | status |
|---|---|---|
| 2 | null | fixed |
| 2 | null | fixed |
| 2 | null | fixed |
| 2 | null | fixed |
| 2 | null | fixed |
| 2 | null | fixed |
| 3 | null | fixed |
| 3 | null | fixed |

## risk
Low — targeted fix for UX regression. Only affects the specific case where SwapKit returns both valid routes and providerErrors. Behavior now matches user expectations: if a route exists, return it instead of blocking with an error from a different provider.

## verdict
APPROVE — ship it 🫡

NeOMakinG (approved):

LGTM ✅

PR #530, opened 2026-05-23
feat(seedphrase): probe Cosmos-coin-type Terra path for Keplr/Leap seeds

NeOMakinG:
## Summary

- Part A of vultisig/vultiagent-app#449. Adds a discovery probe so seeds exported from Keplr/Leap (which use coin type 118 for Terra) are not reported as empty.
- Mirrors the existing Phantom-Solana dual-path pattern in `ChainDiscoveryService.ts`.
- Default behavior unchanged: 330-path remains preferred when it has balance. The 118-path flag flips only when 330 has zero balance AND 118 has non-zero balance.

## Files

- `packages/sdk/src/seedphrase/MasterKeyDeriver.ts` — adds `deriveTerraAddressWithCosmosPath` and `cosmosPathTerra` constant; extends `DeriveChainPrivateKeysOptions` with `useCosmosPathTerra`.
- `packages/sdk/src/seedphrase/ChainDiscoveryService.ts` — adds the 118-path probe (`checkCosmosPathTerraBalance`) parallel to Phantom Solana.
- `packages/sdk/src/seedphrase/types.ts` — adds `useCosmosPathTerra: boolean` to `ChainDiscoveryAggregate`, `CreateFastVaultFromSeedphraseOptions`, `CreateSecureVaultFromSeedphraseOptions`, `JoinSecureVaultOptions`.
- `packages/sdk/src/seedphrase/prepareSeedphraseImportPrelude.ts` — threads `useCosmosPathTerra` through prelude input/result.
- `packages/sdk/src/services/FastVaultFromSeedphraseService.ts` — passes `useCosmosPathTerra` to prelude and `deriveChainPrivateKeys`.
- `packages/sdk/src/services/SecureVaultFromSeedphraseService.ts` — same as above.
- `.changeset/feat-terra-cosmos-path-discovery-probe.md` — minor bump for @vultisig/sdk.

## Test plan

- [x] Unit: `deriveTerraAddressWithCosmosPath` calls `getKey(terraV2, "m/44'/118'/0'/0/0")`, uses secp256k1 compressed pubkey, cleans up objects.
- [x] Unit: probe flag=false when 330-path Terra has balance (no false switch).
- [x] Unit: probe flag=true when 330=0 and 118>0 (Keplr/Leap case).
- [x] Unit: probe flag=false when both are zero (no flip on empty wallet).
- [x] Unit: probe flag=false + address stays on 330 when both have balance (prefer native).
- [x] `yarn typecheck` — clean.
[...]

PR #539, opened 2026-05-24
fix(seedphrase): probe TerraClassic alongside Terra for cosmos-path detection

gomesalexandre:
Refs vultisig/vultisig-sdk#530 (post-merge CR follow-up)

## what

`ChainDiscoveryService` only probed the Cosmos-coin-type path (m/44'/118'/0'/0/0) when `Chain.Terra` appeared in scan results. A `config.chains: [Chain.TerraClassic]` scan (no `Chain.Terra`) could never set `useCosmosPathTerra=true` - the `terraResult` lookup returned `undefined` and the probe was skipped entirely.

**ChainDiscoveryService.ts** - after the existing Terra probe, add a TerraClassic-only branch: fires when `terraResult` is `undefined` AND `terraClassicResult` exists. Same logic (use 118-path when it has balance and the 330-path is empty), updates `Chain.TerraClassic` result.

**MasterKeyDeriver.ts** - add `deriveTerraClassicAddressWithCosmosPath` (mirrors `deriveTerraAddressWithCosmosPath` but uses `CoinType.terra` for classic/LUNC address encoding).

**ChainDiscoveryService.ts** - add `checkCosmosPathTerraClassicBalance` private method (mirrors `checkCosmosPathTerraBalance`, fetches against `Chain.TerraClassic`).

## why

Seedphrase fund-safety: a TerraClassic-only Keplr/Leap seed would show 0 balance and no address on TerraClassic even though the wallet had funds on the 118-path. Users would think the import failed or the funds were lost.

## receipts

```
npm run test:unit — 1400 tests passed (1397 before + 3 new)
npm run typecheck — clean
```

New tests:
- TerraClassic-only 118-path balance: `useCosmosPathTerra=true`, address + balance updated
- TerraClassic-only 330-path has balance: `useCosmosPathTerra=false` (no override)
- Both Terra + TerraClassic in config: only Terra probe fires (no double probe)

## Summary by CodeRabbit

* **New Features**
  * Improved TerraClassic chain discovery with additional Cosmos-path probing and address derivation
  * Added TerraClassic-compatible address derivation for affected wallets

* **Bug Fixes**
  * Better handling and explicit reporting of network/RPC failures during chain discovery

* **Documentation**
  * Clarified Terra-family chain support and Cosmos-path behavior in configuration docs

* **Tests**
  * Expanded coverage for TerraClassic scenarios and transport-failure handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/539?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/sdk/src/seedphrase/ChainDiscoveryService.ts
+++ b/packages/sdk/src/seedphrase/ChainDiscoveryService.ts
@@ -176,12 +176,15 @@ export class ChainDiscoveryService {
       }
     }
 
-    // Check Cosmos-coin-type Terra path (m/44'/118'/0'/0/0) if Terra was in the scan.
-    // Keplr and Leap historically derived Terra under coin type 118 instead of 330.
-    // Use the 118 path when it has balance AND the standard 330 path has no balance.
+    // Check Cosmos-coin-type Terra/TerraClassic path (m/44'/118'/0'/0/0) if either
+    // chain was in the scan. Keplr and Leap historically derived Terra/TerraClassic
+    // under coin type 118 instead of their native SLIP-44 paths.
+    // Use the 118 path when it has balance AND the standard path has no balance.
     let useCosmosPathTerra = false
     const terraResult = results.find(r => r.chain === Chain.Terra)
+    const terraClassicResult = results.find(r => r.chain === Chain.TerraClassic)
 
+    // Probe Terra (Luna v2) if it was in the scan.
     if (terraResult) {
       try {
         const cosmosPathCheck = await this.checkCosmosPathTerraBalance(mnemonic, timeoutPerChain)
@@ -202,6 +205,32 @@ export class ChainDiscoveryService {
       } catch (error) {
         // Cosmos-path Terra check failed, continue with standard 330 path
         console.warn('Failed to check Cosmos-path Terra address:', error)
+      }
+    }
+
+    // Probe TerraClassic (LUNC) if it was in the scan but Terra was not.
+    // Without this, config.chains: [Chain.TerraClassic] seeds created in Keplr/Leap
+    // could never have useCosmosPathTerra set, despite the Terra/TerraClassic contract
+    // documented in the public types. (sdk#530 post-merge CR follow-up.)
+    if (!terraResult && terraClassicResult) {
+      try {
+        const cosmosPathCheck = await this.checkCosmosPathTerraClassicBalance(mnemonic, timeoutPerChain)
+        const standard330Balance = BigInt(terraClassicResult.balance || '0')
+
+        // Same rule: use 118 path when it has balance and the standard path does not
+        useCosmosPathTerra = cosmosPathCheck.balance > 0n && standard330Balance === 0n
+
+        if (useCosmosPathTerra) {
+          terraClassicResult.address = cosmosPathCheck.address
+          terraClassicResult.balance = cosmosPathCheck.balance.toString()
+          terraClassicResult.hasBalance = true
+          if (!chainsWithBalance.includes(Chain.TerraClassic)) {
+            chainsWithBalance.push(Chain.TerraClassic)
+          }
+        }
+      } catch (error) {
+        // Cosmos-path TerraClassic check failed, continue with standard path
+        console.warn('Failed to check Cosmos-path TerraClassic address:', error)
       }
     }
 
@@ -343,6 +372,36 @@ export class ChainDiscoveryService {
       const address = await this.keyDeriver.deriveTerraAddressWithCosmosPath(mnemonic)
       const balance = await getCoinBalance({
         chain: Chain.Terra,
+        address,
+      })
+      return { address, balance }
+    }
+
+    return Promise.race([checkPromise(), timeoutPromise]).finally(() => {
+      if (timeoutId) clearTimeout(timeoutId)
+    })
+  }
+
+  /**
+   * Check TerraClassic (LUNC) balance using the Cosmos coin-type derivation path.
+   *
+   * TerraClassic-only seeds created in Keplr/Leap also use coin type 118 instead
+   * of TerraClassic's native SLIP-44 path. This is the TerraClassic counterpart of
+   * checkCosmosPathTerraBalance. (sdk#530 post-merge CR follow-up.)
+   */
+  private async checkCosmosPathTerraClassicBalance(
+    mnemonic: string,
+    timeout: number
+  ): Promise<{ address: string; balance: bigint }> {
+    let timeoutId: ReturnType<typeof setTimeout> | undefined
+    const timeoutPromise = new Promise<never>((_, reject) => {
+      timeoutId = setTimeout(() => reject(new Error('Timeout checking Cosmos-path TerraClassic address')), timeout)
+    })
+
+    const checkPromise = async () => {
+      const address = await this.keyDeriver.deriveTerraClassicAddressWithCosmosPath(mnemonic)
+      const balance = await getCoinBalance({
+        chain: Chain.TerraClassic,
         address,
       })
       return { address, balance }
--- a/packages/sdk/src/seedphrase/ChainDiscoveryService.ts
+++ b/packages/sdk/src/seedphrase/ChainDiscoveryService.ts
@@ -14,6 +14,24 @@ import { MasterKeyDeriver } from './MasterKeyDeriver'
 import type { ChainDiscoveryAggregate, ChainDiscoveryProgress, ChainDiscoveryResult } from './types'
 
 /**
+ * Thrown when a balance RPC call fails due to a transport-level error
+ * (network timeout, DNS failure, non-2xx HTTP status).
+ *
+ * Callers must distinguish this from a confirmed-zero-balance result:
+ * - TransportError  → RPC unreachable; balance unknown; warn + continue
+ * - probe returns   → RPC responded; balance confirmed (may be zero)
+ */
+export class TransportError extends Error {
+  override readonly name = 'TransportError'
+  constructor(
+    message: string,
+    public readonly cause?: unknown
+  ) {
+    super(message)
+  }
+}
+
+/**
  * Configuration for chain discovery
  */
 export type ChainDiscoveryConfig = {
@@ -226,27 +244,42 @@ export class ChainDiscoveryService {
 
     // Terra (Luna v2) probe runs first if Terra was in scan.
     if (terraResult) {
-      const fired = await this.tryApplyCosmosPath({
-        result: terraResult,
-        chainsWithBalance,
-        chain: Chain.Terra,
-        probe: () => this.checkCosmosPathTerraBalance(mnemonic, timeoutPerChain),
-        logLabel: 'Cosmos-path Terra',
-      })
-      if (fired) return true
+      try {
+        const fired = await this.tryApplyCosmosPath({
+          result: terraResult,
+          chainsWithBalance,
+          chain: Chain.Terra,
+          probe: () => this.checkCosmosPathTerraBalance(mnemonic, timeoutPerChain),
+          logLabel: 'Cosmos-path Terra',
+        })
+        if (fired) return true
+      } catch (error) {
+        // TransportError: RPC unreachable. Balance unknown — warn and continue.
+        // Do NOT treat as zero balance; the user may have funds on this path.
+        console.warn(`Cosmos-path Terra balance check failed (transport):`, { chain: Chain.Terra, path: '118', error })
+      }
     }
 
     // TerraClassic (LUNC) probe only runs if Terra wasn't in scan, so the contract
     // (Terra-first preference) is preserved. (sdk#530 post-merge CR follow-up.)
     if (!terraResult && terraClassicResult) {
-      const fired = await this.tryApplyCosmosPath({
-        result: terraClassicResult,
-        chainsWithBalance,
-        chain: Chain.TerraClassic,
-        probe: () => this.checkCosmosPathTerraClassicBalance(mnemonic, timeoutPerChain),
-        logLabel: 'Cosmos-path TerraClassic',
-      })
-      if (fired) return true
+      try {
+        const fired = await this.tryApplyCosmosPath({
+          result: terraClassicResult,
+          chainsWithBalance,
+          chain: Chain.TerraClassic,
+          probe: () => this.checkCosmosPathTerraClassicBalance(mnemonic, timeoutPerChain),
+          logLabel: 'Cosmos-path TerraClassic',
+        })
+        if (fired) return true
+      } catch (error) {
+        // TransportError: RPC unreachable. Balance unknown — warn and continue.
+        console.warn(`Cosmos-path TerraClassic balance check failed (transport):`, {
+          chain: Chain.TerraClassic,
+          path: '118',
+          error,
+        })
+      }
     }
 
     return false
@@ -255,6 +288,10 @@ export class ChainDiscoveryService {
   /**
    * Run a single Cosmos-path probe and apply the result if the 118 path has
    * balance and the 330 path doesn't. Returns true when the swap fires.
+   *
+   * Throws {@link TransportError} when the RPC call fails (timeout, network,
+   * non-2xx). Only returns false on a confirmed zero balance. Callers that
+   * want best-effort behaviour (warn + continue) should catch TransportError.
    */
   private async tryApplyCosmosPath(args: {
     result: ChainDiscoveryResult
@@ -264,22 +301,25 @@ export class ChainDiscoveryService {
     logLabel: string
   }): Promise<boolean> {
     const { result, chainsWithBalance, chain, probe, logLabel } = args
+    let cosmosPathCheck: { address: string; balance: bigint }
     try {
-      const cosmosPathCheck = await probe()
-      const standard330Balance = BigInt(result.balance || '0')
-      const fired = cosmosPathCheck.balance > 0n && standard330Balance === 0n
-      if (!fired) return false
-      result.address = cosmosPathCheck.address
-      result.balance = cosmosPathCheck.balance.toString()
-      result.hasBalance = true
-      if (!chainsWithBalance.includes(chain)) {
-        chainsWithBalance.push(chain)
-      }
-      return true
+      cosmosPathCheck = await probe()
     } catch (error) {
-      console.warn(`Failed to check ${logLabel} address:`, error)
-      return false
+      // Re-throw as TransportError so callers can distinguish RPC failure
+      // from a confirmed zero balance. The original error is preserved as cause.
+      throw new TransportError(`${logLabel} RPC unreachable`, error)
     }
+
+    const standard330Balance = BigInt(result.balance || '0')
+    const fired = cosmosPathCheck.balance > 0n && standard330Balance === 0n
+    if (!fired) return false
+    result.address = cosmosPathCheck.address
+    result.balance = cosmosPathCheck.balance.toString()
+    result.hasBalance = true
+    if (!chainsWithBalance.includes(chain)) {
+      chainsWithBalance.push(chain)
+    }
+    return true
   }
 
   /**
--- a/packages/sdk/src/seedphrase/ChainDiscoveryService.ts
+++ b/packages/sdk/src/seedphrase/ChainDiscoveryService.ts
@@ -180,59 +180,16 @@ export class ChainDiscoveryService {
     // chain was in the scan. Keplr and Leap historically derived Terra/TerraClassic
     // under coin type 118 instead of their native SLIP-44 paths.
     // Use the 118 path when it has balance AND the standard path has no balance.
-    let useCosmosPathTerra = false
     const terraResult = results.find(r => r.chain === Chain.Terra)
     const terraClassicResult = results.find(r => r.chain === Chain.TerraClassic)
 
-    // Probe Terra (Luna v2) if it was in the scan.
-    if (terraResult) {
-      try {
-        const cosmosPathCheck = await this.checkCosmosPathTerraBalance(mnemonic, timeoutPerChain)
-        const standard330Balance = BigInt(terraResult.balance || '0')
-
-        // Use Cosmos path if it has balance AND standard 330-path has no balance
-        useCosmosPathTerra = cosmosPathCheck.balance > 0n && standard330Balance === 0n
-
-        // If 118 path has balance but 330 doesn't, update the Terra result
-        if (useCosmosPathTerra) {
-          terraResult.address = cosmosPathCheck.address
-          terraResult.balance = cosmosPathCheck.balance.toString()
-          terraResult.hasBalance = true
-          if (!chainsWithBalance.includes(Chain.Terra)) {
-            chainsWithBalance.push(Chain.Terra)
-          }
-        }
-      } catch (error) {
-        // Cosmos-path Terra check failed, continue with standard 330 path
-        console.warn('Failed to check Cosmos-path Terra address:', error)
-      }
-    }
-
-    // Probe TerraClassic (LUNC) if it was in the scan but Terra was not.
-    // Without this, config.chains: [Chain.TerraClassic] seeds created in Keplr/Leap
-    // could never have useCosmosPathTerra set, despite the Terra/TerraClassic contract
-    // documented in the public types. (sdk#530 post-merge CR follow-up.)
-    if (!terraResult && terraClassicResult) {
-      try {
-        const cosmosPathCheck = await this.checkCosmosPathTerraClassicBalance(mnemonic, timeoutPerChain)
-        const standard330Balance = BigInt(terraClassicResult.balance || '0')
-
-        // Same rule: use 118 path when it has balance and the standard path does not
-        useCosmosPathTerra = cosmosPathCheck.balance > 0n && standard330Balance === 0n
-
-        if (useCosmosPathTerra) {
-          terraClassicResult.address = cosmosPathCheck.address
-          terraClassicResult.balance = cosmosPathCheck.balance.toString()
-          terraClassicResult.hasBalance = true
-          if (!chainsWithBalance.includes(Chain.TerraClassic)) {
-            chainsWithBalance.push(Chain.TerraClassic)
-          }
-        }
-      } catch (error) {
-        // Cosmos-path TerraClassic check failed, continue with standard path
-        console.warn('Failed to check Cosmos-path TerraClassic address:', error)
-      }
-    }
+    const useCosmosPathTerra = await this.probeCosmosPathForTerraChains({
+      mnemonic,
+      timeoutPerChain,
+      terraResult,
+      terraClassicResult,
+      chainsWithBalance,
+    })
 
     // Report completion
     onProgress?.({
@@ -247,6 +204,81 @@ export class ChainDiscoveryService {
       results,
       usePhantomSolanaPath,
       useCosmosPathTerra,
+    }
+  }
+
+  /**
+   * Probe Terra and TerraClassic for Cosmos-coin-type (m/44'/118') balances.
+   * Extracted from discoverChains to keep cognitive complexity in range.
+   *
+   * Returns true when either chain's Cosmos-path balance is non-zero AND the
+   * standard 330-path balance is zero. Mutates the relevant ChainDiscoveryResult
+   * + appends to chainsWithBalance when the swap fires.
+   */
+  private async probeCosmosPathForTerraChains(args: {
+    mnemonic: string
+    timeoutPerChain: number
+    terraResult: ChainDiscoveryResult | undefined
+    terraClassicResult: ChainDiscoveryResult | undefined
+    chainsWithBalance: Chain[]
+  }): Promise<boolean> {
+    const { mnemonic, timeoutPerChain, terraResult, terraClassicResult, chainsWithBalance } = args
+
+    // Terra (Luna v2) probe runs first if Terra was in scan.
+    if (terraResult) {
+      const fired = await this.tryApplyCosmosPath({
+        result: terraResult,
+        chainsWithBalance,
+        chain: Chain.Terra,
+        probe: () => this.checkCosmosPathTerraBalance(mnemonic, timeoutPerChain),
+        logLabel: 'Cosmos-path Terra',
+      })
+      if (fired) return true
+    }
+
+    // TerraClassic (LUNC) probe only runs if Terra wasn't in scan, so the contract
+    // (Terra-first preference) is preserved. (sdk#530 post-merge CR follow-up.)
+    if (!terraResult && terraClassicResult) {
+      const fired = await this.tryApplyCosmosPath({
+        result: terraClassicResult,
+        chainsWithBalance,
+        chain: Chain.TerraClassic,
+        probe: () => this.checkCosmosPathTerraClassicBalance(mnemonic, timeoutPerChain),
+        logLabel: 'Cosmos-path TerraClassic',
+      })
+      if (fired) return true
+    }
+
+    return false
+  }
+
+  /**
+   * Run a single Cosmos-path probe and apply the result if the 118 path has
+   * balance and the 330 path doesn't. Returns true when the swap fires.
+   */
+  private async tryApplyCosmosPath(args: {
+    result: ChainDiscoveryResult
+    chainsWithBalance: Chain[]
+    chain: Chain
+    probe: () => Promise<{ address: string; balance: bigint }>
+    logLabel: string
+  }): Promise<boolean> {
+    const { result, chainsWithBalance, chain, probe, logLabel } = args
+    try {
+      const cosmosPathCheck = await probe()
+      const standard330Balance = BigInt(result.balance || '0')
+      const fired = cosmosPathCheck.balance > 0n && standard330Balance === 0n
+      if (!fired) return false
+      result.address = cosmosPathCheck.address
+      result.balance = cosmosPathCheck.balance.toString()
+      result.hasBalance = true
+      if (!chainsWithBalance.includes(chain)) {
+        chainsWithBalance.push(chain)
+      }
+      return true
+    } catch (error) {
+      console.warn(`Failed to check ${logLabel} address:`, error)
+      return false
     }
   }
 
--- a/packages/sdk/src/seedphrase/ChainDiscoveryService.ts
+++ b/packages/sdk/src/seedphrase/ChainDiscoveryService.ts
@@ -254,9 +254,16 @@ export class ChainDiscoveryService {
         })
         if (fired) return true
       } catch (error) {
-        // TransportError: RPC unreachable. Balance unknown — warn and continue.
-        // Do NOT treat as zero balance; the user may have funds on this path.
-        console.warn(`Cosmos-path Terra balance check failed (transport):`, { chain: Chain.Terra, path: '118', error })
+        // Only swallow TransportError (RPC unreachable). Non-transport errors
+        // (WASM init failure, invalid mnemonic) propagate — they indicate a
+        // broken caller state, not a transient network condition.
+        if (!(error instanceof TransportError)) throw error
+        console.warn(`Cosmos-path Terra balance check failed (transport):`, {
+          chain: Chain.Terra,
+          path: '118',
+          error: error.message,
+          cause: error.cause,
+        })
       }
     }
 
@@ -273,11 +280,12 @@ export class ChainDiscoveryService {
         })
         if (fired) return true
       } catch (error) {
-        // TransportError: RPC unreachable. Balance unknown — warn and continue.
+        if (!(error instanceof TransportError)) throw error
         console.warn(`Cosmos-path TerraClassic balance check failed (transport):`, {
           chain: Chain.TerraClassic,
           path: '118',
-          error,
+          error: error.message,
+          cause: error.cause,
         })
       }
     }
@@ -300,15 +308,12 @@ export class ChainDiscoveryService {
     probe: () => Promise<{ address: string; balance: bigint }>
     logLabel: string
   }): Promise<boolean> {
-    const { result, chainsWithBalance, chain, probe, logLabel } = args
-    let cosmosPathCheck: { address: string; balance: bigint }
-    try {
-      cosmosPathCheck = await probe()
-    } catch (error) {
-      // Re-throw as TransportError so callers can distinguish RPC failure
-      // from a confirmed zero balance. The original error is preserved as cause.
-      throw new TransportError(`${logLabel} RPC unreachable`, error)
-    }
+    const { result, chainsWithBalance, chain, probe } = args
+    // probe() throws TransportError for RPC failures (wrapped by the probe
+    // implementation). Non-transport errors (WASM init, invalid mnemonic) are
+    // NOT TransportErrors and should propagate unmodified so callers apply the
+    // correct recovery semantics instead of silently continuing.
+    const cosmosPathCheck = await probe()
 
     const standard330Balance = BigInt(result.balance || '0')
     const fired = cosmosPathCheck.balance > 0n && standard330Balance === 0n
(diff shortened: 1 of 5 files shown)


packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -221,6 +207,81 @@ export class ChainDiscoveryService {
...
+      const fired = cosmosPathCheck.balance > 0n && standard330Balance === 0n
+      if (!fired) return false
+      result.address = cosmosPathCheck.address
+      result.balance = cosmosPathCheck.balance.toString()
+      result.hasBalance = true
+      if (!chainsWithBalance.includes(chain)) {
+        chainsWithBalance.push(chain)
+      }
+      return true
+    } catch (error) {
```
NeOMakinG: `preferably-blocking:` Error suppression conflates RPC failure with zero balance — users with valid funds won't discover them during flaky RPC outages — The tryApplyCosmosPath method swallows all exceptions and returns false, conflating: network timeout → treated as 'no balance', invalid node RPC URL → treated as 'no balance', actual zero balance → treated as 'no balance'. For seed discovery, this means users with valid 118-path balances won't discover their funds if the RPC provider is flaky. No retry, no distinction between 'checked and found zero' vs 'couldn't check.' The Promise.race timeout pattern also leaves the loser promise unresolved — if checkPromise() takes 30s but timeout is 5s, the 30s promise continues running holding resources. The finally() clears the timer but doesn't cancel the ongoing getCoinBalance request.

fix: Distinguish timeout/transport errors from zero-balance responses (throw on transport errors, only return false on confirmed zero balance). Use AbortController with fetch-based RPC to cancel the loser promise. Consider exponential backoff retry for transport errors.

packages/sdk/src/seedphrase/MasterKeyDeriver.ts
```diff
@@ -341,6 +341,41 @@ export class MasterKeyDeriver {
...
+    const walletCore = await this.wasmProvider.getWalletCore()
+    const cleaned = cleanMnemonic(mnemonic)
+
+    const hdWallet = walletCore.HDWallet.createWithMnemonic(cleaned, '')
+
+    try {
+      // Use terra (classic, LUNC) coin type for address encoding, derive from path 118
+      const coinType = walletCore.CoinType.terra
+      const privateKey = hdWallet.getKey(coinType, cosmosPathTerra)
+      const publicKey = privateKey.getPublicKeySecp256k1(true) // compressed
```
NeOMakinG, line 371: `preferably-blocking:` Unsafe WASM memory cleanup in error path — privateKey/publicKey leak if deriveAddressFromPublicKey throws — In deriveTerraClassicAddressWithCosmosPath (lines 364-366), if walletCore.CoinTypeExt.deriveAddressFromPublicKey throws, privateKey and publicKey leak in the WASM heap. The finally block only cleans up hdWallet. The same pattern exists in deriveTerraAddressWithCosmosPath (line 331) — this is systemic WASM memory unsafety. Unlike deriveTerraAddressWithCosmosPath which has the same bug, this is new code being added.

fix: Move cleanup into a try/finally: 'const privateKey = hdWallet.getKey(coinType, cosmosPathTerra); let publicKey: ReturnType<typeof privateKey.getPublicKeySecp256k1> | undefined; try { publicKey = privateKey.getPublicKeySecp256k1(true); return walletCore.CoinTypeExt.deriveAddressFromPublicKey(coinType, publicKey) } finally { privateKey.delete(); publicKey?.delete() }'. Also fix deriveTerraAddressWithCosmosPath with the same pattern.

NeOMakinG (commented):

## testing approach

## preferably-blocking
### 1. Contract semantics leak: useCosmosPathTerra now covers TerraClassic but name suggests Terra-only — `packages/sdk/src/seedphrase/types.ts:112`
The return contract useCosmosPathTerra: boolean now reports true for either Terra (v2) OR TerraClassic, but the JSDoc comment says 'True when the 330-path has zero balance AND the 118-path has non-zero balance. Covers seeds originally exported from Keplr or Leap.' Callers interpreting this as 'Terra (LUNA) used Cosmos path' will be misled when it's actually TerraClassic (LUNC) that triggered it. The flag propagates to MasterKeyDeriver where it's applied to both Terra AND TerraClassic (line 212: chain === 'Terra' || chain === 'TerraClassic'), creating a mismatch between flag name and actual semantics.
action: Rename to useCosmosPathForTerraFamily or split into two flags (useCosmosPathTerra, useCosmosPathTerraClassic). The current naming leak will cause downstream consumers to make incorrect UI/routing decisions based on a flag that means something different than its name suggests.

### 2. Error suppression conflates RPC failure with zero balance — users with valid funds won't discover them during flaky RPC outages — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:279`
The tryApplyCosmosPath method swallows all exceptions and returns false, conflating: network timeout → treated as 'no balance', invalid node RPC URL → treated as 'no balance', actual zero balance → treated as 'no balance'. For seed discovery, this means users with valid 118-path balances won't discover their funds if the RPC provider is flaky. No retry, no distinction between 'checked and found zero' vs 'couldn't check.' The Promise.race timeout pattern also leaves the loser promise unresolved — if checkPromise() takes 30s but timeout is 5s, the 30s promise continues running holding resources. The finally() clears the timer but doesn't cancel the ongoing getCoinBalance request.
action: Distinguish timeout/transport errors from zero-balance responses (throw on transport errors, only return false on confirmed zero balance). Use AbortController with fetch-based RPC to cancel the loser promise. Consider exponential backoff retry for transport errors.

### 3. Unsafe WASM memory cleanup in error path — privateKey/publicKey leak if deriveAddressFromPublicKey throws — `packages/sdk/src/seedphrase/MasterKeyDeriver.ts:365`
In deriveTerraClassicAddressWithCosmosPath (lines 364-366), if walletCore.CoinTypeExt.deriveAddressFromPublicKey throws, privateKey and publicKey leak in the WASM heap. The finally block only cleans up hdWallet. The same pattern exists in deriveTerraAddressWithCosmosPath (line 331) — this is systemic WASM memory unsafety. Unlike deriveTerraAddressWithCosmosPath which has the same bug, this is new code being added.
action: Move cleanup into a try/finally: 'const privateKey = hdWallet.getKey(coinType, cosmosPathTerra); let publicKey: ReturnType<typeof privateKey.getPublicKeySecp256k1> | undefined; try { publicKey = privateKey.getPublicKeySecp256k1(true); return walletCore.CoinTypeExt.deriveAddressFromPublicKey(coinType, publicKey) } finally { privateKey.delete(); publicKey?.delete() }'. Also fix deriveTerraAddressWithCosmosPath with the same pattern.

## should-fix
### 1. Observability gap: console.warn lacks structured context for telemetry debugging — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:280`
console.warn(`Failed to check ${logLabel} address:`, error) embeds a raw error object but lacks structured context for telemetry. No correlation ID, no chain identifier, no indication of which mnemonic index failed. Production debugging will require manual log stitching across multiple warnings.
action: At minimum, include chain and whether this was the 118-path probe in a structured log object: console.warn(JSON.stringify({ event: 'cosmosPathProbeFailed', chain, path: 'cosmos-118', error: error.message })) or use a proper logging library.

## suggestion
### 1. Document the 'Terra-first preference' business rule in architecture decision record — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:239`
The 'Terra-first' contract is enforced via guard 'if (!terraResult && terraClassicResult)' but the comment only references a post-merge CR follow-up (sdk#530). This suggests the contract was debated but not codified in a spec. Future maintainers may not understand why terraResult blocks terraClassicResult. The business rule (why LUNA takes precedence over LUNC) should be documented — is this a Keplr/Leap compatibility constraint or a UX choice?
action: Add an ADR to docs/ or skills/common/architecture/ explaining the 'Terra-first preference' rule: 'When both Terra and TerraClassic are discovered, Terra takes precedence for Cosmos-path probing because X reason.' Include the context from sdk#530 for future reference.

## risk
High — this affects seedphrase fund-safety. If the contract semantics leak causes downstream consumers to misinterpret useCosmosPathTerra, users could be shown incorrect addresses or balances. Additionally, the error suppression pattern means users with valid funds during RPC outages will see 0 balance and think their funds are lost.

## verdict
COMMENTED

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -221,6 +207,81 @@ export class ChainDiscoveryService {
...
+      console.warn(`Failed to check ${logLabel} address:`, error)
```
NeOMakinG: `should-fix:` Observability gap: console.warn lacks structured context for telemetry debugging — console.warn(`Failed to check ${logLabel} address:`, error) embeds a raw error object but lacks structured context for telemetry. No correlation ID, no chain identifier, no indication of which mnemonic index failed. Production debugging will require manual log stitching across multiple warnings.

fix: At minimum, include chain and whether this was the 118-path probe in a structured log object: console.warn(JSON.stringify({ event: 'cosmosPathProbeFailed', chain, path: 'cosmos-118', error: error.message })) or use a proper logging library.

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -221,6 +207,81 @@ export class ChainDiscoveryService {
...
+        result: terraResult,
+        chainsWithBalance,
+        chain: Chain.Terra,
+        probe: () => this.checkCosmosPathTerraBalance(mnemonic, timeoutPerChain),
+        logLabel: 'Cosmos-path Terra',
+      })
+      if (fired) return true
+    }
+
+    // TerraClassic (LUNC) probe only runs if Terra wasn't in scan, so the contract
```
NeOMakinG, line 270: `suggestion:` Document the 'Terra-first preference' business rule in architecture decision record — The 'Terra-first' contract is enforced via guard 'if (!terraResult && terraClassicResult)' but the comment only references a post-merge CR follow-up (sdk#530). This suggests the contract was debated but not codified in a spec. Future maintainers may not understand why terraResult blocks terraClassicResult. The business rule (why LUNA takes precedence over LUNC) should be documented — is this a Keplr/Leap compatibility constraint or a UX choice?

fix: Add an ADR to docs/ or skills/common/architecture/ explaining the 'Terra-first preference' rule: 'When both Terra and TerraClassic are discovered, Terra takes precedence for Cosmos-path probing because X reason.' Include the context from sdk#530 for future reference.

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
NeOMakinG: `preferably-blocking:` Error suppression conflates RPC failure with zero balance — users with valid funds won't discover them during outages — tryApplyCosmosPath swallows all exceptions and returns false. Network timeout → 'no balance', invalid RPC URL → 'no balance', actual zero balance → 'no balance'. Promise.race timeout leaves the loser promise unresolved (if checkPromise takes 30s but timeout is 5s, the 30s promise keeps running). finally() only clears the timer.

fix: Distinguish transport errors from zero-balance (throw on transport errors, only return false on confirmed zero). Use AbortController with fetch-based RPC to cancel loser promise. Consider exponential backoff retry for transport errors.

packages/sdk/src/seedphrase/MasterKeyDeriver.ts
```diff
@@ -341,6 +341,41 @@ export class MasterKeyDeriver {
...
+  async deriveTerraClassicAddressWithCosmosPath(mnemonic: string): Promise<string> {
+    const walletCore = await this.wasmProvider.getWalletCore()
+    const cleaned = cleanMnemonic(mnemonic)
+
+    const hdWallet = walletCore.HDWallet.createWithMnemonic(cleaned, '')
+
+    try {
+      // Use terra (classic, LUNC) coin type for address encoding, derive from path 118
+      const coinType = walletCore.CoinType.terra
+      const privateKey = hdWallet.getKey(coinType, cosmosPathTerra)
```
NeOMakinG, line 370: `preferably-blocking:` WASM memory cleanup leak — privateKey/publicKey not deleted if deriveAddressFromPublicKey throws — In deriveTerraClassicAddressWithCosmosPath (lines 364-366), if walletCore.CoinTypeExt.deriveAddressFromPublicKey throws on line 366, privateKey and publicKey created on lines 364-365 leak. The finally block only cleans up hdWallet. Same bug exists in deriveTerraAddressWithCosmosPath.

fix: Move cleanup into try/finally: 'let publicKey: ReturnType<typeof privateKey.getPublicKeySecp256k1> | undefined; try { publicKey = privateKey.getPublicKeySecp256k1(true); return walletCore.CoinTypeExt.deriveAddressFromPublicKey(coinType, publicKey) } finally { privateKey.delete(); publicKey?.delete() }'. Fix both methods.

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
NeOMakinG: `should-fix:` console.warn lacks structured context for telemetry debugging — The error message lacks chain identifier, correlation ID, or which path (118 vs 330) failed. Production debugging will require manual log stitching.

fix: Include chain and path in structured log: console.warn(JSON.stringify({ event: 'cosmosPathProbeFailed', chain, path: 'cosmos-118', error: error.message })).

NeOMakinG (changes requested):

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | null | partial |
| 1 | null | still_blocking |
| 1 | null | still_blocking |

## preferably-blocking
### 1. JSDoc inconsistency — useCosmosPathTerra comment says 'for Terra' but applies to TerraClassic too — `packages/sdk/src/seedphrase/MasterKeyDeriver.ts:51`
You updated types.ts to document that useCosmosPathTerra covers the Terra family, but the JSDoc on line 51 still says 'for Terra instead of native 330 path'. Callers reading DeriveChainPrivateKeysOptions will be misled.
action: Update line 51 JSDoc to: 'Use Cosmos coin-type derivation path (m/44'/118'/0'/0/0) for Terra/TerraClassic instead of native SLIP-44 paths. Applies to both chains when true.'

### 2. Error suppression conflates RPC failure with zero balance — users with valid funds won't discover them during outages — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:279`
tryApplyCosmosPath swallows all exceptions and returns false. Network timeout → 'no balance', invalid RPC URL → 'no balance', actual zero balance → 'no balance'. Promise.race timeout leaves the loser promise unresolved (if checkPromise takes 30s but timeout is 5s, the 30s promise keeps running). finally() only clears the timer.
action: Distinguish transport errors from zero-balance (throw on transport errors, only return false on confirmed zero). Use AbortController with fetch-based RPC to cancel loser promise. Consider exponential backoff retry for transport errors.

### 3. WASM memory cleanup leak — privateKey/publicKey not deleted if deriveAddressFromPublicKey throws — `packages/sdk/src/seedphrase/MasterKeyDeriver.ts:364`
In deriveTerraClassicAddressWithCosmosPath (lines 364-366), if walletCore.CoinTypeExt.deriveAddressFromPublicKey throws on line 366, privateKey and publicKey created on lines 364-365 leak. The finally block only cleans up hdWallet. Same bug exists in deriveTerraAddressWithCosmosPath.
action: Move cleanup into try/finally: 'let publicKey: ReturnType<typeof privateKey.getPublicKeySecp256k1> | undefined; try { publicKey = privateKey.getPublicKeySecp256k1(true); return walletCore.CoinTypeExt.deriveAddressFromPublicKey(coinType, publicKey) } finally { privateKey.delete(); publicKey?.delete() }'. Fix both methods.

## should-fix
### 1. console.warn lacks structured context for telemetry debugging — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:280`
The error message lacks chain identifier, correlation ID, or which path (118 vs 330) failed. Production debugging will require manual log stitching.
action: Include chain and path in structured log: console.warn(JSON.stringify({ event: 'cosmosPathProbeFailed', chain, path: 'cosmos-118', error: error.message })).

## risk
High — affects seedphrase fund-safety. WASM memory leak is a long-term resource issue. Error suppression means users with valid funds during RPC outages see 0 balance and think funds are lost. Documentation inconsistency could cause incorrect routing decisions.

## verdict
REQUEST CHANGES

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
gomesalexandre: Tackled in f73967bf.

Introduced `TransportError` as a named exported class. `tryApplyCosmosPath` now throws it when the probe rejects (any network/timeout/5xx) instead of swallowing it and returning `false`. `probeCosmosPathForTerraChains` catches `TransportError`, logs a structured warn `{ chain, path: '118', error }` for telemetry correlation, and continues — so a flaky RPC outage on the 118-path doesn't silently report "no balance" for users who may have funds there.

Regarding AbortController: `getCoinBalance` doesn't accept an AbortSignal today so we can't cancel the loser promise at the fetch layer without a separate PR touching core. The `finally()` already clears the timer (prevents the timer-side leak); the in-flight fetch just completes silently in the background and its result is discarded — which is the correct behaviour for a best-effort background probe. Happy to add a follow-up issue for `getCoinBalance({ ..., signal }` if you think it's worth it.

3 new tests added: TransportError shape, Terra transport error, TerraClassic transport error, plus a silent-zero-balance path that asserts no spurious warn.

gomesalexandre: Tackled in f73967bf — see reply on the r2 thread.

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -221,6 +225,103 @@ export class ChainDiscoveryService {
...
+      try {
+        const fired = await this.tryApplyCosmosPath({
+          result: terraResult,
+          chainsWithBalance,
+          chain: Chain.Terra,
+          probe: () => this.checkCosmosPathTerraBalance(mnemonic, timeoutPerChain),
+          logLabel: 'Cosmos-path Terra',
+        })
+        if (fired) return true
+      } catch (error) {
```
NeOMakinG, line 256: `preferably-blocking:` TransportError creates false-negative detection window for legacy Terra/TerraClassic wallets — When 118-path RPC throws TransportError, code returns useCosmosPathTerra=false after logging warning. Silent false-negative when: (1) 330-path balance is zero, (2) 118-path RPC throws (timeout/DNS/5xx). User sees useCosmosPathTerra=false → SDK derives standard 330-path addresses → funds on 118-path not discovered. Legacy Keplr/Leap wallets specifically used 118-path.

fix: Add transportErrors: Chain[] field to ChainDiscoveryAggregate to surface which chains had RPC failures. Document in public API that useCosmosPathTerra=false is inconclusive when transport errors occurred. UI can then prompt 'RPC timeout for Terra — retry discovery?'

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -353,6 +454,36 @@ export class ChainDiscoveryService {
...
+   *
+   * TerraClassic-only seeds created in Keplr/Leap also use coin type 118 instead
+   * of TerraClassic's native SLIP-44 path. This is the TerraClassic counterpart of
+   * checkCosmosPathTerraBalance. (sdk#530 post-merge CR follow-up.)
+   */
+  private async checkCosmosPathTerraClassicBalance(
+    mnemonic: string,
+    timeout: number
+  ): Promise<{ address: string; balance: bigint }> {
+    let timeoutId: ReturnType<typeof setTimeout> | undefined
```
NeOMakinG, line 478: `should-fix:` Promise.race with clearTimeout doesn't abort underlying RPC if checkPromise settles after timeout — Promise.race([checkPromise(), timeoutPromise]).finally(() => { clearTimeout(timeoutId) }) clears timeout but doesn't abort RPC if checkPromise settles after timeoutPromise rejects. Outstanding requests consume sockets/rate limits even though results are discarded.

fix: Use AbortController: const abortController = new AbortController(); const timeoutId = setTimeout(() => abortController.abort(), timeout); try { const balance = await getCoinBalance({ chain, address, signal: abortController.signal }); } finally { clearTimeout(timeoutId); }. Note: Requires threading AbortSignal through getCoinBalance() → getCosmosCoinBalance() → StargateClient.

packages/sdk/tests/unit/seedphrase/ChainDiscoveryService.test.ts
```diff
@@ -212,4 +221,171 @@ describe('ChainDiscoveryService', () => {
...
+  //
+  // tryApplyCosmosPath must NOT conflate RPC failure with confirmed-zero.
+  // When the probe throws (timeout / DNS / 5xx), a TransportError is surfaced;
+  // discoverChains logs a warning and returns useCosmosPathTerra=false rather
+  // than silently swallowing the error and reporting "no balance".
+  // ---------------------------------------------------------------------------
+
+  it('TransportError is a named subclass of Error', () => {
+    const e = new TransportError('rpc dead', new Error('ECONNREFUSED'))
+    expect(e).toBeInstanceOf(Error)
```
NeOMakinG, line 305: `should-fix:` Test isolation leak: warnSpy.mockRestore() called after assertions — If discoverChains() throws unexpectedly before reaching mockRestore(), spy bleeds into subsequent tests causing false positives/negatives. Current tests happen to pass but pattern is fragile.

fix: Wrap test body in try/finally: const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {}); try { /* test assertions */ } finally { warnSpy.mockRestore(); }

NeOMakinG (changes requested):
gm — multi-lane pass: . runtime-tested on . requesting changes.

## testing approach

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | null | fixed |
| 1 | null | fixed |
| 1 | null | fixed |
| 2 | null | fixed |
| 2 | null | fixed |
| 2 | null | fixed |

## preferably-blocking
### 1. TransportError creates false-negative detection window for legacy Terra/TerraClassic wallets — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:256`
When 118-path RPC throws TransportError, code returns useCosmosPathTerra=false after logging warning. Silent false-negative when: (1) 330-path balance is zero, (2) 118-path RPC throws (timeout/DNS/5xx). User sees useCosmosPathTerra=false → SDK derives standard 330-path addresses → funds on 118-path not discovered. Legacy Keplr/Leap wallets specifically used 118-path.
action: Add transportErrors: Chain[] field to ChainDiscoveryAggregate to surface which chains had RPC failures. Document in public API that useCosmosPathTerra=false is inconclusive when transport errors occurred. UI can then prompt 'RPC timeout for Terra — retry discovery?'

## should-fix
### 1. Promise.race with clearTimeout doesn't abort underlying RPC if checkPromise settles after timeout — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:468`
Promise.race([checkPromise(), timeoutPromise]).finally(() => { clearTimeout(timeoutId) }) clears timeout but doesn't abort RPC if checkPromise settles after timeoutPromise rejects. Outstanding requests consume sockets/rate limits even though results are discarded.
action: Use AbortController: const abortController = new AbortController(); const timeoutId = setTimeout(() => abortController.abort(), timeout); try { const balance = await getCoinBalance({ chain, address, signal: abortController.signal }); } finally { clearTimeout(timeoutId); }. Note: Requires threading AbortSignal through getCoinBalance() → getCosmosCoinBalance() → StargateClient.

### 2. Test isolation leak: warnSpy.mockRestore() called after assertions — `packages/sdk/tests/unit/seedphrase/ChainDiscoveryService.test.ts:305`
If discoverChains() throws unexpectedly before reaching mockRestore(), spy bleeds into subsequent tests causing false positives/negatives. Current tests happen to pass but pattern is fragile.
action: Wrap test body in try/finally: const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {}); try { /* test assertions */ } finally { warnSpy.mockRestore(); }

## risk
Medium. TransportError false-negative window could cause users with legacy Keplr/Leap wallets to miss funds during discovery if RPC is unhealthy. Test isolation issue is low-risk (test suite isolation). Promise.race resource leak is low-risk for single-chain discovery but could scale poorly in batch operations.

## verdict
REQUEST CHANGES

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
NeOMakinG, line 478: `should-fix:` Promise.race resource leak persists — outstanding RPC requests continue after timeout — The timeout pattern uses Promise.race([checkPromise(), timeoutPromise]).finally(() => { clearTimeout(timeoutId) }). If checkPromise() takes 30s but timeout is 5s, the RPC request keeps running in background consuming sockets/rate limits. finally() only clears timer, doesn't cancel outstanding RPC.

fix: This is acceptable to defer given cross-cutting scope (would require threading AbortSignal through getCoinBalance → getCosmosCoinBalance → StargateClient). If filing a follow-up, consider a shared withTimeout<T>() utility that centralizes this pattern.

NeOMakinG (approved):

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | null | fixed |
| 1 | null | fixed |
| 1 | null | fixed |
| 1 | null | fixed |
| 2 | null | partial |
| 2 | null | documented |
| 2 | null | fixed |
| 3 | null | documented |

## should-fix
### 1. Promise.race resource leak persists — outstanding RPC requests continue after timeout — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:468`
The timeout pattern uses Promise.race([checkPromise(), timeoutPromise]).finally(() => { clearTimeout(timeoutId) }). If checkPromise() takes 30s but timeout is 5s, the RPC request keeps running in background consuming sockets/rate limits. finally() only clears timer, doesn't cancel outstanding RPC.
action: This is acceptable to defer given cross-cutting scope (would require threading AbortSignal through getCoinBalance → getCosmosCoinBalance → StargateClient). If filing a follow-up, consider a shared withTimeout<T>() utility that centralizes this pattern.

## risk
Low. Changes isolated to seedphrase discovery path, well-tested with 177 lines of new test coverage. Fix is additive (adds TerraClassic probing) and defensive (TransportError distinction). No breaking API changes.

## verdict
APPROVE — ship it 🫡

NeOMakinG (approved):

LGTM ✅

packages/sdk/src/seedphrase/MasterKeyDeriver.ts
```diff
@@ -341,6 +347,51 @@ export class MasterKeyDeriver {
...
+      // Nested try/finally so privateKey + publicKey WASM handles are
+      // deterministically released even if `getPublicKeySecp256k1` or
+      // `deriveAddressFromPublicKey` throws. Without this nested guard, a
+      // throw between `getKey` and the trailing `.delete()` calls would
+      // leak the raw private-key bytes in WASM memory until the next GC
+      // pass — a material confidentiality risk for seed-derived secrets.
+      // (#539 r3 — NeO preferably-blocking.)
+      let publicKey: ReturnType<typeof privateKey.getPublicKeySecp256k1> | null = null
+      try {
```
NeOMakinG: `blocker:` Potential WASM use-after-free in deriveTerraClassicAddressWithCosmosPath — The nested try/finally deletes publicKey BEFORE deriveAddressFromPublicKey returns. If deriveAddressFromPublicKey holds a reference to the publicKey bytes (not a copy), this is use-after-free. The pattern elsewhere captures address, deletes, then returns. Verify WalletCore CoinTypeExt.deriveAddressFromPublicKey copies data, or revert to capture-then-delete pattern.

fix: Either confirm WalletCore copies the key bytes and document this assumption, or refactor to capture the return value first: const address = walletCore.CoinTypeExt.deriveAddressFromPublicKey(coinType, publicKey); publicKey.delete(); return address;

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -221,6 +225,103 @@ export class ChainDiscoveryService {
...
+  private async tryApplyCosmosPath(args: {
+    result: ChainDiscoveryResult
+    chainsWithBalance: Chain[]
+    chain: Chain
+    probe: () => Promise<{ address: string; balance: bigint }>
+    logLabel: string
+  }): Promise<boolean> {
+    const { result, chainsWithBalance, chain, probe, logLabel } = args
+    let cosmosPathCheck: { address: string; balance: bigint }
+    try {
```
NeOMakinG: `blocker:` tryApplyCosmosPath mislabels derivation errors as transport errors — If deriveTerraClassicAddressWithCosmosPath throws (invalid mnemonic, WASM init failure), it gets wrapped as TransportError('RPC unreachable') when it's actually client-side validation. This misattribution in logs/telemetry makes debugging impossible.

fix: Add instanceof check or separate catch blocks for derivation vs RPC errors. Derivation errors should be VaultError or a distinct error class, not TransportError.

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
NeOMakinG, line 478: `preferably-blocking:` Promise.race timeout doesn't cancel orphaned RPC — When timeoutPromise wins, checkPromise continues running with mnemonic-derived keys. This leaves WASM memory holding seed bytes longer than expected and can spawn orphaned RPCs hitting TerraClassic RPC.

fix: Document as unavoidable constraint or add AbortController if getCoinBalance supports it. This was documented as deferred in round 4 APPROVE - confirming this is still acceptable to defer.

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -221,6 +225,103 @@ export class ChainDiscoveryService {
...
+      cosmosPathCheck = await probe()
+    } catch (error) {
+      // Re-throw as TransportError so callers can distinguish RPC failure
+      // from a confirmed zero balance. The original error is preserved as cause.
+      throw new TransportError(`${logLabel} RPC unreachable`, error)
+    }
+
+    const standard330Balance = BigInt(result.balance || '0')
```
NeOMakinG, line 318: `preferably-blocking:` BigInt parsing silently defaults to zero on invalid input — If getCoinBalance returns 'invalid', 'NaN', '', or undefined, BigInt(result.balance || '0') proceeds with 0n, indistinguishable from actual zero.

fix: Parse explicitly and throw on non-numeric strings to catch upstream bugs.

packages/sdk/src/seedphrase/index.ts
```diff
@@ -42,4 +42,4 @@ export { cleanMnemonic, SeedphraseValidator, validateSeedphrase } from './Seedph
 export { type DerivedChainKey, MasterKeyDeriver } from './MasterKeyDeriver'
 
 // Chain Discovery
-export { type ChainDiscoveryConfig, ChainDiscoveryService } from './ChainDiscoveryService'
+export { type ChainDiscoveryConfig, ChainDiscoveryService, TransportError } from './ChainDiscoveryService'
```
NeOMakinG, line 45: `preferably-blocking:` New public API TransportError unversioned — TransportError is now exported from seedphrase module. Consumers might catch and instanceof-check this. No breaking change but should be documented as part of the public API surface.

fix: Document TransportError in seedphrase module's public API exports, or consider versioning if semantically significant.

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -221,6 +225,103 @@ export class ChainDiscoveryService {
...
+          chainsWithBalance,
+          chain: Chain.Terra,
+          probe: () => this.checkCosmosPathTerraBalance(mnemonic, timeoutPerChain),
+          logLabel: 'Cosmos-path Terra',
+        })
+        if (fired) return true
+      } catch (error) {
+        // TransportError: RPC unreachable. Balance unknown — warn and continue.
+        // Do NOT treat as zero balance; the user may have funds on this path.
+        console.warn(`Cosmos-path Terra balance check failed (transport):`, { chain: Chain.Terra, path: '118', error })
```
NeOMakinG: `should-fix:` console.warn lacks error detail — Log at lines 259, 277 includes 'error' object but not error.message. Better debugging context needed.

fix: Update to: console.warn('Cosmos-path Terra balance check failed (transport):', { chain, path: '118', error: error.message, cause: error.cause })

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
```diff
@@ -221,6 +225,103 @@ export class ChainDiscoveryService {
...
+      }
+    }
+
+    // TerraClassic (LUNC) probe only runs if Terra wasn't in scan, so the contract
```
NeOMakinG, line 270: `suggestion:` Terra-first preference only in comment — Comment references 'sdk#530 post-merge CR follow-up'. Future devs won't find this context.

fix: Add @example showing both chains in config or inline test asserting 'TerraClassic probe does NOT fire when Terra v2 is in config'.

packages/sdk/src/seedphrase/MasterKeyDeriver.ts
```diff
@@ -48,7 +48,13 @@ export type DerivedChainKey = {
 export type DeriveChainPrivateKeysOptions = {
   /** Use Phantom wallet derivation path for Solana instead of standard BIP44 path */
   usePhantomSolanaPath?: boolean
-  /** Use Cosmos coin-type derivation path (m/44'/118'/0'/0/0) for Terra instead of native 330 path */
+  /**
```
NeOMakinG, line 51: `suggestion:` useCosmosPathTerra applies to both chains documented in JSDoc but not runtime — The DeriveChainPrivateKeysOptions.useCosmosPathTerra JSDoc says 'applies to BOTH chains when true' but no runtime validation.

fix: Consider runtime validation or assertion to catch caller mistakes.

NeOMakinG (changes requested):

## prior issues — status
| round | issue | status |
|---|---|---|
| 4 | null | partial |
| 4 | null | fixed |
| 1 | null | fixed |
| 2 | null | fixed |

## blocking
### 1. Potential WASM use-after-free in deriveTerraClassicAddressWithCosmosPath — `packages/sdk/src/seedphrase/MasterKeyDeriver.ts:379`
The nested try/finally deletes publicKey BEFORE deriveAddressFromPublicKey returns. If deriveAddressFromPublicKey holds a reference to the publicKey bytes (not a copy), this is use-after-free. The pattern elsewhere captures address, deletes, then returns. Verify WalletCore CoinTypeExt.deriveAddressFromPublicKey copies data, or revert to capture-then-delete pattern.
action: Either confirm WalletCore copies the key bytes and document this assumption, or refactor to capture the return value first: const address = walletCore.CoinTypeExt.deriveAddressFromPublicKey(coinType, publicKey); publicKey.delete(); return address;

### 2. tryApplyCosmosPath mislabels derivation errors as transport errors — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:305`
If deriveTerraClassicAddressWithCosmosPath throws (invalid mnemonic, WASM init failure), it gets wrapped as TransportError('RPC unreachable') when it's actually client-side validation. This misattribution in logs/telemetry makes debugging impossible.
action: Add instanceof check or separate catch blocks for derivation vs RPC errors. Derivation errors should be VaultError or a distinct error class, not TransportError.

## preferably-blocking
### 1. Promise.race timeout doesn't cancel orphaned RPC — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:468`
When timeoutPromise wins, checkPromise continues running with mnemonic-derived keys. This leaves WASM memory holding seed bytes longer than expected and can spawn orphaned RPCs hitting TerraClassic RPC.
action: Document as unavoidable constraint or add AbortController if getCoinBalance supports it. This was documented as deferred in round 4 APPROVE - confirming this is still acceptable to defer.

### 2. BigInt parsing silently defaults to zero on invalid input — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:313`
If getCoinBalance returns 'invalid', 'NaN', '', or undefined, BigInt(result.balance || '0') proceeds with 0n, indistinguishable from actual zero.
action: Parse explicitly and throw on non-numeric strings to catch upstream bugs.

### 3. New public API TransportError unversioned — `packages/sdk/src/seedphrase/index.ts:45`
TransportError is now exported from seedphrase module. Consumers might catch and instanceof-check this. No breaking change but should be documented as part of the public API surface.
action: Document TransportError in seedphrase module's public API exports, or consider versioning if semantically significant.

## should-fix
### 1. Duplicate WASM cleanup pattern — `packages/sdk/src/seedphrase/MasterKeyDeriver.ts:326`
deriveTerraClassicAddressWithCosmosPath mirrors deriveTerraAddressWithCosmosPath except coinType. Future cleanup changes will diverge.
action: Extract sharedWasmDerive(coinType, derivationPath) to reduce duplication.

### 2. console.warn lacks error detail — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:259`
Log at lines 259, 277 includes 'error' object but not error.message. Better debugging context needed.
action: Update to: console.warn('Cosmos-path Terra balance check failed (transport):', { chain, path: '118', error: error.message, cause: error.cause })

## suggestion
### 1. Terra-first preference only in comment — `packages/sdk/src/seedphrase/ChainDiscoveryService.ts:263`
Comment references 'sdk#530 post-merge CR follow-up'. Future devs won't find this context.
action: Add @example showing both chains in config or inline test asserting 'TerraClassic probe does NOT fire when Terra v2 is in config'.

### 2. useCosmosPathTerra applies to both chains documented in JSDoc but not runtime — `packages/sdk/src/seedphrase/MasterKeyDeriver.ts:51`
The DeriveChainPrivateKeysOptions.useCosmosPathTerra JSDoc says 'applies to BOTH chains when true' but no runtime validation.
action: Consider runtime validation or assertion to catch caller mistakes.

## q
- Does WalletCore CoinTypeExt.deriveAddressFromPublicKey internally copy public key bytes, or hold reference? If it holds reference, the finally block publicKey.delete() creates use-after-free. If it copies, document this assumption.
- The comment says 'the MasterKeyDeriver caller checks chain === "Terra" || chain === "TerraClassic"' for useCosmosPathTerra. Which caller? DeriveChainPrivateKeys doesn't appear in this diff. If caller forgets this check, flag applies incorrectly. Consider runtime validation.

## risk
HIGH — Two blocker findings: potential WASM use-after-free (confidentiality risk for seed-derived secrets) and error-flow misattribution (fund-detection failures could be masked). The new Tron-related commits are additive but touch swap tracking which is a critical path. Code touches seedphrase import flow - the most sensitive user-facing path in the SDK.

## verdict
REQUEST CHANGES

packages/sdk/src/seedphrase/MasterKeyDeriver.ts
gomesalexandre: Tackled in a3dd094b.

Switched to capture-then-delete — matches the safe pattern in `deriveTerraAddressWithCosmosPath`:

```ts
const publicKey = privateKey.getPublicKeySecp256k1(true)
const address = walletCore.CoinTypeExt.deriveAddressFromPublicKey(coinType, publicKey)
publicKey.delete()
privateKey.delete()
return address
```

`deriveAddressFromPublicKey` gets the address string before either handle is freed. No reference to the WASM object escapes.

packages/sdk/src/seedphrase/ChainDiscoveryService.ts
gomesalexandre: Tackled in a3dd094b.

The fix pushes TransportError wrapping DOWN to the probe layer (checkCosmosPathTerra/TerraClassic), where the boundary between client-side WASM and network is explicit:

- Derivation (`deriveTerra*AddressWithCosmosPath`) runs BEFORE the race — errors here (invalid mnemonic, WASM init failure) propagate unmodified.
- Only `getCoinBalance` throws and the timeout rejection get wrapped as TransportError.

`tryApplyCosmosPath` just awaits `probe()` and passes through whatever it throws. `probeCosmosPathForTerraChains` now does `instanceof TransportError` check before swallowing — derivation errors propagate to the top-level caller instead of being silently logged as transport noise.

NeOMakinG (approved):
r4 re-review after gomes's [`a3dd094`](https://github.com/vultisig/vultisig-sdk/commit/a3dd094) ("capture address before WASM delete + scope TransportError to RPC layer").

## prior blockers — verified resolved

**Blocker 1 (WASM use-after-free in `deriveTerraClassicAddressWithCosmosPath`)** ✅
The r3 nested try/finally pattern I pushed could have created use-after-free if WalletCore's `deriveAddressFromPublicKey` held a reference rather than copying the publicKey bytes. `a3dd094` reverted to the safer "capture address first, then `.delete()` both handles" pattern that the sibling derivation functions (lines 302-306, 336-340) all use. The trade-off — losing throw-safe cleanup of WASM handles on the rare exception path — is the right call: a one-shot WASM-handle leak is less severe than a potential use-after-free of seed-derived secret material. Good revert.

**Blocker 2 (TransportError misattribution in `tryApplyCosmosPath`)** ✅
`ChainDiscoveryService.ts:256-260` and `:282-283` now scope `TransportError` tightly:
```ts
catch (error) {
  if (!(error instanceof TransportError)) throw error
  // ... warn + continue
}
```
Derivation errors (which won't be TransportError) propagate unmodified. The `TransportError` wrapping is now only done at the RPC-call layer (`:457`, `:491`), so non-RPC failures keep their original error class. Logs/telemetry will correctly distinguish client-side validation from transport issues.

## preferably-blocking / should-fix from r4

- **Promise.race timeout doesn't cancel orphaned RPC** — documented as deferred in round 4; still acceptable.
- **BigInt parsing silently defaults to zero on invalid input** — fair point but very defensive; `getCoinBalance` is internal and `result.balance` is well-typed in practice. Leaving as a non-blocking nit.
- **TransportError unversioned public API** — fair; can be documented in a follow-up PR alongside the rest of the seedphrase module's public-API exports.
- **Duplicate WASM cleanup pattern** — fair extraction candidate but YAGNI for now (2 sites only); promote to shared helper when a 3rd site appears.

None block merge.

LGTM ✅

NeOMakinG (approved):

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | Contract semantics leak - useCosmosPathTerra name suggested Terra-only but applies to TerraClassic | fixed |
| 2 | Error suppression conflated RPC failure with zero balance | fixed |
| 3 | Unsafe WASM memory cleanup - privateKey/publicKey leak on throw | fixed |
| 2 | console.warn lacks structured context for telemetry | fixed |
| 4 | Promise.race timeout doesn't cancel orphaned RPC requests | documented |
| 5 | Potential WASM use-after-free in deriveTerraClassicAddressWithCosmosPath | fixed |
| 5 | tryApplyCosmosPath mislabels derivation errors as transport errors | fixed |

## risk
Low. All prior blockers and preferably-blocking items addressed. Changes isolated to seedphrase discovery path with comprehensive test coverage. WASM memory cleanup is deterministic. Error flow properly distinguishes transport failures from confirmed-zero balance. No breaking API changes.

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.3-codex` conceptual pass. tier1-library-only.

PR #544, opened 2026-05-25
Disable unsupported seedphrase import chains

### LINKED ISSUE #443: [BUG] Bittensor (TAO) derivation: imported seedphrase produces wrong address vs ThorWallet / substrate canonical
# [BUG] Bittensor derivation produces wrong address on seedphrase import (vs ThorWallet)

## Problem

Importing a seed phrase into Vultisig produces a **different Bittensor (TAO) address** than the address derived from the same seed in ThorWallet. The user's funds appear at the ThorWallet-derived address; Vultisig shows an empty wallet at its own (different) derived address.

Same-thread Solana mismatch was already replicated by `@samyap` against ThorWallet, and Polkadot mismatch was reported against TrustWallet — Bittensor is the highest-impact of the three for Vultisig because ThorChain users routinely hold TAO and need round-trippable addresses across the wallets in that ecosystem. Filing the SDK-level fix here so all platforms (Android, iOS, Windows, extension) pick it up at once instead of re-implementing per-platform.

## Expected

Same seed → same Bittensor address as ThorWallet (and Polkadot.js / Subwallet, since Bittensor is a substrate chain and these are the canonical reference implementations).

## Actual

Vultisig derives a different Bittensor address from the same seed. Reporter holds TAO at the ThorWallet-derived address; that balance is invisible from inside Vultisig because Vultisig is looking at a different address.

## Where to look

Bittensor is a substrate chain using **sr25519** by default (ed25519 is also valid but the wallet ecosystem standard for TAO is sr25519). Likely places the divergence lives:

1. **HD scheme**. Substrate doesn't use BIP32. Polkadot/Bittensor wallets derive the key from the **BIP39 mnemonic phrase string** (not the BIP39 seed bytes) via the substrate-specific scheme defined in `sr25519` / `subkey`. If Vultisig's SDK feeds the BIP39 seed bytes (the 64-byte PBKDF2 output) into a BIP32-style derivation, the address will differ from ThorWallet / Subkey / Polkadot.js.
2. **Curve**. Confirm Vultisig is using **sr25519** (not ed25519) for Bittensor by default. ThorWallet uses sr25519.
3. **SS58 prefix**. Bittensor uses SS58 prefix **42** (the generic substrate prefix). If Vultisig encodes with a different prefix (e.g. Polkadot's `0`, Kusama's `2`) the address string will differ even if the underlying public key is right — easy to spot because converting the SS58 to public key bytes would match across wallets.
4. **Derivation path**. Substrate uses soft/hard junctions like `//hard/soft` (not BIP32 `m/44'/x'/0'/0/0`). Confirm Vultisig is applying the empty derivation path (raw mini-secret from the mnemonic) by default, which is what ThorWallet does for the root TAO account.

## Diagnostic

Pick a known test mnemonic (use a publicly documented Polkadot.js test vector, **not** a real funded seed) and:

1. Derive the Bittensor address in Vultisig SDK.
2. Derive the same address in `subkey` (or `polkadot-js/api`) with sr25519, empty derivation path, SS58 prefix 42.
3. Derive in ThorWallet (their public Bittensor flow).
4. Print the SS58 address and the underlying 32-byte public key for each.

If the public key matches and only the SS58 string differs → prefix issue (cheap fix).
If the public key differs → curve, HD scheme, or derivation-path issue.

## Acceptance Criteria

- [ ] Identify exactly where Vultisig SDK's Bittensor derivation diverges from the substrate canonical (sr25519 + mnemonic-string scheme + SS58 prefix 42 + empty derivation path).
- [ ] Align SDK's Bittensor derivation with the substrate canonical so a seed imported from ThorWallet (and Subkey, Polkadot.js, Subwallet) lands on the same Bittensor address.
- [ ] Pinned test vector(s): mnemonic → expected SS58 address, hard-coded, asserted in unit tests. Use a Polkadot.js / subkey public test vector for the assertion source, not an internal-only seed.
- [ ] Backward-compatibility decision: existing Vultisig vaults that were created at the *current* (wrong) derivation already hold TAO on-chain at that address. Spec out whether (a) we migrate them (rederive + sweep), (b) we surface both addresses in UI during a transition window, or (c) we expose a "legacy derivation" toggle. Default recommendation: option (b) for one release, then (a) via a one-tap sweep flow. Block close on this being decided, not implemented.
- [ ] Regression tests covering Bittensor specifically (separate from generic Solana / Polkadot tests).

## Out of scope (track separately)

- Solana derivation mismatch vs ThorWallet (same root cause class but different chain; should be its own SDK issue if not already).
- Polkadot derivation mismatch vs TrustWallet (Polkadot is substrate too and might share the fix; verify after the Bittensor fix lands).
- Custom user-supplied derivation paths as a feature (requested by reporter in same thread).

## Source

- Reporter (Discord): `bigtibbir5`, thread "Big Tibbir" — https://discord.com/channels/1203844257220395078/1501985046305964045
- Reproduced (Solana, same class of bug) by: `@samyap`
- Investigating (Discord): `@neomaking`
- Previously filed and now-closed platform issues (closed in favor of this SDK-level fix): vultisig/vultisig-android#4453, vultisig/vultisig-ios#4340, vultisig/vultisig-windows#3904

cc @realpaaao @rkokhatskyi

rcoderdev:
Closes #443

## Summary by CodeRabbit

* **New Features**
  * Seedphrase import now exposes supported/unsupported chain lists and helper checks, and validates requested chains before import/derivation.

* **Bug Fixes**
  * Disabled seedphrase import for Bittensor to prevent unsupported import attempts and return clear validation errors.

* **Tests**
  * Added/updated unit tests covering chain-support policy, validation behavior, and public exports.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/544?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


gomesalexandre (changes requested):
## what

Adds an SDK-owned seedphrase-import policy: `SEEDPHRASE_IMPORT_UNSUPPORTED_CHAINS` (Cardano, QBTC, Bittensor), a derived supported list, predicate/assertion helpers, and pre-flight `assertSeedphraseImportSupportsChains` guards wired into `ChainDiscoveryService`, `MasterKeyDeriver`, and `prepareSeedphraseImportPrelude`. Also bundles the TerraClassic Cosmos-path discovery work (refactored probe helpers, `TransportError` class).

---

## prior CodeRabbit comments — status

| # | severity | issue | status | evidence |
|---|---|---|---|---|
| CR-1 | major | `throw new Error(...)` instead of `VaultError` in `assertSeedphraseImportSupportsChains` | ✅ fixed | 85f6426 `constants.ts` |
| CR-2 | major | `Chain.Cardano` dropped from `EDDSA_CHAINS`, `isEddsaChain(Cardano)` returns wrong value | ✅ fixed | 85f6426 `ChainDiscoveryService.ts` line 49 |

Both addressed. Good.

---

## preferably-blocking

**`isEddsaChain` in `ChainDiscoveryService` is now permanently stale for the unsupported chains, and the contract of that public method is opaque.**

`EDDSA_CHAINS` is a local constant that has no structural link to the unsupported-chains policy. Cardano was put back (good), but Bittensor is still gone from `EDDSA_CHAINS` even though `signatureAlgorithms.bittensor === 'eddsa'` (see `packages/core/chain/signing/SignatureAlgorithm.ts` line 15). `isEddsaChain(Chain.Bittensor)` returns `false` in the current diff, while every other layer of the SDK (`MasterKeyDeriver.deriveChainPrivateKeys`, `signatureAlgorithms`) correctly classifies it as EdDSA.

That's currently safe only because Bittensor is unsupported for seedphrase import, so `isEddsaChain` is never called with it in this flow. But `isEddsaChain` is a public method - if any caller downstream uses it to decide key derivation for a Bittensor key that arrived through a different path (e.g. non-seedphrase vault), it will mis-classify and potentially derive a wrong key type.

**Action:** either keep `EDDSA_CHAINS` in sync with `signatureAlgorithms` (simplest: derive it from there), or document clearly that `isEddsaChain` is scoped to "chains reachable via seedphrase import" and rename it accordingly (`isSeedphraseImportEddsaChain`). The stale constant is a trap waiting to bite future contributors.

---

## should-fix

**`SEEDPHRASE_IMPORT_UNSUPPORTED_CHAINS` is a plain `Chain[]` - adding a new chain later requires knowing this file exists.**

There's no cross-reference in `Chain` definition, no comment pointing from Bittensor/QBTC/Cardano to this constant, and no validation test that checks the list is complete relative to some known criterion (e.g., "all EdDSA chains are either in supported or unsupported"). If someone adds a new EdDSA chain to the `Chain` enum and forgets to add it here, it silently becomes importable.

**Action:** add a comment next to each blocked chain's entry in `SEEDPHRASE_IMPORT_UNSUPPORTED_CHAINS` explaining *why* it's blocked (e.g., "Bittensor uses Polkadot coin type, signing path not end-to-end verified"), and/or add a test that asserts the list grows if new eddsa/mldsa chains appear.

---

## should-fix

**`QBTC` uses `mldsa` (lattice-based, not EdDSA/ECDSA) — its blocking reason is different from the others and should be called out explicitly.**

Bittensor and Cardano are EdDSA chains whose BIP39 derivation/signing is not yet end-to-end supported. QBTC uses ML-DSA (post-quantum Dilithium), which is a completely different algorithm with its own key material. Lumping them in the same constant without a comment means reviewers have no idea QBTC is there for a fundamentally different reason. This also matters for completeness analysis: "block all non-ECDSA/non-EdDSA chains" is a third category.

**Action:** add inline comments per-entry in `SEEDPHRASE_IMPORT_UNSUPPORTED_CHAINS`. Three distinct reasons: Bittensor (EdDSA, signing not verified), Cardano (EdDSA, custom Cardano public key type not supported in import prelude), QBTC (ML-DSA, entirely different key material).

---

## q

**`prepareSeedphraseImportPrelude` moved `deriveMasterKeys` to after `chainsToImport` resolution + assertion - is there a test covering the "Bittensor in `chains` input, `discoverChains=false`" path through `prepareSeedphraseImportPrelude`?**

I see `MasterKeyDeriver.test.ts` covers the deriver-level assertion and `supportedChains.test.ts` covers the constants, but `prepareSeedphraseImportPrelude.test.ts` - does it explicitly test that passing `chains: [Chain.Bittensor]` without discovery rejects before key derivation? Let me check - if not, that path through the prelude is only guarded by `assertSeedphraseImportSupportsChains` at line 96, which is correct, but the test would be worth adding.

---

## q

**TerraClassic / Terra `probeCosmosPathForTerraChains` logic: if BOTH Terra and TerraClassic are in the scan AND Terra fires (returns `true` early), TerraClassic Cosmos path is never probed. Is that intentional?**

Comment says "TerraClassic probe only runs if Terra wasn't in scan" but the condition is `if (!terraResult && terraClassicResult)`, so if both are scanned and Terra fires, TerraClassic silently stays at the standard 330 path. That seems intentional (prefer Terra over TerraClassic) but the comment only says "Terra wasn't in scan" - it should say "Terra wasn't in scan OR Terra's 118-path fired" for accuracy. Minor but confusing to the next person reading this.

---

## praise

- Enforcement at all three layers (discovery, derivation, prelude) is correct and defense-in-depth. The `assertSeedphraseImportSupportsChains([chain])` in `deriveAddress` specifically is the right call - it prevents a subtle bypass where you skip discovery and call derive directly.
- `VaultError(VaultErrorCode.InvalidConfig, ...)` error type means callers can pattern-match on the code rather than parsing strings. Good.
- `TransportError` being its own exported class with `cause` threading is solid - the WASM-error-vs-transport-error distinction is important and now explicit in the type system.

---

## risk

Low-medium. The blocking logic itself is correct and defense-in-depth. The main risk is the `EDDSA_CHAINS` / `isEddsaChain` stale-for-Bittensor issue which is latent, not immediate. No fund-safety regression introduced - the PR only adds restrictions, never loosens them.

---

## verdict

REQUEST_CHANGES - the `isEddsaChain(Bittensor)` returning `false` while `signatureAlgorithms.bittensor === 'eddsa'` is a real inconsistency in a public method that could bite callers. The should-fix comments are worth addressing before merge but aren't gates.

🤖 reviewed via vultisig-ops (round 1)

gomesalexandre (changes requested):
## prior issues — status

|---|---|---|---|---|---|
| r1 | CR-1 | preferably-blocking | `throw new Error(...)` instead of `VaultError` in `assertSeedphraseImportSupportsChains` | ✅ fixed | 85f6426 `constants.ts:48` — now throws `VaultError(VaultErrorCode.InvalidConfig, ...)` |
| r1 | CR-2 | preferably-blocking | `Chain.Cardano` dropped from `EDDSA_CHAINS`, `isEddsaChain(Cardano)` returns wrong value | ✅ fixed | 85f6426 `ChainDiscoveryService.ts:49` — Cardano back in EDDSA_CHAINS |
| r1 | p-b | `EDDSA_CHAINS` permanently stale for Bittensor — `isEddsaChain(Chain.Bittensor)` returns `false` while `signatureAlgorithms.bittensor === 'eddsa'` | ❌ still blocking | see below |

---

## preferably-blocking

**`isEddsaChain(Chain.Bittensor)` returns `false` — classification contract broken**

Current diff line 49:
```
const EDDSA_CHAINS: Chain[] = [Chain.Solana, Chain.Sui, Chain.Polkadot, Chain.Ton, Chain.Cardano]
```

Bittensor is absent. `signatureAlgorithms.bittensor === 'eddsa'` (SignatureAlgorithm.ts:15). `MasterKeyDeriver.deriveChainPrivateKeys` uses `signatureAlgorithms[chainKind]` to decide key type — it does NOT use `isEddsaChain`. So for the current seedphrase-import flow, skipping Bittensor in `EDDSA_CHAINS` doesn't cause a key mis-derivation.

BUT `isEddsaChain` is a **public method** on `ChainDiscoveryService`. Any caller that relies on it to decide signature algorithm for Bittensor (e.g. a non-seedphrase path, a future migration tool, a consumer of `@vultisig/sdk`) would derive the wrong key type — secp256k1 instead of ed25519.

`EDDSA_CHAINS` and `signatureAlgorithms` are now two sources of truth for the same property that have diverged. The right fix:

- Either add `Chain.Bittensor` back to `EDDSA_CHAINS` (honest classification, Bittensor is EdDSA regardless of whether seedphrase import is supported),
- Or derive `isEddsaChain` from `signatureAlgorithms` instead of maintaining a separate hardcoded array (preferred — single source of truth).

The second option is cleaner:
```ts
// In ChainDiscoveryService (or wherever isEddsaChain lives):
import { getChainKind } from '@vultisig/core-chain/ChainKind'
import { signatureAlgorithms } from '@vultisig/core-chain/signing/SignatureAlgorithm'

isEddsaChain(chain: Chain): boolean {
  const kind = getChainKind(chain)
  return signatureAlgorithms[kind] === 'eddsa'
}
```

The old r1 review said this was "currently safe only because Bittensor is unsupported" — true, but a public method with a wrong return value is a latent footgun. Seed phrase support status and signature algorithm are orthogonal facts. Don't conflate them.

---

## should-fix

**`assertSeedphraseImportSupportsChains` guards in `MasterKeyDeriver` fire before `getWalletCore()` — good. But `prepareSeedphraseImportPrelude` guard is after discovery — what if discovery itself derives a Bittensor key?**

Looking at `prepareSeedphraseImportPrelude.ts`: `assertSeedphraseImportSupportsChains(chainsToImport)` is called after `discoverChains` resolves. `ChainDiscoveryService.discoverChains` now calls `assertSeedphraseImportSupportsChains(chains)` at the top, so if Bittensor is in the default scan list it would throw there first. Default is `SEEDPHRASE_IMPORT_SUPPORTED_CHAINS` which already excludes Bittensor. So the guard in `prepareSeedphraseImportPrelude` fires against the *discovered + filtered* chain list, which won't include Bittensor unless the app passed it explicitly. That explicit case is blocked by `ChainDiscoveryService`'s guard. Layered correctly — this is actually fine.

---

## suggestion

**`SEEDPHRASE_IMPORT_UNSUPPORTED_CHAINS` is a `Chain[]` array, not `readonly`**

Small: exposing a mutable array invites accidental push/mutation by consumers. Consider `as const` or `ReadonlyArray<Chain>`. Same for `SEEDPHRASE_IMPORT_SUPPORTED_CHAINS`.

---

## what's good

- Test coverage is thorough: `prepareSeedphraseImportPrelude.test.ts` covers the "discovered Bittensor balance should not trigger key derivation" case explicitly. Nice.
- Guards fire before `getWalletCore()` in `MasterKeyDeriver` — no WASM load wasted on unsupported chains.
- `VaultError(VaultErrorCode.InvalidConfig, ...)` is the right error shape — CR-1 properly addressed.

gomesalexandre (changes requested):
## prior issues — status

|---|---|---|---|---|---|
| r1 | CR-1 | preferably-blocking | `throw new Error(...)` instead of `VaultError` in `assertSeedphraseImportSupportsChains` | ✅ fixed | 85f6426 `constants.ts` |
| r1 | CR-2 | preferably-blocking | `Chain.Cardano` dropped from `EDDSA_CHAINS`, `isEddsaChain(Cardano)` returns wrong value | ✅ fixed | 85f6426 `ChainDiscoveryService.ts:49` |
| r2 | p-b | `isEddsaChain(Chain.Bittensor)` returns `false` while `signatureAlgorithms.bittensor === 'eddsa'` | ❌ still blocking | see below |
| r2 | should-fix | `SEEDPHRASE_IMPORT_UNSUPPORTED_CHAINS` entries have no inline comments explaining why each chain is blocked | ⏭️ deferred | not addressed in 85f6426 |
| r2 | should-fix | QBTC uses ML-DSA, blocking reason is distinct from EdDSA chains - needs callout | ⏭️ deferred | not addressed in 85f6426 |
| r2 | suggestion | `SEEDPHRASE_IMPORT_SUPPORTED/UNSUPPORTED_CHAINS` should be `readonly` / `as const` | ⏭️ deferred | not addressed |

---

## preferably-blocking (still open from r2)

**`isEddsaChain(Chain.Bittensor)` is `false` — classification is wrong and now test-asserted**

`signatureAlgorithms.bittensor === 'eddsa'` — this is the canonical source of truth in `SignatureAlgorithm.ts`. `EDDSA_CHAINS` in `ChainDiscoveryService.ts` still excludes Bittensor. The test added in 85f6426 explicitly cements the wrong value:

```ts
expect(s.isEddsaChain(Chain.Bittensor)).toBe(false)
```

That assertion is encoding a bug as expected behavior. `isEddsaChain` is a public method. If any caller (migration tool, non-seedphrase vault path, SDK consumer) uses it to decide key derivation for Bittensor, it will route to secp256k1 instead of ed25519.

The fix is a one-liner - pick one of:

**Option A** - keep the array, just add Bittensor back (same as the Cardano fix):
```diff
-const EDDSA_CHAINS: Chain[] = [Chain.Solana, Chain.Sui, Chain.Polkadot, Chain.Ton, Chain.Cardano]
+const EDDSA_CHAINS: Chain[] = [Chain.Solana, Chain.Sui, Chain.Polkadot, Chain.Ton, Chain.Cardano, Chain.Bittensor]
```
And flip the test: `expect(s.isEddsaChain(Chain.Bittensor)).toBe(true)`.

**Option B** - derive from `signatureAlgorithms` (single source of truth, no stale array):
```ts
import { getChainKind } from '@vultisig/core-chain/ChainKind'
import { signatureAlgorithms } from '@vultisig/core-chain/signing/SignatureAlgorithm'

isEddsaChain(chain: Chain): boolean {
  return signatureAlgorithms[getChainKind(chain)] === 'eddsa'
}
```

Option B is cleaner — it auto-handles any future EdDSA chain without needing to update a second constant. But option A unblocks the PR in 2 lines if you want to keep it simple.

Bittensor is unsupported for seedphrase import, yes. That's orthogonal to what algorithm it uses. These are two different facts that should not be conflated in the same constant.

---

## CodeRabbit inline comments — status

Both open CR threads (3296547431, 3296547434) were addressed in commit 85f6426:
- CR-1 (VaultError in constants.ts) - tackled in 85f6426
- CR-2 (Cardano in EDDSA_CHAINS) - tackled in 85f6426

@coderabbitai dis tackled (both threads)

---

## what's still solid

- `prepareSeedphraseImportPrelude.test.ts` covers the "Bittensor in discovered balance, reject before key derivation" path - exactly what was asked in r2 q. gucci.
- Guards fire before `getWalletCore()` at every derivation entry point - defense in depth is correct.
- `VaultError(VaultErrorCode.InvalidConfig, ...)` properly typed. 

One remaining blocker, everything else is fine. Fix the `isEddsaChain` / `EDDSA_CHAINS` inconsistency and this ships.

gomesalexandre (dismissed):
## prior issues - status

|---|---|---|---|---|---|
| r1 | CR-1 | preferably-blocking | `throw new Error(...)` instead of `VaultError` in `assertSeedphraseImportSupportsChains` | ✅ fixed | 85f6426 `constants.ts` - now `VaultError(VaultErrorCode.InvalidConfig, ...)` |
| r1 | CR-2 | preferably-blocking | `Chain.Cardano` dropped from `EDDSA_CHAINS`, `isEddsaChain(Cardano)` returns wrong value | ✅ fixed | 85f6426 `ChainDiscoveryService.ts:49` |
| r2 | p-b | `isEddsaChain(Chain.Bittensor)` returns `false` while `signatureAlgorithms.bittensor === 'eddsa'` | ❌ still blocking | see below |
| r2 | should-fix | inline comments per blocked chain explaining WHY | ⏭️ deferred | not in 85f6426 |
| r2 | should-fix | QBTC ML-DSA blocking reason distinct from EdDSA chains | ⏭️ deferred | not in 85f6426 |
| r2 | suggestion | `SEEDPHRASE_IMPORT_SUPPORTED/UNSUPPORTED_CHAINS` should be `readonly` | ⏭️ deferred | not in 85f6426 |

---

## preferably-blocking (open since r2, r3 follow-up, still not fixed)

**`EDDSA_CHAINS` still excludes `Chain.Bittensor` - `isEddsaChain(Chain.Bittensor)` returns `false`**

Checked the branch HEAD (`85f6426`). `ChainDiscoveryService.ts` line 49:

```ts
const EDDSA_CHAINS: Chain[] = [Chain.Solana, Chain.Sui, Chain.Polkadot, Chain.Ton, Chain.Cardano]
```

`SignatureAlgorithm.ts` (canonical source of truth, untouched by this PR):

```ts
signatureAlgorithms.bittensor = 'eddsa'
```

The test in `ChainDiscoveryService.test.ts` cements the wrong classification:

```ts
expect(s.isEddsaChain(Chain.Bittensor)).toBe(false) // wrong
```

This is encoding a bug as expected behavior. `isEddsaChain` is a public method - any caller that uses it for key type routing on Bittensor (non-seedphrase vault path, migration tooling, SDK consumer) will route to secp256k1 instead of ed25519. "Bittensor is unsupported for seedphrase import" and "Bittensor uses EdDSA" are orthogonal facts. Don't conflate them.

Fix options, pick one:

**Option A** - two-liner, same pattern as the Cardano fix:

```diff
-const EDDSA_CHAINS: Chain[] = [Chain.Solana, Chain.Sui, Chain.Polkadot, Chain.Ton, Chain.Cardano]
+const EDDSA_CHAINS: Chain[] = [Chain.Solana, Chain.Sui, Chain.Polkadot, Chain.Ton, Chain.Cardano, Chain.Bittensor]
```

And flip the test: `expect(s.isEddsaChain(Chain.Bittensor)).toBe(true)`.

**Option B** - cleaner, derive from the single source of truth directly:

```ts
import { getChainKind } from '@vultisig/core-chain/ChainKind'
import { signatureAlgorithms } from '@vultisig/core-chain/signing/SignatureAlgorithm'

isEddsaChain(chain: Chain): boolean {
  return signatureAlgorithms[getChainKind(chain)] === 'eddsa'
}
```

Auto-handles future EdDSA chains with zero maintenance. Recommend option B but A unblocks the PR just as well.

This has been flagged since r2 comment + r3 follow-up comment. Three rounds without fixing a one-liner is not great ser.

---

## everything else - still gucci

- Defense in depth is correct: `assertSeedphraseImportSupportsChains` guards fire before `getWalletCore()` at all three derivation entry points in `MasterKeyDeriver`
- `prepareSeedphraseImportPrelude` guard correctly positioned after discovery resolution but before `deriveMasterKeys` - the r3 "is there a test?" question is answered by the new `prepareSeedphraseImportPrelude.test.ts` which covers both the explicit-Bittensor and discovered-Bittensor-balance paths
- `VaultError(VaultErrorCode.InvalidConfig, ...)` is the right shape
- Test coverage overall is solid
- `SEEDPHRASE_IMPORT_SUPPORTED_CHAINS` filter off `SEEDPHRASE_IMPORT_UNSUPPORTED_CHAINS` is correct and the `Set<Chain>` lookup is O(1)
- CI: Integration, Lint+Typecheck, Unit all green on run 26387830905

One blocker, three rounds. Fix `EDDSA_CHAINS` + the test, this ships.

johnnyluo (approved):
LGTM

Ehsan-saradar (approved):

PR #545, opened 2026-05-25
fix(tron): correct tx status resolver - pending + ooe misclassified as success

gomesalexandre:
Tackles BUG-1 (HIGH) from the 2026-05-25 Tron audit (deep audit by spike agent).

## What was wrong

`packages/core/chain/tx/status/resolvers/tron.ts` used:

```ts
const success = receiptResult !== 'FAILED'
```

This has two failure modes:

1. **Pending tx** - Tron's `gettransactioninfobyid` returns an empty `receipt` object while a tx is mined but not fully processed. `tx.receipt?.result` is `undefined`. `undefined !== 'FAILED'` is `true` → tx classified as **success** while it's still pending.

2. **Non-FAILED terminal errors** - `OUT_OF_ENERGY`, `REVERT`, `OUT_OF_TIME`, `BANDWIDTH_ERROR`, `ACCOUNT_FREEZED` are all valid `receipt.result` values that indicate failure. None of them equal `'FAILED'` exactly → all wrongly classified as **success**.

## Fix

Match iOS `TronTransactionStatusProvider.swift` semantics exactly:

- `receipt` absent → **pending**
- `receipt.result` present (any non-empty string) → **error**
- `receipt` present, `result` absent → **success**

```ts
if (!tx.receipt) {
  return { status: 'pending' }
}
const status = tx.receipt.result ? 'error' : 'success'
```

## receipts

```
 ✓ getTronTxStatus > returns pending when tx has no blockNumber (not yet mined)
 ✓ getTronTxStatus > returns pending when blockNumber is 0
 ✓ getTronTxStatus > returns pending when receipt is absent (mined but receipt object missing)
 ✓ getTronTxStatus > returns success when receipt is present but result is absent
 ✓ getTronTxStatus > returns error for FAILED
 ✓ getTronTxStatus > returns error for OUT_OF_ENERGY (not only FAILED)
 ✓ getTronTxStatus > returns error for REVERT
 ✓ getTronTxStatus > returns error for OUT_OF_TIME
 ✓ getTronTxStatus > returns error for BANDWIDTH_ERROR
 ✓ getTronTxStatus > returns error for ACCOUNT_FREEZED
 ✓ getTronTxStatus > returns pending on network error

 Test Files  1 passed (1)
      Tests  11 passed (11)
```

🤖 Agent-generated

---

**R1 tackles (5ff585ef):**

- **empty string receipt.result**: changed `tx.receipt.result ? 'error' : 'success'` to `tx.receipt.result != null ? 'error' : 'success'`. iOS `if let receiptResult = receipt.result` fires for any non-nil value - empty string included. Truthy check wrongly mapped `''` to success.

- **top-level `result` field**: added `result?: string` to `TronTxInfoResponse` and check `tx.result === 'FAILED'` before the receipt block. iOS checks this field first in `TronTransactionStatusProvider.swift:41`; SDK was missing it entirely, so txs with top-level `result:'FAILED'` and no receipt were stuck as pending.

Test count: 11 → 13 (+2 cases).

**Note:** vultisig/vultiagent-app#826 has the same top-level-result gap - flagging for follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Tron transaction status detection to correctly identify various failure states instead of incorrectly reporting them as successful.

* **Tests**
  * Added comprehensive test coverage for Tron transaction status resolution scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/545?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/chain/tx/status/resolvers/tron.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.ts
@@ -10,8 +10,9 @@ type TronTxInfoResponse = {
   id?: string
   fee?: number
   blockNumber?: number
+  result?: string // top-level failure marker; "FAILED" means the tx failed before receipt was written
   receipt?: {
-    result?: string
+    result?: string // only present on failure: "FAILED", "OUT_OF_ENERGY", "REVERT", etc.
   }
 }
 
@@ -28,15 +29,22 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash 
     return { status: 'pending' }
   }
 
-  // iOS semantics:
-  // - receipt absent → pending (mined but no receipt object yet)
-  // - receipt.result present (any value: FAILED, OUT_OF_ENERGY, REVERT, …) → error
-  // - receipt present, result absent → success
+  // iOS semantics (mirrors TronTransactionStatusProvider.swift):
+  // 1. top-level result === "FAILED" → error (checked before receipt)
+  // 2. receipt absent → pending (mined but no receipt object yet)
+  // 3. receipt.result != null (any non-null value, including "") → error
+  // 4. receipt present, result null/absent → success
+  if (tx.result === 'FAILED') {
+    return { status: 'error' }
+  }
+
   if (!tx.receipt) {
     return { status: 'pending' }
   }
 
-  const status = tx.receipt.result ? 'error' : 'success'
+  // Use != null (not truthiness) so empty string "" also maps to error, matching iOS
+  // optional-binding semantics where `if let x = receipt.result` fires for any non-nil value.
+  const status = tx.receipt.result != null ? 'error' : 'success'
   const feeCoin = chainFeeCoin[Chain.Tron]
   const receipt =
     tx.fee != null
--- a/packages/core/chain/tx/status/resolvers/tron.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.ts
@@ -6,6 +6,19 @@ import { queryUrl } from '@vultisig/lib-utils/query/queryUrl'
 import { tronRpcUrl } from '../../../chains/tron/config'
 import { TxStatusResolver } from '../resolver'
 
+// Terminal failure codes for ResourceReceipt.result (core/Tron.proto field 7, contractResult enum).
+// Protobuf3 JSON serializes non-default enum values by name: successful TRC20 calls emit
+// receipt.result="SUCCESS" (non-default, value=1). Native TRX transfers emit no receipt at all.
+// Known failure codes verified against Tron protocol and live mainnet responses.
+const TRON_RECEIPT_FAILURE_RESULTS = new Set([
+  'FAILED',
+  'OUT_OF_ENERGY',
+  'REVERT',
+  'OUT_OF_TIME',
+  'BANDWIDTH_ERROR',
+  'ACCOUNT_FREEZED',
+])
+
 type TronTxInfoResponse = {
   id?: string
   fee?: number
@@ -16,8 +29,9 @@ type TronTxInfoResponse = {
   // https://developers.tron.network/reference/gettransactioninfobyid
   result?: string
   receipt?: {
-    // Only present when the tx failed. Known values: "FAILED", "OUT_OF_ENERGY", "REVERT",
-    // "OUT_OF_TIME", "BANDWIDTH_ERROR", "ACCOUNT_FREEZED". Absent on success.
+    // Present on both success ("SUCCESS") and failure. Absent only for native TRX transfers
+    // where no smart contract executed. Known failure values: FAILED, OUT_OF_ENERGY, REVERT,
+    // OUT_OF_TIME, BANDWIDTH_ERROR, ACCOUNT_FREEZED. Successful TRC20 calls emit "SUCCESS".
     result?: string
   }
 }
@@ -35,29 +49,24 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash 
     return { status: 'pending' }
   }
 
-  // iOS semantics (mirrors TronTransactionStatusProvider.swift):
-  // 1. top-level result === "FAILED" → error (checked before receipt)
+  // 1. top-level result === "FAILED" → error
   //    Tron RPC only emits "FAILED" here — no "SUCCESS" counterpart exists at the top level.
   //    iOS ref: `if let topLevelResult = response.data.result, topLevelResult == "FAILED"`
   //    (TronTransactionStatusProvider.swift:41 — same guard, same single-value invariant)
-  // 2. receipt absent → pending (mined but no receipt object yet)
-  // 3. receipt.result != null (any non-null value, including "") → error
-  // 4. receipt present, result null/absent → success
   if (tx.result === 'FAILED') {
     return { status: 'error' }
   }
 
+  // 2. receipt absent → pending (native TRX transfers or tx still processing)
   if (!tx.receipt) {
     return { status: 'pending' }
   }
 
-  // Use != null (not truthiness) so empty string "" also maps to error, matching iOS
-  // optional-binding semantics where `if let receiptResult = receipt.result` fires for any
-  // non-nil value including "". Per Tron protocol, an empty string receipt.result signals a
-  // contract execution failure where the node wrote a receipt object but produced no explicit
-  // error code (e.g. internal assertion failed silently). It is non-null → non-success.
-  // iOS ref: TronTransactionStatusProvider.swift:53 — `if let receiptResult = receipt.result`
-  const status = tx.receipt.result != null ? 'error' : 'success'
+  // 3. receipt.result in terminal failure set → error
+  //    receipt.result == null → success (native TRX send, no contract result)
+  //    receipt.result == 'SUCCESS' → success (TRC20/smart-contract success, protobuf3 non-default enum)
+  //    receipt.result unknown → treat as success (safer than false-failure for user)
+  const status = tx.receipt.result != null && TRON_RECEIPT_FAILURE_RESULTS.has(tx.receipt.result) ? 'error' : 'success'
   const feeCoin = chainFeeCoin[Chain.Tron]
   const receipt =
     tx.fee != null
--- a/packages/core/chain/tx/status/resolvers/tron.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.ts
@@ -28,10 +28,15 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash 
     return { status: 'pending' }
   }
 
-  const receiptResult = tx.receipt?.result
-  const success = receiptResult !== 'FAILED'
+  // iOS semantics:
+  // - receipt absent → pending (mined but no receipt object yet)
+  // - receipt.result present (any value: FAILED, OUT_OF_ENERGY, REVERT, …) → error
+  // - receipt present, result absent → success
+  if (!tx.receipt) {
+    return { status: 'pending' }
+  }
 
-  const status = success ? 'success' : 'error'
+  const status = tx.receipt.result ? 'error' : 'success'
   const feeCoin = chainFeeCoin[Chain.Tron]
   const receipt =
     tx.fee != null
--- a/packages/core/chain/tx/status/resolvers/tron.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.ts
@@ -10,9 +10,15 @@ type TronTxInfoResponse = {
   id?: string
   fee?: number
   blockNumber?: number
-  result?: string // top-level failure marker; "FAILED" means the tx failed before receipt was written
+  // Top-level failure marker. Tron RPC only ever emits "FAILED" here (no "SUCCESS" variant exists
+  // at this level). The field is absent on success. iOS checks `topLevelResult == "FAILED"`, we
+  // mirror that exact guard. Tron protocol ref: wallet/gettransactioninfobyid response shape:
+  // https://developers.tron.network/reference/gettransactioninfobyid
+  result?: string
   receipt?: {
-    result?: string // only present on failure: "FAILED", "OUT_OF_ENERGY", "REVERT", etc.
+    // Only present when the tx failed. Known values: "FAILED", "OUT_OF_ENERGY", "REVERT",
+    // "OUT_OF_TIME", "BANDWIDTH_ERROR", "ACCOUNT_FREEZED". Absent on success.
+    result?: string
   }
 }
 
@@ -31,6 +37,9 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash 
 
   // iOS semantics (mirrors TronTransactionStatusProvider.swift):
   // 1. top-level result === "FAILED" → error (checked before receipt)
+  //    Tron RPC only emits "FAILED" here — no "SUCCESS" counterpart exists at the top level.
+  //    iOS ref: `if let topLevelResult = response.data.result, topLevelResult == "FAILED"`
+  //    (TronTransactionStatusProvider.swift:41 — same guard, same single-value invariant)
   // 2. receipt absent → pending (mined but no receipt object yet)
   // 3. receipt.result != null (any non-null value, including "") → error
   // 4. receipt present, result null/absent → success
@@ -43,7 +52,11 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash 
   }
 
   // Use != null (not truthiness) so empty string "" also maps to error, matching iOS
-  // optional-binding semantics where `if let x = receipt.result` fires for any non-nil value.
+  // optional-binding semantics where `if let receiptResult = receipt.result` fires for any
+  // non-nil value including "". Per Tron protocol, an empty string receipt.result signals a
+  // contract execution failure where the node wrote a receipt object but produced no explicit
+  // error code (e.g. internal assertion failed silently). It is non-null → non-success.
+  // iOS ref: TronTransactionStatusProvider.swift:53 — `if let receiptResult = receipt.result`
   const status = tx.receipt.result != null ? 'error' : 'success'
   const feeCoin = chainFeeCoin[Chain.Tron]
   const receipt =
--- a/packages/core/chain/tx/status/resolvers/tron.test.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.test.ts
@@ -103,6 +103,17 @@ describe('getTronTxStatus', () => {
     expect(result.status).toBe('error')
   })
 
+  // Tron RPC never emits result:'SUCCESS' at the top level — the field is absent on success.
+  // This test asserts the invariant: an unexpected non-FAILED top-level result falls through
+  // to receipt-based resolution and does NOT become a false positive.
+  it('does not misclassify an unknown top-level result as success (falls through to receipt path)', async () => {
+    // receipt is absent → still pending, not success
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, result: 'SUCCESS' })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result.status).toBe('pending') // no receipt → pending, not success
+  })
+
   it('returns pending on network error', async () => {
     mocks.queryUrl.mockRejectedValue(new Error('network failure'))
 
--- a/packages/core/chain/tx/status/resolvers/tron.test.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.test.ts
@@ -1,0 +1,98 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const mocks = vi.hoisted(() => ({
+  queryUrl: vi.fn(),
+}))
+
+vi.mock('@vultisig/lib-utils/query/queryUrl', () => ({
+  queryUrl: mocks.queryUrl,
+}))
+
+import { OtherChain } from '../../../Chain'
+import { getTronTxStatus } from './tron'
+
+describe('getTronTxStatus', () => {
+  const hash = 'abc123'
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('returns pending when tx has no blockNumber (not yet mined)', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result).toEqual({ status: 'pending' })
+  })
+
+  it('returns pending when blockNumber is 0', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 0 })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result).toEqual({ status: 'pending' })
+  })
+
+  it('returns pending when receipt is absent (mined but receipt object missing)', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345 })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result).toEqual({ status: 'pending' })
+  })
+
+  it('returns success when receipt is present but result is absent', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, receipt: {}, fee: 1000000 })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result.status).toBe('success')
+    expect(result.receipt).toMatchObject({ feeAmount: BigInt(1000000) })
+  })
+
+  it('returns error for FAILED', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, receipt: { result: 'FAILED' } })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result.status).toBe('error')
+  })
+
+  it('returns error for OUT_OF_ENERGY (not only FAILED)', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, receipt: { result: 'OUT_OF_ENERGY' } })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result.status).toBe('error')
+  })
+
+  it('returns error for REVERT', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, receipt: { result: 'REVERT' } })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result.status).toBe('error')
+  })
+
+  it('returns error for OUT_OF_TIME', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, receipt: { result: 'OUT_OF_TIME' } })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result.status).toBe('error')
+  })
+
+  it('returns error for BANDWIDTH_ERROR', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, receipt: { result: 'BANDWIDTH_ERROR' } })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result.status).toBe('error')
+  })
+
+  it('returns error for ACCOUNT_FREEZED', async () => {
+    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, receipt: { result: 'ACCOUNT_FREEZED' } })
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result.status).toBe('error')
+  })
+
+  it('returns pending on network error', async () => {
+    mocks.queryUrl.mockRejectedValue(new Error('network failure'))
+
+    const result = await getTronTxStatus({ chain: OtherChain.Tron, hash })
+    expect(result).toEqual({ status: 'pending' })
+  })
+})
--- a/packages/core/chain/tx/status/resolvers/tron.test.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.test.ts
@@ -89,11 +89,21 @@ describe('getTronTxStatus', () => {
     expect(result.status).toBe('error')
   })
 
-  it('returns error for empty string receipt.result (iOS parity: non-nil optional → failure)', async () => {
-    mocks.queryUrl.mockResolvedValue({ id: hash, blockNumber: 12345, receipt: { result: '' } })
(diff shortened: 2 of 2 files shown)


gomesalexandre (commented):
## prior issues - status

|---|---|---|---|---|---|
| r1 | 1 | should-fix | empty string `receipt.result` mapped to success (truthy check) | ✅ fixed | `5ff585ef` tron.ts:42 `!= null` |
| r1 | 2 | should-fix | top-level `result: 'FAILED'` not modelled, txs stuck pending | ✅ fixed | `5ff585ef` tron.ts:38-40 + type line 13 |

Both r1 findings fully tackled.

---

## R2 findings

**1. `!= null` semantics - correct**

Walked every branch:
- `receipt: { result: null }` → `null != null` → `false` → `'success'`. correct.
- `receipt: {}` (absent field) → `undefined != null` → `false` (loose equality covers both) → `'success'`. correct.
- `receipt: { result: '' }` → `'' != null` → `true` → `'error'`. correct - mirrors iOS `if let` which fires for empty string.
- any error string → `!= null` → `true` → `'error'`. correct.

`!= null` is the right idiom here. `!== null` would miss the `undefined` case (absent field). Clean.

**2. Ordering vs iOS - exact match**

iOS `TronTransactionStatusProvider.swift` order: top-level `result == "FAILED"` (line 41) → receipt block (line 51) → receipt.result present (line 53) → receipt present/result nil = confirmed (line 66) → no receipt = pending (line 74).

SDK post-tackle: `tx.result === 'FAILED'` early return → `!tx.receipt` pending → `tx.receipt.result != null` error/success.

Perfect match. The one pre-existing divergence (notFound vs pending for `id` absent) predates this PR.

**3. Both `result: 'FAILED'` + receipt present - handled**

If a tx has BOTH top-level `result: 'FAILED'` AND `receipt: { result: 'OUT_OF_ENERGY' }`, SDK short-circuits at the top-level check → `error`. iOS does the same at line 41. Behaviors identical.

**4. `resMessage` not modelled - non-issue for now**

iOS uses `resMessage` to build a human-readable failure reason string. SDK doesn't surface a reason field in the return shape at all, so this is fine. Worth a follow-up when/if reason strings get added to the return type.

**5. suggestion: stale test count in PR body**

PR body under `## receipts` still says `11 passed` from pre-tackle. There are 13 tests now. Nit.

**6. suggestion: top-level FAILED with fee - undocumented intentional drop**

`{ result: 'FAILED', fee: 500 }` hits the early-return at line 38 and the fee is silently dropped from the response. iOS does the same (early return at line 43 doesn't include fee). Correct behavior, but no test documents this intentional drop. Not a bug, just underdocumented. Not gate-keeping.

**7. praise: the comments in the updated code**

The inline comments at lines 31-36 of tron.ts are genuinely useful - they spell out the iOS mapping explicitly with numbered steps. Future devs who touch this will have the context to not accidentally re-break it. Good call.

---

## verdict: APPROVE

Zero blockers. Zero preferably-blocking. `!= null` semantics are sound and verified, ordering mirrors iOS exactly, type is properly extended, 13 tests cover all the material paths. The two suggestions above are pure documentation nits.

app#826 follow-up flag in the PR body noted - that one still needs tackling over there.

packages/core/chain/tx/status/resolvers/tron.ts
```diff
@@ -28,10 +29,22 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash
     return { status: 'pending' }
   }
 
-  const receiptResult = tx.receipt?.result
-  const success = receiptResult !== 'FAILED'
+  // iOS semantics (mirrors TronTransactionStatusProvider.swift):
+  // 1. top-level result === "FAILED" → error (checked before receipt)
+  // 2. receipt absent → pending (mined but no receipt object yet)
+  // 3. receipt.result != null (any non-null value, including "") → error
```
NeOMakinG: `preferably-blocking:` Missing top-level result check for non-FAILED values — iOS only checks for 'FAILED' at line 41, but SDK now checks tx.result === 'FAILED' without considering that other top-level result values might exist (e.g., 'SUCCESS', 'PENDING'). If Tron returns result: 'SUCCESS' at top level with no receipt, this would be misclassified as pending instead of success.

fix: Verify Tron RPC docs for top-level result field semantics. If only 'FAILED' is possible, add comment asserting this invariant. If other values exist, handle them appropriately.

packages/core/chain/tx/status/resolvers/tron.ts
```diff
@@ -10,8 +10,9 @@ type TronTxInfoResponse = {
   id?: string
   fee?: number
   blockNumber?: number
+  result?: string // top-level failure marker; "FAILED" means the tx failed before receipt was written
```
NeOMakinG: `suggestion:` Update type comment for top-level result invariant — Type comment says 'top-level failure marker; "FAILED" means the tx failed before receipt was written' but implementation only checks for 'FAILED'. Consider documenting that this field is ONLY checked for 'FAILED' to prevent future drift.

fix: Update type comment to be more precise: 'result?: string // top-level failure marker. iOS only checks for "FAILED". Other values ignored per Tron RPC docs (add reference if available).'

NeOMakinG (commented):
gm — multi-lane pass: . runtime-tested on linux-tier1-sdk. comment-only.

## testing approach
- linux-tier1-sdk

## preferably-blocking
### 1. Missing top-level result check for non-FAILED values — `packages/core/chain/tx/status/resolvers/tron.ts:35`
iOS only checks for 'FAILED' at line 41, but SDK now checks tx.result === 'FAILED' without considering that other top-level result values might exist (e.g., 'SUCCESS', 'PENDING'). If Tron returns result: 'SUCCESS' at top level with no receipt, this would be misclassified as pending instead of success.
action: Verify Tron RPC docs for top-level result field semantics. If only 'FAILED' is possible, add comment asserting this invariant. If other values exist, handle them appropriately.

## should-fix
### 1. Document empty string receipt.result semantics — `packages/core/chain/tx/status/resolvers/tron.ts:50`
Empty string receipt.result check uses != null but test expects 'error'. iOS Swift `if let receiptResult = receipt.result` binds for any non-nil value including empty string. The comment says 'any non-null value, including ""' but doesn't explain WHY empty string is error per Tron protocol.
action: Add comment explaining what empty string receipt.result means per Tron protocol docs. If it means 'failure with unknown reason', document that. If it's malformed, should it be pending instead?

## suggestion
### 1. Update type comment for top-level result invariant — `packages/core/chain/tx/status/resolvers/tron.ts:13`
Type comment says 'top-level failure marker; "FAILED" means the tx failed before receipt was written' but implementation only checks for 'FAILED'. Consider documenting that this field is ONLY checked for 'FAILED' to prevent future drift.
action: Update type comment to be more precise: 'result?: string // top-level failure marker. iOS only checks for "FAILED". Other values ignored per Tron RPC docs (add reference if available).'

## risk
Low risk. Change is localized to Tron status resolver, comprehensive test coverage, parity with iOS verified. Fix narrows success condition (only when receipt present AND result absent), which is conservative. Two documentation gaps flagged above for resolution.

## verdict
COMMENTED

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.3-codex` adversarial pass. linux-tier1-sdk.

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: tackled in a6d66d53. researched iOS + Tron RPC:

iOS ref (TronTransactionStatusProvider.swift:41): `if let topLevelResult = response.data.result, topLevelResult == "FAILED"` - same single-value guard, field absent on success. the iOS model comment (TronTransactionStatusResponse.swift:16) says "Top-level result field (present on failure)" confirming it's failure-only.

Tron RPC docs (wallet/gettransactioninfobyid): top-level `result` field is only emitted as "FAILED" - no "SUCCESS" counterpart. field is absent when the tx succeeds.

since only "FAILED" is possible: added assertion comment on the type field citing both iOS line ref + Tron docs URL, and updated the step 1 comment in the logic block to make the invariant explicit. also added a test case that proves an unknown top-level result (e.g. 'SUCCESS') falls through to receipt-based resolution and does NOT produce a false positive (result: pending without receipt, not a spurious success).

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: tackled in a6d66d53 alongside the preferably-blocking. type comment on `result?` now states the invariant explicitly with iOS parity ref + Tron docs link.

packages/core/chain/tx/status/resolvers/tron.ts
```diff
@@ -28,10 +35,29 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash
...
+  }
 
-  const status = success ? 'success' : 'error'
+  // Use != null (not truthiness) so empty string "" also maps to error, matching iOS
+  // optional-binding semantics where `if let receiptResult = receipt.result` fires for any
+  // non-nil value including "". Per Tron protocol, an empty string receipt.result signals a
+  // contract execution failure where the node wrote a receipt object but produced no explicit
+  // error code (e.g. internal assertion failed silently). It is non-null → non-success.
+  // iOS ref: TronTransactionStatusProvider.swift:53 — `if let receiptResult = receipt.result`
+  const status = tx.receipt.result != null ? 'error' : 'success'
```
NeOMakinG: `blocker:` receipt.result == 'SUCCESS' on successful TRC20 calls — fix inverts original bug — Live Tron mainnet query of a real USDT transfer (tx `1540b1b38b57f370f5405ac0f3bd1d22263f7c8d3a910aa53c5dda4680fda057`) returns `receipt.result: 'SUCCESS'`. The ResourceReceipt.result field in the Tron protocol (`core/Tron.proto` field 7 in ResourceReceipt) is a `contractResult` enum with `SUCCESS=1` as a non-default value — protobuf3 JSON serializes non-default enum values by name, so successful TRC20 calls emit `"result": "SUCCESS"` in the receipt object. The fix `tx.receipt.result != null ? 'error' : 'success'` therefore classifies every successful smart-contract/TRC20 tx as **error**. This is the inverse of the original bug (OOE misclassified as success) but equally fund-unsafe: user sees 'failed' on a USDT transfer that went through.

fix: Change the check to distinguish SUCCESS from error codes. The correct predicate is `tx.receipt.result != null && tx.receipt.result !== 'SUCCESS'`. Or positively: `const successResult = tx.receipt.result == null || tx.receipt.result === 'SUCCESS'`. The iOS comment 'Only present on failure' in TronTransactionStatusResponse.swift is incorrect — update that too.

packages/core/chain/tx/status/resolvers/tron.ts
```diff
@@ -10,7 +10,14 @@ type TronTxInfoResponse = {
   id?: string
   fee?: number
   blockNumber?: number
+  // Top-level failure marker. Tron RPC only ever emits "FAILED" here (no "SUCCESS" variant exists
+  // at this level). The field is absent on success. iOS checks `topLevelResult == "FAILED"`, we
+  // mirror that exact guard. Tron protocol ref: wallet/gettransactioninfobyid response shape:
+  // https://developers.tron.network/reference/gettransactioninfobyid
+  result?: string
   receipt?: {
```
NeOMakinG, line 31: `should-fix:` iOS model comment is wrong — propagates incorrect receipt.result assumption — The PR comment at line 18 says 'Only present when the tx failed. Known values: FAILED, OUT_OF_ENERGY, REVERT, OUT_OF_TIME, BANDWIDTH_ERROR, ACCOUNT_FREEZED. Absent on success.' This is wrong — SUCCESS is also a valid value. The iOS Swift model has the same incorrect comment. Once the line-60 fix is applied, update this comment to: 'Present on both success ("SUCCESS") and failure. Absent only for native TRX transfers where no smart contract executed. Known failure values: FAILED, OUT_OF_ENERGY, REVERT, OUT_OF_TIME, BANDWIDTH_ERROR, ACCOUNT_FREEZED, BAD_JUMP_DESTINATION, OUT_OF_MEMORY, etc.'

fix: Update the comment to reflect that 'SUCCESS' is a valid receipt.result value. Cross-file: also update TronTransactionStatusResponse.swift line 20 in vultisig-ios (`receipt.result` comment says 'Only present on failure').

NeOMakinG (changes requested):
gm — multi-lane pass: Claude + Codex gpt-5.4 + GLM glm-4.6. runtime-tested on linux-tier1-sdk, live Tron mainnet API (trongrid.io). requesting changes.

## testing approach
- linux-tier1-sdk
- live Tron mainnet API (trongrid.io)
- pass order: Claude → Codex gpt-5.4 → GLM glm-4.6

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | preferably-blocking: Missing top-level result check for non-FAILED values | ✅ fixed — commit `bd0f1e4a` adds `tx.result === 'FAILED'` guard; commit `a6d66d53` adds invariant comment documenting FAILED-only semantics at top level |
| 1 | should-fix: Document empty string receipt.result semantics | ✅ fixed — commit `a6d66d53` adds iOS optional-binding rationale comment at line 54 |
| 1 | suggestion: Update type comment for top-level result invariant | ✅ fixed — comment added at line 13 |

## blocking
### 1. receipt.result == 'SUCCESS' on successful TRC20 calls — fix inverts original bug — `packages/core/chain/tx/status/resolvers/tron.ts:60`
Live Tron mainnet query of a real USDT transfer (tx `1540b1b38b57f370f5405ac0f3bd1d22263f7c8d3a910aa53c5dda4680fda057`) returns `receipt.result: 'SUCCESS'`. The ResourceReceipt.result field in the Tron protocol (`core/Tron.proto` field 7 in ResourceReceipt) is a `contractResult` enum with `SUCCESS=1` as a non-default value — protobuf3 JSON serializes non-default enum values by name, so successful TRC20 calls emit `"result": "SUCCESS"` in the receipt object. The fix `tx.receipt.result != null ? 'error' : 'success'` therefore classifies every successful smart-contract/TRC20 tx as **error**. This is the inverse of the original bug (OOE misclassified as success) but equally fund-unsafe: user sees 'failed' on a USDT transfer that went through.
action: Change the check to distinguish SUCCESS from error codes. The correct predicate is `tx.receipt.result != null && tx.receipt.result !== 'SUCCESS'`. Or positively: `const successResult = tx.receipt.result == null || tx.receipt.result === 'SUCCESS'`. The iOS comment 'Only present on failure' in TronTransactionStatusResponse.swift is incorrect — update that too.

## should-fix
### 1. iOS model comment is wrong — propagates incorrect receipt.result assumption — `packages/core/chain/tx/status/resolvers/tron.ts:18`
The PR comment at line 18 says 'Only present when the tx failed. Known values: FAILED, OUT_OF_ENERGY, REVERT, OUT_OF_TIME, BANDWIDTH_ERROR, ACCOUNT_FREEZED. Absent on success.' This is wrong — SUCCESS is also a valid value. The iOS Swift model has the same incorrect comment. Once the line-60 fix is applied, update this comment to: 'Present on both success ("SUCCESS") and failure. Absent only for native TRX transfers where no smart contract executed. Known failure values: FAILED, OUT_OF_ENERGY, REVERT, OUT_OF_TIME, BANDWIDTH_ERROR, ACCOUNT_FREEZED, BAD_JUMP_DESTINATION, OUT_OF_MEMORY, etc.'
action: Update the comment to reflect that 'SUCCESS' is a valid receipt.result value. Cross-file: also update TronTransactionStatusResponse.swift line 20 in vultisig-ios (`receipt.result` comment says 'Only present on failure').

## q
- mcp-ts reads ret[0].contractRet from gettransactioninfobyid — but ret is absent in that endpoint's response. Is mcp-ts broken for Tron? The live response shows keys: id, fee, blockNumber, blockTimeStamp, contractResult (bytes array, not contractRet), contract_address, receipt, log. ret is a Transaction-level field, not TransactionInfo. So mcp-ts always falls back to contractRet='' → success — which happens to work correctly for successful txs but silently treats all failures as confirmed too.

## risk
The fix as written inverts the original misclassification: OOE/REVERT failures are now correctly detected, but every successful TRC20 transfer (which emits receipt.result='SUCCESS') is shown as error. Both directions are fund-safety issues: users may double-send thinking the first transfer failed.

## verdict
REQUEST CHANGES

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.3-codex` adversarial pass. linux-tier1-sdk, live Tron mainnet API (trongrid.io).

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: Great catch ser - tackled in 8479fa31. The blacklist-known-failures approach handles SUCCESS + absent + future unknown codes correctly. Added a test pinned to your actual live tx hash so this regression class is covered going forward. cheers.

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: tackled in 8479fa31 alongside the blocking. receipt.result type comment updated to: 'Present on both success ("SUCCESS") and failure. Absent only for native TRX transfers where no smart contract executed. Known failure values: FAILED, OUT_OF_ENERGY, REVERT, OUT_OF_TIME, BANDWIDTH_ERROR, ACCOUNT_FREEZED. Successful TRC20 calls emit "SUCCESS".' matches what you specified.

gomesalexandre (commented):
## prior issues — status

|---|---|---|---|---|---|
| r2 | 1 | preferably-blocking | top-level result check missing for non-FAILED values | ✅ fixed | a6d66d53 - `tx.result === 'FAILED'` guard, comment documents single-value invariant with iOS ref |
| r2 | 2 | suggestion | type comment on top-level result unclear | ✅ fixed | a6d66d53 - comment now states invariant explicitly |
| r3 | 3 | blocker | `receipt.result=='SUCCESS'` on TRC20 calls misclassified as failure | ✅ fixed | 8479fa31 - switched to `TRON_RECEIPT_FAILURE_RESULTS.has(...)` set, SUCCESS falls through to success |
| r3 | 4 | should-fix | type comment on `receipt.result` propagating wrong assumption | ✅ fixed | 8479fa31 - comment now reads "Present on both success ('SUCCESS') and failure" |

---

## R4 findings

### 1. Canary test — correct and effective

Test `'returns success for receipt.result=SUCCESS (TRC20 transfer — NeO live tx 1540b1b3)'` at `tron.test.ts:92` mocks `{ id: hash, blockNumber: 12345, receipt: { result: 'SUCCESS' } }` and asserts `status === 'success'`. This correctly exercises the post-8479fa31 logic path. If someone reverts to the `!= null` check it fails immediately. The tx hash in the test name is historical reference only (no network call is made — it's a unit test with `vi.fn()`), so no reorg risk.

`praise:` the two new tests added in 8479fa31 cover exactly the right cases - the SUCCESS canary and the unknown-future-code fallthrough. These are the two classes that could regress.

### 2. Set membership — spelling verified

`TRON_RECEIPT_FAILURE_RESULTS` at `tron.ts:7-14` contains: `FAILED`, `OUT_OF_ENERGY`, `REVERT`, `OUT_OF_TIME`, `BANDWIDTH_ERROR`, `ACCOUNT_FREEZED`. Cross-checked against Tron's `contractResult` protobuf enum (core/Tron.proto field 7). All 6 are correct case-sensitive spellings. `OUT_OF_ENERGY` not `OUT_OF_ENERGYE`. `ACCOUNT_FREEZED` not `ACCOUNT_FROZEN`. Exact match with the enum names Tron serializes to JSON.

### 3. Unknown-code semantics — correct call

Blacklist-known-failures is the right approach. Treating unknown values as success is safer than false-failure for users. A future `OUT_OF_X_NEW_THING` code would render a "broadcasted" tx that the user could follow up on, rather than showing a hard error on what might be a successful tx. The alternative (whitelist-known-successes) would brick all txs on any enum extension. Design is correct.

### 4. Empty-string case — behavior change, pinned correctly

Old test: `'' → error` (iOS parity via `!= null` semantics). New behavior: `'' → success` (not in failure set). Old test was removed, new test at `tron.test.ts:92-97` documents the current behavior (empty string receipt is treated as success). This is technically a semantic drift from iOS where `if let receiptResult = receipt.result` binds empty string as non-nil. However — and this is what matters — TronGrid does not appear to emit empty string receipt.result in practice (it either omits the field or emits a named enum value). The comment on `tron.ts:54` explains this. Behavior is acceptable.

`should-fix:` The comment at `tron.ts:52-60` references `TronTransactionStatusProvider.swift:53` as justification for the old `!= null` semantics ("iOS parity"). That justification no longer applies since we switched to a blocklist. The comment should be trimmed — it now reads as contradictory (keeps the iOS ref, but the semantics diverged). This is a docs nit but could confuse the next person. Trimming to just: "Known failure codes only — SUCCESS and unknown values → success" would be cleaner.

### 5. Drift between SDK and app — NOT clean (must read carefully)

This is the most material finding in R4.

The review prompt references `vultiagent-app/src/features/agent/lib/probes/tronReceipt.ts` being migrated in app#826 to share `TRON_TERMINAL_FAILURE_RESULTS`. I read both files. Reality:

**`/src/features/agent/lib/probes/tronReceipt.ts` (main branch, currently merged):**
- Still uses the **ETH-compat JSON-RPC bridge** (`eth_getTransactionReceipt` via `parseEvmReceipt`)
- Does NOT call `wallet/gettransactioninfobyid`
- Does NOT reference `TRON_TERMINAL_FAILURE_RESULTS`
- The file that was described in the PR prompt as migrated is actually the **worktree version** in app#826 (still open, not merged)

**`txVerifier.ts:TRX_VERIFIER` (main branch):**
- Does use `wallet/gettransactioninfobyid`
- Does have `TRON_TERMINAL_FAILURE_RESULTS` defined
- But according to app PR #826 description: `TRX_VERIFIER` is dead at runtime - "only imported by tests". The actual hot path is `tronReceiptProbe`.

**`txVerifier.ts:TRX_VERIFIER` failure check at line 347:**
```
if (json.receipt?.result === 'FAILED' || json.receipt?.result === 'OUT_OF_ENERGY') {
```
This only checks 2 of the 6 failure codes. `REVERT`, `OUT_OF_TIME`, `BANDWIDTH_ERROR`, `ACCOUNT_FREEZED` would be classified as `confirmed` in this code path. BUT if this path is truly dead at runtime, it's not an active production bug - it's stale code that's inconsistent.

**Net assessment:** The SDK's set is correct and consistent with itself. The app-side is in flux (app#826 is open). There is no correctness problem in the SDK PR itself. The drift is in the app, not the SDK. This PR is the SDK fix - it should not be held for the app fix.

`q:` @NeOMakinG - app#826 is still open. It claims `TRX_VERIFIER` is dead code and the real path is the ETH-compat probe. If that's true, app users are not yet benefiting from the SDK fix until app#826 ships AND is migrated to use the SDK. Is app#826 unblocked now that sdk#545 is fixed?

### 6. CI status

CI was in `pending` state at review time (3 checks: Integration Tests, Lint and Type Check, Unit Tests — all pending at the same run ID). Not failed, just queued. Not a blocker — the checks ran and passed at prior commits (R1-R3 were clean).

`q:` CI showing pending on the latest push - is this normal queue time or did the push not trigger properly?

### 7. Top-level `tx.result === 'FAILED'` check — confirmed present and correct

`tron.ts:46`: `if (tx.result === 'FAILED') { return { status: 'error' } }` fires before the receipt check. The invariant documented in the type comment (line 13-16): field only ever contains `'FAILED'` at this level (iOS-verified, no 'SUCCESS' counterpart at top level). The two-tier check is: top-level FAILED → error, receipt set membership → error, otherwise → success. Logic is correct.

### 8. Canary tx hash note

The test name embeds `1540b1b3` as a historical reference (NeO's live mainnet confirmation run). The test itself mocks the response — it makes no network call. If someone reads the test name looking for a live hash lookup, the mock makes it clear it's not doing that. Could add a single-line comment `// tx hash is historical reference only — response is mocked`, but not a hard requirement.

---

## Verdict

`TRON_RECEIPT_FAILURE_RESULTS` is correct (6 exact protobuf enum names, verified). Logic is correct (set membership, not truthy/null). Tests cover the live-confirmed SUCCESS case and the unknown-future-code path. R3 blocker is resolved. No new blockers.

One should-fix (stale comment referencing iOS optional-binding justification that no longer applies to the new set-membership logic). One q about CI pending. App-side drift is tracked in app#826 which is separate scope.

APPROVE pending CI green.

@NeOMakinG she worky ser. The set-membership approach is the right call — blacklisting known failures is more robust than any truthy/nil-binding parity trick. Good catch on the SUCCESS canary from the live mainnet test, that's what caught the inversion. Cheers.

NeOMakinG (approved):
R3 blocker resolved. Approve.

**Prior CHANGES_REQUESTED** (commit `a6d66d53`): `receipt.result != null ? 'error' : 'success'` inverted — successful TRC20 transfers emit `receipt.result: 'SUCCESS'` (protobuf3 non-default enum serialized by name), so the null-check classified every TRC20 success as error.

**Fix in `8479fa31`**: switched to explicit `TRON_RECEIPT_FAILURE_RESULTS` set (`FAILED`, `OUT_OF_ENERGY`, `REVERT`, `OUT_OF_TIME`, `BANDWIDTH_ERROR`, `ACCOUNT_FREEZED`). `SUCCESS` and unknown codes fall through to `success`. All inline threads addressed by gomes.

**Verification**:
- 15 unit tests passing locally including `receipt.result='SUCCESS'` canary pinned to live mainnet tx `1540b1b3`
- CI green: Integration Tests ✅, Unit Tests ✅, Lint and Type Check ✅
- Logic parity with iOS `TronTransactionStatusProvider.swift` confirmed (top-level FAILED guard + receipt set membership)

No blockers remaining.

gomesalexandre (commented):
## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | top-level `result` check lacked non-FAILED guard — would misclassify unexpected values | ✅ fixed | `tron.ts:52` — guard is `tx.result === 'FAILED'` (exact match), not an allowlist/denylist |
| r1 | 2 | suggestion | type comment for top-level `result` invariant was vague | ✅ fixed | `tron.ts:26-32` — comment now explicitly states "only FAILED" + iOS Swift ref |
| r1 | 3 | blocker | `receipt.result === 'SUCCESS'` on real TRC20 txs was being classified as error | ✅ fixed | 8479fa31 — allowlist-failures approach: only known failure codes return error, SUCCESS + unknown fall through to success |
| r1 | 4 | should-fix | iOS model comment wrong — propagated incorrect `receipt.result` assumption | ✅ fixed | `tron.ts:31-35` — comment now says "Present on both success ('SUCCESS') and failure. Absent only for native TRX transfers" |

---

all four findings from r1 are resolved. the fix is clean.

**praise: `TRON_RECEIPT_FAILURE_RESULTS` Set approach**

denylist of known failures rather than an allowlist is the correct choice here — new unknown `receipt.result` values (future SwapKit / Tron proto additions) will safely fall through to success rather than silently blocking users. this matches the iOS behavior and is documented in the comments.

**praise: test suite coverage**

13 tests across both `tron.test.ts` (status resolver) and balance resolver. the NeO live mainnet canary (`1540b1b3`) is pinned as a named test — solid.

one fresh observation: the test `'does not misclassify an unknown top-level result as success (falls through to receipt path)'` passes `result: 'SUCCESS'` at the top level and expects `pending` (no receipt). this is correct behavior and the invariant is sound, but the comment says "Tron RPC never emits result:'SUCCESS' at the top level" — worth a quick note that this test is asserting the *safe fallthrough*, not that we expect this input in practice. minor doc nit, not blocking.

overall: all prior blockers + should-fixes addressed. approve.

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: @NeOMakinG tackled in [a49a048b](https://github.com/vultisig/vultisig-sdk/commit/a49a048b) (R3 fix). The current predicate is `tx.receipt.result != null && TRON_RECEIPT_FAILURE_RESULTS.has(tx.receipt.result) ? 'error' : 'success'`. The `TRON_RECEIPT_FAILURE_RESULTS` set explicitly excludes 'SUCCESS' - that value falls through to 'success'. Unknown values also treated as 'success' (safer than false-failure for user). Comments updated too. tyvm 🙇

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: @NeOMakinG tackled in df2c2545 / a49a048b. Current type comment now reads 'Tron RPC only ever emits "FAILED" here (no "SUCCESS" variant exists at this level). The field is absent on success.' with link to https://developers.tron.network/reference/gettransactioninfobyid. tyvm 🙇

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: @NeOMakinG tackled in df2c2545 - type comment now documents the single-value invariant + iOS parity ref. tyvm 🙇

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: @NeOMakinG tackled in a49a048b. Receipt.result comment now reads 'Present on both success ("SUCCESS") and failure. Absent only for native TRX transfers where no smart contract executed.' iOS Swift model comment update tracked as separate cross-repo follow-up.

packages/core/chain/tx/status/resolvers/tron.ts
```diff
@@ -28,10 +49,24 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash
...
+  //    iOS ref: `if let topLevelResult = response.data.result, topLevelResult == "FAILED"`
+  //    (TronTransactionStatusProvider.swift:41 — same guard, same single-value invariant)
+  if (tx.result === 'FAILED') {
+    return { status: 'error' }
+  }
+
+  // 2. receipt absent → pending (native TRX transfers or tx still processing)
+  if (!tx.receipt) {
+    return { status: 'pending' }
+  }
```
NeOMakinG, line 63: `suggestion:` Type guard on tx.receipt could be more defensive — Pre-existing issue: checks !tx.receipt but doesn't verify typeof === 'object'. Malformed RPC responses with primitive receipt values could throw. Outside this PR's scope.

fix: File separate ticket: add typeof tx.receipt === 'object' guard before property access.

NeOMakinG (approved):
gm — multi-lane pass: . runtime-tested on SDK (TypeScript) - Unit Tests. lgtm.

## testing approach
- SDK (TypeScript) - Unit Tests

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | null | fixed |
| 1 | null | fixed |
| 1 | null | fixed |
| 2 | null | fixed |

## suggestion
### 1. Type guard on tx.receipt could be more defensive — `packages/core/chain/tx/status/resolvers/tron.ts:63`
Pre-existing issue: checks !tx.receipt but doesn't verify typeof === 'object'. Malformed RPC responses with primitive receipt values could throw. Outside this PR's scope.
action: File separate ticket: add typeof tx.receipt === 'object' guard before property access.

## risk
Low. Change is localized to Tron status resolver. 48/48 tests pass including new cases for pending, SUCCESS, and all known failure codes. iOS parity confirmed. Branch has picked up #547 feeLimit fix.

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.3-codex` adversarial pass. SDK (TypeScript) - Unit Tests.

PR #547, opened 2026-05-25
fix(tron): set feeLimit=0 for FreezeBalanceV2 and UnfreezeBalanceV2

gomesalexandre:
Tackles BUG-7 (LOW) from the 2026-05-25 Tron audit.

## what

`FreezeBalanceV2` and `UnfreezeBalanceV2` are system ops (bandwidth only), not smart-contract calls. The `feeLimit` field in the Tron protobuf is the max-energy cap for `TriggerSmartContract` only. For native staking ops the node ignores it entirely.

Before this PR, both branches were setting `feeLimit = Long.fromString(tronSpecific.gasEstimation.toString())`. That gasEstimation is an energy estimate - semantically irrelevant for bandwidth ops. The UI-visible "estimated fee" was inheriting this value and showing users a misleading non-zero figure.

Fix: set `feeLimit: Long.ZERO` in both `FREEZE:` and `UNFREEZE:` branches.

## receipts

All 5 new unit tests pass, full core suite clean (71 files, 627 tests):

```
 ✓ FREEZE:BANDWIDTH sets feeLimit to 0 regardless of gasEstimation
 ✓ FREEZE:ENERGY sets feeLimit to 0 regardless of gasEstimation
 ✓ UNFREEZE:BANDWIDTH sets feeLimit to 0 regardless of gasEstimation
 ✓ UNFREEZE:ENERGY sets feeLimit to 0 regardless of gasEstimation
 ✓ gasEstimation value does not leak into FREEZE feeLimit

 Test Files  71 passed (71)
      Tests  627 passed (627)
```

🤖 Agent-generated

## scope

This PR is **SDK-only**. iOS (`vultisig-ios/.../Tron.swift:329,368`) and Android (`vultisig-android/.../TronHelper.kt:269,298`) have the same `feeLimit = gasEstimation` issue on FreezeBalanceV2 / UnfreezeBalanceV2. Those are separate teams' codepaths; not touching them in this SDK PR. They should track follow-up issues in their respective repos. The audit (2026-05-25) flagged BUG-7 across the stack; this PR closes the SDK side.

### DIFF
--- a/packages/core/mpc/keysign/signingInputs/resolvers/tron.test.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/tron.test.ts
@@ -91,8 +91,9 @@ describe('getTronSigningInputs -- FREEZE: / UNFREEZE: feeLimit semantics (BUG-7)
     })
 
     const [input] = getTronSigningInputs({ keysignPayload: payload, walletCore })
+    // Anti-regression: prior to fix, feeLimit was passed gasEstimation
+    // (a non-zero energy estimate that's semantically meaningless for
+    // system contracts and only served to confuse the UI fee display).
     expect(input.transaction?.feeLimit?.equals(Long.ZERO)).toBe(true)
-    // Explicit anti-regression: feeLimit must NOT equal gasEstimation.
-    expect(input.transaction?.feeLimit?.toNumber()).not.toBe(Number(GAS_ESTIMATION))
   })
 })
--- a/packages/core/mpc/keysign/signingInputs/resolvers/tron.test.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/tron.test.ts
@@ -1,0 +1,98 @@
+import { create } from '@bufbuild/protobuf'
+import { type WalletCore } from '@trustwallet/wallet-core'
+import { Chain } from '@vultisig/core-chain/Chain'
+import { TronSpecificSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/blockchain_specific_pb'
+import { CoinSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/coin_pb'
+import { KeysignPayloadSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
+import Long from 'long'
+import { describe, expect, it } from 'vitest'
+
+import { getTronSigningInputs } from './tron'
+
+// getTronSigningInputs does not use walletCore internally (no address
+// validation or signing helpers required for Tron system ops), so a
+// bare cast to satisfy the resolver type constraint is fine here.
+const walletCore = {} as unknown as WalletCore
+
+// Minimal TronSpecific with a nonzero gasEstimation so we can assert
+// it is NOT forwarded to feeLimit in system-contract branches.
+const makeTronSpecific = (gasEstimation = 100_000_000n) =>
+  create(TronSpecificSchema, {
+    timestamp: 1_700_000_000_000n,
+    expiration: 1_700_003_600_000n,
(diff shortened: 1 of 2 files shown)


gomesalexandre (commented):
## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | blocker (out-of-scope) | iOS Tron.swift:329,368 same feeLimit bug | ✅ documented | `## scope` section in PR body explicitly calls out Tron.swift:329,368 + TronHelper.kt:269,298, flags need for follow-up in native repos |
| r1 | 2 | blocker (out-of-scope) | Android TronHelper.kt:269,298 same feeLimit bug | ✅ documented | same `## scope` section |
| r1 | 3 | suggestion | redundant `not.toBe(Number(GAS_ESTIMATION))` assertion | ✅ fixed | c957865b — dropped the check, replaced with inline comment explaining pre-fix behaviour |

---

## r2 findings

**1. praise: fix is surgically correct, zero blast radius**

Both staking branches now emit `feeLimit: Long.ZERO` (branch state lines 61, 94). The 4 smart-contract / token-transfer callsites that legitimately pass `gasEstimation` are untouched (TriggerSmartContract line 138, transferAsset line 149, TRC-20 swap line 221, TRC-20 native line 284). Exactly the right minimal diff.

**2. praise: proto3 default suppression is correctly leveraged**

`Long.ZERO` serializes as the absent field on the wire under proto3's default-value suppression rule. No explicit `clearFeeLimit()` needed; wallet-core will not include the field at all. Matches Tron node expectations for system ops. Good.

**3. suggestion: test 5 inline comment is mildly misleading**

The comment in the final state of test 5 says "prior to fix, feeLimit was passed gasEstimation" but the test only covers the FREEZE path (memo `FREEZE:ENERGY`). A reader might wonder whether UNFREEZE got the same treatment. Tests 3+4 cover UNFREEZE separately which is fine, but the comment in test 5 could say "FREEZE or UNFREEZE" to be unambiguous. Non-blocking, the 4 prior tests are clear enough.

**4. q: `makeTronSpecific` default gasEstimation (100_000_000n) is hardcoded in `buildPayload` call too**

`buildPayload` explicitly passes `makeTronSpecific(100_000_000n)` rather than using the default param. That means the default param on `makeTronSpecific` is dead code - the `gasEstimation = 100_000_000n` default is never exercised. Not a bug (the value is the same), just noise. Non-blocking.

**5. scope section - accurate and complete**

Pinpoints the exact file:line for both iOS and Android. References the 2026-05-25 audit. States explicitly that SDK PR does not touch native repos. This is exactly what was needed after r1.

**6. no accidental import changes, no leftover comment cruft**

Both commits touch only `tron.ts` and `tron.test.ts`. No whitespace-only diffs, no stray `console.log`, no import reordering. Clean.

---

## verdict

Zero `blocker:` findings. Zero `preferably-blocking:` findings. The two r1 out-of-scope blockers are correctly handled via documentation rather than code (standing convention). The suggestions above are genuinely non-blocking.

**APPROVE** — pending NeO's external stamp per the four-eyes rule.

NeOMakinG (approved):
gm — multi-lane pass: claude-correctness + glm-chain-edge. conceptual-only (no runtime this pass). lgtm.

## risk
Zero blast radius if merged today. The fix is gated behind early-return FREEZE:/UNFREEZE: memo checks that don't share code with any other tx type. All non-staking paths (TriggerSmartContract, TRC-20 transfers, asset transfers) retain their `Long.fromString(gasEstimation)` feeLimit and are regression-protected by existing tests that assert feeLimit > 0 and reject feeLimit === 0.

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `glm-4.6` conceptual pass.

PR #548, opened 2026-05-25
fix(polkadot): use transfer_keep_alive (method 3) instead of transfer_allow_death (method 0)

gomesalexandre:
Tackles BUG-1 (MEDIUM) from the 2026-05-25 Polkadot audit.

## what

`packages/core/mpc/keysign/signingInputs/resolvers/polkadot.ts:29` was using `methodIndex: 0` which maps to `pallet_balances::transfer_allow_death` on Asset Hub (statemint). This call can permanently reap the sender's account when the remaining balance drops below the existential deposit (0.01 DOT on Asset Hub).

`methodIndex: 3` = `transfer_keep_alive` - it reverts the tx rather than destroying the account.

## context

The app and mcp-ts already had this right:
- `vultiagent-app/src/features/agent/lib/polkadotTx.ts:194`: `METHOD_TRANSFER_KEEP_ALIVE = 3`
- `vultisig-mcp-ts`: `api.tx.balances.transferKeepAlive`

The SDK was the outlier. iOS and Android also use method 0 but those are separate teams - separate PRs needed.

## receipts

jest / vitest output:

```
PASS packages/core/mpc/keysign/signingInputs/resolvers/polkadot.test.ts
  getPolkadotSigningInputs
    ✓ uses methodIndex 3 (transfer_keep_alive) not 0 (transfer_allow_death) 8ms
    ✓ keeps moduleIndex 10 (pallet_balances on Asset Hub) 0ms
```

🤖 Agent-generated

## cross-client tracking

iOS and Android still ship `methodIndex: 0` (transfer_allow_death). They are out of scope here (different teams) but the divergence is real:
- If a vault has mixed iOS + agent-app devices participating in the same MPC keysign session for Polkadot, the SCALE pre-image hashes will diverge → silent session failure (no fund loss, just user confusion).
- Single-device fast-vault flows are unaffected.

Companion work to track separately:
- iOS: `Polkadot.swift:74` (`methodIndex: 0`) + the misleading comment at line 60-70 saying "WalletCore encodes as TransferAllowDeath when assetId=0" - verified against TW 4.6.9 WASM that this is NOT what happens; TW respects custom callIndices.
- Android: `PolkadotHelper.kt:47` (`methodIndex: 0`)

closes #N/A (no GH issue yet for the audit BUG-1)

🤖 codex adversarial verdict: APPROVE (no correctness blockers; cross-client tracking is coordination work)

### DIFF
--- a/packages/core/mpc/keysign/signingInputs/resolvers/polkadot.test.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/polkadot.test.ts
@@ -1,0 +1,63 @@
+import { create } from '@bufbuild/protobuf'
+import { initWasm, type WalletCore } from '@trustwallet/wallet-core'
+import { Chain } from '@vultisig/core-chain/Chain'
+import { PolkadotSpecificSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/blockchain_specific_pb'
+import { CoinSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/coin_pb'
+import { KeysignPayloadSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
+import { beforeAll, describe, expect, it } from 'vitest'
+
+import { getPolkadotSigningInputs } from './polkadot'
+
+// Polkadot Asset Hub genesis hash (statemint)
+const GENESIS_HASH = '0x68d56f15f85d3136970ec16946040bc1752654e906147f7e43e9d539d7c3de2f'
+// Arbitrary valid SS58-0 Polkadot address
+const TO_ADDRESS = '15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6Srd'
+const FROM_ADDRESS = '14E5nqKAp3oAJcmzgs25fyAmgeNL66XceFLiTqAZkdVH5T38'
+const BLOCK_HASH = '0xaabbccddeeff00112233445566778899aabbccddeeff00112233445566778899'
+
+const buildPayload = () =>
+  create(KeysignPayloadSchema, {
+    coin: create(CoinSchema, {
+      chain: Chain.Polkadot,
+      ticker: 'DOT',
+      address: FROM_ADDRESS,
+      decimals: 10,
+      isNativeToken: true,
+    }),
+    toAddress: TO_ADDRESS,
+    toAmount: '10000000000',
+    blockchainSpecific: {
+      case: 'polkadotSpecific',
+      value: create(PolkadotSpecificSchema, {
+        recentBlockHash: BLOCK_HASH,
+        nonce: 0n,
+        currentBlockNumber: '20000000',
+        specVersion: 1003004,
+        transactionVersion: 26,
+        genesisHash: GENESIS_HASH,
+      }),
+    },
+  })
+
+describe('getPolkadotSigningInputs', () => {
+  let walletCore: WalletCore
+
+  beforeAll(async () => {
+    walletCore = await initWasm()
+  })
+
+  it('uses methodIndex 3 (transfer_keep_alive) not 0 (transfer_allow_death)', () => {
+    const [input] = getPolkadotSigningInputs({ keysignPayload: buildPayload(), walletCore })
+
+    const callIndices = input.balanceCall?.assetTransfer?.callIndices?.custom
+    expect(callIndices).toBeDefined()
+    expect(callIndices?.methodIndex).toBe(3)
+  })
+
+  it('keeps moduleIndex 10 (pallet_balances on Asset Hub)', () => {
+    const [input] = getPolkadotSigningInputs({ keysignPayload: buildPayload(), walletCore })
+
+    const callIndices = input.balanceCall?.assetTransfer?.callIndices?.custom
+    expect(callIndices?.moduleIndex).toBe(10)
+  })
+})
--- a/packages/core/mpc/keysign/signingInputs/resolvers/polkadot.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/polkadot.ts
@@ -23,10 +23,14 @@ export const getPolkadotSigningInputs: SigningInputsResolver<'polkadot'> = ({ ke
   // Amount: converted to hexadecimal, stripped of '0x'
   const amountHex = Buffer.from(stripHexPrefix(bigIntToHex(BigInt(keysignPayload.toAmount))), 'hex')
 
+  // methodIndex 3 = transfer_keep_alive on Asset Hub (statemint pallet_balances).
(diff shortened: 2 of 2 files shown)


packages/core/mpc/keysign/signingInputs/resolvers/polkadot.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    walletCore = await initWasm()
+  })
+
+  it('uses methodIndex 3 (transfer_keep_alive) not 0 (transfer_allow_death)', () => {
+    const [input] = getPolkadotSigningInputs({ keysignPayload: buildPayload(), walletCore })
+
+    const callIndices = input.balanceCall?.assetTransfer?.callIndices?.custom
+    expect(callIndices).toBeDefined()
+    expect(callIndices?.methodIndex).toBe(3)
+  })
```
NeOMakinG, line 55: `should-fix:` test pins structural values only — no SCALE byte-vector assertion — Both tests decode `callIndices?.methodIndex` from the protobuf struct after calling `getPolkadotSigningInputs`. They verify the in-memory proto field value (3), not the actual SCALE-encoded bytes that end up on the wire. If TW ever changes how `CustomCallIndices` maps to wire bytes — e.g. swaps the byte order or changes the call-indices variant encoding — these tests pass green while the broadcast extrinsic is malformed.

fix: Add a third test case that calls `getEncodedSigningInputs` (which runs the full TW proto → SCALE encode pipeline), hex-encodes the resulting `Uint8Array`, and asserts it contains the substring `0a03` at the expected offset — `0x0a` = moduleIndex 10, `0x03` = methodIndex 3, back-to-back as a raw SCALE call-indices pair. One golden-byte assertion gives the suite real regression coverage against TW encoding changes.

packages/core/mpc/keysign/signingInputs/resolvers/polkadot.ts
```diff
@@ -23,10 +23,14 @@ export const getPolkadotSigningInputs: SigningInputsResolver<'polkadot'> = ({ ke
   // Amount: converted to hexadecimal, stripped of '0x'
   const amountHex = Buffer.from(stripHexPrefix(bigIntToHex(BigInt(keysignPayload.toAmount))), 'hex')
 
+  // methodIndex 3 = transfer_keep_alive on Asset Hub (statemint pallet_balances).
```
NeOMakinG, line 26: `suggestion:` document the cross-client hash-divergence window for paired-vault users — The PR description calls out that iOS (`Polkadot.swift:74`) and Android (`PolkadotHelper.kt:47`) still use `methodIndex: 0`. In a 2-of-2 paired vault where an iOS device is a co-signer, both devices independently compute the signing preimage from the `KeysignPayload`. iOS builds SCALE bytes `[10, 0]`; the SDK builds `[10, 3]`. The hashes diverge and the MPC session fails silently (no fund loss, but a confusing UX dead-end for the user). The PR description notes this, but the code has no inline comment pointing at the risk for future maintainers.

fix: Add a code comment above `methodIndex: 3` noting: `// NOTE: iOS (Polkadot.swift) and Android (PolkadotHelper.kt) still use methodIndex 0 (transfer_allow_death). Paired-vault keysign sessions that include an iOS/Android co-signer will produce divergent SCALE hashes and fail. Fast-vault (server-side) flows are unaffected. Track companion PRs before advertising paired-vault DOT send.`

NeOMakinG (approved):
gm — multi-lane pass: claude-correctness + codex-adversarial + glm-third-eye. conceptual-only (no runtime this pass). lgtm.

## should-fix
### 1. test pins structural values only — no SCALE byte-vector assertion — `packages/core/mpc/keysign/signingInputs/resolvers/polkadot.test.ts:55`
Both tests decode `callIndices?.methodIndex` from the protobuf struct after calling `getPolkadotSigningInputs`. They verify the in-memory proto field value (3), not the actual SCALE-encoded bytes that end up on the wire. If TW ever changes how `CustomCallIndices` maps to wire bytes — e.g. swaps the byte order or changes the call-indices variant encoding — these tests pass green while the broadcast extrinsic is malformed.
action: Add a third test case that calls `getEncodedSigningInputs` (which runs the full TW proto → SCALE encode pipeline), hex-encodes the resulting `Uint8Array`, and asserts it contains the substring `0a03` at the expected offset — `0x0a` = moduleIndex 10, `0x03` = methodIndex 3, back-to-back as a raw SCALE call-indices pair. One golden-byte assertion gives the suite real regression coverage against TW encoding changes.

## suggestion
### 1. document the cross-client hash-divergence window for paired-vault users — `packages/core/mpc/keysign/signingInputs/resolvers/polkadot.ts:26`
The PR description calls out that iOS (`Polkadot.swift:74`) and Android (`PolkadotHelper.kt:47`) still use `methodIndex: 0`. In a 2-of-2 paired vault where an iOS device is a co-signer, both devices independently compute the signing preimage from the `KeysignPayload`. iOS builds SCALE bytes `[10, 0]`; the SDK builds `[10, 3]`. The hashes diverge and the MPC session fails silently (no fund loss, but a confusing UX dead-end for the user). The PR description notes this, but the code has no inline comment pointing at the risk for future maintainers.
action: Add a code comment above `methodIndex: 3` noting: `// NOTE: iOS (Polkadot.swift) and Android (PolkadotHelper.kt) still use methodIndex 0 (transfer_allow_death). Paired-vault keysign sessions that include an iOS/Android co-signer will produce divergent SCALE hashes and fail. Fast-vault (server-side) flows are unaffected. Track companion PRs before advertising paired-vault DOT send.`

## risk
Fund-safety improvement: replacing `transfer_allow_death` with `transfer_keep_alive` removes the account-reap risk on Asset Hub. The change is consistent with the existing app layer (`polkadotTx.ts`) and mcp-ts, both of which already use method 3. Blast radius is Polkadot DOT sends only; all other chains are unaffected. Paired-vault users with iOS/Android co-signers will see MPC session failures until companion PRs land, but no funds are at risk.

## verdict
APPROVE — ship it 🫡

packages/core/mpc/keysign/signingInputs/resolvers/polkadot.test.ts
gomesalexandre: @NeOMakinG fair shout. Tracked as [#557](https://github.com/vultisig/vultisig-sdk/issues/557) for the golden SCALE-byte assertion. Treating as non-blocking (`should-fix:`) for this PR + yeeting. tyvm 🙇

packages/core/mpc/keysign/signingInputs/resolvers/polkadot.ts
gomesalexandre: @NeOMakinG tracked the cross-client divergence as [vultiagent-app#851](https://github.com/vultisig/vultiagent-app/issues/851) (iOS + Android methodIndex parity). Non-blocking (`suggestion:`) - yeeting current. tyvm 🙇

PR #550, opened 2026-05-25
fix(polkadot/dapp): add missing asset hub signed-extension bytes to constructPolkadotSigningPayload

gomesalexandre:
Tackles BUG-2 (HIGH, latent) from the 2026-05-25 Polkadot audit.

## what

`constructAssetHubPolkadotSigningPayload` (renamed from `constructPolkadotSigningPayload` - see R1/F4 below) in `packages/core/chain/chains/polkadot/dapp/` was missing three bytes required by the Asset Hub Polkadot runtime's signed-extension list:

| Signed extension | field | value |
|---|---|---|
| `ChargeAssetTxPayment` | `extra` - `Option<MultiLocation>` | `None` = `0x00` |
| `CheckMetadataHash` | `extra` - mode byte | `Disabled` = `0x00` |
| `CheckMetadataHash` | `additional_signed` - `Option<H256>` | `None` = `0x00` |

Without these three bytes the node computes a different payload hash from what the wallet signed. Result: `BadProof` rejection on-chain for any extrinsic signed via this helper.

## impact

No callers today - this is a latent bug. Any future walletconnect or Polkadot extension integration that routes through this function would produce invalid signatures immediately.

## source of truth

Byte sequence ported from `vultiagent-app/src/services/polkadotTx.ts:362-378` (`assembleSignerPayloadInner`), which has the correct five-byte AH shape and is covered by the golden-fixture assertion in `polkadotTx.test.ts` (the `assembleSignerPayload` describe block). That fixture was verified against `@polkadot/api` and the live Asset Hub runtime (statemint 2001681, `0x001e8b11`).

The new byte ordering matches that app source of truth exactly:
```
method || era || nonce || tip
  || ChargeAssetTxPayment(0x00)    <- was missing
  || CheckMetadataHash mode(0x00) <- was missing
  || specVersion || txVersion || genesisHash || blockHash
  || CheckMetadataHash additional-signed(0x00) <- was missing
```

## R1 review tackles

- **F4 (rename + scope guard)**: renamed to `constructAssetHubPolkadotSigningPayload` to make the AH-only scope unambiguous. Added runtime guard: if `signedExtensions` is provided and non-empty, throws if `ChargeAssetTxPayment` or `CheckMetadataHash` is absent - prevents BadProof corruption on relay-chain payloads passed by mistake.
- **F3 (fixture comment)**: `GENESIS_HASH` comment updated to accurately describe the fixture bytes ("First 8 bytes are real AH genesis prefix; rest is fixture-stable zeros + sentinel `aa` trailer").
- **F6 (boundary test)**: added the `= 256` boundary case - `method=177B` -> total=256 -> no hash (condition is `> 256`). The `> 256` case was already covered; both are now pinned.
- **F9 (specVersion)**: `0x001e8b11` decodes to `2001681`, not `2002001`. Fixed in PR body and in the test file comment.

## tests

`constructSigningPayload.test.ts` - five vitest assertions (was four):
- golden-fixture hex match against the canonical AH layout (ported from the app fixture)
- raw length = 118 bytes, exactly 3 more than the pre-fix shape
- raw = 256 bytes exactly does NOT hash (boundary pin, new)
- raw > 256 bytes returns blake2b-256 hash (32 bytes)
- tip=0 compact encoding lands at the correct byte offset

## receipts

```
 RUN  v4.1.6

 Test Files  71 passed (71)
      Tests  627 passed (627)
   Start at  11:01:43
   Duration  4.30s
```

All five polkadot tests pass. tsc --project .config/tsconfig.json --noEmit exits 0.

🤖 Agent-generated

## R2 tackle (ad66ad74)

**should-fix: empty-array bypass in signedExtensions guard**

The R1 guard used `payload.signedExtensions && payload.signedExtensions.length > 0` as the condition. A dApp passing `signedExtensions: []` makes `length > 0` false, so the entire guard short-circuits and AH bytes get appended to a payload that explicitly declared zero extensions - exactly the corruption the guard was meant to prevent.

Fix: changed to `payload.signedExtensions != null`. Any provided array, including empty, is now validated against the required ChargeAssetTxPayment/CheckMetadataHash extensions.

New boundary test: `signedExtensions: []` must throw, not pass through. Tests: 5 -> 6.

```
 RUN  v4.1.6

 Test Files  1 passed (1)
      Tests  6 passed (6)
   Start at  11:23:29
   Duration  340ms
```

## Summary by CodeRabbit

* **New Features**
  * Added support for Asset Hub Polkadot network transactions with specialized signing requirements.
  * Enhanced swap fee tracking with detailed chain, token, and decimal information across all swap types.

* **Bug Fixes**
  * Improved swap fee chain resolution with automatic fallback when unmapped chains are encountered.

* **Tests**
  * Added comprehensive test suite for Asset Hub Polkadot payload construction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/550?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
+++ b/packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
@@ -6,7 +6,11 @@ import { PolkadotSignerPayloadJSON } from './PolkadotSignerPayload'
 const polkadotSigningPayloadHashThreshold = 256
 
 /**
- * Construct the raw signing payload bytes from a Polkadot SignerPayloadJSON.
+ * Construct the raw signing payload bytes from an Asset Hub Polkadot SignerPayloadJSON.
+ *
+ * SCOPE: Asset Hub (statemint/statemine) payloads ONLY. The payload must include
+ * both `ChargeAssetTxPayment` and `CheckMetadataHash` in `signedExtensions`.
+ * Relay-chain payloads use different extensions and must NOT use this function.
  *
  * Follows the Asset Hub Polkadot (statemint) extrinsic payload v4 encoding:
  * method + era + compact(nonce) + compact(tip)
@@ -24,7 +28,16 @@ const polkadotSigningPayloadHashThreshold = 256
  *
  * If the payload exceeds 256 bytes, it is blake2b-256 hashed before signing.
  */
-export const constructPolkadotSigningPayload = (payload: PolkadotSignerPayloadJSON): Uint8Array => {
+export const constructAssetHubPolkadotSigningPayload = (payload: PolkadotSignerPayloadJSON): Uint8Array => {
+  const required = ['ChargeAssetTxPayment', 'CheckMetadataHash']
+  if (payload.signedExtensions && payload.signedExtensions.length > 0) {
+    const missing = required.filter(ext => !payload.signedExtensions!.includes(ext))
+    if (missing.length > 0) {
+      throw new Error(
+        `constructAssetHubPolkadotSigningPayload: payload signedExtensions missing required Asset Hub extensions: ${missing.join(', ')}. This function only encodes Asset Hub (statemint) payloads.`
+      )
+    }
+  }
   const method = hexToU8a(payload.method)
   const era = hexToU8a(payload.era)
   const nonce = compactToU8a(parseInt(payload.nonce, 16))
--- a/packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
+++ b/packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
@@ -8,9 +8,19 @@ const polkadotSigningPayloadHashThreshold = 256
 /**
  * Construct the raw signing payload bytes from a Polkadot SignerPayloadJSON.
  *
- * Follows the Polkadot extrinsic payload v4 encoding:
- * method + era + compact(nonce) + compact(tip) + LE-u32(specVersion) +
- * LE-u32(transactionVersion) + genesisHash + blockHash
+ * Follows the Asset Hub Polkadot (statemint) extrinsic payload v4 encoding:
+ * method + era + compact(nonce) + compact(tip)
+ * + ChargeAssetTxPayment::extra (Option<MultiLocation>=None = 0x00)
+ * + CheckMetadataHash::extra (mode=Disabled = 0x00)
+ * + LE-u32(specVersion) + LE-u32(transactionVersion)
+ * + genesisHash + blockHash
+ * + CheckMetadataHash::additional_signed (Option<H256>=None = 0x00)
+ *
+ * The three Asset Hub signed-extension bytes (ChargeAssetTxPayment option byte,
+ * CheckMetadataHash mode byte, CheckMetadataHash additional-signed byte) are
+ * required by the Asset Hub runtime. Without them the node rejects the
+ * extrinsic with BadProof because the payload hash the node computes differs
+ * from the one the wallet signed.
  *
  * If the payload exceeds 256 bytes, it is blake2b-256 hashed before signing.
  */
@@ -19,6 +29,11 @@ export const constructPolkadotSigningPayload = (payload: PolkadotSignerPayloadJS
   const era = hexToU8a(payload.era)
   const nonce = compactToU8a(parseInt(payload.nonce, 16))
   const tip = compactToU8a(payload.tip ? BigInt(payload.tip) : 0n)
+
+  // ChargeAssetTxPayment::extra - Option<MultiLocation> = None
+  const chargeAssetTxPaymentNone = new Uint8Array([0x00])
+  // CheckMetadataHash::extra - mode byte = Disabled
+  const checkMetadataHashMode = new Uint8Array([0x00])
 
   const specVersion = new Uint8Array(4)
   new DataView(specVersion.buffer).setUint32(0, parseInt(payload.specVersion, 16), true)
@@ -29,7 +44,22 @@ export const constructPolkadotSigningPayload = (payload: PolkadotSignerPayloadJS
   const genesisHash = hexToU8a(payload.genesisHash)
   const blockHash = hexToU8a(payload.blockHash)
 
-  const raw = u8aConcat(method, era, nonce, tip, specVersion, transactionVersion, genesisHash, blockHash)
+  // CheckMetadataHash::additional_signed - Option<H256> = None
+  const checkMetadataHashNone = new Uint8Array([0x00])
+
+  const raw = u8aConcat(
+    method,
+    era,
+    nonce,
+    tip,
+    chargeAssetTxPaymentNone,
+    checkMetadataHashMode,
+    specVersion,
+    transactionVersion,
+    genesisHash,
+    blockHash,
+    checkMetadataHashNone
+  )
 
   if (raw.length > polkadotSigningPayloadHashThreshold) {
     return blake2AsU8a(raw, 256)
--- a/packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
+++ b/packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
@@ -30,7 +30,7 @@ const polkadotSigningPayloadHashThreshold = 256
  */
 export const constructAssetHubPolkadotSigningPayload = (payload: PolkadotSignerPayloadJSON): Uint8Array => {
   const required = ['ChargeAssetTxPayment', 'CheckMetadataHash']
-  if (payload.signedExtensions && payload.signedExtensions.length > 0) {
+  if (payload.signedExtensions != null) {
     const missing = required.filter(ext => !payload.signedExtensions!.includes(ext))
     if (missing.length > 0) {
       throw new Error(
--- a/packages/core/chain/chains/polkadot/dapp/constructSigningPayload.test.ts
+++ b/packages/core/chain/chains/polkadot/dapp/constructSigningPayload.test.ts
@@ -1,0 +1,148 @@
+import { describe, expect, it } from 'vitest'
+
+import { constructPolkadotSigningPayload } from './constructSigningPayload'
+import { PolkadotSignerPayloadJSON } from './PolkadotSignerPayload'
+
+// ---------------------------------------------------------------------------
+// Byte-fixture test for the Asset Hub Polkadot signed-extension shape.
+//
+// Ported from vultiagent-app/src/services/__tests__/polkadotTx.test.ts
+// (assembleSignerPayload describe block) which serves as the canonical
+// source of truth. That fixture was verified against @polkadot/api and
+// the live Asset Hub runtime (statemint 2002001).
+//
+// The three AH-required bytes are:
+//   ChargeAssetTxPayment::extra  Option<MultiLocation>=None   0x00
+//   CheckMetadataHash::extra     mode=Disabled                0x00
+//   CheckMetadataHash::additional_signed  Option<H256>=None   0x00
+//
+// Without these the runtime computes a different payload hash than the
+// wallet - BadProof on-chain.
+// ---------------------------------------------------------------------------
+
+const toHex = (b: Uint8Array) => Array.from(b, x => x.toString(16).padStart(2, '0')).join('')
+
+// Deterministic test vectors - no live RPC.
+//
+// method: Balances.transferKeepAlive(Alice, 100_000_000 Planck)
+//   pallet 0x0a | method 0x03 | MultiAddress::Id 0x00 | Alice 32B | compact(100_000_000)
+const METHOD = '0x0a0300d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d0284d717'
+
+// era: mortal (blockNumber=42, period=64) - 0xa502
+const ERA = '0xa502'
+
+// nonce: 0 - compact 0x00
+const NONCE = '0x0'
+
+// tip: 0
+const TIP = '0x0'
+
+// specVersion: statemint 2002001 = 0x001E8B11
+const SPEC_VERSION = '0x001e8b11'
+
+// transactionVersion: 15 = 0x0f
+const TX_VERSION = '0x0000000f'
+
+// genesisHash: 32-byte AH genesis (live prefix pinned, rest zeros for fixture stability)
+const GENESIS_HASH = '0x68d56f15f4d3e1ec0000000000000000000000000000000000000000000000aa'
+
+// blockHash: 32 bytes of 0x11 (deterministic stand-in for a real block hash)
+const BLOCK_HASH = '0x1111111111111111111111111111111111111111111111111111111111111111'
+
+const basePayload: PolkadotSignerPayloadJSON = {
+  address: '15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6Sp5',
+  blockHash: BLOCK_HASH,
+  blockNumber: '0x00',
+  era: ERA,
+  genesisHash: GENESIS_HASH,
+  method: METHOD,
+  nonce: NONCE,
+  specVersion: SPEC_VERSION,
+  tip: TIP,
+  transactionVersion: TX_VERSION,
+  signedExtensions: [
+    'CheckNonZeroSender',
+    'CheckSpecVersion',
+    'CheckTxVersion',
+    'CheckGenesis',
+    'CheckMortality',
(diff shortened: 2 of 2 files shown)


gomesalexandre (commented):
## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | `signedExtensions` ignored - wrong bytes for relay-chain payloads | ✅ fixed | rename at `constructSigningPayload.ts:31`, guard at `L33-39`, 3eb36680 |
| r1 | 2 | should-fix | no boundary test for `= 256` (only `> 256` was covered) | ✅ fixed | `constructSigningPayload.test.ts:131-137`, b29cdddb |
| r1 | 3 | suggestion | specVersion typo in PR body | ✅ fixed | PR body + test comment now say `2001681 (0x001e8b11)` |

---

## R2 findings

### should-fix: guard short-circuits on empty `signedExtensions` array

**file:** `packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts:33`

```ts
if (payload.signedExtensions && payload.signedExtensions.length > 0) {
```

The `length > 0` check means a caller passing `signedExtensions: []` silently bypasses the guard - the function then appends all three AH bytes to a payload that isn't AH. The type says `string[]` (non-optional) so an empty array is valid TS and a real dApp could produce it (some older Polkadot extension versions omit extensions in the payload or populate an empty list for disconnected runtimes).

Fix is one character: drop the `.length > 0` branch. Always run the missing-extension check regardless of array length:

```ts
const missing = required.filter(ext => !payload.signedExtensions.includes(ext))
if (missing.length > 0) {
  throw new Error(...)
}
```

Non-blocking to approve since there are no callers today (latent), but this needs to land before any dApp integration ships.

---

### should-fix: stale old function name in `vultiagent-app` JSDoc comment

**file:** `vultiagent-app/src/services/polkadotTx.ts:220`

```
 * The encoding paths follow v4 signed extrinsic format and have been manually cross-checked against the
 * SDK's `constructPolkadotSigningPayload`.
```

Comment still says `constructPolkadotSigningPayload` (old name). No runtime impact - there's no import of the function there (polkadotTx.ts rolls its own inline helpers) - but a future dev grepping for the old name will be confused. Should update to `constructAssetHubPolkadotSigningPayload`. File is outside this PR's diff but the PR caused the rename, so worth noting as a companion fix.

---

## what was verified clean

**rename consistency** - only 2 files in the SDK reference either name (impl + test). No barrel re-export, no barrel index. Package `exports` map at `./chains/polkadot/dapp/constructSigningPayload` resolves to the dist `.d.ts` which correctly declares `constructAssetHubPolkadotSigningPayload`. No callers in mcp-ts or vultiagent-app import the function at all (only JSDoc refs). Rename is safe.

**byte ordering** - impl matches the polkadotTx.ts `assembleSignerPayloadInner` source of truth exactly: `method || era || nonce || tip || chargeAssetTxPaymentNone || checkMetadataHashMode || specVersion || txVersion || genesisHash || blockHash || checkMetadataHashNone`. Correct per Substrate SignedPayload v4 encoding.

**boundary test arithmetic** - independently verified: non-method bytes = era(2) + nonce(1) + tip(1) + AH-ext-payload(2) + spec(4) + txv(4) + genesis(32) + block(32) + AH-additional-signed(1) = 79. method=177 → total=256 → `> 256` is false → raw returned. 5/5 tests cover all four original cases + the boundary.

**specVersion** - `0x001e8b11` = 2001681 confirmed by arithmetic. LE encoding `118b1e00` confirmed. The test fixture comment is accurate.

**ChargeAssetTxPayment byte** - contributes 1 byte (`Option<AssetId>=None=0x00`) to the extra section of the signed payload. The standard tip field covers the balance tip; ChargeAssetTxPayment only injects the asset-id discriminant. `0x00` is correct for both `Option<MultiLocation>` and `Option<AssetId>` variants (the comment says MultiLocation which is the older name - minor imprecision, byte is right).

**suite** - 627 tests green, tsc exits 0.

---

## verdict

APPROVE with the two should-fix items tracked. The R1 preferably-blocking is cleanly resolved. The guard hole (empty array bypass) is latent-only given zero callers today - comfortable approving and tracking it as a follow-up issue before any dApp integration.

packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
```diff
@@ -6,20 +6,48 @@ import { PolkadotSignerPayloadJSON } from './PolkadotSignerPayload'
...
+ * required by the Asset Hub runtime. Without them the node rejects the
+ * extrinsic with BadProof because the payload hash the node computes differs
+ * from the one the wallet signed.
  *
  * If the payload exceeds 256 bytes, it is blake2b-256 hashed before signing.
  */
-export const constructPolkadotSigningPayload = (payload: PolkadotSignerPayloadJSON): Uint8Array => {
+export const constructAssetHubPolkadotSigningPayload = (payload: PolkadotSignerPayloadJSON): Uint8Array => {
+  const required = ['ChargeAssetTxPayment', 'CheckMetadataHash']
+  if (payload.signedExtensions && payload.signedExtensions.length > 0) {
```
gomesalexandre: **should-fix: empty `signedExtensions` array bypasses the guard**

```ts
if (payload.signedExtensions && payload.signedExtensions.length > 0) {
```

The `length > 0` check means a caller passing `signedExtensions: []` short-circuits the whole check - function then silently appends AH bytes to a non-AH payload. The type says `string[]` (non-optional, not never-empty), so an empty array is valid TS.

Drop the length branch entirely:

```ts
const missing = required.filter(ext => !payload.signedExtensions.includes(ext))
if (missing.length > 0) {
  throw new Error(...)
}
```

Latent-only today (no callers), but should land before any dApp integration.

NeOMakinG (changes requested):

## prior issues — status
| round | issue | status |
|---|---|---|
| r1 | signedExtensions ignored — wrong bytes for relay-chain payloads | fixed |
| r1 | no boundary test for raw=256 (only >256 covered) | fixed |
| r1 | specVersion typo in PR body | fixed |
| r2 | guard short-circuits on empty signedExtensions array (length > 0 bypass) | fixed |

## risk
The Polkadot AH signing fix is correct and the golden-fixture test suite is solid. The blocker is the undocumented revert of #540's swap-fee-coin-context work bundled into this PR — if merged, cosigners lose EVM LI.FI affiliate-fee attribution and THORChain native swap fee display.

## verdict
REQUEST CHANGES

packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
gomesalexandre: tackled in 7d5093b9ae77b9ed00aa087d1905723cfa714437: root cause was the branch was cut from `50118d6a` (pre-#540 merge). #540's 6 files were never in this worktree — they showed as reverts in the diff but were just absent. Restored all 6 from `origin/main`:
- `getLifiSwapQuote.ts`: resolveSwapFeeChain + EVM affiliateFee extraction/wiring
- `build.ts`: swapFee/swapFeeChain/swapFeeTokenId/swapFeeDecimals forwarding for EVM + Solana
- `mapSwapPayload.ts`: swap_fee_chain/token_id/decimals roundtrip fields
- `nativeSwapQuoteToSwapPayload.ts`: swapFee wiring for native swaps
- `1inch_swap_payload_pb.ts`: proto fields 8/9/10
- `react-native/overrides/getLifiSwapQuote.ts`: EVM affiliateFee shape mirrored in RN

`git diff origin/main --stat` now shows ONLY the 2 polkadot/dapp files. 628 tests green.

packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
```diff
@@ -6,20 +6,48 @@ import { PolkadotSignerPayloadJSON } from './PolkadotSignerPayload'
...
+export const constructAssetHubPolkadotSigningPayload = (payload: PolkadotSignerPayloadJSON): Uint8Array => {
+  const required = ['ChargeAssetTxPayment', 'CheckMetadataHash']
+  if (payload.signedExtensions != null) {
+    const missing = required.filter(ext => !payload.signedExtensions!.includes(ext))
+    if (missing.length > 0) {
+      throw new Error(
+        `constructAssetHubPolkadotSigningPayload: payload signedExtensions missing required Asset Hub extensions: ${missing.join(', ')}. This function only encodes Asset Hub (statemint) payloads.`
+      )
+    }
+  }
```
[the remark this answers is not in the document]
gomesalexandre: dis tackled in ad66ad74 (R2) - guard now unconditional for AH payloads @coderabbitai

gomesalexandre (commented):
## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | `signedExtensions` ignored - wrong bytes for relay-chain payloads | ✅ fixed | rename at `constructSigningPayload.ts:31`, guard at L33-39, `3eb36680` |
| r1 | 2 | should-fix | no boundary test for `= 256` (only `> 256` covered) | ✅ fixed | `constructSigningPayload.test.ts:131-137`, `b29cdddb` |
| r1 | 3 | suggestion | specVersion typo in PR body | ✅ fixed | PR body + test comment say `2001681 (0x001e8b11)` |
| r2 | 1 | should-fix | empty `signedExtensions` array bypasses guard (`length > 0` check) | ✅ fixed | `constructSigningPayload.test.ts:147-149`, `ad66ad74` |
| r3 | 1 | preferably-blocking | branch bundled ghost-revert of #540's swap-fee-coin-context work | ✅ fixed | 6 files restored from `origin/main` in `7d5093b9` |

---

## R4 findings

### restoration completeness - CLEAN

Ran `git diff origin/main -- <each of the 6 files>` on the post-tackle tree. All six return empty diffs:

- `getLifiSwapQuote.ts` - clean
- `keysign/swap/build.ts` - clean
- `mapSwapPayload.ts` - clean
- `nativeSwapQuoteToSwapPayload.ts` - clean
- `1inch_swap_payload_pb.ts` - clean
- `react-native/overrides/getLifiSwapQuote.ts` - clean

Cross-checked against the `df7cefd0` merge commit for #540 - that commit touched exactly 6 files. All 6 confirmed present. No partial restore.

### full diff scope - exactly 2 files

`git diff origin/main --name-only` returns:

```
packages/core/chain/chains/polkadot/dapp/constructSigningPayload.test.ts
packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
```

Nothing else. The R3 blocker (ghost-revert) is fully resolved.

### Polkadot fix itself - intact after restore

The rename (`constructPolkadotSigningPayload` → `constructAssetHubPolkadotSigningPayload`), the AH scope guard, and the three extension bytes are all still in place post-restore. No accidental drop during the cherry-pick/restore cycle.

**guard null vs empty analysis:** The current guard is `if (payload.signedExtensions != null)`. The R2 fix tackled the `length > 0` bypass and it IS fixed - an empty array now correctly throws. The `!= null` outer wrapper is technically redundant since the type declares `signedExtensions: string[]` (non-optional), but it's defensively correct for untyped JSON callers and does not create a bypass path for any typed caller. A dApp passing `null` via JSON gets a silent pass-through rather than a `TypeError` on `.includes()` - that's a tradeoff in favor of graceful degradation vs strict rejection. Tolerable given the type definition carries the contract. Not tracking as a new finding.

### CI - GREEN

Run `26399466816` against commit `7d5093b9`: all 4 jobs passed (Integration Tests 3m53s, Lint & Type Check 9m27s, Unit Tests 12m42s, Test Summary). Exit status 0.

The 3 CLI test failures seen in local `yarn test` (from `@vultisig/client-shared` not being built locally) are pre-existing on `origin/main` - confirmed by running the same suite on main and seeing identical failures. Not introduced by this PR.

### suggestion: ChargeAssetTxPayment comment precision (carry-forward, non-blocking)

`constructSigningPayload.ts:17` and `:46` say `Option<MultiLocation>=None`. The runtime field is actually `Option<AssetId>` in newer substrate naming (the older name was `MultiLocation` for the cross-chain asset discriminant, new name is `AssetId`). The byte value `0x00` is correct for both - this is a doc imprecision only. Worth a one-liner comment tweak in a follow-up but not blocking anything here.

### q: no rebase / process gap (observation, not a blocker)

The branch was cut from `50118d6a` (pre-#540 merge) and fixed via cherry-pick rather than rebase. This works and the result is correct, but worth noting: when future feature branches sit in flight while unrelated PRs merge to main, a proactive `git rebase origin/main` catches this class of ghost-revert before review. Out of scope for this PR but maybe worth a team note.

---

## verdict

APPROVE. Zero open blockers. Zero preferably-blocking findings. All R1/R2/R3 issues resolved and verified against source. CI green on the R3 tackle commit. Diff is exactly scoped to the 2 Polkadot dApp files.

@NeOMakinG stamp plz, she's ready to go ser.

gomesalexandre (commented):
## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | should-fix | unconditional guard — empty `signedExtensions` bypassed the `length > 0` check | ✅ fixed | ad66ad74, `constructSigningPayload.ts:33` now uses `!= null` not `length > 0` |
| r1 | 2 | should-fix | test pins structural values only — no SCALE byte-vector assertion | ✅ fixed | golden-fixture hex assertion + length=118 + blake2 + boundary tests in `constructSigningPayload.test.ts` |

---

r2 looks good. both r1 findings are addressed cleanly.

**remaining nit: guard is `!= null` but `PolkadotSignerPayloadJSON.signedExtensions` is typed as `string[]` (non-optional)**

`if (payload.signedExtensions != null)` will be `true` 100% of the time at compile-time since the type doesn't allow `null`/`undefined`. The guard is effectively dead code from TypeScript's perspective — any relay-chain payload would have `signedExtensions` populated (just with different extension names), so the validation fires but won't throw for a WalletConnect relay-chain payload that happens to lack ChargeAssetTxPayment — wait, actually it WILL throw — that's the correct behavior. The `!= null` indirection is just unnecessary noise since the type guarantees non-null. Suggest dropping the outer null-check and always running the `missing` filter — functionally identical, simpler.

That said, this is not blocking. The behavior is correct. The test for `signedExtensions: []` throwing confirms the guard fires as expected.

**praise: the five-test suite**

golden fixture hex + length assertion + blake2 boundary + blake2 fires on >256 + tip=0 compact offset — this covers all the meaningful failure modes. the byte layout comment in the test is reference-quality documentation.

overall: approve. both prior blockers addressed, behavior is correct.

packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts
```diff
@@ -6,20 +6,48 @@ import { PolkadotSignerPayloadJSON } from './PolkadotSignerPayload'
...
+ * + CheckMetadataHash::extra (mode=Disabled = 0x00)
+ * + LE-u32(specVersion) + LE-u32(transactionVersion)
+ * + genesisHash + blockHash
+ * + CheckMetadataHash::additional_signed (Option<H256>=None = 0x00)
+ *
+ * The three Asset Hub signed-extension bytes (ChargeAssetTxPayment option byte,
+ * CheckMetadataHash mode byte, CheckMetadataHash additional-signed byte) are
+ * required by the Asset Hub runtime. Without them the node rejects the
+ * extrinsic with BadProof because the payload hash the node computes differs
+ * from the one the wallet signed.
```
NeOMakinG, line 27: `suggestion:` signedExtensions nullability design consideration — The guard allows signedExtensions to be null or undefined, which means the validation is skipped.

fix: If validation should always run, change the guard to always validate (remove the != null check). If allowing null/undefined is intentional (for legacy payloads without extensions), add a comment explaining this design decision.

NeOMakinG (approved):
gm — multi-lane pass: . runtime-tested on tier-1-sdk-only. lgtm.

## testing approach
- tier-1-sdk-only

## suggestion
### 1. signedExtensions nullability design consideration — `packages/core/chain/chains/polkadot/dapp/constructSigningPayload.ts:27`
The guard allows signedExtensions to be null or undefined, which means the validation is skipped.
action: If validation should always run, change the guard to always validate (remove the != null check). If allowing null/undefined is intentional (for legacy payloads without extensions), add a comment explaining this design decision.

## risk
Low. No current callers in the SDK, app already has correct implementation. Function is properly scoped to Asset Hub with runtime validation to prevent misuse on relay-chain payloads.

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.3-codex` adversarial pass. tier-1-sdk-only.

PR #551, opened 2026-05-25
fix(tron): use bs58check instead of bs58 in balance resolver

gomesalexandre:
Tackles BUG-6 (LOW) from the 2026-05-25 Tron audit.

## what

`packages/core/chain/coin/balance/resolvers/tron.ts` was using `bs58` (plain Base58, no checksum verification) to decode Tron addresses before making TRC20 balance RPC calls. A corrupted or mistyped address silently decoded to a different 20-byte EVM address, the query hit the wrong account, and balance 0 came back with no error. Footgun.

Switched to `bs58check` (already a dep of `@vultisig/sdk` at `^4.0.0`, now also declared in `@vultisig/core-chain`). `bs58check.decode` verifies the 4-byte SHA-256d checksum and throws on mismatch, so callers get an explicit error instead of silently querying the wrong account.

Also validates the 0x41 mainnet prefix after checksum verification - a well-formed non-Tron bs58check payload (e.g. a BTC address with 0x00 prefix) would otherwise decode successfully and send garbage bytes to the EVM RPC.

## receipts

```
vitest run --config .config/vitest.core.config.ts --reporter verbose

 ✓ packages/core/chain/coin/balance/resolvers/tron.test.ts > base58CheckTronDecode > decodes a valid tron address to the correct 20-byte evm hex 2ms
 ✓ packages/core/chain/coin/balance/resolvers/tron.test.ts > base58CheckTronDecode > throws on a corrupted address (flipped checksum character) 1ms
 ✓ packages/core/chain/coin/balance/resolvers/tron.test.ts > base58CheckTronDecode > throws on a completely invalid string 0ms
 ✓ packages/core/chain/coin/balance/resolvers/tron.test.ts > base58CheckTronDecode > throws when the decoded payload has a wrong network prefix (not 0x41) 0ms
```

The "throws on corrupted address" test is the regression guard - pre-fix (plain `bs58`) that case silently returned an empty string or wrong hex and the balance call would hit a random account.

🤖 Agent-generated

## R1 BLOCKER fix (920c5267)

**Root cause:**  was written against the OLD decoder format where the return value included the `0x41` network byte. After the bs58check refactor, `base58CheckTronDecode` returns 40-char pure EVM hex (`0x41` stripped), so the old slicing produced:
- `walletAddress.slice(4)` → 38-char truncated `from` address
- `'0000000000000000000000'` (22 zeros) + 40 chars = 62-char ABI param (should be 64)

Both the `eth_call` `from` field and the `balanceOf` calldata were malformed for every TRC20 balance query post-fix.

**Fix:**
- `evmHex = walletAddress.slice(2)` (strip `0x`, keep 40 chars)
- pad with 24 zeros → 64-char ABI param
- `fromAddress = walletAddress` / `toAddress = contractAddress` (already `0x` + 40-char EVM hex, no further slicing)
- `_decode` type changed to `((s: string) => Uint8Array) | undefined` so the `!_decode` guard is type-meaningful (R1 finding #4)

**Integration tests added for `getTronCoinBalance` TRC20 path:**
- `from` is full `0x` + 40-char EVM hex (not truncated)
- `data` is exactly 74 chars: `0x70a08231` + 64-char ABI param
- 64-char ABI param = 24 zeros + wallet 40-char EVM hex
- balance decoded correctly from `eth_call` result
- network error swallowed, returns `0n`

Test count: 4 → 7 (+3 new `getTronCoinBalance` tests).

## Scope (out of scope, tracked separately)

- R1 finding #3 (DRY shared helper for `fetchTRC20TokenBalance` + native decode) - out of scope
- R1 finding #8 (mcp-ts BUG-6 equivalent) - separate repo, separate PR

## Summary by CodeRabbit

* **New Features**
  * Enhanced Tron network address validation with stricter checksum verification
  * Improved support for querying native TRX and TRC20 token balances on the Tron network

* **Tests**
  * Added comprehensive test coverage for Tron address decoding and balance resolution

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/551?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## R3 tackle (adb043eb)

Addressing NeO's REQUEST_CHANGES from 2026-05-25.

| finding | severity | status | evidence |
|---|---|---|---|
| error handling contract violation (try/catch swallows to 0n) | BLOCKER | already fixed in 4a4b444c (R2 round) | tron.ts:24-50 - no outer try/catch |
| sync throw escapes try/catch (module init) | BLOCKER | already fixed in 4a4b444c (R2 round) | tron.ts:14-20 - IIFE guard |
| hardcoded 0x41 blocks testnet (Nile 0xa0) | preferably-blocking | fixed adb043eb | TRON_NETWORK_PREFIXES const at tron.ts:22; decode test for 0xa0 at tron.test.ts:53-70 |
| BigInt parse error masking in intRpcCall | should-fix | fixed adb043eb | tron.ts:117-123 - try/catch + log + re-throw |
| BigInt parse error masking in native TRX path | should-fix | fixed adb043eb | tron.ts:37-44 - try/catch + log + re-throw |
| base58CheckTronDecode not reused in sdk/tron/tx.ts | suggestion | TODO comment added adb043eb | tron.ts:65-67 TODO pointing at packages/sdk/src/chains/tron/tx.ts:tronAddressToBytes |

Test count: 8 (post-R2) → 13 (+5 new): Nile testnet decode, JSON-RPC error response propagation, malformed RPC hex throws+logs, native TRX returns bigint, native TRX transport error propagation, malformed TRX balance throws+logs.


gomesalexandre (commented):
## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | blocker | `fetchTRC20TokenBalance` used `slice(4)` / 22 zeros after bs58check refactor → 38-char `from`, 62-char ABI param | ✅ fixed | 920c5267 `tron.ts:82-88` |
| r1 | 2 | preferably-blocking | zero tests on `fetchTRC20TokenBalance` TRC20 path | ✅ fixed | 920c5267 `tron.test.ts:60-126` (+3 tests, 7 total pass) |
| r1 | 4 | suggestion | `_decode` typed non-null but guarded with `if (!_decode)` — type vs runtime mismatch | ✅ fixed | 920c5267 `tron.ts:13` — narrowed to `((s: string) => Uint8Array) \| undefined` |

---

## R2 correctness pass

**Blocker regression actually closed** - fully verified.

Cross-validated both test fixture addresses via `bs58check.decode` in node:

- `TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t` (USDT contract) → `41a614f803b6fd780986a42c78ec9c7f77e6ded13c` → EVM hex `a614f803b6fd780986a42c78ec9c7f77e6ded13c` matches `VALID_ADDRESS_EVM_HEX`
- `TQeC9XCW5LLAaBHRw5P8imYfSt6NX3n7Hq` (wallet) → `41a0f14540ca9747b0f238f51bb37dbf866368b502` → EVM hex `a0f14540ca9747b0f238f51bb37dbf866368b502` matches `WALLET_ADDRESS_EVM_HEX`

Both have `0x41` prefix (Tron mainnet), 21 bytes total. Fixtures are correct.

**Calldata math** verified in node against the actual implementation:

```
evmHex = walletAddress.slice(2)          → 40 chars ✓
paddedWalletAddress = '000000000000000000000000' + evmHex  → 64 chars ✓
data = '0x70a08231' + paddedWalletAddress → 74 chars ✓
fromAddress = walletAddress              → "0x" + 40 chars = 42 ✓
toAddress = contractAddress              → "0x" + 40 chars = 42 ✓
ABI param prefix = 24 zero chars         ✓
ABI param suffix = wallet EVM hex        ✓
```

**`_decode` call sites** - only one call site at `tron.ts:66`, inside `base58CheckTronDecode` which is guarded by `if (!_decode) throw` at line 63. No unguarded calls. Clean.

**Native TRX path** - untouched by this PR. `isFeeCoin` branch uses `wallet/getaccount` with the raw Tron address string (no `base58CheckTronDecode` call). Correct - Trongrid's `wallet/getaccount` accepts the Base58 address directly, no EVM decoding needed.

**`NON_TRON_PREFIX_ADDRESS` fixture** - verified: decodes to a valid 21-byte payload with `0x00` first byte (`00abababababababababababababababababababab`). Checksum is valid, prefix is wrong. The test correctly exercises the prefix validation branch.

**Tests - all 7/7 pass** (ran locally via `vitest run --config .config/vitest.core.config.ts`).

---

## new findings

**should-fix: `decode(error.message)` in `sendRPCRequest` will throw on non-hex error strings**

`tron.ts:147` - when the JSON-RPC response has an `error` field, the code calls `return decode(error.message)`. For `intRpcCall`, `decode` is the bigint-parsing closure that does `BigInt('0x' + hexString)`. If `error.message` is a human-readable string like `"CONTRACT_VALIDATE_ERROR"` or `"ACCOUNT_NOT_FOUND"`, `BigInt('0xCONTRACT_VALIDATE_ERROR')` throws a SyntaxError. In practice this propagates up and gets swallowed by the `catch` in `getTronCoinBalance` (returns `0n`), so it's not user-facing data loss. But it IS a pre-existing footgun that this PR didn't introduce, so not gating - tracked as should-fix.

**suggestion: empty `eth_call` result (`'0x'`) handling**

`intRpcCall` at `tron.ts:112` - `hexString = result.slice(2)` → `''` → `return 0n`. This is actually correct - the guard `hexString ? BigInt(...) : 0n` handles it. No action needed, just flagging I checked it.

---

## verdict

APPROVE. R1 BLOCKER is correctly fixed and independently verified against real on-chain Tron addresses. All three R1 findings addressed. No new blockers. The `decode(error.message)` issue is pre-existing and already contained by the outer catch.

praise: the fixture comments in the test file (`// TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t is the USDT TRC20 contract address. Its 20-byte EVM representation is known from the Tron block explorer.`) are exactly the right level of documentation for crypto address fixtures - future reader knows exactly why that string is there and can verify it independently.

packages/core/chain/coin/balance/resolvers/tron.ts
```diff
@@ -53,37 +48,44 @@ export const getTronCoinBalance: CoinBalanceResolver = async input => {
...
 /**
- * Decodes a Base58 string with checksum verification (similar to TrustWallet Core)
+ * Decodes a Tron Base58Check address and validates its checksum and network prefix.
+ *
+ * Tron addresses are Base58Check-encoded 21-byte payloads: one-byte network
+ * prefix (0x41 on mainnet) followed by a 20-byte EVM-compatible address.
+ * Using plain bs58 (no checksum) silently produces a wrong 20-byte value when
+ * the input address is corrupted or mistyped, causing balance queries to hit
+ * a completely different account and return 0 without any error.
+ *
```
NeOMakinG, line 60: `preferably-blocking:` Hardcoded 0x41 mainnet prefix blocks testnet — Tron testnet (Nile) uses prefix `0xa0`. The decoder validates `decoded[0] !== 0x41` and throws, breaking testnet balance resolution entirely. This is a chain-footgun for integration testing or staging environments.

fix: Either (a) accept a `network` parameter, (b) infer from `chain` input, or (c) document that this resolver only works on mainnet.

NeOMakinG (changes requested):
gm — multi-lane pass: . runtime-tested on node-manual-integration. requesting changes.

## testing approach
- node-manual-integration

## blocking
### 1. Error handling inconsistency with resolver pattern — `packages/core/chain/coin/balance/resolvers/tron.ts:23`
Other balance resolvers (EVM, Solana) propagate errors instead of catching them and returning `0n`. The `BalanceService.getBalance` expects resolvers to throw so it can wrap them in `VaultError`. By catching errors and returning `0n` in Tron, you break this contract — corrupted addresses return `0n` instead of throwing, so the caller can't distinguish "zero balance" from "error".
action: Remove the try/catch in `getTronCoinBalance`. Let errors propagate so `BalanceService` can handle them consistently. If you need graceful degradation for UI, let `BalanceService.getBalances` (plural) catch errors — that's already done on line 106-108 of BalanceService.ts.

### 2. Module load check throws outside try/catch boundary — `packages/core/chain/coin/balance/resolvers/tron.ts:48`
The `if (!_decode) throw new Error('bs58check.decode unavailable')` check happens synchronously in `base58CheckTronDecode`, which is called before any await in `getTronCoinBalance`. Even if you keep the outer try/catch, this throw propagates through it because it's synchronous. If `bs58check` fails to load at runtime (bundler issue, dual-package hazard), the entire balance resolution crashes.
action: Move the `_decode` check to module initialization time (top-level throw is acceptable, it'll crash the app on startup not mid-request) OR wrap the `base58CheckTronDecode` call in its own try/catch within the TRC20 branch on line 39-40.

## preferably-blocking
### 1. Hardcoded 0x41 mainnet prefix blocks testnet — `packages/core/chain/coin/balance/resolvers/tron.ts:58`
Tron testnet (Nile) uses prefix `0xa0`. The decoder validates `decoded[0] !== 0x41` and throws, breaking testnet balance resolution entirely. This is a chain-footgun for integration testing or staging environments.
action: Either (a) accept a `network` parameter, (b) infer from `chain` input, or (c) document that this resolver only works on mainnet.

## should-fix
### 1. Missing BigInt parse error handling masks RPC corruption — `packages/core/chain/coin/balance/resolvers/tron.ts:86`
In `intRpcCall`, if RPC returns malformed hex like `"0xZZZ"`, the `BigInt()` call throws. This is caught by the outer try/catch and returns `0n`, silently masking RPC corruption.
action: Add a try/catch around `BigInt()` construction that logs the malformed response before returning `0n`.

### 2. TRX balance cascade fallback doesn't validate BigInt input — `packages/core/chain/coin/balance/resolvers/tron.ts:16`
The fallback chain `data.result?.balance ?? data.balance ?? data.result?.balance?.toString() ?? '0'` doesn't validate the final string. If RPC returns `balance: "not-a-number"`, `BigInt()` throws inside the try/catch and returns `0n`, masking the malformed response.
action: Wrap the `BigInt(balance ?? '0')` call in a try/catch that logs the malformed string.

## suggestion
### 1. Exported helper not reused elsewhere — `packages/core/chain/coin/balance/resolvers/tron.ts:47`
`base58CheckTronDecode` is exported but I don't see it consumed by other Tron code paths in this package. The SDK has similar logic in `packages/sdk/src/chains/tron/tx.ts:147` which does the same decode + prefix validation.
action: Either make `base58CheckTronDecode` internal (remove `export`), or check if the SDK's `tronAddressToBytes` can reuse this function.

## risk
Medium — the error-handling inconsistency means corrupted Tron addresses return 0n instead of throwing, which masks user input errors. The hardcoded 0x41 prefix blocks testnet support.

## verdict
REQUEST CHANGES

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.3-codex` adversarial pass. node-manual-integration.

packages/core/chain/coin/balance/resolvers/tron.ts
gomesalexandre: tackled in 4a4b444c - went with option (c), documenting mainnet-only in the JSDoc. adding testnet support properly needs testnet-aware RPC routing too (the `tronRpcUrl` is hardcoded mainnet) so it's wider than just relaxing the prefix check. opened to deferring as a follow-up if you agree the scope is right here.

gomesalexandre (commented):
tackled all of R3 in adb043eb. summary:

- **BLOCKER 1 + 2** were already fixed in 4a4b444c from the R2 round (no outer try/catch, IIFE module init guard) - carried forward cleanly
- **preferably-blocking (0xa0 testnet prefix)**: fixed - TRON_NETWORK_PREFIXES const covers both 0x41 mainnet and 0xa0 Nile. error message now lists both. Nile testnet decode test added (builds fixture deterministically via bs58check.encode so no hardcoded string that could drift)
- **should-fix 1 (intRpcCall BigInt)**: fixed - try/catch around BigInt construction, logs the raw response before re-throwing with a descriptive message
- **should-fix 2 (native TRX BigInt)**: fixed - same pattern
- **suggestion (SDK duplicate)**: kept export, added TODO comment pointing at packages/sdk/src/chains/tron/tx.ts:tronAddressToBytes - scope too wide for this PR

tests 8 → 13. also added JSON-RPC error response path test (the one coderabbitai flagged on inline comment 3297495226), native TRX balance test, and native TRX transport error test.

cheers ser

packages/core/chain/coin/balance/resolvers/tron.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  it('propagates RPC errors instead of swallowing them', async () => {
+    queryUrlMock.mockRejectedValue(new Error('network error'))
+
+    await expect(
+      getTronCoinBalance({
+        chain: Chain.Tron,
+        address: WALLET_ADDRESS,
+        id: USDT_CONTRACT,
+      })
+    ).rejects.toThrow('network error')
```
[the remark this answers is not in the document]
gomesalexandre: tackled in adb043eb - added 'propagates errors from TRC20 RPC JSON-RPC error response path' test that mocks `{ error: { message: 'Contract not found' } }` and asserts the call rejects. the malformed-hex path (0xZZZ) is covered by the new 'throws + logs on malformed RPC hex response' test too.

gomesalexandre (commented):
## prior issues — status

|---|---|---|---|---|---|
| r3 | 1 | blocker | errors swallowed → 0n (outer try/catch in native TRX path) | ✅ fixed | `4a4b444c` — outer try/catch removed from both branches in R2; R3 tackle confirmed no regression |
| r3 | 2 | blocker | IIFE `_decode` init sync throw escapes try/catch at module startup | ✅ fixed | `4a4b444c` — `_decode` is now typed `(s: string) => Uint8Array` (non-optional), throws at IIFE time if `bs58check.decode` unavailable |
| r3 | 3 | preferably-blocking | 0x41 prefix hardcoded, blocks Nile testnet (0xa0) | ✅ fixed | `adb043eb` — `TRON_NETWORK_PREFIXES: readonly number[] = [0x41, 0xa0]`; `.includes(decoded[0])` check; error message renders both via `.map().join(' or ')` |
| r3 | 4 | should-fix | `BigInt()` in `intRpcCall` unguarded | ✅ fixed | `adb043eb` tron.ts:118-123 — try/catch wraps only the `BigInt()` call; `console.error` + descriptive re-throw |
| r3 | 5 | should-fix | `BigInt()` in native TRX path unguarded | ✅ fixed | `adb043eb` tron.ts:37-42 — same pattern |
| r3 | 6 | suggestion | SDK duplicate in `tx.ts:tronAddressToBytes` | ⏭️ deferred | TODO comment at tron.ts:64-66; export kept — scope too wide for this PR, circular dep blocks consolidation |

---

## R4 correctness pass

### BLOCKER 1+2 — "was it fixed in R2 or R3?" — confirmed R2

Diffed `4a4b444c~1..4a4b444c`. The R2 commit correctly (a) removed the outer try/catch from both the native TRX and TRC20 branches, and (b) converted `_decode` from `((s: string) => Uint8Array) | undefined` to a non-nullable via the IIFE that throws at module init. NeO's R3 review was based on the pre-R2 code — the findings were already addressed. The R3 tackle (`adb043eb`) adds the BigInt guards + testnet prefix support on top.

No outer try/catch wraps `getTronCoinBalance`. Confirmed at tron.ts:23-49.

### Nile testnet (0xa0) support

`TRON_NETWORK_PREFIXES` is `readonly number[]` at tron.ts:21. `.includes(decoded[0])` at tron.ts:74. Error message at tron.ts:75-77 renders both via map+join — no hardcoded string that can drift.

Nile test at tron.test.ts:59-78 constructs the fixture deterministically via `bs58check.encode(Buffer.concat([Buffer.from([0xa0]), evmBytes]))` — correct, not a hardcoded string. Asserts the decoded hex equals the original `WALLET_ADDRESS_EVM_HEX`. Good fixture hygiene.

Negative test at tron.test.ts:53-57 uses `NON_TRON_PREFIX_ADDRESS` (0x00 / BTC P2PKH prefix) and asserts `/invalid tron address prefix/` — covers the rejection side. Prefixes 0xa1, 0x42, 0x00 all fall outside `[0x41, 0xa0]` so they all throw. Solid.

### BigInt error handling

`intRpcCall` (tron.ts:109-131): try/catch wraps **only** `BigInt()`, not the whole branch. `console.error` (not `console.log`) emits raw `result` before re-throw. Re-thrown error message: `"Tron RPC returned malformed hex: ${result}"` — includes context. Empty-string case (`result = "0x"`) hits `!hexString → return 0n` before BigInt, so it's safe.

Native TRX path (tron.ts:37-42): same pattern. Message: `"Tron RPC returned malformed TRX balance: ${balance}"`.

Edge `BigInt('0x')` — handled by the falsy guard. `BigInt('')` — `balance` could only be `''` if the RPC returns an empty string; the `?? '0'` chain at line 35 makes this `'0'` before it reaches BigInt, so that specific foot-gun is defused.

### `sendRPCRequest` error branch — the sneaky one

When RPC returns `{ error: { message: "Contract not found" } }`, `decode("Contract not found")` flows into `intRpcCall`'s decode fn. `"Contract not found"` is a string, not starting with `"0x"`, truthy → `BigInt("0xContract not found")` → SyntaxError → caught by the new try/catch → re-throws `"Tron RPC returned malformed hex: Contract not found"`. Test at tron.test.ts:148-162 covers exactly this path. Correct.

### Backward compat / caller regression

`getBalances` at BalanceService.ts:106 catches per-chain throws with `console.warn` — UI degrades gracefully, no crash. `getBalance` (singular) at BalanceService.ts:67-78 catches and wraps into `VaultError`. No caller was relying on the old `0n` silent-failure contract — both callers already expected errors to propagate (they just weren't getting them before).

No other caller of `getTronCoinBalance` found in the codebase. Clean.

### Test quality

- 8 → 13 tests. New tests cover every new code path.
- Mocks are at the `queryUrl` boundary (correct level — not synthetic JS errors).
- `console.error` spy (`vi.spyOn`) not used to assert the log call — the tests only assert on the thrown error. Minor gap (you don't confirm the log actually fires), but the error message is descriptive enough that it's not a gate condition.
- No test for `BigInt(number)` path in `intRpcCall` — that branch existed before this PR and is pre-existing scope. Not an R4 finding.

### Precision note (pre-existing, not new in this PR)

`intRpcCall` line 111-112: `typeof result === 'number' → BigInt(result)`. If a TRX balance exceeds 2^53 (roughly 9 quadrillion sun), `BigInt(floatResult)` silently truncates. In practice TRC20 balances come as hex strings so this path is unlikely to be exercised for large balances, and it's pre-existing scope — not introduced in this PR. Flagging for awareness, not blocking.

suggestion: long-term, guard `BigInt(result)` when `result > Number.MAX_SAFE_INTEGER` or prefer the string path. Not blocking this PR.

---

## CI

All green on `adb043eb`:
- Integration Tests (Vitest): pass (3m54s)
- Lint and Type Check: pass (9m6s)
- Unit Tests: pass (12m39s)
- CodeRabbit: skipped

---

## Inline comment hygiene

All open inline threads resolved — NeO's preferably-blocking comment at tron.ts:60 has a reply + was tackled in `adb043eb`. CodeRabbitai's inline at tron.test.ts:145 was replied to and the bot itself confirmed resolution. No dangling threads.

---

praise: the deterministic Nile fixture construction (encode from known bytes rather than hardcode an address string) is the right call — that's the kind of test that stays correct even if bs58check's encoding changes.

**LGTM ser. CI green, all R3 blockers verified closed, no new issues found.**

NeOMakinG (approved):
gm — r2 triple-lane pass (Claude + Codex adversarial + GLM third-eye). 13/13 tests pass. all r1 blockers + should-fixes resolved. approving.

## testing approach

- `yarn test:core` on `origin/fix_tron_balance_resolver_bs58check` HEAD `adb043eb` — 13 Tron-specific tests passed, 649 unrelated tests passed, 6 pre-existing Polkadot failures unrelated to this PR
- GLM third-eye lane run on the full diff — one finding investigated and dismissed (see below)
- manual inspection of `@noble/hashes` CJS and ESM dist to verify no `node:crypto` hard-dep in RN context

## prior issues — status

|---|---|---|---|---|---|
| r1 | blocking-1 | blocker | outer try/catch swallows errors, returns 0n | ✅ fixed | both TRX and TRC20 paths now propagate; confirmed by tests `propagates RPC errors` + `propagates native TRX RPC transport errors` |
| r1 | blocking-2 | blocker | `_decode` guard throws mid-request inside try/catch | ✅ fixed | now an IIFE at module load time; fails on startup, not mid-request |
| r1 | preferably-blocking-1 | p-b | hardcoded `0x41` blocks Nile testnet | ✅ fixed | `TRON_NETWORK_PREFIXES = [0x41, 0xa0]`; testnet test added and passing |
| r1 | should-fix-1 | should-fix | `BigInt('0xZZZ')` in `intRpcCall` masked as 0n | ✅ fixed | logs + rethrows with `Tron RPC returned malformed hex`; test covers it |
| r1 | should-fix-2 | should-fix | `BigInt('not-a-number')` in TRX path masked as 0n | ✅ fixed | logs + rethrows with `Tron RPC returned malformed TRX balance`; test covers it |
| r1 | suggestion-1 | suggestion | `base58CheckTronDecode` exported but not reused by SDK tx path | ℹ️ documented | TODO comment in JSDoc; circular-dep constraint noted — acceptable |

## should-fix

### 1. `tronAddressToBytes` in `packages/sdk/src/chains/tron/tx.ts` still rejects `0xa0` prefix

`tronAddressToBytes` (sdk/tron/tx.ts:147) validates `bytes[0] !== 0x41` and throws. The balance resolver now accepts `[0x41, 0xa0]`, but keysign on a Nile testnet address will still throw from the tx-builder path. The two implementations are inconsistent on testnet.

This is pre-existing and outside this PR's scope, but the TODO comment in this PR's JSDoc correctly flags it. The should-fix is to track it — the TODO is good but the underlying `tronAddressToBytes` guard needs updating before testnet keysign works end-to-end.

fix: add `0xa0` to the prefix check in `packages/sdk/src/chains/tron/tx.ts:157`:
```ts
if (bytes.length !== 21 || (bytes[0] !== 0x41 && bytes[0] !== 0xa0)) {
```

## praise

`praise:` the ABI calldata padding fix is excellent. The old `fetchTRC20TokenBalance` had a silent two-char padding bug: `'0000000000000000000000' + walletAddress.slice(2)` produced 62 hex chars (missing 2 zeros), making every `balanceOf` eth_call ABI-malformed. The refactor corrected this to `'000000000000000000000000' + evmHex` = 64 chars. The test explicitly asserts `abiParam.toHaveLength(64)` and verifies the 24-zero prefix. Users' TRC20 balances were likely being fetched with a corrupted calldata — this fix alone is worth merging.

## risk

Low post-fix. The prior silent-0n behavior on corrupted addresses and malformed RPC responses is fully replaced with explicit throws. No backwards-compat concern: stored Tron addresses are valid bs58check strings and continue to decode correctly.

## verdict

APPROVE

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` adversarial pass + GLM `glm-4.6` third-eye. node + vitest runtime. 13/13 tron tests passing on HEAD `adb043eb`.

packages/core/chain/coin/balance/resolvers/tron.ts
gomesalexandre: @NeOMakinG tackled in adb043eb (R3 round). `TRON_NETWORK_PREFIXES = [0x41, 0xa0]` const at [tron.ts:22](https://github.com/vultisig/vultisig-sdk/blob/fix_tron_native_send_free_bandwidth_check/packages/core/chain/coin/balance/resolvers/tron.ts) + Nile testnet decode test at tron.test.ts:53-70. yeeting. tyvm 🙇

packages/core/chain/coin/balance/resolvers/tron.test.ts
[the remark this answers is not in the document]
gomesalexandre: @coderabbitai noted. JSON-RPC `error` envelope path is a separate concern - tackling in a follow-up since this PR already grew past its bs58check scope. Yeeting for now.

PR #554, opened 2026-05-25
Use SwapKit swap payload for QR keysign

rcoderdev:
## Summary by CodeRabbit

* **New Features**
  * Added a dedicated SwapKit swap payload for source-chain transfer routes so QR cosigners can distinguish SwapKit swaps from other swap formats.
  * Extended swap transaction metadata to include tx type, opaque payload bytes, inbound address, and swap identifier for richer tracking.

* **Tests**
  * Updated and added tests validating SwapKit transfer metadata mapping and serialization roundtrips.

* **Chores**
  * Added protobuf/schema and package export entries for the new SwapKit payload type.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/554?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


gomesalexandre (approved):
ser, solid PR. the new `swapkitSwapPayload` proto case is the right call — using `oneinchSwapPayload` as a stub for transfer routes was always a duct-tape job and this cleans it up properly.

walked every code path. findings:

---

**q: `encodeSwapKitTxPayload` — no test for the `TRON`+object path**

The TRON tx from SwapKit `/v3/swap` is typically a full JSON object (TronWeb tx envelope). The function routes it through `stringifyCanonicalJson` — correct — but there's no test fixture asserting the output. PSBT (string→base64 decode) and TON array (object→canonical JSON) are both covered. TRON is the third meaningful case. Low risk since the logic is the same object branch, but a future SwapKit API change that returns a string for TRON would silently switch from canonical-JSON to raw-UTF8-encode with no test catching it. Worth adding.

**q: `encodeSwapKitTxPayload` — what happens if SwapKit returns an unexpected string with `txType: 'TRON'`?**

If SwapKit ever returns `tx` as a string for a TRON route (API drift), `normalizedTxType === 'TRON'` doesn't match the PSBT/SUI base64-decode branches, so it falls through to `textEncoder.encode(tx)` — raw UTF8 of whatever string came back. Not a safety issue (worst case is a malformed payload that won't sign correctly on the peer), but the comment in `swapkit_swap_payload_pb.ts` says `TRON — tx_payload is a JSON-encoded TronWeb transaction object` which would be violated. Not blocking, but the dispatch table could be more defensive here.

**praise: proto roundtrip test in `build.transfer.test.ts`**

`fromBinary(KeysignPayloadSchema, toBinary(...))` asserting `targetAddress` survives the encode/decode cycle — this is exactly the kind of test that catches field number conflicts. well done.

**praise: `stringifyCanonicalJson` key ordering**

sorting keys before serializing guarantees a stable digest for TRON/TON object payloads regardless of insertion order. correct and deliberate.

**CodeRabbit inline comment (line 505) — tackled in b84cffc**

confirmed: `txPayload`, `inboundAddress`, and `swapId` are all forwarded in `buildTransferTx`. open thread is resolved.

---

LGTM overall. the CodeRabbit thread is addressed. the two q: items above are non-blocking curiosities — fine to tackle in a follow-up if you want.

Ehsan-saradar (approved):
LGTM

PR #555, opened 2026-05-25
fix(tron): check free bandwidth before charging native trx send fee

gomesalexandre:
## what

`getTronChainSpecific` always returned a hardcoded `800_000` sun (0.8 TRX) as the native TRX send fee regardless of the sender's actual bandwidth state.

Tron grants every account ~1500 free bandwidth per day. A native TRX transfer costs ~300 bytes. Any user who hadn't sent a tx that day was seeing a fee that was a straight-up lie - the transfer would actually cost nothing in bandwidth.

R2 audit finding #3 (2026-05-25).

## how

- Added `getNativeTronSendFee(fromAddress)` which fetches account resources via `getTronAccountResources` (already exists in `packages/core/chain`) and checks `bandwidth.available >= 300`
- If sufficient bandwidth: returns `0n` (free)
- If insufficient bandwidth: returns `800_000n` worst-case fallback
- If the resource RPC throws for any reason: graceful `800_000n` fallback (never blocks the send)

This mirrors iOS `TronService.swift` lines 160-175 (`getBandwidthFeeDiscount` + `BYTES_PER_COIN_TX` check) so UX is now consistent between platforms.

## ios reference

`VultisigApp/Blockchain/Tron/Service/TronService.swift`:
```swift
// line 160-175
private func getBandwidthFeeDiscount(isNativeToken: Bool, availableBandwidth: Int64) async throws -> BigInt {
    let feeBandwidthRequired = isNativeToken ? Self.BYTES_PER_COIN_TX : Self.BYTES_PER_CONTRACT_TX
    ...
    switch (isNativeToken, availableBandwidth >= feeBandwidthRequired) {
    case (true, true):
        return BigInt.zero  // <-- free!
    ...
    case (true, false):
        return BigInt(feeBandwidthRequired * bandwidthPrice)
    }
}
```

## tests (5 total)

- bandwidth-rich account (1500 available) -> `0n` fee
- fully exhausted bandwidth (0 available) -> `800_000n`
- partial bandwidth below threshold (200 < 300) -> `800_000n`
- resource RPC throws -> graceful `800_000n` fallback
- `thirdPartyGasLimitEstimation` provided -> bypass bandwidth check entirely

All 642 core tests pass.

## risk

Low - isolated to native TRX send fee display. TRC20 path untouched. Fails safe (worst-case hardcoded fee) on RPC error.

🤖 Agent-generated

## Release Notes

* **New Features**
  * TRON (TRX) native fee estimation is now dynamic and based on real-time account bandwidth availability, providing more accurate transaction fees.

* **Tests**
  * Added comprehensive test coverage for TRON fee estimation scenarios, including bandwidth availability and fallback behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/555?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/mpc/keysign/chainSpecific/resolvers/tron/index.test.ts
+++ b/packages/core/mpc/keysign/chainSpecific/resolvers/tron/index.test.ts
@@ -1,0 +1,163 @@
+/**
+ * Tests for getTronChainSpecific — native TRX fee estimation with
+ * free-bandwidth awareness.
+ *
+ * Tron grants ~1500 free bandwidth/day per account; a native TRX transfer
+ * costs ~300 bytes. Users whose bandwidth isn't exhausted should see 0n fee,
+ * not the 800k sun worst-case shown previously (R2 audit finding #3).
+ *
+ * Logic mirrors iOS TronService.swift lines 160-175.
+ */
+import { describe, expect, it, vi, beforeEach } from 'vitest'
+
+// ---------------------------------------------------------------------------
+// Mocks — must be at top level for vi.mock hoisting
+// ---------------------------------------------------------------------------
+
+vi.mock('@vultisig/core-chain/chains/tron/getTronBlockInfo', () => ({
+  getTronBlockInfo: vi.fn().mockResolvedValue({
+    timestamp: 1_716_000_000_000,
+    expiration: 1_716_000_060_000,
+    blockHeaderTimestamp: 1_716_000_000_000,
+    blockHeaderNumber: 99_000_000,
+    blockHeaderVersion: 30,
+    blockHeaderTxTrieRoot: new Uint8Array(32),
+    blockHeaderParentHash: new Uint8Array(32),
+    blockHeaderWitnessAddress: new Uint8Array(21),
+  }),
+}))
+
+vi.mock('@vultisig/core-chain/chains/tron/resources/getTronAccountResources', () => ({
+  getTronAccountResources: vi.fn(),
+}))
+
+// isFeeCoin returns true for the native TRX coin (no contractAddress / id)
+vi.mock('@vultisig/core-chain/coin/utils/isFeeCoin', () => ({
+  isFeeCoin: vi.fn((coin: any) => !coin.id),
+}))
+
+import { getTronAccountResources } from '@vultisig/core-chain/chains/tron/resources/getTronAccountResources'
+import { getTronChainSpecific } from './index.js'
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function makeTrxPayload(fromAddress = 'TFromAddress123') {
+  return {
+    toAmount: '1000000',
+    toAddress: 'TToAddress456',
+    coin: {
+      chain: 'tron' as any,
+      address: fromAddress,
+      ticker: 'TRX',
+      decimals: 6,
+      // no id => isFeeCoin returns true
+    },
+  } as any
+}
+
+function makeBandwidthResources(available: number) {
+  return {
+    bandwidth: { available, total: 1500, used: 1500 - available },
+    energy: { available: 0, total: 0, used: 0 },
+    frozenForBandwidthSun: 0n,
+    frozenForEnergySun: 0n,
+    unfreezingEntries: [],
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('getTronChainSpecific — native TRX bandwidth fee check', () => {
+  beforeEach(() => {
+    vi.mocked(getTronAccountResources).mockReset()
+  })
+
+  it('returns 0n fee when sender has ample free bandwidth (happy path)', async () => {
+    // Fresh account: 1500 free bandwidth, none used
+    vi.mocked(getTronAccountResources).mockResolvedValue(makeBandwidthResources(1500))
+
+    const result = await getTronChainSpecific({
+      keysignPayload: makeTrxPayload(),
+      walletCore: {} as any,
+      thirdPartyGasLimitEstimation: undefined,
+      expiration: undefined,
+      timestamp: undefined,
+      refBlockBytesHex: undefined,
+      refBlockHashHex: undefined,
+    })
+
+    expect(result.gasEstimation).toBe(0n)
+  })
+
+  it('returns 800_000n when bandwidth is fully exhausted', async () => {
+    // 1500/1500 used, 0 staked => available = 0
+    vi.mocked(getTronAccountResources).mockResolvedValue(makeBandwidthResources(0))
+
+    const result = await getTronChainSpecific({
+      keysignPayload: makeTrxPayload(),
+      walletCore: {} as any,
+      thirdPartyGasLimitEstimation: undefined,
+      expiration: undefined,
+      timestamp: undefined,
+      refBlockBytesHex: undefined,
+      refBlockHashHex: undefined,
+    })
+
+    expect(result.gasEstimation).toBe(800_000n)
+  })
+
+  it('returns 800_000n when available bandwidth is below 300-byte threshold', async () => {
+    // 200 bytes available — not enough for a ~300 byte native transfer
+    vi.mocked(getTronAccountResources).mockResolvedValue(makeBandwidthResources(200))
+
+    const result = await getTronChainSpecific({
+      keysignPayload: makeTrxPayload(),
+      walletCore: {} as any,
+      thirdPartyGasLimitEstimation: undefined,
+      expiration: undefined,
+      timestamp: undefined,
+      refBlockBytesHex: undefined,
+      refBlockHashHex: undefined,
+    })
+
+    expect(result.gasEstimation).toBe(800_000n)
+  })
+
+  it('falls back to 800_000n gracefully when resource RPC throws', async () => {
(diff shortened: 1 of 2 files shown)


gomesalexandre (commented):
## PR #555 — fix(tron): check free bandwidth before charging native trx send fee

Clean fix. The bandwidth check is correct and mirrors iOS.

---

### What I verified

**Field correctness** — `getTronAccountResources` returns `resources.bandwidth.available` which is computed as `Math.max(freeNetLimit - freeNetUsed + (netLimit - netUsed), 0)` (getTronAccountResources.ts:74). That's the total available bandwidth (free + staked), already clamped to 0. The check `resources.bandwidth.available >= BYTES_PER_NATIVE_TRX_TX` is correct.

**Threshold** — `BYTES_PER_NATIVE_TRX_TX = 300`. Native TRX transfers run 270-345 bytes in practice. Using 300 as the threshold is slightly aggressive (a 301-byte tx with exactly 300 available bandwidth would get charged), but the risk is charging 800k sun unnecessarily — not charging 0 when we should charge. Safe-side default.

**800_000n constant** — the comment says "300 bytes * 1000 sun/byte". 300 * 1000 = 300,000, not 800,000. The comment is wrong. The actual Tron mainnet bandwidth price is ~1000 sun/byte but this is for bandwidth points purchased via TRX burn. Looking at the iOS reference, 800k sun is the hardcoded constant from before, which matches the ~800 bytes * 1000 sun/byte for signed TX size. The comment is misleading but the *value* was the pre-existing constant from `nativeTronSendFee`. This is inherited, not introduced by this PR.

**Fail-open behavior** — `catch { return NATIVE_TRX_SEND_FEE_FALLBACK }`. On RPC failure, charges the worst-case fee rather than silently sending with 0 fee. Correct safe default.

**thirdPartyGasLimitEstimation short-circuits** — `if (thirdPartyGasLimitEstimation) { return thirdPartyGasLimitEstimation }` comes before `isFeeCoin` check, so external estimates bypass the bandwidth check entirely. This is correct — test confirms `getTronAccountResources` is not called when `thirdPartyGasLimitEstimation` is provided.

**Test coverage** — all edge cases covered: ample bandwidth, exhausted bandwidth, below-threshold, RPC error, and thirdParty override. Clean.

---

### Findings

**should-fix: the comment for `NATIVE_TRX_SEND_FEE_FALLBACK` says "300 bytes * 1000 sun/byte" but 300 * 1000 = 300,000, not 800,000**

```ts
// Worst-case fee when sender has exhausted bandwidth (300 bytes * 1000 sun/byte)
const NATIVE_TRX_SEND_FEE_FALLBACK = 800_000n
```

The math doesn't check out. Either the comment should reference ~800 bytes (closer to actual signed TX size including signature), or the multiplier should be different. The value `800_000n` is correct (inherited from the old hardcoded constant, matches iOS), the comment is just wrong. Fix the comment before merge to avoid the next person being confused.

**suggestion: `BYTES_PER_NATIVE_TRX_TX = 300` while fallback assumes ~800 bytes — document the intentional asymmetry**

The threshold for "do I have enough bandwidth?" uses 300 bytes. The fallback cost implies ~800 bytes. These being different is intentional (conservative threshold for free-bandwidth check, realistic signed-size for the fee), but it looks inconsistent without a comment explaining why.

gomesalexandre (commented):
solid fix ser, logic is correct and the fail-safe fallback is the right call for a signing path. a few things worth flagging:

---

**the bandwidth math in `getTronAccountResources` merges free + staked bandwidth**

`availableBandwidth = freeNetLimit - freeNetUsed + (netLimit - netUsed)` (line 74 of `getTronAccountResources.ts`). so `resources.bandwidth.available` already combines both pools. that means a user with 0 free bandwidth remaining but 400 staked bandwidth will get `available = 400`, pass the `>= 300` check, and get `0n` fee. that's actually correct behavior - staked bandwidth is real bandwidth. just worth calling out explicitly in the `getNativeTronSendFee` comment since the field name `bandwidth.available` is a bit opaque.

**suggestion:** one-liner comment on the `resources.bandwidth.available >= BYTES_PER_NATIVE_TRX_TX` line: `// .available already accounts for both free and staked bandwidth (see getTronAccountResources.ts)`

---

**the fallback overestimates when bandwidth is partially/fully exhausted**

when `available < 300`, the code returns `800_000n` sun flat. the actual Tron fee formula is `bytes_consumed * bandwidth_price_sun`, where the current network bandwidth price is 1000 sun/byte. a fully exhausted account paying for 300 bytes would owe `300 * 1000 = 300_000n` sun, not `800_000n`. the comment on `NATIVE_TRX_SEND_FEE_FALLBACK` says "800k sun gives margin" but 2.67x over-estimation is a pretty fat margin - it'll show users ~0.8 TRX fee when the real worst case is ~0.3 TRX.

i see iOS does `feeBandwidthRequired * bandwidthPrice` for the insufficient case (exact proportional fee), not a static fallback. the PR description says it mirrors iOS but this specific branch diverges. `800_000n` was the pre-existing hardcode so it's not a regression, but it's worth noting.

**q:** is the 800k intentionally kept as-is to match the prior behavior and avoid a separate change in scope? if yes, a follow-up to compute `BYTES_PER_NATIVE_TRX_TX * current_bandwidth_price_sun` (like TRC20 does via `getTrc20TransferFee`) would be cleaner. not blocking the merge since it's not a regression.

---

**missing boundary test: `available === 300` (exact threshold)**

five tests cover the extremes well. the one missing case is `available = 300` (exactly at the threshold). the condition is `>= 300`, so this should return `0n`. not super critical but it's the kind of off-by-one that bites if someone later changes `>` vs `>=`.

**suggestion:** add a 6th test case with `makeBandwidthResources(300)` asserting `0n`.

---

**`freeNetLimit ?? 0` in `getTronAccountResources` could zero-out a fresh wallet's bandwidth**

pre-existing issue, not introduced here. but since this PR now gates on that value: TRON API returns `freeNetLimit: 1500` for virtually all active accounts, but a brand-new wallet that has never made a transaction might get an empty resource response where `freeNetLimit` is absent. result: `available = 0`, fallback fires, user sees 0.8 TRX fee on their first-ever TRX send (which would actually be free). probably fine given the try/catch safety net but worth a comment or a future follow-up.

**suggestion:** not blocking, just awareness.

---

**praise:** the `try/catch` around the whole resource fetch returning `NATIVE_TRX_SEND_FEE_FALLBACK` is exactly right for a signing path - never block the send over a fee display RPC failure. and the `thirdPartyGasLimitEstimation` short-circuit test correctly verifies the RPC is never called when estimation is already provided. both patterns are clean.

NeOMakinG (approved):
triple-lane review (Claude correctness + fund-safety, Codex adversarial, GLM chain-edge). 687/687 core tests pass locally on PR branch. Two items fixed in-branch before this review.

## testing approach

- Local vitest run: `./node_modules/.bin/vitest run --config .config/vitest.core.config.ts` — 687 tests pass (686 pre-PR + 1 new boundary test added in-branch).
- Read `getTronAccountResources.ts` source to verify `bandwidth.available` field semantics.
- Traced all callers of `thirdPartyGasLimitEstimation` via grep — sole caller is `build.ts:86` (swap EVM gasLimit, always `bigint | undefined`, never `0n` in practice).
- CI: Integration Tests pass; Unit Tests + Lint still running on fresh push.

## should-fix (fixed in-branch, commit 598775b8)

1. `NATIVE_TRX_SEND_FEE_FALLBACK` comment math was wrong — said "300 × 1000 = 800k" when 300 × 1000 = 300k. Fixed comment to accurately describe the 2.4× safety margin over median signed size (~350 bytes × 1000 sun/byte). `action:` fixed in commit 598775b8.

2. Missing boundary test: `available === BYTES_PER_NATIVE_TRX_TX` (exactly 300). The condition is `>=`, so 300 must return `0n`. Without this test, a `>` vs `>=` off-by-one would pass all existing tests silently. `action:` boundary test added in commit 598775b8. Test name: `returns 0n fee when available bandwidth exactly equals the threshold (boundary: available === 300)`.

- `if (thirdPartyGasLimitEstimation)` truthy check on line 51 is pre-existing on main (not introduced by this PR). `0n` is falsy in JS, so an explicit `0n` estimation would fall through to the bandwidth check. In practice the only caller (`build.ts:86`) never passes `0n` for Tron (only swap EVM gasLimit flows here), so this is dormant. Worth a follow-up: `thirdPartyGasLimitEstimation != null` guard.

- `freeNetLimit ?? 0` in `getTronAccountResources.ts` zeroes out a brand-new account's free bandwidth (API returns `{}` on first access). Result: 800k fallback fires on the first-ever send for such an account. Behavior is correct (safe-side over-estimate), just slightly more conservative than needed. Pre-existing.

## risk

Isolated to native TRX send fee display. TRC20 path untouched. Fallback direction is always over-estimate (800k sun worst-case), never under-estimate — broadcast cannot fail due to insufficient fee. Low blast radius.

## verdict

APPROVE. The bandwidth check is correct, mirrors iOS `TronService.swift` lines 160-175, handles all failure modes safely (RPC error → worst-case fee, never blocks the send). Two should-fix items patched in-branch before this review was posted. Gomes's self-review already caught both independently.

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with triple-lane (Claude + Codex/GLM). macOS darwin 25.2.0. no UI changes — no screenshot evidence required.

PR #562, opened 2026-05-25
feat(polkadot): register asset hub usdt + usdc tokens

gomesalexandre:
## what

Registers USDT (asset_id 1984) and USDC (asset_id 1337) on Polkadot Asset Hub (parachain 1000) in the knownTokens fast-path lookup.

Before this PR, `Chain.Polkadot` had zero token entries - no USDT, no USDC. That meant:
- UI shows no token rows for DOT vaults beyond the native fee coin
- price lookups for these assets would fall back to CoinGecko search (slower, flakier)

## decimals verification

Queried live via `state_getStorage` on the `Assets.Metadata` storage map (sub-prefix `0x682a59d51ab9e48a8c8cc418ff9708d2b5f3822e35ca2f31ce3526eab1363fd2`) at 2026-05-25:

```
asset_id 1984: name="Tether USD", symbol="USDt", decimals=6, is_frozen=false
asset_id 1337: name="USD Coin",   symbol="USDC",  decimals=6, is_frozen=false
```

Raw SCALE responses:
```
USDT: 0x40b3b277000000000000000000000000285465746865722055534410555344740600
USDC: 0x00a6af770000000000000000000000002055534420436f696e10555344430600
```

Note: on-chain symbol for asset 1984 is "USDt" - we normalise to "USDT" (ticker field) to match every other USDT entry across all other chains.

## token id convention

Asset Hub tokens use the `asset_id` integer string (`'1984'`, `'1337'`) as the id field - same role as an EVM contract address or a Solana SPL mint address. The `knownTokensIndex` lookup lowercases ids for EVM chains; these numeric string ids are case-invariant so the lookup works correctly.

## follow-up PRs

- **PR B**: balance resolver adding a `pallet_assets.Account` branch for non-native Polkadot tokens (R2 audit #5 part 2)
- **PR C**: `build_polkadot_asset_send` using `api.tx.assets.transfer(asset_id, dest, amount)` in mcp-ts (R2 audit #5 part 3)

This PR (registry only) unblocks token row display + price lookups. PRs B + C follow once NeOMakinG confirms the token id shape is right.

## risk

Low - additive only. New chain section in `leanTokens`. No existing paths touched.

🤖 Agent-generated

### DIFF
--- a/packages/core/chain/coin/knownTokens/index.ts
+++ b/packages/core/chain/coin/knownTokens/index.ts
@@ -761,6 +761,24 @@ const leanTokens: Partial<LeanChainTokensRecord> = {
       priceProviderId: 'usd-coin',
     },
   },
+  [Chain.Polkadot]: {
+    // Polkadot Asset Hub (parachain 1000) — pallet_assets tokens.
+    // asset_id is used as the token identifier instead of a contract address.
+    // Decimals verified live via state_getStorage on Assets.Metadata at 2026-05-25.
+    // On-chain symbol for 1984 is "USDt" but we normalise to "USDT" for consistency.
+    '1984': {
+      ticker: 'USDT',
+      logo: 'usdt',
+      decimals: 6,
+      priceProviderId: 'tether',
+    },
+    '1337': {
+      ticker: 'USDC',
+      logo: 'usdc',
+      decimals: 6,
+      priceProviderId: 'usd-coin',
+    },
+  },
   ...knownCosmosTokens,
 }
 
--- a/packages/core/chain/coin/knownTokens/index.ts
+++ b/packages/core/chain/coin/knownTokens/index.ts
@@ -770,6 +770,11 @@ const leanTokens: Partial<LeanChainTokensRecord> = {
       ticker: 'USDT',
       logo: 'usdt',
       decimals: 6,
+      // NOTE: CoinGecko's 'tether' coin doesn't list polkadot-asset-hub in its
+      // platforms map (only Ethereum/Solana/Tron/TON/Near/etc). Price still resolves
+      // correctly via /simple/price?ids=tether but any future platform-verification
+      // logic (e.g. checking coin.platforms['polkadot-asset-hub']) will find nothing
+      // and should special-case this entry.
       priceProviderId: 'tether',
     },
     '1337': {
--- a/packages/core/chain/coin/balance/resolvers/polkadot.test.ts
+++ b/packages/core/chain/coin/balance/resolvers/polkadot.test.ts
@@ -1,0 +1,112 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const queryUrlMock = vi.hoisted(() => vi.fn())
+
+vi.mock('@vultisig/lib-utils/query/queryUrl', () => ({
+  queryUrl: queryUrlMock,
+}))
+
+import { getPolkadotCoinBalance } from './polkadot'
+
+// A valid Polkadot SS58 address (Alice, network prefix 0).
+const VALID_DOT_ADDRESS = '15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6Sp5'
+
+describe('getPolkadotCoinBalance — interim token guard (PR A / #562)', () => {
+  // PR A adds USDT (id=1984) + USDC (id=1337) to knownTokens.
+  // PR B (TBD) will wire pallet_assets.Account queries for those assets.
+  // Until PR B lands the resolver must return 0n for any coin with an id
+  // rather than returning the native DOT System.Account free balance —
+  // which would show the user a misleading DOT balance under a USDT row.
+
+  beforeEach(() => {
+    queryUrlMock.mockReset()
+  })
+
+  it('returns 0n for USDT (asset_id=1984) without hitting RPC', async () => {
+    const balance = await getPolkadotCoinBalance({
+      chain: Chain.Polkadot,
+      address: VALID_DOT_ADDRESS,
+      id: '1984', // TODO(PR B): wire pallet_assets.Account query here
+    })
+
+    expect(balance).toBe(0n)
+    // RPC must NOT be called — guard must short-circuit before any network I/O
+    expect(queryUrlMock).not.toHaveBeenCalled()
+  })
+
+  it('returns 0n for USDC (asset_id=1337) without hitting RPC', async () => {
+    const balance = await getPolkadotCoinBalance({
+      chain: Chain.Polkadot,
+      address: VALID_DOT_ADDRESS,
+      id: '1337', // TODO(PR B): wire pallet_assets.Account query here
+    })
+
+    expect(balance).toBe(0n)
+    expect(queryUrlMock).not.toHaveBeenCalled()
+  })
+
+  it('returns 0n for any non-empty id (generic guard coverage)', async () => {
+    const balance = await getPolkadotCoinBalance({
+      chain: Chain.Polkadot,
+      address: VALID_DOT_ADDRESS,
+      id: '9999',
+    })
+
+    expect(balance).toBe(0n)
+    expect(queryUrlMock).not.toHaveBeenCalled()
+  })
+})
+
+describe('getPolkadotCoinBalance — native DOT path (no id)', () => {
+  beforeEach(() => {
+    queryUrlMock.mockReset()
+  })
+
+  it('queries RPC for native DOT (no id present)', async () => {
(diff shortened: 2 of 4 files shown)


gomesalexandre (commented):
multi-lane review: claude correctness + adversarial pass. draft so posting as COMMENT per rules.

---

## verdict: looks solid, one preferably-blocking, one should-fix, one question

---

## findings

### preferably-blocking: no test coverage for the new Polkadot entries

The Solana Bugs J PR set a precedent that's hard to ignore: `knownTokensSolana.test.ts` pins the exact ids, decimals, priceProviderId, and ticker so future refactors can't silently nuke them. This PR adds two entries to the same data structure, touches zero test files (`git diff main...HEAD --name-only` → only `index.ts`), and the CI suite will just pass because there's no invariant to break.

The risk is real: the Polkadot entries live in a `[Chain.Polkadot]` key under `leanTokens`. If someone later restructures the `knownTokens` export (say, splits it per-chain, or changes how the Polkadot namespace is formed) the entries silently disappear and the UI goes back to showing zero Polkadot tokens with no failing test to catch it.

**action:** mirror what `knownTokensSolana.test.ts` does — create `packages/sdk/tests/unit/tools/token/knownTokensPolkadot.test.ts` pinning:
- `knownTokensIndex[Chain.Polkadot]['1984']` → ticker `'USDT'`, decimals `6`, priceProviderId `'tether'`  
- `knownTokensIndex[Chain.Polkadot]['1337']` → ticker `'USDC'`, decimals `6`, priceProviderId `'usd-coin'`
- `knownTokens[Chain.Polkadot].length >= 2`

The template is already there, copy-paste job.

---

### should-fix: `priceProviderId: 'tether'` has no polkadot-asset-hub platform on CoinGecko

Queried `https://api.coingecko.com/api/v3/coins/tether` live. The `platforms` map returns 11 entries — none with a `polkadot` or `polkadot-asset-hub` key. USDC (`usd-coin`) does: `"polkadot": "1337"`. So USDC is fully ratified by CoinGecko for this network; USDT is not.

```
curl https://api.coingecko.com/api/v3/coins/tether → platforms: { ethereum, solana, tron, ton, near-protocol, ... } — no polkadot key
curl https://api.coingecko.com/api/v3/coins/usd-coin → platforms: { ..., "polkadot": "1337" } ✓
```

Practically this means: CoinGecko knows Tether's on Asset Hub (they list the balance via API), but they don't surface a canonical polkadot entry under the tether coin's `platforms`. The `priceProviderId: 'tether'` slug itself resolves fine for price (`/simple/price?ids=tether` → `$0.999`) so the price lookup still works — the slug is correct. But if the SDK ever gains a "verify chain+address against CoinGecko platforms" step (like some chains do), USDT would fail. Not blocking today but worth a comment in-code so the next person doesn't think it's an oversight.

**action:** add a comment on the USDT entry:
```ts
// priceProviderId 'tether' resolves correctly for price lookups.
// CoinGecko's tether coin does not list polkadot-asset-hub in its
// platforms map (unlike usd-coin which lists "polkadot": "1337"),
// so if we add platform-verification logic later, revisit this.
```

---

### q: balance resolver doesn't handle token ids yet — will this produce silent zero balances?

`getPolkadotCoinBalance` (at `packages/core/chain/coin/balance/resolvers/polkadot.ts`) is hardwired to query `System.Account` (the DOT native balance). It receives `CoinBalanceResolverInput = AccountCoinKey = { chain, id?, address }`. The `id` field (which would be `'1984'` or `'1337'` for these tokens) is never read.

When the UI surfaces the new token rows and tries to fetch balances, the resolver will return the native DOT balance for the address — not the USDT/USDC balance from `pallet_assets.Account`. Since PR B (balance resolver) is explicitly deferred this is by design, but worth confirming: is the current behavior "returns 0 / returns DOT balance" or does the dispatcher gracefully skip tokens with unknown chain-kind ids? Just want to make sure the UI doesn't show a misleading non-zero DOT value for a USDT row.

---

## what's clean

**praise:** decimals verification is bulletproof. Raw SCALE hex inlined in the PR body, on-chain query cited with storage prefix, is_frozen checked, and the "USDt" vs "USDT" normalization decision is explicitly documented in both commit message and PR body. That's the right level of rigor for a live-network data entry.

**praise:** single-file diff, purely additive, zero existing paths touched. Low blast radius, clean scope.

**id slot pattern is correct.** Checked: `knownTokensIndex` lowercases ids via `coin.id.toLowerCase()`. For numeric strings `'1984'.toLowerCase() === '1984'` — no-op, safe. EVM uses checksummed hex, Solana uses base58 mint, TON uses raw address strings, Tron uses base58 contract. Asset Hub numeric string is unconventional but the mechanism is the same: the id is whatever opaque string the chain uses to identify the token. Fine.

**ticker normalization to `'USDT'` is correct.** On-chain symbol is `"USDt"` but all other USDT entries across Ethereum, Solana, Tron, Ton use `'USDT'` (uppercase). CoinGecko `symbol` field is also `'usdt'` — our all-caps convention matches. No collision risk: `knownTokensIndex` is chain-scoped (`Record<Chain, Record<string, KnownCoin>>`), so Tron USDT and Polkadot USDT live in separate namespaces.

**logo slugs `'usdt'` / `'usdc'` are consistent.** Same slugs used on every other chain. No full URL needed.

**CI:** `pending` at review time (just pushed). Lint + Unit + Integration all queued. Will check back but the diff is data-only so the risk of a type error is minimal.

---

@NeOMakinG would love your eyes on the id-slot convention specifically — does the balance resolver PR (PR B) plan to parse `input.id` as the asset_id integer for the `pallet_assets.Account` query? Just want to confirm the handshake before merging.

NeOMakinG (approved):
gm — multi-lane pass: claude correctness + fund-safety + GLM chain-edge. conceptual-only (no runtime this pass). lgtm.

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | `preferably-blocking` — no test coverage for knownTokens Polkadot entries | ✅ fixed in a8ffc66c — added `knownTokensPolkadot.test.ts` (pins id, decimals, priceProviderId, ticker for both tokens) + `polkadot.test.ts` (interim guard + native DOT path) |
| 1 | `should-fix` — CoinGecko `tether` coin missing `polkadot-asset-hub` in platforms map, no comment | ✅ fixed in a8ffc66c — NOTE comment added directly on the USDT priceProviderId field explaining the gap and the risk if platform-verification logic is added later |
| 1 | `q` — balance resolver would return native DOT balance for USDT/USDC rows | ✅ fixed in a8ffc66c — `if (input.id) return BigInt(0)` guard added with clear TODO(PR B) comment; guard exercised by 3 tests in polkadot.test.ts |

## risk
Additive registry entry only. Worst-case on bad data: wrong USDT/USDC balance shown (currently 0 until PR B). Asset IDs verified against polkadot-fellows runtime source (asset_id 1984 = Tether USD, 1337 = USD Coin, both 6 decimals). Balance guard prevents DOT-balance leakage under token rows until PR B lands.

## verdict
APPROVE — ship it 🫡

packages/core/chain/coin/knownTokens/index.ts
```diff
@@ -761,6 +761,29 @@ const leanTokens: Partial<LeanChainTokensRecord> = {
       priceProviderId: 'usd-coin',
     },
   },
+  [Chain.Polkadot]: {
+    // Polkadot Asset Hub (parachain 1000) — pallet_assets tokens.
+    // asset_id is used as the token identifier instead of a contract address.
+    // Decimals verified live via state_getStorage on Assets.Metadata at 2026-05-25.
+    // On-chain symbol for 1984 is "USDt" but we normalise to "USDT" for consistency.
+    '1984': {
```
NeOMakinG, line 769: `suggestion:` Consider documenting exact SCALE storage query pattern for follow-up PR — Current TODO is general. Adding the exact composite key layout (Twox128(\"Assets\") ++ Twox128(\"Account\") ++ Twox64Concat(asset_id) ++ Twox64Concat(account_id)) would help the next reviewer verify the PR B implementation.

fix: Add detailed TODO comment with the exact SCALE key construction pattern

NeOMakinG (approved):

## suggestion
### 1. Consider documenting exact SCALE storage query pattern for follow-up PR — `packages/core/chain/coin/knownTokens/index.ts:769`
Current TODO is general. Adding the exact composite key layout (Twox128(\"Assets\") ++ Twox128(\"Account\") ++ Twox64Concat(asset_id) ++ Twox64Concat(account_id)) would help the next reviewer verify the PR B implementation.
action: Add detailed TODO comment with the exact SCALE key construction pattern

## risk
Low — additive only. New chain section in leanTokens with USDT/USDC entries. Guard in balance resolver prevents wrong data until follow-up PR lands. No existing paths touched.

## verdict
APPROVE — ship it 🫡

PR #577, opened 2026-05-26
feat(mpc): env-gated diag logging for relay-decrypt ghash tag investigation

gomesalexandre:
## what

Adds env-gated diagnostic logging in `@vultisig/sdk` DKLS keysign relay-decrypt path. NO behavior change when the flag is unset. When set, logs body/key/nonce shapes that pinpoint the source of "aes/gcm: invalid ghash tag" errors reported by Pan (Discord 12:52 today) + NakYot (14:28).

## why

Two deep spikes + one sim repro attempt converged on: bug is in `fromMpcServerMessage` relay-decrypt path, wire format and KDF wire-compatible analytically, need device-level byte capture to identify the actual failure mode. Local repro blocked (no test vault can start a successful keysign with VultiServer).

Shipping this lets us capture the bytes in real failed sessions and root-cause the bug.

## how

- Env flag: `VULTISIG_DIAG_MPC_RELAY=1` enables. Default off.
- `packages/core/mpc/message/server.ts` - logs before `decryptWithAesGcm`: `body_len`, `decoded_len`, `nonce_hex`, `first32_hex`, `key_first16`
- `packages/core/mpc/keysign/index.ts` - logs in `processInbound` loop before `inputMessage`: `from`, `to`, `session_id`, `body_len`, `message_id`
- No new deps. Uses `process.env` directly, consistent with `cosigner.ts` convention.

## receipts

- Unit test: logging is OFF by default (no `console.log` spy fires without flag)
- Unit test: logging fires with `VULTISIG_DIAG_MPC_RELAY=1`, correct fields present (`body_len`, `nonce_hex`, `key_first16`)
- 3/3 `server.test.ts` tests pass, CI will confirm

No sim receipt applicable - purely observability, no user-visible surface.

## risk

Low. Pure addition gated by env flag. Default-off matches current behavior exactly. No new deps. The `key_first16` slice exposes only the first 8 bytes of the hex-encoded key (16 chars = 8 bytes of the 32-byte key) - enough to correlate mismatches without leaking the full key.

## followup

Once we capture the bytes in a real failure (Pan / NakYot / any new report), root-cause + ship the actual fix and revert this diag. The companion vultiagent-app wiring (`EXPO_PUBLIC_VULTISIG_DIAG_MPC_RELAY=1` in staging env) is deferred until this SDK PR lands + publishes.

🤖 Agent-generated

## Summary by CodeRabbit

* **New Features**
  * Added support for Polkadot Asset Hub USDT and USDC tokens
  * Added Tron native send memo data support

* **Bug Fixes**
  * Fixed Tron TRC-20 fee estimation to properly account for sender's available energy
  * Resolved spurious fee charges when bandwidth is available on Tron

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/577?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/mpc-diag-relay-logging.md
+++ b/.changeset/mpc-diag-relay-logging.md
@@ -1,0 +1,13 @@
+---
+'@vultisig/core-mpc': patch
+---
+
+feat(mpc): env-gated diagnostic logging for relay-decrypt ghash tag investigation
+
+Adds non-default-on diagnostic logging to `fromMpcServerMessage` and the
+`receiveMessages` keysign relay loop, gated on `VULTISIG_DIAG_MPC_RELAY=1`.
+Logs envelope shape (`body_len`, `decoded_len`, `nonce_hex`, first 32 bytes
+of ciphertext) plus a `key_fingerprint` (sha256-truncated of decoded key
+bytes, NOT raw key material) for cross-node correlation of the persistent
+"aes/gcm: invalid ghash tag" failures. Behavior unchanged when the env flag
+is absent.
--- a/.changeset/polkadot-asset-hub-tron-memo.md
+++ b/.changeset/polkadot-asset-hub-tron-memo.md
@@ -1,0 +1,16 @@
+---
+'@vultisig/core-chain': minor
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': minor
+---
+
+## New
+
+- Polkadot Asset Hub USDT (asset_id 1984) + USDC (asset_id 1337) token registry (#562)
+- Polkadot `pallet_assets.Account` balance resolver for Asset Hub tokens - replaces placeholder 0n guard (#563)
+- Tron native send `data` field (proto field 12) for THORChain memos + exchange deposit memos; `BuildTronSendOptions` and `BuildTrc20TransferOptions` gain optional `data?: Uint8Array` field (#559)
+
+## Fixed
+
+- Tron TRC-20 fee estimate now subtracts sender's available energy before charging TRX (#556)
+- Tron native send free bandwidth check prevents spurious fee charge when bandwidth is available (#555)
--- a/.config/eslint.config.mjs
+++ b/.config/eslint.config.mjs
@@ -50,6 +50,10 @@ export default [
       // WASM files copied by build tools
       '**/public/wallet-core.js',
       '**/public/wallet-core.wasm',
+      // Metro symlink+exports shim — CommonJS module.exports required so
+      // Metro's resolver can pick it up at runtime. Source lint rules don't
+      // apply; this is build-tool plumbing, not application source.
+      'packages/sdk/react-native.js',
     ],
   },
   ...fixupConfigRules(
--- a/package.json
+++ b/package.json
@@ -189,6 +189,8 @@
     "readable-stream": "3.6.2",
     "@noble/hashes": "1.8.0",
     "@lifi/sdk/@solana/web3.js": "1.98.4",
-    "@cosmjs/encoding/@scure/base": "1.2.6"
+    "@cosmjs/encoding/@scure/base": "1.2.6",
+    "tmp@npm:^0.0.33": "0.2.6",
+    "tmp@npm:^0.2.0": "0.2.6"
   }
 }
--- a/packages/core/mpc/keysign/index.ts
+++ b/packages/core/mpc/keysign/index.ts
@@ -147,6 +147,21 @@ export const keysign = async ({
         continue
       }
 
+      // Diagnostic logging for "aes/gcm: invalid ghash tag" investigation.
+      // Enable via VULTISIG_DIAG_MPC_RELAY=1. Remove once root-cause is confirmed.
+      if (process.env.VULTISIG_DIAG_MPC_RELAY === '1') {
+        console.log(
+          '[DIAG-MPC-INBOUND]',
+          JSON.stringify({
+            from: msg.from,
+            to: msg.to,
+            session_id: msg.session_id,
+            body_len: msg.body?.length ?? null,
+            message_id: messageId,
+          })
+        )
+      }
+
       const accepted = session.inputMessage(fromMpcServerMessage(msg.body, hexEncryptionKey))
       if (accepted) {
         processedMessages[cacheKey] = true
--- a/packages/core/mpc/message/server.test.ts
+++ b/packages/core/mpc/message/server.test.ts
@@ -4,6 +4,7 @@ describe('mpc server message encoding', () => {
   afterEach(() => {
     vi.unstubAllGlobals()
     vi.resetModules()
+    delete process.env.VULTISIG_DIAG_MPC_RELAY
   })
 
   it('round-trips when globalThis.Buffer is undefined (browser-like)', async () => {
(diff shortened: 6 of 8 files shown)


NeOMakinG (commented):
## testing approach

triple-lane static review (Claude + Codex + GLM). No runtime required — this is a pure logging addition with no tx-path behavior change. Read `packages/core/mpc/message/server.ts` and `packages/core/mpc/keysign/index.ts` at HEAD `69e70fc0` from canonical clone. Traced: env gate correctness, crypto-material exposure, decrypt-path interference, and test coverage.

## should-fix

**1. `key_first16` logs 8 raw bytes of the AES-256 relay key — key material should never enter logs.**

`packages/core/mpc/message/server.ts`, the `key_first16` field:
```ts
key_first16: hexEncryptionKey.slice(0, 16),  // 16 hex chars = 8 raw bytes of the relay key
```
Even behind `VULTISIG_DIAG_MPC_RELAY=1`, diagnostic logs are persistent: syslog rotation, crash reporters, log shipping pipelines, and operator screenshots all capture them. 8 bytes is not brute-forceable against a 256-bit key, but the principle that key material never enters any log output is load-bearing; making exceptions here creates a precedent that erodes that boundary. If you need to distinguish which key is in use across two nodes, log a non-reversible fingerprint instead:
```ts
key_fingerprint: require('crypto').createHash('sha256').update(hexEncryptionKey).digest('hex').slice(0, 16)
```
That identifies the key without exposing any of its bits.

**2. Double-decode of `body` in the diag branch creates a different failure site than the real decrypt path.**

`packages/core/mpc/message/server.ts`:
```ts
const decoded = Buffer.from(body, encryptedEncoding)  // diag path
// ...
value: Buffer.from(body, encryptedEncoding),           // main decrypt
```
If a malformed message arrives, the exception originates from the diag decode with a stack trace pointing there — not from `decryptWithAesGcm`. This makes the `invalid ghash tag` investigation harder because error attribution points to the wrong site. Cache the decoded buffer once and pass it to both the log and the decrypt:
```ts
const encryptedBuf = Buffer.from(body, encryptedEncoding)
if (process.env.VULTISIG_DIAG_MPC_RELAY === '1') {
  console.log('[DIAG-MPC-RELAY]', JSON.stringify({
    body_len: body.length,
    decoded_len: encryptedBuf.length,
    nonce_hex: encryptedBuf.subarray(0, 12).toString('hex'),
    first32_hex: encryptedBuf.toString('hex').slice(0, 64),
  }))
}
return Buffer.from(
  decryptWithAesGcm({ key: Buffer.from(hexEncryptionKey, 'hex'), value: encryptedBuf }).toString(plainTextEncoding),
  'base64'
)
```

## q

**Q1: Is `first32_hex` useful for the ghash investigation?** Bytes 0–11 are the nonce (already in `nonce_hex`), bytes 12–31 are 20 bytes of raw ciphertext. AES-GCM ciphertext is keystream XOR, not plaintext, so no MPC message content is exposed. Confirm this is worth the log noise vs. just nonce + lengths.

**Q2: Test coverage for `DIAG-MPC-INBOUND` in `keysign/index.ts`.** The negative test only covers `server.ts`. The `DIAG-MPC-INBOUND` log only touches envelope metadata (`from`, `to`, `session_id`, `body_len`, `message_id`) — no crypto fields — so all three review lanes agreed coverage is not required before merge. Noting it in case you want symmetry.

## risk

low-to-medium. The env gate is correctly strict (`=== '1'`), off by default, and the tag-validation path is unaffected — logging fires before the decrypt call and does not wrap or catch it. The only live concern is `key_first16` landing in a log collector if someone enables this in production to debug a user-reported ghash error.

## verdict

comment — two should-fixes, one q. Gate logic and decrypt-path isolation are correct. The key-material logging and double-decode are worth addressing before merge.

---
*triple-lane: Claude (correctness + fund-safety) + Codex gpt-5.4 (adversarial, scoped to contract-shape + error-flow) + GLM-4.6 (third-eye). All three lanes agree on key_first16; nonce assessed non-critical by Codex + Claude, overriding GLM blocker classification.*

NeOMakinG (approved):
## prior issues — status

| r1 | 1 | should-fix | `key_first16` logged 8 raw bytes of the relay AES-256 key | ✅ fixed | replaced with `key_fingerprint: sha256(hexEncryptionKey).slice(0,16)`; source comment + test both document that no key bits reach the log |
| r1 | 2 | should-fix | double-decode — `Buffer.from(body, encryptedEncoding)` ran twice (once in diag branch, once in main decrypt), obscuring error origin | ✅ fixed | `encryptedBuf` decoded once at function top, shared to both sites; comment explains the single-site rationale for the ghash investigation |

Both r1 should-fixes resolved in 4f436e16. Test coverage added: positive shape-match on `key_fingerprint` + negative assertion that `key_first16` never appears. Clean.

gomesalexandre (commented):
## testing approach

Triple-lane: Claude (env gate correctness, crypto-material exposure, decrypt-path interference), Codex adversarial (independent contract-shape + error-flow audit), GLM third-eye (RN bundle import of `crypto` module, process.env availability in browser contexts, base64/hex encoding edge cases).

Traced both diagnostic sites: `server.ts` (decrypt path) and `keysign/index.ts` (inbound loop). Verified env gate (`=== '1'` strict equality). Confirmed `createHash` from Node `crypto` is not available in RN Metro bundle without polyfill — see note below.

---

## prior issues — status

|---|---|---|---|---|---|
| NeO r1 | 1 | should-fix | `key_first16` logs 8 raw bytes of relay key — key material in logs | ✅ fixed | tackled at `4f436e16` — replaced with `key_fingerprint = sha256(key).slice(0, 16)` |
| NeO r1 | 2 | should-fix | double-decode of `body` creates wrong error attribution site for ghash investigation | ✅ fixed | tackled at `4f436e16` — `encryptedBuf` decoded once, shared by diag and decrypt |
| NeO r1 | Q1 | q | `first32_hex` usefulness for ghash investigation | ℹ️ documented | 20 bytes of raw ciphertext, not plaintext; non-reversible per AES-GCM keystream property |
| NeO r1 | Q2 | q | test coverage for `DIAG-MPC-INBOUND` in `keysign/index.ts` | ℹ️ documented | envelope metadata only (no crypto fields); three lanes agreed coverage not required before merge |

---

## q

### Q1 — `import { createHash } from 'crypto'` in RN Metro bundle

`server.ts` now imports `createHash` from Node's `crypto` module. `server.ts` is imported by `keysign/index.ts` which runs in the RN app (not Node). Metro does NOT polyfill Node's `crypto` module by default — `require('crypto')` in RN returns an empty object or throws depending on the Metro version.

This is behind `VULTISIG_DIAG_MPC_RELAY=1` which is set via `process.env` (also Node-only; RN uses `EXPO_PUBLIC_*` via Babel). So both the flag and the `createHash` call are unreachable in production RN builds. The risk is: if a future refactor moves this code somewhere that runs in RN, it silently fails.

Current assessment: safe for this PR's purpose (SDK used in Node/server contexts for diag). But worth noting before the companion `vultiagent-app` wiring lands.

---

## risk

low. Diag logging is behind `=== '1'` gate, off by default. `createHash` import is Node-only and unreachable in production RN builds while the flag is env-gated via `process.env`. Key fingerprint is a non-reversible digest. Decrypt path unchanged.

## verdict

`comment` — NeO's two should-fixes tackled at `4f436e16`. Q1 on `createHash` in RN context is a heads-up for the companion vp wiring, not a blocker here.

---
triple-lane self-review · claude-sonnet-4-6 · HEAD `4f436e16`

NeOMakinG (approved):
## prior issues — status

| r1 | 1 | should-fix | `key_first16` logs 8 raw bytes of relay key — key material in logs | ✅ fixed | `e3432c31` — replaced with sha256-truncated fingerprint |
| r1 | 2 | should-fix | double-decode of `body` creates wrong error attribution site | ✅ fixed | `e3432c31` — `encryptedBuf` decoded once, shared by diag and decrypt |
| r1 | 3 | should-fix | missing changeset | ✅ fixed | `1ca76598` — `.changeset/mpc-diag-relay-logging.md` adds `@vultisig/core-mpc: patch` |
| r3 | Q1 | q | `createHash` from Node `crypto` imported into file that is also bundled by Metro | ℹ️ documented | diag block is `process.env`-gated and unreachable in RN production builds; no change in `1ca76598`, concern is a heads-up for the companion app wiring, not a blocker |

---

## testing approach

Diff-walk only (round ≥ 2 per memory). Verified `server.ts` at HEAD `1ca76598` and the two MPC-relevant commits: `e3432c31` (sha256 fingerprint replace) and `1ca76598` (hex-vs-bytes fix + changeset).

---

## hex-vs-bytes fix

`key_fingerprint: createHash('sha256').update(Buffer.from(hexEncryptionKey, 'hex')).digest('hex').slice(0, 16)`

Correct. `Buffer.from(hexEncryptionKey, 'hex')` decodes the hex string to raw bytes before hashing, so `sha256(rawBytes)` is computed — not `sha256(hexString)`. The fingerprint is stable regardless of caller-side casing (`'cc...cc'` and `'CC...CC'` both decode to the same 32-byte key and produce the same digest). No key bits reach the log.

The case-stability test at `server.test.ts` (lowerKey vs upperKey, `expect(fingerprints[0]).toBe(fingerprints[1])`) directly exercises this invariant.

---

## `keysign/index.ts` scope

The new diagnostic block in the inbound relay loop logs envelope metadata only: `from`, `to`, `session_id`, `body_len`, `message_id`. No `createHash`, no key fingerprint at this site. Scope is clean — it is a companion inbound-message diagnostic site, not a separate fingerprint computation path.

---

## verdict

approve. All r1/r2 findings resolved. The hex-vs-bytes fingerprint fix is mechanically correct and pinned by a case-stability regression test. Changeset present. r3 Q1 (RN `createHash`) remains informational and unchanged.

---
claude-sonnet-4-6 · HEAD `1ca76598`

NeOMakinG (approved):
Re-affirms r3 APPROVE on `1ca76598` (review #4371703467, 2026-05-27T10:40Z). Delta to `4242f422` is two infra commits landed from main: `chore(eslint): ignore packages/sdk/react-native.js metro shim` and `chore(deps): pin tmp >=0.2.6 via resolutions (GHSA-ph9p-34f9-6g65)`. No code change to the MPC diag-logging scope of this PR (server.ts and keysign/index.ts unchanged in the delta). Security pin is a legitimate response to the upstream `tmp` advisory; yarn.lock resolution-only. Re-APPROVED.

PR #579, opened 2026-05-27
fix(cosmos): LCD fallback for account-info AND balance when RPC returns null/0

gomesalexandre:
## what

`getCosmosAccountInfo` silently falls back to `sequence: 0` whenever `StargateClient.getAccount()` returns null. That fallback is correct for accounts that have genuinely never been funded — but it's a **doomed broadcast** for accounts that exist on-chain when the RPC just couldn't decode an extended account type or hit a transient infra issue.

User-reported symptom: `"Broadcast failed (code 32): account sequence mismatch, expected 84, got 0"` on a 303 LUNA Terra v2 send. The chain knew the address had done 84 prior txs; our tx claimed sequence 0 because RPC `getAccount` returned null.

## how

Add an LCD REST fallback after RPC null:

```ts
let accountNumber = accountInfo?.accountNumber
let sequence = accountInfo?.sequence

if (accountInfo === null) {
  const lcd = await fetchAccountViaLcd(chain, address)
  if (lcd) {
    accountNumber = lcd.accountNumber
    sequence = lcd.sequence
  }
}
```

LCD endpoint `/cosmos/auth/v1beta1/accounts/{addr}` returns a JSON envelope we can shape-parse for the standard `BaseAccount` AND extended types (`base_vesting_account.base_account`, `ModuleAccount.base_account`). If BOTH RPC and LCD return null, we fall through to `sequence: 0` (preserves correct behavior for genuinely-new accounts).

LCD uses a different infra path (REST vs Tendermint RPC), is less likely to fail simultaneously, and handles account-type-decode cases the protobuf RPC path misses.

## why this approach

- Smallest surface change — single file in core-chain
- Preserves the "new account → sequence 0" semantics that other callers depend on
- LCD URLs are already configured per-chain in `cosmosRpcUrl.ts` — no new env / config
- Doesn't change return signature; existing callers (cosmos / thor / maya / qbtc resolvers) get the fix transparently

## tests

6 new cases in `getCosmosAccountInfo.test.ts`:
- RPC returns account → use directly (LCD not called)
- RPC null + LCD BaseAccount → uses LCD sequence/accountNumber
- RPC null + LCD PeriodicVestingAccount wrapper → parses nested `base_vesting_account.base_account`
- RPC null + LCD ModuleAccount wrapper → parses nested `base_account`
- RPC null + LCD throws → falls through to 0 (genuinely-new account)
- RPC null + LCD malformed → falls through to 0

## receipts

- Unit tests covering all 6 paths (running locally, will surface in CI)
- Sim verification of the actual LUNA send broadcast succeeding — pending (need a vault loaded with LUNA + access to the broken broadcast path; will follow-up post-merge or via dogfooder)

## refs

User Discord report 2026-05-27 ~AM: 303.485 LUNA broadcast failed with `code 32 ... expected 84, got 0`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved account information retrieval for extended Cosmos account types to prevent invalid transaction broadcasting with incorrect sequence numbers.
  * Fixed coin balance lookup on affected Cosmos environments (e.g., Hermes/React Native) to correctly display funded addresses instead of showing zero balance.

* **Tests**
  * Added comprehensive test coverage for account and balance resolution with LCD fallback scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/579?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


gomesalexandre (commented):
## testing approach

Triple-lane: Claude (LCD fallback shape parsing, cosmosRpcUrl endpoint compatibility, null-path semantics), Codex adversarial (concurrent RPC+LCD, error swallowing, type coercion), GLM third-eye (base64/bech32 encoding in LCD response, chain-specific account type variations, sequence number integer edge cases).

Traced `parseLcdAccount` for all three nesting shapes (flat `BaseAccount`, `base_account` wrapper, `base_vesting_account.base_account` wrapper). Verified `cosmosRpcUrl` values are REST/LCD base URLs (not Tendermint RPC). Spot-checked THORChain Liquify and MayaChain endpoints — both serve `/cosmos/auth/v1beta1/accounts` with valid 3xx-code bech32-format errors on invalid addresses (confirms the path exists).

All 6 test cases in `getCosmosAccountInfo.test.ts` traced. The `Promise.all([getAccount, getBlock])` parallelization is a nice cleanup.

---

## should-fix

none found

---

## q

### Q1 — `cosmosRpcUrl[chain]` is used as the LCD base URL

`cosmosRpcUrl` was designed as REST base URLs (not Tendermint RPC), which is why `/cosmos/auth/v1beta1/accounts` works. But the name is `cosmosRpcUrl` — it's a naming footgun. Future callers who see `cosmosRpcUrl` might assume it's a Tendermint RPC endpoint (`:26657`) and hit the wrong port. Worth a follow-up rename to `cosmosRestUrl` or at minimum a comment clarifying it's the REST/LCD base.

Not a blocker for this PR since the actual URLs are LCD-compatible.

### Q2 — `Number(base.sequence ?? '0')` on a large sequence number

Cosmos sequence numbers fit in uint64 on-chain. At 9 quadrillion txs a JS `Number` would lose precision, but that's effectively impossible in practice. `Number.isFinite` guards against `NaN` from malformed strings. Noting in case the chain is ever used at high frequency (e.g. smart-contract-heavy chains) — `BigInt` would be the safe alternative long-term.

---

## risk

low. LCD fallback only fires on the narrow `RPC returns null` case (extended account type or transient infra). New-account semantics (`sequence=0`) preserved. No change to caller signatures. THORChain/MayaChain/Terra/all tested chains serve the standard LCD account endpoint.

## verdict

`approve` — clean implementation. Q1/Q2 are follow-up notes, not blockers. The fix closes a real user-reported broadcast failure (sequence mismatch on LUNA Terra v2).

---
triple-lane self-review · claude-sonnet-4-6 · HEAD `fa3f0f5`

NeOMakinG (approved):
triple-lane review: Claude primary + GLM third-eye (Codex lane unavailable — companion CLI auth error). Fund-safety surface: `accountNumber`/`sequence` → MPC keysign payload. 6 tests, 0 blockers. APPROVE.

## testing approach

- Traced all 3 call sites: `cosmos.ts:45`, `thor.ts:17`, `maya.ts:11`, `buildCosmosPayload.ts:156,213`
- Read `assertFetchResponse` → confirms non-2xx throws `HttpResponseError` → `fetchAccountViaLcd` catch returns `null` — dual-null path is correct
- Live-verified THORChain LCD: `GET /cosmos/auth/v1beta1/accounts/thor1t60f02r8jvzjrhtnjgfj4ne6rs5wjnejwmj7fh` → `{account_number:"45060",sequence:"361671"}` — standard BaseAccount shape, no custom nesting needed
- Live-verified Cosmos SDK 404 behavior: `GET /cosmos/auth/v1beta1/accounts/cosmos1lml0alh7...` (never-funded address) → HTTP 404 `code:5 "account not found"` — confirms dual-null correctly maps to sequence=0 only for genuinely unfunded accounts, not "LCD returned zero envelope"
- Confirmed MayaChain/THORChain both return gRPC-over-REST from `/cosmos/auth/v1beta1/accounts/` — standard path is live
- Reviewed `parseLcdAccount` nesting priority: `base_vesting_account.base_account → base_account → acc` — correctly handles PeriodicVestingAccount, ContinuousVestingAccount, ModuleAccount

## should-fix

1. The changeset file `.changeset/polkadot-asset-hub-tron-memo.md` bundles Polkadot Asset Hub + Tron entries but does not include an entry for this cosmos LCD fallback fix. Any downstream consumer scanning the changelog will miss that `getCosmosAccountInfo` now silently recovers from extended account types that previously returned sequence:0. Suggest adding a `## Fixed` bullet: `- Cosmos account info LCD fallback for extended account types that RPC cannot decode (vesting, module wrappers) — prevents doomed txs with sequence:0`. action: add one line to the existing `## Fixed` section in that changeset file.

## q

1. `getCosmosAccountInfo.ts:97`: the return still has `pubkey: accountInfo?.pubkey ?? null`. When the LCD path fires, `accountInfo` is `null`, so `pubkey` is always `null` for LCD-resolved accounts even when the LCD response contains a `pub_key` field. Callers (`cosmos.ts`, `thor.ts`, `maya.ts`, `buildCosmosPayload.ts`) don't currently consume `pubkey` — but it's in the return type. Is there a future caller that needs it? If not, no action needed.

## risk

All three call sites only consume `accountNumber` and `sequence` from the return value — not `pubkey`. The fallback is behind a strict `=== null` guard, doesn't fire for the happy path, and degrades to the pre-existing sequence:0 behavior on any LCD failure. Blast radius on merge: a transaction from a vesting/module account that previously failed with sequence mismatch will now succeed. No regressive path identified.

## verdict

APPROVE. Fund-safety critical paths (dual-null → sequence:0, LCD 404 → throw → null, assertFetchResponse non-2xx guard) all verified both in code and against live endpoints. Test suite covers all 5 branches including the dual-null edge case. One should-fix (missing changelog entry, non-blocking).

---
reviewed via [Claude Code](https://claude.com/claude-code), triple-lane (Claude primary + GLM `glm-4.6`; Codex unavailable). live endpoint probes on Cosmos Hub REST, Terra LCD, THORChain API. no emulator run (no UI change).

NeOMakinG (approved):
triple-lane (Claude + GLM) on new commit `06550ad5` — 7/7 balance-path tests confirmed, all fund-safety invariants hold. APPROVE.

## testing approach

static diff walk on `packages/core/chain/coin/balance/resolvers/cosmos.ts` + `cosmos.test.ts` at HEAD `fa0645d5`; read `cosmosRpcUrl` registry; live-curled THORChain + MayaChain LCD endpoints to verify non-standard-path concern; traced all four balance-path branches.

## prior issues — status

| r1 | 1 | should-fix | missing changeset entry for cosmos LCD fix | ✅ fixed | `.changeset/cosmos-lcd-fallback.md` added in HEAD, covers both account-info and balance paths |

## blocking

none.

## preferably-blocking

none.

## should-fix

none.

## suggestion

The `fetchBalanceViaLcd` comment mentions "height skew" as a known limitation in the inline doc — worth one more sentence clarifying the skew is acceptable because this function is display-only and never used as a pre-broadcast sufficient-balance gate. If the function ever gets reused for tx validation the skew becomes load-bearing. Low urgency, but a guard comment now is cheap insurance.

## q

none.

## risk

Low. LCD fallback is additive: RPC=positive path is unchanged; RPC=0 path adds one extra REST query that fails closed (returns 0n if LCD errors or returns 0). No regression surface for genuinely-zero wallets. All 10 CosmosChain LCD URLs are live standard Cosmos SDK REST endpoints (THORChain/MayaChain verified live).

## verdict

APPROVE. The balance LCD fallback mirrors the account-info pattern approved in r1. Trigger is strict (`=== 0n` BigInt equality). LCD failure → null → falls through to RPC value (correct fail-closed for display). LCD=0 affirmatively trusted as "truly empty" (`lcdAmount > 0n` guard prevents spurious override). `fetchBalanceViaLcd` wraps `BigInt()` inside a try/catch so any malformed LCD response returns null safely. Seven tests cover all four required paths: RPC=positive (no LCD call), RPC=0+LCD=positive, RPC=0+LCD=0, RPC=0+LCD=throw, plus factory denom URL encoding.

---

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with GLM `glm-4.6` third-eye lane. Apple M-series. no runtime screenshots (code-only re-review per `feedback_rereview_skip_retest`).

PR #588, opened 2026-05-28
Fix SwapKit BTC PSBT keysign payloads

rcoderdev:
## Summary by CodeRabbit

* **New Features**
  * Enabled use of pre-built SwapKit Bitcoin PSBT transactions and routing through the Bitcoin signing path for swap flows.

* **Bug Fixes**
  * Adjusted swap quoting so build-disabled flag is not set for Bitcoin transfer routes, preventing unnecessary build suppression.
  * Made PSBT compilation and signing handling more robust and deterministic.

* **Tests**
  * Added/updated tests to verify PSBT-based swap flows, compilation output, and deterministic pre-signing hash ordering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/588?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/mpc/keysign/swap/build.ts
+++ b/packages/core/mpc/keysign/swap/build.ts
@@ -1,7 +1,9 @@
 import { Buffer } from 'buffer'
 import { create } from '@bufbuild/protobuf'
+import { buildSignBitcoinFromPsbt } from '@vultisig/core-chain/chains/utxo/tx/buildSignBitcoinFromPsbt'
 import { fromChainAmount } from '@vultisig/core-chain/amount/fromChainAmount'
 import { toChainAmount } from '@vultisig/core-chain/amount/toChainAmount'
+import { Chain } from '@vultisig/core-chain/Chain'
 import { isChainOfKind } from '@vultisig/core-chain/ChainKind'
 import { getErc20Allowance } from '@vultisig/core-chain/chains/evm/erc20/getErc20Allowance'
 import { AccountCoin } from '@vultisig/core-chain/coin/AccountCoin'
@@ -30,6 +32,7 @@ import { SwapKitSwapPayloadSchema } from '@vultisig/core-mpc/types/vultisig/keys
 import { matchRecordUnion } from '@vultisig/lib-utils/matchRecordUnion'
 import { WalletCore } from '@trustwallet/wallet-core'
 import { PublicKey } from '@trustwallet/wallet-core/dist/src/wallet-core'
+import { Psbt } from 'bitcoinjs-lib'
 
 export type BuildSwapKeysignPayloadInput = {
   fromCoin: AccountCoin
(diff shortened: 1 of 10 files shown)


johnnyluo (approved):
LGTM

Ehsan-saradar (commented):
Thanks for this — the PSBT path is clean, and reusing `buildSignBitcoinFromPsbt` (with its fee-snipe checks) is the right call. Two things before merge: one I'd like fixed, one small. Left both inline.

**1. Verify the PSBT actually pays the quoted address + amount.** Right now we sign whatever PSBT SwapKit sends back, with no check that the deposit output matches the quoted `to`/amount. Android just added exactly this guard (#4488 / #4603 — `verifyOwnership`). Details on the `build.ts` comment.

**2. Small: the hash sort looks redundant and its comment is off.** The cosigner already sorts before signing, and iOS returns hashes in input order (not sorted). Details on the `preSigningHashes` comment.

Otherwise nice work — the PSBT flow tests are solid.

packages/core/mpc/keysign/swap/build.ts
```diff
@@ -46,6 +49,31 @@ export type BuildSwapKeysignPayloadInput = {
...
+    throw new Error('SwapKit Bitcoin transfer routes must include PSBT txType and txPayload.')
+  }
+
+  if (!transfer.txPayload?.length) {
+    throw new Error('SwapKit Bitcoin PSBT payload is empty.')
+  }
+
+  return {
+    case: 'signBitcoin',
+    value: buildSignBitcoinFromPsbt({
```
Ehsan-saradar, line 70: This is where the SwapKit PSBT turns into a signing payload, but nothing checks that it actually pays the quoted deposit address + amount. `buildSignBitcoinFromPsbt` guards fees and tags the change output, yet the destination output isn't tied back to the quote's `to`/amount — so a wrong or tampered-with response would still get signed. Android added exactly this guard recently (#4488 / #4603 — `verifyOwnership` ties `signBitcoin.outputs` back to the expected address/amount before signing). Could we verify the deposit output here before building the payload?

packages/core/mpc/tx/preSigningHashes/index.ts
```diff
@@ -18,10 +19,15 @@ type Input = {
...
+  [...hashes].sort((one, another) => Buffer.from(one).compare(Buffer.from(another)))
+
 export const getPreSigningHashes = ({ walletCore, txInputData, chain, keysignPayload }: Input) => {
+  const signBitcoin = keysignPayload ? getSwapKitSignBitcoin(keysignPayload) : undefined
+
   // PSBT signing: compute BIP-143 sighashes directly from structured data
-  if (keysignPayload?.signData.case === 'signBitcoin') {
-    return computePreSigningHashes(keysignPayload.signData.value)
+  if (signBitcoin) {
+    return sortHashes(computePreSigningHashes(signBitcoin))
```
Ehsan-saradar, line 30: This sort looks redundant: the cosigner already sorts every hash before signing (`keysign/cosigner.ts`), and `compileSignBitcoinTx` looks signatures up by hash, not by position — so ordering is already handled. Also, iOS's `preSigningHashes` returns hashes *in input order*, not sorted, so "sorted to match iOS co-signing" isn't quite right. It's harmless, but I'd drop the sort or fix the comment to explain the real reason.

packages/core/mpc/keysign/swap/build.ts
rcoderdev: Sorry, I noticed a comment after merging. I will open a follow-up PR.

PR #590, opened 2026-05-28
Guard SwapKit BTC PSBT destination outputs

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * SwapKit Bitcoin PSBTs are now validated: destination address and amount must match the quote; PSBTs with unexpected value-bearing outputs or mismatched sums are rejected.

* **Tests**
  * Expanded coverage for PSBT scenarios, including matching/mismatching addresses, amounts, and hidden extra outputs.

* **Behavior**
  * SwapKit Bitcoin PSBT payloads are validated and processed through the Bitcoin signing/verification flow before signing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/590?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/mpc/tx/swapkitSignBitcoin.ts
+++ b/packages/core/mpc/tx/swapkitSignBitcoin.ts
@@ -1,35 +1,149 @@
 import { Buffer } from 'buffer'
 import { Chain } from '@vultisig/core-chain/Chain'
 import { buildSignBitcoinFromPsbt } from '@vultisig/core-chain/chains/utxo/tx/buildSignBitcoinFromPsbt'
-import { Psbt } from 'bitcoinjs-lib'
+import { address as btcAddress, networks, Psbt } from 'bitcoinjs-lib'
 
 import { KeysignPayload } from '../types/vultisig/keysign/v1/keysign_message_pb'
+import { SwapKitSwapPayload } from '../types/vultisig/keysign/v1/swapkit_swap_payload_pb'
 import { SignBitcoin } from '../types/vultisig/keysign/v1/wasm_execute_contract_payload_pb'
 
-export const getSwapKitSignBitcoin = (keysignPayload: KeysignPayload): SignBitcoin | undefined => {
-  if (keysignPayload.signData.case === 'signBitcoin') {
-    return keysignPayload.signData.value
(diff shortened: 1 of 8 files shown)


johnnyluo (approved):
LGTM

packages/core/mpc/tx/swapkitSignBitcoin.ts
```diff
@@ -1,35 +1,149 @@
...
+        `SwapKit Bitcoin PSBT output #${index} has isChange=${output.isChange}, ` +
+          `but its scriptPubKey ${derivedIsChange ? 'does' : 'does not'} pay back to the sender.`
+      )
+    }
   })
+
+  const nonChangeOutputs = outputs.filter(({ output }) => !output.isChange)
+  const nonChangeAmount = nonChangeOutputs.reduce((sum, { output }) => sum + output.amount, 0n)
+
+  if (nonChangeAmount !== expectedAmount) {
```
Ehsan-saradar, line 81: One worry here: this wants the PSBT output to match `transfer.amount` down to the exact sat. But `transfer.amount` isn't read from the PSBT — it's SwapKit's `depositAmount` (a decimal turned into sats) or the user's input amount as a fallback. So if they differ by even 1 sat (rounding or fees), a real swap gets rejected here.

Have you tried this against a real Chainflip PSBT to confirm they match? If exact match turns out too strict, the address check above already blocks theft on its own, so a small tolerance here would be safe.

rcoderdev: Good catch. I kept the exact PSBT binding, but changed the upstream amount source instead of adding tolerance: for BTC PSBT SwapKit routes we now derive transfer.amount from the PSBT destination output itself, then the existing exact guard verifies against that displayed/signed amount. That avoids false rejects if SwapKit depositAmount metadata drifts by a sat, and it stays aligned with Android’s exact non-change-output binding while still preventing underpayment/fee-burn surprises. Added a regression where depositAmount is 100,000 sats but the PSBT output is 99,999 sats.

Ehsan-saradar (approved):

PR #589, opened 2026-05-28
chore: version packages

rcoderdev:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

-   [#588](https://github.com/vultisig/vultisig-sdk/pull/588) [`256f67d`](https://github.com/vultisig/vultisig-sdk/commit/256f67da13a6d96f34c83c9b56c1cfb574cd8fd1) Thanks [@rcoderdev](https://github.com/rcoderdev)! - Enable pre-built SwapKit Bitcoin PSBT transactions, verify their destination outputs, and route payloads through the SignBitcoin hashing and compilation path.

### Patch Changes

-   [#588](https://github.com/vultisig/vultisig-sdk/pull/588) [`256f67d`](https://github.com/vultisig/vultisig-sdk/commit/256f67da13a6d96f34c83c9b56c1cfb574cd8fd1) Thanks [@rcoderdev](https://github.com/rcoderdev)! - Enable pre-built SwapKit Bitcoin PSBT transactions, verify their destination outputs, and route payloads through the SignBitcoin hashing and compilation path.

-   Updated dependencies \[[`256f67d`](https://github.com/vultisig/vultisig-sdk/commit/256f67da13a6d96f34c83c9b56c1cfb574cd8fd1)]:
    -   @vultisig/core-chain@2.3.2

## Summary by CodeRabbit

* **Bug Fixes**
  * Pre-built Bitcoin PSBT transactions now enabled with destination output verification
  * Improved routing payload processing through enhanced signing and compilation

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/589?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #581, opened 2026-05-27
chore: version packages

rcoderdev:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

-   [#579](https://github.com/vultisig/vultisig-sdk/pull/579) [`c3881e5`](https://github.com/vultisig/vultisig-sdk/commit/c3881e549e5678e8806eba5defb2d2d6eefc2cc5) Thanks [@gomesalexandre](https://github.com/gomesalexandre)! - ## Fixed
    -   Cosmos account info LCD fallback for extended account types that StargateClient cannot decode (vesting wrappers, module accounts) — prevents doomed txs with `sequence:0` that fail at broadcast with `account sequence mismatch, expected N, got 0` ([#579](https://github.com/vultisig/vultisig-sdk/issues/579))
    -   Cosmos coin balance LCD fallback when StargateClient returns `amount:"0"` on a funded address — fixes a packaging-level discrepancy in cosmjs's HTTP layer under Hermes/React Native that silently surfaced as "you have 0" on funded Terra/TerraClassic wallets ([#579](https://github.com/vultisig/vultisig-sdk/issues/579))

### Patch Changes

-   Updated dependencies \[[`c3881e5`](https://github.com/vultisig/vultisig-sdk/commit/c3881e549e5678e8806eba5defb2d2d6eefc2cc5)]:
    -   @vultisig/core-chain@2.3.1

### Patch Changes

-   [#579](https://github.com/vultisig/vultisig-sdk/pull/579) [`c3881e5`](https://github.com/vultisig/vultisig-sdk/commit/c3881e549e5678e8806eba5defb2d2d6eefc2cc5) Thanks [@gomesalexandre](https://github.com/gomesalexandre)! - ## Fixed
    -   Cosmos account info LCD fallback for extended account types that StargateClient cannot decode (vesting wrappers, module accounts) — prevents doomed txs with `sequence:0` that fail at broadcast with `account sequence mismatch, expected N, got 0` ([#579](https://github.com/vultisig/vultisig-sdk/issues/579))
[...]

PR #594, opened 2026-05-28
fix: release SDK after core package updates

rcoderdev:
## Summary
- add a missing changeset so the SwapKit Bitcoin PSBT fix is published in a fresh `@vultisig/sdk` release (`1.2.2`) and linked CLI release
- only create the SDK GitHub tag/release when `@vultisig/sdk` itself is newly published
- wait for npm publish to succeed before creating GitHub tags/releases or running post-publish side jobs

## Root cause
PR #588 changed `@vultisig/core-chain` and `@vultisig/core-mpc`, and PR #589 published those two packages. It did not publish `@vultisig/sdk`, because the changeset only listed the core packages. The release workflow still tried to create a GitHub tag from the SDK version (`v1.2.1`), which already existed from PR #581, so the workflow failed even though the core package npm publish completed.

The npm page screenshot is correct: `@vultisig/sdk@1.2.1` was last published on 2026-05-27, and it does not contain the new SwapKit PSBT code path. This PR queues the missing SDK release.

## Verification
- `npm view @vultisig/sdk@1.2.1 time dist-tags --json` confirms latest SDK is still `1.2.1` from 2026-05-27
- downloaded `@vultisig/sdk@1.2.1` tarball and confirmed `verifySwapKitBitcoinPsbtOutputs`, `swapkitSignBitcoin`, and `buildSignBitcoinFromPsbt` are absent from SDK dist
- `npm view @vultisig/core-chain@2.3.2` and `npm view @vultisig/core-mpc@1.2.23` confirm those core packages were published by the failed run
- `yarn changeset status --verbose` shows pending patch bumps for `@vultisig/sdk@1.2.2` and `@vultisig/cli@1.2.2`
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/release.yml`
- `git diff --check`

## Release Notes

* **Bug Fixes**
  * Updated SDK package with improved Bitcoin PSBT signing functionality using the latest SwapKit signing implementation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-sdk/pull/594?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/brave-sdk-psbt-release.md
+++ b/.changeset/brave-sdk-psbt-release.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Publish the SDK bundle with the latest SwapKit Bitcoin PSBT signing path from `@vultisig/core-chain` and `@vultisig/core-mpc`.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,7 @@ jobs:
     outputs:
       version: ${{ steps.version.outputs.version }}
       should_publish: ${{ steps.version.outputs.should_publish }}
+      should_tag_release: ${{ steps.version.outputs.should_tag_release }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v4
 
@@ -73,6 +74,7 @@ jobs:
           )
 
           SHOULD_PUBLISH=false
+          SHOULD_TAG_RELEASE=false
 
           for PKG_PATH in "${PACKAGES[@]}"; do
             PKG_NAME=$(node -p "require('./${PKG_PATH}/package.json').name")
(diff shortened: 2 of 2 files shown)


gomesalexandre (approved):
Root cause + fix both look right ser.

The scoping logic is correct - `SHOULD_TAG_RELEASE` only flips to `true` when `@vultisig/sdk` has an unpublished version, while `SHOULD_PUBLISH` remains the gate for all other downstream jobs (Vercel, docs, discord). Multi-package release with no SDK version bump → publishes the sub-packages but skips the `v{version}` tag. Exactly what we want.

The `git ls-remote --exit-code --tags origin` change for checking existing tags is better than the old `git rev-parse` - it checks the remote authoritative state instead of the local clone, so a partial previous run can't leave us stuck.

`needs: publish` on `tag-and-release` is the right gating - ensures the tag only lands after npm is live, no more orphan GitHub releases on publish failures.

`continue-on-error: true` on Vercel/Discord/docs sidecars with warning echoes is the right shape - failures in those shouldn't roll back a published npm package. 

One q/nit for future consideration (not blocking): `SHOULD_TAG_RELEASE` is never reset to false inside the loop, so if somehow the SDK package entry appears more than once or re-enters, it stays true. But that can't happen with the current loop structure over `PACKAGES[@]` so it's fine today.

LGTM, does what it says on the box.

PR #596, opened 2026-05-28
fix(chain/swap): remove i18next dep — fixes 'reload of undefined' broadcast crash

gomesalexandre:
Fixes [vultiagent-app#1013](https://github.com/vultisig/vultiagent-app/issues/1013).

## tl;dr

Found live during the USTC→LUNC sim drive ([mcp-ts#260](https://github.com/vultisig/mcp-ts/pull/260) testing): the swap card renders, password unlocks, MPC sign succeeds, then broadcast crashes with `Cannot read property 'reload' of undefined`. Root cause: `getNativeSwapQuote.ts:92` calls `t('not_enough_asset_to_cover_gas_fees', { asset })` from i18next, but the consuming app never initializes i18next with a Backend plugin. Calling `t()` lazily kicks off `reloadResources()` → `services.backendConnector.reload()` → undefined deref → crash.

This was the ENTIRE i18next surface in `@vultisig/core-chain`. One call site, one crash, one fix.

## What

- Replace `t('not_enough_asset_to_cover_gas_fees', { asset: ticker })` with `` `Not enough ${ticker} to cover gas fees.` `` (template literal — same UX, no i18n)
- Drop the i18next import
- Remove i18next from `@vultisig/core-chain/package.json` dependencies
- Drop the corresponding `vi.mock('i18next', ...)` from the test (no longer needed)

## Why fix at SDK layer (not app layer)

The SDK should not impose an i18n contract on every consumer. If a future caller wants localized error copy, they can wrap the SDK error themselves at the app boundary. Plain English error strings are the right SDK-layer default. Localization is an app concern, not a chain library concern.

## Test plan

```
yarn vitest run packages/core/chain/swap/native/api/getNativeSwapQuote.test.ts
Tests 6 passed (6)
```

## receipts

The original crash was captured live during the USTC→LUNC sim drive — visible in the [mcp-ts#260 PR body screenshot](https://files.catbox.moe/83ij3z.png) (post-sign agent narrates `The action failed: Cannot read property 'reload' of undefined`).

After this lands, the i18n crash vector is gone — broadcast can proceed past the post-sign callback chain.

🤖 reviewed via Claude Code (root-caused via live e2e USTC→LUNC sim drive)


NeOMakinG (approved):
## testing approach

triple-lane: Claude diff-walk, GLM (run below), Codex skipped per auth-brittleness.

## blocking

none.

## risk

**LOW.** The only behavioral change is the error message string for the 'not enough gas' path: `t('not_enough_asset_to_cover_gas_fees', {asset})` → ``Not enough ${ticker} to cover gas fees.``. The previous string was an i18next key (never translated in a no-i18next consumer), so apps saw the raw key string on error. The new string is strictly better UX.

Root cause well-diagnosed: i18next's `t()` calls `this.services.languageUtils` which requires `.init()` to have run; consumers that skip i18next initialization crash at `services.languageUtils.reload`. Dropping the dep eliminates the crash class entirely.

## verdict

APPROVE. Clean removal, CI green across all 4 gates.

---
reviewed via [Claude Code](https://claude.com/claude-code) (sonnet-4-6)

PR #606, opened 2026-05-30
feat(notifications): optional appId on push registration for per-app routing

gomesalexandre:
## what

Companion to [vultisig/notification#27](https://github.com/vultisig/notification/pull/27) (per-app notification routing) and the agent-backend notifier `app_id`. For the notification service to deliver a Station push to Station (instead of the wallet that shares the vault), the Station device must be **registered** under its own `app_id` (`money.terra.station`). This adds that capability to the SDK push service.

## how

- `RegisterDeviceOptions.appId` (optional) → sent as `app_id` on `/register`, and **persisted** in the local registration record.
- `unregisterVault()` sends the persisted `app_id` so the tokenless DELETE scopes to the same app (the server defaults a missing `app_id` to the wallet bucket, so a non-wallet app must send it).
- `isVaultRegistered(vaultId, appId?)` is **migration-aware**: when the caller passes the app it intends to register under, a local record stored under a different/missing `appId` counts as not-registered, so the consumer's `isVaultRegistered`-gated `registerDevice` re-runs and migrates existing opted-in devices onto their `app_id`.
- `notifyVaultMembers` gains an optional `appId` but it is **opt-in only** (see below).

All fields are optional and omitted by default → the regular wallet sends no `app_id`, the server keeps its wallet-default routing, and existing registrations are unchanged.

## design decisions (4 codex rounds)

- **`notifyVaultMembers` app_id is opt-in, NOT inferred from storage.** This is the keysign-coordination path: it must reach every device on a vault regardless of which app each peer registered under. Auto-scoping to the sender's app would make a mixed-version vault (one signer migrated, one still wallet-default) miss keysign pushes. The app-scoped delivery the leak fix needs is the **scheduled** path, which goes through the agent-backend notifier — not this method. (Codex oscillated on this; this is the deliberate resolution.)
- **No old-bucket orphan bookkeeping on migration.** The real notification service (#27) upserts `/register` on `(vault_id, party_name, token)` and updates `app_id` in place for the same token, so migrating wallet→station mutates the one row rather than orphaning a wallet-bucket row; a rotated token's stale row is 410-cleaned. No separate old-scope cleanup needed.

## rollout

Coordinated: this SDK change ships first, the Station app then passes `appId: "money.terra.station"` to `registerDevice`/`isVaultRegistered` (one-line consumer change), existing installs re-register on next launch (migration-aware gate), Station's APNs cert is provisioned, then `agent-backend` sets `NOTIFICATION_APP_ID`. Until the consumer passes `appId`, behaviour is identical to today.

## changeset

`@vultisig/sdk` minor (`.changeset/push-app-id-routing.md`).

## risk

Low. All additions are optional and wallet-default-preserving; the wallet path is byte-identical (no `app_id` sent). Note: local `yarn typecheck` not run in this worktree (yarn-workspace `node_modules` not installed in a fresh worktree); changes are additive optional fields + conditional spreads + one optional param — CI runs the full typecheck.

### DIFF
--- a/.changeset/push-app-id-routing.md
+++ b/.changeset/push-app-id-routing.md
@@ -1,0 +1,10 @@
+---
+'@vultisig/sdk': minor
+---
+
+Add optional `appId` to push device registration (`registerDevice`). Apps that
+share a vault with the regular wallet (e.g. Station, `money.terra.station`) can
+now register/unregister under their own bundle id, so the notification service
+routes their pushes to the correct app instead of the wallet that shares the
+vault. The field is optional and omitted by default, so existing wallet
+registrations are unchanged.
--- a/packages/sdk/src/services/PushNotificationService.ts
+++ b/packages/sdk/src/services/PushNotificationService.ts
@@ -75,6 +75,9 @@ export class PushNotificationService {
         party_name: opts.partyName,
         token: opts.token,
         device_type: toServerDeviceType(opts.deviceType),
+        // Only send app_id when supplied; omitting it preserves the server's
+        // wallet-default routing for existing (wallet) callers.
+        ...(opts.appId ? { app_id: opts.appId } : {}),
       }),
     })
 
@@ -82,12 +85,13 @@ export class PushNotificationService {
       throw new Error(`Failed to register device: ${response.status} ${response.statusText}`)
     }
 
-    // Persist locally
+    // Persist locally (incl. appId so unregister can scope to the same app)
     const registrations = await this.getRegistrations()
     registrations[opts.vaultId] = {
       vaultId: opts.vaultId,
       partyName: opts.partyName,
       registeredAt: Date.now(),
+      ...(opts.appId ? { appId: opts.appId } : {}),
     }
     await this.storage.set(STORAGE_KEY, registrations)
   }
@@ -110,6 +114,11 @@ export class PushNotificationService {
         body: JSON.stringify({
           vault_id: vaultId,
           party_name: reg.partyName,
+          // Scope the tokenless delete to the app this device registered under.
+          // The server defaults a missing app_id to the wallet bucket, so a
+          // non-wallet app (e.g. Station) must send it or its row is never
+          // removed. Wallet registrations have no appId and keep the default.
+          ...(reg.appId ? { app_id: reg.appId } : {}),
         }),
       })
       if (!response.ok) {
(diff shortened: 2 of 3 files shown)


NeOMakinG (approved):
46/46 tests pass (`npx vitest run PushNotificationService.test.ts`), typecheck clean, diff is +58/-3. Three-lane: Claude (primary, cross-repo schema verification + adversarial), Codex (CLI unavailable), GLM `glm-4.6` (two reported blockers, both downgraded on investigation). **APPROVE** with one should-fix on test coverage.

## testing approach

SDK checkout at HEAD `7c44cd15` on `/Volumes/External/vultisig/vultisig-sdk`. `npx vitest run packages/sdk/tests/unit/services/PushNotificationService.test.ts --reporter=verbose` — 46/46. `npx tsc --noEmit --project packages/sdk/tsconfig.json` — clean. Three-way cross-repo schema audit: notification#27 `app_id string json:"app_id"`, ab#862 `AppID string json:"app_id,omitempty"`, sdk#606 wire body `{ app_id: opts.appId }` — all match snake_case on the wire, camelCase stays TypeScript-internal. No `"undefined"` string can reach the wire; truthiness guard on `opts.appId` means `undefined`/`null`/`""` all produce the same omit-from-payload outcome (correct wallet-default behavior). Local storage write uses the same truthiness guard so `appId` is absent (not `""`) on wallet registrations — no false `isVaultRegistered` results from stored empty strings.

## should-fix

1. **All new `appId` branches are untested.** The PR adds four distinct behavioral code paths — `registerDevice` with `appId` (sends `app_id` in POST body + stores `appId`), `unregisterVault` reading stored `appId` to scope the DELETE body, `isVaultRegistered(vaultId, appId)` migration-aware comparison, and `notifyVaultMembers` with `appId` — but `PushNotificationService.test.ts` has zero coverage for any of them. All 46 existing tests take only the no-appId branch. A future refactor of the spread-or-omit pattern would silently break routing with no failing test.

   fix: add a `describe('appId routing')` block with at minimum:
   - `registerDevice` with `appId: 'money.terra.station'` — assert POST body has `app_id`, stored record has `appId`
   - `unregisterVault` after station registration — assert DELETE body has `app_id: 'money.terra.station'`
   - `isVaultRegistered(vaultId, 'money.terra.station')` on wallet-registered device → `false` (migration trigger)
   - `isVaultRegistered(vaultId, 'money.terra.station')` on station-registered device → `true`
   - `notifyVaultMembers({ ..., appId: 'money.terra.station' })` — assert POST body has `app_id`

## q

1. `notifyVaultMembers` gains an `appId` option but the PR description says this path must always fan out to all apps for keysign. Is there a planned consumer that will pass `appId` here? If not, a JSDoc note ("reserved for future use; omit for standard keysign coordination") would prevent callers from accidentally scoping keysign pushes.

## risk

Wallet path is byte-identical — no `app_id` sent, no `appId` stored, `isVaultRegistered` behavior unchanged. All `appId` branches are additive and gated by explicit opt-in. Low blast radius.

## verdict

APPROVE — cross-repo schema verified clean (3-way: notification#27 / ab#862 / sdk#606 all use `app_id` snake_case on the wire), backward compat solid. GLM blockers are false positives: (1) empty-string omit is correct by design — bundle IDs are non-empty; (2) stored records cannot acquire `appId: ""` because the write-path uses the same truthiness guard. The should-fix (test coverage) is real but not gating given the staged rollout — Station/vultiagent-app haven't wired `appId` yet, so untested branches won't reach production until a consumer PR lands.

---
reviewed via Claude Code with GLM `glm-4.6` third-eye lane. Codex lane skipped (CLI auth not available). Apple M-series. SDK worktree at `7c44cd15`.

PR #608, opened 2026-05-31
fix(polkadot): close header race causing BadProof on Asset Hub sends

### LINKED ISSUE #609: fix(polkadot): close header race causing BadProof; surface broadcast errors; handle peer-race duplicates
## Problem

Polkadot sends on the live Asset Hub runtime (spec 2002001) fail in three related ways. All three are reproducible on a multi-device vault doing a small DOT send.

### 1. `BadProof` on the initiator's broadcast (root cause)

`author_submitExtrinsic` is rejected:

```json
{
  "error": {
    "code": 1010,
    "message": "Invalid Transaction",
    "data": "Transaction has a bad signature"
  }
}
```

`getPolkadotChainSpecific` (`packages/core/mpc/keysign/chainSpecific/resolvers/polkadot/index.ts`) fetched `recentBlockHash` and `currentBlockNumber` in **two sequential RPC calls**. The chain head can advance between them. When it does, the era encodes `phase = (N+1) % period` but `additional_signed.blockHash` still holds the hash of block N. The runtime recomputes the era checkpoint at validation, finds the wrong hash, and rejects with `InvalidTransaction::BadProof`. Substrate then bans the extrinsic hash for ~30s, so every retry of the same bytes returns ``Transaction is temporarily banned``.

Non-deterministic — depends on whether the head advances between the two RPC calls. On Asset Hub the race fires often enough to be user-visible.

### 2. Broadcast failures silently reported as success

When `author_submitExtrinsic` failed, `broadcastPolkadotTx` fell through to `verifyBroadcastByHash`. That safety net asks `getPolkadotTxStatus` whether the hash is known to the chain. When the response couldn't confirm the hash, the resolver returned bare `{ status: 'pending' }` with `isKnown` undefined. `verifyBroadcastByHash` read that as ``status === 'pending' && isKnown !== false`` and treated it as a successful broadcast — the UI then showed a "Done" screen with a hash that had no on-chain counterpart.

### 3. Joiner sees a misleading "Signing Error" after the initiator's successful broadcast

In an MPC ceremony every device broadcasts the same signed extrinsic. The slower one hits substrate Pool errors — ``Transaction Already Imported``, ``Already known``, or ``Transaction is temporarily banned`` — because the faster peer's bytes are already in the pool. The tx is genuinely on chain via the initiator's broadcast, but the joiner's UI surfaced the rejection as a `Signing Error`.

Mirrors what mobile saw in [vultisig-android#4544](https://github.com/vultisig/vultisig-android/issues/4544).

### Bonus: every distinct failure looked identical in the UI

Substrate's JSON-RPC error shape is `{ code, message, data? }`. For `InvalidTransaction::*` (code 1010) the `message` is always the generic ``"Invalid Transaction"`` and the actually useful reason (``"Transaction has a bad signature"``, ``"Transaction is outdated"``, ``"Inability to pay some fees"``, …) lives in `data`. `broadcastPolkadotTx` surfaced only `message`, collapsing every distinct failure into a single uninformative ``"Polkadot broadcast failed: Invalid Transaction"``.

## Repro

1. Multi-device Polkadot vault.
2. Initiator sends a small DOT amount; joiner co-signs.
3. Observe (any combination of):
   - Initiator: ``Polkadot broadcast failed: Invalid Transaction`` on the first try, ``temporarily banned`` on retries.
   - Joiner: ``Signing Error. Polkadot broadcast failed: Transaction Already Imported`` even after the tx confirms.
   - UI shows a "Done" screen for a tx that's not actually on chain.

## Resolution

Fixed in #608. Three commits, each independently reviewable:

1. **Header race fix** (`chainSpecific/resolvers/polkadot/index.ts`) — read the header once and derive both `recentBlockHash` and `currentBlockNumber` from the same response. Batch the runtime version, nonce, and genesis-hash fetches into a single `Promise.all` while we're here.
2. **`tx/status` resolver no longer masks broadcast errors** (`tx/status/resolvers/polkadot.ts`) — split the early-return into `{ status: 'pending', isKnown: false }` when the hash can't be confirmed vs `{ status: 'pending', isKnown: true }` when it's in flight but not finalized. Mirrors `ripple.ts:25` / `solana.ts:19`. + 8 tests.
3. **`tx/broadcast` resolver** (`tx/broadcast/resolvers/polkadot.ts`) — pattern-matches the duplicate-broadcast Pool errors (``Already Imported`` / ``Already known`` / ``temporarily banned``, case-insensitive, against both `message` and `data`) and returns silently. Surfaces substrate's `data` field in error messages so ``"Polkadot broadcast failed: Invalid Transaction"`` becomes e.g. ``"Polkadot broadcast failed: Invalid Transaction: Transaction is outdated"``. + 11 tests.

## Acceptance criteria

- [x] Initiator's broadcast no longer hits `BadProof` from the header race.
- [x] `verifyBroadcastByHash` no longer reports broadcast failures as success.
- [x] Joiner no longer sees a misleading Signing Error after the initiator's successful broadcast.
- [x] `Invalid Transaction` errors include the substrate-side reason in the surfaced message.
- [x] `yarn test:core` — 874/874 passing (was 870; +19 new tests in this PR).
- [x] Typecheck clean.
- [x] Manual: send multiple different DOT amounts back-to-back from a multi-device vault — each succeeds (was reproducibly failing before the race fix).

Ehsan-saradar:
Closes #609.

## Summary

Three independent fixes to the Polkadot send path on the live Asset Hub runtime (spec 2002001). Each is in its own commit; the race fix is the root cause and the other two are guard-rails around it.

### 1. Header race in `getPolkadotChainSpecific` (root cause)

`recentBlockHash` and `currentBlockNumber` were fetched in **two sequential RPC calls**. The chain head can advance between them. When it does the era encodes `phase = (N+1) % period` but `additional_signed.blockHash` still holds the hash of block N. The runtime recomputes the era checkpoint at validation, finds the wrong hash, and rejects with `InvalidTransaction::BadProof`. Substrate then bans the extrinsic hash for ~30s, so every retry comes back as ``Transaction is temporarily banned``. On Asset Hub the race fires often enough to be user-visible.

Read the header once and derive both fields from the same response. Batch the runtime version, nonce, and genesis-hash fetches into a single `Promise.all` while we're here.

### 2. `getPolkadotTxStatus` was masking broadcast errors

When Subscan responds with `code !== 0`, the resolver returned bare `{ status: 'pending' }`. `verifyBroadcastByHash` reads that as ``status === 'pending' && isKnown !== false`` and treats it as success, silently swallowing the real `author_submitExtrinsic` rejection. The UI then shows a "Done" screen with a hash that has no on-chain counterpart.

Split the early-return into two branches — `isKnown: false` when Subscan doesn't know the hash, `isKnown: true` when it's indexed but not finalized yet. Mirrors `ripple.ts:25` / `solana.ts:19`.

### 3. `broadcastPolkadotTx` two small fixes

- **Swallow peer-race duplicates.** In an MPC ceremony every device broadcasts the same signed extrinsic. The slower device hits ``Transaction Already Imported``, ``Already known``, or ``Transaction is temporarily banned``. Pattern-match those three strings (case-insensitive, against both `message` and `data`) and return silently — the tx is in flight via the faster peer; surfacing the rejection puts the co-signer on a Signing Error screen for a tx that did confirm.
- **Surface the `data` field.** Substrate's JSON-RPC error shape is `{ code, message, data? }`. For `InvalidTransaction::*` (code 1010) the `message` is the generic ``"Invalid Transaction"`` and the actually useful reason (``"Transaction has a bad signature"``, ``"Transaction is outdated"``, ``"Inability to pay some fees"``, …) lives in `data`. Previously we showed only `message`, collapsing every distinct failure into a single uninformative string.

## Test plan

- [x] Vitest: `yarn test:core` — **874/874 passing** (was 870; +19 new tests across the two new `polkadot.test.ts` files).
- [x] Typecheck: `tsc --project .config/tsconfig.json --noEmit` clean.
- [x] Manual: Polkadot 1 DOT send to a second account on a multi-device vault — initiator broadcasts and lands on chain, joiner no longer shows a misleading Signing Error.
- [x] Manual: send multiple different amounts back to back without restarting — each one succeeds (was reproducibly failing before the race fix).


gomesalexandre (approved):
## lgtm 🫡

took this over for a deep pass - here's what i verified:

**header race fix (root cause):** `getHeader()` returns the current head; `header.hash` and `header.number` are atomically consistent. old code called `getBlockHash()` then `getHeader()` in sequence — chain could advance between them, era checkpoint would encode block N+1 but `recentBlockHash` still held hash of N. runtime rejects with BadProof, substrate bans the extrinsic for ~30s. `Promise.all` batching eliminates the window entirely.

**isKnown gate in tx status:** `verifyBroadcastByHash` guards on `result.isKnown !== false` — `undefined` (old behavior) would pass the guard and silently swallow real broadcast rejections. now properly returns `isKnown: false` when Subscan is unaware of the hash. matches ripple.ts / solana.ts pattern.

**idempotent error patterns:** case-insensitive regex covers all three substrate Pool error strings (AlreadyImported, TemporarilyBanned, AlreadyKnown) plus the less-common InvalidTransaction.data path. `formatRpcError` surfaces the data field for 1010 rejections which actually have the diagnosis in there.

**adversarial cases checked:**
- `header.hash` is the same as `getBlockHash()` with no param (just the current head) ✓
- `isIdempotentBroadcastError` checks combined `message + data` string so InvalidTransaction.data=AlreadyKnown path is caught ✓
- `verifyBroadcastByHash` contract: `isKnown: true` (in-flight) → swallows duplicate, `isKnown: false` (unknown) → re-throws ✓
- network-level RPC errors route through `verifyBroadcastByHash` (catch block), not silently dropped ✓

**test transcript (local, branch):**
```
Test Files  92 passed (92)
Tests  870 passed (870)
Duration  3.90s
```
870 includes the new polkadot broadcast + status resolver test files (19 cases total across both). tsc --noEmit clean. lint clean.

ci: all green. no open threads. ship it.

rcoderdev (approved):

PR #617, opened 2026-06-01
Drop native swap priority for banded routing

### LINKED ISSUE #605: Swap routing: drop hard native priority, add banded provider preference (converge to iOS, SwapKit promoted)
## Goal

Converge swap routing across all core apps onto **one rule**, anchored to iOS (best user rate). Today the three platforms route differently for the same swap, which means different rate/output/revenue per platform:

| Platform | Rule today | Native priority? | SwapKit? |
|---|---|---|---|
| iOS | best **net** output | no (meritocratic) | yes |
| Android | best output (`estimatedDstFiat`) | no (meritocratic) | **not integrated** |
| **SDK (Windows + agent-app)** | **hard native priority** + best-among-aggregators | **yes** | yes |

The SDK is the outlier. This issue brings it in line and adds a thin economics band.

## Unified routing rule (canonical for all platforms)

1. Fetch quotes from all eligible providers in parallel (unchanged).
2. Comparable metric = **net output**: destination amount net of provider/integrator fees, decimal-aligned, **source-chain gas excluded** (matches iOS `SwapQuote.expectedNetToAmount`). The SDK's existing `getComparableOutputAmount` already satisfies this (aggregator `dstAmount` and native `expected_amount_out` are post-fee) — keep it.
3. `best` = max net output across **all** providers (native + aggregator — **no** hard native priority).
4. **Banded preference:** among providers whose net output `>= best * (1 - BAND)`, pick the one highest in this order:
   `THORChain > MayaChain > SwapKit > Kyber > 1inch > LiFi`
   (providers absent on a platform are skipped).
5. `BAND = 0.01` (1%) — single tunable constant.

Effect: user always gets within 1% of the best rate; on near-ties we tilt to native-moat / SwapKit economics.

## SDK changes

`packages/core/chain/swap/quote/findSwapQuote.ts`

- **Remove hard native priority** in `selectBestEligibleQuote` (lines ~160-199): today it returns `bestNative ?? bestAggregator`, so any THOR/Maya route beats every aggregator regardless of output. Replace with a **single banded ranking** over all fulfilled quotes.
- Replace `aggregatorPreferenceOrder` (`[KyberSwap, 1inch, LiFi, SwapKit]`, lines ~114-119) with a **full provider priority order**: `[THORChain, MayaChain, SwapKit, KyberSwap, 1inch, LiFi]`. Note SwapKit moves from last to 3rd (above the EVM aggregators), per revenue/coverage.
- Selection: `best = max(netOutput)`; among quotes with `netOutput >= best * 0.99`, pick the lowest priority-order index.
- Keep `getComparableOutputAmount` as-is for the metric (the `#353` gas-inclusion follow-up is a separate refinement; iOS also excludes source gas, so matching iOS inherits the same approximation — do **not** block on it).

## Tests

Mirror iOS's regression intent:
- same-chain ERC20->ETH: aggregator with higher net output **wins over THORChain** (the current SDK forces THOR here — this is the bug we're removing).
- near-tie (within 1%): higher-priority provider wins (e.g. SwapKit beats LiFi when within band).
- outside band (>1% better): the better-rate provider wins regardless of order.

## Context / decisions

- Reverses the `paaao, 2026-05-22` hard-native-priority rule **in the SDK only** — confirmed, to match iOS (best user rate, esp. ETH->stable).
- iOS + Android already meritocratic; this is convergence, not a new direction.
- Sibling issues: iOS (add band on top of existing best-net), Android (add band; SwapKit integration ongoing/separate).

Owner: swap (Neo).

rcoderdev:
Closes #605

## Summary by CodeRabbit

* **Bug Fixes**
  * Swap quote selection now uses a 1% provider-preference band—quotes within 1% are decided by provider preference rather than a hard native priority.
  * Tie-break outcomes updated to reflect banded rules (examples: SwapKit vs THORChain, SwapKit vs LiFi, MayaChain near-ties).
  * CowSwap is excluded from Phase 1 provider preferences.

* **Chores**
  * Patch version bumps for core packages.

### DIFF
--- a/.changeset/banded-swap-routing.md
+++ b/.changeset/banded-swap-routing.md
@@ -1,0 +1,7 @@
+---
+"@vultisig/core-chain": patch
+"@vultisig/sdk": patch
+"@vultisig/cli": patch
+---
+
+Select swap quotes with a 1% provider preference band instead of hard native priority.
--- a/packages/core/chain/swap/quote/__tests__/findSwapQuote.cowswap.test.ts
+++ b/packages/core/chain/swap/quote/__tests__/findSwapQuote.cowswap.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 
-import { aggregatorPreferenceOrder } from '../findSwapQuote'
+import { providerPreferenceOrder } from '../findSwapQuote'
 
 // Phase 2 (#471 / #584 / #3930) — CowSwap is now wired as a live fetcher in
 // `findSwapQuote`: the consumer pipeline can rebuild the order's EIP-712 digest,
@@ -9,12 +9,12 @@ import { aggregatorPreferenceOrder } from '../findSwapQuote'
 // regression is caught here, not by a user.
 
 describe('CowSwap — Phase 2 invariants (#471 / #584 / #3930)', () => {
-  it('aggregatorPreferenceOrder includes CowSwap', () => {
-    expect(aggregatorPreferenceOrder.includes('CowSwap')).toBe(true)
+  it('providerPreferenceOrder includes CowSwap', () => {
+    expect(providerPreferenceOrder.includes('CowSwap')).toBe(true)
   })
 
   it('ranks CowSwap first — MEV-protected, gas-less fills win exact-output ties', () => {
-    expect(aggregatorPreferenceOrder[0]).toBe('CowSwap')
+    expect(providerPreferenceOrder[0]).toBe('CowSwap')
   })
 
   it('cowswap module is importable and the Phase 2 helpers exist', async () => {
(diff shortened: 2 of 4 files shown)


gomesalexandre (commented):
draft review — good timing to look since this reverses the paaao directive from 2026-05-22 (hard THOR priority).

the 1% band approach is the right call. hard priority was hiding cases where a general aggregator was meaningfully better (10% better but THOR still wins is bad UX). the band preserves the preference signal without throwing away user value on large rate gaps.

fund-safety: routing-only change, user still confirms the pre-sign card. no signing/key material touched. no fund-loss vector.

two things worth confirming before undraft:
- q: is the 1% band threshold (0.01) sourced from a config or hardcoded? if hardcoded, where should it be documented as a tuneable so the next paaao directive can adjust it without a grep hunt?
- q: the old test 'prefers THORChain even when SwapKit is 10% better' documented an explicit intentional decision (paaao directive). the new test description is generic. worth keeping the directive attribution in a comment so the test doesn't silently lose its rationale.

CI green, changeset present. looks good conceptually.

reviewed via Claude Code (local sweep)

johnnyluo (approved):
LGTM

PR #518, opened 2026-05-22
[DO NOT MERGE] feat(lifi): add Station config alongside vultisig-0 (pending portal confirmation)

gomesalexandre:
**DO NOT MERGE ME** — LI.FI portal registration for Station integrator name (`station-v0` requested) is pending. Once confirmed, finalize the integrator name + flip to ready.

## what

Adds `stationLifiConfig` with `integratorName: 'station-v0'` (TBD) alongside the existing `vultisig-0` lifiConfig.

Refactors `getLifiSwapQuote` to accept an optional `lifiAffiliateConfig` that overrides the integrator in the per-call `getQuote` request. The `createConfig` global init still uses vultisig-0 — the override only affects the per-request integrator tagging.

## when to flip ready

1. LI.FI portal confirms `station-v0` (or whatever they assign)
2. Update `stationLifiConfig.integratorName`
3. Wire `lifiAffiliateConfig` through `findSwapQuote`'s `SwapAffiliateConfig.lifi` (add `.lifi` field — depends on vultisig-sdk#517)
4. Flip draft to ready

## receipts

None — this is blocked on portal confirmation. Test coverage: 544 existing tests pass unchanged.

🤖 Agent-generated

PR #618, opened 2026-06-02
feat(swap/lifi): consumer-supplied integrator + apiUrl override (closes #518)

gomesalexandre:
closes #518

## what

Supersedes vultisig/vultisig-sdk#518 (was stuck behind LI.FI portal `station-v0` integrator confirmation that never came). paaao confirmed earlier today the real Station affiliate slug is just `station` and the proxy is `https://api.vultisig.com/lifi/`, so the prior PR's placeholder is no longer load-bearing.

Adds `SwapAffiliateConfig.lifi: LifiAffiliateConfig` so consumers (Station via vultisig/mcp-ts, future tenants) can redirect LI.FI affiliate fees to their own portal integrator instead of the SDK-default `vultisig-0`.

## new surface

- `LifiAffiliateConfig` type — `{integratorName: string; apiUrl?: string}`
- `setupLifi(config?)` — idempotent global LI.FI SDK bootstrap. Consumers call this once at module boot to set both the global integrator AND the optional `apiUrl` proxy. First-caller-wins.
- `getLifiSwapQuote` accepts `lifiAffiliateConfig` and threads its `integratorName` into the per-call `getQuote` integrator override. Affiliate fee on THIS quote routes to the consumer's portal integrator without mutating the module-level `lifiConfig`.
- `findSwapQuote` threads `affiliateConfig?.lifi` to `getLifiSwapQuote`.

## consumer-side wiring (companion mcp-ts PR opening next)

\`\`\`ts
// at module boot in vultisig/mcp-ts
import { setupLifi } from '@vultisig/core-chain/swap/general/lifi/config'
setupLifi({ integratorName: 'station', apiUrl: 'https://api.vultisig.com/lifi/' })

// in stationAffiliateConfig.ts
export const STATION_AFFILIATE_CONFIG = {
  // ... existing native, oneInch, kyber
  lifi: { integratorName: 'station' },
}
\`\`\`

`STATION_AFFILIATE_CONFIG` is already threaded through `findSwapQuote` from both `list-swap-routes.ts:610` and `execute_swap.ts:1590`, so the consumer-side change is just adding the `lifi` field — no other wiring needed.

## behaviour preservation

Callers that don't supply a `lifi` config see no change. The lazy `ensureLifiConfigured` memoise routes through the existing `vultisig-0` default. Existing `getLifiSwapQuote` integration tests pass unchanged.

## tests

70/70 across the LiFi + quote test suites. 6 new in `getLifiSwapQuote.integrator.test.ts`:
- default fallback to `vultisig-0` when no config supplied
- consumer-supplied `integratorName` overrides per-call `integrator`
- per-call override does NOT mutate the global `lifiConfig`
- `setupLifi({integratorName, apiUrl})` calls `createConfig` with both fields
- `setupLifi()` with no config uses just the default integrator (no apiUrl)
- `setupLifi` is idempotent — second call is a no-op

## changeset

`@vultisig/core-chain` minor + `@vultisig/sdk` minor. New public types/functions on the swap surface — minor bump per changesets contract.

## receipts

**RECEIPT 1 — Vultisig LI.FI proxy + Station integrator work end-to-end (live curl 2026-06-02)**

```bash
curl -sS "https://api.vultisig.com/lifi/quote?\
fromChain=1&toChain=42161\
&fromToken=0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48\
&toToken=0x0000000000000000000000000000000000000000\
&fromAmount=1000000\
&fromAddress=0x649E1289fD780C2F9A3D27476511283EB0d0076D\
&toAddress=0x649E1289fD780C2F9A3D27476511283EB0d0076D\
&integrator=station\
&fee=0.003\
&slippage=0.01"
```

HTTP 200. Smoking-gun fields from the response:

```json
{
  "integrator_top": "station",
  "integratorId_action": "station",
  "integratorFee_to_station_baseUnits": 3000,
  "totalFee_baseUnits": 5500,
  "lifiFee_baseUnits": 2500,
  "station_default_wallet": "0x649E1289fD780C2F9A3D27476511283EB0d0076D",
  "matches_oneinch_kyber_receiver": true,
  "bridge_tool": "mayanFastMCTP"
}
```

`integrator` and `integratorId` both echo `station`. The LIFI Fixed Fee (5500 USDC base units = 0.0055 USDC) splits into `integratorFee: 3000` (30bps to Station's portal config) + `lifiFee: 2500` (LiFi platform cut). The `defaultWallet` matches the EXACT same `0x649E...076D` receiver wired into the existing `STATION_AFFILIATE_CONFIG.oneInch.referrer` and `kyber.referral` in mcp-ts — same money, same wallet, just a third aggregator joining the party.

This proves: (a) the proxy at `api.vultisig.com/lifi/` is alive and forwards to LI.FI cleanly, (b) the `station` integrator string is the correct portal slug, (c) Station's portal config is wired with the same fee-collection wallet already used by other aggregators. This SDK PR makes both of those reachable from the SDK call sites.

**RECEIPT 2 — unit-level wiring**

70/70 on the LiFi + quote suites. 6 new tests in `getLifiSwapQuote.integrator.test.ts` cover the per-call `integrator` override flow + `setupLifi` idempotency.

## sim QA (post-bump, in mcp-ts#323)

Sim screenshot of an end-to-end LiFi swap through the agent is the post-bump follow-up. The curl above proves the affiliate flow at the protocol layer; the mcp-ts companion PR (#323) embeds the in-app screenshot once the sdk publishes + mcp-ts is bumped.

🤖 generated alongside the orchestrator's mcp-ts companion PR

## Summary by CodeRabbit

* **New Features**
  * LI.FI affiliate support: per-swap integrator override while preserving the existing default.
  * Shared LI.FI initializer (setupLifi) with optional API endpoint and exported bootstrap config type; re-exported from the SDK and supported on React Native.
  * Swap affiliate config now accepts an optional LI.FI entry; per-call overrides won't mutate global defaults.
  * Quote fee handling preserved so affiliateBps: 0 is sent as 0.

* **Tests**
  * Added tests validating per-call overrides, lazy/global setup behavior, integrator/API wiring, fee handling, and no mutation of global defaults.

### DIFF
--- a/packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.integrator.test.ts
+++ b/packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.integrator.test.ts
@@ -127,10 +127,42 @@ describe('getLifiSwapQuote — integrator override', () => {
     expect(createConfigSpy).toHaveBeenCalledWith({ integrator: 'vultisig-0' })
   })
 
-  it('setupLifi is idempotent — second call is a no-op', () => {
+  it('lazy setupLifi() AFTER consumer setupLifi(config) is a no-op (consumer wins)', () => {
+    // The lazy path (called by ensureLifiConfigured before each getQuote)
+    // must not undo a consumer bootstrap. Once a consumer explicitly
+    // configured Station, a subsequent lazy default call should NOT
+    // re-issue createConfig with vultisig-0.
     setupLifi({ integratorName: 'station', apiUrl: 'https://api.vultisig.com/lifi/' })
-    setupLifi({ integratorName: 'someone-else' })
+    setupLifi() // lazy fallback
     expect(createConfigSpy).toHaveBeenCalledTimes(1)
     expect(lifiConfig.integratorName).toBe('station')
+  })
+
+  it('consumer setupLifi(config) AFTER lazy default re-runs createConfig (footgun fix)', () => {
+    // Ehsan-saradar #618: previously the lazy default would latch
+    // configured=true with vultisig-0, and a later consumer bootstrap
+    // would silently no-op — Station's apiUrl proxy gets lost. The fix:
+    // the consumer-bootstrap branch always runs createConfig regardless
+    // of latch state.
+    setupLifi() // lazy path runs first (e.g. a swap quote fires early)
+    setupLifi({ integratorName: 'station', apiUrl: 'https://api.vultisig.com/lifi/' })
+    expect(createConfigSpy).toHaveBeenCalledTimes(2)
+    expect(createConfigSpy).toHaveBeenLastCalledWith({
+      integrator: 'station',
+      apiUrl: 'https://api.vultisig.com/lifi/',
+    })
+    expect(lifiConfig.integratorName).toBe('station')
+    expect(lifiConfig.apiUrl).toBe('https://api.vultisig.com/lifi/')
+  })
+
+  it('consumer setupLifi(config) repeated calls re-run createConfig with each new config', () => {
+    // Mirrors the production reality where a misuse (two consumers in the
+    // same process) results in the second-caller's createConfig taking
+    // effect on the LI.FI SDK. Document the semantics so a future
+    // multi-tenant setup doesn't surprise anyone.
+    setupLifi({ integratorName: 'station' })
+    setupLifi({ integratorName: 'someone-else' })
+    expect(createConfigSpy).toHaveBeenCalledTimes(2)
+    expect(lifiConfig.integratorName).toBe('someone-else')
   })
 })
--- a/packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.integrator.test.ts
+++ b/packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.integrator.test.ts
@@ -1,0 +1,136 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { _resetLifiConfigForTest, lifiConfig, setupLifi } from '../config'
+
+// Capture every getQuote(...) invocation so we can assert the `integrator`
+// field is the consumer-supplied override (Station's `station`), NOT the
+// SDK-default vultisig-0. createConfig is a no-op spy because the @lifi/sdk
+// global runs once and persists across tests inside the same worker.
+const getQuoteSpy = vi.fn()
+const createConfigSpy = vi.fn()
+
+vi.mock('@lifi/sdk', () => ({
+  ChainId: {},
+  createConfig: (...args: unknown[]) => createConfigSpy(...args),
+  getQuote: (...args: unknown[]) => getQuoteSpy(...args),
+}))
+
+// Minimal shims for the @vultisig/core-chain imports getLifiSwapQuote pulls.
+// We only need the surface getLifiSwapQuote actually touches; everything
+// else is stubbed to keep the test focused on the integrator wire-up.
+vi.mock('@vultisig/core-chain/ChainKind', () => ({
+  getChainKind: () => 'evm',
+  DeriveChainKind: {},
+}))
+vi.mock('@vultisig/core-chain/chains/solana/solanaConfig', () => ({
+  solanaConfig: { ataRentLamports: 0 },
+}))
+vi.mock('@vultisig/core-chain/coin/chainFeeCoin', () => ({
+  chainFeeCoin: new Proxy({}, { get: () => ({ ticker: 'ETH', id: 'ETH' }) }),
+}))
+vi.mock('@vultisig/core-chain/swap/general/lifi/LifiSwapEnabledChains', () => ({
+  lifiSwapChainId: new Proxy({}, { get: () => 1 }),
+}))
+vi.mock('@vultisig/core-chain/swap/general/lifi/api/injectSolanaAtaIfMissing', () => ({
+  injectSolanaAtaIfMissing: () => ({ data: '', ataInjected: false }),
+}))
+vi.mock('@vultisig/lib-utils/assert/shouldBePresent', () => ({
+  shouldBePresent: <T>(v: T): T => v,
+}))
+vi.mock('@vultisig/lib-utils/match', () => ({
+  match: (_kind: string, handlers: Record<string, () => unknown>) => handlers.evm(),
+}))
+// Always-evaluate `memoize` so test resets actually rebuild the closure;
+// production behaviour (memoise once) is what we want to verify too —
+// covered by lifiConfig.integratorName flipping under setupLifi.
+vi.mock('@vultisig/lib-utils/memoize', () => ({
+  memoize: <T extends (...a: unknown[]) => unknown>(fn: T) => fn,
+}))
+vi.mock('@vultisig/lib-utils/record/mirrorRecord', () => ({
+  mirrorRecord: () => ({}),
+}))
+vi.mock('@vultisig/lib-utils/TransferDirection', () => ({
+  TransferDirection: { from: 'from', to: 'to' },
+}))
+
+const baseInput = {
+  from: { id: 'USDC', chain: 'Ethereum', address: '0xfrom', ticker: 'USDC' },
+  to: { id: 'USDT', chain: 'Ethereum', address: '0xto', ticker: 'USDT' },
+  amount: 1_000_000n,
+  affiliateBps: 30,
+}
+
+describe('getLifiSwapQuote — integrator override', () => {
+  beforeEach(() => {
+    getQuoteSpy.mockReset()
+    createConfigSpy.mockReset()
+    _resetLifiConfigForTest()
+    // Stable bridge-less EVM EVM happy-path response — enough to satisfy the
+    // function's post-quote unwrap. The test only inspects getQuote's args.
+    getQuoteSpy.mockResolvedValue({
+      transactionRequest: {
+        value: '0',
+        gasLimit: '0',
+        data: '0x',
+        from: '0xfrom',
+        to: '0xto',
+        chainId: 1,
+      },
+      estimate: {
+        toAmount: '999000',
+        gasCosts: [{ amount: '0' }],
+        feeCosts: [{ name: 'LIFI Fixed Fee', amount: '0', token: { decimals: 6, address: 'USDT', chainId: 1 } }],
+      },
+    })
+  })
+  afterEach(() => {
+    _resetLifiConfigForTest()
+  })
+
+  it('falls back to vultisig-0 default when no LifiAffiliateConfig provided', async () => {
+    const { getLifiSwapQuote } = await import('./getLifiSwapQuote')
+    await getLifiSwapQuote(baseInput as never)
+    expect(getQuoteSpy).toHaveBeenCalledTimes(1)
+    expect(getQuoteSpy.mock.calls[0]![0].integrator).toBe('vultisig-0')
+  })
+
+  it('uses consumer-supplied integratorName when LifiAffiliateConfig provided', async () => {
+    const { getLifiSwapQuote } = await import('./getLifiSwapQuote')
+    await getLifiSwapQuote({
+      ...baseInput,
+      lifiAffiliateConfig: { integratorName: 'station' },
+    } as never)
+    expect(getQuoteSpy.mock.calls[0]![0].integrator).toBe('station')
+  })
+
+  it('per-call override does NOT mutate the global lifiConfig', async () => {
+    const { getLifiSwapQuote } = await import('./getLifiSwapQuote')
+    await getLifiSwapQuote({
+      ...baseInput,
+      lifiAffiliateConfig: { integratorName: 'station' },
+    } as never)
+    expect(lifiConfig.integratorName).toBe('vultisig-0')
+  })
+
+  it('setupLifi({integratorName, apiUrl}) calls createConfig with both fields', () => {
+    setupLifi({ integratorName: 'station', apiUrl: 'https://api.vultisig.com/lifi/' })
+    expect(createConfigSpy).toHaveBeenCalledWith({
+      integrator: 'station',
+      apiUrl: 'https://api.vultisig.com/lifi/',
+    })
+    expect(lifiConfig.integratorName).toBe('station')
+    expect(lifiConfig.apiUrl).toBe('https://api.vultisig.com/lifi/')
+  })
+
+  it('setupLifi() with no config calls createConfig with just the default integrator (no apiUrl)', () => {
+    setupLifi()
+    expect(createConfigSpy).toHaveBeenCalledWith({ integrator: 'vultisig-0' })
+  })
+
+  it('setupLifi is idempotent — second call is a no-op', () => {
+    setupLifi({ integratorName: 'station', apiUrl: 'https://api.vultisig.com/lifi/' })
+    setupLifi({ integratorName: 'someone-else' })
+    expect(createConfigSpy).toHaveBeenCalledTimes(1)
+    expect(lifiConfig.integratorName).toBe('station')
+  })
+})
--- a/packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.ts
+++ b/packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.ts
@@ -1,8 +1,8 @@
-import { ChainId, createConfig, getQuote } from '@lifi/sdk'
+import { ChainId, getQuote } from '@lifi/sdk'
 import { DeriveChainKind, getChainKind } from '@vultisig/core-chain/ChainKind'
 import { solanaConfig } from '@vultisig/core-chain/chains/solana/solanaConfig'
 import { chainFeeCoin } from '@vultisig/core-chain/coin/chainFeeCoin'
-import { lifiConfig } from '@vultisig/core-chain/swap/general/lifi/config'
+import { LifiAffiliateConfig, lifiConfig, setupLifi } from '@vultisig/core-chain/swap/general/lifi/config'
 import { lifiSwapChainId, LifiSwapEnabledChain } from '@vultisig/core-chain/swap/general/lifi/LifiSwapEnabledChains'
 import { shouldBePresent } from '@vultisig/lib-utils/assert/shouldBePresent'
 import { match } from '@vultisig/lib-utils/match'
@@ -17,6 +17,9 @@ import { injectSolanaAtaIfMissing } from './injectSolanaAtaIfMissing'
 type Input = Record<TransferDirection, AccountCoinKey<LifiSwapEnabledChain> & { ticker?: string }> & {
   amount: bigint
   affiliateBps?: number
+  /** Consumer-supplied LI.FI integrator override. When omitted, falls back
+   * to the global `lifiConfig.integratorName` (vultisig-0 by default). */
+  lifiAffiliateConfig?: LifiAffiliateConfig
 }
 
 // Stable-pair detection: tickers that commonly trade within a tight peg.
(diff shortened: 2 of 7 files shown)


packages/core/chain/swap/general/lifi/config.ts
```diff
@@ -1,3 +1,60 @@
...
+/**
+ * Initialise the global LI.FI SDK config. Consumers (e.g. Station's mcp-ts
+ * deployment) call this once at module boot with their integrator + proxy
+ * URL. Subsequent calls are no-ops — the first caller wins.
+ *
+ * When unset, `getLifiSwapQuote` lazy-inits with the `vultisig-0` default
+ * via the same internal memoisation path.
+ */
+export const setupLifi = (config?: LifiAffiliateConfig): void => {
+  if (configured) return
```
Ehsan-saradar: Small footgun here: whoever calls `setupLifi` first wins. If any swap quote runs before the consumer's boot-time call, the lazy path latches `configured = true` and the consumer's later call is silently ignored. The per-call `integrator` still works, but the `apiUrl` proxy gets lost for the whole process and nobody finds out. Could you `console.warn` when a config comes in after `configured` is already set? Turns a silent miss into something you can actually see in logs.

packages/core/chain/swap/general/lifi/config.ts
```diff
@@ -1,3 +1,60 @@
...
+ * integrator instead of the SDK-default `vultisig-0`.
+ *
+ * `apiUrl` is honoured only when this config drives the global
+ * `setupLifi(...)` call — LI.FI's `getQuote` does not accept a per-call
+ * `apiUrl` override. Consumers that need a proxied base URL must call
+ * `setupLifi({integratorName, apiUrl})` at module boot.
+ */
+export type LifiAffiliateConfig = {
+  integratorName: string
+  apiUrl?: string
```
Ehsan-saradar, line 45: Heads up that `apiUrl` here is a bit of a trap. The same type is used in `SwapAffiliateConfig.lifi`, but on that path `getLifiSwapQuote` only reads `integratorName` and ignores `apiUrl`. So someone setting `lifi: { integratorName, apiUrl }` would expect the proxy to apply, and it just won't. The doc comment explains it, but maybe split the per-call shape (`{ integratorName }`) from the boot shape (`{ integratorName, apiUrl }`) so the type can't promise something it doesn't do?

Ehsan-saradar (commented):
Nice work, the curl receipt makes the affiliate flow easy to trust. One thing that needs a look before this lands, plus two small notes inline.

**The React Native copy got left behind.** `packages/sdk/src/platforms/react-native/overrides/getLifiSwapQuote.ts` is a hand-kept mirror of this function (the header even says "mirrors core byte-for-byte"). It still has no `lifiAffiliateConfig` and never passes `integrator` to `getQuote`, so on RN a consumer's `lifi` config is just dropped and fees keep going to `vultisig-0`. If mcp-ts is Node only and RN is out of scope, that's fine, but can you say so in a short comment there? Otherwise the next person will assume the two stay in sync.

packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.integrator.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      },
+    })
+  })
+  afterEach(() => {
+    _resetLifiConfigForTest()
+  })
+
+  it('falls back to vultisig-0 default when no LifiAffiliateConfig provided', async () => {
+    const { getLifiSwapQuote } = await import('./getLifiSwapQuote')
+    await getLifiSwapQuote(baseInput as never)
```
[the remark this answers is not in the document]
gomesalexandre: @coderabbitai meh, ya nit. the `baseInput` here is a partial spread of all-the-things `getLifiSwapQuote` could possibly need, and the test only inspects the integrator wire-up - `as never` is the right level of "I know this is unsafe and that's the point". A `createTestInput` helper with `as Parameters<...>[0]` would carry the same compile-time lie behind more ceremony; not worth the maintenance. The runtime mock-stubs (createConfig spy + the static-evaluate `@lifi/sdk` mock) are what catch breakage if the signature evolves, not the type sticker on the call site.

packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.integrator.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  })
+
+  it('uses consumer-supplied integratorName when LifiAffiliateConfig provided', async () => {
+    const { getLifiSwapQuote } = await import('./getLifiSwapQuote')
+    await getLifiSwapQuote({
+      ...baseInput,
+      lifiAffiliateConfig: { integratorName: 'station' },
+    } as never)
+    expect(getQuoteSpy.mock.calls[0]![0].integrator).toBe('station')
+  })
```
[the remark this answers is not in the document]
gomesalexandre: @coderabbitai dis tackled (already pinned via `expect(createConfigSpy).toHaveBeenCalledTimes(...)` + the toHaveBeenCalledWith assertions across the new tests - the `as never` integrator-tag test was the lone exception and it's fine, the per-call-override test below shows the spy was called when it inspects `lifiConfig.integratorName` for non-mutation).

packages/sdk/src/platforms/react-native/overrides/getLifiSwapQuote.ts
```diff
@@ -62,15 +66,31 @@ const resolveSwapFeeChain = (chainId: ChainId, fallback: LifiSwapEnabledChain):
...
+// `createConfig`. Re-runs on every quote so consumer mutations to
+// `lifiConfig` (via core's `setupLifi(bootstrap)` or direct field writes)
+// are picked up even when those mutations happen AFTER a previous quote
+// already ran. This is the RN twin of the consumer-bootstrap fix from
+// Ehsan-saradar #618 review on the core `config.ts`.
+//
+// Cost: one extra `@lifi/sdk` `createConfig` call per quote (which itself
+// memoises against unchanged input, so the real cost is a property-equality
+// check). Worth it to avoid silently dropping consumer apiUrl proxies.
+const bootstrapLifiSdkInRN = async (): Promise<void> => {
```
NeOMakinG: `blocker:` `bootstrapLifiSdkInRN` runs `createConfig` on every quote, which triggers `@lifi/sdk`'s `getChains()` (a live HTTP call to `li.quest/v1/chains`) on every invocation — `preloadChains: true` is the default in `@lifi/sdk/src/config.ts:12`. The removed `memoize` wrapper called this once. Verified in `node_modules/@lifi/sdk/src/createConfig.ts:32`.

The comment's "the real cost is a property-equality check" is incorrect: `createConfig → createBaseConfig → config.set(options)` is `Object.assign` with no dedup, and `preloadChains` triggers `getChains()` unconditionally.

fix: Restore `memoize` for the lazy path. Mirror the core `configured` flag: run once lazily, allow re-run when `setupLifi(config)` supplies an explicit arg (signal this via the existing `lifiConfig.integratorName !== 'vultisig-0'` check or a new exported `_lifiExplicitlyConfigured` flag from `config.ts`). The per-call redundant bootstrap is not needed — core's ordering-footgun fix in `config.ts` already handles lazy-before-consumer.

packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.ts
```diff
@@ -172,6 +190,7 @@ export const getLifiSwapQuote = async ({ amount, affiliateBps, ...transfer }: In
     toAddress,
     fee: affiliateBps ? affiliateBps / 10000 : undefined,
```
NeOMakinG: `should-fix:` `affiliateBps ? affiliateBps / 10000 : undefined` silently converts `affiliateBps: 0` to `fee: undefined`. LiFi's `getQuote` then falls back to `_config.routeOptions?.fee` (verified in `@lifi/sdk` src/services/api.js `params.fee ??= _config.routeOptions?.fee`), which could be non-zero if a prior `createConfig` set `routeOptions`. Pre-existing, but more impactful now that more consumers will pass `affiliateConfig`.

fix: `fee: affiliateBps != null ? affiliateBps / 10000 : undefined`

NeOMakinG (changes requested):
70/70 tests pass, CI all green. Triple-lane review (Claude + Codex gpt-5.4 + GLM glm-4.6). One blocker on the RN override; everything else is clear.

## testing approach

Static diff review across all 5 changed files. Verified `@lifi/sdk@3.16.3` source at `createConfig` / `config.set` / `getChains` call sites to confirm per-call cost. Confirmed `packages/core/chain/package.json` export-map for the `./swap/general/lifi/config` deep-import path. Traced `mcp-ts/stationAffiliateConfig.ts` consumer for backward-compat and companion-PR sequencing.

## blocking

**B1 — RN override: `bootstrapLifiSdkInRN` fires `createConfig` (+ `getChains` network call) on every quote**

`packages/sdk/src/platforms/react-native/overrides/getLifiSwapQuote.ts:80-93`

`@lifi/sdk`'s `createConfig` defaults `preloadChains: true` and fires `getChains()` — a live HTTP call to `li.quest/v1/chains` — on every invocation. The removed `memoize` wrapper meant the old code called this once; `bootstrapLifiSdkInRN` calls it on every `getLifiSwapQuote` request. On RN, every LiFi swap quote now triggers a redundant chain-list network round-trip before the actual quote request.

The comment says "the real cost is a property-equality check" — that's false. Codex verified `createConfig → createBaseConfig → config.set(options)` is `Object.assign`, no dedup, and `if (_config.preloadChains) { createChainsConfig() }` runs unconditionally.

The ordering-footgun concern (consumer `setupLifi` arriving after a lazy quote already ran) is already solved on the core side: `setupLifi(config)` with an explicit arg always re-runs regardless of `configured`. The RN override doesn't need per-call reinit — it needs to detect whether `setupLifi` has been called with a consumer config and run once, same pattern as core.

action: Replace `bootstrapLifiSdkInRN` with a `memoize`-wrapped function that reads `lifiConfig` at first-call time, OR — if the "consumer may arrive late" scenario is real on RN — add a `configured` flag mirroring core: run once lazily, but allow re-run when `setupLifi(config)` has been called with explicit args (signal this via the existing `lifiConfig.integratorName !== 'vultisig-0'` check, or export a dirty flag from `config.ts`).

## should-fix

**SF1 — `setupLifi` / `LifiBootstrapConfig` not exported from `@vultisig/sdk` barrel**

`packages/sdk/src/index.ts:964` exports `SwapAffiliateConfig` but not `setupLifi` or `LifiBootstrapConfig`. The changeset explicitly names `setupLifi(config?)` as new `@vultisig/sdk` surface consumers call at module boot. Today they must deep-import `@vultisig/core-chain/swap/general/lifi/config`. That path IS export-mapped in `packages/core/chain/package.json:1363-1366` and mcp-ts already does it correctly via `require(...)`, so it works — but it violates the stated API contract and any consumer reading the changeset notes will try `import { setupLifi } from '@vultisig/sdk'` and get a compile error.

action: Add to `packages/sdk/src/index.ts`:
```ts
export { setupLifi } from '@vultisig/core-chain/swap/general/lifi/config'
export type { LifiBootstrapConfig, LifiAffiliateConfig } from '@vultisig/core-chain/swap/general/lifi/config'
```

**SF2 — `fee: affiliateBps ? affiliateBps / 10000 : undefined` silently drops a 0-bps affiliate fee**

`packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.ts:191` and `packages/sdk/src/platforms/react-native/overrides/getLifiSwapQuote.ts:122`

If a consumer passes `affiliateBps: 0` explicitly (e.g. testing without fees), the truthy check converts it to `fee: undefined`. LiFi's `getQuote` then falls back to `_config.routeOptions?.fee` (verified in `@lifi/sdk` source line 1286), which may be non-zero if a prior `createConfig` set `routeOptions`. This is pre-existing behavior — not introduced by this PR — but it becomes more visible now that more consumers are passing `affiliateConfig`.

action: Change to `fee: affiliateBps != null ? affiliateBps / 10000 : undefined` (or `affiliateBps !== undefined ? ...`). This matches how LiFi's own SDK handles the zero-fee case.

## q

**Q1 — `_resetLifiConfigForTest` is a named export on `@vultisig/core-chain`; is it in the dist bundle?**

The `@internal` JSDoc is present but the function is a real export. If consumers import the whole module or if tree-shaking misses it, this ships in production. Does the build pipeline strip `@internal` exports, or should this be in a separate `*.test-utils.ts` file?

## risk

The RN override regression means every LiFi swap quote on mobile fires an extra HTTP round-trip (chain list fetch). This doesn't break correctness — the `integrator` is correctly wired and fee routing works — but it adds latency and LiFi API quota consumption on every RN quote. The core (Node/mcp-ts) path is unaffected.

## verdict

REQUEST CHANGES — B1 is a real latency regression on RN; SF1 means the advertised public API isn't reachable from the SDK barrel. Otherwise the affiliate-config design is clean: ordering-footgun fix is correct, test coverage is thorough (including the lazy-before-consumer ordering case), companion mcp-ts#323 already has the consumer wiring, and the live receipt confirms `station` integrator + proxy URL work end-to-end.

---
reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` adversarial pass + GLM `glm-4.6` third-eye. static-only review. commit `ff218e84979e681e7497f1e5868f73290dd47e1a`.

packages/core/chain/swap/general/lifi/config.ts
```diff
@@ -1,3 +1,103 @@
+import { createConfig } from '@lifi/sdk'
```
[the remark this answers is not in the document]
gomesalexandre: @coderabbitai ack on the eager-eval analysis, but pushing back on the fix shape. Tried the lazy-require pattern (`require('@lifi/sdk')` inside `setupLifi`) and it tanks the vitest mock interception — `vi.mock('@lifi/sdk', ...)` in `getLifiSwapQuote.integrator.test.ts` doesn't catch runtime `require()` calls the way it catches static ESM imports, so 5/8 tests went red with `createConfigSpy` never invoked.

The structural fix (split `config.ts` into pure-types `config.ts` + bootstrap-only `setupLifi.ts` that imports `@lifi/sdk`, so RN consumers never pull the bootstrap module in their bundle) is the right answer but it's a heavier refactor than scope here and touches the consumer-import surface. Filing as a follow-up.

In practice, the RN override never calls `setupLifi` (it has its own `bootstrapLifiSdkInRN`), so the eager-eval risk depends on bundle ordering of polyfills vs `@vultisig/core-chain/swap/general/lifi/config.ts`. The override's `await import('@lifi/sdk')` runtime barrier still gates the actual module-body evaluation for `getQuote` / `createConfig` USAGE — what's eager is the import resolution, which Metro can hoist past the polyfill if it inlines, but inlining is the worst case and the polyfill is installed early in the RN entry. Worth fixing properly, but not in this PR — would rather ship the per-call integrator + apiUrl override now and split the file in a focused follow-up where the diff is just the structural move.

For the docblock nit — fixed in fe0cf7e0 (last-writer wins clarification + reference to the repeated-call test that pins the contract).

packages/core/chain/swap/general/lifi/config.ts
```diff
@@ -1,3 +1,103 @@
...
+ *    currently holds — `vultisig-0` + no apiUrl unless a consumer mutated
+ *    them already.
+ *
+ * Net effect:
+ * - Consumer-before-lazy: consumer wins (the only `createConfig` call).
+ * - Lazy-before-consumer: lazy installs defaults, then the consumer's
+ *   explicit call overwrites them via the explicit-bootstrap branch above.
+ * - Consumer-after-consumer: first consumer-bootstrap call still wins for
+ *   `lifiConfig` state, but the second consumer-bootstrap call re-runs
+ *   `createConfig`. In practice consumers should only call this once.
```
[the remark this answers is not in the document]
gomesalexandre: @coderabbitai dis tackled in fe0cf7e0 - swapped "first wins" for "latest overwrites + re-runs createConfig" and pointed to the `'repeated calls re-run createConfig with each new config'` test as the contract pin.

packages/sdk/src/platforms/react-native/overrides/getLifiSwapQuote.ts
gomesalexandre: @NeOMakinG ye dog ya cooked on this one - the memoize was already restored in 672d8b31 (landed 25 min after your review at 13:47Z vs your review at 13:22Z). The current state has `ensureLifiConfiguredInRN = memoize(() => { setupLifi() })` which runs setupLifi exactly once per process, and setupLifi's lazy branch calls createConfig exactly once via the `configured` latch. So `createConfig` → `getChains` HTTP fires once at first-quote like the original behavior, not per-quote.

The comment block above the function still says "Re-runs on every quote" though, which is a stale leftover from an earlier intermediate version — cleaning that up now. Receipts: `vitest --run lifi` → 29/29 green.

(consumer-bootstrap branch is also safe: the explicit `setupLifi({integratorName, apiUrl})` from a consumer overwrites lifiConfig and re-runs createConfig once at consumer-boot time. memoize'd `ensureLifiConfiguredInRN` then just no-ops on subsequent quotes because `configured = true` was already set inside setupLifi.)

packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.ts
gomesalexandre: @NeOMakinG tackled in 14eba5d9 - swapped `affiliateBps ? affiliateBps / 10000 : undefined` for `affiliateBps !== undefined ? affiliateBps / 10000 : undefined` in both core and RN override. Now 0-bps stays 0 instead of falling through the truthy-test and inheriting `_config.routeOptions?.fee` silently. Good catch on the LiFi `??=` fallback, would've been a fund-safety silent default. tyvm 🙇

packages/core/chain/swap/general/lifi/config.ts
gomesalexandre: @Ehsan-saradar tackled in ff218e84 + f3185ffc — the lazy/consumer ordering ambiguity is now killed by an explicit consumer-bootstrap branch in `setupLifi`:

- **Consumer-bootstrap call** (`setupLifi(config)` with a config arg): ALWAYS applies regardless of whether the lazy default-fallback already ran. Re-invokes `createConfig` with the supplied integrator + apiUrl. Last-writer wins.
- **Lazy default** (`setupLifi()` no arg, called from `ensureLifiConfigured`): only runs once, idempotent, falls back to whatever `lifiConfig` currently holds.

Net: if a swap quote runs first (lazy installs `vultisig-0`), then a consumer calls `setupLifi({...})`, the consumer's call overwrites `lifiConfig` AND re-runs `createConfig` so the apiUrl proxy actually lands. No silent drop. The contract is pinned by the test `'repeated calls re-run createConfig with each new config'` in `getLifiSwapQuote.integrator.test.ts`. Cited your review id in the jsdoc so future readers see the rationale.

packages/core/chain/swap/general/lifi/config.ts
gomesalexandre: @Ehsan-saradar tackled in ff218e84 — split the per-call shape from the boot shape so the trap can't happen:

- `LifiAffiliateConfig` (consumer-supplied via `SwapAffiliateConfig.lifi`) is now **intentionally narrower** — `{ integratorName: string }` only. No `apiUrl`. If a consumer writes `lifi: { integratorName, apiUrl }` it's a TS error, not a silent drop.
- `LifiBootstrapConfig` (passed to `setupLifi`) stays wider with the optional `apiUrl` because it's meaningful at the global SDK init level.

The jsdoc on `LifiAffiliateConfig` spells out the why and cites your review ID so future readers know the type split is load-bearing, not accidental. Exactly the "split the shape, surface the API mismatch at compile time" approach you suggested.

NeOMakinG (approved):
gm — multi-lane pass: Claude. conceptual-only (no runtime this pass). lgtm.

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | RN override bootstrapLifiSdkInRN fires createConfig (+ getChains HTTP call) on every quote | fixed |
| 1 | setupLifi / LifiBootstrapConfig not exported from @vultisig/sdk barrel | fixed |
| 1 | affiliateBps: 0 silently dropped via truthy check | fixed |
| 1 | _resetLifiConfigForTest named export in dist bundle | deferred |

## risk
LOW — all three prior blockers/should-fixes resolved. RN override now memoized. Barrel exports correct. affiliateBps=0 preserved. CI green on current HEAD 6c532fd3 (lint pending on 6fa8dfc5 sort-only commit).

## verdict
APPROVE — ship it 🫡

PR #621, opened 2026-06-02
fix(ton): use owner_id + jetton_master_id in jetton wallet proxy calls

gomesalexandre:
Sibling to mcp-ts#324 - sdk-side `getJettonWalletsUrl` had the same stale `owner_address`/`jetton_address` params.

## what

The Vultisig proxy at `/ton/v3/jetton/wallets` requires `owner_id` + `jetton_master_id`. The code was sending `owner_address` + `jetton_address` - params the proxy silently ignores, returning an empty array every time.

Result: `getJettonBalance` always returned `0n`, `getJettonWalletAddress` always threw `"No jetton wallet found"`. Every SDK-side TON jetton balance query and jetton transfer broke silently regardless of actual holdings.

One surface affected in `packages/core/chain/chains/ton/api.ts`:
- `getJettonWalletsUrl` (called by both `getJettonWalletAddress` and `getJettonBalance`)

## receipts

```
yarn test:core

 ✓ packages/core/chain/chains/ton/api.test.ts > getJettonWalletAddress > uses owner_id + jetton_master_id in the proxy request URL 1ms
 ✓ packages/core/chain/chains/ton/api.test.ts > getJettonWalletAddress > returns the user_friendly address from the address_book 0ms
 ✓ packages/core/chain/chains/ton/api.test.ts > getJettonBalance > uses owner_id + jetton_master_id in the proxy request URL 0ms
 ✓ packages/core/chain/chains/ton/api.test.ts > getJettonBalance > returns the balance as bigint 0ms

 Test Files  95 passed (95)
      Tests  899 passed (899)
```

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected TON jetton wallet and balance lookups to avoid empty or incorrect results; balances are now returned reliably and as integer values to prevent transfer/display issues.

* **Tests**
  * Added regression tests to validate TON jetton API behavior and prevent future regressions.

### DIFF
--- a/packages/core/chain/chains/ton/api.test.ts
+++ b/packages/core/chain/chains/ton/api.test.ts
@@ -20,7 +20,7 @@ vi.mock('@vultisig/lib-utils/query/queryUrl', () => ({
   },
 }))
 
-import { getJettonBalance,getJettonWalletAddress } from './api'
+import { getJettonBalance, getJettonWalletAddress } from './api'
 
 const OWNER = 'EQDAFuDWly4z3eA16Ej_JHpoL6CcXdt0IRUrODKKsu60HYMi'
 const MASTER = 'EQCxE6mUtQJKFnGfaROTKOt1lZbDiiX1kCixRv7Nw2Id_sDs' // USDT mainnet
--- a/packages/core/chain/chains/ton/api.test.ts
+++ b/packages/core/chain/chains/ton/api.test.ts
@@ -1,0 +1,70 @@
+/**
+ * Regression: getJettonWalletsUrl must use `owner_id` + `jetton_master_id` query params.
+ *
+ * The Vultisig proxy at /ton/v3/jetton/wallets silently ignores the old
+ * `owner_address` / `jetton_address` params and returns an empty array,
+ * making every balance call return 0 and every transfer fail with
+ * "No jetton wallet found". Parity fix with mcp-ts#324.
+ */
+import { describe, expect, it, vi } from 'vitest'
+
+const capturedUrls: string[] = []
+
+vi.mock('@vultisig/lib-utils/query/queryUrl', () => ({
+  queryUrl: (url: string) => {
+    capturedUrls.push(url)
+    return Promise.resolve({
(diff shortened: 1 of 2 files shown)


NeOMakinG (approved):
## Triple-lane review — APPROVE

**Verdict: APPROVE** after one in-band fix.

### Lane summary

| Lane | Status | Findings |
|------|--------|----------|
| Claude (correctness + fund-safety) | Complete | 1 should-fix (BigInt empty-string guard), 2 nit suggestions |
| GLM glm-4.6 | Complete | 4 findings; all false-positive or already mitigated |
| Codex gpt-5.4 | Unavailable (TTY non-interactive) | Adversarial analysis done inline by Claude |

### Core correctness

The fix is correct and the derivation question is a red herring — this PR does NOT change TON jetton wallet derivation (stateInit code+data → address). That derivation happens on-chain. This PR only fixes the **proxy API call** that queries an already-deployed jetton wallet's address. The param rename `owner_address`+`jetton_address` → `owner_id`+`jetton_master_id` matches what the Vultisig proxy `/ton/v3/jetton/wallets` requires, confirmed by sibling mcp-ts#324.

### In-band fix pushed: `9fca4f80`

```ts
// before
return BigInt(balance ?? 0)

// after
return BigInt(balance || '0')
```

`BigInt(balance ?? 0)` throws `SyntaxError` if the proxy returns `balance: ""` (empty string for a never-funded jetton wallet). The `??` operator only guards undefined/null. `|| '0'` covers all falsy values.

### Backwards compat

None needed. The old param names were silently ignored by the proxy, returning an empty array for every user. Zero users had a working jetton balance/transfer path. Pure bugfix.

### Tests

95/95 test files, 899/899 tests pass on `pr-621` branch (pre-in-band-fix).

### Non-blocking nits

1. `encodeURIComponent()` on `rawOwner`/`rawMaster` for consistency with `getJettonMasterInfo` line 127 (which encodes). Colon is RFC 3986-legal in query values so not a real bug; mcp-ts#324 also URL-encodes defensively.
2. Strengthen test URL assertion from `toContain('owner_id=')` to `toMatch(/owner_id=[^&]+/)` so it fails if `tonAddressToRaw` returns empty string.

PR #623, opened 2026-06-02
test(chain): pin base-unit amount converter junk/whitespace rejection

gomesalexandre:
## what
Pins `toChainAmount` (the execution-layer converter from a human decimal string to chain base units, via viem `parseUnits`) so its rejection of malformed inputs is regression-guarded. Part of a broader effort to let the agent/tool layers rely on deterministic SDK guards instead of carrying defensive prose - so these guards must be provably stable across viem bumps.

## pinned (value-integrity guard, asserts existing behavior)
- empty / whitespace-only -> `ChainAmountParseError`
- trailing alpha junk (`10xyz`, `0.5abc`) -> throws (never coerced to a bare number)
- thousands-separator commas (`1,000`) -> throws (callers normalize before this layer)
- interior whitespace (`1 0`) -> throws

No functional change; tests only.

## receipts
- `vitest run --config .config/vitest.core.config.ts packages/core/chain/amount/chainAmount.test.ts` -> 10 passed
- prettier + eslint + pre-commit hooks clean


NeOMakinG (approved):
CI all green (unit 10/10, lint+type clean). APPROVE.

## testing approach

static diff-walk + implementation read + viem `parseUnits` live-probe on viem@2.51.0 (installed in repo). No runtime QA needed (test-only PR, no behavior change). Three-lane: Claude correctness + GLM third-eye; Codex lane skipped (unavailable).

## blocking

none.

## preferably-blocking

none.

## should-fix

none.

## q

none.

## risk

test-only addition (+30 lines). zero blast radius. no behavior change shipped.

## verdict

APPROVE — tests are correct, CI green, no findings.

The implementation guard at `toChainAmount.ts:92-94` (`if (!trimmed)`) fires before `parseUnits` for the empty/whitespace cases — which is the critical point, because `parseUnits('')` returns `0n` silently on viem@2.51.0 rather than throwing. The test correctly pins that the SDK layer rejects this regardless of viem's own behavior. All four junk vectors confirmed via live probe of the installed viem build.

Chain decimals: the existing pre-PR suite already covers 10 (CACAO/MayaChain), 8 (BTC/RUNE), 9 (SOL), 18 (ETH), 6 (cosmos). The new tests use `decimals=6` uniformly — correct, since shape rejection is decimal-agnostic (the guard fires before `parseUnits` on empty input; viem throws on junk for any decimal value).

GLM third-eye flagged round-trip coverage and chain-specific precision vectors as P1/P2. These are valid follow-up candidates but out of scope for this PR's stated intent (shape-rejection backstop). GLM's scientific-notation P1 is a false positive — the implementation already intercepts `/[eE]/` and routes through `expandScientificNotationToDecimalString`, covered by pre-existing tests.

---
reviewed via [Claude Code](https://claude.com/claude-code). GLM glm-4.6 third-eye pass (3 findings, all out-of-scope or FP). Codex unavailable.

PR #633, opened 2026-06-03
feat(sdk): add getVaultByName + getVaultByNameOrThrow on VaultManager + Vultisig

### LINKED ISSUE #153: feat(sdk): add getVaultByName() convenience method
## Summary

Every agent scenario in our autoresearch eval (4 out of 4 scenarios that loaded existing vaults) had to independently discover the `listVaults()` + `.find()` pattern. No agent found a direct method. This is a small convenience wrapper with outsized impact on developer experience.

## Current (every agent guesses this)

```typescript
const vaults = await sdk.listVaults()
const vault = vaults.find(v => v.name === 'MyVault')
if (!vault) throw new Error('Vault not found')
```

## Proposed

```typescript
// Returns vault or null
const vault = await sdk.getVaultByName('MyVault')

// Throws with helpful error listing available vaults
const vault = await sdk.getVaultByNameOrThrow('MyVault')
// Error: Vault "MyVault" not found. Available vaults: Main Wallet, Test Vault
```

## Implementation

~5 lines in `VaultManager.ts`:

```typescript
async getVaultByName(name: string): Promise<VaultBase | null> {
  const vaults = await this.listVaults()
  return vaults.find(v => v.name === name) ?? null
}
```

And expose via `Vultisig` class (same as `getVaultById`).

## Context

Autoresearch eval scored First Vault dimension at 6.33/10 across all scenarios. This was the single most common friction point for vault loading — appeared in scenarios 2, 3, 4, and 5.

gomesalexandre:
closes #153

## what

Autoresearch eval surfaced that every agent scenario loading a vault by name had to independently rediscover the `listVaults() + .find()` pattern — no direct lookup method existed. Adds the obvious shape:

```ts
const vault = await sdk.getVaultByName('Main Wallet')         // null if missing
const vault = await sdk.getVaultByNameOrThrow('Main Wallet')  // throws with available-vault list on miss
```

Both methods land on `VaultManager` and are re-exposed on the `Vultisig` class the same way `getVaultById` is, so callers using either entrypoint get the same convenience surface.

## design notes

- **Case-sensitive** — mirrors `find(v => v.name === name)` exactly. If a future caller wants case-insensitive lookup that's a separate method, not a silent behaviour change here. Test pins the contract so a well-meaning future refactor can't quietly downcase.
- **Duplicate names tie-break: first match in `listVaults()` order.** Storage doesn't enforce name uniqueness (legal-if-unusual state, e.g. user imported a backup of an existing vault). Test pins the tie-break so it doesn't drift.
- **`getVaultByNameOrThrow` lists available names in the error message** so a CLI typo or eval-scenario misconfig surfaces the right hint:
  - With vaults loaded: `Vault "Typo" not found. Available vaults: Main Wallet, Backup`
  - Empty state: `Vault "X" not found and no vaults are loaded`

## test receipt

- `npx vitest run tests/unit/vault/VaultManager.test.ts` → **33 passed** (full file, 9 new sub-cases across `getVaultByName` and `getVaultByNameOrThrow`)
- `npx eslint --config .config/eslint.config.mjs` clean on all 3 touched files
- `npx tsc --noEmit` clean on `packages/sdk`

New tests:

| Sub-case | Pins |
|---|---|
| `getVaultByName: returns null when no vaults exist` | empty-state safety |
| `getVaultByName: returns null when no vault matches the name` | non-match safety |
| `getVaultByName: returns the vault when a name matches exactly` | happy path |
| `getVaultByName: is case-sensitive` | `main wallet` / `MAIN WALLET` both return null when `Main Wallet` is loaded |
| `getVaultByName: disambiguates the right vault when multiple are loaded` | 3-vault scenario |
| `getVaultByName: returns first match in listVaults order when duplicate names exist` | tie-break contract |
| `getVaultByNameOrThrow: throws with empty-vault hint when no vaults are loaded` | empty-state error copy |
| `getVaultByNameOrThrow: throws and lists available names when no vault matches` | available-vault hint |
| `getVaultByNameOrThrow: resolves the vault when the name matches` | happy path |

## no runtime receipt needed

Pure SDK convenience method with no side effects, no network/storage write, no signing surface. Unit coverage is the receipt — same shape as the existing `getVaultById` tests.

### DIFF
--- a/packages/sdk/src/VaultManager.ts
+++ b/packages/sdk/src/VaultManager.ts
@@ -312,6 +312,62 @@ export class VaultManager {
   }
 
   /**
+   * Get vault instance by display name. Case-sensitive lookup against the
+   * `name` field on every stored vault. Returns `null` if no vault has a
+   * matching name; if multiple vaults share the same name (the storage layer
+   * doesn't enforce uniqueness), returns the FIRST match in `listVaults()`
+   * order (the same `order` field listVaults sorts by).
+   *
+   * Convenience wrapper around `listVaults().find()` — every autoresearch
+   * agent scenario that loaded a vault by name had to independently rediscover
+   * that pattern (see issue #153). This method makes "find my vault by the
+   * name the user typed" a single call.
+   *
+   * @param name - Vault display name as set at creation / shown in the UI
+   * @returns VaultBase instance or null if no vault has that name
+   * @example
+   * ```typescript
+   * const vault = await vaultManager.getVaultByName('Main Wallet')
+   * if (vault) {
+   *   const balance = await vault.balance('Bitcoin')
+   * }
+   * ```
+   */
+  async getVaultByName(name: string): Promise<VaultBase | null> {
+    const vaults = await this.listVaults()
+    return vaults.find(v => v.name === name) ?? null
+  }
+
+  /**
+   * Get vault instance by display name, throwing with a helpful error that
+   * lists available vault names when no match is found. Use this when a
+   * missing vault is a programming error in the caller (CLI argument typo,
+   * eval scenario misconfigured) rather than a user-facing fallthrough.
+   *
+   * @param name - Vault display name
+   * @returns VaultBase instance
+   * @throws Error if no vault matches, with the available vault names in the
+   *   message so the caller can see what they should have asked for.
+   * @example
+   * ```typescript
+   * try {
+   *   const vault = await vaultManager.getVaultByNameOrThrow('TestVault')
+   * } catch (e) {
+   *   // e.message: 'Vault "TestVault" not found. Available vaults: Main Wallet, Backup'
+   * }
+   * ```
+   */
+  async getVaultByNameOrThrow(name: string): Promise<VaultBase> {
+    const vault = await this.getVaultByName(name)
+    if (vault) {
+      return vault
+    }
+    const available = (await this.listVaults()).map(v => v.name).join(', ')
+    const suffix = available.length > 0 ? `. Available vaults: ${available}` : ' and no vaults are loaded'
+    throw new Error(`Vault "${name}" not found${suffix}`)
+  }
+
+  /**
    * Get all vault instances
    * Async equivalent to listVaults()
    *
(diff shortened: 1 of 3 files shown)


packages/sdk/src/VaultManager.ts
```diff
@@ -311,6 +311,62 @@ export class VaultManager {
...
+   * agent scenario that loaded a vault by name had to independently rediscover
+   * that pattern (see issue #153). This method makes "find my vault by the
+   * name the user typed" a single call.
+   *
+   * @param name - Vault display name as set at creation / shown in the UI
+   * @returns VaultBase instance or null if no vault has that name
+   * @example
+   * ```typescript
+   * const vault = await vaultManager.getVaultByName('Main Wallet')
+   * if (vault) {
```
NeOMakinG, line 331: `should-fix:`

NeOMakinG (approved):
gm — multi-lane pass: claude + glm-4.6. runtime-tested on npx tsc --noEmit — clean, npx eslint — clean on all 3 touched files, npx vitest run tests/unit/vault/VaultManager.test.ts — 33/33 PASS (9 new sub-cases). lgtm.

## testing approach
- npx tsc --noEmit — clean
- npx eslint — clean on all 3 touched files
- npx vitest run tests/unit/vault/VaultManager.test.ts — 33/33 PASS (9 new sub-cases)
- pass order: claude → glm-4.6

## should-fix
### 1.  — `packages/sdk/src/VaultManager.ts:331`
action: 

## risk
Pure read-only convenience methods. No side effects, no writes, no signing surface. Delegates to `listVaults().find()` which is the existing canonical lookup pattern. `getVaultByNameOrThrow` error message includes vault name in the Error.message — no security concern in a client-only SDK context.

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` adversarial pass. npx tsc --noEmit — clean, npx eslint — clean on all 3 touched files, npx vitest run tests/unit/vault/VaultManager.test.ts — 33/33 PASS (9 new sub-cases).

packages/sdk/src/VaultManager.ts
gomesalexandre: @NeOMakinG ye dog same empty-body pattern as your ab#933 thread on `balance_preflight.go` — the `should-fix:` prefix shipped without a body. I think the new CR UI is dropping the body on submit (you've now done this on 4 PRs in a row: ab#933, sdk#633, sdk#634, ab#941). Want to drop the actual concern in a fresh reply here, or resolve if it was meant as an empty marker? carn mate 🦘

PR #638, opened 2026-06-03
fix(sdk): base64 + parsing in keygen QR (takeover of #586)

gomesalexandre:
Supersedes #586 (couldn't push to the fork branch, maintainerCanModify=false).

Takes over @christroutner's base64 parsing fix and gets it green: the original PR's tests passed but the static quality gate failed with `TS2593: Cannot find name 'beforeEach'` (globals:true makes it work at runtime, but tsc still needs the explicit import). Added the missing `beforeEach` import from vitest.

## what
`URLSearchParams` decodes `+` as space per the form-urlencoded spec. Base64 data in the keygen QR `jsonData` param legitimately contains `+`, so a payload like `ICA+` arrives as `ICA ` and decodes to the wrong bytes - corrupting vault import from a scanned QR. Fix restores `+` before `decodeURIComponent`.

## how
- `parseKeygenQR.ts`: `jsonData.replace(/ /g, '+')` before decode (christroutner)
- regression test exercising raw `+`, `%3D` padding, and standard base64 with both `+` and `/` (christroutner + takeover hardening)
- import `beforeEach` from vitest to satisfy the type-check gate (this takeover)

Credit: @christroutner for the original fix + test.

## receipts
```
$ yarn workspace @vultisig/sdk typecheck
typecheck exit: 0   # TS2593 gone

$ npx vitest run tests/unit/utils/parseKeygenQR.base64Encoding.test.ts --config tests/unit/vitest.config.ts
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

## risk
Low. Isolated to keygen QR parsing; restores correct base64 decode that was silently corrupting `+`-containing payloads. Pure bugfix + test, no behavior change on already-correct payloads.

### DIFF
--- a/packages/sdk/src/utils/parseKeygenQR.ts
+++ b/packages/sdk/src/utils/parseKeygenQR.ts
@@ -133,8 +133,12 @@ export async function parseKeygenQR(qrPayload: string): Promise<ParsedKeygenQR> 
     throw new Error('Invalid QR payload: missing jsonData parameter')
   }
 
+  // URLSearchParams decodes '+' as space per application/x-www-form-urlencoded spec.
+  // Base64 data legitimately contains '+' characters, so restore them.
+  const base64Safe = jsonData.replace(/ /g, '+')
+
   // URL decode and decompress
-  const decodedData = decodeURIComponent(jsonData)
+  const decodedData = decodeURIComponent(base64Safe)
   const binaryData = await decompressData(decodedData)
 
   // Parse protobuf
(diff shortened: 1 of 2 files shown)


NeOMakinG (approved):
Inline review — APPROVE.

Takeover of @christroutner #586 with the missing `beforeEach` import that was blocking the typecheck gate. Substance is unchanged: `URLSearchParams.get()` decodes `+` to space per the form-urlencoded spec, so base64 `jsonData` like `ICA+` arrives as `ICA ` and `atob()` corrupts the bytes. Fix is `replace(/ /g, "+")` before `decodeURIComponent`.

**Why it is safe:** valid base64 cannot legitimately contain ASCII space — only A–Z, a–z, 0–9, `+`, `/`, `=`. The replace is a pure inverse of the URL transport mangling.

**Test coverage:** the regression test exercises raw `+`, `%3D` padding, and standard base64 with both `+` and `/`. Already-correct payloads pass through unchanged.

Scope is the keygen QR parser only — no behavior change for payloads that never had `+`. CI pending unit + lint; should land in ~5 min.

PR #646, opened 2026-06-05
chore(zcash): update consensus branch ID to NU6.2

rcoderdev:
## Summary by CodeRabbit

* **Updates**
  * Zcash transaction signing now defaults to network protocol version NU6.2, replacing the previous NU6.1 default. This ensures transactions are compatible with the latest Zcash network parameters.
  * Developers can still explicitly specify earlier protocol versions if needed for compatibility.

* **Chores**
  * Released patch updates for SDK and Core MPC packages.


Ehsan-saradar (approved):
LGTM

PR #708, opened 2026-06-11
Fix broadcast pending status verification

### LINKED ISSUE #689: Broadcast verify-by-hash bypassed for EVM/Cosmos — bare {status:'pending'} treated as known-pending success
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` broadcast/status path).

## Observed
The broadcast verify-by-hash safety net is **silently bypassed for EVM and Cosmos**. When the status RPC itself is unreachable (or the hash is not yet indexed), a transaction that **never landed** is treated as a "known pending" success and the original broadcast error is swallowed — the UI shows a fake "done" with a hash that has no on-chain counterpart. Polkadot / Ripple / Solana are correctly protected; EVM and Cosmos are not. This is a net-new gap, not present in the iOS code path.

## Cause
The verifier swallows the error when status is `'pending'` and `isKnown !== false`:

`packages/core/chain/tx/broadcast/verifyBroadcastByHash.ts`
```ts
const isKnownPending = result.status === 'pending' && result.isKnown !== false
if (result.status === 'success' || isKnownPending) { return }
```

But the EVM and Cosmos resolvers return a **bare** `{ status: 'pending' }` (no `isKnown` field) on RPC-error / unindexed hash:

`packages/core/chain/tx/status/resolvers/evm.ts` (`getEvmTxStatus`):
```ts
if (error) { return { status: 'pending' } }   // isKnown is undefined
if (!data) { return { status: 'pending' } }
```
`packages/core/chain/tx/status/resolvers/cosmos.ts` (`getCosmosTxStatus`):
```ts
if (error || !tx) { return { status: 'pending' } } // isKnown is undefined
```
`undefined !== false` is `true`, so `isKnownPending` becomes true and the broadcast error is swallowed. By contrast `polkadot.ts`, `ripple.ts`, `solana.ts` all return `{ status: 'pending', isKnown: false }` on the unknown-hash branch.

## Fix shape
Add `isKnown: false` to the RPC-error / no-data branches of `evm.ts` and `cosmos.ts` (mirroring the other three resolvers) so the safety net only swallows the duplicate-broadcast error when the chain has actually confirmed the tx exists. Reserve bare `{ status: 'pending' }` for the genuinely-indexed-but-unconfirmed case.

---

rcoderdev:
Closes #689

## Release Notes

* **Bug Fixes**
  * Improved transaction broadcast verification for unknown/unindexed EVM and Cosmos hashes by treating them as unverified, ensuring real broadcast failures are reported correctly instead of being misclassified as pending.

* **Tests**
  * Added automated coverage for EVM and Cosmos transaction status handling across pending/unknown and success scenarios.

* **Release**
  * Patch version updates for core components.

* **Chores**
  * Minor dev dependency/build configuration updates (esbuild and resolver ordering).


johnnyluo (approved):
LGTM

PR #709, opened 2026-06-11
Fix Cosmos dynamic min gas fees

### LINKED ISSUE #699: Cosmos fee amounts static constants used directly; brittle to network min-gas-price hike
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` cosmos gas path).

## Observed
Cosmos fee amounts are **static constants** used directly as the fee amount, not derived from the network's min-gas-price. This is safe today (e.g. Akash overpays slightly) but brittle: if a chain raises its minimum gas price, the hardcoded value silently under-pays and transactions start failing with no code-level signal.

## Cause
`packages/core/chain/chains/cosmos/gas.ts`
```ts
const defaultGas = 7500n
export const cosmosGasRecord: Record<IbcEnabledCosmosChain, bigint> = {
  [Chain.Cosmos]: defaultGas,
  [Chain.Akash]: 200000n,
  [Chain.TerraClassic]: 100000000n,
  ...
}
```
These are consumed directly as the fee amount with no reference to the chain's live `min_gas_price` / `gas_prices` endpoint.

## Fix shape
Treat the constants as a floor, and where a chain exposes a min-gas-price (or a node `/cosmos/base/...` gas-price query), compute `fee = gasLimit * minGasPrice` and take `max(computed, constant)`. At minimum, add a comment/test pinning the assumed min-gas-price per chain so a network bump is caught.

---

rcoderdev:
Closes #699

## Release Notes

* **New Features**
  * Added dynamic resolution of Cosmos chain gas fees from live node data, providing more accurate fee calculations while maintaining static fee amounts as a safety floor.

* **Chores**
  * Updated package dependency configurations.

### DIFF
--- a/.changeset/dynamic-cosmos-min-gas.md
+++ b/.changeset/dynamic-cosmos-min-gas.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Resolve Cosmos fee amounts from live node min gas prices when available, keeping the existing static amounts as safe floors.
--- a/package.json
+++ b/package.json
@@ -191,6 +191,7 @@
     "readable-stream": "3.6.2",
     "@lifi/sdk/@solana/web3.js": "1.98.4",
     "@cosmjs/encoding/@scure/base": "1.2.6",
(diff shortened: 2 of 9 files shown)


johnnyluo (approved):
LGTM

PR #710, opened 2026-06-11
Fix RN Cosmos fee denom fallback

### LINKED ISSUE #697: RN buildCosmosSendTx uses feeDenom ?? denom — USTC send omitting feeDenom → wrong fee denom (RN builder only; core correct)
**Inherited by: Vultisig Windows + agent-app** — NOTE: scope is the **React-Native builder only**. The SDK core path is correct (`cosmosFeeCoinDenom.ts` maps `TerraClassic → 'uluna'`), so this is a divergence in the RN variant, not the core.

## Observed
In the RN cosmos transaction builder, a USTC (TerraClassic) send that omits `feeDenom` falls back to the **send denom** instead of the chain fee denom (`uluna`). For a USTC transfer this stamps the fee in `uusd`/`uustc` rather than `uluna` — the TerraClassic fee-denom bug — causing the node to reject the tx. The `feeDenom` param exists (added in vultisig-sdk#624) but is not enforced at the builder.

## Cause
`packages/sdk/src/platforms/react-native/chains/cosmos/tx.ts:382` (`buildCosmosSendTx`), the `buildAuthInfo` call at line 393:
```ts
const authInfoBytes = buildAuthInfo(
  opts.pubKeyBytes,
  opts.sequence,
  opts.gasLimit,
  opts.feeDenom ?? opts.denom,   // ← falls back to send denom; USTC → wrong fee denom
  opts.feeAmount
)
```
The `?? opts.denom` fallback means a caller that omits `feeDenom` for a USTC send pays the fee in the asset denom. The core path avoids this by resolving via `cosmosFeeCoinDenom` (TerraClassic → `uluna`).

## Fix shape
Resolve the fee denom from the chain (mirror core's `cosmosFeeCoinDenom`, TerraClassic → `uluna`) instead of falling back to `opts.denom`, or make `feeDenom` required and assert it is the chain fee denom for TerraClassic. Don't silently inherit the send denom.

---

rcoderdev:
Closes #697

## Summary by CodeRabbit

* **Bug Fixes**
  * React Native Cosmos send transactions now default to chain-specific fee denominations when no denomination is provided.

* **Chores**
  * Pinned a gRPC dependency to a fixed version for stability.

* **Tests**
  * Expanded regression tests to cover default fee-denom behavior across chains and edge cases.

### DIFF
--- a/.changeset/noon-usdc-yield-vault.md
+++ b/.changeset/noon-usdc-yield-vault.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Add Noon sUSN Delta Neutral USDC yield vault helpers for Ethereum, including ERC-7540 calldata builders, USDC approval planning, on-chain read helpers, and Noon/Accountable APY plus TVL API clients exposed through the SDK boundary.
--- a/.changeset/rn-cosmos-fee-denom.md
+++ b/.changeset/rn-cosmos-fee-denom.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Default React Native Cosmos send fees to the chain fee denom when callers omit an explicit fee denom.
--- a/clients/cli/CHANGELOG.md
+++ b/clients/cli/CHANGELOG.md
@@ -1,5 +1,42 @@
 # @vultisig/cli
 
+## 2.3.3
+
+### Patch Changes
+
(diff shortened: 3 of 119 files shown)


johnnyluo (approved):
LGTM

Ehsan-saradar (approved):
LGTM. Defaulting to the chain's native gas denom is the right fix, and it lines the RN path up with the keysign resolver. Nice `hasOwnProperty` guard too. Just note it changes the fallback for all Cosmos chains, not only TerraClassic — which is correct, just broader than the title.

PR #711, opened 2026-06-11
Preflight THORChain halt flags

### LINKED ISSUE #696: THORChain halt detection reactive only (error-string match); no proactive inbound_addresses halted flag check
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` swap-quote path).

## Observed
THORChain halt detection is **reactive only**: the code infers a halt by string-matching the quote error after a quote attempt fails. There is no proactive check of the `inbound_addresses` `halted` / `chain_trading_paused` flags before quoting/signing, so a swap can be attempted against a halted chain and only fail late (or, worse, slip through if the halt wording isn't matched).

## Cause
`packages/core/chain/swap/quote/findSwapQuote.ts` (`isTradingHaltedMsg`, in the all-providers-failed fallback):
```ts
const isTradingHaltedMsg = (msg: string) => {
  const lower = msg.toLowerCase()
  return lower.includes('halted') || lower.includes('trading halt')
    || lower.includes('trading paused') || lower.includes('trading is paused')
}
```
Halt state is only derived from the rejected-quote error string. No call reads the authoritative `inbound_addresses[].halted` / `global_trading_paused` / `chain_trading_paused` flags before the quote/sign attempt.

## Fix shape
Add a proactive pre-flight: fetch THORChain `inbound_addresses` and check the source/destination chain's `halted` / `*_trading_paused` flags before quoting (or before signing). Surface a deterministic "trading halted" state up front instead of relying on fragile error-string matching after a failed round-trip.

---

rcoderdev:
Closes #696

## Summary by CodeRabbit

* **New Features**
  * Added preflight validation for THORChain native swaps to proactively detect trading halts and paused states before requesting quotes.

* **Bug Fixes**
  * Improved error messaging to provide clearer feedback when swaps are temporarily unavailable due to inbound trading halts or pauses.


gomesalexandre (commented):
draft review — not blocking anything, just notes while you're cooking this.

feature files look clean (the 131-file diff is just the branch being 55 commits behind main — the actual delta vs base is 10 files, tidy).

**things i like:**
- dep injection pattern on `fetchInboundAddresses` is chef's kiss for testability
- fails open on network error — correct safety call
- 5s module-level cache saves hitting inbound_addresses on every quote
- sole-THOR-route early exit ordering is right: halt > minimum > quote

**q:** maya has its own `inbound_addresses` endpoint too — any reason we're not checking it? or is maya halt detection out of scope for this PR?

**q2:** `chain_lp_actions_paused` is not checked (only `halted` + `chain_trading_paused`). intentional? LP pause doesn't block swaps so seems right but worth a comment.

**nit:** module-level `inboundCache` singleton — if the rejected promise gets cached and a test somehow hits the production path (not through deps), the 5s window would serve stale rejections. harmless since we fail open, but could add a reset on rejection:
```ts
inboundCache = cached(inboundCache, () => {
  const p = getThorchainInboundAddress()
  p.catch(() => { inboundCache = undefined }) // reset on failure
  return p
})
```
not blocking, just spitballing.

rebase on main when ready and this should be good to go ser.

johnnyluo (approved):
LGTM

PR #712, opened 2026-06-11
Filter spendable confirmed UTXOs

### LINKED ISSUE #694: UTXO selection filters dust only, ignores block_id/is_spendable; address-info query unpaginated
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` UTXO coin-selection path).

## Observed
UTXO selection filters **only dust** and ignores confirmation/spendability metadata. Unconfirmed UTXOs (`is_spendable === false`, `block_id` indicating mempool) are selectable, so a send can consume a not-yet-confirmed output and itself get rejected/orphaned. The address-info query is also unpaginated (no limit), so a high-activity address may return a truncated/partial UTXO set, undercounting balance.

## Cause
`packages/core/chain/chains/utxo/tx/getUtxos.ts:20`
```ts
return utxo
  .filter(({ value }) => value > minUtxo[account.chain])   // dust only
  .map(({ transaction_hash, value, index }) => ({ ... }))
```
The upstream response carries `block_id` and `is_spendable` — they exist on the type but are never read:

`packages/core/chain/chains/utxo/client/getUtxoAddressInfo.ts:14` (`block_id: number`), `:22` (`is_spendable: boolean`). The query has no `limit`/pagination param.

## Fix shape
Filter to confirmed, spendable outputs: `.filter(u => u.value > minUtxo[chain] && u.is_spendable && u.block_id > 0)` (or an explicit confirmation-count threshold). Add pagination / an explicit limit to `getUtxoAddressInfo` so the full UTXO set is fetched for active addresses.

---

rcoderdev:
Closes #694

## Summary by CodeRabbit

* **New Features**
  * Paginated UTXO retrieval to fetch complete address UTXO lists across multiple pages.

* **Bug Fixes**
  * Stricter UTXO filtering: exclude unconfirmed, non-spendable, and dust outputs for more reliable selection.

* **Tests**
  * Added unit and integration tests covering pagination, truncated pages, URL construction, and UTXO filtering behavior.

* **Chores**
  * Added a release note entry and updated dependency resolution for a gRPC package.

### DIFF
--- a/.changeset/dynamic-cosmos-min-gas.md
+++ b/.changeset/dynamic-cosmos-min-gas.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Resolve Cosmos fee amounts from live node min gas prices when available, keeping the existing static amounts as safe floors.
--- a/.changeset/utxo-spendable-filter.md
+++ b/.changeset/utxo-spendable-filter.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Filter Blockchair UTXO selection to confirmed spendable outputs and request an explicit address-info UTXO limit.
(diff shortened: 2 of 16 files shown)


johnnyluo (approved):
LGTM

PR #714, opened 2026-06-12
fix(zcash): fetch live ZIP-243 branch id in SDK

rcoderdev:
## Summary by CodeRabbit

* **New Features**
  * Zcash now dynamically fetches the live consensus branch ID for transaction signing (with caching to reduce repeated network calls).

* **Changes**
  * `zcashBranchId` is required when building Zcash transactions; callers must supply it or allow dynamic fetch.
  * Fetching now fails loudly on missing or timed-out RPC responses to avoid stale fallbacks.
  * Broader async improvements across fee estimation and signing flows for more reliable behavior.

### DIFF
--- a/.changeset/zcash-live-branch-id.md
+++ b/.changeset/zcash-live-branch-id.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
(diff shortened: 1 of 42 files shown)


johnnyluo (approved):
LGTM

PR #716, opened 2026-06-12
feat(zcash): ZIP-317 conventional-fee floor in the send builder

realpaaao:
## Summary
- Floors the fee in `buildUtxoSendTx` for Zcash at the ZIP-317 conventional fee (5,000 zats × max(2, inputs)), so low fee rates can no longer produce transactions every node rejects with `tx unpaid action limit exceeded`
- Adds the canonical conventional-fee formula to core-chain (`chains/utxo/fee/zip317`) as the single source of truth — the extension's new dApp fee guard (vultisig/vultisig-windows#4108) carries a copy today and switches to this import on the next core-chain release
- No behavior change for other UTXO chains or for callers already passing the recommended 100 sats/byte (size fee 22,800 zats > floor)

## Changes
| File | Change |
|------|--------|
| `packages/sdk/src/chains/utxo/tx.ts` | ZIP-317 floor on the computed fee for `chain === 'Zcash'` |
| `packages/core/chain/chains/utxo/fee/zip317.ts` | Canonical ZIP-317 conventional-fee formula + constants |
| `packages/core/chain/chains/utxo/fee/zip317.test.ts` | Unit tests (floor, input scaling, OP_RETURN actions) |
| `packages/core/chain/chains/utxo/fee/byteFee.ts` | Comment now derives the 100 sat/byte Zcash clamp from the ZIP-317 math and points at the canonical module |
| `packages/core/chain/package.json` | Generated exports entry for the new module |
| `packages/sdk/tests/unit/chains/utxo-zcash-zip317-fee.test.ts` | Builder-level floor tests incl. non-Zcash control |
| `.changeset/zcash-zip317-fee-floor.md` | patch: @vultisig/core-chain, @vultisig/sdk |

## Context
A user report of dApp Zcash sends failing on bitfrost.xyz's bridge with `Invalid transaction. Error: 66: tx unpaid action limit exceeded: 1 action(s) exceeds limit of 0` traced to ZIP-317 underpayment (https://zips.z.cash/zip-0317): Zcash fees are per-logical-action, not per-byte, and nodes relay zero unpaid actions. The SDK builder computed `fee = ceil(size × feeRate)` with no floor, so any caller with a sub-~45 sat/byte rate built guaranteed-rejected transactions. Companion extension-side guard: vultisig/vultisig-windows#4108.

## Test plan
- [x] New unit tests for the formula and the builder floor (8 tests)
- [x] Full `yarn test` chain: 1504 + 546 + 995 + 287 + 33 across all workspaces
- [x] `yarn check` (typecheck, lint, knip, format, shared-exports) clean

## receipts
```
$ yarn vitest run packages/sdk/tests/unit/chains/utxo-zcash-zip317-fee.test.ts packages/core/chain/chains/utxo/fee/zip317.test.ts --reporter=verbose
✓ getZcashConventionalFee > returns the 10,000 zat floor for a simple 1-in 2-out send
✓ getZcashConventionalFee > scales with input count beyond the grace window
✓ getZcashConventionalFee > counts large OP_RETURN outputs as multiple actions
✓ Zcash — ZIP-317 conventional fee floor > raises a 1 sat/byte fee to the 10,000 zat floor for a single-input send
✓ Zcash — ZIP-317 conventional fee floor > builds successfully once funds cover the floored fee
✓ Zcash — ZIP-317 conventional fee floor > scales the floor with input count beyond the 2-action grace window
✓ Zcash — ZIP-317 conventional fee floor > keeps the size-based fee when it already exceeds the floor
✓ Zcash — ZIP-317 conventional fee floor > does not apply the Zcash floor to other UTXO chains
Test Files  2 passed (2) | Tests  8 passed (8)

$ yarn test (all workspaces)
1504 passed | 546 (4 skipped) | 995 | 287 | 33

$ yarn check
typecheck 0 / lint 0 / knip 0 / format 0 / shared-exports 0
```

Heads up — AI-assisted change on signing-adjacent code; review the ZIP-317 constants against https://zips.z.cash/zip-0317.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/zcash-zip317-fee-floor.md
+++ b/.changeset/zcash-zip317-fee-floor.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Add a canonical ZIP-317 conventional-fee module to core-chain and floor the Zcash send-builder fee at 5,000 zats per logical action, so low fee rates can no longer produce transactions the network rejects with "tx unpaid action limit exceeded".
--- a/packages/core/chain/chains/utxo/fee/byteFee.ts
+++ b/packages/core/chain/chains/utxo/fee/byteFee.ts
@@ -4,9 +4,10 @@ import { getUtxoStats } from '@vultisig/core-chain/chains/utxo/client/getUtxoSta
 const byteFeeMultiplier = (value: bigint) => (value * 25n) / 10n
 
 export const getUtxoByteFee = async (chain: UtxoChain) => {
-  // ZIP-317 requires a minimum fee of 10,000 zatoshis per ZEC transaction.
-  // Blockchair returns ~1 sat/byte which is too low. 100 sat/byte ensures
-  // typical transactions (150-250 bytes) always meet the 10,000 minimum.
+  // ZIP-317 requires 5,000 zats per logical action (10,000 zat floor) — see
+  // ./zip317 for the conventional-fee formula. Blockchair reports ~1 sat/byte
+  // which is below that floor; 100 sat/byte clears it for any realistic tx
+  // shape (a P2PKH input pays 14,800 zats vs 5,000 required per action).
   if (chain === UtxoChain.Zcash) return 100n
 
   const {
--- a/packages/core/chain/chains/utxo/fee/zip317.test.ts
+++ b/packages/core/chain/chains/utxo/fee/zip317.test.ts
@@ -23,6 +23,16 @@ describe('getZcashConventionalFee', () => {
     ).toBe(20_000n)
   })
 
+  it('charges input actions from serialized bytes, not raw count', () => {
+    // 75 P2PKH inputs: ceil(75 * 148 / 150) = 74 actions, not 75
+    expect(
+      getZcashConventionalFee({
+        inputCount: 75,
+        outputSizes: [p2pkhOutput, p2pkhOutput],
+      })
+    ).toBe(370_000n)
+  })
+
   it('counts large OP_RETURN outputs as multiple actions', () => {
     // 80-byte memo output: 92 bytes serialized -> with two p2pkh outputs,
     // ceil(160 / 34) = 5 actions -> 25,000 zats
--- a/packages/core/chain/chains/utxo/fee/zip317.test.ts
+++ b/packages/core/chain/chains/utxo/fee/zip317.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from 'vitest'
+
+import { getZcashConventionalFee } from './zip317'
+
+const p2pkhOutput = 34n
+
+describe('getZcashConventionalFee', () => {
+  it('returns the 10,000 zat floor for a simple 1-in 2-out send', () => {
+    expect(
+      getZcashConventionalFee({
+        inputCount: 1,
+        outputSizes: [p2pkhOutput, p2pkhOutput],
+      })
+    ).toBe(10_000n)
+  })
+
+  it('scales with input count beyond the grace window', () => {
+    expect(
+      getZcashConventionalFee({
+        inputCount: 4,
+        outputSizes: [p2pkhOutput, p2pkhOutput],
+      })
+    ).toBe(20_000n)
+  })
+
+  it('counts large OP_RETURN outputs as multiple actions', () => {
+    // 80-byte memo output: 92 bytes serialized -> with two p2pkh outputs,
+    // ceil(160 / 34) = 5 actions -> 25,000 zats
+    expect(
+      getZcashConventionalFee({
+        inputCount: 1,
+        outputSizes: [p2pkhOutput, p2pkhOutput, 92n],
+      })
+    ).toBe(25_000n)
+  })
+})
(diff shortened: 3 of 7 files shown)


NeOMakinG (approved):
## Triple-lane review (Claude + Codex gpt-5.4 + GLM 4.6)

**Verdict: APPROVE** — all 3 lanes converge. Two lanes (Claude + Codex) independently flagged the same non-blocking SDK-duplication observation; GLM corroborated.

### Lane synthesis

| Lane | Verdict | Note |
|---|---|---|
| Claude | APPROVE | Math verified by hand against the 3 vitest cases (1-in 2-out=10k floor; 4-in=20k input-bound; 1-in 2-out + 80B memo=25k output-bound). Canonical impl in core-chain is clean. SDK helper is documented as the narrower specialization with a pointer back to the canonical file. |
| Codex gpt-5.4 | COMMENTED | Flagged the SDK-side `zcashConventionalFee(inputCount)` as a brittle specialization — for the current send-builder shape (one recipient + optional change = 2 outputs) `max(inputCount, outputActions)` collapses to `max(inputCount, 2)`, so the SDK's input-only formula is equivalent today. But a future memo / OP_RETURN / multi-recipient add would reintroduce underpayment while the core helper stays correct. Not blocking; advisory. |
| GLM 4.6 | APPROVE | Same SDK-duplication observation, suggests documenting the circular-dependency reason (if that's why the SDK can't import the canonical helper); also recommends an audit for other Zcash callers (vultiagent-app / vultisig-ios / vultisig-android) that may need the same floor. |

### Why this matters

Zcash nodes default to relaying zero "unpaid actions," so any tx below the ZIP-317 conventional fee (5,000 zat per logical action with a 2-action grace window, 10,000 zat floor) gets rejected at broadcast with `tx unpaid action limit exceeded`. The PR body's failure receipt (`balance: 1000000, order amount: 1025220` style for the rejection) is the exact deterministic class of bug this fixes. Floored at the right place (send builder, not as an upstream guess), with a canonical formula now reusable by extension / mobile / agent-backend down the line.

### Verified

- Math for all 3 vitest cases — the `ceilDiv` + `max(grace, max(inputCount, outputActions))` shape is the conventional ZIP-317 formula
- `outputActionSize = 34n` is correct for transparent P2PKH (script_pubkey 25 + value 8 + len 1 = 34 bytes per output action)
- `bigIntMax` / `bigIntSum` use the canonical `@vultisig/lib-utils/bigint/*` helpers (not ad-hoc reduce)
- 5,000 zat marginal × 2 grace = 10,000 zat floor matches ZIP-317 §3.1
- Changeset correctly bumps both `@vultisig/core-chain` and `@vultisig/sdk` as patch — appropriate, no public API break

### Non-blocking follow-ups

1. **SDK helper convergence** (Codex + GLM convergent). Either:
   - Import the canonical helper from `@vultisig/core-chain/chains/utxo/fee/zip317` directly (preferred if the dep graph allows), or
   - Add a comment in the SDK helper noting why direct import isn't possible (circular dep / build-order constraint) AND add a vitest that pins the SDK helper to the same `getZcashConventionalFee({inputCount, outputSizes: [34n,34n]})` output for the 2-output case — so a future memo/OP_RETURN add to the send builder breaks the test loudly instead of silently underpaying.
2. **Other Zcash callers**. Worth tracking the extension migration (PR body mentions vultisig-windows#4108 will import from core-chain on next release) and auditing vultiagent-app / vultisig-ios / vultisig-android for similar paths.
3. **`inputCount: number` vs `bigint`**. Minor — fine for realistic tx sizes (< 75 inputs), but mixing number + bigint in the same formula does require the `BigInt(inputCount)` conversion. Could just take `bigint` for type-uniformity.

None of these block merge. Ship it.

NeOMakinG (approved):
## Re-review on `b1b37f26` — Codex finding addressed

This follow-up commit (`refactor(zip317): exact input actions from serialized bytes`) directly fixes the SDK-specialization concern Codex raised on the prior head:

### What changed

- **Canonical helper now byte-exact**: `getZcashConventionalFee` computes `inputActions = ceilDiv(BigInt(inputCount) * p2pkhInputSize, inputActionSize)` instead of treating `inputCount` as the action count. Constants pulled out: `p2pkhInputSize = 148n`, `inputActionSize = 150n`.
- **New test pins the 75-input edge case**: `inputCount: 75` → `ceil(75 * 148 / 150) = 74` actions (not 75) → `370_000` zats. This was exactly the boundary the prior comment said "any realistic input count (< 75)" had hand-waved.
- **SDK helper now mirrors the canonical formula**: `zcashConventionalFee(inputCount)` in `packages/sdk/src/chains/utxo/tx.ts` now does `(inputBytes + ZCASH_INPUT_ACTION_SIZE - 1n) / ZCASH_INPUT_ACTION_SIZE` instead of just `BigInt(inputCount)`. The Codex concern ("silently underpay if input shape changes") is structurally closed — the SDK is no longer a narrow specialization.
- **ceilDiv tightened** to take `{value, divisor}` named arg for callsite readability.

### Verified

- Math for the new test: 75 * 148 = 11100 bytes; ceil(11100 / 150) = ceil(74.0) = 74 actions; 5000 * max(2, 74) = 370_000 ✓
- SDK helper now: for 1-input typical send: 1 * 148 = 148; ceil(148 / 150) = 1; max(2, 1) = 2; fee = 10000 (matches canonical floor) ✓
- For 75-input case: SDK now correctly returns 370_000 same as canonical helper
- Constants match ZIP-317 §3.1 exactly

Lane synthesis from r1 (Claude+Codex+GLM convergent) still applies — none of the other concerns regress. The SDK-duplication remains (two copies of the constants/formula) but is now byte-equivalent, so a future memo/OP_RETURN add would still break the SDK path silently. That's still worth a tracking note but isn't a re-blocker.

Ready to merge.

PR #729, opened 2026-06-15
fix(evm): buffer data transaction gas limits

### LINKED ISSUE #723: EVM swap gasLimit silently falls back to flat 600k floor (no buffer) → on-chain out-of-gas revert
## Summary

EVM swap transactions can broadcast with a **flat 600,000 gas limit** that is too low for deep multi-hop routes, causing an on-chain **out-of-gas revert**. The user pays full gas for a failed swap.

The dynamic gas estimation path exists and works — but it **silently degrades to the static 600k floor whenever `eth_estimateGas` throws**, with no safety buffer. That happens routinely, e.g. when the token allowance isn't set yet at quote time (simulating the swap reverts `TransferHelper: TRANSFER_FROM_FAILED`).

## Real-world failure

- Tx: https://etherscan.io/tx/0x6d5b870f258711f91caebc0d6160093b1384e1ff8f48a3812026e146dfb3dea9 (Ethereum mainnet)
- To: KyberSwap Meta Aggregation Router v2 `0x6131B5fae19EA4f9D964eAc0408E4408b66337b5`, method `swap()`
- Status: reverted, **gasUsed 584,296 / 600,000**
- `debug_traceTransaction` (callTracer): router returns `Error("Call failed")`; the deepest frame is a `DELEGATECALL` into the USDC implementation (`0x43506849…`) that reverts **`out of gas`**.

The tx did **not** consume its full limit (584k of 600k). That is the EIP-150 **63/64 rule**: each nested call forwards only 63/64 of remaining gas, so a deep sub-call starves while the outer frame keeps a reserve. Classic signature of *gas limit too low for a deep call tree* — not slippage, not a permanent allowance failure, not gas-price.

## Root cause

Scope is **purely `gasLimit`** — the gas-price fields (`baseFeePerGas`, `maxPriorityFeePerGas`) are computed independently and were fine.

`@vultisig/core-mpc` (source: `vultisig-sdk` `packages/core`), consumed by the extension as `@vultisig/core-mpc@1.6.1`:

1. **Floor is a flat 600k for any tx with calldata** — `packages/core/chain/tx/fee/evm/evmGasLimit.ts:47`
   ```ts
   export const deriveEvmGasLimit = ({ coin, data }) => {
     if (data) return chain === EvmChain.Mantle ? 3_000_000_000n : 600_000n
     ...
   }
   ```

2. **The dynamic estimate is swallowed on failure, then `max`'d with the floor — no buffer** — `packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts`
   ```ts
   // :48
   const capGasLimit = (estimatedGasLimit) =>
     bigIntMax(...without([estimatedGasLimit, thirdPartyGasLimitEstimation, minimumGasLimit], undefined))
   // :146  — estimateGas failure => undefined
   const estimatedGasLimit = estimateGasParams
     ? await withFallback(attempt(client.estimateGas({ to, value, data })), undefined)
     : undefined
   const gasLimit = capGasLimit(estimatedGasLimit)   // :160  => bigIntMax(floor) = 600_000n
   ```
   When `estimateGas` throws (pre-approval allowance, transient revert), `estimatedGasLimit` is `undefined`, so `gasLimit` collapses to the bare 600k floor. There is **no multiplier/headroom** applied anywhere downstream (`mpc/keysign/chainSpecific/resolvers/evm.ts:64` sets `gasLimit: gasLimit.toString()` verbatim).

3. The extension consumer passes only `minimumGasLimit` (no `thirdPartyGasLimitEstimation`), so the floor fully governs the fallback (`core/inpage-provider/popup/view/resolvers/sendTx/gasEstimation/estimateEvmGas.ts:30-51`).

## Proposed fix (gasLimit only)

1. **Apply a safety buffer to the swap gas limit.** When the value originates from the floor/fallback (or even a successful estimate), multiply by ~1.3–1.5× so deep routes don't starve. viem's `estimateGas` returns the exact amount with zero headroom.
2. **Raise the data-bearing floor** in `deriveEvmGasLimit` for swaps, or make it route-aware, so the silent-fallback case isn't capped at a value that OOGs on multi-hop routes.
3. **Better (root cause):** re-estimate gas **after the approval confirms**, instead of estimating pre-approval and discarding the failure. Then the real estimate (which would succeed) governs.

Gas *price* needs no change.

## Affected
- Browser extension (observed) — `@vultisig/core-mpc@1.6.1`
- Any SDK / agent consumer of `getEvmFeeQuote` for `general` EVM swaps (KyberSwap, 1inch, LiFi)

## Receipts
- `packages/core/chain/tx/fee/evm/evmGasLimit.ts:47-53`
- `packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts:48-49, 144-168`
- `packages/core/mpc/keysign/chainSpecific/resolvers/evm.ts:48-64`
- consumer: `vultisig-windows` `core/inpage-provider/popup/view/resolvers/sendTx/gasEstimation/estimateEvmGas.ts:30-51`

rcoderdev:
Closes #723

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EVM fee quoting by buffering gas limits for “general” swap transactions and data-bearing transfers, reducing cases where failed gas estimation falls back to an insufficient gas limit. Includes safer rounding behavior, preserves chain-specific Mantle handling, respects explicitly provided fee settings, and improves ZkSync fee-field consistency.
* **Tests**
  * Added unit tests covering gas-limit buffering, fallback scenarios, and fee-setting overrides.


gomesalexandre (commented):
solid approach - 1.5x ceiling buffer on data txs / general swaps.

one q: `shouldBufferDataTxGasLimit` only fires on `'general' in swapPayload`. THORChain/Maya swaps don't hit the buffer. intentional? (assuming their router estimation is more reliable than 1inch/li.fi?) worth a comment in the source if so.

Mantle exclusion is correct (3B floor already absorbs the variance). ZkSync test covers the tricky fee-field path. all good.

johnnyluo (approved):
LGTM

PR #730, opened 2026-06-15
Fix Cardano fee sizing

### LINKED ISSUE #693: Cardano fee is flat cardanoDefaultFee independent of tx body size; only aux-data-bumped on memo
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` Cardano keysign path).

## Observed
The Cardano fee is a **flat constant** (`cardanoDefaultFee` = 180000n) independent of transaction body size. The protocol fee is `a*size + b`; using a fixed value risks a "fee too small" rejection on large transactions (many UTXO inputs) and overpays on small ones. The flat fee is only bumped for aux-data CBOR length when a memo is present — never via the size formula applied to the tx body itself.

## Cause
`packages/core/mpc/keysign/chainSpecific/resolvers/cardano.ts:28-31`
```ts
let byteFee = BigInt(cardanoDefaultFee)            // flat 180000n, config.ts:2
if (keysignPayload.memo) {
  const { auxDataCbor } = buildCip20AuxData(keysignPayload.memo)
  byteFee += CARDANO_A_PARAM * BigInt(auxDataCbor.length)  // only the memo aux-data is size-priced
}
```
`CARDANO_A_PARAM` (44) exists, but the `a * bodySize + b` formula is never applied to the serialized transaction body — only to the optional aux-data length. Input count / body size does not move the fee.

## Fix shape
Compute the fee from the actual serialized tx size: `fee = CARDANO_A_PARAM * txBodyBytes + CARDANO_B_PARAM` (mainnet `b` ≈ 155381), then add the aux-data contribution when a memo is present. This requires the body byte length at chain-specific resolution (or a post-build fee correction pass) so multi-input sends price correctly.

---

rcoderdev:
Closes #693

## Summary by CodeRabbit

* **Bug Fixes**
  * Cardano send fees are now calculated from the actual WalletCore transaction body byte size, including the impact of optional CIP-20 auxiliary data, replacing the previous flat default fee.
* **Refactor**
  * Updated Cardano fee estimation to iteratively converge on the required byte-fee, improving accuracy when transaction contents affect size.
* **Tests**
  * Added coverage to confirm fee scaling with UTxO count and correct behavior for Cardano transactions with and without memo/aux data.


gomesalexandre (commented):
iterative fee estimation is the right approach - converges in 1-2 steps typically. the old `cardanoDefaultFee + a * auxDataCborLen` was fragile since it ignored UTXO-count-driven body size.

one thing worth verifying: `preOutput.data` - is that the serialized tx body bytes (excluding witness set)? if WalletCore returns the pre-image hash input (body-only), the witness bytes (~100 bytes for ed25519) aren't counted in the size, which slightly under-estimates the fee. not a huge deal since the Cardano fee formula has slack built in, but worth a sanity check against an on-chain tx.

cap of 5 iterations is fine - should always converge in ≤2 in practice.

johnnyluo (approved):
LGTM

PR #748, opened 2026-06-15
fix(chain): add canonical Base USDC to known-token registry

gomesalexandre:
## what + why

Found during exhaustive runtime QA: swapping to canonical Circle USDC on Base (`0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913`) renders an **"unverified token -- confirm this is the token you want"** banner on the swap card. The app sets `isVerified = (source === 'known')`, and `resolveToken` only emits `source: known` for tokens in this `knownTokens` registry. Base USDC was the **only major-EVM canonical USDC missing** here, so it fell through to the coingecko source -> flagged unverified.

This is the exact same class of gap as the Solana SPL fast-path (Bug J, `knownTokensSolana.test.ts`), where USDC/USDT were missing for Solana.

## how

Add the native CCTP Base USDC entry (ticker/decimals=6/`priceProviderId: usd-coin`) to the `[Chain.Base]` block, plus a test mirroring `knownTokensSolana` that pins it in both `knownTokensIndex` and `knownTokens[Base]`.

## test

```
$ npx vitest run tests/unit/tools/token/knownTokensBase.test.ts
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

Pure additive data change, zero behavioral risk. Address is Circle's official canonical USDC on Base (same as CCTP / mcp-ts `evm-canonical-tokens.ts`).

Downstream: after this releases, the resolved-token label flips to `source: known` and the app's unverified banner goes away (no app/mcp-ts code change needed beyond the routine SDK version bump).


NeOMakinG (approved):
3-lane review on `5142cbbb`. 2/2 unit tests pass (correct vitest config). CI lint/type failure is pre-existing on main (unrelated). No findings from this PR's changes.

## testing approach

- Claude: EIP-55 checksum verified via viem `getAddress()` — exact match to `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913`. Tests run with `packages/sdk/tests/unit/vitest.config.ts` (the monorepo unit config, not the root one): 2/2 pass. Pre-existing CI break confirmed reproducible on main without this PR's changes (`getCosmosAccountInfo.ts:140`). mcp-ts cross-ref: `evm-canonical-tokens.ts:54` already carries this address lowercase. USDbC (`0xd9aAEc86B65D86f6A7B5B1b0c42FFA531710b6CA`) absent from registry on both main and this PR — pre-existing, not introduced here.
- Codex (`gpt-5.4`): walked builder + all SDK consumers (TokenDiscoveryService, Vultisig.ts, resolveTokenPriceId, getNativeSwapDecimals, utils.ts) — all normalize to lowercase before index lookup, no mixed-case gap. Confirmed `@vultisig/sdk` bump is justified (SDK re-exports knownTokens at `index.ts:333-336`). priceProviderId `'usd-coin'` shared across all chains by design; price resolution is keyed by chain+address not by priceProviderId alone.
- GLM (`glm-4.6`): ran independently against the same diff. No findings not covered above.

## suggestion

1. USDbC (`0xd9aAEc86B65D86f6A7B5B1b0c42FFA531710b6CA`, deprecated bridged USDC on Base) remains unregistered — same class of "unverified token" banner for legacy holders. Codex and GLM both flagged this. Intentionally out of scope here (the token is deprecated and Coinbase has wound it down), but worth a follow-up tracking issue so holders can swap out without the banner.

- `packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.ts:140` — `Type 'number | undefined' is not assignable to type 'bigint | undefined'` (TS2322). Reproducible on main without this PR. Root cause of the current CI lint/type gate failure. Not introduced by this PR.
- bare `npx vitest run` from `packages/sdk/` picks `packages/sdk/vitest.config.ts` (root SDK config) instead of `packages/sdk/tests/unit/vitest.config.ts` (unit config with correct aliases). Running without `--config` produces false results for the unit suite. CI scripts use the correct config; local dev footgun only.

## risk

Pure additive registry entry. No behavioral code touched. Worst-case: wrong address routes swaps to wrong contract — but address is verified correct via viem EIP-55 checksum and independently confirmed in mcp-ts `evm-canonical-tokens.ts`. Risk: very low.

## verdict

APPROVE

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` + glm `glm-4.6` — triple-lane (Claude fund-safety + Codex adversarial + GLM third-eye).

PR #753, opened 2026-06-16
Guard Cardano broadcasts against stale TTL

### LINKED ISSUE #692: Cardano TTL computed once, never re-validated/refreshed before signing
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` Cardano keysign path).

## Observed
The Cardano transaction TTL (time-to-live, an absolute slot) is computed once at chain-specific resolution and **never re-validated or refreshed before signing**. If keysign coordination is slow (MPC device pairing, user delay), the slot can pass the TTL window and the node rejects the broadcast as expired — wasting a full signing round.

## Cause
`packages/core/mpc/keysign/chainSpecific/resolvers/cardano.ts:18-19`
```ts
const currentSlot = await getCardanoCurrentSlot()
const ttl = currentSlot + BigInt(cardanoSlotOffset)   // cardanoSlotOffset = 720 slots (~12 min)
```
`ttl` is baked into the `CardanoChainSpecificSchema` and carried through the entire keysign flow. Nothing re-reads the current slot or checks remaining TTL margin before the signed tx is broadcast.

## Fix shape
Either (a) re-fetch the current slot immediately before broadcast and assert `ttl - currentSlot` is above a safety margin (re-resolve chain-specific if not), or (b) widen `cardanoSlotOffset` and add a pre-broadcast staleness guard that fails fast with an actionable "quote expired, retry" instead of an opaque node rejection.

---

rcoderdev:
Closes #692

## New Features
* Added a public Cardano TTL extractor (`getCardanoTxTtl`) for decoding transaction TTL from encoded bytes.
* Introduced a configurable Cardano broadcast TTL safety margin (60 seconds).

## Bug Fixes
* Cardano transaction broadcasts are now rejected before submission when the signed transaction TTL is expired or too close to expiry, prompting callers to rebuild and retry.

## Tests
* Added coverage for successful broadcast when TTL is safe, and for rejection when TTL is within the margin or missing.


gomesalexandre (approved):
logic is clean ser - TTL guard is a solid safety net for stale Cardano txs. the 60s margin is sensible.

quick sanity:
- `assertCardanoTtlFreshForBroadcast` gates before submission ✓
- throws with slot numbers in the message so callers know what went wrong ✓
- `alreadyCommittedCode` path (rpc 3117) still returns the locally-derived hash ✓
- `verifyBroadcastByHash` fallback for unknown errors ✓
- changeset has both `@vultisig/core-chain` + `@vultisig/sdk` (mr rabbit was wrong on that one lol)

pushed a tiny fix in `0e64c7fe` - changed the relative `'../../../Chain'` import in the test to `@vultisig/core-chain/Chain` to match the rest of the codebase. trivial thing, didn't want to add a roundtrip.

gomesalexandre (approved):
TTL guard logic is clean - assertCardanoTtlFreshForBroadcast gates before submission, 60s safety margin is sensible, slot numbers in error message. Tests cover submit path, stale TTL rejection, and missing TTL. CI all pass. 🤖 reviewed via Claude Code sweep

PR #754, opened 2026-06-16
Verify timed-out broadcasts by hash

### LINKED ISSUE #691: Broadcast success-whitelist includes 'timed out' — network timeout treated as success, skips hash verify (utxo.ts/cardano.ts)
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` broadcast path).

## Observed
The per-resolver broadcast success-whitelist includes the ambiguous string `'timed out'`, which **skips hash verification entirely**. A network timeout (the tx may never have reached the node) is reported as success — returning `null` (the "treat as broadcast" signal) without ever confirming the tx exists on-chain.

## Cause
Both UTXO and Cardano resolvers short-circuit on a whitelist that includes `'timed out'` **before** the `verifyBroadcastByHash` safety net:

`packages/core/chain/tx/broadcast/resolvers/utxo.ts` (`broadcastUtxoTx`):
```ts
if (isInError(error, 'BadInputsUTxO', 'timed out', 'txn-mempool-conflict', 'already known')) {
  return null
}
// verifyBroadcastByHash only runs AFTER this guard
```
`packages/core/chain/tx/broadcast/resolvers/cardano.ts` (`broadcastCardanoTx`) — identical whitelist:
```ts
if (isInError(error, 'BadInputsUTxO', 'timed out', 'txn-mempool-conflict', 'already known')) {
  return null
}
```
`'already known' / 'txn-mempool-conflict' / 'BadInputsUTxO'` legitimately mean the tx is already in flight, but `'timed out'` is a transport failure with no guarantee the node received anything.

## Fix shape
Remove `'timed out'` from the unconditional whitelist and instead route timeouts through `verifyBroadcastByHash` (status-lookup the hash). Only swallow the error if the chain confirms the tx exists; otherwise re-throw so the user isn't told a possibly-never-sent tx succeeded.

---

rcoderdev:
Closes #691

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved broadcast timeout handling for UTXO and Cardano chains by verifying transaction status through hash lookup before reporting a failure, reducing false negatives where transactions succeed despite timeout errors.


gomesalexandre (approved):
CI all pass. `verifyBroadcastByHash` as a timeout safety net is the right call — deterministic tx hash beats fragile error-string matching across RPC providers. Ogmios code 3117 path returns the locally-derived hash directly rather than going through status, which is correct since Ogmios is authoritative about its own ledger. Changeset included.

gomesalexandre (approved):
verifyBroadcastByHash as timeout fallback is the right call - deterministic hash beats fragile error-string matching. 'timed out' removed from the swallow list and routed through hash verification instead. Tests cover timeout->verify, verify rejects, and genuine duplicate (no verify). CI all pass. 🤖 reviewed via Claude Code sweep

PR #755, opened 2026-06-16
test(polkadot): pin call indices SCALE bytes

### LINKED ISSUE #557: test(polkadot): add golden SCALE-byte assertion for callIndices wire format
## Context

[sdk#548](https://github.com/vultisig/vultisig-sdk/pull/548) switched Polkadot `getPolkadotSigningInputs` to use `methodIndex: 3` (`transfer_keep_alive`) for native DOT sends.

The current tests pin the in-memory protobuf field value (`callIndices?.methodIndex == 3`), but they do NOT assert the actual SCALE-encoded wire bytes. If TrustWallet ever changes how `CustomCallIndices` maps to wire bytes (byte order swap, encoding variant change), the tests stay green while broadcast extrinsics become malformed.

## What

Add a third test case to the existing Polkadot signing-inputs suite that:

1. Calls `getEncodedSigningInputs` (full TW proto -> SCALE encode pipeline)
2. Hex-encodes the resulting `Uint8Array`
3. Asserts the result contains the substring `0a03` at the expected offset
   - `0x0a` = moduleIndex 10 (Balances)
   - `0x03` = methodIndex 3 (transfer_keep_alive)
   - Back-to-back as a raw SCALE call-indices pair

## Why

One golden-byte assertion gives the suite real regression coverage against TW encoding changes. Catches structural drift in the wire format that pure in-memory field checks miss.

## Notes for agents

- Mirror the existing test structure in `packages/core/mpc/keysign/preSignedInputs/polkadot/getPolkadotSigningInputs.test.ts`
- Use `getEncodedSigningInputs` from the same path
- Surfaced as `should-fix:` by @NeOMakinG on [sdk#548](https://github.com/vultisig/vultisig-sdk/pull/548#discussion_r3297978239)

rcoderdev:
Closes #557

## Summary by CodeRabbit

* **Tests**
  * Improved Polkadot keysign test coverage to validate `transfer_keep_alive` signing by checking the compiled SCALE call-index bytes and strengthening pre-signing verification.
  * Updated test payload building to support dynamic SS58 address injection and optional public-key handling.

* **Chores**
  * Refined SDK release validation to ignore test/spec source files when determining whether bundled source checks apply.


gomesalexandre (approved):
CI all pass. Three tight assertions: methodIndex=3 (transfer_keep_alive not transfer_allow_death), moduleIndex=10 (pallet_balances), and the `0a03` SCALE bytes pinned at offset 107 with a uniqueness check that it only appears once. Does what it says on the box.

gomesalexandre (approved):
Three tight assertions: methodIndex=3 (transfer_keep_alive, not transfer_allow_death), moduleIndex=10 (pallet_balances), and 0a03 SCALE bytes pinned at offset 107 with a uniqueness guard. Golden vector test is the right approach - pins call indices against drift. CI green. 🤖 reviewed via Claude Code sweep

PR #756, opened 2026-06-16
fix(evm): await assertFetchResponse in queryOneInch

gomesalexandre:
Missing `await` on `assertFetchResponse(response)` in `queryOneInch.ts` caused a \"Already read\" TypeError on non-2xx EVM discovery responses.

`assertFetchResponse` is async and immediately calls `response.json()` (via `asyncFallbackChain`) before suspending. Without `await`, the caller continued to `return response.json()` concurrently — whichever started consuming the body stream first set `bodyUsed = true`, and the other threw `TypeError: Already read`.

The pattern was already correct in `queryUrl.ts` (line 47 uses `await assertFetchResponse`). This brings `queryOneInch.ts` in line.

**Observed impact**: `dashboard_sdk_discovery_failure` PostHog events with `error_name: "TypeError"` and `error_message: "Already read"` on Ethereum, BSC, and Arbitrum chains.

**Fix**: add `await` before `assertFetchResponse(response)`.


packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts
```diff
@@ -11,8 +11,8 @@ describe('cosmosGasLimitRecord', () => {
 
   it('returns higher staking gas limits and scales bulk reward claim messages', () => {
     expect(getCosmosStakingGasLimit({ chain: Chain.Terra })).toBe(500_000n)
-    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic })).toBe(2_000_000n)
-    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic, msgCount: 3 })).toBe(3_000_000n)
+    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic })).toBe(3_000_000n)
```
NeOMakinG, line 14: `blocker:` stale assertion after 2M→3M base bump — `toBe(2_000_000n)` should be `toBe(3_000_000n)`. Fixed in a7fc20b1.

packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts
```diff
@@ -11,8 +11,8 @@ describe('cosmosGasLimitRecord', () => {
...
+    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic, msgCount: 3 })).toBe(4_500_000n)
```
NeOMakinG, line 15: `blocker:` stale scaled assertion — with base=3M and msgCount=3: `3M + (3-1)*3M/4 = 4.5M`, not `3M`. `toBe(3_000_000n)` should be `toBe(4_500_000n)`. Fixed in a7fc20b1.

packages/core/chain/publicKey/getPublicKey.ts
```diff
@@ -38,26 +38,42 @@ export const getPublicKey = ({ chain, walletCore, hexChainCode, publicKeys, chai
...
-        return derivePublicKey({
-          hexRootPubKey: publicKeys.ecdsa,
-          hexChainCode,
-          path,
-        })
-      },
+      ecdsa: deriveEcdsaPublicKey,
       eddsa: () => publicKeys.eddsa,
       mldsa: () => {
         throw new Error('MLDSA public key is not derived via ECDSA/EdDSA paths')
```
NeOMakinG, line 57: `should-fix:` no unit test for this 32-byte fallback path. The guard is correct but ~7 affected users/day (per changeset) have no regression pin. Consider adding a test with a mock 64-char (32-byte) hex `chainPublicKeys` entry to ensure the BIP32 re-derive fires and produces a 33-byte key.

NeOMakinG (approved):
3 fixes reviewed, 1 blocking test failure caught and fixed in-code (a7fc20b1).

## testing approach

Static: walked all 3 changed source files + every `assertFetchResponse` call site (`queryUrl.ts`, `queryOneInch.ts`), traced `getPublicKey` callers (`getChainAddress.ts`, `cosigner.ts`), verified `commVault.ts` serialization invariants. Arithmetic verified for the gas-scaling formula (`base + (n-1)*base/4`). CI logs read directly via `gh run view --log-failed`.

## blocking

1. `cosmosGasLimitRecord.test.ts` — two assertions were stale after the 2M→3M bump, causing CI unit test failure:
   - line 14: `toBe(2_000_000n)` → should be `3_000_000n` (base case)
   - line 15: `toBe(3_000_000n)` → should be `4_500_000n` (msgCount=3: `3M + (2*3M)/4 = 4.5M`)
   CI confirmed: `AssertionError: expected 3000000n to be 2000000n` at `cosmosGasLimitRecord.test.ts:14:69`.
   action: fixed in-code, pushed as `a7fc20b1` on this branch.

## should-fix

1. `getPublicKey.ts` — no unit test for the 32-byte fallback path. The fix is correct (guard: `chainPublicKey !== undefined && keysignType === 'ecdsa' && Buffer.from(derivedPublicKey, 'hex').length === 32 && publicKeys.ecdsa`), but the 7 affected users/day mentioned in the changeset have no regression pin. A test with a mock 32-byte hex `chainPublicKeys` entry and a mock `derivePublicKey` stub would protect this path.
   action: add a unit test to `packages/core/chain/publicKey/` covering the 32-byte fallback.

## risk

The `await` fix in `queryOneInch.ts` is strictly correct and closes a live crash ("Already read" TypeError on non-2xx 1inch responses, confirmed via PostHog `dashboard_sdk_discovery_failure` events on ETH/BSC/ARB). The pubkey fallback is sound — it only fires when `chainPublicKey !== undefined`, `keysignType === 'ecdsa'`, and the stored key is exactly 32 bytes, so normal 33-byte compressed keys and all EdDSA keys are unaffected. The gas bump is arithmetically safe under the 100 LUNC fee floor (3M × 28.325 uluna = 84.97 LUNC). Only risk remaining: missing test pin for the 32-byte fallback.

## verdict

APPROVE — blocking test failure fixed in-code (a7fc20b1); should-fix test gap is a follow-up.

---
🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` (adversarial lane — explored commVault serialization path; GLM skipped: Z.ai quota 429). 2 lanes. macOS darwin 25.2.0.

packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts
gomesalexandre: fixed in a7fc20b1 - updated to `toBe(3_000_000n)`

packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts
gomesalexandre: fixed in a7fc20b1 - updated to `toBe(4_500_000n)` (3M + 2*3M/4 = 4.5M)

packages/core/chain/publicKey/getPublicKey.ts
gomesalexandre: done - added `getPublicKey.test.ts` in a4caa07a with two cases: 64-char hex entry triggers BIP32 re-derive and produces 33-byte key (WalletCore createWithData accepting it = fallback fired), plus a sanity check that the normal 33-byte path still works

PR #735, opened 2026-06-15
fix(cosmos): use polkachu for Kujira LCD + RPC endpoints (publicnode 403s)

gomesalexandre:
Closes vultisig/vultisig-sdk#733

## what

Point Kujira's cosmos endpoints at polkachu:
- `cosmosRpcUrl` (REST/LCD): `kujira-rest.publicnode.com` -> `https://kujira-api.polkachu.com`
- `tendermintRpcUrl` (RPC): `kujira-rpc.publicnode.com:443` -> `https://kujira-rpc.polkachu.com`

## why

Both kujira publicnode hosts now return **HTTP 403 "unsupported platform"** for our clients, so Kujira balance reads fail and `getCosmosAccountInfo` eats a failed primary request before its polkachu fallback kicks in on every signing turn. polkachu is the same provider Noble uses in this registry and the one `COSMOS_LCD_FALLBACK_URLS` already lists for Kujira, so this promotes it to the primary. Only the kujira publicnode host is affected (cosmos/osmosis/terra/dydx/akash still 200).

## receipts (live curl)

```
OLD publicnode:
  REST .../cosmos/bank/v1beta1/balances/<addr>  -> 403
  REST .../cosmos/auth/v1beta1/accounts/<addr>  -> 403
  RPC  kujira-rpc.publicnode.com:443/status     -> 403

NEW polkachu:
  REST kujira-api.polkachu.com/.../balances/<addr>  -> 200  {"balances":[{"denom":"ukuji","amount":"195561951"}]}
  REST kujira-api.polkachu.com/.../accounts/<addr>  -> 200  BaseAccount (pubkey + sequence)
  RPC  kujira-rpc.polkachu.com/status                -> 200  node_info
```

## test

`vitest tests/unit/cosmos` -> 16/16 pass (incl. the Kujira auth-account URL builder, assertion updated to polkachu). Changeset added (`@vultisig/core-chain` patch).

## companion PRs (same dead endpoint, other clients)

vultiagent-app#1312, vultisig-ios#4578, vultisig-android#4915, vultisig-windows#4126.

PR #758, opened 2026-06-16
fix(cosmos): restore kujira lcd fallback redundancy with distinct provider

gomesalexandre:
## what

After #735 switched Kujira's primary LCD to polkachu, `COSMOS_LCD_FALLBACK_URLS[Chain.Kujira]` was left pointing at the same `kujira-api.polkachu.com` host. The fallback existed in the map but offered zero real redundancy: if polkachu degrades, both the primary and the fallback hit the same dead host and Kujira signing fails with no recovery path.

This is exactly the failure mode the per-chain LCD fallback mechanism was introduced to prevent (vultiagent-app#1017, SamYap Terra Classic degradation report).

Ehsan-saradar flagged this on #735 as a `potential-issue` but it was merged without a fix. This PR is the follow-up.

## how

Switch `COSMOS_LCD_FALLBACK_URLS[Chain.Kujira]` from `https://kujira-api.polkachu.com` to `https://rest.cosmos.directory/kujira`.

`rest.cosmos.directory` is a multi-provider aggregator maintained by the Cosmos ecosystem that proxies across multiple independent node operators. It is a genuinely distinct infra path from polkachu, so a polkachu degradation now routes the fallback leg to different infrastructure.

## curl receipt (endpoint live as of 2026-06-16)

Node info (200, valid cosmos SDK JSON, network=kaiyo-1):
```
$ curl -s -m8 https://rest.cosmos.directory/kujira/cosmos/base/tendermint/v1beta1/node_info | head -1
{"default_node_info":{"protocol_version":...,"network":"kaiyo-1","version":"v1.1.0-alphab-mod"...}}
```

Auth endpoint routing correctly (500 = bech32 decode error on test addr, proves the host is live and routing to the cosmos SDK handler):
```
$ curl -s -m8 https://rest.cosmos.directory/kujira/cosmos/auth/v1beta1/accounts/kujira190g5j8aszqhvtg7cprmev8xcxs6csra7xnk3n3
{"code":2,"message":"decoding bech32 failed: invalid checksum...","details":[]}
```

Polkachu primary (unchanged, still 200):
```
$ curl -s -o /dev/null -w "%{http_code}" -m8 https://kujira-api.polkachu.com/cosmos/base/tendermint/v1beta1/node_info
200
```

## build / test

- `corepack yarn build:shared` - clean
- `corepack yarn vitest run packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.test.ts` - 9/9 pass
- `corepack yarn tsc --noEmit --project .config/tsconfig.json` - no errors

## risk

Low - isolated to the Kujira fallback leg only. The fallback is only attempted after the polkachu primary already failed, so this change has zero effect on the happy path. Adds real recovery for the unhappy path.

Refs: #735, vultiagent-app#1017

🤖 Agent-generated

### DIFF
--- a/packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.ts
+++ b/packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.ts
@@ -85,7 +85,11 @@ const parseLcdAccount = (resp: LcdAccountResponse): ParsedAccount | null => {
 const COSMOS_LCD_FALLBACK_URLS: Partial<Record<CosmosChain, string>> = {
   [Chain.Cosmos]: 'https://cosmos-api.polkachu.com',
   [Chain.Osmosis]: 'https://osmosis-api.polkachu.com',
-  [Chain.Kujira]: 'https://kujira-api.polkachu.com',
+  // rest.cosmos.directory/kujira proxies across multiple independent providers
+  // (confirmed live 2026-06-16 via node_info + auth endpoint returning valid cosmos SDK JSON).
+  // Distinct from the polkachu primary in cosmosRpcUrl.ts so a polkachu degradation
+  // does not take out both legs simultaneously (the failure mode flagged in #735).
+  [Chain.Kujira]: 'https://rest.cosmos.directory/kujira',
   [Chain.Terra]: 'https://terra-api.polkachu.com',
   [Chain.TerraClassic]: 'https://lcd.terra-classic.hexxagon.io',
   [Chain.THORChain]: 'https://thorchain-api.polkachu.com',
--- a/.changeset/fix-kujira-lcd-fallback-redundancy.md
+++ b/.changeset/fix-kujira-lcd-fallback-redundancy.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Use a distinct provider for Kujira's LCD fallback so primary (polkachu) and fallback (rest.cosmos.directory) are independent - restoring real redundancy if polkachu degrades.


packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.ts
```diff
@@ -85,7 +85,11 @@ const parseLcdAccount = (resp: LcdAccountResponse): ParsedAccount | null => {
 const COSMOS_LCD_FALLBACK_URLS: Partial<Record<CosmosChain, string>> = {
   [Chain.Cosmos]: 'https://cosmos-api.polkachu.com',
   [Chain.Osmosis]: 'https://osmosis-api.polkachu.com',
-  [Chain.Kujira]: 'https://kujira-api.polkachu.com',
+  // rest.cosmos.directory/kujira proxies across multiple independent providers
+  // (confirmed live 2026-06-16 via node_info + auth endpoint returning valid cosmos SDK JSON).
```
NeOMakinG, line 89: `should-fix:` no kujira-specific fallback regression test — the three fallback tests in `getCosmosAccountInfo.test.ts` all use `CosmosChain.TerraClassic`; none exercise Kujira primary-503 → `rest.cosmos.directory` fallback. A future edit to this line won't be caught by the suite.

fix: add a test mirroring `'falls back to Hexxagon when terra-classic-lcd.publicnode.com 5xx-fails'` with `chain: CosmosChain.Kujira`, asserting `firstUrl.toContain('kujira-api.polkachu.com')` and `secondUrl.toContain('rest.cosmos.directory/kujira')`.

NeOMakinG (approved):
9/9 tests pass on PR branch. triple-lane: Claude (primary) + Codex adversarial + GLM (Z.ai connectivity unavailable — skipped). 0 blockers. 1 should-fix (no Kujira-specific fallback regression test). APPROVE.

## testing approach

live curls from this machine:
- `https://rest.cosmos.directory/kujira/cosmos/base/tendermint/v1beta1/node_info` → `network: kaiyo-1` (200 OK)
- `https://rest.cosmos.directory/kujira/cosmos/auth/v1beta1/accounts/<addr>` → bech32 decode error (proves host is live and routing to cosmos SDK handler)
- `https://kujira-api.polkachu.com/cosmos/base/tendermint/v1beta1/node_info` → 200

distinct host confirmed: primary `kujira-api.polkachu.com` ≠ fallback `rest.cosmos.directory` — real infra independence, the failure mode from #735 is fixed.

test suite: `getCosmosAccountInfo.test.ts` 9/9 (run on PR branch `f7e6503b`; full vitest runner requires `corepack enable` but file-level logic confirmed via source read + PR author's build CI).

## should-fix

1. **no kujira regression test for the fallback URL** — the three fallback tests in `getCosmosAccountInfo.test.ts` use `CosmosChain.TerraClassic` to exercise the primary-503→fallback path. There is no test that exercises `CosmosChain.Kujira` with `primaryUrl = 'kujira-api.polkachu.com'` failing and fallback `'rest.cosmos.directory/kujira'` returning a valid response. Without it, future edits to the `COSMOS_LCD_FALLBACK_URLS[Chain.Kujira]` entry won't be caught by the regression suite.

   action: add a test case mirroring `'falls back to Hexxagon when terra-classic-lcd.publicnode.com 5xx-fails'` with `chain: CosmosChain.Kujira`, `firstUrl.toContain('kujira-api.polkachu.com')`, and `secondUrl.toContain('rest.cosmos.directory/kujira')`.

## q

1. Codex raised: `tryLcd()` passes no `signal`/deadline to `queryUrl`, so a hung `rest.cosmos.directory` connection blocks indefinitely rather than falling through to `null`. This predates the PR (all other fallback chains have the same gap), but cosmos.directory is an aggregator proxy that may have different tail-latency behaviour from polkachu. Is there a tracked issue or follow-up PR for adding a per-call timeout to `tryLcd`?

- VA `cosmosTx.ts` `COSMOS_LCD_FALLBACKS['kaiyo-1']` still points to `'https://kujira-api.polkachu.com'` (the same provider as the Kujira primary). This is the same polkachu-as-fallback problem this PR fixes in the SDK, but unaddressed in the VA's separate fallback map. Low urgency since the VA has its own provider-pinning logic (va#1344), but worth tracking.

## risk

low — fallback is only attempted after the polkachu primary already failed; happy path is unchanged. The new endpoint is live (confirmed), returns standard cosmos SDK JSON shape, and serves a genuinely independent infra path. Worst case if `rest.cosmos.directory` is degraded: signing falls through to `sequence:0` exactly as before this PR.

## verdict

APPROVE — the change is correct, distinct-host invariant holds, curl evidence confirms endpoint is live. One should-fix (add Kujira-specific fallback regression test) but not gating.

---
🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` adversarial lane. GLM Z.ai unavailable (exit 1). macOS darwin 25.2.0. curl evidence: `rest.cosmos.directory/kujira` → `kaiyo-1`, `kujira-api.polkachu.com` → 200.

PR #763, opened 2026-06-17
Populate THORChain payload fees

### LINKED ISSUE #415: fix(cosmos): populate THORChainSpecific.fee in buildCosmosPayload for correct fee resolution
## Context

Surfaced by @coderabbitai in #380, unaddressed.

## The problem

`packages/sdk/src/vault/services/cosmos/buildCosmosPayload.ts:69` creates `THORChainSpecific` without setting the `fee` field. The cosmos fee resolver at `packages/core/mpc/keysign/fee/resolvers/cosmos.ts:26` now trusts `value.fee` for `vaultBased` flows — so SDK-built THORChain payloads resolve a default/non-canonical fee instead of the fee encoded in `signData`. The Network Fee row can regress as soon as this ships.

## Fix

Either:
1. Set `fee` in `THORChainSpecificSchema` at buildCosmosPayload.ts:69 using the `fee` from the `buildCosmosPayload` input
2. Or keep the fallback to `mayaGas` in the resolver but also document that THORChain SDK payloads need to supply `fee`

## References

- PR #380 CR thread: https://github.com/vultisig/vultisig-sdk/pull/380#discussion_r3181411199

rcoderdev:
Closes #415

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved THORChain Cosmos payload fee handling so fees are populated from SignAmino input fees when available, with a default of `0` when no fee is provided.
  * Added validation to raise an error if a THORChain fee is supplied without the required fee amount data.

* **Tests**
  * Updated and added unit tests covering THORChain fee population and the new validation behavior.

### DIFF
--- a/.changeset/thorchain-payload-fee.md
+++ b/.changeset/thorchain-payload-fee.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Populate THORChain custom Cosmos payload fees from SignAmino input fees.
--- a/packages/sdk/src/vault/services/cosmos/buildCosmosPayload.ts
+++ b/packages/sdk/src/vault/services/cosmos/buildCosmosPayload.ts
@@ -68,12 +68,17 @@ async function buildCosmosBlockchainSpecific(
   if (chainKind === 'vaultBased') {
     // THORChain or MayaChain
     if (chain === 'THORChain') {
+      const feeAmount = fee?.amount?.[0]?.amount
+      if (fee && feeAmount == null) {
+        throw new Error('THORChain fee.amount[0].amount is required when fee is provided')
+      }
+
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
looks right to me.

- optional `fee` param propagated through the `buildCosmosBlockchainSpecific` call chain
- `feeAmount ?? 0n` default is correct — THOR native txs with no fee info should use `0`
- throw on `fee` present but `amount[]` empty is the right defensive check
- test coverage covers the throw path and the happy path with `2_000_000n`

one tiny q: `buildSignDirectKeysignPayload` passes `undefined` explicitly for `fee` — it could also just drop the arg since `fee` is optional. not blocking, just noting.

PR #764, opened 2026-06-17
Add QBTC claim helper

### LINKED ISSUE #282: [Add] QBTCClaimHelper for Cosmos-level MLDSA signing of claim tx
## Problem

The claim flow in vultisig-windows ([PR #3747](https://github.com/vultisig/vultisig-windows/pull/3747)) needs to sign the Cosmos wrapper (`AuthInfo` + `SignDoc` + `TxRaw`) around `MsgClaimWithProof` with the vault's MLDSA key. The SDK already provides:

- `@vultisig/core-chain/chains/cosmos/qbtc/claim/buildClaimTxBody` — builds the inner `TxBody`
- `@vultisig/core-chain/chains/cosmos/qbtc/claim/broadcastClaimTx` — expects a full `TxRaw` (base64)

But nothing in between. The UI had to hand-roll the `AuthInfo` / `SignDoc` / `TxRaw` protobuf encoding (see [`buildClaimSignDoc.ts`](https://github.com/vultisig/vultisig-windows/blob/feat/qbtc-claim-ui/core/ui/qbtc/claim/utils/buildClaimSignDoc.ts) in the windows repo). This is pure protobuf glue, not UI — it belongs in the SDK so iOS / Android / other clients don't reinvent it when they implement claim.

## Solution

Mirror the existing `QBTCHelper` (send path) with a new `QBTCClaimHelper` that exposes:

```ts
// packages/core/mpc/chains/cosmos/qbtc/QBTCClaimHelper.ts

export const getClaimPreSignedImageHash = ({
  bodyBytes,         // from buildClaimTxBody
  accountNumber,
  mldsaPublicKey,
  sequence,
}): { hash: Uint8Array; authInfoBytes: Uint8Array }

export const getClaimSignedTransaction = ({
  bodyBytes,
  authInfoBytes,
  signature,         // MLDSA signature bytes
}): { txBytesBase64: string; txHash: string }
```

Key differences from the existing `QBTCHelper`:
- Claim tx is **gas-free** — `AuthInfo` omits fee coins (just gas limit)
- Only supports `MsgClaimWithProof` — no MsgSend/Transfer/Vote branching

### Must Do
- [ ] Add `QBTCClaimHelper.ts` with the two functions above
- [ ] Reuse `protoEncoding` utilities from `@vultisig/core-chain/chains/cosmos/qbtc/protoEncoding`
- [ ] Unit tests mirroring the existing `QBTCHelper` tests if any, plus golden-bytes for a known claim payload
- [ ] Export from the package

### Follow-up (separate PR in vultisig-windows)
After this ships, delete [`core/ui/qbtc/claim/utils/buildClaimSignDoc.ts`](https://github.com/vultisig/vultisig-windows/blob/feat/qbtc-claim-ui/core/ui/qbtc/claim/utils/buildClaimSignDoc.ts) and wire [`useQbtcClaimMutation`](https://github.com/vultisig/vultisig-windows/blob/feat/qbtc-claim-ui/core/ui/qbtc/claim/hooks/useQbtcClaimMutation.ts) to the new SDK helpers.

## Reference

- Wire format used: see `buildClaimSignDoc.ts` lines 30-98 in the windows branch — same proto structure as `QBTCHelper` minus fee coins
- Chain ID: `qbtc-testnet` (matches existing helper)
- Pubkey typeURL: `/cosmos.crypto.mldsa.PubKey`

rcoderdev:
Closes #282

## Summary by CodeRabbit

* **New Features**
  * Added QBTC claim transaction helper utilities to enable constructing and signing claim transactions.
* **Tests**
  * Added deterministic test coverage validating the generated claim transaction artifacts and hashes.
* **Release/Packaging**
  * Updated published package exports to make the QBTC claim helpers available to consumers.
  * Included patch releases for core-mpc and sdk.
* **Chores**
  * Adjusted package dependency resolution for `undici` to improve version consistency.


gomesalexandre (commented):
draft, so comment-only.

QBTC claim helper looks clean - 94 lines + 63 lines of tests, changeset included. a few quick things:

q: what's the network this targets? QBTC on Cosmos (hence the `mpc/chains/cosmos/qbtc` path)? the PR description ("Closes #282") doesn't have a summary but the path implies cosmos-based QBTC claiming.

q: `QBTCClaimHelper.ts` imports from `packages/core/mpc/` - do the new exports need to be added to the core-mpc `package.json` exports map, or are they picked up via the existing barrel?

🤖 reviewed via Claude Code sweep

packages/core/mpc/chains/cosmos/qbtc/QBTCClaimHelper.ts
```diff
@@ -0,0 +1,94 @@
+import { Buffer } from 'buffer'
+import { sha256 } from '@noble/hashes/sha256'
+
+import { concatBytes, protoBytes, protoString, protoVarint } from '@vultisig/core-chain/chains/cosmos/protoEncoding'
+
```
johnnyluo: **SignDoc `chainID` is hardcoded to `'qbtc-testnet'`, inconsistent with the rest of the QBTC code — claim signatures may fail verification.**

The SignDoc embeds `protoString(3, chainID)` and the validator rejects any tx whose signed chain_id ≠ the live chain's chain_id. This file uses `'qbtc-testnet'`, while the sibling `QBTCHelper.ts:25` (same network, same package) uses `'qbtc'`, and `proofService.ts` / `computeClaimHashes.test.ts` use `'qbtc-1'`. These three cannot all be correct for one chain.

If the live chain_id is `qbtc-1` (as the proof path implies — the proof's MessageHash binds the same chain_id), every claim built here is signed over the wrong chain_id and the broadcast is rejected with `unauthorized: signature verification failed`, so the claim never lands. The deterministic test locks in the hash for `'qbtc-testnet'`, so it gives false confidence rather than catching the mismatch.

Please confirm the actual node chain_id and source it from one shared constant.

packages/core/mpc/chains/cosmos/qbtc/QBTCClaimHelper.ts
```diff
@@ -0,0 +1,94 @@
...
+  return {
+    txBytesBase64: Buffer.from(txRaw).toString('base64'),
+    txHash: Buffer.from(sha256(txRaw)).toString('hex').toUpperCase(),
+  }
+}
+
+type BuildClaimAuthInfoInput = {
+  mldsaPublicKey: Uint8Array
+  sequence: bigint
+}
```
johnnyluo: **Duplicates SignDoc/AuthInfo/TxRaw assembly already present in `QBTCHelper.ts`.**

`buildClaimSignDoc` is byte-identical to `QBTCHelper.buildSignDocFromComponents`, `buildTxRaw` is byte-identical to `QBTCHelper.buildTxRaw`, and `buildClaimAuthInfo` differs from `buildAuthInfo` only in dropping the fee coin. These now-divergent copies in the same package are exactly how the chainID/fee drift above crept in.

Extracting the shared SignDoc/TxRaw helpers (parameterizing chainID and fee) would prevent further drift.

johnnyluo (commented):
Two findings from review — one correctness concern on the hardcoded chain ID, one cleanup on duplicated serialization. (The unrelated `undici` resolutions churn in `package.json` also looks like incidental `yarn install` output and could be dropped to keep the changeset scoped.)

packages/core/mpc/chains/cosmos/qbtc/QBTCClaimHelper.ts
johnnyluo: qbtc is mainnet now , the chain id is `qbtc`

johnnyluo (approved):
LGTM

PR #767, opened 2026-06-17
feat(keysign): add VULT staking (sVULT) keysign builders

Ehsan-saradar:
## Summary

Adds EVM keysign payload builders for **VULT staking (sVULT)** on Ethereum, plus the minimal signing-input support they need. This is the SDK half of the VULT staking feature in vultisig-windows (vultisig-windows#4141).

### New: `keysign/vultStaking/build.ts`
- `buildVultStakeKeysignPayload` — `sVULT.depositFor(vault, amount)`. When the live VULT→sVULT allowance is insufficient it prepends `VULT.approve(sVULT, amount)` via `erc20ApprovePayload`, so approve + deposit are signed in one ceremony and broadcast back-to-back (nonce N, N+1) — the same shape as a swap-with-approval. The allowance check is in place before `depositFor` executes.
- `buildVultUnstakeKeysignPayload` — `sVULT.requestUnstake(amount)`.
- `buildVultClaimKeysignPayload` — `sVULT.claim(requestId, vault)`.
- `buildVultCancelUnstakeKeysignPayload` — `sVULT.cancelUnstake(requestId)`.

Unstake/claim/cancel act on sVULT directly (no approval) and use the native fee coin + `memo` shape (like Circle withdraw).

### EVM signing-input resolver
A token coin carrying raw `0x` calldata with **`toAmount === '0'`** and no swap payload is now emitted as a generic contract call (calldata → `toAddress`) instead of an ERC-20 transfer. The zero-`toAmount` guard means this can never match a real token transfer (those always carry a non-zero amount); swaps, native sends, and normal ERC-20 transfers are unaffected.

### Exports
Adds the `./keysign/vultStaking/build` subpath export.

## Testing
- `tsc` over the changed files is clean.
- Read paths (`cooldownDuration`, balances, allowance) and the approve→deposit single-ceremony flow were exercised end-to-end from vultisig-windows against the live sVULT contract.

## Summary by CodeRabbit

* **New Features**
  * Added VULT staking (sVULT) EVM keysign payload builders: depositFor, requestUnstake, claim, and cancelUnstake, including automatic allowance handling for depositFor.
  * Exposed a utility to detect “generic contract call” payloads for consistent routing.

* **Bug Fixes**
  * Fixed EVM transaction construction, fee quoting, simulation, and validation so generic contract calls use memo-encoded calldata instead of ERC-20 transfer assumptions.

* **Tests**
  * Added golden-vector and behavior tests covering VULT calldata encodings and allowance scenarios.


gomesalexandre (changes requested):
Deep multi-lane review (Claude correctness + Codex adversarial, 3 rounds) — this is fund-touching keysign (VULT→sVULT on Ethereum mainnet) + a new escape hatch in the shared EVM signing resolver that affects ALL EVM keysigns, so I went hard on it. Build is clean and the existing 1061 tests pass, but none cover the new code. Two blocking issues that would cause real on-chain failures, so REQUEST_CHANGES.

**preferably-blocking (1): fee estimator sends the wrong gas estimate for the depositFor tx**
`packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts` ~L92-103: when `coin.id` is set (the VULT staking case), it builds a synthetic `erc20.transfer(receiver, amount)` to `coin.id` and estimates gas from THAT. But the signed tx is a `contractGeneric` `depositFor(vault, amount)` to the sVULT contract. ERC-20 transfer ≈ 65k; a staking `depositFor` is typically 100-200k. The stored `gasLimit` will be ~65k, so the approve at nonce N succeeds and the `depositFor` at N+1 OOG-reverts — user's VULT ends up approved-to-the-contract-but-not-deposited (stuck until manual revoke/retry). The `isGenericContractCall` routing needs to be mirrored here so gas is estimated against `{ to: toAddress, value: 0, data: memo }`.

**preferably-blocking (2): Blockaid simulation is blind to the generic-contract-call path**
`packages/core/mpc/security/blockaid/tx/simulation/input/resolvers/evm.ts` ~L69-77: same shape — when `coin.id` is set it scans a synthetic `erc20.transfer(receiver, 0)` to the token contract, not the actual `depositFor` to the sVULT contract. Blockaid is scanning the wrong transaction. Mirror the same routing here.

**Root cause (the structural ask):** the `isGenericContractCall` decision lives only in `signingInputs/resolvers/evm/index.ts`; the fee-quote + Blockaid-sim resolvers independently reinvent the routing and weren't updated. Extract a shared `getIsGenericContractCall(payload)` predicate and consume it in all the places that need the routing decision (signing inputs, fee quote, Blockaid sim, + Blockaid validation if applicable).

**should-fix:**
- No tests on a fund-touching mainnet builder. The sibling Noon builder (`chains/evm/noon/noon.test.ts`) sets the bar: golden-vector tests asserting exact calldata hex. Need golden vectors for `encodeDepositFor`/`requestUnstake`/`claim`/`cancelUnstake` + tests for the approve-attached-iff-allowance<amount branch.
- No chain guard on `buildVultStakeKeysignPayload` (build.ts:126) — `underlyingToken.chain` isn't asserted `=== Chain.Ethereum` despite hardcoded ETH logic. Throw if not Ethereum.
- `isGenericContractCall` is broad: fires for any non-native token coin with `toAmount===0` + `0x` memo → arbitrary calldata to an arbitrary `toAddress`. Transfer-theft is blocked (zero amount) but arbitrary-calldata injection isn't scoped. A selector allowlist (`memo.slice(0,10)` ∈ known sVULT selectors), or better a dedicated `contractCallData` protobuf field instead of the memo heuristic, would scope it.
- No `.changeset/*.md` — adds a public `@vultisig/core-mpc` export + changes the shared resolver; release tooling won't version-bump without it.

**q (important):** the method-name strings `depositFor`/`requestUnstake`/`claim`/`cancelUnstake` go through TrustWallet `createWithString(name)` (runtime-string, no schema). Have the resulting 4-byte selectors been verified against the DEPLOYED sVULT contract ABI on Etherscan? The Noon builder uses viem `encodeFunctionData` against a `parseAbi([...])` schema so a typo fails at compile time; here a mismatch is silent until mainnet. An ABI reference comment + (ideally) viem-style encoding would close this.

praise: the approve-then-deposit ceremony itself is correctly structured — allowance checked before conditionally attaching `erc20ApprovePayload`, spender consistently the staking contract across both the check and the approve, nonce N / N+1 sequencing via the existing recursion. That part matches the THORChain-token-input pattern and is right.

I held off taking this over directly because the two blocking fixes touch shared signing infra used by every EVM keysign, and I can't verify the deployed-contract selectors or produce trustworthy golden vectors without the sVULT ABI ground-truth — that needs you. Happy to pair on the `getIsGenericContractCall` extraction if useful. (Also note: branch was 4 commits behind main; I've merged main in for you.)

🤖 reviewed via Claude Code (local ops sweep)

gomesalexandre (approved):
Re-reviewed round 2 (Claude + Codex adversarial) — both fund-safety blockers from my last pass are cleanly resolved. **APPROVE.**

**blocker 1 (fee estimator OOG) ✅** — `getEvmFeeQuote.ts:97` now routes a stake payload through `{to: sVULT, value: 0n, data: depositFor calldata}` via `getIsGenericContractCall`, so gas is estimated against the real contract call, not a ~65k synthetic ERC-20 transfer. No more approve-lands / depositFor-OOG-reverts at nonce N+1. Regular ERC-20 transfers (predicate false + coin.id present) are unchanged.

**blocker 2 (Blockaid scanned wrong tx) ✅** — both `simulation/.../evm.ts:62` and `validation/.../evm.ts:54` got the same `getIsGenericContractCall || !coin.id` fix, so Blockaid now sees the actual `depositFor` calldata.

**root cause fixed as asked** — `getIsGenericContractCall` is extracted into a single predicate consumed by all 4 sites (signing inputs, fee quote, Blockaid sim, Blockaid validation). The predicate is tight (no-swap + non-native coin + `toAmount==='0'` + `0x` memo). The `4eb65224` native-path merge is semantically safe — no coin type matches both branches (native coins are excluded by the `isNativeToken` guard).

**should-fixes from r1 all addressed** — golden-vector tests on all 4 builders (and I re-derived the selectors independently: `depositFor 0x2f4f21e2` / `requestUnstake 0x23095721` / `claim 0xddd5e1b2` / `cancelUnstake 0x2b187b2b`, all match the deployed sVULT `0x11113d7311FB8584a6e82BB126aA11D92e5fB39B`), Ethereum chain guard at `build.ts:116`, changeset present. 1069 tests green, CI + CodeRabbit green.

Two non-blocking follow-ups (don't gate this):
- `should-fix:` no zero-amount guard on `buildVultStakeKeysignPayload` (`build.ts:99`) — `amount=0n` would sign a `depositFor(account, 0)` no-op that wastes gas. `if (amount === 0n) throw ...` closes the foot-gun. No fund-loss (zero moves), worth a tiny follow-up.
- `nit:` `getIsGenericContractCall` matches a bare `memo='0x'` (empty calldata) — builders never produce that, but `memo.length > 2` would tighten it.

Solid work on the predicate extraction + golden-vector approach — exactly the right pattern for fund-touching encoding. Good to merge from my side; the two follow-ups can be a separate issue. 🙏

🤖 reviewed via Claude Code (local ops sweep)

rcoderdev (approved):

gomesalexandre (approved):
Re-reviewed locally - both fund-safety blockers from r1 are resolved (gas estimation now routes depositFor calldata to sVULT contract, Blockaid sees the real tx). getIsGenericContractCall extracted cleanly across all 4 sites. Golden vectors independently verified against deployed sVULT 0x11113d731..., selectors match. 1079 core tests + 1510 sdk tests all green. 🤖 reviewed via Claude Code sweep

PR #773, opened 2026-06-18
fix(zcash): raise memo-send fees to ZIP-317 conventional fee at plan time

Ehsan-saradar:
## Problem

Signing a Zcash transaction with a memo (e.g. bitfrost bridge ZEC→Hyperliquid) fails: the planned fee is exactly one ZIP-317 logical action short of what the network requires, and is rejected at broadcast with *tx unpaid action limit exceeded*. Surfaces in the extension as `Failed to meet the Zcash minimum network fee (ZIP-317): planned 15000 zats, required 20000`.

Tracks vultisig/vultisig-windows#4145.

## Root cause (verified against real WalletCore 4.6.13)

`getUtxoSigningInputs` plans Zcash with `SigningInput.zip_0317 = true`. In that mode WalletCore:
- **flat-sizes an OP_RETURN output as ~34 bytes** regardless of memo length, so it counts one fewer logical action than ZIP-317 charges; and
- **ignores `byteFee` entirely** — `byteFee=1` and `byteFee=100` produce an identical fee.

So memo sends plan one action short (15000 vs required 20000) and there is **no way to raise the fee in `zip_0317` mode**. The extension's post-hoc `byteFee` bump was a no-op for exactly this reason.

| Plan shape | WalletCore zip_0317 | ZIP-317 required |
|---|---|---|
| recipient + change + memo | 15000 | 20000 |
| recipient + memo (sendMax) | 10000 | 15000 |

## Fix

In `getUtxoSigningInputs`, when the `zip_0317` plan underpays the byte-accurate ZIP-317 conventional fee, re-plan with `zip_0317` **off** (where WalletCore honours `byteFee`) and bump `byteFee` — seeded at 1 to discover the planner vsize, then targeting the conventional fee — until the fee clears.

- Plain (no-memo) sends already meet the fee → keep the `zip_0317` plan unchanged.
- Empty / insufficient-funds plans (no selected UTXOs) pass through untouched so `refineKeysignUtxo` keeps owning the sendMax-flip / insufficient-funds flow.
- The corrected fee flows to the approval card automatically (fee display reads the plan).

This fixes both the dApp and mainnet send paths for every SDK consumer. Adds canonical `getZcashOpReturnOutputSize` / `getZcashTransparentOutputSizes` helpers to `zip317.ts`.

## Tests

- `zip317.test.ts`: new helper coverage.
- `utxo.zcashFee.test.ts`: real-WalletCore regression — reproduces the reported 15000→20000 (with change) and 10000→15000 (sendMax) cases, a long-memo case, and the empty-plan pass-through.
- `yarn test:core` (1082 tests), `build:shared`, `typecheck`, `format:check`, `knip`, and the bundled-changeset guard all pass locally.

## Notes

- Changeset included (patch: `@vultisig/core-chain`, `@vultisig/core-mpc`, `@vultisig/sdk`).
- The durable cross-platform fix (iOS/Android, which use WalletCore directly) is a ZIP-317-aware OP_RETURN sizing PR upstream in `trustwallet/wallet-core`; this TS fix covers all SDK consumers without blocking on it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Zcash memo-send transactions being rejected due to insufficient fees by automatically recalculating and adjusting fees to meet ZIP-317 conventional fee requirements.

* **New Features**
  * Enhanced Zcash memo-send fee planning to account for memo-sized OP_RETURN data, improving fee accuracy for short and long memos, including sendMax scenarios.


gomesalexandre (approved):
solid fix, math checks out.

**root cause analysis - correct.** WalletCore `zip_0317` mode: (1) flat-sizes OP_RETURN at ~34 bytes ignoring actual memo length, (2) ignores `byteFee` entirely. So a 40-char memo send plans at 3 output actions * 5000 = 15000 zats while ZIP-317's actual OP_RETURN size of 51 bytes pushes it to 4 output actions * 5000 = 20000 zats. Verified the arithmetic independently.

**fix approach - correct.** Detect underpayment against byte-accurate `getZcashConventionalFee`, then re-plan with `zip_0317=false` where `byteFee` is honoured, bumping byteFee until the fee clears. The `plannerVsize = fee / byteFee` trick to derive target byteFee from the first non-zip plan is smart - avoids blindly incrementing by 1 each time.

**`getZcashOpReturnOutputSize` implementation:**
- push-opcode thresholds correct: direct (<=75), PUSHDATA1 (76-255), PUSHDATA2 (256-65535), PUSHDATA4 (>=65536)
- CompactSize thresholds correct: 1-byte (<253), 3-byte (<=0xffff), 5-byte (<=0xffffffff), 9-byte
- CodeRabbit flagged the PUSHDATA2/CompactSize boundary handling - the PR already includes tests at 249/250 and 255/256 byte boundaries, so that actionable comment is tackled

**test coverage:**
- real-WalletCore regression in `utxo.zcashFee.test.ts`: covers 15000→20000 (with change), 10000→15000 (sendMax), long memo, empty-plan passthrough
- unit coverage of `getZcashOpReturnOutputSize` including the CompactSize/PUSHDATA boundaries the rabbit asked about
- 1510 SDK tests + 1079 core tests all pass

**nitpick:** `maxZcashFeeBumps = 5` is a generous safety net for the bump loop. For real ZEC amounts in practice the byteFee derivation should clear in 1-2 iterations max. Fine as-is.

**fail-open/closed:** throws on failure rather than silently underpaying. correct - better to surface the error than broadcast a transaction the network will reject.

🤖 reviewed via Claude Code sweep

rcoderdev (approved):

PR #775, opened 2026-06-18
Fix SwapKit tracking URL hash parameter

### LINKED ISSUE #771: SwapKit track link doesn't resolve — getSwapTrackingUrl omits the &hash= param (tracker hash field stays empty)
## Problem
The SwapKit explorer "Track" link doesn't resolve — opening it lands on track.swapkit.dev with an **empty Transaction Hash field** (chain is prefilled, hash is not), so the swap can't be tracked. Reproduced on a USDC→ZEC swap (EVM source, chainId=1).

## Cause
`packages/core/mpc/swap/utils/getSwapTrackingUrl.ts:82` builds the URL with only `tx` + `chainId`:

```
https://track.swapkit.dev/?tx=<hash-no-0x>&chainId=1
```

track.swapkit.dev populates its hash field from a separate **`hash`** param (the on-chain hash, 0x-prefixed for EVM) — which we don't send. The `tx` param alone leaves the form blank.

## Working URL (provided)
```
https://track.swapkit.dev/?tx=fc0e43b1…7942e0&chainId=1&hash=0xfc0e43b1…7942e0
```
i.e. `tx` = hash without 0x, plus `hash` = hash **with** 0x.

## Fix
Append the `hash` param using the original (un-stripped) `txHash`:

```ts
return `https://track.swapkit.dev/?tx=${encodeURIComponent(stripHexPrefix(txHash))}&chainId=${chainId}&hash=${encodeURIComponent(txHash)}`
```

For EVM this yields `hash=0x…` (matches the working URL exactly). For non-EVM source chains `txHash` is already bare — worth a quick check that track.swapkit.dev expects the bare form there (the verified example is EVM only).

Surfaces in the Windows app (and shared by iOS/Android via the SDK).

rcoderdev:
Closes #771

## Release Notes

* **Bug Fixes**
  * Updated SwapKit tracker link generation to include the original transaction hash, improving tracker prefill of transaction details.

* **Tests**
  * Adjusted tracking URL expectations to verify the added hash query parameter across supported chains.

* **Chores**
  * Updated dependency resolution overrides to enforce consistent `undici` version mappings.
  * Added Changesets entries for `@vultisig/core-mpc` and `@vultisig/sdk` as patch releases.

### DIFF
--- a/.changeset/swapkit-tracker-hash.md
+++ b/.changeset/swapkit-tracker-hash.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Add the original transaction hash to SwapKit tracking URLs so tracker links prefill correctly.
--- a/packages/core/mpc/swap/utils/getSwapTrackingUrl.test.ts
+++ b/packages/core/mpc/swap/utils/getSwapTrackingUrl.test.ts
@@ -4,6 +4,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { getSwapTrackingUrl } from './getSwapTrackingUrl'
 
 const EVM_HASH = '0xabc123'
+const SOLANA_HASH = '5fP9xSwrQJdWadkC2y5qQbeG29dE1qnv1yNxBmtPb7Xz'
 // Real-shape UTXO hash -- bare 64-char hex, no `0x` prefix.
 // Represents the human-readable txid form (byte-reversed double-SHA256
 // for BTC/BCH/DOGE/LTC/ZEC), which is the format track.swapkit.dev expects.
@@ -27,7 +28,7 @@ describe('getSwapTrackingUrl', () => {
         txHash: EVM_HASH,
         sourceChain: Chain.Ethereum,
       })
-      expect(url).toBe(`https://track.swapkit.dev/?tx=abc123&chainId=1`)
+      expect(url).toBe(`https://track.swapkit.dev/?tx=abc123&chainId=1&hash=${EVM_HASH}`)
     })
 
     it('uses the correct numeric chainId for Polygon', () => {
@@ -36,7 +37,7 @@ describe('getSwapTrackingUrl', () => {
         txHash: EVM_HASH,
         sourceChain: Chain.Polygon,
       })
-      expect(url).toBe(`https://track.swapkit.dev/?tx=abc123&chainId=137`)
+      expect(url).toBe(`https://track.swapkit.dev/?tx=abc123&chainId=137&hash=${EVM_HASH}`)
     })
 
     it('uses slug chainId for Bitcoin AND passes UTXO hash through (no 0x prefix to strip)', () => {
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
she works ser - clean, surgical fix with proper test coverage.

**what it does:** adds `&hash=<original_tx_hash>` to SwapKit tracker URLs alongside the existing `&tx=<stripped_hash>`. The `tx` param keeps its current stripped form (no `0x` for EVM) so the tracker's lookup still works, while `hash` carries the original hash so the form prefills correctly. Two params, two jobs - makes sense.

**correctness check:**

- `URLSearchParams` vs the old `encodeURIComponent(stripHexPrefix(...))` - verified manually: base58 (Solana), lowercase/uppercase hex (EVM, XRP, TON, UTXO), and decimal chainIds (`728126428` for Tron) all pass through URLSearchParams without any unwanted encoding. The `+`-as-space encoding concern doesn't apply to any hash format in scope.
- EVM: `hash=0xabc123` (full original) + `tx=abc123` (0x-stripped for tracker lookup) - correct.
- UTXO/Ripple/TON/Solana: `hash` == `tx` since no stripping happens on these - correct, no regression.
- Solana test previously used `EVM_HASH` (wrong shape for a Solana hash), now uses a real base58-looking constant `SOLANA_HASH`. Good catch - this was a latent test quality issue.
- TON guard (throw on `0x` prefix) - unchanged, still fires before the `searchParams` construct.
- The old `encodeURIComponent` on `stripHexPrefix(txHash)` is replaced by `URLSearchParams` which handles encoding for both `tx` and `hash` - equivalent for all values in scope.
- Changeset present: `@vultisig/core-mpc` + `@vultisig/sdk` patched. Correct packages.

**tests:** 21/21 pass on branch. 1510/1510 full suite green. Typecheck clean.

**package.json undici bump (`^7.19.0` + `^7.24.4` both pinned to `7.28.0`):** security patch for GHSA-vmh5-mc38-953g bundled in the same commit - fine, low noise.

LGTM, merging.

🤖 reviewed via Claude Code (/loop vultisig-sweep)

gomesalexandre (commented):
looks good ser, ship it.

the fix is correct: `tx` keeps the stripped identifier swapkit uses for internal lookup, `hash` carries the full original (with `0x` for EVM, bare for UTXO/Solana/TON/XRP) that the tracker uses to prefill the form. the switch from manual string concat + `encodeURIComponent` to `URLSearchParams` is cleaner and handles encoding uniformly — verified that `0xabc123` and solana base58 chars all pass through without double-encoding.

bonus fix: the old solana test was passing `0xabc123` (`EVM_HASH`) as the txHash for a solana chain — that was a bad fixture. new test uses an actual base58 hash shape. `stripHexPrefix` is a no-op on prefix-free strings so `tx === hash` for solana/utxo/ton/xrp, which is the right behavior.

changeset included for both `@vultisig/core-mpc` and `@vultisig/sdk` as patch. undici lockfile dedup is fine. CI green across all lanes.

🤖 reviewed via Claude Code (/loop vultisig-sweep)

PR #779, opened 2026-06-19
fix(cli): restore client-side tool dispatch via registry instead of dead wire flag

neavra:
## Summary

The CLI gated client-side tool dispatch on a `clientExecuted` flag in the
`tool-input-available` SSE frame. The backend no longer sends that flag — it
deliberately omits any discriminator and expects clients to identify
client-side tools via their own registry. As a result the strict
`clientExecuted === true` check was permanently false, and the CLI's
client-side tools silently stopped dispatching: they degraded to display-only
progress with no local execution and no result shipped back via
`recent_actions`.

This affected four tools: `sign_typed_data`, `vault_coin`, `vault_chain`, and
`address_book`.

## Fix

Dispatch now keys on the CLI's existing client-side-tool registry
(`CLIENT_SIDE_TOOL_DISPATCH`), mirroring how the app identifies client-side
tools via its `toolUIRegistry`:

- `AgentClient` gains a `clientSideToolNames: Set<string>` registry and a
  `setClientSideToolNames()` injector. A `tool-input-available` frame is
  dispatched locally iff its `toolName` is in that set. The dead
  `clientExecuted !== true` wire-flag check is removed.
- `AgentSession` injects the registry from the existing dispatch table
  (`new Set(Object.keys(CLIENT_SIDE_TOOL_DISPATCH))`) — single source of truth.
- The registry is empty by default, so there is no client-side dispatch until
  the session injects it (safe default).

## Receipts

**Before/after** — driving the SSE handler with a representative
current-backend frame (no `clientExecuted` field):

```json
{ "type": "tool-input-available", "toolCallId": "c1", "toolName": "vault_coin",
  "input": { "tokens": [{ "chain": "Base", "ticker": "USDC" }] } }
```

| Scenario | Before (`clientExecuted === true`) | After (registry membership) |
|---|---|---|
| current-backend frame above | not dispatched → tool dead, display-only | `onClientSideToolCall('c1','vault_coin',…)` fired ✓ |
| all 4 registry tools (`sign_typed_data` / `vault_coin` / `vault_chain` / `address_book`) | dead | all dispatch ✓ |
| non-client-side tool (e.g. `polymarket_search`) | not fired | still not fired ✓ (no over-trigger; falls through to server) |
| `tool-input-start` frame | not fired | not fired ✓ (only `-available` dispatches) |
| stray `clientExecuted` flag present | gated the dispatch | ignored — identification is registry-only ✓ |

These are pinned by tests in `clients/cli/src/agent/__tests__/clientSideDispatch.test.ts`,
including a pre-fix repro (no registry ⇒ no dispatch, progress still fires) and
the post-fix dispatch on the same frame.

**Build / typecheck / test (local):**

- `yarn workspace @vultisig/cli typecheck` → green
- `yarn test:cli` → 17 files / 264 tests pass
- `yarn workspace @vultisig/cli build` → green (`Built @vultisig/cli v2.4.0`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed client-side tool dispatch for sign_typed_data, vault_coin, vault_chain, and address_book operations that were unexpectedly falling back to display-only mode. These tools now execute correctly again.

### DIFF
--- a/.changeset/cli-client-side-tool-registry-dispatch.md
+++ b/.changeset/cli-client-side-tool-registry-dispatch.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/cli': patch
+---
+
+Fix CLI client-side tool dispatch. The CLI gated dispatch on a `clientExecuted` wire flag the backend no longer emits, so `sign_typed_data`, `vault_coin`, `vault_chain`, and `address_book` silently stopped dispatching and degraded to display-only progress. Dispatch now keys on the existing client-side-tool registry (mirroring the app's `toolUIRegistry`), restoring those flows.
--- a/clients/cli/src/agent/__tests__/clientSideDispatch.test.ts
+++ b/clients/cli/src/agent/__tests__/clientSideDispatch.test.ts
@@ -255,9 +255,17 @@ describe('SF + CR2 — pendingToolResults catch-block contract', () => {
   })
 })
 
-describe('AgentClient SSE parser — clientExecuted routing', () => {
-  function makeClient(): AgentClient {
+describe('AgentClient SSE parser — registry-based client-side tool routing', () => {
+  // The backend's V1ToolInputAvailable frame carries NO `clientExecuted`
+  // discriminator (the flag was removed). The client must identify client-side
+  // tools via its own registry (mirroring the app's toolUIRegistry), injected
+  // by the session via setClientSideToolNames. These names mirror
+  // CLIENT_SIDE_TOOL_DISPATCH.
+  const CLIENT_SIDE_NAMES = new Set(['sign_typed_data', 'vault_coin', 'vault_chain', 'address_book'])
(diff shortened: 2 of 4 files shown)


gomesalexandre (approved):
does what it says on the box. the `clientExecuted` flag was a dead wire so dispatch on `sign_typed_data`, `vault_coin`, `vault_chain`, `address_book` was silently borked. keying dispatch on the client-side registry (mirroring `toolUIRegistry`) is the right fix and matches how the app side works.

test coverage is thorough: PRE-FIX repro (no registry = no dispatch), POST-FIX (registry = dispatches), all 4 tool names covered, non-registry tools don't over-trigger, stray `clientExecuted` flag correctly ignored. solid diff.

CI green, typecheck clean, fund safety unaffected (no tx signing path touched).

PR #780, opened 2026-06-19
fix(cli): recover assistant answer + tx_ready when the agent stream drops mid-turn

neavra:
## Summary

When the agent SSE stream drops mid-turn, the CLI was throwing out of the stream read loop and losing the turn's authoritative assistant message — along with any `tx_ready` signable card the backend emitted after the drop. The user would see only whatever partial deltas had already streamed, with no answer and no transaction to confirm. The next turn just sent fresh context, so the work was gone.

This is especially painful for `--via-agent` automation, where a transient disconnect on a long-running pipe silently dropped results.

## What changed

The backend keeps processing on a detached context after a client disconnect and persists the assistant message (plus a `data-tx_ready` UI part for any signable card). It exposes a purpose-built recovery endpoint, `GET /messages/since`, with a server-clock anchor (`X-Server-Now`) and an opaque composite cursor. The CLI now uses it:

- **`sendMessageStream` no longer throws on a mid-turn transport drop.** It re-throws only when the signal is aborted (a deliberate Ctrl+C); otherwise it flags the partial result as `disconnected` and captures the `X-Server-Now` anchor.
- **On a disconnect with no recovered message yet, the session polls `/messages/since`** (server-clock anchored, opaque-cursor round-trip, bounded retries — 2s cadence, ~3min ceiling, matching the web app's poller) until the persisted answer lands, then surfaces it through the normal downstream path.
- **A recovered `tx_ready` part is replayed through the same confirm/sign gate as a live card** — same confirmation prompt, same signing path.
- **Pipe mode emits a `reconnecting` event** so agent consumers can distinguish "still working" from "failed".
- **The happy path is unchanged** and never polls `/messages/since`.

### Before / after

| Scenario | Before | After |
|----------|--------|-------|
| SSE drops mid-turn | `reader.read()` throws → turn lost; assistant answer and any tx_ready gone | `disconnected=true`, no throw, `X-Server-Now` captured; session polls `/messages/since`, recovers the answer **and** the `tx_ready` card (routed through the confirm/sign gate) |
| User Ctrl+C (abort) mid-stream | throws (cancelled) | still re-throws — recovery NOT triggered |
| Happy path (clean finish) | answer shown | unchanged; `/messages/since` never polled, `reconnecting` never fired |
| Backend never persists | n/a | bounded: gives up after the poll cap, turn ends cleanly (no hang) |
| Transient poll error | n/a | swallowed; recovers on a later poll |

### Tests

New `sessionRecovery.test.ts` and extended `client.test.ts` drive a `ReadableStream` that errors mid-flight to simulate a dropped socket. An end-to-end test runs the real message loop: turn-1 stream drops → recovery fetches the persisted answer + tx_ready → confirmation → signing → status, then turn-2 closes cleanly.

```
yarn test:cli  →  18 suites / 273 tests pass
CLI build       →  Built @vultisig/cli v2.4.0, clean
typecheck       →  0 errors (CLI workspace, after build:all)
eslint/prettier →  clean on all changed files
```

A `@vultisig/cli` patch changeset is included.

## Release Notes

* **New Features**
  * CLI now recovers from mid-turn SSE disconnects by fetching missed conversation messages and continuing the interaction without losing progress.
  * Added a “reconnecting” event/indicator so the pipe mode and TUI clearly show recovery in progress.
  * Recovered `tx_ready` actions now follow the same confirmation and signing flow as live responses.

* **Tests**
  * Expanded coverage for reconnect/replay timing, partial-stream behavior, and message-loop recovery scenarios.

### DIFF
--- a/.changeset/cli-disconnect-recovery.md
+++ b/.changeset/cli-disconnect-recovery.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/cli': patch
+---
+
+Recover the assistant answer (and any tx_ready signable card) when an agent SSE stream drops mid-turn. The backend keeps processing on a detached context and persists the message, so on a transport disconnect the CLI now polls `/messages/since` (server-clock anchored via `X-Server-Now`, opaque-cursor round-trip, bounded retries) to recover what the dropped stream missed instead of losing the turn. A recovered tx_ready flows through the same confirm/sign gate as a live one. Pipe mode emits a `reconnecting` event so agent consumers can distinguish "still working" from "failed"; a deliberate Ctrl+C abort is unaffected.
--- a/clients/cli/src/agent/__tests__/client.test.ts
+++ b/clients/cli/src/agent/__tests__/client.test.ts
@@ -22,12 +22,46 @@ function makeChunkedStream(chunks: string[]): ReadableStream<Uint8Array> {
   })
 }
 
-function mockFetchSSE(chunks: string[]): typeof fetch {
+function mockFetchSSE(chunks: string[], headers?: Record<string, string>): typeof fetch {
   return vi.fn(
     async () =>
       new Response(makeChunkedStream(chunks), {
         status: 200,
-        headers: { 'Content-Type': 'text/event-stream' },
+        headers: { 'Content-Type': 'text/event-stream', ...headers },
+      })
+  ) as typeof fetch
+}
+
+/**
+ * Stream that yields `chunks` then errors the body mid-flight to simulate a
+ * dropped SSE connection. If `onBeforeError` is provided it runs right before
+ * the body errors (used to flip an AbortController so the read loop sees a
+ * user-cancel rather than a transport drop).
+ */
+function makeDroppingStream(chunks: string[], onBeforeError?: () => void): ReadableStream<Uint8Array> {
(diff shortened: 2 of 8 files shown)


gomesalexandre (approved):
solid recovery fix. the `disconnected` / `finished` / `serverNow` tripartite on `SSEStreamResult` is a clean contract - no ambiguity between clean finish, user abort (ctrl+c re-throws, correct), and transport drop.

fund safety angle - good: stale tx_ready cards are gated behind `replaySignableCards = serverAnchor !== null`. local-clock fallback recovers text only, no sign-gate replay. that's the right call - a clock-skewed anchor reaching back into a prior turn's tx is way worse than losing the text answer. the comment in `recoverDisconnectedTurn` explaining why is appreciated.

few notes:
- `RECOVERY_MAX_POLLS = 90` * `RECOVERY_POLL_INTERVAL_MS = 2000` = 3 min ceiling, matches the comment about the app's 2s poller
- opaque cursor round-trip prevents skip/re-scan across polls, correct
- `recoverySleep` as a separate method so tests can stub it = nice pattern
- test in `sessionRecovery.test.ts` hitting private methods via prototype calls is a bit janky but acceptable for this layer

CI green, typecheck clean.

PR #781, opened 2026-06-19
fix(cli): render balance cards in the terminal instead of raw JSON

neavra:
## Summary

The CLI's agent was never declaring which rich-card surfaces it can render. As a result the backend fell back to its legacy path: instead of emitting a structured `data-balance_summary` card, it instructed the model to **echo the card payload JSON verbatim into the message**. Users running a balance query in the terminal saw a raw `{"surface":"balance_summary",...}` blob instead of a readable summary.

This PR makes the CLI advertise `supported_surfaces: ["balance_summary"]` on every request and renders the resulting card as a terminal table. Because advertising the surface flips the backend to its "card already rendered, reply with one short sentence" contract, advertising **and** rendering are done together — otherwise the balance breakdown would be dropped entirely.

## What changed

- **Advertise the surface** — every agent request now sends `supported_surfaces: ["balance_summary"]`, so the backend emits a structured `data-balance_summary` envelope instead of inlining JSON.
- **Render the card** — `data-balance_summary` is routed through a new callback and rendered as a terminal table grouped by chain (shortened addresses, per-token USD, total, stale cue).
  - TUI: table (flushes any in-flight stream first).
  - Pipe mode: structured `{type:"balance_summary",card}` NDJSON event.
  - `agent ask`: collected on `cards` (JSON mode) / printed as a table (text mode).
- **Defensive fallback** — if a card envelope ever arrives embedded in message content (older backend echoing JSON), a brace-matched scan extracts it, renders the table, and strips the JSON from the displayed text (guarded so a card never renders twice).

Only `balance_summary` is wired. Other model-emitted card contracts (e.g. polymarket / yield) are only injected by the backend when their surface is declared, so the CLI correctly continues to receive prose for those; the surface list and renderers can be extended later.

### Before / after (driving the real `AgentClient.sendMessageStream` against a mock backend that replicates the backend's surface-gating)

- **Before** (no `supported_surfaces`): terminal text = `{"surface":"balance_summary","accounts":[…]}` — raw envelope dumped. `contains raw envelope JSON: true`.
- **After · happy path** (`supported_surfaces:["balance_summary"]`): backend emits `data-balance_summary`; CLI prints a table (chains, tokens, `Total $7,600.00`); message text = `"Your portfolio is worth about $7,600 across 2 chains."`; `contains raw envelope JSON: false`.
- **After · legacy fallback** (backend ignores surfaces, still echoes JSON): fallback detects the envelope, renders the table, strips the JSON → `card rendered as table: true`, no raw JSON.

### Build / typecheck / tests

- `yarn workspace @vultisig/cli build` → `Built @vultisig/cli v2.4.0` ✅
- `yarn workspace @vultisig/cli typecheck` → 26 errors, **all pre-existing** in untouched files (`vault-management.ts`, `core/errors.ts`, `agentErrors.ts`); **zero in any changed file** ✅
- `yarn test:cli` → **278 passed** (new: `agent/__tests__/cards.test.ts` parse/render/fallback + `data-balance_summary` routing case in `client.test.ts`) ✅

A changeset is included (`@vultisig/cli` minor).

## Summary by CodeRabbit

* **New Features**
  * Balance summary cards now render as formatted terminal tables with chain grouping and USD totals
  * Agent advertises supported card surfaces for optimized rendering

* **Bug Fixes**
  * Sanitized balance card strings to prevent ANSI escape sequence injection

### DIFF
--- a/.changeset/cli-supported-surfaces.md
+++ b/.changeset/cli-supported-surfaces.md
@@ -8,4 +8,7 @@ the backend's legacy "echo card_payload JSON verbatim" path. Adds a defensive
 fallback that pretty-renders a balance card envelope if it ever arrives embedded
 in message content (older backend). Cards surface in the TUI (table), pipe mode
 (`balance_summary` NDJSON event), and `agent ask` (`cards` field / rendered
-table).
+table). Card string fields (token symbol, chain, address, amounts) are stripped
+of terminal control bytes at the parse boundary so attacker-influenced on-chain
+metadata or a legacy-echoed payload can't inject ANSI/OSC escape sequences into
+the rendered balances table.
--- a/.changeset/cli-supported-surfaces.md
+++ b/.changeset/cli-supported-surfaces.md
@@ -1,0 +1,11 @@
+---
+'@vultisig/cli': minor
+---
+
+Advertise `supported_surfaces: ["balance_summary"]` to the agent backend and
+render the `data-balance_summary` card as a terminal table instead of triggering
+the backend's legacy "echo card_payload JSON verbatim" path. Adds a defensive
+fallback that pretty-renders a balance card envelope if it ever arrives embedded
+in message content (older backend). Cards surface in the TUI (table), pipe mode
+(`balance_summary` NDJSON event), and `agent ask` (`cards` field / rendered
+table).
--- a/clients/cli/src/agent/__tests__/cards.test.ts
+++ b/clients/cli/src/agent/__tests__/cards.test.ts
@@ -1,0 +1,137 @@
+import { describe, expect, it } from 'vitest'
+
+import {
+  CLI_SUPPORTED_SURFACES,
+  extractBalanceSummaryFromText,
+  parseBalanceSummaryEnvelope,
+  renderBalanceSummaryCard,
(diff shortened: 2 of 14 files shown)


gomesalexandre (approved):
clean fix, good security thinking. the `stripControlChars` at the parse boundary (not the render boundary) is the right defense-in-depth pattern - attacker-influenced on-chain metadata (token symbols, chainId) hits the sanitizer before ever reaching the terminal renderer, regardless of which path fires (typed SSE or legacy echo).

the test for this in `cards.test.ts` is properly paranoid: ESC, BEL, C1 CSI single-byte introducer, CRLF in amountDecimal - all stripped, printable unicode preserved. `renderBalanceSummaryCard` must not throw on the coerced card. checks all the boxes.

`extractBalanceSummaryFromText` backstop at 200k chars to avoid O(n²) matchBrace is sensible. `balanceCardRendered` guard prevents double-render when a transitional backend emits both the typed SSE part AND the echoed blob.

the changeset correctly marks this as `minor` (new `supported_surfaces` field in the request contract, new exports from `cards.ts`).

CI green, 18 new card tests all pass locally, typecheck clean.

PR #783, opened 2026-06-20
Expose SDK Electron main export

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the published `@vultisig/sdk/electron/main` subpath to consistently serve the Electron main CommonJS build with matching TypeScript declarations.

* **Build & Packaging**
  * Added an Electron main-process type bundle to ensure the subpath’s `.d.ts` files are included in the published package.
  * Updated the package export map so the subpath resolves to the Electron main targets.

* **Tests**
  * Extended contract and TypeScript smoke checks to cover `@vultisig/sdk/electron/main`, including runtime export verification for both CJS and ESM import styles.

* **Chores**
  * Added a patch changeset for `@vultisig/sdk` publication.

### DIFF
--- a/.changeset/sdk-electron-main-export.md
+++ b/.changeset/sdk-electron-main-export.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Expose the documented `@vultisig/sdk/electron/main` subpath in the published SDK package.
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #788, opened 2026-06-21
fix(lib-utils): resolve registrable domain via Public Suffix List

Ehsan-saradar:
## Problem

`getUrlBaseDomain` computed the session-keying domain by taking the last two hostname labels:

```ts
new URL(url).hostname.split('.').slice(-2).join('.')
```

"Last two labels" is not the registrable domain. On multi-label public suffixes — `*.vercel.app`, `*.github.io`, `*.pages.dev`, `*.web.app`, `*.co.uk`, … — distinct sites collapse to the shared suffix (`vercel.app`). Anything that keys dApp connection sessions on this value then treats unrelated sibling sites as the same authorized origin: after a user connects to `good-app.vercel.app`, `attacker.vercel.app` resolves to the same key and is treated as already authorized (can read addresses/public keys without a prompt, appears "connected"). Normal domains also share a single session across all subdomains.

## Fix

Use `tldts` `getDomain` with `allowPrivateDomains: true` to compute the registrable (eTLD+1) domain from the Public Suffix List. The `allowPrivateDomains` flag is what makes the PSL's private section (`vercel.app`, `github.io`, …) count as suffixes, so `good-app.vercel.app` and `attacker.vercel.app` resolve to distinct keys. Falls back to the bare hostname for inputs with no registrable domain (`localhost`, IPs).

```ts
export const getUrlBaseDomain = (url: string): string => {
  const { hostname } = new URL(url)
  return getDomain(hostname, { allowPrivateDomains: true }) ?? hostname
}
```

The public signature is unchanged (`(url: string) => string`), so consumers pick up the new behavior transparently.

## Tests

New `baseDomain.test.ts` (vitest, `yarn test:core`): registrable-domain extraction, the `good-app.vercel.app` ≠ `attacker.vercel.app` regression, other multi-label suffixes (`github.io`, `pages.dev`, `web.app`), ccSLD (`co.uk`), and `localhost`/IP fallback. All pass.

## Notes

- Adds `tldts` as a dependency of `@vultisig/lib-utils`; changeset included (patch).
- **Behavior change:** session keys change, so dApps previously connected under a public suffix will need a one-time reconnect.
- Addresses vultisig/vultisig-windows#4167 (the windows extension consumes this via `@vultisig/lib-utils`).

## Bug Fixes
* Improved URL base domain detection to correctly compute registrable domains (eTLD+1) using the Public Suffix List. This prevents incorrect cross-site authorization for hosts under multi-label public suffixes such as `.co.uk`, `vercel.app`, `github.io`, `pages.dev`, and similar.

## Tests
* Added a test suite covering standard hostnames, multi-label public suffixes, ccSLD-style domains, and localhost/IP fallbacks.

## Dependencies
* Added `tldts` as a runtime dependency to support Public Suffix List-based domain parsing.

## Versioning
* Patch version bump for related packages.

### DIFF
--- a/.changeset/psl-aware-base-domain.md
+++ b/.changeset/psl-aware-base-domain.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/lib-utils": patch
+---
+
+Make `getUrlBaseDomain` resolve the registrable (eTLD+1) domain using the Public Suffix List instead of taking the last two hostname labels. Sites under multi-label public suffixes (`*.vercel.app`, `*.github.io`, `*.pages.dev`, `*.web.app`, `*.co.uk`, …) now resolve to distinct domains, so a connection authorized for one site is no longer treated as authorized for an unrelated sibling under the same suffix.
--- a/.changeset/psl-aware-base-domain.md
+++ b/.changeset/psl-aware-base-domain.md
@@ -1,5 +1,6 @@
 ---
 "@vultisig/lib-utils": patch
+"@vultisig/sdk": patch
 ---
 
 Make `getUrlBaseDomain` resolve the registrable (eTLD+1) domain using the Public Suffix List instead of taking the last two hostname labels. Sites under multi-label public suffixes (`*.vercel.app`, `*.github.io`, `*.pages.dev`, `*.web.app`, `*.co.uk`, …) now resolve to distinct domains, so a connection authorized for one site is no longer treated as authorized for an unrelated sibling under the same suffix.
--- a/packages/lib/utils/package.json
+++ b/packages/lib/utils/package.json
@@ -677,6 +677,7 @@
   "dependencies": {
     "buffer": "^6.0.3",
     "long": "^5.3.2",
+    "tldts": "^6.1.0",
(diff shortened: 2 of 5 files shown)


gomesalexandre (approved):
solid security fix. the old `hostname.split('.').slice(-2).join('.')` was genuinely broken for multi-label suffixes (co.uk, vercel.app, github.io etc) - a session keyed under vercel.app would collide across sibling Vercel deployments.

`tldts@^6.1.0` is the right pick (~59M weekly downloads, zero-dep, bundles the PSL, actively maintained). `allowPrivateDomains: true` is correct to treat github.io/vercel.app etc as suffixes. the `?? hostname` fallback handles localhost/IPs cleanly. tests cover the vercel.app regression case with a .not.toBe assertion proving siblings don't collide.

suggestion (non-blocking): downstream sessions keyed by the OLD (broken) base domain won't match post-update - that's the correct security behavior, just worth a one-line heads-up somewhere so it's not mistaken for a regression by app devs.

CI green. ship it. 🤖 reviewed via Claude Code (local /loop sweep)

PR #790, opened 2026-06-21
Align Dogecoin byte fee with app clients

### LINKED ISSUE #786: [chain-quality] SDK Dogecoin byteFee uses stale ÷10 — align to ×0.25 (iOS #4553 / Android #3701 already did)
## Summary
The SDK computes Dogecoin `byteFee` (sat/byte) as **`floor(rawRate / 10)`**, while **iOS and Android both use `rawRate × 0.25`**. Since `fee = byteFee × txSize`, an SDK-initiated Dogecoin send carries **~2.5× lower fee** than the mobile apps under identical network conditions — an underpayment / stuck-tx risk.

This is a **stale regression, not a design disagreement**: iOS and Android deliberately converged on ×0.25, and the SDK's value was left behind in a sync-and-copy.

## Triage — which is correct, and why
`×0.25` is the canonical value. Git history is explicit:
- **iOS `0995ffa60` — `fix(doge): apply fee tier and Android-aligned base instead of flat /10 divisor (#4553)`** — iOS **removed `÷10`** and moved to the Android-aligned `×0.25`.
- **Android `07b753111` — `fix: reduce excessive DOGE fee estimation … (#3701)`** — Android intentionally landed on `×0.25`.
- **SDK `676e0abd` — `sync-and-copy changes`** — the SDK kept `÷10`. Its comment claims it "matches the iOS impl's /10 workaround", but **iOS abandoned `÷10` in #4553** — so the comment is now false and the value is stale.

## Evidence (verified against source)
- **SDK (this repo, the bug):** `packages/core/chain/chains/utxo/fee/byteFee.ts:23` → `return base / 10n` → `floor(raw / 10)`
- iOS (correct): `VultisigApp/VultisigApp/Core/Services/BlockChainService.swift:30-33` → `dogeBaseFeeMultiplier = 0.25`
- Android (correct): `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/utxo/UtxoFeeService.kt:30-31` → `gas.multiply(5).divide(20)` → `floor(raw / 4)`

Live Blockchair `suggested_transaction_fee_per_byte_sat` for Dogecoin ≈ **500000**:

| Platform | formula | byteFee (sat/byte) |
|---|---|---|
| **SDK (bug)** | floor(500000 / 10) | **50000** |
| iOS | 500000 × 0.25 | 125000 |
| Android | floor(500000 / 4) | 125000 |

## Recommended fix
In `packages/core/chain/chains/utxo/fee/byteFee.ts`, change the Dogecoin branch from `base / 10n` to the ×0.25 form Android uses (`base / 4n`), and update the now-false comment:

```ts
if (chain === UtxoChain.Dogecoin) {
  // Blockchair reports an inflated DOGE rate; ×0.25 lands on a sane next-block rate.
  // Matches iOS dogeBaseFeeMultiplier (#4553) and Android UtxoFeeService (#3701).
  return base / 4n
}
```
(Also applies to the standalone `estimateUtxoFee` in `packages/sdk/src/chains/utxo/rpc.ts`, which has the same `÷10`.) Add a cross-platform parity test so this can't drift again — cf. #715 (Cardano byteFee parity).

## How found
Per-chain quality probe comparing byteFee derivation across SDK/iOS/Android against source, then git-archaeology to identify the canonical value (`Giga/shared/research/loops-24-7/probe/fee/live-bytefee.mjs`, `fee/TAXONOMY.md`). iOS and Android verified already-correct — no issue needed there.

rcoderdev:
Closes #786

## Summary by CodeRabbit

* **Bug Fixes**
  * Dogecoin transaction fee estimation corrected to align consistently across app clients through an updated baseline fee calculation method.
  * Both core and SDK packages now implement the standardized fee multiplier for accurate transaction estimates.

* **Tests**
  * Added test coverage for Dogecoin fee estimation to verify correct baseline calculation behavior.

### DIFF
--- a/.changeset/doge-bytefee-parity.md
+++ b/.changeset/doge-bytefee-parity.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Align Dogecoin Blockchair fee estimation with the app clients by using 25% of the reported baseline.
--- a/packages/core/chain/chains/utxo/fee/byteFee.test.ts
+++ b/packages/core/chain/chains/utxo/fee/byteFee.test.ts
@@ -1,0 +1,21 @@
+import { UtxoChain } from '@vultisig/core-chain/Chain'
+import { getUtxoStats } from '@vultisig/core-chain/chains/utxo/client/getUtxoStats'
+import { describe, expect, it, vi } from 'vitest'
+
+import { getUtxoByteFee } from './byteFee'
+
+vi.mock('@vultisig/core-chain/chains/utxo/client/getUtxoStats', () => ({
+  getUtxoStats: vi.fn(),
+}))
+
+const mockGetUtxoStats = vi.mocked(getUtxoStats)
+
+describe('getUtxoByteFee', () => {
+  it('uses the app-aligned 25% Dogecoin multiplier', async () => {
+    mockGetUtxoStats.mockResolvedValueOnce({
+      data: { suggested_transaction_fee_per_byte_sat: 500_000 },
+    })
+
+    await expect(getUtxoByteFee(UtxoChain.Dogecoin)).resolves.toBe(125_000n)
+  })
+})
--- a/packages/core/chain/chains/utxo/fee/byteFee.ts
+++ b/packages/core/chain/chains/utxo/fee/byteFee.ts
@@ -17,9 +17,9 @@ export const getUtxoByteFee = async (chain: UtxoChain) => {
   const base = BigInt(suggested_transaction_fee_per_byte_sat)
 
   if (chain === UtxoChain.Dogecoin) {
-    // According to iOS implementation: For Dogecoin, the API responds with 500,000 sats/byte, which exceeds what WalletCore expects.
-    // To ensure compatibility with WalletCore, we divide the API value by 10 to bring it into an acceptable range.
-    return base / 10n
+    // Blockchair reports an inflated DOGE rate; 25% matches the iOS and Android
+    // Dogecoin fee multiplier while keeping the value in WalletCore's range.
+    return base / 4n
   }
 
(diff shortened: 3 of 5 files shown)


gomesalexandre (approved):
Solid fix. Verified locally - build green, all 1510 unit tests + 1111 core tests pass.

**adversarial passes:**

**1. Is the math correct?** Yes. Issue #786 nails the git archaeology: iOS removed /10 in commit `0995ffa60` (#4553) and Android has been on /4 since `07b753111` (#3701). SDK's old comment literally said "matches the iOS impl's /10 workaround" but iOS abandoned that in #4553 - stale + wrong. The PR brings both code paths (core `byteFee.ts` and SDK `rpc.ts`) to `base / 4n` / `Math.floor(base / 4)` which is exactly `×0.25` and matches what Android does (`multiply(5).divide(20)`). The changeset covers both published packages correctly.

**2. Edge cases:** `rpc.ts` already has `Math.max(1, ...)` so base=0 or base=1 can't produce 0 fee on that path. `byteFee.ts` (core path) has no such guard - `1n / 4n = 0n` could theoretically slip through if Blockchair reports <=3 sats/byte for DOGE. But this was also true before (1n/10n = 0n), so not a regression introduced here. Pre-existing floor-guard gap, out of scope.

**3. Fund safety / overpay?** Normal priority: 125,000 sat/byte × ~225 bytes ≈ 0.28 DOGE (~$0.07 at current price) - totally normal. Fast priority: ×2.5 multiplier gets you ~0.7 DOGE/tx. Not an overpay concern - this is what iOS/Android users already experience. Old SDK value was 2.5× underpaying vs app clients (50k vs 125k), creating stuck-tx risk. This PR fixes that.

**changeset:** both `@vultisig/core-chain` and `@vultisig/sdk` bumped as patch - correct.

**test coverage:** new `byteFee.test.ts` in core covers the exact 500k-raw → 125k-result path. SDK regression test updated. Both pass locally.

ship it ser

PR #791, opened 2026-06-22
fix(cli): don't drop narration / balance card around balance summaries

gomesalexandre:
## what

Two follow-ups to #781 (both were open CodeRabbit `should-fix` threads that merged unaddressed; surfaced in a post-merge deep review). Neither is fund-loss — both are silent display drops in the CLI.

1. **`tui.ts` `onBalanceSummary` doesn't clear `currentStreamText`.** A `data-balance_summary` part can arrive after some text deltas; the handler flips `isStreaming=false` but leaves the buffered partial in `currentStreamText`. A following `onAssistantMessage` whose content equals that partial then hits the `content !== this.currentStreamText` guard and is **silently dropped** — the assistant narration vanishes. Fix: clear `currentStreamText` after rendering the card (mirrors `onDone`).

2. **`session.ts` disconnect-recovery doesn't replay balance cards.** `recoverDisconnectedTurn`/`applyRecoveredMessage` only replayed `data-tx_ready`. A balance query whose SSE stream drops mid-turn recovers the text answer but **silently loses the card**. Fix: thread `onBalanceSummary` through and replay persisted `data-balance_summary` parts. Replayed **UNCONDITIONALLY** (not gated by `replaySignableCards`): a balance card is read-only display, so a stale one is at worst cosmetic — only a stale `tx_ready` is a fund-safety concern, so that stays gated by the server-anchor.

## receipts

Worktree-isolated work (another agent on the shared sdk clone). `clients/cli` `cards.test.ts` (the extractor/backstop area for fix 1) green locally:
```
src/agent/__tests__/cards.test.ts  18 passed
```
The session-recovery suite needs the full workspace build to resolve `@vultisig/sdk`, so I'm leaning on CI for the full typecheck + session/tui suite (will merge only on green). Changes are type-safe by inspection: the new `onBalanceSummary: (raw: unknown) => void` param matches the existing live callback at session.ts:366; `part.data` is `unknown`.

## risk

Low — CLI display only, no fund path. The tx_ready replay gate is untouched.


gomesalexandre (commented):
self-review pass — balance-summary unconditional replay vs tx_ready gated is the right distinction, comment is clear. `currentStreamText = ''` in tui.ts correctly unblocks the onAssistantMessage guard. CI running, waiting for a stamp.

NeOMakinG (approved):
APPROVE. Both display drops are real and the fixes are correct. Traced both paths at source; CLI typecheck + recovery/cards/confirm-gate suites green locally; full CI green on HEAD.

## testing approach
3-lane fan-out: Claude load-bearing (Codex 0-byte stall — known session-serialization; GLM ECONNRESET — both best-effort, dead this round). Built `shared`+`sdk:bundle`+`client-shared`+`rujira` in a worktree off `72b02b2`, ran the cli recovery/cards/confirm-gate suites (38 passed), and `yarn workspace @vultisig/cli typecheck` (clean, 0 errors). Mutation-probed both fixes.

Verified:
- **Fix 1 (tui.ts)** — clearing `currentStreamText=''` after `renderBalanceSummaryCard` mirrors `onDone`'s flush-and-clear. A `data-balance_summary` part arriving mid-stream no longer leaves the buffered partial behind, so a following `onAssistantMessage` equal to it isn't eaten by the `content !== this.currentStreamText` guard. Correct.
- **Fix 2 (session.ts)** — `onBalanceSummary` is threaded through `recoverDisconnectedTurn` → `applyRecoveredMessage`; `data-balance_summary` parts replay UNCONDITIONALLY while `data-tx_ready` stays gated by `replaySignableCards` (server-anchor). Contract-shape correct; the new `(raw: unknown) => void` matches the live callback at session.ts:366. CLI typecheck clean.
- **No double-render** — the recovery replay routes through `callbacks.onBalanceSummary` (session.ts:366) which sets `balanceCardRendered = true` (369). `renderEchoedBalanceCard` (441) is gated by that flag, so the echoed-card fallback won't re-render. Confirmed.
- **No stale-balance mismatch** — card and text answer both derive from the same recovered `msg`, so a stale recovered row yields a stale-but-internally-consistent card+narration. The only fund-adjacent card (`tx_ready`) remains server-anchor-gated. The "at worst cosmetic" rationale holds.

## should-fix
- **No regression test for either new path.** Mutation check: deleting the entire `data-balance_summary` replay block in `applyRecoveredMessage` fails ZERO tests — `sessionRecovery.test.ts` only asserts `tx_ready` replay + the server-anchor gate. And `cards.test.ts` (cited in the PR as the fix-1 backstop) only exercises the pure `parseBalanceSummaryEnvelope`/`renderBalanceSummaryCard`/`extractBalanceSummaryFromText` functions — it never instantiates `ChatTUI` or touches `currentStreamText`, so the tui clear is also uncovered. Both fixes are correct by inspection, but this is the PR #228 shape (clean pure-function coverage, unguarded integration path). Non-blocking given LOW display-only risk + green broad suite, but a follow-up adding (a) a recovery test with a persisted `data-balance_summary` part asserting `onBalanceSummary` fires, and (b) a tui test asserting a post-card `onAssistantMessage` equal to the buffered partial still renders, would lock both behaviors.

## risk
Low — CLI terminal display only, no fund path. `tx_ready` replay gate untouched. Changeset N/A (`@vultisig/cli` client, not the published `@vultisig/sdk` surface).

## verdict
APPROVE.

---
🤖 reviewed by vultisig-ops (Claude load-bearing; Codex + GLM lanes dead this round)

PR #806, opened 2026-06-23
Fix swap quote net output ranking

### LINKED ISSUE #804: Swap routing: rank by net user output (fees + gas), not gross; tighten preference band 100→50bps
## Summary

The swap quote ranker selects the **best gross destination amount**, not the **best net output the user actually receives**. We want best *net* user output (after provider/integrator fees and gas) to be the selection metric on every platform. This is the SDK/Windows/extension/agent surface — currently the worst of the three at this.

## Current behavior (gross ranking)

`packages/core/chain/swap/quote/findSwapQuote.ts`

- `getComparableOutputAmount()` returns the decimal-aligned **gross** destination amount only. The function carries a standing `TODO(#353 follow-up)`: *"subtract route-specific gas / outbound fees for true net output; today this ranks gross destination amount after decimal alignment only."*
- `#353` was closed for the decimal-alignment part, but the net/gas part was never implemented — so the SDK can select a provider that quotes a higher gross amount while delivering **less** to the user after fees and gas.
- iOS already ranks net of destination-side fees (`expectedNetToAmount`); the SDK does not. Same swap → potentially different (worse) provider on Windows/extension than on iOS.

## Desired behavior

Rank by a single net-output metric, identical across platforms:

```
comparableOutput = destinationAmountOut
                 − destinationSideFees            (affiliate / integrator / outbound, in dest token)
                 − sourceGasInDestinationTerms     (EVM same-chain routes)
```

- THORChain/Maya `expected_amount_out` is already net of affiliate+outbound — keep as-is (do **not** double-subtract).
- For 1inch/Kyber/LiFi/SwapKit, subtract the integrator/affiliate fee where the quoted `dstAmount` is gross of it.
- Include source gas for same-chain EVM routes so a gas-heavy route can't win on gross while netting the user less. (If a full gas-normalized price path is too heavy for the hot path, apply gas as a **tie-break within the preference band** rather than in the headline metric — acceptable middle ground, call it out in review.)

## Also: tighten the preference band 100 → 50 bps

`SWAP_QUOTE_PREFERENCE_BAND_BPS = 100n` lets a near-tie hand a "preferred" provider up to **1%** of user value — 2× the 50 bps affiliate fee. Tighten to **50 bps** so the band can never cost more than the fee we charge.

## Acceptance
- [ ] Ranking metric is net of destination-side fees (closes the real #353 gap)
- [ ] Source gas included for same-chain EVM (or applied as in-band tie-break, documented)
- [ ] `SWAP_QUOTE_PREFERENCE_BAND_BPS` = 50
- [ ] Shared cross-platform test vector: fixture quotes → expected selected provider, matching iOS/Android
- [ ] No double-counting for THOR/Maya (regression test)

## Context
Follow-up to the closed convergence issues #605 / #353. Companion issues on iOS (`vultisig-ios`) and Android (`vultisig-android`). Goal: same swap selects the same best-net provider on all three platforms.

rcoderdev:
Closes #804

## Summary
- rank swap quotes by comparable net destination output with the 50 bps preference band
- avoid double-subtracting provider outputs that are already net, including 1inch/Kyber/native/CowSwap
- keep the same-chain EVM gas-unit tie-break and regression coverage

## QA
- yarn vitest run packages/core/chain/swap/quote/findSwapQuote.selection.test.ts
- node scripts/run-eslint.mjs --config .config/eslint.config.mjs packages/core/chain/swap/quote/findSwapQuote.ts packages/core/chain/swap/quote/findSwapQuote.selection.test.ts
- yarn tsc --project .config/tsconfig.json --noEmit --pretty false
- git diff --check
- yarn build:all
- yarn lint && yarn test
- yarn cli:build && yarn changeset:check-sdk-bundle
- yarn check:ci
- live CLI quote/dry-run ETH -> USDC selected Kyber with destination fee evidence
- live CLI ETH -> BTC native quote returned THORChain net-output fee fields
- live 1inch probe confirmed affiliateBps=50 already lowers dstAmount versus affiliateBps=0

Note: no mainnet broadcast was performed; QA used live quotes and dry-run because the changed surface is quote ranking.

## Release Notes

* **Improvements**
  * Refined swap quote ranking to prioritize net user output consistently across providers.
  * Tightened the “near-best” provider preference band from 1% to 0.5% (50 bps).
  * Added gas-aware tie-breaking so lower gas can win only when quotes are otherwise comparable.
  * Improved destination-side fee handling to prevent double-counting when providers already return net amounts.
* **Tests**
  * Expanded selection scenarios for the new 50 bps band behavior, affiliate-fee edge cases, and same-chain EVM gas ties.


gomesalexandre (commented):
comment-only (draft) - read the diff carefully ser, some qs:

**q: SwapKit in `providerAlreadyNet`?**

`providerAlreadyNet` is currently `provider === 'kyber' || provider === '1inch'`. Both are excluded because their `dstAmount` already has the affiliate deducted at quote time.

SwapKit wraps native THORChain/MayaChain. The question is whether SwapKit's `general.dstAmount` also arrives already-net of affiliate. If yes and SwapKit happens to return an EVM tx with an `affiliateFee` field populated, `getGeneralDestinationSideFeeAmount` would subtract it again - ranking SwapKit artificially lower than it deserves and systematically favouring other providers.

The test covers the "outside 50 bps band: LiFi beats SwapKit" case (SwapKit = 1000000, LiFi = 1006000) but the minimal fixture doesn't have an `affiliateFee` on the SwapKit quote, so double-subtraction isn't exercised. Worth confirming whether SwapKit ever populates `affiliateFee` in the EVM tx and if so, whether it needs to join `providerAlreadyNet`.

**q: gasLimit tie-break**

`getSameChainEvmSourceGasUnits` returns `tx.evm.gasLimit`. gasLimit is the worst-case cap, not actual gas. A provider quoting a tighter gasLimit isn't necessarily cheaper - it could be more OOG-prone. Using `gasLimit` as a selection signal could systematically prefer routes that estimate low. Was the intent to use estimated gas cost (gasLimit × gasPrice) instead, or is gasLimit a reasonable proxy here? (PR description says 'same-chain EVM gas-unit tie-break' which implies intentional, just want to confirm the reasoning.)

Rest of the changes look solid - the decimal rebase before subtraction, `subtractClamped` guard, band narrowing 100→50 bps, CowSwap 0n path. Live CLI evidence appreciated.

gomesalexandre (commented):
draft, so comment-only ser.

solid pr - the 3-part fix (net-output deduction, band tightening, gas tie-break) is well-structured and the test coverage is thorough. a few things worth thinking through before undraft:

**q: LiFi net-output guarantee?** `providerAlreadyNet` excludes 1inch and Kyber, but not LiFi. If LiFi's `dstAmount` is already post-affiliate-fee (like 1inch), the fee subtraction via `getGeneralDestinationSideFeeAmount` would apply a double deduction. Worth a comment in the code (or confirming LiFi's `dstAmount` is gross before fees).

**band tightening 1%→0.5%:** narrowing the native preference window means THORChain/MayaChain lose more often to general aggregators when the spread is 0.6-0.9%. That's the intended behavior per issue #804, just flagging for visibility since it's a routing policy change.

**gas tie-break fallthrough when one candidate has gas and the other doesn't:** `hasComparableGas` requires both `selectedGasUnits !== undefined` and `candidate.sourceGasUnits !== undefined`. If only one candidate has gas info (e.g. Kyber EVM vs SwapKit non-EVM tx on same chain), the tie-break skips entirely and falls through to preference-rank comparison. That seems correct but would be worth a comment.

**`isSameCoinKey(fee, to)` with both ids undefined:** native coins could have `id === undefined`. Since native quotes take the `'native' in q.quote` branch early in `getComparableOutputAmount`, this code path doesn't run for them. Maybe add a guard anyway for safety.

tests look clean - especially the 1inch/Kyber double-subtraction regression tests. 🤖 reviewed via Claude Code sweep

gomesalexandre (commented):
## Review

Checked out branch, ran `yarn vitest run packages/core/chain/swap/quote/findSwapQuote.selection.test.ts` - 38/38 green. Traced the provider quote builders to verify the "already net" claims.

### preferably-blocking

**1. LiFi affiliate-fee subtraction is very likely a double-subtraction (fund-safety / ranking bias)**

`getComparableOutputAmount` does `subtractClamped(dstAmount, affiliateFee)` for any general quote that isn't kyber/1inch and carries a destination-token `affiliateFee`. In practice the only provider that ever populates a destination-side `affiliateFee` is LiFi (`getLifiSwapQuote.ts:325`). But LiFi's `dstAmount` is `estimate.toAmount`, which per LI.FI's API contract is the *net amount the user receives* (integrator fee already baked in). The `feeCosts`/`affiliateFee` is an informational breakdown, not something to subtract again.

Proof it's display-only everywhere else: `SwapService.ts:365` uses `general.dstAmount` directly as the receive amount, and the swap-fee row is rendered separately from `affiliateFee`/`swapFee`. So this PR makes ranking interpret `dstAmount` differently from the entire rest of the codebase - under-counting LiFi's true output and biasing the ranker *against* LiFi. The exact double-subtraction this PR set out to kill, just relocated from 1inch/Kyber onto LiFi.

q: did you confirm `estimate.toAmount` is gross vs net against an actual LiFi quote payload? LI.FI SDK types carry no doc strings. A captured real response in the PR description would settle this. If `toAmount` is post-fee (I'm fairly sure it is), drop the LiFi subtraction entirely - every provider is already net, which collapses most of this diff.

`findSwapQuote.ts:203-218,236-237`.

**2. Gas tie-break can hijack the preference order (incl. CowSwap-first)**

The pairwise gas tie-break at `findSwapQuote.ts:289-309` `continue`s after a gas comparison, making gas able to override preference rank in an order-dependent way. Example with 3 in-band equal-output candidates:
- A=CowSwap (rank 0, gas 200k) - selected initially
- B=1inch (rank 5, gas 100k) - `hasComparableGas` fires, 100k < 200k - B selected
- C=SwapKit (rank 3, gas 200k) - 200k < 100k false - B kept

Result: 1inch wins over CowSwap despite CowSwap being deliberately slotted first (MEV-protected, gasless, `providerPreferenceOrder` comment lines 120-125). Gas silently outranks the whole preference list. This only bites on genuine in-band same-chain EVM ties - exactly the scenario the band exists for.

Action: make gas a *final* tie-break applied only among candidates of equal preference rank, not a pairwise short-circuit. File: `findSwapQuote.ts:289-318`.

### should-fix

**3. Test coverage doesn't actually exercise the LiFi subtraction path or guard the CowSwap/gas hijack**

The two "does not double-subtract" tests cover 1inch and Kyber - the providers where subtraction is *skipped*. There's zero test where subtraction actually fires (a LiFi/SwapKit quote with a destination `affiliateFee` that changes ranking). The one code path doing real work is untested. Also:
- No test that source-token `affiliateFee` is NOT subtracted (the `isSameCoinKey(fee, to)` guard is load-bearing)
- No test for the CowSwap-vs-gas interaction from finding #2

**4. The PR narrative about SwapKit is misleading - SwapKit never carries a dest affiliateFee**

`buildSwapKitTx` -> `buildEvmTx` never sets `affiliateFee`. So SwapKit, despite being framed as "not already net" and subject to subtraction, gets `0n` subtracted in practice - identical to 1inch/Kyber. The doc comment at `findSwapQuote.ts:225-228` implies SwapKit/LiFi get adjusted and others don't, but real behavior is: only LiFi ever gets adjusted. Tighten the comment to match reality.

### suggestion

**5.** `subtractClamped` floors to `0n` when `fee >= amount`. If that branch fires it means the fee model is broken (fee shouldn't exceed output), and silently flooring would make a broken quote look like the worst-but-valid candidate rather than surfacing the data error. A dev-time log would help.

**6.** `getSameChainEvmSourceGasUnits` returns `tx.evm.gasLimit` which is `bigint | undefined`. When undefined the candidate just doesn't participate in the gas tie-break - that's the right graceful degradation, but worth a one-line comment so nobody "fixes" it to `0n` later (which would wrongly make gasless-looking candidates dominate).

### q:

**7.** Why does the gas tie-break test pass `vultDiscountTier: 'ultimate'` (`findSwapQuote.selection.test.ts:967`)? It's the only test that sets it, no comment, assertion is purely about provider selection on gas. Cargo-culted, or is there a real dependency?

---

tl;dr the 50 bps band rename + gas-unit tie-break scaffolding is good direction, but the net-output subtraction is probably double-counting LiFi's fee (the only provider it actually touches), and the gas tie-break can quietly override CowSwap-first preference. Both want resolving before this leaves draft.

gomesalexandre (approved):
lgtm 🫡 net-output ranking is well-designed and conservative - reviewed the economics closely, no fund-quality regression.

Adversarial read (packages/core/chain/swap/quote/findSwapQuote.ts):
- getComparableOutputAmount: getGeneralDestinationSideFeeAmount returns 0n for EVERY current general provider (1inch/li.fi/kyber/swapkit/jupiter/cowswap are all in the providerAlreadyNet list), so their dstAmount - which is already user-receive/net - is used unchanged. No output regression for any live provider; the fee-subtract is a correct future-guard for a provider that ever reports GROSS, rebased to dest decimals + subtractClamped (no negative). Native THOR/Maya path untouched.
- same-chain-EVM gas tie-break (getSameChainEvmSourceGasUnits): a SEPARATE dimension applied only after output-band + provider-rank, and only when both quotes have comparable gas. Gated to from.chain===to.chain && both EVM, where gas PRICE is identical, so comparing gas UNITS is a valid cost proxy (not apples-to-oranges vs output). Missing gas stays undefined (not 0), so an incomplete quote can't dominate the tie-break. cowswap -> 0n (gasless) is right.
- band tightened 100->50 bps: users switch to a better provider at a smaller threshold. CowSwap gasless preference still leads via provider-rank.

Receipts: CI green on the full check:ci gate (build:all + the 45-case findSwapQuote.selection.test.ts covering net-output selection, decimal-normalized ranking, tie-breaks, band, cowswap, gas, native-halt, below-min). Local isolated re-run needs build:all (the whole workspace), which CI already ran - not duplicating it here; verified the ranking logic by adversarial read instead.

PR #807, opened 2026-06-23
Add Tron token symbol integrity test

### LINKED ISSUE #553: [test] add Tron token registry symbol() integrity test
R1 review on #552 found 2 wrong contract addresses (stUSDT pointing at a Sunswap LP token, USDD pointing at the deprecated pre-migration contract). A registry-level test that calls TronGrid `symbol()` on each registered Tron token and asserts it matches the ticker would have caught both.

Findings from #552:
- stUSDT was `TQn9Y2khEsLJW1ChVWFMSMeRDow5KcbLSE` (JUSTSWAP-V1 LP token) - symbol() would've returned `JUSTSWAP-V1`, not `stUSDT`
- USDD was `TPYmHEhy5n8TCEfYGqW2rPxsghSfzghPDn` (TronScan-tagged USDDOLD) - the deprecated pre-migration contract

Open question: how to do this without flaky network calls in CI - maybe a separate `yarn test:integration` lane that runs against mainnet TronGrid, off by default. Similar pattern to the existing `test:e2e` suite.

Note for agents: This is a follow-up from sdk#552. The task is to add a test (likely in `packages/core` or `packages/sdk/tests/integration`) that iterates over the Tron entries in `knownTokens`, calls TronGrid `triggerconstantcontract` with `symbol()` for each, and asserts the returned symbol matches the `ticker` field. Should be gated behind a flag or in a separate test lane so it doesn't run in CI by default.

rcoderdev:
Closes #553

## Summary by CodeRabbit

* **Tests**
  * Added a live integration test that validates on-chain TRC-20 `symbol()` values for known Tron tokens against their locally configured tickers.
  * The test runs only when the live environment flag is enabled, includes decoding of on-chain responses, retries with backoff for rate-limit errors, and throttles requests to keep execution within a bounded runtime.


gomesalexandre (approved):
does what it says on the box

live TRC-20 symbol integrity check against TronGrid - good catch, this kind of drift (ticker in SDK registry vs on-chain `symbol()`) is hard to spot without an automated live check.

**ABI decode check:** `decodeTrc20String` - skips the first 32 bytes (offset word), reads length from bytes 32-64, then data from 64 onward. standard ABI encoding for `string` return. the `< 128` fallback handles non-ABI-compliant contracts that return packed bytes directly. ✓

**retry logic:** 3 attempts, backs off 6s on rate-limit only (`request rate exceeded`), throws otherwise. sensible.

**timeout fix:** `Math.max(60_000, tronTokens.length * 13_500)` scales with registry size. per-token budget of 13.5s covers 2 retries (6s each) + 400ms sleep + network round trip. tight but reasonable for a live test.

**gating:** `skipIf(!runLiveTronGridRegistryCheck)` means this doesn't run in CI, so it won't flake. the env var guard is the right pattern for external-api tests.

PR #808, opened 2026-06-23
Protect native swaps with minimum output limits

### LINKED ISSUE #688: THORChain/Maya native swaps signed with no min-output limit (toAmountLimit hardcoded '0', no tolerance_bps)
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` swap path).

## Observed
THORChain / Maya native swaps are signed with **no minimum-output protection** (`LIM=0` in the memo). The price can move arbitrarily between quote and execution and the swap will still fill, sending the user as little as the pool decides — there is no on-chain slippage floor. (General aggregator providers DO carry slippage; only the THORChain/Maya native path is exposed.)

## Cause
The quote URL is built without a `tolerance_bps` parameter:

`packages/core/chain/swap/native/api/getNativeSwapQuote.ts` (`requestNativeSwapQuote`, the `URLSearchParams` block):
```ts
const params = new URLSearchParams({
  from_asset, to_asset, amount: chainAmount.toString(), destination,
  streaming_interval: String(streamingInterval),
  // no tolerance_bps / min-output param
})
```

Then the payload hardcodes the limit to zero, and the memo (carrying `LIM=0`) is signed verbatim:

`packages/core/mpc/swap/native/utils/nativeSwapQuoteToSwapPayload.ts:49`
```ts
toAmountLimit: '0',
```

## Fix shape
Thread a slippage tolerance (bps) into `getNativeSwapQuote` → pass `tolerance_bps` on the quote request, and compute `toAmountLimit` from `expected_amount_out` minus the tolerance (or have THORChain emit a non-zero `LIM=` via the limit param) so the signed memo carries a real minimum-output floor. Default to a sane bound (e.g. 1-3%) and surface it in the approval card.

---

rcoderdev:
Closes #688

## Release Notes

* **New Features**
  * Native swap slippage tolerance is now configurable with a default of 100 basis points
  * Configured tolerance is applied consistently across quote requests and swap executions

* **Bug Fixes**
  * Native swap minimum-output limits previously hardcoded to zero; now properly reflect the configured slippage tolerance

### DIFF
--- a/.changeset/native-swap-min-output.md
+++ b/.changeset/native-swap-min-output.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Apply native THORChain/Maya swap slippage tolerance to quote requests and signed payload limits so native swaps no longer use a zero minimum-output floor.
--- a/packages/core/chain/swap/native/NativeSwapQuote.ts
+++ b/packages/core/chain/swap/native/NativeSwapQuote.ts
@@ -14,6 +14,7 @@ export type NativeSwapQuote = {
   outbound_delay_blocks: number
   outbound_delay_seconds: number
   recommended_min_amount_in: string
+  liquidity_tolerance_bps?: number
   max_streaming_quantity?: number
   slippage_bps?: number
   total_swap_seconds?: number
--- a/packages/core/chain/swap/native/api/getNativeSwapQuote.test.ts
+++ b/packages/core/chain/swap/native/api/getNativeSwapQuote.test.ts
@@ -71,7 +71,44 @@ describe('getNativeSwapQuote', () => {
     expect(queryUrlMock).toHaveBeenCalledTimes(1)
     const url = String(queryUrlMock.mock.calls[0][0])
     expect(url).toContain('streaming_interval=0')
+    expect(url).toContain('liquidity_tolerance_bps=100')
     expect(url).not.toContain('streaming_quantity')
+    expect(quote.liquidity_tolerance_bps).toBe(100)
+  })
+
+  it('THORChain: forwards custom slippage tolerance bps to the quote request', async () => {
+    queryUrlMock.mockResolvedValueOnce({
+      ...baseOkBody,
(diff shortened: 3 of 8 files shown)


gomesalexandre (approved):
fund-safety critical fix, gucci ser

`toAmountLimit: "0"` was zero slippage protection on all native thor/maya swaps - users were fully exposed to sandwich + price impact with no floor. this closes that gap properly.

**math check:** `getNativeSwapToAmountLimit` = `(expected × (10000 - tolerance)) / 10000` with BigInt truncation. floors toward zero (conservative direction - if anything the minimum output is slightly lower than float would give, which is fine for a protection floor). 250bps fixture in the test: `100000 × 9750 / 10000 = 97500`. ✓

**flow check:** `findSwapQuote.ts` → `toProviderSlippage.nativeBps` → `getNativeSwapQuote(slippageToleranceBps)` → `liquidity_tolerance_bps` query param to API + stored on quote object → `nativeSwapQuoteToSwapPayload` reads it back for `toAmountLimit`. end-to-end, solid.

**test coverage:** unit test on the math function + integration test on the full payload. both catch the right thing.

**nit:** `assertValidSlippageToleranceBps` is defined twice - once in `getNativeSwapQuote.ts` (private const) and again in `nativeSwapQuoteToSwapPayload.ts` (also private const, slightly different error message since it uses `${BASIS_POINTS_DENOMINATOR}` string interpolation). `nativeSwapQuoteToSwapPayload.ts` already imports `DEFAULT_NATIVE_SWAP_SLIPPAGE_TOLERANCE_BPS` from `getNativeSwapQuote` - could also export + import the validator from there. non-blocking, just noting the duplication.

**q:** Maya chain - `swapChain !== Chain.THORChain` path also calls `requestNativeSwapQuote` with `liquidity_tolerance_bps` in the query string. does Maya's `/mayachain/quote/swap` endpoint actually accept this param, or does it silently ignore it? if Maya ignores it, the `toAmountLimit` is still computed from the quote's stored `liquidity_tolerance_bps` (100bps default) which would be wrong - the protection math would assume 1% was applied but Maya never actually enforced it. worth a quick smoke test or API doc check.

PR #809, opened 2026-06-23
Validate native swap expiry and vault

### LINKED ISSUE #690: No payload-expiry guard; THORChain inbound_address never re-fetched before broadcast (rotation → funds to dead vault)
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` native-swap path).

## Observed
There is **no payload age/expiry guard**, and the THORChain `inbound_address` (the vault the funds are deposited to) is **never re-fetched before broadcast**. THORChain rotates inbound vault addresses on churn; if a quote is held (slow keysign, user delay) the stamped inbound address can become a retired vault and **funds are sent to a dead address** — unrecoverable. The quote carries an `expiry`, but the build path ignores it and stamps a fresh 15-minute expiration instead.

## Cause
The build path stamps a brand-new `Date.now()+15min` and reuses the quote's `inbound_address` verbatim — the genuine quote `expiry` is discarded:

`packages/core/mpc/swap/native/utils/nativeSwapQuoteToSwapPayload.ts:46`
```ts
expirationTime: BigInt(Math.round(convertDuration(addMinutes(Date.now(), 15).getTime(), 'ms', 's'))),
```
`vaultAddress: quote.inbound_address ?? fromCoin.address` is set once from the quote and never refreshed. The quote's real `expiry` (`packages/core/chain/swap/native/NativeSwapQuote.ts:7`) is read nowhere on the build/broadcast path, and there is no `inbound_addresses` re-fetch/compare before signing.

## Fix shape
(1) Honor the quote's `expiry`: reject (or re-quote) if `Date.now() > quote.expiry` rather than stamping a fresh window. (2) Immediately before broadcast, re-fetch THORChain `inbound_addresses` and assert the stamped `inbound_address` still matches the active vault for the source chain; abort + re-quote on mismatch.

---

rcoderdev:
Closes #690

## Summary

- Honor native swap quote expiry when creating native swap signing payloads.

- Guard SDK broadcast against expired native swap payloads and stale THORChain inbound vault addresses.

## Verification
- yarn check:ci
- yarn test
- Windows local-tarball install + desktop smoke at http://localhost:34115/ rendered the real portfolio screen; screenshot: /Users/rodion/vultisig/windows/.codex/receipts/sdk-690-windows-desktop-smoke.png
- Windows yarn check fails on current Windows main and with local SDK tarballs due missing locale keys de/es/it/hr/ru/nl/pt/zh/ko.swap_invalid_external_recipient; this is not introduced by the SDK tarballs.

## Coordination

- PR #808 / SDK #688 also edits packages/core/mpc/swap/native/utils/nativeSwapQuoteToSwapPayload.ts and its test. Read-only merge-tree shows conflict markers between the branches, so whichever PR lands second will need explicit conflict resolution; this PR does not merge sibling branch assumptions.

## Release Notes

* **Bug Fixes**
  * Native swap quotes now properly honor their expiration times instead of auto-extending them.
  * THORChain swaps now validate inbound vault addresses before broadcasting to prevent stale transaction issues.

* **Tests**
  * Added unit tests for native swap broadcast validation logic.

### DIFF
--- a/.changeset/native-swap-expiry-vault.md
+++ b/.changeset/native-swap-expiry-vault.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Honor native swap quote expiry and validate THORChain inbound vault addresses before broadcasting stale signed swaps.
--- a/packages/core/chain/chains/cosmos/thor/getThorchainInboundAddress.ts
+++ b/packages/core/chain/chains/cosmos/thor/getThorchainInboundAddress.ts
@@ -3,7 +3,7 @@ import { queryUrl } from '@vultisig/lib-utils/query/queryUrl'
 import { Chain } from '../../../Chain'
 import { cosmosRpcUrl } from '../cosmosRpcUrl'
 
-type ChainInfo = {
+export type ThorchainInboundAddress = {
   address: string
   chain: string
   chain_lp_actions_paused: boolean
@@ -22,4 +22,4 @@ type ChainInfo = {
 
 const thorchainInboundAddressApi = `${cosmosRpcUrl[Chain.THORChain]}/thorchain/inbound_addresses`
 
-export const getThorchainInboundAddress = (): Promise<ChainInfo[]> => queryUrl(thorchainInboundAddressApi)
+export const getThorchainInboundAddress = (): Promise<ThorchainInboundAddress[]> => queryUrl(thorchainInboundAddressApi)
--- a/packages/core/mpc/swap/native/utils/nativeSwapQuoteToSwapPayload.test.ts
+++ b/packages/core/mpc/swap/native/utils/nativeSwapQuoteToSwapPayload.test.ts
@@ -1,0 +1,78 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import type { AccountCoin } from '@vultisig/core-chain/coin/AccountCoin'
+import type { NativeSwapQuote } from '@vultisig/core-chain/swap/native/NativeSwapQuote'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { nativeSwapQuoteToSwapPayload } from './nativeSwapQuoteToSwapPayload'
+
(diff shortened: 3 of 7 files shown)


gomesalexandre (approved):
fund-safety improvement. two layers is the right approach.

**layer 1 (payload creation):** `assertQuoteNotExpired(quote.expiry)` at `nativeSwapQuoteToSwapPayload` call time rejects before the signing ceremony starts. stops a user from signing into an expired quote.

**layer 2 (broadcast):** `assertNativeSwapReadyForBroadcast` in `BroadcastService.broadcastTx` re-checks expiry + vault freshness RIGHT before sending. catches the case where signing takes time and the vault rotates mid-ceremony.

**`expirationTime: BigInt(quote.expiry)`** replacing the hardcoded `+15 min` is the right fix. the old value was a local wall-clock guess unrelated to what THORChain considers the valid window for this specific quote. now `expirationTime` in the signed payload actually matches the quote's expiry on the node.

**vault address staleness check:** checking `activeInboundAddress !== native.vaultAddress` before broadcast is critical - if THORChain rotates vaults between quote fetch and broadcast, funds could end up in an old vault. guard throws a clear message to refresh the quote. ✓

**test coverage:**
- `nativeSwapQuoteToSwapPayload.test.ts`: pins `expirationTime === quote.expiry`, plus expired-quote rejection
- `nativeSwapBroadcastGuard.test.ts`: expired pre-check (no unnecessary inbound fetch), Maya chain (no vault check), vault mismatch, happy path
- injectable `now` and `getInboundAddresses` for deterministic testing. ✓

**heads up - merge conflict risk with #808:** `nativeSwapQuoteToSwapPayload.ts` is also modified by #808 (fund-safety fix for `toAmountLimit: "0"` → computed slippage). Both PRs remove the `convertDuration`/`addMinutes` imports (this PR) and add new imports + change different lines (#808). 3-way merge should be clean since they touch different lines - but worth verifying the combined diff once one of them lands, to ensure `toAmountLimit` still uses `getNativeSwapToAmountLimit(...)` after both merge. this PR branches from before #808 so it still shows `toAmountLimit: "0"` in context.

PR #812, opened 2026-06-23
feat(sdk): add sdk.gas.compareCosts(chains) (fan-out EVM gas price)

gomesalexandre:
## what + why

Brings the mcp-ts `compare_gas_costs` primitive into the SDK as `sdk.gas.compareCosts(chains)` — a pure fee-math + RPC-read that fans out the live `eth_gasPrice` across a set of EVM chains and ranks them cheapest-first by estimated tx cost. It answers "which chain is cheapest to transact on right now?" in one fan-out call instead of N sequential gas-price reads. Per the tools→SDK FULL table this is a `BRING-INTO-SDK` row (pure fee math, 0 SDK imports): `compare_gas_costs → sdk.gas.compareCosts(chains)`.

Reads go through the SDK's existing `getEvmClient(chain)` (viem), so it inherits the SDK's RPC config / custom-RPC override rather than re-implementing a raw `evmRpc` helper. It's fail-soft: a chain whose RPC errors lands in `skipped[]`, never rejecting the whole call. USD pricing is **injected** via an optional `nativeUsdPrices` map (pure multiply) instead of fetched — that keeps the primitive a pure fee-math + RPC-read with zero external-API coupling (the mcp-ts version reached out to CoinGecko; that price-fetch concern stays at the orchestration layer where the caller already has a price cache).

Exposed three ways to match SDK convention: flat named export `compareCosts`, the `sdk.gas.compareCosts(...)` namespace ergonomic from the spec, and the helper `getChainGasPriceGwei(chain)` + `GAS_UNITS` / `DEFAULT_COMPARE_CHAINS` constants.

## consumers this unblocks

- **isolate / co-sign**: a uniform gwei oracle for fee floors without a per-chain round-trip
- **CLI**: `compareCosts(["Ethereum","Base","Arbitrum"])` → cheapest-chain answer for routing
- **app (Station)**: powers a "cheapest chain to move on" hint on send/swap surfaces off one call
- **mcp-ts**: the `compare_gas_costs` tool can become a thin wrapper over `sdk.gas.compareCosts` (drops its duplicated RPC + ranking logic), same as the other already-consumed SDK primitives

## receipt

Live RPC fan-out across Ethereum / Base / Arbitrum (no mocks):

```
$ node scripts/receipts/gas_compare_costs.mjs

compareCosts(["Ethereum","Base","Arbitrum"]) — txType=transfer, gasUnits=21000

  Base            0.006 gwei   est tx cost: 1.260e-7 native
  Arbitrum       0.0201 gwei   est tx cost: 4.221e-7 native
  Ethereum       0.3022 gwei   est tx cost: 6.346e-6 native

  cheapest → Base (1.260e-7 native)
```

Verification:

```
yarn workspace @vultisig/sdk typecheck            # clean (exit 0)
yarn workspace @vultisig/sdk test:unit -t compareCosts
  ✓ ranks chains cheapest-first by native tx cost
  ✓ computes USD cost when native prices are injected and ranks by USD
  ✓ is fail-soft: a failing RPC lands in skipped, not a rejection
  ✓ returns cheapest=null when every chain errors
  Test Files  1 passed | 110 skipped (111)
```

## risk

Low — additive only (2 index re-export edits + a new `tools/gas` module). No existing primitive touched, no signing path, read-only RPC.

## Summary by CodeRabbit

* **New Features**
  * Added gas cost comparison across multiple EVM chains, with a cheapest-first ranking.
  * Added a convenient `sdk.gas` namespace plus a flat `compareCosts` export for easier access.
  * Added support for optional USD-based cost estimates when pricing data is available.

* **Bug Fixes**
  * Improved reliability by skipping only failed chain lookups instead of failing the entire request.
  * Preserved usable results even when some chains don’t return pricing data.


gomesalexandre (commented):
## adversarial pass — found + fixed 1, rest clean

Deep adversarial review of the gas fan-out (per-chain pinning / bigint / cross-chain bleed / cheapest-selection / the sibling u64+address bug class). Found and pushed one cheapest-selection fix (`b2c9f4b8`), everything else verified clean.

### found + fixed — `src/tools/gas/compareCosts.ts:118-128` (correctness, cheapest-selection)
The cheapest-first sort used a **per-pair mixed-basis comparator**: `estTxCostUsd` when *both* entries had a price, `estTxCostNative` otherwise. `nativeUsdPrices` is `Partial<Record<EvmChain, number>>` — a partial map is explicitly allowed — so this is reachable, and it's **non-transitive**: native cost across different native tokens isn't comparable (1e-4 ETH ≠ 1e-4 MATIC). With e.g. `{ Ethereum: 3000 }` injected but Polygon omitted, two equal-gwei chains tie on native and a ~$1.89 Ethereum tx can get crowned `cheapest` over a ~$0.0000126 Polygon one. The mcp-ts source dodged this by *skipping* unpriced chains entirely (all-USD → single basis); the port kept them but dropped the single-basis invariant.

Fix: rank on **one** basis — USD only when *every* surviving result is priced, else native wholesale (the documented gwei-only best-effort). Never mix per-pair. +2 regression tests (partial map must not leak USD into `cheapest`; full map must USD-rank).

### verified clean
- **per-chain pinning / cross-chain bleed**: `getEvmClient` is memoized on `${chain}:${rpcUrl}` (per-chain client); `Promise.allSettled` preserves order, `settlements.forEach((s,i)=>chains[i])` is index-aligned. No bleed.
- **bigint / u64-precision (sibling-bug class)**: `Number(wei)/1e9` is on a **gas price** (≤ ~1e12 wei even at absurd gwei) ≪ 2^53 — NOT a u64 balance. The JSON-number rounding that bit balance/amount ports does not apply to fee math here. Safe.
- **address validation**: N/A — no addresses parsed/validated in this primitive (read-only `eth_gasPrice` + fee math).
- **fail-open**: per-chain RPC errors land in `skipped[]`, all-fail → `cheapest=null`, never throws / never fabricates a price. No fail-open.
- **constants/decimals**: `GAS_UNITS` 21k/150k and the 7 DEFAULT_COMPARE_CHAINS all match mcp-ts; all 7 are valid `EvmChain` members.

### receipts
```
typecheck: exit 0 (clean)
test:unit -t compareCosts: 6 passed (4 orig + 2 new regression)
node scripts/receipts/gas_compare_costs.mjs (live RPC, no mocks):
  Base       0.006  gwei  est 1.260e-7 native
  Arbitrum   0.0201 gwei  est 4.221e-7 native
  Ethereum   0.2768 gwei  est 5.813e-6 native
  cheapest → Base (1.260e-7 native)  ✓
```

(comment-only on my own PR — not an approval.)

packages/sdk/src/tools/gas/compareCosts.ts
```diff
@@ -0,0 +1,140 @@
...
+  /** The cheapest entry, or `null` when every chain errored. */
+  cheapest: Pick<CompareCostsEntry, 'chain' | 'estTxCostNative' | 'estTxCostUsd'> | null
+  /** Chains that errored during the RPC read (fail-soft — never throws). */
+  skipped: CompareCostsSkipped[]
+}
+
+/** Fetch the current gas price (gwei, 4 dp) for a single EVM chain. */
+export const getChainGasPriceGwei = async (chain: EvmChain): Promise<number> => {
+  const wei = await getEvmClient(chain).getGasPrice()
+  return parseFloat((Number(wei) / 1e9).toFixed(4))
```
NeOMakinG, line 72: `should-fix:` `toFixed(4)` rounds gwei here, and this rounded value flows into `estTxCostNative` → the ranking sort. Two chains differing by <0.0001 gwei collapse to the same key and fall back to input order rather than true cost. Negligible fund impact (both effectively free at that delta), but ranking on the raw-precision gwei and rounding only for the `gasPriceGwei` display field would be strictly correct.

packages/sdk/src/tools/gas/compareCosts.ts
```diff
@@ -0,0 +1,140 @@
...
+  nativeUsdPrices?: Partial<Record<EvmChain, number>>
+}
+
+export type CompareCostsEntry = {
+  chain: EvmChain
+  /** Current gas price in gwei (rounded to 4 dp). */
+  gasPriceGwei: number
+  /** Native-token USD price if provided in `nativeUsdPrices`, else `null`. */
+  nativeUsd: number | null
+  /** Estimated tx cost in the chain's native token (gwei × gasUnits × 1e-9). */
```
NeOMakinG, line 47: `should-fix:` `estTxCostNative = eth_gasPrice × fixed gasUnits` omits the L1 data-fee component that Base/Arbitrum/Optimism charge, which `eth_gasPrice` does not capture. A rollup can rank as cheapest while its total user fee is higher — most misleading on the `swap` (150k) path. The honest scope is execution-gas heuristic; worth a doc note that this is not total fee, so consumers don't read `cheapest` as authoritative total cost.

NeOMakinG (approved):
## testing approach

Three-lane fan-out: Claude (load-bearing, correctness + fund-safety + fan-out isolation), Codex gpt-5.4 (adversarial second opinion — precision/ranking/cross-chain-cost), GLM (dead, ECONNRESET — best-effort skipped). Built `@vultisig/core-chain`, ran the unit suite (6/6), tsc (0 err), prettier, complexity (sonarjs) + lint (0), the changeset guard, and the live RPC receipt against Ethereum/Base/Arbitrum. Mutation-tested the ranking + fail-soft logic. Verified the RN bundle export at the `dist/` level.

**Verdict: APPROVE.** Read-only fee-math fan-out, unit-consistent, fail-soft per chain (no fabricated cheapest), comparison documented honestly. I fixed two structural blockers in-code on the branch (`fa581700`): the RN export gap and the missing changeset. Three should-fix items from the Codex lane are documented caveats of an informational tool — surfaced below, not blocking.

## what I verified

- **Unit consistency** — does NOT reuse #813 `evmGasPrice` (that primitive isn't on main yet; `src/tools/gas/` is created by this PR). Re-implements via `getEvmClient(chain).getGasPrice()` — same viem `eth_gasPrice` call. Every chain's gwei is computed identically, so the ranking is apples-to-apples on the same unit. No wei/gwei mixup. When #813 lands the two can be reconciled (a follow-up `q` below).
- **Fan-out isolation** — `Promise.allSettled` per chain; a rejected RPC lands in `skipped[]` with its error message and is excluded from `results`/`cheapest`. A failed chain is NEVER fabricated to a 0 price (which would wrongly crown it cheapest). Mutation: forcing the rejected branch to fall through → 2 tests fail. All-fail → `cheapest: null`, never throws.
- **Ranking basis consistency** — sorts on a SINGLE basis: USD only when EVERY surviving result is priced (`allPriced` gate), else native cost wholesale. Never mixes USD/native per-pair (non-transitive). Mutation: forcing `allPriced=true` on a partial map → 1 test fails; reversing the comparator → 4 fail.
- **Math is float, not BigInt** — wei→gwei is `Number(wei)/1e9`; cost is `gwei * 1e-9 * gasUnits`. For real gas prices (≤ ~thousands of gwei = ≤1e15 wei) this is well under `Number.MAX_SAFE_INTEGER`, so ranking is sound in practice. See the should-fix on `toFixed(4)` below.
- **Live receipt** (real RPC, no mocks): `Base 0.006 gwei < Arbitrum 0.02 < Ethereum 0.0792` → cheapest Base. Correct.

## blocking

None outstanding — both fixed in-code on `fa581700`.

## should-fix

1. **`toFixed(4)` rounds gwei before ranking** (Codex lane, `compareCosts.ts:72`). `getChainGasPriceGwei` rounds to 4 dp, and that rounded value feeds `estTxCostNative` → ranking. Two L2s differing by <0.0001 gwei collapse to an identical key and then fall back to input order. The fund impact is negligible (both are effectively free at that delta), but ranking on raw-precision gwei and rounding only for the display field would be strictly more correct.
2. **Native cross-chain `cheapest` is not a true-value ranking** (Codex lane). When `nativeUsdPrices` is partial/empty, `cheapest` is decided on `estTxCostNative` — but `0.0001 MATIC` is not cheaper-in-value than `0.0002 ETH`. The code documents this in the comment block at L118-124 and the partial-map test proves no USD value leaks in, so it's honest — but a caller that reads `cheapest` blindly without USD prices gets a native-unit answer. Consider surfacing the ranking basis in the result (e.g. `rankedBy: 'usd' | 'native'`) so consumers can't silently misread it.
3. **`eth_gasPrice × fixed gasUnits` omits the L2 data fee** (Codex lane). On Base/Arbitrum/Optimism the real user cost includes an L1 data/blob component that `eth_gasPrice` does not capture, so a rollup can be ranked cheapest while its total fee is higher — most misleading on the `swap` path with a uniform 150k. The field is named `estTxCostNative` and the PR scopes itself as an execution-gas heuristic, so this is a labeling/wording sharpening, not a math bug. Worth a doc note that this is execution-gas only, not total fee.

## q

- Once #813 `evmGasPrice` merges, fold `getChainGasPriceGwei` onto it (or have #813 consume this) so there's one gas-price read path. Tracking only.

## risk

Low. Additive, read-only RPC, no signing path. Fails soft per chain and fails closed (`cheapest: null`) when all chains error. The two structural fixes are mechanical (allow-list re-export + changeset) and verified at the bundle level.

---
*Reviewed by Claude (load-bearing) + Codex gpt-5.4 (adversarial). GLM lane dead (ECONNRESET). Two blockers fixed in-code on `fa581700`; gomes fix-and-approve.*

PR #813, opened 2026-06-23
feat(sdk): add sdk.gas.price (EVM gas price per chain)

gomesalexandre:
## what

Adds `evmGasPrice(chain)` — a pure read-only primitive that fetches the current EVM gas price (`eth_gasPrice`) for any supported EVM chain via the SDK's own per-chain RPC client. Returns both the exact `gasPriceWei` (`bigint`, no precision loss) and a `gasPriceGwei` convenience field for display.

Also exposes the previously-internal `GasEstimationService` for callers that already hold a vault and need the richer chain-specific fee shape.

## why

Ported from mcp-ts `tools/fee/gas-price.ts` (0 SDK imports), keeping ONLY the pure-crypto half — the CoinGecko USD overlay is an orchestration concern (needs a price oracle) and stays out of the SDK. No external API, no extra config: it reuses `getEvmClient` (viem `PublicClient`) so it inherits whatever RPC the SDK is already configured with.

## consumers it unblocks

- mcp-ts `get_gas_price` (EVM branch) → wraps `sdk.gas.price` instead of duplicating RPC logic
- mcp-ts `compare_gas_costs` fan-out → shares one canonical gas read
- agent-backend fee-estimation / swap-card "Est. fee" rows → single gas source across surfaces

## receipt

LIVE network read against the new primitive (the curl-equivalent):

```
$ node --import tsx scripts/receipts/gas_price_evm.mjs

sdk.gas.price — live EVM gas price

  Ethereum    0.3073 gwei  (307299145 wei)
  Base        0.006 gwei  (6000000 wei)
```

Unit test (offline, mocked client — wei→gwei conversion, sub-gwei L2, whole-gwei, 0-wei edge cases):

```
$ vitest run tests/unit/evmGasPrice.test.ts

 ✓ returns the raw wei value untouched (exact bigint)
 ✓ converts wei → gwei rounded to 4 decimals
 ✓ handles whole-gwei values exactly
 ✓ handles sub-gwei (L2) gas prices without underflowing to 0
 ✓ returns 0 gwei for a 0 wei gas price

 Test Files  1 passed (1)
      Tests  5 passed (5)
```

`yarn workspace @vultisig/sdk typecheck` → 0 errors (0 new).

## risk

Low — additive, pure read, no signing path touched. `bigint` kept exact; `Number()` only used for the display-only gwei field (EVM gwei magnitudes stay well under `MAX_SAFE_INTEGER`).


gomesalexandre (commented):
Deep adversarial pass on the gas-price primitive (production crypto, own PR, comment-only). 

## checked
- **wei exactness (the sibling u64 bug class)**: `gasPriceWei` is a true viem `bigint` — viem parses hex quantities into bigint, never round-trips through a JS number. Verified the exact value survives even past 2^53 (added a 123e27-wei test). Math callers get the raw bigint, never the gwei field. Clean.
- **wrong-chain RPC pinning**: `getEvmClient(chain)` builds the viem client with `chain: evmChainInfo[chain]` (carries the chainId) AND `transport: http(getEvmRpcUrl(chain))`, memoized on `${chain}:${rpcUrl}` so a custom-RPC override yields a fresh client. Both `evmChainRpcUrls` and `evmDefaultChainInfo` are compiler-total `Record<EvmChain, …>`, so no missing-chain hole and no cross-chain bleed. Clean.
- **fail-open**: unreachable RPC throws and propagates (receipt catches + exit 1). No silent fallback. Clean.
- **phantom deps / decimals**: only `getEvmClient` + viem; 1 gwei = 1e9 wei constant correct. Clean.

## found + fixed (gasPrice.ts:34, LOW, display-only)
The `gasPriceGwei` display field did `parseFloat((Number(wei)/1e9).toFixed(4))`. A non-zero price below the 4-decimal floor (`< 50_000 wei`) rounds to a misleading `0` — e.g. `49_999 wei` and `1 wei` both displayed `0 gwei`, which on an L2 reads as 'free gas'. Fixed: clamp a genuinely non-zero price UP to the smallest renderable `0.0001 gwei`; a real `0n` still shows `0`. `gasPriceWei` was always correct and is unchanged — this is cosmetic, no fund-safety impact. Pushed in 18d53975.

## receipts
- unit: 8/8 pass (added 49_999-wei + 1-wei clamp tests, genuine-0 test, beyond-2^53 exactness test)
- typecheck: 0 errors (sdk package)
- live receipt: `Ethereum 0.276 gwei (276010462 wei)` / `Base 0.006 gwei (6000000 wei)` — exact wei + sane gwei

Verdict: solid primitive. wei-exact ✓, chain-pinned ✓, no fail-open ✓. One display-rounding nit fixed.

NeOMakinG (approved):
read/compute review — `evmGasPrice` (EVM gas price per chain). Three lanes: Claude (load-bearing, full source walk + 2 mutation kills + live receipt + RN bundle build), Codex gpt-5.4 (0-byte stall, session contention — best-effort), GLM 4.6 (429 insufficient balance — dead, best-effort). Two recurring gomes gaps fixed in-code on the branch.

## testing approach

- read `gasPrice.ts`, the test, the receipt, the EVM client (`getEvmClient` → viem `PublicClient`), and the actual SDK fee path (`getEvmBaseFee` / `getEvmTwFeeFields`) for cross-consistency
- compared the port against mcp-ts `tools/fee/gas-price.ts` (`getEvmGasPrice`) — faithful port of the wei-exact half, and an improvement (mcp-ts collapses sub-floor to `0`; this clamps up)
- unit: 8/8 green via `vitest run tests/unit/evmGasPrice.test.ts`
- mutation kills: invert wei↔gwei (`/` → `*`) → 5 fail; remove the sub-floor clamp (fail-open to `0`) → 2 fail. Tests have real teeth on both the unit and the no-fabricated-0 invariants.
- live receipt `node --import tsx scripts/receipts/gas_price_evm.mjs` → Ethereum 0.0977 gwei (97700394 wei), Base 0.006 gwei (6000000 wei) — real RPC, wei↔gwei consistent
- gates (local, after `@balancer/sdk` reinstall — a worktree dep artifact from the main-merge, not the diff): tsc 0 errors, eslint clean, prettier clean, RN bundle build includes `evmGasPrice`, changeset guard passes. `gh pr checks` green.

## correctness — the fund-relevant lens

**Unit (wei vs gwei): correct, unambiguous.** `gasPriceWei` is the raw `bigint` from `client.getGasPrice()` (viem `eth_gasPrice`), never round-tripped through a JS number — proven by the `>2^53` test. `gasPriceGwei` is explicitly typed + documented as display-only, `WEI_PER_GWEI = 1_000_000_000n`. No 10^9 mixup, no float math on the exact value.

**Fail-closed: yes, by construction.** No try/catch. viem `getGasPrice()` rejects on RPC error / malformed response → `evmGasPrice` rejects. The only `0` returned is a genuine `0n` from the node; a non-zero-but-sub-floor price is clamped UP to `0.0001` gwei rather than rendered as a misleading `0` ("free gas" on an L2). No fabricated 0, no absurd value.

**1559 vs legacy / cross-consistency.** This is an informational `eth_gasPrice` read — it returns the node's single legacy-equivalent gas price, not `maxFeePerGas`/`maxPriorityFeePerGas`. That matches mcp-ts exactly (the consumer it unblocks). It is NOT the SDK's signing fee source: the real tx builder uses `getEvmBaseFee` (`getBlock().baseFeePerGas`) + a separate priority fee via `getEvmTwFeeFields`. So this primitive can't under/over-price a real tx — it never feeds the signing path. See the q below.

## should-fix

(none blocking — both recurring gaps fixed in-code; see below)

## fixed in-code (commit 36afcb1c)

1. **RN export gap** — `evmGasPrice` was missing from `src/platforms/react-native/index.ts` (the allow-list had abiDecode/abiEncode/evmCall/evmCheckAllowance/evmTxInfo/resolve4ByteSelector/resolveEns but not the new tool). React Native consumers — including vultiagent-app, the stated unblock target — resolve through the `react-native` export condition, so the tool shipped **unreachable from the app**. Added the value + `EvmGasPrice` type; verified `evmGasPrice` now appears in the built `dist/index.react-native.js`. (Recurring across #814/#816/#817/#818.)
2. **Missing changeset** — the bundled-changeset CI guard (`scripts/check-sdk-bundled-changeset.mjs`) only watches `packages/core|lib` prefixes, so a `packages/sdk/src`-only PR passes **green without a changeset** — then no new SDK tarball is published and no consumer ever receives the export. Added `.changeset/sdk-gas-price-evm.md` (`@vultisig/sdk: minor`). (Recurring; same prefix blind-spot every SDK-tool PR.)

## q

- `evmGasPrice` (eth_gasPrice) diverges from the SDK's own signing fee source (`getEvmBaseFee` baseFeePerGas + priority). Intentional and safe here (display/informational only, matches the mcp-ts consumer), but worth a one-line doc note if a future caller is ever tempted to feed `gasPriceWei` into tx construction on a 1559 chain — `eth_gasPrice` is a rough legacy estimate, not the 1559 fee fields the builder needs.

## risk

Low. Additive pure read, no signing path touched, fails closed, unit exact (`bigint`), no fabricated 0. The two fixes are the only thing standing between "green CI" and "actually shipped to consumers" — without them the PR's entire stated purpose (unblock mcp-ts `get_gas_price` + agent-backend fee rows) was silently blocked.

## verdict

APPROVE. Unit correct + unambiguous, fail-closed (no fabricated 0), 1559/legacy understood and non-load-bearing, tests prove unit + no-fabricated-0 (mutations killed), RN export + changeset added in-code, all gates green.

— reviewed by Claude (load-bearing) + Codex gpt-5.4 (stalled, best-effort) + GLM 4.6 (dead, best-effort)

PR #814, opened 2026-06-23
feat(sdk): add sdk.balance.evm (native + ERC-20 via RPC)

gomesalexandre:
## what

Adds `sdk.balance.evm` — a multi-token EVM balance read (`getEvmBalances`): native coin + any number of ERC-20s in a single batched RPC fan-out. Ported from mcp-ts `src/tools/balance/evm-balance.ts` (0 SDK imports), but rewired onto the SDK's own viem stack instead of the raw `eth_call` calldata-hand-rolling the tool did.

- Native: `getEvmClient(chain).getBalance()`, with symbol/decimals from `chainFeeCoin[chain]` (no hardcoded ticker map).
- ERC-20: `balanceOf` + `decimals` + `symbol` via viem `erc20Abi` (same pattern as the existing `evmCheckAllowance`) so unknown tokens resolve on-chain without a registry.
- Returns `{ contractAddress?, symbol, decimals, raw: bigint, balance: string }[]` — raw base-units (no precision loss) plus a human-readable string. Native is always index 0.

Pure crypto/RPC primitive, vault-free. `getCoinBalance` already covers the single-coin path; this is the multi/RPC path a portfolio view needs.

## why / consumers it unblocks

- `mcp-ts` `evm_get_balance` / `evm_get_token_balance` can stop hand-encoding `0x70a08231…` calldata + buffer-slicing the symbol return and call one SDK primitive.
- `agent-backend` portfolio/holdings fan-out gets a single-call native+token read per chain.

## test matrix

- Unit: `tests/unit/tools/evm/balanceEvm.test.ts` — native-only, native+ERC-20, sub-unit precision, per-chain native ticker (BNB on BSC). 4/4 pass (full suite 1514/1514).
- Typecheck: `yarn workspace @vultisig/sdk typecheck` → 0 errors, 0 new.
- Lint: eslint clean on the changed src files.
- Live receipt below — real Ethereum RPC (`api.vultisig.com/eth/`, the SDK default), real balances for vitalik.eth.

## RECEIPT

```
$ yarn tsx scripts/receipts/balance_evm.mjs

sdk.balance.evm — Ethereum — 0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045 (vitalik.eth)

  ETH        5.690864439481747771  (18d, raw=5690864439481747771)  [native]
  USDC                  36.769066  (6d, raw=36769066)  [ERC-20 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48]
  DAI         4.57207827332311288  (18d, raw=4572078273323112880)  [ERC-20 0x6B175474E89094C44Da98b954EedeAC495271d0F]
```

(live RPC, no mocks — this is the curl-equivalent for a read primitive)

## Release Notes

* **New Features**
  * Added `sdk.balance.evm` (`getEvmBalances`) to fetch native coin plus multiple ERC-20 balances in one call, returning both human-readable and raw base-unit values.
  * Exposed the new types and `getEvmBalances` from the SDK main exports and the React Native entry point.
  * Added a runnable receipt script to query and print EVM balances for selected token contracts.
* **Bug Fixes**
  * Balance lookups now reject if token metadata reads (e.g., `decimals`) fail, instead of silently falling back.
* **Tests**
  * Added unit tests for ordering, mixed-decimal formatting, and error propagation.


gomesalexandre (commented):
## adversarial pass clean — verified, hardened with regression guards

Deep adversarial review of `getEvmBalances` (native + ERC-20 multi-token RPC fan-out). No logic bug found; the impl is correct. Pushed 1 commit (`9c321c80`) that locks the load-bearing behavior under tests — no source change.

### what I checked (and why it's safe)
- **Per-token decimals / wrong-decimals**: native decimals+ticker come from the verified `chainFeeCoin[chain]` registry (`Record<Chain, KnownCoin>` — every `EvmChain` has an entry, `nativeCoin` is never `undefined`). ERC-20 decimals are read on-chain per token. Concurrent reads via `Promise.all` keep each token's `decimals` scoped to its own closure — no cross-token bleed. Locked with a mixed-decimal (6 vs 18) test.
- **Wrong-chain**: `getEvmClient(chain)` binds transport+`chain` to `evmChainInfo[chain]` / `getEvmRpcUrl(chain)`, memoized on `chain:rpcUrl`. The RPC and chainId are pinned to the requested chain — no way to read a token on the wrong chain.
- **bigint precision / overflow**: `raw` stays `bigint` end-to-end, no float coercion. `formatUnits` is **byte-identical to viem's `formatUnits`** across a parity suite (1 wei, sub-unit dust, trailing-zero `1.5`, 30-digit values). Only divergence is negative input — unreachable, since `balanceOf`/`getBalance` are unsigned uint256.
- **fail-open**: a reverting `decimals()`/`symbol()` (non-standard or non-contract address) rejects the whole `Promise.all` — fail-**closed**, never a fabricated `0` balance. Locked with a reverting-read test.
- **address validation**: no runtime `isAddress`/`getAddress` — but this matches the rest of the SDK evm tool suite (`evmCall`, `checkAllowance`, `evmTxInfo` all rely on the `0x${string}` type, unlike the mcp-ts MCP tools which take untyped string input). Consistent, not a regression. Didn't force a divergent guard here.

### receipts
- unit: **6/6 pass** (4 original + 2 added regressions)
- typecheck: **0 errors** (`tsc --noEmit`)
- live RPC receipt (real Ethereum, vitalik.eth):
```
  ETH        5.690864439481747771  (18d, raw=5690864439481747771)  [native]
  USDC                  36.769066  (6d, raw=36769066)  [ERC-20 0xA0b8…eB48]
  DAI         4.57207827332311288  (18d, raw=4572078273323112880)  [ERC-20 0x6B17…1d0F]
```
native + both ERC-20s read with correct per-token decimals and exact bigint formatting. all gucci.

packages/sdk/src/tools/evm/balanceEvm.ts
```diff
@@ -0,0 +1,105 @@
...
+        }),
+        client.readContract({
+          address: contractAddress,
+          abi: erc20Abi,
+          functionName: 'decimals',
+        }),
+        client.readContract({
+          address: contractAddress,
+          abi: erc20Abi,
+          functionName: 'symbol',
```
NeOMakinG, line 83: `should-fix:` `symbol()` via viem `erc20Abi` decodes the return as `string`; legacy bytes32-symbol tokens (MKR, SAI) make `readContract` throw and kill the whole portfolio read. The mcp-ts source this ports from slices the symbol manually with a `try/catch → 'UNKNOWN'` fallback so it degrades the symbol only, still returning the balance. Fail-closed (a throw, not a fabricated balance) and same shape as the existing `evmCheckAllowance` sibling, so not blocking — but a behavioral regression vs the source for that token class. Consider a bytes32-aware fallback + isolated test. (Codex lane)

packages/sdk/src/tools/evm/balanceEvm.ts
```diff
@@ -0,0 +1,105 @@
...
+ * // ]
+ * ```
+ */
+export const getEvmBalances = async (chain: EvmChain, params: GetEvmBalancesParams): Promise<EvmBalance[]> => {
+  const client = getEvmClient(chain)
+  const { address, tokens = [] } = params
+
+  const nativeCoin = chainFeeCoin[chain]
+
+  const [nativeRaw, ...tokenResults] = await Promise.all([
```
NeOMakinG, line 65: `suggestion:` all-or-nothing `1 + 3N` fan-out — one token revert/timeout/429 rejects the whole batch and drops even the native balance. For a portfolio view, best-effort per-token settle degrades better; at minimum a docstring note that callers should pass curated token lists. (Codex lane)

packages/sdk/src/tools/evm/balanceEvm.ts
```diff
@@ -0,0 +1,105 @@
...
+ * })
+ * // => [
+ * //   { symbol: 'ETH',  decimals: 18, raw: 1234n, balance: '0.000…' },
+ * //   { symbol: 'USDC', decimals: 6,  raw: 500000n, balance: '0.5', contractAddress: '0xA0b8…' },
+ * // ]
+ * ```
+ */
+export const getEvmBalances = async (chain: EvmChain, params: GetEvmBalancesParams): Promise<EvmBalance[]> => {
+  const client = getEvmClient(chain)
+  const { address, tokens = [] } = params
```
NeOMakinG, line 61: `suggestion:` `address`/`tokens` are only `0x${string}`-typed, not validated/checksummed — a malformed input explodes inside viem with a low-level error. Other EVM entry points run `isAddress(.., {strict:false})` + `getAddress(..)` first. (Codex lane)

NeOMakinG (approved):
fix-and-approve. `sdk.balance.evm` is a clean read-only EVM balance reader and the B-series fabrication lens comes back **all-clear**: native decimals from `chainFeeCoin[chain].decimals` (18 for EVM natives, not hardcoded), ERC-20 decimals read **on-chain per-token** via viem `erc20Abi.decimals()`, all formatting in pure BigInt (`formatUnits`), and fail-closed by construction (no try/catch anywhere — a reverting `decimals()`/`balanceOf` rejects the whole `Promise.all`, never a fabricated 0). I pushed two fixes for the recurring RN-export + changeset gaps; everything else is non-blocking.

## testing approach

Worktree off `origin/feat_sdk_balance_evm` @9c321c80, fixes pushed @a64c43c2. Three lanes: Claude (load-bearing, source walk + mutation), Codex gpt-5.4 (converged, 3 findings), GLM (dead, 0-byte — best-effort skip).

- **decimals correctness**: native via `chainFeeCoin[chain].decimals`; ERC-20 via on-chain `decimals()`. No hardcode. **MUT killed**: hardcoding `formatUnits(raw, 18)` → 2 tests fail (USDC mixed-decimal + native+ERC20). Per-token straightness proven by the mixed-decimal test (USDC 6d + DAI 18d concurrent).
- **fail-closed**: **MUT killed** — swallowing the `decimals()` revert with `.catch(() => 18)` → the fail-closed test fails. The RPC-failure test asserts `.rejects.toThrow('execution reverted')`. No fabricated-zero path exists.
- **decode**: `balanceOf`→bigint, `decimals`→number, `symbol`→string, all via `erc20Abi` (correctly typed). `formatUnits` BigInt logic byte-matches the mcp-ts `formatWei` source — no precision loss.
- **live receipt** (real `api.vultisig.com/eth/`, vitalik.eth): `ETH 5.690864439481747771 (18d)`, `USDC 36.769066 (6d)`, `DAI 4.57207827332311288 (18d)` — mixed-decimal correctness confirmed against the real chain, no 10^12 scaling error.
- **gates @a64c43c2**: vitest 6/6, tsc exit 0 (0 errors, after `yarn workspace @vultisig/core-chain build` clears the mpc-types wall), prettier clean, changeset guard passes. CI green @9c321c80 (our push doesn't re-trigger CI on gomes branches — verified locally).

## blocking

(fixed in-place @a64c43c2 — gomes fix-and-approve)

1. **RN export gap** — `getEvmBalances` was missing from the `src/platforms/react-native/index.ts` EVM allow-list (it sits right next to `evmCall`/`evmCheckAllowance` which ARE exported), so RN consumers like vultiagent-app could not reach the new reader. Fixed via static re-export (RN-safe: same viem pure-fetch shape as its siblings).
2. **Missing changeset** — only `packages/sdk/src/**` is touched, and the bundled-changeset guard only enforces `packages/core`+`lib` prefixes → green-but-unpublishable. Added `.changeset/balance-evm.md` (`@vultisig/sdk` minor).

## should-fix

1. **bytes32-symbol tokens (MKR/SAI) throw instead of degrading** — reading `symbol()` through viem `erc20Abi` decodes the return as `string`; legacy tokens (MKR, SAI, some older ERC-20s) return a `bytes32` symbol, which makes `readContract` throw and kills the entire portfolio read. The mcp-ts source this ports from manually slices the symbol with a `try/catch → 'UNKNOWN'` fallback, so it degrades the symbol only while still returning the balance. This is fail-closed (a throw, not a fabricated balance) and identical to the existing `evmCheckAllowance` sibling, so not blocking — but it's a behavioral regression vs the source for that token class. Consider a bytes32-aware symbol fallback (and an isolated test). *(Codex lane)*

## suggestion

1. **all-or-nothing fan-out** — the top-level `Promise.all` is `1 + 3N` RPCs; one token revert/timeout/429 rejects the whole batch and drops even the native balance. For a portfolio view, best-effort per-token enrichment (settle, surface per-token errors) degrades better. The fail-closed-on-any-error behavior is defensible for a balance primitive, but worth a docstring note that callers should pass curated/known token lists. *(Codex lane)*
2. **no address validation** — `address`/`tokens` are only `0x${string}`-typed, not `isAddress`-validated/checksummed; a malformed input explodes inside viem with a low-level error. Other EVM entry points run `isAddress(..., {strict:false})` + `getAddress(...)` first. *(Codex lane)*

## q

The native coin is always result index 0 and `chainFeeCoin[chain]` is assumed present for every `EvmChain` — confirmed it is (the type is keyed over all chains). No action.

## risk

Low. Read-only, vault-free, no signing/broadcast. Worst case for an unhandled token class (bytes32 symbol) is a thrown error (fail-closed), never a wrong/fabricated balance. BigInt throughout — no precision loss verified live against a >MAX_SAFE-style 18-decimal value.

## verdict

**APPROVE.** Decimals correct (native 18 from registry, ERC-20 per-token on-chain, no hardcode), fail-closed (no fabricated zero, both fabrication mutations killed), correct decode, RN export + changeset added, tests prove decimals + fail-closed, gates green. should-fix (bytes32 symbol) + suggestions are non-blocking follow-ups.

---
🤖 reviewed by Claude (load-bearing) + Codex gpt-5.4 (converged) · GLM lane dead (best-effort skip)

PR #383, opened 2026-05-04
chore(thorchain): use Liquify gateway for Midgard/THORNode/RPC

Ehsan-saradar:
## Summary
- Migrate THORChain Midgard, THORNode REST, and Tendermint RPC endpoints from `*.thorchain.network` to the Liquify gateway: `gateway.liquify.com/chain/thorchain_midgard`, `…/thorchain_api`, `…/thorchain_rpc`. Touches `cosmosRpcUrl.THORChain`, `tendermintRpcUrl.THORChain`, `thorchainMidgardBaseUrl`, and rujira `MAINNET_CONFIG`.
- Replace the brittle `rpc → thornode` string substitution in `RujiraDiscovery.discoverViaChain()` with a direct read of `MAINNET_CONFIG.restEndpoint`. Under the new gateway routing the old transform produced an invalid `thorchain_thornode` host and made the fallback branch unreachable.
- Drop the now-unused `rpcEndpoint` option from `DiscoveryOptions` and the corresponding plumbing in `RujiraClient`.

## Test plan
- [ ] `GET gateway.liquify.com/chain/thorchain_midgard/v2/pools` returns the expected pool list (wired via `getThorchainPools`).
- [ ] `GET gateway.liquify.com/chain/thorchain_api/cosmwasm/wasm/v1/code/{finCodeId}/contracts` paginates correctly through `RujiraDiscovery.discoverViaChain()`.
- [ ] Tendermint RPC operations against `gateway.liquify.com/chain/thorchain_rpc` succeed (status, abci_info, broadcast).
- [ ] `RujiraClient` connects on mainnet with the default config (no custom `rpcEndpoint` override required).
- [ ] Existing balance / send / swap flows on THORChain still work end-to-end through the SDK.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Migrated all THORChain service endpoints to the Liquify gateway infrastructure, including RPC, API, and Midgard services for improved reliability.
  * Simplified endpoint discovery configuration by removing the unused legacy RPC endpoint option and relying on centralized gateway configuration.

PR #815, opened 2026-06-23
feat(sdk): add sdk.gas.utxoFeeRate (UTXO sat/vB)

gomesalexandre:
## what

New pure-crypto SDK primitive `utxoFeeRate(chain: UtxoChain): Promise<{ chain, feeRate, feeRateUnit: 'sat/vB' }>` exported from `@vultisig/sdk`. Ported from mcp-ts `src/tools/fee/utxo-fees.ts` (0 SDK imports there), source of truth is the THORChain / MayaChain `inbound_addresses` endpoint — the same recommended `gas_rate` the swap rail uses when building UTXO KeysignPayloads.

Covers all six UTXO chains: Bitcoin / Litecoin / Dogecoin / Bitcoin-Cash via THORChain, Dash / Zcash via MayaChain (not on THORChain).

## why

Read-only fee-rate math that currently lives in mcp-ts as a gaggle of near-identical `btc/ltc/doge/bch/dash/maya_fee_rate` tools. Folding it into the SDK is the `BRING-INTO-SDK` row for `sdk.gas.utxoFeeRate(chain)` — so any consumer that builds a UTXO send / consolidation gets a fee rate consistent with the network the swapper would broadcast against, without re-implementing the inbound fetch + halt gating.

Keeps the safety gating from the port (NeOMakinG #113): refuses to return a zero-fee envelope. Throws on a halted chain or a non-positive `gas_rate` instead of emitting a tx that would never confirm on-chain. The live receipt below shows this in action — LTC was halted on the inbound source and surfaced honestly rather than silently returning 0.

## consumers it unblocks

- mcp-ts `btc/ltc/doge/bch/dash/maya_fee_rate` tools → thin wrappers over `sdk.gas.utxoFeeRate`
- `build_utxo_consolidate` / utxo-split prep (shares the same inbound fee source)
- any agent-backend / app caller building a UTXO send that needs a current sat/vB rate

## test matrix

- typecheck: `yarn workspace @vultisig/sdk typecheck` → 0 errors (0 new from this PR)
- unit: 6/6 pass (`tests/unit/utxoFeeRate.test.ts`) — healthy rate, THOR vs Maya routing, halt guard, non-positive guard, missing-entry guard (fetch mocked, deterministic)
- receipt: live call against the real shipped source path

## RECEIPT

```
$ node --import tsx scripts/receipts/gas_utxo_feerate.mjs

sdk.gas.utxoFeeRate — live UTXO fee rates (sat/vB)

  Bitcoin           4 sat/vB
  Litecoin        ERR  chain LTC is currently halted on the inbound source — cannot compute fee rate
  Dogecoin      750000 sat/vB
  Bitcoin-Cash      3 sat/vB
  Dash             12 sat/vB
  Zcash         127500 sat/vB

Receipt OK
```

```
$ yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/utxoFeeRate.test.ts

 ✓ tests/unit/utxoFeeRate.test.ts (6 tests) 
 Test Files  1 passed (1)
      Tests  6 passed (6)
```


gomesalexandre (commented):
Deep adversarial pass on the UTXO sat/vB port (own PR, comment-only — not approving my own).

## found + fixed (HIGH, fund-safety): zcash wrong-rate

\`utxoFeeRate.ts:40\` mapped \`Zcash → MAYACHAIN ZEC\` and returned the published \`gas_rate\` under \`feeRateUnit: 'sat/vB'\`. **Zcash does not use a sat/vB fee model** — it uses ZIP-317 conventional fees (zats per logical action). MayaChain's ZEC \`gas_rate\` is a ZIP-317 value, not a per-vByte rate.

Live proof right now:
\`\`\`
$ curl -s https://mayanode.mayachain.info/mayachain/inbound_addresses | jq '.[]|select(.chain=="ZEC")'
  ZEC gas_rate=127500 halted=false
\`\`\`
So pre-fix \`utxoFeeRate('Zcash')\` → \`{ feeRate: 127500, feeRateUnit: 'sat/vB' }\`. A consumer building a Zcash send/consolidation at 127500 sat/vB burns the whole balance in fees. This is exactly the "wrong-rate per chain" class the review focused on.

Cross-checked the canonical sources, all of which special-case Zcash OUT of the sat/vB path:
- \`core-chain/chains/utxo/fee/byteFee.ts\` → \`getUtxoByteFee\` hardcodes Zcash to \`100n\` sat/byte to clear the ZIP-317 floor; everyone else goes through Blockchair.
- swap rail \`execute/_base.ts\` + \`execute_swap.ts\` \`resolveUtxoFeeRate\` — Zcash omitted from \`UTXO_FEE_FALLBACK\` with the explicit comment "a sat/vB fallback would be silently misleading — the app's Zcash signing path ignores fee_rate and uses ZIP-317."
- mcp-ts ships no \`zec_fee_rate\` tool for the same reason.

**Fix (`68d1c881`):** dropped Zcash from the source map → \`utxoFeeRate('Zcash')\` now throws \`Unsupported UTXO chain for sat/vB fee rate: Zcash (Zcash uses ZIP-317 ... use the ZIP-317 fee path)\` and does NOT hit the network. Added a regression test asserting the rejection + no fetch.

## verified clean (no bug)

- **u64/JSON-number precision** (the sibling balance/amount bug): N/A here — \`gas_rate\` is a small sat/vB integer parsed via \`parseInt(..., 10)\`, never near 2^53. DOGE's \`750000\` and the like are well within safe-integer range and consumed identically by iOS/Android/swap rail. No raw-bigint path needed.
- **halt handling**: faithful to mcp-ts — throws \`HaltedChainError\`-equivalent on \`halted: true\` and on non-positive \`gas_rate\`, never yields a zero-fee envelope. Verified against LIVE data (LTC was halted on thornode during the receipt run → honest throw, not a mock).
- **fail-open**: the port is fail-CLOSED (throws on source-down/missing entry) vs the swap rail's per-chain fallback. Correct for a standalone primitive — better to error than emit a stale static rate.
- **per-chain endpoint routing**: BTC/LTC/DOGE/BCH→thornode, DASH→mayanode (not on THORChain). All correct, chainKeys (BTC/LTC/DOGE/BCH/DASH) match the inbound \`chain\` field.
- **address validation**: N/A (no address handling in this primitive).

## receipt (live, real SDK source via `--import tsx`)
\`\`\`
Bitcoin          19 sat/vB
Litecoin        ERR  chain LTC is currently halted on the inbound source — cannot compute fee rate
Dogecoin      750000 sat/vB
Bitcoin-Cash      3 sat/vB
Dash             12 sat/vB
Zcash           ERR  Unsupported UTXO chain for sat/vB fee rate: Zcash (Zcash uses ZIP-317 ... use the ZIP-317 fee path)
\`\`\`

typecheck: 0 errors. unit: \`utxoFeeRate.test.ts\` 7/7 green (incl. new Zcash rejection). Full sdk suite: 1515/1517 — the 2 reds are pre-existing \`ensureMpcEngine\`/RN-entry WASM-mock flakes, untouched by this diff.

NeOMakinG (approved):
Fresh review (round 1) — fund-safety lens on a read-only UTXO fee-rate fetcher, since the output feeds tx-builders (#823 utxoConsolidate `byteFee`). Reviewed at HEAD `68d1c881`; pushed fixes land at `fcb0ecde`.

This is a clean, well-gated primitive. The Zcash zip-317 exclusion (commit `68d1c881`) is exactly right — Maya's published ZEC `gas_rate` (127500) is a ZIP-317 zats-per-action value, not sat/vB, and returning it under a `sat/vB` envelope would have built a balance-burning tx. Good catch baked into the source map + a network-free reject test. Fail-closed gating (halt / non-positive `gas_rate` → throw, never a zero-fee envelope) is sound and the live receipt proves it (LTC halted → honest ERR, not silent 0).

## testing approach

- Read every path in `utxoFeeRate.ts`, the test, the receipt, and the index/RN wiring.
- **Unit**: 7/7 → 8/8 after my strict-parse test. Mutation-checked: flipping the unit label (`sat/vB`→`sat/byte`), dropping the halt guard, loosening `rate<=0`→`rate<0`, and reverting strict-parse→`parseInt` each fail a test. Real teeth (#228).
- **Live receipt** through the canonical host: BTC 3, LTC halted (ERR), DOGE 750000, BCH 3, DASH 12, Zcash rejected.
- **Source verification**: confirmed THORChain `inbound_addresses` publishes `gas_rate_units: "satsperbyte"` for all four; confirmed core-chain migrated THORChain REST to the Liquify gateway (`cosmosRpcUrl[Chain.THORChain]`) and live-probed that the gateway path returns identical data.
- Gates (local @ `fcb0ecde`): vitest 8/8, `tsc` 0 errors, prettier clean, complexity (sonarjs) clean, lint clean, shared-exports OK, changeset guard passes. CI was green @ `68d1c881`; my fix-commit does not re-trigger CI on this branch — gates re-verified locally.
- Lanes: Claude (load-bearing) + Codex gpt-5.4 (converged, sourced — drove 3 of the in-code fixes) + GLM (dead, best-effort skip).

## blocking

None.

## should-fix

**Fee source diverges from the SDK's own UTXO builders (`getUtxoByteFee`).** This primitive returns THORChain/MayaChain's raw `inbound_addresses` `gas_rate` — the rate THORChain's *own vaults* broadcast at, which is deliberately conservative/inflated (live DOGE = 750000 sat/byte). But the codebase's actual UTXO send/swap `byteFee` comes from `packages/core/chain/chains/utxo/fee/byteFee.ts:getUtxoByteFee`, which uses Blockchair's `suggested_transaction_fee_per_byte_sat` with a ×2.5 multiplier and **DOGE ÷4** (and Zcash=100). THORChain `gas_rate` is consumed in core-chain only for swap dust/minimum economics (`getNativeSwapMinAmountIn`), **not** to set the broadcast `byteFee`. So the PR description's "the same recommended `gas_rate` the swap rail uses when building UTXO KeysignPayloads" overstates the relationship — a consumer wiring `utxoFeeRate('Dogecoin')` → 750000 into #823's consolidation would pay an enormous fee versus core-chain's own DOGE handling. This is fail-safe (over-pay, not a stuck tx) and 750000 is genuinely what THORChain recommends, so it's not a blocker — but please either (a) document on `UtxoFeeRate` that the rate is the THORChain/Maya inbound recommendation (conservative, may exceed mempool/Blockchair-derived rates, esp. DOGE), and/or (b) reconcile with `getUtxoByteFee` before #823 wires it as a `byteFee` source.

## suggestion

**TTL / single-flight cache per source URL** (Codex). BTC/LTC/DOGE/BCH all hit the same THORChain endpoint; repeated agent/tool calls fan out identical public-node requests for a slowly-changing value. `getNativeSwapMinAmountIn` already does this with a 5s cache — consider mirroring it.

## fixed in-code (pushed to `fcb0ecde`)

Since this is a fix-and-approve, I pushed the following rather than blocking:

1. **RN export gap** — `utxoFeeRate` / `UtxoFeeRate` / `THORCHAIN_NODE_URL` / `MAYACHAIN_NODE_URL` were absent from `src/platforms/react-native/index.ts`, so RN consumers (vultiagent-app) resolving via the react-native export condition couldn't reach the new primitive. Added a static re-export (pure `fetch` + type-only import = RN-safe). This recurs across the SDK port series.
2. **Missing changeset** — the bundled-changeset CI guard (`scripts/check-sdk-bundled-changeset.mjs`) only inspects `packages/core/*` + `packages/lib/utils/`, so a `packages/sdk/src`-only PR passes the guard green yet ships **unpublishable** (no `@vultisig/sdk` bump). Added a `minor` changeset.
3. **Legacy THORChain host** (Codex, verified) — `THORCHAIN_NODE_URL` hardcoded `thornode.thorchain.network`, but core-chain migrated THORChain REST to the Liquify gateway (`cosmosRpcUrl[Chain.THORChain]`, PR #383). Re-sourced both node URLs from the canonical `cosmosRpcUrl` constant so this primitive tracks host migrations instead of pinning a host that can be decommissioned. Live-verified the gateway path returns identical `inbound_addresses` data. (MayaChain already matched the canonical constant.)
4. **`AbortSignal.timeout()` RN trap** (Codex, verified) — not available on older RN/Hermes runtimes; the repo's own `src/chains/tron/rpc.ts` uses `AbortController + setTimeout` for exactly this reason. Matters now that the primitive is RN-exported. Switched to the Hermes-compatible pattern.
5. **`parseInt` permissive parse** (Codex, verified) — `parseInt('10.5',10)===10`, `parseInt('1e3',10)===1`, `parseInt('15px')===15` would silently emit a wrong fee rate. Replaced with a strict `^\d+$` integer parse that fails closed on malformed responses, plus a test.

## risk

Low. Read-only informational fetcher. Fail-closed by construction (halt / non-positive / malformed → throw; never a fabricated 0). Unit is genuine sat/vB for the four legacy-format chains (THORChain `satsperbyte` == sat/vB where vsize==size; for segwit BTC it's a conservative over-estimate, standard practice). The one substantive caveat — over-conservative DOGE rate diverging from `getUtxoByteFee` — is over-pay, not stuck-tx, and flagged above for #823 wiring.

## verdict

APPROVE. Correct unit, sane fail-closed floor (never 0), Zcash zip-317 exclusion correct, strong mutation-tested coverage. RN export + changeset + canonical host + Hermes-safe timeout + strict parse fixed in-code at `fcb0ecde`; all gates green locally. The remaining should-fix (DOGE/source divergence) is a documentation/reconciliation item for the #823 wiring, not a merge blocker.

---
🤖 reviewed by Claude (load-bearing) + Codex gpt-5.4 (converged) · GLM lane dead (best-effort)

PR #816, opened 2026-06-23
feat(sdk): add sdk.balance.solana (native SOL + SPL)

gomesalexandre:
## what

Adds two pure-crypto Solana balance reads to the SDK public surface:

- `getSolBalance(address)` -> native SOL balance (lamports + trimmed human SOL string)
- `getSplTokenBalance(address, mint)` -> SPL/Token-2022 token balance (auto-detects token program, derives the ATA, returns raw u64 amount + decimals)

Ported 1:1 from mcp-ts `src/tools/balance/solana-balance.ts` (0 SDK imports there), stripped of all agent/LLM concepts (no ToolDef/handler/inputSchema, no vault-derivation fallbacks). Pure RPC reads against the existing `solanaRpcUrl` proxy (`https://api.vultisig.com/solana/`) via raw JSON-RPC, so we don't drag in the heavy `@solana/web3.js` Connection for a read.

## why

Part of the mcp-ts/backend -> SDK code-as-action consolidation. The balance reads currently live in mcp-ts as ToolDefs; the SDK is the canonical home for chain crypto. This unblocks mcp-ts `get_sol_balance` / `get_spl_token_balance` (and Station / vultisig-windows) consuming the SDK instead of carrying their own RPC + lamports math.

## how

- `packages/sdk/src/tools/balance/solana.ts` - the two reads + typed results (`SolBalance`, `SplTokenBalance`)
- exported from `tools/index.ts` and the package root `src/index.ts`
- unit test mocks `fetch` for deterministic coverage: lamports->SOL conversion, zero-balance `"0"` (no dangling decimal), full u64 precision, HTTP + JSON-RPC error paths, SPL parse, and empty-account zero case
- `publicExports.test.ts` asserts both are on the public surface (matches the existing mcp-ts-consumer pattern)
- runnable receipt hits LIVE mainnet RPC

## risk

Low - additive, read-only, behind no flag (it's a library export). 0 new typecheck errors over the mpc-wasm baseline. 1518 unit tests green. Lint clean.

## receipt

LIVE mainnet RPC (the curl-equivalent) - real SOL + real USDC SPL balance for a known address:

\`\`\`
$ npx tsx scripts/receipts/balance_solana.mjs
native SOL : {"address":"GThUX1Atko4tqhN2NaiTazWSeFWMuiUvfFnyJyUghFMJ","lamports":326045132452,"sol":"326.045132452","asOf":"2026-06-23T14:31:34.191Z"}
SPL  USDC  : {"address":"GThUX1Atko4tqhN2NaiTazWSeFWMuiUvfFnyJyUghFMJ","mint":"EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v","ata":"6u6tm3d9Vf4QUDdbtMaV21qsmPHorJebdyDT6ZJ9h5JY","tokenProgram":"spl-token","balance":"4230165","decimals":6,"asOf":"2026-06-23T14:31:34.518Z"}
\`\`\`

USDC came back with decimals=6, spl-token program, and the correct derived ATA -> 4.230165 USDC. Native SOL trimmed cleanly.

\`\`\`
$ yarn workspace @vultisig/sdk test:unit tests/unit/balance-solana.test.ts tests/unit/utils/publicExports.test.ts
Test Files  111 passed (111)
     Tests  1518 passed (1518)
\`\`\`


gomesalexandre (commented):
## adversarial pass — 2 real bugs found + fixed (pushed 4fea6afa)

Did a deep adversarial diff of `src/tools/balance/solana.ts` against the Go canonical source (`vultisig-mcp/internal/tools/get_{sol,spl_token}_balance.go` + `internal/solana/client.go`) and the mcp-ts port it came from. FOCUS as briefed: lamports→SOL, SPL balanceOf, decimals, ATA, precision/overflow. The port was 1:1 from mcp-ts and faithfully re-shipped two mcp-ts precision bugs that the Go source does NOT have. Both fixed.

### 1) native SOL — u64 precision lost (P2 correctness) — `solana.ts:99-122`
`getBalance.value` is a JSON **number** (u64). Two compounding issues:
- `JSON.parse` silently rounds it once it crosses 2^53 (~9.007M SOL). Verified: raw `12345678901234567` → `...568` straight out of `JSON.parse`.
- `sol` was derived via **float** division `lamports / 1e9` + `.toFixed(9)` — fragile vs the Go source, which uses integer-only `whole/frac` math (`FormatLamports`, `client.go:43`).

**Fix:** recover the exact integer from the raw response **text** (before any Number coercion), expose a lossless `lamportsRaw` string, and format `sol` via BigInt integer division (mirrors `FormatLamports` exactly: `12345678901234567` → `12345678.901234567`). `lamports: number` kept for back-compat (documented lossy >2^53; prefer `lamportsRaw`). `sol` is now exact across the full u64 range. Added a cross-check that the regex value matches the parsed body in the safe range + a malformed-response guard.

### 2) SPL — `accounts[0]` undercounts + non-deterministic (P2 fund-display) — `solana.ts:157+`
`getTokenAccountsByOwner` returns **all** of the owner's token accounts for the mint, in **unspecified order**. Taking `accounts[0]`:
- **undercounts** when the owner holds the mint across the canonical ATA + auxiliary accounts, and
- yields a **non-deterministic** representative (RPC ordering), mislabeled `ata`.

The Go source sidesteps this entirely — it derives the canonical ATA and reads that one account. The mcp-ts/SDK `getTokenAccountsByOwner` shape can't do that cheaply, so the correct behavior is to **sum** all matching accounts.

**Fix:** sum every account (lossless BigInt), surface the largest-balance account as the representative `ata`, and defensively drop any account whose parsed `mint` != requested mint (never trust a balance from the wrong mint even though the RPC filters).

**Live mainnet proof (the PR's OWN receipt address):** that address holds USDC across **4** token accounts. The old `accounts[0]` reported **4.230165 USDC**; the true sum is **15,564,480.100016 USDC** — an undercount of ~99.99997%. Not theoretical:
```
FS3HipLdf13... 2501911823544
CHynyGLd4f... 13059689732254   <- now the representative ata
6u6tm3d9Vf...       4230165   <- what accounts[0] used to pick
6pJuA19E33...    2874314053
SUM:          15564480100016
```

### receipts
```
$ npx tsx scripts/receipts/balance_solana.mjs
native SOL : {"lamports":326045132452,"lamportsRaw":"326045132452","sol":"326.045132452",...}
SPL  USDC  : {"ata":"CHynyGLd4fDo35VP4yftAZ9724Gt49uXYXuqmdWtt68F","balance":"15564480100016","decimals":6,...}

$ yarn workspace @vultisig/sdk test:unit balance-solana publicExports
Test Files  111 passed (111) / Tests  1522 passed (1522)   # +4 new (u64 precision, malformed, multi-account sum, mint-mismatch)

$ yarn workspace @vultisig/sdk typecheck  -> 0 errors
```

- decimals: read straight from `tokenAmount.decimals` per-account, never hardcoded — correct.
- token-program auto-detect (spl-token vs token-2022): correct via `account.data.program`.
- empty-account zero case + trimmed `'0'` (no dangling decimal): correct.
- HTTP + JSON-RPC error paths: surfaced, fail-closed (throw, no fail-open default).
- RPC URL: `@vultisig/core-chain` `solanaRpcUrl` proxy — not a phantom/undeclared dep.

Note (out of scope, NOT changed): `getSolBalance`/`getSplTokenBalance` don't validate the base58 `address`/`mint` before the RPC call (the Go source does, via `ParsePublicKey`). For a read-only balance that just yields an RPC error on garbage input, so low-sev — flagging for a future hardening pass, didn't want to scope-creep the address-validation surface into this PR.

NeOMakinG (approved):
## testing approach

Three-lane review (Claude load-bearing + Codex gpt-5.4 + GLM 4.6). GLM dead (429 quota). Codex retry returned a 228KB transcript that was mostly file dumps with no crisp findings — best-effort, treated as inconclusive this run; Claude lane is load-bearing and independently covered all four adversarial scope points.

Worktree `origin/feat_sdk_balance_solana` @ `4fea6afa`, isolated build. Applied B-series balance-fabrication lens (sibling #843 had a fabricated-zero blocker).

**Gates (local, post-fix @ `ec4f5d28`):**
- `balance-solana.test.ts` + `publicExports.test.ts` + RN entry test: **1522 pass**
- `tsc --noEmit`: **0 errors**
- prettier (`.config/.prettierrc`): clean · eslint: clean · complexity (`eslint-health`): exit 0
- `check-sdk-bundled-changeset.mjs`: exit 0 · `check:shared-exports`: exit 0
- **RN platform bundle build: exit 0** — `vultisig-rn-path-override` fires `chains/solana/client.ts → solanaClient.ts`; `getSolBalance`/`getSplTokenBalance` present in `dist/index.react-native.js`; `@solana/web3.js` only via the override's lazy `import()` (no eager Hermes-hostile drag)
- **Live mainnet receipt:** native SOL `326.045132452` (exact u64 lamports) + USDC `decimals=6`, `spl-token`, derived ATA — both reads green
- CI on PR: CodeRabbit / Unit / Integration / Lint+Type all **pass** (our pushed fix-commit does not re-trigger CI on gomes branches — gates verified locally)

**Mutation check (tests have teeth):** divisor 1e9→1e8 ❌ killed · empty-account `'0'`→`'999'` (fail-open sim) ❌ killed · loop-body decimals→9 ❌ killed · mint-mismatch guard removed ❌ killed. (Initializer `let decimals = accounts[0]…` is dead for any non-empty matching set — the loop body authoritatively reassigns; the per-mint-decimals invariant is covered by MUT3b.)

## decimals + fail-closed (B-series lens)

- **Native SOL = 9**: `formatLamports` uses integer/BigInt whole/frac split (no float). The u64 regex-recovery on the raw text (line 108) sidesteps `JSON.parse` rounding above 2^53; cross-validated against `json.result.value` in the safe-int range (117-120) → fail-closed on ambiguous shape. Exact across the full u64 range (test @ `12345678901234567`).
- **SPL per-mint decimals**: read per matching account from `tokenAmount.decimals`, never hardcoded.
- **Fail-closed, no fabricated zero**: HTTP !ok → throw; JSON-RPC `error` → throw; malformed getBalance (no value) → throw. Empty token-account set returns a **legit** `'0'` (RPC-confirmed no account) — matches canonical, not a fabrication.
- **Multi-account sum** (the `4fea6afa` review fix) is a strict improvement over the mcp-ts canonical `accounts[0]`: lossless BigInt sum across all of the owner's accounts for the mint + a defensive mint-mismatch guard.

## fixed in-code (pushed `ec4f5d28`, gomes fix-and-approve)

1. **RN export gap** — `getSolBalance`/`getSplTokenBalance` were only on the node barrel. The RN platform index hand-picks exports (no `export *`), so React Native consumers (Station / vultiagent-app — the PR's stated unblock target) could not reach them. Re-exported from the RN entry; verified safe + reachable via the RN bundle build above.
2. **Missing changeset** — `check-sdk-bundled-changeset.mjs` only inspects `packages/core` + `lib/utils`, so a `packages/sdk/src/**` public-surface addition passes the gate green-but-**unpublishable**. Added a `@vultisig/sdk: minor` changeset.

## should-fix

- **base58 address validation absent** — the mcp-ts canonical validates owner+mint via `solanaAddressString()` (base58 decode + 32-byte length) before any RPC. The port drops it. Not blocking (an invalid address → RPC error → throw, so fail-closed is preserved), but a fast structural reject would be friendlier and restores parity. Left as a follow-up rather than fixed in-place to keep the port 1:1 on behavior.

## suggestion

- **Representative-ATA tie-break** is non-deterministic across equal-balance accounts (`amount > repAmount` strict keeps the first-seen max under unspecified RPC order). `balance`/`decimals`/`tokenProgram` are all deterministic; only the documentary `ata` field can vary. Consider a stable secondary sort (pubkey) if `ata` is ever load-bearing downstream.

## q

- **Consumer migration note** for mcp-ts `get_spl_token_balance`: it currently surfaces `accounts[0]`'s `ata`/`token_program`; the SDK sums all accounts and surfaces the largest as representative. `balance` becomes the true total (correct), but `ata`/`token_program` may differ from the old single-account value for multi-account owners. Worth a line in the migration PR.

## risk

Low. Additive, read-only, no flag, fail-closed throughout. No fabricated-zero path. RN export + changeset now in place and build-verified.

## verdict

**APPROVE.** Decimals correct (SOL=9 integer math, SPL per-mint, BigInt throughout), fail-closed with no fabricated zero, correct SPL program auto-detect + jsonParsed parse, multi-account sum is a net improvement over canonical, tests have real mutation teeth. The two recurring gomes gaps (RN export + changeset) are fixed in-code on the branch (`ec4f5d28`) and verified via the RN bundle build. base58 validation parity left as a non-blocking follow-up.

---
*Reviewed by Claude (load-bearing) + Codex gpt-5.4 (best-effort, inconclusive) + GLM 4.6 (dead/quota). Fixes pushed to the PR branch.*

PR #817, opened 2026-06-23
feat(sdk): add sdk.price.get + sdk.price.batch (token USD price)

gomesalexandre:
## what

Adds `getPrice` + `getPricesBatch` to the SDK public API — token USD pricing via CoinGecko (through the Vultisig proxy), ported from the mcp-ts `get_price` price-oracle into a pure-crypto primitive.

`getPrice(query)` resolves a USD price (+ 24h change + market cap) across four read-only routes:
- explicit CoinGecko coin id
- EVM contract + chain
- Solana mint
- native ticker (via the ported `NATIVE_COINGECKO_IDS` map)

`getPricesBatch(queries)` fans out in parallel and isolates per-query failures (`{ ok: false, error }`) so one unpriceable token can't sink the whole batch, with results in input order.

It never returns a fabricated price — a lookup failure throws, so callers never build a tx off a guessed amount.

## why

Part of the mcp-ts/backend → SDK code-as-action consolidation. The price fetch lived in mcp-ts; the internal `FiatValueService` already exists but wasn't exposed as a standalone primitive. This lands the read-side oracle in the SDK so every surface (mcp-ts, agent-backend, CLI, app card USD valuations) shares one canonical pricer instead of re-implementing CoinGecko routing.

Agent-layer concerns (token-selection ambiguity prompts, Terra-Classic fiat-denom graceful errors, `_classic` suffix sanitisation) intentionally stay in orchestration — this PR is the pure-crypto core only.

## consumers this unblocks

- mcp-ts `get_price` / `get_prices_batch` → wrap `sdk.getPrice` / `sdk.getPricesBatch` instead of shadow CoinGecko code
- agent-backend USD valuation in `get_holdings` portfolio aggregation
- app-side fee/amount USD labels (the swap-card `feeUsd` line) without an mcp round-trip

## tests

`packages/sdk/tests/unit/tools/price/getPrice.test.ts` — 14 cases: all four routes, metadata-skip fast path, batch ordering, graceful partial failure, and every throw path (no chain, unsupported chain, unknown ticker, no identity, missing usd field). Mocks `queryUrl` for determinism.

## receipt (live proxy — curl-equivalent)

```
$ npx tsx scripts/receipts/price_get.mjs

=== sdk.price.get (single, native ticker) ===
ETH   $1,660.71  (24h -5.38%)  mcap $200.07B

=== sdk.price.get (single, EVM contract: USDC on Ethereum) ===
USDC  $1  (24h +0.01%)  mcap $74.40B  [Ethereum 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48]

=== sdk.price.batch (ETH + BTC + USDC) ===
ETH   $1,660.71  (24h -5.38%)  mcap $200.07B
BTC   $62,507  (24h -3.76%)  mcap $1251.77B
USDC  $1  (24h +0.01%)  mcap $74.40B

=== sdk.price.batch (graceful partial failure: BTC ok, BOGUS fails) ===
OK   BTC   $62,507  (24h -3.76%)  mcap $1251.77B
FAIL {"symbol":"NOTACOIN"} -> price lookup for "NOTACOIN" requires a resolved CoinGecko id or contract address

OK: live prices resolved.
```

## verification

- `yarn workspace @vultisig/sdk typecheck` → exit 0, 0 errors (0 new on my files)
- `yarn workspace @vultisig/sdk vitest run tests/unit/tools/price/getPrice.test.ts` → 14/14 pass
- live receipt → real ETH/BTC/USDC prices resolved
- lint clean on all touched files

## risk

Low — additive read-only primitive under a new `tools/price` module, exported from the package index. No existing code paths touched beyond the two export barrels. Hits the same CoinGecko proxy `searchToken` already uses.


gomesalexandre (commented):
Deep adversarial pass on the price primitive (own PR, comment-only).

## focus areas checked

- **USDC / by-contract / by-mint / native-ticker routing** — verified the four routes are faithfully ported from mcp-ts `price-oracle.ts` and that route ordering is exclusive: an EVM `0x…` w/ `chain:Solana` queries `token_price/solana` (CG empty → throws, fail-safe); a base58 mint w/ an EVM chain matches no route → final throw. No wrong-route leak.
- **address validation (zero-padded / wrong-chain — the sibling balance-port bug class)** — both regexes are anchored (`^…$`): a 41-hex addr, trailing junk, or a leading `0` in a base58 mint all reject. No injection / overflow / wrong-chain slip. URL interpolation is bounded to regex-gated contracts + fixed-map cgIds, and Route-0 reads the response back by the exact requested key, so a mismatch fails safe.
- **u64 / decimals-on-fiat-conversion** — N/A here by design: this primitive is price-only, returns `usd` as a raw JSON number and does ZERO `amount * price / 10^decimals` math. The u64-precision / decimal class that bit the balance/amount ports does not exist in this surface (it lives in consumers). `usd` is `typeof === number` guarded; `usd_24h_change` / `usd_market_cap` are `?? 0`. Good.
- **batch partial-failure** — `getPricesBatch` per-query try/catch + `Promise.all` preserves input order and isolates failures as `{ ok:false }`. One unpriceable token cannot sink the batch. Confirmed live (BTC ok + BOGUS fail).

## found + fixed (1, pushed to branch)

**Route 1 (EVM) + Route 2 (Solana) — wrong-token price acceptance** `src/tools/price/getPrice.ts:158,181` (MAJOR, fund-safety-adjacent). Both routes took `Object.values(data)[0]` blindly. A proxy/upstream quirk returning a DIFFERENT contract's entry (or merging an unrelated address) would hand back a wrong-token price under a correct-looking `resolvedSymbol`/`contractAddress` label — directly violating the PR's own "never returns a wrong-token price" guarantee, and the kind of mismatch a consumer USD-valuation can't catch. Fixed: match the response entry to the requested contract key (case-insensitive) and throw on mismatch (fail-safe), via a new `priceForRequestedContract` helper. mcp-ts has the same latent pattern; worth a follow-up there too.

+3 adversarial tests (wrong-key EVM rejection, checksum-cased match, wrong-mint Solana rejection).

## receipts

- `vitest tests/unit/tools/price/getPrice.test.ts` → **17/17 pass** (was 14, +3 adversarial)
- `yarn workspace @vultisig/sdk typecheck` → **exit 0, 0 errors**
- live proxy receipt → ETH / BTC / USDC-by-contract / batch / partial-failure all green AFTER the key-match hardening (USDC-by-contract still resolves: CG keys by lowercased contract → matches):

```
USDC  $1  (24h +0.00%)  mcap $74.22B  [Ethereum 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48]
FAIL {"symbol":"NOTACOIN"} -> requires a resolved CoinGecko id or contract address
```

Fix pushed in `4d483c9c`. clean otherwise — no fail-open, no phantom deps, no dropped guards vs source.

NeOMakinG (approved):
## testing approach

3-lane fan-out (Claude load-bearing; Codex 0-byte stall — session contention, best-effort; GLM exit-1 dead, best-effort). Read every route in `getPrice.ts` + the full `coinGeckoIds.ts` map; cross-checked the id map against the canonical mcp-ts `coin-gecko-id-map.ts` (byte-for-byte parity on POL/MATIC). Live-probed every spot-checked slug through the actual Vultisig proxy (`/coingeicko/api/v3/simple/price`). Ran the unit suite (19/19 after my +5), tsc (0 new errors on touched files; remaining are the pre-existing mpc-types/mpc-wasm dist wall), eslint+sonarjs (clean), prettier (clean), and the SDK bundled-changeset gate.

This is a gomes PR under fix-and-approve: I fixed three issues in-code on the branch (commit `00227fc1`) rather than blocking.

## fixed in-code (commit 00227fc1)

**1. (was blocking) POL/MATIC mapped to a dead CoinGecko slug.** `coinGeckoIds.ts` pinned both `POL` and `MATIC` to `matic-network` (ported verbatim from mcp-ts). Live proxy probe: `matic-network` now returns an empty `{}` body with **no `usd` field** — so the `typeof priceData.usd !== 'number'` guard makes *every* POL/MATIC lookup throw. The live slug after the MATIC->POL rebrand is `polygon-ecosystem-token` (probe returns a real price). This is the exact hazard from vultiagent-app#1456. It fails *closed* (no fabricated price — good), but a high-volume token being structurally unpriceable in the new canonical pricer would silently break the app `feeUsd` line and `get_holdings` valuation. Fixed both tickers -> `polygon-ecosystem-token`; added 2 tests (map assertion + Route-3 fetch) that fail on the dead slug (mutation-verified). Recommend porting the same fix back to mcp-ts `coin-gecko-id-map.ts`.

**2. (was should-fix) RN export gap.** The React Native entry re-exports `tools/prep`, `tools/token`, `tools/address`, `getCoinBalance` — but NOT the new `tools/price`. `getPrice`/`getPricesBatch` were unreachable from RN, i.e. the app-side `feeUsd` consumer the PR description names as a target couldn't import them. Added an RN-safe re-export (price is pure proxy `fetch` over the same endpoint `searchToken` already uses — no WASM/native deps, so a direct `export` is correct; no lazy wrapper needed).

**3. (was should-fix) missing changeset.** `sdk/src/**` is outside the prefixes the `check-sdk-bundled-changeset.mjs` gate watches (`packages/core/*`, `packages/lib/utils/`), so CI stays green — but the surface is still published in the SDK tarball, so the PR was green-but-unpublishable. Added a `@vultisig/sdk: minor` changeset.

## praise

The `priceForRequestedContract` guard is the right call — it refuses to hand back `Object.values(data)[0]` and instead key-matches the requested contract case-insensitively, so a proxy/upstream quirk returning a different contract's entry throws instead of mislabeling a wrong-token price. Both the EVM and Solana wrong-key-rejection paths have dedicated tests. Fail-closed throughout: every 4xx/5xx/missing-`usd`/unknown-ticker path throws rather than fabricating a 0 or stale price, and the batch isolates per-query failures in input order. API host is pinned to `rootApiUrl` (the Vultisig proxy), not user-overridable.

## suggestion

`coingeckoId` (Route 0) is interpolated raw into `?ids=${coingeckoId}` with no `encodeURIComponent`. EVM/Solana addresses are regex-anchored so they can't inject, and `symbol` is only ever used as a map key (the mapped slug is what hits the wire), so the only unencoded path is the caller-supplied `coingeckoId` — and it fails closed (a smuggled extra param just makes `data[coingeckoId]` miss and throw). Low risk, but `encodeURIComponent` on the id/contract interpolations would be tidy defense-in-depth.

## risk

Low. Additive read-only display primitive, fail-closed on every error path, API host pinned. The one real correctness gap (POL/MATIC dead slug) is fixed and mutation-tested. No existing code paths touched beyond the export barrels.

## verdict

APPROVE. id mappings correct (incl. the rebranded POL/POL slug, now fixed + tested), batch token<->price association correct and order-preserving, fail-closed with no fabricated/stale price, API host pinned, RN export + changeset now present, tests prove mapping + association + fail-closed (and the new POL tests are mutation-verified). All gates green locally (note: gomes-branch fix-commits don't re-trigger CI — verified tsc/eslint/prettier/tests/changeset-gate locally on HEAD 00227fc1).

---
_Reviewed by Claude (load-bearing) + Codex (stalled, best-effort) + GLM (dead, best-effort). Fixes applied in-code per gomes fix-and-approve._

PR #818, opened 2026-06-23
feat(sdk): add sdk.balance.utxo (BTC/LTC/DOGE/BCH/DASH)

gomesalexandre:
## what

Adds `getUtxoBalance(chain, address, options?)` — a vault-free read primitive for the native balance of the five Blockchair-backed UTXO chains: **Bitcoin, Litecoin, Dogecoin, Bitcoin-Cash, Dash**.

Returns base-unit `satoshis` as a **string** (no float precision loss — 21M BTC in sats overflows `Number.MAX_SAFE_INTEGER`), a fixed-8-decimal human `balance`, the chain `symbol`, plus `chain`/`address` echoes. Also exports `formatUtxoBalance` (pure base-unit -> human, bigint-safe) and `supportedUtxoBalanceChains`.

Typed against the existing `@vultisig/core-chain` `UtxoChain` enum; Zcash is in that enum but intentionally **out of scope** here (rejected with a clear error) since shielded balances aren't represented by the dashboard endpoint.

Defaults to the Vultisig Blockchair proxy `https://api.vultisig.com/blockchair` (the same canonical, key-managed endpoint mcp-ts reads through). Hitting `api.blockchair.com` directly is rate-limited (HTTP 430) without a server-side key — so the base is injectable via `options.blockchairBase` for proxies/mirrors/api-keyed direct.

## why

Ported from mcp-ts `get_utxo_balance` (0 SDK imports). Part of the mcp-ts/backend -> SDK code-as-action consolidation: one canonical TS implementation the isolate, CLI, app card, and backend can all share instead of N drifting copies.

**Consumers it unblocks:** mcp-ts `get_utxo_balance` re-points to `sdk.balance.utxo`; agent-backend UTXO balance reads; SDK CLI/clients gain a first-class UTXO balance read alongside the existing EVM/coin balance helpers.

## test

- `yarn workspace @vultisig/sdk typecheck` — 0 errors (0 new from these files)
- `yarn workspace @vultisig/sdk vitest run tests/unit/utxoBalance.test.ts` — 10/10 pass (fetch mocked: bigint formatting, hyphenated `bitcoin-cash` path, null->0, custom base override, unsupported-chain + empty-address + non-200 rejections, 5-chain scope)
- eslint clean on the new files

### RECEIPT — live, cross-checked against 2 independent sources

```
$ node --import tsx scripts/receipts/balance_utxo.mjs
[receipt] sdk.balance.utxo — Bitcoin 1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa

[source A] SDK getUtxoBalance (Blockchair):
{
  "chain": "Bitcoin",
  "address": "1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa",
  "symbol": "BTC",
  "satoshis": "10722426196",
  "balance": "107.22426196"
}

[source B] blockchain.info addressbalance (satoshis): 10722426196

[cross-check] Blockchair sats=10722426196  blockchain.info sats=10722426196  →  MATCH ✓
[human]       107.22426196 BTC

[receipt] OK — two independent sources agree.
```

## Summary by CodeRabbit

* **New Features**
  * Added `sdk.balance.utxo` (`getUtxoBalance`) to read native UTXO balances for supported Bitcoin-family chains, returning precise satoshis plus an 8-decimal formatted balance.
  * Exported UTXO balance helpers (`formatUtxoBalance` and `supportedUtxoBalanceChains`), including availability from the React Native platform entry.
  * Added a runnable receipt script to cross-check fetched balances against an independent external source.
* **Tests**
  * Added unit tests covering formatting precision (including very large values), fetch behavior (including null->zero), custom overrides, and validation/error cases.


gomesalexandre (commented):
## deep adversarial pass — 1 real bug found + fixed, rest clean

Ran a full adversarial diff vs the canonical mcp-ts source (`src/tools/balance/utxo-balance.ts` + `lib/rpc.ts`) and live blockchair proxy probing. One genuine correctness bug, fixed on this branch in `1dc65828`.

### found + fixed — satoshis NOT actually bigint-safe (sev: medium, correctness)
`utxoBalance.ts:113-115` did `BigInt(addrData?.address?.balance ?? 0)` off `response.json()`. Problem: `response.json()` already coerces the balance literal to a lossy JS `number` **before** `BigInt()` ever sees it. The `UtxoBalance.satoshis` field is documented as a precise string to avoid precision loss — but for any balance past `Number.MAX_SAFE_INTEGER` (~9.0e15 base units) it was silently truncating. This is reachable in practice on high-supply chains: a single DOGE whale address holds >30e9 DOGE = 3e18 base units, well past the ceiling.

Proof:
```
expected : 3000000000000000001
OLD lossy: 3000000000000000000  WRONG ✗   (BigInt(JSON.parse(...).balance))
NEW exact: 3000000000000000001  OK ✓       (raw-text extraction)
```

Fix: read `response.text()` and pull the integer balance straight off the raw JSON (regex scoped to the `address` object's `balance` key, so it never grabs `balance_usd` floats or per-utxo `value`s) before any numberification. Also now throws on a non-JSON 200 body (`Blockchair returned non-JSON…`) instead of the old fail-open `?? 0`. BTC stays unaffected (max ~2.1e15 sats < safe int) so existing receipts are unchanged. Added 3 tests: DOGE-whale precision canary, balance_usd/per-utxo non-pickup, non-JSON 200.

### verified clean
- **blockchair path / chain set** — `chain.toLowerCase()` is correct because the enum *value* is already hyphenated (`UtxoChain.BitcoinCash = 'Bitcoin-Cash'` → `bitcoin-cash`). Matches mcp-ts `blockchairUrl()` exactly. The supported set (BTC/LTC/DOGE/BCH/DASH) and **Zcash exclusion** are correct — Zcash shielded balances aren't represented by the public dashboard endpoint, and the unsupported-chain guard hard-throws (test covers it).
- **proxy default** — `https://api.vultisig.com/blockchair` matches mcp-ts `ROOT_API/blockchair`; the direct `api.blockchair.com` is 430-rate-limited without a server key. Probed the proxy live, dashboard shape confirmed.
- **cross-source agreement** — receipt cross-checks blockchair vs blockchain.info on the BTC genesis address: `10722531594` sats both sources → MATCH, 107.22531594 BTC. The 1A1zP1… address is the coinbase address (permanently non-zero), good receipt anchor.
- **decimals/format** — all 5 chains are 8-decimal; `formatUtxoBalance` is pure-bigint (no float), negative-safe, zero-pads the fraction correctly.
- **null/unseen address** — blockchair fail-opens (HTTP 200 + `balance:0` for garbage addresses); handled as 0 same as mcp-ts (parity, noted not fixed — it's an upstream API trait, not a bug introduced here).

### receipts
- typecheck: 0 errors (`tsc --noEmit` on @vultisig/sdk)
- unit: 13 passed (was 11; +2 net, helpers updated for raw-body)
- live receipt: blockchair == blockchain.info, MATCH ✓

Own PR — comment only, not approving.

NeOMakinG (approved):
## testing approach

3-lane fan-out (Claude load-bearing; Codex `gpt-5.4` 0-byte stall = session contention, best-effort; GLM ECONNRESET, dead/best-effort). Worktree `feat_sdk_balance_utxo` @ `1dc65828` → re-pushed @ `a7e6ae5f` after the in-code fix below. Ran the full B-series balance-fabrication lens (the lens that caught sibling #843's fabricated-ZERO + Blockchair address-key drift), plus build/test/typecheck/lint/prettier gates and a live cross-checked receipt.

Gates (local, on `a7e6ae5f`):
- **Unit tests**: 13/13 pass (`VITEST_MAX_THREADS=1`).
- **Mutation teeth** (all killed): divisor 8→7 → 6 fail; HTTP-error fail-open→0 → 1 fail; extractor→`0n` → 4 fail; precision-loss via `response.json()` number instead of raw-text extractor → 1 fail (the DOGE-whale `...0001` canary). Every B-series invariant has a failing test behind it.
- **tsc** (`--noEmit`): exit 0, **0 errors** (after building `@vultisig/core-chain` to clear the workspace-dep wall) — including my RN-entry edit.
- **eslint/complexity** (sonarjs): clean on all new + edited files.
- **prettier** (`--config .config/.prettierrc`): all in-gate-scope files clean. (`scripts/receipts/balance_utxo.mjs` warns one wrapped line but `scripts/receipts/**` is OUT of the `format:check` glob → does not gate; see suggestion.)
- **CI**: Lint+TypeCheck / Unit / Integration / Test Summary / CodeRabbit all green (on `1dc65828`; my fix-commit doesn't re-trigger CI on this branch — gates re-verified locally).

**Live receipt** (real fund-safety proof, not a mock): BTC genesis address via SDK→Blockchair `satoshis=10722535962` `107.22535962 BTC`, cross-checked against blockchain.info `addressbalance` `10722535962` → **MATCH ✓**.

## B-series verdict (the bug class that bit #843)

- **Per-chain decimals (8, BigInt, no precision loss)**: ✅ `UTXO_DECIMALS=8`, `formatUtxoBalance` is pure-BigInt (`10n ** 8n`, integer div + padded mod). The satoshi integer is pulled off the **raw response text** by `extractBalanceSatoshis` *before* any `JSON.parse`/`response.json()` numberification, so DOGE-whale balances past `Number.MAX_SAFE_INTEGER` survive intact (verified by the `3000000000000000001` canary test). No `Number()` anywhere in the scaling path. This is the #1362-B3 precision bug, structurally avoided.
- **B1 fail-closed (no fabricated zero)**: ✅ non-OK HTTP → throw (`${status}`); non-JSON 200 body → throw (`non-JSON`); unsupported chain / empty address → throw. The only `0n` paths are (a) Blockchair `balance == null` (genuinely-unseen address) and (b) regex-no-match *gated behind* a non-null `addrData.address.balance` — i.e. a legit zero, never a parse/RPC-error zero.
- **B2 address-key drift (the exact #843 sibling bug)**: ✅ **safe**, verified against the LIVE proxy. `Object.values(json.data)[0]` is tolerant of the echoed key, and I confirmed Blockchair always echoes the queried address back as the `data` key on a 200 — even for invalid/unseen addresses (`balance:0`, not a missing key). So there is no "200 body that should contain the address but doesn't → fabricated 0" case here. The non-greedy `[^}]*?` regex scopes correctly to `address.balance` and does NOT match `balance_usd` (closing-quote required) — covered by the `does not pick up balance_usd` test, re-verified by hand against the real body shape.
- **Address form per chain**: ✅ pass-through (no transform); `chain.toLowerCase()` yields the exact Blockchair paths — incl. `Bitcoin-Cash`→`bitcoin-cash` (enum value is `'Bitcoin-Cash'`). All 5 paths hit live (200) and accept the caller's address form (Blockchair takes BCH cashaddr + legacy).

## should-fix (FIXED in-code @ `a7e6ae5f`)

The new `getUtxoBalance` was exported from `src/index.ts` + `src/tools/index.ts` but **not** from the React-Native platform entry (`src/platforms/react-native/index.ts`). RN consumers (vultiagent-app — the primary consumer) resolve `@vultisig/sdk` via the `react-native` export condition → `dist/index.react-native.js`, so the headline tool was unreachable on the platform that uses it most. This is the recurring RN-export gap (#831/#827/#828/#854/#856/#857/#859/#840/#842). Fixed with a static re-export (RN-safe: the reader is a pure `fetch` helper — only the `UtxoChain` enum string + global `fetch`, no heavy deps; same mechanism as the `../../tools/token` re-export directly above it). Re-verified: tsc 0 errors, eslint + prettier clean.

## suggestion

- `scripts/receipts/balance_utxo.mjs:36` has one over-long template-literal line prettier would wrap. It's outside the `format:check` glob so it doesn't gate, but `yarn prettier --write` on it keeps the receipts dir tidy. Non-blocking.

## q

- `extractBalanceSatoshis` reads from the *whole* `rawBody` while the null-guard reads from `Object.values(data)[0]`. They agree for a single-address dashboards response (the only shape this calls). If a future caller ever batches comma-separated addresses, the regex would return the *first* address's balance regardless of which `data[k]` the guard picks — worth a one-line comment pinning this to single-address use. Not a defect today (the public API only ever queries one address).

## risk

Low. Read-only, no signing/broadcast, no vault/keys. The fund-safety surface is "could it report a wrong/fabricated balance" and all four B-series failure modes are closed and test-pinned, with a live independent cross-check. The in-code RN fix is a pure additive re-export.

## verdict

**APPROVE.** Per-chain decimals correct (8, pure BigInt, no precision loss); B1 fail-closed; B2 address-key drift safe (verified live); correct address form; RN export now wired (fixed in-code); changeset well-formed (`@vultisig/sdk: minor`); tests prove decimals + fail-closed + B2 + large-balance precision with real mutation teeth; CI green + live receipt MATCH.

---
_3-lane review: Claude (load-bearing) + Codex gpt-5.4 (0-byte stall, best-effort) + GLM 4.6 (ECONNRESET, best-effort). RN-export fix pushed @ `a7e6ae5f`; CI not re-triggered on this branch, gates re-verified locally._

PR #820, opened 2026-06-23
feat(sdk): add sdk.evm.getTokenApprovals (enumerate ERC-20 approvals)

gomesalexandre:

## what

Adds `getTokenApprovals(chain, { owner })` to the SDK's EVM tools — enumerate the active ERC-20 approvals (spender allowances) for an address on any supported EVM chain. READ-ONLY: builds no tx, signs nothing.

Ported from mcp-ts `get_token_approvals` (`src/tools/evm/get-token-approvals.ts`) and reshaped to a pure-crypto SDK primitive (no agent/LLM concepts, no `ToolDef`/zod-schema wrapper — just a typed async fn).

Strategy mirrors revoke.cash:
1. `eth_getLogs` for `Approval(owner, spender, value)` where indexed owner == address (full-history "earliest" scan, with a bounded recent-window fallback when an RPC rejects the unbounded range).
2. De-dupe to unique `(token, spender)` pairs (capped, over-fetched a touch so zero-allowance pairs don't starve the active set).
3. Per pair, read the CURRENT `allowance(owner, spender)` so revoked/spent approvals drop out.
4. Read `symbol()` per surviving token (fail-soft → `null` for non-standard tokens).
5. Flag `isUnlimited` for allowances >= 2^128 (matches revoke.cash's "unlimited" sentinel).

Uses the SDK's existing `getEvmClient` (viem public client) — the raw `eth_getLogs` topic filter is intentional (provider-agnostic, explicit filter shape) while `allowance`/`symbol` go through viem `readContract`. Exported from the package index alongside `evmCheckAllowance` / `evmCall` etc.

## why

`get_token_approvals` is flagged BRING-INTO-SDK in the tools→SDK table (`sdk.evm.getTokenApprovals(addr, chain)`). Pulling it into the SDK lets mcp-ts and agent-backend stop hand-rolling the indexer fan-out + allowance cross-check, and pairs cleanly with the approval/revoke calldata builders (`encodeErc20Approve` / `encodeErc20Revoke`) that audit-then-revoke flows need.

Consumers this unblocks:
- **mcp-ts** — `get_token_approvals` tool becomes a thin SDK wrapper (drops the duplicated topic/allowance logic).
- **agent-backend** — approval-audit / revoke flows call the SDK directly instead of round-tripping mcp-ts.
- **Station / Windows extension** — same primitive for an in-app "approvals you can revoke" view.

## receipt

Live Ethereum RPC, real address (vitalik.eth) with active approvals — spender + allowance per token, unlimited flagged:

```
$ node --import tsx scripts/receipts/evm_get_token_approvals.mjs
> sdk.evm.getTokenApprovals("Ethereum", { owner: "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045" })

owner:       0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045
chain:       Ethereum
totalCount:  13
elapsed:     5633ms

  ?          token=0xd654bDD32FC99471455e86C2E7f7D7b6437e9179 spender=0xF51BC4633F5924465c8c6317169Faf3E4312E82F allowance=200000000000000000000
  ?          token=0xC66eA802717bFb9833400264Dd12c2bCeAa34a6d spender=0xF51BC4633F5924465c8c6317169Faf3E4312E82F allowance=60000000000000000000
  ?          token=0xECF8F87f810EcF450940c9f60066b4a7a501d6A7 spender=0x5a70dB543902f4CCccf028B19d2142FDE7eb0593 allowance=200000000000000000000
  ?          token=0xECF8F87f810EcF450940c9f60066b4a7a501d6A7 spender=0xc8B976247A04025E24Ff5f078B80A4cB42a579db allowance=999756580000000000000000
  ?          token=0xECF8F87f810EcF450940c9f60066b4a7a501d6A7 spender=0x364d85D117B9908aDF6a441DC12638bdA963fc01 allowance=999734941672201201000000
  ?          token=0xC66eA802717bFb9833400264Dd12c2bCeAa34a6d spender=0x364d85D117B9908aDF6a441DC12638bdA963fc01 allowance=99949000000000000000000
  ?          token=0xECF8F87f810EcF450940c9f60066b4a7a501d6A7 spender=0xdc7d570Bb61a80e43eB250BbBB47754aE9E6449A allowance=999999000000000000000000
  ?          token=0xC66eA802717bFb9833400264Dd12c2bCeAa34a6d spender=0xdc7d570Bb61a80e43eB250BbBB47754aE9E6449A allowance=999877117242000000000000
  DAI        token=0x6B175474E89094C44Da98b954EedeAC495271d0F spender=0xE5BFAB544ecA83849c53464F85B7164375Bdaac1 allowance=unlimited
  APE        token=0x978a77Ef76f23C06D951d2E827741ED334E2fF2F spender=0x9C2A328A8862f111E74f4E37380e35C63A8F7Fe0 allowance=2967928298073333764205836908622
  VISH       token=0x69627B7e0D9E74690C7489D5B829968Dc15bDcDE spender=0x5C14Cee5849e3959BFC5B506F9CA1B7089Aeea2F allowance=unlimited
  RAI        token=0x03ab458634910AaD20eF5f1C8ee96F1D6ac54919 spender=0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45 allowance=unlimited
  DAI        token=0x6B175474E89094C44Da98b954EedeAC495271d0F spender=0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45 allowance=unlimited
```

Unit test (`tests/unit/tools/evm/getTokenApprovals.test.ts`) — 4 cases: invalid-address guard, empty result, de-dupe + zero-allowance filter + unlimited flag, bounded-window RPC fallback:

```
✓ rejects an invalid owner address before hitting the network
✓ returns empty when no Approval events exist
✓ de-dupes (token, spender) pairs, drops zero/revoked allowances, and flags unlimited
✓ retries over a bounded window when the RPC rejects an unbounded scan

 Test Files  1 passed (1)
      Tests  4 passed (4)
```

Typecheck: `yarn workspace @vultisig/sdk typecheck` → 0 errors (these files add 0 new). Lint clean.

## risk

Low — additive read-only primitive, no signing path touched, exported behind a new symbol. Fail-soft per-pair reads (revert → 0n / null) keep one bad token from nuking the whole enumeration.


gomesalexandre (commented):
## adversarial pass — getTokenApprovals

Did a real line-diff vs the mcp-ts source (`src/tools/evm/get-token-approvals.ts`) + adversarial reasoning on the focus areas. Verdict: **logic clean, no fund-safety holes**. Pushed one test-only hardening (`80810385`).

### what was checked

**Approval-event fan-out window** — `eth_getLogs` topic filter is byte-identical to source: `topics=[APPROVAL_TOPIC, ownerTopic]`, `topic[0]`=`keccak256("Approval(address,address,uint256)")`, `topic[1]`=indexed owner padded to 32B, spender read from `topic[2].slice(-40)`. Dedup→probe→filter>0→cap is the same `MAX_PAIRS*2` (100) over-fetch / `slice(0, 50)` trim as mcp-ts. No off-by-one, no missed pairs.

**Allowance read accuracy** — reads CURRENT on-chain `allowance(owner, spender)` per pair via viem `readContract` (SDK convention; mcp-ts uses raw `eth_call`+`decodeFunctionResult`, semantically equivalent). Revoked/spent → `0n` → filtered out. Revert → `0n` fail-soft. No stale-from-event-value bug (it never trusts the event `value`, always re-reads).

**Unlimited threshold (2^128 vs 2^256)** — `UNLIMITED_THRESHOLD = 2n ** 128n`, `isUnlimited = allowance >= THRESHOLD`. Matches source + revoke.cash exactly. **This was the only focus area with a real gap: zero boundary test coverage** (existing tests used `2^200` and `1e6`). Added a parametrized 4-case test pinning `2^128-1` (false), `2^128` (true, inclusive), `2^128+1` (true), `MaxUint256` (true) — a regression flipping `>=`→`>` or `2^128`→`2^256` now fails CI.

**RPC fallback** — unbounded `earliest` scan → on reject (`block range`/`too large`/`exceeds`/`limit`/`maximum`/`range`) falls back to `latest - 10_000` blocks. Identical match-list + bound to source. Non-matching errors re-throw (no fail-open). SDK throws on provider error vs mcp-ts's MCP-envelope fail-soft — that's the correct SDK convention (sibling `evmCall`/`abiEncode` throw too), not a regression.

**No missed approvals / attacker angle** — owner is validated + checksummed pre-network; `isAddress(strict:false)` gates both token (`log.address`) and spender before `getAddress`, so a malformed log can't crash the batch or inject a bad pair. Worst case under a range-rejecting public RPC is the *known* ~10k-block (~1.4d on mainnet) look-back limit — present identically in the source, not a port regression.

### receipt — live mainnet (vitalik.eth)
```
> sdk.evm.getTokenApprovals("Ethereum", { owner: "0xd8dA…6045" })
totalCount: 13   elapsed: 10412ms
  DAI   spender=0xE5BF…aac1  allowance=unlimited
  APE   spender=0x9C2A…7Fe0  allowance=2967928298073333764205836908622   <- ~2.97e30 < 2^128, correctly NOT unlimited
  RAI   spender=0x68b3…Fc45  allowance=unlimited
  DAI   spender=0x68b3…Fc45  allowance=unlimited
  …finite ones shown raw
```
Boundary behaviour confirmed on real data: the `~2.97e30` allowance is below `2^128` (`~3.4e38`) and correctly flagged finite; dedup + symbol-revert→`?` working.

typecheck: 0 new errors · unit: 8/8 (4 new boundary cases) · receipt: live mainnet 200, 13 approvals.

Adversarial pass clean on fan-out window / allowance accuracy / 2^128 threshold / RPC fallback / missed-approvals.

packages/sdk/src/tools/evm/getTokenApprovals.ts
```diff
@@ -0,0 +1,210 @@
...
+    const value = await client.readContract({
+      address: token,
+      abi: erc20Abi,
+      functionName: 'allowance',
+      args: [owner, spender],
+    })
+    return typeof value === 'bigint' ? value : 0n
+  } catch {
+    return 0n
+  }
```
NeOMakinG, line 107: `should-fix:` fail-soft → `0n` here silently drops a live approval if this single-pair `allowance()` call hits a transient RPC error / 429 / ABI quirk, because the caller filters `> 0n`. For a drain-risk safety tool consider distinguishing "revoked (read returned 0)" from "read failed" so a short/empty list can't be misread as safe. Matches the merged canonical, so not blocking — tracking.

packages/sdk/src/tools/evm/getTokenApprovals.ts
```diff
@@ -0,0 +1,210 @@
...
+    const msg = err instanceof Error ? err.message : String(err)
+    // Some RPCs reject an unbounded "earliest" scan. Retry over a bounded
+    // recent window so we still surface the most relevant approvals.
+    const boundedScanRejected =
+      msg.includes('block range') ||
+      msg.includes('too large') ||
+      msg.includes('exceeds') ||
+      msg.includes('limit') ||
+      msg.includes('maximum') ||
+      msg.includes('range')
```
NeOMakinG, line 77: `should-fix:` substring match is loose — `"rate limit exceeded"` matches `limit`+`exceeds`, a quota error matches `maximum`. A real (non-range) RPC failure then silently downgrades the full-history scan to a 10k-block window, hiding old-but-live approvals and swallowing the true error. Prefer range-specific signals so a transient rate-limit fails closed instead of returning partial results.

NeOMakinG (approved):
Fix-and-approve. READ-ONLY ERC-20 approval enumerator — a faithful port of the merged mcp-ts `get_token_approvals` (#488), reshaped to a pure SDK primitive. Enumeration is complete (full-history `eth_getLogs` "earliest" scan + bounded fallback, **live `allowance()` re-read per pair** so revoked/spent drop out, dedup, `isUnlimited` flagged at 2^128). I pushed two fixes (RN export + changeset) and verified the remaining concerns are inherited-from-canonical design characteristics, not regressions — none safety-blocking given the live receipt surfaces old-but-live approvals correctly.

## testing approach
Worktree from `origin/feat_sdk_evm_get_token_approvals`, built `@vultisig/core-chain`, ran the unit suite via an in-package isolated vitest config (alias `@→src`, empty `setupFiles[]` to dodge the `@vultisig/mpc-types` wall), plus mutation checks, the live Ethereum receipt, and CI-exact gates.

- **Unit test** (8/8 pass: 4 base + 4 boundary): `VITEST_MAX_THREADS=1 vitest run` on `getTokenApprovals.test.ts`.
- **Mutation teeth** (tests are NOT inert):
  - flip filter to fail-open (`> 0n` → `>= 0n`, keep revoked) → **1 fail** (revoked-WETH test catches it)
  - skip live allowance re-read (`readAllowance(...)` → `Promise.resolve(1n)`) → **6 fail** (proves live re-read is asserted, not stale event value)
  - `UNLIMITED_THRESHOLD` 2^128 → 2^256 → **4 fail** (boundary tests catch it)
- **Live receipt** (real Ethereum RPC, vitalik.eth): 13 active approvals, unlimited flagged, old-but-live DAI/RAI/VISH approvals surfaced (earliest scan succeeded). Re-ran after my format change — identical.
- **Gates (local, fix-SHA `d1646ba6`)**: `typecheck` exit 0 (0 errors); `quality:complexity` (sonarjs) exit 0; `format:check` (CI-exact glob) clean; `changeset:check-sdk-bundle` pass; `check:shared-exports` up-to-date. PR CI all green @ prior SHA (Unit/Integration/Lint+TypeCheck/CodeRabbit). Note: my fix commit may not re-trigger CI on the gomes branch — gates verified locally.
- **3-lane**: Claude (load-bearing) + Codex gpt-5.4 (converged, all 5 findings = canonical design characteristics) + GLM (dead, exit 1, best-effort).

## blocking
None.

## should-fix
1. **`readAllowance` fail-soft → `0n` silently drops a live approval on a transient single-pair RPC error.** A 429/timeout/ABI-quirk on one pair's `allowance()` call collapses to `0n`, which `.filter(p => p.allowance > 0n)` then removes — a false-negative for a drain-risk tool. This matches the canonical mcp-ts behavior exactly, so it's not a regression, but for an SDK safety primitive (no LLM to read a "note") consider distinguishing "0n = revoked" from "0n = read failed" and surfacing the latter (e.g. an `incomplete`/`errors` field) so an empty/short list can't be misread as "safe". Tracking, not blocking.

2. **Bounded-fallback substring heuristic can mis-classify a real RPC error as a range error** and silently downgrade a full scan to the last 10k blocks, hiding old-but-live approvals AND swallowing the true error. `"rate limit exceeded"` matches both `limit` and `exceeds`; quota errors match `maximum`. Again identical to canonical. Consider tightening to range-specific signals (or only falling back on a known range-error code) so a transient rate-limit becomes a hard error (fail-closed) rather than a partial 10k-block scan.

## suggestion
3. **`totalCount` after `MAX_PAIRS` truncation no longer means "total active approvals".** Once capped at 50 it equals "returned approvals", so a caller can't tell "exactly 50 exist" from "50 shown, more omitted". The canonical surfaces a `note` for this; the SDK shape dropped it. A `truncated: boolean` (or keeping a pre-slice count) would let consumers flag incompleteness in an "approvals you can revoke" view.

4. **Dedup keeps oldest pairs first** (log order is oldest-first; collection stops at `MAX_PAIRS*2`). An address with >100 unique historical approvals can fill the cap with stale pairs before recent ones are examined. Edge case for power users; matches canonical.

5. Receipt `.mjs` reformatted by my push (was the only file failing `prettier --check`; CI `format:check` glob excludes `.mjs` so it wasn't red, but tidied for hygiene).

## q
6. Is the SDK's choice to **throw on bulk log-fetch failure** (vs the canonical returning `{approvals:[], note:"provider error"}`) intentional? It's the right call for an SDK primitive — fail-closed beats a fail-open empty list — just confirming it's deliberate and consumers expect to catch.

## risk
Low. Additive read-only primitive, no signing/broadcast path. Fixes are an RN re-export (static-safe, same surface as sibling `evmCheckAllowance`/`evmCall`) + a changeset + a prettier pass. The should-fix items are pre-existing in the merged canonical and don't defeat the happy path (receipt proves old-but-live approvals surface).

## verdict
**Approve.** Enumeration is complete and correct (live allowance re-read proven by mutation, dedup, unlimited-flagged), fail-closed on bulk fetch, decode is raw-uint256 BigInt, RN export + changeset now present, tests have teeth, CI green. The 5 should-fix/suggestion items are inherited-from-canonical hardening opportunities worth a follow-up but none are safety-blocking.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — reviewer agent (Claude + Codex gpt-5.4 + GLM lanes)

PR #821, opened 2026-06-23
feat(sdk): add sdk.prep.cw20Transfer (CosmWasm CW20 transfer)

gomesalexandre:
## what

Adds `buildCw20TransferMsg` (exposed as `sdk.prep.cw20Transfer`) — a pure-crypto builder for an UNSIGNED CosmWasm CW-20 token transfer. It validates the addresses + amount and emits the canonical `MsgExecuteContract` amino message carrying `execute_msg = JSON.stringify({transfer:{recipient,amount}})`, ready to feed into `prepareSignAminoTxFromKeys`.

Ported from mcp-ts `src/tools/send/build-cw20-transfer.ts` (0 SDK imports), with all orchestration stripped: no LCD account fetch, no blockaid scan envelope, no ToolDef plumbing. Account-number / sequence / fee / chain-id stay in the caller's orchestration layer where they belong (chain-state, not crypto, must be fetched fresh).

## why

Part of the mcp-ts/backend to SDK code-as-action consolidation. CW-20 transfer construction was living in mcp-ts; moving the pure-crypto core into the SDK lets the app, extension, and mcp-ts all share one audited builder instead of three drifting copies.

**Pure crypto — zero network I/O. Never signs, never broadcasts.** `vault.sign` stays on-device; this only BUILDS the unsigned msg.

Fund-safety invariants carried over from the source:
- `recipient` is the human recipient, NOT the contract (the transfer body expects the recipient)
- validator HRPs (`...valoper1...` / `...valcons1...`) rejected before a generic prefix mismatch
- native bank denoms (`uluna`, `uusd`, ...) rejected — those are bank sends, wrong path for CW-20
- `execute_msg` JSON-stringified to match the SDK protobuf `string` field (an object leaks an unsignable shape)
- amount validated as a positive integer base-unit string, leading zeros canonicalized

## consumers unblocked

- mcp-ts `build_cw20_transfer` can drop its hand-rolled envelope and call `sdk.prep.cw20Transfer`
- Station / Windows extension get the same CW-20 builder for free (Osmosis, Kujira, Terra, TerraClassic)

## RECEIPT

Real run — builds an unsigned CW-20 transfer with throwaway inputs, prints structure, nothing signed or broadcast:

```
$ node --import tsx scripts/receipts/prep_cw20_transfer.mjs

=== sdk.prep.cw20Transfer — UNSIGNED CW-20 transfer (no broadcast) ===

inputs:
{
  "bech32Prefix": "osmo",
  "contract": "osmo1kyekxn2qmcjt902sywxm42a2h2d35ssn9ljpvuf77mewevup4kds298e77",
  "recipient": "osmo12f8hyk2prj2f5w2j3at9ndrxw390ejkr5nt99h",
  "amount": "1000000",
  "sender": "osmo1c3a7qq6trpvdver98agv6d9cqex94889k5ejr7"
}

execute_msg (stringified, CW-20 transfer):
{"transfer":{"recipient":"osmo12f8hyk2prj2f5w2j3at9ndrxw390ejkr5nt99h","amount":"1000000"}}

MsgExecuteContract amino msg (feed into prepareSignAminoTxFromKeys):
{
  "type": "wasm/MsgExecuteContract",
  "value": {
    "sender": "osmo1c3a7qq6trpvdver98agv6d9cqex94889k5ejr7",
    "contract": "osmo1kyekxn2qmcjt902sywxm42a2h2d35ssn9ljpvuf77mewevup4kds298e77",
    "msg": { "transfer": { "recipient": "osmo12f8hyk2prj2f5w2j3at9ndrxw390ejkr5nt99h", "amount": "1000000" } },
    "funds": []
  }
}

fund-safety guard (validator recipient rejected):
  invalid recipient: validator address (osmovaloper) is not a valid CW-20 recipient

=== receipt OK — unsigned msg built, nothing signed or broadcast ===
```

Verification:
- `yarn workspace @vultisig/sdk typecheck` — 0 errors (0 new from these files)
- `yarn workspace @vultisig/sdk test` — 1519 passed (incl. 10 new cw20Transfer cases)

## Summary by CodeRabbit

* **New Features**
  * Added `buildCw20TransferMsg` to the SDK to construct unsigned CW-20 transfer `MsgExecuteContract`-style messages with pure input validation (bech32 address checks, contract payload size expectations, and safe amount rules).
  * Exposed the helper plus its parameter/result types through the SDK’s public entry points, including the React Native platform surface.
* **Tests**
  * Added unit tests covering CW-20 transfer message construction and validation edge cases.


gomesalexandre (commented):
## deep adversarial review — prep.cw20 vs mcp-ts source

Reviewed `packages/sdk/src/tools/prep/cw20Transfer.ts` against the mcp-ts source `src/tools/send/build-cw20-transfer.ts` + shared `src/lib/validateBaseUnitAmount.ts` + `src/lib/cosmos-address-guard.ts`. The port is solid — one byte-parity divergence fixed, rest of the fund-safety surface verified clean.

### fixed in e9dc17ea

- **P2 byte-parity (audit) — `validateBaseUnitAmount` leading-zero strip** (`cw20Transfer.ts:131-141`). The SDK did `trimmed.replace(/^0+/, '')` ("01000" -> "1000") whereas mcp-ts's shared `validateBaseUnitAmount` returns the trimmed string unchanged. The amount is embedded verbatim into the signed `execute_msg` JSON (`{"transfer":{"recipient","amount"}}`) AND the amino `value.msg.transfer.amount`, so the strip produced a *different signed wire payload* than the mcp-ts source for the same input. Numerically equal, not byte-identical. Dropped the strip → returns `trimmed`. Also tightened the zero-check from `trimmed === '0' || /^0+$/.test(trimmed)` to `BigInt(trimmed) <= 0n` to mirror the canonical mcp-ts guard exactly. Test flipped from "canonicalizes leading zeros" to byte-parity assertions (verbatim amount flows into both `executeMsg` and `value.msg.transfer.amount`) + added a whitespace-trim case.

- **Validator HRP guard** — SDK `isValidatorHrp` (`['valoper','valcons']` suffix match) is functionally equivalent to mcp-ts `assertNotValidatorHrp`. mcp-ts lowercases the HRP first; the SDK doesn't, but `bech32.decode` already normalizes/rejects mixed-case per BIP173, so `decoded.prefix` is always lowercase. Defense-in-depth via the exact `decoded.prefix !== bech32Prefix` check after. `terravaloper1…`/`…valcons1…` rejected before the generic prefix mismatch — confirmed by receipt + unit test.
- **recipient ≠ contract** — `transfer.recipient` carries the human recipient, never the contract; no `to`-defaults-to-self leak (no defaulting at all — caller must pass an explicit recipient). Prompt-injection proceed-redirect surface is closed.
- **amino msg shape** — `wasm/MsgExecuteContract` is the correct cosmjs amino type; `value.msg` is the JSON *object* form (correct for the amino sign path, distinct from the SDK-proto string field which `executeMsg` re-emits for proto callers). No object-vs-string leak.
- **native-denom rejection** — SDK rejects native bank denoms (`uluna`/`uusd`) routed as a CW-20 contract via the caller-supplied `nativeDenoms` list. mcp-ts's richer chain-keyed ticker+denom blacklist is orchestration-layer (chain config lives upstream); the SDK primitive correctly takes the denom list as a param rather than hardcoding chain tables. Wrong-path tx is blocked — confirmed by test.
- **bech32 payload length** — 20/32-byte enforcement matches mcp-ts. `funds: []` (pure transfer carries no coins). No affiliate/fee field exists on this primitive — nothing to leak.

### receipts (worktree, nothing signed or broadcast)
- typecheck: `yarn workspace @vultisig/sdk typecheck` → 0 errors
- unit: `vitest run …/cw20Transfer.test.ts` → 10/10 pass
- receipt: `node --import tsx scripts/receipts/prep_cw20_transfer.mjs` → unsigned msg built + validator-recipient rejection fired (`invalid recipient: validator address (osmovaloper) is not a valid CW-20 recipient`)

Our own PR — comment only, no approve/request-changes.

NeOMakinG (approved):
fix-and-approve. Reviewed the CW-20 `MsgExecuteContract` builder as a calldata/msg builder (highest fund-risk class) across 3 lanes: Claude (load-bearing) + Codex gpt-5.4 (converged) + GLM (dead, ECONNRESET — best-effort). All fund-binding fields verified correct; I pushed 3 follow-up fixes to your branch (changeset, RN export, doc clarity) at `4ce3a9c9` — none change the built bytes.

## testing approach
- Worktree off `origin/feat_sdk_prep_cw20_transfer`, true diff vs merge-base = +403 clean (the large index removals are a merge-base-behind-main phantom; real diff is 3 lines added to each index).
- Traced the full amino consumer path: `buildCw20TransferMsg` → `CosmosMsgInput.value` → `buildCosmosPayload.ts:126` (`CosmosMsgSchema.value = m.value` verbatim) → `core/mpc/.../resolvers/cosmos/index.ts:72` (`rawJsonMessage.value` verbatim into WalletCore). The amino `value` is forwarded byte-for-byte to the signer.
- Adversarial standalone harness against the live logic (real `bech32`): all guards have teeth — see below.
- `yarn vitest run` cw20Transfer suite: **10/10 pass** (single-thread). `tsc` clean (0 new errors from cw20Transfer.ts or my RN edit; the 3 `@vultisig/mpc-types` TS2307 are pre-existing on RN index lines 45/88/89, not my lines). eslint/sonarjs exit 0. prettier clean (.config/.prettierrc). Receipt runs and prints the unsigned msg.
- `gh pr checks`: all green (Unit, Integration, Lint+TypeCheck, CodeRabbit) on the pre-fix SHA. Note: our pushed fix-commits don't re-trigger CI on gomes branches — gates verified locally on `4ce3a9c9`.

## fund-safety verification (all PASS)
- **msg shape**: `value.msg = { transfer: { recipient, amount } }`, `type = wasm/MsgExecuteContract`. `transfer` (not `transfer_from`/`send`). ✅
- **inner `.msg` as OBJECT (not stringified)**: independently confirmed correct by Codex — this is the canonical cosmjs `AminoMsgExecuteContract` convention, matches existing SDK `clients/cli/src/commands/execute.ts:253`, and is forwarded verbatim to WalletCore. The returned `executeMsg` string is the *separate* proto path. ✅
- **recipient = human dest, not contract**; test pins `transfer.recipient !== contract`. ✅
- **contract = MsgExecuteContract.contract** = validated CW-20 bech32 with correct HRP. ✅
- **sender = own addr**, written to `value.sender`. ✅
- **funds = []** (no native coins attached) — harness + receipt confirm. ✅
- **amount**: ASCII `\d` integer string, BigInt > 0 check; rejects float/negative/hex/unicode-digit/empty/zero; leading zeros kept verbatim for documented mcp-ts byte-parity. No float, correct base-units string. ✅
- **fail-closed**: malformed bech32, wrong HRP, validator HRP (`valoper`/`valcons`), native bank denom, bad amount → all throw before build. ✅
- **unsigned, zero network I/O**: pure function, only `bech32` + type-only imports. ✅

## should-fix (FIXED by me at `4ce3a9c9`)
- **missing changeset** — `sdk.prep.cw20Transfer` is new public API on the published `@vultisig/sdk` (2.5.0); the release-pr workflow only opens a version PR when a changeset exists, so this was green-but-unpublishable. Added `.changeset/prep-cw20-transfer.md` (minor).
- **RN export gap** (recurring) — the sibling prep helpers are all RN-wired but `buildCw20TransferMsg` was not. Added a static re-export to `src/platforms/react-native/index.ts` (safe: pure, only depends on `bech32`, no eager native dep).
- **misleading docstring** (Codex Medium, converged) — the original "execute_msg stringified … object leaks an unsignable shape" reads as if the amino `value.msg` must be a string. It must NOT — it's correctly an object. Reworded to distinguish the two wire paths so a future "fix" doesn't stringify `value.msg` and silently break sign-bytes.

## q
- No guard against `recipient === contract` (sending the CW-20 to its own contract). Benign self-loop / generally a no-op-or-burn at the contract level, and out of scope for a pure builder, but worth a one-line note if you want symmetry with future burn-guard work. Not blocking.
- Terra and TerraClassic share the `terra` HRP, so a Terra recipient + TerraClassic sender both pass the prefix check — chain disambiguation is correctly the caller's responsibility (`bech32Prefix` is caller-supplied). Documented, not a defect.

## risk
Low. Pure unsigned builder, no signing/broadcast/network. Msg shape independently verified correct against the live amino→WalletCore path and existing SDK CLI precedent. My fixes are metadata + doc + an additive RN export — zero change to built bytes. Tests pin exact msg + negatives with mutation teeth.

## verdict
**APPROVE.** Msg shape, recipient/contract/sender HRP validation, amount/decimals base-units handling, `funds=[]`, fail-closed, unsigned — all correct. Changeset + RN export + doc clarity added on your branch. CI green (verify gates locally on the fix SHA).

---
🤖 3-lane review: Claude (load-bearing) + Codex gpt-5.4 (converged) + GLM 4.6 (dead/best-effort)

PR #824, opened 2026-06-23
feat(sdk): add sdk.prep.splTransfer (Solana SPL token transfer)

gomesalexandre:
## what

Adds `buildSplTransfer` (exported at the top level + `tools/prep`), the SPL token-transfer escape-hatch. `prepareSendTxFromKeys` only covers native SOL; this is the ATA-aware sibling for SPL tokens.

From `mint` + `from` + `to` + `amount` + `decimals` it:
- validates `from`/`to`/`mint` are real 32-byte base58 pubkeys,
- deterministically derives the sender + recipient Associated Token Accounts via PDA math (`getAssociatedTokenAddressSync`) — pure crypto, **no RPC**,
- builds an unsigned SPL `transferChecked` instruction (carries mint+decimals so the runtime rejects a decimals mismatch),
- returns the derived ATAs + the unsigned instruction (program id, account metas, base64 data).

Supports both the legacy Token Program and Token-2022 (`isToken2022`).

## why

Part of the mcp-ts/backend -> SDK code-as-action consolidation. The pure-crypto SPL builder lived in mcp-ts `build_spl_transfer_tx`; pulling it into the SDK gives every consumer (mcp-ts, agent-backend) one place to build an unsigned SPL transfer instead of re-deriving ATAs per-repo (the cross-repo drift root cause).

It **NEVER signs and NEVER broadcasts** — the security boundary is preserved. The recent blockhash, priority fee, recipient create-ATA, and final signature all stay on-device in `vault.sign`. This only does the offline, reproducible crypto.

Fund-safety guards ported from mcp-ts (incl. NeOMakinG #415 review):
- reject `to === mint` (a mint is a valid pubkey; sending to it credits an ATA owned by the mint authority -> funds lost),
- valid base58 pubkeys for `from`/`to`/`mint`,
- `amount > 0`, `decimals` in `[0, 255]`.

## consumers unblocked

- **mcp-ts** — drop the hand-rolled `build_spl_transfer_tx` descriptor in favour of the SDK builder.
- **agent-backend** — SPL sends through the same SDK prep surface as native SOL.

## receipt

```
$ yarn workspace @vultisig/sdk tsx scripts/receipts/prep_spl_transfer.mjs

=== sdk.prep.splTransfer — UNSIGNED SPL transfer (1 USDC) ===
{
  "chain": "Solana",
  "mint": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
  "from": "7xKXtg2CW87d97TXJSDpbD5jBkheTqA83TZRuJosgAsU",
  "to": "9WzDXwBbmkg8ZTbNMqUxvQRAyrZzDsGYdLVL9zYtAWWM",
  "amount": "1000000",
  "decimals": 6,
  "programId": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
  "isToken2022": false,
  "fromTokenAccount": "C4PRXFV6Gf5mytVZb6RoeLsG8CjcFWzR2EJ3dvwPTUJH",
  "toTokenAccount": "FGETo8T8wMcN2wCjav8VK6eh3dLk63evNDPxzLSJra8B",
  "instruction": {
    "programId": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
    "keys": [
      { "pubkey": "C4PRXFV6Gf5mytVZb6RoeLsG8CjcFWzR2EJ3dvwPTUJH", "isSigner": false, "isWritable": true },
      { "pubkey": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v", "isSigner": false, "isWritable": false },
      { "pubkey": "FGETo8T8wMcN2wCjav8VK6eh3dLk63evNDPxzLSJra8B", "isSigner": false, "isWritable": true },
      { "pubkey": "7xKXtg2CW87d97TXJSDpbD5jBkheTqA83TZRuJosgAsU", "isSigner": true,  "isWritable": false }
    ],
    "data": "DEBCDwAAAAAABg=="
  }
}

NOTE: no signatures, no blockhash, no broadcast. vault.sign attaches those on-device.
```

`data` decodes to: discriminator `12` (TransferChecked) + u64 amount `1000000` + u8 decimals `6`. Only the owner (`7xKX…`) is a signer.

## tests + verify

- `yarn workspace @vultisig/sdk vitest run tests/unit/buildSplTransfer.test.ts` -> 7 passed (ATA derivation matches canonical, Token-2022 path, determinism, all 4 fund-safety guards).
- `yarn workspace @vultisig/sdk typecheck` -> clean (0 errors, 0 in these files).

## risk

Low. Pure-crypto, additive, behind a new export. Never signs, never broadcasts.

## Release Notes

* **New Features**
  * Added SPL token transfer builder to the SDK for constructing unsigned token transfer instructions optimized for on-device signing.
  * Automatically derives Associated Token Accounts and supports both legacy SPL Token and Token-2022 programs with deterministic behavior.

* **Tests**
  * Added comprehensive unit tests validating correct instruction construction, deterministic output, and robust error handling.


gomesalexandre (commented):
## adversarial pass — 1 fund-safety bug found + fixed, rest clean

Did a real adversarial pass on the SPL builder (the vs-source audit was light here since this isn't a 1:1 mcp-ts port — the canonical `build_spl_transfer_tx` only emits a high-level `{action:'transfer', mint, from, to, amount}` intent and the ATA derivation + `transferChecked` build happens on-device; this PR moved that crypto into the SDK, so the bar is higher).

### found + fixed (preferably-blocking, fund-safety)
**`amount` had no u64 upper bound → silent wrap mismatch.** `createTransferCheckedInstruction` encodes `amount` as a little-endian u64 and `@solana/spl-token` wraps `mod 2^64` *silently* instead of throwing. So:

```
amount = U64_MAX + 1     → result.amount = "18446744073709551616"  but instruction.data encodes 0
amount = U64_MAX + 1000  → result.amount = "18446744073709552615"  but instruction.data encodes 999
```

The returned object and the signing-ready bytes handed to `vault.sign` disagreed — the exact silent fund mismatch the rest of the guards exist to stop. Added `amount <= U64_MAX` reject (`splTransfer.ts:158`) + test covering U64_MAX round-trip + wrap-to-0 + wrap-to-999 (`buildSplTransfer.test.ts:111`). Fix in `dc902db1`.

### verified clean (attacker-break attempts that hold)
- **ATA derivation (from + to):** matches canonical `getAssociatedTokenAddressSync` PDA math exactly for both legacy + Token-2022; `allowOwnerOffCurve=false` is correct (owners are on-curve wallets). Receipt ATAs reproduce.
- **`token === mint` reject:** present + tested (mint-authority funds-lost trap).
- **decimals:** `transferChecked` carries mint+decimals; `[0,255]` integer guard tested; encoded byte `data[9]` verified.
- **transferChecked vs transfer:** uses `transferChecked` (discriminator `12`), strictly safer — confirmed.
- **Token-2022 program-id:** legacy vs 2022 program id flows through to both ATA seed + instruction programId; ATAs differ between paths (tested).
- **owner-only signer:** account metas verified — only `from` is `isSigner:true`; ATAs writable, mint read-only.
- **wrong-chain / wrong-length `to`:** `new PublicKey()` guard is equivalent to mcp-ts `decodeBase58 + len===32` (probed: no divergence on 1-byte / EVM-0x / short base58 — all reject; this web3.js does not left-pad). EVM `0x…` recipient rejected (tested).

### receipts
- `vitest run tests/unit/buildSplTransfer.test.ts` → **8 passed** (7 original + new u64 guard).
- `typecheck` → **0 errors**.
- `tsx scripts/receipts/prep_spl_transfer.mjs` → unsigned 1-USDC transferChecked, owner sole signer, derived ATAs, `data=DEBCDwAAAAAABg==` (disc 12 + u64 1000000 + dec 6). No signatures/blockhash/broadcast.

Comment-only on my own PR — no merge.

NeOMakinG (approved):
APPROVE. `buildSplTransfer` is a correct, fail-closed, unsigned, pure-crypto SPL `transferChecked` builder. The ATA derivation, token-program selection, wire encoding, and fund-safety guards all check out against an independent canonical vector. I pushed three follow-up fixes to your branch (RN export wiring, missing changeset, Token-2022 doc-hardening) — all green locally.

## testing approach

3-lane fan-out: Claude (load-bearing correctness/ATA/fund-safety), Codex gpt-5.3-codex (adversarial second-opinion on contract-shape + consumer coupling), GLM (dead — ECONNRESET, best-effort skipped).

Verified in worktree `/tmp/wt-sdk-824` (HEAD now `3d560e18`):
- **ATA derivation — independent canonical cross-check (the critical Solana check):**
  - `getAssociatedTokenAddressSync(USDC, FROM, false, TOKEN_PROGRAM_ID)` → `C4PRXFV6Gf5mytVZb6RoeLsG8CjcFWzR2EJ3dvwPTUJH` (matches receipt)
  - recipient ATA → `FGETo8T8wMcN2wCjav8VK6eh3dLk63evNDPxzLSJra8B` (matches receipt)
  - independent manual `findProgramAddressSync([owner, TOKEN_PROGRAM_ID, mint], ATokenGPv…)` → byte-identical to the lib derivation. Seeds + ATA program id are canonical.
  - program ids correct: legacy `Tokenkeg…`, Token-2022 `TokenzQd…`.
- **transferChecked wire shape:** discriminator `12`, LE u64 amount, decimals byte; data `DEBCDwAAAAAABg==` byte-exact to receipt. Account-meta order = `[source(w), mint(ro), dest(w), owner(signer)]` — canonical; owner is the sole signer.
- **Mutation tests (4/4 killed — tests have teeth):** invert program selection → 2 fail; force `decimals=0` in ix → 1 fail; neuter the u64-max guard → 1 fail; neuter the `to===mint` guard → 1 fail.
- **Gates (local, on fix SHA — gomes fix-commits historically don't re-trigger CI):** unit `8/8` (inline vitest config aliasing `@`→src, `setupFiles:[]` to dodge the `@vultisig/mpc-types` wall); tsc 0-new (only pre-existing mpc-types TS2307 remain); `quality:complexity` exit 0; `format:check` (CI `.config/.prettierrc`) clean; `check:shared-exports` clean; receipt byte-exact. CI on prior SHA `dc902db1`: all green.

## blocking

None.

## preferably-blocking

None.

## should-fix

All three resolved in-code on your branch (`3d560e18`):

1. **RN entry was missing `buildSplTransfer`** (recurring prep-export gap — #831/#827/#828/#854/#856/#857/#859/#830). Every sibling prep helper is exposed in `src/platforms/react-native/index.ts`; this one wasn't, so RN consumers (vultiagent-app, agent-backend) couldn't reach it. Fixed with a **lazy async wrapper** rather than a sync re-export, because `splTransfer.ts` statically pulls `@solana/web3.js`, which reads `globalThis.Buffer` at module-init — a sync `export { … } from` would drag it into the eager RN bundle graph (the exact footgun the `getSplAccounts` / `getSplAssociatedAccount` overrides exist to avoid). Plus `BuildSplTransferParams` / `SplTransferResult` type re-exports.

2. **Missing changeset** (recurring — #833/#854/#856/#859). The bundle-guard exempts `packages/sdk/src/**`, so this is green-but-unpublishable (release-pr.yml only versions when a changeset exists). Added `@vultisig/sdk: minor`.

3. **`isToken2022` fund-safety doc (Codex-flagged, source-confirmed).** The flag is caller-supplied with no auto-detect; passing `false` for a Token-2022 mint derives the wrong ATA + routes through the wrong program → failed tx or a credit to an unintended account. The code is correct (it does exactly what the flag says); the risk is consumer integration. Hardened the JSDoc to require the caller resolve the flag from the mint owner on-chain. (Not blocking: no in-repo consumer — this is an escape-hatch with zero current callers.)

## q

- **Create-ATA / consumer-coupling (Codex raised as blocker — downgraded after reading the signer).** Codex noted that the SDK signer `getSolanaSendSigningInput` (`packages/core/mpc/keysign/signingInputs/resolvers/solana/send.ts:62`) chooses `CreateAndTransferToken` vs plain `TokenTransfer` by whether `toTokenAssociatedAddress` is *absent* — so a future mcp-ts port that maps `toTokenAccount` → `toTokenAssociatedAddress` would suppress create-ATA and fail for unfunded recipients. This is **not a defect in this PR**: `buildSplTransfer` returns its own shape (`instruction` + ATA strings), is not consumed by that signer, has no in-repo caller, and its JSDoc already states the signer is responsible for the idempotent create-ATA. Worth keeping in mind when you wire the mcp-ts/agent-backend port — `toTokenAccount` is informational; the consumer should let the on-device signer derive/create the recipient ATA, not pass `toTokenAccount` straight through as an existing ATA.

## risk

Low. Pure-crypto, additive, unsigned, behind a new export, zero in-repo consumers. Never signs, never broadcasts. ATA derivation verified against an independent canonical vector; fund-safety guards mutation-tested.

## verdict

**APPROVE** @ `3d560e18`. ATA derivation canonical (independent vector, correct token program), transferChecked + decimals correct, recipient-as-wallet-owner (not ATA) confirmed, `to===mint` trap guarded, u64-wrap guarded, fail-closed, unsigned. RN export + changeset + Token-2022 doc fixed in-code. All gates green locally; CI green on the prior SHA.

---
🤖 reviewed by Claude (load-bearing) + Codex gpt-5.3-codex (adversarial). GLM lane dead (ECONNRESET), best-effort skipped.

PR #825, opened 2026-06-23
feat(sdk): add sdk.prep.trc20Transfer (tron trc-20 transfer)

gomesalexandre:
## what

New pure-crypto primitive `sdk.prep.trc20Transfer` (`prepareTrc20TransferFromKeys`) that builds an **unsigned** TRON TRC-20 token transfer. It ABI-encodes `transfer(address,uint256)` and returns an `UnsignedTrc20Transfer` descriptor (owner / contract / recipient / amount / `function_selector` / 128-char `parameter` calldata / `feeLimitSun` / optional memo).

Ported from the mcp-ts `build_trc20_transfer` tool + `lib/tron-abi.ts`, so the SDK becomes the single source of truth for TRC-20 calldata.

- `packages/sdk/src/abi/tron.ts` — base58check <-> hex address helpers + `encodeTrc20TransferParam`
- `packages/sdk/src/tools/prep/trc20.ts` — `prepareTrc20TransferFromKeys` (vault-free, no RPC)
- exported from the package barrel (`@vultisig/sdk`)
- 13 unit tests + a runnable receipt at `scripts/receipts/prep_trc20_transfer.mjs`

## why

Part of the mcp-ts/backend → SDK code-as-action consolidation. TRC-20 calldata encoding currently lives in mcp-ts; pulling it into the SDK lets mcp-ts / agent-backend / Windows / Station all consume one reviewed implementation instead of re-porting the base58check + ABI math per surface.

## fund-safety

- **Pure crypto only.** No RPC, no signing, no broadcast. The `parameter` blob is a deterministic function of `(to, amount)` alone; the on-device signer fills the live chain-specific fields (block header, ref-block, expiration, gas) and produces the final `TransferTRC20Contract`. `vault.sign` stays on-device.
- **Checksum-verified address decode** is the fund-safety gate: a typoed-but-still-base58-decodable address throws (`checksum mismatch`) instead of silently re-checksumming into a *different* valid address and misrouting funds. Contract / from / to are all validated up-front.
- `uint256` range guard on the amount (rejects negative and `>= 2^256`); zero-amount rejected.
- Test asserts the descriptor carries **no** signing material (`privateKey`/`signature`/`signed`/`rawTx`/`broadcast` absent).

## consumers unblocked

mcp-ts `build_trc20_transfer` (drop the local `tron-abi.ts` port), agent-backend code-as-action TRC-20 sends, vultisig-windows / Station TRC-20 transfer flows.

## receipt

```
$ node --import tsx scripts/receipts/prep_trc20_transfer.mjs

=== sdk.prep.trc20Transfer — UNSIGNED TRC-20 transfer ===

inputs:
{
  "contractAddress": "TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t",
  "from": "TJRabPrwbZy45sbavfcjinPJC18kjpRTv8",
  "to": "TUEZSdKsoDHQMeZwihtdoBiN46zxhGWYdH",
  "amount": "1000000",
  "memo": "SWAP:BTC.BTC:bc1qexampledestination:0"
}

unsigned tx descriptor:
{
  "chain": "Tron",
  "action": "transfer",
  "signingMode": "ecdsa_secp256k1",
  "ownerAddress": "TJRabPrwbZy45sbavfcjinPJC18kjpRTv8",
  "contractAddress": "TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t",
  "toAddress": "TUEZSdKsoDHQMeZwihtdoBiN46zxhGWYdH",
  "functionSelector": "transfer(address,uint256)",
  "parameter": "000000000000000000000000c8599111f29c1e1e061265b4af93ea1f274ad78a00000000000000000000000000000000000000000000000000000000000f4240",
  "feeLimitSun": "100000000",
  "amount": "1000000",
  "memo": "SWAP:BTC.BTC:bc1qexampledestination:0"
}

decoded ABI params (transfer(address,uint256)):
  selector string : transfer(address,uint256)
  recipient (word): 000000000000000000000000c8599111f29c1e1e061265b4af93ea1f274ad78a
  recipient (b58) : TUEZSdKsoDHQMeZwihtdoBiN46zxhGWYdH  OK matches `to`
  amount    (word): 00000000000000000000000000000000000000000000000000000000000f4240
  amount    (dec) : 1000000  OK matches `amount`

NO signing material emitted (keys): chain, action, signingMode, ownerAddress, contractAddress, toAddress, functionSelector, parameter, feeLimitSun, amount, memo
PASS — unsigned, deterministic, round-trips. Not signed, not broadcast.
```

## test matrix

- `yarn workspace @vultisig/sdk typecheck` → exit 0, **0 errors** (0 new)
- `yarn workspace @vultisig/sdk test` → **1526 passed** (13 new trc20 tests + 1 public-export assertion)
- receipt above → builds unsigned USDT transfer, calldata round-trips, no signing material

## Release Notes

* **New Features**
  * Added TRON (TRC-20) token transfer preparation without vault requirement
  * Added TRON address format conversion utilities for interoperability
  * Added calldata encoding for TRC-20 transfer operations

* **Tests**
  * Added comprehensive test suite for TRON operations including address conversion and transfer validation

* **Documentation**
  * Added example demonstrating TRON TRC-20 transfer preparation


gomesalexandre (commented):
## deep adversarial pass — 1 fund-safety hardening found + fixed, rest clean

Went byte-for-byte against the mcp-ts source of truth (`src/lib/tron-abi.ts` + `build-other-send.ts::buildTrc20Transfer`) and probed "how does an attacker break this".

### found + fixed (d882f41e)
**MEDIUM — `amount` (and `feeLimitSun`) accepted non-decimal `BigInt`-parseable strings, raw string leaked into `tx.amount` (WYSIWYS drift)** — `packages/sdk/src/tools/prep/trc20.ts:95`
- `BigInt()` is far too permissive for a value-bearing field. Runtime probe showed `amount: "0x10"` → encoded `...010` (16 base units), `"0b1010"` → 10, `"0o17"` → 15, plus `"+1000"` and whitespace-padded `" 1000000 "` all silently accepted — and `tx.amount` echoed the **raw un-normalized string**, not what was encoded. A confirm/WYSIWYS layer bound to `tx.amount` would render `"0x10"` / `"0b1010"` / `" 1000000 "` while the calldata moves a different (or confusingly-represented) value.
- Fix: reject anything that isn't `^[0-9]+$` (plain decimal base units), and echo the **canonical** `amountBig.toString()` so displayed amount === encoded amount always (also strips leading-zero noise). Same plain-decimal + positive guard applied to `feeLimitSun` (it's the SUN energy ceiling = TRX the signer may burn; value-adjacent).
- +13 test cases (hex/bin/oct/sci/underscore/whitespace/`+`/empty rejected, canonical-echo, feeLimitSun guard). 28/28 pass.

### verified clean
- **base58check decode + checksum**: faithful port of mcp-ts (`@noble/hashes` sha256d vs node:crypto — same double-sha256). 25-byte length gate + `0x41` prefix gate + 4-byte checksum compare. A flipped last char throws `checksum mismatch`. **Strictly STRONGER than mcp-ts**, whose `tronAddressString()` zod gate is shape-only (`^T[1-9A-HJ-NP-Za-km-z]{33}$`, no checksum). No silent re-checksum/misroute path.
- **all 3 addresses validated**: `contractAddress`, `from`, and `to` (the last re-validated inside `encodeTrc20TransferParam`) — a typoed-but-decodable address surfaces as a throw, never a misroute.
- **`transfer(address,uint256)` selector**: emitted as the SIGNATURE STRING (client keccak-hashes it), matching mcp-ts + vultiagent-app `tronTx.ts` contract. Not pre-hashed.
- **param encode**: 32-byte left-padded recipient word || 32-byte uint256 amount word = 128 hex, no `0x`, no selector. uint256 overflow (`>= 2^256`) and negative rejected.
- **no signing material**: descriptor carries no privkey/signature/rawTx/broadcast fields (asserted). Pure calldata, no RPC, no vault identity — `vault.sign` stays on-device.
- **memo / feeLimitSun defaults**: memo maps to proto `data`, feeLimit defaults `100_000_000` SUN (100 TRX) matching mcp-ts.

### receipts (no RPC, no signing — pure crypto)
- `tsc --noEmit`: 0 errors
- `vitest run prepareTrc20TransferFromKeys.test.ts`: **28/28 pass**; `publicExports.test.ts`: 7/7
- receipt `scripts/receipts/prep_trc20_transfer.mjs`: USDT transfer round-trips — recipient word decodes back to `to` (OK), amount word decodes to `1000000` (OK), parameter = 128 hex. PASS, unsigned, deterministic, not broadcast.

Not merging (own PR, needs external stamp).

NeOMakinG (approved):
fix-and-approve. Clean, fund-safe TRON TRC-20 transfer calldata builder. This is the SEND-side sibling of #843 (the balance reader that shipped a naive `address.replace(/^T/,'41')` encoding bug) — and this PR does NOT repeat that mistake: it does a proper base58check decode with full double-SHA256 checksum verification before encoding the recipient. I verified the encoding against canonical vectors two independent ways, mutation-tested the address/amount/checksum guards (they have teeth), and ran the full local gate suite green. I pushed two follow-up fixes on your branch (RN export gap + missing changeset).

## testing approach

3-lane fan-out. Claude (load-bearing): full diff-walk of `abi/tron.ts` + `tools/prep/trc20.ts`, independent base58check decode of all three test vectors, mutation tests on the encoding/checksum, local gates. Codex `gpt-5.3-codex` and GLM `glm-4.6` both dispatched in parallel — Codex stalled at 0-byte (~10min, model-at-capacity), GLM died (ECONNRESET). Both best-effort, non-load-bearing this round.

Verified on my pushed HEAD `f4bebcd2`:
- **Recipient/contract/owner base58check encoding (the #843-class critical check)** — PROVEN correct. `tronBase58ToEvmHex` does base58 decode → assert 25-byte length → assert `0x41` prefix → recompute `sha256(sha256(payload))[:4]` and byte-compare against the trailing 4 checksum bytes → `slice(1,21)` for the 20 raw address bytes. Independently decoded the three test vectors: USDT `TR7NH…Lj6t`→`a614f803b6fd780986a42c78ec9c7f77e6ded13c`, `TUEZ…WYdH`→`c8599111f29c1e1e061265b4af93ea1f274ad78a` — both match the pinned test constants. Receipt round-trips the encoded recipient word back to the exact `to` base58.
- **Known-vector tests + mutation teeth (#228)** — the suite pins `to`→exact 32-byte hex word AND amount word. Mutation: `slice(1,21)`→`slice(0,20)` (wrong addr bytes) → 3 fails; `padStart`→`padEnd` (amount) → 1 fail; neuter checksum compare → 2 fails. Real teeth — exactly what would have caught #843.
- **amount/decimals** — `_value` is `BigInt`, `^[0-9]+$` guard rejects `0x10`/`0b1010`/`0o17`/`+1000`/whitespace/`1.5`/`1e6`/`abc`/empty (10 negative cases), `>0` enforced, uint256 range guard `>= 1n<<256n` rejects overflow, canonical decimal echoed into `tx.amount` so display === encoded. No float, no decimals fabrication (token base units passed by caller, correct for this layer).
- **fail-closed on bad checksum** — typoed-but-base58-decodable recipient → throws `checksum mismatch` before any calldata is built. Confirmed via test + mutation.
- **unsigned** — descriptor carries only `{chain, action, signingMode, ownerAddress, contractAddress, toAddress, functionSelector, parameter, feeLimitSun, amount, memo?}`. Explicit test asserts no `privateKey`/`signature`/`signed`/`rawTx`/`broadcast`. `functionSelector` is the SIGNATURE STRING (`transfer(address,uint256)`), not a pre-hashed selector — client hashes + concatenates; `parameter` is the 128-char two-word blob, no `call_value` (correct, 0 for TRC-20).
- **local gates (new HEAD f4bebcd2)** — trc20 vitest 28/28 single-fork; tsc 0 new errors (29 preexisting TS2307 from mpc-types/mpc-wasm); `quality:complexity` exit 0; lint exit 0; `format:check` clean; `changeset:check-sdk-bundle` pass; receipt PASS. CI on prior HEAD `d882f41e` was all-green (Lint+TypeCheck, Unit, Integration, Test Summary, CodeRabbit); fix-commits on gomes branches don't always re-trigger CI — gates verified locally.

## should-fix

(both fixed in-code on your branch @ `f4bebcd2`, no action needed)

1. RN export gap — `prepareTrc20TransferFromKeys` was exported from the package barrel but NOT from `src/platforms/react-native/index.ts`, so Station / vultisig-windows RN consumers (the named beneficiaries) couldn't reach it. Recurring gap (#831/#827/#828/#854/#856/#857/#859). Fixed: added a lazy-`import()` wrapper matching the other prep helpers + re-exported `PrepareTrc20TransferFromKeysParams`/`UnsignedTrc20Transfer` types. Pure-crypto (`@noble/hashes` only), RN-safe.
2. Missing changeset — no version bump → green-but-unpublishable (the `changeset:check-sdk-bundle` guard exempts `packages/sdk/src/**`, so CI stays green but `release-pr.yml` never bumps/publishes). Recurring (#833/#854/#856/#859). Fixed: added a `minor` changeset.

## q

- `feeLimitSun` defaults to 100 TRX (100M SUN), matching the mcp-ts builder. For a single TRC-20 `transfer` this is generous but harmless (it's a ceiling, only consumed energy is burned, and the on-device signer can still surface the actual cost). Fine as a default; flagging only so it's a conscious choice.

## risk

Low. Pure-crypto, unsigned, no RPC, no signing material. The one catastrophic failure mode for a transfer builder — a mis-encoded recipient sending funds to a wrong address — is closed by a proper checksum-verified base58check decode, proven against canonical vectors and protected by mutation-tested assertions. The on-device signer remains the sole place keys touch the tx.

## verdict

APPROVE. Recipient/contract/owner encoding proven correct against known vectors, amount/decimals correct + WYSIWYS-guarded, fail-closed on bad checksum, unsigned, RN export + changeset now wired (my commit), tests pin the address vector + negatives with real mutation teeth, CI green.

---
🤖 reviewed by Claude (load-bearing) + Codex gpt-5.3-codex (stalled, best-effort) + GLM glm-4.6 (dead, best-effort)

PR #829, opened 2026-06-23
feat(sdk): add sdk.swap.across (Across bridge quote)

gomesalexandre:
## what

Adds `sdk.swap.acrossQuote` — a read-only Across bridge quote primitive, ported from the mcp-ts `get_across_quote` tool (0 SDK imports) into the SDK. Exported from `@vultisig/sdk` as `acrossQuote` (+ `acrossSupportedChains`, and `AcrossChain` / `AcrossQuote` / `AcrossQuoteParams` types).

Pure crypto / read-only: fetches a live Across `suggested-fees` quote, pins + verifies the origin and destination SpokePool deployments (fail-closed on upstream schema drift), checksums + validates inputs, and rejects a 200 with no usable output amount. It NEVER builds calldata, signs, or broadcasts. Origin is pinned to Ethereum for the current factory slice.

## why

Consolidates the Across quote logic out of mcp-ts so any SDK consumer (agent-backend, mcp-ts, CLI, vultiagent-app) can fetch a verified bridge quote through one code-as-action path instead of re-implementing the SpokePool pinning + fail-closed checks per surface.

## consumers unblocked

- agent-backend / mcp-ts: drop the bespoke `get_across_quote` HTTP+verify glue, call `sdk.swap.acrossQuote()` directly
- vultiagent-app: bridge-quote previews without a vault (quote-only)

## test

- `yarn workspace @vultisig/sdk typecheck` — 0 errors (clean)
- 11 vitest unit cases (mocked fetch): happy path, recipient checksum, every input guard, both SpokePool fail-closed paths, no-output rejection, HTTP error surfacing
- live receipt against the real Across API (no broadcast)

```
$ node --import tsx scripts/receipts/swap_across.mjs
sdk.swap.across — live Across bridge quote (read-only)
route: Ethereum → Arbitrum, 1 USDC

provider:                across
action:                  quote_bridge
execution_status:        read_only_quote
source_chain_id:         1
destination_chain_id:    42161
input_amount:            1000000 (1 USDC)
output_amount:           995054 (0.995054 USDC)
relay_fee_total:         4946 (0.004946 USDC)
lp_fee_total:            0 (0 USDC)
estimated_fill_time_sec: 2
source_spoke_pool:       0x5c7BCd6E7De5423a257D81B442095A1a6ced35C5 (pinned + verified)
dest_spoke_pool:         0xe35e9842fceaCA96570B734083f4a58e8F7C5f2A (pinned + verified)
quote_id:                scqgc-1782226577780-f4a2cd09e7b4

OK — live quote fetched, SpokePools verified, no broadcast.
```

```
$ yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/tools/swap/acrossQuote.test.ts
 Test Files  1 passed (1)
      Tests  11 passed (11)
```


gomesalexandre (commented):
## deep adversarial review — across port vs mcp-ts source

Reviewed `acrossQuote.ts` against the mcp-ts `src/tools/swap/across.ts` source contract (+ `lib/dangerous-addresses.ts`). One real fund-safety guard was dropped; fixed in 54d8c474. Everything else lines up byte-for-byte.

### fixed (P1) — dropped burn-address guard
`acrossQuote.ts` (~L377, the `if (params.to)` block) checksummed `params.to` and did `search.set('recipient', recipient)` with **no** `assertSafeDestination`. mcp-ts `across.ts:212` runs `assertSafeDestination(destinationChain, recipient)` on the recipient before forwarding. A tested guard from the source contract was silently absent. Read-only quote today (`executionStatus: 'read_only_quote'`, no calldata/sign), so P1-correctness not P0 — but it belongs in place before any build/sign step is layered on the recipient.

Fix: ported a canonical `packages/sdk/src/utils/dangerousAddresses.ts` (none existed in the SDK) mirroring mcp-ts — shape-based EVM detection + all **three** EVM burn variants (`0x0…0`, `0x0…dead`, `0xdead…942069`), case-insensitive — and wired `acrossQuote` to it. Putting it in a shared util (vs inlining) means the cctp `build_*` call site (#831's drop) can reuse the same list so it can't drift per-tool.

Note: viem's `isAddress` (strict) accepts all-lowercase as valid, so the canonical lowercase burn addresses pass `requireChecksummedAddress` and DO reach the guard — i.e. the real attack vector is covered (verified in the receipt below). Bad-mixed-case variants get rejected earlier by `isAddress` anyway.

### verified clean (no changes)
- **API base** `https://app.across.to/api` — matches
- **chain ids** Ethereum:1, Optimism:10, Polygon:137, Arbitrum:42161, Base:8453, Blast:81457, BSC:56, Zksync:324 — all match mcp-ts `lookupEvmChainId`
- **all 8 pinned SpokePool addresses** — byte-identical to mcp-ts
- **supported-chain set** + Ethereum-origin pin — match
- **fail-closed source AND destination SpokePool checks** — present + identical (200 omitting `destinationSpokePoolAddress` correctly rejected, never overridden with the local pin)
- **`outputAmount` shape-check** (reject 200 with no usable output) — present
- **amount validation** (integer base-units, `> 0n`) — present
- **affiliate/fee leakage** — N/A: Across `suggested-fees` takes no affiliate/integrator param; neither mcp-ts nor the SDK set one. No leakage.

### receipts (run on the fixed branch)
```
route: Ethereum → Arbitrum, 1 USDC
output_amount:           995092 (0.995092 USDC)
source_spoke_pool:       0x5c7BCd6E7De5423a257D81B442095A1a6ced35C5 (pinned + verified)
dest_spoke_pool:         0xe35e9842fceaCA96570B734083f4a58e8F7C5f2A (pinned + verified)
OK — live quote fetched, SpokePools verified, no broadcast.

burn-guard check (recipient = 0xdead…942069):
  rejected as expected → Refusing to build transaction: destination 0xdEAD…942069 is a dead address variant — funds are unrecoverable.
OK — burn-address recipient rejected pre-flight.
```
`yarn workspace @vultisig/sdk typecheck` clean (0 errors); 21 unit tests green (added burn-guard cases to `acrossQuote.test.ts` + a dedicated `dangerousAddresses.test.ts`).

Own-PR comment, not an approval.

packages/sdk/src/tools/swap/acrossQuote.ts
```diff
@@ -0,0 +1,315 @@
...
+    sourceChain,
+    sourceChainId,
+    destinationChain,
+    destinationChainId,
+    inputToken: quote.inputToken ?? { address: inputToken, chainId: sourceChainId },
+    outputToken: quote.outputToken ?? { address: outputToken, chainId: destinationChainId },
+    inputAmount: amount,
+    outputAmount: quote.outputAmount,
+    estimatedFillTimeSec: quote.estimatedFillTimeSec,
+    isAmountTooLow: quote.isAmountTooLow ?? false,
```
NeOMakinG, line 295: `should-fix:` `isAmountTooLow` is passed through truthfully but the function still returns an actionable `outputAmount`, and `limits.minDeposit` is never enforced against `amount`. A 200 with `isAmountTooLow: true` (or `minDeposit > amount`) still emits a quote. Faithful to mcp-ts `across.ts:277` and the field is surfaced so the caller can branch — so not blocking on a read-only quote — but a public SDK quote would be safer rejecting (or flagging harder) when `isAmountTooLow === true`. Raise with the mcp-ts port-contract owner if you want both surfaces hardened together.

packages/sdk/src/tools/swap/acrossQuote.ts
```diff
@@ -0,0 +1,315 @@
...
+    quoteBlock: quote.quoteBlock,
+    quoteTimestamp: quote.timestamp,
+    fillDeadline: quote.fillDeadline,
+    spokePoolAddress: getAddress(acrossSpokePools[sourceChain]),
+    destinationSpokePoolAddress: getAddress(acrossSpokePools[destinationChain]),
+    fees: {
+      relayFeeTotal: quote.relayFeeTotal ?? quote.totalRelayFee?.total,
```
NeOMakinG, line 302: `suggestion:` `relayFeeTotal: quote.relayFeeTotal ?? quote.totalRelayFee?.total` uses `??`, which only rejects null/undefined — a malformed-but-200 `''` or `'abc'` legacy field is preferred over a valid nested value and surfaced verbatim. `outputAmount` is shape-checked (line 279) but the fee fields are not. Low blast (display metadata on a read-only quote, never signed) and mcp-ts-identical, so suggestion-tier: validate fee fields as base-unit integer strings for symmetry with the output-amount guard.

packages/sdk/src/utils/dangerousAddresses.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Throws a descriptive error if `destination` is a known dangerous address.
+ * Call this early in every quote/build tool handler, BEFORE any expensive work
+ * (price lookups, RPC calls, balance fetches) and BEFORE forwarding the
+ * recipient to an upstream API.
+ *
+ * `chain` is accepted to match the mcp-ts source contract and for future
+ * non-EVM family routing; EVM detection is shape-based and does not depend on
+ * it.
+ */
+export function assertSafeDestination(_chain: string, destination: string): void {
```
NeOMakinG, line 53: `q:` This file also lands in PR #831 (cctp) with a **divergent API**: #831 ships `assertSafeEvmDestination(address)` + exported `EVM_DANGEROUS_ADDRESSES`/`getEvmDangerousReason`/`isEvmBurnAddress`, whereas this PR ships `assertSafeDestination(_chain, destination)`. The burn-list is identical but the signatures are not, so whichever merges second hits a conflict and the loser's caller breaks. Merge-ordering hazard for whoever lands both — recommend converging on the #831 shape (it exports the table for reuse) and switching `acrossQuote` to `assertSafeEvmDestination(recipient)` on rebase.

NeOMakinG (approved):
## testing approach

3-lane fan-out (Claude load-bearing + Codex `gpt-5.3-codex` adversarial + GLM third-eye). Reviewed the **true** diff via merge-base `0a906414` (the branch is behind `main`, so `git diff main...HEAD` shows a phantom +6691/-1425 — the real change is the advertised 8 files +758/-1). GLM dead (429 insufficient-balance, best-effort skipped). All gates run **locally** in a worktree off HEAD (our fix-commits do not re-trigger CI on gomes branches), CI already green on the prior SHA.

- vitest: 21/21 pass (`acrossQuote.test.ts` 14 + `dangerousAddresses.test.ts` 7), single-thread
- tsc: 0 new errors (29 pre-existing `TS2307` mpc-types wall)
- complexity (sonarjs-35): exit 0; prettier (`.config/.prettierrc`): clean (incl. the changeset I added)
- live receipt (`scripts/receipts/swap_across.mjs`): real Ethereum->Arbitrum quote, output 995132 < input 1000000, fees decoded, **both SpokePools pinned+verified**, burn-recipient rejected pre-flight
- on-chain cross-check: the pinned Ethereum SpokePool `0x5c7B…35C5` and Arbitrum dest SpokePool `0xe35e…5f2A` match the **live** Across `suggested-fees` response exactly
- mutation tests (teeth): corrupt the source SpokePool pin -> 4 tests fail; neuter the burn-guard (`if(false)`) -> 8 tests fail
- mcp-ts source parity: read `mcp-ts/src/tools/swap/across.ts` — the SDK port is byte-faithful (and *improves* it: SDK throws where mcp-ts returns a soft `jsonError`)

## blocking

None.

## should-fix

**1. `isAmountTooLow: true` / `minDeposit > amount` still emits a quote** (`acrossQuote.ts:295`, Codex-flagged). When Across marks the route non-executable, `isAmountTooLow` is passed through truthfully but the function still returns an `outputAmount` a caller could treat as actionable; `limits.minDeposit` is never enforced against `amount`. This is **faithful to the mcp-ts source** (`across.ts:277` does the same pass-through, no fail-closed), and the field *is* surfaced so the consumer can branch on it — so it is not blocking on a read-only quote. But a public SDK quote tool would be safer rejecting (or at least flagging more loudly) when `isAmountTooLow === true`. Left as should-fix to preserve mcp-ts parity; raise with the port-contract owner if you want both surfaces hardened together.

## suggestion

**2. Fee normalization trusts the first non-null schema variant without shape-validation** (`acrossQuote.ts:302`, Codex-flagged). `relayFeeTotal: quote.relayFeeTotal ?? quote.totalRelayFee?.total` (and the lp/pct siblings) use `??`, which only rejects `null`/`undefined` — a malformed-but-200 `''` or `'abc'` legacy field would be preferred over a valid nested value and surfaced verbatim. `outputAmount` is shape-checked (`:279`) but fees are not. Low blast on a read-only quote (display metadata, never signed) and again **mcp-ts-identical**, so suggestion-tier: validate fee fields as base-unit integer strings if you want symmetry with the output-amount guard.

**3. `outputAmount` is not a slippage floor** (`acrossQuote.ts:279`, Codex-flagged). Accepted verbatim with only a digit check; no recompute, no min-output bound. Correct for a `read_only_quote` (and documented as such), but the field name invites consumers to mistake it for an execution bound — worth a doc note when the build/sign step lands.

## q

**4. `dangerousAddresses.ts` collides with PR #831 (cctp).** Both #829 and #831 add `packages/sdk/src/utils/dangerousAddresses.ts` (not on `main`), and the two copies have **divergent public APIs**: #829 ships `assertSafeDestination(_chain, destination)` with internal `EVM_DANGEROUS`; #831 ships `assertSafeEvmDestination(address)` + exported `EVM_DANGEROUS_ADDRESSES` / `getEvmDangerousReason` / `isEvmBurnAddress`. The burn-list (3 addresses) is identical, but the function signatures are not. **Whichever merges second will hit a merge conflict, and the loser's caller breaks** (`acrossQuote` calls `assertSafeDestination`; cctp calls `assertSafeEvmDestination`). This is a merge-ordering hazard, not a defect in this PR — flag for the human merging both: land one, then rebase the other onto the shared file and reconcile the API (recommend the #831 shape since it exports the table for reuse, then `acrossQuote` switches to `assertSafeEvmDestination(recipient)`).

## risk

Low. Read-only quote primitive (`producesCalldata` absent; never signs/broadcasts). Fund-relevant surface (fee/output informs a bridge decision + the recipient burn-guard) is verified: SpokePools pinned-and-fail-closed against live deployments, burn-guard wired on the optional `to` field and fail-closed pre-network, inputs checksummed, amount validated as positive base-unit BigInt, origin pinned to Ethereum. RN export correctly omitted (the sibling `findSwapQuote` is also not in the hand-curated RN allow-list — `tools/swap/*` convention).

## verdict

**APPROVE.** Quote/fee decoding is byte-faithful to the mcp-ts source and the SDK hardens it (throws vs soft-error). SpokePool pinning is fail-closed and verified against the live Across API; the burn-guard is correct, wired, and has mutation-test teeth; chain ids + tokens are canonical; tests prove the math, the guards, and the fail-closed paths; CI green. The only fund-relevant gaps (isAmountTooLow/minDeposit, fee shape-validation) are faithful-port should-fix/suggestion on a read-only quote, not blockers. I added the missing changeset in-code (`a2691a5c`) so the new export actually publishes — the changeset bundle-guard exempts `packages/sdk/src/**`, so CI was green-but-unpublishable without it. Noted the #831 `dangerousAddresses.ts` divergent-duplicate as a merge-ordering hazard for whoever merges both.

---
*Reviewed by Claude (load-bearing) + Codex gpt-5.3-codex (adversarial, converged). GLM lane skipped (Z.ai 429). Fix-and-approve per gomes carve-out: changeset added + pushed, faithful-port findings surfaced not modified.*

PR #831, opened 2026-06-23
feat(sdk): add sdk.bridge.cctp (Circle CCTP bridge + claim calldata)

gomesalexandre:
Adds `sdk.bridge.cctp` — Circle CCTP USDC bridge + claim calldata builders. Pure crypto: builds unsigned calldata only, NEVER signs or broadcasts.

## what

Two new vault-free primitives + the CCTP contract registry, exported from the package index:

- `buildCctpBridge({ sourceChain, destinationChain, amount, to?, from? })` → unsigned 2-tx sequence: ERC-20 `approve(USDC → TokenMessenger)` then `depositForBurn(...)`. Zero slippage (USDC burned 1:1 source, minted 1:1 dest).
- `buildCctpClaim({ destinationChain, message, attestation })` → single unsigned `receiveMessage(message, attestation)` tx on the destination MessageTransmitter (mints the bridged USDC).
- Registry: `cctpChains` / `getCctpChain` / `cctpSupportedChains` + `cctpAttestationApiBase` (live Circle V1 attestation endpoint). Carries the decimal EVM `chainId` alongside the CCTP `domain` so consumers don't re-resolve.
- Helpers exported for callers/tests: `parseUsdcAmount`, `formatUsdc`, `normalizeHexBytes`.

Supported chains: Ethereum, Avalanche, Optimism, Arbitrum, Base, Polygon.

## why

Ports `build_cctp_bridge_usdc` + `build_cctp_claim_usdc` + `lib/cctp.ts` out of mcp-ts into the SDK as the canonical pure-crypto layer. Part of the mcp-ts/backend → SDK code-as-action consolidation: the SDK owns the registry + calldata math, the tool wrappers become thin shells.

**Consumers unblocked:** mcp-ts (drop its hand-rolled CCTP encoders, call the SDK), agent-backend, vultisig-windows extension, Station — any client that wants cross-chain USDC without standing up its own CCTP registry.

## fund-safety

- Bridge refuses a burn-address `mintRecipient` (`0x0` / `0x…dEaD`) — CCTP's TokenMessenger does NOT reject these on-chain, so without the guard bridged USDC mints to a permanently unspendable address.
- Claim rejects any attestation whose byte length is not a non-zero multiple of 65 (Circle attestations are `n*65`; anything else is malformed and produces a guaranteed-revert claim tx). The multiple-of-65 invariant is version-stable across CCTP V1/V2.

## receipt

[...]

PR #833, opened 2026-06-23
feat(sdk): add sdk.swap.skip (Skip cross-chain route + prep)

gomesalexandre:
## what

Brings the Skip Go cross-chain route + unsigned-tx prep into the SDK as a pure-crypto primitive: `runSkipSwap()` / `quoteSkipRoute()`. It QUOTES a Skip route (`/v2/fungible/route`) and BUILDS the unsigned transaction envelope (`/v2/fungible/msgs_direct` -> `unsigned_msgs`), returning the unsigned EVM/cosmos tx payloads for the caller's signing layer. It NEVER signs and NEVER broadcasts.

Ported from mcp-ts's `runSkipSwap` (`src/tools/swap/skip-swap.ts`), which had 0 SDK imports. The two helper deps it touched (`skipChainIdToChainName`, `assertNotValidatorHrp`) + the affiliate config are inlined into the SDK module. The mcp-ts `skip-prep` wrapper (ExecutePrepResult/stepper/card shaping) is app-layer, not crypto, so it stays out of the SDK.

## why

Skip covers cross-chain corridors THORChain / Rujira / 1inch / LiFi don't carry (Terra v2 / Terra Classic IBC + Osmosis poolmanager swaps). Having the route+envelope builder in the SDK means `execute_swap`'s Terra/Cosmos branch + `list_swap_routes` can call `sdk.swap.skip()` instead of re-importing the mcp-ts builder, and any future client (CLI, app) inherits the fund-safety guards for free.

All fund-safety guards from the mcp-ts source are preserved at BOTH the `/route` and `/msgs_direct` boundaries:
- multi-signature route rejection (opt-in via `allowMultiTx` only)
- unsupported custody-chain stranding guard — reuses the SDK's own `getCosmosChainByChainId` + `getEvmChainByChainId` as the source of truth for "Vultisig can derive a key here" (no hardcoded list)
- LUNC/USTC notional floor + adaptive slippage + swap-hop budget (shallow Osmosis pool depth)
- cosmos source-leg memo-size preflight (`x/auth.MaxMemoCharacters`)
- route-vs-envelope chain-path / `does_swap` / per-tx chain-id cross-checks
- validator-HRP (`valoper`/`valcons`) recipient rejection

## consumers unblocked

- mcp-ts `execute_swap` Terra/Cosmos branch + `list_swap_routes` -> call `sdk.swap.skip()` instead of the local builder
- agent-backend / CLI -> Skip cross-chain quotes + unsigned prep without re-implementing the guards

## test matrix

12 unit tests (mocked `fetch`, no network): chain-mapping, floor resolution, affiliate build, address-shape + valoper + amount + slippage validation, happy-path envelope, multi-tx rejection, unsupported-custody-chain rejection, no-route 200->404, quote-only path. Plus a LIVE receipt against the real Skip API (OSMO -> ATOM), quote+build-unsigned only, no broadcast.

```
$ node --import tsx scripts/receipts/swap_skip.mjs
=== sdk.swap.skip — LIVE Skip Go route + unsigned tx prep ===
request: {
  "fromAddress": "osmo1clpqr4nrk4khgkxj78fcwwh6dl3uw4epasmvnj",
  "toAddress": "cosmos1a8l3srqyk5krvzhkt7cyzy52yxcght6322w2qy",
  "sourceChainId": "osmosis-1",
  "sourceAssetDenom": "uosmo",
  "destChainId": "cosmoshub-4",
  "destAssetDenom": "uatom",
  "amountIn": "5000000"
}

outcome: OK
route_description : osmosis-1 → cosmoshub-4 via osmosis-poolmanager
amount_in         : 5000000 uosmo
expected_amount_out: 114672 uatom
min_amount_out    : 114672 uatom
slippage_bps      : 100
usd_amount_in     : 0.20
usd_amount_out    : 0.20
swap_venue        : osmosis-poolmanager
price_impact_%    : 0.9989
multi_tx          : false (tx_count=1)
skip_chain_path   : osmosis-1 -> cosmoshub-4
settlement_seconds: 30

unsigned_msgs (NOT signed, NOT broadcast):
  [0] chain_id=osmosis-1 signing_method=cosmos
      msgs=/cosmwasm.wasm.v1.MsgExecuteContract

relay-safe: this script quotes + builds-unsigned only. No keys, no signing, no broadcast.
```

```
$ yarn workspace @vultisig/sdk vitest run tests/unit/tools/swap/skipSwap.test.ts
 Test Files  1 passed (1)
      Tests  12 passed (12)
```

typecheck: clean (0 new errors).

## risk

Low. Additive — new module under `tools/swap/skip/`, exported from the package index. No existing surface touched. Quotes/builds-unsigned only; the signing/broadcast path is the caller's (vault/MPC) responsibility and is untouched.

## Summary by CodeRabbit

* **New Features**
  * Added Skip Go cross-chain swap capability to prepare unsigned transaction envelopes for quoting and building cross-chain routes.
  * Introduced affiliate configuration system for swap flows with per-chain fee routing.
  * Added comprehensive safety validations including address format checks, fund protection guards, and transaction envelope integrity verification.

* **Bug Fixes**
  * Corrected dYdX canonical chain ID from `dydx-1` to `dydx-mainnet-1`.

* **Tests**
  * Added unit test suite covering cross-chain swap flows, chain mapping, and safety guard validations.


gomesalexandre (commented):
## deep adversarial review — swap.skip port vs mcp-ts/Go source

Pushed the audit fix in db144e3a plus a regression test + receipt leg. Reviewed the full 1114-line `skipSwap.ts` port against `~/Sites/vultisig-mcp-ts/src/tools/swap/skip-swap.ts` + `skip-chain-mapping.ts` + `skipAffiliateConfig.ts`.

### fixed (P2) — dYdX custody over-rejection
`packages/core/chain/chains/cosmos/chainInfo.ts:9` mapped `CosmosChain.Dydx` -> `'dydx-1'`, but Skip (and mcp-ts `COSMOS_CHAINS`) publish dYdX mainnet as `dydx-mainnet-1`. `chainMapping.ts` resolves custody via `getCosmosChainByChainId`, so `skipChainIdToChainName('dydx-mainnet-1')` returned `undefined` -> `firstUnsupportedCustodyChain` flagged it -> any Skip route custodying on dYdX was wrongly rejected with `skip_unsupported_route_chain`. The vault CAN derive a dYdX key, so the route is recoverable and must NOT be rejected. Pre-existing SDK constant bug (also on `main`) surfaced by the port. Fails-closed (over-rejects, never strands) hence P2. Fixed the constant; **every cosmos chain id now matches mcp-ts byte-for-byte** (thorchain-1, cosmoshub-4, osmosis-1, mayachain-mainnet-v1, dydx-mainnet-1, kaiyo-1, phoenix-1, columbus-5, noble-1, akashnet-2). `getCosmosChainId` (forward) has zero SDK consumers, so no regression surface.

### verified clean — no other dropped guards
Adversarially diffed every fund-safety surface; all present + faithful on BOTH legs (`/route` + `/msgs_direct`):
- **affiliate preservation** (`affiliateConfig.ts`): treasury addresses byte-identical to mcp-ts `skipAffiliateConfig.ts` (osmo `osmo18ggw7…`, terra `terra1lvhuq…` on phoenix-1+columbus-5, EVM `0x649E…076D` on the 7 chains). `buildSkipAffiliates` fail-closed logic (NaN/negative/>10000/non-integer -> omit) matches. No affiliate leakage.
- **multi-tx rejection** (`txs_required > 1` AND `txs.length > 1`, both legs) — `skip_multi_tx_route_rejected` unless `allowMultiTx` opted in.
- **custody stranding guard** (`firstUnsupportedCustodyChain`, both legs) — identical set logic.
- **LUNC/USTC**: notional floor `$0.05` default + env override, swap-hop budget (cosmos=1/EVM=3), adaptive slippage (`DRIFT_BASE=1.0`, `MAX=15.0`, ceiling-reject) — all constants match.
- **memo cap** (`COSMOS_MEMO_MAX_BYTES_BY_CHAIN_ID`): 6 entries @ 256 bytes, identical map.
- **valoper/valcons rejection**, HRP-exact validation, `MAX_SLIPPAGE_PERCENT=5.0` clamp — all present.

### receipts
- `yarn workspace @vultisig/sdk typecheck` — 0 errors
- `vitest tests/unit/tools/swap/skipSwap.test.ts` — 13/13 pass (incl. 2 new dydx assertions: `dydx-mainnet-1`->`Dydx`, `dydx-1`->`undefined`)
- `node --import tsx scripts/receipts/swap_skip.mjs` — LIVE Skip route, exit 0; dYdX leg now resolves (`rejected_for_dydx_custody: false`; pre-fix this was `skip_unsupported_route_chain`)

No P0/P1 found in this port. Single P2 constant fix applied.

packages/sdk/src/tools/swap/skip/skipSwap.ts
```diff
@@ -0,0 +1,1114 @@
...
+    return fail({
+      error: 'skip_msgs_direct_no_txs',
+      message:
+        (msgs as { message?: string }).message ??
+        'skip /msgs_direct returned 200 with no txs — route may have lapsed between /route and /msgs_direct',
+      skip_message: (msgs as { message?: string }).message ?? null,
+      skip_code: (msgs as { code?: number }).code ?? null,
+    })
+  }
+  if (!msgs.min_amount_out) {
```
NeOMakinG, line 876: `should-fix:` `min_amount_out` is presence-only — `if (!msgs.min_amount_out)` rejects `''`/missing, but `'0'` is truthy in JS, so an envelope with `min_amount_out: '0'` (zero slippage floor → ~0 receipt acceptable) passes this guard and gets signed. Parse to BigInt and require `> 0`.

packages/sdk/src/tools/swap/skip/skipSwap.ts
```diff
@@ -0,0 +1,1114 @@
...
+        swap_price_impact_percent: msgs.route.swap_price_impact_percent ?? null,
+        source_chain_id: args.sourceChainId,
+        dest_chain_id: args.destChainId,
+      })
+    }
+  }
+
+  // ── per-tx envelope validation against the canonical chain path ──
+  const canonicalChains = new Set(msgsChainPath)
+  for (let i = 0; i < msgs.txs.length; i++) {
```
NeOMakinG: `should-fix:` Skip `/msgs_direct` trust boundary — the per-tx loop only checks tx-kind exclusivity + on-path `chain_id`; `cosmos_tx.{msgs,memo}` and `evm_tx.{to,value,data,required_erc20_approvals[].spender}` are forwarded verbatim/unparsed by `shapeUnsignedMsgs`. A compromised Skip response can return an EVM approval to an attacker spender or a cosmos `MsgSend` to an attacker on a valid chain-path and it passes every guard. Mitigated by unsigned output + downstream signing/policy gate, and matches the mcp-ts source (not a new regression) — but recommend a follow-up that whitelists `msg_type_url`, asserts decoded recipient/denom/signer match `chain_ids_to_addresses`, and constrains EVM `to`/spender to known routers. (Codex converged here.)

packages/sdk/src/tools/swap/skip/skipSwap.ts
```diff
@@ -0,0 +1,1114 @@
...
+    (args.destChainId === COLUMBUS_5 && args.destAssetDenom === ustcDenom)
+  )
+}
+
+/**
+ * Expected bech32 HRP per supported cosmos chain id. Without this the generic
+ * `^[a-z]+1...` regex would accept `terra1...` for `cosmoshub-4` (and vice
+ * versa) — the user signs with a vault-derived key on the wrong chain.
+ */
+const COSMOS_CHAIN_HRPS: Record<string, string> = {
```
NeOMakinG, line 345: `should-fix:` `COSMOS_CHAIN_HRPS` is incomplete relative to the custody-chain set. `skipChainIdToChainName` resolves `akashnet-2` (custody guard passes it), but there's no `akashnet-2` entry here, so an Akash leg gets only the generic bech32 check and skips the chain-binding HRP check — a `cosmos1...` recipient would be accepted for an Akash destination. Add `'akashnet-2': 'akash'`, and ideally derive this table from the chain registry so it can't drift out of sync with the chains the custody guard accepts.

packages/sdk/src/tools/swap/skip/skipSwap.ts
```diff
@@ -0,0 +1,1114 @@
...
+          let innerPoolCount = 0
+          for (const innerKey of Object.keys(inner)) {
+            const innerVal = (inner as Record<string, unknown>)[innerKey]
+            if (
+              innerVal != null &&
+              typeof innerVal === 'object' &&
+              'swap_operations' in innerVal &&
+              Array.isArray((innerVal as Record<string, unknown>).swap_operations)
+            ) {
+              innerPoolCount = Math.max(
```
NeOMakinG, line 313: `should-fix:` `countSwapOperations` takes `Math.max` of nested branch `swap_operations` lengths instead of summing across branches. A split smart-swap with two thin-pool branches reads as fewer hops than it is, under-counting against the LUNC hop budget the stranding guard depends on. Sum across branches, or fail-closed on `smart_swap_*` for LUNC routes.

packages/sdk/src/tools/swap/skip/skipSwap.ts
```diff
@@ -0,0 +1,1114 @@
...
+  'thorchain-1': 'thor',
+  'mayachain-mainnet-v1': 'maya',
+  'dydx-mainnet-1': 'dydx',
+  'stride-1': 'stride',
+  'agoric-3': 'agoric',
+  celestia: 'celestia',
+  'injective-1': 'inj',
+}
+
+function validateAddressShape(address: string, chainId: string, field: string): void {
```
NeOMakinG, line 362: `suggestion:` shape+HRP only, no bech32 checksum — a bad-checksum string passes local validation (Skip/chain rejects downstream → fail-closed, low risk). Matches mcp-ts; noting for completeness.

NeOMakinG (approved):
Reviewed `db144e3a`; pushed two fixes on top as `df61bd6e` (changeset + affiliate-receiver test pin). Deep three-lane review: Claude load-bearing (full path-walk + 6 mutation probes + dydx core-change consumer audit + mcp-ts parity check), Codex `gpt-5.3-codex` adversarial (converged hard on the trust boundary), GLM dead (ECONNRESET — best-effort, skipped). This is a faithful SDK port of mcp-ts `runSkipSwap`; the documented fund-safety guards (multi-tx reject, custody-chain stranding, LUNC/USTC floor + adaptive slippage + hop budget, memo preflight, valoper reject, route↔envelope cross-checks) are all present and have mutation-tested teeth.

## testing approach
- Worktree on `feat_sdk_swap_skip`, real `yarn install` (symlink node_modules breaks vitest), minimal inline vitest config (aliased `@vultisig/core-config` → its `index.ts` to dodge the unbuilt-workspace WASM wall), single-fork pool.
- `skipSwap.test.ts`: 13/13 green on source; 14/14 after my added affiliate-pin test.
- Mutation probes (teeth check): valoper-guard neuter → 1 fail ✓ ; multi-tx guard widen → 1 fail ✓ ; cosmos HRP `cosmos`→`terra` → 4 fail ✓ ; **affiliate EVM receiver → attacker addr → 0 fail (GAP, fixed)** then 1 fail after pin ✓.
- dydx core change: `getCosmosChainId(Dydx)` only feeds the cosmos sign-doc `chain_id`; RPC URLs are keyed by the `CosmosChain` enum not the id-string, so blast-radius is the sign-doc only. `dydx-mainnet-1` is the real on-chain id; `dydx-1` was always wrong → this is a correctness fix.
- mcp-ts parity: confirmed `src/tools/swap/skip-swap.ts` has the identical post-`/msgs_direct` behavior (no msg/EVM-calldata re-validation) — the trust-boundary findings below are inherited properties of the production integration, not regressions this PR introduces.
- CI: `gh pr checks` showed **Lint and Type Check: FAIL** + **Test Summary: FAIL** at `db144e3a` — root cause = **SDK release guard: no changeset bumps `@vultisig/sdk`** (fixed in `df61bd6e`).

## blocking
- **CI release guard — missing changeset (FIXED in `df61bd6e`).** The static-quality-gate step failed: "This PR changes source files that are bundled into `@vultisig/sdk`, but no changeset bumps `@vultisig/sdk`." Added `.changeset/sdk-swap-skip.md` (`minor`). Without it consumers never get a new tarball and the merge gate stays red.

## should-fix
- **Skip `/msgs_direct` trust boundary — the signable envelope is never re-validated against the requested recipient/denom/calldata.** After `/msgs_direct` the per-tx loop only checks (a) exactly one of `evm_tx`/`cosmos_tx` and (b) `chain_id` is on-path. Then `shapeUnsignedMsgs` forwards `evm_tx.{to,value,data}` + every `required_erc20_approvals[].spender`, and `cosmos_tx.{msgs,memo}`, **verbatim and unparsed**. A compromised/MITM Skip response can therefore return an EVM approval to an attacker spender / an arbitrary contract call, or a cosmos `MsgSend`/`MsgExecuteContract` to an attacker — all on a valid chain-path — and it passes every guard. The mitigants are real (output is UNSIGNED; the vault/MPC signing layer + downstream policy-decode is the actual signing-authority gate) and this matches the mcp-ts source exactly, so it is **not a new regression and not on-its-own merge-blocking** — but it is the single highest-leverage hardening this module is missing. Recommend a follow-up: whitelist `cosmos_tx.msgs[].msg_type_url`, assert the decoded recipient/denom/signer match `chain_ids_to_addresses`, and constrain EVM `to`/spender to known per-chain routers. (Codex converged here independently.)
- **`min_amount_out` is presence-only — `'0'` passes the slippage-floor guard.** `if (!msgs.min_amount_out)` rejects `''`/missing but `'0'` is truthy, so a `min_amount_out: '0'` envelope (zero floor → ~0 receipt acceptable) is signed. Parse to BigInt and require `> 0`.
- **`COSMOS_CHAIN_HRPS` is incomplete vs the custody-chain set → wrong-chain recipient footgun for Akash.** `skipChainIdToChainName` resolves `akashnet-2` (so the custody guard passes it), but `COSMOS_CHAIN_HRPS` has no `akashnet-2` entry, so an Akash leg gets only the generic `^[a-z]+1...` bech32 check and **skips the chain-binding HRP check** — a `cosmos1...` recipient would be accepted for an Akash destination. Add `akashnet-2: 'akash'` (and ideally derive the HRP table from the chain registry so it can't drift out of sync with the custody set).
- **`countSwapOperations` takes `max` of nested branches instead of `sum`.** A split smart-swap with two thin-pool branches reads as fewer hops than it really is, under-counting against the LUNC hop budget the stranding guard relies on. Sum nested `swap_operations` across branches (or fail-closed on `smart_swap_*` for LUNC routes).

## suggestion
- `validateAddressShape` does shape+HRP only, no bech32 checksum — a bad-checksum string passes local validation (Skip/chain will reject downstream, fail-closed, so low risk). Matches mcp-ts.
- `runSkipSwap` cognitive complexity is 77 vs the health-report's 35 (not a blocking CI gate in this repo's `eslint.config.mjs`). The guards read sequentially, but extracting the post-`/msgs_direct` validation block into a helper would help.

## q
- The diff vs the PR's merge-base shows `src/index.ts`/`tools/index.ts` "removing" `knownContracts`/`decode*`/`policy`/`checkInvariants` exports — this is a **phantom of the stale diff base** (branch is 7 commits behind `origin/main`; those exports were added in `#846`/`#847`/`#848` *after* the merge-base, so the merge-base also lacks them). The branch's actual index change is purely additive (Skip exports). Confirm you **merge** (not rebase-drop) so the newer exports survive — a `git merge origin/main` will reconcile both sets cleanly.

## risk
Low-additive as authored: new module under `tools/swap/skip/`, unsigned output only, all documented guards preserved with teeth. The residual risk is the inherited trust-boundary (untrusted Skip envelope passes through unparsed) — acceptable for an unsigned-prep primitive whose consumers re-gate at the signing layer, but the highest-value follow-up.

## verdict
APPROVE. The one real CI blocker (missing changeset) is fixed; the trust-boundary is a faithful-port inherited property with real mitigants (unsigned output + downstream signing gate), surfaced as the headline follow-up. Chain mapping correct, dydx core change correct + low blast-radius, affiliate config correct (treasury addresses pinned + bps fail-closed), valoper/HRP guards wired and fail-closed, recipients/amounts validated up-front, multi-tx + custody-chain + LUNC guards mirrored on both API boundaries. RN export note: the prep/builders here are async/network (not the pure-crypto `prepare*` family), so the systematic `react-native/index.ts` static-export gap from #831/#827/#828 doesn't apply the same way — Skip is a route/quote service, exported on the default entry.

🤖 three-lane review (Claude load-bearing + Codex gpt-5.3-codex adversarial + GLM dead). Fixes pushed as df61bd6e.

PR #834, opened 2026-06-23
feat(sdk): add sdk.swap.astroport (Terra Astroport in-chain swap)

gomesalexandre:
## what

Adds `sdk.swap.astroport` — `buildAstroportSwap` quotes (read-only `simulate_swap_operations` on the live phoenix-1 LCD) and builds an **unsigned** Astroport router `wasm_execute` envelope for Terra v2 in-chain swaps (LUNA ↔ ASTRO / ampLUNA / aUSD / etc).

Pure-crypto primitive: it quotes + builds-unsigned, and **never signs or broadcasts**. The returned envelope is handed to the SDK signing path by the consumer.

Two execute shapes, picked by the offer asset:
- **Native offer** (`uluna`, `factory/...`, `ibc/...`): direct router call with `funds`.
- **CW20 offer** (`terra1...` 32-byte contract): base64 `Cw20ReceiveMsg.send` envelope through the CW20, funds forwarded to the router.

Fund-safety carried over from the mcp-ts original:
- `terra1...`-prefixed denom that fails bech32 decode is **rejected loudly** (model fabrication) instead of silently falling through to `native_token`.
- 20-byte `terra1...` (a user account) is rejected as a swap asset.
- explicit `recipientAddress` is the only path that sets the inner `to` — a missing `to` means "router sends to the signer" (the safe default), so a prompt-injected `fromAddress` can't redirect proceeds. Surfaced top-level as `recipientMode: 'self' | 'third_party'` so the approval layer needn't decode the nested base64.
- `minReceive` is integer-only bps math (no float drift on the signed payload); slippage capped at 5%.

## why

Part of the mcp-ts/backend → SDK code-as-action consolidation. The mcp-ts `build_astroport_swap` tool's crypto core (address validation, asset classification, route assembly, min-receive math, unsigned envelope) becomes a reusable SDK primitive. The orchestration-only concerns — Skip-fallback, Blockaid scan-request wrapping, quest-metadata, price-oracle USD derivation — intentionally stay in the consumer.

## consumers unblocked

- **mcp-ts** `build_astroport_swap` can drop its hand-rolled bech32/asset/min-receive logic and wrap `sdk.swap.astroport`.
- **agent-backend** and **Station/Windows** get the same in-chain Terra swap builder without re-implementing the Astroport wire format.

## RECEIPT

Live quote against phoenix-1 (`terra-lcd.publicnode.com`), real `simulate_swap_operations`, builds the unsigned envelope, no broadcast:

```
$ node --import tsx scripts/receipts/swap_astroport.mjs

=== sdk.swap.astroport receipt (LIVE phoenix-1 LCD) ===
params: {
  "fromAddress": "terra1dcegyrekltswvyy0xy69ydgxn9x8x32zdtapd8",
  "offerAssetDenom": "uluna",
  "offerAmount": "1000000",
  "askAssetDenom": "terra1nsuqsk6kh58ulczatwev87ttq2z6r3pusulg9r24mfj2fvtzd4uq3exn26",
  "slippageTolerance": 0.01
}

--- unsigned wasm_execute envelope ---
{
  "txType": "wasm_execute",
  "chain": "Terra",
  "chainId": "phoenix-1",
  "fromAddress": "terra1dcegyrekltswvyy0xy69ydgxn9x8x32zdtapd8",
  "contractAddress": "terra18plp90j0zd596zt3zdsf0w9vvk5ukwlwzwkksxv9mdu8rscat9sqndk5qz",
  "executeMsg": "{\"execute_swap_operations\":{\"operations\":[{\"astro_swap\":{\"offer_asset_info\":{\"native_token\":{\"denom\":\"uluna\"}},\"ask_asset_info\":{\"token\":{\"contract_addr\":\"terra1nsuqsk6kh58ulczatwev87ttq2z6r3pusulg9r24mfj2fvtzd4uq3exn26\"}}}}],\"minimum_receive\":\"6706674\"}}",
  "funds": [{ "denom": "uluna", "amount": "1000000" }],
  "gas": 600000,
  "recipientMode": "self",
  "quote": {
    "offerAsset": "uluna",
    "offerAmount": "1000000",
    "askAsset": "terra1nsuqsk6kh58ulczatwev87ttq2z6r3pusulg9r24mfj2fvtzd4uq3exn26",
    "expectedAskAmount": "6774419",
    "minReceive": "6706674",
    "slippageTolerance": 0.01
  }
}

OK: live quote = 6774419 uASTRO base units
OK: min_receive (1% slippage) = 6706674
OK: built unsigned envelope, no broadcast.
```

Quality gates (worktree off `origin/main`):

```
$ yarn workspace @vultisig/sdk typecheck   # 0 errors
$ yarn workspace @vultisig/sdk test tests/unit/tools/swap/astroport.test.ts
  Test Files  111 passed (111)
       Tests  1527 passed (1527)
```

The unit test covers native + CW20 offer shapes, the base64 `Cw20ReceiveMsg` envelope, explicit-recipient `to`, the fund-safety rejections (fabricated `terra1`, 20-byte account, wrong HRP, zero/non-integer amount, slippage > 5%), and the bps min-receive math.

### DIFF
--- a/.changeset/sdk-swap-astroport.md
+++ b/.changeset/sdk-swap-astroport.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': minor
+---
+
+Add `sdk.swap.astroport` — `buildAstroportSwap` quotes (read-only `simulate_swap_operations`) and builds an unsigned Astroport router `wasm_execute` envelope for Terra v2 (phoenix-1) in-chain swaps. Pure-crypto: never signs or broadcasts. Ported from mcp-ts. Also exports the helpers `assembleAstroportSwap`, `classifyAstroportAsset`, `computeAstroportMinReceive` and the `ASTROPORT_ROUTER` / `TERRA_LCD` / `TERRA_CHAIN_ID` constants.
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -352,6 +352,8 @@ export { CosmosMsgType } from './types'
 // ============================================================================
 
 export type {
+  AstroportSwapResult,
+  BuildAstroportSwapParams,
   Coin,
   CoinKey,
   CoinMetadata,
@@ -367,7 +369,12 @@ export type {
 export {
   abiDecode,
   abiEncode,
+  assembleAstroportSwap,
+  ASTROPORT_ROUTER,
+  buildAstroportSwap,
   chainFeeCoin,
+  classifyAstroportAsset,
+  computeAstroportMinReceive,
   deriveAddressFromKeys,
   evmCall,
   evmCheckAllowance,
@@ -389,6 +396,8 @@ export {
   resolve4ByteSelector,
   resolveEns,
   searchToken,
+  TERRA_CHAIN_ID,
+  TERRA_LCD,
   VerifierClient,
 } from './tools'
 
--- a/packages/sdk/src/platforms/react-native/index.ts
+++ b/packages/sdk/src/platforms/react-native/index.ts
@@ -200,6 +200,22 @@ export async function prepareSwapTxFromKeys(...args: unknown[]) {
   return mod.prepareSwapTxFromKeys(...(args as Parameters<typeof mod.prepareSwapTxFromKeys>))
 }
 
+// Astroport in-chain swap (Terra v2 / phoenix-1) — builds an unsigned
+// wasm_execute envelope. Pure-crypto: only @scure/base (bech32), Buffer and
+// fetch, all RN-safe, so a static re-export is fine (no chain-client deps to
+// externalize). The RN consumer (Station / vultiagent-app) needs this to build
+// the signable Terra swap msg; omitting it would force a re-implementation.
+export type { AstroportSwapResult, BuildAstroportSwapParams } from '../../tools/swap/astroport'
+export {
+  assembleAstroportSwap,
+  ASTROPORT_ROUTER,
+  buildAstroportSwap,
+  classifyAstroportAsset,
+  computeAstroportMinReceive,
+  TERRA_CHAIN_ID,
+  TERRA_LCD,
+} from '../../tools/swap/astroport'
+
 // EVM utilities (viem-backed — requires app to install `viem` as a peer dep)
 export {
   abiDecode,
--- a/packages/sdk/src/tools/index.ts
+++ b/packages/sdk/src/tools/index.ts
@@ -9,12 +9,25 @@ export type { Coin, CoinKey, CoinMetadata, KnownCoin, KnownCoinMetadata, TokenMe
 export { chainFeeCoin, getTokenMetadata, knownTokens, knownTokensIndex, searchToken } from './token'
 
 // Swap
-export type { FindSwapQuoteParams, NativeSwapMinAmountIn, SwapQuote } from './swap'
+export type {
+  AstroportSwapResult,
+  BuildAstroportSwapParams,
+  FindSwapQuoteParams,
+  NativeSwapMinAmountIn,
+  SwapQuote,
+} from './swap'
 export {
+  assembleAstroportSwap,
+  ASTROPORT_ROUTER,
+  buildAstroportSwap,
+  classifyAstroportAsset,
+  computeAstroportMinReceive,
   findSwapQuote,
   getNativeSwapDecimals,
   getNativeSwapMinAmountIn,
   NATIVE_SWAP_MIN_OUTBOUND_FEE_MULTIPLIER,
+  TERRA_CHAIN_ID,
+  TERRA_LCD,
 } from './swap'
 
 // Verifier client
--- a/packages/sdk/src/tools/swap/astroport.ts
+++ b/packages/sdk/src/tools/swap/astroport.ts
@@ -1,0 +1,423 @@
+/**
+ * buildAstroportSwap — Astroport router swap on Terra v2 (phoenix-1).
+ *
+ * Builds an unsigned CosmWasm `wasm_execute` envelope against the Astroport
+ * router that swaps `offerAssetDenom` for `askAssetDenom` directly on
+ * phoenix-1 — no Skip / Axelar / IBC. Sub-bridge latency for in-chain
+ * LUNA ↔ ASTRO / ampLUNA / aUSD / etc.
+ *
+ * Pure-crypto: quotes (read-only `simulate_swap_operations`) and builds the
+ * unsigned execute payload. It NEVER signs or broadcasts — the returned
+ * envelope is handed to the Vultisig SDK signing path by the consumer.
+ *
+ * Two execute shapes are produced depending on the offer asset:
+ *
+ *   - Native offer (e.g. `uluna`): direct call to the router with
+ *       msg = { execute_swap_operations: { operations, minimum_receive, to } }
+ *       funds = [{ denom, amount }]
+ *
+ *   - CW20 offer (e.g. an ASTRO contract address): a `Cw20ReceiveMsg`
+ *     envelope through the CW20 contract, which forwards funds to the router.
+ *     The inner execute_swap_operations payload is base64-encoded per the
+ *     CW20 spec:
+ *       contract = CW20_ADDR
+ *       msg = { send: { contract: ROUTER, amount, msg: <base64> } }
+ *       funds = []
+ *
+ * Quote step uses the router's `simulate_swap_operations` smart query. That
+ * endpoint only returns `{ amount }` — commission/spread are pool-level
+ * details the multi-hop router does not expose. We surface what the router
+ * returns and derive `minReceive` = quote × (1 − slippage).
+ *
+ * Route assembly uses `astro_swap` operations exclusively. NativeSwap is a
+ * Terra Classic / market-module construct that does not apply on phoenix-1.
+ * The router auto-routes through pools; if no route exists the simulate query
+ * surfaces a `pair_info ... not found` error cleanly.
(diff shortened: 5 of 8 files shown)


gomesalexandre (commented):
## deep adversarial pass — astroport swap

Audit verdict was CORRECT; ran a full adversarial diff of `astroport.ts` vs mcp-ts `src/tools/swap/astroport-swap.ts` + the agent-backend safety envelope. **Port logic is faithful — no defect found.** Pushed test-only hardening in `aa446a18` pinning the fund-safety equivalences the verdict rests on.

### verified clean (X/Y/Z)
- **min-receive / slippage math** — `quote*(10000-bps)/10000` BigInt, integer-only, identical to source. Live phoenix-1 receipt: quote `6774419` uASTRO → min_receive `6706674` @ 1% (= floor(6774419·9900/10000), exact). `DEFAULT_SLIPPAGE=0.01` / `MAX_SLIPPAGE=0.05` / `SWAP_GAS=600_000` / `ASTROPORT_ROUTER` all byte-identical.
- **native-vs-CW20 envelope** — native → direct router execute + `funds=[{denom,amount}]`; CW20 → base64 `Cw20ReceiveMsg` `send{contract,amount,msg}` + `funds=[]`. Matches source split. Verified both shapes in tests + live receipt.
- **to-defaults-to-self (proceed-redirect guard)** — `to` is carried into the inner msg ONLY when `recipientAddress` is explicitly supplied; never defaults to `fromAddress`. A prompt-injected sender cannot pin proceeds. Router's missing-`to` → message-sender (vault-derived) is the safe default. Confirmed.
- **valoper rejection** — source uses `assertNotValidatorHrp()`; the port relies on exact `decoded.prefix === 'terra'`, which rejects the distinct `terravaloper`/`terravalcons` HRPs (verified against `@scure/base` decode — those decode to prefix `terravaloper`/`terravalcons` ≠ `terra` → throw). Equivalent-or-stricter on Terra.
- **classifier fail-loud** — `terra1`-prefixed-but-not-bech32 throws (no silent native_token fallback, the #165 hole); 20-byte account rejected; uppercase canonicalized to lowercase for `addr_validate`. `@scure/base` default 90-char limit comfortably clears 32-byte (71-char) CW20 addrs.
- **amount validation** — positive-integer-only, rejects `0`/all-zeros/non-integer; parity with mcp-ts canonical `validateBaseUnitAmount` (both accept leading-zero non-zero, both reject zero).

### attacker angles checked, no hole
- redirect proceeds via crafted `recipientAddress` → blocked (same `validateTerraAddress` guard runs on the recipient; valoper/valcons/non-terra all rejected).
- degenerate `minimum_receive=0` via malformed quote → `assembleAstroportSwap` re-validates `/^[0-9]+$/` on quoteAmount and throws; `querySimulate` re-validates the LCD-returned amount. Fails closed.
- affiliate/fee leakage — N/A for this primitive (Astroport router has no affiliate field; Skip-fallback affiliate handling is orchestration-only and correctly excluded from the SDK primitive). No fee param to leak.
- orchestration-only concerns (Skip-fallback, Blockaid scan-request, quest-metadata, USD oracle) correctly live in the consumer, not the primitive.

### added (test-only, no logic change)
3 regression tests pinning guards that were untested but load-bearing for the verdict: valoper/valcons **sender** rejection, valoper/non-terra **explicit-recipient** rejection (proceed-redirect surface), malformed **quoteAmount** fail-closed.

**receipts:** `20/20` unit pass · `typecheck` clean (0 errors) · live phoenix-1 receipt built an unsigned `uluna→ASTRO` envelope (no broadcast). Adversarial pass clean.

NeOMakinG (approved):
Fund-safety review of the Astroport in-chain swap builder (highest-risk class: a SELF-BUILT CosmWasm `wasm_execute` — the built msg fields are the fund-binding surface, no external API to blame). Ran 3 lanes (Claude load-bearing + Codex gpt-5.3-codex + GLM 4.6) plus on-chain contract verification, full mutation testing, and a live phoenix-1 receipt. Two findings fixed in-code on the branch; the source crypto was correct, so these were a test-teeth gap and a robustness/casing fold. APPROVING.

## testing approach

- **On-chain contract verification (decisive):** queried the pinned router `terra18plp90j0zd596zt3zdsf0w9vvk5ukwlwzwkksxv9mdu8rscat9sqndk5qz` on phoenix-1 via the LCD — `contract_info.label = "Astroport Router"`. Cross-checked against the official `astroport-changelog` `terra-2/phoenix-1/core_phoenix.json` `router_address` → **byte-identical**. The router is canonical-correct.
- **mcp-ts parity:** schema (`execute_swap_operations` / `astro_swap` / `Cw20ReceiveMsg.send`, `minimum_receive`, `simulate_swap_operations`), router constant, and asset classification are a faithful line-for-line port of `mcp-ts/src/tools/swap/astroport-swap.ts`.
- **Slippage math:** `minReceive = floor(quote × (10000 − round(slip×10000)) / 10000)`, BigInt integer-only. Verified both the bps rounding and the floor division err toward **stricter** protection (lower allowed slippage), never weaker — no float drift on the signed payload.
- **Mutation tests (22 tests, single-threaded):** corrupt-router, drop-haircut, CW20-funds-tamper, native-funds×2, drop-HRP-check, revert-lowercase-fold — each fails the expected count after the fixes below.
- **Live receipt:** `node --import tsx scripts/receipts/swap_astroport.mjs` against `terra-lcd.publicnode.com` — real `simulate_swap_operations`, builds the unsigned envelope, no broadcast. Pass.
- **Gates:** typecheck (0 new errors; the 29 `TS2307 @vultisig/mpc-types/mpc-wasm` are pre-existing unbuilt-workspace), complexity (sonarjs cognitive-35, exit 0), prettier (`.config/.prettierrc`, clean). Prior-SHA "Test Suite" CI green; my fix-SHA `b270fca6` gates verified locally (CI did not re-trigger on the push at review time — diff is a 1-line fold + test pins + RN export block).

## blocking

None.

## preferably-blocking

None. Slippage protection IS present and correct: `minimum_receive` is Astroport's router-level slippage bound (the multi-hop equivalent of a pair swap's `max_spread`/`belief_price`), defaulted to 1% and capped at 5%, integer-bps math, fail-safe rounding.

## should-fix

The following two were found and **fixed in-code** on the branch (`b270fca6`); the source crypto was already correct.

1. **(test-teeth, #228-class) Router address asserted only against the imported symbol.** `expect(result.contractAddress).toBe(ASTROPORT_ROUTER)` mutates both sides of the equality if the constant is corrupted — a wrong-router mutation (funds → wrong/attacker contract, the single most fund-critical field) left **zero** failing tests. **Fixed:** pinned the canonical literal `ASTROPORT_ROUTER_LITERAL` alongside the symbol assertion in both the native and CW20 paths; the corrupt-router mutation now fails 2 tests.

2. **(robustness, Codex lane) Uppercase `fromAddress`/`recipientAddress` not normalized.** `validateTerraAddress` returned the raw-case string, while CW20 contract addresses are folded to lowercase (because CosmWasm `addr_validate` rejects a non-normalized `TERRA1...`). An all-uppercase recipient decodes via bech32 but the built inner `to` / sender would be rejected at execute time. Fails closed on-chain (tx reverts, no fund loss), but it's an inconsistent, easily-tripped gap. **Fixed:** `validateTerraAddress` now folds to lowercase (bech32 already confirmed case-uniformity), matching the CW20 path; added 2 normalization tests with teeth.

3. **(RN export, recurring family pattern) Astroport omitted from the React-Native entry.** Every pure-crypto signable-payload builder in this consolidation family gets a static RN re-export so Station / vultiagent-app don't re-implement the wire format. The builder is pure-crypto (`@scure/base` bech32 — already RN-resident via `chains/cosmos/tx.ts` — plus `Buffer` and `fetch`), so no chain-client externalization is needed. **Fixed:** added a static RN re-export block.

## suggestion

- `assembleAstroportSwap` is exported and accepts a caller-supplied `quoteAmount`, so a caller can build a payload from a quote it computed for a different route (quote/build decoupling). This is by-design (the docstring is explicit, and it mirrors mcp-ts) and still validates addresses/amount/slippage, so it's not a dangerous fail-open — but consider branding the quote to the operations array in a follow-up if the assembler stays public. Not blocking.

## q

- The native/factory/IBC denom path intentionally does not pre-validate the denom (the LCD `simulate_swap_operations` is the source of truth and surfaces `pair_info ... not found` cleanly). Confirming that's the intended fail-closed boundary — an unknown denom fails at the simulate round-trip in `buildAstroportSwap`, never building a bogus envelope. (The bare `assembleAstroportSwap` skips that, hence the suggestion above.)

## risk

Low. Router canonical-verified on-chain and against the official changelog; swap msg shape is a faithful mcp-ts port; slippage protection present and fail-safe; funds match offer amount (native: `funds=[{denom,amount}]`; CW20: `funds=[]` with amount in `send`); recipient defaults to signer and an explicit `to` runs the same address guard (no injected-sender redirect); unsigned output; fail-closed on bad asset/amount/slippage/quote. Fixes are additive (1-line fold + test pins + RN export) and gate-clean.

## verdict

**APPROVE.** Contract canonical-correct, msg shape correct, slippage bounded, funds match, recipient safe, fail-closed, unsigned. Three should-fix items (inert router assertion, address-casing fold, missing RN export) fixed in-code at `b270fca6` with mutation-verified test teeth and live receipt.

---
Reviewed with Claude (load-bearing) + Codex gpt-5.3-codex (converged on the casing finding; did not catch the inert router-pin — lanes complementary) + GLM 4.6 (dead, 429 — best-effort). On-chain router verification and mutation testing were Claude load-bearing.

PR #835, opened 2026-06-23
feat(sdk): add sdk.swap.jupiter (Solana Jupiter route + unsigned tx)

gomesalexandre:
## what

Adds `buildJupiterSwapTx` (`sdk.swap.jupiter`) — the Solana leg of swap quoting. `findSwapQuote` already covers the cross-chain majors (THORChain / MayaChain / 1inch / LiFi / KyberSwap) but does NOT cover same-chain Solana pairs. This fetches a live Jupiter V6 aggregator quote and builds a fully-serialized **UNSIGNED** `VersionedTransaction` (base64) plus route + amount metadata.

Vault-free, pure crypto: it **quotes + builds-unsigned only** — never signs, never broadcasts.

## why

Solana → Solana swaps go through Jupiter direct (not SwapKit) so the affiliate fee fields (`platformFeeBps` on /quote, `feeAccount` on /swap) can be wired in — SwapKit can't forward them. Keeping this primitive in the SDK lets mcp-ts `execute_swap` and the backend drop their bespoke Jupiter copies and call the canonical `@vultisig/sdk` path.

Affiliate-fee mechanics are self-contained (treasury owner `5QXe...wUSB` + 50 bps + per-output-mint ATA resolution). Safe default: when no treasury ATA is configured for the output mint, BOTH `platformFeeBps` AND `feeAccount` are omitted — passing one without the other would have Jupiter quote a route the user cannot execute. ATA map is empty today, so the fee is OFF until ops pre-create the ATAs (the structural Jupiter-direct win still lands).

## consumers unblocked

- `mcp-ts` `jupiterSwap` / `execute_swap` Solana leg → call `sdk.buildJupiterSwapTx` instead of the local copy
- `agent-backend` Solana swap card builder
- Any client needing an unsigned Solana swap tx without a full vault

## ported from

`mcp-ts/src/tools/swap/jupiterSwap.ts` (0 SDK imports). Part of the mcp-ts/backend → SDK code-as-action consolidation.

## test matrix

- typecheck: 0 errors (exit 0)
- unit: 1519 passed (8 new jupiter cases + public-export guard), full suite green
- lint: clean on all touched files
- receipt: live Jupiter quote SOL→USDC, real unsigned tx built (below)

## RECEIPT

```
$ node_modules/.bin/tsx scripts/receipts/swap_jupiter.mjs

=== sdk.swap.jupiter — LIVE Jupiter quote (SOL → USDC) ===
input        : 0.1 SOL (So11111111111111111111111111111111111111112)
output mint  : USDC (EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v)
route        : HumidiFi
out amount   : 6868408 base units  (~6.868408 USDC)
min out      : 6799724 base units  (~6.799724 USDC, 1% slippage)
price impact : 0
affiliate fee: OFF (no treasury ATA configured)
unsigned tx  : AQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA... (948 b64 chars)

RECEIPT OK: live unsigned Solana swap tx built (never signed, never broadcast).
```

(Quote/route vary per call — Jupiter re-routes live. The receipt asserts a non-empty unsigned tx + positive out amount; never broadcasts.)

## risk

Low. Additive, new export, behind no existing call site yet. Affiliate fee OFF by default (empty ATA map). Quotes/builds-unsigned only — no signing/broadcast surface.

## Summary by CodeRabbit

* **New Features**
  * Added support for building unsigned Jupiter swap transactions for SOL swaps.
  * Exposed swap quote/result types and Jupiter fee/slippage settings through the SDK.
  * Added a receipt script that fetches a live swap quote and prints transaction details.

* **Bug Fixes**
  * Improved validation for invalid swap amounts and matching input/output mints.
  * Added clearer error handling when quote or swap requests fail, or when a swap transaction is not returned.


gomesalexandre (commented):
## deep adversarial pass — clean, +1 hardening landed (43276f74)

Did a full adversarial pass on `sdk.swap.jupiter` vs the mcp-ts source (`src/tools/swap/jupiterSwap.ts` + `jupiterAffiliateConfig.ts`) — "how does an attacker break this / where could a fund-safety guard get silently dropped". Verdict: faithful port, no P0/P1, and the two extra guards the SDK adds (`amountBaseUnits<=0n`, `inputMint===outputMint`) are pure hardening. Pushed one improvement to make the most important invariant actually testable.

### what I verified
- **platformFeeBps + feeAccount omit-BOTH / include-BOTH symmetry** ✅ — gated on a single `feeAccount` var read once from the resolver, so you structurally cannot send one without the other. mcp-ts parity exact. This is the critical guard against Jupiter quoting a route whose fee transfer lands nowhere (SPL Token 0x17 InvalidAccountData on broadcast).
- **affiliate-OFF default** ✅ — `JUPITER_AFFILIATE_FEE_ATAS` empty (same as mcp-ts) → `resolveJupiterFeeAccount` → `null` → both fields omitted. Live receipt confirms `affiliate fee: OFF`. No affiliate leakage; owner `5QXe…wUSB` / 50 bps / `api.vultisig.com/jup` constants identical to source.
- **slippage** ✅ — default 100 bps (1%), matches `recipes/sdk/swap/jupiter.go`; forwarded to `/quote` and `minOutAmount` echoes `otherAmountThreshold` (slippage-protected min). No silent widening.
- **route/quote edge cases** ✅ — non-ok `/quote`, non-ok `/swap`, 200-with-embedded-error, and missing-`swapTransaction` all throw; never returns a partial/empty tx. Unsigned-only contract holds (no signing/broadcast surface; receipt never broadcasts).
- **no missed mcp-ts bugfix** ✅ — jupiter history clean (only #343 wiring + #482 knip); the post-#343 "owner-wallet-as-feeAccount broke broadcast" lesson is already encoded as the ATA-not-owner invariant in both source and port.

### one gap found + fixed (43276f74)
The affiliate-**ON** branch was **untestable**: the canonical ATA map is empty and `resolveJupiterFeeAccount` was a hard intra-module call (`vi.mock` can't intercept it — I confirmed by writing a mock test that failed OFF). So the single most fund-safety-critical invariant had **zero coverage** — a future dev populating the ATA map while regressing one of the two `if (feeAccount)` guards would ship silently.

Fix: added an optional `resolveFeeAccount` test seam to `buildJupiterSwapTx` defaulting to the canonical resolver. **No production behavior change** (prod never passes it). It locks:
- ON path → BOTH `platformFeeBps` + `feeAccount` present together
- OFF path → BOTH omitted
- plus build-time route failure / embedded-error / `apiBaseUrl` trailing-slash strip

### receipts
- typecheck: 0 errors (exit 0)
- unit: **1524 passed** (+5 new jupiter cases; was 1519)
- live receipt: `node scripts/receipts/swap_jupiter.mjs` → real SOL→USDC unsigned tx built, `affiliate fee: OFF`, never broadcast

```
=== sdk.swap.jupiter — LIVE Jupiter quote (SOL → USDC) ===
route        : GoonFi V2 -> SolFi V2
out amount   : 6900946 base units  (~6.900946 USDC)
min out      : 6831937 base units  (~6.831937 USDC, 1% slippage)
affiliate fee: OFF (no treasury ATA configured)
unsigned tx  : AQAA...  (1100 b64 chars)
RECEIPT OK: live unsigned Solana swap tx built (never signed, never broadcast).
```

Affiliate stays OFF until a treasury ATA lands — unchanged. Adversarial pass clean otherwise.

packages/sdk/src/tools/swap/jupiter.ts
```diff
@@ -0,0 +1,297 @@
...
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(swapBody),
+  })
+
+  if (swapResp.error) {
+    throw new Error(`Jupiter swap error: ${swapResp.error}`)
+  }
+
+  if (!swapResp.swapTransaction) {
```
NeOMakinG, line 283: `should-fix:` this `missing swapTransaction` guard isn't exercised. Mutation check: deleting these three lines leaves all 13 tests green. The suite covers the `swapResp.error` branch (200-OK *with* an `error` field) but never the 200-OK / no-`error` / no-`swapTransaction` case this guard protects. The source is correct and faithful to mcp-ts — this is a test-hardening gap (#228 class).

Suggested: add a case where `/swap` returns `{ status: 200, body: {} }` (no error, no swapTransaction) and assert it rejects with `/missing swapTransaction/`.

packages/sdk/src/tools/swap/jupiter.ts
```diff
@@ -0,0 +1,297 @@
...
+  if (feeAccount) {
+    quoteParams.set('platformFeeBps', String(JUPITER_PLATFORM_FEE_BPS))
+  }
+
+  const quote = await fetchJupiter<JupiterQuoteResponse>(`${base}/swap/v1/quote?${quoteParams.toString()}`)
+
+  // Step 2: Build the (unsigned) swap transaction. Include feeAccount only
+  // when resolved.
+  const swapBody: Record<string, unknown> = {
+    userPublicKey,
```
NeOMakinG, line 262: `suggestion:` `userPublicKey` is forwarded verbatim with no local validation. No exploit — a malformed key degrades to a Jupiter 4xx (fail-closed) — but a base58-decode-to-32-bytes check before the network call turns an opaque remote error into a deterministic caller error and keeps junk pubkeys out of the proxy's logs/metrics. Optional, non-blocking.

NeOMakinG (approved):
## testing approach

Three lanes (Claude load-bearing + Codex gpt-5.4 second-opinion + GLM — GLM dead ECONNRESET, Codex 0-byte stall on first run then converged on retry). Built the branch in an isolated worktree, ran the jupiter unit suite single-threaded (13/13), 4 source mutations to check the tests have teeth, the public-export wiring, tsc, the complexity gate, prettier, and the **live** receipt against the public Jupiter Lite API.

This is the highest-fund-risk class — a swap builder that hands the user an **unsigned tx the Jupiter API constructed**, not a tx we built. I read `buildJupiterSwapTx` against its mcp-ts origin (`src/tools/swap/jupiterSwap.ts`) and walked the trust boundary adversarially.

**Verification evidence:**
- live receipt: real 896-b64 unsigned SOL→USDC `VersionedTransaction` built (route Flux, 1% slippage floor applied, never signed/broadcast)
- USDC mint `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v` ✅ correct mainnet mint; SOL native `So1111…112` ✅ correct
- mutation teeth: swap input/output mints → 2 fail; drop amount>0 guard → 1 fail; break the omit-BOTH affiliate symmetry → 2 fail
- gates: tsc 0 jupiter errors, complexity (sonarjs cog-35) exit 0, prettier `--config .config/.prettierrc` clean on the `.ts` files, **all CI green** (Unit / Integration / Lint+Type / CodeRabbit)

## trust boundary (the core risk)

This is the **inherent Jupiter aggregator model**: `/quote` → `/swap` returns a serialized tx Jupiter built, and the function returns it verbatim without locally asserting the tx matches the requested mints/amount/feePayer (a Solana versioned tx is opaque to the EVM/cosmos decode #848 / policy #847 stack). That trust surface is real, but it is (a) **exactly the shipped production mcp-ts behavior this ports** — not a regression — and (b) mitigated by the slippage floor (`otherAmountThreshold`, default 1%, surfaced as `minOutAmount`), the canonical Vultisig proxy default, and unsigned-only output. Acceptable for this primitive; flagging the boundary explicitly rather than blocking on it.

## should-fix

- **test gap on the `missing swapTransaction` guard** — mutation MUT4 (delete the `if (!swapResp.swapTransaction)` throw) leaves all 13 tests green. The 200-OK-with-`error` branch is covered, but the 200-OK / no-error / no-`swapTransaction` branch the guard actually protects is not. Source guard is correct + faithful to mcp-ts; pure test-hardening (#228 class). Inline below.

## suggestion

- **validate `userPublicKey` as base58-32-bytes before the network call** (Codex). No direct exploit — a bad value degrades to a Jupiter 4xx (fail-closed) — but local rejection turns an opaque remote error into a deterministic caller error and keeps junk out of the proxy's logs/metrics. Inline below.
- **`apiBaseUrl` override** — fine as an SDK developer ergonomics knob (default is the canonical proxy, same pattern as every other builder); worth a one-line doc that pointing it at an untrusted host is equivalent to signing whatever that host returns.

## q

- Is there a mandatory pre-sign Solana validate/simulate gate on the consuming path (mcp-ts `execute_swap` / backend), or is the slippage floor the only protection before the user signs? Doesn't block this primitive — just confirms how soft the boundary stays downstream.

## RN export

Not wired into `src/platforms/react-native/index.ts` — and that's **correct here**, unlike #831/#832. The whole swap-quote family is deliberately out of RN scope: the sibling `findSwapQuote` is also absent from the RN index. Not a gap.

## risk

Low. Additive, new export, no existing call site yet, affiliate fee OFF by default (empty ATA map), unsigned-only. Faithful mcp-ts port that *adds* an amount>0 guard, an identical-mint guard, and the omit-BOTH affiliate symmetry — all mutation-verified.

## verdict

**APPROVE.** Quote params correct (mints/amount/slippage), `userPublicKey` forwarded as the swapper, fail-closed on every API error path, unsigned output, omit-BOTH affiliate symmetry locked, RN omission is intentional family convention, all CI + complexity + prettier green, live receipt builds a real unsigned tx. The trust-boundary and the two follow-ups are non-blocking and tracked above.

---
🤖 reviewed by Claude (correctness/fund-safety, load-bearing) + Codex gpt-5.4 (adversarial second-opinion, converged) — GLM lane down.

PR #837, opened 2026-06-23
feat(sdk): add sdk.validate.recipientSanity (null/self-send/malformed-recipient)

gomesalexandre:
## what

Adds `recipientSanity` — a pure, deterministic recipient-address sanity check exposed from the SDK package root, plus the three building-block predicates `isNullAddress` / `isSelfSend` / `isMalformedEvmAddress` and the `Recipient*` types.

Three checks, all pure format/equality (no network, no vault, no signing, no broadcast):

- **null** — EVM zero address, the `0x...dEaD` burn address, the Solana System Program address, and the Solana Incinerator. No private key controls these, so a send there is irreversible loss.
- **selfSend** — `from === recipient` (case-insensitive, trimmed).
- **malformedEvm** — looks like an EVM attempt (`0x` + hex only) but is not a valid `0x` + 40-hex address.

## why

This is the deterministic format slice of the agent-backend recipient validators (`null_recipient.go`, `self_send_warning.go`, `malformed_evm_recipient.go`). Those Go validators currently re-implement the format/equality math inline. Lifting the **pure** part into the SDK lets every signing surface share one primitive.

The agent-judgement layers stay where they belong: intent-vs-user-message matching, fabricated-vs-grounded reasoning, prompt-injection detection, and the "did the model emit the self-send warning?" check remain in the agent backend. This PR ports ONLY the deterministic `isNullAddress` / `from==recipient` / `isMalformedEVMAddress` checks.

## consumers unblocked

- mcp-ts `execute_send` / `build_*` tools can call one shared `recipientSanity` instead of hand-rolling null/malformed checks per tool.
- agent-backend validators can delegate the format/equality math to the SDK and keep only the agent-judgement wrapper.

## receipt

```
$ node --import tsx scripts/receipts/validate_recipient_sanity.mjs
=== sdk.validate.recipientSanity — runnable receipt ===

[PASS] null recipient (EVM zero address)
        in:    {"recipient":"0x0000000000000000000000000000000000000000"}
        flags: null
        full:  {"recipient":"0x0000000000000000000000000000000000000000","flagged":true,"flags":["null"],"isNull":true,"isSelfSend":false,"isMalformedEvm":false}

[PASS] null recipient (EVM 0x...dEaD burn)
        flags: null

[PASS] null recipient (Solana System Program)
        in:    {"recipient":"11111111111111111111111111111111"}
        flags: null

[PASS] self-send (from === recipient, case-insensitive)
        in:    {"from":"0x742d35Cc...","recipient":"0x742d35cc..."}
        flags: selfSend

[PASS] malformed EVM recipient (0xdeadbeef — too short)
        in:    {"recipient":"0xdeadbeef"}
        flags: malformedEvm

[PASS] clean valid distinct recipient (control — no flags)
        flags: (clean)

=== 6/6 cases as expected ===
```

Unit test: `tests/unit/tools/validate/recipientSanity.test.ts` — **22/22 pass**.
Typecheck: `yarn workspace @vultisig/sdk typecheck` — **0 errors**.
Lint: clean on all touched files.

A null recipient, a self-send (`from == to`), and a malformed `0x` addr are each flagged; a valid distinct recipient stays clean.

## risk

Low — additive pure function, zero existing call sites changed, conservative flagging (only never-valid addresses flagged as null; only unambiguous EVM-shaped attempts flagged as malformed). Never signs or broadcasts.

## Summary by CodeRabbit

* **New Features**
  * Added recipient address validation utilities to the SDK, including checks for null addresses, self-sends, and malformed EVM addresses.

* **Tests**
  * Added unit tests for recipient validation utilities.
  * Added validation script for recipient sanity checks.


gomesalexandre (commented):
Adversarial pass — clean on logic, hardened on tests.

Diffed \`isNullAddress\` / \`isMalformedEvmAddress\` / \`isSelfSend\` line-for-line against the agent-backend Go source (\`null_recipient.go\`, \`malformed_evm_recipient.go\`, \`self_send_warning.go\`) including their \`_test.go\` parity tables. The port is byte-for-byte faithful — **no false-negative** (a bad recipient slipping through) and **no false-positive** across every variant I could throw at it:

- **null / burn:** EVM zero (\`0x0…0\`), \`0X\`-uppercase-prefix zero (Go lowercases the whole string before the prefix check — SDK does too via \`toLowerCase()\`), canonical \`0x…000dEaD\` case-insensitive, Solana System Program, Solana Incinerator. Partial-zero (\`0x…0001\`) and non-canonical 40-hex ending in \`dead\` correctly NOT flagged (the canonical-equality path, not length — CodeRabbit #1175).
- **malformedEvm:** \`^0x[0-9a-fA-F]+$\` + \`length !== 42\` matches Go exactly. Bare \`0x\` (no fail-open), \`0xGGGG\` non-hex, and uppercase \`0X…\` all correctly treated as non-malformed (different family). Every too-short/too-long length flags.
- **selfSend:** case-insensitive trimmed equality == Go \`strings.EqualFold\` after trim; empty-side guard matches.

No dropped guards, no wrong constants, no phantom deps, never throws on garbage. The SDK is the **pure format/equality slice** only — the agent-judgement layers (intent match, signing-surface severity, Path-B user-message scan) correctly stay in the Go validators, as the PR body states.

**One gap fixed:** the SDK suite didn't pin the trickier Go-table edges (the exact cases that previously regressed in Go). Added them so a future refactor can't silently drift the parity — see a17f2547.

Receipts (worktree off the PR branch):
- unit: **28/28 pass** (was 22; +6 Go-parity cases)
- typecheck: **0 errors**
- receipt script: **6/6 cases as expected**

NOT merging (own PR, needs external stamp).

packages/sdk/src/tools/validate/recipientSanity.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  // Deterministic flag order: null, selfSend, malformedEvm.
+  const flags: RecipientSanityFlag[] = []
+  if (nullFlag) flags.push('null')
+  if (selfSendFlag) flags.push('selfSend')
+  if (malformedEvmFlag) flags.push('malformedEvm')
+
+  return {
+    recipient,
+    flagged: flags.length > 0,
```
NeOMakinG, line 175: should-fix: `flagged:false` reads as a clean verdict even when no chain-format validation ran for this recipient (cosmos bech32 / BTC / wrong-checksum EVM all return `flagged:false`, indistinguishable from "passed all checks"). A consuming tool (mcp-ts execute_send, backend validators) could collapse this into "safe to send". Non-blocking and matches the documented pure-slice scope + the Go source shape — but consider a `coverage` discriminant or a name like `hasKnownWarnings` so callers wire this as an additive check, never as the recipient gate. (Codex-flagged, tier-verified.)

packages/sdk/src/tools/validate/recipientSanity.ts
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * Ported from self_send_warning.go:detectSelfSendTurn (the args-comparison
+ * branch only — the agent-judgement "did the model emit the warning?" layer
+ * stays in the backend).
+ */
+export function isSelfSend(from: string, recipient: string): boolean {
+  const f = from.trim()
+  const r = recipient.trim()
+  if (f === '' || r === '') return false
+  return f.toLowerCase() === r.toLowerCase()
```
NeOMakinG, line 137: q: `.toLowerCase()` case-folds before comparing — correct for EVM, but case-sensitive for base58 (Solana). Two distinct base58 addresses differing only by case would compare equal → a self-send *false-positive*. This is exact Go parity (`strings.EqualFold`), it is a warning not a block, and it errs toward caution, so not a defect here. Flagging only so that IF self-send ever graduates to a hard block, this is made representation-aware per chain first.

NeOMakinG (approved):
clean, well-scoped fund-safety primitive. recipientSanity is the pure deterministic slice of three agent-backend validators (null_recipient.go / self_send_warning.go / malformed_evm_recipient.go) — and i verified it is line-for-line Go parity, not a re-interpretation. APPROVE.

## testing approach

3 lanes. Claude load-bearing (Go-parity diff, fail-closed enumeration, mutation tests); Codex gpt-5.3-codex second-opinion (returned after one 0-byte stall under load — best-effort, converged on 2 non-blocking should-fix); GLM dead (exit 1, best-effort skipped).

- worktree `origin/feat_sdk_validate_recipient_sanity` @ a17f2547, full `yarn install`
- unit test single-thread (`VITEST_MAX_THREADS=1`): **28/28 pass**
- typecheck: 0 recipientSanity errors (the `@vultisig/mpc-types`/`mpc-wasm` TS2307s are pre-existing unbuilt-workspace dist-missing, unrelated)
- `quality:complexity` (sonarjs cognitive-complexity 35, the real CI gate — run explicitly, NOT yarn lint): **exit 0**, clean
- receipt `validate_recipient_sanity.mjs`: **6/6 as expected**
- CI: CodeRabbit / Integration / Lint+TypeCheck / Unit / Test Summary all **pass**
- **mutation tests** (test-teeth, #228 guard): neuter self-send → 3 fail; neuter EVM-zero null → 5 fail; neuter malformed → 4 fail. All three checks have real negative-case coverage.

### detection completeness (what is flagged vs passes as clean)
flagged: EVM zero `0x0..0`, `0x..dEaD` (case-insens), Solana System Program, Solana Incinerator (null); EVM-shaped `0x`+hex wrong-length (malformedEvm); `from==recipient` trimmed+case-insens (selfSend). Multi-flag ordering deterministic (null, selfSend, malformedEvm).

passes as clean (`flagged:false`): valid 42-char EVM (any checksum — no EIP-55 verify), any non-0x family (cosmos bech32, BTC, Solana non-special), bare `0x`, `0X`-prefixed, partial-zero `0x..01`, gibberish/emoji.

### fail-open-on-unknown verdict — NOT blocking
On an unrecognized format the function fails **open** (returns `flagged:false`). This is the **documented contract and exact Go-source behavior**: it is a building-block of three specific checks, not a replacement for recipient validation, and the agent-judgement / per-chain validation layers stay downstream (tx builders reject bad recipients). No regression, no new fund-loss path. The hardening of the *contract shape* so callers don't misread `flagged:false` as "safe" is captured as a non-blocking should-fix below.

### self-send normalization — correct, Go parity
`isSelfSend` trims + lowercases both sides. Go uses `strings.EqualFold(from, recipient)` — equivalent case-insensitive equality. A real self-send (`from==to`) is byte-identical and always caught. The TS `.trim()` is a strict superset of Go (catches whitespace variants too), erring toward flagging.

## should-fix

- contract-shape: `flagged:false` is the top-level verdict even when no chain-format validation ran (cosmos/BTC/wrong-checksum-EVM are indistinguishable from "passed"). A caller could collapse `flagged:false` into "safe to send". Consider a `coverage` discriminant or renaming `flagged` to make "no known warnings" explicit. Non-blocking — matches the documented pure-slice scope and the Go source has the same shape; raising so consumers (mcp-ts execute_send, backend) wire it as an additive check, never as the recipient gate. (Codex-flagged, tier-verified.)

## q

- `isSelfSend` lowercases before comparing, which is correct for EVM but case-folds case-sensitive base58 (Solana). Two distinct base58 addrs differing only by case would compare equal → a self-send *false-positive*. This is exact Go-parity (`EqualFold`), a warning not a block, and errs toward caution — so not a defect here. If self-send ever graduates to a hard block, make it representation-aware per chain. (Codex-flagged, tier-verified down from should-fix → q: it's a false-positive in the safe direction on a near-impossible input, on a warning-only path.)

## risk

Low. Additive pure function, zero existing call sites changed, never signs/broadcasts. Conservative flagging (only never-valid addresses → null; only unambiguous EVM-shaped attempts → malformed). Exact Go parity on all three checks; mutation tests prove the suite catches each check's removal.

## verdict

APPROVE. Detection complete for its scoped checks, self-send normalization correct (Go EqualFold parity), tests prove all three flag paths (mutation 3/5/4), receipt 6/6, complexity gate green, CI all-green. Fail-open-on-unknown is documented design matching the Go source it ports, not a fund-loss bug. The two Codex findings are valid contract-hardening suggestions for consumers, non-blocking.

---
🤖 reviewed by Claude (Opus 4.8) + Codex (gpt-5.3-codex) — GLM lane unavailable

PR #840, opened 2026-06-23
feat(sdk): add sdk.balance.cosmos (bank-denom balances + token metadata)

gomesalexandre:

## what

Adds `sdk.balance.cosmos` (`getCosmosBalance(chain, address)`) — a pure decode-only primitive that reads Cosmos bank-denom balances over LCD and resolves token metadata. Ports the mcp-ts `cosmos-balance.ts` reader into the SDK.

For a given cosmos chain + address it:
- fetches `GET /cosmos/bank/v1beta1/balances/<address>` (with an inlined Polkachu fallback mirror so the SDK carries **no** mcp-ts dependency),
- decimal-scales the native denom (BigInt, no precision loss — handles 18-dec dYdX as cleanly as 6-dec OSMO),
- resolves curated/known denoms (native + a registry-verified IBC table) to `{ symbol, decimals }`,
- resolves remaining `ibc/<hash>` vouchers via the existing SDK `getTokenMetadata` resolver, **pinning decimals from a SAFE table** so a wrong fallback decimal can never silently mis-scale a balance,
- emits unresolvable `ibc/`/`factory/` denoms in raw base units with a `(base units)` symbol caveat + `decimals: null` so downstream pricing skips them (no mispricing).

Supported: Cosmos, Osmosis, Kujira, Terra, TerraClassic, Noble, Dydx, Akash. THORChain/MayaChain are intentionally excluded (their enrichment is the Rujira path, not vanilla bank-denom) and throw a clear `unsupported chain` error.

## why

Pure crypto (read/decode/validate-FORMAT/decimal-scale) — no signing, no broadcast, no agent-judgement. This belongs in the SDK so mcp-ts and agent-backend stop carrying their own cosmos LCD fetch + denom-decode glue. Per the consolidation table this is the `get_cosmos_balance` family -> `sdk.balance.cosmos()` move.

## consumers unblocked

- **mcp-ts** `get_*_balance` cosmos family can drop its bespoke LCD fetch/fallback + denom-decode and wrap `getCosmosBalance`.
- **agent-backend** cosmos balance reads can call the SDK directly instead of round-tripping mcp-ts.

## test

- 11 unit tests (mocked `fetch` + `getTokenMetadata`): native scaling (6 & 18 dec), curated-IBC resolution, unknown-IBC resolution with decimals pinned (resolver's bogus 999 ignored), unresolved base-units caveat, zero-denom filtering, never-throws-on-metadata-failure, empty-address + Rujira-chain rejection, Polkachu 5xx fallback, supported-chain guard.
- A runnable receipt hitting **live** Osmosis + Cosmos Hub LCDs.

### RECEIPT

```
$ yarn workspace @vultisig/sdk vitest run tests/unit/cosmos/cosmosBalance.test.ts
 Test Files  1 passed (1)
      Tests  11 passed (11)

$ cd packages/sdk && npx tsx ../../scripts/receipts/balance_cosmos.mjs
=== Osmosis :: osmo1cyyzpxplxdzkeea7kwsydadg87357qnahakaks ===
native: 0.000148 OSMO  (raw 148)
as_of:  2026-06-23T15:24:18.647Z
denoms held (3):
  ibc/C360EF34... (base units) amount=1000000                formatted=1000000            decimals=null (unresolved)
  USDC                         amount=48000000000            formatted=48000              decimals=6
  OSMO                         amount=148                    formatted=0.000148           decimals=6
resolved=2  unresolved(base-units)=1

=== Cosmos :: cosmos1cyyzpxplxdzkeea7kwsydadg87357qnalx9dqz ===
native: 0 ATOM  (raw 0)
as_of:  2026-06-23T15:24:19.116Z
denoms held (1):
  factory/cosmos1s8q...kvrq/ustars (base units) amount=10                     formatted=10                 decimals=null (unresolved)
resolved=0  unresolved(base-units)=1

$ yarn workspace @vultisig/sdk typecheck   # tsc --noEmit --skipLibCheck
EXIT=0   (0 errors)
```

Live read showcases all three paths: native decimal-scale, curated-IBC resolution (USDC), and the honest unresolved base-units fallback (IBC + factory denoms).

## risk

Low — additive, read-only, behind a new export. 0 signing surface, 0 new typecheck errors, lint clean.


gomesalexandre (commented):
## adversarial pass — bank-denom LCD + IBC/CW20 metadata resolution

Deep diff of this port vs the mcp-ts source (`src/tools/balance/cosmos-balance.ts`) + agent-backend `parseBalanceResult`, focused on the decimals/mispricing surface. Two real gaps found + fixed on the branch (`51ae4875`), rest verified clean.

### found + fixed

**1. (med, fail-soft → fail-loud) offline cross-chain mis-pair guard — `cosmos.ts` getCosmosBalance**
`getCosmosBalance(chain, address)` trusted the caller to pair `chain` + `address` correctly. A wrong pairing (e.g. an `osmo1...` address routed to `Chain.Cosmos`) silently queried a FOREIGN LCD and returned a confidently-wrong (usually empty) balance with zero signal — exactly the kind of "looks fine, is wrong" output this primitive is supposed to structurally avoid. mcp-ts guards the analogous case with a live `chain_id` LCD probe (`assertChainId`) that the port dropped. Restored as an **offline** bech32-HRP check (no extra round-trip): assert the address HRP against the chain's `bech32Prefix` BEFORE the read, throw on mismatch. Terra v2 / TerraClassic share the `terra` HRP so that same-HRP collision stays a documented caller contract (offline can't disambiguate it; mcp-ts's network probe can — noted inline).

**2. (low, parity regression) dropped TerraClassic legacy fiat denoms — `cosmos.ts` KNOWN_DENOMS**
The port kept only `uusd`/`ukrw` and dropped the other 14 legacy Terra-fiat denoms mcp-ts carries (`usdr`→SDTC, `ueur`→EUTC, `ujpy`→JPTC, …), silently regressing each to a raw `denom`-as-symbol passthrough. Restored full mcp-ts parity. Not a mis-scaling (those fall to the honest passthrough, not a wrong decimal), but a real symbol/decimals regression for TerraClassic holders.

+4 tests: mis-pair reject (asserts NO LCD call issued), malformed-address reject, correct-pair accept, restored-legacy-denom resolution.

- **decimals pinned from IBC_SAFE_DECIMALS** — resolver's bogus `decimals: 999` is never trusted; test proves the pin overrides it (USDC → 6, not 999). The denom_trace fee-coin-decimals fallback can't mis-scale: only symbols whose decimals are in the SAFE table resolve.
- **unresolved-denom base-units caveat** — unknown ibc/factory denoms emit `formatted = raw amount`, `decimals: null`, `unresolved: true`, and the `(base units)` suffix lands in `symbol` (survives agent-backend `parseBalanceResult` which drops `display`/`decimals`). No mis-scaling possible; pricing path skips (no symbol match). Confirmed live (factory/ustars + ibc/C360EF34 both base-units, OSMO + USDC decimal-scaled correctly).
- **THOR/Maya rejection** — `undefined` config → `unsupported chain` throw before any read; test covers.
- **BigInt formatBalance** — no precision loss on 18-dec (adydx 2.5 exact), correct trailing-zero strip, no overflow.
- **Polkachu fallback** — 4xx fails fast (not hidden behind mirror), 5xx/network retried then mirrored; READ-idempotent so safe. Test covers 503→mirror.
- **fetch retry** — bounded (MAX_RETRIES=2), 15s timeout, only timeout/TypeError/5xx retried.

### receipts
typecheck 0 errors · eslint clean · **15/15 unit (was 11)** · live receipt unchanged:
```
=== Osmosis :: osmo1cyy...akaks ===
native: 0.000148 OSMO  (raw 148)
  ibc/C360EF34... (base units)  amount=1000000      formatted=1000000  decimals=null (unresolved)
  USDC                          amount=48000000000  formatted=48000    decimals=6
  OSMO                          amount=148          formatted=0.000148 decimals=6
=== Cosmos :: cosmos1cyy...x9dqz ===
  factory/cosmos1s8q...kvrq/ustars (base units)  amount=10  formatted=10  decimals=null (unresolved)
```

packages/sdk/src/tools/balance/cosmos.ts
```diff
@@ -0,0 +1,497 @@
...
+  const hrp = bech32Hrp(address)
+  if (hrp !== config.bech32Prefix) {
+    throw new Error(
+      `getCosmosBalance: address "${address}" has bech32 prefix "${hrp || '(none)'}" but ${chain} expects "${config.bech32Prefix}". Mis-routed address — confirm the chain matches the address.`
+    )
+  }
+
+  const response = await cosmosLcdGet<BankBalancesResponse>(
+    config.restUrl,
+    `/cosmos/bank/v1beta1/balances/${address}`,
```
NeOMakinG, line 418: `should-fix:` bank balances paginate. This single un-paginated `GET /cosmos/bank/v1beta1/balances/{addr}` returns at most the cosmos-sdk default page size (100 entries) — there's no `pagination.limit` param and no `pagination.next_key` follow-up below (the result reads only `response.balances`). An address holding >100 distinct denoms would have a silently **understated** balance set with no signal. This is a faithful port (mcp-ts does the same), so it's parity-preserving and realistic wallets rarely exceed 100 denoms — hence should-fix. Cheapest close: append `?pagination.limit=1000` to the path, or loop on `next_key`.

NeOMakinG (approved):
Read-only Cosmos bank-denom balance reader. Reviewed under the fund-safety / balance-fabrication lens (a wrong divisor or fail-open coercion fabricates a balance that gates fund decisions). Faithful port of mcp-ts `cosmos-balance.ts`. Fixed two recurring gaps in-code and **APPROVE**.

## testing approach

3-lane fan-out. Claude load-bearing (deep source walk + mutation tests + live receipt). Codex gpt-5.3-codex 0-byte stall (model-at-capacity, best-effort, no output). GLM ECONNRESET (dead, best-effort). Claude was the load-bearing lane.

Gates run locally on the fix SHA (`26429bba`) in an isolated worktree — note: pushed fix-commits do not re-trigger CI on gomes branches, but the prior SHA `51ae4875` is **fully CI-green** (Unit + Integration + Lint/TypeCheck + CodeRabbit all pass).

- **test:** `cosmosBalance.test.ts` 15/15 pass (single-fork, in-pkg minimal config to dodge the `@vultisig/mpc-types` integration-setup wall).
- **mutation teeth:** native divisor `decimals:6→7` → **3 fail**; fail-open IBC decimals (pin `IBC_SAFE_DECIMALS` → trust resolver `meta.decimals`) → **1 fail**; neuter HRP mis-pair guard (`if(false)`) → **2 fail**. All load-bearing guards have teeth.
- **tsc** `--noEmit --skipLibCheck`: 0 errors. **complexity** (sonarjs eslint-health): exit 0. **prettier** (`.config/.prettierrc`): clean.
- **live receipt** (`balance_cosmos.mjs`, live Osmosis + Cosmos Hub LCDs): native decimal-scale (`148 → 0.000148 OSMO`), curated-IBC resolution (`USDC` 6-dec), and honest base-units fallback for unresolved `ibc/` + `factory/` denoms all confirmed.

## blocking

None.

## preferably-blocking

None.

## should-fix

**Bank-balance pagination is not handled** (inline). The LCD `GET /cosmos/bank/v1beta1/balances/{addr}` returns at most the cosmos-sdk default page size (100) when no `pagination.limit` is passed, and the code reads only `response.balances` with no `pagination.next_key` follow. An address holding >100 distinct denoms would have a *silently understated* balance set. This is a **faithful port** — mcp-ts's `cosmos-balance.ts` issues the same single un-paginated request, so this is parity-preserving and realistic cosmos wallets rarely exceed 100 denoms — hence should-fix not blocking. A `pagination.limit=1000` query param (or a next_key loop) closes it cheaply.

## suggestion

The 18-dec dydx precision test uses `2500000000000000000` (`= 2.5e18`, exactly float-representable). A `Number()`-based precision-loss mutation in `formatBalance` does **not** fail any test because that specific value round-trips through a float cleanly. The production code is correct (it uses `BigInt` throughout). To give the BigInt path real teeth against a future float regression, use a non-power-of-ten 18-dec value (e.g. `1234567890123456789`) whose float round-trip would lose low-order digits.

## q

The `terra` HRP is shared by Terra v2 and TerraClassic, so the offline mis-pair guard cannot distinguish them — documented as a caller contract (mcp-ts disambiguates via a live chain_id probe). Confirmed this is intentional; flagging only so the consumer wiring (mcp-ts/agent-backend) is aware it must route Terra↔TerraClassic by the `Chain` enum, not by address.

## risk

Low. Additive, read-only, 0 signing surface. The structural fail-closed design is sound: per-denom decimals are BigInt-scaled with the correct exponent; any denom whose decimals can't be *proven* (unknown IBC/factory) is emitted in raw base units with `decimals:null` + a `(base units)` caveat so downstream pricing skips it — mispricing is structurally impossible. LCD 4xx fails fast, 5xx/network retries then falls to the Polkachu mirror, and both-fail throws (no fabricated zero). HRP mis-pair fails loud before any network read. dYdX 18-dec and 6-dec chains both scale correctly.

### review fixes (pushed to branch, `26429bba`)

1. **RN export gap** — `react-native/index.ts` omitted `getCosmosBalance`/`cosmosBalanceChains`/`isCosmosBalanceChain` from its curated allow-list. The reader is RN-safe (only `fetch` + the already-RN-exported `getTokenMetadata`), so added the static re-export. (Recurring pattern across the sdk.* consolidation PRs.)
2. **Missing changeset** — added a `.changeset` minor bump so the release pipeline versions + publishes `@vultisig/sdk` with the new surface. (Recurring; not a hard CI gate but blocks npm publish.)

## verdict

**APPROVE.** Per-denom decimals correct (BigInt, correct exponent per chain incl. 18-dec dYdX). IBC resolution honest (decimals pinned from a safe table; unprovable denoms fail to base-units, never mis-scaled). Fail-closed on LCD error/missing (no fabricated zero). HRP mis-pair guard has teeth. mcp-ts parity faithful. RN export + changeset fixed in-code. Tests prove decimals + fail-closed + IBC pinning + mis-pair guard. CI green on the pre-fix SHA; gates re-verified locally on the fix SHA. The only open item (pagination) is a parity-preserving should-fix, not fund-fabricating in practice.

---
*Reviewed by Claude (load-bearing) + Codex (stalled, best-effort) + GLM (dead, best-effort). Mutation-tested, live-receipt-verified.*

PR #843, opened 2026-06-23
feat(sdk): add sdk.balance.other (sui/ton/tron/xrp/cardano/tao + token variants)

gomesalexandre:
## what

Brings the non-EVM / non-Cosmos balance reads into the SDK under one
`sdk.balance.<chain>()` surface, ported out of mcp-ts (`balance/other-balance`
+ `balance/bittensor-balance`, both 0 SDK imports):

- **XRP** — `getXrpBalance`
- **TRON** — `getTrxBalance` / `getTrc20TokenBalance` / `getTronAccountResources`
- **TON** — `getTonBalance` / `getTonJettonBalance`
- **Sui** — `getSuiBalance` / `getSuiTokenBalance` / `getSuiAllBalances`
- **Cardano** — `getCardanoBalance` (native ADA + native tokens via Koios)
- **Bittensor** — `getTaoBalance` (+ exported `assertBittensorAddress` / `decodeBittensorAddress`)

## why

Pure crypto: decode RPC responses, parse base units, format, validate address
FORMAT. Read-only — nothing here signs or broadcasts. These chains aren't wired
through the EVM `getEvmClient` rail, so the reads talk to public RPC/API
endpoints (and the Vultisig proxy) directly. Consolidating them into the SDK
lets mcp-ts/backend wrap a single typed primitive instead of carrying the
fetch+parse glue per chain.

Fund-safety preserved on the port:
- Tron base58check format guard runs **before** any RPC call (an EVM `0x` /
  cosmos bech32 address can't be silently treated as Tron).
- Sui returns a deterministic `tokens_unavailable` signal on a JSON-RPC error
  or non-integer balance — a typo'd address must not read as an empty wallet.
- Bittensor SS58 prefix + Blake2b checksum gate done client-side via
  `bs58` + `@noble/hashes` (no `@polkadot/util-crypto` dep added) so a
  Polkadot/Kusama address (shared 32-byte AccountId encoding) is rejected
  before the SCALE storage read.

## consumers unblocked

- **mcp-ts** — `get_sui/ton/trx/xrp/cardano/tao_balance`, `get_trc20/ton_jetton/sui_token_balance`, `get_sui_all_balances`, `get_tron_account_resources` can wrap these instead of re-implementing fetch+parse.
- **agent-backend** — same single typed surface for non-EVM balance reads.

## test matrix

`yarn workspace @vultisig/sdk typecheck` -> exit 0, 0 errors (0 in the new files).
`yarn workspace @vultisig/sdk test` -> 1526 passed, including 16 new cases in `tests/unit/tools/balance/otherBalance.test.ts` (format helper, SS58 fund-safety gate, XRP actNotFound vs error, Tron pre-RPC guard, TRC-20 hex decode, TON state map, Sui native/token split + tokens_unavailable, Cardano lovelace->ADA).

### RECEIPT (live RPC/API, read-only — no signing/broadcast)

```
$ node scripts/receipts/balance_other.mjs
=== sdk.balance.<chain> live receipt ===

[OK] getTonBalance (TON)
{ "chain": "Ton", "ticker": "TON", "address": "EQCD39VS5jcptHL8vMjEXrzGaRcCVYto7HUn4bpAOg8xqB2N",
  "balanceNano": "1592537943858147", "balance": "1592537.943858147",
  "status": "wallet.v3.accountState", "seqno": 343 }

[OK] getTrxBalance (TRON)
{ "address": "TZ4UXDV5ZhNW7fb2AMSbgfAEZ7hWsnYS2g", "balanceSun": 59000, "balanceTrx": "0.059" }

[OK] getTrc20TokenBalance (USDT-TRC20)
{ "address": "TZ4UXDV5ZhNW7fb2AMSbgfAEZ7hWsnYS2g",
  "contractAddress": "TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t",
  "symbol": "USDT", "balance": "0", "decimals": 6 }

[OK] getXrpBalance (XRP)
{ "address": "rEb8TK3gBgk5auZkwc6sHnwrGVJH8DuaLh", "balanceDrops": 99999000, "balanceXrp": "99.999000" }

[OK] getSuiBalance (SUI)
{ "address": "0x00...0005", "chain": "Sui", "ticker": "SUI",
  "balance": "31.018584912", "balanceMist": "31018584912" }

[OK] getCardanoBalance (ADA)
{ "address": "addr1qx2...", "balanceLovelaces": "0", "balanceAda": "0.000000", "nativeTokens": [] }

[OK] getTaoBalance (TAO)
{ "chain": "Bittensor", "address": "5GrwvaEF...HGKutQY",
  "freeRao": "0", "freeTao": "0", "decimals": 9, "note": "Network fee for sends: ~0.0002 TAO" }

=== done (read-only; no signing/broadcast) ===
```


gomesalexandre (commented):
Deep adversarial pass on \`balance.other\` (sui/ton/tron/xrp/cardano/tao + token variants). FOCUS: per-chain address validation/checksum, u64/precision (raw bigint vs JSON-number rounding past 2^53), decimals, wrong-chain confusion. Sibling to the wave-4 u64/address bugs.

## found + fixed

**preferably-blocking: XRP `balanceDrops` u64 precision lost past 2^53** (`otherBalance.ts:72`, sev: high)
`getXrpBalance` parsed the XRPL `Balance` (a decimal **string** on the wire) through `Number()`, which rounds past 2^53 (~9.007e15). XRP supply is ~1e17 drops, so any account >~9M XRP silently lost precision, and `balanceXrp` was derived from the already-lossy float via `.toFixed(6)`. Exact class the wave-4 ports were flagged for.

- Repro: `Number("90000000000123456")` -> `90000000000123460` (corrupted by 4 drops).
- Fix: parse the raw string with `BigInt`, format XRP via bigint arithmetic (mirrors the Cardano ADA path), `balanceDrops` is now the authoritative raw **string**; non-integer `Balance` now throws instead of NaN-ing. Pushed in `a1b72b7`.
- Tests: +`keeps full precision for a >2^53-drop balance`, +`rejects a non-integer Balance`, updated the funded/actNotFound cases for the string type. 18/18 green.

## clean — verified the rest of the surface

- **TRON** — `assertTronAddress` (`/^T[1-9A-HJ-NP-Za-km-z]{33}$/`) gates native + resources + TRC-20 (owner AND contract) before any RPC, rejecting 0x/bech32 (test asserts no RPC fired). TRC-20 balance via `BigInt("0x"+hex)`, decimals/symbol ABI-decoded — raw, no float. **Note (latent, not introduced):** native `getTrxBalance` types `balance` as `number`, so JSON.parse rounds a >9.007B-TRX balance before `BigInt(sun)` — but the wire returns a bare JSON number (confirmed live), so a true fix needs a raw-text parse in the shared `fetchJson`; very-low reachability, left as-is to keep this PR scoped (mcp-ts has the same).
- **Bittensor/TAO** — full SS58 decode: bs58 + 35-byte length + **prefix-42 gate** (rejects Polkadot/Kusama/Acala sharing the 32-byte AccountId) + **blake2b SS58 checksum**. Gate runs BEFORE any RPC (test asserts). free-balance read is u128 LE -> `BigInt`, multi-endpoint fallback. Lossless.
- **TON** nano + jetton, **Sui** mist + `getSuiAllBalances` — all `BigInt(string)`; Sui returns deterministic `tokens_unavailable` (not empty wallet) on RPC error / non-integer balance / typo address; native-vs-token discriminated via the 2 canonical SUI type strings. **Cardano** lovelace->ADA via pure bigint. All raw, no JSON-number path.
- decimals correct per chain (XRP 6, TRX 6, TON 9, SUI 9, TAO 9, ADA 6); no wrong-chain confusion.

typecheck 0 errors · 18/18 unit · live receipt green across all 7 chains (XRP now `balanceDrops:"99999000"` raw string + bigint-formatted `balanceXrp`). Comment-only on my own PR — not approving/merging.

NeOMakinG (approved):
Reviewed across 3 lanes (Claude correctness/fund-safety load-bearing + Codex gpt-5.4 adversarial + GLM — GLM dead 429). I found one **fund-fabrication blocker** in the TRC-20 path, **fixed it in-code** on this branch (`cdcd93d6`) with live proof + mutation-checked tests, and also closed the recurring missing-changeset + missing-RN-export gaps. Approving on the fixed SHA.

## testing approach

Read every chain path against the #1362 B-series balance-fabrication lens (wrong divisor / fail-open zero feeds the agent's fund-decision view). Cross-checked the port against mcp-ts `balance/other-balance.ts`. Drove the TRON path against the **live** `tron-rpc.publicnode.com` node to settle the encoding question with real funded holders. Ran the unit suite (20/20, single-thread minimal in-pkg config — the default config hits the `@vultisig/mpc-types` WASM wall), `tsc` (0 new errors; 29 preexisting `mpc-types`/`mpc-wasm` TS2307 from unbuilt workspace deps), prettier, and the sonarjs-35 complexity gate locally on the fix SHA. CI didn't re-trigger the full suite on the fix-commit (known pattern on these branches) — but the prior SHA `a1b72b7e` had Unit / Integration / Lint+TypeCheck / Test Summary all green, and my change only touches the TRC-20 path (covered by new tests) + a type-only RN re-export + a changeset.

## blocking

**(fixed in-code @cdcd93d6)** `getTrc20TokenBalance` fabricated a zero balance for every funded TRC-20 holder. The owner address was ABI-encoded with `'0'.repeat(24) + address.replace(/^T/, '41')` — a string replace that leaves base58 characters (G/U/Z/R…) in the param, producing a 59-char non-hex string. The TRON node rejects it (`OTHER_ERROR: invalid characters encountered in Hex string`), so the 200 body carries **no** `constant_result`, and `balResp.constant_result?.[0] ?? '0'` then read it as balance `0`.

Live proof:
- `TWd4WrZ9wn84f5x1hZhL4DHvk738ns5jwb` — correct balance **300,000,000 USDT**, SDK reported **0**.
- `TKHuVq1oKVruCGLvqVexFs6dawKv6fQgFs` — correct balance **1,615,944,598 USDT**, SDK reported **0**.

This is the B1 fail-open-zero pattern: a funded wallet reads as empty, hiding funds from the agent's portfolio view. Note this is a faithful port — the same bug ships in production `mcp-ts` `other-balance.ts:433` (worth a follow-up there).

Fix: `tronAddressToAbiParam()` base58check-decodes the address (bs58 was already a dep) to its 21-byte payload (`0x41` + 20-byte EVM-form + 4-byte checksum), validates the prefix/length, and left-pads the 20-byte slice to a 32-byte word. The `balanceOf`/`decimals` reads now **throw** when `constant_result` is absent instead of coercing to `0`. Added two mutation-checked tests (encoding-shape + fail-closed), both verified to fail when the fix is reverted.

## should-fix

- **TRX native precision** — `TrxBalance.balanceSun` is a JS `number` and `response.balance` is consumed as one. TRX supply (71.7B) in sun (1e6) is `7.17e16 > 2^53`, so a whale account (>9M TRX) silently rounds. Because `fetch().json()` already parses to a JS number, fixing this properly needs a raw-text BigInt parse. Faithful to mcp-ts, but a real precision risk on this surface.
- **TON native + jetton fail-open** — `result.balance ?? '0'` / `jetton_wallets?.[0] ?? 0` turn a malformed-but-200 proxy body (or an address-form mismatch that omits the field) into a real-looking zero. Prefer throwing when the expected field is structurally absent.
- **Cardano fail-open** — `infoRes[0]?.balance ?? '0'` / `assetsRes[0]?.asset_list ?? []` read 0 without confirming the echoed `infoRes[0].address` matches the requested address or that the array is non-empty for a known-funded address. Lower blast than TRC-20 (Koios returns only requested addresses and `[]` for genuinely-empty), but worth an address-echo guard.

## q

- **Bittensor `rpcNullable` null → `0n`** — Codex flagged this as fabricating zero on a malformed body. I tier this down: Substrate `state_getStorage` legitimately returns `null` for an absent storage key (a never-funded account), and a malformed no-result/no-error body is indistinguishable from that at the JSON-RPC layer. Standard Substrate semantics — confirm you're comfortable that a (rare) malformed-200 reads as unfunded rather than erroring.
- **`rpc.ts` retries `AbortError`** — `isRetryable` returns true for `AbortError` even though the comment says deterministic timeouts shouldn't retry. With `AbortSignal.timeout(...)` a single timeout fans out to 4 attempts + backoff. Intentional (matches mcp-ts), or should a caller-budget timeout fail fast?

## risk

Low after the fix. Read-only surface — nothing signs or broadcasts. The blocker was a universal fabricated-zero (fund-visibility, not fund-loss); now fixed and proven live. Remaining items are lower-severity fail-opens consistent with the mcp-ts port. RN export is pure-crypto and matches the existing cosmos-staking LCD precedent.

## verdict

APPROVE on `cdcd93d6`. The TRC-20 fabricated-zero blocker is fixed in-code with live evidence and mutation-checked tests; the recurring changeset + RN-export gaps are closed. Per-chain decimals are exact and BigInt-based (XRP 6 / TON 9 / TRX 6 / ADA 6 / SUI 9 / TAO 9), the Bittensor SS58 prefix+checksum gate runs before any RPC (rejects Polkadot/Kusama shared-AccountId confusion), and the address-format guards fail closed.

---
🤖 3-lane review (Claude + Codex gpt-5.4 + GLM). Generated with [Claude Code](https://claude.com/claude-code).

PR #845, opened 2026-06-23
feat(sdk): add sdk.tx.normalize + sdk.tx.split

gomesalexandre:
## what

Two pure, vault-free tx-shape primitives exported from `@vultisig/sdk`:

- **`normalizeTx(result, args)`** — canonicalize a `build_*` tool result into a signing-ready envelope. Wraps a flat result under `tx` + lifts `chain`/`chain_id` to the outer level, leaves an already-nested `swap_tx`/`send_tx`/`tx` untouched (no double-nest), passes `execute_*` prep envelopes through unchanged, and enriches `from_chain`/`to_chain`/`chain` from the originating tool args (with `from_chain` falling back to `chain`). Throws `TxNormalizeError` on a non-routable `execute_*` prep envelope (phantom-card guard, missing `txArgs.tx_encoding`) and on unparseable input.
- **`splitMultiTx(result, args)`** — split a multi-tx build result into an ordered list of per-leg envelopes:
  - approve+swap (`needs_approval === true` + `approval_tx` + `swap_tx`) -> ordered `[approval, swap]`, approval-first
  - generic `transactions[]` (len >= 1) -> per-leg, with parent metadata (`chain`/`provider`/symbols/addresses/decimals) copied onto each leg
  - otherwise a 1-element `[normalizeTx(...)]` no-op

## why

This is the normalize/split half of the agent-backend's Go `enrichBuildResult` (`internal/service/agent/agent.go`) + `splitMultiTx` (`internal/service/agent/tx_sequence.go`), lifted into the SDK so every consumer shares ONE canonical tx-shape contract instead of each re-deriving it.

What deliberately STAYS in the backend (not ported, not pure): SSE `tx_ready` event sequencing, Redis sequence storage/pop/peek/drained-marker lifecycle, `sequence_id`/`sequence_index`/`sequence_total` injection, and the `produces_calldata` dispatch gate. Those are per-turn/per-user runtime + agent-judgement concerns. This module never signs and never broadcasts.

## consumers unblocked

- **mcp-ts** `execute_*` tools — can normalize their build output through one shared helper before emitting
- **SDK CLI executor** (`storeServerTransaction` reads `.swap_tx || .send_tx || .tx`) — split legs are already in the shape it routes on
- **Windows extension / Station** — same canonical envelope for approve+swap and generic multi-tx flows

## receipt

```
==== 1. normalizeTx(flat build_evm_tx result) ====
output: {
  "tx": { "to": "0xA0b8...eB48", "value": "0x0", "data": "0xa9059cbb", "gas": "0x5208", "chain": "Ethereum", "chain_id": "1" },
  "chain": "Ethereum",
  "chain_id": "1",
  "from_chain": "Ethereum"
}

==== 2. normalizeTx(execute_* prep envelope) ====
output: { "txArgs": { "tx_encoding": "evm", ... }, "stepperConfig": { "steps": ["sign","broadcast"] }, "chain": "Ethereum" }
  (NOT double-wrapped under tx; tx_encoding discriminator preserved)

==== 3. splitMultiTx(approve + swap build_swap_tx result) ====
split into 2 ordered legs:
  leg 0 = approval (tx):  { "tx": { "to": "0xA0b8...eB48", "data": "0x095ea7b3" }, "provider": "oneinch", "chain": "Ethereum", "from_symbol": "USDC", ... }
  leg 1 = swap (swap_tx): { "swap_tx": { "to": "0x1111...2A65", "data": "0x12aa3caf" }, "provider": "oneinch", "chain": "Ethereum", "to_symbol": "ETH", ... }

==== 4. splitMultiTx(generic transactions[] array) ====
split into 2 legs:
  leg 0: { "tx": { "to": "0xmorpho", "data": "0x0001" }, "chain": "Base", "provider": "morpho" }
  leg 1: { "tx": { "to": "0xmorpho", "data": "0x0002" }, "chain": "Base", "provider": "morpho" }

All receipt assertions passed. (pure shape transforms — no signing, no broadcast)
```

Run it yourself: `node scripts/receipts/tx_normalize.mjs`

## testing

- `yarn workspace @vultisig/sdk test` — 1527 passed (incl. new `tests/unit/tx/normalize.test.ts`, both functions + every branch: flat-wrap, nested-passthrough, execute_* prep pass + phantom-card throw, from_chain fallback, no-overwrite, raw-string transport, no-mutation, approve+swap ordering, `needs_approval:false` no-split, generic array len>=1, malformed-JSON throw)
- `yarn workspace @vultisig/sdk typecheck` — 0 errors (these files add 0 new)
- lint + prettier clean

## Summary by CodeRabbit

* **New Features**
  * Added transaction normalization utilities to the SDK for standardizing transaction payload shapes
  * Support for splitting multi-leg transactions (e.g., approval + swap scenarios) into separate normalized legs
  * New error handling for malformed or invalid transaction inputs

### DIFF
--- a/packages/sdk/src/tx/normalize.ts
+++ b/packages/sdk/src/tx/normalize.ts
@@ -1,0 +1,233 @@
+/**
+ * Pure tx-shape normalization + multi-tx splitting.
+ *
+ * Ports the *normalize/split* half of the agent-backend's
+ * `enrichBuildResult` (internal/service/agent/agent.go) and `splitMultiTx`
+ * (internal/service/agent/tx_sequence.go) into a vault-free, side-effect-free
+ * SDK primitive.
+ *
+ * What stays in the backend (NOT ported here):
+ *   - SSE event sequencing (tx_ready emission ordering, pendingTxReadyEvents)
+ *   - Redis sequence storage / pop / peek / drained-marker lifecycle
+ *   - sequence_id / sequence_index / sequence_total injection (per-turn,
+ *     per-user, Redis-scoped — runtime concern, not a pure shape transform)
+ *   - the produces_calldata dispatch gate + any agent-judgement
+ *
+ * This module is PURE: it takes an already-built tool result (JSON) and
+ * returns a canonical tx envelope (`normalizeTx`) or an ordered list of
+ * per-leg envelopes (`splitMultiTx`). It never signs, broadcasts, or reads
+ * chain state.
+ */
+
+/** A parsed JSON object whose values are themselves arbitrary JSON. */
+type JsonObject = Record<string, unknown>
+
+/**
+ * A canonical, signing-ready tx envelope. Always carries the inner tx under one
+ * of `tx` / `swap_tx` / `send_tx`, plus optional chain-routing metadata copied
+ * from the build args / parent envelope. Extra fields ride along unchanged.
+ */
+export type NormalizedTx = JsonObject & {
+  tx?: unknown
+  swap_tx?: unknown
+  send_tx?: unknown
+  chain?: string
+  chain_id?: string
+  from_chain?: string
+  to_chain?: string
+}
+
+/**
+ * Chain-routing args from the originating `build_*` tool call. Mirrors the
+ * `{from_chain, to_chain, chain}` probe in Go `enrichBuildResult`. All optional
+ * — a flat single-chain `build_evm_tx` only carries `chain`.
+ */
+export type NormalizeArgs = {
+  from_chain?: string
+  to_chain?: string
+  chain?: string
+}
+
+/** Thrown when a build result can't be parsed into a tx envelope. */
+export class TxNormalizeError extends Error {
+  override readonly name = 'TxNormalizeError'
+
+  constructor(message: string) {
+    super(message)
+  }
+}
+
+const isJsonObject = (v: unknown): v is JsonObject => typeof v === 'object' && v !== null && !Array.isArray(v)
+
+/**
+ * Parse a build result into a plain object. Accepts either a JSON string (the
+ * raw MCP tool result, mirroring the Go `result string` input) or an
+ * already-parsed object. Throws `TxNormalizeError` on anything that isn't a
+ * JSON object (arrays, primitives, malformed JSON) — the Go code returned nil
+ * here; in TS we surface it so callers don't silently drop a tx.
+ */
+const parseResult = (result: string | JsonObject): JsonObject => {
+  if (isJsonObject(result)) return { ...result }
+  let parsed: unknown
+  try {
+    parsed = JSON.parse(result)
+  } catch (e) {
+    throw new TxNormalizeError(`build result is not valid JSON: ${(e as Error).message}`)
+  }
+  if (!isJsonObject(parsed)) {
+    throw new TxNormalizeError('build result is not a JSON object')
+  }
+  return parsed
+}
+
+/**
+ * `execute_*` prep envelopes (mcp-ts#49) are self-describing via `txArgs` +
+ * `stepperConfig` and dispatch on the inner `txArgs.tx_encoding` discriminator.
+ * Wrapping them under `tx` would double-nest and break consumers that switch on
+ * `txArgs.tx_encoding`, so the wrap step is skipped for them.
+ */
+const isExecutePrep = (txMap: JsonObject): boolean => 'txArgs' in txMap && 'stepperConfig' in txMap
+
+/**
+ * Phantom-card guard (vultiagent-app#603 part B): an `execute_*` prep envelope
+ * whose `txArgs` lacks a `tx_encoding` discriminator is structurally malformed —
+ * the client `parseServerTx` cannot route it to any signing branch. The Go code
+ * returned nil here so tx_ready is never emitted; we throw so the caller can
+ * count + drop it explicitly.
+ */
+const assertExecutePrepRoutable = (txMap: JsonObject): void => {
+  const txArgs = txMap['txArgs']
+  const encoding = isJsonObject(txArgs) ? txArgs['tx_encoding'] : undefined
+  if (typeof encoding !== 'string' || encoding === '') {
+    throw new TxNormalizeError('execute_* prep envelope missing txArgs.tx_encoding discriminator (phantom card)')
+  }
+}
+
+/**
+ * Normalize a `build_*` tool result into a canonical tx envelope.
+ *
+ * Mirrors Go `enrichBuildResult`:
+ *   - If the result already carries a nested tx (`swap_tx` / `send_tx` / `tx`),
+ *     or is an `execute_*` prep envelope, it is enriched in place.
+ *   - Otherwise the whole result IS the transaction and gets wrapped under
+ *     `tx` (with `chain` / `chain_id` lifted to the outer level so chain
+ *     resolution keeps working).
+ *   - Chain-routing fields (`from_chain` / `to_chain` / `chain`) are filled
+ *     from `args` when the payload doesn't already carry them.
+ *
+ * Pure: no I/O, no signing, no broadcast. Returns a fresh object.
+ *
+ * @throws {TxNormalizeError} on unparseable input or a non-routable
+ *   `execute_*` prep envelope (phantom card).
+ */
+export const normalizeTx = (result: string | JsonObject, args: NormalizeArgs = {}): NormalizedTx => {
+  let txMap = parseResult(result)
+
+  const executePrep = isExecutePrep(txMap)
+  if (executePrep) {
+    assertExecutePrepRoutable(txMap)
+  }
+
+  const hasNestedTx = 'swap_tx' in txMap || 'send_tx' in txMap || 'tx' in txMap
+
+  // Flat build_* result with no nested tx and not an execute_* prep envelope:
+  // the whole result IS the transaction, wrap it under "tx" and lift chain
+  // metadata to the outer level for downstream chain resolution.
+  if (!executePrep && !hasNestedTx) {
+    const wrapped: JsonObject = { tx: { ...txMap } }
+    if ('chain' in txMap) wrapped['chain'] = txMap['chain']
+    if ('chain_id' in txMap) wrapped['chain_id'] = txMap['chain_id']
+    txMap = wrapped
+  }
+
+  // Enrich with chain-routing info from the originating tool-call args when the
+  // payload doesn't already carry it. from_chain falls back to chain (matches
+  // Go: single-chain ops set from_chain = chain).
+  const { from_chain: argFrom, to_chain: argTo, chain: argChain } = args
+  if (!('from_chain' in txMap)) {
+    if (argFrom) txMap['from_chain'] = argFrom
+    else if (argChain) txMap['from_chain'] = argChain
+  }
+  if (!('chain' in txMap) && argChain) txMap['chain'] = argChain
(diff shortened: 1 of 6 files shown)


gomesalexandre (commented):
Deep adversarial pass on \`normalizeTx\` / \`splitMultiTx\` vs the Go source (\`enrichBuildResult\` in \`agent.go:7898\`, \`splitMultiTx\` + \`wrapSingleTx\` in \`tx_sequence.go:120/184\`). The vs-source audit was light for this one, so I went line-by-line on the fund-safety-adjacent invariants: multi-tx split ordering, per-leg metadata copy, leg drop/reorder, and the normalize wrapping.

**Verdict: adversarial pass clean — faithful port, no logic fix needed.** Verified, with empirical checks via tsx:

- **Approval-before-swap ordering** — Pattern 1 returns \`[approvalLeg (tx), swapLeg (swap_tx)]\`, approval-first, matching Go + the \`sequenceTxReady\` \`txs[1:]\` stash contract. No reorder.
- **Pattern 1 precedence over Pattern 2** — when both \`needs_approval\`+\`approval_tx\`+\`swap_tx\` AND a \`transactions[]\` array co-exist, the approve+swap split wins (checked first, exactly like Go). The generic array is never wrongly wrapped.
- **Strict \`needs_approval === true\`** — string \`"true"\` / number \`1\` do NOT trigger a split (mirrors Go's strict bool unmarshal). No fabricated approval leg.
- **No leg dropped/reordered** — 3+ element \`transactions[]\` preserves index order 0,1,2 with parent metadata on every leg.
- **Per-leg metadata copy** — \`LEG_METADATA_KEYS\` is the same 11-key list (chain/chain_id/from_chain/to_chain/provider/from_symbol/to_symbol/from_address/to_address/from_decimals/to_decimals) as Go's \`metadataKeys\`; control flags (\`needs_approval\`/\`approval_tx\`) correctly do NOT leak onto legs.
- **Flat-wrap path** — non-nested \`build_*\` results wrap under \`tx\` with \`chain\`/\`chain_id\` lifted outward, then \`from_chain\`/\`chain\`/\`to_chain\` enriched from args with the from_chain→chain fallback — byte-for-byte the Go enrich order.
- **execute_* prep + phantom-card guard** — \`txArgs\`+\`stepperConfig\` skips the wrap; missing \`tx_encoding\` throws (Go returned nil → tx_ready never emitted). Equivalent fail-closed.

One intentional divergence noted, not a regression: SDK throws \`TxNormalizeError\` on a malformed JSON *string* where Go passes the original RawMessage through (Go's input is always valid-JSON by construction). Fails loud/closed, not silent — fine.

**Changed (test-only, \`bacb97f8\`):** added 4 regression locks for the above invariants (Pattern-1-precedence, strict-bool, 3-leg order/metadata, no-control-flag-leak) — \`packages/sdk/tests/unit/tx/normalize.test.ts\`. These are the shapes a future refactor (truthy check, reordered patterns, widened metadata copy) would silently break.

**Receipts:** 20/20 unit (16 existing + 4 new) green, \`tsc --noEmit\` 0 errors, \`scripts/receipts/tx_normalize.mjs\` all assertions pass (approve+swap split + generic transactions[] split printed correctly).

packages/sdk/src/tx/normalize.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  // Pattern 1: swap with approval.
+  if (parsed['needs_approval'] === true && 'approval_tx' in parsed && 'swap_tx' in parsed) {
+    return [wrapSingleTx('tx', parsed['approval_tx'], parsed), wrapSingleTx('swap_tx', parsed['swap_tx'], parsed)]
+  }
+
+  // Pattern 2: generic transactions array (len >= 1).
+  const txs = parsed['transactions']
+  if (Array.isArray(txs) && txs.length > 0) {
+    return txs.map(child => wrapSingleTx('tx', child, parsed))
```
NeOMakinG, line 228: `suggestion:` (Codex, downgraded from should-fix) split wraps each `transactions[]` child verbatim under `tx` without inspecting its shape. If a future tool regresses and emits a child that is itself an envelope (`{swap_tx:...}`, `{txArgs:{tx_encoding}}`) or a nested `{transactions:[...]}`, the leg becomes `{tx:{swap_tx:...}}` — value-moving content is *preserved* (not dropped, so not the #848 class), but sits one level deeper than a downstream parser keying on `tx.to`/`swap_tx` expects, so it could read as un-signable rather than rejected. This is faithful Go-parity (`wrapSingleTx` does the same), so not blocking — but as this is now a public primitive, consider a fail-closed guard or explicit recursion for non-flat children, mirrored on the Go side.

packages/sdk/src/tx/normalize.ts
```diff
@@ -0,0 +1,233 @@
...
+    assertExecutePrepRoutable(txMap)
+  }
+
+  const hasNestedTx = 'swap_tx' in txMap || 'send_tx' in txMap || 'tx' in txMap
+
+  // Flat build_* result with no nested tx and not an execute_* prep envelope:
+  // the whole result IS the transaction, wrap it under "tx" and lift chain
+  // metadata to the outer level for downstream chain resolution.
+  if (!executePrep && !hasNestedTx) {
+    const wrapped: JsonObject = { tx: { ...txMap } }
```
NeOMakinG, line 137: `suggestion:` (Codex, downgraded from should-fix) the flat-wrap path promotes any non-nested JSON object into `{tx:{...}}` — `{foo:'bar'}` normalizes 'successfully'. This is by-design Go-parity ('the whole result IS the transaction') and the malformed-JSON path *does* fail-closed (verified: bad JSON/array/primitive all throw). A structural `tx.to`-style guard here would diverge from Go and risk rejecting valid flat builds on chains that don't carry `to` (UTXO/Cosmos), so I'd keep this as-is and let decode/#848 do the shape validation — flagging only so the trade-off is a conscious one.

NeOMakinG (approved):
## testing approach

Three-lane review (Claude load-bearing + Codex gpt-5.3-codex + GLM). GLM dead (ECONNRESET) — best-effort skipped. Reviewed against Go reference `enrichBuildResult` (internal/service/agent/agent.go:7842) + `splitMultiTx`/`wrapSingleTx` (internal/service/agent/tx_sequence.go) in agent-backend @9e50093e.

Fresh worktree `yarn install` (canonical node_modules symlink breaks vitest); ran `tests/unit/tx/normalize.test.ts` via a package-local minimal vitest config (the workspace `tests/integration/setup.ts` imports unbuilt `@vultisig/mpc-types` dist — a pre-existing wall unrelated to this PR). Ran the receipts script, complexity gate, prettier, tsc, plus adversarial probes + 4 source mutations.

- **normalize test**: 20/20 pass
- **receipts** (`node scripts/receipts/tx_normalize.mjs`): all assertions pass, exit 0
- **mutations (all caught — tests have teeth, not #228-inert)**: drop-last-leg → 3 fail; reverse approval/swap order → 3 fail; alter `value` in wrap → 3 fail; fail-open on malformed (throw→return {}) → 1 fail; restored → 20/20
- **fund-field probes**: amount/value preserved as strings (no float coercion), idempotent `normalize(normalize(x))==normalize(x)`, no input mutation, `__proto__` key does not pollute Object.prototype
- **fail-closed**: `{not json` / `[1,2,3]` / `"string"` / `null` / `42` all throw `TxNormalizeError`; bad split string throws
- **gates**: complexity (`run-eslint.mjs --config .config/eslint-health.config.mjs`) exit 0; prettier `--check` clean; tsc adds 0 new errors (all TS2307 are pre-existing unbuilt `@vultisig/mpc-types`/`mpc-wasm`); `gh pr checks` — Unit / Integration / Lint+Type / CodeRabbit all green

## blocking

None.

## preferably-blocking

None.

## should-fix

None.

## suggestion

Codex flagged two hardening opportunities (both downgraded from should-fix after source re-read — see q below for why they are not blocking). Surfaced inline as suggestions: split wrapping a non-flat `transactions[]` child verbatim, and normalize wrapping any arbitrary object under `tx`.

## q

Codex asks whether exact Go-parity is a hard requirement here, given the SDK export broadens the blast radius vs the backend-internal original. My read: the two flagged behaviors (split wraps non-flat children one level deep under `tx`; normalize wraps any non-nested object under `tx` as 'the whole result IS the tx') are **faithful Go-parity** — Go's `wrapSingleTx` and `enrichBuildResult` do exactly the same with no child-shape validation or recursion. Crucially this is **preserve-not-drop**: unlike #848 (which returned only the first inner multicall call and *dropped* the rest), split here keeps every leg's payload verbatim — I probed a child carrying a nested `calls:[decoy, attacker]` array and the attacker call is preserved intact inside the leg's `tx`, not lost. The decode/policy seam (#848/#847) is where structural tx-shape validation belongs; pushing a 'looks-like-a-tx' guard up into normalize risks rejecting valid flat builds for non-EVM chains (UTXO/Cosmos don't carry `to`). So I'd keep parity and harden at decode, but worth a deliberate decision since this is now a public primitive.

## risk

Low. Pure, vault-free, side-effect-free shape transforms — never signs/broadcasts/reads chain state. Lossless on fund-binding fields, fails-closed on malformed (stricter than Go: throws where Go returned nil), and a faithful Go port verified line-for-line. The #848-class drop/hide risk does not materialize: split surfaces all value-moving legs in order with no merge/dedup/drop. Downstream decode (#848, with its multi-value-moving sentinel) and policy (#847) remain the structural-validation gate; this module's contract is shape-canonicalization only, which it meets.

## verdict

APPROVE. Lossless normalize, complete split (no leg drop/hide/reorder — #848-class risk does not apply), fail-closed on malformed, faithful Go-parity, tests with proven mutation teeth, and all CI (incl. complexity + prettier) green.

---
🤖 Reviewed with Claude + Codex (gpt-5.3-codex). GLM lane best-effort (down).

NeOMakinG (approved):
## testing approach

Re-confirm pass (round 2) after the HEAD advance `bacb97f8` → `f967510d`. This is **not** a fresh review — it re-validates the prior APPROVE (`@bacb97f8`, review 4557...) against the new HEAD.

The advance was a forward-merge of `main` (`chore(merge): sync main after #848 + #861`), pulling in #846/#847/#848 plus rcoderdev's #807/#808/#809 native-swap changes + version-packages. The question for re-confirm: did the merge touch #845's own logic?

- **own-files diff** — `git diff bacb97f8 f967510d -- packages/sdk/src/tx/ packages/sdk/tests/unit/tx/` is **empty**. `sdk.tx.normalize` + `sdk.tx.split` source and tests are byte-identical to the SHA I approved. No re-review of the tx logic is warranted; the prior verdict (normalize losslessness + split preserve-not-drop) stands unchanged.
- **CI** — all green at HEAD (Unit, Integration/Vitest, Lint+TypeCheck, Test Summary, CodeRabbit). The tx unit suite is covered by the green Unit Tests job at `f967510d`.

## prior issues — status

| 1 | — | — | (no blocking/should-fix findings; prior pass was a clean APPROVE) | ✅ | nothing to carry; re-confirm only |

## blocking
none.

## should-fix
none.

## risk
none introduced by the forward-merge — #845's own tx surface is untouched. Merge brought only unrelated main commits already reviewed in their own PRs.

## verdict
APPROVE (re-confirm at HEAD `f967510d`). Prior approval stands after a clean `main` forward-merge with no change to `tx.normalize`/`tx.split` logic.

## lanes
Claude load-bearing. Codex/GLM best-effort, skipped for a no-op-merge re-confirm (empty own-files diff = no analytical surface). 3-lane spirit satisfied.

---
🤖 reviewed by vultisig-ops reviewer

PR #847, opened 2026-06-23
feat(sdk): add sdk.policy.evaluate(claim, envelope) -> Verdict

gomesalexandre:
## what

Ports the PURE intent<->envelope field-diff from the Go reference `internal/safety/policy.go` (`EvaluatePolicy` + `FieldDiff` + `CheckInvariants`) into the SDK as a vault-free comparison primitive, surfaced as the `policy` namespace:

```ts
import { policy } from '@vultisig/sdk'

policy.evaluate(claim, envelope)         // -> Verdict  (first blocking decision: PASS | WARN | BLOCK + field diffs)
policy.checkInvariants(invariantInput)   // -> InvariantViolation[]  (every independent I1-I7 violation)
```

Given a structured `IntentClaim {chain, recipient, amount, asset}` and an already-decoded `Envelope`, it computes a field-level `Diff` (chain/recipient/amount/asset mismatch) and asserts the I1-I7 fund-safety invariants.

## why

This is the bare struct-diff the canonical-decoder analysis calls out as "free once `sdk.decode` exists" - the pure-crypto comparison sub-layer of the safety stack. A plain wallet / CLI doing a manual tx wants this exact drift-check on what it is about to sign, with zero agent context. It lifts the comparison logic out of the Go backend so the app card, CLI text-WYSIWYS, and co-sign gate can all diff a claim against a decoded tx through ONE canonical implementation instead of three that drift.

Strictly PURE comparison - it diffs two structured objects and nothing more:
- does NOT extract intent from a natural-language message (agent judgement stays in the backend)
- does NOT decode wire bytes (that is the separate `sdk.decode.fromToolResult` keystone)
- NEVER signs or broadcasts

Faithful 1:1 port: same Slice-0 chain alias map, same units-aware amount-interpretation ladder (closes the 1-wei false-negative sink - `"1"` ETH human vs a 1-wei envelope now BLOCKs), same fail-open-on-undecoded WARN, same exact `bigint` decimal scaling (no float precision loss).

## consumers unblocked

- mcp-ts `hostValidate` / app decoded-intent card / CLI WYSIWYS - one bytes-vs-claim oracle instead of per-surface re-implementations
- agent-backend - consumes the same TS verdict over the boundary (kills the Go<->TS drift the analysis flags)
- the migration shadow-diff (old vs new on the SAME diff -> converges)

The co-sign spend-cap / allowlist / velocity thresholds are PRODUCT POLICY and explicitly stay OUT of the generic SDK (`@vultisig/policy`); this PR ships only the chain/recipient/amount/asset struct-diff + invariants.

## test / receipt

- `packages/sdk/tests/unit/tools/policy/policy.test.ts` - 22 unit tests (mismatch BLOCK, clean PASS, case-insensitive recipient, chain-alias, drift WARN/BLOCK, asset-symbol WARN, fail-open, 1-wei sink, all 7 invariants, amount helpers). Full SDK unit suite stays green (1532 passed).
- `yarn workspace @vultisig/sdk typecheck` -> 0 errors (0 new from this PR).
- Runnable receipt `packages/sdk/scripts/receipts/policy_evaluate.mjs` (self-verifying, exits non-zero on regression, NO broadcast):

```
=== sdk.policy.evaluate(claim, envelope) -> Verdict ===

[1] MISMATCH  claim {send 1 USDC -> 0xAAA}  vs  envelope {0xBBB, 1 USDC}
    result : BLOCK
    reason : recipient mismatch: user claimed "0xaaa", envelope has "0xbbb"
    diff   : [{"field":"recipient","claimed":"0xaaa","observed":"0xbbb"}]

[2] MATCH     claim {send 1 USDC -> 0xAAA}  vs  envelope {0xAAA, 1 USDC}
    result : PASS
    reason : envelope matches intent claim
    diff   : []

[3] DRIFT     claim {send 1 USDC -> 0xAAA}  vs  envelope {0xAAA, 2 USDC}
    result : BLOCK
    reason : amount drift 100.00% exceeds block threshold 1%

[4] INVARIANTS  claim {1 USDC -> 0xAAA on ethereum}  vs  envelope {0xBBB, base}
    violations : I3_chain_matches_intent, I1_recipient_matches_intent

OK — all receipt assertions passed (PURE comparison, no signing/broadcast).
```

Run: `../../node_modules/.bin/tsx scripts/receipts/policy_evaluate.mjs` from `packages/sdk`.

## Release Notes

* **New Features**
  * Added policy evaluation for transaction intents returning Pass/Warn/Block verdicts
  * Introduced fund-safety invariant checks for recipient, chain, amount, and memo fields
  * Added amount drift detection with configurable thresholds
  * Expanded SDK public API with amount parsing, chain aliasing, and policy utilities

* **Tests**
  * Added comprehensive unit test coverage for policy evaluation and invariant validation


gomesalexandre (commented):
## deep adversarial port-review — policy.evaluate vs Go `internal/safety/policy.go` @ `b6956d75c`

Ran the full port diff against the Go reference at its current HEAD. **The policy logic is a faithful 1:1 port — adversarial pass clean, no production-logic fix needed.** Only addition is test hardening (pushed `fb087ba5`).

### what was checked (line-by-line vs Go)
- **`chainsMatch` + `chainAliasMap`** (`chains.ts`) — byte-exact: same early `a===b`, same `?? a` fallback (mirrors Go's `if canonA=="" { canonA=a }`), all **12 alias entries identical** (eth→ethereum, bnb/bsc→bsc, avax→avalanche, cosmos→cosmoshub-4, terra→phoenix-1, terraclassic/lunc→columbus-5, osmosis→osmosis-1, noble→noble-1, dydx→dydx-mainnet-1, akash→akashnet-2).
- **`evaluatePolicy`** (`evaluatePolicy.ts`) — exact: fail-open-on-undecoded → WARN; chain BLOCK **before** recipient **before** amount ordering; min-drift-across-interpretations; thresholds warn 0.1% / block 1%; asset display-name → WARN. Matches `EvaluatePolicy` L91-191.
- **`checkInvariants` I1-I7** (`checkInvariants.ts`) — exact: I3/I1 empty-decoded-field strictness (a dropped chain/recipient is a divergence, not just a non-alias mismatch); I2 zero/nonzero switch + all-interpretations-drift gate; I4 sign-without-confirm; I5 `> balance`; I7 memo drop/rewrite verbatim; I6 provenance (`invariantI6Diffs`) sanitized-string amount compare so `1,000`==`1000` but any rewrite/clear fires. Ordering (I7 before I6) matches Go.
- **amount units-ladder** (`amount.ts`) — exact: `PLAUSIBLE_TOKEN_DECIMALS=[6,8,9,12,18]`, `scaleDecimalClaimToAtomic` exact bigint string arithmetic (no floats, frac>decimals → null), `claimInterpretations` honors `base` / `human+known` / `human+unknown`→ladder / `''`→ladder, `isZeroAmount` strconv-ParseFloat-equivalent regex, `parseAmountBig` rejects floats.
- **recipient/spender diff** — case-insensitive lowercase compare both sides (EVM checksum drift not a mismatch), BLOCK only when both non-empty in `evaluate`, strict empty-decoded-field in I1.

### the one caveat (NOT a defect in this PR)
The audit-flagged numeric-vs-symbolic chain issue (`chainsMatch("base","8453")` → false → spurious BLOCK on typed-EVM txs) is a defect in **#848's decoder** (`decodeEvmTx` emitting `String(tx.chainId)` instead of resolving via an `evmChainIdToSymbol` map like Go's `populateFromEVMTx`), **not** in #847. From #847's side this is **fail-SAFE** (errs to BLOCK, never a silent pass), so the policy is correct as-is; #848 must land the symbolic resolver.

### fail-closed / affiliate
- **Fail-closed confirmed**: undecoded → WARN (no PASS); empty decoded fields treated as divergence in I1/I3; un-resolved numeric chain id → BLOCK not pass.
- **No affiliate surface** in this PR — pure intent↔envelope comparison, no quote/route/fee construction, no treasury constants. Nothing to leak.

### hardening added (`fb087ba5`, +59 LOC tests only, no logic change)
- pins the **symbolic-chain contract with #848**: symbolic↔symbolic PASSes; an un-resolved numeric id (`8453`/`1`) → BLOCK (catches a decoder regression policy-side).
- pins **`chainAliasMap` verbatim** to the Go table → any cross-surface alias drift is a test failure.

### receipts
- `tsx scripts/receipts/policy_evaluate.mjs` → all 4 cases pass (recipient BLOCK, match PASS, 100% drift BLOCK, I1+I3 invariants).
- `vitest tests/unit/tools/policy/policy.test.ts` → **26 passed** (22 existing + 4 new).
- `tsc --noEmit` → **0 errors**.

packages/sdk/src/tools/policy/evaluatePolicy.ts
```diff
@@ -0,0 +1,113 @@
...
+      reason: `chain mismatch: user claimed "${claimChain}", envelope has "${envChain}"`,
+      diff: [{ field: 'chain', claimed: claimChain, observed: envChain }],
+    }
+  }
+
+  // Recipient check: BLOCK on mismatch when both are non-empty. EVM addresses
+  // are case-insensitive; Cosmos bech32 is lowercase by convention.
+  const claimRecipient = lower(claim.recipient)
+  const envRecipient = lower(envelope.recipient)
+  if (claimRecipient !== '' && envRecipient !== '' && claimRecipient !== envRecipient) {
```
NeOMakinG, line 56: `q:` (fail-open asymmetry — faithful to Go, but a consumer footgun) — `evaluatePolicy` only BLOCKs recipient when **both** sides are non-empty. If the decoded envelope **dropped** the recipient (empty) while the claim states one, this returns PASS, not BLOCK. Same for a dropped chain (L44) and a zero/nil envelope amount (L68). This matches the Go `EvaluatePolicy` exactly and is intentional (first-blocking triage), but it means **`evaluate` alone is fail-open on dropped/zero fields** — only `checkInvariants` (I1/I3/I2) catches them, fail-closed. Since the PR pitches this at "a plain wallet / CLI", please document in the `policy` namespace that `checkInvariants` is the fail-closed gate and `evaluate` is triage — otherwise a naive consumer will wire `evaluate` as the allow/deny gate and ship a fail-open path.

packages/sdk/src/tools/policy/evaluatePolicy.ts
```diff
@@ -0,0 +1,113 @@
...
+      if (drift > AMOUNT_DRIFT_WARN_PCT) {
+        diffs.push({ field: 'amount', claimed: claim.amount ?? '', observed: envAmount.toString() })
+      }
+    }
+  }
+
+  // Asset symbol check: WARN on mismatch (could be display-name drift).
+  const claimAsset = upper(claim.asset)
+  const envAsset = upper(envelope.asset?.symbol)
+  if (claimAsset !== '' && envAsset !== '' && claimAsset !== envAsset) {
```
NeOMakinG, line 96: `should-fix:` token **contract** is never compared — only `asset.symbol` (and only to WARN). The envelope carries `asset.contract` but `IntentClaim` has no contract field, so a tx to a fake contract with a matching symbol (e.g. a counterfeit "USDC") never BLOCKs here. Faithful to the Go `IntentClaim` and out of scope for the pure diff (contract/calldata allowlisting = the `sdk.decode` + security seam), so not a blocker — but a one-line doc note that contract identity is **not** part of this verdict would prevent consumers from assuming it is.

NeOMakinG (approved):
## testing approach

Worktree `/tmp/wt-sdk-847` off `feat_sdk_policy_evaluate`. Read all six policy source files + tests + receipt, then diffed the TS port **line-by-line against the Go reference** `agent-backend/internal/safety/policy.go` (the canonical production safety impl). Ran the policy unit suite (`vitest`, single-threaded), the self-verifying receipt, and adversarial `tsx` probes (dropped/zero fields, negative/sci-notation/hex amounts, min-drift-ladder bridging, I6 provenance formatting). Mutation-tested every BLOCK invariant (recipient/chain/amount + I2 drift) to confirm test teeth. Reproduced and **fixed the red CI gate** in-code.

Lanes: Claude load-bearing. Codex (gpt-5.3-codex) stalled 0-byte under concurrent load; GLM dead (ECONNRESET) — both second-opinion lanes best-effort down, noted.

## what I verified

- **Port fidelity (the bar for this PR):** the TS is a faithful 1:1 of the Go `EvaluatePolicy` / `CheckInvariants` / `invariantI6Diffs` / amount helpers — identical guards, branch order, reason strings, and the 12-entry `chainAliasMap` (diff'd byte-for-byte).
- **Fail-closed:** the gate is fail-**closed via `checkInvariants`** (the exhaustive I1–I7 oracle): dropped recipient → I1, dropped chain → I3, zero-envelope-vs-non-zero-claim → I2 (all probe-confirmed firing). See the `q` on `evaluatePolicy` below for the important caveat.
- **Amount integrity:** exact `bigint` string scaling (no float in the scaling path); the `Number()` ratio in `amountDriftPct` matches Go's `big.Float→Float64` and its ~2⁻⁵³ error is irrelevant vs a 1% threshold. The min-drift ladder does **not** bridge a 10× mutation (probed: closest rung drifts 90% → BLOCK). 1-wei sink closed.
- **Tests:** 26 → **28** (I added the two missing `checkInvariants` I2 tests). Mutation-proven teeth on recipient/chain/amount/I2-drift BLOCK paths. Receipt green.

## prior issues — status

_n/a — first review._

## blocking

**(fixed in-code — fix-and-approve)** CI `quality:complexity` (`sonarjs/cognitive-complexity`, threshold 35; lives in `.config/eslint-health.config.mjs`, run by `quality:complexity`, **not** the `yarn lint` config — which is why lint looked green locally) failed on `checkInvariants.ts:26` at complexity **37**. This makes the PR CI-red and unmergeable. Reproduced exactly, then fixed in commit `1c4d51f9` by extracting the I2 amount oracle into a pure `checkAmountInvariant` helper. Behaviour is byte-identical to the inlined Go port (verified via adversarial probes + a mutation that flips the drift-return — the new drift test fails, proving teeth). Complexity gate now `EXIT 0`; lint/format/28-test/receipt all green.

## should-fix

**Token contract is never cross-checked — only the symbol (WARN-only).** The `Envelope` carries `asset.contract` (`types.ts:99`) but `IntentClaim` has no contract field, so an envelope with a matching symbol (`USDC`) and a *different/fake contract address* passes as PASS/WARN, never BLOCK. This is faithful to the Go `IntentClaim` (no contract field there either), and contract/calldata allowlisting is explicitly the separate `sdk.decode` + security seam — so not a port regression and not a blocker. But for a surface labelled a fund-safety verdict gate and aimed at "a plain wallet / CLI", a one-line note in the `policy` namespace docs that contract identity is **not** part of this diff would close a real expectation gap.

## q

1. `checkInvariants` I2 drift branch (`!claimZero && envZero` and the `drifted === interps.length` return) had **zero** test coverage before this PR — neutralizing the drift return kept the suite green (#228-inert). I added two I2 tests (drift>1% under all interpretations; non-zero-claim/zero-envelope) in `1c4d51f9`. Flagging so it's on record that the oracle's I2 path is now covered.
2. A negative claim amount (`"-1"`) yields no positive interpretation, so the amount check is skipped → PASS. Nonsensical input and faithful to Go (`add` requires `Sign() > 0`), but worth confirming no caller can feed an attacker-controlled negative claim into `evaluate` expecting it to constrain the envelope.

## risk

Low. Pure comparison primitive, no signing/broadcast/network/vault. The verdict gate is fail-closed **through `checkInvariants`**; the `evaluatePolicy`-alone fail-open-on-dropped-field asymmetry is faithful to the production Go and documented (the inline `q` records the consumer caveat). The only CI-blocking issue (complexity gate) is fixed and pushed.

## verdict

**APPROVE** (fix-and-approve). Faithful Go port, fail-closed via the invariant oracle, sound amount integrity, mutation-proven deny coverage. The CI complexity blocker is fixed in-code (`1c4d51f9`); the contract-symbol gap and consumer fail-open caveat are documented, not blocking.

---
🤖 Reviewed with Claude Code (Opus 4.8). Codex + GLM lanes were down (best-effort); Claude lane load-bearing.

PR #848, opened 2026-06-23
feat(sdk): add sdk.decode.fromToolResult (canonical calldata -> Envelope)

gomesalexandre:
## what

THE keystone bytes oracle: one decoder that turns a pending tx into a chain-agnostic `Envelope`:

```ts
type Envelope = {
  chain: string          // EVM: numeric chain id from the wire; Cosmos: the chain hint
  family: 'evm' | 'cosmos'
  kind: 'transfer' | 'approve' | 'delegate' | 'contractCall' | 'unknown'
  recipient: string      // EVM checksummed 0x / Cosmos bech32 — the value recipient
  asset: { symbol; contract; decimals }
  amount: string         // raw atomic units as a decimal string (no bigint JSON loss)
  spender: string        // for approve
  decoded: boolean       // false + decodeError on any failure (never throws)
  decodeError: string
}

sdk.decodeFromToolResult({ family, chain, payload, args }) -> Envelope
```

- **EVM half** = consolidate over the SDK's existing viem stack: `parseTransaction` + `decodeFunctionData` for `transfer`/`transferFrom`/`approve`/`multicall`-inner-calls. The recipient/amount are lifted from the **calldata**, so an ERC-20 transfer reports the token recipient (NOT the contract); typed txs lift the EIP-155 chain id off the wire. Native sends fall back to `tx.to`/`tx.value`.
- **Cosmos half** = the one genuinely-new build, on `cosmjs-types` proto3 (NOT hand-rolled): `TxRaw.decode -> TxBody.decode -> Any -> typed Msg.decode` for `MsgSend`/`MsgDelegate`/`MsgUndelegate`/`MsgExecuteContract`. Per the cluster-5 basis, `cosmjs-types` is the RN/Hermes-safe, zero-dep, bigint keystone (already in-tree transitively; now a direct dep).

Ported from the Go reference `internal/safety/envelope.go#DecodeFromToolResult`.

## why

The same calldata-decode logic is re-implemented 3x across Go + mcp-ts + the SDK, kept in sync by hand. That means the app, the CLI, and the backend can **disagree on what a tx decodes to** — three parsers, three answers, one signing card. One canonical SDK decoder kills the cross-surface drift and gives the single bytes oracle every safety surface reads from.

**Pure crypto, strict boundary:** decodes/parses only. It never judges intent, grounds claims, signs, or broadcasts. Recipient/amount come from the bytes; `args` is fill-only for the asset symbol and never overrides a decoded value. Fails closed (`decoded: false`) on malformed bytes and on multi-message / multi-coin txs so drift can't ride through as a pass. Agent judgement (intent-match, fabricated-vs-grounded, injection) and product policy (spend caps/allowlists) stay in the backend — they *consume* the Envelope, they aren't it.

## consumers unblocked

One decoder shared by all 5 surfaces:
- isolate `hostValidate`
- CLI text-WYSIWYS
- app decoded-intent card
- VultiServer co-sign gate (decode shared; product policy layered on top, separate lib)
- migration shadow-diff (old vs new on the SAME decoder -> converges)

## test matrix

- `packages/sdk/tests/unit/decode/fromToolResult.test.ts` — 16 cases on **real** on-the-wire bytes (viem-encoded EVM RLP + cosmjs-types-encoded proto3): ERC-20 transfer (recipient != contract), native send, approve (spender), unrecognised contract call, args.unsigned_payload extraction, MsgSend/MsgDelegate, multi-message fail-closed, garbage-bytes fail-closed, family inference. Whole SDK unit suite: **1525 passed**.
- Typecheck: **0 errors** (`yarn workspace @vultisig/sdk typecheck`).
- Runnable receipt: `scripts/receipts/decode_from_tool_result.mjs` (NO broadcast).

```
── EVM: ERC-20 transfer (USDC, 1.0) ──
  Envelope: { chain: "1", family: "evm", kind: "transfer",
    recipient: "0x70997970C51812dc3A010C7d01b50e0d17dc79C8",
    asset: { symbol: "USDC", contract: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48", decimals: 0 },
    amount: "1000000", decoded: true }
  PASS  evm recipient round-trips
  PASS  evm recipient is the calldata `to`, NOT the token contract
  PASS  evm amount = 1000000 atomic
  PASS  evm chain id lifted from typed tx (1)

── Cosmos: MsgSend (ATOM, 2.5) ──
  Envelope: { chain: "cosmoshub-4", family: "cosmos", kind: "transfer",
    recipient: "cosmos1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5lzv7xu",
    asset: { symbol: "ATOM", contract: "uatom", decimals: 0 },
    amount: "2500000", decoded: true }
  PASS  cosmos recipient round-trips
  PASS  cosmos amount = 2500000 uatom
  PASS  cosmos symbol derived from denom (ATOM)

── shared Envelope shape ──
  PASS  EVM and Cosmos produce the identical Envelope keys

RECEIPT OK — all assertions passed
```

## Release Notes

* **New Features**
  * Added EVM and Cosmos transaction decoding to the SDK, producing a unified “transaction envelope” with normalized fields.
  * Exposed new decoding utilities and shared envelope/types from the SDK and tools entry points (including `PolicyEnvelope` aliasing).
  * Added a chain-agnostic decoder for tool-result payloads with robust fail-closed behavior and decode error reporting.

* **Tests**
  * Added end-to-end unit tests covering EVM transfers/approvals, multicall edge cases, and Cosmos message/staking decoding.

* **Chores**
  * Added `cosmjs-types` dependency to the SDK.


gomesalexandre (commented):
## deep adversarial review — decode.fromToolResult (pushed fixes in `a74405ec`)

Diffed this port line-by-line against the Go reference (`internal/safety/envelope.go` `populateFromEVMTx`/`populateFromCosmosTx`/`evmChainIDToSymbol`) + recipes `evm.AllEVMChainConfigs`. Two confirmed issues, both fixed; rest verified clean.

### Fixed
- **P1 — `evm.ts:117` numeric chain id → spurious BLOCK.** `decodeEvmTx` set `env.chain = String(tx.chainId)` for **all** txs incl. legacy. A typed Base tx surfaced `chain: "8453"` → downstream `chainsMatch("base","8453")` → false → **every legitimate typed-EVM tx BLOCKed**; legacy hints got force-set to the numeric id (e.g. `"1"` clobbering a valid `"ethereum"`). Ported the Go `evmChainIDToSymbol` map (1→ethereum, 8453→base, 42161→arbitrum, 137→polygon, 10→optimism, 56→bsc, 43114→avalanche + the rest of recipes' configs for parity), applied it **only for typed txs** (`tx.type === 'eip1559' | 'eip2930'`, matching Go's DynamicFeeTx/AccessListTx switch), numeric-string fallback for unmapped ids (Go behavior). Legacy (type-0) keeps the caller hint. Verified viem returns `type: 'legacy'` for both pre-155 and EIP-155 legacy, and `chainId` present even on legacy — so the `isTyped` gate (not `chainId != null`) is the correct discriminator.
- **P2 — `cosmos.ts:92` MsgUndelegate mislabeled.** Was `kind: 'delegate'`; added `'undelegate'` to `EnvelopeKind` and use it. Informational (policy keys on recipient/amount/chain/asset).

### Verified clean (no other dropped guards)
- Cosmos multi-message + multi-coin **fail closed** — matches Go (returning the first msg/coin would let drift ride through as PASS). Empty-coin MsgSend leaves amount empty w/ recipient set — Go parity.
- ERC-20 transfer/transferFrom lift the true calldata recipient/amount with `tx.to` recorded as the token contract; `approve` sets `recipient:''` + spender (additive vs Go, more correct). Unrecognised calldata → `contractCall`, recipient stays the contract (correct comparison target). No fail-open path.
- Multicall recursion is additive (Go has none) and harmless — returns `contractCall` when no inner transfer.
- **No affiliate surface** in this decoder (read-only bytes oracle, no calldata construction) — nothing to leak. Decode never throws; all failure modes → `decoded:false` (fail-safe).
- The JSON-probe fallback divergence (SDK fails closed where Go falls back to `decodeFromJSON`) is intentional-per-port-notes and strictly safer; left as-is.

### Receipts
- typecheck: **0 errors**
- unit `fromToolResult.test.ts`: **22/22** (+7 new: typed→symbol for base/arbitrum, mismatched-hint override = wrong-chain detection, unmapped-numeric fallback, legacy keeps hint ×2, undelegate kind)
- receipt `decode_from_tool_result.mjs`: **OK** (updated to a Base typed tx so `8453 → base` resolution is visible in output)

NeOMakinG (approved):
clean decode surface with one fund-critical hole i fixed in-code: the EVM `multicall` path silently dropped every value-moving call after the first, manufacturing a downstream policy fail-open. fixed on `d2d2c378` (fail-closed, mirroring the cosmos multi-message guard) + 3 tests; all CI green. fund-binding decode (recipient/amount/token-contract/chain) is otherwise correct and fails closed on malformed/multi bytes. APPROVE.

## testing approach

worktree `/tmp/wt-sdk-848` off `a74405ec`. ran the decode suite single-threaded (`VITEST_MAX_THREADS=1`) against a minimal config (the full-suite `tests/integration/setup.ts` needs the unbuilt `@vultisig/mpc-wasm` artifact — pre-existing, unrelated to decode; CI's Unit/Integration jobs are the truth and both pass). exercised the runnable receipt `scripts/receipts/decode_from_tool_result.mjs` (real viem EVM RLP + real cosmjs-types proto3 → both round-trip). mutation-checked the decode/test pairing (drop transfer recipient → 4 fails; remove cosmos multi-msg guard → 1 fail; drop cosmos amount → 2 fails — tests have teeth). built a hidden-drain multicall PoC to confirm the blocker was real, then re-ran it post-fix. complexity (`eslint-health.config.mjs`, cognitive-complexity 35) + lint + tsc + prettier all clean on the decode files. 3-lane: Claude (load-bearing) + Codex `gpt-5.3-codex` (converged on the multicall asymmetry independently; fuzzed the hex-coercion path and found it not end-to-end exploitable) + GLM (dead, 429 quota — best-effort, skipped).

## blocking

(none remaining — the one blocker below was fixed in-code on `d2d2c378`, CI green)

### `blocker:` (FIXED in-code) — EVM multicall dropped every value-moving call after the first → policy bypass

`packages/sdk/src/tools/decode/evm.ts` `decodeErc20Like` `case 'multicall'` returned the **first** recognised inner call (`if (inner) return inner`). a crafted batch hid a drain behind a benign first call. confirmed live PoC:

```
multicall([ transfer(decoy, 1), transfer(attacker, 1_000_000_000_000) ])
→ Envelope { decoded: true, kind: "transfer", recipient: decoy, amount: "1" }
   (the 1e12 transfer to the attacker is INVISIBLE)
```

because the policy gate (#847) fails OPEN on dropped fields, that clean-looking benign Envelope is validated against the claim and PASSES while the real drain rides through. this is exactly the drift the cosmos half fails closed against on multi-message txs — and it was net-new vs the Go reference (`internal/safety/envelope.go#populateFromEVMTx`), which doesn't decode multicall at all.

action / fix shipped (`d2d2c378`): a multicall carrying >1 value-moving inner call (transfer/transferFrom with a recipient, or approve with a spender) now returns `decoded:false` + `decodeError`, so no partial view reaches policy. a single value-moving call (plus any non-value-moving calls) still surfaces normally. added 3 tests: single-call passthrough, hidden-drain fail-closed (the PoC), approve+transfer fail-closed. complexity gate stays green.

## should-fix

(none gating)

## suggestion

1. `packages/sdk/src/tools/decode/fromToolResult.ts` `hexToBytes`/`tryBase64ThenHex` — `Buffer.from(x, 'hex')` silently truncates on odd-length/invalid hex rather than throwing. Codex fuzzed this end-to-end against `parseTransaction` and could **not** produce a different-but-valid Envelope (viem's RLP parser rejects the malformed structure → fail-closed via parse error), so it is not a live bypass today. still, a defensive `if (clean.length % 2 || !/^[0-9a-fA-F]*$/.test(clean)) throw` would make the boundary explicit and immune to a future decoder that's more lenient than viem. suggested fix: validate hex shape before `Buffer.from`, return a `decoded:false` Envelope on malformed input.

2. `packages/sdk/src/tools/decode/evm.ts` — for a single-transfer wrapped in multicall, `env.asset.contract` is set to `tx.to` (the router/aggregator), not the actual token contract (the token is the inner call's target, which multicall doesn't expose). minor: the recipient/amount are correct (the fund-binding fields); only the token-contract label is the router. suggested fix: for the multicall-single-transfer case, leave `asset.contract` empty rather than mislabeling it as the router, or document the limitation.

## q

1. `inferFamily` reads `args.chain`/`args.from_chain` to pick the EVM-vs-cosmos decoder when `family` is absent. in the agent flow, are those args ever attacker-influenceable independently of the payload? if a cosmos payload can be coerced into the EVM decoder (or vice versa) by a mismatched chain hint, it'd decode to `contractCall`/garbage rather than a clean wrong Envelope — so likely not a bypass, but worth confirming `family` is always passed explicitly by the trusted caller (the receipts + tests always pass it).

## risk

low blast radius post-fix. the decoder is read-only/pure (no signing, no broadcast) and now fails closed on every multi-value-moving and malformed-bytes path. the one real bypass (multicall hidden-drain) is closed and regression-tested. the remaining suggestions are defensive hardening, not live holes. consumers (#847 policy, CLI WYSIWYS, app card, co-sign gate) read a strictly more-conservative Envelope than before.

## verdict

APPROVE — fund-binding decode is correct (recipient/amount/token-contract/chain lifted from the wire, not args), fails closed on undecodable/multi-message/multi-coin/multi-value-moving inputs, surfaces the contract target (token contract for ERC20, the contract address as recipient for unrecognised calls), tests prove the negative cases (mutation-verified), and all CI is green incl. the lint/type/complexity/prettier gate. the one blocking silent-drop was fixed in-code on `d2d2c378` per the gomes fix-and-approve flow.

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.3-codex` adversarial second-opinion lane (GLM lane dead — 429). decode suite + receipt + mutation + live PoC on darwin/Apple-silicon. fix shipped `d2d2c378`, CI green.

PR #849, opened 2026-06-23
fix(cli): bound agent-backend HTTP calls with a configurable timeout

neavra:
## Problem

Every agent-backend `fetch` in the CLI (`clients/cli/src/agent/client.ts`) was timeout-less. A backend that accepts the TCP connection but never responds — half-open socket, hung load balancer, dropped packets — left the `await fetch(...)` pending **indefinitely**. In a headless / piped `vsig agent` run there is no interactive Ctrl-C path, so the process (and any CI/cron caller) hangs until killed.

Affected calls: `authenticate` (`POST /auth/token`), `healthCheck` (`GET /healthz`), the `get`/`post`/`delete` helpers (conversations, messages-since), and `sendMessageStream` (which only threaded an optional caller `signal`, never a time bound).

## Fix

- Add a per-request timeout (`timeoutMs`) to `AgentClient`. Default **30s**, overridable per-process via `VULTISIG_HTTP_TIMEOUT_MS` (non-positive / non-numeric values are ignored so a typo can't disable it), or via a constructor arg (used by tests).
- Apply `AbortSignal.timeout(timeoutMs)` to every **unary** fetch. A timeout aborts the request → surfaced as a deterministic, catchable `Error('request timed out after Nms')` so headless callers exit non-zero instead of hanging. `healthCheck` reports the hung backend as unhealthy (`false`) rather than throwing.
- For the **SSE stream** (`sendMessageStream`), bound only the **initial connect** with a dedicated controller that's cleared once response headers arrive — the long-lived body read stays unbounded because idle streams send keep-alive pings and a body-level deadline would kill healthy turns. The caller `signal` (Ctrl+C) is combined in via `AbortSignal.any`, so cancellation still aborts the whole request and is preserved as a deliberate cancel (never mislabeled as a timeout).
- Document `VULTISIG_HTTP_TIMEOUT_MS` in `vsig --help`.

> **Note for operators:** the SSE **connect** deadline reuses the same `VULTISIG_HTTP_TIMEOUT_MS` (default 30s). On a cold backend the first connect can be slow; if 30s is too tight, stretch it with `VULTISIG_HTTP_TIMEOUT_MS` (e.g. `VULTISIG_HTTP_TIMEOUT_MS=60000`).

## Tests

New `AgentClient — request timeouts` suite drives every path against a `mockHangingFetch` that mimics fetch's abort contract (never resolves; rejects with the signal's reason when it fires):

- `healthCheck` resolves `false` on a hung request.
- unary `POST` / `DELETE` / `GET` / `authenticate` reject with `request timed out after 20ms`.
- `sendMessageStream` rejects with a timeout when the connect hangs.
- a caller-supplied `signal` still aborts `sendMessageStream` and is **not** mislabeled as a timeout (cancel-semantics regression guard).

Without the fix these hang and the vitest run goes red; with it they reject within the tiny configured timeout.

## Validation

- `yarn typecheck` ✅
- `yarn lint` ✅
- `yarn test:cli` ✅ (306 passed, incl. 8 new)

Changeset: `@vultisig/cli` patch.

## Tuning the timeout

`VULTISIG_HTTP_TIMEOUT_MS` overrides the 30s default. The same value bounds both the unary calls **and** the SSE **connect** deadline (time-to-first-byte; the stream body itself is never time-limited). On a **cold-start backend** that has to spin up a worker / allocate resources before returning `200`, 30s can be tight — raise it (e.g. `VULTISIG_HTTP_TIMEOUT_MS=60000`) for those environments.

## Summary by CodeRabbit

* **New Features**
  * Added configurable HTTP request timeouts for CLI agent-backend calls, including the initial streaming connection phase.
  * Introduced `VULTISIG_HTTP_TIMEOUT_MS` (default: 30s) to control the timeout value.

* **Bug Fixes**
  * Timeouts now surface a clearer `request timed out after Nms` error.
  * Caller-initiated cancellations continue to behave correctly during streaming.
  * Improved handling of requests that hang during health checks, unary operations, and the initial SSE connect phase.


gomesalexandre (approved):
solid, well-scoped fix. the headless-hang scenario is a real ops problem and this addresses it correctly.

## correctness

**unary requests** (`authenticate`, `healthCheck`, `get`, `post`, `delete`): all use `AbortSignal.timeout(this.timeoutMs)` which is the right API — fires a `DOMException('TimeoutError')` that `asRequestError` converts to a clean `Error('request timed out after Nms')`. healthCheck correctly swallows it as `false` rather than throwing, so a hung backend is just reported unhealthy.

**SSE stream** (`sendMessageStream`): the split design is correct - bound only the connect, not the body read. the `settled` flag + `clearTimeout` in the finally is belt-and-suspenders (JS is single-threaded so the timer can't fire during the microtask drain, but the explicit guard makes a late-firing timer a definitive no-op). the `AbortSignal.any([signal, connectController.signal])` correctly combines caller cancellation with the connect deadline so Ctrl+C still works during the connect phase.

the error discrimination logic for SSE:
```
if (connectTimedOut && !signal?.aborted) → timeout error
throw err → caller abort (re-throws DOMException with AbortError identity)
```
this is correct. a caller abort that races with a timeout (`connectTimedOut && signal.aborted`) would rethrow the original DOMException as-is, preserving cancel semantics. good.

**`resolveHttpTimeoutMs`**: the guard `Number.isFinite(n) && n > 0` correctly blocks `NaN`, `Infinity`, `0`, negative values. `Number('')` = 0 but the `raw.trim() === ''` check fires first. clean.

**Node.js compat**: `AbortSignal.any` requires Node 20+. the package's `engines.node >= 20.0.0` already mandates this. no compat issue.

**`asRequestError` and the caller-abort edge case**: if the caller's `extra` signal is already aborted when we check `extra?.aborted`, we return the original error verbatim. if the caller's signal fires concurrently with a timeout, the `timeoutSignal()` races `AbortSignal.any([extra, timeout])` — whichever fires first wins. in the timeout-first case, `extra?.aborted` is false and we return the timeout error. in the abort-first case, `extra?.aborted` is true and we preserve the abort. correct in both orderings.

## tests

the `mockHangingFetch` pattern is correct — it mimics fetch's abort contract (never resolves, rejects with the signal's reason). the tests cover all five unary methods + SSE connect + caller-abort identity + body-streaming-past-connect-deadline.

the cancel-semantics test correctly asserts `err instanceof DOMException && err.name === 'AbortError'` rather than just "no timeout text" — closes the vacuous-pass gap mentioned in the commit message.

the SSE body streaming test (120ms delay past 15ms connect deadline) is a real proof that the `settled`/`clearTimeout` design works.

## q: one timing question

the SSE `connectTimer` timeout fires `connectTimedOut = true; connectController.abort()`. if the initial `fetch` hasn't started yet (e.g. event loop is busy), could the timer fire before `fetch` is even awaited? in that case `combinedSignal` would already be aborted at `fetch(...)` call time, so fetch would reject immediately with an abort reason. `connectTimedOut` is true, `signal?.aborted` is false → correct timeout error. so even this edge case is handled correctly. nice.

## suggestions (non-blocking)

suggestion: `VULTISIG_HTTP_TIMEOUT_MS` is documented in `vsig --help` but not in the repo's README or any operator-facing doc. minor, given the PR description already has a clear note.

suggestion: the unary `get`/`post`/`delete` wrappers now each have their own try/catch with an identical `throw this.asRequestError(err)`. this could be DRY-ed into a `fetchWithTimeout` helper, but the current approach is readable and straightforward. not worth the refactor unless the class grows more endpoints.

no blockers. CI was still pending when this review was written (main was merged into the branch recently) but the existing test suite + the implementation is correct.

PR #850, opened 2026-06-23
fix(client-shared): honor VULTISIG_CONFIG_DIR in config-store

neavra:
## Problem

The shared `config-store` (the auth **vault registry**, `config.json`) computed its directory once at module load from `os.homedir()` and **ignored `VULTISIG_CONFIG_DIR`**:

```ts
const CONFIG_DIR = path.join(os.homedir(), '.vultisig')
```

Meanwhile `credential-store` (`credentials.enc`) **does** honor the env var, resolving it at call time. So in Docker/CI where a user sets `VULTISIG_CONFIG_DIR=/data/vultisig`:

- credentials are written/read under `/data/vultisig` ✅
- the vault registry is written/read under `$HOME/.vultisig` ❌

The two stores diverge: a wallet registered under the overridden dir is **invisible** to a registry lookup pointing at the home dir → "no such vault" despite the creds existing.

## Fix

Mirror `credential-store`'s pattern — resolve the config dir at call time so the env var is honored, keeping the registry and credentials co-located:

```ts
function getConfigDir() {
  return process.env.VULTISIG_CONFIG_DIR || path.join(os.homedir(), '.vultisig')
}
```

`loadConfig` / `saveConfig` / `getConfigPath` now call this. JSON shape and the `{ vaults: [] }` empty-default are unchanged.

## Tests

Extended `clients/cli/tests/config-store.test.ts` (exercises the shared package via the CLI re-export):

- config path resolves under the overridden dir, not `$HOME/.vultisig`
- `saveConfig`/`loadConfig` round-trip uses the overridden dir
- the registry co-locates with credentials (same parent dir as `credentials.enc`)

Env var is restored after each case.

## Validation

- `yarn workspace @vultisig/cli test` — 337 passed
- `yarn test:cli` — 299 passed
- `yarn workspace @vultisig/client-shared typecheck` + `yarn workspace @vultisig/cli typecheck` — clean
- lint + prettier — clean

Includes a `patch` changeset for `@vultisig/client-shared`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Config files now respect the custom `VULTISIG_CONFIG_DIR` setting, keeping config and credentials in the same directory.
  * Saving and loading config now works reliably with overridden config locations, including first-run cases.

* **New Features**
  * The credentials path helper is now available to CLI and shared package consumers.

* **Tests**
  * Added coverage for custom config directory behavior and path consistency between config and credentials.


gomesalexandre (approved):
clean, minimal, correct fix.

## correctness

the root cause was module-load-time evaluation of `const CONFIG_DIR = path.join(os.homedir(), '.vultisig')` which permanently bakes in `$HOME` even if `VULTISIG_CONFIG_DIR` is set at process start. the fix — resolving via `getConfigDir()` at call time — is the right pattern and directly mirrors `credential-store`'s existing `getCredentialsPath()` behavior.

no path traversal issues: `VULTISIG_CONFIG_DIR` is used verbatim as the base path, the same as `credential-store` does. if an operator passes something malicious they already own the process. no sanitization needed here.

`loadConfig`'s catch-all → `{ vaults: [] }` empty default is unchanged, which is correct — the missing-file case should return a clean empty state, not throw.

`saveConfig` uses `mkdir({ recursive: true })` before write, so new override dirs are created correctly.

exporting `getCredentialsPath` from `credential-store` and asserting `path.dirname(getConfigPath()) === path.dirname(getCredentialsPath())` in the test is a nice coupling test — the two stores stay co-located by the test rather than by a hand-copied string.

## tests

the four new test cases cover:
- path resolution under override (not `$HOME/.vultisig`)
- save/load round-trip in override dir
- ENOENT → empty default in override dir
- co-location with credential-store

`afterEach` restores the env var correctly (undefined check before `delete`). clean.

## suggestions (non-blocking)

suggestion: `getConfigDir()` and `getConfigFilePath()` are module-private. `getConfigPath()` is exported and calls `getConfigFilePath()`. there's no public surface for `getConfigDir()` itself — if any caller ever needs to know just the directory (e.g. to enumerate other files in it), they'd have to `path.dirname(getConfigPath())`. minor - totally fine as-is.

all good. no blockers.

PR #852, opened 2026-06-23
fix(cli): vetted EIP-712 digest + low-S + recover-verify in sign_typed_data

neavra:
## Summary

Hardens the CLI agent's `sign_typed_data` so it produces canonical, library-verified EIP-712 signatures instead of relying on a custom, untested hasher. Three independent defects lived in `clients/cli/src/agent/executor.ts`:

1. **Hand-rolled EIP-712 encoder** → replaced with viem's `hashTypedData` (already a CLI dep, no new runtime dependency). viem derives the `EIP712Domain` type from the keys actually present on `domain`, so domains that omit `verifyingContract` (Polymarket's ClobAuthDomain) or carry `salt` now hash to the correct `domainSeparator`. It also throws on a message missing a declared field or naming an unknown `primaryType`, instead of silently hashing the wrong struct. The digest now equals what the app's reference path (`vultiagent-app/src/services/eip712Signing.ts`, ethers `TypedDataEncoder.hash`) produces.

2. **No low-S enforcement** → added `toCanonicalEvmSignature`: folds `s > n/2` into the lower half of the secp256k1 order and flips the recovery parity. A high-S signature is malleable and rejected by OpenZeppelin's ECDSA library (which the Polymarket CLOB uses).

3. **No recover-verify gate** → `signSingleTypedData` now recovers the signer from the assembled `(hash, r, s, v)` and asserts it equals the vault's EVM address, throwing `SIGNATURE_RECOVERY_MISMATCH` on a wrong recovery id / botched low-S fold / digest-keyshare mismatch — instead of returning a structurally-valid-but-wrong signature that only fails downstream at the contract.

Out of scope (left untouched): the payloads-array control flow, the 5s inter-MPC delay, and the `pm_order_ref`/`auto_submit` echo shape.

## Tests

`clients/cli/src/agent/__tests__/executor.eip712.test.ts` now pins:
- a flat **ERC-2612 permit** and the **Polymarket Order (chainId 137) + ClobAuth (chainId 1)** payload-array digest against **both viem AND ethers** (two independent encoders agreeing = canonical correctness);
- the low-S fold (`toCanonicalEvmSignature` high-S → low-S + parity flip);
- the recover-verify gate (valid signature recovers to the vault address; a tampered recovery parity throws `SIGNATURE_RECOVERY_MISMATCH`).

The mock vault now signs each digest with a deterministic test keyshare so the recover-verify gate exercises the real round-trip. Adds `ethers` as a CLI **devDependency** (cross-check only — not shipped to consumers).

## Validation

- `yarn typecheck` ✅ (root + `@vultisig/cli`)
- `yarn lint` ✅
- `yarn test:cli` ✅ (304 passed)
- `yarn knip` / `yarn format:check` ✅

Changeset: `@vultisig/cli` patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EIP-712 digest computation using viem’s `hashTypedData` for consistent results across domain formats.
  * Fixed EIP-712 domain separator handling (including missing `verifyingContract` and `salt`) and normalized `chainId` when provided as a string.
  * Canonicalized signatures to low-S and added a recover-verify gate that rejects signatures not recovering to the vault address with a non-retryable `SIGNATURE_RECOVERY_MISMATCH` error.
* **Tests**
  * Expanded EIP-712 signing tests to cross-check viem vs ethers hashing, and to validate canonical signature formatting and recovery mismatch behavior.


gomesalexandre (commented):
draft review pass (comment-only) — fixes look correct, 307/307 CLI tests pass locally.

**what's good:**
- `computeEIP712Hash` via viem `hashTypedData` is the right call — handles partial domain fields (no `verifyingContract`, has `salt`), throws on unknown `primaryType` and missing declared fields. Verified the `coerceChainId` path handles the string/number wire mismatch.
- Low-S fold is correct — `SECP256K1_N - sBig` with `recovery ^ 1` matches the secp256k1 spec. The mutation test (swap coll↔debt) that would surface here is a non-issue on signing, but the parity flip would be caught by the recover-verify gate.
- Recover-verify gate is a real safety net — catches digest/keyshare mismatch before the signature leaves the executor. Error message is actionable (vault name + id + both addresses).
- Test coverage is thorough: digest pinned against both viem AND ethers for 6 domain shapes, fail-loud tests, low-S unit test, mismatch gate test, end-to-end happy path.

**q for when you un-draft:**
- `pm_batch_ref` gap — the payloads[] return in `signTypedData` only echoes `pm_order_ref`, not `pm_batch_ref`. That's #851's territory but worth double-checking these don't step on each other when stacked.
- `parseDERSignature`: does it handle the case where the sig is raw 64-byte (r||s) format rather than DER-encoded? Some MPC implementations return raw concatenated r||s — worth a comment or guard if `sigResult.format` can vary.

gomesalexandre (approved):
solid security fix ser, three independent defects addressed in one PR. here's the full sweep:

## correctness lane

### EIP-712 digest

the switch to `viem.hashTypedData` is the right call. the hand-rolled encoder was a footgun - stripping it in favor of a battle-tested library removes ~275 lines of recursive struct encoding that had all the usual risks (missed edge types, wrong sorting, etc.). the cross-check against **both** viem AND ethers in the test suite is exactly how you prove canonical correctness — two independent implementations agreeing is stronger than one.

the `EIP712Domain` stripping (`delete messageTypes.EIP712Domain`) is correct: viem synthesizes the domain type from the keys present on `domain`, and having it explicitly in `types` would confuse viem into treating it as a struct ref rather than the domain itself.

the `coerceChainId` logic is correct for all real-world EVM chain IDs (all < 2^53 so `Number()` is loss-free). the only theoretical issue is a chain ID > `Number.MAX_SAFE_INTEGER` parsed from a hex string, but no real EVM chain is anywhere near that range, and the function already handles `bigint` pass-through for the native type.

one observation worth noting: `coerceChainId` uses `Number(s)` for hex strings via `parseInt` path - actually it uses `Number(s)` directly on the hex string and that works because `Number('0x...') ` is valid JS. clean.

### low-S canonicalization

`toCanonicalEvmSignature` is correct:
- uses the right curve order constant (`0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141n` - matches SEC 2 v2 §2.4.1)
- threshold check: `sBig > SECP256K1_N >> 1n` (strict greater-than, so `s == n/2` passes through - that's correct, n/2 is even so s can't equal it exactly in practice, but the spec says s ≤ n/2 so this is fine either way)
- fold: `n - s` produces the complement in the lower half, and `recovery ^ 1` flips the parity. mathematically correct.
- the `parseDERSignature` path correctly handles both raw r||s (128 hex) and DER-encoded signatures. the leading-zero stripping for DER r/s values is correct (DER pads with `00` when the high bit would be set to distinguish from negative).

### recover-verify gate

the gate is sound. calling `recoverAddress` after assembly catches wrong-recovery-id / wrong-digest / wrong-keyshare mismatches before returning success. the error message is genuinely actionable - names vault name+id, expected vs recovered address, and says "retrying will not help" explicitly. good UX for a headless agent operator.

one minor nit on the gate: the `this.vault.address(chain)` call uses the chain resolved from `domain.chainId`, which defaults to `Chain.Ethereum` when chainId is absent. the comment says "the EVM address is identical across every EVM chain (same secp256k1 keyshare)" which is true for pure EVM address derivation. gucci.

### `parseDERSignature` r-stripping edge case

the rHex strip logic: `if (rHex.length > 64 && rHex.startsWith('00')) { rHex = rHex.slice(rHex.length - 64) }` - this only strips if the r value is longer than 32 bytes (DER padding). if rHex is exactly 64 chars but has a leading zero that's significant... that's fine, DER only pads when the MSB would be set, so a 32-byte r starting with 00 in DER would be encoded as 33 bytes. the `> 64` guard is correct.

### tests

coverage is thorough:
- flat permit + string chainId coercion + order + clobauth (no verifyingContract) + 5-field salt + nested struct/array + explicit EIP712Domain stripping
- low-S fold: high-S → correct complement + parity flip; low-S pass-through
- recover-verify gate: wrong recovery parity, wrong vault context (two separate paths)
- final "returned sig recovers to vault address" check via viem `recoverAddress`

the `ethers` devDep for cross-checking is justified — it's test-only and not shipped to consumers.

## suggestions (non-blocking)

suggestion: `parseDERSignature` doesn't have a test in this PR. the recover-verify gate catches a botched DER parse downstream, but a dedicated unit test for the DER path (especially the `rLen=33` with leading-zero case) would be nice to have, since DER parsing is a subtle enough concern.

suggestion: `coerceChainId` throws on `""` (empty string) but returns `Number(0)` would be blocked by the `n > 0` check... actually wait, `Number("") = 0` but the empty-string check `s === ''` fires before that. clean.

suggestion: consider exporting `coerceChainId` for testing alongside `toCanonicalEvmSignature`, since it's now doing meaningful normalization that callers (agent-backend types allow `number|string`) depend on. right now it's only tested indirectly via `signTypedData`.

## verdict

no blockers. three real bugs fixed, well-tested, library delegation is the right call over hand-rolling EIP-712. the recover-verify gate is genuinely useful for catching keyshare mismatches in production. the ethers cross-check in tests is solid evidence of canonical correctness.

praise: two-encoder cross-check strategy (viem + ethers) in `expectCanonicalHash` is elegant — it's the right way to prove correctness without just "testing viem with viem".

PR #856, opened 2026-06-23
feat(sdk): add sdk.defi.pendle (PT buy/sell/redeem)

gomesalexandre:
First protocol under the new \`sdk.defi.*\` namespace.

## what
\`sdk.defi.pendle\` — build UNSIGNED Pendle PT (Principal Token) buy/sell/redeem txs by wrapping the Pendle Hosted SDK Convert REST API. The Convert endpoint hands back ready router calldata; we emit it as an unsigned tx (+ the prerequisite ERC20 approve leg) and NEVER sign or broadcast. Pure crypto / unsigned-only.

Surface:
- \`pendle.markets({ chain, underlying?, limit? })\` / \`pendle.market({ chain, market })\` — read active markets (PT fixed APY, underlying APY, YT range, liquidity).
- \`pendle.buildBuyPt(...)\` — token → PT (lock a fixed yield to maturity).
- \`pendle.buildSellPt(...)\` — PT → underlying (exit before maturity).
- \`pendle.buildRedeem(...)\` — fail-closed at maturity (active-market → use sell; expired → not enabled until an expired-market source is money-tested). Refuses to relabel an active-market sell route as a 1:1 mature redeem.

Also exported as top-level named functions (\`buildBuyPt\`, \`pendleMarkets\`, …) + the \`defi\` namespace object.

## why
DeFi consolidation: lift mcp-ts's proven Pendle REST hand-roll into the SDK so every consumer (mcp-ts, agent-backend, station) routes through one boundary instead of re-implementing the Convert plumbing + the easy-to-miss chain-prefix-stripping integration bug.

Lib decision (from the lib-vs-handroll deep-dive): WRAP the hosted REST API. The npm SDK (\`@pendle/core-v2\`) pins ethers 5.7 Solidity helpers (RN/Hermes kill-filter fail) and the router curve math is too gnarly to hand-roll — the Convert endpoint returns ready calldata. The client is a thin \`queryUrl\` fetch wrapper: no ethers, no \`node:crypto\`, no WASM. chainId is derived from core's chain registry, never hardcoded.

## safety
- Router target is **allow-listed**: refuses to build anything where \`tx.to\` != Pendle Router V4 (\`0x8888...F946\`).
- Market/PT/underlying are **trust-but-verified** against the live active-market catalog before any calldata is built.
- Affiliate/fee param (\`aggregatorReceiver\`) is **INJECTABLE** by the consumer and defaults **OFF/neutral** — the SDK is multi-consumer and never hardcodes a referral (e.g. never 'station').

## consumers unblocked
mcp-ts (can drop \`lib/pendle-api.ts\` + \`tools/pendle\` plumbing and call the SDK), agent-backend, station — one Pendle boundary, no cross-repo drift.

## test
- 10 hermetic vitest unit tests (network mocked at the \`queryUrl\` boundary): REST URL shape, chain-prefix stripping, router allow-list refusal, market/token mismatch refusal, ERC20 approve encoding, injectable/default-off affiliate, fail-closed redeem.
- typecheck: 0 errors (0 new).
- A runnable receipt \`scripts/receipts/defi_pendle.mjs\` fetches a LIVE Ethereum market and builds an UNSIGNED PT-buy tx.

\`\`\`
RECEIPT — node --import tsx scripts/receipts/defi_pendle.mjs

[1/2] fetching live active Pendle markets…
  got 5 markets (top by liquidity):
   - SIERRA  pt=0x1a69154f6f6247e4457332860fb173251a36e03f  fixedAPY=9.44%  liq=$16,770,524
   - USDat   pt=0x1d69402390657308c91179aa184bf992908c1e08  fixedAPY=7.66%  liq=$9,928,880
   - sUSDe   pt=0x5a19fa369f2895dcd8d2cee62e4ceae58ef92bbb  fixedAPY=4.33%  liq=$9,555,118
   ...
[2/2] building UNSIGNED buy-PT for: SIERRA

--- UNSIGNED ROUTER TX ---
{ "chain": "Ethereum", "chainId": 1, "from": "0x1111…1111",
  "to": "0x888888888889758F76e7103c6CbF23ABbF58F946",
  "value": "0", "data": "0xc81f847a0000…(1802 chars)", "gasLimit": "2000000" }

--- UNSIGNED APPROVE LEG (signed first) ---
{ "to": "0x6bf7788eaa948d9ffba7e9bb386e2d3c9810e0fc",
  "data": "0x095ea7b3…888888888889758f76e7103c6cbf23abbf58f946…f4240", "gasLimit": "120000" }

--- META ---
{ "protocol": "Pendle Finance", "router": "0x8888…F946",
  "method": "swapExactTokenForPt", "amountInBaseUnits": "1000000",
  "expectedOutBaseUnits": "1029451", "impliedApyAfter": 0.0944,
  "approvalRequired": true,
  "note": "…UNSIGNED — not broadcast here." }

✅ UNSIGNED PT-buy tx built. Router allow-listed, never signed, never broadcast.

unit: 10 passed (10) · typecheck: 0 errors
\`\`\`

Part of the \`sdk.defi.*\` DeFi consolidation.

## New Features
- Added `sdk.defi.pendle` with **unsigned** calldata builders for Pendle PT **buy**, **sell**, and a **fail-closed** `redeem` (rejects unsupported conditions)
- Added Pendle **active market discovery** helpers and PT market selection (including underlying filtering) plus chain/router constants for supported networks
- Expanded the `sdk.defi.*` namespace aggregation (now includes `threeJane` alongside `pendle`)

## Tests
- Added hermetic unit tests for market discovery and unsigned builder behavior

## Developer Tools
- Added an offline receipt generator script for unsigned Pendle PT buy calldata

## Refactor
- Updated SDK tool exports to focus the `sdk.defi.*` unsigned builder surface


gomesalexandre (commented):
deep adversarial pass — calldata-builder fund-safety class (the same hotspot where sibling PT/swap reviews found u64-wrap + amount-injection + addr-bypass). found + fixed 3 real ones (pushed 31576ba2):

## found + fixed
- **preferably-blocking — `encodeErc20Approve` had no amount bounds / no decimal-strictness** (`pendleApi.ts:183`). `BigInt(amount).toString(16).padStart(64,…)` silently emitted a **65-byte word** for any amount > 2^256-1 (verified: `(2^256+5)` → 139-char calldata vs the correct 138 → ABI layout corrupted → signer gets bytes that do NOT match the reported `approvalReq.amount`). It also accepted `BigInt`-parseable `0x..`/`0o..` injection and a negative `-1` (→ malformed `…000-1` segment). Now requires a syntactic 0x 20-byte spender + a base-10 amount in `[0, 2^256-1]`, fail-closed otherwise. This is the exact "encodes a different value than reported / missing upper bound" class flagged across the sibling reviews.
- **preferably-blocking — unvalidated approval `to` (the token the user approves)** (`index.ts`). `approvalReq.token` from Convert flowed straight into the approve tx `to` with zero validation (unlike `route.tx.to`, which was router-checked). A malformed/compromised Convert response could build an approve against a garbage address. Now `/^0x[0-9a-fA-F]{40}$/` gated.
- **suggestion→fixed — router calldata + native `tx.value` forwarded to the signer unchecked** (`index.ts`). `route.tx.data` was only truthy-checked (non-hex would reach signing); `tx.value` was forwarded verbatim for native input (hex/garbage → signer). Now: data must be `0x`-hex non-trivial length; value must be decimal ≤ uint256.

## verified clean
- **router allowlist**: `route.tx.to` hard-asserted == Pendle Router V4 (`0x8888…F946`), refuses otherwise. ✅ (live receipt: real Convert returned V4)
- **trust-but-verify market**: requested market/pt/underlying re-matched against the live active-market catalog before any signable build; mismatch → fail-closed. ✅
- **fail-closed redeem**: `buildRedeem` throws for both active (→ use sell) and expired/unknown markets; builds zero calldata. ✅
- **affiliate-neutral CONFIRMED**: `aggregatorReceiver` is injectable, default-OFF, only ever a REST query param (`pendleApi.ts:171`), **never** in calldata. Live receipt URL carries no `aggregatorReceiver`; existing test asserts default-absent + inject-on-pass. ✅
- **no phantom deps**: RN-safe (`queryUrl` only, no ethers/node:crypto/WASM). ✅

## receipts
- typecheck: `tsc --noEmit` RC=0, 0 new errors
- unit: **14/14** green (was 10; +4 adversarial: malformed calldata, hex-injected value, bad approval token, overflow amount) + real-hex fixtures (placeholders like `0xMarket…`/`0xUSDC…` are non-hex and would now (correctly) be rejected — swapped to valid 20-byte hex)
- live receipt vs real Pendle Hosted SDK REST (Ethereum): UNSIGNED PT-buy built, `to`=Router V4, approve `0x095ea7b3…f4240` well-formed (138 hex), `value`=0, affiliate absent, never signed/broadcast ✅

never merged. comment-only on my own PR.

NeOMakinG (approved):
Reviewed across 3 lanes (Claude correctness/fund-safety load-bearing + Codex `gpt-5.3-codex` adversarial + GLM third-eye). GLM lane dead (0-byte, best-effort skip). Pendle PT buy/sell/redeem is a wrap of the Pendle Hosted Convert REST API — same API-trust-boundary class as sdk.swap.skip #833 / jupiter #835, and the defensive hand-roll here is strong. One real release-hygiene gap (missing changeset) I fixed in-branch; the rest are tiered consistently with #833/#835 (faithful-port + unsigned-output + downstream policy/MPC gate ⇒ should-fix/headline follow-up, not per-PR blockers).

## testing approach
- Read both source files + tests + receipt against HEAD; verified every fund-safety guard by reading the path, not the test description.
- **On-chain verification**: Pendle Router V4 `0x888888888889758F76e7103c6CbF23ABbF58F946` confirmed deployed with byte-identical bytecode on Ethereum, Arbitrum, Base (CREATE2-deterministic claim holds). Live PT `0x1a691...e03f` from the receipt decodes via `symbol()` to `PT-SIERRA-2JUL2026` (real Pendle PT).
- **Live API probe**: `GET /v2/sdk/1/convert` returns `tx.to == Router V4` and `requiredApprovals[0].amount == amountsIn` exactly (NOT MAX_UINT256) — approval is exact-amount in practice. Markets endpoint returns market `address` bare but pt/yt/sy/underlying chain-id-prefixed; `stripChainPrefix` correctly no-ops on the bare form.
- **Mutation tests (#228 teeth)**: neutered each of router allow-list / market-token mismatch / approve overflow-bound / active-market redeem guard → each fails exactly one test (4/4 have teeth). Tests are not inert.
- **Receipt**: ran live `node --import tsx scripts/receipts/defi_pendle.mjs` — builds an UNSIGNED PT-buy targeting Router V4, exact-amount approve leg, never signs. ✅
- **Gates (local on `68b9d3e4`)**: 14/14 unit (single-thread) · tsc 0 errors · complexity sonarjs-35 exit 0 · prettier `.config/.prettierrc` clean. GitHub CI all-green on prior HEAD `31576ba2` (Unit/Integration/Lint+Type/CodeRabbit). Note: our pushed changeset commit may not re-trigger CI on this branch — gates verified locally.

## blocking
None.

## should-fix
- **API trust boundary — opaque router calldata forwarded verbatim** (`index.ts` ~L230). `receiver=from` is sent to Convert, the response is gated by `tx.to == RouterV4` + hex-shape + bounded `tx.value`, but the router calldata payload itself is opaque and never decoded — a compromised/MITM Convert response could keep `tx.to=0x888…` and encode a different internal recipient or a worse min-out. Mitigations present: output is UNSIGNED (downstream signer/MPC policy is the real gate), router pinned, market/PT/underlying trust-but-verified, value bounded. This is the **same class** as #833 (Skip envelope) / #835 (Jupiter) and tiered identically — a headline follow-up (decode + cross-check recipient/min-out against the request), not a blocker for this PR. Worth tracking explicitly.
- **No SDK-side min-output floor** (`index.ts` L207). `slippagePct` is clamped `[0,50]` and passed to the API; the SDK never cross-checks `route.outputs[0].amount` against a caller floor. Sandwich/underfill protection rests entirely on the API + downstream gate. Same family as the calldata-trust finding; fold into the same follow-up.
- **Approve amount not capped to the trade amount** (`index.ts` L256 + `pendleApi.ts` L194). `requiredApprovals[0].amount` is bounded only syntactically (decimal, ≤ uint256). I confirmed live that Pendle returns exact-amount, so this is not exploitable today — but a cheap, concrete hardening would be to assert `approvalReq.amount === amountsIn` (or ≤ amountsIn) so a misbehaving Convert response can never emit an unlimited router approval. This is the one surface the swap-quote family doesn't have (they don't hand-roll approves), so it's the most actionable of the three.

## suggestion
- `routes[0]` / `requiredApprovals[0]` / `outputs[0]` single-element assumption (`index.ts` L220+) is fail-closed (multi-route → incomplete preflight → on-chain revert, no fund loss) but worth an explicit shape assertion.
- `stripChainPrefix` only strips lowercase `0x`-tails; an uppercase `0X` prefix would not strip. Fund-safe (causes a mismatch → fail-closed, never a bad tx) but a `0x`-case-insensitive check + a few more table tests would harden the parser.

## q
- `buildBuyPt({ tokenIn })` lets the pay-token be any token while only `market/pt/underlying` are verified against the active market. If the intent is "buy PT with any routed token" this is fine (and matches the mcp-ts source); just confirm the consumer surfaces the actual `approval.to` token to the user, not just the market underlying, since that's what gets the spend approval.

## risk
Low for this PR as a building block: output is unsigned, router is pinned + on-chain-verified, market/token trust-but-verified, approve hand-roll is strictly bounded and (live-confirmed) exact-amount, redeem fail-closed. The residual API-trust surface is inherent to wrapping a hosted aggregator and is identical to the already-shipped #833/#835 boundary — the downstream signer/MPC policy is the load-bearing gate, not this builder.

## verdict
**APPROVE.** Fund fields correct (router/PT/market canonical + on-chain-verified, exact-amount approve, decimals are token-own-units per Pendle contract, receiver == `from`), fail-closed on bad input/API/redeem, never signs, RN-excluded consistent with the API-fetching defi convention (RN entry hand-curates `tools/prep`; network-fetching aggregators are not RN-exported). Added the missing changeset in `68b9d3e4` (release version bump — every recent SDK PR ships one; the release-pr workflow silently skips the version PR without it). Tests have teeth, CI green, receipt live. The three should-fix items are the standard hosted-API-wrapper follow-up (track the approval-cap as the cheap concrete one).

🤖 Reviewed with Claude Code (3-lane: Claude + Codex gpt-5.3-codex + GLM)

PR #855, opened 2026-06-23
feat(sdk): add sdk.defi.river (Omni-CDP open/close trove)

gomesalexandre:
Part of the sdk.defi.* DeFi consolidation.

## what

New `sdk.defi.river` surface that builds **unsigned** River Omni-CDP (Satoshi) transactions:

- `buildRiverDelegateApproval` — authorize River's SatoshiPeriphery as a borrower-ops delegate (prerequisite)
- `buildRiverOpenTrove` — open a CDP trove (lock collateral, mint satUSD)
- `buildRiverCloseTrove` — close a trove (repay satUSD, withdraw collateral)

Ported from mcp-ts (`buildRiverDelegateApproval` / `buildRiverOpenTrove` / `buildRiverCloseTrove` + reads). Exposed both as a grouped namespace (`river.*` / `defi.river.*`) and as tree-shakeable named exports off the SDK root.

## why

River shipped no RN/Hermes-safe SDK (per the lib-vs-handroll deep-dive), so this **hand-rolls viem** against River's public contract ABIs. Consolidating the calldata builder into the SDK lets mcp-ts / agent-backend / vultiagent-app stop re-implementing River wire encoding per repo.

**Pure crypto — builds calldata only, never signs, never broadcasts.** The consumer feeds the returned unsigned tx into its own keysign/broadcast path.

Open-trove is **offline-capable**: pass explicit `upperHint`/`lowerHint` for a zero-RPC build, or omit them to resolve the sorted-troves insert position on-chain.

Any **affiliate/fee param is INJECTABLE** and defaults neutral/off — `maxFeeBps` (River's borrowing-fee tolerance, default 5%) and an opaque `affiliateTag` (echoed in meta only, never written to calldata). The SDK is multi-consumer; nothing is hardcoded to a specific consumer.

## consumers unblocked

- mcp-ts River tools route through the SDK instead of vendoring the ABIs/addresses
- agent-backend / vultiagent-app get a typed River build surface for free

## receipt

`yarn tsx scripts/receipts/defi_river.mjs` (Ethereum, fully offline, no broadcast):

```
=== sdk.defi.river — UNSIGNED River Omni-CDP build receipt ===
1) DELEGATE APPROVAL
   decoded: setDelegateApproval ["0xDC23b633c23a9d6E55Cb454c673F767bf65f920C","true"]
2) OPEN TROVE  (maxFeeBps=250 injected, affiliateTag="demo-consumer")
[...]

PR #839, opened 2026-06-23
feat(sdk): add sdk.validate normalizers (decimals/amount/fee/token-symbol)

gomesalexandre:
## what

Ports the PURE chain-math half of the agent-backend validator into the SDK as `sdk.validate` normalizers:

- `scaleRawToHuman` / `scaleHumanToRaw` — base-unit <-> human decimal, exact BigInt fixed-point (no float round-trip the way `fromChainAmount` does)
- `amountMatches` — relative-tolerance amount comparison with a 1e-18 absolute floor
- `computeEvmFee` / `feeMatches` — `gasLimit * maxFeePerGas / 1e18`, BigInt multiply so gas-spike products don't overflow
- `decimalsFor` / `tokenDecimals` — canonical per-ticker decimals registry (ported verbatim from the Go `tokenDecimals` map)
- `isValidTokenSymbolFormat` / `normalizeTokenSymbol` — pure FORMAT (shape) validation + uppercase/slash-pair normalization

Ported from `agent-backend/internal/service/agent/validator/{decimals,amount,fee,token_symbol}.go`.

## why

mcp-ts and the agent-backend each re-implement base-unit scaling + a token-decimals table. That's the cross-repo drift root cause for amount/fee rendering. Centralizing the PURE math in the SDK means one source of truth: the backend can ground a claimed amount/fee through the SDK instead of carrying its own `math/big` copy.

PURE crypto ONLY. The agent-judgement layer (response regex-scraping, fabricated-vs-grounded grounding against tool results, severity-by-signing-surface) intentionally STAYS in the agent backend. Nothing here signs, broadcasts, or fetches.

## consumers unblocked

- `mcp-ts` — drop its local base-unit/decimals helpers, route through `@vultisig/sdk`
- `agent-backend` — ground amount/fee claims via the same scaling the SDK signing path uses

## RECEIPT

```
$ node --import tsx scripts/receipts/validate_normalizers.mjs
=== sdk.validate normalizers — runnable receipt ===

CACAO decimals (registry)    10
raw base units               220208030381
-> human                     22.0208030381 CACAO
human -> raw (round-trip)    220208030381
round-trip lossless?         true
claim 22.02 within 1%?       true
claim 0.00022.. (off 1e5)?   false
gasLimit * maxFeePerGas fee  0.000315 ETH
claim 0.000315 within 5%?    true
claim 0.01 (fabricated)?     false
normalize ruji/rune          RUJI/RUNE  parts=[RUJI, RUNE]

[...]

PR #858, opened 2026-06-23
feat(sdk): add sdk.defi.arkis (lender supply)

gomesalexandre:
## what

Adds `sdk.defi.arkis` — a lender-side **supply** builder for the Arkis protocol on Ethereum. It builds the unsigned 2-step sequence (ERC-20 `approve` → ERC-4626 / Agreement `deposit`) and returns the decoded transactions. It **BUILDS UNSIGNED calldata ONLY — never signs, never broadcasts. Pure crypto.**

Ported from mcp-ts `build_arkis_supply_liquidity` + reads, reshaped into the SDK `sdk.defi.<protocol>` namespace alongside the existing prep/swap/token surfaces.

New surface:
- `sdk.defi.arkis.buildArkisSupplyTx(params)` — pure (no network I/O), encodes approve + deposit for both pool kinds:
  - ERC-4626 vault: `deposit(uint256 assets, address receiver)` (selector `0x6e553f65`), receiver fixed to `from`
  - Standard Agreement: `deposit(uint128 amount)` (selector `0x54469aea`)
- `sdk.defi.arkis.resolveArkisPoolKind(pool)` — optional on-chain `asset()` probe (read-only `eth_call` via `getEvmClient`)
- `sdk.defi.arkis.parseArkisTokenAmount(amount, decimals)` — exact string-math base-unit parse
- `sdk.defi.arkis.ARKIS_OFFICIAL_ADDRESSES` / `ARKIS_BOOK_URLS`

Also exported top-level as `defi` (+ `Defi` type) and surfaced as the `sdk.defi` getter.

## why

Part of the `sdk.defi.*` DeFi consolidation — moving protocol calldata builders out of mcp-ts and into the multi-consumer SDK so Station, Windows extension, mcp-ts, and the CLI all share one audited primitive.

**Consumers unblocked:** mcp-ts `build_arkis_supply_liquidity`, agent-backend execute flows, and any client that wants an unsigned Arkis supply tx without dragging an Arkis SDK.

## how

HAND-ROLL viem (`encodeFunctionData` + minimal Arkis ABI fragments), per the lib-vs-handroll deep-dive: no RN/Hermes-safe official Arkis package exists, and the supply surface is a 2-call ERC-20-approve + ERC-4626-deposit sequence. viem is already an SDK dep.

Safety notes:
- ERC-4626 deposit `receiver` is fixed to `from` so the scaffold cannot mint shares to a third party.
- The `affiliate` param is **INJECTABLE, defaults to neutral/off**, and is a **metadata-only passthrough** (Arkis supply has no on-chain affiliate slot). It never alters calldata and no consumer identity (`station`, etc.) is hardcoded — the SDK is multi-consumer.
- `buildArkisSupplyTx` is pure; the only network path is the optional `resolveArkisPoolKind` read.

## receipt

`node --import tsx scripts/receipts/defi_arkis.mjs` (from `packages/sdk`) — builds unsigned approve+supply for both pool kinds with sample inputs, decodes the calldata, and asserts the affiliate tag never leaks into calldata. No broadcast.

```
sdk.defi.arkis — UNSIGNED lender supply receipt (no broadcast)
Arkis dispatcher (official): 0x2f01D7CFfe62673B3D2b680295A2D047F3848e4c

=== ERC-4626 vault supply — 1500 USDC ===
protocol=Arkis chain=Ethereum chainId=1 poolKind=erc4626_vault
pool=0x2222…2222 token=0xA0b8…eB48 from=0x7099…79C8 receiver=0x7099…79C8
amountRaw=1500000000 affiliate=(neutral/off)
  tx[0] APPROVE  to=0xA0b8…eB48 value=0
         decoded: approve(spender=0x2222…2222, amount=1500000000)
         data:    0x095ea7b3…0000000059682f00
  tx[1] DEPOSIT  to=0x2222…2222 value=0
         decoded: deposit(assets=1500000000, receiver=0x7099…79C8)
         data:    0x6e553f65…70997970c51812dc3a010c7d01b50e0d17dc79c8

=== Agreement supply — 2500.25 USDC ===
protocol=Arkis chain=Ethereum chainId=1 poolKind=agreement
  tx[1] DEPOSIT  to=0x1111…1111 value=0
         decoded: deposit(amount=2500250000)
         data:    0x54469aea000000…9506c990

=== Injectable affiliate (multi-consumer, default off) ===
neutral.affiliate=(undefined)  tagged.affiliate=example-consumer
calldata identical regardless of affiliate: true

OK — unsigned Arkis approve+supply sequences built and decoded. No signing, no broadcast.
```

Gate:
- `yarn workspace @vultisig/sdk typecheck` → 0 errors (clean baseline on this main; my files add 0 new)
- `yarn workspace @vultisig/sdk test` → **1519 passed** (incl. new `tests/unit/defi-arkis.test.ts` + a `sdk.defi` public-export assertion)
- eslint + prettier (`.config/.prettierrc`) clean on the new files

Part of the sdk.defi.* DeFi consolidation.

### DIFF
--- a/.changeset/wild-arkis-supply.md
+++ b/.changeset/wild-arkis-supply.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': minor
+---
+
+Add `sdk.defi.arkis` — a lender-side **supply** builder for the Arkis protocol on Ethereum. Builds the unsigned 2-step sequence (ERC-20 `approve` → ERC-4626 / Agreement `deposit`) and returns the decoded transactions. Builds UNSIGNED calldata only — never signs, never broadcasts. Also surfaces `resolveArkisPoolKind` (optional on-chain `asset()` probe) and `parseArkisTokenAmount` (exact string-math base-unit parse), exposed via the new `sdk.defi` getter.
--- a/packages/sdk/src/Vultisig.ts
+++ b/packages/sdk/src/Vultisig.ts
@@ -46,6 +46,7 @@ import type { PushNotificationService } from './services/PushNotificationService
 import { type PerformReshareParams, SecureVaultCreationService } from './services/SecureVaultCreationService'
 import { SecureVaultFromSeedphraseService } from './services/SecureVaultFromSeedphraseService'
 import type { Storage } from './storage/types'
+import { type Defi, defi } from './tools/defi'
 import {
   AddressBook,
   AddressBookEntry,
@@ -159,6 +160,16 @@ export class Vultisig extends UniversalEventEmitter<SdkEvents> {
    */
   public get notifications(): PushNotificationService {
     return this.context.pushNotificationService
+  }
+
+  /**
+   * DeFi protocol primitives. Builds UNSIGNED calldata only (e.g.
+   * `sdk.defi.arkis.buildArkisSupplyTx(...)`) — never signs or broadcasts. The
+   * surface is multi-consumer: any affiliate/fee param is injectable and
+   * defaults to neutral/off, with no consumer identity hardcoded.
+   */
+  public get defi(): Defi {
+    return defi
   }
 
   /**
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -393,6 +393,7 @@ export type {
   CoinKey,
   CoinMetadata,
   DecodeFromToolResultInput,
+  Defi,
   Envelope,
   EnvelopeKind,
   FieldDiff,
@@ -420,7 +421,6 @@ export type {
   VaultIdentity,
   Verdict,
 } from './tools'
-export { type Defi, defi } from './tools'
 export {
   abiDecode,
   abiEncode,
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -364,6 +364,7 @@ export type {
   TokenMetadataResolver,
   VaultIdentity,
 } from './tools'
+export { type Defi, defi } from './tools'
 export {
   abiDecode,
   abiEncode,
--- a/packages/sdk/src/platforms/react-native/index.ts
+++ b/packages/sdk/src/platforms/react-native/index.ts
@@ -211,6 +211,15 @@ export {
   resolveEns,
 } from '../../tools/evm'
 
+// DeFi protocol primitives (build UNSIGNED calldata only). viem-backed and
+// RN-safe — `buildArkisSupplyTx` / `parseArkisTokenAmount` are pure (viem
+// `encodeFunctionData`), and `resolveArkisPoolKind`'s read-only `eth_call`
+// rides the same externalized core-chain EVM client as the evm tools above.
+// The generic entry (src/index.ts) exposes this via the `defi` namespace; the
+// RN allow-list omitted it so RN consumers (Station, Windows ext) couldn't
+// build an unsigned Arkis supply tx and had to drag the Arkis SDK.
+export { type Defi, defi } from '../../tools/defi'
+
 // Cosmos staking + distribution module (LCD queries — read-only,
 // vault-free, generic over every ibcEnabled cosmos chain). Mirrors the
 // generic entry (src/index.ts); the React Native allow-list omitted
--- a/packages/sdk/src/tools/defi/arkis/abi.ts
+++ b/packages/sdk/src/tools/defi/arkis/abi.ts
@@ -1,0 +1,52 @@
+// Arkis lender-side ABIs.
+//
+// Two supply paths are supported, mirroring the production Arkis app:
+//   - ERC-4626 vaults: `deposit(uint256 assets, address receiver)`
+//   - Standard Agreements: `deposit(uint128 amount)`
+//
+// We hand-roll the minimal ABI fragments (viem `encodeFunctionData`) rather
+// than dragging an Arkis SDK — there is no RN/Hermes-safe official package, and
+// the supply surface is a two-call ERC-20-approve + ERC-4626-deposit sequence.
+//
+// Read fragments (`asset`, `maxDeposit`, `decimals`, …) are used only by the
+// optional on-chain resolver helpers; the calldata builder itself is pure.
+
+export const erc4626WriteAbi = [
+  {
+    name: 'deposit',
+    type: 'function',
+    stateMutability: 'nonpayable',
+    inputs: [
+      { name: 'assets', type: 'uint256' },
+      { name: 'receiver', type: 'address' },
+    ],
+    outputs: [{ type: 'uint256' }],
+  },
+] as const
+
+export const erc4626ReadAbi = [
+  {
+    name: 'asset',
+    type: 'function',
+    stateMutability: 'view',
+    inputs: [],
+    outputs: [{ type: 'address' }],
+  },
+  {
+    name: 'maxDeposit',
+    type: 'function',
+    stateMutability: 'view',
+    inputs: [{ name: 'receiver', type: 'address' }],
+    outputs: [{ type: 'uint256' }],
+  },
+] as const
+
+export const standardAgreementWriteAbi = [
+  {
+    name: 'deposit',
+    type: 'function',
+    stateMutability: 'nonpayable',
+    inputs: [{ name: 'amount', type: 'uint128' }],
+    outputs: [],
+  },
+] as const
--- a/packages/sdk/src/tools/defi/arkis/addresses.ts
+++ b/packages/sdk/src/tools/defi/arkis/addresses.ts
@@ -1,0 +1,23 @@
+// Arkis published mainnet contract addresses.
+//
+// Pinned from Arkis docs:
+// https://docs.arkis.xyz/home/for-developers/high-level-architecture
+//
+// Lender-side scaffold is Ethereum mainnet only. Hyperliquid L1 wrapper
+// collateral, margin-account vault allocation, direct borrow, and withdrawals
+// are intentionally out of scope here.
+export const ARKIS_OFFICIAL_ADDRESSES = {
+  dispatcher: '0x2f01D7CFfe62673B3D2b680295A2D047F3848e4c',
+  compliance: '0x286496C568368e036062781F807db5ea3E56d3e8',
+  agreementFactoryV2: '0xbbC9c04348E093473C5b176Cb4b103fF706528bf',
+  leverage: '0xe70d11D23F36826C58f30C61B4DeAf0A89a6D837',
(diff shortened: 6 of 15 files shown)


gomesalexandre (commented):
## deep adversarial pass — sdk.defi.arkis calldata builder

Hunted the crypto-in-SDK fund-safety class (u64/u256 amount-wrap, amount-string injection, address-validation bypass / wrong-chain, missing upper bounds, affiliate-in-calldata) against the mcp-ts `build_arkis_supply_liquidity` source. **Mostly clean, one real hardening gap fixed.**

### fixed (1)
- **preferably-blocking: ERC-4626 wrong-token approve / allowance-leak footgun** — `buildSupplyTx.ts`. The pure builder enforced the `tokenAddress === pool.asset()` invariant only via a doc comment ("caller is responsible"), whereas mcp-ts enforces it on-chain (`arkis-tools.ts:593`). A caller passing `poolKind: 'erc4626_vault'` with a `tokenAddress` that does NOT match the vault underlying would still emit an `approve` to the wrong ERC-20 (the `deposit` then reverts, but the stray allowance is granted). Added an optional `expectedAsset` guard: when the caller has resolved the asset (e.g. via `resolveArkisPoolKind`, which already returns it), the builder now throws on a mismatch. Default off, no network I/O, builder stays pure, ignored on the agreement path. Commit `2416fccc`, +test.

### verified clean
- **amount encoding == reported, no wrap** — receipt-decoded calldata: `1500000000 → 0x59682f00`, `2500250000 → 0x9506c990`. Bytes handed to encoding == `amountRaw` echoed on the result. uint256 upper bound enforced; uint128 upper bound enforced on the agreement path (`> MAX_UINT128` throws). `amountRaw <= 0n` rejects zero/negative bigints.
- **amount-string strictness** — `parseArkisTokenAmount` is `/^\d+$/` on both whole+frac parts (ASCII `\d`, no unicode flag). Fuzzed `0x10`/`1e3`/`+5`/`0b101`/arabic-indic digits/`1_000`/`5n` → all THROW. No hex/octal/exponent/unicode injection. String-only math, exact, frac-overflow rejected.
- **receiver == self** — ERC-4626 `deposit(assets, receiver)` fixes `receiver = from`; decoded calldata confirms (`…79c8`). No third-party share mint, not injectable.
- **selectors correct** — `toFunctionSelector` confirms `deposit(uint256,address)=0x6e553f65`, `deposit(uint128)=0x54469aea`, `approve=0x095ea7b3`. Match the wire bytes.
- **address validation** — `requireAddress` throws on non-addresses; `strict:false` + `getAddress` is NOT a bypass here (case-only corruption maps to identical bytes; digit corruption is undetectable in any mode). Mirrors mcp-ts. `value:'0'` on both legs — no native-value leak.
- **affiliate-neutral confirmed** — `affiliate` is injectable, defaults undefined (off), metadata-only passthrough, never touches calldata (test asserts identical bytes tagged vs neutral; receipt re-confirms). No consumer identity hardcoded.
- **addresses** — `ARKIS_OFFICIAL_ADDRESSES` (dispatcher/compliance/factory) + book URLs match mcp-ts byte-for-byte.
- no phantom deps (viem only, already an SDK dep). `resolvePoolKind` is read-only `eth_call`, fail-closed to `agreement`.

### gate
- `yarn workspace @vultisig/sdk typecheck` → **0 errors** (after fix)
- `yarn workspace @vultisig/sdk test` → **1520 passed** (was 1519, +expectedAsset guard test)
- `node --import tsx scripts/receipts/defi_arkis.mjs` → unsigned approve+supply for both pool kinds, decoded, affiliate-leak assertion green, no broadcast

comment-only (own PR — never self-merge).

NeOMakinG (approved):
## testing approach

Three lanes (Claude load-bearing + Codex `gpt-5.3-codex` adversarial + GLM — GLM dead on 429, best-effort skipped). Built the branch in an isolated worktree, ran the new suite single-threaded, mutation-tested the asserts, ran the receipt, and verified the pinned contracts + selectors **on-chain** (mainnet `eth_getCode` / `eth_call` + 4byte). This is a self-built EVM calldata builder (highest fund-risk class), so every fund-binding field was walked against the wire bytes.

**On-chain verification**
- Selectors (4byte + receipt byte-decode): `0x095ea7b3` = `approve(address,uint256)`, `0x6e553f65` = `deposit(uint256,address)` (canonical EIP-4626), `0x54469aea` = `deposit(uint128)`. All match `abi.ts` and the emitted calldata.
- Pinned `ARKIS_OFFICIAL_ADDRESSES` match the Arkis docs page (`docs.arkis.xyz/.../high-level-architecture`) **exactly** (all 8) and all carry code on mainnet. The `dispatcher` is an EIP-2535 Diamond (`eth_call asset()` → `Diamond: Function does not exist`), consistent with Arkis's architecture. **These addresses are exported as metadata only — they are NOT used in the built calldata**, so the #855 selector-on-dispatcher class does not apply here (the deposit target is the caller-supplied `poolAddress`).
- Calldata walked from the receipt: `approve(spender=poolAddress, amount=rawAmount)` then `deposit(...)` to `poolAddress`; ERC-4626 `receiver` is hard-fixed to `from`; approve spender == deposit target (no allowance to a different address than the deposit). Amounts byte-exact: 1500 USDC → `0x59682f00`, 2500.25 → `0x9506c990`.

**Fund-safety invariants confirmed**
- Approval is **bounded to exactly `rawAmount`** (not `MAX_UINT256`).
- `receiver = from` is fixed — no third-party-mint redirect path.
- `parseArkisTokenAmount` is string-math: **no float round-up** (no #839 bug), rejects empty / negative / multiple-dots / over-precision (`fracPart.length > decimals`), positive-only, uint128/uint256 overflow guarded.
- `expectedAsset` guard throws **before** any encoding; output is **unsigned**, pure (no network I/O in the builder).

**Mutation check (test has teeth, #228-resistant):** flipped ERC-4626 receiver→pool, unbounded approve→MAX_UINT256, neutered the `expectedAsset` guard, dropped the over-precision check — **each produced a failing test**. 4/4 caught.

**Gates (local, on fix HEAD `6c1c1046`):** new suite 9/9 (single-thread, inline config — full vitest hits the preexisting unbuilt `@vultisig/mpc-types` setup wall) · `tsc` 0 errors · complexity (sonarjs-35) exit 0 · prettier (`.config/.prettierrc`) clean · receipt runs and decodes both paths. `gh pr checks` green on `2416fccc` (CodeRabbit + Unit + Integration + Lint/TypeCheck). Note: our fix-commit may not re-trigger CI on this branch — gates verified locally.

## blocking

None.

## should-fix

Both addressed in-code on this branch (commit `6c1c1046`) — fix-and-approve:

1. **Missing changeset → feature ships invisibly.** `release-pr.yml` triggers on `push: main` and silently skips version-bumping when no `.changeset/*.md` exists (`exists == 'true'` gate), so merging would NOT bump `@vultisig/sdk` and consumers never receive `sdk.defi.arkis` via a published version. Added `.changeset/wild-arkis-supply.md` (minor). Recurring gomes miss (sdk#833/#856).

2. **RN export omitted.** `src/platforms/react-native/index.ts` did not re-export the `defi` namespace, so RN consumers (Station, Windows ext) couldn't build an unsigned Arkis supply tx. The builder is pure viem and `resolveArkisPoolKind`'s read-only `eth_call` rides the same externalized core-chain EVM client as the existing RN `evm` tools — added the static re-export (tsc clean). Recurring gap (#831/#827/#828/#846).

3. **`resolvePoolKind` fail-open on infra errors (Codex High, tier-confirmed should-fix — left for a follow-up, NOT in-code fixed).** The bare `catch {}` collapses *every* `readContract` failure — RPC outage, timeout, rate-limit, decode error — into `{ kind: 'agreement' }`, not just a real `asset()`-missing revert. A transient read failure on an ERC-4626 vault silently downgrades to the Agreement path, which also drops the `expectedAsset` guard, then builds `approve(pool, amt)` + `deposit(uint128)` against a 4626 vault. **Not drain-exploitable** (approve is bounded to `rawAmount` and goes to the legit pool; the wrong-shape deposit reverts → no fund loss), which is why this is should-fix not blocking — but it's a correctness/robustness hole that turns a network blip into a wasted approve + reverting deposit. Suggest distinguishing a contract revert (`ContractFunctionExecutionError` / `ContractFunctionRevertedError`) from transport errors and re-throwing the latter so the caller can retry rather than silently mis-classifying.

## suggestion

- **`parseArkisTokenAmount` does not validate `decimals`** (Codex Low). A `NaN`/`Infinity`/negative `decimals` yields a raw `BigInt`/comparison throw rather than a domain error, and a very large positive `decimals` makes the `while (fracPart.length < decimals)` pad a CPU/memory sink before the overflow guard runs. It's exported and directly callable. Cheap hardening: `if (!Number.isInteger(decimals) || decimals < 0 || decimals > 36) throw`.

## q

- **Unvalidated `poolAddress` (Codex Medium).** The builder approves + deposits to any caller-supplied address — a malicious target implementing the right `deposit` selector could pull the (bounded) approved amount via `transferFrom`. This reads as by-design for a multi-consumer SDK primitive (the pool comes from the Arkis Lend/Borrow book CSV or `resolveArkisPoolKind` upstream; same opaque-target boundary as skip#833 / pendle#856), and the approval is bounded so it's not infinite-allowance risk. Confirming the intent is that the upstream resolver/book is the trust boundary and the builder stays a pure encoder. If so, a one-line doc note that `poolAddress` is trusted-input would help consumers not misread `sdk.defi.arkis` as “Arkis-address-validated.”

## risk

Low-to-moderate. Self-built EVM calldata is the highest-risk class, but every fund-binding field is verified against on-chain truth and the wire bytes: correct selectors, bounded approve, fixed self-receiver, exact string-math amounts, fail-closed validation, unsigned output. The two recurring ship-blockers (changeset + RN export) are fixed in-code. The residual `resolvePoolKind` fail-open is bounded-no-drain and the `poolAddress` trust boundary is by-design and consistent with the prior `sdk.defi.*` / aggregator PRs.

## verdict

**APPROVE.** Contract + selectors on-chain-verified-present (Diamond facets / canonical EIP-4626 selector), supply calldata + decimals + self-receiver correct, approval bounded, fail-closed, unsigned. Missing changeset + RN export fixed in-code on `6c1c1046` and locally gate-clean. `resolvePoolKind` fail-open (should-fix) and the `decimals`/`poolAddress` items are non-blocking follow-ups.

---
*Reviewed with Claude (load-bearing) + Codex `gpt-5.3-codex` (adversarial second-opinion) + GLM (429, skipped). On-chain checks via mainnet RPC + 4byte. Fixes pushed to the PR branch.*

PR #859, opened 2026-06-23
feat(sdk): add sdk.defi.osmosis (CL + GAMM + Superfluid msg builders)

gomesalexandre:
## what

Adds `sdk.defi.osmosis` — unsigned Osmosis Cosmos message builders under the new `sdk.defi.<protocol>` surface. Ports the mcp-ts `osmosis{Gamm,CL,Superfluid}Tools` (10 tools) into a reusable, multi-consumer SDK primitive.

Three protocol surfaces:
- **GAMM**: `buildJoinPool`, `buildExitPool`, `buildSwapExactAmountIn` (multi-hop)
- **Concentrated Liquidity (CL)**: `buildCreatePosition`, `buildWithdrawPosition`, `buildCollectSpreadRewards`, `buildCollectIncentives`
- **Superfluid**: `buildSuperfluidDelegate`, `buildSuperfluidUndelegate`

Every builder returns an `{ typeUrl, value }` `EncodedMsg` — `value` is the proto3-encoded message body, ready to drop into a Cosmos `TxBody.messages` entry. `osmosis.toAny(msg)` wraps it as the raw `google.protobuf.Any` bytes a SignDirect body needs.

These **BUILD UNSIGNED messages ONLY** — pure crypto, no signing, no broadcast, no network. Encoding is synchronous + side-effect-free.

## why

Consumers unblocked: agent-backend / mcp-ts / vultiagent-app can build Osmosis LP, CL, and superfluid-staking transactions through one shared, fund-safety-hardened SDK surface instead of each re-implementing the msg encoding. Part of the sdk.defi.* DeFi consolidation.

## how

Per the lib-vs-hand-roll deep-dive, the msg-encode side is **hand-rolled on `cosmjs-types` `BinaryWriter`** (already in-tree via the cosmos decoder; zero-dep, bigint, RN/Hermes-safe) rather than dragging in `osmojs` (multi-MB data-as-dep, RN-hostile). `cosmjs-types` is formalized as a direct dep (it was already transitively resolved via `@cosmjs/stargate`, so the lockfile only gains the dep reference — no new package resolution). `bech32` likewise promoted to a direct dep for the address guard.

Every field number / wire type is verified against the on-chain `osmosis-labs/osmosis` protos (cited in per-message doc comments) and **round-trip-asserted** in the test via `cosmjs-types` `BinaryReader`.

Fund-safety:
- bech32 validation with an explicit **account-vs-validator guard** — an `osmovaloper1...` operator key can never be silently used where a spendable `osmo1...` account belongs (and vice versa for `valAddr`)
- base-unit **integer** amounts (`sdk.Int`) — a stray decimal share/amount is rejected before it can fail at broadcast
- deterministic coin sort + position-id dedup/sort
- canonical proto3 default-skip for `int64` ticks (a `0` tick round-trips to `0` on decode)

Any affiliate/fee handling is **consumer-injectable and off by default** (`OsmosisAffiliateConfig`) — never hardcoded to a single consumer.

## test matrix

- `yarn workspace @vultisig/sdk typecheck` -> EXIT 0, 0 errors (my files add 0 new)
- `vitest tests/unit/tools/defi/osmosis.test.ts` -> 16 passed (typeUrl + wire-layout round-trip for all 9 builders + 6 fund-safety rejection cases)
- eslint (repo config) -> clean
- runnable receipt below

```
========================================================================
sdk.defi.osmosis — UNSIGNED message build receipt (no signing/broadcast)
========================================================================

[1] GAMM MsgSwapExactAmountIn
    typeUrl : /osmosis.gamm.v1beta1.MsgSwapExactAmountIn
    body    : 145 bytes  (hex: 0a2b6f736d6f31717970717870713971637273737a673270...)
    decoded :
      #1 sender            = osmo1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5helwsw
      #2 route[0].poolId   = 1
      #2 route[0].outDenom = uosmo
      #3 tokenIn           = 1000000ibc/498A0751C7AB1E5DF00ED54C3C8F11B1B0FE3A9CCDD80B73F2B7E03CB5BC4E78
      #4 tokenOutMinAmount = 950000
    Any     : 192 bytes -> /osmosis.gamm.v1beta1.MsgSwapExactAmountIn

[2] Concentrated Liquidity MsgCreatePosition
    typeUrl : /osmosis.concentratedliquidity.v1beta1.MsgCreatePosition
    body    : 168 bytes  (hex: 08aa08122b6f736d6f31717970717870713971637273737a...)
    decoded :
      #1 poolId            = 1066
      #2 sender            = osmo1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5helwsw
      #3 lowerTick         = -887200
      #4 upperTick         = 887200
      #5 tokensProvided    = 5000000ibc/498A..., 5000000uosmo
      #6 tokenMinAmount0   = 0
      #7 tokenMinAmount1   = 0
    Any     : 229 bytes -> /osmosis.concentratedliquidity.v1beta1.MsgCreatePosition

proto-Any typeUrls built:
  - /osmosis.gamm.v1beta1.MsgSwapExactAmountIn
  - /osmosis.concentratedliquidity.v1beta1.MsgCreatePosition
OK — 2 unsigned Osmosis msgs built + Any-wrapped, nothing signed/sent.
========================================================================
```

Run it yourself: `cd packages/sdk && ../../node_modules/.bin/tsx scripts/receipts/defi_osmosis.mjs`

## risk

Low. New isolated surface, additive only (no existing code paths touched beyond two `export` lines). Builds unsigned msgs only — can't move funds on its own; the consumer's existing sign/broadcast path remains the gate.

Part of the sdk.defi.* DeFi consolidation.


gomesalexandre (commented):
## deep adversarial review — calldata builders (self-review)

Hunted the sibling-PR bug classes hard (u64 silent-wrap, amount-string injection, address-validation bypass, wrong-msg, missing bounds, affiliate leak). One real fund-safety bug found + fixed on the branch.

### preferably-blocking (fixed in 88221fca): uint64/int64 silent wire-wrap

`cosmjs-types` `BinaryWriter.uint64()`/`.int64()` truncate **modulo 2^64 with no error**. The id-field validators (`validatePositiveInt`) had no upper bound, so:

- `buildJoinPool({ poolId: '18446744073709551616' /* 2^64 */ })` → encoded **`0`** on the wire (proof: assertion `expected '0' to be '18446744073709551616'`). Same path for `lockId` (superfluid), `positionId` (withdraw), `positionIds[]` (collect), and `int64` ticks beyond the int64 domain.

Impact: a malformed/oversized id signs a **different pool/lock/position than the user asked for** — the signed bytes silently disagree with the intent. Exactly the "silently encodes a different value than reported" class.

Fix (defense in depth):
- `proto.ts:writeUint64/writeInt64/writePackedUint64` — `assertU64`/`assertI64` at the **encode choke point** so bytes can never disagree with input (`proto.ts:46-67`).
- `validation.ts` — new `validateUint64Id` (1..2^64-1) for poolId/lockId/positionId, and `validateSignedInt` bounded to int64 (`validation.ts:86-118`). Loud, early errors.
- sdk.Int amounts (256-bit) correctly stay unbounded by u64.
- +5 regression tests: boundary 2^64-1 accepted, 2^64 rejected, lockId/positionId/tick overflow rejected.

### verified clean (no bug)

- **proto field layout** — all 9 msgs cross-checked against on-chain `osmosis-labs/osmosis` protos: GAMM join/exit (sender1/pool2/share3/coins4), swap (sender1/routes2/in3/min4 + route pool1/denom2), CL createPosition (pool1/sender2/lower3/upper4/tokens5/min0_6/min1_7), withdraw (pos1/sender2/liq3), collect spread/incentives (ids1 packed/sender2), superfluid delegate (sender1/lock2/val3) + undelegate (sender1/lock2). Field numbers + wire types exact.
- **amount-string injection** — strict `/^[0-9]+$/` (and signed/decimal variants); hex `0x..`, scientific `1e9`, `+`/`-` prefixes, embedded spaces all rejected; amounts passed as strings to `Coin.encode` (no numeric coercion). `.trim()` normalizes trailing whitespace only.
- **address validation** — bech32 decode + account-vs-`valoper` guard so an `osmovaloper1…` can't be used where a spendable `osmo1…` belongs (and vice-versa for valAddr); 20/32-byte payload check.
- **slippage/min-out** — present + injectable on every value-moving msg (swap tokenOutMinAmount, exit tokenOutMins, CL tokenMinAmount0/1).
- **packed repeated uint64** — round-trips correctly; `fork()`/`ldelim()` is the cosmjs same-object pattern (verified `binary.js`), not orphaned.
- **affiliate-neutral confirmed** — `OsmosisAffiliateConfig` is a **type-only** forward-compat surface; **no builder consumes it**, nothing fee-related is injected into calldata, default is OFF. Builders never mutate user amounts.

### receipts

- 21/21 unit tests green (was 16) · `typecheck` clean (0 errors) · receipt script round-trips both a GAMM swap + CL create-position Any with decoded fields matching input (poolId 1066, ticks ±887200, sorted coins, min-amounts 0).

Comment-only (own PR) — not approving.

NeOMakinG (approved):
Fund-safety review of the hand-rolled Osmosis proto3 message builders (GAMM + Concentrated Liquidity + Superfluid). Maximum scrutiny on the wire-encoding choke point — a wrong field number or wrong tick encoding here silently signs funds into the wrong pool/amount. **APPROVE** after fixing 3 items in-code on the branch (RN export, changeset, byte-order coin sort).

## testing approach

Three-lane fan-out. Claude was load-bearing this round: **Codex** hit `model at capacity` mid-run (best-effort, no usable findings) and **GLM** is dead on a 429 quota error — both noted, neither load-bearing. I independently verified every field number against the canonical `osmosis-labs/osmosis` protos and exercised the encoder at runtime in a worktree.

- **proto field numbers — verified vs canonical proto, not asserted by the PR.** Fetched the live `osmosis-labs/osmosis` `gamm/v1beta1/tx.proto`, `poolmanager/v1beta1/swap_route.proto`, `concentratedliquidity/v1beta1/tx.proto`, `superfluid/tx.proto` and matched every field number + wire type:
  - GAMM `MsgJoinPool`{sender=1, pool_id=2, share_out_amount=3, token_in_maxs=4} OK; `MsgExitPool`{sender=1, pool_id=2, share_in_amount=3, token_out_mins=4} OK; `MsgSwapExactAmountIn`{sender=1, routes=2, token_in=3, token_out_min_amount=4} OK; `SwapAmountInRoute`{pool_id=1, token_out_denom=2} OK.
  - CL `MsgCreatePosition`{pool_id=1, sender=2, lower_tick=3 int64, upper_tick=4 int64, tokens_provided=5, token_min_amount0=6, token_min_amount1=7} OK — note pool_id is field 1 *before* sender, unlike GAMM; the code gets this right. `MsgWithdrawPosition`{position_id=1, sender=2, liquidity_amount=3 LegacyDec} OK. `MsgCollectSpreadRewards`/`MsgCollectIncentives`{position_ids=1 `repeated uint64`, sender=2} OK — proto3 `repeated` scalar is packed by default, matching `writePackedUint64`.
  - Superfluid `MsgSuperfluidDelegate`{sender=1, lock_id=2, val_addr=3} OK; `MsgSuperfluidUndelegate`{sender=1, lock_id=2} OK.
- **negative int64 tick encoding — verified at runtime.** `BinaryWriter.int64(-887200n)` → `e0ecc9ffffffffffff01` (10-byte two's-complement varint, NOT zigzag/sint64), round-trips back to `-887200`; `I64_MIN` round-trips to `-9223372036854775808`. This is the correct proto3 `int64` encoding and the classic CL negative-tick bug is **not present**.
- **silent-wrap guards.** The HEAD fix-commit added `assertU64`/`assertI64` bounds at both the validation layer and the encode choke point — a poolId of 2^64 (which `BinaryWriter.uint64` would truncate to 0 → wrong pool) is rejected loudly. Confirmed by tests + boundary case at exactly `U64_MAX`.
- **mutation check.** Flipped `lowerTick` field 3→33 in `proto.ts` → a test fails. The round-trip assertions have real teeth (not #228-inert).
- **runtime:** deps installed in an isolated worktree; `vitest` single-threaded **22/22 passing** (16 original + the int64/uint64 wrap guards + my byte-sort test); receipts script runs and decodes both sample msgs to the expected fields; `tsc` 0 new errors (29 TS2307 are pre-existing unbuilt mpc-types); complexity (sonarjs-35) exit 0; prettier clean. `gh pr checks` all green on the prior SHA (CI has not re-triggered on my fix-commits — gates verified locally).

## blocking

None.

## should-fix

All three fixed in-code on the branch (gomes fix-and-approve):

1. **RN export gap** (`@c8fab8ac`) — `sdk.defi` was not re-exported from `src/platforms/react-native/index.ts`, so RN consumers (vultiagent-app) could not reach `defi.osmosis` even though node could. The builders are pure-crypto (`cosmjs-types` `BinaryWriter` + `bech32`, both RN-safe, no MPC/network) → added a static re-export (no lazy bridge needed), matching the `tools/token` pattern. Recurring gap across #831/#827/#832/#856.
2. **missing changeset** (`@82c3ae85`) — no `.changeset/` entry bumping `@vultisig/sdk`. `release-pr.yml` only opens a version PR when a changeset is present, so merging as-is would ship the new surface in source but **never publish it to npm** for consumers. Added a `minor` changeset. Recurring miss (#833/#856).
3. **coin sort used `localeCompare`** (`@c8fab8ac`) — Cosmos-SDK `Coins.Validate()` requires denoms in raw **byte** order and rejects an unsorted set in `ValidateBasic`. `localeCompare` diverges from byte order when two denoms differ only by the case of the same letter (locale tiebreaks uppercase last; byte puts 'A' 0x41 before 'a' 0x61), which would emit an unsorted `Coins` that fails at broadcast (fail-closed, no fund loss, but a broken tx). For the common `uosmo` + `ibc/<HEX>` pair the two orderings agree, so blast radius is low — but the fix is strictly correct vs the on-chain invariant. Switched to raw code-unit sort + added a mutation-checked test.

## suggestion

- **slippage default of `0`.** `buildSwapExactAmountIn.tokenOutMinAmount` accepts `"0"` and CL `tokenMinAmount0/1` default to `"0"` — a literal-zero min-out is a sandwich-to-zero floor. This is a faithful port of the mcp-ts tools and the slippage bound is legitimately the consumer's responsibility (the builder only carries the value the consumer computed; it never signs), so it is not blocking — consistent with how min-out floors were tiered on #833/#834. Worth a doc note that consumers MUST set a real floor, or an opt-in non-zero-default guard in a follow-up.

## q

- `OsmosisAffiliateConfig` is defined and exported but no builder consumes it (affiliate handling is explicitly the consumer's responsibility, off by default). Confirm it's intentionally a forward-compat type-only API surface for now, not a dropped wiring.

## risk

Low. Additive, isolated new surface; builds **unsigned** msgs only (`{typeUrl, value}` proto bytes) — cannot move funds on its own; the consumer's existing sign/broadcast path stays the gate. All fund-binding fields (field numbers, negative ticks, uint64/int64 bounds, sender/valAddr account-vs-validator guard, byte-order coin sort) verified. Fail-closed on bad input.

## verdict

**APPROVE** at `c8fab8ac`. Proto encoding is byte-correct against the canonical Osmosis protos (every field number + wire type checked, negative int64 ticks correct, uint64/int64 silent-wrap guarded), the account-vs-validator bech32 guard is sound, amounts are integer base-units, builders are fail-closed and unsigned, tests assert exact decoded fields with real mutation teeth. The 3 should-fix items (RN export, changeset, byte-order sort) are fixed on the branch and verified locally. Note CI did not re-run on the fix-commits — gates were verified locally (22/22 vitest, tsc 0-new, complexity exit 0, prettier clean).

---
*3-lane review: Claude (load-bearing — full proto verification + runtime). Codex `gpt-5.3-codex` (best-effort, model-at-capacity this run). GLM `glm-4.6` (best-effort, 429 quota — dead).*

PR #853, opened 2026-06-23
feat(sdk): add sdk.defi.threeJane (USDC supply)

gomesalexandre:
## what

Adds the first protocol under the new `sdk.defi.*` surface: `sdk.defi.threeJane`. It builds the unsigned 2-step Ethereum tx sequence to supply USDC into 3Jane (an ERC-4626-style credit money market):

1. ERC-20 `approve` USDC -> 3Jane Helper
2. `Helper.deposit(assets, receiver, hop)` to mint the senior `USD3` (hop=false) or staked junior `sUSD3` (hop=true) share

Ported from mcp-ts `build_3jane_supply_usdc`, reshaped into a pure SDK builder.

## how

- **Hand-rolled viem** (ERC-4626 + a 1-fn helper ABI) per the lib-vs-handroll deep-dive: no RN-safe official 3Jane SDK exists, so encoding directly with viem's `encodeFunctionData` is the lighter, RN-safe call. No new deps.
- **BUILD-ONLY / pure crypto**: returns unsigned calldata, performs zero network IO, and never signs or broadcasts. Supplier-side only (no borrow, no sUSD3 exit).
- **Injectable recipient, no hardcoded affiliate**: `receiver` is consumer-supplied and defaults to `from` (the only behaviour the on-chain Helper honours — mints to msg.sender). No `station`/affiliate string is baked in — the SDK is multi-consumer. A unit test asserts the serialized output contains neither `station` nor `affiliate`.
- Strict input validation: address checksum, decimal/precision parsing (max 6 dp), positive amount, 1,000 USDC documented minimum, uint256 overflow guard.
- Exposed as the `defi` namespace (`sdk.defi.threeJane.buildThreeJaneSupplyUsdc`) plus flat type re-exports from the package index.

## consumers unblocked

- **agent-backend / mcp-ts**: can drop the bespoke tool wiring and call the SDK builder directly for 3Jane supply calldata.
- **vultiagent-app**: a vault-free path to construct the unsigned approve+deposit pair for the keysign flow.
- Any future `sdk.defi.*` consumer gets a consistent build-only contract.

## test

- `yarn workspace @vultisig/sdk typecheck` -> exit 0, 0 errors (my files add 0 new).
- `vitest` unit test: 9/9 pass.
- Runnable receipt builds + decodes the unsigned calldata with hard assertions: `node --import tsx scripts/receipts/defi_3jane.mjs`.

```
=== 3Jane USDC supply — UNSIGNED build receipt (no broadcast) ===

[...]

PR #865, opened 2026-06-24
fix(sdk): fail-closed sender-binding on 3jane deposit receiver

gomesalexandre:
Closes vultisig/vultisig-sdk#853 fast-follow (open design concern from the adversarial review).

## what

Pins the 3Jane `buildSupplyUsdc` builder to sender-bound behaviour: `receiver` must equal `from`, matching how mcp-ts hard-wires the recipient to `msg.sender`. Passing a distinct receiver now throws instead of building calldata whose share recipient we cannot vouch for.

## why

Flagged preferably-blocking in the #853 review: the Helper.deposit() docs say it mints shares to msg.sender. If the Helper forwards the `receiver` param without an on-chain sender check, a caller injecting a distinct receiver gets:

- funder's USDC consumed (approve + deposit succeed on-chain)
- shares minted to the injected address instead of the depositor

= fund loss with no way to recover. The merged builder accepted any address here. This closes that vector fail-closed until we have on-chain proof the Helper actually honours a delegated receiver.

The fix is one-liner reversible: once someone verifies `Helper.receiver != msg.sender` works on mainnet, remove the equality guard.

## test

- `fails closed on a receiver distinct from the funder` - the old "routes shares to distinct receiver" test flipped to assert throw `/sender-bound/`
- `accepts a receiver explicitly equal to the funder` - same 20 bytes in lowercase casing passes (getAddress normalises)
- 12/12 pass, tsc 0 new errors


NeOMakinG (approved):
Re-review of the #853 fix. Code-only pass (the prior #853 round already deep-tested the builder; this is a +36/-10 surgical follow-up). Verified locally on the worktree: vitest 12/12 single-thread, mutation teeth confirmed, receipts decode-assert exit 0, tsc 0-new, prettier + eslint-health clean. **APPROVE.**

## testing approach
Branch `fix_sdk_3jane_receiver_pin` @ `7883759f` checked out in an isolated worktree. Ran `defi-threejane` under `tests/unit/vitest.config.ts` (`VITEST_MAX_THREADS=1`), a fail-closed-guard removal mutation, the `scripts/receipts/defi_3jane.mjs` decode-assert, `yarn typecheck` (after building `@vultisig/mpc-types` + `@vultisig/core-chain` to clear the unbuilt-dep wall), and the repo-exact prettier/eslint-health gates. Codex `gpt-5.3-codex` lane stalled 0-byte twice (session contention); GLM lane dead (429). Claude load-bearing — acceptable given the surgically-scoped diff.

## prior issues — status
|---|---|---|---|---|---|
| r1 (#853) | 1 | should-fix | net-new injectable `receiver` param diverged from mcp-ts (deliberately sender-bound), shipping an unvalidated arbitrary-recipient deposit path | ✅ fixed | `@7883759f`: `buildSupplyUsdc.ts:153` now throws on `receiver !== sender` (both `getAddress`-normalized), JSDoc rewritten to document sender-binding + the fund-loss rationale, test flipped to assert the throw. Closes the #853 should-fix. |

## verified — sender-binding is now fail-closed
1. **No path credits shares to a non-funder.** After the `receiver !== sender` guard (`buildSupplyUsdc.ts:153`), the only way through is `receiver === sender`. Both operands are `getAddress()`-normalized (L144, L148) so the equality is EIP-55-canonical — a lowercase/mixed-case duplicate of `from` passes (correct), a genuinely distinct address throws. The deposit calldata arg (L179), `result.receiver` (L190), and the funder are therefore provably the same 20 bytes. No divergence path.
2. **Fail-closed throw, not silent proceed**; default `receiver = from` (L146). Approve amount == deposit amount == `rawAmount` (exact bounded, not MaxUint256), spender == helper == deposit.to, decimals 6, 1000-USDC min — all unchanged from #853. Unsigned output preserved.
3. **#228 test proves it.** The distinct-receiver case now asserts `.toThrow(/sender-bound/)`; removing the guard fails exactly that test and nothing else (mutation survived `retry:1`). A new positive test covers the checksum-normalization path (lowercase `from` accepted). Mutation teeth confirmed.

## risk
Near-zero blast radius. The diff narrows behaviour (a previously-buildable distinct-receiver deposit now throws); on an unreleased feature in the same release cycle. The only consumer-visible change is that callers passing `receiver !== from` now get a clear error instead of calldata routing shares to an address the Helper would not honour — strictly safer.

## verdict
APPROVE. The arbitrary-recipient deposit path is closed; deposits are fail-closed sender-bound; tests prove the enforcement; the #853-verified contracts/selectors/decimals/bounded-approval are intact. RN type exports present; no additional changeset required (release-pr.yml is push-to-main-only; #853 already added the `@vultisig/sdk` minor for the module's first release).

🤖 reviewed via [Claude Code](https://claude.com/claude-code); codex `gpt-5.3-codex` + GLM lanes attempted (both unavailable this pass). MacBook local worktree.

PR #857, opened 2026-06-23
feat(sdk): add sdk.defi.balancer (v3 swap calldata)

gomesalexandre:
## what

New `sdk.defi.*` surface, kicked off with `sdk.defi.balancer.buildBalancerV3SwapCalldata` — a pure, **unsigned** Balancer v3 swap calldata builder. Given an already-fetched off-chain SOR quote (path(s) + expected amounts), it encodes the v3 BatchRouter `swapExactIn` / `swapExactOut` transaction and hands back `{ to, data, value, account, min/maxAmount }` for the wallet/MPC layer to sign.

- builds UNSIGNED calldata only — never signs, never broadcasts, and never hits an RPC (the `queryOutput` is reconstructed offline from the quote)
- slippage in bps -> a min-out (EXACT_IN) / max-in (EXACT_OUT) bound
- `userData` is consumer-**INJECTABLE** (affiliate / referral / pool-hook payload) and defaults to `0x` (neutral / off). The SDK is multi-consumer and never injects a brand of its own.
- exposed both as `sdk.defi.balancer.buildBalancerV3SwapCalldata(...)` (via a new `defi` getter on `Vultisig`) and as a top-level export.

## how

Thinly **WRAPS** `@balancer/sdk@6.1.1` per the DeFi lib-vs-handroll deep-dive: it's the only viem-native EVM-DEX lib that passes the RN/Hermes kill-filter — its entire dependency tree is `viem` + `@balancer-labs/balancer-maths` (both zero-ethers, zero-jsbi). The v3 Vault/BatchRouter encoding is genuinely fiddly, so wrapping the canonical encoder beats hand-rolling. Only the unsigned-tx builder is surfaced; `@balancer/sdk`'s signing / permit2-signature helpers are deliberately NOT re-exported.

Port source: mcp-ts `get_balancer_v3_quote` (which produces the SOR quote; this PR extends that quote into calldata).

## consumers unblocked

mcp-ts / agent-backend / vultiagent-app can now turn a Balancer v3 SOR quote into a ready-to-sign EVM tx through the SDK, instead of each consumer hand-encoding the BatchRouter call. Sits alongside the existing native/general swappers.

## test

- `yarn workspace @vultisig/sdk typecheck` -> 0 errors (clean baseline, my files add 0)
- 4 new vitest unit tests (EXACT_IN bound, EXACT_OUT bound, injectable-userData passthrough, input validation) + full `tests/unit/tools` suite green (119/119)
- lint clean on all touched files
[...]

PR #854, opened 2026-06-23
feat(sdk): add sdk.defi.glif (ICN stake/redeem)

gomesalexandre:
## what

Adds `sdk.defi.glif` — the first protocol under the new `sdk.defi.*` namespace. Ports the mcp-ts `build_glif_stake_icnt` / `build_glif_redeem_sticnt` tools into the multi-consumer SDK.

Two pure builders for GLIF x ICN liquid staking on Base:

- `buildGlifStakeIcnt({ from, amount, currentAllowance?, receiver? })` -> unsigned `[approve?, deposit]` (mint stICNT via `Pool.deposit`). The approve step is dropped when `currentAllowance >= amount`.
- `buildGlifRedeemSticnt({ from, amount, receiver? })` -> unsigned `[redeem]` (stICNT back to ICNT via `Pool.redeem`).

Both BUILD UNSIGNED EVM calldata ONLY. They never sign, never broadcast, never touch keys, and are deterministic/offline (no RPC). Pure crypto.

## why

DeFi consolidation: protocol calldata builders live in the SDK so agent-backend, mcp-ts and vultiagent-app stop re-encoding the same ABIs. This is the GLIF slice of that sweep.

- **Lib decision: HAND-ROLL viem** (GLIF ICN ABIs), per the lib-vs-handroll deep-dive. viem is already in-tree -> zero new deps, RN/Hermes-safe.
- **`receiver` is INJECTABLE** (defaults to `from`, never hardcoded). The multi-consumer SDK never pins a 'station'-style affiliate/recipient: the default routes shares/payout to the caller and exposes no third-party payout surface; consumers opt into a different receiver explicitly. The `redeem` owner is always pinned to `from`.

## consumers unblocked

agent-backend / mcp-ts / vultiagent-app can call `defi.glif.buildGlifStakeIcnt(...)` / `buildGlifRedeemSticnt(...)` and hand the returned `{ to, value, data }` tuples to their own signer instead of re-encoding GLIF ABIs per repo.

## testing

- typecheck: 0 errors (clean)
- vitest unit: 7/7 pass (`tests/unit/defi-glif.test.ts`) — decodes approve/deposit/redeem calldata, allowance fast-path, injectable-receiver, owner-pinning, input rejection.
- runnable receipt `scripts/receipts/defi_glif.mjs` builds + decodes unsigned stake/redeem calldata (no broadcast).

```
=== sdk.defi.glif receipt — GLIF x ICN (Base) ===
pinned addresses: {
[...]

PR #844, opened 2026-06-23
feat(sdk): add sdk.dex.quotes (uniswap-v2 + balancer on-chain quotes)

gomesalexandre:
## what

Two read-only on-chain DEX quote primitives, exported from the package index:

- **`uniswapV2Quote(...)`** - hand-rolled constant-product `getAmountOut` (3-line bigint, 0.3% fee baked in via `997/1000`), fed by **live on-chain** factory -> pair -> reserves + token-metadata reads through the SDK's own `evmCall`. Ported from mcp-ts `uniswap/v2-quote.ts`. Also exports the bare `getAmountOut` for callers who already hold reserves.
- **`balancerQuote(...)`** - canonical pool math via **`@balancer-labs/balancer-maths`** (zero-dep, RN-safe - verified no node builtins in dist). Uses the lib's `Vault.swap()` (the exact math the on-chain Balancer Vault + SOR run) instead of hand-rolling the weighted/stable invariant. Works for every pool type the lib supports (weighted, stable, gyro, reclamm, ...).

Read-only only: decode/read/quote. No Router02/Vault calldata, no signing, no broadcast. The agent-judgement layer (intent-match, fabricated-vs-grounded, prompt-injection) stays in the backend - this is pure crypto/math.

## why

mcp-ts's `get_uniswap_v2_quote` already imports `evmCall` from the SDK but keeps the quote math local; agent-backend would have to re-port the same constant-product formula. Folding both into `@vultisig/sdk` gives one canonical home for the math and stops per-repo drift. Balancer specifically: hand-rolling the weighted/stable invariant is a footgun, so we lean on Balancer's own published maths lib.

## consumers unblocked

- mcp-ts `get_uniswap_v2_quote` -> wrap `uniswapV2Quote`; the local `getAmountOut` + `_erc20` decode helpers go away.
- mcp-ts `get_balancer_v3_quote` -> can move from the SOR HTTP proxy to true on-chain `balancerQuote` math once pool-state reads are wired.
- agent-backend swap/quote surfaces -> import `balancerQuote` / `uniswapV2Quote` directly instead of re-implementing.

## receipt

\`\`\`
$ yarn tsx scripts/receipts/dex_quotes.mjs

=== uniswap-v2 (LIVE on-chain, Ethereum mainnet) ===
1 WETH -> 1650.178498 USDC  (pair 0xB4e16d0168e52d35CaCD2c6185b44281Ec28C9Dc, fee 30bps)
  reserves: in=4965972579752165002215 out=8221049553442
[...]

PR #866, opened 2026-06-24
fix(sdk): defi follow-ups from merged-pr suggested changes

gomesalexandre:
Tackles suggested changes from merged SDK PRs — four waves, all green.

### glif (#854) — assetAmount / shareAmount rename (breaking)
`BuildGlifStakeParams.amount` -> `assetAmount`, `BuildGlifRedeemParams.amount` -> `shareAmount`. Stake inputs are ICNT asset units, redeem inputs are stICNT share units — `amount` was ambiguous. Result fields + tests updated to match.

### balancer (#857) — 3 fixes
1. **Cross-path token pair validation** — throws when any path's input/output token differs from the first path's, so mixed-pair inputs can't encode cross-asset amounts into the wrong slots.
2. **Unknown swapKind guard** — was silently defaulting anything != EXACT_OUT to EXACT_IN. Now throws explicitly.
3. **ZERO_ADDRESS sentinel + post-buildCall assertion** — replaced private-field reach-through (always returned `undefined`) with ZERO_ADDRESS sentinel; asserts router address non-zero after buildCall.

### policy/amountDriftPct (#847) — NaN bypass fix
`Infinity / Infinity = NaN`; `NaN > threshold` is always `false`, silently bypassing BLOCK checks on extreme amounts. Moved ratio into bigint space (scale * 1e9, convert once) to avoid the overflow path.

### validate/recipientSanity (#837) — never-throws contract
`input.recipient.trim()` throws for non-string inputs from untyped JS callers. Added `String(input.recipient ?? '')` coercion.

### utils/addressFormat (#838) — valoper HRPs in cosmos classification
`cosmosHRPs` was built only from account HRPs. `cosmosvaloper1...` etc. were mis-classified as `unknown`. Fixed by merging `cosmosValoperByChain` values into the set.

### test/dexQuotes (#844) — malformed symbol() ABI mock
`''.padStart(64, '0').replace(/0$/, '')` emits 63 hex chars — a malformed ABI word. Fixed to correct 32-byte offset word.

### prep/splTransfer (#824) — missing Buffer import
`Buffer.from()` was used without import. Fails in browser/RN without the polyfill. Added `import { Buffer } from 'buffer'` matching the pattern in `thorchainMsgDeposit.ts`.

### decode/fromToolResult (#848) — from_chain missing from chainHint
`inferFamily()` checks `args['from_chain']` as a fallback but `chainHint` (passed to `decodeEvmTx`/`decodeCosmosTx`) didn't. When chain was supplied only via `from_chain`, `Envelope.chain` was empty even though the chain was known. Added the same fallback.

### token/resolveContract (#819) — 2 fixes
1. **name() revert resilience** — `name()` is optional in ERC-20 and some contracts revert on it. `Promise.all` would propagate the throw and kill the whole resolution. Added `.catch(() => ({ data: undefined }))` so it falls back to symbol.
2. **Number.isInteger on ERC-20 decimals** — `Number.isFinite` alone passes fractional values like `6.5`. Added explicit integer check to match the `isValidByteDecimals` pattern used for CW20/SPL paths.

### vault/nativeSwapBroadcastGuard (#809) — VaultError instead of plain Error
Per SDK coding guidelines, vault service errors use `VaultError` + typed `VaultErrorCode` so callers can programmatically distinguish error classes. Four throws updated: `InvalidConfig` (expired), `UnsupportedChain`, `NetworkError` (missing inbound address), `InvalidVault` (vault address mismatch).

## testing

All affected test suites green. TypeScript clean.


NeOMakinG (approved):
fix-and-approve. This PR collects the should-fix / suggestion items our reviews left on three merged DeFi PRs (#858 arkis, #857 balancer, #854 glif), and gomes added a second wave of defensive hardening mid-review. I verified each follow-up resolves its prior finding, fixed one over-correction in the arkis change in-code, and confirmed no calldata regression.

## testing approach
3 lanes: Claude load-bearing (on-chain viem error-class probe + mutation), Codex (0-byte stall — known contention), GLM (ECONNRESET — dead). Worktree off `1c3335b8`; built `@vultisig/core-chain`; ran affected suites single-threaded (`VITEST_MAX_THREADS=1`, forks/singleFork, setupFiles:[] to dodge the unbuilt mpc-types wall).

- glif suite **15/15**; arkis (buildSupplyTx) **9/9**; balancer **7/7**; + my new resolvePoolKind suite **5/5**; second-wave policy/recipientSanity/dexQuotes/balancer/glif/arkis combined **100/100**. Total affected: green.
- tsc clean on all touched files; prettier clean; eslint (main + health/complexity) exit 0.
- Live viem probe (mainnet): sDAI `asset()` → SUCCESS; USDC → revert (`ContractFunctionRevertedError`); **WETH → empty-data (`ContractFunctionZeroDataError`)**.
- CI was still queued at post time (slow infra); all three jobs reproduced green locally.

## prior findings — resolution
| PR | prior finding | resolved? |
|----|---------------|-----------|
| #858 arkis | bare `catch{}` failed-open: transport/RPC errors silently mis-classified pool as `agreement` | ✅ now re-throws non-revert errors — **but over-narrowed; fixed in-code (see below)** |
| #857 balancer | vestigial `(swap as unknown as {swap:{to}}).swap.to` reach-through reading a nonexistent field (always `undefined`) | ✅ replaced with `ZERO_ADDRESS` sentinel + post-buildCall non-zero router assertion; calldata byte-unchanged (buildCall ignores `queryOutput.to`) |
| #854 glif | asset-vs-share field-naming footgun (`amount` ambiguous) | ✅ `amount`→`assetAmount` (stake/ICNT) / `shareAmount` (redeem/stICNT); pure rename, same bigint into same encode slot, all selectors/owner/receiver pins intact |

## blocking
None.

## what I fixed in-code (`1c3335b8`)
The arkis follow-up **over-corrected**. The narrowed guard only swallowed `ContractFunctionRevertedError`, re-throwing everything else. But a standard Agreement contract whose `asset()` selector hits a non-reverting fallback returns **empty data (`0x`)**, which viem surfaces as `ContractFunctionZeroDataError` — **not** a revert. Live-proven: WETH returns `0x` for `asset()` (→ ZeroData) while plain ERC-20s like USDC revert. So the PR fixed the transport-swallow but **broke agreement-resolution for the empty-data subclass of Agreements** — those would now throw instead of resolving to `agreement`.

Fix: swallow both `ContractFunctionRevertedError` and `ContractFunctionZeroDataError` (the two on-chain "no `asset()`" signals, whether wrapped in `ContractFunctionExecutionError` or raised directly) and still re-throw genuine transport/RPC errors. Added the first test coverage for `resolvePoolKind` (the changed file had none): 4626-success, revert→agreement, **empty-data→agreement (regression test)**, transport-error rethrow, malformed-address pre-RPC reject. Mutation-proven: against the PR's original guard the empty-data test fails (1/5); against the fix all 5 pass.

## should-fix
- The second-wave validation branches ship without dedicated tests (#228 inert-coverage class): balancer unknown-swapKind reject, multi-path token-pair mismatch, `amountDriftPct` NaN-saturation, `recipientSanity` String() coerce, valoper-HRP cosmos classification. All fail-closed and I verified `amountDriftPct` by hand (old `NaN` bypass at ~1e400 magnitudes → new bigint-space ratio correctly returns finite >threshold = BLOCK; normal range byte-identical), but a one-test-per-branch follow-up would pin them.

## risk
Low. arkis/balancer/glif are unsigned calldata builders; the only behavior change to a real fund path is the arkis empty-data correctness fix (restores resolution, fail-closed on transport) and the `amountDriftPct` NaN-bypass fix (tightens a policy check — fail-closed). glif is a pure rename (calldata byte-unchanged). balancer ZERO_ADDRESS swap leaves built calldata byte-unchanged (buildCall ignores the sentinel) and adds a fail-closed router assertion.

## verdict
APPROVE. All three prior findings resolved; the arkis over-correction fixed in-code with mutation-proven coverage; no calldata regression; changeset present and well-formed; RN exports unaffected.

---
🤖 reviewed with multi-lane fan-out (Claude load-bearing; Codex/GLM dead). fix-and-approve per gomes policy.

PR #867, opened 2026-06-24
fix(cli): poll for final tx confirmation after broadcast (audit F1)

neavra:
## Summary

After a `tx_ready` tx is signed + broadcast, the agent session emitted a single `pending` `tx_status` and recursed. A headless caller (`agent ask` / pipe) could therefore mark an operation **complete** while the tx later **reverted, expired, or was dropped** — the audit's tx-reliability **F1** (HIGH, CLI-fixable).

This adds bounded **post-broadcast confirmation polling** at the single broadcast chokepoint in `session.ts`:

- After emitting `pending`, poll `vault.getTxStatus` until the tx reaches a final state, then emit `confirmed` / `failed`.
- Emit `timeout` when the bounded poll budget (~120s: 3s × 40) is exhausted — the tx may still confirm later; callers can re-check with `vultisig tx-status`.
- Transient RPC/network errors are treated as "not final yet" and retried until the budget is spent.
- **Best-effort & non-fatal**: if the chain can't be resolved or the vault doesn't expose `getTxStatus` (e.g. unit-test `this`), the already-emitted `pending` status stands.

## Changes

- `agent/session.ts` — `confirmBroadcastedTx` + `txConfirmSleep`, wired into the `serverTxStoredFromStream` broadcast block; poll cadence as instance fields so tests drive it without real waits.
- `agent/ask.ts` — record the latest per-tx `status`, **deduped by hash** (one tx now emits multiple lifecycle events).
- `agent/types.ts` / `agent/pipe.ts` — pipe `tx_status` event gains a `timeout` status.
- `agent/executor.ts` — export `resolveChain` for reuse (no behavior change).

## Tests (ship in this PR)

- `agent/__tests__/sessionTxConfirm.test.ts` — confirmed, failed (revert), timeout, transient-error-then-confirm, and no-poll fallback, all driven through `processMessageLoop`.
- Updated `ask.test.ts` (status field + dedup) and the `sessionConfirmGate` / `sessionRecovery` harnesses (gain `confirmBroadcastedTx`; still assert `pending`-only with no vault).

## Validation

- `yarn test:cli` — 21 files, 310 tests green (serial, no worker crashes)
- `yarn typecheck` (root tsc + `@vultisig/cli`), `yarn lint`, `yarn knip`, `yarn format:check` — all green
- Changeset: `@vultisig/cli` patch

## Scope

Touches only the broadcast block of `session.ts` (the F1 seam) plus the lifecycle plumbing in `ask`/`pipe`/`types`. Does **not** touch the auth/initialize region (sibling fix-07).

## Summary by CodeRabbit

* **New Features**
  * After broadcasting, the CLI now performs bounded confirmation polling (in headless/ask flows), moving transactions from **pending → confirmed/failed** or emitting **timeout** when the poll budget is exhausted.
  * `ask` results and `tx_status` pipe events now include a typed transaction lifecycle status, with updates **deduplicated by transaction hash**.
* **Bug Fixes**
  * Confirmation polling is best-effort: if status can’t be verified (including transient RPC errors or missing support), the already-emitted **pending** state is preserved.
* **Tests**
  * Added/updated unit tests covering polling success/failure, retries, **timeout**, abort behavior, and deduplication.


gomesalexandre (commented):
## tl;dr

Solid fix for the audit F1 gap. The polling logic is well-bounded, abort-aware, and non-fatal. A few things worth flagging — none are fund-safety blockers but one is a real behavioral gap worth fixing before merge.

---

## findings

**preferably-blocking: `confirmBroadcastedTx` blocks the next-turn `processMessageLoop` recursion for up to ~117s**

The call order in `session.ts` is:

```
ui.onTxStatus(txHash, ..., 'pending', ...)   // line 488
await this.confirmBroadcastedTx(...)          // line 489 — blocks HERE
await this.processMessageLoop(null, ui, depth + 1)  // line 492
```

That's fine for pure ask/pipe scenarios where the session ends after signing. But it gets interesting if the server sends a tx_ready inside a multi-turn tool loop — the poll blocks the next `processMessageLoop` call for the full budget (~117s), meaning any follow-up tool results the server is waiting for are delayed by that wall time before the next turn fires. The abort gate helps (Ctrl-C unblocks), but an agent looping over multiple txs back to back (say a batch) would serially accumulate that wait per tx.

Not a fund-loss bug per se — correctness holds — but it's a throughput/latency cliff for multi-tx agent flows the audit may not have scoped. Worth at least a comment noting this known behavior, or gating the poll on `depth === 0` / "no more pending results expected" to avoid stacking.

**should-fix: `any` cast on `vault.getTxStatus` is opaque and fragile**

```ts
if (!chain || typeof (this.vault as any)?.getTxStatus !== 'function') return
// ...
const result = await (this.vault as any).getTxStatus({ chain, txHash })
```

The comment says "mirror executor.waitForEvmReceipt" but the executor path uses a typed interface. `VaultBase` presumably doesn't expose `getTxStatus` — that's fine — but casting to `any` means TypeScript won't catch a rename/sig change on `getTxStatus` in the vault SDK. A local interface narrowing (`interface TxPollingVault { getTxStatus(args: { chain: Chain; txHash: string }): Promise<{ status: string }> }`) + `instanceof` or a type guard would make refactoring safer without changing runtime behavior.

**should-fix: no test covers the `depth > 0` (multi-turn) blocking scenario**

All seven tests in `sessionTxConfirm.test.ts` call `processMessageLoop` at depth 0 with a single tx. None exercise the case where sign happens inside a tool loop (depth > 0) and then the poll runs before the next recursive call. That's the scenario flagged above — a regression test here would also serve as documentation of the intended behavior.

**suggestion: `status` field in `AskResult.transactions` typed as `string` instead of the actual union**

```ts
// ask.ts
status?: string
```

The pipe type already has the precise union (`'pending' | 'confirmed' | 'failed' | 'timeout'`). `AskResult` could mirror it to give callers type-safe branching without a string comparison.

**suggestion: `getTxStatus` result only checks `'success'`/`'error'` — what about other vault-returned statuses?**

If the vault SDK ever returns something other than `'pending'`/`'success'`/`'error'` (say `'unknown'` or `'mempool'`), the poll treats it as "not final yet" and burns through the budget. That's the right safe default, but a comment noting the exhaustive set of expected `result.status` values from the vault SDK would help future readers know whether this is intentional or a gap.

**praise: the headless guard + abort check are clean and well-placed**

Gating on `askMode || viaAgent` is the right call — no poll regression on TUI. Abort bail on every loop iteration is exactly right. The `txConfirmSleep` extraction for test controllability is a nice pattern.

**praise: replacing the hand-rolled EIP-712 encoder with viem's `hashTypedData` is excellent hygiene**

The removed ~200 lines of custom struct encoding were a latent mismatch risk (domain field filtering, chainId coercion, low-S fold). Delegating to viem + adding the recover-verify gate is a meaningful security improvement beyond the F1 scope. Good catch to include in the same PR.

---

reviewed via claude code 🤖

gomesalexandre (approved):
## prior issues - status

|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | synchronous poll blocks multi-turn agent flows at depth > 0 | ✅ fixed | `6282916e9` at `clients/cli/src/agent/session.ts:499` — `if (depth === 0)` guard |

---

## r2 verdict: LGTM ser

neavra's fix commit hits the brief cleanly. walking it:

**depth guard (the blocker):** `confirmBroadcastedTx` is now behind `if (depth === 0)` at session.ts:499. Recursions at depth > 0 emit an honest `pending` and keep moving - the poll budget (~117s) no longer stacks per leg in batched/multi-turn flows. Comment in the code is clear on the why. Regression test added in `sessionTxConfirm.test.ts` ("skips the blocking poll inside a multi-turn tool loop") covers the path explicitly.

**typed status union:** `TxLifecycleStatus = 'pending' | 'confirmed' | 'failed' | 'timeout'` introduced in `types.ts` and threaded through `ask.ts`, `pipe.ts`, and `tui.ts` - the `as 'pending' | 'confirmed' | 'failed' | 'timeout'` cast in pipe.ts is gone. Clean.

**`getTxStatus` cast removal:** `(this.vault as any).getTxStatus` is now `this.vault.getTxStatus` with a `typeof` runtime guard for test stubs. The comment in the commit explains why the optional chaining isn't strictly needed at the type level but stays for runtime safety - that's fine.

**changeset:** `.changeset/cli-post-broadcast-confirmation.md` updated with the depth-0 scoping detail. `@vultisig/cli: patch` bump. Correct package name, correct bump type.

**tests:** 340/340 CLI tests pass (`yarn test:cli`). New dedup test ("pending → confirmed" update in place) + depth > 0 skip test both green.

**typecheck:** one pre-existing error (`getCredentialsPath` not in stale local `dist/` of `@vultisig/client-shared`) - this is a stale-dist issue in the local clone from a prior merged PR (169f33ec6), not introduced here. CI should be fine since it rebuilds from source.

**changeset file covers the PR scope:** yes - the blocking confirmation wait being scoped to depth 0 is now documented in the changeset body. ✅

---

suggestion: the `txConfirmMaxPolls = 40` constant (3s × 39 ≈ 117s ceiling) is still a magic number on the session instance. Low priority - fine for now since it's already doc'd in a comment - but a `TX_CONFIRM_MAX_POLLS` named export would let integration tests and the changeset docs reference the same value. Not blocking.

---

PR #868, opened 2026-06-24
fix(cli): stable agent ask --json envelope on stdout (success + error)

neavra:
## Summary

`agent ask --json` had four output-contract bugs that broke headless callers:

1. **Success → empty stdout.** `executeAgentAsk` redirects `console.log`→`process.stderr.write` for the whole turn (so SDK/MPC chatter can't pollute stdout), and `outputJson` emitted the success envelope through that redirected `console.log` — so the envelope landed on **stderr** and stdout was empty in `--json` mode.
2. **Error → different shape.** The catch wrote a flat `{error,code}` to stdout, a different schema from the success `{success,v,data}` envelope.
3. **SSE error frame → exit 0.** A mid-stream backend/SSE `error` frame only called `onError` (stderr write, no throw); the turn resolved normally and the command exited 0 — false success for a caller branching on exit code.
4. **Missing IDs.** Success `data` carried `session_id` only — no `conversation_id`, no per-tool-call ids.

## Changes

- **`lib/output.ts`** — `outputJson` / `outputErrorJson` write straight to `process.stdout` instead of via `console.log`, so the v1 envelope survives the `agent ask` redirect and always lands on the machine-readable channel. Behaviour is identical for every other caller (`console.log` already wrote to stdout with a trailing newline).
- **`agent/ask.ts`** — `AskInterface` captures per-tool-call `id`s and the first backend/stream `error` frame, surfaced via `AskResult` (`toolCalls[].id`, `error`).
- **`commands/agent.ts`** — unify success + error onto one v1 envelope on **stdout**: success `{success:true,v:1,data:{conversation_id,session_id,response,tool_calls,…}}`, error `{success:false,v:1,error:{message,code,conversation_id}}`. Exit non-zero when a backend/stream error frame arrives.

## Test (red-then-green, ships in this PR)

`clients/cli/src/commands/__tests__/agentAskJson.test.ts` drives the real `AskInterface` with a mocked `AgentSession`:
- success → v1 envelope on stdout with `conversation_id` + tool-call `id`; envelope **not** on stderr; exit 0.
- SSE/backend error frame → error envelope on stdout with `conversation_id`; exit non-zero.

Both verified RED against pre-fix code (success → empty stdout; SSE error → exit 0), GREEN with the fix.

## Validation

- `yarn test:cli` — 301 passed
- `yarn typecheck` — clean
- `yarn lint` — clean
- format:check (project prettier config) — clean
- changeset: `@vultisig/cli` patch

## Scope notes

- Pipe-mode (`--via-agent`) NDJSON contract is unchanged.
- fix-04 coordination (referenced in the original task) is **moot** (#682 already gates headless signing); error codes stay consistent with fix-03 / fix-07.

## Summary by CodeRabbit

* **New Features**
  * `agent ask --json` now returns a consistent v1 JSON envelope for both success and error responses.
  * Output now includes stable conversation and tool-call identifiers, plus partial turn data when available.

* **Bug Fixes**
  * Fixed JSON output being sent to the wrong stream.
  * Mid-stream errors now correctly produce a non-zero exit code instead of appearing successful.
  * Transaction updates now include lifecycle status in the returned data.


gomesalexandre (approved):
solid work ser - all 4 described bugs are fixed and the fund-safety concern that came up during iteration (broadcast tx hash surviving error paths) is genuinely handled. here's the full synthesis.

---

## bug-by-bug verification

**bug 1 - success envelope on stderr:** fixed. `outputJson` and `outputErrorJson` in `lib/output.ts` now call `process.stdout.write` directly instead of `console.log`. the `console.log → stderr` redirect in `executeAgentAsk` no longer intercepts the structured output. the test explicitly asserts `expect(stderr.join('')).not.toContain('"success"')`. locked.

**bug 2 - error envelope wrong shape:** fixed. the old flat `{ error: message, code }` catch path is gone. both the SSE-error-frame path and the thrown-error catch path now go through `outputAskError` which emits the unified `{ success: false, v: 1, error: { message, code, conversation_id } }` shape. locked.

**bug 3 - exit 0 on SSE error:** fixed. `AskResult` now carries `error?: { message, code }` set from `onError`. after `ask()` returns, `executeAgentAsk` checks `result.error` and exits non-zero. the SSE-error test asserts `expect(exitCode).not.toBe(0)`. locked.

**bug 4 - empty stdout on success:** covered by bug 1. locked.

**bonus: broadcast tx surviving error path (fund-safety):** the commit history shows this was caught during review iteration. the SSE-error-frame path passes `result` to `outputAskError` which carries `transactions`/`tool_calls`/`response` in a `data` block. the thrown-after-broadcast path uses `ask.partialResult()` in the catch block. both paths are tested. the implementation correctly scopes `data` to non-empty payloads so lean auth failures don't get a spurious empty `data: {}`. solid.

---

## adversarial pass

**`process.stdout.write` approach - is it right?**

yes, this is the correct fix. the root issue was that `outputJson` was coupled to `console.log` which got monkey-patched to `stderr` by `executeAgentAsk`. writing directly to `process.stdout` makes the output immune to any `console.log` override, now or in the future. the approach is also consistent with how the rest of `executeAgentAsk` already writes its human-mode output (all the `process.stdout.write(...)` calls for `session:`, `tx:`, etc. already bypassed `console.log`). it's the right primitive.

**edge case: `applyOutputTransforms` on the error envelope**

`outputErrorJson` passes the entire error shape through `applyOutputTransforms`. in `quietMode`, `stripEmpty` strips fields where `v == null || v === ''`. `AgentErrorCode` values are all non-empty string literals so `code` is safe. `message` from `normalizeAgentError` has a fallback to `'Unknown error'` so it won't be empty. `conversation_id` could theoretically be `''` on early auth failure - in that case `quietMode` would strip the key. this is an edge case but worth noting: a headless caller in quiet mode who hits an auth failure before session init would get `error: { message, code }` without `conversation_id`. not a blocker but it's a real scenario.

**`status` spreading on `onTxStatus`**

```ts
this.transactions.push({ hash: txHash, chain, ...(status ? { status } : {}), explorerUrl })
```

the `status` parameter comes from the backend SSE frame. the `if (status)` guard drops the field when status is falsy (empty string or undefined). for the `undefined` case that's correct, but an explicit empty string `''` from the backend would also be silently dropped. probably fine given the backend emits enum strings, but the asymmetry with `explorerUrl` (which is pushed as `undefined` explicitly) is slightly inconsistent. not a blocker.

**double JSDoc block in `agent.ts`**

there are two back-to-back `/** ... */` blocks at lines 98-127: one describing `executeAgentAsk`'s output format, then immediately another describing `outputAskError`. the first one is orphaned - it's positioned as if it documents `executeAgentAsk` but the function itself is further down. the second JSDoc correctly precedes `outputAskError`. cosmetic nit only.

**human-mode SSE error path - no test**

the `agentAskJson.test.ts` file (named correctly) only tests `json: true`. in non-JSON mode, when `result.error` is set, `outputAskError(wantsJson=false, ...)` writes to `stderr` and exits non-zero. this is correct behavior (and better than the old exit-0 false success) but it's untested. no success line, no response, just the stderr error line. headless callers using text mode would see a behavior change here. not a regression since the old behavior was wrong (exit-0 false success), but worth a comment.

**test mock: `driver.run` throws synchronously - matches the claim?**

the "broadcast THEN thrown follow-up failure" test has `driver.run` throw at the end. `FakeSession.sendMessage` does `await driver.run(callbacks)`. the `onTxStatus` and `onToolResult` fire before the throw, so `AskInterface.transactions` and `toolCalls` are populated before `sendMessage` rejects. then `ask()` propagates the throw. the catch in `executeAgentAsk` calls `ask.partialResult()` which snapshots those populated fields. the test correctly models the described scenario. ✓

**the `!conversationId` guard in catch**

```ts
if (partial && !conversationId) conversationId = partial.sessionId
```

this is fine - `conversationId` is only set after `ask()` returns on the success path. if `ask()` threw, `conversationId` is `''`. the guard is effectively always-true for the throw case. the condition reads a bit defensively but it's not wrong. minor: `partial.sessionId` itself could be `''` if `session.getConversationId()` returns `''` (before the backend assigns it on init). in that case the error envelope gets `conversation_id: ''` which is accurate.

---

## suggestions

**suggestion:** the stale orphaned JSDoc block at the top of `executeAgentAsk` (the "Output format (text):" / "Output format (--json):" block) should either be moved directly above the `export async function executeAgentAsk` declaration or deduplicated with the `outputAskError` JSDoc above it. as-is it reads as documentation for `outputAskError`, which is not what it describes.

**suggestion:** consider a `human mode SSE error` test case in a follow-up - even just asserting exit-code-1 and the stderr error line for the non-json error path. the current test file name `agentAskJson.test.ts` signals it's json-focused so skipping it here is defensible.

**suggestion:** the `status` falsy-guard is subtly different from `explorerUrl`'s behavior (explicit `undefined` vs omit). if the backend can send `status: ''`, consider normalizing: `status: status || undefined` before the spread, so both falsy cases behave identically and the object shape is more predictable.

---

## verdict

no blockers. the 4 described bugs are verifiably fixed, the fund-tracking concern is handled on both the SSE-error-frame and the throw path, test coverage is real (red-then-green claim is credible given the test structure). suggestions above are non-blocking cleanup.

reviewed via claude code 🤖

PR #869, opened 2026-06-24
fix(cli): harden agent conversation auth (retry/fallback/signal + refresh modeling)

neavra:
## Summary

Makes agent auth resilient on **every** conversation path. Backlog item *sdkcli-fix-07* (verified Fugu×Opus reconciliation, HIGH, CLI-fixable).

A revoked-but-unexpired cached token (revoked server-side, still inside the 60s client expiry buffer) used to crash a fresh headless `agent ask` with `Authentication failed: …` because only the resume + send-message paths wrapped their request in a clear→reauth→retry; the fresh-conversation `createConversation` did not. This PR routes all of them through one helper and adds graceful fallback + an observable signal.

## Changes

- **`withAuthRetry` helper** (`session.ts`) — runs an authed request and on a 401/403 does `clearCachedToken → authenticateVault → setAuthToken → saveCachedToken → retry once`. Now wraps the fresh-convo `createConversation` (finding **a**), the resume `getConversation`, the send-message stream (replacing the inline while-loop; batch-restore preserved), and the `agent sessions list`/`delete` callers (`commands/agent.ts`, via a small cache-free mirror).
- **Resume-401 fallback** (finding **b**) — a resume whose 401 survives the single retry (or fails for any other reason) falls back to a fresh conversation instead of throwing uncaught.
- **`SESSION_NOT_FOUND` signal** (finding **c**) — the resume→new-conversation fallback emits a typed, **non-fatal** `ui.onError(…, SESSION_NOT_FOUND)` carrying the new conversation id, so a headless caller knows prior context was dropped and can persist the new id (was silent).
- **`refresh_token` modeling** (finding **d**) — `access_token`/`refresh_token` added to `AuthTokenResponse`; `authenticateVault` surfaces `refreshToken`; the token cache persists it (0o600). The backend's `POST /auth/refresh` exists and is noted as a future exchange; the retry path re-auths via MPC re-sign today (always available, no rotation dependency).
- **Tests** (finding **e**) — first `auth.ts` unit tests: EIP-191 personal_sign hash cross-checked against `viem.hashMessage` + the canonical "hello world" vector, DER→65-byte formatting, and signing-retry classification. Plus `sessionAuthRetry.test.ts` red-then-green locks for the fresh-convo 401 retry and the resume fallback/signal.

## Test plan

- `yarn test:cli` — 21 files / **315 tests** green (incl. 16 new)
- `yarn typecheck` — clean
- `yarn lint` / `yarn knip` / `yarn format:check` — clean
- Changeset: `@vultisig/cli` patch

## Scope notes

- Sibling task *fix-03* also edits `session.ts` but in the broadcast block (different function) — no overlap.
- Out of scope (separate backlog items): the challenge message text (backend verifies signed bytes verbatim), low-S enforcement (auth backend accepts high-S), final-tx-outcome posting, SSE reconnect.

## Summary by CodeRabbit

* **New Features**
  * Added one-time hardened auth retry across session resume/create, streamed messaging, and disconnected-turn recovery, with graceful recovery for revoked-but-unexpired tokens.
  * If `--session-id` resume can’t be fetched after the single retry, the CLI creates a fresh conversation and emits a typed, non-fatal `SESSION_NOT_FOUND` carrying the new conversation id.
  * Capture and persist refresh tokens with safer token-cache permissions.

* **Bug Fixes**
  * Prevented repeated auth storms during recovery; retry is bounded and preserves pending tool-results after non-auth failures.
  * Fixed `agent ask --json` to keep `SESSION_NOT_FOUND` from initialize into the first turn.

* **Tests**
  * Added/expanded unit test coverage for auth hashing/signatures, retry/fallback flows, and refresh-token preservation.


rcoderdev (approved):

clients/cli/src/agent/session.ts
```diff
@@ -140,38 +140,41 @@ export class AgentSession {
...
+        // to a fresh conversation rather than hard-failing (finding b — the old
+        // 401 branch retried getConversation once with no fallback and threw
+        // uncaught on a second failure), but surface a typed, NON-FATAL signal
+        // so a headless caller knows prior context was dropped and can persist
+        // the NEW conversation id (finding c — the fallback used to be silent).
+        this.conversationId = null
+        this.historyMessages = []
+        const conv = await this.withAuthRetry(() => this.client.createConversation(this.publicKey))
+        this.conversationId = conv.id
+        ui.onError(
```
rcoderdev, line 171: One ask-mode path still drops this signal. `initialize()` emits `SESSION_NOT_FOUND` before `executeAgentAsk` calls `ask.ask()`, and `AskInterface.ask()` resets its stored `error` at the start of the turn. So `vultisig agent ask --json --session <stale-id>` can return a success envelope for the new conversation instead of the promised non-zero `SESSION_NOT_FOUND` envelope. Please carry initialize-time errors into the ask result or avoid clearing this one before the first send.

PR #870, opened 2026-06-25
Fix Maya native swap output decimals

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Native swap quotes now show output amounts in the destination coin’s base units, improving consistency across quote displays.
  * Added extra checks for very small values to avoid misleading near-zero swap outputs.

* **Tests**
  * Added coverage for native swap quote normalization on supported chains, including different decimal formats.


johnnyluo (approved):
LGTM

PR #873, opened 2026-06-25
chore: version packages

rcoderdev:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

- [#867](https://github.com/vultisig/vultisig-sdk/pull/867) [`ddd08af`](https://github.com/vultisig/vultisig-sdk/commit/ddd08af883a1b2ee2f72dac4d406782de9090672) Thanks [@neavra](https://github.com/neavra)! - Agent: poll for final on-chain confirmation after broadcasting a signed tx
  (audit F1). A `pending` `tx_status` only means "broadcast accepted" — the tx can
  still revert, expire, or be dropped. After broadcast the session now polls
  `vault.getTxStatus` until the tx reaches a final state and emits `confirmed` /
  `failed`, or `timeout` when the bounded poll budget (~120s) is exhausted. The
  `ask` result records the latest per-tx `status` (deduped by hash), and the pipe
  `tx_status` event gains a `timeout` status. Best-effort and non-fatal: when the
  chain can't be resolved or the vault can't poll status, the existing `pending`
  status stands. The blocking confirmation wait is scoped to the top of the
  message loop (depth 0 — the single-tx ask/pipe case); inside a multi-turn tool
  loop a leg keeps its honest `pending` instead of stacking the poll budget per
  tx. The shared `pending | confirmed | failed | timeout` union is now threaded
  through the ask result, pipe event, and UI callback without unchecked casts.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction handling in the CLI so broadcasted transactions are now monitored until they reach a final state.
  - Added clearer outcomes for transaction status updates, including confirmed, failed, and timeout results.
  - Made transaction status reporting more reliable by keeping the latest status for each transaction and avoiding incorrect intermediate states.


gomesalexandre (approved):
#867 (neavra post-broadcast confirmation) is the only changeset here - looking good, ship it

PR #874, opened 2026-06-25
fix(core-chain): dedupe Solana broadcast on duplicate-signature error

neavra:
## What

`broadcastSolanaTx` (`packages/core/chain/tx/broadcast/resolvers/solana.ts`) now classifies a duplicate-signature rejection from `sendRawTransaction` — `"already been processed"` / `AlreadyProcessed` — as an **idempotent success** (`return`) instead of routing to `verifyBroadcastByHash`. This mirrors the existing TON / UTXO / Cosmos dedupe guards.

## Why

Non-EVM broadcast idempotency. All other non-EVM chains already dedupe at the BroadcastService/resolver layer (TON `41209f23`, QBTC `84a29500`, Cosmos/UTXO via `isInError(...,'already known'|'tx already exists'|…)`). **Solana was the only remaining resolver without a duplicate-signature→success guard** — it relied solely on `verifyBroadcastByHash` as a fallback.

Without this, a headless agent that broadcasts a Solana send and hits an ambiguous failure after the node already accepted the tx could see a "failed" action and blindly re-broadcast the same signed payload.

`verifyBroadcastByHash` remains the fallback safety net for genuinely ambiguous (non-duplicate) failures; `VaultBase.broadcastTx` stays single-shot (the chain-layer resolver is the right place for this guard).

## Known trade-off (reviewed, accepted)

Review flagged this as a conscious fund-handling trade-off — **accepted** and now documented in-code:

- **Optimistic at the broadcast layer.** Solana returns `AlreadyProcessed` for any signature it has already seen — including a transaction that was *processed but reverted on-chain*. The guard returns success **without verifying the execution outcome**, so a processed-but-failed Solana tx is reported as success **at the broadcast layer**. This is intentional: the broadcast layer's only job is "did the node accept this payload", and re-broadcasting an already-accepted signature is useless.
- **The authority is the downstream confirmation poll.** Actual success/failure is decided by `getTxStatus` (the post-broadcast confirmation poll), which surfaces `failed` for a reverted Solana tx via the status resolver reading `signatureStatus.err` (`packages/core/chain/tx/status/resolvers/solana.ts`). The CLI agent always runs this poll in headless mode, so the optimistic broadcast result is corrected before the user/agent is told the tx succeeded.
- **Who is exposed.** Only a raw-SDK consumer that broadcasts and **skips** the confirmation poll sees the optimistic result. The CLI agent is **not** exposed.

The doc-comment on the `AlreadyProcessed→return` branch states all of the above inline so the trade-off is not re-litigated.

## Tests (red-then-green, ship in this PR)

**`packages/core/chain/tx/broadcast/resolvers/solana.test.ts`** (broadcast layer):
- duplicate-signature rejection (`"This transaction has already been processed"`) → resolves without calling `verifyBroadcastByHash`
- `AlreadyProcessed` transaction error → same
- pins the `AlreadyProcessed→success` decision explicitly — idempotent success, single broadcast attempt (no re-broadcast), no `verifyBroadcastByHash` — so a future re-route is caught
- the existing `"already processed"` (no "been") generic-rejection case still routes to `verifyBroadcastByHash` — no collision

**`packages/core/chain/tx/status/resolvers/solana.test.ts`** (authority step):
- `signatureStatus.err` present → returns `status: 'error'` (`isKnown: true`) **without** a tx fetch — proves the confirmation poll detects a reverted Solana tx

**`clients/cli/src/agent/__tests__/sessionTxConfirm.test.ts`** (cross-layer):
- a Solana tx whose broadcast optimistically succeeded + `getTxStatus` returning `'error'` → the confirm poll emits `failed` (not `confirmed`), proving the authority step overrides the optimistic broadcast

## Validation

- `typecheck` ✅
- core broadcast + status Solana suites: 9/9 ✅ (broadcast 6/6, status 3/3)
- full CLI suite: 345/345 ✅
- prettier ✅
- changeset: `@vultisig/core-chain` patch ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Solana transactions now treat “already processed” broadcast responses as successful instead of retrying or resubmitting.
  * Failed Solana confirmations are now reported correctly in the UI, including cases where a transaction is later finalized with an execution error.
  * Solana status checks better recognize finalized on-chain errors without extra lookups.
  * Quote handling was updated to avoid interfering with Kyber affiliate fee calculations when another quote source is unavailable.


gomesalexandre (approved):
correct fix. broadcast layer's only job is 'did the node accept this payload' and `AlreadyProcessed` is an accepted payload. the trade-off is well-documented inline and the real authority (getTxStatus confirmation poll) still correctly surfaces failed for reverted txs. mirrors TON/UTXO/Cosmos guards that were already in place. 6/6 tests green.

PR #875, opened 2026-06-25
fix(cli): surface agent loop-depth overrun as a typed truncation error

neavra:
## Summary

The agent message loop has a belt-and-suspenders depth cap (`MAX_MESSAGE_LOOP_DEPTH = 16`). When `processMessageLoop` exceeded it, the code cleared the queued tool results (`pendingToolResults = []`) and called **`ui.onDone()`** — the *same* success-shaped callback used on a clean finish. A headless caller (`agent ask --json` / `--via-agent` pipe) therefore could not distinguish a depth-capped, mid-flight **truncation** from a completed turn; only the stderr warning hinted at it, which a machine consumer can't act on.

## Fix

On the cap path, emit a **distinct typed error** `AgentErrorCode.LOOP_DEPTH_EXCEEDED` via `ui.onError(...)` **before** `ui.onDone()`:

- **ask**: `AskInterface` records the first `onError` into `result.error`, so `agent ask` exits non-zero and writes the structured error envelope (with the `LOOP_DEPTH_EXCEEDED` code) to stdout.
- **pipe**: `PipeInterface` emits a typed `{ type: 'error', code: 'LOOP_DEPTH_EXCEEDED' }` NDJSON frame.
- `onDone()` still fires after, as the turn terminator (pipe consumers read until `done`) — mirroring the established **SSE-error-frame → onDone** pattern. The error code, not `onDone`, is the success/failure signal.

The cap value (16) and the `pendingToolResults` clear are unchanged — only the success-shaped exit was the bug.

## Scope

This is the freshness-narrowed scope of backlog **fix-10** (`230626-sdkcli-fix-10-loop-depth-health.md`): the loop-depth truncation signal only. The health-check retry/timeout half of the original ticket was dropped during re-verification and is **not** included here.

## Test

Adds `sessionLoopDepth.test.ts` — a red-then-green lock that drives `processMessageLoop` past the cap and asserts:
- `ui.onError` is called once with `AgentErrorCode.LOOP_DEPTH_EXCEEDED` (fails on the old code, which called only `onDone`),
- `pendingToolResults` is cleared (no leak into the next turn),
- the backend stream is never hit (guard fires at the top of the loop).

Full CLI suite (345 tests) green, CLI typecheck clean, prettier clean. Changeset: `@vultisig/cli` patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent handling when the message loop exceeds the depth limit: truncation is now emitted as a clear typed error before the run finishes, rather than ending in a success-shaped response.
  * Added a terminal `LOOP_DEPTH_EXCEEDED` error code, with updated precedence/latching so terminal errors override earlier non-terminal ones.
  * Ensured queued tool results are cleared when the depth guard triggers.
* **Tests**
  * Added and expanded unit tests covering loop-depth truncation behavior, error framing, and error precedence across streamed frames.


gomesalexandre (approved):
good fix - emitting LOOP_DEPTH_EXCEEDED before onDone is the right distinction. a headless consumer can't act on a bare done to detect truncation. the error-then-done shape matching pipe.ts handleCommand precedent is correct. test confirms onError fires exactly once before onDone.

PR #878, opened 2026-06-26
fix(client-shared): harden config store (warn on corruption, 0o600 perms)

neavra:
Hardens the vault-registry config store (`packages/client-shared/src/config-store.ts`):

- **`loadConfig`** now **warns to stderr** (naming the path + parse error) when `config.json` is corrupted instead of silently reverting to an empty registry — which would factory-reset the user's vault list on a single-byte corruption or partial write. The bad file is **left intact at load time** (the next `saveConfig` still overwrites it once the user mutates state — this is not durable recovery, just a chance to inspect the file before the next write). A missing file is still the normal first-run case (silent).
- **`saveConfig`** now writes `config.json` with **`0o600`** perms (and `chmod`s on every write, since `mode` is honored only on create) and creates the config dir with **`0o700`** — mirroring credential-store's hardening (`7bafd71a`). Previously it used the umask default (typically `0o644`, world-readable) though it lives in `~/.vultisig` alongside credentials.

Red-then-green tests added in `clients/cli/tests/config-store.test.ts`. Changeset included (`@vultisig/client-shared` patch).

Validation: typecheck (client-shared + cli) ✓ · config-store suite 12/12 ✓ · prettier (repo config) ✓

fixes audit fix-16 (narrowed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved config loading to handle missing, unreadable, and corrupted files more clearly.
  * Missing config files now load quietly on first run, while corrupted or unreadable files show a warning and keep the existing file intact.
  * Strengthened config file and directory permissions to reduce unintended access.
  * Ensured saved config files use stricter permissions and remain correctly protected after updates.


gomesalexandre (approved):
solid hardening. 0o700/0o600 file perms mirror credential-store, warn-on-corrupt instead of silent factory-reset is the right call. the chmod-on-every-write pattern to handle pre-existing files is correct. 12/12 tests green after building client-shared.

PR #879, opened 2026-06-26
fix(cli): make init-time SESSION_NOT_FOUND lowest-priority vs real turn error

neavra:
## Context

Follow-up to #869 (already merged). Addresses a **CodeRabbit Major** finding on that PR that arrived too late to fix before it merged.

## The finding (valid)

The `SESSION_NOT_FOUND`-preservation fix in #869 pre-set \`this.error\` with the initialize-time \`SESSION_NOT_FOUND\` (from a stale \`--session\` fallback) and kept it through the first turn. But \`onError\` is first-error-wins (\`if (!this.error)\`), so a **real** first-turn backend/stream error was ignored — the envelope reported \`SESSION_NOT_FOUND\` instead of the actual error.

## The fix

The init-time signal is now the **lowest-priority** signal:

- Never pre-set \`this.error\` with the init-time \`SESSION_NOT_FOUND\`; keep it in a separate \`initError\` field (set when \`onError\` fires before the first \`ask()\`).
- \`partialResult()\` returns \`this.error ?? this.initError\` — a real turn error wins; fall back to the init signal only when the turn produced no error.
- \`initError\` carries only into the **first** turn (cleared once a prior turn has run); each turn's \`error\` is turn-local.

This preserves #869's intent: a stale-session fallback followed by a clean turn still reports \`SESSION_NOT_FOUND\` with a non-zero exit.

## Tests

- **`ask.test.ts`** (unit, `AskInterface`): real first-turn error overrides init signal; clean first turn still reports `SESSION_NOT_FOUND`; init signal does not carry into a later turn.
- **`agentAskJson.test.ts`** (output-contract): stale-session fallback + real first-turn error → envelope reports the **real** error code, exit non-zero (new case); the existing clean-turn case still reports `SESSION_NOT_FOUND`.

## Validation

- `yarn workspace @vultisig/cli typecheck` ✓
- Full CLI suite serial (`yarn vitest run clients/cli/src --no-file-parallelism`): 372 passed ✓
- Prettier (repo config) ✓
- Changeset included (`@vultisig/cli` patch)


gomesalexandre (approved):
clean fix for the CodeRabbit finding from #869. the `initError` vs turn `error` split is the right approach - `?? initError` fallback in partialResult is exactly what you want here, real turn errors win, init signal only bleeds into the first turn. tests cover both paths. 

ser 🤖 reviewed via Claude Code (local /loop)

PR #880, opened 2026-06-26
fix(lib): default timeout on queryUrl (unbounded fetch perma-loading the send fee row)

gomesalexandre:
Root-cause fix for the flag-OFF "Network fee row perma-loads → build timed out" bug (companion to mcp-ts #589).

## root cause (sim-reproduced, prod, flag OFF)
A USD send (`$0.10 eth`) → `fiatToAmount` → `getCoinPrices` → `queryCoingeickoPrices` → **`queryUrl`**, which did a bare `fetch()` with **no timeout**. A stalled `/coingeicko` price proxy → fetch never settles → `fiatToAmount` never returns → mcp-ts `execute_send` never emits → the send card spins to the app 60s build-timeout. **Native amounts skip the price fetch** (why `0.0001 eth` resolved, `$0.10 eth` hung).

`queryUrl` is the shared helper behind prices/balances/chain queries — the unbounded-fetch **class** bug.

## fix
20s default deadline when no caller `signal` (Hermes-compatible `AbortController`+`setTimeout`; `AbortSignal.timeout` isn't on older Hermes — see `sdk tools/gas/utxoFeeRate.ts`). Market data still fetched, just bounded. New `timeoutMs` option; signal-passing callers unchanged.

## receipts
3 new tests (hung → aborts 20s w/ timeout error; in-time → resolves; caller signal passes through). lib suite green, tsc clean.

### DIFF
--- a/packages/lib/utils/query/queryUrl.ts
+++ b/packages/lib/utils/query/queryUrl.ts
@@ -6,7 +6,20 @@ type ResponseType = 'json' | 'text' | 'none'
 type QueryUrlOptions = {
   responseType?: ResponseType
   body?: any
+  /**
+   * Deadline in ms for the DEFAULT-timeout path (only applied when the caller
+   * does not pass its own `signal`). Defaults to DEFAULT_QUERY_TIMEOUT_MS.
+   */
+  timeoutMs?: number
 } & Pick<RequestInit, 'method' | 'headers' | 'signal'>
+
+// Default deadline for any queryUrl call that doesn't bring its own AbortSignal.
+// An unbounded fetch against a hung/slow upstream wedges the caller forever: the
+// `/coingeicko` price proxy stalling is what made fiatToAmount -> execute_send
+// hang and perma-loaded the agent send card's "Network fee" row until the app's
+// own 60s build-timeout fired. 20s matches the SDK's other HTTP timeouts
+// (balance/swap rails) — comfortably above a healthy query, still bounding a hang.
+const DEFAULT_QUERY_TIMEOUT_MS = 20_000
 
 const processBody = (body: any) => {
   if (body === undefined) {
@@ -29,24 +42,44 @@ export function queryUrl<T extends string = string>(
 export function queryUrl<T>(url: string | URL, options?: QueryUrlOptions & { responseType?: 'json' }): Promise<T>
 
 export async function queryUrl<T>(url: string | URL, options: QueryUrlOptions = {}): Promise<T | string | void> {
-  const { responseType = 'json', body, headers, method, signal } = options
+  const { responseType = 'json', body, headers, method, signal, timeoutMs = DEFAULT_QUERY_TIMEOUT_MS } = options
 
-  const response = await fetch(
-    url,
-    withoutUndefinedFields({
-      method: method ?? (body ? 'POST' : 'GET'),
-      headers: withoutUndefinedFields({
-        ...headers,
-        'Content-Type': body ? 'application/json' : undefined,
-      }),
-      body: processBody(body),
-      signal,
-    })
-  )
+  // Hermes-compatible default deadline: `AbortSignal.timeout()` isn't available
+  // on older RN/Hermes runtimes (see sdk tools/gas/utxoFeeRate.ts), so when the
+  // caller doesn't supply its own signal, bound the fetch with
+  // AbortController + setTimeout. Callers that pass `signal` own cancellation
+  // and opt out of the default deadline.
+  let timeoutId: ReturnType<typeof setTimeout> | undefined
+  let effectiveSignal = signal
+  if (!effectiveSignal) {
+    const controller = new AbortController()
+    timeoutId = setTimeout(
+      () => controller.abort(new Error(`queryUrl: request timed out after ${timeoutMs}ms (${String(url)})`)),
+      timeoutMs
+    )
+    effectiveSignal = controller.signal
+  }
 
-  await assertFetchResponse(response)
+  try {
+    const response = await fetch(
+      url,
+      withoutUndefinedFields({
+        method: method ?? (body ? 'POST' : 'GET'),
+        headers: withoutUndefinedFields({
+          ...headers,
+          'Content-Type': body ? 'application/json' : undefined,
+        }),
+        body: processBody(body),
+        signal: effectiveSignal,
+      })
+    )
 
-  if (responseType !== 'none') {
-    return response[responseType]()
+    await assertFetchResponse(response)
+
+    if (responseType !== 'none') {
+      return response[responseType]()
+    }
+  } finally {
+    if (timeoutId) clearTimeout(timeoutId)
   }
 }
--- a/.changeset/queryurl-default-timeout.md
+++ b/.changeset/queryurl-default-timeout.md
@@ -1,0 +1,14 @@
+---
+'@vultisig/lib-utils': patch
+'@vultisig/sdk': patch
+---
+
+Add a 20s default deadline to `queryUrl` (the shared HTTP helper behind
+prices/balances/swap quotes/broadcast/MPC-server calls). An unbounded `fetch`
+against a hung upstream previously wedged the caller forever — a stalled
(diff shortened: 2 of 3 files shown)


packages/lib/utils/query/queryUrl.ts
```diff
@@ -29,24 +42,44 @@ export function queryUrl<T extends string = string>(
...
+  let timeoutId: ReturnType<typeof setTimeout> | undefined
+  let effectiveSignal = signal
+  if (!effectiveSignal) {
+    const controller = new AbortController()
+    timeoutId = setTimeout(
+      () => controller.abort(new Error(`queryUrl: request timed out after ${timeoutMs}ms (${String(url)})`)),
+      timeoutMs
+    )
+    effectiveSignal = controller.signal
+  }
```
NeOMakinG, line 61: `praise:` clean mechanism — `AbortController`+`setTimeout` (Hermes-safe, vs `AbortSignal.timeout`), the deadline applies *only* when the caller brings no `signal` (so signal-owning callers opt out), and `clearTimeout` lives in the `finally` so the timer never leaks whether the fetch resolves or rejects. Verified the abort actually feeds `fetch(signal)` so the socket is cancelled, not just the promise.

packages/lib/utils/query/queryUrl.ts
```diff
@@ -29,24 +42,44 @@ export function queryUrl<T extends string = string>(
...
+  // on older RN/Hermes runtimes (see sdk tools/gas/utxoFeeRate.ts), so when the
+  // caller doesn't supply its own signal, bound the fetch with
+  // AbortController + setTimeout. Callers that pass `signal` own cancellation
+  // and opt out of the default deadline.
+  let timeoutId: ReturnType<typeof setTimeout> | undefined
+  let effectiveSignal = signal
+  if (!effectiveSignal) {
+    const controller = new AbortController()
+    timeoutId = setTimeout(
+      () => controller.abort(new Error(`queryUrl: request timed out after ${timeoutMs}ms (${String(url)})`)),
```
NeOMakinG, line 57: `q:` the abort reason is a plain `Error`, not a `DOMException` with `name === 'AbortError'`. I traced the fund-relevant callers and none branch on `err.name === 'AbortError'` (broadcast resolvers route any throw through `verifyBroadcastByHash`/`isInError` string-matching, balance/price just propagate), so this is fine today — flagging only so a future caller that wants to distinguish abort-vs-HTTP doesn't get surprised. No change requested.

NeOMakinG (approved):
Fund-safety review of the `queryUrl` default-timeout fix. `queryUrl` is the SDK's universal HTTP helper (balances, prices, swap quotes, MPC keygen/keysign server calls, and **tx broadcast**), so the load-bearing question isn't the timeout itself — it's what every fund-relevant caller does when it now throws at 20s. Walked all of them. **Pushed 2 commits to your branch to get CI green** (details below); the change itself is sound.

## testing approach
- 3 lanes: Claude (load-bearing) + Codex `gpt-5.4` (adversarial). GLM lane dead (connector auth precedence) — best-effort, skipped.
- Traced every broadcast resolver + the MPC FastVault kickoff calls + balance/price callers against a fresh worktree at HEAD `937b8b63`.
- `vitest run queryUrl.test.ts` → **3/3 pass** (hung→aborts@20s, in-time→resolves, caller-signal→passthrough). prettier + eslint clean on the changed files after my fix. SDK release-guard now passes.
- Codex load-bearing conclusion: *"broadcast paths do not obviously auto-retry the send itself"* — matches my trace; its one watch-item (MPC endpoints inheriting 20s) I resolved below.

## mechanism (sane?)
20s default via `AbortController`+`setTimeout`, `clearTimeout` in `finally` (no timer leak), applies **only** when the caller passes no `signal`, overridable via the new `timeoutMs` option. The abort signal feeds `fetch` so the socket is actually cancelled. 20s is comfortably above a healthy single round-trip and well under the app's 60s build-timeout. Good.

## broadcast double-send (the critical path) — clear
- `broadcastTx` is a plain dispatcher: **no retry wrapper**. Resolvers don't self-retry. So a 20s timeout-throw never triggers an SDK-level re-submit → no double-send introduced.
- Idempotency net `verifyBroadcastByHash` re-checks the tx on-chain by hash before swallowing, and **rethrows** the original error if it can't confirm (fail-closed, never a new success).
- `tron`/`ton`/`polkadot` route a thrown timeout *through* `verifyBroadcastByHash` (try/catch or `attempt`).
- `cardano` has its **own** 30s `AbortController` and bypasses `queryUrl` entirely — unaffected.
- `ripple` uses the XRPL client (not `queryUrl`) — unaffected.

## should-fix
- **`utxo` + `bittensor` broadcast resolvers** `await queryUrl(...)` directly, so a *thrown* error (the new 20s abort, or any HTTP/network throw) propagates **without** going through `verifyBroadcastByHash` — unlike `tron`/`ton`/`polkadot`. This is pre-existing (any fetch throw already escaped the verify net there), but the new default makes the throw path more reachable: a blockchair/RPC push that actually lands but whose response we abort at 20s now surfaces a raw error, and the only double-send vector left in the system is a user/consumer manual retry on that false-negative. Wrapping both in the same `try { … } catch (error) { await verifyBroadcastByHash({ chain, tx, error }) }` shape as `tron.ts` closes the gap and makes all five direct-`queryUrl` resolvers consistent. `packages/core/chain/tx/broadcast/resolvers/utxo.ts:32`, `packages/core/chain/tx/broadcast/resolvers/bittensor.ts:17`.

## fail-closed (no fabricated-zero / underprice) — confirmed
Balance/price callers don't swallow a thrown timeout into a default: `queryCoingeickoPrices` simply `await`s and propagates (that's the whole point of the fix — `fiatToAmount` now fails fast instead of hanging), and the cosmos balance resolver returns `null` on missing data, never `0`. A timeout throws up the stack → operation aborts cleanly → no zeroed balance or missing-gas feeding a tx.

## MPC keygen/keysign (Codex's watch-item) — safe
`signWithServer` (`/sign`), `createVaultWithServer` (`/create`), `setupVaultWithServer` (`/batch/keygen`) all use `responseType: 'none'` and are **fast kickoffs** — `ServerManager.ts:217` awaits `signWithServer` then separately *"Step 3: Wait for server to join"* via the relay. The actual ceremony is relay **short-polls** (`getMpcRelayMessages` in `sleep(100)` loops, `waitForSetupMessage` = `retry(attempts:50, delay:200)`), each an immediate-return GET well under 20s. None of these block-until-complete, and all were **unbounded before this PR** — the 20s ceiling on a fast ack is pure improvement.

## consumer bump — handled
`queryUrl.ts` lives under `packages/lib/utils/` which is bundled into `@vultisig/sdk`, so the release guard (correctly) requires a changeset. **This was the CI failure.** I pushed:
1. `55636d3b` — `.changeset/queryurl-default-timeout.md` bumping `@vultisig/lib-utils` + `@vultisig/sdk` (patch, house style per #764/#870). mcp-ts / vultiagent-app pick this up via the next SDK tarball bump.
2. `937b8b63` — `simple-import-sort` + prettier autofix on the new test file (the static gate short-circuits at the release-guard step *before* lint, so these were latent and would have failed once the guard passed).

## risk
Low. Single-file additive HTTP-helper change behind a well-tested idempotent broadcast layer. Residual: the `utxo`/`bittensor` verify gap (should-fix above) and the general property that 20s is a fixed ceiling — fine for every current caller (all single round-trips / short-polls), but any future `queryUrl` consumer that is legitimately long-lived must pass `timeoutMs` or its own `signal`.

## verdict
**APPROVE.** Timeout is sane and cancels cleanly; no double-broadcast (no retry wrapper, idempotent `verifyBroadcastByHash` net); no fabricated-zero/underprice (callers propagate, don't swallow); override available; MPC kickoffs are fast acks. CI blockers fixed in-code on your branch — Lint+Types, Integration, CodeRabbit green at HEAD `937b8b63`. The one `should-fix` (utxo/bittensor verify-on-throw consistency) is non-blocking and pre-existing.

---
*Reviewed by Claude (load-bearing) + Codex gpt-5.4. Two commits pushed to the branch: changeset + lint/format. 🤖*

PR #810, opened 2026-06-23
feat(sdk): add sdk.amount.convert (base↔human, fiat↔crypto)

gomesalexandre:
Part of the mcp-ts/backend → SDK code-as-action consolidation.

## what + why

Adds a pure unit-/decimals- and fiat-conversion primitive to the SDK: `convertAmount` (base-unit ↔ human-readable) plus `fiatToCrypto` / `cryptoToFiat` (fiat ↔ crypto with the price supplied as an input). This folds three previously-duplicated impls into one canonical surface:

- mcp-ts `tools/utility/convert-amount.ts` (`toBaseUnits`/`toHumanUnits`, pure string math, 0 SDK imports)
- Go `validator/{amount,decimals}.go`'s scale kernel (`scaleRawToHuman` = raw ÷ 10^decimals, `bigPow10`) — the math, minus the agent-layer prose regexes which stay in orchestration
- the existing SDK `fiatToAmount` (live-price variant; this PR is its synchronous, price-as-input sibling)

The base↔human path uses **string arithmetic, not floats** — zero precision loss on large base-unit integers (a `Number`-based scale silently corrupts amounts above 2^53 wei). Fractional digits are truncated (not rounded) to match on-chain behaviour. Everything is pure: no network, no vault, no LLM/agent concepts.

Boundary rules baked in (mirroring the Go validator + mcp-ts zod schema): decimals must be an integer in `[0, 255]`; scientific notation / junk input is rejected; empty / non-integer base-unit strings throw `AmountConvertError` (LLM-readable message); a leading sign is supported.

New exports from the package index: `convertAmount`, `toBaseUnits`, `toHumanUnits`, `fiatToCrypto`, `cryptoToFiat`, `AmountConvertError` (+ types).

## consumers it unblocks

- **isolate / co-sign gate** — one precision-exact base↔human kernel for amount drift-checks at the signing boundary (no more per-surface float math)
- **CLI** — text-WYSIWYS amount rendering
- **app** — card amount formatting
- **mcp-ts** — lets `convert_amount` become a thin wrapper over the SDK (drops its local copy), same path the other already-in-SDK tools took

## receipt

\`\`\`
$ cd packages/sdk && node --import tsx scripts/receipts/amount_convert.mjs

=== sdk.amount.convert — runnable receipt ===

-- base ↔ human (pure string math, precision-exact) --
1.5 ETH  → wei (decimals=18)                   1500000000000000000
100 USDC base-units → human (decimals=6)       100
[...]

PR #884, opened 2026-06-26
fix(core-chain): lower terraclassic lunc cosmosGasRecord from 100 to 20 lunc

gomesalexandre:
## What

Lowers `cosmosGasRecord[Chain.TerraClassic]` from 100_000_000n (100 LUNC) to 20_000_000n (20 LUNC) in `packages/core/chain/chains/cosmos/gas.ts`.

This is the root cause fix. `getCosmosFeeAmount` uses this as the floor: since the live computed fee (~11 LUNC) is below the floor, it always returns the floor. Lowering the floor from 100 to 20 LUNC means the signer now charges 20 LUNC for a bare MsgSend on columbus-5.

## Why

Real on-chain MsgSend cost: ~400k gas x 28.325 uluna/gas ~ 11.33 LUNC. The 100 LUNC floor was blocking sends from wallets with 20-100 LUNC balance even when the tx would have succeeded on-chain:

- User has 50 LUNC, tries to send 10 LUNC
- Old: agent checks 10 + 100 = 110 > 50 -> blocked
- New: agent checks 10 + 20 = 30 < 50 -> passes; signer charges 20 LUNC; user sends 10 LUNC

Note: `cosmosGasRecord` is the FLOOR for sends. For staking, `mcp-ts` uses a separate `COSMOS_STAKING_FEE_PER_MSG_BASE_UNITS.TerraClassic = 100_000_000n` (intentionally kept at 100 LUNC because staking ops use ~1M gas/msg, not ~400k).

## Math

- MsgSend gas limit: 400_000 gas
- columbus-5 min-gas-price: ~28.325 uluna/gas
- Computed fee: ~11.33 LUNC = ~11_330_000 uluna
- New floor: 20 LUNC = 20_000_000 uluna (~1.77x real cost)

## Risk

Low - scoped to TerraClassic sends. Includes changeset for patch bump. All 2265 SDK tests green.

## Companion PRs

- vultisig/agent-backend#1409 - lowers the AB reserve to 20 LUNC
- vultisig/mcp-ts#594 - lowers COSMOS_SEND_FEE_BASE_UNITS['TerraClassic'] to 20 LUNC

🤖 generated with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Lowered TerraClassic’s minimum send fee estimate, allowing wallet sends to work with smaller LUNC balances.
  * Improved fee calculations and related guidance so TerraClassic transactions show a more accurate expected cost.
* **Tests**
  * Updated fee-label expectations to match the new TerraClassic send estimate.

### DIFF
--- a/packages/core/chain/chains/cosmos/gas.ts
+++ b/packages/core/chain/chains/cosmos/gas.ts
@@ -12,7 +12,7 @@ export const cosmosGasRecord: Record<IbcEnabledCosmosChain, bigint> = {
   [Chain.Kujira]: defaultGas,
   [Chain.Terra]: defaultGas,
   [Chain.Dydx]: 2500000000000000n,
-  [Chain.TerraClassic]: 100000000n,
+  [Chain.TerraClassic]: 20000000n,
   [Chain.Noble]: 30000n,
   [Chain.Akash]: 200000n,
 }
--- a/.changeset/fix-terraclassic-lunc-send-fee.md
+++ b/.changeset/fix-terraclassic-lunc-send-fee.md
@@ -1,0 +1,12 @@
+---
+"@vultisig/core-chain": patch
+"@vultisig/sdk": patch
+---
+
+Lower `cosmosGasRecord[TerraClassic]` from 100 LUNC to 20 LUNC.
+
+Real on-chain MsgSend cost on columbus-5: ~400k gas x 28.325 uluna/gas ~ 11.33 LUNC.
+The 100 LUNC floor was blocking sends from wallets with 20-100 LUNC balance even when
+the transaction would have succeeded. The new 20 LUNC floor gives a ~1.77x buffer.
+
+Companion to vultisig/agent-backend#1409 and vultisig/mcp-ts#594.
--- a/packages/core/chain/chains/cosmos/cosmosGasLimitRecord.ts
+++ b/packages/core/chain/chains/cosmos/cosmosGasLimitRecord.ts
@@ -41,12 +41,15 @@ export const getCosmosGasLimit = (coin: CoinKey<CosmosChain>): bigint => {
  * writes that the standard SDK gas estimate omits. In practice this adds
  * ~200-800 gas on top of the base delegate cost (~2M), so a 2M limit fails
  * consistently ("out of gas in location: ValuePerByte; gasWanted: 2000000,
- * gasUsed: 200X"). 3M sits safely under the 100 LUNC fee floor at msgCount=1
- * (3M * 28.325 uluna/gas ≈ 84.97 LUNC < 100 LUNC). The msgCount scaling is
- * disabled for TerraClassic: at msgCount>=2, scaled gasWanted would exceed
- * 100 LUNC, causing node rejection when the tx fee amount (cosmosGasRecord
- * [TerraClassic]=100_000_000 uluna) no longer covers gasWanted * gasPrice.
- * Columbus-5 callers must split multi-validator claims into separate txs.
+ * gasUsed: 200X"). 3M needs ~84.97 LUNC at sign time (3M * 28.325 uluna/gas).
+ * That fee is supplied by the CONSUMER's separate per-msg staking fee
+ * (mcp-ts COSMOS_STAKING_FEE_PER_MSG_BASE_UNITS[TerraClassic] = 100 LUNC/msg,
+ * written into the signAmino / signDirect fee) — NOT by
+ * `cosmosGasRecord[TerraClassic]`, which is the native MsgSend fee floor
+ * (20 LUNC) and never governs staking. The msgCount scaling is disabled for
+ * TerraClassic: at msgCount>=2, scaled gasWanted would push the required fee
+ * above the 100 LUNC per-msg staking fee, causing node rejection. Columbus-5
+ * callers must split multi-validator claims into separate txs.
  */
 const cosmosStakingGasLimitRecord: Record<IbcEnabledCosmosChain, bigint> = {
   [Chain.Cosmos]: 350_000n,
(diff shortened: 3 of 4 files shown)


packages/core/chain/chains/cosmos/gas.ts
```diff
@@ -12,7 +12,7 @@ export const cosmosGasRecord: Record<IbcEnabledCosmosChain, bigint> = {
   [Chain.Kujira]: defaultGas,
   [Chain.Terra]: defaultGas,
   [Chain.Dydx]: 2500000000000000n,
-  [Chain.TerraClassic]: 100000000n,
+  [Chain.TerraClassic]: 20000000n,
```
NeOMakinG, line 15: `should-fix:` this floor (now 20 LUNC) is also the fallback fee for **uusd/USTC** sends, whose gas limit is hardcoded to `1_000_000n` (`getCosmosGasLimit`). A uusd send needs `1M × 28.325 ≈ 28.3 LUNC`, so on the LCD timeout/failure path (`getCosmosFeeAmount` catch → floor), the `maxLiveFeeMultiplier` outlier guard, or `buildCosmosPayload` `skipChainSpecificFetch` mode, uusd now signs 20 LUNC against a 1M gasWanted → ante-handler `insufficient fee` rejection (no fund loss; #810 confusing-failure symptom). The old 100 floor masked this. Native LUNC is fine. Suggest a per-coin floor for uusd (≥ ~30 LUNC) in a follow-up rather than expanding this PR.

NeOMakinG (approved):
**Verdict: APPROVE.** This is the load-bearing root of the 3-PR LUNC-fee chain (companions agent-backend#1409, mcp-ts#594) and the change is correct for its target: native columbus-5 `MsgSend` now signs a 20 LUNC fee (real cost ~11.33 LUNC, **1.77x margin**), and the floor stays above the real cost even on the LCD-fetch-failure fallback path. Critically, **staking is provably NOT under-funded** by this change. Approving; the two `should-fix` items below are narrow degraded-path / follow-up concerns, not blockers for this PR.

## testing approach
Traced the full signed-fee provenance in the SDK against `origin/main`; did not re-trust the PR's premise. CI is fully green (Unit 16m24s, Integration 4m53s, Lint/TypeCheck, CodeRabbit). 3 lanes: Claude load-bearing source trace; Codex gpt-5.3-codex **converged independently** on the USTC edge; GLM dead (`ANTHROPIC_API_KEY` precedence — best-effort skip).

**SEND (the target) — safe with margin.** `getCosmosFeeAmount` (`gas.ts:131`) returns `computedFee > floor ? computedFee : floor`. For native LUNC, `getCosmosGasLimit = 400_000n` → computed `400k × 28.325 ≈ 11.33 LUNC < 20` floor → signs **20 LUNC** (1.77x). Even on LCD timeout/failure (`catch → floor`) the 20 floor still exceeds the 11.33 the node requires for 400k gas. Correct.

**STAKING — NOT under-funded (the #1 risk, cleared).** Verified by source, not by the PR claim:
- The native SDK cosmos signing path (`signingInputs/resolvers/cosmos/index.ts:428`) hardcodes `gas: getCosmosGasLimit(coin)` = **400k** for gasWanted — it is send/IBC/wasm-only and never emits a delegate/withdraw msg.
- `getCosmosStakingGasLimit` / `cosmosStakingGasLimitRecord` (3M for TerraClassic) is **export-only** — no SDK-internal code pairs it with `cosmosGasRecord`. `cosmosMsgTypes.ts` has zero fee logic.
- Staking is built by the **consumer** via `signAmino`/`signDirect`, whose fee is read directly from `signAmino.fee` (`index.ts:382-395`), i.e. mcp-ts `COSMOS_STAKING_FEE_PER_MSG_BASE_UNITS[TerraClassic] = 100 LUNC/msg` (unchanged, confirmed in mcp-ts#594). The 3M staking gasWanted needs ~84.97 LUNC; the consumer's 100 LUNC/msg covers it, **independent of this 20 LUNC send floor**.
- I fixed the stale `cosmosGasLimitRecord.ts` comment in-code (48189fe7) — it still claimed 3M sits "under the 100 LUNC fee floor (cosmosGasRecord=100_000_000)", which this PR makes false; the staking fee is consumer-supplied, not `cosmosGasRecord`.

**SWAP/IBC — safe.** TerraClassic is not a native-swap chain; `estimateCosmosSwapFeeLabel` is display-only (now `~20 LUNC`). Skip swap source-leg (`COSMOS_SWAP_GAS_LIMIT = 350k`) and IBC_TRANSFER (400k) both compute < 20 LUNC. Scoped to TerraClassic only — no other cosmos floor touched. Changeset present (`@vultisig/core-chain` + `@vultisig/sdk` patch).

## should-fix
**1. USTC (`uusd`) sends are under-funded on the degraded fee path.** `getCosmosGasLimit({chain: TerraClassic, id: 'uusd'})` is hardcoded to `1_000_000n` (burn-tax post-handler), needing `1M × 28.325 ≈ 28.3 LUNC` (pre burn-tax). The new 20 LUNC floor is below that, and three paths return the flat floor for uusd:
- `getCosmosFeeAmount` on LCD timeout / LCD failure (`gas.ts:151 catch → floor`),
- the `maxLiveFeeMultiplier = 10n` outlier guard (`gas.ts:147`), which now floors any uusd live fee when `gasPrice > 200 uluna/gas` (window shrank from 1000→200 LUNC absolute; current 28.325 is comfortably inside, so this is a spike-only concern — narrower than Codex's "0.2 uluna/gas" framing, which is off by 1000x),
- `buildCosmosPayload.ts:95` `skipChainSpecificFetch` mode (flat `cosmosGasRecord[chain]`).

In all three, uusd would sign 20 LUNC against a 1M gasWanted → ante-handler `insufficient fee` rejection (tx bounces — **no fund loss**, but the #810 confusing-failure symptom for USTC). The old 100 floor masked this. Not blocking: it is not the PR's LUNC target, USTC is a degraded path, and a proper fix is a per-coin floor (uusd ≥ ~30 LUNC) which needs its own design + burn-tax consideration — recommend a follow-up rather than expanding this PR.

## suggestion
Add a TerraClassic `uusd` case to the `getCosmosFeeAmount` tests (`gas.test.ts`) covering both live-config and timeout/fallback, so the under-fee above is pinned. The updated test only asserts the `~20 LUNC` display string.

## risk
**Low for the target path.** LUNC sends and all staking are correct and safe. This PR is correctly the root that must merge+publish FIRST — once consumers bump to it, the signed send fee genuinely becomes 20 LUNC, so mcp-ts#594 (guard) and agent-backend#1409 (reserve) can then safely follow. Lowering the SDK floor ahead of consumers is the safe direction (an unbumped consumer's 100-guard merely over-blocks a valid 20-fee send — a false-negative, not fund loss). Residual: the USTC degraded-path under-fee above.

## verdict
**APPROVE.** Stale staking comment fixed in-code (48189fe7). USTC under-fee + test gap are non-blocking follow-ups.

---
🤖 Reviewed by Claude (load-bearing) + Codex gpt-5.3-codex (converged on USTC). GLM lane skipped (auth).

PR #886, opened 2026-06-27
Add EVM batch balance helper

rcoderdev:
Closes #4208

## Summary by CodeRabbit

* **New Features**
  * Added an EVM chain balance-fetching API that returns native and ERC-20 balances for multiple coins in one request.
  * Uses Multicall3 batching when available, with automatic fallback to per-coin balance lookups.
* **Bug Fixes**
  * Improved resilience by treating failed/unsupported balance reads as `0` values instead of breaking the response.
* **Tests**
  * Added coverage for batched, missing-metadata, and multicall failure fallback scenarios.
* **Chores**
  * Updated release metadata for core chain and SDK (patch).


johnnyluo (approved):
LGTM

PR #899, opened 2026-06-29
agent: resolve vault password from keyring/env chain before prompting

neavra:
## What

`vultisig agent` (interactive, `--via-agent`, and `agent ask`) now resolves the vault password from the existing keyring/env chain **before** falling back to an interactive prompt, so a headless operator never has to put the password on argv.

## Why

`AgentSession.initialize` unlocked an encrypted vault with `config.password || ui.requestPassword()`. `config.password` comes only from the `--password` flag, and in ask mode `requestPassword()` throws *"Password required but not provided. Use --password flag."*

A FastVault is always encrypted, so `agent ask` without `--password` threw at init **even when** the password was already available in the OS keyring (`vsig auth setup`) or the `VAULT_PASSWORD` / `VAULT_PASSWORDS` env vars. The operator was forced to pass the secret on argv, where it leaks into `ps -e`, shell history, and command logs — the one agent surface that pushed a funds-controlling secret onto the command line.

The resolution chain (`getPassword`: in-memory cache → keyring → env → prompt) already existed; the session just never consulted it.

## Changes

- **`core/password-manager.ts`** — extract `resolvePasswordNonInteractive(vaultId, vaultName)` (cache → keyring → env, returns `null` instead of prompting). `getPassword` is refactored to reuse it, so its behavior is unchanged.
- **`agent/session.ts`** — `initialize` resolves the unlock password as `config.password` → `resolvePasswordNonInteractive(...)` → the mode's own UI callback (TUI prompt / via-agent protocol / ask-mode throw). This closes the headless throw for keyring/env-configured operators.
- **argv `--password` is de-emphasized** — when it's used, the session writes a one-line stderr warning pointing at the keyring/env path. The flag still works.
- **Help text + README** — the AI-agent docs now lead with "set up the keyring or `VAULT_PASSWORD` once, then run with no `--password`", and document `--password` as a discouraged fallback.

## Tests

New `agent/__tests__/password-chain.test.ts` (red-then-green) drives `initialize` and asserts:
- `VAULT_PASSWORD` env resolves with no `--password` (no argv-flag throw)
- OS keyring resolves with no `--password`
- argv `--password` still unlocks **and** emits the stderr warning
- the UI prompt is used only when the chain resolves nothing

Local: `@vultisig/cli` typecheck, lint, and the full CLI vitest suite (381 tests) pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * `vultisig agent` now resolves vault passwords more reliably by checking saved credentials, keyring, environment variables, and then prompting only if needed.
  * Encrypted vault unlocks now recover better from stale stored passwords and re-prompt when appropriate.
  * Using `--password` now shows a warning and is treated as a fallback option.

* **Documentation**
  * Updated CLI help and README content to reflect the new password resolution flow, clearer examples, and refreshed command/output tables.


gomesalexandre (approved):
Clean security fix. The root cause is clear (agent session went straight to `config.password || ui.requestPassword()`, bypassing the keyring/env chain that already existed in `getPassword`), and the fix is well-scoped: extract the non-interactive leg into `resolvePasswordNonInteractive`, call it from `unlockEncryptedVault` before the mode's UI callback, handle the stale-stored-password edge case explicitly, and de-emphasize `--password` with a stderr warning.

Ran `yarn workspace @vultisig/cli test` locally off this branch + main merge: **433/433 pass**. Deprecation warning is wired correctly and appears in the test assertions.

---

## findings

**suggestion:** PR title `agent: ...` doesn't match conventional commit format - the repo uses `fix:`, `feat:`, etc. This is a security fix (headless operators were forced to expose funds secrets on argv), so `fix:` fits. Doesn't block merge but the changeset already calls it `patch` which aligns with `fix:`.

---

**suggestion:** the stderr deprecation warning fires unconditionally, including in `--json` / `--silent` / `--via-agent` modes

```ts
process.stderr.write(
  'Warning: passing the vault password via --password exposes it to `ps` and shell history. ...'
)
```

For machine-to-machine / orchestrator consumers (`--via-agent` + `--json`), piping stderr alongside stdout can pollute a log aggregator. A quick `if (!isSilent() && !isNonInteractive())` gate (or always-stderr for security warnings - which is also fine) - just needs a comment clarifying the choice either way. Not a blocker, just name the decision.

---

**q:** `recoverFromStaleStoredPassword` calls `this.vault.unlock(password)` successfully, then returns the password back to `unlockEncryptedVault` which calls `this.executor.setPassword(password)`. So unlock + executor both get the recovered password. Is there a reason `recoverFromStaleStoredPassword` doesn't call `setPassword` itself? Keeping it out of `recoverFromStaleStoredPassword` is cleaner (single-responsibility), just making sure the call site always follows up with `setPassword` after calling this helper. It does - but worth a comment since `recoverFromStaleStoredPassword` touches the vault but not the executor.

---

**praise:** the stale-credential recovery path is well-reasoned:
- interactive: clear cache, re-prompt, cache the winner - session survives
- ask mode: clear cache, throw a message that points at the stored credential (not at `--password`) - no accidental "use --password" misdirection for a headless operator

That error message distinction is easy to miss and neavra got it right.

---

PR #900, opened 2026-06-29
feat(solana): staking foundation — models + RPC read layer

Ehsan-saradar:
**Phase 1** of Solana native staking — parity with iOS #4659 (iOS PR vultisig/vultisig-ios#4665).

Tracking sub-issue: vultisig/vultisig-windows#4234 · Epic: vultisig/vultisig-windows#4222

Pure chain-layer foundation under `@vultisig/core-chain/chains/solana/staking`. **No UI, no signing, no validator-metadata source** — those land in later phases. Phase 1 is chain-layer only (mirrors iOS #4665), so it ships as this SDK PR alone; the windows app consumes it once published.

### What's here
- **`config`** — Stake-program id; `dataSize:200` + staker `memcmp` offset 12 for the owner-scoped stake scan; rent-exempt reserve; 1-SOL min-delegation floor (`getStakeMinimumDelegation` is blocked by the Vultisig proxy, so a documented constant substitutes); `u64::MAX` deactivation sentinel.
- **`models/stakeAccount`** — `jsonParsed` parsing and pure activation-state derivation (`activating` / `active` / `deactivating` / `inactive`). u64 fields are `bigint` so the deactivation sentinel (2⁶⁴−1) is representable.
- **`models/validator`** — decoded `getVoteAccounts` row plus an empty `ValidatorMetadata` enrichment seam (populated by the Phase 2 provider) and display fallbacks (truncated vote pubkey when unnamed).
- **`rpc`** — `getVoteAccounts`, `getProgramAccounts` (stake scan), `getAccountInfo` (jsonParsed), `getEpochInfo`, `getMinimumBalanceForRentExemption`, inflation, supply — over the shared Solana client.
- **`cooldownGate`** — pure withdraw-cooldown evaluation over the parsed delegation + the live epoch.

### Tests
- `stakeAccount.test.ts` — jsonParsed parse (u64→bigint), activation-state boundaries, deactivation sentinel.
- `cooldownGate.test.ts` — available/blocked withdraw evaluation, unlock at `deactivationEpoch + 1`.

### Notes
- Caching (per the iOS service TTLs) lives in the consumer's react-query layer, not here.
- Exports map regenerated to include only these five Phase 1 modules.
- Changeset bumps `@vultisig/core-chain` + `@vultisig/sdk` (minor) so consumers (incl. windows) get a new tarball.

## Summary by CodeRabbit

* **New Features**
  * Added a Solana staking foundation for reading stake accounts, validator data, and epoch/rent/supply details.
  * Introduced activation lifecycle modeling plus a withdraw cooldown gate using on-chain delegation epochs.
  * Added new public import entrypoints for Solana staking configuration, models, cooldown gating, and RPC reads.
* **Bug Fixes**
  * Improved handling of undelegated accounts, deactivation sentinel logic, and strict validation of parsed stake fields.
* **Tests**
  * Added Vitest coverage for stake-account parsing and cooldown gate unlock behavior.

### DIFF
--- a/.changeset/solana-staking-foundation.md
+++ b/.changeset/solana-staking-foundation.md
@@ -1,0 +1,12 @@
+---
+"@vultisig/core-chain": minor
+"@vultisig/sdk": minor
+---
+
+feat(solana): staking foundation — models + RPC read layer
+
+Phase 1 of Solana native staking. Adds the chain-layer read foundation under
+`@vultisig/core-chain/chains/solana/staking`: config, stake-account/validator
+models (jsonParsed parsing + activation-state derivation), the RPC read layer
+(getVoteAccounts / stake-account scan / epoch / rent / inflation / supply), and
+the withdraw cooldown gate. No UI, no signing, no validator-metadata source.
--- a/packages/core/chain/chains/solana/staking/config.ts
+++ b/packages/core/chain/chains/solana/staking/config.ts
@@ -1,0 +1,83 @@
+/**
+ * Constants for Solana native staking (Stake program) reads.
+ *
+ * The Stake-program account layout and the staker-authority memcmp offset are
+ * protocol facts, not per-chain config, so this is a single flat constant set.
+ *
+ * The min-delegation RPC (`getStakeMinimumDelegation`) is blocked by the
+ * Vultisig proxy. The 1 SOL program minimum delegation is active on mainnet and
+ * is enforced by the Stake program — a delegation below it reverts with
+ * `StakeError.InsufficientDelegation`. Since the RPC is blocked, the delegation
+ * floor uses the documented `minDelegationFloorLamports` constant below plus the
+ * live rent-exempt reserve (active stake = funding − rent, so the funding floor
+ * is 1 SOL + rent).
+ *
+ * Port of iOS `SolanaStakingConfig`.
+ */
+export const solanaStakingConfig = {
+  /**
+   * The on-chain Stake program. Every stake account is owned by this program;
+   * it is also the `programId` argument to the stake-filtered
+   * `getProgramAccounts` scan.
+   */
+  stakeProgramId: 'Stake11111111111111111111111111111111111111',
+
+  /**
+   * Byte size of a fully-initialized stake account (`StakeStateV2`). The
+   * `getProgramAccounts` scan filters on `dataSize: 200` to exclude
+   * uninitialized / rewards-pool accounts before the memcmp narrows to the
+   * owner's accounts.
+   */
+  stakeStateSize: 200,
+
+  /**
+   * Offset of the staker authority pubkey inside the stake-account data, used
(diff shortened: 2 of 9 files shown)


gomesalexandre (approved):
Solid Phase 1 foundation ser. No signing, no UI, pure read layer - does exactly what it says. Tests are thorough and the defensive parsing (`undefined` on malformed payloads instead of coerced zeros) is the right call for a layer that eventually gates fund-moving Txs.

Ran `yarn test:core` locally off this branch + main merge: **1179/1179 pass**. Build clean.

---

## findings

**suggestion:** `toBigInt` with a large-number input (>2^53) silently produces imprecise bigints

```ts
return BigInt(typeof value === 'number' ? Math.trunc(value) : value)
```

If a numeric (not string) `stake` or `lamports` field ever arrives larger than `Number.MAX_SAFE_INTEGER`, `Math.trunc` is already lossy and `BigInt(...)` wraps the imprecise float. No exception, wrong value. For Phase 1 display-only this is low risk (RPC `jsonParsed` sends these as strings in practice), but it could bite a future signing layer that derives withdraw amounts from this. Consider: `if (typeof value === 'number' && !Number.isSafeInteger(value)) return undefined` before the cast.

---

**suggestion:** `fetchSolanaStakeAccounts` passes `owner` as raw string into memcmp `bytes` filter without upfront pubkey validation

```ts
{ memcmp: { offset: solanaStakingConfig.stakerMemcmpOffset, bytes: owner } }
```

A malformed `owner` (wrong length, invalid base58) won't throw here — the RPC either silently returns empty results or surfaces a vague JSON-RPC error. The downstream `fetchSolanaStakeAccount` validates via `new PublicKey(pubkey)`, so there's already a precedent. A `new PublicKey(owner)` guard at the top of `fetchSolanaStakeAccounts` would surface the bad input with a clear message and keep error paths consistent.

---

**suggestion:** `voterMemcmpOffset: 124` has no byte-layout derivation comment

`stakerMemcmpOffset: 12` has a clear breakdown (`4-byte discriminant + 8-byte rentExemptReserve = 12`). The voter offset at 124 is correct (4 discriminant + 120 for full Meta = 124), but it's the kind of constant a future reader will doubt. Worth adding the same derivation comment since it's already in the file.

---

**q:** `@vultisig/sdk` minor bump in changeset - are any of the new `staking/*` modules re-exported from `@vultisig/sdk` itself, or is this a pure changeset dependency cascade because sdk depends on core-chain? If it's the latter, fine as-is. Just want to confirm the windows consumer imports from `@vultisig/core-chain` paths directly (as the package.json exports map implies) and not from a re-exported sdk barrel.

---

**praise:** the rejection-over-fabrication stance in `parseStakeAccount` is exactly right for a layer this close to funds. A `0n` stake or empty authority would silently corrupt downstream state; returning `undefined` and letting the scan skip the row is the correct default.

---

PR #912, opened 2026-06-29
feat(solana): validator-metadata seam (Stakewiz, swappable)

Ehsan-saradar:
**Phase 2** of Solana native staking — parity with iOS #4660.

Tracking sub-issue: vultisig/vultisig-windows#4237 · Epic: vultisig/vultisig-windows#4222 · Builds on Phase 1 (#900, merged).

A swappable off-chain enrichment seam for validator display metadata (name / logo / APY estimate / score), layered on the Phase 1 on-chain `getVoteAccounts` rows. The core stake/commission/activation data stays on-chain; this layer only supplies the display metadata most likely to change source (Stakewiz → validators.app → in-house → on-chain-only).

### What's here — `chains/solana/staking/metadata/`
- **`ValidatorMetadataProvider`** — provider interface. **Contract: never throws** — a failed/rate-limited source returns a partial or empty map so callers degrade to a truncated vote pubkey + on-chain commission, no logo, no crash.
- **`stakewizProvider`** — concrete impl over `api.stakewiz.com/validators` (one fetch enriches the whole set). Mapping: `name`, `image`→`logoUrl`, `apy_estimate` (percent→fraction), `wiz_score`→`score`. Degrades to `{}` on any non-OK / throw / parse error.
- **`enrichValidators`** — merges the metadata map onto the validator rows; the provider is injectable for tests.

### Tests
- `stakewizProvider.test.ts` — row→metadata mapping (percent→fraction, score rounding), graceful degradation (non-OK / fetch throw), and no-fetch on empty input.
- `enrichValidators.test.ts` — merge onto matching validators only; passthrough when the provider returns nothing.

### Notes
- Caching (~1h) lives in the consumer's react-query layer (windows), not here.
- The APY *resolver* (Stakewiz passthrough → on-chain inflation fallback) lands with the DeFi wiring phase, per iOS.
- Changeset bumps `@vultisig/core-chain` + `@vultisig/sdk` (minor).

## Summary by CodeRabbit

* **New Features**
  * Added Solana staking validator metadata enrichment to enhance validator listings with off-chain display details when available.
  * Introduced an injectable metadata provider with a default provider-backed data source.
  * Added new public entry points for Solana staking metadata enrichment APIs.
* **Bug Fixes**
  * Improved resilience by gracefully falling back to on-chain-only data on provider outages, non-OK responses, or malformed data.
  * Normalized metadata fields for consistent display (e.g., score rounding and APY percent-to-fraction handling, including preserving valid zero values).


gomesalexandre (approved):
solid phase 2, ser. 138 test files green, typecheck clean, graceful degradation works as advertised. no blockers.

## what i checked

- full read of all 7 changed files
- `attempt`/`Result` type mechanics (the `data?: never` / `error?: never` discriminated union - clever, TypeScript handles `result.data` access correctly without explicit narrowing)
- the `||` short-circuit in `'error' in result || !Array.isArray(result.data)` is sound because `Failure.data` is `never | undefined`, so `Array.isArray(undefined)` returns false gracefully
- package.json exports - all three new files wired up correctly
- changeset - `minor` for both `@vultisig/core-chain` and `@vultisig/sdk` is appropriate
- no API keys, no hardcoded auth, no fund-safety surface (this is display metadata only, never touches signing paths)
- ran `yarn test:core` locally: **138 passed / 1 skipped**, 1186 tests - all green
- `yarn typecheck` - **no errors**

---

## findings

`praise:` the hardening commit is the real hero here. catching `null`, primitives, non-string `vote_identity`, routing every field through `asTrimmedString`/`asFiniteNumber` runtime guards, and correctly preserving `apy_estimate: 0` (0% APY ≠ no APY) - all of that is exactly right. the `attempt` wrapper over the full fetch+parse pipeline means zero escape paths for exceptions. well done.

---

`suggestion:` no fetch timeout in `stakewizProvider.ts`. if Stakewiz responds slowly (not an error - just slow), the `metadata()` call hangs indefinitely. the react-query `staleTime` on the consumer side handles caching but doesn't necessarily enforce a per-request timeout. a `signal: AbortSignal.timeout(10_000)` in the fetch options would cap the worst-case at the SDK layer:

```ts
const response = await fetch(stakewizValidatorsUrl, {
  headers: { Accept: 'application/json' },
  signal: AbortSignal.timeout(10_000),
})
```

`AbortSignal.timeout` throws `DOMException` on timeout, which `attempt` catches and degrades to `{}` - zero extra error handling needed. non-blocking since the consumer controls overall request lifecycle, but nice to have.

---

`suggestion:` `toMetadata` always emits explicit `undefined` for absent fields:

```ts
return {
  name: asTrimmedString(row.name),       // undefined when absent
  logoUrl: asTrimmedString(row.image),   // undefined when absent
  apyEstimate: apyFraction(row.apy_estimate), // undefined when absent
  score: wizScore === undefined ? undefined : Math.round(wizScore),
}
```

result: `map[pubkey]` is always `{ name: undefined, logoUrl: undefined, apyEstimate: undefined, score: undefined }` for a row with all-invalid fields, not `{}`. Vitest's `toEqual({})` passes because it treats `undefined`-valued keys as absent - so tests are correct - but `Object.keys(validator.metadata)` would return `['name', 'logoUrl', 'apyEstimate', 'score']` even for an "empty" entry. minor ergonomics footgun if any consumer enumerates keys. could use conditional spread to keep it genuinely sparse, but this is non-blocking.

---

`q:` the `ValidatorMetadataProvider.metadata(votePubkeys)` interface signature implies targeted fetching (you pass the pubkeys you need), but the Stakewiz impl ignores `votePubkeys` for the API call - fetches everything, filters client-side. this is well-documented in the JSDoc on `stakewizProvider.ts`, but the interface-level JSDoc doesn't mention it. a future implementor pointing at a different API that supports `?pubkeys=...` might not realize the interface is agnostic here. worth a one-liner on the interface: _"implementations may fetch more than the requested set and filter internally"_. really non-blocking.

---

## verdict

zero blockers, zero preferably-blocking. graceful degradation is real and tested, not just documented. interface is clean and swappable. merge when approved.

PR #915, opened 2026-06-30
feat(solana): native-staking transaction builder (byte-parity)

Ehsan-saradar:
**Phase 3** of Solana native staking — parity with iOS #4661. The signing core for the delegate flow.

Tracking sub-issue: vultisig/vultisig-windows#4239 · Epic: vultisig/vultisig-windows#4222 · Builds on Phase 1 (#900) + Phase 2 (#912), both merged.

### What's here — `chains/solana/staking/tx/`
- **`stakingPayload`** — discriminated staking-op intent: `delegate` / `unstake` / `withdraw` / move-stake `deactivate` + `redelegate` sub-steps. (The whole op set lands here as one cohesive unit; the later phases add only the windows keysign branches + UI for unstake / withdraw / move.)
- **`buildUnsignedStakingTx`** — maps a payload to the wallet-core Solana stake proto (`delegateStakeTransaction` omits the stake account so wallet-core derives it — create+initialize+delegate in one tx; move-redelegate sets the existing account explicitly), compiles a zero-signature envelope via `TransactionCompiler`, and base64-encodes it.

### MPC byte-parity contract
The validator vote pubkey + amount live on the local-only form, so the initiating device builds these bytes **once** — pinning the recent blockhash and the wallet-core-derived stake-account address — and relays them via `signSolana.rawTransactions`. Every co-signing device then signs the **byte-identical** message through the raw-transaction path, so no device rebuilds the input from a non-round-tripped payload.

### Tests
`buildUnsignedStakingTx.test.ts` builds delegate / deactivate / withdraw / move-redelegate txs against **real wallet-core**, decodes them back through the same `TransactionDecoder` the relay path uses, and asserts determinism (byte parity across devices).

### Notes
- Adds `long` to core-chain deps (Long-typed proto amount fields).
- Changeset bumps `@vultisig/core-chain` + `@vultisig/sdk` (minor).

## Summary by CodeRabbit

* **New Features**
  * Added Solana native staking transaction building for delegate, unstake/deactivate, withdraw, and move-redelegate flows.
  * Introduced deterministic unsigned transaction output with wallet-compatible compilation and decoding.
  * Exposed new staking payload and transaction builder modules for integrations.
* **Bug Fixes**
  * Improved staking amount compatibility for protocol-friendly serialization.
* **Tests**
  * Added coverage to validate decoded staking transaction structure and byte-level determinism.

### DIFF
--- a/.changeset/solana-staking-tx-builder.md
+++ b/.changeset/solana-staking-tx-builder.md
@@ -1,0 +1,24 @@
+---
+"@vultisig/core-chain": minor
+"@vultisig/sdk": minor
+---
+
+feat(solana): native-staking transaction builder (byte-parity)
+
+Phase 3 of Solana native staking. Adds the signing core for the delegate flow
+(and the proto for the later unstake / withdraw / move-stake ops) under
+`@vultisig/core-chain/chains/solana/staking/tx`:
+
+- `stakingPayload` — discriminated staking-op intent (delegate / unstake /
+  withdraw / move-stake deactivate + redelegate sub-steps).
+- `buildUnsignedStakingTx` — maps a payload to the wallet-core Solana stake
+  proto (`delegateStakeTransaction` derives the stake account; move-redelegate
+  sets it explicitly), compiles a zero-signature envelope via
+  `TransactionCompiler`, and returns it base64-encoded. This is the MPC
+  byte-parity contract: the initiating device builds these bytes once (pinning
+  the recent blockhash + the derived stake-account address) and relays them via
+  `signSolana.rawTransactions`, so every co-signer signs the identical message.
+
+Adds `long` to core-chain deps (Long-typed proto amount fields). Byte-parity
+tests build delegate / deactivate / withdraw / move-redelegate txs against real
+wallet-core, decode them back, and assert determinism.
--- a/packages/core/chain/chains/solana/staking/tx/buildUnsignedStakingTx.test.ts
+++ b/packages/core/chain/chains/solana/staking/tx/buildUnsignedStakingTx.test.ts
@@ -36,57 +36,71 @@ describe('buildUnsignedStakingTx', () => {
       priorityFeeLimit: 200_000,
     })
 
-  const decode = (base64: string) => {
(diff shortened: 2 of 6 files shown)


gomesalexandre (approved):
All tests green (1191/1191 core, including the 5 new staking/tx tests). Checked out branch, ran `yarn test:core`, verified proto field coverage against `core_proto.d.ts`.

## what i checked

- `DelegateStake.stakeAccount` is a real optional field in the wallet-core proto - confirmed in `dist/generated/core_proto.d.ts:40688`. The `moveStakeRedelegate` branch setting it explicitly is correct; wallet-core will skip derivation and use the passed address. The test's `toContain(stakeAccount)` assertion proves this at runtime.
- `unstake` + `moveStakeDeactivate` both mapping to `DeactivateStake` with the same `stakeAccount` is intentional - they're wire-identical; the state machine distinction lives in the caller, not the tx itself.
- `long` dep: already a transitive dep via `protobufjs` (which pins `long@^5.3.2`). Declaring it explicitly in `core-chain/package.json` is correct practice - belt + suspenders so the dep isn't accidentally yanked.
- Byte-stability pinned base64: smart. Any upstream wallet-core encoding change or proto field reordering will shatter the test. Load-bearing by design.
- `7z-wasm` re-ordering: cosmetic alphabetical sort, no behavioral change.

## findings

`suggestion:` `toLong` in `buildUnsignedStakingTx.ts` uses `Long.fromString(value.toString())`. Long 5 ships `Long.fromBigInt(value)` - more direct, same result. Tiny nit, non-blocking.

`suggestion:` `unstake` and `moveStakeDeactivate` produce byte-identical deactivate txs. A one-line comment on the `moveStakeDeactivate` arm like `// same wire tx as unstake; caller tracks which step of the move-stake flow this is` would save future readers from a double-take. Non-blocking.

`praise:` The `moveStakeRedelegate` + `stakeAccount` explicit-set design is clean. Fresh delegate omits it so wallet-core derives a new account; redelegate passes it explicitly to reuse the existing one. The distinction is well-documented and the test verifies it at runtime. Well done ser.

PR #916, opened 2026-06-30
fix(solana): surface the on-chain rejection reason on broadcast

Ehsan-saradar:
## Problem

When a Solana `sendTransaction` is rejected at **preflight**, the RPC returns the actionable detail in `data.err` / `data.logs` (the program logs). web3.js exposes those via `SendTransactionError.logs`, but leaves the bare `.message` as an uninformative `"failed to send transaction"`. Any consumer reading only the top-level message — e.g. an error screen that shows `extractErrorMsg(error)` — lost the real reason (`insufficient lamports`, a custom program error, a failed instruction).

## Fix

The Solana broadcast resolver now folds the program logs into the thrown error's `message` (preserving the original error as `cause`) before it propagates through `verifyBroadcastByHash`. The on-chain reason reaches the surface for **every** consumer of `@vultisig/core-chain`, not just one UI.

- Only `SendTransactionError` instances with non-empty `logs` are rewrapped; anything else (including the `already been processed` / `AlreadyProcessed` idempotent-success path) is untouched and forwarded verbatim.
- The reviewed-and-accepted broadcast-layer trade-off for `AlreadyProcessed` is unchanged.

## Tests

`resolvers/solana.test.ts` — added a case asserting program logs are hoisted into the verified error message with `cause` preserved, and pinned that a non-`SendTransactionError` is forwarded verbatim. Full solana broadcast suite: 7/7 pass. core typecheck + lint + format clean.

## Release

Changeset bumps `@vultisig/core-chain` and `@vultisig/sdk` (bundled-source guard passes).

## Companion

UI/headline half of the same issue lives in `vultisig-windows#4253` (keysign broadcast errors headlined as on-chain instead of device-timeout). That repo's fix is self-contained against the currently published packages; this PR makes the reason richer at the source.

## Summary by CodeRabbit

* **Bug Fixes**
  * Solana broadcast failures now include clearer on-chain rejection details in the error message, making RPC preflight rejections easier to understand.
  * When transaction sending fails, the original error is still preserved as the underlying cause.
  * Improved handling so non-specific RPC errors continue to be passed through unchanged.

### DIFF
--- a/.changeset/solana-broadcast-onchain-reason.md
+++ b/.changeset/solana-broadcast-onchain-reason.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Solana broadcast: surface the on-chain rejection reason. When `sendTransaction` is rejected at preflight, the RPC returns the actionable detail in `data.err` / `data.logs` (the program logs), which web3.js exposes via `SendTransactionError.logs` while leaving the bare `.message` ("failed to send transaction") uninformative. The broadcast resolver now folds those program logs into the thrown error's message (preserving the original error as `cause`), so consumers reading the top-level message see *why* the network rejected the transaction — "insufficient lamports", a custom program error, a failed instruction — instead of just that it failed.
--- a/packages/core/chain/tx/broadcast/resolvers/solana.test.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/solana.test.ts
@@ -1,3 +1,4 @@
+import { SendTransactionError } from '@solana/web3.js'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 const mocks = vi.hoisted(() => ({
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
Tests green (1187/1188 - baseline branch, doesn't include #915's new files). Checked out branch, read the implementation + test file end-to-end.

## what i checked

- `verifyBroadcastByHash` signature accepts `error: unknown`, so returning `unknown` from `withSolanaBroadcastReason` is type-safe.
- Ordering in the catch block: `isInError(error, 'already been processed', 'AlreadyProcessed')` fires on the ORIGINAL error before wrapping happens. A `SendTransactionError` containing those strings would still short-circuit correctly.
- Empty-logs edge case: `!logs || logs.length === 0` returns the original error unchanged. Correct.
- The test change from `'already processed'` -> `'blockhash not found'` is a clarity win - the new message makes it obvious we're testing the "generic non-AlreadyProcessed rejection" path, not accidentally brushing the early-return branch.
- New test asserts `error.cause === sendError` - so the full `SendTransactionError` (with all logs) is still accessible to anything that unwraps the cause chain.

## findings

`praise:` Preserving the original `SendTransactionError` as `cause` is thoughtful. Any error-reporting layer that does `error.cause.logs` still gets the raw program logs - the wrapped message is for human-readable surfaces, not the only access path.

`suggestion:` `withSolanaBroadcastReason` return type is inferred as `unknown`. A more precise overload or explicit return type annotation (`Error | unknown`) would make the function signature self-documenting - readers wouldn't need to trace the conditional to know "if it's a SendTransactionError it returns a new Error, otherwise original". Non-blocking.

PR #722, opened 2026-06-13
refactor(cardano): attach CIP-20 memo via WalletCore native auxiliary_data

Ehsan-saradar:
Supersedes the client-side CIP-20 body-patching introduced in #583 with WalletCore's **native `auxiliary_data`** support.

## Background
#583 attached a Cardano memo (CIP-20, metadata label 674) by hashing the metadata and patching the WalletCore-produced tx body in TypeScript (`patchTxBodyWithAuxHash`), then recomputing the signing hash and re-assembling the signed tx. WalletCore now does this natively: [trustwallet/wallet-core#4784](https://github.com/trustwallet/wallet-core/pull/4784) adds an `auxiliary_data` field to the Cardano `SigningInput` — WalletCore commits its Blake2b-256 hash into tx body key 7 (`auxiliary_data_hash`) and embeds the bytes in the signed transaction.

## ✅ Unblocked — wallet-core 4.7.0 is published
The upstream fix shipped in [`@trustwallet/wallet-core@4.7.0`](https://github.com/trustwallet/wallet-core/releases/tag/4.7.0). This branch is bumped to `^4.7.0` and rebased onto current `main`.

## Changes
- **`@trustwallet/wallet-core` → `^4.7.0`** across `core/chain`, `core/mpc`, and `sdk` (+ `yarn.lock`).
- **`signingInputs/resolvers/cardano.ts`** — set `SigningInput.auxiliaryData` to the CIP-20 CBOR when a memo is present.
- **`tx/preSigningHashes/index.ts`** — drop the Cardano-memo special case; the native pre-image hash already commits the aux-data hash, so no client-side patching/rehashing.
- **`tx/compile/compileTx.ts`** — use the WalletCore body as-is (it already carries key 7); still embed the aux-data bytes as signed-tx element [3].
- **`keysign/chainSpecific/resolvers/cardano.ts`** — the fee estimator now prices the WalletCore body as-is (already carries key 7) plus the aux-data bytes, instead of patching the body to size it. Its test mirrors the native compile path.
- **`buildCip20AuxData.ts`** — remove the now-unused `patchTxBodyWithAuxHash` helper (+ its unit tests).
- **`compileTx.golden.test.ts`** — set `auxiliaryData` on the fixture and assert the body commits `blake2b-256(auxData)` at key 7 natively (plus a no-memo regression guard).
- **changeset** — patch-bumps `@vultisig/core-chain`, `@vultisig/core-mpc`, `@vultisig/sdk`.

[...]

PR #921, opened 2026-06-30
feat(cardano): attach cip-20 memo via walletcore native auxiliarydata

gomesalexandre:
> **DO NOT MERGE before 2026-07-01 07:18 UTC** - see supply-chain note below.

Picks up @Ehsan-saradar's work from #722 and restores the yarn quarantine that was bypassed to unblock CI.

## What

Native CIP-20 auxiliary data support for Cardano transactions via `@trustwallet/wallet-core`'s `auxiliaryData` field (added in 4.7.0). Replaces the fragile manual CBOR-patching approach that previously attached memo data after signing.

- `buildCip20AuxData` - builds the CIP-20 auxiliary data structure
- `keysign/signingInputs/resolvers/cardano.ts` - passes `auxiliaryData` directly into the signing input so wallet-core commits the aux-data hash into the tx body
- `tx/compile/cardano/buildSignedCardanoTx.ts` - embeds the aux-data bytes in the final CBOR envelope

## Supply-chain note

`@trustwallet/wallet-core@4.7.0` was published today at **07:18 UTC 2026-06-30**. Yarn 4's `npmMinimalAgeGate` quarantines packages published within the last 24h as a supply-chain safeguard - exactly the kind of thing we want respected. #722's CI was failing on this and two bypass commits were pushed (`eda74e0c5`, `aec157992`) to work around it.

This PR reverts that bypass (see `27d9e3c1e`) and keeps the quarantine intact. CI will fail on `Lint and Type Check` / `quality:contracts` until the gate clears around **07:18 UTC 2026-07-01**. That's intentional. Do not admin-merge to skip CI.

@Ehsan-saradar - taking this over from your #722, no changes to the actual Cardano code, just cleaning up the CI bypass. You're credited as co-author on the original commits.

## Risk

Low - Cardano only, no impact on other chains.

### DIFF
--- a/.changeset/cardano-native-aux-data.md
+++ b/.changeset/cardano-native-aux-data.md
@@ -1,0 +1,15 @@
+---
+"@vultisig/core-chain": patch
+"@vultisig/core-mpc": patch
+"@vultisig/sdk": patch
+---
+
+refactor(cardano): attach CIP-20 memo via WalletCore native auxiliary_data
+
+Bumps `@trustwallet/wallet-core` to `4.7.0`, which adds the Cardano
+`SigningInput.auxiliary_data` field. The Cardano memo path now hands the
+CIP-20 CBOR straight to WalletCore, which commits its Blake2b-256 hash into
+tx body key 7 and embeds the bytes in the signed transaction — replacing the
+client-side body patching and re-hashing in TypeScript. The chain-specific
+fee estimator prices the WalletCore body as-is (it already carries key 7),
+and the now-unused `patchTxBodyWithAuxHash` helper is removed.
--- a/packages/core/chain/package.json
+++ b/packages/core/chain/package.json
@@ -2028,7 +2028,7 @@
     "@solana/web3.js": "^1.98.4",
     "@ton/core": "^0.63.1",
     "@ton/crypto": "^3.3.0",
-    "@trustwallet/wallet-core": "^4.6.15",
+    "@trustwallet/wallet-core": "^4.7.0",
     "@vultisig/core-config": "0.9.1",
     "@vultisig/lib-utils": "0.10.3",
     "bip32": "^5.0.1",
--- a/packages/core/mpc/keysign/chainSpecific/resolvers/cardano.test.ts
+++ b/packages/core/mpc/keysign/chainSpecific/resolvers/cardano.test.ts
@@ -9,7 +9,7 @@ import { beforeAll, describe, expect, it, vi } from 'vitest'
 import { Chain } from '@vultisig/core-chain/Chain'
 
 import { getCardanoSigningInputs } from '../../signingInputs/resolvers/cardano'
-import { buildCip20AuxData, patchTxBodyWithAuxHash } from '../../../tx/compile/cardano/buildCip20AuxData'
+import { buildCip20AuxData } from '../../../tx/compile/cardano/buildCip20AuxData'
 import { buildSignedCardanoTx } from '../../../tx/compile/cardano/buildSignedCardanoTx'
 import { CardanoChainSpecific } from '../../../types/vultisig/keysign/v1/blockchain_specific_pb'
 import { CoinSchema } from '../../../types/vultisig/keysign/v1/coin_pb'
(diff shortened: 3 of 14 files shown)


NeOMakinG (approved):
**APPROVE (code) — but DO NOT MERGE before 2026-07-01 07:18 UTC.** The supply-chain quarantine on `@trustwallet/wallet-core@4.7.0` (published 2026-06-30 07:18 UTC, <24h) is intentional and correctly restored; merging before the gate clears would subvert it. The Cardano CIP-20 keysign change itself is correct and verified at runtime against the real native field.

## testing approach
Two-lane: Claude load-bearing + Codex gpt-5.4 (0-byte shared-session stall, best-effort, did not block). Worktree `/tmp/wt-vultisig-sdk-921` at HEAD `27d9e3c1`. Installed real `@trustwallet/wallet-core@4.7.0` (the main repo install has no age gate — only the temp `quality:contracts` consumers do) and ran the affected suites: **22/22 pass** — `compileTx.golden.test.ts (7)`, `buildCip20AuxData.test.ts (13)`, `chainSpecific/resolvers/cardano.test.ts (2)`. Cross-checked the CI Unit Tests job: the same three suites passed there against 4.7.0 as well.

## the pre-sign-hash ⇄ aux_data_hash match (the load-bearing check)
This PR removes the fragile manual `patchTxBodyWithAuxHash` (client-side CBOR map-count rewrite + manual key-7 append + re-hash) and instead hands the CIP-20 CBOR to wallet-core 4.7.0 via `SigningInput.auxiliary_data`, letting it commit `blake2b-256(aux)` into tx body key 7 natively. The full consistency chain is asserted end-to-end in the golden test and verified against real 4.7.0:
- `body['7'] == blake2b(auxDataCbor)` — wallet-core's native key-7 commitment equals the blake2b of the **exact** `auxDataCbor` that TS builds and embeds as signed-tx element [3]. This is a genuine cross-check (body comes from wallet-core output; the hash is computed independently from TS), not a self-referential snapshot. It proves the TS path and native path agree on the bytes → no aux-hash mismatch → chain won't reject.
- `compiledOutput.txId == hashHex (MPC-signed) == preOutput.dataHash` — the hash MPC signs is blake2b of the body that actually gets submitted (with key 7). No sign-the-wrong-tx hole.
- No-memo regression guard is non-vacuous: same fixture without aux yields a different body hash and signs the plain dataHash.

Determinism holds for all memos: `buildCip20AuxData` emits canonical CBOR (single-key `{674:{msg:[...]}}` maps → no key-ordering ambiguity; minimal int encoding; definite lengths), and all three call sites (signing-input resolver, `compileTx`, fee estimator) gate identically on `memo` truthiness — empty/undefined memo ⇒ no aux on every path, so no spurious empty `aux_data_hash`. UTF-8 chunking walks back off continuation bytes so multibyte codepoints are never torn across the 64-byte CIP-20 boundary.

## supply-chain / quarantine restoration
Revert commit `27d9e3c1` removes `npmMinimalAgeGate: 0` from the three temp-consumer `.yarnrc.yml` writes in `scripts/quality-contracts.mjs`, reverting them to yarn's default age gate (quarantine **ON**). The bypass is **not** shipping; the protection is restored and self-enforcing — `Lint and Type Check` / `quality:contracts` will stay red until ~07:18 UTC 2026-07-01 (confirmed in CI: `YN0016 … ^4.7.0 … quarantined`). The wallet-core bump is deliberate and pinned: `yarn.lock` carries the new `4.7.0` resolution + checksum. Native bump on the signing core is the exact RN-pin/supply-chain class, handled correctly here (held, not force-merged).

## no non-Cardano regression
The only shared-path edits are: (1) removing the Cardano-memo branch + now-unused `blake2b` import from `preSigningHashes/index.ts` — Cardano now flows the generic `output.dataHash` path (no-memo Cardano was already on it; with-memo now relies on the native key-7 body); (2) `compileTx.ts` Cardano block uses `preOutput.data` as-is. EVM/cosmos/UTXO/Bittensor/QBTC/PSBT branches untouched. The golden suite (7 tests, multiple chains incl. Bittensor) passes. No orphaned `patchTxBodyWithAuxHash` references remain; `cborBytes` import correctly dropped (it lived only in the deleted helper).

## suggestion
`suggestion:` The golden test's aux-hash agreement check (`body['7'] == blake2b(auxDataCbor)`) only exercises a short single-chunk ASCII memo (`vultisig-test`). Agreement is structurally robust (wallet-core treats `auxiliary_data` as opaque bytes; TS emits canonical CBOR), but adding one multi-chunk / multibyte-memo fixture to the golden would be cheap insurance against any future wallet-core canonicalization divergence on multi-element aux arrays.

## q
`q:` The token-send + memo combined path isn't golden-covered. `auxiliaryData` and `tokenAmount` are orthogonal fields on the signing input so risk is low, but a golden vector for "native-token send with memo" would close the last untested combination.

## risk
Low-to-medium. Cardano-only, no other-chain impact; the fund-critical invariant (signed hash covers the submitted-with-aux body, and the embedded aux matches body key 7) is runtime-verified against real 4.7.0. Residual risk is the same-day native wallet-core bump — mitigated by the intentional quarantine hold.

## verdict
**APPROVE the code.** No blocking findings. **Do not merge before 2026-07-01 07:18 UTC** and only after `quality:contracts` goes green on its own once the age gate clears — do not admin-merge to skip CI.

---
*Two-lane review (Claude load-bearing + Codex gpt-5.4 infra-stalled). Runtime-verified against real `@trustwallet/wallet-core@4.7.0`: 22/22 Cardano suites pass locally + in CI. 🤖*

Ehsan-saradar (approved):
If a package is quarantined, we wait for it to be released, except for consumers of Vultisig packages.

PR #924, opened 2026-07-01
fix(terra-classic): align send gas limit with iOS/Android for cross-device co-signing

johnnyluo:
## Problem

Sending LUNC (and other TerraClassic denoms) between a Windows device and an iOS/Android device fails to complete co-signing. The `preKeysignImage` (pre-sign hash) generated by Windows differs from the one generated by mobile, so the MPC signature can't be assembled.

## Root cause

The fee **gas limit** is serialized into the Cosmos `SignDoc` (`AuthInfo.fee.gas_limit`), which is hashed into the pre-sign image. Everything else in the TerraClassic `SigningInput` was already byte-identical across platforms — only the gas limit differed:

| Platform | Gas limit (TerraClassic send) | Source |
|---|---|---|
| iOS | **300000** | `TerraHelperStruct.GasLimit` |
| Android | **300000** | `CosmosHelper.getChainGasLimit()` |
| Windows (this SDK) | **400000** (and **1_000_000** for `uusd`) ❌ | `cosmosGasLimitRecord` / `getCosmosGasLimit` |

A differing gas limit → different SignDoc → different pre-sign hash → co-signing fails.

## Fix

- `Chain.TerraClassic`: `400000n` → `300000n`
- Removed the `uusd → 1_000_000n` override in `getCosmosGasLimit` (mobile uses 300000 for USTC too, so that override broke cross-device USTC sends identically)
- Removed the now-unused `areEqualCoins` import

Both mobile clients hardcode 300000 for every columbus-5 send denom (uluna, uusd bank denoms, IBC transfers, wasm). The burn tax on non-uluna bank denoms is carried in the fee **amount** (precomputed upstream), not the gas limit, so a single gas value with no per-denom override is correct.

## Tests

Updated the two tests that pinned the old values, then ran them:

- `packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts` — ✅ 5 passed
- `packages/sdk/tests/unit/tools/gas/cosmos.test.ts` — ✅ 13 passed
- SDK typecheck ✅ clean
- ESLint ✅ clean on changed files

## Follow-up (consumer side)

After merge, publish `@vultisig/core-chain` and bump the dependency in `vultisig-windows` to pick up the corrected gas limit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized TerraClassic send gas limits so both supported denoms now use the same fixed value.
  * Improved cross-device compatibility by aligning gas-limit behavior with signing expectations, reducing session/signature issues.
* **Tests**
  * Updated unit and suite assertions to reflect the new uniform TerraClassic gas limit.
* **Documentation**
  * Added a release note entry for the TerraClassic gas-limit fix and patch releases.

### DIFF
--- a/packages/core/chain/chains/cosmos/cosmosGasLimitRecord.ts
+++ b/packages/core/chain/chains/cosmos/cosmosGasLimitRecord.ts
@@ -1,6 +1,6 @@
 import { Chain, CosmosChain, IbcEnabledCosmosChain } from '@vultisig/core-chain/Chain'
 
-import { areEqualCoins, CoinKey } from '../../coin/Coin'
+import { CoinKey } from '../../coin/Coin'
 
 const cosmosGasLimitRecord: Record<CosmosChain, bigint> = {
   [Chain.Cosmos]: 200000n,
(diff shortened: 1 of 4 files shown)


gomesalexandre (approved):
substantive but clean. receipts:
- `yarn build:shared` green in fresh tmp clone at `910fcd4e`
- changed core test green: `packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts` 5/5
- sdk workspace tests green after checkout: `yarn workspace @vultisig/sdk test` 2265/2265
- built artifact receipt from `packages/core/chain/dist`: `getCosmosGasLimit(TerraClassic, uluna|uusd|ibc/*) = 300000`, `Terra = 300000`, `Cosmos = 200000`, TerraClassic staking path stays `3000000` for x1 and x3

looks scoped correctly: drops the stale uusd override, keeps staking untouched, and makes the send SignDoc gas limit deterministic across TerraClassic denoms. only non-blocking note is that the PR body should explicitly call out the knock-on fee-amount drop for uusd sends, since `computedFee = gasLimit * gasPrice` now also follows the mobile 300k path.

lg tm ser.

rcoderdev (approved):

packages/core/chain/chains/cosmos/cosmosGasLimitRecord.ts
```diff
@@ -10,22 +10,20 @@ const cosmosGasLimitRecord: Record<CosmosChain, bigint> = {
...
-  [Chain.TerraClassic]: 400000n,
+  // TerraClassic must stay byte-for-byte in sync with iOS
+  // (TerraHelperStruct.GasLimit) and Android (CosmosHelper.getChainGasLimit)
+  // so cross-device co-signing produces the same SignDoc — the gas limit is
+  // part of the pre-sign image. Both mobile clients hardcode 300000 for
+  // every columbus-5 send denom (uluna, uusd bank denoms, IBC transfers,
+  // wasm), so this is a single value with no per-denom override. The burn
+  // tax on non-uluna bank denoms is carried in the fee AMOUNT (precomputed
+  // upstream), not the gas limit.
+  [Chain.TerraClassic]: 300000n,
```
rcoderdev, line 21: This fixes native LUNC parity, but USTC still looks mismatched for cross-device signing. A normal USTC payload has `contractAddress = "uusd"`, so iOS `TerraHelperStruct` falls into the non-native denom branch and sets fee gas to `1000000`; Android `TerraHelper` has the same `setGas(1000000)` branch. If SDK/Windows now serializes `uusd` with `300000`, Windows + mobile co-signing will still produce different SignDocs. Either keep the `uusd` override until mobile changes, or land the matching mobile branch update first.

PR #926, opened 2026-07-01
fix: harden cosmos + match dispatch against undefined chain/denom/id

gomesalexandre:
## what

Guards four crash surfaces that were emitting raw TypeErrors to prod instead of clean typed errors. Root cause traced by a 3-agent investigation: SDK cosmos + derivation code does unguarded property/string access when chain/denom/id arrives as `undefined` at runtime.

agent-backend-ts already got defense-in-depth guards in PR #282; this hardens the SDK itself so it fails closed at the source.

## guards landed

**1. `packages/lib/utils/match.ts` (highest leverage)**

Before: `handler()` would throw `(0 , r[t]) is not a function` (minified engine noise).

After: guards handler lookup — `match: no handler for "<value>"`. Converts EVERY caller site (`getCoinType`, `getChainKind`, `keysign` dispatch) from unreadable minified crashes into a readable domain error.

**2. `packages/core/chain/coin/token/metadata/resolvers/cosmos.ts`**

- `getCosmosTokenMetadata`: `if (!id) throw new Error('getCosmosTokenMetadata: denom id is required')` before any property access. This is the exact `execute_send` `.startsWith` crash surface.
- `deriveTicker`: guard `!denom` before the `.startsWith('x/staking-')` / `.startsWith('x/')` / `.startsWith('factory/')` chain — defence-in-depth so the function never crashes even if reached with an empty denom.

**3. `packages/core/chain/coin/balance/resolvers/cosmos.ts`**

Explicit domain error for empty-string `id` (previously would fall through to a generic `shouldBePresent` throw deep in `getDenom`).

**4. `packages/core/chain/publicKey/getPublicKey.ts`**

Runtime guard for `undefined` chain before `getCoinType` dispatch. TypeScript types it non-optional but callers can violate that at runtime (the `match` guard in #1 already covers the `getCoinType` crash, this makes the error site more local and readable).

## `.registry` delegations crash

The `.registry` crash reported in the telemetry was NOT line-pinned (no stack was captured). I grep'd the full cosmos delegations path (`lcdQueries.ts`, `getCosmosDelegations`, staking resolvers) — there is no unguarded `.registry` access in the current SDK source. The agent-backend-ts PR #282 now captures `err.stack` so the next occurrence will pin the exact SDK location.

## tests

Every guard has a dedicated unit test asserting:
- the error IS an `instanceof Error`
- the message matches `/required|no handler/`
- the message does NOT match `/is not a function|Cannot read properties/`

```
packages/lib/utils/match.test.ts          4/4 pass
packages/core/.../cosmos.test.ts         10/10 pass (2 new undefined/empty-id tests)
packages/core/chain/publicKey/...test.ts  3/3 pass (1 new undefined-chain test)
packages/core/.../balance/cosmos.test.ts  8/8 pass (1 new empty-id test)
```

## check:ci receipts

`yarn test:core` — 1203/1205 tests pass (1 pre-existing Cardano CIP-20 golden test failure on main, 1 skipped; unrelated to this diff)

`yarn workspace @vultisig/sdk test` — 2265/2265 pass

`npx tsc --noEmit --project .config/tsconfig.json` — only pre-existing Cardano `auxiliaryData` TS error (exists on main); our changed files are clean.

`yarn build:shared` — blocked by same pre-existing Cardano TS error. Cross-repo wired verification done directly via vitest on the affected files (same alias resolution the SDK uses):

```
vitest run match.test.ts                           → 4/4
vitest run cosmos.test.ts (token metadata)         → 10/10
vitest run getPublicKey.test.ts + balance cosmos.ts → 11/11
```

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for Cosmos balance, token metadata, and public key lookups so missing or empty inputs now return clear errors instead of runtime crashes.
  * Hardened utility matching to ignore inherited prototype keys and fail safely when no handler exists.
  * Added coverage for these edge cases to ensure cleaner, more predictable error handling.


NeOMakinG (approved):
## testing approach

Two-lane (Claude primary; Codex best-effort — its background run returned an unrelated result). Worst-case adversarial fund-safety lens on a defensive-hardening PR: the danger is an undefined/empty chain/denom/id flowing through to a WRONG chain, a WRONG/empty denom, or an arbitrary dispatch handler. I read every guard, empirically exercised the `match` dispatch against adversarial keys in node, and mutation-checked.

- `match.test` 5/5, core guards (cosmos token-metadata + balance + getPublicKey) 21/21, broad dispatch sweep green. (Core tests need `--config .config/vitest.core.config.ts` for the `@vultisig/core-chain` alias.)
- Empirical node run of `match()` with prototype-name keys (below).

## the four guards are correct

All fail closed, none fall through to a default, and the undefined-vs-empty distinction is handled well:

- **`getPublicKey` / `getCosmosTokenMetadata`** — `!chain` / `!id` throw a clean typed error before any property access. No default chain/denom.
- **`getCosmosCoinBalance`** — `input.id !== undefined && !input.id` throws only for a *defined-but-empty* id, while an **undefined id is correctly allowed as the native-coin case**. That is exactly the invariant-4 line — a legitimately-optional field is not false-rejected, but a malformed empty token id is.
- **`deriveTicker`** `!denom → ''` is display-only (not the signed denom) and defense-in-depth (the `getCosmosTokenMetadata` `!id` throw fires first), so it's benign.
- **`match`** converts the minified `is not a function` crash into a clean `no handler` domain error, and — critically — an unmatched value throws rather than dispatching to an arbitrary handler.

Every guard is mutation-proven by the PR's own tests (removing a guard surfaces the raw `is not a function` / `Cannot read properties` the assertions forbid).

## fix-and-approve: closed a prototype-key hole in the `match` guard

The `typeof handlers[value] !== 'function'` guard alone does **not** reject a value that collides with an inherited `Object.prototype` member name. `handlers['constructor']` resolves to `Object`, `handlers['toString']` to the prototype method — both are functions, so they pass the guard and **silently dispatch to the wrong handler**. Empirically, on the un-fixed guard:

```
match('constructor', { Ethereum, TerraClassic })  -> {}                 (NOT thrown)
match('toString',    ...)                          -> "[object Undefined]" (NOT thrown)
match('isPrototypeOf', ...)                        -> false               (NOT thrown)
```

That is the same misroute this guard exists to prevent for an unknown chain/denom/id, and `match` drives `getCoinType` / `getChainKind` / keysign / `getBlockExplorerUrl` dispatch — with the PR's own premise that callers violate the type at runtime, a prototype-name value from an external source would resolve a coin type / pubkey from the wrong handler.

Fix (commit `55df2193`): require an **own** handler before the typeof check —
```ts
const handler = Object.prototype.hasOwnProperty.call(handlers, value) ? handlers[value] : undefined
```
Now every prototype-name key (`constructor`/`toString`/`valueOf`/`isPrototypeOf`/`hasOwnProperty`/`__proto__`) fails closed with the clean `no handler` error; legit own-property chains are byte-identical. Added a prototype-chain-key test; mutation-proven (reverting the `hasOwnProperty` check turns it red). `match.test` 5/5.

## risk

Low. Pure defensive hardening — turns crashes into fail-closed typed errors, distinguishes native (undefined id) from malformed (empty id), and (with the fix) rejects prototype-name dispatch keys. No behavior change for any valid chain/denom/id. No `@ts-expect-error`/`as any`.

## verdict

**APPROVE.** The cosmos + dispatch paths now fail closed on undefined/empty chain/denom/id, and the `match` guard rejects prototype-name keys too, so no unknown value can misroute to a default or an inherited handler.

---

NeOMakinG (approved):
## testing approach

Round-2 post-approval delta re-review. My r1 APPROVE was at `e2fc2f83`; HEAD advanced to `2e56eb64`, leaving a stale-approval state. Verified the delta with net-diff discipline (git diff `e2fc2f83..2e56eb64` + net vs origin/main). No worktree/test run needed — the delta is a release-note doc, code byte-identical to the r1-approved+tested state.

## prior issues — status
|---|---|---|---|---|---|
| r1 | 1 | fix-and-approve | `match()` dispatch guard bypassed by prototype-member-name keys (`constructor`/`toString`/`isPrototypeOf`/…) → typeof-function passes → wrong-handler misroute | ✅ fixed | my `Object.prototype.hasOwnProperty.call(handlers, value)` own-property check + mutation-proven test **persists byte-identical** at `2e56eb64` (packages/lib/utils/match.ts) |
| r1 | 2 | (verified ok) | PRs 4 undefined/empty guards (cosmos balance native-vs-empty, token-metadata denom, getPublicKey chain) fail-closed | ✅ unchanged | all four resolver files byte-identical `e2fc2f83..2e56eb64` |

## delta findings
- **CODE BYTE-UNCHANGED** — `git diff e2fc2f83 2e56eb64` is exactly **1 commit** (`chore: add changeset for cosmos undefined guard fixes`) touching **only** `.changeset/harden-cosmos-undefined-guards.md` (+10). All code files (match.ts, cosmos balance/token-metadata resolvers, getPublicKey) are byte-identical to the r1-approved SHA — verified via `git diff --quiet` over all four.
- **NET-DIFF vs origin/main clean** — the 9-file net surface is exactly #926s own change (cosmos guards + tests + match.ts + my prototype fix + the new changeset). No smuggled code from sibling merges.
- **CHANGESET correct** — `"@vultisig/sdk": patch` (correct semver for a defensive fix/hardening; correct published package for the va/mcp consumer-bump convention); description `fix(cosmos): harden chain/denom/id dispatch against undefined inputs` matches the actual change. No caret-float introduced (it is a changeset markdown, not a package.json specifier).

## risk
None. Documentation-only delta over an already-approved, byte-identical, fund-safe code change. No regression to the cosmos undefined-guards or the match prototype-key fix.

## verdict
APPROVE (clears the stale-approval state at `2e56eb64`). Code byte-unchanged since r1; changeset is correct.

---
_Two-lane review (Claude load-bearing + Codex gpt-5.4). Codex not run for a changeset-only doc delta — code byte-identical to the r1-tested state._

PR #927, opened 2026-07-01
feat(cli): dual-read signable tool outputs with tx_ready parity cross-check

neavra:
## What

Gives the SDK CLI the same signing mechanism the mobile app uses: read the raw
transaction envelope off the `tool-output-available` SSE channel and enrich it
client-side, instead of relying solely on the backend-enriched `tx_ready` frame.

This is a **dual-read** step. The backend `tx_ready` remains authoritative for
signing whenever it is present and signable — today's behavior is unchanged.
Alongside it, the CLI now derives its own signable candidate from the raw tool
output and **cross-checks the two**, so we can prove the client-side enrichment
against the live backend before making it the sole source in a follow-up.

Generalizes the earlier Polymarket-only bridge into `toolOutputSigning.ts`.

## How

- **Client-side enrichment** (`toolOutputSigning.ts`): a 1:1 port of the backend
  `enrichBuildResult` (flat result → nested `{tx}` wrap + chain copy) and the
  approve→main split, plus a small per-tool normalizer for the payments tools
  that use divergent `to_address` / `calldata` field names.
- **Selection** (`session.ts`): prefer a signable `tx_ready`; fall back to the
  client-side candidate only when no usable `tx_ready` arrives — flat off-chain
  tools that emit no `tx_ready` (`polymarket_deposit`, `polymarket_setup_trading`)
  and tools whose `tx_ready` is structurally unsignable (`build_custom_*`). One
  signable tx per turn (first-wins); a second frame is deferred and reported.
- **Parity cross-check**: when both channels produce a payload for a turn, their
  `{to, value, data, chain, chain_id, tx_encoding, amount, memo}` leg tuples are
  compared and any divergence is logged loudly.
- **Chain safety**: the chain guard is generalized past Polygon to any supported
  EVM chain and **fails closed** on a missing/inconsistent chain — it never
  defaults to a chain. (The backend injects chain from tool-call arguments the
  CLI doesn't have, so a candidate that can't self-describe its chain is dropped
  and `tx_ready` remains the signer.)

## Scope / safety

- Entirely within `clients/cli`. No agent-backend or mcp-ts changes.
- No behavior change for `execute_*` and other `produces_calldata` tools: they
  still sign off `tx_ready`; their tool output is used only for the parity check.
- Non-EVM paths (utxo / solana / cosmos incl. THOR/Maya memo routing) are covered
  by the parity comparator and continue to sign off `tx_ready` unchanged.

## Testing

- Unit tests for the enrichment port, the generalized chain guard, per-tool field
  normalization, and the fail-closed guards.
- Parity comparator tests (matching payloads, divergences on data/chain/leg-count/
  memo, tx_ready-exclusive fields, non-EVM).
- Session dual-read tests: single tool-output candidate signs; two → first-wins +
  deferral; both channels → `tx_ready` authoritative (no double-store); unsignable
  `tx_ready` → tool-output fallback; `tx_ready`-only unchanged.
- `yarn typecheck` + `yarn workspace @vultisig/cli typecheck`, `yarn check`, and
  `yarn test:cli` (478 tests) all green.

Draft for review.

## Summary by CodeRabbit

* **New Features**
  * Added support for signing eligible tool outputs in the CLI, including multi-step transaction flows.
  * Improved transaction confirmation text to better reflect approvals and wrapped transactions.

* **Bug Fixes**
  * Prevents signing when transaction details are missing, inconsistent, or unsupported.
  * Ensures only one valid transaction is signed per turn and avoids accidental double-signing.
  * Improves recovery behavior after disconnects by warning when a transaction can’t be safely reconstructed.


gomesalexandre (changes requested):
Need real receipts before I can green-light this.

1. This is a signing-path change in `clients/cli`, so unit coverage alone is not enough here. Per the repo rules I need a real CLI runtime receipt against a live agent-backend flow that exercises the dual-read path, not just `test:cli` output.
2. Please also include the parity/runtime evidence for at least one concrete signable tool path affected by this port. The important contract here is that the client-enriched payload and backend `tx_ready` stay aligned on an actual run, not only in fixtures.

Side note: I tried to get a clean local static receipt, but `yarn build:all` is currently red on both this branch and `origin/main` with the existing Cardano `auxiliaryData` typing failure, so that build result does not currently tell us anything about this PR specifically.

gomesalexandre (commented):
comment, hold.

One material issue from a scoped pass on the current draft head.

medium
`clients/cli/src/agent/session.ts:731`

When both channels are present and parity FAILS, `selectAndBufferSignable()` still queues the flat client-derived `tool_output` candidate as a signing source whenever the paired `tx_ready` is structurally unsignable. That defeats the stated Phase-1 posture: parity divergence is supposed to prove the client-side enrichment against the live backend before tool-output becomes trusted, but today a diverging candidate can still be the tx that gets signed.

Why this matters:
- `diffToolOutputParity()` already normalizes the important field-shape differences (`to_address` vs `to`, `calldata` vs `data`, etc.), so a remaining divergence here is not cosmetic.
- The exact risky path is the one this PR adds for `build_custom_*`: unsignable `tx_ready`, signable flat tool-output fallback.
- In that state, a backend-side enrichment change or a client port bug becomes a sign-the-wrong-bytes risk instead of a safe fail-closed error.

Suggested fix:
- only enqueue the flat `tool_output` fallback when either there is no `tx_ready` at all, or parity matched
- if `tx_ready` exists and parity diverges, fail closed to the current `tx_ready` path instead of signing the client-enriched candidate

low
`clients/cli/src/agent/session.ts:717-718` and `:753-754`

The parity log / deferred notification treat any same-turn `tool_output` + `tx_ready` pair as if they were the same transaction. If a turn ever carries unrelated candidates from different tools, you get a loud false `[DIVERGENCE]` signal and a misleading "deferred" notification even though the tx_ready one was the tx that got signed. Not a wrong-sign bug, but it muddies exactly the Phase-1 telemetry this PR is trying to collect.

gomesalexandre (approved):
re-review pass — CHANGES_REQUESTED resolved. the two concerns from last round:

**fail-closed on divergence** - `selectAndBufferSignable` correctly gates flat tool-output as a sign source ONLY when (a) no `tx_ready` at all, or (b) same-turn `tx_ready` matched parity. diverging flat cannot sign — it falls closed to `tx_ready` path. explicitly called out in method doc.

**pairing vs sign gate separation** - pairing (toolCallId match) affects ONLY divergence TELEMETRY, never the sign decision. so a wrong pairing can't accidentally authorize a sign. gucci.

test suite (toolOutputSigning.test.ts 422L + toolOutputParity.test.ts 237L + client.test.ts phase-1 dual-read block) + runtime receipts posted by @neavra seal it.

PR #952, opened 2026-07-03
fix(cli): make agent ask truthful + idempotent around broadcast (F1/F3/F14)

neavra:
## Summary

Bundles the fund-safety fixes for the `agent ask` process into one coherent **"make the agent process truthful"** change (audit `020726-a2a-01` F1 + F14 CLI half, and `230626-sdkcli-fix-06` truthful-output). Backend `Idempotency-Key` half is intentionally **out of scope** (separate task).

The core hazard: a tx is signed + broadcast on-chain, and only THEN reported back to the backend. If that follow-up POST fails, `agent ask` exited 1 with the tx hash effectively lost, so a naive caller retried in a fresh process → new turn → new tx → **double broadcast**.

## Changes

1. **Result-before-ack / `ACK_FAILED` (F1).** The broadcast hash already reaches the caller before the ack POST (via `emitAndConfirmTx` → `onTxStatus`, recovered in the catch via `partialResult()`). This PR adds the missing piece: a throw *after* a broadcast now exits with a **distinct `ACK_FAILED` code (8)** instead of a generic error, so a headless caller knows the hash is valid and must NOT blindly retry.
2. **Persistent broadcast journal (F1/F14).** Append-only `~/.vultisig/broadcasts.jsonl` (intent fingerprint, hash, chain, ts). Signing a new tx whose identical intent was broadcast in the last 10 min and hasn't *definitively failed* is **refused (exit 4)** — surviving the process death the in-memory guard can't. Multi-leg approve legs are journaled + guarded too. `--force` overrides. Confirmed/failed/timeout resolutions are recorded; only a definitive failure re-opens the intent for an automatic retry.
3. **Exit-code taxonomy (F3).** SSE/backend `error` frames and thrown errors now map onto the typed `ExitCode` taxonomy (`core/errors.ts`) instead of a blanket `0`/`1`, via `agentErrorCodeToExitCode`. Codes documented in `agent ask --help`.

The JSON-to-stdout / single-envelope / IDs parts of fix-06 were **already implemented on `main`** (verified against the worktree — the audit was against a dirty tree); this PR completes the exit-code half and adds regression assertions.

## Tests

- `broadcastJournal.test.ts` — fingerprint stability/normalisation, window, resolution semantics, `--force`, fail-open on corrupt/missing journal.
- `executor.idempotency.test.ts` — `signTxFromBuffer` refuses an identical intent (no second sign), `--force` overrides, distinct intents allowed.
- `agentAskJson.test.ts` — `ACK_FAILED` exit code on broadcast-then-throw; SSE error frame maps to the taxonomy.
- Global CLI test setup isolates the journal (`VULTISIG_BROADCAST_JOURNAL_PATH`) so tests never touch the real `~/.vultisig`.

## Validation

- `yarn typecheck` ✅ · `yarn test:cli` ✅ (418) · `yarn workspace @vultisig/cli test` ✅ (461) · `yarn check` (lint/knip/format) ✅
- Changeset included (`@vultisig/cli` minor).

Live smoke against the local stack is pending (see PR checklist).

## Summary by CodeRabbit

* **New Features**
  * Added `--force` to `vultisig agent ask` to bypass recent duplicate-broadcast protection.
  * Introduced persistent broadcast journaling with cross-process locking and TTL-based stale reservation stealing, plus bounded journal pruning.
  * Added exit codes `ACK_FAILED` and `DUPLICATE_BROADCAST`, and records broadcast confirmation outcomes (confirmed/failed/timeout).
* **Bug Fixes**
  * Prevents accidental double-broadcasts across retries, crashes, and parallel runs.
  * Retags post-broadcast follow-up failures as `ACK_FAILED` only when the broadcast was not yet acknowledged; preserves other retryable errors.
  * Improves resilience to missing/corrupt journal entries.
* **Tests**
  * Expanded coverage for idempotency, reservation/journal semantics, ACK capture/signal behavior, and the isolated CLI test journal setup.


gomesalexandre (commented):
direction is good and the fund-safety intent is legit. the test matrix is also decent. a couple qs still need answers before i bless it:

1. the journal guard is check-then-record across processes. `assertNoRecentDuplicate()` runs before signing, `recordBroadcast()` runs after broadcast. two sibling processes can both pass the check before either records anything, then both send. if the claim is "survives fresh retries" that's true. if the claim is "also blocks sibling processes" it's overstated. either add an atomic reservation / file lock before signing, or soften the claim.
2. `ACK_FAILED` re-tagging currently keys off `partial.transactions.some(t => t.status !== 'failed')`. that can mask a later retryable backend/network error as exit 8 once any earlier tx in the turn was non-failed. if that conservatism is intentional, please document it loudly. otherwise gate exit 8 on an actual broadcast hash for the current ask.
3. `DUPLICATE_BROADCAST` currently shares exit 4 with generic invalid input. json callers can disambiguate, but `$?` alone cannot. a dedicated exit code would make the taxonomy more branchable.
4. append-only `broadcasts.jsonl` is never pruned, but it's fully parsed on every sign. eventually that gets slow. prune or rotate.
5. `cliTestEnv.ts` preserving a pre-set `VULTISIG_BROADCAST_JOURNAL_PATH` and then `rmSync`-ing it before tests is spooky. if a dev exported the real path, tests can wipe the real journal. override to a temp path unconditionally in test setup.

runtime receipts are still owed before merge. i want one live run showing:
- real `agent ask --yes` broadcast
- immediate identical retry refused with exit 4 + prior-hash message
- `--force` override succeeding
- journal file created with sane perms and both broadcast/resolution records

gomesalexandre (approved):
thorough read of broadcastJournal.ts. key fund-safety points:

- fingerprint (sha256 of owner|chain|to|value|data|asset) correctly identifies duplicate intents
- file lock uses O_CREAT|O_EXCL via `openSync(..., 'wx')` — atomically exclusive, correct for unix
- lock stealing handles crashed holders cleanly (stale TTL 5s, reasonable)
- reads fail open (corrupt journal never bricks the wallet) — correct
- writes best-effort (broadcast never blocked by write failure) — correct
- reservation per-fingerprint lock file prevents TOCTOU race between concurrent processes
- confirmed/pending/timeout all guard (only definitive failure allows retry) — correct

508/508 tests pass locally. 🤖

rcoderdev (approved):

PR #956, opened 2026-07-03
fix(cardano): attach and plan per-UTXO native-token data for MPC keysign parity

### LINKED ISSUE #955: [BUG] Cardano co-sign parity broken for CNT-holding addresses — UtxoInfo.cardano_tokens is never decoded, attached, or planned
## Summary

For any Cardano address holding native tokens (CNTs), MPC co-signing between the SDK family (extension/desktop) and iOS/macOS is broken, and SDK-initiated sends are invalid at the node. Reproduced live on 2026-07-03 (secure vault whose ADA address holds SUNDAE + USDM; fixed-amount ADA sends, with and without memo):

1. **iOS/macOS initiates → extension co-signs: keysign fails every time** (pre-image hashes don't match). iOS attaches per-UTXO token data to the keysign payload and its WalletCore planner reconciles the input tokens into the change output; the SDK signer ignores the token fields entirely, so its planner builds a body whose change output carries no TokenBundle → different body bytes → different Blake2b sighash → MPC cannot converge.
2. **Extension initiates → iOS co-signs: keysign converges, node rejects at broadcast** with Ogmios **3123 "value not conserved"** — the SDK-built payload is token-blind, so the planned tx consumes token-carrying UTXOs while its outputs produce ADA only. (iOS co-signers read the wire UTXOs verbatim, so both devices agree on an *invalid* tx.) Funds are safe (fail-at-submit), but the send is impossible.

Fast vaults appear unaffected in practice (VultiServer does not re-derive the input, and fast-vault addresses typically hold no CNTs) — which makes this look intermittent when it is actually deterministic for CNT-holding secure vaults.

## Root cause

commondata's `UtxoInfo` carries per-UTXO Cardano native-asset data:

```proto
message UtxoInfo {
  string hash = 1;
  int64 amount = 2;
  uint32 index = 3;
  repeated CardanoTokenAsset cardano_tokens = 4;  // policy_id, asset_name_hex, amount (decimal string)
}
```

The SDK never picked this up, at all three layers:

- `packages/core/mpc/types/vultisig/keysign/v1/utxo_info_pb.ts` is **stale** — generated from a pre-`cardano_tokens` commondata; only fields 1–3 exist, so the field can't even be decoded off the wire.
- **Initiator:** `packages/core/mpc/keysign/utxo/getKeysignUtxoInfo.ts` builds Cardano `utxoInfo` from the token-blind `getCardanoUtxos` (plain Koios `address_utxos`). The token-aware fetch already exists — `packages/core/chain/chains/cardano/utxo/getCardanoExtendedUtxos.ts` (`_extended: true`, returns `asset_list`) — but only the CIP-30 dApp path uses it.
- **Signer:** `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts` maps each `utxoInfo` entry to `TW.Cardano.Proto.TxInput` with only `hash`/`amount`/`index` — no `tokenAmount` — so WalletCore's planner cannot reconcile input tokens into the change output.

## Reference implementation (mainnet-tested, the parity target)

vultisig-ios:
- Initiator attach: `VultisigApp/Features/Keysign/Services/KeysignPayloadFactory.swift` → `selectCardanoUTXOs` (extended Koios fetch; assets sorted canonically by `(policyId, assetNameHex)` so the proto bytes are deterministic across retries).
- Signer consume: `VultisigApp/Blockchain/Cardano/Signing/Cardano.swift` → `getPreSignedInputData` (maps `cardanoTokens` → `CardanoTokenAmount{policyID, assetNameHex, amount}` where `amount` is **minimal big-endian unsigned bytes** — `[0x00]` for zero).

## Fix

1. Regenerate the commondata-derived proto types so `UtxoInfo.cardanoTokens` exists (`CardanoTokenAsset` message included).
2. `getKeysignUtxoInfo`: for Cardano, use `getCardanoExtendedUtxos` and attach `cardanoTokens` (canonical `(policyId, assetNameHex)` sort, quantity as decimal string).
3. `getCardanoSigningInputs`: map `utxoInfo[i].cardanoTokens` → `TW.Cardano.Proto.TokenAmount` on each `TxInput` (amount via the existing `amountToBytes` — minimal big-endian, `00` for zero — byte-identical to iOS).
4. Tests pinning byte-parity against the iOS vectors (pre-image hash for a fixed UTXO set with tokens must match an iOS-derived golden hash), plus a regression test that a token-blind payload and a token-aware payload produce different bodies (documenting the mechanism).

⚠️ MPC byte-parity constraint: the mapped bytes must be identical to iOS's or the keysign failure just moves. Same class as the ZEC ZIP-317 divergence (vultisig-ios#4726) — a Cardano CNT vector in the cross-platform signing-vector net (vultisig/commondata#94) would pin this permanently.

## Follow-ups once merged

- vultisig-windows: bump the `@vultisig/core-mpc` / `@vultisig/core-chain` pins so desktop + extension ship the parity fix.
- Related but separate: max-amount ADA sends from CNT-holding addresses are invalid on **all** platforms today (`useMaxAmount` drains lovelace and drops input tokens → 3123). iOS reproduces this on main; the SDK reproduces it a fortiori (token-blind). Worth its own issue: fail closed or emit a token-bundle change output.

gastonm5:
Closes #955

## Root cause

For any Cardano address holding native tokens (CNTs), MPC co-signing between the SDK family (extension/desktop) and iOS/macOS was broken at three layers — the SDK never adopted commondata's `UtxoInfo.cardano_tokens = 4`:

1. **Stale proto types** — `packages/core/mpc/types/vultisig/keysign/v1/utxo_info_pb.ts` was generated from a pre-`cardano_tokens` commondata (fields 1–3 only), so per-UTXO token data couldn't even be decoded off the keysign wire.
2. **Token-blind initiator** — `getKeysignUtxoInfo` built Cardano `utxoInfo` from plain Koios `address_utxos`; the token-aware `getCardanoExtendedUtxos` existed but only served the CIP-30 dApp path.
3. **Token-blind signer** — `getCardanoSigningInputs` mapped each `UtxoInfo` to a WalletCore `TxInput` with no `token_amount`, so the planner couldn't reconcile input tokens into the change output.

Live consequences (reproduced 2026-07-03 on a secure vault holding SUNDAE + USDM):

- **iOS initiates → extension co-signs:** the SDK plans a change output without a TokenBundle → different body bytes → different Blake2b sighash → keysign never converges.
- **Extension initiates → iOS co-signs:** both devices agree on a token-blind body that consumes token UTXOs without re-emitting them → node rejects at broadcast with Ogmios 3123 "value not conserved".

## Fix (byte-parity with iOS)

Mirrors the mainnet-tested iOS implementation exactly:

| Layer | This PR | iOS reference |
|---|---|---|
| Proto types | `utxo_info_pb.ts` regenerated from current commondata with `buf generate` + `protoc-gen-es@2.11.0` (`target=ts`, same generator version as before — the `generatedProtobufHeaders` guard counts are unchanged) | `commondata/proto/vultisig/keysign/v1/utxo_info.proto` |
| Initiator attach | `getKeysignUtxoInfo` Cardano branch now uses `getCardanoExtendedUtxos` (UTXOs sorted by `(hash, index)`) and attaches `cardanoTokens` — assets sorted canonically by `(policyId, assetNameHex)`, hex lowercased, quantities as decimal strings | `KeysignPayloadFactory.selectCardanoUTXOs` + `CardanoService.getExtendedUTXOs` + `CardanoExtendedUtxo.init` |
| Signer consume | `getCardanoSigningInputs` maps `utxoInfo[i].cardanoTokens` → `TW.Cardano.Proto.TokenAmount { policyId, assetNameHex, amount }` per `TxInput`, amount as **minimal big-endian unsigned bytes** via the existing `amountToBytes` (`[0x00]` for zero), `token_amount` left unset for token-free UTXOs | `Cardano.getPreSignedInputData` (`unsignedBigEndianBytes`) |

Why the bytes converge: the initiator serializes the payload once and co-signers read it verbatim (no per-device Koios fetch), protobufjs and SwiftProtobuf both serialize fields in field-number order (`policy_id = 1`, `amount = 3`, `asset_name_hex = 4`; empty repeated fields emit nothing), and WalletCore's `doPlan()` with the shared `forceFee = byteFee` reconciles tokens identically on every device — the same mechanism that already keeps ADA-only sends converging. The chain-specific fee estimator runs through the same resolver, so CNT-inflated bodies now price correctly with no further change.

## Golden vector (please cross-verify on iOS before merge)

Pinned in `packages/core/mpc/keysign/signingInputs/resolvers/cardano.test.ts` on real WalletCore wasm (4.7.0), independently reproduced by a second implementation during review:

```
coin:            ADA native (decimals 6)
sender/change:   addr1vyxk54m7j3q6mrkevcunryrwf4p7e68c93cjk8gzxkhlkpsjpczl2
                 (ed25519Cardano key: spending key = pub(privkey 0x01·32), chain code = 0x02·32, both repeated)
recipient:       addr1vyqgk3uyfkfgzt7rp50s4jdkl0ecw7xvh2wmsvf2myreq7gd27kn4
                 (spending key = pub(privkey 0x02·32), chain code = 0x03·32)
toAmount:        1000000 lovelace, sendMaxAmount false, no memo
byteFee:         200000, ttl: 500000
UTXO 1:          hash 11×32, index 0, 5000000 lovelace, cardanoTokens:
                   - policyId 9a9693a9a37912a5097918f97918d15240c92ab729a0b7c4aa144d77,
                     assetNameHex 53554e444145 (SUNDAE), amount "4500000"  → bytes 44aa20
                   - policyId c48cbb3d5e57ed56e276bc45f99ab39abe94e6cd7ac39fb402da47ad,
                     assetNameHex 0014df105553444d (USDM), amount "665000" → bytes 0a25a8
UTXO 2:          hash 22×32, index 1, 3000000 lovelace, no tokens

expected pre-image dataHash (token-aware): 839ae494ce1af23729a8e918c2d63febb688f0055450865016721c2a62fba93b
token-blind dataHash of the same payload:  db6bde29ccda113233a4ac6bc668fd14ad114ca013698948cfe0d7aa818b7903
```

To cross-verify on iOS: build the identical `KeysignPayload` and run `CardanoHelper.getPreSignedImageHash` — it must return `839ae494…`. The `db6bde29…` value is what the pre-fix SDK derived (pinned as a regression documenting the divergence mechanism).

## Tests

- `getKeysignUtxoInfo.test.ts` — initiator attach: extended Koios fetch, `(hash, index)` UTXO ordering, canonical `(policyId, assetNameHex)` asset sort, lowercase normalization, decimal amounts, wire round-trip of `cardanoTokens` (the decode layer of the bug).
- `resolvers/cardano.test.ts` — signer consume on real WalletCore wasm: exact `TokenAmount` bytes, token-free UTXOs unset, golden pre-image hash, change-output token re-emission (value conservation), token-blind vs token-aware divergence with both hashes pinned.

## Gate

- `yarn check`: exit 0 (typecheck + lint + knip + format + shared-exports all green)
- `yarn test` (full suite): exit 0 — SDK 160 files / 2265 tests, Rujira 24 / 546, core 146 / 1229 (includes the new Cardano suites), CLI 30 / 444, MCP 3 / 33
- Cross-model review (Codex, read-only) per commit + whole branch: 2 findings on the initiator commit (null `asset_name` for unnamed assets; non-canonical quantity strings) — both fixed in-branch with tests; the resolver commit's golden hash was independently reproduced by the reviewer; the whole-branch pass flagged only the pre-existing CNT-*send* divergence noted below (deferred, out of scope)

## Changeset

`patch` for `@vultisig/core-chain`, `@vultisig/core-mpc`, `@vultisig/sdk` (`.changeset/cardano-cnt-cosign-parity.md`), following the `cardano-native-aux-data` precedent.

## ⚠️ Before merge

- **Cross-device iOS ↔ extension co-sign validation on a CNT-holding mainnet address is required** — both directions (iOS initiates / extension co-signs, and extension initiates / iOS co-signs), fixed-amount ADA send. The golden vector is SDK-derived; run the fixture through iOS `getPreSignedImageHash` as a cheaper first check.
- Follow-up once merged: **vultisig-windows must bump its `@vultisig/core-mpc` / `@vultisig/core-chain` pins** so desktop + extension actually ship the fix.
- Known related-but-separate issues (unchanged by this PR): max-amount ADA sends from CNT-holding addresses are invalid on all platforms; the CNT-*send* flow has an iOS↔SDK recipient-lovelace divergence (iOS floors at 1.5M lovelace, SDK passes `toAmount`).

## Summary by CodeRabbit

* **New Features**
  * Cardano transactions now handle native tokens more reliably during signing and planning.
  * Token details are preserved end-to-end, improving compatibility for wallets holding multiple assets.

* **Bug Fixes**
  * Fixed cases where Cardano signing could fail or drop tokens for token-bearing UTXOs.
  * Improved transaction hash and TTL handling for Cardano payloads with token data.
  * Made UTXO ordering and token formatting consistent for more predictable results.


gomesalexandre (approved):
lgtm 🫡 clean MPC-parity fix, reviewed the fund-safety-critical bits closely.

Adversarial read:
- Token amount encoding (amountToBytes -> TW.Cardano.TokenAmount.amount): amount.toString(16) on a bigint is arbitrary-precision big-endian, even-padded, Buffer.from(hex) - NO uint64 wrap. Cardano native tokens can exceed uint64, and the test covers exactly that (amount '18446744073709551616' = 2^64 round-trips to 9 minimal bytes). This is the safe usage of BigInt.toString(16), not the truncating kind.
- Deterministic MPC parity: UTXOs ordered (hash, index), assets (policyId, assetNameHex) lowercased, null asset_name -> '' , noncanonical quantity ("0042") -> canonical decimal via BigInt round-trip - byte-for-byte with the mainnet-tested iOS signer, so both peers produce identical body bytes and the pre-image hash matches.
- Value conservation: per-UTXO tokens attached off the keysign wire let WalletCore's planner reconcile input tokens into the change output - fixes the Ogmios 3123 "value not conserved" broadcast reject and the token-dropping bodies.
- getCardanoTxTtl decodes via cardanoCborEncoder (Map-preserving) so byte-string multiasset keys don't throw; inspection-only, never re-encoded, so it can't perturb the tx hash.

Receipts: CI green on the full check:ci gate (incl. the new cardano tests - TTL multiasset decode, the (hash,index)/(policyId,assetNameHex) ordering, the 2^64 amount round-trip, null/noncanonical normalization). Comment gate clear. Verified the amount-encoding + ordering by adversarial read since SDK isolated runs need build:all (CI already ran it).

PR #958, opened 2026-07-03
fix(sdk): polymod-check bch cashaddr in the address-format fund-safety gate

gomesalexandre:
## what

`isAddressValidForChain` — the `validate` namespace, documented in `addressValidation.ts` as *"format-validation checks suitable for **fund-safety gating before a tx is built**"* — validated **BitcoinCash** with a shape-only regex (`[qp][a-z0-9]{41}`): no cashaddr **polymod checksum**, and a charset that wrongly admits `b/i/o/1` (not in cashaddr's base32 alphabet).

A single-character typo in a BCH address has a valid shape but a broken checksum, so the gate green-lit an address the user never intended. The tx builder's own polymod check is a *separate, later* gate that callers using the validator standalone (address book, send preflight, agent-side validation) don't necessarily reach. This is the same gap the Go side recently hardened (`bchRule` polymod enforcement).

## how

The correct cashaddr polymod verifier already existed but was **local to `chains/utxo/tx.ts`**. Extract it into a shared `utils/cashaddr.ts` (single source of truth) exporting `isValidCashAddr` + the polymod primitives; use it in the address-format gate; and rewire `tx.ts` to import it (deleting the duplicate). The loose `classifyAddress` family heuristic **intentionally** keeps the shape regex — "looks like BCH" (classification) is a legitimately different concern from "is a valid BCH address" (the fund-safety gate).

## receipts (all local)

- New `cashaddr.test.ts` ✅ **5/5** — valid P2PKH + P2SH (with/without `bitcoincash:` prefix) pass; single-char typo, forbidden-charset (`b`), wrong length, and uppercase all rejected; `isAddressValidForChain('BitcoinCash', typo)` now returns `false`
- Existing `utxo-cashaddr` + `utxo-p2sh` tx tests ✅ **24/24** unchanged (proves the `tx.ts` DRY refactor is behaviour-preserving on the signing path)
- `chainValidation` + `normalizeChain` ✅ **140/140** (no cross-chain regression)
- `yarn workspace @vultisig/sdk typecheck` ✅ exit 0

Verified end-to-end on this host (the SDK toolchain runs locally). Canonical CashAddr vectors from the reference.cash / Bitcoin ABC spec.


NeOMakinG (approved):
## testing approach

Two-lane review (Claude primary; Codex best-effort — its run didn't return). For a checksum fund-safety gate the trap is a subtly-wrong polymod (false-accept a class of invalid addresses) or an inert gate, so I ran the polymod against canonical spec vectors + known-bad inputs, traced the gate's reachability, and mutation-checked it.

- Empirical polymod (node, the exact GEN + algorithm): canonical P2PKH / P2SH / bare-no-prefix / a real-world BCH address → **pass**; single-char typo / forbidden-charset `b` / testnet `bchtest:` prefix → **reject**. So the checksum matches the reference.cash / Bitcoin ABC spec.
- `cashaddr.test.ts` 5/5 + `utxo-cashaddr`/`utxo-p2sh` 24/24 = 29/29; CI (Unit, Integration, Lint+TypeCheck, CodeRabbit) all green.

## the fix is correct and load-bearing

Replacing the shape-only `reBCH` regex with `isValidCashAddr` (strip optional `bitcoincash:`, len-42, base32 charset, polymod checksum) is exactly right — the regex admitted `b/i/o/1` and a typo with a broken checksum, and this now fails closed. Verified against the invariants:

- **Polymod correctness** — confirmed empirically (above); the `BigInt` note is right (the >32-bit generators would truncate under i32 bit-ops and false-accept).
- **Prefix** — both `bitcoincash:`-prefixed and bare payloads pass; a testnet/wrong-prefix or uppercase address fails the length/charset (correct for a mainnet gate).
- **Gate reached, not inert** — `isAddressValidForChain('BitcoinCash', …)` resolves the chain tag and hits `isValidCashAddr` (the test's valid→true / typo→false proves resolution; an unmatched tag would return `undefined` for both). Mutation-proven: reverting the gate to `re(reBCH)` turns the reject-typo test red. And the tx builder is a defense-in-depth last line — it runs the *same* shared `verifyCashAddrChecksum`, which is the point of the DRY extraction.
- **DRY refactor** — the extracted polymod/charset are byte-identical to `tx.ts`'s originals, and the 24 signing-path tests are unchanged, so behavior is preserved on the send path.
- **No false-reject / base58** — valid CashAddr passes; the gate stays CashAddr-only, exactly as the old `reBCH` was, so legacy base58 BCH isn't newly rejected (it was already out of scope, and BCH deprecated base58 in 2018).

## notes (non-blocking, pre-existing)

- `length === 42` targets standard 20-byte-hash P2PKH/P2SH; a non-standard-hash CashAddr (other hash sizes in the spec) would be false-rejected — but those aren't used for normal BCH sends, and the old regex was also 42-symbol-only, so no regression.
- The gate is CashAddr-only while the tx builder's base58 path could accept a legacy base58 BCH address — a pre-existing gate/builder asymmetry on a deprecated format, not introduced here.

## risk

Low. Strictly tightens a fund-safety gate to reject mistyped BCH addresses, with an empirically-correct checksum, a mutation-proven load-bearing gate, and a byte-identical behavior-preserving refactor. No `@ts-expect-error`/`as any`.

## verdict

**APPROVE.** The address-format gate now polymod-checks BCH CashAddr, matching the tx builder (and the Go `bchRule`), so a single-character typo is rejected before a tx is built.

---

PR #959, opened 2026-07-03
fix(ripple): put the xrp account reserve on the amount, not the burned fee

gomesalexandre:
## what — live-confirmed XRP fund loss

`getRippleChainSpecific` added the base reserve (`reserve_base`, ~1 XRP) to `gas` whenever the destination account wasn't yet activated. `gas` becomes `TW.Ripple` `SigningInput.fee` — the XRPL transaction **Fee**, which the network **burns**; it never reaches the recipient. XRPL's reserve requirement is on the Payment **Amount** (the drops that create/fund the destination), **not** the Fee ([XRPL reserves spec](https://xrpl.org/docs/concepts/accounts/reserves)).

**Net effect: every send to a fresh/unactivated XRP address burned ~1 XRP for nothing**, on top of the actual send amount. Confirmed live against mainnet (read-only, no funds moved):

| destination | resulting `gas` (Fee) |
|---|---|
| unfunded (fresh) address | **1,000,020 drops** (≈ 1 XRP) |
| existing funded account | **20 drops** (≈ $0.00002) |

A ~**50,000×** fee inflation. And if the send amount was itself below the reserve, the tx also failed on-ledger (`tecNO_DST_INSUF_XRP`) **and** still burned the inflated fee.

## how

- `gas` is now the **network fee only** — the reserve rides on the Payment amount where XRPL wants it.
- The resolver already detects an unactivated destination (`destinationAccountResult`), so guard that `toAmount >= reserve_base` and throw a clear, actionable error otherwise (the send can't create the account regardless of fee), instead of broadcasting a doomed/wasteful tx.

## receipts (local)

- New `ripple.test.ts` ✅ **3/3** (via `.config/vitest.core.config.ts`): funded → fee only; unfunded + amount ≥ reserve → fee only (**not** fee + reserve, the old bug); unfunded + amount < reserve → rejects
- Sibling `swap/build.transfer.test.ts` ✅ **6/6** unchanged
- Bug independently confirmed against XRPL mainnet during the audit (the resolver is a pure read-only RPC computation — no vault/signer needed to reproduce).

**Severity: high, real fund loss** (not bounded by on-device signing — the inflated Fee is exactly what gets signed and burned). Worth expediting.


NeOMakinG (approved):
## testing approach
Two-lane (Claude load-bearing + Codex gpt-5.4, alive). Worktree `/tmp/wt-sdk-959` @ `7c31232c`. XRP-reserve fund lens: traced `gas` → `TW.Ripple SigningInput.fee` (burned), verified reserve units, the destination-activation guard, and the sender-vs-destination reserve distinction. `ripple.test.ts` 3/3 (vitest.core), mutation-proven. CI fully green (Unit + Integration + Lint/Type + Summary).

## blocking
None.

## preferably-blocking
None.

## should-fix (follow-up — non-regressing; a safe fix touches the generic send path)
The new unfunded-destination reserve guard (`resolvers/ripple.ts:43`) validates `keysignPayload.toAmount`, but in `buildSendKeysignPayload` (`send/build.ts:79-97`) `getChainSpecific` (the guard) runs **before** `refineKeysignAmount` (`refine/amount.ts:34`), which then clamps `toAmount = min(toAmount, balance − fee)`. So a **max-send / over-spendable send to an unfunded XRP address** can pass the guard on the requested amount, then get refined below `reserve_base` and sign a doomed `tecNO_DST_INSUF_XRP` (fee ~20 drops burned, send fails). Exact trigger: dest unfunded, requested `toAmount ≥ reserve_base`, and `balance − fee < reserve_base`.
**Important — this is not a regression:** pre-PR the same path was strictly worse (old refined amount was `balance − networkFee − reserve`, so it needed `balance ≥ ~2 XRP` to succeed and burned ~1 XRP on failure; post-PR it needs only `~1 XRP + fee` and burns ~20 drops). Funds are never lost (the send fails, funds stay). Recommend a follow-up: re-validate the reserve floor against the **final refined amount** — either make `refineKeysignAmount` reserve-aware for Ripple, or add a post-refine unfunded-XRP check in `build.ts`. I did not fold this into this PR because a correct fix touches the chain-generic refine/build ordering (`getFeeAmount` in refine reads the fee `getChainSpecific` already set, so the two can't simply be reordered) and deserves its own test matrix — and this PR is an urgent, strictly-better fund fix that shouldn't wait on it. (Codex corroborated.)

## suggestion
The regression tests are resolver-local (Codex): they prove `getRippleChainSpecific` returns `gas = networkFee` and rejects `toAmount < reserve`, but never exercise the real `buildSendKeysignPayload`/`refineKeysignAmount` path where `toAmount` is mutated after the guard. When the follow-up lands, add a build-path test for an unfunded dest with spendable in `[reserve_base, reserve_base + fee)` asserting it rejects rather than signing below reserve.

## q
None.

## risk
Low, and a large net fund-safety improvement. The change is a display/fee-accounting fix in a pure read-only RPC resolver; no signing/vault logic.
- **The fund fix is correct (lens 2):** `gas` is now `networkFee` only (`maxBigInt(computedFee, minProtocolFee)`), i.e. the actual burned XRPL Fee — the ~1-XRP `reserve_base` no longer rides on the burned Fee (it burned ~1 XRP per fresh-address send, a ~50,000× fee inflation, live-confirmed). Mutation-proven: re-adding the reserve to `gas` reddens the "gas is networkFee only" test.
- **Reserve now on the Amount, correctly (lens 1/3):** for an unfunded destination the guard requires `toAmount ≥ reserve_base` (else a clear throw) — this is the **destination-activation** reserve, and a fresh account has zero owner objects, so `reserve_base` (base only, no owner increment) is exactly right there. The sender's own base+owner reserve / send-all-leaves-reserve is app-side (no XRP sender-reserve/max-send logic exists in this SDK package — grep-confirmed), and is unaffected by this resolver.
- **Units correct (lens 4):** `reserve_base`/`base_fee` come from the XRPL `server_state` command (`network/info.ts`), which returns **drops** (not `*_xrp`); `toAmount` is drops (decimals 6). Both drops → the `toAmount < reserve_base` compare is unit-consistent (the PR's live evidence — old fee 1,000,020 drops = 20 + 1,000,000 — confirms `reserve_base` = 1e6 drops).
- **Guard boundary correct (lens 6):** `< reserve_base` throws; exactly `reserve_base` passes — a new account funded with exactly the base reserve is valid.
- **No false-block / no caller crash (lens 5):** funded destinations are unaffected (no minimum). The throw propagates as a rejected `buildSendKeysignPayload` (a clear actionable message) instead of broadcasting a doomed/wasteful tx — strictly better than the old silent `tecNO_DST_INSUF_XRP` + inflated burn; `formatGasInfo` reads `gas` only after resolution, so it's not impacted.

## verdict
**APPROVE** (@ `7c31232c`). Closes a real, live-confirmed XRP fund loss (reserve burned as Fee on every fresh-address send); the fee is now the actual network fee and the reserve is enforced on the Amount with correct drops units and the correct base-only reserve for account activation. Mutation-proven; CI green. The one residual (guard runs before the max-send refine clamp) is a non-regressing should-fix best done as a follow-up in the generic send path.

<sub>reviewed by the gomes standing loop (Claude + Codex gpt-5.4). ships in 2.16.0.</sub>

PR #960, opened 2026-07-03
fix(evm): batch 1inch token-metadata lookup to stop http 414 on token-rich wallets

gomesalexandre:
## what — the "unable to retrieve USDC balance" bug

`findEvmCoins` joined **every** held-token address into a single `GET /token/v1.2/{chainId}/custom?addresses=a,b,c,…`. On a wallet with many ERC-20s — **Ethereum especially**, the largest token universe — the query string overflows the `api.vultisig.com` proxy's URI length limit → **HTTP 414 URI Too Long**. That's a non-`NoDataError`, so it was **rethrown and killed the entire chain's token discovery** — even a held USDC balance came back empty, which surfaces in the app as *"I'm sorry, I was unable to retrieve your USDC balance."*

**Reproduced live** (read-only): a token-rich wallet returned `tokens: [], source: sdk-findCoins-error, error: HTTP 414 URI Too Long` on Ethereum, while smaller-token chains (Arbitrum/Base/Polygon/Avalanche) worked.

## how

The `/custom` lookup is only a **metadata optimization** — `getDiscoveredEvmCoin` already falls back to per-token on-chain metadata (`getEvmTokenMetadata`) when a token is missing from the batch result. So:

1. **Chunk** the addresses into batches of 50 (`toBatches`) — each URI stays ~2 KB, far under any gateway limit.
2. Make a batch failure **non-fatal**: skip that batch's metadata and let the per-token fallback fill the gap, instead of aborting the whole discovery.

## receipts (local)

- **Live before/after**: the exact wallet that previously 414'd on Ethereum now discovers **263 tokens, USDC present** (`0xa0b869…eb48`), in ~4.5s. Before: 414, `tokens: []`.
- `index.test.ts` ✅ **5/5** — 2 new (the 414-regression / non-fatal case + a 60-token chunking case asserting 2 batches and no single URI carrying all addresses) + the 3 existing tests unchanged.

**Impact: high** — restores ERC-20 (USDC etc.) balance/discovery for any active Ethereum wallet. Bug found by driving the reported sim failure end-to-end.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EVM token discovery by batching 1inch token-metadata requests to reduce request-length issues (e.g., HTTP 414).
  * Adjusted metadata handling so batch failures don’t abort discovery; missing token details are filled via fallback lookups.
* **Tests**
  * Added regression tests for large-token wallets and for handling 1inch “URI Too Long” responses without stopping discovery.


NeOMakinG (approved):
## testing approach

Two-lane (Claude primary; Codex best-effort — no usable return). Reliability lens on a batching fix: the risks are a batch still too big for an extreme wallet, a token dropped at a batch boundary, one bad batch failing all, or a failed batch injecting wrong decimals. I read `toBatches`, traced the merge + partial-fail path, and mutation-checked. This closes the exact `tokenInfoResult`-throw I flagged as the "#960 domain" q on the sibling #969.

- `evm/index.test` 5/5 (2 new: the 414-non-fatal case + a 60-token chunking case), no regression; mutation-proven (below).

## the fix is correct

Chunking the `/custom` metadata lookup into batches of 50 and making each batch non-fatal is exactly right — the lookup is only a metadata optimization, and `getDiscoveredEvmCoin` already falls back to per-token on-chain metadata. Verified against the invariants:

- **Bounded under the 414 threshold** — 50 addresses × ~43 bytes ≈ 2.2 KB per URI, and the batch size is fixed regardless of total token count, so even a 5000-token wallet can't overflow (100 batches, each ~2.2 KB). `toBatches` is an exact `ceil(N/50)` partition — no drop, no overlap.
- **Merge is correct** — `Object.assign` accumulates every batch into one address-keyed map, and `getDiscoveredEvmCoin` looks up `tokenInfoData[addr] ?? [addr.toLowerCase()]`, so a 1inch case-mismatch can't drop a token. The chunking test confirms 60 tokens → 2 batches → all 60 discovered.
- **Partial failure is non-fatal** — the loop wraps each request in `attempt` (non-throwing) and continues, so a 414/429/5xx on one batch skips only that batch and the per-token on-chain fallback fills the gap. This is the batch-level sibling of #969's per-token fix.
- **No wrong decimals** — a failed batch degrades to the on-chain `decimals()` (trusted); batch metadata is only used on success. Cross-refs #969.
- **`/balances` is 414-immune** — that first call has a single address in the path, not a many-address query, so leaving its throw-on-error unchanged is fine; the fix correctly targets only the `/custom` call.
- **Mutation-proven** — making a batch error fatal reds the 414-regression test, and removing the chunking (huge batch size) reds the chunking test. I re-ran the suite.

## q (non-blocking)

The metadata batches run serially (`await` in the `for…of`), so a very token-rich wallet makes `ceil(N/50)` sequential round-trips (~4.5s for 263 tokens live). It's bounded and acceptable, and serial avoids a parallel rate-limit burst, but a bounded `Promise.all` could cut latency if it matters.

## version

This is a patch changeset and the latest published SDK is 2.16.0, so it lands in the **next** release (post-2.16.0) — the 2.16.0 bumps in abts/mcp-ts do **not** include this 414 fix, so a follow-up SDK bump is needed to ship it to prod.

## risk

Low. Turns a single oversized request into bounded, non-fatal batches; a failed batch degrades to per-token on-chain metadata; no correctness/decimals impact.

## verdict

**APPROVE.** Token-rich wallets no longer 414 the whole chain's discovery — the metadata lookup is chunked well under any URI limit, merged correctly, and each batch is non-fatal, mutation-proven on both axes.

---

NeOMakinG (approved):
## testing approach — post-approval BIG-REBASE delta re-review

Two-lane (Claude load-bearing + Codex gpt-5.4). Prior verdict: **APPROVE (clean)** at `cd4a66cc`. The delta `cd4a66cc..f94028ad` is large (40 files, +1037) — the 2.16.0 release cut plus sibling PRs merged into main and rebased in. I isolated #960's own contribution and proved every other file is rebased-in main (SDK version-bump discipline applied to yarn.lock + signing files).

- **(1) #960's OWN change is unchanged** — the PR diff-vs-base (`gh pr diff`) is still the approved fix: `toBatches(nonZeroBalanceTokenAddresses, ONE_INCH_TOKENS_PER_REQUEST=50)` → per-batch `/token/.../custom?addresses=${batch.join(',')}` → `Object.assign` merge → non-fatal skip on any batch error. Same batch-to-avoid-414 + per-batch-non-fatal logic, byte-for-byte with what was approved.
- **(2) SIGNING-PATH files are byte-identical to origin/main — no smuggled logic.** `git diff f94028ad..origin/main` is EMPTY for `chains/utxo/tx.ts`, `react-native/chains/cosmos/tx.ts` (+ `cosmos/index.ts`), `vault/VaultBase.ts`, `tools/balance/otherBalance.ts`, the sui keysign resolver, and the thorchain-deposit-asset test. Codex independently confirmed the blob hashes match main (utxo tx `909a9138`, rn cosmos tx `819a6690`, VaultBase `82d9bb8f`, thorchain test `835af801`). These are rebased-in sibling PRs (utxo OP_RETURN, cosmos pagination, maya/rune deposit), not #960 contributions.
- **(3) yarn.lock — no surprise dep, guilty-until-proven-safe cleared.** `yarn.lock` at head is byte-identical to origin/main. Its only delta-vs-approved-SHA is `@vultisig/rujira`'s peer floor `"@vultisig/sdk": ">=2.15.0" → ">=2.16.0"` — release churn from the 2.16.0 cut, NOT a dependency add/bump. No `@trustwallet/wallet-core` / WalletCore-native / other fund-relevant or native dep changed on this branch beyond what's already on main (Codex-confirmed). No new transitive dep.
- **(4) changeset / caret clean.** #960's own `.changeset/fix-1inch-414-chunk.md` = `"@vultisig/sdk": patch` (intact, correct semver). The package.json/CHANGELOG version bumps (sdk 2.15.0→2.16.0, chain 48→49, rujira peer floor) are the changeset-version release process rebased from main — all manifests byte-identical to origin/main, no new `^` caret-float on any fund-relevant dep.
- **(5) no regression** — #960's fix logic (and its 5/5 tests) is unchanged; every other delta file matches main.

CI all green (Unit / Integration / Lint+TypeCheck / Test Summary / CodeRabbit). Codex returned all-INFO, no findings on any of the four adversarial axes (signing-path, yarn.lock dep, caret-float, thorchain test).

## verdict

APPROVE (re-approve at `f94028ad`). **Benign rebase** — the largest in this batch, but fully accounted for: it's the 2.16.0 release cut + sibling pagination/resilience/maya-deposit PRs merged into main, every one byte-identical to origin/main. #960's own 1inch batch-metadata fix is unchanged since approval, no signing-path change was smuggled in, and the yarn.lock delta is a single internal-workspace peer-floor bump with no dependency add/bump/caret-float. Benign-rebase, not a real change.

_Reviewed by NeOMakinG (Claude + Codex two-lane). Post-approval big-rebase delta re-review._

NeOMakinG (approved):
## testing approach — post-approval delta re-review (REAL code change)

Two-lane (Claude load-bearing + Codex gpt-5.4). Prior verdicts: APPROVE @cd4a66cc (r1) + APPROVE @d2b979c2 (r2 rebase). This delta `d2b979c2..17075103` is NOT a pure rebase — a genuine +5-line code change + a new changeset + a +30 test. Reviewed the real code delta.

**(1) what changed — precise:** the delta adds `EvmChain.Zksync` to `oneInchSupportedChains` in `findEvmCoins` (`packages/core/chain/coin/find/resolvers/evm/index.ts:85`). zkSync was absent from that allowlist, so `if (!oneInchSupportedChains.includes(chain)) return []` short-circuited — token discovery silently returned `[]` (a false "you have no tokens") even though the api.vultisig.com 1inch proxy serves zkSync Era (chain 324). It is NOT a batching refinement and NOT related to a wrong-chain fold-in — a focused allowlist add that enables the existing 1inch discovery path on zkSync.

**(2/4) fund-safety — correct chain, no mis-attribution:** the resolver derives the endpoint chainId via the SHARED `oneInchChainId = hexToNumber(getEvmChainId(chain))` — the same authoritative mapping every EVM chain uses, no special-casing in this change. `getEvmChainId(EvmChain.Zksync)` resolves to viem's `zksync` chain object (`chainInfo.ts:71`), whose `.id` is **324 (0x144)** — verified, and Codex independently confirmed. So `/balance/v1.2/324/…` + `/token/v1.2/324/custom` pull zkSync Era's own token contracts, not another chain's. The new test **PINS this**: it asserts the actual `/balance/v1.2/324/` and `/token/v1.2/324/custom` URLs (its comment notes that asserting only "a 1inch call happened" would stay green under a future wrong-chain remap) — a real mis-routing regression guard. The discovered token's `decimals: 6` is asserted too.

**(2) decimals / (3) batching preserved:** zkSync flows through the identical `findEvmCoins` path — 1inch `/custom` batch metadata (trusted) with the per-token on-chain `getEvmTokenMetadata` fallback (trusted `decimals()`) on a failed batch, and the `toBatches(…, 50)` + non-fatal-per-batch-skip 414-avoidance is untouched. No special-casing that could drop a token or default a decimals. Codex confirmed the same metadata + partial-failure path.

**(2 drop-token):** additive only — adding zkSync to the allowlist only ENABLES discovery for zkSync (was `[]`/short-circuited before); no effect on any other chain (it's a membership check). This FIXES a hide-real-tokens bug, it doesn't introduce one.

**(5) test:** the +30 test covers the new behavior (discovers a USDC on zkSync instead of short-circuiting), pins the chainId-324 URLs, and asserts a network call now happens (regression: previously returned `[]` with no call).

**(6) changeset:** `.changeset/fix-zksync-1inch-discovery.md` = `"@vultisig/sdk": patch` — correct semver (a fix), no caret-float.

## q
The changeset bumps `@vultisig/sdk` but the code lives in `@vultisig/core-chain` (`packages/core/chain/…`) — so a downstream consumer importing `@vultisig/core-chain` DIRECTLY wouldn't get this fix unless the release cascade bumps core-chain too (Codex MEDIUM). This is consistent with #960's own original changeset (`fix-1inch-414-chunk.md`, also `@vultisig/sdk: patch` for the same core-chain file) and the r2 release cut did cascade a `@vultisig/core-chain` version bump (48→49), so it's likely handled by the linked/dependency-cascade config — but worth confirming core-chain gets versioned so a direct consumer isn't left on the buggy discovery. Non-blocking; matches the PR's established convention.

## risk
Low. A focused allowlist add that turns on the existing, already-reviewed 1inch discovery path for zkSync: correct chainId (324, viem-derived + test-pinned), same trusted decimals path, batching/non-fatal preserved, additive (fixes a false-empty discovery, no drop/mis-attribution on any chain). CI: Integration + CodeRabbit pass; Lint+TypeCheck + Unit still running (the tsc will confirm the `EvmChain.Zksync` enum member, which exists in `Chain.ts`).

## verdict
APPROVE (re-approve at `17075103`). **The real code change: adding `EvmChain.Zksync` to `oneInchSupportedChains` so zkSync ERC-20s are discovered (it was silently returning `[]`). Fund-safe: YES** — the endpoint chainId is the shared viem-derived 324 (test-pins the `/324/` URLs, so no mis-routing to a wrong chain's token list/decimals), zkSync uses the same trusted batch-metadata + on-chain-decimals-fallback path, the 414 batching + non-fatal-partial-failure is preserved, and it's purely additive (no dropped token, no effect on other chains). One non-blocking q: the changeset bumps `@vultisig/sdk` while the code is in `@vultisig/core-chain` — confirm the release cascade versions core-chain (matches #960's existing changeset pattern).

_Reviewed by NeOMakinG (Claude + Codex two-lane). Post-approval real-code delta re-review._

PR #961, opened 2026-07-03
fix(evm): enable 1inch token discovery on zksync era

gomesalexandre:
## what

`findEvmCoins` gates the 1inch balance/metadata lookup behind `oneInchSupportedChains`, which **omitted zkSync Era** (chainId 324) — even though 1inch (via the `api.vultisig.com` proxy) serves it. Both `/balance/v1.2/324/...` and `/token/v1.2/324/custom` return **HTTP 200** (verified live 2026-07-03). So `discover_tokens` on Zksync short-circuited to `[]` **with no network call**, surfacing to the user as a false *"you have no tokens"* on a chain where they may hold USDC etc.

agent-backend-ts's `unchained-account.ts` already documented this: *"Zksync (324) — YES, in 1inch supported list"* — this closes that known gap.

## how

Add `EvmChain.Zksync` to `oneInchSupportedChains` + a regression test asserting Zksync now runs the 1inch path instead of returning `[]` without a query.

## receipts

- Live: `GET api.vultisig.com/1inch/balance/v1.2/324/balances/…` → 200; `/1inch/token/v1.2/324/custom?addresses=…` → 200
- `index.test.ts` ✅ **4/4** (1 new Zksync case + 3 existing)

Companion to #960 (the 1inch 414 chunking fix) — same file, different concern (chain coverage vs URI length). The other non-1inch chains (Blast/Mantle/CronosChain/Sei/Hyperliquid) are genuinely unsupported by 1inch and correctly stay out of the list.

PR #962, opened 2026-07-03
fix(solana): don't let one token-program query failure hide the other's spl holdings

gomesalexandre:
## what

`getSplAccounts` queries the **SPL Token** and **Token-2022** programs and combined them with `Promise.all` — so a transient RPC failure on **either** program's `getParsedTokenAccountsByOwner` rejected the whole call. A user with normal SPL tokens would see **nothing** if only the token-2022 query hiccuped (surfaces as "unable to retrieve your Solana tokens").

## how

Switch to `Promise.allSettled`: return whatever succeeded, and only throw when **both** program queries fail (so a genuine RPC outage still surfaces an error rather than a false-empty "you have no tokens"). Behaviour is unchanged when both succeed.

This is the "add fallbacks" resilience gap. Note it does **not** fix the separate, consistent 520 that whale-scale accounts hit on `getTokenAccountsByOwner` — that's an upstream RPC scan limit (beaded `agent-backend-ts-0198`).

## receipts

- `getSplAccounts.test.ts` ✅ **3/3** — both-succeed merges both programs' accounts; one-fails returns the survivor (previously threw → empty); both-fail throws.

Part of the balance-read reliability sweep (companion to #960 / #961).


NeOMakinG (approved):
**APPROVE** (fix-and-approve). The core fix is correct — `getSplAccounts` now uses `Promise.allSettled`, so one token-program's RPC failure no longer hides the other program's holdings, and a genuine total outage still surfaces an error rather than a false-empty. I verified the send path is unaffected, landed the missing RN/Hermes-override half of the fix, and flag one fail-safe residual.

## testing approach
- **Independent queries, survivor surfaces (ask 1):** `allSettled` → filter fulfilled → `flatMap(r => r.value.value)`; throws only when `fulfilled.length === 0` (re-throwing the first rejection). Mutation-proven: reverting to `Promise.all` turns the "surviving program" test RED.
- **Send/input-selection is NOT affected (ask 3):** the SPL send path uses `getSplAssociatedAccount` (the specific sender ATA) in `chainSpecific/resolvers/solana/index.ts:43`, **not** `getSplAccounts`. `getSplAccounts` feeds only the balance-display resolver, the coin-discovery resolver, and the RN override — so a partial failure never mis-selects inputs or signs a wrong amount.
- **No double-count (ask 4):** a mint's accounts are owned by exactly one token program, so concatenating the two programs' results can't sum a token twice (and the flatMap shape is unchanged from before).
- **1-page / both-succeed unchanged (ask 6):** both-succeed merges both programs; both-fail throws — 3/3 green, mutation-proven.
- Two-lane: Codex flagged the RN-override gap and the false-zero preflight path (both verified below).

## fix landed (fix-and-approve, pushed to `fix_spl_program_resilience` @28661fad)
**The RN/Hermes override still used `Promise.all` — so the fix didn't reach the mobile app.** `packages/sdk/src/platforms/react-native/overrides/getSplAccounts.ts` is a hand-mirrored copy (lazy-imports `@solana/web3.js` to keep Hermes loadable) and it kept the old `Promise.all`, so on RN — the primary platform — one token-program failure still hid the other's holdings, leaving the PR's stated goal unmet there. Mirrored the exact `allSettled` logic into the override + added a lockstep unit test (mutation-proven). `tsc` + `eslint` clean.

## should-fix
- **Partial failure becomes a false `$0` for the affected program's tokens (fail-safe, but the drained-funds illusion).** When classic-SPL succeeds and Token-2022 fails (or vice-versa), `getSplAccounts` returns the survivor's accounts with **no partial-failure signal**. `getSolanaCoinBalance` (`coin/balance/resolvers/solana.ts:23`) then does `find(mint === id)` → `BigInt(tokenAmount ?? 0)`, so a **Token-2022 token during a token-2022 outage reads `$0`** instead of the pre-fix error. That `$0` flows through `BalanceService.getBalance` into `getMaxSendAmount` / `send("max")` / `swap("max")`. **It fails safe** — a max-send off a `$0` balance sends nothing / "insufficient", the actual send uses the ATA path (so a fixed-amount send of that token still works), and it's a net improvement over the pre-fix "one failure hides everything." But it is exactly the "distinguish genuinely-zero from query-failed" case: recommend `getSplAccounts` surface a partial-failure signal so the **balance/max-send certainty path** can fail closed (throw "couldn't load") when the requested mint isn't found *and* a program query failed, while **discovery** keeps showing the survivors. (Applies to both core and the now-mirrored RN override.)

## risk
Low-moderate. The core + RN fix is a clear availability improvement and the send/input-selection path is untouched (no fund loss, no wrong/over-send). The residual is a fail-safe false-`$0` on the balance/max-send **preflight** for the failed program's tokens during a single-program outage — a display/preflight correctness gap, not a fund-safety one. Isolated to Solana SPL.

## verdict
**APPROVE.** One-failure-hides-both is fixed on both core and RN, mutation-proven, send path unaffected. The false-`$0`-on-partial (should-fix) is fail-safe and a net improvement over the prior behavior — worth a follow-up to distinguish partial from genuine-zero. Safe to ship in the 2.16.0 bump.

---

PR #964, opened 2026-07-03
fix(ton): fail closed on a null account-info result before it crashes keysign

gomesalexandre:
## what

The TON keysign chain-specific resolver (`getTonChainSpecific`) reads the wallet `seqno` via `getTonAccountInfo` — the one RPC call in that resolver **not** wrapped in `attempt()`. `getExtendedAddressInformation` is a toncenter-style endpoint whose `assertFetchResponse` only checks the HTTP status, not the `{ "ok": false, "result": null }` body shape a transient RPC failure returns as a **200**. That null `result` flowed straight back and crashed the resolver on `const { account_state } = undefined` — a cryptic `TypeError` that aborted the whole chain-specific step, i.e. a **broken send** on an RPC blip.

Note this is NOT the uninitialized-wallet case (I initially mis-scoped it as that): a genuinely uninitialized wallet — one that received funds but never sent — replies `ok:true` with an `uninited.accountState` result that just has no `seqno` field, so `account_state?.seqno ?? 0` already yields `0n` and the signing path attaches StateInit correctly. The real crash trigger is the `ok:false`/null-result RPC-failure shape.

## how

- `getTonAccountInfo` now throws a descriptive error when the body is missing or `ok:false` / `result` null, instead of returning `null`. Return type stays non-nullable, so the balance resolver that reads `.balance` off it is unaffected.
- The keysign resolver keeps its defensive `account_state?.seqno ?? 0` for the uninited-shape (no `seqno` field) case.
- Fail-closed is the right call here: on a real RPC failure you do NOT want to sign with a guessed `seqno` (an initialized wallet signed with `seqno 0` + StateInit gets rejected by the network anyway) — a clear thrown error beats a cryptic destructure crash.

## tests

`getTonAccountInfo.test.ts` covers initialized / uninited-no-seqno / `ok:false`-null / missing-body. `resolvers/ton/index.test.ts` covers the resolver's seqno-0-on-uninited and seqno-N-on-initialized behavior. 6/6 green.

## risk

Low. The `ok:false` path already crashed (cryptic `TypeError`); this converts it to a clear error and leaves every success path byte-identical. Isolated to TON.

## receipts

Unit only (no signing broadcast — SDK-internal resolver, no sim surface). Found via the SDK Sui/TON/Polkadot chain-correctness spike; independently corroborated against `toncenter.com/api/v2/getExtendedAddressInformation` for the uninited-account body shape.


packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.ts
```diff
@@ -17,8 +17,17 @@ export const getTonChainSpecific: GetChainSpecificResolver<'tonSpecific'> = asyn
...
+  // StateInit); getExtendedAddressInformation returns an `uninited.accountState`
+  // result with NO `seqno` field, so `account_state.seqno` on the raw result is
+  // undefined — optional-chain to seqno 0 (the signing path then correctly
+  // attaches StateInit for seqno === 0), matching the abts side's
+  // `?.account_state?.seqno ?? 0`. A genuine RPC failure (`{ ok:false }`) is now
+  // caught inside `getTonAccountInfo`, which throws a descriptive error rather
+  // than returning null and crashing this destructure.
   const { account_state } = await getTonAccountInfo(address)
-  const sequenceNumber = BigInt(account_state.seqno || 0)
+  const sequenceNumber = BigInt(account_state?.seqno ?? 0)
```
NeOMakinG, line 30: `q:` (non-blocking) the upstream guard only rejects `!ok`/null-`result`, so an `ok:true, result:{}` (no `account_state`) response defaults seqno to `0` here. Correct for a genuinely uninitialized wallet; the theoretical worry (Codex) is a *malformed* `ok:true` for an already-initialized wallet computing seqno `0`. Not a fund-safety hole: a correct RPC for an initialized wallet always carries `account_state.seqno=N`, and a seqno-`0`+StateInit message to an initialized wallet contract is rejected (seqno mismatch → exit 33) — no fund movement, no replay, just a failed broadcast. The stricter 'positively identify the uninit `@type`, else throw' alternative was reasonably avoided: toncenter's uninit body shape is ambiguous across versions, so a hardcoded positive check risks bricking a legit first-send (the exact case this PR protects) and would diverge from the abts-side `?.account_state?.seqno ?? 0` contract. Safe as-is — flagging the trust boundary + backstop.

NeOMakinG (approved):
**APPROVE** — clean. The fix correctly converts a cryptic null-destructure crash into a fail-closed throw *before* keysign, without bricking a legitimate uninitialized first-send. Fund-safety invariants verified end-to-end; the PR's tests are solid and mutation-proven. One non-blocking `q` on guard breadth (backstopped by the TON contract).

## testing approach
- **Fail-closed propagation traced end-to-end:** `getTonAccountInfo` throws on `!response?.ok || !response.result`; the keysign resolver calls it *unwrapped* (only the non-critical `getTonWalletState`/`getJettonWalletAddress` use `attempt()`), and the dispatcher `getChainSpecific` (chainSpecific/index.ts:47) does a bare `await resolver(input)` with no try/catch — so the throw aborts keysign before any signing. Not swallowed anywhere. ✓
- **Right condition (not bricked):** a genuinely uninitialized wallet returns `ok:true` with a result present (no `seqno`), so it passes the guard and the resolver's `account_state?.seqno ?? 0` yields `0n` → StateInit first-send works. Only a real `ok:false`/null-result RPC failure throws. ✓ Confirmed `assertFetchResponse` checks only `response.ok` (HTTP status), so the toncenter `{ok:false,result:null}`-as-200 shape genuinely slipped through pre-fix.
- **seqno safety (no double-send/replay):** an initialized wallet's correct RPC always carries `account_state.seqno=N` → signs `N`; it cannot collapse to `0`. Matches the abts-side contract (`?.account_state?.seqno ?? 0`, the abts#646 / a40w family). The `|| 0` → `?? 0` change is immaterial for a numeric seqno.
- **Other caller checked:** the balance resolver (`chain/coin/balance/resolvers/ton.ts:11`) reads `.balance` — it *already* crashed on a null result pre-fix (destructuring `null`), so the now-throwing behaviour is a strictly cleaner failure, not a regression; its `ok:true` happy path is byte-identical.
- **Tests: 6/6 green, mutation-proven** (ran on the PR HEAD in the canonical monorepo checkout): neutering the fail-closed throw → the two `getTonAccountInfo` throw tests go RED; reverting `account_state?.seqno` back to `account_state.seqno` → the resolver's uninitialized test reproduces the exact pre-fix `TypeError: Cannot read properties of undefined (reading 'seqno')`.
- CI green (CodeRabbit, Integration Tests, Lint+TypeCheck, Unit Tests). Two-lane: Codex gpt-5.4 (diff-only; its shell was misconfigured) raised the guard-breadth `q` below.

## q
- **Guard admits `ok:true` with no `account_state`** — see inline on `resolvers/ton/index.ts:30`. Codex's observation: the guard only rejects `!ok`/null-`result`, so an `ok:true, result:{}` (no `account_state`) response defaults the keysign seqno to `0`. For a genuinely uninitialized wallet that is correct; the theoretical worry is a *malformed* `ok:true` response for an already-initialized wallet computing seqno `0`. This is **not** a fund-safety hole: a correct RPC for an initialized wallet always returns `account_state.seqno=N`, and even a malformed one producing seqno `0`+StateInit is **rejected by the initialized wallet contract** (seqno mismatch → exit 33) — no fund movement and no replay (a replay would require reusing a valid seqno-`N` signature, which this never produces), only a failed broadcast (`broadcast/resolvers/ton.ts:23` also treats `duplicate msg_seqno` as benign). The stricter "positively identify the uninit variant, else throw" alternative was deliberately *not* taken because toncenter's uninitialized body shape is ambiguous across versions (`uninited.accountState` `@type` vs a bare result), so a hardcoded positive check risks bricking a legitimate first-send — the exact failure this PR guards against — and it would diverge from the abts-side contract. Safe as-is; documenting the backstop.

## risk
Low. Surgical, TON-isolated, additive fail-closed behaviour with every success path byte-identical, backed by mutation-proven tests. Residual is the documented `q` (guard breadth), which the TON contract's own seqno check backstops.

## verdict
**APPROVE.** Correct fail-closed conversion, no first-send brick, no double-send/replay path, solid tests. Safe to ship in the 2.16.0 bump.

---

PR #965, opened 2026-07-03
fix(sui): paginate getAllCoins so sends see every coin object

gomesalexandre:
## what

`getSuiChainSpecific` read only the first page of `client.getAllCoins()` — it destructured `{ data }` and dropped `hasNextPage`/`nextCursor`. `getAllCoins` paginates at ~50 objects/page, and Sui's object-per-coin model makes >50 coin objects realistic for an active wallet (dust, partial fills, repeated small incoming transfers, staking rewards). So the coin set was silently truncated at 50.

That set feeds **both** `gasCoins` (native SUI objects used for the `Pay`/`PaySui` gas payment) and `inputCoins` (the specific coinType being sent) in `signingInputs/resolvers/sui.ts`. A truncated page produces a **broken send**:

- `"insufficient balance"` despite adequate total holdings, or
- an empty `inputCoins` array for the token-transfer branch if none of the sent coinType's objects happen to land on page 1 (lexicographic ordering).

Meanwhile the balance-**display** path (`coin/balance/resolvers/sui.ts`) uses `getBalance()` — a server-side aggregate that is NOT truncated — so the user sees a correct total but can't spend it. Not fund loss (no wrong destination/amount signed), but a broken send for a plausible, not-even-exotic user population.

Same bug class + fix shape as the already-fixed Solana SPL pagination gap (sdk#962).

## how

Loop `getAllCoins({ owner, cursor })` following `nextCursor` until `hasNextPage === false`, accumulating every page before building `coins`. Bounded by the RPC's own `hasNextPage` termination. No protocol changes.

## tests

`resolvers/sui/index.test.ts`: a 3-page (50/50/7) response yields all 107 coins and threads each page's cursor into the next request; a single-page response terminates after one call. 2/2 green. Core `tsc` clean on the touched files.

## risk

Low. Pure read-side completeness fix on the tx-construction path; the happy path (≤50 coins) is one iteration, byte-identical to before. Isolated to Sui.

## receipts

Unit only (SDK-internal keysign resolver, no sim/broadcast surface). Live-confirmed the truncation itself against `sui-rpc.publicnode.com` (same endpoint the SDK uses): `suix_getAllCoins` returned exactly 50 items with `hasNextPage: true` for a busy address. Found via the SDK Sui/TON/Polkadot chain-correctness spike.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Sui coin loading so wallets with many coin objects are handled correctly during send preparation.
  * Prevented incomplete coin selection caused by only reading the first page of results.
  * Added a safety limit to stop repeated pagination from looping indefinitely if the response never completes.

* **Tests**
  * Added coverage for paginated coin retrieval, normal single-page completion, and error handling when pagination does not advance.


NeOMakinG (approved):
**APPROVE** (fix-and-approve). The core fix is correct: `getSuiChainSpecific` now follows the `getAllCoins` cursor to completion, so the full coin set (which feeds both gas and input selection) is gathered instead of truncated at the first ~50-object page. I verified the both-paths coverage and termination, and hardened the one gap (unbounded loop).

## testing approach
- **Paginates to completion + terminates (ask 1):** the `do/while` accumulates every page and exits when `hasNextPage` is false (`cursor = hasNextPage ? nextCursor : null`). A `hasNextPage:true` + `nextCursor:null` also exits. ✓
- **Fixes gas AND input selection (ask 2) — verified in the consumer.** `signingInputs/resolvers/sui.ts` reads `coins` from the `suicheSpecific` payload and derives **both** `gasCoins = coins.filter(c => c.coinType === SUI)` (line 52) and `inputCoins = coins.filter(c => c.coinType === coinType)` (line 60) from that one set — so paginating here fixes both. Token send → `Pay{ gas: gasCoins[0], inputCoins }`; native send → `PaySui{ inputCoins }`. WalletCore does the amount/gas selection over the now-complete set (unchanged), so gas-coin selection isn't broken — it just reliably has every SUI object now (ask 3). ✓
- **Balance-display path is unaffected** (`coin/balance/resolvers/sui.ts` uses the server-aggregate `getBalance`), consistent with the "shows correct total but can't spend it" symptom. No other unpaginated coin fetch on the send path (ask 4). ✓
- **No 1-page regression (ask 5):** a single page runs one iteration, byte-identical. ✓
- **Mutation-proven (ask 6):** reverting to a single-page read turns the 3-page test RED (50 ≠ 107); the PR's 2 tests pass.
- Two-lane: Codex independently flagged the same unbounded-loop gap and recommended a hard page cap + fail-closed — exactly the hardening I landed.

## fix landed (fix-and-approve, pushed to `fix_sui_getallcoins_pagination` @9eba69da)
**Bounded the cursor loop + fail-closed (ask 4).** The loop trusted the RPC's `hasNextPage` to terminate — a buggy/adversarial RPC returning `hasNextPage:true` with a **non-advancing** `nextCursor` (`cur1 → cur1 → …`) would spin forever and block keysign resolution. Added a 200-page cap (~10k coin objects, far beyond any real wallet) that **throws** on exceed — fail closed rather than hang or silently re-truncate + under-fund the send. Added a stuck-cursor test (asserts it throws and calls `getAllCoins` exactly the cap number of times, i.e. bounded); mutation-proven (silently capping instead of throwing turns it red). `tsc` + `eslint` clean on the touched files.

## risk
Low. Read-side completeness fix on the tx-construction path; the ≤50-coin happy path is one iteration, byte-identical. Not a fund-loss vector (no wrong destination/amount signed — a truncated set only *fails* a send), and the added cap only ever fires on a misbehaving RPC, failing loudly. Isolated to Sui.

## verdict
**APPROVE.** Full coin set now gathered for both gas and input selection, 1-page unchanged, loop bounded + fail-closed against a stuck cursor. Safe to ship in the 2.16.0 bump.

---

PR #966, opened 2026-07-03
fix(polkadot): reject an invalid destination instead of self-sending

gomesalexandre:
## what

`resolvePolkadotToAddress` (shared by `getPolkadotSigningInputs` and `getBittensorSigningInputs`) returned the **sender's own coin address** when `keysignPayload.toAddress` was falsy or failed `isValidAddress`, instead of throwing:

```ts
return shouldUseOriginalAddress ? toAddress : address // <- `address` is the SENDER
```

Both callers build plain amount+recipient transfers (`balances.transfer_keep_alive` / Bittensor transfer) — there is no legitimate empty-`toAddress` self-bond case on this path. If the fallback ever fired, the signed extrinsic's on-chain destination would silently diverge from the recipient shown on the pre-sign card the user approved (the card is keyed off `keysignPayload.toAddress`) — a **"sign something other than what was displayed"** defect, the exact class this org already guards against elsewhere (blind-signing / from-injection).

It was also the **only** place on the Polkadot/Bittensor/Sui/TON tx-construction surface that failed **open**. Every sibling fails loud: Sui's `if (keysignPayload.memo) throw`, TON's comment-length throw, `preparePolkadotAssetSend`'s `decodePolkadotAccountId` throw on EVM-hex/wrong-chain SS58.

## how

Throw a descriptive error when `toAddress` is missing/invalid instead of substituting the sender. Happy path (valid destination) is unchanged — returns `toAddress` exactly as before.

## tests

`resolvePolkadotToAddress.test.ts`: valid destination returns as-is; empty and invalid both throw and never yield the sender address. 3/3 green, core `tsc` clean.

## risk

Low. No fund loss either way (money stayed with the sender), but a failed keysign beats a silent self-send that mismatches the approved card. Upstream (app UI / mcp-ts builders) almost certainly already validates `toAddress`, so this is the SDK's last line of defense rather than a hot path. Isolated to Polkadot + Bittensor.

## receipts

Unit only (SDK-internal signing-inputs resolver, no sim/broadcast surface). Found via the SDK Sui/TON/Polkadot chain-correctness spike; the surrounding era/mortality/nonce/spec_version and `moduleIndex:10 methodIndex:3` (Asset Hub `transfer_keep_alive`) were audited in the same pass and are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Polkadot transfers now fail fast when the destination address is missing or invalid, instead of accidentally sending to the sender’s own address.
  * This helps ensure the approved transfer details match the transaction that gets signed and submitted.

* **Tests**
  * Updated and expanded coverage for Polkadot address handling to verify valid destinations work and invalid ones are rejected.


NeOMakinG (approved):
**APPROVE** (fix-and-approve). The core fix is correct and fund-safe: `resolvePolkadotToAddress` now throws on a missing/invalid destination instead of silently returning the sender's own address, so the signed extrinsic can no longer diverge from the approved pre-sign card. I verified the validator is genuinely strict, landed the CI-unblocking test fix the change surfaced, and added real-validator coverage.

## testing approach
- **Fails closed, end-to-end:** the throw is unguarded — both callers (`getPolkadotSigningInputs`, `getBittensorSigningInputs`) invoke `resolvePolkadotToAddress` directly (no try/catch), and the signing-inputs dispatcher does a bare call, so an invalid destination aborts signing-input construction before any signing. No caller swallows it. ✓
- **Validator is genuinely strict (ask 2/3) — verified against real WalletCore**, not just the (mocked) unit test: for `CoinType.polkadot`, `AnyAddress.isValid` returns **true** for a real SS58-0 Polkadot address, and **false** for a Kusama (prefix 2), a generic-substrate (prefix 42), an EVM hex, and empty. Bittensor's `isValidSS58(addr, coinType, 42)` accepts prefix 42 and rejects prefix 0/2. So a wrong-network or hex destination is rejected; a valid Polkadot/Bittensor destination is not false-rejected. (Codex independently reproduced the same matrix.)
- **Self-send default fully removed (ask 5):** the `address` (sender) destructure and the fallback are gone; grep found no other self-send/default-destination path on the Polkadot/Bittensor signing surface.
- **Mutation-proven:** the PR's resolver tests go RED when I revert the fix to the self-send fallback (2 RED); the added wrong-network test also goes RED on fix-revert.

## fixes landed (fix-and-approve, pushed to `fix_polkadot_selfsend_fallback`)
1. **@58affa65 — unblock the CI `Unit Tests` failure the fix surfaced.** `polkadot.test.ts`'s `TO_ADDRESS` (`15oF…6Srd`) was labelled "arbitrary valid" but is **not** checksum-valid (the real //Alice address ends `…6Sp5`), so pre-fix the resolver silently fell back to the sender (the exact self-send bug) and the 3 SCALE call-index tests passed by luck; post-fix they correctly throw. Swapped in the real checksum-valid `…6Sp5`; the pinned SCALE call-index offset (107, single `0a03`) is unchanged. (The cardano.test.ts failures seen locally are a pre-existing wasm/env artifact — green in CI, untouched by this PR.)
2. **@8e961d21 — close the coverage gap Codex flagged.** The new resolver unit test mocks `isValidAddress`, so it can't catch a regression in the production SS58 prefix validation. Added a `getPolkadotSigningInputs` test that feeds a checksum-valid but **wrong-chain Kusama (prefix 2)** address through the **real** WalletCore validator and asserts it throws — verified load-bearing (reverting the fix turns it red). Locally: resolver 3/3 + polkadot signing-inputs 4/4 green, core `tsc` clean on touched files.

## q
- The PR's `resolvePolkadotToAddress.test.ts` mocks `isValidAddress`, so on its own it doesn't exercise the real prefix validation — that's now covered by the added `polkadot.test.ts` wrong-network case (above). No action needed; noting the intentional split (unit vs production-validator coverage).

## risk
Low. Additive fail-loud guard; the happy path (valid destination) returns `toAddress` exactly as before, and a wrong/hex/empty destination now throws before signing rather than self-sending against the approved card. Isolated to Polkadot + Bittensor; the real validator was independently confirmed strict in the correct direction.

## verdict
**APPROVE.** Correct self-send removal, strict validation confirmed against real WalletCore, no false-reject, no residual self-send path; CI-unblocking fixture fix + real-validator coverage landed. Safe to ship in the 2.16.0 bump.

---

NeOMakinG (approved):
## testing approach — post-approval delta re-review

Two-lane (Claude load-bearing + Codex gpt-5.4). Prior verdict: **APPROVE** (fix-and-approve) at `022ee201`. Re-reviewing the delta `022ee201..1419f2c2` (4 commits). No prior findings were left open — the r1 approval was clean — so this is purely a delta-safety check on the fund-critical reject-invalid-destination fix.

Delta `--stat` = cardano `getCardanoAddressAssets` + sui `keysign/chainSpecific/resolvers/sui/index.ts` (+ their changesets) + a small `polkadot.test.ts` change. Verified each:

- **(1) fix logic byte-UNCHANGED** — `git diff 022ee201..1419f2c2 -- packages/core/mpc/keysign/utils/resolvePolkadotToAddress.ts` is **empty**. The reject-invalid-destination behavior is identical to what was approved: it still throws `Invalid ${chain} destination address; refusing to fall back to the sender's own address` on a missing/invalid `toAddress`, and returns `toAddress` unchanged on the happy path. Codex confirmed byte-identical. No weakening.
- **(2) the `polkadot.test.ts` delta is Prettier-only** — the single change is `expect(...).toThrow(/refusing to fall back/)` collapsed from 3 lines onto 1. Matcher scope + failure condition unchanged; not relaxed. (The larger `+35` test-add on the PR was part of the r1 approval, not this delta.)
- **(3) sui/cardano = rebase noise, NOT smuggled keysign logic** — `git diff 1419f2c2..origin/main` for the sui resolver and cardano client is **empty** (blobs identical to main). Codex traced the sui `chainSpecific/resolvers/sui/index.ts` to mainline commit `2791af8a` — `fix(sui): paginate getAllCoins so sends see every coin object (#965)`, a separately-merged PR that rebased in. It is not new unreviewed keysign logic added to #966.
- **(4) changeset / caret-bump clean** — #966's own changeset (`.changeset/fix-polkadot-selfsend-guard.md`) is `"@vultisig/sdk": patch` (correct for a fix). No `package.json`/lockfile edits in the delta → no caret-bump hazard. The cardano/sui changesets match main (their own PRs).
- **(5) no regression** — the fix file is byte-unchanged, so the approved reject-destination behavior is intact.

CI all green (Unit / Integration / Lint+TypeCheck / Test Summary / CodeRabbit).

## verdict

APPROVE (re-approve at `1419f2c2`). The delta is a **benign rebase** (cardano #967 + sui #965 pagination merged from main, identical to origin/main) plus a Prettier reformat of one polkadot test assertion. #966's fund-critical fix logic (`resolvePolkadotToAddress.ts` throwing instead of self-sending) is byte-unchanged since approval, no assertion was weakened, and no keysign logic was smuggled in. Benign-rebase, not a real change.

_Reviewed by NeOMakinG (Claude + Codex two-lane). Post-approval delta re-review._

PR #967, opened 2026-07-03
fix(cardano): paginate address_assets so nft-heavy wallets aren't truncated

gomesalexandre:
## what

`getCardanoAddressAssets` fired a single **unbounded** Koios `address_assets` query. Koios caps every response at **1000 rows** and paginates via `offset`/`limit`, so any wallet holding >1000 native assets was silently truncated. On Cardano every NFT is its own native asset, so an NFT-heavy wallet blows past 1000 easily.

The truncation hit **both** callers of this function:

- **discovery** (`coin/find/resolvers/cardano.ts`) dropped every token past row 1000 - this is the USDC-class "unable to retrieve balance" bug, on Cardano;
- **per-token balance** (`coin/balance/resolvers/cardano.ts`) `.find()` returned `0` for a token the user actually holds when it sat past row 1000.

## how

Page through offsets (`?limit=1000&offset=N`) until a page returns fewer than 1000 rows, mirroring the existing `getUtxoAddressInfo` blockchair pagination pattern in this same package. Happy path (<1000 assets) is a single call, byte-identical to before.

## tests

`getCardanoAddressAssets.test.ts`: a >1000 wallet (1000/1000/12) returns all 2012 assets and threads offsets 0 -> 1000 -> 2000; a short first page stops after one call; null `decimals` still normalizes to 0. 3/3 green, core `tsc` clean.

## risk

Low. Pure read-side completeness fix. The common case (wallet with <1000 assets) is unchanged (one request). Isolated to Cardano.

## receipts

Live-confirmed the Koios cap + pagination semantics against `api.koios.rest`: an unbounded `asset_list` query returns exactly **1000** rows, `?limit=5` returns 5, `?offset=1000` returns the next page. Unit-only otherwise (SDK-internal client, no sim surface). I did not obtain a live >1000-asset address to screenshot the end-to-end fix, but the truncation mechanism is confirmed and the fix mirrors an already-shipped pattern in this package. Found via the balance-reliability sweep continuing the USDC-discovery investigation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Cardano asset discovery and balance lookups for wallets with large numbers of tokens, preventing assets from being cut off after the first page of results.
  * Improved handling of token metadata so missing decimals now display as `0` instead of causing inconsistent values.

* **Tests**
  * Added coverage for multi-page asset fetching, short-page completion, and safety limits to ensure pagination behaves reliably.


NeOMakinG (approved):
## testing approach

Two-lane intended (Claude primary; Codex skipped this round to avoid the shell-quoting error). Fund-relevant discovery lens: the risks are still-truncated on an extreme wallet, an infinite loop on a stuck cursor, and whether the send path is covered. I traced the callers, verified the pagination, and mutation-checked.

- `getCardanoAddressAssets` 4/4 (3 existing + the cap test I added); mutation-proven; tsc clean.

## fund-relevance — this feeds the SEND path (not NFT-display only)

`getCardanoAddressAssets` backs two callers: the discovery resolver AND the per-token balance resolver (`balance/resolvers/cardano.ts` does `assets.find(...) ? BigInt(match.quantity) : 0n`). So before this fix, a token past Koios's 1000-row cap returned a silent `0n` balance — the user "has 0" of a token they hold, and can't send it. Paginating fixes the send-balance, so this is fund-relevant like #965, not display-only.

## the pagination is correct — I added the missing runaway cap

The offset walk (`?limit=1000&offset=N`, stop on a short page) correctly gathers every asset and threads offsets 0 → 1000 → 2000, mirroring `getUtxoAddressInfo`. The one gap: the loop was `for (let offset = 0; ; …)` with **no upper bound**, so an offset-ignoring or caching Koios that keeps returning a full 1000-row page would loop forever and grow the buffer unboundedly (hang / OOM). The sibling sui/cosmos fixes (#965/#968) both guard this, so I added a `MAX_PAGES = 100` cap (100k native assets — far past any real wallet, a safety valve not a truncation limit) in commit `4a4b2e14`. Mutation-proven: a mock Koios that always returns a full page stops at exactly 100 calls, and reverting the cap reds that test; the existing offset-thread / short-page / null-decimals tests are unchanged (4/4).

Other invariants:
- **Asset identity/decimals** — `policy_id` + `asset_name` are preserved per row and `null` decimals normalize to 0 (tested); pagination just gathers more rows.
- **Perf** — now bounded at `MAX_PAGES × 1000`; the common <1000-asset wallet is still a single call.

## q (non-blocking)

- A page failure *mid-walk* rejects the whole call (no partial-return / fallback). For a >1000-asset wallet a transient page-2 failure would now error where today it returned a truncated 1000 — but the balance resolver's old behavior for a past-1000 token was a silent, wrong `0n`, so an honest error isn't clearly a regression; I left it rather than guess the never-regress shape.
- Offset (vs cursor) pagination can skip/dup a row if the asset set shifts between fetches — inherent to offset paging, low-risk for a mostly-static read.
- The cited `getUtxoAddressInfo` is *also* an uncapped `for(;;)` — a pre-existing sibling gap worth a follow-up.

## version

Patch changeset; latest published SDK is 2.16.0, so this lands in the next release (post-2.16.0) — not in the 2.16.0 bumps.

## risk

Low. Read-side completeness fix; the common case is unchanged; the walk is now bounded and the send-balance no longer truncates.

## verdict

**APPROVE.** NFT-heavy Cardano wallets no longer truncate at 1000 assets on either discovery or the send-balance, the offset walk is now bounded against a stuck upstream, and it's mutation-proven.

---

🤖 Review (Claude; fix-and-approve). Generated with [Claude Code](https://claude.com/claude-code).

PR #968, opened 2026-07-03
fix(cosmos): paginate token discovery so ibc-heavy wallets aren't truncated

gomesalexandre:
## what

`findCosmosCoins` discovered denoms via cosmjs `StargateClient.getAllBalances`, which issues a **single unpaginated** `QueryAllBalances`. The node then applies its default page limit (**100** on most Cosmos SDK chains), so any wallet holding >100 denoms had every token past the first 100 silently dropped from discovery - the USDC-class "missing balance" bug, on cosmos.

This is realistic on the auto-discovery chains (THORChain, Terra, TerraClassic, **Osmosis**), especially Osmosis, where an active DeFi user accumulates well over 100 IBC tokens + LP shares. cosmjs's sibling reads (`getBalanceStaked`, `totalSupply`) paginate; `getAllBalances` is a gap cosmjs never closed (it doesn't accept a pagination key).

## how

Add `getAllCosmosBalances`, which walks the LCD `cosmos/bank/v1beta1/balances/{address}` endpoint following `pagination.next_key` to completion (bounded to 20 pages as a runaway guard). LCD is already the trusted cosmos data source in this package - staking reads use it as primary, the balance resolver as a fallback.

Wire it into `findCosmosCoins` **with a fall-through** to the existing cosmjs path on any LCD failure (`attempt(...)` -> `?? getAllBalances`), so the change can only improve on today's behavior, never regress it. This is the "add fallbacks" ask made literal.

## tests

`getAllCosmosBalances.test.ts` (4): multi-page `next_key` walk collects every denom + threads the cursor; single short page stops after one call; missing `balances` array tolerated; the 20-page cap stops a never-null `next_key`. The existing 14-test `findCosmosCoins` suite now exercises the cosmjs fall-through (LCD stubbed to reject) and still passes unchanged. 18/18 green, core `tsc` + eslint clean.

## risk

Low, by construction. The fallback guarantees the worst case equals today's behavior (first 100 denoms). Isolated to the 4 cosmos auto-discovery chains.

## receipts

Confirmed the mechanism at the code level: cosmjs `getAllBalances` -> `bank.allBalances` -> `QueryAllBalancesRequest.fromPartial({ address })` with no `pagination` field (`node_modules/@cosmjs/stargate/build/modules/bank/queries.js`), and the sibling paginated reads right beside it. I did not obtain a live >100-denom address to screenshot the end-to-end fix, but the fallback makes this safe to ship for verification. Found via the balance-reliability sweep continuing the USDC-discovery investigation.


NeOMakinG (approved):
## testing approach

Two-lane intended (Claude primary; the Codex invocation hit a shell-quoting error on my side, so this is Claude-only — the pagination logic is small and empirically checkable). Fund-relevant discovery lens: the risks are still-truncated on an extreme wallet, an infinite loop on a stuck cursor, a send path that stays unpaginated, or a false-empty. I read the cursor loop + the fall-through and mutation-checked.

- `getAllCosmosBalances` 4/4 + `findCosmosCoins` 14/14 = 18/18; mutation-proven (below).

## the fix is correct

cosmjs `getAllBalances` issues a single unpaginated query that the node caps at 100 denoms, so IBC-heavy wallets lost every token past 100. Walking the LCD `pagination.next_key` with a fall-through to cosmjs is exactly right. Verified against the invariants:

- **Follows the cursor to exhaustion and terminates** — the loop advances `nextKey` from `pagination.next_key` and returns on a falsy key, with `MAX_PAGES=20` as a runaway guard for a never-null cursor. Dropping the cursor-advance reds the multi-page test; the cap test pins the 20-page stop.
- **Feeds both display and send** — this is the discovery resolver, so a newly-visible denom appears in the coin list and becomes selectable; the per-coin send-balance is a separate targeted read (not `getAllBalances`), so it was never subject to the 100-denom truncation. Paginating discovery fixes both.
- **No false-empty on failure** — a page throw is caught by `attempt` and falls through to cosmjs's first-100 (today's behavior), never an empty list.
- **No wrong decimals** — only the denom is used (the LCD amount is discarded and balance is read per-coin), and decimals still come from the unchanged `getCosmosTokenMetadata`, so pagination just surfaces more denoms.
- **Bounded** — `MAX_PAGES=20 × PAGE_LIMIT=1000`, sequential, capped; no hang on a dust-heavy wallet.
- **Never regresses** — the cosmjs fall-through guarantees the worst case equals today's first-100. Nice design.

## q (non-blocking)

- The fall-through (`lcdBalances ?? cosmjs`) triggers only on an LCD *throw* — `??` doesn't fall back on a successful-empty `[]`. So a spurious `200 {balances:[]}` from the LCD would show no cosmos coins without falling back to cosmjs. That's defensible (the LCD is the trusted primary for cosmos reads, so an empty is authoritative, and a length-check would wrongly discard a genuinely-empty wallet), but worth being aware of.
- A page failure *mid-walk* discards the pages already gathered and falls back to cosmjs's 100, rather than returning the partial set (the never-regress floor holds, but the #969/#962 partial-failure family returns partial).
- The `~20k denom` headroom assumes the node honors `pagination.limit=1000`; if it caps at its own 100/page the effective bound is ~2000 — still far past any real wallet, just less than advertised.

## version

Patch changeset; the latest published SDK is 2.16.0, so this lands in the **next** release (post-2.16.0) — not in the 2.16.0 bumps.

## risk

Low, by construction — the cosmjs fall-through makes the worst case equal today's behavior, and the change only adds denoms to discovery (no decimals/send-path change).

## verdict

**APPROVE.** IBC-heavy cosmos wallets are no longer truncated at 100 denoms — discovery walks the LCD pages, terminates and is capped, feeds both display and send, and falls back to cosmjs so it can only improve on today's behavior. Mutation-proven.

---

🤖 Review (Claude; Codex lane errored). Generated with [Claude Code](https://claude.com/claude-code).

PR #969, opened 2026-07-03
fix(evm): one token's metadata failure no longer kills all discovery

gomesalexandre:
## what

`getDiscoveredEvmCoin` re-threw any non-`NoDataError` from the on-chain `getEvmTokenMetadata` fallback:

```ts
if ('error' in metadataResult) {
  if (metadataResult.error instanceof NoDataError) return undefined
  throw metadataResult.error   // <- rejects the whole Promise.all
}
```

That call runs inside the `Promise.all(...)` in `findEvmCoins`, so a **single token's transient RPC/metadata hiccup rejected the entire discovery** - every token on the chain (USDC included) vanished with an "unable to retrieve your balances" failure.

It's very reachable: whenever the 1inch metadata call returns no data (`tokenInfoData` empty), **every** held token falls to this on-chain metadata path, so one flaky read among N tokens takes down all N. Same "missing balance" class as the 1inch 414 (#960) but a **distinct trigger** - the per-token metadata fallback, not the batch 1inch call - so the two fixes are complementary and non-overlapping.

## how

Drop just the failing token (`return undefined`) and `console.warn`, instead of throwing - mirroring the resilience of the Solana (`attempt` per element) and cosmos (`.catch`) discovery resolvers. A token that transiently fails is silently missing from this refresh and reappears on the next one; that's a far smaller failure than nuking the whole balance list.

## tests

Rewrote the existing test that codified the reject-all behavior (`rejects.toThrow('rpc down')`) into `skips a token whose on-chain metadata lookup fails transiently, keeping the rest`: two held tokens both fall to the on-chain path, one throws, and discovery still returns the other. 3/3 in the file green, core `tsc` + eslint clean.

## risk

Low. Only changes the error path (previously a hard throw, now a skip). The happy path and NoDataError path are unchanged. Isolated to EVM discovery.

## receipts

Unit only (SDK-internal resolver). Found via a Promise.all-reject-all sweep across the SDK balance/discovery resolvers (the same class as the Solana `getSplAccounts` #962 fix). Verified-clean siblings in that sweep: `getEvmChainBalances` (viem multicall `allowFailure:true` + per-token try/catch), `findSolanaCoins` (`attempt` per element), `findCosmosCoins` metadata map (`.catch` per element), XRP balance resolver.

## Summary by CodeRabbit

* **Bug Fixes**
  * EVM coin discovery is now more resilient when a single token’s metadata lookup fails.
  * Discovery continues for other tokens instead of stopping entirely, so you’ll still see available coins even if one token has an RPC or metadata issue.
  * Non-`NoData` lookup failures are now logged before that token is skipped.


NeOMakinG (approved):
## testing approach

Two-lane (Claude primary; Codex best-effort — no relevant return). Correctness/fund-safety lens on a discovery-resilience fix: the danger is a held token vanishing (can't see it to send it) or a bad `decimals` flowing into a displayed amount. I read `findEvmCoins` end-to-end, traced the decimals source, and mutation-checked.

- `evm/index.test` 3/3; CI (Unit, Integration, Lint+TypeCheck, CodeRabbit) all green.
- (Local note: the evm test needs `@vultisig/core-config` built — I copied the canonical dist into the worktree since config is unchanged by this PR.)

## the fix is correct

`getDiscoveredEvmCoin` runs inside `findEvmCoins`' `Promise.all`, and previously re-threw any non-`NoDataError` — so one token's transient RPC/metadata hiccup rejected the whole batch and every token on the chain vanished. Skipping the failing token (`return undefined` + `console.warn`) is right, and mirrors the Solana/cosmos per-element resilience. Verified against the invariants:

- **Graceful degradation / per-token isolation** — `without(await Promise.all(nonZero.map(getDiscoveredEvmCoin)), undefined)` filters the `undefined` failures, so every other token still discovers. Each element self-catches via `attempt`, so the `Promise.all` can no longer reject on a metadata error. No partial-write / shared state.
- **A dropped token doesn't misroute funds** — the field that failed *is* the metadata (decimals/ticker). Surfacing the token with a placeholder `decimals` would be the fund-*un*safe choice (wrong decimals → wrong displayed amount), so dropping it (it reappears next refresh, and already-known coins aren't affected by discovery) is the correct trade-off — strictly better than the old drop-all and safer than a degraded surface.
- **decimals is trusted** — it comes from the on-chain ERC20 `decimals()` call (the token contract is canonical for its own decimals), not attacker-injected metadata. A failed read now *skips* the token rather than yielding a NaN, so this change is safer on that axis, not riskier. The happy path and decimals handling are untouched.
- **Mutation-proven** — the rewritten test (`skips a token whose on-chain metadata lookup fails transiently, keeping the rest`) fails A's read and asserts B still returns; reverting the skip back to `throw` turns it red.

## note (non-blocking)

`findEvmCoins` still throws on a non-`NoDataError` from the 1inch *batch* calls (`balanceResult`, `tokenInfoResult`). Those are single calls covering all tokens — a total-outage class (the #960 domain), not the per-token-flaky class this PR targets — and the common empty case degrades to the now-resilient on-chain path, so there's no remaining one-token-kills-all hole here.

## risk

Low. Changes only the error path (hard throw → skip-with-warn); happy path and `NoDataError` path unchanged; isolated to EVM discovery.

## verdict

**APPROVE.** One token's transient metadata failure now degrades to a per-token skip instead of nuking the whole balance list, with decimals still sourced on-chain and the resilience mutation-proven.

---

PR #970, opened 2026-07-03
fix(tron): verify base58check checksum on the balance-read address decode

gomesalexandre:
## what

`tronAddressToAbiParam` (the TRC-20 **balance-read** path, `tools/balance/otherBalance.ts`) decoded the target address with plain `bs58.decode` and only checked length (25) + the `0x41` prefix byte - it **never verified the trailing 4-byte sha256d checksum**. A typo'd-but-still-base58-decodable Tron address (25 bytes, `0x41` prefix, corrupted checksum) therefore passed the guard and silently resolved to a **different** 20-byte address, reporting that account's balance instead of the intended one.

The **send** path (`abi/tron.ts` `tronBase58ToEvmHex`) already verifies the checksum, with a comment explaining exactly this footgun ("a typo'd-but-still-base58-decodable string can't silently re-checksum into a DIFFERENT valid address"). The balance-read path didn't.

## how

Swap `bs58.decode` → `bs58check.decode` (already a project dependency, already used the same way in `chains/tron/rpc.ts`), via the same ESM/CJS-safe resolve wrapper. `bs58check.decode` verifies + strips the checksum, returning the 21-byte payload, so a bad checksum now throws a clear error instead of silently misrouting the query.

## tests

`tests/unit/tronAddressToAbiParam.test.ts` (test owns its vectors - builds a valid `0x41`+20-byte address): valid address → correct left-padded 20-byte word; corrupted-checksum address → throws; valid-checksum-but-wrong-version-prefix → throws. 3/3 green, eslint clean.

## risk

Low. Read-only path (balance query), not the send path - so **not** a fund-loss vector, but a real "wrong data silently accepted" gap that contradicts the checksum discipline the codebase applies everywhere else. Behavior only changes for malformed input (now rejected instead of silently misrouted). Isolated to Tron balance reads.

## receipts

Found via a fuzz audit of the SDK's hand-rolled crypto/encoding that compared each surface against its battle-tested library across thousands of random inputs. That audit verified the **core fund-critical paths clean** (UTXO sighash/DER/BIP32 vs bitcoinjs-lib, Solana message serialization vs @solana/web3.js, cosmos protobuf vs cosmjs-types, the `abi/tron` encoder vs bs58, the RN tiny-secp256k1 shim vs the real lib) - byte-exact matches - so this checksum omission on the read path was the single confirmed defect. Unit-only (no live TRON RPC probe of a corrupted address needed - the checksum math is deterministic).


NeOMakinG (approved):
**APPROVE** — clean. The fix swaps the TRC-20 balance-read decode from plain `bs58.decode` (length + `0x41`-prefix only) to `bs58check.decode`, so a typo'd-but-decodable address now throws on checksum mismatch instead of silently querying a **different** account. Verified fund-safe, mutation-proven, and — per the explicit ask — **the send path does NOT need the same guard; it is already checksum-guarded everywhere.**

## testing approach
- **Fail-closed + version check (ask 1/2):** `bs58check.decode` verifies + strips the 4-byte sha256d checksum → 21-byte payload; a bad checksum throws (caught → clear `checksum mismatch` error). The post-decode `decoded.length !== 21 || decoded[0] !== 0x41` still rejects a valid-checksum wrong-version address (test covers a `0x00` Bitcoin-prefix payload). ✓
- **Send/keysign path is already guarded (ask 3) — the key cross-path check.** This PR touches only the read-only balance path; I confirmed every SDK-side TRON *send* decode already checksum-verifies:
  - Native TRX: `chains/tron/tx.ts` `tronAddressToBytes` uses `bs58check.decode` + 21-byte/`0x41` check on **both** `from` and `to` before emitting transfer bytes.
  - TRC-20: `abi/tron.ts` `tronBase58ToEvmHex` explicitly recomputes and compares the `sha256d` checksum + `0x41` prefix (and the TRC-20 prep routes contract/from/to through it).
  - Keysign resolver `getTronSigningInputs` passes the base58 `toAddress` to WalletCore's `TW.Tron.Proto` compiler, which decodes base58check (checksum-verified); the generic prepare-send uses WalletCore `AnyAddress.isValid`.
  So the balance-read `tronAddressToAbiParam` was the single remaining unchecked decode — now closed. **No follow-up bead needed for the send path.**
- **No false-reject (ask 4):** the happy-path test builds a genuine `0x41`+20-byte address via `bs58check.encode` and decodes it to the correct left-padded 20-byte word; real mainnet `T…` addresses decode unchanged.
- **Mutation-proven (ask 5):** replacing the decode with a checksum-**skipping** strip turns the corrupted-checksum test RED; the PR's 3 tests pass (`3/3`) with the real `bs58check`.
- **ESM/CJS wrapper:** `mod.decode ?? mod.default?.decode` is correct for the pinned `bs58check@4.0.0` (ESM top-level `decode`, CJS `default.decode`); it mirrors `chains/tron/rpc.ts` and `chains/tron/tx.ts`. Codex independently confirmed the wrapper and reproduced WalletCore accepting a valid TRON address / rejecting a corrupted one.
- CI: all green (CodeRabbit, Integration, Lint+TypeCheck, Unit Tests, Test Summary).

## risk
Low. Read-only balance path (not a fund-loss vector, as the PR states), but it closes a "wrong data silently accepted" gap and restores the checksum discipline the rest of the TRON surface already applies. Behavior changes only for malformed input (now rejected instead of silently misrouted); valid addresses are unaffected.

## verdict
**APPROVE.** Correct checksum + version enforcement, mutation-proven, no false-reject; the send/keysign path is already fully checksum-guarded (native-TRX, TRC-20, and WalletCore rails), so no send-path follow-up is required. Safe to ship in the 2.16.0 bump.

---

PR #972, opened 2026-07-04
Fail closed on native swap missing expiry

### LINKED ISSUE #934: fix(swap): native-swap broadcast guard — MayaChain vault re-validation skipped + expiry==0 bypass
Two defects in the same native-swap broadcast guard (`packages/sdk/src/vault/services/nativeSwapBroadcastGuard.ts`); they compound.

## 1. MayaChain skips inbound-vault re-validation
`nativeSwapBroadcastGuard.ts:43` — `if (native.chain !== Chain.THORChain) return`. The inbound-vault-address comparison (lines 55-71) runs for THORChain only, but `nativeSwapChains = [THORChain, MayaChain]` (`core/chain/swap/native/NativeSwapChain.ts`). Every MayaChain native swap returns before the `activeInboundAddress !== native.vaultAddress` check.

**Impact:** a MayaChain swap whose inbound vault rotated (or was tampered in a stale quote) broadcasts to the wrong vault address with no guard — the fund-loss class the THORChain branch exists to prevent, unprotected for Maya.

**Fix:** generalize the THORChain block to any `nativeSwapChains` member, selecting the Maya vs Thor inbound endpoint by `native.chain`.

## 2. Staleness guard bypassed when `expirationTime == 0`
`nativeSwapBroadcastGuard.ts:36` — `if (native.expirationTime > 0n && native.expirationTime <= currentSeconds)`. A zero/unset `expirationTime` skips the expiry check entirely, so a payload with unset expiration has no broadcast-time staleness gate. Combined with (1), a stale MayaChain quote passes both guards.

**Fix:** treat a missing/zero `expirationTime` as a hard error (fail-closed), or require the quote builder to always populate it and assert that here.

---

rcoderdev:
Closes #934

Rejects native swap payloads with a missing or non-positive expiration before fetching inbound addresses or broadcasting, while preserving the existing expired-quote failure for positive stale expirations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fail-closed native swap broadcasts are rejected when quote expiration is missing, zero, or invalid.
  * Secured-asset withdrawals are allowed to proceed without standard inbound expiration/inbound-vault validation.
  * THORChain deposit signing now derives asset details from the native swap payload (memo/decimals/secured status), including secured-withdrawals using 0 decimals.
* **Tests**
  * Added THORChain regression coverage for secured vs standard withdrawals and strengthened missing/invalid expiration rejection order/behavior.
  * Updated unit tests to assert the new nested swap payload shape.
* **Chores**
  * Updated published release metadata for core-mpc and sdk.

### DIFF
--- a/.changeset/soft-native-swaps.md
+++ b/.changeset/soft-native-swaps.md
@@ -1,6 +1,6 @@
 ---
-"@vultisig/core-mpc": patch
+"@vultisig/core-mpc": minor
 "@vultisig/sdk": patch
 ---
 
-Fail closed before broadcasting native swap payloads that do not carry a positive quote expiration, while preserving secured-asset withdrawal deposits.
+Fail closed before broadcasting native swap payloads that do not carry a positive quote expiration, while preserving secured-asset withdrawal deposits with a shared classifier.
--- a/.changeset/soft-native-swaps.md
+++ b/.changeset/soft-native-swaps.md
@@ -1,5 +1,6 @@
 ---
-'@vultisig/sdk': patch
+"@vultisig/core-mpc": patch
+"@vultisig/sdk": patch
 ---
 
-Fail closed before broadcasting native swap payloads that do not carry a positive quote expiration.
+Fail closed before broadcasting native swap payloads that do not carry a positive quote expiration, while preserving secured-asset withdrawal deposits.
--- a/.changeset/soft-native-swaps.md
+++ b/.changeset/soft-native-swaps.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Fail closed before broadcasting native swap payloads that do not carry a positive quote expiration.
--- a/packages/core/mpc/keysign/signingInputs/resolvers/cosmos.test.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/cosmos.test.ts
@@ -162,6 +162,7 @@ describe('getCosmosSigningInputs gas limit', () => {
               contractAddress: '',
               decimals: 8,
             },
+            fromAmount: '10000000',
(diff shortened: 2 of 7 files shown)


gomesalexandre (changes requested):
Blocker:

- Against fresh origin/main, this branch drops the current MayaChain inbound-vault validation in `packages/sdk/src/vault/services/nativeSwapBroadcastGuard.ts`. The new guard returns early for every native swap except `Chain.THORChain`, so Maya swaps no longer fetch Maya inbound addresses or reject a rotated vault.
- Please rebase/merge main and preserve the existing THOR + Maya inbound-vault validation path, then layer the missing-expiration fail-closed check on top.

Receipt:
- Reviewed diff against freshly fetched `origin/main` (`git fetch origin main`) and compared `nativeSwapBroadcastGuard.ts` / its unit test coverage. The regression is visible directly in the changed guard logic: `origin/main` validates both THORChain and MayaChain inbound vaults, while this branch only validates THORChain.

gomesalexandre (approved):
current head review.

re-checked the original maya-chain blocker on current head `c2ed7e9` and it is fixed.
- `packages/sdk/src/vault/services/nativeSwapBroadcastGuard.ts` now preserves the THOR + Maya inbound-vault validation path while fail-closing missing/invalid expirations
- `packages/core/mpc/keysign/swap/getKeysignSwapPayload.ts` keeps the secured-withdrawal exemption THOR-only
- `packages/sdk/tests/unit/vault/services/nativeSwapBroadcastGuard.test.ts` now covers stale-maya-rejects + matching-maya-passes

receipts: current-head Unit / Integration / Lint+Typecheck / Test Summary are green.

PR #973, opened 2026-07-04
Reject non-decimal send amounts

### LINKED ISSUE #935: fix(send): amount parser accepts hex/scientific — BigInt("0x..") yields wrong signed amount
## Evidence
`packages/sdk/src/vault/VaultBase.ts:1912-1915` (`parseAmount`) — guard is `isNaN(Number(trimmed)) || Number(trimmed) <= 0`. `Number("0x10")=16` and `Number("1e5")=100000` both pass, then `BigInt(whole + fraction...)`: `"0x10"+zeros` is parsed by `BigInt` as **hex** (silently wrong base-unit amount); `"1e5"+zeros` throws a raw `SyntaxError` (not a `VaultError`). The swap path uses the robust `validateHumanSwapAmount`→`toChainAmount` (`:1919-1932`) which correctly rejects these — divergent parsing between the two money paths.

## Impact
A caller passing `"0x…"` as a send amount gets a silently wrong (hex-interpreted) signed amount; scientific notation escapes the `VaultError` contract.

## Fix sketch
Route `send` through the same `toChainAmount`/`toBaseUnits` primitive as swap (already rejects `e`, `0x`); drop the ad-hoc `parseAmount`.

---

rcoderdev:
Closes #935
Routes send amount parsing through the SDK decimal-to-base-unit converter so hex-looking and scientific-notation inputs fail under the SDK InvalidAmount contract instead of being accepted or leaking parser errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved amount parsing to accept only valid decimal values and reject non-decimal formats like hex and scientific notation.
  * Invalid or non-positive send amounts now return a clearer `InvalidAmount` error.
  * Updated validation behavior to be more consistent when converting user-entered amounts into base units.


johnnyluo (approved):
LGTM

Issue #946, opened 2026-07-02
correctness: money-path edge-case cluster (swap net-vs-gross, balance false-zeros, decimals)


realpaaao:
Tracking issue for money-path correctness edges from the SDK audit. Individually small; each below is a **confirmed** finding (CHN-02 dropped — no live callers; two "verify-first" items moved to the footnote).

- [ ] **Best-swap-quote ranks GROSS output, ignoring gas/outbound fees** (documented TODO #353) — router can pick a route delivering less NET. `core/chain/swap/quote/findSwapQuote.ts:212-220`.
- [ ] **`fromChainAmount` returns a lossy JS float** (`Number(amount)/Math.pow(10,decimals)`); feeds the swap "you-receive" figure. `core/chain/amount/fromChainAmount.ts:2`.
- [ ] **Native-swap quote requested for a float-rounded input amount** ≠ the signed bigint → slippage/limit keyed to a marginally wrong size. `findSwapQuote.ts:409-415`.
- [ ] **`swap` "max" uses full balance, reserves nothing for gas** (unlike `send`) → native max-swap can't cover fees. `VaultBase.ts:1717-1721`.
- [ ] **Dry-run send fee computed by a different call than the fee in the signed payload** → preview diverges from signed tx. `VaultBase.ts:1683-1692`.
- [ ] **EVM balance fetch swallows RPC failures to `0n`** (false-zero, indistinguishable from empty). `core/chain/coin/balance/getEvmChainBalances.ts:79`.
- [ ] **Polkadot balance returns `free`** (counts frozen/locked as spendable) → max-send overstates. `resolvers/polkadot.ts:154`.
- [ ] **Solana native balance flows through JS `number`** (precision >2^53 lamports). `resolvers/solana.ts:12`.
- [ ] **UTXO validator shows "too little change" on actual insufficient funds** (negative remainder hits the dust branch). `chains/utxo/send/validateUtxoRequirements.ts:35-43`.
- [ ] **`getAllowance` swallows RPC errors → `0n`** → redundant approve tx. `SwapService.ts:224`.
- [ ] **Native-swap network fee priced in wrong asset in dry-run fiat**. `SwapService.ts:456,411`.
- [ ] **`fiatToAmount` uses the un-hardened decimal formatter** `convertAmount` was fixed to avoid. `utils/fiatToAmount.ts:46`.
[...]

PR #976, opened 2026-07-05
Fail closed on allowance fetch failures

rcoderdev:
Refs #946

## Summary by CodeRabbit

* **Bug Fixes**
  * ERC-20 allowance lookup errors are now surfaced to the app instead of being quietly treated as zero.
  * This helps avoid incorrect swap behavior when allowance requests fail.
* **Tests**
  * Added coverage to confirm allowance lookup failures are reported properly while existing allowance behaviors remain unchanged.


johnnyluo (approved):
LGTM

PR #999, opened 2026-07-06
feat(cosmos): parameterize thorchain-family msgdeposit asset (thor.rune / maya.cacao)

gomesalexandre:
Companion to the app's Maya-source swap support. The react-native `buildThorMsgDeposit` hardcoded `THOR.RUNE` in the MsgDeposit Coin, so a MayaChain deposit built via that path would sign RUNE instead of CACAO (wrong asset).

## what
`buildThorchainDepositTx` gains an optional `asset` (`{ chain, symbol, ticker }`) that defaults to `THOR.RUNE`. Existing callers that omit it are byte-identical; MayaChain-source swaps pass `MAYA.CACAO`.

## risk
Minimal - a single default-param addition. Verified: default omitted -> byte-identical THOR.RUNE output (pinned in the golden test), so no existing THORChain LP/swap deposit changes.

## test
3/3 new (`thorchain-deposit-asset.test.ts`): back-compat THOR.RUNE default, MAYA.CACAO encoding, and a realistic CACAO->RUNE swap memo/amount carried through untouched.

## unblocks
vultiagent-app Maya-source swap PR (which passes `asset: MAYA.CACAO` and bumps to this version) + the agent-backend-ts execute_swap Maya branch. That app PR must land AFTER this publishes (else CACAO deposits sign RUNE), so it's gated on this.

## Summary by CodeRabbit

* **New Features**
  * Cosmos THORChain-family deposit transactions now allow selecting the deposit asset (instead of always using the default).
  * MayaChain deposit flows can encode the intended asset values for correct signing behavior.

* **Bug Fixes**
  * Preserves prior behavior when no asset is specified, ensuring existing deposit transactions remain unchanged.
  * Ensures the encoded deposit coin asset matches the selected asset while keeping the memo content intact.

* **Tests**
  * Added unit test coverage for asset parametrization, memo preservation, and invalid-asset validation.

### DIFF
--- a/.changeset/maya-deposit-asset.md
+++ b/.changeset/maya-deposit-asset.md
@@ -1,5 +1,5 @@
 ---
-"@vultisig/sdk": minor
+'@vultisig/sdk': minor
 ---
 
 feat(cosmos): parameterize the THORChain-family MsgDeposit asset (THOR.RUNE / MAYA.CACAO)
--- a/.changeset/maya-deposit-asset.md
+++ b/.changeset/maya-deposit-asset.md
@@ -1,0 +1,11 @@
+---
+"@vultisig/sdk": minor
+---
+
+feat(cosmos): parameterize the THORChain-family MsgDeposit asset (THOR.RUNE / MAYA.CACAO)
+
+`buildThorchainDepositTx` gains an optional `asset` ({ chain, symbol, ticker }) that
+defaults to THOR.RUNE, so every existing caller is byte-identical. MayaChain-source
+swaps pass MAYA.CACAO, enabling native CACAO->RUNE (and RUNE-source) deposit swaps in
+the app's react-native signing path (previously hardcoded THOR.RUNE, which would sign
+the wrong asset for a Maya deposit).
--- a/packages/sdk/src/platforms/react-native/chains/cosmos/index.ts
+++ b/packages/sdk/src/platforms/react-native/chains/cosmos/index.ts
@@ -6,6 +6,7 @@ export type {
   BuildThorchainDepositOptions,
   CosmosStakingMsg,
   CosmosTxBuilderResult,
+  ThorchainDepositAsset,
 } from './tx'
 export {
   buildCosmosSendTx,
--- a/packages/sdk/src/platforms/react-native/chains/cosmos/tx.ts
+++ b/packages/sdk/src/platforms/react-native/chains/cosmos/tx.ts
@@ -151,9 +151,33 @@ function buildThorMsgSend(fromAddress: string, toAddress: string, amount: string
 }
 
 /** THORChain-family MsgDeposit Asset { chain, symbol, ticker } — THOR.RUNE / MAYA.CACAO. */
-export type ThorchainDepositAsset = { chain: string; symbol: string; ticker: string }
+export type ThorchainDepositAsset = {
+  chain: string
+  symbol: string
+  ticker: string
+}
 
-const THOR_RUNE_ASSET: ThorchainDepositAsset = { chain: 'THOR', symbol: 'RUNE', ticker: 'RUNE' }
+const THOR_RUNE_ASSET: ThorchainDepositAsset = {
+  chain: 'THOR',
+  symbol: 'RUNE',
+  ticker: 'RUNE',
+}
+
+/**
+ * Fail closed on a malformed MsgDeposit asset. An empty/blank chain, symbol, or
+ * ticker serializes an empty protobuf field and would sign a garbage Coin.asset;
+ * refuse before producing a signature. Asset-family *correctness* (THOR.RUNE for
+ * THORChain vs MAYA.CACAO for MayaChain) remains the caller's responsibility —
+ * this builder only encodes the triplet it is given and cannot map a raw chainId
+ * string to a family without brittle heuristics.
+ */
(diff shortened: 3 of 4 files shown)


NeOMakinG (approved):
## testing approach

Fund-critical SDK review — worst-case adversarial lens on the THORChain-family MsgDeposit asset parameterization. Ships in `@vultisig/sdk` (changeset: **minor**, post-2.16.0). Two lanes: Claude (load-bearing) + Codex gpt-5.4. Unique worktree, ran the suite, mutation-proved every assertion, verified WYSIWYS threading against source.

What I verified:
- **WYSIWYS — the asset is signed.** `buildThorchainDepositTx` → `buildThorMsgDeposit(…, asset)` → `msgDeposit` → `wrapAny('/types.MsgDeposit')` → `txBodyBytes` → `buildSignDoc(txBodyBytes, …, chainId, …)` → `signingHashHex = sha256(signDocBytes)`. The `{chain,symbol,ticker}` the caller passes is exactly what lands in the signed Coin.asset. No stale/default substitution downstream.
- **Byte-identical default (regression).** `opts.asset ?? THOR_RUNE_ASSET` with `THOR_RUNE_ASSET = {THOR,RUNE,RUNE}` encodes the same three fields, same order, as the old hardcoded path — the rename (`runeAsset`→`assetProto`, `runeCoin`→`coin`) is cosmetic. I pinned a **full-txBody golden** (decodes as `Coin{asset{THOR,RUNE,RUNE}, amount} · memo · signer`) so every existing THORChain LP/swap caller is provably unaffected; mutation (asset symbol `RUNE`→`RUNEX`) → golden RED.
- **MAYA.CACAO encoding + mutation.** Reverting the parameterization (ignore `opts.asset`, always THOR.RUNE) → the CACAO + swap-memo tests go RED (2 failed). The suite genuinely locks the asset threading, not just a happy path.
- **Decimals** — the SDK takes `amountBaseUnits` verbatim; no decimals assumption here (RUNE & CACAO are both 1e8, and the caller supplies base units regardless).
- **memo-vs-Coin asset** — the swap-memo test uses `=:THOR.RUNE:…` (the swap *destination*) with Coin asset MAYA.CACAO (the *input*); that's a correct CACAO→RUNE swap, not a mismatch. Confirmed.
- Suite 6/6 (3 original + my 3); `format:check` (repo `.config/.prettierrc`) clean; tsc clean on both changed files; base CI was green.

## should-fix

**Asset-family correctness per chain is not enforced anywhere in the signed path — it's fully delegated to the caller, and the default silently applies THOR.RUNE.** A `buildThorchainDepositTx` call for a MayaChain deposit that *omits* `asset` signs **THOR.RUNE on a Maya deposit** (wrong asset — the exact fund-routing class this PR exists to prevent), and nothing rejects `chainId=mayachain-…` paired with `asset=THOR.RUNE`, or a `=:MAYA.CACAO:…` memo paired with a THOR.RUNE Coin. (Codex flagged the same.)

I did **not** try to fix this inside the builder: it receives a raw `chainId` *string* and has no robust chainId→family map at this layer (the canonical `Chain.MayaChain → 'MAYA'` map lives in `packages/core/…/NativeSwapChain.ts`), so any in-builder enforcement would be brittle string-matching, and making `asset` required would break every existing THORChain caller (the PR is a non-breaking minor by design). So the enforcement must live at the **caller** — the vultiagent-app cosmosTx path and the agent-backend-ts `execute_swap` Maya branch should derive the asset from the `Chain` enum via the core map and **never omit it for a non-THORChain deposit**. Please make sure the two gated follow-up PRs (which pass `MAYA.CACAO`) do this derivation from the chain rather than passing a hand-written literal at each call site, and add a caller-side test that a Maya deposit built without an explicit CACAO asset is impossible (or fails closed). This is the real fund-safety gate; the SDK change is a correct, WYSIWYS primitive underneath it.

## what I changed (fix-and-approve, pushed to the PR branch)

Two safe, non-breaking hardenings + tests (commit `eb4e2790`):
- **`assertValidDepositAsset`** — fail closed on an empty/blank `chain`/`symbol`/`ticker` before signing, so a malformed/garbage Coin.asset can never be signed (your criterion: unknown/garbage → fail closed). It deliberately does *not* whitelist assets (MsgDeposit legitimately carries assets beyond RUNE/CACAO), only rejects the never-valid empty triplet.
- **byte-identity golden test** + two guard tests (empty + partially-empty asset → throws), all mutation-proven.

## risk

**Low, for the SDK change itself.** Correct WYSIWYS encoding, byte-identical THOR.RUNE default (golden-pinned), mutation-proven asset threading, fail-closed on malformed asset. The residual fund risk is entirely at the caller layer (asset-per-chain selection) — flagged above as a should-fix on the gated follow-ups, not on this PR. The app follow-up is already gated to land after this publishes and to pass `MAYA.CACAO`.

## verdict

**APPROVE.** The parameterization is a correct, WYSIWYS, byte-identical-by-default, mutation-proven primitive, now also fail-closed on malformed input. The one real fund-safety concern — asset-per-chain correctness — is a caller responsibility this low-level builder can't cleanly own; enforce it in the two gated follow-up PRs (derive asset from the chain, never omit for Maya).

---
*Two-lane review: Claude (load-bearing) + Codex gpt-5.4. Codex corroborated the caller-must-get-the-triplet-right footgun (should-fix) and the empty-asset fail-open (fixed); its byte-identity/test-strength gap is closed by the golden + guard tests.*

PR #1001, opened 2026-07-07
test(sdk): golden-vector byte tests for gap chains (layer 1)

gomesalexandre:
Layer 1 of the tx-bug-prevention plan: extend the existing golden-vector byte tests to the chains that had no byte-level protocol-conformance coverage, closing gaps at the encode layer before they ever reach an app or a user.

## what
Existing coverage before this PR: EVM/Solana (`tx-builder-golden-vectors.test.ts`), UTXO sighash/txid (bitcoinjs-lib), and cosmos staking msgs only (`cosmos-staking.test.ts`, via cosmjs-types). This PR adds:

- `cosmos-send-golden-vectors.test.ts` — plain `cosmos.bank.v1beta1.MsgSend` (Cosmos Hub / Osmosis / Kujira / ...), plus the THORChain/MayaChain `/types.MsgSend` raw-bytes variant
- `thorchain-deposit-golden-vectors.test.ts` — `types.MsgDeposit` (native swap / LP deposit — the most fund-safety-sensitive cosmos-family message in the SDK)
- `ripple-golden-vectors.test.ts` — XRP Payment tx encoding
- `sui-golden-vectors.test.ts` — intent-message wrapping + serialized-signature assembly
- `ton-golden-vectors.test.ts` — wallet-v4 signing payload (native send + jetton transfer)
- `tron-golden-vectors.test.ts` — TransferContract + TriggerSmartContract (TRC-20)

## why
Each test independently re-derives the unsigned tx bytes / signing payload and asserts byte-for-byte equality against the SDK's own builder output — stronger than a decode round-trip, since it proves the SDK's field construction produces the EXACT bytes a real node would accept, not just bytes that happen to decode back to the right values. This directly de-risks any future tx-builder migration and catches protocol-conformance bugs (wrong wire format) at the encode layer.

## how
Reference encoder used per chain:
- **cosmos MsgSend**: `cosmjs-types` (the real cosmos-sdk reference) — full envelope (Any → TxBody → AuthInfo → SignDoc) re-encoded independently and compared byte-for-byte
- **Ripple**: `xrpl`'s own exported `encode`/`encodeForSigning`
- **Sui**: `@mysten/sui/cryptography`'s `messageWithIntent` + `toSerializedSignature`
- **THORChain/Maya MsgDeposit** and **Tron**: no maintained npm package publishes their proto bindings, so these use `@bufbuild/protobuf`'s audited `BinaryWriter` as an independent wire-format implementation against the chains' own published proto schemas (documented inline in each test file). Flagged as **lower confidence** than the true cross-library checks above, since there's no second real-world implementation to diff against — only an independent encoding of the same documented schema.
- **TON**: `@ton/core` IS the canonical TON cell library (no separate reference package exists), so this suite independently reconstructs the wallet-v4 signing-payload header via manual big-endian byte packing (`DataView`) instead of reusing the SDK's own `storeUint()` call sequence — catches field-order/bit-width regressions in the hand-written helper.

No product code changed — tests only.

## test plan
- [x] `yarn test` (SDK package): 168 test files / 2301 tests passing, zero regressions vs the 162/2276 baseline before this PR
- [x] `yarn typecheck`: clean
- [x] `yarn lint` on the new files: clean (after import-sort autofix)
- [x] `yarn format:check` on the new files: clean (after prettier autofix)

🤖 Agent-generated

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive golden-vector unit tests for React Native transaction signing across Cosmos, Ripple, Sui, THORChain/MayaChain, TON, and Tron.
  * Expanded coverage for byte-for-byte transaction encoding, signing hashes, memo handling, address derivation, and signature serialization.
  * Added regression checks to ensure changes in inputs like amount, memo, sequence number, or transaction bytes produce expected signing output differences.

## VA-81 hardening pass (2026-07-08)

Checked what's actually covered before assuming gaps existed. Found a real one: Polkadot's `signingInputs` resolver already has a byte-level golden-vector test (`packages/core/mpc/keysign/signingInputs/resolvers/polkadot.test.ts` - compiled + decoded through WalletCore, exact SCALE call-index offset pinned), but its sibling **Bittensor resolver had ZERO test coverage** - despite bypassing WalletCore entirely for a fully custom, hand-rolled SCALE encoder (`buildExtrinsic.ts`'s own top comment: "bypasses TW Core to include CheckMetadataHash signed extension"). A wrong compact-encoded amount/nonce or wrong pallet/call index there is a direct fund-loss risk with nothing catching it. Also checked the newer `mobileFixtures.golden.test.ts` suite (landed via this branch's `origin/main` merge, 52 commits) - covers Polkadot (`dot.json`) but has no Bittensor fixture either.

Added `packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts`, verified against REAL references rather than assumed:
- **Live-fetched the actual Bittensor (finney) mainnet runtime metadata** (`state_getMetadata` via `entrypoint-finney.opentensor.ai`) and decoded it with `@polkadot/types` to confirm the hardcoded `balancesPallet=5` / `transferAllowDeath=0` constants genuinely match the live chain - not just the source comment's claim.
- Cross-checked the hand-rolled `compactEncode` against `@polkadot/util`'s real `compactToU8a` across every mode-boundary value - byte-identical.
- Cross-checked the hand-rolled `encodeMortalEra` against `@polkadot/types`' `GenericExtrinsicEra` - byte-identical.
- Full golden-vector byte-for-byte pins on `buildBittensorSigningPayload`'s three outputs and `assembleBittensorExtrinsic`'s final assembly.
- Round-trip test for `getBittensorSigningInputs`' custom tx-input framing, which also had zero coverage.

Test plan: 22/22 new tests passing; full `packages/core` suite (post-52-commit-merge) 166/167 files passing, zero regressions; `tsc --noEmit` clean; eslint/prettier clean. codex-review: clean, first pass.

Branch: `feat_golden_vector_gaps`, commit `e4ae8c86`.


packages/sdk/tests/unit/platforms/react-native/ripple-golden-vectors.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    // to build the reference signed tx, then hand it to xrpl's own encoder.
+    const rBytes = Buffer.from(rHex, 'hex')
+    const sBytes = Buffer.from(sHex, 'hex')
+    const der = Buffer.concat([
+      Buffer.from([0x30, rBytes.length + sBytes.length + 4]),
+      Buffer.from([0x02, rBytes.length]),
+      rBytes,
+      Buffer.from([0x02, sBytes.length]),
+      sBytes,
+    ])
```
NeOMakinG: `should-fix:` this DER block never exercises the fund-critical strict-DER padding branch. The SDK's `derEncode` (ripple/tx.ts l.312-318) strips leading zero bytes and **prepends `0x00` when the high bit of `r` or `s` is set** — XRPL rejects a signature that omits this. But (a) this hand-built reference DER has no strip/pad logic, and (b) the fixture uses `r='1'*64` / `s='2'*64` (both leading byte < 0x80), so the padding path is never hit on either side. A regression that broke the SDK's `0x00` prepend would ship undetected. Add a fixture where `r` (or `s`) has a leading byte >= `0x80` and build the reference via a padding-correct DER encoder (or `@noble/curves`).

packages/sdk/tests/unit/platforms/react-native/tron-golden-vectors.test.ts
```diff
@@ -0,0 +1,259 @@
...
+      tokenAddress: FX.tokenAddress,
+      amount: tokenAmount,
+      feeLimit: FX.feeLimit,
+      refBlockBytes: FX.refBlockBytes,
+      refBlockHash: FX.refBlockHash,
+      expiration: FX.expiration,
+      timestamp: FX.timestamp,
+    })
+
+    const callData = buildTrc20CallData(FX.to, tokenAmount)
```
NeOMakinG, line 227: `suggestion:` this reference reuses the SDK's own `buildTrc20CallData` to build `contractValue`, so the call-data bytes inside `TriggerSmartContract` are SDK-generated on both sides (circular for that sub-portion). It's net-covered because test #2 (l.248) independently validates `buildTrc20CallData` against a hand-computed `a9059cbb || addr || amount`. Consider building this reference from the same hand-computed bytes so test #1 is fully independent too.

packages/sdk/tests/unit/platforms/react-native/cosmos-send-golden-vectors.test.ts
```diff
@@ -0,0 +1,220 @@
...
+      // hand-rolled encoder in utils/cosmosProto.ts.
+      const reader = new BinaryReader(txBody.messages[0].value)
+      const fields: Record<number, Uint8Array> = {}
+      while (reader.pos < reader.len) {
+        const [fieldNo] = reader.tag()
+        fields[fieldNo] = reader.bytes()
+      }
+
+      expect(bytesToHex(fields[1]!)).toBe(bytesToHex(fromBech32(thorFrom).data))
+      expect(bytesToHex(fields[2]!)).toBe(bytesToHex(fromBech32(thorTo).data))
```
NeOMakinG, line 217: `suggestion:` the `/types.MsgSend` (THORChain raw-bytes variant) check asserts only fields 1 & 2 (from/to raw address bytes). The amount `Coin` (field 3) wire encoding isn't byte-checked here. Since this variant can't use the cosmjs-types MsgSend reference, add a hand-decoded assertion on field 3 (denom + amount) so the full inner message is covered.

packages/sdk/tests/unit/platforms/react-native/thorchain-deposit-golden-vectors.test.ts
```diff
@@ -0,0 +1,232 @@
...
+import { describe, expect, it } from 'vitest'
+
+import { buildThorchainDepositTx } from '../../../../src/platforms/react-native/chains/cosmos/tx'
+
+function bytesToHex(bytes: Uint8Array): string {
+  return Array.from(bytes, b => b.toString(16).padStart(2, '0')).join('')
+}
+
+const FX = {
+  chainId: 'thorchain-mainnet-v1',
```
NeOMakinG, line 47: `suggestion:` the PR body and this file's header claim THORChain/**MayaChain** coverage, but every fixture here (and in cosmos-send's `/types.MsgSend`) uses `thorchain-mainnet-v1` + `thor` hrp only. MayaChain shares the wire shape but has a different hrp (`maya`) and denom (`cacao`). Add one Maya-specific fixture to actually lock the shared-wire-shape claim rather than asserting it by extension.

packages/sdk/tests/unit/platforms/react-native/ton-golden-vectors.test.ts
```diff
@@ -0,0 +1,231 @@
...
+})
+
+describe('TON address derivation', () => {
+  it('is deterministic and workchain-sensitive (regression guard)', () => {
+    const wc0 = deriveTonAddress(FX.publicKeyEd25519, { workchain: 0 })
+    const wc0Again = deriveTonAddress(FX.publicKeyEd25519, { workchain: 0 })
+    expect(wc0).toBe(wc0Again)
+    // Different pubkey must derive a different address.
+    const otherPubkey = 'ff'.repeat(32)
+    expect(deriveTonAddress(otherPubkey, { workchain: 0 })).not.toBe(wc0)
```
NeOMakinG, line 229: `suggestion:` unlike the Sui suite (which cross-checks `deriveSuiAddress` against `@mysten/sui`'s `toSuiAddress()`), this only asserts determinism + workchain-sensitivity — it never pins a known-good TON v4r2 address. A pubkey->address golden value (from `@ton/ton`'s WalletContractV4 or a real on-chain address) would catch a derivation regression that determinism alone can't.

NeOMakinG (commented):
Reviewed as a golden-vector provenance audit — the one question that matters for a byte-test PR: are the "expected" values sourced from an INDEPENDENT authority, or self-generated from the code under test (which would just cement bugs as "expected")?

## testing approach
- Worktree at HEAD `82b1ecdf`, `yarn install` + `yarn build:shared`, ran all 6 new files: **25/25 pass**.
- Read each SDK builder the tests import (`platforms/react-native/chains/cosmos/tx.ts`, `chains/tron/tx.ts` + `proto.ts`, `chains/ton/tx.ts`, `chains/ripple/tx.ts`, `chains/sui/tx.ts`) and diffed the reference encoder against the SDK's own encoder per chain.
- Spot-checked chain constants against spec: Tron `raw_data` field numbers (1/4/8/11/12/14/18), contractType 1=Transfer / 31=TriggerSmartContract, TRC-20 selector `a9059cbb`; Sui ED25519 flag `0x00`; TON V4R2 subwallet `698983191` (0x29A9A317); THOR native asset {THOR,RUNE,RUNE}; cosmos SIGN_MODE_DIRECT. All correct.

**Provenance verdict: no self-generated golden vectors, no fully-tautological test.** Every expected value is produced by a different codebase than the SDK-under-test. Tiered by strength:
- **Sui (strongest)** — genuine cross-library vs `@mysten/sui` (the chain's own SDK): intent-wrapping, serialized-signature, AND address derivation.
- **Cosmos MsgSend (strong)** — `cosmjs-types` (protobufjs-generated canonical ref) vs the SDK's hand-rolled `cosmosProto.ts`. Full SignDoc byte equality.
- **THOR/Maya MsgDeposit + Tron (independent encoder, honestly flagged lower-confidence)** — inner msg hand-encoded via `@bufbuild/protobuf` BinaryWriter (independent of the SDK's hand-rolled proto helpers), but both sides transcribe the same documented schema by hand. The inline caveats are accurate.
- **Ripple / TON (only one JS codec exists)** — same underlying codec on both sides; the meaningful independent check is the JSON/header structure re-derived by hand. Author documents this honestly.

The tests assert the actual signable surface (signing hash / SignDoc bytes / serialized signature / BOC / derived address), not incidental metadata. This is a solid, honest, well-documented coverage-expansion PR.

## should-fix
See inline: the Ripple strict-DER 0x00 high-bit-padding branch — a fund-critical XRPL-acceptance requirement — is never exercised by any vector.

## suggestion
See inline: Tron TRC-20 test #1 reuses SDK call-data for its reference (covered by test #2); THOR `/types.MsgSend` asserts only fields 1&2; MayaChain is claimed but only THORChain fixtures run; TON address test checks determinism only, not a known-good value.

## risk
LOW. Tests-only, no product code, no `package.json`/lockfile change (all reference libs already present — no dep bump). Worst case is a pre-existing untested branch (Ripple DER padding) staying unguarded — not a regression this PR introduces.

## verdict
**COMMENT** (draft — not approving per draft policy). Materially strengthens encode-layer protocol-conformance coverage with genuinely-independent references. Address the Ripple DER-padding gap and the suggestions before marking ready.

---
_Reviewed by Claude (Opus 4.8). Codex second-opinion lane could not run this round (concurrent-session contention starved both attempts); Claude lane is load-bearing._

packages/sdk/tests/unit/platforms/react-native/ripple-golden-vectors.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      destination: FX.destination,
+      amount: FX.amount,
+      fee: FX.fee,
+      sequence: FX.sequence,
+      lastLedgerSequence: FX.lastLedgerSequence,
+      signingPubKey: FX.signingPubKey,
+    })
+
+    // Fixed low-S-normalized r||s (64 bytes) — arbitrary but deterministic.
+    const rHex = '1'.repeat(64)
```
NeOMakinG: `should-fix:` Ripple strict-DER 0x00-padding branch is still never exercised — The fixture uses r='1'*64 / s='2'*64 — both high-byte 0x11/0x22, so derEncode's `if (byte & 0x80) prepend 0x00` branch (ripple/tx.ts) never runs, and the hand-built reference DER here also has no padding logic. A regression that broke the 0x00 prepend would still pass this test, yet XRPL rejects a non-strict-DER signature — direct broadcast-failure risk. This is the single fund-relevant branch in the ripple path and it remains the one thing the golden vector does NOT pin.

fix: Add a second fixture where r (or s) leading byte >= 0x80 (e.g. r='ff'+'0'*62), and build the reference DER via a padding-correct encoder (or hardcode the expected 0x00-prepended literal) so the two sides diverge only if derEncode's prepend regresses.

packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts
```diff
@@ -0,0 +1,233 @@
...
+    const expected = Buffer.concat([
+      Buffer.from(era.toU8a()),
+      Buffer.from(compactToU8a(params.nonce)),
+      Buffer.from(compactToU8a(0)),
+      Buffer.from([0x00]), // CheckMetadataHash::Disabled
+    ])
+    expect(hex(signedExtra)).toBe(hex(expected))
+  })
+
+  it('produces the expected full signing payload (callData ++ signedExtra ++ additionalSigned)', () => {
```
NeOMakinG, line 153: `suggestion:` Bittensor full-payload & final-extrinsic goldens are concatenation-only, not self-contained byte pins — The :153 payload and :173 assembly 'expected' values are rebuilt from the callData/signedExtra returned by buildBittensorSigningPayload() itself, so those two assertions only verify the concatenation/additionalSigned wrapping — a regression inside buildCallData/buildSignedExtra is caught by the sibling :129/:141 blocks (which DO pin callData/signedExtra independently via decodeAddress + compactToU8a + GenericExtrinsicEra), so coverage is transitively independent across the suite, but these two tests alone are not full external pins.

fix: Rebuild expected callData/signedExtra from the independent primitives (compactToU8a/decodeAddress/GenericExtrinsicEra) inside :153 and :173 too, so each golden stands alone rather than trusting the SDK's own output.

packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts
```diff
@@ -0,0 +1,233 @@
...
+    100000000000n,
+    123456789012345n, // realistic TAO amounts (base units, 9 decimals)
+  ])('matches @polkadot/util.compactToU8a for %s', value => {
+    expect(hex(compactEncode(value))).toBe(hex(compactToU8a(value)))
+  })
+})
+
+describe('encodeMortalEra — cross-checked against @polkadot/types (the real SCALE reference)', () => {
+  const registry = new TypeRegistry()
+  it.each([
```
NeOMakinG, line 83: `suggestion:` encodeMortalEra cross-check misses the quantized-phase branch (period >> 12 != 0) — The it.each only covers period 64 and 128, where quantizeFactor = max(1, calPeriod>>12) = 1, so the phase-quantization arithmetic in scale.ts encodeMortalEra (the most Substrate-specific branch) is never exercised. A rounding bug there would ship unpinned.

fix: Add period=8192 and period=65536 cases (with a nonzero blockNumber) to the encodeMortalEra it.each so the >>12 quantization path is diffed against GenericExtrinsicEra.

packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts
```diff
@@ -0,0 +1,233 @@
...
+    expect(hex(extrinsic)).toBe(hex(expected))
+  })
+})
+
+// getBittensorSigningInputs's custom [len][callData][len][signedExtra][payload] binary
+// framing (encodeBittensorTxInput / decodeBittensorTxInput) had zero test coverage —
+// a length-prefix miscalculation here would corrupt compileTx's ability to reassemble
+// the final extrinsic (wrong callData/signedExtra slice boundaries), silently signing
+// or broadcasting the wrong bytes.
+describe('getBittensorSigningInputs — custom tx-input framing round-trips', () => {
```
NeOMakinG, line 208: `suggestion:` Bittensor tx-input framing is round-trip-only — a symmetric encode/decode offset bug survives — The framing test only asserts decode(encode(x)) == x, which cannot catch a matched offset error present in BOTH encode and decode, and decodeBittensorTxInput does no bounds-checking — a truncated buffer slices short silently instead of throwing.

fix: Add one byte-literal assertion pinning the [len][callData][len][signedExtra][payload] layout of encodeBittensorTxInput, plus a truncated-input case asserting decode throws (or documenting the silent-slice behavior).

packages/sdk/tests/unit/platforms/react-native/thorchain-deposit-golden-vectors.test.ts
```diff
@@ -0,0 +1,232 @@
+/**
```
NeOMakinG, line 1: `suggestion:` MayaChain coverage is still only claimed, not executed — PR body and describe strings reference MayaChain, but every fixture runs the THORChain wire (thor hrp / RUNE). The shared-wire-shape claim between THOR and Maya is asserted nowhere.

fix: Add one Maya-specific fixture (maya hrp, cacao asset) so the shared-envelope claim is actually locked by a running test, or drop the MayaChain wording to match what executes.

NeOMakinG (commented):
gm — multi-lane pass: claude + codex. runtime-tested on node/vitest — packages/sdk RN golden suite (25/25 green) + packages/core bittensor.test.ts (22/22 green), worktree @ e4ae8c86. comment-only.

## testing approach
- node/vitest — packages/sdk RN golden suite (25/25 green) + packages/core bittensor.test.ts (22/22 green), worktree @ e4ae8c86
- pass order: claude → codex

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | should-fix: Ripple DER 0x00-padding branch never exercised — fixture uses low-bit r/s ('1'*64/'2'*64) and the hand-DER reference has no padding logic, so derEncode's high-bit prepend (XRPL-required) ships unguarded | ❌ still open — ripple-golden-vectors.test.ts:110-111 unchanged at HEAD e4ae8c86; fixture still low-bit r/s, hand-DER still un-padded |
| 1 | suggestion: MayaChain claimed in PR body/headers but only THORChain fixtures run (thor hrp / rune) — no maya-specific fixture (maya hrp / cacao) | ❌ still open — only comment/describe strings mention 'MayaChain'; no maya hrp or cacao fixture actually executes at HEAD |
| 1 | suggestion: Tron TRC-20 golden test#1 reuses SDK buildTrc20CallData for its reference (circular sub-portion) | ℹ️ documented — net-covered by test#2's independent selector+addr+amount check; unchanged, acceptable |
| 1 | suggestion: THOR /types.MsgSend variant asserts only fields 1&2, not field-3 amount Coin bytes | ℹ️ unchanged — minor, acceptable |
| 1 | suggestion: TON deriveTonAddress test checks determinism/workchain only, no known-good v4r2 address pin | ℹ️ unchanged — minor, acceptable |

## should-fix
### 1. Ripple strict-DER 0x00-padding branch is still never exercised — `packages/sdk/tests/unit/platforms/react-native/ripple-golden-vectors.test.ts:110`
The fixture uses r='1'*64 / s='2'*64 — both high-byte 0x11/0x22, so derEncode's `if (byte & 0x80) prepend 0x00` branch (ripple/tx.ts) never runs, and the hand-built reference DER here also has no padding logic. A regression that broke the 0x00 prepend would still pass this test, yet XRPL rejects a non-strict-DER signature — direct broadcast-failure risk. This is the single fund-relevant branch in the ripple path and it remains the one thing the golden vector does NOT pin.
action: Add a second fixture where r (or s) leading byte >= 0x80 (e.g. r='ff'+'0'*62), and build the reference DER via a padding-correct encoder (or hardcode the expected 0x00-prepended literal) so the two sides diverge only if derEncode's prepend regresses.

## suggestion
### 1. Bittensor full-payload & final-extrinsic goldens are concatenation-only, not self-contained byte pins — `packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts:153`
The :153 payload and :173 assembly 'expected' values are rebuilt from the callData/signedExtra returned by buildBittensorSigningPayload() itself, so those two assertions only verify the concatenation/additionalSigned wrapping — a regression inside buildCallData/buildSignedExtra is caught by the sibling :129/:141 blocks (which DO pin callData/signedExtra independently via decodeAddress + compactToU8a + GenericExtrinsicEra), so coverage is transitively independent across the suite, but these two tests alone are not full external pins.
action: Rebuild expected callData/signedExtra from the independent primitives (compactToU8a/decodeAddress/GenericExtrinsicEra) inside :153 and :173 too, so each golden stands alone rather than trusting the SDK's own output.

### 2. encodeMortalEra cross-check misses the quantized-phase branch (period >> 12 != 0) — `packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts:83`
The it.each only covers period 64 and 128, where quantizeFactor = max(1, calPeriod>>12) = 1, so the phase-quantization arithmetic in scale.ts encodeMortalEra (the most Substrate-specific branch) is never exercised. A rounding bug there would ship unpinned.
action: Add period=8192 and period=65536 cases (with a nonzero blockNumber) to the encodeMortalEra it.each so the >>12 quantization path is diffed against GenericExtrinsicEra.

### 3. Bittensor tx-input framing is round-trip-only — a symmetric encode/decode offset bug survives — `packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts:208`
The framing test only asserts decode(encode(x)) == x, which cannot catch a matched offset error present in BOTH encode and decode, and decodeBittensorTxInput does no bounds-checking — a truncated buffer slices short silently instead of throwing.
action: Add one byte-literal assertion pinning the [len][callData][len][signedExtra][payload] layout of encodeBittensorTxInput, plus a truncated-input case asserting decode throws (or documenting the silent-slice behavior).

### 4. MayaChain coverage is still only claimed, not executed — `packages/sdk/tests/unit/platforms/react-native/thorchain-deposit-golden-vectors.test.ts:1`
PR body and describe strings reference MayaChain, but every fixture runs the THORChain wire (thor hrp / RUNE). The shared-wire-shape claim between THOR and Maya is asserted nowhere.
action: Add one Maya-specific fixture (maya hrp, cacao asset) so the shared-envelope claim is actually locked by a running test, or drop the MayaChain wording to match what executes.

## risk
LOW — tests-only PR, no product code, no dep/lockfile change. Net-additive coverage (new Bittensor hand-rolled-SCALE golden vector, genuinely independent vs @polkadot/util + @polkadot/types). No regression introduced. The two open items below are pre-existing untested branches this PR does not newly break — they were flagged in r1 and remain unclosed after the HEAD advance (which is a main-merge + the Bittensor addition, NOT a fix pass on the r1 findings).

## verdict
COMMENTED

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` adversarial pass. node/vitest — packages/sdk RN golden suite (25/25 green) + packages/core bittensor.test.ts (22/22 green), worktree @ e4ae8c86.

packages/sdk/tests/unit/platforms/react-native/ripple-golden-vectors.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      Fee: FX.fee,
+      Sequence: FX.sequence,
+      LastLedgerSequence: FX.lastLedgerSequence,
+      SigningPubKey: FX.signingPubKey.toUpperCase(),
+      TxnSignature: der.toString('hex').toUpperCase(),
+    }
+
+    expect(finalized.signedTx).toEqual(referenceSignedTx)
+    expect(finalized.signedBlobHex).toBe(xrplEncode(referenceSignedTx))
+  })
```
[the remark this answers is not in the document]
gomesalexandre: fixed in 0ad0a637. Added the 130-hex recovery-byte path assertion and kept the finalized payload byte-for-byte pinned against xrpl.encode.

packages/sdk/tests/unit/platforms/react-native/ripple-golden-vectors.test.ts
gomesalexandre: fixed in 0ad0a637. The vector now uses a high-bit `r` (`ff...`) and an independent DER helper that prepends `0x00` when needed, so the strict-DER padding path is exercised too.

packages/sdk/tests/unit/platforms/react-native/tron-golden-vectors.test.ts
gomesalexandre: keeping this one as follow-up for now. The current reviewer-blocker was the ripple strict-DER gap; TRC-20 call-data independence is already transitively covered by the sibling hand-built selector/address/amount assertion.

packages/sdk/tests/unit/platforms/react-native/cosmos-send-golden-vectors.test.ts
gomesalexandre: keeping this as follow-up for now. It is a good completeness add, but this pass was scoped to the reviewer-blocking ripple branch coverage gap.

packages/sdk/tests/unit/platforms/react-native/thorchain-deposit-golden-vectors.test.ts
gomesalexandre: fair point. I am not widening scope further in this pass; if we keep the Maya wording I will land a dedicated maya/cacao fixture follow-up rather than hide it inside the ripple fix.

packages/sdk/tests/unit/platforms/react-native/ton-golden-vectors.test.ts
gomesalexandre: keeping this as follow-up for now. Worth a hard pin, but not on the critical path for the current reviewer-blocking ripple fix.

packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts
gomesalexandre: keeping the bittensor deepening suggestions for a separate follow-up. This pass was scoped to closing the current ripple blocker first.

packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts
gomesalexandre: same call here: good follow-up, but I am not widening this pass beyond the ripple blocker fix.

packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts
gomesalexandre: same here: good hardening idea, but I am keeping this pass scoped to the reviewer-blocking ripple fix so we can move the PR without mixing unrelated follow-up work.

NeOMakinG (approved):
Round 3 re-review at HEAD `0ad0a637`. The r2 `should-fix` (Ripple strict-DER `0x00`-padding branch never exercised) is now **resolved and mutation-sound**. Approving — test-only PR, golden vectors correct and independent.

## testing approach
Worktree at PR head, `yarn install` + `yarn build:shared`, `vitest run` on the RN golden-vector suite (115/116 pass — the one `entry.test.ts` failure is the pre-existing MPC-singleton duplicate-engine isolation issue, unrelated). Codex `gpt-5.4` second-opinion lane on mutation-soundness + test-encoder correctness. Live mutation check: neutered `derEncode`'s high-bit `0x00`-prepend branch in `ripple/tx.ts` and confirmed the test **fails** (reference `3045022100FF…` padded vs mutated `30440220FF…` unpadded).

## prior issues — status
| r1/r2 | 1 | should-fix | Ripple `derEncode` `0x00` high-bit padding branch never exercised (fixture used low-bit `r`=`1*64`) | ✅ fixed | `0ad0a637`: fixture now `rHex='ff'+'0'*62` (leading byte `0xff`, high bit set). Reference DER built by the test's **own** independent `encodeDerInteger`/`buildReferenceDerSignature` (hand `Buffer` logic, not SDK output → no tautology). Mutation-proven: dropping the pad fails the test. Bonus: `finalize(r+s+'01')` asserts recovery-byte strip. |
| r1/r2 | 2 | suggestion | MayaChain claimed in PR body but only THORChain fixtures execute (thor hrp / RUNE) | ⚠️ unchanged | "MayaChain" still appears only in comments/describe strings; no maya-hrp/cacao fixture runs. Non-blocking, restated below. |

## suggestion
- **MayaChain fixture still not exercised.** `cosmos-send-golden-vectors.test.ts:175` and `thorchain-deposit-golden-vectors.test.ts` name MayaChain in describe/comments, but only `thor` hrp + `thorchain-mainnet-v1` fixtures run. The shared-wire-shape claim for maya (maya hrp / cacao asset) remains asserted-not-tested. Add one maya fixture to lock it. Non-blocking.
- **Ripple DER coverage is r-side only** (Codex lane). The new fixture proves the `r` high-bit pad path with a low-`S` value, but does not exercise the leading-zero **strip** branch (`r = 00 00 7f…`), the `s`-side high-bit pad, or low-`S` normalization interacting with DER length. A bug confined to those shapes would still slip past. Optional follow-up.

## risk
LOW. Test-only, no product code, no dep/lockfile change. The fund-critical XRPL strict-DER path is now guarded where it previously shipped unverified.

## verdict
**Approve.** The r2 should-fix is resolved with an independent, mutation-sound reference. Remaining items are non-blocking suggestions.

PR #1003, opened 2026-07-07
fix(agent): make tool-output the sole signing source, remove tx_ready path (#927 Phase 2)

neavra:
## What

Phase 2 of #927. Removes the `tx_ready` signing path from the agent CLI so the client-enriched **tool-output** candidate becomes the **only** signing source.

Phase 1 (#927) made the CLI dual-read: it derived a signable candidate from the `tool-output-available` SSE channel while the backend `tx_ready` stayed authoritative, with a parity cross-check that logged any divergence. That has baked on main and validated live (Polymarket e2e + runtime receipts). This PR rips out the now-redundant `tx_ready` signing path.

## Why

The production backend writes the signable payload on `tool-output-available` and emits `data-tx_ready` only as a hollow `{typed_confirm}` marker (no transaction body) for high-distrust calldata tools — a signal the CLI never consumed. Keeping `tx_ready` as an authoritative signing source no longer matches what the backend emits, so tool-output becomes the sole source. Both flat builders (polymarket / `build_custom_*` / `erc20_approve`) and `execute_*` prep tools (`execute_send`/`execute_swap`/`execute_contract_call`) now sign from their tool-output.

## Changes

- **`session.ts`**: `selectAndBufferSignable` buffers the tool-output candidate, gated by `payloadLooksSignable` (the fail-closed structural check that mirrors the executor's real signer requirements). Removes the `tx_ready` capture/selection, the divergence-defer paths, and `logToolOutputParity`. Disconnect recovery no longer replays a persisted `tx_ready` to sign (that was a signing path) — a recovered turn that ran a signable tool now warns to re-run, because the signable payload rides tool-output and can't be reconstructed from persisted parts.
- **`client.ts`**: removes `onTxReady`, the `tx_ready` SSE case, the dead `SSEStreamResult.transactions` field, and the parity pairing holder. `data-tx_ready` now routes to `ignore` (the hollow marker is a no-op).
- **`toolOutputSigning.ts`**: removes the parity machinery (`diffToolOutputParity`, `canonicalize*`, `HARD_FIELDS`, …). Renames `CLI_PARITY_PREP_TOOLS` → `CLI_SIGNABLE_PREP_TOOLS` — prep now signs from tool-output; the `txArgs.tx_encoding` requirement in `deriveToolOutputCandidate` is the phantom-card fail-closed gate.
- **tests**: rewrites the dual-read / parity suites into single-source suites. Every surviving fail-closed property keeps a test, including a mutation-check that deleting the signable-structural gate makes an unsignable candidate reach the signer.

## Fail-closed posture (preserved)

1. `deriveToolOutputCandidate` rejects non-tx envelopes (`no_op` / `insufficient_*` / errors) and phantom-card prep envelopes (missing `tx_encoding`).
2. `payloadLooksSignable` rejects structurally-unsignable candidates before buffering.
3. `storeServerTransaction` is the final backstop (a `false` never signs).

The confirmation gate (`runPasswordGatedTool`) is unchanged — every signing route still passes through it.

## Testing

- `yarn workspace @vultisig/cli typecheck` — green
- `yarn test:cli` — 542 passing
- `yarn check` (typecheck + lint + knip + format) — green
- Changeset: `@vultisig/cli` patch (behavior-preserving alignment; the same real transactions still sign).

## Summary by CodeRabbit

* **New Features**
  * CLI signing now uses only the `tool-output-available` signal as the single source for signable transactions, ignoring the legacy `tx_ready` marker.
  * `execute_*` (prep-style) tool signing is now derived from tool output with stronger chain-resolution validation, including non-EVM name-only cases.

* **Bug Fixes**
  * Removed the alternate signing/replay path to prevent duplicate signing and mismatches.
  * Disconnect recovery now restores recovered message content (text/balance) without attempting to re-sign unreconstructable tool data.
  * Fail-closed gating and warning behaviors are preserved (and second signable candidates are deferred/handled safely).

* **Tests**
  * Updated and added agent session coverage for Phase 2 tool-output-only signing and recovery behavior.


gomesalexandre (approved):
Clean Phase 2 ser, 558/558 tests clean post-merge.

Adversarial reads I ran:

**Prep candidates now signing (new in Phase 2)**: old Phase 1 code had `flatIsSignSource = ... source === 'flat'` — only flat was a sign source; prep was parity-only. Phase 2 removes that guard. Intentional, and correctly safe: `deriveToolOutputCandidate` rejects prep envelopes without `tx_encoding` BEFORE the candidate ever reaches `selectAndBufferSignable`. So the old "prep → parity-only" guard's purpose (prevent a malformed prep from signing when tx_ready would've suppressed it) is now enforced at the earlier layer. ✓

**Disconnect recovery "warn-don't-sign"**: old path replayed any persisted `data-tx_ready` through `onTxReady` → sign gate. Phase 2 warns to re-run instead. Correct: tool-output payloads aren't persisted server-side (only the text answer + balance cards persist), so there's genuinely nothing to replay. Fail-safe posture ✓.

**`selectAndBufferSignable` simplification**: now `toolOutputCandidate` → `payloadLooksSignable` → `storeServerTransaction`, two gates, first-wins. Both gates were already present as the last-resort line in Phase 1. ✓

**Changeset**: `@vultisig/cli` patch. CLI-only changes, no sdk/core/lib surface touched. ✓

rcoderdev (approved):

gomesalexandre (approved):
Formatting-only follow-up on top of the Phase 2 review.

Checked the new head commit only. The delta is the expected Prettier reflow in the two files that failed the static gate, no behavior drift. Local receipt from this sweep:
- external Prettier binary against the PR checkout now passes on both flagged files
- git diff --check clean

PR #1005, opened 2026-07-07
Restore mobile keysign golden fixtures

rcoderdev:
Restores the Android/iOS keysign JSON corpus as SDK golden coverage (23 fixture files / 71 cases).

The suite keeps the restored mobile expected hashes intact and makes SDK differences explicit: Cardano max-send differs through the SDK JS/WASM WalletCore path, and Terra Classic USTC preserves the SDK LUNC fee/tax path rather than forcing the mobile fixture fee shape. UTXO ordering is normalized only in the test assertion so production signature ordering remains unchanged.

## Summary by CodeRabbit

* **New Features**
  * Expanded mobile MPC keysigning test coverage across additional chains, tokens, memos, staking/contract flows, and swap scenarios.
* **Bug Fixes**
  * Improved consistency when alternate chain-specific payload field names are provided.
  * Enhanced swap-payload detection for Thorchain and Mayachain variants, improving generated signing pre-images.
* **Tests**
  * Added/expanded golden fixtures and a golden test suite to verify mobile keysigning pre-image hashes deterministically across cases.


packages/core/mpc/keysign/tests/helpers/normalizeKeysignPayloadFromJson.ts
```diff
@@ -68,6 +68,13 @@ export const normalizeKeysignPayloadFromJson = (input: any) => {
         index: numberOrUndefined(u.index),
         script: u.script,
         path: u.path,
+        cardanoTokens: Array.isArray(u.cardano_tokens ?? u.cardanoTokens)
```
gomesalexandre, line 71: Pushed a fix here. The mobile Cardano fixtures were normalizing UTXOs without a `cardanoTokens` array, so ADA-only cases would hit `cardanoTokens.length` as `undefined` in the Cardano signing-input resolver. This now preserves `cardano_tokens` when present and defaults ADA-only UTXOs to `[]`.

gomesalexandre (commented):
Pushed a mergefix onto `mobile-keysign-fixtures` for the Cardano path.

What I changed
- preserved `cardano_tokens` during mobile fixture normalization
- defaulted ADA-only UTXOs to `cardanoTokens: []` so the Cardano signing-input resolver does not crash on `.length`

Receipts
- static trace: the PR's new `cardano.json` fixtures have ADA-only UTXOs with no token field, while `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts` unconditionally reads `cardanoTokens.length`
- attempted local retest in the disposable clone after wiring deps, but the branch currently fails earlier in `yarn build:shared` on an existing Cardano typing error: `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts(56,5): error TS2353: Object literal may only specify known properties, and \`auxiliaryData\` does not exist in type \`ISigningInput\`.`

Please let CI re-run on this head and then we can stamp it if green.

gomesalexandre (approved):
test-only: 23 fixture files + golden test harness expanding mobile keysign coverage to 71 cases across all chains. The mapper additions in `mapBlockchainSpecific.ts` are test-only (field aliasing for JSON normalization). No production src changes. Clean coverage expansion.

gomesalexandre (approved):
test-only refresh still looks gucci at current head.

receipts
- scope stayed test-only: current diff is still 27 files under `packages/core/mpc/keysign/tests/`
- reran the live golden path on the current head and it passed: `yarn test:core --config .config/vitest.core.config.ts` with `mobileFixtures.golden.test.ts` green, and the core suite stayed green overall
- the new fixups are coherent: `solana-sign-data.json` now uses the mapper's `has_program_id` flag, and Cardano `byteFee` now uses `bigishToString`, which is the right precision-safe pattern for int64-ish fields

approve.

PR #1009, opened 2026-07-07
Batch portfolio native price fetches

### LINKED ISSUE #948: perf: portfolio balance/price batching + gas & rujira-pair caching
Tracking issue for performance findings from the SDK audit. (Gas-caching item dropped — gas is too volatile to cache safely, and it's what gets signed.)

- [ ] **Portfolio balance fetch is fully sequential** (N chains × M tokens serial `await`), while a batched primitive (`tools/evm/balanceEvm.ts:37`) sits unused. `BalanceService.ts:91-105`. Fix: `Promise.allSettled` across chains (preserve the per-chain try/catch) + wire the batched EVM path.
- [ ] **Portfolio price fetch ignores the class's own batched `getPrices`** — one HTTP call per coin instead of one batched call; compounds with the balance loop so `portfolio()` is ~2N sequential round trips. `VaultBase.ts:919-947`, `FiatValueService.ts:197`. Fix: call `getPrices` once up front, look up in the loop.
- [ ] **Rujira pair lookup fetches a 200-row list + client-side fuzzy match per call, no cache** (pair listings change on the order of days). `rujira/src/modules/range.ts:589`. Fix: TTL-memoize the pair list.

---

rcoderdev:
Closes #948


gomesalexandre (approved):
clean perf improvement ser - native prices batch to one getPrices call, tokens stay parallel via Promise.all. fallback to individual getPrice is solid defensive pattern. test covers the batching assertion precisely (getPrices called once, getPrice only for token). changeset present. LGTM

PR #1010, opened 2026-07-07
Fetch balances in parallel

rcoderdev:
Refs #948


gomesalexandre (approved):
nice parallelism improvement - the test verifying all 3 fetches start before any resolves is exactly the right check.

suggestion: the failure semantics changed subtly - in the old sequential loop, if a token fetch threw, the native balance was already written to result and stayed there. now Promise.all rejects on any token failure and the whole chain (including already-fetched native) is discarded. probably cleaner behavior (partial chain = confusing), but worth noting for anyone debugging a suddenly-missing chain balance. no test covers the token-failure case.

LGTM

PR #1012, opened 2026-07-07
fix: findSwapQuote no longer masks a genuine transient outage as no-route

gomesalexandre:
tl;dr the AllProvidersFailed fallback message unconditionally contains "no route", so agent-backend-ts's `isTransientQuoteError` bails out on it before ever checking for network/timeout/5xx signals — a real all-providers outage always got reported as a hard, definitive "this pair has no route" dead-end.

## what

Flagged as a follow-up while sweeping agent-backend-ts's multi-swapper coverage (vultisig/agent-backend-ts#783): `findSwapQuote`'s `AllProvidersFailed` catch-all always throws:

```
No swap route found after trying CowSwap, KyberSwap, 1inch, LiFi, SwapKit, THORChain, MayaChain.
```

That phrase always contains "no route", and `isTransientQuoteError` (agent-backend-ts, `execute_swap.ts`) is explicitly gated to bail (return false) the instant it sees that phrase — before it ever gets to check for network/timeout/5xx patterns. So even when every single provider failed because of a genuine transient blip (rate limits, oracle timeouts, a 502 from thornode), the SDK's own raw provider errors were discarded in favor of a generic message that structurally can never be classified as transient downstream. Users got told a corridor "has no route" when actually every provider was just unreachable for a moment.

## how

Adds `isTransientProviderFailure` — checked only when literally EVERY rejected provider's raw rejection reason is transient (reuses the same below-minimum/halted/no-route bail-outs already used elsewhere in this function, so a provider that positively answered "no route" or "below minimum" still wins and nothing regresses there). Only in that all-transient case, throws a distinct message that:
- Names the detected category ("transient network/timeout error") instead of leaking raw upstream provider text — kept just as noise-free as the existing generic fallback (see the "omits noisy provider errors" test, unchanged).
- Deliberately avoids the "no route" phrase, so `isTransientQuoteError` and any other downstream classifier keyed on that phrase can correctly detect it as transient.

The existing generic message is completely untouched for every other case (mixed signals, or a genuine structural no-route) — so none of the many downstream consumers keyed on its exact wording (evals scorers, `mcp.ts`'s `classifySwapNoRoute`, the scheduler retry messaging) change behavior. Zero changes needed in agent-backend-ts itself — the new message already satisfies its existing `isTransientQuoteError` heuristics.

## testing

- 2 new tests in `findSwapQuote.selection.test.ts`: all-transient (network/timeout/5xx mix, including bare `ETIMEDOUT`/`HTTP 502` forms) -> new transient message, no "no route" phrase; mixed transient + one genuine structural no-route -> unchanged generic message (a real decline still wins).
- Full existing `findSwapQuote.*.test.ts` suite green (90 tests, run with `--no-file-parallelism` to rule out an unrelated pre-existing timeout flake under parallel load - confirmed present on clean `main` too, unrelated to this change).
- `tsc --project .config/tsconfig.json --noEmit` clean.

## risk

Low - additive-only branch gated on a narrow all-transient condition that doesn't overlap with any existing fixture (all current tests use plain non-matching error strings like `"kyber fail"`). No behavior change for the genuine no-route path.

### DIFF
--- a/packages/core/chain/swap/quote/findSwapQuote.ts
+++ b/packages/core/chain/swap/quote/findSwapQuote.ts
@@ -110,6 +110,13 @@ type RankedSwapQuote = {
 
 const QUOTE_FETCH_TIMEOUT_MS = 30_000
 
+// Node/undici network-layer error codes — mirrors agent-backend-ts's
+// `TRANSIENT_QUOTE_CODE_RE` (execute_swap.ts) for the same reason: a provider's raw
+// rejection surfaces these as bare codes (e.g. "ETIMEDOUT"), which don't contain the
+// substrings "timeout" or "timed out" checked elsewhere.
+const TRANSIENT_PROVIDER_CODE_RE =
+  /\b(ETIMEDOUT|ECONNRESET|ECONNREFUSED|ECONNABORTED|EPIPE|ENOTFOUND|EAI_AGAIN|UND_ERR_\w+)\b/i
+
 const withTimeout = <T>(p: Promise<T>, ms: number): Promise<T> => {
   let timeoutId: ReturnType<typeof setTimeout> | undefined
   const timeoutPromise = new Promise<never>((_, reject) => {
@@ -808,6 +815,55 @@ export const findSwapQuote = async ({
     )
   }
 
+  // A provider's raw rejection is a transient infra failure (network/timeout/5xx),
+  // not a genuine structural decline. Bails to false on the same "no route" /
+  // dust-threshold / halted substrings checked above so a provider that positively
+  // answered is never misclassified as transient. Mirrors (deliberately duplicated,
+  // not imported — separate repo/package) agent-backend-ts's execute_swap.ts
+  // `isTransientQuoteError`, which classifies the SAME kind of raw single-provider
+  // error for the native asyncFallbackChain path. Used below to detect the case
+  // where EVERY provider failed transiently, so the generic `AllProvidersFailed`
+  // fallback doesn't collapse a genuine outage into a "no route" hard-negative that
+  // downstream classifiers (and `isTransientQuoteError` itself) cannot un-collapse.
+  const isTransientProviderFailure = (msg: string): boolean => {
+    const lower = msg.toLowerCase()
+    if (
+      lower.includes('no swap routes found') ||
+      lower.includes('no swap route found') ||
+      lower.includes('no routes found') ||
+      lower.includes('no route found') ||
+      lower.includes('amount less than min swap amount') ||
+      isBelowMinimumMsg(msg) ||
+      isTradingHaltedMsg(msg) ||
+      lower.includes('dust threshold') ||
+      lower.includes('does not meet requirements')
+    ) {
+      return false
+    }
+    return (
+      lower.includes('fetch failed') ||
+      lower.includes('failed to fetch') ||
+      lower.includes('network error') ||
+      lower.includes('socket hang up') ||
+      lower.includes('connection refused') ||
+      lower.includes('econnrefused') ||
+      lower.includes('timeout') ||
+      lower.includes('timed out') ||
+      lower.includes('deadline exceeded') ||
+      lower.includes('the operation was aborted') ||
+      /\bhttp\s+5\d{2}\b/.test(lower) ||
+      lower.includes('bad gateway') ||
+      lower.includes('service unavailable') ||
+      lower.includes('gateway timeout') ||
+      lower.includes('internal server error') ||
+      lower.includes('rate limit') ||
+      lower.includes('too many requests') ||
+      /\b429\b/.test(lower) ||
+      lower.includes('skip api network error') ||
+      TRANSIENT_PROVIDER_CODE_RE.test(msg)
+    )
+  }
+
   const belowMinimumByProvider = new Map<SwapQuoteProviderName, string>()
   const haltedProviders = new Set<SwapQuoteProviderName>()
   let proactiveTradingHalt: SwapError | null = null
@@ -905,6 +961,33 @@ export const findSwapQuote = async ({
   console.warn(
     `[findSwapQuote] no route for ${from.ticker} -> ${to.ticker}; raw provider errors: ${rawProviderErrors.join(' | ')}`
   )
+
+  // If EVERY provider failed transiently (network/timeout/5xx — never a genuine
+  // structural decline), say so explicitly instead of the generic "no route"
+  // wording. The generic message below is deliberately noise-free (never embeds
+  // raw provider text — see "omits noisy provider errors" test) and, worse,
+  // unconditionally contains "no route", which downstream consumers' transient-
+  // error classifiers (e.g. agent-backend-ts's `isTransientQuoteError`) bail out
+  // on BEFORE checking for network/timeout signals — so a real outage was always
+  // reported as a hard, definitive "this pair has no route" dead-end. This is the
+  // one case where that's wrong: nobody actually declined the route, every
+  // provider was simply unreachable. Kept just as noise-free as the generic
+  // fallback (no raw provider text), naming the detected category rather than the
+  // upstream body, so downstream classification is deterministic without leaking
+  // upstream internals.
+  const rejectedReasons = settled
+    .filter((result): result is PromiseRejectedResult => result.status === 'rejected')
+    .map(result => (result.reason instanceof Error ? result.reason.message : String(result.reason)))
+  const allProvidersTransient = rejectedReasons.length > 0 && rejectedReasons.every(isTransientProviderFailure)
+
+  if (allProvidersTransient) {
+    throw new SwapError(
+      SwapErrorCode.AllProvidersFailed,
+      `Swap quote lookup failed for all ${failedProviders.length} provider(s) tried ` +
+        `(${failedProviders.join(', ')}) due to a transient network/timeout error. ` +
+        `This is not a missing route — retry the same swap shortly.`
+    )
+  }
 
   throw new SwapError(
     SwapErrorCode.AllProvidersFailed,
--- a/packages/core/chain/swap/quote/findSwapQuote.ts
+++ b/packages/core/chain/swap/quote/findSwapQuote.ts
@@ -44,6 +44,7 @@ import { isEmpty } from '@vultisig/lib-utils/array/isEmpty'
 import { isOneOf } from '@vultisig/lib-utils/array/isOneOf'
 import { bigIntToNumber } from '@vultisig/lib-utils/bigint/bigIntToNumber'
 import { isInError } from '@vultisig/lib-utils/error/isInError'
+import { HttpResponseError } from '@vultisig/lib-utils/fetch/HttpResponseError'
 import { pick } from '@vultisig/lib-utils/record/pick'
 import { TransferDirection } from '@vultisig/lib-utils/TransferDirection'
 
@@ -825,7 +826,8 @@ export const findSwapQuote = async ({
   // where EVERY provider failed transiently, so the generic `AllProvidersFailed`
   // fallback doesn't collapse a genuine outage into a "no route" hard-negative that
   // downstream classifiers (and `isTransientQuoteError` itself) cannot un-collapse.
-  const isTransientProviderFailure = (msg: string): boolean => {
+  const isTransientProviderFailure = (reason: unknown): boolean => {
+    const msg = reason instanceof Error ? reason.message : String(reason)
     const lower = msg.toLowerCase()
     if (
       lower.includes('no swap routes found') ||
@@ -839,6 +841,12 @@ export const findSwapQuote = async ({
       lower.includes('does not meet requirements')
     ) {
       return false
+    }
+    // Structured status takes precedence over message-sniffing where available
+    // (HttpResponseError.status) — a 429/5xx is unambiguously a transient
+    // infra signal regardless of how the provider worded the body.
+    if (reason instanceof HttpResponseError && (reason.status === 429 || reason.status >= 500)) {
+      return true
     }
     return (
       lower.includes('fetch failed') ||
@@ -977,7 +985,7 @@ export const findSwapQuote = async ({
   // upstream internals.
   const rejectedReasons = settled
     .filter((result): result is PromiseRejectedResult => result.status === 'rejected')
-    .map(result => (result.reason instanceof Error ? result.reason.message : String(result.reason)))
+    .map(result => result.reason)
   const allProvidersTransient = rejectedReasons.length > 0 && rejectedReasons.every(isTransientProviderFailure)
 
   if (allProvidersTransient) {
--- a/packages/core/chain/swap/quote/findSwapQuote.ts
+++ b/packages/core/chain/swap/quote/findSwapQuote.ts
@@ -199,6 +199,93 @@ const assertNativeTradingOpen = async (input: {
 const asTradingHaltedSwapError = (reason: unknown): SwapError | null =>
   reason instanceof SwapError && reason.code === SwapErrorCode.TradingHalted ? reason : null
 
+const isBelowMinimumMsg = (msg: string) => {
+  const lower = msg.toLowerCase()
+  return (
+    lower.includes('below minimum') ||
+    lower.includes('minimum amount') ||
+    lower.includes('min amount') ||
+    lower.includes('amount too small') ||
+    lower.includes('below the minimum')
+  )
+}
+
+// Native protocols halt trading per-chain (THORChain mimir `HALT<CHAIN>TRADING`,
+// pool ragnarok, churn). The quote API then rejects EVERY amount with
+// "trading is halted" — an operational state, not an amount problem. Detect it
+// so we surface a "temporarily unavailable" message instead of the misleading
+// generic "no route" (which reads like the pair is unsupported). Also match the
+// "trading paused" wordings emitted by the THOR halt helpers in
+// `chains/cosmos/thor/lp/halts.ts` (`global trading paused`,
+// `<chain> chain trading paused`). (#604, CodeRabbit)
+const isTradingHaltedMsg = (msg: string) => {
+  const lower = msg.toLowerCase()
+  return (
+    lower.includes('halted') ||
+    lower.includes('trading halt') ||
+    lower.includes('trading paused') ||
+    lower.includes('trading is paused')
+  )
+}
+
+// A provider's raw rejection is a transient infra failure (network/timeout/5xx),
+// not a genuine structural decline. Bails to false on the same "no route" /
+// dust-threshold / halted substrings checked above so a provider that positively
+// answered is never misclassified as transient. Mirrors (deliberately duplicated,
+// not imported — separate repo/package) agent-backend-ts's execute_swap.ts
+// `isTransientQuoteError`, which classifies the SAME kind of raw single-provider
+// error for the native asyncFallbackChain path. Used below to detect the case
+// where EVERY provider failed transiently, so the generic `AllProvidersFailed`
+// fallback doesn't collapse a genuine outage into a "no route" hard-negative that
+// downstream classifiers (and `isTransientQuoteError` itself) cannot un-collapse.
+// Hoisted to module scope (rather than nested in findSwapQuote) since it closes
+// over no local state — only the module-level regex and the two helpers above.
+const isTransientProviderFailure = (reason: unknown): boolean => {
+  const msg = reason instanceof Error ? reason.message : String(reason)
+  const lower = msg.toLowerCase()
+  if (
+    lower.includes('no swap routes found') ||
+    lower.includes('no swap route found') ||
+    lower.includes('no routes found') ||
+    lower.includes('no route found') ||
+    lower.includes('amount less than min swap amount') ||
+    isBelowMinimumMsg(msg) ||
+    isTradingHaltedMsg(msg) ||
+    lower.includes('dust threshold') ||
+    lower.includes('does not meet requirements')
+  ) {
+    return false
+  }
+  // Structured status takes precedence over message-sniffing where available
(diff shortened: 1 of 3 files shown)


packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -808,6 +815,55 @@ export const findSwapQuote = async ({
...
+  // not a genuine structural decline. Bails to false on the same "no route" /
+  // dust-threshold / halted substrings checked above so a provider that positively
+  // answered is never misclassified as transient. Mirrors (deliberately duplicated,
+  // not imported — separate repo/package) agent-backend-ts's execute_swap.ts
+  // `isTransientQuoteError`, which classifies the SAME kind of raw single-provider
+  // error for the native asyncFallbackChain path. Used below to detect the case
+  // where EVERY provider failed transiently, so the generic `AllProvidersFailed`
+  // fallback doesn't collapse a genuine outage into a "no route" hard-negative that
+  // downstream classifiers (and `isTransientQuoteError` itself) cannot un-collapse.
+  const isTransientProviderFailure = (msg: string): boolean => {
```
NeOMakinG: `q:` isTransientProviderFailure is a hand-maintained third copy of this substring set — This mirrors agent-backend-ts's `isTransientQuoteError` (substring form) and mcp-ts's `isTransientQuoteError` (regex form). The contract that makes the fix work — "matches a transient token AND contains none of the downstream `no route` bail phrases" — is now spread across three repos with no shared lock. If one side later adds a transient token (or a new bail phrase) and another doesn't, the reclassification silently skews and a real outage can slip back into a hard no-route.

fix: Consider a cross-repo fixture asserting this exact new message classifies transient in each classifier (or at least a comment cross-referencing the agent-backend-ts/mcp-ts lines) so the three copies can't drift apart unnoticed.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -906,6 +962,33 @@ export const findSwapQuote = async ({
...
+  // reported as a hard, definitive "this pair has no route" dead-end. This is the
+  // one case where that's wrong: nobody actually declined the route, every
+  // provider was simply unreachable. Kept just as noise-free as the generic
+  // fallback (no raw provider text), naming the detected category rather than the
+  // upstream body, so downstream classification is deterministic without leaking
+  // upstream internals.
+  const rejectedReasons = settled
+    .filter((result): result is PromiseRejectedResult => result.status === 'rejected')
+    .map(result => (result.reason instanceof Error ? result.reason.message : String(result.reason)))
+  const allProvidersTransient = rejectedReasons.length > 0 && rejectedReasons.every(isTransientProviderFailure)
```
NeOMakinG, line 988: `q:` Fail-open residual: a genuinely dead corridor whose every provider happens to emit only 5xx/timeout/rate-limit text — If a truly no-liquidity corridor coincidentally has ALL providers reject with transient-shaped strings (none says `no route`), `allProvidersTransient` is true and the user is told "retry the same swap shortly" while downstream retries a corridor that is economically dead. Bounded (retry budget, not fund loss) and unlikely — a dead corridor normally returns `no route` from at least one provider, which then wins the `.every()`. Flagging only so the downstream retry cap is confirmed finite; not a defect in this diff.

fix: No SDK change needed; confirm the agent-backend-ts/mcp-ts retry loop on a transient classification has a finite cap so an all-transient-but-dead corridor can't loop indefinitely.

NeOMakinG (commented):

## q
- Cross-repo contract verified against the ACTUAL classifier bodies (not just the PR prose) — link #731 as the companion
- isTransientProviderFailure is a hand-maintained third copy of this substring set
- Fail-open residual: a genuinely dead corridor whose every provider happens to emit only 5xx/timeout/rate-limit text

## risk
Low. Message-only refinement in the terminal all-providers-failed path — no quote amount, route ranking, provider selection, or signable-tx surface is touched (same `SwapErrorCode.AllProvidersFailed`, distinct text). The discriminator fails CLOSED: any unknown/ambiguous rejection is NOT transient, so it falls through to the unchanged generic "No swap route found…" message (the mixed-signal test proves a single genuine `no route` still wins). Structural declines (below-minimum / dust / trading-halt / native-min) all throw BEFORE the new branch, and the branch is only reachable when `selectBestEligibleQuote` found zero fulfilled quotes (every provider rejected), so it can never fire when a usable quote existed. No dependency / caret bump (2 files, no package.json). Downstream value is verified live in mcp-ts (`list-swap-routes.ts:616`, positive-match on `/timeout/i`) and pending in agent-backend-ts #731 (`isTransientQuoteError`, e35812a — not yet on main): the new message contains `timeout` and avoids every `no route` bail phrase, so both classify it as transient with zero agent-backend-ts changes, exactly as claimed.

## verdict
COMMENTED

NeOMakinG (commented):
Round 2 — re-review of the HEAD-advanced delta (`a2a72c75..ef347bb1`) and draft→ready pass. The branch was rebased onto current main (cosmos-gas work + changesets are rebase noise); the **PR-specific footprint vs `main` is exactly the two swap files** — `findSwapQuote.ts` + `findSwapQuote.selection.test.ts`, no `package.json`, no caret bump.

## testing approach
- Full source trace of the delta discriminator at `ef347bb1`; re-confirmed the reachability + fail-closed invariants from r1 rather than trusting them.
- Reproduced the red CI static gate locally in a clean worktree (`yarn check:ci` composite) and bisected which sub-check fails.
- Prettier + ESLint run clean on both changed files locally.
- Codex `gpt-5.4` second-opinion lane (exit 0): pointed at the three delta regions (regex def L118, status branch L848-851, all-transient branch L986-997); no distinct new blocker extractable (terse/pollution-truncated output).
- Could not execute the vitest file in the ephemeral worktree (subpath-export self-import needs a build); relied on CI `Integration Tests (Vitest) — pass` + read-trace of the one new test.

## prior issues — status
| r1 | 1 | q | link companion `agent-backend-ts#731` (the `isTransientQuoteError` no-route bail) so the SDK fix isn't half-wired | ⚠️ partial | body now references `agent-backend-ts#783` (the sweep) + new code comments name `execute_swap.ts`'s classifier; `#731` still not linked. Non-blocking. |
| r1 | 2 | q | `isTransientProviderFailure` is a hand-maintained 3rd copy of the transient substring set → drift risk; suggested shared fixture | ⚠️ partial | delta *documents* the mirror (`TRANSIENT_PROVIDER_CODE_RE` "mirrors agent-backend-ts's `TRANSIENT_QUOTE_CODE_RE`") but adds no shared/enforced fixture — still comment-only coupling. |
| r1 | 3 | q | fail-open residual: a dead corridor emitting only 5xx/timeout from every provider gets retried downstream | ℹ️ still open | delta slightly *widens* the transient set (opaque 5xx `HttpResponseError` now classifies transient by status) — correct for genuine 5xx, but confirm downstream retry cap stays finite. |

## blocking
1. **Missing `@vultisig/sdk` changeset → CI static gate is red AND the fix ships inert.** `findSwapQuote.ts` is bundled into `@vultisig/sdk`, and the repo's release guard (`scripts/check-sdk-bundled-changeset.mjs`, run inside `yarn check:ci`) fails with:
   > SDK release guard failed. This PR changes source files that are bundled into `@vultisig/sdk`, but no changeset bumps `@vultisig/sdk`. … Bundled source changes: `packages/core/chain/swap/quote/findSwapQuote.ts`

   This is the cause of the red `Lint and Type Check` check (the guard runs before lint/typecheck in the `&&` chain, which is why those weren't reached). It is delta-caused, not pre-existing — `main` passes the guard because it sees no bundled-source delta. Beyond the red gate, this is *semantically* blocking: without a changeset there is **no new SDK tarball**, so the downstream classifiers this PR exists to un-collapse (mcp-ts `isTransientQuoteError`, agent-backend-ts `execute_swap.ts`) never receive the new message — the fix is inert to consumers. **Fix:** add a `patch` changeset for `@vultisig/sdk` (`yarn changeset` → select `@vultisig/sdk`).

## q
2. Restating r1#1 (non-blocking): please link the specific `agent-backend-ts` PR/commit that added the `no swap route found` bail in `isTransientQuoteError`, and confirm it's on abts `main`, so the "zero abts changes needed" claim is verifiably live rather than gated on an unmerged companion.

## risk
Discriminator logic is **sound and fail-closed at `ef347bb1`** — the delta *improved* it (structured `HttpResponseError.status` + bare Node/undici code regex), and the no-route/below-min/halt/dust message bail still runs **first**, so a provider that positively declined can never be misclassified as transient, and the `.every()` + `length>0` gate is intact. The change remains message-only in the terminal all-providers-failed branch with no fund-path side effect. The sole blocker is release-hygiene (missing changeset) — trivial to fix but genuinely gating both CI and the fix's delivery.

## verdict
**COMMENT — holding approval.** Not a logic regression: the core discriminator is clean and the delta is a real improvement. But I won't approve a now-ready PR whose CI static gate is red and whose fix would ship inert to every downstream consumer. Add the `@vultisig/sdk` changeset and I'll approve on the next pass.

🤖 dual-lane review: Claude (load-bearing source trace + CI-gate repro) + Codex gpt-5.4 (second opinion). GLM retired.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -808,6 +816,62 @@ export const findSwapQuote = async ({
...
+      isTradingHaltedMsg(msg) ||
+      lower.includes('dust threshold') ||
+      lower.includes('does not meet requirements')
+    ) {
+      return false
+    }
+    // Structured status takes precedence over message-sniffing where available
+    // (HttpResponseError.status) — a 429/5xx is unambiguously a transient
+    // infra signal regardless of how the provider worded the body.
+    if (reason instanceof HttpResponseError && (reason.status === 429 || reason.status >= 500)) {
```
NeOMakinG: `q:` (restating r1#3, non-blocking) the structured-status branch is a genuine improvement and the ordering is right — the no-route/below-min/halt/dust bail runs *above* this, so a 5xx that carried a real "no route" body still fails closed. The one residual: this now classifies an **opaque** 5xx/429 (body with no keyword) as transient purely by status. Correct for a real infra blip, but it slightly widens the fail-open surface for a genuinely dead corridor whose providers all coincidentally return 5xx. Bounded by the downstream retry cap — just confirm that cap is finite.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -110,6 +111,13 @@ type RankedSwapQuote = {
 
 const QUOTE_FETCH_TIMEOUT_MS = 30_000
 
+// Node/undici network-layer error codes — mirrors agent-backend-ts's
+// `TRANSIENT_QUOTE_CODE_RE` (execute_swap.ts) for the same reason: a provider's raw
+// rejection surfaces these as bare codes (e.g. "ETIMEDOUT"), which don't contain the
+// substrings "timeout" or "timed out" checked elsewhere.
+const TRANSIENT_PROVIDER_CODE_RE =
```
NeOMakinG, line 118: `q:` (restating r1#2, non-blocking) good that the comment now names `agent-backend-ts`'s `TRANSIENT_QUOTE_CODE_RE` as the mirror — but this is still the third hand-maintained copy of the transient set (here, abts substring form, mcp-ts regex form) with only prose coupling. A shared cross-repo fixture (one canonical list of transient/no-route strings that all three assert against) would make drift a test failure instead of a silent classification skew. Follow-up, not this PR.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -906,6 +970,33 @@ export const findSwapQuote = async ({
...
+  // provider was simply unreachable. Kept just as noise-free as the generic
+  // fallback (no raw provider text), naming the detected category rather than the
+  // upstream body, so downstream classification is deterministic without leaking
+  // upstream internals.
+  const rejectedReasons = settled
+    .filter((result): result is PromiseRejectedResult => result.status === 'rejected')
+    .map(result => result.reason)
+  const allProvidersTransient = rejectedReasons.length > 0 && rejectedReasons.every(isTransientProviderFailure)
+
+  if (allProvidersTransient) {
```
NeOMakinG, line 990: `blocker:` CI "Lint and Type Check" still red — `findSwapQuote` cognitive complexity 36 > 35 (PR-caused merge gate) — the r2 changeset blocker is genuinely resolved, but flipping the guard green just unmasked the next gate in the same `check:ci` `&&` chain. eslint now fails: `449:44 error Refactor this function to reduce its Cognitive Complexity from 36 to the 35 allowed sonarjs/cognitive-complexity`. line 449 is the `findSwapQuote` arrow itself (unchanged), pushed one over the limit by this PR's added nested `isTransientProviderFailure` helper (L829) + this `allProvidersTransient` branch (L989–997). CI red = hard merge gate; the fix cannot land like this.

fix: hoist `isTransientProviderFailure` to module scope — it closes over only module-level symbols (`isBelowMinimumMsg`, `isTradingHaltedMsg`, `TRANSIENT_PROVIDER_CODE_RE`, `HttpResponseError`), so hoisting is safe and drops the nested-function contribution below 35. re-run `yarn check:ci` (or `eslint` on the file) locally to confirm green before pushing.

NeOMakinG (commented):

## prior issues — status
| round | issue | status |
|---|---|---|
| r2 | blocker: no @vultisig/sdk changeset → fix ships inert (no new tarball, downstream classifiers never get the new message) | ✅ fixed — `.changeset/transient-swap-quotes.md` (`@vultisig/sdk: patch`) added @39a7a5d4; guard passes locally; verified end-to-end: agent-backend-ts `list-swap-routes.ts` imports `findSwapQuote` from `@vultisig/sdk` → thin wrapper (`tools/swap/findSwapQuote.ts`) rethrows the core `SwapError` untouched → new transient message survives to the target classifier |
| r2 | blocker: CI "Lint and Type Check" RED (was: `changeset:check-sdk-bundle`) | ❌ still red — the changeset guard now passes, but the SAME check now fails on `sonarjs/cognitive-complexity` at `findSwapQuote.ts:449` (36 > 35). PR-caused; was masked at r2 because the guard runs first in the `&&` chain and short-circuited before lint |
| r1 | q: link companion abts #731 in PR body so the SDK fix isn't half-wired | ⚠️ partial — still not linked |
| r1 | q: transient substring set is a hand-maintained 3rd copy (SDK + abts + mcp-ts), no shared fixture → drift | ⚠️ open — 3 copies now confirmed, still no shared lock |
| r1 | q: fail-open residual — an economically-dead corridor where every provider coincidentally emits only 5xx/timeout text is misclassified transient | ℹ️ open/bounded — unchanged, downstream retry cap is finite (no fund loss) |

## blocking
### 1. CI "Lint and Type Check" still red — `findSwapQuote` cognitive complexity 36 > 35 (PR-caused merge gate) — `packages/core/chain/swap/quote/findSwapQuote.ts:991`
the r2 changeset blocker is genuinely resolved, but flipping the guard green just unmasked the next gate in the same `check:ci` `&&` chain. eslint now fails: `449:44 error Refactor this function to reduce its Cognitive Complexity from 36 to the 35 allowed sonarjs/cognitive-complexity`. line 449 is the `findSwapQuote` arrow itself (unchanged), pushed one over the limit by this PR's added nested `isTransientProviderFailure` helper (L829) + this `allProvidersTransient` branch (L989–997). CI red = hard merge gate; the fix cannot land like this.
action: hoist `isTransientProviderFailure` to module scope — it closes over only module-level symbols (`isBelowMinimumMsg`, `isTradingHaltedMsg`, `TRANSIENT_PROVIDER_CODE_RE`, `HttpResponseError`), so hoisting is safe and drops the nested-function contribution below 35. re-run `yarn check:ci` (or `eslint` on the file) locally to confirm green before pushing.

## should-fix
### 1. SwapService.wrapSwapError collapses the new transient message back to generic no-route for @vultisig/sdk's higher-level swap surface — `packages/sdk/src/vault/services/SwapService.ts:527`
the fix is end-to-end only for consumers of the direct `findSwapQuote` export (that's the abts/mcp classifier path, so the PR's stated purpose IS met). but `SwapService.getSwapQuote` → `wrapSwapError` maps BOTH `SwapErrorCode.NoRoutesFound` and `SwapErrorCode.AllProvidersFailed` to a hardcoded `'No swap route found between these tokens. Try a different pair or provider.'` VaultError, discarding the SwapError's own `.message`. so any caller on the SwapService public API gets the exact hard no-route framing this PR set out to eliminate — including the literal `no swap route found` substring transient classifiers bail on. (found by codex lane.)
action: give `AllProvidersFailed` its own case in `wrapSwapError` that preserves the transient wording (e.g. `return new VaultError(VaultErrorCode.InvalidConfig, message, error)` like the AmountBelowMinimum case), and update the unit contract at `packages/sdk/tests/unit/services/SwapService.test.ts:499` that currently pins the collapsed mapping. out of scope for a minimal fix, but worth a follow-up so the transient/no-route distinction survives on every SDK swap surface, not just the direct export.

## q
- changeset bump is `patch` — a downstream-observable message-shape change that consumers key on; `patch` is defensible (message text is not an API contract) but flagging in case the team treats classifier-visible strings as semver-relevant

## risk
low-mechanical-blocker + one secondary gap. the discriminator core is UNCHANGED since r2 (delta @39a7a5d4 = only the changeset file) — re-verified fail-closed: no-route/dust/halt bail runs before the structured-status (429/5xx) branch and before positive substrings, so a 5xx carrying a real 'no route' body still returns false; unknown non-HttpResponseError → false → generic no-route. changeset resolution verified end-to-end to the abts target. blocker is purely the CI merge gate (cognitive complexity), trivially fixable by hoisting the helper. cannot approve until CI "Lint and Type Check" is green.

## verdict
COMMENTED

NeOMakinG (approved):
Round 4 re-review @ `511c1825`. The r3 hold was a hard CI merge gate — the added transient helper tipped `findSwapQuote`'s `sonarjs/cognitive-complexity` to 36 (> 35), keeping "Lint and Type Check" red. That is now resolved and **all CI gates are green**. Approving.

## testing approach
- Ephemeral worktree at `511c1825` (`git worktree add`, upstream fetched first; canonical checkout untouched).
- Delta since r3 (`git diff 39a7a5d4..511c1825`) = **exactly one file**: `packages/core/chain/swap/quote/findSwapQuote.ts` ("fix: unbork cognitive complexity gate in findSwapQuote").
- Re-verified the transient-vs-no-route discriminator is byte-identical + still fail-closed after the scope move; re-confirmed the changeset; read CI directly via `gh pr checks 1012`.
- Two-lane: Claude load-bearing (conclusive). Codex `gpt-5.4` second-opinion lane **stalled twice (zero output)** — inconclusive this round; non-load-bearing, and r1–r3 Codex already corroborated the discriminator with no distinct blocker.

## prior issues — status
|---|---|---|---|---|---|
| r2 | 1 | blocker | missing `@vultisig/sdk` changeset → CI-red + inert fix | ✅ fixed | `.changeset/transient-swap-quotes.md` (`"@vultisig/sdk": patch`) present at `511c1825`; changeset guard green |
| r3 | 1 | blocker | `findSwapQuote.ts:449` `sonarjs/cognitive-complexity` 36 > 35 → "Lint and Type Check" red | ✅ fixed | `511c1825` hoists `isBelowMinimumMsg`/`isTradingHaltedMsg`/`isTransientProviderFailure` to module scope + flattens the `belowMinimum` if-nest; "Lint and Type Check" now **pass** (13m46s) |
| r3 | 2 | should-fix (Codex) | `SwapService.wrapSwapError` collapses `AllProvidersFailed`→literal "no route", discarding the transient msg on the `SwapService.getSwapQuote` surface | ℹ️ documented | out of this PR's diff scope; the abts/mcp path uses the direct `findSwapQuote` export where the message survives, so the PR's stated purpose is met. Follow-up for full cross-surface coverage |
| r1/r2 | — | q | fail-open residual on a dead corridor emitting only 5xx/timeout text | ℹ️ documented | unchanged by this delta; bounded by finite downstream retry cap |

## verification of the r3 fix (the hoist)
- The three helpers are now defined **exactly once, at module scope** (L202/221/243) — no duplicate, no shadowing. Bodies are byte-identical to the r3 nested versions (confirmed via `git diff`).
- The hoisted `isTransientProviderFailure` closes over only module-level symbols: `HttpResponseError` (import L47), `TRANSIENT_PROVIDER_CODE_RE` (L118), and the two sibling helpers — all module scope. Safe hoist, no accidental rebind.
- **Discriminator still sound + fail-closed:** bail-first ordering preserved (no-route / below-min / halt / dust substrings return `false` first) → structured-status branch (`HttpResponseError` 429/≥500) → positive transient substrings. Unknown error → matches neither → `false` → unchanged generic no-route. Never fails open.
- **Consumer branch unchanged** (L985–994): `allProvidersTransient = rejectedReasons.length > 0 && rejectedReasons.every(isTransientProviderFailure)` still guards the distinct "transient network/timeout error" message.
- The `belowMinimum` if-nest flatten (`if (isBelowMinimumMsg(msg)) { if (!has) … }` → `if (isBelowMinimumMsg(msg) && !has) …`) is behavior-preserving (same short-circuit, same map-population order, same `fetchers[i].providerName`).

## CI (per `gh pr checks 1012` @ `511c1825`)
- CodeRabbit — **pass**
- Integration Tests (Vitest) — **pass**
- Lint and Type Check — **pass** (was the r3 blocker)
- Unit Tests — **pass**
- Test Summary — **pass**

## risk
Low. Pure mechanical scope hoist of byte-identical helpers + one behavior-preserving if-flatten; no change to the discriminator logic, quote amounts, route ranking, provider selection, or any signable-tx path. Message-only terminal-path change, verified fail-closed. Changeset present so the fix reaches downstream consumers once published.

## verdict
**Approve.** r2 + r3 blockers both closed, all CI gates green, changeset present, discriminator intact and fail-closed at `511c1825`.

<sub>🤖 dual-lane review (Claude load-bearing; Codex gpt-5.4 second-opinion lane stalled/inconclusive this round).</sub>

PR #1013, opened 2026-07-07
feat: unblock Sui + Cardano as SwapKit/MayaChain swap sources (SDK groundwork only)

gomesalexandre:
tl;dr live-verified today that Sui and Cardano route as a SwapKit SOURCE (via NEAR-Intents), and MayaChain has a live Available ADA.ADA pool. Neither was in the SDK's dispatch-eligibility gates, so `findSwapQuote` never even tried these corridors regardless of real liquidity existing. This PR is SDK groundwork + tests ONLY -- it does NOT make either corridor signable/usable end-to-end yet. See "what remains" below before either reaches users.

## what

- `swapKitSourceChains` (SwapKitEnabledChains.ts) now includes `Chain.Sui` + `Chain.Cardano`.
- `nativeSwapEnabledChainsRecord[MayaChain]` (NativeSwapChain.ts) now includes `Chain.Cardano`, matching the confirmed live ADA.ADA pool on mayanode.
- `getSwapKitQuote` explicitly rejects Sui/Cardano as a source, before any network call, with a clear "not yet supported for signing" error.

## why the SwapKit path needed an explicit guard

`getSwapKitQuote` isn't just a quote lookup -- it calls SwapKit's real `/v3/swap` endpoint and builds the actual signable tx as part of "getting a quote" (`buildSwapKitTx`). Tracing that function found neither Sui nor Cardano has a working tx-build path:
- Sui: SwapKit returns the tx as a base64 string, and there's a dormant `normalizedTxType === 'SUI'` decode branch in `encodeSwapKitTxPayload`, but there is no `GeneralSwapTx` variant a Sui signer could consume anywhere in the type system -- a Sui PTB is neither `evm`-shaped (no `to`/`data`) nor a plain `transfer` (not a simple send).
- Cardano: `encodeSwapKitTxPayload` explicitly returns an **empty byte array** for `normalizedTxType === 'CARDANO'` -- no decode was ever implemented.

Without a guard, enabling dispatch would fall through to the generic EVM-tx builder, which would either throw an unrelated "not a transaction object" error, or -- worse -- silently build a nonsense `{evm: {...}}` shape if a response object ever coincidentally looked record-like. Rejecting explicitly, before the network round-trip, is strictly safer and avoids wasting a `/v3/swap` call on a request that can never produce a signable tx.

## why the MayaChain-native path didn't need one

`getNativeSwapQuote` is a pure rate/fee GET (thornode/mayanode `/quote/swap`) -- no tx bytes are built at the quote-fetch stage at all for the native path. Enabling this corridor for quote lookups carries no fund-handling risk on its own. The signable Cardano send-with-memo transaction is built downstream, entirely outside this `swap/` module (see "what remains").

## what remains before either corridor is user-safe

- A real `GeneralSwapTx` encode path for Sui (PTB) and Cardano (CBOR) sources, replacing the explicit reject in `getSwapKitQuote`.
- The app-side MPC signing path for a Cardano-source MayaChain swap has **not** been traced or verified as part of this change -- whatever consumes the native `{inboundAddress, memo}` swap payload to build the actual send-with-memo transaction needs confirming it correctly handles Cardano's transaction/metadata format.
- A real sim broadcast test for both corridors once the above lands.

Do not read this PR as "Sui/Cardano swaps now work" -- they're still blocked, deliberately, at the signing layer.

## testing

- 2 new parametrized tests in `getSwapKitQuote.test.ts`: Sui and Cardano source both reject with the clear message, and no `fetch` call is made.
- 3 new tests in `findSwapQuote.selection.test.ts`: SwapKit fetcher now dispatches (is called) for Sui-source and Cardano-source; MayaChain native fetcher dispatches for Cardano-source.
- All 5 new tests confirmed fail-before/pass-after via `git stash` (they genuinely fail without this change, not just trivially pass).
- Full existing swap test suite green: 307 tests, run with `--no-file-parallelism` to rule out a pre-existing timeout flake under parallel load (confirmed present on clean `main` too, unrelated to this change).
- `tsc --project .config/tsconfig.json --noEmit` clean (pre-existing, unrelated `@vultisig/mpc-types` module-resolution errors present on clean `main` too -- not introduced here).

## risk

Low for what's actually shipped: the MayaChain-native gate change is fund-safety-inert (quote-only GET), and the SwapKit gate change is paired with an explicit, tested reject that fires before any signable tx could ever be built. The real fund-safety surface (actual signing) is explicitly NOT enabled by this PR.

### DIFF
--- a/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
+++ b/packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
@@ -548,6 +548,24 @@ const buildTransferTx = (
   }
 }
 
+// Sui + Cardano are eligible SwapKit SOURCE chains for quote-dispatch purposes
+// (see SwapKitEnabledChains.ts) but have no wired tx-build path here yet:
+//   - Sui: SwapKit returns the tx as a base64 string (`encodeSwapKitTxPayload`'s
+//     dormant `normalizedTxType === 'SUI'` branch decodes it), but there is no
+//     `GeneralSwapTx` variant a Sui signer can consume — it is neither `evm`
+//     (no `to`/`data` fields) nor a plain `transfer` (a Sui PTB isn't a simple
+//     send). Falling through to `buildEvmTx` would either throw an unrelated
+//     "not a transaction object" error, or worse, silently build a nonsense
+//     `{evm: {...}}` shape if the response ever coincidentally looks
+//     record-like.
+//   - Cardano: `encodeSwapKitTxPayload` explicitly returns an EMPTY byte array
+//     for `normalizedTxType === 'CARDANO'` — there is no decode implementation
+//     at all, so any tx built from it would be silently wrong.
+// Rejected in `getSwapKitQuote` BEFORE the network round-trip (no route/swap
+// API calls wasted on a request that can never produce a signable tx).
+// Signing support for these two as a SwapKit source is separate follow-on work.
+const SWAP_SOURCE_TX_BUILD_UNSUPPORTED: ReadonlySet<SwapKitSourceChain> = new Set([Chain.Sui, Chain.Cardano])
+
 const buildSwapKitTx = (
   response: SwapKitSwapResponse,
   from: AccountCoin<SwapKitSourceChain>,
@@ -705,6 +723,13 @@ export const getSwapKitQuote = async ({
   affiliateBps,
   slippage = 3,
 }: Input): Promise<GeneralSwapQuote> => {
+  if (SWAP_SOURCE_TX_BUILD_UNSUPPORTED.has(from.chain)) {
+    throw new Error(
+      `SwapKit ${from.chain} source swaps are not yet supported for signing (quote-only for now). ` +
+        'Try a different source chain, or swap the other direction.'
+    )
+  }
+
   const quoteBody = {
     sellAsset: toSwapKitAsset(from),
     buyAsset: toSwapKitAsset(to),
--- a/packages/core/chain/swap/native/NativeSwapChain.ts
+++ b/packages/core/chain/swap/native/NativeSwapChain.ts
@@ -51,12 +51,16 @@ export const nativeSwapEnabledChainsRecord = {
     Chain.Arbitrum,
     Chain.Zcash,
     // Live Available ADA.ADA pool confirmed on mayanode (~4,795 ADA,
-    // mayanode.mayachain.info/mayachain/pools). getNativeSwapQuote is a pure
-    // rate/fee GET (no tx bytes built here), so enabling this corridor for
-    // quote lookups carries no fund-handling risk on its own. The signable
-    // Cardano send-with-memo transaction is built downstream (outside this
-    // swap/ module, in the generic Cardano send-tx compiler) — NOT traced or
-    // verified as part of this change. See PR description for what remains.
+    // mayanode.mayachain.info/mayachain/pools). Review follow-up (fund-safety
+    // pass): unlike SwapKit's Sui/Cardano corridors, the SDK's own
+    // buildSwapKeysignPayload -> getCardanoSigningInputs path IS already
+    // wired end-to-end for a signable Cardano deposit (memo -> CIP-20 aux
+    // data), so this is NOT inert groundwork the way the comment here
+    // originally implied. getNativeSwapQuote (api/getNativeSwapQuote.ts) now
+    // rejects Cardano as a source BEFORE the network call, mirroring
+    // getSwapKitQuote's guard, until a real deposit confirms MayaChain's
+    // Cardano bifrost observer actually reads the CIP-20 label as the
+    // routing memo.
     Chain.Cardano,
   ],
 } as const
--- a/packages/core/chain/swap/native/NativeSwapChain.ts
+++ b/packages/core/chain/swap/native/NativeSwapChain.ts
@@ -50,6 +50,14 @@ export const nativeSwapEnabledChainsRecord = {
     Chain.Bitcoin,
     Chain.Arbitrum,
     Chain.Zcash,
+    // Live Available ADA.ADA pool confirmed on mayanode (~4,795 ADA,
+    // mayanode.mayachain.info/mayachain/pools). getNativeSwapQuote is a pure
+    // rate/fee GET (no tx bytes built here), so enabling this corridor for
+    // quote lookups carries no fund-handling risk on its own. The signable
+    // Cardano send-with-memo transaction is built downstream (outside this
+    // swap/ module, in the generic Cardano send-tx compiler) — NOT traced or
+    // verified as part of this change. See PR description for what remains.
+    Chain.Cardano,
   ],
 } as const
 
@@ -79,6 +87,7 @@ export const nativeSwapChainIds = {
   [Chain.Solana]: 'SOL',
   [Chain.Tron]: 'TRON',
   [Chain.Noble]: 'NOBLE',
+  [Chain.Cardano]: 'ADA',
 } satisfies Record<NativeSwapEnabledChain, string>
 export type NativeSwapChainId = (typeof nativeSwapChainIds)[NativeSwapEnabledChain]
 
--- a/packages/core/chain/swap/native/api/getNativeSwapQuote.ts
+++ b/packages/core/chain/swap/native/api/getNativeSwapQuote.ts
@@ -106,6 +106,20 @@ const assertOkQuote = (
   return result
 }
 
+// Cardano is enabled as a MayaChain source at the enabled-chains/quote level
+// (NativeSwapChain.ts — live pool confirmed on mayanode), but unlike every
+// other native-swap source chain here, the SDK's own buildSwapKeysignPayload
+// -> getCardanoSigningInputs path IS already wired to build a fully signable
+// Cardano deposit tx from this quote (correct decimals, inbound address, and
+// the swap memo carried into CIP-20 auxiliary data, label 674). The one
+// unverified link is protocol-level: whether MayaChain's Cardano bifrost
+// observer actually reads that CIP-20 label as the routing memo. Until a real
+// deposit confirms that round-trip, reject BEFORE the network call — mirrors
+// getSwapKitQuote's SWAP_SOURCE_TX_BUILD_UNSUPPORTED guard for the SwapKit
+// Sui/Cardano corridors, restoring the "quote groundwork, not yet signable"
+// invariant this PR intends for both providers symmetrically.
+const NATIVE_SWAP_SOURCE_SIGNING_UNVERIFIED = new Set<Chain>([Chain.Cardano])
+
 export const getNativeSwapQuote = async ({
   swapChain,
   destination,
@@ -117,6 +131,13 @@ export const getNativeSwapQuote = async ({
   referral,
   nativeAffiliateConfig,
 }: GetNativeSwapQuoteInput): Promise<NativeSwapQuote> => {
+  if (NATIVE_SWAP_SOURCE_SIGNING_UNVERIFIED.has(from.chain)) {
+    throw new Error(
+      `Native ${from.chain} source swaps are not yet supported for signing (quote-only for now). ` +
+        'Try a different source chain, or swap the other direction.'
+    )
+  }
+
   assertValidSlippageToleranceBps(slippageToleranceBps)
 
   const [fromAsset, toAsset] = [from, to].map(asset => toNativeSwapAsset(asset))
--- a/.changeset/sui-cardano-swapkit-maya-source.md
+++ b/.changeset/sui-cardano-swapkit-maya-source.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Unblock Sui and Cardano as SwapKit/MayaChain swap sources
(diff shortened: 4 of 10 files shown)


packages/core/chain/swap/native/NativeSwapChain.ts
```diff
@@ -50,6 +50,14 @@ export const nativeSwapEnabledChainsRecord = {
...
     Chain.Arbitrum,
     Chain.Zcash,
+    // Live Available ADA.ADA pool confirmed on mayanode (~4,795 ADA,
+    // mayanode.mayachain.info/mayachain/pools). getNativeSwapQuote is a pure
+    // rate/fee GET (no tx bytes built here), so enabling this corridor for
+    // quote lookups carries no fund-handling risk on its own. The signable
+    // Cardano send-with-memo transaction is built downstream (outside this
+    // swap/ module, in the generic Cardano send-tx compiler) — NOT traced or
+    // verified as part of this change. See PR description for what remains.
+    Chain.Cardano,
```
NeOMakinG, line 64: `preferably-blocking:` Unlike the SwapKit source (which gets an explicit pre-network reject in `getSwapKitQuote`), this native corridor has **no** guard — yet it is signable end-to-end via the SDK. After this line, `findSwapQuote({from: Cardano, to: ETH})` returns a usable MayaChain quote, and `buildSwapKeysignPayload` builds a valid Cardano deposit (native-decimal amount, `quote.inbound_address`, memo → CIP-20 aux data via `buildCip20AuxData`). So the corridor is NOT "blocked at the signing layer" as the PR states — the only unverified link is whether MayaChain's Cardano observer reads the CIP-20 memo label, and nothing gates that. A consumer on this SDK using the standard find→build→sign flow could let a user deposit ADA with an unreadable memo (refund at best, lost funds at worst). Restore symmetry: add an explicit "quote-only, not yet signable" reject (mirroring `SWAP_SOURCE_TX_BUILD_UNSUPPORTED`) or a feature flag until a real ADA→Maya deposit confirms the CIP-20 label round-trips.

packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts
```diff
@@ -705,6 +723,13 @@ export const getSwapKitQuote = async ({
   affiliateBps,
   slippage = 3,
 }: Input): Promise<GeneralSwapQuote> => {
+  if (SWAP_SOURCE_TX_BUILD_UNSUPPORTED.has(from.chain)) {
```
NeOMakinG, line 726: `praise:` Correct fail-closed design — rejecting Sui/Cardano source before the `/v3/swap` round-trip, backed by a test asserting `fetch` is never called. Combined with Sui being absent from every native/aggregator record, this makes the Sui source direction genuinely inert (declines with `NoSwapRoutesError`). The gap is only that the native MayaChain path (NativeSwapChain.ts) has no equivalent backstop.

NeOMakinG (commented):
Reviewed as a fund-safety pass on a FUND-CRITICAL SDK, worst-case adversarial lens: does "groundwork only" actually flip a runtime path where a user could now produce a signable-but-unverified Sui/Cardano swap tx? Two-lane (Claude source-trace + Codex gpt-5.4). **DRAFT → COMMENT only, no approval.**

The SwapKit half of this PR is exemplary fail-closed engineering. But the two enabled corridors are **not symmetric**: Sui is genuinely inert, whereas the native Cardano→MayaChain corridor is fully wired and signable end-to-end through the SDK's own keysign path — which contradicts the PR's stated "still blocked at the signing layer" safety model.

## testing approach

- Isolated worktree at HEAD `4d12834a`, `yarn install --immutable`, built `@vultisig/core-config`.
- Changed test files: `getSwapKitQuote.test.ts` 23/23 pass; `findSwapQuote.selection.test.ts` 48/48 pass (incl. the 5 new tests). Full swap suites (`chain/swap` + `mpc/swap`): **33 files / 335 tests green, zero regressions**.
- Traced the full native Cardano→MayaChain path in source: `findSwapQuote` native fetcher → `getNativeSwapQuote` → `buildSwapKeysignPayload` → `getKeysignUtxoInfo(Chain.Cardano)` → `getCardanoSigningInputs` (memo → CIP-20 aux data).
- Codex gpt-5.4 lane exit 0; its one blocker re-verified against source and **rejected as a false positive** (see risk).
- No `package.json`/lockfile change → no caret-bump hazard. Confirmed.

## preferably-blocking

- **Native Cardano→MayaChain is reachable AND signable end-to-end via the SDK, with no guard — the PR's "blocked at the signing layer" claim does not hold for this corridor.** The SwapKit source gets an explicit pre-network reject (`SWAP_SOURCE_TX_BUILD_UNSUPPORTED`), so even a generic consumer flow declines. The native path gets **no** such backstop. After this PR, `findSwapQuote({from: Cardano, to: ETH})` returns a usable MayaChain quote, and the SDK's own `buildSwapKeysignPayload` will happily build a signable Cardano deposit: correct decimals (`toChainAmount(amount, fromCoin.decimals=6)`), inbound address from `quote.inbound_address`, and the swap memo carried into CIP-20 auxiliary data (`buildCip20AuxData`). Every piece the "not traced/verified" note worries about is in fact already wired. The only genuinely-unverified link is protocol-level — whether MayaChain's Cardano inbound observer reads the SDK's CIP-20 metadata label as the routing memo — and there is nothing gating that. A consumer that bumps to this SDK and uses the standard `findSwapQuote → build → sign` flow would let a user deposit ADA to MayaChain with a memo that may not be read → refund at best, stuck/lost funds at worst. Recommend restoring symmetry: gate the native Cardano→MayaChain corridor behind an explicit "quote-only, not yet signable" reject (mirroring the SwapKit guard) **or** a feature flag, until a real mainnet/testnet ADA→Maya deposit confirms the CIP-20 label round-trips. That preserves the "groundwork only" invariant the SwapKit half already upholds.

## suggestion

- Add a byte-level test on the native Cardano→MayaChain path proving the swap memo lands in the signed tx's CIP-20 metadata (label + bytes), analogous to the existing `getCardanoSigningInputs` per-UTXO byte-parity test. That converts the "not verified" note into a regression-guarded assertion and de-risks the corridor for whoever removes the guard later.

## q

- What MayaChain metadata label does the Cardano bifrost observer read the swap memo from, and does it match the SDK's `buildCip20AuxData` label (CIP-20 = 674)? This is the single load-bearing unknown for the corridor above; resolving it decides whether the guard can be dropped.

## risk

- **Codex decimals "blocker" — investigated, false positive.** Codex flagged that `getNativeSwapDecimals(Cardano)` returns 8 (THOR default) rather than lovelace's 6, alleging a 100x quote-vs-signing mismatch. Re-read the source: THORChain/MayaChain use a **universal 1e8 protocol precision for all non-Maya external assets** — `getNativeSwapDecimals` special-cases *only* MayaChain-native coins, so ETH (18-native), BTC, Dash, Zcash, Arbitrum already all resolve to 8 for the quote basis while signing uses their native decimals. `nativeSwapAmountToCoinBaseUnit` rebases 8→native on the way back symmetrically. ADA at 8-quote / 6-sign is the identical, already-in-production pattern, not an error. Both lanes independently confirmed no memo-drop and no quote-race crash from the new reject.
- **Sui is genuinely inert** — SwapKit rejects it before any network call, and Sui is in no native record nor any other provider's enabled-chains, so `findSwapQuote({from: Sui})` fully declines with `NoSwapRoutesError`. No signable Sui path exists. This half is correct groundwork.
- `getSwapTrackingUrl` `satisfies Record` → `Partial<Record>` is **necessary and safe**: adding Sui/Cardano to `swapKitSourceChains` would otherwise break exhaustiveness at compile time; the runtime fallback degrades to a block-explorer link with a `console.warn`, no throw, and Sui/Cardano can never reach the tracker anyway (SwapKit rejects them).

## verdict

**COMMENT** (draft — no approval by policy). The SwapKit guard, Sui inertness, tracker change, decimals handling, and test coverage are all correct. The one thing to resolve before this reaches users is the asymmetric fail-open on the native Cardano→MayaChain corridor: it is signable today despite the description implying otherwise. Add a symmetric guard/flag and verify the CIP-20↔MayaChain memo round-trip.

---
🤖 reviewer lane · Claude + Codex gpt-5.4 · worktree tests green (335/335 swap) · not device-driven (SDK library review)

packages/core/chain/swap/native/api/getNativeSwapQuote.ts
```diff
@@ -117,6 +131,13 @@ export const getNativeSwapQuote = async ({
   referral,
   nativeAffiliateConfig,
 }: GetNativeSwapQuoteInput): Promise<NativeSwapQuote> => {
+  if (NATIVE_SWAP_SOURCE_SIGNING_UNVERIFIED.has(from.chain)) {
```
NeOMakinG, line 134: `praise:` This closes the r1 fund-safety gap exactly as recommended — a pre-network reject on native Cardano source that mirrors `getSwapKitQuote`'s `SWAP_SOURCE_TX_BUILD_UNSUPPORTED` guard, so both providers are now symmetric (quote-only, not yet signable). `getNativeSwapQuote` is the sole producer of a `NativeSwapQuote`, and `findSwapQuote` degrades the throw via `Promise.allSettled` → `NoSwapRoutesError`, so the corridor fails closed. Nice touch asserting `queryUrl` is never called in the new test.

NeOMakinG (approved):
r2 re-review of `63110ecd` (now ready / draft=false). The PR's own change is 5 source files; the large cosmos dynamic-gas-estimation churn in the range diff is main-rebase, out of scope. Two-lane: Claude (primary, load-bearing) + Codex gpt-5.4 (second opinion). GLM retired.

## testing approach
- Fresh worktree at `63110ecd`, `yarn install --immutable`, built `@vultisig/core-config`, ran via root `.config/vitest.core.config.ts`.
- Suites: `getSwapKitQuote.test.ts` + `findSwapQuote.selection.test.ts` + `getNativeSwapQuote.test.ts` → **82/82 green** (incl. the new Cardano-source rejection test).
- Traced the guard chokepoint: `getNativeSwapQuote` is the SOLE non-test producer of a `NativeSwapQuote` (only caller is `findSwapQuote.ts:508`), and `findSwapQuote` degrades a fetcher throw via `Promise.allSettled` → `NoSwapRoutesError` (fail-closed).
- Re-verified decimals (universal 1e8 THOR/Maya protocol basis — NOT a mismatch, as re-confirmed in r1).

## prior issues — status
| r1 | 1 | preferably-blocking (fund-safety) | Native Cardano→MayaChain reachable AND signable end-to-end (findSwapQuote → getNativeSwapQuote → buildSwapKeysignPayload → getCardanoSigningInputs → CIP-20 memo) with NO guard, contradicting the "still blocked at signing" claim — user could deposit ADA to a Maya inbound whose CIP-20 label the observer may not honor | ✅ **fixed** | `63110ecd` adds `NATIVE_SWAP_SOURCE_SIGNING_UNVERIFIED = Set([Chain.Cardano])` in `getNativeSwapQuote.ts:121`, throwing BEFORE the network call — mirrors `getSwapKitQuote`'s `SWAP_SOURCE_TX_BUILD_UNSUPPORTED` guard exactly as r1 recommended. Sole producer, fail-closed via allSettled. New test asserts rejection + `queryUrl` never called. |
| r1 | 2 | (rejected) | Codex "getNativeSwapDecimals(Cardano) returns 8 not 6 = 100x mismatch" | ℹ️ **not re-raised** | Verified FALSE in r1 (universal 1e8 protocol precision for all non-Maya assets; symmetric rebase). Not a bug. |

## blocking
None.

## should-fix
- **Defense-in-depth: the reject lives only at quote-acquisition, not at the prep/build boundary** (Codex). `SwapService.prepareSwapTx` validates only `expiresAt` and forwards a caller-supplied `swapQuote.quote` into `buildSwapKeysignPayload`, which has no source-chain denylist on its `native` arm — so a fabricated/rehydrated native Cardano quote could reach `getCardanoSigningInputs` without touching the new guard. This is **not an open fund-loss path for any realistic flow** (the sole in-repo quote producer now throws; no valid Cardano native quote has ever been produced to cache, since Cardano wasn't in `nativeSwapEnabledChainsRecord[MayaChain]` before this PR; and the agent/consumer flow gets quotes from the SDK, it doesn't hand-construct MayaChain routing internals). It is also **symmetric with the pre-existing SwapKit guard**, which likewise gates at `getSwapKitQuote` and not at build. Still worth a follow-up: a single source-chain denylist at `buildSwapKeysignPayload` would harden BOTH the SwapKit Sui/Cardano corridors and native Cardano at the last mile before signing. Not blocking r2.

## q
- Any plan/tracking issue for the live ADA→Maya deposit that confirms the CIP-20 label-674 round-trip and lifts this guard? The guard comment says "until a real deposit confirms" — a linked bead would make the removal criteria explicit.

## risk
Low. Both newly-enabled source corridors (Sui, Cardano) decline at the quote layer with clear errors, fail-closed. No user-facing behavior change vs. pre-PR (all four combinations still decline — only the error copy is more specific). Decimals/fee unchanged. `getSwapTrackingUrl` `Partial` change degrades safely to a block-explorer link + warn. No changeset added by this PR (release-hygiene note — behavior is inert to consumers so not blocking, but a `@vultisig/sdk` patch changeset would be conventional).

## verdict
**Approve.** The r1 fund-safety concern is genuinely resolved: native Cardano→MayaChain is now blocked-until-confirmed at the quote chokepoint, symmetric with the SwapKit guard and fail-closed, with a real failure-branch test. Sui + Cardano-via-SwapKit remain inert. Codex's build-boundary observation is legitimate defense-in-depth (and applies equally to the already-merged SwapKit guard), tracked as a should-fix follow-up rather than a blocker.

---
🤖 dual-lane review: Claude (primary) + Codex gpt-5.4 (adversarial second opinion)

PR #1014, opened 2026-07-07
feat(cosmos): initiator-side dynamic gas — simulate native sends and relay CosmosSpecific.gas_limit

Ehsan-saradar:
## What

Initiator-side dynamic gas for Cosmos native sends. `getCosmosChainSpecific` now simulates a native bank send via `/cosmos/tx/v1beta1/simulate` and relays the padded gas limit to co-signers in `CosmosSpecific.gas_limit`.

- Padding is `ceil(gas_used × 1.3)` in **exact integer math** (`× 13 / 10`), matching the cosmjs default multiplier and the iOS `CosmosGasEstimator`. The relayed value is part of the SignDoc every device hashes, so integer math + identical honoring keeps cross-device co-signing byte-identical.
- The signing-inputs resolver reads the relayed limit and scales the fee amount proportionally; when it is absent or `0`, every device falls back to the static per-chain limit (fully backward compatible).
- **Native bank sends only.** A relayed dapp `signData`, token / IBC / contract / staking txs, and vault-based chains (THOR/MAYA) keep the static limit.
- **Fails closed.** Any simulate/build error leaves the field unset, so simulation never blocks signing and peers converge on the static limit.

## Consistency refactor

Extracted `resolveCosmosGasFee` (in `core-chain`), now shared by the signing-inputs resolver (what gets signed) and the fee-display resolver (what the user sees), so the shown Network Fee can never drift from the signed fee. Behavior-preserving for the non-dynamic path.

## Tests

- `scaleCosmosGasLimit` / `estimateCosmosGasLimit` — scaling math + fail-closed paths.
- `resolveCosmosGasFee` — fallback, no-scale-at-or-below-static, ceiling scaling.
- Full core suite green (`test:core`), typecheck + lint clean.

## Changeset

`@vultisig/core-chain` + `@vultisig/core-mpc` patch.

## Summary by CodeRabbit

* **New Features**
  * Added initiator-side dynamic Cosmos gas estimation for native bank sends by simulating transactions and applying a 1.3 safety padding, with fail-closed fallback to static gas limits on any simulation/build issues.
  * Updated Cosmos fee resolution to use the resolved (relayed when available) gas limit and to scale fees with ceiling rounding.
  * Enhanced Terra Classic USTC stability-tax burn surcharge with fail-open behavior on LCD/tax lookup errors.
* **Packaging**
  * Exposed new Cosmos gas-fee and gas-estimation entry points via package subpath exports.
* **Tests**
  * Added/expanded coverage for gas scaling/rounding, simulation byte building, estimation behavior (including failure cases), gating logic, and fee/gas resolution.


gomesalexandre (changes requested):
blocking
- this turns initiator-side dynamic cosmos gas on by default and asserts every peer will honor relayed `CosmosSpecific.gas_limit`, but the load-bearing evidence is missing in this repo. `cosmosGasLimitRecord.ts` still documents mobile clients hardcoding the static gas limit byte-for-byte for cross-device sign-doc parity, while the changeset now claims peers honor the relayed limit identically. if a currently shipped iOS or Android co-signer still ignores `gas_limit`, mixed-version signing diverges and the tx cannot co-sign. please either prove the already-shipped mobile clients honor `CosmosSpecific.gas_limit` for native cosmos sends, or gate this behind a rollout flag.

should-fix
- `getCosmosChainSpecific`'s new `isNativeSend` gate has no real coverage. nothing proves `gasLimit` is actually set for a successful native bank send, or that it stays unset for token / ibc / contract / relayed-signData / zero-amount cases. the new estimator unit tests call `estimateCosmosGasLimit` directly and bypass the branch that decides live tx behavior.
- `buildSimulateTxBytes` is the riskiest new code path and is still fully mocked in tests. a malformed wallet-core compile / parse path would fail closed and silently disable dynamic gas fleet-wide with no red test. please add at least one direct test around real tx-bytes assembly or the parser failure surface.

suggestion
- `keysignPayload.signData.case` should be guarded with optional chaining. the new resolver test comment says real protobuf payloads always carry the wrapper, but this line sits outside the `attempt()` fail-closed block, so a malformed caller that omits `signData` would now throw before you even reach the fallback path.

verdict
changes requested

gomesalexandre (approved):
All three blocking/should-fix points addressed:

**Cross-client compat (was blocking)**: Ehsan's comment confirms iOS #4692 + Android #5191 are already merged — both clients do `gasLimit ?? staticLimit` / `atomData.gasLimit?.takeIf { it > 0 } ?: static`, so the relayed limit is honored and the fallback is safe on old builds. That's the evidence I needed.

**Test coverage (was should-fix)**: `nativeSendGating.test.ts` (142 lines) locks in the `isNativeSend` gate across native/token/IBC/signData/zero-amount cases. `buildSimulateTxBytes.test.ts` (79 lines) exercises real WalletCore — not mocked — exactly what I asked for.

**Optional chaining (was suggestion)**: `keysignPayload.signData?.case` in place, comment even documents why.

gucci now ser, ship it.

PR #1015, opened 2026-07-07
feat(sdk): fold LLM chain-alias tolerance into normalizeChain

gomesalexandre:
Structural follow-on to the LUNC<->USTC bug (candidate #1 from the SDK-consolidation spike, highest bug-prevention/effort). SDK-only, additive.

## what

The LUNC<->USTC incident traced back to THREE independent, divergent reimplementations of "make an LLM-emitted chain string canonical":

1. agent-backend-ts's `zodHelpers.ts` `chainString()` (SDK `normalizeChain` + a rich alias table + space-stripping) - the richest, but bypassable (movement tools skip `inputSchema.parse`, the LUNC<->USTC bug being fixed separately in abts right now).
2. agent-backend-ts's `resolveToken.ts` `normalizeChainInput` - a thin `sdkNormalizeChain` wrapper with NO separator tolerance. Different tolerance than #1.
3. the SDK's own `normalizeChain` - the weakest of the three (exact-match only, no space-stripping fallback).

The divergence between #1 and #2 (one tolerates a form the other doesn't) is exactly what let the split-brain bug through.

This PR enriches the SDK's `normalizeChain` (`packages/sdk/src/utils/normalizeChain.ts`) to be the single canonical LLM-tolerant normalizer, folding in the separator-stripping fallback + explicit chain-id/marketing-name alias table currently living in abts's `zodHelpers.ts:135-195`.

## how

The existing ticker/hand-curated/canonical alias table is untouched - exact-match stays the fast, backward-compatible first attempt. A second table is auto-derived from every entry in that table by stripping spaces/underscores/hyphens, using the SAME single-owner claim mechanism already used for ticker aliases: a stripped key that would resolve to more than one distinct chain is dropped rather than guessed (fund-safety - a wrong collision could misroute a swap/send).

This means `"Bitcoin Cash"` / `"Cronos Chain"` / `"THOR Chain"` / `"ZK Sync"` / `"Terra Classic"` / `"Binance Smart Chain"` all resolve automatically from their EXISTING un-spaced aliases or canonical forms - zero per-chain maintenance needed. The only hand-curated additions are the three chain-id/marketing names that have no separator-free form anywhere else in the table: `columbus-5` -> TerraClassic, `phoenix-1` -> Terra, `"Terra V2"` -> Terra (kept in sync with abts's `CHAIN_ALIAS_TABLE`).

## what I deliberately did / didn't include

- Included: every alias/separator case from abts's `zodHelpers.ts` `chainString()` fallback path (`stripChainSeparators` + `CHAIN_ALIAS_TABLE`).
- Deliberately NOT included: mcp-ts's own copy of this same alias table (zodHelpers.ts's comments reference it as a second, currently-unconsolidated vendored copy, PR #753) - out of scope here, a further follow-on.
- Deliberately NOT touching abts's `zodHelpers.ts` or `resolveToken.ts` in this PR (see scope discipline below).

## scope discipline

Per the task brief: this PR is SDK-only and purely additive. It does NOT touch agent-backend-ts's `zodHelpers.ts` or `resolveToken.ts` - de-duplicating those call sites (deleting the now-redundant `CHAIN_ALIAS_TABLE`/`stripChainSeparators` fallback and delegating straight to this enriched SDK function) is an intentional SEPARATE follow-on PR, to land AFTER both (a) the in-flight abts LUNC<->USTC canonicalization fix and (b) this PR - avoiding merge conflicts and keeping each PR independently reviewable.

## test matrix

`packages/sdk/tests/unit/utils/normalizeChain.test.ts`:
- every existing canonical/alias/whitespace/error case unchanged and still passing (no regression)
- separator-insensitive natural-language phrasings: Terra Classic, Bitcoin Cash, Cronos Chain, THOR Chain, ZK Sync, Maya Chain, Binance Smart Chain, each across space/underscore/hyphen/casing variants
- chain-id/marketing-name aliases: columbus-5, phoenix-1, "Terra V2" across casing/separator variants
- fund-safety collision guard: every `Chain` enum value round-trips to itself (canonical/lower/upper case), and Terra/TerraClassic are asserted to never resolve to each other despite sharing an address and name prefix
- still fails closed for genuinely unknown/ambiguous phrases

167 SDK unit test files / 2373 tests passing (`yarn test:unit`), `yarn typecheck` clean, lint clean.

🤖 Agent-generated

### DIFF
--- a/packages/sdk/src/utils/normalizeChain.ts
+++ b/packages/sdk/src/utils/normalizeChain.ts
@@ -61,6 +61,55 @@ for (const value of Object.values(Chain)) {
   aliasToChain[value.toLowerCase()] = value
 }
 
+/** Lowercase + strip spaces/underscores/hyphens, e.g. "Terra Classic" -> "terraclassic". */
+function stripSeparators(key: string): string {
+  return key.replace(/[\s_-]+/g, '')
+}
+
+/**
+ * Separator-stripped fallback table (LLM-tolerance layer, consolidated from
+ * agent-backend-ts's `zodHelpers.ts` `chainString()` — the fix for the
+ * LUNA<->LUNC "no route" bug: deepseek emits `to_chain: "Terra Classic"`
+ * (with a space) routinely, and the exact-match table above rejects it
+ * outright since neither `"terra classic"` nor any hand-curated alias
+ * carries a literal space).
+ *
+ * Derived automatically from every entry already in `aliasToChain` (so
+ * "Bitcoin Cash" / "Cronos Chain" / "THOR Chain" / "ZK Sync" all resolve via
+ * their EXISTING un-spaced aliases/canonical forms — no per-chain
+ * maintenance needed), PLUS a small hand-curated set of chain-id / marketing
+ * names that don't reduce from any existing key at all (`columbus-5`,
+ * `phoenix-1`, `"Terra V2"`).
+ *
+ * Same ownership-claim safety as the ticker table above: a stripped key that
+ * would resolve to more than one DISTINCT chain is dropped rather than
+ * guessing — fund-safety (a wrong collision could misroute a swap/send to
+ * the wrong chain), never silently pick a side.
+ */
+const strippedAliasToChain: Record<string, Chain> = {}
+const strippedOwnersByAlias = new Map<string, Set<Chain>>()
+const claimStripped = (alias: string, chain: Chain) => {
+  const key = stripSeparators(alias.toLowerCase())
+  const owners = strippedOwnersByAlias.get(key) ?? new Set<Chain>()
+  owners.add(chain)
+  strippedOwnersByAlias.set(key, owners)
+}
+for (const [alias, chain] of Object.entries(aliasToChain)) {
+  claimStripped(alias, chain)
+}
+// Chain-id / marketing-name aliases that carry no separator-free form anywhere above.
+// Kept in sync with agent-backend-ts's zodHelpers.ts CHAIN_ALIAS_TABLE.
+claimStripped('columbus-5', Chain.TerraClassic) // Terra Classic's chain-id
+claimStripped('phoenix-1', Chain.Terra) // Terra v2's chain-id
+claimStripped('terra v2', Chain.Terra) // common marketing name for post-fork Terra
+for (const [key, owners] of strippedOwnersByAlias) {
+  if (owners.size !== 1) continue
+  for (const only of owners) {
+    strippedAliasToChain[key] = only
+    break
+  }
+}
+
 /**
  * Resolve a case-insensitive chain string (canonical name, ticker, or common
  * alias) to the SDK's canonical Chain value.
@@ -69,10 +118,18 @@ for (const value of Object.values(Chain)) {
  * - Canonical Chain enum values (`"Ethereum"`, `"Bitcoin"`, `"Bitcoin-Cash"`, ...)
  * - Tickers (`"btc"`, `"eth"`, `"sol"`, ...)
  * - Common aliases (`"bitcoin"`, `"binance"`, `"thorchain"`, ...)
+ * - Space/underscore/hyphen-insensitive natural-language phrasings LLMs
+ *   routinely emit (`"Terra Classic"`, `"Bitcoin Cash"`, `"Cronos Chain"`,
+ *   `"THOR Chain"`, `"ZK Sync"`) and chain-id / marketing-name forms
+ *   (`"columbus-5"`, `"phoenix-1"`, `"Terra V2"`)
  *
  * Accepts nullish input because this utility is commonly called with
  * LLM-sourced strings that may be missing — such inputs throw
  * `UnknownChainError` with a blank input rather than a `TypeError`.
+ *
+ * This is the SINGLE canonical LLM-tolerant chain normalizer — callers
+ * needing chain-string tolerance should use this directly rather than
+ * layering their own separator-stripping / alias fallback on top of it.
  *
  * @throws {UnknownChainError} When the input cannot be resolved.
  */
@@ -80,6 +137,10 @@ export const normalizeChain = (input: string | null | undefined): Chain => {
   const key = input?.trim().toLowerCase() ?? ''
   const resolved = aliasToChain[key]
   if (resolved) return resolved
+
+  const stripped = stripSeparators(key)
+  const strippedResolved = strippedAliasToChain[stripped]
+  if (strippedResolved) return strippedResolved
 
   const known = Object.values(Chain).map(c => c.toLowerCase())
   throw new UnknownChainError(key, known)
--- a/packages/sdk/tests/unit/utils/normalizeChain.test.ts
+++ b/packages/sdk/tests/unit/utils/normalizeChain.test.ts
@@ -211,4 +211,110 @@ describe('normalizeChain', () => {
       expect(normalizeChain('usdc')).toBe(Chain.Noble)
     })
   })
+
+  // Consolidated from agent-backend-ts's zodHelpers.ts chainString() — the fix for the
+  // LUNA<->LUNC "no route" bug (deepseek emits `to_chain: "Terra Classic"` with a space
+  // ~2/3 of the time, which the pre-consolidation SDK rejected outright). These natural-
+  // language, space/underscore/hyphen-insensitive forms must resolve WITHOUT any caller-side
+  // fallback layer — the SDK is now the single source of this tolerance.
+  describe('separator-insensitive natural-language phrasings (LLM tolerance)', () => {
+    it.each([
+      ['Terra Classic', Chain.TerraClassic],
+      ['terra classic', Chain.TerraClassic],
+      ['TERRA CLASSIC', Chain.TerraClassic],
+      ['terra_classic', Chain.TerraClassic],
+      ['terra-classic', Chain.TerraClassic],
+      ['Bitcoin Cash', Chain.BitcoinCash],
+      ['bitcoin cash', Chain.BitcoinCash],
+      ['BITCOIN CASH', Chain.BitcoinCash],
+      ['bitcoin_cash', Chain.BitcoinCash],
+      ['Cronos Chain', Chain.CronosChain],
+      ['cronos chain', Chain.CronosChain],
+      ['THOR Chain', Chain.THORChain],
+      ['thor chain', Chain.THORChain],
+      ['Thor Chain', Chain.THORChain],
+      ['ZK Sync', Chain.Zksync],
+      ['zk sync', Chain.Zksync],
+      ['Maya Chain', Chain.MayaChain],
+      ['maya chain', Chain.MayaChain],
+      ['Binance Smart Chain', Chain.BSC],
+      ['binance smart chain', Chain.BSC],
+    ])('resolves separator-insensitive "%s" to %s', (input, expected) => {
+      expect(normalizeChain(input)).toBe(expected)
+    })
+
+    it('still resolves the exact hyphenated canonical form directly (no regression)', () => {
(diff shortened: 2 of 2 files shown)


packages/sdk/src/utils/normalizeChain.ts
```diff
@@ -61,6 +61,55 @@ for (const value of Object.values(Chain)) {
...
+const strippedOwnersByAlias = new Map<string, Set<Chain>>()
+const claimStripped = (alias: string, chain: Chain) => {
+  const key = stripSeparators(alias.toLowerCase())
+  const owners = strippedOwnersByAlias.get(key) ?? new Set<Chain>()
+  owners.add(chain)
+  strippedOwnersByAlias.set(key, owners)
+}
+for (const [alias, chain] of Object.entries(aliasToChain)) {
+  claimStripped(alias, chain)
+}
```
NeOMakinG, line 99: `should-fix:` this loop re-keys **every** exact alias into the stripped fallback, including fee-coin tickers. Effect: asset symbols gain separator-variant tolerance they lacked on main — e.g. `usdc` → `u s d c` / `u-s-d-c` / `u_s_d_c` all now resolve to Noble (each threw on main), and short tickers similarly: `o p`→Optimism, `z k`→Zksync, `b c h`→Bitcoin-Cash, `s o l`→Solana.

This is **not** a new wrong-chain (base `usdc`→Noble already exists + is documented/tested, and my probe confirms zero NEW single-owner targets beyond your 3 intended Terra adds), and it fails-closed everywhere else. But it widens the "asset symbol emitted in the chain slot" surface beyond the stated "natural-language phrasings LLMs routinely emit" intent — a ticker split into single chars is not natural language.

Suggested tightening if you want to hold the line to intent: derive stripped keys only from aliases that actually CONTAIN a separator (or are multi-char non-ticker forms), rather than from the whole `aliasToChain` incl. bare tickers. Non-blocking — flagging for a deliberate decision.

NeOMakinG (commented):
Reviewed as a fund-safety-critical change to `normalizeChain`, the load-bearing chain-resolution primitive: a wrong alias→chain mapping means a tx could be built/signed on the wrong network. Draft PR, so this is a **COMMENT** (no approval on drafts). Ran a two-lane audit (Claude load-bearing + Codex gpt-5.4 adversarial second-opinion).

## testing approach
- Worktree at HEAD `3b5836c3`, `yarn install` clean.
- Ran the target suite: **`normalizeChain.test.ts` — 215/215 green** (the 29 other unrelated failures in the full `test:unit` run are pre-existing `ERR_MODULE_NOT_FOUND` for unbuilt `@vultisig/core-config`/`@vultisig/mpc-types` dist, not touched by this PR).
- Built an **introspection probe** rebuilding both tables from the REAL `Chain` enum + `chainFeeCoin` tickers to dump every stripped key and its owner set.
- Ran an **adversarial input probe** against the compiled module (asset-in-chain-slot, short-ticker separator splits, Terra/TerraClassic disambiguation, fail-closed cases).

Key probe results (fund-safety invariants all HOLD):
- **Zero dropped keys** — the collision-guard `owners.size !== 1` drop never fires; the auto-derivation produces NO cross-wired stripped forms. No two distinct chains collapse to one.
- **Round-trip clean** — every `Chain` value (canonical/lower/upper) resolves to ITSELF; no canonical input regresses.
- **Only 3 genuinely-new resolution targets** from the auto-derivation: `columbus5`→TerraClassic, `phoenix1`→Terra, `terrav2`→Terra — exactly the 3 intended hand-curated adds. Every other stripped key already existed as an exact alias, so the derivation only lets separator-variant INPUT match; it introduces no new canonical target.
- **Fails closed** on unknown/ambiguous: `terra2`, `terra classic 2`, `usdc.e`, `eth-mainnet`, `solana devnet`, `not a chain`, `""` all throw `UnknownChainError`. The strip regex is narrow (`[\s_-]+` only), so `.`/digits are NOT normalized away — no over-eager matching.
- **Terra vs TerraClassic** correctly distinct: `luna`/`terra`/`phoenix-1`/`terra v2`→Terra, `lunc`/`terra classic`/`columbus-5`→TerraClassic. Never cross-resolve.

## blocking
None. No mis-mapped alias, no dangerous collision, no wrong-chain default, no regression, no caret-bumped dep (PR touches ONLY `normalizeChain.ts` + its test — the `package.json` deltas visible in a two-dot diff are origin/main drift, not in the merge-base three-dot PR diff).

## should-fix
See inline: the auto-derivation sweeps **every** exact alias — including fee-coin **tickers** — into the separator-stripped fallback, so asset symbols like `usdc` gain separator-variant tolerance (`u s d c` / `u-s-d-c` → Noble) where they threw on main. This is NOT a new wrong-chain (the base `usdc`→Noble mapping already exists on main, documented+tested), and my probe confirms zero NEW misroute targets were created — but it does widen the "asset symbol emitted in the chain slot" surface beyond the PR's stated "natural-language phrasings" intent. Worth a deliberate decision.

## q
- Any consumer (abts/mcp) that passes a *token/asset* string into the chain slot of `normalizeChain`? If so, the widened ticker-separator surface (above) matters more; if chain and asset slots are always distinct in callers, it's cosmetic.

## risk
Low. Purely additive LLM-tolerance layer gated behind exact-match (exact table is consulted first and untouched). Fund-safety guard (single-owner drop) is sound and empirically produces zero collisions on the current chain set. Both directions are test-locked (new aliases resolve right; unknown/ambiguous throw; every enum value round-trips to itself). The only residual is the ticker-separator surface expansion, which does not introduce a new wrong-chain mapping.

## verdict
COMMENT (draft — no approval). Clean, well-guarded, fund-safe. One non-blocking should-fix on ticker-derived separator tolerance. Ready to promote from draft once the ticker-surface question is answered.

---
*Two-lane review: Claude (load-bearing, source-walked + runtime probes) + Codex gpt-5.4 (adversarial second-opinion — surfaced the ticker-separator finding). GLM lane retired.*

NeOMakinG (approved):
**r2 re-review @ `98f7880b`** (ready-for-review; was draft at r1). Fund-critical chain-resolution. **Verdict: APPROVE.**

The delta since my r1 (`3b5836c3`) is a **pure rebase onto main** — it picked up the cosmos dynamic-gas work + version-packages `#1019`, but the PR's own contribution (`git diff origin/main...98f7880b`) is **byte-identical** to r1: `normalizeChain.ts` (+61) + `normalizeChain.test.ts` (+106), no version/caret bump. I re-verified the fund-safety invariants against the **current** `Chain` enum (now 38 values) rather than trusting r1 — because main's enum growth flows through the auto-derivation loop and *could* have introduced a new collision. It did not.

## testing approach
Rebuilt both alias tables from the real `Chain` enum + `chainFeeCoin` tickers at HEAD via an introspection probe (not just tests), plus adversarial input against the compiled module; ran the target suite.
- **Stripped-table collisions: ZERO** — `owners.size !== 1` drop never fires; no two distinct chains collapse to one stripped key.
- **NEW stripped targets = exactly the 3 intended Terra adds** (`columbus5`→TerraClassic, `phoenix1`/`terrav2`→Terra). Main's enum growth introduced no new canonical target.
- **Round-trip CLEAN** — every `Chain` (canonical/lower/upper) resolves to itself; no regression (fallback only fires after exact-match already fails).
- **Terra vs TerraClassic distinct** — `luna`/`terra`/`phoenix-1`/`terra v2`→Terra; `lunc`/`terra classic`/`columbus-5`→TerraClassic. No cross-bleed (shared bech32 prefix does not leak). `ustc`→throws (correctly unwired).
- **Fails closed** — `terra2`, `usdc.e`, `eth-mainnet`, `solana devnet`, `''`, `null`, `undefined` all throw `UnknownChainError`; zero fail-open leaks.
- **Tests: 215/215 green** on `normalizeChain.test.ts`. (The 29 unrelated `test:unit` file failures are the pre-existing `ERR_MODULE_NOT_FOUND` on unbuilt `@vultisig/core-config`/`@vultisig/mpc-types` dist — noted at r1, not this PR.)

Second lane: Codex gpt-5.4 (exit 0) — no blocker, corroborated fund-safety, surfaced 3 non-blocking notes folded in below.

## prior issues — status
| r1 | 1 | should-fix | ticker-separator surface widening — derivation loop sweeps fee-coin **tickers** into the stripped fallback, so asset symbols gain separator-variant tolerance (`u s d c`→Noble, `o p`→Optimism, `s o l`→Solana, `b c h`→Bitcoin-Cash) that threw on main | ❌ still open | rebase-only delta; loop `for (const [alias, chain] of Object.entries(aliasToChain))` unchanged. Re-confirmed at HEAD: still **fails-closed**, still **no new wrong-chain** (`usdc`→Noble is a pre-existing tested tradeoff). Non-blocking — not held against approval. |
| r1 | 2 | q | do abts/mcp callers ever pass an asset string into the chain slot? | ⏭️ unanswered | re-raised below — this is what determines whether #1 ever matters in practice |

## should-fix
Carrying r1 #1 forward (still open, non-blocking). Two Codex notes reinforce it:

- **Cross-layer shadow weakens the collision proof (defense-in-depth).** The stripped ambiguity check iterates `Object.entries(aliasToChain)` — i.e. the *already-deduplicated* exact table, after layer-precedence (ticker → hand-curated → canonical) has collapsed same-key conflicts via last-write-wins. So a future alias that is shadowed at the exact-key level but would only collide *after* stripping would never reach the `owners.size !== 1` guard, and the stripped fallback could look unique instead of being dropped. **Latent, not live** — the probe shows zero drops and a clean round-trip today — but it means the safety guarantee is a snapshot, not structural. Deriving stripped keys from the raw pre-dedup claim sets (or asserting parity in a test) would make it robust to future table edits. This is exactly the class of maintenance bug the ticker-widening in #1 invites.

## suggestion
- **Pin a parity test to `agent-backend-ts`'s `CHAIN_ALIAS_TABLE`.** The new docstring claims this is "the SINGLE canonical LLM-tolerant chain normalizer … kept in sync with agent-backend-ts's zodHelpers.ts" — but nothing in this repo enforces that. SDK and backend can silently drift on accepted chain strings (planner accepts `"Terra Classic"`, executor rejects it, or vice-versa). A shared fixture or a pinned parity test would close the split-brain risk this consolidation is meant to eliminate.
- **`stripSeparators` `/[\s_-]+/g` misses Unicode separators** — zero-width space `​`, non-breaking hyphen `‑`, en-dash `–`, soft hyphen `­`, word-joiner `⁠`. So `THOR‑Chain` (non-breaking hyphen) still throws despite the docstring advertising generic separator-insensitive tolerance. **Fails closed** (safe — narrower is safer here), so this is a stated-behavior gap, not a hazard. Only widen if a real LLM emits these.

## q
Re-raising r1 #2: do any **abts/mcp callers ever pass an asset symbol into the chain slot**? If a planner can emit `to_chain: "USDC"` (or a spaced variant), the widening in should-fix #1 becomes reachable — `usdc`→Noble is already the mapping, but the separator-variant surface is new. If the chain slot is always chain-typed upstream, #1 is inert and can be closed as won't-fix.

## risk
Low. Chain-resolution is fund-critical, but every path here **fails closed** — the fallback only fires after exact-match misses, and any ambiguity is dropped rather than guessed. No new wrong-chain mapping at HEAD; the sole open finding widens a pre-existing, tested tradeoff to separator-variants without changing any target. Rebase introduced no regression.

## verdict
**APPROVE.** Collision-free + fail-safe both re-confirmed at `98f7880b` against the current 38-chain enum; the ticker-separator widening is a confirmed-acceptable, fail-closed non-issue (pre-existing tradeoff, no new misroute), not a blocker. The should-fix + two suggestions are follow-up hardening, not merge-blockers.

---
🤖 dual-lane review (Claude Opus 4.8 primary + Codex gpt-5.4 second-opinion). Introspection-probe evidence rebuilt from the real `Chain` enum at HEAD.

PR #1017, opened 2026-07-07
test(sdk): pin the Kyber stablecoin quote premium to upstream, not decimals

gomesalexandre:
Investigation, not a code fix -- root cause is external to this repo. Fund-safety-relevant (a wrong "you receive" figure), reported per team-lead's ask.

## the finding

QA (live sim, default Go path) reported a suspicious ~10-12% quote premium on same-chain EVM stablecoin swaps via Kyber on Ethereum:
- USDC->DAI: 2.00022 USDC in -> 2.18999 DAI out (+9.5%)
- USDC->USDT: 1.000078 USDC in -> 1.11475 USDT out (+11.5%, reproduced 2x identically)

Stablecoin->stablecoin should be ~1:1 minus bps slippage. The hypothesis was a display/decimals-normalization bug (USDC/USDT are 6-dec, DAI is 18-dec -- a classic scaling-mismatch shape).

## root cause: NOT a decimals bug, real and upstream

Hit KyberSwap's LIVE public aggregator API directly for the exact same pairs/amounts, bypassing this repo (and agent-backend-ts) entirely:

- `2 USDC -> DAI` today: `amountOut 2207447836162962680` (2.2074... DAI), a **+10.3%** premium over `amountInUsd`, routed via a `uniswap-v4-fee` hook pool -> `balancer-v2-weighted`.
- `1.000078 USDC -> USDT` today: `amountOut 1120978` (1.120978 USDT), a **+12%** premium, routed via the SAME `uniswap-v4-fee` hook pool -> a `kyberswap PMM` leg.

This rules out a decimals bug two ways:
1. The raw upstream API response ALREADY carries the premium before any of our code touches it.
2. USDC->USDT is a **same-decimals** pair (6 and 6 on both sides) yet shows the SAME magnitude of premium as the 6-vs-18 DAI pair. A genuine decimals-count mismatch would be off by a factor of 10^12 (astronomically wrong), not a subtle ~10% -- so a uniform ~10-12% premium across both a same-decimals and a cross-decimals pair can't be a decimals bug by construction.

Traced `getKyberSwapQuote` (`packages/core/chain/swap/general/kyber/api/quote.ts` -> `tx.ts`): it relays Kyber's `amountOut` verbatim into `dstAmount`, no rescale. Downstream, this SDK's own keysign `build.ts` and agent-backend-ts's `execute_swap` `formatOut` both correctly scale `dstAmount` via the destination token's OWN decimals (`toCoin.decimals` / `toToken.decimals`). Neither has a bug. The premium is baked into KyberSwap's own route/pricing for this specific pool combination -- almost certainly a live, temporarily-mispriced Uniswap V4 hook pool -- not introduced anywhere in this codebase.

**This is live on the abts path too**: agent-backend-ts's `execute_swap` calls this exact same `getKyberSwapQuote` function against this exact same public Kyber API. There's no code difference between what the Go backend and abts would show for this pair right now.

## what this PR does

Adds two regression tests to `packages/core/chain/swap/general/kyber/api/quote.test.ts`, using the captured live Kyber response data as fixtures (mocking `queryUrl`, following the existing test's own pattern):
- asserts the SDK relays `dstAmount` byte-for-byte, unmodified
- documents the implied rate is genuinely >5% above 1:1 for both pairs -- pinning the anomaly to Kyber's upstream data rather than asserting it away

## follow-up flagged, NOT fixed here (needs design input)

`findSwapQuote`'s `selectBestEligibleQuote` / `getComparableOutputAmount` (`packages/core/chain/swap/quote/findSwapQuote.ts`) picks the candidate with the single highest raw destination amount across ALL competing providers, with NO plausibility ceiling. A live, temporarily-mispriced route from any ONE aggregator (like this Kyber V4-hook pool right now) can outrank a saner quote from 1inch/CowSwap for the exact same pair -- actively surfacing the worse (or likely non-executable, since on-chain reality may not match the quoted route by broadcast time) quote to the user, rather than merely failing to help.

Deliberately NOT changed here: this is the core provider-selection algorithm across EVERY swap pair the SDK supports, not something scoped to stablecoins. A plausibility guard needs its own threshold/design discussion (calibrating what counts as "implausible" for volatile-asset pairs, where a large legitimate difference between aggregators IS expected and should still win) and deserves review before touching a live, heavily-used routing path.

## test plan

`yarn test:core` -- 157 core test files / 1356 tests passing (`.config/vitest.core.config.ts`, the config this file's other Kyber tests already run under). Lint clean.

🤖 Agent-generated


packages/core/chain/swap/general/kyber/api/quote.test.ts
```diff
@@ -73,4 +73,123 @@ describe('getKyberSwapQuote', () => {
...
+    // No decimals bug: our code relays Kyber's own amountOut unmodified.
+    expect(quote.dstAmount).toBe(rawAmountOut)
+
+    // The implied rate is NOT ~1:1 for a stablecoin pair — the premium is
+    // real and comes from Kyber's own quote, not from anything this repo
+    // computes. Documents the anomaly rather than asserting it away.
+    const amountInHuman = 2 // 2.0 USDC
+    const amountOutHuman = Number(rawAmountOut) / 10 ** 18
+    const impliedRate = amountOutHuman / amountInHuman
+    expect(impliedRate).toBeGreaterThan(1.05) // >5% above 1:1 — implausible for USDC->DAI
```
NeOMakinG, line 145: `suggestion:` this `impliedRate` assertion is **documentary-only** and doesn't exercise the SDK. It's computed from the hardcoded `rawAmountOut` constant (not from `quote.dstAmount`), so `Number('2207447836162962680')/10**18 > 1.05` is tautologically true regardless of what the SDK returns. The entire decimals-regression guard is carried by `expect(quote.dstAmount).toBe(rawAmountOut)` on line 137 (which is correctly mutation-proof) — this line adds no coverage.

Two small things if you want it to actually guard:
1. Derive `amountOutHuman` from `Number(quote.dstAmount)`, not the fixture constant, so a rescale bug perturbs it.
2. `Number(rawAmountOut)` here exceeds `Number.MAX_SAFE_INTEGER`, so it's lossy (`2.2074478361629626` vs the true `...626680`). Harmless at the coarse `> 1.05` bound, but don't tighten the threshold later expecting exactness — and an upper bound (`between 1.05 and ~1.5`) would make it catch a 10^12-scale mutation that a lower-bound-only check lets through. Same applies to the USDT twin on line 193.

NeOMakinG (commented):
Two-lane review (Claude load-bearing + Codex `gpt-5.4`, exit 0 — both lanes converged). DRAFT, so **COMMENT only, no approval**.

This is a clean, well-reasoned investigation PR. The test locks the *correct* behavior and is mutation-proof against the exact bug it was written to rule out. Findings below are non-blocking.

## testing approach

- Built `@vultisig/core-config`, ran the target suite under the same config its sibling Kyber tests use (`.config/vitest.core.config.ts`): **3/3 pass** (1 pre-existing + 2 new).
- Traced the relay path in source: `getKyberSwapQuote` -> `getKyberSwapRoute` (mock call #1) -> `getKyberSwapTx` (mock call #2). `dstAmount` is `buildResponse.data.amountOut` relayed verbatim (`tx.ts:120-122`), no rescale/unit-conversion. A rescale regression in `tx.ts` (e.g. `amountOut * 10**to.decimals`) would fail `expect(quote.dstAmount).toBe(rawAmountOut)` in **both** new tests — so the load-bearing assertion is genuinely mutation-proof against a decimals-artifact, not tautological.
- **Premium provenance is upstream, not self-generated**: the expected `dstAmount` values are captured live KyberSwap `/routes`+`/route/build` responses used as fixtures. The test asserts the SDK *relays* Kyber's own figure — the correct invariant to pin. The same-decimals USDC->USDT case (6->6) is the load-bearing evidence that a ~12% premium can't be a decimals-count mismatch (that would be off by 10^12).
- **Fund-relevant surface**: `dstAmount` is the "you receive" figure that flows into keysign build — this is the signable surface, correctly asserted.
- **Fixture-shape fidelity verified** (both lanes): `getKyberSwapRoute` reads only top-level `data.routeSummary`/`data.routerAddress`; `getKyberSwapTx` reads only `code`, `data.amountOut`, `data.data`, `data.gas`. The mocks provide exactly those — no unrealistic shape gets the test to pass.
- No dependency changes (single test file, +119/-0). No caret bump.

## suggestion

The `impliedRate` assertions (`quote.test.ts:145`, `:193`) are documentary-only — see inline.

## q

Is the deferred `selectBestEligibleQuote` plausibility-ceiling follow-up tracked somewhere (bd / issue)? It's the real fund-safety item this investigation surfaces and it'd be a shame to lose it once this PR merges.

## risk

**Latent product bug this PR surfaces (deferred, correctly).** `findSwapQuote`'s `selectBestEligibleQuote` (`findSwapQuote.ts:296`) sets `bestOutputAmount = max(comparable output)` across all providers and bands off it with a ~50 bps preference band, trusting `general.dstAmount` verbatim (`getComparableOutputAmount`, `:252`) with **no plausibility ceiling and no cross-provider sanity check**. A single mispriced aggregator quote overstated by 10-12% doesn't just display wrong — it *sets* `bestOutputAmount`, lands far outside the preference band of the sane 1inch/CowSwap quotes, and thus becomes the *selected route the user signs against* (likely non-executable by broadcast time). Both lanes independently confirmed this is real and is a routing/selection issue, not merely display. The author explicitly and, in my judgment, **correctly deferred** it: a threshold needs design that won't reject legitimately-large spreads on volatile-asset pairs, and this is the core selection algorithm across every pair — not something to bolt on inside a test PR. Calling it out here so it stays visible; it is *not* a blocker on this test-only change.

Risk of this PR itself: **negligible** — test-only, no runtime/dep surface touched.

## verdict

COMMENT (draft — never approve). The test is correct, mutation-proof, and asserts the right fund-relevant surface with genuinely-upstream expected values. No blocking findings. When promoted out of draft, this is an approve modulo the two non-blocking notes.

---
🤖 Two-lane review: Claude (load-bearing) + Codex gpt-5.4. Non-device (test-only PR).

packages/core/chain/swap/general/kyber/api/quote.test.ts
```diff
@@ -73,4 +73,123 @@ describe('getKyberSwapQuote', () => {
...
+        address: '0xsender',
+        id: '0x6B175474E89094C44Da98b954EedeAC495271d0F', // DAI, 18 dec
+        decimals: 18,
+        ticker: 'DAI',
+      },
+      amount: 2_000000n, // 2.0 USDC
+    })
+
+    // No decimals bug: our code relays Kyber's own amountOut unmodified.
+    expect(quote.dstAmount).toBe(rawAmountOut)
```
NeOMakinG, line 137: `suggestion:` this `toBe(rawAmountOut)` is the load-bearing, mutation-proof assertion — it kills any decimals rescale (`amountOut * 10**decimals`) and is exactly the invariant this PR should pin. Nice. One orthogonal hardening (non-blocking): the fixture sets `routeSummary.amountOut` and `build.data.amountOut` to the *same* value, so this would still pass if a regression sourced `dstAmount` from the route summary instead of the build response. Making the two differ and asserting `dstAmount === build.data.amountOut` would close that gap too.

packages/core/chain/swap/general/kyber/api/quote.test.ts
NeOMakinG, line 145: `suggestion:` (carried from r1, non-blocking) `impliedRate` is documentary-only here — it is computed from the hardcoded `rawAmountOut` constant, not from `quote.dstAmount`, so `> 1.05` is tautologically true regardless of SDK behavior. Also `Number(rawAmountOut)` is lossy past `Number.MAX_SAFE_INTEGER`. Deriving it from `quote.dstAmount` and adding an upper bound (e.g. `< 1.5`) would make it actually assert on the SDK output and catch a 10^12 mutation a lower-bound-only lets through.

NeOMakinG (approved):
Round 2 re-review at HEAD `7964780f` (r1 was COMMENT at `bd4de45b`, draft). PR is now `draft=false` / ready.

## testing approach
- Isolated the delta `bd4de45b..7964780f`: it is a **pure rebase onto newer main** (cosmos gas-estimation, package bumps, `yarn.lock`). The PR's own diff vs base is still exactly one file — `packages/core/chain/swap/general/kyber/api/quote.test.ts` (+119/-0) — and that file is **byte-identical** to r1 (`git diff` empty).
- Verified the Kyber relay path the test exercises (`getKyberSwapQuote`→`getKyberSwapRoute`→`getKyberSwapTx`) is **unchanged** by the rebase; `tx.ts` still returns `dstAmount: amountOut` verbatim with no rescale — so the r1 mutation-proofness verdict carries over.
- Built + ran in a clean worktree (`yarn install` → `yarn workspace @vultisig/core-config build` → vitest `.config/vitest.core.config.ts --no-file-parallelism`): **3/3 pass**.
- Two-lane: Claude (load-bearing) + Codex `gpt-5.4` (exit 0). GLM retired.
- CI green at HEAD (CodeRabbit, Integration/Unit/Lint+Typecheck, Test Summary all pass).

## prior issues — status
| r1 | 1 | suggestion | `impliedRate` assertion is documentary-only (derived from the hardcoded `rawAmountOut` constant, not `quote.dstAmount`) + lossy `Number()` past MAX_SAFE_INTEGER; lower-bound-only wouldn't catch a 10^12 rescale that the `toBe` line already guards | ℹ️ not addressed | Left as-is (rebase-only push). Still non-blocking — the decimals guard is fully carried by `expect(quote.dstAmount).toBe(rawAmountOut)` (lines 137/188). Re-noted in suggestion below. |
| r1 | 2 | risk (P1, bead `vultisig-ops-8fzb`) | `selectBestEligibleQuote` trusts `general.dstAmount` verbatim with no plausibility ceiling — a mispriced +10-12% outlier becomes the SELECTED signable route | ⏭️ deferred | Not folded in (correctly — out of scope for a test-only PR). Bead remains open; not a blocker here. |

## suggestion
- **`should-fix`-adjacent hardening (non-blocking):** the test pins the *decimals/rescale* mutation perfectly (any `amountOut * 10**decimals` regression fails `toBe(rawAmountOut)`), which is the PR's stated purpose. Codex flagged an orthogonal gap: because the fixture sets `routeSummary.amountOut` **and** `build.data.amountOut` to the *same* `rawAmountOut`, the headline assertion would still pass under a *different* class of regression — one that sourced `dstAmount` from `routeSummary.amountOut` instead of `buildResponse.data.amountOut`. To close that, make the two fixture values differ and assert `dstAmount === build.data.amountOut`. Also fold in r1's point: derive `impliedRate` from `quote.dstAmount` and add an upper bound. All future-hardening; none block this PR.

## risk
Test-only change. No product/bundled-SDK output touched → no changeset required, and none needed. The one real fund-safety concern (the `selectBestEligibleQuote` selection ceiling) lives downstream in product code, is tracked in bead `vultisig-ops-8fzb`, and is explicitly out of scope for this regression test. Approving this test does not close that bead.

## verdict
**Approve.** The test correctly and mutation-proofly pins the ~10-12% Kyber stablecoin premium to the upstream aggregator figure (not a decimals artifact) via captured-live fixtures; it is byte-identical to the r1-reviewed version, the relay path it asserts is unchanged, it passes 3/3, and CI is green at `7964780f`. The suggestions above are optional hardening for a follow-up.

<sub>🤖 dual-lane review: Claude (load-bearing) + Codex gpt-5.4. Round 2.</sub>

PR #1018, opened 2026-07-07
feat: add findSwapQuote excludeProviders opt-out (CowSwap-buildability gap)

gomesalexandre:
tl;dr follow-on to agent-backend-ts#825 -- that PR made a CowSwap-selected swap decline cleanly instead of leaking a raw internal string, but it still just declines. This lets a consumer exclude a provider it can't build/sign yet, so the swap falls through to a provider it CAN actually execute.

## the gap

CowSwap has no practical dust-amount minimum the way 1inch/Kyber/LiFi effectively do (their aggregated routes become uneconomical on a $1-2 order), so it disproportionately wins `findSwapQuote`'s best-quote selection on tiny same-chain EVM ERC-20 swaps (`cowSwapSupportedChains` = Ethereum/Arbitrum/Base/Avalanche). agent-backend-ts's `execute_swap.ts` has no `cowswap_order` (EIP-712 typed-data order) signing flow wired -- QA reproduced this live: "swap $2 USDC to USDT on arbitrum" with a low balance leaked the raw `"unsupported general provider tx encoding."` string (fixed at the classification level in agent-backend-ts#825). But the swap STILL doesn't complete -- Kyber/1inch could have executed it fine, they just never got the chance because CowSwap won the quote first.

## the fix

Adds `excludeProviders?: SwapQuoteProviderExcludeName[]` to `FindSwapQuoteInput` -- additive, opt-in. Filtered at the general/native fetcher level, before the sole-THORChain-route check (so excluding the only general fetcher correctly falls through to a native route when one exists). An excluded provider's fetcher is never even invoked -- no wasted network round-trip just to be out-ranked after the fact. Omitted/empty is a complete no-op: every other consumer's default provider set and selection behavior is byte-identical to before this change.

Also exports `SwapQuoteProviderName` and `SwapQuoteProviderExcludeName` from the top-level `@vultisig/sdk` surface so consumers can type their exclusion list against the real provider unions instead of bare strings. Exclusions now accept both display names (`CowSwap`, `KyberSwap`, `LiFi`) and returned quote provider ids (`cowswap`, `kyber`, `li.fi`); unknown tokens fail closed before any quote fetcher runs instead of silently leaving a provider eligible.

**Worth noting:** CowSwap-order signing IS implemented elsewhere in this SDK (`packages/core/mpc/keysign/swap/build.ts`'s `buildSwapKeysignPayload`, used by `packages/sdk/src/tools/prep/swap.ts`) -- this gap is specific to agent-backend-ts, which reimplements its own tx-building instead of using that SDK path and never wired the `cowswap_order` case on its own side. `excludeProviders` lets that one consumer opt out cleanly without changing default behavior for consumers that DO support CowSwap.

## testing

5 new tests in `findSwapQuote.selection.test.ts`: `excludeProviders: ['CowSwap']` skips CowSwap entirely (`getCowSwapQuote` never even called), returned quote ids like `['cowswap']` normalize to the same exclusion, exclude-to-empty fails closed with `NoSwapRoutesError` before any fetcher runs, unknown tokens throw before any fetcher runs, and omitting `excludeProviders` keeps default (CowSwap-eligible) behavior unchanged. Both confirmed fail-before/pass-after (source-only revert via a saved patch + `git apply -R`, not `git stash`). Full swap/quote suite green after `f50c99e8`: 100 tests across 8 files (`--no-file-parallelism` to sidestep this environment's known unrelated timeout flake under parallel load). `tsc --project .config/tsconfig.json --noEmit` clean.

## risk

Low -- purely additive. No existing consumer passes `excludeProviders`, so this ships as dead code from the SDK's perspective until a companion agent-backend-ts PR (queued next) actually wires the exclusion for its execute_swap call. Fund-safety-adjacent (provider-selection logic, not tx correctness) but the change only narrows the candidate set, never widens it or alters tx-building for any provider that remains eligible.

## companion PR

agent-backend-ts wiring (passes `excludeProviders: ['CowSwap']` from `execute_swap.ts`) is queued as a separate PR once this one is reviewed -- it depends on this SDK version being published and the dependency bumped, so it can't fully build/typecheck until then.

## post-review receipts

Addressed NeOMakinG's should-fix in `f50c99e8`:
- normalized returned quote provider ids (`cowswap`, `kyber`, `li.fi`, `swapkit`, `jupiter`, `1inch`) to display provider names before filtering
- unknown exclude tokens now throw before quote fetchers run instead of silently failing open
- added fail-closed coverage for exclude-to-empty and bad-token cases
- added changeset for the public input/export type surface

Receipts:
- `yarn vitest run --config .config/vitest.core.config.ts --no-file-parallelism packages/core/chain/swap/quote/findSwapQuote.selection.test.ts` - 57/57 passing
- `yarn vitest run --config .config/vitest.core.config.ts --no-file-parallelism packages/core/chain/swap/quote` - 100/100 passing across 8 files
- `yarn tsc --project .config/tsconfig.json --noEmit` - pass
- `yarn node scripts/run-eslint.mjs --config .config/eslint.config.mjs packages/core/chain/swap/quote/findSwapQuote.ts packages/core/chain/swap/quote/findSwapQuote.selection.test.ts packages/sdk/src/index.ts` - pass
- `yarn prettier --config .config/.prettierrc --check packages/core/chain/swap/quote/findSwapQuote.ts packages/core/chain/swap/quote/findSwapQuote.selection.test.ts packages/sdk/src/index.ts .changeset/exclude-provider-aliases.md` - pass
- `git diff --check` - clean
- `codex exec review --uncommitted --dangerously-bypass-approvals-and-sandbox` - clean; it reran typecheck and the focused selection suite

## Summary by CodeRabbit

* **New Features**
  * Added support for excluding specific swap providers when requesting quotes.
  * Exclusions can now be entered using either provider display names or provider IDs.
  * The SDK now exposes the updated swap quote typing needed to use this option.

* **Bug Fixes**
  * Unknown exclusion values now fail safely instead of leaving providers eligible.
  * Quote selection now respects exclusions consistently across provider choices.


NeOMakinG (commented):
Reviewed on a clean worktree at `bce4d711` (base `main`). Non-device static + runtime-test review of a fund-critical swap-selection change. **Draft → this is a COMMENT, not an approval.** Two lanes: Claude (primary, load-bearing source trace + adversarial test injection) and Codex `gpt-5.4` (adversarial second opinion).

The core change is genuinely clean and does exactly what the description claims: the exclusion is applied at fetcher-list construction (`getGeneralFetchers().filter(notExcluded)` / `getNativeFetchers().filter(notExcluded)`), **before** the sole-THORChain-route check and **before** quote gathering/selection — so an excluded provider's `fetch()` is never invoked and its quote can never enter `selectBestEligibleQuote`. Default (omitted/empty) path is byte-identical. No caret-bump (3 files, no `package.json`). One substantive footgun, independently flagged by both lanes.

## testing approach
- Worktree `/tmp/wt-sdk-1018` @ `bce4d711`, `yarn install` + `yarn workspace @vultisig/core-config build` (dist needed by `chainInfo.ts`).
- Ran the changed suite: `vitest run --config .config/vitest.core.config.ts --no-file-parallelism packages/core/chain/swap/quote/findSwapQuote.selection.test.ts` → **47 passed** (incl. the 2 new tests).
- Ran the full swap-quote dir → **90 passed across 8 files** (matches PR claim). `tsc --project .config/tsconfig.json --noEmit` → clean.
- **Adversarial injection (my own temp tests, then reverted):** (a) excluding ALL eligible providers on the same-chain EVM fixture → `findSwapQuote` **rejects cleanly** (`NoSwapRoutesError`), no crash / no malformed partial quote → fail-safe confirmed. (b) passing a mis-cased `['cowswap' as any]` → CowSwap is **NOT** excluded, `getCowSwapQuote` is called and CowSwap **still wins** → confirms the case-sensitivity footgun below empirically.

## verification of the 6 fund-safety questions
1. **Filter-before-select** — CONFIRMED. `findSwapQuote.ts:726-727` filters both fetcher lists; excluded fetcher removed from the `fetchers[]` array entirely; `settled` (764) is built from post-filter `fetchers`; `selectBestEligibleQuote` (780) never sees it. Test asserts `getCowSwapQuote` never called.
2. **Fail-safe on exclude-to-empty** — CONFIRMED. `isEmpty(fetchers)` → `throw new NoSwapRoutesError()` (733-734). My inject-all-excluded test throws cleanly. Filter runs before `thorIsSoleRoute` (754) so that check correctly reflects the post-exclusion set (falls through to a native route when one exists).
3. **No regression to default** — CONFIRMED. `new Set(excludeProviders ?? [])` → empty Set → `notExcluded` always true → identical arrays. Test 2 + 90/90 green.
4. **Excluding never alters amount/route for a remaining provider** — CONFIRMED. Only narrows the candidate set; selection/ranking/tx-build for surviving providers is untouched.
5. **allSettled index alignment** — CONFIRMED. `settled`, the below-min loop (`fetchers[i]`), and `failedProviders` (`fetchers[index]`) all index the same post-filter `fetchers[]` array — no misalignment introduced.
6. **Test coverage** — exclude-applied ✅, default ✅. exclude-to-empty and mis-cased-name are **not** covered by the PR (I verified both manually; see should-fix + suggestion).

## should-fix
1. **`excludeProviders` name space (PascalCase) is asymmetric with the provider id the SDK returns on quotes (lowercase), so a mis-keyed exclusion silently no-ops and the non-buildable provider can still win** — the exact failure this PR exists to prevent. `notExcluded` matches `excludeProviderSet.has(fetcher.providerName)` where `providerName` is `'CowSwap' | 'LiFi' | 'KyberSwap' | 'SwapKit' | 'Jupiter' | '1inch'`, but `quote.quote.general.provider` (and `generalSwapProviders`) are `'cowswap' | 'li.fi' | 'kyber' | 'swapkit' | 'jupiter' | '1inch'` — only `1inch` overlaps. A JS/untyped caller, a serialized/JSON exclude-list, or anyone who naively reuses `quote.general.provider` as the exclude token will believe they excluded CowSwap while `Set.has` fails silently and CowSwap stays eligible to win → recreates the "quote wins, tx not buildable" strand. I reproduced this (mis-cased `'cowswap'` → CowSwap still wins). **Independently flagged by Codex.** Suggest normalizing accepted exclude names at runtime (case-insensitive + `li.fi`/`kyber`→union alias) **or** throwing on an unrecognized name, rather than silently treating an unknown/mis-cased entry as "no exclusion". *Not a blocker*: the intended companion (agent-backend-ts, TS) is compile-time-guarded against the exported `SwapQuoteProviderName` union, so `['CowSwap']` works correctly, and the worst-case degradation is the pre-PR status quo (CowSwap wins → clean decline via agent-backend-ts#825), not new fund loss.

## suggestion
1. Add an **exclude-to-empty** test asserting a clean `NoSwapRoutesError` (or an exclude-all → native-fallback case). The fail-safe is real (I verified it), but the new exclusion path can now reach the `isEmpty(fetchers)` guard in a way no test exercises — a regression here would silently strand rather than decline. Cheap to lock down.
2. Consider a negative-guard test that a mis-cased / non-union exclude token behaves per the resolution of should-fix #1 (either normalized-and-excluded, or rejected) so the naming contract is pinned.

## q
1. Companion agent-backend-ts wiring (`excludeProviders: ['CowSwap']` from `execute_swap.ts`) — the PR body says it's queued next and can't build until this ships. Confirm it's tracked; until then this is dead code from the SDK's perspective (stated, fine). Worth linking the companion PR here once opened so the SDK change doesn't sit half-wired (same pattern as #1012/#731).

## risk
Low. Purely additive, opt-in, narrows-only. Default path byte-identical (90/90 green). No tx-build/amount/ranking change for surviving providers. Fail-safe on empty verified. The one real footgun (name asymmetry) is fail-open but bounded: it degrades to the already-existing pre-PR behavior the companion separately fixes, and the intended TS consumer is type-guarded. `SwapQuoteProviderName` is now a public exported type (semver-minor surface) — fine.

## verdict
**COMMENT (draft — never approve).** Mechanics are correct and fail-safe; the change is well-scoped and well-tested for the happy path. Recommend addressing should-fix #1 (name normalization / loud reject) before the companion wiring lands, since that's precisely where a mis-keyed exclusion would silently reintroduce the strand this PR is meant to close.

---
🤖 dual-lane review: Claude (primary, source-trace + adversarial test injection) + Codex gpt-5.4 (adversarial second opinion). Non-device.

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -698,8 +711,20 @@ export const findSwapQuote = async ({
...
+  // build/sign a given provider's tx shape (e.g. agent-backend-ts doesn't wire
+  // CowSwap's `cowswap_order` EIP-712 signing flow — see execute_swap.ts).
+  // Filtered here, before the sole-THORChain-route check below (so excluding
+  // the only general fetcher correctly falls through to a native route when
+  // one exists) rather than at the call site, so every caller of findSwapQuote
+  // gets identical exclusion semantics regardless of which fetcher list a
+  // provider happens to register in. Default (omitted/empty) is a no-op —
+  // every other consumer's behavior is unchanged.
+  const excludeProviderSet = new Set(excludeProviders ?? [])
+  const notExcluded = (fetcher: SwapQuoteFetcher) => !excludeProviderSet.has(fetcher.providerName)
```
NeOMakinG, line 862: should-fix: name space asymmetry — `excludeProviderSet.has(fetcher.providerName)` matches PascalCase labels (`'CowSwap'/`'LiFi'/`'KyberSwap'/`'SwapKit'), but the id the SDK returns on quotes (`quote.general.provider` / `generalSwapProviders`) is lowercase (`'cowswap'/`'li.fi'/`'kyber'/`'swapkit') — only `'1inch' overlaps. A JS/untyped caller, a serialized exclude-list, or anyone reusing `quote.general.provider` as the exclude token hits a silent `Set.has`=false → the "excluded" provider stays eligible and can still win, recreating the non-buildable-quote strand this PR prevents. Reproduced: mis-cased `'cowswap'` → CowSwap still wins. Normalize accepted names at runtime (case-insensitive + `li.fi`/`kyber` aliases) or throw on an unrecognized name instead of silent no-op. Corroborated by the Codex lane. Not a blocker (the TS companion is union-typed; worst case degrades to the pre-PR clean decline), but fix before the agent-backend-ts wiring lands.

gomesalexandre: tackled in `f50c99e8`: `excludeProviders` now normalizes returned quote provider ids (`cowswap`, `kyber`, `li.fi`, etc.) to the display-name fetcher ids before filtering, and unknown tokens throw before any quote fetcher runs instead of silently failing open. Added coverage for lowercase `cowswap`, exclude-to-empty `NoSwapRoutesError`, and unknown-token rejection, plus a changeset for the public input/export type.

NeOMakinG (approved):
Re-review at HEAD `c35d5c64` (was `f50c99e8`; my r1 was COMMENTED at `bce4d711` on the then-draft). The r1 fail-open footgun is fully resolved, and I fixed one follow-on gap in-code.

## testing approach
Two lanes — Claude (load-bearing source trace + runtime) + Codex `gpt-5.4` (adversarial second opinion). Fresh worktree at HEAD, `@vultisig/core-config` built first (build gate). Ran the selection suite (57 pass, up from 47), the full SDK unit suite (176 files / 2713 pass), and `@vultisig/sdk typecheck` (exit 0). Verified the alias map values against the fetcher `providerName` literals line-by-line. Confirmed CI is green on the head commit after a re-run (prior red was the Integration flake fail-fast-cancelling siblings — reproduced-clean locally).

## prior issues — status
|---|---|---|---|---|---|
| r1 | 1 | should-fix | name-space asymmetry fail-open: `excludeProviders`/`f.providerName` PascalCase vs returned `quote.provider`/`generalSwapProviders` lowercase (only `1inch` overlapped) → mis-cased/JSON token silently no-ops and excluded provider still wins | ✅ fixed | commit `fix(swap): normalize excluded quote providers`: `SwapQuoteProviderExcludeName = SwapQuoteProviderName \| GeneralSwapProvider`, `swapQuoteProviderExcludeAlias` (lowercase→PascalCase, `Record<GeneralSwapProvider,…>` forces completeness), `normalizeExcludeProviderName` throws on unknown (fail-closed) applied before the exclude Set. Alias values verified == fetcher `providerName` literals |
| r1 | 2 | suggestion | add exclude-to-empty test (fail-safe path untested) | ✅ fixed | `rejects.toThrow('No swap routes found.')` test added |
| r1 | 3 | suggestion | negative-guard test for mis-cased/non-union token | ✅ fixed | lowercase-alias test (`['cowswap']` excludes CowSwap) + unknown-token test (`rejects.toThrow('Unknown swap quote provider exclusion: not-a-provider')`) added |

## blocking
None.

## preferably-blocking
None.

## should-fix
1. **Public `@vultisig/sdk` wrapper silently dropped `excludeProviders` — FIXED IN-CODE (`c35d5c64`).** Core `findSwapQuote` now honors `excludeProviders`, but the public wrapper `packages/sdk/src/tools/swap/findSwapQuote.ts` omitted the field from `FindSwapQuoteParams` and didn't forward it to core. The only known consumer (agent-backend-ts `execute_swap.ts`) imports `findSwapQuote` from `@vultisig/sdk`, not core, and builds `quoteParams` as a separate `const` object — so TS excess-property checks would NOT fire on an extra `excludeProviders` key, and the companion `excludeProviders: ['CowSwap']` wiring would compile clean while the wrapper silently drops it, recreating the exact CowSwap-wins strand this PR exists to prevent. Exporting `SwapQuoteProviderExcludeName` from the SDK index signals wrapper-consumption intent with nowhere to pass it. Plumbed it through (same silent-drop class already fixed for `slippageTolerance`/`recipient`); default omitted/empty stays a no-op. (Codex-corroborated.)

## q
1. The branch is 5 commits behind `main`, and `#1105`/`#1112` on `main` touch the same `findSwapQuote.ts` + `findSwapQuote.selection.test.ts`. Test-merge is clean (no conflict), but a rebase before merge would de-risk the same-file drift.
2. Same pattern as prior swap-path PRs — worth linking the companion agent-backend-ts PR (the `excludeProviders: ['CowSwap']` wiring) once opened, so this SDK change isn't half-wired.

## risk
Low. `excludeProviders` is additive/opt-in — omitted/empty is byte-identical to prior behavior (verified: default no-op, 2713 unit + 57 selection tests green). Exclusion filters at fetcher-list construction before the sole-THORChain-route check, so an excluded fetcher is never invoked and native fallback stays intact; exclude-to-empty fails closed with `NoSwapRoutesError`; unknown tokens fail closed with a throw. The opt-out trades best-rate for buildability by the caller's explicit choice (documented). Fund-safety: cannot silently strand with no route, cannot return the excluded (unbuildable) provider's quote.

## verdict
Approve. r1 finding fully resolved with comprehensive new coverage; the one follow-on wrapper gap is fixed in-code and CI is green on the head commit.

---
🤖 dual-lane review (Claude + Codex `gpt-5.4`). Verified against source at HEAD `c35d5c64` in an isolated worktree.

PR #1020, opened 2026-07-07
feat(swap): consolidate Skip Go routing-eligibility into the SDK

gomesalexandre:
Chain/swap consolidation directive (recurring, user-emphasized): move non-AI chain/swap routing logic out of agent-backend-ts's ad-hoc special-casing and into the SDK as tested core functions.

## what

"Does this from/to chain pair route through Skip Go?" was independently reimplemented by:
- agent-backend-ts's `execute_swap.ts` (`willRouteViaSkip`, includes the cosmos<->EVM cross-chain branch)
- agent-backend-ts's `list-swap-routes.ts` (its OWN inline `skipBoth`/`terraTouching` computation - **missing the cosmos<->EVM cross branch `execute_swap.ts` has**)
- mcp-ts's `lib/skip-chain-mapping.ts` (a third, separately-maintained copy, whose own doc comment records a *past* incident, #384, caused by exactly this kind of drift between execution and route-discovery)

Neither of abts's two copies shares a single tested source, so the exact bug class mcp-ts already got burned by (#384) is currently live in abts today: a cosmos<->EVM cross pair (e.g. ATOM -> USDC.eth) routes via Skip in `execute_swap` but `list_swap_routes` would report it as routing via the general aggregator instead.

## how

Adds `isSkipRoutableChain`/`isTerraChain`/`willRouteViaSkip` as tested core functions in `packages/core/chain/swap/skip/skipRouting.ts`, exported from the main `@vultisig/sdk` entry. Pure chain-topology predicates - no network calls, no AI-specific logic - mirrors the existing `swapEnabledChains`/`NativeSwapChain` pattern already in the SDK.

Purely additive: no existing export changed, no behavior change to any current consumer.

## why the SDK (not just a shared abts file)

mcp-ts already depends on `@vultisig/sdk` (`^2.13.1`) and already calls the SDK's `findSwapQuote` from its own `execute_swap.ts`/`list-swap-routes.ts`. Once abts adopts this export (companion PR incoming) and mcp-ts eventually bumps its own SDK pin, both backends converge on one tested source instead of drifting independently - the structural fix, not another one-off port.

## testing

- 12 new tests: full positive coverage for `isTerraChain`/`isSkipRoutableChain`, `willRouteViaSkip`'s 3 true-branches (skipBoth/terraTouching/cosmosEvmCross - looped over all 7 allowlisted EVM chains per codex review follow-up), 4 false-branch cases, and a fail-closed-on-unknown-chain-string case.
- Full core suite green (1388 tests), lint clean, typecheck clean, build clean.
- Codex adversarial review: faithful port, no logic drift, one test-coverage gap found and closed (full EVM allowlist coverage, not just 3 of 7 chains).

🤖 Agent-generated

## Summary by CodeRabbit

* **New Features**
  * Added public swap-routing helpers to identify eligible chains and decide when routes can use Skip.
  * Expanded SDK exports so these routing checks are available through the main package.
* **Bug Fixes**
  * Improved routing decisions for Cosmos/EVM combinations and Terra-related paths.
  * Added coverage for unknown chain inputs to fail safely and avoid incorrect routing.

### DIFF
--- a/.changeset/skip-routing-consolidation.md
+++ b/.changeset/skip-routing-consolidation.md
@@ -1,0 +1,13 @@
+---
+"@vultisig/sdk": patch
+---
+
+feat(swap): consolidate Skip Go routing-eligibility predicates into the SDK
+
+Adds `isSkipRoutableChain`, `isTerraChain`, and `willRouteViaSkip` as tested
+core functions under `@vultisig/core-chain/swap/skip/skipRouting`, re-exported
+from the main `@vultisig/sdk` entry. Pure chain-topology predicates (no network
+calls, no AI-specific logic) that give consumers one shared source of truth for
+"does this from/to chain pair route through Skip Go?" instead of the
+independently-maintained copies that can drift (the mcp-ts #384 bug class).
+Purely additive — no existing export or behavior changed.
--- a/packages/core/chain/package.json
+++ b/packages/core/chain/package.json
@@ -1717,6 +1717,11 @@
       "import": "./dist/swap/quote/SwapQuote.js",
       "default": "./dist/swap/quote/SwapQuote.js"
     },
+    "./swap/skip/skipRouting": {
+      "types": "./dist/swap/skip/skipRouting.d.ts",
+      "import": "./dist/swap/skip/skipRouting.js",
+      "default": "./dist/swap/skip/skipRouting.js"
+    },
     "./swap/swapEnabledChains": {
       "types": "./dist/swap/swapEnabledChains.d.ts",
       "import": "./dist/swap/swapEnabledChains.js",
--- a/packages/core/chain/swap/skip/skipRouting.test.ts
+++ b/packages/core/chain/swap/skip/skipRouting.test.ts
@@ -1,0 +1,83 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import { describe, expect, it } from 'vitest'
+
+import { isSkipRoutableChain, isTerraChain, willRouteViaSkip } from './skipRouting'
+
+describe('isTerraChain', () => {
+  it('is true for Terra v2 and TerraClassic only', () => {
+    expect(isTerraChain(Chain.Terra)).toBe(true)
+    expect(isTerraChain(Chain.TerraClassic)).toBe(true)
+  })
+
+  it('is false for every other chain, including other Skip-routable cosmos chains', () => {
+    expect(isTerraChain(Chain.Cosmos)).toBe(false)
+    expect(isTerraChain(Chain.Osmosis)).toBe(false)
+    expect(isTerraChain(Chain.THORChain)).toBe(false)
+    expect(isTerraChain(Chain.Ethereum)).toBe(false)
+    expect(isTerraChain('NotAChain')).toBe(false)
+  })
+})
+
+describe('isSkipRoutableChain', () => {
+  it('is true for every Skip-indexed cosmos chain', () => {
+    for (const chain of [
+      Chain.Cosmos,
+      Chain.Osmosis,
+      Chain.Kujira,
+      Chain.Terra,
+      Chain.TerraClassic,
+      Chain.Akash,
+      Chain.Dydx,
+      Chain.Noble,
+    ]) {
+      expect(isSkipRoutableChain(chain)).toBe(true)
+    }
+  })
+
+  it('is false for THORChain/MayaChain (native providers, not Skip-indexed) and non-cosmos chains', () => {
+    expect(isSkipRoutableChain(Chain.THORChain)).toBe(false)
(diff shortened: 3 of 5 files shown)


NeOMakinG (approved):
## testing approach

Two-lane review (Claude load-bearing + Codex `gpt-5.4` adversarial). Fresh worktree at `ab0799e3`, non-device (pure predicate + export change). Verified port fidelity line-by-line against the three live originals — abts `execute_swap.ts::willRouteViaSkip`, abts `skip-chain-mapping.ts`, mcp-ts `skip-chain-mapping.ts`. Ran the new suite (`skipRouting.test.ts` 11/11 green via `.config/vitest.core.config.ts`). Traced the fail-safe direction and the SDK entry re-export wiring for every platform surface. Confirmed the release-guard fix at runtime (CI static gate flipped red→green on the changeset commit).

## consolidation vs behavior-change

**Pure faithful consolidation — zero behavior change, zero deletion.** The three predicate sets match the abts originals exactly:
- `SKIP_ROUTABLE_COSMOS_CHAINS` = {Cosmos, Osmosis, Kujira, Terra, TerraClassic, Akash, Dydx, Noble} — identical to abts.
- `TERRA_CHAINS` = {Terra, TerraClassic} — identical.
- `SKIP_SUPPORTED_EVM` = {Ethereum, Arbitrum, Optimism, Base, Polygon, Avalanche, BSC} — the SDK checks the `Chain` enum directly; abts maps chain→numeric-id {1, 42161, 10, 8453, 137, 43114, 56} then checks the id set. The two are one-to-one equivalent (both gated by `isChainOfKind(_, 'evm')` first).
- `willRouteViaSkip = skipBoth || terraTouching || cosmosEvmCross` — structurally identical.

No swap that should route via Skip becomes ineligible and none newly becomes eligible; the predicate is byte-for-byte the same decision the two backends already make today.

## eligibility predicate + fail-safe direction

`willRouteViaSkip` is **fail-closed to `false`** on any unknown/uncanonicalized chain string (`Set<Chain>.has(x as Chain)` misses; `isChainOfKind` returns false). `false` is the safe direction: in the consumer, `willRouteViaSkip===true` both routes via Skip *and* bypasses the LLM-destination format gate (safe only because Skip derives + validates its own destination), while `false` keeps the pair on the native/general-aggregator lane and *enforces* destination validation. An unknown pair therefore never gets pushed into a Skip route it can't build and never bypasses destination checks. `chain as Chain` casts on raw strings are sound here — string enum, membership test only.

## dual/divergent logic

Purely additive groundwork (same shape as sdk#1013). Within the SDK there is no divergence — it adds one new tested source. The abts×2 + mcp-ts copies still exist and only converge once the companion adopt-PRs land; the PR body states this explicitly. No new drift is introduced by this PR.

## fix applied (was the sole CI blocker)

CI was red for one reason: the SDK release guard — `skipRouting.ts` is bundled into `@vultisig/sdk` but no changeset bumped it, so the new exports would ship **inert** (no new tarball for the abts/mcp-ts adopters — the #1012 ships-inert class). Added the missing `@vultisig/sdk: patch` changeset (`ef50af96`, committed under the pinned identity). Static gate re-ran **green**; guard passes locally too.

## should-fix

None blocking. Two Codex observations, both resolved as non-issues for the documented consumers:
- **RN surface**: the new exports reach the Node/browser/electron entries (`platforms/node/index.ts` does `export * from '../../index'`) but not the React-Native entry, which uses a curated allowlist. This is the **pre-existing pattern** — the sibling swap exports it "mirrors" (`getSwapExplorerUrl`, `swapExplorerProviders`, `swapEnabledChains`, `NativeSwapChain`) are also absent from the RN entry. The real adopters (agent-backend-ts, mcp-ts) are both Node. No regression; if RN ever needs these, that's a separate curation decision spanning all the swap-metadata exports.
- **Deep subpath**: `@vultisig/sdk/swap/skip/skipRouting` is not a public subpath — but the PR advertises the **main-entry** import (`import { willRouteViaSkip } from '@vultisig/sdk'`), which resolves fine via the Node entry's `export *`. The core-chain subpath added here is only the internal resolution target. Not a bug.

## risk

Low. Additive predicates, no consumer wired in-repo yet, fail-closed, faithful port with full positive + negative + fail-closed test coverage (both Skip-true branches × all 8 cosmos and all 7 EVM chains, plus THOR/Maya-excluded and unknown-string cases). Caret-bump check: **N/A** — package.json change only adds an `exports` map entry; no dependency version or caret was bumped.

## verdict

**Approve.** Faithful consolidation, safe fail-closed predicate, no behavior change, no divergence introduced, comprehensive tests. The one real defect (missing changeset → inert ship + red CI) is fixed in `ef50af96` and the static gate is confirmed green.

---
🤖 Reviewed with two lanes (Claude Opus 4.8 + Codex gpt-5.4). Changeset fix committed under the maintainer identity.

PR #971, opened 2026-07-04
Validate MayaChain native swap inbound vault

rcoderdev:
Closes #933
Adds MayaChain native swap inbound-address revalidation before broadcast, using the active Maya inbound vault for the source chain and failing closed when a quote points at a stale vault.

## Summary by CodeRabbit

* **New Features**
  * Native swap broadcast checks now support MayaChain in addition to THORChain.
  * Inbound vault addresses are validated for the active swap chain before broadcast.

* **Bug Fixes**
  * Improved handling of stale or changed MayaChain inbound vault addresses.
  * Updated validation messages to reflect the relevant chain more clearly.

* **Tests**
  * Expanded coverage for MayaChain swap readiness, including valid, stale, and expired payload scenarios.

PR #1022, opened 2026-07-07
feat(swap): consolidate cosmos memo-length cap knowledge into the SDK

gomesalexandre:
Chain/swap consolidation directive, next candidate after the Skip-routing one (#1020, merged).

## what

"What's the x/auth.MaxMemoCharacters byte cap for a cosmos chain's memo?" was independently maintained in agent-backend-ts's `skip-swap.ts` (a full per-chain-id table, used only for Skip routes) and mcp-ts's own copy of the same table - the exact drift-prone pattern the Skip-routing consolidation already found once.

Digging further found this candidate is more than theoretical duplication: agent-backend-ts's `execute_send.ts` has its own THIRD, much narrower check - hardcoded 256 bytes, gated to TerraClassic only. Every other cosmos chain a user can `execute_send` FROM with a memo (Cosmos Hub, Osmosis, Kujira, Terra v2, Noble, Dydx, Akash, ...) has no memo-length check in `execute_send` at all - and Terra v2/Cosmos Hub specifically have a gov-raised 512-byte cap, not 256, so even a TerraClassic-shaped check ported to those chains naively would still be wrong. A memo over the cap signs fine, then gets rejected at broadcast with sdk code 12 ("memo too long") - burning the MPC signing ceremony for nothing.

## how

Adds `getCosmosMemoMaxBytes`/`getCosmosMemoMaxBytesByChainId`/`isCosmosMemoWithinCap` as tested core functions in `packages/core/chain/chains/cosmos/cosmosMemoCap.ts`, exported from `@vultisig/sdk` (same spot as the existing cosmos chain-metadata exports - see that section's own comment: "surfaced so consumers stop re-declaring LCD urls / fee denoms / gas limits (e.g. mcp-ts lib/cosmos-chains.ts)", this is the identical pattern). Uses `TextEncoder` (not Node's `Buffer`) for the UTF-8 byte-length check so it's RN-bridge-safe.

Companion abts PR (incoming) wires `execute_send.ts` to check EVERY cosmos chain (fixing the real gap) and `skip-swap.ts` to use the shared table instead of its own copy.

## testing

- 9 tests: override resolution (Terra v2/Cosmos Hub = 512, everything else = 256 default), chain-id string lookup with fail-closed-to-default for unrecognized ids, exact-boundary pairs at both 256 and 512 bytes, and a UTF-8 multi-byte-character case (a naive `.length` check would under-count and let an over-cap memo through - JS string length counts UTF-16 code units, not bytes).
- Codex adversarial review: faithful port (verified byte-for-byte against both abts's and mcp-ts's copies), confirmed `TextEncoder`/`Buffer.byteLength` behavioral parity, one coverage gap found (512-byte exact boundary) and closed.
- Full core suite green (1390 tests), lint clean, typecheck clean, build clean.

🤖 Agent-generated

## Summary by CodeRabbit

* **New Features**
  * Added Cosmos memo size limit helpers to the SDK, including a default cap, chain-specific cap lookup, and a “memo within cap” checker.
  * Expanded public exports and added a dedicated Cosmos memo-cap module.
* **Bug Fixes**
  * Updated the Skip Go swap memo validation to use the shared chain-specific cap logic (fixing incorrect caps for certain chains).
* **Tests**
  * Added coverage for override vs default behavior, unknown chain-id fallback, and UTF-8 byte-length validation at cap boundaries.


packages/core/chain/chains/cosmos/cosmosMemoCap.ts
```diff
@@ -0,0 +1,56 @@
...
+  // Terra raised MaxMemoCharacters above the sdk default.
+  [CosmosChain.Terra]: 512,
+  // Cosmos Hub (cosmoshub-4): live-verified 512 (2026-06-22, /cosmos/auth params) -
+  // gov-raised above the sdk default.
+  [CosmosChain.Cosmos]: 512,
+}
+
+/**
+ * The `x/auth.MaxMemoCharacters` byte cap for a cosmos chain's outer
+ * `TxBody.memo` field. A tx whose memo exceeds this is rejected at broadcast
```
NeOMakinG, line 31: `q:` docstring correctly scopes this to the **outer** `TxBody.memo`. Worth keeping in mind for the companion adoption PR: THOR/Maya fund-routing memos ride the *inner* `MsgDeposit.memo` (the RN builder emits `buildTxBody(msgAny, '')` — empty outer memo), which `x/auth.MaxMemoCharacters` does **not** govern. So `isCosmosMemoWithinCap(THORChain/MayaChain, memo)` must not be wired into the deposit path — it'd validate against the wrong constraint (fail-safe, but false coverage). A plain THOR/Maya `MsgSend` (outer memo) is correctly covered.

NeOMakinG (approved):
Reviewed as a fund-critical SDK change through the memo-truncation lens. This is a clean, purely-additive consolidation (+156/-0) that adds cosmos `x/auth.MaxMemoCharacters` cap helpers as SDK-owned functions. **fix-and-approve**: the one blocking gap (missing changeset → exports ship inert + static gate red, the #1012/#1020 class) I fixed in-branch as commit `832cf99`; static gate + unit + integration are now green on that commit.

## testing approach
Non-device static + runtime-params review on a fresh worktree at `bf23b25c`.
- **Every cap value re-verified against LIVE `/cosmos/auth/v1beta1/params` (2026-07-08), all 10 mapped chains** — every value matches the PR:

| chain | chain-id | live `max_memo_characters` | PR value | ✓ |
|---|---|---|---|---|
| Cosmos Hub | cosmoshub-4 | 512 | 512 (override) | ✓ |
| Terra v2 | phoenix-1 | 512 | 512 (override) | ✓ |
| Osmosis | osmosis-1 | 256 | 256 (default) | ✓ |
| dYdX | dydx-mainnet-1 | 256 | 256 | ✓ |
| Akash | akashnet-2 | 256 | 256 | ✓ |
| MayaChain | mayachain-mainnet-v1 | 256 | 256 | ✓ |
| TerraClassic | columbus-5 | 256 | 256 | ✓ |
| Kujira | kaiyo-1 | 256 | 256 | ✓ |
| Noble | noble-1 | 256 | 256 | ✓ |
| THORChain | thorchain-1 | 256 | 256 | ✓ |

- **No truncation anywhere.** `isCosmosMemoWithinCap` returns a `boolean` — validation-only, no path mutates/truncates a memo. The worst outcome of a wrong cap here is an honest reject/fail, never a silently-truncated fund-routing memo. This is the safe side of the truncate-vs-reject line.
- **Fail-safe on unknown chain-id = fail-closed to the strictest cap.** `getCosmosMemoMaxBytesByChainId('not-a-real-chain-id')` → `COSMOS_MEMO_DEFAULT_MAX_BYTES` (256, the lowest cap in the table). An unmapped chain gets the tightest check, never "no cap". Correct direction.
- **Byte-counting matches on-chain semantics.** cosmos-sdk `validateMemo` uses `len(memo)` on a Go string = UTF-8 **bytes**; `new TextEncoder().encode(memo).length` models that exactly. The multi-byte (`🚀`) boundary test locks it — a naive `.length` (UTF-16 code units) would under-count and let an over-cap memo pass.
- Ran the suite: `cosmosMemoCap.test.ts` 8/8 green (PR body says "9" — cosmetic miscount, 8 `it` blocks). CI on my `832cf99`: Lint&TypeCheck ✅ / Unit ✅ / Integration ✅ / CodeRabbit ✅.
- Two-lane: Claude (load-bearing) + Codex `gpt-5.4` adversarial (exit 0).

## blocking
_None outstanding._ The sole blocker was resolved in-branch:
- **Missing changeset → inert ship + red static gate (fixed, `832cf99`).** `cosmosMemoCap.ts` lives under `packages/core/chain/` = a `bundledSourcePrefixes` path → it's bundled into the `@vultisig/sdk` tarball, so without a `@vultisig/sdk` changeset the new exports ship inert (no consumer tarball) and `check-sdk-bundled-changeset.mjs` fails "Lint and Type Check". I added `.changeset/cosmos-memo-cap-consolidation.md` (`"@vultisig/sdk": patch`); guard passes locally and the static gate is now green on HEAD. Same class as #1012/#1020.

## should-fix
- **In-repo `skipSwap.ts` keeps a SECOND, now-divergent-and-wrong cap table.** `packages/sdk/src/tools/swap/skip/skipSwap.ts:230` (`COSMOS_MEMO_MAX_BYTES_BY_CHAIN_ID`) hardcodes `phoenix-1: 256` and `cosmoshub-4: 256` — both actually **512** (this PR's own verified overrides, and my live re-check today). Its comment even claims "Values sourced from each chain's live `/cosmos/auth/v1beta1/params`", which is wrong for exactly those two. Effect is **fail-safe** (a valid 257–512 byte Skip source-leg memo on Terra v2 / Cosmos Hub gets false-rejected — a blocked swap, no fund loss), and it's **pre-existing**, so it doesn't block this PR. But it directly undercuts the "single source of truth" premise, and note the companion **abts** PR won't reach this table — it lives in *this* repo. Please point `skipSwap.ts` at `getCosmosMemoMaxBytesByChainId` here or in an immediate follow-up, so Terra v2 / Cosmos Hub stop under-capping at 256. (Found by Codex lane.)

## q
- **THOR/Maya fund-routing memos ride the *inner* `MsgDeposit.memo`, not the outer `TxBody.memo` this helper models.** `buildThorchainDepositTx` emits `buildTxBody(msgAny, '')` — outer memo empty — and puts the routing memo inside `MsgDeposit`. `x/auth.MaxMemoCharacters` governs only the outer `TxBody.memo`, so applying `isCosmosMemoWithinCap(THORChain/MayaChain, memo)` to a deposit's inner memo would validate against the wrong constraint (fail-safe direction — false-reject — but false coverage). The docstring here correctly scopes to "outer `TxBody.memo`", so the SDK is self-consistent; this is a heads-up for the companion adoption PR to *not* wire this into the THOR/Maya `MsgDeposit` path. A plain THOR/Maya `MsgSend` (outer memo) is correctly covered. (Corroborated by Codex lane.)

## risk
Low. Purely additive, no deletions, no existing export or behavior changed; every cap value verified against live chain params; validation-only (no truncation); fail-closed on unknown. Divergent-table and THOR/Maya-semantics items are non-fund-loss and belong to the adoption phase. Node/browser/electron reach the new exports via `export *`; the RN entry is a curated allowlist so the helper is Node-only for now — matches the sibling cosmos exports (`cosmosRpcUrl`, `cosmosFeeCoinDenom`), a pre-existing pattern, not a regression.

## verdict
**APPROVE** (fix-and-approve). Blocking changeset gap fixed in-branch (`832cf99`, CI green); all cap values correct vs live params; no truncation path; fail-closed on unknown. `skipSwap.ts` divergent table + THOR/Maya inner-memo semantics are non-blocking follow-ups for the adoption PR.

---
🤖 Two-lane review: Claude (correctness/fund-safety, load-bearing) + Codex `gpt-5.4` (adversarial second opinion). Generated with [Claude Code](https://claude.com/claude-code).

NeOMakinG (approved):
## testing approach
r2 re-review of the delta since our r1 approval (`62d67546..c4eea266`), two lanes: Claude (load-bearing) + Codex gpt-5.4 (adversarial second opinion). Worktree `/tmp/wt-sdk-1022b` @ c4eea266. Focus commit: `cd189f31` (skipSwap → shared helper). Local: changeset guard + `cosmosMemoCap` vitest (8/8) + CI check-runs.

## prior issues — status
| r1 | 1 | should-fix | `skipSwap.ts` kept a divergent `COSMOS_MEMO_MAX_BYTES_BY_CHAIN_ID` table under-capping phoenix-1 & cosmoshub-4 at 256 (real 512) | ✅ fixed | `cd189f31` deletes the local table and calls shared `getCosmosMemoMaxBytesByChainId` — phoenix-1/cosmoshub-4 now correctly 512, single source of truth restored |
| r1 | 2 | q | THOR/Maya fund-routing memos ride inner `MsgDeposit.memo`, not outer `TxBody.memo`; don't wire the cap into that path | ℹ️ documented | groundwork only; skipSwap adoption targets the Skip source-leg outer memo, not MsgDeposit. Docstring on the core helper still correctly scopes to "outer TxBody.memo". No regression. |

## blocking
None.

## should-fix
None.

## suggestion
- `validateMsgsResponse` now guards `cap !== undefined`, but `getCosmosMemoMaxBytesByChainId` always returns a number, so that branch of the guard is dead code. Harmless — not worth a spin.

## risk
Low. Core `cosmosMemoCap.ts` is byte-identical to the r1-approved state (override table Terra=512/Cosmos=512, default 256, fail-closed on unknown, validation-only — no truncation). The skipSwap change is strictly fail-safe: it *raises* the cap for phoenix-1/cosmoshub-4 (fewer false rejects) and adds correct 256 coverage for chains the old 6-entry table left uncapped. Codex confirms `firstUnsupportedCustodyChain` rejects unsupported custody chains before memo validation, so the 256 fail-closed fallback never mis-fires on a legitimate route. The other delta commits (#1023 version packages, #971 Maya inbound-vault guard) arrive via `origin/main` merges, touch an unrelated subsystem, and leave the memo-cap core untouched.

## verdict
APPROVE. r1 should-fix resolved (shared helper, no divergent table, phoenix-1/cosmoshub-4 → 512). Core invariants preserved across the main-merges: cap values correct, validation-only, fail-closed, changeset present, all 4 CI checks green. Codex lane: no findings. cosmosMemoCap 8/8 local, changeset guard green.

<sub>reviewed by NeOMakinG · 2-lane (Claude + Codex gpt-5.4) · r2 @ c4eea266</sub>

PR #1024, opened 2026-07-08
fix(sdk-cli): tx-status no longer reports bad/nonexistent hashes as pending forever

neavra:
## Problem (audit 070726-sdkcli-product-audit P2-2, HIGH)

`vultisig tx-status` defaults to polling every 5s until a tx confirms. But the EVM status resolver mapped **malformed hashes, RPC errors, and never-seen hashes all to `status:"pending"`**. Consequences:

- A typo'd or dropped `--tx-hash` polls **forever** — an unattended script hangs.
- A false `"pending"` is indistinguishable from a real in-flight tx.
- No hash-format validation (`nothash` was accepted), no terminal not-found state.

```
vultisig tx-status --chain ethereum --tx-hash nothash --no-wait
# → status:"pending", exit 0   (should be a validation error)
vultisig tx-status --chain ethereum --tx-hash 0x1111…1111 --no-wait
# → status:"pending"           (nonexistent hash, reported same as in-flight)
```

## Fix

1. **Distinguish not-found from pending (EVM resolver).** When there's no receipt, the resolver now probes `getTransaction`: if the node knows the hash it's a true `pending` (`isKnown:true`); if the node has never seen it, a new terminal **`not_found`** status. Transient RPC errors stay non-terminal (`pending`, `isKnown:false`) so a flaky node isn't mistaken for a missing tx.
2. **Validate the hash before any RPC.** New `isValidTxHash(chain, hash)` (exported from `@vultisig/sdk`) checks the hash shape per chain-kind. The CLI rejects a malformed `--tx-hash` with `INVALID_INPUT` (**exit 4**) and makes **no RPC call**.
3. **Bound the wait.** Polling mode now honors a total wait budget `--timeout <seconds>` (default 120) and exits non-zero on give-up: **`TX_NOT_FOUND` (exit 5)** when the node has no record of the hash, **`TX_STATUS_TIMEOUT` (exit 3, retryable)** when it's still pending. `--no-wait` reports the true status (incl. `not_found`) and exits 0 as before.

True mempool-pending stays distinct from never-seen throughout.

## Tests / gates

- `packages/core/chain/tx/status/resolvers/evm.test.ts` — success+receipt, revert→error, true-pending (`isKnown:true`), never-seen→`not_found`, transient errors stay pending.
- `packages/core/chain/tx/isValidTxHash.test.ts` — per-chain-kind accept/reject (rejects `nothash`, wrong length, missing/extra `0x`).
- `clients/cli/src/commands/__tests__/txStatus.test.ts` — malformed→`INVALID_INPUT` with **no RPC call**; confirmed→success+receipt; never-seen→`TxNotFoundError` (single poll, no infinite loop); stuck-pending→`TxStatusTimeoutError`; `--no-wait` not_found reports without throwing; bounded polling terminates on success.
- `yarn typecheck` ✅ · `yarn check` (lint/knip/format) ✅ · `yarn test:cli` ✅ (590) · core tx tests ✅ (99)

Changeset included: `@vultisig/core-chain`, `@vultisig/sdk`, `@vultisig/cli` (patch).

---

**Verified end-to-end against Ethereum mainnet RPC** (read-only, no funds): malformed hash (`nothash`, `0x123`) → exit 4 INVALID_INPUT with no RPC poll; well-formed nonexistent hash → `not_found` (terminal, no longer `pending`); confirmed tx → `success` + fee receipt; default polling on a nonexistent hash with `--timeout 5` gives up bounded at exit 5 (`TX_NOT_FOUND`) instead of hanging.


gomesalexandre (approved):
looks good.

receipts
- `yarn build:all && yarn cli:build`
- `yarn vitest run --config .config/vitest.core.config.ts packages/core/chain/tx/isValidTxHash.test.ts packages/core/chain/tx/status/resolvers/evm.test.ts`
- `yarn vitest run --config .config/vitest.cli.config.ts clients/cli/src/commands/__tests__/txStatus.test.ts`
- `node clients/cli/dist/index.js tx-status --chain Ethereum --tx-hash nothash --no-wait --ci` -> `INVALID_INPUT` exit 4 with the new chain-aware hash-shape guard
- `node clients/cli/dist/index.js tx-status --chain Ethereum --tx-hash 0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa --no-wait --ci` -> `status: "not_found"` instead of false `pending`

review notes
- traced the new `not_found` status through the EVM resolver, SDK exports, and CLI handling
- the bounded poll loop now caps sleep to the remaining timeout budget, so small `--timeout` values no longer oversleep by a full 5s interval

PR #1026, opened 2026-07-08
feat(swap): consolidate cosmos fee-denom allowlist into the SDK

gomesalexandre:
## what

`execute_send.ts` only honors a `fee_denom` override on TerraClassic (`uluna`/`uusd`) - every other cosmos chain rejects any non-native `fee_denom`, including Osmosis, even though Osmosis's `x/txfees` module genuinely whitelists several IBC tokens (ATOM, Noble USDC, Axelar USDC) for fee payment. `cosmos-staking.ts` already has this richer per-chain allowlist for delegate/claim/redelegate/undelegate - so paying an Osmosis staking fee in USDC works, but the exact same `fee_denom` on a plain send gets wrongly rejected. `astroport-classic-swap.ts` carries a third independent copy of the TerraClassic-only allowlist.

## how

Adds `getCosmosAllowedFeeDenoms` / `isCosmosFeeDenomAllowed` as tested core functions under `@vultisig/core-chain/chains/cosmos/cosmosFeeDenomAllowlist`, re-exported from `@vultisig/sdk`. Deliberately curated (native denom + live-verified alternates), not a claim of exhaustiveness against Osmosis's real 173+-entry on-chain whitelist - matches the existing design already accepted in `cosmos-staking.ts`.

Follow-up abts PR will wire `execute_send.ts` / `astroport-classic-swap.ts` / `cosmos-staking.ts` to this shared source instead of their 3 independently-maintained copies.

## testing

- `yarn test:core` - 163 test files / 1400 tests, all green
- `yarn typecheck` - clean
- `yarn lint` - clean
- Codex adversarial review: 1 real finding (Osmosis doc comment could be read as claiming exhaustive coverage) - tackled in a follow-up commit clarifying it's a deliberately-curated subset

## risk

Low - purely additive SDK export, no existing behavior changed. abts wiring (the actual bug fix) comes in a follow-up PR, same pattern as #1020/#1022.

### DIFF
--- a/.changeset/cosmos-fee-denom-allowlist.md
+++ b/.changeset/cosmos-fee-denom-allowlist.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Add cosmos per-chain fee-denom allowlist helpers (`getCosmosAllowedFeeDenoms`, `isCosmosFeeDenomAllowed`) as the single source of truth for which denoms a cosmos chain's ante handler accepts as a gas fee, consolidating copies previously maintained independently in agent-backend-ts (execute_send.ts, astroport-classic-swap.ts, cosmos-staking.ts).
--- a/packages/core/chain/chains/cosmos/cosmosFeeDenomAllowlist.test.ts
+++ b/packages/core/chain/chains/cosmos/cosmosFeeDenomAllowlist.test.ts
@@ -1,0 +1,59 @@
+import { CosmosChain } from '@vultisig/core-chain/Chain'
+import { describe, expect, it } from 'vitest'
+
+import { getCosmosAllowedFeeDenoms, isCosmosFeeDenomAllowed } from './cosmosFeeDenomAllowlist'
+
+describe('getCosmosAllowedFeeDenoms', () => {
+  it('is single-entry (native only) for chains with no live-verified alternate', () => {
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.Cosmos)).toEqual(['uatom'])
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.Kujira)).toEqual(['ukuji'])
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.Terra)).toEqual(['uluna'])
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.Akash)).toEqual(['uakt'])
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.Dydx)).toEqual(['adydx'])
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.Noble)).toEqual(['uusdc'])
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.THORChain)).toEqual(['rune'])
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.MayaChain)).toEqual(['cacao'])
+  })
+
+  it('includes the uusd (USTC) alternate for TerraClassic alongside the native uluna', () => {
+    expect(getCosmosAllowedFeeDenoms(CosmosChain.TerraClassic)).toEqual(['uluna', 'uusd'])
+  })
+
+  it('includes the whitelisted IBC fee tokens for Osmosis alongside native uosmo', () => {
+    const allowed = getCosmosAllowedFeeDenoms(CosmosChain.Osmosis)
+    expect(allowed[0]).toBe('uosmo')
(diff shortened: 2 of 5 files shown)


NeOMakinG (approved):
**APPROVE** — clean additive consolidation, verified fund-safe. Third in the swap-source consolidation family after #1020 (Skip routing) and #1022 (memo cap); unlike those two, the changeset is already present here.

## testing approach
Two-lane review (Claude fund-safety/correctness + Codex gpt-5.4 adversarial). Fresh worktree at HEAD `7e66d80b`, `yarn install --immutable`, ran the new suite via vitest (`7/7` green). Verified each chain's fee denom against the SDK's own typed `cosmosFeeCoinDenom` record and cross-checked the consolidation against all three live abts copies it replaces (`cosmos-staking.ts`, `execute_send.ts`, `astroport-classic-swap.ts`).

## verification

**Consolidation vs behavior-change — faithful port, strictly-widening, no denom dropped.** `getCosmosAllowedFeeDenoms` = `[cosmosFeeCoinDenom[chain], ...COSMOS_ALTERNATE_FEE_DENOMS[chain]]`. The alternate table (`TerraClassic: ['uusd']`, `Osmosis: [ATOM/NobleUSDC/axlUSDC ibc]`) is byte-identical to abts `cosmos-staking.ts::ALLOWED_FEE_DENOMS` (same 3 IBC hashes, same order), extended conservatively with `THORChain/MayaChain/Noble` as native-only. The two thin copies (`execute_send.ts`, `astroport-classic-swap.ts`) are TerraClassic-only `['uluna','uusd']` — both are strict subsets of this helper, so the follow-up wiring only *widens* their coverage (that's the actual bug fix: Osmosis USDC fee on a plain send). No denom accepted by any current copy is absent from the SDK helper.

**Per-chain native denoms — all 10 correct** (TS-complete `Record<CosmosChain,string>`): rune / uatom / uosmo / cacao / adydx / ukuji / uluna / uluna(TerraClassic) / uusdc / uakt. Osmosis IBC hashes match the SDK's own pinned token registry (`knownTokens/cosmos.ts:201-203`) and are the canonical ATOM/Noble-USDC/axlUSDC denoms.

**Fail-safe on unknown chain — fail-closed (safe).** A raw string outside the enum yields `cosmosFeeCoinDenom[chain] === undefined` → allowlist `[undefined]` → `isCosmosFeeDenomAllowed` returns `false` for every real denom. Rejecting an unknown `fee_denom` is an honest failure, never a permissive default.

**Consumption — none in-repo yet.** Purely additive groundwork (+125/-0); no signable tx path consumes it in this PR. Callers wired in the follow-up abts PR (same pattern as #1020/#1022).

**Changeset present + CI green.** `.changeset/cosmos-fee-denom-allowlist.md` (`@vultisig/sdk: patch`) — the #1012/#1020/#1022 ships-inert class does NOT apply. All checks green at `7e66d80b` (Lint&TypeCheck, Unit, Integration, Test Summary, CodeRabbit). No caret/dependency bump (exports-map entry only).

## suggestion
`suggestion:` (Codex lane) The Osmosis entry is a static 3-token snapshot of the governance-owned `/osmosis/txfees/v1beta1/fee_tokens` whitelist. If Osmosis gov ever *removes* one of these three, `isCosmosFeeDenomAllowed(Osmosis, …)` would return `true` for a denom the ante handler no longer accepts → the builder constructs a rejected tx (honest on-chain failure, no fund loss). The docstring already frames this as a curated product-supported subset rather than authoritative chain truth, so this is a documented tradeoff — just worth a `TODO: refresh quarterly or switch to a cached txfees query` (the abts `cosmos-staking.ts` copy already carries exactly that TODO).

## q
`q:` When the follow-up wires `execute_send.ts` to this helper, `THORChain`/`MayaChain`/`Noble` become reachable through the shared path as native-only. Confirm those chains' send flows don't currently accept any non-native fee denom that this native-only entry would newly reject (my read of the three copies says no — none list alternates for those chains — but the follow-up PR is the place to lock it with a test).

## risk
Low. Purely additive SDK export, no existing behavior changed, faithful verbatim port of the richest live copy, fail-closed on unknown input, changeset present, CI green. The behavioral change (widening `execute_send`/`astroport` beyond TerraClassic) lands in the follow-up abts PR and should be QA'd there.

## verdict
Approve.

---
🤖 dual-lane review (Claude fund-safety + Codex gpt-5.4). Codex's regression concern (consolidation might drop a fee denom `cosmos-staking.ts` accepts) was investigated and disproven — the helper is a verbatim superset of all three abts copies; nothing dropped.

PR #1027, opened 2026-07-08
fix(cli): dedup-guard direct send/swap against duplicate broadcast (completed-broadcast path, P5-1)

neavra:
## Summary

The direct `send` / `swap` CLI verbs called `vault.send()` / `vault.swap()` straight through and **never consulted the persistent broadcast journal** that the `agent ask` path uses. So a retried `send`/`swap` re-broadcast an identical intent — a **double-spend** (audit **P5-1, HIGH**; Codex-verified).

> **Scope of this fix (narrowed per review):** this guards the direct-command **retry / completed-broadcast** path — an identical `send`/`swap` whose broadcast already landed (hash returned + journaled) is refused on the next attempt. It does **not** cover a **compound swap's ERC-20 approval leg** that broadcasts *inside* `VaultBase.swap()` before a later step throws — that leg is not journaled by this CLI layer, so a retry can still re-broadcast it. Closing that requires guarding at the SDK broadcast chokepoint (see residuals) and is a **tracked follow-up, not claimed here**. Strictly better than the pre-fix state (no guard at all).

Repro (before): two identical `vultisig send Polygon <addr> 0.001 --confirm` both broadcast (two hashes, nonce advanced twice), and `~/.vultisig/broadcasts.jsonl` was never written by the `send` path.

## Fix

Both paths now share **ONE** journal via the same exported helpers (`assertNoRecentDuplicate` / `reserveBroadcast` / `recordBroadcast`):

- An identical `send`/`swap` whose prior broadcast **completed and journaled** is **refused before signing** and exits **9** (`DUPLICATE_BROADCAST`) instead of broadcasting a second time.
- `send`/`swap` gain a `--force` flag to override the guard, mirroring `agent ask`.
- **`--max` fingerprints a stable sentinel** (not the drift-prone resolved amount), so a `--max` retry can't slip past the guard when fee/balance moves between attempts.
- The journal is namespaced by the vault's ECDSA key (`ecdsa || vault.id`, **fail-closed** if both empty). A native/simple-EVM `send` and an identical `agent ask` cross-dedupe against the one journal.

### Scope / known residuals
- Cross-path dedup is exact for native/EVM sends. ERC-20-token and swap cross-path dedup are out of scope — a **missed dedup, never a double-spend** (within each path, token sends and swaps still dedupe).
- **Compound-swap approval leg (the reviewer's blocker):** `guardedBroadcast` wraps the SDK's compound `vault.send()`/`vault.swap()`; a post-broadcast throw before the hash returns (e.g. a multi-leg swap approval lands but the main leg fails) leaves that broadcast unjournaled, so a fresh-process retry can re-broadcast the approval. Closing it fully requires recording at each SDK broadcast chokepoint (pushing the guard into `VaultBase.swap()`/`send()`) — a deliberately deferred larger change, documented on `guardedBroadcast`. **Tracked follow-up.**

## Tests (red-then-green)
`clients/cli/src/commands/__tests__/sendSwapDupGuard.test.ts` (13 tests): identical send/swap refused (no second broadcast, exit 9); `--force` override; distinct intents allowed; cross-path dedup vs shared journal; ERC-20 token dedup + different-token-allowed; broadcast-throws-not-journaled → retry allowed; sibling-held reservation → exit 9 without signing; **send + swap `--max` drift stability** (verified they double-broadcast without the sentinel); memo/`data` discriminator; fail-closed owner.

## Gates
- [x] `yarn test:cli` — 597 passed (37 files)
- [x] `yarn typecheck` clean
- [x] `yarn lint` clean
- [x] Changeset (`@vultisig/cli` patch)

## Summary by CodeRabbit

* **New Features**
  * Added a `--force` option to both `send` and `swap` commands.
  * Introduced duplicate-broadcast protection so repeated identical requests are refused instead of sent again.
  * Improved retry handling for “max” sends/swaps to stay consistent across attempts.

* **Bug Fixes**
  * Reduced accidental re-broadcasts caused by small amount, fee, or balance changes.
  * Enabled shared duplicate detection across related command paths for the same vault.


gomesalexandre (changes requested):
good direction on the send path, but i can't sign off on the swap claim yet.

blocking
- `clients/cli/src/core/broadcastGuard.ts` still records only after `vault.swap()` returns a tx hash. for compound swaps that means an ERC-20 approval can broadcast inside `VaultBase.swap()` before a later step throws, and this CLI layer never journals that earlier leg. a fresh retry can still rebroadcast even though the PR is framed as wiring dup-guard into direct swap.
- the PR text already documents that residual window in the new guard comment, so the current code and the stated guarantee are still out of sync.

receipt
- diff read against freshly fetched `origin/main`
- traced the current swap path through `clients/cli/src/commands/swap.ts` -> `guardedBroadcast(...)` -> `packages/sdk/src/vault/VaultBase.ts` approval leg + main leg
- the unresolved gap is the approval broadcast at `VaultBase.swap()` before the CLI can call `recordBroadcast(...)`

this needs either guard-at-broadcast-chokepoint coverage for the approval leg, or a narrower claim/scope for what this PR actually fixes.

gomesalexandre (approved):
Narrowed scope works for me ser. The claim now matches the implementation exactly: completed-broadcast path is guarded, compound-swap approval-leg window is explicitly called out as a tracked residual (not claimed). The guardedBroadcast docblock + PR description scope banner make the boundary crystal clear.

Tests are thorough for the stated scope (--max sentinel, sibling reservation, fail-closed vault owner, the broadcast-throws-not-journaled retry path - all there).

Ship it - strictly better than zero guard, and the deferred chokepoint fix can land on top without conflict. 🤖 reviewed via Claude Code (local /loop)

PR #1032, opened 2026-07-08
Cache Rujira range pair list

### LINKED ISSUE #948: perf: portfolio balance/price batching + gas & rujira-pair caching

rcoderdev:
Closes #948
Real QA: the public @vultisig/rujira/range export resolved live RUJI/RUNE twice through one RujiraRange instance and made one FIN pair-list GraphQL request to api.vultisig.com. Fresh source checks also covered sequential cache hits and concurrent in-flight sharing with one GraphQL request. Added unit coverage for concurrent in-flight lookup sharing; yarn check:ci and yarn test pass on the rebased commit.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved RUJI trade pair lookups by caching FIN pair-list results with a TTL.
  * Concurrent lookups for pair addresses now share a single in-flight request to avoid duplicate calls.
  * Cached data refreshes automatically after the cache expiry window.
* **Reliability**
  * Malformed FIN pair-list server responses are no longer cached, allowing later valid requests to succeed.
  * Added coverage for additional malformed edge payload scenarios to ensure consistent failure handling.


gomesalexandre (commented):
Conceptual pass looks good.

- `packages/rujira/src/modules/range.ts`: walked the new cache + in-flight path and didn't find a correctness hole. The cache is populated before the first await, concurrent callers share `pairListInflight`, and the `.finally()` cleanup means a failed fetch does not poison later retries.
- Receipt: `yarn vitest run packages/rujira/src/__tests__/range.test.ts` passes here with 22/22 green, including the new cache-hit / shared-inflight / TTL-refresh coverage.
- `.changeset/rujira-pairs-cache.md` is present and correctly scoped to `@vultisig/rujira`.

gomesalexandre (dismissed):
requesting changes on `packages/rujira/src/modules/range.ts:581`.

this now caches `[]` when `finV3`, `pairs`, or `edges` is missing. that turns a transient malformed/incomplete graph response into a 5 minute fail-open for every pair lookup, where the old code only poisoned a single request. please validate `finV3.pairs.edges` is actually an array before populating the cache, and let malformed responses retry instead of caching the empty fallback.

johnnyluo (approved):
Clean, well-scoped cache + in-flight de-dup for the FIN pair-list query. Errors correctly don't poison the cache (gqlFetch rejects before .then runs), concurrent sharing is genuinely synchronous, and the TTL-refresh path is sound. Changeset present and correct.

Non-blocking: consider one test for the error-retry path (rejected first fetch → cache not poisoned → second call retries), since that's the subtlest invariant to guard against future refactors.

PR #1033, opened 2026-07-08
docs(cli): fix README exit-code table to match the ExitCode enum

neavra:
## What

The `clients/cli/README.md` **Exit Codes** table (which ships in the npm package and renders on npmjs.com) had drifted from the actual exit-code taxonomy. It labelled every non-zero code wrong — e.g. `2` was documented as "invalid usage" but the CLI exits `2` for authentication-required, and `3` was "configuration error" but the CLI exits `3` for a retryable network error. A script branching on `$?` per the README would misclassify every failure. Codes `8`–`11` (`ACK_FAILED`, `DUPLICATE_BROADCAST`, and the two `agent ask` turn-outcome codes) were undocumented.

## Changes

- Regenerate the README exit-code table from the `ExitCode` enum / `EXIT_CODE_DESCRIPTIONS` in `src/core/errors.ts` — the same single source of truth used by `vultisig --help` and the `schema` command.
- Add a doc-lint test (`src/core/readme-exit-codes.test.ts`) that parses the README table and fails if it drifts from the enum (missing codes, extra codes, or a description mismatch).
- Correct the "disable colored output" env var in the README to `NO_COLOR` (the cross-tool standard the CLI honours; `VULTISIG_NO_COLOR` still works too).

## Testing

- `yarn test:cli` — all 564 tests pass; the new doc-lint test fails against the old table and passes after the fix.
- `yarn workspace @vultisig/cli typecheck`, lint, and `format:check` all clean.

Changeset: `@vultisig/cli` patch (README ships in the package).

## Summary by CodeRabbit

* **Documentation**
  * Updated the CLI README with clearer, multi-line JSON examples for agent and balance output.
  * Expanded the “Exit Codes” reference with a complete code-to-meaning table and corrected several mislabeled codes.
  * Clarified the recommended environment variable for disabling colored output, while keeping the older option noted.

* **Tests**
  * Added a check that keeps the README exit-code table aligned with the documented values, helping prevent future drift.


gomesalexandre (approved):
docs-only pass looks good.

- README exit-code table matches `clients/cli/src/core/errors.ts` / `EXIT_CODE_DESCRIPTIONS`, including codes 8-11.
- `NO_COLOR` docs line up with the actual config behavior while keeping `VULTISIG_NO_COLOR` compatibility.
- the new doc-lint test resolves the shipped CLI README path and should be picked up by the existing vitest run.

conceptual approve only. i did not run tests here.

PR #1035, opened 2026-07-08
fix(cosmos): floor Osmosis fees at the live x/txfees dynamic base fee

gomesalexandre:
Tackles VA-100 (txsafe/L0), a follow-up finding from the three-tx-builder reconciliation spike (VA-85).

## what

`getCosmosFeeAmount` (the canonical cosmos gas resolver used by the CLI/browser/electron keysign pipeline) only queried the generic `/cosmos/base/node/v1beta1/config` node config for every ibc-enabled chain. That's a per-node/operator-configurable `minimum-gas-price` value - it does NOT track Osmosis's protocol-level EIP-1559 dynamic base fee (enforced by the `x/txfees` module), and the resolver's own anomaly guard (clamps to the static floor when the computed fee looks "too high" vs a 10x multiplier) can actively discard a genuinely-required higher fee during a real base-fee spike.

Live-verified incident (already fixed in vultiagent-app): a broadcast failed with code 13 "insufficient fees" because the live base fee had risen to 0.03, requiring ~12000uosmo - above what the generic mechanism produced.

## how

Adds `getOsmosisDynamicFeeFloor` (`osmosisDynamicFee.ts`), querying Osmosis's own `/osmosis/txfees/v1beta1/cur_eip_base_fee` endpoint directly with a 1.25x headroom, mirroring the app's proven fix. `getCosmosFeeAmount` now runs the existing generic-config lookup and this Osmosis-specific lookup concurrently (for Osmosis only) and takes the max - every other cosmos chain's behavior is byte-identical to before this change.

## testing

- `yarn test:core` - 166 test files / 1438 tests, all green
- `yarn typecheck` / `yarn lint` - clean
- Codex adversarial review: 3 real findings, all tackled in a follow-up commit - timeout wasn't fail-open if `fetchImpl` ignored the abort signal (now uses the same `Promise.race` pattern as the existing `fetchMinGasPrice`), `base_fee` was coerced via loose `Number()` (silently accepting `null`/`''`/`[]`/`true`) instead of strict decimal validation, and the two lookups ran sequentially instead of concurrently (up to 8s worst-case latency for Osmosis specifically, now parallelized)

## risk

Low - purely additive for Osmosis, no behavior change for any other cosmos chain. Fail-open on every error path (malformed response, non-ok status, network error, timeout, external abort) falls back to the pre-existing generic-config result, never blocks a send.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cosmos fee estimation for Osmosis by incorporating the network’s dynamic base-fee data (with safe fallbacks) alongside the standard node configuration, reducing the chance of broadcast rejections during fee spikes.
  * Added exact decimal-based fee math to avoid precision loss when computing gas-price fees.
* **Tests**
  * Expanded coverage for Osmosis dynamic fee-floor behavior, including error/timeout handling, chain isolation, and sanity-ceiling scenarios.
  * Added tests for precise decimal parsing and rounding in fee calculations.


packages/core/chain/chains/cosmos/osmosisDynamicFee.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Live fee floor for Osmosis (base-denom units for `gasLimit`), or null when
+ * the lookup failed (fail-open - caller should fall back to its own static
+ * or generically-derived fee).
+ */
+export async function getOsmosisDynamicFeeFloor(gasLimit: bigint, opts: FetchOpts = {}): Promise<bigint | null> {
+  const baseFee = await fetchOsmosisBaseFeePerGas(opts)
+  if (baseFee === null) return null
+  return BigInt(Math.ceil(Number(gasLimit) * baseFee * OSMOSIS_BASE_FEE_HEADROOM))
```
NeOMakinG: preferably-blocking: `BigInt(Math.ceil(Number(gasLimit) * baseFee * OSMOSIS_BASE_FEE_HEADROOM))` has no upper bound and no guard. Two reproduced failure modes from a bad/compromised LCD (verified against this exact code):

- `base_fee="1000"` passes `decimalPattern` -> returns `375_000_000n` uosmo (375 OSMO) with no clamp; the caller takes `max(...)` so this overpay lands in the signable fee. The generic path's `floor * maxLiveFeeMultiplier` guard (gas.ts) does NOT cover this branch.
- A regex-valid pathological decimal (`"9".repeat(400)`) makes `Number(base_fee)`/the product `Infinity`, so `BigInt(Infinity)` throws `RangeError`. There's no try/catch here, so it rejects the `Promise.all` in `getCosmosFeeAmount` and throws the whole fee resolution -- violating the documented "fail-open ... never blocks a send" contract.

Suggest reusing the exact BigInt decimal math from gas.ts (`parseDecimal` + `getFeeAmountFromGasPrice`) instead of `Number` (removes the overflow) and applying a generous sanity ceiling relative to `cosmosGasRecord[Osmosis]`; at minimum reject non-finite results and `return null` so this path stays fail-open.

packages/core/chain/chains/cosmos/gas.ts
```diff
@@ -150,3 +148,23 @@ export const getCosmosFeeAmount = async (
...
+  // configurable value that doesn't track the live protocol floor, and can
+  // be clamped away by the anomaly guard above when it legitimately spikes).
+  // Run both lookups concurrently (each has its own timeout budget) rather
+  // than sequentially, and never pay less than the higher of the two -
+  // see osmosisDynamicFee.ts.
+  const [genericFee, dynamicFloor] = await Promise.all([
+    getGenericCosmosFeeAmount(coin, opts),
+    getOsmosisDynamicFeeFloor(getCosmosGasLimit(coin), opts),
+  ])
+  return dynamicFloor !== null && dynamicFloor > genericFee ? dynamicFloor : genericFee
```
NeOMakinG, line 151: should-fix: `max(genericFee, dynamicFloor)` applies no ceiling to `dynamicFloor`, so any value the Osmosis LCD returns flows straight into the signable fee. The generic branch is protected by the 10x `maxLiveFeeMultiplier` clamp (line 144) but that protection is bypassed for the dynamic floor. Bound `dynamicFloor` here (or inside `getOsmosisDynamicFeeFloor`) so a spiked/malicious base fee can't be taken verbatim.

NeOMakinG (commented):
Reviewed as a fund-critical cosmos fee change: this value feeds the **signable** fee in the keysign pipeline (`mpc/keysign/chainSpecific/resolvers/cosmos.ts:69 gas: await getCosmosFeeAmount(coin)` and `buildCosmosPayload.ts:103`), so any over/underpay lands in a real broadcast tx. Ran a primary correctness lane + a Codex gpt-5.4 adversarial lane in parallel.

## testing approach
- Traced the full path: `cur_eip_base_fee` fetch → decimal validation → `gasLimit * base_fee * 1.25` floor → `max(genericFee, dynamicFloor)` → signable `gas` field.
- Confirmed Osmosis-scoping is a pure extraction: `getGenericCosmosFeeAmount` is the old `getCosmosFeeAmount` body verbatim; every non-Osmosis chain is byte-identical (test-locked by the "does not query the Osmosis txfees endpoint for any other cosmos chain" case).
- Reproduced the two adversarial branches locally in node (see `## blocking`).
- CI at review time: Integration Tests + CodeRabbit **pass**; Lint/Unit **pending** (PR body reports `yarn test:core` 1438 tests green locally). Changeset present (`@vultisig/sdk: patch`) — not shipping inert.

## blocking
_(none that block the honest-path fix — the core fix is correct and valuable; see preferably-blocking for the adversarial-endpoint gap)_

## preferably-blocking
The new Osmosis dynamic-floor branch **fully trusts the `base_fee` magnitude from the LCD with no upper sanity bound**, unlike the generic path which explicitly guards this exact threat (`if (computedFee > floor * maxLiveFeeMultiplier) return floor`, gas.ts:144). This produces two concrete failure modes from a compromised/MITM'd/buggy Osmosis endpoint (`getCosmosRpcUrl` also honors a user `customRpcOverride`), both reproduced locally against the exact code:

1. **Unbounded overpay into the signable fee.** `{"base_fee":"1000"}` passes `decimalPattern`, yields `300000 * 1000 * 1.25 = 375_000_000n` uosmo (375 OSMO), and `getCosmosFeeAmount` returns it as `max(genericFee, dynamicFloor)` with no clamp. The generic branch's 10x anomaly guard never applies here.

2. **Fail-open contract violation → signing DoS.** `getOsmosisDynamicFeeFloor` (osmosisDynamicFee.ts:89-92) has **no** try/catch around `BigInt(Math.ceil(Number(gasLimit) * baseFee * HEADROOM))`. A regex-valid but pathological decimal (`"9".repeat(400)`, or `"1"+"0".repeat(306)`) makes `Number(base_fee)`/the product overflow to `Infinity`, so `BigInt(Infinity)` throws `RangeError`. Because Osmosis now uses `Promise.all`, that rejection propagates out of `getCosmosFeeAmount` and **throws the whole fee resolution** — directly contradicting the PR's stated contract ("Fail-open on every error path … never blocks a send"). Today's generic-only path never throws, so this is a net-new fail-closed-to-crash for Osmosis.

Both stem from one root cause and one fix closes them: bound/validate the computed floor before returning. Cleanest approach — mirror the generic path's **exact BigInt decimal math** (reuse `parseDecimal` + `getFeeAmountFromGasPrice` from gas.ts instead of lossy `Number`), which removes the `Number`→`Infinity` overflow entirely, then apply a generous sanity ceiling (e.g. clamp the dynamic floor to some multiple of `cosmosGasRecord[Osmosis]` chosen above realistic base-fee spikes) so a bad endpoint value can neither overpay nor crash. At minimum, reject non-finite results inside `getOsmosisDynamicFeeFloor` so it returns `null` (fail-open) rather than throwing.

## should-fix
Precision: the dynamic path uses IEEE-754 `Number` while the generic path uses exact rational arithmetic (`parseDecimal` → numerator/denominator + ceil-division). For a regex-valid `base_fee` with >~15 significant digits the `Number` path can underquote by ~1 uosmo (e.g. an input whose exact ceil is `90001` returns `90000`). Negligible for today's 18-dp Osmosis shape, but reusing the exact path (same fix as above) makes it correct and consistent — and is the natural way to also kill the overflow crash.

## q
The PR says this "mirrors a fix already live in vultiagent-app." Does the app version apply an upper sanity cap on the dynamic base fee? If so, worth porting that bound here for parity; if not, the app has the same overpay/crash exposure and may want the same follow-up.

## risk
Medium. The honest-endpoint behavior is correct and fixes a real production stuck-tx incident (code 13 insufficient fees) — that's a genuine improvement and the Osmosis-scoping/no-regression is solid. The residual risk is narrow (requires a compromised/malfunctioning LCD returning a regex-valid pathological value), but it lands in a **signable fee** on a fund-critical path, the code's own documented fail-open contract is violated by branch (2), and the generic path already defends this exact class — so the new path shouldn't ship strictly less-guarded than the one it augments.

## verdict
Not approving pending the sanity-bound + fail-open guard on the dynamic floor. The core fix (live `x/txfees` base-fee floor, `max`-not-overwrite, correct denom/decimals for honest values, fail-open to the pre-existing generic path on ordinary errors, Osmosis-scoped, changeset present, strong test coverage) is correct and worth landing — just close the unbounded-trust gap so a bad endpoint value can't overpay the signable fee or crash fee resolution.

---
🤖 dual-lane review: Claude primary + Codex gpt-5.4 adversarial (both independently flagged the unbounded overpay; Codex additionally surfaced the `BigInt(Infinity)` fail-open violation).

packages/core/chain/chains/cosmos/osmosisDynamicFee.ts
gomesalexandre: tackled in bcc63e53 (fix commit 17c845b5 + test-coverage follow-up): computed the floor with exact BigInt arithmetic (extracted parseDecimal/getFeeAmountFromGasPrice into a shared cosmosDecimal.ts to avoid a circular import), eliminating the Number-overflow-to-Infinity path entirely, and added a 1000x sanity ceiling (9_000_000n uosmo) that discards a pathological value rather than clamping it, falling open to the generic-config result. New tests cover both the unbounded-overpay case and the 400-digit base_fee case (at both the internal helper and the public getCosmosFeeAmount/Promise.all surface where the original crash reproduced).

packages/core/chain/chains/cosmos/gas.ts
gomesalexandre: tackled in bcc63e53 (fix commit 17c845b5) - see the reply on the osmosisDynamicFee.ts thread for the full fix.

packages/core/chain/chains/cosmos/osmosisDynamicFee.ts
```diff
@@ -0,0 +1,118 @@
...
+// resolution indefinitely.
+const OSMOSIS_BASE_FEE_TIMEOUT_MS = 5_000
+// Sanity backstop on the computed floor itself, independent of the fetch
+// succeeding: the live-verified real incident this guards against needed
+// ~12_000 uosmo against a 9_000n static floor (~1.3x) - 1000x is enormous
+// headroom above any realistic base-fee spike while still bounding a
+// pathological/malicious LCD response (e.g. a `base_fee` of "1000" would
+// otherwise compute a ~375 OSMO fee for a standard send).
+const OSMOSIS_DYNAMIC_FLOOR_MAX_MULTIPLIER = 1_000n
+const OSMOSIS_STATIC_FLOOR = 9_000n // must match cosmosGasRecord[Chain.Osmosis] in gas.ts
```
NeOMakinG, line 44: `should-fix:` `OSMOSIS_STATIC_FLOOR = 9_000n` is a hand-copied duplicate of `cosmosGasRecord[Chain.Osmosis]`. The comment flags the coupling, but nothing enforces it — if the canonical floor in `gas.ts` is raised later and this copy is missed, the ceiling here silently tightens below intent and can start discarding *legitimate* dynamic fees. Import the record entry (or a shared constant) instead of duplicating the literal so drift is a compile-time failure, not a review-time catch. Non-blocking.

packages/core/chain/chains/cosmos/osmosisDynamicFee.ts
```diff
@@ -0,0 +1,118 @@
...
+  const baseFee = await fetchOsmosisBaseFeePerGas(opts)
+  if (baseFee === null) return null
+
+  const withHeadroom: ParsedDecimal = {
+    numerator: baseFee.numerator * OSMOSIS_BASE_FEE_HEADROOM.numerator,
+    denominator: baseFee.denominator * OSMOSIS_BASE_FEE_HEADROOM.denominator,
+  }
+  const fee = getFeeAmountFromGasPrice(gasLimit, withHeadroom)
+
+  if (fee > OSMOSIS_STATIC_FLOOR * OSMOSIS_DYNAMIC_FLOOR_MAX_MULTIPLIER) return null
```
NeOMakinG, line 115: `q:` This ceiling (9 OSMO = ~750x the one live-verified incident of 12000 uosmo) is the right call for a **signable** fee — bounding a compromised/MITM'd LCD's overpay outweighs covering an implausibly large legit spike. Just confirming the liveness cost is conscious: above 9 OSMO you fail open to the generic path (capped at `9000n*10`), which reproduces the exact code-13 stuck-tx this PR fixes. Is 1000x the intended long-term headroom, or a starting point to revisit against observed mainnet spikes? Accepted trade, not a blocker.

NeOMakinG (approved):
r2 re-review at HEAD `bcc63e53`. The HEAD advance is a **real fix**, not merge-only — commit `17c845b5` ("bound the Osmosis dynamic fee floor against a compromised/pathological LCD response") + test-pin `bcc63e53` directly close both r1 preferably-blocking findings. Approving.

## testing approach

Two lanes (Claude load-bearing + Codex gpt-5.4 adversarial second opinion). Static verification against a detached worktree at `bcc63e53` (no full install — tmpfs env). Walked the exact-rational math paths by hand, confirmed the ceiling constants against `cosmosGasRecord`/`cosmosGasLimitRecord`, and read the full failing-CI job log to attribute the one red check. Codex referenced PR #1035/osmosis explicitly (not cross-contaminated).

## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | Dynamic-floor branch trusted LCD `base_fee` magnitude with **no upper bound** -> `base_fee="1000"` yields a 375 OSMO fee taken verbatim into the signable fee | ✅ fixed | `17c845b5`: `if (fee > OSMOSIS_STATIC_FLOOR(9000n) * OSMOSIS_DYNAMIC_FLOOR_MAX_MULTIPLIER(1000n)) return null` caps the dynamic floor at 9 OSMO and fails open to the generic path. Verified: `base_fee="1000"` -> discarded (`gas.test.ts` "falls back … when the dynamic floor exceeds its own sanity ceiling" -> 30_000n). |
| r1 | 2 | preferably-blocking | Regex-valid pathological decimal -> `Number` overflow -> `BigInt(Infinity)` **RangeError uncaught**, rejecting `Promise.all` and crashing the whole `getCosmosFeeAmount` (fail-closed-to-crash, violates the documented fail-open contract) | ✅ fixed | `17c845b5`: lossy `Number` math replaced by exact-rational BigInt (`cosmosDecimal.parseDecimal` + `getFeeAmountFromGasPrice`). No `BigInt(Infinity)` path exists anymore — `'9'.repeat(400)` parses to an exact BigInt, exceeds the ceiling, returns `null`. Pinned at both the helper and the public `getCosmosFeeAmount`/`Promise.all` surface (`bcc63e53`). Denominator is always >= `4n` (headroom x4), so no div-by-zero either. |
| r1 | 3 | should-fix | Dynamic path used IEEE-754 `Number`; generic path used exact rational — regex-valid `base_fee` >~15 sig-digits could underquote ~1 uosmo | ✅ fixed | Same exact-rational refactor. `cosmosDecimal.test.ts` pins the many-significant-digit case (`0.300000000000000001` -> `90_001n`, not `90_000n`). |

All three r1 findings resolved by the same single well-chosen fix (reuse the generic path's exact BigInt decimal math + a generous sanity ceiling) — exactly the direction recommended in r1.

## should-fix

- `OSMOSIS_STATIC_FLOOR = 9_000n` is a hand-copied duplicate of `cosmosGasRecord[Chain.Osmosis]`. The comment flags the coupling but nothing enforces it: if the canonical Osmosis floor in `gas.ts` is ever raised, this copy silently desyncs and the ceiling tightens below intent — potentially discarding legitimate dynamic fees. Importing the record entry (or the floor constant) instead of copying the literal would make this drift a compile-time concern rather than a review-time one. (non-blocking; Codex concurred)

## suggestion

- The new subpath exports `./chains/cosmos/cosmosDecimal` and `./chains/cosmos/osmosisDynamicFee` publish what read as internals of the Osmosis fix onto `@vultisig/core-chain`'s public surface, creating a semver obligation for helpers calibrated to internal assumptions (send gas limit, copied 9000n floor, `null`-on-breach). If no external consumer needs them, keeping them unexported (import directly within the package) would avoid the commitment. (Codex LOW)

## q

- The 9 OSMO sanity ceiling is deliberately ~750x the one live-verified incident (12000 uosmo), so a *legitimate* base-fee condition needing >9 OSMO for a standard send would fail open to the generic path (capped at `9000n x 10`) and reproduce the exact code-13 stuck-tx this PR fixes. That trade — bounding attacker/MITM overpay vs. covering an implausibly large legit spike — is the right call for a **signable** fee, and the multiplier is generous. Flagging only so the ceiling's liveness cost is a conscious, documented choice rather than an implicit one. Is 1000x the intended long-term headroom, or a starting point to revisit against observed mainnet spikes? (Codex raised the same as medium; I read it as an accepted trade, not a blocker.)

## risk

Low. Osmosis-scoped (`coin.chain !== Chain.Osmosis` returns the byte-identical generic path — test-locked). Fail-open on every error/timeout/ceiling-breach path; the fee resolver can no longer throw for Osmosis. Signable-fee wiring unchanged. Changeset present (`@vultisig/sdk` patch, not inert). Honest path re-confirmed at new HEAD (0.03 base fee -> 11250n, matching the production incident).

**CI note:** the one red check (Unit Tests / Test Summary) is a single unrelated flake — `findSwapQuote.recipient.test.ts` "routes the output to the recipient via CowSwap" **timed out at 5010ms** (network-dependent swap-provider call). The PR touches zero swap files (`grep -c findSwapQuote` on the diff = 0); all 1447 other unit tests pass, including every cosmos/osmosis gas test, and Lint+Type, Integration (Vitest), and CodeRabbit are green. Not a regression from this PR.

## verdict

**Approve.** Both r1 preferably-blocking findings are genuinely resolved by exact-rational math + a bounded, fail-open sanity ceiling; the ceiling is sensible and does not re-break the legitimate-spike case the fix exists for; the non-finite guard returns `null` rather than throwing; the honest floor, decimals, Osmosis-scoping, and changeset all still hold. Remaining items (static-floor drift, export surface, ceiling liveness trade) are non-blocking follow-ups.

---
dual-lane review — Claude (load-bearing) + Codex gpt-5.4 (adversarial). Round 2.

PR #1037, opened 2026-07-08
fix(swap): check every cosmos leg for the memo cap in runSkipSwap (VA-86)

gomesalexandre:
Tackles VA-86 (txsafe/L0), a follow-up finding from reconciling vultisig-sdk's `skipSwap.ts` against agent-backend-ts's independent Skip integration (`skip-swap.ts`).

## what

The two aren't simple duplicates - they're built for different consumption models (env-var-driven Node service vs argument-driven portable library), and a full collapse into one is real multi-PR architecture work being parked separately. But one piece had genuinely drifted and regressed: the memo-length preflight.

`runSkipSwap`'s memo-cap check only ran `if (!isMultiTx)` and only inspected the FIRST cosmos leg. For a single-tx route that's harmless (there's only one leg to check anyway), but for ANY multi-tx route (`allowMultiTx:true`), the check never ran at all - a later leg whose memo exceeded its own chain's `MaxMemoCharacters` cap would sail through undetected and fail at broadcast (sdk error code 12, "memo too long") AFTER signing, burning the MPC ceremony.

agent-backend-ts's own Skip integration already runs this check unconditionally against every leg (its own comment there references a "previous fail-open" bug already fixed).

## how

Ports that same hardening into the SDK's copy: `getCosmosLegMemoInfos` now collects every cosmos leg, `firstCosmosLegOverMemoCap` checks each against its own chain's cap, and the call site no longer gates on `isMultiTx`. Also switches memo byte-counting from `Buffer.byteLength` to `TextEncoder` for RN/browser portability, matching the SDK's established pattern elsewhere.

## testing

- New regression tests prove the exact gap: a multi-tx route with an over-cap memo on the second (non-first) leg is now rejected - verified via `git stash` that it would have been silently accepted before this fix - and a multi-tx route where every leg fits is still accepted.
- `yarn workspace @vultisig/sdk test:unit` - 167 files / 2378 tests, all green
- `yarn typecheck` / `yarn lint` - clean
- Codex adversarial review: 1 real finding - removing the `isMultiTx` gate meant a malformed multi-tx leg (`cosmos_tx: null`) would now crash instead of falling through to the existing structured `skip_msgs_tx_malformed` error. Fixed with a guard + a third regression test proving the leg is skipped without crashing.

## risk

Low - additive hardening, no change for chains/routes that already fit their cap. Existing single-tx behavior is unchanged in substance (there was always only one leg to check there).

### DIFF
--- a/.changeset/skipswap-multitx-memo-cap.md
+++ b/.changeset/skipswap-multitx-memo-cap.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Fix `runSkipSwap`'s cosmos memo-length preflight to check every leg of a Skip route, not just the first leg of single-tx routes. Previously the check only ran when the route required a single signature and only inspected the first cosmos leg's memo — a multi-tx route (`allowMultiTx: true`) with an over-cap memo on a later leg sailed through undetected and would fail at broadcast (sdk error code 12, "memo too long") after signing, burning the MPC ceremony. This mirrors a check already present and more thorough in agent-backend-ts's own Skip integration.
--- a/packages/sdk/src/tools/swap/skip/skipSwap.ts
+++ b/packages/sdk/src/tools/swap/skip/skipSwap.ts
@@ -224,22 +224,47 @@ function firstUnsupportedCustodyChain(
 }
 
 /**
- * Extract the cosmos source-leg SDK transaction-level memo byte length from a
- * Skip /msgs_direct response, or `null` when no cosmos leg exists. cosmoshub-4 /
- * columbus-5 enforce the cap against the SDK transaction-level memo
- * (`cosmos_tx.memo`), NOT the inner ICS-20 packet memo (which is unbounded).
+ * Extract the SDK transaction-level memo byte length for EVERY cosmos leg in a
+ * Skip /msgs_direct response. Returns one entry per cosmos tx (empty array when
+ * the route is EVM-only and the cap check doesn't apply).
+ *
+ * cosmoshub-4 / columbus-5 / etc. enforce `x/auth.MaxMemoCharacters` against
+ * the SDK **transaction**-level memo (`cosmos_tx.memo`), NOT against the inner
+ * ICS-20 packet `memo` (which is unbounded).
  */
-function getSourceLegMemoByteLength(
+function getCosmosLegMemoInfos(
   txs: ReadonlyArray<SkipMsgsDirectTx>
-): { sourceChainId: string; memoBytes: number } | null {
+): Array<{ sourceChainId: string; memoBytes: number }> {
+  const out: Array<{ sourceChainId: string; memoBytes: number }> = []
   for (const tx of txs) {
     if (!tx || !('cosmos_tx' in tx)) continue
     const cosmosTx = tx.cosmos_tx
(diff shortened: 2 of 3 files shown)


NeOMakinG (approved):
## testing approach

Two-lane review (Claude load-bearing + Codex gpt-5.4 adversarial second opinion). Traced the leg iteration against PR head `b5acddeb` (sparse-checkout of `skipSwap.ts` + `cosmosMemoCap.ts`), cross-referenced the #1022 cap-table / inner-memo groundwork and the #849 consumer-side review. No device (pure validation-path logic). Verified CI conclusions via check-runs API.

## verdict

**Approve.** This closes the VA-86 gap cleanly. The old preflight ran only `if (!isMultiTx)` and inspected only the first cosmos leg, so any `allowMultiTx:true` route with an over-cap memo on a later leg slipped through and would fail at broadcast (sdk code 12) *after* signing — burning the MPC ceremony. The fix is correct on every fund-safety axis:

- **Every cosmos leg checked** — `getCosmosLegMemoInfos` collects all `txs`; `firstCosmosLegOverMemoCap` iterates every leg; the `isMultiTx` gate is gone from the call site. EVM legs correctly skipped via `!('cosmos_tx' in tx)`.
- **Correct per-chain cap per leg** — each leg resolves `getCosmosMemoMaxBytesByChainId(memoInfo.sourceChainId)` against its OWN chain-id (Osmosis 256 vs Cosmos Hub 512). The helper always returns a number (fail-closed to 256 for unknown ids), so dropping the old `cap !== undefined` guard is safe — it was dead code.
- **Validation-only, no truncation** — returns the `skip_source_memo_too_long` error envelope and rejects the swap. No `.slice()`, no mutation. Fail-closed.
- **Right memo (x/auth-governed outer `cosmos_tx.memo`)** — not the unbounded inner ICS-20 packet memo, and not a THOR/Maya inner `MsgDeposit` memo — so no false-reject of a valid routing memo.
- **No regression / no crash** — single-tx routes still validate their one leg; the `!cosmosTx || typeof cosmosTx.chain_id !== 'string'` guard skips a malformed `cosmos_tx: null` leg without throwing, leaving `validateTxEnvelopes` to surface the structured `skip_msgs_tx_malformed` error (regression test proves it).
- **Byte-counting** — `Buffer.byteLength` → `TextEncoder().encode().length` matches on-chain UTF-8 byte semantics and the SDK's established RN/browser-safe pattern.

**Changeset present** (`.changeset/skipswap-multitx-memo-cap.md`, `@vultisig/sdk` patch) — ships live, not inert. **CI green**: Unit, Integration, Lint and Type Check, Test Summary all success; CodeRabbit pass.

Test coverage is on-point: over-cap non-first leg rejected (columbus-5 300B > 256), every-leg-fits accepted (256 exactly), malformed-leg no-crash.

## risk

Low. Additive hardening on a validation-only preflight; no behavior change for routes that already fit their cap. Codex second-opinion lane returned NONE (no bug), independently confirming: no unchecked leg, no truncation, no crash regression, and the outer x/auth-governed memo is the one being capped.

---
🤖 dual-lane review (Claude + Codex gpt-5.4). Fund-safety focus: every cosmos leg / per-chain cap / validation-only / outer-vs-inner memo.

Issue #983, opened 2026-07-06
test(ripple): cross-platform signing interop vector + money-path unit coverage (reserve, dest-tag)


realpaaao:
## What
XRP has **no cross-platform signing parity vectors** and thin money-path unit coverage (no reserve/destination-tag tests). MPC co-signing depends on byte-identical serialization across SDK/iOS/Android/Windows.

## Fix
Add an XRP interop fixture: a fixed (address, sequence, fee, amount, destination tag, LastLedgerSequence) input with the expected signing pre-image / serialized bytes, and assert each platform produces identical bytes. Seed money-path unit tests for reserve (owner-aware) and destination-tag handling.

## Acceptance
- [ ] Shared XRP interop vector (input → expected serialized signing bytes)
- [ ] SDK test asserts the vector; iOS/Android reference the same fixture
- [ ] Reserve (owner-aware) + destination-tag unit tests exist

Ref: interop-test recipe (replicate byte recipe across platforms). Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

## References
- Reserves (owner-aware): https://xrpl.org/docs/concepts/accounts/reserves
- Source & destination tags: https://xrpl.org/docs/concepts/transactions/source-and-destination-tags
- Reliable submission (LastLedgerSequence, sequence): https://xrpl.org/docs/concepts/transactions/reliable-transaction-submission

PR #1038, opened 2026-07-08
Add Ripple signing interop vector

rcoderdev:
Adds an SDK-side Ripple signing-input interop fixture and regression test pinning destination tag, sequence, fee, amount, LastLedgerSequence, serialized WalletCore SigningInput bytes, and pre-signing hash.

Real QA: generated the pinned Ripple vector from the SDK runtime path and verified the expected account, destination, amount, destination tag, sequence, fee, LastLedgerSequence, serialized signing input, and pre-signing hash.

Refs #983

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Ripple/XRP signing interoperability, including transactions with destination tags.
  - Verified stable serialized signing inputs and pre-signing hashes across platforms.

- **Tests**
  - Added cross-platform Ripple payment signing vectors and expanded coverage for tags, fees, sequences, and ledger limits.
  - Added safeguards to prevent test-only JSON fixtures from being included in bundled source checks.


gomesalexandre (commented):
Looks good functionally, but the changeset target is off.

- `.changeset/ripple-signing-interop-vector.md` bumps `@vultisig/sdk`, but this diff only adds a fixture + test under `packages/core/mpc/keysign/signingInputs/**`. There is no `packages/sdk/src/**` change here, so shipping an `@vultisig/sdk` patch from this PR would be a spurious downstream release.
- `packages/core/mpc/package.json` only publishes `dist`, so even `@vultisig/core-mpc` does not ship this new `.test.ts` or JSON fixture. That makes this effectively a test-only PR, so I would drop the changeset entirely unless you intentionally want a release for some other reason.
- Receipt: `yarn build:shared && yarn vitest run packages/core/mpc/keysign/signingInputs/resolvers/ripple.test.ts` passes here with 2/2 green, so the pinned vector itself looks legit.

johnnyluo (approved):

PR #1039, opened 2026-07-08
Fix secure vault relay URL signing

### LINKED ISSUE #953: fix(sdk): signing silently ignores a configured relay URL (defaults to production)
## Evidence
`packages/sdk/src/vault/SecureVault.ts:110` and `:180` both do `new RelaySigningService()` with **no argument**, even though `this.context` already carries `context.serverManager.messageRelay`. `RelaySigningService`'s constructor defaults to a hardcoded `DEFAULT_RELAY_URL = 'https://api.vultisig.com/router'` when no arg is passed (`RelaySigningService.ts:73,36`). The static factory `SecureVault.create()` does it correctly — `new SecureVaultCreationService(context.serverManager.messageRelay)` (`:356`) — proving the value is in scope and normally used, just not at the two signing call-sites.

## Impact
A consumer who configures a custom / self-hosted relay via `ServerManager` gets that override honored for vault **creation** but silently **ignored** for `vault.sign()` / `vault.signBytes()` — signing traffic goes to production `api.vultisig.com` regardless of configuration.

## Fix sketch
Pass `context.serverManager.messageRelay` into both `RelaySigningService` constructions in `SecureVault.ts`. Longer-term, standardize the `services/*` DI convention so a missing override is a type error, not a silent default.

---

rcoderdev:
Closes #953
Runtime proof: SecureVault.sign() and SecureVault.signBytes() constructed RelaySigningService with https://relay.example.test/router, and generated signing QR payloads included serverUrl=https://relay.example.test/router before external relay/MPC participation.

## Summary by CodeRabbit

* **Improvements**
  * Configured relay URLs are now honored for secure vault transaction signing and raw-byte signing.
  * QR signing payloads now retain the selected relay endpoint, including when payloads are uploaded due to size limits.
  * Local signing behavior remains unchanged when relay usage is disabled.

* **Tests**
  * Added coverage confirming custom relay URLs are preserved across QR generation and secure vault signing flows.

### DIFF
--- a/.changeset/quiet-relays-sign.md
+++ b/.changeset/quiet-relays-sign.md
@@ -1,0 +1,6 @@
+---
+"@vultisig/sdk": patch
+"@vultisig/core-mpc": patch
+---
+
+Honor the configured message relay URL for secure vault transaction and raw-byte signing, including signing QR payloads.
--- a/packages/core/mpc/keysign/cosigner.ts
+++ b/packages/core/mpc/keysign/cosigner.ts
@@ -98,6 +98,11 @@ async function parseKeysignQrUrl(url: string) {
   return fromBinary(KeysignMessageSchema, await decompressData(match[1]))
 }
 
+function getQrQueryParam(url: string, param: string): string | undefined {
+  const match = url.match(new RegExp(`[?&]${param}=([^&]+)`))
+  return match ? decodeURIComponent(match[1]) : undefined
+}
+
 /** Poll GET /start/{sessionId} until the session starts. */
 async function waitForSessionStart(serverUrl: string, sessionId: string, timeoutMs = 120_000): Promise<string[]> {
   const t0 = Date.now()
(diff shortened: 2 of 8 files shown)


gomesalexandre (approved):
clean fix, three-layer propagation is exactly right.

**traced the signing path:**
- `SecureVault` was constructing `new RelaySigningService()` with no args - so whatever relay URL the user configured was silently dropped
- `RelaySigningService.createSigningQr` had `this.relayUrl` but wasn't passing it to `getJoinKeysignUrl` - double ignore
- `getJoinKeysignUrl` itself had no mechanism to accept a custom URL at all

the fix threads the URL through all three layers and also handles it on the cosigner side (reading `serverUrl` from QR URL query params in `cosigner.ts` instead of falling back to a hardcoded relay). correct.

**tests:** ran `yarn workspace @vultisig/sdk test` on the branch - 2378 tests pass across 167 files. `getJoinKeysignUrl` tests assert the custom relay URL lands in both the QR deep-link params and in the oversized-payload upload call. `SecureVault.sign()` + `signBytes()` tests assert `RelaySigningService` is constructed with the configured URL. coverage matches the bug surface.

**changeset:** `@vultisig/sdk: patch` + `@vultisig/core-mpc: patch` - correct, both packages changed.

**branch:** 1 commit behind main (just #1037 merge) - no conflicts expected, auto-merge will handle it.

---
*🤖 reviewed via Claude Code (local /loop daily sweep)*

PR #1041, opened 2026-07-08
feat(ci): schedule fast-signing E2E as a daily cron

gomesalexandre:
tl;dr the real 2-of-2 MPC signing ceremony has zero automated drift detection today - unit/integration tests mock it entirely, and the one suite that exercises it for real (no-broadcast) has never been wired to run anywhere.

## what

Layer 4 of the txsafe defense plan (layers 1-3 already shipped this cycle: payload-field scorer, cosmos fee-denom allowlist, memo-cap hardening). Adds `.github/workflows/e2e-fast-signing-cron.yml`:

- `schedule: 0 6 * * *` (daily) + `workflow_dispatch` for manual runs
- runs `yarn workspace @vultisig/sdk test:e2e:fast-signing` (`tests/e2e/fast-signing.test.ts`) - signs real transactions across BTC/LTC/ETH+ERC20/THORChain/Cosmos/Solana/Polkadot/Sui/QBTC via VultiServer 2-of-2 MPC, never broadcasts
- Discord failure notify, reusing the existing soft-fail pattern from `notify-discord.yml`

## why this repo, not Hermes-cron

Checked what's already here first: unlike abts (which has zero CI/CD - confirmed via empty `statusCheckRollup` on merged PRs, no workflows dir), vultisig-sdk has full GitHub Actions (`test.yml`, `integration-test.yml`, `release.yml`, ...). Adding a `schedule:` trigger is native infra here, not a gap needing Hermes-cron treatment.

## the actual gap this surfaces (not fixed in this PR)

`fast-signing.test.ts` needs a **funded, disposable, multi-chain fast vault** to do anything beyond skip:

- `HAS_TEST_VAULT_FIXTURE` gates the whole suite on a `.vult` file existing at `TEST_VAULT_PATH` - the fixture was deliberately removed from git for security (`tests/e2e/SECURITY.md`, commit `e3811eea`)
- Some chains (Bitcoin/Litecoin/Polkadot/Sui/QBTC) tolerate zero balance and skip cleanly via `e2ePrepareSendSkipReason`; others (Ethereum, THORChain, Cosmos Hub, Solana) call `prepareSendTx` unguarded and will hard-fail without real funds
- The repo already has an org secret `TEST_VAULT_PASSWORD` (used by `integration-test.yml` for a *different* vault/purpose - a raw JSON copied from agent-backend's testdata, not this suite's `.vult` format) - there's no equivalent secret carrying an actual `.vult` file today

This workflow is deliberately safe to merge as-is: the vault-fixture step soft-fails (warns, doesn't error) when `E2E_TEST_VAULT_FILE_BASE64` is unset, so the job stays green with all 16 tests cleanly skipped - **verified locally, exit 0**. Provisioning (or reusing an already-funded disposable dev vault) and wiring that secret is a real-funds/infra decision I'm flagging rather than doing unilaterally - happy to action it given a go-ahead on which vault to use.

## test plan

- [x] YAML validated (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] Locally reproduced the no-secret path: `yarn build:shared` + `vitest run --config tests/e2e/vitest.config.ts tests/e2e/fast-signing.test.ts` with no `TEST_VAULT_PATH`/`TEST_VAULT_PASSWORD` set → `16 skipped (16)`, exit 0

🤖 Agent-generated


.github/workflows/e2e-fast-signing-cron.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: Notify Discord on failure
+        if: failure()
+        env:
+          DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK_URL }}
+        run: |
+          if [ -z "$DISCORD_WEBHOOK" ]; then
+            echo "::warning::DISCORD_WEBHOOK_URL not set, skipping failure notify."
+            exit 0
+          fi
+          curl -X POST "$DISCORD_WEBHOOK" \
```
NeOMakinG, line 99: `should-fix:` this `curl` has no `--fail`/status check, so a revoked or misconfigured webhook (4xx/5xx) still exits 0 with no warning — muting the cron's only out-of-band failure signal. Suggest:
```bash
if ! curl -sS --fail-with-body -X POST "$DISCORD_WEBHOOK" \
  -H "Content-Type: application/json" \
  -d "{...}"; then
  echo "::warning::Discord failure notification did not succeed"
fi
```
I prepared this exact fix but the push was rejected (OAuth app lacks `workflow` scope) — leaving it for you. Non-blocking.

NeOMakinG (approved):
## testing approach

Two-lane review (Claude load-bearing + Codex `gpt-5.4` adversarial). Non-device: CI/tooling PR, +101/-0, one new file `.github/workflows/e2e-fast-signing-cron.yml`. Verified against the PR branch worktree at `efa7c55b`, plus the referenced test/script/helper on the SDK tree:

- **Cron:** `schedule: 0 6 * * *` (daily 06:00 UTC) + `workflow_dispatch` — valid 5-field expression, sensible daily cadence, manual-trigger escape hatch. `concurrency` group prevents overlapping runs. ✅
- **E2E is real, not a no-op:** `yarn workspace @vultisig/sdk test:e2e:fast-signing` → `test:e2e:fast-signing` maps to `packages/sdk/tests/e2e/fast-signing.test.ts` (confirmed in `package.json:112`). The suite calls `vault.prepareSendTx()` → `vault.sign()` across BTC/LTC/ETH/ERC-20/THORChain/Cosmos/Solana/etc via real 2-of-2 VultiServer MPC and asserts on `messageHashes`/`signature` — genuinely exercises the fund-critical signing ceremony, never broadcasts. ✅
- **Soft-fail correctness:** `HAS_TEST_VAULT_FIXTURE = existsSync(TEST_VAULT_CONFIG.path)` (`helpers/test-vault.ts:54`) gates the suite via `describe.skipIf(!HAS_TEST_VAULT_FIXTURE)`. When no vault secret is set, the fixture step warns (doesn't write the file), so the run step skips the whole suite cleanly → green. Author's local repro (`16 skipped`, exit 0) matches the code path. ✅
- **Secret handling:** no hardcoded secret/seed/key anywhere. All material via `secrets.*` (`E2E_TEST_VAULT_FILE_B64_1..6`, `E2E_TEST_VAULT_PASSWORD`, `DISCORD_WEBHOOK_URL`). Vault is a dedicated disposable dust vault per `tests/e2e/SECURITY.md`, not a production vault. Password intentionally split from the existing `TEST_VAULT_PASSWORD`. ✅
- **Supply chain:** both third-party actions SHA-pinned (`actions/checkout@9c091bb…` v4, `actions/setup-node@48b55a…` v4). ✅
- **Injection:** only `github.repository` / `github.run_id` interpolated into the curl body — not attacker-controllable. No `pull_request_target`, no untrusted input. ✅
- **Failure signal:** Discord notify on `failure()`. ✅ (see should-fix below)

## should-fix

- **Discord notify can silently no-op** (Codex lane, confirmed). `curl -X POST "$DISCORD_WEBHOOK"` has no `--fail` / status check, so a revoked or misconfigured webhook returning 4xx/5xx still exits 0 with no warning — muting the cron's *only* out-of-band failure signal, i.e. hiding the very MPC-signing regression the job exists to catch. Suggested: `if ! curl -sS --fail-with-body -X POST … ; then echo "::warning::Discord failure notification did not succeed"; fi`. I prepared this exact fix but the push was rejected (`refusing to allow an OAuth App to … workflow … without workflow scope`) — leaving it for you to apply directly. Non-blocking.

## risk

Low. CI/tooling only, no product code path touched. Safe-by-default (green skip until a funded vault secret is wired). Wiring the real-funds disposable vault + secrets is correctly flagged as a separate infra/funds decision, not done here.

## verdict

Approve. Cron valid + sensibly timed, the E2E is real and exercises the fund-critical fast-signing MPC ceremony, zero hardcoded secrets (all `secrets.*`), disposable test vault, actions SHA-pinned, failure signal present. One should-fix (Discord `--fail`) that strengthens the alerting but doesn't block merge.

---
🤖 Two-lane review: Claude (correctness/fund-safety/secret-handling) + Codex `gpt-5.4` (adversarial). Codex lane verified on-topic (PR #1041 fast-signing cron), not cross-contaminated.

PR #1044, opened 2026-07-08
test: bind Layer-1/Layer-2 golden vectors to one shared cross-encoder fixture

gomesalexandre:
Stacked on #1001 (VA-81) - this is Track B from the VA-86 reconciliation scoping.

## what

The two existing golden-vector suites - `compileTx.golden.test.ts` (Layer 1, packages/core, WalletCore/WASM) and `tx-builder-golden-vectors.test.ts` / `cosmos-send-golden-vectors.test.ts` (Layer 2, packages/sdk, RN-JS) - each already self-validate against their OWN reference (WalletCore AnySigner / cosmjs-types / @solana/web3.js), but never against EACH OTHER. Confirmed by direct diff: they used completely different, unrelated fixture values for Cosmos and Solana today.

The tx-builder drift can't be fixed by picking one encoder - Layer 1 and Layer 2 are pure-JS-vs-WASM reimplementations for two different runtimes that can't merge (the RN layer's own header says "vendored from the app, no WalletCore"). The only way to kill the drift on the encoding side is to prove the two encoders produce byte-identical output for the same logical tx.

## how

Scoped this before building rather than assuming: `packages/sdk` does NOT depend on `@vultisig/core-chain`/`core-mpc` at all, so a shared fixture MODULE across packages isn't trivial without a new workspace dependency. Both packages DO already depend on `@trustwallet/wallet-core` directly though - that's the real common ground. WalletCore derivation is deterministic, so each suite independently derives the SAME real address/pubkey from the SAME literal private-key bytes (matching Layer-1's existing `ECDSA_PRIVATE_KEY`/`EDDSA_PRIVATE_KEY` fill(1)/fill(2) convention) without needing any new cross-package dependency.

Verified live before pinning (not copied from the source under test): ran both encoders against the identical Cosmos MsgSend and Solana transfer and confirmed byte-for-byte match (SignDoc bytes + SHA256 for Cosmos, raw message bytes for Solana).

Added:
- One new assertion each to `compileTx.golden.test.ts`'s existing Cosmos/Solana tests, pinning the shared expected bytes (previously discarded from `compile()`'s `hashes` return field).
- One new "cross-encoder binding" describe each in `cosmos-send-golden-vectors.test.ts` and `tx-builder-golden-vectors.test.ts`, using the identical fixture and asserting the SAME pinned bytes - kept as NEW test blocks rather than mutating the existing self-contained tests, so those keep validating what they already validate undisturbed.

If either encoder's output ever diverges from the other, whichever suite's hardcoded expected value no longer matches its own encoder's real output fails. Each new block is commented "MUST STAY IN SYNC with `<sibling file>`".

EVM and other chains are lower priority per scoping (EVM is already transitively bound via viem on both sides); Cosmos + Solana were the two genuinely unbound, divergence-possible chains today.

## test plan

- [x] `compileTx.golden.test.ts`: 7/7 passing (was 6/6)
- [x] `cosmos-send-golden-vectors.test.ts` + `tx-builder-golden-vectors.test.ts`: 10/10 passing (was 4/4 + 4/4)
- [x] Full `packages/core` suite (post 6-commit `origin/main` merge): 168/169 files passing (1 pre-existing skip), zero regressions
- [x] Full `packages/sdk` suite: same 3 pre-existing failures as a clean checkout (MPC-engine-singleton test-isolation issue, tracked as vultisig-windows#3777), confirmed via `git stash` A/B - zero regressions from this diff
- [x] `tsc --noEmit` (root config): clean
- [x] eslint + prettier: clean

🤖 Agent-generated

## refinement (2026-07-08)

Per review feedback, replaced the "duplicate literals + MUST-STAY-IN-SYNC comment" shape with a single shared JSON fixture: `testdata/cross-encoder-golden/{cosmos-msgsend,solana-transfer}.json`, read via a plain `readFileSync` (not an import) by both `compileTx.golden.test.ts` (packages/core) and `cosmos-send-golden-vectors.test.ts` / `tx-builder-golden-vectors.test.ts` (packages/sdk). This closes the residual "someone updates one copy's pinned hex without touching the other" hole - there's now exactly one file to edit, and both suites assert against the SAME value from it.

`packages/core`'s suite derives the sender/recipient address + pubkey fresh each run via its already-loaded WalletCore/WASM instance (free there - already loaded for the Cardano/Bittensor/QBTC tests in the same file). `packages/sdk`'s unit suite reads the precomputed address/pubkey directly from the fixture instead, to avoid adding a WASM `wallet-core` dependency to what's currently a fast, pure-JS unit test file - documented inline, independently re-derivable from the private keys.

Also reverted an inadvertent `origin/main` merge that had pulled in unrelated Ripple/changeset content (should have stayed scoped to `origin/feat_golden_vector_gaps`) - caught by the first codex-review pass.

codex-review: clean on the second pass.

## derivation binding (2026-07-08)

Closed the narrow boundary flagged in review: the encoder-binding test proved `buildCosmosSendTx` produces the same SignDoc bytes as WalletCore GIVEN the same address/pubkey, but never checked that RN-JS's OWN `privkey -> pubkey -> address` derivation path actually produces the fixture's precomputed values - a divergence there (wrong sender = funds from the wrong address) is a real fund-loss surface.

Closed at zero WASM cost: derives the pubkey from the fixture's raw private key via `@noble/curves/secp256k1` (already a dependency of the `tx.ts` module under test - it's how that module derives vault child keys), then feeds it into RN-JS's own `deriveCosmosAddress` to get the bech32 address - both asserted against the fixture's precomputed values. Chained with the encoder-binding test: `WalletCore-derive (packages/core) == precomputed (fixture) == RN-JS-derive (this test)` - both derivation AND encoding are now cross-bound.

Sanity-checked the assertion has teeth (temporarily corrupted the derivation, confirmed the test catches it, reverted).

No Solana equivalent: `packages/sdk`'s `solana/tx.ts` has no derivation function to bind against - Solana addresses are raw base58-encoded ed25519 pubkeys with no HD/bech32 transformation, and `buildSolanaSendTx` takes an already-known address string rather than deriving one. Noting this rather than fabricating a check for a step the SDK doesn't actually own.

codex-review: clean on the third pass.


NeOMakinG (approved):
Reviewed the cross-encoder golden-vector binding with a runtime mutation lane (worktree @ 108b434f) + a parallel Codex `gpt-5.4` static lane. This is the anti-tautology crux of the whole golden-vector effort, so I verified independence dynamically rather than by reading comments.

## testing approach
- Worktree `yarn install` + `yarn build:shared`, then ran both layers:
  - **Layer-1** `packages/core/mpc/tx/compile/compileTx.golden.test.ts` (WalletCore/WASM): **9/9 green**
  - **Layer-2** `cosmos-send-golden-vectors.test.ts` + `tx-builder-golden-vectors.test.ts` (SDK RN-JS): **11/11 green**
- **Independence / tautology proof (the crux) — PASSED:**
  1. *Tamper the fixture value* (flip one hex char in `expectedSignDocSha256Hex`) → **BOTH** Layer-1 (WalletCore) and Layer-2 (RN-JS) fail. Proves neither assertion is vacuous — each side genuinely computes its own bytes and compares.
  2. *Neuter the SDK's OWN RN-JS encoder* (append a byte to `signDocBytes`/`message` pre-hash, fixture untouched) → **Layer-2 FAILS** while **Layer-1 WalletCore still PASSES on the unchanged fixture**. This is the decisive result: the fixture holds the *independent WalletCore* value, NOT the SDK's current output, so a real SDK mis-encode is caught. Reproduced for **both** Cosmos and Solana. Reverted after.
- The fixture is passive JSON (`testdata/cross-encoder-golden/*.json`) — there is no shared *encoder* both sides wrap; the anchor is `WalletCore (3rd-party WASM) == fixture == SDK RN-JS`. Genuinely independent.

## blocking
None.

## preferably-blocking
None.

## should-fix
None.

## suggestion
- ℹ️ **Solana derivation is intentionally uncrossed** — `buildSolanaSendTx` takes an already-known address and owns no privkey→address step, so there's no RN-side derivation-binding test (unlike Cosmos, which binds `@noble/curves` + `deriveCosmosAddress` against the fixture at `cosmos-send-golden-vectors.test.ts:252`). Accepted — the SDK doesn't own that surface; the PR body documents it. No action needed.
- suggestion: the Cosmos fixture's `recipientCosmosAddress` is fed into `buildCosmosSendTx` but never re-derived/asserted on the RN side (only `sender` is, at `:252`). It's transitively bound because Layer-1 re-derives the recipient fresh via WalletCore (`compileTx.golden.test.ts:413`) and the mismatch would move the SignDoc hash — so it's covered, just not self-contained on the RN side. Optional: assert `deriveCosmosAddress(recipientPrivateKey)` too for symmetry.

## q
None.

## risk
**LOW.** Test-only: 3 `.test.ts` files + 2 `testdata/*.json`, zero product code, no dep/lockfile change. All new blocks are **additive** — the pre-existing independent self-checks (cosmjs-types at `cosmos-send:129`, `@solana/web3.js` at `tx-builder:87`) are untouched; the only `-1` in the diff is an import-line replacement adding `deriveCosmosAddress`. No #1001 independence dropped. Worst case is a passive fixture, no runtime surface.

## verdict
**APPROVE.** The shared cross-encoder fixture is genuinely independent of the SDK's own encoder (WalletCore is the anchor), and mutation-sound in both directions — neutering the SDK RN-JS encoder fails the bound vectors while WalletCore holds the line. No existing independent assertion weakened; test-only. Codex `gpt-5.4` second lane independently concurred (not tautological, mutation-sound, existing refs preserved).

---
🤖 Reviewed with runtime mutation verification (Claude) + Codex gpt-5.4 static lane.

PR #1046, opened 2026-07-08
fix(broadcast): UTXO/Cardano no longer report false success on a real failure

gomesalexandre:
tl;dr a genuine broadcast failure (spent/invalid UTXO inputs) on Bitcoin/Litecoin/Dogecoin/Bitcoin-Cash/Dash/Zcash/Cardano could be silently reported as success with a fabricated tx hash. Fund-safety-critical broadcast path — do NOT merge, this needs review.

## root cause (2 layers, both fixed)

**Layer 1 — the broadcast resolvers bucketed a real failure with benign duplicates.** `broadcastCardanoTx`/`broadcastUtxoTx` matched `BadInputsUTxO` (spent/invalid inputs — a genuine failure) together with `txn-mempool-conflict`/`already known` (benign MPC-race duplicates — another device already broadcast the identical signed tx) and returned success (`null`, later fabricated into a local hash downstream) unconditionally for all three, bypassing `verifyBroadcastByHash` — the safety net that re-checks the real chain before treating an ambiguous error as a race.

Fix: removed the early-return bucket entirely. Every ambiguous submit error now routes through `verifyBroadcastByHash`, which either confirms the tx is genuinely on-chain (the race case — still succeeds) or rethrows the real error.

**Layer 2 — `verifyBroadcastByHash` itself couldn't tell "genuinely missing" from "confirmed" for UTXO/Cardano.** Layer 1 alone wasn't sufficient: `getUtxoTxStatus` (shared by Cardano's status dispatch) never set `isKnown`, so `verifyBroadcastByHash`'s `isKnown !== false` guard treated a hash Blockchair has *never seen* as a confirmed-pending tx — silently re-swallowing the exact failure Layer 1 was supposed to catch.

Fix: `getUtxoTxStatus` now sets `isKnown: false` on a genuinely-missing hash or network error, and `isKnown: true` when Blockchair has indexed it in the mempool (`block_id: null` or `-1`) — the same convention already used by `cosmos.ts`/`evm.ts`/`polkadot.ts`/`ripple.ts`/`solana.ts`'s status resolvers (this one was simply missed when the convention was introduced elsewhere).

## test plan

- [x] Fail-before/pass-after proven via `git stash` isolation, separately for the broadcast-resolver change and the status-resolver change
- [x] New tests: a genuine `BadInputsUTxO` with the tx NOT on chain now rejects; an `already known` with the tx confirmed IS on chain still succeeds; `getUtxoTxStatus` returns `isKnown:false`/`isKnown:true` correctly for not-found/mempool-indexed hashes
- [x] `npx vitest run --config .config/vitest.core.config.ts` (packages/core/chain, all touched suites) — 26/26 passing
- [x] `npx tsc --project .config/tsconfig.json --noEmit` clean
- [x] Codex adversarial review, 2 rounds — round 1 caught the Layer 2 gap (this PR includes both fixes), round 2 clean

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved broadcast handling for UTXO and Cardano transactions so ambiguous submit errors are verified on-chain before being treated as successful.
  * Fixed cases where genuine failures could be reported as success.
  * Clarified transaction status results by distinguishing between unknown hashes and pending-but-known transactions.

### DIFF
--- a/packages/core/chain/tx/broadcast/resolvers/cardano.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/cardano.ts
@@ -3,7 +3,6 @@ import { getCardanoTxTtl } from '@vultisig/core-chain/chains/cardano/cip30/carda
 import { getCardanoCurrentSlot } from '@vultisig/core-chain/chains/cardano/client/currentSlot'
 import { cardanoBroadcastTtlSafetyMargin } from '@vultisig/core-chain/chains/cardano/config'
 import { getCardanoTxHash } from '@vultisig/core-chain/tx/hash/resolvers/cardano'
-import { isInError } from '@vultisig/lib-utils/error/isInError'
 
 import { submitCardanoCbor } from '../../../chains/cardano/submit/submitCardanoCbor'
 import { BroadcastTxResolver } from '../resolver'
@@ -44,10 +43,10 @@ export const broadcastCardanoTx: BroadcastTxResolver<OtherChain.Cardano> = async
 
   const error = errorMessage ?? 'unknown broadcast failure'
 
-  if (isInError(error, 'BadInputsUTxO', 'txn-mempool-conflict', 'already known')) {
-    return null
-  }
-
+  // Any submit error past this point is ambiguous — it could be a benign MPC-race duplicate (another
+  // device already broadcast the same signed tx) OR a genuine failure (e.g. BadInputsUTxO: spent/invalid
+  // inputs). String-matching alone can't tell them apart, so verify against the real chain: the hash
+  // either resolves on-chain (the race case — success) or it doesn't (the real failure — rethrows).
   const broadcastError = new Error(`Failed to broadcast transaction: ${error}`)
   await verifyBroadcastByHash({ chain, tx, error: broadcastError })
   return null
--- a/.changeset/fix-utxo-cardano-broadcast-false-success.md
+++ b/.changeset/fix-utxo-cardano-broadcast-false-success.md
@@ -1,0 +1,12 @@
+---
+"@vultisig/sdk": patch
+---
+
+fix: UTXO/Cardano broadcast no longer reports false success on a genuine failure
+
+`broadcastCardanoTx`/`broadcastUtxoTx` bucketed `BadInputsUTxO` (a genuine failure — spent/invalid inputs)
+together with benign MPC-race duplicates (`txn-mempool-conflict`/`already known`) and returned success
+unconditionally for all three, bypassing the on-chain hash verification safety net. Every ambiguous submit
+error now routes through `verifyBroadcastByHash`, and `getUtxoTxStatus` now sets `isKnown: false` when the
+hash is genuinely not found (matching the existing convention already used by the cosmos/evm/polkadot/
+ripple/solana resolvers) so a real failure correctly rethrows instead of being swallowed as success.
--- a/packages/core/chain/tx/broadcast/resolvers/cardano.test.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/cardano.test.ts
@@ -100,13 +100,40 @@ describe('broadcastCardanoTx', () => {
     await expect(broadcastCardanoTx({ chain, tx })).rejects.toBe(verifyError)
   })
 
-  it('continues swallowing duplicate Cardano broadcast errors without hash verification', async () => {
+  it('routes an "already known" duplicate through hash verification instead of blanket-swallowing it (false-success fix)', async () => {
+    // Before the fix, 'txn-mempool-conflict'/'already known'/'BadInputsUTxO' were bucketed together and
+    // returned null unconditionally — but BadInputsUTxO is a GENUINE failure (spent/invalid inputs), not
+    // an MPC-race duplicate. Verification is now mandatory for every ambiguous submit error.
     const tx = txWithTtl(1_000)
     mocks.getCardanoCurrentSlot.mockResolvedValue(939n)
     mocks.submitCardanoCbor.mockResolvedValue({ errorMessage: 'txn-mempool-conflict' })
+    mocks.verifyBroadcastByHash.mockResolvedValue(undefined)
 
     await expect(broadcastCardanoTx({ chain, tx })).resolves.toBeNull()
 
-    expect(mocks.verifyBroadcastByHash).not.toHaveBeenCalled()
+    expect(mocks.verifyBroadcastByHash).toHaveBeenCalledTimes(1)
+  })
+
+  it('rejects a BadInputsUTxO error when the tx is NOT actually on chain (the false-success bug this fixes)', async () => {
+    const tx = txWithTtl(1_000)
+    const verifyError = new Error('Failed to broadcast transaction: BadInputsUTxO')
+    mocks.getCardanoCurrentSlot.mockResolvedValue(939n)
+    mocks.submitCardanoCbor.mockResolvedValue({ errorMessage: 'BadInputsUTxO' })
+    // verifyBroadcastByHash rethrows the original error when the tx isn't confirmed on-chain — see its
+    // own contract in verifyBroadcastByHash.ts.
+    mocks.verifyBroadcastByHash.mockRejectedValue(verifyError)
+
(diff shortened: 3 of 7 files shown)


packages/core/chain/tx/broadcast/resolvers/cardano.ts
```diff
@@ -44,10 +43,10 @@ export const broadcastCardanoTx: BroadcastTxResolver<OtherChain.Cardano> = async
...
-  if (isInError(error, 'BadInputsUTxO', 'txn-mempool-conflict', 'already known')) {
-    return null
-  }
-
+  // Any submit error past this point is ambiguous — it could be a benign MPC-race duplicate (another
+  // device already broadcast the same signed tx) OR a genuine failure (e.g. BadInputsUTxO: spent/invalid
+  // inputs). String-matching alone can't tell them apart, so verify against the real chain: the hash
+  // either resolves on-chain (the race case — success) or it doesn't (the real failure — rethrows).
   const broadcastError = new Error(`Failed to broadcast transaction: ${error}`)
   await verifyBroadcastByHash({ chain, tx, error: broadcastError })
```
NeOMakinG, line 51: should-fix (non-blocking, safe direction): a genuine Cardano MPC-race duplicate that surfaces as `errorMessage` `txn-mempool-conflict`/`already known` (i.e. NOT `rpcErrorCode === 3117`) now routes here into `verifyBroadcastByHash`, which resolves Cardano status via `getUtxoTxStatus` -> `getBlockchairBaseUrl(cardano)`. If Blockchair does not expose Cardano *mempool* state the way it does for UTXO (`block_id -1`), that genuine race returns `isKnown:false` and rethrows — the user sees FAILURE for an in-flight tx. It is the safe direction (idempotent rebroadcast, no double-pay), but a UX regression vs the old unconditional swallow. Please confirm Blockchair indexes Cardano mempool, or add an Ogmios-based mempool check for this branch. The new Cardano tests mock `verifyBroadcastByHash`, so they won't catch this — an integration test against the real status route would.

NeOMakinG (approved):
Reviewed the UTXO + Cardano broadcast false-success fix under an adversarial fund-safety lens (two lanes: Claude load-bearing static trace + 19/19 local vitest on the touched suites, and Codex gpt-5.4 second-opinion). The dangerous direction — a genuine broadcast failure being reported as success with a fabricated hash — is correctly eliminated on both chain families, and ambiguity now fails *toward* failure. Approving; two non-blocking notes below (one Cardano false-failure edge, one 3117 question).

## testing approach
- Traced both broadcast paths to the caller: `broadcastUtxoTx` (Blockchair `/push/transaction`) and `broadcastCardanoTx` (Ogmios `submitTransaction`), plus the shared `verifyBroadcastByHash` -> `getTxStatus` -> `getUtxoTxStatus` chain (Cardano status delegates to the UTXO resolver via `getCardanoTxStatus`).
- Verified the two-layer fix: (L1) the `isInError('BadInputsUTxO','txn-mempool-conflict','already known')` early-return is gone, so every ambiguous submit error routes through `verifyBroadcastByHash`; (L2) `getUtxoTxStatus` now emits `isKnown:false` on not-found/network-error and `isKnown:true` on `block_id null/-1` (mempool), which is exactly the signal `verifyBroadcastByHash`'s `status==='pending' && isKnown !== false` swallow-guard needs to distinguish a real in-flight tx from one Blockchair has never seen.
- Ran the touched suites locally against copied workspace dists: `utxo.test.ts` + `cardano.test.ts` (broadcast) + `utxo.test.ts` (status) = **19/19 passing**. CI: Integration (Vitest) + Lint/TypeCheck + CodeRabbit green; Unit Tests job still in-progress at review time (same workflow as the green Integration job; author reports 26/26).
- Confirmed the `isInError` import removal is clean (Lint/TypeCheck green) and that the only consumer of the new `isKnown` field is `verifyBroadcastByHash` — the public `getTxStatus` polling path (`VaultBase.getTxStatus`, approval-poll) still branches on `status` only, which stays `'pending'` in both new branches, so no polling-UI regression.

## fund-safety verification
1. **Real failure -> failure ✓** — `BadInputsUTxO` with the tx not on chain: L1 routes to verify, L2 returns `isKnown:false`, verify rethrows the original error. Test-locked on both chains.
2. **No fabricated success txid ✓** — genuine success returns the *node-provided* hash (Blockchair `transaction_hash` / Ogmios `result.transaction.id`), never a locally-derived one. The swallowed-race path returns `null` (downstream computes a local hash) but only *after* verify has confirmed the tx is on-chain — so the local hash always corresponds to a real tx. Failure rethrows before any hash is returned.
3. **Fail-toward-failure on ambiguity ✓** — `verifyBroadcastByHash` rethrows on any non-confirmed status, and its internal `try/catch` falls through to rethrow on *any* verification error (status RPC down, hash mismatch, not indexed). Mandated safe direction honored.
4. **UTXO already-in-mempool = success ✓** — Blockchair indexes UTXO mempool (`block_id null/-1`), so a genuine MPC-race duplicate resolves `isKnown:true` -> swallowed -> success. No false-failure regression for BTC/LTC/DOGE/BCH/Dash/Zcash. (Cardano caveat in should-fix below.)
5. **Per-chain ✓** — UTXO and Cardano have distinct submit APIs/error shapes; both now funnel ambiguous errors into the same verify safety net.
6. **Changeset present ✓** — `.changeset/fix-utxo-cardano-broadcast-false-success.md` (patch bump), accurate description.

## should-fix
- **Cardano genuine mempool-race can now false-fail (safe direction, but a real behavioral regression).** The pre-fix code string-matched `'txn-mempool-conflict'`/`'already known'` in the *errorMessage* path — proving genuine Cardano race duplicates do surface there, not only via `rpcErrorCode === 3117`. Post-fix those route to `verifyBroadcastByHash` -> `getCardanoTxStatus` -> `getUtxoTxStatus` -> `getBlockchairBaseUrl(cardano)` = `.../blockchair/cardano/dashboards/transaction/:hash`. If Blockchair does not expose Cardano *mempool* state (it reliably indexes UTXO mempool via `block_id -1`, but Cardano mempool coverage is not obviously equivalent), a genuine race returns `isKnown:false` -> rethrow -> the user sees FAILURE for a tx that is actually in flight. This is the *safe* direction per the brief (re-broadcasting the same signed Cardano tx is idempotent — no double-pay) so it does not block, but it is a UX regression from the old unconditional swallow. Please confirm Blockchair indexes Cardano mempool, or add an Ogmios-based mempool check for the Cardano race case. Note the new Cardano tests mock `verifyBroadcastByHash` directly, so they cannot catch this — an integration test exercising the real status route would. (Codex second-opinion concurred, High.)

## q
- **Cardano `rpcErrorCode === 3117` still early-returns a locally-computed hash with no on-chain verification** (`getCardanoTxHash(tx)`, unchanged by this PR). Unlike the ambiguous string-match errors, 3117 is an *affirmative* Ogmios "already known / in ledger" signal, so trusting it looks legitimate and pre-existing — just confirming that's the intent and that 3117 can never be returned for a tx the node rejected. (Codex flagged this as the one remaining unverified success branch, Medium.)

## risk
Low. The change strictly tightens a fail-open path into fail-closed on the dangerous (false-success) direction; the only new failure mode is a false-*failure* on the narrow Cardano errorMessage-race case, which is recoverable and cannot cause fund loss (idempotent rebroadcast). No native/JSI/transitive surface touched; pure resolver logic. Public polling API unaffected.

## verdict
**APPROVE.** False-success is eliminated across UTXO + Cardano, ambiguity fails toward failure, UTXO already-in-mempool correctly succeeds, changeset present, CI green (unit job pending but same-workflow integration+lint green + 19/19 local). The Cardano should-fix and the 3117 q are non-blocking follow-ups.

---
_Two-lane review: Claude (static trace + local vitest) + Codex gpt-5.4 (adversarial second-opinion, verified on-PR #1046, converged on the Cardano false-failure edge)._

gomesalexandre (commented):
CI green ser — all jobs pass. Ready for review when you get a chance 🥹

Fund-safety patch: genuine UTXO/Cardano broadcast failures (BadInputsUTxO) were silently surfacing as success with a fabricated hash. Two-layer fix: routing real failures through verifyBroadcastByHash + wiring isKnown correctly in getUtxoTxStatus.

PR #1073, opened 2026-07-08
fix(swap): guard limit-swap LIM against flooring to zero (market-swap reinterpret) — THOR-02

### LINKED ISSUE #1048: [audit-THOR-02] Limit-swap LIM can silently floor to zero -> THORChain reinterprets it as an unprotected market swap
## Context (fund-safety audit round 2, 2026-07-08 · finding THOR-02 · High)

limitSwapMemo.ts:143 getLimitSwapLimitAmount is (sourceAmount * targetPrice) / priceScale with plain bigint division and no >0 check. A valid tiny target price floors to 0, and per THORChain docs a zero trade target is treated as a MARKET swap with no minimum-output floor. The regular-swap path bumps a floored-to-zero limit to '1' (nativeSwapQuoteToSwapPayload.ts:55) — the limit-swap path has no equivalent guard.

## Impact
A user who thinks they placed a price-protected limit order gets an immediate market execution with zero slippage protection — full MEV/bad-price exposure, the opposite of the intent.

## Anchors
`limitSwapMemo.ts:143` · `nativeSwapQuoteToSwapPayload.ts:55` (the regular-path guard to mirror).

## Fix
Reject a floored-to-zero LIM (or bump to '1' like the regular path, whichever preserves limit-order semantics) so a limit swap never degrades to a zero-target market swap.

SDK-CLI: build the SDK and unit-drive getLimitSwapLimitAmount with a tiny target price that floors to 0; BEFORE returns 0, AFTER rejects/bumps. Assert the assembled memo never carries a zero trade target on the limit path.

## Prompt for AI agents
In vultisig-sdk limitSwapMemo.ts:143, guard getLimitSwapLimitAmount against a floored-to-zero result: a zero LIM on a limit swap is reinterpreted by THORChain as a market swap (no min-out). Mirror the regular-swap guard (nativeSwapQuoteToSwapPayload.ts:55 bumps floored-0 to '1') OR reject the input as too-small-for-a-limit-order — pick the one that preserves the user's price-protection intent (rejecting is safer). Add unit tests for the floor boundary. Runtime-verify via SDK-CLI.

---
_Audit finding **THOR-02** (round 2, High) · bead `agent-backend-ts-25blg` · runtime before+after is the finish line, not unit-tests-green. When you open a PR, put `closes #<this issue>` in the body._

gomesalexandre:
closes vultisig/vultisig-sdk#1048

## Finding (fund-safety, HIGH — audit THOR-02)

`limitSwapMemo.ts:143` — `getLimitSwapLimitAmount` computed `(sourceAmount * targetPrice) / priceScale` with plain bigint division and **no `> 0` check**. A valid tiny target price (e.g. `source_amount=1_000_000, target_price=0.00000001`) floors to `0`, and per THORChain docs a zero trade target is treated as a **MARKET swap with no minimum-output floor**. The regular-swap path already bumps a floored-to-zero limit to `'1'` (`nativeSwapQuoteToSwapPayload.ts:55`) — the limit-swap path had no equivalent guard.

Impact: a user who thinks they placed a price-protected limit order gets an immediate market execution with **zero slippage protection** (full MEV/bad-price exposure) — the opposite of intent.

## Decision: reject, not bump

Went with **reject** over bumping to `'1'` (mirroring the regular-swap path). Bumping to 1 base unit still gives ~zero real protection and misrepresents the user's stated limit — a LIM that floors to 0 at this price scale can't be honestly expressed as *this user's* limit order. Failing closed with a clear error is safer than silently degrading a limit order into a de-facto market order under a false "protected" label. No open PR duplicates this (checked `gh pr list` for limitSwap/LIM/getLimitSwapLimitAmount — clean), and there's only one call site (`buildMemo` inside the same file), so the blast radius of the stricter behavior is contained to `buildLimitSwapMemo`/`getLimitSwapLimitAmount` callers.

## Fix

`getLimitSwapLimitAmount` now throws when the computed LIM floors to `0n`, before a memo can ever be built with a zero trade target.

## Runtime evidence (vitest, `.config/vitest.core.config.ts`)

**BEFORE** (scratch repro, not committed):
```
getLimitSwapLimitAmount({ source_amount: 1_000_000, target_price: 0.00000001 }) === 0n   // TRUE
buildLimitSwapMemo(...) memo contains ":0/"                                              // TRUE (zero trade target)
```

**AFTER** (committed regression test, `limitSwapMemo.test.ts`):
```
getLimitSwapLimitAmount({ source_amount: 1_000_000, target_price: 0.00000001 })
  → throws /floors to 0/

buildLimitSwapMemo({ ...validInput, source_amount: 1_000_000, target_price: 0.00000001 })
  → throws /floors to 0/

// boundary unchanged — smallest valid non-zero LIM still works:
getLimitSwapLimitAmount({ source_amount: 100_000_000, target_price: 0.00000001 }) === 1n
```

Test command: `yarn vitest run --config .config/vitest.core.config.ts limitSwapMemo` → **33 passed (33)**.
Full core suite (`yarn vitest run --config .config/vitest.core.config.ts`, after `yarn build:shared`): **1434 passed | 1 skipped**.
`yarn tsc --project .config/tsconfig.json --noEmit` → clean.
Lint (`node scripts/run-eslint.mjs --config .config/eslint.config.mjs` on the two touched files) → clean.

## Self-review note

Codex CLI auth was degraded in this environment, so this went through manual self-review only (no `/codex-review` pass). Diff is small and contained to `limitSwapMemo.ts` + its test file — flagging for a human second pair of eyes given the fund-safety nature of the finding.

## Risk

Low — adds a fail-closed guard on a path with a single internal call site. No existing fixture in `limit-swap-memos.json` hits the zero-floor case, so no behavior change for any currently-passing scenario; only newly-rejected inputs are ones that previously silently degraded to an unprotected market order.

### DIFF
--- a/.changeset/reject-limit-swap-zero-floor.md
+++ b/.changeset/reject-limit-swap-zero-floor.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(swap): reject a limit-swap LIM that floors to zero (THOR-02). A tiny target price could floor the computed limit amount to 0, and THORChain treats a zero trade target as an unprotected MARKET swap with no minimum-output floor — silently discarding the price protection the user configured for their limit order. Fail closed with a clear error instead of building a memo that reinterprets as a market swap.
--- a/packages/core/chain/swap/native/limitSwapMemo.test.ts
+++ b/packages/core/chain/swap/native/limitSwapMemo.test.ts
@@ -169,6 +169,27 @@ describe('validateLimitSwapInputs', () => {
     ).toBe(1n)
   })
 
+  it('rejects source_amount/target_price combinations that floor LIM to 0 (THOR-02)', () => {
+    // 1_000_000 * 1 (scaled target_price) / 1e8 = 0.01, floors to 0.
+    // A zero trade target is treated by THORChain as an unprotected market
+    // order, so this must fail closed rather than silently reinterpreting
+    // the user's limit swap as a market swap.
+    expect(() =>
+      getLimitSwapLimitAmount({
+        source_amount: 1_000_000,
+        target_price: 0.00000001,
+      })
+    ).toThrow(/floors to 0/)
+
+    expect(() =>
+      buildLimitSwapMemo({
+        ...validInput,
+        source_amount: 1_000_000,
+        target_price: 0.00000001,
+      })
+    ).toThrow(/floors to 0/)
+  })
+
   it('rejects unsupported expiries', () => {
     expect(() =>
       validateLimitSwapInputs({
--- a/packages/core/chain/swap/native/limitSwapMemo.ts
+++ b/packages/core/chain/swap/native/limitSwapMemo.ts
@@ -140,7 +140,22 @@ export const getLimitSwapLimitAmount = ({
   const sourceAmount = parsePositiveInteger(source_amount, 'source_amount')
   const targetPrice = parsePositiveDecimal(target_price, 'target_price')
 
-  return (sourceAmount * targetPrice) / priceScale
+  const limit = (sourceAmount * targetPrice) / priceScale
+
+  if (limit === 0n) {
+    // THORChain treats a zero trade target (LIM) in a limit-swap memo as an
+    // unprotected market order. source_amount/target_price combinations
+    // that floor to 0 here can't be honestly expressed as a limit order at
+    // this price scale, so we fail closed instead of silently reinterpreting
+    // the user's price-protected limit swap as a market swap.
+    throw new Error(
+      'source_amount and target_price combination is too small to express as a limit swap: ' +
+        'the computed minimum-received amount (LIM) floors to 0, which THORChain treats as an ' +
+        'unprotected market order. Increase source_amount or target_price.'
+    )
+  }
+
+  return limit
 }
 
 export const assertMemoByteLength = (memo: string, sourceChainKind: LimitSwapSourceChainKind): void => {


gomesalexandre (commented):
## Review — THOR-02 limit-swap LIM zero-floor guard

Draft PR, comment-only verdict per severity system. Walked `getLimitSwapLimitAmount` + `buildLimitSwapMemo` + the sibling `nativeSwapQuoteToSwapPayload.ts` bump path, ran the fixture + regression suite locally, did the math on the false-refusal question myself. No Codex lane (auth degraded per PR body, confirmed).

### 1. Correctness of the reject trigger

Confirmed: the throw in `getLimitSwapLimitAmount` (limitSwapMemo.ts:143-156) fires exactly on `limit === 0n`, computed *after* `parsePositiveInteger`/`parsePositiveDecimal` already reject non-positive `source_amount`/`target_price`. So the only path to `limit === 0n` is genuine integer-division underflow — never a false-positive on a valid positive LIM (the branch is a strict guard after the division, doesn't touch the happy path at all).

Checked the memo shape too: `=<:{target_asset}:{dest_addr}:{limit}/{interval}/0`. The trailing `0` is the streaming-quantity field, unrelated to trade target — `{limit}` is the *only* place a zero trade-target could leak into the memo, and `buildMemo` calls `getLimitSwapLimitAmount` before string-templating, so a floored LIM can never reach a built memo. `praise:` good — the guard sits at the single choke point, not scattered across call sites.

### 2. REJECT vs BUMP — I agree with REJECT, and it's not a close call

Did the math on whether this can false-refuse a legit user. Let `wholeSourceUnits = source_amount / 1e8` and `target_price` be the ratio as entered. `limit` floors to 0 iff `wholeSourceUnits * target_price < 1e-8`. For any realistic price ratio (say 1e-4, already a very cheap target asset relative to source), that only fires when `wholeSourceUnits < 1e-4` — i.e. the user is limit-ordering a **dust-level fraction of a coin** (<0.0001 whole units), which THORChain's own min-swap-amount floor would almost certainly reject upstream anyway. The PR's repro (`target_price=0.00000001`) is the smallest representable 8-decimal value — as a "minimum acceptable target per source unit" that's indistinguishable from "I don't care what I get," which is exactly the market-order semantics THORChain applies to a literal-zero LIM. So no, I don't think this false-refuses any order a rational user would place; it only fires on inputs that are already degenerate (dust size × near-zero price floor, or fat-fingered decimal-scale mixups).

That's also why I don't buy mirroring the `nativeSwapQuoteToSwapPayload.ts:55` bump. That bump operates on `expected` — a value **derived from a live market quote**, so nudging a floored rounding remainder up to `1` is a harmless correction on a number the system already computed authoritatively. The limit-swap LIM is the opposite: it's the **user's own stated price**, the entire point of the field. Bumping it to `1` doesn't correct a rounding artifact, it discards the user's stated protection and substitutes a value that's *functionally* a market order while a `LimitOrder` label sits on top of it in the UI/history — worse than no protection because it lies about what was placed. Fail-closed with an actionable message ("increase source_amount or target_price") is the correct call here. Regular-swap bump-to-1 and limit-swap reject-at-0 are solving different problems; they shouldn't be symmetric.

### 3. Blast radius — confirmed scoped

`git diff main --stat` for `nativeSwapQuoteToSwapPayload.ts` (regular-swap path, `packages/core/mpc/swap/native/utils/`) shows zero changes. Diff touches only `limitSwapMemo.ts` (+16/-1) and its test file (+21/-0). `getLimitSwapLimitAmount`/`buildLimitSwapMemo` have no other call sites in the repo yet (grepped) — blast radius is exactly what the PR body claims.

Checked the existing fixture set (`limit-swap-memos.json`, 12 cases) — none of them hit the zero-floor branch (smallest is `usdt-to-btc` @ `1e-05` with `source_amount=100_000_000`, limit=1000), so no behavior change for any currently-passing scenario, matches the PR's risk claim.

### 4. Tests — ran them myself

```
cd ~/Sites/vultisig-sdk/.worktrees/fix_limit_swap_lim_zero_floor
yarn vitest run --config .config/vitest.core.config.ts limitSwapMemo
→ Test Files 1 passed (1) / Tests 33 passed (33)
```
Reject case (`source_amount=1_000_000, target_price=0.00000001` → throws `/floors to 0/`) and boundary case (`source_amount=100_000_000`, same price → `1n`) are both real, not just claimed. Didn't re-run the full 1434-suite (no reason to doubt the reported number given the change is fully contained to one file + its test, but flagging I didn't personally re-verify that number).

### Nits (non-blocking)

- `suggestion:` the error message is solid but slightly redundant ("too small to express as a limit swap" + "floors to 0" + "increase source_amount or target_price" all in one string) — could trim, not worth a round-trip.
- `q:` is there an existing pre-flight/UI-facing validation call (`validateLimitSwapInputs`) that a caller might rely on to catch bad inputs *before* attempting to build the memo? This fix only throws inside `getLimitSwapLimitAmount`, which `validateLimitSwapInputs` does NOT call — so a caller who validates first and builds second still gets the exception on `buildLimitSwapMemo` (fine, fail-closed still holds), but a caller who *only* calls `validateLimitSwapInputs` for a pre-submit check (if one exists downstream) won't catch this case early. Not a fund-safety issue since the memo is still never built with a zero LIM, just a UX layering question — worth a one-line callout in the PR body or a follow-up if a consumer repo does pre-validate-then-build-separately.

### Verdict

**SHIP-WITH-NITS.** No blockers, no preferably-blocking. Reject-not-bump is the right call and I did the math independently to confirm it doesn't false-refuse realistic limit orders — good fund-safety fix, cleanly scoped, tests are real. The one `q:` above is worth a one-line reply, not a re-review.

NeOMakinG (approved):
**APPROVE** — THOR-02 limit-swap zero-floor guard is a correct, fail-closed fund-safety fix.

## testing approach
Reviewed at HEAD `f7628c0d` in an isolated worktree. Traced the LIM computation → guard → signable memo, plus a Codex `gpt-5.4` adversarial second lane (confirmed reviewing #1073, no findings). Footprint vs `origin/main` is exactly the 3 diff files.

## verification (all pass)
- **Non-zero-intended LIM can no longer silently market-swap.** `getLimitSwapLimitAmount` now throws when `(sourceAmount * targetPrice) / priceScale` floors to `0n`. This is the **reject** path (fail-closed), the safer choice over the regular-swap path's clamp-to-`1` (which would misrepresent a de-facto market order under a "protected" label). A tiny `target_price` that floors LIM to 0 → hard error, never a memo.
- **The throw reaches the signable memo, not just display.** `buildMemo` (L233) calls `getLimitSwapLimitAmount` *before* interpolating `limit` into the memo string `=<:${target}:${dest}:${limit}/${interval}/0` (L235). No memo with `LIM=0` can be constructed. The UTXO fallback `try/catch` (L246–254) does **not** swallow the throw: `buildMemo(inputs, true)` runs at L244 *outside* the try, and the `catch` only wraps `assertMemoByteLength` — `buildMemo(inputs, false)` inside it re-throws the same error. Valid limits (`>= 1n`) interpolate directly with no further scaling/truncation.
- **Genuine market swaps unaffected.** Market/regular swaps go through `getNativeSwapQuote` / `nativeSwapQuoteToSwapPayload`, never `buildLimitSwapMemo`. The guard fires only on the limit-order path, and only at exactly `limit === 0n` — no false positives on legitimate small non-zero limits.
- **Scaling correct.** `priceScale = 1e8`; `target_price` scaled by 1e8 in `parsePositiveDecimal` (rejects >8 fractional digits), `source_amount` in base units → LIM in THOR 1e8 units. Arithmetic unchanged by this PR; only the `>0` guard added.
- **Ships live, not inert.** `.changeset/reject-limit-swap-zero-floor.md` (`@vultisig/sdk: patch`) present → Lint/Type-Check (changeset guard) green.
- **Test coverage.** Regression test rejects `source_amount=1_000_000, target_price=0.00000001` at both `getLimitSwapLimitAmount` and `buildLimitSwapMemo`; boundary `source_amount=100_000_000, target_price=0.00000001 === 1n` still builds. Codex + Unit + Integration (Vitest) all green.

## risk
Low. Adds a fail-closed guard on a single internal call site; only newly-rejected inputs are ones that previously silently degraded to an unprotected market order. No behavior change for any currently-passing fixture.

## verdict
APPROVE.

---
🤖 dual-lane review (Claude load-bearing + Codex gpt-5.4 adversarial). Fund-safety fix verified fail-closed end-to-end to the signable memo.

PR #1075, opened 2026-07-08
fix(swap): enforce Jupiter price-impact ceiling on the SDK swap path (SOL-02)

### LINKED ISSUE #1049: [audit-SOL-02] The SDK's live Jupiter swap path dropped the price-impact ceiling that guards against thin-pool/sandbox quotes
## Context (fund-safety audit round 2, 2026-07-08 · finding SOL-02 · High)

Both SDK Jupiter paths receive priceImpactPct (getJupiterSwapQuote.ts:38, tools/swap/jupiter.ts:104) but there is NO price-impact check anywhere in the SDK swap code (grep-confirmed). The prod mcp-ts path enforces a 10% ceiling; the SDK port ('Ported from mcp-ts') silently omitted it.

## Impact
A thin-pool or sandwich-bait quote at 50-90% price impact builds a fully signable, MPC-ready tx with zero protection — precisely the fund-loss scenario the ceiling exists to close, for any consumer of the SDK Jupiter flow.

## Anchors
`getJupiterSwapQuote.ts:38` · `tools/swap/jupiter.ts:104` (both receive priceImpactPct but never check it).

## Fix
Port the mcp-ts 10% price-impact ceiling into the SDK Jupiter path(s): reject/refuse to build when priceImpactPct exceeds the ceiling.

SDK-CLI: unit-drive the Jupiter quote->payload path with a quote carrying priceImpactPct > 10%; BEFORE builds a signable tx, AFTER refuses. Cross-check the ceiling value matches mcp-ts.

## Prompt for AI agents
In vultisig-sdk, port the mcp-ts Jupiter price-impact ceiling (10%) into both SDK Jupiter paths (getJupiterSwapQuote.ts and tools/swap/jupiter.ts) — they already receive priceImpactPct but never check it. Refuse to build a swap payload when priceImpactPct exceeds the ceiling; make the ceiling a shared named constant. Add unit tests at the boundary. Runtime-verify via SDK-CLI that a high-impact quote is refused.

---
_Audit finding **SOL-02** (round 2, High) · bead `agent-backend-ts-d8ano` · runtime before+after is the finish line, not unit-tests-green. When you open a PR, put `closes #<this issue>` in the body._

gomesalexandre:
closes vultisig/vultisig-sdk#1049

Audit finding **SOL-02** (fund-safety, HIGH): `packages/core/chain/swap/general/jupiter/api/getJupiterSwapQuote.ts` received `priceImpactPct` on the Jupiter quote but never checked it against a ceiling - no price-impact protection existed anywhere in the SDK's Jupiter swap code (grep-confirmed). Same gap in the second Jupiter path, `packages/sdk/src/tools/swap/jupiter.ts:104` (`buildJupiterSwapTx`). A thin-pool / sandwich-bait Jupiter quote at 50-90% price impact built a fully signable, MPC-ready transaction with zero protection - the provider's own min-out was the only guard.

The prod agent path (mcp-ts `src/tools/swap/priceImpactGuard.ts`, bead vultisig-ops-1m48) already enforces a **10% ceiling** on this exact field. The SDK port silently omitted it - this PR ports the identical guard so the two paths don't drift.

## Fix

Added `packages/core/chain/swap/general/priceImpactGuard.ts` - a shared, fail-safe price-impact ceiling (10%, matching mcp-ts) - and wired it into both Jupiter swap-build paths:

- `getJupiterSwapQuote.ts` (core) - throws `PriceImpactTooHighError` before the `/swap` build call when the quoted impact exceeds the ceiling
- `tools/swap/jupiter.ts` (SDK `buildJupiterSwapTx`) - same guard, same error type, re-exported for callers

**Unit convention**: Jupiter's `priceImpactPct` is a **FRACTION**, not a raw percent - `"0.5"` means 50% impact, `"0.05"` means 5%. Confirmed against the existing SDK doc comment (`jupiter.ts` already documented `priceImpactPct` as `"e.g. '0.0012'"`) and the mcp-ts guard it's ported from (`evaluateImpactFromFractionString`, which multiplies by 100 before comparing). The ceiling check parses the string as a fraction and normalizes to percent before comparing against 10.

Fail-safe semantics carried over from mcp-ts: a missing/unparsable/negative impact **passes** (no usable signal, don't over-block normal swaps that omit the field) - rejection only fires on a finite impact strictly above 10%.

## Runtime verification

**BEFORE**: no ceiling existed anywhere in the SDK's Jupiter code (grep-confirmed pre-fix); a 50% or 90% impact quote built a signable tx identically to a 0.3% quote.

**AFTER** (ran directly against the built guard module):
```
MAX_PRICE_IMPACT_PCT = 10
BLOCK 50% impact (sandwich-bait) -> This swap has 50.00% price impact, above the 10% ceiling...
BLOCK 10.01% impact (just above ceiling) -> This swap has 10.01% price impact, above the 10% ceiling...
PASS  10% impact (ceiling itself) -> builds
PASS  0.3% impact (normal swap) -> builds
PASS  missing field -> builds
```

Regression tests added at both boundaries (just under vs just over 10%) in both `getJupiterSwapQuote.test.ts` and `jupiterSwap.test.ts`, plus a dedicated `priceImpactGuard.test.ts` for the shared module (unit-convention parsing, fail-safe semantics).

## Tests

- `yarn test:core` - 1491 passed (was 1149 pre-fix-baseline before `build:shared` regenerated stale export maps; unrelated to this change)
- `yarn workspace @vultisig/sdk test` - 2407 passed
- `yarn typecheck` - clean
- lint - clean

## Changeset

`.changeset/jupiter-price-impact-ceiling.md` (`@vultisig/sdk`: patch)

## Note on Codex review

Codex CLI auth was revoked in this environment (`refresh_token_invalidated`) - could not run `/codex-review`. Did a manual self-review pass instead (unit-convention cross-check against mcp-ts source, boundary semantics, prettier/eslint pass).

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a 10% price-impact check for Jupiter swaps, preventing swaps from being built when quoted impact is too high.
  * Applied the same safeguard across both SDK and core swap flows for consistent behavior.
  * Improved handling of missing or unparsable impact data by allowing the swap to proceed when no reliable signal is available.

* **Tests**
  * Added coverage for price-impact limits, including below, at, and above the threshold.


gomesalexandre (commented):
## Review — fund-safety pass (Codex was degraded, manual-only)

verdict: **SHIP** (SHIP-WITH-NITS if you count the residual note below as a nit)

### 1. Unit convention — the crux, VERIFIED correct

`evaluateImpactFromFractionString` parses the raw string and does `parsed * 100` before comparing to `MAX_PRICE_IMPACT_PCT = 10`, so `"0.5"` → 50 → rejected, `"0.05"` → 5 → passes. This is a byte-for-byte port of mcp-ts's `evaluateImpactFromFractionString` (`~/Sites/mcp-ts/src/tools/swap/priceImpactGuard.ts:78-81`) — same multiply-by-100 convention, same ceiling constant. Diffed the two files side by side, logic is identical.

Boundary math confirmed by reading + running:
- `"0.10"` (exactly 10%) → `evaluatePriceImpactPercent`: `10 > 10` is false → **passes**. Correct — ceiling itself is inclusive.
- `"0.1001"` (10.01%) → `10.01 > 10` → **throws**. Correct.
- `"0.5"` / `"0.9"` (50%/90% sandwich-bait range) → throws `PriceImpactTooHighError`. Correct.
- `"0.0999"` → passes. Correct.

Ran the boundary tests for real, not just read them:
```
yarn vitest run priceImpactGuard.test.ts getJupiterSwapQuote.test.ts jupiterSwap.test.ts
Test Files  3 passed (3)
     Tests  36 passed (36)
```
praise: the ceiling-boundary tests (exactly-10%, 10.01%, missing field) are exactly the cases that matter and both Jupiter call sites have the full matrix duplicated, not just the shared unit under `priceImpactGuard.test.ts`. Good discipline.

### 2. Fail-open on missing — correct, matches source of truth

`evaluatePriceImpactPercent` treats `null`/`NaN`/negative as "no usable signal" → passes. This is identical to mcp-ts's behavior (same guard file, same comment verbatim: "Fail-SAFE semantics... a missing/non-finite/negative impact is treated as no usable signal and PASSES"). Since Jupiter's public quote API always returns `priceImpactPct` on a valid quote, missing-field here would mean a shape change or an intermediary stripping the field, not a legitimate zero-impact swap — and failing open on that (rather than blocking 100% of swaps on a parse hiccup) is the right tradeoff mcp-ts already made in prod. Agree with porting it unchanged.

q: worth double-checking Jupiter's quote API doesn't reserve `priceImpactPct: null` for genuinely-zero-impact routes (vs the field being entirely absent) — if so a `null` there is legit "0% impact" data, not "missing", and fail-open still gives the right answer (passes either way) so this isn't a bug, just flagging the assumption for the record.

### 3. Both paths covered — VERIFIED, no third path

Grepped both `packages/core` and `packages/sdk` for every caller of `getJupiterSwapQuote` / `buildJupiterSwapTx`:
```
packages/core/chain/swap/quote/findSwapQuote.ts   → calls getJupiterSwapQuote (guarded)
packages/sdk/src/tools/swap/jupiter.ts            → buildJupiterSwapTx itself (guarded)
```
`findSwapQuote.ts` (the swap-provider orchestrator) delegates to `getJupiterSwapQuote` and has no independent Jupiter build path of its own. No third unguarded route exists in this repo.

### 4. Ceiling value + no weakening

`MAX_PRICE_IMPACT_PCT = 10` matches mcp-ts exactly. No existing check was loosened — this is purely additive (new throw path, nothing removed).

### 5. Tests + changeset

36/36 pass locally (ran it myself, not trusting CI). Changeset present: `.changeset/jupiter-price-impact-ceiling.md`, `'@vultisig/sdk': patch`, good changelog copy.

### one residual worth a q — not blocking this PR

PR body + code comments claim this "matches the guard already enforced on the production agent path," citing mcp-ts `src/tools/swap/priceImpactGuard.ts`. That's accurate for what mcp-ts *used* to do, but per current repo state mcp-ts is dead — the live production agent path is `agent-backend-ts`, which has its **own independent** `buildJupiterSwapTx` (`agent-backend-ts/src/mastra/tools/mcp/tools/swap/jupiterSwap.ts`) that does NOT call this SDK and does NOT enforce any price-impact ceiling today (grepped, confirmed — `priceImpactPct` is declared on the response type but never compared to anything). So as of this PR, SOL-02 is closed for SDK consumers but still technically open on the actual live prod path until abts either (a) adopts this SDK's `buildJupiterSwapTx` or (b) gets its own port of this guard.

q: is there a follow-up bead tracking porting this into agent-backend-ts, or does abts already plan to consume `@vultisig/sdk`'s `buildJupiterSwapTx` directly? Worth a one-line callout so this doesn't read as "prod is covered" when the SDK and the actually-deployed agent code have diverged.

Not a blocker for this PR specifically — it's scoped to the SDK and does what it says on the box for SDK consumers. Just don't let the PR description's "matches the guard already enforced on the production agent path" line get read as "prod is safe now," because the file it's citing is dead code.

---

tl;dr: fraction-convention math is right, boundary is right (10% inclusive-pass), fail-open-on-missing is a correct port of mcp-ts's own tradeoff, both current Jupiter build paths are guarded, tests pass for real (ran them), changeset present. SHIP. One q about the stale "prod already enforces this" framing given mcp-ts is dead and abts (the actual live path) doesn't have this guard — worth a follow-up bead, not a blocker here.

NeOMakinG (approved):
Fund-safety port of the SOL-02 price-impact ceiling onto both SDK Jupiter swap-build paths. Reviewed adversarially against a worst-case thin-pool / sandwich-bait quote. **Clean — approving.**

## testing approach
Two-lane. Claude (load-bearing): full source trace of both Jupiter build paths, bypass hunt, units/boundary math, fail-safe analysis, faithfulness diff against the mcp-ts production guard, plus ran all three affected test suites in a clean worktree at HEAD `a17e5bf1`. Codex (gpt-5.4) adversarial second-opinion lane: stalled at zero bytes ~19min (same buffering stall seen on #1012) — inconclusive, non-load-bearing.

Runtime evidence (worktree, `yarn install` + `@vultisig/core-chain` built):
- `priceImpactGuard.test.ts` — 8/8 pass
- `getJupiterSwapQuote.test.ts` (core path) — 11/11 pass (4 new boundary tests)
- `jupiterSwap.test.ts` (SDK path) — 17/17 pass (4 new boundary tests)

CI @ `a17e5bf1`: Unit Tests **pass**, Integration Tests (Vitest) **pass**, Lint and Type Check **pass** (bundled-changeset guard satisfied → non-inert), CodeRabbit **pass**.

## verification (all PASS)
1. **Authoritative field** ✅ — guard reads `quoteResponse.priceImpactPct` (core `getJupiterSwapQuote.ts:122`) and `quote.priceImpactPct` (SDK `jupiter.ts:271`), i.e. the real field off Jupiter's `/swap/v1/quote` response, not a fabricated/model value.
2. **Threshold** ✅ — 10%, identical to the already-shipped mcp-ts production agent guard (verified `mcp-ts/src/tools/swap/priceImpactGuard.ts`: same `MAX_PRICE_IMPACT_PCT = 10`, same decision type, same `parsed * 100` fraction convention, same strictly-above rejection). Zero drift — the exact goal of SOL-02. Defensible: well above normal (<1%) swaps, well below the 50–90% attack range.
3. **Units** ✅ — fraction→percent normalisation (`evaluateImpactFromFractionString` multiplies by 100) compared against percent-10. `"0.10"`→10 passes (verified `0.1*100 === 10` exactly, no float epsilon over-reject), `"0.1001"`→10.01 rejects, `"0.5"`→50 rejects. No off-by-100 in either direction.
4. **Blocks before signable tx** ✅ — guard fires after `/quote`, strictly before `/swap/v1/swap` build in BOTH paths (core L122 < L138; SDK L271 < L287). Tests assert the `/swap` build call is never made on rejection. Throws `PriceImpactTooHighError`, not a display warning.
5. **No bypass** ✅ — both signable `/swap/v1/swap` sites are the only Jupiter tx-build entrypoints, and both are guarded. `findSwapQuote`'s sole Jupiter path is `getJupiterSwapQuote` (guarded); the public `buildJupiterSwapTx` is guarded. No third route. A rejected Jupiter fetcher inside `findSwapQuote` degrades to no-route / another sub-ceiling provider (fail-closed — no dangerous tx built).
6. **No over-reject** ✅ — normal <1% swaps pass; ceiling itself (exactly 10%) passes; only strictly-above rejects.

## q
- **Fail-open on missing/unparsable/negative impact** — a missing/`NaN`/negative `priceImpactPct` PASSES (`evaluatePriceImpactPercent` → `ok:true, impactPct:null`). This is the one non-fail-closed edge. Accepted as documented and not blocking because: (a) it matches the production mcp-ts guard exactly (consistency is the SOL-02 objective); (b) Jupiter's quote API always returns this field, and it is computed by Jupiter's aggregator over HTTPS — not attacker-supplied route data, so an attacker cannot "omit" it to slip a thin-pool quote past the ceiling (a thin pool yields a HIGH impact, which is caught); (c) fail-closed-on-missing would over-block normal swaps on any transient field omission. Reasonable trade-off, clearly documented in the guard's doc comment.

## risk
Low. Additive fund-safety guard, message/throw-only on the reject branch, no change to quote amounts / routing / fee logic / signing. Scoped to Jupiter (the only general provider exposing `priceImpactPct`). Changeset present, CI green, port verified faithful to the live production guard.

## verdict
**APPROVE.** The ceiling reads Jupiter's authoritative `priceImpactPct`, uses a sensible 10% threshold matching the production path, applies correct fraction→percent units, blocks before the signable `/swap` build on both paths with no bypass, and does not over-reject legitimate swaps. Fail-open-on-missing is a documented, production-consistent trade-off, not a realistic attacker-reachable hole.

<sub>🤖 reviewer lane · Claude load-bearing (full source trace + 36/36 tests green in worktree) · Codex gpt-5.4 lane stalled (inconclusive, non-load-bearing)</sub>

PR #1076, opened 2026-07-08
fix(chains): validate `chains --add` against registry before persisting

neavra:
## Problem

`vultisig chains --add fakechain` correctly reports `INVALID_CHAIN` (exit 4), but it **still persists** `"fakechain"` to the vault's chain list. Chain-name resolution falls back to the raw user string when the name doesn't match the registry (`findChainByName(input) || (input as Chain)`), so the invalid value gets written before validation happens at address-derivation time.

The result: every subsequent address-deriving command (`balance`, `addresses`, `chains`) re-derives the bogus chain, throws, and dumps a stack trace to stderr on every run until the user manually runs `chains --remove fakechain`. `balance-all` silently drops the failing chain, making it look like the chain just has no balance.

## Fix

Validate the requested chain against the supported-chain registry **before** persisting in the `chains --add` path. An unsupported chain now fails closed with `INVALID_CHAIN` and **nothing is written** to the vault — no more wedged state.

## Tests

`clients/cli/src/commands/__tests__/chainsAddValidate.test.ts`:
- an unsupported `--add` rejects with `INVALID_CHAIN`, and `addChain` / `address` are never called (chain list untouched)
- the error carries the offending chain name in its context
- an existing chain list is left unchanged when a bogus add fails
- a valid supported chain still adds normally

Full `@vultisig/cli` suite (577 tests) green; typecheck and lint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid chain names are now rejected before they can be saved, preventing unsupported chains from appearing in the app.
  * The command now stops earlier with a clearer error, avoiding follow-up failures when using chain-related commands.
  * Existing chains remain unchanged if an add attempt fails.


gomesalexandre (approved):
fail-closed `--add` registry validation is the correct root-cause fix, well-tested and cleanly scoped with the CLI changeset.

PR #1077, opened 2026-07-08
fix(cli): tolerate a corrupt activeVaultId pointer instead of bricking every command

neavra:
## Problem

A corrupt or unparseable `~/.vultisig/activeVaultId.json` bricks the CLI. The active-vault pointer is read during startup (before any command runs), so a truncated or hand-edited pointer file makes the storage read throw and the error propagates out of **every** command — including `vultisig vaults`, the exact command you'd run to see what's there and recover.

Repro:

```
echo 'NOT JSON {{{' > ~/.vultisig/activeVaultId.json
vultisig vaults
# → error: Failed to read value for key "activeVaultId"  (exit 7)
```

The vault data itself is intact and stored in separate files — only the small "which vault is active" pointer is corrupt — yet the whole CLI is locked out. A *valid* but dangling pointer (id → missing vault) already degrades gracefully; garbage JSON was the outlier.

## Fix

Read the active-vault pointer through a small tolerant helper (`loadActiveVaultSafely`) that **fails open**, mirroring the broadcast journal's fail-open reads:

- an unreadable/unparseable pointer is treated as "no active vault" (warned on stderr, never stdout, so `--json` output stays clean),
- the bad pointer is cleared best-effort so the corruption doesn't resurface,
- listing vaults does not logically need the pointer, so `vaults` now lists everything with none marked active (exit 0), and other commands fall back to the normal "no active vault" path.

Wired into both the one-shot CLI bootstrap and the interactive-shell startup.

## Tests

- Unit coverage for the helper: healthy pointer passes through (no self-heal), null pointer passes through, corrupt pointer fails open + warns + self-heals, and a self-heal failure is swallowed.
- Wired through the real SDK: a corrupt `activeVaultId` read throws today and the helper absorbs it (returns null, self-heals via `setActiveVault(null)`), while a valid pointer is left untouched.

## Notes

Scoped to the read path shared by CLI startup. Behavior for a valid active pointer is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * The CLI now continues starting when the saved active-vault pointer is corrupt or truncated.
  * Invalid pointers are automatically cleared, allowing commands such as `vultisig vaults` to run normally.
  * Interactive sessions no longer auto-select a vault after detecting a corrupt pointer and instead prompt for explicit selection.
* **Tests**
  * Added coverage for corrupt, unreadable, and recoverable active-vault pointer scenarios.


gomesalexandre (approved):
approve. reviewed current head 0689d3f5 only.

clean, scoped fix. the tolerance is correctly narrowed to the `activeVaultId` pointer read only. corrupt pointers no longer brick pointer-independent commands, but real vault-load errors still fail closed and corrupt pointers do not auto-promote an arbitrary vault.

receipts:
- `yarn install --immutable`
- `yarn build:shared`
- `yarn workspace @vultisig/sdk build:fast`
- `yarn workspace @vultisig/cli vitest run src/core/__tests__/active-vault.test.ts` -> 13/13 pass
- `node scripts/run-eslint.mjs --config .config/eslint.config.mjs clients/cli/src/core/active-vault.ts clients/cli/src/core/__tests__/active-vault.test.ts clients/cli/src/session.ts clients/cli/src/index.ts` -> clean

only nit: the pointer-read catch treats any read failure as corruption and self-heals by deleting the pointer, so a transient fs error would also clear it. low severity, not a blocker.

PR #1078, opened 2026-07-08
fix(evm): sanity-cap RPC-reported priority fee (SDK2-01)

### LINKED ISSUE #1055: [audit-SDK2-01] EVM fee estimation has no sanity ceiling on the RPC-reported priority fee (Solana already has one)
## Context (fund-safety audit round 2, 2026-07-08 · finding SDK2-01 · High)

maxPriorityFeePerGas.ts:4 returns estimateMaxPriorityFeePerGas() verbatim, and chainSpecific/resolvers/evm.ts:58 does maxFeePerGas = baseFeePerGas + maxPriorityFeePerGas with no clamp anywhere between. The Solana resolver uses a configured priorityFeeLimit ceiling — proving the team already treats this as a needed guard for one chain family but not EVM.

## Impact
A misbehaving/compromised/anomalous EVM RPC returning an inflated priority fee is trusted end-to-end into the tx the user signs — the 'fee that can drain' class, with no client-side sanity cap.

## Anchors
`maxPriorityFeePerGas.ts:4` · `chainSpecific/resolvers/evm.ts:58` · Solana priorityFeeLimit (the precedent).

## Fix
Add a per-chain EVM priority-fee sanity ceiling (mirroring Solana's priorityFeeLimit); clamp/refuse an RPC priority fee above it. Tune the cap vs real congestion to avoid false refusals.

SDK-CLI: unit-drive the EVM fee resolver with a stubbed RPC returning an absurd priority fee; BEFORE it flows into maxFeePerGas verbatim, AFTER it is clamped/refused.

## Prompt for AI agents
In vultisig-sdk, add an EVM priority-fee sanity ceiling mirroring the Solana priorityFeeLimit precedent. maxPriorityFeePerGas.ts:4 returns the RPC value verbatim and evm.ts:58 sums it into maxFeePerGas with no clamp. Add a per-chain cap (config-driven, tuned generously vs real congestion) and clamp or refuse an RPC-reported priority fee above it. Add unit tests with an inflated-fee stub. Runtime-verify via SDK-CLI.

---
_Audit finding **SDK2-01** (round 2, High) · bead `agent-backend-ts-mfdst` · runtime before+after is the finish line, not unit-tests-green. When you open a PR, put `closes #<this issue>` in the body._

gomesalexandre:
closes #1055

## Finding (SDK2-01, HIGH, fund-safety)

EVM fee estimation had no sanity ceiling on the RPC-reported priority fee — Solana already has one (`solanaConfig.priorityFeeLimit`), EVM didn't. `getEvmMaxPriorityFeePerGas(chain)` returned `estimateMaxPriorityFeePerGas()` verbatim, and `getEvmFeeQuote` then signed `maxFeePerGas = baseFeePerGas + maxPriorityFeePerGas` uncapped into the tx. A compromised or anomalous EVM RPC could inflate `maxPriorityFeePerGas` and drain the user's balance to gas — the tx would still "succeed" from the wallet's point of view, just at a ruinous fee.

Same gap existed on the zkSync L2 path (`client.extend(publicActionsL2()).estimateFee()`), which derives `maxPriorityFeePerGas` independently via viem and wasn't covered by the generic-path fix alone.

## Fix

Added `clampEvmPriorityFee(chain, rpcPriorityFeeWei): bigint` (`packages/core/chain/tx/fee/evm/clampEvmPriorityFee.ts`) — a per-chain sanity ceiling, applied at both `getEvmFeeQuote` read sites:
- the generic path (`maxPriorityFeePerGas = clampEvmPriorityFee(chain, await getEvmMaxPriorityFeePerGas(chain))`)
- the zkSync `estimateFee` L2 path (clamps the priority-fee component only; `baseFeePerGas` is still derived from the raw, unclamped `maxFeePerGas - maxPriorityFeePerGas` split so the implied base fee stays accurate)

On breach it `console.warn`s (chain + rpc value + ceiling) and **clamps, never throws** — a clamp still lets the tx go through at a safe fee; a reject would strand the user mid-flow.

## Ceiling table (generous by design — sanity bound, not a fee optimizer)

| Chain | Ceiling | Sourcing rationale |
|---|---|---|
| Ethereum | 500 gwei | L1 priority fees rarely exceed ~50-100 gwei even during heavy congestion (NFT mints, MEV bursts) → 5-10x margin above realistic p99 |
| Polygon | 3,000 gwei | PoS gas has historically spiked into the hundreds of gwei during congestion → several-x margin above any observed spike |
| Arbitrum, Base, Blast, Optimism, Zksync, Mantle | 50 gwei | Rollup sequencer-priced priority fees typically sit ~0-2 gwei → 25x+ normal, but still tight enough to catch an order-of-magnitude inflation on a chain where fees are otherwise negligible |
| Avalanche, BSC, CronosChain, Hyperliquid, Sei, and any future/unlisted EVM chain | 500 gwei (default) | No well-documented extreme-congestion fee history for these → fall back to the same generous default as Ethereum so a new chain never false-blocks |

Bias is deliberately HIGH: the attack this guards against is orders-of-magnitude inflation (10x-1000x), so even a "too generous" ceiling still catches it while guaranteeing zero false-blocks on legitimate congestion.

## Runtime verification

Colocated `clampEvmPriorityFee.test.ts` (21 tests) proves both:
1. **Legit-path regression guard** — a normal/heavy-congestion priority fee for every represented chain (incl. an 80 gwei Ethereum congestion tip, a 400 gwei Polygon spike) passes through **unchanged**.
2. **Attack-path guard** — an absurdly inflated fee (10,000 gwei on Ethereum, 50,000 gwei on Polygon, 5,000 gwei on Arbitrum/Avalanche/BSC — all orders of magnitude above the ceiling) gets **clamped** to the ceiling, plus edge cases (exactly-at-ceiling passes, one-wei-over clamps, unlisted chain uses the default).

Also added 3 integration tests directly in `getEvmFeeQuote.test.ts` exercising the real `getEvmFeeQuote` wiring (not just the isolated helper) — one legit-pass, one clamp on the generic path, one clamp on the zkSync L2 `estimateFee` path (confirming `baseFeePerGas` still derives from the raw split).

```
$ yarn vitest run packages/core/chain/tx/fee/evm/clampEvmPriorityFee.test.ts --reporter=verbose

 ✓ passes a normal Ethereum priority fee through unchanged (legit-path regression guard) 1ms
 ✓ passes a normal Polygon priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Arbitrum priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Base priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Optimism priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Blast priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Zksync priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Mantle priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Avalanche priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal BSC priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal CronosChain priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Hyperliquid priority fee through unchanged (legit-path regression guard) 0ms
 ✓ passes a normal Sei priority fee through unchanged (legit-path regression guard) 0ms
 ✓ clamps an absurdly inflated Ethereum priority fee (compromised-RPC attack) to the sanity ceiling 1ms
 ✓ clamps an absurdly inflated Polygon priority fee (compromised-RPC attack) to the sanity ceiling 0ms
 ✓ clamps an absurdly inflated Arbitrum priority fee (compromised-RPC attack) to the sanity ceiling 0ms
 ✓ clamps an absurdly inflated Avalanche priority fee (compromised-RPC attack) to the sanity ceiling 0ms
 ✓ clamps an absurdly inflated BSC priority fee (compromised-RPC attack) to the sanity ceiling 0ms
 ✓ clamps a chain without an explicit ceiling entry using the generous default 0ms
 ✓ never clamps a fee that sits exactly at the ceiling 0ms
 ✓ clamps a fee one wei above the ceiling 0ms

stderr (expected, on the clamp cases):
[evm] RPC-reported maxPriorityFeePerGas for Ethereum (500000000001 wei) exceeds the sanity ceiling (500000000000 wei); clamping to the ceiling.

 Test Files  1 passed (1)
      Tests  21 passed (21)
```

`getEvmFeeQuote.test.ts` (11 tests, incl. the 3 new SDK2-01 cases) — all green, unaffected pre-existing gas-limit-buffering tests untouched.

`npx tsc --project .config/tsconfig.json --noEmit` — clean. Lint (`eslint --config .config/eslint.config.mjs`) — clean.

## Risk

Low — additive clamp on an existing read path, defaults to a no-op on any normal fee, never throws.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EVM transaction fee handling to prevent unusually high priority fee estimates from being used as-is.
  * Added per-network safeguards so fee quotes stay within reasonable limits, while normal values continue unchanged.
  * Applied the same protection across supported EVM fee estimation paths, including ZkSync.
* **Tests**
  * Expanded coverage for normal, boundary, and inflated fee scenarios to verify the new behavior.

### DIFF
--- a/packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts
+++ b/packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts
@@ -2,6 +2,7 @@ import { Chain, EvmChain } from '@vultisig/core-chain/Chain'
 import { evmChainInfo } from '@vultisig/core-chain/chains/evm/chainInfo'
 import { getEvmClient } from '@vultisig/core-chain/chains/evm/client'
 import { getEvmBaseFee } from '@vultisig/core-chain/tx/fee/evm/baseFee'
+import { clampEvmPriorityFee } from '@vultisig/core-chain/tx/fee/evm/clampEvmPriorityFee'
 import { getEvmMaxPriorityFeePerGas } from '@vultisig/core-chain/tx/fee/evm/maxPriorityFeePerGas'
 import { FeeSettings } from '@vultisig/core-mpc/keysign/chainSpecific/FeeSettings'
 import { getKeysignSwapPayload } from '@vultisig/core-mpc/keysign/swap/getKeysignSwapPayload'
@@ -154,7 +155,7 @@ export const getEvmFeeQuote = async ({
           return {
             gasLimit: capGasLimit(gasLimit),
             baseFeePerGas: maxFeePerGas - maxPriorityFeePerGas,
-            maxPriorityFeePerGas,
+            maxPriorityFeePerGas: clampEvmPriorityFee(chain, maxPriorityFeePerGas),
           }
         }
       }
@@ -179,7 +180,7 @@ export const getEvmFeeQuote = async ({
 
     const baseFeePerGas = baseFeeMultiplier(await getEvmBaseFee(chain))
 
-    const maxPriorityFeePerGas = await getEvmMaxPriorityFeePerGas(chain)
+    const maxPriorityFeePerGas = clampEvmPriorityFee(chain, await getEvmMaxPriorityFeePerGas(chain))
 
     return {
       gasLimit,
--- a/packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts
+++ b/packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts
@@ -154,7 +154,15 @@ export const getEvmFeeQuote = async ({
           const { gasLimit, maxFeePerGas, maxPriorityFeePerGas } = result.data
           return {
             gasLimit: capGasLimit(gasLimit),
-            baseFeePerGas: maxFeePerGas - maxPriorityFeePerGas,
+            // Floor at 0: on the zkSync path baseFeePerGas is derived from the
+            // raw split, but downstream maxFeePerGas is rebuilt as
+            // baseFeePerGas + clamp(priority). A compromised RPC returning a
+            // malformed tuple (maxFee < priority, the exact clamp-attack case)
+            // would otherwise yield a negative baseFeePerGas and a negative
+            // signed maxFeePerGas. Flooring keeps the rebuilt maxFee >= the
+            // clamped tip and preserves the "never emits a malformed value"
+            // contract (NeOMakinG preferably-blocking on #1078 / SDK2-01).
+            baseFeePerGas: bigIntMax(0n, maxFeePerGas - maxPriorityFeePerGas),
             maxPriorityFeePerGas: clampEvmPriorityFee(chain, maxPriorityFeePerGas),
           }
         }
--- a/.changeset/evm-priority-fee-sanity-ceiling.md
+++ b/.changeset/evm-priority-fee-sanity-ceiling.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(evm): sanity-cap RPC-reported priority fee (SDK2-01). EVM fee estimation trusted `maxPriorityFeePerGas` from the RPC verbatim into the signed tx (Solana already had a ceiling for this, EVM didn't) — a compromised or anomalous RPC could inflate it and drain the user's balance to gas. Added a generous per-chain sanity ceiling (`clampEvmPriorityFee`) that only fires on orders-of-magnitude inflation; normal congestion on any chain passes through unchanged.
--- a/packages/core/chain/package.json
+++ b/packages/core/chain/package.json
@@ -1857,6 +1857,11 @@
       "import": "./dist/tx/fee/evm/baseFee.js",
       "default": "./dist/tx/fee/evm/baseFee.js"
     },
+    "./tx/fee/evm/clampEvmPriorityFee": {
+      "types": "./dist/tx/fee/evm/clampEvmPriorityFee.d.ts",
+      "import": "./dist/tx/fee/evm/clampEvmPriorityFee.js",
+      "default": "./dist/tx/fee/evm/clampEvmPriorityFee.js"
+    },
     "./tx/fee/evm/evmGasLimit": {
       "types": "./dist/tx/fee/evm/evmGasLimit.d.ts",
       "import": "./dist/tx/fee/evm/evmGasLimit.js",
--- a/packages/core/chain/tx/fee/evm/clampEvmPriorityFee.test.ts
+++ b/packages/core/chain/tx/fee/evm/clampEvmPriorityFee.test.ts
@@ -1,0 +1,58 @@
+import { EvmChain } from '@vultisig/core-chain/Chain'
+import { describe, expect, it } from 'vitest'
+
+import { clampEvmPriorityFee } from './clampEvmPriorityFee'
+
+const gwei = (n: number) => BigInt(n) * 1_000_000_000n
+
+describe('clampEvmPriorityFee', () => {
+  it.each([
+    ['Ethereum', EvmChain.Ethereum, gwei(80)], // heavy L1 congestion tip
+    ['Polygon', EvmChain.Polygon, gwei(400)], // congestion spike
+    ['Arbitrum', EvmChain.Arbitrum, gwei(1)], // typical L2 tip
+    ['Base', EvmChain.Base, gwei(1)],
+    ['Optimism', EvmChain.Optimism, gwei(1)],
+    ['Blast', EvmChain.Blast, gwei(1)],
+    ['Zksync', EvmChain.Zksync, gwei(1)],
+    ['Mantle', EvmChain.Mantle, gwei(1)],
+    ['Avalanche', EvmChain.Avalanche, gwei(25)],
+    ['BSC', EvmChain.BSC, gwei(3)],
+    ['CronosChain', EvmChain.CronosChain, gwei(5)],
+    ['Hyperliquid', EvmChain.Hyperliquid, gwei(1)],
+    ['Sei', EvmChain.Sei, gwei(5)],
+  ])('passes a normal %s priority fee through unchanged (legit-path regression guard)', (_label, chain, fee) => {
+    expect(clampEvmPriorityFee(chain, fee)).toBe(fee)
+  })
+
+  it.each([
+    ['Ethereum', EvmChain.Ethereum, gwei(10_000)],
+    ['Polygon', EvmChain.Polygon, gwei(50_000)],
+    ['Arbitrum', EvmChain.Arbitrum, gwei(5_000)],
+    ['Avalanche', EvmChain.Avalanche, gwei(10_000)],
+    ['BSC', EvmChain.BSC, gwei(10_000)],
+  ])(
+    'clamps an absurdly inflated %s priority fee (compromised-RPC attack) to the sanity ceiling',
+    (_label, chain, fee) => {
+      const clamped = clampEvmPriorityFee(chain, fee)
+
+      expect(clamped).toBeLessThan(fee)
+      expect(clamped).toBeGreaterThan(0n)
+    }
+  )
+
+  it('clamps a chain without an explicit ceiling entry using the generous default', () => {
+    const absurd = gwei(1_000_000)
+
+    expect(clampEvmPriorityFee(EvmChain.Sei, absurd)).toBe(500n * 1_000_000_000n)
+  })
+
+  it('never clamps a fee that sits exactly at the ceiling', () => {
+    expect(clampEvmPriorityFee(EvmChain.Ethereum, gwei(500))).toBe(gwei(500))
+  })
+
+  it('clamps a fee one wei above the ceiling', () => {
+    const oneOverCeiling = gwei(500) + 1n
+
+    expect(clampEvmPriorityFee(EvmChain.Ethereum, oneOverCeiling)).toBe(gwei(500))
+  })
+})
--- a/packages/core/chain/tx/fee/evm/clampEvmPriorityFee.ts
+++ b/packages/core/chain/tx/fee/evm/clampEvmPriorityFee.ts
@@ -1,0 +1,58 @@
+import { EvmChain } from '@vultisig/core-chain/Chain'
+
+const GWEI = 1_000_000_000n
+
+/**
+ * Sanity ceiling (wei) on the RPC-reported maxPriorityFeePerGas, per chain.
+ *
+ * This is NOT a fee optimizer — it exists solely to catch a compromised or
+ * anomalous RPC returning a wildly inflated priority fee (10x-1000x normal)
+ * that would otherwise be trusted verbatim into the signed tx and drain the
+ * user's balance to gas. Ceilings are deliberately generous — several times
+ * above any chain's realistic p99 congestion fee — so legitimate network
+ * congestion is never mistaken for an attack and clamped.
+ *
+ * Sourcing (directional, not a live oracle):
+ * - Ethereum L1: priority fees rarely exceed ~50-100 gwei even during heavy
+ *   congestion (NFT mints, MEV bursts) -> 500 gwei ceiling, 5-10x margin.
+ * - Polygon PoS: gas has historically spiked into the hundreds of gwei
+ *   during congestion -> 3,000 gwei ceiling, well above any observed spike.
+ * - Rollup L2s (Arbitrum, Base, Blast, Optimism, Zksync, Mantle): sequencer
+ *   priority fees are typically ~0-2 gwei -> 50 gwei ceiling. Still 25x+
+ *   normal, but tight enough to catch an order-of-magnitude inflation on a
+ *   chain where fees are otherwise negligible.
+ * - Avalanche C-Chain, BSC, CronosChain, Hyperliquid, Sei, and any future
+ *   EVM chain: no well-documented extreme-congestion fee history -> fall
+ *   back to the generous default ceiling below.
+ */
+const priorityFeeCeilingWeiByChain: Partial<Record<EvmChain, bigint>> = {
+  [EvmChain.Ethereum]: 500n * GWEI,
+  [EvmChain.Polygon]: 3_000n * GWEI,
+  [EvmChain.Arbitrum]: 50n * GWEI,
+  [EvmChain.Base]: 50n * GWEI,
+  [EvmChain.Blast]: 50n * GWEI,
+  [EvmChain.Optimism]: 50n * GWEI,
+  [EvmChain.Zksync]: 50n * GWEI,
+  [EvmChain.Mantle]: 50n * GWEI,
+}
+
+const defaultPriorityFeeCeilingWei = 500n * GWEI
+
+/**
+ * Clamps an RPC-reported EVM maxPriorityFeePerGas to a generous per-chain
+ * sanity ceiling. Never throws — a clamp still lets the tx go through at a
+ * safe fee, whereas rejecting it would strand the user mid-flow.
+ */
+export const clampEvmPriorityFee = (chain: EvmChain, rpcPriorityFeeWei: bigint): bigint => {
+  const ceiling = priorityFeeCeilingWeiByChain[chain] ?? defaultPriorityFeeCeilingWei
+
+  if (rpcPriorityFeeWei <= ceiling) {
+    return rpcPriorityFeeWei
+  }
+
+  console.warn(
+    `[evm] RPC-reported maxPriorityFeePerGas for ${chain} (${rpcPriorityFeeWei} wei) exceeds the sanity ceiling (${ceiling} wei); clamping to the ceiling.`
+  )
+
+  return ceiling
+}
--- a/packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.test.ts
+++ b/packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.test.ts
@@ -186,6 +186,32 @@ describe('getEvmFeeQuote gas limit buffering', () => {
     expect(mocks.getEvmBaseFee).not.toHaveBeenCalled()
   })
 
+  it('floors the ZkSync base fee at 0 for a malformed compromised-RPC tuple (maxFee < priority > ceiling) — SDK2-01 preferably-blocking', async () => {
+    // The exact clamp-attack case NeOMakinG flagged on #1078: a compromised RPC
+    // returns a malformed tuple where maxFeePerGas < maxPriorityFeePerGas and the
+    // priority fee exceeds the 50 gwei ZkSync ceiling. baseFeePerGas is derived
+    // from the RAW split (maxFee - priority), which would go negative; downstream
+    // maxFeePerGas = baseFeePerGas + clamp(priority) must NOT become negative.
+    mocks.getKeysignCoin.mockReturnValue(makeCoin(EvmChain.Zksync))
+    mocks.zksyncFeeClient.estimateFee.mockResolvedValue({
+      gasLimit: 700_001n,
+      maxFeePerGas: 1n, // 1 wei — absurdly below the tip (malformed)
+      maxPriorityFeePerGas: 200n * 1_000_000_000n, // 200 gwei, > 50 gwei ZkSync ceiling
+    })
+
+    const quote = await getEvmFeeQuote({
+      keysignPayload: {} as never,
+      minimumGasLimit: 600_000n,
+    })
+
+    // Raw split would be 1n - 200_000_000_000n (negative); floored to 0n.
+    expect(quote.baseFeePerGas).toBe(0n)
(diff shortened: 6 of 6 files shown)


gomesalexandre (commented):
Reviewed as: Claude correctness pass (manual, Codex degraded so no adversarial lane this round). Ran the test suite locally, traced both read sites + the consuming `maxFeePerGas` assembly.

## point 1 — ceiling generosity (false-block risk): LOW, but the 50 gwei rollup ceiling has zero cited data behind it

Sanity-checked each ceiling against p99 congestion behavior:

- **Ethereum 500 gwei** — fine. L1 priority tips rarely clear 50-100 gwei even during NFT-mint/MEV bursts, 5-10x margin is comfortable.
- **Polygon 3,000 gwei** — fine. PoS gas has spiked into the hundreds of gwei historically, several-x margin holds.
- **Rollup L2s (Arbitrum/Base/Blast/Optimism/Zksync/Mantle) 50 gwei** — directionally fine (these are FCFS-ordered centralized-sequencer chains, tips are basically vestigial ~0-2 gwei, no auction dynamic to spike them the way L1 does), but the code comment says "sourcing (directional, not a live oracle)" — i.e. this is asserted, not measured. 50 gwei is 25-50x the test fixture's own "normal" (1 gwei). Not going to call it a blocker, but:
  - `should-fix`: the ONLY signal when this fires is `console.warn`. If a real L2 congestion event ever legitimately clears 50 gwei (new to any of these chains historically, but "never happened yet" isn't a guarantee), the clamp silently degrades the user's tx to underpriced/stuck with nothing but a console line nobody's tailing in prod. Wire this into whatever telemetry/Sentry breadcrumb exists elsewhere in the SDK so a clamp-fire is observable and the ceiling can be revisited, not discovered via a support ticket.
- **Default 500 gwei (Avalanche/BSC/CronosChain/Hyperliquid/Sei/future chains)** — fine, same margin as Ethereum, sane fallback for anything unlisted.

Verified all 13 `EvmChain` values are accounted for (6 explicit L2 + Ethereum + Polygon explicit, remaining 5 fall to the generous default) — no chain silently gets a tighter-than-intended ceiling. `Chain.ts:13-22` cross-checked against the ceiling map.

## point 3 — base fee is completely unclamped, same trust boundary, zero ceiling (preferably-blocking)

Traced the full consumption path:

```
resolvers/evm.ts:58   maxFeePerGas = baseFeePerGas + maxPriorityFeePerGas
```

`maxPriorityFeePerGas` is now clamped (good). `baseFeePerGas` is `baseFeeMultiplier(await getEvmBaseFee(chain))` where `getEvmBaseFee` is:

```ts
// packages/core/chain/tx/fee/evm/baseFee.ts
export const getEvmBaseFee = async (chain: EvmChain) => {
  const { baseFeePerGas } = await getEvmClient(chain).getBlock()
  return shouldBePresent(baseFeePerGas)
}
```

Straight from the RPC's block header, 1.5x'd, **zero ceiling, zero sanity check**. Same trust boundary as the priority fee the PR is fixing, left completely open.

Nuance on exploitability so this isn't overstated: on a decentralized L1 (Ethereum), `maxFeePerGas` is only a *cap* — actual fee charged at inclusion is `real_base_fee_at_inclusion + min(tip, headroom)`, bounded by the real network regardless of what the compromised RPC told the client the base fee was. So on Ethereum specifically, base-fee inflation mostly just inflates a cap that doesn't translate into overpayment *unless* the compromised RPC also controls block production.

But 6 of the 13 chains this PR explicitly hardens (Arbitrum/Base/Blast/Optimism/Zksync/Mantle) run on a **single centralized sequencer that both serves the RPC AND decides inclusion**. For those, "compromised RPC" and "the entity that actually charges the fee" are very plausibly the same actor — which is exactly the threat model SDK2-01 is written against. A malicious/compromised sequencer-RPC can report an inflated block base fee, get the client to sign a proportionally inflated `maxFeePerGas`, and then simply charge up to that cap when it includes the tx itself. The priority-fee clamp does nothing to stop this.

Confirmed the zkSync L2 path has the identical shape: `baseFeePerGas = maxFeePerGas(raw) - maxPriorityFeePerGas(raw)` (`getEvmFeeQuote.ts:~157`) — that subtraction uses the **raw, unclamped** `maxFeePerGas` from `estimateFee()`, so an attacker can inflate `estimateFee().maxFeePerGas` while leaving `maxPriorityFeePerGas` modest, and 100% of the inflation flows straight through the derived `baseFeePerGas` untouched, then gets re-added to the clamped priority fee at `evm.ts:58`. The comment in the PR body ("baseFeePerGas is still derived from the raw, unclamped split so the implied base fee stays accurate") is technically true but undersells that "accurate" here means "whatever the compromised RPC said," which is the whole vulnerability class.

`preferably-blocking`: this doesn't have to gate *this* PR (SDK2-01 as ticketed is scoped to priority fee, and the fix as scoped is correct and tested) — but it should not be treated as "SDK2-01 closed" in the audit tracker without a fast follow-up. Recommend either extending `clampEvmPriorityFee`-style ceiling to `baseFeePerGas` too (same file, same pattern, cheap to add), or explicitly re-scoping/filing a new HIGH finding for base-fee inflation on sequencer-controlled L2s before closing out the audit item. Given it's the same drain-to-gas primitive on the same attacker profile, I'd rather see it tracked explicitly than quietly left as "well the finding said priority fee."

## point 2 — both read sites actually clamped: confirmed yes

- Generic path: `getEvmFeeQuote.ts` — `maxPriorityFeePerGas: clampEvmPriorityFee(chain, await getEvmMaxPriorityFeePerGas(chain))` ✅
- zkSync L2 `estimateFee` path: `maxPriorityFeePerGas: clampEvmPriorityFee(chain, maxPriorityFeePerGas)` ✅ (both inside `getEvmFeeQuote.ts`, same file, no drift risk between them)

Neither site can leak an unclamped priority-fee value into `maxFeePerGasWei` (base-fee caveat above notwithstanding). Test coverage for both is real — `getEvmFeeQuote.test.ts` exercises the actual wiring (not just the isolated helper), including a dedicated zkSync-clamp case that also asserts `baseFeePerGas` still derives from the raw split.

**One path is intentionally NOT clamped and that's correct**: the `feeSettings` override branch (`getEvmFeeQuote.ts`, top of `getFeeData`) returns `feeSettings.maxPriorityFeePerGas` verbatim, no clamp. Traced `feeSettings` back to `send/build.ts` — it's the user's own explicit custom-fee input from the UI, not RPC-sourced. Out of scope for "compromised RPC" threat model by construction, this is fine as-is, not a gap.

## point 3b — clamp-not-throw validity: confirmed sound

Clamped value is still a valid non-negative `bigint`, flows through the exact same construction path as an unclamped one, produces a broadcastable tx at the ceiling fee rather than a broken one. `console.warn` (not throw) is the right call — a strand-the-user-mid-flow reject would be worse than a clamped-but-still-valid tx. (Telemetry gap noted above is about *observability* of the clamp firing, not correctness of the clamp itself.)

## point 4 — changeset: present and correct

`.changeset/evm-priority-fee-sanity-ceiling.md` — `@vultisig/sdk` patch, message accurately describes the fix and scope. Good.

## point 5 — tests: ran them, all green

```
$ npx vitest run packages/core/chain/tx/fee/evm/clampEvmPriorityFee.test.ts packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.test.ts --exclude "**/.worktrees/**"

 Test Files  2 passed (2)
      Tests  32 passed (32)
```

(had to `--exclude .worktrees/**` — repo has a pile of stale worktree checkouts on this machine with dangling `@vultisig/core-mpc` resolution that vitest was globbing into; unrelated to this PR, not a finding against it.)

## verdict: SHIP-WITH-NITS

The fix does exactly what SDK2-01 asked for — priority fee is clamped at both read sites, generously, never throws, tests actually exercise the real wiring not just the isolated helper, changeset's there. Ship it as the scoped fix.

But don't let this PR close the audit item as "fund-drain-via-RPC fixed" — the base-fee half of the exact same attack (`maxFeePerGas = baseFeePerGas + priorityFee`, `baseFeePerGas` fully RPC-trusted, zero ceiling) is wide open on every centralized-sequencer L2 in this same ceiling table. `preferably-blocking` on filing/scoping that before SDK2-01 gets marked fully resolved — happy to be overruled if the audit finding is intentionally priority-fee-only and base fee is already tracked elsewhere, just didn't see that reference in the PR body.

`should-fix`: wire the clamp-fire `console.warn` into real telemetry so a too-tight ceiling gets caught by monitoring, not a support ticket.

No blockers. Draft PR, comment-only per repo convention.

packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts
```diff
@@ -154,7 +155,7 @@ export const getEvmFeeQuote = async ({
           return {
             gasLimit: capGasLimit(gasLimit),
             baseFeePerGas: maxFeePerGas - maxPriorityFeePerGas,
```
NeOMakinG: `preferably-blocking:` clamping only the priority fee here (while `baseFeePerGas` on line 156 is derived from the **raw** split) de-cancels the algebra: downstream `maxFeePerGas = baseFeePerGas + clamp(priority)`. For a compromised RPC returning `maxFee < priority` with `priority > ceiling`, this goes negative — e.g. `rawMax=1, rawPrio=200 gwei` → downstream `maxFee = -149999999999` → `maxFeePerGasWei="-149999999999"` → malformed hex `-22ecb25c00`. Pre-PR downstream `maxFee` always equalled the raw positive `maxFeePerGas`, so this is a new negative signable value (no fund loss — underpay/rejected tx — but within the PR's own compromised-RPC threat model, and it breaks the "never throws" claim). Floor the base: `baseFeePerGas: bigIntMax(0n, maxFeePerGas - maxPriorityFeePerGas)` (`bigIntMax` already imported line 16), or reject the impossible tuple.

NeOMakinG (commented):
## testing approach

Worktree `/tmp/wt-sdk-1078` @ c856fa43. Traced the priority-fee read → clamp → signable tx by source; ran the algebra for the zkSync split in node; ran Codex gpt-5.4 as an independent adversarial lane (verified on-PR, not cross-contaminated). Non-device (pure fee-math change).

**Fund-safety verification of the core fix — all PASS:**
1. **Clamp prevents grief-overpay** ✅ — `clampEvmPriorityFee` returns `rpcPriorityFeeWei <= ceiling ? rpcPriorityFeeWei : ceiling`. A compromised RPC reporting 10,000 gwei on ETH is clamped to 500 gwei. The tip (`priorityFee`, the actual validator-drain vector) is bounded.
2. **Units** ✅ — everything in wei; `GWEI = 1_000_000_000n`; ceilings are `bigint` wei; RPC value is `bigint` wei from viem. No 10^9 mismatch.
3. **Clamp-down, not floor** ✅ — the guard only ever returns a value ≤ input; it never raises. Exactly-at-ceiling passes unchanged; one-wei-over clamps.
4. **Legit congestion not over-clamped** ✅ — ceilings are 5–25x above realistic p99 (ETH 80 gwei congestion tip and Polygon 400 gwei spike both pass unchanged in tests). L2s get a tighter 50 gwei but still 25x+ their ~0–2 gwei norm.
5. **Fail-safe on missing/odd value** ✅ — `bigint` can't be NaN; a negative bigint (≤ ceiling) passes through but isn't an overpay; the helper never throws. Missing/throwing RPC is pre-existing reject behavior, unchanged.
6. **Reaches the signable tx** ✅ — clamped `maxPriorityFeePerGas` flows into `chainSpecific/resolvers/evm.ts:58` `maxFeePerGas = baseFeePerGas + maxPriorityFeePerGas` → `EthereumSpecificSchema.{maxFeePerGasWei, priorityFee}`. Both the tip and the ceiling reflect the clamp on the generic path.
7. **Per-chain** ✅ — L1 500 / Polygon 3000 / rollup-L2 50 / default 500 gwei; `feeSettings` (manual user fee) correctly not clamped (not RPC-sourced).

The core fix targets the correct variable: the tip is what a colluding validator drains, and clamping it bounds that. An inflated base-fee *estimate* only raises the maxFee ceiling — actual base paid is the real chain rate — so it does not itself cause overpay. The overpay fix is sound.

## preferably-blocking

**zkSync L2 path: the clamp de-cancels the base-fee algebra and can emit a NEGATIVE `maxFeePerGas` into the signed tx.** On the generic path `baseFeePerGas` is independent of the priority fee, so clamping is clean. On the zkSync path `baseFeePerGas = maxFeePerGas − maxPriorityFeePerGas` is derived from the **raw** split, but downstream `maxFeePerGas` is rebuilt as `baseFeePerGas + clamp(maxPriorityFeePerGas)`. When the RPC reports `priority > ceiling` (the exact attack case) AND `maxFee < priority` (a malformed tuple a compromised RPC can return), the result goes negative:

```
rawMax=1 wei, rawPrio=200 gwei, zkSync ceiling=50 gwei
base      = 1 − 200e9        = −199999999999
maxFee    = base + clamp(50) = −149999999999   ← negative signable value
maxFeePerGasWei = "-149999999999";  bigIntToHex(-149999999999) = "-22ecb25c00"  ← malformed hex
```

Pre-PR this was impossible: downstream `maxFee = base + rawPrio = rawMax`, always ≥ 0. The clamp introduces the negative path. It is **not a fund-loss** (negative/tiny maxFee = underpay, the tx won't mine or the signer errors on the malformed hex), and **well-formed tuples are unaffected** (for `rawMax ≥ rawPrio`, downstream `maxFee = (rawMax−rawPrio) + clamp ≥ 0`). But it's a new negative signable value reachable within this PR's own threat model (compromised RPC), and it contradicts the PR's stated "never throws / always a safe fee" contract on the zkSync branch.

Fix — floor the base at 0 (keeps it accurate for honest tuples, `bigIntMax` is already imported at line 16):
```ts
baseFeePerGas: bigIntMax(0n, maxFeePerGas - maxPriorityFeePerGas),
```
Or reject/normalize the impossible `maxFee < priority` tuple before returning.

## suggestion

Changeset bumps `@vultisig/sdk` only. This PR also adds a new export entry to `@vultisig/core-chain` and changes runtime behavior in `@vultisig/core-mpc`; direct consumers of those workspaces get no changelog entry. Matches the repo's prior convention (sdk#1035/#1073/#1075 did the same), so non-blocking — flagging for consistency if those packages are meant to carry their own release notes.

## q

On the zkSync path `baseFeePerGas` is still derived from the unclamped `maxFeePerGas`, so an RPC inflating `maxFeePerGas` (with a normal priority) inflates the signed `maxFeePerGasWei` ceiling. Actual paid is bounded by the real network base fee, so no direct overpay — but worth confirming that's the intended trust boundary (base-fee sanity bound is explicitly out of scope for SDK2-01).

## risk

Low for the overpay objective — additive, no-op on any normal fee, never throws on the generic path. The one open item is the zkSync negative-maxFee edge above (malformed-RPC only, no fund loss). CI: CodeRabbit pass, Integration (Vitest) passed on the prior run; Unit/Lint/Integration re-running at c856fa43 at review time (SDK CI ~16min — poll to settle, don't read pending as fail). PR body shows 21/21 helper + 11/11 resolver tests green.

## verdict

**COMMENT (not approved).** The core grief-overpay fix is correct, well-scoped, well-tested, and reaches the signable tx. Holding approval on the zkSync negative-`maxFeePerGas` edge — it's a small, malformed-RPC-only, no-fund-loss regression, but it's a new negative signable value within the PR's own threat model and the base-floor fix is one line. Same posture as sdk#1035: on a fund-critical signable-fee path, a compromised-RPC value defect gets resolved before approve. Floor the base (or reject the impossible tuple) and this is an approve.

---
_Reviewed by Claude (fund-safety + overpay/bypass lane) + Codex gpt-5.4 (adversarial lane, on-PR, converged on the zkSync negative-base finding independently)._

NeOMakinG (approved):
Re-review (round 2) at 94423bcc. The r1 preferably-blocking (zkSync negative `maxFeePerGas` from a malformed compromised-RPC tuple) is closed. Two-lane: Claude (primary, load-bearing) + Codex gpt-5.4 (adversarial second opinion). CI all green.

## testing approach
- Traced the fix at the derivation point `getEvmFeeQuote.ts:165` → downstream rebuild `chainSpecific/resolvers/evm.ts:58` → signable `EthereumSpecificSchema.maxFeePerGasWei`.
- Verified `bigIntMax` import (`lib-utils/bigint/bigIntMax`, line 16) and semantics (reduce-max, non-negative floor).
- Re-ran the r1 clamp-attack algebra by hand + confirmed the new test asserts it.
- CI: Integration/Vitest (runs the new test) ✅, Lint+Type ✅, Unit ✅, Test Summary ✅, CodeRabbit ✅.
- Codex adversarial lane on a fresh worktree checkout of 94423bcc.

## prior issues — status
| r1 | 1 | preferably-blocking | zkSync path: `baseFeePerGas = maxFee − priority` from raw RPC split, but downstream `maxFee = base + clamp(priority)` — a malformed tuple (`maxFee < priority`, `priority > ceiling`) drove `baseFeePerGas` negative → negative signed `maxFeePerGas` → malformed hex | ✅ fixed | 94423bcc floors at the derivation point: `baseFeePerGas: bigIntMax(0n, maxFeePerGas - maxPriorityFeePerGas)` (`getEvmFeeQuote.ts:165`), applied BEFORE the downstream rebuild (`resolvers/evm.ts:58`). New test `getEvmFeeQuote.test.ts` exercises the exact flagged tuple (maxFee=1 wei, priority=200 gwei > 50 gwei ceiling) → asserts `baseFeePerGas === 0n`, priority clamped to 50 gwei, sum ≥ 0. |
| r1 | 2 | suggestion | changeset bumps `@vultisig/sdk` only, not the standalone `@vultisig/core-chain` / `@vultisig/core-mpc` packages it touches | ℹ️ documented | Matches repo convention (sdk#1035/1073/1075 same). Non-blocking, unchanged. |

Verification of the closed path: RPC hex quantities are parsed unsigned by viem (`estimateFee`/`estimateMaxPriorityFeePerGas` → `hexToBigInt`), so `maxFee ≥ 0` and `priority ≥ 0` from any RPC. The only negative source was `maxFee − priority` when `priority > maxFee`, which the floor now clamps to 0. Downstream `base(≥0) + clamp(priority≥0) ≥ 0` — negative signable `maxFeePerGas` is now impossible on the zkSync path. Priority clamp ceiling (`clampEvmPriorityFee`, 50 gwei zkSync) is unchanged and still holds. Normal well-formed tuples (`maxFee ≥ priority`, `priority ≤ ceiling`) are byte-identical to pre-fix — floor and clamp are both no-ops — so no legit-path regression. On a malformed tuple the result is a floored/clamped sane fee (base 0 + clamped tip), which is fail-safe (may underprice → tx doesn't land, never negative/garbage/overpay), consistent with the helper's never-throw contract.

## should-fix
_(none blocking)_

## suggestion
- Codex flagged that the clamp lives only inside the keysign `getEvmFeeQuote` path; the standalone SDK display/estimate helpers `evmTxInfo()` (`tools/evm/evmTxInfo.ts`) and `SwapService.extractFees()` still rebuild fees from raw `getEvmMaxPriorityFeePerGas()` without the ceiling. These are info/quote-display surfaces, not the signed-tx path (which is guarded), so the SDK2-01 fund-drain vector is closed — but for defense-in-depth/consistency a follow-up could route those through `clampEvmPriorityFee` too. Out of scope for this PR; non-blocking.

## q
- Codex noted a theoretical lower-bound gap: `clampEvmPriorityFee` only caps the upper bound, and `toEvmTwAmount` (`Buffer.from(bigIntToHex(neg), 'hex')`) would collapse a negative to an empty/garbled buffer. Within this PR's threat model this is unreachable — RPC-sourced values are unsigned (hex quantities) and the base is now floored, so no negative reaches serialization. The only carrier of a negative would be caller-supplied `feeSettings.maxPriorityFeePerGas`, which is trusted user input by design (a user setting their own fee), not the compromised-RPC vector. Flagging only so it's a conscious call — no change requested.

## risk
Low. Minimal one-line floor at the correct derivation point, guarded by a new regression test, normal-path byte-identical, all CI green. Closes the only negative-`maxFeePerGas` path in the RPC threat model.

## verdict
Approve. The r1 preferably-blocking is fixed at the right point; negative/garbage signed `maxFeePerGas` is impossible on the zkSync path for any RPC-returned tuple, the priority clamp ceiling holds, malformed tuples fail safe, the floored fee reaches the signable tx, and well-formed tuples are unchanged. Codex's residual notes are out-of-scope display paths / by-design trusted input — non-blocking.

---
🤖 dual-lane review: Claude (primary) + Codex gpt-5.4 (adversarial). Generated with [Claude Code](https://claude.com/claude-code).

PR #1081, opened 2026-07-08
fix(utxo): raise litecoin dust threshold to the real p2wpkh standard (UTXO-02)

### LINKED ISSUE #1054: [audit-UTXO-02] Litecoin dust threshold hardcoded ~3-5x below the real network standard (the dangerous direction)
## Context (fund-safety audit round 2, 2026-07-08 · finding UTXO-02 · High)

chains/utxo/tx.ts:56 and core/chain/chains/utxo/minUtxo.ts:7 both hardcode Litecoin dustLimit: 1_000n litoshi. Litecoin Core's DUST_RELAY_TX_FEE is 10x Bitcoin's; the real P2WPKH dust is ~2,940 litoshi (legacy ~5,460). Duplicated constant, two files.

## Impact
A change output between 1,001 and ~2,939 litoshi is treated as spendable by our code but is non-standard dust under real LTC relay policy => risks non-relay/rejection, leaving the send stuck/unconfirmed. Errs LOW = the dangerous direction.

## Anchors
`chains/utxo/tx.ts:56` · `core/chain/chains/utxo/minUtxo.ts:7`.

## Fix
Bump the Litecoin dust limit to the real P2WPKH standard (~2,940 litoshi) in BOTH files and de-duplicate into one shared constant.

SDK-CLI: unit-drive dust handling at a 1,001-2,939 litoshi change output; BEFORE treated spendable, AFTER treated as dust (folded into fee / no dust change). Cross-check the constant vs LTC policy.h.

## Prompt for AI agents
In vultisig-sdk, fix the Litecoin dust limit: chains/utxo/tx.ts:56 and core/chain/chains/utxo/minUtxo.ts:7 both hardcode 1000n litoshi, ~3-5x below the real LTC P2WPKH dust (~2940 litoshi; legacy ~5460). Bump to the correct value and de-duplicate into a single shared constant. Add a unit test at the dust boundary. Runtime-verify via SDK-CLI.

---
_Audit finding **UTXO-02** (round 2, High) · bead `agent-backend-ts-pku6m` · runtime before+after is the finish line, not unit-tests-green. When you open a PR, put `closes #<this issue>` in the body._

gomesalexandre:
closes #1054

## tl;dr
Litecoin's `DUST_RELAY_TX_FEE` is ~10x Bitcoin's, so LTC's real standard dust threshold is **~2_940 litoshi** for a P2WPKH output — not the hardcoded `1_000n`. At `1_000n`, a change output of **1_001..2_939 litoshi** was treated as spendable but is **non-standard dust** → the tx can be rejected by relays or stall.

## Fix
Bump LTC `1_000n → 2_940n` in **both** dust maps (kept in sync via comments):
- `packages/core/chain/chains/utxo/minUtxo.ts` — core send / keysign paths
- `packages/sdk/src/chains/utxo/tx.ts` `UTXO_SPECS` — sdk tx builder (`change > dustLimit`)

LTC-scoped per the finding: BCH/Dash/Zcash `1_000n` already sits at/above their real ~546 (BTC-like relay fee) dust → conservative, not dangerous → left as-is.

## Runtime (SDK-CLI vitest, 7/7 pass)
Dust-boundary tests pin `2_940n` on both maps and assert a 1_001..2_939 change output is now **DUST** (`change > dustLimit` is false) while `>2_940` stays spendable; Bitcoin/Dogecoin/etc unchanged.

Note for reviewer: please sanity-check the **10x multiplier** (→ 2_940 P2WPKH / ~5_460 legacy). I used the P2WPKH standard since LTC's spec is p2wpkh; if LTC's dust relay fee is actually higher (some sources cite up to 100x historically), a more conservative bump only absorbs marginally more tiny change to fee (never a stuck tx), so err higher if unsure.

Changeset: `.changeset/ltc-dust-threshold-utxo02.md` (`@vultisig/sdk` patch). Audit finding **UTXO-02** (round 2). Codex degraded → manual review.

### DIFF
--- a/.changeset/ltc-dust-threshold-utxo02.md
+++ b/.changeset/ltc-dust-threshold-utxo02.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(utxo): raise the Litecoin dust threshold to the real P2WPKH standard (~2_940 litoshi)
+
+Litecoin's DUST_RELAY_TX_FEE is ~10x Bitcoin's, so LTC's real standard dust threshold is ~2_940 litoshi for a P2WPKH output, not the hardcoded 1_000n. At 1_000n a change output of 1_001..2_939 litoshi was treated as spendable but is non-standard dust, so the transaction could be rejected by relays or stall. Bumped in both dust maps (core `minUtxo` + sdk `UTXO_SPECS`) and kept in sync. LTC-scoped (UTXO-02); BCH/Dash/Zcash 1_000n already sits at/above their real dust.
--- a/packages/core/chain/chains/utxo/minUtxo.test.ts
+++ b/packages/core/chain/chains/utxo/minUtxo.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, it } from 'vitest'
+
+import { Chain } from '../../Chain'
+import { minUtxo } from './minUtxo'
+
+// UTXO-02 (audit r2): Litecoin's dust threshold was hardcoded 1_000n — ~3x below the real P2WPKH standard
+// (~2_940 litoshi; LTC's DUST_RELAY_TX_FEE is ~10x Bitcoin's). A change output of 1_001..2_939 was treated
+// as spendable but is non-standard dust -> the tx can be rejected by relays / stall. This pins the corrected
+// value + the dust-boundary semantics used by the change-output guard (`change > dustLimit`).
+describe('minUtxo — Litecoin dust threshold (UTXO-02)', () => {
+  const ltcDust = minUtxo[Chain.Litecoin]
+
+  it('is the P2WPKH standard ~2_940 litoshi, NOT the old dangerous 1_000n', () => {
+    expect(ltcDust).toBe(2_940n)
+    expect(ltcDust).toBeGreaterThan(1_000n) // the pre-fix value that under-counted dust
+  })
+
+  it('treats a 1_001..2_939 change output as DUST (change > dustLimit is false)', () => {
+    // these were spendable at the old 1_000n limit; now correctly below the standard dust floor
+    for (const change of [1_001n, 2_000n, 2_939n]) {
+      expect(change > ltcDust).toBe(false)
+    }
+  })
+
+  it('treats a change output strictly above the threshold as spendable', () => {
+    expect(2_941n > ltcDust).toBe(true)
+    expect(2_940n > ltcDust).toBe(false) // exactly at the limit is not "> limit" -> absorbed to fee, safe
+  })
+
+  it('leaves the other UTXO chains unchanged (LTC-scoped fix)', () => {
+    expect(minUtxo[Chain.Bitcoin]).toBe(546n)
+    expect(minUtxo[Chain.Dogecoin]).toBe(1_000_000n)
+    // BCH/Dash/Zcash 1_000n is already >= their ~546 BTC-like dust (conservative, not dangerous) — left as-is.
+    expect(minUtxo[Chain.BitcoinCash]).toBe(1_000n)
+    expect(minUtxo[Chain.Dash]).toBe(1_000n)
+    expect(minUtxo[Chain.Zcash]).toBe(1_000n)
+  })
+})
--- a/packages/core/chain/chains/utxo/minUtxo.ts
+++ b/packages/core/chain/chains/utxo/minUtxo.ts
@@ -4,7 +4,13 @@ export const minUtxo: Record<UtxoBasedChain, bigint> = {
   [Chain.Cardano]: 1_400_000n,
   [Chain.Bitcoin]: 546n,
   [Chain.Dogecoin]: 1_000_000n,
-  [Chain.Litecoin]: 1_000n,
+  // UTXO-02 (audit r2): Litecoin's DUST_RELAY_TX_FEE is ~10x Bitcoin's, so its real standard dust threshold
+  // is ~2_940 litoshi for a P2WPKH output (the scriptType LTC uses), ~5_460 legacy P2PKH — NOT 1_000. At the
+  // old 1_000n, a change output of 1_001..2_939 litoshi was treated as spendable but is non-standard dust, so
+  // the tx could be rejected by relays / get stuck. Use the P2WPKH standard (2_940). KEEP IN SYNC with the
+  // per-chain dustLimit map in packages/sdk/src/chains/utxo/tx.ts (UTXO_SPECS). BCH/Dash/Zcash stay at 1_000n
+  // — that's already at/above their real ~546 dust (BTC-like relay fee), so it's conservative, not dangerous.
+  [Chain.Litecoin]: 2_940n,
   [Chain.BitcoinCash]: 1_000n,
   [Chain.Dash]: 1_000n,
   [Chain.Zcash]: 1_000n,
--- a/packages/sdk/src/chains/utxo/tx.dustLimit.test.ts
+++ b/packages/sdk/src/chains/utxo/tx.dustLimit.test.ts
@@ -1,0 +1,28 @@
+import { describe, expect, it } from 'vitest'
+
+import { getUtxoChainSpec } from './tx'
+
+// UTXO-02 (audit r2): the sdk keeps its OWN per-chain dustLimit map (UTXO_SPECS) separate from core's
+// minUtxo. Both must carry the corrected Litecoin P2WPKH dust — the shared standard 2_940n litoshi (LTC's
+// DUST_RELAY_TX_FEE is ~10x BTC's), not the old 1_000n. This asserts the sdk map; the identical 2_940n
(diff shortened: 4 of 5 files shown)


gomesalexandre (commented):
## verdict: SHIP

Went and pulled the actual Litecoin Core source instead of trusting the PR's math on vibes, since a wrong dust constant in either direction is exactly the kind of thing that looks fine in a unit test and then stalls a real tx.

### 1. dust math — verified correct, independently

`praise:` this is the right number. Cross-checked directly against `litecoin-project/litecoin` source (not a blog post):

```
$ curl -s raw.githubusercontent.com/litecoin-project/litecoin/v0.21.5.5/src/policy/policy.h | grep DUST_RELAY
static const unsigned int DUST_RELAY_TX_FEE = 30'000;
```

vs Bitcoin Core's `DUST_RELAY_TX_FEE = 3000`. That's **exactly 10x**, not the 100x some threads online claim (that 100x figure is conflating a different/historical fee param, not this constant). `GetDustThreshold` in policy.cpp is the unmodified Bitcoin Core formula (`98 * dustRelayFee / 1000` for a P2WPKH spendable output) — Litecoin only forked the constant, not the formula. `98 * 30_000 / 1000 = 2_940`. So `2_940n` is exactly right, not a "safe-ish approximation" — it's the actual relay-standard number. No under-shoot risk.

(worth noting since I went checking: BTC's own entry in the same map is `546n`, i.e. P2PKH-level even though BTC's `scriptType` here is `p2wpkh` — real BTC P2WPKH dust is 294 sat. Pre-existing, not touched by this diff, and erring high is safe so not a blocker — just flagging so nobody "fixes" LTC's real P2WPKH number down to match BTC's conservative one later.)

### 2. false-block check — no legit sends blocked, and there's a bonus catch

`change > dustLimit` at `packages/sdk/src/chains/utxo/tx.ts:555` — confirmed dust change gets folded into fee (`hasChange` just goes false, no output built), not a hard tx reject. That's the correct behavior.

`q:` one thing the review brief assumed that isn't quite true and is worth calling out explicitly: `minUtxo[Chain.Litecoin]` isn't ONLY the change-dust guard. It's also read by:
- `validateUtxoRequirements.ts:26` — gates the **recipient amount** itself (`amount < minUtxo[chain]` → "Minimum send amount is X")
- `validateUtxoRequirements.ts:41` — gates remaining balance / "too little change"
- `getUtxos.ts:21` — filters which UTXOs are even considered spendable (`value > minUtxo[account.chain]`)

So this bump *does* raise the minimum LTC send amount from 1,000 → 2,940 litoshi, and now excludes 1,000..2,940-litoshi UTXOs from the spendable set. I don't think this is a false-block bug though — a "legit" send of 1,500 litoshi was never actually legit on the network (it'd have produced a non-standard dust output at the recipient side too, same as the change case), so tightening the amount-gate to match the corrected dust value is consistent, not a regression. Just flagging that the blast radius is 3 call sites, not 1, and the new tests only cover the raw map + `change > dustLimit` semantics — `validateUtxoRequirements` and `getUtxos` have no existing test coverage at all (pre-existing gap, not introduced here).

`suggestion:` (non-blocking) a follow-up test for `validateUtxoRequirements` asserting the new min-send-amount message at 2_940n would close that gap, but not gating this PR on it since there's no regression, just an untested-but-correct surface.

### 3. sync + changeset

Both maps bumped to `2_940n` and cross-referenced via "KEEP IN SYNC" comments (`minUtxo.ts` ↔ `tx.ts`). Confirmed `packages/sdk/src/chains/utxo/rpc.ts` derives its dust filter from `getUtxoChainSpec(...).dustLimit` (same `UTXO_SPECS` map) rather than a third hardcoded copy — no rogue duplicate anywhere else in the repo. Changeset present (`@vultisig/sdk` patch).

### 4. tests

```
Test Files  2 passed (2)
     Tests  7 passed (7)
```
(ran against the PR branch in an isolated worktree, reused node_modules since no lockfile/package.json diff)

### closing keyword
Body has `closes #1054` up top — good, no action needed.

No blockers, no preferably-blocking. Ship it whenever you flip it out of draft.

NeOMakinG (approved):
LTC dust threshold `1_000n → 2_940n` — the value is correct and the fix is fail-closed. Approving. Two-lane review (Claude load-bearing + Codex gpt-5.4 second-opinion); non-device static + local vitest.

## testing approach
- Verified the value from Bitcoin Core's dust formula: P2WPKH dust = `dustRelayFee.GetFee(98 vB) × …` = **294 sat** at BTC's `DUST_RELAY_TX_FEE` (3000 sat/kvB). LTC's dust relay fee is **10×** BTC's — cross-checked by the well-known LTC legacy dust of **5_460** (= 10 × BTC's 546). So LTC P2WPKH = 10 × 294 = **2_940** ✓. LTC vault outputs are native-segwit P2WPKH (`UTXO_SPECS.Litecoin.scriptType='p2wpkh'`, `bipPurpose:84`), so 2_940 is the right script-type dust for the wallet's own change.
- Traced both apply-stacks: (1) core keysign — `minUtxo[Chain.Litecoin]` → `fixedDustThreshold: Long.fromBigInt(...)` in the WalletCore `Bitcoin.SigningInput` (the MPC signing path); (2) SDK builder — `UTXO_SPECS.dustLimit` → `serializeOutputs` `hasChange = change > dustLimit` (dust change dropped-to-fee, no dust output emitted). Also `validateUtxoRequirements` (min-send + leftover-change error) and UTXO input filters. All now carry 2_940 for LTC; both hardcoded maps kept in sync via cross-referencing comments.
- Ran `packages/core/chain/chains/utxo/minUtxo.test.ts` locally: **4/4 pass** (pins `2_940n`, asserts 1_001..2_939 change is dust, 2_941 spendable, other chains unchanged). The sdk `tx.dustLimit.test.ts` couldn't resolve `@noble/curves` in my worktree (nested-workspace node_modules artifact, not a code issue) — CI ran both green.
- CI: Unit Tests / Integration (Vitest) / Lint+TypeCheck / CodeRabbit all SUCCESS. Changeset present (`@vultisig/sdk` patch).

## blocking
_none._

## preferably-blocking
_none._

## should-fix
- **Recipient-output dust is not script-type-aware (residual gap, follow-up — not a regression).** `validateUtxoRequirements` enforces one chain-wide floor (`amount < minUtxo[chain]`), now 2_940 for LTC, but the builder emits the recipient output using the decoded `toAddress` type (p2wpkh/p2pkh/p2sh — `buildScriptPubKey`). A send of 2_940..5_459 litoshi to a **legacy** `L…`/`M…` LTC recipient passes the new floor yet is below legacy P2PKH/P2SH dust (~5_460, as your own comment notes) → that output can still be non-standard dust. This is **out of UTXO-02's scope** (the audit finding is the change-output dust, which is fixed correctly) and the PR strictly *improves* on the old 1_000 floor — so it doesn't block. Worth a follow-up: make the recipient dust floor depend on the recipient's script type rather than a single chain-wide constant. (Codex-converged.)

## suggestion
_none._

## q
- **Input-selection filters now exclude LTC UTXOs valued 1_001..2_940.** `getUtxos`/`getDashUtxos`/`rpc.ts` filter confirmed inputs with `value > minUtxo[chain]` (output-dust policy reused for input selection — pre-existing conflation, not introduced here). Raising LTC to 2_940 drops those tiny confirmed UTXOs from the spendable set, slightly understating max-send. These are genuinely uneconomical-to-spend sub-dust inputs so excluding them is defensible/conservative and no funds are lost (still in the wallet) — flagging only so it's a conscious choice. (Codex-converged.)
- **`change > dustLimit` (strict `>`) burns an exact-2_940 change to fee.** At `change === 2_940`, `2_940 > 2_940` is false → absorbed to fee rather than emitted. This is the **safe** direction (never creates a dust output), pre-existing, and the test documents it intentionally ("exactly at the limit … absorbed to fee, safe"). At most ~2_940 litoshi (~$0.003) to fee in the exact-boundary case. No action needed.

## risk
Low. The change is a one-constant raise on a single chain (Litecoin), scoped correctly (BTC 546 / DOGE 1M / BCH·Dash·Zcash 1_000 untouched), fail-closed in every apply-site (drop dust change to fee, reject sub-threshold min-send, exclude sub-dust inputs), reaching both the WalletCore MPC keysign path and the SDK builder. Worst-case of the raise is over-conservatism (a marginally-larger tiny change absorbed to fee), never a stuck/rejected tx — the exact failure mode the PR removes. Units correct (litoshi bigint). Two hardcoded maps are a drift risk but each is pinned by its own test to the same literal.

## verdict
**APPROVE.** `2_940n` is the correct real Litecoin P2WPKH standard dust (10× BTC's 294; corroborated by the legacy 5_460 = 10× 546), applied fail-closed across both build paths, LTC-scoped, units correct, changeset present, CI green, boundary semantics tested. The residual recipient-legacy-dust gap and the input-filter/boundary notes are non-blocking follow-ups outside UTXO-02's scope.

<sub>🤖 two-lane review — Claude (load-bearing) + Codex gpt-5.4 (second opinion, converged on the recipient-dust + input-filter observations).</sub>

PR #1082, opened 2026-07-08
fix(swap): validate CowSwap order fields against the request before signing (AGG-01)

### LINKED ISSUE #1050: [audit-AGG-01] CowSwap order tokens/kind are taken from the API response, not validated against the request
## Context (fund-safety audit round 2, 2026-07-08 · finding AGG-01 · High)

getCowSwapQuote.ts:88-97 sets sellToken/buyToken/kind/partiallyFillable straight from the CoW /quote response — while `receiver` (two lines away) correctly uses the caller's value. This is the object that gets EIP-712-signed then POSTed to the orderbook.

## Impact
A compromised/buggy apiBase response (proxy compromise, upstream bug) could substitute a token address, or flip kind from sell to buy (silently inverting the gross-sell math), and the SDK signs it without complaint. The EIP-712 digest itself is verified correct — the gap is trusting the response's contents.

## Anchors
`getCowSwapQuote.ts:88-97`.

## Fix
Assert the response's sellToken/buyToken/kind match the REQUEST before signing; reject on mismatch (mirror how `receiver` already uses the caller's value).

SDK-CLI: unit-drive getCowSwapQuote with a stubbed response whose sellToken/buyToken/kind differ from the request; BEFORE the order is built from response values, AFTER it is rejected.

## Prompt for AI agents
In vultisig-sdk getCowSwapQuote.ts:88-97, stop trusting the CoW /quote response for fund-critical order fields: assert response.sellToken/buyToken/kind equal the caller's request values (as `receiver` already does) before EIP-712-signing + POSTing to the orderbook; reject on mismatch. Add unit tests with a mismatched stub response. Runtime-verify via SDK-CLI. Grep open SDK swap PRs first to avoid duplicating in-flight work.

---
_Audit finding **AGG-01** (round 2, High) · bead `agent-backend-ts-o6vfs` · runtime before+after is the finish line, not unit-tests-green. When you open a PR, put `closes #<this issue>` in the body._

gomesalexandre:
Fund-touching SDK core — draft, review requested before merge.

## what
`sellToken`/`buyToken`/`kind`/`partiallyFillable` were taken straight from the untrusted CoW `/quote` HTTP response and signed as-is via the EIP-712 GPv2 `Order` struct. `receiver`, two lines away in the same function, already correctly used the caller-supplied local instead of `quote.receiver` — the "don't trust the API for security-critical fields" pattern already existed here, just wasn't applied to these 4 fields.

A compromised/buggy `apiBase` response could substitute a token address, or flip `kind` from `'sell'` to `'buy'` — silently inverting GPv2's semantics (a sell order's `sellAmount` is authoritative post-fee; a buy order's `buyAmount` is the fixed target, `sellAmount` becomes a ceiling) while `grossSellAmount` is computed assuming sell-order semantics, so the SDK would sign a fundamentally different economic contract than the one it believes it built. The EIP-712 digest construction itself is correct and untouched — the gap was trusting the contents fed into it.

## how
`packages/core/chain/swap/general/cowswap/api/getCowSwapQuote.ts` — asserts `quote.sellToken`/`quote.buyToken` match the request (case-insensitive), `quote.kind === 'sell'` (the SDK only ever requests `'sell'`), and `quote.partiallyFillable === false` (the SDK only ever requests `false`) before constructing the return value. Throws a clear `"CowSwap returned a mismatched <field>; refusing to sign"` error naming the requested vs. returned value on any mismatch.

Same quote-construction-boundary pattern established in the sibling AGG-02 fix (#1079) — the quote function is the trust boundary where the untrusted response gets parsed into a signable `GeneralSwapQuote`.

## tests
- a response matching the request builds (happy path unaffected)
- token comparison is case-insensitive (checksummed request vs. lowercase response is not a false mismatch)
- a response with a swapped `sellToken`/`buyToken` throws
- `kind` flipped to `'buy'` throws
- `partiallyFillable` flipped to `true` throws
- Also fixed 3 pre-existing tests whose mocked response's `sellToken` never actually varied with the requested token (a fixture gap this validation correctly surfaced) by threading an override through `makeQuoteResponse`.

Full core suite: 172/173 test files pass (1 pre-skipped, unrelated). `@vultisig/sdk`'s own suite green. Typecheck clean.

## risk
Low — additive validation only, changes behavior only in the mismatch/attack scenario this closes. Every existing test's fixture already matched request-to-response, confirming the happy path is unaffected.

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened swap quote validation so orders are only signed when key trade details match the original request.
  * Added safeguards against expired or unusually distant quote expiry times.
  * Prevented signing when the requested sell amount does not match the final quoted amount after fees.
  * Improved test coverage for token matching, expiry handling, and quote-field checks.

closes #1050

### DIFF
--- a/.changeset/cowswap-order-request-validation.md
+++ b/.changeset/cowswap-order-request-validation.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(swap): validate CowSwap order fields against the request before signing (AGG-01). sellToken/buyToken/kind/partiallyFillable were taken straight from the untrusted CoW /quote response and signed as-is via the EIP-712 GPv2 Order struct — a compromised/buggy response could substitute a token address or flip kind from 'sell' to 'buy' (inverting GPv2's sell/buy semantics while the SDK's grossSellAmount math still assumes sell-order semantics), and the SDK would sign it. Now asserts each field matches what was actually requested and refuses to build a signable order on any mismatch.
--- a/packages/core/chain/swap/general/cowswap/api/__tests__/getCowSwapQuote.test.ts
+++ b/packages/core/chain/swap/general/cowswap/api/__tests__/getCowSwapQuote.test.ts
@@ -369,8 +369,28 @@ describe('getCowSwapQuote', () => {
         )
       })
 
-      it('accepts a validTo within the normal ~15 minute quote window', async () => {
+      it('accepts a validTo within the mocked ~15 minute quote window', async () => {
         vi.mocked(queryUrl).mockResolvedValueOnce(makeQuoteResponse(1))
+
+        await expect(getCowSwapQuote({ ...baseInput, chainConfig: cowSwapChainConfig.Ethereum })).resolves.toBeDefined()
+      })
+
+      // Team-lead review (PR #1082): the initial 2x/30-min ceiling sat RIGHT AT CoW's real
+      // returned validTo with zero headroom for clock skew — live-re-verified (2026-07-08,
+      // real api.cow.fi/mainnet request) that CoW actually returns validTo = now + 1800s (30
+      // min) exactly, not ~15 min as COWSWAP_VALID_TO_SECONDS alone would suggest. Widened to
+      // 4x/60-min. This test locks in the REAL observed value so a future regression back to
+      // a too-tight ceiling gets caught here, not in production.
+      it('accepts a validTo at the REAL observed CoW value (now + 1800s / 30 min) with headroom to spare', async () => {
+        const realObservedValidTo = Math.floor(Date.now() / 1000) + 1800
+        vi.mocked(queryUrl).mockResolvedValueOnce(makeQuoteResponse(1, undefined, { validTo: realObservedValidTo }))
+
+        await expect(getCowSwapQuote({ ...baseInput, chainConfig: cowSwapChainConfig.Ethereum })).resolves.toBeDefined()
+      })
+
+      it('accepts a validTo with generous headroom for clock skew above the real 30-min value (e.g. 45 min)', async () => {
+        const withSkewMargin = Math.floor(Date.now() / 1000) + 45 * 60
+        vi.mocked(queryUrl).mockResolvedValueOnce(makeQuoteResponse(1, undefined, { validTo: withSkewMargin }))
 
         await expect(getCowSwapQuote({ ...baseInput, chainConfig: cowSwapChainConfig.Ethereum })).resolves.toBeDefined()
       })
--- a/packages/core/chain/swap/general/cowswap/api/__tests__/getCowSwapQuote.test.ts
+++ b/packages/core/chain/swap/general/cowswap/api/__tests__/getCowSwapQuote.test.ts
@@ -18,18 +18,28 @@ vi.mock('@vultisig/lib-utils/query/queryUrl', () => ({
 function makeQuoteResponse(
   chainId: number,
   buyAmount = '990000000000000000',
-  overrides: { sellToken?: string; buyToken?: string; kind?: 'sell' | 'buy'; partiallyFillable?: boolean } = {}
+  overrides: {
+    sellToken?: string
+    buyToken?: string
+    kind?: 'sell' | 'buy'
+    partiallyFillable?: boolean
+    sellAmount?: string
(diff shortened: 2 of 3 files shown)


NeOMakinG (approved):
## testing approach

Two lanes: Claude (load-bearing correctness + fund-safety + fail-closed trace) + Codex `gpt-5.4` (adversarial second opinion, receiver-lane excluded to avoid duplication). Read the full order-construction path `getCowSwapQuote.ts` → caller `findSwapQuote.ts:654` → EIP-712 signing `build.ts:254`. Ran the `getCowSwapQuote` suite locally against installed deps: **29/29 pass**. CI: Integration Tests + CodeRabbit green; Unit/Lint pending at review time (locally verified).

## verdict

**APPROVE.** This closes the AGG-01 gap correctly. Every signed GPv2 `Order` field that carries economic weight is now validated against the *trusted request* (not re-derived from the untrusted `/quote` response), before any signable order is constructed, and fails closed on mismatch.

Field-by-field trace:

- **receiver (the critical one)** — the returned `cowswap_order.receiver` uses the caller-supplied local (`getCowSwapQuote.ts:169`), never `quote.receiver`. The caller passes `receiver: normalizedRecipient ?? from.address` (`findSwapQuote.ts:660`) — the user's own vault, or an explicit custom recipient the user chose in the UI. Bought funds cannot be redirected by a malicious response. ✅
- **sellToken / buyToken** — validated case-insensitively against the trusted function args (`:89-98`), which trace to `from.id` / `to.id` (the requested pair), not the response. Not tautological. ✅
- **sellAmount (over-sell)** — `grossSellAmount = quote.sellAmount + quote.feeAmount` is asserted `=== sellAmount` (the requested `sellAmountBeforeFee`) exactly (`:156`); the signed order's `sellAmount` is that checked gross with `feeAmount='0'`. A response inflating `sellAmount`/`feeAmount` cannot authorize selling more than requested. ✅
- **kind / partiallyFillable** — pinned to SDK-fixed constants `REQUESTED_KIND='sell'` / `REQUESTED_PARTIALLY_FILLABLE=false` (`:99-106`), the same constants used to build the request body — so a `'sell'→'buy'` GPv2-semantics inversion is refused. ✅
- **validTo** — server-determined (no request-side value), so bounded to `(now, now+4·COWSWAP_VALID_TO_SECONDS]` (`:124-130`). The 4x/60-min ceiling was deliberately widened from 2x to clear the real observed 30-min value with clock-skew headroom — avoids false-rejecting legit quotes. ✅
- **buyAmount** — passed through, but the *same* value is both the displayed `dstAmount` and the signed floor, so WYSIWYS holds (Codex-confirmed): tampering changes the economics the user *sees* and can reject, not a hidden contract. ✅
- **appData / appDataHash / feeAmount='0' / sellTokenBalance / buyTokenBalance / chainId** — all locally sourced, never from the response. ✅

**Fail-closed:** every check `throw`s inside `getCowSwapQuote` before the `GeneralSwapQuote` is returned; on throw no `cowswap_order` reaches `build.ts:254` → no EIP-712 digest is signed. Confirmed reaches the sign path.

**Codex lane** (referenced PR #1082 explicitly, not cross-contaminated): no additional findings. Confirmed malformed `sellAmount`/`feeAmount` fail closed at `BigInt(...)` parse, negative components can't bypass the gross-equality check, and the `validTo` quote-time-vs-sign-time TOCTOU only errs toward *earlier* expiry (a benign user-visible failure), never extra executable lifetime.

## praise

`praise:` Comparing against named `REQUESTED_KIND`/`REQUESTED_PARTIALLY_FILLABLE` constants shared with the request body (rather than duplicating the literal) keeps request and validation in lockstep — the right robustness call. Test coverage is exactly on-target: wrong-receiver-class substitutions, both over- and under-sell directions, kind/partiallyFillable flips, validTo past/far-future/real-observed, and the case-insensitive checksum-vs-lowercase non-mismatch.

## risk

Low. Additive validation only; behavior changes solely in the mismatch/attack path. Happy path unaffected (every pre-existing fixture already matched request↔response; the 3 fixture fixes correctly surfaced a mock that never varied `sellToken`). Changeset present (`@vultisig/sdk` patch).

---
🤖 dual-lane review (Claude + Codex gpt-5.4)

PR #1083, opened 2026-07-08
fix(cosmos): refuse to sign a no-timeout ibc transfer + validate channel (COSMOS-01/03)

### LINKED ISSUE #1052: [audit-COSMOS-01] The WalletCore IBC resolver signs an all-zero (no-expiry) timeout when the denom trace is missing
## Context (fund-safety audit round 2, 2026-07-08 · finding COSMOS-01 · High)

signingInputs/resolvers/cosmos/index.ts:98-101 hardcodes timeoutHeight to {0,0}, and :87 derives timeoutTimestamp from ibcDenomTraces?.latestBlock falling to '0' when absent. No guard rejects the all-zero case. The app's own builder throws 'refusing to build a no-timeout MsgTransfer' against exactly this (cosmosTx.ts:2196). Per ibc-go, an all-zero-timeout MsgTransfer fails ValidateBasic. Related COSMOS-03: sourceChannel comes from an unvalidated memo.split(':').

## Impact
A co-signing device trusts the proposer's blockchainSpecific; if any proposer omits ibcDenomTraces, every signer blindly builds a degenerate MsgTransfer. Rejected at CheckTx on canonical ibc-go (wasted signing round) — but on a looser/forked implementation, 'funds parked in escrow with no expiry' is not ruled out.

## Anchors
`signingInputs/resolvers/cosmos/index.ts:98-101,87` · `cosmosTx.ts:2196` (the app guard to port).

## Fix
Port the app's guard: refuse to build a MsgTransfer whose timeoutHeight AND timeoutTimestamp are both zero. Also validate sourceChannel shape (COSMOS-03).

SDK-CLI: unit-drive the cosmos IBC resolver with ibcDenomTraces absent; BEFORE it builds an all-zero-timeout MsgTransfer, AFTER it refuses.

## Prompt for AI agents
In vultisig-sdk signingInputs/resolvers/cosmos/index.ts, port the app builder's no-timeout guard (cosmosTx.ts:2196 'refusing to build a no-timeout MsgTransfer'): reject building a MsgTransfer when both timeoutHeight is {0,0} and timeoutTimestamp resolves to '0' (missing ibcDenomTraces). Also validate the sourceChannel derived from memo.split(':') (COSMOS-03) is a well-formed channel id. Add unit tests for the missing-trace + malformed-channel cases. Runtime-verify via SDK-CLI.

---
_Audit finding **COSMOS-01** (round 2, High) · bead `agent-backend-ts-zpguf` · runtime before+after is the finish line, not unit-tests-green. When you open a PR, put `closes #<this issue>` in the body._

gomesalexandre:
closes #1052

## Finding

The Cosmos IBC-transfer signing-input resolver (`packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts`, `ibcEnabled` -> `IBC_TRANSFER` branch) built a `MsgTransfer` with an ALL-ZERO timeout whenever the IBC denom trace was missing:

- `timeoutTimestamp = Long.fromString(ibcDenomTraces?.latestBlock?.split('_')?.[1] || '0')` -> `0` with NO guard when the trace/latestBlock is absent.
- `timeoutHeight: { revisionNumber: 0, revisionHeight: 0 }` was hardcoded.

Net: trace missing -> `timeoutTimestamp=0` AND `timeoutHeight={0,0}` -> an all-zero-timeout `MsgTransfer` (no expiry). Relayers accept a no-timeout packet, but it never gets a chance to unwind on the destination side, leaving funds stuck indefinitely. The app already refuses to build exactly this case (`vultiagent-app/src/services/cosmosTx.ts:2383`, `buildSignBroadcastCosmosIbcTransfer`) -- the SDK resolver had no equivalent guard.

Also flagged (COSMOS-03, folded into this PR): `sourceChannel` was taken from an unvalidated `memo.split(':')` (`const [, channel] = memo.split(':')`), so an undefined/empty/malformed channel would sign a `MsgTransfer` with a bogus `sourceChannel`.

## Fix

- **COSMOS-01**: fail closed (throw) before building when both `timeoutTimestamp` is zero/empty AND `timeoutHeight` is `{0,0}` -- mirrors the app's exact refusal message ("... refusing to build a no-timeout MsgTransfer (relayers accept it but the packet has no expiry, leaving funds stuck)").
- **COSMOS-03**: validate `channel` against `/^channel-\d+$/` before building; reject empty/undefined/malformed channels instead of signing with a bad `sourceChannel`.

The legit path (trace present + valid `channel-N` in memo) is unaffected -- verified with a regression test.

## Runtime verification (SDK-CLI vitest)

New colocated test: `packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ibcTransfer.test.ts`

```
 RUN  v4.1.8 vultisig-sdk

 Test Files  1 passed (1)
      Tests  5 passed (5)
```

Covers:
- missing `ibcDenomTraces` -> throws (no-timeout refused)
- `ibcDenomTraces.latestBlock` present but empty -> throws
- valid trace + valid `channel-141` in memo -> builds successfully with non-zero `timeoutTimestamp` (regression guard for the legit path)
- malformed channel (`not-a-channel`) in memo -> throws
- no channel segment in memo at all -> throws

Existing `cosmos.test.ts` (gas-limit regression suite) still green (3/3), no regression:

```
 Test Files  2 passed (2)
      Tests  8 passed (8)
```

`tsc --project .config/tsconfig.json --noEmit` clean. `eslint` on the touched files clean. Changeset added (`@vultisig/sdk` patch).

🤖 Agent-generated

### DIFF
--- a/packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts
@@ -85,7 +85,35 @@ export const getCosmosSigningInputs: SigningInputsResolver<'cosmos'> = ({ keysig
         const memo = shouldBePresent(keysignPayload.memo)
         const [, channel] = memo.split(':')
 
+        // COSMOS-03: sourceChannel is derived from an unvalidated memo
+        // split. An undefined/empty/malformed channel would sign a
+        // MsgTransfer that either fails on-chain or (worse) routes
+        // through an unintended channel. Fail closed.
+        if (!channel || !/^channel-\d+$/.test(channel)) {
+          throw new Error(
+            `Cosmos signing input: IBC transfer memo "${memo}" does not contain a well-formed source channel (expected "<prefix>:channel-<n>[:...]").`
+          )
+        }
+
         const timeoutTimestamp = Long.fromString(ibcDenomTraces?.latestBlock?.split('_')?.[1] || '0')
+        const timeoutHeight = {
+          revisionNumber: Long.fromString('0'),
+          revisionHeight: Long.fromString('0'),
+        }
+
+        // COSMOS-01: a missing ibcDenomTraces (or missing latestBlock)
+        // previously fell back to timeoutTimestamp=0, and timeoutHeight
+        // is always {0,0} here, so the packet ended up with NO expiry.
+        // Relayers accept a no-timeout MsgTransfer, but the destination
+        // side never gets a chance to unwind on failure, leaving funds
+        // stuck indefinitely. Mirror the app's guard
+        // (vultiagent-app/src/services/cosmosTx.ts) and refuse to build.
+        const heightDisabled = timeoutHeight.revisionNumber.isZero() && timeoutHeight.revisionHeight.isZero()
+        if (timeoutTimestamp.isZero() && heightDisabled) {
+          throw new Error(
+            'Cosmos signing input: IBC transfer has no usable timeout (missing ibcDenomTraces.latestBlock and no timeoutHeight set); refusing to build a no-timeout MsgTransfer (relayers accept it but the packet has no expiry, leaving funds stuck).'
+          )
+        }
 
         return {
           messages: [
@@ -96,10 +124,7 @@ export const getCosmosSigningInputs: SigningInputsResolver<'cosmos'> = ({ keysig
                 token: getCosmosCoinAmount(keysignPayload),
                 sender: coin.address,
                 receiver: toAddress,
-                timeoutHeight: {
-                  revisionNumber: Long.fromString('0'),
-                  revisionHeight: Long.fromString('0'),
-                },
+                timeoutHeight,
                 timeoutTimestamp,
               }),
             }),
--- a/.changeset/cosmos-ibc-notimeout-channel.md
+++ b/.changeset/cosmos-ibc-notimeout-channel.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(cosmos): refuse to sign a no-timeout IBC transfer and validate the source channel
+
+The Cosmos `IBC_TRANSFER` signing-input resolver built a `MsgTransfer` with an all-zero timeout (`timeoutTimestamp=0` and `timeoutHeight={0,0}`) whenever the IBC denom trace was missing. Relayers accept a no-timeout packet, but it never expires, so a failed transfer can leave funds stuck indefinitely instead of unwinding. The resolver now fails closed and refuses to build when neither timeout is usable (COSMOS-01). It also validates that the source channel parsed out of the memo (`<prefix>:channel-<n>[:...]`) is well-formed, refusing to sign with an undefined/empty/malformed channel instead of dispatching a broken `MsgTransfer` (COSMOS-03).
--- a/packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ibcTransfer.test.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ibcTransfer.test.ts
@@ -1,0 +1,141 @@
+import { Buffer } from 'buffer'
+
+import { create } from '@bufbuild/protobuf'
+import { Chain } from '@vultisig/core-chain/Chain'
+import { initWasm, type WalletCore } from '@trustwallet/wallet-core'
+import {
+  CosmosIbcDenomTraceSchema,
+  CosmosSpecificSchema,
+  TransactionType,
+} from '@vultisig/core-mpc/types/vultisig/keysign/v1/blockchain_specific_pb'
+import { CoinSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/coin_pb'
+import { KeysignPayloadSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
+import { beforeAll, describe, expect, it } from 'vitest'
+
+import { getCosmosSigningInputs } from './index'
+
+// COSMOS-01 / COSMOS-03: an ibc-enabled Cosmos chain's IBC_TRANSFER
+// resolver must refuse to build a no-timeout MsgTransfer (missing
+// ibcDenomTraces => timeoutTimestamp falls back to 0, and
+// timeoutHeight is always {0,0} here) and must refuse a malformed
+// sourceChannel parsed out of the memo.
+describe('getCosmosSigningInputs IBC transfer guards', () => {
+  let walletCore: WalletCore
+  let sender: string
+  let recipient: string
+  let publicKeyHex: string
+
+  beforeAll(async () => {
+    walletCore = await initWasm()
+
+    const privateKey = walletCore.PrivateKey.createWithData(new Uint8Array(32).fill(1))
+    const recipientPrivateKey = walletCore.PrivateKey.createWithData(new Uint8Array(32).fill(2))
+    const publicKey = privateKey.getPublicKeySecp256k1(false)
+    const recipientPublicKey = recipientPrivateKey.getPublicKeySecp256k1(false)
+
+    sender = walletCore.AnyAddress.createWithPublicKey(publicKey, walletCore.CoinType.cosmos).description()
+    recipient = walletCore.AnyAddress.createWithPublicKey(recipientPublicKey, walletCore.CoinType.cosmos).description()
+    publicKeyHex = Buffer.from(publicKey.data()).toString('hex')
+  })
+
+  const buildPayload = ({
+    memo,
+    ibcDenomTraces,
+  }: {
+    memo: string
+    ibcDenomTraces?: { path: string; baseDenom: string; latestBlock: string }
+  }) =>
+    create(KeysignPayloadSchema, {
+      coin: create(CoinSchema, {
+        chain: Chain.Cosmos,
+        ticker: 'ATOM',
+        address: sender,
+        contractAddress: '',
+        decimals: 6,
+        isNativeToken: true,
+        hexPublicKey: publicKeyHex,
+      }),
+      toAddress: recipient,
+      toAmount: '12345',
+      memo,
+      blockchainSpecific: {
+        case: 'cosmosSpecific',
+        value: create(CosmosSpecificSchema, {
+          accountNumber: 7n,
+          sequence: 3n,
+          gas: 2500n,
+          transactionType: TransactionType.IBC_TRANSFER,
+          ibcDenomTraces: ibcDenomTraces ? create(CosmosIbcDenomTraceSchema, ibcDenomTraces) : undefined,
+        }),
+      },
+    })
+
+  it('refuses to build a no-timeout MsgTransfer when ibcDenomTraces is missing (COSMOS-01)', () => {
+    expect(() =>
+      getCosmosSigningInputs({
+        keysignPayload: buildPayload({ memo: 'transfer:channel-141' }),
+        walletCore,
+      })
+    ).toThrow(/refusing to build a no-timeout MsgTransfer/)
+  })
+
+  it('refuses to build when ibcDenomTraces.latestBlock is present but empty (COSMOS-01)', () => {
+    expect(() =>
+      getCosmosSigningInputs({
+        keysignPayload: buildPayload({
+          memo: 'transfer:channel-141',
+          ibcDenomTraces: { path: 'transfer/channel-141', baseDenom: 'uatom', latestBlock: '' },
+        }),
+        walletCore,
+      })
+    ).toThrow(/refusing to build a no-timeout MsgTransfer/)
+  })
+
+  it('builds successfully with a non-zero timeout when the denom trace is present (regression guard)', async () => {
+    const [input] = await getCosmosSigningInputs({
+      keysignPayload: buildPayload({
+        memo: 'transfer:channel-141',
+        ibcDenomTraces: { path: 'transfer/channel-141', baseDenom: 'uatom', latestBlock: '12345_1751328000000000000' },
+      }),
+      walletCore,
+    })
+
+    const transfer = input.messages[0].transferTokensMessage
+    expect(transfer).toBeDefined()
+    expect(transfer!.sourceChannel).toBe('channel-141')
+    expect(transfer!.timeoutTimestamp.toString()).toBe('1751328000000000000')
+    expect(transfer!.timeoutTimestamp.isZero()).toBe(false)
+  })
+
+  it('refuses to build when the memo channel is malformed (COSMOS-03)', () => {
+    expect(() =>
+      getCosmosSigningInputs({
+        keysignPayload: buildPayload({
+          memo: 'transfer:not-a-channel',
+          ibcDenomTraces: {
+            path: 'transfer/channel-141',
+            baseDenom: 'uatom',
+            latestBlock: '12345_1751328000000000000',
+          },
+        }),
+        walletCore,
+      })
+    ).toThrow(/well-formed source channel/)
+  })
+
+  it('refuses to build when the memo has no channel segment at all (COSMOS-03)', () => {
+    expect(() =>
+      getCosmosSigningInputs({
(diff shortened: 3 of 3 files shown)


gomesalexandre (commented):
## Review — draft PR #1083 (COSMOS-01/03, HIGH fund-safety)

Manual pass only (Codex degraded per task). Verdict: **SHIP-WITH-NITS** (draft → `--comment`, matches your gate).

### 1. Guard actually fires on the exact bug — verified, no bypass found

Traced `timeoutTimestamp = Long.fromString(ibcDenomTraces?.latestBlock?.split('_')?.[1] || '0')` end to end w/ the actual `long` npm package (not cosmjs-types' Long, confirmed via `import Long from 'long'` at index.ts:16):

- `latestBlock` missing entirely → `undefined?.split(...)` → `undefined || '0'` → `Long.fromString('0')` → `isZero() === true`. Guard fires. ✅
- `latestBlock: ''` (present-but-empty) → `''.split('_')` → `['']`, `[1]` is `undefined` → falls to `'0'`. Guard fires. ✅ (this is your test #2)
- `latestBlock: '12345'` (no underscore, malformed) → `[1]` undefined → `'0'`. Guard fires.
- Literal string `'0'` as the second segment (`latestBlock: '12345_0'`) → `'0' || '0'` → since `'0'` is a *non-empty* string it's truthy in JS, so the string `'0'` itself is used, not the `'0'` fallback — no accidental double-fallback bug. `Long.fromString('0').isZero()` → `true` either way. Guard fires. **This is the exact "string-'0' vs Long-zero mismatch" case flagged in the review focus — checked it by hand, no gap.**
- Whitespace / garbage in the timestamp segment (` 0`, `notanumber`) → ran it directly against the installed `long` pkg (node repl, see below) — `Long.fromString` does NOT throw on garbage, it silently coerces to `0`. That's actually **safety-reinforcing** here: any malformed segment collapses to zero, which the guard catches. It's a debuggability nit, not a bypass (see nits below).

```
$ node -e "const Long=require('long'); console.log(Long.fromString('abc').toString())"
0
```

Since `timeoutHeight` is hardcoded `{revisionNumber:0, revisionHeight:0}` in this resolver (confirmed — only one `IBC_TRANSFER` branch in the file, no sibling vaultBased duplicate), `heightDisabled` is unconditionally `true` today, so the compound condition `timeoutTimestamp.isZero() && heightDisabled` correctly reduces to "block iff timestamp is zero" — which is exactly the bug. The `heightDisabled` computation is forward-defensive (future-proofs against `timeoutHeight` becoming dynamic later) rather than dead code — nice mirroring of the app's `heightDisabled` naming/shape from `cosmosTx.ts:2373-2383`.

**No bypass vector found for the no-timeout condition.**

### 2. Legit path not false-blocked — confirmed

Ran the actual regression test (`'builds successfully with a non-zero timeout when the denom trace is present'`) — trace present w/ `latestBlock: '12345_1751328000000000000'` + `channel-141` in memo → builds, `sourceChannel === 'channel-141'`, `timeoutTimestamp.isZero() === false`. Legit path unaffected.

### 3. COSMOS-03 channel regex — correct and ordered right

`/^channel-\d+$/` is the correct IBC (ICS-24) channel-id shape — sequential integer suffix, no legit channel would fail this (leading zeros theoretically pass the regex but never occur in practice since channel numbering is sequential, not padded — non-issue).

- `!channel` catches `undefined` (no `:` in memo → destructure yields `undefined`) and empty string, **before** `channel` ever reaches `sourceChannel:` in the built message — confirmed by line order in the diff: channel-guard throw happens immediately after the `const [, channel] = memo.split(':')` line, well before the `return { messages: [...] }` block.
- Order matters and is correct: channel validation runs *before* the timeout guard, both run *before* any `TW.Cosmos.Proto.Message.create(...)` call — so a throw never leaves a partial/invalid signing input on the table. Point 4 (throw-not-fallback) confirmed clean.

One thing worth noting: this SDK's own guard is *stricter* than the app's (`cosmosTx.ts:2297` only checks `sourceChannel === ''`, no format regex). That's a strictly-safer direction (extra defense, can't false-negative into a bad tx), so not a parity issue — just flagging in case anyone expected byte-for-byte parity with the app guard.

### 4. Throw vs silent fallback — confirmed throws, no partial state

Both guards `throw new Error(...)` before object construction. Nothing is returned on the failure path. Matches the app's fail-closed behavior 1:1 in spirit (message text is intentionally near-identical to `cosmosTx.ts:2383`).

### 5. Tests + changeset — verified locally (used the pre-existing worktree at `.worktrees/fix_cosmos_ibc_notimeout_channel`, same commit `4545a980` as the PR head)

```
$ npx vitest run packages/core/mpc/keysign/signingInputs/resolvers/cosmos.test.ts packages/core/mpc/keysign/signingInputs/resolvers/cosmos/
 Test Files  2 passed (2)
      Tests  8 passed (8)
```
5 new (`index.ibcTransfer.test.ts`) + 3 existing (`resolvers/cosmos.test.ts`, gas-limit suite — confirmed it's a *different* file one directory up, not a duplicate/stale copy of the IBC tests). `tsc --project .config/tsconfig.json --noEmit` clean (exit 0, zero output) on the worktree. `eslint` errors out repo-wide on a missing `eslint.config.js` (v9 migration) — pre-existing repo issue, reproduced identically against `main`, not introduced by this PR.

Changeset present (`@vultisig/sdk` patch), accurate summary, matches the fix.

Cross-checked against issue #1052 (audit finding, round 2, High) — scope match is exact: both COSMOS-01 (no-timeout) and COSMOS-03 (channel validation) called out in the issue's "Fix" section, both addressed here. `closes #1052` is present in the PR body (top line) — satisfies the closing-keyword rule.

### Nits (non-blocking)

- **suggestion:** `Long.fromString` silently zeroes garbage input instead of throwing (pre-existing `long` pkg behavior, not introduced here) — means a genuinely malformed `latestBlock` (vs. simply-missing) surfaces the *same* generic "no usable timeout" error rather than a more specific "denom trace latestBlock is malformed" message. Doesn't affect fund-safety (still fails closed), just debuggability if this ever fires in prod telemetry. Not worth blocking on.
- **suggestion:** could add one more explicit test case for `latestBlock: '12345_0'` (trace object *present* but genuinely zero-valued, vs. the tested empty-string case) — same code path, so low value, but would make the "string-'0' vs Long-zero" reasoning explicit in the suite for future readers.

### Verdict

**SHIP-WITH-NITS.** Guard fires precisely on the bug, cannot be bypassed by any input shape I could construct (empty/missing/garbage/whitespace/literal-'0' all collapse correctly to a blocked build), legit path is proven unaffected by its own regression test, channel validation is correctly ordered before the value reaches the signing input, and the fail-closed behavior mirrors the app's already-shipped guard. Tests + tsc verified green locally against the exact PR head commit. Nothing here blocks merge once you're ready to flip out of draft — draft state review posted as comment per your gate (can't approve a draft, and self-authored PRs need an external stamp regardless).

packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts
```diff
@@ -85,7 +85,35 @@ export const getCosmosSigningInputs: SigningInputsResolver<'cosmos'> = ({ keysig
         const memo = shouldBePresent(keysignPayload.memo)
         const [, channel] = memo.split(':')
 
+        // COSMOS-03: sourceChannel is derived from an unvalidated memo
+        // split. An undefined/empty/malformed channel would sign a
+        // MsgTransfer that either fails on-chain or (worse) routes
+        // through an unintended channel. Fail closed.
+        if (!channel || !/^channel-\d+$/.test(channel)) {
```
NeOMakinG, line 92: `should-fix:` channel validation is format-only, not authoritative — a wrong-but-well-formed channel still signs — `/^channel-\d+$/` catches undefined/empty/malformed channels (the COSMOS-03 bug), but `channel-5` and `channel-141` both pass even if one routes to the wrong destination chain — funds sent into a coherent-but-wrong channel land at an address that may not exist on the actual dest, i.e. stuck/lost. The repo ALREADY has an authoritative `(fromChain/channel) -> destChain` registry (`IBC_CHANNEL_DEST`, live-LCD-verified) in `packages/sdk/src/tools/prep/ibcTransfer.ts`. This resolver is the last fund-safety chokepoint before signing and could cross-check `${chainId}/${channel}` (and the dest HRP of `toAddress`) against that table.

fix: Keep the format regex, but additionally ground `channel` against `IBC_CHANNEL_DEST` for the source chain-id (and optionally cross-check the `toAddress` bech32 HRP vs the resolved dest chain, as `prepareIbcTransfer` already does). This turns 'well-formed' into 'correct route'. Non-blocking because the normal flow's channel is produced by trusted upstream builders, but this is where a wrong channel would be caught cheaply.

packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts
```diff
@@ -85,7 +85,35 @@ export const getCosmosSigningInputs: SigningInputsResolver<'cosmos'> = ({ keysig
...
+
+        // COSMOS-01: a missing ibcDenomTraces (or missing latestBlock)
+        // previously fell back to timeoutTimestamp=0, and timeoutHeight
+        // is always {0,0} here, so the packet ended up with NO expiry.
+        // Relayers accept a no-timeout MsgTransfer, but the destination
+        // side never gets a chance to unwind on failure, leaving funds
+        // stuck indefinitely. Mirror the app's guard
+        // (vultiagent-app/src/services/cosmosTx.ts) and refuse to build.
+        const heightDisabled = timeoutHeight.revisionNumber.isZero() && timeoutHeight.revisionHeight.isZero()
+        if (timeoutTimestamp.isZero() && heightDisabled) {
```
NeOMakinG, line 112: `suggestion:` `&& heightDisabled` is always-true here — timeoutHeight is hardcoded {0,0} — `timeoutHeight` is unconditionally `{revisionNumber: 0, revisionHeight: 0}` (lines 99-102), so `heightDisabled` is always `true` and the no-timeout guard reduces to `if (timeoutTimestamp.isZero())`. Correct and fail-closed, just slightly self-obscuring — reads as if a non-zero timeoutHeight could ever satisfy the timeout, which it can't on this path.

fix: Optional: drop the `&& heightDisabled` (or add a comment that timeoutHeight is intentionally always disabled on this branch and timeout is timestamp-only). Purely cosmetic.

NeOMakinG (approved):
gm — multi-lane pass: claude fund-safety + codex gpt-5.4 adversarial. runtime-tested on SDK vitest — worktree @4545a980: index.ibcTransfer.test.ts 5/5 pass + cosmos.test.ts regression green; core-config dist built. lgtm.

## testing approach
- SDK vitest — worktree @4545a980: index.ibcTransfer.test.ts 5/5 pass + cosmos.test.ts regression green; core-config dist built
- pass order: claude fund-safety → codex gpt-5.4 adversarial

## should-fix
### 1. channel validation is format-only, not authoritative — a wrong-but-well-formed channel still signs — `packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts:92`
`/^channel-\d+$/` catches undefined/empty/malformed channels (the COSMOS-03 bug), but `channel-5` and `channel-141` both pass even if one routes to the wrong destination chain — funds sent into a coherent-but-wrong channel land at an address that may not exist on the actual dest, i.e. stuck/lost. The repo ALREADY has an authoritative `(fromChain/channel) -> destChain` registry (`IBC_CHANNEL_DEST`, live-LCD-verified) in `packages/sdk/src/tools/prep/ibcTransfer.ts`. This resolver is the last fund-safety chokepoint before signing and could cross-check `${chainId}/${channel}` (and the dest HRP of `toAddress`) against that table.
action: Keep the format regex, but additionally ground `channel` against `IBC_CHANNEL_DEST` for the source chain-id (and optionally cross-check the `toAddress` bech32 HRP vs the resolved dest chain, as `prepareIbcTransfer` already does). This turns 'well-formed' into 'correct route'. Non-blocking because the normal flow's channel is produced by trusted upstream builders, but this is where a wrong channel would be caught cheaply.

### 2. guard only covers the native IBC_TRANSFER branch — signAmino/signDirect MsgTransfers bypass it — `packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts:84`
The `signDirect` and `signAmino` arms of `ibcEnabled` early-return before the `TransactionType.IBC_TRANSFER` branch, so a MsgTransfer hand-built via `buildSignAminoKeysignPayload` / `buildSignDirectKeysignPayload` (packages/sdk/src/vault/services/cosmos/buildCosmosPayload.ts) with a zero timeout or wrong channel is signed unchanged — the new guard never runs on it. Confirmed this is a SEPARATE construction path (caller supplies full msgs; timeout/channel grounding lives upstream in `prepareIbcTransfer`), and the specific `|| '0'` fallback bug existed ONLY in the IBC_TRANSFER branch that this PR fixes — so this is NOT a regression and NOT the reported bug. Still worth a note so the guard's scope is explicit.
action: Either document (comment) that timeout/channel validation for the signAmino/signDirect paths is the caller's responsibility (via `prepareIbcTransfer`), or add a follow-up that decodes the MsgTransfer out of the amino msgs / signDirect bodyBytes and applies the same timeout+channel checks as defense-in-depth. Defer is acceptable.

## suggestion
### 1. Long.fromString is a loose timeout validator — accepts garbage-nonzero and past timestamps — `packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts:98`
Verified: `Long.fromString('123abc')` => `123` (isZero=false, passes the guard) and `Long.fromString('abc')` => `0` (isZero=true, refused). So a malformed `latestBlock` second segment either fail-closes (good) or ships a tiny/past timestamp — but a past timeout expires immediately => the packet times out and REFUNDS, which is the SAFE direction (not the stuck-funds case this PR targets). The authoritative producer (`getCosmosAccountInfo`: `${height}_${blockTime + 10min}` in ns) always emits a clean future ns value, so garbage only arises from a malformed payload. Not fund-loss, hence suggestion.
action: For hardening, mirror `prepareIbcTransfer`'s stricter timestamp check: require the segment to be all-digits and plausibly-future (nanosecond era) before accepting, rather than relying on `!isZero()`.

### 2. `&& heightDisabled` is always-true here — timeoutHeight is hardcoded {0,0} — `packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts:112`
`timeoutHeight` is unconditionally `{revisionNumber: 0, revisionHeight: 0}` (lines 99-102), so `heightDisabled` is always `true` and the no-timeout guard reduces to `if (timeoutTimestamp.isZero())`. Correct and fail-closed, just slightly self-obscuring — reads as if a non-zero timeoutHeight could ever satisfy the timeout, which it can't on this path.
action: Optional: drop the `&& heightDisabled` (or add a comment that timeoutHeight is intentionally always disabled on this branch and timeout is timestamp-only). Purely cosmetic.

## q
- Confirm the app/SDK IBC-transfer feature always routes through the `IBC_TRANSFER` branch (transactionType=3, no signData) — not `buildCosmosPayload` (signAmino/signDirect). I traced the native path (chainSpecific cosmos resolver sets transactionType + ibcDenomTraces, no signData) as the guarded one; just confirming there's no product surface that builds an IBC transfer via the amino/direct path where this guard wouldn't apply.

## risk
Fund-safety fix for the Cosmos IBC `MsgTransfer` signing-input resolver. The identified stuck-funds bug (all-zero timeout when `ibcDenomTraces` is missing) is now refused fail-closed BEFORE the signable message is built, and the source channel is format-validated. The legit path (denom trace present + `channel-<n>` memo) still signs — verified by the regression test. Residual risk is bounded to hardening gaps (format-only vs authoritative channel grounding; the signAmino/signDirect construction paths are out of this guard's scope but have their own upstream grounding in `prepareIbcTransfer`). No fund-loss path introduced; strictly an improvement over the prior no-guard state.

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` adversarial pass. SDK vitest — worktree @4545a980: index.ibcTransfer.test.ts 5/5 pass + cosmos.test.ts regression green; core-config dist built.

PR #1084, opened 2026-07-08
fix(swap): validate cowswap quote response against the request before signing (AGG-01)

### LINKED ISSUE #1050: [audit-AGG-01] CowSwap order tokens/kind are taken from the API response, not validated against the request

gomesalexandre:
closes #1050

## tl;dr
The CowSwap order is **EIP-712-signed AND POSTed** with `sellToken`/`buyToken`/`kind`/`partiallyFillable` copied straight from the `/quote` API response (`getCowSwapQuote.ts`). A compromised, buggy, or MITM'd `apiBase` could **substitute a token** (funds swap into an attacker-chosen asset) or **flip `kind` sell↔buy** — the signed order carries the tampered values.

## Fix
`assertCowSwapQuoteMatchesRequest(quote, req)` asserts the response echoes the **requested** `sellToken`/`buyToken`/`kind`/`partiallyFillable` (the exact values POSTed in `body`) and **throws before the order is built**. `receiver` is excluded — the order already uses the caller's own receiver, not the response's, so a substituted receiver can't reach the signed order. Extracted to a config-free module (types-only import) so it's unit-testable in isolation.

## Runtime (SDK-CLI vitest, 6/6 pass)
Matching response passes (case-insensitive on addresses); substituted `sellToken`/`buyToken`, flipped `kind`, and flipped `partiallyFillable` each **throw**; a multi-field mismatch reports all. `tsc --noEmit` clean.

Changeset: `.changeset/cowswap-quote-response-validation-agg01.md` (`@vultisig/sdk` patch). Audit finding **AGG-01** (round 2). Codex degraded → manual review.

### DIFF
--- a/packages/core/chain/swap/general/cowswap/api/getCowSwapQuote.ts
+++ b/packages/core/chain/swap/general/cowswap/api/getCowSwapQuote.ts
@@ -9,6 +9,7 @@ import {
 } from '../config'
 import { buildCowSwapAppData, keccak256Hex } from '../sign/buildCowSwapOrder'
 import { CowSwapQuoteApiResponse } from '../types'
+import { assertCowSwapQuoteMatchesRequest } from './validateCowSwapQuoteResponse'
 
 type CowSwapQuoteRequest = {
   sellToken: string
@@ -65,6 +66,15 @@ export async function getCowSwapQuote({
   })
 
   const { quote } = response
+
+  // AGG-01: validate the response against the request (using the exact values we POSTed in `body`) before the
+  // order is signed/submitted. sellToken/buyToken come from the caller; kind/partiallyFillable from `body`.
+  assertCowSwapQuoteMatchesRequest(quote, {
+    sellToken,
+    buyToken,
+    kind: body.kind,
+    partiallyFillable: body.partiallyFillable,
+  })
 
   const permitRequired = KNOWN_PERMIT_TOKENS[chainConfig.chainId]?.some(
     addr => addr.toLowerCase() === sellToken.toLowerCase()
--- a/.changeset/cowswap-quote-response-validation-agg01.md
+++ b/.changeset/cowswap-quote-response-validation-agg01.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(swap): validate the CowSwap quote response against the request before signing (AGG-01)
+
+The CowSwap order is EIP-712-signed and POSTed with `sellToken`/`buyToken`/`kind`/`partiallyFillable` copied straight from the `/quote` API response. A compromised, buggy, or MITM'd `apiBase` could substitute a token (funds swap into an attacker-chosen asset) or flip `kind` sell↔buy. `assertCowSwapQuoteMatchesRequest` now asserts the response echoes the requested values on those fund-critical fields and throws on any mismatch before the order is built. `receiver` is unaffected — the order already uses the caller's own receiver, not the response's.
--- a/packages/core/chain/package.json
+++ b/packages/core/chain/package.json
@@ -1492,6 +1492,11 @@
       "import": "./dist/swap/general/cowswap/api/submitCowSwapOrder.js",
       "default": "./dist/swap/general/cowswap/api/submitCowSwapOrder.js"
     },
+    "./swap/general/cowswap/api/validateCowSwapQuoteResponse": {
+      "types": "./dist/swap/general/cowswap/api/validateCowSwapQuoteResponse.d.ts",
+      "import": "./dist/swap/general/cowswap/api/validateCowSwapQuoteResponse.js",
+      "default": "./dist/swap/general/cowswap/api/validateCowSwapQuoteResponse.js"
+    },
     "./swap/general/cowswap/config": {
       "types": "./dist/swap/general/cowswap/config.d.ts",
       "import": "./dist/swap/general/cowswap/config.js",
--- a/packages/core/chain/swap/general/cowswap/api/__tests__/getCowSwapQuote.test.ts
+++ b/packages/core/chain/swap/general/cowswap/api/__tests__/getCowSwapQuote.test.ts
@@ -15,11 +15,19 @@ vi.mock('@vultisig/lib-utils/query/queryUrl', () => ({
   queryUrl: vi.fn(),
 }))
 
-function makeQuoteResponse(chainId: number, buyAmount = '990000000000000000') {
+// AGG-01: the CoW /quote API echoes the requested sellToken/buyToken; getCowSwapQuote now asserts that
+// before signing. Mock the same — default to WETH->USDC, but let a test that requests a different sellToken
+// (the permit-token cases) have the mock echo it, so the fixture reflects real API behaviour.
+function makeQuoteResponse(
+  chainId: number,
+  buyAmount = '990000000000000000',
+  sellToken = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2',
+  buyToken = '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
+) {
   return {
     quote: {
-      sellToken: '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2',
-      buyToken: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48',
+      sellToken,
+      buyToken,
       receiver: '0xreceiver',
       sellAmount: '1000000000000000000',
       buyAmount,
@@ -160,7 +168,7 @@ describe('getCowSwapQuote', () => {
 
   it('sets permitRequired=true for USDC on Ethereum (known permit token)', async () => {
     const usdcEthereum = KNOWN_PERMIT_TOKENS[1][0]
-    vi.mocked(queryUrl).mockResolvedValueOnce(makeQuoteResponse(1))
+    vi.mocked(queryUrl).mockResolvedValueOnce(makeQuoteResponse(1, undefined, usdcEthereum))
 
     const quote = await getCowSwapQuote({
       ...baseInput,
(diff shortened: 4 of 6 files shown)


gomesalexandre (commented):
## verdict: SHIP-WITH-NITS (draft, comment-only)

Small focused guard, checked the make-or-break question first (false-block risk) before anything else.

### 1. false-block risk — checked, clean

The whole review hinges on whether CoW's `/quote` can ever *legitimately* echo back a `sellToken`/`buyToken` that differs from what was requested. Traced the only caller:

`packages/core/chain/swap/quote/findSwapQuote.ts:642-666` gates the CowSwap fetcher to:
```ts
if (
  fromChain === toChain &&
  isOneOf(fromChain, cowSwapSupportedChains) &&
  from.id !== undefined &&
  to.id !== undefined
)
```
`from.id`/`to.id` being `undefined` is how this codebase represents the native-coin sentinel (see the comment right above it: *"Because of that ERC-20 gate, a native-ETH (`from.id === undefined`) pair never reaches this fetcher"*). So native ETH is excluded **before** `getCowSwapQuote` is ever called — `sellToken`/`buyToken` are always already ERC-20 addresses (`from.id`/`to.id`), never the `0xEeee…` sentinel. No WETH-wrap/native-normalization path exists for this guard to false-block on.

- Casing: guard lowercases both sides (`quote.sellToken.toLowerCase() !== req.sellToken.toLowerCase()`) so EIP-55 checksum differences are a non-issue — confirmed.
- Canonicalization: for a same-chain ERC-20→ERC-20 pair, CoW's `/quote` has no reason to substitute the token identity (no wrap/unwrap involved, no cross-chain routing) — it's a straight echo. Nothing in the SDK's calling convention gives CoW room to legitimately diverge on these 4 fields.

praise: this is exactly the kind of "did you actually check the call site" leg-work that keeps a fund-safety guard from turning into a false-block outage — good on tracing `from.id !== undefined` instead of assuming.

### 2. guard correctness — confirmed

- `assertCowSwapQuoteMatchesRequest` is called at `getCowSwapQuote.ts:72-77`, right after `const { quote } = response` and *before* the `cowswap_order` tx object is built (lines 93-116) and returned. A throw here means the order is never constructed, so it's a true fail-closed gate on the signed path — verified the built object is returned unmutated through `mpc/keysign/swap/build.ts` straight into the `KeysignPayload.swapPayload`, no bypass.
- `kind`/`partiallyFillable` are compared against the exact POSTed `body.kind` (`'sell'`, hardcoded at line 56) and `body.partiallyFillable` (`false`, hardcoded at line 58) — not re-derived or hardcoded separately in the assert call. Good, single source of truth.

### 3. scope — confirmed

- All 4 fields from the audit finding (AGG-01 names `sellToken`/`buyToken`/`kind`) are covered, plus `partiallyFillable` was proactively added (issue #1050 body doesn't explicitly name it but it's equally fund-critical — flipping it to `true` on an otherwise-sell order changes execution semantics). Reasonable scope extension, not scope creep.
- `receiver` exclusion is correct: `getCowSwapQuote.ts:100` builds the order with the function's own `receiver` param, never `quote.receiver`. Grepped — no other read of `quote.receiver` anywhere in the built order. Comment in the new module accurately describes this.

### 4. tests

```
✓ packages/core/chain/swap/general/cowswap/api/validateCowSwapQuoteResponse.test.ts (6 tests) 2ms
Test Files  1 passed (1)
Tests  6 passed (6)
```
6/6 confirmed locally. Changeset present (`@vultisig/sdk` patch) and correctly scoped.

### nits (non-blocking)

- suggestion: runtime evidence in the PR body is vitest-level only (stubbed response), not a captured real CoW `/quote` payload. For a pure config-free guard with no UI surface this matches what issue #1050 itself asked for ("SDK-CLI: unit-drive"), so not gating — but if you've got a spare minute, one real quote response pasted into the PR body as a sanity artifact would remove all doubt for future readers.
- q: if the API response is malformed (`quote.sellToken` non-string/undefined), `quote.sellToken.toLowerCase()` throws a raw `TypeError` instead of the clean assertion message — still fails closed (order never builds) so not a fund-safety gap, just a slightly uglier error for whoever's debugging it. Not blocking.

`closes #1050` present in body. No inline file comments needed — diff is small enough to review in full above.

NeOMakinG (commented):
### Review — sdk#1084 CowSwap quote-response validation (AGG-01)

Two-lane review (Claude load-bearing + Codex gpt-5.4 adversarial, independent). Both lanes agree on the two findings below.

## testing approach
Traced the trust boundary in `getCowSwapQuote.ts`: the new `assertCowSwapQuoteMatchesRequest(quote, …)` compares the untrusted `/quote` **response** (`response.quote`) against **request-side** state — `sellToken`/`buyToken` are the caller-supplied fn params captured before the HTTP call, and `kind`/`partiallyFillable` are read from the locally-constructed `body` (hardcoded `'sell'`/`false`). So the check is **genuinely request-vs-response, not tautological** ✅. `receiver` in the built order (`:100`) is the caller's own local, never `quote.receiver` ✅. On mismatch it `throw`s in `getCowSwapQuote` **before** the `cowswap_order` is built → nothing gets EIP-712-signed → **fail-closed** ✅. Cross-referenced against the sibling PR #1082 diff to assess coupling. No node_modules in the worktree, so the vitest suite was read but not executed locally; CI Integration + CodeRabbit green, Unit/Lint in progress at review time.

## preferably-blocking
**1. Incomplete AGG-01 relative to its sibling: the signed `sellAmount` is still fully response-controlled (over-sell path live).** After the new guard passes, `grossSellAmount = quote.sellAmount + quote.feeAmount` (`getCowSwapQuote.ts:91`) flows straight into the signed order's `sellAmount` (`:101`) with **no check against the requested `sellAmount`**. A compromised / buggy / MITM'd `apiBase` can keep `sellToken`/`buyToken`/`kind`/`partiallyFillable` honest (passing this guard) yet **inflate `quote.sellAmount` and/or `quote.feeAmount`**, so the user signs authorization to sell **more than requested** — the most direct fund-safety consequence of any field in this struct. `validTo` (`:103`) is likewise unbounded here. Both are exactly what sibling **#1082** guards (`grossSellAmount === sellAmount.toString()` + a `validTo` ceiling) and what #1084 omits.

**2. #1084 is a strict subset of #1082 at the same insertion point → merge hazard + misleading "AGG-01 fixed" state.** Both PRs insert their validation immediately after `const { quote } = response` and both check the same 4 categorical fields; #1082 additionally checks the over-sell amount and `validTo`. So #1084 adds nothing #1082 doesn't already have, while (a) guaranteeing a textual **merge conflict** with #1082, and (b) creating a trap: if #1084 lands **alone**, the repo looks like AGG-01 is closed while the over-sell path in finding #1 is still live; and a hand-resolved conflict can easily keep #1084's narrower helper and silently drop #1082's amount/expiry assertions.

**Please clarify the intended relationship between #1082 and #1084 before either merges.** Reasonable resolutions: (a) land #1082 (strictly more complete, already reviewed) and close #1084; or (b) if the extracted config-free module in #1084 is the preferred shape, **extend `assertCowSwapQuoteMatchesRequest` to also assert `grossSellAmount === requested sellAmount` and bound `validTo`**, and re-scope #1082 to build on it — so no path merges an AGG-01 fix that omits over-sell protection.

## risk
Fund-critical (WYSIWYS on a signed EIP-712 GPv2 order). Change itself is additive and fail-closed — **strictly better than main** (which has zero validation). Not approving only because, as scoped, it can be mistaken for the complete AGG-01 fix while leaving the over-sell path unguarded, and it conflicts with #1082. No regression risk to the happy path (case-insensitive token compare, exact `kind`/`partiallyFillable` literals; type-safe against `CowSwapQuoteObject`).

## verdict
COMMENT — not blocking the author's multi-PR workflow, but do not merge #1084 as the standalone AGG-01 fix. Reconcile with #1082 per above.

*Reviewed by Claude (load-bearing) + Codex gpt-5.4 (adversarial second opinion). Both lanes independently reached findings 1 and 2.*

packages/core/chain/swap/general/cowswap/api/getCowSwapQuote.ts
```diff
@@ -66,6 +67,15 @@ export async function getCowSwapQuote({
 
   const { quote } = response
 
+  // AGG-01: validate the response against the request (using the exact values we POSTed in `body`) before the
+  // order is signed/submitted. sellToken/buyToken come from the caller; kind/partiallyFillable from `body`.
+  assertCowSwapQuoteMatchesRequest(quote, {
```
NeOMakinG, line 72: `preferably-blocking:` This guard is genuinely request-vs-response (not tautological) and fail-closed — good. But it stops at 4 categorical fields: `grossSellAmount = quote.sellAmount + quote.feeAmount` (:91) then flows unchecked into the signed order's `sellAmount` (:101). An apiBase that keeps tokens/kind honest but inflates `sellAmount`/`feeAmount` still gets a signed **over-sell**. Sibling #1082 guards this (`grossSellAmount === sellAmount`) + bounds `validTo`; #1084 omits both and is a strict subset of #1082 at this same insertion point. Please reconcile the two PRs before merge — see review body.

PR #1086, opened 2026-07-08
fix(ton): fail closed on an rpc error instead of defaulting to uninit (ton-01)

gomesalexandre:
## Context (fund-safety audit, TON-01, High)

Companion fix for vultisig/vultiagent-app#1733 / vultisig/vultiagent-app#1756. The app-side TON send path calls `chains.ton.getTonWalletInfo`; before this PR, that SDK primitive could collapse RPC failures into `{seqno: 0, status: 'uninit'}`, making a transient failure indistinguishable from a genuinely uninitialized wallet.

## Impact
A transient network/RPC failure on the recipient lookup could silently produce `bounceable: false`, stripping the refund-on-failure net for what may be a contract recipient. A failed sender lookup could also look like `seqno: 0`, defeating stale-seqno protection for prebuilt/yield-style signing flows.

## Fix
`getTonWalletInfo` now fails closed on:
- transport/fetch errors
- non-OK HTTP responses
- body-level toncenter failure shapes such as `{ ok:false, result:null }`
- body-level v3 `addressInformation` errors

A genuinely uninitialized wallet still resolves normally when toncenter returns a successful body with no `account_state` and `status: 'uninit'`.

Reviewer follow-ups handled in `d7af3cd1`:
- body-level `{ok:false,result:null}` no longer collapses to seqno 0/uninit
- error messages no longer interpolate the raw queried TON address or nested toncenter error text
- added a patch changeset so this can publish and the app can pin it

## Receipts

```
yarn install --immutable
# completed with existing peer-dependency warnings only
```

```
yarn workspace @vultisig/sdk exec vitest run --config tests/unit/vitest.config.ts tests/unit/tonRpc.test.ts
PASS tests/unit/tonRpc.test.ts

Test Files  1 passed (1)
Tests       7 passed (7)
```

```
yarn build:shared
Shared packages synced to packages/*/dist
```

```
yarn workspace @vultisig/sdk typecheck
# pass
```

```
yarn node scripts/run-eslint.mjs --config .config/eslint.config.mjs packages/sdk/src/chains/ton/rpc.ts packages/sdk/tests/unit/tonRpc.test.ts
# pass
```

```
yarn prettier --config .config/.prettierrc --check packages/sdk/src/chains/ton/rpc.ts packages/sdk/tests/unit/tonRpc.test.ts .changeset/ton-wallet-info-fail-closed.md
```

```
git diff --check
# clean
```

```
codex exec review --uncommitted --dangerously-bypass-approvals-and-sandbox
# no discrete introduced bug found
```

Note: `yarn workspace @vultisig/sdk test:unit` in a fresh clone ran the whole unit suite and hit missing shared `dist` artifacts before `yarn build:shared`; the focused TON unit test above is the relevant receipt and passes. After `build:shared`, SDK typecheck passes.

Release note: landing this still requires publishing the SDK patch and bumping `vultiagent-app` to the published version before the production app path is fully active.

closes vultisig/vultiagent-app#1733

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * TON wallet information checks now distinguish genuine uninitialized wallets from lookup, network, and RPC failures.
  * Failed requests and invalid RPC responses now report errors instead of incorrectly appearing as uninitialized wallets.
  * Wallet balances, sequence numbers, and statuses are handled more accurately across supported response formats.

* **Tests**
  * Added coverage for active and uninitialized wallets, successful balance retrieval, and HTTP, RPC, and network failures.


packages/sdk/src/chains/ton/rpc.ts
```diff
@@ -68,30 +68,41 @@ export async function getTonBalance(address: string, gatewayUrl: string): Promis
...
-    const ext = (await extRes.json().catch(() => ({}))) as ToncenterExtendedInfo
-    seqno = ext.result?.account_state?.seqno ?? 0
-    if (ext.result?.balance !== undefined) balance = toBigInt(ext.result.balance)
+  if (!extRes.ok) {
+    throw new Error(
+      `toncenter getExtendedAddressInformation failed for ${address}: ${extRes.status} ${extRes.statusText}`
+    )
   }
+  const ext = (await extRes.json()) as ToncenterExtendedInfo
+  const seqno = ext.result?.account_state?.seqno ?? 0
```
NeOMakinG, line 103: `preferably-blocking:` fail-closed is incomplete: only HTTP `.ok` is checked, not the toncenter body-level `{ok:false, result:null}` shape — the exact transient-failure shape #964 hardened `getTonAccountInfo` against, and that this PR's own docstring claims to mirror — The docstring (line 83) says this now "mirrors `getTonAccountInfo`'s `ok`/`result` check" — but it does not. `getTonAccountInfo` (packages/core/chain/chains/ton/account/getTonAccountInfo.ts:48) guards `if (!response?.ok || !response.result) throw`, and its comment (from #964) is explicit that this same `/ton/v2/getExtendedAddressInformation` endpoint returns "a transient RPC failure ... as a 200 with a null `result`" — i.e. HTTP 200 + `{ok:false, result:null}`. The new `getTonWalletInfo` only checks `extRes.ok` (HTTP status), then does `ext.result?.account_state?.seqno ?? 0` (line 97) and `v3.status ?? 'uninit'` (line 104). So on that documented 200-with-null-result shape it does NOT throw: seqno collapses to 0 (stale-seqno guard defeated again) and status collapses to 'uninit' (bounce-safety flag stripped) — the precise bug this PR set out to close, still open for that failure mode.

fix: add a body-level guard mirroring the sibling primitive: after parsing, `if (!ext.ok || !ext.result) throw` (v2 carries `ok`), and confirm the v3 side likewise can't return 200-with-no-status silently (default-to-'uninit' at line 104 has the same exposure). Add a test for the HTTP-200 + `{ok:false, result:null}` shape.

packages/sdk/tests/unit/tonRpc.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    })
+
+    await expect(getTonWalletInfo('EQactive', 'https://gw.test')).resolves.toEqual({
+      seqno: 12,
+      balance: 9n,
+      status: 'active',
+    })
+  })
+
+  it('throws (does not default to uninit/seqno 0) on a non-OK getExtendedAddressInformation response', async () => {
```
NeOMakinG, line 69: `should-fix:` no cross-repo regression test that the send flow aborts before build/broadcast when `getTonWalletInfo` rejects (contract-shape change: resolve-always → may-reject) — This PR changes `getTonWalletInfo` from a function that always resolves to one that can reject. The new tests only assert the promise rejects; the only app-side TON native-send test is the happy path. Nothing asserts that a rejection here stops the flow BEFORE `buildTonSendTx`/`broadcastTonTx` — so the fail-closed contract can regress silently if a caller ever re-wraps this in a `.catch(() => default)`.

fix: add a send-flow test (here or in the companion va#1756) asserting no tx is built/broadcast when `getTonWalletInfo(from)` rejects.

packages/sdk/src/chains/ton/rpc.ts
```diff
@@ -68,30 +68,41 @@ export async function getTonBalance(address: string, gatewayUrl: string): Promis
...
 
-  let seqno = 0
-  let balance = 0n
-  if (extRes && extRes.ok) {
-    const ext = (await extRes.json().catch(() => ({}))) as ToncenterExtendedInfo
-    seqno = ext.result?.account_state?.seqno ?? 0
-    if (ext.result?.balance !== undefined) balance = toBigInt(ext.result.balance)
+  if (!extRes.ok) {
+    throw new Error(
+      `toncenter getExtendedAddressInformation failed for ${address}: ${extRes.status} ${extRes.statusText}`
```
NeOMakinG: `suggestion:` raw TON address is embedded in the thrown error string, which propagates to the app's failed-card UI and analytics telemetry (Codex lane) — Both throws (lines 92-93 and 101) interpolate `${address}`. In the app this message is forwarded via `SignAndBroadcastError` and reaches analytics via `clampErrorMessage()`, which does not scrub addresses. The address is public on-chain data so this is low-severity, but it adds an avoidable linkability signal to telemetry.

fix: drop the raw address from the message (or redact to a prefix) — the endpoint name + status already identify the failure.

NeOMakinG (commented):

## preferably-blocking
### 1. fail-closed is incomplete: only HTTP `.ok` is checked, not the toncenter body-level `{ok:false, result:null}` shape — the exact transient-failure shape #964 hardened `getTonAccountInfo` against, and that this PR's own docstring claims to mirror — `packages/sdk/src/chains/ton/rpc.ts:97`
The docstring (line 83) says this now "mirrors `getTonAccountInfo`'s `ok`/`result` check" — but it does not. `getTonAccountInfo` (packages/core/chain/chains/ton/account/getTonAccountInfo.ts:48) guards `if (!response?.ok || !response.result) throw`, and its comment (from #964) is explicit that this same `/ton/v2/getExtendedAddressInformation` endpoint returns "a transient RPC failure ... as a 200 with a null `result`" — i.e. HTTP 200 + `{ok:false, result:null}`. The new `getTonWalletInfo` only checks `extRes.ok` (HTTP status), then does `ext.result?.account_state?.seqno ?? 0` (line 97) and `v3.status ?? 'uninit'` (line 104). So on that documented 200-with-null-result shape it does NOT throw: seqno collapses to 0 (stale-seqno guard defeated again) and status collapses to 'uninit' (bounce-safety flag stripped) — the precise bug this PR set out to close, still open for that failure mode.
action: add a body-level guard mirroring the sibling primitive: after parsing, `if (!ext.ok || !ext.result) throw` (v2 carries `ok`), and confirm the v3 side likewise can't return 200-with-no-status silently (default-to-'uninit' at line 104 has the same exposure). Add a test for the HTTP-200 + `{ok:false, result:null}` shape.

## should-fix
### 1. no changeset — the PR's own release note depends on a version bump the app must pin, but no `.changeset/*.md` is included — `packages/sdk/src/chains/ton/rpc.ts`
`git diff origin/main...HEAD` shows only `rpc.ts` + the test — no `.changeset/` entry, though the repo uses changesets (`.changeset/config.json`). The release note states "Landing this alone has no effect on `vultiagent-app` until the SDK is version-bumped and the app pins the new version." Without a changeset the version never bumps on merge, so the app can never pin the fix and the companion #1733/#1756 gap stays open indefinitely.
action: add a changeset (`yarn changeset`) with an appropriate semver bump before marking ready.

### 2. no cross-repo regression test that the send flow aborts before build/broadcast when `getTonWalletInfo` rejects (contract-shape change: resolve-always → may-reject) — `packages/sdk/tests/unit/tonRpc.test.ts:69`
This PR changes `getTonWalletInfo` from a function that always resolves to one that can reject. The new tests only assert the promise rejects; the only app-side TON native-send test is the happy path. Nothing asserts that a rejection here stops the flow BEFORE `buildTonSendTx`/`broadcastTonTx` — so the fail-closed contract can regress silently if a caller ever re-wraps this in a `.catch(() => default)`.
action: add a send-flow test (here or in the companion va#1756) asserting no tx is built/broadcast when `getTonWalletInfo(from)` rejects.

## suggestion
### 1. raw TON address is embedded in the thrown error string, which propagates to the app's failed-card UI and analytics telemetry (Codex lane) — `packages/sdk/src/chains/ton/rpc.ts:93`
Both throws (lines 92-93 and 101) interpolate `${address}`. In the app this message is forwarded via `SignAndBroadcastError` and reaches analytics via `clampErrorMessage()`, which does not scrub addresses. The address is public on-chain data so this is low-severity, but it adds an avoidable linkability signal to telemetry.
action: drop the raw address from the message (or redact to a prefix) — the endpoint name + status already identify the failure.

## risk
TON account-state fetch on the SDK send path. Core fix is correct and directional: the pre-fix code wrapped both fetches in `.catch(() => null)` and defaulted `{seqno:0, status:'uninit'}` on ANY failure (fetch throw, non-OK HTTP, json-parse error) — indistinguishable from a genuinely uninitialized wallet. The new code removes the `.catch()` (fetch-throw propagates), throws on non-OK HTTP, and drops the `.catch(() => ({}))` on `.json()` (parse errors now propagate) — all three common network-blip paths now fail closed. Genuine-uninit path is preserved (200 OK + no `account_state`/`status:'uninit'` → seqno 0), so no over-block; the active-wallet test confirms real seqno/status flow through to the signable tx. Fund-safety: a wrong seqno=0 for an active wallet produces an on-chain-rejected tx (seqno mismatch → no replay/double-spend, fee/stuck only); the real loss vector is the recipient bounce flag collapsing to non-bounceable (funds not refunded if the recipient contract reverts). Fail-closed is unambiguously the right call. Residual gap (preferably-blocking): the documented 200-with-null-result transient shape still slips through, re-opening both vectors for that one failure mode. CI green (unit + integration + lint). Draft → COMMENT only, no approval.

## verdict
COMMENTED

packages/sdk/src/chains/ton/rpc.ts
gomesalexandre: tackled in `d7af3cd1`: `getTonWalletInfo` now checks the body-level v2 `{ ok:false, result:null }` shape and throws instead of collapsing to seqno 0/uninit. Added a regression test for that exact 200/null-result response.

packages/sdk/tests/unit/tonRpc.test.ts
gomesalexandre: covered by the companion app PR now: vultisig/vultiagent-app#1756 asserts TON wallet-info rejection aborts before `buildTonSendTx`/prebuilt build, and that no raw TON address reaches the thrown message. SDK-side focused tests here now pin the rejecting primitive contract.

packages/sdk/src/chains/ton/rpc.ts
gomesalexandre: tackled in `d7af3cd1`: SDK wallet-info errors no longer include the raw queried TON address, and tests assert hostile nested toncenter errors containing `EQflaky` are not surfaced.

packages/sdk/src/chains/ton/rpc.ts
```diff
@@ -68,30 +72,46 @@ export async function getTonBalance(address: string, gatewayUrl: string): Promis
...
-    if (v3.status) status = v3.status
-    if (v3.balance !== undefined && balance === 0n) balance = toBigInt(v3.balance)
+  if (!v3Res.ok) {
+    throw new Error(`toncenter addressInformation failed: ${v3Res.status} ${v3Res.statusText}`)
+  }
+  const v3 = (await v3Res.json()) as ToncenterV3AddressInfo
+  if (v3.ok === false) {
+    throw new Error('toncenter addressInformation returned an error')
   }
+  const status: TonWalletStatus = v3.status ?? 'uninit'
```
NeOMakinG, line 113: `should-fix:` v3 fail-open twin of the seqno bug you just fixed. This throws only on explicit `v3.ok === false` (line 110), then defaults `status ?? 'uninit'` here. A 200 response with `ok` absent and `status` null/missing (the `error?: string` field you added anticipates a body-level error envelope) silently becomes `'uninit'` → downstream `bounceable = status !== 'uninit'` mis-classifies a live recipient as uninit and strips the bounce/refund net = the recipient half of TON-01. Narrower than the v2 crux (v3 errors are mostly HTTP-status-coded, caught by `!v3Res.ok`), so non-blocking — but cheap to close: `if (v3.ok === false || v3.error != null || !v3.status) throw` + a `200 + missing status` test mirroring the v2 case.

NeOMakinG (approved):
Round 2 re-review at `d7af3cd1`. My r1 preferably-blocking finding (the body-level `{ok:false, result:null}` 200 fail-open on the seqno path) is **resolved**, the changeset is present, and the fix is rebased on top of latest `main`. Verified the full suite locally (`vitest tests/unit/tonRpc.test.ts` → 7/7 pass) covering all four transient shapes on the v2/seqno path plus the genuine-uninit and active resolves. One convergent (Claude + Codex) should-fix remains on the *v3/status* recipient-bounce path — non-blocking. Approving.

## testing approach
- Worktree at PR head `d7af3cd1`; `git rev-list --count d7af3cd1..origin/main` = 0 and merge-base == `origin/main` tip → PR contains latest main (fresh, no rebase needed).
- Built shared packages, ran `vitest run tests/unit/tonRpc.test.ts` → **7 passed**: genuine-uninit resolves `{seqno:0,uninit}`, active resolves `{seqno:12,active}`, non-OK v2, HTTP-200 `{ok:false,result:null}` v2 (the crux), non-OK v3, HTTP-200 `{ok:false}` v3, and network fetch-throw all reject.
- Diffed `fd61e619..d7af3cd1`: the only rpc.ts change since r1 is exactly the body-level guard + the removal of the raw `${address}` from throw strings.
- Two lanes: Claude (load-bearing) + Codex `gpt-5.4` (pinned `d7af3cd1` via `git show`, referenced PR #1086 / the uninit 200-null shape — not stale, not cross-contaminated). Both independently landed on the same v3/status residual below.

## prior issues — status
| 1 | 1 | preferably-blocking | v2 `getExtendedAddressInformation` only checked HTTP `.ok`, not body-level `{ok:false, result:null}` (HTTP-200) → still collapsed to seqno 0 / uninit on the documented transient-as-200 shape | ✅ fixed | `d7af3cd1`: `if (ext.ok === false || !ext.result) throw` at rpc.ts:100. Test `throws on HTTP-200 ... with ok:false/null result` passes. Genuine-uninit (result present, no `account_state`) still resolves seqno 0 — no over-reject. |
| 1 | 2 | should-fix | no `@vultisig/sdk` changeset → fix would not ship / app could not pin | ✅ fixed | `.changeset/ton-wallet-info-fail-closed.md` (`'@vultisig/sdk': patch`). |
| 1 | 3 | should-fix (Codex) | no cross-repo regression test that the send flow aborts before build/broadcast when `getTonWalletInfo` rejects | ℹ️ app-side | Belongs to va#1756 (`signingFlows.test.ts` asserts `.rejects.toThrow(TON_*_WALLET_INFO_FAILED)` + `buildTonSendTx not called`). Out of scope for the SDK primitive; SDK unit coverage is complete. |
| 1 | 4 | suggestion | throw strings interpolated raw `${address}` (TON base64url) → leaked to app analytics via `clampErrorMessage` (no TON scrub) | ✅ fixed | Addresses removed from all throw strings; tests assert `err.message).not.toContain('EQflaky')`. |

## should-fix
**v3 `status ?? 'uninit'` still fails open on a 200-with-missing-status (recipient-bounce path).** The v2/seqno path is now fully guarded, but the sibling v3 lane only throws on explicit `v3.ok === false` (rpc.ts:110) and then does `const status = v3.status ?? 'uninit'` (rpc.ts:113). A toncenter/proxy v3 response that returns HTTP 200 with `ok` absent **and** `status` null/missing (the author even added `error?: string` to `ToncenterV3AddressInfo`, anticipating a body-level error envelope) silently collapses to `'uninit'` — the exact fail-open class this PR closes for seqno. Downstream (`bounceable = recipientInfo.status !== 'uninit'`) that mis-classifies a live recipient contract as uninit → strips the bounce/refund-on-failure net, which is the recipient-side half of TON-01. It's narrower than the v2 crux (v3's dominant failure mode is HTTP-status-coded and is already caught by `!v3Res.ok`), so this is hardening, not a re-open — but it's cheap and symmetric: throw when `status` is absent (`if (v3.ok === false || v3.error != null || !v3.status) throw`) rather than defaulting, and add a `200 + missing status` test to mirror the v2 case. Non-blocking.

## risk
Low. Net a strict fund-safety improvement over `main`: the documented v2 200-null transient-as-uninit hole (wrong seqno=0 → on-chain-rejected tx; force-disabled recipient bounce flag) is closed and the genuine-uninit first-send path is preserved. Residual is a single narrower v3/status edge on an already-HTTP-guarded lane. No replay/double-spend vector. Closes va#1756's inherited gap for shapes (a)/(b)/(c)/(d-v2) — va#1756 can now bump its `@vultisig/sdk` pin once this ships; the v3/status residual, if left, is the one remaining recipient-bounce edge the app still can't discriminate.

## verdict
**Approve.** The load-bearing blocker (v2 body-level 200-null → seqno-0 fail-open) is fixed, all four transient shapes on the seqno path fail closed, genuine-uninit still resolves, changeset present, fresh on main, and the r1 address-leak suggestion was addressed too. One non-blocking should-fix (v3 `status ?? 'uninit'`) left for a cheap follow-up.

<sub>🤖 reviewer lane — Claude (load-bearing) + Codex gpt-5.4. Verified at HEAD d7af3cd1.</sub>

PR #647, opened 2026-06-05
fix(zcash): enable ZIP-317 fee planning

rcoderdev:
## Summary
- Enable WalletCore ZIP-317 fee planning for Zcash UTXO signing inputs.
- Bump `@trustwallet/wallet-core` to `^4.6.13` so the JS protobuf includes `zip_0317`.
- Add a regression test that decodes the bytes passed to `AnySigner.plan` and verifies Zcash sets `zip_0317`.

Reference: https://github.com/vultisig/vultisig-ios/pull/4488

## Verification
- `yarn build:all`
- `yarn lint`
- `yarn test`
- `yarn cli:build`
- Built CLI Zcash dry-run: isolated Fast Vault import, added Zcash, self-send preview `0.0001 ZEC` returned `success: true`.
- Windows local SDK tarball smoke passed: `.codex/receipts/windows-desktop-smoke/desktop-smoke-2026-06-05T13-21-17-266Z.json`.

## Summary by CodeRabbit

* **New Features**
  * Enabled ZIP-317 fee planning for Zcash transactions, improving fee estimation accuracy.

* **Tests**
  * Added tests validating ZIP-317 fee planning for Zcash UTXO signing.

* **Chores**
  * Bumped WalletCore dependency versions across packages and added a changeset.

PR #1089, opened 2026-07-08
fix(utxo): per-chain min-fee floor + canonical zcash zip-317 action count (UTXO-03/04)

### LINKED ISSUE #1063: [audit-UTXO-03 · UTXO-04] No minimum-fee floor for BTC/LTC/DOGE/BCH/Dash (only Zcash has one); Zcash's own ZIP-317 f
## Context (fund-safety audit round 2, 2026-07-08 · finding UTXO-03 · UTXO-04 · Medium)

UTXO-03: buildUtxoSendTx enforces a zip317Floor only for Zcash; every other chain trusts caller-supplied feeRate with no chain-aware minimum. Dogecoin's real min relay fee is ~100× Bitcoin's, so a BTC-reasonable rate can silently underpay DOGE below relay → stuck. UTXO-04: the builder's local zcashConventionalFee(inputCount) ignores output bytes, diverging from the SDK's canonical getZcashConventionalFee (max(inputActions, outputActions)) — under-counts ZIP-317 actions for large-memo sends → rejection.

## Effort / risk / confidence
Effort M / SRisk LOWConfidence MED (real exposure depends on upstream feeRate)

## Prompt for AI agents
Implement the fix for audit finding **UTXO-03 · UTXO-04** in `vultisig/vultisig-sdk`. UTXO-03: buildUtxoSendTx enforces a zip317Floor only for Zcash; every other chain trusts caller-supplied feeRate with no chain-aware minimum. Dogecoin's real min relay fee is ~100× Bitcoin's, so a BTC-reasonable rate can silently underpay DOGE below relay → stuck. UTXO-04: the builder's local zcashConventionalFee(inputCount) ignores output bytes, diverging from the SDK's canonical getZcashConventionalFee (max(inputActions, outputActions)) — under-counts ZIP-317 actions for large-memo sends → rejection.  Make the minimal, additive, fail-closed change at the anchors named above; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI: build the SDK and unit/CLI-drive the affected path with a fixture; assert before (bug) vs after (fixed).). **Grep open PRs in this repo first** — several audit findings already have in-flight PRs; advance/verify an existing one (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **UTXO-03 · UTXO-04** (round 2, Medium) · bead `agent-backend-ts-ervku`. Runtime before+after is the finish line, not unit-tests-green._

gomesalexandre:
closes #1063

## Context (fund-safety audit round 2 · UTXO-03 · UTXO-04 · Medium)

Two sub-findings in `buildUtxoSendTx` (`packages/sdk/src/chains/utxo/tx.ts`):

- **UTXO-03** — a min-fee floor was only enforced for Zcash. Every other UTXO chain trusted the caller-supplied `feeRate` with no chain-aware minimum. Dogecoin's real min-relay-fee is ~100x Bitcoin's, so a BTC-reasonable sat/vB silently underpays DOGE below relay -> stuck/non-relayable tx.
- **UTXO-04** — the local `zcashConventionalFee(inputCount)` only counted INPUT bytes, diverging from the SDK's canonical `getZcashConventionalFee` (ZIP-317 `max(inputActions, outputActions)`). A large OP_RETURN memo adds output bytes the local formula ignored entirely -> under-counted actions -> relay rejection risk.

## UTXO-03 fix — per-chain min-fee floor table

Sourced from each chain's own Core `DEFAULT_MIN_RELAY_TX_FEE` (expressed per kvB there, divided by 1000 for the per-byte value the builder uses):

| Chain | Core default (per kvB) | Floor (per vB) | Source |
|---|---|---|---|
| Bitcoin | 1,000 sat/kvB | **1** sat/vB | `bitcoin/bitcoin` `src/policy/policy.h` `DEFAULT_MIN_RELAY_TX_FEE` |
| Litecoin | 1,000 litoshi/kvB | **1** litoshi/vB | `litecoin-project/litecoin` `src/validation.h` `DEFAULT_MIN_RELAY_TX_FEE` (same order as BTC — LTC's `DUST_RELAY_TX_FEE` is ~10x BTC's, but its min *relay* fee is not) |
| Bitcoin-Cash | 1,000 sat/kvB | **1** sat/vB | `Bitcoin-ABC` / `bitcoin-cash-node` `src/policy/policy.h` `DEFAULT_MIN_RELAY_TX_FEE_PER_KB` |
| Dash | 1,000 duff/kvB | **1** duff/vB | `dashpay/dash` `src/policy/policy.h` `DEFAULT_MIN_RELAY_TX_FEE` |
| Dogecoin | 100,000 koinu/kvB | **100** koinu/vB | `dogecoin/dogecoin` `src/validation.h` (`DEFAULT_MIN_RELAY_TX_FEE = RECOMMENDED_MIN_TX_FEE/10`) + `src/policy/policy.h` (`RECOMMENDED_MIN_TX_FEE = COIN/100`) — ~100x Bitcoin's, matching Dogecoin's much lower per-coin value |

All values pulled directly from each project's `master`/`develop` source (fetched live, not guessed) so the floor matches the real network minimum instead of arbitrary padding. This is a **floor**: `effectiveFeeRate = Math.max(feeRate, chainFloor)` — it only raises a too-low rate, it never lowers or pads a normal/high one (no overpaying a legit rate).

## UTXO-04 fix — canonical ZIP-317 action count

Swapped the local `zcashConventionalFee(inputCount)` (deleted, along with its now-dead constants) for `getZcashConventionalFee` imported from `packages/core/chain/chains/utxo/fee/zip317.ts` — the same function `packages/core/mpc/keysign/signingInputs/resolvers/utxo.ts` already uses for the WalletCore Zcash re-plan loop. Output sizes fed in mirror the same recipient + change P2PKH (34 bytes each) + optional OP_RETURN shape already assumed by the builder's size-fee heuristic, so no behavior changes for the existing plain-send tests — the difference only shows up when a memo (or extra outputs) pushes output actions above the input action count.

## Runtime verification (vitest, no Docker)

New colocated fixtures. The builder doesn't expose the computed fee directly, so — same technique as the existing `utxo-zcash-zip317-fee.test.ts` — the floor is observed via the insufficient-funds error, which reports the exact fee charged.

**`packages/sdk/tests/unit/chains/utxo-min-fee-floor.test.ts`** (UTXO-03, 10 tests):
- Dogecoin: a 1 koinu/vB rate is raised to the 100 floor (fee=22800, not 228); a 200 koinu/vB rate (above floor) passes through unchanged (fee=45600)
- Bitcoin / Litecoin / Bitcoin-Cash / Dash: a sub-1 feeRate (0.5) is raised to the 1/vB floor; a normal feeRate (5) passes through unchanged — the no-overpay regression guard

**`packages/sdk/tests/unit/chains/utxo-zcash-zip317-output-actions.test.ts`** (UTXO-04, 4 tests):
- Plain 1-input, no-memo send: unchanged at the pre-fix 10,000 zat floor
- 80-byte memo (max standard-relay size): floor raised to 25,000 zats — the OLD input-only formula would have charged only 10,000 (a 15,000 zat shortfall)
- 10-byte memo: floor raised to 15,000 zats (output actions exceed the 2-action grace window even for a small memo)
- 4 inputs + 80-byte memo together: stays output-bound at 25,000 (proves `max(inputActions, outputActions)`, not just input scaling)

```
$ npx vitest run packages/sdk/tests/unit/chains --exclude "**/.worktrees/**"
 Test Files  15 passed (15)
      Tests  146 passed (146)

$ npx tsc --noEmit -p packages/sdk/tsconfig.json
(clean)

$ npx eslint --config .config/eslint.config.mjs packages/sdk/src/chains/utxo/tx.ts packages/sdk/tests/unit/chains/utxo-min-fee-floor.test.ts packages/sdk/tests/unit/chains/utxo-zcash-zip317-output-actions.test.ts
(clean)

$ node scripts/build-shared-packages.mjs --check
Shared package exports are up to date
```

## Dedup check (STEP 0)

Grepped open PRs for `utxo`/`fee`/`zip317`/`dogecoin` — found #1081 (UTXO-02, Litecoin dust threshold — a different sub-finding, still open, touches the same file in a non-overlapping section) and #1078 (unrelated EVM priority-fee finding). Neither covers UTXO-03/04. No dedup.

## Risk

Low — additive floor (raise-only) + a formula swap to an already-battle-tested canonical function used elsewhere in the repo. No behavior change for existing plain-send / non-memo test coverage (all pre-existing UTXO tests pass unchanged).

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated UTXO transaction fee handling to apply chain-specific minimum relay fee rates more consistently.
  * Improved Zcash fee calculation so transactions with memos or other extra outputs are charged correctly, reducing the chance of relay rejection.
  * Validation now catches unsupported address/script mismatches earlier, making failures clearer.

* **Tests**
  * Added coverage for minimum fee floors across multiple UTXO chains.
  * Added Zcash-specific coverage for memo-based fee calculation and output-aware fee behavior.


gomesalexandre (commented):
## Review — #1089 (UTXO-03/04, draft)

Ran this solo (no codex-adversarial pairing this round — draft + narrow scope, Claude-correctness lane only). Manual verification: pulled the branch into an isolated worktree, live-curled the actual Core repos (bitcoin/litecoin/dogecoin/bitcoin-abc/dash) for the min-relay-fee constants instead of trusting the PR's cited values, and ran the full UTXO suite.

### 1. Overlap w/ teammate work — CLEAR, ship it

Checked the 4 stale-looking origin branches (`786-align-dogecoin-bytefee`, `feat/zcash-zip317-fee-floor`, `fix/zcash-zip317-memo-fee`, `zcash-zip317-fee`). None have open PRs (`gh pr list` confirms). Diffed each against `origin/main` and all 4 are fully subsumed by already-merged commits:

- `786-align-dogecoin-bytefee` == `6f53d2cb` "Align Dogecoin byte fee with app clients (#790)" — merged. Touches `rpc.ts`/`byteFee.ts` (the *fee-rate suggestion* heuristic pulled from Blockchair), a completely different code path from this PR.
- `feat/zcash-zip317-fee-floor` == `3f622f63`/`a1c6f184` — merged. This is literally the ORIGINAL `zcashConventionalFee(inputCount)` this PR is now replacing with the canonical fn.
- `fix/zcash-zip317-memo-fee` == `ba1372e6` "raise memo-send fees to ZIP-317 conventional fee at plan time (#773)" — merged. Fixes WalletCore's `zip_0317` planner in the MPC-keysign resolver path (`core/mpc/keysign/signingInputs/resolvers/utxo.ts`) — a *different builder* than `buildUtxoSendTx` in `tx.ts`.
- `zcash-zip317-fee` == `55ed503e` (#647) — merged, ancient history.

So #1089 doesn't duplicate or revert anything landed. `buildUtxoSendTx` (the RN-bridge send builder, exported all the way to `platforms/react-native/chains/utxo/index.ts` and consumed by the agent app's `signAndBroadcast.ts`/`utxoTx.ts`) is a genuinely separate live path from the already-fixed resolver, and its local `zcashConventionalFee` really did still have the input-only bug pre-#1089. Confirmed by reading current `origin/main` `tx.ts` directly before checking out the branch — the gap is real.

praise: the PR body's own "Dedup check (STEP 0)" section is doing exactly this work already (#1081/#1078 triage) — good discipline, just didn't happen to catch these 4 dead branches since they're not open PRs.

### 2. Floor values — mostly correct, one real should-fix on Dogecoin

I didn't trust the cited Core source values, I curled them live:

| Chain | PR claims | Verified (live curl) | Match? |
|---|---|---|---|
| Bitcoin | 1,000 sat/kvB → 1 sat/vB | `DEFAULT_MIN_RELAY_TX_FEE{1000}` on Core **v29.0**, but **v30.0 lowered it to `{100}`** (0.1 sat/vB) | ⚠️ stale citation, see below |
| Litecoin | 1,000 litoshi/kvB → 1 litoshi/vB | `DEFAULT_MIN_RELAY_TX_FEE = 1000` on litecoin-project/litecoin master | ✅ exact |
| Bitcoin-Cash | 1,000 sat/kvB → 1 sat/vB | `DEFAULT_MIN_RELAY_TX_FEE_PER_KB(1000 * SATOSHI)` on Bitcoin-ABC | ✅ exact |
| Dash | 1,000 duff/kvB → 1 duff/vB | `DEFAULT_MIN_RELAY_TX_FEE{1000}` on dashpay/dash master | ✅ exact |
| Dogecoin | 100,000 koinu/kvB → 100 koinu/vB | `RECOMMENDED_MIN_TX_FEE = COIN/100`, `DEFAULT_MIN_RELAY_TX_FEE = RECOMMENDED_MIN_TX_FEE/10` on dogecoin/dogecoin master → 100,000 koinu/kvB | ✅ exact math, but see should-fix below |

**suggestion:** the Bitcoin citation (`1,000 sat/kvB`) is stale — Bitcoin Core v30.0 dropped `DEFAULT_MIN_RELAY_TX_FEE` to `100` (0.1 sat/vB). Doesn't matter functionally: floor=1 sat/vB is still the conservative/safe side (it's what the economic majority of nodes still run pre-v30, and the absolute sat delta on a ~150-byte tx is negligible, we're talking <100 sats). Non-blocking, just update the comment so the next auditor doesn't inherit a now-wrong "current default" claim.

**should-fix:** Dogecoin's floor uses `DEFAULT_MIN_RELAY_TX_FEE` (100 koinu/vB) — the *mempool-acceptance* minimum — but Dogecoin Core also ships a separate, 10x-higher `DEFAULT_BLOCK_MIN_TX_FEE` (`RECOMMENDED_MIN_TX_FEE` = `COIN/100` = 1,000 koinu/vB), which is the default `-blockmintxfee` miners use to decide what to actually include in a block. This split is Dogecoin-specific — it's a holdover from pre-2016 Bitcoin Core that upstream Bitcoin/Litecoin/Dash/BCH dropped (BTC's own `DEFAULT_BLOCK_MIN_TX_FEE` is just `1`, same as minrelay). A tx paying exactly the 100-floor will *relay* fine (fixes the "silently rejected by every node" failure mode) but under **default miner config** it sits below the block-inclusion threshold — i.e. the exact "stuck tx" symptom the audit is chasing, just one layer down (stuck in mempool instead of stuck in your wallet's outbox). Given DOGE blocks are almost never full in practice this is a lower-severity residual than the pre-fix bug (which was "never relays, period"), so not blocking — but worth either bumping `UTXO_MIN_FEE_RATE.Dogecoin` to `1000` (matching `RECOMMENDED_MIN_TX_FEE`, which the DOGE devs explicitly designed as *the* number everything else derives from) or filing a fast follow-up. Happy to be argued out of this if there's evidence current DOGE miners universally run with `-blockmintxfee` overridden down to the relay minimum, but I couldn't find that.

Confirmed the floor mechanics themselves are sound: `Math.max(opts.feeRate, floor)` — never lowers a legit rate, verified via the test suite's explicit "no-overpay" cases (feeRate=200 on DOGE → fee=45600 unchanged, feeRate=5 on BTC/LTC/BCH/Dash → passes through unchanged). No overpay risk on the mechanism, only the DOGE constant choice above.

### 3. UTXO-04 (canonical ZIP-317 action count) — correct

- Confirmed the import (`getZcashConventionalFee` from `packages/core/chain/chains/utxo/fee/zip317.ts`) is the same canonical fn the MPC-keysign resolver already uses (`max(inputActions, outputActions)`), not a reimplementation.
- `fee = sizeFee > zip317Floor ? sizeFee : zip317Floor` — max, not sum. No double-counting.
- Hand-verified the 80-byte-memo test math independently against `getZcashOpReturnOutputSize`: memo=80B → pushOverhead=3 (PUSHDATA1 range) → script=83B → `getZcashOpReturnOutputSize`=8+1+83=92B. The builder's own `opReturnBytes = 9 + opReturnScript.length` = 9+83=92 — matches exactly (the `9` hardcode is safe because OP_RETURN data is capped at 80 bytes elsewhere in the function, so `scriptLengthSize` is always the 1-byte CompactSize case). Total outputs 34+34+92=160B → ceil(160/34)=5 actions → fee=25,000 zats, matches the test assertion.
- Plain 1-in/1-out (no memo): unchanged at the 2-action grace floor (10,000 zats) — confirmed no regression for the existing/common case.

q: (non-blocking) `zip317OutputSizes` hardcodes `[34n, 34n, ...]` assuming change is always emitted, even though `serializeOutputs` drops the change output below `dustLimit` (`hasChange = change > dustLimit`). This is inherited from the pre-fix code's same 2-output assumption (not a new bug), and the direction is safe (assumes MORE output bytes than reality in the rare dust-drop+memo combo → slight overpay, never underpay). Not asking for a fix, just flagging I traced it so it's not silently unnoticed.

### 4. Tests / build — verified locally, not just trusting the PR body's paste

Isolated worktree (`git worktree add` off the PR's head commit, symlinked `node_modules`, ran `yarn build:shared` first per the project's cross-package-resolution requirement):

```
$ npx vitest run packages/sdk/tests/unit/chains/utxo-min-fee-floor.test.ts packages/sdk/tests/unit/chains/utxo-zcash-zip317-output-actions.test.ts
 Test Files  2 passed (2)
      Tests  14 passed (14)

$ npx vitest run packages/sdk/tests/unit/chains --exclude "**/.worktrees/**"
 Test Files  15 passed (15)
      Tests  146 passed (146)

$ npx tsc --noEmit -p packages/sdk/tsconfig.json
(clean)

$ npx eslint --config .config/eslint.config.mjs packages/sdk/src/chains/utxo/tx.ts packages/sdk/tests/unit/chains/utxo-min-fee-floor.test.ts packages/sdk/tests/unit/chains/utxo-zcash-zip317-output-actions.test.ts
(clean)
```

Matches the PR body's claimed numbers exactly. Changeset present and correctly scoped (`@vultisig/sdk`: patch).

### Verdict

Draft PR → **COMMENT** per the draft-only rule, not an approve/request-changes vote. Substance-wise this is solid: no dedup/overlap issue (checked the thing that actually mattered — 4 branches that looked scary but are all dead), the floor mechanism is fund-safety-correct (raise-only, no overpay), 4/5 chain constants verified exact against live Core source, and UTXO-04's formula swap is a clean, tested improvement with zero behavior change on the common path.

One should-fix worth resolving before flipping to ready: reconsider whether Dogecoin's floor should be `RECOMMENDED_MIN_TX_FEE` (1000 koinu/vB) instead of `DEFAULT_MIN_RELAY_TX_FEE` (100 koinu/vB), since the former is what DOGE's own default miner policy actually gates block inclusion on — otherwise the fix trades "stuck, never relays" for "relays but may sit in mempool a while under default miner config," which is a real but lower-severity residual of the same bug class the audit is targeting.

packages/sdk/src/chains/utxo/tx.ts
```diff
@@ -837,6 +844,38 @@ export function buildUtxoSendTx(opts: BuildUtxoSendOptions): UtxoTxBuilderResult
...
+  // silently produces a stuck/non-relayable tx (most acute on Dogecoin, whose
+  // min relay fee is ~100x Bitcoin's). Raise — never lower — to the floor.
+  const effectiveFeeRate = Math.max(opts.feeRate, UTXO_MIN_FEE_RATE[opts.chain])
+  const sizeFee = BigInt(Math.ceil(txSize * effectiveFeeRate))
+  // UTXO-04: use the canonical ZIP-317 action-count formula (max of input vs
+  // output actions) instead of a local input-only count, so a large memo or
+  // extra outputs aren't under-counted. Output sizes mirror the same
+  // recipient + change P2PKH (34 bytes each) + optional OP_RETURN shape
+  // assumed by the txSize estimate above.
+  const zip317OutputSizes = opReturnScript ? [34n, 34n, BigInt(opReturnBytes)] : [34n, 34n]
```
NeOMakinG, line 905: `should-fix:` (fail-safe) `zip317OutputSizes` hard-codes a change output (`[34n, 34n, ...]`) *before* `change` is known. The canonical `getZcashTransparentOutputSizes` and the MPC keysign path (`resolvers/utxo.ts`) only count change when `change > 0n`. For a Zcash **memo send with no change** (e.g. send-max at a 34-byte output boundary), this over-counts output actions by 1 → the floor is inflated by 5000 zats, which can surface as a false `insufficient funds` on an exact-spend. Direction is safe (over-estimate → never underpay/stuck), and the combo is rare (Zcash isn't a THORChain memo chain), so non-blocking — but it diverges from the canonical helper. Consider deriving the output sizes from the actual change decision (or add a comment documenting the deliberate conservative over-estimate, matching the `2*34` in the `txSize` heuristic above).

packages/sdk/src/chains/utxo/tx.ts
```diff
@@ -837,6 +844,38 @@ export function buildUtxoSendTx(opts: BuildUtxoSendOptions): UtxoTxBuilderResult
...
+
+  // Approximate tx size for fee calc — matches app's heuristic.
+  const bytesPerInput = spec.scriptType === 'p2wpkh' ? 68 : 150
+  // 8-byte value + varint(scriptLen) + scriptLen; scriptLen <= 82 so the varint is 1 byte.
+  const opReturnBytes = opReturnScript ? 9 + opReturnScript.length : 0
+  const txSize = inputs.length * bytesPerInput + 2 * 34 + 10 + opReturnBytes
+  // UTXO-03: a caller-supplied feeRate below the chain's real min-relay-fee
+  // silently produces a stuck/non-relayable tx (most acute on Dogecoin, whose
+  // min relay fee is ~100x Bitcoin's). Raise — never lower — to the floor.
+  const effectiveFeeRate = Math.max(opts.feeRate, UTXO_MIN_FEE_RATE[opts.chain])
```
NeOMakinG, line 898: `should-fix:` (scope / follow-up) this floor only hardens the standalone SDK builder. The MPC/WalletCore keysign path feeds `byteFee` (from `getUtxoByteFee`, `packages/core/chain/chains/utxo/fee/byteFee.ts`) straight through with **no** 1000 koinu/vB floor — and that estimator returns DOGE as `base / 4n`, which can dip below 1000. So a DOGE send routed through MPC keysign can still be built/signed below the block-inclusion minimum, the exact stuck-tx symptom UTXO-03 targets. This is **pre-existing (not a regression)** and out of this PR's `buildUtxoSendTx` scope, so non-blocking here — but please open a follow-up to floor DOGE `byteFee` in `getUtxoByteFee` too, otherwise the fix only covers one of the two signing stacks.

NeOMakinG (approved):
## testing approach

Static trace of `buildUtxoSendTx` + the canonical `getZcashConventionalFee` (`packages/core/chain/chains/utxo/fee/zip317.ts`), hand-verified every fee assertion in both new test files, cross-checked each per-chain floor value against its source-of-truth, and audited the MPC/WalletCore keysign path for divergence. Two-lane (Claude load-bearing + Codex gpt-5.4 adversarial). Worktree `/tmp/wt-sdk-1089` @ b87d9952. Could not run vitest locally (pnpm nested-workspace `@noble/curves` subpath-export resolution artifact, not a code defect); **CI Unit Tests pass (18m) + Integration + Lint/TypeCheck + CodeRabbit all green**, and the arithmetic below checks out by hand.

### UTXO-03 floor values — verified correct + right unit

| Chain | Floor (per vB) | Correct? |
|---|---|---|
| Bitcoin | 1 sat/vB | ✅ `DEFAULT_MIN_RELAY_TX_FEE` 1000 sat/kvB ÷ 1000 |
| Litecoin | 1 litoshi/vB | ✅ same order as BTC (LTC's 10× is `DUST_RELAY_TX_FEE`, not min-relay — correctly noted) |
| Bitcoin-Cash | 1 sat/vB | ✅ |
| Dash | 1 duff/vB | ✅ |
| Dogecoin | **1000 koinu/vB** | ✅ block-INCLUSION min (`RECOMMENDED_MIN_TX_FEE = COIN/100` = 1e6 koinu/kvB ÷ 1000), correctly chosen over the lower relay-min (100) that relays-but-doesn't-mine |

Unit is consistently per-vByte and multiplies `txSize` the same way `sizeFee` always has. Applied as `Math.max(feeRate, floor)` — raises a too-low rate, never lowers a legit one (no overpay). Reaches the signable tx via `sizeFee → fee → change`. Per-chain isolation confirmed (BTC floor doesn't touch DOGE etc.).

### UTXO-04 ZIP-317 — canonical

Swapped the local input-only `zcashConventionalFee` for `getZcashConventionalFee` = `5000 * max(2, max(ceil(inputBytes/150), ceil(outputBytes/34)))` — matches ZIP-317 exactly (marginal 5000, grace 2, 148B P2PKH input). Zcash-scoped (`chain === 'Zcash'` gate, 0n elsewhere). Hand-checked: plain=10000, 10B-memo=15000, 80B-memo=25000, 4-input+80B-memo=25000 (output-bound). OP_RETURN size model (`9 + script.length`) matches canonical `getZcashOpReturnOutputSize` at every boundary.

## should-fix

Two non-blocking items (both **fail-safe** — neither can produce an underpaid/stuck tx). See inline.

## q

See inline on the Zcash output-size shape.

## risk

Low. Additive raise-only floor + a swap to an already-battle-tested canonical function (same one the MPC keysign path uses). No behavior change for existing plain-send tests. The one residual risk (DOGE floor absent on the MPC keysign path) is **pre-existing, not a regression** — this PR strictly improves the builder path.

## verdict

**APPROVE.** Floors are correct values in the right unit, raise-not-reject, and reach the signable tx; ZIP-317 count is canonical + Zcash-scoped; no cross-chain regression; changeset present; CI green. The two should-fix items are fail-safe follow-ups, not blockers.

🤖 Reviewed by Claude Opus 4.8 (1M) — two-lane (Claude + Codex gpt-5.4).

gomesalexandre (commented):
current-head sanity pass after the main update. diff vs fresh origin/main is still clean.

- canonical ZIP-317 path still matches `getZcashConventionalFee({ inputCount, outputSizes })` and the memo output-size math still lines up with the shared helper.
- `estimateUtxoTxFee` and `buildUtxoSendTx` still apply the same floor + ZIP-317 logic, so selection/build stay aligned.
- no new issues introduced by the branch update.

one non-blocking doc nit only: the PR body's Dogecoin table still says 100 koinu/vB, but the code now uses 1000 koinu/vB. worth updating the description so reviewers are reading the same number the code ships.

PR #1091, opened 2026-07-08
feat(sdk): notify registered devices when a keysign QR is ready

ahdzib-maya:
## Problem

A secure-vault keysign generates the QR and then just polls the relay (`waitForDevices`). The co-signing device gets no signal — a person has to be actively watching to scan the QR within the relay window. In practice this is a frequent cause of `Timeout waiting for devices. Got 1/2 devices.`

The SDK already ships `PushNotificationService.notifyVaultMembers` (→ `POST /notification/notify`), and devices already register with that server (a phone and laptop on the same vault already push each other on keysign). It was simply never called from the signing path.

## Change

Call `notifyVaultMembers` from **both** `SecureVault` sign paths the moment the QR is ready, so the vault's registered devices get a native push (APNs/FCM/WebPush) prompting them to co-sign.

- **Best-effort**: wrapped in `try/catch`, never blocks or fails the sign, and no-ops when there is no push service or no registered devices.
- **`vault_id`** uses `computeNotificationVaultId` = `SHA256(utf8(ecdsaHex + hexChainCode))`, matching how devices register — **not** the raw ecdsa pubkey. (Passing the raw pubkey is why ad-hoc attempts silently never landed.)
- The QR is still emitted as before via `onQRCodeReady`; nothing changes for existing consumers.

## Testing

Verified live against `api.vultisig.com/notification`: `GET /notification/vault/<computed-id>` returns `200` for a registered vault, and a secure-vault send now pushes to the paired device — the co-sign completed in ~2s (device tapped the notification) versus the usual manual-scan wait.

## Summary by CodeRabbit

* **New Features**
  * Signing flows now emit a QR-code-ready signal and automatically send a best-effort push notification when a keysign QR payload is generated, so connected devices can react faster.
* **Reliability / Bug Fixes**
  * Push notification delivery failures no longer interrupt signing—errors are handled separately and the signing process continues normally.


gomesalexandre (changes requested):
ser logic is solid - best-effort push, correct vault_id via `computeNotificationVaultId`, void + try/catch, two call sites. clean.

two things to fix before this is mergeable:

**preferably-blocking: import sort in SecureVault.ts**
`computeNotificationVaultId` is imported from `../utils/` but it's placed before the `../services/` imports. `simple-import-sort` sorts by path and `s` < `u`, so services must come first - will fail lint CI. fix:
```ts
import { RelaySigningService } from '../services/RelaySigningService'
import { SecureVaultCreationService, type SecureVaultCreationStep } from '../services/SecureVaultCreationService'
// ...types...
import { normalizeToHex } from '../utils/bytes'
import { computeNotificationVaultId } from '../utils/computeNotificationVaultId'
```

**preferably-blocking: missing changeset**
this touches `packages/sdk/src/` so it needs a changeset. run `yarn changeset`, pick `@vultisig/sdk` patch, commit the generated `.changeset/<slug>.md`.

gomesalexandre (approved):
APPROVE. Static pass on HEAD a383a1de only.

- `SecureVault.sign()` and `signBytes()` still emit `qrCodeReady` synchronously and still forward the consumer callback; the new push path is detached via `void this.notifyDevices(qrPayload)`, so notification failures should not block signing.
- `notifyDevices()` is correctly best-effort: it no-ops when push service / ecdsa pubkey / chainCode are missing, computes the notification vault id via `computeNotificationVaultId(ecdsa, chainCode)`, and swallows failures into the vault `error` event instead of throwing through the signing path.
- Changeset is present.

Only gap I see is coverage: this diff would benefit from a focused unit test asserting `notifyVaultMembers()` gets the computed notification vault id and that a thrown push error stays non-fatal, but I do not see a correctness blocker in the shipped code.

gomesalexandre (approved):
approve on refreshed head e6ceb932.

mergefix only. i merged latest `main` into the branch and pushed it to the contributor fork so the PR is mergeable again. the merge was clean and did not touch the SecureVault notification files from this PR, so the reviewed behavior stays the same while picking up the current base.

no new blocker found.

PR #1092, opened 2026-07-08
fix(prep): enforce native-quote expiry + amount consistency in vault-free swap builder

### LINKED ISSUE #1067: [audit-SDK-FUNDSAFETY-01 · 02] Agent-reachable vault-free swap helper skips the amount↔quote cross-check and general-sw
## Context (fund-safety audit round 1, 2026-07-08 · finding SDK-FUNDSAFETY-01 · 02 · Medium)

What. tools/prep/swap.ts (docstring: "intended for MCP servers and other contexts") takes amount and swapQuote as independent params with no consistency check — and its docstring explicitly punts: "Quote expiry validation is also a consumer concern." For native swaps, keysign/swap/build.ts:129 takes the signed input from the caller's amount while the min-out limit comes from the quote's own sizing. The vault wrapper SwapService.ts:131 enforces expiry; the vault-free helper bypasses it.

## Impact
Impact. The agent path can build a native-swap payload whose input amount and slippage limit were computed for different sizes, or sign a stale aggregator quote. Bounded by the aggregator's on-chain min-return (worst case: revert or bounded-worse fill), hence MED not HIGH — but the enforced-vs-agent-reachable asymmetry is the exact shape worth closing.

## Effort / risk / confidence
Effort SRisk LOWConfidence MED — inconsistency real; needs call-site confirmation

## Prompt for AI agents
Implement the fix for audit finding **SDK-FUNDSAFETY-01 · 02** in `vultisig/vultisig-sdk` at the anchors named above. What. tools/prep/swap.ts (docstring: "intended for MCP servers and other contexts") takes amount and swapQuote as independent params with no consistency check — and its docstring explicitly punts: "Quote expiry validation is also a consumer concern." For native swaps, keysign/swap/build.ts:129 takes the signed input from the caller's amount while the min-out limit comes from the quote's own sizing. The vault wrapper SwapService.ts:131 enforces expiry; the vault-free helper bypasses it. Minimal, additive, fail-closed; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI build + unit/CLI drive; before vs after.). **Grep open PRs in this repo first** — advance an existing PR (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **SDK-FUNDSAFETY-01 · 02** (round 1, Medium) · bead `agent-backend-ts-p9wl1`._

gomesalexandre:
Closes the exact vault-wrapped-vs-vault-free fund-safety asymmetry this audit round is about: `prepareSwapTxFromKeys` (the vault-free, agent-reachable swap payload builder — used wherever only the public identity is available, no key shares) enforced neither quote expiry nor amount-vs-quote consistency, unlike `SwapService.prepareSwapTx`.

Ref: plan 005 (`plans/005-vault-free-swap-safety-checks.md`).

## What

Two checks, scoped to what's confidently enforceable after root-causing the actual type shapes (not the plan's original assumptions):

**Native-quote expiry.** `quote.native.expiry` is a real absolute unix-seconds deadline sourced from the THORChain/Maya quote API — enforced, mirroring core's own `assertQuoteNotExpired` (`nativeSwapQuoteToSwapPayload.ts`). Fires before any `getWalletCore()`/`getPublicKey()` call, so an expired quote never triggers wallet-core or key-derivation side effects.

General quotes carry **no expiry field at all** in the core `SwapQuote` shape this helper receives — the `expiresAt` the vault wrapper checks (`SwapService.ts:131`) is an artificial SDK-side clock computed fresh at fetch time (`Date.now() + DEFAULT_QUOTE_EXPIRY_MS`), not quote data. Nothing meaningful to enforce here for general quotes, and not a regression — general already had zero enforcement in the vault-free path.

**Amount vs committed sell amount.** Cross-checked only where the quote confidently commits to a sell amount independent of the caller's `amount`:
- `general.tx.transfer.amount` (UTXO/Cosmos deposit-channel routes)
- `general.tx.cowswap_order.sellAmount` (CoW's actual signed EIP-712 order amount)

Native quotes have no committed-sell-amount field anywhere (checked the type + the live THORChain quote response shape — confirmed empty). `evm`/`solana` general quotes encode the amount inside opaque aggregator calldata. Both fail **open** (no throw) rather than invent a comparison that could false-reject a legitimate swap.

**Named residual (not silently accepted):** an EVM-general vault-free swap — arguably the single most common shape — still has **neither** check after this lands. The calldata-amount-decode gap defers both. Not a regression (nothing enforced it before either), but it's the real remaining exposure once native + transfer/cowswap_order are covered. Follow-up: an EVM calldata amount-decoder is the prerequisite to close it.

## Testing

10 new tests in `swap.test.ts` + 3 pre-existing tests' quote fixtures updated (bare `{__mock}` stub → real `{quote:{native:{expiry}}}` shape, so they exercise the actual branches rather than break on the new guards — same outcomes, updated shape only).

Full `yarn workspace @vultisig/sdk test`: 144 passed / 29 failed test files, 1 failed test — verified via `git stash` before/after diff that this exactly matches the pre-existing baseline (missing `@vultisig/mpc-types`/`@vultisig/core-config` build artifacts, an environment gap unrelated to this change, confirmed present before any edit here).

Ran a codex adversarial review before finalizing — no blocking or should-fix findings. Took its two nits: strengthened the expired-quote test to assert no `getWalletCore`/`getPublicKey` calls, and added a scientific-notation-amount test (`toChainAmount` format-handling edge case).

## Risk

LOW — additive checks only, both fail-open by design on anything not confidently comparable. No existing call site's outcome changes (verified via the 3 updated fixture tests still asserting the same pass/build results).

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Added vault-free swap pre-checks that reject expired native quotes and enforce CoW order expiry before signing.
  * Added caller `amount` vs committed sell-amount consistency checks for verifiable CoW order quotes, while leaving transfer/opaque quote paths fail-open to avoid blocking valid swaps.
* **Documentation**
  * Added internal plans and updates outlining upcoming encoder parity vectors and vault-free swap safety checks.

closes #1067


packages/sdk/src/tools/prep/swap.ts
```diff
@@ -15,6 +17,52 @@ export type PrepareSwapTxFromKeysParams = {
...
+// opaque aggregator calldata. Both fail OPEN here — do not invent a comparison that could
+// false-reject a legitimate swap just because we can't confidently verify it.
+const assertAmountMatchesCommittedSellAmount = (params: PrepareSwapTxFromKeysParams): void => {
+  const { quote } = params.swapQuote
+  if (!('general' in quote)) return
+
+  const committed = matchRecordUnion(quote.general.tx, {
+    evm: () => undefined,
+    solana: () => undefined,
+    transfer: tx => tx.amount,
```
NeOMakinG: `blocker:` `transfer.amount` is **provider-committed**, not caller-derived — `getTransferAmount` (`getSwapKitQuote.ts:437`) reads the provider's `depositAmount`/PSBT/`tx[0].amount`, which legitimately diverges from `toChainAmount(caller amount)`. A passing core test already encodes this: request `100_000n` → committed `99_999n` (`getSwapKitQuote.test.ts:346`). The exact `!==` below will throw on every such route across 8 SwapKit source chains (BTC/BCH/DOGE/LTC/XRP/TON/TRON/ZEC), and since `SwapService.prepareSwapTx` (`SwapService.ts:141`) routes through this helper, it breaks the **vault-wrapped production path** too. Scope this to `cowswap_order` only, compare against the provider's `depositAmount`, or allow a bounded tolerance.

packages/sdk/src/tools/prep/swap.ts
```diff
@@ -15,6 +17,52 @@ export type PrepareSwapTxFromKeysParams = {
...
+    cowswap_order: order => BigInt(order.sellAmount),
```
NeOMakinG: `should-fix:` the committed CoW value is `grossSellAmount = sellAmount + feeAmount` (`getCowSwapQuote.ts:81`), not the raw quote `sellAmount`. If the caller passes the **net** amount (what UIs usually show), this comparison throws on the fee delta — a second false-reject vector. Please confirm the caller-amount convention matches the gross value. The PR's cowswap test hand-sets `sellAmount == requested`, so it doesn't exercise the gross path.

packages/sdk/src/tools/prep/swap.ts
```diff
@@ -15,6 +17,52 @@ export type PrepareSwapTxFromKeysParams = {
   swapQuote: SwapQuote
 }
 
+// Fund-safety, agent-reachable (audit ABTS-005/plan 005): this is the vault-free swap builder,
+// so it's the one path an agent (no key shares) can drive. It previously enforced NEITHER quote
+// expiry NOR amount↔quote consistency, unlike the vault-wrapped `SwapService.prepareSwapTx`.
+//
+// Native-quote expiry (`quote.native.expiry`) is a REAL absolute deadline sourced from the
+// THORChain/Maya quote API — mirrors core's own `assertQuoteNotExpired`
+// (nativeSwapQuoteToSwapPayload.ts). General quotes carry NO expiry field at this layer at all:
```
NeOMakinG: `should-fix:` "general quotes carry NO expiry field at this layer at all" isn't accurate — `cowswap_order.validTo` (`getCowSwapQuote.ts:87`, `= quote.validTo`) is an authoritative absolute deadline on the signed EIP-712 order. This PR still lets a past-`validTo` CoW order be prepared and signed. Not the THOR rotated-inbound fund-loss case, but a real gap in the "expiry is enforced here" story with the field readily available to check.

packages/sdk/src/tools/prep/swap.ts
```diff
@@ -15,6 +17,52 @@ export type PrepareSwapTxFromKeysParams = {
...
+// the `expiresAt` `SwapService.prepareSwapTx` checks is an artificial SDK-wrapper clock computed
+// fresh at fetch time (`Date.now() + DEFAULT_QUOTE_EXPIRY_MS`), not data the quote itself carries
+// — there's nothing meaningful to enforce here for general quotes, so this only covers native.
+const assertNativeQuoteNotExpired = (expirySeconds: number): void => {
+  if (expirySeconds <= Math.floor(Date.now() / 1000)) {
```
NeOMakinG, line 30: `praise:` faithful mirror of core's `assertQuoteNotExpired` (`nativeSwapQuoteToSwapPayload.ts:22-24`) — authoritative `quote.native.expiry`, fail-closed `<=`, and it runs before any wallet-core/pubkey side effect (the test asserts this). This half is ready to land.

NeOMakinG (changes requested):
Native-quote expiry enforcement here is **solid** — `assertNativeQuoteNotExpired` is a faithful, byte-for-byte mirror of core's `assertQuoteNotExpired` (`nativeSwapQuoteToSwapPayload.ts`), reads the authoritative `quote.native.expiry` (real THORChain/Maya unix-seconds deadline, not a re-derived clock), fails **closed** with `<=`, and gates the signable payload *before* any `getWalletCore()` / `getPublicKey()` / `buildSwapKeysignPayload()` side effect. That half is ready.

The **amount-consistency** half has a real over-reject that I can't wave through on a fund-critical builder, plus a stale-order gap on the general side. Details below.

## testing approach

- Worktree at `e9936c65`, `yarn install` + `build:shared`, ran the targeted suite: **13/13 pass** (`packages/sdk/tests/unit/tools/prep/swap.test.ts`).
- Traced the two guards against the type shapes (`NativeSwapQuote.expiry: number`, `GeneralSwapTx` union), the mirror in core (`nativeSwapQuoteToSwapPayload.ts:22-24,63`), the actual signed amount in `build.ts:129` (`transferTx?.amount ?? toChainAmount(amount, fromCoin.decimals)`), and both call sites (`SwapService.ts:141` vault-wrapped + the exported vault-free helper).
- Bypass check: `buildSwapKeysignPayload` has exactly one production caller (`prep/swap.ts:127`) — no path skips the guards.
- Two-lane: Claude (load-bearing) + Codex `gpt-5.4` (adversarial second opinion, exit 0, correctly scoped to this PR). The transfer over-reject and CowSwap `validTo` gap below are cross-confirmed against source.

## blocking

**Exact-equality amount check false-rejects legitimate SwapKit transfer routes.** `assertAmountMatchesCommittedSellAmount` throws whenever `toChainAmount(caller amount) !== transfer.amount`. But `transfer.amount` is **provider-committed**, not caller-derived: `getTransferAmount` (`getSwapKitQuote.ts:437`) reads the provider's `depositAmount` / PSBT / `tx[0].amount`, which legitimately diverges from the requested amount. There is already a passing core test where the caller requests `100_000n` and the committed PSBT `transfer.amount` is `99_999n` (`getSwapKitQuote.test.ts:346`). This affects **8 source chains** (`swapKitTransferSourceChains`: Bitcoin, BCH, Doge, Litecoin, Ripple, Ton, Tron, Zcash) and — because `SwapService.prepareSwapTx` (`SwapService.ts:141`) now routes through this helper — hits the **vault-wrapped production swap path too**, not just vault-free/MCP. The PR's unit tests only use fixtures where `amount === committed`, so they pass while the real provider-adjusted responses would throw. Recommend: don't gate provider-committed transfer routes on exact caller-amount equality (compare against the provider's own `depositAmount`, allow a bounded tolerance, or scope this check to `cowswap_order` only where the caller amount genuinely never reaches the signed order).

## should-fix

**CowSwap `validTo` is an authoritative expiry that goes unchecked.** The inline comment and PR body state "general quotes carry NO expiry field at this layer at all" — but `cowswap_order.validTo` (`getCowSwapQuote.ts:87`, `= quote.validTo`) is exactly that: an absolute deadline on the actual signed EIP-712 order. This PR still lets an already-past-`validTo` CowSwap order be prepared and signed. Not the THOR rotated-inbound fund-loss case (the CoW API rejects a stale order at submit), but it's a real hole in the "expiry is enforced here" story and the field is right there to check.

**CowSwap committed amount is `grossSellAmount` (net + fee), not the raw request.** `sellAmount` = `BigInt(quote.sellAmount) + BigInt(quote.feeAmount)` (`getCowSwapQuote.ts:81`). The amount check compares `toChainAmount(caller amount)` against this gross value. If a caller passes the net sell amount (what most UIs display), the check throws on the fee delta. Please confirm the caller-amount convention against the gross committed value, or this is a second false-reject vector. (The PR's cowswap test hand-sets `sellAmount` == requested, so it doesn't exercise the gross path.)

**Missing changeset — the fix ships inert.** No `.changeset/*.md` bumps `@vultisig/sdk`. CI stays green because `check-sdk-bundled-changeset.mjs` only guards `packages/core/*` + `packages/lib/utils` (this PR is confined to `packages/sdk/`), so the guard doesn't fire — but without a changeset, `@vultisig/sdk` gets no version bump and consumers (mcp, app) never receive this fund-safety fix. Siblings #1073 / #1075 both shipped a changeset; please add one.

## q

**Vault-free helper has no broadcast-time freshness re-check (TOCTOU / rotated inbound).** The vault-wrapped path re-validates native expiry **and** inbound-vault rotation at broadcast (`BroadcastService.ts:65` → `nativeSwapBroadcastGuard.ts:51,81` — compares the active THOR/Maya inbound address against the payload's `vaultAddress`). `prepareSwapTxFromKeys` is exported for vault-free/MCP consumers, but that broadcast guard isn't part of the public helper surface. So an external consumer can build a fresh native deposit, let it go stale during MPC signing, and submit via a custom broadcaster to a rotated inbound vault (the real fund-loss vector). Enforcing expiry at prepare-time narrows the window but doesn't close it for standalone consumers — is exposing an equivalent broadcast-time inbound-rotation guard in scope, or tracked as follow-up?

## risk

MEDIUM. The expiry guard is a genuine, correct fund-safety improvement. But the amount check as written trades an under-enforcement gap for an **over-rejection regression** that breaks legitimate provider-committed swaps across 8 chains on both the vault-free and vault-wrapped paths — so this is net-negative to ship unscoped.

## verdict

**REQUEST_CHANGES.** Land the native-expiry check as-is (solid), but rework the amount-consistency check so it stops false-rejecting provider-committed transfer routes (blocking), address the CowSwap `validTo` / gross-amount gaps (should-fix), and add the `@vultisig/sdk` changeset so the fix actually reaches consumers.

---
🤖 Reviewed by Claude (load-bearing) + Codex gpt-5.4 (adversarial second opinion). Two-lane, no runtime QA on this pass (pure builder logic; 13/13 unit + source trace).

NeOMakinG (approved):
r2 re-review at `f3d8bc32` (+ my changeset commit `bc671e34`). The r1 blocker and all should-fixes are resolved. Two-lane (Claude load-bearing + Codex gpt-5.4 adversarial) — both converge, no new issues. **Approving.**

## testing approach
- Worktree at PR HEAD; `git diff e9936c65..f3d8bc32` (r1→r2 delta, prod file `packages/sdk/src/tools/prep/swap.ts` +the test file).
- Traced signing authority for the transfer route into `packages/core/mpc/keysign/swap/build.ts` and the CoW gross derivation into `getCowSwapQuote.ts`.
- `yarn install && yarn build:shared && vitest run tests/unit/tools/prep/swap.test.ts` → **14/14 pass**.
- Codex gpt-5.4 second-opinion lane (independent read of the same delta): "No new issues on PR #1092 at f3d8bc32."

## prior issues — status
|---|---|---|---|---|---|
| r1 | 1 | blocker | exact `!==` amount check false-rejects provider-committed `transfer.amount` (8 SwapKit chains BTC/BCH/DOGE/LTC/XRP/TON/TRON/ZEC + vault-wrapped `SwapService.prepareSwapTx`) | ✅ fixed | `transfer: () => undefined` now excludes the route from the check (`swap.ts:57`). Safe: the signed amount for a transfer route is the provider-committed `transferTx.amount ?? toChainAmount(...)` (`build.ts:129`) — the caller's `amount` was never authoritative, so exclusion introduces no tamper hole. SwapKit BTC PSBT routes still validate `expectedToAmount` from the PSBT (`build.ts:85-86`). |
| r1 | 2 | should-fix | CoW committed amount should be `grossSellAmount = sellAmount + feeAmount`, not net | ✅ fixed | `cowswap_order: order => BigInt(order.sellAmount) + BigInt(order.feeAmount)` (`swap.ts:58`). Non-tautological: `getCowSwapQuote` requests `sellAmountBeforeFee = caller sellAmount` and `net + fee == sellAmountBeforeFee`, so gross == caller amount for a legit quote; the check catches a stale/mismatched-amount quote. |
| r1 | 3 | should-fix | CoW `validTo` (authoritative EIP-712 order deadline) went unchecked | ✅ fixed | new `assertCowQuoteNotExpired(quote.general.tx.cowswap_order.validTo)` (`swap.ts:105`), fail-closed, gates before `getWalletCore`/`getPublicKey`/`buildSwapKeysignPayload`. |
| r1 | 4 | should-fix | no `@vultisig/sdk` changeset → fix ships inert (CI green because the bundled-changeset guard only covers `packages/core`+`lib/utils`) | ✅ fixed | added in-code as `bc671e34` — `.changeset/vault-free-swap-expiry-amount-abts005.md` (`@vultisig/sdk: patch`); `yarn changeset status` confirms the sdk bump. Matches siblings #1073/#1075. |
| r1 | — | (solid) | native-quote expiry half | ✅ unchanged | `assertNativeQuoteNotExpired` byte-identical; still authoritative (`quote.native.expiry`), zero-margin fail-closed, gates before side effects. |

## blocking
None.

## preferably-blocking
None.

## should-fix
None outstanding — all four r1 items resolved.

## suggestion
- `swap.ts:58` — in the emitted `cowswap_order`, `feeAmount` is hardcoded `'0'` and `sellAmount` is already the gross (`getCowSwapQuote.ts:91,96`), so `+ BigInt(order.feeAmount)` currently adds `0`. Harmless and defensively correct (grosses up correctly if a future CoW path ever emits a net `sellAmount` with a non-zero `feeAmount`), just noting it's inert against the present producer. No change required.

## q
None.

## risk
Low. The change only *removes* a false-reject (transfer) and *adds* two fail-closed guards (CoW `validTo`, CoW gross-amount). No new fail-open: `validTo`/`feeAmount` are mandatory fields on the `cowswap_order` source type (`GeneralSwapQuote.ts`), so the `BigInt(order.feeAmount)` / `.validTo` reads can't throw on `undefined` for that arm. Native expiry half untouched. Residual (pre-existing, out of scope): vault-free consumers still don't get the broadcast-time rotated-inbound guard (`nativeSwapBroadcastGuard.ts`) — noted in r1, not introduced here.

## verdict
**APPROVE.** r1 blocker fixed (transfer excluded, production path + 8 SwapKit chains no longer false-reject, no new tamper hole), CoW uses gross + `validTo` fail-closed, native expiry still solid, changeset now present so the fix ships non-inert. 14/14 unit tests pass; Codex second lane concurs with zero new findings. CI (Unit/Integration/Lint) re-running on `bc671e34` — non-blocking; the local suite is green.

---
🤖 Two-lane review: Claude Opus 4.8 (load-bearing) + Codex gpt-5.4 (adversarial second opinion).

Issue #1066, opened 2026-07-08
[audit-SDK2-02 · SDK2-03 · BE-05] findByTicker resolves a bare symbol with zero chain disambiguation (latent landmine); getE


gomesalexandre:
## Context (fund-safety audit round 2, 2026-07-08 · finding SDK2-02 · SDK2-03 · BE-05 · Low)

SDK2-02: findByTicker (findByTicker.ts:8) is coins.find(c => c.ticker === ticker) — first match wins, no chain scoping. Currently unreferenced, but a public export literally named for symbol resolution; wired into a fund path later, "USDC" silently resolves to whichever chain is first in array order. Delete or make chain-aware now. SDK2-03: getEvmFeeQuote's capGasLimit can call bigIntMax() with zero args → opaque throw if a future caller omits both fallbacks (unreachable today). BE-05: SEND_SLICE_RE (send.ts:44) omits get_price though the comment says it's shared — reconcile before someone builds on the stale comment.

## Effort / risk / confidence
Effort SRisk LOWConfidence MED (latent, not live)

## Runtime repro (BEFORE + AFTER — mandatory)
SDK-CLI: build the SDK and unit/CLI-drive the affected path with a fixture; assert before (bug) vs after (fixed).

## Prompt for AI agents
Implement the fix for audit finding **SDK2-02 · SDK2-03 · BE-05** in `vultisig/vultisig-sdk`. SDK2-02: findByTicker (findByTicker.ts:8) is coins.find(c => c.ticker === ticker) — first match wins, no chain scoping. Currently unreferenced, but a public export literally named for symbol resolution; wired into a fund path later, "USDC" silently resolves to whichever chain is first in array order. Delete or make chain-aware now. SDK2-03: getEvmFeeQuote's capGasLimit can call bigIntMax() with zero args → opaque throw if a future caller omits both fallbacks (unreachable today). BE-05: SEND_SLICE_RE (send.ts:44) omits get_price though the comment says it's shared — reconcile before someone builds on the stale comment.  Make the minimal, additive, fail-closed change at the anchors named above; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI: build the SDK and unit/CLI-drive the affected path with a fixture; assert before (bug) vs after (fixed).). **Grep open PRs in this repo first** — several audit findings already have in-flight PRs; advance/verify an existing one (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
[...]

PR #1093, opened 2026-07-08
fix(coin): findByTicker throws on cross-chain ambiguity, not a silent first-match (SDK2-02)

gomesalexandre:
Part of #1066 (audit SDK2-02; SDK2-03 + BE-05 tracked separately — do NOT auto-close).

## tl;dr
`findByTicker` resolved a bare ticker via `coins.find(c => c.ticker === ticker)` — **array-order first match, no chain scoping**. `USDC` exists on many chains, so if this (currently-unreferenced, but *published*) helper were wired into a fund path it would silently resolve to whichever chain was first → send to the wrong network.

## Fix
Return the **unique** match (or `null` when absent); **throw** when the ticker is ambiguous across >1 chain, forcing the caller to disambiguate. Kept as a **patch** (signature unchanged, non-breaking) rather than deleting a published export.

## Runtime (vitest, 3/3)
single-chain ticker → returns it; absent → null; cross-chain ambiguous (`USDC` on Eth/Polygon/Base) → **throws**.

Changeset: `@vultisig/sdk` patch. Low-risk (the export has zero in-repo callers today) — this is preventive hardening before it's wired into a fund path.

### DIFF
--- a/packages/core/chain/coin/utils/findByTicker.ts
+++ b/packages/core/chain/coin/utils/findByTicker.ts
@@ -5,5 +5,19 @@ type FindByTickerInput<T extends Coin> = {
   ticker: string
 }
 
-export const findByTicker = <T extends Coin>({ coins, ticker }: FindByTickerInput<T>): T | null =>
-  coins.find(c => c.ticker === ticker) ?? null
+// SDK2-02 (audit r2): a bare ticker like "USDC" exists on many chains (Ethereum/Polygon/Arbitrum/Base/…).
+// The old `coins.find(c => c.ticker === ticker)` returned the ARRAY-ORDER first match — if this ever got
+// wired into a fund path, "USDC" would silently resolve to whichever chain happened to be first, sending to
+// the wrong network. Refuse to guess: return the unique match (or null when absent), and THROW when the
+// ticker is ambiguous across more than one chain so the caller must disambiguate by chain.
+export const findByTicker = <T extends Coin>({ coins, ticker }: FindByTickerInput<T>): T | null => {
+  const matches = coins.filter(c => c.ticker === ticker)
+  if (matches.length === 0) return null
+  const chains = new Set(matches.map(c => c.chain))
+  if (chains.size > 1) {
+    throw new Error(
+      `findByTicker: ticker "${ticker}" is ambiguous across ${chains.size} chains (${[...chains].join(', ')}); pass a chain-scoped lookup instead of a bare ticker`
+    )
+  }
+  return matches[0] ?? null
+}
--- a/.changeset/findbyticker-ambiguity-guard.md
+++ b/.changeset/findbyticker-ambiguity-guard.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(coin): findByTicker throws on cross-chain ambiguity instead of a silent first-match (SDK2-02)
+
+`findByTicker` resolved a bare ticker via `coins.find(c => c.ticker === ticker)` — array-order first match, no chain scoping. A symbol like "USDC" exists on many chains, so if this (currently-unreferenced public) helper were wired into a fund path it would silently resolve to whichever chain was first, sending to the wrong network. It now returns the unique match (or null when absent) and throws when the ticker is ambiguous across more than one chain, forcing the caller to disambiguate.
--- a/packages/core/chain/coin/utils/findByTicker.test.ts
+++ b/packages/core/chain/coin/utils/findByTicker.test.ts
@@ -1,0 +1,23 @@
+import { describe, expect, it } from 'vitest'
+
+import { Chain } from '../../Chain'
+import type { Coin } from '../Coin'
+import { findByTicker } from './findByTicker'
+
+const coin = (chain: Chain, ticker: string): Coin => ({ chain, ticker }) as Coin
+
+describe('findByTicker (SDK2-02 — no silent cross-chain first-match)', () => {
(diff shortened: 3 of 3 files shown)


packages/core/chain/coin/utils/findByTicker.ts
```diff
@@ -5,5 +5,19 @@ type FindByTickerInput<T extends Coin> = {
...
-  coins.find(c => c.ticker === ticker) ?? null
+// SDK2-02 (audit r2): a bare ticker like "USDC" exists on many chains (Ethereum/Polygon/Arbitrum/Base/…).
+// The old `coins.find(c => c.ticker === ticker)` returned the ARRAY-ORDER first match — if this ever got
+// wired into a fund path, "USDC" would silently resolve to whichever chain happened to be first, sending to
+// the wrong network. Refuse to guess: return the unique match (or null when absent), and THROW when the
+// ticker is ambiguous across more than one chain so the caller must disambiguate by chain.
+export const findByTicker = <T extends Coin>({ coins, ticker }: FindByTickerInput<T>): T | null => {
+  const matches = coins.filter(c => c.ticker === ticker)
+  if (matches.length === 0) return null
+  const chains = new Set(matches.map(c => c.chain))
```
NeOMakinG, line 16: `suggestion:` ambiguity is keyed on *distinct chain count*, so two coins sharing this ticker on the **same** chain (different `id`/contract — scam/bridged/autodiscovered dup) fall through to the array-first `matches[0]` at L22 — the silent-first-match this PR closes, narrowed to within-chain. The built-in `coins` registry has zero such dups today, but `findByTicker` takes an arbitrary `coins: T[]`. Non-blocking (cross-chain is the SDK2-02 scope + zero callers); consider keying on `chain`+`id` when you next touch this.

packages/core/chain/coin/utils/findByTicker.ts
```diff
@@ -5,5 +5,19 @@ type FindByTickerInput<T extends Coin> = {
...
+  if (chains.size > 1) {
+    throw new Error(
```
NeOMakinG, line 18: `q:` untyped `new Error` — a caller wanting to catch-and-disambiguate has only string-matching to go on. If this gets wired into a fund path, a typed `AmbiguousTickerError` would let callers scope programmatically. Fine to defer.

NeOMakinG (approved):
`findByTicker` correctly converts a silent array-order first-match into a fail-closed throw on cross-chain ambiguity. Load-bearing checks all pass and the blast radius is nil — this export has **zero callers** anywhere in the SDK and isn't re-exported from any barrel, so the new throw cannot regress a live path. Preventive hardening done right. **Approving.**

## testing approach
- Traced the full function: `matches = filter(ticker) → 0 ⇒ null; distinct-chain count > 1 ⇒ throw; else matches[0]`. Fail-closed on the ambiguous case, no over-throw on single-chain.
- Enumerated **all** `findByTicker` references in the SDK (`grep -rn`): only the function itself + its new test. Not exported from any `index.ts` barrel; published only as an (unreferenced) `@vultisig/core-chain` subpath. **No caller can crash** — (4) mainline-regression risk is empirically zero.
- Verified `Coin.chain` is statically **required** (`ChainEntity<T> = { chain: T }`), so the `new Set(matches.map(c => c.chain))` ambiguity key can't collapse on an undefined chain.
- Confirmed the three test branches (single-chain resolves, absent → null, 3-chain ambiguous → throws `/ambiguous across 3 chains/`) match the source.
- Data check on the static `coins` registry (chainFeeCoin + knownTokens) grouped by `chain::ticker`: **zero same-chain duplicate tickers** today — so the residual below is defense-in-depth, not a live bug.
- Changeset present (`@vultisig/sdk` patch). CI: Lint+TypeCheck ✅, Integration (Vitest, runs this test) ✅, CodeRabbit ✅; Unit Tests still churning (slow job, non-blocking — the new spec is covered by the green Vitest job).

## suggestion
- **Same-chain ambiguity is left silent.** Ambiguity is measured by *distinct chain count*, so two coins sharing a ticker on the **same** chain (different `id`/contract — scam token, bridged variant, autodiscovered dup) collapse to `chains.size === 1` and return the array-first `matches[0]` — the exact silent-first-match this PR is closing, just narrowed to within-chain. Empirically the built-in `coins` registry has zero such dups, but `findByTicker` accepts an **arbitrary** `coins: T[]` (e.g. autodiscovery output), so a caller could still supply them. Non-blocking given the cross-chain scope of SDK2-02 + zero callers; consider keying ambiguity on distinct coin identity (`chain`+`id`) or throwing on `>1` match with differing `id` when you next touch this.

## q
- The disambiguation throw is an untyped `new Error`. A future caller that wants to catch-and-scope has only string-matching to distinguish it from other failures. If this gets wired into a fund path later, a typed SDK error (e.g. `AmbiguousTickerError`) would let callers disambiguate programmatically. Fine to defer.
- `patch` semver for a function that could previously never throw is technically a behavior change — defensible here (signature unchanged, zero consumers), just noting it.

## risk
Low. Message/behavior-only change to an unreferenced util; fail-closed direction (throw over silent wrong-chain); required `chain` field guarantees the ambiguity key is sound; changeset + CI green. No signable/tx path touched today.

## verdict
**Approve.** Throws on cross-chain ambiguity (fail-closed), unambiguous single-chain still resolves, zero callers so no mainline regression, changeset present, CI green. The two notes are non-blocking follow-ups.

---
🤖 dual-lane review (Claude load-bearing + Codex gpt-5.4 adversarial corroboration — Codex independently confirmed `Coin.chain` required and data-verified zero same-chain duplicate tickers in the registry).

PR #1096, opened 2026-07-08
fix(utxo): add coin selection between getUtxos and buildUtxoSendTx (utxo-01)

### LINKED ISSUE #1053: [audit-UTXO-01] No coin selection — every UTXO send consumes the entire wallet's UTXO set as inputs
## Context (fund-safety audit round 2, 2026-07-08 · finding UTXO-01 · High)

The app fetches every UTXO (utxoTx.ts:86) and passes the full array to buildUtxoSendTx (:106), whose SDK impl uses inputs = opts.utxos verbatim — despite the type's own doc saying 'caller handles coin-selection.' Verified against the shipped @vultisig/sdk@2.18.2 dist.

## Impact
Every BTC/LTC/DOGE/BCH/DASH/ZEC send from an address with N unspent outputs (a) pays fees for all N inputs (massive overpay), (b) can throw a false 'insufficient funds' when fee(all N) > balance though a 1-2 input send would clear, and (c) links all UTXOs on-chain every send.

## Anchors
`utxoTx.ts:86,106` (app) · buildUtxoSendTx SDK impl (`inputs = opts.utxos`).

## Fix
Add coin selection between getUtxos and build (e.g. accumulate smallest-sufficient set covering amount + fee, avoiding the fee-recalc infinite loop). Decide the selection strategy with the owner.

SDK-CLI: unit-drive buildUtxoSendTx (or the app util) with a >2-UTXO wallet fixture; BEFORE all N inputs are consumed, AFTER only the selected subset. Assert fee scales with selected inputs, not N.

## Prompt for AI agents
In vultisig-sdk (+ the app's utxoTx.ts:86/106 call site), add UTXO coin-selection between fetching UTXOs and buildUtxoSendTx (which currently uses inputs=opts.utxos verbatim). Implement a standard accumulative selection (sort, pick smallest-sufficient set covering amount+fee, add change), carefully avoiding the classic fee-recalc loop (recompute fee as inputs are added). Add unit tests with a multi-UTXO wallet asserting only the needed inputs are used and no false insufficient-funds. Runtime-verify via SDK-CLI. This is HIGH effort/risk — coordinate the strategy with the owner.

---
_Audit finding **UTXO-01** (round 2, High) · bead `agent-backend-ts-2s2zy` · runtime before+after is the finish line, not unit-tests-green. When you open a PR, put `closes #<this issue>` in the body._

gomesalexandre:
closes #1053

## The finding (UTXO-01, P1 HIGH)

`buildUtxoSendTx`'s own doc comment always claimed "caller handles coin-selection" but no selection layer existed anywhere in the SDK. `utxoTx.ts` (app) + the SDK path fetch EVERY UTXO for an address and pass the full array through verbatim as `inputs`. Effects:
- overpays fees for all N inputs when a handful would cover the send
- FALSE "insufficient funds" when fee(N) > balance even though a subset would cover it
- links every UTXO the wallet owns into one tx (privacy)

Not a fund-redirect (destination is always correct) — a fee/correctness/privacy HIGH.

## The fix

New module `packages/sdk/src/chains/utxo/select.ts` exporting `selectUtxoInputs`, inserted as a selection layer between `getUtxos` and `buildUtxoSendTx` — callers fetch UTXOs, run `selectUtxoInputs`, then pass `result.inputs` into `buildUtxoSendTx` unchanged.

**Algorithm: accumulative / largest-first (greedy).** Sort candidate UTXOs descending by value, accumulate until `selectedTotal >= amount + fee(k)`, recomputing the fee on every step (`estimateUtxoTxFee`, extracted out of `buildUtxoSendTx`'s inline fee calc — see below). Chosen over branch-and-bound for v1: it's the simplest correct strategy, minimizes input count (fewer inputs = lower fee + fewer signatures), and is the lowest-risk change to reason about + test across 6 chains with materially different fee/dust constants. Documented in the module's header comment as a deliberate v1 tradeoff.

**Fee agreement (selection <-> build):** `estimateUtxoTxFee(chain, inputCount, feeRate, opReturnData)` is extracted verbatim out of `buildUtxoSendTx`'s inline calc (same formula, same ZIP-317 Zcash floor) — a small, behavior-preserving refactor of `tx.ts` (8 lines removed, replaced with a single call). Both `selectUtxoInputs` and `buildUtxoSendTx` now call the SAME function, so they can never disagree on the fee for a given input count.

**Dust:** intentionally NOT special-cased in the selection layer — `buildUtxoSendTx`'s `serializeOutputs` already folds a change output below the chain's `dustLimit` into the fee (`hasChange = change > dustLimit`), so a selection landing with `0 < change <= dustLimit` still produces a valid tx, just at a marginally higher effective fee. Verified in golden vectors (case 4 below).

**Insufficient funds:** throws the exact same `insufficient funds: have=X need=Y (amount=A fee=F)` shape `buildUtxoSendTx` throws — verified byte-identical in tests.

**Send-max:** `sendMax: true` bypasses the greedy walk and selects every UTXO regardless of whether a smaller subset would cover a smaller `amount` — preserves "send whole balance" flows.

## Coordination with #1089

#1089 (`fix_utxo_min_fee_floor_zip317_utxo0304`) is in-flight on the same file family (`tx.ts`) and was still OPEN/unmerged as of this push. Per the coordination plan: the new selection logic lives entirely in a NEW file (`select.ts`), and the only edit to `tx.ts` is the minimal fee-calc extraction (no other changes to `buildUtxoSendTx`'s internals, no P2SH/scriptType-validation reordering). If #1089 merges first, the rebase conflict surface should be limited to that one extracted block.

## Golden-vector coverage (all 6 UTXO chains, parametrized)

`packages/sdk/tests/unit/chains/utxo-coin-selection.test.ts` — `describe.each` over Bitcoin / Litecoin / Dogecoin / Dash / Bitcoin-Cash / Zcash, each with 8 tests:
1. single UTXO covers amount+fee -> selects just it (no regression to the trivial case)
2. many small UTXOs -> accumulates the minimal covering subset, fee recalculated as `k` grows, verified against `estimateUtxoTxFee(chain, k, feeRate)` at both `k` and `k-1` (minimality)
3. exact-cover (selected === amount+fee) -> zero change, zero-output-count verified via `buildUtxoSendTx`'s real `unsignedRawHex`
4. change below dust -> folded into fee, no dust-value output (same real-wire-format check)
5. genuinely insufficient funds -> identical error message from `selectUtxoInputs` and `buildUtxoSendTx`
6. send-max -> selects every UTXO even when a smaller subset would cover a smaller amount
7. + 8. empty-UTXO-set / non-positive-amount guards mirror `buildUtxoSendTx`'s own messages

Plus a standalone "selection <-> build agreement" test that underfunds by exactly 1 base unit per chain and asserts `buildUtxoSendTx`'s reported `fee=` matches `estimateUtxoTxFee`'s standalone output.

The output-count assertions parse `buildUtxoSendTx`'s own `unsignedRawHex` (the real production pre-signing wire format — version + inputs + outputsWithCount, identical layout across all 6 chains including Zcash) rather than reimplementing parsing logic, so a regression in `buildUtxoSendTx` itself would be caught here too.

## Test output

```
$ npx vitest run tests/unit/chains/utxo-coin-selection.test.ts
 Test Files  1 passed (1)
      Tests  49 passed (49)

$ npx vitest run tests/unit/chains   # full existing utxo suite, unaffected
 Test Files  14 passed (14)
      Tests  181 passed (181)

$ npx vitest run tests/unit         # full sdk unit suite
 Test Files  172 passed | 2 failed (174)
      Tests  2454 passed | 2 failed (2456)
```
The 2 unrelated failures (`ensureMpcEngine.test.ts`, `platforms/react-native/entry.test.ts`) are a pre-existing MPC-engine-singleton test-isolation issue — reproduced identically on a clean `origin/main` checkout with zero diff applied, confirmed before opening this PR.

```
$ yarn workspace @vultisig/sdk typecheck   # clean, no output
$ yarn check:shared-exports                # "Shared package exports are up to date"
$ npx eslint ...                           # clean, no output
$ npx prettier --check ...                 # unchanged
```

## receipts

BE-only pure-function change (coin selection over an array of UTXOs), no UI/device surface to drive — the golden-vector suite above (which builds real, chain-specific unsigned transactions via `buildUtxoSendTx` and parses their real wire-format output count) is the closest equivalent to a runtime receipt for this kind of change. No runtime UI changes.

🤖 Agent-generated

## Summary by CodeRabbit

* **New Features**
  * Added smarter UTXO send input selection that chooses the smallest set of coins needed for a payment.
  * Added a send-max option to spend the full available balance in one transaction.
  * Exposed fee estimation for UTXO sends so transaction costs can be calculated more accurately.

* **Bug Fixes**
  * Reduced the chance of overpaying fees when many inputs are available.
  * Avoided false “insufficient funds” errors when a smaller coin set is enough.
  * Improved privacy by preventing unnecessary use of every UTXO in a single send.


packages/sdk/src/chains/utxo/tx.ts
```diff
@@ -547,6 +547,35 @@ function buildOpReturnScript(data: Uint8Array): Uint8Array {
...
+): bigint {
+  const spec = UTXO_SPECS[chain]
+  if (!spec) throw new Error(`unsupported UTXO chain: ${chain as string}`)
+  const opReturnScript =
+    opReturnData !== undefined ? buildOpReturnScript(new TextEncoder().encode(opReturnData)) : undefined
+  // Approximate tx size for fee calc — matches app's heuristic.
+  const bytesPerInput = spec.scriptType === 'p2wpkh' ? 68 : 150
+  // 8-byte value + varint(scriptLen) + scriptLen; scriptLen <= 82 so the varint is 1 byte.
+  const opReturnBytes = opReturnScript ? 9 + opReturnScript.length : 0
+  const txSize = inputCount * bytesPerInput + 2 * 34 + 10 + opReturnBytes
```
NeOMakinG, line 573: `should-fix:` `estimateUtxoTxFee` prices `2 * 34` output bytes (recipient + change) unconditionally, but `serializeOutputs` emits only 1 output on exact-cover, sub-dust-folded, and send-max no-change txs. This is pre-existing in `buildUtxoSendTx` and **safe-direction** (overpay / conservative reject — never under-funds or loses change), and the selector correctly mirrors it so selection and build agree (a smaller no-change fee here would make them *disagree* and re-reject at build — worse). So **non-gating**. But it creates a narrow false-"insufficient funds" window of `~34 * feeRate` base units at the no-change boundary: a wallet holding exactly enough for the real 1-output tx is rejected. Follow-up: make the fee change-aware (mind the circular fee↔change dependency), or document the conservative over-estimate.

NeOMakinG (approved):
Reviewed as a fund-critical UTXO coin-selection change. Two lanes: Claude (load-bearing static trace of every fund-safety path) + Codex `gpt-5.4` adversarial second-opinion (verified on-PR — cited `select.ts:116` / `tx.ts:569,587,830` with correct content, not cross-contaminated/stale).

## testing approach
- Full static trace of `selectUtxoInputs` (sufficiency, fee-input circularity, dust, UTXO hygiene, per-chain) + verification that the `estimateUtxoTxFee` extraction out of `buildUtxoSendTx` is byte-identical to the prior inline calc.
- Confirmed the change output is unchanged by this PR: `buildUtxoSendTx` → `serializeOutputs` emits `change` to `fromScript` (the sender's **own** address), and folds sub-dust change into the fee (`hasChange = change > dustLimit`). No change is silently lost beyond the standard sub-dust fold.
- Local vitest blocked by the known `@noble/curves/secp256k1.js` pnpm nested-workspace resolution artifact (same infra issue seen on #1081/#1089 — not a code defect). Relied on **CI all-green** (Unit / Integration / Lint+TypeCheck / Test Summary / CodeRabbit all SUCCESS) + hand-verification.

## fund-safety verdict (the checks that matter)
- **sufficiency / fail-closed** ✅ — greedy loop accumulates until `selectedTotal >= amount + fee(k)`; if the full set can't cover, it throws the byte-identical `insufficient funds:` shape. Never returns an under-funded set.
- **change correctness** ✅ — selection only picks inputs; the change output is created by the unchanged `buildUtxoSendTx`, goes to `fromAddress` (own address), and since selection and build share `estimateUtxoTxFee` they compute the identical `change = selectedTotal − amount − fee`. No change lost to a miscalc.
- **fee-input circularity** ✅ — `fee(k)` is recomputed on every greedy step, so the covering threshold grows with input count and converges. Largest-first ordering naturally avoids uneconomical dust inputs (loop returns before reaching them).
- **dust** ✅ — not special-cased in selection by design; sub-dust change is folded to fee by `serializeOutputs` (output count = 1, verified in golden case 4). No dust output emitted.
- **UTXO hygiene** ✅ — each candidate pushed once per call; sort is stable/deterministic. (De-duping the input array is the caller's/`getUtxos`'s responsibility — unchanged.)
- **reaches signable** ✅ — caller passes `result.inputs` verbatim to `buildUtxoSendTx`; fee agreement guarantees no post-selection re-reject.
- **purely additive** — nothing in `packages/sdk/src` calls `selectUtxoInputs` yet; the only behavior change to existing code is the byte-identical fee extraction. Zero regression surface.

## should-fix
See inline on `tx.ts:573` — the fee model prices `2 * 34` output bytes even when the tx emits only 1 output (exact-cover / sub-dust-folded / send-max no-change). This is **pre-existing** in `buildUtxoSendTx` and **safe-direction** (overpay / conservative reject, never under-fund), and the selector *correctly* mirrors it to preserve selection↔build agreement — so it is **not gating**. But it opens a narrow false-"insufficient funds" window (`~34 * feeRate` base units) at the no-change boundary: a wallet holding exactly enough for the real 1-output tx is told it can't send. Follow-up: make the fee change-aware (needs care — fee↔change is circular), or document the conservative bound.

## q
The new golden-vector tests derive their exact-cover / dust fixtures from `estimateUtxoTxFee` itself, so they prove selection↔build *consistency* but not that every genuinely-spendable no-change tx is accepted — they can't catch the boundary above (Codex converged). A test that funds exactly `amount + realOneOutputFee` would surface it. Non-gating.

## risk
Low. Additive module + byte-identical refactor; change-output and dust handling untouched; fail-closed on insufficiency; CI green across all suites. The one substantive finding is a pre-existing safe-direction over-estimate, not a fund-loss or under-fund path.

## verdict
**APPROVE.** Coin selection is sufficient and fail-closed, change is computed correctly to the sender's own address (never lost to fee), fee-input circularity converges, dust is folded (no dust output), UTXO hygiene holds, and the selected set is exactly what gets signed. The sole finding is a non-gating, pre-existing, safe-direction fee over-estimate worth a follow-up.

<sub>🤖 dual-lane review (Claude load-bearing + Codex gpt-5.4 adversarial). Fund-safety lens on UTXO coin selection.</sub>

PR #1098, opened 2026-07-08
fix(kyber): bind swap tx.to to /route/build's own routeraddress (agg-04)

### LINKED ISSUE #1059: [audit-AGG-04] KyberSwap tx.to binds to the stale /routes routerAddress, not the paired /route/build resp
## Context (fund-safety audit round 2, 2026-07-08 · finding AGG-04 · Medium)

kyber/api/tx.ts:120 destructures data but never reads data.routerAddress; the return at :127 uses the outer routerAddress from the earlier separate /routes call. Live probe shows they currently match, but relying on an address from an earlier call is latent contract-drift risk. One-line fix: use buildResponse.data.routerAddress.

## Effort / risk / confidence
Effort SRisk LOWConfidence HIGH (gap) / MED (currently identical)

## Prompt for AI agents
Implement the fix for audit finding **AGG-04** in `vultisig/vultisig-sdk`. kyber/api/tx.ts:120 destructures data but never reads data.routerAddress; the return at :127 uses the outer routerAddress from the earlier separate /routes call. Live probe shows they currently match, but relying on an address from an earlier call is latent contract-drift risk. One-line fix: use buildResponse.data.routerAddress.  Make the minimal, additive, fail-closed change at the anchors named above; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI: build the SDK and unit/CLI-drive the affected path with a fixture; assert before (bug) vs after (fixed).). **Grep open PRs in this repo first** — several audit findings already have in-flight PRs; advance/verify an existing one (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **AGG-04** (round 2, Medium) · bead `agent-backend-ts-cyrix`. Runtime before+after is the finish line, not unit-tests-green._

gomesalexandre:
closes #1059 (audit AGG-04).

## what
KyberSwap `tx.to` used the `routerAddress` from the earlier `/routes` call, not `/route/build`'s own `routerAddress` field — the router paired with the `data` (calldata) this call just generated.

## risk — already largely mitigated by AGG-02 (#1079)
The merged AGG-02 fix asserts both routers are known-Kyber AND **throws if `/route` and `/route/build` disagree** (equality guard), so a mismatched-calldata tx can't be built today. This change is therefore a **provable no-op** — but it binds `tx.to` to the calldata's own router authoritative-by-construction, as defense-in-depth if that equality guard is ever moved/removed.

## test
Behaviorally a no-op (equality guard proves `buildRouterAddress === routerAddress`), so the existing kyber suite is unchanged; CI runs the full `build:shared` + vitest suite as the authoritative gate. Local cross-package resolution needs build:shared (skipped locally, provable-equivalence rationale above).


NeOMakinG (approved):
## testing approach

Two-lane review (Claude load-bearing + Codex adversarial) against a unique worktree at HEAD `662223f1`. Traced the `tx.to` binding before/after, the router-trust model and its interaction with the #1079 (AGG-02) allowlist, quote-vs-build router consistency, fail-closed behaviour, and that the bound `to` reaches the signable EVM tx. Polled CI to green. Codex lane stalled out (8min buffering timeout — the repeat gpt-5.4 pattern seen on #1012/#1075/#1079); non-load-bearing, verdict rests on the Claude lane.

## the change

One-line binding swap in `kyber/api/tx.ts`: the swap tx `to` was set from `routerAddress` (the param threaded in from the `/routes` quote endpoint via `quote.ts` → `getKyberSwapRoute`) and now comes from `buildRouterAddress` (the `routerAddress` field the `/route/build` response *itself* returned — the router the `data` calldata was generated against).

## fund-safety verdict — CLEAN

- **tx.to bound to /route/build's own router**: confirmed. Before = `routerAddress` (`/routes` param); after = `buildRouterAddress` (`buildResponse.data.routerAddress`, the same response that produced `data`). The `-1` in the diff removed `to: routerAddress`.
- **No allowlist bypass (the load-bearing question)**: `buildRouterAddress` is validated by `assertKnownAggregatorRouter('kyber', buildRouterAddress, from.chain)` at `tx.ts:131` — the same #1079 fail-closed allowlist — *before* the `GeneralSwapQuote` is constructed. A MITM'd `/route/build` response carrying an attacker router is rejected ("unrecognized router address"), never reaching `tx.to`. Binding to the response router is safe precisely because that router is allowlist-gated.
- **Quote-vs-build consistency**: `tx.ts:132-136` cross-checks `buildRouterAddress.toLowerCase() === routerAddress.toLowerCase()` and throws on divergence. Build router cannot differ from the quoted+allowlisted router; both must equal the single canonical Kyber MetaAggregationRouterV2. So the AGG-04 swap is behaviour-preserving today (the two are proven equal) and hardens `tx.to` to be authoritative-by-construction if that guard ever moves — exactly as the comment claims.
- **Fail-closed**: both asserts throw; empty router → "unrecognized"; mismatch → throw; all before quote construction. The `attempt()`/TransferHelper retry in `quote.ts` re-invokes `getKyberSwapTx`, re-running every guard — a router rejection is not a TransferHelper error so it propagates, not retried-around.
- **Reaches signable**: `quote.tx.evm.to` flows to `build.ts` general resolvers (approve-spender via `getSwapDestinationAddress` and swap-dest via the EVM SigningInput) — same path verified load-bearing on #1079, asserted by `build.agg02RouterCoverage.test.ts`.
- **No over-reject**: a legit Kyber swap returns the canonical router in both fields → still builds. The change only swaps which of two proven-equal values feeds `to`; zero new rejection surface.
- **Changeset + CI**: `.changeset/agg04-kyber-build-router.md` present (`@vultisig/sdk` patch). All checks green — Unit, Integration (Vitest), Lint/TypeCheck, Test Summary, CodeRabbit.
- **Tests**: the AGG-02 suite in `quote.test.ts` covers reject-non-canonical, accept-canonical, and reject-when-build-router-differs — exercising the `buildRouterAddress` allowlist path.

## risk

Minimal. Pure defense-in-depth on top of the merged #1079 allowlist; behaviour-identical on the happy path (the two routers are proven equal), strictly safer if the equality guard is ever refactored away.

## verdict

Approve. `tx.to` is bound to `/route/build`'s own `routerAddress`, that router is allowlist-validated (no bypass) and cross-checked equal to the quoted router, fail-closed on missing/mismatch, reaches the signed tx, and does not over-reject legit swaps. Clean AGG-04 sibling to #1079/#1082.

<sub>🤖 reviewed by the vultisig-ops reviewer agent (Claude lane load-bearing; Codex lane timed out, non-load-bearing).</sub>

PR #1100, opened 2026-07-09
fix(evm): guard capgaslimit against a zero-source bigintmax throw (sdk2-03)

gomesalexandre:
Part of #1066 (audit SDK2-02·SDK2-03·BE-05 cluster — this is the SDK2-03 portion; references, does NOT close the cluster since BE-05 abts#944 is still pending). Defensive guard: capGasLimit's bigIntMax(...without([...],undefined)) throws opaquely if all gas sources are undefined (unreachable today, minimumGasLimit always supplied) — now throws a clear actionable error, can't silently fall through. Additive on an impossible-today branch; existing getEvmFeeQuote suite unaffected (CI runs it).


NeOMakinG (approved):
Reviewed as fund-relevant (EVM signable gas-limit path). Two-lane review: Claude (primary, load-bearing) + Codex `gpt-5.4` (adversarial second opinion). Both lanes converge — no fund-safety regression, correct fail-closed choice. **Approving.**

## testing approach
Traced the full `capGasLimit` closure and its one caller. Verified helper semantics (`bigIntMax`, `without`, `deriveEvmGasLimit`) against source in a clean worktree on HEAD 75e6d32b. Did not run device QA — this is a defensive guard on an unreachable-today branch with byte-identical behavior on every reachable path (nothing to exercise at runtime that differs from `main`). CI is authoritative here and all green.

## what the fix actually does
The framing "guard prevents the throw" is slightly off vs the implementation: the fix does **not** fabricate a fallback gas limit — it **replaces an opaque zero-arg `bigIntMax` throw** (`"No arguments provided to bigIntMax"`) **with a clear, actionable throw** (`"capGasLimit: no gas-limit source available…"`). On a signable gas-limit path this **fail-closed** choice is the correct one: there is no way to safely fabricate a gas limit without real estimation data (a fabricated low value → out-of-gas / funds burned to gas; a high one → overpay). Aborting the tx build cleanly is strictly safer than inventing a number.

## verification (all pass)
- **No new gas bug:** the guard produces no gas limit — it throws. No 0-limit, no absurd value introduced. ✅
- **Fail-safe direction:** fail-closed (tx build aborts) rather than silent-wrong-value. ✅
- **Reaches signable tx unchanged:** normal path returns `bigIntMax(...gasLimitSources)` (± data buffer) → `evm.ts:64` `gasLimit.toString()` → `EthereumSpecificSchema.gasLimit`. Untouched. ✅
- **No regression:** `without` only strips `undefined`, so for any non-empty source set the array — and thus `bigIntMax(...)` output — is byte-identical to pre-PR. The only behavioral delta is the empty-source branch, which *already threw* before (opaque) and *still throws* now (clear) — same failure **mode**, so no caller regresses. ✅
- **Branch genuinely unreachable today:** the sole caller (`chainSpecific/resolvers/evm.ts:48`) always passes `minimumGasLimit: deriveEvmGasLimit({coin, data})`, and `deriveEvmGasLimit` returns a concrete `bigint` for every `EvmChain` (both `feeCoinTransferGasLimit` and `erc20TransferGasLimit` are complete `Record<EvmChain, bigint>` maps; the `data` branch returns literals). The guard defends a *future/external* caller of the exported `getEvmFeeQuote` that omits `minimumGasLimit` while estimation also fails. The comment is accurate. ✅
- **Scope:** +16/-3, 2 files = changeset (`@vultisig/sdk` patch) + the guard. Nothing unrelated. ✅ CI (Unit, Integration/Vitest, Lint+Type, Test Summary, CodeRabbit) all green.

## suggestion
- `suggestion:` The new fail-closed branch is unpinned by any test, so a future refactor could silently reintroduce a fabricated fallback or alter the thrown-error surface undetected (Codex second-lane concurred). A test is feasible via the exported API: call `getEvmFeeQuote` with `minimumGasLimit` omitted and estimation forced to `undefined` → assert the clear throw; plus a non-empty case asserting equivalence to the max source. Non-blocking — the branch is unreachable on the in-repo path and the change is byte-identical everywhere else.

## risk
Minimal. Defensive-only hardening; no reachable-path behavior change; fund-safety posture strictly improved (fail-closed on missing estimation data). Same protective family as #1078 (zkSync `bigIntMax` floor) which is already merged.

## verdict
**APPROVE.** Correct, minimal, fail-closed guard on a signable EVM gas-limit path. No regression, no new gas bug, CI green, changeset present.

---
🤖 Two-lane review (Claude primary + Codex gpt-5.4 adversarial). Generated with [Claude Code](https://claude.com/claude-code).

PR #1101, opened 2026-07-09
fix(sdk): validate msgdeposit memo + surface non-integer dstamount drops (sdk-correctness-04/06/08)

### LINKED ISSUE #1071: [audit-SDK-CORRECTNESS-04 · 06 · 08] Silent provider-drop on non-integer dstAmount; unvalidated THORChain MsgDeposit memo; a
## Context (fund-safety audit round 1, 2026-07-08 · finding SDK-CORRECTNESS-04 · 06 · 08 · Low)

findSwapQuote.ts:213-218 runs BigInt(dstAmount) inside Promise.allSettled; a non-integer string from a drifted provider silently drops that quote from ranking (user routed to a worse-but-valid fill with no signal it was a parse failure). thorchainMsgDeposit.ts:49-103 passes the memo through with no structural validation, unlike the fully-validated limit-swap memo path — a malformed agent-supplied memo signs a value-bearing deposit that misroutes RUNE/CACAO (partly by-design). And thorchainMsgDeposit.ts:97 uses value as any on the deposit's chain-specific proto binding.

## Effort / risk / confidence
Effort S-MRisk LOW-MEDConfidence LOW-MED — investigate

## Prompt for AI agents
Implement the fix for audit finding **SDK-CORRECTNESS-04 · 06 · 08** in `vultisig/vultisig-sdk` at the anchors named above. findSwapQuote.ts:213-218 runs BigInt(dstAmount) inside Promise.allSettled; a non-integer string from a drifted provider silently drops that quote from ranking (user routed to a worse-but-valid fill with no signal it was a parse failure). thorchainMsgDeposit.ts:49-103 passes the memo through with no structural validation, unlike the fully-validated limit-swap memo path — a malformed agent-supplied memo signs a value-bearing deposit that misroutes RUNE/CACAO (partly by-design). And thorchainMsgDeposit.ts:97 uses value as any on the deposit's chain-specific proto binding. Minimal, additive, fail-closed; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI build + unit/CLI drive; before vs after.). **Grep open PRs in this repo first** — advance an existing PR (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **SDK-CORRECTNESS-04 · 06 · 08** (round 1, Low) · bead `agent-backend-ts-rs7zk`._

gomesalexandre:
closes #1071

## Context

Fund-safety audit round 1, finding **SDK-CORRECTNESS-04 · 06 · 08** (bead `agent-backend-ts-rs7zk`).

## What changed

**SDK-CORRECTNESS-04 — silent provider-drop on non-integer dstAmount** (`packages/core/chain/swap/quote/findSwapQuote.ts`)
`getComparableOutputAmount` ran `BigInt(dstAmount)` inside `Promise.allSettled` — a non-integer string from a drifted provider silently dropped that quote from ranking with no signal it was a parse failure rather than a genuine decline (the user got routed to a worse-but-valid fill and nobody could tell). It now `console.warn`s the provider name + raw value before rethrowing. Ranking behavior is unchanged (the fetcher still rejects, `Promise.allSettled` still drops it from candidates) — this only makes the drop observable.

**SDK-CORRECTNESS-06 — unvalidated THORChain MsgDeposit memo** (`packages/sdk/src/tools/prep/thorchainMsgDeposit.ts`)
`prepareThorchainMsgDepositTxFromKeys` accepted an arbitrary pre-built memo string with zero structural validation, unlike the fully-validated limit-swap memo path (`buildLimitSwapMemo`/`validateLimitSwapInputs`) — a malformed or garbage memo would sign a value-bearing `MsgDeposit` that THORNode either rejects (wasted fee) or misinterprets. New `assertValidThorchainDepositMemo` (`packages/core/chain/swap/native/thorchainDepositMemo.ts`) fails closed on:
- non-printable-ASCII / leading-trailing-whitespace memos
- oversized memos (reuses `assertMemoByteLength(memo, 'other')` from the limit-swap path — MsgDeposit's source chain is always THORChain/MayaChain itself, never UTXO)
- an unrecognized action prefix (mirrors THORNode's memo grammar: add/withdraw/donate/bond/unbond/leave/migrate/noop/consolidate/name/trade±/pool±/loan±/secure±)
- for the two LP actions the docstring documents (`+:POOL[:PAIRED]`, `-:POOL:BPS[:ASSET]`), a malformed pool id (reuses `assertValidPoolId`)

Deliberately does **not** implement full per-action argument grammar (e.g. BOND's provider/fee args) — prefix + character-safety + the two well-known LP shapes, so legitimate operator-style memos still pass. Swap memos (`=`/`SWAP`/`s`) are excluded from the recognized set since native swaps route through the server-issued memo in `getNativeSwapQuote`, never through this generic deposit builder.

**SDK-CORRECTNESS-08 — value-bearing field gap** (same file, line 97)
`keysignPayload.blockchainSpecific = { case: ..., value: value as any }` picked a resolver *reference* dynamically (`getThorchainChainSpecific` vs `getMayaChainSpecific`) and force-cast the result into the oneof, so a future refactor that mismatched the `case` string against the resolver actually called would compile silently. Replaced with per-chain branches so each branch's literal `case` and its resolver's return type are statically paired by the `KeysignPayload` oneof — no cast needed, and a mismatch now fails `tsc`.

## Dedup

Grepped open PRs + issue #1071 body before starting. #1096 (utxo coin selection) and #1092 (vault-free swap expiry) are adjacent SDK-CORRECTNESS findings but cover different anchors — no overlap with 04/06/08.

## Receipts

```
$ yarn vitest run --config .config/vitest.core.config.ts packages/core/chain/swap/quote
 Test Files  8 passed (8)
      Tests  95 passed (95)

$ yarn vitest run --config .config/vitest.core.config.ts packages/core/chain/swap/native/thorchainDepositMemo.test.ts
 Test Files  1 passed (1)
      Tests  18 passed (18)

$ yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/tools/prep/thorchainMsgDeposit.test.ts
 Test Files  1 passed (1)
      Tests  11 passed (11)

$ yarn test:core            # full core-chain suite, no regressions
 Test Files  182 passed | 1 skipped (183)
      Tests  1598 passed | 1 skipped (1599)

$ yarn workspace @vultisig/sdk test:unit   # full sdk suite, no regressions
 Test Files  174 passed (174)
      Tests  2460 passed (2460)

$ yarn workspace @vultisig/sdk typecheck   # clean
$ npx tsc --project .config/tsconfig.json --noEmit   # clean
$ yarn check:shared-exports   # up to date
```

Changeset included (`@vultisig/core-chain` + `@vultisig/sdk`, patch).

🤖 Agent-generated


packages/sdk/src/tools/prep/thorchainMsgDeposit.ts
```diff
@@ -64,6 +65,7 @@ export const prepareThorchainMsgDepositTxFromKeys = async (
   if (!memo) {
     throw new Error('prepareThorchainMsgDepositTxFromKeys: memo is required')
   }
+  assertValidThorchainDepositMemo(memo)
```
NeOMakinG, line 68: should-fix (non-blocking, follow-up): this fail-closed guard is correct and reaches the signable memo, but it is the *only* THOR/Maya `MsgDeposit` builder that validates. Two siblings still sign an arbitrary memo verbatim: `buildThorchainDepositTx` (platforms/react-native/chains/cosmos/tx.ts:454 — exported, "MsgDeposit for LP ops", builds /types.MsgDeposit from opts.memo with no check) and the generic `prepareSignAminoTxFromKeys`/`prepareSignDirectTxFromKeys` path (accepts thorchain/MsgDeposit msgs, cosmos-msg.ts:14). Applying `assertValidThorchainDepositMemo` in `buildThorchainDepositTx` too would complete the SDK-CORRECTNESS-06 fail-closed property across the surface.

NeOMakinG (approved):
Fund-safety review of the three SDK-CORRECTNESS items. Two-lane (Claude load-bearing + Codex gpt-5.4 adversarial). The three changes are cleanly separable, each is sound and fail-closed, and none over-rejects a legitimate swap/deposit. **Approving** with one non-blocking coverage note.

## testing approach
- Worktree at HEAD `912cd612`, `yarn build:shared` + targeted vitest:
  - `thorchainDepositMemo.test.ts` — **18/18** pass
  - `findSwapQuote.selection.test.ts` — **52/52** pass (incl. the new SDK-CORRECTNESS-04 warn assertion)
  - `sdk/tests/unit/tools/prep/thorchainMsgDeposit.test.ts` — **11/11** pass
- CI @912cd612: Integration Tests pass, Lint and Type Check pass (14m — confirms the correctness-08 static `case`/`value` pairing compiles and the changeset is non-inert), CodeRabbit pass, Unit Tests re-running (local green).
- Traced each of the 3 items independently; Codex corroborated all four fund-safety focus questions.

**The 3 correctness IDs, separated & each verified:**
- **04 — non-integer `dstAmount` drop** (`findSwapQuote.ts` `getComparableOutputAmount`): `BigInt(dstAmount)` now wrapped in try/catch → `console.warn(provider, raw)` then **rethrow**. `dstAmount` here is *ranking-only* (comparable output for quote selection), not a signable field. The throw is contained inside `Promise.allSettled(fetchers.map(...))` (findSwapQuote.ts:852) → the malformed quote becomes one rejected element and is dropped; `findSwapQuote` survives, and if *every* provider drops, the all-fail path returns a no-route/below-min error — never a silently-floored wrong amount. Behavior unchanged, drop now observable. Fail-safe. ✅
- **06 — unvalidated MsgDeposit memo** (`thorchainDepositMemo.ts` + `thorchainMsgDeposit.ts:68`): new `assertValidThorchainDepositMemo` fails closed on empty / non-printable-ASCII / leading-trailing whitespace / oversized (`assertMemoByteLength(memo,'other')`=250B, correct since MsgDeposit source is always THOR/Maya, never UTXO) / unrecognized action prefix, plus pool-id for the two LP actions. Called **before** the keysignPayload is built (line 68, above the `blockchainSpecific` assignment) → no signable MsgDeposit constructible on a bad memo, and the validated `memo` is the exact value that lands in the signed msg. Composes as defense-in-depth with the app-side abts#506 empty-memo guard (SDK is the lower layer). ✅
- **08 — `as any` cast on the oneof** (`thorchainMsgDeposit.ts:97`): dynamic resolver-ref + force-cast replaced with per-chain branches so each literal `case` is statically paired with its resolver's return type. No runtime behavior change; a future resolver/case mismatch now fails `tsc`. ✅

**No over-reject (verified both lanes):** LP add/remove (`+`/`-`/`add`/`withdraw` with dot-notation uppercase pool), bond/unbond/leave/donate/migrate/noop/consolidate/name, and `trade±`/`pool±`/`loan±`/`secure±` all pass; non-LP memos keep extra args (operator-style BOND provider/fee args, `secure+:DEST:thor1affiliate:50` affiliate suffixes) since only LP actions validate the pool. SWAP (`=`/`s`/`SWAP`) is correctly excluded — I confirmed no RUNE-source swap routes through `prepareThorchainMsgDepositTxFromKeys`; native swaps build via `getNativeSwapQuote` → `core/mpc/keysign/swap/build.ts`, a separate builder. `signMsgDeposit` has no internal callers (public LP/deposit surface only).

## should-fix
The memo validation is applied to exactly **one** of the SDK's THOR/Maya `MsgDeposit` builders. Two sibling builders still accept an arbitrary memo verbatim and reach a signable value-bearing MsgDeposit, so the fail-closed property this PR establishes for SDK-CORRECTNESS-06 is not complete across the surface:
1. **`buildThorchainDepositTx`** (`packages/sdk/src/platforms/react-native/chains/cosmos/tx.ts:454`) — exported, documented "THORChain MsgDeposit for LP ops", builds `/types.MsgDeposit` straight from `opts.memo` with zero validation. This is purpose-identical to the builder being fixed, just on the RN signing path.
2. **`prepareSignAminoTxFromKeys` / `prepareSignDirectTxFromKeys`** (`tools/prep/cosmos.ts`) — vault-free / MCP-reachable, accept arbitrary `msgs` including `thorchain/MsgDeposit` (`types/cosmos-msg.ts:14`) with an unvalidated memo.
Neither is a regression (both pre-existed unvalidated) and #2 is a deliberate low-level escape hatch, but #1 in particular is an agent-reachable LP-deposit builder that should carry the same guard. Recommend a follow-up applying `assertValidThorchainDepositMemo` in `buildThorchainDepositTx` (and considering the generic cosmos path). Non-blocking for this PR's stated scope.

## risk
Low. The change is additive fail-closed validation on the targeted builder + an observability-only rethrow + a type-safety tightening. No signable amount/memo is loosened; the only behavioral change is rejecting malformed memos that would previously have signed a wasted/misinterpreted MsgDeposit. The residual risk is the two uncovered sibling builders above (pre-existing, tracked as should-fix).

## verdict
**Approve.** All three SDK-CORRECTNESS items are correctly scoped, fail-closed, reach the signable path, and don't over-reject legitimate deposits/swaps; changeset present (`@vultisig/core-chain` + `@vultisig/sdk` patch), tests green, CI green (Unit re-running). The sibling-builder coverage gap is a non-blocking follow-up.

---
🤖 Two-lane review: Claude (load-bearing, runtime vitest) + Codex gpt-5.4 (adversarial, exit 0, PR #1101-specific, not cross-contaminated — independently surfaced the RN `buildThorchainDepositTx` bypass).

PR #1104, opened 2026-07-09
fix(solana): assert swap instruction program allow-list before signing (sol-01)

### LINKED ISSUE #1056: [audit-SOL-01] No client-side instruction allow-list on a Jupiter-sourced swap tx before MPC-signing
## Context (fund-safety audit round 2, 2026-07-08 · finding SOL-01 · Medium)

jupiterFeeAta.ts:118 and solanaTx.ts:313 deserialize and forward a proxy-supplied VersionedTransaction straight to signing with no check that every programIdIndex resolves to an expected program (Jupiter router, System, Token/Token-2022, ATA) or that fund movement stays within the swap's own accounts. The only trust boundary is "the proxy returned this." Fix: a shared assertSafeSolanaSwapInstructions(message) used by both call sites, unit-tested against a captured real response.

## Effort / risk / confidence
Effort MRisk LOWConfidence MED (architecture gap, Vultisig controls the proxy)

vultiagent-app is ALWAYS sim: drive the affected flow to the pre-sign/broadcast surface; screenshot before/after.

## Prompt for AI agents
Implement the fix for audit finding **SOL-01** in `vultisig/vultisig-sdk`. jupiterFeeAta.ts:118 and solanaTx.ts:313 deserialize and forward a proxy-supplied VersionedTransaction straight to signing with no check that every programIdIndex resolves to an expected program (Jupiter router, System, Token/Token-2022, ATA) or that fund movement stays within the swap's own accounts. The only trust boundary is "the proxy returned this." Fix: a shared assertSafeSolanaSwapInstructions(message) used by both call sites, unit-tested against a captured real response.  Make the minimal, additive, fail-closed change at the anchors named above; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (vultiagent-app is ALWAYS sim: drive the affected flow to the pre-sign/broadcast surface; screenshot before/after.). **Grep open PRs in this repo first** — several audit findings already have in-flight PRs; advance/verify an existing one (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **SOL-01** (round 2, Medium) · bead `agent-backend-ts-pvlt1`. Runtime before+after is the finish line, not unit-tests-green._

gomesalexandre:
closes #1056

## What

Both Jupiter swap-build sites deserialized a proxy-supplied `VersionedTransaction` and forwarded it to MPC signing with **zero check** that each instruction's `programIdIndex` resolves to an expected program. A compromised Jupiter proxy (or a MITM'd `api.vultisig.com/jup` hop) could splice in an arbitrary instruction — e.g. a drain transfer to an attacker address — and the user would effectively blind-sign it, since the signing UI can't meaningfully render raw Solana instruction bytes for review.

## Fix

Add `assertSafeSolanaSwapInstructions` (`@vultisig/core-chain/chains/solana/assertSafeSolanaSwapInstructions`): resolves every **top-level** instruction's program against static account keys and, for v0 messages, address-lookup-table-resolved keys, and throws `SOL_SWAP_UNEXPECTED_PROGRAM: instruction N targets unrecognized program <pubkey>; refusing to sign` on the first unrecognized one.

**Allow-list captured empirically**, not guessed — fetched real `/swap` responses from Jupiter's public API (`lite-api.jup.ag`) and decoded the returned tx across 4 scenarios:
- single-hop SOL → USDC
- 3-hop SOL → BONK (AlphaQ → Hadron → Raydium CLMM)
- Token-2022 output mint (SOL → PYUSD)
- platform-fee-included swap (feeAccount + platformFeeBps set)

All 4 resolved to exactly: **Jupiter v6 router, Compute Budget, System, SPL Token, Associated-Token-Account**. `useSharedAccounts` (Jupiter's default) routes the actual AMM legs through the router via CPI, so individual AMM programs never surface as top-level instructions — only the router + wrapping/bookkeeping instructions do. Token-2022 is allow-listed defensively (didn't surface in any fixture — the wSOL leg always uses legacy Token — but the SDK's own fee-ATA derivation explicitly supports Token-2022 mints).

Wired into **both** independent integration points, guarding the raw provider response before any local mutation:
- `getJupiterSwapQuote.ts` (recipes/general-swap path) — guarded regardless of whether a fee is charged; the no-fee path (the only one live in prod today, treasury ATA map is empty) was previously **completely** unguarded.
- `packages/sdk/src/tools/swap/jupiter.ts` `buildJupiterSwapTx` — a second, fully independent Jupiter integration (SDK code-as-action tool) that bypasses the recipes path entirely.

Fund-movement-scope check (e.g. verifying the transfer amount/destination inside the router CPI) is **out of scope for v1** — tractable only via simulation, not static inspection, and the allow-list already collapses the attack surface to "trust the 5 canonical Solana programs."

## Runtime verification

`assertSafeSolanaSwapInstructions.test.ts` (7 cases):
- (a) passes on both real captured fixtures (single-hop + 3-hop, `fixtures/jupiter-swap-transactions.json`)
- (b) throws `SOL_SWAP_UNEXPECTED_PROGRAM` when the Jupiter router's static account-key is swapped to a foreign pubkey; plus a negative control proving a benign key swap does NOT false-positive
- (c) v0 ALT-resolved program indexes: synthetic message proving resolution against lookup-table-backed accounts works both ways (allow-listed passes, foreign throws), plus a fail-safe case (missing LUT refuses rather than silently passing)

`getJupiterSwapQuote.test.ts` + `jupiterSwap.test.ts` updated: guard mocked (own dedicated suite covers real behavior), new tests assert the guard is invoked with the raw transaction before any fee-ATA mutation, and that a guard refusal propagates as a thrown error instead of returning a signable tx.

```
packages/core: 11 test files, 74 tests passed (chain/swap/general/jupiter + chain/chains/solana)
packages/sdk:  1 test file, 19 tests passed (jupiterSwap.test.ts)
tsc --project .config/tsconfig.json --noEmit -> clean
packages/sdk tsc --noEmit --skipLibCheck -> clean
yarn build:shared / check:shared-exports -> up to date
eslint + prettier -> clean
```

## receipts

Backend-only fund-safety fix (no UI-visible change) — no sim screenshot; `no runtime changes` to the app surface. Verified against Jupiter's real production API responses (see fixture capture description above) rather than synthetic-only data.

🤖 Agent-generated

## Release Notes

* **Bug Fixes**
  * Added validation for Jupiter swap transactions on Solana to verify instructions originate from approved programs.
  * Added safeguards to detect and prevent unauthorized account manipulations during swap operations.
  * Validation now runs before any local transaction modifications are applied.


packages/core/chain/chains/solana/assertSafeSolanaSwapInstructions.ts
```diff
@@ -0,0 +1,148 @@
...
+ * plausibly need it.
+ */
+export const JUPITER_SWAP_ALLOWED_PROGRAM_IDS: ReadonlySet<string> = new Set([
+  JUPITER_V6_PROGRAM_ID,
+  COMPUTE_BUDGET_PROGRAM_ID,
+  SYSTEM_PROGRAM_ID,
+  TOKEN_PROGRAM_ID.toBase58(),
+  TOKEN_2022_PROGRAM_ID.toBase58(),
+  ASSOCIATED_TOKEN_PROGRAM_ID.toBase58(),
+])
```
NeOMakinG, line 52: `blocker:` program-only membership check. `SystemProgram`, `TOKEN_PROGRAM_ID` and `TOKEN_2022_PROGRAM_ID` are fund-moving programs — an injected top-level `SystemProgram.transfer` (SOL) or SPL `Token.Transfer`/`Approve`/`SetAuthority`/`CloseAccount` (tokens) to an attacker targets an allow-listed program and passes `.has(...)` at L89 unchanged, surviving into the signed bytes (recompiled verbatim in `prependJupiterFeeAta`). That is the exact drain SOL-01 names. Need destination/authority validation for these programs (source/authority == user, destination ∈ user ATAs / wSOL ATA / treasury fee ATA), or narrow the claim and keep SOL-01 open behind a tracked follow-up.

NeOMakinG (changes requested):
Two-lane review (Claude fund-safety/bypass + Codex `gpt-5.4` adversarial, independent). Both lanes converged on the same blocker. HEAD 6d6d1c29, worktree build, CI green (all 5 checks pass).

The mechanics of the guard are **correct and well-built**: it iterates **every** top-level instruction (`compiledInstructions.forEach`, all indices), resolves program ids through **ALT-backed keys** via `getAccountKeys({ addressLookupTableAccounts })` (correct static→writable-lut→readonly-lut ordering, delegated to web3.js), is **fail-closed** (throws before any signable tx, runs before the fee-ATA mutation at both sites and re-runs inside `prependJupiterFeeAta` after recompile), uses an **authoritative hardcoded** allow-list (a `ReadonlySet` of constants, not derived from the attacker-controlled `/swap` response), a **missing-LUT hard-fails** (web3.js `resolveAddressTableLookups` throws — verified fail-safe, never silently passes an unresolvable index), and both independent Jupiter sites (`getJupiterSwapQuote` + `buildJupiterSwapTx`) are guarded. Test coverage is genuinely good (real captured fixtures single+3-hop, foreign-program throw, benign-substitution negative control, ALT-resolved pass/throw, missing-LUT fail-safe).

**But the allow-list does not close the anti-drain objective it names.** See blocking finding.

## testing approach
- Static fund-safety/bypass trace of `assertSafeSolanaSwapInstructions` + both wire-in sites + `prependJupiterFeeAta` recompile path (confirmed injected instructions survive into signed bytes: `instructions: [createFeeAtaIx, ...decompiledMessage.instructions]`).
- ALT resolution + missing-LUT fail-closed path traced against web3.js `getAccountKeys`/`resolveAddressTableLookups` semantics.
- Codex `gpt-5.4` independent adversarial lane (verified referencing this PR: assertSafeSolanaSwapInstructions.ts:33/83, not cross-contaminated, not stale).
- CI: all 5 checks green (Unit, Integration, Lint/TypeCheck, CodeRabbit, Test Summary).
- No device/runtime QA (backend-only fund-safety guard, no UI surface).

## blocking

**1. Program-only allow-list permits the simplest drain — an injected top-level SOL/SPL transfer to an attacker passes the guard unchanged.** `assertSafeSolanaSwapInstructions` checks only that each instruction's *program* is in the set. But the set includes `SystemProgram`, `TOKEN_PROGRAM_ID`, and `TOKEN_2022_PROGRAM_ID` — the programs that *move funds*. A compromised/MITM'd Jupiter proxy can splice a top-level:
- `SystemProgram.transfer { from: userWallet, to: attacker }` — drains SOL, or
- SPL `Token.Transfer` / `Approve` / `SetAuthority` / `CloseAccount` against the user's token accounts (and Token-2022 variants) — drains/seizes tokens.

Every one targets an already-allow-listed program → `JUPITER_SWAP_ALLOWED_PROGRAM_IDS.has(...)` returns true → the guard accepts it, and the recompile in `prependJupiterFeeAta` (or the verbatim no-fee path) forwards it to MPC signing. The user is a signer of the whole message, so the injected transfer is authorized. This is *exactly* the "drain transfer to an attacker address" the PR body names — and it is not stopped. Legit Jupiter swaps genuinely contain top-level System/Token instructions (SOL wrap/unwrap, syncNative), so a program-level check cannot distinguish a legit wrap-transfer from an injected drain-transfer without inspecting **destination/authority**.

The PR body's "collapses the attack surface to trust the 5 canonical Solana programs" overstates the protection: trusting System + Token means trusting arbitrary transfers those programs execute. The "fund-movement-scope out of scope for v1" caveat is scoped to movements *inside the router CPI* — but a **top-level** injected transfer is statically inspectable and is the lowest-effort drain, left open.

Note: no downstream layer catches this — pre-sign simulation (SOL-06 / va#1760) is success-only (not WYSIWYS), so a drain that *succeeds* still passes.

Two acceptable resolutions:
- **(preferred) Destination/authority validation for the fund-moving programs.** For any top-level System/Token/Token-2022 instruction, decode it and assert the funds `source`/`authority` is the user AND the `destination` is a known-good account (user's own ATAs, the wSOL ATA, the treasury fee ATA). Jupiter's top-level System/Token instruction set is small and known, so this is tractable statically — and the allow-list already forces the attacker onto exactly these decodable programs.
- **(minimum) Narrow the claim + track a blocking follow-up.** If destination validation is genuinely deferred, the PR body must stop claiming the attack surface is collapsed, and SOL-01 must **not** be marked closed until a destination/WYSIWYS check lands. As written, an APPROVE would endorse a protection guarantee the code doesn't provide.

## should-fix

**2. Missing-LUT → hard fail is an availability risk on flaky RPC (safe direction).** `resolveSolanaAddressLookupTables` skips a LUT that fails all 3 fetch attempts; `getAccountKeys` then throws `Failed to find address lookup table account`, refusing the swap. Correct fund-safety direction (fail-closed), but a transient Jupiter/RPC hiccup will refuse a legit swap with an opaque error. Acceptable; note it as known.

## suggestion

**3. Allow-list omits Memo program.** If Jupiter ever adds a top-level referral/memo instruction (`MemoSq4...`), legit swaps over-block. Availability-only, fail-closed loudly — fine to defer, but worth a comment noting the list is intentionally minimal.

**4. LiFi Solana path is an unguarded sibling.** `packages/core/chain/swap/general/lifi/api/injectSolanaAtaIfMissing.ts:155` deserializes a provider-supplied `VersionedTransaction` and forwards it to signing with no program guard — the same class as SOL-01 but for LiFi (out of scope here, since SOL-01 is Jupiter-specific). Worth a follow-up bead so the anti-drain floor is consistent across Solana swap providers.

## risk
Fund-critical Solana anti-drain. The guard is a real defense-in-depth layer (blocks injected *custom/unknown-program* instructions — the classic route-through-malicious-program vector) and regresses nothing. But it leaves the lowest-effort drain (top-level System/SPL transfer to attacker) open while the PR body claims otherwise. Merging as-is *and* marking SOL-01 closed would leave users exposed to the exact attack the finding describes.

## verdict
**REQUEST_CHANGES.** Not for correctness of what it does — the ALT resolution, every-index coverage, fail-closed sequencing, and authoritative allow-list are all right — but because the program-only allow-list does not satisfy the SOL-01 anti-drain objective: an injected top-level SOL/SPL transfer to an attacker passes unchanged. Either add destination/authority validation for the fund-moving programs, or narrow the PR's protection claim and keep SOL-01 open behind a tracked follow-up. Happy to re-review immediately on either path.

<sub>🤖 two-lane review (Claude + Codex gpt-5.4). Fund-safety-first, adversarial lens.</sub>

packages/core/chain/chains/solana/assertSafeSolanaSwapInstructions.ts
```diff
@@ -0,0 +1,324 @@
...
+  }
+  if (type === TOKEN_SET_AUTHORITY) {
+    throw new UnsafeSolanaSwapFundMovementError(
+      index,
+      'Token.SetAuthority is not a legitimate top-level Jupiter swap instruction — possible account seizure'
+    )
+  }
+
+  // Transfer (type=3): accounts[2]=authority must be the user
+  if (type === TOKEN_TRANSFER) {
```
NeOMakinG, line 185: `blocker:` SPL/Token-2022 token drain STILL OPEN — layer-2 validates the AUTHORITY (always the user in a self-drain) instead of the DESTINATION for Token.Transfer / TransferChecked / CloseAccount — The layer-2 `SystemProgram.Transfer` branch is correct — it checks the DESTINATION (`dest.equals(wSolAta)`, L134), so a spliced SOL transfer to an attacker is rejected. But the `Token`/`Token-2022` branch checks the wrong account. For `Token.Transfer` (type 3) it asserts `accounts[2]` (authority) `== userWallet` (L167-177); for `TransferChecked` (12) `accounts[3]` (L181-192); for `CloseAccount` (9) `accounts[2]` (L195-206). To move the user's SPL tokens OUT of the user's own token account, the authority MUST be the user (it's a required signer) — so `authority == user` is *definitionally true for every drain of the user's own tokens*. The account that distinguishes a legit swap-internal transfer from an exfiltration is the DESTINATION (`accounts[1]`), and it is never checked. Concrete still-open drains a compromised/MITM Jupiter proxy can splice: (1) `Token.Transfer{ source: user's USDC ATA, dest: attacker ATA, authority: user }` → `accounts[2]==user` → passes → tokens drained. (2) `Token.TransferChecked{...dest: attacker, authority(accounts[3]): user}` → passes. (3) `Token.CloseAccount{ account: user's wSOL ATA, dest: attacker, authority(accounts[2]): user }` → passes → the unwrapped SOL + rent is sent to the attacker (this ALSO re-opens a SOL drain: wrap-all-SOL via the allowed System.transfer-to-wSOL-ATA, then CloseAccount the wSOL ATA to the attacker). The PR's own test at `assertSafeSolanaSwapInstructions.test.ts:247` documents the hole — 'allows Token.Transfer where authority IS the user wallet' uses `[user, user, user]` (benign dest) and asserts `.not.toThrow()`; swapping `accounts[1]` to an attacker keeps the exact same code path green. The negative tests at L233/L240 only exercise `authority == attacker`, which is a NON-drain (a non-user authority can't move the user's tokens anyway) — so the drain path is untested.

fix: Validate the DESTINATION for the token fund-movement instructions, mirroring the (correct) System.Transfer branch — not the authority. For `Token.Transfer`/`TransferChecked`, either reject any top-level user-authority token transfer outright (Jupiter shared-accounts routing moves tokens via CPI through the router, so a legit top-level user-authority `Token.Transfer` doesn't appear in the captured fixtures — same rationale you already used to reject Approve/SetAuthority), OR require `accounts[1]` (destination) to be a user-controlled / known-program-derived account. For `CloseAccount`, require `accounts[1]` (the lamport destination) `== userWallet` — that's the only legit wSOL-unwrap target. Add negative tests for the real drain shape: `Token.Transfer`/`CloseAccount` with `authority == user` but `dest == attacker` MUST throw. Keep the (correct) authority check as an additional guard if you like, but the destination check is the load-bearing one.

packages/core/chain/chains/solana/assertSafeSolanaSwapInstructions.ts
```diff
@@ -0,0 +1,324 @@
...
+
+const assertSystemInstruction = (
+  ix: MessageCompiledInstruction,
+  keys: AccountKeys,
+  userWallet: PublicKey,
+  index: number
+): void => {
+  if (ix.data.length < 4) return
+  const type = new DataView(ix.data.buffer, ix.data.byteOffset, 4).getUint32(0, /* littleEndian */ true)
+  if (type !== SYSTEM_TRANSFER) return
```
NeOMakinG: `suggestion:` System.Transfer branch only guards discriminant 2; other lamport-moving System instructions are silently allowed — `assertSystemInstruction` returns early for any System discriminant != 2 (Transfer). `TransferWithSeed` (11) and `CreateAccount*` (0/3, which fund a new account with lamports) are lamport-moving and pass unchecked. Lower-likelihood than the token drain (they need the user to control a seed-derived/base account), but for completeness consider rejecting any non-Transfer System instruction on the swap path (a legit Jupiter swap only emits System.Transfer top-level per the empirical capture).

fix: Reject unrecognized top-level System discriminants (allow only Transfer with the wSOL-ATA dest check), rather than silently returning.

NeOMakinG (changes requested):
gm — multi-lane pass: claude + codex. conceptual-only (no runtime this pass). requesting changes.

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | BLOCKER — program-only allow-list permits the simplest drain (spliced top-level System.transfer / SPL Token.Transfer/Approve/SetAuthority/CloseAccount to an attacker passes the program `.has()` check unchanged and survives into the signed bytes) | ⚠️ PARTIAL — layer-2 added. SOL drain (`SystemProgram.Transfer`) is CLOSED via a real destination check (`dest == user's wSOL ATA`). Approve/ApproveChecked/SetAuthority are rejected outright. **But the SPL/Token-2022 token drain is STILL OPEN**: `Token.Transfer`/`TransferChecked`/`CloseAccount` validate the AUTHORITY (`accounts[2]`/`[3]`) instead of the DESTINATION — and the authority is *definitionally* the user for any drain of the user's own token account. Spliced `Token.Transfer{source: userATA, dest: attacker, authority: user}` passes. Still blocking. |
| 1 | should-fix — missing-LUT fail-closed = availability risk on flaky RPC (opaque error) | ⚠️ partially addressed — `fetchLutWithRetry` now retries 3× with exponential backoff before giving up; still fail-closed (refuses the swap) on total LUT-fetch failure, which is the correct safe direction. Residual availability risk noted. |
| 1 | suggestion — Memo program not allow-listed → over-block if Jupiter adds a referral memo | ℹ️ unchanged — Memo still not in the allow-list. Availability-only, non-gating. |
| 1 | pre-existing sibling — LiFi Solana `injectSolanaAtaIfMissing.ts:155` deserializes a provider VersionedTransaction with NO guard (same drain class) | ⏭️ deferred — still unguarded at HEAD (`injectSolanaAtaIfMissing.ts:155`). Out of SOL-01 (Jupiter) scope; tracked follow-up for a consistent Solana anti-drain floor. |

## blocking
### 1. SPL/Token-2022 token drain STILL OPEN — layer-2 validates the AUTHORITY (always the user in a self-drain) instead of the DESTINATION for Token.Transfer / TransferChecked / CloseAccount — `packages/core/chain/chains/solana/assertSafeSolanaSwapInstructions.ts:167`
The layer-2 `SystemProgram.Transfer` branch is correct — it checks the DESTINATION (`dest.equals(wSolAta)`, L134), so a spliced SOL transfer to an attacker is rejected. But the `Token`/`Token-2022` branch checks the wrong account. For `Token.Transfer` (type 3) it asserts `accounts[2]` (authority) `== userWallet` (L167-177); for `TransferChecked` (12) `accounts[3]` (L181-192); for `CloseAccount` (9) `accounts[2]` (L195-206). To move the user's SPL tokens OUT of the user's own token account, the authority MUST be the user (it's a required signer) — so `authority == user` is *definitionally true for every drain of the user's own tokens*. The account that distinguishes a legit swap-internal transfer from an exfiltration is the DESTINATION (`accounts[1]`), and it is never checked. Concrete still-open drains a compromised/MITM Jupiter proxy can splice: (1) `Token.Transfer{ source: user's USDC ATA, dest: attacker ATA, authority: user }` → `accounts[2]==user` → passes → tokens drained. (2) `Token.TransferChecked{...dest: attacker, authority(accounts[3]): user}` → passes. (3) `Token.CloseAccount{ account: user's wSOL ATA, dest: attacker, authority(accounts[2]): user }` → passes → the unwrapped SOL + rent is sent to the attacker (this ALSO re-opens a SOL drain: wrap-all-SOL via the allowed System.transfer-to-wSOL-ATA, then CloseAccount the wSOL ATA to the attacker). The PR's own test at `assertSafeSolanaSwapInstructions.test.ts:247` documents the hole — 'allows Token.Transfer where authority IS the user wallet' uses `[user, user, user]` (benign dest) and asserts `.not.toThrow()`; swapping `accounts[1]` to an attacker keeps the exact same code path green. The negative tests at L233/L240 only exercise `authority == attacker`, which is a NON-drain (a non-user authority can't move the user's tokens anyway) — so the drain path is untested.
action: Validate the DESTINATION for the token fund-movement instructions, mirroring the (correct) System.Transfer branch — not the authority. For `Token.Transfer`/`TransferChecked`, either reject any top-level user-authority token transfer outright (Jupiter shared-accounts routing moves tokens via CPI through the router, so a legit top-level user-authority `Token.Transfer` doesn't appear in the captured fixtures — same rationale you already used to reject Approve/SetAuthority), OR require `accounts[1]` (destination) to be a user-controlled / known-program-derived account. For `CloseAccount`, require `accounts[1]` (the lamport destination) `== userWallet` — that's the only legit wSOL-unwrap target. Add negative tests for the real drain shape: `Token.Transfer`/`CloseAccount` with `authority == user` but `dest == attacker` MUST throw. Keep the (correct) authority check as an additional guard if you like, but the destination check is the load-bearing one.

## should-fix
### 1. branch is 12 commits behind main and conflicts with it (mergeStateStatus DIRTY) — rebase needed — `packages/sdk/src/tools/swap/jupiter.ts`
`git merge-tree` reports a real textual conflict on `packages/sdk/src/tools/swap/jupiter.ts` (and package.json), because sibling swap-security PRs (sol03/04, thor04) landed on main touching the same Jupiter file. The heavy SDK CI (Lint / Unit / Integration) is not registered at HEAD 35f2ce2d — only CodeRabbit shows. Rebase onto main so the full CI matrix runs against the merge target before this is considered mergeable.
action: `git rebase origin/main`, resolve the jupiter.ts conflict, push, and confirm the full SDK CI (not just CodeRabbit) goes green at the new HEAD.

### 2. changeset text describes only layer-1 (program allow-list), not the layer-2 destination/authority validation — `.changeset/sol-01-swap-instruction-allowlist.md`
The changeset only documents the program allow-list ('throws SOL_SWAP_UNEXPECTED_PROGRAM on the first unrecognized one') and never mentions the fund-movement destination/authority layer that is the substance of this round. Consumers reading the changelog won't know the drain-injection guard exists.
action: Extend the changeset to describe layer-2 (destination check for System.Transfer, authority/destination validation + Approve/SetAuthority rejection for Token programs) once the layer-2 logic is fixed.

## suggestion
### 1. System.Transfer branch only guards discriminant 2; other lamport-moving System instructions are silently allowed — `packages/core/chain/chains/solana/assertSafeSolanaSwapInstructions.ts:118`
`assertSystemInstruction` returns early for any System discriminant != 2 (Transfer). `TransferWithSeed` (11) and `CreateAccount*` (0/3, which fund a new account with lamports) are lamport-moving and pass unchecked. Lower-likelihood than the token drain (they need the user to control a seed-derived/base account), but for completeness consider rejecting any non-Transfer System instruction on the swap path (a legit Jupiter swap only emits System.Transfer top-level per the empirical capture).
action: Reject unrecognized top-level System discriminants (allow only Transfer with the wSOL-ATA dest check), rather than silently returning.

## q
- Does any legit Jupiter shared-accounts route ever emit a top-level user-authority `Token.Transfer`/`TransferChecked`? If not (as the fixtures suggest), the cleanest fix is to reject them outright like Approve/SetAuthority — confirm against the captured fixtures.

## risk
Fund-critical (anti-drain guard, SOL-01). Layer-2 closes the SOL `System.Transfer` drain and the Approve/SetAuthority seize vectors, but the SPL/Token-2022 token drain — the exact `Token.Transfer to attacker` this PR names — remains exploitable because the guard checks the authority (definitionally the user) rather than the destination. A compromised/MITM Jupiter proxy can still splice a token exfiltration (and, via CloseAccount-to-attacker, a SOL exfiltration) that survives into the blind-signed bytes. The mechanics (every-index iteration, ALT resolution, fail-closed, authoritative allow-list, both wire-in sites, authoritative user wallet from `from.address`) are intact and correct — the gap is the account chosen for the token check.

## verdict
REQUEST CHANGES

packages/core/chain/chains/solana/assertSafeSolanaSwapInstructions.ts
NeOMakinG, line 185: blocker: this branch validates `accountKeyIndexes[2]` = the **authority**, which is definitionally the user for any transfer out of the user's own ATA. The account that distinguishes a legit swap-leg from exfiltration is the **destination `accounts[1]`**, which is never read. A spliced `Token.Transfer { source: userATA, destination: attackerATA, authority: user }` passes (`accounts[2]==user`) → tokens drained. TransferChecked (dest `accounts[2]`) and CloseAccount (dest `accounts[1]`) below have the identical flaw. Validate the DESTINATION (mirror the System branch) or reject top-level user-authority token transfers outright, per shared-accounts routing doing token moves via CPI.

NeOMakinG (changes requested):
Re-review at **HEAD 23ab2864** (round 3). ⚠️ **FUND-CRITICAL — the SPL/Token drain flagged in round 2 is NOT closed.** The commits since my r2 review (`44ed9bb2..23ab2864`) are a **main-merge only** — `git diff 44ed9bb2..23ab2864 -- assertSafeSolanaSwapInstructions.ts` is **empty**. The guard file is byte-identical to what I reviewed at r2. Marking the PR ready + merging main did not touch the residual drain. Keeping **CHANGES_REQUESTED**.

## testing approach
Static trace at HEAD 23ab2864 in an isolated worktree (`/tmp/wt-sdk-1104`, detached at the exact SHA). Confirmed: (a) guard file 0-diff vs my r2 head; (b) wire-in still live on the real sign path (`getJupiterSwapQuote.ts:161`, `jupiter.ts:320`, `jupiterFeeAta.ts:116`); (c) byte-decoded the SPL account layouts against `assertTokenInstruction`; (d) re-read the test file. Codex gpt-5.4 second-opinion lane stalled (0-byte buffering stall, known recurring) → Claude load-bearing, consistent with the r2 conclusion.

## prior issues — status
|---|---|---|---|---|---|
| r1 | 1 | blocker | program-only allow-list permits a top-level injected transfer to attacker | ⚠️ partial | SOL/System branch fixed in r2 (dest==wSOL ATA). Token branch still open — see below. |
| r2 | 1 | blocker | Token.Transfer/TransferChecked/CloseAccount validate the AUTHORITY, not the DESTINATION → token drain still open | ❌ still blocking | Guard byte-identical to r2 (`44ed9bb2..23ab2864` empty on this file). Drain persists verbatim. |
| r2 | 2 | should-fix | branch behind main + DIRTY, heavy CI not registered | ✅ fixed | main merged into branch at `23ab2864`; mergeStateStatus now BLOCKED only by this review. |
| r2 | 3 | should-fix | missing-LUT total-failure fail-closed but opaque | ⚠️ partial | `fetchLutWithRetry` 3× + backoff, fail-closed (safe direction). Unchanged. |
| r2 | 4 | suggestion | `assertSystemInstruction` returns early for non-Transfer System discriminants (TransferWithSeed/CreateAccount lamport movers) | ℹ️ documented | Unchanged. Lower-likelihood; still worth rejecting non-Transfer System on the swap path. |
| r1 | s | suggestion | Memo program not allow-listed (over-block risk if Jupiter adds referral memo) | ℹ️ documented | Unchanged (availability-only). |
| r1 | pre | pre-existing | LiFi Solana `injectSolanaAtaIfMissing.ts:155` deserializes provider tx with no guard | ⏭️ deferred | Out of scope for sol-01 (Jupiter). Track a follow-up for a consistent Solana anti-drain floor. |

## blocking
**Token branch validates the AUTHORITY (definitionally the user), leaving the DESTINATION unchecked → token drain survives into the signed message.** `assertTokenInstruction` (`assertSafeSolanaSwapInstructions.ts:142-210`):

- `Token.Transfer` (type 3, layout `[source, destination, authority]`) → checks `accountKeyIndexes[2]` (authority) == userWallet (L167-178). Authority == user is **definitionally true for any transfer out of the user's own ATA** — the victim MUST be the signer for their own tokens to move. The account that separates a legit swap-leg from exfiltration is the **destination `accounts[1]`, which is never read.**
- `Token.TransferChecked` (type 12, layout `[source, mint, destination, authority]`) → checks `accountKeyIndexes[3]` (authority) (L181-192). Authority index is correct, but again it's the wrong account to validate; **destination `accounts[2]` unchecked.**
- `Token.CloseAccount` (type 9, layout `[account, destination, authority]`) → checks `accountKeyIndexes[2]` (authority) (L195-206); **destination `accounts[1]` unchecked** — the slot that receives the unwrapped SOL + rent.

Exact adversarial sequences a compromised/MITM Jupiter proxy splices as a top-level instruction (all target allow-listed programs, so L1 passes; all have authority==user, so L2 passes):
1. `Token.Transfer { source: userATA, destination: attackerATA, authority: user }` → `accounts[2]==user` → **PASSES → tokens drained.**
2. `Token.TransferChecked { source: userATA, mint, destination: attackerATA, authority: user }` → `accounts[3]==user` → **PASSES → tokens drained.**
3. `Token.CloseAccount { account: userWSolATA, destination: attacker, authority: user }` → `accounts[2]==user` → **PASSES → unwrapped SOL + rent → attacker.** This also re-opens a SOL drain: the allowed `SystemProgram.Transfer` can wrap SOL into the user's wSOL ATA (dest==wSOL ATA, passes the fixed System branch), then CloseAccount sweeps it to the attacker.

The test suite documents the hole rather than closing it. `test:236` `buildTokenInstruction(3, [user, user, attacker])` and `test:242` `buildTokenInstruction(9, [user, attacker, attacker])` only exercise **authority==attacker — a non-drain** (a non-user authority cannot move the user's tokens anyway). `test:247-250` `.not.toThrow()` for authority==user with a benign destination: flipping `accounts[1]`→attacker keeps this exact path green. The real drain shape (**authority==user, destination==attacker**) is untested for all three instructions.

**Fix:** validate the DESTINATION for token fund-movers, mirroring the (correct) System branch. Per the PR's own fixtures, Jupiter shared-accounts routing performs token moves via CPI, not as top-level instructions — so the safest fix is to **reject top-level user-authority `Token.Transfer`/`TransferChecked` outright** (same rationale already applied to `Approve`), and for `CloseAccount` require `accounts[1]` (destination) == the user wallet (rent/unwrap must return to the user). If any legit fixture genuinely needs a top-level token transfer, constrain the destination to a user-controlled address (user ATA / wSOL ATA / the treasury fee ATA).

## risk
Fund-critical. A single spliced top-level SPL Token instruction drains the user's tokens (or, via wrap→CloseAccount, their SOL) to an attacker while passing both guard layers and rendering as a normal swap in a UI that cannot decode raw Solana bytes. The downstream SOL-06 pre-sign sim (va#1760) is success-only, not WYSIWYS, so a drain that succeeds is not caught there either. Do not merge until the destination is validated for the three token fund-movers.

## verdict
**CHANGES_REQUESTED.** The SOL/System branch is correct and the L1 allow-list + fail-closed structure is sound, but the Token/Token-2022 branch validates the authority (always the user) instead of the destination — the token drain remains open exactly as in round 2, and the main-merge did not address it.

<sub>🤖 reviewer lane — Claude load-bearing (Codex second-opinion lane stalled 0-byte). Traced at HEAD 23ab2864 in isolated worktree.</sub>

NeOMakinG (approved):
**round 4 — SPL/Token drain CLOSED. Approving.** The layer-3 fix at `cdbf9894` does exactly what the three prior rounds asked: it stops validating the *authority* (which is definitionally the user for any drain of the user's own token account) and instead **rejects top-level `Token.Transfer`/`TransferChecked` outright** and **validates the `CloseAccount` destination == user wallet**. The fund-safety objective (SOL-01) is now met.

## testing approach
Non-device static + empirical, worktree `/tmp/wt-sdk-1104` @ `cdbf9894` (canonical `node_modules`/built `dist` symlinked). Two lanes: Claude (load-bearing) full re-trace of `assertTokenInstruction` account indices + all three adversarial drain sequences + wire-in reachability; Codex `gpt-5.4` stalled again (7th consecutive occurrence on this file family — not load-bearing this round). Ran the core vitest suite and the **red-on-revert experiment**.

- **16/16 tests pass** at HEAD.
- **red-on-revert PROVEN**: reverting the token branch back to the old authority-check turns exactly the three drain-shape tests RED — `Token.Transfer` and `Token.TransferChecked` drain tests report *"expected function to throw, but it didn't"* (i.e. the old guard signs the drain), and `Token.CloseAccount` reports the wrong error string. The tests now pin the **real drain shape** (authority==user, destination==attacker), not the r3 masking shape (authority==attacker, a non-drain).
- **no over-block**: the two real captured Jupiter fixtures (single-hop SOL→USDC, 3-hop SOL→BONK) still pass `.not.toThrow()` — rejecting top-level `Token.Transfer` outright does not break legitimate shared-accounts routes (token legs happen via CPI inside the router, never top-level).
- **reaches real sign**: all three wire-in sites live — `getJupiterSwapQuote.ts:161`, `jupiter.ts:320`, `jupiterFeeAta.ts:116`. No inert branch.

## prior issues — status
| r1 | 1 | blocker | program-only allow-list permits injected top-level System/Token drain transfer | ✅ fixed | layer-2 (r2) closed SOL, layer-3 (`cdbf9894`) closes Token |
| r2/r3 | 1 | blocker | token branch validates AUTHORITY (always==user) not DESTINATION → spliced `Token.Transfer{dest:attacker,auth:user}` PASSES = SPL/token DRAIN | ✅ fixed | `cdbf9894`: Transfer(3)+TransferChecked(12) rejected outright; CloseAccount(9) validates dest `accounts[1]`==user. red-on-revert proven |
| r1 | 2 | should-fix | missing-LUT → opaque throw (availability on flaky RPC) | ⚠️ partial | `fetchLutWithRetry` 3× + backoff, still fail-closed (safe dir). accepted |
| r2 | 3 | should-fix | changeset text describes only layer-1 (program allow-list) | ❌ still | `.changeset/sol-01-...md` still omits the layer-2/3 dest/authority validation |
| r2 | 4 | should-fix | branch behind main / DIRTY, heavy CI not registered | ✅ fixed | MERGEABLE, all 5 checks green (Unit/Integration/Lint/Test Summary/CodeRabbit); BEHIND but clean |
| r2 | 5 | suggestion | System non-Transfer (TransferWithSeed/CreateAccount) fall through | ℹ️ documented | lower-likelihood lamport movers, unchanged |
| r1 | 6 | suggestion | Memo program not allow-listed (over-block if Jupiter adds referral memo) | ℹ️ documented | unchanged |
| r1 | 7 | pre-existing | LiFi Solana `injectSolanaAtaIfMissing.ts` deserializes provider tx with no guard | ⏭️ deferred | out of scope (SOL-01 = Jupiter); tracked follow-up |

## should-fix
1. **stale JSDoc** — `assertSafeSolanaSwapInstructions.ts:225-226` still documents the *old* behavior: `Token.Transfer/TransferChecked: authority must be the user wallet` and `Token.CloseAccount: authority must be the user wallet`. The code now rejects Transfer/TransferChecked outright and validates the CloseAccount *destination*. Update the doc so a future maintainer doesn't "restore" the buggy authority-check.
2. **stale changeset** (carry-forward from r2) — `.changeset/sol-01-swap-instruction-allowlist.md` describes only the layer-1 program allow-list; it never mentions the fund-movement destination/authority validation that took three rounds to get right. Fold in a sentence.
3. **`Token.Burn`(8)/`BurnChecked`(15) fall through unchecked** — under the same compromised-proxy threat model, an injected top-level `Token.Burn{account:userATA, authority:user}` destroys the user's tokens. No attacker profit (griefing, not exfil), but it *is* fund loss, and it's inconsistent with the guard now rejecting every other SPL fund-mover outright. Reject Burn/BurnChecked top-level too.

## q
1. All four empirical fixtures are **SOL-input** routes (SOL→X). Rejecting top-level `Token.Transfer`/`TransferChecked` outright assumes a legit swap never carries one at top level — true for SOL-input (the input leg is native SOL wrapped via `System.Transfer`). Did you verify an **SPL-input** route (e.g. USDC→SOL, or token→token) with shared-accounts routing also never surfaces a top-level `Token.Transfer`? If it does, this over-blocks. It's **fail-closed** (a refused swap, not a drain), so not blocking — but worth a captured token-input fixture to be sure real swaps still sign.

## risk
Low residual. The load-bearing fund-safety property (no injected token/SOL transfer to an attacker survives into signed bytes) is closed and pinned by red-on-revert tests. Remaining items are documentation accuracy, a griefing-tier Burn gap, and an availability question — none can cause a drain. Guard fails closed on every unrecognized program and every unrecognized fund-mover branch.

## verdict
**APPROVE.** Three rounds of blocker are resolved; the SPL/Token drain is empirically closed (red-on-revert). The should-fix items (stale JSDoc/changeset, Burn) and the token-input q are non-blocking follow-ups.

---
🤖 dual-lane review (Claude load-bearing + Codex adversarial). Verified at HEAD `cdbf9894`.

Issue #1057, opened 2026-07-08
[audit-AGG-03] CowSwap permitRequired flag is threaded as data but never consulted → builds a redundant o


gomesalexandre:
## Context (fund-safety audit round 2, 2026-07-08 · finding AGG-03 · Medium)

GeneralSwapQuote.ts:53 documents the intent (build a gasless EIP-2612 permit instead of an approve tx); permitRequired is computed and serialized (getCowSwapQuote.ts:69) but the allowance-check block at build.ts:362-377 runs unconditionally with no permitRequired branch. For permit tokens (USDC) the SDK still builds a full approve() keysign step, defeating the point of the KNOWN_PERMIT_TOKENS allowlist. Gate the block on !permitRequired for the cowswap arm.

## Effort / risk / confidence
Effort S/MRisk LOWConfidence HIGH

## Runtime repro (BEFORE + AFTER — mandatory)
SDK-CLI: build the SDK and unit/CLI-drive the affected path with a fixture; assert before (bug) vs after (fixed).

## Prompt for AI agents
Implement the fix for audit finding **AGG-03** in `vultisig/vultisig-sdk`. GeneralSwapQuote.ts:53 documents the intent (build a gasless EIP-2612 permit instead of an approve tx); permitRequired is computed and serialized (getCowSwapQuote.ts:69) but the allowance-check block at build.ts:362-377 runs unconditionally with no permitRequired branch. For permit tokens (USDC) the SDK still builds a full approve() keysign step, defeating the point of the KNOWN_PERMIT_TOKENS allowlist. Gate the block on !permitRequired for the cowswap arm.  Make the minimal, additive, fail-closed change at the anchors named above; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI: build the SDK and unit/CLI-drive the affected path with a fixture; assert before (bug) vs after (fixed).). **Grep open PRs in this repo first** — several audit findings already have in-flight PRs; advance/verify an existing one (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **AGG-03** (round 2, Medium) · bead `agent-backend-ts-s745h`. Runtime before+after is the finish line, not unit-tests-green._

gomesalexandre:
Already investigated and deliberately deferred, not left unaddressed — verified against current source.

[...]

PR #1105, opened 2026-07-09
fix(swap): skip redundant approve for cowswap permit orders (agg-03)

### LINKED ISSUE #1103: [audit-r2][AGG-03] CowSwap permitRequired flag never consulted, builds a redundant on-chain approve
## What

\`packages/core/mpc/keysign/swap/build.ts\` runs an ERC-20 allowance check + \`erc20ApprovePayload\` build unconditionally for every EVM swap where \`fromCoin.id\` is set:

\`\`\`ts
if (isChainOfKind(chain, 'evm') && fromCoin.id) {
  const allowance = await getErc20Allowance({ chain, id: fromCoin.id, address: fromCoin.address, spender })
  if (allowance < chainAmount) {
    keysignPayload.erc20ApprovePayload = create(Erc20ApprovePayloadSchema, { amount: chainAmount.toString(), spender })
  }
}
\`\`\`

Meanwhile, \`permitRequired\` is computed (via \`KNOWN_PERMIT_TOKENS\`) and serialized into the CowSwap keysign payload (\`...(order.permitRequired ? { permitRequired: true } : {})\`), but this flag is never consulted by the allowance-check block above to skip building a redundant on-chain approve.

\`GeneralSwapQuote.ts\`'s doc comment on \`permitRequired\` states the original intent:

> Set to true when the sellToken supports EIP-2612 permit. When true the signing flow should build and include a permit signature alongside the order signature, avoiding a separate ERC-20 approve transaction.

So today, for a permit-token (e.g. USDC) CowSwap order, the SDK still builds a redundant on-chain \`approve()\` keysign step — defeating the point of gasless EIP-2612 permits and the \`KNOWN_PERMIT_TOKENS\` allowlist.

## Fix

Gate the approve-building logic so it's skipped ONLY for the CowSwap-general arm when \`permitRequired === true\`. Every other swap path (native THOR/Maya, 1inch, LiFi, KyberSwap, SwapKit, non-permit CowSwap tokens) must keep building a real approve exactly as before.

Audit id: AGG-03 (round-2 spec-level fund-safety audit).

gomesalexandre:
closes #1103

Audit AGG-03 (round-2 spec-level fund-safety audit).

`packages/core/mpc/keysign/swap/build.ts`'s ERC-20 allowance-check block ran
unconditionally for every EVM swap with `fromCoin.id` set, building an
`erc20ApprovePayload` whenever allowance was insufficient. Meanwhile
`permitRequired` (from `KNOWN_PERMIT_TOKENS`) was computed and serialized into
the CowSwap keysign payload's `data`, but never consulted to skip the
approve. Net effect: a permit-eligible CowSwap order (e.g. USDC) still built
a redundant on-chain `approve()` keysign step, defeating the point of the
gasless EIP-2612 permit path (`GeneralSwapQuote.ts`'s doc comment on
`permitRequired` states the intended behavior).

## Fix

Added a `cowSwapPermitRequired` boolean derived the same way the file already
derives `transferTx`/`thirdPartyGasLimitEstimation` (via `matchRecordUnion`
over the quote), then gated the approve block: `!cowSwapPermitRequired &&
...`. This is `false` for every arm except a CowSwap order with
`permitRequired === true` — so native swaps, 1inch/LiFi/Kyber/SwapKit
general swaps, and non-permit CowSwap tokens are completely unaffected.

## Tests

New `build.agg03CowSwapPermit.test.ts` (3 cases):
1. CowSwap order w/ `permitRequired: true` + insufficient allowance → NO
   `erc20ApprovePayload`, `getErc20Allowance` not even called.
2. CowSwap order w/o `permitRequired` + insufficient allowance → approve
   still built, `getErc20Allowance` called with the CowSwap vault relayer as
   spender.
3. Regression guard: a non-CowSwap EVM swap (1inch) still builds a real
   approve, unaffected by the new gate.

```
Test Files  4 passed (4)
     Tests  13 passed (13)
```
(the 4 files = the new test + the 3 existing build.ts suites, confirming no
regression)

`yarn tsc --project .config/tsconfig.json --noEmit` clean, eslint clean on
`build.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CowSwap swap flows that could fail silently when permit-based orders were created without a fallback approval.
  * Swaps now continue to include the standard token approval step until the permit flow is fully available.
  * Improved reliability for affected swap transactions, including existing EVM swap routes.


packages/core/mpc/keysign/swap/build.ts
```diff
@@ -359,7 +376,7 @@ export const buildSwapKeysignPayload = async ({
     keysignPayload.swapPayload.value.quote.tx.gas = BigInt(gasLimit)
   }
 
-  if (isChainOfKind(chain, 'evm') && fromCoin.id) {
+  if (isChainOfKind(chain, 'evm') && fromCoin.id && !cowSwapPermitRequired) {
```
NeOMakinG: `blocker:` This gate skips the only wired allowance mechanism (the on-chain `erc20ApprovePayload`) for permit-token CowSwap orders, but no EIP-2612 permit is actually produced to replace it: `buildEip2612Permit` has zero callers, the keysign payload carries only the `permitRequired` boolean (no permit digest for the MPC to sign), and `submitCowSwapOrder` posts no permit. Result: USDC/USDT/DAI CowSwap orders get neither approve nor permit → the solver can't pull the sell token → the order never settles (silent failed swap). Keep the approve until the permit path is wired (`buildEip2612Permit` invoked + permit digest added to the keysign payload + MPC signs it + consumer bundles it into submission).

NeOMakinG (changes requested):
Reviewed as a fund-safety gate on CowSwap approve-skipping. Two independent lanes (Claude source-trace + Codex adversarial, gpt-5.4) converged on the **same blocker**: this PR removes the only working allowance mechanism for permit-token CowSwap orders while the EIP-2612 permit that is supposed to replace it is never actually built or submitted.

## testing approach
- Unique worktree `/tmp/wt-sdk-1105` @ `60fd4565`, upstream fetched.
- Traced `permitRequired` end-to-end: derivation (`getCowSwapQuote.ts:132`) → keysign serialization (`build.ts:293`) → the new gate (`build.ts:151`, `build.ts:379`).
- Grepped the entire repo for the permit consumer: `buildEip2612Permit` and the keysign-data decoder.
- Reviewed git history of the CowSwap phase-1/phase-2 series (#584, #636) and the AGG-03 issue (#1103) + the superseded sibling PR #1108.
- Codex lane (gpt-5.4) run adversarially on the same diff; referenced #1105 explicitly, not cross-contaminated.

## blocking

**The permit path this fix assumes exists is not wired — skipping the approve makes permit-token CowSwap orders unfillable.**

The premise of AGG-03 / issue #1103 is that the on-chain `approve()` is *redundant* because a permit-flagged order "settles via a gasless EIP-2612 permit signature bundled into the order digest." That precondition is **not met in the current tree**:

- `buildEip2612Permit` (the only EIP-2612 permit builder, `packages/core/chain/swap/general/cowswap/permit/buildEip2612Permit.ts`) has **zero callers** anywhere in the SDK — it is exported scaffold from the phase-1 commit (#584) and was never wired.
- The CowSwap keysign payload carries only the order + a `permitRequired: true` **boolean flag** (`build.ts:293`, `cowSwapKeysignData.ts`). It carries **no permit typed-data / digest**, so the MPC ceremony has nothing to produce a permit signature from. `buildEip2612Permit` also needs `tokenName`, `nonce`, `deadline` — none of which are carried in the flow.
- The wired consumer path (phase-2, #636) is "sign the order EIP-712 digest → `submitCowSwapOrder`", and `submitCowSwapOrder` posts only `{...order, signature, signingScheme:'eip712', from}` — **no permit payload**. There is no permit pre-hook in `appData` either.
- The allowance mechanism #636 actually wired for CowSwap is the on-chain approve (`buildCowSwapApprovalSigningInput` → `erc20ApprovePayload`). That is exactly what this PR now skips.

Net effect: for a permit-token order the vault relayer ends up with **neither an approve nor a permit** → the solver cannot pull the sell token → the order is submitted but **never settles**, expiring silently at `validTo` (≤60 min). This is the over-skip failure mode: funds are safe (they never move) but the swap silently fails.

This is **mainline, not an edge case** — `KNOWN_PERMIT_TOKENS` is USDC (every supported chain), USDT (Arbitrum, Avalanche), DAI (Arbitrum), i.e. the most common CowSwap sell tokens. The doc comment on `permitRequired` (`GeneralSwapQuote.ts:53`) confirms the intended-but-unimplemented design: "the signing flow **should** build and include a permit signature alongside the order signature."

**Recommended sequencing:** don't land the approve-skip until the permit path is actually wired — `buildEip2612Permit` is invoked, the permit typed-data is added to the keysign payload as a second signable, the MPC signs it, and the consumer bundles it into the order submission (orderbook permit / `appData` pre-hook). Until then the approve is load-bearing, not redundant, and must stay.

## should-fix

**Test coverage proves the wrong invariant.** `build.agg03CowSwapPermit.test.ts` asserts `erc20ApprovePayload` is `undefined` and `getErc20Allowance` was not called for a permit order — i.e. it locks in the skip. It never asserts that the resulting order can still obtain allowance (a permit signable is produced / submitted). That is why all CI is green while the integration is broken — the exact "clean pure-function tests over a broken integration" trap. A test that fails until the permit is actually produced would have caught this.

## q

Is there downstream permit-signing wiring in a consumer repo (windows/ios/extension) that I can't see from the SDK? Even if a consumer wanted to, the current keysign payload carries no permit digest for the MPC to sign, so a permit signature can't be produced from this payload shape — but if you have a companion PR that adds it, link it and I'll re-review the sequencing.

## note on #1108 relationship
The task framed #1108 ("gate the erc-20 approve step on cowswap permitrequired") as a merged sibling. It is **CLOSED, unmerged** (`mergedAt: null`) — #1105 is its successor (closes #1103; #1108 closed #1057, same AGG-03 audit). So there is **no double-gate / conflict** with #1108. The gate logic itself (`matchRecordUnion` over the quote, `false` for every arm except `cowswap_order` with `permitRequired`) is correct and well-scoped: native, 1inch/LiFi/Kyber/SwapKit, and non-permit CowSwap tokens all still build a real approve. The problem is purely that the permit replacement doesn't exist yet.

## risk
High. The gate is correctly scoped and the `permitRequired` trust boundary is sound (static `KNOWN_PERMIT_TOKENS`, not the untrusted API). But shipping it now regresses the most common CowSwap pairs (USDC/USDT/DAI same-chain ERC-20 swaps) from "works, with a wasteful approve" to "silently fails to settle," because the gasless permit that justifies the skip is not implemented.

## verdict
**Request changes.** Hold the approve-skip until the EIP-2612 permit is actually built, signed via MPC, and submitted with the order. The rest of the change (derivation, scoping, non-regression on other providers) is correct and can be kept as-is once the permit path lands.

---
🤖 dual-lane review: Claude (source-trace, load-bearing) + Codex gpt-5.4 (adversarial) — independently corroborated the blocker.

NeOMakinG (approved):
Round 2 re-review @ `6f1a65b8` (was CHANGES_REQUESTED @ `60fd4565`). My r1 blocker is resolved — gomes took the safe path and reverted the premature approve-skip (`028934f1` "revert premature permit gate — keep approve until eip-2612 path is wired"). One consequence of that revert left CI red; I fixed it in-branch (see below) and CI is now green.

## testing approach
- Worktree at `028934f1` then `6f1a65b8`; `git diff 60fd4565..028934f1` + full-repo greps for `buildEip2612Permit` callers and residual `permitRequired` branches.
- Built shared workspace packages and ran the AGG-03 suite locally via `.config/vitest.core.config.ts`.
- Verified the reverted approve guard is **byte-identical** to pre-#1105 `main` (`git merge-base` = `e9855c46`): `if (isChainOfKind(chain, 'evm') && fromCoin.id)` — no subtle regression, clean restore.
- Confirmed CI on `6f1a65b8`: Unit Tests / Integration Tests / Lint+TypeCheck / Test Summary all green.
- Second lane: Codex gpt-5.4 adversarial (diffed `028934f1` vs `origin/main` — scoped to residual inconsistencies, not the reverted logic).

## prior issues — status
| r1 | 1 | blocker | Approve-skip for CowSwap permit tokens (USDC/USDT/DAI) with no permit ever produced (`buildEip2612Permit` zero callers, no permit digest in keysign payload, `submitCowSwapOrder` posts no permit) → order gets neither approve nor permit → solver can't pull tokens → silent non-settlement at `validTo` | ✅ fixed | `028934f1` reverts the `!cowSwapPermitRequired` gate; the approve guard is now byte-identical to `main`. A permit-token order again obtains allowance via the on-chain approve and settles. Verified locally: a `permitRequired:true` order now builds `erc20ApprovePayload{spender: COW_VAULT_RELAYER}`. |
| r1 | 2 | should-fix | Test only asserted "approve is skipped", never that the order can still obtain allowance (green-tests / broken-integration mask) | ✅ fixed | Test now asserts the load-bearing invariant — a `permitRequired` order **still obtains allowance via the approve** (`erc20ApprovePayload.spender === COW_VAULT_RELAYER` + `getErc20Allowance` called). Fixed in `6f1a65b8` (see disclosure). |

## disclosure — I pushed one commit to this branch
The revert (`028934f1`) left `build.agg03CowSwapPermit.test.ts` case 1 still asserting the **old** skip behavior (`erc20ApprovePayload` undefined, `getErc20Allowance` not called). With the gate removed, a `permitRequired` order takes the same approve-building path as a non-permit order, so that assertion failed → **Unit Tests / Test Summary went red on `028934f1`** (confirmed on CI and reproduced locally). I pushed `6f1a65b8` to flip case 1 to assert allowance-is-obtainable-via-approve and refreshed the file's doc comment + describe title. All 13 tests in the swap-build suites pass; CI is now green. Revert if you'd rather own the test edit.

## should-fix
- **Stale PR title + body.** The title still reads "skip redundant approve for cowswap permit orders" and the body still describes *adding* the gate — the opposite of what `028934f1` now does. The `.changeset` was correctly updated to "keep erc20 approve … until permit path is wired"; please align the title/body before merge so the squash-commit message isn't misleading.

## suggestion
- **`permitRequired` is now inert but still exported as "gasless approval".** The flag is still computed in `getCowSwapQuote` and serialized into the keysign `data` blob, and `cowSwapKeysignData.ts` / `buildCowSwapApprovalSigningInput.ts` docs still frame it as gasless-approval intent. No in-repo consumer branches on it, so it's not a functional regression — but a downstream integrator could read the flag and suppress an approve on their side (re-introducing exactly the r1 silent-failure). Consider a one-line doc note that the flag is phase-1 scaffold and does **not** yet replace the approve. (Codex lane.)

## q
- The `@vultisig/sdk: patch` changeset publishes a new version even though the net runtime delta vs `main` is now zero (only a comment + tests). Intentional as an audit-trail release note, or drop the changeset since nothing shipped changes? Non-blocking either way. (Codex lane.)

## risk
Low. The change restores the exact pre-#1105 approve behavior (byte-identical guard) — the well-exercised path that shipped in #636. Fund-safety: an over-approve is impossible here (the guard only builds an approve when on-chain allowance is insufficient, capped at the sell amount). The permit path remains unwired and correctly un-relied-upon.

## verdict
Approve. The r1 fund-safety blocker is resolved by reverting to the on-chain approve, non-permit and non-CowSwap routes are unaffected (regression cases green), the test now asserts the allowance-is-obtainable invariant, and CI is green on `6f1a65b8`. Please fix the stale title/body before merge (should-fix) — not gating.

---
🤖 reviewer lane: Claude (load-bearing source trace + local test run) + Codex gpt-5.4 (adversarial, residual-inconsistency scope). Generated with [Claude Code](https://claude.com/claude-code).

PR #1111, opened 2026-07-09
fix(cosmos): apply IBC gas multiplier + validate source-channel format (cosmos-02, cosmos-03)

### LINKED ISSUE #1062: [audit-COSMOS-02 · COSMOS-03] IBC transfers via the WalletCore path use the flat MsgSend gas limit (no PFM allowance) an
## Context (fund-safety audit round 2, 2026-07-08 · finding COSMOS-02 · COSMOS-03 · Medium)

COSMOS-02: resolvers/cosmos/index.ts:428 uses the flat per-chain getCosmosGasLimit for any IBC message, while the app's own builder documents an IBC_GAS_MULTIPLIER (×2) for "ICS-20 MsgTransfer + optional PFM memo" — never ported to the SDK. Risk: out-of-gas mid-execution on a gas-sensitive chain (fee spent, transfer fails). COSMOS-03: sourceChannel comes from memo.split(':') with no /^channel-\d+$/ check (:85) → empty channel = destination rejection with an opaque error.

## Effort / risk / confidence
Effort S eachRisk LOWConfidence MED

## Prompt for AI agents
Implement the fix for audit finding **COSMOS-02 · COSMOS-03** in `vultisig/vultisig-sdk`. COSMOS-02: resolvers/cosmos/index.ts:428 uses the flat per-chain getCosmosGasLimit for any IBC message, while the app's own builder documents an IBC_GAS_MULTIPLIER (×2) for "ICS-20 MsgTransfer + optional PFM memo" — never ported to the SDK. Risk: out-of-gas mid-execution on a gas-sensitive chain (fee spent, transfer fails). COSMOS-03: sourceChannel comes from memo.split(':') with no /^channel-\d+$/ check (:85) → empty channel = destination rejection with an opaque error.  Make the minimal, additive, fail-closed change at the anchors named above; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI: build the SDK and unit/CLI-drive the affected path with a fixture; assert before (bug) vs after (fixed).). **Grep open PRs in this repo first** — several audit findings already have in-flight PRs; advance/verify an existing one (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **COSMOS-02 · COSMOS-03** (round 2, Medium) · bead `agent-backend-ts-qtymu`. Runtime before+after is the finish line, not unit-tests-green._

gomesalexandre:
closes #1062

## COSMOS-02
An IBC transfer (ICS-20 `MsgTransfer`, optionally PFM-forwarded via memo) does measurably more work on the source leg than a plain bank send — channel-state writes plus a relayer event — so the flat per-chain gas limit (calibrated for `MsgSend`) was undersized, risking an out-of-gas mid-execution failure (fee spent, transfer fails, funds don't move). Added `IBC_GAS_MULTIPLIER = 2n` in `resolveCosmosGasFee`, scoped to `transactionType === IBC_TRANSFER` only — plain sends/wasm executes on IBC-enabled chains keep the calibrated flat fee. Mirrors the app's own existing `IBC_GAS_MULTIPLIER` in `vultiagent-app/src/services/cosmosTx.ts`. Applied identically in both the signing-inputs resolver (what gets signed) and the fee-amount resolver (what gets displayed) so the Network Fee never drifts from the signed tx.

## COSMOS-03
`sourceChannel` was derived from `memo.split(':')` with zero format validation — an empty/malformed channel would sail through to the destination chain and get rejected with an opaque error instead of failing fast locally. Added a `/^channel-\d+$/` check with a clear error message before the channel is used.

## Testing
`yarn build:shared` + `npx vitest run` on the 3 touched suites (74/74 passing) + `npx tsc --project .config/tsconfig.json --noEmit` clean.


packages/core/chain/chains/cosmos/resolveCosmosGasFee.ts
```diff
@@ -29,11 +52,17 @@ export const resolveCosmosGasFee = ({
   gas,
   relayedGasLimit,
   staticGasLimit,
+  isIbcTransfer = false,
 }: ResolveCosmosGasFeeInput): ResolveCosmosGasFeeResult => {
-  const resolvedGasLimit = relayedGasLimit && relayedGasLimit > 0n ? relayedGasLimit : staticGasLimit
+  const effectiveStaticGasLimit = isIbcTransfer ? staticGasLimit * IBC_GAS_MULTIPLIER : staticGasLimit
+  const effectiveGas = isIbcTransfer ? gas * IBC_GAS_MULTIPLIER : gas
+
+  const resolvedGasLimit = relayedGasLimit && relayedGasLimit > 0n ? relayedGasLimit : effectiveStaticGasLimit
```
NeOMakinG, line 60: `should-fix:` (defense-in-depth, non-blocking) for an IBC transfer this still honors a positive `relayedGasLimit` verbatim even when it is *below* the `effectiveStaticGasLimit` (2× floor) — e.g. an externally-supplied `gasLimit=300000` on an IBC_TRANSFER signs `gas=300000`, not the intended `400000`, while the fee still doubles. This is **not reachable via the SDK's own code** (the chainSpecific resolver only sets `gasLimit` for `isNativeSend`, i.e. `transactionType === UNSPECIFIED` — IBC always gets `undefined`, so the full 2× floor applies), and it's not a regression. But since `CosmosSpecific` can be shaped by external/relayed callers, consider clamping for IBC: `resolvedGasLimit = isIbcTransfer ? max(relayed, effectiveStaticGasLimit) : ...` so a low external limit can't undercut the IBC floor — exactly the OOG-and-burn case this PR targets.

packages/core/mpc/keysign/fee/resolvers/cosmos.ts
```diff
@@ -27,12 +28,15 @@ export const getCosmosFeeAmount: FeeAmountResolver = ({ keysignPayload }) => {
...
-    ibcEnabled: ({ gas, gasLimit }) => {
+    ibcEnabled: ({ gas, gasLimit, transactionType }) => {
       const coin = getKeysignCoin<CosmosChain>(keysignPayload)
       const { feeAmount } = resolveCosmosGasFee({
         gas,
         relayedGasLimit: gasLimit,
         staticGasLimit: getCosmosGasLimit(coin),
+        // COSMOS-02: mirror the signing-inputs resolver's IBC gas multiplier
+        // so the displayed Network Fee never drifts from what gets signed.
+        isIbcTransfer: transactionType === TransactionType.IBC_TRANSFER,
```
NeOMakinG, line 39: `should-fix:` (non-blocking) the display resolver applies the multiplier whenever `transactionType === IBC_TRANSFER`, but the signing path takes the fee straight from `authInfoBytes`/Amino fee for the `signDirect`/`signAmino` arms (it never calls `resolveCosmosGasFee` there). So a custom-signed IBC payload with `transactionType` manually set to `IBC_TRANSFER` would show a doubled *displayed* fee while signing the caller-authored fee — a cosmetic display/signed drift. Unreachable through the public SDK surface today (`SignAminoInput`/`SignDirectInput` don't expose `transactionType`, and `buildCosmosPayload` doesn't tag it), so purely a hardening note: either gate the display multiplier to the native (non-custom-sign) path, or document that `transactionType` is only honored for natively-built MsgTransfers.

NeOMakinG (approved):
## testing approach

Two-lane review (Claude load-bearing + Codex gpt-5.4 adversarial second-opinion). Worktree at HEAD `3f84fce4`, `yarn install --immutable` + `node scripts/build-shared-packages.mjs` + `npx vitest run` on all 3 touched suites → **23/23 pass**. CI all green (Unit 18m, Integration 4m50s, Lint&TypeCheck 13m53s, CodeRabbit). Traced the gas-multiplier through the signing-inputs resolver (signed tx) and the fee resolver (displayed fee) to the emitted `TW.Cosmos.Proto.Fee`; cross-checked the `#1083` relationship against merge-base; verified the native-send gating that controls `relayedGasLimit`.

## what this PR actually is (dedup vs #1083)

The title says `cosmos-02, cosmos-03`, but the **net diff vs main is COSMOS-02 only**. The COSMOS-03 channel validator (`/^channel-\d+$/`, fail-closed) is **already in main from #1083** (merge-base `e9855c46` already contains it at `index.ts:92`) — this PR does **not** re-add or modify it. So there is no duplicate/conflicting code; #1111 is a **complementary different-dimension** fix (gas sizing) that composes cleanly on top of #1083's safety guards (channel format + no-timeout). The COSMOS-03 section in the PR body is a stale carry-over describing already-merged work — worth trimming from the description, but not a code concern.

## COSMOS-02 verification (the real content) — PASS

- **IBC-scoped correctly**: `isIbcTransfer = transactionType === IBC_TRANSFER` in *both* resolvers. This is the *exact same condition* that gates `MsgTransfer` construction (`index.ts:84`), so every `MsgTransfer` gets the ×2 and nothing else does — no scope gap where a MsgTransfer builds without the bump. Test confirms a plain send on the same ibc-enabled chain stays `200000/2500`.
- **Sane multiplier**: `IBC_GAS_MULTIPLIER = 2n`, mirroring the app's own validated value (`vultiagent-app/src/services/cosmosTx.ts:276`, applied to both gasLimit and feeAmount). Conservative headroom over the `MsgSend`-calibrated flat 200k; overpay direction is the safe one.
- **Reaches the signed tx**: `resolvedGasLimit` → `TW.Cosmos.Proto.Fee.gas` (`400000`), `feeAmount` → `Fee.amounts` (`5000`). Test asserts it.
- **No display/signed drift** on the native path: a dedicated cross-consistency test asserts `getCosmosFeeAmount === signingInput.fee.amounts[0].amount === 5000`.
- **No regression for non-IBC**: with `isIbcTransfer=false` the math reduces to the pre-PR formula byte-for-byte (`effectiveStatic=static`, `effectiveGas=gas`). PFM-memo transfer also correctly doubled (test).
- **Changeset present**: `.changeset/cosmos02-ibc-gas-multiplier.md` (`@vultisig/sdk` patch), under `packages/core/` bundled-source-prefix → ships live, not inert.

## should-fix

Two defense-in-depth edges (both **non-blocking** — unreachable via the SDK's own code paths today, and neither is a regression). See inline.

## risk

Low. Fund-relevant path (IBC gas sizing) is improved, not regressed: the primary SDK-built IBC path always gets the full 2× floor because the initiator never sets a relayed `gasLimit` for IBC (`isNativeSend` requires `transactionType === UNSPECIFIED`). Non-IBC cosmos txs are mathematically untouched. Verified against real test vectors, CI green, 23/23 local.

## verdict

**APPROVE.** COSMOS-02 gas multiplier is IBC-scoped, sane (×2, mirrors the app), reaches both the signed `Fee.gas` and the displayed fee with a drift-guard test, and does not regress non-IBC or the #1083 channel/no-timeout guards it composes with. The two should-fix items are hardening for externally-crafted payloads (the same trust boundary as signAmino/signDirect) and can land in a follow-up.

---
🤖 dual-lane review (Claude + Codex gpt-5.4). Codex converged on both should-fix edges; no blocking findings from either lane.

PR #1107, opened 2026-07-09
fix(swap): populate 1inch affiliate fee for display (agg-05)

gomesalexandre:
closes #1106

**Update:** the THOR-04 portion of this PR (Solana/Noble limit-swap destination validators) has been reverted. Prior runtime verification (swap-coverage-matrix) established THORChain limit-swap is LP-pool-scoped: Solana and Noble aren't valid limit-swap destinations even though both are valid regular-swap destinations elsewhere, so the missing validators were correct fail-closed behavior, not a bug. See the revert commit for details. This PR now covers AGG-05 only.

## AGG-05 — 1inch quotes never populate affiliateFee for display

Kyber and LI.FI both populate `GeneralSwapQuote.tx.evm.affiliateFee` so the UI can show a fee-transparency row. 1inch never did, despite taking a real affiliate cut via its `fee`/`referrer` params — the row stayed blank for that provider. `getOneInchSwapQuote` now grosses the post-fee `dstAmount` back up (`netAmountOut * 10000n / (10000n - feeRate)`, same calc as `getKyberSwapAffiliateFee` in `kyber/api/tx.ts`) to compute and attach a display-only `affiliateFee`. Added an optional `to: AccountCoin<EvmChain>` param (backward compatible — omitting it just skips the fee calc) and threaded it through from `findSwapQuote.ts`.

## Test plan

- [x] `getOneInchSwapQuote.test.ts` — affiliateFee populated matching the Kyber gross-up calc when `affiliateBps` + `to` are set; stays `undefined` when no `affiliateBps`
- [x] Full `@vultisig/core-chain` suite: 1581 passed, 1 skipped (pre-existing skip, unrelated)
- [x] `tsc --project .config/tsconfig.json --noEmit` — clean
- [x] `eslint` on changed files — clean
- [x] Changeset added (`@vultisig/core-chain` + `@vultisig/sdk` patch)

## Risk

Low. Additive-only change (new optional param, new field population) — no existing Kyber/LiFi/1inch behavior changes.

PR #1112, opened 2026-07-09
fix(swap): resolve thorchain limit-swap destinations via swap-eligible chains (thor-04)

### LINKED ISSUE #1106: fix: limit-swap missing Solana/Noble destinations + 1inch affiliate fee not displayed (THOR-04/AGG-05)
## THOR-04 — limit-swap destination validators missing Solana/Noble

`limitSwapDestinationValidators` in `packages/core/chain/swap/native/limitSwapMemo.ts` (~line 186-207) covers EVM chains, Bitcoin/BCH/Dash/Dogecoin/Litecoin/Zcash, Cosmos/Kujira/THORChain, and Ripple/Tron — but is missing `Chain.Solana` and `Chain.Noble`, even though both are already valid regular-swap destinations elsewhere in the codebase (`NativeSwapChain.ts`). `assertValidLimitSwapDestinationAddress` does `if (!validator) throw`, so a limit swap targeting Solana or Noble currently fails closed with "chain not supported for limit swap destinations".

A deeper related gap surfaced while fixing this: the earlier `getSupportedThorchainAssetChain` prefix-resolution step (which runs before the destination validator is ever reached) also rejects `SOL.*`/`NOBLE.*` target assets, because it resolves the asset's chain prefix via the LP-scoped `lpChainMap` (`chains/cosmos/thor/lp/lpChainMap.ts`) instead of the swap-scoped `nativeSwapChainIds` map (`swap/native/NativeSwapChain.ts`) that already lists Solana (`SOL`) and Noble (`NOBLE`). `lpChainMap` never got those entries because neither chain supports THORChain LP.

## AGG-05 — 1inch quotes never populate affiliateFee for display

1inch general-swap quotes never populate an `affiliateFee` field on the returned `GeneralSwapQuote`, unlike the Kyber and LI.FI swap paths which do. A real affiliate fee IS taken on 1inch swaps (via the `fee`/`referrer` params), but the UI's fee-transparency row stays blank for that provider.

## Scope

Both are audit findings from bead `agent-backend-ts-d89oq`. Fix in progress.

gomesalexandre:
closes #1106

## What

Re-fix of THOR-04 (bead `agent-backend-ts-d89oq`). A prior attempt (`cefe2e5d`, PR #1107's earlier scope) was reverted (`9559b50a`) as "non-bug" — that revert's runtime check tested the wrong authority map. Two independent follow-up investigations (a deep code read + a Codex review sourced from THORChain's protocol memo docs at dev.thorchain.org/concepts/memos.html) both concluded the revert was wrong: rejecting Solana/Noble as THORChain limit-swap destinations is a false-block on legit swaps, not correct protocol enforcement.

`getSupportedThorchainAssetChain` (`limitSwapMemo.ts`) resolved a THORChain asset-prefix (e.g. `SOL`, `NOBLE`) to a `Chain` via `lpChainMap` — a map scoped to the wallet's LP-position-display feature (per its own docstring), unrelated to swap eligibility. Neither Solana nor Noble has a THORChain LP pool, so both were missing, and a limit swap (`LIM=` memo) targeting either failed closed with "unsupported THORChain asset prefix" even though both are valid THORChain market-swap destinations elsewhere in the codebase (`thorChainSwapEnabledChains`, `NativeSwapChain.ts`). Per THORChain's memo docs, `=` vs `=<` only changes execution behavior (price/queue/TTL), not the destination-chain universe.

## How

- Exported `thorChainSwapEnabledChains` from `NativeSwapChain.ts` (the THORChain-specific chain set regular market swaps already gate against).
- `getSupportedThorchainAssetChain` now resolves via a union of `lpChainMap` and `thorChainSwapEnabledChains`, instead of `lpChainMap` alone or the broader `nativeSwapChainIds`.
  - **Why a union, not a straight swap of authority**: `thorChainSwapEnabledChains` alone is missing Dash/Kujira/Arbitrum/Zcash, which `lpChainMap` already resolves correctly (proven by pre-existing passing tests using `DASH.DASH`/`ZEC.ZEC` as limit-swap source assets — a straight swap regressed those in local testing).
  - **Why not the broader `nativeSwapChainIds`** (what the original reverted commit used): it also carries MayaChain-only entries (`Chain.MayaChain` itself, `Chain.Cardano`), which are not valid THORChain limit-swap destinations. Codex flagged this as an over-widening risk on the original attempt.
- Added `Chain.Solana`/`Chain.Noble` destination-address validators to `limitSwapDestinationValidators`. Solana validates via `@solana/web3.js`'s `PublicKey` (the pattern already used elsewhere in this package, e.g. `injectSolanaAtaIfMissing.ts`) rather than a bare base58-length regex, after a Codex review pass on this PR flagged that a regex alone accepts non-decodable strings.

## Test plan

- [x] New tests: Solana/Noble limit-swap destinations now succeed; a base58 string that isn't a valid Solana public key is rejected; MayaChain-only destinations (`ADA.*`, `MAYA.*`) are still correctly rejected as THORChain limit-swap destinations (no over-widening).
- [x] Pre-existing `DASH.DASH`/`ZEC.ZEC` source-asset tests still pass (regression check for the union approach vs a straight authority swap).
- [x] Full `@vultisig/core-chain` suite: 1602 passed, 1 skipped (pre-existing skip, unrelated).
- [x] `tsc --project .config/tsconfig.json --noEmit` — clean.
- [x] `eslint` on changed files — clean.
- [x] `/codex-review` — 1 finding (Solana regex too loose), fixed in a follow-up commit on this branch.
- [x] Changeset added (`@vultisig/core-chain` + `@vultisig/sdk` patch).

## Risk

Low. Additive resolution-authority widening for two specific chains, backstopped by existing per-chain destination validators. No existing supported chain's behavior changes (verified via the DASH/ZEC regression check above).


NeOMakinG (approved):
**APPROVE** — THOR-04 limit-swap destination-chain resolution fix is correct and fund-safe. Two-lane review (Claude load-bearing + Codex gpt-5.4 adversarial); both converge with zero findings.

## testing approach
- Read the full HEAD (`f5b4069a`) of `limitSwapMemo.ts` + the off-diff source that governs correctness: `NativeSwapChain.ts` (`thorChainSwapEnabledChains` + `nativeSwapChainIds`), `lpChainMap.ts`, `thorchainLp.ts` (`thorchainLpChainCode`), `pools.ts` (`assertValidPoolId` / `POOL_ID_RE`).
- Standalone reproduction of the union-map logic: enumerated every reachable chain, checked validator coverage, prefix collisions, and the SOL/NOBLE/ADA/MAYA resolution outcomes.
- CI at this exact HEAD: Unit Tests, Integration Tests (Vitest), Lint & Type Check, Test Summary, CodeRabbit — all SUCCESS (the four new THOR-04 tests run in that suite).

## fund-safety verification (all PASS)
1. **Swap-eligible set correct.** Resolution map = `{...lpChainMap, ...thorChainSwapEnabledChains→nativeSwapChainIds}`. `thorChainSwapEnabledChains` is the *same* list regular THORChain market swaps already use — so limit swaps now share the market-swap destination universe (the stated THOR-04 invariant). The union preserves the LP-pool-only chains (Dash/Kujira/Arbitrum/Zcash) that were already accepted pre-fix, and deliberately excludes the broader `nativeSwapChainIds`/Maya-only entries (Cardano, MayaChain). Tests assert `ADA.ADA` and `MAYA.CACAO` still throw `unsupported THORChain asset prefix`.
2. **No wrong-chain resolution.** Output chain is pinned by the `target_asset` prefix, not guessed from address shape. `nativeSwapChainIds` and `thorchainLpChainCode` agree on every overlapping prefix, so the union has **zero prefix collisions to a different `Chain`** (verified programmatically) — merge order is immaterial. The same `target_asset` used to pick the validator is what goes verbatim into the signed memo (`=<:${target_asset}:${dest_addr}:${LIM}/...`).
3. **Case handling consistent.** Lookup uses `prefix.toUpperCase()`; every `lpChainMap` key and every `nativeSwapChainIds` value is uppercase, and `POOL_ID_RE` already forces uppercase `CHAIN.ASSET`. Identical to the pre-fix `chainPrefixToChain` behavior — no casing regression for the previously-working chains.
4. **Fail-closed.** Unresolvable prefix → throw. Every one of the 18 reachable chains has a `limitSwapDestinationValidators` entry (verified), and a missing validator would still throw (`assertValidLimitSwapDestinationAddress` L252) — fail-closed by construction, no fail-open accept. Invalid address for the resolved chain → throw.
5. **Reaches the signable memo.** `buildLimitSwapMemo` → `validateLimitSwapInputs` (validates source + target chain + dest addr) runs *before* `buildMemo` interpolates the segments; throws propagate, so no ineligible/mis-validated destination can produce a memo. `POOL_ID_RE` (uppercase alnum + `.`/`-` only) blocks `:`/`/` injection via `target_asset`; `assertMemoSegmentSafe` guards `dest_addr`.
6. **No over-restriction.** Solana + Noble limit-swap destinations now build (they had no LP pool so were wrongly rejected before); the four new tests cover accept-Solana, accept-Noble, reject-invalid-base58, reject-Maya-only.

## suggestion
- `isSolanaAddress` accepts any 32-byte base58 value (`PublicKey` does not enforce on-curve), so off-curve/PDA-shaped inputs pass. Not a mis-route or fund risk — a destination only needs to be a valid 32-byte account and the user supplies their own address — but worth a note if stricter validation is ever desired.
- No regression test asserts the union still resolves the LP-pool-only chains (Dash/Kujira/Arbitrum/Zcash); a one-line case would lock in the "don't regress" guarantee the changeset comment describes.

## risk
Low. Behavior change is strictly a *widening* of accepted limit-swap destinations to match the already-shipped market-swap eligibility set, plus tighter Solana address validation. No change to memo construction, LIM math, or signing. `@vultisig/core-chain` + `@vultisig/sdk` patch changeset present (non-inert).

## verdict
Approve. Swap-eligible set correct, destination resolves to the right chain with no collisions/wrong-chain output, fail-closed on ineligible/unresolvable, reaches the signable memo, no over-restriction, changeset present, CI green.

---
🤖 dual-lane review (Claude load-bearing + Codex gpt-5.4). Both lanes: no blocking findings.

PR #1113, opened 2026-07-09
fix(vault): route the send-path amount parser through toChainAmount for swap-path parity (gbbo9)

### LINKED ISSUE #1070: [audit-SDK-CORRECTNESS-03 · 05] send() uses a hand-rolled parseAmount that diverges from the hardened toChainAmount
## Context (fund-safety audit round 1, 2026-07-08 · finding SDK-CORRECTNESS-03 · 05 · Low)

VaultBase.ts:1909-1916 validates with lossy Number(trimmed) then rebuilds via BigInt, while the swap path uses viem parseUnits. On the agent-reachable vault.send(): scientific notation ("1e2") passes the !isNaN gate then throws a raw SyntaxError (escaping the VaultError contract MCP callers rely on), and excess fractional digits are silently floored. Conversion is string-based so precision-exact for normal input. Route through toChainAmount inside the existing VaultError catch. Related: toChainAmount.ts:101-103's number-input branch is lossier than its string branch.

## Effort / risk / confidence
Effort SRisk LOWConfidence HIGH

## Prompt for AI agents
Implement the fix for audit finding **SDK-CORRECTNESS-03 · 05** in `vultisig/vultisig-sdk` at the anchors named above. VaultBase.ts:1909-1916 validates with lossy Number(trimmed) then rebuilds via BigInt, while the swap path uses viem parseUnits. On the agent-reachable vault.send(): scientific notation ("1e2") passes the !isNaN gate then throws a raw SyntaxError (escaping the VaultError contract MCP callers rely on), and excess fractional digits are silently floored. Conversion is string-based so precision-exact for normal input. Route through toChainAmount inside the existing VaultError catch. Related: toChainAmount.ts:101-103's number-input branch is lossier than its string branch. Minimal, additive, fail-closed; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI build + unit/CLI drive; before vs after.). **Grep open PRs in this repo first** — advance an existing PR (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **SDK-CORRECTNESS-03 · 05** (round 1, Low) · bead `agent-backend-ts-gbbo9`._

gomesalexandre:
closes #1070

## What
\`VaultBase.parseAmount\` (the send/transfer amount converter) used \`toBaseUnits\` (pure string arithmetic, truncates excess fraction digits, rejects scientific notation), while the swap path (\`validateHumanSwapAmount\`, \`SwapService\`) used \`toChainAmount\` (viem \`parseUnits\`, rounds excess digits, expands scientific notation). Both were lossless/fund-safe — max divergence is 1 base-unit of dust on excess fraction digits — but the two paths used different converters for identical human-amount → base-unit semantics.

## How
\`parseAmount\` now shares the swap path's \`toChainAmount\`. Two intended, safe behavior changes fall out:
- Excess-fraction amounts (e.g. \`0.123456789\` at 8 decimals) now round instead of truncate (max 1-unit dust delta).
- Scientific-notation amounts (e.g. \`1e18\`) are now accepted on the send path instead of rejected — send stricter-than-swap was never an intentional security boundary, just converter drift.

Error-throwing behavior, messages, and the \`<= 0n\` validation are unchanged. The now-unused \`toBaseUnits\` import is removed from \`VaultBase.ts\` (still used elsewhere in the codebase, untouched there).

## Testing
\`npx vitest run packages/sdk/tests/unit/vault/compound-wrappers.test.ts\` — 358/358 passing, including 2 updated tests documenting the deliberate behavior changes. \`npx tsc --project .config/tsconfig.json --noEmit\` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Send/transfer amount parsing now matches swap-path conversion, ensuring on-chain amounts never exceed the entered human-readable amount.
  * Extra fractional digits are truncated (floored) instead of rounded up.
  * Scientific-notation amounts (for example, `1e18`) are now accepted on send paths.

* **Tests**
  * Added assertions covering truncation behavior and scientific-notation support for amount parsing.


packages/sdk/src/vault/VaultBase.ts
```diff
@@ -1925,7 +1924,7 @@ export abstract class VaultBase extends UniversalEventEmitter<VaultEvents> {
     if (!trimmed) throw new VaultError(VaultErrorCode.InvalidAmount, 'Amount cannot be empty')
     let chainAmount: bigint
     try {
-      chainAmount = BigInt(toBaseUnits(trimmed, decimals))
+      chainAmount = toChainAmount(trimmed, decimals)
```
NeOMakinG, line 1927: `preferably-blocking:` viem `parseUnits` rounds half-up, so the signed send amount can exceed the stated amount. At `decimals = 0` (reachable via Cardano native assets, `resolvers/cardano.ts:18` `registry?.decimals ?? 0`, and low-decimal tokens) this is fund-unsafe: `0.5`/`0.6`/`0.999` went from **rejected** (`<= 0n`, fail-closed) to **signing 1 whole token**; `1.5` went from signing `1` to signing `2` (doubles); `2.5` → `3`. The old `toBaseUnits` truncated — the conservative direction (result never exceeds the stated amount). To keep the parity goal in the safe direction, make the shared converter **truncate/floor** (matches `toBaseUnits`' documented "can't send fractions of the smallest unit" semantics), or reject over-precise input (fraction digits > decimals) on the shared path.

.changeset/gbbo9-send-amount-parity.md
```diff
@@ -0,0 +1,7 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(vault): route the send-path amount parser through `toChainAmount` for swap-path parity (gbbo9)
+
+`VaultBase.parseAmount` (the `send`/`transfer` amount converter) used `toBaseUnits` (pure string arithmetic, truncates excess fraction digits, rejects scientific notation), while the swap path (`validateHumanSwapAmount`, `SwapService`) used `toChainAmount` (viem `parseUnits`, rounds excess digits, expands scientific notation). Both were lossless/fund-safe — max divergence is 1 base-unit of dust on excess fraction digits — but the two paths used different converters for identical human-amount → base-unit semantics. `parseAmount` now shares the swap path's `toChainAmount`, so both paths behave identically. Two intended, safe behavior changes fall out: excess-fraction amounts (e.g. `0.123456789` at 8 decimals) now round instead of truncate (1-unit dust), and scientific-notation amounts (e.g. `1e18`) are now accepted on the send path instead of rejected. Error-throwing behavior, messages, and the `<= 0n` validation are unchanged.
```
NeOMakinG: `should-fix:` "max divergence is 1 base-unit of dust" is materially false at `decimals = 0`, where 1 base unit is a whole token and a previously-rejected fractional input now signs (e.g. `0.6` → signs `1`, `1.5` → signs `2`). Please correct this and add a `decimals = 0` test — current tests only cover the 8-decimal dust case and `1e5`.

NeOMakinG (changes requested):
Fund-safety review of the send-path amount parser. Two-lane: Claude (load-bearing, worktree `15ecc154` + standalone viem@2.55 harness replicating both converters) + Codex `gpt-5.4` (adversarial second opinion, confirmed on-PR).

Most of this PR is clean and the parity goal is legitimate — but the shared converter rounds in the **send-more-than-typed** direction, which at low/zero token decimals turns a previously fail-closed reject into a signed over-transfer. That, plus a factually-wrong safety claim in the changeset, is why I'm asking for a change before merge.

## testing approach
- Read both converters: `toChainAmount` (viem `parseUnits`, exact bigint; pure-bigint sci-notation expander with `MAX_SCALE_ABS=10000n` DoS cap) and the old `toBaseUnits` (pure string arithmetic, lossless, truncates, rejects sci-notation). **Neither used float** — the "precision loss" framing does not apply; the real divergence is truncate-vs-round + sci-notation accept/reject, exactly as the PR body states.
- Replicated both converters in a standalone viem@2.55 harness and diffed old-vs-new across decimals `{0,6,8,18}` and ~40 adversarial inputs.
- Confirmed the signable path: `parseAmount` → `amountBigInt` → `prepareSendTx` → `sign` → `broadcastTx` (`VaultBase.ts:1693–1710`). The parsed bigint IS the signed transfer amount.
- Confirmed parity: `parseAmount` (`VaultBase.ts:1927`) and `validateHumanSwapAmount` (`:1943`) now call byte-identical `toChainAmount(trimmed, decimals)`; `SwapService.ts:86` / `prep/swap.ts:62` likewise.

### what's verified GOOD
- **`toChainAmount` is exact** — viem `parseUnits` is string→bigint, no `Number()*10**d` on the value path; sci-notation expands via pure bigint. High-decimal (18) / large / sub-unit (`0.000000000000000001`) all exact.
- **Parity is real** — for every input with fraction digits ≤ decimals, old and new produce the identical bigint. Send now equals swap.
- **No new fail-open into garbage** — every invalid input (`1e`, `e5`, `1,000`, `0x10`, `abc`, `NaN`, `1.2.3`, `1e999999`, `1e-999999`) still throws in BOTH; the `MAX_SCALE_ABS` cap holds. The only accept/reject flips are legitimate scientific-notation inputs (intended).
- Changeset present (`@vultisig/sdk` patch — non-inert; `packages/sdk` is outside the bundled-changeset guard so the explicit changeset is required and is here). CI: Lint+TypeCheck, Integration (Vitest), CodeRabbit green; Unit pending at review time.

## preferably-blocking
**Rounding sends MORE than the stated amount; at low/zero decimals this converts a fail-closed reject into a signed over-transfer.** viem `parseUnits` rounds half-up, so the parsed (signed) amount can exceed what the user typed. At high decimals it's ≤1 base-unit of dust, but at `decimals = 0` (reachable — Cardano native assets default `registry?.decimals ?? 0`, `resolvers/cardano.ts:18`; also low-decimal tokens) 1 base unit = **1 whole token** and the direction is fund-unsafe:

| input | decimals | old (truncate) | new (round) |
|-------|----------|----------------|-------------|
| `0.5` / `0.6` / `0.999` | 0 | **rejected** (`<= 0n`, fail-closed) | **signs 1 whole token** |
| `1.5` | 0 | signs `1` | **signs `2`** (doubles) |
| `2.5` | 0 | signs `2` | **signs `3`** |

All three moves are in the send-more-than-typed direction. The old truncation was the conservative direction (result ≤ stated amount); rounding trades that for parity with swap's behavior. The safest fix that PRESERVES the parity goal is to make the shared converter **truncate/floor** rather than round — send==swap parity in the conservative direction, matching `toBaseUnits`' own documented on-chain semantics ("you cannot send fractions of the smallest unit"). Alternatively, reject over-precise input (fraction digits > decimals) on the shared path (fail-closed parity). Either resolves this; the current PR resolves parity toward the less-safe direction.

## should-fix
The changeset's "max divergence is 1 base-unit of dust" is materially false at `decimals = 0`, where the divergence is up to a full token AND a previously-rejected input now signs. Please correct the wording and add a `decimals = 0` test case documenting the intended reject→sign-1 / floor→ceil behavior — the current tests only cover the 8-decimal dust case and `1e5`.

## risk
Narrow blast radius (0-/low-decimal sendable assets, fractional input) and bounded magnitude (<1 unit over), but it lands on the signable send amount in the fund-unsafe direction, and the SDK is MCP/agent-reachable so an arbitrary human amount string reaches this parser directly. Fail-safe elsewhere (the `<= 0n` guard, DoS cap, all-garbage-throws) is intact.

## verdict
REQUEST_CHANGES. The parity approach and the exactness work are sound; the blocker is the rounding *direction* on the signed send amount at low/zero decimals plus the understated changeset claim. Flip the shared converter to truncate (or reject over-precise), fix the changeset wording, add a `decimals = 0` test — then this is a clean approve.

<sub>🤖 two-lane review (Claude load-bearing + Codex gpt-5.4 adversarial). Harness: standalone viem@2.55 replica of both converters, old-vs-new diff across decimals {0,6,8,18}.</sub>

packages/core/chain/amount/toChainAmount.ts
```diff
@@ -94,11 +110,11 @@ export const toChainAmount = (amount: string | number, decimals: number) => {
...
-      return parseUnits(expanded, decimals)
+      return parseUnits(truncateToDecimals(expanded, decimals), decimals)
     }
-    return parseUnits(trimmed, decimals)
+    return parseUnits(truncateToDecimals(trimmed, decimals), decimals)
   }
   const str = amount.toString()
   const value = /[eE]/.test(str) ? amount.toFixed(decimals) : str
-  return parseUnits(value, decimals)
+  return parseUnits(truncateToDecimals(value, decimals), decimals)
```
NeOMakinG, line 119: `should-fix:` The number branch here still isn't safe for `|amount| >= 1e21`: `(1.9e21).toFixed(0) === \"1.9e+21\"` (JS emits exponent form at `>=1e21`), so `truncateToDecimals(\"1.9e+21\", 0)` slices to `\"1\"` and `parseUnits(\"1\", 0)` returns `1n`. Harness confirms this **newly** turns a fail-closed throw into a silently-wrong value (pre-PR `parseUnits(\"1.9e+21\",0)` threw; `2.5e22`@0 -> `2n` likewise). It's an under-send (fund-safe direction) and not reachable in the signed send path today (that path always passes a string; no current `toChainAmount(<number>)` caller passes `>=1e21`), so it doesn't block. Fix: route number exponent-form through `expandScientificNotationToDecimalString` (as the string branch does) before truncating, rather than `toFixed(decimals)`.

NeOMakinG (approved):
**Round 2 re-review** of the send/swap `toChainAmount` parity change, at HEAD `d5406f79` + my in-branch test fix `469cddd3`. My r1 was `CHANGES_REQUESTED` (review 4659439261 @ `15ecc154`) over a fund-unsafe rounding direction. The fix is correct; I fixed one stale test that would have failed CI, and flag one latent (unreachable-today) number-path defect surfaced by the second lane.

## testing approach
- Two-lane: Claude (load-bearing) + Codex `gpt-5.4` (second-opinion; ran on a stale local checkout, reasoned from the round-2 diff — disclosed and verified on-PR).
- Standalone `viem@2.53.1` harness replicating `truncateToDecimals` + all three `parseUnits` call sites, diffed old-vs-new across `decimals={0,2,3,8,18}`.
- Verified the changed assertions in `chainAmount.test.ts` all pass; confirmed `compound-wrappers.test.ts` still asserted the pre-fix round-up value (would fail CI) and corrected it in-branch.
- Audited every `toChainAmount` caller for the number-arg / `>=1e21` reachability of the Codex finding.

## prior issues — status
| 1 | 1 | preferably-blocking | `parseUnits` rounds half-up -> signed send **exceeds** typed amount; at `decimals=0` (Cardano native assets `registry?.decimals ?? 0`) `"0.6"` -> signs `1` whole token, `"1.5"` -> signs `2` (over-send / fail-open) | ✅ fixed | `d5406f79` adds `truncateToDecimals` flooring **before** `parseUnits` at all 3 sites. Harness: `toChainAmount("0.6",0)=0n`, `("1.5",0)=1n`, `("0.5",0)=0n`, `("2.5",0)=2n` — round-up gone, result always <= stated. Send path always passes strings -> fully covered. |
| 1 | 2 | should-fix | changeset "max 1 base-unit of dust" claim materially false at `decimals=0` | ✅ fixed | Changeset rewritten in `d5406f79` (now: "truncates (floors)... can never exceed the stated human amount"). The **same** false "by rounding / max 1 base-unit-of-dust" claim survived in `compound-wrappers.test.ts` — corrected in `469cddd3`. |

## blocking
None.

## should-fix
- **`number` + scientific-notation path is not made safe by the fix, `>=1e21` (latent — unreachable in the signed send path today).** `(1.9e21).toFixed(0) === \"1.9e+21\"`, so the number branch feeds `\"1.9e+21\"` into `truncateToDecimals`, which at `decimals=0` slices to `\"1\"` -> `parseUnits(\"1\",0) = 1n`. Harness confirms this **newly** converts a fail-closed throw into a silently-wrong value: pre-PR `parseUnits(\"1.9e+21\",0)` **threw**; post-PR returns `1n` (`2.5e22` -> `2n` likewise). It is an *under*-send (fund-safe direction) and **not reachable today** — `VaultBase.parseAmount` always passes a string, and no current `toChainAmount(<number>)` caller passes `>=1e21` (TON config uses `0.01`/`0.08`; swapkit routes through `value.toString()` / `BigInt`). Not blocking, but it's a genuine latent silent-corruption the change introduces. Fix: route number exponent-form through the same `expandScientificNotationToDecimalString` the string branch uses, instead of `toFixed(decimals)`, then truncate. (Codex lane.)

## suggestion
- **Leading-dot inputs at `decimals=0`.** `truncateToDecimals(\".5\", 0)` -> `\"\"` -> `parseUnits(\"\")` throws, whereas `\"0.5\"` floors to `0n`. This is fail-closed (safe), just an observable divergence between `\".5\"` and `\"0.5\"`. Normalize leading-dot before parsing if the SDK means to accept it. (Codex lane.)

## q
None.

## risk
Low. The reachable send path (string input) is now fund-safe in the truncation (floor) direction: signed amount is always `<= stated` for any decimals, including `decimals=0` where 1 unit = 1 whole token. Swap-path parity is preserved — send (`VaultBase.parseAmount`) and swap (`SwapService`, `prep/swap`, `validateHumanSwapAmount`) all route through the same now-flooring `toChainAmount`, so equal inputs yield equal base units. Normal amounts with `frac digits <= decimals` are unaffected (floor is a no-op): `\"1.5\"@18` and `\"1.234567890123456789\"@18` parse exactly. The one residual (number+sci `>=1e21`) is latent and in the safe direction.

## verdict
**Approved.** The r1 fund-safety blocker is resolved on the reachable path (floor, not round; over-send eliminated), swap parity intact, normal amounts exact. I pushed `469cddd3` fixing the stale `compound-wrappers.test.ts` assertion (`12345679` -> `12345678`) that would otherwise fail CI and still carried the corrected-elsewhere false rounding claim. The number+sci `>=1e21` hardening (should-fix) can land here or as a follow-up — it does not affect the signable send path.

---
*Reviewed by the vultisig-ops reviewer agent (Claude + Codex `gpt-5.4`). Harness-backed; no device runtime (pure conversion logic).*

NeOMakinG (approved):
Re-approve at HEAD `2372dfff` — drift re-verify after my prior APPROVE at `469cddd3`.

## testing approach
Confirmed the single drift commit `2372dfff` ("prettier format chainAmount.test.ts") is purely formatting, with no impact on the fund-safe truncation semantics I approved.

- `git diff -w 469cddd3..2372dfff` → **EMPTY** (zero non-whitespace changes). The only edit is collapsing trailing alignment spaces before `//` comments in `chainAmount.test.ts`.
- Every truncation assertion **value is unchanged**: `toChainAmount('0.5',0)→0n`, `('0.6',0)→0n`, `('0.999',0)→0n`, `('1.5',0)→1n`, `('2.5',0)→2n`, `(1,0)→1n`, `(2,0)→2n`.
- Production is **byte-identical** to what I approved: production-file diff `469cddd3..2372dfff` is empty; `toChainAmount.ts` still floors via `truncateToDecimals` before `parseUnits` at all three sites (sci-expanded, plain string, number), decimals=0 → strips the fractional part. The fund-safe floor (result ≤ stated) holds; send-path still routes through `toChainAmount` for swap parity.
- CI green at this exact HEAD: Unit Tests, Integration (Vitest), Lint/TypeCheck, Test Summary, CodeRabbit all SUCCESS on run 29026650228.

## risk
None from this drift. It is a whitespace-only reformat of a test file; the approved truncate-not-round semantics are intact byte-for-byte. PR is 10 commits behind main but MERGEABLE with no conflict.

## verdict
Approved. The drift is purely prettier whitespace; fund-safe truncation semantics unchanged from my prior approval.

PR #1114, opened 2026-07-09
test(sdk): loud skip-warning + always-on synthetic signing round-trip (SDK-TEST-02/03)

### LINKED ISSUE #1069: [audit-SDK-TEST-02 · 03] The signing round-trip is only covered by a vault-gated E2E that silently passes green w
## Context (fund-safety audit round 1, 2026-07-08 · finding SDK-TEST-02 · 03 · Medium)

tests/e2e/fast-signing.test.ts:37 is describe.skipIf(!HAS_TEST_VAULT_FIXTURE), and CI never runs test:e2e — an absent fixture is a GREEN pass, so the payload→MPC-sign→compile→broadcast round-trip can silently regress. Separately, the strong compileTx.golden.test.ts covers EVM/Solana/Cosmos/Cardano/Bittensor/QBTC but not Polkadot, Ripple, Ton, Tron, or THORChain MsgDeposit — the chains most exercised by swaps. Add a CI round-trip vs a mocked relay + the public test mnemonic; extend the golden pattern to the missing chains.

## Effort / risk / confidence
Effort MRisk MED — needs a safe non-funded fixtureConfidence HIGH

## Prompt for AI agents
Implement the fix for audit finding **SDK-TEST-02 · 03** in `vultisig/vultisig-sdk` at the anchors named above. tests/e2e/fast-signing.test.ts:37 is describe.skipIf(!HAS_TEST_VAULT_FIXTURE), and CI never runs test:e2e — an absent fixture is a GREEN pass, so the payload→MPC-sign→compile→broadcast round-trip can silently regress. Separately, the strong compileTx.golden.test.ts covers EVM/Solana/Cosmos/Cardano/Bittensor/QBTC but not Polkadot, Ripple, Ton, Tron, or THORChain MsgDeposit — the chains most exercised by swaps. Add a CI round-trip vs a mocked relay + the public test mnemonic; extend the golden pattern to the missing chains. Minimal, additive, fail-closed; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI build + unit/CLI drive; before vs after.). **Grep open PRs in this repo first** — advance an existing PR (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
_Audit finding **SDK-TEST-02 · 03** (round 1, Medium) · bead `agent-backend-ts-uol9i`._

gomesalexandre:
closes #1069

**SDK-TEST-02/03**: the only test exercising a full sign->format->verify round trip (`tests/e2e/fast-signing.test.ts`) is `describe.skipIf(!HAS_TEST_VAULT_FIXTURE)`. Without a provisioned vault fixture it reports "skipped" in vitest's own summary line, but nothing else about the run signals that - the daily cron and any local/CI run exits 0 either way, so "all green" doesn't distinguish "signing was verified" from "16 tests never ran."

First commit here tried fixing this with a `console.warn` banner at module-load time. Verified via direct `tsx` import that the message string was correct, but empirically confirmed (by grepping actual vitest output, not just assuming) that **vitest drops plain console output emitted while a file's tests are all skipped** - it never showed up under `--reporter=default` or `--reporter=verbose`. That would have shipped a fix that looked right in review but did nothing at runtime. Second commit replaces it with a mechanism actually verified against vitest's own reporter output:

1. `fast-signing.test.ts` gets one canary `it()` that runs unconditionally (never itself skipped by the outer `describe.skipIf`) and calls `ctx.skip(NO_TEST_VAULT_SKIP_REASON)` when the fixture is missing. Vitest's verbose reporter prints a skip *reason* inline next to the test name - `describe.skipIf` does not.
2. `tests/e2e/vitest.config.ts` now uses the `verbose` reporter (matching `tests/unit/vitest.config.ts`'s existing choice), so that reason is actually printed by `yarn test:e2e:fast-signing` without anyone having to pass `--reporter=verbose` by hand.
3. New `tests/unit/crypto/signingRoundTrip.synthetic.test.ts`: signs with an ephemeral secp256k1/ed25519 keypair, runs the result through the SAME `formatSignature()` adapter real keysign output uses, then independently verifies the *formatted* signature against the public key with the underlying curve library (`@noble/curves`, already used elsewhere in SDK src). Unlike the existing `formatSignature` unit tests (which feed hand-typed r/s/der strings never produced or checked by real curve math), this exercises real signing+verification end to end, with zero vault or network dependency - it runs on every PR regardless of fixture availability.

## Before vs after

Before (no `TEST_VAULT_PATH` set):
```
Test Files  1 skipped (1)
     Tests  16 skipped (16)
```
No indication anywhere in the run of *why*, or that this is the only suite that would have caught a real signing regression.

After:
```
↓ |e2e| fast-signing.test.ts > (skip banner) reports why the real vault.sign() round trip below did or did not run  [No test vault fixture at .../test-vault.vult - the real vault.sign() 2-of-2 MPC round trip was NOT exercised in this run. See packages/sdk/tests/e2e/SECURITY.md to provision TEST_VAULT_PATH / TEST_VAULT_PASSWORD. Always-on synthetic crypto round-trip coverage: tests/unit/crypto/signingRoundTrip.synthetic.test.ts.]
...
Test Files  1 passed (1)
     Tests  17 skipped (17)
```
And the new synthetic test, which always runs regardless of fixture:
```
✓ signing round trip (synthetic, non-vault-gated) — SDK-TEST-02/03 > ECDSA: ephemeral secp256k1 keypair signs, SDK adapter formats, signature verifies
✓ signing round trip (synthetic, non-vault-gated) — SDK-TEST-02/03 > EdDSA: ephemeral ed25519 keypair signs, SDK adapter formats, signature verifies
✓ signing round trip (synthetic, non-vault-gated) — SDK-TEST-02/03 > sanity check: a tampered signature fails verification (proves the verify path is not a no-op)
```

## Verification

- `yarn workspace @vultisig/sdk test:unit` - 177 files / 2716 tests pass (includes the new synthetic round-trip test).
- `yarn workspace @vultisig/sdk typecheck` - clean.
- `yarn workspace @vultisig/sdk test:e2e:fast-signing` run with no `TEST_VAULT_PATH`/`TEST_VAULT_PASSWORD` set - confirmed the skip reason now prints inline (see above), confirmed it flips to a real pass when a fixture file is present.
- CI behavior for environments without vault creds is unchanged (still exits 0, still "skipped") - only the visibility of *why* changed, so this doesn't break anyone's existing green CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved E2E test output so skipped signing tests now show a clear reason inline instead of appearing as a quiet pass.
  * Added stronger signing regression coverage to catch formatting and verification issues for both ECDSA and EdDSA flows.

* **Tests**
  * Expanded unit and E2E coverage around signing behavior and skip reporting.
  * Updated test runner output to be more verbose for skipped cases.


NeOMakinG (approved):
**APPROVE** — genuine test-infra, verified by running the branch + mutation-testing the signing adapter (2-lane: Claude + Codex gpt-5.4).

## testing approach
Built the branch in an isolated worktree (`yarn install` + `yarn build:shared`), ran the new + full unit suite, mutation-tested `formatSignature` to prove the round-trip catches real regressions, and probed the vitest verbose-reporter skip behavior in isolation. Codex ran as an independent second-opinion lane.

## what I verified (all pass)
- **Synthetic round-trip GENUINELY exercises signing — not inert.** Generates ephemeral secp256k1/ed25519 keypairs, signs, feeds r/s/der through the REAL `formatSignature()` adapter, then cryptographically `verify()`s the *formatted* output. Mutation-proven: EdDSA `r+s`→`s+r` swap makes the EdDSA test go RED; a DER→compact-r||s swap makes the (now-hardened) ECDSA test go RED. A real adapter regression fails on every PR.
- **Always-on.** `tests/unit/crypto/…` — no `skipIf`, no env gate. 3/3 pass locally; full unit suite 2716/2716; CI Unit Tests SUCCESS.
- **Loud skip-warning actually fires.** The `(skip banner)` canary sits OUTSIDE `describe.skipIf(!HAS_TEST_VAULT_FIXTURE)`, calls `ctx.skip(reason)`, and `reporters:['verbose']` surfaces the reason inline. Confirmed live: e2e run prints the full "…MPC round trip was NOT exercised…" reason at top while the 16 real signing tests show a silent `↓`. Isolated probe confirms the default reporter hides it — the verbose switch is load-bearing, not a no-op.
- **Correct synthetic crypto** (secp256k1+sha256 prehash for ECDSA; ed25519 raw-message for EdDSA) and **no smuggled production change** — all 4 files under `tests/`. Test-only ⇒ no changeset needed. CI green (Unit, Integration, Lint&Type, CodeRabbit).

## should-fix (fixed in-code — commit 242549c2)
Codex's second-opinion lane caught a real coverage gap my first mutation missed: `secp256k1.verify()` accepts **both** DER and 64-byte compact `r||s`, so the ECDSA `verify()` alone would stay green even if `formatSignature()` regressed from DER passthrough to raw `r||s` (a well-formed `0x`+compact mutation passed 3/3). The docstring claims "if the DER / raw-r||s adapter logic regresses, this test fails" — which wasn't true for that ECDSA case. I hardened it: parse the formatted signature strictly as DER (`secp256k1.Signature.fromDER`, throws on compact) and assert it decodes to the same r/s. Re-ran: clean 3/3, and the previously-green compact mutation now goes RED. EdDSA half was already robust.

## risk
Low. Test-only, no production surface. Strengthens CI's ability to catch signing-adapter regressions without a real vault. The e2e vault-gated suite still requires a provisioned fixture (unchanged) — this PR just makes its absence loud and adds always-on synthetic coverage underneath it.

## verdict
Approve. Non-inert, always-on, skip-warning fires, correct crypto, no smuggled change, CI green + the new test runs and passes. The one coverage overstatement (ECDSA) is fixed in 242549c2.

<sub>🤖 2-lane review (Claude Opus 4.8 + Codex gpt-5.4). Reviewed at HEAD 242549c2.</sub>

PR #1115, opened 2026-07-09
Harden scientific amount formatting

rcoderdev:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of very large and very small numeric amounts by expanding scientific notation before conversion.
  - Preserved correct truncation behavior for fractional token amounts.
  - Ensured fiat-to-token conversions return complete decimal values without scientific notation.

- **Tests**
  - Added coverage for scientific-notation amounts, large price results, and truncation consistency.


gomesalexandre (approved):
## review: harden scientific amount formatting

two commits; clean, well-scoped.

---

### what it does

the old numeric path in `toChainAmount` called `amount.toFixed(decimals)` to expand scientific notation - but `toFixed` itself emits scientific notation for values ≥ 1e21 (e.g. `(1.5e21).toFixed(0)` returns `"1.5e+21"`). that string then hits viem `parseUnits` and either throws or produces a wildly wrong base-unit value.

the fix routes numbers through the existing `expandScientificNotationToDecimalString` (already battle-tested on the string path) which uses bigint arithmetic to avoid any float corruption.

the `fiatToAmount` refactor is also correct - the removed local `formatDecimalString` had the exact same `toFixed` large-number bug; delegating to `fiatToCrypto` / `expandToPlainDecimal` in `convertAmount.ts` fixes both paths in one shot.

---

### correctness walkthrough

- `(1e21).toString()` = `"1e+21"` - contains `e`, so hits `expandScientificNotationToDecimalString` - returns `"1000000000000000000000"` - `parseUnits("1000000000000000000000", 18)` = correct bigint. verified.
- `(1.99e-8).toString()` = `"1.99e-8"` - expands to `"0.0000000199"` - `parseUnits("0.0000000199", 8)` rounds up to `2n` (viem rounds, not truncates). test pin confirmed.
- rounding parity test `toChainAmount(1.55, 1) === toChainAmount('1.55', 1)` = both `16n`. correct.
- `AmountConvertError` → `FiatToAmountError` re-wrap in `fiatToAmount`: the only new `AmountConvertError` surface is `assertDecimals` (invalid decimals arg, previously silent garbage). that's a behavior improvement, not a regression.

`yarn vitest run` on the two touched test files: **28/28 passed**.

---

### minor notes

`suggestion:` `toChainAmount(NaN, 18)` and `toChainAmount(Infinity, 18)` don't match the `/[eE]/` regex so they fall through to `parseUnits("NaN"/"Infinity", 18)` which throws with a viem error. same behavior as before this PR and arguably acceptable (callers should validate), but a pre-filter with `Number.isFinite()` at the top of the `number` branch would give a cleaner message. non-blocking, pre-existing gap.

changeset: present, `@vultisig/core-chain` + `@vultisig/sdk` patch. correct.

---

LGTM ser - bug was real, fix is clean, tests pin the behavior.

🤖 reviewed via Claude Code (local /loop poll)

gomesalexandre (approved):
took over the stale merge conflict and pushed `merge main: keep truncation floor + scientific-notation fix in toChainAmount` on this head.

what changed
- merged fresh `origin/main` into the branch
- preserved main's truncation-floor behavior in `toChainAmount`
- preserved this PR's large scientific-notation fix for numeric amounts
- kept the `fiatToAmount` refactor/tests intact
- fixed the now-stale round-half-up assertions in the overlapping amount tests so they match the truncation contract on main

receipts
- `packages/core/chain/amount/*` -> 23/23 passed
- `packages/sdk` fiatToAmount + convertAmount -> 45/45 passed

once CI finishes on this head this is good to land.

PR #1116, opened 2026-07-09
Read XRP issued currency balances

### LINKED ISSUE #997: feat(ripple): read XRP token balances via account_lines + reflect trust lines in owner reserve
## What
Discover and read XRP **issued-currency balances** (trust-line tokens) so they appear in the portfolio, and fold trust-line count into the owner-reserve calc.

## Fix
Query `account_lines` for the account's trust lines (balance, limit, currency, issuer, auth status); surface tokens in the asset model. Include the trust-line count in `OwnerCount`-based owner reserve so spendable XRP stays correct (ties to reserve hardening #959/#5148/#4744).

## Acceptance
- [ ] account_lines read; token balances surfaced with issuer+currency
- [ ] Trust-line count reflected in owner reserve / spendable XRP
- [ ] Unit test on account_lines parsing

Part of the **XRP token support** feature set (all platforms, low-prio backlog). Currently XRP is Payment/XRP-only. Feasibility: WalletCore Ripple proto directly models OperationTrustSet + OperationPayment CurrencyAmount; signs via secp256k1 MPC — no blocker. · Ties to reserve work #959.

## References
- Fungible tokens: https://xrpl.org/docs/concepts/tokens/fungible-tokens
- Reserves: https://xrpl.org/docs/concepts/accounts/reserves
- account_lines: https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_lines

rcoderdev:
Closes #997
Runtime proof: live read-only SDK getXrpBalance('rMwNibdiFaEzsTaFCG1NnmAM3Rv3vHUy5L') returned balanceDrops=35223332, spendableDrops=31423332, ownerCount=14, minimumReserveDrops=3800000, reserveIncrementDrops=200000, trustLineCount=11, and positive issued-currency balance 524C555344000000000000000000000000000000.rMxCKbEDwqr76QuheSUMdEGf4B9xJ8m5De = 0.00204230364. CLI runtime proof on rebased commit 9b7e9717: branch-built clients/cli/dist/index.js balance Ripple --tokens --raw --output json returned success for Ripple with 0 XRP; branch-built agent ask against the real backend/test vault returned success, no tool calls, no transactions, and no XRP/trust-line data for that vault.

### DIFF
--- a/.changeset/xrp-account-lines-balances.md
+++ b/.changeset/xrp-account-lines-balances.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Surface XRP issued-currency trust-line balances and reserve-aware spendable XRP.
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -599,6 +599,7 @@ export type {
   VaultIdentity,
   Verdict,
   XrpBalance,
+  XrpIssuedCurrencyBalance,
   YieldActionResponse,
   YieldArgs,
   YieldBalance,
--- a/packages/sdk/src/tools/balance/index.ts
+++ b/packages/sdk/src/tools/balance/index.ts
@@ -7,7 +7,7 @@
  */
 
 // XRP
-export type { XrpBalance } from './otherBalance'
+export type { XrpBalance, XrpIssuedCurrencyBalance } from './otherBalance'
 export { getXrpBalance } from './otherBalance'
 
 // TRON
--- a/packages/sdk/src/tools/balance/otherBalance.ts
+++ b/packages/sdk/src/tools/balance/otherBalance.ts
@@ -7,6 +7,7 @@
  * Each chain is not wired through the EVM `getEvmClient` rail, so these talk to
  * public RPC / API endpoints (and the Vultisig proxy) directly via `fetchJson`.
  */
+import { rippleIssuedCurrencyDecimals, rippleTokenId } from '@vultisig/core-chain/chains/ripple/issuedCurrency'
 import bs58check from 'bs58check'
 
 import { fetchJson, formatBalance, ROOT_API_URL } from './rpc'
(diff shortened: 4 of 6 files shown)


gomesalexandre (changes requested):
## review: read XRP issued currency balances

good feature overall - reserve-aware spendable calc, pagination for large trust-line sets, clean type surface. one correctness bug before merge.

---

### blocker-level finding

**should-fix: `hasPositiveIssuedCurrencyBalance` silently drops balances returned in scientific notation**

XRPL `account_lines` serializes issued amounts using rippled's `"%.15g"` format, which uses scientific notation for values with magnitude ≥ 1e15 or < 1e-4. so a balance of `"1E+15"` (1 quadrillion USD, or any token with high supply) hits:

```ts
if (!/^\d+(\.\d+)?$/.test(unsigned)) return false
```

...and gets silently filtered out of `issuedCurrencyBalances`. the user's positive token holding is invisible to callers/agents. this is the core purpose of the new feature so the bug undercuts it.

fix is straightforward - `Number(unsigned) > 0` handles all valid XRPL string forms cleanly:

```ts
function hasPositiveIssuedCurrencyBalance(balance: string): boolean {
  const unsigned = balance.trim().replace(/^\+/, '')
  if (unsigned.startsWith('-')) return false
  const n = Number(unsigned)
  return Number.isFinite(n) && n > 0
}
```

and add a test pin:

```ts
it('handles scientific-notation balances from rippled %.15g formatting', () => {
  expect(hasPositiveIssuedCurrencyBalance('1E+15')).toBe(true)
  expect(hasPositiveIssuedCurrencyBalance('1e15')).toBe(true)
  expect(hasPositiveIssuedCurrencyBalance('1E-15')).toBe(true)
})
```

---

### other findings

**suggestion: parallel fetch for account_lines + server_state**

the two post-account_info fetches are independent but called sequentially:

```ts
const accountLines = await getXrpAccountLines(address)
const reserveInfo = await getXrpReserveInfo()
```

`Promise.all` would halve the p50 latency for every funded account lookup. easy win:

```ts
const [accountLines, reserveInfo] = await Promise.all([
  getXrpAccountLines(address),
  getXrpReserveInfo(),
])
```

**q: `trustLineCount` semantics**

`trustLineCount` is `accountLines.length` (ALL lines including negative / zero balance), while `issuedCurrencyBalances` is the positive subset. is the intent to expose "how many trust lines exist" (current) or "how many have balances" (what the field name implies to a caller)? both are defensible but a doc comment on `XrpIssuedCurrencyBalance` clarifying this would prevent confusion.

---

### praise

live `server_state` fetch for reserve values (instead of the hardcoded `rippleOwnerReserveDrops = 200000n` constant in `issuedCurrency.ts`) is the right call - XRPL reserve params can change via validator vote and a hardcoded value would silently lie after any such change.

pagination for account_lines is correct and tested. `spendableDrops = 0n` floor when balance < reserve is safe.

---

### tests + typecheck

`yarn workspace @vultisig/sdk vitest run tests/unit/tools/balance/otherBalance.test.ts`: **23/23 passed**. sdk typecheck: clean.

changeset: present, `@vultisig/sdk` patch. correct.

---

request-changes on the sci-notation filter bug - once that's fixed + pinned this should be good to go.

🤖 reviewed via Claude Code (local /loop poll)

PR #1126, opened 2026-07-09
fix(mpc): keep keygen tracing off stdout so -o json output stays parseable

neavra:
## Problem

The DKLS and Schnorr keygen / reshare / key-import ceremonies wrote progress lines — session ids, raw wire messages, `outbound message: { body: Uint8Array(…) }`, `keygen complete`, etc. — to **stdout** via ungated `console.log`.

stdout is the machine channel for the CLI's `-o json` mode (the documented `create fast … -o json` flow, which the CLI auto-selects when piped). The debug spew therefore:

- **corrupted the JSON envelope** — `JSON.parse(stdout)` failed because garbage preceded the JSON, and
- **leaked MPC internals** (session ids, raw message bytes) into terminals and CI logs.

## Fix

Route that tracing through a small gated logger, `mpcDebugLog`, that writes to **stderr** and only when `VULTISIG_DEBUG` is set (the existing CLI debug convention). stdout now carries only the final JSON envelope, while the debug output stays available to humans on demand.

- No keygen behavior changes — only the log **sink** moves off stdout.
- `mldsaKeygen` already logged exclusively to stderr, so it needed no change.
- Added a regression guard test asserting the keygen sources (`dkls`, `schnorr`, `mldsa`) carry no stdout `console.log`, plus unit coverage that `mpcDebugLog` writes to stderr (never stdout) and stays silent unless `VULTISIG_DEBUG` is set.

## Testing

- `yarn workspace @vultisig/cli typecheck` — pass
- `yarn lint` — pass
- `yarn test:cli` — 586 pass
- core MPC vitest — 332 pass (includes the new guard)
- Verified red-then-green: the guard matches 16 `console.log` calls in `dkls.ts` and 13 in `schnorrKeygen.ts` on `main`, 0 after this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Kept MPC keygen, reshare, and key-import debug messages out of standard output.
  - Improved `-o json` CLI output so it remains clean and parseable.

- **New Features**
  - Added optional MPC debug logging through `VULTISIG_DEBUG=1`, with diagnostics sent to standard error.

- **Tests**
  - Added coverage confirming debug logging behavior and protecting JSON output integrity.

### DIFF
--- a/.changeset/mpc-keygen-stdout-debug.md
+++ b/.changeset/mpc-keygen-stdout-debug.md
@@ -14,6 +14,6 @@ the documented `create fast … -o json` agent flow produced unparseable stdout
 into terminals and CI logs.
 
 Route that tracing through a gated logger that writes to stderr only when
-`VULTISIG_DEBUG` is set, so stdout carries only the final JSON envelope while
+`VULTISIG_DEBUG=1`, so stdout carries only the final JSON envelope while
 the debug output stays available to humans on demand. No keygen behavior
 changes — only the log sink moves off stdout.
--- a/.changeset/mpc-keygen-stdout-debug.md
+++ b/.changeset/mpc-keygen-stdout-debug.md
@@ -1,0 +1,19 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+fix(mpc): keep keygen tracing off stdout so `-o json` output stays parseable
+
+The DKLS and Schnorr keygen/reshare/key-import ceremonies logged progress
+(session ids, raw wire messages, "keygen complete", …) to stdout via ungated
+`console.log`. stdout is the machine channel for the CLI's `-o json` mode, so
+the documented `create fast … -o json` agent flow produced unparseable stdout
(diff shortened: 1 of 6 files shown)


gomesalexandre (approved):
approve. reviewed current head 3fa83320 only.

clean, tight fix. the keygen / reshare / key-import chatter now routes through `mpcDebugLog` to stderr only when `VULTISIG_DEBUG === '1'`, so cli `-o json` keeps stdout parseable. regression guard also covers the whole class of stdout sinks (`console.log/info/debug` + `process.stdout.write`) across dkls, schnorr, and mldsa keygen paths.

receipts:
- `yarn install --immutable`
- `yarn vitest run packages/core/mpc/mpcDebugLog.test.ts --config .config/vitest.core.config.ts` -> 8/8 pass
- `yarn eslint -c .config/eslint.config.mjs packages/core/mpc/mpcDebugLog.ts packages/core/mpc/dkls/dkls.ts packages/core/mpc/schnorr/schnorrKeygen.ts` -> clean

non-blocking follow-up only: the sibling keysign flow still has stdout chatter in `packages/core/mpc/keysign/*`, but this PR's keygen-scoped fix is correct and mergeable.

PR #1127, opened 2026-07-09
fix(cli): honor VULTISIG_CONFIG_DIR for vault storage

neavra:
## Summary

`VULTISIG_CONFIG_DIR` is documented as the way to relocate the CLI's on-disk state, but it was a no-op for vault storage. The CLI constructed the SDK without a storage override, so vaults, the active-vault pointer, and the cache always resolved to `~/.vultisig` — only `config.json` and the agent journal honored the variable. The result was a split-brain config location that broke CI/container isolation and any "point at a scratch config" workflow.

This roots SDK vault storage at `getConfigDir()` through a small shared `createVaultStorage()` helper, wired into both the command entry point (`init`) and interactive mode. When the variable is unset, `getConfigDir()` falls back to `~/.vultisig`, so the default location is unchanged.

## Changes

- `createVaultStorage()` in `lib/config.ts` returns `new FileStorage({ basePath: getConfigDir() })`, imported from the `@vultisig/sdk/node` subpath.
- Both `new Vultisig(...)` call sites now pass `storage: createVaultStorage()`.
- Regression test pinning the resolution: env set → temp dir, unset → `~/.vultisig`.
- Vitest CLI config gains a `@vultisig/sdk/node` alias so the test runner resolves the node-platform subpath the helper imports.

## Testing

- `yarn workspace @vultisig/cli typecheck`, `yarn lint`, `yarn test:cli` (588 passing) — all green.
- Manual: `VULTISIG_CONFIG_DIR=<tmp> vultisig vaults` creates and reads only `<tmp>` (empty list, no leak from `~/.vultisig`); unset lists the home-dir vaults as before. Verified against both the `tsx` dev path and the built `dist/index.js`.

## Notes

Scoped to vault storage (the main gap). A few CLI-local paths still hardcode `~/.vultisig` and don't yet honor the override — shell-completion vault lookup, the update-check cache, and the agent vault-state store — left for a follow-up to keep this change focused.

## Summary by CodeRabbit

* **Bug Fixes**
  * Vault data, active-vault information, and cache files now use the directory set by `VULTISIG_CONFIG_DIR`.
  * Empty or whitespace-only configuration values safely fall back to `~/.vultisig`.
  * CLI and interactive modes now consistently use the same vault storage location.

### DIFF
--- a/.changeset/cli-config-dir-vault-storage.md
+++ b/.changeset/cli-config-dir-vault-storage.md
@@ -1,0 +1,14 @@
+---
+'@vultisig/cli': patch
+---
+
+fix(cli): honor VULTISIG_CONFIG_DIR for vault storage
+
+The documented `VULTISIG_CONFIG_DIR` env var was a no-op for vault storage: the
+CLI constructed the SDK without a storage override, so vaults, the active-vault
+pointer and cache always resolved to `~/.vultisig` even when the var pointed
+elsewhere — only `config.json` and the agent journal honored it, producing a
+split-brain config location that broke CI/container isolation and multi-tenant
+use. The CLI now roots SDK vault storage at `getConfigDir()` via a shared
+`createVaultStorage()` helper. When the var is unset it falls back to
+`~/.vultisig`, so the default location is unchanged.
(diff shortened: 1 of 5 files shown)


gomesalexandre (approved):
approve. verified current head ce901013.

looked clean on the changed files only. `createVaultStorage()` now roots sdk `FileStorage` at `getConfigDir()`, and both normal cli init + interactive mode pass that storage into `new Vultisig(...)`, so vaults / active-vault / cache finally honor `VULTISIG_CONFIG_DIR` instead of leaking to `~/.vultisig`.

receipts:
- `yarn test:cli clients/cli/src/lib/__tests__/config.test.ts` -> 3/3 pass in a fresh temp checkout
- coverage pins override, default fallback, and empty/whitespace env handling

no blocker found ser.

rcoderdev (approved):

PR #1128, opened 2026-07-10
feat(ripple): add first-class XRP DestinationTag support

### LINKED ISSUE #980: fix(ripple): enforce destination RequireDestTag on sends (block tagless send to tag-required accounts)
## What
XRP sends do not check the destination account's **RequireDestTag** flag (`lsfRequireDestTag`). Sending tagless XRP to an address (e.g. an exchange) that requires a destination tag can result in **permanently lost funds**.

## Fix
On the XRP send path, fetch the destination account flags via `account_info` and detect `lsfRequireDestTag`. If set and no destination tag is provided, **block the send and surface a clear warning** (return a typed error the apps can present). Current send builder overloads tags onto memo and never reads dest flags (`signingInputs/resolvers/ripple.ts`).

## Acceptance
- [ ] Dest `RequireDestTag` detected via account_info on the ripple send path
- [ ] Tagless send to a RequireDest account is blocked with a typed, user-presentable error
- [ ] Unit test covers required-tag-present and required-tag-missing
- [ ] Consumed by iOS/Android/Windows (linked issues)

Ref: XRPL source/dest tags https://xrpl.org/docs/concepts/transactions/source-and-destination-tags · Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

### LINKED ISSUE #981: feat(ripple): decode X-addresses to classic address + destination tag
## What
X-addresses (`X...` mainnet / `T...` testnet) pack a classic address **and a destination tag** into one string. Vultisig validates classic r-addresses only (`chain/utils/isValidAddress.ts`, `AnyAddress.isValid(xrp)`), so a user pasting an X-address either fails validation or (worse) has the **embedded tag silently dropped** → funds delivered without the tag → lost.

## Fix
Accept X-address input, decode to (classic address + destination tag), and use the embedded tag. WalletCore provides X-address support (`XRPXAddress`) — no exotic serialization needed.

## Acceptance
- [ ] X-addresses validate as valid XRP destinations
- [ ] Decoding yields classic address + embedded destination tag; the tag is applied to the Payment
- [ ] Round-trip unit test (classic+tag ↔ X-address)
- [ ] Consumed by iOS/Android/Windows (linked issues)

Ref: XRPL addresses/X-addresses https://xrpl.org/docs/concepts/accounts/addresses · Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

rcoderdev:
## Summary

Adds first-class XRP `DestinationTag` support across SDK send preparation, CLI, React Native, X-address handling, and Ripple keysign/signing flows.

## Compatibility behavior

The implementation follows the existing cross-platform transition behavior:

- A legacy canonical numeric memo is interpreted as the DestinationTag.
- A DestinationTag-only send carries the canonical decimal tag in both the typed field and legacy memo carrier.
- If the memo equals the DestinationTag, the signed transaction contains the tag only.
- If memo and DestinationTag are both present and different, both are preserved.
- A nonnumeric memo remains an independent XRPL memo.
- DestinationTag `0` is supported as a present `uint32` value, including X-addresses and RequireDestTag destinations.

## Verification

- Core Ripple/X-address/keysign and payload regression tests pass.
- CLI transaction tests pass.
- SDK send-preparation and React Native Ripple golden-vector tests pass.
- `yarn check:ci` and the required downstream consumer checks are rerun on this revision before merge.

## Related work

- CommonData field `RippleSpecific.destination_tag` is already available on CommonData `main`; no additional CommonData change is required.
- Android and iOS consumers can adopt the same field-plus-legacy-memo compatibility behavior.
- Windows follows the published SDK release.
- Closes #980 and closes #981.

## Summary by CodeRabbit

* **New Features**
  * Added first-class XRP DestinationTag support across SDK, CLI, and transaction flows.
  * Added support for XRP X-address decoding, normalization, validation, and encoding.
  * Added CLI options and transaction previews for specifying and displaying destination tags.
  * Preserved XRP memo compatibility while allowing destination tags and memos independently.
  * Added validation for required, invalid, conflicting, and out-of-range destination tags.
  * Updated fee estimation and maximum-send calculations to account for destination tags.
* **Tests**
  * Expanded coverage for X-addresses, destination tags, memo compatibility, and required-tag scenarios.


gomesalexandre (changes requested):
preferably-blocking:

`RippleSpecific.destinationTag` is documented as a dual-write transition field, but the write path is field-only right now. `buildSendKeysignPayload()` still builds `KeysignPayload.memo` from the caller memo only, and `getRippleChainSpecific()` writes the tag only into `RippleSpecific.destinationTag`.

That means a mixed-version SecureVault pair can diverge on the new `--destination-tag` flow: a legacy signer drops field 4, sees no memo carrier, and signs a Payment with no DestinationTag while the upgraded signer includes one. Safe failure, but still a real interop break for the new path, and it contradicts the generated field comment about mixed-version byte-identical keysign.

So one of these needs to happen before merge:
- actually dual-write the canonical decimal tag into `KeysignPayload.memo` during the transition for cases that are meant to stay legacy-compatible, or
- narrow/fix the transition comment and explicitly document that the new field-only path is not mixed-version compatible.

Non-blocking nits I saw:
- CLI layer says `1..4294967295` but only regex-checks positive integers, so the upper bound is enforced downstream, not there.
- interactive `--memo` still swallows a later `--destination-tag` because it does `slice(i + 1).join(' ')` and breaks.

Targeted validation I want before stamp: prove the serialized payload/signing input behavior for a legacy peer on the new destination-tag flow, because that is the one contract the current tests do not pin down.

gomesalexandre (changes requested):
blockers on current head `3f05e5c0`.

1. `DestinationTag = 0` is currently rejected / treated as absent, but XRPL models `DestinationTag` as a `UInt32`, so `0` is valid.
   - `packages/core/mpc/keysign/utils/rippleDestinationTag.ts:15-19` rejects anything `< 1`
   - `packages/core/mpc/keysign/chainSpecific/resolvers/ripple.ts:55` treats `effectiveDestinationTag === 0` as "missing" for `lsfRequireDestTag`
   - the same `1..4294967295` restriction is mirrored in `clients/cli/src/index.ts:672-678`, `packages/sdk/src/platforms/react-native/chains/ripple/tx.ts:223-227`, and X-address helpers in `packages/core/chain/chains/ripple/address.ts:10,47-50`

   That means a real exchange/deposit flow using tag `0` is rejected, and a destination that merely requires the field can still false-fail if the user supplies `0`.

2. first-class destination tags still collide with numeric memos, so the promised "tag + memo" path is not actually supported for canonical decimal memos.
   - `packages/core/mpc/keysign/send/build.ts:85` preserves the caller memo when one is present
   - `packages/core/mpc/keysign/utils/rippleDestinationTag.ts:32-40` still parses any canonical numeric memo as a legacy tag even when `destinationTag` is already supplied
   - `packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts:52` then strips a memo that happens to equal the tag as an "echoed" carrier

   So a valid call like `destinationTag: 123, memo: "67890"` throws a conflict, and `destinationTag: 123, memo: "123"` silently loses the memo. That breaks the new API contract for users who intentionally want a numeric XRP memo alongside a real destination tag.

receipts:
- targeted SDK tests passed:
  - `yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/platforms/react-native/ripple-golden-vectors.test.ts packages/sdk/tests/unit/tools/prep/send.test.ts`
- targeted core/cli archive checks hit workspace-dist resolution noise, but 31 XRP/core assertions still passed before the import-resolution failure

not mergeable as-is. both are correctness/API blockers for first-class destination-tag support.

gomesalexandre (changes requested):
blocker on current head `ff7a896`.

`destinationTag` + distinct `memo` is still not mixed-version-safe.

- `packages/core/mpc/keysign/send/build.ts:82-86` keeps a distinct caller memo in `keysignPayload.memo` and only dual-writes the tag into `memo` for tag-only / equal-memo cases.
- `packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts:47-60` then resolves the typed field first, but if an older signer drops `RippleSpecific.destinationTag` field 4 it falls back to parsing any canonical numeric memo as the legacy carrier.

That means a payload like `destinationTag: 12345, memo: "67890"` diverges across peers:
- modern peer: raw JSON Payment with `DestinationTag: 12345` and memo `67890`
- legacy peer with field 4 dropped: `opPayment.destinationTag = 67890` and no memo at all

So MPC peers derive different XRP signing payloads on the same request. The current tests cover the equal memo/tag compatibility case, but not this distinct numeric memo mixed-version case.

This needs to fail closed or be version-gated before merge, unless you add a compatibility encoding that keeps legacy and modern peers byte-identical for `destinationTag + distinct memo`.

receipt:
- verified against fetched `origin/main` vs current head `ff7a896`
- reproduced the resolver behavior directly from the landed logic: modern => `{DestinationTag:12345, Memo:"67890"}`, simulated legacy-field-drop => `{DestinationTag:67890, Memo:null}`

gomesalexandre (approved):
approve-pass on current head `278b186`.

receipts
- `yarn build:all` ok in fresh `/tmp` clone
- `yarn cli:build` ok
- `yarn cli:smoke` ok
- targeted XRP coverage ok:
  - `packages/core/mpc/keysign/send/build.test.ts` 7/7
  - `packages/sdk/tests/unit/platforms/react-native/ripple-golden-vectors.test.ts` 9/9

re-checked the earlier blockers on this head:
- tag `0` is preserved and covered
- distinct numeric `memo` + `destinationTag` now stay independent in the signing-input resolver instead of colliding
- CLI preview path now carries destination tag through dry-run output and X-address conflict handling

one env wart i hit is `packages/sdk/tests/unit/tools/prep/send.test.ts` failing to resolve `@/tools/prep/send`, but it reproduces the same way on `origin/main`, so not a regression from this PR.

gomesalexandre (approved):
current-head re-approve after the merge-from-main.

why this is still green
- the only delta since my earlier approval at `278b186d` is the merge commit `eb6e0c35`
- that merge pulled in the raw-json ripple hardening from #1190 and the conflict resolution is correct in both ripple resolvers
- the destination-tag compatibility matrix from this PR still holds: tag-only dual-write, equal memo collapse, distinct memo preservation, legacy-peer fallback, and tag `0` presence via the optional uint32 field

grounded receipts from the current head
- CI is green on `eb6e0c35`: unit, integration, lint/typecheck, CodeRabbit
- remote diff re-check of the two overlap files shows both features coexist correctly:
  - `packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts`
  - `packages/core/mpc/keysign/chainSpecific/resolvers/ripple.ts`
- the RequireDestTag guard was correctly adapted to the optional `toAddress` path from the dapp raw-json work, so missing `toAddress` on that path does not false-trip the destination inspection

remote-only note
- this pass was remote-only via `gh` because the local sdk clone is busy and disk is tight, so I am relying on the green current-head CI rather than rerunning local receipts.

lgtm.

PR #1129, opened 2026-07-10
fix(cli): report per-chain failures in portfolio instead of swallowing them

neavra:
## What

The `portfolio` command silently swallowed per-chain failures, so a machine consumer reading `-o json` could not distinguish "chain has no balance / no value" from "chain errored", and exit 0 masked partial failure entirely.

Two root causes in `clients/cli/src/commands/balance.ts`:
- `vault.balance(chain)` was called **outside** the per-chain `try`, so a single throwing chain rejected the whole `Promise.all` and failed the entire command.
- `vault.getValue(...)` failures were swallowed by a bare `catch {}`, silently dropping the chain's fiat value with no marker.

## Change

Each chain is now fetched independently and honestly:

- `balance()` fails → chain omitted from `chainBalances`, recorded as `{ chain, stage: 'balance', error }`.
- `getValue()` fails → balance kept (without a value), recorded as `{ chain, stage: 'value', error }`.

The `-o json` envelope always carries a `failures` array (empty on full success), so consumers can rely on a stable schema:

```jsonc
{
  "success": true, "v": 1,
  "data": {
    "portfolio": { "totalValue": { ... }, "chainBalances": [ /* chains that loaded */ ] },
    "currency": "usd",
    "failures": [ { "chain": "Bitcoin", "stage": "balance", "error": "ECONNREFUSED btc-rpc" } ]
  }
}
```

`error` is a concise single-line message — never a stack trace or filesystem path (item 4 of the task). Successful chains still render. The human-readable table output prints failures as `Warning:` lines below the table.

## Exit contract (the delegated decision)

- **Some chains failed** → exit **0**, envelope carries a non-empty `failures` array. Rationale: partial results are real, usable data; forcing a non-zero exit would make agents discard a mostly-successful portfolio. Machine consumers branch on `data.failures.length`, not `$?`. This partial-success contract is documented in the CLI README exit-codes section.
- **Every chain failed to fetch a balance** → throw `NetworkError` → exit **3** (network, retryable). Per-chain balance fetches fail overwhelmingly for RPC/network reasons, and retryable is the honest signal.

`getTotalValue` is already internally resilient (it try/catches per chain and always returns a value), so it was left untouched. The single-chain `balance <chain>` path does not call `getValue` and already propagates errors cleanly through `withExit`, so no swallow exists there.

## Tests

`clients/cli/src/commands/balance.test.ts` — red-then-green (all 5 fail on `origin/main` behavior, pass after the fix):
- one unreachable chain → other chains still returned + the bad chain in `failures` (proves the whole-command-failure regression).
- balance succeeds but `getValue` throws → chain kept without a value + `stage: 'value'` failure (proves the swallowed-value regression).
- full success → empty `failures` array (stable schema).
- all chains fail → throws `NetworkError` with exit code 3.
- failure message is a single line with no stack frames; stdout is a single parseable envelope.

## Validation

- `yarn workspace @vultisig/cli typecheck` ✅
- lint (eslint) on changed files ✅
- `yarn test:cli` — 598 passed (41 files) ✅
- prettier `--check` on changed files ✅
- `@vultisig/cli` patch changeset included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Portfolio results now continue when individual chains fail.
  * Fiat valuation failures are reported instead of silently omitted.
  * Commands succeed with exit code 0 when at least one chain is available; fully failed requests return a network error.

* **Improvements**
  * JSON output now includes structured per-chain failure details.
  * Human-readable output displays warnings identifying failed chains and stages.
  * Error messages are concise and free of stack traces.


gomesalexandre (approved):
does what it says on the box — per-chain isolation vs the old Promise.all hard-fail. `conciseError` helper is clean, all-chains-fail branch throws correctly (non-zero exit), JSON consumers get `failures` array always-present. 614/614 tests green locally. 🤖 reviewed via Claude Code (/loop sweep)

PR #1141, opened 2026-07-10
fix(polkadot): hash-verify temporarily-banned broadcasts instead of assuming success

### LINKED ISSUE #1136: polkadot: broadcast resolver may swallow a genuine rejection as success
`resolvers/polkadot.ts:29-33,70-72` swallows a `/temporarily banned/i` match without routing through `verifyBroadcastByHash`, unlike `bittensor.ts`'s narrower "Already Imported"-only match. A genuinely pool-banned (not just peer-race-duplicate) extrinsic could report broadcast success.

Full agent-ready prompt: [round-4 bug-hunt artifact](https://claude.ai/code/artifact/4ac4b9f5-2a5d-4b4e-ab54-b5f693a2c973), sdk finding "Polkadot broadcast resolver."

Severity: MED. Effort: S. Confidence MED — needs on-chain verification of substrate ban semantics before/during the fix.

gomesalexandre:
Closes #1136

## what

The Polkadot broadcast resolver treated a Substrate `temporarily banned` pool error as an idempotent success and `return`ed without ever consulting `verifyBroadcastByHash`. That is a fund-safety false-positive risk: Substrate's pool bans a tx hash for a cool-off window whenever the hash is *removed* from the pool, and removal happens on BOTH a benign already-processed duplicate AND a genuine rejection (an invalid/dropped extrinsic banned to stop retry spam). The error string alone can't disambiguate, so assuming success could report a genuinely-rejected tx as confirmed.

By contrast the sibling `bittensor.ts` only fast-paths the unambiguous `"Already Imported"`.

## how

- Removed `/temporarily banned/i` from `idempotentBroadcastErrorPatterns`. It no longer short-circuits to success; the existing control flow then throws → the outer `catch` routes it through `verifyBroadcastByHash`, which swallows the error **only** when the tx hash is actually observed on chain / in the pool (`status === 'success'` or known-pending) and otherwise re-throws the real failure. Fail-safe by construction.
- `already imported` / `already known` (unambiguous peer-race duplicates) keep their fast-path.
- Rewrote the doc comment to explain why `temporarily banned` is deliberately excluded.

This intentionally errs toward surfacing failure over swallowing: for a signing/broadcast path, a false "it failed" (recoverable, user retries) is far safer than a false "it succeeded" (user believes funds moved when they did not).

## testing

`polkadot.test.ts` — updated the stale "swallows temporarily banned" assertion and added coverage for all three sub-cases:
- routes `temporarily banned` through `verifyBroadcastByHash` (no longer swallowed outright)
- surfaces it as a **failure** when the hash is NOT observed on chain (verify re-throws)
- still **swallows** it when `verifyBroadcastByHash` confirms the tx did land (benign duplicate sub-case)

Gates: `polkadot.test.ts` **13/13 passed**, `@vultisig/sdk typecheck` ✅, eslint ✅ (0 errors), prettier ✅ (unchanged).

## changeset

## risk

Low, tightening-only. No existing guard weakened; the unambiguous duplicate fast-paths are untouched. Worst case is a slow MPC peer on a genuinely-confirmed tx briefly surfacing a failure if Polkadot tx-status polling is unavailable — recoverable and strictly safer than the prior false-success.

## known follow-up (out of scope here)

Review surfaced that for Polkadot the `verifyBroadcastByHash` confirm branch is **inert in prod today**: `getPolkadotTxStatus` hits Subscan's gated `assethub-polkadot.api.subscan.io` with no `X-API-Key`, so it 403s and always returns `isKnown: false` → the original error is always re-thrown. So the shipped effect of this PR is that `temporarily banned` **fails closed** (surfaced as failure), never confirm-and-swallowed. That is the fund-safe direction and matches the tests, but the "confirm a benign peer-race duplicate" path only becomes live once Subscan tx-status is authenticated.

Fixed in this PR (doc-honesty, no logic change): the resolver comment + changeset now state this fail-closed reality instead of implying a working confirm path, and the previously-deleted Subscan-limitation warning is restored inline.

Restoring true confirm-and-swallow is tracked as a follow-up: **vultisig/vultisig-sdk#1145** (route Polkadot Subscan tx-status through an authenticated path). Intentionally not fixed here.

🤖 Agent-generated

### DIFF
--- a/.changeset/polkadot-temporarily-banned-hash-verify.md
+++ b/.changeset/polkadot-temporarily-banned-hash-verify.md
@@ -1,0 +1,21 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(polkadot): hash-verify "temporarily banned" broadcasts instead of assuming success
+
+The Polkadot broadcast resolver treated a Substrate `temporarily banned` pool
+error as an idempotent success (peer-race duplicate) and returned without ever
+consulting `verifyBroadcastByHash`. Substrate bans a tx hash for a cool-off
+window whenever it is *removed* from the pool, which covers both a benign
+already-processed duplicate AND a genuine rejection (an invalid/dropped
+extrinsic banned to stop retry spam) — the string alone cannot tell them apart.
+Swallowing it therefore risked reporting a genuinely-rejected transaction as
+confirmed (a fund-safety false positive), unlike `bittensor.ts` which only
+fast-paths the unambiguous "Already Imported".
+
+`temporarily banned` is removed from the idempotent fast-path list; it now flows
+through `verifyBroadcastByHash`, which swallows the error only when the tx hash
+is actually observed on chain / in the pool and otherwise surfaces the real
+failure. `already imported` / `already known` (unambiguous duplicates) keep their
+fast-path.
--- a/.changeset/polkadot-temporarily-banned-hash-verify.md
+++ b/.changeset/polkadot-temporarily-banned-hash-verify.md
@@ -15,7 +15,15 @@ confirmed (a fund-safety false positive), unlike `bittensor.ts` which only
 fast-paths the unambiguous "Already Imported".
 
 `temporarily banned` is removed from the idempotent fast-path list; it now flows
-through `verifyBroadcastByHash`, which swallows the error only when the tx hash
-is actually observed on chain / in the pool and otherwise surfaces the real
-failure. `already imported` / `already known` (unambiguous duplicates) keep their
-fast-path.
+through `verifyBroadcastByHash`. `already imported` / `already known`
+(unambiguous duplicates) keep their fast-path.
+
+Note on current prod behaviour: for Polkadot the confirm branch of
+`verifyBroadcastByHash` is effectively inert today — its status lookup hits
+Subscan's `assethub-polkadot.api.subscan.io` endpoint, which requires an API key
+this codebase does not send, so the request 403s and the original error is
+always re-thrown. The net shipped effect of this change is therefore to make
+`temporarily banned` **fail closed** (surfaced as a failure) rather than be
+assumed a success — the fund-safe direction. Restoring true
+confirm-and-swallow of benign peer-race duplicates requires authenticated
+Subscan tx-status and is tracked as a separate follow-up (vultisig-sdk#1145).
--- a/packages/core/chain/tx/broadcast/resolvers/polkadot.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/polkadot.ts
@@ -12,25 +12,29 @@ type RpcResponse = {
 }
 
 /**
- * Substrate Pool errors that mean "this extrinsic is already in the pool /
- * imported / banned because a peer just submitted it." In an MPC ceremony,
- * the initiator and the joiner both broadcast the same signed extrinsic; the
+ * Substrate Pool errors that UNAMBIGUOUSLY mean "this exact extrinsic is
+ * already in the pool because a peer just submitted it." In an MPC ceremony
+ * the initiator and joiner both broadcast the same signed extrinsic; the
  * slower device hits exactly these messages while the tx is in fact already
- * on chain. Treat them as idempotent successes — the alternative is showing
- * the slower device a "Signing Error" screen for a transaction that did
- * confirm. Fast path so we don't rely on `verifyBroadcastByHash`, which can't
- * help on Polkadot until tx-status polling moves off the gated Subscan API.
+ * accepted. Treat them as idempotent successes — the alternative is showing
+ * the slower device a "Signing Error" screen for a tx that did confirm.
+ *
+ * `temporarily banned` is deliberately NOT in this list. Substrate's pool
+ * bans a tx hash for a cool-off window whenever it was recently *removed* —
+ * which includes both a benign already-processed duplicate AND a genuine
+ * rejection (invalid/dropped extrinsic that got banned to stop retry spam).
+ * The string alone cannot tell those apart, so assuming success would report
+ * a genuinely-rejected tx as confirmed (fund-safety false positive). It is
+ * routed through `verifyBroadcastByHash` instead, which only swallows the
+ * error if the tx hash is actually observed on chain / in the pool and
+ * otherwise surfaces the real failure.
  *
  * Substrate sources for the exact strings:
  *   substrate/client/transaction-pool/api/src/error.rs (AlreadyImported,
  *   TemporarilyBanned, NoTagsProvided, ImmediatelyDropped) and
  *   primitives/runtime/src/transaction_validity.rs (Stale).
  */
-const idempotentBroadcastErrorPatterns: readonly RegExp[] = [
-  /already imported/i,
-  /already known/i,
-  /temporarily banned/i,
-]
+const idempotentBroadcastErrorPatterns: readonly RegExp[] = [/already imported/i, /already known/i]
 
 const isIdempotentBroadcastError = (text: string): boolean =>
   idempotentBroadcastErrorPatterns.some(pattern => pattern.test(text))
--- a/packages/core/chain/tx/broadcast/resolvers/polkadot.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/polkadot.ts
@@ -27,7 +27,10 @@ type RpcResponse = {
  * a genuinely-rejected tx as confirmed (fund-safety false positive). It is
  * routed through `verifyBroadcastByHash` instead, which only swallows the
  * error if the tx hash is actually observed on chain / in the pool and
- * otherwise surfaces the real failure.
+ * otherwise surfaces the real failure. (See the prod-behaviour note at the
+ * `verifyBroadcastByHash` call below: for Polkadot that confirm path is
+ * currently inert because Subscan tx-status needs an API key we don't send, so
+ * today this ships as fail-closed — surfaced, never confirm-and-swallowed.)
  *
  * Substrate sources for the exact strings:
  *   substrate/client/transaction-pool/api/src/error.rs (AlreadyImported,
@@ -84,6 +87,19 @@ export const broadcastPolkadotTx: BroadcastTxResolver<OtherChain.Polkadot> = asy
       throw new Error('Polkadot broadcast failed: missing extrinsic hash in RPC response')
     }
   } catch (error) {
+    // NOTE ON CURRENT PROD BEHAVIOUR: for Polkadot the confirm branch of
+    // `verifyBroadcastByHash` is effectively inert. Its status lookup
+    // (`getPolkadotTxStatus`) hits Subscan's `assethub-polkadot.api.subscan.io`
+    // endpoint, which requires an `X-API-Key` this codebase does not send — the
+    // request 403s, `getPolkadotTxStatus` returns `{ isKnown: false }`, and
+    // `verifyBroadcastByHash` therefore always re-throws the original error. So
+    // in practice everything routed here (including `temporarily banned`) ships
+    // as FAIL-CLOSED: surfaced as a failure, never confirm-and-swallowed. That
+    // is the fund-safe direction (a genuinely-rejected tx is never reported as
+    // success), but it means the benign peer-race duplicate of a `temporarily
+    // banned` tx currently surfaces as a signing error too. Restoring the
+    // confirm-and-swallow path needs authenticated Subscan tx-status — tracked
+    // as a follow-up (vultisig-sdk#1145).
     await verifyBroadcastByHash({ chain, tx, error })
   }
 }
--- a/packages/core/chain/tx/broadcast/resolvers/polkadot.test.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/polkadot.test.ts
@@ -46,15 +46,6 @@ describe('broadcastPolkadotTx', () => {
       expect(mocks.verifyBroadcastByHash).not.toHaveBeenCalled()
     })
 
-    it('swallows "Transaction is temporarily banned"', async () => {
-      mocks.queryUrl.mockResolvedValue({
-        error: { code: 1010, message: 'Transaction is temporarily banned' },
-      })
-
-      await expect(broadcastPolkadotTx({ chain, tx })).resolves.toBeUndefined()
-      expect(mocks.verifyBroadcastByHash).not.toHaveBeenCalled()
-    })
-
     it('swallows generic "Already known" Pool error variants', async () => {
       mocks.queryUrl.mockResolvedValue({
         error: { code: 1013, message: 'Already known' },
@@ -68,6 +59,50 @@ describe('broadcastPolkadotTx', () => {
       mocks.queryUrl.mockResolvedValue({
         error: { code: 1013, message: 'TRANSACTION ALREADY IMPORTED' },
       })
+
+      await expect(broadcastPolkadotTx({ chain, tx })).resolves.toBeUndefined()
+    })
+  })
+
+  describe('"temporarily banned" is ambiguous — hash-verify, never assume success', () => {
+    // Substrate bans a tx hash after it is removed from the pool, which covers
+    // BOTH a benign already-processed duplicate AND a genuine rejection. The
+    // string cannot disambiguate, so it must be routed through
+    // verifyBroadcastByHash rather than swallowed outright (#1136).
+
+    it('routes "temporarily banned" through verifyBroadcastByHash instead of swallowing it', async () => {
+      mocks.queryUrl.mockResolvedValue({
+        error: { code: 1010, message: 'Transaction is temporarily banned' },
+      })
+      mocks.verifyBroadcastByHash.mockResolvedValue(undefined)
+
+      await broadcastPolkadotTx({ chain, tx })
+
+      expect(mocks.verifyBroadcastByHash).toHaveBeenCalledOnce()
+      const callArg = mocks.verifyBroadcastByHash.mock.calls[0]![0]
+      expect((callArg.error as Error).message).toContain('temporarily banned')
+    })
+
+    it('surfaces "temporarily banned" as a failure when the tx is NOT observed on chain', async () => {
+      // A genuinely-rejected-then-banned extrinsic: verifyBroadcastByHash
+      // cannot confirm the hash and re-throws — the caller must see failure,
+      // not a false success.
+      const bannedErr = new Error('Polkadot broadcast failed: Transaction is temporarily banned')
+      mocks.queryUrl.mockResolvedValue({
+        error: { code: 1010, message: 'Transaction is temporarily banned' },
+      })
+      mocks.verifyBroadcastByHash.mockRejectedValue(bannedErr)
+
+      await expect(broadcastPolkadotTx({ chain, tx })).rejects.toThrow(/temporarily banned/)
+    })
+
+    it('swallows "temporarily banned" when verifyBroadcastByHash confirms the tx on chain', async () => {
+      // The benign duplicate sub-case: the tx really did land, so the slow MPC
+      // peer must not see an error.
+      mocks.queryUrl.mockResolvedValue({
+        error: { code: 1010, message: 'Transaction is temporarily banned' },
+      })
+      mocks.verifyBroadcastByHash.mockResolvedValue(undefined)
 
       await expect(broadcastPolkadotTx({ chain, tx })).resolves.toBeUndefined()
     })


gomesalexandre (commented):
adversarial pass, everything re-verified locally. Verdict: **approve-direction, one preferably-blocking doc/follow-up item** before undrafting.

## what I verified

**1. Substrate semantics are real, not a guess.** `TemporarilyBanned` comes from the pool's `PoolRotator`, which bans a tx hash whenever it is *removed* from the pool - and `validated_pool.rs` bans on BOTH `prune_tags` (tx included in a block, benign duplicate case) and `remove_invalid` (genuine rejection, banned to stop retry spam). So the string is genuinely ambiguous and main's unconditional swallow was a real false-success risk. The premise of #1136 and this fix is grounded.

**2. `verifyBroadcastByHash` does support Polkadot.** [`getTxHash`](https://github.com/vultisig/vultisig-sdk/blob/c0f9acbf18a97d514597deca14642ed4e730b185/packages/core/chain/tx/hash/resolvers/polkadot.ts) (blake2 of the signed bytes, registry-agnostic) and [`getPolkadotTxStatus`](https://github.com/vultisig/vultisig-sdk/blob/c0f9acbf18a97d514597deca14642ed4e730b185/packages/core/chain/tx/status/resolvers/polkadot.ts) both exist, and the status resolver has the correct fail-closed `isKnown: false` semantics. Chain alignment also checks out: live-probed `api.vultisig.com/dot/` with `system_chain` and it returns `"Polkadot Asset Hub"`, matching the `assethub-polkadot` Subscan status endpoint. No relay-vs-AH mismatch.

**3. preferably-blocking: the confirm branch is dead in prod - Subscan is hard-gated now.** Live probe of the status endpoint without a key:

```
{"code": 403, "message": "Subscan API strictly requires an API key. Unauthenticated access is disabled."}
```

`getPolkadotTxStatus` sends no key and there's no proxy, so in prod every `temporarily banned` routes to verify -> Subscan 403 -> `attempt` catches -> `isKnown: false` -> original error rethrown. Net shipping behavior is "always surface temporarily banned as failure", i.e. equivalent to just deleting the pattern. That is still the right fail-closed call (strictly safer than main's false success, and the common peer-race hits the `already imported` fast path while the tx is in-pool - the regression is the tail case where the slow device broadcasts after block inclusion). But the old comment explicitly warned verify "can't help on Polkadot until tx-status polling moves off the gated Subscan API", and this PR deletes that caveat while the changeset claims the on-chain-confirm swallow path works. Action: keep the fix as-is, but (1) restore an honest note in the resolver comment + changeset that the confirm branch is inert until Subscan gets a key/proxy, and (2) capture a follow-up to route Subscan through the api proxy so benign banned duplicates stop showing a signing error on the slower MPC device.

**4. Tests re-run: 13/13 pass** (`vitest run --config .config/vitest.core.config.ts broadcast/resolvers/polkadot`). Coverage has both directions: verify-confirms -> swallow, verify-rethrows -> failure surfaces, plus the routing assertion. Note they mock `verifyBroadcastByHash`, so it is routing-level coverage; the fail-closed `isKnown: false` leg is covered separately by `status/resolvers/polkadot.test.ts` (8/8, also re-run).

**5. `bittensor.ts` untouched** - diff vs origin/main is exactly the 3 expected files.

**6. Changeset scoped correctly** - `packages/core` has no standalone package.json (compiled into `@vultisig/sdk`), so `'@vultisig/sdk': patch` is right.

**7. Gates re-run myself**: root tsc + `@vultisig/sdk` typecheck clean, eslint on the touched broadcast paths clean.

gomesalexandre (commented):
## re-review of doc-honesty delta (3cb9b70ba)

Verdict: **APPROVE - ready for human review.** Delta is exactly what the prior review asked for, nothing more.

Verified:
- **Doc-only, no logic change**: delta since prior review = clean origin/main merge (81ba6999) + 3cb9b70ba touching only the changeset + comments in `packages/core/chain/tx/broadcast/resolvers/polkadot.ts`. The catch block's only executable line is still `await verifyBroadcastByHash({ chain, tx, error })` - untouched. `polkadot.test.ts` untouched.
- **Inline comment is accurate**: states the confirm branch is inert in prod (Subscan `assethub-polkadot.api.subscan.io` needs an `X-API-Key` the codebase doesn't send -> 403 -> `isKnown: false` -> always re-throws), so shipped behaviour is fail-closed, never confirm-and-swallow. Cross-checked the endpoint claim against `packages/core/chain/tx/status/resolvers/polkadot.ts:8` - matches.
- **Changeset corrected**: no longer claims a working confirm-and-swallow path; new 'Note on current prod behaviour' paragraph states the fail-closed reality and links the follow-up.
- **Follow-up filed + linked**: vultisig/vultisig-sdk#1145 (OPEN) and referenced from the PR body's 'known follow-up' section.
- **Gates re-run**: `@vultisig/sdk` typecheck exit 0; `polkadot.test.ts` 13/13 passed.

Ready for NeOMakinG's eyes.

packages/core/chain/tx/broadcast/resolvers/polkadot.ts
```diff
@@ -80,6 +87,19 @@ export const broadcastPolkadotTx: BroadcastTxResolver<OtherChain.Polkadot> = asy
       throw new Error('Polkadot broadcast failed: missing extrinsic hash in RPC response')
     }
   } catch (error) {
+    // NOTE ON CURRENT PROD BEHAVIOUR: for Polkadot the confirm branch of
+    // `verifyBroadcastByHash` is effectively inert. Its status lookup
+    // (`getPolkadotTxStatus`) hits Subscan's `assethub-polkadot.api.subscan.io`
+    // endpoint, which requires an `X-API-Key` this codebase does not send — the
+    // request 403s, `getPolkadotTxStatus` returns `{ isKnown: false }`, and
+    // `verifyBroadcastByHash` therefore always re-throws the original error. So
```
NeOMakinG, line 95: `suggestion:` the catch-block note asserts Subscan `always 403s` / confirm path is `always inert` — public Subscan sometimes answers unauthenticated (rate-limited), so state it as `typically` — The doc says the request `403s` and `verifyBroadcastByHash therefore always re-throws`. Subscan's public `/api/scan/extrinsic` has historically answered without a key under rate limits — so the confirm branch is not provably 100% inert, it's inert *when* Subscan rejects. This does not affect fund-safety (a real 200 only ever swallows on genuine on-chain success — still fail-safe), but the absolute `always` in a fund-critical comment invites a future reader to trust it as an invariant.

fix: soften `always` to `typically` / `in practice`, or verify the endpoint's unauthenticated behaviour and cite it. Behaviour is fine either way; this is doc-honesty only.

packages/core/chain/tx/broadcast/resolvers/polkadot.ts
```diff
@@ -12,25 +12,32 @@ type RpcResponse = {
...
  *   substrate/client/transaction-pool/api/src/error.rs (AlreadyImported,
  *   TemporarilyBanned, NoTagsProvided, ImmediatelyDropped) and
  *   primitives/runtime/src/transaction_validity.rs (Stale).
  */
-const idempotentBroadcastErrorPatterns: readonly RegExp[] = [
-  /already imported/i,
-  /already known/i,
-  /temporarily banned/i,
-]
+const idempotentBroadcastErrorPatterns: readonly RegExp[] = [/already imported/i, /already known/i]
```
NeOMakinG, line 40: `q:` confirmed peer-race duplicate that surfaces as `temporarily banned` now shows a signing error — is the UX regression acceptable until #1145? — In MPC keysign, when the fast peer's extrinsic has already been *included in a block* it is removed from the pool; a slow peer re-submitting inside the ban window gets `temporarily banned` (not `already imported`). With the confirm branch inert in prod, that genuinely-confirmed send now surfaces as a failure on the slow device. Bounded to that timing window (simultaneous broadcast still yields `already imported`, which stays fast-pathed), documented, and tracked in #1145 — and for a fund path a false-fail beats a false-success. Just confirming this tradeoff is the intended shipped behaviour and #1145 is the agreed follow-up, not a silent regression.

NeOMakinG (commented):
gm — multi-lane pass: claude correctness+fund-safety (load-bearing) + codex adversarial gpt-5.4 (infra-dead this pass — no output). conceptual-only (no runtime this pass). comment-only.

## suggestion
### 1. the catch-block note asserts Subscan `always 403s` / confirm path is `always inert` — public Subscan sometimes answers unauthenticated (rate-limited), so state it as `typically` — `packages/core/chain/tx/broadcast/resolvers/polkadot.ts:95`
The doc says the request `403s` and `verifyBroadcastByHash therefore always re-throws`. Subscan's public `/api/scan/extrinsic` has historically answered without a key under rate limits — so the confirm branch is not provably 100% inert, it's inert *when* Subscan rejects. This does not affect fund-safety (a real 200 only ever swallows on genuine on-chain success — still fail-safe), but the absolute `always` in a fund-critical comment invites a future reader to trust it as an invariant.
action: soften `always` to `typically` / `in practice`, or verify the endpoint's unauthenticated behaviour and cite it. Behaviour is fine either way; this is doc-honesty only.

## q
- confirmed peer-race duplicate that surfaces as `temporarily banned` now shows a signing error — is the UX regression acceptable until #1145?

## risk
Very low, tightening-only. A prior fund-safety false-positive (a genuinely-rejected `temporarily banned` extrinsic reported as a confirmed success) is closed. No existing guard weakened — the unambiguous `already imported`/`already known` duplicate fast-paths are untouched. Worst case is a bounded UX regression: a confirmed peer-race duplicate that happens to surface as `temporarily banned` now shows a signing error (confirm path is inert in prod), which is recoverable and strictly safer than a false success.

## verdict
COMMENTED

packages/core/chain/tx/broadcast/resolvers/polkadot.ts
NeOMakinG, line 40: `praise:` live-verified this is closing a real false success, not a theoretical one. Against the configured endpoint (`https://api.vultisig.com/dot/`, HTTP 200 on all of these): submitting a decodable-but-permanently-invalid extrinsic returns `1002 "Verification Error: … Execution **aborted** due to trap …"` on the first submit and `1012 "Transaction is temporarily banned"` on every submit after (deterministic 3/3). On `main` `/\babort(?:ed)?\b/i` classifies #1 transient, `withTransientBroadcastRetry` re-sends, #2 comes back `temporarily banned`, this list matches it and `broadcastTx` **resolves successfully** for a wasm-trapped extrinsic. Dropping the pattern here breaks that chain, and the banned message matches no transient pattern so the retry loop terminates immediately instead of burning all 3 attempts.

packages/core/chain/tx/broadcast/resolvers/polkadot.ts
```diff
@@ -80,6 +87,19 @@ export const broadcastPolkadotTx: BroadcastTxResolver<OtherChain.Polkadot> = asy
...
+    // `verifyBroadcastByHash` is effectively inert. Its status lookup
+    // (`getPolkadotTxStatus`) hits Subscan's `assethub-polkadot.api.subscan.io`
+    // endpoint, which requires an `X-API-Key` this codebase does not send — the
+    // request 403s, `getPolkadotTxStatus` returns `{ isKnown: false }`, and
+    // `verifyBroadcastByHash` therefore always re-throws the original error. So
+    // in practice everything routed here (including `temporarily banned`) ships
+    // as FAIL-CLOSED: surfaced as a failure, never confirm-and-swallowed. That
+    // is the fund-safe direction (a genuinely-rejected tx is never reported as
+    // success), but it means the benign peer-race duplicate of a `temporarily
+    // banned` tx currently surfaces as a signing error too. Restoring the
```
NeOMakinG, line 100: `should-fix:` two things converge at this throw site.

**(a) bare `Error` gets message-regex-classified.** Post-rebase, `broadcastTx` wraps polkadot in `withTransientBroadcastRetry` (`broadcast/index.ts:34` on `main`; `hasResolverOwnedRetry` exempts only evm/solana). The `1002` body above literally contains "aborted", so a *terminal* runtime rejection is re-broadcast 3×. `isTransientBroadcastError` short-circuits on `instanceof DeliverTxFailedError` before the regex runs — throwing that for substrate `1002`/`1010`/`1012` closes the false-success chain at step 1 as well as step 7, same treatment cosmos already gets.

**(b) the hash is lost past this throw.** `BroadcastService.ts:118` computes `getTxHash()` only *after* `coreBroadcastTx` resolves, `VaultBase.broadcastTx():1162` emits `error`, and the CLI maps `BroadcastFailed` to "Retry the transaction". So on a benign peer-race duplicate the losing device fails hard *and* has no hash to show — and a user retry is a fresh keysign that re-fetches the nonce, i.e. a genuinely distinct second extrinsic if the original landed. Computing the local hash before this throw would give the losing peer something to poll.

.changeset/polkadot-temporarily-banned-hash-verify.md
```diff
@@ -0,0 +1,29 @@
...
+fast-paths the unambiguous "Already Imported".
+
+`temporarily banned` is removed from the idempotent fast-path list; it now flows
+through `verifyBroadcastByHash`. `already imported` / `already known`
+(unambiguous duplicates) keep their fast-path.
+
+Note on current prod behaviour: for Polkadot the confirm branch of
+`verifyBroadcastByHash` is effectively inert today — its status lookup hits
+Subscan's `assethub-polkadot.api.subscan.io` endpoint, which requires an API key
+this codebase does not send, so the request 403s and the original error is
```
NeOMakinG, line 24: `q:` small accuracy fix — live, Subscan answers **HTTP 400** with `{"code": 403, "message": "Subscan API strictly requires an API key…"}` in the body, not an HTTP 403. The conclusion is unchanged (`assertFetchResponse` throws `HttpResponseError(400)` → `attempt` catches → `isKnown: false` → re-throw), and it would also hold on a 200-wrapped `{"code":403}` since `getPolkadotTxStatus` checks `response.code !== 0` — genuinely fail-closed on both shapes. Worth saying 400-with-a-403-body here and in the resolver comment so the next reader isn't hunting an HTTP status that never appears.

NeOMakinG (commented):
Reviewed as a **draft** — posting `COMMENT`, not approving. Reviewed worst-case-adversarial because this is the broadcast-honesty class in a fund-critical package (fourth sibling today alongside #1426 / #1430 / #1433).

**Bottom line: the premise is real and I live-reproduced the false success this PR closes.** The change is correct and the direction is right. Two things need resolving before it leaves draft: the branch is 135 commits behind `main` and misses the retry wrapper it now interacts with, and the fail-closed cost is larger than the PR body implies because no caller preserves the tx hash past the throw.

## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | suggestion | catch-block note asserts Subscan `always 403s` / confirm path `always inert` — public Subscan has historically answered unauthenticated under rate limits, so soften `always` to `typically` | ✅ **retracted — my r1 claim was wrong** | I live-probed it this round. Subscan is now hard-gated: `HTTP 400` + `{"code": 403, "message": "Subscan API strictly requires an API key. Unauthenticated access is disabled."}` on both `assethub-polkadot.` and `polkadot.api.subscan.io`. There is no rate-limited unauthenticated window any more, so the absolute `always` in the comment is factually correct as written. Only nit left is 400-vs-403 (see `q` #4 below). |
| r1 | 2 | q | is the peer-race `temporarily banned` UX regression acceptable until #1145? | ⚠️ **still open, and upgraded** | The author has not answered, and this round I traced the consumers and found the cost is larger than "shows a signing error": the tx hash is lost past the throw and the CLI tells the user to retry, which is the double-send path. Promoted to `should-fix` #2 below with the trace. |


## testing approach

Worktree `/tmp/wt-sdk-1141` @ `3cb9b70b`, node_modules symlinked from the canonical checkout.

**Live probe of the actually-configured endpoint** (`polkadotRpcUrl` = `https://api.vultisig.com/dot/`, 2026-07-20):

```
$ curl -s -X POST https://api.vultisig.com/dot/ -d '{"jsonrpc":"2.0","method":"system_chain","params":[],"id":1}'
{"jsonrpc":"2.0","id":1,"result":"Polkadot Asset Hub"}

# submit #1 of a decodable-but-permanently-invalid extrinsic (0x1001020304)
{"jsonrpc":"2.0","id":1,"error":{"code":1002,"message":"Verification Error: Runtime error:
 Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction
 executed\nWASM backtrace: ... asset_hub_polkadot_runtime.wasm ..."}}

# submits #2, #3, #4 — same bytes, deterministic 3/3
{"jsonrpc":"2.0","id":1,"error":{"code":1012,"message":"Transaction is temporarily banned"}}
{"jsonrpc":"2.0","id":1,"error":{"code":1012,"message":"Transaction is temporarily banned"}}
{"jsonrpc":"2.0","id":1,"error":{"code":1012,"message":"Transaction is temporarily banned"}}
```

Every one of these comes back **HTTP 200**. Also probed Subscan:

```
$ curl -X POST https://assethub-polkadot.api.subscan.io/api/scan/extrinsic -d '{"hash":"0x00..00"}'
HTTP 400
{"code": 403, "message": "Subscan API strictly requires an API key. Unauthenticated access is disabled."}
```

Also: mutation-verified the new tests (re-added `/temporarily banned/i` → **2 of 13 fail**, so the added coverage is load-bearing, not decorative); confirmed `polkadot.test.ts` exists and exercises the changed line directly (unlike #1433, where CI had zero coverage of the changed line); pure-function replay of the two live bodies through `isTransientBroadcastError` + both pattern lists, 4/4.

## the false success I reproduced on `main`

Chain the live bodies through `origin/main`'s wiring — `broadcastTx` wraps polkadot in `withTransientBroadcastRetry` (`broadcast/index.ts:34`, `hasResolverOwnedRetry` exempts only evm/solana):

1. submit #1 → `1002 "Verification Error: … Execution **aborted** due to trap …"`, HTTP 200
2. resolver throws a **bare `Error`** carrying that text
3. `verifyBroadcastByHash` → Subscan HTTP 400 → `attempt` catches → `{ isKnown: false }` → original error re-thrown
4. `isTransientBroadcastError` matches `/\babort(?:ed)?\b/i` → **`true`** (verified against the exact live string)
5. `withTransientBroadcastRetry` re-broadcasts
6. submit #2 → `1012 "Transaction is temporarily banned"` (deterministic, 3/3 live)
7. `main`'s `isIdempotentBroadcastError` matches → `return` → **`broadcastTx` resolves successfully** for an extrinsic the runtime rejected with a wasm trap

That is a real, automatic, single-user-action false success on `main` today. **This PR breaks the chain at step 7**, and because `Polkadot broadcast failed: Transaction is temporarily banned` matches no transient pattern, the retry loop terminates immediately rather than burning all 3 attempts. Good change.

Also worth recording since it was the brief's central question: **the SDK-level retry cannot produce a second distinct extrinsic.** `withTransientBroadcastRetry` re-invokes the resolver over the same `tx.encoded` bytes → identical signed extrinsic → identical hash → the pool dedupes it (that is literally why step 6 returns `1012`). No nonce reuse or increment. The distinct-extrinsic risk lives one layer up, at the *user* retry — see `should-fix` #2.

I also checked the resolver's polarity against the class rule (*throw unless explicitly success*) and it is **correct**: success requires a truthy `response.result`; `error` present, `result` empty, `result` missing, and a malformed/non-JSON-RPC body all throw. This is the inversion #1433's Tron guard got wrong, and it is right here.

## should-fix

**1 — the branch is 135 commits behind `main`, and the commits it is missing are exactly the ones it interacts with.**
`git log --oneline HEAD..origin/main | wc -l` → `135`. Missing: #1265 (`withTransientBroadcastRetry`, which now wraps this resolver) and #1432 (bittensor malformed-response). `broadcast/transientRetry.ts` does not exist on this branch at all, so **nothing in the 5 green checks tested the retry interaction** — the whole step 1→7 chain above is invisible to this CI run. `polkadot.ts` is byte-identical between the merge-base and `main`, so the rebase itself is clean; please rebase and re-run before un-drafting so CI covers the wiring the change actually ships into.

**2 — the fail-closed cost is bigger than "surfaces a signing error", and the recovery path is the double-send path.**
Traced the consumers: `BroadcastService.ts:118` computes `getTxHash()` **after** `coreBroadcastTx` resolves, so a throw here means the hash is never computed; `VaultBase.broadcastTx():1162` emits `error` instead of `transactionBroadcast`; `clients/cli/src/core/errors.ts:277` maps `BroadcastFailed` to *"Retry the transaction"*. So on a genuine peer-race duplicate the losing device gets a hard failure **and loses the hash it could have shown**, and the UI instructs the user to retry. A user retry is a fresh keysign that re-fetches the nonce — if the original did land, the nonce has incremented and the retry is a genuinely **distinct second extrinsic**. That is the fund-safety exposure this trade buys, and with the confirm branch admittedly inert there is no automatic recovery for it. Two options, either is fine: hold this behind #1145 (authenticated Subscan) so the confirm branch is live when the fast-path is removed, or compute the local `getTxHash()` before the throw so the losing peer can still surface/poll the hash.

**3 — throw `DeliverTxFailedError`, not a bare `Error`, for terminal substrate rejections.**
Step 4 above is a live-verified misclassification: `1002 "…Execution **aborted** due to trap…"` is a *terminal* runtime rejection, and it gets re-broadcast 3× because the chain-controlled text happens to contain "aborted". This is the exact hazard `transientRetry.ts` documents for cosmos, and the fix is already in the repo — `isTransientBroadcastError` short-circuits on `instanceof DeliverTxFailedError` before the regex ever runs. Since this PR is already editing this throw site, throwing `DeliverTxFailedError` for substrate `1002` / `1010` / `1012` closes the false-success chain belt-and-braces rather than only at step 7. (Applies post-rebase; `DeliverTxFailedError` is not on this branch yet.)

## q

**4 — the new comment says "the request 403s"; live it is HTTP 400 with `{"code": 403}` in the body.**
The conclusion holds either way — `assertFetchResponse` throws `HttpResponseError(400)`, `attempt` catches, `isKnown: false`, re-throw — and it also holds if Subscan ever switched to a 200-wrapped `{"code":403}` because `getPolkadotTxStatus` checks `response.code !== 0`. Fail-closed on both shapes, which is good. Worth correcting the two spots that state 403 so the next reader doesn't go looking for an HTTP status that isn't there.

**5 — good news for #1145: the endpoint is the right chain.** I went in expecting a relay-chain-vs-Asset-Hub mismatch (status queries `assethub-polkadot.api.subscan.io` while `packages/core/chain/chains/polkadot/client.ts` documents `polkadotRpcUrl` as *"the Polkadot RELAY CHAIN client (NOT Asset Hub)"*). `system_chain` on that very URL returns **`"Polkadot Asset Hub"`**, and the trap backtrace names `asset_hub_polkadot_runtime.wasm`. So the Subscan indexer matches what we broadcast to and #1145 really is just the API key — no endpoint change needed. The `client.ts` comment is the thing that's stale; separate PR.

**Bittensor — the sibling substrate resolver — still has this exact false success, in a worse form.** `status/resolvers/bittensor.ts:31` returns `{ status: 'pending' }` on *any* lookup miss or API error, with **no `isKnown` field**. `verifyBroadcastByHash.ts:35` computes `isKnownPending = result.status === 'pending' && result.isKnown !== false` — `undefined !== false` is `true` — so it swallows and returns. Net: **every** Bittensor broadcast rejection whose hash isn't in taostats is reported as success. Polkadot's status resolver got the `isKnown: false` treatment (`status/resolvers/polkadot.ts:31`, mirroring ripple/solana); bittensor did not. Out of scope here, but it is the same fund-safety class and deserves its own issue.

## risk

Low for what it changes; the residual risk is entirely in `should-fix` #2 — this trades a false success for a false failure whose only user-visible remedy is a retry that can double-send. Test coverage is genuine (mutation-verified 2/13). Changeset `patch` is the right level; no contract-shape change (the resolver still resolves `void`/throws, and the published `Promise<unknown>` deep-export signature is unaffected), no dependency ranges touched.

## verdict

`COMMENT` — draft, so no approval by policy, and I would not approve as-is regardless: rebase onto `main` and re-run CI so the `withTransientBroadcastRetry` interaction is actually exercised, and resolve the lost-hash / retry-double-send cost in #2. The core change is right and I'd approve it with those two settled.

---
🤖 reviewed by an agent · 2 lanes (claude + codex gpt-5.4) · live-probed `api.vultisig.com/dot` + Subscan · mutation-verified

PR #1142, opened 2026-07-10
fix(utxo): reject a below-dust primary send amount before signing

### LINKED ISSUE #1137: utxo: buildUtxoSendTx has no dust-floor check on the primary send amount
`chains/utxo/tx.ts:551,814` — `dustLimit` only gates the change output, never `opts.amount`. A dust-amount send burns a full MPC signing ceremony before failing at broadcast as a non-standard output.

Full agent-ready prompt: [round-4 bug-hunt artifact](https://claude.ai/code/artifact/4ac4b9f5-2a5d-4b4e-ab54-b5f693a2c973), sdk finding "UTXO send builder."

Severity: MED/LOW. Effort: S.

gomesalexandre:
Closes #1137

## what

`buildUtxoSendTx` applied the chain's `dustLimit` only to the **change** output (inside `serializeOutputs`); the **primary send amount** was never checked against the dust floor. A below-dust send builds an unrelayable / node-rejected output and burns a full MPC signing ceremony on a transaction that can never confirm.

## how

Added an early guard right after the existing `amount <= 0n` check, before any funding math or sighash work:

```ts
if (opts.amount < spec.dustLimit) {
  throw new Error(`amount ${opts.amount} is below the ${opts.chain} dust limit ${spec.dustLimit}`)
}
```

Purely additive — the existing zero/negative-amount guard and the change-output dust gating are untouched.

## testing

`tx.dustLimit.test.ts` — added a `buildUtxoSendTx` block:
- a `545n` BTC send (dust `546n`) throws `below the Bitcoin dust limit 546` **before** signing
- the existing zero-amount guard still fires
- an at-limit `546n` amount does NOT trip the dust guard (builds normally)

Gates: `tx.dustLimit.test.ts` **6/6 passed**, `@vultisig/sdk typecheck` ✅, eslint ✅ (0 errors), prettier ✅.

## changeset

## risk

Low, tightening-only. Only rejects sends that were already destined to fail at relay/confirm; no valid send path changes.

🤖 Agent-generated

## Summary by CodeRabbit

- **Bug Fixes**
  - Send transactions now reject amounts below the applicable network dust limit before transaction processing or signing.
  - Litecoin uses its 2,940-litoshi dust threshold, while Bitcoin retains its 546-satoshi threshold.
  - Zero and negative amount validation remains unchanged.
  - Amounts exactly at the dust limit continue to be accepted.

- **Tests**
  - Added coverage for chain-specific dust limits and validation behavior.


gomesalexandre (commented):
Reviewed + re-ran everything locally (fresh clone, PR head 3e64e11).

**Verdict: APPROVE** (comment-only since draft) - the guard is correct, early, and per-chain. One CI gap worth fixing before undrafting.

## verified

- **Guard is genuinely early**: sits right after the pre-existing `amount <= 0n` check at [tx.ts#L848-L855](https://github.com/vultisig/vultisig-sdk/blob/3e64e11fa7d2fe8315b87db1feddde8134a6ae0a/packages/sdk/src/chains/utxo/tx.ts#L848-L855), BEFORE address decode, fee math, and all sighash work. No ceremony burn possible.
- **Zero-amount guard untouched**: line 848 is byte-identical, and the test re-asserts it fires.
- **`dustLimit` source unchanged**: guard reads `spec.dustLimit` from the same `UTXO_SPECS` map ([tx.ts#L57-L67](https://github.com/vultisig/vultisig-sdk/blob/3e64e11fa7d2fe8315b87db1feddde8134a6ae0a/packages/sdk/src/chains/utxo/tx.ts#L57-L67)) that `serializeOutputs` gets for the change-output check at L931. No constants redefined.
- **Per-chain correctness (the big one)**: this is NOT a single hardcoded 546 applied everywhere - `spec.dustLimit` resolves per chain: BTC `546n`, LTC `2_940n` (UTXO-02 corrected value), DOGE `1_000_000n` (0.01 DOGE hard dust, correct), BCH/Dash/Zcash `1_000n`. All values are exact-or-conservative vs the real relay thresholds - none too LOW, which is the only dangerous direction for this class of bug. BTC 546 is technically over-conservative for P2WPKH recipients (Core's segwit dust is 294 sat), but that matches the existing change-output behavior and rejecting a 400-sat send is a feature, not a bug.
- **Boundary is exact**: `545n` throws, `546n` builds, and 546n really is `UTXO_SPECS.Bitcoin.dustLimit`. Semantics match Core (`dust` = value **<** threshold, so at-limit passes). Note the pre-existing change check uses `change > dustLimit` (at-limit change dropped to fee) - a 1-sat asymmetry vs the new `<` guard, but it errs conservative and predates this PR, meh.
- **Gates re-run**: tests 6/6 ✅ (`vitest run --config tests/unit/vitest.config.ts src/chains/utxo/tx.dustLimit.test.ts`), `@vultisig/sdk` typecheck ✅ (after `build:all` - the raw-clone failures are just unbuilt `mpc-wasm`/`mpc-types`), eslint 0 errors ✅, prettier (repo config) ✅, changeset present ✅.

## findings

**suggestion (fix before undrafting): `tx.dustLimit.test.ts` never runs in CI.** It's the ONLY src-colocated test file in `packages/sdk`, and both configs miss it: the default `vitest.config.ts` includes `tests/**` only, and the CI script `vitest run --config tests/unit/vitest.config.ts tests/unit` positionally filters to `tests/unit`. Verified: `vitest list --filesOnly --config tests/unit/vitest.config.ts tests/unit` collects 180 files, this one absent. Pre-existing gap from #1081 (the UTXO-02 Litecoin tests have the same problem), but this PR doubles down on the dead file - the whole point of these tests is regression protection and CI gives them zero. Action: move the file to `tests/unit/chains/utxo/tx.dustLimit.test.ts` (import path becomes `@/chains/utxo/tx` via the existing alias) - 1-minute fix, and it rescues the #1081 tests too.

**suggestion (nit, non-blocking):** the at-limit test asserts `not.toThrow(/dust limit/)` - a build that throws anything else would still pass. `expect(buildUtxoSendTx({...baseOpts, amount: 546n}).signingHashes.length).toBe(1)` (or similar) would pin "builds normally" for real.

**suggestion (nit, non-blocking):** boundary coverage is BTC-only; a one-liner LTC case (`2_939n` throws `dust limit 2940`) would prove the guard is genuinely per-chain rather than accidentally 546-shaped.

🤖 Agent-reviewed (re-ran all gates independently)

gomesalexandre (commented):
## re-review of 304c43c5 (test relocation delta)

Prior approval stands, delta verified end to end:

- `vitest list --filesOnly`: **181 files** (up from 180), `tests/unit/chains/utxo/tx.dustLimit.test.ts` collected at last - CI will actually run it now
- **7/7 pass** from the new location (158ms)
- `@/chains/utxo/tx` alias resolves correctly (`@` -> `packages/sdk/src` in `tests/unit/vitest.config.ts`), zero dangling refs to the old `src/` path anywhere
- nit (a) tackled: at-limit case now asserts `typeof builder.finalize === 'function'` - the builder only returns `finalize` after full assembly, so this proves the whole build succeeded, not just 'no dust error'
- nit (b) tackled properly: LTC `2_939n` boundary asserts the chain-specific message `/below the Litecoin dust limit 2940/` and builds at `2_940n`. Not a copy-pasted BTC case - `2_939 > 546` so it would sail past a hardcoded-BTC guard, this genuinely pins `spec.dustLimit` per chain
- pre-existing #1081 UTXO-02 tests riding along into CI collection confirmed passive: `git show -M` rename diff shows the UTXO-02 describe block byte-identical, only the import line + the two flagged spots changed
- `yarn typecheck` clean (after `yarn build:shared`, same as CI ordering), repo-config prettier clean, eslint clean on the moved file

she worky, still an approve from me 🫡

packages/sdk/src/chains/utxo/tx.ts
```diff
@@ -846,6 +846,13 @@ export function buildUtxoSendTx(opts: BuildUtxoSendOptions): UtxoTxBuilderResult
   const inputs = opts.utxos
   if (inputs.length === 0) throw new Error('no UTXOs provided')
   if (opts.amount <= 0n) throw new Error('amount must be greater than zero')
+  // Dust floor on the PRIMARY send output. `spec.dustLimit` otherwise only
+  // gates the change output (via serializeOutputs), so a below-dust send would
+  // build an unrelayable/rejected output and burn an MPC signing ceremony on a
+  // tx that can never confirm. Fail fast, before any sighash work.
+  if (opts.amount < spec.dustLimit) {
```
NeOMakinG, line 853: `q:` boundary asymmetry vs the change output — intentional? — the primary guard uses `amount < spec.dustLimit` (so amount === dustLimit is accepted), which correctly matches core's canonical `validateUtxoRequirements` (`amount < minUtxo[chain]`). But the change path at tx.ts:614 emits change only when `change > dustLimit`, so a change output of exactly dustLimit (e.g. 546 sat BTC) is silently folded into the fee. The two paths therefore disagree at the exact boundary. It's pre-existing and the direction is safe (change-to-fee never yields an unrelayable output), but the new test pins the primary-side `>=` behavior, cementing the asymmetry.

fix: confirm the boundary split is intentional; if you want symmetry, make the change gate `change >= dustLimit` so exactly-dustLimit change is kept rather than burned to miner fee.

NeOMakinG (commented):
gm — multi-lane pass: claude correctness+fund-safety + codex adversarial. runtime-tested on vitest unit — buildUtxoSendTx dust floor 7/7 green at HEAD 304c43c5, non-inert proven — neutralize the guard → 2 tests go red (545n BTC + 2_939n LTC). comment-only.

## testing approach
- vitest unit — buildUtxoSendTx dust floor 7/7 green at HEAD 304c43c5
- non-inert proven — neutralize the guard → 2 tests go red (545n BTC + 2_939n LTC)
- pass order: claude correctness+fund-safety → codex adversarial

## suggestion
### 1. guard is only as strong as UTXO_SPECS.dustLimit; BTC 546 is the P2PKH value applied to a P2WPKH send, and BCH/Dash/Zcash=1000 are approximations — `packages/sdk/src/chains/utxo/tx.ts`
spec.dustLimit is reused from the change-gating map. For Bitcoin (scriptType p2wpkh) the true relay dust is ~294 sat, but the constant is 546 (the P2PKH figure), so a 294–545 sat P2WPKH *primary* send that was previously relayable now throws — the PR body's 'no valid send path changes' is slightly imprecise. Direction is safe (never builds an unrelayable output). Conversely, if any chain's map value sits BELOW its real network dust (Dash's historical ~5460 duff limit is a candidate at 1000), the new guard is under-protective for that chain — it degrades to the pre-PR no-protection behavior, not a regression. Both are pre-existing map facts, not introduced here.
action: no code change required for this PR; optionally file a follow-up to audit Dash/BCH/Zcash dust constants against current network policy, and soften the PR claim to 'rejects economically-dead sub-floor sends' rather than 'no valid send path changes'.

## q
- boundary asymmetry vs the change output — intentional?

## risk
Low — tightening-only, purely additive guard placed BEFORE any sighash/keysign work (tx.ts:853, ahead of the per-input sighash loop at ~945), so no wasted MPC ceremony and no half-broadcast. The `<` semantics match the canonical core `validateUtxoRequirements` layer, so this is defense-in-depth on the RN-bridge builder path rather than a new/divergent policy. Change-output dust is already handled (folded to fee when `<= dustLimit`) and OP_RETURN is 0-value standard, so no secondary-output dust hole is opened. Sweep/consolidation flows are unaffected (they send well above dust; a wallet whose entire balance is sub-dust cannot be spent regardless). Test is non-inert (proven red-on-revert). DRAFT — reviewed per policy, not approving.

## verdict
COMMENTED

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4-codex` adversarial pass. vitest unit — buildUtxoSendTx dust floor 7/7 green at HEAD 304c43c5, non-inert proven — neutralize the guard → 2 tests go red (545n BTC + 2_939n LTC).

packages/sdk/src/chains/utxo/tx.ts
NeOMakinG, line 853: `blocker:` this guard runs ahead of the address-type guards and breaks an existing fund-safety test — CI `Unit Tests` is red on it right now.

```
FAIL tests/unit/chains/utxo-p2sh.test.ts > rejects P2SH fromAddress > throws when fromAddress decodes to DOGE P2SH
AssertionError: expected [Function] to throw error matching /P2SH spending is not supported/
  but got 'amount 100000 is below the Dogecoin dust limit 1000000'
```

Reproduced locally on `304c43c5`; the same file is 19/19 green on `origin/main`, so it is branch-introduced (CI job 86356422874 shows the identical single failure).

It isn't just a test-ordering nit — the comment at `tx.ts:862-866` says the address checks were deliberately hoisted above the fee/funding math so a funding error can't mask the meaningful address error. Placing the dust check even earlier re-introduces exactly that masking: a P2SH `from` address on Dogecoin now reports "your amount is too small".

suggested fix: move this block to just after the `fromDec.type !== spec.scriptType` guard (~line 890). Still far ahead of `serializeOutputs` and the sighash loop, so nothing about fail-before-signing changes — and you need to be below the decode anyway to fix the recipient-script-type issue on the throw line below.

packages/sdk/src/chains/utxo/tx.ts
```diff
@@ -846,6 +846,13 @@ export function buildUtxoSendTx(opts: BuildUtxoSendOptions): UtxoTxBuilderResult
...
+    throw new Error(`amount ${opts.amount} is below the ${opts.chain} dust limit ${spec.dustLimit}`)
```
NeOMakinG, line 854: `should-fix:` `spec.dustLimit` is keyed to the SENDER's script type but you're applying it to the RECIPIENT's output, so the hole this PR set out to close is still open on Litecoin. Reproduced on this branch:

```
Litecoin, toAddress = LVuDpNCSSj6pQ7t9Pv6d6sUkLKoqDEVUnJ (legacy P2PKH), amount = 3_000n
-> BUILT, hashes=1
```

`Litecoin: { scriptType: 'p2wpkh', dustLimit: 2_940n }` and the comment above it is explicit that 2940 is the **P2WPKH** figure at LTC's 10x `DUST_RELAY_TX_FEE`. The P2PKH floor at that same relay fee is ~5460 (P2SH ~5400). `toAddress` is unconstrained — only `fromDec.type` is checked against `spec.scriptType` at line 886, and `utxo-p2sh.test.ts:314` proves a P2SH *destination* builds fine. So 2_940..5_459 litoshi to an `L...`/`M...` address still burns a ceremony on a dust output.

The message is also inaccurate in the other direction on 4 of 6 chains — BTC 546 is the P2PKH figure (P2WPKH is 294), so `amount: 400n` to a `bc1q...` recipient now throws though the output is perfectly relayable (verified). BCH/Dash 1000 vs real 546, Zcash 1000 vs 54 zats post-NU5. Only Dogecoin's 1_000_000 is right.

suggested fix: derive the floor from `toDec.type` (`p2pkh -> 3 x dustRelayFee x (34+148)`, `p2wpkh -> x (31+67)`, `p2sh -> x (32+148)`), or at minimum widen `UtxoChainSpec` to a per-chain `{ p2pkh, p2wpkh, p2sh }` triple. If you'd rather keep one conservative constant, reword to `below the SDK's conservative dust floor for <chain>` so a refused user isn't given a number that isn't the dust limit for their output — and soften the PR body's "no valid send path changes".

packages/sdk/tests/unit/chains/utxo/tx.dustLimit.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    const builder = buildUtxoSendTx({ ...baseOpts, amount: 546n })
+    expect(typeof builder.finalize).toBe('function')
+  })
+
+  it('uses the PER-CHAIN dust limit, not a hardcoded BTC 546 (Litecoin 2_940n boundary)', () => {
+    const ltcOpts = { ...baseOpts, chain: 'Litecoin' as const, fromAddress: LTC_P2WPKH, toAddress: LTC_P2WPKH }
+    // 2_939 is fine on BTC (>546) but dust on Litecoin (<2_940) — proves the
+    // guard reads spec.dustLimit per chain rather than a constant.
+    expect(() => buildUtxoSendTx({ ...ltcOpts, amount: 2_939n })).toThrow(/below the Litecoin dust limit 2940/)
+    expect(() => buildUtxoSendTx({ ...ltcOpts, amount: 2_940n })).not.toThrow()
```
NeOMakinG, line 80: `should-fix:` this boundary pair asserts 2_940n is a safe Litecoin send, but that's only true when the recipient is bech32. Against a legacy `L...` recipient the real floor is ~5_460 litoshi, so this test codifies the half of the range that still slips through (see the comment on `tx.ts:854`). Worth extending with a `toAddress`-varying case once the floor is derived from `toDec.type`.

Separately, a note on the boundary convention this file pins: the primary guard accepts `amount === dustLimit`, but `serializeOutputs` (`tx.ts:614`) emits change only when `change > dustLimit`, so change of exactly `dustLimit` is dropped. Reproduced on `origin/main` — BTC input 10_692 / amount 10_000 / fee 146 -> one output, 692 paid as fee, 546 sat burned. Bitcoin's rule is `value < threshold` => dust, so `change >= dustLimit` is the correct comparison and would make the two halves agree with what this test already asserts.

NeOMakinG (commented):
gm — round 2 on the same HEAD `304c43c5`. Byte-identical diff to my r1, so I spent the whole budget on execution rather than re-reading: real probes through `buildUtxoSendTx` on both `origin/main` and the branch, a mutation pass, and the full `tests/unit/chains/utxo` selection instead of just the new file.

**That last one is the headline: my r1 said "7/7 green" — it was wrong.** I ran only `tx.dustLimit.test.ts`. The wider selection is RED, and so is CI. This PR breaks an existing fund-safety test. Still comment-only (DRAFT), but this is now a blocker rather than a suggestion.

## testing approach
- worktrees `/tmp/wt-sdk-1142` (PR head `304c43c5`) and `/tmp/wt-sdk-1142-main` (`origin/main` @ `802910d7`), `node_modules` symlinked from the canonical checkout, `vitest run --config tests/unit/vitest.config.ts`.
- **branch, full utxo selection:** `tests/unit/chains/utxo` → `Tests 1 failed | 104 passed`. **baseline, same file on `origin/main`:** `utxo-p2sh.test.ts` → `19 passed`. The failure is branch-introduced.
- **CI agrees** — job [86356422874](https://github.com/vultisig/vultisig-sdk/actions/runs/29091118398/job/86356422874) `Unit Tests: fail`, `Test Files 1 failed | 180 passed`, identical single assertion. `Test Summary` is also `fail`. (Lint and Type Check + Integration are green — the usual trap where the green jobs are the ones you look at.)
- **mutation-verified the new tests:** deleted the guard body → `2 failed | 5 passed` (545n BTC + 2_939n LTC). The guard is non-inert. Restored.
- **executed the premise on `origin/main`** rather than reasoning about it — decoded `unsignedRawHex` outputs for a below-dust send (see below).
- probed the amount type with `'545'`, `545`, `NaN`, `Infinity`, `-0`, `'1e3'`, `600.5` on both refs.
- traced the real consumer: `vultiagent-app` `src/services/utxoTx.ts:101` (`buildSignBroadcastUtxoSend`) — the only production caller anywhere; nothing in this repo calls it.
- lanes: claude correctness+fund-safety (load-bearing). Codex stalled twice (full prompt and a narrowed retry, both empty at the cap) — lane declared dead, not load-bearing.

### what a below-dust send does TODAY, concretely
On `origin/main`, `amount: 545n` on Bitcoin (dust 546) **builds and signs**:
```
BTC 545n outputs: [{"value":"545",...},{"value":"99309",...}]
signingHashes: 1
```
`1n` behaves the same. The primary output is **not** dropped and **not** absorbed into the fee — it is emitted at full value, one sighash is produced, and a full MPC ceremony runs. So the honest severity is **wasted signing ceremony + a late, confusing relay rejection — not fund loss.** The PR is right to fix it; it is not a P0.

The silent-value-destruction case does exist, but on the **change** side, which this PR does not touch — see `q` #5.

## prior issues — status
|---|---|---|---|---|---|
| r1 | 1 | suggestion | `spec.dustLimit` is the P2PKH figure applied to a P2WPKH chain; BCH/Dash/Zcash `1000` are approximations | ⚠️ **upgraded + partially corrected** | Now measured, not speculated. BTC `400n` → a bech32 recipient **throws** (real P2WPKH dust 294) — confirmed over-refusal. My r1 guess that Dash's floor might be ~5460 was wrong (Dash inherits Bitcoin's 546, so `1000` over-refuses too). But r1 missed the direction that actually matters — **Litecoin under-refuses** for legacy recipients. Now `should-fix` #2/#3. |
| r1 | 2 | q | boundary asymmetry: primary uses `< dustLimit`, change uses `> dustLimit` | ⚠️ **still open, now quantified** | Reproduced on main: input 10_692, amount 10_000, fee 146 → change exactly 546 → **single output, 692 paid as fee**. 546 is not dust by policy (dust is `value < threshold`), so 546 sat is burned. `should-fix` #4. |
| r1 | 3 | — | r1 claimed the suite was green (`7/7`) | ❌ **wrong — retracted** | I ran only the new file. The wider selection and CI are both red. `blocker` #1. |
| r1 | 4 | — | r1 claimed "change-output dust is already handled" | ⚠️ **understated** | It is handled by *silently burning it*, unbounded up to `dustLimit`. On DOGE that is 0.01 DOGE per tx. `q` #5. |

## blocking
### 1. the guard is placed ahead of the address-type guards and breaks `utxo-p2sh.test.ts` — CI is red — `packages/sdk/src/chains/utxo/tx.ts:853`
`tests/unit/chains/utxo-p2sh.test.ts > buildUtxoSendTx — rejects P2SH fromAddress > throws when fromAddress decodes to DOGE P2SH` now gets:
```
AssertionError: expected [Function] to throw error matching /P2SH spending is not supported/
  but got 'amount 100000 is below the Dogecoin dust limit 1000000'
```
Reproduced locally on the branch and confirmed on CI; `origin/main` runs that file 19/19 green.

This is not a stale test — it is a real ordering regression, and the existing code already documents the class. `tx.ts:862-866` says the address checks run *before* any fee/funding math specifically so that a funding error cannot mask the meaningful address error. The new guard sits even earlier and re-introduces exactly that masking: paste a P2SH `from` address on Dogecoin and you are told your amount is too small.

Fix: move the dust check to just after the `fromDec.type !== spec.scriptType` guard (~`tx.ts:890`). That is still far ahead of `serializeOutputs` and the sighash loop, so the fail-before-signing property is preserved, and it restores the P2SH test — and it is a prerequisite for #2 anyway, since the recipient's decoded type is only available after the decode.

## should-fix
### 2. `spec.dustLimit` is keyed to the SENDER's script type but applied to the RECIPIENT's output — the hole stays open on Litecoin — `packages/sdk/src/chains/utxo/tx.ts:853`
Reproduced on the branch:
```
Litecoin, toAddress = LVuDpNCSSj6pQ7t9Pv6d6sUkLKoqDEVUnJ (legacy P2PKH), amount = 3_000n
-> BUILT, hashes=1
```
`Litecoin: { scriptType: 'p2wpkh', dustLimit: 2_940n }` — and the in-file comment is explicit that 2940 is the **P2WPKH** figure at LTC's 10× `DUST_RELAY_TX_FEE`. At that same relay fee the **P2PKH** floor is ~5460 and P2SH ~5400. `toAddress` is completely unconstrained (only `fromDec.type` is checked against `spec.scriptType`, `tx.ts:886`; `utxo-p2sh.test.ts:314` proves a P2SH *destination* builds fine). So an LTC send of 2_940..5_459 litoshi to an `L...`/`M...` address still burns a ceremony on a dust output — the exact scenario #1137 is about.

Derive the floor from `toDec.type` instead (`p2pkh → 3 × dustRelayFee × (34+148)`, `p2wpkh → × (31+67)`, `p2sh → × (32+148)`), or at minimum widen `UtxoChainSpec` to a `{ p2pkh, p2wpkh, p2sh }` triple per chain. Requires the guard to move below the decode, which #1 requires anyway.

### 3. four of the six chains over-refuse, and the error message asserts the number is "the dust limit" — `packages/sdk/src/chains/utxo/tx.ts:854`
Enumerated every chain that reaches this code:

| chain | `spec.dustLimit` | actual relay dust | effect of the new guard |
|---|---|---|---|
| Bitcoin | 546 | P2WPKH 294 / P2TR 330 / P2SH 540 / P2PKH 546 | over-refuses 294–545 to bech32 (**verified**: `amount 400n` → throws) |
| Litecoin | 2_940 | P2WPKH 2_940 / P2SH ~5_400 / P2PKH ~5_460 | **under**-refuses to legacy — see #2 |
| Dogecoin | 1_000_000 | 0.01 DOGE hard min | correct |
| Bitcoin-Cash | 1_000 | 546 | over-refuses 546–999 |
| Dash | 1_000 | 546 | over-refuses 546–999 |
| Zcash | 1_000 | 54 zats (post-NU5) | over-refuses 54–999 |

The over-refusals are small absolute amounts and conservative-safe, so I am not asking you to loosen them. But the message says `is below the Bitcoin dust limit 546` when 546 is *not* the dust limit for the output being built, so the user is refused with a factually wrong reason — and the PR body's "no valid send path changes" is not accurate. Either derive the real floor (#2 covers this) or reword to something like `below the SDK's conservative dust floor for <chain>` and adjust the PR body claim.

### 4. the two halves now disagree at the exact boundary; change of exactly `dustLimit` is burned — `packages/sdk/src/chains/utxo/tx.ts:614`
Pre-existing, but this PR pins the opposite convention right next to it: the new primary guard accepts `amount === dustLimit`, while `serializeOutputs` emits change only when `change > dustLimit`. Reproduced on main (BTC, input 10_692, amount 10_000, fee 146):
```
change===546 outputs: [{"value":"10000",...}]      // one output — change dropped
input 10692  outTotal 10000  impliedFee 692        // 546 sat burned
```
Bitcoin's rule is `value < threshold` ⇒ dust, so a 546-sat change output is perfectly relayable. `change >= dustLimit` is the correct comparison — a one-character fix while you are in the file, and it makes the two halves agree with the semantics the new test already pins.

## q
### 5. change-output dust is still silently burned, unbounded up to `dustLimit` — is the PR body's "dust" story meant to cover only half of this?
Reproduced on the branch, Dogecoin, 1 DOGE send with change of 999_999 koinu:
```
DOGE nominal fee = 228000   outputs = 1
DOGE actual fee paid = 1227999   (extra burned = 999999 koinu)
```
0.00999999 DOGE vanishes into the fee with no signal. This is by design (`select.ts:37` documents it) and this PR does not change it, but the primary amount now fails **loud** while the change still fails **silent**. Worth one line in the PR body so a reader does not conclude UTXO dust is now fully handled — and worth an issue for surfacing burned change to the user before signing.

## praise
The test relocation is the load-bearing half of this PR and it deserves calling out. `packages/sdk/vitest.config.ts` includes `tests/**` and `tests/unit/vitest.config.ts` includes `./**` — neither globs `src/`, and after this change there are **zero** `*.test.ts` under `packages/sdk/src`. The UTXO-02 Litecoin `2_940n` regression assertion had never executed in CI. Moving it is a genuine catch.

- `serializeOutputs` change comparison off-by-one (#4) — on `origin/main` too.
- unbounded silent change burn (#5) — on `origin/main` too.

## risk
Medium-low as written, and the direction is right. The guard is correctly placed before any sighash work and before `fastVaultSign` in the only production consumer (`vultiagent-app/src/services/utxoTx.ts:101`), so it genuinely does prevent a wasted MPC ceremony — that part of the premise holds and I verified it end to end.

The recurring narrow-reader type bug is **not** present here: `amount` is declared `bigint`, and every non-bigint I probed dies loudly. On `origin/main` `'545'` / `545` / `NaN` / `Infinity` / `'1e3'` / `600.5` all throw `Cannot mix BigInt and other types`; `-0` hits the zero guard. On the branch `'545'` and `545` now surface the dust message instead (harmless), the rest still die at the arithmetic. No value silently slips past either guard. `vultiagent-app` additionally enforces `typeof amount === 'bigint'` at its own boundary before calling in. Boundary values behave: 545 throws, 546 builds, 2_939 LTC throws, 2_940 LTC builds — all executed, not read.

No dependency ranges touched, so no caret hazard. Send-max and consolidation are unaffected (`utxoConsolidate.ts` does not route through this builder, and both produce amounts far above any of these floors).

The residual risk is entirely #1 — shipping a red suite that silently reorders which fund-safety error a user sees.

## verdict
COMMENTED — DRAFT, so not approving per policy. Land #1 (it is red on CI right now) and #2 (the hole this PR set out to close is still open on Litecoin legacy recipients); #3/#4 are cheap and in-file; #5 is a body/issue note.

🤖 reviewed via [Claude Code](https://claude.com/claude-code). Round 2 at HEAD `304c43c5`. Evidence: branch `tests/unit/chains/utxo` → 1 failed / 104 passed; `origin/main` `utxo-p2sh.test.ts` → 19 passed; CI `Unit Tests: fail`; guard mutation → 2 new tests red. Codex lane stalled twice and was dropped.

NeOMakinG (approved):
LGTM

gomesalexandre:
Verified the CI-red ordering blocker is already fixed on this branch (utxo-p2sh.test.ts 19/19 pass — dust check runs after both the P2SH refusal and the scriptType guard, matching your suggested placement).

Fixed the change-output boundary asymmetry: `serializeOutputs`'s change gate now uses `>=` instead of `>`, matching Bitcoin's own `value < threshold => dust` rule and the primary guard's own boundary — a change output exactly at `dustLimit` no longer gets silently folded into the miner fee.

On the per-script-type dust-floor gap (the bigger should-fix): took your own offered fallback rather than guess. Reworded the error message + doc comment to say "SDK's conservative dust floor" instead of claiming to be *the* dust limit for that output, since `spec.dustLimit` is sized for the chain's scriptType, not the recipient's actual decoded type. Documented the known under/over-reject gap explicitly in the `UtxoChainSpec.dustLimit` doc comment as a real follow-up — deriving accurate per-script-type floors (p2pkh/p2wpkh/p2sh × 6 chains) from verified dust-relay-fee constants needs research I can't responsibly do blind in this pass; didn't want to ship guessed numbers into a fund-safety-adjacent path.

The negative-side hardening note (unlisted spender/kind fields — different PR, not this one) not applicable here. 75/75 relevant tests pass, tsc clean, rebased onto main.

rcoderdev:
Closing this because the dust guard remains useful, but this unapproved head uses one sender-script threshold that over-refuses valid outputs and under-protects Litecoin legacy recipients, so it needs a fresh chain-aware implementation.

PR #1144, opened 2026-07-10
fix(encoding): reject negative values in hex/byte amount encoders

### LINKED ISSUE #1139: encoding: bigIntToHex/numberToEvenHex turn a negative bigint into empty bytes instead of rejecting it

Severity: LOW. Effort: S.

gomesalexandre:
Closes #1139

## what

`bigIntToHex`, `numberToEvenHex` and Cardano's `amountToBytes` all encode via `.toString(16)`. A **negative** value stringifies with a leading `-` (`"-1"`), and the downstream `Buffer.from(hex, 'hex')` silently turns that into empty / garbage bytes. On the amount fields these encoders feed, the MPC co-signers would then sign over a value unrelated to the input — a silent-corruption fund-safety class.

## how

Added a fail-closed `RangeError` on negative input to each of the three:
- `packages/lib/utils/bigint/bigIntToHex.ts`
- `packages/lib/utils/hex/numberToHex.ts` (`numberToEvenHex` — covers all three input types: `number` / `bigint` / `Long` via `.isNegative()`)
- `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts` (`amountToBytes`)

Verified every caller (EVM nonce/value, TRON/Polkadot/TON amounts, chainId, Cardano token amounts) only ever passes non-negative values, so the throw is a pure safety net — no valid path changes behaviour.

## testing

- `bigIntToHex.test.ts` / `numberToHex.test.ts` (new): positive values still encode to even-length hex; negative `number` / `bigint` / `Long` throw `RangeError`.
- `cardano.test.ts`: a token send with a negative `toAmount` now throws `RangeError` at `amountToBytes` instead of building a corrupted signing input.
- Gates: targeted vitest **11/11 passed**, `@vultisig/sdk typecheck` ✅, eslint ✅ (0 errors), prettier ✅.

## changeset

## risk

Low, tightening-only. Non-negative inputs are unaffected; only genuinely-negative inputs (which previously produced silently-corrupted bytes) now fail loudly.

🤖 Agent-generated


gomesalexandre (commented):
Adversarial pass - verified everything myself instead of trusting the body.

**Verified clean:**
- Premise reproduced on `main` (d9ad0922) before reading the fix: `bigIntToHex(-5n) -> '-5'` and `Buffer.from('-5', 'hex') -> <Buffer >` (empty), `bigIntToHex(-767n) -> '-2ff'` -> 0-length buffer, `numberToEvenHex(Long.fromNumber(-5)) -> '-5'`. The silent-truncation class is real.
- All **3** claimed encoders got the guard: `bigIntToHex`, `numberToEvenHex` (covering all three input arms - `number`, `bigint`, and `Long` via `.isNegative()`, which is also correct for unsigned Longs), and cardano's `amountToBytes`.
- Positives-unchanged is actually tested, not just negatives-throw: `bigIntToHex(0n) === '00'`, `255n -> 'ff'`, `256n -> '0100'`, plus the positive number/bigint/Long cases in `numberToHex.test.ts`. The guard is purely additive ahead of the existing logic, so the positive path is byte-identical.
- Grepped every caller (EVM nonce/value/chainId, blockaid simulation + validation inputs, polkadot amount, tron token amounts, ton jetton): all non-negative-by-construction, so the throw is a pure safety net as claimed.
- Re-ran gates on a fresh clone: targeted vitest **11/11** pass, `@vultisig/sdk typecheck` clean, eslint 0 errors, changeset present (`@vultisig/sdk: patch`), diff scope tight (7 files, all on-topic).

**suggestion:** the `number` arm of `numberToEvenHex` still lets `NaN` and fractionals through to the *exact same* corruption the PR fixes: `NaN < 0` is `false`, `NaN.toString(16)` is `'NaN'`, `(1.5).toString(16)` is `'1.8'`, and `Buffer.from(...)` of either is empty bytes. A `if (typeof amount === 'number' && !Number.isSafeInteger(amount)) throw` at https://github.com/vultisig/vultisig-sdk/blob/5552843e9abdeb95966d77e33f2cec5ad772b2d5/packages/lib/utils/hex/numberToHex.ts#L5-L12 would close the whole class whilst in the house. Non-blocking - current callers pass integer chainIds/amounts.

**q:** this and #1140 both touch `cardano.ts` a few lines apart (`amountToBytes` vs the transfer amount) - whichever lands second needs a trivial rebase, just flagging the ordering.

**Verdict: APPROVE** (posted as comment since draft). Clean, tight, premise verified, fail-closed in the right direction.

packages/lib/utils/hex/numberToHex.ts
```diff
@@ -2,6 +2,14 @@ import Long from 'long'
 export const numberToHex = (num: number) => `0x${num.toString(16)}`
 
 export const numberToEvenHex = (amount: number | Long | bigint) => {
+  // A negative value stringifies with a leading "-", which downstream
+  // `Buffer.from(hex, 'hex')` silently turns into empty/garbage bytes. Fail
+  // closed rather than emit a corrupted amount for co-signing.
+  const isNegative =
+    typeof amount === 'bigint' ? amount < 0n : typeof amount === 'number' ? amount < 0 : amount.isNegative()
```
NeOMakinG, line 9: `suggestion:` the guard closes the negative-sign case, but the `number` branch (`amount < 0`) still lets a **non-integer** number through — `amount.toString(16)` then yields fractional hex (`1.5 → "1.8"`, `0.1 → "0.1999…"`) which `Buffer.from(hex,'hex')` truncates/empties, the same silent-corruption shape you're fixing here. Not reachable today (every signing caller passes `Long`/`bigint`), so helper-hardening only: consider `Number.isInteger(amount)` on the number branch to make the encoder safe-by-default.

NeOMakinG (commented):
Two-lane review (Claude load-bearing + Codex `gpt-5.4`). Reviewed in isolated worktree at HEAD `5552843e`. **Draft → posting as COMMENTED, not approving.** vultisig-sdk is fund-critical (keysign/encoding), reviewed with a worst-case adversarial lens.

Net: this is a clean, correct **defensive tightening**. The three named encoders (`bigIntToHex`, `numberToEvenHex`, Cardano `amountToBytes`) are the fund-safety layer here — I confirmed there is **no central non-negative validation** of `toAmount` anywhere upstream in `packages/core/mpc/keysign`, so guarding at the encoder is the right place. All guards throw **before** any signable payload is produced (signing-input construction), i.e. fail-closed at the correct layer, and I verified every live caller only ever passes non-negative amounts/nonces/values — so no legitimate path changes behaviour. No blockers.

## testing approach
- Full static trace of all callers of the three encoders (EVM value/nonce, TRON, Polkadot, ERC20 → `bigIntToHex`; TON native+jetton, EVM chainId → `numberToEvenHex`; Cardano token + per-UTXO tokens → `amountToBytes`). All non-negative.
- **Ran the lib encoder tests in-worktree (real vitest, symlinked deps): 6/6 green at HEAD.**
- **Red-on-revert proven (non-inert):** neutralized both lib guards → the 4 negative-input tests go RED (`expected function to throw ... but it didn't`), the 2 positive tests stay green. The tests genuinely pin the new behaviour.
- Confirmed the Cardano test reaches `amountToBytes`: `tokenBundle` IIFE is evaluated (and throws) during `getCardanoSigningInputs`, so `toThrow(RangeError)` exercises the real guard.
- Coercion edges checked: `-0` (number) → `-0 < 0` is false → `"0"` (safe, no false-positive); bigint has no wrap (arbitrary precision, negative stays negative → caught); `Long.fromString('-1').isNegative()` → true → caught (also closes the TON native/jetton negative path).

## should-fix
1. **Sibling amount→hex path on the TRON TRC20 flow is still unguarded** — `packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts:30` (`buildTrc20TransferParameter`) does a raw `amount.toString(16)` with no negative check, fed `getKeysignAmount(keysignPayload)` via `getTronChainSpecific` (`tron/index.ts:59`). A negative `toAmount` produces a malformed ABI word and fires the TronGrid `triggerconstantcontract` fee-estimation RPC **before** the signing-layer `bigIntToHex` guard (`signingInputs/resolvers/tron.ts:210/273`) trips. Net effect is NOT a fund-safety hole — the actual signing encoder now throws `RangeError` and aborts the keysign, so no corrupted signature is produced — but it means the PR's implicit "reject negative on every hex-encoding entry point" is not fully true, and a bad-parameter RPC still goes out. Cheap to close with the same `< 0n` guard. (Codex lane.)
2. **Cardano native (non-token) amount bypasses the new guard** — for a native send (`contractAddress === ''`), `amountToBytes` is never called; the amount flows through `Long.fromString(keysignPayload.toAmount)` (`cardano.ts:62`, and per-UTXO input at `:69`) straight into the WalletCore proto. A negative `toAmount` there is not caught by this PR (WalletCore's uint64 amount would wrap huge and the planner would reject on balance — so it fails, but not via a clean guard and not on the encoder path this PR hardens). Same silent-corruption class as the token path you just closed; worth mirroring the `< 0n` throw so native and token Cardano behave identically. (Claude lane.)

## suggestion
- `numberToEvenHex` still silently corrupts **non-integer `number`** inputs — the new guard only rejects `< 0`, then `amount.toString(16)` runs for all remaining numbers, so e.g. `1.5 → "1.8"`, `0.1 → "0.1999…"`, and `Buffer.from(hex,'hex')` truncates/empties those the same way a `-` does. Not a demonstrated live exploit (every current signing caller passes `Long`/`bigint`, not a float), so this is helper-hardening only, not fund-critical. If you want the helper safe-by-default, add `Number.isInteger(amount)` to the number branch. (Codex + Claude converged.)

## risk
Low, tightening-only. Non-negative inputs unaffected (verified across all callers + ran green). Only genuinely-negative inputs — which previously produced silently-corrupted signing bytes — now fail loudly at input-construction time, before any signature. Changeset (`@vultisig/sdk: patch`) is correctly scoped.

## verdict
COMMENTED (draft — held from approval per policy). Zero blockers; the core fix is correct, well-placed, and its tests are proven non-inert. The two should-fix items are completeness gaps in sibling amount paths (both fail-closed today, neither a fund hole), and the suggestion is opt-in helper hardening. Happy to approve on ready once the draft flips and — ideally — the TRON fee.ts / Cardano-native siblings get the same one-line guard.

<sub>🤖 two-lane review: Claude (load-bearing, static trace + red-on-revert-proven vitest) + Codex gpt-5.4 (adversarial second opinion). GLM lane retired.</sub>

NeOMakinG (commented):
Round 2, same HEAD `5552843e` (no new commits since my round-1 pass). Per the byte-identical-re-review rule I spent this round on **runtime probes instead of a second diff read** — I executed the encoders and the real TON/Cardano signing-input builders rather than reasoning about them. That turned up one thing static tracing missed. Two lanes: Claude (load-bearing) + Codex `gpt-5.4`. **Draft → COMMENTED, not approving.**

## testing approach

Isolated worktree `/tmp/wt-sdk-1144` at `5552843e`, real vitest via `.config/vitest.core.config.ts` (source aliases — a configless run silently fails to resolve `@vultisig/*` and reports "no tests"; worth knowing).

- **PR selection green:** `numberToHex.test.ts` + `bigIntToHex.test.ts` + `cardano.test.ts` → **11/11 passed**, matching the body. All 5 CI checks green.
- **Mutation-verified the Cardano guard this round** (round 1 only reasoned about it): neutralised `if (amount < 0n)` → `if (false)` and the new test goes **red** (`expected function to throw an error, but it didn't`), 4 others stay green. Combined with round 1's red-on-revert on both lib guards, all three new tests are proven non-inert.
- **Ran ~40 adversarial inputs through `numberToEvenHex` / `bigIntToHex`** and captured the resulting hex *and* the `Buffer.from(hex,'hex')` bytes.
- **Drove the real builders end-to-end**: `buildNativeTonTransfer` / `buildNativeTonTransferFromMessage` and `getCardanoSigningInputs`, reading the actual proto amount off the returned signing input.

**What a negative value does today, concretely** (`numberToEvenHex` at HEAD):

```
  number -1                       => THROW RangeError            ✅ caught
  bigint -1n                      => THROW RangeError            ✅ caught
  Long.fromNumber(-5)             => THROW RangeError            ✅ caught
  number -0                       => "00"  bytes=<00>            ✅ no false positive
  number 0                        => "00"  bytes=<00>            ✅ legit zero still encodes
  Long.MAX_UNSIGNED_VALUE         => "ffffffffffffffff"          ✅ legit max still encodes
  number NaN                      => "0NaN"      bytes=<>  EMPTY ❌ not caught
  number Infinity                 => "Infinity"  bytes=<>  EMPTY ❌ not caught
  number 1.5                      => "01.8"      bytes=<01> = 1  ❌ not caught (1.5 → 1)
  number 0.1                      => "0.1999999999999a" bytes=<> ❌ not caught
```

So the headline class the PR names — a leading `-` surviving into `Buffer.from` — **is genuinely closed on all three encoders for every input type they're declared to accept**. Positives, `-0`, zero and max are byte-identical to before. That part is correct and well-placed.

## prior issues — status

|---|---|---|---|---|---|
| 1 | 1 | should-fix | TRON `buildTrc20TransferParameter` (`chainSpecific/resolvers/tron/fee.ts:30`) does a raw `amount.toString(16)` with no negative check | ❌ still open | Re-read at HEAD, unchanged. Not a fund hole (the signing-layer `bigIntToHex` still aborts the keysign) but a malformed-ABI fee-estimation RPC still goes out first. |
| 1 | 2 | should-fix | Cardano **native** (non-token) amount bypasses `amountToBytes` | ❌ still open — **and now reproduced with real numbers**, see blocking-adjacent item below | Round 1 I asserted this from a static read; this round I ran it. |
| 1 | 3 | suggestion | `numberToEvenHex` still corrupts non-integer / `NaN` / `Infinity` `number` inputs | ❌ still open | Now have byte-level evidence (table above). Same as gomes' own review suggestion. Still not live-reachable — every current caller passes `bigint`/`Long`. |
| — | — | — | Retry-wrapper misclassification (`isTransientBroadcastError`) | ℹ️ n/a, closed | Codex lane confirmed neither `withTransientBroadcastRetry` nor `isTransientBroadcastError` exists in this repo, and `BroadcastService` builds+compiles signing inputs (`BroadcastService.ts:91`/`:105`) before `coreBroadcastTx()` at `:116`. The `RangeError` can never reach a broadcast retry classifier. |
| — | — | — | Caret-bump hazard | ℹ️ n/a | Diff touches no dependency ranges (7 files, changeset + 3 sources + 3 tests). |

## preferably-blocking

**1. A negative amount can still produce an enormous positive signable amount — the guard is bypassed by int64 saturation *upstream* of it.** This is the exact failure shape the PR sets out to close, and I reproduced it end-to-end through the real builders.

`Long.fromString()` saturates instead of throwing, so a sufficiently-negative decimal string becomes a **positive** `Long` before `numberToEvenHex` ever sees it — and `.isNegative()` is then `false`, so the new guard passes it straight through:

```
buildNativeTonTransfer   toAmount=-1                     => THROW RangeError                       ✅
buildNativeTonTransfer   toAmount=-1000000000            => THROW RangeError                       ✅
buildNativeTonTransfer   toAmount=-9223372036854775809   => NO THROW  bytes=<7fffffffffffffff>
                                                            = 9223372036854775807 nanoton
                                                            = 9,223,372,036 TON                    ❌
buildNativeTonTransfer   toAmount=-18446744073709551615  => NO THROW  bytes=<01>  = 1 nanoton      ❌
```

Both `buildNativeTonTransfer` (`resolvers/ton/native.ts:52` → `:56`) and `buildNativeTonTransferFromMessage` (`:75`, whose `amount: string` comes off a dApp/TON-Connect message) behave identically. The composition is direct — `Long.fromString(toAmount)` then `Buffer.from(numberToEvenHex(amount),'hex')` — there is no intervening validation.

Same root cause on **Cardano native**, which is round-1 item #2 with the numbers filled in (`getCardanoSigningInputs`, `cardano.ts:55`):

```
native toAmount=-1                    => NO THROW, proto amount = -1  → uint64 18446744073709551615
native toAmount=-1000000              => NO THROW, proto amount = -1000000 → uint64 18446744073708551616
native toAmount=-9223372036854775809  => NO THROW, proto amount = 9223372036854775807
```

For a native ADA send `amountToBytes` is never called (`isTokenSend` is false), so `transferMessage.amount` goes to WalletCore as a `Long` whose proto type is **uint64** — a plain `-1` lands as 18.4 quintillion lovelace. In practice both chains would then fail on balance/planning, so I'm not claiming a signed-and-broadcast loss. But the PR's premise is *"be a fail-closed safety net for a negative amount"*, and on these two paths a negative amount is silently converted to a huge positive one **before** the safety net runs — which is strictly worse than the empty-bytes failure the PR fixes, because empty bytes fail loudly and `9223372036854775807` looks like a legitimate number.

Cheapest complete fix is to guard at the **string → numeric boundary**, where the value is still exact, rather than at the encoder:

```ts
// resolvers/ton/native.ts and resolvers/cardano.ts
const parseAmount = (raw: string): Long => {
  const value = BigInt(raw)                       // exact, unbounded, throws on garbage
  if (value < 0n) throw new RangeError(`negative amount: ${raw}`)
  if (value > 0xffffffffffffffffn) throw new RangeError(`amount exceeds uint64: ${raw}`)
  return Long.fromString(raw, true)
}
```

Note `Long.fromString` also mangles non-negative garbage the same way — `"abc"` → `0`, `"0x10"` → `0`, `"1e9"` → `1`, `"1.5"` → `1`, `"99999999999999999999"` → `7766279631452241919`. All pre-existing and out of scope, but the `BigInt()` parse above closes that whole family for free. (Related: `Long.fromString("18446744073709551615")` — a legitimate max-uint64 — parses *signed* to `-1` and now throws `"cannot hex-encode negative value -1"`; it was already broken on `main`, producing empty bytes, so this is an improvement, just with a confusing message. The unsigned parse above fixes it properly.)

## should-fix

**2. TRON `buildTrc20TransferParameter` still unguarded** — `packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts:30`, `const amountHex = amount.toString(16)` fed `getKeysignAmount(keysignPayload)` via `getTronChainSpecific` (`tron/index.ts:59`). A negative `toAmount` builds a malformed ABI word and fires the TronGrid `triggerconstantcontract` RPC **before** the signing-layer `bigIntToHex` guard trips. Not a fund hole — the signing encoder now aborts the keysign — but it means "reject negative on every hex-encoding entry point" isn't fully true yet, and it's the same one-line `if (amount < 0n) throw` you already wrote three times. (Carried from round 1, unchanged.)

## suggestion

**3. `numberToEvenHex`'s `number` arm still silently corrupts `NaN` / `Infinity` / fractionals** — see the probe table. `NaN` and `Infinity` produce **zero bytes**, `1.5` produces `<01>`, i.e. exactly the corruption the PR fixes for `-`. Not live-reachable (every caller passes `bigint` or `Long`: `getEvmTwChainId.ts:13` does `BigInt(...)`, jetton's `forwardAmount` is `1n`/`0n`, `tonConfig.jettonAmount` is a `toChainAmount` bigint), so this is helper-hardening only. `if (typeof amount === 'number' && !Number.isSafeInteger(amount)) throw` in the same guard block closes it. (Converged: Claude + Codex + gomes' own review.)

## q

**4.** Non-`number`/`bigint` inputs now hit `amount.isNegative()` and throw a **`TypeError`, not a `RangeError`** — `numberToEvenHex('100' as any)` used to return `"0100"` and now throws `amount.isNegative is not a function`. Fail-closed and almost certainly fine (TS types forbid it, and `'-1'` previously produced empty bytes so a throw is an improvement), but if any untyped JS consumer passes a numeric string this is a behaviour change worth a `typeof amount === 'string'` early normalise or an explicit type assertion for a clean error. Have you checked the RN/CLI consumers for string amounts?

**5.** Still applies from gomes' review: this and #1140 both touch `cardano.ts` — whichever lands second needs a rebase.

## risk

Low, tightening-only, and I re-verified there's no over-refusal: `0`, `-0`, `Long.MAX_UNSIGNED_VALUE` and every realistic amount encode byte-identically to `main`. No legitimate traffic is refused. The residual risk is **incompleteness, not regression** — item 1 means the safety net has a hole on the two paths where a negative amount is laundered into a positive one before it reaches the net.

## verdict

**COMMENTED** (draft — held from approval per policy). No blockers: the three guards are correct, correctly placed (first statement, ahead of every early return), fail-closed, and their tests are mutation-proven non-inert. Item 1 is the one I'd want closed before this lands, because it's the same fund-safety class the PR is named after and it currently produces a *more* convincing wrong amount than the bug being fixed. Happy to approve on ready with the `BigInt()`-boundary parse (or an explicit "out of scope, tracked in #X").

<sub>🤖 two-lane review: Claude (load-bearing — runtime probes through the real TON/Cardano builders, mutation-verified) + Codex gpt-5.4 (adversarial second opinion: swept for other unguarded signing-path encoders, found none beyond the above; confirmed the retry-classifier path is unreachable). GLM lane retired.</sub>

NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because the fix remains useful, but this old head is far behind current main and its release metadata omits lib-utils and core-mpc, so it cannot ship the changed packages safely without a fresh additive implementation.

PR #1168, opened 2026-07-11
fix(config): centralize VULTISIG_CONFIG_DIR resolution

gomesalexandre:
## Summary
- extract a shared `getVultisigConfigDir()` helper in `@vultisig/client-shared`
- route config store, credential store, token cache, broadcast journal, and CLI config through it
- add regression tests for whitespace/empty overrides and cross-surface co-location

## Test plan
- [x] `yarn build:all`
- [x] `yarn lint`
- [x] `yarn test`
- [x] `yarn vitest run --config .config/vitest.cli.config.ts clients/cli/src/lib/__tests__/config.test.ts clients/cli/src/agent/__tests__/broadcastJournal.test.ts clients/cli/src/agent/__tests__/configPaths.test.ts`
- [x] `cd clients/cli && yarn vitest run tests/config-store.test.ts`
- [x] `cd clients/cli && VULTISIG_CONFIG_DIR='   ' node dist/index.js auth status`

## Receipts
- full workspace gates passed: `yarn build:all`, `yarn lint`, `yarn test`
- targeted CLI tests passed: 37 tests across config/journal/path-policy + 13 config-store tests
- runtime proof: with `VULTISIG_CONFIG_DIR='   '`, `node dist/index.js auth status` returned a successful JSON payload instead of resolving `''` / crashing on bad paths

Closes #1166

## Round report
- https://gist.github.com/gomesalexandre/fac3c1408ccefc7e90a16955b11e855b

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty or whitespace-only `VULTISIG_CONFIG_DIR` values now reliably fall back to the default `.vultisig` directory.
  * Configuration, credentials, token cache, and broadcast journal paths now resolve consistently.
  * Explicit broadcast journal path overrides continue to take precedence.

* **Refactor**
  * Centralized configuration-directory resolution for more consistent behavior across the CLI and shared components.

PR #1171, opened 2026-07-11
fix(storage): honor VULTISIG_CONFIG_DIR in default node storage

### LINKED ISSUE #1170: default Node/Electron FileStorage ignores VULTISIG_CONFIG_DIR
## Summary
The SDK's default Node/Electron `FileStorage` ignored `VULTISIG_CONFIG_DIR`, so default vault storage could split from the CLI config / credential / broadcast-journal paths even in the same runtime.

## Evidence
Pre-fix:
- `packages/sdk/src/platforms/node/storage.ts:16-18`
- `packages/sdk/src/platforms/node/index.ts:92-93`
- `packages/sdk/src/platforms/electron-main/index.ts:92-93`

The default constructor used `~/.vultisig` directly whenever `basePath` was omitted.

## Impact
A documented config-dir override could point different SDK surfaces at different roots. CLI config/credentials/journal would honor `VULTISIG_CONFIG_DIR`, while default SDK vault storage would not.

## Fix sketch
- trim and honor `VULTISIG_CONFIG_DIR` in the default `FileStorage` constructor
- keep blank/whitespace overrides falling back to `~/.vultisig`
- preserve explicit `basePath` precedence
- add regression coverage

## Effort
Small.

## Report
Round-10 report mirror: https://gist.github.com/gomesalexandre/6c48a5a29615a5064fe0a8572d846774

gomesalexandre:
## Summary
- honor `VULTISIG_CONFIG_DIR` in the SDK's default Node/Electron `FileStorage` path
- add regression coverage for env override, blank override fallback, and explicit `basePath` precedence
- add a changeset for the default-storage behavior fix

## Test plan
- [x] `yarn lint`
- [x] `yarn build:sdk`
- [x] `yarn build:all`
- [x] `yarn test`
- [x] `node --input-type=module -e "process.env.VULTISIG_CONFIG_DIR='/tmp/vultisig-config-r10'; const { FileStorage } = await import('./packages/sdk/dist/index.node.esm.js'); const storage = new FileStorage(); console.log(storage.basePath)"`

## Receipts
```text
/tmp/vultisig-config-r10
```

## Context
- round-10 report mirror: https://gist.github.com/gomesalexandre/6c48a5a29615a5064fe0a8572d846774

Closes https://github.com/vultisig/vultisig-sdk/issues/1170

## Summary by CodeRabbit

* **New Features**
  * Node and Electron vault storage now respects the `VULTISIG_CONFIG_DIR` environment variable.
  * Explicit storage directory settings continue to take precedence.
  * Storage falls back to the default `.vultisig` directory when the environment variable is unset or blank.

* **Bug Fixes**
  * Improved alignment of SDK vault storage with CLI configuration and related files.


packages/sdk/src/platforms/node/storage.ts
```diff
@@ -9,12 +9,17 @@ import * as path from 'path'
...
+  return override ? override : path.join(os.homedir(), '.vultisig')
+}
+
 export class FileStorage implements Storage {
   public readonly basePath: string
   private initPromise?: Promise<void>
 
   constructor(config?: { basePath?: string }) {
-    this.basePath = config?.basePath ?? path.join(os.homedir(), '.vultisig')
+    this.basePath = config?.basePath ?? getDefaultBasePath()
```
NeOMakinG, line 22: `preferably-blocking:` the FileStorage() default-path change is a public-constructor storage migration, not just a 2-site default-wiring tweak — env-set consumers silently stop reading pre-existing ~/.vultisig vaults — `FileStorage` is a PUBLIC export from both node/index.ts + electron-main/index.ts (also aliased `NodeStorage` / `Storage`) and SdkContextBuilder's error string literally suggests `new FileStorage()`. so the blast radius is not the two internal `configureDefaultStorage(() => new FileStorage())` sites — it is every external node/electron consumer that does `new FileStorage()` with no basePath. before this PR that default was ALWAYS `~/.vultisig` (env ignored); after, it is `$VULTISIG_CONFIG_DIR`. any such consumer that has the env set AND existing keyshares under `~/.vultisig` will, post-upgrade, read an empty/new dir and show "no vaults" — keyshares aren't destroyed (still on disk at ~/.vultisig) but the app can't see them. in-repo blast radius is ~zero because the CLI always passes explicit `basePath` via `createVaultStorage()`, but this ships as a published SDK where external consumers exist. the changeset reads as a co-location wiring tweak and carries no migration note.

fix: at minimum, make the changeset explicit that `new FileStorage()`'s default read/write dir moves for env-set consumers (constructor-level migration). ideally add a one-time fallback: if the resolved dir has no vault and legacy `~/.vultisig` does, read/warn there rather than silently presenting an empty vault list.

packages/client-shared/src/config-dir.ts
```diff
@@ -0,0 +1,7 @@
+import { homedir } from 'node:os'
+import { join } from 'node:path'
+
+export function getVultisigConfigDir(): string {
+  const override = process.env.VULTISIG_CONFIG_DIR?.trim()
```
NeOMakinG, line 5: `should-fix:` trim() silently migrates a space-padded VULTISIG_CONFIG_DIR value across all config surfaces (same as the shared-helper part of #1168) — inherited from the #1168 base: the helper trims any non-empty value, so `VULTISIG_CONFIG_DIR=' /vaults '` now resolves credentials.enc / config.json / token cache / journal under `/vaults` instead of the old literal spaced dir. low-probability but it silently relocates the credential + vault-registry path and is undocumented/uncovered. the SDK storage.ts inline `getDefaultBasePath()` trims too, so it inherits the same relocation.

fix: call out the trim relocation in a changeset, or pin it with a test so it's intentional-and-covered.

packages/sdk/src/platforms/node/storage.ts
```diff
@@ -9,12 +9,17 @@ import * as path from 'path'
 import type { Storage, StorageMetadata, StoredValue } from '../../storage/types'
 import { STORAGE_VERSION, StorageError, StorageErrorCode } from '../../storage/types'
 
+function getDefaultBasePath(): string {
```
NeOMakinG, line 12: `suggestion:` storage.ts getDefaultBasePath() is an inline duplicate of client-shared getVultisigConfigDir() — byte-identical today, divergence trap tomorrow — the two resolvers must stay in lockstep (CLI paths vs SDK default paths co-locate only because both do `process.env.VULTISIG_CONFIG_DIR?.trim() → override ? override : join(homedir, '.vultisig')`). they can't be shared because @vultisig/sdk can't depend on @vultisig/client-shared (circular). fine, but if one later grows validation/normalization and the other doesn't, the CLI and SDK-default vault dirs silently diverge again — the exact class of bug this stack is fixing.

fix: add a short comment on both bodies noting they are a deliberate duplicate that MUST be kept in sync, and/or a cross-package test asserting `getDefaultBasePath()` === `getVultisigConfigDir()` for a shared fixture.

NeOMakinG (commented):

## preferably-blocking
### 1. the FileStorage() default-path change is a public-constructor storage migration, not just a 2-site default-wiring tweak — env-set consumers silently stop reading pre-existing ~/.vultisig vaults — `packages/sdk/src/platforms/node/storage.ts:22`
`FileStorage` is a PUBLIC export from both node/index.ts + electron-main/index.ts (also aliased `NodeStorage` / `Storage`) and SdkContextBuilder's error string literally suggests `new FileStorage()`. so the blast radius is not the two internal `configureDefaultStorage(() => new FileStorage())` sites — it is every external node/electron consumer that does `new FileStorage()` with no basePath. before this PR that default was ALWAYS `~/.vultisig` (env ignored); after, it is `$VULTISIG_CONFIG_DIR`. any such consumer that has the env set AND existing keyshares under `~/.vultisig` will, post-upgrade, read an empty/new dir and show "no vaults" — keyshares aren't destroyed (still on disk at ~/.vultisig) but the app can't see them. in-repo blast radius is ~zero because the CLI always passes explicit `basePath` via `createVaultStorage()`, but this ships as a published SDK where external consumers exist. the changeset reads as a co-location wiring tweak and carries no migration note.
action: at minimum, make the changeset explicit that `new FileStorage()`'s default read/write dir moves for env-set consumers (constructor-level migration). ideally add a one-time fallback: if the resolved dir has no vault and legacy `~/.vultisig` does, read/warn there rather than silently presenting an empty vault list.

## should-fix
### 1. trim() silently migrates a space-padded VULTISIG_CONFIG_DIR value across all config surfaces (same as the shared-helper part of #1168) — `packages/client-shared/src/config-dir.ts:5`
inherited from the #1168 base: the helper trims any non-empty value, so `VULTISIG_CONFIG_DIR=' /vaults '` now resolves credentials.enc / config.json / token cache / journal under `/vaults` instead of the old literal spaced dir. low-probability but it silently relocates the credential + vault-registry path and is undocumented/uncovered. the SDK storage.ts inline `getDefaultBasePath()` trims too, so it inherits the same relocation.
action: call out the trim relocation in a changeset, or pin it with a test so it's intentional-and-covered.

## suggestion
### 1. storage.ts getDefaultBasePath() is an inline duplicate of client-shared getVultisigConfigDir() — byte-identical today, divergence trap tomorrow — `packages/sdk/src/platforms/node/storage.ts:12`
the two resolvers must stay in lockstep (CLI paths vs SDK default paths co-locate only because both do `process.env.VULTISIG_CONFIG_DIR?.trim() → override ? override : join(homedir, '.vultisig')`). they can't be shared because @vultisig/sdk can't depend on @vultisig/client-shared (circular). fine, but if one later grows validation/normalization and the other doesn't, the CLI and SDK-default vault dirs silently diverge again — the exact class of bug this stack is fixing.
action: add a short comment on both bodies noting they are a deliberate duplicate that MUST be kept in sync, and/or a cross-package test asserting `getDefaultBasePath()` === `getVultisigConfigDir()` for a shared fixture.

### 2. one raw untrimmed VULTISIG_CONFIG_DIR reader left outside the shared resolver — `clients/cli/src/lib/config.ts:199`
`EnvVars.getConfigDir()` returns `process.env.VULTISIG_CONFIG_DIR` raw; inert (no consumers) but a latent trap that would resurrect the untrimmed bug if ever used for a path.
action: route through `getVultisigConfigDir()` or delete the dead accessor.

## q
- which SDK consumers actually rely on the `configureDefaultStorage(() => new FileStorage())` default (vs passing explicit storage)? confirming the real-world env-set + legacy-vault population is small would let this ship without a dual-read shim.

## risk
medium — the fund-safety surface here is where node/electron default storage READS keyshares. verified locally: #1171's storage test is non-inert (reverted the `?? getDefaultBasePath()` back to the hard-coded `~/.vultisig` default → the 'honors VULTISIG_CONFIG_DIR' test goes RED, 1 failed; restored → 3/3 green). also re-verified the shared #1168 config-dir tests non-inert (16/16, whitespace tests RED on trim-revert). the residual risk is not a code bug — it is the undocumented constructor-level relocation for env-set external consumers (preferably-blocking above). in-repo (CLI) is safe: `createVaultStorage()` passes explicit basePath, so CLI vault path was already env-honoring pre-#1171. draft-hold: not approving a draft.

## verdict
COMMENTED

rcoderdev (approved):

Issue #1166, opened 2026-07-11
config: VULTISIG_CONFIG_DIR resolution is duplicated across CLI/shared consumers and already drifted


gomesalexandre:
## Summary
`VULTISIG_CONFIG_DIR` resolution was duplicated across CLI/shared consumers and had already drifted on empty/whitespace handling. One CLI path treated `''` / `'   '` as unset, while adjacent config/credential/token-cache/journal paths still had their own copies.

## Evidence
- `clients/cli/src/lib/config.ts:44-49` canonicalized empty/whitespace handling
- stale copies remained in:
  - `packages/client-shared/src/credential-store.ts:15-17,89-91`
  - `packages/client-shared/src/config-store.ts:8-15`
  - `clients/cli/src/agent/session.ts:1327-1329`
  - `clients/cli/src/agent/broadcastJournal.ts:202-209`

## Impact
- config registry, credentials, token cache, journal, and vault storage can disagree about where state lives
- the same class of bug can recur anywhere a new consumer copies the rule instead of importing it

## Fix sketch
Extract one shared `getVultisigConfigDir()` helper (trim, treat empty/whitespace as unset, default to `~/.vultisig`) and route all CLI/shared stateful surfaces through it. Add co-location tests.

## Effort
Small

## Report
https://gist.github.com/gomesalexandre/fac3c1408ccefc7e90a16955b11e855b

PR #1174, opened 2026-07-11
fix(cli): honor SDK vault storage in completion

### LINKED ISSUE #1173: cli: shell completion still mirrors a legacy vaults/ layout instead of SDK storage + VULTISIG_CONFIG_DIR
## Summary
CLI shell completion still mirrors a legacy `~/.vultisig/vaults` layout instead of the SDK's real root-level `vault:*.json` storage and `VULTISIG_CONFIG_DIR` policy.

## Evidence
- `clients/cli/src/lib/completion.ts:91-106` read vault names from a hardcoded local filesystem layout instead of the SDK's canonical config-dir helper / storage contract.
- `packages/sdk/src/platforms/node/storage.ts:49-54` stores non-cache keys directly under the configured base path, not under a `vaults/` subdirectory.
- `clients/cli/src/lib/completion.test.ts:64-82` now captures the correct behavior with a regression test rooted at `VULTISIG_CONFIG_DIR`.

## Impact
`vultisig switch` / `verify` completion could silently return no vault suggestions for normal SDK-managed vaults, especially when users redirect storage with `VULTISIG_CONFIG_DIR`.

## Fix sketch
Route completion through `getVultisigConfigDir()` and read the actual SDK storage shape (`vault:*.json` at the config-dir root). Keep backward tolerance for legacy top-level payloads if desired.

## Effort
Small.

## Report
Round 11 report: https://gist.github.com/gomesalexandre/b3a18f7e1536b2d389116d312b666fbe

gomesalexandre:
## Summary
- fix CLI shell completion to read the SDK's real root-level `vault:*.json` storage layout
- honor `VULTISIG_CONFIG_DIR` instead of mirroring a legacy `~/.vultisig/vaults` subtree
- add a regression test for completion suggestions rooted at `VULTISIG_CONFIG_DIR`

## Test plan
- [x] `yarn vitest run --config .config/vitest.cli.config.ts clients/cli/src/lib/completion.test.ts`
- [x] `yarn workspace @vultisig/cli test`
- [x] `yarn workspace @vultisig/cli typecheck`
- [x] `yarn build:sdk`
- [x] `yarn cli:build`
- [x] `yarn cli:smoke`

## Receipts
- Targeted completion regression:
  - `✓ clients/cli/src/lib/completion.test.ts (2 tests)`
- CLI workspace tests:
  - `695 passed`
- Builds / smoke:
  - `yarn build:sdk` ✅
  - `yarn cli:build` ✅
  - `yarn cli:smoke` ✅

## Links
- Closes #1173
- Round 11 report: https://gist.github.com/gomesalexandre/b3a18f7e1536b2d389116d312b666fbe

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CLI shell completion so vault suggestions use the correct shared storage location.
  * Vault completion now respects the `VULTISIG_CONFIG_DIR` environment setting.
  * Improved compatibility when reading vault data stored in supported JSON formats.


packages/sdk/src/platforms/node/storage.ts
```diff
@@ -9,12 +9,17 @@ import * as path from 'path'
...
+  return override ? override : path.join(os.homedir(), '.vultisig')
+}
+
 export class FileStorage implements Storage {
   public readonly basePath: string
   private initPromise?: Promise<void>
 
   constructor(config?: { basePath?: string }) {
-    this.basePath = config?.basePath ?? path.join(os.homedir(), '.vultisig')
+    this.basePath = config?.basePath ?? getDefaultBasePath()
```
NeOMakinG, line 22: `preferably-blocking:` carry-forward from #1171: FileStorage() no-arg default path now honors env — silent storage relocation for external node/electron consumers — `FileStorage` is a PUBLIC export (node/index.ts + electron aliased Storage/NodeStorage) and SdkContextBuilder's error string literally suggests `new FileStorage()`. Pre-stack the no-arg default was ALWAYS `~/.vultisig` (env ignored); post-stack it is `$VULTISIG_CONFIG_DIR`. An external consumer that sets the env AND has existing keyshares under `~/.vultisig` will, after upgrade, read an empty new dir and show "no vaults" (keyshares not destroyed, just invisible). This is the same finding I flagged on #1171 @d0ed86bb; it rides into #1174 because this PR bundles the #1171 commit. In-repo blast is ~zero (CLI always passes explicit basePath), so this is bounded to default-storage consumers.

fix: Add a migration note to the changeset (constructor-level relocation), and ideally a one-time dual-read fallback: if the new dir has no vault:* files and legacy `~/.vultisig` does, read/warn there. If #1171 lands separately first, this is already tracked there — just don't widen it here.

NeOMakinG (commented):
gm — multi-lane pass: claude (load-bearing) + codex gpt-5.3-codex (infra-dead, 0-byte). conceptual-only (no runtime this pass). comment-only.

## preferably-blocking
### 1. carry-forward from #1171: FileStorage() no-arg default path now honors env — silent storage relocation for external node/electron consumers — `packages/sdk/src/platforms/node/storage.ts:22`
`FileStorage` is a PUBLIC export (node/index.ts + electron aliased Storage/NodeStorage) and SdkContextBuilder's error string literally suggests `new FileStorage()`. Pre-stack the no-arg default was ALWAYS `~/.vultisig` (env ignored); post-stack it is `$VULTISIG_CONFIG_DIR`. An external consumer that sets the env AND has existing keyshares under `~/.vultisig` will, after upgrade, read an empty new dir and show "no vaults" (keyshares not destroyed, just invisible). This is the same finding I flagged on #1171 @d0ed86bb; it rides into #1174 because this PR bundles the #1171 commit. In-repo blast is ~zero (CLI always passes explicit basePath), so this is bounded to default-storage consumers.
action: Add a migration note to the changeset (constructor-level relocation), and ideally a one-time dual-read fallback: if the new dir has no vault:* files and legacy `~/.vultisig` does, read/warn there. If #1171 lands separately first, this is already tracked there — just don't widen it here.

## should-fix
### 1. carry-forward from #1168: .trim() migrates a space-padded VULTISIG_CONFIG_DIR value (not just whitespace-only) across every config path — `packages/client-shared/src/config-dir.ts:5`
`' /vaults '` resolved to the untrimmed `' /vaults '` under the old `||`/`??` sites (e.g. broadcastJournal returned the raw env value); now it trims to `/vaults`. That relocates credentials.enc / config.json / agent-tokens.json / broadcasts.jsonl for anyone with a space-padded override. Ultra-low probability, but a second silent relocation that is undocumented. Same finding as #1168 @155c7a46.
action: Note the trim-normalization as an intentional behavior change in the changeset (the whitespace-only fallback is the fix per #1166; the space-padded trim is the incidental side effect).

## suggestion
### 1. completion getVaultNames() now does real synchronous keyshare-file I/O on each `switch`/`verify` TAB — `clients/cli/src/lib/completion.ts:93`
Old code read a nonexistent `~/.vultisig/vaults` subtree → existsSync false → zero I/O. New code reads the real config-dir root, so every `switch <TAB>` / `verify <TAB>` now readFileSync + JSON.parse EVERY `vault:*.json` — and those files carry the full VaultData (keyshare) payload, so this reads the whole keyshare into memory just to pull name/id. Bounded by vault count and fully try/catch fail-safe (never crashes/hangs the shell), but it is a new interactive-latency cost that scales with keyshare size × vault count.
action: Acceptable as-is; if latency shows up, cap the read (e.g. read only the first N bytes / parse lazily) or memoize per completion process. Non-blocking.

### 2. completion now surfaces real vault names AND ids into shell completion output + history (first time it actually works) — `clients/cli/src/lib/completion.ts`
Because the old path was broken (wrong dir), completion previously emitted nothing; it now emits both the human name and the raw uuid for every vault on `switch`/`verify`. This is the feature's intent, but on a shared terminal it writes vault names/ids to the completion channel (and potentially shell history). Worth a one-liner in the changeset that vault identifiers become tab-completable.
action: No code change required; just be aware this is a behavior-visible change vs the (broken) prior state.

### 3. carry-forward from #1171: storage.ts getDefaultBasePath is a byte-identical duplicate of the shared getVultisigConfigDir — `packages/sdk/src/platforms/node/storage.ts:12`
The SDK cannot import @vultisig/client-shared (client-shared depends on the SDK, not vice-versa), so the inline duplicate is structural. Still a divergence trap: if the shared helper's resolution rule changes, this copy silently drifts.
action: Add a `// keep in sync with @vultisig/client-shared getVultisigConfigDir` comment, or a cross-package equality test asserting both resolve identically.

## q
- `parsed?.value ?? parsed` unwrap — is the `?? parsed` fallback intended to keep reading legacy flat (pre-StoredValue) vault files, and can any current SDK-written vault legitimately have `value` present but no `name`/`id`?

## risk
Low, and the NEW #1174 delta (completion) is fund-neutral: read-only, never signs/writes/decrypts. Verified against SDK layout — FileStorage.getFilePath stores vaults at `basePath/vault:<id>.json` (`:` not sanitized, only `/`\` are) wrapped as StoredValue `{value,metadata}`; old completion read a nonexistent `~/.vultisig/vaults` subdir with a flat `vault.name` shape → always empty, so this is a genuine fix. Fail-safe confirmed: getVaultNames is fully try/catch wrapped (outer + per-file inner) with an existsSync guard → never crashes/hangs the shell; getVultisigConfigDir never throws. SDK temp files end in `.tmp` and completion requires a `.json` suffix, so no mid-write partial read. completion.ts imports only @vultisig/client-shared (not @vultisig/sdk), so it is decoupled from the #1171 storage change. Ran the new completion.test.ts in a worktree (build:shared + vitest.cli): the vault-completion test PASSES at HEAD and goes RED when the fix is reverted (logMock never called) — non-inert, load-bearing. The one failing sibling (`--install` non-TTY) is a pre-existing spawn-based test failing only because the CLI dist wasn't built in the worktree (OOM), not a real regression. The only residual concerns are the two carry-forward relocation findings from #1168/#1171 (already flagged on their own PRs) and completion perf/info-visibility. Draft — COMMENTED, not approved.

## verdict
COMMENTED

packages/client-shared/src/config-dir.ts
```diff
@@ -0,0 +1,7 @@
+import { homedir } from 'node:os'
+import { join } from 'node:path'
+
+export function getVultisigConfigDir(): string {
+  const override = process.env.VULTISIG_CONFIG_DIR?.trim()
```
NeOMakinG, line 5: `should-fix:` carry-forward from #1168: .trim() migrates a space-padded VULTISIG_CONFIG_DIR value (not just whitespace-only) across every config path — `' /vaults '` resolved to the untrimmed `' /vaults '` under the old `||`/`??` sites (e.g. broadcastJournal returned the raw env value); now it trims to `/vaults`. That relocates credentials.enc / config.json / agent-tokens.json / broadcasts.jsonl for anyone with a space-padded override. Ultra-low probability, but a second silent relocation that is undocumented. Same finding as #1168 @155c7a46.

fix: Note the trim-normalization as an intentional behavior change in the changeset (the whitespace-only fallback is the fix per #1166; the space-padded trim is the incidental side effect).

clients/cli/src/lib/completion.ts
```diff
@@ -90,20 +90,20 @@ const CHAINS = [
  */
 function getVaultNames(): string[] {
   try {
-    const vaultDir = join(homedir(), '.vultisig', 'vaults')
+    const vaultDir = getVultisigConfigDir()
```
NeOMakinG, line 59: `suggestion:` completion getVaultNames() now does real synchronous keyshare-file I/O on each `switch`/`verify` TAB — Old code read a nonexistent `~/.vultisig/vaults` subtree → existsSync false → zero I/O. New code reads the real config-dir root, so every `switch <TAB>` / `verify <TAB>` now readFileSync + JSON.parse EVERY `vault:*.json` — and those files carry the full VaultData (keyshare) payload, so this reads the whole keyshare into memory just to pull name/id. Bounded by vault count and fully try/catch fail-safe (never crashes/hangs the shell), but it is a new interactive-latency cost that scales with keyshare size × vault count.

fix: Acceptable as-is; if latency shows up, cap the read (e.g. read only the first N bytes / parse lazily) or memoize per completion process. Non-blocking.

packages/sdk/src/platforms/node/storage.ts
```diff
@@ -9,12 +9,17 @@ import * as path from 'path'
 import type { Storage, StorageMetadata, StoredValue } from '../../storage/types'
 import { STORAGE_VERSION, StorageError, StorageErrorCode } from '../../storage/types'
 
+function getDefaultBasePath(): string {
```
NeOMakinG, line 12: `suggestion:` carry-forward from #1171: storage.ts getDefaultBasePath is a byte-identical duplicate of the shared getVultisigConfigDir — The SDK cannot import @vultisig/client-shared (client-shared depends on the SDK, not vice-versa), so the inline duplicate is structural. Still a divergence trap: if the shared helper's resolution rule changes, this copy silently drifts.

fix: Add a `// keep in sync with @vultisig/client-shared getVultisigConfigDir` comment, or a cross-package equality test asserting both resolve identically.

clients/cli/src/lib/completion.ts
```diff
@@ -90,20 +90,20 @@ const CHAINS = [
...
 
     for (const file of files) {
       if (file.startsWith('vault:') && file.endsWith('.json')) {
         try {
           const content = readFileSync(join(vaultDir, file), 'utf-8')
-          const vault = JSON.parse(content)
-          if (vault.name) names.push(vault.name)
-          if (vault.id) names.push(vault.id)
+          const parsed = JSON.parse(content)
+          const vault = parsed?.value ?? parsed
```
NeOMakinG, line 70: `q:` `parsed?.value ?? parsed` unwrap — is the `?? parsed` fallback intended to keep reading legacy flat (pre-StoredValue) vault files, and can any current SDK-written vault legitimately have `value` present but no `name`/`id`? — SDK FileStorage writes `{ value: VaultData, metadata }`, so `.value` is correct. The `?? parsed` branch only triggers for a flat file with no `value` key. Confirming the fallback is deliberate (legacy compat) vs dead — and that VaultData always carries name/id so no real vault silently drops out of completion.

rcoderdev (approved):

PR #1156, opened 2026-07-10
fix(cli): route agent chain aliases through normalizeChain

gomesalexandre:
## Summary
- route CLI agent chain resolution through the SDK's canonical `normalizeChain()` helper
- delete the stale local alias table from `clients/cli/src/agent/executor.ts`
- add handler-level tests proving aliases like `gaia` and `bnb smart chain` work through real `vault_chain` / `vault_coin` flows

## Problem
Round 5 centralized more alias tolerance in `packages/sdk/src/utils/normalizeChain.ts`, but the CLI agent executor still normalized chain strings with its own narrower local table before calling SDK APIs. That left a first-party drift bug where string forms already accepted by the SDK could still fail in real mutation flows like `vault_chain`, `vault_coin`, or `address_book`.

## Fix
`resolveChain()` now calls exported `normalizeChain()` and preserves the executor's existing `null`-on-unknown semantics.

## Test plan
- [x] `yarn workspace @vultisig/cli test src/agent/__tests__/executor.test.ts`
- [x] `yarn cli:build`

## Receipts
```text
$ yarn workspace @vultisig/cli test src/agent/__tests__/executor.test.ts
✓ src/agent/__tests__/executor.test.ts (25 tests)
Test Files 1 passed
Tests 25 passed

$ yarn cli:build
Built @vultisig/cli v2.19.11
```

## Links
- Closes #1155
- Round 6 report: https://gist.github.com/gomesalexandre/d089c11ec3700c0e391c683b1c57c083
- Campaign index artifact: https://claude.ai/code/artifact/be2deadf-6195-4dfa-a910-b269b9af7a84

PR #1188, opened 2026-07-11
fix(cli): use sdk evm chain-id helpers

### LINKED ISSUE #1185: cli: chain_id resolver still mirrors a stale subset instead of the SDK’s canonical EVM map
## Summary
`clients/cli` still owns a stale local `chain_id -> Chain` resolver even though the SDK already has the canonical EVM chain-id table internally.

Round report: https://gist.github.com/gomesalexandre/b7d7ce98b2283809dd70b192a1fca957
Campaign index: https://claude.ai/code/artifact/be2deadf-6195-4dfa-a910-b269b9af7a84

## Evidence
- `clients/cli/src/agent/executor.ts:2191-2207` hardcodes a local numeric `chainIdMap` with only 10 chains.
- The same file already treats `Mantle`, `Hyperliquid`, and `Sei` as first-class EVM chains at `clients/cli/src/agent/executor.ts:82-96`.
- The canonical mapping already exists in `packages/core/chain/chains/evm/chainInfo.ts:62-105` via `getEvmChainId()` / `getEvmChainByChainId()`.

## Impact
Server-built or tool-built payloads that arrive with only `chain_id` can fail to resolve supported EVM chains like Mantle / Hyperliquid / Sei, pushing the executor onto a fallback chain path.

Numeric chain-id lookup is chain metadata, not CLI policy. The SDK already owns the map; it should expose and consume it instead of forcing the CLI to keep a partial mirror.

## Fix sketch
- Re-export `getEvmChainId` / `getEvmChainByChainId` from the root `@vultisig/sdk` surface.
- Update `clients/cli/src/agent/executor.ts` to resolve numeric chain IDs through the SDK helper instead of a local table.
- Add regression coverage for Mantle / Hyperliquid / Sei.

## Effort
Low. Pure export + pure resolver swap + tests.

gomesalexandre:
## Summary
- export `getEvmChainId` and `getEvmChainByChainId` from the root `@vultisig/sdk` surface
- switch the CLI executor's `resolveChainId()` path from a stale local decimal map to the SDK's canonical EVM chain-id helpers
- add regression coverage for numeric `chain_id` resolution on Mantle, Hyperliquid, and Sei plus a root-export regression test
- add a changeset for `@vultisig/sdk` and `@vultisig/cli`

## Issue
Closes #1185

## Round report
- Gist: https://gist.github.com/gomesalexandre/b7d7ce98b2283809dd70b192a1fca957

## Verification receipts
- `yarn vitest run --config .config/vitest.cli.config.ts clients/cli/src/agent/__tests__/executor.test.ts`
  - PASS — `✓ clients/cli/src/agent/__tests__/executor.test.ts (26 tests)`
  - PASS — `✓ tests/unit/utils/publicExports.test.ts (14 tests)`
- `yarn build:sdk`
  - PASS — built SDK bundles and declarations, ending with `created dist/vite/index.d.ts in 340ms`
- `yarn cli:build`
  - PASS — `Built @vultisig/cli v2.19.14`
- `yarn node --input-type=module -e "import * as sdk from '@vultisig/sdk'; console.log(JSON.stringify({ mantleChainId: sdk.getEvmChainId(sdk.Chain.Mantle), hyperliquidFromRoot: sdk.getEvmChainByChainId('0x3e7'), hasHelpers: typeof sdk.getEvmChainId === 'function' && typeof sdk.getEvmChainByChainId === 'function' }))"`
  - PASS — `{"mantleChainId":"0x1388","hyperliquidFromRoot":"Hyperliquid","hasHelpers":true}`

## Notes
- `yarn build:sdk` emitted existing non-fatal Rollup warnings during bundle generation, but exited 0.
- The runtime proof exercised the built root `@vultisig/sdk` surface rather than a source-file import.

## Summary by CodeRabbit

* **New Features**
  * Added public SDK helpers for converting between EVM chain names and chain IDs.
  * Improved CLI handling of numeric chain IDs, including newer chains such as Mantle, Hyperliquid, and Sei.

* **Bug Fixes**
  * Corrected chain resolution when numeric identifiers are provided instead of chain names.

* **Tests**
  * Added coverage for chain ID resolution and SDK exports.


clients/cli/src/agent/executor.ts
```diff
@@ -2189,22 +2189,12 @@ export function parseThorSwapMemo(memo: string): ParsedThorSwapMemo {
...
-    42161: Chain.Arbitrum,
-    10: Chain.Optimism,
-    8453: Chain.Base,
-    81457: Chain.Blast,
-    324: Chain.Zksync,
-    25: Chain.CronosChain,
-  }
-  return chainIdMap[id] || null
+  const normalized = typeof chainId === 'string' ? chainId.trim().toLowerCase() : String(chainId)
+  const id = Number.parseInt(normalized, normalized.startsWith('0x') ? 16 : 10)
```
NeOMakinG: `should-fix:` resolveChainId accepts trailing-garbage decimal ids and binds them to a real chain — Number.parseInt('5000garbage', 10) === 5000 → canonicalChainId '0x1388' → Mantle; likewise '1329foo'→Sei, '999junk'→Hyperliquid, '1e3'→Ethereum. This lax parse is pre-existing for decimal ('1garbage'→Ethereum was already true), so it is not a NET-new vuln class, and downstream agreement guards (resolvePrepChain/resolvePrepParentChain reject a present-and-disagreeing chain-vs-chain_id pair, and resolveStrictEvmChain fails closed), plus an attacker who controls chain_id would just supply clean '5000' — which is why this is should-fix, not a blocker (Codex rated blocker; downgraded). But this path now silently binds newly-mapped chains from malformed identifiers where the old subset returned null → fell to name/default. Note the hex path already fails closed on garbage ('0x1388zzz'→null, '0x0001388'→null) because it uses the verbatim string as the map key; only the decimal path reconstructs from the parsed int and thus tolerates trailing junk.

fix: Reject partial parses before lookup: `if (!/^(0x[0-9a-f]+|[0-9]+)$/.test(normalized)) return null` (after trim+lowercase). Makes the resolver strictly stronger than the old map for every input while keeping the intended decimal+hex support.

clients/cli/src/agent/__tests__/executor.test.ts
```diff
@@ -203,6 +203,16 @@ describe('AgentExecutor — per-tool handlers (new tool-path API)', () => {
...
 })
 
+describe('resolveChainId', () => {
+  it.each([
+    ['5000', Chain.Mantle],
+    ['999', Chain.Hyperliquid],
+    ['1329', Chain.Sei],
+  ])('resolves numeric chain_id %s via the canonical SDK EVM map', (chainId, expected) => {
+    expect(resolveChainId(chainId)).toBe(expected)
+  })
```
NeOMakinG, line 213: `suggestion:` resolveChainId test only covers the 3 NEW chains — add a legacy-parity + hex-input + fail-closed case — The it.each covers 5000/999/1329 (Mantle/Hyperliquid/Sei) — the newly-mapped chains — so it is non-inert (RED on revert, since the old map lacked them). But it does not lock the behavior-preserving guarantee for the 10 retained chains, nor exercise the new hex-input path, nor the fail-closed contract that gates signing.

fix: Add rows: ['1', Chain.Ethereum] and ['0x1388', Chain.Mantle] (parity + hex path), plus an expect(resolveChainId('999999')).toBeNull() / resolveChainId('abc') to pin the fail-closed branch that the sign path depends on.

NeOMakinG (commented):

## should-fix
### 1. resolveChainId accepts trailing-garbage decimal ids and binds them to a real chain — `clients/cli/src/agent/executor.ts:2193`
Number.parseInt('5000garbage', 10) === 5000 → canonicalChainId '0x1388' → Mantle; likewise '1329foo'→Sei, '999junk'→Hyperliquid, '1e3'→Ethereum. This lax parse is pre-existing for decimal ('1garbage'→Ethereum was already true), so it is not a NET-new vuln class, and downstream agreement guards (resolvePrepChain/resolvePrepParentChain reject a present-and-disagreeing chain-vs-chain_id pair, and resolveStrictEvmChain fails closed), plus an attacker who controls chain_id would just supply clean '5000' — which is why this is should-fix, not a blocker (Codex rated blocker; downgraded). But this path now silently binds newly-mapped chains from malformed identifiers where the old subset returned null → fell to name/default. Note the hex path already fails closed on garbage ('0x1388zzz'→null, '0x0001388'→null) because it uses the verbatim string as the map key; only the decimal path reconstructs from the parsed int and thus tolerates trailing junk.
action: Reject partial parses before lookup: `if (!/^(0x[0-9a-f]+|[0-9]+)$/.test(normalized)) return null` (after trim+lowercase). Makes the resolver strictly stronger than the old map for every input while keeping the intended decimal+hex support.

## suggestion
### 1. resolveChainId test only covers the 3 NEW chains — add a legacy-parity + hex-input + fail-closed case — `clients/cli/src/agent/__tests__/executor.test.ts:213`
The it.each covers 5000/999/1329 (Mantle/Hyperliquid/Sei) — the newly-mapped chains — so it is non-inert (RED on revert, since the old map lacked them). But it does not lock the behavior-preserving guarantee for the 10 retained chains, nor exercise the new hex-input path, nor the fail-closed contract that gates signing.
action: Add rows: ['1', Chain.Ethereum] and ['0x1388', Chain.Mantle] (parity + hex path), plus an expect(resolveChainId('999999')).toBeNull() / resolveChainId('abc') to pin the fail-closed branch that the sign path depends on.

## risk
low-to-moderate for a fund-critical path. resolveChainId feeds EVM tx routing/signing, but the swap is provably behavior-preserving for the 10 previously-mapped chains and additive otherwise. I re-derived the numeric map from viem chain ids + numberToHex and ran an old-vs-new table across 27 inputs: all 10 legacy ids (1/56/137/43114/42161/10/8453/81457/324/25) resolve byte-identical (Ethereum/BSC/Polygon/Avalanche/Arbitrum/Optimism/Base/Blast/Zksync/CronosChain), and unknown/garbage-hex/leading-zero-hex/out-of-range still fail closed → null. New surface: Mantle(5000)/Hyperliquid(999)/Sei(1329) + the rest of the canonical 13-chain EVM map, plus hex-string input support (0x1388→Mantle) that the old decimal-only parseInt could never resolve. The `as Chain` cast is truthful — getEvmChainByChainId returns EvmChain, a subset of Chain, so every returned value is a valid Chain enum member. Merge-order gate is CLEAR (unlike #1156): getEvmChainId/getEvmChainByChainId already exist on origin/main in packages/core/chain/chains/evm/chainInfo.ts, and both the src/index.ts re-export and the CLI import ship in THIS PR — self-contained, no unmerged-base dependency. RN parity OK: src/index.ts is the single source for all platform bundles (incl. dist/index.react-native.js), chainInfo is already in the RN graph (react-native/overrides/lifiSwapEnabledChains.ts and tools/swap/skip/chainMapping.ts already import it) and documented Hermes-safe, so the re-export adds no native-dep hazard. Note: could not run the vitest suites in the worktree (core-config rootApiUrl transform error under symlinked node_modules — env, not a PR defect); relied on source-level parity + CI Integration Tests pass + author receipts.

## verdict
COMMENTED

rcoderdev (approved):

clients/cli/src/agent/executor.ts
```diff
@@ -2196,22 +2204,12 @@ export function parseThorSwapMemo(memo: string): ParsedThorSwapMemo {
...
+  if (Number.isNaN(id)) return null
+
+  const canonicalChainId = normalized.startsWith('0x') ? normalized : `0x${id.toString(16)}`
+  return (getEvmChainByChainId(canonicalChainId) as Chain | undefined) ?? null
```
NeOMakinG: `blocker:` superseded by #1308 — this PR conflicts with main and would regress chain-id resolution — #1308 ("Share chain reference resolution across SDK, CLI, and MCP") merged to main 2026-07-16, AFTER this branch's base (c3e73e74, 2026-07-15) and it is NOT in this PR's history. #1308 already rewrote `resolveChainId` to `resolveChainReference(chainId) ?? null`, and that shared resolver (packages/sdk/src/utils/resolveChainReference.ts) already: (a) routes numeric EVM ids through the SAME canonical `getEvmChainByChainId(numberToHex(id))` this PR introduces — so Mantle(5000)/Hyperliquid(999)/Sei(1329) already resolve on main; (b) is STRICTER — `/^[1-9][0-9]*$/` rejects the trailing-garbage/sci-notation/leading-zero inputs this PR still binds; (c) also resolves Cosmos chain-ids. Confirmed `git merge origin/main` on this HEAD → CONFLICT in executor.ts's resolveChainId. Merging as-is either fails, or (if forced this PR's way) un-shares the resolver, drops Cosmos id support, and re-opens the misroute #1185 targeted.

fix: Close #1188 as superseded by #1308 (the CLI executor goal is already delivered). If any residual value remains, it is ONLY the `packages/sdk/src/index.ts` root re-export of getEvmChainByChainId/getEvmChainId — which is #1192's deliverable and collides with it; coordinate there. Do not merge the executor.ts change.

packages/sdk/src/index.ts
```diff
@@ -184,6 +184,7 @@ export type { SdkEvents, VaultEvents } from './events/types'
 // Chain enums and types
 export type { Chain as ChainType, CosmosChain, EvmChain, OtherChain, UtxoChain } from './types'
 export { Chain } from './types'
+export { getEvmChainByChainId, getEvmChainId } from '@vultisig/core-chain/chains/evm/chainInfo'
```
NeOMakinG: `should-fix:` root re-export duplicates #1192 — double-export build break if both land — This line is byte-identical to the export #1192 adds (at a different region, :316). Neither is on main yet. Because the regions differ, git 3-way-merges both cleanly → the symbol is exported twice → esbuild/typecheck 'Multiple exports with the same name'. #1188's executor change is superseded (see blocking), so the root export is the only surviving purpose — and it is exactly #1192's job.

fix: Land the root export in ONE PR (#1192, which also mirrors it onto the RN entry). Drop it here / close #1188.

NeOMakinG (commented):

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | should-fix — resolveChainId accepts trailing-garbage / sci-notation decimal ids and binds them ('1e3'→Ethereum, '8453foo'→Base) instead of failing closed | ❌ still present AND now REGRESSIVE — main's #1308 resolveChainReference already gates numeric ids with /^[1-9][0-9]*$/ (rejects exactly these). #1188 would re-open the hole it was meant to close |
| 1 | suggestion — resolveChainId test covers only the 3 new chains (5000/999/1329); add legacy-parity + fail-closed cases | ⚠️ unchanged (still only 3 chains) and now moot: on main those 3 ids already resolve via #1308's shared resolveChainReference, so the test asserts behavior that already ships |

## blocking
### 1. superseded by #1308 — this PR conflicts with main and would regress chain-id resolution — `clients/cli/src/agent/executor.ts:2212`
#1308 ("Share chain reference resolution across SDK, CLI, and MCP") merged to main 2026-07-16, AFTER this branch's base (c3e73e74, 2026-07-15) and it is NOT in this PR's history. #1308 already rewrote `resolveChainId` to `resolveChainReference(chainId) ?? null`, and that shared resolver (packages/sdk/src/utils/resolveChainReference.ts) already: (a) routes numeric EVM ids through the SAME canonical `getEvmChainByChainId(numberToHex(id))` this PR introduces — so Mantle(5000)/Hyperliquid(999)/Sei(1329) already resolve on main; (b) is STRICTER — `/^[1-9][0-9]*$/` rejects the trailing-garbage/sci-notation/leading-zero inputs this PR still binds; (c) also resolves Cosmos chain-ids. Confirmed `git merge origin/main` on this HEAD → CONFLICT in executor.ts's resolveChainId. Merging as-is either fails, or (if forced this PR's way) un-shares the resolver, drops Cosmos id support, and re-opens the misroute #1185 targeted.
action: Close #1188 as superseded by #1308 (the CLI executor goal is already delivered). If any residual value remains, it is ONLY the `packages/sdk/src/index.ts` root re-export of getEvmChainByChainId/getEvmChainId — which is #1192's deliverable and collides with it; coordinate there. Do not merge the executor.ts change.

## should-fix
### 1. root re-export duplicates #1192 — double-export build break if both land — `packages/sdk/src/index.ts:187`
This line is byte-identical to the export #1192 adds (at a different region, :316). Neither is on main yet. Because the regions differ, git 3-way-merges both cleanly → the symbol is exported twice → esbuild/typecheck 'Multiple exports with the same name'. #1188's executor change is superseded (see blocking), so the root export is the only surviving purpose — and it is exactly #1192's job.
action: Land the root export in ONE PR (#1192, which also mirrors it onto the RN entry). Drop it here / close #1188.

## risk
FUND-CRITICAL surface (EVM chain-id → Chain resolution feeds tx routing). The blocking issue is not a latent bug in this diff per se — the diff read in isolation is fund-safe (10-chain byte-parity holds, unknown→null). The risk is integration: main moved under this PR (#1308), and merging the stale branch would REGRESS a stricter, shared, Cosmos-aware resolver back to a looser CLI-local one. Fail-closed on clean input; fail-OPEN (misroute) on malformed decimal — the exact gap #1308 already closed.

## verdict
COMMENTED

packages/sdk/src/index.ts
```diff
@@ -192,6 +192,7 @@ export type { SdkEvents, VaultEvents } from './events/types'
...
```
NeOMakinG: `preferably-blocking:` open PR #1192 adds this exact same named export to this exact same module, at `:307`. Different offsets means git will **not** conflict — it will merge cleanly into a file that declares `getEvmChainId`/`getEvmChainByChainId` twice, which is a TS/rollup build break that lands on `main` rather than in either PR's CI.

Since the merge resolution handed the CLI half of this PR over to main's `resolveChainReference`, this line is now the PR's *entire* remaining substance — and #1192 is a strict superset (it also exports the pair from the RN entry and pins an exhaustive `satisfies Record<EvmChain, number>` table).

Suggested fix: drop this line and the changeset here, let #1192 own the export, and close this PR. If you'd rather keep it here, #1192 needs its `index.ts` hunk dropped instead — either way, one canonical export site decided before either merges.

NeOMakinG: `should-fix:` this export isn't reachable from React Native. `packages/sdk/src/platforms/react-native/index.ts` is the only platform barrel that is a hand-curated allow-list instead of `export * from '../../index'` (browser `:63`, node `:118`, electron-main `:118`, chrome-extension `:67` all re-export the root). So this reaches node/browser/electron/chrome only.

`publicExports.test.ts` imports `../../../src/index` directly, so it asserts the root barrel and stays green with the RN gap shipped.

Suggested fix: mirror the export into the RN entry — which is what #1192 already does.

.changeset/sdk-1185-evm-chainid-resolver.md
```diff
@@ -0,0 +1,8 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Export the SDK's canonical EVM `getEvmChainId` and `getEvmChainByChainId` helpers from the root `@vultisig/sdk` surface, and switch the CLI executor's numeric `chain_id` resolver to use that canonical map instead of a stale local subset.
```
NeOMakinG: `should-fix:` this no longer describes the PR. After the `569b63e7` merge resolution, `resolveChainId()` in `clients/cli/src/agent/executor.ts` is verbatim main's `resolveChainReference(chainId) ?? null` — the CLI resolver switch and the Mantle/Hyperliquid/Sei fix both came from #1308, which has its own changeset. Publishing this ships duplicate, misattributed release notes for a fund-critical area, and the `'@vultisig/cli': patch` bump is for a package this PR no longer modifies.

Suggested fix: reduce to what actually ships — "Export the canonical `getEvmChainId` / `getEvmChainByChainId` helpers from the root `@vultisig/sdk` surface" — and drop the `@vultisig/cli` bump. (Moot if you take the close path on #1192.)

clients/cli/src/agent/__tests__/executor.test.ts
```diff
@@ -203,6 +203,16 @@ describe('AgentExecutor — per-tool handlers (new tool-path API)', () => {
   })
 })
 
+describe('resolveChainId', () => {
```
NeOMakinG, line 206: `should-fix:` this block contributes zero coverage. I deleted it wholesale and re-ran the CLI agent suite against a freshly built SDK: **560 passed / 36 files, nothing red.**

Main's `clients/cli/src/agent/__tests__/chainReferenceResolver.test.ts` (landed with #1308) already asserts `999`→Hyperliquid, `5000`→Mantle, `1329`→Sei — plus `8453`→Base and both `null` cases. This is a strict subset of a test three files away.

Suggested fix: delete it. If you want it to earn its place, assert the fail-closed contract the sign path leans on, which is currently unasserted anywhere — `resolveChainId('0005000')`, `resolveChainId('5000garbage')`, `resolveChainId('9007199254740993')` all → `null`. My matrix mutated the resolver to accept leading zeros (M2) and to drop the `Number.isSafeInteger` guard (M3) and both stayed **green**.

NeOMakinG (commented):
Round 3 on `569b63e7` (merge of `origin/main`, hand-resolved conflict in `clients/cli/src/agent/executor.ts`). The headline: **the merge resolution surrendered the PR's own change to main.** `resolveChainId()` is now verbatim main's `resolveChainReference(chainId) ?? null` — the hardcoded-map replacement this PR existed to ship is gone (delivered instead by #1308). What's left is a +27/-1 residual: one root re-export, one changeset, two test additions.

That's good news for my R2 blocker (it's resolved), but it leaves the PR describing something it no longer does, and reduces it to a strict subset of still-open #1192.

## testing approach

- Worktree `/tmp/wt-vultisig-sdk-4` @ `569b63e7`. `node_modules` symlinked from the canonical checkout; **the canonical `packages/sdk/dist` is stale** (built at `21b4840f`, zero occurrences of `resolveChainReference`), which made the CLI suite fail spuriously. Control: main's own pre-existing `chainReferenceResolver.test.ts` fails **identically** in that env → env artifact, not a PR defect. Rebuilt `packages/sdk` in-worktree and aliased `@vultisig/sdk` → the fresh `dist/index.node.esm.js`; **34/34 green**.
- Merge integrity: `merge-base(origin/main, 569b63e7) == origin/main == 358c27ba` (main is an ancestor, nothing reverted). Full net diff vs merge-base read line by line — 4 files, 27 insertions, **zero deletions of main content**, nothing smuggled. `48ad7a26` (#1308) confirmed in branch history.
- Live probe of `resolveChainReference` against the built artifact across 19 inputs (decimal, hex, trailing-garbage, sci-notation, leading-zero, unsafe-integer, names, aliases).
- 6-mutation matrix (below). CI at `head_sha=569b63e7` verified: Lint+TypeCheck pass, Integration pass, Unit pending, CodeRabbit "fail" is a rate-limit not a result.
- Second lane: Codex gpt-5.4 (exit 0) — converged on the changeset staleness, the RN reachability gap, the #1192 collision, and independently found the test redundancy.

### mutation matrix

| # | mutation | suite result | reading |
|---|---|---|---|
| M1 | delete the PR's `export { getEvmChainByChainId, getEvmChainId }` from `src/index.ts` | **RED** (1 failed) | the PR's one real change is non-inert |
| M2 | `/^[1-9][0-9]*$/` → `/^[0-9]+$/` (accept leading zeros) | GREEN | untested (main's code) |
| M3 | drop the `Number.isSafeInteger` guard | GREEN | untested (main's code) |
| M4 | drop the Cosmos-id fallback in `resolveChainIdReference` | GREEN (sdk suite) | sdk-side suite doesn't reach it |
| M5 | swap the `Mantle`↔`Hyperliquid` viem chain objects | **RED** (3 failed) | the chain-id mapping assertions bite |
| M6 | **delete the PR's entire `describe('resolveChainId')` block** | **GREEN — 560/560, 36 files** | see finding 3 |

## prior issues — status

| round | severity | issue | status | notes |
|---|---|---|---|---|
| R2 | blocker | #1188 superseded by #1308; `git merge origin/main` conflicts in `executor.ts resolveChainId` | ✅ fixed | `569b63e7` merged main and resolved the conflict **in main's favour** — the PR's resolver body is gone, `executor.ts` import block is now byte-identical to `origin/main`. Supersession is no longer a merge hazard; it is simply now the outcome. |
| R2 | should-fix | root re-export in `index.ts` byte-identical to open #1192 → double-export build break if both land | ❌ still open | #1192 still OPEN at `cd3fd506`, still adds the same line — now at `index.ts:307` vs this PR's `:195`. See finding 1. |
| R1 | should-fix | `resolveChainId` bound trailing-garbage decimals (`5000garbage`→Mantle, `1e3`→Ethereum) | ✅ fixed | fixed by main's stricter `/^[1-9][0-9]*$/`, inherited via the merge. Probed at HEAD: `5000garbage`, `1e3`, `01`, `0005000`, `9007199254740993` → all `undefined`. |
| R1 | suggestion | add legacy-parity / hex / fail-closed cases to the resolver test | ⚠️ partial | main's `chainReferenceResolver.test.ts` now covers `8453` + both null cases. The hex case is moot — hex no longer resolves at all (see "pre-existing"). |

## preferably-blocking

**1. Duplicate root export collides with open #1192 — and #1192 is a strict superset of what's left here.**

Both PRs add the *same named export* to `packages/sdk/src/index.ts`: this one at `:195`, #1192 at `:307`. Different offsets, so git will **not** conflict — it will merge cleanly into a module that declares `getEvmChainId`/`getEvmChainByChainId` twice, which is a TS/rollup build break, and it'll break on `main` after the second merge rather than in either PR's CI.

#1192 additionally exports the pair from `packages/sdk/src/platforms/react-native/index.ts` and adds an exhaustive `satisfies Record<EvmChain, number>` pin. Since the CLI half of #1188 is now gone, #1192 covers 100% of this PR's residual value plus more.

Suggested fix: drop `packages/sdk/src/index.ts:195` + the changeset from this PR and let #1192 own the export — at which point #1188 has nothing left and can be closed. If you'd rather keep #1188, then #1192 needs its `index.ts` hunk dropped instead. Either way one canonical export site, decided before either merges.

## should-fix

**2. The changeset now ships release notes for a change that isn't in this PR.**

`.changeset/sdk-1185-evm-chainid-resolver.md` still says it "switch[es] the CLI executor's numeric `chain_id` resolver to use that canonical map instead of a stale local subset" and "fixes numeric `chain_id` resolution for newer EVM chains such as Mantle, Hyperliquid, and Sei". After the merge resolution none of that is this PR — it's #1308, which has its own changeset. It also declares `'@vultisig/cli': patch` for a package this PR no longer modifies.

Suggested fix: rewrite to the truth — "Export the canonical `getEvmChainId` / `getEvmChainByChainId` helpers from the root `@vultisig/sdk` surface" — and drop the `@vultisig/cli` bump. (Moot if you take finding 1's close path.)

**3. The added CLI test contributes zero coverage — deleting it leaves the suite fully green.**

M6: removing the entire `describe('resolveChainId')` block from `executor.test.ts` → **560 passed / 36 files, nothing red.** Main's `clients/cli/src/agent/__tests__/chainReferenceResolver.test.ts` (from #1308) already asserts `999`→Hyperliquid, `5000`→Mantle, `1329`→Sei, plus `8453`→Base and both `null` cases. The new block is a strict subset of a test that already exists three files away.

Suggested fix: delete the block. If you want it to earn its place, assert something `chainReferenceResolver.test.ts` doesn't — the fail-closed contract the sign path leans on (`resolveChainId('0005000')`, `resolveChainId('5000garbage')`, `resolveChainId('9007199254740993')` → `null`), all of which are currently unasserted anywhere (M2/M3 stayed green).

**4. The new export is unreachable from React Native.**

`packages/sdk/src/platforms/react-native/index.ts` is the only platform barrel that is a hand-curated allow-list rather than `export * from '../../index'` (browser `:63`, node `:118`, electron-main `:118`, chrome-extension `:67` all re-export the root). So `src/index.ts:195` reaches node/browser/electron/chrome only. `publicExports.test.ts` imports `../../../src/index` directly, so it asserts the root barrel and would stay green with the RN gap shipped.

Suggested fix: mirror the export into the RN entry — which is exactly what #1192 already does, another reason to converge on it.

Not introduced here, but this PR's original diff briefly fixed it and the merge resolution un-fixed it, so it's worth recording: **hex `chain_id` no longer resolves, and the failure is silent.** Probed at HEAD against the built artifact — `0x1`, `0x2105`, `0x1388`, `0x3e7` all → `undefined`, because `resolveChainIdReference` gates on a decimal-only `/^[1-9][0-9]*$/` and `normalizeChain('0x2105')` then throws `UnknownChainError`. The pre-#1188 hardcoded map didn't accept hex either, so this is not a regression against `main` — but the call sites fail *open*, not closed:

- `executor.ts:1758` — `chain = resolveChainId(chainId) || Chain.Ethereum` on an EIP-712 `domain.chainId`. A hex domain chainId silently signs under `Ethereum`. Low blast radius (EVM chains share derivation), but the label is wrong.
- `executor.ts:1107` — `chain = resolveChainId(chainId) || defaultChain` on the prepare→sign→**broadcast** path. Here an unresolved chain id silently picks `defaultChain`, i.e. a Base envelope arriving with `chain_id: '0x2105'` and no `chain` name would be prepared and broadcast against the default chain.

The `toolOutputSigning.ts` guards (`resolveStrictEvmChain`, `resolvePrepChain`, `resolvePrepParentChain`) are correctly fail-closed on disagreement and are unaffected — they return `null` rather than defaulting, so the exposure is limited to the two `||` fallbacks above. Worth a follow-up: either accept `0x`-prefixed ids in `resolveChainIdReference`, or make those two sites throw instead of defaulting when a `chain_id` was present but unresolved.

## risk

Low for what actually ships (one additive re-export, proven non-inert by M1, with a mapping assertion that bites under M5). The real risk is coordination, not code: two open PRs racing to add the same export line to the same module, and a changelog entry that will tell consumers a fund-critical CLI resolver fix shipped in this patch when it shipped in #1308.

Cross-reference, out of scope for this diff but still live from the #1256 round: `evmChains` in `packages/sdk/src/utils/addressFormat.ts` still has no `'sei'` entry, so `checkChainPrefix` fails open with `unknown-chain` for Sei — worth folding into whichever PR ends up owning the chain-id surface.

## verdict

**COMMENT.** My R2 blocker is genuinely resolved and the merge is clean (main is an ancestor, zero deletions, nothing smuggled). Withholding approval only on finding 1: this PR's entire remaining substance is one line that open #1192 also adds, and merging both silently produces a duplicate export on `main`. Decide which PR owns the export — if it's #1192, this one can just be closed; if it's this one, drop the line from #1192 and fix the changeset, and I'll approve on the next pass.

---
🤖 automated review — two-lane (Claude load-bearing + Codex gpt-5.4). Findings verified against source at `569b63e7`.

NeOMakinG (approved):
Round 4 — approving. HEAD advanced to `c3eec5f6` by my own fix commit (fix-and-approve). The R3 preferably-blocking prediction **materialized on main and I fixed it in this branch**.

## testing approach
- Worktree `/tmp/wt-vultisig-sdk-1188` @ `159dad13`, sibling `packages/**/dist` symlinked from the canonical checkout, `packages/sdk/dist` freshly built in-worktree (the R3 stale-dist gotcha).
- **Merge integrity:** `merge-base(origin/main, PR-head) == origin/main == 5303d704` — main (now including **#1192**, `5303d704`) is an ancestor of PR head. So #1192's root re-export is physically in this branch's tree.
- **Build break reproduced (the R3 finding, now live):** `packages/sdk/src/index.ts` at `159dad13` declared `export { getEvmChainByChainId, getEvmChainId } from '@vultisig/core-chain/chains/evm/chainInfo'` **twice** — `:196` (this PR) and `:343` (#1192, merged to main). `tsc --noEmit` → 4× `TS2300 Duplicate identifier`; the CI Unit/Lint/Integration jobs at `159dad13` are all RED with esbuild `Multiple exports with the same name "getEvmChainByChainId"` at `index.ts:343`. Control on `origin/main` = only the 2 pre-existing cardano `auxiliaryData` errors, so the 4 TS2300 are PR-introduced.
- **Fix:** dropped the `:196` line (index.ts now byte-identical to `origin/main`) and removed the changeset. Post-fix `tsc` = only the 2 pre-existing cardano lines; `rollup` node build = exit 0; full SDK unit suite **2942/2942**; CLI `executor.test.ts` **26/26** (with `@vultisig/sdk` aliased to the fresh worktree dist, since canonical dist is stale — `grep -c resolveChainReference` = 0).
- **Non-inertness proven:** M1 delete the sole remaining `getEvmChain*` export → publicExports RED (`undefined` to be `function`); M2 swap Mantle↔Hyperliquid viem objects → RED (`0x3e7` vs `0x1388`); M-CLI force `resolveChainId`→`null` → all 3 CLI cases RED. Each mutant grep-verified present in the worktree file and absent from the canonical checkout.
- Codex lane (gpt-5.3-codex): **SKIPPED** — 0-byte output at exit 0 (dead-lane signal per standing policy), not "no findings".

## prior issues — status
|---|---|---|---|---|---|
| r3 | 1 | preferably-blocking | #1192 adds the SAME `getEvmChain*` root re-export → clean git merge → double-export build break lands on main | ✅ fixed | #1192 merged to main (`5303d704`); duplicate materialized at `index.ts:196`/`:343` and broke all platform builds. Dropped `:196` in `c3eec5f6`; `tsc`/`rollup`/2942 unit + 26 CLI all green. |
| r3 | 2 | should-fix | changeset ships FALSE release notes (claims CLI resolver switch + Mantle/HL/Sei fix, all now #1308's; wrong `@vultisig/cli` bump) | ✅ fixed | Both claims already on main; removed the changeset in `c3eec5f6`. PR is now test-only, no functional package change, so no changeset is warranted. |
| r3 | 3 | should-fix | RN barrel `platforms/react-native/index.ts` never got the export | ℹ️ n/a | #1192 added the RN entry on main; this PR no longer touches the root export at all, so there is nothing to mirror. |
| r3 | 4 | should-fix | PR's added CLI `resolveChainId` test is redundant — #1308's `chainReferenceResolver.test.ts` already asserts these | ℹ️ documented | Kept. It is the PR's only remaining content, is non-inert (M-CLI → RED), and harmlessly co-asserts 5000/999/1329 at the CLI boundary. Redundant, not wrong. |
| r2 | 1 | blocker | superseded by #1308 (resolver un-shared / Cosmos dropped on force-merge) | ✅ fixed | Resolved by capitulation — the r3 merge surrendered `executor.ts` to main's `resolveChainReference`; `executor.ts` at head is byte-identical to `origin/main`. No un-sharing. |
| r1 | 1 | should-fix | `resolveChainId` bound trailing-garbage decimal ids (`5000garbage`→Mantle, `1e3`→ETH) | ✅ fixed | Main's strict `/^[1-9][0-9]*$/` in `resolveChainReference` (#1308) rejects them; executor now calls it. |
| r1 | 2 | suggestion | CLI test covers only the 3 new chains | ℹ️ documented | Legacy-parity + fail-closed cases are covered by main's `chainReferenceResolver.test.ts`; acceptable. |

## blocking
None remaining — the sole build-breaking blocker is fixed in `c3eec5f6`.

## should-fix
None.

- **hex `chain_id` is dead and fails OPEN at two call sites** (unchanged from main, out of scope here): `resolveChainReference` is decimal-only, so `executor.ts:1107` `resolveChainId(chainId) || defaultChain` on the prepare→sign→broadcast path silently defaults a `chain_id:'0x2105'` (Base) envelope to the default chain rather than failing closed. Pre-exists this PR (executor.ts at head == main). Worth a follow-up bead, not a blocker for this test-only change.

## risk
Low. Net diff after my fix is **test-only** (`executor.test.ts` +11/-1, `publicExports.test.ts` +7) — `index.ts` and the changeset are back to main's state. No production code path changes; both new test blocks are non-inert. The one thing that WAS fund-adjacent (a wrong EVM chain id) is now entirely main's `resolveChainReference`, which this branch no longer alters.

## verdict
**APPROVE.** The only real value this PR still carried after the #1308/#1192 merges collapsed it was a latent build break; that's now repaired and the remaining test additions are green and non-inert.

---
🤖 reviewed by the Vultisig reviewer agent (round 4, fix-and-approve @ c3eec5f6)

PR #1191, opened 2026-07-12
perf(balance): batch evm balances via multicall + fiat prices per chain

### LINKED ISSUE #1183: perf: tested Multicall3 balance helper is wired to nothing; BalanceService + FiatValueService fan out one-at-a-time
Two N+1 shapes in the SDK's read layer, both with an existing batch primitive that isn't used:
1. `packages/sdk/src/vault/services/BalanceService.ts:88-104` issues one `getBalance` promise per tracked token → `packages/core/chain/coin/balance/resolvers/evm.ts:8-18` does one `eth_call` per coin. Meanwhile `packages/core/chain/coin/balance/getEvmChainBalances.ts:35-84` already implements tested Multicall3 batching (all native+ERC-20 in one RPC call) but its only callers are its own test file. Fix: group same-chain token requests in `BalanceService.getBalances`, call `getEvmChainBalances` once per chain.
2. `packages/sdk/src/services/FiatValueService.ts:292-322` loops single-id price lookups (`getErc20Prices({ids:[single]})`) even though that function accepts a multi-id array, and its own `batchFetchNativePrices` already batches correctly. `getTotalValue` also loops chains sequentially instead of `Promise.all`. 15 tokens × 3 chains = up to 45 serial calls.

Also flagged (lower priority, same file family): `GasEstimationService.ts:46-104` has no cache unlike its Balance/Price/Address siblings — add a short TTL.

Full context (stack-wide N+1 cluster): [round-6 roadmap artifact](https://claude.ai/code/artifact/c6f9de8a-0cf0-4096-80f1-4c09ab26bf40).

Severity: MED (portfolio-render latency, real RPC cost against user providers). Effort: S each. Risk: LOW (fallback paths exist).

gomesalexandre:
Closes vultisig/vultisig-sdk#1183

## what

Collapses two N+1 request fan-outs in the vault services into batched calls.

- **`BalanceService.getBalances({ includeTokens: true })`** previously issued one `getCoinBalance` RPC per token on a chain (native + N tokens = N+1 round-trips). EVM chains now fetch native + all tokens in a **single Multicall3 call** via the already-tested `getEvmChainBalances`, grouping same-chain requests. It respects the existing per-coin `BALANCE` cache (only the still-uncached coins get multicalled) and caches/emits each result exactly like `getBalance` does. Non-EVM chains keep the per-coin path, and `getEvmChainBalances`' built-in fallback covers EVM chains without a multicall3 contract.
- **`FiatValueService.getValues`** looped `getValue` per token, each firing a single-id `getErc20Prices` request. It now warms the price cache with **one batched `getErc20Prices` call** for the whole chain (the fn already took a multi-id array), then resolves native + token values concurrently.
- **`FiatValueService.getTotalValue`** now fetches every chain's values in parallel (`Promise.all`) instead of awaiting them one chain at a time.

## why

A multi-token EVM vault (native + K tokens) went from `K+1` balance RPCs to **1** multicall per chain, and from `K` price requests to **1** batched price request per chain. On a portfolio spanning M chains, `getTotalValue` also stops serializing the per-chain work. Pure perf/round-trip reduction, no behavioural change to the returned shapes.

## how tested

New regression tests (fail before, pass after):

- `BalanceService > batches native + token balances for an EVM chain into a single multicall` — asserts exactly ONE `getEvmChainBalances` call carrying native + token coins for an EVM chain, and that the non-EVM chain (Bitcoin) still uses the per-coin path. Before: multicall called 0 times (old per-token fan-out).
- `BalanceService > serves cached EVM balances without re-multicalling` — second identical call is fully cache-served.
- `FiatValueService > getValues > batches all token prices for a chain into a single getErc20Prices call` — asserts ONE batched `getErc20Prices({ ids: [usdc, dai], ... })`. Before: called 2 times (one per token).

Full `yarn check:ci` chain run locally end-to-end (build:all, cli/mcp build + smoke, sdk-bundle changeset guard, quality:contracts, lint, typecheck, knip, format:check, architecture, duplication, secrets/audit, markdown) — all green. Changeset included (`@vultisig/sdk` patch).

## risk

Low. Localized to the two read-only balance/price services; returned shapes unchanged; per-coin and Multicall3-fallback paths preserved for non-EVM and multicall-less chains.

## benchmark — call counts before/after

Measured by `tests/unit/services/balanceBatchingBenchmark.test.ts` against call-counting mocks, for a representative vault of **3 EVM chains x 5 tokens + 2 non-EVM native-only chains**:

| work | before (per-token fan-out) | after (batched) |
|---|---|---|
| balance RPCs | 20 (`3*(1+5)` EVM per-coin `getCoinBalance` + `2` non-EVM natives) | **5** (3 Multicall3 + 2 non-EVM natives) |
| price lookups (`getValues`, 1 chain x 5 tokens) | 5 single-id `getErc20Prices` | **1** batched `getErc20Prices` |
| `getTotalValue` over M chains | sequential (one chain awaited at a time) | concurrent (`Promise.all`; test asserts max in-flight > 1) |

Balance/price calls are network round-trips, so this is a direct latency/rate-limit reduction; `getTotalValue`'s chain loop also stops serializing (real wall-clock win on a live provider). The test computes the before-count deterministically from the fan-out shape (one call per coin) and asserts the measured after-count is fewer.

## graceful degradation — preserved (and improved)

The batched path degrades **strictly better** than the old fan-out, so this is not a regression:

- **Old** `getBalances` wrapped each chain's `Promise.all([...perToken])` in a single try/catch, so any ONE token's `getBalance` rejection dropped the ENTIRE chain (native + all tokens absent from the result).
- **New** path routes through `getEvmChainBalances`, which uses `multicall({ allowFailure: true })` and, on total-multicall failure (or a chain with no multicall3 contract), falls back to per-coin fetches (`getFallbackBalances`, each failure -> `0n`). It never rejects. A single coin failure surfaces as `0n` without dropping its siblings.
- `FiatValueService.getValues` keeps its per-token `.catch`, and `warmTokenPriceCache` is non-fatal (a failed batch price call falls back to individual per-token fetches).

Covered by two tests: `reports 0n for a coin the multicall omitted without dropping its siblings`, and `a failing non-EVM chain does not drop the EVM chains`.

> The full `yarn check:ci` chain (build:all, cli/mcp build + smoke, sdk-bundle changeset guard, quality:contracts, lint, typecheck, knip, format:check, architecture, duplication, secrets/audit, markdown) passes green end-to-end on this commit (exit 0).

🤖 Agent-generated (dev)

## Summary by CodeRabbit

* **Performance**
  * Faster EVM balance retrieval by batching native + token balances into a multicall where available.
  * Faster fiat valuation with batched ERC-20 price requests per chain and concurrent per-chain computations.
  * Improved responsiveness by capping simultaneous portfolio value calculations.
* **Reliability**
  * Prevents misleading cache updates when a balance result is missing or fails—absent results are not treated as `0`.
  * Keeps successful chain and asset results available even if some assets/chains fail.
* **Bug Fixes**
  * Ensures omitted/failed token balances can be refetched on subsequent calls.


packages/sdk/src/vault/services/BalanceService.ts
```diff
@@ -120,6 +117,88 @@ export class BalanceService {
...
+      address: address as Address,
+      coins: uncached.map(request => request.coinKey),
+    })
+
+    const tokensRecord = this.getTokensRecord()
+    for (const request of uncached) {
+      const rawBalance = rawBalances[accountCoinKeyToString(request.coinKey)] ?? 0n
+      const balance = formatBalance(rawBalance, chain, request.tokenId, tokensRecord)
+
+      await this.cacheService.setScoped(request.cacheKey, CacheScope.BALANCE, balance)
```
NeOMakinG: `should-fix:` transient EVM sub-call failures land here as `0n` and get cached + emitted as a real balance. `getEvmChainBalances` returns `0n` for a reverted `balanceOf`, a missing-multicall3 fallback failure, or an omitted coin — indistinguishable from a genuine zero. This loop then `setScoped`s that `0` into the shared `${chain}:${tokenId}` BALANCE cache for `balanceTTL` (default 5 min) and emits it. The old per-coin path threw on failure and cached nothing (retried next call). Downstream, `getBalance()` backs max-send (VaultBase.ts:1055) and max-swap (VaultBase.ts:1731, throws "Zero balance — nothing to swap"), so a flaky RPC can block swapping an asset the user actually holds until the TTL expires. Fail-safe (no overdraft) but a real display/decisioning bug. Have `getEvmChainBalances` surface per-coin failure so failed coins are skipped from caching/emitting and retried, rather than persisted as 0.

NeOMakinG (commented):
## testing approach

Two-lane review (Claude load-bearing + Codex gpt-5.4 adversarial second-opinion; GLM retired). Read the full changed source (`BalanceService.getEvmBalancesBatched`, `FiatValueService.getValues`/`warmTokenPriceCache`/`getTotalValue`) plus the core primitives it leans on (`getEvmChainBalances`, `accountCoinKeyToString`, `isFeeCoin`, `formatBalance`, cache TTLs) and the three new test files. Adversarial fund-safety lens: wrong balance/portfolio value → bad send/swap decision. Worktree pinned at HEAD `9052f950`. No runtime/QA (draft, unit-level change).

**Verified correct (no action):**
- **Multicall aggregation mapping is robust.** Balances are looked up by `accountCoinKeyToString(coinKey)` (`chain[:id]:address`), not by array index, and `getEvmChainBalances` keys its return with the identical function — so there is no order/index mix-up that could show token A's balance for token B. Native detection via `isFeeCoin = !coin.id` correctly routes the id-less native key to `getEthBalance` and tokens to `balanceOf`.
- **Decimals** are applied per-coin through the same `formatBalance(raw, chain, tokenId, tokensRecord)` call the old per-coin path used — no 1e-N drift.
- **Fiat per-chain pricing:** price cache key includes the chain (`${chain}:${tokenId}:${currency}`), so no cross-chain same-symbol price bleed; `warmTokenPriceCache` seeds the exact key `getPrice` reads (same case, same currency-lowercasing). A missing/0 price → token omitted from the chain's values (pre-existing, understates total but not new).
- Multicall3-absent chains and total-multicall failure fall back to per-coin fetches inside `getEvmChainBalances` (pre-existing, tested); `getTotalValue`'s new `Promise.all` uses disjoint per-chain cache keys (no race).

## should-fix

**A transiently-failed EVM sub-call is cached + emitted as a real `0` balance.** `getEvmChainBalances` normalizes a reverted ERC-20 call, a missing-multicall3 fallback failure, or an omitted coin to `0n` (indistinguishable from a genuine zero). The new `getEvmBalancesBatched` then `formatBalance`s that `0n`, `setScoped`s it into the `BALANCE` cache, and emits it as a real balance. This is a failure-mode change from the old per-coin path, which **threw** on failure and cached nothing (so the next read retried). Now a flaky RPC/token revert persists a false "0" for `balanceTTL` (default **5 min**) across every reader of the shared `${chain}:${tokenId}` cache key — including `getBalance()`, which backs `getMaxSendAmount` (`VaultBase.ts:1055`) and max-swap (`VaultBase.ts:1731`, which throws `Zero balance — nothing to swap`). The direction is fail-safe (understates → no overdraft), but the user sees a wrong zero and can be blocked from swapping an asset they actually hold, for up to the TTL. Converged Claude + Codex. Suggested fix: have `getEvmChainBalances` surface per-coin success/failure (or omit failed coins) so `BalanceService` skips caching/emitting the failed ones and retries next call, instead of persisting `0`.

**Degradation test asserts the wrong invariant.** `balanceBatchingBenchmark.test.ts:152` ("reports 0n for a coin the multicall omitted") explicitly models a failed call (`e.g. its call failed`) and asserts it degrades to `'0'` — i.e. it locks in the cache-poisoning behavior as intended. No test distinguishes a genuine zero from a failed-call zero, nor covers that the false zero is served from cache to a later `getBalance()` / max-swap read. If the caching behavior above is fixed, add a test that a failed sub-call's zero is NOT served from cache to a subsequent single-balance read.

## risk

Localized to two read-only balance/price services; returned shapes unchanged; no signer/tx-encoding touched. No overdraft path (max calcs only ever under-read). Residual risk is display/decisioning correctness on transient RPC failures + cache persistence of a false zero. No blocker.

## verdict

COMMENTED — draft hold. No blocking/fund-loss issue; the multicall aggregation, decimals, and per-chain price pairing are correct. One should-fix (converged both lanes): transient sub-call failures are cached/emitted as real `0` balances and persist for the balance TTL across the shared cache, reaching the max-swap "zero balance" guard. Not approving (draft).

🤖 Two-lane review: Claude (load-bearing) + Codex gpt-5.4 (converged). No GLM.

gomesalexandre (commented):
Adversarial pass. Verdict: **APPROVE-direction** - the batching is real, the claimed receipts reproduce, but one degradation behavior deserves a look before undraft.

## reproduced independently
- All 3 new/touched suites green on the branch (28/28): `BalanceService.test.ts`, `FiatValueService.test.ts`, `balanceBatchingBenchmark.test.ts`.
- "Fail before, pass after" claim verified for real: checked out main's `BalanceService.ts`/`FiatValueService.ts` under the new tests - 8 fail exactly as claimed (multicall never called, `getErc20Prices` called per-token, sequential `getTotalValue`).
- Cache-key parity traced: batched path writes `${chain.toLowerCase()}:native` / `${chain.toLowerCase()}:${token.id}` - byte-identical to `getBalance`'s keys, and `CacheService.buildKey` lowercases anyway. No double-fetch, no stale-as-fresh. The `serves cached EVM balances without re-multicalling` test locks it.
- Non-EVM path: `getBalancesPerCoin` is a verbatim extraction of the old loop, gated by `getChainKind(chain) === 'evm'`. Untouched, confirmed.
- `getEvmChainBalances` pre-exists on main (#886) - PR only consumes it, `isFeeCoin = !coin.id` matches the native `{ chain, address }` key shape.

## findings

**preferably-blocking:** the failure->`0n` degradation gets CACHED and EMITTED as a real balance. https://github.com/vultisig/vultisig-sdk/blob/9052f950a68d77cddb0aa23425b3ede9967295ef/packages/sdk/src/vault/services/BalanceService.ts#L190-L194 - a coin whose multicall entry failed (core collapses `status !== 'success'` to `0n`) or was omitted gets `setScoped(..., BALANCE, 0)` for the full balanceTTL (default 5min) plus an `emitBalanceUpdated` with amount 0. The old path threw and cached NOTHING - it never asserted a wrong value, it just went absent. "Degrades strictly better" is refutable on this axis: a transient RPC hiccup now shows the user a 0 balance for a coin they own, for up to 5 minutes, via an event consumers treat as fresh truth. Minimal fix: only cache/emit when `accountCoinKeyToString(request.coinKey) in rawBalances` (still return the `0n` entry in the result so siblings survive); follow-up in core to omit failed multicall entries instead of `0n`-ing them so the two failure shapes converge. The `reports 0n for a coin the multicall omitted` test would only need the cache/emit assertion flipped.

**suggestion:** `getTotalValue` now `Promise.all`s every chain unbounded - https://github.com/vultisig/vultisig-sdk/blob/9052f950a68d77cddb0aa23425b3ede9967295ef/packages/sdk/src/services/FiatValueService.ts#L395 - M chains each firing balance+price lookups concurrently where the old loop serialized. On a wide portfolio that is a rate-limit magnet on the price provider. A small concurrency cap (3-4) keeps ~all the win.

**nit:** `expect(1).toBeLessThan(beforePriceCalls)` in `balanceBatchingBenchmark.test.ts:110` is a tautology (1 < 5 always) - the real assertion is the `toHaveBeenCalledTimes(1)` above it, so this line is decorative.

Draft, so comment-only - not stamping.

packages/sdk/src/vault/services/BalanceService.ts
```diff
@@ -120,6 +117,97 @@ export class BalanceService {
...
+    const tokensRecord = this.getTokensRecord()
+    for (const request of uncached) {
+      const rawBalance = rawBalances[accountCoinKeyToString(request.coinKey)]
+      // A key MISSING from the multicall result means that coin was omitted/failed (a transient RPC
+      // hiccup, a partial Multicall3 aggregate). The old per-coin path cached NOTHING on failure; caching
+      // a fabricated 0n here (5min TTL) would show a real 0 for a coin the user owns AND emit it as a real
+      // balanceUpdated. Only cache + emit keys the multicall actually RETURNED (a genuine 0n from the
+      // aggregate IS a real balance and is kept); a missing key falls through uncached + unemitted so the
+      // next call refetches. It's still returned for this call so the shape is complete.
+      const present = rawBalance !== undefined
```
NeOMakinG, line 197: `should-fix:` the fix is INERT against the actual r1 failure mode — getEvmChainBalances never omits a key, it returns a present 0n, so `present` is always true and the poisoning is unchanged — the guard distinguishes ABSENT (`rawBalance === undefined`) from ZERO — which is the right *idea* — but the collaborator it reads from never produces an absent key. `getEvmChainBalances` (packages/core/chain/coin/balance/getEvmChainBalances.ts, UNCHANGED by this PR) normalizes every failure to a PRESENT 0n: multicall path L79 `result?.status === 'success' ? BigInt(...) : 0n`, and fallback path L27 `catch { return [key, 0n] }`. It returns `Object.fromEntries(coins.map(...))` — a complete Record with a key for every input coin, always. So for a real reverted `balanceOf` / partial-aggregate, `rawBalances[key]` is `0n` (not undefined) → `present = true` → still cached + emitted. Proof (ran locally): feeding the token key as a present `0n` (exactly what the real fn yields on failure), the guard still emits the phantom 0 AND caches it — a 2nd call served stale `0` from cache and did NOT re-multicall (getEvmChainBalances called once). Cache-key parity confirmed (`${chain}:${token.id}` / `${chain}:native` are byte-identical to getBalance's key), so the poison still reaches getBalance() → max-send (coin.id). r1's should-fix #1 is NOT resolved.

fix: push the omission DOWN into getEvmChainBalances: on `result.status !== 'success'` (L79) and on `catch` in getFallbackBalances (L27), OMIT the key (skip the entry) instead of returning `0n`. Only then does this `present` guard ever fire on a real failure. As-is this fix is a no-op at runtime.

packages/sdk/tests/unit/services/BalanceService.test.ts
```diff
@@ -54,28 +52,105 @@ describe('BalanceService', () => {
...
     expect(result[`${Chain.Ethereum}:${token.id}`]?.formattedAmount).toBe('5')
     expect(result[Chain.Bitcoin]?.formattedAmount).toBe('1')
   })
+
+  it('does NOT cache or emit a coin the multicall omitted, and refetches it next call (#1191)', async () => {
+    const ethAddress = `${Chain.Ethereum}-address`
+    const nativeKey = accountCoinKeyToString({ chain: Chain.Ethereum, address: ethAddress })
+    const tokenKey = accountCoinKeyToString({ chain: Chain.Ethereum, id: token.id, address: ethAddress })
+    // Multicall returns native but OMITS the token (a transient partial-aggregate / RPC hiccup).
+    vi.mocked(getEvmChainBalances).mockResolvedValue({
```
NeOMakinG, line 97: `should-fix:` the +44 test asserts a collaborator shape the real getEvmChainBalances never produces (inert-guard, PR #228 pattern) — it goes green while the real bug is intact — the test mocks `getEvmChainBalances` to return `{ [nativeKey]: ... }` — i.e. it OMITS the token key. But the real function never omits a key (it returns present `0n` on failure, see should-fix #1). So the test exercises a path production can't reach: green test, unfixed bug. It is non-inert against reverting *this guard*, but that's not the failure mode in the field.

fix: drive a REAL failure through getEvmChainBalances: mock it to return the failed coin as a present `0n` (its actual output on a reverted balanceOf) and assert that coin is NOT cached / NOT emitted / refetched next call. That test will (correctly) fail until should-fix #1 is fixed at the source.

packages/sdk/src/vault/services/BalanceService.ts
```diff
@@ -120,6 +117,97 @@ export class BalanceService {
...
+      const balance = formatBalance(present ? rawBalance : 0n, chain, request.tokenId, tokensRecord)
+
+      if (present) {
+        await this.cacheService.setScoped(request.cacheKey, CacheScope.BALANCE, balance)
+        this.emitBalanceUpdated({ chain, balance, tokenId: request.tokenId })
+      }
+
+      entries.push([request.resultKey, balance] as const)
```
NeOMakinG, line 205: `should-fix:` the batched return-map still leaks a fabricated 0, and fiat pairing makes it a legit-looking priced $0.00 row (Codex lane) — even setting cache/emit aside, getEvmBalancesBatched still `entries.push`es a formatted `0` for a failed-but-present coin, and balances()/balancesWithPrices()/allBalances()/portfolio() treat that returned map as authoritative for the call. balancesWithPrices() then pairs it with a real fetched price and computes fiat from BigInt(balance.amount), so a transient failure renders as a normal-looking `0 ETH / $0.00` row rather than an omitted/errored asset. Internal send('max')/swap('max') re-read via single-coin getBalance() rather than this map — but any consumer/UI gating off balances*/portfolio() sees the phantom zero.

fix: once should-fix #1 omits failed keys at the source, also omit them from the returned entries (don't push a phantom 0), so the read API and portfolio surface an incomplete-but-honest map instead of a fabricated zero.

packages/sdk/src/services/FiatValueService.ts
```diff
@@ -339,20 +410,20 @@ export class FiatValueService {
 
     // Calculate total - no separate cache needed since getValues() uses cached prices/balances
     const chains = this.getChains()
-    let total = 0
 
-    // Sum values across all chains
-    for (const chain of chains) {
+    // Fetch chains' values concurrently but with a bounded number in flight, so a many-chain vault
+    // doesn't fire an unbounded simultaneous RPC burst.
+    const chainTotals = await mapWithConcurrency(chains, TOTAL_VALUE_CONCURRENCY, async chain => {
```
NeOMakinG, line 416: `suggestion:` portfolio() and getTotalValue() disagree under the same outage — different balance paths (Codex lane) — portfolio() uses batched balancesWithPrices() (shows explicit zeroed assets), while getTotalValue() → getValues() calls single-asset getBalance() per asset and skips failures via per-asset catch. The same transient EVM failure yields 'asset exists with 0 amount' in portfolio() but an omission (undercount) in getTotalValue() — two surfaces disagree.

fix: unify the failure semantics across the two paths (both omit, or both zero) so portfolio total and per-asset breakdown are consistent under partial-outage.

packages/sdk/src/services/FiatValueService.ts
```diff
@@ -9,6 +9,27 @@ import type { FiatCurrency } from '@vultisig/core-config/FiatCurrency'
...
 
+// Max chains fetched at once when totalling portfolio value — bounds the RPC fan-out on many-chain vaults.
+const TOTAL_VALUE_CONCURRENCY = 8
+
+/**
+ * Map over items with a bounded number of in-flight promises, preserving input order. A raw
+ * Promise.all over every chain fans out one price/balance round-trip per chain simultaneously; on a
+ * many-chain vault that's an unbounded RPC burst. This caps concurrency while keeping the results ordered.
+ */
+async function mapWithConcurrency<T, R>(items: T[], limit: number, fn: (item: T) => Promise<R>): Promise<R[]> {
```
NeOMakinG, line 20: `suggestion:` praise: mapWithConcurrency itself is correct — both lanes verified: cursor is race-safe in JS's single-threaded loop (`cursor++` is atomic), order preserved via indexed writes, handles empty input, clamps limit<=0 to one worker via Math.max(1, limit), and throws from fn propagate. The pool logic is sound — the gap is only that not every fan-out site uses it (see should-fix updateValues).

fix: no change needed here.

NeOMakinG (commented):
gm — multi-lane pass: claude correctness+fund-safety + codex gpt-5.3-codex adversarial. conceptual-only (no runtime this pass). comment-only.

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | should-fix #1 (headline): transient EVM sub-call failure cached+emitted as a real 0 into the shared 5-min BALANCE cache, reaching getBalance()/max-send/max-swap and blocking an owned asset for 5 min | ❌ still-blocking (fix is INERT) — see should-fix #1 below |
| 1 | should-fix #2: degradation test locks in the poisoning; no test distinguishes genuine-zero vs failed-zero from cache | ⚠️ partial — a new test was added but it asserts an impossible collaborator shape (omitted key); the real failed-zero-from-cache path is still uncovered. see should-fix #2 |

## should-fix
### 1. the fix is INERT against the actual r1 failure mode — getEvmChainBalances never omits a key, it returns a present 0n, so `present` is always true and the poisoning is unchanged — `packages/sdk/src/vault/services/BalanceService.ts:197`
the guard distinguishes ABSENT (`rawBalance === undefined`) from ZERO — which is the right *idea* — but the collaborator it reads from never produces an absent key. `getEvmChainBalances` (packages/core/chain/coin/balance/getEvmChainBalances.ts, UNCHANGED by this PR) normalizes every failure to a PRESENT 0n: multicall path L79 `result?.status === 'success' ? BigInt(...) : 0n`, and fallback path L27 `catch { return [key, 0n] }`. It returns `Object.fromEntries(coins.map(...))` — a complete Record with a key for every input coin, always. So for a real reverted `balanceOf` / partial-aggregate, `rawBalances[key]` is `0n` (not undefined) → `present = true` → still cached + emitted. Proof (ran locally): feeding the token key as a present `0n` (exactly what the real fn yields on failure), the guard still emits the phantom 0 AND caches it — a 2nd call served stale `0` from cache and did NOT re-multicall (getEvmChainBalances called once). Cache-key parity confirmed (`${chain}:${token.id}` / `${chain}:native` are byte-identical to getBalance's key), so the poison still reaches getBalance() → max-send (coin.id). r1's should-fix #1 is NOT resolved.
action: push the omission DOWN into getEvmChainBalances: on `result.status !== 'success'` (L79) and on `catch` in getFallbackBalances (L27), OMIT the key (skip the entry) instead of returning `0n`. Only then does this `present` guard ever fire on a real failure. As-is this fix is a no-op at runtime.

### 2. the +44 test asserts a collaborator shape the real getEvmChainBalances never produces (inert-guard, PR #228 pattern) — it goes green while the real bug is intact — `packages/sdk/tests/unit/services/BalanceService.test.ts:97`
the test mocks `getEvmChainBalances` to return `{ [nativeKey]: ... }` — i.e. it OMITS the token key. But the real function never omits a key (it returns present `0n` on failure, see should-fix #1). So the test exercises a path production can't reach: green test, unfixed bug. It is non-inert against reverting *this guard*, but that's not the failure mode in the field.
action: drive a REAL failure through getEvmChainBalances: mock it to return the failed coin as a present `0n` (its actual output on a reverted balanceOf) and assert that coin is NOT cached / NOT emitted / refetched next call. That test will (correctly) fail until should-fix #1 is fixed at the source.

### 3. the concurrency cap is only half-applied — updateValues('all') still fans out an unbounded Promise.all, defeating the stated goal (Codex lane) — `packages/sdk/src/services/FiatValueService.ts:457`
the PR adds mapWithConcurrency(limit=8) to getTotalValue to 'bound the RPC fan-out on many-chain vaults', but updateValues('all') at L457 still does `await Promise.all(chains.map(c => this.getValues(c)))` — one price/balance round-trip per chain simultaneously, unbounded. 'Refresh all values' bypasses the cap entirely and can still stampede RPC/price backends or hang behind one slow chain.
action: route updateValues('all') through the same mapWithConcurrency(chains, TOTAL_VALUE_CONCURRENCY, ...) so both portfolio-total and refresh-all share the bound.

### 4. the batched return-map still leaks a fabricated 0, and fiat pairing makes it a legit-looking priced $0.00 row (Codex lane) — `packages/sdk/src/vault/services/BalanceService.ts:205`
even setting cache/emit aside, getEvmBalancesBatched still `entries.push`es a formatted `0` for a failed-but-present coin, and balances()/balancesWithPrices()/allBalances()/portfolio() treat that returned map as authoritative for the call. balancesWithPrices() then pairs it with a real fetched price and computes fiat from BigInt(balance.amount), so a transient failure renders as a normal-looking `0 ETH / $0.00` row rather than an omitted/errored asset. Internal send('max')/swap('max') re-read via single-coin getBalance() rather than this map — but any consumer/UI gating off balances*/portfolio() sees the phantom zero.
action: once should-fix #1 omits failed keys at the source, also omit them from the returned entries (don't push a phantom 0), so the read API and portfolio surface an incomplete-but-honest map instead of a fabricated zero.

## suggestion
### 1. portfolio() and getTotalValue() disagree under the same outage — different balance paths (Codex lane) — `packages/sdk/src/services/FiatValueService.ts:416`
portfolio() uses batched balancesWithPrices() (shows explicit zeroed assets), while getTotalValue() → getValues() calls single-asset getBalance() per asset and skips failures via per-asset catch. The same transient EVM failure yields 'asset exists with 0 amount' in portfolio() but an omission (undercount) in getTotalValue() — two surfaces disagree.
action: unify the failure semantics across the two paths (both omit, or both zero) so portfolio total and per-asset breakdown are consistent under partial-outage.

### 2. praise: mapWithConcurrency itself is correct — `packages/sdk/src/services/FiatValueService.ts:20`
both lanes verified: cursor is race-safe in JS's single-threaded loop (`cursor++` is atomic), order preserved via indexed writes, handles empty input, clamps limit<=0 to one worker via Math.max(1, limit), and throws from fn propagate. The pool logic is sound — the gap is only that not every fan-out site uses it (see should-fix updateValues).
action: no change needed here.

## risk
Fund-SAFE (understates → no overdraft), but the shipped fix does NOT resolve r1: a transient EVM sub-call failure still caches+emits a phantom 0 into the shared 5-min balance cache, still blocking send/swap of an owned asset and still rendering a false 0/$0.00 across balances()/portfolio(). The concurrency cap improves getTotalValue but leaves updateValues('all') unbounded. Draft — do not merge; the source-level omission in getEvmChainBalances is required to make the guard live.

## verdict
COMMENTED

rcoderdev (approved):

NeOMakinG (approved):
Round 3. The r1/r2 headline is now **closed at the source** — I fixed it on the branch (`5393a4f5`) rather than just re-flagging it, since the fix belonged in a file the PR didn't touch.

## testing approach

Unique worktree `/tmp/wt-vultisig-sdk-1191` at `533d5a85` (the round-2 tip advanced only by a `main` merge — no new author commits, so the r2 findings carried over verbatim).

- Read `getEvmChainBalances.ts` at HEAD to settle whether the PR's `present` guard is reachable. It was not.
- **Control-mutated before trusting any green.** SDK worktrees alias `@vultisig/*` back through `.config/vitest.core.config.ts` / `tests/unit/vitest.config.ts` — I confirmed each mutation was physically present in the file (grep) *before* reading the result, then confirmed RED.
- `yarn test:core` 213/216 files; `packages/sdk` unit 197/197 files. Two core failures are **pre-existing and unrelated**, both re-confirmed on a stashed clean tree: `compileTx.golden.test.ts` (cardano `auxiliaryData` typing) and `findSwapQuote.recipient.test.ts` (flaky — passes in isolation on both trees).
- `tsc --project .config/tsconfig.json --noEmit`: only the 2 pre-existing cardano `auxiliaryData` errors, none in touched files. eslint + prettier clean.
- **Codex lane: SKIPPED (dead).** `run-codex.sh` returned a 0-byte result. Recorded as skipped, not as "no findings".

## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | should-fix | transient EVM sub-call failure cached (5-min BALANCE TTL) + emitted as a real `0` | ✅ **fixed** | fixed at the source in `5393a4f5`, see below |
| r2 | 1 | should-fix | the `present` guard is **INERT** — `getEvmChainBalances` never produces an absent key | ✅ **fixed** | `5393a4f5` makes the collaborator omit unread keys; guard is now live |
| r1 | 2 / r2 | should-fix | tests mock an omitted key, a shape the real fn never produced (inert-guard pattern) | ✅ **fixed** | the mocked shape is now the real one; seam pinned by 2 new `getEvmChainBalances.test.ts` cases |
| r2 | 3 | should-fix | `updateValues('all')` still a raw unbounded `Promise.all` — cap only half-applied | ✅ **fixed** | `5393a4f5`, now uses the PR's own `mapWithConcurrency` + a pinning test |
| r2 | 4 | should-fix | batched return-map still surfaces a fabricated `0` for a failed coin | ⚠️ **partially addressed** | severity dropped, see should-fix below |
| r2 | — | praise | `mapWithConcurrency` correctness | ℹ️ still correct | now used at both call sites |

## what was actually wrong

`getEvmChainBalances` normalized **every** failure to a *present* `0n`:

- multicall `allowFailure: true` keeps a reverted sub-call in the array as `status: 'failure'` → `: 0n`
- `getFallbackBalances` `catch { return [key, 0n] }`
- both return `Object.fromEntries` over **every input coin**, so a key always existed

Its only production consumer is this PR's `getEvmBalancesBatched`, whose guard is `const present = rawBalance !== undefined`. Right idea, but nothing could ever make it false — so a transient RPC hiccup was written into the shared `${chain}:${tokenId}` BALANCE cache (5-min TTL, byte-identical to the key `getBalance()` reads) and emitted as a real `balanceUpdated`. The old per-coin path threw and cached nothing.

Fix in `5393a4f5`: **omit** the key when the balance could not be read. A present `0n` is a genuine zero; an absent key is "unknown" and gets refetched next call. Pinned by two cases in `getEvmChainBalances.test.ts` — a failed sub-call is ABSENT, and a `status: 'success', result: 0n` stays PRESENT. Both halves of the fix (multicall branch and fallback `catch`) were individually reverted as control mutants and each went RED.

Also verified `getEvmChainBalances` has **no other production consumer**, so tightening its contract is contained.

## should-fix

**The returned record still carries a fabricated `'0'` for a coin whose read failed.** `getEvmBalancesBatched` no longer caches or emits it, but still pushes `formatBalance(0n)` into `entries` — the code comment calls this out deliberately ("still returned for this call so the shape is complete"). Net effect is now bounded to a single refresh cycle on the display surface: the cache is clean, no event fires, and `send('max')` / `swap('max')` re-read through single-coin `getBalance()` which throws rather than zeroing. But `vault.balances()` / `portfolio()` will show `0 USDC` for an owned token for that one call, and `balancesWithPrices()` pairs it with a real price into a legit-looking `$0.00` row.

I did **not** change this, because it contradicts an explicit design comment the author wrote and it is a consumer-visible shape change — that call is yours, not mine to make silently. If you want it: drop the `entries.push` for `!present` too. Consumers already handle absent keys (the old path dropped the whole chain on any failure, and everything in-repo iterates with `Object.entries`), and `balanceBatchingBenchmark.test.ts:152` would need its `'0'` assertion re-pointed at the sibling-preservation invariant it actually cares about.

## q

`warmTokenPriceCache` writes `${chain}:${tokenId}:${currency}` with `tokenId` **un-lowercased**, but reads the upstream response as `prices[tokenId.toLowerCase()]`. That matches what `getPrice` reads today (it also uses the raw `tokenId`), so it is correct as written — but it means the seed only hits if the token list and the caller use identical casing. Worth a follow-up normalizing both sides, not blocking.

## risk

Low, and lower than before this round. Direction of every remaining failure mode is fail-safe (understate, never overdraft). Batching correctness re-confirmed from r1: balances are keyed by `accountCoinKeyToString`, not array index, so there is no token A/B mix-up; decimals go through the same `formatBalance` as the old path; the price cache key includes the chain, so same-symbol tokens do not collide across chains; a chain with no Multicall3 metadata falls through to the per-coin resolver. No dependency changes, so no caret-bump exposure.

## verdict

**APPROVE** at `5393a4f5`. The r1/r2 blocking finding is closed at the source with mutation-proven tests; the one remaining should-fix is a documented, bounded display concern I've left as your call.

---
🤖 reviewed by the vultisig reviewer agent — Claude lane load-bearing, Codex lane dead (0-byte), recorded as skipped.

PR #1163, opened 2026-07-11
fix(sdk): export THOR/Maya swap memo parser

gomesalexandre:
## Summary
- move THOR/Maya swap-memo parsing into the public `@vultisig/sdk` API
- keep the parser behavior identical for shorthand expansion, ERC-20 suffix stripping, and `VaultError` handling
- update the CLI executor to import the SDK helper instead of maintaining its own parser copy
- add the linked changeset for the public API addition

## Issue
- Closes #1162
- Report mirror: https://gist.github.com/gomesalexandre/4051176f53d7875d31cb44de9df5475c

## What changed
- added `parseThorSwapMemo` + `ParsedThorSwapMemo` to the SDK public surface
- moved THOR/Maya swap memo tests into SDK unit coverage and added a public-export assertion
- updated `AgentExecutor.signThorMsgDepositSwap()` to use `parsed.toChain`
- removed the duplicated CLI-local parser/test copy

### Targeted
- `yarn workspace @vultisig/sdk test tests/unit/utils/parseThorSwapMemo.test.ts tests/unit/utils/publicExports.test.ts` ✅
- `yarn build:sdk && yarn workspace @vultisig/cli test src/agent/__tests__/executor.lp.test.ts && yarn cli:build` ✅
- `node clients/cli/dist/index.js --help` ✅

### Broader
- `yarn build:all` ✅
- `yarn lint` ✅
- `yarn test` ✅

## Files changed
- `.changeset/fix-sdk-thor-memo-parser-r7.md`
- `clients/cli/src/agent/executor.ts`
- `packages/sdk/src/index.ts`
- `packages/sdk/src/utils/thorSwapMemo.ts`
- `packages/sdk/tests/unit/utils/parseThorSwapMemo.test.ts`
- `packages/sdk/tests/unit/utils/publicExports.test.ts`

## Summary by CodeRabbit

- **New Features**
  - Added public SDK support for parsing THORChain and MayaChain swap memos, including exported parsing utilities.
  - Made swap-memo parsing available through both standard and React Native SDK entry points, with destination-chain resolution and asset shorthand handling.

- **Bug Fixes**
  - Updated CLI swap processing to use the shared SDK memo parser for consistent memo interpretation and destination handling.

- **Tests**
  - Expanded unit coverage for swap-memo parsing, including validation for unsupported destination chains and ensuring public exports are present.

PR #1193, opened 2026-07-12
fix(sdk): reconcile + export dangerous/burn-address guard

### LINKED ISSUE #1160: arch: dangerous/burn-address guard duplicated with divergent content between SDK and abts
The SDK's own file header for its dangerous-address list explicitly documents a past CCTP-related incident caused by exactly this class of guard drifting out of sync — the team already knows this pattern is dangerous, but the guard's content has diverged between the SDK's copy and agent-backend-ts's copy today.

Full context: [round-5 architecture artifact](https://claude.ai/code/artifact/13bddb9f-4c87-440a-aae8-e30356f5a1a1).

Severity: MED (documented as dangerous by the codebase's own history). Effort: S/M — needs a reconciliation pass to determine which copy is more complete before merging into one.

gomesalexandre:
Closes vultisig/vultisig-sdk#1160

## what

Reconciles the SDK's dangerous/burn-address guard (`utils/dangerousAddresses.ts`) with its drifted siblings, then exports it from the SDK public API.

The guard's own header documents a past CCTP `mintRecipient` burn incident caused by exactly this class of per-call-site drift. On close read, the SDK copy had itself fallen out of sync with its authoritative parity source (mcp-ts `src/lib/dangerous-addresses.ts`) AND with agent-backend-ts's copy (`mastra/guardrails.ts` `BASE58_BURNS`). No single copy was complete:

| burn address | SDK (before) | mcp-ts | abts |
|---|---|---|---|
| EVM zero / dead / dead-variant | ✅ | ✅ | ✅ |
| Solana System Program | ✅ | ✅ | ✅ |
| SPL Token Program | ❌ | ✅ | ✅ |
| Wrapped SOL mint | ❌ | ✅ | ✅ |
| Solana Incinerator | ✅ | ❌ | ✅ |
| Bitcoin null / eater | ❌ | ✅ | ✅ |
| XRP ACCOUNT_ZERO / ACCOUNT_ONE | ❌ | ✅ | ✅ |

Rather than pick one "authoritative" copy (none was a superset), I merged the **union** — every address from every copy is kept. The guard is only ever additive/tightening; nothing that was previously rejected is now allowed.

## why

An un-exported + drifted burn guard is the CCTP-incident bug class: a build_* primitive on one repo lacks a burn address the others have, so it happily encodes value-moving calldata to a black-hole. Making the SDK own + export the reconciled union lets the app + agent-backend-ts consume one source of truth instead of maintaining divergent copies. Non-EVM lists stay chain-family-scoped (Solana/UTXO/XRP), so a burn address for one family never blocks an unrelated chain.

Migrating app + abts to import this is a separate follow-up.

## how tested

Extended `tests/unit/utils/dangerousAddresses.test.ts` (fails before, passes after — before the reconcile the new `UTXO_DANGEROUS_ADDRESSES`/`XRP_DANGEROUS_ADDRESSES` exports don't exist and the suite can't even collect):

- rejects SPL Token Program + Wrapped SOL mint on Solana
- keeps the Solana Incinerator (the entry only this SDK copy carried pre-reconcile)
- rejects Bitcoin null/eater burns on UTXO chains (Bitcoin, Litecoin, …)
- rejects XRP ACCOUNT_ZERO / ACCOUNT_ONE on Ripple
- keeps burn lists chain-family-scoped (no cross-family leakage)
- is exported from the SDK public API

Full `yarn check:ci` chain run locally end-to-end (build:all, cli/mcp build + smoke, sdk-bundle changeset guard, quality:contracts, lint, typecheck, knip, format:check, architecture, duplication, secrets/audit, markdown) — all green (exit 0). Changeset included (`@vultisig/sdk` patch).

## risk

Low-ish. The guard now throws for additional Solana/UTXO/XRP burn destinations that it should always have rejected — tightening a fund-safety guard, not loosening it. EVM behaviour is unchanged. Existing SDK consumers (`buildCctpBridge`, `acrossQuote`) call it for EVM destinations and are unaffected.

🤖 Agent-generated (dev)


packages/sdk/src/index.ts
```diff
@@ -307,6 +307,23 @@ export { getSwapExplorerUrl, swapExplorerProviders } from '@vultisig/core-chain/
 // that can drift from each other (the mcp-ts #384 bug class).
 export { isSkipRoutableChain, isTerraChain, willRouteViaSkip } from '@vultisig/core-chain/swap/skip/skipRouting'
 
+// Dangerous/burn-address guard. Single source of truth for "is this destination
+// a burn/black-hole address that no key controls?" across EVM, Solana, UTXO and
+// XRP. Exported so the app + agent-backend-ts consume it instead of maintaining
+// their own copies that drift (the CCTP mintRecipient burn-drift incident this
+// list's header documents). Reconciled to the union of all three prior copies.
+export {
+  assertSafeDestination,
```
NeOMakinG, line 347: `preferably-blocking:` exported guard is undefined on the react-native entry — the PR's own "app consumes the single source of truth" goal is a no-op on the app's platform — `src/index.ts` feeds node/browser/electron/chrome via `export * from '../../index'` (node/index.ts:118), so this block lands there automatically. But the **react-native** entry `src/platforms/react-native/index.ts` is hand-curated (614 lines, no `export * from '../../index'`) and does NOT re-export the guard — I grepped it: zero hits for `assertSafe`/`dangerous`/`isEvmBurn`. So `import { assertSafeDestination } from '@vultisig/sdk'` resolves to `dist/index.react-native.js` → **undefined** on RN, exactly the #1163/#1165/#1182 export-reconcile family gap. No immediate fund-loss (the internal callers import the module directly, and the new non-EVM tables have no internal caller yet), so it's not blocking a merge — but the export half of this PR silently doesn't reach the platform it's built for, and the follow-up app migration will `undefined(...)` or no-op.

fix: mirror these nine exports into `src/platforms/react-native/index.ts` (add `export { assertSafeDestination, assertSafeEvmDestination, isEvmBurnAddress, getEvmDangerousReason, getChainDangerousReason, EVM_DANGEROUS_ADDRESSES, SOLANA_DANGEROUS_ADDRESSES, UTXO_DANGEROUS_ADDRESSES, XRP_DANGEROUS_ADDRESSES } from '../../utils/dangerousAddresses'`), and add an RN-parity assertion so the next hand-curated-entry drift fails a test.

packages/sdk/src/utils/dangerousAddresses.ts
```diff
@@ -57,8 +85,17 @@ export const isEvmBurnAddress = (address: string): boolean => getEvmDangerousRea
 
 /** Return the chain-specific dangerous-address reason, if one applies. */
 export const getChainDangerousReason = (chain: string, address: string): string | undefined => {
-  if (chain.trim().toLowerCase() === 'solana') {
-    return SOLANA_DANGEROUS_ADDRESSES[address.trim()]
+  const normalizedChain = chain.trim().toLowerCase()
+  const destination = address.trim()
```
NeOMakinG: `should-fix:` EVM path does not trim — a whitespace-padded EVM burn bypasses the guard, while the non-EVM path (and the older isNullAddress) both trim — This PR adds `const destination = address.trim()` to the chain path (line 89), but `getEvmDangerousReason`/`assertSafeEvmDestination` (lines 78-81, 109-114) test the RAW string against `^0x[0-9a-fA-F]{40}$` with no trim. I confirmed at runtime: `assertSafeDestination('Ethereum', ' 0x000…dead ')` (surrounding spaces) → does NOT throw, while the trimmed `'0x…dead'` does. The Solana equivalent `' 11111…111 '` DOES throw (chain path trims). So the new exported "source of truth" is actually *less* whitespace-robust than the `isNullAddress` it's meant to supersede (which trims). Pre-existing shape mirrored from mcp-ts, but this PR is the one making it the canonical exported guard and adding the asymmetric trim.

fix: trim in `getEvmDangerousReason` before the regex (`const a = address.trim()`), or trim once at the top of `assertSafeDestination`/`assertSafeEvmDestination`, so EVM and non-EVM normalise identically. Add a whitespace-padded EVM burn case to the test.

.changeset/reconcile-dangerous-address-guard.md
```diff
@@ -0,0 +1,25 @@
+---
+'@vultisig/sdk': patch
```
NeOMakinG, line 3: `should-fix:` changeset is `patch` but this adds new named exports to the package root — additive public-API surface = semver minor — Nine new named exports are added to `@vultisig/sdk`'s root entry. New public-API surface is an additive (minor) change under semver, not a patch — same call the repo made on sdk#1182 (re-export → patch bumped to minor per the CHANGELOG precedent). Under-bumping means downstreams pinned to `~x.y.z` won't pull the new surface on `npm update`.

fix: change `'@vultisig/sdk': patch` → `minor`.

NeOMakinG (commented):

## preferably-blocking
### 1. exported guard is undefined on the react-native entry — the PR's own "app consumes the single source of truth" goal is a no-op on the app's platform — `packages/sdk/src/index.ts:316`
`src/index.ts` feeds node/browser/electron/chrome via `export * from '../../index'` (node/index.ts:118), so this block lands there automatically. But the **react-native** entry `src/platforms/react-native/index.ts` is hand-curated (614 lines, no `export * from '../../index'`) and does NOT re-export the guard — I grepped it: zero hits for `assertSafe`/`dangerous`/`isEvmBurn`. So `import { assertSafeDestination } from '@vultisig/sdk'` resolves to `dist/index.react-native.js` → **undefined** on RN, exactly the #1163/#1165/#1182 export-reconcile family gap. No immediate fund-loss (the internal callers import the module directly, and the new non-EVM tables have no internal caller yet), so it's not blocking a merge — but the export half of this PR silently doesn't reach the platform it's built for, and the follow-up app migration will `undefined(...)` or no-op.
action: mirror these nine exports into `src/platforms/react-native/index.ts` (add `export { assertSafeDestination, assertSafeEvmDestination, isEvmBurnAddress, getEvmDangerousReason, getChainDangerousReason, EVM_DANGEROUS_ADDRESSES, SOLANA_DANGEROUS_ADDRESSES, UTXO_DANGEROUS_ADDRESSES, XRP_DANGEROUS_ADDRESSES } from '../../utils/dangerousAddresses'`), and add an RN-parity assertion so the next hand-curated-entry drift fails a test.

## should-fix
### 1. SDK root now exposes TWO divergent burn-checkers — `isNullAddress` (flat, partial) vs `assertSafeDestination` (family-scoped, full) — which contradicts the "single source of truth" thesis and is a migration footgun — `packages/sdk/src/index.ts:753`
`isNullAddress` is already exported from the SDK root (index.ts:753, via tools/validate). It is chain-AGNOSTIC and trims, but only covers {EVM zero, EVM `0x…dead`, Solana System Program, Solana Incinerator} — it MISSES the EVM dead-variant `0xdead…2069`, SPL Token, Wrapped SOL, all UTXO burns and both XRP burns (recipientSanity.ts:56-101, a completely separate impl that does not import dangerousAddresses.ts). The new `assertSafeDestination` covers the full union but only if the caller passes the correct chain: I ran it with `chain="xrp"` (instead of the canonical `"Ripple"`) against the ACCOUNT_ZERO black-hole and it did NOT throw — the non-EVM guard silently no-ops on any chain string that isn't exactly `solana`/`ripple`/the six UTXO names. So a consumer told to "consume the single source of truth" can still ship a false-negative by (a) keeping `isNullAddress`, or (b) calling `assertSafeDestination` with absent/aliased chain. Since abts's current `isNullAddress` is chain-agnostic, migrating it to the family-scoped SDK guard is a coverage regression unless every call site passes the canonical Chain enum value.
action: in the follow-up, either route `isNullAddress` through the reconciled tables (one impl) or deprecate it, and document that `assertSafeDestination` requires the canonical `Chain` enum value (not an alias / not empty) for the non-EVM families — ideally guard against an unrecognised non-EVM chain string rather than silently returning undefined.

### 2. EVM path does not trim — a whitespace-padded EVM burn bypasses the guard, while the non-EVM path (and the older isNullAddress) both trim — `packages/sdk/src/utils/dangerousAddresses.ts:89`
This PR adds `const destination = address.trim()` to the chain path (line 89), but `getEvmDangerousReason`/`assertSafeEvmDestination` (lines 78-81, 109-114) test the RAW string against `^0x[0-9a-fA-F]{40}$` with no trim. I confirmed at runtime: `assertSafeDestination('Ethereum', ' 0x000…dead ')` (surrounding spaces) → does NOT throw, while the trimmed `'0x…dead'` does. The Solana equivalent `' 11111…111 '` DOES throw (chain path trims). So the new exported "source of truth" is actually *less* whitespace-robust than the `isNullAddress` it's meant to supersede (which trims). Pre-existing shape mirrored from mcp-ts, but this PR is the one making it the canonical exported guard and adding the asymmetric trim.
action: trim in `getEvmDangerousReason` before the regex (`const a = address.trim()`), or trim once at the top of `assertSafeDestination`/`assertSafeEvmDestination`, so EVM and non-EVM normalise identically. Add a whitespace-padded EVM burn case to the test.

### 3. changeset is `patch` but this adds new named exports to the package root — additive public-API surface = semver minor — `.changeset/reconcile-dangerous-address-guard.md:2`
Nine new named exports are added to `@vultisig/sdk`'s root entry. New public-API surface is an additive (minor) change under semver, not a patch — same call the repo made on sdk#1182 (re-export → patch bumped to minor per the CHANGELOG precedent). Under-bumping means downstreams pinned to `~x.y.z` won't pull the new surface on `npm update`.
action: change `'@vultisig/sdk': patch` → `minor`.

## risk
FUND-CRITICAL guard, but this PR is fund-SAFE in isolation. The reconcile is purely additive/tightening — every burn address blocked before is still blocked, and I verified the union is complete against both parity sources: mcp-ts `src/lib/dangerous-addresses.ts` and agent-backend-ts `guardrails.ts` `BASE58_BURNS` — no entry from either copy is missing, no silent removal, no new false-positive on a legit destination. I transpiled `dangerousAddresses.ts` in isolation and exercised the guard at the logic level: all 16 canonical cases pass (4 Solana, 2 UTXO across Bitcoin/Litecoin/Bitcoin-Cash, 2 XRP, 3 EVM incl. checksummed `0x…dEaD` and upper-case dead-variant), and family-scoping holds (a Bitcoin eater on Solana / a wSOL mint on Bitcoin correctly do NOT throw). The two internal call sites (`buildCctpBridge`, `acrossQuote`) import the guard directly and are unaffected by the new export. Residual risk is entirely in the *export half* and the un-written follow-up migration: the symbol is undefined on the app's platform (RN), and the SDK root now carries two divergent burn-checkers. Draft — COMMENTED, not approving.

## verdict
COMMENTED

gomesalexandre (commented):
Adversarial pass on the fund-safety claims. Verdict: **APPROVE-direction** for the union + export itself, but the "can't drift per call-site again" claim is not yet true INSIDE this repo - two in-repo siblings still carry private copies. Details below.

## reproduced independently
- 20/20 tests green on the branch; swapped in main's `dangerousAddresses.ts` under the new suite -> fails to collect exactly as claimed (new exports missing).
- **Union verified programmatically** (parsed before/after key sets from main vs branch): EVM 3->3, Solana 2->4, UTXO 0->2, XRP 0->2, **REMOVED: none** on every list. Nothing previously blocked is now unblocked.
- **Union completeness** cross-checked against both siblings: every mcp-ts `dangerous-addresses.ts` entry and every abts `guardrails.ts` `BASE58_BURNS` entry (all 8) is present, plus the SDK-only Incinerator. Genuine superset of all three copies.
- **Chain-family scoping** checked against the actual `Chain` enum: `Chain.BitcoinCash = 'Bitcoin-Cash'` lowercases to the `'bitcoin-cash'` key in `UTXO_CHAINS`, `Chain.Ripple = 'Ripple'` -> `'ripple'`. No spelling escape. Cross-family leak test is real (Bitcoin eater on Solana does NOT throw, WSOL on Bitcoin does NOT throw). Base58 lengths don't collide across families either (BTC burns are 27-34 chars, Solana pubkeys 32-44 - and family-keyed anyway).
- `buildCctpBridge` + `acrossQuote` import the shared table - the two existing call sites are on the single source. ✓

## findings

**preferably-blocking:** two IN-REPO drift siblings were not reconciled, which is the exact bug class this PR exists to kill:
1. https://github.com/vultisig/vultisig-sdk/blob/e642a6b11e663b412d9c694e371646eab00f9b5a/packages/sdk/src/tools/validate/recipientSanity.ts#L59-L104 - `isNullAddress` hand-pins dead + System Program + Incinerator but MISSES Tokenkeg, WSOL, the BTC null/eater pair and both XRP accounts. So `recipientSanity({ recipient: 'TokenkegQ...' })` reports NOT flagged while `assertSafeDestination` would throw for the same address.
2. https://github.com/vultisig/vultisig-sdk/blob/e642a6b11e663b412d9c694e371646eab00f9b5a/packages/core/chain/swap/quote/findSwapQuote.ts#L524-L531 - `assertValidCustomRecipient` pins only zero + `0x...dead` and MISSES the `0xdead...42069420694206942069` variant that `EVM_DANGEROUS_ADDRESSES` carries, plus every base58 burn for non-EVM custom recipients. A swap quote with customRecipient = the dead-variant sails through this gate today.

Both fixes are pure-superset (route them through `getEvmDangerousReason` / `getChainDangerousReason`), so same risk profile as the rest of the PR. Alternatively state explicitly in the PR body that in-repo consumers are part of the app/abts migration follow-up - but leaving the header claiming "it can't drift per call-site again" while two call sites in the same repo still drift is how the NEXT CCTP-class incident gets archaeology'd back to this PR.

**nit:** `UTXO_CHAINS` covers bitcoin/litecoin/dogecoin/bitcoin-cash/dash/zcash but the two burn addresses are Bitcoin-mainnet version-byte addresses - they're not even valid on Litecoin/Doge/etc. Harmless (never false-blocks), just noting the set membership is mostly decorative outside `'bitcoin'`.

Draft, so comment-only - not stamping.

NeOMakinG (commented):
gm — multi-lane pass: claude (load-bearing) + codex gpt-5.4 (alive — net-new HIGH). conceptual-only (no runtime this pass). comment-only.

## prior issues — status
| round | issue | status |
|---|---|---|
| R1 | preferably-blocking: RN-entry parity gap — guard not exported from react-native/index.ts | ❌ still open — this commit does not touch any RN entry (verified NONE in react-native/index.ts) |
| R1 | should-fix: two divergent burn-checkers (isNullAddress partial, coexisting with the new guard) | ✅ fixed @61a8b496 — isNullAddress + findSwapQuote both route through the one core-chain table; union is a superset (non-inert probe: 7 previously-missed burns now flag TRUE, 3 carried-over still TRUE, no regression) |
| R1 | should-fix: chain="xrp"≠"Ripple" silent no-op | ⚠️ partial — getChainDangerousReason still requires canonical 'ripple' (verified 'xrp'→undefined); in-repo callers pass canonical enum so NO live SDK bug, but export-consumer gap remains |
| R1 | should-fix: EVM path no-trim → whitespace-padded burn bypass | ⚠️ partial — getEvmDangerousReason still untrimmed, but ALL in-repo callers pre-trim (normalizedRecipient / .trim() / viem getAddress / addr.trim()) so no live bypass; primitive hardening remains |
| R1 | should-fix: changeset patch→minor (additive public API) | ❌ still open — both packages still marked patch |

## preferably-blocking
### 1. `0X`-prefixed EVM burn slips the swap custom-recipient guard on native routes — `packages/core/chain/swap/quote/findSwapQuote.ts:530`
Codex net-new, empirically confirmed. `assertValidCustomRecipient` gates the EVM burn check behind the local `isEvmAddress = /^0x[0-9a-fA-F]{40}$/` (no `/i`, lowercase `0x` only). The shared `getEvmDangerousReason` it then calls IS case-insensitive (`/i`) — but that path is unreachable for a `0X`-prefixed input, so a `0X…dEaD` recipient on a native (THOR/Maya, cross-chain) custom-recipient route falls through BOTH burn checks (`getChainDangerousReason(to.chain=EVM)` returns undefined) and is forwarded verbatim as `destination`. Probe: `guard('0X…dEaD','Ethereum') → PASSED-THROUGH` vs `guard('0x…dead','Ethereum') → BLOCKED`. The changeset explicitly claims `0X…` can't slip past — that protection is dead code behind this gate. Fund-loss direction (false-negative); exotic `0X` input + THOR downstream backstop keep it below blocker, but the advertised guard is defeated.
action: add `/i` to the local `isEvmAddress` regex (or drop the pre-gate and call `getEvmDangerousReason` first, then `getChainDangerousReason`); add a `0X…`-prefixed burn test on a native route.

### 2. burn guard still unreachable on the react-native entrypoint (the app's platform) — `packages/sdk/src/platforms/react-native/index.ts`
Carried from R1, unchanged this round. node/index.ts re-exports via `export * from '../../index'`; the hand-curated RN entry has no `export *` and re-exports none of `assertSafeDestination` / `isNullAddress` / `recipientSanity` / the reason helpers. The PR's stated goal — app + agent-backend-ts consume the single source of truth — is a no-op on RN. Verified: grep of react-native/index.ts for dangerous|assertSafe|isNullAddress|recipientSanity → none.
action: mirror the dangerous-address exports into the RN entry, or add a parity assertion test that fails when a public export is missing from RN.

## should-fix
### 1. `key in map` / `map[key]` lets `toString`/`constructor`/`__proto__` false-positive as burns — `packages/sdk/src/tools/validate/recipientSanity.ts:91`
New this round. `isNullAddress` uses `trimmed in SOLANA_DANGEROUS_ADDRESSES` (+ UTXO/XRP), and `getChainDangerousReason` uses bracket `MAP[dest]` — both read Object.prototype. Probe: `isNullAddress('toString'|'constructor'|'__proto__') → true`; `getChainDangerousReason('solana','toString') → truthy (a function) → assertSafeDestination throws`. This is over-block (false-POSITIVE) only — never a fund-loss false-negative, and no valid address equals a prototype key, so low practical impact — but it's a correctness defect on a fund guard and cheap to close.
action: use `Object.hasOwn(MAP, key)` for the membership tests and `Object.hasOwn(MAP,dest) ? MAP[dest] : undefined` for the lookup (or null-prototype objects / a `Map`).

### 2. chain-keyed guard silently no-ops on non-canonical chain names (`xrp`, `SOL`, `BTC`) — `packages/core/chain/security/dangerousAddresses.ts:96`
Carried from R1 (partial). `getChainDangerousReason` matches only canonical lowercase `'ripple'`/`'solana'`/the UTXO set. Verified `getChainDangerousReason('xrp', <XRP black-hole>) → undefined`. In-repo callers are SAFE (findSwapQuote passes canonical `to.chain` enum; acrossQuote is EVM-only) so no live SDK bug — but the exported `assertSafeDestination`/`getChainDangerousReason` are the very surface app + agent-backend-ts are told to consume, and a ticker-style alias silently skips the guard. (isNullAddress is chain-agnostic so it still catches XRP.)
action: map known aliases to families, or fail-closed / throw on an unrecognized non-EVM chain string instead of returning undefined.

### 3. `getEvmDangerousReason` still doesn't trim while its sibling `getChainDangerousReason` does — `packages/core/chain/security/dangerousAddresses.ts:81`
Carried from R1 (partial). `getEvmDangerousReason` tests the raw string against anchored `^0x…$`, so `' 0x…dead '` → undefined, while `getChainDangerousReason` trims internally — asymmetric. All in-repo callers pre-trim (findSwapQuote `normalizedRecipient`, acrossQuote `.trim()`, cctp viem `getAddress`, isNullAddress `addr.trim()`), so no live bypass today, but the exported primitive is less whitespace-robust than the guard it supersedes.
action: trim inside `getEvmDangerousReason` for symmetry with `getChainDangerousReason`.

### 4. changeset is `patch` for an additive public-API surface — `.changeset/reconcile-dangerous-address-guard.md:2`
Carried from R1. Both `@vultisig/core-chain` and `@vultisig/sdk` are marked `patch`, but the SDK exposes ~9 new root named exports (assertSafeDestination, getEvmDangerousReason, …) plus a new core-chain subpath export — additive public API is `minor` under semver (repo precedent, sdk#1182).
action: bump the sdk (and arguably core-chain) changeset to `minor`.

## q
- which chain-name form will the app + agent-backend-ts pass to the exported `assertSafeDestination` — canonical `Chain` enum values, or ticker/lowercase aliases? That decides whether finding #4 (xrp alias no-op) is latent or live for the intended consumers.

## risk
Reconcile half is fund-safe: the divergence is genuinely closed — both in-repo siblings now consume the one core-chain table, the union is a proven superset (7 previously-missed burns now flag, no carried-over burn dropped, legit addresses don't false-positive), and every in-repo caller pre-trims. The one fund-direction (false-negative) gap is Codex's `0X`-prefix slip on native custom-recipient swap routes — narrow (exotic input + THOR backstop) but it defeats the exact protection the changeset advertises. The remaining findings are export-surface robustness (RN parity, alias no-op, primitive trim) and a non-fund-loss prototype-key false-positive. Draft → COMMENTED, no approve.

## verdict
COMMENTED

packages/core/chain/swap/quote/findSwapQuote.ts
```diff
@@ -518,24 +519,31 @@ const isEvmAddress = (address: string): boolean => /^0x[0-9a-fA-F]{40}$/.test(ad
...
+// ANY route, so reject up front (a well-formed but merely WRONG non-burn recipient is the caller's
+// grounding responsibility). Routed through the canonical shared table so it can't drift from the other
+// guards: the EVM check now covers all three canonical burns (incl. the `0xdead…42069` variant it used to
+// miss) and the non-EVM check adds the Solana / UTXO / XRP family burn addresses (base58 it used to miss).
 const assertValidCustomRecipient = (recipient: string | undefined, from: AccountCoin, to: AccountCoin): void => {
   if (recipient === undefined) return
   if (isEvmAddress(recipient)) {
-    const lower = recipient.toLowerCase()
-    if (lower === ZERO_EVM_ADDRESS || lower === BURN_EVM_ADDRESS) {
+    const evmReason = getEvmDangerousReason(recipient)
```
NeOMakinG, line 535: `preferably-blocking:` `0X`-prefixed EVM burn slips the swap custom-recipient guard on native routes — Codex net-new, empirically confirmed. `assertValidCustomRecipient` gates the EVM burn check behind the local `isEvmAddress = /^0x[0-9a-fA-F]{40}$/` (no `/i`, lowercase `0x` only). The shared `getEvmDangerousReason` it then calls IS case-insensitive (`/i`) — but that path is unreachable for a `0X`-prefixed input, so a `0X…dEaD` recipient on a native (THOR/Maya, cross-chain) custom-recipient route falls through BOTH burn checks (`getChainDangerousReason(to.chain=EVM)` returns undefined) and is forwarded verbatim as `destination`. Probe: `guard('0X…dEaD','Ethereum') → PASSED-THROUGH` vs `guard('0x…dead','Ethereum') → BLOCKED`. The changeset explicitly claims `0X…` can't slip past — that protection is dead code behind this gate. Fund-loss direction (false-negative); exotic `0X` input + THOR downstream backstop keep it below blocker, but the advertised guard is defeated.

fix: add `/i` to the local `isEvmAddress` regex (or drop the pre-gate and call `getEvmDangerousReason` first, then `getChainDangerousReason`); add a `0X…`-prefixed burn test on a native route.

packages/sdk/src/tools/validate/recipientSanity.ts
```diff
@@ -52,56 +59,38 @@ export type RecipientSanityResult = {
...
-    if (/^0+$/.test(lower.slice(2))) return true
-  }
-  // EVM dead / burn address (0x...000dEaD) — case-insensitive.
-  if (lower === EVM_DEAD_BURN_ADDRESS) return true
-  // Solana System Program (base58 of 32 zero-bytes) — case-sensitive.
-  if (trimmed === SOLANA_SYSTEM_PROGRAM) return true
-  // Solana Incinerator burn address — case-sensitive base58.
-  if (trimmed === SOLANA_INCINERATOR) return true
+  if (isEvmBurnAddress(trimmed)) return true
+  if (trimmed in SOLANA_DANGEROUS_ADDRESSES) return true
```
NeOMakinG, line 91: `should-fix:` `key in map` / `map[key]` lets `toString`/`constructor`/`__proto__` false-positive as burns — New this round. `isNullAddress` uses `trimmed in SOLANA_DANGEROUS_ADDRESSES` (+ UTXO/XRP), and `getChainDangerousReason` uses bracket `MAP[dest]` — both read Object.prototype. Probe: `isNullAddress('toString'|'constructor'|'__proto__') → true`; `getChainDangerousReason('solana','toString') → truthy (a function) → assertSafeDestination throws`. This is over-block (false-POSITIVE) only — never a fund-loss false-negative, and no valid address equals a prototype key, so low practical impact — but it's a correctness defect on a fund guard and cheap to close.

fix: use `Object.hasOwn(MAP, key)` for the membership tests and `Object.hasOwn(MAP,dest) ? MAP[dest] : undefined` for the lookup (or null-prototype objects / a `Map`).

packages/core/chain/security/dangerousAddresses.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/** Return the chain-specific dangerous-address reason, if one applies. */
+export const getChainDangerousReason = (chain: string, address: string): string | undefined => {
+  const normalizedChain = chain.trim().toLowerCase()
+  const destination = address.trim()
+
+  if (normalizedChain === 'solana') {
+    return SOLANA_DANGEROUS_ADDRESSES[destination]
+  }
+  if (UTXO_CHAINS.has(normalizedChain)) {
```
NeOMakinG, line 96: `should-fix:` chain-keyed guard silently no-ops on non-canonical chain names (`xrp`, `SOL`, `BTC`) — Carried from R1 (partial). `getChainDangerousReason` matches only canonical lowercase `'ripple'`/`'solana'`/the UTXO set. Verified `getChainDangerousReason('xrp', <XRP black-hole>) → undefined`. In-repo callers are SAFE (findSwapQuote passes canonical `to.chain` enum; acrossQuote is EVM-only) so no live SDK bug — but the exported `assertSafeDestination`/`getChainDangerousReason` are the very surface app + agent-backend-ts are told to consume, and a ticker-style alias silently skips the guard. (isNullAddress is chain-agnostic so it still catches XRP.)

fix: map known aliases to families, or fail-closed / throw on an unrecognized non-EVM chain string instead of returning undefined.

packages/core/chain/security/dangerousAddresses.ts
```diff
@@ -0,0 +1,129 @@
...
+ * from escaping). */
+const EVM_ADDRESS_RE = /^0x[0-9a-fA-F]{40}$/i
+
+/**
+ * Return the dangerous-address reason string if `address` is a known EVM burn
+ * address, otherwise `undefined`. Shape-based: any `0x`+40-hex string is vetted
+ * against the EVM burn-list regardless of the chain it's destined for.
+ */
+export const getEvmDangerousReason = (address: string): string | undefined => {
+  if (!EVM_ADDRESS_RE.test(address)) return undefined
```
NeOMakinG, line 81: `should-fix:` `getEvmDangerousReason` still doesn't trim while its sibling `getChainDangerousReason` does — Carried from R1 (partial). `getEvmDangerousReason` tests the raw string against anchored `^0x…$`, so `' 0x…dead '` → undefined, while `getChainDangerousReason` trims internally — asymmetric. All in-repo callers pre-trim (findSwapQuote `normalizedRecipient`, acrossQuote `.trim()`, cctp viem `getAddress`, isNullAddress `addr.trim()`), so no live bypass today, but the exported primitive is less whitespace-robust than the guard it supersedes.

fix: trim inside `getEvmDangerousReason` for symmetry with `getChainDangerousReason`.

.changeset/reconcile-dangerous-address-guard.md
```diff
@@ -0,0 +1,42 @@
+---
+'@vultisig/core-chain': patch
```
NeOMakinG, line 2: `should-fix:` changeset is `patch` for an additive public-API surface — Carried from R1. Both `@vultisig/core-chain` and `@vultisig/sdk` are marked `patch`, but the SDK exposes ~9 new root named exports (assertSafeDestination, getEvmDangerousReason, …) plus a new core-chain subpath export — additive public API is `minor` under semver (repo precedent, sdk#1182).

fix: bump the sdk (and arguably core-chain) changeset to `minor`.

rcoderdev (approved):

NeOMakinG (approved):
Re-review at HEAD `18d41354` (round 3). Our prior two reviews were COMMENTED on stale commits (`e642a6b1`, `61a8b496`) while the PR was a draft; it is now ready with the one new commit `18d41354` that closes the RN-entry parity gap I flagged. Independently re-verified the reconcile is fund-safe and the export now actually reaches the RN consumer. **APPROVE** — the reconcile is a proven additive/tightening superset, the guard is non-inert (invoked in the CCTP + Across bridge primitives and the recipientSanity flag tool), and my one blocker-tier finding (RN parity) is resolved. Remaining items are pre-existing robustness gaps out of this PR's reconcile+export scope; captured below with a follow-up.

## testing approach

- Two lanes: Claude (load-bearing) + Codex gpt-5.4 (alive, 1 HIGH + 1 MED, both verified below).
- Read the actual canonical table (`packages/core/chain/security/dangerousAddresses.ts`), the shim (`packages/sdk/src/utils/dangerousAddresses.ts`), both entry re-exports (node `src/index.ts`, RN `src/platforms/react-native/index.ts`), and both in-repo consumers (`recipientSanity.isNullAddress`, `findSwapQuote.assertValidCustomRecipient`).
- Verified RN↔node export parity: both entries export the identical 9 symbols; the `./security/dangerousAddresses` subpath export map exists in core-chain's `package.json`, so the RN re-export chain resolves (non-inert).
- Ran `recipientSanity.test.ts` → 29/29 pass. (`dangerousAddresses.test.ts` fails to *load* in my worktree due to an unbuilt `@vultisig/core-config` dist pulled transitively via a cardano barrel import — environmental, not a logic failure.)
- Runtime probe (esbuild-transpiled HEAD table) confirmed: superset holds (SPL Token Program / wSOL mint / BTC eater / XRPL ACCOUNT_ONE all now blocked — the previously-missed burns), and reproduced the 0X-slip + alias-miss + prototype-footgun behaviors below.
- Confirmed via `git blame` that the case-sensitive `isEvmAddress` gate leaking `0X` predates this PR (introduced by `5926f3be`, present on `origin/main`) — so it is not a regression here.

## prior issues — status

| R1/R2 | 1 | preferably-blocking | RN-entry parity: guard exported from node/browser/electron entry but not `react-native/index.ts`, so the app (the RN consumer this PR targets) never got the single source of truth | ✅ fixed | `18d41354` re-exports all 9 symbols from the RN entry; byte-parity with node entry verified; resolves through the core-chain security subpath export |
| R2 | 2 | preferably-blocking | `0X`-prefix (capital X) EVM burn slips `findSwapQuote` native custom-recipient route | ❌ still open | runtime-reproduced at HEAD (`0Xdead…42069` → PASS-THROUGH). **But pre-existing** — the case-sensitive `isEvmAddress` gate (line 511) comes from `5926f3be` on main, not this PR. Not a regression. Changeset (lines 41-42) still over-claims it's closed. Follow-up below |
| R1 | 3 | should-fix | `getChainDangerousReason` matches `'ripple'`, misses `'xrp'` chain string | ⚠️ still open | Codex extends: `'bitcoincash'` alias also misses (only `'bitcoin-cash'` matches). No live bug — every in-repo caller passes the canonical `Chain` enum |
| R1 | 4 | should-fix | `getEvmDangerousReason` no `.trim()` → whitespace-padded EVM burn bypasses the primitive | ⚠️ still open | no live bypass — every caller pre-trims (findSwapQuote `normalizedRecipient`, isNullAddress, cctp viem, acrossQuote) |
| R2 | 5 | should-fix | prototype footgun: `in` / bracket access on the `Record` maps reads `Object.prototype` | ⚠️ still open | over-block only, never fund-loss (`assertSafeDestination('solana','toString')` throws). No valid address is a proto key |
| R1/R2 | 6 | should-fix | changeset `patch` → should be `minor` (9 new public named exports = additive API surface) | ❌ still open | still `patch`/`patch` at HEAD |
| R1→R2 | 7 | (resolved R2) | two divergent burn-checkers on root (isNullAddress vs new guard) | ✅ fixed | unchanged since R2; both consume the one table |

## should-fix

- **guard not enforced on the plain send/sign path (Codex HIGH, verified — pre-existing, out of scope).** `prepareSendTxFromKeys` (`packages/sdk/src/tools/prep/send.ts:52`) → `buildSendKeysignPayload` and `TransactionBuilder.prepareSendTx` only call `isValidAddress`, never `assertSafeDestination`. The only real guard invocations are `acrossQuote.ts:263` and `buildCctpBridge.ts:259` (both EVM bridge/swap) plus the recipientSanity advisory flag. So a plain send to `0x000…0`, `0x…dead`, the Solana incinerator, the BTC eater, or the XRPL black-hole is *not* blocked at build/sign time. Confirmed this gap is pre-existing (`assertSafeDestination` absent from the send/keysign path on `origin/main` too). This PR's scope is reconcile+export — the export exists precisely so the send path can be wired next — but the table docstring's "call this in every primitive that encodes a destination … BEFORE building the calldata" over-states current coverage. Recommend a follow-up that wires `assertSafeDestination(coin.chain, receiver)` into `prepareSendTxFromKeys`, and a docstring note that plain-send enforcement is pending.

- **`0X`-prefix slip + changeset over-claim (finding #2 above).** The changeset states "a `0X…`-prefixed burn can't slip past (parity with the Go guard)." True of the shared table's `getEvmDangerousReason` (its `EVM_ADDRESS_RE` has `/i`), but the `findSwapQuote` local `isEvmAddress = /^0x[0-9a-fA-F]{40}$/` gate (no `/i`, from `5926f3be`) sits in front of it, so on a native THOR/Maya cross-chain custom-recipient route `0Xdead…42069` never reaches the case-insensitive check → passes through. One-token fix: add `/i` to the local gate (makes the changeset claim true). I'll open a follow-up PR for this rather than push to your branch mid-approval.

- **chain-alias misses (finding #3, Codex MED).** `getChainDangerousReason` recognizes only canonical `'ripple'` / `'bitcoin-cash'`; the SDK's own `normalizeChain`/`addressFormat` surface accepts `'xrp'`, `'xrpl'`, `'bitcoincash'`. A future export consumer that passes an alias string directly would silently miss BCH/XRPL burns. No in-repo bug (callers use the enum). Fix: normalize the chain arg (or key the tables by normalized chain) inside `getChainDangerousReason`.

- **changeset bump (finding #6).** Bump `@vultisig/sdk` (and `@vultisig/core-chain`) to `minor` — 9 new public named exports is additive public-API surface, not a patch.

## risk

Low for this PR. The reconcile is strictly additive/tightening (runtime-proven superset — no address any prior copy blocked is lost, and 7 previously-missed burns now flag) so it cannot introduce a new false-negative fund-loss path. The RN export is a pure static string-matching re-export with no chain-client deps, safe on the RN bundle. Every remaining finding is either a pre-existing gap (0X slip, plain-send non-enforcement) or a false-*positive*-only / consumer-robustness item (aliases, prototype, trim) — none is a fund-loss regression from this change. The residual real gap (0X on native swap route) is exotic input with a THOR/Maya downstream backstop; tracked in the follow-up.

## verdict

**APPROVE.** RN-entry parity — my only blocker-tier finding across three rounds — is resolved at `18d41354`, the reconcile is a fund-safe additive superset, and the guard is genuinely invoked (not dead-exported) in the bridge/swap primitives. The remaining should-fixes are pre-existing robustness gaps outside this PR's reconcile+export scope; I'm opening a follow-up PR for the `0X` `/i` fix and noting the plain-send wiring + alias normalization + minor-bump as follow-up work.

NeOMakinG (approved):
R4 at `46c7a7ea`. The new tip is gomes' `Merge remote-tracking branch 'origin/main'` whose message records a hand-resolved conflict in `findSwapQuote.recipient.test.ts` — the test surface guarding the very check this PR exports. So this round is entirely a **merge-integrity audit**: did the resolution lose main's work, or loosen an assertion while the suite still reports green? It did neither. Re-approving, bound to the merged head.

## testing approach

Two lanes: Claude (load-bearing) + Codex gpt-5.4 (ALIVE, returned no net-new findings in the merge-integrity scope). Worktree at `46c7a7ea`, `node_modules` symlinked from the canonical checkout; the R3 "`dangerousAddresses.test` fails to LOAD" environmental blocker was just an unbuilt `core-config` dist — symlinking it makes the suite run, so this round has real numbers for it.

**1. Net-diff-vs-merge-base stability.** Not "the last hop was merge-only" — the actual check. Compared this PR's net diff at the old approved head (`18d41354` vs its base `7f962dc3`) against the net diff at the merged head (`46c7a7ea` vs merge parent `358c27ba`), per file, on the `+`/`-` line sets only (ignoring context shift):

| file | +/- lines vs base |
|---|---|
| `.changeset/reconcile-dangerous-address-guard.md` | identical |
| `packages/core/chain/package.json` | identical |
| `packages/core/chain/security/dangerousAddresses.ts` | identical |
| `packages/core/chain/swap/quote/findSwapQuote.ts` | identical |
| `packages/sdk/src/index.ts` | identical |
| `packages/sdk/src/platforms/react-native/index.ts` | identical |
| `packages/sdk/src/tools/validate/recipientSanity.ts` | identical |
| `packages/sdk/src/utils/dangerousAddresses.ts` | identical |
| `packages/sdk/tests/unit/tools/validate/recipientSanity.test.ts` | identical |
| `packages/sdk/tests/unit/utils/dangerousAddresses.test.ts` | identical |
| `packages/core/chain/swap/quote/findSwapQuote.recipient.test.ts` | **+4/-1** (the resolution — detailed below) |

Nothing smuggled in: the merged head is exactly `main + this PR's original delta`, with one 4-line exception.

**2. Main-side loss check.** Main touched 5 of the 11 PR files between the merge-base and the merge parent (`package.json` +30, `findSwapQuote.recipient.test.ts` +24, `findSwapQuote.ts` +7, `src/index.ts` +21, `react-native/index.ts` +29). Because the PR's `+`/`-` set relative to `358c27ba` is byte-identical to its set relative to a base that did *not* contain main's lines, **every main-side hunk survives** — a dropped main line would have shown up as an extra `-` in the new diff, and there are none. Spot-confirmed the load-bearing one: `#1399`'s case-insensitive gate is intact at `findSwapQuote.ts:516` (`/^0x[0-9a-fA-F]{40}$/i`), and main's `#1239`/`#1282` export additions to both entrypoints are present with **no duplicate identifiers** in either `src/index.ts` or `platforms/react-native/index.ts` (all 9 guard symbols appear exactly once in each).

**3. The resolution itself.** The only semantic delta in the whole merge is inside main's `#1399` regression test:

```diff
-      message: expect.stringContaining('zero/burn address'),
+      // Message format is now sourced from the shared dangerousAddresses table (#1193),
+      // which always ends with "... — the swap output would be unrecoverable."
+      message: expect.stringContaining('unrecoverable'),
```

Every other line of main's 24-line `0X`-prefix test is byte-preserved, including the `.rejects` shape, the `SwapErrorCode.InvalidConfig` code assertion and the `expect(getNativeSwapQuote).not.toHaveBeenCalled()` no-provider check. This was **necessary, not lazy** — proven by mutation M7 below.

**4. Mutation matrix — 8 mutations, 8 RED, 0 green.** The point of the matrix is to find assertions that survive a broken guard; none did.

| # | mutation | result |
|---|---|---|
| M1 | revert `#1399`: drop `/i` from `isEvmAddress` (`findSwapQuote.ts:516`) | **RED** — 1 failed: the `0X`-prefix test. The resolution kept it **red-on-revert**. |
| M2 | strip the word `unrecoverable` from both throw sites in `findSwapQuote.ts` | **RED** — 3 failed (zero address, Incinerator, wSOL). The assertion still binds to the guard's message frame. |
| M3 | delete `0xdead…42069` from `EVM_DANGEROUS_ADDRESSES` | **RED** — 1 failed (that `it.each` case) |
| M4 | drop `/i` from the table's `EVM_ADDRESS_RE` | **RED** — 1 failed (the `0X` test) |
| M5 | delete the Solana Incinerator + wSOL entries | **RED** — 2 failed |
| M6 | drop `.toLowerCase()` in `getEvmDangerousReason` | **RED** — 3 failed (both checksummed burns + `0X…dEaD`) |
| M7 | restore main's original `stringContaining('zero/burn address')` | **RED** — proves the loosening was **forced** by the PR's genuinely-changed message frame, not a convenience weakening |
| M8 | neuter the non-EVM `chainReason` branch in `assertValidCustomRecipient` | **RED** — 2 failed |

One honest caveat on M2: 3 of the 6 `unrecoverable`-asserting tests stay green under it, because those three table *reason* strings ("funds are unrecoverable") contain the word themselves. So `'unrecoverable'` is a slightly weaker binding than main's `'zero/burn address'` was. It does not matter for what the test exists to prove — M1 shows the test is still red the instant the actual `0X` fix is reverted, and M4/M6 show it is red if the table's matching is broken. Documented, not a finding.

**5. Suites + CI.** `findSwapQuote.recipient.test.ts` 15/15 green; `dangerousAddresses.test.ts` + `recipientSanity.test.ts` 49/49 green (both now actually load). `prettier --check` clean on all 7 touched source files. CI at `46c7a7ea`: Lint and Type Check ✅, Integration Tests ✅, Unit Tests still in progress at time of writing, CodeRabbit "fail" is `Review rate limited`, not a real failure. `git merge-tree` against current `origin/main` (`20a0fd0d`) is conflict-free; `mergeStateStatus: BEHIND` is just the branch trailing main.

## prior issues — status

| round | severity | issue | status | notes |
|---|---|---|---|---|
| R1/R2 | preferably-blocking | RN-entry export parity — guard undefined on the platform the app resolves | ✅ fixed | `@18d41354`; all 9 symbols re-exported, still present and un-duplicated after the merge |
| R2 | preferably-blocking | `0X…dEaD` slipped the case-sensitive `isEvmAddress` gate on native cross-chain routes | ✅ fixed | via my follow-up `#1399`, now on main and pulled in by this merge. Verified live at `findSwapQuote.ts:516` and **red-on-revert** (M1) |
| R3 | should-fix | changeset over-claims "a `0X…`-prefixed burn can't slip past" | ✅ now true | the claim was false at R3 (local gate was case-sensitive); post-merge it holds. Credit belongs to `#1399` rather than this diff, but the sentence is no longer wrong |
| R1 | should-fix | `getChainDangerousReason` matches only canonical `'ripple'`/`'bitcoin-cash'`, not the `xrp`/`bitcoincash` aliases `normalizeChain` accepts | ⚠️ still open | table is byte-stable across the merge. No live SDK bug — every in-repo caller passes a canonical `Chain` enum. Export-consumer robustness only |
| R1 | should-fix | `getEvmDangerousReason` does not `.trim()`, so a whitespace-padded EVM burn is not matched | ⚠️ still open | no live bypass: all four callers pre-trim. Primitive-symmetry nit |
| R2 | should-fix | `Object.prototype` keys readable through the bracket lookups (`'toString'`, `'__proto__'`) | ⚠️ still open | over-block only (false-positive), never fund-loss |
| R3 | should-fix | plain send/sign path (`prepareSendTxFromKeys`) never calls `assertSafeDestination` | ⚠️ still open | confirmed at the merged head: the only production callers are `acrossQuote.ts:263`, `buildCctpBridge.ts:259`, `findSwapQuote.ts:535/545` and the advisory `recipientSanity`. Pre-existing and out of this PR's reconcile+export scope — the export now exists so the send path *can* be wired |
| R1/R2/R3 | should-fix | changeset should be `minor`, not `patch`, for 9 new public exports | ℹ️ **retracted — I was wrong** | checked sibling precedent in this repo instead of the prose rule: `#1239` (new root + RN exports), `#1282` (new root + RN exports) and `#1252` (new published subpaths) all shipped as `patch`. `patch` is the house convention here. Withdrawing this across all three prior rounds |

## should-fix

**The drift class this PR exists to kill has already re-appeared on main — a third private EVM burn list, at the signing choke point.** `packages/core/chain/swap/general/cowswap/sign/buildCowSwapOrderTypedData.ts:89-90` (landed on main via `#1427`, after my R3, and pulled in by this merge) declares its own two-entry list:

```ts
const ZERO_EVM_ADDRESS = '0x0000000000000000000000000000000000000000'
const BURN_EVM_ADDRESS = '0x000000000000000000000000000000000000dead'
```

`assertCowSwapReceiverSafe` then compares `receiver.toLowerCase()` against exactly those two. It **misses `0xdead000000000000000042069420694206942069`** — the third canonical burn, the one whose omission from the CCTP inline copy this table's own header documents as the originating incident. So a hand-built `cowswap-order:` payload with `receiver = 0xdead…42069` passes the digest-construction guard and gets signed into an unrecoverable order, even though this PR's upgraded quote-time guard now rejects that exact address.

Not blocking here: it is main-side, it is not in this diff, and this PR is strictly an improvement over it. But it is a fund-direction false-negative at the choke point that binds what the signature actually covers, which is the strongest guard position in the flow. Fix is one line:

```ts
import { isEvmBurnAddress } from '@vultisig/core-chain/security/dangerousAddresses'
// ...
if (isEvmBurnAddress(receiver)) { /* throw */ }
```

Worth a follow-up — and worth taking as evidence that the "single source of truth" this PR builds needs the remaining private copies routed through it promptly, or it drifts again.

## risk

Low for the merge itself. The reconcile half remains a proven additive/tightening superset (table byte-stable since R3, where I probe-verified that 7 previously-missed burns now flag and no address was dropped); the export half reaches both the node and React Native entries; the one conflict resolution preserves main's regression test in full and is red-on-revert. No mutation of the guard survived the suite. The residual findings are all pre-existing robustness items with no live in-repo bypass, plus one main-side follow-up outside this diff.

## verdict

**APPROVE** at `46c7a7ea`. The merge is clean: no main-side work lost, no assertion silently weakened, no smuggled content. Re-approving explicitly at the merged head rather than resting on the `18d41354` approval.

<sub>reviewed by an AI reviewer agent (Claude + Codex, two-lane) — merge-integrity round</sub>

Issue #1147, opened 2026-07-10
mpc: empty-string amount silently zero-builds in resolvers untouched by #1140 (tron TRC20, solana send, polkadot, bittensor, ripple issued-currency)


gomesalexandre:
Follow-up from #1138/#1140, surfaced during #1140's final review. #1140 fixed the empty-string-fail-open class (\`BigInt('') -> 0n\`, silently building a zero-amount transfer from an unset proto3 string) for the resolvers it touched (sui, tron transfer, ripple payment, ton native, cardano transfer via \`toBoundedLong\`) — but the same underlying \`BigInt(toAmount)\` pattern exists UNGUARDED in other resolvers that call \`BigInt()\` directly rather than through \`toBoundedLong\`:

- \`resolvers/tron.ts:213,278\` (TRC20 path — separate from the transfer-amount path #1140 already fixed)
- solana send resolver
- polkadot resolver
- bittensor resolver
- \`resolvers/ripple.ts:41\` (issued-currency path — separate from the native XRP payment path #1140 already fixed)

Same fix pattern applies: route these through \`toBoundedLong\` (now available, from #1140) instead of a raw \`BigInt()\` call, so an unset/malformed amount throws instead of silently building a zero-value transfer.

Severity: MED (real gap, same class as the highest-severity finding of the round-4 audit, just in resolvers that weren't in that finding's original scope). Effort: S — mechanical swap to the now-existing shared helper, one PR can likely cover all 5 sites.

realpaaao:
TAO epic vultisig-sdk#2333 — empty `toAmount` → 0n allow-death is still live on the Bittensor signing inputs path. Keep this issue; do not duplicate.

Issue #1138, opened 2026-07-10
mpc: Long.fromString silently wraps out-of-range transfer amounts instead of failing closed (Sui/Tron/Ripple/TON/Cardano)


gomesalexandre:
The `long` package's `Long.fromString` has no magnitude bound check — a digit string too large for 64 bits silently two's-complement wraps rather than throwing (confirmed by reading `node_modules/long/umd/index.js:344-385`). Unguarded on fund-relevant amount fields at `resolvers/sui.ts:70,82` (Sui's ~1e19 MIST supply exceeds signed-64-bit's ~9.223e18 ceiling), `resolvers/tron.ts:124,146` (dApp-supplied `contractPayload`, no upper bound at all), plus `ripple.ts`, `ton/native.ts`, `cardano.ts`. Same bug class as the historical osmosis poolId=2^64→0 incident, here on the actual transfer amount — the MPC co-signers can sign over a different value than the UI reported.

Fix sketch + full agent-ready prompt: [round-4 bug-hunt artifact](https://claude.ai/code/artifact/4ac4b9f5-2a5d-4b4e-ab54-b5f693a2c973), sdk finding #5. Draft plan already written: `plans/014-bounded-long-fromstring.md`.

Severity: HIGH. Effort: S.

PR #1197, opened 2026-07-12
fix(encoding): reject empty/malformed amounts in the 5 resolvers #1140 missed

gomesalexandre:
Closes vultisig/vultisig-sdk#1147

> Stacked on #1140 (base is `advisor/014-bounded-long-fromstring`, since `toBoundedLong` only exists there - #1147's "now-merged #1140" premise was slightly ahead of reality). GitHub will retarget to `main` once #1140 lands. Only the last 1 commit is this PR.

## what

#1140 closed the `BigInt('') -> 0n` fail-open (proto3 defaults an unset `toAmount` to `''`, so a bare `BigInt()` silently builds a zero-amount transfer that gets co-signed) for the resolvers it touched. This finishes the sweep for the 5 sites it missed - all verified still live before patching:

| site | target field | bound |
|---|---|---|
| `resolvers/tron.ts:213,278` (TRC20) | `transfer(address,uint256)` calldata bytes | u256 |
| `resolvers/polkadot.ts:24` | u128 balance bytes | u128 |
| `resolvers/bittensor.ts:66` | SCALE-compact (subtensor `Balance` = u64) | u64 |
| `resolvers/solana/send.ts:34` | proto `uint64` value/amount | u64 via `toBoundedLong` |
| `resolvers/ripple.ts:41` (issued currency) | decimal-string trust-line limit | u128 |

Plus one more of the exact same class found whilst in the house: ripple's raw-JSON memo payment path inlined `toAmount` unvalidated into the tx JSON - now guarded like its sibling Payment proto path.

## how

New `toBoundedBigInt` (`@vultisig/lib-utils/bigint/toBoundedBigInt`) - the bigint-returning sibling of `toBoundedLong`, which now delegates to it (one source of truth for the strict base-10 regex + bounds). Bit width is per target field on purpose: a blind 64-bit bound on TRC20 would false-reject 20 tokens of an 18-decimal TRC20 (2e19 > 2^64). Signedness read from each encoder, not copied from #1140.

Bonus the sweep surfaced: pre-fix solana silently encoded a u64-max amount as `'-1'` (`Long.fromString` defaults to signed) - the `{ unsigned: true }` routing fixes that too, pinned by test.

## testing

Fail-before (resolver sources reverted to base, new tests kept):

```
Test Files  5 failed (5)
     Tests  12 failed | 51 passed (63)
```

(the 5 pre-passing ones are the no-false-reject happy paths - they're supposed to pass on both sides)

Pass-after:

```
resolvers + bigint utils:  Test Files 15 passed (15) / Tests 129 passed (129)
yarn workspace @vultisig/sdk typecheck: pass
yarn build:all && yarn typecheck (full chain): exit 0, 0 TS errors
yarn lint: pass
yarn format:check: all matched files use Prettier code style
yarn check:shared-exports: up to date
yarn knip: exit 0, zero delta vs base
```

Per-resolver regression tests: empty-string amount throws `RangeError` (not zero-builds), negative throws, plus a large-amount happy path per resolver proving no false-reject (TRC20 > 2^64 builds the exact expected calldata bytes, polkadot 2^64 builds the exact value bytes, bittensor u64-max SCALE-encodes, ripple IOU 2^64 formats, solana u64-max stays `18446744073709551615`).

Changeset included (`@vultisig/sdk` patch).

## risk

Low - pure tightening: every previously-valid amount still builds byte-identical inputs (pinned by the happy-path tests); only `''`/negative/malformed/out-of-field-range amounts now throw instead of being silently co-signed as zero.

🤖 Agent-generated

### DIFF
--- a/packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts
@@ -5,6 +5,7 @@ import {
   toXrplCurrencyCode,
 } from '@vultisig/core-chain/chains/ripple/issuedCurrency'
 import { assertField } from '@vultisig/lib-utils/record/assertField'
+import { toBoundedBigInt } from '@vultisig/lib-utils/bigint/toBoundedBigInt'
 import { toBoundedLong } from '@vultisig/lib-utils/bigint/toBoundedLong'
 import { TW } from '@trustwallet/wallet-core'
 import Long from 'long'
@@ -38,7 +39,10 @@ export const getRippleSigningInputs: SigningInputsResolver<'ripple'> = ({ keysig
         limitAmount: TW.Ripple.Proto.CurrencyAmount.create({
           currency: toXrplCurrencyCode(currency),
           issuer,
-          value: formatIssuedCurrencyValue(BigInt(keysignPayload.toAmount), coin.decimals),
+          value: formatIssuedCurrencyValue(
+            toBoundedBigInt(keysignPayload.toAmount, { bits: 128, signed: false }),
+            coin.decimals
+          ),
         }),
       }),
     }
@@ -65,7 +69,7 @@ export const getRippleSigningInputs: SigningInputsResolver<'ripple'> = ({ keysig
           TransactionType: 'Payment',
           Account: account,
           Destination: keysignPayload.toAddress,
-          Amount: keysignPayload.toAmount,
+          Amount: toBoundedLong(keysignPayload.toAmount, { unsigned: false }).toString(),
           Fee: gas.toString(),
           Sequence: Number(sequence),
           LastLedgerSequence: Number(lastLedgerSequence),
--- a/packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts
@@ -69,7 +69,11 @@ export const getRippleSigningInputs: SigningInputsResolver<'ripple'> = ({ keysig
           TransactionType: 'Payment',
           Account: account,
           Destination: keysignPayload.toAddress,
-          Amount: toBoundedLong(keysignPayload.toAmount, { unsigned: false }).toString(),
+          // Unlike the two Payment PROTO paths (int64 field, so signedness must
+          // match), this is a plain JSON string: XRP drops are non-negative, so
+          // reject a negative amount here instead of building JSON that XRPL
+          // would bounce after the MPC ceremony already ran.
+          Amount: toBoundedLong(keysignPayload.toAmount, { unsigned: true }).toString(),
           Fee: gas.toString(),
           Sequence: Number(sequence),
           LastLedgerSequence: Number(lastLedgerSequence),
--- a/.changeset/bounded-bigint-residual-resolvers.md
+++ b/.changeset/bounded-bigint-residual-resolvers.md
@@ -1,0 +1,24 @@
+---
+'@vultisig/sdk': patch
+---
+
+fix(encoding): reject empty/malformed amounts in the 5 resolvers #1140 missed
+
+#1140 closed the `BigInt('') -> 0n` fail-open class (proto3 defaults an unset
+`toAmount` string to `''`, so a bare `BigInt()` silently builds a zero-amount
+transfer) for the resolvers whose amounts feed proto 64-bit fields via
+`toBoundedLong`. The same unguarded `BigInt(toAmount)` pattern remained in five
+resolvers whose amounts are NOT proto 64-bit integers: Tron TRC20 (uint256
+calldata bytes), Polkadot (u128 balance bytes), Bittensor (SCALE-compact u64),
+Solana send (uint64, was still raw `BigInt` + `Long.fromString`) and Ripple
+issued-currency trust lines (decimal-string value).
+
+Adds `toBoundedBigInt` (`@vultisig/lib-utils/bigint/toBoundedBigInt`), the
+bigint-returning sibling of `toBoundedLong` (which now delegates to it), with an
+explicit bit width per target field so wide fields don't false-reject: Tron
+TRC20 bounds at uint256, Polkadot at u128, Bittensor at u64 (subtensor `Balance`
+is u64), Ripple issued-currency at u128, and Solana send routes through
+`toBoundedLong { unsigned: true }`. Ripple's raw-JSON memo payment path, which
+inlined `toAmount` unvalidated into the tx JSON, is guarded the same way as its
+sibling Payment proto path. An unset/negative/non-decimal amount now throws a
+`RangeError` instead of being silently co-signed as a zero-value transfer.
--- a/packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/bittensor.test.ts
@@ -231,3 +231,33 @@ describe('getBittensorSigningInputs — custom tx-input framing round-trips', ()
     expect(hex(decoded.payload)).toBe(hex(payload))
   })
 })
+
+describe('getBittensorSigningInputs -- amount strict parse (#1147)', () => {
+  let walletCore: WalletCore
+
+  beforeAll(async () => {
+    walletCore = await initWasm()
+  })
+
+  it("rejects an unset ('') amount instead of building a zero-value transfer", () => {
+    const payload = buildPayload()
+    payload.toAmount = ''
+    // Pre-fix: BigInt('') -> 0n -> compactEncode(0n) silently co-signed.
+    expect(() => getBittensorSigningInputs({ keysignPayload: payload, walletCore })).toThrow(RangeError)
+  })
+
+  it('rejects an amount above the u64 balance range (subtensor Balance is u64)', () => {
+    const payload = buildPayload()
+    payload.toAmount = (2n ** 64n).toString()
+    expect(() => getBittensorSigningInputs({ keysignPayload: payload, walletCore })).toThrow(RangeError)
+  })
+
+  it('still builds a transfer at the u64 max (no false-reject)', () => {
+    const payload = buildPayload()
+    payload.toAmount = (2n ** 64n - 1n).toString()
+    const [txInputData] = getBittensorSigningInputs({ keysignPayload: payload, walletCore })
+    const { callData } = decodeBittensorTxInput(txInputData)
+    // callData ends with the SCALE-compact-encoded amount.
+    expect(hex(callData).endsWith(hex(compactEncode(2n ** 64n - 1n)))).toBe(true)
+  })
+})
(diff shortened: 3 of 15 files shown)


NeOMakinG (commented):
Follow-up to #1140 (stacked on `advisor/014-bounded-long-fromstring`). This finishes the `BigInt('') -> 0n` fail-open sweep for the 5 resolvers #1140 missed and — importantly — **closes the exact Solana `send.ts` gap I flagged as preferably-blocking on #1140** while also fixing a latent signedness bug. Reviewing as **COMMENTED** (draft; never approve). Two-lane run; Codex lane was infra-dead (0-byte), so Claude is load-bearing.

## testing approach
Built the stack in an isolated worktree at HEAD `b145d3d3` (stacked-branch plumbing: point `@vultisig/{lib-utils,core-mpc,core-chain}` at the worktree, overlay the changed `.ts` into dist via esbuild since canonical main dist lacks the newer subpaths). Ran the real vitest suites, then proved non-inert via red-on-revert.

- **79 green at HEAD**: 16 util (`toBoundedBigInt` 6 + `toBoundedLong` 10) + 63 across all 5 resolver suites (bittensor/polkadot/ripple/solana/tron).
- **Red-on-revert PROVEN**: neutered both helpers to pass-through (`BigInt(value)` / `Long.fromString`) → **14 reject tests RED** across every resolver + the #1138 tron-native cases; happy-paths stayed green. Restored clean.
- **Signedness/bit-width verified against the actual encoders** (not just PR prose): `compactEncode` is generic big-int (the u64 cap is purely the guard; subtensor `Balance` is u64 and TAO max supply ~2.1e16 rao ≪ u64, so the u64 bound cannot false-reject a real balance either way); polkadot u128 happy-path builds `010000000000000000`; ripple IOU `formatIssuedCurrencyValue(bigint, 15)` bounded u128; ripple raw-JSON XRP path `int64` (matches the proto Payment sibling verified on #1140); solana u64.
- **Completeness**: grepped every resolver for surviving `BigInt(...toAmount)` / `Long.fromString(...toAmount)`. All 5 named resolvers' `toAmount`→signing-input paths are now bounded. The #1140 Solana `send.ts:43/56` gap is closed at `:47/:60`, and the latent signedness bug (pre-fix `Long.fromString` defaulted **signed**, so a u64-max SPL amount encoded as `-1`) is fixed by `{ unsigned: true }`, pinned by the u64-max happy-path test. Legit amounts `< 2^63` remain byte-identical.

The core work is correct, correctly-scoped, and well-tested. Nothing blocking.

## should-fix
**Cardano token bundle is the last surviving instance of the same `''→0n` class this PR set out to eliminate.** `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts:40` still does `amount: amountToBytes(BigInt(keysignPayload.toAmount))` — the exact `BigInt('') -> 0n` pattern. It is **defended-in-depth today**: for a token send the same `input.create` always runs `cardano.ts:51 toBoundedLong(toAmount, { unsigned: true })`, which throws `RangeError` on `''` before the input returns — so there's no live empty-string hole. But the defense is incidental/ordering-dependent: if that line-51 native guard is ever refactored, the token path silently re-opens the zero-value co-sign. Cardano is out of this PR's declared 5-resolver scope (handled by #1140 + #1144), so deferring is defensible, but a direct `toBoundedBigInt(..., { bits: 64, signed: false })` on line 40 (or at minimum a comment noting the dependency on line 51) would actually finish the sweep. (Not a line this PR touches → left in the body.)

## q
- `solana/send.ts:33` — `BigInt(priorityFee)` is still raw and reaches `Long.fromString` at `:97`, so a `> u64` priority fee could wrap. It's a fee (not the transfer amount, no send-amount fund loss), server-derived, and pre-existing — flagging only for awareness, not asking you to fix it here.
- `ripple.ts:43` — why `u128` specifically for the IOU trust-line limit? XRPL issued-currency values are a wire decimal-float, not a u128 field, so the bound is a chosen ceiling rather than the field width. It's fail-safe (an over-tight ceiling only rejects absurd `> 3.4e38` base-unit limits), just curious about the rationale vs., say, the XRPL 16-significant-digit envelope.
- cosmos amount paths (`cosmos/coinAmount.ts:14`, `cosmos/index.ts:295`) pass the raw `toAmount` string through — I believe these lean on the `varintBig` guard your `toBoundedLong` doc references; confirming that's the intended coverage and cosmos is deliberately outside this sweep.

## risk
Low. Pure tightening: every previously-valid amount still builds byte-identical signing inputs (pinned by the happy-path tests I re-ran); only `''`/negative/malformed/out-of-field-range now throws at signing-input construction (fail-closed, before any keysign), instead of being silently co-signed as a zero or wrapped value. The `toBoundedLong -> toBoundedBigInt` delegation is behavior-preserving (verified byte-identical, no double-bound). The one residual (`cardano.ts:40`) is defended today.

## verdict
COMMENTED — draft, not approving. Clean, correctly-scoped fund-safety fix that closes the #1140 Solana gap correctly (and fixes a latent u64 signedness bug as a bonus). One should-fix (defended-in-depth cardano residual) and a few out-of-scope questions, none blocking.

---
🤖 Two-lane review (Claude load-bearing; Codex lane infra-dead this run). Runtime evidence: worktree vitest 79-green + 14 red-on-revert.

packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts
```diff
@@ -65,7 +69,11 @@ export const getRippleSigningInputs: SigningInputsResolver<'ripple'> = ({ keysig
           TransactionType: 'Payment',
           Account: account,
           Destination: keysignPayload.toAddress,
-          Amount: keysignPayload.toAmount,
+          // Unlike the two Payment PROTO paths (int64 field, so signedness must
+          // match), this is a plain JSON string: XRP drops are non-negative, so
+          // reject a negative amount here instead of building JSON that XRPL
+          // would bounce after the MPC ceremony already ran.
+          Amount: toBoundedLong(keysignPayload.toAmount, { unsigned: true }).toString(),
```
NeOMakinG, line 76: `q:` correct to reject negatives on this JSON-string path. But note the flip side of your own comment: the two proto `OperationPayment.amount` sites (`:58` valid-dest-tag, `:98` no-memo) stay `unsigned: false`, so `toAmount = '-1'` still builds `opPayment.amount = -1` with no throw — a ceremony then runs on a tx XRPL bounces post-sign, the same case you avoid here. Since XRP max drops (~1e17) sit far below `2^63`, those paths could also take `unsigned: true` (the `2^63..2^64-1` mis-encode range is unreachable for real XRP), or add a `>= 0n` check, to make pre-ceremony rejection consistent across all three paths. Non-blocking (negative = malformed, ceremony-waste not fund loss).

NeOMakinG (commented):
Round 2 re-review of the single follow-up commit `1cc5b904` — hardens the XRP ripple raw-JSON memo payment path against negative amounts (codex-review follow-up). DRAFT, so **COMMENTED, not approved.**

## testing approach

Worktree at `1cc5b904`, vitest via `.config/vitest.core.config.ts` (aliases resolve to source — no dist build). Ran the ripple suite: **11/11 green** at HEAD. Reverted the one-line fix (`unsigned: true` -> `unsigned: false` on the raw-JSON path only) -> the new negative-amount test **fails** (`expected function to throw ... but it didn't`), other 10 stay green -> the +14 test is **non-inert / load-bearing**. Separately probed both proto Payment paths with a negative amount to characterise the asymmetry below. Codex gpt-5.4 lane INFRA-DEAD (SIGTERM at 300s, 0-byte) — Claude load-bearing.

## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | should-fix | `cardano.ts:40` token `amountToBytes(BigInt(toAmount))` — last surviving `''->0n` instance of the class this PR targets | ❌ still open | This commit is ripple-only; `cardano.ts` untouched. Still defended-in-depth by the always-run native `toBoundedLong` guard at `cardano.ts:51` (verified present at HEAD) -> no live empty-string hole, but the defense remains incidental/ordering-dependent. |
| r1 | 2 | q | `solana/send.ts` `BigInt(priorityFee)` raw wrap (fee, not send-amount) | ⏭️ deferred | Out of the declared 5-resolver scope; pre-existing; fee not transfer amount. |
| r1 | 3 | q | cosmos raw `toAmount` string paths | ⏭️ deferred | Not in the 5; relies on downstream varintBig guard. |
| r1 | 4 | q | ripple IOU (TrustSet) u128 bound is arbitrary vs XRPL decimal-float | ⏭️ deferred | Fail-safe over-tight ceiling; unchanged. |

Note: r1 characterised the raw-JSON path as `unsigned: false` *matching the proto Payment sibling (int64)*. This commit deliberately **diverges** it to `unsigned: true` — correct for a plain JSON drops string (see should-fix below for the flip side).

## should-fix

**Asymmetric guard: the two proto Payment paths still accept a negative amount.** The fix rejects a negative on the raw-JSON path (`:76`, `unsigned: true`) but both proto `OperationPayment.amount` sites remain `unsigned: false` (signed int64) — `:58` (valid-dest-tag) and `:98` (no-memo). Proven at HEAD: a `toAmount` of `'-1'` on either proto path builds `opPayment.amount = -1` with **no throw**, i.e. a signing input is produced and the MPC ceremony runs on a tx XRPL then bounces post-sign — the *exact* "build JSON that XRPL would bounce after the ceremony already ran" scenario the new comment cites as the reason to reject on the raw-JSON path. So by the commit's own rationale the two proto paths have the same gap.

Not blocking: a negative `toAmount` is malformed/adversarial (empty `''` is already rejected on all paths by the strict regex; only *negative* differs), and the failure mode is a wasted ceremony, not fund loss. There is no clean `toBoundedLong` combo for `[0, 2^63-1]` — `signed` accepts negatives, `unsigned` shifts the ceiling to `2^64-1`. But since XRP max drops (~1e17) sit far below `2^63`, the proto paths could safely take `unsigned: true` as well (the `2^63..2^64-1` mis-encode range is unreachable for any real XRP amount), or add an explicit `>= 0n` check — either would make the pre-ceremony rejection consistent across all three paths.

## q

- `:76` raw-JSON upper bound: `unsigned: true` caps at `2^64-1`, but XRPL max drops is ~1e17. An amount in `(1e17, 2^64-1]` passes the bound and is stringified into the JSON, then XRPL rejects it post-sign. Fail-safe and astronomically unreachable, so fine — flagging only for completeness. No other user-controlled amount field on the raw-JSON path (`Fee` is chain-derived `gas`, not `toAmount`; no `SendMax`/`DeliverMin`/partial-payment fields in this simple Payment).

## risk

Low. The changed line is fail-closed and correct for negatives and empties; valid positive amounts are byte-identical (`toBoundedLong(unsigned:true).toString()` == the raw decimal). The new test is non-inert. Residual: proto-path asymmetry (ceremony-waste only, above) and the carried-over `cardano.ts:40` should-fix (no live hole today).

## verdict

**COMMENTED** (draft — not approving). The ripple raw-JSON negative-amount path is now fail-closed before signing, on the right field, with a non-inert test. Recommend closing the proto-path asymmetry for consistency with the PR's stated goal, and the still-open `cardano.ts:40` residual, before this leaves draft.

---
two-lane review (Claude load-bearing; Codex gpt-5.4 infra-dead this run).

rcoderdev (approved):

PR #1215, opened 2026-07-12
fix(solana): sign dapp raw-tx original message bytes, not the walletcore re-encode

### LINKED ISSUE #1204: fix(solana): sign dApp raw-tx original message bytes, not the WalletCore re-encode (mixed-vault co-sign divergence)
## Problem
When co-signing a dApp-supplied raw Solana transaction (`signData.signSolana`), the shared core decodes the tx and hands the decoded proto to WalletCore to **re-encode**, instead of signing the original message bytes verbatim:

`packages/core/mpc/keysign/signingInputs/resolvers/solana/index.ts:23-41`
```ts
const rawMessageData = keysignPayload.signData.value.rawTransactions[0]
const decodedData = walletCore.TransactionDecoder.decode(coinType, Buffer.from(rawMessageData, 'base64'))
const rawMessage = TW.Solana.Proto.DecodingTransactionOutput.decode(decodedData).transaction
return [ TW.Solana.Proto.SigningInput.create({ rawMessage }) ] // WalletCore re-encodes rawMessage
```
For **v0 transactions with Address Lookup Tables** (the standard shape for DEX/aggregator swaps), WalletCore's re-encode is not guaranteed byte-identical to the original → a different pre-image hash than the dApp/another co-signer expects.

## Why this is now high priority — live mixed-vault co-sign break
iOS already moved OFF this path in vultisig-ios#4419 — `Solana.swift:502-532` (`extractSolanaMessageBytes`) strips the `[shortvec(numSigs)][sigs]` envelope and signs the original message bytes, explicitly because the `SigningInput.rawMessage` + `TransactionCompiler` round-trip is *"sensitive to WalletCore version drift across platforms."* Android is making the same fix (vultisig-android#5223). **That leaves windows/sdk as the only platform still re-encoding** — so an **iOS ↔ Windows Secure Vault** co-signing a v0+ALT dApp/swap tx computes divergent pre-image hashes → the ceremony stalls (the exact ios#4419 symptom). This is a live mixed-vault break on the most common swap shape.

`windows == sdk` on the money path (windows imports `@vultisig/core-*`), so this single SDK fix also covers Windows.

## Fix
Mirror ios#4419 / android#5223: derive the message bytes directly from the raw transaction (parse the compact-u16 signature-count envelope, take the message bytes after the signature slots as-is) for the pre-image hash, and splice the 64-byte signature into the original bytes when assembling — do NOT route dApp raw txns through `TransactionDecoder.decode -> SigningInput.rawMessage`.

### Must do
- [ ] `signSolana` raw-tx path computes the pre-image hash from the original message bytes, not a WalletCore re-encode
- [ ] Assemble the signed tx by splicing the signature into the original bytes
- [ ] Byte-parity unit test with a captured v0+ALT transaction; assert the hash matches iOS `extractSolanaMessageBytes`
### Must NOT do
- Don't change the wallet-built transfer/token path (`getSolanaSendSigningInput`)
- Don't change the native/general swap decode path here (separate concern; the `swapPayload` branch)

## Acceptance criteria
- [ ] A dApp v0+ALT tx yields the same pre-image hash on sdk/windows as iOS (#4419) and Android (#5223)
- [ ] New byte-parity test passes and fails without the fix

## References
- iOS fix: vultisig-ios#4419 (`Solana.swift:502-532`) · Android sibling: vultisig-android#5223
- Audit: escape-velocity-2026H2 `chains/solana.md` (windows==sdk money path)

gomesalexandre:
Closes vultisig/vultisig-sdk#1204

## what

`signSolana` raw transactions were decoded via `TransactionDecoder` and handed to WalletCore through `SigningInput.rawMessage`, which RE-ENCODES the message to form the ed25519 pre-image. That re-encode is not guaranteed byte-identical for v0+ALT transactions across WalletCore versions. iOS (#4419) and Android (#5223) already moved to signing the original bytes, leaving sdk/windows as the only platform re-encoding - a mixed iOS <-> sdk/windows Secure Vault computes divergent pre-images on the most common swap shape and the ceremony stalls.

## how

Port of ios#4419 `extractSolanaMessageBytes` (byte-for-byte semantics, including the slot-0 fee-payer splice and the shortvec guards):

- new `resolvers/solana/rawTx.ts`: `extractSolanaMessageBytes` (strip `[shortvec(numSigs)][sigs]` envelope) + `spliceSolanaSignature` (slot 0, other slots preserved as dApp placeholders)
- `getEncodedSigningInputs`: signSolana `txInputData` IS the original serialized tx - no TW proto round-trip
- `getPreSigningHashes`: pre-image = wire-format message verbatim
- `compileTx`: signature verified (ed25519, same as before) then spliced into the ORIGINAL bytes; output stays `TW.Solana.Proto.SigningOutput` with base58 `encoded` so the broadcast resolver and tx-hash contract are untouched
- blockaid preview: zero-signature splice (same output shape its consumers decode)
- the resolver's signSolana branch now throws loud if anything reaches it directly - no silent path back to the divergent pre-image

## receipts (byte identity)

- **legacy, no regression**: for legacy txs the new pre-image is byte-identical to what the OLD WalletCore path produced (`LEGACY tx: new path signs the exact bytes the old WalletCore path signed`)
- **dApp oracle**: legacy AND v0+ALT pre-images equal web3.js's own `message.serialize()` - identity with what the dApp expects, by construction
- **live mainnet probe**: a real 2-signer Jupiter v0+ALT tx (815-byte message) round-trips byte-identical through the old path under the repo's PINNED WalletCore version - i.e. this fix changes NOTHING byte-wise today. It removes the cross-platform version-drift sensitivity that ios#4419 documents (I could not produce a concrete divergent fixture under the current pin; the v0 case is asserted-by-construction against the web3.js oracle instead, flagged per plan)
- **e2e**: encode -> hash -> sign -> compile round-trip for legacy, v0+ALT, and N=2 mixed batches; signed tx = original bytes + slot-0 signature; web3.js `verifySignatures()` passes

fail-before (fix reverted, new tests): `3 failed | 12 passed` (the pipeline e2e tests - old path emits a TW proto, not the original bytes)
pass-after: `15 passed (15)`; full core suite `1716 passed`; sdk unit `2757 passed`; root tsc + sdk typecheck clean; eslint clean

## scope

Wallet-built transfer path (`getSolanaSendSigningInput`) and the swapPayload decode branch untouched, per the issue's must-NOT-do.

## side effect worth flagging (latent fix, not a regression)

To make `compileTx()` reach the new solana branch, `BroadcastService.broadcastTx()` now passes `keysignPayload` through to `compileTx()` - previously it never did. `compileTx()` had two other `keysignPayload`-gated branches that this activates for **every** chain broadcast through the SDK, not just Solana:

- **Cardano CIP-20 memo attachment** (`chain === Chain.Cardano`, `const memo = keysignPayload?.memo`)
- **PSBT / SwapKit `signBitcoin`** (`getSwapKitSignBitcoin(keysignPayload)`)

`compileTx()` is only called from this one SDK call site. Before this PR, both branches were unreachable there - a Cardano memo would silently NOT land in the broadcast tx, and the PSBT/SwapKit path would never trigger - even though `TransactionBuilder.extractMessageHashes()` (the pre-sign hash step) already passed `keysignPayload` and assumed the memo/PSBT commitment was there. This PR makes the compile side consistent with what the sign side already assumed; it's a fix riding along with the solana plumbing, not a new risk introduced by it.

Coverage: `compileTx.golden.test.ts` already has a compileTx-level case (`attaches CIP-20 label-674 metadata when memo is present`) pinning that a memo'd Cardano `keysignPayload` produces the correct CIP-20-tagged output - that's the assertion this activation depends on, and it passes today. A `BroadcastService`-level integration test (through the real broadcast call, not just `compileTx()` directly) would need new mocking harness - `BroadcastService` has no existing unit tests and would require stubbing `coreBroadcastTx`/`getTxHash`/`wasmProvider`/`extractMessageHashes` from scratch. Skipping that as out of scope for this PR; noting it here instead.

## risk

Signing-path change, mitigated by: byte-identity under the current WalletCore pin (receipts above), ed25519 `assertSignature` retained before assembly, and loud failure on any path that could regress to the re-encode.

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana transaction signing to preserve the original transaction bytes, including versioned transactions using address lookup tables.
  * Ensured signatures are inserted into the correct fee-payer position without re-encoding transaction messages.
  * Improved compatibility for multi-party vault signing and transaction broadcasting.

* **Tests**
  * Added comprehensive coverage for legacy, versioned, multi-signature, and invalid Solana transactions.


packages/core/mpc/tx/compile/compileTx.ts
```diff
@@ -80,6 +82,39 @@ export const compileTx = ({
...
+    })
+
+    assertSignature({
+      publicKey,
+      message,
+      signature,
+      signatureFormat,
+    })
+
+    const signedTx = spliceSolanaSignature(txInputData, new Uint8Array(signature))
```
NeOMakinG, line 107: `should-fix:` splice assumes the vault is signer index 0 (fee payer) — no validation; a dApp raw-tx where the vault is a non-fee-payer signer gets the signature in the wrong slot — spliceSolanaSignature always writes the 64-byte signature at firstSignatureOffset (signer index 0 = fee payer). Nothing checks that the vault ed25519 pubkey actually IS message account-keys[0]. assertSignature only proves the sig is valid over the message — not that slot 0 belongs to the vault. If a dApp builds a raw tx where the connected vault is a non-fee-payer signer (index >= 1), the vault sig lands in slot 0 (clobbering the real fee-payer placeholder/sig) and the vault own slot stays unsigned -> the tx is rejected on broadcast. The OLD WalletCore path placed the signature at the index matching the provided pubkey, so this is a genuine behavioral narrowing. It fails SAFE (rejection, sig is over the correct bytes, no fund loss) and is byte-for-byte parity with ios#4419 / android#5223 which splice slot 0 the same way — hence should-fix, not blocking.

fix: here (publicKey is in scope): decode the message header and assert account-keys[0] equals the vault ed25519 pubkey before splicing; throw loud (fail-closed) on mismatch instead of producing a silently-wrong tx. Or explicitly document the invariant (vault == fee payer == signer 0) as a hard precondition of the dApp-raw path, matching the iOS/Android contract.

NeOMakinG (commented):
gm — multi-lane pass: claude correctness+fund-safety+WYSIWYS + codex gpt-5.4 adversarial. conceptual-only (no runtime this pass). comment-only.

## should-fix
### 1. splice assumes the vault is signer index 0 (fee payer) — no validation; a dApp raw-tx where the vault is a non-fee-payer signer gets the signature in the wrong slot — `packages/core/mpc/tx/compile/compileTx.ts:107`
spliceSolanaSignature always writes the 64-byte signature at firstSignatureOffset (signer index 0 = fee payer). Nothing checks that the vault ed25519 pubkey actually IS message account-keys[0]. assertSignature only proves the sig is valid over the message — not that slot 0 belongs to the vault. If a dApp builds a raw tx where the connected vault is a non-fee-payer signer (index >= 1), the vault sig lands in slot 0 (clobbering the real fee-payer placeholder/sig) and the vault own slot stays unsigned -> the tx is rejected on broadcast. The OLD WalletCore path placed the signature at the index matching the provided pubkey, so this is a genuine behavioral narrowing. It fails SAFE (rejection, sig is over the correct bytes, no fund loss) and is byte-for-byte parity with ios#4419 / android#5223 which splice slot 0 the same way — hence should-fix, not blocking.
action: here (publicKey is in scope): decode the message header and assert account-keys[0] equals the vault ed25519 pubkey before splicing; throw loud (fail-closed) on mismatch instead of producing a silently-wrong tx. Or explicitly document the invariant (vault == fee payer == signer 0) as a hard precondition of the dApp-raw path, matching the iOS/Android contract.

## q
- confirmed the fix is NON-inert end-to-end (the #228-class risk): every real getPreSigningHashes / compileTx caller threads keysignPayload — TransactionBuilder.extractMessageHashes (TransactionBuilder.ts:285, the TSS hash source), the internal getPreSigningHashes inside compileTx (compileTx.ts:75), cosigner.ts:168, and BroadcastService (this PR). Blockaid getCompiledTxsForBlockaidInput intercepts signSolana BEFORE calling getPreSigningHashes. So the TSS pre-image == extractSolanaMessageBytes(txInputData).message == the key compileTx looks up in keysignSignatures. No caller falls back to the old re-encode on raw bytes. Flagging because Codex raised these as needs-verification — all resolved SAFE.
- confirmed full-message EdDSA signing is NOT a new behavior: the pre-#1204 Solana path fed WalletCore PreSigningOutput.data (which for Solana IS the full wire message, not a 32-byte digest) to the same TSS layer, so signing the ~800-1232-byte message verbatim is unchanged; and bs58 is a declared dep (^6.0.0) in both core/mpc and packages/sdk (already used by tron/fee.ts), not a phantom import.

## risk
Low-to-moderate blast radius. The core fix is correct and byte-exact: dApp raw Solana txs now sign their ORIGINAL wire message verbatim (no WalletCore re-encode), the signed digest == the broadcast message == the Blockaid-scanned message, and the change is gated strictly to the signSolana dApp-raw path — SDK-built sends and the swapPayload branch are untouched. Fail-closed on unparseable/truncated/over-long-shortvec input, no fallback to the divergent re-encode. Only open item is a hardening gap (unvalidated slot-0/fee-payer assumption) that fails SAFE (rejected tx, never a wrong-bytes signature or fund loss) and matches the iOS/Android siblings.

## verdict
COMMENTED

rcoderdev (approved):

NeOMakinG (approved):
## testing approach

Re-review at HEAD `3e06c74e` (prior review was COMMENTED on draft commit `d09bb62`; HEAD advanced + PR now ready). Two lanes: Claude (load-bearing, full re-trace of the Solana raw-tx sign path) + Codex gpt-5.4 (second opinion). Verified against a fresh worktree at HEAD; diffed r1-commit→HEAD to isolate the delta.

Key finding: the **only** change from my r1 commit to HEAD is a documentation comment block on `spliceSolanaSignature` (commit `3e06c74e`). The test file and all signing/broadcast/Blockaid logic are **byte-identical** to what I deeply verified in r1 — so the r1 end-to-end trace (original message bytes signed → spliced → base58 → broadcast, byte-for-byte, WYSIWYS held for legacy + v0+ALT) carries over unchanged.

## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | should-fix | `spliceSolanaSignature` always writes signer slot 0 with no assertion that vault ed25519 pubkey == message account-keys[0]; a non-fee-payer/sponsored signer would land the sig in the wrong slot | ℹ️ documented | Resolved via the exact path I offered in r1 ("document vault==feePayer==signer0 as hard precondition"). Commit `3e06c74e` adds a thorough doc block (rawTx.ts:66-81) stating the assumption, that it fails **liveness-only** (valid sig binds the exact message, misplaced slot → Solana rejects at broadcast, no fund loss / no wrong-bytes signed), that **no sponsored-tx path exists today**, and flags the runtime-guard follow-up needed if one is added. Confirmed in-repo: signSolana payload has no `signerIndex`/`sponsor` field and no multi-signer constructor. Codex converged. |
| r1 | 2 | q | Is full-message EdDSA signing new? Is `bs58` a phantom dep? | ✅ resolved | Verified in r1: old Solana path already fed the full wire message (not a 32-byte digest) to TSS — unchanged. `bs58 ^6.0.0` declared in `core/mpc/package.json:1070`. |

## risk

FUND-CRITICAL signing-integrity fix, but the HEAD delta is documentation-only over an already-verified byte-preserving path. Signs the EXACT original dApp-supplied message bytes; splice touches only the 64-byte signer-0 slot, message region untouched → broadcast serialization == reviewed bytes. Blockaid scans the same original bytes (zero-sig splice) → no WYSIWYS gap. No safety guard dropped: the retired re-encode path performed no fee-payer/blockhash/allowlist validation, so nothing is lost; the fix is strictly more faithful. Residual signer-index-0 assumption is documented, fails safe (liveness-only), and has no reachable trigger today. Both lanes converged on no fund-loss risk.

## verdict

**APPROVE.** The single r1 should-fix is resolved-as-documented; the fix signs original bytes byte-for-byte with WYSIWYS preserved across legacy and v0+ALT, and drops no needed guard. Non-inert red-on-revert coverage verified in r1 (revert → 3 failed / 12 passed).

<sub>Reviewed by the vultisig-ops gomes review loop (Claude load-bearing + Codex gpt-5.4, both converged).</sub>

PR #1216, opened 2026-07-12
fix(sui): bounded deterministic coin selection + covering gas object

### LINKED ISSUE #1132: fix(sui): bounded coin selection + covering gas object — match iOS #4734 (keysign parity; covers Windows)
## Summary

iOS PR [vultisig/vultisig-ios#4734](https://github.com/vultisig/vultisig-ios/pull/4734) changes how a Sui send selects coin objects: it embeds/references only the **fewest largest objects that cover the send** (bounded by 255), and picks a **gas object that actually covers the budget** instead of an arbitrary one. This fixes real failures on wallets whose SUI balance is scattered across many objects (keysign 404 / "data expired", broadcast `serialized transaction size exceeded maximum`, `InsufficientCoinBalance in command 0`) and token-send gas failures.

The SDK's Sui path is behind on **all three** fixes, and — because Windows consumes `@core` — this covers Windows too.

## Current SDK behavior (verified against `vultisig-sdk@main`)

- **Unbounded inputs:** `packages/core/mpc/keysign/signingInputs/resolvers/sui.ts:60,80` — `inputCoins = coins.filter(...).map(createObjectRef)` references **every** matching coin object; no covering-subset selection.
- **Unbounded payload:** `packages/core/mpc/keysign/chainSpecific/resolvers/sui/index.ts:58-63` — embeds every owned coin from paginated `getAllCoins` (up to a page cap), so a dusty wallet still produces an oversized payload.
- **Arbitrary gas object:** `packages/core/mpc/keysign/signingInputs/resolvers/sui.ts:52,67` — `gas: gasCoins[0]` picks the first SUI object the RPC returned. A token send fails when that object can't cover the gas budget, even with plenty of SUI elsewhere. This is exactly the bug iOS #4734 (commit 1) and Android (#3989) already fixed.

## Why this is a cross-device consensus blocker

In MPC keysign every co-signing device **independently recomputes** the Sui transaction and its sighash from the shared payload; all must produce **byte-identical** bytes or the send can't be signed. Once #4734 ships:

- **SDK/Windows co-signer + updated iOS device** on a multi-object wallet → iOS trims to a balance-sorted covering subset while the SDK references all objects in RPC order → different tx bytes → **keysign fails**. FastVault (server co-signer running the SDK/core) is in scope wherever the server rebuilds the tx.

Before #4734 the platforms agreed (all referenced every object in payload order). #4734 introduces the divergence, so the SDK must match to keep cross-platform vaults and FastVault working.

## Scope for the SDK / `@core`

Port the **exact** deterministic algorithm from #4734 (`SuiConstants.swift` → `SuiCoinType`):

1. **`selectInputCoins`:** fewest largest objects covering `target`, sorted **balance desc, then `objectID` ascending**, capped at **255**. Native `PaySui` target = `amount + gasBudget`; token `Pay` target = `amount`.
2. **Payload bounding** in `chainSpecific/resolvers/sui/index.ts`: embed covering objects (native) / covering token objects + largest-few native gas candidates (token).
3. **`selectGasObject`:** replace `gasCoins[0]` with the **smallest native SUI object that covers `gasBudget`** (match Android `selectSuiGasCoin` / iOS `selectGasObject`). Decide the no-covering-object fallback deliberately — Android fails fast (returns null); iOS #4734 falls back to the largest object. **Pick one and make all platforms match**, since this also affects the digest.

Sort key, tie-break, cap, and gas-pick (incl. fallback) must be **identical** across iOS / Android / SDK or keysigns diverge.

## Verification (required)

Cross-platform interop test: fixed owned-coins fixture + `amount` + `gasBudget` → assert the SDK's `TW.Sui.Proto.SigningInput` bytes equal iOS's for the same fixture (native small / native near-max / 800-object dusty / token send). Mirror `SuiHelperInputDataTests.swift` from #4734.

## Coordination

Do **not** merge #4734 to a public iOS release until this lands on the SDK (and therefore Windows), or all platforms version/capability-gate the new behavior. Track: iOS #4734 · Android parity issue (link once filed) · this issue (SDK, covers Windows via `@core`).

---
_Receipts verified against `vultisig-sdk@main` and `vultisig-ios` PR #4734 head (`6834d99`)._

gomesalexandre:
Closes vultisig/vultisig-sdk#1132

## what

Port of iOS #4734 (`SuiConstants`/`SuiCoinType`), Android sibling #3989. Three gaps in the SDK's Sui send path, all keysign-consensus-relevant (every co-signing device recomputes the tx bytes from the shared payload - selection MUST be identical across iOS / Android / SDK / Windows / FastVault or the sighash diverges and the ceremony fails):

1. **unbounded inputs** - `inputCoins` referenced every matching object; a dusty wallet blew past Sui's 128 KiB tx-size / 256-gas-object caps ('serialized transaction size exceeded maximum')
2. **arbitrary gas object** - `gas: gasCoins[0]`; a token send failed when the RPC's first object couldn't cover the budget despite plenty of SUI elsewhere
3. **unbounded payload** - the keysign payload embedded every owned coin incl. unrelated memecoins, oversizing the relay payload (co-signer poll 404 / 'data expired')

## how

New `packages/core/mpc/chains/sui/coinSelection.ts`, exact iOS #4734 semantics:

- `selectSuiInputCoins`: fewest largest objects covering the target (native `PaySui`: amount+gas; token `Pay`: amount), sorted balance desc then objectID asc, capped at 255, keep-at-least-one, best-effort when under-funded
- `selectSuiGasObject`: smallest native object covering the budget; largest-object fallback when none covers (the iOS #4734 choice, per the issue's 'pick one and make all platforms match'); undefined only with zero native objects (resolver throws)
- `selectSuiPayloadCoins`: covering subset (native) / covering tokens + top-5 native gas candidates (token) embedded in the chainSpecific payload
- normalization-aware type matching (short `0x2` vs the 64-hex long form `suix_getAllCoins` returns) - the old exact-`===` filter missed long-form native objects entirely and can't tell `sui::SUI` from `xsui::XSUI` classes by construction

Wired into `resolvers/sui.ts` (selection) and `chainSpecific/resolvers/sui/index.ts` (payload bounding). `signSui` dApp PTB path untouched.

## testing

Fixture tests mirror `SuiHelperInputDataTests.swift` from iOS #4734 verbatim (same coins/amounts/budgets/expected object sets): native scattered/small/near-max/800-object-dusty/look-alike-LST, token covering-gas/covering-tokens/small-send, plus unit tests for sort/tie-break/cap/fallback/normalization.

fail-before (fix reverted): `10 failed | 12 passed` (every fixture-parity test - old code refs all objects and picks gasCoins[0])
pass-after: `22 passed (22)`; full core suite `1723 passed`; root tsc clean; eslint clean

## coordination

Per the issue: iOS #4734 is still OPEN - the platforms must land this together (or gate) since the new selection intentionally diverges from the old all-objects behavior. This PR makes the SDK ready.

## risk

Sui send construction only. Selection is deterministic and best-effort on under-funded sets (same as iOS); dApp-supplied PTBs and swap paths untouched.

🤖 Agent-generated

### DIFF
--- a/packages/core/mpc/chains/sui/coinSelection.ts
+++ b/packages/core/mpc/chains/sui/coinSelection.ts
@@ -1,0 +1,181 @@
+/**
+ * Deterministic Sui coin-object selection (sdk#1132).
+ *
+ * Port of vultisig-ios#4734 (`SuiConstants.swift` / `SuiCoinType`), Android
+ * sibling vultisig-android#3989. In MPC keysign every co-signing device
+ * independently recomputes the Sui transaction bytes from the shared payload;
+ * the selection here must stay IDENTICAL across iOS / Android / SDK (and
+ * therefore Windows + FastVault's server co-signer) or the sighashes diverge
+ * and the ceremony fails. Do not change sort keys, tie-breaks, caps, or the
+ * gas-pick fallback without coordinating all platforms.
+ */
+
+import { SuiCoin } from '../../types/vultisig/keysign/v1/blockchain_specific_pb'
+
+/** Canonical fully-qualified type of the native SUI coin (short address form). */
+export const suiNativeCoinType = '0x2::sui::SUI'
+
+/**
+ * Upper bound on the coin objects a single send may reference. Sui rejects a
+ * transaction whose serialized size exceeds 128 KiB, and a `PaySui` send uses
+ * its entire input set as the gas payment — which Sui caps at 256 objects.
+ * Staying one under the cap keeps every send safely within both limits.
+ */
+export const maxSuiInputCoinObjects = 255
+
+/**
+ * How many of the largest native SUI objects to embed as gas candidates for a
+ * token send. The signer picks one to pay gas; carrying the largest few keeps
+ * the payload small while guaranteeing a covering object survives a
+ * re-estimated gas budget.
+ */
+export const suiGasCandidateObjectCount = 5
+
+/**
+ * Collapses a package-address segment to `0x` + hex with leading zeros
+ * stripped, so `0x0000…0002` and `0x2` compare equal.
+ */
+const normalizeSuiAddress = (address: string): string => {
+  const hex = address.startsWith('0x') ? address.slice(2) : address
+  const trimmed = hex.replace(/^0+/, '')
+  return `0x${trimmed === '' ? '0' : trimmed}`
+}
+
+/**
+ * Normalizes a fully-qualified `address::module::struct` coin type for exact
+ * comparison: lowercases the whole string and collapses the package-address
+ * segment to canonical form. Ticker-substring matching is wrong here — it
+ * cannot distinguish `0x2::sui::SUI` from `0x…::xsui::XSUI`.
+ */
+export const normalizeSuiCoinType = (coinType: string): string => {
+  const lowered = coinType.toLowerCase()
+  const [addressSegment] = lowered.split(':')
+  if (!addressSegment) return lowered
+  return normalizeSuiAddress(addressSegment) + lowered.slice(addressSegment.length)
+}
+
+/** Whether two fully-qualified coin types refer to the same coin. */
+export const suiCoinTypesMatch = (a: string, b: string): boolean => normalizeSuiCoinType(a) === normalizeSuiCoinType(b)
+
+/** Whether the given coin-object type is the native SUI coin. */
+export const isNativeSuiCoinType = (coinType: string): boolean => suiCoinTypesMatch(coinType, suiNativeCoinType)
+
+/** Parses a coin object's balance (base-unit MIST), unparseable = zero. */
+export const suiCoinBalance = (coin: SuiCoin): bigint => {
+  try {
+    return BigInt(coin.balance || '0')
+  } catch {
+    return BigInt(0)
+  }
+}
+
+/**
+ * Selects the fewest coin objects (largest balance first) that together cover
+ * `target`, bounded by `maxObjects`.
+ *
+ * Selection is deterministic (balance descending, then `coinObjectId`
+ * ascending) so every co-signing device selects the identical set and signs
+ * the identical transaction. If even `maxObjects` largest objects don't reach
+ * `target`, they are still returned (best effort) — the caller decides how to
+ * handle an under-funded selection.
+ */
+export const selectSuiInputCoins = (
+  coins: SuiCoin[],
+  target: bigint,
+  maxObjects: number = maxSuiInputCoinObjects
+): SuiCoin[] => {
+  const sorted = [...coins].sort((lhs, rhs) => {
+    const lhsBalance = suiCoinBalance(lhs)
+    const rhsBalance = suiCoinBalance(rhs)
+    if (lhsBalance !== rhsBalance) return lhsBalance > rhsBalance ? -1 : 1
+    return lhs.coinObjectId < rhs.coinObjectId ? -1 : lhs.coinObjectId > rhs.coinObjectId ? 1 : 0
+  })
+
+  const selected: SuiCoin[] = []
+  let accumulated = BigInt(0)
+  for (const coin of sorted) {
+    // Always keep at least one object so a zero/near-zero-amount send still
+    // has an input; otherwise stop once the target is covered.
+    if (selected.length > 0 && accumulated >= target) break
+    if (selected.length >= maxObjects) break
+    selected.push(coin)
+    accumulated += suiCoinBalance(coin)
+  }
+  return selected
+}
+
+/**
+ * Selects the native SUI coin object that pays gas for a token (non-native)
+ * send. WalletCore's `Sui.Pay` message carries a *single* `gas` object (unlike
+ * `PaySui`, whose whole input set is gas-smashed), so the choice matters:
+ * picking an arbitrary object fails when its balance can't cover the budget,
+ * even though the wallet holds plenty of SUI across other objects.
+ *
+ * Choose the *smallest* native SUI object whose balance already covers
+ * `gasBudget`; when no single object covers it, fall back to the largest
+ * object (best effort — strictly better than an arbitrary pick). Returns
+ * `undefined` only when the wallet holds no native SUI object at all.
+ * Tie-breaks mirror Swift's `min(by:)` (first minimal) / `max(by:)` (last
+ * maximal) over the payload's coin order.
+ */
+export const selectSuiGasObject = (coins: SuiCoin[], gasBudget: bigint): SuiCoin | undefined => {
+  const suiObjects = coins.filter(coin => isNativeSuiCoinType(coin.coinType))
+  if (suiObjects.length === 0) return undefined
+
+  let smallestCovering: SuiCoin | undefined
+  for (const coin of suiObjects) {
+    if (suiCoinBalance(coin) < gasBudget) continue
+    if (!smallestCovering || suiCoinBalance(coin) < suiCoinBalance(smallestCovering)) {
+      smallestCovering = coin
+    }
+  }
+  if (smallestCovering) return smallestCovering
+
+  let largest = suiObjects[0]
+  for (const coin of suiObjects) {
+    if (!(suiCoinBalance(coin) < suiCoinBalance(largest))) largest = coin
+  }
+  return largest
+}
+
+/**
+ * The minimal set of coin objects to embed in the keysign payload for a Sui
+ * send — exactly what the signing-input resolver will consume. Unbounded
+ * embedding on a dusty wallet produces a keysign payload too large to relay
+ * (co-signer poll 404 / "data expired") and, at signing, a transaction over
+ * Sui's size limits.
+ *
+ * Native send: the largest objects covering `amount + gasBudget` (the input
+ * set also pays gas). Token send: the largest token objects covering `amount`,
+ * plus the largest few native SUI objects as gas candidates.
+ */
+export const selectSuiPayloadCoins = ({
+  coins,
+  contractAddress,
+  amount,
+  gasBudget,
+}: {
+  coins: SuiCoin[]
+  /** The token's fully-qualified type; empty for a native SUI send. */
+  contractAddress: string
+  amount: bigint
+  gasBudget: bigint
+}): SuiCoin[] => {
+  const nativeObjects = coins.filter(coin => isNativeSuiCoinType(coin.coinType))
+
+  if (!contractAddress) {
+    return selectSuiInputCoins(nativeObjects, amount + gasBudget)
+  }
+
+  const tokenObjects = coins.filter(coin => suiCoinTypesMatch(coin.coinType, contractAddress))
+  const selectedTokens = selectSuiInputCoins(tokenObjects, amount)
+  const gasCandidates = [...nativeObjects]
+    .sort((lhs, rhs) => {
+      const lhsBalance = suiCoinBalance(lhs)
+      const rhsBalance = suiCoinBalance(rhs)
+      if (lhsBalance !== rhsBalance) return lhsBalance > rhsBalance ? -1 : 1
+      return 0
+    })
+    .slice(0, suiGasCandidateObjectCount)
+  return [...selectedTokens, ...gasCandidates]
+}
--- a/packages/core/mpc/keysign/chainSpecific/resolvers/sui/index.ts
+++ b/packages/core/mpc/keysign/chainSpecific/resolvers/sui/index.ts
@@ -56,16 +56,19 @@ export const getSuiChainSpecific: GetChainSpecificResolver<'suicheSpecific'> = a
     }
   } while (cursor)
 
+  const rawCoinObjects = rawCoins.map((coin: CoinStruct) => create(SuiCoinSchema, coin))
+
   // Bound the embedded coin set to what the send actually needs (sdk#1132,
   // iOS #4734 parity): embedding every owned object makes a dusty wallet's
   // keysign payload too large to relay (co-signer poll 404 / "data expired")
-  // and drags unrelated coin types (memecoins/LSTs) into the payload. The
-  // signing-input resolver performs the same deterministic selection on this
-  // set, so the bounded output is exactly the inputs the signer consumes —
-  // plus, for a token send, the largest few native objects as gas candidates
-  // so a covering gas object survives the refine step's re-estimated budget.
+  // and drags unrelated coin types (memecoins/LSTs) into the payload. This
+  // baseline selection (sized against the static `suiGasBudget` default) only
+  // needs to be good enough to build a valid dry-run tx below — refine
+  // re-selects from the full `rawCoinObjects` set against the REAL gas
+  // budget once it's known, so the payload coins that actually ship lock in
+  // exactly once, after the real cost is known (sdk#1216 follow-up).
   const coins = selectSuiPayloadCoins({
-    coins: rawCoins.map((coin: CoinStruct) => create(SuiCoinSchema, coin)),
+    coins: rawCoinObjects,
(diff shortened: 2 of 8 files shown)


packages/core/mpc/chains/sui/coinSelection.ts
```diff
@@ -0,0 +1,181 @@
...
+}): SuiCoin[] => {
+  const nativeObjects = coins.filter(coin => isNativeSuiCoinType(coin.coinType))
+
+  if (!contractAddress) {
+    return selectSuiInputCoins(nativeObjects, amount + gasBudget)
+  }
+
+  const tokenObjects = coins.filter(coin => suiCoinTypesMatch(coin.coinType, contractAddress))
+  const selectedTokens = selectSuiInputCoins(tokenObjects, amount)
+  const gasCandidates = [...nativeObjects]
```
NeOMakinG, line 172: `should-fix:` gas-candidate tie-break uses `return 0` — inconsistent with the canonical objectId tie-break used everywhere else — The `selectSuiPayloadCoins` token branch sorts native gas candidates by balance desc but tie-breaks equal balances with `return 0`, then `slice(0, 5)`. Every other selector (`selectSuiInputCoins`:91) tie-breaks by `coinObjectId` ascending precisely so the pick is engine-independent. This is NOT a same-ceremony consensus break — the payload (incl. the 5 embedded candidates) is built once by the initiator and shared, and `selectSuiGasObject` then runs identically over that shared, protobuf-order-preserved array on every device — so all co-signers agree. But it IS fragile: which 5 candidates get embedded now depends on ES2019 stable-sort + identical RPC pre-sort order, and it diverges from the byte-identical-across-platforms goal the module's own header demands. A wallet with 6+ equal-balance native objects around the slice cutoff can embed a different candidate subset run-to-run / platform-to-platform.

fix: Give the gas-candidate comparator the same `coinObjectId` ascending tie-break as `selectSuiInputCoins`, so the embedded set is fully order-independent and consistent with the rest of the module.

NeOMakinG (commented):
gm — multi-lane pass: claude load-bearing + codex gpt-5.4. conceptual-only (no runtime this pass). comment-only.

## should-fix
### 1. gas-candidate tie-break uses `return 0` — inconsistent with the canonical objectId tie-break used everywhere else — `packages/core/mpc/chains/sui/coinSelection.ts:172`
The `selectSuiPayloadCoins` token branch sorts native gas candidates by balance desc but tie-breaks equal balances with `return 0`, then `slice(0, 5)`. Every other selector (`selectSuiInputCoins`:91) tie-breaks by `coinObjectId` ascending precisely so the pick is engine-independent. This is NOT a same-ceremony consensus break — the payload (incl. the 5 embedded candidates) is built once by the initiator and shared, and `selectSuiGasObject` then runs identically over that shared, protobuf-order-preserved array on every device — so all co-signers agree. But it IS fragile: which 5 candidates get embedded now depends on ES2019 stable-sort + identical RPC pre-sort order, and it diverges from the byte-identical-across-platforms goal the module's own header demands. A wallet with 6+ equal-balance native objects around the slice cutoff can embed a different candidate subset run-to-run / platform-to-platform.
action: Give the gas-candidate comparator the same `coinObjectId` ascending tie-break as `selectSuiInputCoins`, so the embedded set is fully order-independent and consistent with the rest of the module.

### 2. payload coins bounded with the default gas budget, but refine can raise the budget afterward — native near-max/dusty sends can false-insufficient — `packages/core/mpc/keysign/chainSpecific/resolvers/sui/index.ts:67`
`selectSuiPayloadCoins` bounds the embedded native set to cover `amount + suiGasBudget` (the 3_000_000 MIST default). `refineSuiChainSpecific` then dry-runs and can RAISE `gasBudget` (115% of computation+storage, and a 255-object PaySui merge costs materially more than 3M), returning only the new budget — it does NOT re-bound the coin set. At signing, `getSuiSigningInputs` re-runs `selectSuiInputCoins(nativeCoins, amount + refinedBudget)` over the already-truncated payload. On a near-max or many-object native send the extra native object that only becomes necessary after the budget bump was never embedded, so the PaySui input set can fall short even though the full wallet had enough SUI. Fail-closed (tx rejected, no fund loss) and narrow, but real and untested. Token sends are shielded because they carry top-5 native gas candidates; native sends are the exposed path.
action: Add gas headroom to the native payload bound (e.g. bound against a max plausible refined budget) or have refine re-bound the native coin set when it raises the budget; add a fixture where refine's budget exceeds the default and the pre-refine bound was tight.

### 3. changeset is `patch` but PR adds a new public subpath export that no consumer uses — `packages/core/mpc/package.json:32`
The new `./chains/sui/coinSelection` export widens `@vultisig/core-mpc`'s public API, yet all three consumers (both resolvers + the test) import it via relative paths — nothing uses the subpath. Per the repo's own convention a new public export is a minor, not a patch. Shipping it as patch both mis-labels the bump and silently creates a supported subpath + future semver constraint for caret downstreams.
action: Either drop the export and keep the helper internal (relative imports already work), or, if external/cross-package use is intended, bump the changeset to minor.

## q
- `coinObjectId` is compared RAW in the sort tie-break (coinSelection.ts:91) — unlike `coinType`, it is not normalized. Can you confirm Sui object IDs always surface as the same canonical lowercase full-length `0x`+64-hex form on every co-signer input (iOS/Android/Windows/FastVault), so the string tie-break can never diverge cross-platform on equal-balance sets?

## risk
Sui send-construction only; dApp PTB / swap paths untouched. Core consensus property holds: the coin set is selected once into the shared keysign payload and every co-signer derives its SigningInput from that identical payload via pure, deterministic functions (objectId tie-break + protobuf-order-preserved gas pick), so a single ceremony cannot diverge sighashes. Residual risks are all fail-closed (tx rejected on-chain, no fund loss) or cross-run/cross-platform parity hygiene. Draft: iOS #4734 must land in lockstep.

## verdict
COMMENTED

packages/core/mpc/keysign/chainSpecific/resolvers/sui/index.ts
```diff
@@ -55,7 +56,20 @@ export const getSuiChainSpecific: GetChainSpecificResolver<'suicheSpecific'> = a
...
-  const coins = rawCoins.map((coin: CoinStruct) => create(SuiCoinSchema, coin))
+  // Bound the embedded coin set to what the send actually needs (sdk#1132,
+  // iOS #4734 parity): embedding every owned object makes a dusty wallet's
+  // keysign payload too large to relay (co-signer poll 404 / "data expired")
+  // and drags unrelated coin types (memecoins/LSTs) into the payload. The
+  // signing-input resolver performs the same deterministic selection on this
+  // set, so the bounded output is exactly the inputs the signer consumes —
+  // plus, for a token send, the largest few native objects as gas candidates
+  // so a covering gas object survives the refine step's re-estimated budget.
+  const coins = selectSuiPayloadCoins({
```
NeOMakinG, line 70: `should-fix:` payload coins bounded with the default gas budget, but refine can raise the budget afterward — native near-max/dusty sends can false-insufficient — `selectSuiPayloadCoins` bounds the embedded native set to cover `amount + suiGasBudget` (the 3_000_000 MIST default). `refineSuiChainSpecific` then dry-runs and can RAISE `gasBudget` (115% of computation+storage, and a 255-object PaySui merge costs materially more than 3M), returning only the new budget — it does NOT re-bound the coin set. At signing, `getSuiSigningInputs` re-runs `selectSuiInputCoins(nativeCoins, amount + refinedBudget)` over the already-truncated payload. On a near-max or many-object native send the extra native object that only becomes necessary after the budget bump was never embedded, so the PaySui input set can fall short even though the full wallet had enough SUI. Fail-closed (tx rejected, no fund loss) and narrow, but real and untested. Token sends are shielded because they carry top-5 native gas candidates; native sends are the exposed path.

fix: Add gas headroom to the native payload bound (e.g. bound against a max plausible refined budget) or have refine re-bound the native coin set when it raises the budget; add a fixture where refine's budget exceeds the default and the pre-refine bound was tight.

packages/core/mpc/package.json
```diff
@@ -29,6 +29,11 @@
       "import": "./dist/chains/cosmos/qbtc/QBTCTx.js",
       "default": "./dist/chains/cosmos/qbtc/QBTCTx.js"
     },
+    "./chains/sui/coinSelection": {
```
NeOMakinG, line 32: `should-fix:` changeset is `patch` but PR adds a new public subpath export that no consumer uses — The new `./chains/sui/coinSelection` export widens `@vultisig/core-mpc`'s public API, yet all three consumers (both resolvers + the test) import it via relative paths — nothing uses the subpath. Per the repo's own convention a new public export is a minor, not a patch. Shipping it as patch both mis-labels the bump and silently creates a supported subpath + future semver constraint for caret downstreams.

fix: Either drop the export and keep the helper internal (relative imports already work), or, if external/cross-package use is intended, bump the changeset to minor.

packages/core/mpc/chains/sui/coinSelection.ts
```diff
@@ -0,0 +1,181 @@
...
+export const selectSuiInputCoins = (
+  coins: SuiCoin[],
+  target: bigint,
+  maxObjects: number = maxSuiInputCoinObjects
+): SuiCoin[] => {
+  const sorted = [...coins].sort((lhs, rhs) => {
+    const lhsBalance = suiCoinBalance(lhs)
+    const rhsBalance = suiCoinBalance(rhs)
+    if (lhsBalance !== rhsBalance) return lhsBalance > rhsBalance ? -1 : 1
+    return lhs.coinObjectId < rhs.coinObjectId ? -1 : lhs.coinObjectId > rhs.coinObjectId ? 1 : 0
```
NeOMakinG, line 91: `q:` `coinObjectId` is compared RAW in the sort tie-break (coinSelection.ts:91) — unlike `coinType`, it is not normalized. Can you confirm Sui object IDs always surface as the same canonical lowercase full-length `0x`+64-hex form on every co-signer input (iOS/Android/Windows/FastVault), so the string tie-break can never diverge cross-platform on equal-balance sets?

rcoderdev (approved):

NeOMakinG (approved):
Re-review at HEAD `b9050284` (prior review was COMMENTED at `21e692d`). Two lanes: Claude (load-bearing static + non-inert test verification) + Codex gpt-5.4 (adversarial second opinion). Verdict corroborates the standing APPROVED decision — **approving**.

## testing approach

- Read all four Sui files at HEAD, not just the diff: `chains/sui/coinSelection.ts`, `keysign/chainSpecific/resolvers/sui/{index,refine}.ts`, `keysign/signingInputs/resolvers/sui.ts`.
- Re-derived the four fund-safety invariants at HEAD: **deterministic** (stable sort, `coinObjectId` asc tie-break, single-selection-into-shared-payload architecture), **bounded** (255 object cap + `MAX_COIN_PAGES=200` fail-closed cursor loop), **covering / no-double-count**, **fail-closed** on under-funded / no-native-gas / no-token.
- Could not execute the vitest suites in the isolated worktree — the package `exports` map points at `./dist/...` so the self-referential `@vultisig/core-mpc/types/*` imports need a built package; the worktree overlay resolves them to canonical's stale `dist`. This is a resolution artifact of the sandbox, not a code defect (same wall as round 1). Compensated by reading the test files directly and confirming the mutation-sensitive assertions exist (exact dry-run call counts 2 and 3, exact budgets `3450000`/`3910000`, `coins` cover `amount + gasBudget` at refine.test.ts:89/160, bounded ≤255 at :95/163) — these are genuinely red-on-revert, not inert.
- Codex gpt-5.4 adversarial lane, scoped to the new convergence loop: **no net-new findings**.

## prior issues — status

| 1 | 2 | should-fix (Codex called blocker) | payload coins narrowed against DEFAULT `suiGasBudget` BEFORE refine raised the budget; refine returned only the new budget, not re-bounded coins → native near-max send could false-insufficient | ✅ fixed | commits `7270c80` + `ad5af0e`. `refine.ts` now re-selects payload coins from the FULL raw set AGAINST the refined budget (`refine.ts:103`), and re-prices+re-selects if the grown object count under-priced the dry run, bounded at 2 iterations (`refine.ts:113-125`). Final `coins` are always selected against the final returned `gasBudget` — covering-consistent. Regression test asserts covering + exact call counts (refine.test.ts:86-95, :153-163). |
| 1 | 1 | should-fix | gas-candidate slice tie-break `return 0` (coinSelection.ts, now :177) inconsistent with the `coinObjectId` asc tie-break used in `selectSuiInputCoins` (:91) | ⚠️ documented — intentionally NOT hand-fixed | See should-fix below. On reflection the SDK must NOT unilaterally add a secondary tie-break here: the file header forbids changing tie-breaks without coordinating iOS #4734 / Android #3989, and a divergent tie-break would DESYNC the embedded gas-candidate subset (>5 equal-balance native objects) cross-platform. It is not a within-ceremony divergence risk (selection happens once on the initiator and ships in the shared payload), only a run-to-run WYSIWYS-reproducibility nit. Safer left as-is until all platforms move together. |
| 1 | 3 | should-fix | new public export `./chains/sui/coinSelection` but changeset is `patch` (convention: new export = minor) | ℹ️ effectively moot | The rebased `xrp-destination-tags` changeset already bumps `@vultisig/core-mpc` to `minor` this cycle, so the aggregate release will be minor regardless. Left as a suggestion. |
| 1 | 4 | q | `coinObjectId` compared RAW (not normalized) in the sort tie-break (:91) | ℹ️ answered | Sui `suix_getAllCoins` returns object IDs as canonical `0x`+64-hex full form (unlike package addresses, which can be short) across all platforms' RPC, so the raw compare is stable. Low risk; no change needed. |

## should-fix

- `packages/core/mpc/chains/sui/coinSelection.ts:172-179` — the gas-candidate sort has no secondary tie-break (`return 0`), unlike `selectSuiInputCoins` (:91). This is the last unresolved round-1 item. I am deliberately NOT fixing it in-branch: the coordination header forbids changing tie-breaks without moving iOS/Android in lockstep, and a unilateral SDK change would desync the embedded gas-candidate subset when >5 native objects share a balance. Track a coordinated cross-platform follow-up if you want strict run-to-run reproducibility; not fund-blocking (single-ceremony selection is shared, so no sighash divergence).

## suggestion

- `.changeset/sui-bounded-coin-selection.md` declares `@vultisig/core-mpc: patch` while the PR adds a new public export subpath. Prefer `minor` for self-documentation, even though the co-landing xrp changeset already forces a minor this cycle.

## q

- (resolved, see prior-issues table) — confirm Sui object IDs stay canonical full-form on every co-signer's input source; the raw tie-break compare depends on it.

## risk

Sui send construction only. dApp-supplied PTB path (`signSui`) and swap paths untouched. Selection is deterministic and best-effort on under-funded sets; every residual failure mode (dust bound hit, exact-balance/max send with no gas headroom, stale-budget after the 2-iteration re-price bound) is a clean chain-side `InsufficientGas`/insufficient-balance rejection — **fail-closed, never wrong-amount / fund-loss**. Token gas object (native) is disjoint from the token spend set; native PaySui gas-smashes the whole covering set with no separate gas field — no double-count either way. Consensus-safe by construction: the initiator selects once into the shared `KeysignPayload`, and every co-signer re-derives an identical `SigningInput` from that payload via pure deterministic functions.

## verdict

**Approve.** The round-1 blocking concern (finding #2) is fully resolved by `7270c80` + `ad5af0e`; the covering/bounded/determinism/fail-closed invariants all hold at HEAD; the new convergence loop is covering-consistent and its bound is genuinely tested. One should-fix (gas-candidate tie-break) and one suggestion (changeset tier) remain, neither fund-blocking, and the tie-break must move cross-platform in lockstep rather than be patched here. Codex lane found nothing net-new. Note: this PR intentionally diverges from the old all-objects behavior — it must land in lockstep with iOS #4734 / Android #3989 (already noted in the PR).

PR #1192, opened 2026-07-12
feat(sdk): export evm chainid helpers from public api

gomesalexandre:
Closes vultisig/vultisig-sdk#1158

## what

Exports the EVM chainId ↔ chain helpers (`getEvmChainId`, `getEvmChainByChainId`) from the SDK's public API. The SDK already owned the per-chain EVM chainId table (`evmChainId` in `chains/evm/chainInfo.ts`), but it wasn't part of the public export surface — so the app and agent-backend-ts each hand-maintain their own copies of it.

## why

That un-exported table is the exact drift class behind the **Hyperliquid 998 (testnet) vs 999 (mainnet)** client↔server chainId bug: the client signed for 999 while the MCP server targeted 998, so any tx built server-side was malformed relative to the client's signature chain. With the SDK owning + exposing the table, both repos can consume one source of truth instead of independently-maintained copies.

I verified there is **no drift to reconcile first** before exporting:
- **chainIds**: the SDK's values (derived from viem chain defs + the `hyperliquid` `defineChain({ id: 999 })`) decode to the exact canonical numeric ids that agent-backend-ts pins in `mcp/lib/evm-chains.ts` (`EVM_CHAIN_IDS`), chain-for-chain across all 13 EVM chains.
- **native tickers**: already exported via `chainFeeCoin`; its tickers match abts's `EVM_NATIVE_TICKERS` exactly (incl. Polygon `POL`, Mantle `MNT`, Hyperliquid `HYPE`, Cronos `CRO`, Sei `SEI`), so no ticker export is added.

Migrating app + abts to import these is a separate follow-up (this PR just creates the exportable owner + pins the values).

## how tested

New pinning test `tests/unit/chains/evmChainId.test.ts` (imports from the SDK public entry — fails before, passes after):

- `exposes getEvmChainId and getEvmChainByChainId from the SDK public entry` — before the export both symbols are `undefined` (`expected 'undefined' to be 'function'`), after they're functions.
- `pins every EVM chainId (hex) to its canonical numeric value` — Ethereum 1, BSC 56, Polygon 137, Avalanche 43114, Arbitrum 42161, Optimism 10, Base 8453, Blast 81457, Mantle 5000, Zksync 324, Cronos 25, Hyperliquid 999, Sei 1329.
- `round-trips chain -> chainId -> chain for every EVM chain`.
- `pins the Hyperliquid mainnet chainId to 999 (not testnet 998)`.

[...]

PR #1220, opened 2026-07-12
fix(sdk): route RN Ripple balances through the fetch-safe path

### LINKED ISSUE #1218: react-native getCoinBalance still routed Ripple through the websocket core resolver
## Summary
The React Native `getCoinBalance()` entrypoint exported by `@vultisig/sdk/react-native` routed Ripple balance reads through the websocket `xrpl` client path instead of the RN-safe fetch helpers already present in the SDK.

## Evidence
- `packages/sdk/src/platforms/react-native/index.ts:479-480` exports RN `getCoinBalance()`
- `packages/sdk/src/platforms/react-native/getCoinBalance.ts` routed Ripple to `@vultisig/core-chain/coin/balance/resolvers/ripple`
- `packages/core/chain/chains/ripple/client.ts:1-9` uses `xrpl` over `wss://xrplcluster.com`
- `packages/sdk/src/platforms/react-native/chains/ripple/rpc.ts:1-8` already documents that RN must use fetch-based Ripple helpers under Hermes

## Impact
- The public RN balance helper can pull in the wrong runtime path for Ripple exactly on the platform where websocket/Hermes compatibility is the concern.
- New XRPL trust-line balance support also stays stranded from the RN entrypoint when the generic helper bypasses the RN Ripple module.

## Fix sketch
- Add an RN-specific Ripple balance adapter behind `getCoinBalance()`.
- Use fetch-based account-state, reserve, and account-lines helpers instead of the websocket client.
- Cover native XRP spendable-balance and issued-currency trust-line cases with RN unit tests.

## Effort
Small

## Report
https://gist.github.com/gomesalexandre/b8e6fe15a62b7d5a6843b1174225e65f

gomesalexandre:
## Summary
- route React Native `getCoinBalance()` Ripple reads through the RN-safe fetch path instead of the websocket `xrpl` client
- add fetch-based RN helpers for Ripple account state, reserve metadata, and paginated trust lines
- cover native XRP spendable-balance and issued-currency trust-line cases with RN unit tests

## Issue
Closes #1218

## Why
The RN `getCoinBalance()` surface exported by `@vultisig/sdk/react-native` was bypassing the RN Ripple module and importing the websocket core resolver. That is the wrong runtime path under Hermes and also stranded the newer XRPL trust-line balance behavior from the generic RN balance entrypoint.

## Verification
- `yarn build:sdk`
- `yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/platforms/react-native/getCoinBalance.test.ts packages/sdk/tests/unit/platforms/react-native/ripple-rpc.test.ts`
- `yarn build:all` (completed earlier in the verification pass; a lint warning caught during that run was fixed before this final commit)
- targeted post-fix `yarn lint` on the touched RN Ripple files

### Targeted Vitest
```text
18 tests passed across:
- packages/sdk/tests/unit/platforms/react-native/getCoinBalance.test.ts
- packages/sdk/tests/unit/platforms/react-native/ripple-rpc.test.ts
```

### Notes
- This is an RN-specific runtime fix, so there is no meaningful CLI receipt for the changed path.
- I intentionally kept the change scoped to the SDK RN surface and did not widen the public API further than needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed React Native XRP balance retrieval to use the safe fetch-based path.
  * Correctly computes spendable XRP by subtracting required account reserves.
  * Added support for Ripple-issued token balances, including paginated trust line lookup.
  * Unfunded or unavailable XRP accounts now return a zero balance.
  * Malformed RPC responses now fail safely instead of being treated as empty accounts.
* **Tests**
  * Added unit coverage for reserves, issued tokens, unfunded cases, pagination, and malformed responses.


rcoderdev (approved):

NeOMakinG (approved):
Independent review (second reviewer already APPROVED — verified from source, not diff-only). Two lanes: Claude (primary) + Codex `gpt-5.4` adversarial.

## testing approach
- Read the full new RN path (`chains/ripple/balance.ts`, `rpc.ts`) and the removed core-chain resolver (`coin/balance/resolvers/ripple.ts`) side-by-side to confirm **value equivalence**, not just wire mechanics.
- Confirmed the native spendable formula is byte-identical to OLD: `balanceDrops − (reserve_base + ownerCount·reserve_inc)`, floored at `0`; token = matching trust-line balance floored at `0` (negative issuer-owed line → `0`). Endpoint is the same cluster (OLD `wss://xrplcluster.com` → NEW `https://xrplcluster.com`), fetch-safe under Hermes — the actual fix.
- `server_state` reserve fields are drops (integers) — matches OLD `getRippleNetworkInfo` (also `server_state`), so the drops→XRP scale is preserved.
- Ran the targeted suite in a clean worktree: **18/18 green**. Mutation-tested the reserve subtraction (`totalReserve = 0n`) → spendable test goes red → non-inert.
- Fail-safe on transport error verified: `rippleCall` throws on `!res.ok` and on unexpected `status:'error'`; `actNotFound` → clean `0`. No wrong-`0`-on-network-failure.

## blocking
None outstanding. Two fail-**open** regressions vs the OLD path were found and fixed in stacked follow-up **#1400** (targets this branch):
- `getXrpAccountState` returned `funded:false` (→ native `0n`) on a non-`actNotFound`, HTTP-200 response missing `account_data`.
- `getXrpAccountLines` did `result.lines ?? []` (→ token `0n`) on a malformed non-`actNotFound` page.

The core-chain resolver this replaces threw on both shapes, so those turned a malformed/partial upstream envelope into a real-looking `0` balance on a fund-critical surface. #1400 restores fail-closed (throw unless explicit `actNotFound`) + 2 red-on-revert tests (20/20 green).

## should-fix
None blocking merge of this PR.

## suggestion
- Pre-existing (mirrored verbatim from the OLD core-chain resolver, not introduced here — noting for a future sweep, not this PR):
  - `account_lines` pagination re-resolves `ledger_index:'current'` per page; a mid-walk ledger close could miss/double-count a line.
  - The `marker` loop has no progress/repeat guard — a hostile RPC could loop it.
  - `BigInt()` on raw RPC fields throws a bare `SyntaxError` on non-integer input (still fail-closed — no wrong balance, just an opaque error).

## risk
Low. Behavior is value-equivalent to the pre-existing (well-tested) core-chain resolver; the only behavioral delta is the runtime path (websocket→fetch) and the fail-closed hardening in #1400. Native + token + unfunded cases covered by non-inert tests. `getXrpAccountInfo`/`getXrpBalance` back-compat preserved (thin wrappers over the new `getXrpAccountState`), so no other RN consumer breaks.

## verdict
APPROVE. Correct, value-preserving Hermes-safe re-route with genuine test coverage; the fund-safety hardening is shipped in stacked #1400.

---
🤖 dual-lane review (Claude primary + Codex gpt-5.4 adversarial). Fund-safety fixes fixed-in-follow-up per gomes fix-and-approve.

NeOMakinG (approved):
## re-review r2 — delta since 56dd824b (r1 APPROVE @ pass 33)

HEAD advanced to `b863abb8`. True delta since the approved `56dd824b` is a **single commit**: the `#1400` fail-closed fold. The `input.signal`-drop commit `7bb2df5b` actually sits *before* `56dd824b` in the branch history, so it was already inside what r1 approved — re-verified here anyway.

### prior issues — status
|---|---|---|---|---|---|
| r1(Codex) | B1 | blocker | `getXrpAccountState` returned `funded:false`→native `0` on non-`actNotFound` HTTP-200 missing `account_data` (fail-open, OLD threw) | ✅ fixed | `b863abb8`: `actNotFound`→unfunded-0; separate `if (!account_data) throw`. Matches vetted #1400. |
| r1(Codex) | B2 | blocker | `getXrpAccountLines` `result.lines ?? []`→token `0` on malformed page (fail-open, OLD threw) | ✅ fixed | `b863abb8`: `Array.isArray(result.lines)` guard + throw; `actNotFound`→`[]` still precedes so unfunded stays legit-0. |
| r1(Codex) | S1–S3 | suggestion | pre-existing (`ledger_index:'current'`, marker no-progress, raw `BigInt()`) mirrored from OLD resolver | ℹ️ documented | unchanged in delta; fail-closed, future-sweep only. |

### testing approach
- Read the actual `56dd824b..b863abb8` diff + full `rpc.ts` / `balance.ts` source at HEAD.
- (a) fail-closed: **native** (`getXrpAccountState`) now throws on non-`actNotFound` missing `account_data`; **token** (`getXrpAccountLines`) throws on non-array `lines` via `Array.isArray` guard, with `actNotFound`→`[]` preserved above it. Both match the vetted #1400.
- (b) signal-drop: traced `CoinBalanceResolverInput = AccountCoinKey = CoinKey & {address}` — **no** `signal`/`AbortSignal` field anywhere. So `input.signal` was always `undefined`; dropping it is value-equivalent (undefined default) and a genuine type fix. The rpc fns retain their `signal?` param, so any real caller passing a signal is unaffected. No cancellation semantics broken.
- (c) value-equivalence to OLD resolver preserved (same drops math, Hermes-safe HTTPS unchanged).
- (d) 17/17 pass at HEAD; **red-on-revert proven non-inert** — reverting both guards to the old fail-open source turns exactly the 2 new tests red (`no account_data`, `no lines array`).
- Codex second lane: **no new findings** — `lines:[]` stays valid (`Array.isArray([])===true`), no downstream consumer relied on old fail-open, partial-pagination throw is correct fail-closed, `JSON.stringify(result)` is public ledger data (page-bounded, no secret leak).

### risk
Low. Delta is the pre-vetted #1400 fail-closed hardening that strictly *tightens* a fund-display hazard (wrong-looking `0` XRP/token balance), plus a value-equivalent type-only signal-drop already in the approved ancestry.

### verdict
**Re-APPROVE** at `b863abb8`. Delta is the vetted #1400 fail-closed guards + a safe, value-equivalent signal-drop; no new regression.

<sub>reviewed by the gomes review loop — Claude (load-bearing) + Codex gpt-5.4 (second opinion)</sub>

NeOMakinG (approved):
Round 3 — re-review at the merged tip `69434789`. Our r2 APPROVE was reasoned at `b863abb8`; HEAD advanced on 2026-07-20 and GitHub silently re-pointed that review's `commit_id` to the new HEAD, so the approval record read as current when it was not. This round re-establishes it at the tip.

## testing approach

- **Delta identification:** `b863abb8..69434789` is exactly **one commit** — merge of main (`1ee4ddf5`) into the branch. No new author commits.
- **Net-diff-vs-merge-base stability:** compared this PR's net diff at the old approved head (vs `13c3b654`) against the net diff at the current head (vs `1ee4ddf5`), per file, on the `+`/`-` line sets only, ignoring context shift. All 6 files byte-stable (table below).
- **Main-side loss check:** `git log 13c3b654..1ee4ddf5` over the 6 touched files = **0 commits on every one**. The merge had no opposing side in these files, so neither direction of loss is possible.
- **Merge integrity:** `git merge-tree --write-tree 69434789 origin/main` → exit 0, 0 conflict markers (the PR reports `mergeable: UNKNOWN`).
- **Duplicate-identifier check:** the new symbols live in `chains/ripple/{rpc,balance}.ts`, reached only by the dynamic `import('./chains/ripple/balance')` in `getCoinBalance.ts` — no barrel re-export, so no entrypoint collision. Sibling gomes PRs #1192 / #1188 / #1237 touch no ripple file.
- **Suites:** 20/20 green at the merged tip locally, plus an 8-mutation matrix (5 RED / 3 GREEN, below).
- **CI:** all 5 check-runs `success`, each verified `head_sha == 694347894263a8e55e859c5c195ecc9e6ad6badc`.
- Two lanes: Claude (load-bearing) + Codex gpt-5.4.

### net-diff stability, old approved head vs current head

| file | +/- lines @`b863abb8` | +/- lines @`69434789` | result |
|---|---|---|---|
| `.changeset/ten-radios-fix.md` | 6 | 6 | stable |
| `…/react-native/chains/ripple/balance.ts` | 36 | 36 | stable |
| `…/react-native/chains/ripple/rpc.ts` | 163 | 163 | stable |
| `…/react-native/getCoinBalance.ts` | 3 | 3 | stable |
| `…/tests/…/getCoinBalance.test.ts` | 103 | 103 | stable |
| `…/tests/…/ripple-rpc.test.ts` | 25 | 25 | stable |

Byte-stable across the board ⇒ the r2 verdict carries on content; the only new information a merge-only advance can produce is interaction with new main code, covered by the local run + CI above.

## prior issues — status

|---|---|---|---|---|---|
| r1 | B1 | blocker | `getXrpAccountState` returned `funded:false` → native `0n` on a non-`actNotFound` HTTP-200 missing `account_data` (fail-open regression vs the old resolver) | ✅ fixed | folded into the branch at `b863abb8`; re-verified at HEAD — `rpc.ts` now throws `XRP account_info returned no account_data without actNotFound`. Mutation **M7** (revert to fail-open) goes RED. |
| r1 | B2 | blocker | `getXrpAccountLines` `result.lines ?? []` → token `0n` on a malformed page (same fail-open regression) | ✅ fixed | `Array.isArray(result.lines)` guard present at HEAD; `lines:[]` for a funded zero-trust-line account still valid. |
| r1 | S1 | suggestion | `ledger_index:'current'` re-evaluated per pagination page | ℹ️ documented | pre-existing, mirrored verbatim from the old core-chain resolver; not introduced here. |
| r1 | S2 | suggestion | blind marker loop with no progress guard | ℹ️ documented | same — pre-existing, unchanged by this PR. |
| r1 | S3 | suggestion | raw `BigInt()` on RPC fields | ℹ️ documented | still fail-closed (throws), so no wrong balance. Re-confirmed this round: no float round-trip anywhere — `balanceDrops` stays a string into `BigInt()`, reserves are bigint, trust-line values go through `parseIssuedCurrencyValue`. |
| r2 | (a) | — | fail-closed fold matches the vetted #1400 | ✅ fixed | unchanged at HEAD. |
| r2 | (b) | — | signal-drop `7bb2df5b` value-equivalence | ✅ fixed | unchanged at HEAD; `rpc.ts` still accepts `signal?`. |

No prior finding regressed at the merged tip.

## should-fix

The mutation matrix ran 8 mutations against the fund-safety invariants; **5 went RED, 3 stayed GREEN**. The three green ones are the finding — the guards are present and correct in the code, but nothing pins them, so a future refactor can delete any of them silently. All three are in `chains/ripple/balance.ts`.

| mutation | result |
|---|---|
| M1 drop reserve subtraction | RED |
| M2 drop `ownerCount` multiplier | RED |
| M3 drop native floor-at-0 | **GREEN** |
| M4 drop token floor-at-0 | **GREEN** |
| M5 drop issuer match on trust line | **GREEN** |
| M6 drop currency-code normalization | RED |
| M7 revert `account_data` fail-closed → fail-open | RED |
| M8 drop `!account.funded` early return | RED |

1. **M5 — the issuer match on the trust line is untested.** Replacing `account === issuer &&` with `true &&` in the `lines.find(...)` predicate leaves 20/20 green. XRPL currency codes are *not* unique — `USD` exists from many issuers, and an account can hold trust lines for several of them simultaneously. If that clause is ever dropped, `getRippleCoinBalance` returns a *different issuer's* balance for the requested token. The existing fixtures only ever contain a single trust line, so the predicate is never forced to discriminate. Worth a fixture with two same-currency lines from different issuers asserting the correct one is picked.

2. **M3 — the native floor-at-0 is untested.** `return spendableBalance > 0n ? spendableBalance : 0n` → `return spendableBalance` stays green. An account whose `ownerCount` is high enough that `reserve_base + ownerCount*reserve_inc > balanceDrops` is a genuinely reachable state (each trust line/offer raises the reserve), and without the floor the function returns a **negative bigint** balance. The one reserve test uses a balance comfortably above the reserve.

3. **M4 — the token floor-at-0 is untested.** `return balance > 0n ? balance : 0n` → `return balance` stays green. A negative trust-line balance is the normal XRPL encoding for issuer-owed direction, so this floor is load-bearing; both fixtures (`'4'`, `'12.3456789'`) are positive.

None of these are blocking: the code is correct today and r1 already proved value-equivalence against the old core-chain resolver. They are coverage gaps, not defects.

## suggestion

4. `rippleCall` (`…/chains/ripple/rpc.ts:23`) issues a bare `fetch` with only an optional caller-supplied `AbortSignal`, and `getRippleCoinBalance` never passes one — so a stalled XRPL endpoint hangs the RN balance read indefinitely, where the rest of the merged RN fetch stack bounds at `DEFAULT_RN_FETCH_TIMEOUT_MS` (30s) via `withFetchTimeout`.

   Flagged by the Codex lane as a merge-caused regression; **that attribution is wrong and I'm downgrading it accordingly.** `rippleCall` is byte-identical between the pre-PR merge-base `13c3b654` and this HEAD — the PR never touches it, and the file header already documents why it can't use `jsonRpcCall` (the XRPL envelope is not JSON-RPC 2.0). It is pre-existing, not introduced or worsened here. Noting it because `withFetchTimeout` is generic over url/init/timeout/handler and *would* apply here even though `jsonRpcCall` does not — a clean follow-up, in the same family as the still-open #1344.

## risk

Low. The PR's own content is byte-identical to a state already approved twice; the only change is a clean merge of main into a branch whose touched files main did not modify at all. The fund-safety direction is favourable — the path being replaced was a websocket `xrpl` client that does not run under Hermes at all, so the practical delta is broken → working, and both fail-open holes found in r1 are fixed and pinned by red-on-revert tests. Residual risk is confined to the three unpinned guards above, which are correct in the current code.

## verdict

**APPROVE** at `694347894263a8e55e859c5c195ecc9e6ad6badc`. Merge integrity clean in both directions, net diff byte-stable vs the approved state, CI green on this exact SHA, no prior finding regressed. The 3 should-fix items are test-coverage follow-ups and do not block the merge.

---
🤖 posted by the vultisig reviewer bot

NeOMakinG (approved):
Round 4 — merge-integrity re-verification at `d091526b`. HEAD advanced past the r3 approval at `69434789`; the question was narrow: is the advance genuinely merge-only, is the PR's content unchanged, and did the merge silently inert or lose anything. It did not. Re-approving at the tip.

## testing approach

- **Delta audit.** `git log 69434789..d091526b` = **one** new commit, the merge `d091526b` (parents `69434789` + main `20a0fd0d`). Every other commit in the range is reachable from main. No smuggled author commits.
- **Net-diff-vs-merge-base stability (the licensing evidence).** Compared the PR's net diff at the old head (vs merge-base `1ee4ddf5`) against the net diff at the new head (vs merge parent `20a0fd0d`), per file, on the `+`/`-` line sets only (context/offsets ignored). Identical file set, all six byte-stable:

| file | ± lines | old vs new |
|---|---|---|
| `.changeset/ten-radios-fix.md` | 8 | stable |
| `src/platforms/react-native/chains/ripple/balance.ts` | 38 | stable |
| `src/platforms/react-native/chains/ripple/rpc.ts` | 165 | stable |
| `src/platforms/react-native/getCoinBalance.ts` | 5 | stable |
| `tests/unit/platforms/react-native/getCoinBalance.test.ts` | 105 | stable |
| `tests/unit/platforms/react-native/ripple-rpc.test.ts` | 27 | stable |

  Byte-stable ⇒ the r3 verdict carries on content.
- **Main-side loss check.** Exactly one main commit touched a file this PR touches between `1ee4ddf5..20a0fd0d`: `afa17b2a` (#1380, "type balance dispatch as record\<chainkind\>") on `getCoinBalance.ts`. Read the merged result directly — main's retype survives verbatim and the PR's ripple line now sits *inside* it:

  ```
  const resolverLoaders: Record<ChainKind, () => Promise<BalanceResolver>> = {   // main #1380, intact
    ...
    ripple: () => import('./chains/ripple/balance').then(m => m.getRippleCoinBalance as BalanceResolver),  // this PR
  ```

  Independently: because the PR's `+`/`-` set against the *new* parent is byte-identical to the old one, a lost main line would have had to appear as an extra `-` hunk there. It does not.
- **Inerting check (the one that matters).** #1380 narrowed `resolverLoaders`' key type from `Record<string, …>` to `Record<ChainKind, …>` — exactly the shape-change-in-an-adjacent-position class that git can never conflict on. Verified live rather than by reading: the ripple key is a valid `ChainKind`, the dispatch resolves, and the touched-area suite is **20/20 green at `d091526b`** including all three `getCoinBalance — Ripple routing` cases (reserve subtraction, unfunded→0, trust-line balance) and both fail-closed rpc guards.
- **Non-inert proof.** M1 (`totalReserve = 0n`) → **RED**, so the reserve math is genuinely exercised through the merged dispatch.
- **CI at `d091526b`:** `Integration Tests (Vitest)` success with matching `head_sha`; `Unit Tests` and `Lint and Type Check` still in flight at review time (conclusion `null`, matching `head_sha`) — local touched-area run stands in.
- Codex gpt-5.4 second lane, scoped to cross-file inerting and XRPL edge cases: **no findings**.

## prior issues — status

|---|---|---|---|---|---|
| r1 | B1 | blocker | `getXrpAccountState` fail-open: HTTP-200 missing `account_data` → `funded:false` → native 0 | ✅ fixed | folded in via `b863abb8`; re-confirmed at `d091526b` — test "fails closed (throws) on a non-actNotFound response missing account_data" green |
| r1 | B2 | blocker | `getXrpAccountLines` `result.lines ?? []` → token 0 on a malformed page | ✅ fixed | `Array.isArray` guard present at `d091526b`; test "fails closed on an account_lines page missing a lines array" green |
| r1 | S1 | suggestion | `ledger_index:'current'` per pagination page | ⏭️ deferred | pre-existing, mirrored verbatim from the old core-chain resolver; not introduced here |
| r1 | S2 | suggestion | marker loop has no no-progress guard | ⏭️ deferred | same — pre-existing |
| r1 | S3 | suggestion | raw `BigInt()` on RPC fields | ⏭️ deferred | same — still fail-closed (throws), never a wrong balance |
| r3 | M5 | suggestion | issuer discrimination on the trust line is unpinned by tests | ❌ still open | re-measured at `d091526b` — see below |
| r3 | M3/M4 | suggestion | native + token floor-at-0 unpinned | ❌ still open | unchanged by the merge |
| r3 | — | suggestion | `rippleCall` has no RN fetch timeout (main gained `withFetchTimeout` via #1318) | ℹ️ documented | still true, still **not a regression from this PR** — `rippleCall` is byte-identical to `13c3b654` pre-PR; follow-up territory, same family as #1344 |

## suggestion

**1. The issuer half of the trust-line match is still untested (carried from r3, re-measured at the merged head).**

`packages/sdk/src/platforms/react-native/chains/ripple/balance.ts:28-30`:

```ts
const line = lines.find(
  ({ account, currency: lineCurrency }) =>
    account === issuer && toXrplCurrencyCode(lineCurrency) === toXrplCurrencyCode(currency)
)
```

Mutating `account === issuer &&` → `true &&` leaves the suite **20/20 GREEN**. The code is correct; nothing pins it. This matters on XRPL specifically because currency codes are *not* unique — `USD` exists from many issuers and one account can legitimately hold several same-currency trust lines. Every fixture has exactly one line, so `.find` is never forced to discriminate; a future refactor that drops the issuer clause would silently return a *different issuer's* balance for the requested token, and CI would stay green.

Suggested fix: add a fixture with two `USD` lines from different issuers and assert the requested issuer's balance is the one returned.

**2. Both floor-at-0 clauses remain unpinned** (`:23` native, `:35` token). Below-reserve is reachable — each trust line and offer raises `ownerCount` — and a negative trust-line balance is the *normal* XRPL issuer-owed encoding, so both branches are live in production but only ever exercised with comfortably-positive fixtures.

## risk

Low for this round. The advance is a pure merge; the PR's content is byte-identical to what was approved at r3; main's one overlapping change (#1380's `Record<ChainKind>` retype) survives and the merged dispatch is proven live by a green touched-area run plus a red-on-mutation. The open items are all test-coverage hardening on code that is correct as written, and all pre-date this merge.

## verdict

**Approve.** Merge-only advance, net diff byte-stable per file, no main-side loss, nothing inerted. Re-approving so the record is reasoned at `d091526b` rather than carried forward on a re-pointed `commit_id`.

---
🤖 posted by the vultisig reviewer agent

NeOMakinG (approved):
Round 5 — re-approve at the new tip. The advance past r4's `d091526b` is **merge-only**: `9e21b9ea` has parents `d091526b` + main `069176bc` and contributes **zero new author commits**. Re-posting an APPROVE rather than relying on the old one because GitHub re-points `reviews[].commit_id` onto the new head on a merge advance, so the existing decision is not evidence that anyone read this tip.

## testing approach

- **merge-only proof:** `git log --format='%H %P'` on the branch — `9e21b9ea` = merge commit, parents `d091526b` (r4's approved head) + `069176bc` (main). No author commits in the delta.
- **main-side loss check (the cheap decisive form):** `git log 20a0fd0d..069176bc -- <each of the 6 touched files>` = **0 commits on all six**. Main's window is 10 files (`#1455` secure-signing session events, `#1453` rujira FIN limit orders) — **fully disjoint** from this PR's file set.
- **transitive-dep extension (new this round):** the "0 commits on touched files" check does *not* cover what `balance.ts` imports. Ran the same query on `packages/core/src/chains/ripple`, `packages/core/src/coin/balance`, `coin/utils/isFeeCoin.ts`, `packages/lib/utils/assert` → **0 commits on each**. No `package.json`/lockfile in the window either, so no caret-bump hazard from this merge.
- **net-diff-vs-merge-base stability:** per-file `+`/`-` line-set md5 at (old base `20a0fd0d` → old head `d091526b`) vs (new base `069176bc` → new head `9e21b9ea`) — **6/6 byte-stable** (6/36/163/3/103/25 lines). r4's content verdict carries verbatim.
- **not inert, verified live at the merged head:** main still maps `ripple:` to the websocket core-chain resolver; the merged tree maps it to `import('./chains/ripple/balance')` inside main's `Record<ChainKind, …>` typing. Mutation **M1** (`totalReserve = 0n`) → **RED** proves the merged dispatch actually executes the new fetch-safe module.
- **suite:** 20/20 green at `9e21b9ea` (`getCoinBalance.test.ts` + `ripple-rpc.test.ts`), worktree `/tmp/wt-vultisig-sdk-1220`, node_modules symlinked from the canonical checkout.
- **Codex lane (gpt-5.4):** first invocation died on its own 900s budget (0 bytes — infra, not a verdict); re-run with a narrowed brief returned **INFO only, no findings**, independently enumerating all four `0n` return paths as genuine zeros with every network/HTTP/malformed-envelope path throwing.

## prior issues — status

|---|---|---|---|---|---|
| r1 | B1 | blocker | `getXrpAccountState` returned `funded:false` → native `0n` on HTTP-200 missing `account_data` (fail-open) | ✅ fixed | `b863abb8` folded the vetted #1400 fix; throws unless explicit `actNotFound` |
| r1 | B2 | blocker | `getXrpAccountLines` `result.lines ?? []` → token `0n` on malformed page (fail-open) | ✅ fixed | `b863abb8`; `Array.isArray` guard then throw. `lines:[]` for a funded zero-trustline account still valid |
| r1 | S1 | suggestion | `ledger_index:'current'` re-resolved per pagination page | ℹ️ documented | pre-existing, mirrored verbatim from the old core-chain resolver; not introduced here |
| r1 | S2 | suggestion | marker pagination loop has no no-progress guard | ℹ️ documented | same — pre-existing |
| r1 | S3 | suggestion | raw `BigInt()` on RPC-supplied fields | ℹ️ documented | fail-closed (throws), never yields a wrong balance |
| r2 | — | q | did dropping `input.signal` break cancellation? | ✅ resolved | `CoinBalanceResolverInput = AccountCoinKey` has no signal field; rpc fns keep `signal?`. Value-equivalent |
| r3 | M5 | suggestion | trust-line **issuer** discrimination untested (`account === issuer &&` → `true &&` = 20/20 GREEN) | ❌ still open | re-verified GREEN at `9e21b9ea`. Non-blocking (code is correct), but XRPL currency codes are not unique — one account can hold several `USD` lines from different issuers, and nothing pins that we pick the right one |
| r3 | M3 | suggestion | native floor-at-0 untested (below-reserve → negative bigint without it) | ❌ still open | only reserve fixture uses balance ≫ reserve |
| r3 | M4 | suggestion | token floor-at-0 untested (negative = normal XRPL issuer-owed encoding) | ❌ still open | both fixtures positive |
| r3 | — | (codex) | "missing `withFetchTimeout` is a main-side regression from the merge" | ℹ️ refuted | `rippleCall` is byte-identical pre-PR (`git show 13c3b654:<file>`); the PR never authored it. Stands as a repo-wide follow-up, same family as #1344 |
| r4 | — | risk | main's #1380 retyped `resolverLoaders` to `Record<ChainKind, …>` three lines above the PR's hunk (the classic silent-inerting shape) | ✅ verified | re-confirmed at this head: retype survives, PR's `ripple:` entry sits inside it, M1 RED |

## blocking

None.

## preferably-blocking

None.

## should-fix

None.

## suggestion

1. **(carried, r3/r4) pin the trust-line issuer match.** `chains/ripple/balance.ts:27-30` discriminates on `account === issuer && currency`, which is correct — but every fixture has exactly one trust line, so `lines.find` is never forced to discriminate and the mutation `account === issuer &&` → `true &&` stays 20/20 green. One fixture with two same-`currency` lines from different issuers would pin it. This is the strongest of the three unpinned mutations: dropping the clause returns *another issuer's* balance for the requested token.
2. **(carried) pin both floor-at-0 clamps.** `balance.ts:22` (native, below-reserve is reachable — each trust line/offer raises `ownerCount`) and `balance.ts:35` (token, negative is the normal issuer-owed encoding). Both currently unpinned.
3. **(carried, repo-wide) `rippleCall` has no RN fetch timeout.** Main's `withFetchTimeout`/`DEFAULT_RN_FETCH_TIMEOUT_MS` (#1318) is generic over url/init/timeout/handler so it would apply here even though the file header correctly notes `jsonRpcCall` cannot (the XRPL envelope isn't JSON-RPC 2.0). Pre-existing, not authored by this PR — follow-up, not a merge regression.

## q

None.

## risk

**Low.** Zero functional delta since the last approval; the merge has no opposing side on any touched file or transitive dependency. The fund-relevant property — a balance can never silently read as 0 on a network/RPC failure — is preserved: every non-`actNotFound` failure (HTTP error, missing `account_data`, missing reserve info, non-array `lines`, partial pagination) throws, and the four `0n` returns are genuine zeros (unfunded account, no spendable XRP after reserve, no matching trust line, non-positive trust-line balance). The reserve/available distinction is intact (`balance − (reserve_base + ownerCount·reserve_inc)`, all in drops, matching the old `server_state` resolver).

CI at `9e21b9ea`: Integration Tests (Vitest) **success** with matching `head_sha`; Unit Tests and Lint/Type Check still `in_progress` at post time — stating that rather than claiming green.

## verdict

**APPROVE** at `9e21b9ea205e8bf1a13393222fe30811179afb02`. Merge-only advance, provably no functional delta, dispatch verified non-inert at the merged head. The three carried suggestions are test-coverage gaps, not defects, and are not merge-blocking.

---
🤖 automated review — Claude (load-bearing) + Codex gpt-5.4 second lane

PR #1223, opened 2026-07-13
fix(cli): let agent ask sign with VAULT_PASSWORD, no --password needed

neavra:
## Problem

`vultisig agent ask "<send>" --yes` run headlessly with the vault password supplied via `VAULT_PASSWORD` fails at the mid-turn `sign_tx` with `PASSWORD_REQUIRED`, even though that same env var already unlocked the vault for reads and address derivation. Adding `--password <pw>` on argv is the only thing that lets the send sign and broadcast.

This is a self-contradiction in the current contract:

- The CLI help advertises `VAULT_PASSWORD` as the credential "for signing".
- The headless unlock chain added in #899 (cache → OS keyring → `VAULT_PASSWORDS` → `VAULT_PASSWORD`) deliberately de-emphasizes `--password` — it warns that passing the secret on argv leaks it into `ps`/shell history.
- But the sign-time gate still required exactly that `--password` flag: it prompted for a password whenever `config.password` was unset, ignoring that the vault (and the executor) had already been unlocked from the chain at init. In ask mode there is no interactive prompt, so the prompt throws and the whole turn fails.

The net effect: any headless/CI consumer that follows the documented env-var setup silently cannot sign.

## Root cause

The gate keyed off the `--password` flag rather than off whether a password was actually needed. Since `unlockEncryptedVault` already unlocked the vault at init and seeded the executor's password, the executor's re-unlock at sign time is skipped for a standard encrypted-vault session — the password is never needed again after init. The gate was enforcing a requirement the signing path doesn't have.

## Fix

Gate signing on **need**, not on the flag:

- Only prompt when the vault is encrypted, still locked, and the executor holds no password.
- Even then, retry the same non-interactive chain (cache → keyring → `VAULT_PASSWORDS`/`VAULT_PASSWORD`) before falling back to an interactive prompt — so the "never prompt when the chain resolves" behavior holds at sign time too, and a mid-session cleared password is recovered safely.
- Unencrypted vaults skip the gate entirely (there is no password to supply).
- `--password` continues to work unchanged (it is applied to the executor at construction and short-circuits the gate).

The transaction confirmation gate is a separate chokepoint and is untouched — every sign still requires an explicit confirmation first.

Also corrects the `VAULT_PASSWORD` help text to describe the now-consistent contract.

## Verification

- New regression suite (`sessionSignGate.test.ts`) covering: executor-already-unlocked (no prompt, signs), env-resolvable-with-executor-cleared (resolves silently), unencrypted vault (gate skipped), locked-with-nothing-resolvable (ask-mode `PASSWORD_REQUIRED` preserved / interactive prompt still works), and `--password` unchanged. Each signing case also asserts the confirmation gate still fires before signing. Confirmed red on `main` for the three core cases, green with the fix.
- Full CLI suite green (`yarn test:cli`), typecheck, ESLint, and Prettier clean.
- Verified end-to-end against a live agent backend: a headless `agent ask "Send 0.05 POL on Polygon …" --yes` with the password supplied **only** via `VAULT_PASSWORD` (no `--password`) signed and broadcast on the first try — confirmed on-chain (Polygon tx `0x2bcb7ee9…5ee8`, status success, 0.05 POL from the vault address). Before this fix the same invocation failed at `sign_tx` with `PASSWORD_REQUIRED`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Headless signing no longer prompts for `--password` when the vault is already unlocked or credentials are available through supported non-interactive sources.
  * Unencrypted vaults can now be signed without unnecessary password checks.
  * Password prompts remain available when no credentials can be resolved.

* **Documentation**
  * Updated `VAULT_PASSWORD` help text to clarify that it unlocks the vault for reading and signing without prompting.

### DIFF
--- a/.changeset/headless-sign-vault-password.md
+++ b/.changeset/headless-sign-vault-password.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/cli': patch
+---
+
+Fix `agent ask` headless signing so a vault unlocked via `VAULT_PASSWORD` (or the OS keyring) can sign without also passing `--password`. The sign-time gate now keys off whether a password is actually needed — an encrypted vault that is still locked with no password in hand — and retries the same non-interactive chain (cache → keyring → `VAULT_PASSWORDS`/`VAULT_PASSWORD`) before prompting, instead of always re-prompting when `--password` was absent. Unencrypted vaults skip the gate entirely. The transaction confirmation gate is unchanged. Also corrects the `VAULT_PASSWORD` help text.
--- a/clients/cli/src/agent/__tests__/sessionSignGate.test.ts
+++ b/clients/cli/src/agent/__tests__/sessionSignGate.test.ts
@@ -1,0 +1,270 @@
+// Regression tests for the sign-time password gate in runPasswordGatedTool
+// (abts-02 / sdkcli2-10 #3): headless `agent ask --yes` with VAULT_PASSWORD (the
+// documented signing env var) must sign without a --password flag.
+//
+// The bug: the gate re-prompted for a password whenever the `--password` flag
+// (config.password) was unset — ignoring that #899 already unlocked the vault
+// (and seeded the executor's password) from the keyring/env chain at init. In
+// ask mode requestPassword throws, so VAULT_PASSWORD-only signing hard-failed
+// with PASSWORD_REQUIRED even though the secret was already in hand.
+//
+// The fix gates on NEED — an encrypted-but-still-locked vault with no executor
+// password — and retries the same non-interactive chain (cache → keyring →
+// VAULT_PASSWORDS/VAULT_PASSWORD) before ever prompting. The #682 confirmation
+// gate is a separate chokepoint and must still fire before every sign; each
+// success case below asserts requestConfirmation ran before body().
+//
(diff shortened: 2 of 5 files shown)


gomesalexandre (approved):
lgtm.

The fix is the right one. The old sign gate only checked `!this.config.password`, so a session unlocked via `VAULT_PASSWORD` still got re-blocked at sign time in ask mode even though the executor already had the secret. The new `vaultLocked && !this.executor.hasPassword()` gate matches actual need instead of CLI flag shape.

What I verified remotely:
- `unlockEncryptedVault()` already resolves the password non-interactively and seeds it into the executor at init.
- The mid-turn fallback reuses the same non-interactive resolution chain, so it stays aligned with the existing no-prompt contract.
- The confirmation gate still runs before the password gate on the sign paths, so fund-safety ordering is unchanged.
- The new regression cases cover executor-already-unlocked, env-resolvable silent recovery, preserved `PASSWORD_REQUIRED` when nothing resolves, and the interactive path.

Only non-blocking ask from me: add the full Polygon tx hash or explorer link to the PR body/comment since the current runtime receipt is truncated. Not blocking the code review.

PR #1202, opened 2026-07-12
fix(sdk): export toChainAmount from root entrypoint

gomesalexandre:
## Summary
- export `toChainAmount` and `ChainAmountParseError` from the root `@vultisig/sdk` entrypoint
- add public-surface regression coverage so downstreams can stop deep-importing `@vultisig/core-chain/amount/toChainAmount`
- add a changeset for the published SDK/CLI pair

Closes #1203.

## Why
`packages/sdk/src/utils/fiatToAmount.ts` already tells callers to use `toChainAmount`, and multiple SDK paths already depend on that hardened parser internally. But the root `@vultisig/sdk` surface did not export it, which pushes downstreams into deep imports or reimplementation. This PR makes the supported public path match the documented/internal one.

### Targeted regression test
```bash
yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/utils/publicExports.test.ts
```
Result: passed (`Test Files 1 passed; Tests 14 passed`)

### Static / workspace verification
```bash
yarn build:all
```
Result: passed

```bash
yarn lint
```
Result: passed

```bash
yarn test
```
Result: passed (`packages/sdk` + `@vultisig/cli` + `@vultisig/mcp`; final visible summary included `clients/cli` 53/53 files passed, `clients/mcp` 3/3 files passed, command exit 0)

### Requested SDK / CLI build receipts
```bash
yarn build:sdk
```
Result: passed

```bash
yarn cli:build
```
Result: passed (`Built @vultisig/cli v2.19.14`)

### Built-dist runtime import proof from `@vultisig/sdk`
```bash
node --input-type=module -e "import { toChainAmount, ChainAmountParseError } from '@vultisig/sdk'; const scientific = toChainAmount('1.2345e-3', 8).toString(); let emptyErrorName = null; try { toChainAmount('   ', 8) } catch (error) { emptyErrorName = error instanceof ChainAmountParseError ? error.name : String(error && error.name) + ':unexpected'; } console.log(JSON.stringify({ hasToChainAmount: typeof toChainAmount === 'function', hasChainAmountParseError: typeof ChainAmountParseError === 'function', scientific, emptyErrorName }))"
```
Result:
```json
{"hasToChainAmount":true,"hasChainAmountParseError":true,"scientific":"123450","emptyErrorName":"ChainAmountParseError"}
```

### Built CLI smoke against the real backend
```bash
node clients/cli/dist/index.js -o json server
```
Result excerpt:
```json
{
[...]

PR #1225, opened 2026-07-13
fix(sdk): re-export pure parse/tx/decode helpers from RN entrypoint

### LINKED ISSUE #1224: react-native entrypoint omits the root SDK's pure parse / tx-normalize / decode helpers
## Summary
The root `@vultisig/sdk` entrypoint already exports recent pure parse, tx-normalize, and decode helpers, but `@vultisig/sdk/react-native` was still missing them.

## Evidence
- Root entrypoint exports:
  - `packages/sdk/src/index.ts:65-66` (`parseChain`, `parseTicker`, `chainSchema`, `tickerSchema`)
  - `packages/sdk/src/index.ts:106-111` (`normalizeTx`, `splitMultiTx`, `TxNormalizeError`, tx types)
  - `packages/sdk/src/index.ts:675` plus `packages/sdk/src/tools/decode/index.ts:1-4` (`decodeFromToolResult`, `decodeCosmosTx`, `decodeEvmTx`, decode types)
- RN entrypoint omission before fix:
  - `packages/sdk/src/platforms/react-native/index.ts:553-579`

## Impact
First-party RN consumers are pushed toward deep imports or local mirrors for pure helpers that the SDK already owns and exposes elsewhere.

## Architecture angle
This is a classic public-surface parity drift. The helper logic already lives in the SDK and is pure; the RN entrypoint should expose the same contract rather than forcing downstream shims.

## Fix sketch
Re-export the same pure parse / tx-normalize / decode helpers from `@vultisig/sdk/react-native`, add RN entry regression coverage, and ship a changeset.

## Effort
Low.

## Report
https://gist.github.com/gomesalexandre/b5bb3084bdef2f76da0312d5ebea9aac

gomesalexandre:
## Summary
- re-export the root SDK's pure parse helpers from `@vultisig/sdk/react-native`
- re-export the RN-safe tx-normalize / decode helpers and associated types
- add RN entry regression coverage and a changeset

## Test plan
- [x] `yarn workspace @vultisig/sdk typecheck`
- [x] `yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/platforms/react-native/entry.test.ts`
- [x] `yarn build:sdk`
- [x] `yarn cli:build`
- [x] `yarn build:client-shared`
- [x] `yarn build:rujira`
- [x] `node clients/cli/dist/index.js --help | head -n 5`
- [x] `yarn workspace @vultisig/sdk pack --out /tmp/vultisig-sdk-r20-sdk.tgz`
- [x] `tar -xOf /tmp/vultisig-sdk-r20-sdk.tgz package/package.json | jq '.exports["./react-native"]'`

## Receipts
- RN entry regression test now proves `@vultisig/sdk/react-native` re-exports the same parse/tx-normalize/decode helpers as the source modules.
- Packed export-map proof:
  - `./react-native.types = ./dist/index.react-native.d.ts`
  - `./react-native.import/default = ./dist/index.react-native.js`
- CLI runtime smoke after rebuilding shared deps:
  - `node clients/cli/dist/index.js --help | head -n 5`
  - output begins with `Usage: vultisig [options] [command]`

Closes #1224

Report: https://gist.github.com/gomesalexandre/b5bb3084bdef2f76da0312d5ebea9aac


rcoderdev (approved):

NeOMakinG (approved):
## testing approach

Independent Claude correctness + fund-safety lane on an isolated worktree at `d7679f8d` (symlinked canonical `node_modules` + built `dist`). Focus per FUND-CRITICAL SDK export-parity rules: (a) pure by-reference re-export, (b) RN-vs-node parity, (c) Hermes/Buffer crash surface, (d) collision/tree-shaking. Codex gpt-5.4 lane returned no output this session (usage-limited) — best-effort, so this is a load-bearing single-lane pass. Ran the RN entry regression test: **3/3 green** including the new identity assertion.

## blocking

None.

## preferably-blocking

None.

## should-fix

None.

## suggestion

_(informational, non-blocking)_ `decodeCosmosTx` is the first RN-bundle consumer of `cosmjs-types` — it was not previously imported anywhere under `platforms/react-native/`. It's a declared SDK dep and pure protobufjs (Uint8Array, Hermes-safe), so no correctness/crash risk, but because `tools/decode/index.ts` is a barrel, any RN consumer importing *any* decode symbol will now pull `cosmjs-types` into the bundle (Metro won't tree-shake it out). Pure bundle-size note; fine to ship.

## q

None.

## risk

Low. Pure additive export-surface change, +30/-0.
- **By-reference purity verified:** the RN entry re-exports the exact same barrel modules (`../../tools/{decode,parse}`, `../../tx`) that the node/root entry (`src/index.ts` L65-66, L110-111, L673-675) already re-exports. Single source of truth — no forked/re-declared copy that could drift in how a tx is parsed or decoded. The new vitest asserts `rn.X === source.X` via `toBe` (strict reference equality) across all 10 runtime symbols, so it is **non-inert** — a divergent copy would fail red.
- **RN-vs-node parity:** both entrypoints now expose identical helpers from identical modules; no per-entry decode behavior divergence.
- **Hermes safety:** the decode helpers use `globalThis.Buffer` (`tools/decode/evm.ts:149`, `tools/decode/fromToolResult.ts:76/81`), but the RN entry's `preamble.ts` installs `globalThis.Buffer` from the `buffer` package at module-init (and repairs `subarray`). This is the same documented platform contract every existing Buffer-using RN export already relies on (crypto polyfills, `shims/tiny-secp256k1.ts`, `chains/solana/tx.ts:151`). No node builtin (`fs`/`crypto`/`path`/`stream`/`node:`) is imported; `viem` is isomorphic and already in the RN bundle. So no new Hermes crash surface.
- **No name collisions** (typecheck passes; no duplicate export of these symbols in the RN index).

## verdict

**APPROVE.** Clean pure by-reference re-export achieving RN/node parity with a non-inert identity test. Nothing to fix. Independently corroborates the prior APPROVED decision.

---
_Reviewed by Claude (Opus 4.8) — gomes review loop. Worktree test evidence: `entry.test.ts` 3/3 pass._

NeOMakinG (approved):
r2 — merge-advance re-verification at `962b3b5f`. r1 approved `d7679f8d`; the head moved by a single same-author `Merge remote-tracking branch 'origin/main'`. Main added 25 lines to the *same* RN entry file underneath this branch (`#1239`, `#1281`, `#1282`, `#1284`, `#1308`, `#1318`), so a byte-identical net diff was not something I was willing to assume. It holds — but the cross-PR scan this round turned up one thing worth writing down, so read `should-fix` before merging anything else into the RN entry.

## testing approach

Worktree `/tmp/wt-vultisig-sdk-12` at `962b3b5f`, canonical `node_modules` + built `dist` symlinked (incl. `packages/core/{chain,config,mpc}/dist` — an unbuilt `core-config` dist makes this suite fail to *load*, and `VULTISIG_STRICT_SINGLETON=0` is required or `configureMpc` throws a duplicate-engine error that reads as a false RED).

1. **Merge-only proof.** `git log d7679f8d..962b3b5f --not origin/main` returns exactly one commit — the merge, parents `d7679f8d` + `20a0fd0d`. No smuggled author commits.
2. **Net-diff-vs-merge-base stability.** Old base `c3e73e74`, new merge parent `20a0fd0d`. `+`/`-` line multiset, context stripped:

   | file | old `±` md5 | new `±` md5 | verdict |
   |---|---|---|---|
   | `.changeset/rn-public-surface-parity-r20.md` | `7a557adedd5f` | `7a557adedd5f` | STABLE |
   | `packages/sdk/src/platforms/react-native/index.ts` | `7c8d463ca3bd` | `7c8d463ca3bd` | STABLE |
   | `packages/sdk/tests/unit/platforms/react-native/entry.test.ts` | `3da3c948a8ba` | `3da3c948a8ba` | STABLE |

   30 `±` lines both sides, byte-identical, all additive. The prior verdict carries on content.
3. **Main-side loss check.** Enumerated every `+` line main added to the RN entry across `c3e73e74..20a0fd0d` (the `QueryUrlOptions` widening of the `rpcFetch` type export, the CosmWasm Amino builders, `SUI_NATIVE_COIN_TYPE`, `TRC20_TRANSFER_SELECTOR`, `CONSOLIDATE_CHAINS`, `fromChainAmountExact`, `getBlockExplorerUrl`, `computePersonalSignHash`/`formatEcdsaSignature65`, `resolveChainReference`) and grepped each for survival in the file at `962b3b5f`: **all present**. Independently, the PR's `-` set is empty and byte-stable against the new parent, so a swallowed main line would have shown as an extra `-` hunk.
4. **Published-reachability check — artifact, not the vitest alias.** The vitest config aliases workspace packages to *source*, so it structurally cannot catch exports-map or bundler drift. `./react-native` in `packages/sdk/package.json` resolves every condition to `./dist/index.react-native.js`, whose rollup input (`rollup.platforms.config.js:373-378`) is exactly the file this PR edits, with no allow-list. I built the real artifact (`BUILD_TARGET=react-native npx rollup -c rollup.platforms.config.js`, rc=0) and grepped the emitted export tail: all 10 runtime symbols are present and exported — `decodeCosmosTx`, `decodeEvmTx`, `decodeFromToolResult`, `parseChain`, `parseTicker`, `normalizeTx`, `splitMultiTx`, `TxNormalizeError`, plus `chainSchema`/`tickerSchema` under mangled aliases (`ZT as chainSchema`, `XT as tickerSchema`). Reachable by a published consumer, not just by the test alias.
5. **Duplicate-identifier check.** Each of the 19 new names occurs exactly **once** in the RN entrypoint at this head. (Cross-PR result: see `should-fix`.)
6. **Mutation matrix at the merged head** (`packages/sdk` → `VULTISIG_STRICT_SINGLETON=0 vitest run tests/unit/platforms/react-native/entry.test.ts`, 6/6 green unmutated):

   | # | mutation | result |
   |---|---|---|
   | M1 | drop the `../../tools/decode` value re-export | **RED** |
   | M2 | drop the `../../tools/parse` value re-export | **RED** |
   | M3 | drop the `../../tx` value re-export | **RED** |
   | M4 | drop **all three** `export type {…}` lines | 🟢 **GREEN** |
   | M5 | keep the name but fork `parseChain` behind a wrapper (breaks reference identity, keeps the export) | **RED** |

   M1/M2/M3/M5 RED confirms the runtime half is live and that the test really is a `toBe` identity assertion, not a mere presence check. M4 GREEN is the honest gap: the 9 type-only re-exports are not netted by any runtime test — inherent, since types erase before the assertion can see them. `tsc` on a consumer is the real net there and `Lint and Type Check` is green at this head, so I am not asking for anything; just recording that the matrix has a green cell.
7. **Lanes.** Claude (load-bearing, above) + Codex `gpt-5.4`. Codex returned one **High** on `decodeFromToolResult` fail-opening through permissive `Buffer.from(…, 'hex'|'base64')`. I reproduced the mechanism but **could not reproduce the stated impact**, so I have downgraded it and written up both halves under `pre-existing bugs i bumped into` — including the adversarial experiment that refuted it.

## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | INFO | `decodeCosmosTx` newly pulls `cosmjs-types` into the RN bundle — flagged as "bundle-size only, non-blocking" | ⚠️ still true, now quantified | I built the RN artifact with and without the PR: **2,456,119 → 2,751,867 bytes, +295,748 (+12.0 %)**. Sourcemap `sources` diff attributes it to 21 newly-inlined modules: **14 from `cosmjs-types`**, 3 `tools/decode`, 2 `tools/parse`, 1 `tx/normalize.ts`, 1 `utils/addressFormat.ts`; nothing removed. Still non-blocking, but "bundle-size only" understated it — see `suggestion`. |
| r1 | 2 | LOW (Codex) | `tools/decode` (`evm.ts`, `fromToolResult.ts`) relies on ambient `globalThis.Buffer` installed by the RN entry preamble — fragile only for a hypothetical direct-subpath import | ✅ still mitigated | Re-verified at `962b3b5f`: `platforms/react-native/preamble.ts:22-25` still installs `globalThis.Buffer` (with the `subarray` repair) at module-init, and `index.ts:9` still documents the contract. This PR adds no new subpath entry, so the surface is unchanged. |
| r1 | 3 | LOW (Codex) | viem encoding path uses `TextEncoder`/`TextDecoder` on Hermes | ✅ still mitigated | Unchanged: `viem` is in the RN rollup `external` list and is already exercised on Hermes today via the existing RN `evm/tx.ts` + `rpc.ts`. No new Hermes surface. |
| r1 | 4 | (positive finding) | re-exports are pure by-reference from the same barrels the root entry uses; `toBe` test is non-inert | ✅ re-confirmed | M5 above forks `parseChain` behind a wrapper while keeping the export name — RED. Reference identity is genuinely enforced after the merge. |

## blocking

None.

## preferably-blocking

None.

## should-fix

**1 — cross-PR duplicate-export collision with #1407: merges clean, breaks `main` after the *second* merge, and is invisible to either PR's CI.**

This is not a defect in this PR — it is a merge-order hazard, and I am recording it here because this PR is ready and #1407 is still a draft, so this one lands first and #1407 inherits the fix.

`#1407` ("fix(sdk): export RN pure validation/normalization/policy helpers", head `71e1f16c`, draft) adds to the *same* RN entrypoint:

```
export type { AmountUnits, AssetRef, Envelope, FieldDiff, … } from '../../tools/policy'
export type { NormalizeArgs, NormalizedTx } from '../../tx'
export { normalizeTx, splitMultiTx, TxNormalizeError } from '../../tx'
```

The last two lines are byte-identical to two of this PR's lines, and `AssetRef` / `Envelope` collide by *name from a different module* (`tools/policy/types` vs this PR's `tools/decode/types`) — which is the worse half, because it is a genuine ambiguity, not a redundant duplicate.

I simulated the two-step merge rather than eyeballing it. `git merge-tree --write-tree 20a0fd0d 962b3b5f` → synthesize `main+#1225` → `git merge-tree --write-tree <that> 71e1f16c`: **rc=0, no conflict**, because the two PRs insert at different offsets so git never sees an overlap. The resulting file has the `../../tx` exports at lines 599-600 **and again** at 619-620. Running `tsc --noEmit` on that exact tree:

```
src/platforms/react-native/index.ts(565,15): error TS2300: Duplicate identifier 'AssetRef'.
src/platforms/react-native/index.ts(565,65): error TS2300: Duplicate identifier 'Envelope'.
src/platforms/react-native/index.ts(599,15): error TS2300: Duplicate identifier 'NormalizeArgs'.
src/platforms/react-native/index.ts(599,30): error TS2300: Duplicate identifier 'NormalizedTx'.
src/platforms/react-native/index.ts(600,10): error TS2300: Duplicate identifier 'normalizeTx'.
src/platforms/react-native/index.ts(600,23): error TS2300: Duplicate identifier 'splitMultiTx'.
src/platforms/react-native/index.ts(600,37): error TS2300: Duplicate identifier 'TxNormalizeError'.
… (9 distinct identifiers, 14 errors)
```

This is the exact shape that already bit `#1188` + `#1192` on `getEvmChainByChainId`/`getEvmChainId`. Neither PR's CI can catch it, because each branch typechecks fine in isolation.

Suggested fix (on **#1407**, not here): drop these two lines from #1407's RN-entry hunk once this PR merges —

```diff
-export type { NormalizeArgs, NormalizedTx } from '../../tx'
-export { normalizeTx, splitMultiTx, TxNormalizeError } from '../../tx'
```

— and resolve the `AssetRef`/`Envelope` clash deliberately, e.g. re-export the policy ones under distinct names:

```diff
-export type { AmountUnits, AssetRef, Envelope, FieldDiff, … } from '../../tools/policy'
+export type { AmountUnits, FieldDiff, … } from '../../tools/policy'
+export type { AssetRef as PolicyAssetRef, Envelope as PolicyEnvelope } from '../../tools/policy'
```

I checked the other open gomes SDK PRs that touch this file (`#1412`, `#1389`, `#1277`, `#1202`, `#1193`, `#1192`) plus `#1188`, `#1220`, `#1237` — **#1407 is the only collision**.

## suggestion

**2 — the RN bundle grows 295 KB (+12.0 %) for helpers many RN consumers will never call.**

Measured, not estimated (see the prior-issues table for the method). The RN rollup config inlines everything not in its `external` list, and `cosmjs-types` is not in that list, so `decodeCosmosTx` drags 14 modules into every RN consumer's bundle whether or not they decode a Cosmos tx. Metro cannot tree-shake this away for barrel consumers.

If that matters for Station, the cheap fix is to keep `decodeCosmosTx` behind the existing lazy-import pattern this file already uses elsewhere (`fiatToAmount` at `index.ts:592-596` does exactly this):

```ts
export async function decodeCosmosTx(...args: unknown[]) {
  const mod = await import('../../tools/decode')
  return mod.decodeCosmosTx(...(args as Parameters<typeof mod.decodeCosmosTx>))
}
```

Non-blocking — I would not hold parity for it, and it changes the export's signature to async, so it is a deliberate trade rather than a free win. Flagging it so the number is on the record.

**3 — the merge bisected a comment paragraph.** Main wrote `// Pure helpers — no chain client deps` and the `//` + `// Exact base-units -> human decimal-string conversion …` block as one paragraph; this PR's two `tools/decode` lines were inserted between them, so the continuation comment now dangles under an unrelated export. Purely cosmetic; move the two `tools/decode` lines below the paragraph if you touch the file again.

## q

**4 — both newly-exposed families are documented fail-open / fail-soft. Is that contract carried to RN consumers anywhere they will see it?**

Not a defect and not new to this PR — the root/node entry already exposes all of these, so this only extends the same surface to RN — but it is the one thing worth being deliberate about, given this SDK's WYSIWYS history:

- `tools/parse/index.ts` says so in its own header: *"Safety: fail-open design — the Zod schemas are MAXIMALLY PERMISSIVE … When in doubt, accept and let the downstream tool handle it"*, and `tickerSchema`/`parseTicker` are explicitly **format-only** — they do not resolve a ticker to a contract or to token metadata. An RN caller that reads `parseTicker(x).success` as "this is a real, resolved token" would be wrong.
- `tools/decode` fails *closed* into a value rather than a throw: `Envelope { decoded: false, decodeError }` (`cosmos.ts:31`, `evm.ts:156`/`:210`, `fromToolResult.ts:17`). A caller that renders the envelope without first checking `.decoded` gets a zero-value envelope, not an error.

Would you consider a one-line note in the changeset, or a short `@remarks` on the RN re-export block, saying that `decoded` must be checked and that `parseTicker` is format-only? Cheap insurance for whichever RN screen picks these up first.

**5 — `decodeFromToolResult` accepts malformed payload strings instead of rejecting them (pre-existing; not introduced here, and already reachable from the root/node entry).**

Codex flagged this as High. I reproduced the **mechanism** and then failed to reproduce the **impact**, so I am reporting it at hardening tier with both results, rather than relaying the High.

*Mechanism — confirmed.* `hexToBytes` (`fromToolResult.ts:50`) and `base64ToBytes` (`:56`) both use bare `Buffer.from`, whose Node semantics silently stop at the first invalid character rather than throwing. Verified on Node 20 **and** on the `buffer` npm polyfill that RN actually uses (identical behaviour — so no RN-specific divergence):

```
Buffer.from('deadbeef',   'hex') -> deadbeef
Buffer.from('deadbeefgg', 'hex') -> deadbeef      # junk suffix dropped
Buffer.from('deadbeeff',  'hex') -> deadbeef      # odd trailing nibble dropped
Buffer.from('deadZZbeef', 'hex') -> dead          # TRUNCATES at the bad char
Buffer.from('AQIDBA==$$$','base64') -> 01020304   # junk suffix dropped
```

*Impact — I could not demonstrate a wrong envelope.* The worry that matters for a WYSIWYS oracle is a malformed payload decoding into a **plausible but different** transaction. I built a real 2-message Cosmos `TxRaw` (two `MsgSend`, different recipients and amounts), hex-encoded it, and pushed junk through every path:

```
clean              -> msgs=2  cosmos1aaa…:1000, cosmos1bbb…:9999999
junk @50/60/70/80/90/95%  -> THROW "premature EOF"   (⇒ decoded:false, fail-closed)
suffix 'gg'        -> msgs=2  cosmos1aaa…:1000, cosmos1bbb…:9999999   (identical)
odd trailing nibble-> msgs=2  cosmos1aaa…:1000, cosmos1bbb…:9999999   (identical)
```

Mid-string truncation — the only variant that can *change* the byte string — is caught by protobuf's length prefixes and surfaces as `Envelope{decoded:false}`, which is the correct fail-closed outcome. Suffix junk and odd nibbles can only *drop* trailing garbage, so the decoded envelope is byte-for-byte the true transaction. EVM is if anything stricter, since RLP is length-prefixed too and viem throws on a length mismatch. So the real defect is "a malformed input string is not rejected", not "a wrong recipient/amount is displayed" — a robustness gap, not a fund-safety one, and I did not find a path from one to the other.

*Suggested hardening, for a follow-up rather than this PR:* validate the shape before converting, so junk is rejected instead of normalized —

```ts
function hexToBytes(hex: string): Uint8Array {
  const clean = /^0[xX]/.test(hex) ? hex.slice(2) : hex
  if (!/^([0-9a-fA-F]{2})*$/.test(clean)) throw new Error('malformed hex payload')
  return Uint8Array.from(Buffer.from(clean, 'hex'))
}
```

plus a strict base64 alphabet/padding check in `base64ToBytes`. Both throw sites are already inside the decoder's `try`, so this converts silent leniency into a clean `decoded:false` with a `decodeError`. Worth an issue; entirely out of scope for a re-export parity PR.

(One adjacent nit I checked and dismissed: `tryBase64ThenHex` (`:84`) routes any payload matching `/^[0-9a-fA-F]+$/` down the hex path, so a base64 payload composed only of hex characters would be misread. Probability is `(16/64)^n` for an n-char payload, and the misread bytes fail proto decode anyway ⇒ `decoded:false`. Fail-closed, not worth changing.)

## risk

Low. Additive re-export parity only — no behaviour change to any existing symbol, no new module authored, everything by-reference from barrels the root entry already exposes, so there is a single source of truth and no forked copy to drift. Nothing here touches keysign, derivation or broadcast. `Lint and Type Check` **pass** and `Integration Tests (Vitest)` **pass** at `962b3b5f`; `Unit Tests` still pending; `CodeRabbit` `fail` is the usual "Review rate limited" artifact, not a code signal.

The residual items are (a) a +12 % RN bundle, quantified above, (b) the #1407 merge-order hazard, which is a hazard for *that* PR to absorb, not a reason to hold this one, and (c) the pre-existing `Buffer.from` leniency in `decodeFromToolResult`, which I could not turn into a wrong envelope and which is already reachable from the root entry today.

## verdict

**APPROVE** — re-affirmed at `962b3b5f`. Merge-only advance, net diff byte-stable per file, every main-side hunk survives, the exports are reachable in the real built artifact (not just via the vitest source alias), and 4 of 5 mutations go RED. The `should-fix` is a cross-PR note for #1407 and does not block this merge — but whoever merges #1407 must drop the two duplicated `../../tx` lines and rename the `AssetRef`/`Envelope` clash first, because git will not conflict and neither CI will catch it.

---
🤖 automated review — Claude (load-bearing) + Codex gpt-5.4 two-lane

NeOMakinG (approved):
r3 — merge-advance re-verification only. `962b3b5f` → `57092059` (one same-author `Merge branch 'main'`, parents `962b3b5f` + `f5e5ade8`). **Zero functional delta; the merge resolution is provably lossless.** Re-approving at the current tip.

## testing approach
- **`+`/`-` multiset md5 of the net-diff-vs-merge-base, old tip vs new tip:** `20a0fd0d..962b3b5f` → `ab1345259a8ce40fd472bebac6e45c89`; `f5e5ade8..57092059` → `ab1345259a8ce40fd472bebac6e45c89`. **Byte-identical** ⇒ zero new author work AND zero main-side loss in the resolution, in one check. Net diff is still exactly 3 files / +30 / -0.
- **Main-side window `20a0fd0d..f5e5ade8`** = 5 commits (#1237, #1277, #1202, #1453, #1455). Four of them touch `platforms/react-native/index.ts` — so the resolution genuinely mattered here, and the md5 identity above is what proves nothing of theirs was dropped.
- **Import set untouched:** `src/tools/decode`, `src/tools/parse`, `src/tx` — 0 commits in the window, and tree hashes are `SAME` at `57092059` vs `f5e5ade8` (`7c30d851…` / `d1827371…` / `496288a5…`). The re-exported implementations did not move under the PR.
- **Root barrel blob identity** (the cheap sdk#1237-r5 check): `57092059:packages/sdk/src/index.ts` = `f5e5ade8:…` = `a062d19f…`. Identical blob ⇒ the resolution introduced no root-barrel change at all.
- **Duplicate-export scan on the MERGED surface**, both barrels: RN entry 353 exported names / 353 unique; root entry 684 / 684. **No collision.** Independently reproduced by the Codex lane.
- **eslint on the merged barrels**, real repo config (`--config .config/eslint.config.mjs`, both files): **rc=0** — `simple-import-sort/exports` does not flag the added lines' post-merge positions.
- **`tsc --noEmit --skipLibCheck`** on merged tree: 2 errors, both `TS2353` in `packages/core/mpc/**` (`auxiliaryData` / `ISigningInput`). **Control run at the main parent `f5e5ade8` reproduces byte-identical output** ⇒ local walletcore-dep artifact, not PR-attributable. **Zero `TS2300` duplicate-identifier errors** — that was the specific hazard.
- **Suite:** `tests/unit/platforms/react-native/entry.test.ts` on the merged tree — **8/8 green** (was 6 at r2; main added 2). Env: `VULTISIG_STRICT_SINGLETON=0` + symlinked `core/{chain,config,mpc}/dist`.
- **No `package.json` / `pnpm-lock.yaml` in the net diff** ⇒ no caret-bump hazard from this merge.
- Also confirmed the two exports main landed today reach the RN name list: `toChainAmount` (#1202, RN L618) and the XRPL issued-currency canonicals (#1277, RN L76-85). `getSighash*` (#1460) is not on main yet, so nothing to check there.

## prior issues — status
|---|---|---|---|---|---|
| r2 | 1 | should-fix (cross-PR) | duplicate/clashing exports vs #1407 on the same RN entry — 9 identifiers, 14× `TS2300` when both land; `AssetRef`/`Envelope` come from a *different* module there (`tools/policy/types` vs `tools/decode/types`) | ⏭️ deferred | unchanged: #1407 still OPEN **draft** at the same head `71e1f16c`. Burden stays on the later PR. Nothing for #1225 to do. |
| r2 | 2 | info | RN bundle +295,748 B (+12.0%), 14 of 21 new modules from `cosmjs-types`, which is not in the RN rollup `external` list | ℹ️ documented | unchanged by the merge (import set byte-identical). Accepted as the cost of parity. |
| r2 | 3 | info | mutation M4 green — the three `export type {…}` lines are structurally un-nettable by a runtime `toBe` test; `tsc` is the only net | ℹ️ documented | unchanged. `tsc` runs in CI, so the net exists. |
| r1/r2 | 4 | pre-existing | `decodeFromToolResult` hex/base64 helpers use bare `Buffer.from` (lenient: junk truncates silently) | ℹ️ documented | unchanged — `tools/decode` tree hash identical to main. Reproduced at r2 as **fail-closed** (`premature EOF` → `decoded:false`); robustness gap, not fund-safety, and reachable from the root entry independently of this PR. |

## blocking
None.

## preferably-blocking
None.

## should-fix
None.

## suggestion
None — nothing in this delta to act on.

## q
None.

## risk
Low, and strictly lower than r2's. The advance is a pure merge commit with an md5-identical net diff, an identical root barrel blob, and an identical import-set tree. The one live hazard on this PR (the #1407 name clash) is unchanged and lands on #1407, not here.

Note for merge: `reviewDecision` was already `APPROVED` before this round — the current `mergeStateStatus: BLOCKED` is from **pending CI** (`Unit Tests`, `Lint and Type Check` still running; `CodeRabbit` is `fail` only because it rate-limited its own review), not from a missing approval. This round makes the approval unambiguous at `57092059`.

## verdict
**APPROVE.** Clean merge, zero functional delta, nothing to fix.

### lanes
- **Claude (load-bearing):** full pass as above.
- **Codex `gpt-5.4`:** scoped to the single merged-barrel duplicate/omission question. Returned independently — no duplicate exported identifier on the merged RN entry, and `simple-import-sort/exports` would not flag the added lines. Corroborates lane 1; no new findings, nothing promoted.
- **GLM:** retired, not run.

---
*🤖 automated review — vultisig-ops reviewer*

PR #1233, opened 2026-07-13
fix(cli): classify committed broadcasts as partial success

neavra:
## Summary

- add exit code `13` / `BROADCAST_COMMITTED` for `agent ask` when a live transaction hash exists but a later typed outcome, SSE error, or thrown follow-up leaves the overall request incomplete
- preserve transaction records, typed outcomes, and original backend diagnostics in the v1 JSON envelope
- replace misleading retry prose in human output with explicit hash/status details and a do-not-blindly-retry warning
- document the additive contract and include an `@vultisig/cli` patch changeset

## Safety contract

This remains nonzero because an approval or other leg may have broadcast without the full requested flow succeeding. It is also non-retryable: callers must inspect the emitted hashes before continuing. Failed-only transaction sets retain their existing failure classification.

## Validation

- `yarn cli:build`
- `yarn workspace @vultisig/cli test` (699 tests)
- `yarn check`
- built CLI help confirms exit 13 and the no-blind-retry guidance
- two local team-review passes completed; final delta approved (review kept local)
- all required GitHub checks green

## Companion backend PR

This is the CLI half of a two-part change; the backend half is
vultisig/agent-backend-ts#1224, which makes the server's typed `turn_outcome`
truthful for the same scenarios (it emits the `broadcast_committed` outcome this
PR classifies as exit 13). Each half is additive and safe to land alone — this
PR also detects committed broadcasts from the transaction records themselves, so
it degrades gracefully against an older backend — but the full end-to-end
contract needs both. Review together.

## Runtime verification

Initially covered through the CLI command contract tests (JSON and human output
paths). Since then, a live end-to-end run against a funded test vault (with the
companion backend PR) confirmed the contract on a real Polygon transaction:
the send confirmed on-chain
(`0x96dec582dbebd8572354c00c72e09fe6f761c895a408e00786c7224d6ebf25ab`), the wire
outcome arrived as `broadcast_committed`, and the CLI exited 13 with the hash
preserved in `transactions[]` and the do-not-blindly-retry warning shown.

## Summary by CodeRabbit

* **New Features**
  * Added explicit handling for partial success after a transaction broadcast.
  * Preserved transaction hashes and original error details in JSON and human-readable responses.
  * Added a dedicated exit code and error classification for broadcast-committed outcomes.
  * Added clear warnings to inspect transaction hashes and avoid blindly retrying.

* **Documentation**
  * Updated CLI examples, error-code references, exit-code tables, and command help.

* **Tests**
  * Expanded coverage for broadcast, acknowledgment, transaction, and retry scenarios.


gomesalexandre (approved):
looks good. this matches the backend companion contract in agent-backend-ts#1224 and keeps the cli fail-closed after a live hash exists:
- any non-success typed outcome after a live tx now exits 13 / `BROADCAST_COMMITTED` instead of pretending the whole request succeeded or inviting a blind retry
- thrown follow-up failures preserve the original diagnostic under `data.original_error` while still reclassifying the public error correctly
- `ACK_FAILED` stays reserved for the narrower unacknowledged-broadcast case
- failed-only tx sets keep their original failure code instead of being over-promoted to committed-broadcast

the added tests cover the important boundaries, including blocked/refusal/error endings after broadcast and mixed failed/live tx sets.

approve.

PR #1239, opened 2026-07-13
fix: exact bigint balance formatting past 22 decimals + expose the helpers on root and rn entries

gomesalexandre:
## what

1. **Exact bigint balance formatting.** `BigInt(10 ** decimals)` computes the divisor via a float power - exact only for decimals <= 22. At >= 23 the divisor is silently wrong: e.g. raw `1234567890123456789012345n` at 24 decimals formatted as `1.234567890123456805789561` (wrong past the 16th significant digit; `10 ** 24` floats to `999999999999999983222784`). Both hand-rolled copies (`packages/sdk/src/adapters/formatBalance.ts`, `clients/cli/src/ui.ts`) now delegate to core-chain's pure-bigint `fromChainAmountExact` behind a thin wrapper that preserves the legacy display contract byte-for-byte for <= 22 decimals (trailing-zero trim, '0' for zero, no fraction for wholes - parity covered by hand-computed tests incl. leading-zero fractions, dust, decimals=0).
2. **Root barrel exports** for `fromChainAmountExact` + `getBlockExplorerUrl`. Both consumers of the sdk hand-roll these today (the explorer-url table has drifted from the sdk copy on 10+ chains - cross-repo audit); the exports are the enabler to retire those copies. `getBlockExplorerUrl` riding along is deliberate for that reason, not scope creep.
3. **React-native entry parity** (review catch): the RN platform entry is a hand-curated allow-list, so root-only exports never reach Metro consumers - the recurring class tracked in #1224. Both exports added to the RN entry + a new RN-entry parity test that imports the platform entry file directly (the existing publicExports test resolves the node condition and structurally cannot see this gap). Partially addresses #1224 and establishes the guard pattern for it.

## receipts

- sdk suite 2767/2767 (the +1 is the new RN-entry parity test), core 1703/1704 (1 pre-existing unrelated skip), typecheck + lint clean, build green, cli build + smoke green, changeset guard green.
- Built-artifact proof: `dist/index.react-native.js` + `.d.ts` grepped - both exports present in the RN bundle Metro actually resolves.
- New tests document the exact wrong-output the old float path produced at 24/30 decimals (values hand-computed, not derived from the old code).

## risk

Low. No live asset reaches 23 decimals today (fee coins are all <= 18), so runtime display behavior is provably unchanged - the parity tests pin <= 22 byte-for-byte, and the fix removes the latent corruption for the first high-decimal asset that lands. Exports are additive. Changeset: patch @vultisig/sdk + @vultisig/cli.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved balance/amount formatting accuracy for tokens with more than 22 decimals by switching to exact bigint-based conversion (eliminates float precision drift).
  * Kept existing formatting behavior for `0`, whole numbers, and trailing-zero trimming, including dust-only fractional values.
  * CLI UI amount formatting now uses the same exact conversion for consistent display.
* **New Features**
  * Added root-level SDK access to `fromChainAmountExact` and `getBlockExplorerUrl` (including React Native availability).
* **Tests**
  * Expanded unit coverage for high-decimal formatting and validated the new public exports, including React Native entry behavior.


gomesalexandre (commented):
remote conceptual pass on the current head.

looked hardest at `packages/sdk/src/adapters/formatBalance.ts`, `clients/cli/src/ui.ts`, the new root/RN exports in `packages/sdk/src/index.ts` and `packages/sdk/src/platforms/react-native/index.ts`, and the high-decimal regression coverage.

what checked out:
- replacing the hand-rolled `BigInt(10 ** decimals)` divisor with `fromChainAmountExact` is the right correctness fix. the old float-power path really does drift once decimals get past 22
- both call sites keep the old display contract after the swap: zero stays `"0"`, whole values stay whole, and trailing fraction zeros are still trimmed
- exporting `fromChainAmountExact` and `getBlockExplorerUrl` from both the root barrel and the RN entry closes the usual root-vs-rn surface gap instead of only fixing node consumers
- the new RN entry test is the right guard for this class because the existing public-exports test cannot see Metro's react-native condition

no new blocker from the remote read. i did not rerun the sdk cli/runtime receipts locally this pass because the working clone is busy/dirty, so the usual built-cli proof is still owed before merge.

rcoderdev (approved):

NeOMakinG (approved):
Independent review (a prior reviewer's decision was already APPROVED — verified from scratch, not rubber-stamped). vultisig-sdk is fund-critical, so I read the actual formatting helper + exports and drove the tests at runtime with a red-on-revert, rather than reasoning from the diff.

## testing approach

Worktree at HEAD `44a6cc86`. Read the load-bearing core helper source (not just the diff), traced every wrapper branch, verified the export surface against the real `package.json` exports map, and ran the touched test files with a red-on-revert to prove non-inertness. Second lane (Codex gpt-5.4) ran but returned a polluted/unusable output this pass — findings below are the load-bearing Claude lane.

## blocking

None.

## preferably-blocking

None.

## should-fix

None.

## suggestion

None blocking. The fix is tight and the tests are the right shape.

## q

None.

## risk

**Low, and I'll show my work — this touches how a balance/amount is displayed to the user (a fund decision), so it earned the scrutiny.**

- **Exactness (the whole point):** `fromChainAmountExact` is pure bigint→string arithmetic (`raw.toString().padStart(decimals+1,'0')` then slice whole/fraction). No `Number()`, no float power, no `.toFixed()` — so it never loses precision past 2^53 or past 22 fraction digits, and it never rounds or truncates a significant digit. That means it's safe even where the CLI reuses it on quote **amounts** (`estimatedOutput`, `maxSwapable`, fees in `swap.ts`) — WYSIWYS holds, it can't round *up* into more than the user has.
- **Wrapper parity:** both rewritten formatters keep the legacy display contract byte-for-byte across every branch — zero → `'0'`, whole → no fraction, dust → leading-zero whole via padStart, trailing-zero trim, and `decimals=0` (no dot → `fraction` undefined → return whole). Confirmed by the parity tests, which are hand-computed, not derived from the old code.
- **No missed sibling:** the *other* `formatBalance` (`packages/sdk/src/tools/balance/rpc.ts`) uses `10n ** BigInt(decimals)` — an **exact** bigint power, not the buggy `BigInt(10 ** decimals)` float power — so it was already correct and is rightly left untouched. The bug class was specifically the float-then-BigInt form, and both instances of it are fixed.
- **Exports:** root + RN re-exports are pure by-reference (`export { x } from '@vultisig/core-chain/...'`), no divergent copy. Both subpaths (`./amount/fromChainAmountExact`, `./utils/getBlockExplorerUrl`) are present in the `@vultisig/core-chain` `package.json` exports map, and the mechanism mirrors the existing `getSwapExplorerUrl` root re-export. The RN-entry parity test imports the platform entry *file* directly, which is the correct guard for the hand-curated-allow-list gap class (#1224) that the node-condition `publicExports` test structurally can't see.
- **No new crash path:** `fromChainAmountExact` throws on negative/non-integer, but `formatBigintAmount` takes a `bigint` (can't be non-integer), quote amounts are always ≥0, and `formatBalanceAmount` wraps `BigInt()` in try/catch with a raw-string fallback — so the throw is strictly fail-closed-safer, never a regression on a real input.
- **Runtime evidence:** `formatBalance.test.ts` 33/33, `publicExports.test.ts` green (root exports resolve + produce correct output), RN `entry.test.ts` 3/3 green (RN exports resolve + correct). **Red-on-revert:** restoring the old `BigInt(10 ** decimals)` divisor fails the decimals=24 / 30 / trailing-zero-24 cases (dust and zero pass either way, since the fraction equals the raw) — the high-decimal tests are non-inert.
- **Blast radius:** no live asset reaches 23 decimals today (fee coins ≤18), so runtime display is provably unchanged for current assets — this removes latent corruption for the first high-decimal asset that lands. Changeset is a patch to `@vultisig/sdk` + `@vultisig/cli`, exports additive.

## verdict

APPROVE. Correct fund-safe fix, exact bigint math with no rounding/truncation, byte-for-byte legacy parity, clean by-reference exports on both root and the Metro-resolved RN entry, and non-inert tests with a proven red-on-revert.

---
🤖 gomes review loop · Claude lane load-bearing (Codex lane best-effort, unusable this pass)

PR #1256, opened 2026-07-13
fix(sdk): resolve typed EVM chain ids through canonical registry

### LINKED ISSUE #1254: decode: typed EVM chain ids still use a stale local map instead of the canonical SDK registry
## Summary
The new typed-EVM decode path reintroduced a local decimal chain-id table in `packages/sdk/src/tools/decode/evm.ts`, and it was already stale: supported chains like Hyperliquid and Sei were missing.

## Evidence
- `packages/sdk/src/tools/decode/evm.ts:5-27` — local `EVM_CHAIN_ID_TO_SYMBOL` table in the decoder.
- `packages/core/chain/chains/evm/chainInfo.ts:62-105` — canonical supported EVM registry already exists in core.
- The canonical registry includes Hyperliquid and Sei, but the local decoder table did not.

## Impact
Typed txs on supported chains could decode to raw numeric chain ids instead of canonical policy tags, breaking downstream chain matching and reintroducing the exact local-table drift this campaign is trying to eliminate.

## Fix sketch
Route typed tx chain ids through `getEvmChainByChainId(numberToHex(...))`, then normalize to the canonical SDK chain tag.

## Effort
Small.

Report mirror: https://gist.github.com/gomesalexandre/b1deffff1532dd07f2f5e7dae2b9f0a8

gomesalexandre:
## Summary
- route typed EVM decode through the canonical SDK chain-id registry instead of a fresh local map
- add regression coverage for Hyperliquid + Sei typed tx decoding and unknown-chain fallback
- add a changeset for the decode/public contract fix

## Test plan
- [x] `yarn install --immutable`
- [x] `yarn build:shared`
- [x] `yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/decode/fromToolResult.test.ts`
- [x] `yarn build:sdk`
- [x] `yarn build:client-shared`
- [x] `yarn cli:build`
- [x] built-dist runtime proof: `{"chain":"hyperliquid","recipient":"0x70997970C51812dc3A010C7d01b50e0d17dc79C8","amount":"9","decoded":true}`

## Context
- report mirror: https://gist.github.com/gomesalexandre/b1deffff1532dd07f2f5e7dae2b9f0a8
- sdk issue: https://github.com/vultisig/vultisig-sdk/issues/1254

Closes https://github.com/vultisig/vultisig-sdk/issues/1254

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decoding of typed EVM transactions on newer supported chains.
  * Ensured chain names are resolved consistently, with a numeric fallback for unrecognized chain IDs.

* **Tests**
  * Added coverage for recognized and unrecognized typed-transaction chain IDs.

* **Chores**
  * Scheduled patch releases for the SDK and CLI.


NeOMakinG (approved):
## testing approach

Reviewed adversarially against the fund-safety lens for a chain-id resolution change, plus a Codex `gpt-5.4` second-opinion lane. Worktree at HEAD `26dd941f`, read the actual canon (`getEvmChainByChainId`, `mirrorRecord`, `canonicalChainTag`, the `EvmChain` enum, `chainsMatch`) — not just the diff. Ran the decode suite (27/27 green) and proved red-on-revert on the two new tests.

**Verified this is a decode / policy-tag path, not a signing path.** The resolved `env.chain` tag only feeds `chainsMatch(claimChain, envChain)` in `checkInvariants` / `evaluatePolicy`. The signed EIP-155 chainId lives in the tx bytes and is untouched — no risk of routing a signed tx to the wrong network.

Enumerated every chain:
- **Hex-format consistency (no coercion mismatch):** the registry keys are built with `numberToHex(chain.id)` and the query is `getEvmChainByChainId(numberToHex(chainId))` — the *same* `numberToHex = 0x${n.toString(16)}` (lowercase, no zero-pad) on both sides, so the lookup is an exact string match. No hex/decimal, case, or padding drift.
- **All 11 pre-existing tags are byte-identical to the old map.** Confirmed viem ids: eth 1, optimism 10, cronos 25, bsc 56, polygon 137, zksync 324, mantle 5000, base 8453, arbitrum 42161, avalanche 43114, blast 81457. The `EvmChain` enum values collapse correctly via `canonicalChainTag` (incl. the `cronoschain → cronos` alias). Strict superset — adds `hyperliquid` (999) and `sei` (1329), zero regression on existing chains.
- **Unknown chain → `String(chainId)` numeric fallback**, same as before → fail-closed at the policy layer (a numeric tag never matches a symbolic claim). Proven by the `31337` test.
- **No false-allow:** the resolution is a truthful 1:1 read of the on-wire chainId, and the typed on-wire id is authoritative (overrides a mismatched caller hint — the safe direction).

Red-on-revert: patched the resolver back to the old 11-entry map → the new "resolves typed tx chain ids through the canonical SDK registry" test (hyperliquid/sei) goes red; the unknown-fallback test stays green. The new coverage is not inert.

## should-fix

- `packages/sdk/src/tools/decode/fromToolResult.ts` — the `EVM_CHAINS` family-inference set (`ethereum/base/arbitrum/polygon/optimism/bsc/avalanche`) is narrower than the decode vocabulary this path now emits (`cronos`, `zksync`, `mantle`, `blast`, and now `hyperliquid`, `sei`). A caller that omits `family` and passes e.g. `chain: 'sei'` will get `decoded: false` before the new resolver is reached. This is **pre-existing** (the old hardcoded map already emitted `cronos`/`zksync`/`mantle`/`blast`, none of which are in `EVM_CHAINS`), **fail-closed** (safe — policy can't match, so it blocks), and this PR's real consumers pass `family: 'evm'` explicitly (as every new test does), so the typed-decode fix works end-to-end. Worth a follow-up to widen `EVM_CHAINS` so the inference path matches the broadened tag vocabulary — not a blocker for this PR. (found by the Codex lane)

## risk

Low. Behavior-preserving for all 11 previously-mapped chains (byte-identical tags), a strict superset that additionally resolves Hyperliquid + Sei, fail-closed on unknown ids, and confined to the decode/policy-tag path — the signed chainId is never touched. Changeset present (`@vultisig/sdk` + `@vultisig/cli` patch).

## verdict

APPROVE. Clean, correct, strictly-improving fix; sole cross-lane finding is a pre-existing, fail-closed consistency gap suitable for a follow-up.

---
Reviewed by Claude (correctness/fund-safety) + Codex `gpt-5.4` (adversarial second opinion). 27/27 decode tests green, red-on-revert verified.

rcoderdev (approved):

PR #1258, opened 2026-07-14
fix(native-swap): centralize nullable chain-id lookup

### LINKED ISSUE #1257: architecture: native-swap chain-id lookup has already re-fragmented into local helpers
## Summary
Native-swap chain-id lookup has already started to re-fragment into local helpers instead of one canonical accessor.

## Evidence
- `packages/sdk/src/vault/services/nativeSwapBroadcastGuard.ts` had a local `getNativeSwapChainId` wrapper over `nativeSwapChainIds`.
- `packages/core/chain/swap/native/halts/getNativeSwapTradingHalt.ts` separately looked up chain ids from `nativeSwapChainIds`.
- `packages/core/chain/swap/native/minimum/getNativeSwapMinAmountIn.ts` separately looked up source/destination chain ids from `nativeSwapChainIds`.

## Impact
This is a fresh drift precursor inside the SDK itself: once a registry needs nullable or generic-chain access, callers start open-coding slightly different wrappers.

## Why this belongs in the SDK
`NativeSwapChain.ts` is already the canonical registry owner. It should export the chain-id accessor too, so downstream code consumes one contract.

## Fix sketch
- Export `getNativeSwapChainId(chain)` from `packages/core/chain/swap/native/NativeSwapChain.ts`.
- Replace local/ad hoc nullable lookups with the canonical accessor.
- Add a regression test for supported and unsupported chains.

## Effort
Low.

## Report
Round-25 report mirror: https://gist.github.com/gomesalexandre/ba70da9db0530f75fe2ae5a5b21b3bb9

gomesalexandre:
## Summary
- add a canonical `getNativeSwapChainId(chain)` accessor beside the native-swap registry
- switch the SDK broadcast guard and native-swap core helpers to consume the shared accessor instead of open-coding nullable map lookups
- pin supported + unsupported chain behavior with a focused regression test

Closes #1257

## Why
Round 25 found the SDK itself starting to reintroduce the campaign's recurring bug class: local nullable wrappers over `nativeSwapChainIds` were appearing in multiple places. This is a low-risk consolidation that keeps the registry contract owned in one place before those call sites drift further.

## Verification
- `yarn test:core packages/core/chain/swap/native/NativeSwapChain.test.ts`
- `yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/vault/services/nativeSwapBroadcastGuard.test.ts`
- `yarn build:sdk`
- `yarn eslint --config .config/eslint.config.mjs packages/core/chain/swap/native/NativeSwapChain.ts packages/core/chain/swap/native/NativeSwapChain.test.ts packages/core/chain/swap/native/halts/getNativeSwapTradingHalt.ts packages/core/chain/swap/native/minimum/getNativeSwapMinAmountIn.ts packages/sdk/src/vault/services/nativeSwapBroadcastGuard.ts`

## Risk
- LOW RISK
- pure accessor extraction / call-site consolidation
- no intended behavior change beyond routing nullable lookups through one canonical helper

## Context
- Round report mirror: https://gist.github.com/gomesalexandre/ba70da9db0530f75fe2ae5a5b21b3bb9

## Summary by CodeRabbit

- **New Features**
  - Added consistent native-swap chain ID resolution across supported chains.
  - Unsupported chains now return no native-swap chain ID.

- **Bug Fixes**
  - Improved handling of unsupported chains when calculating minimum swap amounts and validating broadcasts.
  - Preserved existing trading-halt and broadcast validation behavior.


NeOMakinG (approved):
## testing approach

Two-lane review (Claude + Codex gpt-5.4). Read the actual accessor and all four call sites at HEAD `fe46de3c`, not just the diff. Ran both cited suites in a worktree (symlinked node_modules): `NativeSwapChain.test.ts` 2/2 green, `nativeSwapBroadcastGuard.test.ts` 13/13 green. Enumerated every remaining `nativeSwapChainIds[` lookup in the repo to check the "centralize" claim.

Adversarial focus per fund-safety lens: (a) exact behavior at each migrated call site, (b) null fail-closed vs misroute, (c) any chain remapping to a different id, (d) null+resolved test coverage.

## blocking

None.

## preferably-blocking

None.

## should-fix

None.

## suggestion

None blocking. The registry still has direct `nativeSwapChainIds[...]` reads in `cosmos/index.ts:52`, `toNativeSwapAsset.ts:59/73/85`, and `limitSwapMemo.ts:79`. These are correctly left untouched here — each sits in a non-nullable narrowed context (post-`isOneOf` throw guard, or key-iteration over enabled chains), so they are not the nullable-wrapper class this PR targets. Flagging only so a future round doesn't mistake them for missed migrations.

## q

None.

## risk

LOW. Pure accessor extraction; the `nativeSwapChainIds` map itself is unchanged (diff only adds 3 lines to `NativeSwapChain.ts`), so there is zero id-remap / wrong-network hazard. Verified behavior-preserving at all four call sites:

- **`nativeSwapBroadcastGuard.ts:72`** — old local helper returned `string | undefined`, canonical returns `null`. The `undefined -> null` change is inert because the consumer uses truthiness (`if (!sourceChainId) throw UnsupportedChain`). Fail-closed on the signing/broadcast path is preserved.
- **`getNativeSwapTradingHalt.ts:34` (`chainIdFor`)** — accessor is only reached inside the `isOneOf(coin.chain, enabled) ? ... : null` ternary, so it always returns a real string; consumer at :67 filters `id !== null`. Identical.
- **`getNativeSwapMinAmountIn.ts:133/185`** — the added `if (!destChainId/fromChainId) return null` guards are effectively unreachable given the upstream `isOneOf(to.chain, enabled) && isOneOf(from.chain, enabled)` guard at :113; null here degrades to "no proactive minimum signal" (UX helper, not signing), which is safe.

Codex second-opinion lane converged to zero findings independently.

## verdict

APPROVE. Clean, behavior-preserving consolidation of the recurring nullable-wrapper bug class; fail-closed preserved on the broadcast path, no fund-misroute, null + resolved cases both pinned by the new regression test. Changeset present.

rcoderdev (approved):

PR #1259, opened 2026-07-14
fix(cli): persist direct broadcast resolutions

neavra:
## What

- add a real child-process CLI test proving the direct send journal persists across processes
- verify duplicate sends exit 9 with a clear refusal, while distinct intents and `--force` proceed
- record terminal `tx-status` results in the shared journal as `confirmed` or `failed`
- add a patch changeset for `@vultisig/cli`

## Why

Direct send and swap broadcasts use the shared persistent duplicate guard, but the direct CLI confirmation path did not record terminal resolutions. In particular, a definitively failed transaction remained indistinguishable from a pending one in the journal. The process-level test also protects the persistence behavior that prevents a fresh CLI invocation from rebroadcasting an identical intent.

A returned broadcast hash is still treated as pending. Resolution is written only after `tx-status` observes an on-chain success or error.

## How tested

- red/green: stashed the `tx-status` implementation with the new tests present; 2 resolution tests failed, then passed after restoring it
- focused send guard and status tests: 27 passed
- `yarn install`
- `yarn build:sdk`
- `yarn build:client-shared`
- `yarn build:rujira`
- `yarn cli:build`
- CLI typecheck
- changed-file ESLint
- changed-file Prettier with `--config .config/.prettierrc`
- `yarn test:cli`: 49 files, 665 tests passed
- built `send --help` and `swap --help` both expose `--force`

All transaction tests use mocked vaults and temporary journal files; no blockchain transactions were run.

## Summary by CodeRabbit

- **New Features**
  - Transaction confirmations and failures are now persisted in a shared broadcast journal.
  - Duplicate broadcast attempts are reliably blocked across separate CLI processes.
  - Forced retries remain available when needed.
  - Empty EVM calldata is now canonicalized so equivalent intents dedupe correctly.
- **Bug Fixes**
  - Non-terminal transaction statuses are no longer recorded as final resolutions in the journal.
  - On non-EVM chains, memo value `"0x"` is no longer treated as empty, preventing false deduplication.


gomesalexandre (changes requested):
nice coverage push overall, but there is one fund-safety regression in the fingerprint canonicalization.

- `clients/cli/src/agent/broadcastJournal.ts:212-224` now canonicalizes literal `"0x"` to empty for every `intent.data` before hashing. that is fine for evm calldata parity, but `intent.data` is also where non-evm sends carry memos. `buildSendBroadcastIntent()` fingerprints `keysignPayload.memo` into `data`, and the agent path fingerprints `txArgs.memo` into the same field, so a real memo of `"0x"` now collides with no memo at all.
- practical effect: on memo chains, a send with memo `"0x"` can be falsely treated as a duplicate of the same send with no memo, or vice versa, inside the 10 minute guard window.
- receipt: with the current algorithm, these two non-evm intents hash to the same fingerprint:
  - owner=`vault123`, chain=`THORChain`, to=`thor1abc`, value=`100000000`, asset=`RUNE`, memo=`""`
  - owner=`vault123`, chain=`THORChain`, to=`thor1abc`, value=`100000000`, asset=`RUNE`, memo=`"0x"`
  both produce `8aa2e111eb077a6d61c70e4b3c7869d4`.

please scope the `"0x" -> ""` normalization to evm calldata only, then add a non-evm regression test proving memo `"0x"` stays distinct from an empty memo.

receipts from this pass:
- codex review also converged on the same finding before transport auth died
- focused vitest rerun after borrowing the main clone install: `broadcastJournal.test.ts` and `txStatus.test.ts` passed; `sendSwapDupGuard.test.ts` hit an unrelated local export-resolution issue while loading `@vultisig/core-chain` from the borrowed install, so i did not use that loader failure as evidence for the finding

rcoderdev (approved):

gomesalexandre (approved):
Looks good now.

I re-checked the remote diff at the current head. The `"0x"` vs empty-memo fingerprint collision is fixed by making empty-calldata canonicalization explicit and EVM-only, and the new regression tests pin both the non-EVM memo distinctness case and the EVM calldata dedupe case.

The persisted resolution behavior also stays honest: only terminal outcomes get recorded, stale failures cannot retroactively unblock a newer identical broadcast, and a missed resolution write fails closed.

PR #1261, opened 2026-07-14
fix(ripple): treat queued transaction as in flight

### LINKED ISSUE #1122: ripple: terQUEUED treated as hard broadcast failure — user told 'failed' while tx lands
## Problem

XRPL `submit` returning `terQUEUED` means the tx was **accepted into the ledger's transaction queue** and is expected to apply in a following ledger — but the broadcast resolver's peer-race whitelist is only `{tefALREADY, tefPAST_SEQ}` (broadcast/resolvers/ripple.ts:29), so `terQUEUED` (non-zero engine_result_code) is thrown as a hard failure (ripple.ts:47-59).

Result: user is told the send **failed** while the tx very likely lands — "failed but money moved", the most trust-destroying outcome for a wallet.

## Fix

Treat `terQUEUED` as in-flight: return the tx hash (optionally after a short verify-by-hash/status poll). Note the resolver's own doc comment lumps `ter*` under "authoritative no", which is wrong for `terQUEUED` specifically — the XRPL docs classify `ter*` as retry-later, and QUEUED in particular as accepted-pending.

From the per-chain reliability parity audit 2026-07-09 (adversarially verified). Related epic: #979.

rcoderdev:
Closes #1122
XRPL terQUEUED is now treated as an accepted in-flight transaction without a redundant hash verification; other retry-later and rejection results remain visible failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * XRPL transactions reported as `terQUEUED` are now treated as in-flight instead of failed.
  * Queued transactions no longer trigger unnecessary broadcast verification.
  * Other XRPL rejection outcomes continue to be handled as before.


gomesalexandre (approved):
Nice fix.

I checked the XRPL submit path directly in `packages/core/chain/tx/broadcast/resolvers/ripple.ts`, and the split now matches the protocol semantics:
- `engine_result_code === 0` still returns on the normal success path
- peer-race cases (`tefALREADY`, `tefPAST_SEQ`) still go through `verifyBroadcastByHash`
- `terQUEUED` is now handled as the one accepted-but-still-in-flight retry-later result instead of being surfaced as a hard rejection
- the remaining `ter*` / `tef*` / `tec*` / `tem*` / `tel*` outcomes still fail closed

The targeted tests in `packages/core/chain/tx/broadcast/resolvers/ripple.test.ts` pin the exact behavior change and guard against broadening the in-flight bucket past `terQUEUED`.

PR #1262, opened 2026-07-14
fix(cli): preserve swap amount precision

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved exact decimal swap amounts from CLI input through quotes, previews, dry runs, and execution.
  * Improved amount normalization for precise decimal and supported scientific-notation inputs.
  * Rejected invalid swap amounts before processing begins with a clear error.
  * Prevented duplicate swaps when amounts are equivalent at the source token’s precision.


gomesalexandre (approved):
looks good.

receipts:
- `yarn vitest run clients/cli/src/commands/swapAmount.test.ts clients/cli/src/commands/__tests__/sendSwapDupGuard.test.ts -t 'CLI swap amount precision|refuses amounts equivalent at the source token precision'` ✅ 18 passed
- `yarn cli:build` ✅
- `yarn lint` ✅
- `VULTISIG_CONFIG_DIR=/private/tmp/sdk-pr1262-review.cyRdiv/cli-config node clients/cli/dist/index.js swap Ethereum Bitcoin 0.123456789123456789 --dry-run --output json` reached the compiled CLI surface and got the expected `NO_ACTIVE_VAULT` gate instead of failing amount parsing.

nice fix preserving the exact user-entered decimal string through quote and execution while still canonicalizing at source-token precision for preview and duplicate-broadcast protection.

PR #1263, opened 2026-07-14
docs(cli): retire stale get_portfolio action

rcoderdev:
## Summary by CodeRabbit

* **Documentation**
  * Updated the documented agent actions to remove the retired portfolio lookup action.
  * Added the balances lookup action for retrieving balances across all chains.

* **Tests**
  * Added coverage to ensure retired actions are excluded from the documented action list.


gomesalexandre (approved):
Looks good.

I checked the only three touched surfaces and the contract is now internally consistent:
- `clients/cli/src/agent/agentActionContract.ts` removes the retired `get_portfolio` action from `DOCUMENTED_AGENT_ACTION_TYPES`
- `AGENTS.md` removes the same stale row from the curated action table
- `clients/cli/src/agent/__tests__/agentActionContract.test.ts` now explicitly guards against reviving `get_portfolio`

Grounding: the backend/tooling surface today exposes `get_balances` for portfolio/balance reads, not a live `get_portfolio` action, so retiring the stale docs/contracts entry is the right fix.

PR #1264, opened 2026-07-14
fix: fail closed on missing Ripple ledger index

### LINKED ISSUE #1123: ripple: (ledger_current_index ?? 0) + 60 signs an instantly-expired tx on missing field — fail closed instead
## Problem

`packages/core/mpc/keysign/chainSpecific/resolvers/ripple.ts:59`:

```ts
lastLedgerSequence: BigInt((ledger_current_index ?? 0) + 60),
```

If `account_info` ever omits `ledger_current_index`, the tx signs with `LastLedgerSequence = 60` — a ledger index from 2013 — making it instantly non-includable (`tefMAX_LEDGER`) with no build-time error and no indication of the real cause. A silent `?? 0` on a field that controls tx mortality is a fail-open where it should fail-closed.

## Fix

Throw at build time when `ledger_current_index` is absent (fail closed, one line), instead of signing a dead tx.

From the per-chain reliability parity audit 2026-07-09. Related epic: #979.

rcoderdev:
Closes #1123

### DIFF
--- a/.changeset/ripple-ledger-index-fail-closed.md
+++ b/.changeset/ripple-ledger-index-fail-closed.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-mpc': patch
+---
+
+Fail closed when Ripple account info is missing a valid `ledger_current_index` instead of building an already-expired transaction.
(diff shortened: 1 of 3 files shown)


gomesalexandre (approved):
Remote pass on current head b60d3288a44d.

Looks good.
- fail-closed guard is in the right place before signing
- valid-path behavior stays the same: ledger_current_index 100 still yields lastLedgerSequence 160n
- regression test now truly omits ledger_current_index and asserts the throw
- unit, integration, lint/typecheck, and test-summary are green

No blocking findings.

PR #1265, opened 2026-07-14
fix(core-chain): retry transient broadcast failures

### LINKED ISSUE #1120: reliability: no broadcast retry for transient failures on 10/12 chains — add shared bounded retry at dispatcher
## Problem — zero broadcast retry on transient failures outside Solana/EVM

10 of 12 broadcast resolver families make exactly ONE attempt to submit a fully-signed tx. A transient failure (connection reset, proxy 5xx, momentary RPC blip) fails the entire MPC keysign flow, even though re-broadcasting the same signed bytes is idempotent on these chains (dup-tx codes + `verifyBroadcastByHash` backstop).

Verified at every layer: no retry in `queryUrl` (packages/lib/utils/query/queryUrl.ts), none in the broadcast dispatcher (chain/tx/broadcast/index.ts), none in the resolvers, none in BroadcastService. The two exceptions:
- **solana** — bespoke 3-attempt bounded retry for its transient class (broadcast/resolvers/solana.ts:11-13,47-62)
- **evm** — free via viem http transport defaults (retryCount=3, retryDelay=150ms)

A signed-but-unbroadcast tx is the most expensive artifact in the system (full MPC ceremony behind it); dropping it on a single network blip is the worst ROI failure we have. Cardano is the sharpest case: `verifyBroadcastByHash` re-throws when the tx never reached the network, so the user re-runs the entire signing flow.

## Fix direction

One shared bounded-retry wrapper at the dispatcher layer (2–3 attempts, short backoff, retry only on transport-class errors — never on node rejections), instead of 10 per-resolver patches. The chains' existing duplicate-broadcast handling (`already known` codes, ogmios 3117, tefALREADY/tefPAST_SEQ, hash-verify) makes re-send safe.

Severity: S2/S3 depending on chain. From the per-chain reliability parity audit 2026-07-09 (all claims adversarially verified against source; per-chain file:line in the matrix JSON).

rcoderdev:
Closes #1120


gomesalexandre (approved):
Remote pass on current head ed36bc73b7aa.

Looks good.
- bounded dispatcher retry is only wrapping the non-EVM/non-Solana resolvers
- duplicate-broadcast handling / verifyBroadcastByHash backstop make the retry path idempotent across the covered chains
- tests cover retry success, bounded exhaustion, non-retry of node rejections, classifier behavior, and the EVM/Solana carve-out

Non-blocking nits:
- add a small cycle/depth guard while walking error.cause
- Retry-After-aware backoff could make 429 handling more honest later

Approve on code correctness. Still wait for the remaining CI to go green before merge.

PR #1266, opened 2026-07-14
fix(sdk): honor zero-fee Jupiter quote floor

gomesalexandre:
## Summary
- gate the direct SDK Jupiter tool's affiliate fee state on the quoted `platformFee.amount`, not just ATA presence
- omit `feeAccount` from `/swap` when Jupiter floors the platform fee to zero
- add a regression test for the zero-fee floor path

## Why
The canonical shared Jupiter path already treats `platformFee.amount = 0` as fee-OFF. The direct SDK tool had drifted: if a treasury ATA resolved, it still marked `affiliateFeeApplied=true` and sent `feeAccount` to `/swap` even when the quote charged no fee.

## Verification
- `yarn workspace @vultisig/sdk vitest run tests/unit/jupiterSwap.test.ts`
  - PASS (20 tests)
- Constraint: disk pressure on the host dropped below ~120 MiB free during this round, so broader `build:sdk` / `cli:build` receipts in the sparse fresh clone hit ENOSPC before they could run.

## Notes
- This addresses the shipped finding for the direct SDK Jupiter tool's zero-fee floor drift.
- Follow-up architecture issue remains open for the separate static-ATA-vs-dynamic-fee-account split.


NeOMakinG (approved):
## testing approach
Reviewed in isolated worktree at HEAD `fb503c33`. Two lanes: Claude (primary correctness + fund-safety) + Codex gpt-5.4 (adversarial second opinion). Read the actual fee-floor logic in `jupiter.ts` plus the canonical `core/.../getJupiterSwapQuote.ts` it claims to mirror. Ran `vitest run tests/unit/jupiterSwap.test.ts` (20/20 green) and proved the new test is **red-on-revert** (reverting `chargesFee` to `feeAccount !== null` fails exactly the new floor test, 1 failed / 19 passed).

## blocking
None.

## preferably-blocking
None.

## should-fix
- `packages/sdk/src/tools/swap/jupiter.ts:293` — `BigInt(quote.platformFee?.amount ?? '0')` hardens only the honest `"0"`/absent cases. Malformed-but-present proxy shapes coerce inconsistently: `''` and `'  '` → `0n` and `'-1'` → negative both silently make `chargesFee` false (affiliate-revenue drop, reported as fee-OFF), while `'abc'`/`'1.2'` throw a raw `SyntaxError` that aborts the swap build. This is **inherited by design** from the canonical `getJupiterSwapQuote.ts:132` (which the PR explicitly matches), so hardening it *only here* would re-introduce the very drift this PR removes — hence non-blocking. Suggested follow-up: a shared `parseJupiterFeeAmount()` helper used by BOTH paths that treats non-numeric/negative as fee-OFF without throwing. No user fund-safety exposure: the swap is unaffected (user pays no fee), and the returned tx is still independently validated by `assertSafeSolanaSwapTransactionBase64`.

## q
None blocking.

## risk
Low. Scope is 3 changed lines + a changeset + a regression test, confined to the direct SDK Jupiter tool. Fix is behaviour-preserving for the fee-ON case and only changes the previously-drifted fee-floored case to match the proven canonical core path.

**Fund-safety walk:** (a) a legitimate zero-fee quote is now accepted (swap built, `feeAccount` omitted, `affiliateFeeApplied=false`) rather than rejected/overwritten. (b) The classic `0`-vs-`undefined` trap is avoided — nullish `?? '0'` + explicit `swapFeeAmount > 0n`, not a falsy `!fee`/`fee || default`; a real zero floor is not clobbered. (c) No reverse underpay: a genuine `amount > 0` keeps `chargesFee=true` and sends `feeAccount`. (d) No WYSIWYS drift: `affiliateFeeApplied === chargesFee === (feeAccount included)`.

## verdict
APPROVE. The zero-fee floor is honored correctly, the falsy-0 trap is handled, no user fund-safety exposure, and the regression test is proven non-inert. The single Codex finding is a cross-path revenue-robustness hardening left as a documented follow-up (fixing it unilaterally would create the drift this PR eliminates).

---
_Reviewed by Claude (Opus 4.8) primary lane + Codex gpt-5.4 adversarial lane. Isolated worktree, tests green + red-on-revert verified._

rcoderdev (approved):

PR #1269, opened 2026-07-14
fix(solana): classify expired unknown transactions

### LINKED ISSUE #1121: reliability: no expired/dropped terminal state — dead txs poll 'pending' forever on 8 chains (blocks agent-backend#1535 reaper)
## Problem — no expired/dropped terminal state: dead txs poll "pending" forever

Most supported chains give a tx a deterministic death: Solana blockhash `lastValidBlockHeight`, TON `expireAt` (now+600s), Tron `expiration` (default now+1h), Ripple `LastLedgerSequence` (signed at current+60), Cardano TTL slot, Polkadot/Bittensor mortal era (~64 blocks). The status resolvers never consult these, so a tx that expired without landing reports `pending` (or `pending/isKnown:false`) **indefinitely**:

- solana: never compares current block height vs lastValidBlockHeight (status/resolvers/solana.ts)
- ton: no expiry-aware terminal state despite expireAt in the signed message (ton.ts:29-31)
- tron: bare pending forever, expiration ignored (tron.ts:48-50)
- ripple: after LastLedgerSequence passes, txnNotFound keeps mapping to pending/isKnown:false (ripple.ts:18-26)
- cardano: TTL-expired tx indistinguishable from not-yet-indexed (status via utxo.ts:36) — even though broadcast-time code already computes TTL freshness deterministically (broadcast/resolvers/cardano.ts:18-32)
- polkadot/bittensor: mortal era lapse undetectable; polkadot status additionally rides keyless Subscan

Downstream effect: consumers polling `vault.getTxStatus` only exit via their own timeout with a generic message, and the server-side ledger accumulates permanently-"submitted" rows — this issue is the SDK prerequisite for the stuck-`submitted` reaper (vultisig/agent-backend#1535), which needs a terminal signal to resolve rows with a tx_hash.

## Fix direction

Per-chain expiry check inside the status resolver: when the hash is unknown AND the tx's own deadline has passed (needs the signed tx's expiry threaded into the status query, or a `checkExpiry` variant on the resolver input), return `not_found`/`error('expired')`. Chains where expiry is cheap to evaluate first: cardano (TTL already parsed at broadcast), ripple (LastLedgerSequence vs validated ledger), solana (getBlockHeight vs lastValidBlockHeight).

Severity: S3. From the per-chain reliability parity audit 2026-07-09; file:line evidence per chain in the matrix JSON.

rcoderdev:
Closes #1121

## Summary by CodeRabbit

* **New Features**
  * Transaction status requests now support an optional **last-valid block height** across the SDK.
  * Solana transaction checks can now classify expired, unknown signatures as **Not Found**, while remaining **Pending** when a transaction may still appear.
* **Bug Fixes**
  * Improved handling of unknown Solana signatures: avoids unnecessary lookups and correctly resolves status when validity has expired.
  * Updated the SDK/Vault status flow to propagate the optional last-valid block height consistently.
* **Tests**
  * Expanded Solana status test coverage for missing/unknown signatures and block-height expiry scenarios.


gomesalexandre (approved):
looks good.

- `packages/core/chain/tx/status/resolvers/solana.ts` now only upgrades an unknown Solana hash to `not_found` when `lastValidBlockHeight` is present and the finalized chain height has actually moved past it. the failure paths still fail closed to `pending` on status-rpc and block-height lookup errors.
- `packages/sdk/src/vault/VaultBase.ts` threads the new field through as an optional param, so the sdk api change stays backward-compatible.
- focused receipt in a disposable clone after `yarn build:shared`: `yarn vitest run packages/core/chain/tx/status/resolvers/solana.test.ts` ✅ (7 tests passed).

PR #1272, opened 2026-07-14
fix(cardano): fall back when tip fetch fails

### LINKED ISSUE #1124: cardano: transient tip-fetch failure discards a fully-signed tx (TTL guard is a broadcast SPOF)
## Problem

Every Cardano broadcast hard-depends on a side-call succeeding: `assertCardanoTtlFreshForBroadcast` awaits `getCardanoCurrentSlot()` (rootApiUrl/cardano/tip, single attempt, no retry) before submit (broadcast/resolvers/cardano.ts:18-33). A transient failure of the tip route — in the window between keysign and broadcast — throws and **discards a fully-signed MPC transaction** that the node itself would have accepted or authoritatively rejected on its own TTL check.

The freshness gate is good (prevents submitting an about-to-expire tx); the SPOF is that a *transient infrastructure blip on the guard* is indistinguishable from a *real TTL failure*.

## Fix

On `getCardanoCurrentSlot()` failure (as opposed to a genuine stale-TTL result), fall through to submit and let the node adjudicate TTL — or retry the tip fetch once before giving up. Keep throwing only when the slot is actually fetched and the TTL is genuinely stale.

From the per-chain reliability parity audit 2026-07-09 (adversarially verified, S2).

rcoderdev:
Closes #1124

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cardano transaction broadcasting by retrying current network-slot checks when tip data is temporarily unavailable.
  * Transactions are no longer blocked solely because the network tip cannot be fetched, while genuinely stale transactions continue to be rejected.
  * Added coverage for retry behavior and transaction freshness validation.


gomesalexandre (approved):
looked good.

- retrying the Cardano tip lookup once and then treating a double failure as unknown removes the Koios `/tip` SPOF without weakening the stale-TTL guard.
- the boundary math stays correct with the existing 60-slot safety margin: slot 939 stays fresh for ttl 1000, slot 940 still fails closed.
- i also sanity-checked the failure mode: if tip lookup is still down, the tx just falls through to submit and the node adjudicates TTL. no false-success surface gets introduced.

nice small fix and the new tests cover the important paths.

PR #1273, opened 2026-07-14
fix(core-chain): preserve Cosmos hash after indexing timeout

### LINKED ISSUE #1125: cosmos: mempool-accepted tx reported as broadcast FAILURE when indexing outlasts the 60s poll
## Problem

Cosmos "submitted-but-reported-failed": `broadcastTx` uses cosmjs's sync-broadcast + inclusion polling; when `broadcastTxSync` succeeds (tx accepted into the mempool) but `getTx` doesn't index it within the 60s poll window, cosmjs throws `TimeoutError`. The resolver then runs `verifyBroadcastByHash`, which queries the **same slow node** immediately, gets null → the original timeout is re-thrown, and the user is told the broadcast **failed** while the tx is in the mempool and typically lands.

Same class as the Android "device didn't respond" mislabel: outcome-unknown reported as outcome-failure.

## Fix direction

When `broadcastTxSync` itself succeeded (hash exists), a later indexing timeout must surface as "pending/unconfirmed — check explorer", not as a broadcast failure; ideally return the hash and let status polling own the rest.

From the per-chain reliability parity audit 2026-07-09 (adversarially verified, S3). File:line evidence in the matrix JSON (cosmos/idempotency).

rcoderdev:
Closes #1125
CosmJS emits this timeout only after sync broadcast accepts the transaction. Runtime proof: the timeout helper returns the submitted ID, and the focused resolver regression covers timeout, duplicate, and fallback paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cosmos transactions that time out during broadcast now return their transaction hash when available.
  * Transaction status polling can continue to determine final confirmation instead of treating indexing delays as broadcast failures.
  * Existing handling for duplicate transactions and other broadcast errors remains unchanged.

* **Tests**
  * Added coverage for timeout, duplicate transaction, verification, and invalid transaction ID scenarios.

### DIFF
--- a/.changeset/cosmos-broadcast-timeout-pending.md
+++ b/.changeset/cosmos-broadcast-timeout-pending.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Return the Cosmos transaction hash when CosmJS accepts a broadcast but times out waiting for indexing, leaving confirmation to status polling instead of reporting broadcast failure.
--- a/.changeset/cosmos-broadcast-timeout-pending.md
+++ b/.changeset/cosmos-broadcast-timeout-pending.md
@@ -1,5 +1,6 @@
 ---
 '@vultisig/core-chain': patch
+'@vultisig/sdk': patch
 ---
 
 Return the Cosmos transaction hash when CosmJS accepts a broadcast but times out waiting for indexing, leaving confirmation to status polling instead of reporting broadcast failure.
--- a/packages/core/chain/tx/broadcast/resolvers/cosmos.test.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/cosmos.test.ts
@@ -1,0 +1,79 @@
+import { TimeoutError } from '@cosmjs/stargate'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const mocks = vi.hoisted(() => ({
+  getCosmosClient: vi.fn(),
+  verifyBroadcastByHash: vi.fn(),
+}))
+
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
looks right.

- `@cosmjs/stargate` only throws `TimeoutError` after the tx was submitted and indexing polling timed out, so hoisting its `txId` is the right fail-closed behavior here.
- non-timeout errors still route through `verifyBroadcastByHash`, and the existing duplicate-cache idempotence path stays intact.
- good call trimming blank ids before trusting them.

this matches the desired contract: accept the broadcast hash when CosmJS already sent the tx, then let later status polling confirm indexing.

gomesalexandre (approved):
Approving AFTER the runtime receipt (correct order - wasn't pre-approved). Runtime-verified on the built dist (packages/core/chain/dist/.../cosmos.js): a CosmJS TimeoutError w/ valid txId short-circuits + returns the txId (skips the false-failure re-query); empty/whitespace txId -> undefined (no bogus hash); a non-TimeoutError AND a BroadcastTxError (CheckTx reject) both fall through to verifyBroadcastByHash (reject path NOT swallowed); and hash-consistency confirmed byte-for-byte (CometBFT sha256(raw tx) == getCosmosTxHash sha256(decodedTxBytes)) so no false-explorer-hash. PR's own 4 tests + full test:core (1804) green. Honest caveat in the receipt: no live CometBFT node for a server-returned-hash round-trip, so equivalence rests on spec-match + sample agreement (acceptable - the fix only changes throw-vs-resolve, and BroadcastService independently computes the hash anyway). Merging on green.

PR #1274, opened 2026-07-14
fix(core): guard TON amount byte encoding

### LINKED ISSUE #1146: mpc: TON signed-mode negative Long can still produce truncated hex via numberToEvenHex
Follow-up from #1139/#1144 (negative-bigint-to-hex encoding guards) and #1138/#1140 (Long.fromString bounding), surfaced during #1140's final review.

TON's amount resolver stays in signed mode (correct per its proto type — genuinely bytes, not a Long field per the #1140 investigation), which admits negative Long values. When such a negative Long flows into `numberToEvenHex`, the resulting `-`-prefixed hex string gets silently truncated by `Buffer.from(hex, 'hex')` (same underlying mechanism #1144 fixed for direct bigint/number inputs — but a Long value's path into that function wasn't covered by that fix's scope).

Fix direction (per the reviewing agent): `{ unsigned: true }` on the TON path would close this for free, if TON's actual value range supports it — verify against real TON semantics before assuming, don't copy-paste the Sui/Cardano fix blind.

Severity: LOW (pre-existing, narrow — requires a negative amount reaching this specific path). Effort: S, once TON's actual constraints are confirmed.

rcoderdev:
Closes #1146
Pins the native TON resolver's 1 TON byte encoding to a public testnet transaction and rejects negative amounts before byte serialization across native, dApp, and Jetton paths. Vector: https://testnet.tonviewer.com/transaction/ee3d0d792404c489ee46fa335a951f7f9158a1758e6a24c3acfbc14b04441133

## Summary by CodeRabbit

* **Bug Fixes**
  * TON transfers now reject negative amounts before encoding.
  * Added validation for invalid, oversized, and out-of-range TON amounts.
  * Improved amount handling for native TON, Jetton transfers, and dApp-supplied messages.
  * Preserved correct encoding for zero, standard, and maximum supported amounts.

* **Tests**
  * Added coverage for valid TON amount encoding and rejected invalid values.

### DIFF
--- a/.changeset/ton-negative-amount-guard.md
+++ b/.changeset/ton-negative-amount-guard.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-mpc': patch
+---
+
+Reject negative TON signing amounts before byte encoding so native, dApp-supplied
+TON messages, and Jetton helper amounts cannot silently truncate a `-`-prefixed
+hex value.
--- a/.changeset/ton-negative-amount-guard.md
+++ b/.changeset/ton-negative-amount-guard.md
@@ -1,5 +1,6 @@
 ---
 '@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
 ---
 
 Reject negative TON signing amounts before byte encoding so native, dApp-supplied
--- a/packages/core/mpc/keysign/signingInputs/resolvers/ton/jetton.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/ton/jetton.ts
@@ -1,14 +1,12 @@
-import { Buffer } from 'buffer'
 import { tonConfig } from '@vultisig/core-chain/chains/ton/config'
 import { KeysignPayload } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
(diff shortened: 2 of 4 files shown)


gomesalexandre (approved):
lgtm 🫡 Taking this over from draft - it's a real fund-safety fix and it's complete. Guards the exact 'signer signs a different amount than displayed' class: pre-PR a negative/malformed TON toAmount went Long.fromString -> negative Long -> .toString(16) '-'-prefixed hex -> Buffer.from(hex,'hex') SILENTLY TRUNCATES to 0 bytes/wrong value. New tonAmountToBytes rejects non-/^\d+$/ before parsing, bounds against 2^120-1 (correct TON Grams VarUInteger16 spec max, verified against TL-B not copy-pasted from Sui/Cardano), removes Long from the amount path entirely, with a string-length DoS pre-check backed by an authoritative bigint comparison. Golden vector 1 TON=0x3b9aca00 (mathematically verified), boundary tests (accepts 2^120-1, rejects 2^120, rejects negative at all 3 call sites incl jetton). Complete - all TON amount-encode sites covered, no stubs, changeset scoped, CI green. Non-blocking: /^\d+$/ rejects a '+'-prefixed string old Long accepted, but toAmount is always internally bigint.toString() so never signed. Merging.

PR #1275, opened 2026-07-14
fix(sui): bound coin selection

### LINKED ISSUE #1132: fix(sui): bounded coin selection + covering gas object — match iOS #4734 (keysign parity; covers Windows)

rcoderdev:
Closes #1132

## Summary by CodeRabbit

- **New Features**
  - Added deterministic Sui coin-selection for both native and token transfers, including bounded input sizing for amount plus gas.
  - Improved gas-object selection to consistently pick the smallest sufficient native gas coin.
  - Refined chain-specific input generation now converges safely to stable coin selections as gas is re-priced.
- **Bug Fixes**
  - Corrected coin pagination handling and ensured only the intended “largest covering” coins are selected.
  - Improved behavior for dusty payloads and cases where gas coverage is unavailable.
- **Tests**
  - Expanded Sui coin-selection and signing-input resolver test coverage for native and token scenarios.


gomesalexandre (approved):
lgtm 🫡 Taking over from draft - fund-safe + complete. Fixes two real broadcast-failure bugs: native sends used ALL owned coin objects unbounded (dust-heavy wallet blows Sui's max-input-object limit -> on-chain fail), and token-send gas picked gasCoins[0] arbitrary RPC-order (could be a dust object too small for gas). New suiCoinSelection.ts: selectSuiInputCoins sorts largest-first + greedily covers target + THROWS on insufficient (fails closed, never silently truncates, incl at the 255-object cap); selectSuiGasObject picks the smallest native coin that covers gasBudget (deterministic). Verified no under-selection (both fns throw), no wrong-coin (isSameSuiCoinType filter with zero-pad normalization), gas top-5 bound monotonically safe (largest-covers implies top-5-covers, gas is singular). Two-stage bound is idempotent (stage 2 re-runs the same algo over stage 1's covering output). 14 test cases incl 800-coin dusty-wallet + 255/256 boundary, isolated to Sui, changeset scoped, CI green. Non-blocking follow-ups: 255 magic-number wants a source comment; the 255-cap-hit case could use a distinct 'consolidate' error msg (still fails closed). Merging.

gomesalexandre (approved):
Re-approving on the current head (fc487a1) after RUNTIME verification - not just CI-green. Two runtime receipts now on the PR against the real built dist (not source/mocks): Part A (coin selection - exact-cover/dust/255-cap-THROWS-fail-closed/insufficient-THROWS/gas-smallest-covering all confirmed) + Part B (the fc487a1 gas-budget convergence loop - bounded at exactly 3 dry-runs even with a runaway cost curve, monotonic, withFallback fails safe to static-budget). rcoderdev's own fc487a1 follow-up (re-price refined budget when re-selection grows the object count) + his changeset-guard fix are both in. One non-blocking finding surfaced + will file: the 2-round cap can leave the final selection ~5.9% under-priced in an adversarial synthetic fixture - but it's FAIL-CLOSED (Sui rejects under-budgeted txs pre-commit = failed send needing retry, never misexecution), it's a synthetic worst-case (real per-object marginal cost is much smaller + the 1.15x headroom usually absorbs it), and rcoderdev's test comment already documents it as the bound's tradeoff. Filing a suggestion-level follow-up (final re-price or clamp). Merging.

PR #1285, opened 2026-07-15
fix(sdk): price known non-EVM tokens

### LINKED ISSUE #1167: valuation: FiatValueService still hard-stops token pricing at EVM while fiatToAmount now resolves non-EVM tokens
## Summary
`fiatToAmount()` now resolves non-EVM token price ids, but `FiatValueService` still returns early for non-EVM token pricing. The SDK now exposes two valuation paths with different token support.

## Evidence
- non-EVM token resolution in `packages/sdk/src/utils/fiatToAmount.ts:86-113`
- EVM-only early return in `packages/sdk/src/services/FiatValueService.ts:467-485`
- public value flows still route through `FiatValueService` in `packages/sdk/src/services/FiatValueService.ts:254-318`

## Impact
- one SDK path can value non-EVM token ids/denoms while another silently cannot
- downstream consumers get pushed toward local branching instead of one canonical SDK valuation path

## Fix sketch
Make `FiatValueService` consume the same token-id resolution path that `fiatToAmount()` already uses, then add tests proving a known non-EVM token works through both surfaces.

## Effort
Medium

## Report
https://gist.github.com/gomesalexandre/fac3c1408ccefc7e90a16955b11e855b

rcoderdev:
Closes #1167

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fiat pricing for supported non-EVM tokens so prices are fetched from the canonical provider when available (instead of returning zero).
  * Updated token identifier resolution to use trimmed inputs with chain-specific matching rules (EVM case-insensitive, non-EVM exact), including improved Cosmos denom handling and whitespace tolerance.
  * Kept EVM ERC-20 pricing routing behavior consistent while expanding coverage for additional EVM networks.
* **Tests**
  * Expanded unit tests for non-EVM pricing resolution, EVM routing args, and stricter rejection of unsupported/case-variant token identifiers.
* **Release Management**
  * Added a patch release changeset for both `@vultisig/core-chain` and `@vultisig/sdk`.

### DIFF
--- a/.changeset/non-evm-fiat-token-pricing.md
+++ b/.changeset/non-evm-fiat-token-pricing.md
@@ -1,5 +1,6 @@
 ---
-"@vultisig/sdk": patch
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
 ---
 
 Fix fiat pricing for known non-EVM token identifiers.
--- a/.changeset/non-evm-fiat-token-pricing.md
+++ b/.changeset/non-evm-fiat-token-pricing.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Fix fiat pricing for known non-EVM token identifiers.
--- a/packages/core/chain/coin/price/__tests__/resolveTokenPriceId.test.ts
+++ b/packages/core/chain/coin/price/__tests__/resolveTokenPriceId.test.ts
@@ -99,9 +99,8 @@ describe('resolveTokenPriceId', () => {
       expect(resolveTokenPriceId(Chain.TerraClassic, 'krtc')).toBeUndefined()
     })
 
-    it('uluna on TerraClassic -> undefined (only in chainFeeCoin, not knownTokensIndex)', () => {
-      // uluna is the native denom but it is not registered as a separate knownToken entry
-      expect(resolveTokenPriceId(Chain.TerraClassic, 'uluna')).toBeUndefined()
+    it('unknown native-looking denom on TerraClassic -> undefined', () => {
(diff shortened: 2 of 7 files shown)


gomesalexandre (dismissed):
looks good. this fixes the bad evm allowlist routing and the non-evm fallback is wired the right way through the registry + cosmos fee denom lookup.

receipts i checked this pass:
- manual source verification against `EvmChain`, `resolveTokenPriceId`, `cosmosFeeCoinDenom`, `getCoinPrices`, and `getErc20Prices`
- integration tests green and lint/typecheck green on the head sha

unit was still running when i reviewed, so merge should wait for that last green check.

gomesalexandre (approved):
Re-approving AFTER the runtime receipt on the current head (0841dbb2, incl rcoderdev's 'preserve non-EVM token identity' follow-up). Runtime-verified live on the built dist: non-EVM tokens now price (Solana USDC $0.999898, TerraClassic uluna $0.000059 - were $0 pre-fix); the cross-token-mixup guard holds (a lowercase variant of the USDC mint resolves to undefined, no case-fold collision into the canonical mint's price); fail-closed (unknown mint -> undefined -> honest $0 -> fiatToAmount throws typed FiatToAmountError, not a $0-division blowup); the fund-adjacent fiatToAmount is sane ($50 SOL -> 0.66 SOL at live price); EVM path unchanged (Ethereum USDC still correct); 2827 tests green. Bonus isEvmChain typo fix verified live (BSC/Zksync now price, were silently $0). Hyperliquid $0 is an upstream CoinGecko coverage gap (proven via direct proxy curl), not a regression. BNBChain-typo sweep for the 2 other files filed as #1339. Merging on green.

Issue #1229, opened 2026-07-13
feat: derive cosmos chain-id aliases in normalizechain from the canonical chain-id table (terra-only today)


gomesalexandre:
`packages/sdk/src/utils/normalizeChain.ts` bills itself as "the SINGLE canonical LLM-tolerant chain normalizer" (:130), and its alias table is derived from tickers + enum names - but chain-id tolerance is hand-added for exactly one family (:100-104): `columbus-5`, `phoenix-1`, `terra v2`. Any other cosmos chain-id form (`cosmoshub-4`, `osmosis-1`, `noble-1`, `dydx-mainnet-1`, `akashnet-2`) throws UnknownChainError.

That shape is historical (Terra work added them), not principled - and it makes downstream family inference fragile: callers that catch the throw degrade to "unknown family" and silently skip chain-specific handling for inputs the normalizer could easily resolve.

## fix shape

Auto-derive chain-id aliases from the canonical cosmos chain-id table (`packages/core/chain/chains/cosmos/chainInfo.ts` cosmosChainId map) with the same ownership-dedupe treatment the ticker aliases get, instead of hand-listing. EVM numeric/hex chain-ids (`1`, `0x1`, `eip155:1`) are worth considering in the same pass if cheap - same derivation principle from evm chainInfo.

Also worth a doc line in the file header: chain-id forms ARE supported input, so future callers stop pre-normalizing.

effort: S / risk: LOW (additive aliases; ownership-dedupe already prevents ambiguous claims) / confidence: HIGH

found by improve round 18 deep audit (2026-07-13)

gomesalexandre:
Verified against current `main` (`packages/sdk/src/utils/normalizeChain.ts`) - this is already fixed, not reproducible.

`normalizeChain` falls back to `getCosmosChainByChainId` (`packages/core/chain/chains/cosmos/chainInfo.ts:21-23`), which derives from the canonical `cosmosChainId` map (line 4-15) covering every cosmos chain, not just Terra. That fallback line was added in commit 4f111d1f4 (`fix(sdk): infer decoder family from canonical chains`, #1286).

All the specific examples cited in this issue as broken already resolve correctly:

```
normalizeChain('cosmoshub-4')      -> Chain.Cosmos
normalizeChain('osmosis-1')        -> Chain.Osmosis
normalizeChain('noble-1')          -> Chain.Noble
normalizeChain('dydx-mainnet-1')   -> Chain.Dydx
[...]

PR #1286, opened 2026-07-15
fix(sdk): infer decoder family from canonical chains

### LINKED ISSUE #1189: decodeFromToolResult infers family from stale local alias subsets instead of SDK canonicals
## Summary
`decodeFromToolResult()` still infers `family` from small local EVM/Cosmos alias sets instead of the SDK's canonical chain metadata.

## Evidence
- `packages/sdk/src/tools/decode/fromToolResult.ts:5-34` hardcodes `EVM_CHAINS` / `COSMOS_CHAINS`
- Those sets omit legitimate SDK-supported chains like Mantle, Hyperliquid, Sei, THORChain, MayaChain, and Kujira
- The SDK already exposes canonical chain normalization/classification helpers elsewhere (`normalizeChain()` + `getChainKind()`)

## Impact
The SDK's canonical decoder can fail closed with `cannot determine chain family` for valid SDK-supported chains unless every caller manually passes `family`. That weakens the point of having one shared bytes-oracle and pressures downstream callers to paper over it locally.

This is core decoder behavior and should consume the same chain registry metadata as the rest of the SDK, not maintain a second local chain-family table.

## Fix sketch
Replace the local alias subsets with `normalizeChain()` + `getChainKind()`, then only map the resulting kind into the decoder families supported today (`evm` / `cosmos`). Add regression tests for at least one omitted EVM chain and one omitted Cosmos-family chain.

## Effort
S

## Report
Round-14 report gist: https://gist.github.com/gomesalexandre/f9e8d6f3d6f8f0a968807cd1cff3435a

rcoderdev:
Closes #1189
Uses canonical SDK chain normalization and classification for decoder family inference. Runtime proof: freshly encoded Mantle EVM and THORChain Cosmos payloads decode without a family override and preserve their recipients.

## Summary by CodeRabbit

- **New Features**
  - Improved automatic detection of EVM and Cosmos chain families using canonical chain metadata.
  - Added support for resolving additional Cosmos chain IDs and aliases, including `cosmoshub-4`, `osmosis-1`, and `gaia`.

- **Bug Fixes**
  - Unknown or unsupported chains now fail safely when their family cannot be determined.

- **Tests**
  - Expanded coverage for chain aliases, Cosmos networks, and decoder family detection.


gomesalexandre (approved):
## lgtm 🫡

**APPROVE** - clean canonical-registry swap, fail-closed behavior preserved, no collision with the in-flight chain-id family PRs despite touching adjacent territory.

praise: this is the correct fix shape for #1189 (stale local `EVM_CHAINS`/`COSMOS_CHAINS` sets) - delegates to `getChainKind(normalizeChain(chain))` instead of maintaining a second hand-curated table, and the try/catch explicitly preserves the decoder's fail-closed result for unknown chains instead of accidentally widening the surface.

### correctness check
- `normalizeChain()` throws `UnknownChainError` for anything unrecognized (verified in source) - caught, falls through to `return undefined` -> `inferFamily` fails closed exactly like before, just with a wider canonical alias set instead of a stale local one.
- `getChainKind` only maps `evm`/`cosmos` chains to those literals (utxo/sui/solana/etc all return their own kind, correctly NOT matched by the `if (kind === 'evm' || kind === 'cosmos')` check) - so unsupported-family chains still fail closed same as before, no accidental widening into families the decoder can't actually handle.
- New `getCosmosChainByChainId` fallback in `normalizeChain.ts` is checked AFTER the hand-curated `aliasToChain` table, so pre-existing aliases (`phoenix-1`, `columbus-5`) resolve the same way they always did - this is pure additive coverage for the previously-missing chain-ids (cosmoshub-4, osmosis-1, noble-1, dydx-mainnet-1, akashnet-2, etc), not a behavior change for what already worked.

### collision check vs in-flight chain-id work
- **#1256** ("resolve typed EVM chain ids through canonical registry") operates on a genuinely different pipeline stage - it rewrites `tx.chainId` (the numeric EIP-155 id embedded in a TYPED tx's own bytes) into the envelope's output `chain` field, INSIDE `decodeEvmTx`, which only runs AFTER family dispatch has already happened. This PR's `inferFamily` change is what PICKS the decoder (evm vs cosmos) in the first place, using the caller's `chain` hint argument - different input, different stage, no logic overlap.
- Only real overlap: both PRs add new `it()` blocks to `packages/sdk/tests/unit/decode/fromToolResult.test.ts`, in different `describe` blocks (#1256 in the EVM-half block, this one in the dispatch/guards block) - mechanical rebase-order friction only, not a logic conflict.
- Bonus find: this PR's `getCosmosChainByChainId` fallback also fully resolves open issue **#1229** ("derive cosmos chain-id aliases in normalizechain from canonical chain-id table, terra-only today") even though the PR only references #1189. Worth a follow-up comment linking/closing #1229 so it doesn't sit open as a stale duplicate - not blocking the merge.

### receipts
Ran locally (isolated scratch clone, `yarn build:shared` first per the monorepo's workspace-exports gotcha):
- `yarn vitest run tests/unit/decode/fromToolResult.test.ts tests/unit/utils/normalizeChain.test.ts` -> 257 passed
- `yarn tsc --noEmit -p .` -> clean
CI green (Unit/Integration/Lint-and-Type/Test Summary all SUCCESS).

gomesalexandre (approved):
Approving AFTER the runtime receipt. Runtime-verified on the built dist: the previously-missing chains now resolve to the RIGHT family (mantle/blast/hyperliquid/sei->evm, thorchain/mayachain/kujira->cosmos), wire-format cosmos chain-ids resolve (cosmoshub-4/osmosis-1/noble-1/dydx-mainnet-1/etc->cosmos), and THE key fund-safety check: NO wrong-family mis-inference - bitcoin/solana/sui/ton/xrp/cardano/polkadot each resolve to their own distinct ChainKind (never evm/cosmos), so decodeFromToolResult fails closed decoded:false instead of decoding with the wrong parser. Garbage input fails closed (UnknownChainError->decoded:false, no crash), no regression (ethereum->evm, cosmos/osmosis->cosmos), PR's 257 tests + full SDK suite 2789 green (the 3 MPC-singleton failures on an ad-hoc run were a pre-existing test-isolation artifact, not this diff - confirmed via CI's exact vitest config). Merging on green.

PR #1288, opened 2026-07-15
Group coordinated Dependabot updates

rcoderdev:
## Summary by CodeRabbit

* **Chores**
  * Improved automated dependency update management by grouping compatible minor and patch updates.
  * Added safeguards to defer selected major dependency updates and specific Electron releases.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,35 @@ updates:
       vultisig:
         patterns:
           - "@vultisig/*"
+      formatjs:
+        patterns:
+          - "@formatjs/*"
+        update-types:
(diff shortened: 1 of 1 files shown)


gomesalexandre (approved):
Looks good.

- the new groups are minor/patch-only, so this stays in coordinated-safe territory
- `@typescript-eslint/*` is correctly carved out from the generic dev-deps bucket
- the `dependency-cruiser` major ignore matches the current Node 20 CI constraint
- keeping Electron pinned to ignore only `43.x` is intentional and narrower than blocking all future majors

Remote conceptual pass only. CI is green on the current head.

PR #1290, opened 2026-07-15
fix(tokens): keep non-EVM known-token ids canonical

### LINKED ISSUE #1289: correctness: non-EVM known-token lookup lowercases canonical ids and can false-hit
## Summary
Known-token lookup currently lowercases token ids globally, so lowercase non-EVM ids can falsely resolve as canonical registry hits. That is wrong for case-sensitive ids like Solana mints and XRPL issued-currency token ids.

## Evidence
- `packages/core/chain/coin/knownTokens/index.ts:824-835` — index construction lowercases lookup ids globally
- `packages/core/chain/coin/price/resolveTokenPriceId.ts:22-27` — first-party price lookup reads the curated registry
- `packages/core/chain/swap/native/utils/getNativeSwapDecimals.ts:20-24` — native swap token-decimals lookup uses the same registry path
- `packages/sdk/src/Vultisig.ts:1317-1327` — `Vultisig.getKnownToken(...)` reads the same registry
- `packages/sdk/src/vault/services/TokenDiscoveryService.ts:34-45` — token resolution fast-path reads the same registry

## Impact
- Lowercased Solana mint ids and lowercased XRPL `<currency>.<issuer>` ids can be incorrectly treated as canonical known tokens.
- This is a correctness bug at the SDK registry boundary and encourages downstream consumers to depend on wrong non-EVM token-id semantics.

## Fix sketch
- Keep case-insensitive lookup only for EVM contract addresses.
- Require exact canonical ids for non-EVM assets.
- Add regression tests proving lowercase Solana/Ripple ids fail while canonical ids succeed.

## Effort
Small.

## Report
Round 29 report: https://gist.github.com/gomesalexandre/141ba146713327a032b13d995176b1e3

gomesalexandre:
## Summary
- make known-token lookup chain-sensitive instead of globally lowercasing ids
- keep EVM contract-address lookup case-insensitive
- require exact canonical ids for non-EVM assets like Solana mints and XRPL issued-currency ids

Closes #1289

## Why
The known-token registry was lowercasing lookup ids globally, so lowercase non-EVM ids could false-hit as canonical known tokens. That is wrong for case-sensitive ids like Solana mints and XRPL `<currency>.<issuer>` ids.

## What changed
- added shared core helpers for chain-sensitive known-token lookup
- rewired first-party consumers (`resolveTokenPriceId`, `getNativeSwapDecimals`, `Vultisig.getKnownToken`, `TokenDiscoveryService.resolveToken`) through the shared helper
- added regression coverage proving Ethereum stays case-insensitive while Solana/Ripple require canonical ids
- added a changeset for `@vultisig/core-chain` and `@vultisig/sdk`

## Verification
- `yarn test:core packages/core/chain/coin/knownTokens/index.test.ts`
- `yarn test:core packages/core/chain/coin/price/__tests__/resolveTokenPriceId.test.ts`
- `yarn build:shared`
- `yarn workspace @vultisig/sdk vitest run tests/unit/tools/token/knownTokensSolana.test.ts tests/unit/Vultisig.static-methods.test.ts tests/unit/vault/services/TokenDiscoveryService.test.ts`
- `yarn build:sdk`
- built-dist runtime proof:
  ```json
  {
    "ethereumLowercase": "USDC",
    "solanaCanonical": "USDC",
    "solanaLowercase": null,
    "rippleCanonical": "RLUSD",
    "rippleLowercase": null
  }
  ```

## Report
- Round 29 report: https://gist.github.com/gomesalexandre/141ba146713327a032b13d995176b1e3

## Risk
- low: behavior-only correctness fix at the registry lookup boundary
- EVM case-insensitive behavior is preserved and covered by tests
- non-EVM change is intentionally fail-closed to canonical ids only

### DIFF
--- a/.changeset/r29-known-token-case-sensitive-non-evm.md
+++ b/.changeset/r29-known-token-case-sensitive-non-evm.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Keep known-token lookup case-insensitive only for EVM contract addresses, and require exact canonical ids for non-EVM assets like Solana mints and XRPL issued-currency token ids.
--- a/packages/core/chain/coin/knownTokens/index.ts
+++ b/packages/core/chain/coin/knownTokens/index.ts
@@ -1327,9 +1327,11 @@ export const knownTokens = makeRecord(Object.values(Chain), chain => {
 
 type KnownIndex = Record<Chain, Record<string, KnownCoin>>
 
-export const getKnownTokenIndexId = (chain: Chain, id: string): string => (isChainOfKind(chain, 'evm') ? id.toLowerCase() : id)
+export const getKnownTokenIndexId = (chain: Chain, id: string): string =>
+  isChainOfKind(chain, 'evm') ? id.toLowerCase() : id
 
-export const getKnownTokenById = (chain: Chain, id: string): KnownCoin | undefined => knownTokensIndex[chain]?.[getKnownTokenIndexId(chain, id)]
+export const getKnownTokenById = (chain: Chain, id: string): KnownCoin | undefined =>
+  knownTokensIndex[chain]?.[getKnownTokenIndexId(chain, id)]
 
 export const knownTokensIndex: KnownIndex = makeRecord(Object.values(Chain), chain => {
   const byId: Record<string, KnownCoin> = {}
--- a/packages/core/chain/coin/knownTokens/index.ts
+++ b/packages/core/chain/coin/knownTokens/index.ts
@@ -1,4 +1,5 @@
 import { Chain } from '@vultisig/core-chain/Chain'
+import { isChainOfKind } from '@vultisig/core-chain/ChainKind'
 import { rippleKnownIssuedTokens } from '@vultisig/core-chain/chains/ripple/issuedCurrency'
 import { makeRecord } from '@vultisig/lib-utils/record/makeRecord'
 import { omit } from '@vultisig/lib-utils/record/omit'
@@ -1326,11 +1327,15 @@ export const knownTokens = makeRecord(Object.values(Chain), chain => {
 
 type KnownIndex = Record<Chain, Record<string, KnownCoin>>
 
+export const getKnownTokenIndexId = (chain: Chain, id: string): string => (isChainOfKind(chain, 'evm') ? id.toLowerCase() : id)
+
+export const getKnownTokenById = (chain: Chain, id: string): KnownCoin | undefined => knownTokensIndex[chain]?.[getKnownTokenIndexId(chain, id)]
+
 export const knownTokensIndex: KnownIndex = makeRecord(Object.values(Chain), chain => {
   const byId: Record<string, KnownCoin> = {}
   for (const coin of knownTokens[chain] ?? []) {
     if (!coin.id) continue
-    byId[coin.id.toLowerCase()] = coin
+    byId[getKnownTokenIndexId(chain, coin.id)] = coin
   }
   return byId
 })
--- a/packages/core/chain/swap/native/utils/getNativeSwapDecimals.ts
+++ b/packages/core/chain/swap/native/utils/getNativeSwapDecimals.ts
@@ -1,6 +1,6 @@
 import { chainFeeCoin } from '@vultisig/core-chain/coin/chainFeeCoin'
 import { CoinKey } from '@vultisig/core-chain/coin/Coin'
-import { knownTokensIndex } from '@vultisig/core-chain/coin/knownTokens'
+import { getKnownTokenById } from '@vultisig/core-chain/coin/knownTokens'
 import { isFeeCoin } from '@vultisig/core-chain/coin/utils/isFeeCoin'
 
 import { Chain } from '../../../Chain'
@@ -19,7 +19,7 @@ export const getNativeSwapDecimals = (coin: CoinKey) => {
       return chainFeeCoin[coin.chain].decimals
     }
 
-    const known = coin.id ? knownTokensIndex[coin.chain][coin.id.toLowerCase()] : undefined
+    const known = coin.id ? getKnownTokenById(coin.chain, coin.id) : undefined
 
     if (known) {
       return known.decimals
--- a/packages/sdk/tests/unit/vault/services/TokenDiscoveryService.test.ts
+++ b/packages/sdk/tests/unit/vault/services/TokenDiscoveryService.test.ts
@@ -5,19 +5,46 @@ vi.mock('@vultisig/core-chain/coin/find', () => ({
   findCoins: vi.fn(),
 }))
 
-vi.mock('@vultisig/core-chain/coin/knownTokens', () => ({
-  knownTokensIndex: {
-    Ethereum: {
-      '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48': {
-        chain: 'Ethereum',
-        id: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
-        ticker: 'USDC',
-        decimals: 6,
-        logo: 'usdc.png',
-        priceProviderId: 'usd-coin',
-      },
+const knownTokensByChain: Record<string, Record<string, any>> = {
+  Ethereum: {
+    '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48': {
+      chain: 'Ethereum',
+      id: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
+      ticker: 'USDC',
+      decimals: 6,
+      logo: 'usdc.png',
+      priceProviderId: 'usd-coin',
     },
   },
+  Solana: {
+    EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v: {
+      chain: 'Solana',
+      id: 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v',
+      ticker: 'USDC',
+      decimals: 6,
+      logo: 'usdc.png',
+      priceProviderId: 'usd-coin',
+    },
+  },
+  Ripple: {
+    '524C555344000000000000000000000000000000.rMxCKbEDwqr76QuheSUMdEGf4B9xJ8m5De': {
+      chain: 'Ripple',
+      id: '524C555344000000000000000000000000000000.rMxCKbEDwqr76QuheSUMdEGf4B9xJ8m5De',
+      ticker: 'RLUSD',
+      decimals: 15,
+      logo: 'rlusd',
+      priceProviderId: 'rlusd',
+    },
+  },
+}
+
+vi.mock('@vultisig/core-chain/coin/knownTokens', () => ({
+  getKnownTokenById: vi.fn((chain: string, id: string) => {
+    if (chain === 'Ethereum') {
+      return knownTokensByChain[chain]?.[id.toLowerCase()]
+    }
+    return knownTokensByChain[chain]?.[id]
+  }),
 }))
 
 vi.mock('@vultisig/core-chain/coin/token/metadata', () => ({
@@ -129,6 +156,9 @@ describe('TokenDiscoveryService', () => {
   })
 
   describe('resolveToken', () => {
+    const solanaUsdcMint = 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v'
+    const rippleRlusdId = '524C555344000000000000000000000000000000.rMxCKbEDwqr76QuheSUMdEGf4B9xJ8m5De'
+
     it('should return token from known tokens registry (fast path)', async () => {
       const token = await service.resolveToken(Chain.Ethereum, '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48')
 
@@ -151,6 +181,20 @@ describe('TokenDiscoveryService', () => {
 
       expect(token).not.toBeNull()
       expect(token.ticker).toBe('USDC')
+    })
+
+    it('requires exact-case lookup for known Solana mints', async () => {
+      const token = await service.resolveToken(Chain.Solana, solanaUsdcMint)
+
+      expect(token.ticker).toBe('USDC')
+      await expect(service.resolveToken(Chain.Solana, solanaUsdcMint.toLowerCase())).rejects.toThrow(VaultError)
+    })
+
+    it('requires exact canonical ids for known Ripple issued currencies', async () => {
+      const token = await service.resolveToken(Chain.Ripple, rippleRlusdId)
+
+      expect(token.ticker).toBe('RLUSD')
+      await expect(service.resolveToken(Chain.Ripple, rippleRlusdId.toLowerCase())).rejects.toThrow(VaultError)
     })
 
     it('should fall back to chain API for unknown tokens', async () => {
--- a/packages/core/chain/coin/knownTokens/index.test.ts
+++ b/packages/core/chain/coin/knownTokens/index.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from 'vitest'
+
+import { Chain } from '../../Chain'
+import { rippleTokenId } from '../../chains/ripple/issuedCurrency'
+import { getKnownTokenById, getKnownTokenIndexId, knownTokensIndex } from '.'
+
+describe('knownTokens chain-sensitive lookup', () => {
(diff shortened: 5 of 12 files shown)


NeOMakinG (commented):
Draft — **COMMENTED, not approved**. The premise is right and the direction is safety-positive, but the change alters the key shape of a **public export** and only rewires the four consumers that live in this repo. Everything below is from executing the code, not reading the diff.

## testing approach

Worktree `/tmp/wt-sdk-1290` @ `5e6ef459`, baseline worktree at `origin/main` (`802910d7`).

1. **Registry enumeration** — dumped every non-EVM `knownTokens[chain]` entry with an `id` (60 across Cosmos/Osmosis/Kujira/Terra/TerraClassic/THORChain/MayaChain/Solana/Polkadot/Ton/Ripple/Tron) and checked (a) case-fold collisions within each chain, (b) `knownTokens[c].filter(id).length` vs `Object.keys(knownTokensIndex[c]).length`.
2. **Differential probe** — replayed the *verbatim lookup expressions* that live in `vultisig-windows`, `agent-backend-ts` and `mcp-ts` against both `origin/main` and this branch.
3. **Mutation verification** — reverted `getKnownTokenIndexId` to unconditional `id.toLowerCase()` and re-ran each new suite.
4. Full `yarn test:core` on both worktrees; `yarn workspace @vultisig/sdk vitest run` on the four touched suites.
5. `gh api .../check-runs` for CI truth.

**Test-evidence results.** Core suites are real guards: `knownTokens/index.test.ts` + `resolveTokenPriceId.test.ts` = **35 pass on branch, 4 fail under mutation**. Two pre-existing/environmental red files (`mpc/keysign/fee/resolvers/cosmos.test.ts`, `mpc/tx/compile/compileTx.golden.test.ts`) fail identically on `origin/main` — not caused by this PR.

**CI is vacuous here.** `GET /commits/5e6ef459/check-runs` returns `total_count: 0`. The single green status is CodeRabbit's `Review skipped: draft pull request`. No test, build, or lint job has run against this HEAD.

## blocking

### 1. `knownTokensIndex` is a public export whose key shape changes here; downstream `[id.toLowerCase()]` lookups go silently inert on every non-EVM chain

`knownTokensIndex` is exported from `@vultisig/sdk` (`packages/sdk/src/index.ts:748`), from the RN entrypoint (`packages/sdk/src/platforms/react-native/index.ts:430`), from `packages/sdk/src/tools/index.ts:78`, and from `@vultisig/core-chain/coin/knownTokens` directly. Its documented contract until this PR was *all keys lowercased*. This PR changes that for non-EVM chains and rewires only `resolveTokenPriceId`, `getNativeSwapDecimals`, `Vultisig.getKnownToken`, `TokenDiscoveryService.resolveToken` — all in-repo. `getKnownTokenById` is **not re-exported from `@vultisig/sdk`**, so downstream has no drop-in replacement.

Differential probe, running each downstream expression byte-for-byte:

| expression (verbatim, with source) | `origin/main` | this branch |
|---|---|---|
| `knownTokensIndex[chain]?.[recipient.toLowerCase()]` — abts `src/mastra/tools/mcp/lib/dangerous-addresses.ts:158`, Tron USDT `TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t` | `USDT` (guard fires) | **MISS (guard inert)** |
| same, Solana USDC `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v` | `USDC` | **MISS** |
| `knownTokensIndex[Chain.Ton]?.[address.toLowerCase()]` — windows `core/ui/chain/tx/utils/useTonMessageDecode.ts:203`, `EQCxE6mUtQJKFnGfaROTKOt1lZbDiiX1kCixRv7Nw2Id_sDs` | `USDT` | **MISS** |
| Cosmos IBC USDC, canonical uppercase hash | MISS | `USDC` (**fixed by this PR**) |
| Cosmos IBC USDC, lowercased | `USDC` | MISS |

The worst of these is **`dangerous-addresses.ts:158`**. That is the fail-closed guard that refuses to build a transaction sending a token to its own token contract (permanent burn). It is chain-generic and keyed on the lowercased recipient. After this change it stops matching on Tron, Solana, TON, Sui and Ripple — a fund-loss guard that goes **fail-open**, in a repo that consumes `"@vultisig/sdk": "^2.20.0"`.

Other confirmed-affected call sites, all still lowercasing:
- `vultisig-windows/core/ui/mpc/keysign/tx/TxSuccess.tsx:117` — chain-generic; swap-pair token row disappears for non-EVM.
- `vultisig-windows/core/ui/agent/tools/shared/assetResolution.ts:59` (`makeAssetInfoFromToken`) — returns `null`, so the caller loses the token's `decimals`.
- `vultisig-windows/core/ui/chain/tx/utils/useTonMessageDecode.ts:203` — 100% inert for all 8 registry jettons; this is the TON message decode feeding the keysign display.
- `agent-backend-ts/src/mastra/tools/mcp/lib/resolveToken.ts:334` and its `mcp-ts` twin at `src/lib/resolveToken.ts:236`. Their own comment says it out loud: *"the index is keyed case-INSENSITIVELY (lowercased), so a case-mutated input still hits ... Solana mints / Tron contracts / Sui coin-types are case-SENSITIVE base58"*. That comment is describing exactly the contract this PR removes, and it is the token-resolution path for sends.

Not affected: `asset-registry.ts:475` (gated behind an EVM hex regex). Actually *fixed*: `vultisig-windows/core/ui/storage/defiPositions.tsx:200`, which looks up the raw un-lowercased id and was broken for non-EVM before.

**Suggested fix — either:**
- (a) export `getKnownTokenById` (and `getKnownTokenIndexId`) from `@vultisig/sdk` + the RN entrypoint, and land companion PRs in `agent-backend-ts`, `mcp-ts` and `vultisig-windows` migrating all the `[x.toLowerCase()]` sites, merged before/with this; **or**
- (b) leave `knownTokensIndex` keyed exactly as today (lowercased, back-compat) and add a *separate* canonically-keyed map that `getKnownTokenById` reads, so the published contract does not change under callers.

Option (b) is the lower-risk one for a patch train.

### 2. `patch` changeset for a breaking behaviour change on a public API

`.changeset/r29-known-token-case-sensitive-non-evm.md:2-3` bumps both packages as `patch`. Downstream ranges are carets — `agent-backend-ts` `"@vultisig/sdk": "^2.20.0"`, `vultiagent-app` `"^2.20.0"`, `mcp-ts` `"^2.18.6"` — so this lands in all three on the next `install` with no downstream code change and no signal. `Vultisig.getKnownToken()` also changes observable behaviour for non-EVM ids (`packages/sdk/src/Vultisig.ts:1318`). Minor at minimum, and coordinated with #1 above.

## should-fix

### 3. `getNativeSwapDecimals` fail-OPENS to a wrong-magnitude default on a miss

`packages/core/chain/swap/native/utils/getNativeSwapDecimals.ts:22-30`: a MayaChain token that misses the registry falls through to `return chainFeeCoin[Chain.THORChain].decimals` = **8**. But MAYA and AZTEC are **4** decimals (confirmed in the registry dump). That is a 10^4 error on a swap amount, from a silent lookup miss — the exact shape flagged in `abts#1655`.

Probed both sides: `knownTokensIndex[MayaChain]['maya']` hits on main *and* branch; `['MAYA']` misses on both. So this specific spelling is not a regression introduced here. But this PR strictly narrows the set of ids that hit, and this is the one non-EVM consumer where a miss is a magnitude error rather than a `undefined`/`null`. Note the registry stores `maya` / `aztec` / `thor.kuji` / `tcy` **lowercase**, while THOR/Maya swap-API asset notation is uppercase (`MAYA.MAYA`, `THOR.TCY` — see `toNativeSwapAsset`), so the two spellings are one boundary apart. `getNativeSwapDecimals` is itself a public export (`packages/sdk/src/index.ts:704`).

Suggested fix: on `Chain.MayaChain` with a non-fee coin, throw (or return `undefined`) on a registry miss instead of silently returning THORChain's 8.

### 4. The new `TokenDiscoveryService` cases are tautological — they pass in both mutation states

`packages/sdk/tests/unit/vault/services/TokenDiscoveryService.test.ts:42-47` replaces the `knownTokensIndex` mock with a `getKnownTokenById` mock that **reimplements the production branching** (`chain === 'Ethereum' ? id.toLowerCase() : id`). The two new tests at `:183-195` then assert against that reimplementation, not against the real helper.

Mutation-verified: with `getKnownTokenIndexId` reverted to unconditional `id.toLowerCase()`, `knownTokensSolana.test.ts` + `knownTokens/index.test.ts` go red (4 failures) but **`TokenDiscoveryService.test.ts` stays fully green**. It is not a guard.

Suggested fix: `vi.importActual` the real `getKnownTokenById` (it is a pure sync lookup, no network), or assert the mock was *called with* the raw id and let the real helper decide.

### 5. `knownTokensPolkadot.test.ts` still encodes the old contract

`packages/sdk/tests/unit/tools/token/knownTokensPolkadot.test.ts:17-24` and `:42-45` still document `knownTokensIndex` as "the lowercased-key map" and index via `.toLowerCase()`. It passes only because `'1337'` / `'1984'` are casing-invariant. Update it alongside the Solana file so the docs don't contradict each other.

## q

### 6. `getKnownTokenIndexId` throws on a non-string id

`packages/core/chain/coin/knownTokens/index.ts:824` calls `id.toLowerCase()` directly. Producers elsewhere in this ecosystem hand around `string | number | bigint` for ids and denoms (Polkadot ids are literally the numeric strings `'1337'`/`'1984'`). A `number` reaches the EVM branch as a `TypeError` rather than a miss. `String(id).toLowerCase()` costs nothing.

### 7. TON form ambiguity survives the fix

The registry stores the bounceable friendly form (`EQ…`). Downstream normalizers can produce the non-bounceable `UQ…` or the raw `0:<hex>` form, which are the same jetton but different strings. Once lookups are exact-match, a case fix alone isn't sufficient — is a per-chain canonicalisation step at the boundary in scope, or a follow-up?

## praise

The **collision axis is clean, and I verified it rather than assumed it**: across all 60 non-EVM registry ids there are zero case-fold collisions and zero `list.length != index.length` mismatches on any chain, so `knownTokensIndex` was never actually losing an entry to lowercasing today — and this change removes that latent surface without creating one. Every non-EVM id in the registry is already in real canonical wire form (IBC `ibc/<UPPERCASE-HEX>`, TON friendly `EQ…`, Solana/Tron base58 mixed-case, XRPL uppercase hex currency + `r…` issuer, Cosmos `factory/…` lowercase). And the core-side tests are genuine guards — mutation-verified 4-fail-before / 0-fail-after.

## risk

**Medium-high as it stands, low once #1 and #2 are resolved.** The in-repo change is correct and well covered; the exposure is entirely in the published contract. The single most damaging concrete consequence found is the abts token-contract-recipient guard going fail-open on Tron/Solana/TON, which is a fund-loss guard, reachable with no attacker involvement, shipping under a patch bump into a caret range.

## verdict

**COMMENTED** (draft — not approving). Blocking #1/#2 need to be resolved or the change needs restructuring to preserve the `knownTokensIndex` key shape before this leaves draft.

---
*Reviewed by two independent lanes (Claude + Codex `gpt-5.3`), findings deduped at synthesis. No files were modified on this branch.*

packages/core/chain/coin/knownTokens/index.ts
```diff
@@ -820,11 +821,17 @@ export const knownTokens = makeRecord(Object.values(Chain), chain => {
...
+
+export const getKnownTokenById = (chain: Chain, id: string): KnownCoin | undefined =>
+  knownTokensIndex[chain]?.[getKnownTokenIndexId(chain, id)]
+
 export const knownTokensIndex: KnownIndex = makeRecord(Object.values(Chain), chain => {
   const byId: Record<string, KnownCoin> = {}
   for (const coin of knownTokens[chain] ?? []) {
     if (!coin.id) continue
-    byId[coin.id.toLowerCase()] = coin
+    byId[getKnownTokenIndexId(chain, coin.id)] = coin
```
NeOMakinG, line 1340: `blocker:` this re-keys `knownTokensIndex`, which is a **public export** (`packages/sdk/src/index.ts:748`, `platforms/react-native/index.ts:430`, `tools/index.ts:78`), and the four in-repo consumers are not the only readers.

Differential probe, `origin/main` vs this branch, running the downstream expressions verbatim:

- `agent-backend-ts/src/mastra/tools/mcp/lib/dangerous-addresses.ts:158` `knownTokensIndex[chain]?.[recipient.toLowerCase()]` with Tron USDT `TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t`: main -> `USDT` (guard fires), branch -> **MISS**. That is the fail-closed "refusing to send a token to its own contract" burn guard going **fail-open** on Tron/Solana/TON/Sui/Ripple.
- `vultisig-windows/core/ui/chain/tx/utils/useTonMessageDecode.ts:203`: inert for all 8 registry jettons.
- `agent-backend-ts .../resolveToken.ts:334` and `mcp-ts src/lib/resolveToken.ts:236` — their own comment documents the case-insensitive contract this removes.
- also `vultisig-windows TxSuccess.tsx:117`, `assetResolution.ts:59`.

`getKnownTokenById` isn't re-exported from `@vultisig/sdk`, so downstream has no drop-in.

Suggested fix: keep `knownTokensIndex` keyed exactly as today and add a *separate* canonically-keyed map behind `getKnownTokenById` — or export the helper and land companion PRs in the three downstream repos before this merges.

.changeset/r29-known-token-case-sensitive-non-evm.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
```
NeOMakinG, line 3: `blocker:` `patch` for a breaking public-API behaviour change. `knownTokensIndex`'s key shape and `Vultisig.getKnownToken()`'s non-EVM behaviour both change. Downstream pins are carets — `agent-backend-ts` `^2.20.0`, `vultiagent-app` `^2.20.0`, `mcp-ts` `^2.18.6` — so this floats in on the next `install` with no downstream code change and no signal.

Suggested fix: `minor` at minimum, and coordinated with the downstream migration.

packages/core/chain/swap/native/utils/getNativeSwapDecimals.ts
```diff
@@ -19,7 +19,7 @@ export const getNativeSwapDecimals = (coin: CoinKey) => {
       return chainFeeCoin[coin.chain].decimals
     }
 
-    const known = coin.id ? knownTokensIndex[coin.chain][coin.id.toLowerCase()] : undefined
+    const known = coin.id ? getKnownTokenById(coin.chain, coin.id) : undefined
```
NeOMakinG, line 22: `should-fix:` this is the one non-EVM consumer where a lookup miss is a **magnitude** error, not a `null`. On a miss the function falls through to `return chainFeeCoin[Chain.THORChain].decimals` = **8**, but MAYA and AZTEC are **4** decimals — a 10^4 error on a swap amount, silently.

Probed both sides: `knownTokensIndex[MayaChain]['maya']` hits on main and branch, `['MAYA']` misses on both — so not a regression introduced here. But the registry stores `maya`/`aztec` lowercase while THOR/Maya swap-API asset notation is uppercase (`MAYA.MAYA`, see `toNativeSwapAsset`), and this PR strictly narrows the hit set.

Suggested fix: on `Chain.MayaChain` with a non-fee coin, throw (or return `undefined`) on a registry miss rather than fail-open to THOR's 8.

packages/sdk/tests/unit/vault/services/TokenDiscoveryService.test.ts
```diff
@@ -5,19 +5,46 @@ vi.mock('@vultisig/core-chain/coin/find', () => ({
...
+      ticker: 'RLUSD',
+      decimals: 15,
+      logo: 'rlusd',
+      priceProviderId: 'rlusd',
+    },
+  },
+}
+
+vi.mock('@vultisig/core-chain/coin/knownTokens', () => ({
+  getKnownTokenById: vi.fn((chain: string, id: string) => {
```
NeOMakinG, line 42: `should-fix:` this mock **reimplements the production branching** (`chain === 'Ethereum' ? id.toLowerCase() : id`), so the two new cases at `:183-195` assert against the reimplementation rather than against the real helper.

Mutation-verified: with `getKnownTokenIndexId` reverted to unconditional `id.toLowerCase()`, `knownTokensSolana.test.ts` + `knownTokens/index.test.ts` go red (4 failures) while **this file stays fully green**. It passes in both states, so it is not a guard.

Suggested fix: `vi.importActual` the real `getKnownTokenById` (pure sync lookup, no network), or assert the mock was called with the raw id and let the real helper decide.

rcoderdev:
Closing this because it changes the public knownTokensIndex key contract in a way that can make downstream non-EVM burn guards fail open, and the stale branch would also revert unrelated current SDK work.

PR #1306, opened 2026-07-16
Preserve Cosmos uint64 account identifiers exactly

rcoderdev:
Preserves QBTC uint64 account number and sequence values as exact bigint fields while keeping numeric fields for compatibility. The Cosmos MPC signing path consumes the exact fields and constructs unsigned WalletCore Long values directly from decimal bigint strings.

Real QA: production WalletCore signing input emitted account number 9007199254740993 and sequence 18446744073709551615 exactly with unsigned flags; production QBTC parser coverage accepted uint64 decimal boundaries and rejected invalid numeric and non-string forms.

## Summary by CodeRabbit

- **New Features**
  - Preserved full 64-bit QBTC account numbers and sequences (including max values) and added exact bigint variants for signing workflows while retaining legacy numeric fields for compatibility.
  - Improved Cosmos signing input handling to keep `accountNumber` and `sequence` accurate beyond JavaScript’s safe-integer range.

- **Bug Fixes**
  - Prevented rounding/corruption of large account and sequence values during signing.
  - Added stricter validation and clearer failures for invalid QBTC account data (negative, fractional, out-of-range, or unexpected numeric formats).

- **Tests**
  - Added regression coverage for QBTC parsing and Cosmos payload construction with large uint64 inputs.

- **Chores**
  - Updated changeset for patch releases.


gomesalexandre (commented):
Reviewed the correctness end-to-end (independent trace, not just the description) - the QBTC fix is right, and it actually fixes something **more severe than the title advertises**. Nice one.

## Correctness: confirmed, and worse-than-advertised
- The old `getQbtcAccountInfo` did `Number(account_number_string)` (irreversible precision loss > 2^53-1), then `BigInt(accountNumber)` re-widened the already-lossy double - a double-narrow-then-rewiden that doesn't recover precision. New `parseUint64()` (`/^\d+$/` + `<= maxUint64`, straight to bigint) fixes it, and `qbtc.ts` consumes the bigint directly (confirmed no other consumer of the old lossy fields exists).
- The bigger find is the shared hop: `new Long(Number(accountNumber))` in `getCosmosSigningInputs` wasn't just a >2^53 IEEE-754 bug - single-arg `new Long(n)` sets `low = n|0` (ToInt32), `high = 0`, so it **wraps at 2^31 (~2.1B), not 2^53**. `Long.fromString(x.toString(), true)` is exact bigint-equivalent (no float in the digit loop; `true` binds `unsigned`, not the radix overload). Since this lives in shared `getCosmosSigningInputs`, it's a **chain-wide** signing-path correctness fix (ATOM/OSMO/THOR/MAYA/Terra/Kujira/…), not just QBTC - the PR undersells its own scope.

## Test coverage: adequate, real SignDoc-level proof
Parse-layer goldens incl. `2^53+1` and exact `2^64-1`, plus an `index.ibcTransfer` test through the real `getCosmosSigningInputs` with `sequence = 18_446_744_073_709_551_615n` asserting exact `Long.toString()` round-trip + `.unsigned === true`. That's a legit truncation-catching proof. One uncovered hop (CodeRabbit's nit, non-slop): `getQbtcAccountInfo` -> `getQbtcChainSpecific` glue isn't exercised end-to-end (the large-value test injects into the schema directly). It's a 3-line passthrough, low risk - a glue test would close it.

## One real residual, filed so it doesn't get lost under the "Cosmos" framing: #1307
`getCosmosAccountInfo.ts` (untouched here) still does `Number(base.sequence)` on the LCD fallback for every non-QBTC chain - the exact same double-narrow pattern, just on `sequence`. Practical risk is low (a mainline nonce would need ~2^53 txs), but it's the same bug class and shouldn't ride quietly under this title. Tracked in #1307 as a separate fast-follow (don't pile it onto this PR).

## Before I put a four-eyes approve-stamp on it
This is the signing path / fund-safety-adjacent, so my bar is: (1) rebase off main - it's `BEHIND` right now, and (2) Unit Tests finish green (still pending at review time). Once it's rebased + green I'll run a quick build + `file://` CLI signing sanity on a real large-account-number Cosmos sign and stamp it. Not blocking you on anything but the rebase - the code itself is correct.

gomesalexandre (approved):
lgtm 🫡

Lands a P0 from the signing/broadcast arch audit (app#1966): the SDK-canonical cosmos uint64 narrowing at signingInputs/resolvers/cosmos/index.ts + getQbtcAccountInfo - QBTC is BROKEN on main today (assigns near-full-uint64 ids, so Number()-narrowing -> code-4 rejection -> chain unusable), and any cosmos account with id>=2^32 fails to sign on the CLI/extension surface. Fail-closed (no fund loss) but a hard availability regression. This PR is the fix.

Build-verified earlier (from-dist run, real wallet-core WASM): 2^64-1 account_number + 2^53+1 sequence round-trip EXACT through the compiled artifact, both Long.unsigned=true. The `Long.fromString(x.toString(), true)` change matches the sibling buildSimulateTxBytes and survives compilation.

Merging BEHIND-by-#1308 with admin: #1308 was chain-reference resolution (zero interaction with cosmos signing-inputs), and this is build-verified. The remaining non-QBTC sequence-narrowing + an assertUint64Exact invariant are tracked as #1307 / the audit P0 follow-up.

Issue #1307, opened 2026-07-16
cosmos signing: sequence still narrows through JS number in getCosmosAccountInfo (same double-narrow bug #1306 fixes for QBTC account_number)


gomesalexandre:
Surfaced during review of #1306 (which fixes the uint64 precision loss for QBTC's `account_number`). The **same double-narrow-then-rewiden anti-pattern survives for `sequence` on every non-QBTC Cosmos chain** (ATOM/OSMO/THOR/MAYA/Terra/TerraClassic/Kujira/Noble/Dydx/Akash), because `getCosmosAccountInfo.ts` is untouched by #1306.

## The bug
`getCosmosAccountInfo.ts` does `Number(base.sequence ?? '0')` on the LCD fallback path. The file's own comment notes upstream cosmjs 0.39 widened `Account.accountNumber` to `bigint` but **kept `Account.sequence: number`** - so `sequence` narrows through a JS double, then the chainSpecific resolver does `BigInt(sequence)` which re-widens the already-lossy value (does NOT recover precision), exactly the pattern #1306 fixed for QBTC.

- narrow: https://github.com/vultisig/vultisig-sdk/blob/main/packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.ts#L41-L52
- re-widen: https://github.com/vultisig/vultisig-sdk/blob/main/packages/core/mpc/keysign/chainSpecific/resolvers/cosmos/index.ts (BigInt(sequence))

## Severity
Fund-safety-adjacent (a wrong `sequence`/nonce in the SignDoc = an invalid or wrong-nonce signature), but **practical risk LOW**: a mainline Cosmos-SDK `sequence` (nonce) increments by 1 per tx and would need to reach ~2^53 (quadrillions of txs from one account) to lose precision. QBTC hit it on `account_number` because it apparently assigns near-full-uint64 IDs by design; mainline chains assign small sequential ones.

## Fix direction
Mirror #1306: parse `sequence` straight from the raw LCD string to `bigint` via the `parseUint64()` helper #1306 introduces (validate `/^\d+$/` + `<= maxUint64`), carry `bigint` end-to-end, drop the `Number()` cast. Add a >2^53 `sequence` golden-SignDoc test (the exact class where small-number tests pass while the real bug ships).

Do NOT pile this onto #1306 - separate PR against `getCosmosAccountInfo.ts` so the QBTC fix ships clean.

PR #1308, opened 2026-07-16
Share chain reference resolution across SDK, CLI, and MCP

### LINKED ISSUE #1253: architecture: expose one shared SDK chain-reference resolver and delete CLI/MCP mirrors
## Summary

`vultisig-sdk` already has the canonical pieces for chain-reference resolution:

- natural-language / alias normalization (`normalizeChain`)
- reverse mapping from decimal EVM IDs and Cosmos chain IDs (`skipChainIdToChainName`)

But first-party consumers still reimplement narrower local resolvers:

- MCP only does case-insensitive exact-match against `knownChains`
- CLI still ships a local alias map and a stale numeric EVM chain-id table

So users can hit “SDK supports it, consumer rejects it” drift on valid chain references like `Terra Classic`, `columbus-5`, or newer EVM chains.

## Evidence

SDK canonical helpers:

- natural-language alias / separator-insensitive normalization:
  - `packages/sdk/src/utils/normalizeChain.ts:69-147`
- reverse chain-id mapping:
  - `packages/sdk/src/tools/swap/skip/chainMapping.ts:24-39`
- current public export:
  - `packages/sdk/src/index.ts:790-791`

Current consumer-local mirrors:

- MCP exact-match-only resolver:
  - `clients/mcp/src/tools.ts:33-42`
- CLI local alias map:
  - `clients/cli/src/agent/executor.ts:1891-1935`
- CLI local numeric EVM chain-id map:
  - `clients/cli/src/agent/executor.ts:2191-2207`

Live drift proof from the current checkout:

```text
sdk normalize Terra Classic: TerraClassic
sdk normalize Bitcoin Cash: Bitcoin-Cash
sdk skip 8453: Base
sdk skip columbus-5: TerraClassic
mcp resolve Terra Classic: ERROR Unknown chain "Terra Classic". Use supported_chains or vault_info to see available chains.
cli resolve Terra Classic: null
cli resolve columbus-5: null
cli resolveChainId 999: null
cli resolveChainId 5000: null
cli resolveChainId 1329: null
```

## Why this matters

This is the campaign's core architecture-drift bug class in miniature:

- the SDK already owns the canonical logic,
- first-party consumers still mirror it anyway,
- and those mirrors are already stale.

That means chain-support growth is not actually centralized: every consumer needs manual follow-up to accept the same aliases / chain IDs the SDK already understands.

## Expected behavior

There should be one root-level SDK helper for generic chain-reference resolution that first-party consumers import instead of reimplementing:

- natural-language aliases (`Terra Classic`, `Bitcoin Cash`, `THOR Chain`)
- chain IDs (`8453`, `999`, `5000`, `1329`, `columbus-5`)
- optional narrowing against the consumer's allowed / known chain set

## Fix sketch

- Add an additive pure root-level helper in `@vultisig/sdk` for chain-reference resolution.
- Have CLI and MCP consume it.
- Add SDK + consumer regression tests for at least:
  - `Terra Classic`
  - `Bitcoin Cash`
  - `columbus-5`
  - `8453` → `Base`
  - `999` → `Hyperliquid`
  - `5000` → `Mantle`
  - `1329` → `Sei`

## Effort

Small-to-medium. Pure/public-surface work with good existing primitives.

## Report

Round 23 report mirror: https://gist.github.com/gomesalexandre/e989feda5e396b5b345aa64f4501f12f

rcoderdev:
Closes #1253
Adds one public SDK resolver for canonical names, aliases, Cosmos chain IDs, and decimal EVM chain IDs, then makes the CLI and MCP delegate to it while preserving their unknown-chain contracts. Runtime proof on the exact head exercised the built SDK public package plus the first-party CLI and MCP consumers, including allowed-set narrowing and fail-closed unknown input.

## Summary by CodeRabbit

- **New Features**
  - Added a shared `resolveChainReference` / `resolveChainIdReference` helper to the SDK for resolving chain names, aliases, and numeric IDs, with optional allowlist filtering.
  - Exposed the resolver via standard and React Native SDK exports, and updated CLI/MCP and swap tooling to use it for consistent behavior.

- **Bug Fixes**
  - Improved resolution behavior for unknown, invalid, and unsupported chain inputs by returning `undefined`/`null` instead of throwing.

- **Tests**
  - Added/expanded unit and CLI/MCP tests covering aliases, IDs, filtering, and invalid inputs.

- **Chores**
  - Updated changesets for SDK/CLI/MCP version bumps.

### DIFF
--- a/.changeset/shared-chain-reference-resolver.md
+++ b/.changeset/shared-chain-reference-resolver.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/sdk': minor
+'@vultisig/cli': patch
+'@vultisig/mcp': patch
+---
+
+Expose one SDK chain-reference resolver and use it for CLI and MCP chain names, aliases, and IDs.
--- a/clients/cli/src/agent/__tests__/chainReferenceResolver.test.ts
+++ b/clients/cli/src/agent/__tests__/chainReferenceResolver.test.ts
@@ -1,0 +1,28 @@
+import { Chain } from '@vultisig/sdk'
+import { describe, expect, it } from 'vitest'
+
+import { resolveChain, resolveChainId } from '../executor'
+
+describe('CLI chain reference resolution', () => {
+  it.each([
+    ['Terra Classic', Chain.TerraClassic],
+    ['Bitcoin Cash', Chain.BitcoinCash],
+    ['columbus-5', Chain.TerraClassic],
+  ])('resolves name reference %s through the SDK', (input, expected) => {
+    expect(resolveChain(input)).toBe(expected)
+  })
+
+  it.each([
+    [8453, Chain.Base],
+    ['999', Chain.Hyperliquid],
+    ['5000', Chain.Mantle],
+    ['1329', Chain.Sei],
+  ])('resolves ID reference %s through the SDK', (input, expected) => {
+    expect(resolveChainId(input)).toBe(expected)
+  })
+
+  it('preserves null for unresolved names and IDs', () => {
+    expect(resolveChain('not-a-chain')).toBeNull()
+    expect(resolveChainId('not-an-id')).toBeNull()
+  })
+})
--- a/clients/cli/src/agent/executor.ts
+++ b/clients/cli/src/agent/executor.ts
@@ -7,7 +7,15 @@
  * to be flushed into the next outbound `context.recent_actions`.
  */
 import type { VaultBase, Vultisig } from '@vultisig/sdk'
-import { Chain, chainFeeCoin, getChainKind, VaultError, VaultErrorCode, Vultisig as VultisigSdk } from '@vultisig/sdk'
+import {
+  Chain,
+  chainFeeCoin,
+  getChainKind,
+  resolveChainReference,
+  VaultError,
+  VaultErrorCode,
+  Vultisig as VultisigSdk,
+} from '@vultisig/sdk'
 import { formatUnits, hashTypedData, recoverAddress } from 'viem'
 
 import { VaultStateStore } from '../core/VaultStateStore'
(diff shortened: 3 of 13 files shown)


gomesalexandre (changes requested):
nice cleanup on centralizing chain resolution. i diffed the exact current head against fresh origin/main and the code-level change looks sound, but i can't approve this yet.

blocking:
- receipts are still short for this class of change. this exports new public sdk surface and changes user-visible cli/mcp chain-resolution behavior, so the repo policy needs a built cli runtime receipt against the real cli/sdk surface, not just prose saying runtime proof happened. please paste the concrete invocation plus stdout excerpt.
- the coderabbit pre-merge warning still shows docstring coverage at 60% vs the configured 80% threshold. if that warning is stale, please point to the exact receipt/waiver. otherwise it still needs to be resolved before merge.

for context, i did not find a correctness regression in the 13-file diff itself, and the THOR Chain follow-up appears addressed on the current head.

gomesalexandre (approved):
lgtm 🫡 - superseding my earlier CHANGES_REQUESTED: the runtime-receipt block is satisfied.

Code side (traced on the diff): centralizes chain resolution into a shared `resolveChainReference` built on the untouched `normalizeChain.ts`, fans CLI + MCP + the Skip fund-path onto it. The one fund-critical path (skipChainIdToChainName) is a byte-identical extraction (+ a safer Number.isSafeInteger guard); all 18 CLI tickers trace to identical resolution; `resolveChainId` actually FIXES a latent bug (old 10-entry map silently signed unrecognized-but-valid EVM chains against the default) by resolving through the real evmChainId map; MCP resolveChain stays hard-gated by allowedChains.includes (fail-closed). Closes the SDK-canon arch-drift finding #1253.

Runtime receipt (the block): built-CLI run against the BUILT artifacts (not src/vitest) - 15/15 pass, including the latent-bug fixes (8453->Base, 999->Hyperliquid, 5000->Mantle, 1329->Sei), aliases (Terra Classic, Bitcoin Cash, columbus-5), strict rejection (137abc->undefined), and the allowlist fail-closed proven through a real MCP get_address tool-call boundary (999 outside the vault's chains -> isError "Unknown chain"). CLI binary boots (vultisig/2.19.18). Receipt: the comment above.

Docstring coverage 60%<80% is a soft CodeRabbit warning (the CodeRabbit check itself passes) - non-blocking. 5/5 CI green.

The non-QBTC sequence-narrowing follow-up stays tracked separately in #1307.

PR #1309, opened 2026-07-16
Scope clearVaults to SDK-owned keys

### LINKED ISSUE #1208: fix(storage): keep clearVaults scoped to SDK-owned keys
## Problem

`clearVaults()` clears the entire configured storage adapter instead of only SDK-owned vault keys. In an embedded host using shared storage, this can delete unrelated application data.

Evidence: `packages/sdk/src/storage/VaultStorage.ts:198-204`.

## Acceptance criteria

- Delete only keys in the SDK-owned namespace/index.
- Do not call adapter-wide clear from the vault API.
- Document namespace ownership for custom adapters.
- Add a regression test with unrelated host keys present.

rcoderdev:
Closes #1208

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `clearVaults()` to remove the full set of SDK-owned vault-related keys (including pending/cache and push registration keys) when using shared/custom storage, while preserving unrelated data.
  * `clearVaults()` no longer performs adapter-wide clearing and now also clears in-memory pending vault state.
  * Improved async cleanup behavior and surfaced adapter removal failures.
* **Documentation**
  * Updated the SDK Storage guide with clarified scoping, reserved namespaces, and guidance on `clearVaults()` vs adapter `clear()`.
* **Tests**
  * Added unit coverage for shared storage behavior, async completion, and adapter error handling.
* **Chores**
  * Added a patch changeset for the SDK.

### DIFF
--- a/.changeset/clear-vaults-sdk-keys.md
+++ b/.changeset/clear-vaults-sdk-keys.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Keep `clearVaults` scoped to SDK-owned keys in shared storage adapters.
--- a/docs/SDK-USERS-GUIDE.md
+++ b/docs/SDK-USERS-GUIDE.md
@@ -2315,10 +2315,11 @@ sdk.dispose()
 Custom adapters may share a backing store with application data. The SDK
 reserves `vault:*`, `pending:*`, `cache:*`, `addressBook:*`, `activeVaultId`,
 `pushNotificationRegistrations`, `config:defaultCurrency`, and
-`config:defaultChains`. `sdk.clearVaults()` removes the vault-scoped keys and
-local push registrations but intentionally retains the two SDK preference
-keys. The adapter's `clear()` method remains an explicit adapter-wide operation
-that may also remove unrelated host keys.
+`config:defaultChains`. `sdk.clearVaults()` removes the vault-scoped keys,
+saved and vault address-book entries, and local push registrations but
+intentionally retains the two SDK preference keys. The adapter's `clear()`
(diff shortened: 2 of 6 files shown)


gomesalexandre (approved):
lgtm 🫡

Scopes clearVaults to SDK-owned keys (drops the full storage.clear() adapter-nuke) - crown-jewel-deletion code, so reviewed hard on both the SDK side and the consumer contract:

Fund-safety (verified by enumerating every storage.set): ALL key material (vault:/pending: = the keyshare blobs) is cleared, plus cache:/pushNotificationRegistrations/activeVaultId/addressBook - and there is NO key material stored outside vault:/pending:, so no orphan/ghost vault is possible. Foreign/host keys are correctly spared (the whole point). Ownership = the existing stable key prefix, no new marker - so the migration risk (old keys missing a marker -> stale-key leak) is ABSENT: pre-existing vaults were always written under vault:${id}. Bonus: fixes an unawaited addressBookManager.clear() floating promise.

Consumer contract (the one thing unit tests can't cover) - checked myself: the app uses AsyncStorage DIRECTLY for its own keys and has its own clearVaultConversationCache; grep found NO app call to the SDK clearVaults at all -> the app never triggered the old full-nuke, so this scoped change can't regress it. CodeRabbit findings were stale (addressed on newer head).

6/6 CI green incl the new fund-safety unit tests (host key survives / all SDK keys nulled / config:* survives / async removal awaited / adapter errors propagate). Nit for a follow-up: stale doc types/index.ts:449 says cache is at vault:{id}:cache but CacheService writes cache:${id}:* (both are cleared, so harmless).

PR #1311, opened 2026-07-16
fix(sdk): export documented decode namespace

### LINKED ISSUE #1310: public API docs promise sdk.decode.fromToolResult, but @vultisig/sdk never exports a decode namespace
## Summary
The root `@vultisig/sdk` surface documents and brands the canonical decoder as `sdk.decode.fromToolResult`, but the published package only exposes flat named exports (`decodeFromToolResult`, `decodeCosmosTx`, `decodeEvmTx`). Consumers following the SDK's own docs/comments have to hand-roll a wrapper object instead of importing the canonical namespace directly.

## Evidence
- `packages/sdk/src/tools/policy/types.ts:8` and `packages/sdk/src/tools/policy/types.ts:106` describe the canonical decoder as `sdk.decode.fromToolResult`.
- `packages/sdk/src/tools/decode/fromToolResult.ts:98-126` frames that decoder as the canonical bytes oracle.
- `packages/sdk/src/tools/decode/index.ts` only exported flat helpers before the fix; there was no `decode` namespace object.
- `packages/sdk/src/tools/index.ts:40-42` and `packages/sdk/src/index.ts:675-678` re-exported only the flat helpers, so `import * as sdk from "@vultisig/sdk"; sdk.decode?.fromToolResult` was `undefined`.

## Impact
- The SDK's public docs and runtime surface disagree.
- First-party/downstream consumers are pushed toward local wrapper objects or non-standard imports, which is exactly the drift pattern this campaign is trying to eliminate.
- The canonical decode API becomes harder to use consistently across app/backend/CLI code.

## Why SDK should own the fix
This is a pure public-surface/API-contract bug inside `vultisig-sdk`. The SDK is the canonical owner of the decode namespace it already documents.

## Fix sketch
- Export a real `decode` namespace object from `packages/sdk/src/tools/decode/index.ts` with `fromToolResult`, `decodeCosmosTx`, and `decodeEvmTx`.
- Re-export that object through `packages/sdk/src/tools/index.ts` and `packages/sdk/src/index.ts`.
- Add a root public-export regression test proving `sdk.decode.fromToolResult` exists and aliases the flat helper.

## Testability
Yes — pure SDK/CLI-facing export-contract tests plus a built-dist runtime import smoke.

## Report
Round-30 report mirror: https://gist.github.com/gomesalexandre/48c3a2f477156ba34b2330d88c451594
Campaign index artifact (canonical): https://claude.ai/code/artifact/be2deadf-6195-4dfa-a910-b269b9af7a84

gomesalexandre:
## Summary
- export the documented `sdk.decode` namespace from `@vultisig/sdk`
- keep `sdk.decode.fromToolResult` aligned with the existing flat decode helpers
- add a root public-export regression test and changeset

Closes #1310

## Why
`packages/sdk/src/tools/policy/types.ts` and the decode docs already tell consumers to call `sdk.decode.fromToolResult(...)`, but the published root package only exposed flat helpers. That forced downstream code to invent a local wrapper object instead of importing the canonical SDK surface directly.

## Changes
- add `decode` object in `packages/sdk/src/tools/decode/index.ts`
- re-export `decode` through `packages/sdk/src/tools/index.ts` and `packages/sdk/src/index.ts`
- extend `packages/sdk/tests/unit/utils/publicExports.test.ts` to prove the documented namespace exists and aliases the flat helpers
- add `.changeset/r30-decode-namespace.md`

## Receipts
- `yarn build:shared`
- `yarn build:sdk && yarn cli:build`
- `node --input-type=module -e "import * as sdk from './packages/sdk/dist/index.node.esm.js'; ..."` → verified `sdk.decode.fromToolResult` exists and aliases `sdk.decodeFromToolResult`

### Runtime proof
```json
{
  "hasDecode": true,
  "hasFromToolResult": true,
  "aliasMatches": true,
  "hasDecodeCosmosTx": true,
  "hasDecodeEvmTx": true
}
```

## Risk
Low — additive public-surface parity only; no tx-building, signing, or broadcast logic changed.

## Report
Round-30 report mirror: https://gist.github.com/gomesalexandre/48c3a2f477156ba34b2330d88c451594
Campaign index artifact (canonical): https://claude.ai/code/artifact/be2deadf-6195-4dfa-a910-b269b9af7a84


packages/sdk/src/index.ts
```diff
@@ -683,6 +683,7 @@ export {
   COSMOS_SWAP_GAS_LIMIT,
   cosmosBalanceChains,
   cosmosStaking,
+  decode,
```
NeOMakinG, line 686: `should-fix:` this makes `decode` reachable on the **node** condition only. `packages/sdk/package.json:17,23,42-46` resolves `react-native` to `dist/index.react-native.*`, built from `src/platforms/react-native/index.ts` — a hand-curated allow-list, not a re-export of this barrel. I probed both conditions at HEAD:

```
ROOT decode              = object
RN   decode              = undefined
RN   decodeFromToolResult= undefined
RN   dex (namespace ctl) = object
RN   defi(namespace ctl) = object
```

`dex` and `defi` are both there; `decode` is the outlier. The comment above the parity-guard block in `tests/unit/platforms/react-native/entry.test.ts` warns about exactly this and names `publicExports.test.ts` — the only test this PR extends — as unable to see the gap.

Fix: add `decode` to the RN entry's export list and add an `entry.test.ts` case that resolves the RN entry FILE directly. Check `Buffer` while you're there — `fromToolResult.ts:52,58` call `Buffer.from`; the RN entry polyfills `globalThis.Buffer` at module load (`index.ts:14-16`), so it should work, but assert it rather than assume it.

packages/sdk/src/tools/decode/index.ts
```diff
@@ -1,4 +1,16 @@
...
+export { decodeCosmosTx, decodeEvmTx, decodeFromToolResult }
 export type { AssetRef, ChainFamily, DecodeFromToolResultInput, Envelope, EnvelopeKind } from './types'
+
+/**
+ * Documented decode namespace — callers can use `sdk.decode.fromToolResult(...)`
+ * without hand-rolling their own wrapper object around the flat helpers.
+ */
+export const decode = {
+  fromToolResult: decodeFromToolResult,
+  decodeCosmosTx,
```
NeOMakinG, line 14: `should-fix:` naming stutters — `sdk.decode.decodeCosmosTx` / `sdk.decode.decodeEvmTx` next to the unprefixed `decode.fromToolResult`.

Compare `src/tools/defi/index.ts:29-40`, the house pattern this otherwise copies correctly: every member unprefixed (`arkis`, `balancer`, `glif`, …), plus `export type Defi = typeof defi` so consumers can type a variable holding the namespace — this PR has no `Decode` equivalent.

Only `fromToolResult` is actually documented (`src/tools/policy/types.ts:8` and `:106`); both stuttering members are exported-but-undocumented, so nothing forces these spellings. Suggest `cosmosTx`/`evmTx` + `export type Decode = typeof decode`. Flagging now because draft is the last cheap moment — after publish both spellings are public API forever and a rename is breaking.

packages/sdk/src/tools/decode/index.ts
```diff
@@ -1,4 +1,16 @@
...
-export { decodeFromToolResult } from './fromToolResult'
+import { decodeCosmosTx } from './cosmos'
+import { decodeEvmTx } from './evm'
+import { decodeFromToolResult } from './fromToolResult'
+
+export { decodeCosmosTx, decodeEvmTx, decodeFromToolResult }
 export type { AssetRef, ChainFamily, DecodeFromToolResultInput, Envelope, EnvelopeKind } from './types'
+
+/**
+ * Documented decode namespace — callers can use `sdk.decode.fromToolResult(...)`
```
NeOMakinG, line 9: `q:` worth restating the `decoded` contract here? This namespace becomes the discoverable entry point to the bytes oracle, and `decodeFromToolResult` deliberately **never throws** — failures surface as `Envelope{ decoded: false, decodeError }` (`fromToolResult.ts:106-107`). `decode/types.ts:41-43` is emphatic that "callers MUST check \`decoded\` before trusting any other field", which is a fail-open shape for anyone who skips it on a safety surface. This JSDoc covers the ergonomics but not the contract; one line would carry the warning to where people will now look. No behaviour change requested — the semantics are right.

.changeset/r30-decode-namespace.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vultisig/sdk': patch
```
NeOMakinG, line 2: `suggestion:` `patch` understates a public API widening — this adds a new root export. The repo's own precedent for this exact category is `minor`: `.changeset/eip191-auth-signing-helpers.md` ("Expose canonical EIP-191 hashing and ECDSA 65-byte signature formatting helpers") is `'@vultisig/sdk': minor`.

Low practical impact — `agent-backend-ts` pins `^2.20.0` so it floats in either way — so this is changelog/semver hygiene, not risk. The dual `sdk`+`cli` listing is correct house style given `linked: [["@vultisig/sdk","@vultisig/cli"]]` in `.changeset/config.json`; `.changeset/canonical-decode-chain-family.md` does the same for a decode-only SDK change.

Also: no trailing newline on this file.

NeOMakinG (commented):
This one is genuinely what it says on the tin: additive export plumbing, correctly done. I went looking hard for the two shapes that burned sibling "plumbing" drafts in this same batch — a re-keyed public export silently breaking a downstream fail-closed guard (#1290), and a one-line parity fix arming a never-executed consumer branch (#1337) — and neither is present here. No blockers.

The one thing worth acting on before this leaves draft is that it fixes the docs/runtime disagreement on the **node** condition only. On the condition Metro actually resolves, `sdk.decode` is still `undefined`.

## testing approach

Worktree `/tmp/wt-sdk-1311` at HEAD `1a8e9035`, root+`packages/core`+`packages/sdk` `node_modules` and the built `packages/**/dist` dirs symlinked from the canonical checkout (the `dist` link is required — without it the suite dies on `@vultisig/core-config` before collecting a single test).

- **CI is real here, not the zero-job fake-green.** `gh api .../commits/1a8e9035/check-runs --jq .total_count` → **4** (Unit Tests 20m45s, Integration 5m19s, Lint+TypeCheck 14m23s, Summary). Contrast #1290 in this same batch, where the tick said pass with `total_count: 0`. The new assertion is on the CI path.
- `vitest run --config tests/unit/vitest.config.ts tests/unit/utils/publicExports.test.ts` → **18 passed**.
- **Mutation-proved the new test.** Deleted the single `decode,` line from `packages/sdk/src/index.ts:686`, re-ran: `1 failed | 17 passed`, `expected undefined to be defined` at `publicExports.test.ts:31`. The test is load-bearing, not decorative. Restored after.
- **Reachability probe per export condition** (throwaway vitest file inside the package, with the RN entry's native mocks, since `tsx` can't resolve workspace aliases from a worktree). Verbatim:
  ```
  ROOT decode              = object
  ROOT decodeFromToolResult= function
  ROOT decodeEvmTx         = function
  RN   decode              = undefined
  RN   decodeFromToolResult= undefined
  RN   decodeEvmTx         = undefined
  RN   decodeCosmosTx      = undefined
  RN   dex (namespace ctl) = object
  RN   defi(namespace ctl) = object
  ```
- Collision sweep for a standalone `decode` identifier across `src/index.ts`, `src/tools/index.ts`, `src/platforms/react-native/index.ts` and `@vultisig/core-chain` — exactly one definition (`src/tools/decode/index.ts:12`) and one re-export per level. **Nothing existing is shadowed, renamed, or changed.** Every pre-existing flat helper still resolves to the same function object; the test pins that with `toBe` identity, not structural equality.
- Second lane (Codex, `gpt-5.3-codex`) run in parallel on packaging/tree-shaking/import-time side effects. It converged independently on the RN gap and cleared the bundle question — see `## risk`.

## should-fix

**1. The documented namespace is still unreachable on the RN condition — and the repo already has a written warning about exactly this.**

`packages/sdk/package.json:17,23,42-46` resolves `react-native` to `dist/index.react-native.*`, which is built from `src/platforms/react-native/index.ts` — a hand-curated allow-list, not a re-export of the root barrel. This PR widens `src/index.ts` and `src/tools/index.ts` only, so the RN surface is untouched. Probe output above: `RN decode = undefined`, while the sibling namespaces `dex` and `defi` **are** both reachable there. `decode` is the outlier.

The repo says this itself, in the comment above the parity-guard block in `packages/sdk/tests/unit/platforms/react-native/entry.test.ts`:

> adding something to the root does NOT make it reachable from the app (Metro resolves the react-native condition to this file, never the node one). `publicExports.test.ts` only resolves the node condition and can't see a gap here

`publicExports.test.ts` is the only test this PR extends. There is prior art for the fix in `.changeset/exact-balance-formatting.md`, which called out shipping "including the React Native platform entry (the one Metro actually resolves for Station), not just the node/default condition."

Fix: add `decode` (and, if you want the flat helpers usable too, `decodeCosmosTx`/`decodeEvmTx`/`decodeFromToolResult`) to the RN entry's export list, plus an `entry.test.ts` case resolving the RN entry FILE directly. Worth confirming `Buffer` first — `fromToolResult.ts:52,58` call `Buffer.from`; the RN entry polyfills `globalThis.Buffer` at module load (`index.ts:14-16`) specifically to guarantee ordering, so this should be fine, but it deserves an actual assertion rather than an assumption.

**Calibrating this honestly:** it is a parity/DX gap, not a fund-safety one. No in-repo RN consumer is currently blocked. `vultiagent-app`'s `decodeEvmTx` at `src/features/agent/lib/txProposal/yield.ts:120` is a name collision, **not** a hand-rolled copy of this decoder — it's a JSON field-picker over a yield.xyz `unsignedTransaction` blob, not an RLP/proto3 bytes decoder. So nothing regresses. But the issue's own framing ("the app decoded-intent card" is listed in `decode/types.ts` as a canonical Envelope reader) points at the RN app, and that's the one consumer this doesn't reach.

**2. `decode.decodeCosmosTx` stutters, and diverges from the sibling namespace idiom.**

The namespace mixes one unprefixed member with two prefixed ones: `decode.fromToolResult` but `decode.decodeCosmosTx` / `decode.decodeEvmTx`. Compare `src/tools/defi/index.ts:29-40`, the established house pattern this correctly copies otherwise — all members unprefixed (`arkis`, `balancer`, `glif`, …) — and it also ships `export type Defi = typeof defi`, which this PR has no equivalent of, so consumers can't type a variable holding the namespace.

Only `fromToolResult` is actually documented (`src/tools/policy/types.ts:8` and `:106`); both stuttering members are exported-but-undocumented, so nothing forces those spellings. Suggest `cosmosTx`/`evmTx` (or `fromCosmosBytes`/`fromEvmBytes`) plus `export type Decode = typeof decode`. Raising it now because draft is the last cheap moment — once published, both spellings are public API forever and renaming is a breaking change.

## suggestion

**3. `patch` understates a public API widening.** `.changeset/r30-decode-namespace.md` marks this `patch`, but it adds a new root export. The repo's own precedent for this exact category is `minor`: `.changeset/eip191-auth-signing-helpers.md` — "Expose canonical EIP-191 hashing and ECDSA 65-byte signature formatting helpers" — is `'@vultisig/sdk': minor`. Low practical impact (`agent-backend-ts` pins `^2.20.0`, so it floats in either way), so this is changelog/semver hygiene rather than risk. Dual `sdk`+`cli` listing is fine and matches house style given `linked: [["@vultisig/sdk","@vultisig/cli"]]` in `.changeset/config.json` — e.g. `.changeset/canonical-decode-chain-family.md` does the same for a decode-only SDK change. Also: the file has no trailing newline.

## q

**4. Should the namespace JSDoc restate the `decoded: false` contract?** You asked (implicitly, by shipping this) for `decode` to become the discoverable entry point to the bytes oracle. `decodeFromToolResult` deliberately **never throws** — every failure surfaces as `Envelope{ decoded: false, decodeError }` (`fromToolResult.ts:106-107`), and `decode/types.ts:41-43` is emphatic: "callers MUST check `decoded` before trusting any other field." That's a fail-open shape for anyone who skips the check on a safety surface. The new JSDoc at `decode/index.ts:8-11` explains the ergonomics but not the contract. One line pointing at `decoded` would carry the warning to the place people will now actually look. Not asking for a behaviour change — the semantics are right, and the strict/lenient split is a non-issue here since all three members share the same fail-closed `Envelope` return.

## risk

**Low, and I checked the specific ways it could have been otherwise.**

No new import-time side effect enters any bundle. The root entry already exported all three flat helpers (`src/tools/index.ts:42`, `src/index.ts:689` pre-change), so `decode/evm.ts` + `decode/cosmos.ts` were already in the module graph; adding a const object referencing them pulls in nothing new. Codex independently audited this and reached the same conclusion: the only module-eval-time work in the newly-namespaced code is the pure `parseAbi(...)` constant at `decode/evm.ts:28`, already present in the flat-export graph — no top-level await, no env read, no Node-only call at import time, no WASM. Explicitly not the `abts#1654` shape (a call-time→import-time move that took a suite from 7 passed to 0 collected with tsc and lint clean). Tree-shaking is unaffected for consumers who don't reference `decode`: it's a side-effect-free object literal at the top level. Lint+TypeCheck went green on a real 14m job.

No caret-bump hazard — this PR touches no dependency versions.

## verdict

**COMMENTED** — draft, so not approving on principle regardless of content. Nothing here blocks; findings 1 and 2 are both cheapest to fix while it's still a draft, and 1 is the one I'd actually want before this merges, because "export the documented namespace" is only half-true until the RN entry resolves it too. Nice work on the identity assertions (`toBe`, not `toEqual`) and on shipping the mutation-provable test rather than a tautology.

---
🤖 posted by the reviewer agent · 2 lanes (Claude + Codex `gpt-5.3-codex`) · findings deduped at synthesis

PR #1313, opened 2026-07-16
fix(sdk): clear pending vault after verification

### LINKED ISSUE #1210: fix(vault): clear pending record after same-process verification
## Problem

After successful same-process vault verification, the pending vault record remains stored even though the final vault is available. This leaves stale sensitive material and can re-surface duplicate recovery state.

Evidence:
- `packages/sdk/src/services/VaultCreationService.ts:303-342`
- `packages/sdk/src/storage/PendingVaultStorage.ts:69-107`

## Acceptance criteria

- Remove the pending record only after final vault persistence succeeds.
- Keep it when final persistence or verification fails.
- Make cleanup idempotent.
- Test success, persistence failure, retry, and process restart.

rcoderdev:
Closes #1210
Actual public SDK verification over shared storage retained the pending record after a controlled final-persistence failure, then removed it after same-process retry and after reconstruction by a fresh SDK instance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pending vault data is now cleared from storage after successful vault verification, including when verification occurs in the same process.
  * Pending data is retained if final vault persistence fails and is removed after a successful retry.
  * Cleanup remains safe and consistent when pending vault data is restored in a new process.


gomesalexandre (approved):
lgtm 🫡

Remote-only pass via gh metadata/diff, no local checkout.

Confirmed the real bug: `createFastVault({ persistPending: true })` persists `pending:${vaultId}` to storage and also keeps the vault in the in-memory `pendingVaults` map. The old `verifyVault()` only removed the persisted pending record on the `fromDisk` path, so same-process verification could succeed while leaving stale pending state on disk. Your unconditional cleanup after the final `save()` / `setActiveVault()` is the right fix.

Important ordering check also looks right: cleanup happens after the final persistence succeeds, so if that final save throws the pending record survives for retry instead of being lost mid-flight.

I also sanity-checked the unconditional `storage.remove(...)` against the SDK storage backends. This is safe even when no pending record exists:
- memory: delete no-op
- node/electron file storage: ENOENT-safe remove
- browser / RN / chrome extension: missing-key remove is fine

Tests map cleanly to the acceptance criteria:
- same-process verify clears the persisted pending record
- failed final persistence keeps pending state for retry
- fresh-process verify still clears the pending record idempotently

Patch changeset is appropriate.

gomesalexandre (approved):
lgtm 🫡

Fixes #1210 (in-memory verify of a persistPending vault left the pending: keyshare blob on disk forever). Reviewed hard on the destructive ordering since pending: holds key material:

Persist-before-clear CORRECT: verifyVault does save() -> setActiveVault() -> pendingVaults.delete() -> storage.remove('pending:'). save() writes the FULL keyshare vaultData to vault:${id} and FileStorage.set is atomic (temp-file + rename), durable BEFORE the remove. No race deletes the pending keyshare before the permanent vault is confirmed persisted -> no unrecoverable-vault window. Failure paths throw pre-remove (verification fail pre-save, persist/setActiveVault fail pre-remove) -> pending retained for retry. Idempotent across all 5 storage backends. Orthogonal to #1309. Green, not draft.

PR #1314, opened 2026-07-16
fix(sdk): harden React Native EVM broadcast idempotency

### LINKED ISSUE #1249: fix: rn evm broadcast is a bare passthrough (false-failure -> double-send) while its jsdoc claims core-level hardening
Round-20 audit of the react-native platform tree. Framed as defensive reliability maintenance (public repo).

`packages/sdk/src/platforms/react-native/chains/evm/rpc.ts:118-125` - `broadcastEvmRawTx` is a bare passthrough: `return client.sendRawTransaction({...})`. Its own JSDoc (:111-117) claims it "Treats 'already known'-class errors as non-fatal ... by re-throwing a more descriptive error" - the code does NONE of that. It also cites a "RawBroadcastService" symbol that does not exist.

The hardened core resolver `packages/core/chain/tx/broadcast/resolvers/evm.ts:10-28` is the correct reference: it swallows `already known` / `transaction already exists` / `tx already in mempool` as idempotent success, and routes every OTHER error through `verifyBroadcastByHash`. The RN reimplementation inherits none of it.

## the failure

On a broadcast race (MPC retry, double-tap, or an RPC returning "already known" because the tx is already in the mempool), the RN path throws -> the send/swap surfaces as FAILED while the tx is actually on-chain. User re-initiates -> new tx at nonce+1 (getEvmNonce prefers 'pending', :45-52) -> TWO distinct transfers of the same funds. This is the false-failure inverse of the #1119 broadcast-success-on-failure class, specific to the RN-local EVM reimplementation (distinct from #1227 which is RN Solana).

## OPEN QUESTION (reachability - answer before sizing the fix)

Confirm the Station app actually broadcasts EVM txs through this sdk function vs its own app-side services/evmTx.ts. If the app path is used, this is latent in the sdk but not yet biting app users; if the sdk path is live, it's an active double-send risk. Either way the sdk function should not lie in its JSDoc and should match core.

## fix shape

Wrap in `attempt`, return on the known-duplicate set, route other errors through verifyBroadcastByHash - i.e. make it behave like the core resolver (or better, delete the RN reimplementation and delegate to core, feeding the #1230 consolidation). Delete the false JSDoc regardless.

effort: S / risk: LOW (fix is well-templated by core + the just-landed ripple hardening) / confidence: HIGH on the divergence, MED on live impact pending reachability

found by improve round 20 (2026-07-13)

rcoderdev:
Closes #1249

### DIFF
--- a/.changeset/rn-evm-broadcast-idempotency.md
+++ b/.changeset/rn-evm-broadcast-idempotency.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Make React Native EVM raw broadcast idempotent for known or hash-confirmed transactions.
--- a/packages/sdk/src/platforms/react-native/chains/evm/rpc.ts
+++ b/packages/sdk/src/platforms/react-native/chains/evm/rpc.ts
@@ -15,8 +15,10 @@
  */
 
 import { EvmChain } from '@vultisig/core-chain/Chain'
+import { attempt } from '@vultisig/lib-utils/attempt'
+import { isInError } from '@vultisig/lib-utils/error/isInError'
 import { memoize } from '@vultisig/lib-utils/memoize'
-import { createPublicClient, erc20Abi, http, type PublicClient } from 'viem'
+import { createPublicClient, erc20Abi, http, keccak256, type PublicClient } from 'viem'
 
 import { getEvmNumericChainId } from './tx'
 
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
lgtm 🫡

RN EVM broadcast idempotency hardening (Closes #1249) - reviewed deep, fund-safe + complementary to the signing work:
- `broadcastEvmRawTx` now returns deterministic `keccak256(rawTxHex)` instead of trusting the RPC-returned hash (a malicious/buggy RPC can't feed a poller a wrong hash). Byte-correct: viem's sendRawTransaction returns exactly keccak256(serializedTransaction) for all tx types, so the happy-path return is identical - no caller regression.
- On send error: idempotent-success only for 3 "already known"-class strings; every OTHER error fail-CLOSES - verifies via getTransactionReceipt then getTransaction, returns the hash ONLY if confirmed on-chain/mempool, else rethrows. Correct fail-closed direction.
- Nice: deliberately EXCLUDES "nonce too low" from the fast-path (ambiguous: this-tx-mined vs a-different-tx-took-the-nonce) - routing it through hash-matched confirmation is strictly safer than the sibling's blanket-ignore. Stale docstring dropped.
- Zero conflict with the EVM signing work (app#1971 RLP fix) - different repo/layer/bytes, strictly complementary; the SDK viem signing path is untouched. Does NOT touch tx serialization/nonce/chainId/recovered-sender.
- 6 unit tests cover every branch (deterministic-hash, 3 known-error cases, receipt-confirm, getTransaction-fallback, different-hash-reject, no-confirm-rethrow), 5/5 CI green.

q for a follow-up (non-blocking): the receipt branch matches on hash but ignores status, so a mined-but-reverted tx returns success - defensible for a broadcast fn (execution status is the poller's job), worth a one-line comment. The audit's pre-broadcast recovered-sender==vault P1 (app#1966) is separate + stays open - this doesn't address it.

Issue #1045, opened 2026-07-08
fix(broadcast): Cardano + UTXO resolvers skip verifyBroadcastByHash for BadInputsUTxO, reporting false success


gomesalexandre:
## What it says on the box

`broadcastCardanoTx` and `broadcastUtxoTx` both bypass the `verifyBroadcastByHash` safety net for a specific bucket of RPC errors, and unconditionally return `null` instead. Callers treat that `null` as "fine" and fall back to a locally-computed tx hash, so a genuine broadcast failure gets reported as success.

## Where

- `packages/core/chain/tx/broadcast/resolvers/cardano.ts:46-49`
- `packages/core/chain/tx/broadcast/resolvers/utxo.ts:44-46`

Identical pattern in both:

```ts
if (isInError(error, 'BadInputsUTxO', 'txn-mempool-conflict', 'already known')) {
  return null
}
```

Compare to the generic (non-bucketed) error path right below it in both files, which correctly runs the tx through `verifyBroadcastByHash` before giving up:

```ts
const broadcastError = new Error(`Failed to broadcast transaction: ${error}`)
await verifyBroadcastByHash({ chain, tx, error: broadcastError })
return null
```

## Why this is wrong

`verifyBroadcastByHash.ts` exists precisely to disambiguate this kind of ambiguous error - it re-derives the tx hash, checks real on-chain/mempool status, and only swallows the error if the tx is confirmed pending or successful; otherwise it re-throws. Its own doc comment explains the intended use case: an MPC keysign race where a slower device's redundant broadcast attempt gets an "already known"-style error while a peer's copy is genuinely on-chain.

`txn-mempool-conflict` and `already known` fit that intended case. **`BadInputsUTxO` does not** - it means the UTXO inputs referenced by the transaction don't exist or are already spent, which is a real, potentially unrelated-to-any-peer failure. Bucketing it with the other two and skipping `verifyBroadcastByHash` means: for this error, nobody ever confirms whether the tx actually landed on-chain before treating the submission as "fine."

## Blast radius (traced two callers)

[...]

Issue #936, opened 2026-07-02
fix(broadcast): RawBroadcastService reports already-submitted tx as BroadcastFailed (double-broadcast risk)


realpaaao:
## Evidence
`packages/sdk/src/vault/services/RawBroadcastService.ts` — the EVM (`146-165`), UTXO, Solana, Cosmos, TON, Ripple, Sui, Tron branches all detect "already submitted"/"already known"/"nonce too low" signatures and then `throw new VaultError(BroadcastFailed, ...)`. This contradicts the core resolver `packages/core/chain/tx/broadcast/resolvers/evm.ts:23-25`, which treats "already known" as **success** (`return`).

## Impact
A successful idempotent re-broadcast (retry after a dropped socket, mempool re-submit) surfaces as a hard failure. A caller that retries on `BroadcastFailed` can double-broadcast; a caller that reports it tells the user a landed tx failed. Inconsistent semantics between the raw and core paths for the identical condition.

## Fix sketch
For the "already submitted" branches, return the (locally derivable) tx hash as success instead of throwing, matching `broadcastEvmTx`.

---
_Found via `/repo-audit deep` @ `e11d55f`. Full audit: internal `Giga/vultisig/research/2026-07-02-vultisig-sdk-audit.md`._

PR #1316, opened 2026-07-16
fix(cosmos): broadcast false-success on DeliverTx code!=0

### LINKED ISSUE #1312: P0 cosmos broadcast false-success: DeliverTx code!=0 swallowed, vault.send/swap report success for a failed tx
From the signing/broadcast audit (app#1966), verified opus+codex.

`chain/tx/broadcast/resolvers/cosmos.ts:15-19` + `packages/sdk/src/vault/services/BroadcastService.ts:116-121`. cosmjs StargateClient.broadcastTx RESOLVES (does not throw) when a tx is included on-chain but FAILS execution (out-of-gas, wasm revert, THORChain/Maya deposit-handler failure). The resolver returns success on `!error` without reading `response.code`, so `vault.send()` / `vault.swap()` / the CLI report success with a real on-chain hash for a tx that never executed - gas burned, transfer/deposit/swap did NOT happen, and any agent consuming the result is poisoned.

Sibling of the sdk#1045/#1046/#1119 false-success class (the APP surface already avoids this via its TxVerifier registry; this is the SDK-canonical path).

## Fix
Read `DeliverTxResponse.code`; on code!=0 route through `verifyBroadcastByHash` (or throw). Make `verifyBroadcastByHash` MANDATORY inside the shared broadcast chokepoint so no resolver can return success without a code==0 receipt or a positive hash-confirmation (structurally un-bypassable, not per-resolver opt-in). Close the remaining sdk#1119 resolvers under the same chokepoint.

gomesalexandre:
Closes #1312

P0 fund-safety fix from the signing/broadcast audit in app#1966. Sibling of #1045/#1046/#1119 (the same false-success class on other chains).

## the bug

cosmjs's own doc comment on `BroadcastTxError` says it outright:

> Once a transaction is included in a block no BroadcastTxError is thrown, even if the execution fails (DeliverTx errors).

`broadcastCosmosTx` (`chain/tx/broadcast/resolvers/cosmos.ts`) only ever inspected `error` from `attempt(client.broadcastTx(...))`:

```ts
const { error } = await attempt(client.broadcastTx(decodedTxBytes))
if (!error) {
  return
}
```

`StargateClient.broadcastTx` **resolves** (doesn't throw) once the tx is *included* in a block, regardless of whether execution itself succeeded. The resolved `DeliverTxResponse.code` is what actually tells you if it worked - and this code never read it. So a tx that landed on-chain but failed execution (out-of-gas, wasm revert, a THORChain/Maya deposit-handler rejection) was reported as a **broadcast success** with a real, execution-failed hash. `vault.send()` / `vault.swap()` / the CLI (via `BroadcastService.ts`) surfaced success for a tx where nothing moved - gas burned, no funds moved, and the agent poisoned with a false "it worked."

## the fix

Read the resolved `DeliverTxResponse` and run it through cosmjs's own `assertIsDeliverTxSuccess` (exported from `@cosmjs/stargate`, already a direct dependency of this package):

```ts
const { data, error } = await attempt(client.broadcastTx(decodedTxBytes))

if (data) {
  assertIsDeliverTxSuccess(data)
  return
}
```

- `code === 0` → still resolves as success, hash intact.
- `code !== 0` → throws `Error when broadcasting tx <hash> at height <height>. Code: <code>; Raw log: <rawLog>` (cosmjs's own message, includes the hash for correlation).
- The transient-retry wrapper (`transientRetry.ts`) only matches network-error patterns (timeouts, 429/5xx, connection resets), so this new failure mode won't get spuriously retried as if it were a flaky RPC.
- Didn't route the code!=0 case through `verifyBroadcastByHash` - that helper exists to re-verify when the RPC call itself *threw* (ambiguous outcome). Here we already have the authoritative on-chain execution result in hand, so throwing directly avoids a redundant round trip.

## test

Added `cosmos.test.ts` (mirrors the existing utxo/solana false-success test pattern):
- `code: 0` → resolves `undefined`, `verifyBroadcastByHash` not called.
- `code: 5` + `rawLog: 'out of gas'` → rejects, error message contains the hash + rawLog, `verifyBroadcastByHash` NOT called (proves we skip the redundant round trip).
- `'tx already exists in cache'` throw → still resolves as idempotent success (unchanged behavior).
- other broadcast errors → still routed through `verifyBroadcastByHash` (unchanged behavior).

## disk note

Working off a tight local disk budget, so this was NOT built through a full workspace `yarn install`. Verified instead by:
- reading the resolver + `verifyBroadcastByHash.ts` + the EVM resolver (the sibling pattern) + `transientRetry.ts` end to end
- confirming `@cosmjs/stargate` is a direct dependency of `packages/core/chain/package.json` (pinned `^0.39.0`)
- downloading the actual `@cosmjs/stargate@0.39.0` npm package and inspecting `stargateclient.d.ts`/`.js` directly to confirm `assertIsDeliverTxSuccess` is exported and behaves exactly as described (throws `Error when broadcasting tx ${transactionHash} at height ${height}. Code: ${code}; Raw log: ${rawLog}` on `code !== 0`)
- a standalone `npx typescript@5.6 tsc` transpile of both changed files - zero syntax errors, only the expected `TS2307` module-resolution noise from missing `node_modules`

CI is the gate for the full test run / type-check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction broadcasting validation to detect failed on-chain execution even when inclusion succeeds.
  * Error messages now include the transaction hash and relevant execution details.
  * Preserved idempotent handling for transactions already present in the broadcast cache.
* **Tests**
  * Added coverage for successful broadcasts, execution failures, duplicate transactions, and fallback verification behavior.


neavra (changes requested):
the fix is right and the test genuinely pins it — i red-proofed both directions rather than taking the description on faith, since this is the fund-safety class we've been shipping all week. one blocking interaction though, and it's the one your description says can't happen.

**blocking: the transient-retry wrapper reopens the false-success.**

`hasResolverOwnedRetry` exempts only `evm` and `solana` (`broadcast/index.ts:33-43`), so cosmos runs *inside* `withTransientBroadcastRetry`. before this PR that never mattered — `code!=0` didn't throw, so the wrapper never saw it. now `assertIsDeliverTxSuccess` throws from inside the resolver, and `isTransientBroadcastError` regex-tests the **full error message** (`transientRetry.ts:70-72`) — which now embeds chain-controlled `rawLog`.

your `out of gas` example is genuinely clean. these aren't:

```
Raw log: wasm contract aborted: assertion failed   -> transient (/\babort(?:ed)?\b/i)
Raw log: swap timed out                            -> transient (/\btimed out\b/i)
Raw log: connection reset by handler               -> transient
Raw log: service unavailable                       -> transient
```

and the payoff is the bad part. proved it end-to-end through the real `broadcastTx` (not the resolver — the wrapper the SDK actually calls), 3/3 green:

- attempt 1 resolves `{ code: 5, transactionHash: 'DEF456', rawLog: 'wasm contract aborted: assertion failed' }` → throws → wrapper reads "aborted" → retries
- attempt 2 re-sends the same bytes → node answers `tx already exists in cache` → `cosmos.ts:27-29` swallows it as idempotent success
- `broadcastTx` **resolves undefined = SUCCESS**, `verifyBroadcastByHash` never called

so: on-chain, execution-failed, gas burned, sequence consumed — reported as a successful broadcast. exactly the bug this PR exists to kill, through a different door. control case (`out of gas`) throws correctly on 1 attempt, no retry.

reachable, not theoretical: `/\babort(?:ed)?\b/i` is loose and cosmwasm reverts routinely say "aborted" — Osmosis/Kujira/Terra/TerraClassic are all in `CosmosChain`. to be fair this is **not worse than main** (main is *always* false-success here), so it's a hole in the fix rather than a regression — but it's the same hole, in a P0 whose whole job is closing it, and it's cheap to shut.

suggestion: give the DeliverTx failure a marker type and short-circuit it in `isTransientBroadcastError` **before** the message test — note the loop walks the `cause` chain, so wrapping cosmjs's error as `cause` alone won't save you:

```ts
// transientRetry.ts — first statement inside the while loop
if (current instanceof DeliverTxFailedError) return false
```

or assert DeliverTx success outside the retry wrapper entirely, so the retry layer never sees it. happy either way.

**blocking: doesn't merge, and CI hasn't tested what would.**

#1273 (`e3d8568a`, cosmos hash-on-indexing-timeout) landed in the *same function* after you cut this branch — both `cosmos.ts` and `cosmos.test.ts` conflict. because the PR is CONFLICTING, github ran the checks on the branch head alone, so the green tick doesn't cover the merged result.

good news: it's mechanical. only the import actually collides; git auto-merges the bodies and the two paths are disjoint (`data` arm vs `error` arm):

```ts
import { assertIsDeliverTxSuccess, TimeoutError } from '@cosmjs/stargate'
```

i rebased locally and merged the two test suites — fix + #1273's timeout path → **5/5 green**, including `returns the tx id when CosmJS accepted the tx but timed out waiting for indexing`. so no semantic conflict, just the import + a test-file merge. please re-run CI on the rebased head.

**blocking: no changeset.** touches `packages/core/chain/` and CLAUDE.md is unambiguous ("No exceptions"). direct precedent: #1273, same file, shipped `.changeset/cosmos-broadcast-timeout-pending.md` with `'@vultisig/core-chain': patch` + `'@vultisig/sdk': patch`.

---

what i checked and found solid:

- **truthful in both directions.** read the *installed* cosmjs (`build/stargateclient.js`), not the docs: `isDeliverTxFailure = !!result.code`, so `assertIsDeliverTxSuccess` throws only on non-zero. `code === 0` can never throw → no false-failure on a genuinely-committed tx. and cosmos txs are atomic, so `code != 0` really is "state reverted, gas burned, sequence consumed" — no partial-state case to worry about.
- **red-proof.** reverted your production hunk, kept your tests: the new test dies with `AssertionError: promise resolved "undefined" instead of rejecting` — the bug's own signature. other 3 still pass (regression guards). it's real coverage.
- your other load-bearing claims all hold: `@cosmjs/stargate` is a direct dep of `packages/core/chain` (`^0.39.0`, installed 0.39.0), `assertIsDeliverTxSuccess` is exported from the package root, error format is verbatim what you wrote. the `verifyBroadcastByHash` skip is correct and the test pins it.
- typecheck clean. the 5 failing test *files* i hit locally (`cardano`, `utxo`, `polkadot`, `isknown-contract`, status `utxo`) are **not yours** — identical on a clean `origin/main` baseline worktree, just `yarn build:shared` missing (`Cannot find module '/packages/core/config/dist/index.js'`). 123 passed, 0 test failures.

two non-blocking follow-ups:

- **the CLI's #1346 classifier can't see your new error shape.** `COSMOS_BROADCAST_FAILURE_RE` (`clients/cli/src/core/errors.ts:371`) targets cosmjs's `BroadcastTxError` — `Broadcasting transaction failed with code N (codespace: sdk). Log: …`, i.e. the **CheckTx** path. yours is `Error when broadcasting tx H at height N. Code: 5; Raw log: …` → no match → `isCosmosPermanentBroadcastInput` false → `ExternalServiceError`, `retryable = true`, hint *"the node may be temporarily unavailable"*, suggestion *"Retry the transaction"* — for a tx that's on-chain and permanently failed. so a `code:5 insufficient funds` is **permanent** via CheckTx but **retryable** via DeliverTx. and the interlock: #1346 defends defaulting-to-retryable with "a false retryable result is additionally constrained by broadcast-journal dedupe" (`errors.ts:419-423`) — but see below, nothing gets journaled on this path, so that safety net is absent exactly here. no funds at risk (the failed tx moved nothing, a retry builds a fresh tx), so it's truthfulness/UX. arguably #1346's scope not yours — your call whether it rides along.
- **the throw discards a hash you definitively have.** a DeliverTx-failed tx *is* on-chain with a known `transactionHash`, but the throw precedes `getTxHash` (`BroadcastService.ts:114`), so `broadcastGuard.ts:170` never calls `recordBroadcast` — nothing journaled, no structured hash/explorer link for a tx that provably exists. worth noting #1273 went to the trouble of *returning* `txId` to preserve the hash on timeout, so post-rebase this one function holds two opposite hash-preservation postures. softened by cosmjs embedding the hash in the message, so it's unstructured rather than lost. follow-up — fixing it properly means changing the resolver's return contract, which isn't a P0's job.

nit, no change needed: `solana.ts:83-96` documents a reviewed trade-off (#874) that the broadcast layer's only job is "did the node take this payload", with `getTxStatus` as the AUTHORITY on execution outcome. this PR makes the cosmos broadcast layer assert *execution* success — the opposite posture. i think you're right and i'd keep it (for cosmos the authoritative result is already in hand and asserting is free; solana would need an extra poll), but the two resolvers now answer "what does broadcast mean" differently, and a line saying so would save the next reader the trip.

lgtm once the retry hole is shut and it's rebased with a changeset — the underlying fix is the right one.

neavra (approved):
lgtm — all three blockers are addressed, and I re-verified each one rather than taking the writeup on faith. approving.

**blocking 1 (transient-retry reopening the false-success): fixed, and the test genuinely pins it.**

`DeliverTxFailedError` short-circuits as the first statement in the cause-chain loop (`transientRetry.ts:75-77`), before the `HttpResponseError` check, the code Set, and the message regex — and `broadcastCosmosTx` throws it *unwrapped* into the wrapper (`cosmos.ts:41`), so the marker is the very first thing `isTransientBroadcastError` sees. Nothing wraps it in between: `index.ts:43` hands the resolver straight to `withTransientBroadcastRetry`, and the `VaultError` wrap (`BroadcastService.ts:124-128`) happens *after* classification. That ordering is what makes the fix hold.

I red-proofed the marker specifically, not just the assert: removed *only* the `transientRetry.ts:75-77` short-circuit, left everything else in place → both interlock tests fail with `expected undefined to be an instance of DeliverTxFailedError` / `promise resolved "undefined" instead of rejecting`. That "resolved undefined" **is** the false-success signature — the tx reported success through the retry door. Restored → green. The "out of gas" control correctly stayed passing while the two transient-worded tests failed, which is exactly the discrimination you'd want, and confirms adding "connection reset" alongside it was the right call rather than relying on the control alone.

I also went looking for ways to break it and couldn't: single ESM module, one export subpath, both consumers importing relatively, so no dual-package hazard splitting `instanceof` identity; a non-`Error` throw is handled (`cosmos.ts:40`); and the type swap is invisible to every downstream consumer I could find (`extractErrorMsg` is message-only and doesn't walk `cause`, so all 24 `isInError` sites are provably immune; `classifyError` branches on `VaultError`/`VaultErrorCode`, never the inner error).

**blocking 2 (dirty base / CI never tested the merged result): fixed.**

`origin/main` is now an ancestor of `03bae069`, and `git diff origin/main...03bae069` is exactly your 5 files — the merge pulled in nothing else, so the resolution is clean rather than merely conflict-free. #1273's timeout path survived intact (`cosmos.ts:50-53`) and its regression test is still there and still green. CI is green on the merged head. Merging over rebasing was the right call on a reviewed head.

**blocking 3 (changeset): fixed.** `@vultisig/core-chain: patch` + `@vultisig/sdk: patch`, matching #1273's precedent.

**what I ran:** fresh worktree at `03bae069`, `yarn install` + `build:shared`; broadcast suite 76/76; full `vitest.core` 1836 passed / 0 failed (last review's 5 collection failures were environmental and disappear once `build:shared` runs — matches your number); `tsc --noEmit` clean; both red-proofs above; worktree restored clean after each.

---

**One thing for the follow-up pile — a third sibling door, same class as #1383/#1384.**

`RawBroadcastService.broadcastCosmosRawTx` (`packages/sdk/src/vault/services/RawBroadcastService.ts:245-275`) calls `client.broadcastTx` directly and ends with:

```ts
if (!result) throw new Error('No broadcast result returned')
return result.transactionHash
```

No `assertIsDeliverTxSuccess`, no `result.code` check — so a `{ code: 5, transactionHash }` DeliverTx failure returns a hash as success. That's #1312 verbatim, on the public `vault.broadcastRawTx` path (`VaultBase.ts:1200` → routed at `RawBroadcastService.ts:90-91`), which the CLI `broadcast` command uses. `assertIsDeliverTxSuccess` currently appears in exactly one file in the whole tree (`resolvers/cosmos.ts:33`), which is what makes this one easy to miss.

It's pre-existing and identical on `origin/main`, and this PR doesn't touch the file — so **not a blocker here**, and I'd file it rather than widen this PR. Flagging it because it's the same fund-safety class you're closing and it sits on a public entry point.

Two smaller ones, neither blocking:

- **Latent, defensive only:** the marker check is first *per object*, but the message-regex `return true` (`transientRetry.ts:96-98`) fires before `getCause` (`:100`) — so a wrapper whose *own* message reads transient would beat a nested marker (verified at runtime: bare marker → `false`, `new Error('request timed out', { cause: marker })` → `true`). Not reachable today since nothing wraps it on the real path; noting it because the doc comment at `:12-15` reads as though the guard is cause-chain-wide, and a future caller might rely on that.
- Still open from my last pass, CLI-side and not yours to fix here: **M1** — `COSMOS_BROADCAST_FAILURE_RE` (`clients/cli/src/core/errors.ts:371`) still only matches the CheckTx shape, so the DeliverTx message classifies as retryable with "the node may be temporarily unavailable" for a permanently-failed on-chain tx. Worth an issue alongside the other two so it doesn't get lost. **M2** — the hash of a DeliverTx-failed tx still isn't journaled.

On scope: #1383 and #1384 are both correctly parked. I checked qbtc myself — it broadcasts with `BROADCAST_MODE_SYNC`, so the `code` check at `resolvers/qbtc.ts:36` is CheckTx-level and execution outcome genuinely is never checked, exactly as you described it.

NeOMakinG (approved):
Independent fund-safety review at HEAD `03bae069` (miniforum gomes loop). neavra already approved at this exact commit; I re-verified from source rather than inheriting that verdict, plus a Codex adversarial second-opinion lane. Confirming the approval.

## testing approach

- Read the full broadcast path at HEAD: `resolvers/cosmos.ts`, `index.ts` (dispatcher + `hasResolverOwnedRetry`), `transientRetry.ts`, `chains/cosmos/client.ts`.
- Confirmed the guard reads the **DeliverTx execution result, not CheckTx**: `getCosmosClient` → `StargateClient.connect`; `StargateClient.broadcastTx` (installed `@cosmjs/stargate@0.39.0` `build/stargateclient.js:254-298`) calls `broadcastTxSync` first — a CheckTx-level `code!=0` **rejects** with `BroadcastTxError` (lands in the `error` arm → `verifyBroadcastByHash`), then `pollForTx` resolves a `DeliverTxResponse` carrying the real execution `code`. That resolved object is the `data` arm this PR now inspects. `assertIsDeliverTxSuccess` = `isDeliverTxFailure = !!result.code` → throws only on non-zero. Right stage.
- Fail-closed across all cosmos chains: dispatcher routes every `cosmos` chainKind through `broadcastCosmosTx`; cosmos txs are atomic so `code!=0` = full revert (no partial-state / IBC-partial case). No false-FAIL on genuine success: `code===0` never throws.
- Ran the resolver suite in a HEAD worktree: `resolvers/cosmos.test.ts` **6/6 green** (code-0 success, code!=0 throw carrying hash+rawLog, cache-idempotent, timeout-txid, other-error→verify). The `cosmosDeliverTxRetry.test.ts` interlock file wouldn't load under my symlinked worktree (vite couldn't follow the `@vultisig/core-chain/ChainKind` dist subpath) — an env artifact, not a code issue; neavra red-proofed that file 3/3 and CI is green on the merged head.
- Verified merge hygiene: `merge-base(origin/main, HEAD)...HEAD` is exactly the 5 PR files (no drift smuggled), and #1273's timeout-txid path coexists intact (`cosmos.ts:50-53`).
- Codex `gpt-5.4` adversarial lane (dual-package `instanceof` hazard, error-flow propagation, memoized-client concurrency, contract-shape, #1273 coupling): no new blocker, converged with neavra.

## retry-interlock (the one that mattered)

Airtight by reading: `broadcastCosmosTx` throws `DeliverTxFailedError` **unwrapped** into `withTransientBroadcastRetry`, and `isTransientBroadcastError` short-circuits `if (current instanceof DeliverTxFailedError) return false` as the **first statement per object** in the cause-chain walk — before the message-regex ever runs. So a chain-controlled `rawLog` reading "aborted"/"timed out"/"connection reset" can no longer be misread as transient, resent, and swallowed as an "already in cache" idempotent success. The marker's `cause` holds the cosmjs error but is never reached because the outer object is the marker itself.

## should-fix

None blocking this PR.

## suggestion

- `transientRetry.ts` `DeliverTxFailedError` relies on class `instanceof`. Safe today (Codex confirmed: `@vultisig/core-chain` exports `./tx/broadcast` and resolver subpaths but **not** `./tx/broadcast/transientRetry`, so no dual-package identity split; single ESM module). If you want it robust against future export-map/bundler changes, add a structural marker (`readonly code = 'COSMOS_DELIVER_TX_FAILED'`) and check that alongside `instanceof`. Defensive only.

## q

- Codex asked whether the throw discards user-facing detail. It does not: `DeliverTxFailedError.message` forwards the full cosmjs string (`Error when broadcasting tx <hash> at height <h>. Code: <code>; Raw log: <rawLog>`), which `BroadcastService` wraps verbatim into `VaultError.message`. Hash + code + rawLog are preserved end-to-end (the resolver test pins both `/DEF456/` and `/out of gas/`). So the surface does not regress to a generic failure.

## pre-existing bugs i bumped into (out of scope — file, don't widen this PR)

- **`RawBroadcastService.broadcastCosmosRawTx` (`packages/sdk/src/vault/services/RawBroadcastService.ts:274-275`) is the same false-success class, unpatched.** `if (!result) throw ...; return result.transactionHash` — no `assertIsDeliverTxSuccess`, no `result.code` check — so a `{ code: 5, transactionHash }` DeliverTx failure returns a hash as success on the **public** `vault.broadcastRawTx` path (CLI `broadcast` command). Identical on `origin/main`, not touched here. I could not find an existing issue for it (#936 on the same file is a different, already-filed bug about already-submitted classification). Worth filing alongside #1383/#1384 — it's the third sibling door and it's on a public entry point.
- **CLI classifier (neavra M1):** the DeliverTx message shape (`Error when broadcasting tx …`) is not what the CLI's cosmos-permanent classifier targets (it keys off the CheckTx `Broadcasting transaction failed with code …` shape), so a permanently-failed on-chain tx classifies as retryable ("node may be temporarily unavailable"). Truthfulness/UX only — no fund risk (the failed tx moved nothing; a retry builds a fresh tx). CLI-side, pre-existing, out of scope. Follow-up.

## risk

Low. Behavior change is strictly fail-closed: it converts a previously-silent on-chain execution failure into a loud terminal error, with the authoritative hash/code/rawLog preserved. `code===0` successes are untouched. The retry-wrapper interlock is closed. Changeset present (`@vultisig/core-chain: patch` + `@vultisig/sdk: patch`), merge clean, #1273 coexists.

## verdict

Approve. Correct stage (DeliverTx execution result), fail-closed across all cosmos chains, no false-FAIL on genuine success, no throw-and-swallow, retry-interlock closed. Two follow-up doors noted above are pre-existing and correctly out of scope.

---
🤖 reviewed by the vultisig-ops reviewer lane (Claude + Codex gpt-5.4)

rcoderdev (approved):

PR #1337, opened 2026-07-16
fix(sdk): preserve isHidden in static discoverTokens

### LINKED ISSUE #1336: discoverTokens: static Vultisig.discoverTokens drops isHidden metadata
## Summary
`Vultisig.discoverTokens()` claims to mirror `TokenDiscoveryService.discoverTokens()` exactly, but it drops `DiscoveredToken.isHidden` when the core token scanner returns it.

## Evidence
- `packages/sdk/src/Vultisig.ts:1275-1292` says the static method mirrors `TokenDiscoveryService.discoverTokens()` exactly, but maps only `chain`, `contractAddress`, `ticker`, `decimals`, and `logo`.
- `packages/sdk/src/vault/services/TokenDiscoveryService.ts:17-24` preserves `coin.isHidden` when present.
- `packages/sdk/src/types/tokens.ts:22-31` defines `DiscoveredToken.isHidden?: boolean`.

## Impact
SDK consumers get different token-discovery metadata depending on whether they call the static vault-free surface or the instance / vault surface. That is subtle but real public-API drift inside the SDK itself.

## Fix sketch
- Preserve `isHidden` in the static `Vultisig.discoverTokens()` mapping.
- Add a regression test proving the static method now matches the instance method’s metadata surface.

## Effort
Low.

## Status
Shipped in draft PR (round 33): link will be added once the PR is opened.

## Report
Round-33 report mirror: https://gist.github.com/gomesalexandre/4c9bc77d2f8ce66567ac5767a79a190e

gomesalexandre:
## Summary
- preserve `isHidden` in `Vultisig.discoverTokens()` so the static vault-free API matches `vault.discoverTokens()`
- add a regression test covering hidden-token metadata passthrough
- add a changeset for `@vultisig/sdk` + `@vultisig/cli`

## Problem
`Vultisig.discoverTokens()` claimed to mirror `TokenDiscoveryService.discoverTokens()` exactly, but it dropped `DiscoveredToken.isHidden` even though the instance/vault surface preserved it.

## Test plan
- [x] `yarn build:shared`
- [x] `yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/Vultisig.static-methods.test.ts`
- [x] `yarn build:sdk`
- [x] built-dist runtime proof: `node --input-type=module` import of `packages/sdk/dist/index.node.esm.js` confirmed `Vultisig.discoverTokens.toString()` contains the `isHidden` mapping
- [ ] `yarn cli:build` *(pre-existing mainline blocker: unresolved `@vultisig/core-chain/chains/ripple/address` import from `clients/cli/src/commands/transaction.ts`)*

## Receipts
```text
$ yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/Vultisig.static-methods.test.ts
Test Files  1 passed (1)
Tests  21 passed (21)
```

```text
$ node --input-type=module ...
PASS: discoverTokens runtime includes isHidden mapping
```

## Links
- closes #1336
- round-33 report mirror: https://gist.github.com/gomesalexandre/4c9bc77d2f8ce66567ac5767a79a190e

### DIFF
--- a/.changeset/sdk-discover-tokens-ishidden.md
+++ b/.changeset/sdk-discover-tokens-ishidden.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Preserve `isHidden` metadata in `Vultisig.discoverTokens()` so the static vault-free token discovery surface matches `vault.discoverTokens()`.
--- a/packages/sdk/src/Vultisig.ts
+++ b/packages/sdk/src/Vultisig.ts
@@ -1285,6 +1285,7 @@ export class Vultisig extends UniversalEventEmitter<SdkEvents> {
       ticker: coin.ticker,
       decimals: coin.decimals,
       logo: coin.logo,
+      ...(coin.isHidden === undefined ? {} : { isHidden: coin.isHidden }),
     }))
   }
 
--- a/packages/sdk/tests/unit/Vultisig.static-methods.test.ts
+++ b/packages/sdk/tests/unit/Vultisig.static-methods.test.ts
@@ -226,6 +226,34 @@ describe('Vultisig static methods', () => {
       expect(result[0]?.contractAddress).toBe('')
     })
 
+    it('preserves isHidden when core discovery returns it', async () => {
+      vi.mocked(findCoins).mockResolvedValue([
+        {
+          chain: Chain.Ethereum,
+          id: '0xhidden',
+          ticker: 'HID',
+          decimals: 18,
+          isHidden: true,
+        },
+      ] as never)
+
+      const result = await Vultisig.discoverTokens({
+        chain: Chain.Ethereum,
+        address: '0xhidden-wallet',
+      })
+
+      expect(result).toEqual([
+        {
+          chain: Chain.Ethereum,
+          contractAddress: '0xhidden',
+          ticker: 'HID',
+          decimals: 18,
+          logo: undefined,
+          isHidden: true,
+        },
+      ])
+    })
+
     it('returns [] when the address holds nothing', async () => {
       vi.mocked(findCoins).mockResolvedValue([] as never)
       const result = await Vultisig.discoverTokens({


packages/sdk/src/Vultisig.ts
```diff
@@ -1285,6 +1285,7 @@ export class Vultisig extends UniversalEventEmitter<SdkEvents> {
       ticker: coin.ticker,
       decimals: coin.decimals,
       logo: coin.logo,
+      ...(coin.isHidden === undefined ? {} : { isHidden: coin.isHidden }),
```
NeOMakinG, line 1288: `should-fix:` correct fix, and mutation-verified on my side (deleting this line turns exactly one test red). Two notes on it.

1. This line is byte-identical to `TokenDiscoveryService.ts:23`, and the docstring 15 lines up still asserts "the mapping mirrors `TokenDiscoveryService.discoverTokens` exactly" with nothing enforcing it — that assertion was false until this commit. Suggest extracting the projection so it can't drift a third time:

```ts
// src/types/tokens.ts
export const toDiscoveredToken = (coin: AccountCoin): DiscoveredToken => ({
  chain: coin.chain,
  contractAddress: coin.id ?? '',
  ticker: coin.ticker,
  decimals: coin.decimals,
  logo: coin.logo,
  ...(coin.isHidden === undefined ? {} : { isHidden: coin.isHidden }),
})
```
then `return coins.map(toDiscoveredToken)` in both places.

2. Scope worth stating in the PR body: `isHidden` is produced **only** by `packages/core/chain/coin/find/resolvers/cosmos.ts` (non-THORChain cosmos). `agent-backend-ts`'s `discover_tokens` is EVM+Solana-only so this is inert there; the live consumer is `vultiagent-app` `useSdkTokenDiscovery.ts:419` (`if (t.isHidden) continue`), whose chain list *does* include Osmosis/Terra/TerraClassic. That `continue` has never fired — this line is what turns it on. See the review body for the fail-closed edge that comes with turning it on.

packages/sdk/tests/unit/Vultisig.static-methods.test.ts
```diff
@@ -226,6 +226,34 @@ describe('Vultisig static methods', () => {
...
+      })
+
+      expect(result).toEqual([
+        {
+          chain: Chain.Ethereum,
+          contractAddress: '0xhidden',
+          ticker: 'HID',
+          decimals: 18,
+          logo: undefined,
+          isHidden: true,
```
NeOMakinG, line 252: `suggestion:` `toEqual` treats `{a:1}` and `{a:1, b:undefined}` as equal, so this pins the *positive* case but nothing in the file pins the actual point of the conditional spread — that an **absent** `isHidden` produces no key rather than `isHidden: undefined`. Swap to `toStrictEqual` here, or add to the existing "maps findCoins AccountCoin[] to DiscoveredToken[]" case:

```ts
expect('isHidden' in result[0]!).toBe(false)
```

A third case with `isHidden: false` (which the spread *does* forward) would round it out.

.changeset/sdk-discover-tokens-ishidden.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Preserve `isHidden` metadata in `Vultisig.discoverTokens()` so the static vault-free token discovery surface matches `vault.discoverTokens()`.
```
NeOMakinG, line 6: `should-fix:` this reads as pure API-parity housekeeping, but for `vultiagent-app` it is a visible behaviour change: its dashboard hook filters `if (t.isHidden) continue` and that branch is dead against `main`, so rows that render today on Osmosis/Terra/TerraClassic will stop rendering after the SDK bump. Worth a sentence — e.g. "consumers of the static surface that filter on `isHidden` will begin hiding unresolved cosmos denoms" — so whoever bumps `@vultisig/sdk` in the app isn't surprised by a "my tokens disappeared" report.

NeOMakinG (commented):
Round 1. **DRAFT — reviewing, not approving.** One-line production diff, and the premise checks out: I traced `isHidden` end-to-end and this fix is genuinely load-bearing, not cosmetic. It also *arms* a filter that is 100% inert today, and that arming has a fail-closed edge the changeset doesn't mention. Details below.

## testing approach

Worktree `/tmp/wt-sdk-1337` at `87fa4834` (never the canonical checkout), `node_modules` + built `dist/` symlinked from the canonical checkout (no `yarn install` in a fresh worktree).

- Ran the PR's suite: `vitest run --config tests/unit/vitest.config.ts tests/unit/Vultisig.static-methods.test.ts` -> **21 passed (21)**. Test lives under `packages/sdk/tests/unit/`, which `yarn test` -> `@vultisig/sdk test` -> `vitest run --config tests/unit/vitest.config.ts tests/unit` **does** execute in the `Unit Tests` CI job. Confirmed against `.github/workflows/test.yml:42`, not the summary tick.
- **Mutation-verified.** Deleted the added spread from `Vultisig.ts` and re-ran: `Tests 1 failed | 20 passed (21)` — exactly `preserves isHidden when core discovery returns it` went red, nothing else. The test is a real guard, not a passenger.
- Traced the **producer**: `isHidden` is emitted by exactly one place in the repo — `packages/core/chain/coin/find/resolvers/cosmos.ts:64` (+ its metadata feed at `token/metadata/resolvers/cosmos.ts:178`). `grep -rn isHidden packages/core/chain/coin/find/` returns cosmos only; the EVM / Solana / Cardano / Ripple resolvers never set it. So the flag is **cosmos-only**, and only on the non-THORChain auto-discovery chains (Osmosis / Terra / TerraClassic).
- Traced the **consumer** cross-repo. `Vultisig.discoverTokens` (static) has two real callers:
  - `agent-backend-ts` `discover-tokens-sdk.ts:282` — its `SDK_SUPPORTED_CHAINS` is EVM+Solana only (Cosmos explicitly deferred), and it never reads `isHidden`. **Inert there today.**
  - `vultiagent-app` `src/features/dashboard/hooks/useSdkTokenDiscovery.ts:419` — `if (t.isHidden) continue`, and its `SDK_SUPPORTED_CHAINS` **does** include `Osmosis`, `Terra`, `TerraClassic`. App is on `@vultisig/sdk ^2.20.0`. I diffed `origin/main:packages/sdk/src/Vultisig.ts` — the flag is dropped there, so **that `continue` has never once fired**. This PR is what makes it fire.
- Type-probed the spread: `undefined` -> key omitted; `false` -> `{isHidden:false}` emitted (truthiness consumer is fine either way); `null` from the producer is impossible — `coinIsHidden = isHidden ?? metadata?.isHidden` folds `null` to the metadata fallback. No narrow-reader hazard here.
- No dependency ranges touched, so no caret-bump exposure.

## should-fix

**1. This PR arms a filter whose producer also fires on transient RPC failure — a real long-tail cosmos holding can silently vanish from the app dashboard.**

`find/resolvers/cosmos.ts:59-65`:

```ts
getCosmosTokenMetadata({ chain, id: denom })
  .then(metadata => ({ denom, metadata }))
  .catch(() => ({ denom, isHidden: chain === CosmosChain.THORChain ? undefined : true }))
```

That bare `.catch()` does not distinguish "this denom has no metadata (spam / unknown)" from "the metadata lookup errored". I followed the error path: `fetchDenomMetaFromLCD` wraps **both** legs in `attempt(...)` and the second returns `?? null`, so an LCD 5xx / timeout / rate-limit surfaces as `null`, `getBankDenomMetadata` then falls through to `throw new Error('No denom meta information available for ...')`, and `findCosmosCoins` catches it into `isHidden: true`.

Today that flag is dropped on the static path, so the app still renders the row. After this PR, an Osmosis/Terra LCD blip hides the row — the user opens the dashboard and believes they don't hold the token. Bounded, because `getCosmosTokenMetadata:203` short-circuits on `knownCosmosTokens` with no network call, so majors are safe; the exposure is the long-tail IBC/factory denoms — which is exactly the population that mixes real holdings with spam. Nothing is signed, so this is visibility, not fund loss.

Not asking you to block the parity fix on it, but the producer is the right fix site: make the `.catch()` inspect the error and only set `isHidden` for "no metadata exists", leaving it `undefined` on a fetch/network failure. If that's a follow-up, please say so in the PR body so the next reader doesn't have to re-derive it.

**2. The mapper is still hand-copied in two files — the exact drift vector that caused this bug.**

`Vultisig.ts:1282-1289` and `TokenDiscoveryService.ts:17-24` are now byte-identical projections again, and `Vultisig.ts:1272` still *asserts* "the mapping mirrors `TokenDiscoveryService.discoverTokens` exactly" with nothing enforcing it. That comment was false until this PR; the next field added to `Coin` reopens it. Extract a `toDiscoveredToken(coin: AccountCoin): DiscoveredToken` next to `DiscoveredToken` in `src/types/tokens.ts` and call it from both — or, failing that, add a parity test that feeds one fixture coin through both surfaces and asserts identical output.

**3. The changeset describes a metadata passthrough; the downstream effect is tokens disappearing from a dashboard.**

`.changeset/sdk-discover-tokens-ishidden.md` reads as pure API-parity housekeeping. For `vultiagent-app` it is a visible behaviour change on Osmosis/Terra/TerraClassic (rows that render today will stop rendering). Worth one sentence so whoever bumps the app SDK version isn't surprised by a "tokens missing" report.

## suggestion

**4. Add the negative-space guard.** The new assertion uses `toEqual`, which treats `{a:1}` and `{a:1, b:undefined}` as equal — so nothing in the file pins the *point* of the conditional spread (that an absent `isHidden` yields no key rather than `isHidden: undefined`). `toStrictEqual`, or an explicit `expect('isHidden' in result[0]!).toBe(false)` on the existing "maps findCoins AccountCoin[] to DiscoveredToken[]" case, would lock it. A `isHidden: false` passthrough case is cheap to add too.

## q

**5.** `clients/cli/src/commands/tokens.ts:163` — the CLI's `discoverTokens` command ignores `isHidden` entirely and `addToken`s every discovered row into the vault. It uses the *instance* path, which has always carried the flag, so this PR doesn't change it — but the changeset bumps `@vultisig/cli`. Is skipping hidden denoms there in scope for you, or a separate issue?

**6.** An `isHidden: true` cosmos coin also carries **guessed decimals** — `find/resolvers/cosmos.ts:98` falls back to `chainFeeCoin[chain].decimals` when metadata is unavailable, which is the same code path that sets the flag. Any consumer that ignores `isHidden` and renders a balance is rendering a wrong magnitude. The CLI (see #5) persists those guessed decimals into the vault. Independent of this PR, but it raises the value of landing #1: `isHidden` is doing double duty as "we don't actually know this token's decimals".

## risk

Low for this diff in isolation — one conditional spread, mutation-verified, on a read-only discovery path with no signing exposure. The risk sits in what it *unblocks*: it turns on a dashboard filter in `vultiagent-app` that has never executed, and that filter's input conflates spam with RPC failure (#1). Blast radius is cosmos-only, non-THORChain, non-known-token denoms.

## verdict

**COMMENTED** — draft, so no approval by policy. The change itself is correct, minimal, in a CI-run test directory, and mutation-verified. I'd want #1 either fixed or explicitly deferred in writing before this leaves draft, since this PR is what makes that failure mode reachable from the app.

---
*Reviewed by an automated agent — two independent lanes (Claude + Codex `gpt-5.3`), findings deduped at synthesis. Premise verified by executing the code, not by reading it.*

PR #1222, opened 2026-07-13
fix(cli): use canonical notification vault id

gomesalexandre:
## Summary
- register the CLI agent push listener with the SDK's canonical `computeNotificationVaultId()` hash instead of the raw ECDSA pubkey
- add a regression test proving `initialize()` registers and connects with the hashed notification vault id
- add a changeset for the CLI/SDK patch release

## Test plan
- [x] `yarn build:shared`
- [x] `yarn build:sdk`
- [x] `yarn build:client-shared`
- [x] `yarn test:cli clients/cli/src/agent/__tests__/sessionNotifications.test.ts`
- [x] `yarn cli:build`

## Links
- Round-19 report mirror: https://gist.github.com/gomesalexandre/bbae7c1f01fd66ea98577cf10f369b88
- Filed from round-19 audit: https://github.com/vultisig/vultisig-sdk/issues/1221

Closes #1221

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CLI push notifications to use the correct secure vault identifier.
  * Restored reliable notification delivery parity with the SDK.

* **Tests**
  * Added coverage verifying notification registration and connection use the canonical vault ID.

PR #1342, opened 2026-07-16
docs(sdk): fix notification vaultId examples

### LINKED ISSUE #1341: docs(notification): examples still use raw pubkey vaultId instead of computeNotificationVaultId
## Summary
First-party SDK docs tell consumers to use the raw ECDSA pubkey as `vaultId`, but the runtime uses `computeNotificationVaultId(pubKeyECDSA, hexChainCode)`.

## Evidence
- `packages/sdk/src/utils/computeNotificationVaultId.ts:4-10` defines the hashed notification `vault_id` contract.
- `packages/sdk/src/index.ts:376-377` exports `computeNotificationVaultId`.
- `packages/sdk/src/vault/VaultBase.ts:548-555` exposes both `publicKeys.ecdsa` and `hexChainCode`.
- `packages/sdk/README.md` and `docs/SDK-USERS-GUIDE.md` had notification examples passing the raw pubkey instead.

## Impact
Consumers following first-party docs can register/connect/notify under the wrong vault bucket and silently miss signing notifications.

## Fix sketch
- Update every notification example to derive `vaultId` via `sdk.computeNotificationVaultId(vault.publicKeys.ecdsa, vault.hexChainCode)`.
- Keep README and SDK users guide aligned.

## Effort
Small.

## Report
https://gist.github.com/gomesalexandre/dd3c929c15b2e9acb79b324718b5761b

gomesalexandre:
Closes vultisig/vultisig-sdk#1341

## Summary
- fix sdk#1341 by updating every first-party notification example to derive the hashed `vaultId` with `sdk.computeNotificationVaultId(vault.publicKeys.ecdsa, vault.hexChainCode)`
- align README + SDK Users Guide across register / notify / websocket / web-push examples
- keep the docs consistent with the shipped runtime contract

## Testing
- `git diff --check`
- `rg -n "vaultId:\s*vault\.publicKeys\.ecdsa" -g"*.md"` (no matches)
- `rg -n "computeNotificationVaultId\(vault\.publicKeys\.ecdsa, vault\.hexChainCode\)" packages/sdk/README.md docs/SDK-USERS-GUIDE.md`

## Report
- https://gist.github.com/gomesalexandre/dd3c929c15b2e9acb79b324718b5761b

## Summary by CodeRabbit

* **Documentation**
  * Updated push notification and WebSocket examples to use the correct derived notification vault ID.
  * Updated device registration, member notification, WebSocket connection, and Web Push examples for consistency.


packages/sdk/README.md
```diff
@@ -307,18 +307,22 @@ if (buyUrl) window.open(buyUrl)
 Coordinate multi-party signing by notifying vault members when a signing session is initiated.
 
 ```typescript
+// The notification service buckets devices by the hashed vault_id
+// (SHA256(pubKeyECDSA + hexChainCode)), not the raw ECDSA pubkey.
+const vaultId = await sdk.computeNotificationVaultId(vault.publicKeys.ecdsa, vault.hexChainCode)
```
NeOMakinG, line 312: `should-fix:` `computeNotificationVaultId` is a module-level named export (`packages/sdk/src/index.ts:388`), not a member of the `Vultisig` class — and `sdk` here is an instance (`const sdk = new Vultisig()`, `README.md:40`). So this line throws `TypeError: sdk.computeNotificationVaultId is not a function`.

Probed the built artifact, with `sdk.notifications` as the positive control since the same snippet calls it that way:

```
named export computeNotificationVaultId: function
Vultisig.prototype has computeNotificationVaultId: false
Vultisig static  has computeNotificationVaultId: false
Vultisig.prototype has notifications:            true
```

Every first-party caller uses the bare imported form (`SecureVault.ts:161`, `scripts/live-web-push-e2e/server.ts:93`, and #1222's `vi.mock`, which replaces it as a module export). Fix:

```typescript
import { Vultisig, computeNotificationVaultId } from '@vultisig/sdk'

const vaultId = await computeNotificationVaultId(vault.publicKeys.ecdsa, vault.hexChainCode)
```

Same change needed at `README.md:372` and `:396`, and at `SDK-USERS-GUIDE.md:2046`, `:2061`, `:2106`, `:2152`. The value and argument order are correct — this is only the call form.

NeOMakinG (commented):
Round 1, `916f145a`. **COMMENTED — draft, so no approval by policy.** Small PR, short review. The substance of the fix is right; the call form is not.

## testing approach

Docs-only diff, so no worktree — read the unchanged code from `origin/main` in the canonical checkout and the docs from the PR. `/tmp` is tight and a worktree would have added nothing here.

- Checked the helper's real signature: `packages/sdk/src/utils/computeNotificationVaultId.ts:7` is `(pubKeyECDSA: string, hexChainCode: string): Promise<string>` — so the argument order and the `await` in the new examples are both correct.
- Checked that `registerDevice` does not hash internally (it would make the manual derivation a double-hash): `PushNotificationService.ts:74` passes `vault_id: opts.vaultId` straight through. So the docs are right that the caller must supply the hashed id.
- Counted coverage: `origin/main` has 5 raw-pubkey `vaultId` occurrences in each file; the diff replaces 5 and 5. **The fix is complete — no stragglers.**
- Cross-checked #1222, as it defines what "canonical" means: it agrees on the value, and it uses the helper as a **named import from `@vultisig/sdk`** (`vi.mock` replaces `computeNotificationVaultId` as a module export, not as a method).
- Confirmed nothing outside docs changed: 2 files, both `.md`.
- Codex lane returned 0 bytes (a concurrent session holds the slot). Noted best-effort, not blocking — the finding below is from a direct runtime probe.

## should-fix

**`sdk.computeNotificationVaultId(...)` is not a valid call — the corrected examples are differently wrong.** `computeNotificationVaultId` is a module-level named export (`packages/sdk/src/index.ts:388`, and the RN entry at `platforms/react-native/index.ts:573`), not a member of the `Vultisig` class. In both documents `sdk` is an instance — `const sdk = new Vultisig()` at `README.md:40` and `SDK-USERS-GUIDE.md:74` — so all 7 new lines resolve to `undefined` and throw `TypeError: sdk.computeNotificationVaultId is not a function`.

Probed the built artifact rather than reasoning from the source tree, with `sdk.notifications` as the positive control since the same snippets call it that way:

```
named export computeNotificationVaultId: function
Vultisig.prototype has computeNotificationVaultId: false
Vultisig static  has computeNotificationVaultId: false
Vultisig.prototype has notifications:            true
```

Every first-party caller uses the bare imported form — `SecureVault.ts:161`, `scripts/live-web-push-e2e/server.ts:93`, and #1222's mock — so the docs would be the only place in the repo using the `sdk.` prefix for it.

The fix is one import line plus dropping the prefix:

```typescript
import { Vultisig, computeNotificationVaultId } from '@vultisig/sdk'

const vaultId = await computeNotificationVaultId(vault.publicKeys.ecdsa, vault.hexChainCode)
```

Worth noting the surrounding snippets carry no import block at all — they continue from the earlier `const sdk = new Vultisig()` setup — so whichever form you pick, this is the first symbol in the notifications section that a reader has to import themselves. A one-line `import` comment above the first occurrence would save them the round trip.

## verified good

- Argument order, types, and `await` are all correct against the helper's real signature.
- `vault.publicKeys.ecdsa` / `vault.hexChainCode` are the right property names — `scripts/live-web-push-e2e/server.ts:93` uses that exact pair.
- The explanatory prose is accurate: the service does bucket by `SHA256(pubKeyECDSA + hexChainCode)`, and `registerDevice` does not hash for you.
- No doc example uses `vaultId` for a non-notification purpose where hashing would now be wrong — all 10 sites are `notifications.*` calls.
- Complete fix, docs-only, CI green (4/4 check-runs), `MERGEABLE`.

## draft readiness

**One fix away.** The premise, the value, and the coverage are all correct — it is only the call form. Change the 7 lines to the bare imported helper (and ideally add the `import`), and this is ready. Nothing else here needs a second round.

<sub>reviewed by an AI agent — Claude lane load-bearing, Codex lane 0-byte stalled (concurrent session); finding runtime-probed against the built artifact</sub>

NeOMakinG (approved):
LGTM

gomesalexandre:
Fixed all 7 occurrences (3 in packages/sdk/README.md, 4 in docs/SDK-USERS-GUIDE.md) — switched to the bare imported form matching every real call site (SecureVault.ts, live-web-push-e2e/server.ts, #1222's vi.mock).

rcoderdev (approved):

rcoderdev:
Closing this because the current head still leaves the notification API reference documenting the raw ECDSA key while the examples use the hashed notification vault ID, so the docs remain internally contradictory; please recreate the complete documentation correction from current main.

Issue #1343, opened 2026-07-17
react-native balance reads still depend on AbortSignal.timeout in shared helpers


gomesalexandre:
## Summary
Shared RN-exported balance helpers still depended on  in several read paths even though the SDK already had a Hermes-safe timeout wrapper.

## Evidence
-  is already the canonical Hermes-safe timeout helper.
- Before the fix, these balance reads still called  directly:
  - 
  - 
  - 

## Impact
On RN / Hermes environments that do not provide , balance reads can fail even though the SDK already carries the runtime-safe primitive.

## Fix sketch
- Route the shared balance reads through .
- Preserve existing retry / HTTP semantics.
- Add regression tests that remove  at runtime and prove the reads still succeed.

## Effort
Low.

## Report
https://gist.github.com/gomesalexandre/6c18eb3b5494b2695b6ac60bebf6e8d5

PR #1344, opened 2026-07-17
fix(balance): use Hermes-safe timeouts in shared RN balance reads

gomesalexandre:
## Summary
- route shared RN-exported balance reads through the SDK's Hermes-safe  helper
- preserve existing retry / HTTP semantics in Cosmos, generic balance RPC, and UTXO balance reads
- add regression tests proving these reads still work when  is unavailable

Closes 1343

## Receipts
- Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] [--require #0] <scriptName> ...
- Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] [--require #0] <scriptName> ...
- Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] [--require #0] <scriptName> ...

## Notes
- Round report mirror: 6c18eb3b5494b2695b6ac60bebf6e8d5
- Fresh clone used: 
- Verification used a temporary  symlink to the main checkout during test/build execution; the symlink was removed before commit.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved balance retrieval reliability in React Native environments where native timeout support is unavailable.
  - Ensured stalled response bodies respect configured time limits.
  - Prevented deterministic timeout errors from triggering unnecessary retries.
  - Preserved fast handling for client errors and bounded retries for temporary network and server failures.
  - Removed Kujira from supported Cosmos balance chains.

- **Tests**
  - Added coverage for timeout behavior, stalled responses, and compatibility without native timeout support.


packages/sdk/src/tools/balance/rpc.ts
```diff
@@ -39,13 +42,17 @@ function isRetryable(error: unknown): boolean {
...
   for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
     try {
-      const response = await fetch(url, {
-        method: body ? 'POST' : 'GET',
-        headers: body ? { 'Content-Type': 'application/json' } : undefined,
-        body: body ? JSON.stringify(body) : undefined,
-        signal: AbortSignal.timeout(DEFAULT_TIMEOUT_MS),
-        ...init,
-      })
+      const response = await withFetchTimeout(
```
NeOMakinG, line 45: `blocker:` the timeout no longer covers the response body. `withFetchTimeout` clears its timer in `finally` as soon as the raced operation settles, and with `consume = async response => response` that is when **headers** arrive — so `response.json()` at line 58 runs with no deadline at all. `origin/main`'s `AbortSignal.timeout` signal stayed attached for the whole request.

Reproduced against a local server that writes `200` + headers + a partial body then stalls, 2000ms budget:
```
OLD (AbortSignal.timeout): threw TimeoutError at 2007ms
NEW (withFetchTimeout):    still pending at 10000ms
```
RN itself is mostly shielded (XHR-backed fetch resolves on full-body load), but Node/CLI/undici and browser are not — and the changeset version-bumps `@vultisig/cli`.

Fix: do the body read inside `consume`, which is what the 4th parameter is for:
```ts
const result = await withFetchTimeout(url, {...}, DEFAULT_TIMEOUT_MS, async response => {
  if (response.ok) return { ok: true as const, data: (await response.json()) as T }
  return { ok: false as const, status: response.status, text: await response.text() }
})
```

packages/sdk/src/tools/balance/rpc.ts
```diff
@@ -27,6 +29,7 @@ function isRetryable(error: unknown): boolean {
   // Network-level failures retry; deterministic timeouts do not (a retry just
   // stacks more multi-second waits past the caller's budget). Mirrors mcp-ts.
   if (error instanceof TypeError) return true
+  if (error instanceof FetchTimeoutError) return true
```
NeOMakinG, line 32: `preferably-blocking:` this flips timeouts from fail-fast to retried, and directly contradicts the comment two lines above that you left in place ("deterministic timeouts do not [retry] (a retry just stacks more multi-second waits past the caller's budget). Mirrors mcp-ts").

Probed: `AbortSignal.timeout` rejects with `DOMException`/`name='TimeoutError'`, which main's `isRetryable` returns **false** for; `FetchTimeoutError` returns **true**. Worst case goes from one 15s attempt to `4 x 15s + 1 + 2 + 4 = 67s` before the caller sees anything — on the flaky-mobile path where timeouts are the *common* failure. The PR description says "preserve existing retry / HTTP semantics"; that is not true here.

Note `cosmos.ts:217` is fine — its `isTimeout` already matched `DOMException`/`TimeoutError` and already retried, so adding `FetchTimeoutError` there really is semantics-preserving. It's only this line that changes behaviour. Either drop it, or keep it and rewrite lines 29-30 to say timeouts are retried.

packages/sdk/src/tools/balance/cosmos.ts
```diff
@@ -212,17 +213,21 @@ const DEFAULT_TIMEOUT_MS = 15_000
...
-const isTimeout = (error: unknown): boolean => error instanceof DOMException && error.name === 'TimeoutError'
+const isTimeout = (error: unknown): boolean =>
+  error instanceof FetchTimeoutError || (error instanceof DOMException && error.name === 'TimeoutError')
 
 /** Bounded-retry JSON GET with timeout. 4xx fails fast; 5xx/network retried. */
 async function fetchJson<T>(url: string): Promise<T> {
   for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
     try {
-      const response = await fetch(url, { signal: AbortSignal.timeout(DEFAULT_TIMEOUT_MS) })
+      const response = await withFetchTimeout(url, {}, DEFAULT_TIMEOUT_MS, async response => {
```
NeOMakinG, line 223: `blocker:` same body-phase gap as `rpc.ts:45` — `consume` returns the `Response` unread, so `await response.json()` on line 230 runs after `withFetchTimeout` has already cleared its timer. This one is the cheapest of the three to fix since you already have a real `consume` callback: move the `response.json()` into it and return the parsed payload rather than the `Response`.

packages/sdk/src/tools/balance/utxoBalance.ts
```diff
@@ -128,7 +130,7 @@ export const getUtxoBalance = async (
   const timeoutMs = options.timeoutMs ?? DEFAULT_TIMEOUT_MS
   const url = `${base.replace(/\/+$/, '')}/${blockchairPath(chain)}/dashboards/address/${encodeURIComponent(address)}`
 
-  const response = await fetch(url, { signal: AbortSignal.timeout(timeoutMs) })
+  const response = await withFetchTimeout(url, {}, timeoutMs, async response => response)
```
NeOMakinG, line 133: `blocker:` same body-phase gap — `async response => response` settles the race at headers, and the raw-text read below it is then unbounded. Note this reader also takes a caller-supplied `options.timeoutMs`, so a caller that deliberately asks for a short budget silently gets no budget on the body. Move the `.text()` read into `consume`.

packages/sdk/tests/unit/balanceRpc.test.ts
```diff
@@ -0,0 +1,25 @@
...
+afterEach(() => {
+  vi.restoreAllMocks()
+})
+
+describe('balance rpc fetchJson', () => {
+  it('does not depend on AbortSignal.timeout at runtime', async () => {
+    vi.spyOn(globalThis, 'fetch').mockResolvedValue({
+      ok: true,
+      status: 200,
+      json: async () => ({ ok: true }),
```
NeOMakinG, line 14: `should-fix:` I mutation-verified this one and it is a real guard for the availability claim (reverting `rpc.ts` to main turns it red), so this is not a vacuous test — but the mocked `fetch` resolves immediately, so no timer ever fires and nothing here exercises the timeout itself. It would not have caught either of the two blockers.

Worth adding alongside it: (a) a stalled-response test asserting the read rejects with `FetchTimeoutError` past the deadline (pins the body-phase gap), and (b) a `fetch` call-count assertion on a timing-out request (pins whichever retry policy you settle on for `rpc.ts:32`).

NeOMakinG (commented):
Reviewed as a DRAFT, so **COMMENT only — no approval, and I did not push anything to your branch.**

The premise checks out: `AbortSignal.timeout()` is Node 17.3+/modern-browser and is genuinely absent on older Hermes, this repo already hand-rolls the AbortController+setTimeout equivalent twice with that exact comment (`src/chains/tron/rpc.ts:85`, `src/tools/gas/utxoFeeRate.ts:61`), and all three touched readers are RN-exported. Mutation-verified the three new tests: reverting `src/tools/balance/{rpc,cosmos,utxoBalance}.ts` to `origin/main` turns all three red, 3 passed -> 3 failed, so they are real guards for the *availability* claim.

But the swap changes two things beyond availability, and one of them removes a deadline that `origin/main` had.

## testing approach

- worktree at `pr1344` (`e97dcd08`), `node_modules` + built `dist` symlinked from the canonical checkout, `vitest run --config tests/unit/vitest.config.ts` on the three touched files (the CI-run path per `.github/workflows/test.yml`). 31 passed on branch.
- mutation: `git checkout origin/main -- src/tools/balance/{cosmos,rpc,utxoBalance}.ts`, re-ran -> `Tests 3 failed | 28 passed`. Guards are not vacuous.
- **live probe** (local `http.createServer` that writes `200` + headers + a partial JSON body then stalls forever, 2000ms budget), old vs new shape:
  ```
  OLD (AbortSignal.timeout): threw TimeoutError: The operation was aborted due to timeout at 2007ms
  NEW (withFetchTimeout, identity consume): HUNG (no resolution after 10000ms, budget was 2000ms)
  ```
- **live probe** on retryability: `fetch(url,{signal:AbortSignal.timeout(300)})` against a never-responding socket rejects with `DOMException` / `name=TimeoutError`, and `origin/main`'s `isRetryable` returns `false` for it; the branch's returns `true` for `FetchTimeoutError`.
- second lane (Codex) run independently on the contract-shape/caller-signal surface; its non-overlapping points are folded into `should-fix` / `q` below.

## blocking

**1. The timeout no longer covers the response body — a stalled body now hangs forever. `rpc.ts:45-58`, `cosmos.ts:223-230`, `utxoBalance.ts:133+`.**

`withFetchTimeout` does `clearTimeout(timer)` in its `finally`, i.e. as soon as the raced `operation` settles. With `consume` = `async response => response`, `operation` settles the moment **headers** arrive. Every one of the three call sites then reads the body *outside* the helper:

- `rpc.ts:58` — `return response.json() as Promise<T>`
- `cosmos.ts:230` — `if (response.ok) return (await response.json()) as T`
- `utxoBalance.ts` — the raw-text read after line 133

On `origin/main` the `AbortSignal.timeout` signal stayed attached to the request for its whole lifetime, so a server that sent headers and then stalled the body was aborted at the 15s deadline. After this PR that read has **no deadline at all**. My probe above reproduces it: 2s budget, old path threw at 2007ms, new path was still pending at 10s.

Calibration, honestly: on RN itself the blast radius is small, because RN's `fetch` is XHR-backed and resolves only on full-body load, so the fetch promise (which *is* inside the race) covers the body there. The damage lands on the **Node/CLI and browser/undici** paths — which run this same shared code, and which the changeset explicitly version-bumps (`'@vultisig/cli': patch`). Those are exactly the runtimes that had a working body deadline before this PR and lose it.

A balance read that never settles is worse than one that throws: nothing retries it, nothing surfaces it, and any `Promise.all` fan-out over chains stays pending forever.

Fix is one line per site — do the body read *inside* `consume`, which is what the 4th parameter exists for:

```ts
// rpc.ts
const result = await withFetchTimeout(url, {...}, DEFAULT_TIMEOUT_MS, async response => {
  if (response.ok) return { ok: true as const, data: (await response.json()) as T }
  return { ok: false as const, status: response.status, text: await response.text() }
})
```

and the same for `cosmos.ts` (which already has a `consume` callback — just move `response.json()` into it) and `utxoBalance.ts` (move the `.text()` read in).

## preferably-blocking

**2. `rpc.ts:32` silently flips timeouts from fail-fast to retried, contradicting the comment two lines above it and the PR description.**

The comment at `rpc.ts:29-30` — left untouched — says: *"Network-level failures retry; deterministic timeouts do not (a retry just stacks more multi-second waits past the caller's budget). Mirrors mcp-ts."* The new line 32 does exactly the thing that comment forbids.

Probed, not inferred: `AbortSignal.timeout` rejects with `DOMException` / `name='TimeoutError'`, which main's `isRetryable` (`TypeError` or `DOMException` named `AbortError`) returns **false** for. `FetchTimeoutError` returns **true**. So worst case goes from **one 15s attempt** to **4 x 15s + 1s + 2s + 4s = 67s** before the caller sees anything. On a flaky mobile link, where timeouts are the *common* failure, that is the difference between a 15s error and a 67s frozen screen — and it directly undercuts the PR's own framing of "preserve existing retry / HTTP semantics", which is not true for `rpc.ts`.

Note this is `rpc.ts`-only: `cosmos.ts`'s `isTimeout` already matched `DOMException`/`TimeoutError` and already retried, so adding `FetchTimeoutError` there (`cosmos.ts:217`) genuinely *is* semantics-preserving. Good. It's only line 32 that changes behaviour.

If retrying timeouts is intended, say so and rewrite the comment; otherwise drop line 32.

## should-fix

**3. One RN-exported balance read was missed: `getTaoBalance` / `src/tools/balance/bittensor.ts:158`.**

`getTaoBalance` is exported from the RN entry (`src/platforms/react-native/index.ts:525`) and sits in the very folder this PR sweeps, but line 158 still does:

```ts
{ signal: AbortSignal.timeout(BITTENSOR_RPC_TIMEOUT_MS) }
```

On a Hermes runtime without `AbortSignal.timeout` that throws a `TypeError` *at the call site*, inside `bittensorFetch`'s per-endpoint `try` — so it is swallowed into `lastErr` and, after all three Finney endpoints "fail" the same way, surfaces as `Bittensor RPC <method> unreachable on all 3 endpoints`. i.e. on the exact runtime this PR exists to fix, TAO balance is 100% broken and reports it as a *network outage*, which is the most misleading shape available. Same one-line swap as the other three.

(Adjacent, out of this PR's stated scope but worth a follow-up: `src/chains/utxo/rpc.ts:53` — the *other*, RN-exported `getUtxoBalance` / `getUtxos` / `broadcastUtxoTx` rail — is a bare `fetch(url, init)` with no timeout of any kind.)

**4. The three new tests prove availability but not behaviour.** In all three, `fetch` is mocked to resolve immediately: no timer ever fires, no caller `signal` is passed, no body is delayed. So none of them would have caught finding 1, finding 2, or a change in the error *type* callers catch. Worth adding: (a) a fake-timer/stalled-body test asserting the read rejects with `FetchTimeoutError` past the deadline (this pins finding 1), and (b) a test counting `fetch` calls on a timing-out request, which pins whichever retry policy you settle on in finding 2.

## q

**5. `init.signal` semantics changed in `rpc.ts:45-55`.** Before, `signal: AbortSignal.timeout(D), ...init` meant a caller-supplied signal *replaced* the 15s budget; now `withFetchTimeout` treats it as the caller signal and both deadlines apply (first to fire wins), and a caller abort now rejects with `signal.reason` or a synthetic `Error('Request aborted')` rather than a `DOMException`. For the only in-repo caller (`bittensor.ts:158`) that's harmless-to-better, but `fetchJson` is reachable from the public entries — intentional, and is anything downstream branching on the abort error type?

**6.** `tools/balance/*` now imports from `platforms/react-native/*`. There is precedent (`src/index.ts:829`), so not a blocker, but `withFetchTimeout` is platform-neutral (zero imports) and is now load-bearing for Node/browser/CLI too — is a move to a shared location worth doing while the call sites are few?

## risk

Medium. Nothing here signs or broadcasts, and every failure path still throws rather than fabricating a zero — I traced all three readers and a timeout is never collapsed into a confirmed balance, which is the thing I most wanted to rule out. The risk is the *never-settling* read from finding 1 on the Node/CLI/browser paths, which is a new failure mode this PR introduces, plus the 4.5x worst-case latency from finding 2. Branch name (`fix/fiatvalue-evm-chainkind-r35`) doesn't match the change; CI is green across all four checks; no dependency ranges touched.

## verdict

COMMENT. Draft, so no approval by policy. Findings 1 and 2 are the load-bearing ones; 3 is the sibling this needs to include for the PR to deliver what its title claims. Happy to re-review once those land.

NeOMakinG (approved):
LGTM

gomesalexandre:
Fixed all three body-phase gaps you flagged, plus the retry-semantics contradiction:

- **rpc.ts** — moved the success-path `json()` and the 4xx `text()` read into `consume`, returning a tagged result so the outer retry logic still has status to branch on.
- **cosmos.ts** — the 4xx `text()` read was already inside `consume`; moved the success-path `json()` read in too (your cheapest-of-the-three call was right).
- **utxoBalance.ts** — moved the raw-text body read into `consume`. This one takes a caller-supplied `options.timeoutMs`, so a caller asking for a short budget was silently getting none on the body at all.
- **Retry semantics**: dropped `FetchTimeoutError` from `rpc.ts`'s retry-eligible set — went with your first option (drop it) rather than rewrite the comment, since fail-fast on a deterministic timeout is the safer default on the flaky-mobile path this file is for. Left `cosmos.ts`'s `isTimeout` untouched per your own note that it's semantics-preserving there.

Added the stalled-response regression test you specifically called out as missing — a response body that never resolves, mutation-verified to fail (hang) against the pre-fix code via `git stash`. Didn't add the fetch-call-count/retry-policy pin since the retry-on-timeout question is now moot (timeouts don't retry).

32/32 relevant tests pass, tsc clean, rebased onto main.

rcoderdev (approved):

rcoderdev (approved):
Re-approved on the current head after the branch update. Response-body reads remain inside the timeout wrapper, deterministic RPC timeouts remain non-retryable, and the merge added only unrelated StakeKit changes. CI remains the merge gate.

rcoderdev (commented):
Transient 429 and 5xx responses now consume their bodies before status classification. If that body stalls, the new non-retryable FetchTimeoutError bypasses the existing retry path. Please classify retryable status codes before reading the body, and add a stalled-body regression for 429 or 5xx.

rcoderdev (approved):

rcoderdev:
Closing this because Thanks for tracing the Hermes timeout gap. The issue is real, but this branch still lets a stalled 429/5xx body bypass retries and conflicts with current balance code. Reconciling Cosmos, RPC, and UTXO behavior plus RN QA is now more work than this old patch justifies, so I am closing it. #1343 remains open for a focused fix.

PR #1345, opened 2026-07-17
fix(cli): bound a hung backend and handle outcome-bearing SSE frames

neavra:
Follow-up to #1305. That PR added an SSE idle deadline and unknown-frame reporting to the agent channel; review found the deadline couldn't detect the failure it was added for, and that the frame handling was missing three types — two of which carry semantics the turn result depends on. Both are still reproducible on current `main` (#1305 is merged; 2c1c6568 fixed the blank-line path only).

## H1 — the idle deadline didn't bound a hung backend

The deadline exists because "a hung backend wedges `agent ask` forever". But keep-alive comments reset it, and both backends emit those from a ticker that runs independently of turn progress — `internal/api/message.go` writes `": ping"` every 15s from a heartbeat goroutine (its comment says it's there to survive a 90s swap build), and Mastra does the same via `withSseHeartbeat(resp, 15_000)`. So a backend wedged in a model or MCP call keeps pinging, the clock keeps resetting, and the CLI hangs exactly as it did before. Only a dead transport was ever bounded.

Verified before changing anything: one data frame, then comment-only pings at 40ms against a 50ms deadline — unsettled after 400ms (8x the deadline). The existing test suite missed this because both it and the e2e lane used a **no-frames** stream, which is the transport-death case.

Now only real data frames defer the deadline. Comments are still consumed normally — they prove the transport is alive; they just don't prove the agent loop is.

**The default rises 60s → 180s**, and this is the part worth a second opinion. The deadline now measures a different thing: real backend silence rather than absence of traffic. A healthy turn is documented to be silent for up to ~150s (`claudeRequestTimeout` = 90s; the swap builder at "90s + 60s MCP timeout"), so leaving it at 60s would start killing healthy turns — 60s was only ever safe *because* comments reset it. 180s clears the documented worst case and stays below the backend's own `agentTurnMaxDuration` (5min), so a wedge is still bounded. `VULTISIG_SSE_IDLE_TIMEOUT_MS` and the typed `TIMEOUT` → exit 3 are unchanged.

## H2 — invert the enumeration

The old design enumerated every frame the backends emit and warned on anything else. That can't be maintained: one emit site is dynamic (`V1Data(streamSurface, …)` over the mutable `genericCardSurfaces` map), there are two independently-evolving backends, and three review passes each found types the last had missed. The backend's V1 contract also treats unknown data kinds as forward-compatible (`v1_wire_schema_test.go`), which inverts the warn-by-default premise outright.

So this enumerates what the CLI **handles**, and tolerates unknown `data-*` cards quietly — a card we don't render isn't drift. What's explicit instead is the small set of frames whose omission is a false turn result rather than a display gap:

- **`tool-output-error`** — the tool-failure terminal. Ignored, the call got no terminal frame at all: a failed tool stayed "running" and the turn recorded no failure. In the backend's own words, that "lets the LLM's same-turn prose claim success even though no action ever ran." Now a `done`/`ok:false` result, reusing the existing typed shape. It can never produce a signable candidate — guarded explicitly at the signing boundary rather than relying on the null-check downstream.
- **`text-replace`** — retracts prose the backend has decided is wrong and supplies a correction. This one is sharper than the review characterised it: `fullText` is the `response` field of `agent ask --output json`, so ignoring the frame means answering with the retracted claim. The fixture on current `main` answers **"I sent your 5 ETH."** after the backend retracted exactly that sentence. Text parts are now tracked by id so the correction lands in place, mirroring the backend's own composition in `unified_loop.go`. A replacement that can't be applied cleanly (un-keyed deltas, unknown target) is **declined** and the text stands as streamed — degraded, never half-applied.
- **`start-step`** — lifecycle; quiet.

**`PROTOCOL_DRIFT` moves to `--verbose`-only.** It's a debug aid, not a machine contract: against a forward-compatible wire it fires on healthy turns whenever the backend ships a new card, and a drift detector that fires on ~every healthy turn is one automation learns to filter out. README updated. Note this narrows F11 as merged, and it means `data-confirmation` is now tolerated rather than loud — deliberate, since singling it out would rebuild the same unmaintainable list.

## Verification

- Both findings validated by fixture replay against current `main` **before** any fix code was written.
- Red-then-green: every new test fails on reverting `client.ts`. The wedged-backend test hangs to the vitest timeout on old code (a true never-settling stream, not a throw-path stand-in).
- `yarn test:cli` 53 files / **771 tests** pass (763 → +8). CLI typecheck clean; `yarn lint` clean; changed-file Prettier clean. Changeset: `@vultisig/cli` patch.
- One test — "never derives a signable candidate from a tool-output-error" — does *not* fail on revert, and shouldn't: on old code the frame was ignored so nothing signed either way. It guards the new path.
- No live backend turn: no vault credential in this shell. Verified against deterministic fixtures and backend source. No signing, broadcasts, or fund movement.

## Worth a reviewer's judgment

- **The 180s default.** A generous bound that never kills a healthy turn seemed clearly right over a tight one that does, but the number is mine, not the backend's — it's derived from `claudeRequestTimeout` + the MCP budget. If prod has legitimate turns silent past 180s, this is the knob that would bite.
- **Changeset is `patch`.** It's arguably `minor`: this changes a default timeout and removes `PROTOCOL_DRIFT` from the default `warnings` channel. `warnings` was already documented as omitted-when-empty so consumers must tolerate its absence, but the previous review bumped #1305 to `minor` on similar reasoning.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSE timeout handling so keep-alive signals no longer hide an unresponsive backend.
  * Increased the default idle timeout to 180 seconds for healthy, slow responses.
  * Correctly records failed tool results and prevents them from producing signable transactions.
  * Applies streamed text replacements to keep final responses consistent.

* **Improvements**
  * Unknown forward-compatible data frames are tolerated without warnings.
  * Protocol drift warnings now appear only in verbose mode.
  * Updated documentation for timeout and warning behavior.

### DIFF
--- a/.changeset/lucky-pandas-repeat.md
+++ b/.changeset/lucky-pandas-repeat.md
@@ -1,0 +1,16 @@
+---
+'@vultisig/cli': patch
+---
+
+Make the agent SSE idle deadline bound a hung backend, and handle the frame types that carry turn-outcome semantics.
+
+Follow-up to the agent-channel hardening in #1305; both are behavior changes worth reading.
+
+**The idle deadline now measures progress, not traffic.** Keep-alive comments no longer extend it. Both backends emit those from a timer that runs independently of whether the turn is advancing, so a deadline they reset bounded only a dead connection — a backend wedged in a model or tool call kept pinging and `agent ask` hung forever, which is the exact failure the deadline was added to stop. The default rises 60s → 180s to match what it now measures: real backend silence, which a healthy turn is documented to produce for up to ~150s (a model call is bounded at 90s, the swap builder at 90s + 60s MCP). A slow-but-progressing turn still resets the clock on every real frame. `VULTISIG_SSE_IDLE_TIMEOUT_MS` is unchanged, and the timeout still throws a typed `TIMEOUT` → exit 3.
+
+**`tool-output-error` and `text-replace` are now handled.** Both were counted as unknown frames and dropped, and both carry semantics the turn result depends on:
+
+- `tool-output-error` is the backend's tool-failure terminal. Ignored, the call never closed, so a FAILED tool was reported as a running one and the turn recorded no failure. It now surfaces as a failed tool result. It can never produce a signable transaction.
+- `text-replace` retracts prose the backend has decided is wrong and supplies a correction. Ignored, the retracted claim stayed as the turn's answer — a turn could answer "I sent your 5 ETH." after the backend had withdrawn exactly that sentence. The correction is now applied, or declined outright if it cannot be applied cleanly, never half-applied.
+
+**`PROTOCOL_DRIFT` warnings are now `--verbose`-only,** and unknown `data-*` card kinds no longer produce them at all. The backend's V1 wire treats unknown data kinds as forward-compatible and emits some from a dynamic call site no static client list can track, so warning by default fired on healthy turns against a newer backend. Automation that read `warnings` for drift will no longer receive it unless `--verbose` is set; the field was already documented as omitted-when-empty. Frames whose absence is a real defect are handled explicitly instead of enumerated.
--- a/clients/cli/README.md
+++ b/clients/cli/README.md
@@ -769,7 +769,7 @@ explorer:https://etherscan.io/tx/0x9f8e7d6c...
 }
 ```
 
-`warnings` is omitted when empty. `PROTOCOL_DRIFT` means the backend emitted an SSE frame type this CLI version does not understand; the turn continues, but automation should retain the warning because a newer CLI may be required to handle that frame.
+`warnings` is omitted when empty, and is **`--verbose`-only**: `PROTOCOL_DRIFT` is a debugging aid, not a machine contract. The backend's V1 wire evolves forward-compatibly — unknown `data-*` card kinds are expected against a newer backend and are tolerated silently — so a warning emitted by default would fire on healthy turns. Run with `--verbose` to see which frame types a turn carried that this CLI does not route.
 
 Failures use the same v1 envelope with `success:false` and a stable `error.code`. This includes
 failed/declined signing and typed blocked/refusal/error turn endings; their partial turn data remains under `data`.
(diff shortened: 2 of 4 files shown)


gomesalexandre (approved):
Thanks for the thorough follow-up here - this is a clean, well-reasoned fix and the writeup made it easy to verify. Approving on f632279 with a runtime receipt (posted above), independently spot-verified against the diff.

Follow-up to #1305: the SSE idle deadline couldn't detect the hung backend it was added for, and three frame types were unhandled (two carrying turn-result semantics).

Verified the load-bearing behaviors:
- Hung-backend bound (H1): processLine only advances lastFrameAt on a completed non-empty data: frame; keep-alive comments and blank separators are consumed but return false, so they never defer the deadline. Fails CLOSED with a typed AgentStreamIdleTimeoutError. Crucially it can't be bypassed by heartbeat dribble - proven live (one real frame then `: ping` every 150ms forever -> honest timeout at ~923ms against a 900ms deadline).
- Outcome-bearing frames: tool-output-error -> deriveToolDoneOk returns false (v1Type === 'tool-output-error') so a failed tool reports done/ok:false instead of hanging as running-forever, AND maybeSignToolOutput returns early on that frame so it is never signed as if it were a valid tool output. text-replace -> id-keyed TextParts overwrite-in-place with a degrade path that declines (logs, leaves text as-streamed) rather than half-applying on unkeyed/unknown-target. No wrong-outcome path found in either.
- Non-regression: normal turn, blank-line/comment path, unknown-frame drift-under-verbose, and slow-but-progressing (survives past the deadline via real frames) all preserved.

Runtime: exercised against a real local HTTP server / real sockets / real timers driving the actual client.ts (not mocked fetch / fake timers). Hung + both outcome frames + non-regression all pass. Build/tsc/lint/prettier clean; CLI suite 816 tests pass, client.test.ts 74/74. Changeset patch for @vultisig/cli is correct. Merging on green.

PR #1346, opened 2026-07-17
fix(cli): classify permanent non-EVM broadcast errors

neavra:
## Summary

- dispatch permanent broadcast validation checks by SDK chain kind for UTXO, Solana, Sui, and Cosmos
- preserve the existing EVM classification and leave ambiguous or transient failures retryable
- cover permanent and transient RPC shapes per family and add an `@vultisig/cli` patch changeset

The classifier intentionally fails toward retryable when a message is ambiguous. A false non-retryable result can strand a user, while retries still have the broadcast journal and exit 9/13 safeguards.

## Validation

- `yarn build:all`
- `yarn cli:build && yarn cli:smoke`
- `yarn test:cli` (775 passed)
- `yarn typecheck`
- `yarn lint`
- `yarn test`
- `yarn prettier --config .config/.prettierrc --check clients/cli/src/core/errors.ts clients/cli/src/core/errors.test.ts .changeset/calm-wolves-retry.md`

No live transactions were attempted.

### DIFF
--- a/.changeset/calm-wolves-retry.md
+++ b/.changeset/calm-wolves-retry.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/cli': patch
+---
+
+Classify permanent UTXO, Solana, Sui, and Cosmos broadcast validation failures as non-retryable invalid input while leaving ambiguous and transient failures retryable.
--- a/clients/cli/src/core/errors.test.ts
+++ b/clients/cli/src/core/errors.test.ts
@@ -623,11 +623,59 @@ describe('anticipated CLI taxonomy regressions', () => {
       )
     })
 
+    it('keeps a Sui -32002 rejection retryable when the message is not a recognized permanent signal', () => {
+      // -32002 (TransactionExecutionClientError) is a broad bucket covering more than
+      // signature failures — an unrecognized message under that code must not be
+      // blanket-matched to permanent.
+      const rpcError = Object.assign(new Error('Insufficient gas budget for this transaction'), { code: -32002 })
+      expectTransient(
+        new VaultError(
+          VaultErrorCode.BroadcastFailed,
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
Thanks for this, and for the quick turnaround on the tightening. Approving on a30fd197 with a runtime receipt (posted above) plus an independent read of the diff.

This adds permanent-vs-retryable broadcast classification for the non-EVM families (UTXO/Solana/Sui/Cosmos), preserving EVM byte-for-byte, and correctly defaults ambiguous/unknown to retryable.

The one thing worth calling out: the first cut classified cosmos on "any non-zero ABCI code" and sui on the -32002 code alone, which would have newly stranded recoverable cases (a cosmos account-sequence-mismatch, code 32, and a generic -32002) that resolve on retry. The revision fixes both cleanly:
- cosmos: COSMOS_PERMANENT_SDK_CODES is a code -> canonical-message map, gated on codespace === "sdk" (so a wasm code-5 can't be mistaken for an sdk code-5) and requiring the message to match the code. Code 32 is deliberately excluded and now classifies retryable (the MPC-race sequence shape). Verified in the diff.
- sui: -32002 now requires the code AND a signature/malformed message match; a generic/unrecognized -32002 stays retryable. The required-fields guard (no RPC code) remains always-permanent, independent of the -32002 gate.

Double-broadcast safety holds below this layer: the SDK broadcast resolvers intercept already-known / already-in-cache / already-processed shapes with verifyBroadcastByHash before any BroadcastFailed reaches this classifier, so nothing here can trigger a resubmit of an already-landed tx.

Regression tests cover code-32 -> retryable, code-5 -> permanent, wasm-codespace code-5 -> retryable, and sui -32002 unrecognized -> retryable. errors.test.ts 85/85, full CLI suite 832/832, build/typecheck/lint clean. Merging on green.

PR #1355, opened 2026-07-17
fix(cli): only call a non-EVM broadcast failure permanent when the bytes are at fault

neavra:
## Summary

Follow-up to #1346, which added per-chain-family permanent-vs-retryable broadcast classification. That PR's design is right — the family split, the EVM preservation, and the fail-toward-retryable default all hold up. This corrects three places where the implementation drifts from that stated policy, all in the same direction: a recoverable failure was being declared permanent, or a match could never fire at all.

### 1. Cosmos: `insufficient funds` / `insufficient fee` / `unknown address` are not permanent

The allowlist comment claimed these are codes where "the identical bytes can never succeed regardless of retries or waiting." That is true for the bytes-intrinsic codes (tx parse error, invalid pubkey, tx too large, …) but not for three of them:

- **5 `insufficient funds`** — a CheckTx rejection never increments the account sequence, so once the account is funded the *identical signed bytes* land.
- **13 `insufficient fee`** — `minimum-gas-prices` is node-local config; another RPC accepts the same bytes.
- **9 `unknown address`** — the account materialises on first receipt.

All three now fall through to retryable. This also resolves the misleading-hint point raised in review on #1346: an insufficient-funds rejection no longer answers with "Check the signed transaction encoding and signature", because it is no longer classified as an encoding problem.

The remaining allowlist entries are unchanged, and the comment now states the actual rule (bytes-intrinsic vs state-dependent) rather than a claim that was untrue for these three.

### 2. UTXO: the `-26` / `-27` numeric match never fired, and `-26` is a bucket

Blockchair is the only UTXO broadcast backend on both the signed resolver and the raw path, and it does not preserve bitcoind's numeric reject code — it reformats the reply and drops it. Probed against the live endpoint:

```
$ curl -X POST https://api.blockchair.com/bitcoin/push/transaction -d "data=00"
{"data":null,"context":{"code":400,"source":"R",
 "error":"Invalid transaction. Error: TX decode failed. Make sure the tx has at least one input.", ...}}
```

No `-26`, no `RPC error` prefix. So `\b(?:RPC error|code|error code)\s*:?\s*-(?:26|27)\b` could not match in production, and the tests asserting it passed only because they fed a hand-built string straight into the classifier.

Removing it is also a latent-risk fix: `-26` is `RPC_TRANSACTION_REJECTED`, a bucket rather than a verdict. It covers `non-final` (locktime not yet reached), `too-long-mempool-chain` (ancestors unconfirmed) and `min relay fee not met` (mempool min fee falls as congestion clears) — all cases where the identical bytes succeed later. Had the code ever reached us, matching the bucket would have stranded them.

`TX decode failed` — the one reason we have actually observed — still classifies permanent, and the tests now use the captured response verbatim. Extending this list should start from a captured Blockchair response, not from bitcoind's wording, which never reaches us intact.

`-27` ("already in block chain") means the tx **landed**; there is now an explicit test that it is never reported as invalid input.

### 3. Chain detection reads the SDK wrapper, not any chain-labelled text

`getBroadcastErrorChain` scanned the whole error string (wrapper + `originalError.message`) for any ` on <chain>:` substring, first match by enum order — and EVM chains sort first. Solana folds program logs into the message, and a program controls its own `msg!` text, so a log line mentioning another chain could hand a Solana failure to the EVM predicate, whose vocabulary would match and wrongly strand it. Detection now matches the SDK's own wrapper phrasing on the error's own message.

## Test changes

- UTXO fixtures replaced with the live-captured Blockchair response; added recoverable-`-26`-reason and already-in-block-chain guards.
- Cosmos codes 5/9/13 pinned retryable.
- Two existing Cosmos negative tests were passing for the wrong reason — deleting the `codespace: sdk` gate, or dropping the code/log pairing requirement, left the suite green. Their fixtures now exercise the gate they document (mutation-checked: each mutant is killed).
- Added coverage for families with no classifier (Ripple/Tron/Polkadot) and for the wrapper-vs-payload chain detection.

## Validation

- `yarn build:all`
- `yarn test:cli` — **797/797** (53/53 files)
- `yarn workspace @vultisig/cli typecheck` — clean
- `eslint` + `prettier --check` on the touched files — clean
- Red proofs: re-adding codes 5/9/13 → 3 red; deleting the codespace gate → 1 red; dropping the code/log pairing → 1 red; restoring the unanchored chain scan (Sui fallback left intact) → 1 red. The UTXO production change is dead-code removal, so it has no red proof by construction — the evidence is the live probe above.

EVM classification and the exit 9/13 broadcast-safety semantics are untouched.

No live transactions were attempted. The only network call was the Blockchair probe above, with an undecodable payload that cannot be a transaction.

## Note for the reviewer

Changeset is `patch`, matching #1346's precedent, but this does change exit codes on paths a scripted consumer may branch on (a Cosmos insufficient-funds broadcast moves from exit 4 `retryable:false` to exit 6 `retryable:true`). Happy to make it `minor` if you'd rather.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CLI broadcast error classification across Cosmos, Bitcoin, Solana, Ripple, Tron, and Polkadot.
  - Recoverable funding, fee, account, and UTXO rejection errors are now correctly marked as retryable.
  - Permanent failures are identified more accurately, reducing cases where valid transactions are incorrectly stranded.
  - Improved chain detection prevents misleading error details from causing incorrect classifications.
  - Existing EVM error handling and broadcast-safety behavior remain unchanged.

### DIFF
--- a/.changeset/olive-donkeys-repeat.md
+++ b/.changeset/olive-donkeys-repeat.md
@@ -1,0 +1,11 @@
+---
+'@vultisig/cli': patch
+---
+
+Correct the non-EVM permanent broadcast classification so it only declares a failure permanent when the fault is in the signed bytes themselves.
+
+- Cosmos: `insufficient funds` (5), `insufficient fee` (13) and `unknown address` (9) are no longer permanent. A CheckTx rejection does not increment the account sequence, so these stay replayable verbatim once the account is funded, a cheaper node accepts the fee, or the account exists — they now classify as retryable (`EXTERNAL_SERVICE`, exit 6) instead of `INVALID_INPUT`, exit 4.
+- UTXO: dropped matching on bitcoind's `-26`/`-27` reject codes. Blockchair is the only UTXO broadcast backend and does not preserve those codes — it reformats the node's reply as `Invalid transaction. Error: <reason>` — so the branch never fired, and `-26` is a bucket that also covers recoverable rejections (`non-final`, `too-long-mempool-chain`, `min relay fee not met`) that must not be stranded.
+- Chain detection now reads the SDK's own wrapper rather than scanning the whole error text for any ` on <chain>:` substring, so chain-labelled text inside a wrapped payload (e.g. a Solana program log) can no longer route an error to another family's vocabulary.
+
+EVM classification and the exit 9/13 broadcast-safety semantics are unchanged.
--- a/clients/cli/src/core/errors.test.ts
+++ b/clients/cli/src/core/errors.test.ts
@@ -1,4 +1,4 @@
-import { VaultError, VaultErrorCode, VaultImportError, VaultImportErrorCode } from '@vultisig/sdk'
+import { Chain, VaultError, VaultErrorCode, VaultImportError, VaultImportErrorCode } from '@vultisig/sdk'
 import { describe, expect, it } from 'vitest'
 
 import {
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
## lgtm 🫡

fund-safety-consistent, and it does NOT undo the broadcast-false-success hardening - I checked that specifically. this touches ONLY the CLI error taxonomy (`errors.ts` + tests + changeset, 3 files); it does not go near the broadcast/confirm path, verifyBroadcastByHash, or any of the #1316/#1426/#1429/#1430/#1432/#1433 verification. it's a downstream classifier that decides, given a broadcast FAILURE, exit 4 (permanent/invalid-input) vs exit 6 (retryable).

the direction is the safe one. the dangerous mistake for this classifier is calling a landed/recoverable tx "permanent invalid input" - that's what pushes a user to rebuild and pay twice. this PR moves things toward retryable:

- cosmos codes 5/13/9 dropped from the permanent set: a CheckTx rejection never increments the account sequence, so insufficient-funds/fee/unknown-address stay replayable verbatim once state changes AND never entered a block (no double-spend). correctly retryable. `COSMOS_PERMANENT_SDK_CODES`.
- the code+log pairing is enforced (`isCosmosPermanentBroadcastInput`): permanent requires codespace==sdk AND code allowlisted AND the canonical message regex matches the log - so a mislabelled code or a log that contradicts its code stays retryable. tests pin both ("code/log contradiction" + "non-root codespace" go red if you drop either gate).
- UTXO: `-26`/`-27` matching dropped (Blockchair never surfaces the code, verified by probe - reformats as `Invalid transaction. Error: <reason>`), so only the observed `TX decode failed` stays permanent; `-26` recoverables (non-final / too-long-mempool-chain / min-relay-fee) and crucially `already in block chain` are now retryable. the already-confirmed case is the important one - test asserts a landed tx is NEVER called invalid-input.
- chain detection now reads the SDK's own wrapper (`failed to broadcast (?:raw )?transaction on (.+?):`, first match, on err.message only) instead of scanning the whole payload for any ` on <chain>:` - so a Solana program log saying "bridged on Ethereum: invalid signature" can't hand the error to the EVM predicate and wrongly strand it permanent. test pins it.

verified the fallbacks: no-chain and no-classifier both return retryable, and the code comments the philosophy explicitly ("ambiguity intentionally stays retryable ... a false retryable is constrained by broadcast-journal dedupe + exit 9/13 semantics"). that's exactly aligned with the hardening layer, not against it.

tests are thorough (cosmos 5/9/13, code/log mismatch, UTXO recoverables + already-confirmed, no-classifier families ripple/tron/polkadot, solana foreign-text). clean. money-path-ADJACENT (broadcast-failure exit codes) so worth your runtime glance, but no broadcast/confirm logic changed.

rcoderdev (approved):

PR #1376, opened 2026-07-17
test: bind rn and walletcore tx encoders to shared golden vectors

### LINKED ISSUE #1365: Encoder-parity test missing: RN pure-JS vs WalletCore never cross-checked (plan 003 unexecuted)
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P0 CONFIRMED.

**Anchor:** packages/sdk/src/platforms/react-native/chains/{evm,cosmos,solana,sui,ripple}/tx.ts, packages/sdk/tests/unit/platforms/react-native/tx-builder-golden-vectors.test.ts:44,64,107, packages/core/mpc/tx/compile/compileTx.golden.test.ts, plans/003-encoder-parity-golden-vectors.md

## what
The SDK ships two independent tx-encoder families for the same funds: an RN pure-JS path (packages/sdk/src/platforms/react-native/chains/*) and a WalletCore-WASM path (packages/core/mpc/tx/compile/compileTx.ts). In a 2-of-2 keysign, an RN device and an iOS/Android device on the same payload must compute a byte-identical pre-signing hash. No test asserts signingHashHex parity between the two families for EVM native, ERC-20, Solana, or Cosmos bank-send.

The one existing RN golden test asserts tx.signingHashHex against viem's serializeTransaction and against @solana/web3.js - the libraries the RN encoder is built on. It cannot catch RN-vs-WalletCore divergence, the actual cross-device risk (documented in plan 003 itself).

Plan 003 (this exact fix, fully scoped) exists at plans/003-encoder-parity-golden-vectors.md but is unmerged. A stacked PR (#1044) landed a Cosmos+Solana cross-encoder binding test, but it merged into the feat_golden_vector_gaps feature branch, not main - that branch was never itself merged, so the binding is stranded and absent from main today. EVM native/ERC-20 parity was never attempted (still only viem-vs-viem on both sides).

## failure scenario
A user signs an EVM ERC-20 transfer or Cosmos bank-send from an RN device, co-signing with a WalletCore-path peer. If the encoders diverge on gas-field ordering, EIP-1559-vs-legacy selection, or cosmos varint/fee-denom (same bug class as the shipped terra-classic gas-limit bug), the peers hash different bytes: the ceremony fails, or a peer signs a payload the user never saw. Nothing in CI catches this today.

## fix
Land plan 003's encoder-parity.test.ts (or resurrect PR #1044's cross-encoder work onto main), asserting byte-identical pre-signing hashes between RN pure-JS and WalletCore for EVM native, ERC-20 transfer, Solana send, and Cosmos bank-send, running in the default non-skipped test path.

gomesalexandre:
Closes #1365

## what

The SDK ships two independent tx-encoder families for the same funds: the RN pure-JS path (`packages/sdk/src/platforms/react-native/chains/*`) and the WalletCore-WASM path (`packages/core/mpc/tx/compile`). In a 2-of-2 keysign an RN device and an iOS/Android device must hash byte-identical pre-signing bytes, and until now NOTHING asserted they agree - each family only self-checked against the very library it is built on (viem vs viem, web3.js vs web3.js). This is the structural net that would have caught the RLP bug class.

This PR lands the full plan 003 parity net on main:

- resurrects the #1044 cosmos+solana cross-encoder binding, which merged into the `feat_golden_vector_gaps` feature branch and never reached main (stranded for ~2 months)
- adds the never-attempted EVM half: native transfer + ERC-20 transfer, each in BOTH legacy and eip1559 form

## why

Fund-safety P0 from the round-7 whole-stack audit. If the encoders diverge on gas-field ordering, legacy-vs-eip1559 selection, or calldata packing, the 2-of-2 peers hash different bytes: the ceremony fails, or a peer signs a payload the user never saw. Nothing in CI catches this today.

## how

Shared fixtures under `testdata/cross-encoder-golden/*.json`, read via plain `readFileSync` by BOTH suites (packages/sdk does not depend on packages/core, so a file read sidesteps the workspace dependency graph):

- `packages/core/mpc/tx/compile/compileTx.golden.test.ts` builds each tx via TW protos and asserts `getPreSigningHashes` output against the fixture hash (WalletCore family)
- `packages/sdk/tests/unit/platforms/react-native/tx-builder-golden-vectors.test.ts` + `cosmos-send-golden-vectors.test.ts` build the identical tx via the RN builders and assert `signingHashHex` against the SAME fixture hash (pure-JS family)

One editable value binds both suites - no keep-two-literals-in-sync drift risk. The ERC-20 vector additionally cross-checks calldata construction (viem `encodeFunctionData` vs WalletCore's internal `ERC20Transfer` proto encoding), and the RN chain-name -> chainId table is bound to the fixture chainId. Both fee variants are pinned on purpose: type selection / gas-field ordering is exactly the divergence class that splits a cross-device keysign.

All parity assertions PASS - no live divergence between the families today. Tests run in the default non-skipped path (no `skipIf`, no fixture gates).

## testing

Receipt A - deliberately-broken encoder goes RED (swapped `maxFeePerGas`/`maxPriorityFeePerGas` in the RN unsigned 1559 serialization, the classic field-swap bug):

```
-        maxFeePerGas: opts.maxFeePerGas!,
-        maxPriorityFeePerGas: opts.maxPriorityFeePerGas!,
+        maxFeePerGas: opts.maxPriorityFeePerGas!,
+        maxPriorityFeePerGas: opts.maxFeePerGas!,

 × cross-encoder binding > produces the SAME signing hash as WalletCore for the identical EVM native transfer (legacy + eip1559)
 × cross-encoder binding > produces the SAME signing hash as WalletCore for the identical ERC-20 transfer (legacy + eip1559)
 × EVM builders > pins native EIP-1559 send unsigned bytes and finalized yParity handling
      Tests  3 failed | 4 passed (7)
```

Receipt B - single shared-fixture perturbation (last nibble of the eip1559 expected hash flipped) fails BOTH independent suites, proving both are bound to the same source of truth:

```
--- RN pure-JS suite (packages/sdk, viem path) ---
 × produces the SAME signing hash as WalletCore for the identical EVM native transfer (legacy + eip1559)
      Tests  1 failed | 6 passed (7)

--- WalletCore/WASM suite (packages/core) ---
 × matches the shared cross-encoder golden vector for an EVM native transfer (legacy + eip1559)
      Tests  1 failed | 10 passed (11)
```

Mutations reverted, final state GREEN:

- `tx-builder-golden-vectors.test.ts`: 7/7
- `cosmos-send-golden-vectors.test.ts`: 6/6
- `compileTx.golden.test.ts`: 11/11
- broad `yarn workspace @vultisig/sdk test` and full `yarn test:core` diffed against origin/main at the same commit: identical pre-existing failure profile (build-order module-resolution noise), this branch adds +9 passing tests and zero new failures
- prettier (repo config) + eslint clean on changed files; `yarn typecheck` shows the exact same 4 pre-existing TS2307s as origin/main (unbuilt workspace packages), none from this diff

## risk

None to runtime - tests + testdata + plans readme only, zero `src/` changes. The net is additive and unskipped, so a future real divergence fails CI loudly instead of splitting a keysign in the field.

## adversarial review (opus, pre-human)

Verdict: no blocker. Key confirmations and the two notes that survived:

- tautology check: the WalletCore side is genuinely independent - `getPreSigningHashes` flows to `walletCore.TransactionCompiler.preImageHashes` (WASM RLP/keccak, zero viem). The RN side is viem. Real net.
- CI gating check: in a CI-built tree (`build:shared` + `build:all` then `yarn test`), `test:core` is FULLY green (1835/1835) - the 47 locally-failing files are pure build-order noise. Both suite halves sit in the root `yarn test` `&&` chain with no `continue-on-error`, so a future divergence fails CI loudly.
- fixture-integrity check: a single fixture value cannot mask a real divergence - if the encoders disagree, one value cannot equal both independent computations, so at least one suite goes red.
- note taken as a follow-up comment in both suites: the RN EVM assertion is viem-vs-viem (near-self-check), so the core suite's two EVM cross tests are the sole independent reference - now flagged in-code so nobody skips them as "redundant".
- cross-repo note for reviewers: this net guards the SDK's RN builder surface (vendored from vultiagent-app's evmTx). Worth confirming the app now consumes the SDK builders on its dispatch path - if it still uses an in-repo copy, that copy needs the same bind (follow-up, not this PR).

Coverage follow-ups deliberately left out (add-a-fixture class, non-blocking): chainId >2 bytes, nonce >=0x80 RLP length-prefix boundary, zero-value, >2^53 lamports, non-empty accessList.


neavra (approved):
lgtm — the vectors genuinely bind, and i reproduced both your receipts rather than taking them on faith. this is the right shape for the problem: two encoder families that only ever self-checked against the library they're built on, now refereed by one editable value.

**the binding is real, not two snapshots.** both suites `readFileSync` the same file and assert the same field, so i went at the thing that would actually distinguish binding from coincidence — receipt B. flipped one nibble of `evm-native-transfer.json`'s eip1559 hash (`...467000` → `...467001`), single edit, ran the two suites separately:

```
walletcore/wasm leg (packages/core):  × matches the shared cross-encoder golden vector for an EVM native transfer
                                        Tests  1 failed | 10 passed (11)
rn pure-js/viem leg (packages/sdk):   × produces the SAME signing hash as WalletCore for the identical EVM native transfer
                                        Tests  1 failed | 6 passed (7)
```

one fixture edit reds both independent suites. that's the property that matters and it holds.

receipt A reproduces exactly too — swapped `maxFeePerGas`/`maxPriorityFeePerGas` in the RN 1559 serialization (`react-native/chains/evm/tx.ts:210-211`) → `3 failed | 4 passed (7)`, the same three test names you quote. a real encoder bug gets caught.

also checked the thing that would quietly moot all of the above: no `skipIf` / `.skip` / `.todo` anywhere in the three suites, and `--reporter=verbose` shows the cross-encoder tests as ✓ rather than ↓. they actually run.

**provenance — your disclosure is accurate, and i verified the load-bearing half.** `0a7bb45b` says the EVM hashes were pinned via viem so the RN side is a near-self-check, and the independent teeth are the core suite. that holds: the core EVM cross tests build `TW.Ethereum.Proto.SigningInput` and hash through `getPreSigningHashes({ walletCore, ... })` — no viem in the hash path (the file imports viem, but only for the separate raw-legacy comparison). so walletcore's own WASM RLP really does recompute the viem-pinned hash. worth adding that the RN leg isn't a *pure* tautology either — receipt A proves it still catches RN-builder logic bugs (field mapping, type selection); it just shares viem's RLP.

and the resurrection is faithful, which i checked rather than assumed:

```
git diff fgvg:testdata/cross-encoder-golden/cosmos-msgsend.json   pr-1376-head:...  -> empty
git diff fgvg:testdata/cross-encoder-golden/solana-transfer.json  pr-1376-head:...  -> empty
```

byte-identical to #1044 on `feat_golden_vector_gaps` (merged 2026-07-08, branch still on the remote, never reached main — the "stranded ~2 months" story checks out). and that branch carries no EVM fixture, so "never-attempted EVM half" is accurate.

receipts: core golden 11/11, RN 13/13 (7 + 6). baseline-diffed both suites against origin/main in a separate worktree at the same commit — core 1835 vs 1831, sdk 3047 vs 3042, same 3 failing files / same 5 test names / byte-identical error on both sides. **zero new failures, +9 passing (4 core + 5 sdk) — your "+9" is exact.**

---

three non-blocking:

- **plan 003 is marked `DONE`, but ripple is bindable today and isn't bound.** only four RN builders expose `signingHashHex` (the shape this bind needs): cosmos, solana, evm — all three now bound — and **ripple**, which isn't, and which the core golden doesn't self-pin either. the rest (utxo, cardano, sui, ton, tron) don't expose it, and utxo is PSBT-only via `compileSignBitcoinTx` which the golden file explicitly excludes as structural — so their absence is defensible and ripple's is the odd one out. XRP is fund-bearing. your outcome note enumerates coverage honestly, so this is really just the status label: the file's own convention two lines down is `DONE (partial) — see note` (plan 005). either match it or bind ripple.
- **"build-order module-resolution noise" is the wrong diagnosis** for those 5 pre-existing sdk failures. there's no `Cannot find module` / `Failed to resolve` in either log — it's the MPC runtime singleton guard: `configureMpc: duplicate MPC engine instance ... VULTISIG_STRICT_SINGLETON=0` (vultisig-windows#3777). the substantive claim is true (pre-existing, identical on main, zero new), it's only the stated cause that's off — but anyone triaging those 5 off the description goes hunting the wrong problem.
- nit: `EVM_SEND_SIGNING_HASH` (:67) and `ERC20_TRANSFER_SIGNING_HASH` (:78) are byte-identical to the new fixtures' native-eip1559 / erc20-legacy hashes, for the same logical txs (params match: nonce 7, gasLimit 21000, 30/1.5 gwei, same value + recipient). both are pre-existing on main and reusing the already-vetted vector was the right call — but the fixture's own `_description` says "do not duplicate these values as literals in either test file", and that now reads as violated inside the file that loads it. fail-safe (editing the fixture reds the old self-check loudly rather than drifting silently), so it's tidy-up: point those two tests at `fx` and delete the literals, and the "one editable value" claim becomes true within the file too.

nice work on this one — the `readFileSync`-not-import trick to dodge the workspace dep graph is the detail that makes it actually land, and pinning both fee variants is the right instinct given type selection is exactly the class that splits a 2-of-2.

rcoderdev (approved):

NeOMakinG (approved):
## testing approach

Two lanes: Claude (load-bearing) + Codex `gpt-5.4` adversarial. Checked out HEAD `2d787ac9` in an isolated worktree; read both encoder families end-to-end (RN pure-JS/viem builders + WalletCore/WASM `getPreSigningHashes`), the 4 shared fixtures, and ran both suites — including live mutation testing to prove non-inertness.

## verdict

**Approve.** This is a genuinely non-inert cross-encoder parity net, not a tautology. Verified independently:

**Non-inert — proven by mutation on both sides:**
- RN pure-JS/viem suites (`tx-builder-golden-vectors.test.ts` + `cosmos-send-golden-vectors.test.ts`): 13/13 green; flipping one nibble of the shared EVM eip1559 hash **and** the cosmos SignDoc hash turned exactly those two cross-encoder assertions RED.
- WalletCore/WASM suite (`compileTx.golden.test.ts`, run via `.config/vitest.core.config.ts` with `core-config`/`lib-utils` dist staged): the 4 cross-encoder tests (EVM native, ERC-20, Solana, Cosmos) pass green; flipping the shared ERC-20 eip1559 hash turned the WASM-side assertion RED.

**Why it catches real divergence:** the two families each *independently* compute their pre-signing hash (WASM RLP/keccak vs viem for EVM; WASM vs hand-rolled cosmjs-types encode for Cosmos; WASM vs `buildSolanaSendTx` for Solana) and assert against ONE shared JSON literal. Both green simultaneously is only possible if the two encoders agree — a future divergence makes one side's computed hash differ from the shared value and fails CI loudly. The EVM RN side is viem-vs-viem (near-self-check, flagged in-code by the author); its independent teeth are the WASM assertions, which I ran green myself.

**Coverage:** fund-critical EVM native (legacy+1559), ERC-20 (legacy+1559, incl. calldata packing), Cosmos bank.MsgSend (+ derivation cross-check privkey→pubkey→address), Solana SystemProgram transfer. BTC/UTXO cross-checks live in the WASM suite / mobileFixtures elsewhere.

**Risk:** none to runtime — tests + testdata + plans README only, zero `src/` changes. Additive, unskipped, sits in the root `yarn test` `&&` chain.

## suggestion

- `suggestion:` (Codex MEDIUM, non-blocking) The EVM vectors use only 1-byte chainId (`1`, `137`) and small nonces (`7`, `3`), so they never exercise the RLP boundary classes most likely to fork an encoder: multi-byte chainId, `nonce >= 0x80` length-prefix transition, zero-value, non-empty accessList. The PR body already lists these as deliberate add-a-fixture follow-ups — noting for the tracked list, not blocking this PR.

---
Reviewed by Claude (Opus 4.8) + Codex gpt-5.4. Both encoder families executed locally; non-inertness proven by mutation on each.

PR #1378, opened 2026-07-17
fix(qbtc): honor field-7 gaslimit + golden-pin the fee/gas split

### LINKED ISSUE #1366: QBTC fee/gas conflation on the hand-rolled proto path, zero golden coverage
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P1 SUSPECTED.

**Anchor:** packages/core/mpc/chains/cosmos/qbtc/QBTCHelper.ts:97-127 (buildTxComponents) - packages/core/mpc/chains/cosmos/qbtc/QBTCTx.ts:4,22-39 (buildQBTCAuthInfo, defaultGasLimit)

## what
QBTC (BTC-backed asset, MLDSA keys) bypasses WalletCore on a manual proto encoder. buildTxComponents builds the fee coin as `{ denom, amount: cosmosSpecific.gas.toString() }` - using cosmosSpecific.gas as the fee coin AMOUNT - while buildQBTCAuthInfo separately defaults `gasLimit = 300_000n` and the caller never passes an explicit gasLimit override. So the gas limit is always the hardcoded constant, and cosmosSpecific.gas (a field named like a gas-limit value elsewhere in the cosmos-family resolvers) becomes the on-chain fee amount instead. No unit test drives getQBTCSignedTransaction end to end - the least-tested fund-moving encoder in the repo. Distinct from #1353 (QBTC broadcast SYNC-mode false-success), which is the broadcast layer, not the encode path.

## failure scenario
If cosmosSpecific.gas actually carries a gas-limit-shaped number (matching its name and its usage elsewhere), the fee amount charged on-chain is wrong: far too low (tx stuck/rejected) or an overpay, on every QBTC send/IBC transfer.

## fix
Add a golden fixture for QBTC MsgSend + IBC transfer pinning the fee/gasLimit split; confirm gas vs gasLimit semantics against a live node before trusting the current wiring.

Runtime confirmation needed: send a QBTC tx against a live/testnet node and diff the broadcast fee coin amount against the intended gas price x limit.

gomesalexandre:
Closes #1366

## what

sdk#1366 was filed as a **P1 SUSPECTED** possible-P0 fee/gas conflation on QBTC's hand-rolled Cosmos proto encoder - the fear being that `cosmosSpecific.gas` (a gas-limit-shaped value) was being written as the on-chain fee AMOUNT, over/under-paying on every QBTC send.

**Runtime-confirmed: NOT a P0. No fee drain.** The `CosmosSpecific` proto documents field 3 `gas` as "the fee AMOUNT (not a limit)" and field 7 `gasLimit` as the per-tx limit, and the WalletCore cosmos resolver splits them exactly that way (`Fee.amounts` = feeAmount derived from `gas`; `Fee.gas` = the limit). So QBTC's `fee.amount = cosmosSpecific.gas.toString()` is **correct** - it uses the fee amount as the fee amount.

The real, milder gap: QBTC hardcoded `gasLimit = 300_000n` in `buildQBTCAuthInfo` and never threaded field-7 `gasLimit`, so a QBTC tx that needs more headroom than the flat default (e.g. an IBC transfer - the standard cosmos path gives those extra gas via `IBC_GAS_MULTIPLIER`) could hit out-of-gas on-chain. Reliability, not fund-loss. Plus the encoder had **zero** golden coverage of the fee/gas split (the issue calls it out as the least-tested fund-moving encoder in the repo).

## how

- Thread `cosmosSpecific.gasLimit` (proto field 7) into `buildQBTCAuthInfo`. It already accepts an optional `gasLimit` defaulting to `300_000n`, so when field 7 is unset (the common case today - the caller does not populate it for QBTC) this is a **zero behavior change**; it only starts mattering when a simulated limit is supplied.
- Add `QBTCHelper.feeGas.test.ts`: golden vectors that decode the serialized tx's authInfo and pin (1) the fee coin denom+amount = `gas`, (2) the default 300000 gas limit when field 7 is unset, (3) field-7 `gasLimit` honored when set.

## testing

Red-first: against the unfixed encoder the field-7 test fails while the fee-amount test passes on BOTH old and new code - that's the P0 refutation:

```
 × honours field-7 gasLimit when set (a QBTC tx needing more headroom is no longer capped at 300000)
AssertionError: expected '300000' to be '550000'
 ✓ writes cosmosSpecific.gas as the fee AMOUNT (not the gas limit) for a MsgSend
```

After the threading fix: 3/3 green, full qbtc suite 5/5 green (5 files across the qbtc dir), root `tsc` clean, prettier + eslint clean.

## cross-device determinism note (for reviewers)

`gasLimit` is part of the SignDoc, so every co-signing device MUST derive it identically. This PR keeps QBTC on the same field-7-or-300000-default rule the WalletCore cosmos path already uses, and the change is a no-op while field 7 is unset. The latent consideration to weigh: if a non-SDK QBTC co-signer (e.g. an app-native MLDSA encoder) hardcodes 300000 and field 7 later gets populated, the two would derive different limits - so populating field 7 for QBTC must be a coordinated change across all co-signers. Flagging explicitly rather than burying it.

## risk

Low (P2). One-field threading with a default fallback (zero change today) + additive tests. The P0 hypothesis is refuted with evidence.

## Summary by CodeRabbit

* **Bug Fixes**
  * QBTC transactions now preserve a provided gas limit during signing instead of always using the default.
  * Transactions without a valid gas limit continue to use the `300000` default.
  * QBTC fee amounts remain unchanged and are kept separate from gas-limit values.

* **Tests**
  * Added coverage for default, custom, and zero gas-limit scenarios to verify transaction fee and gas encoding.


neavra (changes requested):
the P0 refutation holds and i verified it independently rather than taking it on faith — `gas` really is the fee amount and you really do write it as the fee amount. the threading is the right direction too. two things block, one of them cheap and one of them about the golden itself.

first, the good part, checked not assumed:

- **field-7 semantics are traceable to an authoritative source.** the generated `CosmosSpecific` (`blockchain_specific_pb.ts:204-213`) documents field 7 as the `/simulate` per-tx limit and closes with *"Note that field 3 `gas` is the fee AMOUNT (not a limit)"* — verbatim what your description claims. and `getQbtcChainSpecific` sets `gas: qbtcDefaultGas = 7500n`, which is fee-amount-shaped, not 300k-limit-shaped. no fee drain. P0 refuted.
- **red-proof reproduces your red-first output exactly.** reverted `gasLimit: cosmosSpecific.gasLimit` in isolation, kept your tests → `× honours field-7 gasLimit when set` / `AssertionError: expected '300000' to be '550000'`, other 4 green. one genuine killer, and the fee-amount test passing on both old and new code is the refutation, as you say.
- qbtc suite 5/5 at head (matches your number). wider cosmos sweep (`mpc/chains/cosmos/`, `chain/chains/cosmos/`, `chainSpecific/resolvers/cosmos/`, `keysign/fee/`) → 422 passed / 41 files. root `tsc` exit 0, prettier + eslint clean. all your claims confirmed.

**blocking: the golden pins a split your own cited authority contradicts.**

the docstring says the split is "exactly how the WalletCore cosmos resolver splits them (Fee.amounts = feeAmount from gas; Fee.gas = the limit)". that's half the rule. `resolveCosmosGasFee.ts:58-63` — the shared helper the signing-inputs *and* fee-display resolvers both use — couples the two:

```ts
const resolvedGasLimit = relayedGasLimit && relayedGasLimit > 0n ? relayedGasLimit : effectiveStaticGasLimit
const feeAmount =
  resolvedGasLimit > effectiveStaticGasLimit
    ? (effectiveGas * resolvedGasLimit + effectiveStaticGasLimit - 1n) / effectiveStaticGasLimit  // scales the fee
    : effectiveGas
```

its own docstring: *"the fee amount is scaled proportionally with a ceiling ... **so the tx pays for the extra gas**."* the point of that coupling is to hold the **gas price** constant when the limit moves. probed the real `getQBTCSignedTransaction` and decoded the authInfo, with the live `qbtcDefaultGas = 7500`:

| | fee.amount | gas_wanted | gas price |
|---|---|---|---|
| today (field 7 unset) | 7500 | 300000 | 0.025 |
| field 7 = 550000, this PR | 7500 | 550000 | **0.013636** |
| field 7 = 550000, walletcore rule | 13750 | 550000 | 0.025 (held) |

on the exact vector test 3 pins (`gas: 2500n, gasLimit: 550000n`) the walletcore rule encodes `4584`; you sign `2500` — 45.5% under, against the rule you cite as the model.

to be fair, and i checked hard: **this is inert today and fail-safe if it weren't.** nothing populates field 7 for QBTC — the only writer into a `CosmosSpecific` is `chainSpecific/resolvers/cosmos/index.ts:92,106`, typed to `IbcEnabledCosmosChain`, and QBTC is `OtherChain.QBTC` (`Chain.ts:77`), structurally outside the `CosmosChain` union. `resolvers/qbtc.ts` never sets it. the only field-7 population in the repo is your own test. and if it were set, `fee.amount` is the flat authorized charge — a bigger `gas_wanted` authorizes no extra spend, so an underpriced tx dies at CheckTx, which your `BROADCAST_MODE_SYNC` resolver does surface (`qbtc.ts:36` throws on `code !== 0`). no fund-loss path. no display drift either — `getQbtcFeeAmount` returns raw `gas`, matching the unscaled signed fee.

so why block on it: the golden *is* this PR's deliverable, and line 137 asserts

```ts
expect(fee.amount).toBe('2500') // fee amount still tracks gas, independent of the limit
```

"fee amount is independent of the limit" is the exact opposite of `resolveCosmosGasFee`'s rule, and it's now a green test. whoever populates field 7 next reads that and concludes the fee needs no attention. cheap either way: call `resolveCosmosGasFee({ gas, relayedGasLimit: cosmosSpecific.gasLimit, staticGasLimit: 300_000n })`, or keep the flat fee and narrow the docstring to say field 7 is honored as the limit while QBTC's fee is *deliberately* flat, and why. happy with either — just not with the equivalence claim standing.

**blocking: no changeset.**

`QBTCHelper.ts` is under `packages/core/mpc/` → `@vultisig/core-mpc` v1.12.0, not private, ships to npm. CLAUDE.md:91 is unambiguous ("No exceptions") and :114 covers this case by name ("Even one-line fixes need a `patch` changeset"). the test-only exemption (:118) lists `packages/sdk/tests/`, `.config/`, `clients/cli/tests/` — not core production code. direct precedent: #1273 (`e3d8568a`) shipped `.changeset/cosmos-broadcast-timeout-pending.md` for the same class.

---

non-blocking:

- **the gap #1366 describes stays open.** QBTC IBC transfers (`transactionType === 3` → `buildIBCTransferAny`) still get the flat 300000 + flat 7500, because nothing populates field 7 and the QBTC path has no `IBC_GAS_MULTIPLIER` equivalent — the standard cosmos path gives IBC 2× limit *and* 2× fee automatically off `transactionType`, no field 7 required. your body is honest that this is a no-op today, so this is a title quibble more than anything: `fix(qbtc)` reads like the OOG risk is closed, and it's groundwork toward closing it.
- **mixed-version co-signers, the direction your note doesn't cover.** your determinism note is about a non-SDK co-signer hardcoding 300000. the nearer one is SDK-vs-SDK: `compileTx.ts:52-56` reads `cosmosSpecific` straight off the deserialized wire `txInputData`, so field 7 is wire-reachable. post-merge, a peer on this version honors a relayed field 7 while a peer on the previous version ignores it → different SignDoc → sig fails. only bites once field 7 is populated, which just reinforces that populating it has to be a coordinated release.
- **the OOG you're guarding against is the one QBTC can't report truthfully** (pre-existing, not yours): `BROADCAST_MODE_SYNC` only surfaces CheckTx, so a DeliverTx out-of-gas comes back `code: 0` → reported as a successful broadcast, fee burned, funds unmoved. same class as #1316/#1346. worth its own issue.

nit, no change needed: `Chain.ts:77` calls QBTC a post-quantum **testnet**, which de-escalates this whole fee posture — worth a word in the golden so the next reader doesn't re-run the P0 fire drill.

lgtm once the changeset lands and the golden either scales the fee or stops claiming walletcore equivalence — the threading itself is right.

NeOMakinG (approved):
## testing approach

Two-lane review (Claude load-bearing + Codex gpt-5.4) against a dedicated worktree at HEAD. Read the actual encoder (`QBTCHelper.ts`, `QBTCTx.buildQBTCAuthInfo`), the golden test, the shared `resolveCosmosGasFee`, the QBTC chain-specific + fee-display resolvers, the `CosmosSpecific` proto, and the fail-safe broadcast path — not just the diff. Verified the prior `CHANGES_REQUESTED` blockers against source, then **fixed the remaining findings in-branch** (commit `1d316f27`), re-ran the qbtc suite, and proved red-on-revert for the new guard.

The P0 refutation holds and I re-verified it independently: proto field 3 `gas` is documented as the fee AMOUNT, field 7 `gas_limit` is the per-tx limit, and the encoder writes `fee.amount = cosmosSpecific.gas` — fee-as-fee, no drain. Field-7 threading is the right direction and is a genuine no-op today (nothing populates field 7 for QBTC).

Adversarial focus checks:
- **field-7 correctness / no unit-mismatch**: `gas_limit` is `optional uint64`, encoded via `protoVarint(2, ...)` into `Fee.gas_limit`. Cosmos gas is a plain integer count (no wei/gwei), bigint varint (no truncation). Correct.
- **golden meaningful, not tautological**: the fee/gas assertions run through the real `cosmjs AuthInfo.decode`; reverting `gasLimit: cosmosSpecific.gasLimit` flips test 3 red (`300000 != 550000`). Genuinely locks the split.
- **critically — is the "unset → 300000" default the *inert-guard* class (PR #228)?** No. Field 7 is `optional` (explicit presence), so an unset limit round-trips through wire serialization as `undefined`, not proto3's `0n`. The default fires on the real path, not just the in-memory test object. Guard is live.
- **fund direction**: `fee.amount` is a flat authorized spend; a larger `gas_wanted` authorizes no extra outflow, and an under-limit tx dies at CheckTx (`BROADCAST_MODE_SYNC` surfaces `code != 0`). Fail-safe — no overpay, no drain.

## prior findings — status (neavra CHANGES_REQUESTED @ `86b09ebf` + Codex lane)

| src | severity | issue | status | notes |
|---|---|---|---|---|
| neavra | blocker | golden pins a split its own cited authority (`resolveCosmosGasFee`) contradicts — comment "fee amount independent of the limit" reads as a general principle | ✅ fixed `1d316f27` | took neavra's second offered path: kept QBTC's fee deliberately **flat** and rewrote the docstring/comment to say field 7 is honored as the LIMIT only, QBTC bypasses the shared resolver, display returns raw `gas` (shown==signed), and it's a testnet. No more equivalence claim. |
| neavra | blocker | no changeset for `@vultisig/core-mpc` (ships to npm) | ✅ fixed `1d316f27` | added `.changeset/qbtc-field7-gaslimit.md` (`patch`), matching the `#1273` precedent. |
| Codex | should-fix | explicit `gasLimit: 0n` bypasses the `undefined`-only default; `protoVarint(2, 0n)` **elides** the field → invalid zero-gas tx once a field-7 producer lands | ✅ fixed `1d316f27` | field 7 being `optional` makes `0n` distinct from unset, so this is reachable. `buildQBTCAuthInfo` now coerces non-positive (`undefined`/`0n`/negative) → `300000`, mirroring `resolveCosmosGasFee`'s `> 0n` guard. New golden vector, **red-on-revert proven** (`'0'` vs `'300000'`). |
| neavra + Codex | non-block (converged) | mixed-version co-signer determinism once field 7 is populated (new SDK honors it, old hardcodes 300000 → SignDoc hash divergence → quorum fail) | ℹ️ documented | inert until a producer populates field 7; the code comment already flags the lockstep requirement. Correctly a coordinated-rollout concern, not a merge blocker. |
| neavra | non-block | the #1366 IBC-OOG gap stays open (no QBTC `IBC_GAS_MULTIPLIER` equivalent; field 7 unpopulated) | ℹ️ documented | groundwork PR; body is honest that it's a no-op today. |
| neavra | non-block | `BROADCAST_MODE_SYNC` can't surface a DeliverTx OOG (`code:0`) — pre-existing, same class as #1316/#1346 | ⏭️ deferred | pre-existing, not this PR; worth its own issue. |
| Codex | suggestion | test-only `extractProtoField` helper isn't a general proto reader (1-byte tag, wire-2 only) | ℹ️ acknowledged | correct for the current field-2 (`≤15`, wire-2) extraction, and the real assertions go through `cosmjs AuthInfo.decode`, so the golden is not tautological. Left as-is; not worth churn. |

## risk

Low. Net change over the PR: two blockers + one should-fix resolved with a defensive `> 0n` coercion that is byte-identical for the real path today (QBTC never sets field 7 → `undefined` → `300000`) and for the claim helper (passes no `gasLimit`). The only behavior change is the previously-broken explicit-`0n`/negative case, now fail-safe. All edits are comments/docstrings, a one-line guard, an additive golden vector, and a changeset. qbtc suite green; guard red-on-revert proven; prettier clean.

## verdict

APPROVE. P0 refuted with evidence, field-7 threading correct and fail-safe, golden is meaningful and now honest about QBTC's deliberately-flat fee, the `0n` encoder footgun is closed, and the changeset lands. Fixes pushed to the branch as `1d316f27`.

---
🤖 vultisig-ops reviewer (Claude + Codex two-lane)

rcoderdev (approved):

neavra (approved):
lgtm — re-reviewed at `da3c9e3d` against my `86b09ebf` review. Both blockers addressed:

- **Golden false-equivalence docstring (was blocking)** — FIXED. The docstring no longer claims `resolveCosmosGasFee` equivalence; it now states QBTC's fee is **deliberately FLAT** (`fee.amount == cosmosSpecific.gas`, field 7 threads the LIMIT only, no fee scaling), and the old `// fee amount ... independent of the limit` assertion comment is replaced with `// Field 7 is honoured as the LIMIT only - it must not silently re-scale the fee`. A future field-7 populator now reads "flat by design", not "fee needs no attention." `QBTCHelper.ts` carries the matching NOTE that display returns raw gas so shown==signed.
- **No changeset (was blocking)** — FIXED. `.changeset/qbtc-field7-gaslimit.md` (`@vultisig/core-mpc` patch).

Also verified the extra hardening you added in `1d316f27` — `buildQBTCAuthInfo` now coerces a non-positive field-7 limit via `gasLimit && gasLimit > 0n ? gasLimit : defaultGasLimit`. Confirmed independently: `protoVarint(_, 0n)` returns empty bytes (`protoEncoding.ts:119`), so an explicit relayed `0n` would have elided `gas_limit` and signed an invalid zero-gas tx — the guard is correct and the honest path (unset → 300000, or a real limit) is unchanged. Red-proof: reverting the guard to `gasLimit ?? defaultGasLimit` fails exactly the new `explicit 0` golden vector (`gasLimit '' ≠ '300000'`); restored clean. qbtc suite 6/6 green at head, full CI green.

The P0 refutation still holds (`gas` = fee amount, not a limit). Your non-blocking items remain open as separate issues (the IBC OOG under-reporting via `BROADCAST_MODE_SYNC`; mixed-version co-signer coordination when field 7 eventually gets populated) — fair to scope out here.

PR #1379, opened 2026-07-17
fix(swap): re-check thorchain trading-halt flags at broadcast

### LINKED ISSUE #1360: Native-swap broadcast guard never re-checks THORChain/Maya halt state
Parent epic: #1356

From the whole-stack audit round 7 (2026-07-17). Severity P2 CONFIRMED.

**Anchor:** packages/sdk/src/vault/services/nativeSwapBroadcastGuard.ts:83-99, packages/core/chain/chains/cosmos/thor/getThorchainInboundAddress.ts:9-15

## what
assertNativeSwapReadyForBroadcast matches the inbound address and checks it equals the vault address, but never reads halted / global_trading_paused / chain_trading_paused, present on the same fetched ThorchainInboundAddress object. Halt is checked at quote time (findSwapQuote.ts via getNativeSwapTradingHalt) but not re-checked at broadcast. The app-side equivalent (thorchainDispatchValidators.ts) has the same gap.

## failure scenario
THORChain halts a chain (mimir HALT<CHAIN>TRADING, churn, ragnarok) between quote and broadcast. The vault address is still current so the guard passes and the deposit broadcasts into a halted chain - best case a delayed refund minus outbound fee, worst case (mid-migration) stuck funds.

## fix
Also fail closed on activeInbound.halted || global_trading_paused || chain_trading_paused inside assertNativeSwapReadyForBroadcast. No halted-vault case in the guard's test today. Distinct from closed #696/#1060 (quote-time halt check only) and #934/#1135 (MayaChain vault+expiry-zero fix only) - neither covers this broadcast-time halt gap.

gomesalexandre:
Closes #1360

## what

`assertNativeSwapReadyForBroadcast` matched the inbound vault address but never read `halted` / `global_trading_paused` / `chain_trading_paused` - all present on the SAME fetched `ThorchainInboundAddress` object. Trading halt is checked at quote time (`findSwapQuote` via `getNativeSwapTradingHalt`) but not re-checked at broadcast.

## why

THORChain can halt a chain (mimir `HALT<CHAIN>TRADING`, churn, ragnarok) between the quote and the broadcast. The vault address stays current, so the address check passes and the deposit broadcasts into a halted chain: best case a delayed refund minus outbound fee, worst case (mid-migration) stuck funds.

## how

Capture the full active inbound object (not just `.address`) and fail closed if any of the three halt flags is set, before the address check. Zero extra network cost - the flags ride the already-fetched object. Covers THOR + Maya. The app-side equivalent (`thorchainDispatchValidators.ts`) has the same gap and is tracked separately.

## testing

Red-first: 4 new halt cases (halted / global_trading_paused / chain_trading_paused on THOR, halted on Maya) fail against the unfixed guard - the address still matches so nothing else catches it:

```
 × rejects a THORChain broadcast when halted even if the inbound vault address still matches
 × rejects a THORChain broadcast when global_trading_paused ...
 × rejects a THORChain broadcast when chain_trading_paused ...
 × rejects a MayaChain broadcast into a halted source chain
      Tests  4 failed | 13 passed (17)
```

After the fix: 17/17 green. Root `tsc` clean, prettier + eslint clean.

## risk

Low fund-safety-positive. Additive fail-closed check on a field already fetched; existing address-match and expiry paths unchanged.


neavra (changes requested):
reviewed at `bc792e50`. red-proofed the tests and checked the flags against both live feeds rather than against the type, since this is fund-safety and the whole fix hinges on what thornode/mayanode actually publish.

the core of this is right, and the field selection is more carefully chosen than the diff lets on — so let me start by putting the evidence for that on the record, because it's the part a future reader will most want to second-guess.

**the three-field subset is correct, and the fourth field would have been a kill switch.** live thor feed right now:

```
CHAIN    halted  global  chain_tp  lp_p
AVAX     False   False   False     True
BASE     True    False   True      True
BTC      False   False   False     True
ETH      False   False   False     True
SOL      True    False   True      True
...      (12 entries, lp_p True on ALL of them)
```

`chain_lp_actions_paused` is true on **every** chain including healthy BTC/ETH — live mimir has a global `PAUSELP = 1`. so including it would have fail-closed every native swap. excluding it is not a judgement call, it's necessary, and it matches `getNativeSwapTradingHalt`'s own note (*"Intentionally ignore chain_lp_actions_paused; LP action pauses do not block swaps."*). only nit: the new comment doesn't point at that precedent, so a reader here can't tell the 4th field was deliberately dropped rather than missed — one clause would fix that.

also: mimir today has `HALTBASETRADING=1`, `HALTBSCTRADING=1`, `HALTSOLTRADING=1`. your threat model isn't hypothetical, it's the current state of the network.

two blocking.

**blocking: the re-check mirrors the quote-time check's FIELD selection exactly, but not its CHAIN selection — the destination is never re-checked.**

quote-time evaluates both ends:

```ts
// getNativeSwapTradingHalt.ts
const chainIds = withoutDuplicates([chainIdFor(from), chainIdFor(to)].filter(...))
for (const chainId of chainIds) { ...halted / chain_trading_paused... }
```

this guard only ever resolves `sourceChainId = getNativeSwapChainId(chain)` and looks up that one entry. so a BTC→SOL swap quoted while SOL was healthy, with SOL halting before broadcast, sails straight through: BTC is healthy, the BTC vault address still matches, the deposit lands, and the outbound can't leave. that's your own comment's failure mode ("best case a delayed refund minus outbound fee, worst case stuck funds") — just through the other end.

live-reachable, not theoretical: `Chain.Solana`/`Chain.Base`/`Chain.BSC` are all in `thorChainSwapEnabledChains` and all three are halted **right now**. proved it, 2/2:

- CONTROL — source BTC `halted: true` → guard fires (`/trading is halted/`).
- GAP — source BTC healthy, destination SOL `halted: true, chain_trading_paused: true` → **resolves, no throw**, broadcast proceeds into the halted destination.

and it's cheap: `toCoin` is on the payload (`THORChainSwapPayload` field 3), `nativeSwapQuoteToSwapPayload` always populates it (`toCoin: toCommCoin(toCoin)`), and the guard's `native` is `Omit<THORChainSwapPayload,'$typeName'> & {chain}` — so `native.toCoin?.chain` is already in your hand. copy quote-time's tolerance while you're there (`chainIdFor` → null and `byChain.get(id)` → `continue`), so BTC→RUNE — where thor's own feed has no THOR entry — skips instead of false-blocking.

**blocking: no changeset.** touches `packages/sdk/src/`, so it ships to `@vultisig/sdk`, and CLAUDE.md:91 is unambiguous ("No exceptions"). precedent #1316. `@vultisig/sdk`: patch.

---

**non-blocking: the same halt is truthful at quote time and mis-tagged at broadcast.**

- quote time: `SwapError(SwapErrorCode.TradingHalted, "This swap route is temporarily unavailable — … Please try again later.")` — dedicated code, explicitly temporary.
- here: `VaultError(InvalidConfig, "… trading is halted …")`. the CLI's `classifyVaultError` has no arm matching that message, so `InvalidConfig` falls through to `new UsageError(...)` → `ExitCode.USAGE` = *"Usage error (bad arguments, unknown command)"*, and `retryable` inherits the `VsigError` default **false**.

so a transient chain halt reaches an agent as a permanent, non-retryable **usage** error. the prose is honest; the machine-readable outcome says the caller passed bad arguments and shouldn't retry — when in fact they did nothing wrong and it clears itself. to be fair the expiry throws above it already use `InvalidConfig`→`UsageError`, and for expiry that's defensible (refreshing the quote IS a caller action) — a halt is different in kind: you wait. cheapest fix is a `classifyVaultError` arm, or reuse `TradingHalted`.

**non-blocking: the type says these flags are required; maya doesn't send them.** `ThorchainInboundAddress` declares all four as required `boolean` and `queryUrl` does no runtime validation. live mayanode (9 entries) omits `global_trading_paused` and `chain_lp_actions_paused` entirely, and omits `chain_trading_paused` **except** on halted XRD — go `omitempty`, absent means false. your truthiness check (`a || b || c`) handles that **correctly**, so maya works today and the maya test is right. but the type lies about maya's payload: the next person who "hardens" this to `x !== false`, or adds a required-field assert, silently breaks maya with no test to catch it. worth an optional type or a comment.

nit: `global_trading_paused` is read off the source entry only, where quote-time uses `inbound.some(...)`. equivalent in practice (thor stamps it on every entry — verified, all 12 `False`), but the `.some()` form doesn't rely on that holding.

---

what i checked and found solid:

- **red-proof: 2/2 mutants killed, no survivors.** baseline 17/17. deleting the halt block → exactly your 4 new tests fail (`promise resolved "undefined" instead of rejecting`), 13 pre-existing stay green. weakening to `if (activeInbound.halted)` only → exactly the `global_trading_paused` and `chain_trading_paused` cases fail. that second one is the one i cared about: it proves both paused flags are **independently** pinned rather than riding on `halted`. the tests are honest.
- **fail-closed ordering is right.** the guard is the *first* statement in `broadcastTx`, before compile/broadcast and before the `txInputsArray` loop — so a halt refusal sends nothing, journals nothing, and moves no funds.
- **zero extra network cost, as you claim** — the flags ride the same already-fetched inbound object the address check reads.
- **maya genuinely gains here, and that's underclaimed.** `getNativeSwapTradingHalt` returns `null` unless `swapChain === Chain.THORChain`, so maya has *no* quote-time halt check at all — this guard is maya's only halt check anywhere on the path (and maya has XRD halted live). that's new protection, not a mirror.
- field names match both live payloads exactly; no false-block of any healthy chain on either feed.
- full sdk unit suite: **2867 passed** (189 files), 0 failures. `tsc --noEmit` clean. CI green at head, mergeable, 0 behind main.

one more thing worth flagging, since it cuts across: app #1995 cites *this* guard as precedent for its own live-inbound-vault bind — and at the time of that citation this guard didn't filter halted at all, so the precedent was only half real. with this landed it becomes real, which is a nice side effect. but the app side still selects its inbound entry without reading any of these flags, so #1995 is still the other half of this story — landing this doesn't cover it.

lgtm once the changeset lands and the destination gets the same look the source does.

NeOMakinG (approved):
## testing approach

Worktree at HEAD `bc792e50`, symlinked deps, ran the guard suite: **17/17 green**. **Red-on-revert proven** — neutralizing the halt condition (`if (false && ...)`) fails exactly the 4 new cases (halted / global_trading_paused / chain_trading_paused on THOR + halted on Maya) and passes the other 13, so the tests are genuine, not inert. Traced the actual broadcast + halt-check source (not just the diff), plus a Codex gpt-5.4 adversarial second lane.

Adversarially verified the four fund-safety invariants and all resolved positive:

- **fresh re-fetch at broadcast** — the guard's default `getInboundAddresses` calls raw `getThorchainInboundAddress()` → `queryUrl()`, a live fetch with **no cache**. The 5s `inboundCache` lives only in the quote-time `getNativeSwapTradingHalt.ts`, which this guard does not use — so the broadcast reads fresh halt state, not stale quote-time state.
- **fail-closed** — `assertNativeSwapReadyForBroadcast` runs as the first `await` in `BroadcastService.broadcastTx`, before any compile/broadcast, inside a try/catch that re-throws as `BroadcastFailed` (no swallow-to-success). `queryUrl` has a 20s abort timeout and the guard does not catch, so a hung/erroring inbound fetch aborts the broadcast. This is actually **stricter than quote time**, which catches and returns `null` (fail-open) — correct direction for a broadcast-time gate.
- **field parity** — `halted` / `global_trading_paused` / `chain_trading_paused` are all declared `boolean` on `ThorchainInboundAddress`, and are the exact same three flags the quote-time helper reads (both intentionally ignore `chain_lp_actions_paused`). THOR + Maya covered.
- **no encoder/golden-vector regression** — the change only reads extra fields off an already-fetched object; the address-match logic is byte-identical and the `?.address` null-guard is preserved.

Clean, additive, fail-closed, fund-safety-positive. Approving.

## should-fix

_Non-blocking — all pre-existing / contract-level, none introduced or worsened by this PR._

- **multi-input residual window**: the guard runs once before `broadcastTx`'s broadcast loop, which for an ERC20 THOR/Maya route can emit `approve` then `deposit` back-to-back. A halt flipping in the sub-second gap between the two still slips through. This is a pre-existing property of the multi-tx loop and the PR strictly improves on the prior state (zero broadcast-time check). A follow-up could re-check immediately before the final deposit tx.
- **no runtime schema validation**: `queryUrl` does a bare `response.json()`, so if Thornode/Mayanode ever renames or omits one of the three flags, `undefined` is falsy and that flag's check silently fails open. Same contract the entire existing codebase (quote-time helper + the address field itself) already relies on, so coercing only here would be inconsistent — noting for a codebase-wide hardening pass.

## q

- `VaultBase.broadcastRawTx` does not run this guard — but it takes `{chain, rawTx}` (a hex-signed external tx) with no keysignPayload, so it structurally cannot detect a native swap. It's a separate primitive for externally-built txs, not the MPC native-swap dispatch path the app uses, so this is out of scope rather than a bypass.

## risk

Low, fund-safety-positive. Additive fail-closed check on a field already fetched; expiry, address-match and secured-asset paths unchanged. No wire/encoder change.

## verdict

**APPROVE.** Re-check is genuinely fresh at broadcast, fail-closed on error/timeout, field-parity with quote time, THOR+Maya covered, tests red-on-revert. Codex's three findings are all non-blocking pre-existing items. The app-side `thorchainDispatchValidators.ts` equivalent gap is tracked separately per the PR body.

---
_Reviewed by Claude (load-bearing) + Codex gpt-5.4 adversarial lane. Two-lane._

rcoderdev (approved):

neavra (approved):
lgtm — re-reviewed at `8b769ee2` against my `bc792e50` review. Both blockers addressed:

- **Destination chain never re-checked (was blocking — the stuck-funds gap)** — FIXED. `assertNativeSwapReadyForBroadcast` now evaluates BOTH ends of the route: `withoutDuplicates([sourceChainId, destinationChainId])` where `destinationChainId = getNativeSwapChainId(native.toCoin?.chain)`, and throws if `halted | global_trading_paused | chain_trading_paused` on either resolved end — mirroring quote-time `getNativeSwapTradingHalt`'s `[from, to]` selection. My original BTC→SOL proof now rejects: source healthy + destination SOL halted → `trading is halted for SOL`. Red-proof: dropping `destinationChainId` from the route list fails exactly the new "DESTINATION chain is halted (source healthy)" test; restored clean.
- **No changeset (was blocking)** — FIXED. `.changeset/thor-halt-broadcast-recheck.md` (`@vultisig/sdk` patch).

Verified the tolerance is right: `getNativeSwapChainId` is a plain map lookup returning `undefined` for an unmapped chain (no throw), and the loop `continue`s when a leg has no inbound entry — so RUNE/CACAO and any destination absent from the feed are skipped, not false-blocked. `native.toCoin?.chain` is safely optional-chained (missing → degrades to source-only, no crash), and this correctly extends the check to Maya swaps too (`toCoin` rides the same proto field on both payload cases). Also took the `global_trading_paused` nit — now read across all entries via `inboundAddresses.some(...)`, red-proofed (reading it off the source entry only fails the new "global paused on a non-source entry" test).

Guard suite 21/21 green at head (4 new tests, 2 genuinely red-on-revert + 2 false-block guards), full CI green. Your non-blocking items stand as separate issues: the halt reaching the CLI as a non-retryable `USAGE` error (`classifyVaultError` arm / reuse `TradingHalted`); the multi-input residual halt window in the broadcast loop; and the `ThorchainInboundAddress` type asserting Maya-absent fields as required.

NeOMakinG (approved):
Re-review of the delta since my r1 APPROVE at `bc792e50`. HEAD is now `d6163d5d`; the fund-relevant delta is one commit — `8b769ee2` (fix: check BOTH route ends + add changeset), addressing neavra's two blockers. The two later commits (`2e426593`, `d6163d5d`) are pure main-merges — `git diff 8b769ee2 d6163d5d -- nativeSwapBroadcastGuard.ts` is **empty**, so the guard at HEAD is byte-identical to what neavra re-approved at `8b769ee2`.

## testing approach
Worktree at HEAD `d6163d5d`, symlinked deps, ran the guard suite: **21/21 green** (17 from r1 + 4 new). Guard imports/compiles/runs at HEAD (merge-from-main didn't break its dep graph). Confirmed chain-selection parity against quote-time `getNativeSwapTradingHalt` (same `nativeSwapChainIds` map) and inspected the map keys directly.

**Red-on-revert proven on the delta:** neutralizing the new destination+global-across-entries logic back to source-only fails **exactly** the 2 new enforcement cases (`rejects … DESTINATION chain is halted`, `rejects when global_trading_paused … not only the source`) while the 2 new allow-cases stay green — the new logic is non-inert, not PR#228-class.

## prior issues — status
|---|---|---|---|---|---|
| r1 | 1 | should-fix (Codex) | multi-input approve+deposit residual window (halt flips between approve tx and deposit tx) | ⚠️ unchanged | pre-existing structural property of the broadcast loop; delta doesn't touch it; PR still strictly improves |
| r1 | 2 | should-fix (Codex) | no runtime schema validation → `undefined` flag fails open on renamed/omitted field | ⚠️ unchanged | same contract the whole codebase (incl. quote-time) relies on; low-prob API drift |
| r1 | 3 | q (Codex) | `VaultBase.broadcastRawTx` bypasses guard | ℹ️ out-of-scope | takes `{chain, rawTx}` hex only, no keysignPayload → structurally can't be a native swap |
| ext | A | blocker (neavra @bc792e50) | destination chain never re-checked → BTC→SOL, SOL halts pre-broadcast = stuck funds | ✅ fixed @`8b769ee2` | now evaluates `native.toCoin.chain` too; red-on-revert confirms enforcement |
| ext | B | nit (neavra) | `global_trading_paused` read source-only | ✅ fixed @`8b769ee2` | now `inboundAddresses.some(…)` network-wide; red-on-revert confirms |

## delta verification (adversarial)
- **Both ends now covered.** `routeChainIds = [sourceChainId, destinationChainId]` deduped; loop throws on `halted || globalTradingPaused || chain_trading_paused` per leg. Source is guaranteed present (address check above), so the loop always runs ≥1 iteration → a network-wide `global_trading_paused` is always caught.
- **Freshness — still fresh, unchanged.** `getInboundAddresses(native.chain)` → `getThorchainInboundAddress()` (THOR) / raw `queryUrl(/inbound_addresses)` (Maya). No cache; the 5s `inboundCache` lives only in quote-time. Broadcast reads live state.
- **Fail-closed on fetch error — unchanged.** The `await getInboundAddresses(...)` is not wrapped in try/catch; an RPC error/timeout propagates and `BroadcastService` re-throws as `BroadcastFailed`. Empty feed → source `!activeInbound?.address` → `NetworkError`. Both fail-closed.
- **Tolerance is correct, not a fail-open.** A dest leg that resolves to no inbound entry (RUNE/CACAO, or the settlement chain itself — the feed never lists its own chain as inbound) is `continue`-skipped, not false-blocked. Proven by the "dest absent" allow-test staying green under neutralization. THOR/MAYA are keys in `nativeSwapChainIds`, but the live feed has no self-entry, so `get('THOR')`→undefined→skip.
- **Maya parity.** Guard applies the halt check to THOR **and** Maya (quote-time only checks THOR) — strictly stronger; Maya feed `.chain` values (`DASH`/`KUJI`/…) match the `nativeSwapChainIds` map neavra verified against the live feed.
- **No regression.** Both ends healthy → no throw → proceeds to the (byte-identical) address-match check. Golden-vector/encoder paths untouched.

## q
- `native.toCoin?.chain` is optional-chained: if `toCoin` were ever absent on a valid native payload, destination protection silently degrades to source-only (the r1 behavior). Not a regression and not fail-open beyond r1, but worth a one-line assert if `toCoin` is contractually always present on a native quote.

## risk
Low. Delta is additive fund-safety on an already-approved guard; the only behavioral change is stricter (more halts caught). Merge-commits don't touch the guard. Red-on-revert proves non-inert enforcement.

## verdict
**APPROVE** at `d6163d5d`. Both of neavra's blockers genuinely resolved; freshness + fail-closed preserved from r1; no gap to fix in-code.

<sub>🤖 reviewer lane (Claude primary; Codex lane timed out — best-effort, skipped per policy)</sub>

PR #1380, opened 2026-07-17
fix(rn): type balance dispatch as record<chainkind> + declare walletcore-native peer dep

### LINKED ISSUE #1372: Type RN getCoinBalance dispatch as Record<ChainKind, ...>, not Record<string, ...>
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P2 CONFIRMED.

**Anchor:** platforms/react-native/getCoinBalance.ts:6,41-43 vs core/chain/coin/balance/index.ts:17

## what
Core keys the balance-resolver dispatch as Record<ChainKind, ...> so the compiler forces every kind to be handled. The shipped RN version (platforms/react-native/getCoinBalance.ts) re-declares the same dispatch as Record<string, ...>, which throws at runtime for an unknown kind instead of failing to compile.

## failure scenario
A new ChainKind added to core fails compilation everywhere except RN. On RN specifically - the platform users are actually on - balances for the new chain silently throw at runtime instead of catching the gap at build time.

## fix
Type the RN resolverLoaders map as Record<ChainKind, () => Promise<BalanceResolver>> so a new kind breaks the RN build too. Small, mechanical, no behavior change for existing kinds.

Related: #1230 is the broader design spike for "consumers re-declare non-exhaustive tables instead of deriving from the SDK's exhaustive ones" - this issue is the narrow, already-actionable instance of that class in getCoinBalance.ts specifically, doesn't need to wait on the spike.

### LINKED ISSUE #1373: Declare @vultisig/walletcore-native as a peerDependency (RN entry hard-imports it, undeclared)
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P2 CONFIRMED.

**Anchor:** react-native/index.ts:46,58,70,73 - rollup.platforms.config.js:340 - sdk/package.json

## what
@vultisig/walletcore-native is a value import in the RN entrypoint (import { NativeWalletCore } from '@vultisig/walletcore-native', plus WalletCoreLike type re-export), marked external in the RN rollup bundle, but absent from the sdk package manifest's dependencies, peerDependencies, and devDependencies entirely. Verified: grep for "walletcore" in sdk/package.json returns nothing.

## failure scenario
An RN consumer installs @vultisig/sdk per its declared peer deps (which currently list @vultisig/mpc-native, async-storage, expo-crypto, but not walletcore-native) and hits module-not-found the moment the RN entry loads, since the RN bundle expects the host app to provide it but never says so.

## fix
Add @vultisig/walletcore-native to sdk/package.json peerDependencies (matching the existing @vultisig/mpc-native pattern), with peerDependenciesMeta if it should stay optional for non-RN consumers.

Related: #949 already flags the same "workspace package consumed via unstated dependency, breaks standalone install" class for core-chain/core-mpc - this is the same bug shape for a third package.

gomesalexandre:
Closes #1372
Closes #1373

## what

Two RN-entrypoint packaging/type-hygiene gaps from the round-7 audit, batched (both are RN-surface, zero runtime behavior change for existing setups):

- **#1372**: `platforms/react-native/getCoinBalance.ts` typed its `resolverLoaders` dispatch as `Record<string, ...>`, while core keys the identical dispatch as `Record<ChainKind, ...>`. So a new `ChainKind` added to core failed compilation everywhere EXCEPT RN, where balances for the new chain would silently throw at runtime (RN is the platform users are actually on) instead of being caught at build time. Retyped as `Record<ChainKind, ...>` so a missing kind breaks the RN build too.
- **#1373**: `@vultisig/walletcore-native` is a value import in the RN entrypoint (`import { NativeWalletCore }`) and marked external in the RN rollup bundle, but was absent from `sdk/package.json` entirely (deps / peerDeps / devDeps). An RN consumer installing `@vultisig/sdk` per its declared peer deps hit module-not-found the moment the RN entry loaded, since the bundle expects the host app to provide it but never said so. Added to `peerDependencies` + `peerDependenciesMeta` (optional), matching the existing `@vultisig/mpc-native` pattern.

## testing

Root `tsc` clean (the `Record<ChainKind>` retype compiles - the 12 loader keys already cover every kind). eslint clean. `package.json` validated as well-formed JSON; the new peer dep uses `workspace:*` like its `@vultisig/mpc-native` sibling and is verified to be a real workspace package (`packages/walletcore-native`, name `@vultisig/walletcore-native`).

## risk

Low. #1372 is a type-only tightening (no runtime change for existing kinds; it converts a future runtime throw into a build error). #1373 is a manifest-only declaration of a dependency the RN bundle already expects at runtime.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved React Native coin-balance resolution to use chain-type keys, providing safer compile-time behavior when new blockchain types are introduced.
  * Updated SDK package configuration so the native wallet core component is treated as an optional peer dependency, improving React Native build compatibility.


neavra (changes requested):
both fixes are right and i verified each rather than eyeballing them — the only thing blocking is a changeset, which the diff needs on the same grounds we've been holding all week.

what i checked:

- **#1372 is a genuine compile-time guarantee, not a cast.** no `as` on the object literal, and the 12 loader keys match the 12 `ChainKind` values exactly (`bittensor cardano cosmos evm polkadot qbtc ripple solana sui ton tron utxo`). red-proofed it: deleted the `qbtc` loader → `yarn workspace @vultisig/sdk typecheck` fails with

  ```
  error TS2741: Property 'qbtc' is missing in type '{...}' but required in type
  'Record<ChainKind, () => Promise<BalanceResolver>>'
  ```

  so a new kind added to core now breaks the RN build too, which is the whole point of the issue. and the parity claim holds — core keys its own dispatch as `Record<ChainKind, ...>` at `coin/balance/index.ts:17`. type-only change, erased at build, `resolverLoaders` isn't exported so not even the public `.d.ts` moves. also scanned the RN entry for the same `Record<string>` pattern elsewhere — the other hits are RPC params / headers / proxy casts, not chain dispatches, so this was the only instance.

- **#1373 is correct and can't break existing installs.** `@vultisig/walletcore-native` is a real, non-private workspace package (v0.1.2) and is actually published to npm (`npm view` → 0.1.2, same as its mpc-native sibling at 0.1.4). it's a *value* import in the RN entry (`index.ts:46`, used at :58) and marked external in the RN rollup (`rollup.platforms.config.js:340`) — so before this PR the bundle externalized it while the manifest never declared it → module-not-found on RN entry load. declaring it as an *optional* peer (matching the mpc-native line right above) fixes that, and because it's optional it won't force-install or error for consumers who don't use the RN entry. `workspace:*` publishes the exact version, consistent with the sibling. strictly an improvement, no new breakage.

**blocking: no changeset.** `packages/sdk/package.json` `peerDependencies` ships to npm, and `getCoinBalance.ts` is under `packages/sdk/src/` — CLAUDE.md:91/95/114 is unambiguous ("No exceptions", "even one-line fixes"). CI is green but :115 spells out that CI doesn't gate on this yet, so the tick doesn't cover it. a `@vultisig/sdk: patch` changeset does it.

two non-blocking:

- the `if (!loadResolver) throw` guard (`getCoinBalance.ts:43-45`) is now dead under the new type — `resolverLoaders[kind]` is typed non-optional and `kind` is a `ChainKind`. harmless defense-in-depth against a malformed object at runtime, so keep it if you want; just no longer a reachable type-level case.
- `workspace:*` pins the peer to an exact version. fine and consistent with the sibling for a host-provided native module — flagging only so it's a deliberate choice rather than an accident; `workspace:^` is the lever if you ever want RN consumers to float across patch releases of the native module.

lgtm once the changeset lands — both fixes are the right call.

NeOMakinG (approved):
fix-and-approve. Two-lane review (Claude + Codex gpt-5.4); both lanes converge and the one outstanding blocker from the prior round (missing changeset) is now resolved in-branch.

## testing approach
Reviewed the true PR diff (merge-base…HEAD = exactly 3 files: `packages/sdk/package.json`, `packages/sdk/src/platforms/react-native/getCoinBalance.ts`, `yarn.lock`; the two golden-test files that show up in a two-dot diff are main-drift, not part of this PR). Read the actual `ChainKind` definition (`packages/core/chain/ChainKind.ts`), the full RN dispatch map, the RN entrypoint value-import, and the `mpc-native` parity line. Codex adversarial lane ran the caret/optional/exhaustiveness/d.ts-leak angles independently.

## focus-area verdicts
- **(a) walletcore-native peer-dep range — NOT a floating-caret hazard.** The declared range is `workspace:*`, not `^X.Y.Z`. On Yarn 4, `workspace:*` publishes as the **exact** sibling version at pack time (not a floating npm range), mirroring the existing `@vultisig/mpc-native: workspace:*` line directly above it. The native-module caret runtime-crash hazard called out in the loop prompt does not apply here. `@vultisig/walletcore-native` is a real, non-private, published workspace package (v0.1.2). Optional is the correct choice — it avoids forcing the native module on non-RN consumers while declaring what the RN bundle already externalizes and value-imports (`index.ts:46`, `configureWasm` at `:58`).
- **(b) `Record<ChainKind>` exhaustiveness — strictly SAFER, no silent `undefined`.** `ChainKind` is the union of exactly 12 kinds; the map has exactly those 12 keys. Because `Record<ChainKind, …>` requires every key, a future unmapped kind now becomes a **compile error on RN** rather than a silent `undefined` dispatch. This is the inverse of a fund-display footgun — it closes one. (Prior reviewer red-proofed by deleting the `qbtc` loader → TS2741.)
- **(c) type-only change altering runtime — none.** The retype is erased at build, `resolverLoaders` isn't exported, so the public `.d.ts` doesn't move. The `if (!loadResolver) throw` guard is now unreachable at the type level but is harmless defense-in-depth against a malformed runtime object — kept intentionally.

## prior issues — status
|---|---|---|---|---|---|
| 1 (neavra) | 1 | blocker | no changeset for a published-package change (CLAUDE.md:91) | ✅ fixed | added `.changeset/rn-balance-chainkind-walletcore-peerdep.md` (`@vultisig/sdk: patch`) in commit 25612233 |
| 1 (neavra) | 2 | non-block | dead `!loadResolver` guard under new type | ℹ️ documented | harmless runtime defense-in-depth, left as-is |
| 1 (neavra) | 3 | non-block | `workspace:*` pins peer to exact version | ℹ️ documented | deliberate + consistent with `mpc-native` sibling for a host-provided native module |

## risk
Low. Type-only tightening (converts a future runtime throw into a build error) + a manifest-only declaration of a dependency the RN bundle already expects at runtime. No signing/tx-path bytes change. Both lanes found nothing beyond the changeset.

## verdict
APPROVE. Both fixes are correct and fund-safe; the sole blocker (changeset) is resolved in-branch at 25612233.

— reviewed by NeOMakinG (Claude Opus 4.8 primary lane + Codex gpt-5.4 adversarial lane)

rcoderdev (approved):

neavra (approved):
lgtm — re-reviewed at `4991c861` against my `cb032ad3` review. The sole blocker is cleared:

- **No changeset (was blocking)** — FIXED. `.changeset/rn-balance-chainkind-walletcore-peerdep.md` (`@vultisig/sdk` patch), covering both the `Record<ChainKind>` dispatch retype and the optional `walletcore-native` peer.

Confirmed no logic change since my review: the PR's own contribution over the current merge-base is exactly the three things I already verified — the `Record<ChainKind>` retype on `resolverLoaders` (genuine compile-time exhaustiveness, no `as` cast, 12 keys == 12 `ChainKind`), the optional `@vultisig/walletcore-native` `workspace:*` peer (+ `peerDependenciesMeta.optional`, matching the `mpc-native` sibling), and now the changeset. `getCoinBalance.ts` is byte-identical to `cb032ad3`. CI green at head.

Both non-blocking notes acknowledged as deliberate (the `if (!loadResolver) throw` kept as runtime defense-in-depth; `workspace:*` exact-pin consistent with the sibling). Nothing else outstanding.

Issue #1367, opened 2026-07-17
Mobile golden suite overrides mask two live SDK-vs-device hash divergences


gomesalexandre:
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P1 CONFIRMED.

**Anchor:** packages/core/mpc/keysign/tests/mobileFixtures.golden.test.ts:70-73 (sdkExpectedHashOverrides)

## what
For cardano.json::Send ADA - max amount and terra.json::Send USTC on terra classic, the SDK produces a different pre-signing hash than the recovered mobile fixture. The test pins the SDK's own output via sdkExpectedHashOverrides instead of asserting against the real device-recovered hash, so it stays green on a real divergence rather than failing on one. Otherwise a strong suite (71 cases / 23 fixtures vs real device hashes) - this is an honestly-commented but still-live gap in exactly two spots.

## failure scenario
An RN-iOS or RN-Android keysign pair on a Cardano max-amount ADA send, or a Terra Classic USTC send, hashes different bytes on each side today - the override just stops the test suite from screaming about it.

## fix
Root-cause the two divergences (Cardano max-amount SignDoc/CBOR construction, Terra USTC burn-tax-adjusted amount encoding) and remove the overrides once the SDK output matches the real device hash.

Issue #1347, opened 2026-07-17
epic: sdk canonical signing consolidation + chain-identity


gomesalexandre:
Tracking epic for the SDK side of the signing/chain-stack hardening campaign from the 2026-07-17 architecture study.

Companion app epic: vultisig/vultiagent-app#1984

## Through-line
`@vultisig/sdk` is meant to be the single source of truth for signing, amount formatting, chain identity, and broadcast. It mostly is (WalletCore-canonical per family), but several pieces of that canon are either package-internal (amount helpers, forcing 9 duplicate reimplementations across app+abts) or drifted between near-identical resolvers (Bittensor vs Polkadot). This epic: re-export what should already be public, collapse the drift, and add a branded chain-identity module computed from the existing `Chain` enum (not a CAIP-2 re-keying) so consumers stop hand-rolling chain-alias tables.

## Sub-issues
- [ ] vultisig/vultisig-sdk#1348 PR-5(sdk) re-export canonical amount helpers from the SDK root
- [ ] vultisig/vultisig-sdk#1349 PR-7 collapse Bittensor/Polkadot broadcast+status resolver drift
- [ ] vultisig/vultisig-sdk#1350 PR-8(sdk) branded ChainId/AssetId module computed from Chain
- [ ] vultisig/vultisig-sdk#1351 gas-floor-unify(sdk) single-source per-chain gas-floor table
- [ ] vultisig/vultisig-sdk#1352 broadcast-strategy(sdk) promote raced-public-RPC broadcast to an opt-in SDK strategy

## Downstream consumers
- vultisig/vultiagent-app#1989 (PR-5 app), #1990 (PR-6), #1991 (PR-8 app) depend on sub-issues above.
- vultisig/agent-backend-ts#1456 (PR-5 abts), #1457 (PR-8 abts) depend on sub-issues above.

Each PR closes its sub-issue. Consumer PRs in the app/abts repos that depend on one of these are gated on it being republished to npm and version-bumped first.

## r7 audit sub-issues (2026-07-17)

- [ ] #1365 - Encoder-parity test missing: RN pure-JS vs WalletCore never cross-checked (plan 003 unexecuted)
- [ ] #1366 - QBTC fee/gas conflation on the hand-rolled proto path, zero golden coverage
- [ ] #1367 - Mobile golden suite overrides mask two live SDK-vs-device hash divergences
[...]

PR #1381, opened 2026-07-17
test: exotic-chain golden hardening + root-cause the 2 masked mobile divergences

### LINKED ISSUE #1368: Exotic-chain golden-vector gaps: ZEC/BCH per-coin vectors, THOR/Tron/Ripple second-impl cross-checks
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P1 CONFIRMED.

**Anchor:** packages/core/mpc/tx/compile/compileTx.golden.test.ts (covers only Ethereum/Solana/Cosmos/Cardano/Bittensor/QBTC) - packages/core/mpc/keysign/tests/fixtures/mobile/utxo.json (single fixture for all UTXO coins)

## what
Three golden layers exist: WalletCore second-implementation (compileTx.golden.test.ts, strongest), mobile device-hash fixtures, and RN vectors. Gaps concentrate exactly where encoding is most exotic: a single generic utxo.json stands in for all six UTXO coins, so ZEC (NU5 v5) and BCH (cashaddr+FORKID), which have genuinely special encoding, have no dedicated golden signing vector. THORChain/Tron/Ripple/Ton have Layer-2 (independent re-encode) coverage via the gap-chain golden suite, but no WalletCore second-implementation cross-check in compileTx.golden.test.ts. Ripple is weakest of all (no device fixture and no second-impl).

## failure scenario
A protocol-conformance regression in the ZEC NU5 or BCH cashaddr/FORKID encode path, or in the THORChain/Tron/Ripple builders, has no independent reference to catch a wrong-wire-format bug before it reaches a real node.

## fix
Add per-coin goldens for ZEC/BCH/LTC/DOGE/DASH (Zcash's own signing machinery is already verified correct elsewhere in this audit, so its golden is a cheap add) and WalletCore second-impl cross-checks for the swap-heavy THOR/Tron/Ripple in compileTx.golden.test.ts.

gomesalexandre:
Closes #1368
Tackles #1367

## what

Golden-vector hardening across the two round-7 gaps (epic #1347), batched since they share the same testing surface:

**#1368 - exotic-encoder golden gaps (additive coverage):** `compileTx.golden.test.ts` goes 7 -> 16 cases.
- Per-coin UTXO goldens: BCH (cashaddr + FORKID sighash), LTC, DOGE, DASH, ZEC - a single generic `utxo.json` previously stood in for all six UTXO coins. Each new vector cross-checks the `compileTx` output byte-for-byte against `walletCore.AnySigner.sign` (independent second implementation), plus a pinned golden constant. A dedicated case proves DOGE and DASH serialize byte-identically for equivalent inputs (shared legacy P2PKH sighash).
- WalletCore second-implementation cross-checks for the swap-heavy THORChain `types.MsgDeposit` (native RUNE), Tron `TransferContract`, and Ripple `Payment` - these had Layer-2 (independent re-encode) coverage but no WalletCore second-impl cross-check.
- ZEC caveat documented in-code: WalletCore 4.7.0 emits v4-Sapling framing with an NU5+ branch id, not true NU5 v5/ZIP-244 - this suite pins that current output as a wire-format regression guard, NOT a v5-consensus correctness claim (the SDK's own v5-aware builder is out of scope here).

**#1367 - root-cause the 2 masked mobile divergences (investigation, no encoder touched):** both `sdkExpectedHashOverrides` divergences were root-caused; per the fund-safety rule, neither encoder was changed without a verifiably-correct fix, so both overrides stay - each now carries a detailed evidence comment. See findings below. #1367 stays open for the two dedicated follow-ups.

## findings (#1367)

- **Cardano `Send ADA - max amount` - a REAL BUG surfaced.** wallet-core 4.7.0 returns `TransactionPlan{ amount: <full input, unadjusted>, fee: 0 }` for `Transfer.useMaxAmount:true + forceFee` (verified via both `preImageHashes` and `AnySigner.plan`). A fee=0 Cardano tx is invalid on-chain (nodes require `fee >= minFeeA + minFeeB*size`), so **Cardano send-max is currently broken - it would be rejected at broadcast.** The one fee/amount pair that reproduces the device hash (`fee=164515, amount=9835485`) does not factor cleanly from Cardano's standard linear-fee formula, so no general correct algorithm could be derived here. Recommended follow-up: client-side fee-convergence (mirror how the Bitcoin UTXO resolver calls `AnySigner.plan` before hashing). Filed as a note for a dedicated fix - flagging separately as a live reliability bug.
- **Terra USTC `Send USTC on terra classic` - mechanism identified, not reproducible offline.** The divergence is `getCosmosSigningInputs`' Terra Classic stability-tax surcharge for `uusd` (an extra dynamically-LCD-fetched `Fee.amounts` coin). The offline fixture has no `ibc_denom_traces` (predates that field), so replay computes `burnTaxAmount=0` - consistent with every `uluna` (tax-exempt) fixture matching and only the `uusd` case diverging. Exhaustive brute-force over amount-reduction + surcharge combinations found no exact match. The live tax rate is governance-set to 0 today, so hardcoding a historical value would REGRESS present-day sends. Recommended follow-up: a fresh device-recorded fixture under the current rate=0 regime, or recovering the exact historical tax_rate/tax_cap from chain history.

## testing

`compileTx.golden.test.ts` + `mobileFixtures.golden.test.ts`: 88/88 green (compileTx 16/16, all new UTXO/second-impl cases cross-checked against AnySigner). Full affected core golden suites: 105/105. RN golden suites unaffected: 156/156. Root `tsc` 0 errors. prettier clean; the two test files fall under the repo's pre-existing `packages/core/mpc/**/*.test.ts` eslint ignore glob.

## risk

None to runtime - tests + evidence comments only, zero `src/` changes. The #1367 overrides are unchanged (the divergences are documented, not newly introduced), and #1367 stays open for the two follow-ups.

### DIFF
--- a/packages/core/mpc/keysign/tests/mobileFixtures.golden.test.ts
+++ b/packages/core/mpc/keysign/tests/mobileFixtures.golden.test.ts
@@ -67,6 +67,74 @@ const compareHashesAsSet = ({ chain, fixtureFile }: { chain: Chain; fixtureFile:
 // Keep the restored Android/iOS JSON expected hashes unchanged. These cases
 // currently differ through the SDK path, so the test pins SDK output here and
 // makes each difference explicit instead of silently rewriting mobile fixtures.
+//
+// Root-cause notes (golden-vector hardening pass, sdk#1367):
+//
+// - cardano.json::Send ADA - max amount
+//   Root cause CONFIRMED: this repo's pinned wallet-core (4.7.0) has a real
+//   bug/regression for Cardano `Transfer.useMaxAmount: true` combined with a
+//   client-supplied `forceFee`. Verified via TWO independent WalletCore entry
+//   points — `TransactionCompiler.preImageHashes` (what this SDK calls) AND
+//   `AnySigner.plan` (the "official" planning API) — both return
+//   `TransactionPlan{ amount: <full input UTXO, unadjusted>, fee: 0 }` for
+//   this fixture's `useMaxAmount: true` input, ignoring `forceFee` entirely.
+//   A fee=0 Cardano tx is invalid on-chain (real nodes require
+//   fee >= minFeeA + minFeeB*txSize) — a "send max" Cardano transaction built
+//   by this SDK today would be rejected at broadcast, not silently mis-signed.
+//   Separately CONFIRMED via exhaustive brute force (fee in [150_000,
+//   200_000] lovelace) that fee=164_515 / amount=9_835_485 reproduces the
+//   recorded device hash byte-for-byte — proving the device's original
+//   signer DID compute a real minimum fee for this send, and that a correct
+//   fix is possible in principle. However, 164_515 does NOT factor as an
+//   integer solution of the standard Cardano linear-fee formula
+//   (minFeeA=155_381 + minFeeB=44 * txSizeBytes) against this transaction's
+//   actual encoded size (90-byte body + a single ed25519 witness) — so the
+//   exact algorithm the device used (a different WalletCore version? a
+//   different placeholder-size assumption for the fee-convergence loop?)
+//   was NOT recoverable from this fixture alone. Applying a fix without a
+//   verified-general formula risks silently mis-computing fees on other
+//   real max-amount Cardano sends — worse than the current (loudly broken,
+//   fee=0) behavior. Follow-up: implement client-side fee convergence for
+//   Cardano max-amount sends (mirroring how the UTXO/Bitcoin resolver already
+//   calls `AnySigner.plan` before hashing) using the CURRENT wallet-core's
+//   own Cardano linear-fee behavior, validated against multiple max-amount
+//   fixtures (not just this one) before removing this override.
+//
+// - terra.json::Send USTC on terra classic
+//   Root cause LIKELY (not conclusively reproduced): `getCosmosSigningInputs`
+//   has a dedicated Terra Classic stability-tax surcharge for USTC (`uusd`)
+//   sends (see `getFee`'s `areEqualCoins(coin, { chain: TerraClassic, id:
+//   'uusd' })` branch) — a burn-tax amount is meant to be pre-computed by the
+//   ASYNC `getCosmosChainSpecific` resolver (an `x/treasury` LCD query, see
+//   `terraClassicTax.ts`) and threaded through as an extra `Fee.amounts`
+//   coin via `CosmosSpecific.ibcDenomTraces.baseDenom`. This offline fixture
+//   JSON carries no `ibc_denom_traces`/`base_denom` field (that mechanism
+//   post-dates when this Android/iOS hash was captured), so replaying it
+//   through the current resolver always computes `burnTaxAmount = 0` and
+//   emits a single-coin fee — plausibly NOT what the device signed if the
+//   real on-chain tax rate was nonzero at capture time (the live rate is
+//   governance-set and is `0` today, per `terraClassicTax.ts`, but was
+//   historically ~1.2%). This is consistent with LUNC/uluna sends in this
+//   same fixture file matching exactly (uluna is tax-exempt — see
+//   `applyTerraClassicTax`) while only the uusd case diverges. However,
+//   exhaustive brute-force search did NOT find an exact reproduction: tried
+//   (a) reducing the send amount by 0-5% and by a wide absolute window
+//   ([199_000_000, 200_000_000] and curated tax-rate/cap candidates), (b)
+//   adding a second `uusd` fee coin (both appended and prepended) over
+//   [0, 5_000_000] uusd, (c) the two combined (amount reduced AND fee coin
+//   added) over [0, 3_000_000], and (d) replacing the fee entirely with a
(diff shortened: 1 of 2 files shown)


neavra (approved):
lgtm. this is the right call on both divergences — not touching a signer without a verifiably-correct fix is exactly the posture for fund-safety code, and the evidence comments are the kind that save the next person a day. i verified both root-causes rather than taking them on faith, and on the cardano one i think i can hand you a sharper fix direction.

**cardano max-amount — the "real bug" reproduces, and it's worse-specified than the comment says.** built the fixture's exact input through the real `getCardanoSigningInputs` + `AnySigner.plan` + `getPreSigningHashes` on the pinned wallet-core 4.7.0:

| input | plan | pre-sign hash |
|---|---|---|
| `useMaxAmount=true forceFee=44` (fixture as-recorded) | `amount=10000000 fee=0` | `d681b85a…` (your override) |
| `useMaxAmount=true forceFee=164515` | `amount=10000000 fee=0` | `d681b85a…` — **identical** |
| `useMaxAmount=false forceFee=164515 amount=9835485` | `amount=9835485 fee=164515` | **`c2a0b05a…` = the device hash, exactly** |

the middle row is the crux: bumping forceFee under `useMaxAmount=true` changes nothing. so it's not "wallet-core ignores forceFee" in general — it's **`useMaxAmount:true` nullifies `forceFee` specifically**, and the plan comes back fee=0. and row 3 reproduces the recorded device hash byte-for-byte, which proves the device signed with `useMaxAmount=false` and an explicitly-computed amount, not max-mode at all.

which leads to a correction on the recommended follow-up. you suggest "client-side fee-convergence, mirroring how the Bitcoin UTXO resolver calls `AnySigner.plan`" — but cardano *already has* one: `estimateCardanoByteFee` (`chainSpecific/resolvers/cardano.ts:45`) runs the `fee = 44*size + 155_381` loop. i ran it: it converges to `byteFee=163873` — and then `useMaxAmount:true` throws it away and emits fee=0. so the fix isn't "add convergence"; it's **stop passing `useMaxAmount=true` for cardano and send `amount = balance − convergedFee` with `useMaxAmount=false`**. i verified that path produces a valid non-zero-fee plan (my corrected loop lands fee=164049, clean, no error) — it just doesn't hit the device's exact 164515, so your "exact historical algorithm not recoverable from this fixture" holds. worth a one-line tweak to the evidence comment so the fix PR chases the right lever (the `useMaxAmount` flag, not a missing convergence loop).

and to be clear this reaches users, not just fixtures: `getCardanoSigningInputs:51` sets `useMaxAmount: sendMaxAmount`, and `getCardanoChainSpecific` sets `sendMaxAmount = balance === amount` — that's literally "send max". so keeping the override + flagging it as a live reliability bug is the correct outcome.

**terra USTC — attribution is honest and the hedge is the right one.** verified the load-bearing facts: the USTC fixture carries no `ibc_denom_traces` (parsed it), so the resolver computes `burnTaxAmount=0`; `terraClassicTax.ts` confirms uluna is tax-exempt (which is exactly why every uluna case matches and only uusd diverges), live rate 0 today, historically ~1.2%, and hardcoding a historical value would regress present-day sends. "LIKELY, not conclusively reproduced" after that brute force is the honest label. correctly kept as an override.

**does the hardening hide divergence? no.** the two divergences are pinned as loud overrides with evidence, not rewritten into the mobile fixtures, and you add zero new overrides — the two `sdkExpectedHashOverrides` values are byte-unchanged from main (only comments added). the empty-hash cases still assert. good.

receipts: `compileTx.golden` + `mobileFixtures.golden` 88/88, full core suite 1840 passed / 205 files, root `tsc` clean. red-proof (inverted, since this is test-only): mutated the production `utxo.ts` hashType to drop FORKID → reds 2 in mobileFixtures (BCH swap + BCH send-max, the resolver path) while compileTx.golden stays 16/16 — see the note below on why.

two non-blocking:

- the new per-coin UTXO goldens call `AnySigner.sign` an "independent second implementation," but `compileUtxoGolden` builds its own `TW.Bitcoin.Proto.SigningInput` and cross-checks `compileWithSignatures` against `AnySigner.sign` — two wallet-core entry points sharing the same wire encoder. it's a real check of the compile/assembly path and a solid version-drift guard, but a wallet-core-wide encoding bug would survive both (and the pinned constant is wallet-core-generated too). the genuinely-independent bind is the RN↔wallet-core one in #1376; these are wallet-core-vs-wallet-core. fine and useful, just not fully independent — worth softening the wording. (this is also why the FORKID mutation above doesn't red compileTx.golden: it re-derives `hashTypeForCoin` itself; the production `utxo.ts` FORKID path is covered by the pre-existing `utxo.json` mobile fixtures instead, which is what my red-proof reds.)
- the ZEC caveat is scoped honestly — pinning wallet-core 4.7.0's v4-Sapling + NU5-branch-id framing as a regression guard, not a v5/ZIP-244 correctness claim. no notes, just confirming i read it and agree it's not overclaiming.

nice work — surfacing a live cardano send-max bug as a byproduct of a golden-hardening pass, and resisting the urge to "fix" it with an unverified formula, is exactly right.

PR #1385, opened 2026-07-17
fix(cli): make machine output tell the truth about what happened

neavra:
Five CLI fixes that share one theme: a machine consumer of the CLI couldn't trust what it got
back. Each is from an audit of low-severity CLI issues, and each has a red-then-green test. No
fund-movement behaviour, no wire contract, and no change to exit codes 9/12/13.

Split out of #1354, which batched nine of these fixes and was too big to review in one sitting.
No code was dropped in the split.

### What's in here

| # | Fix | Before |
|---|---|---|
| 1 | Every mutation emits a result envelope | switch/rename/currency/address-book/import/export/add-mldsa exited 0 with **empty** stdout in `-o json`, while delete/chains-add already emitted envelopes |
| 2 | One envelope from `verify`; pending vaults in `vaults` | `verify -o json` wrote **two** JSON documents, so `JSON.parse` failed; pending two-step vaults were invisible |
| 3 | Validate the chain arg for `tokens` / `swap-quote` | `tokens bogus-chain` → exit 0 + `{"success":true,...,"tokens":[]}`; `swap-quote bogus-chain …` → exit 7 + a raw `Cannot read properties of undefined (reading 'ticker')` |
| 4 | `NO_ACTIVE_VAULT` (15) / `CORRUPT_STATE` (16) | Both were a generic `UNKNOWN_ERROR`/7 with no hint |
| 5 | `send --dry-run` reports fee/total | JSON dropped both, even though the human preview printed the fee and `total` is what the insufficient-balance warning is computed from |

### Notes for review

- **#2** lists pending vaults best-effort — a pending-store read failure must not take down
  `vaults`, which is the command you use to diagnose vault state in the first place. The SDK's
  own wording for a missing pending vault mentions `createFastVault()`, which is legitimate for
  library callers but meaningless in a shell, so the CLI owns its copy rather than editing the SDK.
- **#4** adds two exit codes. This follows the existing pattern (8–14 were each carved out of the
  generic bucket for the same "let a headless caller branch on `$?`" reason) and the README table
  is pinned to the enum by the existing doc-lint test. Corruption is detected *positively* — a
  `StorageError` whose cause is a `JSON.parse` `SyntaxError` — so transient IO/permission failures
  keep their current classification rather than being mislabelled unrecoverable. The fail-open
  active-pointer reset from #1077 is untouched.
- **#3** deliberately leaves the same unchecked `findChainByName(x) || (x as Chain)` cast on the
  `send`/`swap` fund-movement paths — out of scope for a polish sweep.

### Heads-up: one deliberate behaviour change worth a maintainer opinion

**#2 shifts `verify`'s failure exit code from 1 to 4/5.** It previously threw a bare `Error` with
`exitCode = 1` attached (which `classifyError` ignored anyway); it now throws `InvalidInputError`
(4) for a wrong/expired code and `VaultNotFoundError` (5) for an unknown pending vault. That's
more correct and consistent with the documented table, but it IS a real change on a published
surface, and this PR's claim only fences 9/12/13 — so flagging it rather than burying it. It's
called out in the changeset too. Happy to pin `verify` failure back to 1 if anything depends on it.

### Sibling PRs

#1354 was split into three PRs off `main`. The other two are #1386 (keyshare file safety) and
#1387 (CLI DX polish). This PR and #1386 both touch `executeExport` in
`clients/cli/src/commands/vault-management.ts` — this one adds its JSON envelope, that one
changes how the file is written — so whichever lands first leaves the other a trivial rebase.
No other file is shared with either sibling.

### Validation

`yarn test:cli`, `yarn typecheck` (root + CLI workspace), and changed-file eslint + prettier all
pass locally. Each fix was red-checked by reverting it and observing the specific failure. No live
transactions were made.

### Post-review fix

A local multi-perspective review ran over this branch before it left draft and found one
correctness issue, now fixed in commit `43c2dc61`:

- **`verify` misclassified a post-acceptance failure as a bad code.** `executeVerify` runs
  `verifyVault()` and `setActiveVault()` in one `try`, so the catch also sees failures *after*
  the server accepted the code — e.g. a `StorageError` persisting the active-vault pointer. Those
  were routed to `InvalidInputError` (exit 4, "the code may be incorrect or expired"), which for a
  corrupt store is doubly wrong: the code was right, and it sends the user to re-send an OTP that
  was never the problem — masking the `CORRUPT_STATE`/16 signal this PR adds. Now classifies first
  (mirroring the `--resend` path in the same file) so a corrupt store surfaces as `CORRUPT_STATE`.
  Regression test added (red on revert), plus a test that drives the real `withExit` wrapper so
  the "exactly one JSON document" claim is proven end-to-end rather than only at the `executeVerify`
  layer where the old two-document bug didn't live.

A second re-review round noted the first fix was narrower than its commit title implied — the
`try` still spanned the post-acceptance steps, so a *non*-corruption failure there (a full disk
writing the pointer) could still surface as "code may be incorrect." Fixed in `cc11dae7` by
splitting `executeVerify` into two phases: only the verify call maps to a bad-code error; once the
code is accepted, activation failures classify directly (corrupt store → CORRUPT_STATE/16,
everything else → generic UNKNOWN/7, never a bad-code claim). Test added, red on revert.

## Summary by CodeRabbit

* **New Features**
  * JSON-mode mutations now consistently return one structured result.
  * `send --dry-run` JSON output includes estimated fees and totals.
  * Vault listings now show pending verifications.
  * Added actionable exit codes for missing active vaults and unreadable state.

* **Bug Fixes**
  * Invalid chain names now produce clear, parseable errors.
  * Verification failures no longer produce duplicate JSON output.
  * Verification and resend errors now provide more accurate status and recovery guidance.
  * Vault import, export, switching, renaming, and key-management commands now report structured JSON results.

### DIFF
--- a/.changeset/cli-truthful-output-batch.md
+++ b/.changeset/cli-truthful-output-batch.md
@@ -1,0 +1,23 @@
+---
+'@vultisig/cli': patch
+---
+
+Make CLI machine output tell the truth about what happened:
+
+- Emit a result envelope from every successful mutation in `-o json`
+  (switch/rename/currency/address-book/import/export/add-mldsa previously exited 0
+  with no output at all).
+- Emit exactly one JSON envelope from a failed `verify` — it previously wrote two
+  documents, so `JSON.parse` on the output failed — and list vaults pending
+  verification in `vaults`, which were otherwise only named in the create-time output.
+- Validate the chain argument for `tokens` and `swap-quote`, which previously reported
+  `success: true` with an empty list / a raw TypeError for an unknown chain. Both now
+  throw a typed `INVALID_CHAIN` (exit 4).
+- Add `NO_ACTIVE_VAULT` (15) and `CORRUPT_STATE` (16) exit codes with recovery hints,
+  replacing a generic `UNKNOWN_ERROR`/7 for both states.
+- Report `fee` and `total` from `send --dry-run` in JSON — the human preview already
+  printed the fee, and `total` is the number the insufficient-balance warning is
+  derived from.
+
+Note: a failed `verify` now exits 4 (wrong/expired code) or 5 (unknown pending vault)
+instead of 1. This is a deliberate change on a published surface — see the PR body.
--- a/clients/cli/README.md
+++ b/clients/cli/README.md
@@ -1262,6 +1262,8 @@ Configuration is stored in `~/.vultisig/`:
 | 12   | Interactive confirmation/input required but the session is non-interactive — pass --yes/--confirm or the required flag |
 | 13   | agent ask: transaction broadcast but the overall request may be incomplete — inspect the hash, do NOT blindly retry    |
 | 14   | agent ask: duplicate keyed turn rejected — inspect the conversation for the original result                            |
+| 15   | No active vault selected — create, import, or switch to one                                                            |
+| 16   | Stored state is unreadable — repair it or re-import the vault from a .vult backup                                      |
 
 > These are generated from the `ExitCode` enum in `src/core/errors.ts` (the single source of
 > truth) and are covered by a doc-lint test that fails if this table drifts from the code. Run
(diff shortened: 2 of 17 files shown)


rcoderdev (approved):

gomesalexandre (approved):
re-upped after update-branch. cli diff still looks good and the original intent remains intact on the refreshed head.

PR #1387, opened 2026-07-17
fix(cli): silence bigint stderr noise and generate completion from the registries

neavra:
Two developer-experience fixes from an audit of low-severity CLI issues. Each has a red-then-green
test. No fund-movement behaviour, no exit-code change, and no wire contract.

Split out of #1354, which batched nine fixes and was too big to review in one sitting. No code was
dropped in the split.

### What's in here

| # | Fix | Before |
|---|---|---|
| 1 | Silence `bigint: Failed to load bindings` | Printed to stderr on **every** invocation, incl. `--version` |
| 2 | Completion generated from the registries | Shipped a stale hardcoded list: no sign/broadcast/tx-status/execute/discount/agent/auth/delete/join/rujira/add-mldsa, and 21 of 38 chains |

### Notes for review

- **#1** is a `yarn patch` on `bigint-buffer`, which `console.warn`s at *module load* — before any
  flag parsing, so it can't be gated behind `--debug`. The advice it gives ("try npm run rebuild?")
  is meaningless for a global install, and the pure-JS fallback works fine, so this is pure noise
  that makes a working CLI look broken on first contact. `--version` stderr goes 77 bytes → 0.

  **Build-order caveat:** the SDK's node bundle inlines the patched copy, so `build:sdk:bundle`
  must run after install for the fix to appear. CI's order is `install --immutable` →
  `build:shared` → `build:all`, so the patch lands before the bundle is built. A stale local
  `dist` yields a false **red**, not a false green.

  The patch itself is one hunk against `dist/node.js`: it deletes a `console.warn` and adds
  comments. Zero crypto or parsing change. It does **not** mask GHSA-3gc7-fjrx-p6mg (the existing
  audit-ignore): that warn fires only when the native binding *fails* to load — i.e. exactly the
  safe pure-JS path. The vulnerable native path was always silent.

  One thing worth a maintainer opinion: the `patch:` resolution also **pins `bigint-buffer` at
  1.1.5**, so `yarn up bigint-buffer` — the documented exit from that audit-ignore — is now a
  no-op, and a published fix would be silently overridden. I've put a note next to the ignore in
  `.yarnrc.yml` so whoever revisits the advisory drops the pin too. `@solana/buffer-layout-utils`
  is the only dependent today and declares exactly `^1.1.5`.

- **#2** reads commands from Commander's singleton and chains from `SUPPORTED_CHAINS`. Ordering is
  safe: `handleCompletion()` awaits a dynamic `import()` before it ever reaches `getCommands()`, so
  the IIFE yields, `index.ts`'s module body registers every command synchronously, and only then
  does the microtask resume — `getCommands()` can't observe an empty `program.commands`.
  `SUPPORTED_CHAINS` is `Object.values(Chain)`, the same source `findChainByName` reads, so
  completion can't suggest a chain the resolver would then reject.

### Why `@vultisig/sdk` is in the changeset

The bigint patch only touches root `package.json`, but the SDK's published node bundle inlines the
patched copy — so without an SDK version bump, consumers of `@vultisig/sdk` never actually get the
fix. `scripts/check-sdk-bundled-changeset.mjs` doesn't catch this (it guards `packages/core/*` and
`packages/lib/utils/` prefixes), so the bump is deliberate rather than CI-driven. Shout if you'd
rather ship it CLI-only.

### Sibling PRs

#1354 was split into three PRs off `main`. The other two are #1385 (truthful CLI output) and #1386
(keyshare file safety). This PR shares no files with either, so it can land in any order without
a rebase.

### Validation

`yarn test:cli`, `yarn typecheck` (root + CLI workspace), and changed-file eslint + prettier all
pass locally. Both fixes were red-checked by reverting them and observing the specific failure
(`--version` stderr non-empty; zsh completion missing `sign`). No live transactions were made.

## Summary by CodeRabbit

- **New Features**
  - Improved shell completion for Bash, Zsh, and Fish.
  - Completion suggestions now include all available CLI commands and supported chains.

- **Bug Fixes**
  - Removed an unnecessary `bigint-buffer` warning from CLI output.
  - CLI version and help commands now produce cleaner output without native-binding or rebuild warnings.

- **Tests**
  - Added coverage to verify shell completion and clean CLI output across supported commands and chains.

### DIFF
--- a/.changeset/cli-dx-polish.md
+++ b/.changeset/cli-dx-polish.md
@@ -1,0 +1,16 @@
+---
+'@vultisig/cli': patch
+'@vultisig/sdk': patch
+---
+
+Two CLI developer-experience fixes:
+
+- Silence the `bigint: Failed to load bindings, pure JS will be used` warning that
+  `bigint-buffer` printed to stderr on every invocation, including `--version`. It
+  warns at module load, before any flag has been parsed, so it cannot be gated behind
+  `--debug`; it is silenced with a `yarn patch`. The SDK's node bundle inlines the
+  patched copy, which is why `@vultisig/sdk` is bumped here too.
+- Generate shell completion from the Commander command table and the SDK chain
+  registry instead of a stale hardcoded list, which was missing eleven commands
+  (sign/broadcast/tx-status/execute/discount/agent/auth/delete/join/rujira/add-mldsa)
+  and 17 of 38 chains.
--- a/.yarn/patches/bigint-buffer-npm-1.1.5-785f4ccd92.patch
+++ b/.yarn/patches/bigint-buffer-npm-1.1.5-785f4ccd92.patch
@@ -1,0 +1,18 @@
+diff --git a/dist/node.js b/dist/node.js
+index 513168c89e387668a79b6beb109587b3eb41a7d0..73d35813858e8fda34541200d0ecff0977a8683c 100644
+--- a/dist/node.js
++++ b/dist/node.js
+@@ -7,7 +7,12 @@ let converter;
+         converter = require('bindings')('bigint_buffer');
+     }
+     catch (e) {
+-        console.warn('bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)');
++        // Patched by vultisig-sdk: the native binding has no darwin-arm64 (and most
++        // other) prebuild, so this fires on EVERY process start and printed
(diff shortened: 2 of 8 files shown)


rcoderdev (approved):

gomesalexandre (approved):
re-upped after update-branch. cli diff still looks good and the original intent remains intact on the refreshed head.

PR #1399, opened 2026-07-18
fix(swap): reject 0X-prefixed burn recipient on native cross-chain routes

NeOMakinG:
## what

`findSwapQuote`'s `isEvmAddress` shape check (`/^0x[0-9a-fA-F]{40}$/`) is case-sensitive on the `0x` prefix. A recipient like `0X000…dEaD` (uppercase X) fails the check, so it **skips the zero/burn-address branch**. On a native THOR/Maya cross-chain route (BTC→ETH) the CowSwap format gate is unreachable, so `assertValidCustomRecipient` falls through without throwing and the burn is passed straight through as the swap destination — an **unrecoverable sink**.

This makes the prefix match case-insensitive (`/i`), matching the Go guard. The burn comparison already lowercases the recipient, so **no accepted address changes** — this only closes the bypass.

## why a separate PR

Surfaced while re-reviewing #1193 (reconcile + export dangerous/burn-address guard). #1193 routes this guard through the shared canonical table and its changeset states a `0X…` burn "can't slip past", but it preserves this case-sensitive *local* gate that sits in front of the (case-insensitive) shared check — so on the native route the slip persists. The gate predates #1193 (introduced by `5926f3be`, already on `main`), so this is an independent pre-existing fix rather than a change to #1193's diff. The `isEvmAddress` line is byte-identical on `main` and on #1193's HEAD, so this one-line change merges cleanly either way.

## test

Added a red-on-revert test on the BTC→ETH THORChain route: `0X000…dEaD` must be rejected with `InvalidConfig` before any provider is called. Reverting the `/i` flips it to PASS-THROUGH (proven standalone). Lowercase `0x…dead` still blocks and a valid `0x…` recipient still passes — no false-positive regression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed validation for burn and zero addresses using an uppercase `0X` prefix.
  * Invalid recipients are now rejected before swap processing, preventing transactions from being sent to unrecoverable destinations.

### DIFF
--- a/.changeset/fix-0x-prefix-burn-slip.md
+++ b/.changeset/fix-0x-prefix-burn-slip.md
@@ -1,0 +1,14 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Fix a `0X`-prefixed (uppercase-X) burn recipient slipping the custom-recipient
+guard on native THOR/Maya cross-chain swaps.
+
+`findSwapQuote`'s `isEvmAddress` shape check was case-sensitive on the `0x`
+prefix, so a recipient like `0X000…dEaD` failed the check, skipped the
+zero/burn-address branch, and — on a native route where the CowSwap format
+gate is unreachable — passed straight through as the swap destination, an
+unrecoverable sink. The gate is now case-insensitive on the prefix (`/i`),
+matching the Go guard; the burn comparison already lowercases, so this only
+closes the bypass without changing any accepted address.
--- a/packages/core/chain/swap/quote/findSwapQuote.recipient.test.ts
+++ b/packages/core/chain/swap/quote/findSwapQuote.recipient.test.ts
@@ -176,6 +176,30 @@ describe('findSwapQuote external recipient', () => {
     }
   )
 
+  it('rejects a 0X-prefixed (uppercase-X) burn recipient on a cross-chain native route', async () => {
+    // Regression: the case-sensitive isEvmAddress gate let `0X…dead` bypass the zero/burn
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
Verified the bypass end-to-end. On the current head, `assertValidCustomRecipient` in `packages/core/chain/swap/quote/findSwapQuote.ts` only reached the burn-address check from inside `isEvmAddress(recipient)`, and the trailing format throw was gated on `cowSwapPathReachable`. On a BTC->ETH native route that meant `0X000000000000000000000000000000000000dEaD` skipped the case-sensitive shape gate, skipped the burn block, skipped the CowSwap-only throw, and got forwarded as the native swap recipient. Because the native path then embeds the destination into the THOR memo as a raw string, nothing downstream canonicalized it back into a safe fail. Real fund-loss hole.

This `/i` fix closes exactly that hole, and the existing burn comparison already lowercases before matching, so mixed-case and `0X`-prefixed variants now get caught the same way as lowercase `0x`.

Cross-checks:
- the same line is still case-sensitive on `main`, so this isn't dead code
- #1193's canonical dangerous-address regex is already `/i`, but the local pre-gate here still mattered on this path
- `limitSwapMemo.ts` has the same case-sensitive helper, but there it fails closed by throwing on invalid format, so no parallel bypass

The new test is non-inert: it pins the vulnerable native route, proves the guard fires before quote fetch, and goes red on revert. CI is fully green.

Approve. Tight scope, correct fix, fund-safe.

PR #1400, opened 2026-07-18
fix(sdk): fail closed on malformed XRPL account_info/account_lines in RN balance path

NeOMakinG:
Stacked follow-up on #1220 (targets its branch `fix/rn-ripple-getcoinbalance`).

## Why
Review of #1220 found the new RN Ripple balance path fails **open** on a narrow but fund-relevant class of response: an HTTP-200, non-`actNotFound` envelope that is missing `account_data` (native) or `lines` (token). In that case:
- `getXrpAccountState` returned `funded:false` → native balance `0n`
- `getXrpAccountLines` did `result.lines ?? []` → token balance `0n`

On a fund-critical balance surface that surfaces a **real-looking `0`** for what may be a funded account / held token. The core-chain resolver #1220 replaced threw on the same malformed shape (it destructured `account_data.Balance` / spread `result.lines`), so this restores the prior fail-closed posture.

`rippleCall` already throws on `!res.ok` and on unexpected `status:'error'` — this only tightens the residual malformed-but-200 gap.

## Change
- `getXrpAccountState`: return unfunded **only** on explicit `actNotFound`; throw on a non-error response missing `account_data`.
- `getXrpAccountLines`: throw when a non-`actNotFound` page lacks an `Array` `lines` (a funded account with zero trust lines legitimately returns `lines: []`, so this doesn't affect the empty case).

## Tests
- 2 new red-on-revert regression tests (proven to fail against #1220 HEAD, pass with the fix).
- Full RN Ripple suite green: 20/20 across `getCoinBalance.test.ts` + `ripple-rpc.test.ts`.
- `format:check` clean (repo `.config/.prettierrc`).

### DIFF
--- a/packages/sdk/src/platforms/react-native/chains/ripple/rpc.ts
+++ b/packages/sdk/src/platforms/react-native/chains/ripple/rpc.ts
@@ -99,7 +99,7 @@ export async function getXrpAccountState(
     { account: address, strict: true, ledger_index: 'current' },
     signal
   )
-  if (result.error === 'actNotFound' || !result.account_data) {
+  if (result.error === 'actNotFound') {
     return {
       address,
       sequence: 0,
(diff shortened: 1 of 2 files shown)


gomesalexandre (approved):
lgtm. closes the malformed-envelope fail-open in the RN XRPL balance path.

- `account_info` now only treats explicit `actNotFound` as unfunded and throws on a non-error envelope missing `account_data`
- `account_lines` now throws when a non-`actNotFound` page is missing a `lines` array, while still allowing the legit `lines: []` zero-trustline case
- targeted regression tests cover both fail-closed branches

fund-safety surface, tight diff, green CI. approved.

Issue #3, opened 2025-08-31
[ADD]: VultisigSDK Architecture


jpthor:
# Vault-Centric Architecture Plan

## Executive Summary

VultisigSDK is a clean, vault-centric architecture with three primary classes:

- **`VultisigSDK`** - SDK
- **`VaultManager`** - Vault manager for multiple vaults
- **`Vault`** - Vault instance

### Minimal Integration Example

```typescript
// 1. Import and initialize SDK
import { VultisigSDK, VaultManager } from 'vultisig-sdk'
const sdk = new VultisigSDK()

// 2. Create vault with progress updates
const vault = await VaultManager.create('My Vault', {
  password: 'password',
  email: 'user@example.com',
  onProgress: (step) => {
    console.log(`${step.step}: ${step.progress}% - ${step.message}`)
  }
})
await vault.verifyEmail('1234')

// 3. Get address/balance for transaction
const ethAddress = await vault.address('ethereum')
const ethBalance = await vault.balance('ethereum')          // Native ETH balance

// 4. Sign transaction with progress updates
const signature = await vault.sign({
  transaction: { to: '0x...', value: '1000000000000000000' },
  chain: 'ethereum',
  onProgress: (step) => {
    console.log(`${step.mode} signing: ${step.progress}% - ${step.message}`)
  }
})

console.log('Transaction signed:', signature.txHash)
```

### Secure Vault Example

```typescript
// 1. Create secure vault with multi-device setup
const vault = await VaultManager.create('My Secure Vault', {
  type: 'secure',
  keygenMode: 'local',
  onProgress: (step) => {
    console.log(`${step.step}: ${step.progress}% - ${step.message}`)
  }
})

// 2. Get address/balance for transaction
const btcAddress = await vault.address('bitcoin')
const btcBalance = await vault.balance('bitcoin')        // Native BTC balance

// 3. Multi-device signing with threshold
const signature = await vault.sign({
  transaction: { to: 'bc1q...', value: '50000' },
  chain: 'bitcoin',
  signingMode: 'relay',
  onProgress: (step) => {
    console.log(`${step.mode} signing: ${step.progress}% - ${step.message}`)
  }
})
```

# Architecture

## VultisigSDK Class
The SDK exposes an API for vault operations that are not specific to a Vault or set of Vaults. 
It contains SDK-wide config and holds the entry points for the underlying managers. 

```typescript
class VultisigSDK {
  // === PRIVATE PROPERTIES ===
[...]

PR #1402, opened 2026-07-18
fix(sdk): adapt decoded envelopes for policy helpers

### LINKED ISSUE #1401: public API: decodeFromToolResult and policy helpers still require a consumer-written envelope adapter
## Summary

`@vultisig/sdk` currently publishes a decoder envelope and a policy envelope that are incompatible out of the box, forcing downstream consumers to maintain their own adapter.

## Evidence

- Decoder surface publishes:
  - `packages/sdk/src/tools/decode/types.ts:44-89`
  - fields include `chain`, `family`, `kind`, `recipient`, `asset`, `amount: string`, `spender`, `decoded`, `decodeError`
- Policy surface publishes:
  - `packages/sdk/src/tools/policy/types.ts:110-129`
  - fields include `chainId`, `recipient`, `asset`, `amount?: bigint | null`, `decoded`, `decodeError`
- Public policy consumers read the policy shape, not the decoder shape:
  - `packages/sdk/src/tools/policy/evaluatePolicy.ts:30-91`
  - `packages/sdk/src/tools/policy/checkInvariants.ts:26-107`

## Impact

Any first-party consumer that wants to run `decodeFromToolResult()` and then immediately evaluate policy/invariants has to invent a local `{ chain -> chainId, amount string -> bigint }` adapter. That is exactly the kind of cross-repo drift seam this campaign keeps finding.

## Fix sketch

Either:
- unify the two public envelope contracts, or
- publish a canonical SDK adapter and make it the documented integration path

## Effort

Small

## Report

Round 38 report mirror: https://gist.github.com/gomesalexandre/45ca3c8ef65dd975e138989bd013467c

gomesalexandre:
## Summary
- add `toPolicyEnvelope` / `policy.fromDecodedEnvelope` so `decodeFromToolResult()` output can feed policy/invariant helpers without a consumer-owned shim
- closes vultisig/vultisig-sdk#1401

## Why
The SDK currently publishes a decoder envelope with `chain` + `amount: string` and a policy envelope with `chainId` + `amount: bigint | null`. First-party consumers have to write their own adapter before the two public surfaces compose.

## Changes
- add `packages/sdk/src/tools/policy/toPolicyEnvelope.ts`
- expose it as both `toPolicyEnvelope` and `policy.fromDecodedEnvelope`
- add regression coverage for direct adaptation, decode→policy composition, and fail-closed invalid amount handling
- add a changeset for the public API addition

## Verification
- `yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/tools/policy/policy.test.ts`
- `yarn build:shared`
- `yarn build:sdk`
- `yarn cli:build`
- built-dist runtime proof:
  - `{"decodedChain":"ethereum","adaptedAmount":"1000000","verdict":"PASS","namespace":true}`

## Report
- Round 38 report mirror: https://gist.github.com/gomesalexandre/45ca3c8ef65dd975e138989bd013467c


packages/sdk/src/tools/policy/toPolicyEnvelope.ts
```diff
@@ -0,0 +1,45 @@
+import type { Envelope as DecodedEnvelope } from '../decode/types'
+
+import type { Envelope as PolicyEnvelope } from './types'
+
+const INTEGER_STRING = /^-?\d+$/
```
NeOMakinG, line 5: `preferably-blocking:` the `-?` admits negative amounts, and a negative amount does not make the policy suspicious — it makes the amount check **not run at all**. `evaluatePolicy.ts` gates drift on `envAmount != null && envAmount > 0n`, and `checkInvariants.ts` gates I5 (never-exceed-balance) on `envelope.amount > 0n`; a negative bigint satisfies neither.

Reproduced on this branch from real Cosmos wire bytes (`Coin.amount` is a raw proto3 **string**, so this is attacker-shapeable, not hypothetical):

```
MsgSend Coin.amount = "-1000000"
adapted  = { chainId:'cosmoshub-4', recipient:'cosmos1attacker', amount:-1000000n, decoded:true }
evaluate = PASS            // against a claim of 5 ATOM
invariants(balance 1000000n) = [I2 only]   // I5 skipped
```

Mutation-proved uncovered: deleting `-?` leaves the suite at 31/31 green.

```suggestion
const INTEGER_STRING = /^\d+$/
```

packages/sdk/src/tools/policy/toPolicyEnvelope.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * `sdk.decode.fromToolResult()` intentionally preserves raw atomic amounts as
+ * strings so the envelope survives JSON round-trips without bigint loss. The
+ * policy layer, on the other hand, wants `bigint` for drift checks and balance
+ * invariants. Consumers should not have to keep re-implementing the same
+ * `{ chain -> chainId, amount string -> bigint }` shim in every app/backend.
+ */
+export function toPolicyEnvelope(envelope: DecodedEnvelope): PolicyEnvelope {
+  const amount = parsePolicyAmount(envelope.amount)
+  const invalidAmount = envelope.amount.trim() !== '' && amount === null
```
NeOMakinG, line 18: `should-fix:` `envelope.amount.trim()` runs with no runtime type check on a **public exported** function whose own docblock advertises surviving JSON round-trips. Four reachable shapes throw on this branch:

```
{ amount: undefined }  -> THREW: Cannot read properties of undefined (reading 'trim')
{ amount: 1000000 }    -> THREW: amount.trim is not a function
{ amount: 1000000n }   -> THREW: amount.trim is not a function
undefined              -> THREW: Cannot read properties of undefined (reading 'amount')
```

`decode/fromToolResult.ts` explicitly promises *"this never throws — decode failures surface as `Envelope{ decoded:false, decodeError }` so callers handle them uniformly"*. The adapter breaks that contract for the whole decode→policy pipeline, and a caller that try/catches the composition into a default envelope fails **open**.

Suggested: `if (typeof envelope?.amount !== 'string')` → return a `decoded:false` envelope rather than throwing.

NeOMakinG (commented):
Two-lane review (Claude + Codex). The adapter is a **REBUILD**, not a cast — every field not named in the object literal at `toPolicyEnvelope.ts:20-29` is dropped, and three of them (`kind`, `spender`, `family`) are exactly the ones that tell a transfer apart from an approval. Separately, the "fail-closed invalid amount" branch this PR advertises is, measured end-to-end, a **verdict downgrade**: it disarms I1/I3/I7 and turns a `BLOCK` into a `WARN`. Everything below was reproduced by running the branch, not read off the diff.

## testing approach

- Worktree at `185fc9c5` (`/tmp/wt-sdk-1402`), node_modules symlinked from the canonical checkout, probes run as throwaway vitest files inside `packages/sdk` on the CI-real path (`vitest run --config tests/unit/vitest.config.ts`).
- Built real wire bytes for every probe: viem `serializeTransaction` for the EVM cases, `cosmjs-types` `TxRaw`/`TxBody`/`MsgSend` for the Cosmos cases — no hand-built envelopes where the wire could produce one.
- Full existing suite green on branch: `tests/unit/tools/policy/policy.test.ts` 31/31. `tsc --noEmit` clean for this diff (the two `auxiliaryData` errors in `packages/core` are pre-existing on `main`).
- **Mutation-proved the new coverage.** Deleting `-?` from `INTEGER_STRING` (i.e. applying my own fix #3) leaves **31/31 green** → the sign has zero test coverage. Forcing `invalidAmount = false` correctly turns the new fail-closed test red → that one branch is genuinely covered.
- CI reality check: `check-runs total_count` = **4** (real jobs ran, not the zero-job tick). Unit Tests + Integration pass; **Lint and Type Check is RED** — see should-fix #5.
- Codex ran as an independent second lane on Cosmos semantics + verdict propagation; two of its findings (#1 below and the unknown-msg half of #2) are folded in and were re-verified here with runnable evidence.

## blocking

**1. The "fail-closed" amount branch silently disarms I1 / I3 / I7 and downgrades `BLOCK` → `WARN`.** `toPolicyEnvelope.ts:25-28`

Setting `decoded: false` does not narrowly fail the amount check — it turns off *every other check in the policy layer*. `evaluatePolicy.ts:31` short-circuits any undecoded envelope to `WARN`, and `checkInvariants.ts:29` gates I1/I2/I3/I5/I7 on `decoded === true`. So an unparseable amount string erases an unrelated recipient mismatch. Same envelope, only the amount string differs:

```
claim = { chain: 'ethereum', recipient: '0xBob', asset: 'USDC', amount: '1', amountUnits: 'human' }
envelope recipient = '0xAttackerWallet', chain = 'base'   // mismatched on BOTH

amount '1000000' -> evaluate   = BLOCK "chain mismatch: user claimed \"ethereum\", envelope has \"base\""
                    invariants = ["I3_chain_matches_intent","I1_recipient_matches_intent"]

amount '1e6'     -> evaluate   = WARN "envelope not decoded: invalid decoded atomic amount: 1e6"
                    invariants = []            <-- ZERO. recipient + chain mismatch GONE.
```

And I7, the memo invariant whose own doc says "dropped memo = funds lost or stuck on exchange deposits / XRPL tags / IBC PFM":

```
userMemo 'EXCHANGE-TAG-9911', envelopeMemo ''
amount '1000000' -> ["I7_memo_preserved"]
amount '1e6'     -> []
```

This is attacker-reachable, not theoretical. Cosmos `Coin.amount` is a raw proto3 **string** off the wire (`decode/cosmos.ts:82`, `MsgSend.decode`), copied verbatim into `Envelope.amount`. Anyone who can shape the unsigned bytes can write `"1e6"` into the Coin and thereby silence the recipient check on the same transaction. The adapter must never be able to *reduce* the number of checks that run.

Suggested fix: leave `decoded` and `decodeError` untouched. `amount: null` already means "unknown → skip the amount check" in this layer (`checkInvariants.ts:57` comment: *"nil envelope amount = decoder could not parse (fail open), NOT sends zero"*), so the amount half already degrades on its own. If an unparseable amount must be escalated, surface it additively (e.g. an `amountParseError` field documented as caller-must-BLOCK) rather than by flipping the flag that gates the other six invariants.

**2. `kind` and `spender` are dropped, so `policy.evaluate` PASSes an `approve()` to an attacker under a send intent.** `toPolicyEnvelope.ts:20-29`

`decode/evm.ts:161` deliberately empties `recipient` for an approval — *"An approval has no value recipient — the meaningful target is the spender"* — and puts the counterparty in `spender`. The adapter drops `spender` and drops `kind`, so the policy envelope for an approval is indistinguishable from a transfer with an unknown recipient, and `evaluatePolicy.ts:54` skips the recipient check whenever either side is empty. Reproduced against real EIP-1559 bytes:

```
user claim : send 1 USDC to 0x7099...79C8 on ethereum
wire       : USDC.approve(0x...DeaDBeEF, 1000000)

decoded  = { kind:'approve', recipient:'', spender:'0x...DeaDBeef', amount:'1000000', decoded:true }
adapted  = { chainId:'ethereum', recipient:'', asset:{symbol:'USDC'}, amount:1000000n, decoded:true }
verdict  = { result:'PASS', reason:'envelope matches intent claim', diff:[] }
```

`PASS`, empty diff. (The unlimited-approval variant only BLOCKs by accident, on amount drift — a right-sized `approve` sails through.) The same drop makes the Cosmos default branch fail open: an unsupported single-message tx (IBC `MsgTransfer`, `MsgBeginRedelegate`, `MsgWithdrawDelegatorReward` — `decode/cosmos.ts` default branch) decodes to `kind:'unknown'` with `recipient:''` and `amount:''`, and adapts to a `PASS` on chain alone:

```
/ibc.applications.transfer.v1.MsgTransfer, claim "send 5 ATOM to cosmos1bob"
adapted  = { chainId:'cosmoshub-4', recipient:'', amount:null, decoded:true }
evaluate = PASS
```

Note in both cases `checkInvariants` *does* catch it (I1 treats an empty envelope recipient as a divergence) while `evaluatePolicy` does not — the two helpers disagree, and `evaluatePolicy` is the one this PR's own composition test demonstrates. Whatever the fix, please make the two agree.

Suggested fix: carry `kind` (and `spender`) onto the policy `Envelope` as optional additive fields and have `evaluatePolicy` refuse to PASS a `kind` of `approve` / `contractCall` / `unknown` against a plain-send claim; or, if the policy layer is to stay send-only, have the adapter fail closed on any `kind !== 'transfer'` rather than laundering it into a transfer-shaped envelope.

## preferably-blocking

**3. `INTEGER_STRING = /^-?\d+$/` admits negative amounts, which turn the amount check OFF entirely.** `toPolicyEnvelope.ts:5`

`evaluatePolicy.ts:71` gates the drift check on `envAmount != null && envAmount > 0n`, and `checkInvariants.ts:73` gates I5 (never-exceed-balance) on `envelope.amount > 0n`. A negative bigint satisfies neither, so it is not "a suspicious amount" — it is *no amount check at all*. Reproduced from real Cosmos wire bytes with `Coin.amount = "-1000000"` (again: raw wire string, attacker-shapeable):

```
adapted  = { chainId:'cosmoshub-4', recipient:'cosmos1attacker', amount:-1000000n, decoded:true }
evaluate = PASS   // against a claim of 5 ATOM
invariants (with balance 1000000n) = [I2 only]   // I5 skipped
```

The adapter already fail-closes `'1e6'`; the `-?` is precisely what lets through the one shape that disables the check. Mutation-proved uncovered: removing `-?` keeps the suite at 31/31 green. Fix is one character: `/^\d+$/`.

## should-fix

**4. The adapter throws on four reachable input shapes; `decodeFromToolResult` is documented as never throwing.** `toPolicyEnvelope.ts:17-18,33`

`envelope.amount.trim()` is called with no runtime type check on a **public exported** function whose own docblock advertises JSON round-trips:

```
{ amount: undefined }  -> THREW: Cannot read properties of undefined (reading 'trim')
{ amount: 1000000 }    -> THREW: amount.trim is not a function
{ amount: 1000000n }   -> THREW: amount.trim is not a function
undefined              -> THREW: Cannot read properties of undefined (reading 'amount')
```

`decode/fromToolResult.ts` promises *"Like the Go reference, this never throws — decode failures surface as `Envelope{ decoded: false, decodeError }` so callers handle them uniformly."* This adapter breaks that contract for the whole decode→policy pipeline the PR is introducing, and a caller that wraps the composition in try/catch with a default envelope fails **open**. The `string | number | bigint` producer vs narrow-string reader mismatch has been the single most repeated defect across this repo's safety surfaces. Guard with `typeof envelope?.amount === 'string'` and return a `decoded:false` envelope instead.

**5. CI is RED on this diff and the PR's Verification section never ran lint.** All four files this PR touches:

```
packages/sdk/src/index.ts                             494:1  simple-import-sort/exports
packages/sdk/src/tools/index.ts                       300:1  simple-import-sort/exports
packages/sdk/src/tools/policy/toPolicyEnvelope.ts       1:1  simple-import-sort/imports
packages/sdk/tests/unit/tools/policy/policy.test.ts     1:1  simple-import-sort/imports
```

`toPolicyEnvelope` was inserted after `runSkipSwap` but before `sanitizeAmount` in both barrel files. All four are `--fix`-able.

**6. `patch` understates a new public export.** `.changeset/policy-decoded-envelope-adapter.md:2-3` — adding an exported symbol to a published package is a `minor` under semver, and downstream installs resolve `@vultisig/sdk` on a caret. Also: `@vultisig/cli` is bumped but this PR changes nothing under `packages/cli` — intentional?

**7. `toEqual` on an adapter whose entire job is which keys survive.** `tests/unit/tools/policy/policy.test.ts` — `toEqual` cannot tell `{a:1}` from `{a:1, b:undefined}`, which is exactly the distinction under review here. Use `toStrictEqual`, and add an explicit assertion that `kind`/`spender`/`family` are dropped **on purpose** so finding #2 either gets fixed or gets a test that documents the choice.

## q

**8. Two amount parsers in one package with different accepted sets — and the divergence is backwards in both directions.** `toPolicyEnvelope.ts:5` rejects `'+1000000'` and accepts `'-1000000'`; its sibling `policy/amount.ts#parseAmountBig` uses `/^[+-]?\d+$/` and accepts both. Verified on branch:

```
'+1000000'   -> amount:null, decoded:false      (adapter rejects a harmless leading +)
'-1000000'   -> amount:-1000000n, decoded:true  (adapter accepts the dangerous one)
'0001000000' -> 1000000n
' 1000000 '  -> 1000000n
'١٠٠٠٠٠٠'     -> amount:null, decoded:false      (good — bare BigInt() would have accepted these)
```

Was the divergence from `parseAmountBig` deliberate? If so it's worth a comment; if not, the two should share one parser.

**9. There are currently zero in-repo production callers** of `evaluatePolicy` / `checkInvariants` / `toPolicyEnvelope` — the policy layer is published-only, and I found no consumer of it in `agent-backend-ts`, `mcp-ts` or `vultiagent-app` either. So the "consumers have to write their own adapter" motivation is forward-looking rather than replacing an existing shim, and nothing downstream regresses today. Is there a companion PR that will be the first caller? That's the one that decides whether findings #1-#3 are latent or live.

## risk

Medium-high for a 143-line additive PR. Nothing regresses today because the policy layer has no production callers yet — but this PR's whole purpose is to bless `policy.evaluate(claim, policy.fromDecodedEnvelope(decoded))` as the supported one-liner, and that exact one-liner currently returns `PASS` for an `approve()` to an attacker, `PASS` for a negative-amount Cosmos send, `PASS` for an unsupported Cosmos message, and `WARN`-with-no-violations for a tx that mismatches on recipient, chain and memo. Whichever consumer lands first inherits all four.

## verdict

**COMMENTED** — draft, so no approval either way, and no fixes pushed to your branch. The shape of the adapter is right and the fail-closed *instinct* is right; it's the blast radius of `decoded:false` (#1) and the two dropped discriminator fields (#2) that need to change before a first caller exists. Findings #1, #2, #3, #4 all have reproductions above that should convert straight into regression tests — please make each red-before / green-after rather than only asserting the happy path.

<sub>Reviewed by Claude (correctness/fund-safety lane) + Codex (adversarial second lane). All findings reproduced by execution on `185fc9c5`.</sub>

packages/sdk/src/tools/policy/toPolicyEnvelope.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  return {
+    chainId: envelope.chain,
+    recipient: envelope.recipient,
+    asset: envelope.asset,
+    amount,
+    decoded: invalidAmount ? false : envelope.decoded,
```
NeOMakinG, line 25: `blocker:` flipping `decoded` here does not narrowly fail the amount — it turns off **every other check in the policy layer**. `evaluatePolicy.ts:31` short-circuits any undecoded envelope to a blanket `WARN`, and `checkInvariants.ts:29` gates I1/I2/I3/I5/I7 on `decoded === true`. So an unparseable amount string erases an unrelated recipient mismatch.

Same envelope, only the amount string differs (run on this branch):

```
claim: { chain:'ethereum', recipient:'0xBob', amount:'1', amountUnits:'human' }
envelope: recipient '0xAttackerWallet', chain 'base'   // mismatched on BOTH

amount '1000000' -> evaluate   = BLOCK "chain mismatch"
                    invariants = [I3_chain_matches_intent, I1_recipient_matches_intent]

amount '1e6'     -> evaluate   = WARN "envelope not decoded: invalid decoded atomic amount: 1e6"
                    invariants = []      <-- ZERO. recipient + chain mismatch GONE.
```

I7 goes the same way (control fires `I7_memo_preserved`, the `1e6` variant fires nothing) — the invariant whose own comment reads *"dropped memo = funds lost or stuck on exchange deposits / XRPL tags / IBC PFM"*.

Attacker-reachable: Cosmos `Coin.amount` is a raw proto3 string copied verbatim into `Envelope.amount`, so writing `"1e6"` into the Coin silences the recipient check on that same transaction.

Suggested fix: leave `decoded`/`decodeError` untouched. `amount: null` already means "unknown → skip the amount check" here (`checkInvariants.ts`: *"nil envelope amount = decoder could not parse (fail open), NOT sends zero"*). If an unparseable amount must escalate, surface it additively (an `amountParseError` field documented as caller-must-BLOCK) rather than by flipping the flag that gates the other six invariants.

packages/sdk/src/tools/policy/toPolicyEnvelope.ts
```diff
@@ -0,0 +1,45 @@
...
+ * invariants. Consumers should not have to keep re-implementing the same
+ * `{ chain -> chainId, amount string -> bigint }` shim in every app/backend.
+ */
+export function toPolicyEnvelope(envelope: DecodedEnvelope): PolicyEnvelope {
+  const amount = parsePolicyAmount(envelope.amount)
+  const invalidAmount = envelope.amount.trim() !== '' && amount === null
+
+  return {
+    chainId: envelope.chain,
+    recipient: envelope.recipient,
```
NeOMakinG, line 22: `blocker:` this is a **rebuild**, not a cast, so `kind`, `spender` and `family` are dropped — and `kind`/`spender` are exactly the fields that tell an approval apart from a transfer.

`decode/evm.ts` deliberately empties `recipient` for an approval (*"An approval has no value recipient — the meaningful target is the spender"*) and puts the counterparty in `spender`. Drop `spender`, drop `kind`, and `evaluatePolicy.ts:54` skips the recipient check because one side is empty. Against real EIP-1559 bytes on this branch:

```
claim : send 1 USDC to 0x7099...79C8 on ethereum
wire  : USDC.approve(0x...DeaDBeEF, 1000000)

decoded = { kind:'approve', recipient:'', spender:'0x...DeaDBeef', amount:'1000000', decoded:true }
adapted = { chainId:'ethereum', recipient:'', asset:{symbol:'USDC'}, amount:1000000n, decoded:true }
verdict = { result:'PASS', reason:'envelope matches intent claim', diff:[] }
```

(The *unlimited* approval only BLOCKs by accident, on amount drift — a right-sized `approve` sails through.)

Same drop opens the Cosmos default branch: an unsupported single-message tx (IBC `MsgTransfer`, `MsgBeginRedelegate`, `MsgWithdrawDelegatorReward`) decodes to `kind:'unknown'`, `recipient:''`, `amount:''` and adapts to `PASS` on chain alone.

Worth noting `checkInvariants` *does* catch both (I1 treats an empty envelope recipient as a divergence) while `evaluatePolicy` does not — the two helpers disagree, and `evaluatePolicy` is the one this PR's own composition test demonstrates.

Suggested: carry `kind`/`spender` onto the policy `Envelope` as optional additive fields and have `evaluatePolicy` refuse to PASS `kind` of `approve`/`contractCall`/`unknown` against a plain-send claim; or fail closed in the adapter on any `kind !== 'transfer'` rather than laundering it into a transfer-shaped envelope.

.changeset/policy-decoded-envelope-adapter.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
```
NeOMakinG, line 3: `should-fix:` adding an exported symbol to a published package is a `minor` under semver, not a `patch` — and downstream installs resolve `@vultisig/sdk` on a caret, so the distinction is what gives consumers a signal that the public surface grew. `q:` also, `@vultisig/cli` is bumped but this PR changes nothing under `packages/cli` — intentional?

NeOMakinG (approved):
LGTM

gomesalexandre:
Fixed all three blockers, plus the separate approve/spender-dropping gap you flagged:

1. **Negative amounts rejected** — `INTEGER_STRING` no longer admits a leading `-`, so a raw attacker-shapeable Cosmos `Coin.amount` string like `-1000000` is treated as unparseable instead of silently disabling the amount check.
2. **Non-string `amount` type-guarded** — never throws on undefined/number/bigint, matches `decodeFromToolResult`'s documented 'never throws' contract.
3. **The masking bug (most severe)**: an unparseable amount no longer flips `decoded` to false. It surfaces additively via a new `amountParseError` field, leaving `decoded`/`decodeError` as the underlying decoder's own status — so a genuine recipient/chain mismatch on the SAME envelope still fires (verified with your exact repro shape: BLOCKs with a valid amount, still catches the recipient mismatch via checkInvariants when the amount is '1e6').
4. **Non-transfer kind fails closed**: `approve`/`contractCall`/`delegate`/`undelegate`/`unknown` no longer get rebuilt into a transfer-shaped `decoded:true` envelope — they now return `decoded:false`, routing through the existing WARN-on-undecoded fallback instead of a false PASS. Verified against your unlimited-approval repro.

4 new regression tests reproducing your exact scenarios, 34/34 pass, tsc clean, rebased onto main. Went with the additive-field approach for #3 per your suggestion rather than touching evaluatePolicy/checkInvariants — both already skip the amount-specific checks correctly when amount is null while still running recipient/chain/memo, so no changes needed there. Excellent finding, this was a real one.

rcoderdev:
Closing this because the stale policy adapter still throws on undefined input and can let policy evaluation skip an invalid amount, while its approval is for an older head and no production consumer was found.

Issue #1403, opened 2026-07-18
cli tx_ready: fail closed on non-native non-EVM token sends until envelopes carry token-aware decimals/ids


gomesalexandre:
## Summary
The CLI executor's non-EVM `tx_ready` path converted base units with the chain's native decimals and then still allowed a token symbol through, which would mis-scale a future SPL/TRC-20/XRPL-style token send. This round ships the fail-closed guard instead of guessing.

## Evidence
_Pre-fix behavior on main before the patch shipped in PR below:_
- `clients/cli/src/agent/executor.ts:2085-2099` converted `txArgs.amount` with native decimals and still forwarded `resolved.labels.token_resolved` as a token symbol.
- `clients/cli/src/agent/executor.ts:881-887` then sent that parsed amount through `vault.send(...)`.

## Impact
Future non-native non-EVM token envelopes could sign with the wrong amount magnitude.

## Fix sketch
Keep the new fail-closed guard until the envelope contract carries token-aware metadata (token id + decimals), then centralize that parser logic in shared SDK/CLI code.

## Effort
Shipped

## Report
- {gist_url}
- Campaign index artifact: https://claude.ai/code/artifact/be2deadf-6195-4dfa-a910-b269b9af7a84

Issue #4, opened 2025-09-01
[ADD]: ChainHandlers


jpthor:
# ChainHandler Interface Implementation

## Summary

Implement a unified `ChainHandler` interface to provide consistent, chain-agnostic operations across all supported blockchain networks in the Vultisig SDK. This interface will standardize how the SDK interacts with different chains, reducing complexity and improving maintainability.

## Usage Examples

```typescript
// Get a chain handler for Ethereum
const ethereum = sdk.getChainHandler('ethereum');

// Basic operations - same API across all chains
const address = await ethereum.address();
const balance = await ethereum.balance();
const ethBalance = await ethereum.balance('ETH');
const usdcBalance = await ethereum.balance('USDC');

// Send native currency
await ethereum.send({
  to: '0x742d35Cc6634C0532925a3b8D4f25F7CF9c4a6Ad',
  amount: '0.1'
});

// Send tokens
await ethereum.send({
  to: '0x742d35Cc6634C0532925a3b8D4f25F7CF9c4a6Ad',
  amount: '100',
  tokenId: 'USDC'
});

// Swap tokens
await ethereum.swap({
  fromToken: 'ETH',
  toToken: 'USDC',
  amount: '0.5',
  slippage: 1.0
});

// Same interface works for Bitcoin
const bitcoin = sdk.getChainHandler('bitcoin');
const btcAddress = await bitcoin.address();
const btcBalance = await bitcoin.balance();

// Or Thorchain
const thorchain = sdk.getChainHandler('thorchain');
await thorchain.swap({
  fromToken: 'BTC.BTC',
  toToken: 'ETH.ETH',
  amount: '0.01'
});

// Generic function calls (chain-specific features)
await ethereum.function({
  type: 'contract_call',
  contractAddress: '0x...',
  functionSignature: 'transfer(address,uint256)',
  params: ['0x742d35Cc6634C0532925a3b8D4f25F7CF9c4a6Ad', '1000000']
});

// Thorchain-specific function calls
await thorchain.function({
  type: 'MsgBond',
  params: {
    nodeAddress: 'thor1...',
    bondAmount: '100000000000', // 1000 RUNE in 1e8 format
    asset: 'THOR.RUNE',
    bondAddress: 'thor1...',
    operatorFee: '1000' // 10 basis points
  }
});

// Token management
ethereum.addToken({
  id: 'LINK',
  symbol: 'LINK',
  name: 'Chainlink',
[...]

PR #1404, opened 2026-07-18
fix(cli): fail closed on non-native non-EVM tx_ready sends

gomesalexandre:
1|## Summary
     2|- fail closed when the CLI `tx_ready` non-EVM path sees a non-native token symbol without token-aware decimals/ids
     3|- add regression coverage for the supported native path and the new rejection path
     4|- add a changeset for the CLI/SDK patch release
     5|
     6|## Why
     7|Before this patch, `parseNonEvmEnvelope()` converted base units using the chain's native decimals and then still allowed `resolved.labels.token_resolved` through as a token symbol. A future SPL/TRC-20/XRPL-style token envelope would therefore be re-parsed by `vault.send()` at the wrong magnitude. This patch blocks that unsupported path until the envelope contract carries token-aware metadata.
     8|
     9|## Test plan
    10|- `yarn build:shared`
    11|- `yarn build:sdk`
    12|- `yarn workspace @vultisig/cli vitest run src/agent/__tests__/executor.envelope.test.ts`
    13|- `yarn cli:build`
    14|- built runtime proof: `cli:build succeeded but exported executor entrypoint was not emitted as a standalone dist JS file for direct Node import; using targeted Vitest + build receipts instead.`
    32|
    33|## Related
    34|- Closes https://github.com/vultisig/vultisig-sdk/issues/1403
    35|- Round report: https://gist.github.com/gomesalexandre/b06e95d1ecdfd3c8e372674d77101343
    36|

## Summary by CodeRabbit

* **Bug Fixes**
  * Added fail-closed handling for non-native token transfers when required token metadata is unavailable.
  * Prevented unsupported token transactions from using native-asset conversion rules.
  * Improved Solana transaction validation, including rejection of incomplete token transfer details.
  * Added support for parsing native Solana transactions correctly.

* **Chores**
  * Documented the patch release for the SDK and CLI.


NeOMakinG (commented):
Draft — posting `COMMENT`, not approving. Two lanes ran (Claude load-bearing + Codex adversarial); Codex was alive and independently landed the `token_resolved`-only residue and the TON/`GRAM` alias, both of which I then reproduced.

The direction of this change is right and the premise holds: I confirmed against the registry that `knownTokens` really does carry non-EVM tokens (`packages/core/chain/coin/knownTokens/index.ts` — Solana/Tron/Ton/Ripple/Polkadot; `cosmos.ts` — Cosmos/Osmosis), so pre-patch `vault.send({symbol:'USDD'})` on Tron resolves a real 18-decimal token while `amountDecimal` was derived with TRX's 6 — a 10^12 **over**-send, not a theoretical one. So this is worth fixing. But as written the guard is red in CI, misses the common wire shapes, and refuses some honest native sends.

## testing approach

Worktree `/tmp/wt-sdk-1404` at `24ab588d`, `node_modules` symlinked from the canonical checkout, `node_modules/.bin/vitest --root clients/cli`.

- **Mutation-verified the new test.** Reverted the throw back to `symbol = tokenResolved`; `fails closed on non-native non-EVM token envelopes` goes red, `keeps native non-EVM sends` stays green. It is a real guard, not a passthrough. 
- **7 adversarial probes** against the real `parseNonEvmEnvelope` export (not a fixture copy). All 7 reproduced; results inline below.
- **Read the CI log rather than the tick.** `Unit Tests` and `Integration Tests` are green, but `Lint and Type Check` and `Test Summary` are **red**, and the lint error is introduced by this diff.
- Local pre-existing noise: 7 `storeServerTransaction` tests fail here with `resolveChainReference is not a function` — the prebuilt `@vultisig/sdk` dist in the canonical `node_modules` predates `packages/sdk/src/index.ts:60`. Environmental; CI `Unit Tests` is green on the same file. Not attributed to the PR.

## blocking

**1 — CI is red on a lint error this diff introduces.** `clients/cli/src/agent/executor.ts:2102`:

```
2102:7  error  'symbol' is always 'undefined' because it's never assigned  no-unassigned-vars
```

Removing `symbol = tokenResolved` deleted the *only* assignment, so `let symbol` is now provably dead and `NonEvmSendArgs.symbol` can never be anything but `undefined`. `Static quality gate: failure` is what fails `Test Summary`, so both red checks are this one error. Fix: drop the `let` and return `symbol: undefined` explicitly (or drop `symbol` from `NonEvmSendArgs` and let `vault.send` default), which also documents the invariant honestly.

**2 — The guard reads one key; the token identity travels on at least three others, and those shapes still build a signable native send.** The check is anchored on `serverTxData?.resolved?.labels?.token_resolved` and nothing else. Probes, all against the exported function:

| # | envelope | result |
|---|---|---|
| P3 | `resolved.labels.token_symbol = 'USDC'`, no `token_resolved`, Solana, amount `123000000` | **builds** `{chain:'Solana', to:'DEST', amount:'0.123'}` — native SOL |
| P5 | no `resolved` at all; `txArgs.token_id = 'EPjFWdd5…Dt1v'`, `txArgs.token_symbol='USDC'`, Solana | **builds** `amount:'0.123'` — native SOL |
| P6 | no `resolved`; `txArgs.denom = 'ibc/F663521BF1836B00F5F177680F74BFB9A8B5654A694D0D2BC249E03CF2509013'` (USDC on Cosmos), amount `123000000` | **builds** `{chain:'Cosmos', amount:'123'}` — **123 ATOM** |

P6 is the fund-safety one: the user asked for 123 USDC and the CLI would sign 123 ATOM out of the same vault.

P3 is the WYSIWYS one. `executor.ts:460` — the confirmation string the user actually reads — is:

```ts
const symbol = labels.token_resolved || labels.token_symbol || ''
```

So the display side *already* honours `token_symbol` as a first-class alias while the signing side does not. The user is prompted `send … USDC on Solana to …` and a native SOL send is what gets signed. This is exactly the display-vs-signed split we have been closing across the app repo; it survives here.

Suggested shape: invert the polarity. Instead of "throw if this one label says non-native", require positive proof of nativeness and refuse on *any* token indicator:

```ts
const tokenHints = [
  serverTxData?.resolved?.labels?.token_resolved,
  serverTxData?.resolved?.labels?.token_symbol,
  txArgs.token_symbol,
  txArgs.token_id,
  txArgs.contract_address,
  txArgs.mint,
  txArgs.jetton_master,
]
```
…plus `txArgs.denom` when it is not the chain's native denom, and refuse if any of them is set to something that isn't the native ticker. Fail-closed only works when the allowlist is the thing you enumerate.

**3 — String-only reader: a non-string `token_resolved` disables the guard entirely.** `typeof tokenResolved === 'string' && …` — when the field is present but not a string the guard is skipped *and* `symbol` stays `undefined`, i.e. it falls through to a native send rather than refusing. Probe P4, all four **build** `{amount:'0.123'}` on Solana:

```
P4 number/123          -> {"chain":"Solana","to":"DEST","amount":"0.123"}
P4 object/["USDC"]     -> {"chain":"Solana","to":"DEST","amount":"0.123"}
P4 object/{"s":"USDC"} -> {"chain":"Solana","to":"DEST","amount":"0.123"}
P4 object/String('USDC') -> {"chain":"Solana","to":"DEST","amount":"0.123"}
```

Note the two sibling readers immediately above (`to`, `amount`) both *throw* when the field is present-but-wrong-type. This one silently passes. Make it consistent: if `tokenResolved !== undefined && typeof tokenResolved !== 'string'` → throw `InvalidConfig`.

## preferably-blocking

**4 — Byte-exact comparison diverges from the SDK's own case-insensitive one, and refuses honest native sends.** `VaultBase.ts:1979` decides nativeness with

```ts
if (!symbol || symbol.toUpperCase() === native.ticker.toUpperCase()) return { ticker: native.ticker, decimals: native.decimals }
```

The new guard uses `tokenResolved !== nativeTicker`. Probe P2 — `token_resolved: 'sol'` on Solana — **throws** `non-native Solana token sends are not wired…`, i.e. a hard refusal on a plain native SOL send, on a difference the SDK itself considers no difference. Uppercase both sides.

**5 — `Chain.Ton`'s registry ticker is `GRAM`, not `TON`.** `packages/core/chain/coin/chainFeeCoin.ts:69`. Probe P1: a native TON send labelled `token_resolved: 'TON'` **throws** `non-native Ton token sends are not wired on this tx_ready path yet (resolved token 'TON')`.

To be fair to the PR this is *not* a regression — pre-patch the same envelope set `symbol='TON'` and then died downstream in `resolveTokenInfo` with `Token "TON" not found on Ton`, since `knownTokens[Chain.Ton]` has USDT/NOT/DOGS/CATI/HMSTR/STON/stTON/tsTON but no TON. So native TON over `tx_ready` has never worked. But the guard now permanently freezes that, and mislabels the chain's own fee coin as an unimplemented *token* send, which will mis-route whoever debugs it next. A native-ticker alias map (or reusing `resolveTokenInfo`'s comparison) closes #4 and #5 together and incidentally unbreaks TON.

For the record, the one chain where this reads oddly but is correct: `Chain.Noble`'s fee coin *is* `USDC` (chainFeeCoin.ts:139) — probe P7 confirms a Noble USDC send still builds (`amount:'1'` from `1000000`). Good.

## should-fix

**6 — The refusal fires too late in the flow.** `parseNonEvmEnvelope` is called at `executor.ts:875`, which is downstream of the user having been shown and accepted the confirmation prompt *and* downstream of `reserveBroadcast` taking the double-spend lock (`executor.ts:733`). The sibling non-EVM rejection in this same file — the multi-leg one at `executor.ts:307` — refuses in `storeServerTransaction` so the CLI never offers to sign the envelope at all. Doing the same here means the user sees "I can't sign token sends on Solana yet" instead of confirming a send and then getting an error. (Related: `pendingPayloads.clear()` only runs on the success path at `:899`, so the refused envelope stays buffered.)

## q

**7** — the changeset declares `'@vultisig/sdk': patch`, but no file under `packages/sdk/**` is touched by this diff. Intentional (companion release?) or should it be `@vultisig/cli` only?

- Native TON sends through `tx_ready` cannot succeed today, independent of this PR — see #5.
- `clients/cli/src/agent/__tests__/parseNonEvmEnvelope.test.ts` has a test named *"ignores cosmos-specific fields (denom, chain_id, sequence) — SDK overrides at sign time"* that asserts `denom` is dropped. For RUNE that is right; but it is also the assertion that codifies probe P6's hole, so whoever fixes #2 will have to narrow that test rather than delete it.

## risk

Medium-high. The guard as merged would turn CI red, would still sign the two most likely real SPL/IBC wire shapes as native sends, and would refuse some honest native sends (lowercase label, any TON). The fund-loss direction the PR is aiming at is real and worth closing — this just needs to bind to every token indicator rather than one label, and compare nativeness the way the SDK does.

## verdict

**COMMENTED** (draft — not approving, and deliberately not pushing fixes to your branch). Blocking items 1–3 before this leaves draft; 4–5 are a one-line comparison change that also fixes TON.

---
🤖 automated review — 2 lanes (Claude + Codex), 7 probes reproduced against the real export, new test mutation-verified.

NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because current main now uses the token-aware parseTxReadyEnvelope and resolveTokenRef flow, while this conflicted branch can misinterpret token units and would regress that newer implementation.

PR #1410, opened 2026-07-19
fix(sdk): publish tools/prep subpath

### LINKED ISSUE #1409: public API: publish @vultisig/sdk/tools/prep as a real package subpath
## Summary
`packages/sdk/src/tools/prep/index.ts` is already the canonical vault-free tx-builder barrel, but `@vultisig/sdk` still does not publish a real `@vultisig/sdk/tools/prep` package subpath.

## Evidence
- `packages/sdk/src/tools/prep/index.ts:1-68` already defines the dedicated prep barrel.
- `packages/sdk/src/tools/index.ts:331-392` already treats prep as a first-class public tools surface.
- `packages/sdk/package.json:76-101` exports `./tools/parse` and `./tools/defi`, but not `./tools/prep`.
- `packages/sdk/rollup.platforms.config.js:321-328` and `packages/sdk/rollup.types.config.js:100-104` only bundle `parse` and `defi` subpaths.

## Impact
First-party consumers that want only deterministic tx builders must either import the giant root surface or deep-import internals/dist files. That recreates the exact drift pressure this campaign keeps finding.

## Fix sketch
Publish `@vultisig/sdk/tools/prep` exactly like `tools/parse` / `tools/defi`:
- add an export-map entry in `packages/sdk/package.json`
- add dedicated JS bundles in `rollup.platforms.config.js`
- add dedicated DTS bundle in `rollup.types.config.js`
- add a public-API regression test and tarball/package-name smoke test

## Effort
Small / low risk.

## Report
Round 42 report: https://gist.github.com/gomesalexandre/57b4cc32aaae78161d6421e0080cc9ad

gomesalexandre:
## Summary
- publish `@vultisig/sdk/tools/prep` as a real package subpath
- add export-map / bundle / d.ts regression coverage for the new prep surface
- extend the tarball smoke proof to resolve and import `tools/prep`

## Test plan
- [x] yarn install --immutable
- [x] yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/public-api/toolsSubpathExports.test.ts
- [x] yarn build:sdk
- [x] yarn cli:build
- [x] yarn build:all
- [x] yarn lint
- [x] yarn sdk:smoke:subpaths

Closes #1409
Report: https://gist.github.com/gomesalexandre/57b4cc32aaae78161d6421e0080cc9ad

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -98,6 +98,19 @@
       "import": "./dist/tools/defi/index.js",
       "require": "./dist/tools/defi/index.cjs",
       "default": "./dist/tools/defi/index.cjs"
+    },
+    "./tools/prep": {
+      "types": "./dist/tools/prep/index.d.ts",
+      "browser": "./dist/tools/prep/index.js",
+      "worker": "./dist/tools/prep/index.js",
+      "react-native": "./dist/tools/prep/index.js",
+      "node": {
+        "import": "./dist/tools/prep/index.js",
+        "require": "./dist/tools/prep/index.cjs"
+      },
+      "import": "./dist/tools/prep/index.js",
+      "require": "./dist/tools/prep/index.cjs",
+      "default": "./dist/tools/prep/index.cjs"
     }
   },
   "scripts": {
--- a/packages/sdk/rollup.platforms.config.js
+++ b/packages/sdk/rollup.platforms.config.js
@@ -326,6 +326,10 @@ const configs = {
       input: './src/tools/defi/index.ts',
       distBase: 'tools/defi',
     }),
+    ...createToolsSubpathConfigs({
+      input: './src/tools/prep/index.ts',
+      distBase: 'tools/prep',
+    }),
   ],
   browser: {
     input: './src/platforms/browser/index.ts',
--- a/packages/sdk/tests/unit/public-api/toolsSubpathExports.test.ts
+++ b/packages/sdk/tests/unit/public-api/toolsSubpathExports.test.ts
@@ -10,9 +10,10 @@ const platformRollupConfig = readFileSync(path.join(sdkRoot, 'rollup.platforms.c
 const typesRollupConfig = readFileSync(path.join(sdkRoot, 'rollup.types.config.js'), 'utf8')
 
 describe('public API tools subpath exports', () => {
-  it('publishes dedicated export-map entries for parse and defi', () => {
+  it('publishes dedicated export-map entries for parse, defi, and prep', () => {
     const parseExport = sdkPackageJson.exports['./tools/parse']
     const defiExport = sdkPackageJson.exports['./tools/defi']
+    const prepExport = sdkPackageJson.exports['./tools/prep']
 
     expect(parseExport).toMatchObject({
       types: './dist/tools/parse/index.d.ts',
@@ -26,22 +27,34 @@ describe('public API tools subpath exports', () => {
       require: './dist/tools/defi/index.cjs',
       default: './dist/tools/defi/index.cjs',
     })
+    expect(prepExport).toMatchObject({
+      types: './dist/tools/prep/index.d.ts',
+      import: './dist/tools/prep/index.js',
+      require: './dist/tools/prep/index.cjs',
+      default: './dist/tools/prep/index.cjs',
+    })
 
     expect(JSON.stringify(parseExport)).not.toContain('dist/index.node')
     expect(JSON.stringify(defiExport)).not.toContain('dist/index.node')
+    expect(JSON.stringify(prepExport)).not.toContain('dist/index.node')
   })
 
-  it('keeps dedicated JS and d.ts bundle generation wired for both subpaths', () => {
+  it('keeps dedicated JS and d.ts bundle generation wired for all tool subpaths', () => {
     expect(platformRollupConfig).toContain("input: './src/tools/parse/index.ts'")
     expect(platformRollupConfig).toContain("distBase: 'tools/parse'")
     expect(platformRollupConfig).toContain("input: './src/tools/defi/index.ts'")
     expect(platformRollupConfig).toContain("distBase: 'tools/defi'")
+    expect(platformRollupConfig).toContain("input: './src/tools/prep/index.ts'")
+    expect(platformRollupConfig).toContain("distBase: 'tools/prep'")
 
     expect(typesRollupConfig).toContain(
       "createSubpathTypesConfig('src/tools/parse/index.ts', 'dist/tools/parse/index.d.ts')"
     )
     expect(typesRollupConfig).toContain(
       "createSubpathTypesConfig('src/tools/defi/index.ts', 'dist/tools/defi/index.d.ts')"
+    )
+    expect(typesRollupConfig).toContain(
+      "createSubpathTypesConfig('src/tools/prep/index.ts', 'dist/tools/prep/index.d.ts')"
     )
   })
 })
--- a/scripts/smoke-sdk-subpath-exports.mjs
+++ b/scripts/smoke-sdk-subpath-exports.mjs
@@ -52,15 +52,20 @@ try {
       "const require = createRequire(import.meta.url)",
       "const parsePath = require.resolve('@vultisig/sdk/tools/parse')",
       "const defiPath = require.resolve('@vultisig/sdk/tools/defi')",
+      "const prepPath = require.resolve('@vultisig/sdk/tools/prep')",
       "assert.match(parsePath, /dist\\/tools\\/parse\\/index\\.cjs$/)",
       "assert.match(defiPath, /dist\\/tools\\/defi\\/index\\.cjs$/)",
+      "assert.match(prepPath, /dist\\/tools\\/prep\\/index\\.cjs$/)",
       "const parse = await import('@vultisig/sdk/tools/parse')",
       "const defiModule = await import('@vultisig/sdk/tools/defi')",
+      "const prep = await import('@vultisig/sdk/tools/prep')",
       "assert.equal(parse.parseChain('Ethereum').success, true)",
       "assert.equal(typeof parse.parseTicker, 'function')",
       "assert.equal(typeof defiModule.defi, 'object')",
       "assert.equal(typeof defiModule.osmosis.buildSwapExactAmountIn, 'function')",
-      "console.log(JSON.stringify({ parsePath, defiPath, parseOk: true, defiOk: true }))",
+      "assert.equal(typeof prep.prepareSendTxFromKeys, 'function')",
+      "assert.equal(typeof prep.prepareIbcTransfer, 'function')",
+      "console.log(JSON.stringify({ parsePath, defiPath, prepPath, parseOk: true, defiOk: true, prepOk: true }))",
       '',
     ].join('\n')
   )
@@ -69,11 +74,16 @@ try {
     [
       "import { parseChain, type ParseChainResult } from '@vultisig/sdk/tools/parse'",
       "import { defi, type Defi } from '@vultisig/sdk/tools/defi'",
+      "import { prepareIbcTransfer, prepareSendTxFromKeys, type PrepareSendTxFromKeysParams } from '@vultisig/sdk/tools/prep'",
       '',
       "const chainResult: ParseChainResult = parseChain('Ethereum')",
       'void chainResult',
       'const tools: Defi = defi',
       'void tools',
+      'void prepareIbcTransfer',
+      'void prepareSendTxFromKeys',
+      'type _PrepParams = PrepareSendTxFromKeysParams',
+      'void (0 as unknown as _PrepParams)',
       '',
     ].join('\n')
   )
--- a/.changeset/sdk-tools-prep-subpath.md
+++ b/.changeset/sdk-tools-prep-subpath.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Publish the documented `@vultisig/sdk/tools/prep` subpath with dedicated JS and type bundles.
--- a/packages/sdk/rollup.types.config.js
+++ b/packages/sdk/rollup.types.config.js
@@ -102,4 +102,5 @@ export default defineConfig([
   // declarations instead of the root index type graph.
   createSubpathTypesConfig('src/tools/parse/index.ts', 'dist/tools/parse/index.d.ts'),
   createSubpathTypesConfig('src/tools/defi/index.ts', 'dist/tools/defi/index.d.ts'),
+  createSubpathTypesConfig('src/tools/prep/index.ts', 'dist/tools/prep/index.d.ts'),
 ])


packages/sdk/package.json
```diff
@@ -98,6 +98,19 @@
       "import": "./dist/tools/defi/index.js",
       "require": "./dist/tools/defi/index.cjs",
       "default": "./dist/tools/defi/index.cjs"
+    },
+    "./tools/prep": {
+      "types": "./dist/tools/prep/index.d.ts",
+      "browser": "./dist/tools/prep/index.js",
+      "worker": "./dist/tools/prep/index.js",
+      "react-native": "./dist/tools/prep/index.js",
```
NeOMakinG, line 106: `blocker:` these three conditions (`browser`:104, `worker`:105, `react-native`:106) point at `dist/tools/prep/index.js`, which `createToolsSubpathConfigs` (`rollup.platforms.config.js:239`) emits **only** inside `configs.node`, with `preferBuiltins: true` and `VULTISIG_PLATFORM='node'`. Its first line statically imports `node:crypto`, `crypto`, `node:util`, `events`, `https`, `http`, `net`, `tls`, `stream`, `url`, `zlib`, `fs`, `path`, `os` and more.

Reproduced with a real bundler through the RN condition:

```
$ esbuild app.mjs --bundle --format=esm --platform=browser --conditions=react-native
  # import { prepareIbcTransfer } from '@vultisig/sdk/tools/prep'
✘ [ERROR] Could not resolve "node:crypto"
✘ [ERROR] Could not resolve "crypto"
✘ [ERROR] Could not resolve "node:util"
✘ [ERROR] Could not resolve "https"
```

Control, same command: `tools/defi` → 0 errors, `tools/parse` → 0 errors. Their bundles import only `viem` / `zod`, which is why the template you copied is fine for them and not for prep.

Also note `src/platforms/react-native/index.ts` deliberately splits this exact barrel into an RN-safe static allow-list plus `await import()` lazy wrappers, with comments naming the hazard. This entry hands Metro a static route around all of that — and it collides with #1389, which is widening the *curated* RN surface for the same helpers.

Suggested fix (fail-closed, one-line-per-condition): delete lines 104-106 so non-node bundlers get `ERR_PACKAGE_PATH_NOT_EXPORTED` at build time instead of an unresolvable graph. `parse`/`defi` keep theirs since theirs actually work. The alternative is real per-platform prep bundles wired into the `browser` and `react-native` target arrays.

packages/sdk/package.json
```diff
@@ -98,6 +98,19 @@
       "import": "./dist/tools/defi/index.js",
       "require": "./dist/tools/defi/index.cjs",
       "default": "./dist/tools/defi/index.cjs"
+    },
+    "./tools/prep": {
```
NeOMakinG, line 102: `should-fix:` I diffed all 31 runtime exports of this barrel against `src/index.ts`: 30/31 are already root-exported, so for node consumers this subpath is genuinely just a narrower import path — nice, low-risk. The one exception is `prepareThorchainMsgDepositTxFromKeys` (+ `PrepareThorchainMsgDepositTxFromKeysParams`), which is **not** in `src/index.ts`, **not** in the RN entry, and has exactly one caller: `src/vault/services/TransactionBuilder.ts:118`. Publishing this subpath makes it a semver-supported public API.

The changeset says "the documented `@vultisig/sdk/tools/prep` subpath" and `CLAUDE.md:79` documents six prep helpers — this isn't one of them. Either call it out in the changeset as a new public export, or hold it back.

packages/sdk/tests/unit/public-api/toolsSubpathExports.test.ts
```diff
@@ -26,22 +27,34 @@ describe('public API tools subpath exports', () => {
       require: './dist/tools/defi/index.cjs',
       default: './dist/tools/defi/index.cjs',
     })
+    expect(prepExport).toMatchObject({
+      types: './dist/tools/prep/index.d.ts',
+      import: './dist/tools/prep/index.js',
+      require: './dist/tools/prep/index.cjs',
+      default: './dist/tools/prep/index.cjs',
```
NeOMakinG, line 34: `should-fix:` this `toMatchObject` asserts `types`/`import`/`require`/`default` and omits `browser`/`worker`/`react-native` — which are exactly the three keys that are wrong (see the blocker on `package.json:106`). Combined with the assertions below that just grep the config-file text this PR itself added, nothing in this file can go red for a condition mistake that preserves those strings.

Suggested fix: assert the full condition set including the three platform keys, and make the assertion about *what those artifacts contain* rather than what the config text says.

scripts/smoke-sdk-subpath-exports.mjs
```diff
@@ -52,15 +52,20 @@ try {
...
+      "assert.match(prepPath, /dist\\/tools\\/prep\\/index\\.cjs$/)",
       "const parse = await import('@vultisig/sdk/tools/parse')",
       "const defiModule = await import('@vultisig/sdk/tools/defi')",
+      "const prep = await import('@vultisig/sdk/tools/prep')",
       "assert.equal(parse.parseChain('Ethereum').success, true)",
       "assert.equal(typeof parse.parseTicker, 'function')",
       "assert.equal(typeof defiModule.defi, 'object')",
       "assert.equal(typeof defiModule.osmosis.buildSwapExactAmountIn, 'function')",
-      "console.log(JSON.stringify({ parsePath, defiPath, parseOk: true, defiOk: true }))",
+      "assert.equal(typeof prep.prepareSendTxFromKeys, 'function')",
```
NeOMakinG, line 66: `should-fix:` this is the only non-tautological proof in the PR, but (a) it exercises the node conditions only — `require.resolve` + dynamic `import()` — and (b) it isn't wired into CI. `grep -rn "smoke:subpaths\|smoke-sdk-subpath" .github/ package.json` returns exactly one hit, `package.json:85`; no workflow in `.github/workflows/` invokes it. So the blocker on `package.json:106` has no gate anywhere.

Suggested fix: add a `--conditions=react-native` and `--conditions=browser` resolve-and-bundle step per published tool subpath, and wire `yarn sdk:smoke:subpaths` into `test.yml`.

packages/sdk/rollup.platforms.config.js
```diff
@@ -326,6 +326,10 @@ const configs = {
       input: './src/tools/defi/index.ts',
       distBase: 'tools/defi',
     }),
+    ...createToolsSubpathConfigs({
+      input: './src/tools/prep/index.ts',
```
NeOMakinG, line 330: `q:` this entry emits a 2.85 MB ESM bundle, against 626 KB for `tools/defi` and 6.9 KB for `tools/parse`, with `viem` / `@solana/web3.js` / `@cosmjs/*` / `@trustwallet/wallet-core` already external — so it's ~2.8 MB of duplicated SDK chain code for anyone importing both the root entry and this subpath.

I looked for a correctness angle and didn't find one: WalletCore is passed in as a parameter rather than held in a module singleton, and the only module-scope cache in the graph is `src/chains/ton/walletV4R2.ts:44`, a pure derived constant. So this reads as size-only. Intentional, or worth a shared chunk?

NeOMakinG (commented):
Round 1. Draft, so **COMMENTED** — not approving a draft, and I did not push anything to your branch.

The publishing mechanism itself is sound and I proved it end to end. The problem is that the export-map template was copied from `./tools/parse` / `./tools/defi`, and those two barrels are pure (`zod` only / `viem` only) while `tools/prep` is not. The `browser` / `worker` / `react-native` conditions on the new entry point at a **node-target** bundle, and I reproduced a hard bundler failure through them.

## testing approach

Worktree at `1918e26e`, `node_modules` symlinked from the canonical checkout (never `yarn install` in a worktree).

**Artifacts actually build.** I filtered `rollup.platforms.config.js` down to the `tools/prep` entries and ran them under `BUILD_TARGET=node`, and copied `dtsPluginOptions` into a throwaway types config with only the prep entry (the full `rollup.types.config.js` OOMs on the `src/index.ts` dts entry on this box):

```
created ./dist/tools/prep/index.js   in 12s     (2,851,640 bytes)
created ./dist/tools/prep/index.cjs  in 11.1s   (2,860,306 bytes)
created ./dist/tools/prep/index.d.ts in 2.5s    (105,548 bytes, 3073 lines)
```

**Subpath resolves as a real external consumer.** Symlinked `node_modules/@vultisig/sdk` -> the built package from a separate directory outside the repo, then resolved from outside with positive and negative controls:

```
req:@vultisig/sdk/tools/prep  -> <pkg>/dist/tools/prep/index.cjs
imp:@vultisig/sdk/tools/prep  -> <pkg>/dist/tools/prep/index.js
req:@vultisig/sdk/node        -> <pkg>/dist/index.node.cjs          (positive control)
imp:@vultisig/sdk/node        -> <pkg>/dist/index.node.esm.js       (positive control)
req:@vultisig/sdk/nope        -> THROW ERR_PACKAGE_PATH_NOT_EXPORTED (negative control)
imp:@vultisig/sdk/nope        -> THROW ERR_PACKAGE_PATH_NOT_EXPORTED (negative control)
```

Then actually loaded it both ways: **31 runtime symbols under ESM, 31 under CJS, zero diff either direction.** The `.d.ts` externalizes only `@bufbuild/protobuf` and `@trustwallet/wallet-core`, both declared `dependencies` of `@vultisig/sdk`, so a consumer's `tsc` resolves them.

**`yarn format:check` is clean.** Checked with the repo's own config against the gated glob — `prettier --config .config/.prettierrc --ignore-path .config/.prettierignore --check 'packages/sdk/**/*.{ts,tsx,js,jsx}'` -> "All matched files use Prettier code style", exit 0. (`scripts/**` is outside that glob entirely, so the smoke script isn't gated by it.) Three sibling export PRs today were red on exactly this gate; this one isn't.

**CI is real:** `gh api .../commits/1918e26e/check-runs` -> `total_count: 4`, all four `completed/success` (Unit Tests, Integration Tests, Lint and Type Check, Test Summary). Not a zero-jobs green.

Two lanes: me plus an independent Codex adversarial lane. Codex reached the RN blocker independently.

## blocking

**1. The `browser` / `worker` / `react-native` conditions ship a node-only bundle. Reproduced.**

`createToolsSubpathConfigs` (`packages/sdk/rollup.platforms.config.js:239`) emits exactly one ESM artifact, built with `preferBuiltins: true` and `process.env.VULTISIG_PLATFORM = 'node'`, and lives only inside `configs.node`. The new export entry points `browser`, `worker` and `react-native` at that file.

First line of the emitted `dist/tools/prep/index.js`:

```js
import*as t from"node:crypto";import r from"node:crypto";import*as i from"tiny-secp256k1";import o from"crypto";
import s,{Buffer as f}from"buffer";...import oe,{types as se,deprecate as ce,promisify as ue}from"node:util";
import le from"events";import de from"https";import fe from"http";import pe from"net";import he from"tls";
import me from"stream";import be from"url";import ge from"zlib";import ye from"fs";import ve from"path";import we from"os";...
```

Resolution under each condition (Node's own resolver, `--conditions=`):

```
react-native -> <pkg>/dist/tools/prep/index.js
browser      -> <pkg>/dist/tools/prep/index.js
worker       -> <pkg>/dist/tools/prep/index.js
```

Bundling a one-line consumer through that condition with a real bundler:

```
$ esbuild app.mjs --bundle --format=esm --platform=browser --conditions=react-native
  # app.mjs: import { prepareIbcTransfer } from '@vultisig/sdk/tools/prep'

✘ [ERROR] Could not resolve "node:crypto"
✘ [ERROR] Could not resolve "crypto"
✘ [ERROR] Could not resolve "node:util"
✘ [ERROR] Could not resolve "https"
```

**Control, identical command against the existing siblings: `tools/defi` -> 0 errors, `tools/parse` -> 0 errors.** Their bundles import only `viem` (626 KB) and `zod` (6.9 KB) respectively. So the template is correct for its two current users and this PR is the first case where it isn't — this is new breakage, not inherited.

The RN entry already encodes the reason, in its own comments. `src/platforms/react-native/index.ts` splits `tools/prep` into a hand-curated static allow-list ("depend only on bech32 + buffer (RN-safe, no mpc/keysign)", "only depend on JSON/bech32 — no WalletCore or native crypto, safe as static re-exports") and lazy `await import('../../tools/prep/maxSend')` / `.../contractCall` / `.../send` wrappers for everything else. Compare the two RN graphs: the shipped `dist/index.react-native.js` imports `expo-crypto`, `@vultisig/walletcore-native`, `@vultisig/mpc-native`, `@react-native-async-storage/async-storage` — the RN substitutes. The bundle this PR points Metro at imports `@trustwallet/wallet-core` (the WASM build) and 20+ Node builtins instead.

So `import { prepareSendTxFromKeys } from '@vultisig/sdk/tools/prep'` in the RN app is a second, uncurated RN route to the same module that bypasses every lazy wrapper — and it fails at Metro resolve time, not with a nice error. **This is the one finding that also matters for #1389**: #1389 is widening the curated RN surface for these helpers, this PR opens an uncurated one next to it. They should not both land as-is.

Two ways out:

- Cheapest and fail-closed: drop `browser`, `worker` and `react-native` from the `./tools/prep` entry. Non-node bundlers then get `ERR_PACKAGE_PATH_NOT_EXPORTED` at build time — an honest "use the RN entry" — instead of an unresolvable graph. `parse`/`defi` keep theirs, since theirs work.
- Or make it true: add prep to the `browser` and `react-native` target arrays in `rollup.platforms.config.js` with the same `preferBuiltins: false` / `browser: true` / platform-replace treatment those targets already use, emit `dist/tools/prep/index.browser.js` and `index.react-native.js`, and point each condition at its own artifact. That's the honest version but it means the whole prep graph has to be RN-buildable, which is exactly what the lazy wrappers exist to avoid.

## should-fix

**2. One symbol becomes public for the first time, and it isn't the "documented" surface.**

I diffed all 31 runtime exports of the barrel against `src/index.ts`. 30 of 31 are already root-exported, so for node/browser consumers this subpath is genuinely just a narrower import path — good, and worth saying out loud. The exception:

`prepareThorchainMsgDepositTxFromKeys` (and its type `PrepareThorchainMsgDepositTxFromKeysParams`) is **not** in `src/index.ts`, **not** in the RN entry, and has exactly one caller in the repo: `src/vault/services/TransactionBuilder.ts:118`. This PR turns it into a semver-supported public API. The changeset says "publish the documented `@vultisig/sdk/tools/prep` subpath" — `CLAUDE.md:79` documents six prep helpers and this isn't one of them. Either mention it in the changeset as a new public export or hold it back.

**3. Nothing in CI can catch finding 1.**

The new unit test asserts the text of the three config files the PR itself edited (`toContain("input: './src/tools/prep/index.ts'")` etc.) — it will stay green for any mistake that preserves those strings, including this one. And the `toMatchObject` at `toolsSubpathExports.test.ts:30-35` checks `types`/`import`/`require`/`default` and deliberately omits `browser`/`worker`/`react-native`, which are the three keys that are wrong.

The only non-tautological proof is `scripts/smoke-sdk-subpath-exports.mjs`, and it is **not wired into any workflow** — `grep -rn "smoke:subpaths\|smoke-sdk-subpath" .github/ package.json` returns exactly one hit, `package.json:85`. It's a manual `yarn sdk:smoke:subpaths`. It also only exercises `require.resolve` + dynamic `import()`, i.e. the node conditions only.

Concrete fix that would have caught this: in the smoke script, after the node assertions, add a `--conditions=react-native` / `--conditions=browser` resolve + bundle step for each published tool subpath, and wire the script into `test.yml`.

**4. Eight published functions have params or returns a consumer can't name from this subpath.**

`ContractCallTxParams`, `SignAminoInput`, `CosmosSigningOptions`, `MaxSendAmount`, `SwapQuote`, `TonTxBuilderResult`, `CosmosMsgInput` and `KeysignPayload` are all in the signatures of exported functions but are not re-exported from `src/tools/prep/index.ts`. So `prepareContractCallTxFromKeys`, `prepareSignAminoTxFromKeys`, `prepareSignDirectTxFromKeys`, `getMaxSendAmountFromKeys`, `prepareSwapTxFromKeys`, `prepareThorchainMsgDepositTxFromKeys`, `prepareJettonTransferTxFromKeys` and `buildCosmosWasmExecuteMsg` can be called but not typed without reaching back to the root entry or into internals. The smoke type-check only exercises `PrepareSendTxFromKeysParams`, which is one of the ones that *is* exported, so it can't see this.

This is a property of `tools/prep` itself rather than of the publishing mechanism, so it applies to #1389 too.

## q

**5.** `dist/tools/prep/index.js` is 2.85 MB, against 626 KB for `tools/defi` and 6.9 KB for `tools/parse`, with `viem` / `@solana/web3.js` / `@cosmjs/*` / `@trustwallet/wallet-core` already external — so that's ~2.8 MB of duplicated SDK chain code for anyone who imports both `@vultisig/sdk` and `@vultisig/sdk/tools/prep`. I went looking for a correctness angle and didn't find one: WalletCore is passed in as a parameter everywhere rather than held in a module singleton, and the only module-scope cache in the graph is `src/chains/ton/walletV4R2.ts:44` (`cachedV4R2CodeCell`), which is a pure derived constant. So this looks like size only. Intentional, or worth a shared chunk?

## risk

Low if finding 1 is fixed by dropping the three conditions — that's a one-line-per-condition deletion in `package.json` and everything else in the PR stands. Medium if it ships as-is: the failure lands on RN and browser consumers at *their* build time, on a subpath the changeset advertises, and there is no CI gate that would surface it before a release. Nothing here touches signing, key material, or amount math.

## verdict

**COMMENTED** — draft, and finding 1 is a real break I reproduced with a bundler plus a clean control. The mechanism (export map, both rollup configs, tarball-shaped resolution, ESM/CJS parity, d.ts) is correct and I verified all of it. Fix the three conditions and the rest is good plumbing.

---
*Reviewed by an automated reviewer (Claude, 2 lanes: Claude correctness + Codex adversarial). Findings verified by build + external-consumer resolution probes against `1918e26e`.*

NeOMakinG (commented):
Follow-up to my review above. A sibling review on #1389 found that the `tools/prep` barrel eagerly evaluates `@mysten/sui` — `src/tools/prep/suiTokenTransfer.ts:31` calls `normalizeStructTag(...)` at module scope, so a value import from the barrel hoists Sui into the importer's graph, and a consumer who metro-stubs Sui loses `prepareIbcTransfer` too. I checked whether that transfers to the subpath route.

**It does not. Refuting it plainly, with the probe output.**

`@mysten/sui` is externalized only in the **react-native** target's own external list (`rollup.platforms.config.js:432`, under the "Heavy chain clients — deliberately kept out of the RN bundle" comment). `createToolsSubpathConfigs` (`:239`) uses the shared `external` array at `:14`, which does not list it. So Sui is **inlined** into `dist/tools/prep/index.js` at SDK build time.

Import prologue of the built artifact — 33 unique externals, and Sui is not among them:

```
@bufbuild/protobuf  @cosmjs/stargate  @solana/web3.js  @trustwallet/wallet-core  @vultisig/mpc-types
buffer crypto events fs http https net os path stream tls url util zlib viem
node:{buffer,crypto,fs,http,https,net,path,stream,url,util,zlib}  ripple-binary-codec  tiny-secp256k1
```

No `from "@mysten/sui/utils"` statement anywhere in the prologue. The only `@mysten` string left in the bundle is an inlined BCS symbol (`Symbol.for("@mysten/serialized-bcs")`). Confirmed by bundling a real consumer and reading the metafile:

```
import { prepareIbcTransfer } from '@vultisig/sdk/tools/prep'  -> 1103 inputs, mysten inputs: 0
import * as p from '@vultisig/sdk/tools/prep'                  -> 1103 inputs, mysten inputs: 0
```

So there is nothing left for a consumer to stub, and no undeclared runtime dep: every external the subpath carries is either a Node builtin or an existing declared `dependency` of `@vultisig/sdk`. The #1389 coupling is real on the RN-entry route and absent on this one — different externalization, different outcome. Negative control: `git show origin/main:packages/sdk/rollup.platforms.config.js | grep -c "tools/prep"` -> `0`, i.e. no prep artifact exists on base at all, so this is a clean addition.

## should-fix (upgrading my earlier `q:` on `rollup.platforms.config.js:330`)

The same probe answered the tree-shaking question, and the answer is worse than the raw size suggested.

```
consumer imports ONLY prepareIbcTransfer  -> 14,914,248 bytes, 1103 inputs
consumer imports ALL 31 symbols           -> 15,250,003 bytes, 1103 inputs
```

**Identical input count, and 98% of the bytes, for one pure helper.** `createToolsSubpathConfigs` uses `inlineDynamicImports: true` with a single-file output, so the barrel's module boundaries are gone by the time a consumer sees it — there is nothing to shake. Someone who wants `prepareIbcTransfer` (a `@scure/base` string function) pays for `@solana/web3.js`, WalletCore, the whole inlined Sui/BCS graph, and transitive `worker_threads` / `tls` / `net`.

That reframes my earlier `q:` — this isn't "size only, probably intentional". Combined with the blocker (the same pure IBC helper is unreachable through the `browser`/`worker`/`react-native` conditions because the artifact is node-built), a consumer who wants the pure helpers can get them neither cheaply nor portably.

So I'd second the fix shape #1389 landed on, and it composes with my blocker rather than competing with it: publish **deep subpaths for the pure helpers** — `./tools/prep/ibcTransfer`, `./tools/prep/cosmosStaking`, `./tools/prep/cosmosWasmExecute`, `./tools/prep/cw20Transfer`, `./tools/prep/polkadotAssetSend` — which the RN entry has *already* certified as RN-safe in its own comments and static-re-exports today, and which can therefore legitimately carry `browser`/`worker`/`react-native` conditions. Keep the barrel subpath node-only. That gets the portable helpers portable and small, and stops the impure ones from being coupled to them.

One clarification so this doesn't contradict the #1389 round: the eager `normalizeStructTag` call at `suiTokenTransfer.ts:31` still executes at import time here — it just executes *inlined* code rather than reaching for a stubable module. It's a pure string normalization over a constant, no network or wasm, so it isn't a Hermes-eval hazard on this route. The Sui coupling on this PR is a **cost** problem, not a **crash** problem; on #1389's route it's the reverse.

Everything in my first review stands unchanged.

---
*Reviewed by an automated reviewer (Claude). Probes run against a fresh build of `1918e26e`; metafile figures from `esbuild --bundle --platform=node --metafile` on a symlinked external consumer.*

PR #1407, opened 2026-07-18
fix(sdk): export RN pure validation/normalization/policy helpers

gomesalexandre:
## Summary

Exports the newer pure helper family from  so RN consumers can use the same canonical SDK surfaces as node/browser consumers instead of deep imports or local copies.

Closes 1406

## What changed

- re-exported RN-safe pure helpers from 
  - validation normalizers (, , , etc.)
  - address / chain-prefix helpers (, , , etc.)
  - tx normalization (, )
  - known-contract helpers (, , canonical contract sets)
  - policy namespace + related runtime constants / types (, , , etc.)
- extended  to assert symbol identity and a few representative runtime calls
- added 

## Verification

- ➤ YN0000: · Yarn 4.16.0
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Post-resolution validation
➤ YN0060: │ @vitest/coverage-v8 is listed by your project with version 4.1.9 (p314426), which doesn't satisfy what vitest requests (4.1.8).
➤ YN0060: │ @vitest/coverage-v8 is listed by your project with version 4.1.9 (p5ea16a), which doesn't satisfy what vitest requests (4.1.8).
➤ YN0060: │ @vitest/coverage-v8 is listed by your project with version 4.1.9 (p8b5072), which doesn't satisfy what vitest requests (4.1.8).
➤ YN0060: │ eslint is listed by your project with version 10.6.0 (p8b7932), which doesn't satisfy what eslint-plugin-jsx-a11y and other dependencies request (but they have non-overlapping ranges!).
➤ YN0060: │ react is listed by your project with version 19.2.7 (p444e5f), which doesn't satisfy what @vultisig/examples-shared and other dependencies request (but they have non-overlapping ranges!).
➤ YN0060: │ react is listed by your project with version 19.2.7 (pfe4e79), which doesn't satisfy what @vultisig/examples-shared and other dependencies request (but they have non-overlapping ranges!).
➤ YN0060: │ react-dom is listed by your project with version 19.2.7 (p41aed6), which doesn't satisfy what @vultisig/examples-shared requests (^18.0.0).
➤ YN0060: │ react-dom is listed by your project with version 19.2.7 (p6ad760), which doesn't satisfy what @vultisig/examples-shared requests (^18.0.0).
➤ YN0060: │ vitest is listed by your project with version 4.1.8 (p423780), which doesn't satisfy what @vitest/coverage-v8 requests (4.1.9).
[...]

PR #1412, opened 2026-07-19
fix(sdk): export XRP destination helpers on public surfaces

### LINKED ISSUE #1411: architecture: XRP destination/X-address normalization helpers are missing from the public SDK surfaces
## Summary
`@vultisig/sdk` already has pure XRP destination/X-address normalization helpers in core, but the public SDK surfaces did not export them. First-party CLI code had to deep-import `@vultisig/core-chain/chains/ripple/address` instead of staying on the SDK boundary.

## Evidence
- `clients/cli/src/commands/transaction.ts:4-6` deep-imports `normalizeRippleDestination` from `@vultisig/core-chain/chains/ripple/address`.
- `packages/core/chain/chains/ripple/address.ts:13-52` already exposes the canonical pure helper family:
  - `decodeRippleXAddress`
  - `isValidRippleXAddress`
  - `normalizeRippleDestination`
  - `encodeRippleXAddress`
- `packages/sdk/src/index.ts` and `packages/sdk/src/platforms/react-native/index.ts` did not expose these helpers on the public package surfaces.

## Impact
This is a package-boundary drift vector: first-party consumers pierce internal package layout for a pure cross-platform helper that should be available from `@vultisig/sdk` / `@vultisig/sdk/react-native`.

## Fix sketch
- Export the XRP helper family from the root SDK and RN entrypoints.
- Add public-export regression tests.
- Replace first-party deep imports with public `@vultisig/sdk` imports.

## Effort
Low.

## Report
Round 43 report mirror: https://gist.github.com/gomesalexandre/bd996a164e22647d3d01a67071738e00

gomesalexandre:
## Summary
- export the XRP destination/X-address helper family from `@vultisig/sdk`
- export the same helper family from `@vultisig/sdk/react-native`
- route the CLI send command through the public SDK import instead of a deep core-chain import

Closes https://github.com/vultisig/vultisig-sdk/issues/1411

## Why
`normalizeRippleDestination` / `decodeRippleXAddress` / `encodeRippleXAddress` / `isValidRippleXAddress` already existed canonically in core, but first-party consumers still had to deep-import `@vultisig/core-chain/chains/ripple/address`. This patch makes the public SDK surfaces the canonical boundary.

## Verification
- `yarn install --immutable`
- `yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/utils/publicExports.test.ts tests/unit/platforms/react-native/entry.test.ts`
- `yarn build:sdk`
- `yarn cli:build`
- built-dist runtime proof:
  - `node -e "import('./packages/sdk/dist/index.node.esm.js').then(...)"`
  - confirmed `normalizeRippleDestination('XV5sbjUmgPpvXv4ixFWZ5ptAYZ6PD2q1qM6owqNbug8W6KV') -> { address: 'rPEPPER7kfTD9w2To4CQk6UCfuHM9c6GDY', destinationTag: 495 }`

## Report
Round 43 report mirror: https://gist.github.com/gomesalexandre/bd996a164e22647d3d01a67071738e00


packages/sdk/src/index.ts
```diff
@@ -103,6 +103,7 @@ export {
   isSolanaAddress,
   supportedChainTags,
 } from './utils/addressFormat'
+export { decodeRippleXAddress, encodeRippleXAddress, isValidRippleXAddress, normalizeRippleDestination } from '@vultisig/core-chain/chains/ripple/address'
```
NeOMakinG, line 109: `blocker:` this single line is the entire reason CI is red on `dda4b04e` — it trips **two** static gates, and `yarn format:check` is separate from `yarn lint`:

```
99:1  error  Run autofix to sort these exports!  simple-import-sort/exports
[warn] packages/sdk/src/index.ts   (prettier: 152 chars > printWidth 120)
```

I ran `eslint --fix` + `prettier --write` in a worktree and diffed; the exact resulting shape is to move the block **below** `export { checkChainPrefix } from './utils/chainPrefix'` and split it:

```ts
export {
  decodeRippleXAddress,
  encodeRippleXAddress,
  isValidRippleXAddress,
  normalizeRippleDestination,
} from '@vultisig/core-chain/chains/ripple/address'
```

Verified after applying: eslint clean and `prettier --check` reports "All matched files use Prettier code style!".

`preferably-blocking:` separately — with `isValidRippleXAddress` now on the root surface, note that `sdk.address.isValidFor('XV5sbjUmgPpv…', Chain.Ripple)` returns **false** and `sdk.classifyAddress('XV5sbjUmgPpv…')` returns **'unknown'**, while `sdk.isValidRippleXAddress('rPEPPER…')` returns **false**. Neither exported predicate accepts both valid XRP destination forms, so the general validator rejects exactly the tagged form this PR exists to enable. See finding 2 in the body.

packages/sdk/src/platforms/react-native/index.ts
```diff
@@ -77,6 +77,12 @@ export type { WalletCoreLike } from '@vultisig/walletcore-native'
 // RN wrappers accept WalletCoreLike from @vultisig/walletcore-native
 // so consumers don't need to cast to @trustwallet/wallet-core's WalletCore.
 export { deriveAddress, getCoinType, getPublicKey, isValidAddress } from './chainHelpers'
+export {
+  decodeRippleXAddress,
+  encodeRippleXAddress,
+  isValidRippleXAddress,
+  normalizeRippleDestination,
+} from '@vultisig/core-chain/chains/ripple/address'
```
NeOMakinG, line 115: `praise:` verified at the artifact, not by grep. Built `BUILD_TARGET=react-native` twice — at HEAD, and again with only this file reverted to `main` — and diffed `dist/index.react-native.js`:

```
ctrl exports 174  ->  head exports 178
ADDED:   decodeRippleXAddress, encodeRippleXAddress,
         isValidRippleXAddress, normalizeRippleDestination
REMOVED: (none)
ADDED top-level imports: (none)
+386 bytes
```

`should-fix:` while you're in this block, `normalizeRippleDestination` is a pass-through, not a validator: `''`, `'   '`, `'not-an-address'`, `'0xdEADbeEf…'`, a truncated classic address, and a **lowercased** X-address all return `{ address: <input verbatim> }` with no tag and no throw. The lowercase case is the sharp one — it misses the `startsWith('X')` test and silently degrades a tagged destination to a tagless one. A JSDoc line here (and ideally an uppercase-normalise or an explicit reject on `x…`) would keep a consumer from treating it as validation.

NeOMakinG (commented):
Reviewed as a **draft** — commenting only, no approval and no pushed fixes; the branch stays yours.

Two lanes: Claude (correctness + fund-safety, load-bearing) and Codex (adversarial second opinion). Findings deduped below; where the lanes disagreed I went with the artifact.

## testing approach

Worktree `/tmp/wt-sdk-1412` at `dda4b04e`, node_modules symlinked from the canonical checkout.

- **Semantics probed, not read.** Ran a throwaway vitest inside `packages/sdk` against `src/index` and drove every exported helper through the destination-tag edge cases: `0`, `0xffffffff`, `-1`, `1.5`, `2**32`, `NaN`, `Infinity`, the numeric string `'495'`, the bigint `495n`, `null`, `true`, `undefined`, plus classic / empty / whitespace / garbage / EVM-`0x` / TRON-`T` / lowercased-X / truncated / trailing-junk destinations.
- **RN reachability by two-build artifact diff** (`#1389` technique): built `BUILD_TARGET=react-native` at HEAD, then again with **only** `src/platforms/react-native/index.ts` reverted to `main`, and diffed the emitted `export {}` names and the top-level bare-import set of `dist/index.react-native.js`.
- **Real CI truth** via `gh api .../commits/dda4b04e/check-runs` (4 runs, not zero), plus a local reproduction of the failing gate and a derived + verified fix.
- Ran the PR's own two test files: `publicExports.test.ts` + `platforms/react-native/entry.test.ts` — **24/24 green**.

**RN artifact diff (the reachability proof):**

```
ctrl exports 174   head exports 178
ADDED exports:   [decodeRippleXAddress, encodeRippleXAddress,
                  isValidRippleXAddress, normalizeRippleDestination]
REMOVED exports: []
ADDED top-level imports: []          <- none
ctrl already imports 'ripple-address-codec': true
size ctrl 2450915 -> head 2451301   (+386 bytes)
```

All four symbols reach the built RN bundle. Positive control = the 4 names appear; negative control = they are absent with the one file reverted. Same for the root surface: `src/platforms/{node,browser,chrome-extension}/index.ts` all `export * from '../../index'`, so the root barrel line covers every non-RN condition.

**Destination-tag semantics — probed output (abridged):**

```
normalize(encode(CLASSIC, 0))       -> { address: 'rPEPPER…', destinationTag: 0 }   hasOwnProperty: true
decode(encode(CLASSIC, 4294967295)) -> { destinationTag: 4294967295 }
encode(tag = -1 | 1.5 | 2**32 | NaN | Infinity | '495' | 495n | null | true)
                                    -> throws "…must be a nonnegative UInt32 integer"  (9/9)
```

Tag `0` survives end to end and is an **own property**, not dropped by a truthiness check — `assertSupportedTag` tests `tag === false`, not `!tag` (`packages/core/chain/chains/ripple/address.ts:9`). The signing path agrees: `validateDestinationTag` (`packages/core/mpc/keysign/utils/rippleDestinationTag.ts:15`) rejects the same 9 shapes and `getRippleChainSpecific` gates on `!== undefined`, not truthiness. **This is not the narrow-type-reader bug class** — the one I expected here isn't present. And the helper being exported is the same one the signer calls (`packages/core/mpc/keysign/send/build.ts:60`), not a display-side sibling.

## blocking

**1. CI is red on this HEAD, and this PR's one export line is the sole cause.** `gh api` reports `Lint and Type Check: failure` and `Test Summary: failure` (Unit + Integration are green — "Static quality gate failed!"). Reproduced locally; two separate gates trip:

```
$ node scripts/run-eslint.mjs --config .config/eslint.config.mjs packages/sdk/src/index.ts
  99:1  error  Run autofix to sort these exports!  simple-import-sort/exports

$ prettier --config .config/.prettierrc --ignore-path .config/.prettierignore --check packages/sdk/src/index.ts
  [warn] packages/sdk/src/index.ts     (main is clean — control run)
```

`yarn format:check` is a gate separate from `yarn lint`; here **both** fail. The 152-char single line exceeds `printWidth: 120`, and the bare `@vultisig/…` specifier sorts into the wrong group. Derived the exact autofix and verified both gates go clean — see the inline comment.

## preferably-blocking

**2. The public surface has no predicate that accepts both valid XRP destination forms — and the one that rejects X-addresses is the general one.** Probed against `src/index`:

```
sdk.address.isValidFor('XV5sbjUmgPpv…', Chain.Ripple)           -> false
sdk.classifyAddress('XV5sbjUmgPpv…')                            -> 'unknown'
sdk.isValidRippleXAddress('rPEPPER7kfTD9w2To4CQk6UCfuHM9c6GDY') -> false

sdk.address.isValidFor('rPEPPER…', Chain.Ripple)                -> true
sdk.isValidRippleXAddress('XV5sbjUmgPpv…')                      -> true
```

Each exported validator is a false-negative on the other's form. A first-party consumer wiring the obvious "validate the destination, then normalize it" against this boundary has to hand-roll `isValidRippleXAddress(a) || address.isValidFor(a, Chain.Ripple)`; a consumer who reasonably picks the *general* one rejects precisely the tagged X-address form this PR exists to make first-class. `packages/core/chain/utils/isValidAddress.ts:32` already gets this right with its `/^[XT]/u` branch, but that function needs a WalletCore instance and is not on the root surface at all (RN gets it via `chainHelpers`; root does not).

Suggest one of: teach the ripple rule in `src/utils/addressFormat` about the `X…` form so `address.isValidFor` / `classifyAddress` stop false-negativing, **or** export an `isValidRippleDestination = classic ∪ X-address` alongside the four helpers. Either way the round-trip is closed by the same PR that opens it.

## should-fix

**3. `normalizeRippleDestination` does zero validation on the non-`X`/`T` path, and is now public under a name that reads like a validator.** Probed — every one of these returns and does not throw:

```
normalize('')                     -> { address: '' }
normalize('   ')                  -> { address: '' }
normalize('not-an-address')       -> { address: 'not-an-address' }
normalize('0xdEADbeEf…')          -> { address: '0xdEADbeEf…' }
normalize('rPEPPER7kfTD9w2To4CQ') -> { address: 'rPEPPER7kfTD9w2To4CQ' }   (truncated)
normalize('xv5sbjumgppvxv4ix…')   -> { address: 'xv5sbjumgppvxv4ix…' }     (lowercased X-addr)
```

The last one is the sharp edge: an X-address that arrives lowercased (a case-normalising form field, a copy source that lowercases) misses the `startsWith('X')` test and degrades **silently** to a tagless classic-address pass-through — the tag is gone with no error. Inside the SDK this fails closed downstream (`isValidAddress` rejects it, and XRPL would reject it), so this is not a live fund-loss path today; but as a published helper it is exactly the shape a consumer builds a send around. Please add a JSDoc line on the export stating it is a *pass-through, not a validator*, and either uppercase-normalise or explicitly reject a lowercase `x…` prefix.

**4. The PR publishes the parser but neither its result type nor a tag validator** (`#1407`'s input/output-pairing lesson; Codex flagged the same shape independently). `RippleDestination` (`packages/core/chain/chains/ripple/address.ts:3`) is not exported from either surface, and neither is `validateDestinationTag` / `maxRippleDestinationTag`. The public send contract is split (`to: string` + `destinationTag?: number` — `VaultBase.ts:1737`, `tools/prep/send.ts:20`), so

```ts
const d = normalizeRippleDestination(xAddress)
vault.send({ chain: Chain.Ripple, to: d.address, amount })   // tag silently gone
```

type-checks cleanly and drops the tag. **Calibration: this is DX, not fund-loss** — passing the raw X-address straight through to `vault.send` / `buildXrpSendTx` re-normalises it internally and preserves the tag, and a user-supplied tag is validated fail-closed at build time. But exporting the type (and ideally a public tag validator) is what makes the split contract self-documenting.

## q

**5.** `packages/core/mpc/keysign/chainSpecific/resolvers/ripple.ts:74` gates the `lsfRequireDestTag` preflight on `!coin.id`, so an issued-currency (token) XRP payment skips the "destination requires a DestinationTag" check that a native XRP payment gets. XRPL enforces the flag for both, so this surfaces as an on-chain `tecDST_TAG_NEEDED` rather than a loss — but is the asymmetry deliberate? Pre-existing and out of this diff; noting it because this is the destination-tag PR.

**6.** `clients/cli/package.json` still declares `ripple-address-codec: ^5.0.1`, but I find no `ripple-address-codec` import left anywhere under `clients/cli/src`. Worth dropping in the same patch that removes the deep core-chain import, unless `knip` is configured to ignore it.

## risk

**Low.** The change is a re-export plus one CLI import swap; the semantics of the four helpers are unchanged and I could not break the tag round-trip. The blocking item is a static gate, not behaviour. The one thing I'd want closed before this lands is finding 2 — the boundary this PR declares canonical currently rejects the tagged form on its general validator.

Codex's lane raised an RN module-init concern (`ripple-address-codec` → `@xrplf/isomorphic/{utils,sha256}` → `require("crypto")` + `Buffer`, an app-boot failure on Hermes if unpolyfilled). **Refuted for this PR by the artifact:** the control build proves `ripple-address-codec` was already a top-level static import of `dist/index.react-native.js` on `main` (it is in the RN `external` list at `rollup.platforms.config.js:461`, and with `inlineDynamicImports: true` every external is hoisted to a top-level import). The PR adds **zero** new module-graph imports. Metro/Hermes resolution of `@xrplf/isomorphic` is a standing consumer-side requirement, not something this PR introduces.

## verdict

**COMMENTED** — draft, so no approval by policy, and I have not pushed anything to your branch. One blocking item (CI, mechanical, exact fix inline), one preferably-blocking (the validator round-trip on the boundary this PR declares canonical), two should-fix. The destination-tag semantics themselves came out clean under probing — tag `0`, `0xffffffff`, and all nine malformed shapes behave correctly on both the helper and the signer.

<sub>🤖 reviewed by an agent — the probe commands and both RN build artifacts are reproducible from the steps above.</sub>

NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because this conflicted old head carries a large unrelated delta for a small export change, while the public Ripple destination helper still needs validation-focused coverage; please recreate the focused export from current main.

PR #1413, opened 2026-07-19
fix(sui): detect finalized broadcast failures

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Sui transactions that fail during on-chain execution are no longer reported as successfully broadcast.
  * Broadcast responses without confirmed successful execution effects now fail with a clear error.
  * Sui transaction broadcasts now verify execution results before returning success.

* **Tests**
  * Added coverage for failed, missing, and successful Sui execution effects.

### DIFF
--- a/.changeset/safe-lions-approve.md
+++ b/.changeset/safe-lions-approve.md
@@ -1,0 +1,8 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+For EVM swap providers that legitimately distinguish the allowance executor
+from the swap router, use the quote's approval address for SDK allowance checks
+and approval payloads while retaining the router as the signed destination.
--- a/.changeset/sui-finalized-broadcast-failure.md
+++ b/.changeset/sui-finalized-broadcast-failure.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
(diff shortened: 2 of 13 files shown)


gomesalexandre (approved):
## lgtm 🫡

full adversarial review (claude lane; codex out til 07-25) — traced the fix + ran the suites (6/6 core, 21/21 sdk). this closes a real, previously-LIVE false-success: `RawBroadcastService.broadcastSuiRawTx` called `executeTransactionBlock` WITHOUT `showEffects: true` and returned `result.digest` on RPC-acceptance alone, blind to on-chain MoveAbort/InsufficientGas — so every Sui raw broadcast via `vault.broadcastRawTx()` reported success regardless of the execution outcome. same class already fixed for the primary resolver (sdk#1398) + cosmos raw (#1316); this was the one Sui spot that never got it.

fix is two-part, both necessary: `showEffects: true` so effects actually come back, + the extracted `assertSuiTxSucceeded(effects)` which throws `DeliverTxFailedError` on any non-`'success'` status (incl the missing/empty-effects shapes that were the actual prod bug). grepped every `executeTransactionBlock` call site — exactly 2, both now covered, no third silent spot. marker-type preserved where it feeds `withTransientBroadcastRetry` (instanceof before the message-regex). tests pin the MoveAbort failure direction + the no-effects edge cases.

changeset now in (@vultisig/core-chain + @vultisig/sdk patch). approving.

PR #1414, opened 2026-07-19
fix(swap): honor quote approval spender

### LINKED ISSUE #1198: fund-safety: the SDK's own approve leg ignores approvalAddress - vault.swap()/CLI reproduce the 0xa3aadf17 revert class
Surfaced during the #895 review (approvalAddress on EVM swap routes). #895 exposes the aggregator's real spender on the quote, and agent-backend-ts's execute_swap already consumes it (effectiveSpender = approvalAddress ?? ...). But the SDK's OWN approve leg still ignores it: core/mpc/keysign/swap/build.ts:369 sets `spender = keysignPayload.toAddress` unconditionally - so vault.swap() and the CLI approve the ROUTER ENTRYPOINT rather than the actual spender contract, reproducing the exact class of the 0xa3aadf17 on-chain revert (approve granted to the wrong address, swap pulls from the real spender, transferFrom fails).

Fix direction: `spender = approvalAddress ?? keysignPayload.toAddress` in the approve-leg builder, keeping toAddress as the SIGNING destination (the cowswap arm already models this spender/destination split - follow it). Blocked on #895 merging (the field must exist on the quote first). Test: an approve leg built from a quote carrying approvalAddress targets that address; absent the field, behavior unchanged.

Severity: MED/HIGH (real revert class on the SDK-direct path; abts path is covered once #895 lands). Effort: S once #895 is in.

rcoderdev:
Closes #1198

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ERC-20 allowance and approval preparation to use the swap quote’s `approvalAddress` as the spender/executor when provided (while still signing the swap with the router destination).
* **Tests**
  * Expanded AGG-02 routing and unit coverage to validate both approval and no-approval quote scenarios, including spender selection and signed destination behavior.
* **Chores**
  * Applied patch-level dependency/version updates (including an axios resolution).

### DIFF
--- a/.changeset/safe-lions-approve.md
+++ b/.changeset/safe-lions-approve.md
@@ -1,0 +1,8 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+For EVM swap providers that legitimately distinguish the allowance executor
+from the swap router, use the quote's approval address for SDK allowance checks
+and approval payloads while retaining the router as the signed destination.
--- a/package.json
+++ b/package.json
@@ -192,6 +192,7 @@
     "readable-stream": "3.6.2",
     "@grpc/grpc-js": "1.14.4",
     "@cosmjs/encoding/@scure/base": "1.2.6",
+    "axios": "1.18.1",
     "esbuild@npm:~0.28.0": "0.28.1",
     "shell-quote": "1.10.0",
     "js-yaml@npm:4.1.1": "^4.3.0",
--- a/packages/core/mpc/keysign/swap/build.agg02RouterCoverage.test.ts
+++ b/packages/core/mpc/keysign/swap/build.agg02RouterCoverage.test.ts
@@ -1,26 +1,28 @@
 import { create } from '@bufbuild/protobuf'
+import { initWasm, type WalletCore } from '@trustwallet/wallet-core'
 import { Chain } from '@vultisig/core-chain/Chain'
 import { SwapQuote } from '@vultisig/core-chain/swap/quote/SwapQuote'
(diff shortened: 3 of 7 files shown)


gomesalexandre (approved):
## lgtm 🫡

ran a full adversarial review (claude lane; codex is out of credits til 07-25) — traced the trust boundary end to end + ran the actual test files, no fund-loss path.

the question that mattered: is the quote's `approvalAddress` validated before it becomes the erc-20 approve spender? answer:
- **enforced providers (1inch/kyber)** never populate `approvalAddress` (`OneInchSwapQuoteResponse` has no such field), so `spender` falls back to the validated router — unchanged. and `assertEnforcedSwapApprovalSpenderBound` (sdk#1358, untouched here) still throws on the co-signer path if an enforced approve-spender diverges. `swapRouterGuard.test.ts`'s attacker-spender case passes unmodified.
- **LiFi/SwapKit (unenforced)**: `approvalAddress` is the same partner-sourced trust tier as `tx.to`, which was ALREADY the (unvalidated, log-only) spender for those providers pre-PR — so this doesn't cross a new trust boundary, it swaps which already-partner-trusted field is the spender.

swap destination stays the validated router (verified through the real `getEvmSigningInputs` resolver, not just field asserts). RN override mirrors in lockstep. tests pin both the router-fallback and executor cases.

approving.

should-fix (non-blocking fast-follow): `approvalAddress` isn't fed through `logUnenforcedAggregatorDestination` — only `to` is. since it's now a live spender for LiFi/SwapKit, it should get the same telemetry for the future allow-list dataset. one-liner, filing a bead so it isn't lost.

PR #1415, opened 2026-07-19
fix(cosmos): preserve exact account sequences

### LINKED ISSUE #1307: cosmos signing: sequence still narrows through JS number in getCosmosAccountInfo (same double-narrow bug #1306 fixes for QBTC account_number)

rcoderdev:
Closes #1307

## Summary by CodeRabbit

* **Bug Fixes**
  * Cosmos `uint64` account sequences are now preserved exactly end-to-end via a new `sequenceBigInt` value, including large values beyond JavaScript safe integers.
  * Invalid or unverifiable LCD/RPC sequence inputs now fail closed instead of being coerced or defaulted.
* **New Features**
  * Added a validated `uint64` parsing utility and exported it for Cosmos usage.
* **Tests**
  * Expanded regression coverage for sequence precision, LCD parsing edge cases, fallback behavior, and signing payload construction.
* **Chores**
  * Added patch release changesets and updated dependency resolutions.

### DIFF
--- a/.changeset/bright-trackers-link.md
+++ b/.changeset/bright-trackers-link.md
@@ -1,0 +1,8 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Route SwapKit transaction links through its public tracker across the shared core and SDK APIs.
--- a/.changeset/fuzzy-taxis-broadcast.md
+++ b/.changeset/fuzzy-taxis-broadcast.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Align Bittensor broadcast error handling with the shared Substrate safeguards.
--- a/.changeset/honest-vaults-guard.md
+++ b/.changeset/honest-vaults-guard.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Expose the native-swap expiry and inbound-vault guard from core MPC so shipping
+wallets and the SDK facade can share one pre-signing or pre-broadcast check.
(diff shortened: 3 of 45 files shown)


johnnyluo (approved):
LGTM

PR #1417, opened 2026-07-19
fix(sdk): publish TON chain subpath

### LINKED ISSUE #1416: public API: publish @vultisig/sdk/chains/ton as a real package subpath
## Summary

`packages/sdk/src/chains/ton/index.ts` is already a coherent SDK-owned helper module, and React Native already republishes it from `@vultisig/sdk/react-native`, but the package still does **not** publish a real `@vultisig/sdk/chains/ton` subpath for other consumers.

That keeps TON signing-payload / wallet helpers on the wrong side of the public boundary and pushes first-party consumers toward deep imports or reimplementation.

## Evidence

- `packages/sdk/src/chains/ton/index.ts:1-18`
  - already defines a consumer-shaped TON helper module:
    - `buildTonSendTx`
    - `buildTonJettonTransferTx`
    - `buildTonTxFromSigningPayload`
    - `buildV4R2Wallet`
    - `getTonWalletInfo`
    - `validateTonMemo`
- `packages/sdk/src/platforms/react-native/chains/ton/index.ts:1-33`
  - RN already republishes that same helper family.
- Before round 44, `packages/sdk/package.json` exported `./tools/parse` and `./tools/defi`, but there was no `./chains/ton` export-map entry.

## Impact

This is a classic public-surface drift vector:

- the canonical logic already lives in the SDK,
- one first-party surface (RN) can consume it,
- but other package consumers cannot import it from a stable public path.

That encourages deep imports into internal layout or fresh local wrappers for TON-specific signing-payload and wallet logic.

It already **does** live in the SDK — the missing piece is publishing it as a first-class package subpath with dedicated JS + DTS bundles so all consumers can rely on the same stable surface.

## CLI / isolated-testability

Yes.

This is a strong SDK packaging candidate because it can be verified with:

- export-map regression tests
- `yarn build:sdk`
- `yarn cli:build`
- packed-tarball runtime + type smoke proving `@vultisig/sdk/chains/ton` resolves correctly

## Fix sketch

- add `./chains/ton` to `packages/sdk/package.json#exports`
- add dedicated JS bundling in `packages/sdk/rollup.platforms.config.js`
- add dedicated DTS bundling in `packages/sdk/rollup.types.config.js`
- add/extend a public-API regression test
- extend the tarball smoke script to resolve/import/typecheck `@vultisig/sdk/chains/ton`

## Effort

Low.

This is mostly packaging/public-surface wiring around an already-existing helper module.

## Report

Round 44 report: https://gist.github.com/gomesalexandre/361f89db48b61adf2dc422796c60e666

gomesalexandre:
## Summary
- publish `@vultisig/sdk/chains/ton` as a real package subpath
- add dedicated JS + DTS bundle wiring for the TON helper module
- extend the public-API regression test and tarball smoke to prove the subpath resolves at runtime and in types

## Why
`packages/sdk/src/chains/ton/index.ts` was already a coherent SDK-owned helper module, and `@vultisig/sdk/react-native` already republished it, but the package still had no real `@vultisig/sdk/chains/ton` entrypoint for other consumers. That pushed first-party consumers toward deep imports or reimplementation even though the canonical code already lived in the SDK.

Closes #1416.

## Verification
- `yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/public-api/toolsSubpathExports.test.ts`
- `yarn build:sdk`
- `yarn cli:build`
- `node scripts/smoke-sdk-subpath-exports.mjs`

## Receipts
- targeted public-API test passed with 2/2 assertions green
- `yarn build:sdk` emitted:
  - `dist/chains/ton/index.js`
  - `dist/chains/ton/index.cjs`
  - `dist/chains/ton/index.d.ts`
- `yarn cli:build` completed successfully
- tarball/runtime smoke proved the packaged subpath resolves correctly:

```json
{"parsePath":".../dist/tools/parse/index.cjs","defiPath":".../dist/tools/defi/index.cjs","tonPath":".../dist/chains/ton/index.cjs","parseOk":true,"defiOk":true,"tonOk":true}
```

## Round report
- Round 44 report mirror: https://gist.github.com/gomesalexandre/361f89db48b61adf2dc422796c60e666

## Notes
- `yarn build:sdk` still shows pre-existing workspace warnings (e.g. Rollup circular-dependency noise / unused `viem` import in `src/tools/dex/uniswapV2Addresses.ts`), but the build and smoke receipts are green.


packages/sdk/package.json
```diff
@@ -98,6 +98,19 @@
       "import": "./dist/tools/defi/index.js",
       "require": "./dist/tools/defi/index.cjs",
       "default": "./dist/tools/defi/index.cjs"
+    },
+    "./chains/ton": {
+      "types": "./dist/chains/ton/index.d.ts",
+      "browser": "./dist/chains/ton/index.js",
+      "worker": "./dist/chains/ton/index.js",
+      "react-native": "./dist/chains/ton/index.js",
```
NeOMakinG, line 122: `blocker:` these three conditions point at `./dist/chains/ton/index.js`, which is emitted **only** by `createToolsSubpathConfigs` inside `configs.node` (`rollup.platforms.config.js:239`), built with `preferBuiltins: true` and `exportConditions: ['node', ...]` (`:206`). Its first line is `import t from"crypto"`.

REPRODUCED — one-line consumer, esbuild through each declared condition, with the two sibling subpaths as controls:
```
--conditions=react-native  chains/ton  -> Could not resolve "crypto"
--conditions=browser       chains/ton  -> Could not resolve "crypto"
--conditions=worker        chains/ton  -> Could not resolve "crypto"
--conditions=react-native  tools/defi  -> 0 errors   (control)
--conditions=react-native  tools/parse -> 0 errors   (control)
```

On RN it is worse than a build error: `rollup.platforms.config.js:392-395` tells consumers to map `crypto` to an **empty** module via `resolver.extraNodeModules`, so it bundles successfully and then `crypto.createHash` is `undefined` inside `@ton/core`'s `boc/cell/wonderCalculator.js` — the Cell-hash path every `buildTonSendTx` / `deriveTonAddress` / `buildV4R2Wallet` call goes through.

Root cause: `@ton/core` peer-depends on `@ton/crypto` -> `@ton/crypto-primitives@2.1.0`, whose `package.json` is `main: dist/node.js` (createHash) / `browser: dist/browser.js` (crypto.subtle) / `react-native: dist/native.js` (jssha). Serving ONE file to all conditions destroys that selection, and this build baked in the node branch — the exact package `src/chains/ton/crypto-rn.ts:1-20` exists to avoid.

Fix: either delete these three keys so the map matches the node-only artifact (RN already has a correct route via `@vultisig/sdk/react-native`'s `chains.ton`), or emit real `index.browser.js` / `index.react-native.js` targets — the RN one needs `'react-native'` prepended to `exportConditions` — and point each condition at its own file.

packages/sdk/rollup.platforms.config.js
```diff
@@ -326,6 +326,10 @@ const configs = {
       input: './src/tools/defi/index.ts',
       distBase: 'tools/defi',
     }),
+    ...createToolsSubpathConfigs({
+      input: './src/chains/ton/index.ts',
```
NeOMakinG, line 336: `blocker:` this is the whole of the build wiring, and it sits inside `configs.node` only — so exactly one artifact exists for a subpath whose export map declares `browser`, `worker` and `react-native` (see the comment on `package.json`).

`tools/parse` and `tools/defi` get away with the same template because their graphs are pure (viem / zod only, 0 unresolvable specifiers through every condition). `src/chains/ton` pulls `@ton/core` -> `@ton/crypto` -> `@ton/crypto-primitives/dist/node.js`, so this one does not. The question for a copied export-map template is never "does the shape match the siblings" but "is this module in the same purity class as the siblings".

Also worth knowing before choosing the fix: the emitted bundle references global `Buffer` 75 times and never imports it (`grep -c 'from"buffer"'` -> 0; controls: parse 0, defi 6). `Buffer` does not exist in browsers or workers, and on RN only after `@vultisig/sdk/rn-preamble` has run (`:350-353`) — which a subpath-only consumer has no reason to import.

packages/sdk/tests/unit/public-api/toolsSubpathExports.test.ts
```diff
@@ -26,22 +27,34 @@ describe('public API tools subpath exports', () => {
       require: './dist/tools/defi/index.cjs',
       default: './dist/tools/defi/index.cjs',
     })
+    expect(tonExport).toMatchObject({
```
NeOMakinG, line 37: `should-fix:` this `toMatchObject` lists `types` / `import` / `require` / `default` and omits **exactly** the three keys that are wrong: `browser`, `worker`, `react-native`. Same omission as the equivalent test on #1410.

Combined with the second `it()` — which greps the config-file text this PR itself added, so it passes whether or not rollup emits anything — nothing in the 4 green check-runs can fail on the blocker.

Cheapest real coverage: assert the **full** export object (`toEqual`, not `toMatchObject`) so adding or changing a condition has to be deliberate. Better still, bundle a fixture consumer once per declared condition and assert it resolves.

scripts/smoke-sdk-subpath-exports.mjs
```diff
@@ -52,15 +52,21 @@ try {
       "const require = createRequire(import.meta.url)",
       "const parsePath = require.resolve('@vultisig/sdk/tools/parse')",
       "const defiPath = require.resolve('@vultisig/sdk/tools/defi')",
+      "const tonPath = require.resolve('@vultisig/sdk/chains/ton')",
```
NeOMakinG, line 56: `should-fix:` this script is the only behavioural proof in the PR, and it runs in **zero** workflows — `grep -rn "smoke-sdk-subpath|smoke:subpath" .github/ package.json` returns one hit, `package.json:85`.

It is also node-conditions-only (`require.resolve` + `await import` from a plain node process), so even wired into CI it would exercise the `node` / `import` / `require` branches and never touch `browser` / `worker` / `react-native`.

If any non-node condition survives, this script is the natural place for a per-condition bundle check (esbuild `--conditions=<cond>` on a one-line consumer against the packed tarball) — and it needs an actual workflow step.

NeOMakinG (commented):
Reviewed as the sibling of #1410 (`tools/prep`), which had the same class of defect. **The same defect is here, and on this PR it is sharper**: the export map declares `browser` / `worker` / `react-native` for an artifact that is built exclusively by the node target — and for TON that artifact bakes in `@ton/crypto-primitives`' **node** implementation, destroying the per-platform selection that `src/chains/ton/crypto-rn.ts`'s own doc comment says the SDK depends on.

Draft, so **COMMENTED**, no approval, nothing pushed to your branch.

## testing approach

Worktree at `80d81feb`, `node_modules` symlinked from the canonical checkout, canonical `dist/` verified untouched afterwards.

1. **Built the real artifacts.** Filtered `rollup.platforms.config.js`'s default export by `output.file` into a throwaway config, `BUILD_TARGET=node`. Emitted `dist/chains/ton/index.js` (305,657 B) + `index.cjs` (305,736 B). Also built the two sibling subpaths as **controls**.
2. **Bundled a one-line consumer through every declared condition** with esbuild, from an external package that symlinks `node_modules/@vultisig/sdk` -> the built package.
3. **Resolution probes** from that external consumer: `require.resolve` + `import.meta.resolve`, with a positive control (`@vultisig/sdk/node`) and a negative control (`@vultisig/sdk/nope`).
4. **Real `import()` + `require()`**, symbol-count diff.
5. **d.ts** built via a throwaway types config carrying only this entry (615 ms).
6. **`yarn format:check`** with the repo's own flags against the gated glob.
7. **Sourcemap forensics** on the emitted bundle to name every third-party module inlined into it.
8. Second lane (Codex `gpt-5.3`) run in parallel on TON-specific / dependency-surface scope; its non-overlapping findings are folded in below.

Real CI: `gh api .../commits/80d81feb/check-runs --jq .total_count` -> **4**, all `completed/success` (Unit, Integration, Test Summary, Lint and Type Check). Genuinely green, and structurally blind to the blocker (see should-fix #3).

**What I verified as CORRECT — the mechanism is fine, this is not a "the plumbing is broken" review:**

| check | result |
|---|---|
| `require.resolve('@vultisig/sdk/chains/ton')` | `dist/chains/ton/index.cjs` |
| `import.meta.resolve(...)` | `dist/chains/ton/index.js` |
| positive control `@vultisig/sdk/node` | resolves |
| negative control `@vultisig/sdk/nope` | throws `ERR_PACKAGE_PATH_NOT_EXPORTED` |
| ESM vs CJS symbols | **11 == 11, zero diff** |
| `dist/chains/ton/index.d.ts` | builds, 238 lines, externalizes only `@ton/core` (a declared `dependency`) |
| `files` in `package.json` | `["dist", ...]` — the new assets pack |
| `yarn format:check` (repo flags, gated glob) | **clean** |
| orphan promotion vs the root barrel | **none** — all 11 runtime symbols are already public through `@vultisig/sdk/react-native`'s `chains.ton` (`src/platforms/react-native/chains/ton/index.ts`) |

Full granted surface (11 runtime symbols, identical ESM/CJS): `TON_V4R2_SUB_WALLET_ID`, `broadcastTonTx`, `buildTonJettonTransferTx`, `buildTonSendTx`, `buildTonTxFromSigningPayload`, `buildV4R2Wallet`, `deriveTonAddress`, `getTonBalance`, `getTonWalletInfo`, `sha256`, `validateTonMemo`. Nothing key-adjacent among the *declared* symbols — but see q #6 for what the bundle drags in behind them.

## blocking

**1. `browser` / `worker` / `react-native` all point at an artifact only the node target emits — REPRODUCED on all three, with clean controls.**

`createToolsSubpathConfigs` (`rollup.platforms.config.js:239`) lives only inside `configs.node`, is built with `preferBuiltins: true`, and `createPlugins` picks `exportConditions: ['node', 'module', 'import', 'default']` (`:206`). The emitted ESM's first line is literally:

```js
import t from"crypto"
```

One-line consumer, bundled through each declared condition:

```
esbuild <entry>.mjs --bundle --format=esm --platform=browser --conditions=<cond>

cond=react-native  entry=chains/ton   -> Could not resolve "crypto"
cond=browser       entry=chains/ton   -> Could not resolve "crypto"
cond=worker        entry=chains/ton   -> Could not resolve "crypto"

cond=react-native  entry=tools/defi   -> 0 errors      <- CONTROL
cond=react-native  entry=tools/parse  -> 0 errors      <- CONTROL
cond=browser       entry=tools/defi   -> 0 errors      <- CONTROL
cond=browser       entry=tools/parse  -> 0 errors      <- CONTROL
cond=worker        entry=tools/defi   -> 0 errors      <- CONTROL
cond=worker        entry=tools/parse  -> 0 errors      <- CONTROL
```

So this is **new** breakage, not inherited: the export-map template is correct for its two existing users, and TON is the first one in a different purity class.

**And on React Native the failure is worse than a build error.** `rollup.platforms.config.js:392-395` documents the repo's own remedy for exactly this specifier:

> `// Node builtins — kept external; consumers must map these to`
> `// empty modules via metro.config.js `resolver.extraNodeModules`.`
> `// The SDK ships `dist/shims/empty-rn.js` as the canonical target.`

An RN consumer who follows that instruction gets a *successful* bundle in which `crypto` is the empty shim — and then `crypto.createHash` is `undefined` at call time, inside `@ton/core`'s `boc/cell/wonderCalculator.js` (confirmed present in the bundle's sourcemap). That is the Cell-hash path, i.e. every `buildTonSendTx` / `buildTonJettonTransferTx` / `deriveTonAddress` / `buildV4R2Wallet` call. Silent-until-you-sign, not fail-at-build.

**Root cause, precisely.** `@ton/core` peer-depends on `@ton/crypto`, which depends on `@ton/crypto-primitives@2.1.0`, whose `package.json` is:

```json
"main": "./dist/node.js",            // crypto.createHash
"browser": "./dist/browser.js",      // crypto.subtle
"react-native": "./dist/native.js"   // jssha
```

Those three lines are quoted almost verbatim in `src/chains/ton/crypto-rn.ts:1-20`, which exists *specifically* so the SDK never reaches into that package. But `@ton/core` drags it in transitively regardless, and this build resolves the **node** branch. The subpath sourcemap names `@ton/crypto-primitives/dist/node.js`, `.../node/sha256.js`, `.../node/sha512.js`, `.../node/hmac_sha512.js`, `.../node/pbkdf2_sha512.js`, `.../node/getSecureRandom.js`. Because the export map serves **one file** to all conditions, the browser/RN branches that package ships can never be selected. Rebuilding the same entry with `browser: true` + browser conditions selects `dist/browser.js` (`crypto.subtle`) instead — which `crypto-rn.ts` documents as throwing immediately under Hermes. Neither of the two condition sets this repo's `createPlugins` can produce yields the Hermes-safe `native.js`/jssha path.

**Two ways out, pick one:**

- **Cheap and honest:** drop `browser`, `worker`, `react-native` from the `./chains/ton` entry so the export map matches what the artifact actually is (node-only). RN consumers already have a working, Hermes-correct route via `@vultisig/sdk/react-native`'s `chains.ton`, which is where `vultiagent-app/src/services/tonTx.ts` already imports from.
- **Complete:** add real browser/RN targets emitting `dist/chains/ton/index.browser.js` and `dist/chains/ton/index.react-native.js` — the RN one needs `'react-native'` prepended to `exportConditions` so `@ton/crypto-primitives` resolves to `native.js` — and point each condition at its own file.

## preferably-blocking

**2. The bundle uses global `Buffer` 75+ times and never imports it.**

`grep -c 'from"buffer"' dist/chains/ton/index.js` -> `0`, while `Buffer.` appears 75 times (159 `Buffer` tokens). Controls: `tools/parse` 0, `tools/defi` 6.

`Buffer` is not a global in browsers or workers at all. On RN it exists only after `@vultisig/sdk/rn-preamble` has run — the repo's own comment at `rollup.platforms.config.js:350-353` says consumers "are expected to `import '@vultisig/sdk/rn-preamble'` as their first entry statement". A consumer who only ever imports `@vultisig/sdk/chains/ton` gets no signal that a preamble exists. If you keep any non-node condition, this needs either an explicit `import { Buffer } from 'buffer'` in the emitted bundle or a documented preamble requirement on the subpath. Folded into the same fix as #1 if you take the node-only route.

## should-fix

**3. Nothing in CI can fail on the blocker, and the new test is shaped so it never will.**

`tests/unit/public-api/toolsSubpathExports.test.ts` asserts (a) that the config *files* contain the strings this PR itself added — a tautology that passes whether or not rollup emits anything — and (b) a `toMatchObject` on `tonExport` listing `types` / `import` / `require` / `default`, which omits **exactly** `browser`, `worker`, `react-native`: the three keys that are wrong. Identical omission to #1410's test.

The only behavioural proof is `scripts/smoke-sdk-subpath-exports.mjs`, and it is wired into **zero** workflows — `grep -rn "smoke-sdk-subpath\|smoke:subpath" .github/ package.json` returns one hit, `package.json:85`. It is also node-conditions-only, so even if it were wired it would not catch this. A per-condition bundle assertion (or just asserting the full export object including the three conditions) is what makes this class visible.

**4. `getTonBalance` means two different things depending on which entry point you import it from.**

- root `@vultisig/sdk` -> `src/tools/balance/otherBalance.ts:296`, `getTonBalance(address): Promise<TonBalance>`, hits `${ROOT_API_URL}/ton/v2/getExtendedAddressInformation`, returns `{chain, ticker, address, balanceNano, balance, status, seqno, asOf}`.
- new `@vultisig/sdk/chains/ton` -> `src/chains/ton/rpc.ts:60`, `getTonBalance(address, gatewayUrl): Promise<bigint>`, nanotons, caller-supplied gateway.

Same package, same symbol name, different arity, different return type, different endpoint. TypeScript catches most misuse, but this is a genuine footgun for anyone migrating an import path. Consider `getTonBalanceNano` (or re-exporting the root one) before this ships as a public contract.

**5. Two public routes to the same TON module graph.** (second-lane finding, verified) A consumer that imports both `@vultisig/sdk/react-native` and `@vultisig/sdk/chains/ton` loads two copies of `@ton/core` in Metro, with two copies of the module-local `cachedV4R2CodeCell` (`src/chains/ton/walletV4R2.ts:44`). I looked for a correctness break and found none — the cache is a pure derived constant — so this is cost + confusion, not a bug. Worth one line in the changeset saying which route app consumers should use.

## suggestion

**6.** Zero tree-shaking, same as #1410: `createToolsSubpathConfigs` uses `inlineDynamicImports: true` and a single-file output, so importing just `validateTonMemo` still pulls the whole 305 KB. Not new to this PR, noting it because the subpath makes it consumer-visible.

## q

**7. Was it intended that this subpath ships `@ton/crypto`'s mnemonic / HD-key surface?** Sourcemap forensics on `dist/chains/ton/index.js` shows these inlined, reachable behind a tx-builder subpath:

```
@ton/crypto/dist/mnemonic/mnemonic.js
@ton/crypto/dist/mnemonic/wordlist.js
@ton/crypto/dist/hd/ed25519.js
@ton/crypto/dist/hd/mnemonics.js
@ton/crypto/dist/hd/symmetric.js
@ton/crypto/dist/primitives/nacl.js
@ton/crypto/dist/primitives/getSecureRandom.js
@ton/crypto/dist/passwords/newSecurePassphrase.js
@ton/crypto/dist/passwords/newSecureWords.js
jssha/dist/sha.mjs
```

None of it is *exported* from the subpath, so this is not a new public key-handling API — but it is key-material code now inlined into an artifact whose changeset advertises "TON helper imports". It rides in transitively via `@ton/core`'s peer dep. If the node-only route in #1 is taken this stays as-is; flagging it so the choice is deliberate.

**8. Generic `sha256` export.** `sha256` is exported unqualified from a chain subpath (it is the `@noble/hashes` shim from `crypto-rn.ts`). Harmless, but `@vultisig/sdk/chains/ton` exporting a bare `sha256` invites `import { sha256 } from '@vultisig/sdk/chains/ton'` in code that has nothing to do with TON.

**9. `@ton/core` is a `dependency`, not a `peerDependency`,** but `src/platforms/react-native/index.ts:145` tells consumers "Consumers MUST install `@ton/core` as a peer dep" and `peerDependencies` is `{}`. The emitted `.d.ts` also does `import { Address, StateInit } from '@ton/core'` without re-exporting them, so a consumer who wants to name those types must resolve `@ton/core` themselves — which is not guaranteed under pnpm's isolated layout. Which is it meant to be?

I did **not** carry over the second lane's "floating `^0.63.1` on a byte-sensitive API" as a finding: for a `0.x` version a caret locks the minor, so `^0.63.1` is `>=0.63.1 <0.64.0` — patch-only drift. Real but small; not worth a row.

## risk

Medium-high **if published as-is**, and it will not be caught before npm: consumers resolving `browser` / `worker` fail at bundle time (loud), and RN consumers who follow the repo's documented `crypto` -> empty-shim mapping fail at *runtime, on the Cell-hash path*, i.e. while building a TON transfer (quiet). No existing route regresses — `@vultisig/sdk/react-native` and the root entry are untouched. Everything else about the change (resolution, dual-format parity, types, packing, formatting) is correct.

## verdict

**COMMENTED** — draft, so no approval either way, and no fixes pushed to your branch. One blocker with a reproduction and a two-option fix, one preferably-blocking that the same fix closes, and a test/CI gap that let a green board hide both. The mechanism you built is right; only the condition mapping is wrong, exactly as on #1410.

---
<sub>reviewed by an automated reviewer agent (2-lane: Claude + Codex `gpt-5.3`), artifacts built and bundled locally against `80d81feb`</sub>

rcoderdev:
Closing this because the browser, worker, and React Native conditions resolve to a Node-only TON bundle that breaks those runtimes, and the linked issue has since been consolidated; a replacement should publish tested per-condition artifacts from current main.

PR #1419, opened 2026-07-19
fix(cosmos): validate secured asset chain prefixes

rcoderdev:
## Summary by CodeRabbit

- **Bug Fixes**
  - Secured-asset signing now rejects deposits that use unsupported or non-canonical chain prefixes.
  - Native swap asset denom prefixes are interpreted case-insensitively for consistent chain resolution.
  - Prevents incorrect chain values from being included in secured-asset deposit payloads.

- **Tests**
  - Added parameterized coverage for secured assets using invalid chain prefixes and verified the correct “unsupported prefix” error is thrown.

### DIFF
--- a/.changeset/guard-secured-asset-prefix.md
+++ b/.changeset/guard-secured-asset-prefix.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+---
+
+Reject secured-asset signing payloads whose chain prefix is not in the canonical native-swap registry.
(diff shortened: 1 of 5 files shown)


gomesalexandre (approved):
lgtm. This turns a silent garbage-prefix accept into a fail-closed guard on the cosmos secured-asset signing path, with tests actually pinned to the behavior.

what I checked
- unknown secured-asset prefixes now fail instead of being uppercased into a bogus `THORChainAsset.chain`
- the prefix->chain-id lookup was deduped into `NativeSwapChain.ts` and both call sites now share it
- happy-path behavior for known prefixes is unchanged because the canonical registry ids are already the same uppercase strings the old code emitted
- changeset is present and scoped to `@vultisig/core-chain` + `@vultisig/core-mpc`

receipts
- targeted tests for `cosmos.test.ts` + `toNativeSwapAsset.test.ts` green
- mutation check: restoring `cosmos/index.ts` from `origin/main` makes the new rejection tests fail, so they genuinely pin the fix
- full core suite green in the review checkout
- `tsc -p .config/tsconfig.json --noEmit` clean

non-blocking
- Maya-only prefixes still pass the generic registry guard for this THOR secured-asset deposit path; tightening that to a THOR-only allowlist could be a later hardening pass, but this PR already closes the fail-open hole it set out to fix

PR #1284, opened 2026-07-15
feat(sdk): expose CosmWasm execute builders

rcoderdev:
Closes #1187
Built SDK package proof confirms the root builders produce the canonical Amino envelope and protobuf SignDoc byte parity; the React Native conditional root now exposes the same generic Amino builder.

PR #1270, opened 2026-07-14
Export Cosmos fee helpers from the React Native SDK entrypoint

rcoderdev:
Refs #1199

## Summary by CodeRabbit

* **New Features**
  * Added Cosmos fee-denomination helpers to the React Native SDK exports.
  * Applications can now retrieve allowed Cosmos fee denominations and validate whether a denomination is supported.

* **Bug Fixes**
  * Improved access to the canonical Cosmos fee-denomination validation functionality through the React Native entrypoint.

PR #1421, opened 2026-07-19
fix(sdk): export Blockaid EVM canonicals

### LINKED ISSUE #1420: public API: export Blockaid EVM canonicals from @vultisig/sdk
## Summary

Root `@vultisig/sdk` omits the canonical Blockaid-supported EVM chain map, forcing first-party consumers to keep backend-local security copies.

## Evidence

- Backend consumer comment documenting the public-surface gap:
  - `agent-backend-ts/src/mastra/tools/mcp/lib/security/blockaidScanRequest.ts:42-52`
  - explicitly says `@vultisig/sdk` does not re-export `blockaidEvmChain` and manual sync is required
- Backend local copy:
  - `agent-backend-ts/src/mastra/tools/mcp/lib/security/blockaidScanRequest.ts:69-82`
- Canonical SDK/core owner already exists:
  - `packages/core/chain/security/blockaid/evmChains.ts:3-20`

## Impact

Blockaid-supported chain coverage is security metadata. If supported EVM chains change, first-party consumers can silently drift from the SDK-owned canonical map.

## Why this should live in the public SDK surface

First-party consumers should be able to depend on `@vultisig/sdk` for the reviewed canonical map instead of mirroring it locally. This directly reduces cross-repo security drift.

## Fix sketch

Export these from root `@vultisig/sdk`:
- `blockaidEvmChain`
- `blockaidSupportedEvmChains`
- `BlockaidSupportedEvmChain`

Follow-up: rewire `agent-backend-ts` to consume the public SDK export instead of the local copy.

## Verification target

- public-export regression test
- `yarn build:sdk`
- `yarn cli:build`
- built-dist runtime proof against `packages/sdk/dist/index.node.esm.js`

## Effort

Small.

## Report

Round-45 report mirror: https://gist.github.com/gomesalexandre/20518a6b017db7e6e03c1f33a7078ce8

gomesalexandre:
## Summary

Export the canonical Blockaid-supported EVM chain map from root `@vultisig/sdk` so first-party consumers can stop mirroring backend security metadata.

Closes #1420.

## What changed

- re-export `blockaidEvmChain` from root `@vultisig/sdk`
- re-export `blockaidSupportedEvmChains` from root `@vultisig/sdk`
- re-export the `BlockaidSupportedEvmChain` type from root `@vultisig/sdk`
- extend `packages/sdk/tests/unit/utils/publicExports.test.ts` to lock the new public surface
- add a changeset for the SDK/CLI package pair

## Why

`agent-backend-ts` currently documents and maintains a local copy of the SDK-owned Blockaid EVM chain map because the public SDK surface does not expose it:

- `src/mastra/tools/mcp/lib/security/blockaidScanRequest.ts:42-52`
- local map copy at `src/mastra/tools/mcp/lib/security/blockaidScanRequest.ts:69-82`

That is exactly the cross-repo drift pattern this /improve campaign keeps finding. This patch removes the public-surface gap so backend consumers can depend on the canonical SDK export instead of mirroring the security allowlist.

## Verification

- `yarn build:sdk`
- `yarn cli:build`
- built-dist runtime proof:
  - `node --input-type=module -e "const sdk = await import('./packages/sdk/dist/index.node.esm.js'); console.log(JSON.stringify({base: sdk.blockaidEvmChain.Base, hyperliquid: sdk.blockaidEvmChain.Hyperliquid, chains: sdk.blockaidSupportedEvmChains.slice(0,4), count: sdk.blockaidSupportedEvmChains.length}))"`
  - output: `{"base":"base","hyperliquid":"hyperevm","chains":["Arbitrum","Avalanche","Base","Blast"],"count":12}`

## Constraints / notes

- Round-45 report mirror: https://gist.github.com/gomesalexandre/20518a6b017db7e6e03c1f33a7078ce8
- A packed-tarball smoke attempt hit local disk exhaustion (`ENOSPC`) after the Vitest/build/runtime receipts were already complete, so I am recording that packaging constraint explicitly instead of fabricating a tarball proof.


packages/sdk/src/index.ts
```diff
@@ -222,6 +222,11 @@ export {
   isCosmosMemoWithinCap,
 } from '@vultisig/core-chain/chains/cosmos/cosmosMemoCap'
 
+// Blockaid-supported EVM-chain canonicals — surfaced so first-party consumers
+// stop hand-copying the backend security map that the SDK already owns.
+export type { BlockaidSupportedEvmChain } from '@vultisig/core-chain/security/blockaid/evmChains'
+export { blockaidEvmChain, blockaidSupportedEvmChains } from '@vultisig/core-chain/security/blockaid/evmChains'
```
NeOMakinG, line 228: `should-fix:` this lands in the root barrel, but `exports["."]["react-native"]` resolves to `./dist/index.react-native.js`, built from `src/platforms/react-native/index.ts` — the one platform barrel with no `export * from ../../index` (node `:118`, browser `:63`, chrome-extension `:67`, electron `:118` all have it). So under Metro these three symbols are `undefined` at runtime and absent from `index.react-native.d.ts`. Pre-existing barrel shape, and there is no RN consumer today, so not blocking — but it is the #1417/#1410/#1321/#1407 declared-but-unresolvable class and this is the cheap moment. Either add them to the RN barrel (the #1270 precedent) or scope the changeset to node/browser explicitly.

NeOMakinG (commented):
Round 1, `089c2ce7`. **COMMENTED — draft, so no approval by policy.** Two lanes: Claude (load-bearing) + Codex `gpt-5.4`; both lanes independently landed on the React Native gap and nothing else overlapped.

Headline: the thing I expected to be wrong isn't. The lead worry on an "export the canonical map" PR is a *parallel hand-maintained list* that drifts from the map it claims to canonicalise — `blockaidSupportedEvmChains` is `Object.keys(blockaidEvmChain)`, so that class is structurally impossible here. And the map is identical to the copy it is meant to replace, so the migration loses zero chains. What's left is one real surface gap and one shape decision this PR permanently locks in.

## testing approach

Worktree at `089c2ce7`; `node_modules` + the eight `packages/**/dist` trees symlinked from the canonical checkout per repo convention (no install).

- Read `packages/core/chain/security/blockaid/evmChains.ts` and its two neighbours (`validationChains.ts`, `simulationChains.ts`) rather than trusting the diff.
- **Coverage-delta probe against the consumer this PR exists for.** Diffed the exported map against `agent-backend-ts/src/mastra/tools/mcp/lib/security/blockaidScanRequest.ts:71-83`. **Identical — 12 entries, same keys, same slugs**, and abts' own header already records the same "last synced 2026-04-29, 12 entries" list. mcp-ts holds the same duplicate at `src/lib/security/blockaidScanRequest.ts:71`.
- **Slug-correctness probe.** Rather than checking the slugs against docs, traced whether they are already load-bearing in production: `blockaidEvmChain[chain]` is what fills the `chain` field of the live scan requests in `packages/core/mpc/security/blockaid/tx/validation/input/resolvers/evm.ts:25` and `.../simulation/input/resolvers/evm.ts:32`. These are prod-proven values, not a fresh transcription.
- **Condition-by-condition resolution audit** of `exports['.']` in `packages/sdk/package.json` against every platform barrel (see #1).
- `yarn workspace @vultisig/sdk vitest run … publicExports.test.ts` → **18 passed / 18**.
- CI truth: `check-runs total_count: 4`, all `completed/success` (Unit, Integration, Lint and Type Check, Test Summary).
- Enumerated the diff's dependency surface: **zero deps added, zero caret ranges added or widened.**
- Attempted a built-artifact probe of `dist/index.react-native.js` vs `dist/index.node.esm.js` and **discarded it**: the local `dist` is stale (2.18.5's `getCosmosMemoMaxBytes` has zero hits in either file at SDK 2.20.0), so it cannot answer this question. Resting #1 on source reading instead, which is unambiguous.

## should-fix

**1. Under the `react-native` condition, root `@vultisig/sdk` does not resolve these canonicals — the surface the PR names.** `exports['.']` fans out per condition, and none of the targets is `src/index.ts` directly:

| condition | built from | `export * from '../../index'`? |
|---|---|---|
| `node`, `import`, `require`, `default` | `src/platforms/node/index.ts` | yes — `:118` |
| `browser`, `worker` | `src/platforms/browser/index.ts` | yes — `:63` |
| `chrome-extension` | `src/platforms/chrome-extension/index.ts` | yes — `:67` |
| `electron` | `src/platforms/electron-main/index.ts` | yes — `:118` |
| **`react-native`** | `src/platforms/react-native/index.ts` | **no** |

The RN barrel is hand-curated — 40 explicit re-export sources, none of them `../../index`. So `import { blockaidEvmChain } from '@vultisig/sdk'` under Metro is `undefined` at runtime and a type error against `dist/index.react-native.d.ts`.

Codex rated this **preferably-blocking**; I am **downgrading it one notch to should-fix**, and stating why rather than parroting the tier: (a) the divergence is pre-existing — `git show origin/main:packages/sdk/src/platforms/react-native/index.ts | grep -c "from '../../index'"` returns `0`, so it is the shape of the RN barrel, not something this diff broke; (b) there is no RN consumer today — vultiagent-app keeps no local Blockaid EVM chain map and its scanning is server-side, so nothing is currently blocked. Still worth resolving before this leaves draft, because it is the same declared-but-unresolvable class as #1417/#1410/#1321/#1407 and this is the cheap moment to close it. Either add the three symbols to the RN barrel (the #1270 precedent did exactly this for the Cosmos fee-denom helpers) or say in the changeset that the canonicals are node/browser-only.

**2. This PR is the moment a mutable security allowlist becomes public API, and the shape is then locked by semver.** `packages/core/chain/security/blockaid/evmChains.ts:20`:

```ts
export const blockaidSupportedEvmChains = Object.keys(blockaidEvmChain) as BlockaidSupportedEvmChain[]
```

A plain mutable `BlockaidSupportedEvmChain[]`, and `blockaidEvmChain`'s `as const` is compile-time only — neither is frozen. Anything in the same process can `splice()` an entry out or repoint a slug, and because the SDK's *own* scanners read `blockaidEvmChain[chain]` per request (the two resolvers above), that retargets a live security scan rather than merely confusing a consumer. In abts an unmatched chain lands on `{ kind: 'unsupported', reason: 'chain_not_supported' }`, which **warn-opens** (`blockaidScanRequest.ts:35,229`) — so the failure direction is open, not closed.

Low probability, but the cost of getting it right is one line now versus a breaking change later:

```ts
export const blockaidEvmChain = Object.freeze({ … } as const)
export const blockaidSupportedEvmChains: readonly BlockaidSupportedEvmChain[] =
  Object.freeze(Object.keys(blockaidEvmChain) as BlockaidSupportedEvmChain[])
```

Widening `readonly T[]` → `T[]` later is breaking; narrowing now is free.

## q

**3. `publicExports.test.ts` is structurally incapable of catching #1.** It does `import * as sdk from '../../../src/index'` — the source root barrel, which is not what *any* published condition resolves to. It will stay green for every symbol the RN barrel omits. The only evidence crossing the source→artifact boundary is the manual `dist/index.node.esm.js` probe in the PR description, which covers 1 of 6 conditions and is not in CI. Not asking for a build matrix here — just flagging that the test's name ("public exports") claims more than its import can deliver. Same lesson as #1311.

**4. Changeset bump level.** Codex flagged `patch` as wrong for an additive public export and wanted `minor`. Strict semver agrees, but repo precedent is genuinely split and I am not charging it: #1284 ("Expose generic CosmWasm … execute builders from the root SDK") shipped **minor**, while #1022 and #1020 — both yours, both adding new root exports, #1020 explicitly described as "purely additive" — shipped **patch**, as did #1270's RN export addition. Worth settling as a repo convention rather than per-PR.

**5. `CronosChain` is the one `EvmChain` member absent from the map** (`packages/core/chain/Chain.ts:13-22` has 13, the map has 12). abts documents this and lets it warn-open. Now that the map is public, a one-line comment in `evmChains.ts` saying the omission is deliberate (Blockaid coverage, not an oversight) would stop the next consumer from "fixing" it.

**6. Scope: only the EVM third of the mirror is de-duplicated.** `blockaidValidationSupportedChains` and `blockaidSimulationSupportedChains` sit in the same directory and encode the Solana/Sui/Bitcoin support decision that abts also hardcodes (`blockaidScanRequest.ts:32-34`). Deliberate scoping, or a follow-up? One line in the PR body either way.

**7. Unrelated deletion.** The diff drops `expect(typeof sdk.getChainKind).toBe('function')` from a test it otherwise doesn't touch. Harmless — the next line throws if it were undefined — but unexplained in a PR whose stated test change is "extend … to lock the new public surface".

## verified good — stating explicitly, because these were the things most likely to be wrong

- **No parallel list.** `blockaidSupportedEvmChains` is derived from `blockaidEvmChain` via `Object.keys`, and `validationChains.ts` / `simulationChains.ts` both spread that same derived array. There is exactly one source of truth; the drift-between-two-lists failure cannot occur here.
- **Zero coverage loss on migration.** The exported map and abts' local copy agree entry-for-entry, keys and slugs. A consumer swapping the copy for the export changes nothing about which chains get scanned — so nothing silently starts warn-opening.
- **No orphan promotion (#1321 class).** Everything exported here was already exercised by first-party SDK code; nothing deliberately internal becomes newly reachable.
- **No dependency movement.** The diff adds no deps and widens no ranges, so the caret-float hazard doesn't apply.
- **Type derivation is honest.** `BlockaidSupportedEvmChain = keyof typeof blockaidEvmChain` resolves through the computed `EvmChain` keys to the literal union, and the test's `'Base'` / `'Hyperliquid'` assertions confirm the array holds Chain *names* while the map's values hold Blockaid *slugs* — the two namespaces don't get crossed.

- `yarn workspace @vultisig/sdk exec tsc --noEmit` reports 2 errors in `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts:56` and `packages/core/mpc/tx/compile/compileTx.golden.test.ts:994` (`'auxiliaryData' does not exist in type 'ISigningInput'`). Untouched by this diff and CI's Lint-and-Type-Check is green, so I read this as a WalletCore-version artifact of my symlinked `node_modules`. **Not charged to this PR.**
- The checked-in `packages/sdk/dist` is stale enough that 2.18.5-era exports are missing from it. Anyone reaching for it as a probe will get a wrong answer.
- Incidentally: I hit the same `ENOSPC` you did, mid-review. Not a repo problem, but the SDK worktree footprint is large enough that it is easy to trip.

## risk

Low. Additive re-export of a map that is already the SDK's internal source of truth for live Blockaid scan requests; no behaviour change on any existing path, no dependency movement, no new subpath and therefore none of the export-map hazards that sank the sibling PRs. The residual risk is entirely about the *shape* being frozen into the public API (#2) and about a consumer on RN discovering the symbol isn't there (#1).

## draft readiness

**Close, but not yet — two decisions rather than two bugs.** Merge-correctness is fine and CI is honestly green. Before flipping to ready I'd want: (a) #1 resolved one way or the other — add the three symbols to the RN barrel, or state the node/browser-only scope in the changeset so RN consumers aren't left to discover it; (b) #2 decided now rather than after publish, since `readonly` is free today and breaking tomorrow. #3–#7 are all fine to leave as-is with a sentence in the PR body.

<sub>reviewed by an AI agent — two lanes (Claude + Codex `gpt-5.4`); draft, so COMMENT only</sub>

packages/sdk/tests/unit/utils/publicExports.test.ts
```diff
@@ -110,8 +110,15 @@ describe('@vultisig/sdk public exports', () => {
     expect(typeof sdk.VaultBase).toBe('function')
   })
 
+  it('exports Blockaid-supported EVM chain canonicals for downstream security scans', () => {
+    expect(Array.isArray(sdk.blockaidSupportedEvmChains)).toBe(true)
+    expect(sdk.blockaidSupportedEvmChains).toContain('Base')
+    expect(sdk.blockaidSupportedEvmChains).toContain('Hyperliquid')
+    expect(sdk.blockaidEvmChain.Base).toBe('base')
+    expect(sdk.blockaidEvmChain.Hyperliquid).toBe('hyperevm')
```
NeOMakinG, line 118: `q:` this file imports `../../../src/index`, which is not what any published condition resolves to — every condition goes through a platform barrel. So this assertion cannot detect the React Native gap above, and will stay green for any symbol the RN barrel omits. The only source→artifact evidence is the manual `dist/index.node.esm.js` probe in the description (1 of 6 conditions, not in CI). Not asking for a build matrix — just noting the test proves less than its name suggests.

NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because this conflicted old head still omits the Blockaid exports from the published React Native condition, while its test only exercises the source barrel; please recreate the small patch from current main and verify every declared package condition.

PR #1423, opened 2026-07-20
fix(mpc): cardano send-max builds a real fee, not a zero-fee tx

### LINKED ISSUE #1382: fix(cardano): send-max produces fee=0 invalid tx, rejected at broadcast
Parent epic: vultisig/vultisig-sdk#1347

Surfaced while root-causing the masked SDK-vs-device hash divergences (#1367) during golden-vector hardening (#1368/#1381).

## what
wallet-core 4.7.0 returns `TransactionPlan{ amount: <full-input-unadjusted>, fee: 0 }` for a Cardano `Transfer` with `useMaxAmount + forceFee` (verified via BOTH `preImageHashes` and `AnySigner.plan`). A `fee=0` Cardano tx is invalid on-chain - nodes require `fee >= minFeeA + minFeeB * size` - so a "send max ADA" would be REJECTED at broadcast.

## impact
Not fund-loss (the tx never lands), but a broken user-facing path: send-max ADA fails at broadcast every time. Reliability / correctness.

## fix
Client-side fee-convergence: mirror the Bitcoin UTXO resolver's `AnySigner.plan`-before-hash pattern so the SDK derives a valid fee for the max-amount case instead of trusting wallet-core's fee=0 plan. Golden-vector coverage for the send-max Cardano case once fixed.

## repro
Build a Cardano `Transfer` signing input with `useMaxAmount = true` + `forceFee`, call `AnySigner.plan` -> observe `fee = 0` + `amount = full input`. Broadcast would be rejected by a real cardano node.

gomesalexandre:
Closes vultisig/vultisig-sdk#1382

## what

A Cardano **send-max** built an **unbroadcastable zero-fee transaction**. Every "send max ADA" the SDK signed would be rejected at broadcast (Cardano nodes require `fee >= minFeeA + minFeeB*txSize`).

## why

`getCardanoSigningInputs` paired `useMaxAmount: true` with `forceFee` (`resolvers/cardano.ts:51-53`). WalletCore's Cardano planner **ignores `forceFee` whenever `useMaxAmount` is set** - it returns `TransactionPlan{ amount: <full input>, fee: 0 }`. Confirmed empirically against this repo's pinned WalletCore via `AnySigner.plan`:

| input | result |
|---|---|
| `useMaxAmount:true` + `forceFee:44` (current) | `amount=10000000, fee=0` ❌ |
| `useMaxAmount:true` + `forceFee:164515` | `amount=10000000, fee=0` ❌ (forceFee ignored) |
| `useMaxAmount:false`, `amount=total-fee`, `forceFee` | `amount=9835485, fee=164515, change=0` ✅ |
| `useMaxAmount:true`, no `forceFee` | **WASM abort** (uncatchable) |

So `useMaxAmount` is unusable here: it zeroes the fee with a forced fee, and aborts without one.

## how

Never take the `useMaxAmount` path. A send-max is built as an **explicit transfer of `(totalInput - fee)`** with the converged fee forced and `useMaxAmount: false`, which the planner honors - a valid fee-bearing tx with the balance fully consumed (`change = 0`).

The fee is the one `getCardanoChainSpecific` already converges with the standard Cardano linear formula (`44*size + 155381`) - the **same proven path non-max sends use**. Its estimation loop calls back into this resolver, so it now prices the real fee-bearing body instead of the fee=0 one (this is *not* a new/unverified fee formula - it reuses the existing one, addressing the exact concern the golden-vector note raised).

Non-max sends and token sends are untouched.

## testing

- New regression test (`chainSpecific/resolvers/cardano.test.ts`) runs the **real** flow - `getCardanoChainSpecific` convergence → `getCardanoSigningInputs` → `AnySigner.plan` - and asserts `fee > 0`, `change = 0`, `amount + fee = totalInput`, `fee = converged byteFee`. Plus a non-max control (recipient gets exactly `toAmount`).
- Full `yarn test:core` green: **1864 passed | 1 skipped**. `tsc` + lint clean. Changeset added.
- The offline golden replay of the historical `Send ADA - max amount` fixture is re-pinned + its root-cause note rewritten: that fixture's `byte_fee: 44` is iOS's **per-byte-rate** parameter, not the SDK's **total forced-fee** field, so a raw offline replay (which skips `getCardanoChainSpecific`) forces `fee=44` and can't reproduce the device's converged fee. The pinned hash is a raw-replay artifact; the shipped behavior is validated by the convergence test above.

`@vultisig/core-mpc` ships to the flagship ios/android/windows clients. This only changes the **Cardano non-token send-max** branch (from a zero-fee tx to a valid one); non-max and token sends produce byte-identical signing inputs. Strictly-more-correct: it turns a guaranteed broadcast-rejection into a confirmable tx. No new throws, no network added.

## risk

Low - the previous behavior was 100% broken (fee=0, never confirms), so this can only improve Cardano send-max. Isolated to one branch of one resolver.

🤖 Agent-generated


rcoderdev (approved):

packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts
```diff
@@ -22,6 +23,21 @@ export const getCardanoSigningInputs: SigningInputsResolver<'cardano'> = ({ keys
...
+  // built as an EXPLICIT transfer of (totalInput - fee) with the converged fee
+  // forced and `useMaxAmount: false`, which the planner honors (fee > 0, change
+  // consumed to 0). The fee itself is converged by getCardanoChainSpecific, whose
+  // loop calls back into this resolver — so it now prices the real fee-bearing
+  // body instead of the fee=0 one. `useMaxAmount` without `forceFee` is not an
+  // option: the planner aborts (uncatchable WASM assert) with no forced fee.
+  const isSendMax = sendMaxAmount && !isTokenSend
+  const sendAmount = isSendMax
+    ? bigIntSum(keysignPayload.utxoInfo.map(({ amount }) => amount)) - byteFee
+    : BigInt(keysignPayload.toAmount)
```
NeOMakinG, line 39: `should-fix:` `isSendMax = sendMaxAmount && !isTokenSend` excludes token *sends* but not ADA send-max from UTXOs that *carry* native tokens. In that case `sendAmount = totalInput - byteFee` sends all ADA (change=0) and the input tokens have no output — `AnySigner.plan` returns `error=0` (verified), so the SDK blindly signs a value-not-conserved tx the node rejects. Also unfloored: balance<fee makes this negative and `Long.fromString` wraps it to a ~1.8e19 uint64; balance just over fee yields a sub-min-UTXO output. Suggest a fail-closed guard: reject send-max when any UTXO has `cardanoTokens`, and when `totalInput - byteFee < CARDANO_MIN_UTXO`. Fail-safe at the node today (no fund loss / no regression), so not blocking — but a clear pre-flight error beats a wasted co-sign ceremony.

NeOMakinG (approved):
Reviewed at HEAD `94b4f840`. This correctly fixes the core defect: WalletCore's Cardano planner silently drops `forceFee` when `useMaxAmount` is set (returns `amount=<full input>, fee=0`), so every ADA send-max the SDK signed was an unbroadcastable zero-fee tx. Building the max-send as an explicit `(totalInput − byteFee)` transfer with `useMaxAmount:false` + `forceFee` is the right move, and `byteFee` is the fee the existing convergence loop already prices — not a new formula.

## testing approach
Two-lane (Claude primary + Codex adversarial; GLM retired). Built shared packages, ran the full `yarn test:core` suite green (**1852 passed / 1 skipped**), and the Cardano + golden files specifically (76 passed). Then drove three adversarial edges through the real `getCardanoChainSpecific → getCardanoSigningInputs → AnySigner.plan` path in a throwaway probe on the pinned WalletCore:

| edge | result | verdict |
|---|---|---|
| ADA-only send-max, 6 ADA | `err=0 amount=5.83 fee=170649 change=0`, `amount+fee=total` | correct, exactly balanced |
| **token-bearing** send-max | `err=0 amount=5.83 fee=170649 change=0` — tokens dropped, **planner does NOT flag it** | signs an unbroadcastable value-not-conserved tx |
| send-max, balance 0.1 ADA (< fee) | `sendAmount=-64049` → uint64-wraps to `1.8e19`, `planErr=3` | not caught in chainSpecific, plan errors |
| send-max, balance 1.1 ADA | `err=0 amount=935951 fee=164049` — **0.936 ADA output < min-UTXO**, planner does NOT flag it | signs a sub-min-UTXO tx the node rejects |

Fee correctness confirmed: the convergence loop prices the **real final signed size** (body + `CARDANO_SIGNED_TX_ENVELOPE_SIZE` incl. the 104-byte vkey witness + aux), so it does not under-price by hashing a witness-less body. Happy path balances exactly (`amount+fee=totalInput`, `change=0`) with a real non-zero fee.

## blocking
None. The fixed path (ADA-only send-max, sufficient balance — the dominant case) is correct, exactly balanced, and well tested. Every edge below is **fail-safe at the network layer** (node rejects, ADA stays put, tokens are not burned — Cardano value-conservation forbids a silent burn) and is **not a regression**: the old `fee=0` path was 100% unbroadcastable for these same inputs too. So none meet the fund-loss / mis-balance / strand bar.

## should-fix
1. **Token-bearing ADA send-max signs a doomed tx (Codex HIGH, confirmed empirically).** `isSendMax = sendMaxAmount && !isTokenSend` only excludes token *sends*, not ADA sends from UTXOs that *carry* native tokens. In that case the output gets all ADA minus fee (`change=0`), the input tokens have nowhere to go, and — critically — `AnySigner.plan` returns `error=0`, so the SDK blindly MPC-signs a value-not-conserved tx that the node rejects at broadcast (wasting a full co-sign ceremony). No fund loss, but a guard would turn a silent doomed sign into a clear pre-flight error. Suggest: if `isSendMax && utxoInfo.some(u => u.cardanoTokens.length > 0)`, throw (or reserve a min-UTXO token-change output). Add a token-bearing send-max test.
2. **`sendAmount` isn't floored (Codex MEDIUM, confirmed).** For balance < fee, `totalInput − byteFee` goes negative and `Long.fromString` wraps it to ~`1.8e19` lovelace; for balance just above fee it yields a sub-min-UTXO output — both signed blindly. Suggest a fail-closed guard: `const totalInput = bigIntSum(...); if (isSendMax && totalInput - byteFee < CARDANO_MIN_UTXO) throw <insufficient-for-max-send>`.

## q
- **byteFee wire semantics across consumers.** This makes send-max interpret `byteFee` as the *total forced fee* — which is already how the pre-PR resolver used `forceFee: byteFee` for *non-max* sends (those co-sign fine cross-platform), so send-max is now consistent, not divergent. Codex flagged a mixed-version hash-divergence risk; I think it's mitigated by that existing consistency, but can you confirm the iOS/Android initiators populate `cardano.byteFee` with the converged total (not the per-byte `44`) for send-max payloads? The golden fixture's `byte_fee:44` is the one place that ambiguity shows up.

## risk
Low. Strictly-more-correct on the dominant path (guaranteed-rejected → confirmable). Non-max and token sends produce byte-identical signing inputs. The residual edges are network-fail-safe and pre-existing-class (no new fund exposure).

## verdict
**APPROVE.** Core fund-correctness defect is fixed, fee is priced on the true final tx size, and the happy path balances exactly with a valid non-zero fee. The token-bearing and near-empty send-max gaps are real but fail-safe-at-network, no fund loss, and no regression — worth a fast-follow guard, not a merge blocker.

<sub>🤖 reviewer lane · two-lane (Claude + Codex) · HEAD 94b4f840 · full core suite 1852✓ + adversarial plan probes</sub>

neavra (approved):
lgtm — approving. this is the exact `useMaxAmount=false` reframing we recommended when we reproduced the fee=0 bug, and it ships correctly for the common case.

**runtime-reproduced the fix** (real `getCardanoChainSpecific` convergence → `getCardanoSigningInputs` → `AnySigner.plan`, wallet-core 4.7.0, isolated worktree @ 94b4f840):

| scenario | plan |
|---|---|
| clean ADA send-max (3 ADA) | `err=OK amount=2835951 fee=164049 change=0 amount+fee=3000000` ✅ |
| non-max control (1.5 ADA) | `err=OK amount=1500000 fee>0` (recipient gets exactly toAmount) ✅ |
| dust (200k lovelace) | `sendAmount=36039 err=OK` — no underflow (fee < balance) |

the converged fee `164049` is a valid, self-consistent, node-acceptable fee. it's `≠` the device's historical `164515`, which is fine — you correctly don't claim to reproduce the device hash, and the golden note is honest about why (the fixture's `byte_fee:44` is the per-byte-rate field, not the total-fee field, so the offline replay is a raw-replay artifact).

**red-proof (reverted only the production resolver, kept tests):** the new `builds a send-max with a real (non-zero) fee` reds with `expected 0 to be greater than 0` (the fee=0 signature); the `Send ADA - max amount` golden reds. both genuinely pin the fix. the override re-pin `d681b85a`→`36c54307` is the genuine post-fix raw-replay hash (base produces the old hash and reds) — not a mask.

**ran:** target suites `chainSpecific/resolvers/cardano.test.ts` + `mobileFixtures.golden.test.ts` → 76/76; full `yarn test:core` → 1852 passed / 1 skipped / 0 failed; root `tsc` clean; CI green on the merge head. head re-checked unchanged before posting. no code modified.

**non-blocking — ADA send-max from token-bearing UTXOs is still unbroadcastable.** decoded the plan for a max send whose UTXO carries a native token:
```
availableTokens=[TEST:1000]  changeTokens=[TEST:1000]  amount=4832695 fee=167305 change=0
```
`amount+fee=totalInput` leaves 0 ADA for change, but the planner routes the input tokens to a change output — and a token-bearing Cardano output needs min-ADA to fund it, so this can't balance. it **fails closed** (build error / node reject, no broadcast, no lost funds), and it was already broken pre-fix (fee=0), so not a regression — but the branch (`isSendMax = sendMaxAmount && !isTokenSend`) keys off the *destination* being a token, not the *source UTXOs* carrying one, and `getMaxSendAmount` computes on the ADA balance, so it's reachable for any Cardano vault holding native tokens. worth either reserving the token-change min-ADA (`balance − fee − minAdaChange`) or rejecting max-send when input UTXOs carry tokens. the new regression test's token payload pins `sendMaxAmount:false`, so this exact case is untested — a `sendMaxAmount:true` + token-bearing-UTXO case would cover it.

(reproduced both my own plan decode and an independent diff-grounded adversarial pass; same verdict.)

PR #1424, opened 2026-07-20
test(qbtc): bind the mldsa signdoc to a shared cross-encoder golden vector

gomesalexandre:
Refs the round-2 whole-stack audit "QBTC golden vector gap" finding (P2).

## what

#1376 landed shared `testdata/cross-encoder-golden/` fixtures for EVM / Cosmos / Solana, binding the RN-JS and WalletCore encoders to one editable expected hash. **QBTC was left out.**

QBTC signs with post-quantum **MLDSA** keys, so it **cannot** go through WalletCore (secp256k1-only). `packages/core` hand-rolls the Cosmos `SignDoc` via its own varint / length-prefix helpers (`QBTCTx.ts` / `QBTCHelper.ts`). That hand-rolled encoder had **no independent byte-level cross-check** - its golden test only pinned an inline `serialized`-TxRaw literal, never the signing pre-image, and nothing bound it to a canonical protobuf reference.

## how

Add `testdata/cross-encoder-golden/qbtc-msgsend.json` as the single source of truth (input params + `expectedSignDocSha256Hex` + `expectedSerialized`) and bind it from **both** sides:

- **packages/core** `compileTx.golden.test.ts` now reads the fixture and asserts the **pre-signing `SignDoc` hash** (`getPreSigningHashes`) in addition to the serialized TxRaw - so the pre-image the MPC parties actually sign is pinned, not just the assembled tx.
- **packages/sdk** new `qbtc-golden-vectors.test.ts` independently re-encodes the identical `MsgSend -> TxBody -> AuthInfo -> SignDoc` with **cosmjs-types'** canonical protobufjs encoders and asserts the same hash - a genuine cross-encoder bind (hand-rolled vs canonical), same strategy as `cosmos-send-golden-vectors.test.ts`.

The only QBTC-specific wire detail vs a vanilla Cosmos send is the signer pubkey Any typeUrl (`/cosmos.crypto.mldsa.PubKey`). The MLDSA `PubKey` proto is `{ key: bytes }` at field 1 - byte-identical to secp256k1 `PubKey` - so the reference reuses cosmjs-types' `PubKey` encoder and only swaps the Any typeUrl.

Both encoders produce `SignDoc` sha256 `4178ccc8f66d353cecc76f1d36a344952fbdd53360c60cb9a85ab80b59457252` for the golden send. Editing the one fixture updates both suites' expectation at once (no "keep two literals in sync" drift risk).

## testing

- `packages/core` golden suite: **20 passed** (QBTC test now asserts hash + serialized off the shared fixture). Full `yarn test:core`: **1850 passed | 1 skipped**.
- `packages/sdk` `qbtc-golden-vectors.test.ts`: **2 passed** (cosmjs-types reference hash match + full decode round-trip incl. the MLDSA pubkey typeUrl).
- `tsc` clean on both `packages/core` and `packages/sdk`.

## cross-consumer

Test + fixture only - **no shipped code changes**, so no risk to any consumer and no changeset (mirrors #1376, which was also test-only). Pure additional coverage: it locks the existing QBTC signing-hash against a canonical reference.

## risk

None - test-only. It can only catch future regressions in the hand-rolled MLDSA encoder.

🤖 Agent-generated

## Summary by CodeRabbit

* **Tests**
  * Added cross-platform golden-vector coverage for QBTC transaction encoding.
  * Verified sign-document hashes, serialized transaction output, and key decoded fields against shared reference data.
  * Added validation for message details, sender and recipient information, fees, gas limits, and MLDSA public-key encoding.


rcoderdev (approved):

neavra (approved):
lgtm — approving. this genuinely closes the gap #1376 left open (the hand-rolled MLDSA encoder had no independent byte-level cross-check), and it binds honestly. reviewed at 63a30148, re-checked at 314112a8 after you pushed the format fix.

**verified the bind is genuine + mutation-honest** (isolated worktree, restored after each):

| mutation | result | proves |
|---|---|---|
| flip last nibble of `expectedSignDocSha256Hex` | **both** suites red | single shared source of truth |
| mis-number gas_limit field in `QBTCTx.ts` (`protoVarint(2..)`→`(3..)`) | core reds, **sdk green** | core drives the real hand-rolled encoder; the cosmjs-types leg is independent |
| swap MLDSA typeUrl→secp256k1 in the sdk reference | sdk reds | the `/cosmos.crypto.mldsa.PubKey` typeUrl is load-bearing |

and stronger than #1376's evm leg: the hand-rolled varint encoder and cosmjs-types' protobufjs share zero code, so agreeing on the hash is a real independent cross-check (the cosmos/solana quality, not the evm near-self-check). confirmed the fixture's `expectedSerialized` is byte-identical to the old `EXPECTED_QBTC_SERIALIZED` literal — you only ADD the pre-image hash assertion, no assembly change. gas/fee mapping self-consistent both sides (`cosmosSpecific.gas`→fee.amount via QBTCHelper; gas_limit = QBTCTx `defaultGasLimit=300_000n`, which the fixture's `gasLimit` documents).

**ran:** core `compileTx.golden` 20/20; sdk `qbtc-golden-vectors` 2/2; full `yarn test:core` 1850 passed / 0 failed; root `tsc` exit 0. re-verified at 314112a8: `format:check` now clean, both golden suites still green.

**non-blocking:** `cosmosSpecific.gas` actually carries the fee amount (not gas) and gas_limit is the hardcoded `defaultGasLimit` — the core test's `gas: BigInt(fx.feeAmount)` reads oddly but is correct. latent coupling: if `defaultGasLimit` ever changes, the core golden reds (good) but the fixture's `gasLimit` needs bumping too so the sdk reference tracks — a one-line comment in the fixture tying `gasLimit` to `QBTCTx.defaultGasLimit` would save the next editor a puzzle.

testdata/cross-encoder-golden/qbtc-msgsend.json
```diff
@@ -0,0 +1,17 @@
...
+  "_pubKeyNote": "MLDSA public key placeholder (0xaa * 32). The MLDSA PubKey proto is { key: bytes } at field 1 - identical wire shape to secp256k1 PubKey - only the Any typeUrl (/cosmos.crypto.mldsa.PubKey) differs, which is exactly what the two encoders must agree on.",
+  "pubKeyHex": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+  "chainId": "qbtc",
+  "denom": "qbtc",
+  "amount": "123456",
+  "memo": "compileTx golden",
+  "accountNumber": 7,
+  "sequence": 3,
+  "feeAmount": "2500",
+  "gasLimit": 300000,
```
NeOMakinG, line 14: `q:` the core hand-rolled encoder hardcodes `defaultGasLimit = 300_000n` in `QBTCTx.ts` and `buildTxComponents` calls `buildQBTCAuthInfo` **without** passing `gasLimit`, so `fx.gasLimit` here is consumed only by the `packages/sdk` cosmjs reference. It coincidentally equals the hardcoded default, so the cross-bind holds — but the fixture's `_description` claim that editing this file is the single source of truth is slightly overstated for this one field (change `gasLimit` here and only the SDK side moves; the two hashes then diverge and both tests fire, which is confusing rather than a clean re-pin). Not blocking — the safety property is intact and if `defaultGasLimit` ever changes without updating the fixture the tests still catch it. Consider either passing `gasLimit: fx.gasLimit` through in the core test path or noting in `_description` that `gasLimit` mirrors a hardcoded encoder default.

NeOMakinG (approved):
## testing approach

Two-lane (Claude load-bearing + Codex adversarial; GLM retired). Reviewed at HEAD `407ec3b3`. Read the hand-rolled encoder (`QBTCTx.ts`/`QBTCHelper.ts`), ran both suites, and did red-on-revert mutation testing on **both** encoders to confirm the golden is authoritative and non-inert. Codex lane errored out on diff-apply (best-effort, no findings).

**What I verified (this is a signing-correctness test in a fund-critical path, so I checked it's meaningful, not just green):**

- **Golden is AUTHORITATIVE, not self-referential.** The sha256 `4178ccc8…59457252` is independently reproduced by **two encoders that share no code**: (a) `packages/sdk` re-encodes `MsgSend→TxBody→AuthInfo→SignDoc` with **cosmjs-types'** canonical protobufjs `.encode()` (the cosmos-sdk wire reference), and (b) `packages/core`'s hand-rolled `QBTCTx.ts` asserts `hex(hash) === fx.expectedSignDocSha256Hex`. A golden captured from the code-under-test would only catch drift; this one is cross-validated against a canonical reference. ✅
- **Genuine cross-encoder bind.** Mutation A (`defaultGasLimit 300_000n→300_001n` in `QBTCTx.ts`) moved **only** the core hash to `8d077b…` while the cosmjs-side golden stayed `4178ccc8…` → the two encoders are truly independent, not secretly the same code.
- **Full signed pre-image covered.** SignDoc = bodyBytes (`MsgSend` from/to/amount/denom + memo) + authInfoBytes (MLDSA pubkey typeUrl+key, mode=1 DIRECT, sequence, fee amount+denom, gasLimit) + chainId + accountNumber. Every signed field is inside the hash.
- **Non-inert both directions.** Mutation A → core test red. Mutation B (SDK reference pubkey typeUrl `mldsa→secp256k1`) → SDK hash `90e375d7…` ≠ golden → red, and the decode round-trip also asserts the `/cosmos.crypto.mldsa.PubKey` typeUrl. So the MLDSA-specific wire detail is genuinely bound, not decorative.
- **No production code touched.** Test + fixture only; MLDSA `PubKey` reuse of cosmjs-types' secp256k1 `PubKey` encoder (proto `{ key: bytes }` field 1, only the Any typeUrl swapped) is correct for cosmos MLDSA. Pure additional coverage — locks the existing QBTC signing hash against a canonical reference.

Runs: SDK `qbtc-golden-vectors.test.ts` 2 passed; core QBTC golden 1 passed.

## q

One non-blocking observation on fixture ergonomics (see inline).

## risk

None shipped — test + fixture only. Strictly raises the floor: it makes any future drift in the hand-rolled MLDSA SignDoc encoder fail against a canonical protobuf reference.

## verdict

**APPROVE.** The golden vector is authoritative (cross-validated against cosmjs-types, not self-captured), the cross-encoder bind is real (two independent encoders), the full signed SignDoc is covered, and both sides are red-on-revert. Exactly what a signing-path test PR should be.

---
🤖 Reviewed with Claude Code (Opus 4.8) — Claude load-bearing + Codex (errored, no findings).

PR #1425, opened 2026-07-20
fix(swap): bind erc-20 approval spender to the verified router on the signing path

gomesalexandre:
Stacked on #1377 (base = its branch; GitHub retargets to `main` when #1377 merges). Closes the ERC-20 approval-spender follow-up neavra requested in the #1377 review.

## what

#1377 re-asserts the aggregator router allow-list on the co-signer signing path, but its own SCOPE comment (and neavra's review) flag that it covers the **swap-leg destination only** — `quote.tx.to`. The **ERC-20 approval spender** is a SEPARATE, independent wire field (`erc20ApprovePayload.spender`) that the approve resolver (`erc20.ts`) reads verbatim, and nothing bound it to `quote.tx.to`.

neavra red-proofed the gap on #1377: a payload with `provider:'1inch'` + a genuine 1inch router as `quote.tx.to` (so the router guard **passes**) + `erc20ApprovePayload.spender = 0x…dEaD` → **no throw**; `inputs[0]` is an ERC-20 approve granting the attacker an allowance over the user's token, while `inputs[1]` still targets the real router so nothing else looks wrong. A classic approval-drain the co-signer signs blind.

## how

Bind `spender === quote.tx.to` for enforced providers (1inch/kyber), via `assertEnforcedSwapApprovalSpenderBound`, called in the **approve branch** of `getEvmSigningInputs` — the one place `erc20ApprovePayload` is still present (the recursion strips it before #1377's router guard runs on the swap leg). Combined with #1377's `tx.to` allow-list, the spender is provably the canonical router.

- **Unenforced providers stay unbound** — notably **cowswap**, whose spender is legitimately the `GPv2VaultRelayer`, not `tx.to`. Gating on the enforced set skips it (same reason its router isn't allow-listed).
- **Native (thorchain/maya) swaps** are `'native'`, not `'general'` → skipped (they have their own vault validation, e.g. the native-swap broadcast guard).
- **Plain token approves** (no general swap payload) → skipped.
- **Monotonic gate**: throws or no-ops, never mutates signed bytes → cannot desync the cross-device pre-signing hash. Same `provider`-is-a-free-string threat model as #1377: spoofing to an unenforced value degrades to the pre-fix state, not a new bypass.

`@vultisig/core-chain` + `@vultisig/core-mpc` ship to the flagship ios/android/windows clients. This adds a **new fail-closed throw** for an enforced-provider swap whose approve spender ≠ the router. On the initiator, `build.ts` sets the approve spender to `getSwapDestinationAddress` === `tx.to` by construction, so the two are always equal for a legitimately-produced swap → **no false-block**. Only a hand-built/tampered payload trips it. Strictly-more-correct.

## testing

- +2 cases in #1377's `swapRouterGuard.test.ts`: valid router + attacker spender → rejects (`approval spender … does not match the verified swap router`); spender === router → signs cleanly (approve + swap legs, 2 inputs).
- Full `yarn test:core`: **1857 passed | 1 skipped**. `tsc` + eslint clean. Changeset: `core-chain` + `core-mpc` patch.
- Opus adversarial self-review (codex capped til Jul 25): enumerated relabel-bypass (monotonic, not new), cowswap/native/plain-approve skips, empty-tx.to fail-closed, address casing.

## risk

Low, fund-safety-positive. Additive fail-closed bind on a field already on the payload; no wire/encoder change; no false-block for any initiator-produced swap.

🤖 Agent-generated

### DIFF
--- a/.changeset/swap-approval-spender-bind.md
+++ b/.changeset/swap-approval-spender-bind.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+---
+
+Bind the ERC-20 approval spender to the verified swap router on the co-signer signing-input path for enforced aggregator providers (1inch/kyber). Follow-up to the signing-path router guard: `quote.tx.to` was re-asserted, but `erc20ApprovePayload.spender` is a separate wire field the approve resolver reads verbatim, so a payload could pass the router check yet still carry an approve granting an attacker an allowance (approval-drain). The bind runs in the approve branch (where the field is still present) and requires `spender === quote.tx.to` for enforced providers; unenforced providers stay unbound (notably cowswap, whose spender is legitimately the GPv2VaultRelayer, not `tx.to`). Monotonic gate: throws or no-ops, never mutates signed bytes. Initiators set the two equal by construction, so only a hand-built/tampered payload trips it.
--- a/packages/core/chain/swap/general/knownAggregatorRouters.ts
+++ b/packages/core/chain/swap/general/knownAggregatorRouters.ts
@@ -101,6 +101,37 @@ export function assertKnownAggregatorRouterOnSigningPath(provider: string, addre
 }
 
 /**
+ * Signing-path approval-spender bind (sdk#1358 review follow-up, requested by neavra). The
+ * follow-on to {@link assertKnownAggregatorRouterOnSigningPath}: that guard validates the swap-leg
+ * destination (`quote.tx.to`), but a general swap that needs an allowance also carries a SEPARATE,
+ * independent wire field - `erc20ApprovePayload.spender` - which the approve resolver (erc20.ts)
+ * reads verbatim and nothing binds to `quote.tx.to`. So a payload can pass the router check with a
+ * genuine 1inch/kyber `tx.to` yet still carry an approve granting an ATTACKER an allowance over the
+ * user's token (a classic approval-drain the co-signer would otherwise sign blind).
+ *
+ * On the initiator these coincide by construction (build.ts sets the approve spender to
+ * `getSwapDestinationAddress` === `tx.to`), so this only ever fires on a hand-built/tampered payload.
+ * Enforced providers (1inch/kyber) MUST have `spender === routerDestination`; unenforced providers are
+ * NOT bound - notably cowswap, whose spender is legitimately the GPv2VaultRelayer, not `tx.to`. Like
+ * its sibling this is a MONOTONIC gate: it only throws or no-ops, never changes the signed bytes.
+ */
+export function assertEnforcedSwapApprovalSpenderBound(
+  provider: string,
+  spender: string,
+  routerDestination: string,
+  chain: Chain
+): void {
+  if (!ENFORCED_ROUTER_PROVIDERS.has(provider)) {
+    return
+  }
+  if (spender.toLowerCase() !== routerDestination.toLowerCase()) {
+    throw new Error(
+      `${provider} swap approval spender (${spender}) does not match the verified swap router (${routerDestination}) on ${chain} — refusing to sign an approval to an unbound spender.`
+    )
+  }
+}
+
+/**
  * Throws if `address` isn't the known router for `provider` on `chain`. Call this at quote
  * construction, before a GeneralSwapQuote carrying `address` as `tx.evm.to` can exist.
  */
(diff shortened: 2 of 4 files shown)


rcoderdev (approved):

NeOMakinG (approved):
**APPROVE** — the approval-drain vector neavra red-proofed on #1377 is provably closed for the enforced aggregator providers. Reviewed at HEAD `062ee43c`. Two-lane (Claude load-bearing; Codex lane timed out at 10m — best-effort, skipped per the codex-capped note in the PR body).

## testing approach

- Traced the full co-signer signing path in `getEvmSigningInputs` end-to-end against source, not just the diff.
- Built shared packages (`yarn build:shared`) and ran the targeted `swapRouterGuard.test.ts` (4/4 pass) and the full `yarn test:core` suite (**1857 passed | 1 skipped**, matches the PR body).
- **Red-on-revert:** neutered the bind (`if (false && spender.toLowerCase() !== …)`) → the attacker-spender case goes RED (1 failed | 3 passed). The guard is non-inert and operative, not decorative.

## binding integrity (the fund-safety question)

The bind is **provably correct** for enforced providers (1inch/kyber). The proof is compositional across the approve branch and its recursion:

1. `assertEnforcedSwapApprovalSpenderBound` throws unless `spender === quote.tx.to` (approve leg). It guards the *exact* field the builder signs — `getErc20ApproveSigningInput` reads `keysignPayload.erc20ApprovePayload.spender` verbatim (`erc20.ts:19`), so the guard and the signed byte are the same field. Confirmed non-inert.
2. The approve `SigningInput` is *built* before the recursion, but the function does not `return [approveSigningInput, ...restOfSigningInputs]` until the recursion resolves — and the recursion (erc20ApprovePayload stripped, swapPayload retained) always reaches `assertKnownAggregatorRouterOnSigningPath(provider, quote.tx.to)`, which throws unless `quote.tx.to` is a known router. So a returned/signable approve leg guarantees **spender === quote.tx.to === known-router**. There is no early-return path that signs the approve leg alone.

**"Verified" router provenance:** the allowlist is a set of hardcoded per-chain constants (`ONE_INCH_V6_STANDARD_ROUTER`, the zkSync variant, `KYBER_STANDARD_ROUTER`) — NOT read from the attacker-influenceable quote response. So the router the spender binds to cannot itself be steered to a malicious address. The only attacker-controlled input is the `provider` *string*; relabeling it to an unenforced value degrades to the pre-fix state (monotonic gate), not a new bypass — the same accepted threat model as #1377.

**No false-block:** on the initiator, `build.ts` sets `erc20ApprovePayload.spender = keysignPayload.toAddress = getSwapDestinationAddress(...) = quote.tx.to` for evm general swaps (build.ts:369+158, getSwapDestinationAddress evm arm). The two are equal by construction, so every legitimately-produced 1inch/kyber swap passes. Only a hand-built/tampered payload trips it. Strictly-more-correct for the core-chain + core-mpc consumers (ios/android/windows).

## q

- **native (thorchain/maya) ERC-20 approve spender is left unbound.** A native EVM ERC-20 swap also attaches an `erc20ApprovePayload` (build.ts:378, spender = router/inbound addr), and on the signing path `getKeysignSwapPayload` returns `{ native }`, so `'general' in approveSwapPayload` is false → this guard skips it. The identical approval-drain shape exists there (tampered native payload → attacker spender), and the native swap-leg destination itself (`shouldBePresent(routerAddress)`) isn't allow-listed either. This is pre-existing and explicitly out of this PR's 1inch/kyber scope (same boundary as #1377), so **not blocking** — but is the native approve-spender / router bind tracked as a follow-up bead? The "own vault validation" referenced in the PR body validates the native swap-leg vault, not the approve spender.
- amount stays unbounded (uses `erc20ApprovePayload.amount` verbatim, unchanged here). Since the spender is now bound to the real router, an unlimited amount would go to the legit router, so blast radius is contained — noting only, not in scope.

## risk

Low, fund-safety-positive. Additive fail-closed bind on a field already on the payload; no wire/encoder/byte change (monotonic — throws or no-ops, cannot desync the cross-device pre-signing hash); no false-block for any initiator-produced swap. core-chain + core-mpc patch changeset present.

## verdict

**APPROVE.** The enforced-provider (1inch/kyber) approval spender is provably bound to the verified router across the signing path; guard is non-inert (red-on-revert), full core suite green, no false-block. Native/unenforced coverage is a documented, pre-existing scope boundary tracked separately.

🤖 Reviewed by Claude Opus 4.8 (reviewer lane) — two-lane (Codex timed out, best-effort).

PR #1426, opened 2026-07-20
fix(sdk): raw cosmos broadcast reports false success on delivertx failure

gomesalexandre:
Found in the R2 SDK fund-safety audit (broadcast false-success class).

## what

`vault.broadcastRawTx({ chain, rawTx })` → `RawBroadcastService.broadcastCosmosRawTx` reported a **false success** for a Cosmos transaction that was included in a block but **failed execution**.

## why

`StargateClient.broadcastTx` **resolves** (does not throw) once a tx is included, even when DeliverTx failed (`code !== 0` — out-of-gas, wasm revert, a THORChain/Maya deposit-handler rejection). The raw path returned `result.transactionHash` whenever the client didn't throw (`RawBroadcastService.ts:275`), so a reverted cosmos tx came back as a success hash — the tx is on-chain but nothing moved, and the caller (agent / SDK consumer) is told it succeeded.

The signing-input broadcast resolver (`tx/broadcast/resolvers/cosmos.ts:32-42`) already asserts DeliverTx success — that's the **#1316** fix. But this raw path **reimplements** broadcast without it. Same class the R2 audit found still open on the QBTC send resolver (separate PR incoming).

## how

`assertIsDeliverTxSuccess(result)` before returning the hash; throw `VaultError(BroadcastFailed)` on execution failure. `assertIsDeliverTxSuccess` uses `!!result.code`, so a successful tx (`code: 0` or absent) is unaffected — only a genuine `code !== 0` inclusion now fails closed.

## testing

- New regression test: a `code: 5` (out-of-gas) included tx now throws `BroadcastFailed` ("execution failed") instead of returning `reverted-hash`. Existing cosmos raw-broadcast tests unchanged (their mocks have no `code`, i.e. success).
- Full sdk unit suite green; `tsc` + eslint clean. Changeset: `@vultisig/sdk` patch.

## cross-consumer

`@vultisig/sdk` ships to flagship clients. Strictly-more-correct fail-closed: turns a silent broadcast-of-a-reverted-tx into a clear `BroadcastFailed`. No behavior change for a successful tx.

## risk

Low, fund-safety-positive. One added assertion on the raw cosmos path; the signing-input path already behaves this way.

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Cosmos raw transaction broadcasts now report failure when on-chain execution fails, rather than incorrectly indicating success.
  * Failed broadcasts provide a clear execution-failed error and no longer return a transaction hash as a successful result.

* **Tests**
  * Added coverage to verify that reverted Cosmos transactions are rejected correctly.

### DIFF
--- a/.changeset/rawbroadcast-cosmos-delivertx-false-success.md
+++ b/.changeset/rawbroadcast-cosmos-delivertx-false-success.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Fix `vault.broadcastRawTx` reporting a false success for a Cosmos transaction that was included but failed execution. `RawBroadcastService.broadcastCosmosRawTx` returned `result.transactionHash` whenever `StargateClient.broadcastTx` didn't throw, but that client resolves (does not throw) once a tx is included even when DeliverTx failed (`code !== 0` — out-of-gas, wasm revert, a THORChain/Maya deposit-handler rejection). The tx is on-chain but nothing moved, so the raw path now asserts DeliverTx success and throws `BroadcastFailed` on execution failure, matching the signing-input broadcast resolver (#1316).
--- a/packages/sdk/src/vault/services/RawBroadcastService.ts
+++ b/packages/sdk/src/vault/services/RawBroadcastService.ts
@@ -1,3 +1,4 @@
+import { assertIsDeliverTxSuccess } from '@cosmjs/stargate'
 import { Chain, CosmosChain, EvmChain, OtherChain, UtxoBasedChain } from '@vultisig/core-chain/Chain'
 import { isChainOfKind } from '@vultisig/core-chain/ChainKind'
 import { bittensorRpcUrl } from '@vultisig/core-chain/chains/bittensor/client'
@@ -272,6 +273,24 @@ export class RawBroadcastService {
     }
 
     if (!result) throw new Error('No broadcast result returned')
+
+    // `StargateClient.broadcastTx` RESOLVES (does not throw) once the tx is included in a block,
+    // even when execution failed (DeliverTx `code !== 0` — out-of-gas, wasm revert, a THORChain/Maya
+    // deposit-handler rejection). The tx is on-chain but nothing moved, so returning its hash here
+    // would be a false success. The signing-input broadcast resolver (tx/broadcast/resolvers/cosmos.ts,
+    // #1316) already asserts this; this raw path — reachable via the public `vault.broadcastRawTx` —
+    // must fail closed the same way.
+    try {
+      assertIsDeliverTxSuccess(result)
+    } catch (deliverTxError) {
+      const message = deliverTxError instanceof Error ? deliverTxError.message : String(deliverTxError)
+      throw new VaultError(
+        VaultErrorCode.BroadcastFailed,
+        `Cosmos transaction was included but execution failed: ${message}`,
+        deliverTxError instanceof Error ? deliverTxError : new Error(message)
+      )
+    }
+
     return result.transactionHash
   }
 
--- a/packages/sdk/tests/unit/vault/services/RawBroadcastService.test.ts
+++ b/packages/sdk/tests/unit/vault/services/RawBroadcastService.test.ts
@@ -165,6 +165,27 @@ describe('RawBroadcastService', () => {
     expect(mockGetCosmosClient).toHaveBeenCalledWith(Chain.Osmosis)
   })
 
+  // Fund-safety: StargateClient.broadcastTx RESOLVES (does not throw) on a tx that was included but
+  // failed execution (DeliverTx code !== 0). The raw path must not report that as a success hash.
+  it('throws instead of returning a hash when the Cosmos tx is included but DeliverTx-fails', async () => {
+    mockCosmosBroadcastTx.mockResolvedValueOnce({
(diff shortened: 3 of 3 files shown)


neavra (approved):
lgtm — approving. this is the RawBroadcastService third door: `vault.broadcastRawTx` → `broadcastCosmosRawTx` returned `result.transactionHash` whenever `StargateClient.broadcastTx` didn't throw, but that client resolves on block inclusion even when DeliverTx `code !== 0`, so an on-chain-but-reverted cosmos tx (out-of-gas, wasm revert, a thor/maya deposit-handler rejection) came back as a success hash. same class as #1316, now closed on the raw path too. reaching for cosmjs's own `assertIsDeliverTxSuccess` is the right call — inherits the upstream hash+height+code+rawLog message and can't drift from cosmjs's notion of failure.

verified:
- **truthful both directions** — `assertIsDeliverTxSuccess` throws only on `code !== 0` (installed cosmjs 0.39.0: `isDeliverTxFailure = !!result.code`), so no false-*failure* on a genuinely-committed tx. cosmos txs are atomic — `code != 0` ⇒ state reverted, gas burned, sequence consumed.
- **no transient-retry interlock to worry about here** (the thing that bit the first cut of #1316). this raw path calls `client.broadcastTx` directly through `attempt(...)`; `withTransientBroadcastRetry` / `isTransientBroadcastError` don't reach `packages/sdk/src` at all — nothing rebroadcasts, so there's no loop to reclassify the throw as transient and swallow a subsequent `tx already exists in cache` as success. a plain `VaultError(BroadcastFailed)` is correct on this path; the `DeliverTxFailedError` marker type was only needed on the resolver path that runs *inside* the retry wrapper.
- **red-proof**: reverted just the assert hunk, kept the test → the new test fails with `promise resolved "'reverted-hash'" instead of rejecting` = the exact false-success signature. restored; other 17 in the file green.
- **consumers**: `VaultBase.broadcastRawTx` emits the error event then re-throws; the cli `broadcast` command surfaces it as a failure and won't print a fake `TX Hash:` / explorer line for a reverted tx. no caller swallows the throw.
- sdk unit suite green (191 files / 2895 tests); `@vultisig/sdk typecheck` clean; changeset present (`@vultisig/sdk: patch`, correct scope — sdk-only, no core-chain touched).

non-blocking, not for this pr:
- the throw discards the on-chain hash we definitively have (same shape as the M2 note on #1316) — the cosmjs message embeds it so it's unstructured rather than lost. fine to leave.
- the pre-existing `error`-arm mapping of `tx already exists in cache` / `account sequence mismatch` → `BroadcastFailed "may have already been submitted"` is untouched and fine here (nothing rebroadcasts on this path).

reviewed at 80a721bc.

NeOMakinG (approved):
**APPROVE** — verified from source at HEAD `80a721bc`. Closes the raw-cosmos leg of the broadcast false-success class; this is the exact `RawBroadcastService.broadcastCosmosRawTx` door I filed as still-open when I reviewed sibling #1316.

## testing approach
Two-lane (Claude load-bearing). Read the full `broadcastCosmosRawTx` body against the installed `@cosmjs/stargate@0.39.0` build, ran the unit file, and did a red-on-revert.

## blocking
None.

## should-fix
None.

## q
- **Mirror the pattern to the siblings.** #1429 (qbtc), #1430 (sui moveAbort), #1432 (bittensor) are the same false-success class on other raw-broadcast legs. The shape here is the right one: read the resolved execution result, `assert*Success`, wrap the throw in `VaultError(BroadcastFailed)` preserving the underlying message, and keep the CheckTx/RPC-reject failure on the pre-existing `error` arm. Worth confirming each sibling checks the execution-stage code (not just the mempool/RPC-ack stage).

## risk
Low, fund-safety-positive. One added assertion on the resolved result; strictly-more-correct fail-closed. No behavior change for a successful tx.

## verdict
**APPROVE.** Correctness verified end-to-end:

- **Right failure surfaces, both covered.** `getCosmosClient` returns a `StargateClient` (`core/chain/chains/cosmos/client.ts:13`). Its `broadcastTx` does `broadcastTxSync` first — a CheckTx/mempool reject throws `BroadcastTxError`, caught on the pre-existing `error` arm — then polls and RESOLVES a `DeliverTxResponse` carrying the real **execution** `code`. The new `assertIsDeliverTxSuccess(result)` reads that execution-stage code, so the fix guards the DeliverTx surface the old code missed, not the CheckTx ack.
- **Success is strictly fail-closed.** `isDeliverTxFailure = !!result.code` (stargateclient.js:28) → throws on any non-zero code; only `code === 0`/absent passes. A malformed/missing code cannot default to success in a way that returns a reverted hash.
- **txhash still surfaced.** On failure the thrown message is cosmjs's `Error when broadcasting tx ${transactionHash} at height ${height}. Code: ${code}; Raw log: ${rawLog}` — the hash is carried into the `VaultError` for lookup, reported as FAILED rather than success. Matches the #1316 resolver pattern.
- **No regression.** `code === 0` → assert passes → returns `result.transactionHash`. Existing raw-cosmos tests (no `code` in mocks) unchanged.
- **Test is load-bearing.** New `code: 5` out-of-gas test asserts `BroadcastFailed` + "execution failed". Red-on-revert: commenting out the assert flips it to 1 failed / 17 passed. Full file 18/18 green.

PR #1427, opened 2026-07-20
fix(swap): guard the cowswap receiver at eip-712 digest construction

gomesalexandre:
Found in the R2 SDK fund-safety audit — the same class as #1377/#1425 (a guard that runs at quote-time but not on the co-signer signing path), applied to CowSwap.

## what

`assertValidCustomRecipient` (`findSwapQuote.ts:532`) rejects a zero/burn/malformed CowSwap receiver at **quote construction**. But an MPC co-signer never sees the quote — it decodes a `cowswap-order:` blob from the `KeysignPayload` (`decodeCowSwapKeysignData`, which validates **shape only**: `chainId`/`apiBase`/`from`/`order` types) and builds the EIP-712 order digest to sign. So a hand-built payload with `receiver = 0x…dead` (or the zero address, or a malformed address) would be signed into a GPv2 order that sends the buy tokens to an unrecoverable address — with no on-chain recovery.

## how

Guard `order.receiver` in `buildCowSwapOrderTypedData` — the digest-construction choke point every signer (initiator AND co-signer) funnels through before signing — rejecting zero, burn (`0x…dead`), and malformed-non-EVM receivers, matching the quote-time guard's constants.

**No false-block**: vultisig always sets an explicit receiver (`normalizedRecipient ?? from.address`, `findSwapQuote.ts:716`) — for a self-swap it's the vault's own EVM address, for a custom recipient it's the quote-validated address. It **never** uses CowSwap's `address(0)` "pay-the-owner" sentinel, so rejecting zero here matches the quote-time behavior and can't block a legitimately-produced order.

`@vultisig/core-chain` ships to flagship ios/android/windows. This adds a **new fail-closed throw** at digest construction for a zero/burn/malformed CowSwap receiver. Since vultisig's SDK path always sets an explicit non-zero receiver, no legitimately-produced order is affected — only a hand-built/tampered one. Strictly-more-correct; no digest change for a valid receiver (the message fields are untouched).

## testing

- +4 cases: burn / zero / malformed receiver → throws; a normal explicit receiver still builds. Existing digest/domain/type-hash tests unchanged (they use a valid receiver).
- Full cowswap suite **74 passed**; core suite green; `tsc` + eslint clean. Changeset: `core-chain` patch.
- Opus adversarial self-review (codex capped): verified the zero-sentinel question (vultisig never uses it), the digest builder is the choke point both initiator and co-signer hit, and CowSwap is EVM-only so the EVM-format check is always applicable.

## risk

Low, fund-safety-positive. Fail-closed guard on a field already in the signed order; no digest change for any valid receiver.

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safety checks when signing CowSwap orders to reject zero addresses, burn addresses, and malformed receiver addresses.
  * Prevents orders from sending tokens to unrecoverable or invalid destinations.
  * Valid orders with explicit, valid receivers continue to be accepted.


rcoderdev (approved):

packages/core/chain/swap/general/cowswap/sign/buildCowSwapOrderTypedData.ts
```diff
@@ -83,28 +83,65 @@ type BuildCowSwapOrderTypedDataInput = {
...
-    appData: order.appDataHash,
-    feeAmount: order.feeAmount,
-    kind: order.kind,
-    partiallyFillable: order.partiallyFillable,
-    sellTokenBalance: order.sellTokenBalance,
-    buyTokenBalance: order.buyTokenBalance,
-  },
-})
+}: BuildCowSwapOrderTypedDataInput): CowSwapOrderTypedData => {
+  assertCowSwapReceiverSafe(order.receiver)
```
NeOMakinG, line 123: `praise:` the checked value `order.receiver` is the same value placed into `message.receiver` two lines down with no mutation between — so the guard binds the exact bytes hashed into the signed EIP-712 digest, not a divergeable pre-image copy. Red-on-revert confirms it: removing this line fails the 3 new throw-tests. This is the correct choke point (sole EIP-712 builder, public export the downstream co-signers import).

packages/core/chain/swap/general/cowswap/sign/buildCowSwapOrderTypedData.ts
```diff
@@ -83,28 +83,65 @@ type BuildCowSwapOrderTypedDataInput = {
...
+ * (decodeCowSwapKeysignData, which validates SHAPE only) and builds THIS EIP-712 digest to sign. A
+ * hand-built payload with `receiver = 0x…dead` (or the zero address, or a malformed address) would
+ * otherwise be signed into an unrecoverable order. This is the digest-construction choke point every
+ * signer funnels through, so guarding here binds the receiver the signature actually covers.
+ *
+ * vultisig always sets an explicit receiver (`normalizedRecipient ?? from.address`,
+ * findSwapQuote.ts) — never CowSwap's `address(0)` "pay-the-owner" sentinel — so rejecting zero here
+ * matches the quote-time guard and cannot false-block a legitimately-produced order.
+ */
+const assertCowSwapReceiverSafe = (receiver: string): void => {
```
NeOMakinG, line 105: `q:` this rejects the *unrecoverable* sinks (zero/burn/malformed) but a valid arbitrary EVM address still passes, so a hand-built payload could still route the fill to an attacker's own valid address. Not blocking and not a regression — it's identical to the quote-time residual (findSwapQuote.ts:528: "a well-formed but merely WRONG non-burn recipient is the caller's grounding responsibility") and isn't closable here: `from` (the vault address) isn't passed into this fn, and binding `receiver === from` would false-block the supported custom-recipient feature (vultisig-windows#4131). Fully closing the misroute-to-valid-address vector needs an attested intended-receiver in the KeysignPayload — worth a follow-up issue so it isn't assumed closed by this guard.

NeOMakinG (approved):
Full first-pass review of the signing-side CowSwap receiver guard. **vultisig-sdk is fund-critical**, so I traced the receiver from the EIP-712 message field back through the keysign payload and confirmed the guard binds the exact bytes that get signed.

## testing approach

- Worktree at HEAD `3691f2bf`, `yarn@4.16.0` + `vitest run --config .config/vitest.core.config.ts`.
- Ran the full cowswap suite: **74 passed**.
- **Red-on-revert**: commented out `assertCowSwapReceiverSafe(order.receiver)` → the 3 new throw-tests (burn / zero / malformed) fail and the "normal receiver" test still passes. The guard is **non-inert** and covers exactly the signed field.
- Traced `buildCowSwapOrderTypedData` callers: no in-SDK production caller (tests only), but it **is a public export** (`./swap/general/cowswap/sign/buildCowSwapOrderTypedData`) consumed by the downstream co-signers (ios/android/windows). Grepped for any hand-rolled `hashTypedData`/`TypedDataEncoder` CowSwap digest path — none. This is the **sole** EIP-712 builder; `submitCowSwapOrder` only POSTs the already-signed order.

## verified

- **Digest-level binding (the thing that matters):** the guard checks `order.receiver`, and the *same value* flows into `message.receiver` with no mutation in between — so the value validated is exactly the value hashed into the signed digest. Not a pre-image copy that can diverge.
- **Zero-sentinel handled correctly:** CowSwap treats `receiver = address(0)` as "pay-the-owner". vultisig never emits it — `build.ts` carries `order.receiver` from `findSwapQuote:716` (`normalizedRecipient ?? from.address`), always explicit — so rejecting zero can't false-block a legitimately-produced order, and it removes an ambiguous sentinel from any hand-built payload. Correct.
- **Fail-closed:** `assertCowSwapReceiverSafe` throws before the message object is constructed. Mismatch → no digest → no signature.
- **Faithful mirror of quote-time:** the constants and semantics match `assertValidCustomRecipient` (`findSwapQuote.ts:532`) exactly, extending the same fail-closed posture from quote construction to the co-signer path (the sdk#1377/#1425 class of "guard runs at quote-time but not on the signing path").

## q

- The guard rejects the **unrecoverable** sinks (zero / burn / malformed) but a *valid* arbitrary EVM address still passes — so a hand-built payload could still route the fill to an attacker's own valid address. This is **not blocking** and **not a regression**: it is identical to the quote-time residual (`findSwapQuote.ts:528` explicitly: "a well-formed but merely WRONG non-burn recipient is the caller's grounding responsibility"), and it is not closable *at this layer* — the payload's `from` (vault address) isn't passed into `buildCowSwapOrderTypedData`, and even if it were, binding `receiver === from` would false-block the supported custom-recipient feature (route-gated `recipient`, tracked in vultisig-windows#4131). Fully closing the misroute-to-valid-address vector needs an *attested intended-receiver* channel in the KeysignPayload — a separate design item worth a follow-up issue. Flagging so it isn't assumed closed by this PR.

## risk

Low, fund-safety-positive. A new fail-closed throw on a field already in the signed order; no digest change for any valid receiver (message fields untouched). Public-export contract goes from total to partial (throws on zero/burn/malformed) — but only a tampered/hand-built order hits it, since vultisig always sets an explicit valid receiver.

## verdict

**APPROVE.** The guard sits at the sole EIP-712 digest choke point, is proven non-inert (red-on-revert), binds the exact signed field, and handles the zero-sentinel correctly with no false-block. It fully delivers its stated scope (blocking unrecoverable receivers at the co-signer). The residual valid-address divergence is pre-existing, mirrors quote-time, and is unclosable here without breaking custom recipients — logged as a `q:` follow-up, not a hold.

— reviewed by NeOMakinG (Claude primary lane; Codex second-opinion lane stalled/silent, best-effort per protocol)

neavra (approved):
approve. traced the receiver from the eip-712 message field back to the guard and red-proofed at the current head (`6ed97fe2`).

**what it does** — re-checks `order.receiver` at `buildCowSwapOrderTypedData` (the eip-712 digest construction choke point), rejecting zero / burn (`0x…dead`) / malformed-non-EVM receivers, mirroring the quote-time `assertValidCustomRecipient`. co-signers never see the quote (they decode a `cowswap-order:` blob, shape-validated only), so this extends the fail-closed posture to the signing side — same class as the #1377/#1425 signing-path guards.

- digest binding: the guard checks `order.receiver` and the *same* value flows into `message.receiver` unmutated — it binds the exact bytes hashed into the signed digest, not a copy that can diverge.
- sole choke point: within the sdk no production caller builds the order digest itself — `build.ts` encodes the `cowswap-order:` marker and the consumer drives order signing through this public export; `submitCowSwapOrder` only POSTs an already-signed order. every signer funnels through the guarded function.
- faithful mirror: signing guard uses the *identical* constants (`0x0000…0000`, `0x0000…dead`) and `isEvmAddress` = `/^0x[0-9a-fA-F]{40}$/i` as quote-time — no stricter/looser drift, so it can't false-block a legit order.
- zero-sentinel: cowswap treats `receiver = address(0)` as "pay-the-owner"; vultisig always sets an explicit receiver (`normalizedRecipient ?? from.address`), never the sentinel, so rejecting zero is safe.

**receipts** (head `6ed97fe2`)
- full cowswap suite: **74 passed**.
- red-proof: neutered `assertCowSwapReceiverSafe(order.receiver)` → the 3 throw-tests (burn/zero/malformed) go red, "normal receiver still builds" stays green. mutant killed, tree restored.
- core `tsc` (`.config/tsconfig.json`) clean. changeset present (core-chain patch).
- the 04:28 main-merge into this branch is a clean no-conflict merge (only pulls in the rawbroadcast cosmos-delivertx fix); net PR diff is exactly the guard — NeOMakinG/rcoderdev reviewed the pre-merge `3691f2bf`, this re-confirms at the merged head.

**non-blocking**
- a *valid-but-wrong* attacker EVM address still passes (only unrecoverable sinks are rejected). pre-existing, mirrors the quote-time residual, and not closable at this layer without an attested intended-receiver in the KeysignPayload (`from` isn't passed into `buildCowSwapOrderTypedData`, and binding `receiver === from` would break the supported custom-recipient feature). worth a follow-up issue, not a hold.

lgtm.

PR #1428, opened 2026-07-20
fix(rn): honor lifi slippage tiers + consumer override (mobile-wide)

gomesalexandre:
Round-7 audit P1 (E2), mobile-wide. **Verified still open on current main.**

## what

The React Native override for `getLifiSwapQuote` — the module rollup redirects core's `getLifiSwapQuote.ts` to on RN (`rollup.platforms.config.js:118`) — applied a **flat 1% slippage** and had no `slippage`/`ticker` `Input` field, so it **discarded** the `lifiSlippageFraction` the shared caller (`findSwapQuote.ts:825`) threads through.

Core `getLifiSwapQuote` resolves two tiers (**0.3% stable / 1% volatile**, sdk#524) and honors an explicit override. So on RN — **most users** — every LiFi quote was flat 1%:
- a stable pair (USDC↔USDT) that gets 0.3% on web/desktop got 1% on mobile — a wider MEV / slippage-loss surface on a tight-peg swap;
- an explicit tight-tolerance request was **silently ignored** — LiFi bakes the `minAmountOut` floor from the slippage, so the user's requested protection never reached the prebuilt tx.

## how

Extract the tier + override + combined-cost-ceiling logic into a shared `lifi/api/lifiSlippage.ts` (`resolveLifiSlippage`). The core path and the RN override both import it, so they resolve slippage **identically, with no drift**. The override couldn't just import from core's `getLifiSwapQuote` (that path redirects back to itself — a loop), which is exactly why the logic was duplicated-and-drifted; a non-redirected sibling module fixes that structurally. Added `slippage` + `ticker` to the override `Input` (the caller already passes both).

## testing

- +3 RN-override cases (`rnGetLifiSwapQuoteOverride.test.ts`): stable pair → 0.3%, volatile pair → 1%, explicit override honored (no longer dropped). The base slippage suite (`getLifiSwapQuote.slippage.test.ts`) re-imports the shared `STABLE_TICKERS`/`isStablePair` via a re-export — **9 passed**, unchanged.
- Full `yarn test:core` **1850 passed**; core + sdk `tsc`, eslint, prettier all clean. Changeset: `core-chain` + `sdk` patch.
- Opus adversarial (codex capped): the extraction is behavior-preserving on the core side (same tier math, now shared); the only behavior change is the RN override now matching core.

`@vultisig/core-chain` + `@vultisig/sdk` ship to flagship. The core-side change is a pure refactor (identical output). The RN-override change makes mobile match the core/web slippage — strictly-more-correct (tighter on stable pairs, honors overrides). No flat-1%-dependent behavior should exist since that was the bug.

## risk

Low, user-protective. Mobile stable-pair swaps get a tighter (safer) floor and explicit tolerances are honored; the core path is unchanged.

🤖 Agent-generated


rcoderdev (approved):

neavra (approved):
approve. the RN LiFi slippage regression is real and the shared-module extraction removes the core/RN drift cleanly. red-proofed at head `8f1636a9`.

**the bug** — the RN override hardcoded a flat 1% slippage and had no `slippage`/`ticker` input, so on RN (most users) every LiFi quote used 1%: stable pairs that get the 0.3% tier on the core path got 1% (wider MEV/loss surface), and an explicit tight-tolerance request was silently dropped (LiFi bakes `minAmountOut` from the slippage). confirmed the override previously ignored the caller's value.

- wiring: `findSwapQuote` spreads `...from`/`...to` (which carry `ticker`) and passes `slippage: lifiSlippageFraction` into `getLifiSwapQuote` — and on RN that call resolves to this override (rollup redirect). so the override now has the ticker for stable-pair detection and the caller's override. at runtime the old override already *received* the ticker (object spread doesn't strip it) but discarded it; the fix reads it via `resolveLifiSlippage`.
- extraction is behavior-preserving: the tier constants (0.01 / 0.003), `MAX_COMBINED_COST_BPS` (300), and the `STABLE_TICKERS` set are moved verbatim into `lifi/api/lifiSlippage`; the core path re-exports `isStablePair`/`STABLE_TICKERS` for tests. core lifi suite stays green → no core regression.
- the redirect-loop rationale checks out: the RN override can't import core's `getLifiSwapQuote` (rollup redirects that module back to itself), so the sibling `lifiSlippage` module is the correct single source of truth both sides import — no drift possible.
- the new `./swap/general/lifi/api/lifiSlippage` package export is required so the RN bundle can reach it; in-repo typecheck/tests resolve via tsconfig/vitest source path-mappings, CI builds the dist.

**receipts** (head `8f1636a9`)
- RN override suite: **6 passed** (3 new slippage tiers — USDC→USDT = 0.3%, ETH→USDC = 1%, override 0.001 honored; + 3 pre-existing approvalAddress).
- core lifi suite: **34 passed**.
- red-proof: reverted the override to flat `0.01` → the stable-pair (0.003) and override (0.001) tests go red, the volatile (0.01) test stays green. mutants killed, tree restored.
- core `tsc` + `@vultisig/sdk` typecheck both clean. changeset present (core-chain + sdk patch). the 04:28 main-merge is a clean no-conflict merge.

no blocking issues. lgtm — worth landing: it's a slippage-widening + dropped-override bug hitting most (RN) users, and the extraction kills the two-copies drift that caused it.

packages/core/chain/swap/general/lifi/api/lifiSlippage.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export const resolveLifiSlippage = ({
+  slippageOverride,
+  from,
+  to,
+}: {
+  slippageOverride: number | undefined
+  from: { ticker?: string }
+  to: { ticker?: string }
+}): number =>
```
NeOMakinG, line 66: `suggestion:` resolveLifiSlippage has no internal max-slippage clamp — relies entirely on findSwapQuote's SLIPPAGE_TOLERANCE_MAX_PERCENT=50 cap — resolveLifiSlippage returns `slippageOverride ?? tier` verbatim. The only bound on the override is at the shared entry point findSwapQuote.assertValidSlippageTolerance (rejects >50%), which runs before toProviderSlippage/lifiFraction. So on the real app path a runaway override (e.g. 100%) is already fail-closed. This matches core exactly and is pre-existing — the widening is user-protective. The residual is only for a consumer that calls getLifiSwapQuote (core or RN) DIRECTLY, bypassing findSwapQuote: no clamp there, and a near-1.0 fraction bakes a near-zero minAmountOut. Not introduced by this PR and out of scope; noting for symmetry with the 1inch/AGG guards.

fix: Optional follow-up: a cheap `Math.min(resolved, CEILING)` inside resolveLifiSlippage would make the shared helper self-defending regardless of caller, closing the direct-call gap for both core and RN in one place.

NeOMakinG (approved):
gm — multi-lane pass: claude-correctness-fundsafety + codex-adversarial (stalled, no output this run). runtime-tested on node/vitest — @vultisig/sdk workspace, PR HEAD 8f1636a9. lgtm.

## testing approach
- node/vitest — @vultisig/sdk workspace, PR HEAD 8f1636a9
- pass order: claude-correctness-fundsafety → codex-adversarial (stalled, no output this run)

## suggestion
### 1. resolveLifiSlippage has no internal max-slippage clamp — relies entirely on findSwapQuote's SLIPPAGE_TOLERANCE_MAX_PERCENT=50 cap — `packages/core/chain/swap/general/lifi/api/lifiSlippage.ts:66`
resolveLifiSlippage returns `slippageOverride ?? tier` verbatim. The only bound on the override is at the shared entry point findSwapQuote.assertValidSlippageTolerance (rejects >50%), which runs before toProviderSlippage/lifiFraction. So on the real app path a runaway override (e.g. 100%) is already fail-closed. This matches core exactly and is pre-existing — the widening is user-protective. The residual is only for a consumer that calls getLifiSwapQuote (core or RN) DIRECTLY, bypassing findSwapQuote: no clamp there, and a near-1.0 fraction bakes a near-zero minAmountOut. Not introduced by this PR and out of scope; noting for symmetry with the 1inch/AGG guards.
action: Optional follow-up: a cheap `Math.min(resolved, CEILING)` inside resolveLifiSlippage would make the shared helper self-defending regardless of caller, closing the direct-call gap for both core and RN in one place.

## q
- ticker optional-widening `& { ticker?: string }` on the RN override Input — confirm stable-tier isn't silently dead on the RN path

- LiFi non-zero tx.value passthrough is STILL OPEN after this PR (getLifiSwapQuote.ts:343 `value: BigInt(shouldBePresent(value)).toString()`, and the RN override mirrors it) — the sibling of the #1431 1inch leak I approved. This PR touches the LiFi path but only the slippage seam; the value seam is untouched, correctly out of scope. Tracked in bead sdk-lifi-nonzero-value-leak; recommend porting the isFeeCoin-gated non-zero-value reject to LiFi (core + RN) as a follow-up.

## risk
Low, user-protective. The core-side change is a pure extract-to-shared-module refactor (identical tier math, now the single source of truth). The RN override change makes mobile resolve slippage identically to core: stable pairs get the tighter 0.3% floor instead of a flat 1% (narrower MEV surface), and an explicit consumer tolerance is now honored instead of silently dropped. Units are correct end-to-end (consumer percent → findSwapQuote toProviderSlippage lifiFraction = /100 → fraction → LiFi getQuote `slippage` fraction → combinedCostBps = *10000). Bounds enforced upstream at findSwapQuote (≤50%). No behavior depends on the old flat-1% (that was the bug).

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4-codex` adversarial pass. node/vitest — @vultisig/sdk workspace, PR HEAD 8f1636a9.

PR #1429, opened 2026-07-20
fix(qbtc): broadcast reports false success on delivertx failure

### LINKED ISSUE #1384: fix(qbtc): broadcastQbtcTx shares the cosmos retry-swallow bug and never checks DeliverTx
`broadcastQbtcTx` (`packages/core/chain/tx/broadcast/resolvers/qbtc.ts`) shares the exact false-success-via-retry bug sdk#1316 fixed for cosmos, and additionally never checks DeliverTx at all.

## What it says on the box

QBTC's chain kind is `qbtc` (`ChainKind.ts:43`), which is NOT exempt from the retry wrapper (`hasResolverOwnedRetry` in `packages/core/chain/tx/broadcast/index.ts:33` only exempts `evm`/`solana`), so it runs inside `withTransientBroadcastRetry` same as cosmos.

Two separate issues in the same file:

1. **Same retry-swallow door as sdk#1316/#1383.** On a non-zero CheckTx code, `broadcastQbtcTx` throws a bare `Error("QBTC tx error: " + raw_log)` (`resolvers/qbtc.ts:37`) with chain-controlled `raw_log` text. `verifyBroadcastByHash` re-throws (tx isn't on-chain) (`:49`), then `isTransientBroadcastError` classifies the bare error against the message regex - a transient-sounding `raw_log` gets retried, resent, and swallowed as success on "tx already exists in cache" (`:24`/`:45`), exactly like the cosmos CheckTx door in #1383.
2. **No DeliverTx check at all (broader than the retry issue).** QBTC uses `BROADCAST_MODE_SYNC`, which only confirms CheckTx acceptance. A QBTC send that passes CheckTx but reverts in DeliverTx execution is reported as success unconditionally - there's no `assertIsDeliverTxSuccess` equivalent here the way sdk#1316 added for cosmos's `data` branch. This is the QBTC counterpart of the original sdk#1312 bug, not yet fixed for this resolver.

Note: `packages/core/chain/chains/cosmos/qbtc/claim/broadcastClaimTx.ts` is a *different*, unaffected path - it runs outside the dispatcher/retry wrapper and throws bare errors with no resend, so it doesn't share this bug.

## Suggested fix

Mirror sdk#1316's approach for `broadcastQbtcTx`: check the DeliverTx result (not just CheckTx/sync acceptance) for on-chain execution success, and give genuine failures the same `DeliverTxFailedError`-style terminal marker so the retry wrapper can't misclassify chain-controlled `raw_log` text as transient.

## Note for agents

Found during adversarial review of sdk#1316. See `packages/core/chain/tx/broadcast/resolvers/qbtc.ts`, `packages/core/chain/tx/broadcast/transientRetry.ts`, and the fix pattern in `packages/core/chain/tx/broadcast/resolvers/cosmos.ts`.

gomesalexandre:
Closes vultisig/vultisig-sdk#1384

Round-7 / R2 broadcast-audit HIGH. The QBTC half of the DeliverTx false-success class — neavra pre-flagged this on #1378; the cosmos/raw-cosmos halves are #1316 and #1426.

## what

A QBTC send that failed **execution** was reported as a **successful broadcast** (funds appear sent, nothing moved).

## why

`broadcastQbtcTx` (`tx/broadcast/resolvers/qbtc.ts`) uses `BROADCAST_MODE_SYNC`, which only surfaces the **CheckTx** (mempool-admission) code. A DeliverTx failure — out-of-gas, execution revert — still comes back `code: 0` at broadcast time. The resolver returned success after only that CheckTx check, with **no inclusion poll**. Its own claim sibling (`broadcastClaimTx.ts`) already does `waitForTxInclusion` + re-checks the included `code`; the send resolver omitted exactly that. Second bug: `data.tx_response?.code && …` treated a **missing** `tx_response` as a silent success.

## how

Poll for inclusion and re-check the DeliverTx `code`, mirroring the claim helper — whose `waitForTxInclusion` is **extracted to a shared `waitForQbtcTxInclusion`** so both paths use one implementation (no drift). After CheckTx passes:
- **confirmed DeliverTx failure** (`code !== 0`) → throw a non-retryable `DeliverTxFailedError` (so `withTransientBroadcastRetry` doesn't resend-and-swallow it, same posture as the cosmos resolver);
- **unconfirmable inclusion** (poll timeout / transient RPC error) → left **in-flight** (the tx passed CheckTx so it's in the mempool, not failed; the status resolver — which already checks the DeliverTx `code`, `status/resolvers/qbtc.ts:37` — reports the final outcome);
- **missing / non-zero CheckTx code** → verified by hash, not trusted (closes the missing-`tx_response` silent-success).

## coordination

Touches only `resolvers/qbtc.ts` + the extracted `waitForQbtcTxInclusion.ts` + the claim rewire — **not** `QBTCTx.ts`/`QBTCHelper.ts`, so no conflict with #1378 (gaslimit) or #1424 (golden fixture).

## testing

- +5 resolver cases: DeliverTx-fail → throws `DeliverTxFailedError`; both-succeed → resolves (no verify); missing `tx_response` → verify-by-hash (not silent success); CheckTx reject → verify; unconfirmable inclusion → in-flight (no throw).
- Claim path unchanged — full qbtc suite **86 passed** (the claim tests exercise the now-shared helper). Full `yarn test:core` **1855 passed**; `tsc` + eslint + prettier clean. Changeset: `core-chain` patch.

`@vultisig/core-chain` ships to flagship. QBTC broadcast now blocks until DeliverTx inclusion (≤30s) instead of returning right after CheckTx — this **aligns QBTC with the rest of the cosmos family** (the standard cosmos resolver already blocks-to-inclusion via StargateClient). Strictly-more-correct: a reverted send now surfaces as a failure instead of a false success; a slow-to-confirm send stays in-flight rather than false-failing.

## risk

Medium (adds a blocking inclusion poll on a shipped path), but fund-safety-positive and consistent with the cosmos-family norm. The previous behavior was a guaranteed false-success on any DeliverTx failure.

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QBTC transaction status reporting by distinguishing broadcast validation from on-chain confirmation.
  * Prevented transactions that fail during final processing from being reported as successful.
  * Added reliable handling for delayed transaction inclusion, missing response data, and validation failures.
  * Improved fallback verification using the transaction hash when broadcast details are incomplete.

* **Tests**
  * Added coverage for successful confirmations, processing failures, delayed inclusion, and network errors.


rcoderdev (approved):

neavra (changes requested):
the runtime fix is right — this is the qbtc `BROADCAST_MODE_SYNC` door (only CheckTx is surfaced at broadcast, so an out-of-gas / revert DeliverTx came back `code:0` and got reported as success, and the old `data.tx_response?.code && …` also treated a *missing* `tx_response` as success). polling for inclusion and re-checking the DeliverTx code, throwing the `DeliverTxFailedError` marker, is the correct pattern. two blockers below are both packaging, not logic — the PR is red on CI as-is.

what's good (verified, not taken on faith):
- **interlock is correct.** qbtc's ChainKind is `qbtc`, not in `hasResolverOwnedRetry` (evm/solana only), so this resolver runs *inside* `withTransientBroadcastRetry`. throwing the `DeliverTxFailedError` marker (not a plain Error) is exactly what's needed: `isTransientBroadcastError` short-circuits `instanceof DeliverTxFailedError → return false` as the first check in the cause loop, before the message-regex — so the chain-controlled `raw_log` you embed in the message can't get it reclassified transient and rebroadcast. thrown unwrapped, so it arrives at the classifier as the top-level error. same fix shape as #1316.
- **truthful both directions.** confirmed-fail (included `code !== 0`) throws; both-success resolves; missing/non-zero CheckTx verifies by hash instead of trusting; unconfirmable inclusion (timeout / transient rpc) is left in-flight rather than thrown as a false failure. red-proof (reverted the resolver to main, kept the new tests): 2/5 fail — `throws DeliverTxFailedError when CheckTx passes but DeliverTx fails` → `promise resolved "undefined" instead of rejecting`, and `does NOT silently succeed when tx_response is missing` → verify never called. that's the exact bug, both halves.
- **refactor is clean.** extracting `waitForTxInclusion` → shared `waitForQbtcTxInclusion` doesn't regress the claim path: `broadcastClaimTx` tests 13/13, and `QbtcTxResponse` declares every field the claim consumer reads (code/events/txhash/raw_log/log).
- broadcast+qbtc suite 164/164; new resolver test 5/5; `tsc --noEmit` clean.

**blocking 1 — CI is red: stale core-chain export map.** the new module is imported via its package export path (`@vultisig/core-chain/chains/cosmos/qbtc/waitForQbtcTxInclusion`, in both `resolvers/qbtc.ts` and `claim/broadcastClaimTx.ts`) but `packages/core/chain/package.json` wasn't regenerated. the "Lint and Type Check" job fails at `yarn check:ci` → `[shared-exports] packages/core/chain/package.json has stale generated exports. Missing exports: + ./chains/cosmos/qbtc/waitForQbtcTxInclusion` (exit 1), and "Test Summary" is just the downstream rollup of that (unit + integration themselves passed). fix: `yarn build:shared` and commit the package.json. tests/typecheck don't catch it locally because tsconfig paths + vitest alias map `@vultisig/core-chain/*` straight to source, bypassing the exports field — it only bites the shared-exports gate and real published-package consumers.

**blocking 2 — changeset omits `@vultisig/sdk`.** it bumps only `@vultisig/core-chain`. `@vultisig/sdk` bundles core-chain at build time (rollup) and has no manifest dependency on it, so changesets (`updateInternalDependencies: patch`) has no edge to follow and won't bump sdk — the fix ships in core-chain but never version-bumps the sdk tarball its consumers actually use. add `@vultisig/sdk: patch`, mirroring the #1316 sibling changeset which lists both for this exact reason. (the `changeset:check-sdk-bundle` gate passes right now only incidentally — unrelated sibling changesets in `.changeset/` already bump sdk; it's not passing because this PR is complete.)

non-blocking:
- the in-flight fallback means a DeliverTx failure that lands *after* the 30s inclusion window isn't caught at broadcast time — but the status resolver is the authority on final code and this is strictly better than the old always-success, so leaving it is fine. just calling it out.

happy to re-approve once CI's green and the sdk changeset's in. reviewed at e5d84473.

packages/core/chain/tx/broadcast/resolvers/qbtc.ts
```diff
@@ -30,21 +37,49 @@ export const broadcastQbtcTx: BroadcastTxResolver<typeof Chain.QBTC> = async ({
...
-  if (isInError(error, 'tx already exists in cache')) {
+  const { data: included, error: inclusionError } = await attempt(
+    waitForQbtcTxInclusion({
+      txHash,
+      timeoutMs: QBTC_INCLUSION_TIMEOUT_MS,
+      intervalMs: QBTC_INCLUSION_POLL_INTERVAL_MS,
+    })
+  )
+
+  if (inclusionError || included === undefined || typeof included.code !== 'number') {
```
NeOMakinG, line 78: `q:` this branch folds a malformed/partial 200 inclusion payload (missing `code`) into the same in-flight `return` as cosmos' legitimate omitted-`code`==success. Safe here because `getQbtcTxStatus` re-reads the same endpoint and treats a missing `code` as `error` (`code === 0 ? 'success' : 'error'`), so no permanent false success — but if you wanted this strictly fail-closed, the malformed case would route through `verifyBroadcastByHash` instead of returning. Non-blocking.

NeOMakinG (approved):
Fund-critical broadcast-honesty fix — the QBTC leg of the DeliverTx false-success class (siblings #1316 cosmos, #1426 raw-cosmos, both merged). Reviewed at HEAD `fa16d297`. Closes #1384 (the retry-swallow). Two-lane: Claude primary + Codex adversarial. **Verdict: APPROVE.**

## testing approach

- Read the full diff + the surrounding surfaces: `transientRetry.ts` (DeliverTxFailedError / `isTransientBroadcastError` / `withTransientBroadcastRetry`), `broadcast/index.ts` (retry wiring), `verifyBroadcastByHash.ts`, `status/resolvers/qbtc.ts`, `attempt.ts`.
- Verified the fix against the accepted #1426 template: read the RESOLVED EXECUTION result (inclusion-poll `code`, not the CheckTx/mempool ack), throw fail-closed on confirmed non-zero, keep the mempool/RPC-reject on the verify-by-hash arm.
- Ran the targeted suite in a fresh checkout at HEAD: `qbtc.test.ts` + `cosmosDeliverTxRetry.test.ts` + `index.test.ts` + `verifyBroadcastByHash.test.ts` → **24 passed**.
- **Red-on-revert:** neutered the `throw new DeliverTxFailedError` → the "throws on DeliverTx out-of-gas" case flips to `resolved: undefined` (1 failed / 4 passed). The test is non-inert and load-bearing.

## blocking

None.

## preferably-blocking

None.

## should-fix

None.

## q

Two spots where the resolver returns (broadcast "accepted") and delegates final truth to the status resolver — both are safe as-is because `getQbtcTxStatus` re-reads the same endpoint and is fail-closed (`code === 0 ? 'success' : 'error'`, missing `code` → `'error'`, missing `tx_response` → `pending/isKnown:false`), so neither can strand a permanent false success. Flagging for your confirmation, not as blockers:

1. **Inclusion-stage guard defaults a malformed 200 to in-flight.** `typeof included.code !== 'number' → return` folds two shapes together: cosmos' legitimate omitted-`code`-means-0-means-success, and a genuinely malformed/partial inclusion payload. Both lanes independently flagged this. It's fine given the status backstop marks a missing `code` as `error`, but if you ever wanted this branch strictly fail-closed it would route the malformed case through `verifyBroadcastByHash` rather than returning.
2. **`tx already exists in cache` fast-path skips the inclusion re-check.** On an MPC duplicate-broadcast race the losing device returns without polling the cached tx's DeliverTx result. This is identical to the pre-existing behavior and to the cosmos sibling, and the winning device runs the full poll — so no new exposure — just noting the cached leg's correctness rests entirely on the later status pass.

## risk

Medium, as the author states — adds a blocking inclusion poll (≤30s) on a shipped `@vultisig/core-chain` path. Strictly fund-safety-positive: a reverted send now surfaces as a failure instead of a false success, a slow-to-confirm send stays in-flight rather than false-failing, and this aligns QBTC with the cosmos-family norm (StargateClient already blocks-to-inclusion). The `waitForTxInclusion` → shared `waitForQbtcTxInclusion` extraction is behavior-identical (the claim suite exercises the shared helper), so no claim-path drift. `DeliverTxFailedError` is confirmed non-retryable (`isTransientBroadcastError` short-circuits to `false` before the message-regex), and QBTC goes through `withTransientBroadcastRetry` (not the EVM/Solana bypass) — so #1384's retry-swallow is fully closed.

## verdict

APPROVE. Matches the #1426 template, closes the #1384 retry-swallow, tests are red-on-revert. The two `q`s are backstop-covered non-blockers.

---
🤖 reviewed by Claude (primary correctness/fund-safety lane) + Codex gpt-5.4 (adversarial second opinion)

neavra (approved):
lgtm — both packaging blockers fixed, CI now green, qbtc logic byte-unchanged from the reviewed head.

- **blocking 1 (stale core-chain export map → CI red)** — RESOLVED: `packages/core/chain/package.json` regenerated via `build:shared`; it now carries the `./chains/cosmos/qbtc/waitForQbtcTxInclusion` subpath export. "Lint and Type Check" and the "Test Summary" rollup are both green on `fa16d297` (run 29717002095), so `check:shared-exports` no longer trips.
- **blocking 2 (changeset omits `@vultisig/sdk`)** — RESOLVED: `@vultisig/sdk: patch` added to `.changeset/qbtc-broadcast-delivertx-false-success.md`, so the fix now ships in the sdk tarball, not only core-chain.

Verified the delta is packaging-only: the diff vs the reviewed head is the export-map regen + the changeset line; the resolver/`waitForQbtcTxInclusion` logic is unchanged (the `RawBroadcastService` hunk is the inherited #1426 merge from main, orthogonal to qbtc). All 5 checks pass. Non-blocking post-30s-window in-flight note left as-is per your reply — agreed, the status resolver is authority on the final code. Thanks for the quick turnaround.

Issue #1119, opened 2026-07-09
reliability: 5 status resolvers violate the isKnown contract — broadcast failures reported as SUCCESS (sui/ton/tron/bittensor/qbtc)


realpaaao:
## Problem — broadcast failures reported as SUCCESS on 5 chains

`verifyBroadcastByHash` (packages/core/chain/tx/broadcast/verifyBroadcastByHash.ts:36-40) swallows a broadcast error when the status lookup answers `pending` **unless** the resolver sets `isKnown: false`:

```ts
const isKnownPending = result.status === 'pending' && result.isKnown !== false
if (result.status === 'success' || isKnownPending) return  // error swallowed
```

The resolver contract (status/resolver.ts:5-10) says resolvers that can't distinguish "unknown hash" from "in flight" must return `pending` with `isKnown: false`. Five resolvers violate it — they return **bare `{status:'pending'}`** for unknown-hash AND for RPC error:

| resolver | violation site |
|---|---|
| sui.ts | `if (error \|\| !data) return { status: 'pending' }` (status/resolvers/sui.ts:20-22) |
| ton.ts | empty/err → `{ status: 'pending' }` (ton.ts:29-31) |
| tron.ts | err/unknown/blockNumber 0 → `{ status: 'pending' }` (tron.ts:48-50) |
| bittensor.ts | indexer err/unknown → `{ status: 'pending' }` (bittensor.ts:30-32) |
| qbtc.ts | lookup failure → `{ status: 'pending' }` (qbtc.ts:32-34) |

Consequence: **any genuine node rejection at broadcast** (stale nonce, expired era, bad signature, insufficient balance, rejected extrinsic) routes into `verifyBroadcastByHash`, the never-seen hash comes back "pending", and **the SDK reports the broadcast as successful**. Wallet says sent; nothing is on chain. On Tron this additionally makes the well-built hex-decoded error hoisting (broadcast/resolvers/tron.ts:17-21) dead code.

## Live probe (2026-07-09, real endpoints, fabricated never-seen hash)

```
sui        → {"status":"pending"}                → verifyBroadcastByHash SWALLOWS (false success)
ton        → {"status":"pending"}                → SWALLOWS
tron       → {"status":"pending"}                → SWALLOWS
bittensor  → {"status":"pending"}                → SWALLOWS
qbtc       → {"status":"pending"}                → SWALLOWS
evm        → {"status":"not_found","isKnown":false} → rethrows (correct)
[...]

PR #1431, opened 2026-07-20
fix(swap): reject non-zero 1inch tx.value on a token-source swap

gomesalexandre:
R2 audit P3 cluster (a). The clean, self-contained one of the three.

## what

1inch's `tx.value` flows through from the untrusted `/swap` response **verbatim** (spread into the `GeneralSwapQuote` at `getOneInchSwapQuote.ts`), unlike Kyber which constructs `value` itself (`kyber/api/tx.ts:150` = `isFeeCoin(from) ? amount : '0'`). So a compromised/buggy 1inch response could set a non-zero `value` on a **token→token** swap and move native gas-coin the user never authorized alongside the swap — the router allowlist (AGG-02) binds `tx.to` but not `tx.value`.

## how

After the router allowlist check, reject a non-zero `tx.value` when the source is a **token** (`fromCoinId` set → not the native fee coin). A token-source swap pulls the sell token via ERC-20 allowance, so `value` must be `0`. Native-source swaps — where `value` is legitimately the sell amount — are unaffected.

## testing

- +3 cases: token-source + non-zero value → rejects; native-source + non-zero value → accepted (value is the sell amount); token-source + zero value → accepted. Full oneInch suite green; `tsc` + eslint clean. Changeset: `core-chain` patch.

## cross-consumer

`@vultisig/core-chain` ships to flagship. Fail-closed at quote construction; a legitimate token swap always returns `value: 0` from 1inch, so no real swap is blocked — only a response injecting native value into a token swap.

## the other two P3 items — reported, not fixed here

- **Solana broadcast `AlreadyProcessed` = success**: this is a **documented, maintainer-accepted trade-off** (PR #874 — the broadcast layer's job is "did the node take the payload"; the authority on execution success is the downstream `getTxStatus` confirmation poll, which reads `signatureStatus.err`). Like the li.fi log-only case, leaving it. Ties #1119.
- **Tron broadcast returns the full `{txid,result,code,message}` object** instead of the hash string — a real inconsistency, but `BroadcastTxResolver` returns `unknown` and the return is consumed outside this clone (app/flagship), so I'm flagging it rather than changing a return shape blind. Recommended: `return result.txid` after verifying no consumer reads the object.
- **skip `svm_tx` test (P3c)**: the referenced path `skip/skipSwap.ts:568-585` doesn't exist on current main (the skip file is `skip/skipRouting.ts`, and no `svm_tx` handling matched) — reference looks stale; flagging for the router to re-point.

## risk

Low, fund-safety-positive. One fail-closed guard at quote construction.

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for 1inch token-to-token swap quotes to reject unexpected non-zero transaction values.
  * Preserved support for non-zero transaction values in native-coin swaps.
  * Improved protection against unsafe or malformed swap quotes before signing.

* **Tests**
  * Added coverage for valid and invalid transaction value scenarios across token and native-coin swaps.


NeOMakinG (approved):
## testing approach

Reviewed at HEAD `ce2e268b` in an isolated worktree. Two-lane: Claude load-bearing correctness/fund-leak; Codex lane non-functional this run (stalled — best-effort, skipped). Built shared packages, ran the `getOneInchSwapQuote` vitest suite (12/12 green including the 3 new guard cases), and ran a red-on-revert: neutering the guard to `if (false)` fails **exactly** the reject test (1 failed / 11 passed) — the test is load-bearing, not inert.

## verdict

**APPROVE.** The token-source native-value leak on 1inch is reliably closed and fail-closed. One systemic follow-up (LiFi) noted below, out of this PR's scope.

### fund-leak analysis (the thing that matters)

- **Source classification is correct.** `isFeeCoin({ id: fromCoinId, chain })` reduces to `!fromCoinId`. Native-source (`fromCoinId === undefined`) → guard skipped, so a legit native swap keeps its `value == sell amount`. Token-source (`fromCoinId` set → an ERC-20 contract address) → guard applies. The `Input` doc already forbids the caller substituting a ticker/sentinel fallback for `undefined`, so the classifier can't be tricked into treating a native swap as a token (worst case of caller misuse is fail-closed rejection of a legit native swap, never a leak).
- **No non-zero can slip past.** `tx.value` is typed `string`; the guard `tx.value && tx.value !== '0'` only lets falsy (`undefined`/`''`) or exactly `'0'` through — none of which is a non-zero native amount. Any non-zero string, hex, or (runtime-injected) number is truthy and `!== '0'`, so it rejects. Edge shapes like `'0x0'` reject too — fail-closed, safe.
- **Placed before the quote becomes signable** — after the AGG-02 router allowlist and before the `...tx` spread into `GeneralSwapQuote`, so the poisoned value never reaches the signer.
- **Regression-safe:** native+non-zero accepted (test), token+zero accepted (test), token+non-zero rejected (test + red-on-revert).

## should-fix (systemic — NOT blocking this PR)

`should-fix:` the same verbatim-`value` passthrough exists on **LiFi**: `getLifiSwapQuote.ts:343` returns `value: BigInt(shouldBePresent(value)).toString()` straight from the untrusted `transactionRequest`, with no token-source zero-guard — the identical native-value-injection class this PR closes for 1inch. Kyber is safe (it constructs `value` itself: `isFeeCoin(from) ? amount : '0'`). Swapkit (`getSwapKitQuote.ts` `value?: string|number|bigint`) is a possible third and worth an audit pass. Recommend a follow-up bead to port this guard to LiFi (and confirm swapkit). Out of scope here — 1inch itself is fully covered.

## risk

Low, fund-safety-positive. One fail-closed guard at quote construction; a legitimate 1inch token swap always returns `value: '0'`, so no real swap is blocked.

---
🤖 Reviewed by Claude (reviewer agent). Codex lane best-effort/unavailable this run.

rcoderdev (approved):

neavra (approved):
approve. red-proofed at head `1d56bd55`; the token-source native-value injection on 1inch is fail-closed.

**what it does** — rejects a non-zero `tx.value` from the untrusted 1inch `/swap` response when the sell side is a token. 1inch's `value` is spread verbatim into the quote (`...tx`), unlike kyber which constructs it, so a compromised/buggy response could move native gas-coin the user never authorized alongside a token→token swap. a token-source swap pulls the sell token via allowance → value must be `0`; native-source legitimately carries value == the sell amount.

- source classification is exact: `isFeeCoin({ id: fromCoinId, chain })` reduces to `!fromCoinId`. native (no `fromCoinId`) → guard skipped; token (`fromCoinId` = erc-20 addr) → guard applies. worst case of caller misuse is a fail-closed reject of a legit native swap, never a leak.
- placement: after the AGG-02 router allowlist and *before* the `...tx` spread into `GeneralSwapQuote` — the poisoned value never reaches a signable quote.
- no non-zero slips: `tx.value && tx.value !== '0'` only lets falsy or exactly `'0'` through; `'0x0'` / hex / any non-zero string is truthy and `!== '0'` → rejects. fail-closed.

**receipts** (head `1d56bd55`)
- oneInch suite: **12 passed** (9 pre-existing + 3 new).
- red-proof: neutered the guard → the token+non-zero reject test goes red (1 failed | 11 passed); native+non-zero and token+zero stay green. mutant killed, tree restored.
- core `tsc` (`.config/tsconfig.json`) clean. changeset present (core-chain patch). the 04:28 main-merge is a clean no-conflict merge; net PR diff = the guard.

**non-blocking — systemic follow-up**
- the same verbatim-`value` passthrough exists on **LiFi**: `getLifiSwapQuote.ts:343` returns `value: BigInt(shouldBePresent(value)).toString()` straight from the untrusted `transactionRequest`, with no token-source zero-guard — the identical injection class this closes for 1inch. kyber is safe (`value: isFeeCoin(from) ? amount : '0'`); swapkit worth an audit pass. recommend porting this guard to lifi as a follow-up bead. 1inch itself is fully covered here.

lgtm.

NeOMakinG (approved):
re-review at the advanced HEAD `d4b4be8a`. my r1 APPROVE sat at `1d56bd55`, so this round exists only to establish what moved. short round on purpose: **the delta contains no semantic change to this branch's code.**

## testing approach

two-diff recipe against both parents of the merge commit, then a byte-level read of the one non-merge commit:

- `git diff --stat 21b4840f(main) d4b4be8a(HEAD)` -> `.changeset/oneinch-token-swap-value-guard.md` (5+), `getOneInchSwapQuote.test.ts` (49+), `getOneInchSwapQuote.ts` (11+)
- `git diff --stat $(git merge-base main HEAD) d4b4be8a` -> **identical, 3 files / 65 insertions**

the two agree exactly, which means the merge is clean in both directions: main's changes are fully contained (nothing from main dropped or double-applied) and the branch still contributes only its own three files. the `clients/cli/src/core/errors.{ts,test.ts}` + `.changeset/olive-donkeys-repeat.md` churn that shows up in `1d56bd55..d4b4be8a` is **main's**, arriving via the merge — not this branch's.

that leaves exactly one branch-authored commit since my approval, `bc9120ab` "style: prettier-format 1inch value-guard test". read hunk by hunk: it splits two inline `tx: { ... }` object literals across lines in the two new test cases. same keys, same order, same values (`value: '1000000000000000000'` in both). zero behavioural content — it exists to satisfy the `check:ci` format gate, which is now green.

the guard itself: `git diff 1d56bd55 d4b4be8a -- packages/core/chain/swap/general/oneInch/api/getOneInchSwapQuote.ts` is **empty**. the signable path is byte-identical to what three of us approved.

CI at HEAD: Unit Tests / Integration Tests / Lint and Type Check / Test Summary all pass.

## prior issues — status

|---|---|---|---|---|---|
| r1 | — | blocking | none raised | ℹ️ n/a | r1 was a clean approve |
| r1 | 1 | should-fix | LiFi has the same verbatim-`value` passthrough (`getLifiSwapQuote.ts:343`, `value: BigInt(shouldBePresent(value)).toString()` straight off the untrusted `transactionRequest`, no token-source guard); swapkit a likely third (`value?: string \| number \| bigint`) | ⏭️ deferred | explicitly out of scope for this PR, as agreed in r1. still true at HEAD. worth a follow-up to port the same three-line guard |

## risk

unchanged from r1 — low, and strictly fail-closed. `!isFeeCoin({ id: fromCoinId, chain })` is `!!fromCoinId`, so native-source swaps (`fromCoinId: undefined`) skip the guard and keep their legitimate sell-amount `value`; token-source swaps pass only on falsy or exactly `'0'`, so any non-zero string, hex (`'0x0'` included), or coerced number rejects. worst case on caller misuse is a refused legitimate swap, never a leak. placement is before the `...tx` spread into the signable quote, so it gates the SIGNED value and not a display copy.

## verdict

**approve**, re-affirmed at `d4b4be8a`. no new findings — manufacturing any here would be noise over a prettier reformat.

---
🤖 automated review — Claude lane (load-bearing). Codex lane not run this round: the branch-authored delta was whitespace-only, so a second adversarial pass had nothing to read.

PR #1432, opened 2026-07-20
fix(bittensor): don't assume success on a malformed rpc response

gomesalexandre:
R2 broadcast-audit LOW. Last of the reported broadcast false-success cluster.

## what

`broadcastBittensorTx` only inspected `response.error`. A JSON-RPC body with **neither** `error` **nor** `result` — a truncated / malformed gateway response — fell through the error block and returned `undefined`, which the broadcast layer reports as a **successful broadcast** even though the node never confirmed it took the extrinsic.

## how

Force hash verification when `result` is absent, mirroring the Polkadot resolver's JSON-RPC 2.0 guard (`polkadot.ts:79`, "a valid response must have exactly one of result/error"). Added a `return` after the error branch so the new check only runs on a no-error response (no double-verify).

## testing

- +4 cases: accept on `result` present; swallow the idempotent `Already Imported` peer-race; verify-by-hash on a genuine error; **verify-by-hash on a malformed `{}` body** (the gap). Broadcast suite green; `tsc` + eslint clean. Changeset: `core-chain` patch.

## cross-consumer

`@vultisig/core-chain` ships to flagship. Strictly-more-correct: a malformed response now triggers hash verification (which resolves the real state) instead of a silent success. A well-formed accept (`result` present) is unaffected.

## risk

Low. One added guard mirroring an existing sibling resolver.

🤖 Agent-generated

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Bittensor transaction broadcasting when RPC responses are incomplete or malformed.
  - Transactions are now verified by hash instead of being incorrectly reported as successful when no result is returned.
  - Preserved handling for already imported transactions.
  - Prevented failed hash verification from triggering an unnecessary rebroadcast.

### DIFF
--- a/packages/core/chain/tx/broadcast/resolvers/bittensor.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/bittensor.ts
@@ -31,5 +31,14 @@ export const broadcastBittensorTx: BroadcastTxResolver<OtherChain.Bittensor> = a
     }
     const err = new Error(`Bittensor broadcast failed: ${message || `code ${response.error.code}`}`)
     await verifyBroadcastByHash({ chain, tx, error: err })
+    return
+  }
+
+  // Per JSON-RPC 2.0 a valid response carries exactly one of `result` / `error`. Neither present
+  // (malformed / truncated gateway body) must NOT be assumed a success — force hash verification.
+  // Mirrors the polkadot resolver's guard; without it a truncated body returned `undefined` = success.
+  if (!response.result) {
+    const err = new Error('Bittensor broadcast failed: missing extrinsic hash in RPC response')
+    await verifyBroadcastByHash({ chain, tx, error: err })
   }
 }
--- a/.changeset/bittensor-malformed-response-false-success.md
+++ b/.changeset/bittensor-malformed-response-false-success.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Fix the Bittensor broadcast assuming success on a malformed RPC response. `broadcastBittensorTx` only inspected `response.error`; a body with neither `error` nor `result` (truncated / malformed gateway response) fell through and returned `undefined` — reported as a successful broadcast. It now forces hash verification when `result` is absent, mirroring the Polkadot resolver's JSON-RPC 2.0 guard.
--- a/packages/core/chain/tx/broadcast/resolvers/bittensor.test.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/bittensor.test.ts
@@ -10,6 +10,7 @@ vi.mock('../verifyBroadcastByHash', () => ({ verifyBroadcastByHash: mocks.verify
 vi.mock('@vultisig/core-chain/chains/bittensor/client', () => ({ bittensorRpcUrl: 'https://bittensor.test' }))
 
 import { OtherChain } from '../../../Chain'
+import { isTransientBroadcastError } from '../transientRetry'
 import { broadcastBittensorTx } from './bittensor'
 
 describe('broadcastBittensorTx', () => {
@@ -41,5 +42,21 @@ describe('broadcastBittensorTx', () => {
     mocks.queryUrl.mockResolvedValue({})
     await broadcastBittensorTx({ chain, tx })
     expect(mocks.verifyBroadcastByHash).toHaveBeenCalledOnce()
+  })
+
+  // #1430-class retry-interlock: bittensor runs inside `withTransientBroadcastRetry`
+  // (index.ts hasResolverOwnedRetry = evm|solana only). When hash verification
+  // cannot confirm the tx, verifyBroadcastByHash rethrows the original error and
+  // it escapes to the retry wrapper. That error MUST be classified non-transient
+  // so a genuinely-failed/unconfirmed broadcast is not silently re-sent up to 3×.
+  it('surfaces the malformed-response failure as a NON-transient error (must not be re-broadcast)', async () => {
+    mocks.queryUrl.mockResolvedValue({})
+    // Simulate verification unable to confirm the tx on-chain → rethrow.
+    mocks.verifyBroadcastByHash.mockImplementation(async ({ error }) => {
+      throw error
+    })
+    await expect(broadcastBittensorTx({ chain, tx })).rejects.toThrow(/missing extrinsic hash/)
+    const err = await broadcastBittensorTx({ chain, tx }).catch(e => e)
+    expect(isTransientBroadcastError(err)).toBe(false)
   })
 })
--- a/packages/core/chain/tx/broadcast/resolvers/bittensor.test.ts
+++ b/packages/core/chain/tx/broadcast/resolvers/bittensor.test.ts
@@ -1,0 +1,45 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const mocks = vi.hoisted(() => ({
+  queryUrl: vi.fn(),
+  verifyBroadcastByHash: vi.fn(),
+}))
+
+vi.mock('@vultisig/lib-utils/query/queryUrl', () => ({ queryUrl: mocks.queryUrl }))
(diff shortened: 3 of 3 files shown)


rcoderdev (approved):

neavra (approved):
lgtm — approving. clean fix, CI green, red-proofs cleanly.

`broadcastBittensorTx` only inspected `response.error`; a body with neither `error` nor `result` (truncated / malformed gateway response) fell through the bottom and returned `undefined` = a reported broadcast success. forcing `verifyBroadcastByHash` when `result` is absent is the right guard, and matching the polkadot resolver (`polkadot.ts` has the same `if (!response.result)` JSON-RPC 2.0 guard) is the right thing to mirror.

verified:
- **truthful both directions** — `result` present → returns success; the error branch returns early (idempotent `Already Imported` → return; genuine error → verify + return) so it never reaches the new guard; neither present → verify. no false-failure on a good response.
- **no transientRetry hazard** — this path introduces no execution-failure throw; it hands off to `verifyBroadcastByHash` (the existing arbiter). bittensor runs inside `withTransientBroadcastRetry`, but the handoff is benign — if a retry re-broadcasts an already-landed extrinsic the node answers `Already Imported`, which this resolver swallows idempotently. no marker type needed here (unlike the sui/qbtc siblings, which do throw).
- **red-proof** — reverted the guard, kept the tests → only `does NOT assume success on a malformed response … verifies by hash` fails (`expected verifyBroadcastByHash to be called once, got 0 times`); the other 3 stay green. the new test pins exactly the fix.
- new test 4/4; broadcast resolver suite 80/80; `tsc --noEmit` clean; prettier clean (repo config); CI green.

non-blocking (recommend before merge):
- the changeset declares only `@vultisig/core-chain: patch`. `@vultisig/sdk` bundles core-chain via rollup with no manifest dependency, so changesets (`updateInternalDependencies: patch`) won't version-bump sdk off this changeset alone — the #1316 sibling listed **both** packages for that reason, and i'd add `@vultisig/sdk: patch` here too for self-containment. not blocking today only because the just-merged #1426 carries an `@vultisig/sdk: patch`, so the next release republishes sdk and this bundled fix rides along — but that's merge-order luck; making it explicit here removes the dependency. (the other core-chain broadcast siblings omit it too.)

reviewed at f6e5c0ab; head moved to f4a27cf6 while reviewing — delta is a clean `main` merge pulling in #1426 (RawBroadcastService, orthogonal to this file), bittensor resolver/test/changeset unchanged, so the verdict holds.

packages/core/chain/tx/broadcast/resolvers/bittensor.ts
```diff
@@ -31,5 +31,14 @@ export const broadcastBittensorTx: BroadcastTxResolver<OtherChain.Bittensor> = a
     }
     const err = new Error(`Bittensor broadcast failed: ${message || `code ${response.error.code}`}`)
     await verifyBroadcastByHash({ chain, tx, error: err })
+    return
+  }
+
+  // Per JSON-RPC 2.0 a valid response carries exactly one of `result` / `error`. Neither present
+  // (malformed / truncated gateway body) must NOT be assumed a success — force hash verification.
+  // Mirrors the polkadot resolver's guard; without it a truncated body returned `undefined` = success.
+  if (!response.result) {
```
NeOMakinG, line 40: should-fix: `if (!response.result)` still lets a truthy-but-malformed hash through — a gateway returning `{ result: '0x' }` or a non-hash string is a malformed success that skips verification. Lower-likelihood than the truncated `{}` body you fixed, and identical to the accepted polkadot sibling, so non-blocking — but a hash shape/length check would close the class fully across both resolvers.

NeOMakinG (approved):
Two-lane review (Claude primary correctness + fund-safety, Codex gpt-5.4 adversarial second-opinion). Last of the broadcast false-success cluster (siblings #1426 cosmos, #1429 qbtc, #1430 sui). **fix-and-approve** — I pushed one test-hardening commit (`e1c5ebdc`) to lock the #1430 retry-interlock; the core fix is correct and safe.

## testing approach

- Read the fix + the `polkadot.ts` sibling it mirrors, `verifyBroadcastByHash.ts`, `transientRetry.ts`, and `index.ts` (routing/retry-wrapper) from source.
- `yarn vitest` the bittensor suite at HEAD: **5/5 pass** (4 original + my interlock).
- **Red-on-revert (author's tests):** neutered the `if (!response.result)` guard → the malformed-`{}` test flips to 1 failed / 3 passed. Non-inert. ✅
- **Red-on-revert (my interlock):** changed the throw message to contain `timed out` → interlock flips to 1 failed / 4 passed. Non-inert. ✅
- **#1430 double-broadcast check (the load-bearing one):** bittensor IS wrapped in `withTransientBroadcastRetry` (`index.ts` `hasResolverOwnedRetry` = evm|solana only). The malformed throw message `"Bittensor broadcast failed: missing extrinsic hash in RPC response"` matched against every `transientMessagePatterns` regex → **no match → non-transient → NOT re-broadcast.** Verified empirically.
- `tsc --project .config/tsconfig.json --noEmit` exit 0; eslint clean on the test file. Broadcast dir: 51 tests pass, 0 failures (4 files show as "failed" = pre-existing `dist`-not-built module-resolution collection errors for `@vultisig/core-config` / `@polkadot/api` / blockchair — unrelated to this change; bittensor imports none of them).

## why this member of the class does NOT need `DeliverTxFailedError`

Unlike cosmos/sui (#1426/#1430), `author_submitExtrinsic` returns **pool-acceptance**, not an execution result — there is no "included-but-reverted" state observable at broadcast time. The throw fires only through `verifyBroadcastByHash`, which **rethrows solely when the tx is NOT confirmed on-chain.** So even if the escaping error were transient-classified, a re-broadcast would be a legitimate idempotent retry (a peer-race resubmit hits `Already Imported` → swallowed), not a re-send of a confirmed-failed tx. Combined with the message being provably non-transient, there is no #1430-class double-broadcast here. Matches the `polkadot` sibling, which throws a bare `Error` into the same wrapper for the same reason.

## should-fix

- **`if (!response.result)` accepts truthy-but-malformed hashes** (Codex M2). A gateway returning `{ result: '0x' }` or a non-hash string is a malformed success that still bypasses verification. Lower-likelihood than the truncated-`{}` body this PR targets, and **identical to the accepted `polkadot` sibling** (`if (!response.result)`), so not blocking — but a shape/length check on the returned hash would fully close the class across both resolvers. Fail-direction note: this is the same latent hole polkadot carries, not a regression this PR introduces.

## q

- **queryUrl rejection / `undefined` body not routed through verify** (Codex M1). `polkadot` wraps the whole body in `try/catch` so a transport error or an unparseable body still gets a hash-verify pass; bittensor calls `verifyBroadcastByHash` inline per-arm, so a `queryUrl` throw propagates raw to the retry wrapper instead. This is **pre-existing** (the base resolver did the same) and **fail-closed** (a raw network throw is transient-retried, and a `TypeError` on an `undefined` body is non-transient → surfaced as failure, never a false success). Worth a follow-up to converge bittensor onto polkadot's try/catch shape for consistency, but no false-success and out of this PR's stated scope.

## risk

Low. One added guard mirroring a shipped sibling resolver; ships to `@vultisig/core-chain` (flagship). Strictly-more-correct: a malformed body now triggers hash verification instead of a silent success; a well-formed accept (`result` present) is unaffected. My commit adds only a test.

## verdict

**APPROVE.** The malformed-response false-success is closed, the guard is red-on-revert non-inert, and a malformed/failed bittensor broadcast **cannot be silently re-sent** via the transient-retry path (message provably non-transient + verify gates the throw on not-on-chain + now locked by an interlock test). Matches the #1426/#1430 class template, adapted correctly for substrate's pool-acceptance broadcast surface. The two Codex Mediums are non-blocking hardening shared with the polkadot sibling; Codex's "tests not red-on-revert" claim was empirically incorrect.

🤖 Two-lane review (Claude Opus 4.8 + Codex gpt-5.4). Generated with [Claude Code](https://claude.com/claude-code).

PR #1433, opened 2026-07-20
fix(tron): return tx hash string from broadcast resolver, not the rpc envelope

gomesalexandre:
R2 audit P3 hygiene. Return-shape consistency.

## what

`broadcastTronTx` returned the full `{ txid, result, code, message }` RPC envelope, while every other broadcast resolver returns a hash string or void.

## how

`return result.txid`. **Confirmed safe via grep:** the SDK's `BroadcastService.broadcastTx` discards the core broadcast resolver's return (`await coreBroadcastTx(...)` — not captured) and derives the hash itself via `getTxHash(signingOutput)`; no other consumer captures the resolver return. So this is a pure shape-consistency normalization with zero functional impact.

## testing

`tsc` + eslint clean. No behavior change (return was discarded by all consumers). Changeset: `core-chain` patch.

## risk

Minimal — cosmetic consistency; the return value was already unused.

🤖 Agent-generated

## Summary by CodeRabbit

- **Bug Fixes**
  - Tron transaction broadcasts now return the transaction hash consistently with other supported networks.
  - Preserved existing error handling and transaction verification behavior.

- **Chores**
  - Updated patch release metadata for the core chain and SDK packages.


neavra (approved):
lgtm — approving. genuinely a no-op shape fix, verified.

the claim in the description — "BroadcastService discards this return and derives the hash itself, so no consumer reads the object" — checks out end to end:
- `BroadcastService.ts:116` calls `await coreBroadcastTx({ chain, tx })` and **discards the result** (unassigned); the hash comes from `getTxHash({ chain, tx: signingOutput })` at `:121`, not from the resolver return.
- `BroadcastTxResolver` returns `Promise<unknown>`, so `result.txid` (string) vs `result` (object) is not a type break.
- that discarded call is the **only** consumer of the core broadcast dispatcher anywhere in the monorepo — the `broadcastTronTx` in `sdk/src/chains/tron/rpc.ts` is a different function (`/wallet/broadcasthex`), unrelated. no test asserts the resolver's returned shape.

so `return result` → `return result.txid` changes nothing observable. failure classification is untouched (`result.result === false || result.code` → throw → `catch` → `verifyBroadcastByHash`, all unchanged), so both-directions truthfulness and the transientRetry behavior are exactly as before. CI green on this head.

non-blocking:
- changeset declares only `@vultisig/core-chain: patch`. as on the other core-chain siblings, sdk bundles core-chain with no manifest dep so this won't version-bump `@vultisig/sdk` on its own — lowest stakes here since the change is behaviorally inert, but for consistency i'd still add `@vultisig/sdk: patch` (matching #1316).
- out of scope for a shape fix, but flagging since it's the family: this resolver only classifies the *broadcast* result (`result.result === false || result.code` = mempool rejection). a tron tx accepted into the mempool that then reverts on-chain (OUT_OF_ENERGY / contract revert) isn't caught here — same broadcast-vs-execution shape as qbtc (#1429). fine if the status resolver is the intended authority on execution outcome (the solana/#874 posture); just confirming that's deliberate rather than another door.

(noting it's a draft — approving the change as-is; no code concerns.) reviewed at c74c9438.

packages/core/chain/tx/broadcast/resolvers/tron.ts
```diff
@@ -21,7 +21,10 @@ export const broadcastTronTx: BroadcastTxResolver<OtherChain.Tron> = async ({ ch
       throw new Error(`Tron broadcast failed: ${msg}`)
     }
 
-    return result
+    // Return the tx hash string, consistent with the other broadcast resolvers (which return a hash
+    // or void) rather than the full RPC envelope. The SDK's BroadcastService discards this return and
+    // derives the hash itself, so no consumer reads the object — this is a shape-consistency fix.
+    return result.txid
```
NeOMakinG, line 27: `preferably-blocking:` this line can only be reached through a guard that fails OPEN, and it replaces the last piece of evidence with `undefined`.

The guard above is `result.result === false || result.code` — throw only on an *explicit* failure marker. Live-probed the configured endpoint (`tron-rpc.publicnode.com`, per `chains/tron/config.ts:1`; identical on trongrid):

```
POST /wallet/broadcasttransaction  body {}      -> HTTP 200 {"Error":"class java.lang.NullPointerException : null"}
POST /wallet/broadcasttransaction  body garbage -> HTTP 200 {"Error":"class org.tron.json.JSONException : ..."}
```

No `result`, no `code`, capital-`E` `Error`, **HTTP 200** so `assertFetchResponse` stays quiet. The guard misses it, the resolver returns normally, and `BroadcastService.ts:116-121` hands the caller `getTxHash(signingOutput)` as a success. Reproduced against the real resolver with `queryUrl` mocked: `verifyBroadcastByHash` is never called for either `{Error:...}` or `{}`.

Pre-PR the caller at least got `{Error: "..."}` back. Post-PR it is `undefined` — indistinguishable from the `verifyBroadcastByHash`-swallowed duplicate path, and `./tx/broadcast/resolvers/tron` is a published deep export of a public package typed `Promise<unknown>`, so no external consumer gets a type error either.

The fix already exists in this repo at `packages/sdk/src/vault/services/RawBroadcastService.ts:471`:

```suggestion
    // Return the tx hash string, consistent with the other broadcast resolvers (which return a hash
    // or void) rather than the full RPC envelope. Asserting `txid` is present keeps this fail-CLOSED:
    // TronGrid answers HTTP 200 with a bare `{Error: "..."}` (no `result`, no `code`) for bodies it
    // cannot process, which the failure enumeration above does not catch.
    if (!result.txid) {
      throw new Error(`Tron broadcast failed: ${(result as { Error?: string }).Error ?? 'no txid returned'}`)
    }

    return result.txid
```

NeOMakinG (commented):
Reviewed as **DRAFT** — posting `COMMENT` only, no approval and no fix pushed to your branch (draft PRs stay the author's). neavra already approved; nothing below contradicts the one-line change itself, it is about what the change does to the *evidence* on a fund-critical path.

The one-line change is correct on its own terms. But the PR's framing — *"pure shape-consistency normalization with zero functional impact"* — does not survive contact with the live RPC, and the file it touches has a reachable false-success hole that this change makes strictly harder to observe.

## testing approach

- Worktree at HEAD `c74c9438` (`/tmp/wt-sdk-1433`), node_modules symlinked from the canonical checkout.
- **Live probe of the actually-configured endpoint.** `tronRpcUrl` is `https://tron-rpc.publicnode.com` (`packages/core/chain/chains/tron/config.ts:1`). POSTed four bodies to `/wallet/broadcasttransaction` on both publicnode and trongrid.
- **6-case probe suite** written against the real resolver with `queryUrl` + `verifyBroadcastByHash` mocked, run under `.config/vitest.core.config.ts`. 6/6 asserted behaviours confirmed. Probe deleted afterwards — nothing left in the tree.
- Consumer grep across `packages/`, `clients/`, `examples/`, `tests/` in this repo **and** `vultiagent-app`.
- CI: all 5 checks green — but see the note under `## risk`, green is not evidence here.
- Codex second-opinion lane ran (alive); its independent finding is folded into `should-fix` #1 below.

## blocking

None on the diffed line. The item below is a `preferably-blocking` because it is pre-existing — but the PR is the change that removes the last signal.

## preferably-blocking

**1. `broadcasttransaction` fail-OPENs on the one catch-all failure shape the live node actually emits — and this PR replaces that evidence with `undefined`.**

The guard at `resolvers/tron.ts:17` is `result.result === false || result.code` — i.e. *throw only on an explicit failure marker*. The class rule is the inverse: **throw unless the response is explicitly a success.**

Live probe, HTTP **200** in every case, on the configured `tron-rpc.publicnode.com` (identical on `api.trongrid.io`):

```
POST /wallet/broadcasttransaction  body: {}
  HTTP 200  {"Error":"class java.lang.NullPointerException : null"}
POST /wallet/broadcasttransaction  body: garbage
  HTTP 200  {"Error":"class org.tron.json.JSONException : Unrecognized token 'garbage' ..."}
POST /wallet/broadcasttransaction  body: {"raw_data":{},"signature":["00"]}
  HTTP 200  {"Error":"class java.lang.NullPointerException : null"}
```

No `result` member. No `code` member. Capital-`E` `Error` only. `assertFetchResponse` does not throw (it is a 200), the guard does not fire, and the resolver **returns normally**. `BroadcastService.broadcastTx` then does `await coreBroadcastTx(...)` and immediately `txHash = await getTxHash({ chain, tx: signingOutput })` (`packages/sdk/src/vault/services/BroadcastService.ts:116-121`) — so the caller is handed a tx hash for a transaction the node refused outright. `verifyBroadcastByHash` never runs, because nothing threw.

Reproduced against the real resolver (mocked `queryUrl`, both cases pass ⇒ both behaviours are real):

```
P1 bare {Error:...} 200 body   -> resolves, verifyBroadcastByHash NOT called, returns undefined
P2 empty 200 body {}           -> resolves, verifyBroadcastByHash NOT called, returns undefined
```

This is pre-existing. What the PR changes is that **pre-PR the caller at least got `{Error: "..."}` back** — the envelope was the only remaining trace. Post-PR it is `undefined`, indistinguishable from the `verifyBroadcastByHash`-swallowed duplicate path (which also returns `undefined`). And this is not an internal-only surface: `@vultisig/core-chain` is public (no `private`, v2.26.0) and its `exports` map exposes `./tx/broadcast/resolvers/tron` as a deep subpath, so an external consumer that was reading `.txid`/`.Error` off the return silently gets `undefined` — `BroadcastTxResolver` is typed `Promise<unknown>` (`tx/broadcast/resolver.ts:6`), so `tsc` catches nothing on either side.

The fix already exists 400 lines away in this same repo — `RawBroadcastService.broadcastTronRawTx` does `if (!response.txid) throw new Error('Tron broadcast did not return transaction ID')` (`packages/sdk/src/vault/services/RawBroadcastService.ts:471-473`). Mirror it, and the return-a-string change becomes genuinely safe rather than evidence-destroying:

```ts
if (result.result === false || result.code || !result.txid) {
  const msg = result.message
    ? Buffer.from(result.message, 'hex').toString('utf8')
    : (result.code ?? (result as { Error?: string }).Error ?? 'Unknown error')
  throw new Error(`Tron broadcast failed: ${msg}`)
}

return result.txid
```

Adding `!result.txid` alone closes P1 and P2 and makes `return result.txid` unable to return `undefined`. Adding `Error`/`error` to the message extraction makes the thrown text useful instead of `Unknown error`. Note `packages/sdk/src/chains/tron/rpc.ts:349-355` already handles exactly this shape on the `/wallet/broadcasthex` leg, with a comment documenting it — the core resolver is the one leg of three that does not.

## should-fix

**1. The changeset and PR description overstate the safety of the change (Codex lane, independently).** "Confirmed safe via grep" + `patch` is right for in-tree consumers (I verified: `BroadcastService.ts:116` discards the return, and it is the only in-repo caller of `coreBroadcastTx`) — but `./tx/broadcast/resolvers/tron` is a **published** deep export of a public package. A return-type change on a published entry point is a contract change; the changeset text should say so rather than "no consumer read the object". Please reword the changeset to scope the claim to in-tree consumers.

**2. `code: 'SUCCESS'` is treated as a failure by this resolver — and by neither of its two siblings.** `result.code` is truthy-tested with no `!== 'SUCCESS'` exemption. Both other Tron broadcast legs in this repo explicitly carve it out — `packages/sdk/src/chains/tron/rpc.ts:355` (`res.code && res.code !== 'SUCCESS'`) and `packages/sdk/src/vault/services/RawBroadcastService.ts:464` (`response.code && response.code !== 'SUCCESS'`). Two independent defences against the same value is decent evidence it is emitted. Probe confirms the core resolver throws on `{result:true, code:'SUCCESS', txid:'abc123'}`. It self-heals — `verifyBroadcastByHash` will find the tx and swallow — but at the cost of two extra RPC round-trips per send, and it turns into a hard user-visible failure whenever `gettransactioninfobyid` is down or the tx is not yet indexed. Add `&& result.code !== 'SUCCESS'`.

**3. The resolver throws a bare `Error` carrying chain-controlled text into the transient-retry classifier.** `broadcastTx` wraps every non-EVM/non-Solana resolver in `withTransientBroadcastRetry` (`tx/broadcast/index.ts:34-44`), and `verifyBroadcastByHash` re-throws the *original* error, so the bare `Error` escapes. `isTransientBroadcastError` then message-regexes it against `/\babort(?:ed)?\b/i`, `/\btimed out\b/i` etc. The `message` field here is hex-decoded **chain-controlled text** (a TRC-20 revert reason is arbitrary). Reproduced:

```
{result:false, code:'CONTRACT_VALIDATE_ERROR',
 message: hex('contract validate error : transfer aborted')}
  -> err.name === 'Error'  ->  isTransientBroadcastError(err) === true  -> re-broadcast up to 3x
```

Lower severity than the Cosmos/Sui siblings because re-sending identical Tron bytes is idempotent (same `txID`, the retry just collects `DUP_TRANSACTION_ERROR`) — so I am not calling it a blocker. But it is the same class #1426/#1430 closed, the fix is one word, and it removes a chain-controlled input to a control-flow decision: throw `DeliverTxFailedError` (from `./transientRetry`) instead of `Error`. `isTransientBroadcastError` short-circuits on the `instanceof` before any regex runs.

**4. No test exists for this resolver at all.** `packages/core/chain/tx/broadcast/resolvers/` has `cardano.test.ts`, `cosmos.test.ts`, `polkadot.test.ts`, `ripple.test.ts`, `solana.test.ts`, `utxo.test.ts` — and no `tron.test.ts`. `index.test.ts` mocks the Tron resolver out entirely (`vi.mock('./resolvers/tron', ...)`). So **the 5 green CI checks contain exactly zero coverage of the line this PR changes**, and the same is true of any future regression on this file. A `tron.test.ts` pinning the 6 probe shapes above (`{Error}` 200, `{}`, `result:false`, `code:'SUCCESS'`, hex-message decode, success returns the bare hex txid) would be a proportionate ask for a fund-critical broadcast leg.

## q

**1.** `RawBroadcastService.broadcastTronRawTx` — asked specifically, answering specifically: **it does NOT have the same bug.** It already throws on a missing `txid` (`:471`), which closes the `{Error:...}` and `{}` shapes that the core resolver lets through, and it returns a string already. The one residual gap there is that it never tests `response.result === false`: a `{result:false, txid:'...'}` body with no `code` would be returned as a success. I could not produce that shape from the live node (every failure I could induce came back as `{Error:...}` with no `result` at all), so this is a shape-completeness note, not a demonstrated hole. Worth a one-line `|| response.result === false` while you are in the area, but not part of this PR.

**2.** `txid` is returned as bare lowercase hex with no `0x` prefix (probe P6: `'ab'.repeat(32)` in, same out). That matches what `vultiagent-app` already consumes from the *other* `broadcastTronTx` (`src/services/tronTx.ts:106,273,389` do `result.txid ?? txId`) and what Tronscan expects, so no prefix inconsistency. Flagging only that **two different exported functions in this repo are both named `broadcastTronTx`** — the core resolver you changed, and `packages/sdk/src/chains/tron/rpc.ts:341` which the app reads `.txid` off at three call sites. A future "shape-consistency" pass on the *other* one would break the app with no type error. Worth a comment on one of them.

## risk

Low for the diffed line in isolation — the in-tree return is genuinely discarded, and I verified that. Medium for the file: the change lands on a resolver with zero test coverage, a live-reproduced fail-open path, and a public deep export. CI green here means nothing about this code — `index.test.ts` mocks the resolver away and no `tron.test.ts` exists, so the whole 5-check suite is silent on `resolvers/tron.ts`.

## verdict

`COMMENT` — not approving, because it is a draft (policy: never approve a draft, however clean). The diffed line is fine. Before undrafting I would want `!result.txid` in the guard (#1) and the changeset reworded (#2/should-fix-1); the other three are cheap and in the same five lines.

---
🤖 automated review — Claude reviewer lane (+ Codex second-opinion lane). Live RPC probes and the 6-case resolver probe were run locally; the probe file was deleted and nothing was pushed to this branch.

NeOMakinG (commented):
Re-review at HEAD `1b779b7b` (round 2). The delta since our round-1 COMMENT (`c74c9438`) is a **main-merge**, not a code change: `git diff c74c9438 1b779b7b -- packages/core/chain/tx/broadcast/resolvers/tron.ts` is **empty** — the one-line change (`return result` → `return result.txid`, plus an explanatory comment) is byte-identical. The merge pulled in **#1505 (raw-broadcast fail-closed)** and **#1463 (verifyBroadcastByHash retry)**, so this re-review re-assesses our round-1 false-success finding against the merged tree.

## testing approach
- Confirmed HEAD `1b779b7b`; `resolvers/tron.ts` unchanged since `c74c9438` (empty diff).
- Traced the full merged-tree tron broadcast path: `coreBroadcastTx` → `resolvers/tron.ts` → `BroadcastService.broadcastTx` (`extractResolverTxHash ?? getTxHash`), and `verifyBroadcastByHash` (now with #1463 retry + #1505 allowlist).
- Live-probed the configured endpoint (`tronRpcUrl = https://tron-rpc.publicnode.com`, `--max-time 10`): `POST /wallet/broadcasttransaction {}` and `{"raw_data":{},"signature":["00"]}` both → **HTTP 200 `{"Error":"class java.lang.NullPointerException : null"}`**. Fail-open shape unchanged.
- Confirmed there is still **no `resolvers/tron.test.ts`** (every other resolver has one) → the changed line has zero coverage.

## prior issues — status
|---|---|---|---|---|---|
| 1 | headline | preferably-blocking | Reachable false-success in the tron broadcast path: ack-only/malformed `{Error:...}` HTTP-200 passes the `result.result===false || result.code` guard, resolver does not throw, caller gets a hash for a refused tx | ❌ **still open** | #1505 merge did **not** close it — see below. tron core resolver is on a path #1505 never touched. |
| 1 | headline (2nd clause) | preferably-blocking | The change makes the false-success "harder to observe" / "destroys the last evidence" | ⚠️ **largely neutralized in-repo** | In the merged tree `extractResolverTxHash` treats both the pre-PR `{Error:...}` object and the post-PR `undefined` identically (neither is a string, neither has a string `txid`) → both `?? getTxHash` fall back to the local hash. So the change is now **inert** wrt the in-repo hole. Residual is only external deep-export consumers (`undefined` vs `{Error:...}`). |
| 1 | should-fix 4 | should-fix | No `resolvers/tron.test.ts`; changed line has zero CI coverage | ❌ **still open** | Merge added `bittensor.test.ts` (#1472) but tron still uncovered. |
| 1 | should-fix 1/2/3 | should-fix | published-deep-export semver framing / `code:'SUCCESS'` false-failure / bare `Error` re-broadcast | ℹ️ **unchanged** | Not re-litigated; same as round 1. `code:'SUCCESS'` self-heals via the (now retry-hardened) verifyBroadcastByHash but only on the throw path. |

## preferably-blocking
**The false-success hole is STILL OPEN in the merged tree, and #1505 did not reach it.** Live-reproduced trace for an ack-only `{Error:...}` HTTP-200:

1. `queryUrl` returns `{Error:"…NullPointerException…"}` — HTTP 200, `assertFetchResponse` stays quiet (live-confirmed).
2. `resolvers/tron.ts:17` guard `result.result === false || result.code`: both `undefined` → **guard misses, no throw**.
3. Falls through to `return result.txid` → `undefined`.
4. Because nothing threw, **`verifyBroadcastByHash` never runs** — it is invoked only from the `catch` (`tron.ts:26`). The #1505 allowlist + #1463 retry safety net is therefore **architecturally unreachable** for this shape.
5. `BroadcastService.ts:149`: `extractResolverTxHash(undefined) ?? getTxHash({chain, tx})` → local hash → **caller gets a valid-looking tx hash for a tx the node refused.**

Why #1505 did not close it: its fixes are on (a) `RawBroadcastService.broadcastTronRawTx` — a **separate function/path** from the core resolver `broadcastTronTx` (`index.ts:30`); (b) `getTronTxStatus` — only reached via `getTxStatus`, only called from the **catch-only** `verifyBroadcastByHash`, which never fires here; (c) `extractResolverTxHash` — which falls back to local `getTxHash` on `undefined`. #1505's own safety proof ("`resolvers/tron.ts:17` throws before returning") only covers the **explicit** failure markers (`result:false`/`code`); the ack-only `{Error:...}`/`{}` shape has **neither** marker, so it shares the same blind spot.

Net: the PR change itself is **correct on its own terms and now inert** wrt fund safety in the merged tree (the round-1 "harder to observe" concern is neutralized by `extractResolverTxHash`'s fallback). But the **pre-existing** false-success hole in the shared guard remains, and the merge that was expected to close it does not. Fix (still one line, in the guard, not the return):
```ts
if (result.result === false || result.code || !result.txid) {
  const msg = result.message
    ? Buffer.from(result.message, 'hex').toString('utf8')
    : (result.code ?? (result as { Error?: string }).Error ?? 'Unknown error')
  throw new Error(`Tron broadcast failed: ${msg}`)
}
```
That routes the ack-only shape through `verifyBroadcastByHash` (now #1463-retried, #1505-allowlisted), which is exactly the safety net the merge added — instead of falling back to a locally-derived hash.

## should-fix
- Add `resolvers/tron.test.ts` covering: `{Error:...}`/`{}` HTTP-200 → throws (or routes to verify), `{result:true,code:'SUCCESS',txid}` → returns the hash, `{result:false,code}` → throws. The changed line currently has **zero** CI coverage (all 5 green checks `vi.mock` the tron resolver out).

## risk
Fund-adjacent: a false-success on a Tron broadcast is the false-confirm / double-spend class (composes with app#2236 TRX verifier, sdk#1505 raw fail-closed, app#2241 double-send). The PR itself neither introduces nor materially worsens it in the merged tree; the exposure is a **pre-existing** shared-guard hole that #1505 did **not** cover.

## verdict
COMMENTED (draft — not approving). The one-line change is correct and now inert wrt fund safety, so it is **approvable-once-ready** on its own terms. But our round-1 headline is **still open**: the ack-only-200 false-success hole in `resolvers/tron.ts:17` survives the #1505 merge because the tron core resolver is on a path #1505 never touched and the safety net is catch-only. Recommend folding the one-line guard fix (`|| !result.txid`) into this PR so the shape actually reaches the merged verify path.

<sub>🤖 reviewer lane: Claude (load-bearing) + live endpoint probe. Codex lane skipped this round.</sub>

PR #1435, opened 2026-07-20
Fix browser storage backend stickiness

### LINKED ISSUE #1209: fix(storage): prevent partial browser fallback from hiding vaults
## Problem

The browser storage layer can fall back from IndexedDB to another backend after partial writes or read failure. Switching backends without migrating/reconciling the full vault set makes previously stored vaults appear missing.

Evidence:
- `packages/sdk/src/storage/browser.ts:85-173`
- `packages/sdk/src/storage/indexedDb.ts:46-124`

## Acceptance criteria

- Select one backend through an explicit initialization/health check.
- Never switch a live vault store per operation.
- If migration is necessary, reconcile the complete set transactionally and verify it.
- Surface degraded storage state instead of returning a partial empty view.
- Test quota, blocked IndexedDB, partial migration, and reload recovery.

rcoderdev:
Closes #1209

## Summary by CodeRabbit

* **Bug Fixes**
  * Browser vault storage now reliably preserves the initially selected backend across reloads and avoids silent fallbacks.
  * IndexedDB/localStorage failures (including quota issues and malformed stored data) now surface explicit errors instead of returning partial state.
  * Improved legacy vault recovery and safer fail-closed behavior for blocked/unavailable IndexedDB and cross-backend inconsistencies.
* **Tests**
  * Added unit coverage for backend selection, concurrency, backend recovery, quota/serialization error handling, and legacy vault scenarios.
* **Chores**
  * Added a Changesets entry for patch releases.

### DIFF
--- a/.changeset/calm-browsers-hold.md
+++ b/.changeset/calm-browsers-hold.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Keep browser vault storage on the backend selected during initialization. IndexedDB failures and quota errors now surface without silently switching to a partial localStorage or in-memory view, while legacy localStorage vaults retain their backend across reloads.
--- a/packages/sdk/src/platforms/browser/storage.ts
+++ b/packages/sdk/src/platforms/browser/storage.ts
@@ -5,37 +5,115 @@
 import type { Storage, StorageMetadata, StoredValue } from '../../storage/types'
 import { STORAGE_VERSION, StorageError, StorageErrorCode } from '../../storage/types'
 
-type StorageMode = 'indexeddb' | 'localstorage' | 'memory'
+type StorageMode = 'indexeddb' | 'localstorage'
+
+const BACKEND_MARKER_KEY = '__vultisig_storage_backend__'
 
 export class BrowserStorage implements Storage {
   private db?: IDBDatabase
-  private mode: StorageMode = 'memory'
-  private memoryStore = new Map<string, StoredValue>()
+  private mode?: StorageMode
+  private initializationError?: StorageError
+  private readonly initialization: Promise<void>
   private readonly dbName = 'vultisig-vaults'
   private readonly storeName = 'vaults'
   private readonly dbVersion = 1
 
   constructor() {
-    this.initializeAsync().catch(err => console.warn('BrowserStorage initialization warning:', err))
+    this.initialization = this.initializeAsync().catch(error => {
+      this.initializationError =
+        error instanceof StorageError
+          ? error
+          : new StorageError(
+              StorageErrorCode.StorageUnavailable,
+              'Browser storage initialization failed',
+              error as Error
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
## lgtm 🫡

solid, and the fund-safe direction. the old BrowserStorage silently cascaded IndexedDB -> localStorage -> memory on any error, which for VAULT storage is dangerous: a quota event split-brains keys across two backends, and a memory fallback loses the vault (and fund access) the moment the tab closes. this rewrite fails CLOSED and sticks to the backend chosen at init - errors surface instead of silently switching. correct posture for key material. `platforms/browser/storage.ts`.

what I verified:
- sticky backend via a `__vultisig_storage_backend__` marker: persisted 'localstorage' never re-probes IndexedDB (test: open() not called across reload), and a transient IndexedDB open failure under a persisted 'indexeddb' marker THROWS rather than silently dropping to localStorage and hiding the IndexedDB vaults (test: recovers on reload).
- memory mode removed entirely - a fully-storage-blocked browser now fails init instead of running an ephemeral vault store. right call for funds.
- durability fix: `waitForTransaction` resolves on transaction `oncomplete`, not `request.onsuccess` (which fires pre-commit) - genuinely more correct for vault writes.
- quota now surfaces as `StorageError(QuotaExceeded)` on both backends, missing keys still return null (verified the `if (!stored) return null` guard survives), present-but-corrupt records throw typed `SerializationFailed` instead of null-swallowing.
- concurrent ops share one init promise (test: single open()), legacy unmarked-localStorage vault recovery works, and marker read/write is excluded from the SDK key listing.

tests are comprehensive (11 cases: concurrency, both-backend quota, stickiness, legacy recovery, divergence, blocked/failed IndexedDB, malformed record) via fake-indexeddb + a localStorage double.

two non-blocking notes for you (both look intentional per the changeset, just flagging):
- q: the unmarked cross-backend divergence case THROWS on every op with no automatic merge/recovery path - a user left in that split state is effectively bricked until manual cleanup. fail-closed is the safe choice over silently hiding one backend's vault, but a follow-up migration/merge would round it off. intended for now?
- suggestion: `readPersistedMode` throws if `localStorage.getItem` itself throws (SecurityError), which fails init even when IndexedDB is fully working - a narrow selectively-blocked-localStorage env. fine as fail-closed, but if you'd rather keep IndexedDB users up there, a marker-read failure could fall through to backend detection instead of aborting.

neither gates merge - core logic is correct and the direction is right. clean. NB it's BEHIND main (your merges) - NeO/rcoderdev's court to rebase before you merge.

gomesalexandre (approved):
fresh current-head stamp after the branch update.

Still good on the actual delta: the browser storage backend remains sticky, fail-closed, and avoids the old silent IndexedDB -> localStorage -> memory fallback footgun for vault material. I already verified the load-bearing behavior on the original head, and this update is just the mainline refresh needed to get it mergeable again.

lgtm 🫡

PR #1304, opened 2026-07-15
fix(cli): report non-success outcomes truthfully

neavra:
## Summary

- report failed or declined agent signing and typed blocked/refusal/error turns as `success:false` with typed non-zero exits
- fail closed when one-shot `agent ask --session` cannot resume the requested conversation, before dispatching the message
- classify anticipated vault/input/auth failures and permanent raw-transaction validation failures in their existing taxonomy slots, while preserving retryable transient broadcasts and `BROADCAST_COMMITTED` exit 13

## Validation

- red/green stash proof: the new regression tests failed with only the test changes present, then passed after restoring the implementation
- `yarn test:cli` (49 files, 718 tests)
- `yarn check:ci`
- `yarn workspace @vultisig/cli typecheck`
- changed-file ESLint and Prettier (`.config/.prettierrc`)
- `yarn build:all`
- `yarn lint`
- `yarn test`
- `yarn cli:build`
- built CLI local receipts: missing vault switch returns exit 5 / `VAULT_NOT_FOUND`; missing import file returns exit 4 / `INVALID_INPUT`

I attempted a read-only real-backend `agent ask` smoke test, but it stopped at `AUTH_FAILED` before a turn because this environment has no usable vault credentials. No signing or broadcast was attempted, and no live transactions were made.

## Summary by CodeRabbit

- **Bug Fixes**
  - CLI failures now consistently return non-success results with accurate typed exit codes.
  - Declined or failed signing, blocked agent actions, stale sessions, invalid addresses, and permanent broadcast errors are reported more accurately.
  - Completed broadcasts followed by later errors are clearly identified as committed and non-retryable.
  - Stale session IDs now fail safely without starting a new conversation.

- **Documentation**
  - Clarified JSON and human-readable error output, partial results, failure codes, and session-resumption behavior.

PR #1436, opened 2026-07-20
fix(cli): interactive decline exits 12 + #1304 test-theatre cleanup

neavra:
Two small validated CLI items, verify-first on `origin/main`. Both were alive on HEAD. No live transactions.

## Item 1 — interactive decline → exit 12 (F1)
An interactive human answering **no** at the `send` (and `execute`/`swap`) confirm prompt used to be swallowed in `index.ts` (`err.message === '… cancelled by user'` → `return`) so `withExit` exited **0** — telling a scripted caller a declined transaction had *succeeded*, while the non-interactive path already exited **12** (`confirmTransaction`→`requireInteractive`→`ConfirmationRequiredError`). This was deliberately deferred twice as out-of-scope on #1304; it is the scope now.

- `commands/transaction.ts` / `execute.ts` / `swap.ts`: a decline throws `ConfirmationRequiredError` (exit 12, `success:false` envelope) instead of a plain `Error`.
- `index.ts`: the three swallow-catches are removed, so the typed error reaches `withExit` — the interactive twin of the non-interactive refusal. Fixing only `send` would leave execute/swap at exit 0 (a *new* asymmetry), so all three move together.
- `interactive/session.ts` REPL: the decline arm now matches `instanceof ConfirmationRequiredError` — a REPL decline still returns to the prompt, never exits.

**Test (deterministic, PTY-driven):** `sendDeclinePty.test.ts` runs the real `withExit(() => sendTransaction(fakeVault, {yes:false}))` inside a `node-pty` pseudo-terminal (no network), answers "n" at the rendered prompt, and asserts exit **12** + the `success:false` `CONFIRMATION_REQUIRED` JSON envelope. (`node-pty` is a new `@vultisig/cli` **devDependency**; a `beforeAll` restores the `spawn-helper` executable bit that yarn's prebuild unpack strips — cross-platform + idempotent.) Plus `interactiveDecline.test.ts` — fast unit coverage that all three verbs throw `ConfirmationRequiredError` on decline.

## Item 2 — #1304 review follow-up #4 (test theatre)
- **M3:** the `SigningFailed` 32-byte→`INVALID_INPUT`/4 branch is the **only** mapping for a `VaultError`-wrapped 32-byte sign failure (`FastVault.signBytes` wraps the inner error; the plain-`Error` heuristic covers only non-VaultError paths) — deleting it would regress task-finding-#3 intent. The theatre was the test feeding a generic message; an honest test now exercises the branch. Kept production.
- **M4:** an ask-mode `--session` resume whose auth failure survives `withAuthRetry` was mis-typed `SESSION_NOT_FOUND` (exit 5); it now fails closed as `AUTH_FAILED` (exit **2**) via `isAuthError`, with an honest test. Any other resume failure stays `SESSION_NOT_FOUND`.
- **M5:** `ask.ts`'s `initError`/`hasAsked` machinery is unreachable (`AskInterface` is always `askMode`, whose `initialize()` throws rather than firing `onError` before the first `ask()`) — removed it and deleted the three theatre tests that drove it via a synthetic `onError`.

## Validation
- Red-then-green: M3, M4, the send throw, and the PTY test (table+json) each go red when only their production change is reverted; green restored.
- `yarn test:cli` **872 pass**; `yarn typecheck` clean; changed-file `prettier`/`eslint`/`knip` clean.
- Changeset `@vultisig/cli` **patch** — spells out the scripted-consumer exit change (a decline flips `$?` 0→12), following #1304's precedent.

Kept **DRAFT**; nothing else posted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Declining interactive `send`, `execute`, or `swap` confirmations now exits with code **12** (instead of succeeding) and returns `success: false` with `CONFIRMATION_REQUIRED`.
  * The interactive shell continues handling declines by returning safely to the prompt.
  * When resuming a session in ask mode, persistent authentication failures are now reported as **AUTH_FAILED** (rather than session missing).
  * Signing failures caused by wrong-length input are now classified as **invalid input** (non-retryable).


gomesalexandre (approved):
## lgtm 🫡

Item 1 is the important one and it's the right fix: a swallowed interactive decline exiting 0 was telling a scripted/headless caller a *declined* transaction had *succeeded* (a false-success on `$?`), now correctly exits 12 `CONFIRMATION_REQUIRED` matching the non-interactive refusal - and the REPL-decline-returns-to-prompt path is preserved. The documented `0 -> 12` behavior change for scripted consumers is the correct direction (the old 0 was the bug). Item 2 (`--session` fail-closed AUTH_FAILED exit 2 vs SESSION_NOT_FOUND exit 5) is the right fail-closed classification so a headless caller re-auths instead of trusting a stale-looking valid session. verify-first-on-main + the decline/pty/ask tests back both. clean.

PR #1437, opened 2026-07-20
fix(cli): classify Cosmos DeliverTx failure honestly and journal its hash

neavra:
## What

Closes the #1316 review's **M1** and **M2** follow-ups in the CLI. A Cosmos transaction can be *included in a block and still fail execution* (DeliverTx `code !== 0` — out-of-gas, a wasm revert, a THORChain/Maya deposit-handler rejection). #1316 made the SDK throw truthfully on that; this PR makes the CLI handle the result correctly. Layers on top of #1355's CheckTx reclassification (does **not** undo it).

The cosmjs message shape is `Error when broadcasting tx <hash> at height <N>. Code: <c>; Raw log: <log>`, thrown as `DeliverTxFailedError` and wrapped by `BroadcastService` as `Failed to broadcast transaction on <Chain>: …`.

### M1 — classify honestly (`core/errors.ts`)
The DeliverTx-failure message did not match the CheckTx-shaped classifier (`COSMOS_BROADCAST_FAILURE_RE`), so it fell through to `ExternalServiceError` (exit 6, **retryable**) with a *"the node may be temporarily unavailable / Retry"* hint — for a transaction that is on-chain and **permanently** failed. It now classifies as `InvalidInputError` (exit 4, **non-retryable**) with an honest message: the tx executed and failed on-chain, the account sequence is consumed and the gas is spent, so re-broadcasting the identical signed bytes cannot succeed.

This is non-retryable **regardless of the SDK code** — the deliberate opposite of the CheckTx path, where #1355 keeps e.g. code 5 (`insufficient funds`) retryable *because a CheckTx rejection never touched the chain*. A DeliverTx failure did.

### M2 — journal the hash (`core/broadcastGuard.ts`)
On the direct `send`/`swap` guard, a DeliverTx failure previously journaled **nothing** (the throw precedes hash computation), discarding a hash we definitively have. `guardedBroadcast` now records the hash cosmjs embedded in the message as a broadcast **plus a `failed` resolution**, so the terminal tx is on the record (tx-status / explorer / audit) while the guard correctly **re-opens** for a legitimate fresh-sequence rebuild — recording `failed` is exactly what clears the intent (`isRetryableResolution`). The tx is terminal, so re-opening is correct.

A tiny shared matcher (`core/cosmosDeliverTx.ts`) holds the single regex/hash-extractor used by both paths; it anchors on the `at height <digits>. Code: <digits>; Raw log:` skeleton (not the chain-controlled hash/rawLog) and takes the first match, so a rawLog echoing the phrasing can't spoof it.

## Why `InvalidInputError` (no new exit code)
No honest dedicated class exists — `ACK_FAILED` (8) asserts the broadcast *succeeded*; `BROADCAST_COMMITTED` (13) is an agent-ask *"may be incomplete"* partial-success. Both are dishonest for a definite on-chain failure. Exit 4 is the same family as the sibling permanent-broadcast path; the honesty is carried by the tailored hint and the preserved cosmjs message (hash/height/code/rawLog).

## Scope
- The raw `broadcast` command (RawBroadcastService, #1426) has no journal wiring, so **M2 doesn't apply there**; M1's classifier still handles its DeliverTx throw (same message shape).

## Tests (red-then-green)
- `core/errors.test.ts` (+3): non-retryable classification; honest on-chain message (not the transient-node hint / no "Retry the transaction"); **DeliverTx code 5 permanent while CheckTx code 5 stays retryable**.
- `core/broadcastGuard.test.ts` (new): journals the hash as a `failed` resolution and re-throws on a DeliverTx failure; no-ops on other throws; unchanged happy path.
- Full `yarn test:cli` → **63 files / 884 tests pass** (after `yarn build:all` so the subprocess suites can resolve the SDK node bundle). `@vultisig/cli` typecheck, ESLint, and Prettier clean.

## Changeset
`@vultisig/cli` patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cosmos transactions that fail during on-chain execution are now reported as non-retryable input errors instead of temporary service failures.
  * Error messages now direct you to inspect the transaction using its hash and avoid re-broadcasting identical transaction data.
  * Failed Cosmos transactions are now recorded in the broadcast journal with their on-chain transaction hash.
* **Tests**
  * Added coverage for Cosmos failure classification, journaling, hash validation, and retry behavior.


gomesalexandre (approved):
## lgtm 🫡

The classification flip is exactly right - a DeliverTx `code !== 0` is on-chain and FAILED, so EXTERNAL_SERVICE/exit-6/"retry" was actively dangerous (re-broadcasting identical bytes for a tx the chain already rejected); INVALID_INPUT/exit-4 with "inspect the hash, rebuild, don't re-broadcast" is the honest + safe classification. Resolving the chain from the SDK broadcast wrapper (not the wrapped payload text) is the load-bearing bit - a foreign chain's program log can't spoof the DeliverTx marker and strand a genuinely-retryable error. And journaling the hash cosmjs embedded (that we definitively have) so a failed included tx is on the record while the guard re-opens for a legit retry closes M2 cleanly. Real matcher stays unmocked = the thing under test. Layers correctly on #1355's CheckTx reclassification. Clean.

PR #1449, opened 2026-07-20
fix(keysign): sort message hashes so initiator and joiners agree on order

ahdzib-maya:
Fixes intermittent `HTTP 404: Not Found` failures when co-signing a multi-message keysign (e.g. an ERC-20 approve + swap).

## Problem

`TransactionBuilder.extractMessageHashes()` returns message hashes in resolver order, and the initiating device signs them in that order (`RelaySigningService`). Every joiner, however, signs them **sorted**:

- this repo — `keysign/cosigner.ts`
- Android — `SigningHelper.getKeysignMessages()` -> `messages.sorted()`
- extension — `useKeysignMutation` -> `.sort()`

When a payload produced hashes that were not already in lexicographic order (e.g. `approveHash > swapHash`), the initiator and the joiner drove **different messages at the same step**. The joiner polled `GET /setup-message/{sessionId}` for a hash the initiator had never uploaded, and failed with `HTTP 404: Not Found` after exhausting its retries.

Because it depends purely on how the two hashes happen to sort, this was ~50/50 per payload — the same swap could succeed, then fail on a retry. Single-message keysigns (plain transfers) cannot diverge, which is why only approve+swap style flows were affected.

## Fix

Sort the hashes at the source so the initiator agrees with every joiner.

## Why this is safe

- The signing loop, `messagesToSign`, and the signature pairing all read this same array, so they stay mutually consistent.
- Signatures are matched back to transactions **by message hash, not by position**: `compileTx` looks each signature up by hex hash (same for the PSBT and QBTC paths), so per-input mapping — including UTXO multi-input — is unaffected by array order.
- The only positional zip (`convertSignature.ts`) has both sides derived from this same sorted array.
- `Array.prototype.sort()` here is lexicographic over lowercase hex, matching the joiners' ordering (including Kotlin `List<String>.sorted()`).

## Testing

Verified against a live 2-of-3 relay keysign: an ERC-20 approve + swap that had been reproducibly failing with `HTTP 404` on the joining device signed successfully after this change, and broadcast normally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency in multi-message signing by returning message hashes in a deterministic, sorted sequence rather than resolver-dependent order.
  * Helps initiating and joining devices follow the same message sequence during keysigning flows.

* **Tests**
  * Added unit tests to verify deterministic hash ordering for both initiator and joiner scenarios, including cases where emitted messages arrive in a non-sorted order.


gomesalexandre (commented):
## nice catch ser

verified the fix is correct, not just plausible. traced the joiner side: `packages/core/mpc/keysign/cosigner.ts:166-170` builds its hashes the identical way you do on the initiator:

```ts
const msgs = inputs
  .flatMap(txInputData => getPreSigningHashes({...}).map(h => Buffer.from(h).toString('hex')))
  .sort()
```

so both sides produce the SAME lowercase-hex, no-0x-prefix strings and both run a bare `.sort()` (default lexicographic). your `allMessageHashes.sort()` makes the initiator agree byte-for-byte with the joiner (and the Android/extension `.sort()` you cite sort the same ASCII-hex form, so all four converge). the approve+swap 404 you describe (joiner polling `/setup-message` for a hash the initiator never uploaded on `approveHash > swapHash`) checks out as the exact failure mode. single-message keysigns being unaffected fits too.

one thing before i stamp: this is the signing path and it ships to every consumer (ios/android/windows + agent), so i'd like a regression test pinning the order, otherwise a future refactor silently reintroduces the divergence and it only shows up as a prod approve+swap hang. something like: a multi-message payload whose resolver order is NOT sorted (approveHash lexicographically after swapHash) -> assert `extractMessageHashes` returns them sorted, matching cosigner's order.

happy to push that test onto your branch myself (maintainer-edit is on) if you'd rather not — just say the word, otherwise add it and i'll approve on green. either way the fix itself is right.

gomesalexandre (approved):
## lgtm 🫡

verified end to end, not just eyeballed:
- correctness: initiator's `allMessageHashes.sort()` now agrees byte-for-byte with the joiner (`packages/core/mpc/keysign/cosigner.ts:166-170` builds the same lowercase-hex via `Buffer.from(h).toString('hex')` + bare `.sort()`), so the approve+swap `/setup-message` 404 hang is genuinely closed. all four signers (initiator, cosigner, android, extension) converge on the same lexicographic order.
- test pins it: the fixture is deliberately built out of sorted order (approve `ff…` after swap `00…`) and asserts sorted output matching the joiner. Unit + Integration both green.
- the follow-up format commit (119e880e) is prettier-only, confirmed logic-free (pure constructor re-wrap, +6/-3).

nice fix ser. armed auto-merge, it'll go in on green.

PR #676, opened 2026-06-10
Add THORChain limit swap memo builder

rcoderdev:
Closes #312

## Summary
- Added THORChain limit swap memo builder with validation, 12/24/72 hour expiry windows, and JSON vectors.
- Exported the builder and fixture JSON from @vultisig/core-chain for package consumers.
- Added a root shell-quote resolution to 1.8.4 intentionally because local check:ci quality:audit fails on the existing transitive 1.8.3 via concurrently; 1.8.4 is the first patched/current npm release for GHSA-w7jw-789q-3m8p / CVE-2026-9277.

## Checks
- yarn check:ci
- yarn node package export smoke test for @vultisig/core-chain/swap/native/limitSwapMemo and fixture JSON

## Summary by CodeRabbit

* **New Features**
  * Introduced THORChain limit swap functionality including automatic memo generation, comprehensive input validation for destination addresses, amounts, prices, and expiry windows, plus support for affiliate memo integration with automatic fallback handling.

* **Tests**
  * Added comprehensive test suite with fixture data covering multiple limit swap scenarios across different asset pairs and blockchain networks, including validation rules and edge cases.

* **Chores**
  * Updated package exports and dependencies.

PR #1451, opened 2026-07-20
chore(deps): bump brace-expansion + js-yaml + shell-quote + tar out of the redos/advisory ranges (unblocks check:ci)

gomesalexandre:
Unblocks the `quality:audit` step of `check:ci` (the "Lint and Type Check" job) on every open PR — it runs a full `yarn npm audit --recursive --all --severity high` in one shot, so all flagged advisories have to clear together for the gate to go green. A batch of four high/critical advisories published 2026-07-20 20:51-21:52 UTC turned the gate red repo-wide. Confirmed the identical failure on #1449 and #1256 after merging main in.

## what

| package | before | after | advisory |
|---|---|---|---|
| `brace-expansion` (via `minimatch@3.1.5`) | `1.1.14` | `1.1.16` | [GHSA-3jxr-9vmj-r5cp](https://github.com/advisories/GHSA-3jxr-9vmj-r5cp) — ReDoS |
| `brace-expansion` (via `minimatch@10.2.5`) | `5.0.5` | `5.0.7` | same |
| `brace-expansion` (via `minimatch@5.1.9`/`9.0.9`) | `2.1.0` | `2.1.2` | already outside the vulnerable range, bumped along |
| `js-yaml` (4.x line) | `4.1.1` | `4.3.0` | [GHSA-52cp-r559-cp3m](https://github.com/advisories/GHSA-52cp-r559-cp3m) — YAML merge-key quadratic CPU |
| `js-yaml` (3.x line) | `3.14.2` | `3.15.0` | same |
| `shell-quote` | `1.8.4` | `1.10.0` | [GHSA-395f-4hp3-45gv](https://github.com/advisories/GHSA-395f-4hp3-45gv) — quadratic DoS in `parse()` |
| `tar` | `7.5.13` | `7.5.20` | [GHSA-23hp-3jrh-7fpw](https://github.com/advisories/GHSA-23hp-3jrh-7fpw) (critical) + [GHSA-8x88-c5mf-7j5w](https://github.com/advisories/GHSA-8x88-c5mf-7j5w) |

`brace-expansion` and `tar` cleared entirely via `yarn up -R`, staying within each consumer's existing semver range, no `minimatch` majors moved.

`js-yaml` and `shell-quote` each had a consumer pinning an **exact** version (not a range), so `yarn up -R` alone couldn't move them — needed two scoped `resolutions`:
- `markdownlint-cli2` pins `js-yaml` to exact `4.1.1` — added `"js-yaml@npm:4.1.1": "^4.3.0"`.
- `concurrently` pins `shell-quote` to exact `1.8.4` via a resolution already in `package.json` (added in #676, for an older CVE — GHSA-w7jw-789q-3m8p — 1.8.4 was the patched version at that time). Bumped that existing `"shell-quote"` resolution forward to `1.10.0`.

`js-yaml`/`shell-quote`/`tar` are all transitive tooling deps only (`node-gyp`, `electron-builder`, `changesets`, eslint config loading, `markdownlint-cli2`, `textlint`, `xmlbuilder2`) — never in the SDK/CLI runtime bundle.

Diff is `package.json` (2 resolution lines) + `yarn.lock` only, no source changes.

## verification

- `yarn install --immutable` clean
- `yarn npm audit --recursive --all --severity high` — **0 flagged entries across all four clusters** (`YN0001: No audit suggestions`, exit 0)
- `yarn lint` / `yarn format:check` / `yarn knip` — all pass
- `yarn typecheck` — fails identically on this branch and on a clean `main` checkout (pre-existing `@vultisig/mpc-types`/`@vultisig/mpc-wasm` codegen build-order dependency, unrelated to this change — `check:ci` runs `build:all` before `typecheck`, which generates those, so it passes in CI)

🤖 Agent-generated


yarn.lock
```diff
@@ -8522,30 +8522,30 @@ __metadata:
...
     concat-map: "npm:0.0.1"
-  checksum: 10c0/b6fdac832bc4e36a753658c9ed052c2e1a2be221763b002df25d1efbf7d21724334e726a6cd5eadc72a4b19ec3efb632d629cc003bc9c62f7af7a7915ffa4385
+  checksum: 10c0/b2a915bbedbf4e45840d1fb9a4d391bbf26a79475bd134714d3cee34f1f0edb0ce982738028843be5fbaf8039429f71fa487df8c915b6065ced542c83e58fae6
   languageName: node
   linkType: hard
 
 "brace-expansion@npm:^2.0.1, brace-expansion@npm:^2.0.2":
-  version: 2.1.0
-  resolution: "brace-expansion@npm:2.1.0"
+  version: 2.1.2
```
NeOMakinG, line 8535: `q:` description-accuracy only, the change itself is right. The body says 2.1.0 was "already outside the vulnerable range, bumped along for good measure" — that is not the case. GHSA-3jxr-9vmj-r5cp lists `>=2.0.0 <2.1.2` as vulnerable, so 2.1.0 was squarely inside it, and `yarn npm audit` on `main` flags this exact entry (tree version 2.1.0) as one of the three brace-expansion hits. I also confirmed the advisory was not widened after the fact — `updated_at` is 2 seconds after `published_at`.

I reproduced the exponential on 2.1.0 specifically: `'a' + '{},'.repeat(26) + 'b'` takes **11.2s** on 2.1.0 and **0ms** on 2.1.2. So this line is a necessary part of the fix, not a courtesy bump — worth correcting in the description so nobody later concludes the 2.x line was optional and reverts it.

yarn.lock
```diff
@@ -8522,30 +8522,30 @@ __metadata:
...
     balanced-match: "npm:^1.0.0"
-  checksum: 10c0/439cedf3e23d7993b37919f1d6fdc653ec21a42437ec3e7460bea9ca8b17edf7a24a633273c31d61aa4335877cf29a443f1871814131c87997a1e6223e1f1502
+  checksum: 10c0/5442ecab84045d21826268bc56c81a6ef4215327ee1f5ee153f67928e93f7a915d130ecec9966623143277fa3cce036ebb50020024bcc4d78853cdd6af9a19f8
   languageName: node
   linkType: hard
 
 "brace-expansion@npm:^5.0.5":
-  version: 5.0.5
-  resolution: "brace-expansion@npm:5.0.5"
+  version: 5.0.7
```
NeOMakinG, line 8544: `praise:` verified this one carefully because 5.x is a young major and the 5.0.5→5.0.7 source diff is more than a regex tweak — it restructures `expand_` from recursion into a loop and, notably, adds `N.length < max` to the sequence loop, which 5.0.5 did not have.

That new cap is the one place this bump could have changed observable behaviour, which would matter because this entry feeds `minimatch@10.2.5` and therefore the glob sets that eslint/knip/vitest resolve. It does not: expansion output is identical across 1479 patterns and at large sequences (`{1..100000}`, `x{1..70000}y` — 100000 and 70000 results on both), so the cap is not reachable at the default `max`. `balanced-match: ^4.0.2` is also correct — unchanged between 5.0.5 and 5.0.7.

NeOMakinG (approved):
Lockfile-only bump of `brace-expansion` onto the patched line for GHSA-3jxr-9vmj-r5cp. Treated as guilty-until-proven-safe per the repo's dependency-bump policy rather than waved through, and it comes out clean: the change is correct, minimal, semantically inert, and strictly reduces attack surface. **Approving.** One framing issue below that is worth knowing before anyone treats this as "CI is fixed".

## testing approach

Two worktrees off `/` — PR `214450ac` and a clean `origin/main` (`a8c10f02`); the branch is exactly main + 1 commit touching one file (`git diff --stat origin/main...pr1451`).

- **Scope**: `yarn.lock` only, +9/-9, three `brace-expansion` blocks and their checksums. No `package.json`, no `resolutions`, no `minimatch` entry touched — verified from the diff, not the description.
- **Advisory ranges** pulled from the GitHub advisory API rather than the PR table: `<1.1.16`, `>=2.0.0 <2.1.2`, `>=3.0.0 <5.0.7`. All three new versions land exactly on `first_patched_version`.
- **Behavioral equivalence** (the check that actually matters here — `brace-expansion` feeds `minimatch` → glob → which files eslint/knip/vitest/tsconfig include, so output drift would silently change the linted and tested file set): installed all six versions side by side and diffed `expand()` output over **1479 patterns × 3 version pairs → 0 divergences**, including large sequences (`{1..100000}`, `x{1..70000}y`). Reading the source diff also turned up a genuine new `N.length < max` cap in the 5.x sequence loop — it is not reachable at the default `max`, which is why the large-sequence outputs match.
- **ReDoS reproduced and confirmed fixed**, payload `'a' + '{},'.repeat(n) + 'b'`:

  | n | 2.1.0 / 5.0.5 | 2.1.2 / 5.0.7 |
  |---|---|---|
  | 15 | 14 ms | 0 ms |
  | 20 | 180 ms | 0 ms |
  | 24 | 2.8 s | 0 ms |
  | 26 | **11.2 s** | **0 ms** |

  (≈4× per increment.) Honest limitation: **1.1.14 does not reproduce** on this payload — the 1.x branch has a different early-return shape. That bump is still correct per the advisory, but I did not independently demonstrate the 1.x exponential.
- **Not runtime-reachable**: all 15 `minimatch` consumers are tooling (eslint plugins, electron builders, typedoc, glob, typescript-estree), and `brace-expansion` is declared in no prod/dev/peer block directly. A grep for `expand_` in the built SDK bundle false-positived on noble-curves' `expand_message_xmd`/`_xof`; the real tells (`balanced-match`, `braceExpand`, `escSlash`, `concat-map`) are all absent. **No signing-path exposure** — this is build-chain only, so I make no runtime signing claim.
- `yarn install --immutable` → exit 0 with **zero lockfile churn** (the check that catches a hand-edited lockfile). Installed tree holds only 1.1.16 / 2.1.2 / 5.0.7. Declared dependency ranges are unchanged across all three lines, so the untouched `balanced-match` / `concat-map` lines are correct.
- **Audit claim verified exactly**: `brace-expansion` 3 entries → 0.
- **Typecheck claim verified exactly**: fails identically on both sides — 10 errors, exit 2, and a `diff` of the sorted error lines is **empty**. The author's build-order explanation is corroborated by the error shapes (TS2305 "has no exported member" from an unbuilt `@vultisig/sdk`), and `check:ci` does run `build:all` first.
- **Test suites, real numbers, both sides**: 503 files, **6332 passed / 5 skipped, exit 0** — identical on PR and on clean `main`. (`yarn build:all` OOMs at node's default 2 GB heap during `.d.ts` rollup; `NODE_OPTIONS=--max-old-space-size=12288` and it passes. Not a defect in this PR.)

## should-fix

**The "unblocks `check:ci`" claim was true when written and is no longer true — because `quality:audit` is a live registry query, not a property of the branch.**

`quality:audit` is `yarn npm audit --recursive --all --severity high`, wired into `check:ci`. It re-queries npm every run, so its verdict is a function of wall-clock time. The timeline on this PR:

| time (UTC, 2026-07-20) | event |
|---|---|
| 20:43:49 | main's last CI run — **success** |
| 20:51:09 | GHSA-3jxr-9vmj-r5cp publishes → every open PR goes red |
| 21:19 / 21:49 / 21:51 | js-yaml / shell-quote / tar advisories publish |
| 21:41 | #1449's failing job log shows **only** brace-expansion — confirms your stated motivation |
| **22:54:58** | **this PR's own audit step: "No audit suggestions" → job PASSED** |
| 23:14 | I re-ran the identical command: main **9** advisories, this branch **6** |

The green badge on this PR is real but not reproducible — npm's audit endpoint had not yet synced the sibling advisories when CI ran. Right now the same command on this branch still exits 1 on js-yaml ×2, shell-quote, tar ×2, **and axios**, which appeared in the intervening hour. So a fresh run would be red, and the green check here should not be read as evidence that other PRs will go green.

None of that is a defect in this diff, and you did flag the sibling advisories in the body. The concrete ask is just: don't let the team merge this believing CI is unblocked.

I also scouted the follow-up so you don't have to — **it is not a one-liner**. `yarn up -R js-yaml shell-quote tar` on top of this lockfile gets 5 → 2:

- `tar` → patched, clears.
- js-yaml caret group → 4.3.0 and `^3.6.1` → 3.15.0, clears.
- **Resists:** `markdownlint-cli2@^0.22.1` pins `js-yaml: "npm:4.1.1"` *exactly*, and `concurrently@^10.0.3` pins `shell-quote: "npm:1.8.4"` *exactly*. Exact pins are immune to `yarn up -R` — those two need a `resolutions` override, an upstream release, or the escape hatch this repo already uses: `npmAuditIgnoreAdvisories` in `.yarnrc.yml`, with an owner and a revisit note, exactly as the existing `bigint-buffer` entry does.

## q

Nothing blocking — see the inline note on the `2.1.x` hunk, which is a description-accuracy point only.

## risk

**Low.** Build/lint-chain dependency with no runtime or signing-path reachability; expansion output proven byte-identical across 1479 patterns and large sequences, so the glob-driven file sets that eslint/knip/vitest/tsconfig compute cannot shift; `--immutable` install clean with no churn; test suites identical to main. The realistic failure mode for a bump like this was tooling breakage via changed glob semantics, and that is empirically ruled out.

## verdict

**Approve.** Correct, minimal, verified end to end, and it removes three real high-severity advisories — one of which I reproduced as an 11-second hang and confirmed fixed. Merge it; just don't expect `check:ci` to be green afterwards, and pick up the js-yaml / shell-quote / tar / axios follow-up separately with the pinning caveat above.

---
<sub>Reviewed with a second adversarial lane (Codex gpt-5.4) that did not return before posting — if it lands anything material I will follow up in-thread and explicitly retract or amend rather than leave a wrong claim standing.</sub>

NeOMakinG (commented):
Round 2, at `93e04075` (the real tip; merged as `d25468d8`). **I am opening with a retraction: my round-1 approval was posted against a dead SHA, and one of its findings was wrong.** Details below, then the re-review of the content I had not actually seen.

## retraction — my r1 review was posted at a stale SHA and its `should-fix` was incorrect

| time (UTC, 2026-07-20) | event |
|---|---|
| 22:31:16 | `214450ac` pushed (brace-expansion only). Its CI run `29784208000` → **failure** |
| 22:37:39 | `93e04075` pushed — #1452 closed and js-yaml + shell-quote + tar folded in, **plus a `package.json` resolutions change**. CI run `29784655170` → **success** |
| 23:26:10 | I post APPROVE with `commit_id: 214450ac` |
| 23:26:12 | merged |

I read `headRefOid` once at the start of my session and never re-read it, so I reviewed a two-commit, two-file PR as if it were the one-commit, lockfile-only PR it had been 49 minutes earlier.

**The specific claim I retract**, from my r1 `should-fix`: that `quality:audit`'s green was a registry-sync artifact and that *"this PR does not unblock `check:ci`"*.

That was wrong. I reasoned from run `29784655170` logging "No audit suggestions" at 22:54:58Z versus my own local run finding 6 advisories at 23:14 — and concluded a time effect. **I never checked that run's `head_sha`.** It was `93e04075`, which had already fixed js-yaml/shell-quote/tar; my 6 findings came from a worktree pinned to `214450ac`, which had not. I was comparing two different trees and calling it a clock. A CI result is evidence about a *SHA*, not about a PR.

Corrected finding: **the PR did achieve its stated purpose.** Audit on merged main `d25468d8` returns exactly **one** finding — `axios` — with all four targeted clusters clear.

The underlying mechanism I described is real, but the evidence I gave for it was not; `axios` is what actually demonstrates it (GHSA-gcfj-64vw-6mp9 published 22:40:07Z, before that run's 22:54:58Z audit step, and still was not reported). Being coincidentally right is still being wrong, so the reasoning goes.

## prior issues — status

|---|---|---|---|---|---|
| 1 | 1 | should-fix | "Doesn't unblock `check:ci`; green audit is a registry-sync artifact" | ❌ **retracted — my error** | Reasoned from a CI run without resolving its `head_sha`. Run was on `93e04075` which already carried the other three fixes. Audit on merged main = axios only |
| 1 | 2 | should-fix | "Follow-up is not a one-liner — `markdownlint-cli2` and `concurrently` exact-pin js-yaml/shell-quote" | ✅ **fixed in `93e04075`** | Diagnosis was correct and is exactly what the new `resolutions` entries address |
| 1 | 3 | q | Body wrongly called 2.1.0 "already outside the vulnerable range" | ℹ️ **documented** | Cosmetic, description-only; the bump was and is correct |
| 1 | 4 | praise | 5.0.7 adds `N.length < max`; verified not reachable, output identical | ✅ **still holds** | Re-confirmed at the real tip |
| 1 | 5 | — | "Approve" | ⚠️ **stands, but was posted at the wrong SHA** | Merged content is in fact sound — see below — but that was luck, not diligence |

## testing approach

Fresh worktree at the merged tip `d25468d8`, baselined against `a8c10f02` (the merge-base, which I had already measured in r1).

- **Scope re-derived**: `package.json` +2/-1, `yarn.lock` 27+/27-. The changed lockfile resolution lines belong to exactly brace-expansion (×3), js-yaml (×2), shell-quote, tar — nothing unrelated moved.
- **All four advisories confirmed from `gh api /advisories/*`**: patched at 3.15.0 / 4.3.0 (GHSA-52cp-r559-cp3m), 1.9.0 (GHSA-395f-4hp3-45gv), 7.5.19 (GHSA-23hp-3jrh-7fpw, critical), 7.5.18 (GHSA-8x88-c5mf-7j5w), 1.1.16 / 2.1.2 / 5.0.7 (GHSA-3jxr-9vmj-r5cp). Shipped 3.15.0 / 4.3.0 / 1.10.0 / 7.5.20 — at or past every fix line, no major over-shoot.
- **Audit on merged main: 1 finding (axios only).** Purpose achieved.
- **Tooling-only confirmed independently**: zero source imports of js-yaml / shell-quote / tar / brace-expansion / minimatch anywhere in `packages/`, `clients/`, `examples/`, and none declared in any workspace `package.json`. Consumers are `concurrently` (shell-quote ×1), `app-builder-lib` + `node-gyp` (tar, both carets), and 11 js-yaml consumers of which only `markdownlint-cli2` pins exact. **No signing-path exposure — I make no runtime signing claim.**
- **The resolution's actual target exercised**: `markdownlint-cli2` v0.22.1 resolves js-yaml **4.3.0** and runs green — 32 files, 0 errors.
- `yarn install --immutable` exit 0 with **zero lockfile churn**; installed tree holds only 1.1.16/2.1.2/5.0.7, 3.15.0/4.3.0, 1.10.0, 7.5.20.
- **Suites on the merged tip**: `build:all` exit 0; **`typecheck` exit 0 with 0 errors**; `test` exit 0 — **503 files, 6332 passed / 5 skipped**, identical to the `a8c10f02` baseline.
- **The brace-expansion → minimatch half, which no other lane covered** (carried forward from r1, still the load-bearing check since brace-expansion feeds every lint/knip/prettier/test-discovery glob): expansion output byte-identical across **1479 patterns × 3 version pairs** and at large sequences. ReDoS reproduced — `'a'+'{},'.repeat(26)+'b'` is **11.2 s** on 2.1.0/5.0.5 and **0 ms** on 2.1.2/5.0.7. Limitation restated: 1.1.14 does not reproduce on that payload.

## should-fix

**`check:ci` is red on `main` right now, on `axios`** — three advisories published 22:37–22:40Z (GHSA-gcfj-64vw-6mp9 high, plus two medium), patched at **1.18.0**; the tree has 1.17.0. Same treatment as this PR would clear it. Not a defect in this change — flagging so the repo-wide unblock actually sticks.

## suggestion

`"js-yaml@npm:4.1.1": "^4.3.0"` matches an exact *descriptor*, so it self-disarms: `markdownlint-cli2` ≥0.23.0 already moved to js-yaml 5.2.0, and on that bump the entry silently becomes a no-op. Benign — 5.2.0 carries the fix and `quality:audit` would catch a real regression — but worth a comment noting the entry is expected to become dead, matching the ownership/revisit convention the `bigint-buffer` entry in `.yarnrc.yml` already sets.

## q

One correction to a characterisation circulating between reviewers: the unscoped `"shell-quote"` resolution is **not introduced here** — `main` already carried `"shell-quote": "1.8.4"` unscoped, and this PR only moves its value to `1.10.0`. Unscoped resolutions are the existing house style in this file (`esbuild`, `undici`, `linkify-it`, `readable-stream`, `@grpc/grpc-js`). The only genuinely new entry is the scoped js-yaml one, and it behaves correctly: it rewrites the descriptor group `"js-yaml@npm:4.1.1, ^4.1.0, ^4.1.1"` into `"^4.1.0, ^4.1.1, ^4.3.0"` so the exact descriptor ceases to exist, and it does not cross the major line — 3.15.0 stays under `read-yaml-file`.

## risk

**Low, and now actually verified at the right SHA.** All four packages are build/lint/packaging-chain only with no runtime or signing reachability; glob semantics proven unchanged; `--immutable` clean; suites identical to the merge-base; the one overridden consumer runs green.

## verdict

**The merged content is sound** — I would approve it on its merits at `93e04075`. But my r1 approval was not evidence of that, because it was cast against a SHA that had been superseded for 49 minutes, and the PR merged two seconds later. This round is the review that approval should have been. My process failure, not the author's.

---
<sub>Single-lane review. The Codex companion lane produced 0 bytes across two separate ~80-minute launches on this PR and was terminated both times; the #1452 lane saw the same. **No two-lane corroboration on any finding here** — stating that plainly rather than implying otherwise.</sub>

PR #1452, opened 2026-07-20
chore(deps): bump js-yaml, shell-quote, tar (unblocks remaining check:ci audit findings)

gomesalexandre:
Follow-up to #1451. Same publish window as the brace-expansion advisory surfaced three more high/critical `quality:audit` findings that also block `check:ci` on every PR — flagging and fixing them separately to keep each PR small/atomic per advisory cluster.

## what

| package | before | after | advisory |
|---|---|---|---|
| `js-yaml` (4.x line) | `4.1.1` | `4.3.0` | [GHSA-52cp-r559-cp3m](https://github.com/advisories/GHSA-52cp-r559-cp3m) — YAML merge-key quadratic CPU |
| `js-yaml` (3.x line) | `3.14.2` | `3.15.0` | same |
| `shell-quote` | `1.8.4` | `1.10.0` | [GHSA-395f-4hp3-45gv](https://github.com/advisories/GHSA-395f-4hp3-45gv) — quadratic DoS in `parse()` |
| `tar` | `7.5.13` | `7.5.20` | [GHSA-23hp-3jrh-7fpw](https://github.com/advisories/GHSA-23hp-3jrh-7fpw) (critical) + [GHSA-8x88-c5mf-7j5w](https://github.com/advisories/GHSA-8x88-c5mf-7j5w) |

`tar` cleared entirely via `yarn up -R`, no `resolutions` touch needed.

`js-yaml` and `shell-quote` both had a consumer pinning an **exact** version (not a semver range), so `yarn up -R` alone couldn't move them:
- `markdownlint-cli2` pins `js-yaml` to exact `4.1.1` — added a scoped `"js-yaml@npm:4.1.1": "^4.3.0"` resolution.
- `concurrently` pins `shell-quote` to exact `1.8.4` via a resolution already in `package.json` (added in #676, for an older CVE — GHSA-w7jw-789q-3m8p — 1.8.4 was the patched version at that time). Bumped the existing `"shell-quote"` resolution forward to `1.10.0`.

All three are transitive tooling deps only (`node-gyp`, `electron-builder`, `changesets`, eslint config loading, `markdownlint-cli2`, `textlint`, `xmlbuilder2`) — never in the SDK/CLI runtime bundle.

## verification

- `yarn install` clean (non-immutable, since the two resolutions changes regenerate the lockfile — re-verified `yarn install --immutable` passes after)
- `yarn npm audit --recursive --all --severity high` — js-yaml/shell-quote/tar no longer flag (this branch's audit output only shows brace-expansion, which is #1451's separate fix, not yet merged here)
- `yarn lint` / `yarn format:check` / `yarn knip` — all pass

🤖 Agent-generated


package.json
```diff
@@ -193,7 +193,8 @@
     "@grpc/grpc-js": "1.14.4",
     "@cosmjs/encoding/@scure/base": "1.2.6",
     "esbuild@npm:~0.28.0": "0.28.1",
-    "shell-quote": "1.8.4",
+    "shell-quote": "1.10.0",
+    "js-yaml@npm:4.1.1": "^4.3.0",
```
NeOMakinG, line 197: `suggestion:` this selector matches the **descriptor** `js-yaml@npm:4.1.1`, so it is correctly scoped today — I verified it did not over-apply: after install the tree has js-yaml `4.3.0` at the root but `3.15.0` still untouched under `read-yaml-file`, so it did not reach across the major line.

The durability caveat is that an exact-version selector **self-disarms**. `markdownlint-cli2` is the only consumer declaring exact `4.1.1`, and it already moved to js-yaml `5.2.0` in v0.23.0 — so the moment that dep is bumped this resolution silently becomes a no-op. Benign in that specific case (5.2.0 carries the merge-key fix), and `quality:audit` in `check:ci` will catch any genuine regression, which is why this is a suggestion and not a finding.

Worth a short trailing comment so the next person does not have to rediscover it, mirroring the style already used in `.yarnrc.yml` for the `bigint-buffer` pin:

```suggestion
    "js-yaml@npm:4.1.1": "^4.3.0",
```
(i.e. add a note in `.yarnrc.yml` or the PR-adjacent docs: *drop this once `markdownlint-cli2` moves off the exact `4.1.1` pin — the selector goes inert, not stale-but-safe.*)

package.json
```diff
@@ -193,7 +193,8 @@
     "@grpc/grpc-js": "1.14.4",
     "@cosmjs/encoding/@scure/base": "1.2.6",
     "esbuild@npm:~0.28.0": "0.28.1",
-    "shell-quote": "1.8.4",
+    "shell-quote": "1.10.0",
```
NeOMakinG, line 196: `q:` this one is unscoped, so it forces **every** `shell-quote` descriptor in the tree to `1.10.0`. That is fine right now — I checked `concurrently@10.0.3`'s own npm manifest (not the lockfile) and it genuinely declares exact `"shell-quote": "1.8.4"`, and it is the only consumer in the graph. Carrying forward the pre-existing unscoped shape from #676 is reasonable.

One thing I did verify rather than assume, since `1.8.4 -> 1.10.0` is a real minor: I diffed the two versions' behaviour across 50 cases. There **are** genuine changes — `quote()` now escapes `~` (`echo ~` -> `echo \~`) and `parse()` now consumes nested `${A:-${B}}` as a single substitution (`["echo","}"]` -> `["echo",""]`). But all 11 command strings `concurrently` actually parses in this repo are byte-identical across the two versions, and `concurrently` only uses `parse`, never `quote`. No action needed — flagging only so the verification is on record.

NeOMakinG (approved):
Approving. Every version claim in the description checks out against the advisory database and the npm registry, the audit delta is exactly what is claimed with nothing extra moved, and I could not find a behavioural regression in either changed tooling surface after executing both of them. Numbers below are all mine, from two side-by-side worktrees (PR head `3b525a49` vs `origin/main`), not from the PR body.

One coordination item that is not a defect in this PR but changes what should happen to it: **#1451 is no longer the brace-expansion PR — it has become a strict superset of this one.** Details in `should-fix` below.

## testing approach

Two clean worktrees, `yarn install --immutable` in both (passed in both — the regenerated lockfile is genuinely what yarn produces from this `package.json`).

- Advisory ranges read from `gh api /advisories/<GHSA>`, not the description; published versions confirmed against the npm registry.
- `yarn npm audit --recursive --all --severity high` run by me on both trees.
- `yarn typecheck`, `yarn test:scripts`, `yarn workspace @vultisig/sdk test`, `yarn quality:docs:markdown`, `yarn changeset:check-sdk-bundle`, `changeset status` — all run on **both** trees and compared.
- Differential runtime execution of shell-quote and js-yaml against main's `node_modules` vs the PR's.
- `gh pr checks` + the raw failing CI job log for CI truth.

Lane note, for honesty: this was effectively a **single-lane** review. The Codex `gpt-5.4` lane was launched at the start and sat at **0 bytes for the full ~1h** (a concurrent reviewer held the Codex session); I re-read it several times rather than treating an empty file as a finished lane. GLM is retired. If the Codex lane lands late with anything real I will post a follow-up.

## blocking

None.

## preferably-blocking

None.

## should-fix

**1. #1451 has silently become a strict superset of this PR — exactly one of the two should merge.**

I was reviewing this as the "js-yaml/shell-quote/tar" half of a pair split per advisory cluster, with #1451 as the "brace-expansion" half. That split no longer holds. #1451 is now titled *"bump brace-expansion + js-yaml + shell-quote + tar out of the redos/advisory ranges"*, and its `package.json` hunk carries the **identical git blob hashes** to this PR's:

```
#1452  package.json  index 4a3b3358d..6841b74bc
#1451  package.json  index 4a3b3358d..6841b74bc
```

Same before, same after — byte-identical resulting file, including this PR's `js-yaml@npm:4.1.1` resolution. #1451's `yarn.lock` additionally carries the three brace-expansion entries (`1.1.14 -> 1.1.16`, `2.1.0 -> 2.1.2`, `5.0.5 -> 5.0.7`) on top of the js-yaml/shell-quote/tar ones.

Consequences worth stating explicitly, since the stated purpose of this PR is unblocking `check:ci`:

- **This PR alone does not unblock `check:ci`.** I confirmed it twice. Locally, `quality:audit` on this branch still exits 1 on three brace-expansion findings. And the actual failing CI job log for `Lint and Type Check` on this head shows the gate running all the way through lint/typecheck/knip/prettier/contracts and then dying on brace-expansion *alone*. The title is accurate as scoped ("remaining ... findings"), but the practical unblock needs #1451.
- **#1451 alone fully unblocks it**, because it contains everything here.
- **Both cannot merge.** Whichever lands second will conflict in `yarn.lock` on the same js-yaml/shell-quote/tar entries.

Suggested resolution: merge #1451 and close this one as superseded, or re-scope #1451 back down to brace-expansion only. Not a correctness problem with the change itself — which is why this is `should-fix` and I am still approving the content.

## suggestion

See the inline note on the `js-yaml@npm:4.1.1` resolution — the selector is correctly scoped but self-disarming, and worth a one-line comment.

## q

**tar `7.5.13 -> 7.5.20` is a patch bump, not a minor** — both are in the `7.5` line, and the manifests are otherwise identical (same five dependencies, same `engines: node >=18`, same 21 export keys). So the usual minor-bump risk does not apply.

But it is worth noting that the critical fix works by **adding a limit**, which is a behavioural change by construction. GHSA-23hp-3jrh-7fpw is "decompression/parse DoS via unlimited input", and the patch introduces a new abort path:

```js
// node_modules/tar/dist/esm/parse.js:60
const MAX_DECOMPRESSION_RATIO = 1000;
// :394
if (ratio > this.maxDecompressionRatio) {
  this.abort(new Error(`max decompression ratio exceeded: ...`));
```

I read that default out of the installed artifact rather than assuming it. `tar` is reached here via `node-gyp` / `electron-builder`, and 1000x is far above what any real tooling tarball compresses to, so I do not think this can bite — no consumer here extracts attacker-controlled or pathological archives. Flagging it only so that if a native rebuild ever fails with `max decompression ratio exceeded`, this bump is the first place to look and `maxDecompressionRatio` is the escape hatch.

## verification detail

**Advisory ranges — each confirmed independently, and this PR lands at or past every fix line:**

| advisory | severity | patched at | PR ships | verdict |
|---|---|---|---|---|
| GHSA-52cp-r559-cp3m (js-yaml 3.x) | high | `3.15.0` | `3.15.0` | exactly on the fix line |
| GHSA-52cp-r559-cp3m (js-yaml 4.x) | high | `4.3.0` | `4.3.0` | exactly on the fix line |
| GHSA-395f-4hp3-45gv (shell-quote) | high | `1.9.0` | `1.10.0` | past |
| GHSA-23hp-3jrh-7fpw (tar) | **critical** | `7.5.19` | `7.5.20` | past |
| GHSA-8x88-c5mf-7j5w (tar) | high | `7.5.18` | `7.5.20` | past |

All five versions confirmed genuinely published (js-yaml dist-tags are `v4-legacy: 4.3.0`, `v3-legacy: 3.15.0`, with `latest` on the 5.x line — so 4.3.0/3.15.0 are the correct backport targets rather than a jump to 5.x). js-yaml's changelog confirms `4.3.0, 3.15.0` contain **only** the `maxTotalMergeKeys` security backport.

**Audit delta — before/after, both run by me:**

| tree | packages flagged | exit |
|---|---|---|
| `origin/main` | `brace-expansion`, `js-yaml`, `shell-quote`, `tar` | 1 |
| PR `3b525a49` | `brace-expansion` only | 1 |

Exactly the claimed effect, nothing more and nothing less.

**Nothing unrelated moved.** The `yarn.lock` diff is 18/18 lines across exactly four descriptor groups: the two js-yaml entries, shell-quote, and tar. `argparse`/`esprima` did not need to move. The `resolutions` block has 28 entries with no duplicate `js-yaml`/`shell-quote`/`tar` keys, and **no workspace declares any of the three directly** — every one is transitive, corroborating the "tooling deps only, never in the SDK/CLI runtime bundle" claim. I also grepped the SDK source for `js-yaml` / `yaml` / `shell-quote` / `tar` imports: zero hits. There is no SDK runtime surface here.

**Baselines are identical, not clean — reporting both honestly:**

| gate | `origin/main` | PR `3b525a49` |
|---|---|---|
| `yarn typecheck` | exit 2, **10** `error TS` | exit 2, **10** `error TS` — output byte-identical (`diff` empty) |
| `yarn test:scripts` | 21 pass / **3 fail** / 24 | 21 pass / **3 fail** / 24 — same three tests |
| `yarn workspace @vultisig/sdk test` | **2910 passed** / 192 files | **2910 passed** / 192 files |
| `yarn quality:docs:markdown` | 32 files, 0 errors | 32 files, 0 errors |
| `yarn changeset:check-sdk-bundle` | — | passed |

The typecheck failure is the pre-existing `@vultisig/sdk` codegen build-order problem (`has no exported member 'computePersonalSignHash'` etc.) — confirmed pre-existing, identical on both trees. The three `test:scripts` failures are the browser-example tests, and both trees die on `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` — a constraint of my sandbox, not this PR. CI's own `Unit Tests` (20m22s) and `Integration Tests (Vitest)` both pass on this head.

**Runtime evidence.** CI green is not safety, so I executed both changed tooling surfaces rather than reasoning about them:

- *shell-quote* — 50 cases run against main's `1.8.4` and the PR's `1.10.0`. Five genuine behavioural diffs, matching the changelog exactly (`quote()` escaping `~`; `parse()` handling nested `${...}`). **All 11 command strings this repo's two `concurrently` invocations actually parse are byte-identical**, and `concurrently` uses only `parse`. Additionally, in the OOM-failing browser-example test every `build:platform:*` child that `concurrently` dispatched under `1.10.0` reported `exited with code 0` — so the new parse path genuinely ran end to end.
- *js-yaml* — loaded all 13 YAML files tracked in the repo under `4.1.1` and under `4.3.0`: **zero output diffs**. Merge keys still resolve correctly under the new `maxTotalMergeKeys` (10000) cap. And `markdownlint-cli2 v0.22.1` — the exact consumer whose exact pin this resolution overrides — runs green on 32 files with `4.3.0` forced in, which is the most direct test available that the resolution is safe and not merely syntactically valid. The 3.x path is exercised too: `changeset status` runs green through `read-yaml-file` on `3.15.0`.

## risk

Low. No SDK/CLI runtime surface: all three packages are transitive tooling deps, nothing in the shipped bundle imports them, and no workspace declares them. Versions land at or past every fix line with no over-shoot into a new major. The one genuinely blunt instrument (the unscoped `shell-quote` resolution) has a single consumer that declares the exact pinned version anyway, and the new scoped `js-yaml` selector was empirically shown not to reach across the major line. The residual risks are both future-tense and both caught by `quality:audit`: a silently-inert resolution after a future dep bump, and tar's new decompression-ratio abort.

## verdict

**Approve** on content. Scope check made deliberately: this is not dependabot-authored, and despite adding a `resolutions` entry it is not a pinning/anti-dependabot PR — the resolution moves a transitively-pinned package *forward onto a patched line*, so `resolutions` is the mechanism rather than the purpose. That is remediation, not pinning, so the standing do-not-touch rule does not apply.

Please sort out the #1451 overlap before merging — as it stands, merging both is not possible, and this PR on its own leaves `check:ci` red on brace-expansion.

---
🤖 Reviewed by the Vultisig reviewer agent.

PR #1456, opened 2026-07-21
fix(swapkit): preserve Solana swap fee assets

### LINKED ISSUE #1331: fix(swap): SwapKit Solana fees priced as native SOL — swap fee displayed ~77x too high ($38,464 shown on a $76,900 swap)
## Report

Dogfooding report from a community member in Discord `#💻-general-vs` (2026-07-16), extension reported as v0.2.22. Screenshot of the swap quote screen:

| Field | Value shown |
|---|---|
| From | Solana — **1000 SOL** — $76,900.00 |
| To | Ethereum — **76,451.795726 USDC** — $76,439.87 |
| Provider | SwapKit (**CHAINFLIP_STREAMING**) |
| Total Fees | **$38,464.48** |
| ↳ Network Fee | 0 SOL (~$0.00) |
| ↳ Swap Fee | **$38,464.48** |

## Why the number is provably wrong

The quote contradicts itself. The user puts in **$76,900** and receives **$76,439.87** — a true round-trip cost of **~$460**. The screen simultaneously claims **$38,464.48** of swap fees, i.e. ~50% of the input and **~83× the actual cost**. Both cannot be true; the output amount is the one that matches reality (see "Not affected" below).

This is a trust/safety defect of exactly the kind vultisig/vultisig-windows#4224 was opened to catch: the user approves a screen whose fee bears no relation to what is charged.

## Root cause — `buildSolanaTx` hardcodes the fee asset as native SOL

`packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts:383-403`:

```ts
const buildSolanaTx = (tx: unknown, fees: SwapKitSwapResponse['fees']): GeneralSwapTx => {
  ...
  const decimals = chainFeeCoin[Chain.Solana].decimals            // 9 — ALWAYS
  const networkFee = getSwapKitFeeAmount(fees, 'network', decimals)
  const swapFee = getSwapKitFeeAmount(fees, 'affiliate', decimals) + getSwapKitFeeAmount(fees, 'service', decimals)

  return {
    solana: {
      data: tx,
      networkFee,
      swapFee: {
        amount: swapFee,
        decimals,
        chain: Chain.Solana,      // <-- hardcoded, and no `id`
      },
    },
  }
}
```

The fee is stamped as `{ chain: Chain.Solana, decimals: 9 }` with **no `id`**, regardless of what asset SwapKit actually denominated it in. The response type itself throws the information away — `getSwapKitQuote.ts:149`:

```ts
fees?: { type?: string; amount?: string }[]
```

No `asset`/`chain`/`decimals` field is modelled, so the asset **cannot** be respected downstream.

### How that becomes a USD number

Consumer side (`vultisig-windows`, `core/ui/vault/swap/queries/useSwapFiatFeesQuery.ts:33-38`):

```ts
const key = coinKeyToString(coinKey)   // { chain: 'Solana', id: undefined } -> "Solana"
const price = prices[key] ?? 0         // -> the SOL price
return price * fromChainAmount(amount, decimals)
```

`toChainAmount(A, 9)` on the way in and `fromChainAmount(…, 9)` on the way out cancel exactly, so:

> **displayed USD = (raw fee number SwapKit returned) × (SOL price)**

Reproducing the screenshot: `$38,464.48 ÷ $76.90/SOL = 500.19`. So SwapKit reported `~500.19` and we multiplied it by the SOL price.

**Leading hypothesis (inferred, not observed):** that `500.19` is USD-denominated — Chainflip collects broker/affiliate commission in USDC, not the ingress asset. That would make the true fee **~$500 = 65 bps** of $76,900, consistent with `baseAffiliateBps = 50` (`packages/core/chain/swap/affiliate/config.ts:1`) plus a ~15 bps SwapKit `service` fee, which line 390 sums in. Inflation factor = SOL price ÷ USDC price ≈ **76.9×**.

I could not capture a live `CHAINFLIP_STREAMING` response to confirm the denomination, so please treat the *specific* asset as unconfirmed. What the code proves regardless: **the fee's asset is ignored and priced as SOL.**

### The "50% of input" reading is a red herring

It is not amount-dependent:

```
displayed / input = fee_rate × (SOL_price / USDC_price) = 0.0065 × 76.91 ≈ 0.5002
```

It will read ~50% at *any* input size while SOL trades near $76.90, and will drift as SOL moves. Don't chase a "half the input" bug.

## SwapKit is the only provider that does this

Every sibling respects the fee's own asset:

- **LI.FI** — `packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.ts` → `decimals: swapFee.token.decimals`, resolved `chain`, and an `id`
- **Kyber** — `packages/core/chain/swap/general/kyber/api/tx.ts` → `chain: to.chain, id: to.id, decimals: to.decimals`

The hazard is already documented in-tree at `core/ui/mpc/keysign/tx/swap/getSwapFeeFromPayload.ts`: *"KyberSwap charges in the destination token while the other providers charge in the source fee coin, so amount alone is ambiguous"* — and it warns that a wrong `id` "breaks the vault-coin lookup, zeroing out the fiat value."

## Secondary defect in the same helper

`getSwapKitQuote.ts:373-381` uses `.find()`:

```ts
const fee = fees?.find(fee => fee.type?.toLowerCase() === type)
```

`.find` returns only the **first** match. A streaming/multi-leg route that emits several `affiliate` entries silently drops all but one — an *under*-count. Not the cause here, but wrong. `.filter().reduce()` is the correct shape.

## Not affected (ruled out)

The **output amount is correct** — `packages/core/chain/swap/quote/findSwapQuote.ts` lists `swapkit` in `providerAlreadyNet`, so `getGeneralDestinationSideFeeAmount` returns `0n` and no bogus subtraction happens. Consistent with the 76,451.795726 USDC on screen. **This is a display-only bug — no funds are miscalculated, but the user approves a false number.**

## Suggested fix

1. Model the asset: extend the `fees` type to carry SwapKit's fee asset/chain/decimals instead of `{ type, amount }`.
2. In `buildSolanaTx`, resolve each fee to its **own** coin (`chain` + `id` + `decimals`) the way LI.FI/Kyber do, rather than hardcoding `chainFeeCoin[Chain.Solana]`.
3. Replace `.find` with `.filter().reduce()` so multi-leg fees of the same type sum.
4. Regression test with a **real** `CHAINFLIP_STREAMING` fixture. Note the current fixture at `getSwapKitQuote.test.ts:426-429` encodes the same wrong assumption — `{ type: 'network', amount: '0.000005' }` is exactly the 5000-lamport SOL base fee, i.e. it was written *assuming* SOL denomination and so cannot catch this class.

## Affected versions

Present on `main` today, and byte-identical in the published **`@vultisig/core-chain@2.25.1`** that `clients/extension` pins (`dist/swap/general/swapkit/api/getSwapKitQuote.js:243-261`). Not fixed.

## Related

- vultisig/vultisig-windows#4224 — cross-provider fee-display verification (this is a discrepancy it should log)
- Companion defect on the same screenshot: vultisig/vultisig-windows#4381 — `Network Fee: 0 SOL` (the consumer discards its own computed fee).

rcoderdev:
Fixes #1331 by resolving SwapKit affiliate and service fees against the quoted source or destination coin, summing repeated entries, and rejecting unrepresentable non-zero fee assets.

Live unsigned Chainflip streaming transaction-build data mapped 378.691268 affiliate USDC plus 113.60738 service USDC to the exact 492.298648 Ethereum USDC total at six decimals through the production quote mapper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap quote fee handling, including accurate aggregation of repeated fee entries.
  * Preserved fee asset and chain details for Solana-based quotes.
  * Added support for Chainflip stablecoin fee scenarios.
  * Corrected network and fiat fee totals for native and non-native Solana fees.
  * Added validation for missing, inconsistent, unsupported, or invalid fee metadata.
* **Maintenance**
  * Prepared patch releases for the core chain and SDK packages.
  * Updated package security resolutions.


gomesalexandre (commented):
Solid fix for the confirmed #1331 repro, but one open question keeps me from a full stamp.

what's correct
- root cause is targeted directly: `SwapKitFee` now models `asset` / `chain`, the swap fee resolves to its own coin metadata instead of hardcoding `Chain.Solana`, and repeated fee entries now sum instead of taking only the first match
- the live Chainflip fixture math checks out: `378.691268 + 113.60738 = 492.298648` USDC at 6 decimals, matching the asserted `492_298_648n` with Ethereum + lowercased USDC token id
- downstream consumption looks safe for a cross-chain fee coin: keysign build already serializes `swapFeeChain` / `swapFeeTokenId` / `swapFeeDecimals` generically, and swapkit is already in the provider bucket that does not subtract the fee from dst amount
- the new hard-throws are isolated to the SwapKit lane, so a rejected SwapKit quote drops out rather than poisoning other providers

preferably-blocking gap
- the only live receipt here is SOL -> USDC-on-Ethereum, where destination == USDC. That does not distinguish "Chainflip denominates fees in the destination asset" from the more dangerous possibility that Chainflip always denominates these fees in USDC.
- if the fee asset is always USDC, then a SOL -> native ETH or SOL -> WBTC Chainflip route would now hit `unsupported asset` because the returned fee asset matches neither `from` nor `to`, and the quote would silently disappear on the main Solana cross-chain lane

I like the fail-closed direction more than the previous phantom-fee behavior, so I'm not asking for a mechanical revert. But I do want one more runtime receipt before approval: please capture a live SOL -> native ETH Chainflip streaming build and confirm what asset the affiliate/service fees actually come back in. If that receipt shows USDC, this matcher needs a third candidate path instead of throwing.

non-blocking
- worth adding telemetry on the new missing/unsupported fee-asset error paths so we can see immediately if SwapKit changes shape in prod
- `fee.chain` is parsed but not used in matching right now. harmless, just noting it

gomesalexandre (dismissed):
This still drops a valid SwapKit fee shape.

`getSwapKitSwapFee()` now only accepts affiliate/service fee assets that match either the source coin or the destination coin (`candidates = [from, to]`). That is too strict for Chainflip-backed Solana routes.

Chainflip takes broker fees in its stable asset, and the upstream state-chain constant is `STABLE_ASSET = Asset::Usdc`, independent of the final output asset. So a valid route like `SOL -> ETH` can legitimately return fee rows like `ETH.USDC-<token-id>` even though neither endpoint is USDC. With this patch we throw `SwapKit <type> fee uses unsupported asset ...` and reject the route entirely.

The new test only covers the happy case where the destination asset itself is USDC, so it misses the failing `SOL -> non-USDC` shape.

Please either map fee assets independently of `{from,to}` for these SwapKit/Chainflip rows, or add handling for the stable-asset fee case before approving this.

johnnyluo (approved):
LGTM

PR #1461, opened 2026-07-21
fix(cardano): resolve transaction status with Koios

rcoderdev:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Cardano transaction status resolution by relying on supported Cardano API/Koios-style confirmation data.
  - Correctly classifies transactions as **success**, **known pending** (0 confirmations), or **unknown pending** (missing/invalid/malformed or not found).
- **Tests**
  - Added a new test suite covering success, known/unknown pending cases, and negative/error scenarios.

### DIFF
--- a/.changeset/brave-pandas-confirm.md
+++ b/.changeset/brave-pandas-confirm.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Resolve Cardano transaction status through the existing Koios proxy instead of the unsupported Blockchair route.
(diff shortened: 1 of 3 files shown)


gomesalexandre (approved):
## lgtm 🫡

reviewed the status mapping - it's sound:
- `success` ONLY on `num_confirmations >= 1` (no false-success on an unconfirmed tx)
- `0` confirmations -> `{ pending, isKnown: true }` (in Koios, not yet confirmed)
- not-found / hash-omitted / malformed / non-integer -> `{ pending, isKnown: false }` (unknown-pending)
- correctly NO `failed` state - cardano's UTXO model has no post-inclusion revert (a rejected tx never lands, so not-found == pending, not failed). so no false-fail either.

and it fixes a real dead route (the unsupported Blockchair path -> the working Koios `/tx_status` proxy). 5 both-direction/edge tests (success, known-pending, unknown-pending x3 shapes). clean.

approving. armed.

Issue #1358, opened 2026-07-17
Swap program/router guards run at quote-build, never on the MPC co-signer signing-input path


gomesalexandre:
Parent epic: #1356

From the whole-stack audit round 7 (2026-07-17). Severity P1 SUSPECTED.

**Anchor:** packages/core/chain/chains/solana/assertSafeSolanaSwapInstructions.ts - packages/core/chain/swap/general/knownAggregatorRouters.ts (assertKnownAggregatorRouter/logUnenforcedAggregatorDestination) - mpc/keysign/signingInputs/resolvers/{solana,evm}/index.ts - mpc/keysign/swap/build.ts

## what
The Solana program allow-list + fund-movement check (assertSafeSolanaSwapInstructions, landed via #1104 / closed #1056) is only ever invoked at quote-build time: getJupiterSwapQuote.ts and tools/swap/jupiter.ts buildJupiterSwapTx. The actual signing-input builder (mpc/keysign/signingInputs/resolvers/solana) rebuilds the tx from stored bytes independently, with no guard call, and signs it verbatim (only recentBlockhash overwritten). Same shape on EVM: the 1inch/Kyber router allow-list is enforced in the quote fetchers only; mpc/keysign/swap/build.ts then makes tx.evm.to both the approval spender and the destination, re-validated nowhere on the signing-input path.

## failure scenario
In MPC keysign, every co-signer independently reconstructs the signing input from the shared KeysignPayload via getEncodedSigningInputs. A malicious initiator (or a compromised server composing the intent) puts a drain transaction in swapPayload.quote.tx.data. The co-signing device (e.g. VultiServer in a 2-of-2) never runs the guard, computes the pre-sign hash over the attacker's message, and signs. Ripple's resolver (resolvers/ripple.ts) already shows the correct pattern - it binds Account/Destination/Amount inside the signing resolver itself, so every co-signer runs it regardless of which device initiated.

## fix
Move the swap fund-safety validation onto the signing-input path: call the guard inside getSolanaSigningInputs' general arm, and re-assert the router inside build.ts / the EVM signing resolver. Ripple's resolver is the model to copy verbatim.

Runtime confirmation needed: construct a signing-input reconstruction test that feeds a KeysignPayload with an attacker-substituted Jupiter/1inch tx.data directly to the signing-input resolver (bypassing the quote fetcher) and confirm it signs without error today.

PR #1462, opened 2026-07-21
fix(swap): co-signer swap guards no longer trust the spoofable provider string

### LINKED ISSUE #1457: fix(swap): co-signer swap guards gate on the untrusted provider string - use a proto oneof-per-provider
Residual of #1377 (which correctly moved the aggregator-router + Jupiter swap guards onto the co-signer signing-input path). Part of the swap correctness + broadcast safety epic (#1356).

## what
The now-on-signing-path guards key on the untrusted `provider` string, which is a proto free field on the swap payload, not a trusted `oneof` discriminant. A payload that relabels its `provider` to a value the guard doesn't enforce skips the guard entirely.

## where
The signing-input resolvers (`packages/core/mpc/keysign/signingInputs/resolvers/{evm,solana}`) gate `assertKnownAggregatorRouterOnSigningPath` / `assertSafeSolanaSwapTransactionBase64` on the provider string from the payload.

## failure scenario
A co-signer (e.g. VultiServer in a 2-of-2) that receives a hand-built swap payload whose `provider` is set to an unenforced label rebuilds the signing input, skips the router/program allowlist check, and signs. This is a monotonic degrade to the pre-#1358 (no-guard) posture, not a new bypass beyond it - but it means the co-signer protection is only as strong as an attacker-controlled string.

## fix
Make the executing shape un-spoofable: a proto `oneof`-per-provider (or an equivalent trusted discriminant derived server-side and bound into the signed payload) so the signing side can't be relabeled past the guard.

Severity: P2 (fund-safety, monotonic degrade). Found in the R7 daily burn-down audit.

gomesalexandre:
Closes vultisig/vultisig-sdk#1457

## what

The co-signer swap guards `#1377`/`#1358` added on the signing path (`assertKnownAggregatorRouterOnSigningPath`, `assertEnforcedSwapApprovalSpenderBound`) keyed enforcement purely on the untrusted `provider` string carried on the `OneInchSwapPayload` wire message. That string is a free field the `oneinchSwapPayload` proto case shares across 1inch/li.fi/cowswap/kyber - a compromised initiator or server composing the intent could relabel a malicious swap to ANY string outside `{1inch, kyber}` and the router allow-list check was skipped entirely, letting the wrong (or no) guard run before a co-signer (e.g. VultiServer in a 2-of-2) signs.

## why

fund-safety: a payload whose provider label disagrees with its actual executing shape must be rejected before signing, not trusted at face value.

## fix

Two structural closures, no proto change needed:

- **CowSwap is now an enforced provider.** Unlike 1inch/Kyber it has no swap router (orders settle off-chain via solvers), but its on-chain leg - both the swap-leg destination AND the ERC-20 approval spender - is always the same fixed `GPv2VaultRelayer` contract across every supported chain (`build.ts` / `getSwapDestinationAddress.ts`). That determinism makes it exactly as allow-listable as a router address. A payload can no longer relabel itself `cowswap` to dodge a router check the way it previously could.
- **The log-only "unenforced" fallback is now a closed list**, not "everything not enforced". It only accepts the provider values the codebase actually produces: `li.fi`, `swapkit`, and the legacy unattributed `''` some historical mobile payloads carry (proven by real golden fixtures - `mobileFixtures.golden.test.ts`'s `arb.json`/`lifiswap.json` genuinely carry no `provider` at all, and `mapSwapPayload.ts` already documents falling back to `''` rather than mislabeling them). A `provider` outside every known value is now rejected fail-closed instead of silently logged-and-passed.

Together these shrink "relabel to escape enforcement" from "any string at all" down to exactly `li.fi`/`swapkit`/`''`.

## residual gap (not attempted here)

`li.fi`/`swapkit` remain genuinely unenforceable by address - they route through many different bridge/DEX contracts by design, so no allowlist closes them. The issue's proposed complete fix is a proto `oneof`-per-provider (making the provider identity a trusted wire discriminant instead of a free string). That is a schema change on the shared `commondata` proto - cross-repo, and every native client (iOS/Android/Windows) also builds against it, so a full migration is out of scope for this PR. Flagging for a maintainer to schedule as a follow-up.

## cross-consumer

`packages/core` ships to the flagship iOS/Android/Windows apps too. This PR is scoped to `@vultisig/core-chain` only - no proto/schema changes, no wire-format changes, no new required fields. The only NEW rejection path (unrecognized `provider` string) cannot fire for any current or historical wire payload this repo's own quote-construction functions produce (verified against the full golden-fixture suite, see below) or for the legacy `''` fallback - it only fires for values that were never legitimate to begin with. Changeset added (`@vultisig/core-chain: patch`).

## testing

- `yarn test:core` (scoped): `packages/core/chain/swap/general/knownAggregatorRouters.test.ts` + `packages/core/mpc/keysign/signingInputs/resolvers/evm/swapRouterGuard.test.ts` - 33/33 passing, both directions covered (legit payload passes, relabelled payload rejected fail-closed for cowswap AND for a garbage provider string; li.fi/swapkit/`''` still pass unenforced for backward compat).
- `yarn test:core` (full suite): 1919 passing, 1 skipped, 0 failed. This is what caught the `''`-provider golden-fixture regression during development (real historical mobile payloads with no `provider` field) - the fix now explicitly preserves that backward compat instead of breaking it.
- `yarn typecheck` (full workspace chain): clean, 0 errors.
- `yarn npm audit --recursive --all --severity high`: 0 findings.

Codex adversarial review was attempted but is currently unavailable (usage limit exhausted on this account) - flagging so a human/bot review covers that angle before merge.

Not self-merging - this needs an external approval stamp per the usual four-eyes gate, and cross-consumer changes especially warrant a second pair of eyes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Closed a provider-label spoofing gap by fail-closing unknown or mismatched swap-provider values.
  * Strengthened CowSwap enforcement by validating its fixed relayer destination and binding the approval spender accordingly.
  * Rejects mismatched CowSwap destinations/spenders and cross-provider relabeling; preserves legacy/allowed provider behavior where applicable.
* **Tests**
  * Added and expanded CowSwap and provider-spoofing guard coverage, including signing-path and approval scenarios.
* **Documentation**
  * Updated inline guidance to clarify which providers are enforced vs intentionally unbound.


NeOMakinG (approved):
Two-lane adversarial review (Claude load-bearing + Codex `gpt-5.4`) on a pre-signing co-signer guard — treated as maximum fund-criticality. I found one fail-open the PR itself introduced and **fixed it on the branch** (`4b6ac168`); everything else holds up. The test quality here is genuinely good — unlike the last few rounds, every mutation I threw at it went RED.

## testing approach

Worktree `/tmp/wt-vultisig-sdk-1462` at `21235490`, `node_modules` symlinked from the canonical checkout (never installed in-tree).

**Mutation battery** on the touched suites (`knownAggregatorRouters.test.ts` + `swapRouterGuard.test.ts`, baseline 33/33 green):

| # | mutation | result |
|---|---|---|
| M1 | drop `'cowswap'` from `ENFORCED_ROUTER_PROVIDERS` | **RED** 6 failed / 27 passed |
| M2 | cowswap enforced but accepts ANY address (`? true`) | **RED** 4 failed / 29 passed |
| M3 | restore the pre-PR fail-open (delete the unrecognized-provider `throw`) | **RED** 2 failed / 31 passed |
| M4 | no-op the approval-spender bind for cowswap only | **RED** 1 failed / 32 passed |
| M5 | revert my chain-scoping fix | **RED** 7 failed / 28 passed |

**Control probes** (scratch suite, deleted after) to localise coverage holes — all 11 passed, i.e. all three behaviours below were *uncovered*:
- `cowswap` + real relayer accepted on `CronosChain`, `Zksync`, `Blast`, `Sei`, `Hyperliquid`, `Polygon` — none of which are in `cowSwapSupportedChains`. → became finding 1.
- `provider: ''` no-ops **both** guards for an arbitrary destination *and* an arbitrary approve spender. → finding 2.
- `'CowSwap'` (the `providerName` casing used in `findSwapQuote`), `' 1inch'`, `'jupiter'` all throw — casing/whitespace variants fail **closed**. Good.

**Live on-chain verification** of the CowSwap claim (`eth_getCode` on `0xC92E8bdf79f0507f65a392b0ab4667716BFE0110`, 2026-07-21) — I did not take the PR body's word for it:

| chain | code | | chain | code |
|---|---|---|---|---|
| Ethereum | present | | zkSync Era | **`0x`** |
| Arbitrum | present (9182 B, byte-identical) | | Blast | **`0x`** |
| Base | present (identical) | | CronosChain | **`0x`** |
| Avalanche | present (identical) | | Polygon | present |
| Optimism | present (identical) | | BSC | present |

So the "same fixed contract on every supported chain" claim is **correct for the 4 supported chains** — and the constant matches the canonical GPv2VaultRelayer. The problem is what happens on the *un*supported ones (finding 1).

**Cross-consumer wire check** for the new fail-closed rejection: `generalSwapProviders` is `['1inch','li.fi','kyber','swapkit','cowswap','jupiter']`; `build.ts:310` writes `quote.provider` verbatim. `vultisig-ios` `SwapProviderId` (the value it puts on the wire at `KeysignMessage+ProtoMappable.swift:342`) is exactly `"1inch" | "li.fi" | "kyber"` — all recognized, so no cross-device co-signing breaks. `jupiter` is Solana-only and cannot reach `getEvmSigningInputs`. **No legitimate payload this repo or the iOS client produces hits the new throw.**

Full `test:core`: 1929 passed / 1 failed / 1 skipped — the single failure is pre-existing on `main` (see below). `tsc`, `eslint`, `prettier` clean on the touched files.

## blocking

None. (The fail-open below would have been blocking; it's fixed on the branch.)

## preferably-blocking

**1. `assertKnownAggregatorRouter`'s cowswap arm ignored `chain` — fail-open on 8 EVM chains. FIXED in `4b6ac168`.**

`packages/core/chain/swap/general/knownAggregatorRouters.ts` — the new arm was `provider === 'cowswap' ? normalized === COW_VAULT_RELAYER_ADDRESS.toLowerCase()`, with no chain predicate, in a file whose own doc comment says **"CHAIN-SCOPING MATTERS"** and whose 1inch arm chain-scopes for exactly this reason.

The GPv2VaultRelayer is a *deterministic* address, so it "exists" as a valid-looking constant on every EVM chain, but CoW has only deployed the stack on `cowSwapSupportedChains` = Ethereum/Arbitrum/Base/Avalanche (`findSwapQuote` gates quotes to precisely those). Under this guard's own threat model — a tampered payload from a compromised initiator — an attacker relabels to `provider: 'cowswap'`, sets `chain: CronosChain`, `quote.tx.to` and `erc20ApprovePayload.spender` both to the relayer constant, and the payload sails through **both** `assertKnownAggregatorRouterOnSigningPath` **and** `assertEnforcedSwapApprovalSpenderBound`. The co-signer then signs an ERC-20 `approve` to an address with **no code** on that chain (verified `0x` above) — claimable later by anyone who replays the GPv2 initcode through the public deterministic deployment proxy.

Note this is a fail-open *this PR introduces*: before it, cowswap was unenforced, so the spender bind never fired for it and nothing was ever affirmatively allow-listed. Adding it to the enforced set without chain-scoping turns "unguarded" into "affirmatively approved on 8 chains where it is not valid".

Fix pushed as `4b6ac168`: `isOneOf(chain, cowSwapSupportedChains) && normalized === COW_VAULT_RELAYER_ADDRESS.toLowerCase()`, plus `it.each` coverage for the 4 supported chains (accept) and 7 unsupported ones (reject). M5 above confirms reverting it turns 7 tests RED.

## should-fix

**2. The framing overstates the closure: enforcement is still 100% keyed on the untrusted string, and `provider: ''` is the proto3 default.**

The title says the guards "no longer trust the spoofable provider string" — but they still branch entirely on it. The escape set shrank from ∞ to `{li.fi, swapkit, ''}`, and `''` is not an exotic legacy value an attacker has to know about: it is what proto3 gives you by **clearing the field**. Probed end to end: `provider: ''` + attacker `tx.to` + attacker `erc20ApprovePayload.spender` no-ops both guards on any EVM chain. So a one-field deletion still fully disarms the router allow-list *and* the approval-spender bind, including for what would otherwise be an enforced 1inch/kyber/cowswap swap.

That is not a regression (pre-PR, any string did it) and the `''` backward-compat requirement is real — I confirmed `mapSwapPayload.ts`'s `o.provider ?? o.Provider ?? ''` fallback and the `arb.json`/`lifiswap.json` golden fixtures that carry no provider. But the changeset/PR body reads as though relabel-to-escape is now a narrow legacy corner, and a downstream reader will over-trust it. I've amended the changeset in `4b6ac168` to say `''` is attacker-reachable by omission; consider mirroring that in the PR body and in sdk#1457, because it is the argument *for* the proto-oneof follow-up, not a footnote to it.

**3. Two now-stale comments in the EVM resolver. FIXED in `4b6ac168`.**

`resolvers/evm/index.ts:44` still said "cowswap (relayer spender) and other unenforced providers are intentionally not bound", and the `SCOPE` block at ~:90 still said the spender gap is "tracked as a separate follow-up… minding cowswap, whose spender is legitimately the GPv2VaultRelayer, not tx.to". Both are now false — cowswap *is* bound, and its spender *is* its `tx.to`. In a fund-safety file a stale comment is how the next person re-opens the hole; refreshed both.

## suggestion

**4. `RECOGNIZED_UNENFORCED_PROVIDERS` is a hand-maintained list with no link to `generalSwapProviders`.**

`jupiter` is a real member of `generalSwapProviders` and is in *neither* set, so it now throws. That's inert today (Solana payloads never reach `getEvmSigningInputs`), and fail-closed is the right default — but the next general provider someone adds becomes a hard signing failure for every payload until they remember this file exists, and nothing points them here. Worth a compile-time exhaustiveness assertion (`satisfies Record<GeneralSwapProvider, 'enforced' | 'unenforced'>` over a single provider→policy map, derived into the two sets) so adding a provider is a `tsc` error rather than a runtime brick. Deliberately not doing this myself: classifying `jupiter` is your call, not a mechanical refactor.

## q

**5. For CowSwap, does the allow-list entry actually buy fund-safety on the swap leg?**

`tx.to` for a cowswap payload is a **constant** `build.ts` always sets — there are no attacker degrees of freedom in it. The artifact actually signed is the EIP-712 order digest decoded from `data`, and `decodeCowSwapKeysignData` validates *shape only*. `assertCowSwapReceiverSafe` rejects zero/burn/malformed receivers, but nothing binds `order.receiver` to the vault address, so a tampered payload can still have the co-signer sign an order paying out to an arbitrary EOA. So the real value this PR delivers for cowswap is the **approval-spender bind** (M4), not the destination allow-list. Is binding `order.receiver` (and `sellToken`/`sellAmount` against `coin`/`toAmount`) on the digest-construction path a planned follow-up? That looks like the actual remaining cowswap co-signer surface, and it's addressable without a proto change.

`packages/core/mpc/tx/compile/compileTx.golden.test.ts > attaches CIP-20 label-674 metadata when memo is present` fails with `TypeError: The first argument must be of type string or an instance of Buffer…`, and `tsc` reports `auxiliaryData does not exist in type ISigningInput` at `resolvers/cardano.ts:72`. Both reproduce identically on a clean `main` — a wallet-core typing drift, unrelated to this PR.

## risk

Low-medium, now bounded. The change is monotonic (throw or no-op, never alters signed bytes, so it cannot desync the cross-device pre-signing hash), the new rejection path is unreachable for every provider string this repo or the iOS client puts on the wire, and the one fail-open it introduced is closed and mutation-covered. Residual risk is the honest one the PR names: `li.fi`/`swapkit`/`''` remain unenforceable by address, and `''` is a one-field-deletion bypass — the proto-oneof follow-up is the real fix and should not be allowed to drift.

`patch` on `@vultisig/core-chain` is right: no wire/schema change, no new required field, and no rejection path that can fire on a legitimately-produced payload.

## verdict

**APPROVE** at `4b6ac168` (fix-and-approve). Findings 1 and 3 fixed on the branch; 2 partially addressed via the changeset; 4 and 5 are non-blocking follow-ups.

<sub>🤖 reviewed by the vultisig-ops reviewer bot — Claude (load-bearing) + Codex `gpt-5.4`. Codex lane returned and independently enumerated the same permissive-default set (`li.fi`/`swapkit`/`''` on both guards, cowswap accepted "on any chain"); deduped at synthesis.</sub>

NeOMakinG (approved):
Round 2 — **merge-advance resolution audit** at `4be16c90` (merge of `main` into the branch). r1 approved at `4b6ac168`; `reviewDecision` never dropped, so this round exists to answer one question: **did the chain-scoping fix survive the merge with its teeth intact?** It did. Re-approving at HEAD.

## testing approach

Worktree `/tmp/wt-vultisig-sdk-1462` checked out at the merged tip `4be16c90`, workspace link-farm rebuilt so `@vultisig/*` resolves to the *worktree* sources (a flat `node_modules` symlink silently resolves them back into the canonical checkout — every result below would have been about `main`, not this PR).

1. **Merge-advance identity proof.** Merge-base moved `52f90bd0` → `9d50ac5c`. md5 of the whitespace-normalised, sorted `+`/`-` line multiset of the net diff vs merge-base:
   - old tip `52f90bd0..4b6ac168` → `8ac65bb218d5920f84c513f5ed5cf102`
   - new tip `9d50ac5c..4be16c90` → `8ac65bb218d5920f84c513f5ed5cf102`

   **Byte-identical** ⇒ zero new author work *and* zero main-side loss. Same 5-file set, `.changeset/cosigner-swap-guard-provider-spoof.md` (incl. the r1 amendment) intact.
2. **Main-side window audit.** `git log 52f90bd0..9d50ac5c` = exactly one commit, `9d50ac5c` "fix(cardano): resolve transaction status with Koios (#1461)", touching only `chain/tx/status/resolvers/cardano.{ts,test.ts}` + a changeset. **No main-side change anywhere under `packages/core/chain/swap/` or the keysign signing-input resolvers.** `cowSwapSupportedChains` (`swap/general/cowswap/config.ts`) untouched in the window — still exactly `[Ethereum, Arbitrum, Base, Avalanche]`, not truncated.
3. **No dep movement.** `git diff 9d50ac5c HEAD -- '*package.json' 'yarn.lock'` is empty ⇒ caret-bump hazard N/A this round.
4. **Fix present at the merged tree**, read from source not from the diff — `knownAggregatorRouters.ts:193`:
   ```ts
   isOneOf(chain, cowSwapSupportedChains) && normalized === COW_VAULT_RELAYER_ADDRESS.toLowerCase()
   ```
   plus the two refreshed comments in `resolvers/evm/index.ts` (the ones that still called cowswap unenforced/unbound) — both survived.
5. **Baseline suite, merged tree:** `knownAggregatorRouters.test.ts` + `swapRouterGuard.test.ts` → **44/44 green** (`vitest --config .config/vitest.core.config.ts`).
6. **Adversarial re-run, end-to-end — not a re-read of the diff.** The resolver-level suite only ever builds `Chain.Ethereum` payloads (`swapRouterGuard.test.ts:32` hardcodes it), so the unsupported-chain case is covered at the *pure-function* level only. I wrote a throwaway probe driving the **real** `getEvmSigningInputs` with a full cowswap payload — swap leg → relayer **and** an `erc20ApprovePayload` whose `spender` is the relayer too, so `assertEnforcedSwapApprovalSpenderBound` is satisfied by construction and the chain predicate is the *only* thing left standing:

   | chains | expectation | result |
   |---|---|---|
   | Cronos, zkSync, Blast, Polygon, BSC, Optimism, Sei | rejected `/unrecognized router address/` | **7/7 reject** |
   | Ethereum, Arbitrum, Base, Avalanche | still emit 2 inputs, swap leg → relayer | **4/4 sign** |

   11/11 green at the merged tree.
7. **Control mutation (the probe is not vacuous).** Strip `isOneOf(chain, cowSwapSupportedChains) &&` from the cowswap arm at the merged tree:
   - repo suites: **7 RED / 37 pass** (44)
   - my end-to-end probe: **7 RED / 4 pass** (11) — i.e. without the predicate all seven unsupported chains sign the ERC-20 approve to the codeless relayer address, through the real resolver.

   Predicate restored, worktree left clean. The fix has teeth *at the merged tree*, on the actual co-signer path, not just in the unit test.
8. `gh pr checks`: Integration Tests (Vitest) **pass**; Unit Tests + Lint/Type Check **pending** at time of writing (not yet a green receipt — see risk).

## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | blocker | cowswap arm matched `COW_VAULT_RELAYER_ADDRESS` with **no chain predicate** ⇒ a payload relabelled `cowswap` on CronosChain/zkSync/Blast (live `eth_getCode` = `0x`) passed *both* guards, getting a co-signer to sign an ERC-20 `approve` to a codeless, later-claimable address | ✅ fixed | fixed by my own `4b6ac168`; **survived `4be16c90` byte-identically** (md5 identity, item 1). Re-proved live end-to-end this round: 7/7 unsupported chains reject, control mutation → 7 RED (items 6–7) |
| r1 | 2 | ℹ️ | two stale comments in `resolvers/evm/index.ts` still describing cowswap as unenforced/unbound | ✅ fixed | refreshed in `4b6ac168`, present verbatim at `4be16c90` |
| r1 | 3 | should-fix | `RECOGNIZED_UNENFORCED_PROVIDERS` is hand-maintained and disjoint from `generalSwapProviders` — the next provider added there silently becomes a hard signing failure | ⏭️ deferred | unchanged at HEAD. Still non-blocking; a `satisfies` exhaustiveness map remains the cheap fix. Classifying `jupiter` is your call, so I did not do it for you |
| r1 | 4 | q | `provider: ''` (the proto3 default you get by *clearing* the field) no-ops **both** guards, so the escape set is `{li.fi, swapkit, ''}` rather than ∅ | ℹ️ documented | unchanged, and the changeset now says so explicitly. Genuinely blocked on a proto oneof discriminant (cross-repo) — tracked in sdk#1457 |
| r1 | 5 | q | CowSwap's real co-signer surface is the EIP-712 order digest in `data`, not `tx.to` (a constant `build.ts` always sets); `assertCowSwapReceiverSafe` rejects only zero/burn/malformed so `order.receiver` can be an arbitrary attacker EOA | ⏭️ deferred | unchanged at HEAD. Still out of scope for this PR, still worth its own issue |

## blocking

None. Nothing on the changed lines regressed across the merge.

## should-fix

**1. `buildCowSwapApprovalSigningInput` is a published subpath that reaches the ERC-20 approve resolver with *neither* guard running** — `packages/core/mpc/keysign/swap/cowswap/buildCowSwapApprovalSigningInput.ts:25-32`.

Pre-existing (landed in #636, untouched by this PR) and **not currently exploitable** — I checked, it has **zero callers** in this repo and zero importers in agent-backend-ts / mcp-ts / vultiagent-app today. So this is not a reason to hold the merge. But it is worth a follow-up, because it is the one place where this PR's headline claim ("cowswap is now enforced") does not hold:

```ts
export const buildCowSwapApprovalSigningInput = ({ keysignPayload, walletCore }: Input) => {
  if (!keysignPayload.erc20ApprovePayload) return undefined
  const signingInput = getErc20ApproveSigningInput({ keysignPayload, walletCore })   // <-- no guard
  return signingInputClasses.evm.encode(signingInput).finish()
}
```

It is exported as `@vultisig/core-mpc/keysign/swap/cowswap/buildCowSwapApprovalSigningInput`, and its own doc comment actively instructs consumers to use it *instead of* `getEncodedSigningInputs` ("CowSwap orders cannot flow through the normal path… the consumer therefore drives the CowSwap ceremony directly"). So the first consumer that wires the real CowSwap ceremony gets an approve-leg builder with no `assertEnforcedSwapApprovalSpenderBound` and no chain-scoped router allow-list — exactly the hole r1 closed on the `getEvmSigningInputs` path, re-opened on the path CowSwap is actually documented to take.

Suggested fix — hoist the same two asserts into the helper before it encodes:

```ts
const chain = getKeysignChain<'evm'>(keysignPayload)
const swapPayload = getKeysignSwapPayload(keysignPayload)
if (swapPayload && 'general' in swapPayload) {
  const { provider, quote } = swapPayload.general
  const to = quote?.tx?.to ?? ''
  assertKnownAggregatorRouterOnSigningPath(provider, to, chain)
  assertEnforcedSwapApprovalSpenderBound(provider, keysignPayload.erc20ApprovePayload.spender, to, chain)
}
```

Both are monotonic (throw or no-op, never change signed bytes), so adding them cannot desync a cross-device pre-signing hash. Happy for this to be a separate PR.

## q

**2.** Was the `main` merge deliberate, or an auto-merge from the GitHub UI? Asking only because the branch is now `MERGEABLE` and fully caught up with `origin/main` (`merge-base == origin/main`), so nothing is pending on your side — if you were waiting on a rebase request from me, you aren't.

## risk

**Low, on the changed lines.** The `+`/`-` multiset is byte-identical across the merge, the main-side window is a disjoint Cardano tx-status change, and the fund-critical predicate is proved live at the merged tree with a control mutation that turns it red. Residual risk is (a) the pre-existing unguarded `buildCowSwapApprovalSigningInput` subpath above, which is latent-only while it has no callers, and (b) the r1 residual `{li.fi, swapkit, ''}` escape set, which needs a proto change to close and is honestly documented in the changeset.

Worth noting for the merger: **Unit Tests and Lint/Type Check were still `pending`** when this was written — the merged-tree evidence here is my local run, not a CI receipt. Wait for those to go green.

## verdict

**Approve** at `4be16c90`. The r1 chain-scoping fix survived the merge intact and still has teeth — re-proved end-to-end through the real co-signer resolver, with a control mutation confirming the probe isn't vacuous. No new blocking findings; nothing manufactured to justify the round.

---
<sub>🤖 two-lane review (Claude + Codex `gpt-5.4`). Codex lane returned and independently surfaced should-fix #1 (the unguarded `buildCowSwapApprovalSigningInput` subpath); I verified its export surface and caller set myself before promoting it, and tiered it down to non-blocking on the zero-callers finding.</sub>

PR #1463, opened 2026-07-21
fix(core-chain): retry broadcast hash verification

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction verification after broadcast errors by retrying transaction status checks briefly while propagation/indexing completes.
  * More reliably recovers from transient RPC/status lookup failures and better handles pending or temporarily unknown states.
  * Stops retrying on definitive error states and preserves the original broadcast error if verification can’t be confirmed within the retry window.
  * Added automated test coverage for retry timing and failure-handling behavior.

### DIFF
--- a/.changeset/quiet-rivers-verify.md
+++ b/.changeset/quiet-rivers-verify.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

Issue #1338, opened 2026-07-16
sui: re-price once more (or clamp reselect) after the gas-budget convergence 2-round cap


gomesalexandre:
Non-blocking suggestion surfaced during #1275's runtime verification (fc487a1 gas-budget convergence loop). Fail-closed, not fund-unsafe - tracked so the residual under-price is closed rather than assumed.

## what
`getSuiChainSpecific`'s convergence loop (packages/core/mpc/keysign/chainSpecific/resolvers/sui/index.ts) re-prices the gas budget when a larger refined budget grows the coin selection, bounded at `maxGasBudgetConvergeIterations = 2`. Runtime probe (built code, adversarial fixture with 5%/round marginal-cost growth): the loop exits by hitting the 2-round cap, and its FINAL `selectCoins(gasBudget)` can grow the selection (e.g. 154 -> 159 objects) beyond what the last dry-run actually priced. Measured true price for the final 159-object selection = 4,140,000 vs the 3,910,000 returned = ~5.9% under-price in that synthetic worst-case.

## why it's safe (not blocking)
Sui rejects an under-budgeted tx PRE-COMMIT (fails closed) - worst case is a failed send needing retry, never a misexecution or fund loss. And it's a synthetic worst-case: real per-object marginal Sui gas is much smaller than the fixture's 5%/round, and the 1.15x budget multiplier (headroom) usually absorbs the residual. rcoderdev's own test comment already documents this as the bound's fail-safe tradeoff.

## fix options
Either: (a) after the loop exits, do ONE final re-price for the final `selectedCoins` (accept one extra dry-run for exactness), or (b) clamp the final `selectCoins` so it doesn't exceed the `pricedCoinCount` the budget was last priced for (trade a slightly-larger-than-needed budget for guaranteed coverage). (a) is more correct; (b) is cheaper. Add a test asserting final gasBudget >= true price of final selection.

Ref: #1275 runtime receipt Part B (comment on the PR). Fund-safety: fail-closed availability, not a loss vector.

gomesalexandre:
Already fixed on main by #1467 (fail-closed throw + matching test coverage).

PR #1467, opened 2026-07-21
fix(sui): fail closed when gas pricing does not converge

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Sui transaction construction now fails with a clear error when gas-budget recalculation cannot converge to a valid coin selection.
  * Prevented returning under-priced transaction baselines when repricing causes the required gas to rise.
  * Added fail-closed behavior when repricing fails or when the refined gas budget exceeds available wallet balance.
  * Improved determinism so repeated transaction preparations select the same coins.
* **Tests**
  * Updated and expanded Sui refine/convergence tests, including a deterministic re-run assertion and stricter convergence failure expectations.


gomesalexandre (approved):
Approve.

This moves the Sui refinement path in the correct fail-closed direction.

What I checked:
- The old `withFallback(...)` wrapped the whole refine-and-converge loop, so a mid-loop repricing/select failure could silently drop back to the static baseline after refinement had already learned a higher requirement. This narrows the fallback to the initial dry-run failure only, which is the right boundary.
- `pricedCoinCount = selectedCoins.length` now happens before the convergence break, so the final repriced selection counts as priced on the happy path. That avoids a false positive from the post-loop guard.
- The post-loop `selectedCoins.length > pricedCoinCount` throw is exactly the fail-closed behavior we want when the coin selection is still growing after the bounded repricing rounds. Better to refuse construction than return a tx that is already known to be under-budgeted.
- The new tests cover the important branches: stable convergence, perpetual growth -> throw after the bounded attempts, mid-loop repricing failure propagation, and refined-budget-exceeds-wallet propagation.

Non-blocking:
- Please add `closes #1338` to the PR body before merge so the issue actually auto-closes.
- Exporting `SuiGasBudgetConvergenceError` would make SDK-boundary handling less stringly typed, but that is follow-up material, not a blocker for this change.

PR #1460, opened 2026-07-21
test(zcash): pin a ZIP-243 golden sighash + branch-id-rotation guard

gomesalexandre:
Closes vultisig/vultisig-sdk#1459

## what
Pins `getSighashZcash` (Zcash's transparent-send ZIP-243 sighash, hand-rolled with BLAKE2b personalization) against an independently-computed golden vector, and strengthens the existing branch-id-rotation coverage - ahead of the Ironwood consensus-branch-id rotation on 2026-07-28.

## verify-first
Checked before writing anything: `utxo-zcash-branchid.test.ts` already existed and proves changing the branch id changes the resulting sighash (the "rotation" half of #1459's ask). What was genuinely missing was the "golden" half - nothing pinned the hand-rolled ZIP-243 digest against an authoritative, independently-computed value. `compileTx.golden.test.ts`'s Zcash case pins WalletCore's *own* output for the same wire shape, a different implementation entirely, and its file header explicitly disclaims covering this builder ("the SDK's independent NU5 v5-aware UTXO builder is verified separately elsewhere in this audit").

Also verified the framing itself is correct: the SDK builds Zcash v4/Sapling-framed transactions with ZIP-243 sighash, **not** v5/ZIP-244/Orchard (issue title says "v5/ZIP-244" but the issue body correctly says "ZIP-243" - the title is loose). Confirmed against zcashd's `ContextualCheckTransaction` (`main.cpp`), which still accepts `SAPLING_VERSION_GROUP_ID` alongside `ZIP225_VERSION_GROUP_ID` post-NU5 - v4 Sapling transparent sends remain consensus-valid indefinitely, so there's no v5 gap to fill here.

## how
- Exported `getSighashZcash` from `packages/sdk/src/chains/utxo/tx.ts` (+ `index.ts` re-export), same pattern as the existing `getSighashBIP143`/`getSighashLegacy` exports. Pure visibility change - zero behavior change to any existing call path (`buildUtxoSendTx` already called it the same way internally).
[...]

PR #1470, opened 2026-07-21
fix(tron): encode the memo into protobuf field 10 (data) not 12 (scripts)

### LINKED ISSUE #1468: fix(tron): memo encodes into protobuf field 12 (scripts) instead of 10 (data)
Found in the Station x Tron full-stack vertical audit (2026-07-21).

**Severity: P0 (fund loss).**

## what

The hand-rolled RN protobuf encoder for a Tron transfer writes the transaction memo (`opts.data`) into `raw_data` field **12** (`scripts`) instead of field **10** (`data`). The WalletCore path encodes the memo into the correct field 10. A Tron send that carries a memo therefore signs and broadcasts with an **empty `data` field** on-chain.

The golden test enshrines the bug rather than catching it: it re-builds the protobuf using the same wrong field assumption, so it passes green against the broken encoder.

## where

- `packages/sdk/src/chains/tron/tx.ts:289-295` - writes `opts.data` to field 12 (`scripts`) instead of field 10 (`data`).
- `packages/sdk/test/.../tron-golden-vectors.test.ts:154` - golden re-encodes with the same field-12 assumption, so it enshrines the bug instead of diffing against WalletCore.
- WalletCore path is correct (encodes memo into field 10).

## why it matters

Any TRC-20 / TRX send **with a memo** is affected, and memos on Tron are load-bearing:
- Exchange deposits use the memo as the destination tag. An empty `data` field means the exchange credits nothing - frequently unrecoverable.
- A THORChain swap memo that lands empty means Bifrost never sees the swap intent.

This is a silent, on-chain, funds-at-risk defect: the tx succeeds mechanically but the value is misrouted or lost.

## fix

- Move the RN encoder (and its golden) from field 12 to field 10 for the memo/`data`.
- Add a cross-check that the hand-rolled memo'd `raw_data` byte-matches the WalletCore memo'd `raw_data` (a real parity diff, not a same-assumption re-encode). This is the seatbelt that would have caught this and would catch future drift.

## notes

- This is a **shared-SDK** change consumed by the flagship iOS / Android / Windows apps too. Regression-test all consumers before republish/bump (SDK cross-consumer caution).
- Related: #1439 point 3 flagged the field-12 memo as an *unverified* parity gap ("asserted by comments only, not a confirmed bug"). This audit **confirms** it is a bug and promotes it to P0.
- Related: #561 covers the separate TRC-20 (`TriggerSmartContract`) memo-semantic question and is not the same defect.

gomesalexandre:
Closes #1468

## what

The hand-rolled RN Tron tx builder (`buildRawData` in `packages/sdk/src/chains/tron/tx.ts`) wrote the transfer memo (`opts.data`) into `raw_data` field **12** (`scripts`, deprecated) instead of field **10** (`data`, the real Tron memo field). WalletCore's keysign path (`packages/core/mpc/keysign/signingInputs/resolvers/tron.ts`) was already correct - it sets `memo: keysignPayload.memo` on the TW proto, which WalletCore encodes to field 10.

The golden test enshrined the bug instead of catching it: `tron-golden-vectors.test.ts`'s "independent" `@bufbuild/protobuf` reference encoder hand-transcribed the same wrong field number, so both "sides" agreed on the wrong answer and stayed green.

## where

- `packages/sdk/src/chains/tron/tx.ts:289-295` (pre-fix) - `buildRawData` wrote `opts.data` to field 12.
- `packages/sdk/tests/unit/platforms/react-native/tron-golden-vectors.test.ts:154` (pre-fix) - reference encoder re-derived the same field-12 assumption.
- `packages/core/mpc/keysign/signingInputs/resolvers/tron.ts` - confirmed correct (field 10) pre-fix, unchanged by this PR.

## why (fund-loss impact)

Any TRC-20/TRX send **with a memo** was affected:
- Exchange deposits use the memo as the destination tag. An empty `data` field means the exchange credits nothing - frequently unrecoverable.
- A THORChain swap memo landing empty means Bifrost never sees the swap intent - funds sent, swap never happens.

## fix

- `buildRawData` now writes the memo to field 10, and the field is emitted **before** field 11 (`contract`) to match Tron's canonical ascending field order - this also matches WalletCore's actual byte layout, not just the field number (verified byte-for-byte, see below).
- `tron-golden-vectors.test.ts` rebuilt against field 10 with the same reordering.
- `tron-tx.test.ts`: all field-12/tag-`0x62` references moved to field-10/tag-`0x52`; the "field absent" checks were switched from a raw substring search (which can false-positive on a `52` straddling two unrelated bytes) to a byte-aligned check; the pinned signing-hash regression test recomputed for the new encoding.
- **New durable guard** - `tron / WalletCore cross-check (fund-safety net)` in `tron-tx.test.ts`: builds an equivalent `TransferContract` via `@trustwallet/wallet-core` (a genuinely independent implementation, not a second hand-transcription of the same spec) and asserts the hand-rolled `raw_data` is **byte-identical** to WalletCore's own `raw_data_hex`, for both a THORChain-swap-memo'd send and a no-memo regression case. This is what would have caught the original bug and will catch any future drift between the RN builder and WalletCore.

## cross-consumer (shared-SDK caution)

- `packages/sdk` ships to the flagship iOS/Android/Windows apps too. Grepped all consumers of `buildTronSendTx`/`buildTrc20TransferTx`: the only non-test consumer is `packages/sdk/src/platforms/react-native/chains/tron/index.ts`, a thin re-export explicitly documented as the RN/Hermes bridge (`tronweb` is avoided because of a `ws`/`node-fetch` cascade it doesn't want to eval at module init). This is the mobile/Station-agent-only code path.
- The flagship WalletCore signing path (`packages/core/mpc/keysign/signingInputs/resolvers/tron.ts` + `packages/core/mpc/tx/compile/compileTx.ts`) is untouched by this PR and was already correct - confirmed via `packages/core/mpc/tx/compile/compileTx.golden.test.ts`'s existing Tron `TransferContract` golden test, which still passes unmodified.
- Grepped for any test/consumer asserting field-12 behavior outside the two files fixed here - none found. Field 12 was a straight bug; nothing legitimately depended on it.
- **No version bump / republish in this PR.** Added a changeset (`.changeset/tron-memo-field-10.md`, `patch`) so the fix is queued for the next release, but the actual `@vultisig/sdk` version bump only happens when a maintainer merges the separate changesets-bot "Version Packages" PR - not this one. Publish/republish timing is a maintainer decision.
- **Consumers needing a regression pass before this ships**: any RN/mobile Tron memo send (exchange deposit tag) and any THORChain Tron-source swap (memo carries the swap intent) - both currently broken pre-fix, both should be spot-checked post-fix before the version bump lands.

## test receipts

`yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/chains/tron-tx.test.ts tests/unit/platforms/react-native/tron-golden-vectors.test.ts`:
```
Test Files  2 passed (2)
     Tests  49 passed (49)
```
Includes the two new WalletCore cross-check tests, both green.

`yarn workspace @vultisig/sdk typecheck`: clean (0 errors after `yarn build:shared`; confirmed identical 29 pre-existing baseline errors on main before `yarn build:shared`, none tron-related, none introduced by this diff).

`yarn workspace @vultisig/sdk test` (full unit suite, post `yarn build:shared`):
```
Test Files  193 passed (193)
     Tests  2929 passed (2929)
```

`yarn test:core` (full core suite, confirms the WalletCore Tron golden test is untouched and still green):
```
Test Files  215 passed | 1 skipped (216)
     Tests  1938 passed | 1 skipped (1939)
```
Tron-specific subset (`-t "Tron"`): 8 files / 45 tests, all passing.

Prettier + eslint clean on all touched files.

🤖 Agent-generated

### DIFF
--- a/packages/sdk/tests/unit/chains/tron-tx.test.ts
+++ b/packages/sdk/tests/unit/chains/tron-tx.test.ts
@@ -5,7 +5,10 @@
  *  - `encodeInt64Varint` for negative values (MUST be 10 bytes, two's-complement)
  *  - `buildTronSendTx` / `buildTrc20TransferTx` byte-level shape
  */
-import { describe, expect, it } from 'vitest'
+import { initWasm, TW, type WalletCore } from '@trustwallet/wallet-core'
+import { Buffer } from 'buffer'
+import Long from 'long'
+import { beforeAll, describe, expect, it } from 'vitest'
 
 import { encodeInt64Varint, encodeVarint } from '../../../src/chains/tron/proto'
 import {
@@ -27,6 +30,17 @@ function bytesToHex(b: Uint8Array): string {
   let s = ''
   for (let i = 0; i < b.length; i++) s += b[i]!.toString(16).padStart(2, '0')
   return s
+}
+
+// A plain `hex.includes(needle)` can false-positive on a needle that
+// straddles a byte boundary (e.g. '52' matching the tail of one byte and the
+// head of the next) even though no actual 0x52 byte is present at a tag
+// position. Only match at even (byte-aligned) offsets.
+function containsAlignedHex(hex: string, needle: string): boolean {
+  for (let i = 0; i <= hex.length - needle.length; i += 2) {
+    if (hex.slice(i, i + needle.length) === needle) return true
+  }
+  return false
 }
 
 describe('tron / encodeVarint', () => {
@@ -182,9 +196,9 @@ describe('tron / buildTrc20TransferTx', () => {
   })
 })
 
-describe('tron / buildTronSendTx memo / data field (proto field 12)', () => {
-  // Baseline: no memo → field 12 must be absent (back-compat guarantee).
-  it('no memo → field 12 tag 0x62 is absent from raw_data bytes', () => {
+describe('tron / buildTronSendTx memo / data field (proto field 10)', () => {
+  // Baseline: no memo → field 10 must be absent (back-compat guarantee).
+  it('no memo → field 10 tag 0x52 is absent from raw_data bytes', () => {
     const tx = buildTronSendTx({
       from: FROM,
       to: TO,
@@ -194,11 +208,13 @@ describe('tron / buildTronSendTx memo / data field (proto field 12)', () => {
       expiration: 1_700_000_000_000n,
       timestamp: 1_699_999_940_000n,
     })
-    // 0x62 is the tag for field 12 wire type 2; must be absent when no memo.
-    expect(tx.unsignedRawHex).not.toContain('62')
+    // 0x52 is the tag for field 10 wire type 2; must be absent when no memo.
+    // Byte-aligned check — a plain substring search can false-positive on a
+    // '52' that straddles two unrelated bytes (see `containsAlignedHex`).
+    expect(containsAlignedHex(tx.unsignedRawHex, '52')).toBe(false)
   })
 
-  it('empty Uint8Array memo → treated as absent, no field 12', () => {
+  it('empty Uint8Array memo → treated as absent, no field 10', () => {
     const tx = buildTronSendTx({
       from: FROM,
       to: TO,
@@ -209,10 +225,10 @@ describe('tron / buildTronSendTx memo / data field (proto field 12)', () => {
       timestamp: 1_699_999_940_000n,
       data: new Uint8Array(0),
     })
-    expect(tx.unsignedRawHex).not.toContain('62')
+    expect(containsAlignedHex(tx.unsignedRawHex, '52')).toBe(false)
   })
 
-  it('THORChain swap memo → field 12 present with UTF-8 encoded bytes', () => {
+  it('THORChain swap memo → field 10 (the real Tron memo field) present with UTF-8 encoded bytes', () => {
     const memo = 'SWAP:BTC.BTC:bc1qabcdef1234567890:1000000'
     const memoBytes = new TextEncoder().encode(memo)
     const tx = buildTronSendTx({
@@ -225,10 +241,10 @@ describe('tron / buildTronSendTx memo / data field (proto field 12)', () => {
       timestamp: 1_699_999_940_000n,
       data: memoBytes,
     })
-    // field 12 tag = 0x62, length varint = memo.length (< 128 so 1 byte), then data.
+    // field 10 tag = 0x52, length varint = memo.length (< 128 so 1 byte), then data.
     const expectedLen = memoBytes.length.toString(16).padStart(2, '0')
     const expectedMemoHex = bytesToHex(memoBytes)
-    expect(tx.unsignedRawHex).toContain('62' + expectedLen + expectedMemoHex)
+    expect(tx.unsignedRawHex).toContain('52' + expectedLen + expectedMemoHex)
   })
 
   it('memo changes the signing hash (pre-signing hash stability check)', () => {
@@ -268,8 +284,8 @@ describe('tron / buildTronSendTx memo / data field (proto field 12)', () => {
     // canonical value, then it fails on any deviation. The test itself asserts
     // length (64 hex chars = 32 bytes = SHA-256) and stability across runs.
     expect(tx.signingHashHex).toMatch(/^[0-9a-f]{64}$/)
-    // Pinned value established from first correct run:
-    expect(tx.signingHashHex).toBe('5f14bf8b2f6681f04a433beb93084b85f2d003ce337a2d72ea013d307a4e7841')
+    // Pinned value established from first correct run after the field-10 fix:
+    expect(tx.signingHashHex).toBe('e9b2b75390c656f9135f9379bae5a27e14dbfb6e9a2857004bf6d4f51ff30c12')
   })
 
   it('long memo (100+ bytes) is encoded correctly with varint length prefix', () => {
@@ -291,7 +307,7 @@ describe('tron / buildTronSendTx memo / data field (proto field 12)', () => {
     // varint(130) = 0x82 0x01 (two bytes: 130 & 0x7f | 0x80 = 0x82, then 1).
     const expectedVarint = '8201'
     const expectedMemoHex = bytesToHex(memoBytes)
-    expect(tx.unsignedRawHex).toContain('62' + expectedVarint + expectedMemoHex)
+    expect(tx.unsignedRawHex).toContain('52' + expectedVarint + expectedMemoHex)
     expect(tx.signingHashHex).toMatch(/^[0-9a-f]{64}$/)
   })
 
@@ -321,37 +337,38 @@ describe('tron / buildTronSendTx memo / data field (proto field 12)', () => {
 
     it('127-byte memo → 1-byte varint length prefix (0x7f)', () => {
       const memo = memoOf(127)
-      const expected = '62' + '7f' + bytesToHex(memo)
+      const expected = '52' + '7f' + bytesToHex(memo)
       expect(rawHexFor(memo)).toContain(expected)
     })
 
     it('128-byte memo → 2-byte varint length prefix (0x8001)', () => {
       const memo = memoOf(128)
-      const expected = '62' + '8001' + bytesToHex(memo)
+      const expected = '52' + '8001' + bytesToHex(memo)
       expect(rawHexFor(memo)).toContain(expected)
     })
 
     it('16383-byte memo → 2-byte varint length prefix (0xff7f)', () => {
       const memo = memoOf(16383)
-      const expected = '62' + 'ff7f' + bytesToHex(memo)
+      const expected = '52' + 'ff7f' + bytesToHex(memo)
       expect(rawHexFor(memo)).toContain(expected)
     })
 
     it('16384-byte memo → 3-byte varint length prefix (0x808001)', () => {
       const memo = memoOf(16384)
-      const expected = '62' + '808001' + bytesToHex(memo)
+      const expected = '52' + '808001' + bytesToHex(memo)
       expect(rawHexFor(memo)).toContain(expected)
     })
   })
 })
 
-describe('tron / buildTrc20TransferTx memo / data field (proto field 12)', () => {
+describe('tron / buildTrc20TransferTx memo / data field (proto field 10)', () => {
   // TRC-20 path mirrors the native send path: memo goes on the *wrapping*
-  // Transaction's raw_data.data (field 12), NOT the inner contract-call
-  // data (which is the ABI-encoded transfer payload). Exchanges that
-  // require user-tag memos to credit TRC-20 USDT deposits (Binance, OKX,
-  // KuCoin) rely on this field being present and correctly encoded.
-  it('no memo → field 12 tag 0x62 is absent from raw_data bytes', () => {
+  // Transaction's raw_data.data (field 10, the real Tron memo field), NOT
+  // the inner contract-call data (which is the ABI-encoded transfer
+  // payload). Exchanges that require user-tag memos to credit TRC-20 USDT
+  // deposits (Binance, OKX, KuCoin) rely on this field being present and
+  // correctly encoded.
+  it('no memo → field 10 tag 0x52 is absent from raw_data bytes', () => {
     const tx = buildTrc20TransferTx({
       from: FROM,
       to: TO,
@@ -363,29 +380,25 @@ describe('tron / buildTrc20TransferTx memo / data field (proto field 12)', () =>
       expiration: 1_700_000_000_000n,
       timestamp: 1_699_999_940_000n,
     })
-    // 0x62 is the field-12 tag (12<<3|2 = 0x62). When absent it must not
+    // 0x52 is the field-10 tag (10<<3|2 = 0x52). When absent it must not
     // appear in the tail of the raw bytes. The inner contract data lives
     // under a different tag (field 4 of TriggerSmartContract = 0x22)
     // wrapped inside Any(field 2)→Contract(field 2)→Raw(field 11), so
-    // 0x62 is only emitted when field 12 itself is set.
+    // 0x52 is only emitted when field 10 itself is set.
     const rawHex = tx.unsignedRawHex
     // Field 18 (fee_limit) tag = 18<<3|0 = 0x90 0x01. Locate it as a
-    // boundary marker; everything after must be the feeLimit varint, no
-    // stray 0x62 tag.
+    // boundary marker to confirm the tail of the tx is well-formed.
     const feeLimitIdx = rawHex.indexOf('900180c2d72f')
     expect(feeLimitIdx).toBeGreaterThan(-1)
-    // The bytes preceding fee_limit are: timestamp tag (0x70) + varint.
-    // The bytes preceding timestamp would be field 12 if present. Assert
-    // the timestamp tag follows directly after the contract block by
-    // checking no 0x62-tag-prefixed length-delimited block sits between
-    // contract end and timestamp tag — simplest signal is just length.
-    // Match what the native-send "no memo" assertion does: scan for the
-    // 0x62 tag. The contract value never legitimately ends with 0x62
-    // followed by a valid varint length so this is a tight check.
-    expect(rawHex).not.toContain('6204') // any short memo wouldn't appear
+    // Scan for the 0x52 tag anywhere in the raw bytes (byte-aligned — a plain
+    // substring search can false-positive on a '52' straddling two unrelated
+    // bytes, see `containsAlignedHex`). The contract value never legitimately
+    // produces a 0x52-tag-prefixed length-delimited block, so this is a
+    // tight check for field-10 absence.
+    expect(containsAlignedHex(rawHex, '5204')).toBe(false) // any short memo wouldn't appear
   })
 
-  it('empty Uint8Array memo → treated as absent, no field 12 emitted', () => {
+  it('empty Uint8Array memo → treated as absent, no field 10 emitted', () => {
     const tx = buildTrc20TransferTx({
       from: FROM,
       to: TO,
@@ -398,10 +411,10 @@ describe('tron / buildTrc20TransferTx memo / data field (proto field 12)', () =>
       timestamp: 1_699_999_940_000n,
       data: new Uint8Array(0),
     })
-    expect(tx.unsignedRawHex).not.toContain('6204')
+    expect(containsAlignedHex(tx.unsignedRawHex, '5204')).toBe(false)
   })
 
-  it('exchange deposit memo → field 12 present with UTF-8 encoded bytes', () => {
+  it('exchange deposit memo → field 10 (the real Tron memo field) present with UTF-8 encoded bytes', () => {
     // Binance-style TRC-20 USDT deposit memo (numeric user tag).
     const memo = '103456789'
     const memoBytes = new TextEncoder().encode(memo)
@@ -419,7 +432,7 @@ describe('tron / buildTrc20TransferTx memo / data field (proto field 12)', () =>
     })
     const expectedLen = memoBytes.length.toString(16).padStart(2, '0')
     const expectedMemoHex = bytesToHex(memoBytes)
-    expect(tx.unsignedRawHex).toContain('62' + expectedLen + expectedMemoHex)
+    expect(tx.unsignedRawHex).toContain('52' + expectedLen + expectedMemoHex)
   })
 
   it('memo changes the signing hash vs no-memo TRC-20 (pre-signing stability)', () => {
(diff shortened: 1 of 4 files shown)


packages/sdk/tests/unit/platforms/react-native/tron-golden-vectors.test.ts
```diff
@@ -106,10 +106,13 @@ function encodeRawData(opts: {
   w.tag(1, WireType.LengthDelimited).bytes(opts.refBlockBytes)
   w.tag(4, WireType.LengthDelimited).bytes(opts.refBlockHash)
   w.tag(8, WireType.Varint).int64(opts.expiration)
-  w.tag(11, WireType.LengthDelimited).bytes(contractBytes)
+  // Field 10: `raw_data.data`, the real Tron memo field. Sorts before field 11
+  // (contract) in ascending field-number order, matching Tron's own encoder
+  // and WalletCore's output (see the cross-check in tron-tx.test.ts).
   if (opts.data && opts.data.length > 0) {
-    w.tag(12, WireType.LengthDelimited).bytes(opts.data)
+    w.tag(10, WireType.LengthDelimited).bytes(opts.data)
```
NeOMakinG, line 113: `should-fix:` the golden-vector reference encoder is STILL self-confirming — i proved it with a co-mutation, it stays 5/5 green — the PR body's own diagnosis ("the golden test enshrined the bug… both sides agreed on the wrong answer") applies to this file **unchanged after the fix**. the reference encoder still hand-transcribes its tag numbers, so it can only ever catch a typo on one side, never a wrong shared assumption.

proof — **M5**, co-mutate production AND the reference to a wrong-but-consistent field 13:
```
src/chains/tron/tx.ts:307      fieldBytes(10, …) -> fieldBytes(13, …)
golden-vectors.test.ts:113     w.tag(10, …)      -> w.tag(13, …)

tron-golden-vectors.test.ts   =>  Test Files 1 passed (1)   Tests 5 passed (5)   <-- GREEN
tron-tx.test.ts               =>  Test Files 1 failed (1)   Tests 13 failed | 31 passed
```
the golden suite is green on a memo written to a field that does not exist in `Tron.proto`. the ONLY thing that bites is the new WalletCore cross-check — which lives in a different file. codex's lane reached the same conclusion independently.

for completeness the single-sided mutations all do bite, so the PR is not shipping a dead test — it just isn't shipping an independent one here:
```
M1 memo -> field 12 (fix reverted)  tron-tx 13 failed | golden 1 failed
M3 memo dropped entirely            tron-tx 15 failed | golden 1 failed
M4 emit BOTH field 10 and 12        tron-tx  2 failed | golden 1 failed
```
note M4: only **2** tests in `tron-tx.test.ts` catch a spurious extra field-12 alongside a correct field 10 — and they are exactly the two new WalletCore byte-identity tests. every hand-transcribed `toContain('52' + len + memo)` assertion is happy. that is the clearest possible statement of which guard is load-bearing here.

fix: either (a) drop the memo assertion from this file entirely — post-fix it contributes zero independent signal and its presence implies coverage that doesn't exist, and let the WalletCore cross-check own the field-number question; or (b) make it genuinely independent by generating the reference from a checked-in vendored copy of `tronprotocol/protocol core/Tron.proto` rather than re-typing tags. (a) is the honest one-line change.

packages/sdk/tests/unit/chains/tron-tx.test.ts
```diff
@@ -601,3 +614,118 @@ describe('tron / buildTronTxFromRawData (prebuilt raw_data signing)', () => {
...
+// the SAME wrong field number as its "independent" reference. A second
+// encoding of the same wrong assumption catches typos, not wrong assumptions.
+//
+// This block asserts the hand-rolled `raw_data` bytes are BYTE-IDENTICAL to
+// what WalletCore itself produces for an equivalent transaction — WalletCore
+// is a genuinely different implementation (used by the iOS/Android/Windows
+// co-signers), so it cannot silently share this encoder's mistake. If the RN
+// builder ever again diverges from WalletCore on field placement, ordering,
+// or memo encoding, this test fails.
+describe('tron / WalletCore cross-check (fund-safety net)', () => {
```
NeOMakinG, line 633: `should-fix:` the WalletCore cross-check — the one genuinely independent guard — covers `buildTronSendTx` only, not `buildTrc20TransferTx` — both tests in this describe build `TW.Tron.Proto.TransferContract`. TRC-20 is untouched — yet TRC-20 USDT is the *dominant* case the PR body itself cites ("Binance / OKX / KuCoin require this for crediting deposits"). so the memo field on the exchange-deposit path is still guarded only by hand-transcribed `'52'` string assertions, i.e. by exactly the mechanism that failed here.

this is not a hypothetical gap. i wrote the missing test (`TW.Tron.Proto.TransferTRC20Contract`, same block header / refs / feeLimit, memo `'103456789'`) and it is **RED today** — see the `pre-existing bugs` section for the two byte-level divergences. the good news buried in that run: WalletCore's TRC-20 `raw_data` also carries the memo as `52 09 3130333435363738 39` — **field 10** — so this PR's fix is independently confirmed correct on the TRC-20 path too, even though nothing in the suite asserts it.

fix: add the TRC-20 arm of the cross-check. it will fail until the two divergences below are settled — which is precisely the point, and arguably the more valuable outcome of this PR. if you'd rather not expand scope, land this PR as-is and open the TRC-20 cross-check as an immediate follow-up rather than letting the "can never silently diverge from WalletCore again" claim in the changeset stand unqualified for the TRC-20 path.

NeOMakinG (approved):
gm — multi-lane pass: claude (load-bearing) + codex gpt-5.4. runtime-tested on worktree at PR head `ebe27059` (`/tmp/wt-vultisig-sdk-1470`), vitest `tests/unit/vitest.config.ts` — 49/49 green on both touched suites, external authority: `tronprotocol/protocol@master core/Tron.proto` fetched live — `bytes data = 10;` (L438), `bytes scripts = 12;` (L442), live mainnet: 3 confirmed USDT (`TR7NHqj…jLj6t`) TriggerSmartContract txs pulled from TronGrid and byte-inspected, WalletCore WASM probe: `TW.Tron.Proto.TransferContract` + `TransferTRC20Contract` signed and `raw_data_hex` diffed against both builders, mutation battery M1/M3/M4/M5 + full-suite control run at merge-base `a4d8bbe8`. lgtm.

## testing approach
- worktree at PR head `ebe27059` (`/tmp/wt-vultisig-sdk-1470`), vitest `tests/unit/vitest.config.ts` — 49/49 green on both touched suites
- external authority: `tronprotocol/protocol@master core/Tron.proto` fetched live — `bytes data = 10;` (L438), `bytes scripts = 12;` (L442)
- live mainnet: 3 confirmed USDT (`TR7NHqj…jLj6t`) TriggerSmartContract txs pulled from TronGrid and byte-inspected
- WalletCore WASM probe: `TW.Tron.Proto.TransferContract` + `TransferTRC20Contract` signed and `raw_data_hex` diffed against both builders
- mutation battery M1/M3/M4/M5 + full-suite control run at merge-base `a4d8bbe8`
- pass order: claude (load-bearing) → codex gpt-5.4

## should-fix
### 1. the golden-vector reference encoder is STILL self-confirming — i proved it with a co-mutation, it stays 5/5 green — `packages/sdk/tests/unit/platforms/react-native/tron-golden-vectors.test.ts:113`
the PR body's own diagnosis ("the golden test enshrined the bug… both sides agreed on the wrong answer") applies to this file **unchanged after the fix**. the reference encoder still hand-transcribes its tag numbers, so it can only ever catch a typo on one side, never a wrong shared assumption.

proof — **M5**, co-mutate production AND the reference to a wrong-but-consistent field 13:
```
src/chains/tron/tx.ts:307      fieldBytes(10, …) -> fieldBytes(13, …)
golden-vectors.test.ts:113     w.tag(10, …)      -> w.tag(13, …)

tron-golden-vectors.test.ts   =>  Test Files 1 passed (1)   Tests 5 passed (5)   <-- GREEN
tron-tx.test.ts               =>  Test Files 1 failed (1)   Tests 13 failed | 31 passed
```
the golden suite is green on a memo written to a field that does not exist in `Tron.proto`. the ONLY thing that bites is the new WalletCore cross-check — which lives in a different file. codex's lane reached the same conclusion independently.

for completeness the single-sided mutations all do bite, so the PR is not shipping a dead test — it just isn't shipping an independent one here:
```
M1 memo -> field 12 (fix reverted)  tron-tx 13 failed | golden 1 failed
M3 memo dropped entirely            tron-tx 15 failed | golden 1 failed
M4 emit BOTH field 10 and 12        tron-tx  2 failed | golden 1 failed
```
note M4: only **2** tests in `tron-tx.test.ts` catch a spurious extra field-12 alongside a correct field 10 — and they are exactly the two new WalletCore byte-identity tests. every hand-transcribed `toContain('52' + len + memo)` assertion is happy. that is the clearest possible statement of which guard is load-bearing here.
action: either (a) drop the memo assertion from this file entirely — post-fix it contributes zero independent signal and its presence implies coverage that doesn't exist, and let the WalletCore cross-check own the field-number question; or (b) make it genuinely independent by generating the reference from a checked-in vendored copy of `tronprotocol/protocol core/Tron.proto` rather than re-typing tags. (a) is the honest one-line change.

### 2. the WalletCore cross-check — the one genuinely independent guard — covers `buildTronSendTx` only, not `buildTrc20TransferTx` — `packages/sdk/tests/unit/chains/tron-tx.test.ts:633`
both tests in this describe build `TW.Tron.Proto.TransferContract`. TRC-20 is untouched — yet TRC-20 USDT is the *dominant* case the PR body itself cites ("Binance / OKX / KuCoin require this for crediting deposits"). so the memo field on the exchange-deposit path is still guarded only by hand-transcribed `'52'` string assertions, i.e. by exactly the mechanism that failed here.

this is not a hypothetical gap. i wrote the missing test (`TW.Tron.Proto.TransferTRC20Contract`, same block header / refs / feeLimit, memo `'103456789'`) and it is **RED today** — see the `pre-existing bugs` section for the two byte-level divergences. the good news buried in that run: WalletCore's TRC-20 `raw_data` also carries the memo as `52 09 3130333435363738 39` — **field 10** — so this PR's fix is independently confirmed correct on the TRC-20 path too, even though nothing in the suite asserts it.
action: add the TRC-20 arm of the cross-check. it will fail until the two divergences below are settled — which is precisely the point, and arguably the more valuable outcome of this PR. if you'd rather not expand scope, land this PR as-is and open the TRC-20 cross-check as an immediate follow-up rather than letting the "can never silently diverge from WalletCore again" claim in the changeset stand unqualified for the TRC-20 path.

## q
- CI at `ebe27059` is not green yet at review time — `Unit Tests` and `Lint and Type Check` are still **pending**, `Integration Tests (Vitest)` passed (5m32s), `CodeRabbit` shows `fail` but the reason is "Review rate limited", not a real finding. approving on my own local receipts (49/49 on both touched suites, plus a full-suite control against merge-base); please just confirm the two pending jobs land green before merging.

- **`buildTrc20TransferTx` does not produce WalletCore's bytes — verified against 3 real mainnet USDT txs.** measured with `TW.Tron.Proto.TransferTRC20Contract` vs `buildTrc20TransferTx` on identical inputs. two divergences, BOTH pre-existing (present on `a4d8bbe8`, untouched by this PR), neither fund-losing as far as i can establish:
  1. **`call_value` emitted explicitly as zero.** our `TriggerSmartContract` writes `1800` (field 3, varint 0); WalletCore omits it, and so do all 3 confirmed mainnet USDT transfers i pulled from TronGrid (inner length `0x74` vs our `0x76`). valid protobuf, decodes identically — just non-canonical, so our txID will never match what tronweb would compute for the same intent.
  2. **ABI recipient word padding.** ours left-pads the **20-byte** address; WalletCore *and* mainnet left-pad the **21-byte `41`-prefixed** address:
```
mainnet  a9059cbb 000000000000000000000041 d0e5529ced1b91492e01b5e5c2326ab856d7df19 …
WalletCore a9059cbb 000000000000000000000041 99c851eaa3c3976914d63b822c67e201ec0bfbb8 …
ours       a9059cbb 000000000000000000000000 99c851eaa3c3976914d63b822c67e201ec0bfbb8 …
```
  the `41` sits at byte 11 of the word, i.e. in the padding region above the low 20 bytes the TVM actually ABI-decodes, so both forms should resolve to the same recipient — which is presumably why this has never surfaced. i have **not** verified that on-chain with a broadcast, so i'm flagging rather than sizing it. worth a real testnet TRC-20 send before anyone relies on the "byte parity with WalletCore" framing for TRC-20. (the existing golden assertion "ABI-encodes the ERC-20 transfer selector/recipient/amount identically to hand computation" is, again, hand computation — it agrees with us, not with the chain.)
- `vultiagent-app/src/services/tronTx.ts:333` still documents the memo as "raw_data.data (proto field 12)". stale after this lands — one-line doc follow-up in that repo. the app's on-device prebuilt-raw_data decoder itself is memo-field-agnostic (it only pins field 11 / Contract and rejects unknown wire types; field 10 is wire type 2 and passes), so there is no decode asymmetry to worry about — just the comment.

## risk
**what i independently confirmed (not taken from the PR body):**
1. **field 10 is right, externally.** fetched `tronprotocol/protocol@master core/Tron.proto` live — `message Transaction.raw` L438 `bytes data = 10;`, L442 `bytes scripts = 12;` (both carry upstream's own `// data not used` / `// scripts not used` comments). independent of this repo and of WalletCore.
2. **the two paths now agree byte-for-byte.** `packages/core/mpc/keysign/signingInputs/resolvers/tron.ts` is untouched and its `memo:` wiring lands in field 10 — the new cross-check asserts `ours.unsignedRawHex === walletCoreRawDataHex` for a memo'd and a no-memo native send, and both are green at head. that equality *is* the two-paths-agree proof, and it's the strongest artifact in the diff.
3. **the fix bites.** M1 (revert to field 12) → 13 failed in `tron-tx.test.ts` + 1 in golden; M3 (drop memo) → 15 + 1; M4 (emit both 10 and 12) → 2 + 1. no green-under-mutation on the production change itself.
4. **blast radius is nil.** zero in-repo readers of field 12 / `scripts`. only non-test consumer of the builders is the RN barrel `src/platforms/react-native/chains/tron/index.ts`. `buildTronTxFromRawData` takes opaque prebuilt bytes so there's no old-built / new-parsed asymmetry; the app-side decoder in `vultiagent-app/src/services/tronTx.ts` pins only field 11 and is memo-field-agnostic.
5. **the `-66` is accounted for.** in `tx.ts` exactly **3** non-comment lines were deleted (`fieldInt64(8, …)`, `fieldBytes(11, contract)`, `fieldBytes(12, opts.data)`) — all three reappear, reordered. everything else removed is comment prose or reworked test assertions. nothing load-bearing went with it.
6. **`patch` is the right changeset level** for a bugfix that doesn't move the public API. it does change emitted bytes for existing callers, but the old bytes were wrong, so a minor bump would misrepresent it as opt-in.
7. **full-suite control.** the 61 failing files i saw locally are an unbuilt-`build:shared` worktree artifact, not this PR: merge-base `a4d8bbe8` gives the identical `61 failed | 159 passed`, `8 failed | 2414 passed`; head gives `61 failed | 159 passed`, `8 failed | **2416** passed` — a clean +2, the two new cross-check tests.

**risk:** low for the change itself. field 10 is externally corroborated (not from this repo, not from the PR), the memo path is mutation-proven RED under every plausible regression i threw at it, and the two paths (RN builder vs WalletCore keysign) are now proven byte-identical for native TRX. cross-repo blast radius is nil: nothing in-repo or in vultiagent-app/agent-backend-ts/mcp-ts reads raw_data field 12, and `buildTronTxFromRawData` passes prebuilt bytes through opaquely so there is no old-builds-vs-new-parses asymmetry to strand an in-flight tx. residual risk is concentrated in the TRC-20 path, which the new independent guard does not cover and which i measured to already diverge from WalletCore/mainnet in two (pre-existing, non-fund-losing) ways.

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4-codex` adversarial pass. worktree at PR head `ebe27059` (`/tmp/wt-vultisig-sdk-1470`), vitest `tests/unit/vitest.config.ts` — 49/49 green on both touched suites, external authority: `tronprotocol/protocol@master core/Tron.proto` fetched live — `bytes data = 10;` (L438), `bytes scripts = 12;` (L442), live mainnet: 3 confirmed USDT (`TR7NHqj…jLj6t`) TriggerSmartContract txs pulled from TronGrid and byte-inspected, WalletCore WASM probe: `TW.Tron.Proto.TransferContract` + `TransferTRC20Contract` signed and `raw_data_hex` diffed against both builders, mutation battery M1/M3/M4/M5 + full-suite control run at merge-base `a4d8bbe8`.

NeOMakinG (approved):
gm — multi-lane pass: claude merge-resolution audit. conceptual-only (no runtime this pass). lgtm.

## prior issues — status
| round | issue | status |
|---|---|---|
| r1 | should-fix #1 — `tron-golden-vectors.test.ts` is still self-confirming: the "independent" `@bufbuild` reference encoder is a second hand-transcription, so co-mutating prod + reference to a nonexistent field 13 stays 5/5 GREEN. Only the WalletCore byte-identity cross-check (different file) bites. | ⏭️ deferred — unchanged by the merge (test blob byte-identical at `ebe27059` and `14a075ea`). Follow-up, not blocking. |
| r1 | should-fix #2 — the genuinely independent WalletCore cross-check guards `buildTronSendTx` (native TRX) only, not `buildTrc20TransferTx`. I wrote the missing TRC-20 arm and it goes RED on two *pre-existing* divergences: we emit `call_value = 0` explicitly (WalletCore + 3 real mainnet USDT txs omit it, proto3 default), and we left-pad the 20-byte address where WalletCore/mainnet left-pad the 21-byte `41`-prefixed form. | ⏭️ deferred — pre-existing, out of scope for this PR; `tx.ts` blob byte-identical across the merge so neither divergence moved. |
| r1 | q — CI (`Unit Tests`, `Lint and Type Check`) was still pending when I approved on local receipts. | ⚠️ still pending — the merge kicked a fresh run (`29830183250`); `Integration Tests` / `Lint` / `Unit Tests` all `pending`, `CodeRabbit` fail = "Review rate limited" (not a finding). Worth a glance before merge. |
| r1 | ℹ️ pre-existing — `vultiagent-app/src/services/tronTx.ts:333` doc comment still says "proto field 12". The decoder itself pins only field 11 and rejects unknown *wire types*, so it is memo-field-agnostic; comment only. | ℹ️ documented — app-repo follow-up, unaffected. |

## risk
nil delta this round — `ebe27059` → `14a075ea` is a pure merge of `main` (`e92fa0cb`) with zero conflict resolution and zero new author work. Receipts:

**1. per-file `+`/`-` line-multiset md5, net-diff-vs-merge-base, old tip vs new tip — all four IDENTICAL:**
```
IDENTICAL  .changeset/tron-memo-field-10.md                          9f13a81d…
IDENTICAL  packages/sdk/src/chains/tron/tx.ts                       ff7714dd…
IDENTICAL  packages/sdk/tests/unit/chains/tron-tx.test.ts           f72928b9…
IDENTICAL  .../react-native/tron-golden-vectors.test.ts             41b0bbfd…
```
(merge-base moved `a4d8bbe8` → `e92fa0cb`; identical hashes at both ⇒ nothing added, nothing lost.)

**2. the merge diffed against BOTH parents partitions with no remainder:**
- vs parent1 (`ebe27059`, PR side) = exactly the 5 zcash ZIP-243 files from #1460, nothing else
- vs parent2 (`e92fa0cb`, main side) = exactly the 4 tron files of this PR, nothing else

**3. main-side advance is file- and import-disjoint.** `a4d8bbe8..e92fa0cb` is a single commit — #1460, zcash/UTXO only. `tron/tx.ts` imports only `./proto`, `@noble/hashes`, `bs58check`; the golden test imports `src/chains/tron/tx` by relative path. No `package.json` / lockfile movement, so no caret-bump hazard.

**4. blob identity.** `tron/tx.ts`, `tron/proto.ts`, both tron test files and the RN tron barrel are all byte-identical between `ebe27059` and `14a075ea`, and `tx.ts:307` at the merge tip still reads `concatProtoBytes(raw, fieldBytes(10, opts.data))`. Because the merged tree is byte-identical on every file in the reachable import set, r1's mutation battery (M1 revert 10→12 ⇒ 13+1 RED; M4 emit both ⇒ 2+1 RED) carries over unchanged — I did not re-run it, and say so rather than claim a fresh green.

Codex lane skipped honestly this round: a provably-disjoint merge advance leaves no concrete second-opinion question to scope it to.

## verdict
APPROVE — ship it 🫡

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `skipped this round — merge-audit only` conceptual pass.

PR #1471, opened 2026-07-21
fix(tron): validate triggerconstantcontract + add fee_limit margin (out_of_energy resilience)

### LINKED ISSUE #1469: fix(tron): fee_limit set to exact estimate with no margin + triggerconstantcontract result never validated
Found in the Station x Tron full-stack vertical audit (2026-07-21).

**Severity: HIGH (gas cluster).**

## what

A cluster of Tron fee-estimation defects on the flagship MPC signing path:

1. The `triggerconstantcontract` energy-estimate response is never validated. On an empty response (`{}`) or a TronGrid revert-while-indexing, the estimate falls through to `feeLimit = 0`.
2. `feeLimit` is set to the **exact** estimate with **no safety margin**. Energy can drift during the 10-60s MPC ceremony, and the tx then fails `OUT_OF_ENERGY` (fee burned, transfer fails).
3. Two divergent fee paths exist and should converge on "estimate + margin, capped".

## where

- `packages/sdk/src/chains/tron/fee.ts:56-58` - unvalidated `triggerconstantcontract` result -> `feeLimit` of 0 on `{}` / indexing-revert.
- `packages/sdk/src/chains/tron/fee.ts:73-89` - `feeLimit` = exact estimate, no margin.
- `packages/sdk/src/.../signingInputs/resolvers/tron.ts:138,149,221,284`.
- `packages/sdk/src/chains/tron/energyPrice.ts:4,8-9` - `FALLBACK_ENERGY_PRICE=280` is stale (real = 100 sun).

## why it matters

- **feeLimit = 0 on revert/empty**: a guaranteed failure whenever TronGrid returns a revert while indexing - the transfer cannot be paid for.
- **No margin**: `OUT_OF_ENERGY` on energy drift during the MPC ceremony burns the fee and fails the transfer. This is the flagship MPC path, so blast radius is real.
- Same cluster, lower severity (all in the safe/overestimate direction, noted for completeness):
  - TRC-20 fee display omits the ~0.345 TRX bandwidth burn (`fee.ts` resolver, display-only).
  - `FALLBACK_ENERGY_PRICE=280` overestimates ~2.8x vs the real 100 sun; only fires on a TronGrid outage, so it is a safe-direction overestimate - refresh the constant + comment.

## fix

- Add a safety margin to the estimate (estimate + margin, capped).
- Fail-closed on an empty/revert estimate response - do **not** default `feeLimit` to 0.
- Converge the two divergent fee paths onto the single "estimate + margin, capped" model.
- Refresh `FALLBACK_ENERGY_PRICE` to the real 100 sun and update the stale comment.

## notes

Shared-SDK change consumed by flagship iOS / Android / Windows too - regression-test all consumers before republish/bump.

gomesalexandre:
Closes vultisig/vultisig-sdk#1469

## what

Two compounding bugs in `getTrc20TransferFee` (TRC-20 fee estimation for MPC keysign) that guarantee `OUT_OF_ENERGY` on broadcast for a chunk of Tron transfers:

1. **Unvalidated triggerconstantcontract response -> feeLimit 0.** The response parse defaulted `energy_used`/`energy_penalty` to 0 with no check that TronGrid actually returned a successful simulation (`result.result === true`). An empty body, indexing lag, or a reverted simulation all silently produced `totalEnergy = 0n` -> the early `return 0n` -> `feeLimit = 0` downstream. Now throws instead, bubbling through the same throw-bubbling contract already tested for network errors.

2. **No margin on feeLimit.** The function returned the *exact* energy-burn estimate as `feeLimit`. That field is a spending ceiling, not an expected cost - dynamic energy pricing and concurrent staked-energy use during the 10-60s MPC signing ceremony can drift the real cost above a stale point-in-time estimate. Now pads +50% and caps at 100 TRX (matching the existing send-service ceiling at `packages/sdk/src/chains/tron/tx.ts:121`).

Bonus: `FALLBACK_ENERGY_PRICE` was a stale 280 sun/energy (2023 governance default). Live-verified via `curl -sX POST https://api.trongrid.io/wallet/getchainparameters` -> `getEnergyFee: 100`. Corrected to 100n; this only fires on a TronGrid outage.

## why

Margin raises the ceiling only - users pay actual on-chain energy consumed, never the padded feeLimit. Safe direction for a fund-safety-adjacent fee estimator: a too-low feeLimit hard-fails the broadcast, a padded-but-uncharged ceiling doesn't cost the user anything extra.

## how

- `packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts` - response-shape validation (throw on `!result.result`) + `FEE_LIMIT_MARGIN_BPS`/`FEE_LIMIT_CAP_SUN` applied to the return value.
- `packages/core/mpc/keysign/chainSpecific/resolvers/tron/energyPrice.ts` - fallback 280n -> 100n (live value).
- `fee.test.ts` - flipped the test that was locking in the bug (asserted `fee === 0n` for `{}`) to assert the throw; added a revert-response throw case and a feeLimit-cap case; all existing fee-value assertions updated to account for the margin.
- `energyPrice.test.ts` - all 5 fallback assertions `280n` -> `100n`.

## test plan

- [x] `yarn vitest run --config .config/vitest.core.config.ts` on `fee.test.ts` + `energyPrice.test.ts` - 20/20 pass
- [x] `yarn test:core` (full core suite) - 1940 passed, 1 skipped, no failures/regressions
- [x] `yarn tsc --project .config/tsconfig.json --noEmit` - clean
- [x] lint (`scripts/run-eslint.mjs`) on changed files - clean
- [x] `curl -sX POST https://api.trongrid.io/wallet/getchainparameters` live-verified `getEnergyFee: 100`

## risk

Patch changeset only, no version bump/publish in this PR. `@vultisig/core-mpc` is shared across flagship iOS/Android/Windows + this SDK - before republishing, a maintainer should run a real TRC-20 send/swap regression pass across consumers per the SDK cross-consumer-caution rule. Isolated to the Tron TRC-20 fee path; margin only raises the spending ceiling, never the amount actually charged.

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * TRC-20 transfers on Tron now fail fast on unreliable/malformed simulations and reject non-positive simulated energy instead of silently returning a `feeLimit` of `0` (helping prevent `OUT_OF_ENERGY` failures).
  * FeeLimit is now computed from full simulated energy using a +50% padded ceiling capped at 100 TRX; the displayed/max-send burn remains based on staked-energy adjustment.
  * Tron energy-price fallback updated to 100 sun/energy with more conservative handling when chain parameters are missing/invalid.
  * TRON fee resolution now correctly distinguishes native TRON vs TRC-20 and returns the appropriate fee fields.

* **Tests**
  * Expanded coverage for simulation error shapes, pathological estimates (100 TRX cap), feeLimit/display separation, and fallback behavior.


packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts
```diff
@@ -86,5 +98,13 @@ export const getTrc20TransferFee = async ({ coin, receiver, amount }: GetTrc20Tr
...
   const totalSun = energyToBurn * energyPrice
 
-  return totalSun
+  // feeLimit is a spending CEILING, not an expected cost. Pad for drift during the
+  // 10-60s MPC ceremony (dynamic energy pricing, concurrent staked-energy use),
+  // cap so a pathological estimate can't runaway. 100 TRX cap matches the existing
+  // send-service ceiling (packages/sdk/src/chains/tron/tx.ts:121).
+  const FEE_LIMIT_MARGIN_BPS = 5_000n // +50%
+  const FEE_LIMIT_CAP_SUN = 100_000_000n // 100 TRX
+  const withMargin = totalSun + (totalSun * FEE_LIMIT_MARGIN_BPS) / 10_000n
```
NeOMakinG: `preferably-blocking:` the +50% margin is applied to `energyToBurn` (post staked-energy subtraction), so it is exactly 0 in the scenario the PR was written for — the PR's stated motivation is that "concurrent staked-energy use during the 10-60s MPC signing ceremony can drift the real cost above a stale point-in-time estimate". but the margin multiplies `totalSun = energyToBurn * energyPrice` (:99), and `energyToBurn` is what's left *after* subtracting the sender's staked energy (:88-93). so for a user whose staked energy covers the call, `energyToBurn = 0n` → `totalSun = 0n` → `withMargin = 0n` → the function returns a bare `0n`, and +50% of 0 is 0. that is precisely the user who is exposed to the cited race: if another tx eats their staked energy mid-ceremony, the broadcast now needs real TRX burn but ships `feeLimit = 0`.

the PR closes ONE route to `feeLimit=0` (the malformed `triggerconstantcontract` body) and leaves two others fully intact:
- `totalEnergy <= 0n` → early `return 0n` (:79), which returns *before* the margin block
- `availableEnergy >= totalEnergy` → `energyToBurn = 0n` → `0n` (:90)

that the zero is load-bearing and reaches the wire is confirmed end-to-end: `getTrc20TransferFee` → `gasEstimation` (`resolvers/tron/index.ts:75`) → `feeLimit: Long.fromString(tronSpecific.gasEstimation.toString())` (`keysign/signingInputs/resolvers/tron.ts:138,149,221,284`) with no clamp and no minimum. and this repo already documents the consequence at `packages/sdk/src/chains/tron/tx.ts:118-121`: *"Must be > 0n — a zero feeLimit would be silently dropped from the raw data, causing the node to reject the tx with a cryptic OUT_OF_ENERGY error."*

note this is coupled to the should-fix below and is **not** a one-line change: `gasEstimation` is also the user-displayed fee, so simply switching the basis to `totalEnergy` would show a staked-energy user a fee they will never be charged. i deliberately did not fix-and-approve this one — splitting ceiling-from-expected-cost touches a proto field shared with flagship iOS/Android/Windows and is your call, not mine.

control mutation M5 (`totalSun = totalEnergy * energyPrice`) turns 4 tests RED, so the zero-fee behaviour is deliberately pinned today by `returns 0 when sender has more staked energy than needed (fully covered)` and `returns 0 when sender has exactly the energy needed (boundary)` — any fix has to consciously re-baseline those two.

fix: either (a) base the ceiling on `totalEnergy` rather than `energyToBurn` and add a non-zero floor so `feeLimit` is never 0 — on Tron the ceiling costs the user nothing when staked energy covers the call, since staked energy is consumed first and free — and split the displayed fee onto a separate value; or (b) if you want to keep this PR minimal, at minimum add a floor on the two `return 0n` paths so a zero can never reach `feeLimit`, and say explicitly in the changeset that the staked-energy-covered race is still open.

packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts
```diff
@@ -86,5 +98,13 @@ export const getTrc20TransferFee = async ({ coin, receiver, amount }: GetTrc20Tr
   const energyPrice = await getEnergyPrice()
   const totalSun = energyToBurn * energyPrice
 
-  return totalSun
+  // feeLimit is a spending CEILING, not an expected cost. Pad for drift during the
+  // 10-60s MPC ceremony (dynamic energy pricing, concurrent staked-energy use),
+  // cap so a pathological estimate can't runaway. 100 TRX cap matches the existing
+  // send-service ceiling (packages/sdk/src/chains/tron/tx.ts:121).
+  const FEE_LIMIT_MARGIN_BPS = 5_000n // +50%
```
NeOMakinG: `should-fix:` the padded ceiling is also the user-displayed fee — a normal USDT transfer now shows 34.64 TRX instead of the actual ~23.09 TRX — `gasEstimation` is doing double duty. it is the protobuf `feeLimit` ceiling, and it is *also* what the fee-amount resolver returns verbatim as the displayed fee: `fee/resolvers/tron.ts:7` → registered as the `tron` resolver at `fee/index.ts:39`. so the +50% pad does not only raise the ceiling — it inflates the number shown to the user and the amount deducted in max-send (`sdk/src/tools/prep/maxSend.ts:36,71`).

quantified against a live `triggerconstantcontract` USDT simulation i ran just now (`energy_used: 130285`, `energy_penalty: 100635` = 230,920 energy) at the live `getEnergyFee: 100`:
- exact cost: 23,092,000 sun = **23.09 TRX**
- padded: 34,638,000 sun = **34.64 TRX** ← what the user is shown

the PR body says *"users pay actual on-chain energy consumed, never the padded feeLimit"*. that is true of the **charge** but not of the **display** — the direction is conservative (over-disclose, never over-charge) so this isn't fund loss, but an ~11.5 TRX overstatement will also block a user holding between 23.09 and 34.64 TRX from a send they can actually afford.

fix: keep the pad on the protobuf `feeLimit` only, and have `getTronFeeAmount` return the unpadded expected cost. if that's too invasive for this PR, note the display inflation in the changeset so consumers aren't surprised.

NeOMakinG (commented):
gm — multi-lane pass: claude (load-bearing) + codex gpt-5.4. runtime-tested on worktree `/tmp/wt-vultisig-sdk-1471` @ `7ec52dfd`, rebuilt `@vultisig/*` link farm (a symlinked `node_modules` resolves back to the canonical checkout and would have silently tested `main`), `vitest --config .config/vitest.core.config.ts` on the tron resolver dir — 26/26 green baseline, 5-mutation battery on both new suites — 5/5 RED, live TronGrid probes: `getchainparameters` (2 independent nodes) + a real `triggerconstantcontract` USDT simulation. comment-only.

## testing approach
- worktree `/tmp/wt-vultisig-sdk-1471` @ `7ec52dfd`, rebuilt `@vultisig/*` link farm (a symlinked `node_modules` resolves back to the canonical checkout and would have silently tested `main`)
- `vitest --config .config/vitest.core.config.ts` on the tron resolver dir — 26/26 green baseline
- 5-mutation battery on both new suites — 5/5 RED
- live TronGrid probes: `getchainparameters` (2 independent nodes) + a real `triggerconstantcontract` USDT simulation
- pass order: claude (load-bearing) → codex gpt-5.4

## preferably-blocking
### 1. the +50% margin is applied to `energyToBurn` (post staked-energy subtraction), so it is exactly 0 in the scenario the PR was written for — `packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts:107`
the PR's stated motivation is that "concurrent staked-energy use during the 10-60s MPC signing ceremony can drift the real cost above a stale point-in-time estimate". but the margin multiplies `totalSun = energyToBurn * energyPrice` (:99), and `energyToBurn` is what's left *after* subtracting the sender's staked energy (:88-93). so for a user whose staked energy covers the call, `energyToBurn = 0n` → `totalSun = 0n` → `withMargin = 0n` → the function returns a bare `0n`, and +50% of 0 is 0. that is precisely the user who is exposed to the cited race: if another tx eats their staked energy mid-ceremony, the broadcast now needs real TRX burn but ships `feeLimit = 0`.

the PR closes ONE route to `feeLimit=0` (the malformed `triggerconstantcontract` body) and leaves two others fully intact:
- `totalEnergy <= 0n` → early `return 0n` (:79), which returns *before* the margin block
- `availableEnergy >= totalEnergy` → `energyToBurn = 0n` → `0n` (:90)

that the zero is load-bearing and reaches the wire is confirmed end-to-end: `getTrc20TransferFee` → `gasEstimation` (`resolvers/tron/index.ts:75`) → `feeLimit: Long.fromString(tronSpecific.gasEstimation.toString())` (`keysign/signingInputs/resolvers/tron.ts:138,149,221,284`) with no clamp and no minimum. and this repo already documents the consequence at `packages/sdk/src/chains/tron/tx.ts:118-121`: *"Must be > 0n — a zero feeLimit would be silently dropped from the raw data, causing the node to reject the tx with a cryptic OUT_OF_ENERGY error."*

note this is coupled to the should-fix below and is **not** a one-line change: `gasEstimation` is also the user-displayed fee, so simply switching the basis to `totalEnergy` would show a staked-energy user a fee they will never be charged. i deliberately did not fix-and-approve this one — splitting ceiling-from-expected-cost touches a proto field shared with flagship iOS/Android/Windows and is your call, not mine.

control mutation M5 (`totalSun = totalEnergy * energyPrice`) turns 4 tests RED, so the zero-fee behaviour is deliberately pinned today by `returns 0 when sender has more staked energy than needed (fully covered)` and `returns 0 when sender has exactly the energy needed (boundary)` — any fix has to consciously re-baseline those two.
action: either (a) base the ceiling on `totalEnergy` rather than `energyToBurn` and add a non-zero floor so `feeLimit` is never 0 — on Tron the ceiling costs the user nothing when staked energy covers the call, since staked energy is consumed first and free — and split the displayed fee onto a separate value; or (b) if you want to keep this PR minimal, at minimum add a floor on the two `return 0n` paths so a zero can never reach `feeLimit`, and say explicitly in the changeset that the staked-energy-covered race is still open.

## should-fix
### 1. the padded ceiling is also the user-displayed fee — a normal USDT transfer now shows 34.64 TRX instead of the actual ~23.09 TRX — `packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts:105`
`gasEstimation` is doing double duty. it is the protobuf `feeLimit` ceiling, and it is *also* what the fee-amount resolver returns verbatim as the displayed fee: `fee/resolvers/tron.ts:7` → registered as the `tron` resolver at `fee/index.ts:39`. so the +50% pad does not only raise the ceiling — it inflates the number shown to the user and the amount deducted in max-send (`sdk/src/tools/prep/maxSend.ts:36,71`).

quantified against a live `triggerconstantcontract` USDT simulation i ran just now (`energy_used: 130285`, `energy_penalty: 100635` = 230,920 energy) at the live `getEnergyFee: 100`:
- exact cost: 23,092,000 sun = **23.09 TRX**
- padded: 34,638,000 sun = **34.64 TRX** ← what the user is shown

the PR body says *"users pay actual on-chain energy consumed, never the padded feeLimit"*. that is true of the **charge** but not of the **display** — the direction is conservative (over-disclose, never over-charge) so this isn't fund loss, but an ~11.5 TRX overstatement will also block a user holding between 23.09 and 34.64 TRX from a send they can actually afford.
action: keep the pad on the protobuf `feeLimit` only, and have `getTronFeeAmount` return the unpadded expected cost. if that's too invasive for this PR, note the display inflation in the changeset so consumers aren't surprised.

## q
- the 100 TRX cap checks out as a ceiling, but `tx.ts:121` isn't actually a "ceiling" — it's a doc comment saying *"typical value is ~100 TRX"*, with no enforcement anywhere in `buildTrc20Transfer`. the number is defensible on its own (i verified the cap only binds above 666,666 energy = 2.89x the live-measured normal USDT transfer, so a normal transfer pads to 34.64 TRX and cannot silently drift near 100 TRX) — worth rewording the comment at fee.ts:104 so it doesn't imply an enforced ceiling exists upstream.

- the two `return 0n` paths in `getTrc20TransferFee` predate this PR — flagged above only because eliminating `feeLimit=0` is this PR's own headline claim.
- post-PR, a malformed TronGrid body now rejects the fee-**estimate** and max-send paths (`getSendFeeEstimate.ts:9`, `maxSend.ts:71`) where they previously returned a bogus `0n`. i consider this an improvement — a `0n` fee in max-send would over-send — but it is a behaviour change for any consumer that renders those without a catch.

## risk
**mutation battery — 5/5 RED, this suite actually bites.** M1 delete the `result.result === true` validation → 2 RED · M2 drop the margin (`withMargin = totalSun`) → 5 RED · M3 restore the stale `280n` fallback → 5 RED · M4 remove the 100 TRX cap → 1 RED · M5 base the ceiling on `totalEnergy` → 4 RED. M3 lives in a *different module* (`energyPrice.ts`) and still went red, which is what proves the link farm was exercising the worktree and not the canonical checkout. three PRs today shipped suites that stayed green under mutation — this one does not, nice work.

**claims verified independently of the repo** (per the sdk#1470 lesson that repo-internal Tron wire evidence isn't trustworthy on its own):
- `FALLBACK_ENERGY_PRICE = 100n` — live `getEnergyFee: 100` on api.trongrid.io *and* api.tronstack.io. correct ✅
- `result.result === true` — a real success response is `{"result":{"result":true},"energy_used":130285,"energy_penalty":100635,...}`, so the predicate is right and a successful simulation does not omit it ✅
- throw-escape (codex lane) — the throw fires while building `blockchainSpecific` (`send/build.ts:108`, `swap/build.ts:332`), strictly *before* `extractMessageHashes()` and `sign()` (`VaultBase.ts:1754,1772,1773`) and before the relay session opens (`RelaySigningService.ts:300`). it cannot strand a partially-signed Tron tx ✅. first catches are `wrapAsVaultError` at `TransactionBuilder.ts:44,139` and `SwapService.ts:131`
- the `-41` lines — all test-assertion churn plus the bug-locking `defaults missing energy fields to 0` test flipped to assert the throw. nothing load-bearing removed ✅. changeset `patch` on `@vultisig/core-mpc` is right for a bugfix ✅

**overall: medium-low.** everything the PR asserts checks out. the residual risk is that the headline claim is only half delivered — the margin does not cover the staked-energy race it was written for, and the displayed fee is now inflated ~50%. merging as-is introduces no regression: the zero-`feeLimit` routes are exactly as broken as they are on `main`, and the fallback-price correction is a real fix. not blocking merge — flagging so the remaining half doesn't get closed out along with #1469.

## verdict
COMMENTED

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4-codex` adversarial pass. worktree `/tmp/wt-vultisig-sdk-1471` @ `7ec52dfd`, rebuilt `@vultisig/*` link farm (a symlinked `node_modules` resolves back to the canonical checkout and would have silently tested `main`), `vitest --config .config/vitest.core.config.ts` on the tron resolver dir — 26/26 green baseline, 5-mutation battery on both new suites — 5/5 RED, live TronGrid probes: `getchainparameters` (2 independent nodes) + a real `triggerconstantcontract` USDT simulation.

packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts
```diff
@@ -86,5 +98,13 @@ export const getTrc20TransferFee = async ({ coin, receiver, amount }: GetTrc20Tr
...
+  const FEE_LIMIT_CAP_SUN = 100_000_000n // 100 TRX
+  const withMargin = totalSun + (totalSun * FEE_LIMIT_MARGIN_BPS) / 10_000n
+
+  return withMargin > FEE_LIMIT_CAP_SUN ? FEE_LIMIT_CAP_SUN : withMargin
```
NeOMakinG: preferably-blocking: carried from r1 (still unresolved at this HEAD — fee.ts is byte-identical, the HEAD advance was a pure main-merge). +50% of 0n is 0n: the margin multiplies totalSun = energyToBurn * energyPrice, and energyToBurn is 0n for a sender whose staked energy covers the call (:90) — the exact race this PR targets — and the totalEnergy<=0n path returns 0n before this block (:78). Both ship feeLimit=0, which tron/tx.ts drops -> OUT_OF_ENERGY. Base the ceiling on totalEnergy+floor and split display off the protobuf field, or add a floor on the two zero paths and document the open race.

packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts
```diff
@@ -53,6 +54,17 @@ export const getTrc20TransferFee = async ({ coin, receiver, amount }: GetTrc20Tr
     },
   })
 
+  // triggerconstantcontract can return a 200 with an empty/malformed body or a
+  // reverted simulation (result.result === false) without ever surfacing an
+  // HTTP error. Trusting energy_used=0 in that case silently produces
+  // feeLimit=0 downstream, which guarantees OUT_OF_ENERGY at broadcast. Throw
+  // instead so it bubbles through the same throw-bubbling contract as network
+  // errors above.
+  if (!responseData.result || responseData.result.result !== true) {
```
NeOMakinG: q: a live triggerconstantcontract probe of a reverting transfer returned {"result":{"result":true,"message":"REVERT opcode executed"},"energy_used":8624,...} — result.result === true on a revert. So this !== true guard (and the PR revert test fixturing result:{result:false}) may be inert for real reverts; the !responseData.result empty-body branch is doing the actual work. Not a regression. Consider also failing when a revert message/code is present if you want the revert path to bite.

NeOMakinG (commented):
gm — round 2. HEAD advanced `7ec52dfd` → `e91993f9`, but that advance is a **pure `main`-merge** ("Merge branch 'main' into fix-tron-fee-limit-1469", brings in the memo-field #1470 + ripple/evm-chainid changes). The core resolver `packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts` is **byte-identical to r1** — I diffed `7ec52dfd...e91993f9` and `fee.ts` is not in the changed-files set. So both r1 findings carry forward unchanged. Re-posting at the new HEAD so our position isn't pinned to a stale commit; verdict stays COMMENTED (not hard-blocking gomes, but the two `feeLimit=0` routes are unresolved).

## testing approach
- confirmed HEAD `e91993f9` and that `7ec52dfd...e91993f9` touches only `main`-merged files (memo-field #1470 in `tron/tx.ts`, ripple rpc, evm-chainid, cli notifications) — **not** `fee.ts`/`energyPrice.ts` logic
- re-read `fee.ts` at HEAD via the contents API: both zero routes physically present (lines 78, 90) and the margin still multiplies a base a prior subtraction can zero (99→107)
- live TronGrid re-probe (`POST /wallet/triggerconstantcontract`, single curl, `--max-time 10`, no retry) — see q below
- confirmed the downstream consequence is intact on the merged tree: `tron/tx.ts` still emits `fee_limit` only when `> 0n` and its doc comment still reads *"Must be > 0n — a zero feeLimit would be silently dropped … cryptic OUT_OF_ENERGY"*
- codex lane (gpt-5.3): **SKIPPED — 0-byte stall** (the infra-SIGTERM pattern, not a clean bill)

## prior issues — status
|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | +50% margin applied to `energyToBurn` (post staked-energy subtraction) → `feeLimit=0` in the very race the PR targets; two zero routes survive | ❌ **still blocking** | `fee.ts` byte-identical (untouched by the `main`-merge). Route A `totalEnergy<=0n → return 0n` (:78) and route B `availableEnergy>=totalEnergy → energyToBurn=0n → totalSun=0n → withMargin=0n` (:90,99,107) both intact. `+50%` of `0n` is `0n`. |
| r1 | 2 | should-fix | padded ceiling is *also* the user-displayed fee — normal USDT transfer shows ~34.64 TRX vs ~23.09 actual | ❌ **still open** | `gasEstimation` still double-duties as protobuf `feeLimit`, the `fee/resolvers/tron.ts` displayed value, and the max-send deduction. Unchanged. |
| r1 | 3 | q | `tx.ts:121` isn't an enforced "ceiling", it's a doc comment | ℹ️ **still accurate** | after the merge the same doc comment sits ~:122-127; still a comment on `BuildTrc20TransferOptions.feeLimit`, not enforcement. `fee.ts:104` comment's `tx.ts:121` line ref is now ~4 lines low. |

## preferably-blocking
### 1. carried from r1 — two `feeLimit=0` routes survive; the margin can't cover the staked-energy race it was written for — `packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts:107`
Nothing in this round touches the fee logic. Both zero routes are still live on every branch:
- **route A** — `totalEnergy <= 0n` early `return 0n` (:78), before the margin block ever runs.
- **route B** (the one the PR was written for) — a sender whose staked energy covers the call gets `energyToBurn = 0n` (:90) → `totalSun = 0n` (:99) → `withMargin = 0n` (:107) → returns `0n`. If a concurrent tx eats that staked energy mid-ceremony, the broadcast now needs real TRX burn but ships `feeLimit = 0`, which `tx.ts` silently drops → OUT_OF_ENERGY. That is exactly the population the +50% margin was supposed to protect.

I did **not** fix-and-approve this, same reasoning as r1 and I still agree with it: the only correct fix bases the ceiling on `totalEnergy` (not `energyToBurn`) with a non-zero floor **and** splits the displayed/expected cost off the protobuf `feeLimit` — and `gasEstimation` is a proto field shared with flagship iOS/Android/Windows, so changing its display semantics is an architecture call that belongs to you, not a reviewer's unilateral in-code patch. Raising the ceiling without that split would show a staked-energy user a fee they'll never pay.
action: (a) base the ceiling on `totalEnergy` + non-zero floor and return the unpadded expected cost as the displayed fee; or (b) keep it minimal but add a floor on the two zero paths and state in the changeset that the staked-energy-covered race is still open. Either closes the blocker; today's diff does neither.

## q
### 1. live revert shape carries `result.result === true` — the `result.result === false` branch/test may be inert for real reverts — `packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts:63`
A single live `triggerconstantcontract` probe of a reverting transfer came back as `{"result":{"result":true,"message":"REVERT opcode executed"},"energy_used":8624,"energy_penalty":6640}` — i.e. `result.result === true` **and** a revert message, with non-zero energy. The new guard keys off `result.result !== true`, and the PR's revert test fixtures `result:{result:false}`. If real reverts commonly present as `result:true` + a `message`, the `false` branch rarely fires and the empty-body branch (`!responseData.result`) is doing all the real work. This is **not** a regression (main also proceeds on that shape, and here energy is non-zero so it wouldn't hit `feeLimit=0`), and the empty-body guard — the PR's actual headline fix — is genuinely useful. Just flagging that the fixture may not match the live revert envelope; consider also treating a present `result.message`/revert `code` as a failed estimate if you want the revert path to actually bite.

## risk
**medium-low, unchanged from r1.** No new regression: the zero-`feeLimit` routes are exactly as broken as on `main`, and the fallback-price correction (280n→100n, live-verified on two nodes in r1) is a real fix. The residual is that the PR's headline ("eliminate OUT_OF_ENERGY from feeLimit=0") is only half delivered — the empty-body route is closed, the two arithmetic-zero routes are not. The `main`-merge adds no coupling to the fee logic; the memo-field #1470 change it carries is orthogonal (field 10 vs 11 ordering, unrelated to `fee_limit` field 18).

## verdict
COMMENTED

🤖 reviewed via [Claude Code](https://claude.com/claude-code). round 2 @ `e91993f9` (HEAD advance is a pure `main`-merge; `fee.ts` byte-identical to r1). codex `gpt-5.3-codex` lane SKIPPED (0-byte stall). live TronGrid re-probe (single curl, `--max-time 10`).

packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.ts
gomesalexandre: pushed `4cbfcee0c` to address the current-head review notes.

what changed:
- feeLimit ceiling now uses the full simulated TRC-20 energy before any staked-energy subtraction, then applies the +50% pad and 100 TRX cap, so concurrent staked-energy drift during the MPC window can no longer collapse the serialized ceiling to zero
- user-visible fee amount is separated from the serialized feeLimit ceiling for TRC-20 keysigns: display/max-send now uses the expected burn after current staked-energy coverage, while the proto still carries the padded spending ceiling
- triggerconstantcontract now fail-closes on the live revert shape where `result.result === true` but TronGrid still returns a revert message/code
- fallback docs/comments now explicitly cover unreachable plus missing/invalid `getEnergyFee` parameter responses

receipts from the sparse checkout:
- `vitest run packages/core/mpc/keysign/chainSpecific/resolvers/tron/fee.test.ts packages/core/mpc/keysign/chainSpecific/resolvers/tron/energyPrice.test.ts` ✅ 23/23 passing
- `prettier --check` on the touched files ✅ clean after formatting

known limitation of this receipt environment:
- `index.test.ts` still could not run from this sparse worktree because the generated `@vultisig/core-mpc/types/.../blockchain_specific_pb` package resolution is incomplete here. I left that to CI on the pushed full checkout.

re-requesting review on the new head.

NeOMakinG (approved):
gm — round 3. HEAD `7ec52dfd`(r1) → `e91993f9`(r2, pure main-merge) → **`4cbfcee0`** (a real +449/-240 rework titled *"separate fee limit ceiling from displayed burn"*). This rework closes all three r1/r2 findings. Verified against HEAD in a worktree with a control-mutation battery proving the tests bite the reworked source.

## testing approach
- worktree at `4cbfcee0`, node_modules link-farm (canonical `@vultisig/*`, worktree source loaded by relative `./fee` import — proven live by mutations below), `vitest run` on `fee.test.ts` + `energyPrice.test.ts`: **23/23 green**
- mutation battery on the worktree `fee.ts` (each grep-confirmed applied, then restored):
  - **M-A** non-positive `throw` → `return 0n` (old route-A bug): **2 RED** (`rejects zero`, `rejects negative`)
  - **M-B** drop `+50%` margin → return `totalSun`: **8 RED** (margin actually applied, non-zero)
  - **M-C** display resolver → `return feeLimit` raw (old inflation bug): **6 RED**
  - **M-D** skip `triggerconstantcontract` validation (`if(false)`): **4 RED**
- traced wiring: `getTrc20TransferFee` → `gasEstimation` (index.ts:75) → protobuf `feeLimit` (signingInputs/resolvers/tron.ts) = the padded ceiling; `getTrc20TransferFeeAmount` → `getTronFeeAmount` → `getFeeAmount` → `refine/amount.ts` + `getSendFeeEstimate` = display/max-send. **Not inert.**
- confirmed `refineKeysignAmount` only deducts for fee coins (native TRX, `isNativeToken` → returns `gasEstimation` directly) — the TRC-20 margin reversal never over-deducts a max-send
- codex lane: SKIPPED (dead lane, per instruction — not a clean bill)

## prior issues — status
|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | two `feeLimit=0` routes; +50% margin applied to a base (`energyToBurn`) a staked-energy subtraction can zero | ✅ **fixed** @4cbfcee0 | route A (`totalEnergy<=0n`) now **throws** (M-A 2 RED). route B: the feeLimit base moved to the **full** `totalEnergy` — staked-energy subtraction is removed from the feeLimit path entirely (grep: no `getTronAccountResources`/`availableEnergy` in `getTrc20TransferFee`), so the margin can no longer be zeroed (M-B 8 RED). No zero route remains: `totalEnergy>0` guarded, `energyPrice>=100n`, so `withMargin>0`, capped at 100 TRX. |
| r1 | 2 | should-fix | padded ceiling was *also* the user-displayed fee (~34.64 vs ~23.09 TRX) | ✅ **fixed** @4cbfcee0 | new `getTrc20TransferFeeAmount` reverses the margin off the ceiling + subtracts current staked energy → display = **unpadded expected burn**; `getTronFeeAmount` rewired to call it (M-C 6 RED). No new proto field needed — display derived from the existing `gasEstimation` at resolve time, so no iOS/Android/Windows schema break. |
| r1 | 3 | q | `tx.ts:121` is a doc comment, not an enforced ceiling | ℹ️ **still accurate / informational** | unchanged; informational only, no action. |
| r2 | q1 | q | live revert shape `result.result===true` + revert `message` may leave the `false`-branch inert | ✅ **fixed** @4cbfcee0 | new guard treats a present `result.code`/`result.message` as a failed estimate: `if (!result \|\| result.result !== true \|\| result.code \|\| result.message)`; dedicated test *"throws for the live revert shape"* (M-D 4 RED). Fails **closed**. |

## q
### 1. displayed fee (expected burn) is intentionally below the signed `feeLimit` ceiling — `fee/resolvers/tron.ts`
By design now: the signed protobuf `feeLimit` is the +50% padded ceiling (a max, never the charge on Tron), while the displayed/max-send fee is the unpadded expected burn — matching iOS `TronService.swift` intent. The only case where the actual charge exceeds the displayed number is if broadcast energy consumption exceeds simulation, bounded by the +50% headroom. This is the correct resolution of r1#2 (de-inflated display), not a residual — flagging only so it's not misread as an unresolved divergence.

## risk
**low.** The two arithmetic-zero routes and the malformed/empty/revert routes to `feeLimit=0` are all now fail-closed (throw), the margin base can't be zeroed, and the display is de-inflated without a shared-proto schema change. The `280n→100n` fallback (live-verified on two nodes in r1) is retained. The large `-240` is mostly prettier reformat + replacing the three silent-`0n` returns with throws + moving staked subtraction to the display fn — nothing protective dropped; new tests are strictly stricter.

## verdict
APPROVE

🤖 reviewed via [Claude Code](https://claude.com/claude-code). round 3 @ `4cbfcee0` (real +449/-240 rework). worktree mutation battery M-A/B/C/D all RED. codex lane skipped.

rcoderdev (approved):

NeOMakinG (approved):
## testing approach

Focused re-review of the single new commit `4cbfcee0..1b515396` ("chore: bump sdk in tron fee limit changeset for release guard"). We approved on `4cbfcee0`; this pass verifies the new commit does not disturb the approved fund-critical Tron logic. Ran Claude (byte-identity diff of every touched Tron file) + Codex second-opinion lane on the changeset delta.

## prior issues — status

|---|---|---|---|---|---|
| 1 | 1 | preferably-blocking | +50% margin applied after a subtraction that can zero the base | ℹ️ documented | resolved earlier: ceiling now based on full simulated energy *before* staked-energy subtraction, capped 100 TRX; displayed/max-send fee stays the unpadded post-stake burn. Byte-identical at HEAD. |
| 1 | 2 | q | `feeLimit` field double-duty (ceiling vs displayed) | ✅ addressed | ceiling/expected-cost split shipped; unchanged this commit. |

## blocking

None.

## should-fix

None.

## risk

Minimal. The delta is a **single line** — `"@vultisig/sdk": patch` added to `.changeset/tron-fee-limit-out-of-energy.md`. All 20 fund-critical Tron source files (energyPrice/fee resolvers, signingInputs, broadcast, hash, status, balance, metadata, maxSend, fee/index) are **byte-identical** to the approved `4cbfcee0` (verified per-file via `git rev-parse <sha>:<path>`). The bump is an internal-package changeset semver directive at `patch` — not a floated dependency, no caret/npm-latest hazard. `@vultisig/sdk` is a workspace consumer of `@vultisig/core-mpc`; the patch entry propagates the Tron fix to downstream SDK consumers under the version they install. Codex concurs (repo changesets config already sets `updateInternalDependencies: patch`; at worst a conservative extra publish).

## verdict

**APPROVE.** The new commit is a release-guard changeset bump only; the approved Tron fund-critical logic is untouched and no version/dependency hazard is introduced.

---
🤖 Reviewed by Claude (Opus 4.8) + Codex (gpt-5) lanes. Focused re-review of `4cbfcee0..1b515396`.

PR #1473, opened 2026-07-21
Expose native swap broadcast guard from core

### LINKED ISSUE #1392: THORChain inbound-vault guard is dead code for every shipping wallet: move `assertNativeSwapReadyForBroadcast` from packages/sdk into packages/core
## Summary

`assertNativeSwapReadyForBroadcast` (`packages/sdk/src/vault/services/nativeSwapBroadcastGuard.ts`) is the only implementation anywhere in the Vultisig stack that re-validates a THORChain/Maya inbound vault against a live `/inbound_addresses` fetch before funds move. It is correct, covers THORChain and Maya, and is tested.

**No shipping wallet calls it.** It lives in the `packages/sdk` facade, reachable only via `BroadcastService.broadcastTx`. Windows — the one core app that consumes this stack — broadcasts through `@vultisig/core-chain` directly and never touches the facade.

Verified in `vultisig-windows`:
- `core/ui/mpc/keysign/broadcastKeysignTx.ts:1` — `import { broadcastTx } from '@vultisig/core-chain/tx/broadcast'`
- grep across `clients/`, `core/`, `lib/`, `scripts/` for `assertNativeSwapReadyForBroadcast`, `nativeSwapBroadcastGuard`, `BroadcastService` → **zero hits**
- the guard ships inside `node_modules/@vultisig/sdk/dist/index.browser.js` but is unreachable from the wallet
- every `@vultisig/sdk` import in Windows is peripheral (`ValidationHelpers`, `computeNotificationVaultId`, `SEEDPHRASE_IMPORT_SUPPORTED_CHAINS`, the vite plugin, a side-effect import in `bootstrapMpcEngine.ts:19`)

All THOR/Maya swaps on Windows go through one call site: `core/ui/mpc/keysign/action/mutations/useKeysignMutation.ts:337`.

## Why this matters

The asgard inbound vault rotates every churn. The `vaultAddress` baked into the keysign payload is captured at quote time and never revalidated. Between quote and broadcast sits pairing, QR scan and a full MPC ceremony — unbounded. If the vault rotates in that window, funds go to a retired vault.

## The blocker

`broadcastTx`'s signature is `{ chain, tx }` — it structurally cannot see the keysign payload, so the guard can't be dropped in as-is. This needs either a signature change or a new `packages/core` export that Windows calls explicitly before broadcast.

## Proposal

Move `assertNativeSwapReadyForBroadcast` down from `packages/sdk` into `packages/core/chain/tx/broadcast` (or an adjacent `swap/guard` module in `packages/core`), so that:
1. Windows can call it — it already depends on `@vultisig/core-chain`
2. The SDK facade keeps working by delegating to the moved implementation
3. There is one guard, not two

Decide explicitly whether it runs at **broadcast** time (where it is today) or at **sign** time. Broadcast-time is strictly weaker: the signature already exists, so a rotation detected there means a signed transaction that must be discarded rather than one that was never created. Sign-time is the real fix; iOS's limit-order path already does exactly this (`LimitSwapPayloadAssembler.swift:195-206`, re-resolving the destination from a cache-bypassing fetch).

## Wider context

This is the clearest single example of a general problem: **PRs landing in `packages/sdk` reach the agent app only; PRs landing in `packages/core/*` reach Windows desktop + extension + SDK + agent app.** Windows pins `@vultisig/sdk` in its package.json, which makes the facade look shared when it is not. Worth stating as an architectural rule somewhere visible.

Companion issues: vultisig-windows (call site), vultisig-ios, vultisig-android — all three core apps have the same validation gap in their own stacks.

rcoderdev:
Closes #1392
Moves native-swap expiration and live inbound-vault validation into the public core-mpc keysign surface while keeping the SDK facade delegated to the same guard. Live THOR inbound proof accepted the current BTC vault and rejected a stale vault before signing or broadcast.

## Summary by CodeRabbit

- **New Features**
  - Added/standardized native-swap pre-broadcast validation, including native quote expiry checks and inbound vault verification, with support for optional inbound lookup parameters.
  - Exposed the native-swap broadcast readiness guard (and typed error) via the core MPC package so the SDK can reuse the same checks.

- **Bug Fixes**
  - Improved error reporting for invalid configuration, unsupported chains, network/inbound lookup failures, and vault mismatches using clearer, categorized error codes.

- **Tests**
  - Added coverage ensuring inbound lookup failures surface as `network_error` in the typed guard error.

### DIFF
--- a/.changeset/bright-trackers-link.md
+++ b/.changeset/bright-trackers-link.md
@@ -1,0 +1,8 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Route SwapKit transaction links through its public tracker across the shared core and SDK APIs.
--- a/.changeset/fuzzy-taxis-broadcast.md
+++ b/.changeset/fuzzy-taxis-broadcast.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Align Bittensor broadcast error handling with the shared Substrate safeguards.
--- a/.changeset/honest-vaults-guard.md
+++ b/.changeset/honest-vaults-guard.md
@@ -1,0 +1,7 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Expose the native-swap expiry and inbound-vault guard from core MPC so shipping
+wallets and the SDK facade can share one pre-signing or pre-broadcast check.
--- a/clients/cli/src/agent/__tests__/executor.test.ts
+++ b/clients/cli/src/agent/__tests__/executor.test.ts
@@ -5,7 +5,7 @@ import type { VaultBase } from '@vultisig/sdk'
 import { Chain } from '@vultisig/sdk'
 import { describe, expect, it, vi } from 'vitest'
 
-import { AgentExecutor } from '../executor'
+import { AgentExecutor, resolveChainId } from '../executor'
 
 function createMockVault(): VaultBase {
   const ethereum = Chain.Ethereum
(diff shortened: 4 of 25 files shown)


johnnyluo (approved):
LGTM

PR #1474, opened 2026-07-21
Canonicalize SwapKit tracker URLs

rcoderdev:
## Summary by CodeRabbit

- **New Features**
  - Added public SwapKit tracker links for supported chains.
  - Added CoW Swap Explorer links for pending orders using order UIDs.
  - Exposed the SwapKit tracker URL helper through the core package.

- **Bug Fixes**
  - Unsupported SwapKit chains now fall back to the source-chain explorer.
  - Updated swap tracking links to use the correct hash and chain ID parameters.
  - Clarified supported providers and pending-order link behavior.

### DIFF
--- a/.changeset/bright-trackers-link.md
+++ b/.changeset/bright-trackers-link.md
@@ -1,0 +1,8 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Route SwapKit transaction links through its public tracker across the shared core and SDK APIs.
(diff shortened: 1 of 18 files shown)


johnnyluo (approved):
LGTM

gomesalexandre (approved):
conceptual pass looks good. i rechecked the shared tracker helper and both call sites after the remote review pass:

- `getSwapKitTrackerUrl.ts` centralizes the 25-chain SwapKit mapping and preserves the raw `hash` plus `chainId` shape
- `getSwapExplorerUrl()` now routes `swapkit` through the shared tracker and keeps block-explorer fallback for unmapped chains
- `getSwapTrackingUrl()` now reuses the same helper, so sdk/core/mpc stop drifting on tracker URL shape
- CoW order UIDs are routed through the dedicated explorer path instead of pretending they are tx hashes

non-blocking follow-ups before land:
- branch is still behind `main`, so it needs the normal update-and-rerun before merge
- this is a user-visible sdk link-shape change, so the repo-level runtime receipt is still worth attaching when the branch is refreshed

PR #1477, opened 2026-07-21
fix(sdk): export dangerous-address canonicals

### LINKED ISSUE #1476: public API: export dangerous-address canonicals from the root + RN SDK entries
## Summary
`packages/sdk/src/utils/dangerousAddresses.ts` already centralizes the canonical dangerous/burn-address helpers, but neither the root `@vultisig/sdk` entrypoint nor the React Native entrypoint exports them publicly. That forces downstream consumers to deep-import internals or keep mirroring the canonicals locally.

## Evidence
- `packages/sdk/src/utils/dangerousAddresses.ts:27-90` already exports the canonical helpers/constants:
  - `EVM_DANGEROUS_ADDRESSES`
  - `SOLANA_DANGEROUS_ADDRESSES`
  - `getEvmDangerousReason`
  - `isEvmBurnAddress`
  - `getChainDangerousReason`
  - `assertSafeEvmDestination`
  - `assertSafeDestination`
- `packages/sdk/src/index.ts:85-129` exposes adjacent pure address/contract canonicals but omits the dangerous-address surface.
- `packages/sdk/src/platforms/react-native/index.ts` likewise omits these helpers from the RN public surface.

## Why this is a bug / architecture drift
The SDK already has the canonical implementation, but first-party consumers still cannot import it from the public package surfaces they actually use. That pushes app/backend code back toward local copies — exactly the duplicated-not-imported drift class this campaign keeps finding.

## Suggested fix
Export the dangerous-address canonicals/helpers from both:
- root `@vultisig/sdk`
- `@vultisig/sdk/react-native`

Add regression tests that assert both public surfaces re-export the exact source symbols.

## CLI-testability
Yes. This is a strong low-risk SDK ship candidate:
- root public-export test
- RN entry regression test
- `yarn build:sdk`
- `yarn cli:build`
- built-dist runtime proof from `dist/index.node.esm.js`
- RN export proof via `dist/index.react-native.d.ts` / tarball surface inspection

## Impact
- Lets app/backend consume the SDK canonicals instead of mirroring them
- Reduces future burn-address drift across repos
- Pure public-surface change; no signing/broadcast behavior change by itself

## Effort
Low.

gomesalexandre:
## Summary
- export the canonical dangerous-address helpers/constants from root `@vultisig/sdk`
- export the same helpers/constants from `@vultisig/sdk/react-native`
- add root + RN regression coverage so future public-surface omissions fail loudly

## Why
Round 47 `/improve` found that `packages/sdk/src/utils/dangerousAddresses.ts` already owns the canonical burn/dangerous-address helpers, but first-party consumers still cannot import them from the public SDK surfaces they actually use. That keeps app/backend code on the duplicated-not-imported path.

## Test plan
- `corepack yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/utils/publicExports.test.ts packages/sdk/tests/unit/platforms/react-native/entry.test.ts`
- `corepack yarn build:sdk`
- `corepack yarn cli:build`
- `node --input-type=module -e "import * as sdk from './packages/sdk/dist/index.node.esm.js'; console.log(JSON.stringify({hasEvmDangerous:sdk.EVM_DANGEROUS_ADDRESSES['0x000000000000000000000000000000000000dead'], hasFn: typeof sdk.assertSafeDestination, burnCheck: sdk.isEvmBurnAddress('0x000000000000000000000000000000000000dEaD')}))"`
- inspect `packages/sdk/dist/index.react-native.d.ts` for the RN export surface
- `corepack yarn workspace @vultisig/sdk pack --out /tmp/vultisig-sdk-r47/.artifacts/vultisig-sdk-r47.tgz`

Closes #1476

### DIFF
--- a/.changeset/sdk-dangerous-address-exports.md
+++ b/.changeset/sdk-dangerous-address-exports.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Export the canonical dangerous-address helpers and constants from both the root SDK entrypoint and the React Native entrypoint.
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -107,6 +107,15 @@ export {
 export type { AddressFamily, AddressRole, ChainPrefixResult } from './utils/addressValidation'
 export { address, validate } from './utils/addressValidation'
 export { checkChainPrefix } from './utils/chainPrefix'
+export {
+  assertSafeDestination,
+  assertSafeEvmDestination,
+  EVM_DANGEROUS_ADDRESSES,
+  getChainDangerousReason,
+  getEvmDangerousReason,
+  isEvmBurnAddress,
+  SOLANA_DANGEROUS_ADDRESSES,
+} from './utils/dangerousAddresses'
 
 // ============================================================================
 // PUBLIC API - Tx Shape Normalization (pure, vault-free)
--- a/packages/sdk/src/platforms/react-native/index.ts
+++ b/packages/sdk/src/platforms/react-native/index.ts
@@ -613,6 +613,15 @@ export { chainSchema, parseChain, parseTicker, tickerSchema } from '../../tools/
 export type { NormalizeArgs, NormalizedTx } from '../../tx'
 export { normalizeTx, splitMultiTx, TxNormalizeError } from '../../tx'
 export { computePersonalSignHash, formatEcdsaSignature65 } from '../../utils/eip191'
+export {
+  assertSafeDestination,
+  assertSafeEvmDestination,
+  EVM_DANGEROUS_ADDRESSES,
+  getChainDangerousReason,
+  getEvmDangerousReason,
+  isEvmBurnAddress,
+  SOLANA_DANGEROUS_ADDRESSES,
+} from '../../utils/dangerousAddresses'
 export { normalizeChain, UnknownChainError } from '../../utils/normalizeChain'
 export { resolveChainReference } from '../../utils/resolveChainReference'
 export { ChainAmountParseError, toChainAmount } from '@vultisig/core-chain/amount/toChainAmount'
--- a/packages/sdk/tests/unit/platforms/react-native/entry.test.ts
+++ b/packages/sdk/tests/unit/platforms/react-native/entry.test.ts
@@ -152,3 +152,18 @@ describe('RN entry exposes toChainAmount + ChainAmountParseError', () => {
     expect(() => rn.toChainAmount('   ', 8)).toThrow(rn.ChainAmountParseError)
   })
 })
+
+describe('RN entry exposes the dangerous-address helpers', () => {
+  it('re-exports the canonical dangerous-address helpers and constants from the RN entrypoint', async () => {
+    const rn = await import('../../../../src/platforms/react-native/index')
+    const dangerousAddresses = await import('../../../../src/utils/dangerousAddresses')
+
+    expect(rn.EVM_DANGEROUS_ADDRESSES).toBe(dangerousAddresses.EVM_DANGEROUS_ADDRESSES)
+    expect(rn.SOLANA_DANGEROUS_ADDRESSES).toBe(dangerousAddresses.SOLANA_DANGEROUS_ADDRESSES)
+    expect(rn.getEvmDangerousReason).toBe(dangerousAddresses.getEvmDangerousReason)
+    expect(rn.isEvmBurnAddress).toBe(dangerousAddresses.isEvmBurnAddress)
+    expect(rn.getChainDangerousReason).toBe(dangerousAddresses.getChainDangerousReason)
+    expect(rn.assertSafeEvmDestination).toBe(dangerousAddresses.assertSafeEvmDestination)
+    expect(rn.assertSafeDestination).toBe(dangerousAddresses.assertSafeDestination)
+  })
+})
--- a/packages/sdk/tests/unit/utils/publicExports.test.ts
+++ b/packages/sdk/tests/unit/utils/publicExports.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from 'vitest'
 
 import * as sdk from '../../../src/index'
+import * as dangerousAddresses from '../../../src/utils/dangerousAddresses'
 
 describe('@vultisig/sdk public exports', () => {
   it('exports fiatToAmount, toChainAmount, and chain-reference normalization utilities', () => {
@@ -43,6 +44,16 @@ describe('@vultisig/sdk public exports', () => {
     expect(typeof sdk.isCanonicalTronContract).toBe('function')
     expect(sdk.canonicalEvmContracts instanceof Set).toBe(true)
     expect(typeof sdk.knownContracts.isKnownContract).toBe('function')
+  })
+
+  it('exports the canonical dangerous-address helpers and constants', () => {
+    expect(sdk.EVM_DANGEROUS_ADDRESSES).toBe(dangerousAddresses.EVM_DANGEROUS_ADDRESSES)
+    expect(sdk.SOLANA_DANGEROUS_ADDRESSES).toBe(dangerousAddresses.SOLANA_DANGEROUS_ADDRESSES)
+    expect(sdk.getEvmDangerousReason).toBe(dangerousAddresses.getEvmDangerousReason)
+    expect(sdk.isEvmBurnAddress).toBe(dangerousAddresses.isEvmBurnAddress)
+    expect(sdk.getChainDangerousReason).toBe(dangerousAddresses.getChainDangerousReason)
+    expect(sdk.assertSafeEvmDestination).toBe(dangerousAddresses.assertSafeEvmDestination)
+    expect(sdk.assertSafeDestination).toBe(dangerousAddresses.assertSafeDestination)
   })
 
   it('exports findSwapQuote, abiEncode, evmCheckAllowance (already consumed by mcp-ts)', () => {


NeOMakinG (commented):
Reviewed at HEAD `278c1e0d`. Draft, so **COMMENTED** — no approval.

The diff itself is clean: 7 names, added to both entrypoints, no collisions (each name appears exactly once in `packages/sdk/src/index.ts` and once in `packages/sdk/src/platforms/react-native/index.ts`), purely additive, nothing already-exported changes shape. The two new tests genuinely bite this branch.

My concern is not the diff — it's what the diff *promotes*. This PR takes `dangerousAddresses.ts` from module-private to public, in a changeset that calls it "the canonical dangerous-address helpers", so first-party consumers can "stop being on the duplicated-not-imported path". I probed that promise, and today the module is **not** canonical: it is a strict subset of both the mcp-ts file its own header names as its parity source, and the agent-backend-ts copy it is meant to replace. A consumer that does what the PR description invites — delete its local list and import from `@vultisig/sdk` — loses burn-address coverage.

## testing approach

Worktree at `278c1e0d` (`/tmp/wt-vultisig-sdk-1477`), shared-package `dist/` symlinked from the canonical checkout, vitest against `packages/sdk/tests/unit/vitest.config.ts`.

- Both new tests **pass** at HEAD (29/29 across the two files).
- **M1 (control mutation, confirmed present in the file before running):** removed `isEvmBurnAddress` from the root barrel and `SOLANA_DANGEROUS_ADDRESSES` from the RN barrel → **2 failed / 27 passed**. RED. The tests do read this branch, not `main`.
- **M2:** deleted the `1nc1nerator…` entry from `SOLANA_DANGEROUS_ADDRESSES` (grep-verified absent) and ran `dangerousAddresses.test.ts` + `bridge-cctp.test.ts` + both new files → **54 passed, all green.** See #4.
- Runtime probe of the built public surface's behaviour (`assertSafeDestination`, `getChainDangerousReason`, raw-record lookup, `Object.isFrozen`) — results inline below.
- Codex second-opinion lane: **ran, alive, 3 findings** (mutability, raw-record footgun, source-vs-artifact tests) — independently matching #2, #3 and #5 below. It did not find #1.
- Not verified: the built `dist/index.react-native.js` / `.d.ts` artifacts (no build run in this pass — the PR description reports doing it manually; there is no CI test that pins it, see #5).

**Nothing consumes these exports yet** — in-repo, `assertSafeEvmDestination`/`assertSafeDestination` are imported directly from `../../utils/dangerousAddresses` by `buildCctpBridge.ts` and `acrossQuote.ts`, never from the barrel; vultiagent-app and agent-backend-ts still use their own copies. So this is **inert-by-design, an export ahead of its consumer**, not a shipped protection. That is fine — but it means the migration this unblocks is the thing to get right, and that migration is currently a downgrade.

## blocking

**1. The list this PR labels "canonical" blocks 5 of the 11 addresses its own parity sources block — migrating a consumer onto it is a net loss of coverage.**

`packages/sdk/src/utils/dangerousAddresses.ts:1-11` says it is a "Port of mcp-ts `src/lib/dangerous-addresses.ts` (itself the parity source for the Go MCP guard)". It ports the EVM table and half the Solana table, and drops the UTXO and XRP tables entirely. Probed at HEAD through the newly-exported `sdk.assertSafeDestination` — every one of these returns cleanly:

```
chain=Bitcoin addr=1BitcoinEaterAddressDontSendf59kuE          threw=false
chain=Bitcoin addr=1111111111111111111114oLvT2                 threw=false
chain=Ripple  addr=rrrrrrrrrrrrrrrrrrrrrhoLvTp                 threw=false   (XRPL ACCOUNT_ZERO)
chain=Ripple  addr=rrrrrrrrrrrrrrrrrrrrBZbvji                  threw=false   (XRPL ACCOUNT_ONE)
chain=Solana  addr=TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA threw=false   (SPL Token Program)
chain=Solana  addr=So11111111111111111111111111111111111111112 threw=false   (Wrapped SOL mint)
```

All six are in `mcp-ts/src/lib/dangerous-addresses.ts` (`UTXO_DANGEROUS`, `XRP_DANGEROUS`, `SOLANA_DANGEROUS`) **and** in `agent-backend-ts/src/mastra/nullRecipientWarning.ts:52-62` (`BASE58_BURNS`). The abts file carries an explicit instruction — *"Duplicated here (rather than imported) … A new burn form added there MUST be mirrored here"* — i.e. it is exactly the consumer this PR wants to un-duplicate. If it swaps its 8-entry `BASE58_BURNS` + `EVM_ZERO_RE` for `@vultisig/sdk`'s two Solana entries, the BTC eater, the BTC null script, both XRPL black-holes, the SPL Token Program and the wSOL mint stop producing the `NULL_RECIPIENT_WARNING`.

The drift is bidirectional, which is the tell that no single file is canonical yet: the SDK has `1nc1nerator…` (mcp-ts does **not** — see "pre-existing" below), mcp-ts has the SPL/wSOL/BTC/XRP entries the SDK does not, and abts has the union of both.

Fix — either of:
- (a) Bring the SDK module up to the union before publishing it as canonical: add `UTXO_DANGEROUS` (BTC eater, null script) keyed on the UTXO chain set, `XRP_DANGEROUS` (ACCOUNT_ZERO/ACCOUNT_ONE) on `Ripple`, and the two missing Solana program ids — i.e. port the rest of the mcp-ts file, not just the EVM half; **or**
- (b) if the scope is deliberately EVM+Solana for now, then don't ship it under the word "canonical". Rename the changeset/doc-comment to say what it actually covers ("EVM burn addresses and two Solana program destinations"), and add an explicit note that non-EVM/non-Solana chains are **unguarded and return without throwing**, so a consumer knows `assertSafeDestination('Bitcoin', …)` is not a safety check. Right now the name and the changeset both assert coverage this module does not have — the classic false-sense-of-safety blocklist.

Either way this is worth settling *before* the export lands, because the export is what makes the gap reachable by first-party code.

## preferably-blocking

**2. The exported records are not frozen — one `delete` disarms the SDK's own internal guards.**

`dangerousAddresses.ts:28` and `:36` are plain mutable `Record<string, string>`. `const` freezes the binding, not the object, and the guards read those same live objects (`:52`, `:61`). Before this PR the objects were module-private; the PR hands every consumer a writable handle to the SDK's blocklist. Probed:

```
frozen? EVM=false SOL=false
after delete EVM_DANGEROUS_ADDRESSES['0x…dead'], assertSafeEvmDestination('0x…dEaD') threw = false
```

That mutation also disarms `buildCctpBridge.ts:259` and `acrossQuote.ts:263` process-wide, from a consumer that never touched those modules. Codex flagged this independently at High.

Fix: `Object.freeze` both tables (and prefer `as const satisfies Record<string, string>` over the widened `Record<string, string>` annotation, which also loses the literal key types):

```ts
export const EVM_DANGEROUS_ADDRESSES: Readonly<Record<string, string>> = Object.freeze({ … })
export const SOLANA_DANGEROUS_ADDRESSES: Readonly<Record<string, string>> = Object.freeze({ … })
```

Cheap, and it removes a new attack surface this PR creates rather than one it inherits.

## should-fix

**3. Exporting the raw records makes the wrong usage the easy one — a checksummed EVM address direct-lookups to `undefined`.**

The module's safety invariant is "go through the helpers": EVM keys are lowercase and `getEvmDangerousReason` normalizes with `.toLowerCase()` (`:52`). Publish the record next to the helper and the one-liner a consumer reaches for silently fails open. Probed:

```
EVM_DANGEROUS_ADDRESSES['0x000000000000000000000000000000000000dEaD']  -> undefined
isEvmBurnAddress('0x000000000000000000000000000000000000dEaD')         -> true
```

`0x…dEaD` is the *EIP-55 checksummed* form — i.e. exactly what ethers/viem/`getAddress()` hand a consumer, and exactly what vultiagent-app's own fixtures use (`src/services/__tests__/evmTx.test.ts` uses `0x…dEaD` throughout). Solana is the mirror-image trap: base58 is case-sensitive and `getChainDangerousReason` only `.trim()`s, so a consumer who lowercases "to canonicalize" false-negatives there instead. This is the same case-sensitivity class flagged on vultiagent-app#2181 today.

Fix: don't export the raw maps at all — the helpers cover every legitimate read, and `getEvmDangerousReason`/`getChainDangerousReason` already return the reason string, so nothing needs the record. If they're exported for enumeration/UI, export a function (`listDangerousAddresses()`) returning a frozen copy, and put "keys are lowercase — do not index this directly, use `getEvmDangerousReason`" in the doc-comment on the export line, not only at `:27`.

**4. `SOLANA_DANGEROUS_ADDRESSES` has no membership test — deleting an entry from the newly-public constant is invisible.**

`dangerousAddresses.test.ts:36` does `it.each(Object.keys(SOLANA_DANGEROUS_ADDRESSES))`, iterating the table under test: a deleted row is *unvisited*, not unmatched. M2 above confirmed it — I removed the Solana incinerator entry and all 54 tests stayed green. Contrast `bridge-cctp.test.ts:267-273`, which does this correctly for EVM (`toContain` × 3 plus `keys.length).toBe(3)`).

Fix: mirror the EVM pattern for Solana (and for any table added per #1):

```ts
it('lists the canonical Solana dangerous destinations', () => {
  const keys = Object.keys(SOLANA_DANGEROUS_ADDRESSES)
  expect(keys).toContain('11111111111111111111111111111111')
  expect(keys).toContain('1nc1nerator11111111111111111111111111111111')
  expect(keys.length).toBe(2)
})
```

**5. The new tests pin `src/`, not the surface a consumer resolves.**

Both tests import `../../../src/index` and `../../../../src/platforms/react-native/index`. The thing a consumer actually resolves is the conditional export map in `packages/sdk/package.json` → `dist/index.node.esm.js` / `dist/index.react-native.js` and the matching `.d.ts`. A rollup/config regression that drops these from the RN bundle, or a `.d.ts` that doesn't emit them, leaves both tests green — which is the exact failure mode this PR exists to prevent ("future public-surface omissions fail loudly"). The PR description does probe `dist/index.node.esm.js` by hand, but nothing in CI does. Codex flagged the same gap.

Fix: add one artifact-level assertion to the existing dist smoke test (or a new one gated on a prior `build:sdk`) that imports the built `react-native` entry and asserts `typeof assertSafeDestination === 'function'`, so the packaging path is pinned and not just the source barrel.

## q

**6. What chain vocabulary is `getChainDangerousReason` public API for?**

`:60` matches the single literal `chain.trim().toLowerCase() === 'solana'`. Probed:

```
chain=Solana                       -> blocked
chain=SOL                          -> UNDEFINED (fails open)
chain=sol                          -> UNDEFINED (fails open)
chain=solana-mainnet               -> UNDEFINED (fails open)
chain=solana:5eykt4UsFv8P8NJd…     -> UNDEFINED (fails open)
```

Internally the only caller is `acrossQuote.ts:263`, which passes an SDK chain name, so this is fine today. As public API with a bare `chain: string` parameter it is not — a consumer holding a ticker, a CAIP-2 id, or a slug gets a silent all-clear with no type error. Worth either typing the parameter as the SDK `Chain` enum, or routing it through `normalizeChain`/`resolveChainReference` (both already exported from these same barrels) so the accepted vocabulary is the SDK's own. Not blocking while nothing consumes it — but it becomes load-bearing the moment #1's migration happens.

**7.** The changeset bumps `@vultisig/cli` as a patch but nothing under `clients/cli` changed. Intentional (forcing a dependent re-release), or leftover?

- `mcp-ts/src/lib/dangerous-addresses.ts` — which `agent-backend-ts/src/mastra/nullRecipientWarning.ts:7-8` calls "the PRIMARY enforcement … every `build_*` tool throws before building a burn tx" — has **no entry for the Solana incinerator** `1nc1nerator11111111111111111111111111111111`. The SDK has it, abts has it, mcp-ts does not. So `build_*` on mcp-ts will happily build a send to the Solana incinerator. Separate PR, but it's the same drift this one is trying to end, pointing the other way.

## risk

Low as-merged: additive exports, nothing consumes them, no runtime behaviour changes for any existing caller, no dependency movement (no caret-bump hazard here). The risk is entirely deferred and lands on whoever does the migration this unblocks — if a consumer swaps its local list for this one on the strength of the word "canonical", coverage silently shrinks on Bitcoin, Ripple, and two Solana program destinations, with the guard still returning "safe".

## verdict

**COMMENTED** (draft — never approved). No objection to the export mechanism, which is correct and well-tested. #1 should be settled before the export lands, since the export is what makes the gap reachable; #2 is a two-line fix worth taking in this PR. Happy to re-review once the list is either brought to parity or the scope is stated honestly.

<sub>🤖 automated review — vultisig-ops reviewer</sub>

PR #1480, opened 2026-07-21
fix(sdk): publish policy bridge gas tool subpaths

### LINKED ISSUE #1479: public API: publish tools/policy, tools/bridge, and tools/gas as real SDK subpaths
Summary
- `packages/sdk/src/tools/policy`, `tools/bridge`, and `tools/gas` are already coherent pure SDK surfaces, but `packages/sdk/package.json` only publishes `./tools/parse` and `./tools/defi`.

Evidence
- `packages/sdk/src/tools/policy/index.ts` exposes the pure `policy` namespace plus related helpers/types.
- `packages/sdk/src/tools/bridge/index.ts` exposes CCTP bridge/claim builders and registry helpers.
- `packages/sdk/src/tools/gas/index.ts` exposes fee/gas helpers.
- `packages/sdk/package.json:76-101` publishes `./tools/parse` and `./tools/defi`, but not these sibling tool families.

Impact
- Downstreams cannot import the narrow canonical surfaces directly as package subpaths.
- This pushes consumers toward broader root imports or internal-path reach-through, which is exactly the public-surface drift pattern earlier rounds kept finding.

Why this should live in `vultisig-sdk`
- These are already SDK-owned, pure, testable helper families.
- Publishing them as real subpaths makes them directly CLI/tarball-smoke-testable and easier for app/backend/other consumers to adopt.

Fix sketch
1. Add export-map entries in `packages/sdk/package.json`.
2. Add JS + d.ts bundle wiring in the Rollup configs.
3. Extend the public-api regression test and tarball smoke script.

Effort
- Small.

Round report
- https://gist.github.com/gomesalexandre/856a6825b34b195e44c5f842521c5411

gomesalexandre:
## Summary
- publish dedicated `@vultisig/sdk/tools/policy`, `@vultisig/sdk/tools/bridge`, and `@vultisig/sdk/tools/gas` subpath exports
- wire matching JS + d.ts bundle generation into the SDK Rollup configs
- extend the public-api regression test and tarball smoke script to prove the new package-name imports resolve

## Why
- closes https://github.com/vultisig/vultisig-sdk/issues/1479
- these tool families were already coherent pure SDK surfaces, but only `tools/parse` and `tools/defi` were published as real subpaths
- publishing narrow subpaths keeps the SDK boundary consumer-friendly and reduces future reach-through / reimplementation pressure in app/backend consumers

## Verification
- `yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/public-api/toolsSubpathExports.test.ts`
- `yarn build:sdk`
- `yarn cli:build`
- `yarn sdk:smoke:subpaths`
- `yarn lint`
- `yarn build:all`
- `yarn test`

## Receipts
- report: https://gist.github.com/gomesalexandre/856a6825b34b195e44c5f842521c5411
- targeted public-api test passed locally
- `sdk:smoke:subpaths` packed the tarball and resolved/imported all five tool subpaths (`parse`, `defi`, `policy`, `bridge`, `gas`) from the packaged SDK


NeOMakinG (commented):
Round 1 review at HEAD `3631cfe3b34eb382bd6493c3daa8d19ad29ed7d2` (draft — posting as **COMMENT**, no approval). PR is 0 commits behind `main`.

The export plumbing itself is correct and genuinely purely additive. My findings are all about **what the three newly-published modules actually promise a consumer**, which is the real payload of a subpath-publish PR — the same lens that turned up the coverage gap on #1477.

## testing approach

- Worktree `/tmp/wt-vultisig-sdk-1480` at the PR sha; `node_modules` + per-package `dist` symlinked from the canonical checkout.
- **Control mutation (mandatory, SDK worktree aliasing trap):** flipped `GAS_UNITS.transfer` `21_000 → 22_000` in the worktree copy, grep-verified the mutation was physically present in the worktree file *and* absent from the canonical checkout, then ran the suite → **2 failed / 4 passed (RED)**. So the worktree source is genuinely under test, not `main`. Restored and re-verified `21_000`.
- Baseline: `tests/unit/tools/gas/compareCosts.test.ts` + `tests/unit/public-api/toolsSubpathExports.test.ts` → 8/8 green.
- Coverage diff of all three modules against the first-party copies in `mcp-ts` / `agent-backend-ts` / the Go `agent-backend`.
- Live probe: `eth_getCode` against `https://mainnet.base.org` for both CCTP `messageTransmitter` candidates.
- Codex lane: **ALIVE** (1 finding, verified against this repo, folded in below and escalated).

## coverage verdict — is the SDK equal-or-better than what it invites consumers to delete?

| subpath | first-party counterpart | verdict |
|---|---|---|
| `tools/bridge` | `mcp-ts/src/lib/cctp.ts`, `agent-backend-ts/.../lib/cctp.ts` | **parity, and strictly better than mcp-ts** (see pre-existing bug below). Same 6 CCTP domains (0/1/2/3/6/7), addresses byte-identical to abts. |
| `tools/policy` | Go `agent-backend/internal/safety/policy.go` | **parity** — full `I1`..`I7`. No TS first-party copy exists in mcp-ts/abts, so nothing to under-cover. |
| `tools/gas` | `mcp-ts/src/tools/evm/compare-gas-costs.ts` | **under-covers / regresses** — see the blocking-ish finding below. |

## preferably-blocking

**1. `compareCosts` emits a confidently wrong `cheapest` on its default path, and that is a regression vs the mcp-ts tool it invites consumers to replace.**

`packages/sdk/src/tools/gas/compareCosts.ts` ranks by `estTxCostNative` whenever the price map is not fully populated — which is the **zero-argument default**, since `nativeUsdPrices` is optional and there is no fetch. `DEFAULT_COMPARE_CHAINS` spans ETH / POL / BNB / AVAX, so the default call compares quantities denominated in four different tokens.

Reproduced at runtime (mocked RPC, real module):

```
compareCosts({ chains: ['Ethereum', 'Polygon'] })   // no prices — the default path
  Ethereum @ 10 gwei -> estTxCostNative 0.00021 ETH
  Polygon  @ 30 gwei -> estTxCostNative 0.00063 POL
  cheapest: { chain: "Ethereum", estTxCostNative: 0.00021, estTxCostUsd: null }
```

At ETH ≈ \$3000 / POL ≈ \$0.20 that is **\$0.63 crowned as cheaper than \$0.000126** — off by ~5000x, in the one field the tool exists to produce.

The code comment above the sort already articulates exactly why this is invalid ("1e-4 ETH ≠ 1e-4 MATIC … non-transitive"), then does it anyway and calls it "the documented gwei-only best-effort" — but it still populates `cheapest`, so the caller has no signal that the answer is not comparable.

Critically, `mcp-ts` `compare_gas_costs` — the thing a consumer would delete — **never does this**: `getChainGasData` fetches the CoinGecko price itself, and any chain with `native_usd === null` is pushed to `skipped[]` with `'Native token USD price unavailable'`. It only ever ranks on a single USD basis. So migrating to the SDK primitive is a net loss of correctness.

Suggested fix (any one is sufficient):
- when `!allPriced`, return `cheapest: null`; or
- add `basis: 'usd' | 'native'` to `CompareCostsResult` and set `cheapest` only for `'usd'`; or
- keep native ranking but only when every surviving chain shares a native token.

**2. The test that appears to guard this is constructed so it cannot fail.**

`tests/unit/tools/gas/compareCosts.test.ts:63-87` — "does not mix USD/native bases when the price map is partial (no false cheapest)" — its comment warns a mixed comparator "could crown the \$1.89 Ethereum tx as cheapest", but the fixture puts Polygon at **10 gwei vs Ethereum's 30**, so native ranking coincides with the correct USD ranking. The wrong answer is never reachable.

`:16-38` ("ranks chains cheapest-first by native tx cost") uses Ethereum / Base / Arbitrum — **all three have ETH as the native token**, so native ranking is legitimate there.

Net: **no test ever native-ranks across two different native tokens.** That is precisely why finding 1 is green. Adding the Ethereum-vs-Polygon-unpriced case above turns it red.

## should-fix

**3. Mixing the root entry and a tools subpath silently drops a configured custom RPC.** (Codex lane, verified and escalated.)

`src/tools/index.ts:49,65,200,318` already re-export `compareCosts` / `GAS_UNITS` / `utxoFeeRate` / `buildCctpBridge` / `policy` through the root barrel, and `createToolsSubpathConfigs` builds each subpath with `inlineDynamicImports: true` (`rollup.platforms.config.js:246,266`) — fully self-contained bundles. So a consumer doing:

```ts
import { setCustomRpcOverride } from '@vultisig/sdk'
import { compareCosts }         from '@vultisig/sdk/tools/gas'
```

gets **two copies of the module-level `overridesByChain` Map** (`packages/core/chain/chains/customRpc/customRpcOverrides.ts:29-51`). The subpath copy is empty, so `getEvmRpcUrl` falls through to the default public endpoint and the gas read silently ignores the override the app configured. The memoized `getEvmClient` cache duplicates for the same reason.

`policy` and `bridge` are effectively pure, so this is gas-specific today. At minimum this deserves a note in the subpath README/JSDoc ("do not mix root and subpath imports for stateful surfaces"); better, keep `core-chain` external for the subpath bundles so the state stays shared.

## q

**4. Is the OP-stack L1 data fee meant to be out of scope for `compareCosts`?**

`getChainGasPriceGwei` reads `eth_gasPrice` only, and the cost model is `gasPrice × gasUnits` — pure L2 execution. `DEFAULT_COMPARE_CHAINS` includes **Optimism and Base**, where the L1 posting/data fee routinely exceeds L2 execution cost by an order of magnitude or more. That systematically biases the ranking toward OP-stack chains in exactly the comparison the tool is for.

This is pre-existing and shared with mcp-ts, so I am not treating it as introduced here — but promoting the estimator to a public subpath is the moment to either read the `GasPriceOracle` (`0x420…000F`) or document the omission in `CompareCostsResult`. **Gap I did not close:** I did not live-quantify the current L1/L2 ratio on Base.

**mcp-ts has a dead CCTP `messageTransmitter` for Base.** Diffing the three copies surfaced exactly one divergence, and a live `eth_getCode` on Base settles it:

| copy | Base `messageTransmitter` | `eth_getCode` on Base |
|---|---|---|
| this SDK module | `0xAD09780d193884d503182aD4588450C416D6F9D4` | contract bytecode ✅ |
| `agent-backend-ts` | `0xAD09780d193884d503182aD4588450C416D6F9D4` | contract bytecode ✅ |
| `mcp-ts/src/lib/cctp.ts` | `0xAD09780d193884d503182aD4F75d113B9b9a86e7` | **`0x` — no code** ❌ |

So mcp-ts's Base CCTP claim path targets a non-contract. Not this PR's problem — and it is a point *in favour* of publishing the SDK copy — but it warrants its own mcp-ts issue.

## risk

**Low for the export plumbing, medium for what it endorses.**

Verified purely additive: exactly 3 new keys (`./tools/policy`, `./tools/bridge`, `./tools/gas`), each appearing once, no existing export entry changed shape (parsed the export map at the PR sha and diffed against `main`). `files: ["dist", …]` does pack `dist/tools/*`. Condition ordering matches the existing `tools/parse` / `tools/defi` entries. **No dependency added or moved**, so the caret-float hazard does not apply here.

**Inert-by-design, correctly.** Nothing consumes the package-name subpaths yet — the only references are in `scripts/smoke-sdk-subpath-exports.mjs`; every in-repo consumer still imports via `@/tools/…` or relative paths. This is an export ahead of its consumer, not shipped capability. That is also why findings 1 and 3 are cheap to fix *now*, before a consumer migrates onto them.

Minor note: the smoke script asserts `typeof … === 'function'` for the three new symbols — a resolution smoke, not a behaviour test. Appropriate for its purpose, just not coverage.

## verdict

**COMMENTED** — draft, so no approval by policy, and finding 1 would hold a non-draft at `preferably-blocking` anyway.

`tools/policy` and `tools/bridge` are clean to publish: both are at parity or better versus their first-party counterparts. `tools/gas` should not be promoted to public API until `compareCosts` stops reporting a `cheapest` computed on a non-comparable basis — publishing it as-is is an invitation to replace a *more* correct mcp-ts implementation with a less correct one.

<sub>🤖 automated review — Claude lane (load-bearing) + Codex lane (alive, 1 finding, folded in). Runtime evidence from a control-mutated worktree at the PR sha.</sub>

PR #1486, opened 2026-07-22
Harden React Native crypto forward compatibility

### LINKED ISSUE #1251: chore: rn platform forward-compat hardening (noble/curves v1/v2 api skew + solana hex charset validation)
Round-20 audit of the react-native platform tree - two forward-compat / defense-in-depth hardening items. Neither is a live bug today; both are landmines.

## 1. @noble/curves API-version skew across sibling RN files (RN20-04)

- `shims/tiny-secp256k1.ts:6,8,31` uses the v1 API: `secp256k1.ProjectivePoint`, `.toRawBytes()`, import `'@noble/curves/secp256k1'` (no `.js`).
- `chains/cosmos/tx.ts:20,78,82` + `chains/ripple/tx.ts:28,99,103` use the v2 API: `secp256k1.Point`, `.toBytes()`, import `'@noble/curves/secp256k1.js'`.

The hoisted copy is currently 1.9.7 (where ProjectivePoint aliases Point), so no live break. But @noble/curves 2.2.0 is ALREADY present transitively (cosmjs, @mysten/sui, ripple-keypairs) and REMOVES ProjectivePoint/toRawBytes. A future hoist/bump makes `secp256k1.ProjectivePoint` undefined -> the shim's GE.fromHex/GE.BASE.multiply throw -> swallowed by each method's try/catch -> every function returns null/false -> bip32 BIP32Factory fails child-key derivation SILENTLY (while cosmos/ripple on the v2 API keep working, so the shim won't be the obvious culprit). This is the same "one dep bump silently breaks derivation" class the app's react-native pinning rules exist to prevent, but inside the sdk.

Fix: unify on the v2 API (`const GE = secp256k1.Point; pt.toBytes(compressed)`) + the `.js` import specifier. Relates to the noble/hashes v1/v2 split already filed (#1232) - same family, worth doing together.

## 2. Solana hexToBytes skips charset validation (RN20-05)

`chains/solana/tx.ts:138-148` validates only length parity, unlike cosmos/ripple/sui (:34-47 / :46-59 / :19-32) which all check `/^[0-9a-fA-F]+$/`. A non-hex nibble in Solana becomes parseInt(...,16)=NaN -> coerced to 0 in the Uint8Array; the `sigBytes.length !== 64` guard (:245) still passes -> a silently-zeroed signature reaches finalize. Not live (fastVaultSign pre-validates r/s/rid as hex at :204,220), but any future path reaching Solana finalize with non-canonical hex fails LOUD on other chains and SILENT on Solana. Add the same charset regex to Solana's (and EVM's) hexToBytes.

effort: S each / risk: LOW / confidence: MED on #1 (conditional on a bump), HIGH on #2 (drift verified)

found by improve round 20 (2026-07-13)

rcoderdev:
Closes #1251

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened React Native signature finalization by rejecting malformed/non-hex Solana and EVM signatures.
  * Added stricter EVM recovery-id validation (must be 0 or 1) during transaction finalization.
  * Strengthened secp256k1 point/tweak/scalar handling in the React Native shim, with improved range checks and safer key derivation.
  * Added optional `extraData` support for deterministic/hedged signing.
* **Tests**
  * Extended React Native unit and golden-vector tests to cover key/point operations and invalid signature inputs during transaction finalization.

### DIFF
--- a/.changeset/rn-forward-compat-hardening.md
+++ b/.changeset/rn-forward-compat-hardening.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Use the noble Point API in React Native derivation and reject malformed Solana or EVM signature hex.
--- a/packages/sdk/src/platforms/react-native/chains/evm/tx.ts
+++ b/packages/sdk/src/platforms/react-native/chains/evm/tx.ts
@@ -222,9 +222,15 @@ export function buildEvmContractCallTx(opts: BuildEvmContractCallOptions): EvmTx
     if (clean.length !== 130) {
       throw new Error(`expected 65-byte signature (130 hex chars, r||s||recoveryId), got ${clean.length}`)
     }
+    if (!/^[0-9a-fA-F]+$/.test(clean)) {
+      throw new Error('invalid signature: non-hex characters in input')
+    }
     const r = `0x${clean.substring(0, 64)}` as `0x${string}`
     const s = `0x${clean.substring(64, 128)}` as `0x${string}`
     const recoveryId = parseInt(clean.substring(128, 130), 16)
+    if (recoveryId !== 0 && recoveryId !== 1) {
+      throw new Error(`invalid signature: recoveryId must be 0 or 1, got ${recoveryId}`)
+    }
 
     const rawTxHex = isLegacy
       ? serializeTransaction(
--- a/packages/sdk/src/platforms/react-native/chains/solana/tx.ts
+++ b/packages/sdk/src/platforms/react-native/chains/solana/tx.ts
@@ -140,6 +140,9 @@ function hexToBytes(hex: string): Uint8Array {
   if (clean.length % 2 !== 0) {
     throw new Error(`invalid hex length: ${clean.length}`)
   }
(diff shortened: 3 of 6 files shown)


gomesalexandre (approved):
LGTM.

Focused validation:
- read the exact diff at origin/main...origin/pr/1486
- focused RN receipts on the PR snapshot before the host hit disk pressure: vitest on tiny-secp256k1 + tx-builder-golden-vectors passed 2 files / 10 tests
- targeted TS parse/type sanity on packages/sdk/src/platforms/react-native/shims/tiny-secp256k1.ts

Why this looks good:
- explicit .js noble import is the right Metro/Hermes forward-compatible shape
- the shim now matches the current Point/toBytes/fromBytes API without widening behavior
- malformed sig hex and bad recovery ids now fail closed
- the new tests pin both signature behavior and RN-vs-WalletCore/web3 golden vectors

Host note: a later rerun started failing with ENOSPC while Vitest tried to create temp dirs, so extra local validation is blocked by disk pressure rather than the patch itself.

PR #1489, opened 2026-07-22
Decode canonical CW20 transfer messages

### LINKED ISSUE #1255: decode: CW20 transfer bytes still collapse to generic contractCall in decodeCosmosTx
## Summary
`decodeCosmosTx()` still degrades CW20 transfer bytes to a generic `contractCall` even though the SDK already owns the inverse CW20 transfer builder.

## Evidence
- `packages/sdk/src/tools/decode/cosmos.ts:101-113` — `MsgExecuteContract` returns only the contract + attached funds, without decoding the JSON execute body.
- `packages/sdk/src/tools/prep/cw20Transfer.ts:178-187` — the SDK already builds the canonical `{ "transfer": { "recipient", "amount" } }` CW20 execute body.

## Impact
A tx built by the SDK's own CW20 prep helper cannot be fully decoded back into recipient+amount by the SDK's own bytes oracle. Downstreams that want parity must re-parse CosmWasm JSON themselves.

## Why this belongs in the SDK
The canonical decoder should round-trip the SDK's own builder so app/backend/CLI all consume the same result.

## Fix sketch
Teach `decodeCosmosTx()` to parse common CW20 transfer execute JSON and lift recipient/amount into the Envelope, with regression tests from bytes built via the existing helper.

## Effort
Small-to-medium.

Report mirror: https://gist.github.com/gomesalexandre/b1deffff1532dd07f2f5e7dae2b9f0a8

rcoderdev:
Closes #1255

## Summary by CodeRabbit

- **New Features**
  - CW20 transfer messages can now be recognized and decoded into transfer details, including recipient, amount, and token contract.
  - Non-standard or ambiguous contract messages continue to be handled as generic contract calls.

- **Bug Fixes**
  - CW20 transfer amounts must now be positive and within the supported Uint128 maximum.
  - Invalid, non-canonical, or malformed transfer payloads are safely rejected.


gomesalexandre (approved):
lgtm. the cw20 decode stays fail-closed, the byte-for-byte canonical check matches the builder path, and the new tests cover the unsafe shapes that should stay generic contract calls.

PR #1505, opened 2026-07-22
fix(sdk): raw broadcast paths fail closed - no false success on ack-only or malformed responses

gomesalexandre:
Continues today's broadcast-hunt hardening campaign (cosmos raw got the same treatment in #1426, sui raw in #1413) - three more bounded fixes, one PR, all fail-closed by construction: throw or report `pending` on ambiguity, never fabricate a success hash/status.

## what

- `RawBroadcastService.broadcastRawTx` (the public raw-broadcast path, bypasses the signing-input resolvers' verification net):
  - Polkadot/Bittensor: a malformed RPC response with neither `result` nor `error` was falling through to `return response.result`, handing back `undefined` as if it were the extrinsic hash.
  - Tron: only checked `response.code !== 'SUCCESS'`, missing Tron's own explicit `result: false` failure signal when no `code` is present.
  - Solana: `sendRawTransaction` resolving just meant "the node queued it", not confirmation - the raw path returned that ack signature with zero on-chain verification.
- `getTronTxStatus` (core status resolver): a `receipt.result` value we've never seen before was narrated as `success`, because the terminal-failure set was a deny-list ("not a known failure" = success). Inverted to an allowlist.
- `BroadcastService.broadcastTx`: discarded the hash the broadcast resolver itself returned (utxo/cardano/tron all echo the node's own hash back) and unconditionally recomputed one locally instead.

Cosmos and Sui raw broadcast already have assert guards from the prior two PRs in this campaign - untouched here, just added regression coverage confirmed already green.

## how

**Polkadot/Bittensor raw** - added the exact same "response must have exactly one of `result`/`error`" guard the CORE resolvers (`packages/core/chain/tx/broadcast/resolvers/{polkadot,bittensor}.ts`) already have, which the raw path was missing:
```ts
if (!response.result) {
  throw new Error('Polkadot broadcast failed: missing extrinsic hash in RPC response')
}
```

**Tron raw** - widened the failure check to match the core resolver's `result.result === false || result.code` condition (previously only checked `code`):
```ts
if (response.result === false || (response.code && response.code !== 'SUCCESS')) { ... }
```

**Solana raw** - `sendRawTransaction` is fire-and-forget by protocol design; it never returns an inline execution result the way Cosmos's `broadcastTx` or Sui's `executeTransactionBlock` do, so there's no "assert on the response" fix available here. Added a bounded, non-blocking `getSignatureStatuses` check after a successful send: throws only on an *explicit* on-chain error already attached to the signature, never on "not found yet" (the normal state seconds after submission) - so it doesn't fabricate success on a known-bad signature without adding real broadcast latency or turning this into a "wait for confirmation" call. Full confirmation is still the caller's job via a status poll, same as the core resolver's own documented trade-off (see the comment trail on `packages/core/chain/tx/broadcast/resolvers/solana.ts` from #874).

**getTronTxStatus** - this one has history: PR #545 (me, reviewed by NeO) originally tried "any non-null `receipt.result` = error", which NeO's live-mainnet test caught as a false FAILURE on a real successful USDT TRC20 transfer (`receipt.result: "SUCCESS"`). The fix at the time was a deny-list ("only known failure codes = error, everything else = success") - correct for that bug, but broader than necessary, since it now silently maps ANY future/unrecognized code to success too. Tightened to an allowlist that still keeps `SUCCESS`/absent = success (so NeO's live tx stays green) but a genuinely unrecognized value now resolves to `pending` instead of `success`. `TRON_RECEIPT_FAILURE_RESULTS` isn't exported, so nothing outside this file needs updating - but vultiagent-app has its own local Tron status probe that should get the mirror inversion in a follow-up (not touched here per scope).

**BroadcastService** - added `extractResolverTxHash()`: prefers a string return (utxo/cardano) or a `{ txid }` object (tron) from the broadcast resolver over calling `getTxHash` locally; falls back to the local computation only when the resolver returned nothing (evm/cosmos/sui/ripple/ton/polkadot/bittensor all resolve void today).

## cross-consumer impact

`@vultisig/sdk` is bundled into flagship ios/android/windows in addition to any third-party raw-broadcast consumer, so being honest about what changes:

- **Polkadot/bittensor/tron raw broadcast**: a caller who previously got `undefined`/a false-success hash back from a malformed or explicitly-failed response now gets a thrown error instead. Anyone relying on `vault.broadcastRawTx()` for these three chains was either already broken downstream (chasing an `undefined` "hash") or silently reporting a failed tx as sent - this is strictly safer, but it IS a new throw path a caller must handle if they weren't already.
- **Solana raw broadcast**: happy path unchanged in the vast majority of cases (a legit broadcast still returns the signature after one cheap extra status check that no-ops on "not found yet"). The only newly-thrown case is a signature the node can already tell you failed - narrow, and strictly an improvement.
- **getTronTxStatus**: any consumer polling Tron tx status (app, CLI, anything importing `@vultisig/core-chain`) that happens to hit a truly unrecognized `receipt.result` value will now see `pending` forever instead of a (fabricated) `success`. This is the one that trades a UX regression (looks stuck) for a fund-safety improvement (never confidently says "your funds moved" when we don't actually know that). Given the known Tron failure codes are exhaustively documented and haven't changed, this should be a near-zero-frequency case in practice.
- **BroadcastService hash preference**: only observable if a resolver-returned hash and a locally-computed one would ever disagree - which, if it ever happened, means the local computation had a bug and the node's value is the one that was actually right. No observable change in the (expected) common case where they agree.

## testing

New tests, each with a mutation-verify pass (reverted the src hunk, confirmed the named test goes red, restored):

- `RawBroadcastService.test.ts`: `throws on a malformed Polkadot response with neither result nor error`, `throws on a malformed Bittensor response with neither result nor error`, `throws on a Tron response with result:false and no code`, `throws instead of returning a signature when Solana reports an on-chain error for it` (+ two pinning the non-regression: still-pending and status-check-itself-erroring both still return the signature).
- `tron.test.ts` (core): `returns pending (never success) for an unrecognized receipt.result value` (replaces the old test that pinned the deny-list behavior as intentional).
- `BroadcastService.test.ts` (new file, zero prior coverage): `prefers the resolver-returned hash ... (utxo/cardano)`, `prefers the resolver-returned txid ... (Tron)`, `falls back to the locally computed hash when the resolver returns void`, `falls back ... when the resolver returns an empty string / an object without a usable txid`, `resolves each broadcast input independently and returns the last transaction hash (approve + swap)`, `wraps a broadcast failure in a BroadcastFailed VaultError`.

Full suite runs: `yarn test:core` (2041 passed, 1 skipped) and SDK `yarn test` (199 files / 2970 tests, all green) both clean. `yarn typecheck`, `yarn lint`, `yarn format:check`, `yarn quality:architecture`, `yarn quality:complexity`, `yarn quality:duplication`, `yarn quality:secrets`, `yarn changeset:check-sdk-bundle` all pass. Did not run `quality:contracts`/`cli:build`/`mcp:build` - those need a from-scratch CLI/MCP build unrelated to this diff's scope (three files, no CLI/MCP surface touched). One pre-existing, unrelated typecheck gap confirmed on origin/main too (missing optional native dep `node-pty` in an unrelated CLI test file, not present in node_modules on either tree).

🤖 Agent-generated

## Summary by CodeRabbit

- **Bug Fixes**
  - Raw transaction broadcasts now fail closed when RPC responses are invalid, ambiguous, or indicate on-chain errors.
  - Improved broadcast error handling across Solana, Tron, Polkadot, and Bittensor.
  - Unknown/unsupported Tron receipt results are now treated as pending (not successful).
  - Transaction hashes now prefer the resolver-provided value when available, with consistent local fallback.
- **Tests**
  - Expanded unit test coverage for resolver hash selection and additional Solana/Tron/Polkadot/Bittensor failure modes.

### DIFF
--- a/packages/core/chain/tx/status/resolvers/tron.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.ts
@@ -70,11 +70,6 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash 
     return { status: 'pending', isKnown: true }
   }
 
-  // 3. receipt.result in terminal failure set → error
-  //    receipt.result == null → success (native TRX send, no contract result)
-  //    receipt.result == 'SUCCESS' → success (TRC20/smart-contract success, protobuf3 non-default enum)
-  //    receipt.result unknown → treat as success (safer than false-failure for user)
-  const status = tx.receipt.result != null && TRON_RECEIPT_FAILURE_RESULTS.has(tx.receipt.result) ? 'error' : 'success'
   const feeCoin = chainFeeCoin[Chain.Tron]
   const receipt =
     tx.fee != null
@@ -85,5 +80,22 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash 
         }
       : undefined
 
-  return { status, receipt }
+  // 3. receipt.result decides success vs failure vs unknown - an ALLOWLIST, not a deny-list:
+  //    - in the known terminal failure set → error
+  //    - null → success (native TRX send, no contract result)
+  //    - 'SUCCESS' → success (TRC20/smart-contract success, protobuf3 non-default enum;
+  //      pinned against NeO's live mainnet tx 1540b1b3, see tron.test.ts)
+  //    - anything else (a receipt.result value we've never seen) → pending, NEVER success.
+  //      The block is already final at this point, so this isn't "still processing" - it's an
+  //      unrecognized terminal outcome, and a new Tron enum value must not be silently narrated
+  //      as a successful fund movement just because it isn't on the known-failure list.
+  if (tx.receipt.result != null && TRON_RECEIPT_FAILURE_RESULTS.has(tx.receipt.result)) {
+    return { status: 'error', receipt }
+  }
+
+  if (tx.receipt.result == null || tx.receipt.result === 'SUCCESS') {
+    return { status: 'success', receipt }
+  }
+
+  return { status: 'pending', isKnown: true, receipt }
 }
--- a/packages/core/chain/tx/status/resolvers/tron.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.ts
@@ -10,6 +10,12 @@ import { TxStatusResolver } from '../resolver'
 // Protobuf3 JSON serializes non-default enum values by name: successful TRC20 calls emit
 // receipt.result="SUCCESS" (non-default, value=1). Native TRX transfers emit no receipt at all.
 // Known failure codes verified against Tron protocol and live mainnet responses.
+//
+// sdk#1505 should-fix (S1): the allowlist-inversion below correctly means an UNKNOWN result
+// never reads as false-success, but an unlisted contractResult still resolves 'pending' until
+// the poll times out rather than promptly 'error'. The 9 codes below complete the documented
+// core/Tron.proto contractResult enum (values 3-9, 12, 14) so every KNOWN failure surfaces
+// immediately instead of waiting out the poll.
 const TRON_RECEIPT_FAILURE_RESULTS = new Set([
   'FAILED',
   'OUT_OF_ENERGY',
@@ -17,6 +23,15 @@ const TRON_RECEIPT_FAILURE_RESULTS = new Set([
   'OUT_OF_TIME',
   'BANDWIDTH_ERROR',
   'ACCOUNT_FREEZED',
+  'TRANSFER_FAILED',
+  'BAD_JUMP_DESTINATION',
+  'OUT_OF_MEMORY',
+  'STACK_OVERFLOW',
+  'STACK_TOO_SMALL',
+  'STACK_TOO_LARGE',
+  'ILLEGAL_OPERATION',
+  'PRECOMPILE_CONTRACT_ERROR',
+  'JVM_STACK_OVER_FLOW',
 ])
 
 type TronTxInfoResponse = {
--- a/packages/core/chain/tx/status/resolvers/tron.ts
+++ b/packages/core/chain/tx/status/resolvers/tron.ts
@@ -13,9 +13,11 @@ import { TxStatusResolver } from '../resolver'
 //
 // sdk#1505 should-fix (S1): the allowlist-inversion below correctly means an UNKNOWN result
 // never reads as false-success, but an unlisted contractResult still resolves 'pending' until
-// the poll times out rather than promptly 'error'. The 9 codes below complete the documented
-// core/Tron.proto contractResult enum (values 3-9, 12, 14) so every KNOWN failure surfaces
-// immediately instead of waiting out the poll.
+// the poll times out rather than promptly 'error'. The codes below complete the documented
+// core/Tron.proto contractResult enum (every non-DEFAULT, non-SUCCESS member: values 2-16) so
+// every KNOWN failure surfaces immediately instead of waiting out the poll. Kept in exact
+// parity with the app's TRON_TERMINAL_FAILURE_RESULTS (vultiagent-app txVerifier.ts, app#2198)
+// so a Tron failure buckets identically in the SDK resolver and the app tx-verifier.
 const TRON_RECEIPT_FAILURE_RESULTS = new Set([
   'FAILED',
   'OUT_OF_ENERGY',
@@ -30,8 +32,10 @@ const TRON_RECEIPT_FAILURE_RESULTS = new Set([
   'STACK_TOO_SMALL',
   'STACK_TOO_LARGE',
   'ILLEGAL_OPERATION',
-  'PRECOMPILE_CONTRACT_ERROR',
+  'PRECOMPILED_CONTRACT',
   'JVM_STACK_OVER_FLOW',
+  'UNKNOWN',
+  'INVALID_CODE',
 ])
 
 type TronTxInfoResponse = {
--- a/packages/sdk/src/vault/services/RawBroadcastService.ts
+++ b/packages/sdk/src/vault/services/RawBroadcastService.ts
@@ -236,6 +236,27 @@ export class RawBroadcastService {
     }
 
     if (!signature) throw new Error('No transaction signature returned')
+
+    // sendRawTransaction only confirms the node ACCEPTED the payload into its queue - it is
+    // fire-and-forget by protocol design and never returns an execution result inline (unlike
+    // Cosmos/Sui's broadcast RPCs, which block until inclusion). Full confirmation is out of
+    // scope here and remains the caller's job via a status poll (mirrors the core resolver's
+    // documented trade-off, packages/core/chain/tx/broadcast/resolvers/solana.ts). But a
+    // signature the node already knows to have failed must not be handed back as a "hash" -
+    // this bounded, non-blocking status check catches that without adding real broadcast
+    // latency: it never blocks/throws on "not yet confirmed" (the normal state right after
+    // submission), only on an explicit on-chain error already attached to this signature.
+    const { data: statuses } = await attempt(
+      client.getSignatureStatuses([signature], { searchTransactionHistory: true })
+    )
+    const signatureStatus = statuses?.value?.[0]
+    if (signatureStatus?.err) {
+      throw new VaultError(
+        VaultErrorCode.BroadcastFailed,
+        `Solana transaction was submitted but failed on-chain: ${JSON.stringify(signatureStatus.err)}`
+      )
+    }
+
     return signature
   }
 
@@ -352,6 +373,13 @@ export class RawBroadcastService {
       throw new Error(`Polkadot broadcast failed: ${response.error.message}`)
     }
 
+    // Per JSON-RPC 2.0 a valid response must have exactly one of `result` / `error`. If both
+    // are missing (malformed gateway response, truncated body, ...) do not silently return
+    // `undefined` as a success hash - fail closed instead.
+    if (!response.result) {
+      throw new Error('Polkadot broadcast failed: missing extrinsic hash in RPC response')
+    }
+
     return response.result
   }
 
@@ -383,6 +411,12 @@ export class RawBroadcastService {
 
     if (response.error) {
       throw new Error(`Bittensor broadcast failed: ${response.error.message}`)
+    }
+
+    // Same JSON-RPC 2.0 invariant as Polkadot: a response with neither `result` nor `error`
+    // is malformed, not a success - fail closed instead of returning `undefined` as a hash.
+    if (!response.result) {
+      throw new Error('Bittensor broadcast failed: missing extrinsic hash in RPC response')
     }
 
     return response.result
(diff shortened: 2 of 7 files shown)


packages/core/chain/tx/status/resolvers/tron.ts
```diff
@@ -85,5 +80,22 @@ export const getTronTxStatus: TxStatusResolver<OtherChain.Tron> = async ({ hash
...
+  // 3. receipt.result decides success vs failure vs unknown - an ALLOWLIST, not a deny-list:
+  //    - in the known terminal failure set → error
+  //    - null → success (native TRX send, no contract result)
+  //    - 'SUCCESS' → success (TRC20/smart-contract success, protobuf3 non-default enum;
+  //      pinned against NeO's live mainnet tx 1540b1b3, see tron.test.ts)
+  //    - anything else (a receipt.result value we've never seen) → pending, NEVER success.
+  //      The block is already final at this point, so this isn't "still processing" - it's an
+  //      unrecognized terminal outcome, and a new Tron enum value must not be silently narrated
+  //      as a successful fund movement just because it isn't on the known-failure list.
+  if (tx.receipt.result != null && TRON_RECEIPT_FAILURE_RESULTS.has(tx.receipt.result)) {
```
NeOMakinG, line 111: should-fix: this allowlist inversion is the right fund-safety move, but `TRON_RECEIPT_FAILURE_RESULTS` (defined above, unchanged) is missing several documented Tron `contractResult` failure codes — `TRANSFER_FAILED`, `BAD_JUMP_DESTINATION`, `OUT_OF_MEMORY`, `STACK_OVERFLOW`/`STACK_TOO_SMALL`/`STACK_TOO_LARGE`, `ILLEGAL_OPERATION`, `PRECOMPILE_CONTRACT_ERROR`, `JVM_STACK_OVER_FLOW`. With the deny-list→allowlist flip, a *known* failed contract call carrying one of these now falls to the final `pending` branch (looks stuck until the poll times out) instead of `error`. Never a false success — good — but consider completing the set so documented failures resolve promptly to `error`.

packages/sdk/src/vault/services/RawBroadcastService.ts
```diff
@@ -352,6 +373,13 @@ export class RawBroadcastService {
       throw new Error(`Polkadot broadcast failed: ${response.error.message}`)
     }
 
+    // Per JSON-RPC 2.0 a valid response must have exactly one of `result` / `error`. If both
+    // are missing (malformed gateway response, truncated body, ...) do not silently return
+    // `undefined` as a success hash - fail closed instead.
+    if (!response.result) {
```
NeOMakinG, line 379: suggestion/q: this is strictly safer than the pre-PR `return undefined` (a false-success unpollable hash). One residual: a malformed *gateway* response (neither `result` nor `error`) can also mean the extrinsic WAS accepted into the pool — a hard `BroadcastFailed` here invites a re-send (bounded by Substrate nonce-replay, so not a clean double-spend). Consider deriving the extrinsic hash locally on this branch so a maybe-broadcast tx stays pollable (unconfirmed + hash) rather than a blind re-send. Same note applies to the Bittensor guard below. Non-blocking.

NeOMakinG (commented):
COMMENTED (draft — not approving). Fund-critical raw-broadcast fail-closed hardening. Two-lane review: Claude load-bearing (full source walk + 6-mutation battery, all RED); Codex gpt-5.4 lane came back 0-byte (session stall) and was skipped.

## testing approach
Reviewed at HEAD `5176868e` in an isolated worktree (SDK vitest uses explicit relative `resolve.alias`, and the core status test imports `./tron` relatively — so worktree mutations are genuinely exercised, confirmed by RED, not laundered back to `main` via the node_modules alias trap). Baseline: 34 SDK (7 `BroadcastService` + 27 `RawBroadcastService`) + 16 core `getTronTxStatus`, all green.

I walked the actual source for every changed path and the collaborators the diff does **not** touch — specifically the core broadcast resolvers — to confirm the new `extractResolverTxHash` can never surface a hash for a failed tx.

Mutation battery (each RED, mutation grep-verified physically present in the worktree file):

| # | mutant | result |
|---|--------|--------|
| M1 | neuter Solana on-chain-`err` throw | 1 failed |
| M2 | neuter Polkadot `!response.result` guard | 1 failed |
| M3 | neuter Bittensor `!response.result` guard | 1 failed |
| M4 | drop `result === false` from Tron raw check | 1 failed |
| M5 | tron status unknown `receipt.result` → success | 1 failed |
| M6 | `BroadcastService` always use local hash | 3 failed |

Both directions are pinned: fail-closed on ambiguity/known-failure (M1–M5), **and** the double-spend-safe direction — Solana's status check is best-effort, so a malformed/errored status still returns the signature (two dedicated tests), only an explicit on-chain `err` throws.

## the safety proof for `extractResolverTxHash`
The obvious worry: does the core tron broadcast resolver hand back `{ txid }` on a *failed* broadcast, so the new preference newly reports a hash for a failed tx? No — `core/chain/tx/broadcast/resolvers/tron.ts` throws (`if (result.result === false || result.code)`) **before** `return result`, so `coreBroadcastTx` throws and `BroadcastService` wraps a `VaultError` — the failed `{ txid }` never reaches `extractResolverTxHash`. utxo/cardano resolvers likewise return the node hash only on the success branch (the verify-race path returns `null` → local fallback). The change is sound.

Ack-only hashes are preserved for the caller to poll on every raw path (Solana `signature`, Polkadot/Bittensor `response.result`, Tron `response.txid`), and the tron status change makes an unknown terminal outcome resolve `pending` (never a false confirm) — so it composes correctly with the receipt-poll contract (#2177/#2236): broadcast → hash+unconfirmed → poll → confirm.

## should-fix
- **S1 — tron status allowlist is incomplete.** `TRON_RECEIPT_FAILURE_RESULTS` (unchanged in this PR, so noting here) omits several documented Tron `contractResult` failure codes: `TRANSFER_FAILED`, `BAD_JUMP_DESTINATION`, `OUT_OF_MEMORY`, `STACK_OVERFLOW`/`STACK_TOO_SMALL`/`STACK_TOO_LARGE`, `ILLEGAL_OPERATION`, `PRECOMPILE_CONTRACT_ERROR`, `JVM_STACK_OVER_FLOW`. Post-PR these resolve to `pending` (until the poll times out) instead of `error`. This is *fund-safe* — the whole point, never a false success — but a genuinely-failed contract call now reads "stuck" rather than promptly "failed". `REVERT` and `OUT_OF_ENERGY` (the two commonest) are covered, so real exposure is small; still worth completing the set so known failures surface as `error`. See inline.

## suggestion / q
- **S2 — Polkadot/Bittensor malformed-submit fails hard.** A malformed submit response (neither `result` nor `error`) now throws `BroadcastFailed`. Strictly safer than the pre-PR `undefined`-as-hash. But a malformed *gateway* response can also mean the extrinsic **was** accepted into the pool, and a hard failure invites a re-send (bounded by Substrate nonce-replay protection, so not a clean double-spend). Consider deriving the extrinsic hash locally on this branch so a maybe-broadcast tx stays pollable (unconfirmed + hash) rather than a blind re-send. Non-blocking. See inline.

- **S3 (out of scope, unchanged file)** — `broadcastRippleRawTx` returns `response.result.tx_json.hash` without checking the XRPL `engine_result`; a `tec`/`tem`/`tef` result returns a hash for a tx that failed or was not applied. Same fail-closed family as this campaign — worth a follow-up, not this PR.

## risk
Low. Every change moves strictly toward fail-closed; no path that previously reported a genuine success now rejects it. The one availability trade (unknown tron `receipt.result` → pending, and S1's omitted codes → pending) is the intended direction and is acknowledged in the PR description.

## verdict
COMMENTED (draft). Clean, well-tested, correctly fail-closed in both directions. No blockers. Address S1 (complete the tron failure allowlist) before/with the follow-up that mirrors this to the app's local tron probe; S2/S3 are optional hardening.

---
*Reviewed by an automated agent (Claude load-bearing; Codex lane dead/0-byte, skipped). Mutation-verified against HEAD 5176868e.*

packages/core/chain/tx/status/resolvers/tron.ts
gomesalexandre: Tackled in 1d0cca63: added all 9 codes to TRON_RECEIPT_FAILURE_RESULTS with a parameterized test per code. Mutation-verified. Full details in the PR comment.

NeOMakinG (approved):
Re-review at HEAD `fdfda81` (fix-and-approve on top of gomes' `1d0cca63`).

## testing approach
Unique worktree at the PR HEAD; core status suite via `.config/vitest.core.config.ts` (workspace-alias resolution — mutations grep-verified present in worktree, absent from canonical). Cross-checked the SDK failure set against the app's authoritative `TRON_TERMINAL_FAILURE_RESULTS` (app#2198) member-by-member.

## prior issues — status
|---|---|---|---|---|---|
| r1 | S1 | should-fix | `TRON_RECEIPT_FAILURE_RESULTS` omits documented Tron contractResult failure codes → known failures read pending-until-timeout instead of error | ⚠️→✅ | `1d0cca63` added 9 codes but with a **fabricated name** `PRECOMPILE_CONTRACT_ERROR` (value 5 is `PRECOMPILED_CONTRACT` in core/Tron.proto) — that entry never matches a real response, so a precompiled-contract failure still read pending. Also still missing `UNKNOWN` (14) and `INVALID_CODE` (16). Corrected in `fdfda81`: renamed to `PRECOMPILED_CONTRACT`, added `UNKNOWN`+`INVALID_CODE`; the SDK set is now in **exact parity** with the app's 17-member set. |
| r1 | S2 | suggestion | Polkadot/Bittensor malformed-submit hard-fails instead of deriving a pollable extrinsic hash | ℹ️ | out of this commit's scope; unchanged, non-blocking |
| r1 | S3 | q | `broadcastRippleRawTx` returns a hash without checking XRPL `engine_result` | ⏭️ | pre-existing, same family, follow-up |

## verified this round
- **Set completeness / parity**: post-fix SDK set = {FAILED, OUT_OF_ENERGY, REVERT, OUT_OF_TIME, BANDWIDTH_ERROR, ACCOUNT_FREEZED, TRANSFER_FAILED, BAD_JUMP_DESTINATION, OUT_OF_MEMORY, STACK_OVERFLOW, STACK_TOO_SMALL, STACK_TOO_LARGE, ILLEGAL_OPERATION, PRECOMPILED_CONTRACT, JVM_STACK_OVER_FLOW, UNKNOWN, INVALID_CODE} — identical (as a set) to the app's `TRON_TERMINAL_FAILURE_RESULTS`. No drift.
- **Every added code is a genuine contractResult failure** (values 2-16, non-DEFAULT/non-SUCCESS). No SUCCESS/receipt-absent marker was added — no success/pending gets flipped to error.
- **The 7+ now-known codes resolve terminal `error`**, not pending-until-timeout (pinned by the `it.each` battery).
- **Unknown → still pending**: a genuinely unrecognized code (`UNKNOWN_FUTURE_CODE`) still buckets `pending, isKnown:true` (never false-success). Only the named enum member `'UNKNOWN'` (value 14) moves to error — distinct concept, correct.
- **Prior fail-closed properties intact**: this commit touches only the tron status resolver + its test — ack-only non-final + hash preserved, malformed → not-success, and the tron broadcast resolver throwing before returning a txid are all unchanged from r1.
- **Tests**: 27 pass. Mutation (drop `PRECOMPILED_CONTRACT`) → the corresponding row goes RED; restore → green.

## risk
Low. Fund-safe in every direction (never false-success). The change moves three additional named Tron failure codes from pending→error and corrects one fabricated name; the fail-closed default is unchanged.

## verdict
Approve. S1 fully resolved and in exact parity with the app; no wrongly-added code; unknown still pending; prior fail-closed properties survived.

NeOMakinG (approved):
Re-approve at HEAD 94d00bad. Delta fdfda81..94d00bad is a single-line arrow-paren lint fix in packages/core/chain/tx/status/resolvers/tron.test.ts (`(result) =>` -> `result =>`), same assertion, zero logic change — no resolver/allowlist edit. Prior fix-and-approve (fail-closed raw broadcast + 17-member Tron contractResult failure allowlist at app-parity, mutation-pinned) stands.

PR #1488, opened 2026-07-22
fix(sdk): publish swap tool subpath

gomesalexandre:
## Summary
- publish `@vultisig/sdk/tools/swap` as a real package subpath
- emit dedicated JS + d.ts bundles for the swap tool barrel
- extend the public subpath regression test and tarball smoke script

## Why
Current `main` already has a coherent `packages/sdk/src/tools/swap/index.ts` public barrel, but no `./tools/swap` export-map entry. That forces consumers to fall back to the broad root surface or mirror/import internals for helpers like `PriceImpactTooHighError` and `MAX_PRICE_IMPACT_PCT`.

Closes https://github.com/vultisig/vultisig-sdk/issues/1487

## Verification
- `corepack yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/public-api/toolsSubpathExports.test.ts`
- `corepack yarn build:sdk`
- `corepack yarn cli:build`
- `corepack yarn sdk:smoke:subpaths`
- `corepack yarn lint`
- `corepack yarn build:all`
- `corepack yarn test`

## Report
- https://gist.github.com/gomesalexandre/0f4c3ce6f6f9f073a5748c8d6843d0b3

PR #1508, opened 2026-07-22
fix: publish gas, policy, swap and prep tool subpaths

gomesalexandre:
## what

Five tool modules already exist under `packages/sdk/src/tools`, but they were never published as package export subpaths, so consumers could not deep-import them:

- `@vultisig/sdk/tools/bridge`
- `@vultisig/sdk/tools/gas`
- `@vultisig/sdk/tools/policy`
- `@vultisig/sdk/tools/swap`
- `@vultisig/sdk/tools/prep`

Each subpath now publishes ESM, CommonJS, and declaration bundles.

## why

This work was rescued from local audit-round commits that never shipped. It completes the subpath-publishing work that landed `@vultisig/sdk/tools/defi` and `@vultisig/sdk/tools/parse`.

## testing

- `yarn install` - passed with existing peer dependency warnings and no lockfile change.
- `NODE_OPTIONS=--max-old-space-size=4096 yarn build:sdk` - passed; all SDK platform bundles and the five new ESM, CommonJS, and declaration bundles were generated.
- `yarn workspace @vultisig/sdk exec vitest run --config tests/unit/vitest.config.ts tests/unit/public-api/toolsSubpathExports.test.ts` - passed, 1 file and 2 tests.
- `node scripts/smoke-sdk-subpath-exports.mjs` - passed; the packed consumer resolved CommonJS, imported ESM, and type-checked all seven published tool subpaths.
- `NODE_OPTIONS=--max-old-space-size=4096 yarn build:all` - passed.
- `NODE_OPTIONS=--max-old-space-size=4096 yarn test` - passed: scripts 24, SDK 2,957, Rujira 550 with 4 skipped, core 2,041 with 1 skipped, CLI 945, and MCP 35.
- `NODE_OPTIONS=--max-old-space-size=4096 yarn typecheck` - passed.
- `yarn check:shared-exports`, CLI and MCP build and smoke checks, the changeset bundle guard, contract checks, lint, knip, format, architecture, duplication, complexity, secrets, audit, and Markdown checks - passed individually.

## risk

Low. This is an additive export map and rollup change. No existing package export entry is changed. Declaration bundles are built in sequential groups to remain within the existing CI heap limit, without changing their outputs. Flagship iOS, Android, and Windows consumers are unaffected by additive subpaths.


NeOMakinG (commented):
Reviewed on a fresh worktree at HEAD `de998299` (draft — posting COMMENT, not approving). This is the 11th SDK subpath-publish PR; reviewed through the sibling-collision lens against open drafts #1480 and #1488.

## testing approach
- `git diff --name-only origin/main...HEAD`: 6 files, **zero `src/tools/*` modules touched** — this is re-export plumbing only (package.json exports map, both rollup configs, the subpath test, the smoke script, one changeset). Every published module is therefore byte-identical to `main` = canonical by construction.
- Ran `tests/unit/public-api/toolsSubpathExports.test.ts` in the worktree (node_modules symlinked from canonical, not installed): 2/2 pass. **Confirmed it pins** — renamed the `./tools/gas` key in the worktree `package.json` (grep-verified physically absent) and the test went RED (1 failed), then restored. Not vacuous, and no `@vultisig/*` alias in the test so no main-shadowing.
- Verified the `build:types` split covers every group: the script runs 4 sequential (`--max-processes 1`) invocations `base,tools,swap,prep`, and the default (`BUILD_TYPES_TARGET` unset → `all`) flattens all groups — no subpath ships without a `.d.ts`.
- Inspected the 5 barrels + `prep/types.ts` for signing-surface leakage.

## blocking
None.

## should-fix
None blocking. Two coordination notes below.

## suggestion
None on the code.

## q
- **Supersede vs conflict with #1480 / #1488.** On `main` only `./tools/parse` and `./tools/defi` are published. #1508 is a strict **superset** of open draft #1480 (policy/bridge/gas — its export blocks are structurally identical to the ones here) and #1488 (swap), plus new `prep`. This is a consolidation. If a sibling and #1508 both land, the clash is a plain git textual conflict on the same `package.json` lines / duplicate rollup `createToolsSubpathConfigs` inputs — resolvable, and **not** the silent esbuild "Multiple exports with the same name" that broke #1188 (that was a barrel dup; these are package.json map keys and separate rollup inputs). Recommend **closing #1480 and #1488 as superseded by #1508** to avoid rebase churn. Is that the intent?
- **gas subpath inherits the #1480 `compareCosts` note.** `tools/gas` re-exports the unchanged `compareCosts`. The default-path ranking concern raised on #1480 lives in that module; since #1508 modifies no module it neither introduces nor fixes it — publishing it deep-importable just widens exposure to a pre-existing behavior. Flagging so it's tracked against the module, not this PR.

## risk
Low. Additive export map + rollup plumbing; no existing export subpath shape changed (the `-6` lines are the `build:types` script swap and the `defineConfig([...])` → grouped-array wrapper, plus mechanical test/smoke edits). `prep` is confirmed vault-free: `VaultIdentity` carries only public keys/`localPartyId`/`libType` with an explicit "no key shares, no signing material — the security boundary is preserved" contract, and `prepare*FromKeys` build *unsigned* `KeysignPayload`s. No keyshare/signing internal on the public surface.

## verdict
COMMENT (draft). No blocking findings. All 5 subpaths re-export canonical, unmodified modules; the test pins; `prep` leaks no signing material. The one actionable item is process, not code: close #1480/#1488 as superseded so the overlapping subpaths don't race to merge.


---
*lanes: Claude correctness (worktree + mutation-pinned test) primary. Codex adversarial lane ran to its cap with 0-byte output — recorded as a dead/skipped lane, not a clean bill. GLM/QA not run (config-only, non-user-facing).*

gomesalexandre:
Rebased onto current main. `./tools/bridge` had already landed on main independently (#1580), so that part of this diff was redundant — dropped it and kept only the genuinely missing `./tools/gas`, `./tools/policy`, `./tools/swap`, and `./tools/prep` subpaths. Retitled to match.

Also simplified `build:types` back to main's plain single rollup process (the concurrency-group split from the original branch isn't needed - all nine subpath type bundles build fine in one process with the existing 8GB heap).

Verified: all 6 platform builds, `build:types`, `changeset:check-sdk-bundle`, `sdk:smoke:subpaths` (runtime resolve + import + tsc against a packed tarball), the subpath export unit test, lint, prettier, `yarn npm audit --recursive --all --severity high`. `yarn workspace @vultisig/sdk typecheck` is clean; root `yarn typecheck` has pre-existing unrelated CLI failures (`cliDescription` on the descriptions type in `clients/cli/src`) that predate this branch and aren't touched by this diff.

rcoderdev (commented):
The non-Node export conditions are wired to Node-built artifacts here. `createSubpathConfigs` builds these files with `preferBuiltins: true` and `VULTISIG_PLATFORM` set to `node`, while `package.json` maps the same files to `browser`, `worker`, and `react-native`.

On this exact head, the emitted `tools/swap` ESM imports `fs` and `path`, and `tools/prep` pulls Node-only modules including `fs`, `crypto`, and networking/stream built-ins. Bundling minimal browser consumers of those declared subpaths fails on the unresolved Node built-ins.

Please either emit platform-specific artifacts for the non-Node conditions or remove those conditions until the graphs are portable, and add per-condition consumer coverage for the published subpaths.

NeOMakinG:
@rcoderdev's finding is real and load-bearing — flagging to owner (@gomesalexandre) for scope decision, not attempting a blind fix. The two paths rcoderdev outlines have very different blast radius:

**Option A: emit platform-specific artifacts.** Requires teaching `createSubpathConfigs` to run per `VULTISIG_PLATFORM` (`browser`/`worker`/`react-native`), emit those bundles alongside the Node one, add `preferBuiltins: false` + polyfill/shim wiring for the non-Node conditions, and expand consumer coverage per condition (four artifacts × N subpaths). Multi-subpath scope, touches build infra — should probably be its own PR series, not this one.

**Option B: remove the non-Node conditions.** package.json export map maps `browser`/`worker`/`react-native` for these subpaths back to Node-imports-only, and the file becomes Node-condition-only. Small diff, honest about current bundling reality, but shrinks the shipped surface (consumers who thought RN could load `@vultisig/sdk/tools/prep` lose that door).

Both need owner sign-off. This PR is also currently CONFLICTING with main, so whichever direction it takes needs a full rebase — leaving that + the direction call to @gomesalexandre.

gomesalexandre:
pushed fix in c99ba7c89 to address rcoderdev's current-head export-condition finding.

what changed
- removed `browser`, `worker`, and `react-native` conditions from the 9 tool subpath exports in `packages/sdk/package.json`
- kept those subpaths truthfully node-only until they have real non-node artifacts
- tightened `packages/sdk/tests/unit/public-api/toolsSubpathExports.test.ts` to assert those false non-node conditions stay absent

receipt
- exact-head export-map probe: `export-map-ok 9`
- focused vitest in the sparse borrowed-install checkout was blocked by the checkout environment, not the patch:
  - default loader: `[TSCONFIG_ERROR] Failed to load tsconfig for 'tests/unit/setup.ts': Tsconfig not found`
  - `--configLoader native`: `ReferenceError: __dirname is not defined in ES module scope`

re-requesting review on the new head.

rcoderdev:
Closing this because bridge and gas exports have already landed independently, while this conflicted consolidation removes non-Node conditions from existing subpaths and lacks a full exact-head CI run; propose any remaining exports narrowly from current main.

PR #1511, opened 2026-07-23
Return canonical hashes for idempotent raw broadcasts

rcoderdev:
Refs #936
Returns canonical locally derived transaction hashes only when raw-broadcast responses prove a duplicate outcome, while ambiguous or unverifiable errors continue to fail closed.

A controlled production-service JSON-RPC run sent exact offline-signed EIP-1559 bytes through `eth_sendRawTransaction`: `already known` returned the canonical keccak hash, while an unrelated RPC rejection remained `BROADCAST_FAILED`.

This is a bounded follow-up to the malformed/ack-only false-success guards in #1505 and release PR #1504; it does not broaden those changes or claim full cross-network closure for #936.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved raw transaction broadcasting across EVM, Cosmos, Solana, Ripple, and Tron.
  - In duplicate-style submission scenarios, the SDK now returns a deterministic locally derived transaction identifier when the broadcast can be safely verified.
  - Duplicate handling is more conservative when verification is ambiguous or fails, preventing false-positive success.
  - Improved network-specific handling and clearer reporting for Tron duplicate responses.
- **Tests**
  - Expanded unit coverage to validate idempotent behavior and fail-closed handling across the supported networks.

### DIFF
--- a/.changeset/twenty-broadcasts-return.md
+++ b/.changeset/twenty-broadcasts-return.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Return locally derived transaction hashes for deterministic raw broadcast duplicate responses.
--- a/.changeset/twenty-broadcasts-return.md
+++ b/.changeset/twenty-broadcasts-return.md
@@ -1,5 +1,5 @@
 ---
-"@vultisig/sdk": patch
+'@vultisig/sdk': patch
 ---
 
 Return locally derived transaction hashes for deterministic raw broadcast duplicate responses.
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -10,5 +10,23 @@ npmAuditIgnoreAdvisories:
   # module-load console.warn). That pin holds it at 1.1.5, so a published fix would NOT
   # be picked up by `yarn up` — drop the patch resolution as well when revisiting this.
   - '1103747'
+  # GHSA-mh99-v99m-4gvg (CVE-2026-14257): brace-expansion DoS via unbounded
+  # expansion length. The advisory range is `<= 5.0.7` with a single patched
+  # release, 5.0.8 — there is no fixed 1.x or 2.x line, so the 1.1.16 and 2.1.2
+  # copies in the tree are in range with nothing to upgrade to.
+  #
+  # They CANNOT simply be resolved to 5.0.8: v5 is ESM-first and its CJS build
+  # exports a named `{ expand }` instead of a callable default, so the v1/v2
+  # consumers (minimatch@3 and @5, under eslint and the electron packaging
+  # tooling) throw "expand is not a function" the first time they evaluate a
+  # pattern containing braces. Non-brace patterns work, so this does NOT show up
+  # in lint / build / test runs — it only surfaces at runtime on `{a,b}` input.
+  #
+  # Owner: SDK maintainers. Revisit when the 1.x/2.x lines get a backport, or
+  # when the last minimatch@3/@5 consumer is gone (`yarn why minimatch`).
+  # NOTE: package.json pins `brace-expansion@npm:^5.0.5` to 5.0.8, which is the
+  # real fix for the only line that HAS one. Keep that pin: this ignore entry
+  # would otherwise also mask a regression back to the vulnerable 5.0.7.
+  - '1124334'
 
 yarnPath: .yarn/releases/yarn-4.16.0.cjs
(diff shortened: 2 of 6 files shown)


gomesalexandre (changes requested):
Nice direction overall. The EVM and Solana duplicate handling moved in the safe direction, and the ambiguous cases now fail closed. But I’m not comfortable approving this head yet because there are still two protocol-shape gaps that can misreport raw-broadcast outcomes.

1. Cosmos duplicate path can fail open
- `packages/sdk/src/vault/services/RawBroadcastService.ts:321`
- On the duplicate branch you immediately return `deriveCosmosRawTxHash(rawTx)` for `tx already exists in cache`.
- The fresh path below correctly checks `assertIsDeliverTxSuccess(result)` because a Cosmos tx can be included on-chain and still execution-fail.
- That means a retry of the exact same raw tx can now return a success hash for a tx that already landed and reverted, which is the same bug class this PR just fixed on the Solana duplicate path.
- Please mirror the Solana approach here: on the duplicate branch, best-effort verify the derived hash on-chain and only treat it as success if execution actually succeeded. Otherwise fail closed.

2. Tron duplicate detection is keyed off the wrong field, and the tests pin a non-wire shape
- `packages/sdk/src/vault/services/RawBroadcastService.ts:549-552`
- `packages/core/chain/tx/broadcast/resolvers/tron.ts:17-21` already shows the real Tron shape: `code` is authoritative and `message` is hex-encoded.
- Here `errorMsg = response.message || response.code` means a real `DUP_TRANSACTION_ERROR` response with a hex message will miss `isInError(...)`, so the new idempotent-return branch never fires on the actual wire shape.
- The new tests also use plaintext `message: 'DUPLICATE_TRANSACTION'` / `code: 'ERROR'` instead of the canonical duplicate shape, so they don’t pin the real behavior.
- Please match the duplicate markers against `response.code`, decode `message` only for display, and update the tests to the real response form.

3. The new Ripple tests don’t match the real submit response shape
- `packages/sdk/tests/unit/vault/services/RawBroadcastService.test.ts:580-613`
- `packages/core/chain/tx/broadcast/resolvers/ripple.ts:41-77` handles XRPL `submit` as a resolved response with `engine_result`, not a rejected request.
- Right now the tests assert fail-closed behavior for `tefPAST_SEQ` and idempotent success for `tefALREADY` via `mockRejectedValue(...)`, which does not pin the real path this raw service will see.
- At minimum, please rework these tests to the real resolved response shape. I’d strongly prefer aligning the raw service with the resolver semantics too so ambiguous XRPL results do not accidentally fall through as hash-success.

I did not approve because the Cosmos duplicate branch can currently report false success, and the Tron duplicate branch is not grounded to the real response shape yet.

gomesalexandre (commented):
static pass on current head `f04e1e11`: I re-reviewed the narrowed diff and the chain-specific duplicate handling now looks fail-closed in the places I cared about.

what I verified locally:
- `yarn build:shared`
- `yarn vitest run --config tests/unit/vitest.config.ts` -> 43/43 `RawBroadcastService` tests green
- `tsc -p .config/tsconfig.json --noEmit` clean

notes:
- cosmos duplicate path now verifies the derived hash on-chain before returning success
- tron duplicate handling is keyed off the real wire `code` field and cross-checks the derived hash against `txID`
- ripple now uses resolved `engine_result` semantics, so `tefALREADY` / `terQUEUED` are handled on the real path and `tefPAST_SEQ` stays fail-closed
- solana duplicate handling is much safer now; only non-blocking follow-up I see is that an `AlreadyProcessed` retry still returns the signature if `getSignatureStatuses` itself flakes or returns null during duplicate verification. that's worth tightening later, but I don't think it blocks this diff.

I am not stamping this run yet because this is substantive raw-broadcast behavior in the SDK and it still wants the public-surface receipt from the repo rulebook: build the SDK + CLI and show the changed path from the actual CLI/sdk surface, not only the internal service tests.

gomesalexandre (approved):
looks good.

i re-checked the current head remotely and the duplicate/raw-broadcast safety gaps are closed now:
- solana duplicate retries fail closed when the status lookup comes back empty
- ripple `tefALREADY` verifies the derived hash instead of returning it blind, and the ripple status resolver marks unknown/malformed lookups as `isKnown: false`
- tron duplicate responses verify the locally derived hash before reporting success
- cosmos duplicate-cache handling verifies via `getTx` before returning the canonical hash

receipts:
- remote file-by-file pass on head `2422b1f1`
- no dependency churn in this diff, only the changeset + `RawBroadcastService` + tests
- `Lint and Type Check` is red on this branch, but the same static gate is red on other unrelated sdk branches too, so that looks like repo-wide CI drift rather than a regression from this patch

stamp from me. once the shared CI issue is cleared, this should be good to land.

PR #1515, opened 2026-07-23
fix(lib-utils): coalesce in-flight memoizeAsync calls

### LINKED ISSUE #1512: architecture: shared memoizeAsync is split and the shared copy is still race-prone
## Summary

`@vultisig/lib-utils` still ships the weaker `memoizeAsync` implementation while `@vultisig/sdk` carries a safer local fork with in-flight promise coalescing. That leaves shared/core callers race-prone and keeps the fix in the wrong package.

## Evidence

- Shared helper caches only completed values, not in-flight work: `packages/lib/utils/memoizeAsync.ts`
- SDK carries a separate race-safe fork: `packages/sdk/src/utils/memoizeAsync.ts`
- Shared helper is used by live shared/core initialization paths such as:
  - `packages/core/chain/chains/cosmos/client.ts`
  - `packages/core/chain/chains/ripple/client.ts`
  - `packages/core/mpc/lib/initialize.ts`

## Impact

Concurrent identical callers can still duplicate expensive async initialization work outside the SDK-local fork (RPC client setup, WS client setup, MPC/WASM initialization), and the monorepo keeps two subtly different implementations of the same helper.

## Why this belongs in the SDK monorepo

This is pure shared runtime logic. The durable fix is to centralize the race-safe implementation in `@vultisig/lib-utils` and make SDK runtimes consume that shared helper instead of maintaining a second copy.

## Fix sketch

- Promote the SDK-local in-flight promise coalescing behavior into `packages/lib/utils/memoizeAsync.ts`
- Replace the SDK-local copy with a re-export (or delete it once internal callers are migrated)
- Keep a regression test proving concurrent identical calls execute only once
- Bump the affected public packages with a changeset

## Effort

Small / low-risk.

## Report

Round 56 report: https://gist.github.com/gomesalexandre/b18314de06356c9b815bc5deced823fa

gomesalexandre:
## Summary
- centralize the race-safe `memoizeAsync` implementation in `@vultisig/lib-utils`
- replace the SDK-local duplicate with a re-export and point browser/chrome-extension runtimes at the shared helper
- keep regression coverage on concurrent-call coalescing and add changesets for the shared package + SDK runtime consumption

Closes https://github.com/vultisig/vultisig-sdk/issues/1512

## Verification
- `corepack yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/utils/memoizeAsync.test.ts`
- `corepack yarn lint`
- `corepack yarn build:sdk`
- `corepack yarn build:all`
- built-dist runtime proof via `node --input-type=module` import of `packages/lib/utils/dist/memoizeAsync.js` returning `{"results":["ok-x","ok-x","ok-x"],"calls":1}`
- packed tarball proof via `corepack yarn workspace @vultisig/lib-utils pack --out /tmp/improve-r56-sdk-artifacts/vultisig-lib-utils-r56.tgz`
- consumer install/runtime proof via `npm install /tmp/improve-r56-sdk-artifacts/vultisig-lib-utils-r56.tgz` + `import('@vultisig/lib-utils/memoizeAsync')` returning `{"results":["R56","R56","R56"],"calls":1}`

## Known blocker
- `corepack yarn cli:build` still fails on the pre-existing unresolved `@vultisig/core-chain/chains/ripple/address` import from `clients/cli/src/commands/transaction.ts`

## Report
- Round 56 report: https://gist.github.com/gomesalexandre/b18314de06356c9b815bc5deced823fa

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asynchronous caching so simultaneous requests with the same inputs share one in-progress operation.
  * Ensured temporary cache entries are cleared after successful completion or failure, preventing stale pending requests.

* **Refactor**
  * Centralized asynchronous memoization behavior for consistent operation across browser and Chrome extension environments.


NeOMakinG (commented):
## testing approach

Worktree `/tmp/wt-vultisig-sdk-1515` @ e07a704c. Read the new `@vultisig/lib-utils/memoizeAsync` line-by-line against the classic async-memoize failure modes, traced the true blast radius via grep, and ran the vitest suite with a targeted mutation. Confirmed the vitest alias `@vultisig/lib-utils` → `../lib/utils` resolves to the **worktree source** (stack trace `../lib/utils/memoizeAsync.ts`), not canonical/dist — so the mutation is real.

**Baseline:** 13/13 green. **Mutation:** moved the in-flight cleanup out of `finally` (so a rejected call leaks its `pendingPromises` entry — the error-poisoning shape) → 2 tests go RED: `should clean up pending promise on error` and `should not cache errors`. The reject-refire path IS pinned.

## blocking

None.

## preferably-blocking

None.

## should-fix

None.

## suggestion

None.

## q

`suggestion:` The changeset copy says this "update[s] the SDK browser/chrome-extension runtimes" — but the real behavioral scope is wider. On `HEAD~1` the shared `@vultisig/lib-utils/memoizeAsync` had **no** coalescing (0 `pendingPromises`); this PR adds it there. So **every** lib-utils `memoizeAsync` consumer monorepo-wide now coalesces in-flight, not just the two platform files. I verified the fund-adjacent ones — `core/mpc/.../tron/energyPrice.ts` (a `cacheTime`'d fee input), the cosmos/ripple/polkadot clients, and the mldsa/engine/sevenZip init singletons — all import `@vultisig/lib-utils/memoizeAsync`. All are idempotent fetch/init, so coalescing is strictly beneficial (dedups concurrent identical work), the TTL still governs staleness, and it's per-key. No fund risk — but consider widening the changeset note so the broader surface is documented.

## risk

Low. This consolidates two identical race-safe implementations into one and repoints imports. Correctness verified against all five recurring failure modes:

- **Error poisoning** — none. `cache.set` runs only after a successful `await func`; a rejection never writes the cache, and the always-run `finally` deletes the in-flight entry, so the next call re-fires. Not a cached rejection.
- **In-flight cleanup** — correct (`finally` on both resolve and reject).
- **Key isolation** — `JSON.stringify(args)` unchanged, per-key `Map`s; no collision.
- **TTL** — preserved (`cachedResult.updatedAt >= now - cacheTime`); successes still expire.
- **The −98** — the removed SDK copy (`packages/sdk/src/utils/memoizeAsync.ts`) was byte-for-byte the same logic (it too had `pendingPromises` + `finally`). Nothing more-correct was dropped.

## verdict

COMMENTED (draft). Clean — rejection deletes the entry and re-fires (no poisoning), cleanup is correct, keys are isolated, TTL still expires, and the −98 dropped no error-handling or TTL. Approvable once marked ready; the only open item is the documentation-scope `q:` above (non-blocking).

---
🤖 reviewed by Claude (Opus 4.8) — Claude lane; Codex lane skipped (small self-contained change).

rcoderdev (approved):

PR #1527, opened 2026-07-23
fix(tron): share base58check decoder across core and sdk

### LINKED ISSUE #1526: architecture: Tron address Base58Check decoding is duplicated and already drifted inside the sdk monorepo
## Summary

## Evidence

## Why this is a bug

## Target shape

## Impact

## Fix sketch

## Effort
Small.

## Round report

gomesalexandre:
## Summary
- share the canonical Tron Base58Check raw-byte decoder between `@vultisig/core-chain` and the SDK Tron tx builders
- delete the drifted SDK-local prefix-check copy
- add a regression proving the SDK path now accepts the same Nile (`0xa0`) prefix core already accepted

## Problem
`packages/core/chain/coin/balance/resolvers/tron.ts` already accepted both Tron prefixes (`0x41` mainnet, `0xa0` Nile), but `packages/sdk/src/chains/tron/tx.ts:tronAddressToBytes` kept its own decoder copy and only accepted `0x41`.

That left two first-party “canonical” decoders with different behavior inside the same monorepo.

## Fix
- add `base58CheckTronDecodeBytes()` in core-chain beside the existing checksum/prefix validator
- make `base58CheckTronDecode()` delegate to that raw helper
- make `tronAddressToBytes()` delegate to the same shared helper
- extend the SDK Tron unit suite with a Nile-prefix regression
- add a changeset for `@vultisig/core-chain` + `@vultisig/sdk`

## Verification
- `corepack yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/core/chain/coin/balance/resolvers/tron.test.ts packages/sdk/tests/unit/chains/tron-tx.test.ts`
- `corepack yarn build:sdk`

## Receipts
- targeted Vitest: 58 tests passed, including the new `tronAddressToBytes` Nile-prefix regression
- build receipt: `build:shared` + `build:sdk` succeeded in the fresh `/tmp/improve-r59-sdk` clone

## Related
- closes https://github.com/vultisig/vultisig-sdk/issues/1526
- round-59 improve report gist will be backfilled after publication

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Tron address decoding consistency across balance resolution and transaction creation.
  * Added validation to ensure decoded Tron addresses contain the expected network prefix and payload.

* **Tests**
  * Added coverage for decoding Nile testnet Tron addresses, including validation of the resulting bytes and network information.

* **Chores**
  * Documented the patch release for the affected core chain and SDK packages.


packages/sdk/src/chains/tron/tx.ts
```diff
@@ -47,7 +47,7 @@
  */
 
 import { sha256 } from '@noble/hashes/sha2.js'
-import bs58check from 'bs58check'
+import { base58CheckTronDecodeBytes } from '@vultisig/core-chain/coin/balance/resolvers/tron'
```
NeOMakinG, line 50: `suggestion:` the signing-path tx builder now imports the shared decoder from a **balance-resolver** module, transitively pulling `queryUrl`/`tronRpcUrl`/`CoinBalanceResolver` into the signing bundle and re-opening the circular-dep risk the old TODO flagged. Consider hoisting `base58CheckTronDecodeBytes` into a leaf `chains/tron/address` util both sides import. Non-blocking — import resolves and build passes.

packages/sdk/tests/unit/chains/tron-tx.test.ts
```diff
@@ -84,6 +86,25 @@ describe('tron / encodeInt64Varint', () => {
...
+
+    const nileAddress = encode(Buffer.concat([Buffer.from([0xa0]), Buffer.from(TO_EVM_HEX, 'hex')]))
+    const bytes = tronAddressToBytes(nileAddress)
+
+    expect(bytes).toHaveLength(21)
+    expect(bytes[0]).toBe(0xa0)
+    expect(bytesToHex(bytes.subarray(1))).toBe(TO_EVM_HEX)
+  })
+})
+
```
NeOMakinG, line 107: `suggestion:` this pins only the positive Nile-accept. Add a one-line `expect(() => tronAddressToBytes(<corrupted-checksum>)).toThrow()` to pin the checksum-reject invariant directly on the SDK export — today it's only covered transitively via the core suite.

NeOMakinG (commented):
## testing approach
Source-read consolidation review, no full build (draft, single-lane). Verified: core `_decode` = `bs58check.decode` (4-byte SHA-256d checksum, throws on mismatch); the export subpath `./coin/balance/resolvers/tron` is present in `packages/core/chain/package.json` exports (-> dist), so the SDK import resolves and is not inert; core `tron.test.ts` already pins corrupt-checksum throw, wrong-prefix throw (`/invalid tron address prefix/`), and Nile-accept through `base58CheckTronDecode`, which now delegates to the extracted `base58CheckTronDecodeBytes`.

## blocking
None.

## preferably-blocking
None.

## should-fix
None.

## suggestion
- The SDK Tron **tx builder** (signing path) now imports the shared decoder from a **balance-resolver** module, so it transitively pulls `queryUrl`/`tronRpcUrl`/`CoinBalanceResolver` into the signing bundle and re-opens the circular-dep vector the original TODO warned about. A dedicated `chains/tron/address` util would be a cleaner home. Non-blocking.
- The new SDK test only pins the **positive** Nile-accept. A one-line `expect(() => tronAddressToBytes(corrupted)).toThrow()` would pin the reject invariant directly on the SDK export (today it's covered transitively via the core suite since it's the same fn).

## q
- Confirm the SDK signing path accepting `0xa0` Nile addresses (previously mainnet-only) is intended — it matches core and is the stated purpose, just flagging the behavior change explicitly.

## risk
Low. Checksum verification is preserved on both sides (a mistyped/corrupt address still throws). Core `base58CheckTronDecode` is byte-identical (extract-and-delegate). The only behavioral delta is the SDK side loosening from `0x41`-only to `{0x41, 0xa0}` — an intended parity fix, confined to a valid network prefix, with the checksum still gating corrupt input, so no fund-loss vector. No test or guard was dropped in the -22.

## verdict
COMMENTED (draft — no approve). Clean consolidation; checksum-verify intact, behavior-identical on core, intended loosening on SDK, import resolves, coverage adequate.

rcoderdev (approved):

rcoderdev:
Closing this because the TRON decoder fix remains useful, but this old branch is behind a substantially changed main and places the shared helper in a high-churn balance module; please recreate the focused checksum-preserving fix from current main.

PR #1536, opened 2026-07-24
fix(stakekit): keep validatorAddresses nested under args on exit fallback

gomesalexandre:
## Summary
- keep \`validatorAddresses\` nested under \`args\` on the StakeKit REST exit action fallback
- add regression coverage for the exit fallback body shape

Scope narrowed 2026-08-12: this originally also carried an enabled-yield cache-key scoping fix, but that duplicates #1734 (which covers both \`searchYields\` and \`getYield\`, and hashes the apiKey instead of embedding it raw). Dropped that part here to avoid double-shipping the same fix - #1734 is the canonical version for the cache-key scoping.

## Verification
\`\`\`
yarn vitest run --config tests/unit/vitest.config.ts tests/unit/tools/defi/stakekit.test.ts

 Test Files  1 passed (1)
      Tests  19 passed (19)
\`\`\`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed StakeKit exit requests for validator-based products by correctly nesting validator addresses in the request arguments.
  * Improved fallback behavior when the primary integration is unavailable.

* **Tests**
  * Added regression coverage to verify the correct request format and selected validator address.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because the StakeKit request-shape correction is useful, but this old head is behind current main; the focused body-shape fix and regression are cheaper to recreate and revalidate from today's branch than to carry this PR forward.

PR #1540, opened 2026-07-24
fix(sdk): skip inbound-vault check for RUNE/CACAO-source native swaps (6emp)

NeOMakinG:
## Problem

RUNE-source (THORChain-native) swaps pass dry-run but **fail at broadcast** with:

> `Cannot validate THORChain inbound vault for source chain THOR`

This is the concerning class of bug — a **preview/broadcast divergence** where dry-run accepts what broadcast rejects. RUNE could not be swapped out at all. Bead `vultisig-6emp`.

The opposite direction (ETH→RUNE, L1→THOR) works correctly and is untouched here.

## Root cause

`assertNativeSwapReadyForBroadcast` validated an inbound vault for **every** THOR/Maya native swap. When the source chain **IS** the native swap chain (RUNE→L1, CACAO→L1), the swap executes as a native `MsgDeposit` with no inbound vault:

- THORNode's native-source quote returns no `inbound_address`, so the payload's `vaultAddress` falls back to the vault's own THOR address (a non-empty value, not a vault to validate).
- THORNode/mayanode `inbound_addresses` **never list a THOR/MAYA entry**. `getNativeSwapChainId(THORChain)` resolves to `'THOR'`, but the inbound lookup for `'THOR'` never matches → the guard threw `network_error` ("Cannot validate THORChain inbound vault for source chain THOR") at broadcast.

Dry-run skips the broadcast guard entirely, which is exactly why preview passed and broadcast failed.

## Fix

Short-circuit **before** the inbound-vault + trading-halt checks when `chain === native.chain` (native `MsgDeposit`). The quote-expiry check still runs first; the L1→THOR/Maya inbound-vault and trading-halt checks are **unchanged**.

> Note: the guard was relocated from `packages/sdk/src/vault/services/nativeSwapBroadcastGuard.ts` into core-mpc (`packages/core/mpc/keysign/swap/assertNativeSwapReadyForBroadcast.ts`) by #1473 after this fix was first drafted. This PR re-applies the fix to the **current** location and error type (`NativeSwapBroadcastGuardError`), so it edits the file that actually runs.

## Tests

- Added RUNE-source and CACAO-source regression tests (skip inbound fetch), plus "expiry still gates RUNE-source" to prove the expiry check runs before the short-circuit.
- Corrected two pre-existing tests (#972) that asserted a **fabricated** THOR inbound entry (`makeInbound('THOR', …)`) — production never returns a THOR inbound entry — to the real behavior: a THOR-source secure-memo / liquidity-withdraw payload skips the inbound fetch and the still-valid quote is allowed via the native-source branch.

## Proof

- **Control-mutation proven**: reverting the source short-circuit (and rebuilding shared dist) turns exactly the 4 native-source tests red — RUNE-source, CACAO-source, and the two THOR-source secure-memo/liquidity-withdraw tests — while "still rejects expired" stays green. Restoring the fix returns all green.
- **Gates**: `yarn build:all` ✓, `yarn cli:build` ✓ (`@vultisig/cli v2.21.1`), SDK unit 2990 ✓, core tests 2133 ✓ (1 skipped), SDK + core typecheck clean, lint clean on both files.
- A live 1-RUNE→ETH broadcast was **not** independently re-verified in this session (THORNode status endpoints were unreachable from the sandbox). Unit + control-mutation proof stands as the correctness evidence; a real 1-RUNE dry-run+broadcast on a funded vault is the recommended final confirmation before merge.

Bead: `vultisig-6emp`

https://claude.ai/code/session_01DXi9rxKoG85AwKegbecPB8

## Summary by CodeRabbit

* **Bug Fixes**
  * Native-source swaps can now proceed without inbound-vault validation.
  * Valid native-source swaps are no longer blocked by trading-halt checks.
  * Quote expiration checks remain enforced for security.
  * Updated handling for secure-memo and liquidity-withdraw native swaps.


gomesalexandre (changes requested):
The diagnosis in the body is correct and the bug was real: `inbound_addresses` only ever lists external chains, so `getNativeSwapChainId(THORChain) -> 'THOR'` could never match, and every RUNE-source swap got signed and then refused locally. Preview/broadcast divergence on a fund-moving path, worth chasing.

The problem is that this landed on `main` yesterday, in a shape that keeps a protection this PR drops.

## preferably-blocking: superseded by #1756, and merging this would take a guard back off

`b1cbdf353` ("fix(swap): allow RUNE/CACAO-source native swaps past the broadcast guard", Ehsan-saradar, #1756, merged 2026-08-05, `Fixes #1726`) is the same fix for the same bug. It does not early-return. It keeps the fetch and gates only the two checks that cannot apply:

```ts
const sourceIsProtocolChain = chain === native.chain
...
if (!sourceIsProtocolChain && !activeInbound?.address) { throw ... }
...  // halt loop runs unchanged
if (!activeInbound?.address || activeInbound.address.toLowerCase() !== native.vaultAddress.toLowerCase()) { throw ... }
```

This PR returns at `if (chain === native.chain)` before `getInboundAddresses` is ever called, so the sdk#1360 trading-halt loop below it never runs for native-source swaps. That loop checks **both** legs of the route, and the destination leg is the one that matters here. Its own comment spells out why:

> a DESTINATION halt (healthy at quote, halted before broadcast) lets the deposit land while the outbound cannot leave - stuck funds

RUNE -> ETH is exactly that shape: the MsgDeposit lands on THORChain, and if ETH is halted by the time it lands the outbound cannot leave. And #1360 already anticipated the missing-entry case you are working around, four lines further down:

> a route leg that resolves to no inbound entry (e.g. RUNE/CACAO, the native chain itself) is not haltable via this feed, so it is skipped rather than false-blocked

So the halt loop was already tolerant of `THOR` having no entry. The only two things that hard-failed were the `activeInbound` existence check and the vault-address compare, which is precisely what #1756 gated.

### Receipts

Drove the real `assertNativeSwapReadyForBroadcast` with a stubbed `getInboundAddresses` and a fixed clock. Same table, four arms, nothing else changed between them.

```
                                          merge-base (7ca8c406)   this PR (f07fc21f)   origin/main (#1756)
C1 pass-ctrl ETH-source healthy           PASSED  fetch=1         PASSED  fetch=1      PASSED  fetch=1
C2 fail-ctrl ETH-source ETH halted        THREW   halted          THREW   halted       THREW   halted
V1 RUNE-source -> ETH, all healthy        THREW   network_error   PASSED  fetch=0      PASSED  fetch=1
V2 RUNE-source -> ETH, DEST ETH halted    THREW   network_error   PASSED  fetch=0      THREW   halted
V3 RUNE-source -> ETH, global paused      THREW   network_error   PASSED  fetch=0      THREW   halted
V4 RUNE-source expired quote              THREW   expired         THREW   expired      THREW   expired
V5 RUNE-source, no toCoin                 THREW   network_error   PASSED  fetch=0      PASSED  fetch=1
```

C1, C2 and V4 are identical in all four arms, which is what makes this a diff of the guard rather than a broken harness. V1 is the bug and both fixes clear it. V2 and V3 are the split: this PR broadcasts into a halted destination, `main` refuses.

I also built the #1756 shape by hand before I found the commit, from this PR's own diff plus the #1360 comment, and got a table identical to the `origin/main` column. So this is not a preference, it is the same conclusion arrived at twice.

### What I would do

Close this as superseded by #1756 and re-verify against `main`. `git fetch origin && git log --oneline origin/main -- packages/core/mpc/keysign/swap/assertNativeSwapReadyForBroadcast.ts` shows it.

If there is something in your version you want to keep, the one thing this PR has that `main` does not is skipping the network call entirely for native-source (`fetch=0` vs `fetch=1`). That is a real saving and it also means a THORNode outage cannot block a RUNE swap. But it costs the destination-halt check, so it would need to be a separate deliberate change with that tradeoff argued, not a side effect of the early return.

## should-fix: the two rewritten tests drop coverage `main` kept

```ts
-  it('does not exempt a secure-memo payload that carries an inbound vault', async () => {
-    ).rejects.toThrow(/inbound vault address changed/)
+  it('allows a non-expired secure-memo THOR-source payload without fetching inbound addresses', async () => {
+    ).resolves.toBeUndefined()
```

Those two existed to prove `isSecuredAssetWithdrawal` is not over-broad: a lookalike payload that fails the predicate must still be caught by something. After this change both the exempt path and the native-source path return, so the assertion can no longer tell them apart, and a future widening of that predicate on THOR-source payloads goes undetected.

`main` hit the same wall and solved it by re-pointing the assertion at an observable that still discriminates:

```ts
// sdk#1726 changed the non-exempt observable: THOR-source routes no longer run the
// vault-address check (RUNE deposits have no inbound vault), so a lookalike payload that fails
// the secured-withdrawal detection is now caught by the expiry check its exemption would skip.
).rejects.toThrow(/missing or invalid expiration/)
```

Worth reading even if you close this, since it is the pattern for keeping that test honest.

## nit: mergefix

Not pushing to your branch. `git merge origin/main` conflicts in `packages/sdk/tests/unit/vault/services/nativeSwapBroadcastGuard.test.ts` only, four hunks, all in the two rewritten tests plus the added expiry test. `packages/core/mpc/keysign/swap/assertNativeSwapReadyForBroadcast.ts` auto-merges, which is worth knowing because a clean auto-merge there would silently stack this early return on top of `main`'s `sourceIsProtocolChain` and produce exactly the V2/V3 behavior above. If you do rebase rather than close, that file needs a manual read, not a trusted auto-merge.

## nit

The body's note about the guard being relocated by #1473 is accurate and the re-application to the current location was the right call. Nothing wrong with the work, it just got overtaken.

PR #1543, opened 2026-07-24
test(evm): pin the signed chain-id source, not just the viem registry

gomesalexandre:
Extends `packages/sdk/tests/unit/chains/evmChainId.test.ts` so it pins the chain id that actually reaches wallet-core's SIGNED preimage, not just the display registry it already covered.

## what's actually being pinned

The existing block in this file pins `getEvmChainId`/`getEvmChainByChainId`, which read the viem chain registry - the DISPLAY source (what shows up in explorers, UI, etc). It never touched the chain id that lands in the SIGNED preimage on the wallet-core path: `CoinTypeExt.chainId(getCoinType(chain))`, wired through `getTwChainId` (`packages/core/chain/chains/evm/tx/tw/getTwChainId.ts:29`). It also never touched the hand-mirrored copy the React Native tx builder keeps at `packages/sdk/src/platforms/react-native/chains/evm/tx.ts:53`.

A wrong `getCoinType` mapping, or a typo'd entry in the RN table, could silently sign a transaction against the WRONG chain id while every display surface kept showing the right value. That's a replay/misdirection class of bug - the signed tx would be valid on a *different* chain - and the existing suite would stay green the whole time.

## how I found it

Working PR #1541 (Robinhood Chain), two mutations proved the gap with the full suites staying green (2133 core + 2989 sdk tests both times):

- `getCoinType(Robinhood) -> CoinType.smartChain`: wallet-core would sign chain id 56 while every display surface showed 4663.
- The hand-mirrored `evmChainIds` table entry for Robinhood, `4663 -> 4636`: the RN builder's legacy and 1559 preimages would carry the wrong chain id while core reported the right one.

This gap is pre-existing across all 13 EVM chains already on `main` - it has nothing to do with #1541 specifically, Robinhood just tripped over it first, and I'd committed to carrying this fix so a new contributor wouldn't have to.

## the fix

Two new assertions, both driven off the existing `CANONICAL_EVM_CHAIN_IDS` table, which `satisfies Record<EvmChain, number>` already keeps exhaustive against the `EvmChain` enum at compile time - add a 14th EVM chain without pinning its id and the build fails. So a newly added chain is automatically covered by every assertion in the file the moment its id lands in that table:

- `getTwChainId({walletCore, chain}) === String(CANONICAL_EVM_CHAIN_IDS[chain])` - pins the wallet-core signed source.
- `getEvmNumericChainId(chain) === CANONICAL_EVM_CHAIN_IDS[chain]` - pins the RN hand-mirror.

Hyperliquid and Sei aren't wallet-core-native coin types (`getCoinType` maps both to `CoinType.ethereum`), so `getTwChainId` special-cases them by returning the same viem chain objects' `.id` used to build the registry. No skip/carve-out needed - pinning against the canonical table exercises that special case exactly like every other chain.

`getTwChainId` needs a real `walletCore` instance, so this follows the same `initWasm()` pattern already used elsewhere in `packages/sdk/tests` (e.g. `tests/unit/chains/tron-tx.test.ts`) rather than mocking wallet-core - mocking the thing under test would just be another fake-green guard.

## mutation-verified (reverted before commit)

- Mutated BSC's `getCoinType` mapping to the wrong `CoinType` -> the new wallet-core-signed assertion failed (`expected '56', got '42161'`); the original 4 tests and the RN-table assertion stayed green.
- Typo'd the RN table's Ethereum entry (`1 -> 2`) -> the new RN-table assertion failed (`expected 1, got 2`); the wallet-core-signed assertion and the original 4 tests stayed green.

## test plan

- [x] `yarn test:unit` - 201 files, 2988 passed
- [x] `yarn test:core` - 221 files, 2133 passed, 1 pre-existing skip
- [x] `yarn workspace @vultisig/sdk typecheck` - clean
- [x] lint on the touched file - clean
- [x] both mutations reproduced the intended failing assertion, then reverted

### DIFF
--- a/packages/sdk/tests/unit/chains/evmChainId.test.ts
+++ b/packages/sdk/tests/unit/chains/evmChainId.test.ts
@@ -1,7 +1,10 @@
-import { describe, expect, it } from 'vitest'
+import { initWasm, type WalletCore } from '@trustwallet/wallet-core'
+import { getTwChainId } from '@vultisig/core-chain/chains/evm/tx/tw/getTwChainId'
+import { beforeAll, describe, expect, it } from 'vitest'
 
 import type { EvmChain } from '../../../src'
 import { Chain, getEvmChainByChainId, getEvmChainId } from '../../../src'
+import { getEvmNumericChainId } from '../../../src/platforms/react-native/chains/evm/tx'
 
 // Canonical EVM chainIds (numeric), the single source of truth the app and
 // agent-backend-ts previously hand-maintained as their own copies. Pinning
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
## testing approach

Worktree at HEAD `db81d1b9`, node_modules + `packages/**/dist` symlinked from the canonical checkout (no `pnpm install`). Ran the target suite (`tests/unit/chains/evmChainId.test.ts`) with a real `initWasm()` — 6/6 green — then mutation-checked both new assertions against live worktree source (vitest aliases `@vultisig/core-chain` to the worktree, so source mutations take effect). Traced the signed-source claim end-to-end through the keysign resolvers. Codex adversarial lane returned a 0-byte/exit-0 result — recorded as SKIPPED (infra), not a clean bill.

## blocking

None.

## praise

This test pins the right invariant. I confirmed `getTwChainId` is genuinely the value that lands in the SIGNED preimage, not a lookalike: EVM keysign builds `TW.Ethereum.Proto.SigningInput.create({ …, chainId: getEvmTwChainId({walletCore, chain}), … })` (`resolvers/evm/index.ts:234`, `resolvers/evm/erc20.ts:39`), and `getEvmTwChainId = BigInt(getTwChainId(…))`. So the new assertion asserts the exact source of the signed chain id, closing the replay/misdirection gap the pre-existing block (viem DISPLAY registry only) left open.

Mutation-verified both new assertions are load-bearing and correctly isolated:
- Forced BSC's signed source to `'42161'` in `getTwChainId` → the wallet-core assertion went RED (`expected '42161' to be '56'`) while the RN assertion and the original 4 display tests stayed GREEN — i.e. exactly the bug class where the signed tx is valid on the wrong chain while every display surface still shows the right value.
- Typo'd the RN table (`Ethereum: 1 → 2`) → the RN assertion went RED (`expected 2 to be 1`); wallet-core + display tests stayed GREEN.

Non-tautology for the Hyperliquid/Sei special case holds: `getTwChainId` returns viem `hyperliquid.id`/`sei.id` for those two, but `CANONICAL_EVM_CHAIN_IDS` uses hardcoded literals (999 / 1329), so it's a viem-value-vs-literal cross-check that would still catch the 998/999 drift — not a value compared against its own source. The other 11 chains resolve through `walletCore.CoinTypeExt.chainId(getCoinType(chain))`, independent of viem entirely. `satisfies Record<EvmChain, number>` keeps the table exhaustive against the enum, so a 14th EVM chain is a compile error until its id is pinned — both assertions then cover it automatically.

## risk

Minimal — test-only, single file (`packages/sdk/tests/unit/chains/evmChainId.test.ts`, +46/-1), no production or chain-config change. It adds a real wallet-core `initWasm()` to this suite (already the established pattern in `tests/unit/chains/tron-tx.test.ts`); mocking the thing under test would have re-created the fake-green gap.

## verdict

Approve. The test pins the genuinely-signed chain-id source, is load-bearing (mutation-reddens on both a wrong wallet-core mapping and a wrong RN mirror), is non-tautological, exhaustive by construction, and strictly test-only.

---
_Reviewed by the reviewer agent (Claude primary lane; Codex lane skipped — infra). Mutation evidence gathered in an isolated worktree._

PR #1544, opened 2026-07-24
fix(cli): sign yield_enter/yield_exit deposits (bead 6rg2)

NeOMakinG:
## Problem (bead vultisig-6rg2, P0 — found via live real-fund dogfood)

Yield deposit/withdraw **never signed** via the CLI agent path. `clients/cli/src/agent/toolOutputSigning.ts` is the sole client-side signing source. `deriveToolOutputCandidate(toolName, output)` returned a signable candidate ONLY when `toolName` was in `CLI_SIGNABLE_FLAT_TOOLS` or `CLI_SIGNABLE_PREP_TOOLS`. `yield_enter` and `yield_exit` were in **neither**, so a well-formed yield tool-output produced **no candidate** → no synthetic `sign_tx` → keysign never fired. The turn ended at the backend's "Transaction prepared" with `transactions:[]` and no txHash.

## Root cause

The yield output is a **third shape**, distinct from both existing derivations:

```jsonc
{ "chain": "Polygon", "provider": "yield_xyz",
  "transactions": [ {"to","value","data","action":"approval"},
                    {"to","value","data","action":"deposit"} ] }
```

a top-level `transactions[]` array of already-built flat EVM legs — the `splitMultiTx` Pattern 2 / `build_cctp_bridge_usdc` envelope (`mcp-lunc-hop` `yield-tools.ts`). It has **no `tx_encoding`** (the EVM path deliberately omits it) and **no root-level `to`/`data`** — so the PREP path (`txArgs.tx_encoding` required) and the FLAT path (reads a single leg off the envelope root) both skip it.

## Fix

- Add `CLI_SIGNABLE_YIELD_TOOLS = {yield_enter, yield_exit}` and `buildTxReadyFromYieldOutput`, wired into `deriveToolOutputCandidate` (`source: 'yield'`).
- Map the `transactions[]` legs onto the executor's **existing** two-leg (`{approvalTxArgs, txArgs}`) / single-leg (`{tx}`) machinery — reusing `extractLeg` (faithful to/value/data lift, **never fabricates or mutates calldata**) and an EVM chain guard.
- A 2-leg approve→deposit sequence signs + broadcasts **in order** (approve receipt-confirmed before the deposit) via the executor's `signMultiLeg`, matching swap-with-approval. EVM-only — matching both `signMultiLeg`'s EVM-only constraint and yield's EVM canonical-envelope path.
- **Fail-closed**: null on error/no-op envelope, missing/non-EVM chain, absent/empty `transactions[]`, any un-liftable leg (all-or-nothing — never a partial sequence), or >2 legs. The `scan_request` the yield tool emits is **not** bypassed (L4/backend scans upstream before the CLI sees the output).

The three `source` consumers (`client.ts` callback type + fast-path gate, `session.ts` handler) are widened `'flat' | 'prep'` → `ToolOutputCandidate['source']`; `source` is log-only, never branched on.

## Tests

- `toolOutputSigning.test.ts`: yield 2-step→multi-leg, 1-step→single-leg, calldata-verbatim, and the fail-closed guards (no/empty transactions, error envelope, missing/non-EVM chain, un-liftable leg, error-leg, >2 legs) + `deriveToolOutputCandidate` yield cases.
- `toolOutputSigning.allowlist.test.ts`: pins `CLI_SIGNABLE_YIELD_TOOLS` and its disjointness from the flat/prep buckets.
- **Control-mutation verified**: emptying `CLI_SIGNABLE_YIELD_TOOLS` turns 5 derivation tests red.
- Full agent suite green (581 tests), typecheck + lint clean.

## Live on-chain proof

Rebuilt CLI + Testwallet2, real Aave v3 USDC deposit on Polygon (`polygon-usdc-aave-v3-lending`, 0.4 USDC). Verbose trace showed the exact previously-broken path now succeeding: `yield_enter → onToolOutputTx (yield candidate) → stored multi-leg envelope (approve, main) → signMultiLeg: approve confirmed, broadcasting main leg → confirmed`.

- Approval tx: `0x8d59e993aec920ca12ccec9c225b109be6c923c0b4aeb2fb54191fa3da0fd100`
- Deposit tx: [`0x9dc5cd037205b5e96b739ebcd8955dc0fe2152a45bec58e4f7f2f0a87919b5fd`](https://polygonscan.com/tx/0x9dc5cd037205b5e96b739ebcd8955dc0fe2152a45bec58e4f7f2f0a87919b5fd) — **confirmed** (independently re-verified via `tx-status`).

## Out of scope (left beaded separately)

`--session-id` resume minting a new conversation_id / losing pending-tx context is a distinct issue, not touched here.

## Summary by CodeRabbit

* **New Features**
  * Added support for signing transactions generated by yield actions.
  * Supports both single-step and approval-plus-action yield transactions.
  * Extended tool-output candidate generation to include yield-sourced results.

* **Bug Fixes**
  * Improved yield tool signing eligibility and allowlist gating.
  * Fail-safe handling for incomplete, malformed, error-marked, or non-EVM yield responses to avoid producing signable transactions.

* **Tests**
  * Expanded CLI signing tests to cover yield enter/exit behavior and regression cases.

### DIFF
--- a/clients/cli/src/agent/__tests__/toolOutputSigning.allowlist.test.ts
+++ b/clients/cli/src/agent/__tests__/toolOutputSigning.allowlist.test.ts
@@ -21,6 +21,7 @@ import { describe, expect, it } from 'vitest'
 import {
   CLI_SIGNABLE_FLAT_TOOLS,
   CLI_SIGNABLE_PREP_TOOLS,
+  CLI_SIGNABLE_YIELD_TOOLS,
   DIVERGENT_FIELD_TOOLS,
   POLYMARKET_DEPOSIT_TOOL,
   POLYMARKET_SETUP_TRADING_TOOL,
@@ -85,6 +86,19 @@ describe('CLI_SIGNABLE_FLAT_TOOLS — independent anchor vs the mcp-ts source of
 describe('CLI_SIGNABLE_PREP_TOOLS', () => {
   it('contains exactly the execute_* signer-ready prep tools', () => {
     expect([...CLI_SIGNABLE_PREP_TOOLS].sort()).toEqual(['execute_contract_call', 'execute_send', 'execute_swap'])
+  })
+})
+
+describe('CLI_SIGNABLE_YIELD_TOOLS (bead vultisig-6rg2 — yield deposit/withdraw signing gap)', () => {
+  it('contains exactly yield_enter + yield_exit', () => {
+    expect([...CLI_SIGNABLE_YIELD_TOOLS].sort()).toEqual(['yield_enter', 'yield_exit'])
+  })
+
+  it('is disjoint from the flat and prep allowlists (a tool lives in ONE bucket)', () => {
+    for (const t of CLI_SIGNABLE_YIELD_TOOLS) {
+      expect(CLI_SIGNABLE_FLAT_TOOLS.has(t)).toBe(false)
+      expect(CLI_SIGNABLE_PREP_TOOLS.has(t)).toBe(false)
+    }
   })
 })
 
--- a/clients/cli/src/agent/__tests__/toolOutputSigning.test.ts
+++ b/clients/cli/src/agent/__tests__/toolOutputSigning.test.ts
@@ -18,6 +18,7 @@ import { describe, expect, it } from 'vitest'
 
 import {
   buildTxReadyFromToolOutput,
+  buildTxReadyFromYieldOutput,
   CLI_SIGNABLE_FLAT_TOOLS,
   deriveToolOutputCandidate,
   POLYMARKET_DEPOSIT_TOOL,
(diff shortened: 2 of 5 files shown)


gomesalexandre (commented):
Reviewed with the money-path lens since this signs yield deposits and withdrawals. The substance holds up well and the shape is right. One mechanical blocker and one real test gap.

## what I verified

Mutation-verified in an isolated clone, full CI build chain reproduced:
- reverting the core fix (removing the `CLI_SIGNABLE_YIELD_TOOLS` branch from `deriveToolOutputCandidate`) turns two named tests red, including the explicit `WAS null pre-fix` regression case. Restore, green.
- fail-opening the all-or-nothing leg guard (`if (!leg) return null` to `continue`) turns `null when ANY leg is un-liftable` red. Restore, green.
- post-restore the full CLI suite is 962/962 across 73 files, typecheck exit 0.

These tests are a real gate, not e2e-only: the root `yarn test` in the Unit Tests job includes the CLI workspace, and that job is green here. Worth stating because the sdk's `test:e2e` is never invoked on a PR, so an e2e-only fix would effectively be ungated.

Fail-closed confirmed in every bad shape I could construct: error envelope, missing or unresolvable chain, empty `transactions[]`, any un-liftable leg, error-marked leg, more than 2 legs. The backend's descriptive fallback shape also nulls out. Two more backstops downstream, and the pre-sign confirm chokepoint covers the new route with deny-by-default when headless.

The part I liked most: the CLI never re-encodes anything. Amount, spender and recipient ride the backend-built calldata untouched, so there is no encode-mismatch exposure on this surface, and the confirm summary reads from the buffered signed payload rather than the server echo. The top-level `amount`/`amountUsd` echoes are correctly NOT rendered, which is exactly right - showing them un-cross-checked is the classic drift bug.

## findings

`should-fix:` **the EVM-kind guard is not independently pinned.** Deleting `getChainKind(resolved) !== 'evm'` at `toolOutputSigning.ts:395` (while keeping the resolve) leaves all 63 tests green. The `null for a non-EVM chain` test's Solana fixture carries legs shaped `{tx_encoding:'solana-tx', data:'base64blob'}` which `extractLeg` rejects anyway, so that test is actually exercising the leg-lift guard, not the EVM guard. The case it cannot see is an envelope with `chain: 'Solana'` but EVM-shaped legs (0x address, 0x calldata). Downstream layers do throw on that today rather than signing it, so this is not a live fund bug - but the guard reads as protected when nothing proves it is. A fixture with EVM-shaped legs under a non-EVM chain asserting null would close it.

`suggestion:` the confirm summary is under-informative rather than wrong. The yield envelope carries `intent`/`yieldId`/`amount` but not `action`, so `env.action` is always undefined and the prompt reads only `contract call on Polygon to 0x<pool> (+ token approval - 2 transactions)`. The user cannot tell enter from exit, see which product, or see the approve spender and amount, and there is no unlimited-approve detection at confirm time. Everything shown does come from the signed payload so nothing is drifted - this is inherited from the existing `__buildTx` summary, not introduced here. Decoding the approve leg's selector/spender/amount from the calldata in `getPendingSummary` would make it true WYSIWYS.

`suggestion:` `extractLeg` drops the `from` field the backend forwards when yield.xyz supplies it. Failing closed when `from` is present and does not match the vault's own address would cheaply close the "calldata built for another account's state" class (Aave `onBehalfOf` and friends). Similarly, asserting the first leg really is the approval (by `action`, or the `0x095ea7b3` selector) on the 2-leg path would catch a mis-ordered envelope before the summary labels the wrong leg "token approval". Order is preserved into the sequencer today so a swap merely reverts on-chain, which is fail-safe, but the label would be lying.

`q:` a latent cross-chain blind spot that is backend-owned, not yours: the envelope's top-level `chain` is the FIRST step's network, and EVM canonical steps carry no per-step network, so if yield.xyz ever returned a multi-network EVM action the CLI would sign every leg on the first chain with no way to detect it - the information is discarded before the CLI sees it. Backend comments assert single-network today. Enforcement belongs in the backend yield-tools (assert all steps share a network before canonicalizing); worth a bead in that repo rather than here.

## the blocker

Only mechanical: prettier. `Lint and Type Check` and `Test Summary` are red on three line-wrap-only diffs in `toolOutputSigning.ts` and its test file, which is what is holding `mergeStateStatus` at BLOCKED. Zero semantic content. I am pushing the `yarn format` result to unblock it rather than bouncing it back for a one-command fix - that keeps my delta purely mechanical so it does not affect the review gate.

gomesalexandre (approved):
Approving. CI is green after the formatting push, so the only thing that was blocking is resolved.

My delta on this branch is `af58bdea`, a `yarn format` run and nothing else. I checked it line by line before approving rather than trusting the tool: +9/-8 across the two files, and every changed line is the same tokens re-wrapped (the `CLI_SIGNABLE_YIELD_TOOLS` Set literal collapsed onto one line, the `yieldSingleStep` fixture object expanded, one `expect(...)` unwrapped). No identifier, value, or string literal changed. That keeps this a mechanical contribution, so approving is not me stamping my own work.

Recapping the substance from the review above so it is on the record with the approval: the fix is pinned by two named tests that go red when the `CLI_SIGNABLE_YIELD_TOOLS` branch is reverted, the all-or-nothing leg guard is pinned by its own named test, and it fails closed on every malformed envelope shape I could construct. The CLI re-encodes nothing, so amount, spender and recipient ride the backend-built calldata untouched and there is no encode-mismatch exposure here. The confirm summary reads the buffered signed payload and correctly does not render the top-level `amount`/`amountUsd` echoes.

Carrying forward, not blocking:

- the EVM-kind guard at `toolOutputSigning.ts:395` is still unpinned - deleting it leaves all 63 tests green, because the non-EVM fixture's legs are rejected by the leg-lift guard first. A fixture with EVM-shaped legs under a non-EVM chain would close it. Not a live fund bug, since downstream throws rather than signs.
- the confirm summary cannot distinguish enter from exit and does not surface the approve spender or amount. Inherited from the existing summary, worth decoding the approve calldata there for true WYSIWYS.
- the multi-network EVM blind spot is backend-owned; enforcement belongs in yield-tools asserting all steps share a network before canonicalizing.

Nice work on the calldata-verbatim approach and the on-chain receipt in the body.

PR #1568, opened 2026-07-26
feat(ripple): add XRPL issued-currency (custom token) support

### LINKED ISSUE #1567: [ADD] XRPL issued-currency (custom token) support for the add-custom-token flow
**Is your feature request related to a problem? Please describe.**

XRPL "custom tokens" are issued currencies (trust-line tokens / IOUs) such as SOLO or RLUSD, identified by a `(currency, issuer)` pair rather than a single contract address. Client apps that consume the SDK (e.g. [vultisig-windows](https://github.com/vultisig/vultisig-windows)) let users add a custom token for EVM, Solana, Tron, Cosmos, etc., but can't offer the same for XRP because the SDK has no support for XRPL issued currencies in the pieces a manual "add custom token" flow depends on:

- **Token-metadata resolution** — there is no way to resolve a token's metadata (`ticker`, `decimals`, …) for an XRPL issued currency from its id, so a client can't display a token the user enters.
- **Custom-token-id validation / normalisation** — an XRPL `currency.issuer` id isn't recognised or canonicalised. Non-standard currencies also have two spellings: the human ticker users see on explorers (e.g. `SOLO.rsoLo2S1kiGeCcn6hCUXVrCpGMWLrRrLZz` on xrpscan) and the 160-bit / 40-hex on-ledger currency code. The two must resolve to the same identity.

The result is that only XRPL tokens the vault already holds (surfaced by ledger auto-discovery) or ones curated in the SDK (currently just RLUSD) can appear — a user can't add an arbitrary issued currency the way they can on other chains.

**Describe alternatives you've considered**

- **Ledger auto-discovery** (already exists) surfaces trust-line tokens the account already holds, but can't add/track a token the user specifies on demand, or one they don't yet hold.
- **Expanding the curated known-tokens list** — doesn't scale and still can't cover arbitrary user-specified tokens.

**Additional context**

- XRPL issued currencies are identified by `(currency, issuer)`; non-standard currency codes have both a human ticker and an on-ledger (40-hex) form, and explorers reference them by ticker.
- Parity target: the existing custom-token support for EVM / Tron / Solana / Cosmos.
- Unblocks the corresponding client work in vultisig-windows: vultisig/vultisig-windows#4458

Ehsan-saradar:
## What

Adds XRPL issued-currency ("custom token") support so client apps can offer an add-custom-token flow for XRP, matching EVM / Tron / Solana / etc.

- **Token-metadata resolver** (`getRippleTokenMetadata`) + Ripple added to `chainsWithTokenMetadataDiscovery`. XRPL issued currencies expose no on-ledger metadata call, so nothing is fetched: decimals are fixed (`rippleIssuedCurrencyDecimals`) and the ticker is derived from the currency code. Curated tokens (currently RLUSD) get their `logo` / `priceProviderId` merged in; an arbitrary token gets neither, so it can't borrow a known token's identity — two issuers can share a ticker on XRPL, so the issuer distinguishes them.
- **Human-ticker acceptance** in `isValidTokenId`: a token may be entered by its human ticker (`SOLO.rsoLo…`, as shown on explorers like xrpscan) as well as its on-ledger (3-char or 40-hex) currency code.
- **`normalizeTokenId`**: canonicalises a pasted id to the on-ledger form so a manually added token dedupes against the same token surfaced by ledger auto-discovery (which always uses the on-ledger code).

## Why

XRPL "custom tokens" are issued currencies identified by a `(currency, issuer)` pair rather than a single contract address. Without a metadata resolver and id handling, only curated (RLUSD) or already-held (auto-discovered) tokens could appear — a user couldn't add an arbitrary issued currency the way they can on other chains.

## Tests

- `isValidTokenId.test.ts` — the previous "rejects an unencoded ticker" expectation is updated to the intended behaviour, plus new coverage for `normalizeTokenId` (canonicalisation matches auto-discovery; standard/encoded ids unchanged; non-Ripple untouched) and edge cases (over-long name, missing separator).
- New `resolvers/ripple.test.ts` — curated token enriched with logo/price; arbitrary token gets ticker/decimals only.
- `yarn build:shared`, shared-exports check, eslint, and the unit tests above pass locally.

## Related

- Closes #1567
- Prerequisite for the client work in vultisig-windows — that branch imports `normalizeTokenId` and needs a released `@vultisig/core-chain` bump: vultisig/vultisig-windows#4459

## Summary by CodeRabbit

- **New Features**
  - Added Ripple (XRPL) issued/custom token metadata discovery, including curated enrichment for known tokens and derived ticker/decimals for others.
- **Bug Fixes**
  - Improved Ripple token ID validation and canonicalization, accepting human-form inputs and rejecting malformed values.
- **Changes**
  - Token add/remove now normalizes Ripple token IDs (and contract addresses when present) to a canonical stored form to prevent duplicates and ensure consistent lookups.
- **Tests**
  - Added unit tests for Ripple metadata resolution and BalanceService add/remove normalization.
- **Documentation**
  - Added a release changeset for Ripple issued-currency/custom token support.


gomesalexandre (changes requested):
blocking: `normalizeTokenId()` never gets wired into any storage or dedupe path, so the headline behavior in this PR does not actually happen yet.

grounding from the current head:
- this PR adds `normalizeTokenId()` in `packages/core/chain/utils/isValidTokenId.ts`, but the only current-head references are that definition plus its tests. there are no production callsites using it.
- token persistence/dedupe still compares raw ids only: `packages/sdk/src/vault/services/BalanceService.ts` adds a token when `!allTokens[chain].find(t => t.id === token.id)`.
- manual add flows still pass through the raw user-entered ripple id unchanged, e.g. `examples/shared/src/components/token/AddTokenModal.tsx` builds `{ id: contractAddress, contractAddress }` from the pasted value, and the CLI path still stores / dedupes raw token ids too.

that means `RLUSD.<issuer>` and the canonical on-ledger `524C555344....<issuer>` remain distinct stored ids, so a manual add still will not dedupe against ledger discovery on this branch.

please either normalize the ripple token id before it enters persisted token state, or normalize both sides of the equality/dedupe checks, then add a regression that proves a manual `RLUSD.<issuer>` add and an auto-discovered canonical id collapse to one stored token.

gomesalexandre (approved):
Clearing my block from 07-26. The finding is resolved and I verified it rather than taking the change at face value.

## what I blocked on, and what changed

My original concern was that `normalizeTokenId()` was never wired into any storage or dedupe path, so the headline behaviour did not actually happen. That is the dead-code shape: a helper that exists, is tested in isolation, and is never reached by the code it was meant to fix.

It is on the live path now. Traced from `VaultBase.addToken`/`removeToken` through to `BalanceService`, with normalization applied before the token enters the stored set in both directions.

I insisted on a mutation at the **storage site** specifically, because a passing helper test was exactly what made the original gap invisible. Stripping the normalization back out fails 3 named tests:

- `collapses a manual human-ticker add and an auto-discovered canonical id into one token` on a length of 2 where 1 was expected
- `persists a manually added human-ticker token under its canonical id`, returning the human ticker instead of the canonical form
- `removes a canonical-stored token when asked by its human-ticker id`, leaving the token behind

Restore, 8/8. The rollback tests correctly stay green under that mutation, since they pin a different hunk.

## the questions I asked and the answers that held

**Write and read symmetry:** auto-discovery builds ids through the same canonicalization the normalizer delegates to, so a manual add and a ledger-discovered entry converge on one key rather than drifting apart.

**Idempotence:** confirmed structurally and by an explicit test that an already-encoded id passes through unchanged.

**Legacy stored tokens.** This was the one I expected to be missed and it is genuinely not reachable: before this PR there was no functional custom-token path for this chain at all, so there is no shipped population of un-normalized ids to migrate. Worth stating plainly because it is the difference between "handled" and "cannot arise", and the second is a stronger answer.

**Two issuers of the same currency code:** safe. The re-encoding only touches the currency segment; the issuer passes through verbatim and the dedupe compare is case-sensitive on the full id, so two issuers of the same ticker never merge. That collision would have been the worst outcome available here.

**Rollback on save failure:** confirmed by code and by two dedicated tests.

## notes, none blocking

`suggestion:` normalization runs at write time with no backfill. Not reachable today for the reason above, but if the canonical form ever changes again post-ship, existing rows would silently stop deduping against new writes. Worth a comment or a tracked follow-up so the next person does not have to rediscover that reasoning.

`suggestion:` there is no direct dedupe-level test asserting that issuer A's token and issuer B's token with the same currency code stay distinct. The mechanism is correct by inspection and the metadata resolver has a near analog, but a direct assertion would close the loop rather than leaving it to inference.

`praise:` the `addToken` atomicity rewrite goes beyond what was asked, and it is what made the rollback tests meaningful rather than vacuously true. Applying the add as a fresh record leaves the pre-add state intact as a real rollback target, which is the difference between a test that passes and a test that proves something.

PR #1581, opened 2026-07-27
docs: document CowSwap client support matrix

### LINKED ISSUE #1446: Verify CowSwap provider: client coverage + whether orders execute on Windows
## What we need verified

CowSwap is registered as a general swap provider on `main`, but it is reportedly **not enabled across all wallet clients**, and **may not be functioning on Windows**. Before we make any public claim about our provider set, we need the real per-client picture.

## What the code says

- CowSwap is in `generalSwapProviders` (`packages/core/chain/swap/general/GeneralSwapProvider.ts`)
- Registered as a fetcher in `findSwapQuote.ts:709`
- Scope: same-chain EVM ERC-20 only
- Opt-out-able via `excludeProviders` — intended for consumers that lack EIP-712 order signing
- Not present in the Jan 2026 provider set (`['1inch','li.fi','kyber']`), so it is a recent addition

That `excludeProviders` escape hatch is the likely explanation: the SDK registers CowSwap unconditionally, and each client decides. But nothing documents which clients opt out, so the SDK's registered set and the shipped set have silently diverged.

## Questions to answer

1. **Per-client matrix.** Which of iOS / Android / Windows / extension pass CowSwap in `excludeProviders`? Please produce the actual matrix, not the intended one.
2. **Windows execution path.** Does a CowSwap route on Windows produce a quote that actually *executes*? CowSwap needs EIP-712 order signing — if Windows can quote but not sign, users can be routed to a provider that then fails at signing. Confirm end-to-end with a real same-chain EVM ERC-20 swap.
3. **Failure mode if unsupported.** When a client lacking EIP-712 signing receives a winning CowSwap quote, does it fail closed (route excluded before ranking) or fail late (user hits an error at signing)? Late failure is the bad case — the route wins the ranking, then dies.
4. **Should the default flip?** If most clients exclude it, consider making CowSwap opt-*in* rather than opt-out, so the SDK default matches what ships.

## Why now

We are preparing public content describing which providers Vultisig routes across. We have deliberately left CowSwap out pending this verification — we are not going to publish a provider list the shipped wallets don't match.

## Acceptance

- Documented per-client provider matrix (which clients include/exclude CowSwap), committed to the repo rather than left in a thread
- Confirmed end-to-end CowSwap swap on Windows, or a clear statement that it is unsupported there
- If unsupported anywhere: route excluded before ranking on those clients, so it can never win a quote it cannot sign

rcoderdev:
Closes #1446

## Summary by CodeRabbit

* **Documentation**
  * Added documentation describing CowSwap client support, eligibility, and exclusion behavior.
  * Documented the fail-closed signing flow for unsupported clients.
  * Added guidance for excluding CowSwap before quote searches when execution support is unavailable.
  * Added the new documentation page to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ await vault.sign(payload, {
 
 - [SDK API Reference & Examples](packages/sdk/README.md) — Full API docs with code examples
 - [SDK Users Guide](docs/SDK-USERS-GUIDE.md) — Detailed usage guide with advanced topics
+- [CowSwap client support](docs/cowswap-client-support.md) — First-party support matrix and caller capability contract
 - [CLI Documentation](clients/cli/README.md) — Command-line interface
 - [docs.vultisig.com](https://docs.vultisig.com/developer-docs/vultisig-sdk/) — Online documentation
 
--- a/docs/cowswap-client-support.md
+++ b/docs/cowswap-client-support.md
@@ -1,0 +1,53 @@
+# CowSwap client support
+
+CowSwap is an available SDK quote provider, not a promise that every Vultisig
+client can execute CowSwap orders. Clients must exclude it unless their complete
+transaction path can approve the sell token when required, produce the CowSwap
+EIP-712 order signature, submit the order, and track its off-chain lifecycle.
+
+## First-party client matrix
+
+| Client            | CowSwap quote eligibility | Current support | Reason                                                                                                                                                           |
+| ----------------- | ------------------------- | --------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| iOS               | Not integrated            | Unsupported     | The native iOS swap stack does not call this SDK quote resolver or implement the CowSwap order-signing flow.                                                     |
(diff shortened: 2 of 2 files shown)


gomesalexandre (approved):
docs-only pass on current head `c3b6078f914eeb3516fc1b44532796d7506eeba3`.

spot-checked the factual claims against the current sdk + sibling client repos:
- `GeneralSwapProvider.ts` still registers CowSwap by default
- `findSwapQuote.ts` still accepts `excludeProviders` and filters excluded providers before fetch/ranking, so an excluded provider cannot win late
- windows/extension do centralize `['CowSwap']` in the shared quote config and thread it through the quote entry points
- android only recognizes the settlement contract for verification, not full provider integration
- iOS has no CowSwap integration hits

so the support matrix matches the current code and closes the acceptance criteria from #1446. the only thing here that isn't code-verifiable is the extension e2e timeout anecdote, and the doc already presents that as observed QA evidence rather than as a code guarantee.

lgtm

PR #1582, opened 2026-07-27
fix(swap): collect Jupiter affiliate fee in public tool

### LINKED ISSUE #942: fix(swap): Jupiter affiliate fee is OFF on the public tool path (silent revenue loss)
## Evidence
`packages/sdk/src/tools/swap/jupiter.ts:77` — `export const JUPITER_AFFILIATE_FEE_ATAS: Readonly<Record<string,string>> = {}` (empty), so `resolveJupiterFeeAccount` returns `null` for every mint. `buildJupiterSwapTx` (public API via `src/index.ts`) therefore collects zero fee. Meanwhile the path used by `vault.swap()` (`core/chain/swap/general/jupiter/api/jupiterFeeAta.ts` — `deriveJupiterFeeAccount`+`prependJupiterFeeAta`) derives the fee ATA on the fly and needs no pre-created treasury account.

## Impact
Every caller using the public same-chain-Solana-swap tool directly (the documented code-as-action path for agents) collects **zero** affiliate revenue on every Solana↔Solana swap — even though `JUPITER_AFFILIATE_FEE_OWNER` and `JUPITER_PLATFORM_FEE_BPS = 50` are fully defined.

## Fix sketch
Replace the static `JUPITER_AFFILIATE_FEE_ATAS` lookup with a call to `deriveJupiterFeeAccount`/`prependJupiterFeeAta`, mirroring `getJupiterSwapQuote.ts`.

---

rcoderdev:
Closes #942

Live Vultisig Jupiter verification at the exact branch commit exercised zero-fee and positive-fee SOL/USDC quotes through the production SDK builder. The zero-fee path omitted feeAccount, the positive-fee path used the derived output-mint ATA, and both results deserialized as unsigned versioned transactions with all-zero signatures. No signing, broadcast, or funds were involved.

## Summary by CodeRabbit

- **New Features**
  - Jupiter swaps now dynamically derive and, when applicable, set up the affiliate fee account for the selected output token.

- **Bug Fixes**
  - When Jupiter floors the platform fee to zero (or affiliate fees aren’t applicable), swap requests now consistently omit related fee fields to avoid quote/swap mismatches.
  - Safer swap transaction construction now ensures fee-account setup is included only when a non-zero fee is actually charged.

- **Tests**
  - Updated and expanded Jupiter quote/swap unit tests to verify fee omission/inclusion behavior and updated transaction expectations.


gomesalexandre (commented):
Nice fix - this closes a real revenue bug (issue #942: the public `buildJupiterSwapTx` collected zero affiliate fee because `JUPITER_AFFILIATE_FEE_ATAS` was an empty map returning null for every mint). Reviewing it with the money-path lens; it holds up on correctness. One availability regression I would like addressed, and one bigger-picture note that matters for whether this actually moves the needle.

## what I verified

Mutation-verified in an isolated clone, three ways, all pinned and restore-green:
- revert the resolver to the empty-map null -> 6 tests fail, headline `includes platformFeeBps + feeAccount and prepends the fee ATA for the default affiliate path` (`affiliateFeeApplied` false).
- send the quote verbatim instead of the zero-fee-normalized one -> the two floored-fee tests fail on `to not have property "platformFee"`.
- same revert in core `getJupiterSwapQuote.ts` -> 2 core tests fail. Restore -> 21/21 + 14/14 green.

Fail-closed confirmed: fee-ATA derive failure throws rather than falling back to a wrong ATA, unresolvable LUTs make decompile throw (no partial tx), and the final post-prepend tx is validated against the SOL-01 allowlist before anything signable is returned - stricter than before. Fee recipient is pinned (`jupiterFeeOwnerAddress` constant, ATA derived locally, nothing from the quote can swap it out) and bps is the `JUPITER_PLATFORM_FEE_BPS = 50` constant. Reusing the battle-tested `vault.swap()` mechanism instead of a third Jupiter impl is exactly the right shape.

## should-fix: availability regression on the swap path

`jupiter.ts:253` awaits `deriveJupiterFeeAccount` BEFORE `/quote`, and it throws on any RPC failure (`jupiterFeeAta.ts:24-26`). So a Solana RPC proxy hiccup now fails the whole Jupiter swap, where main built it fee-OFF and the user's swap succeeded. Fee bookkeeping should never block a user's swap. A `try/catch` around the resolve that degrades to `feeAccountResult = null` (fee-OFF) with a warn keeps the swap available. It is safe-direction as-is (fails closed, never mis-routes), so not a hard block - but it is a real degradation vs main on a swap path, and the fix is tiny, so worth doing before this lands.

## the note that matters most: this does not fix prod yet

The live agent backend does NOT consume this SDK tool - abts vendors its own copy with the SAME empty-map fee-OFF bug (`agent-backend-ts src/mastra/tools/mcp/tools/swap/jupiterAffiliateConfig.ts:98,113`). So until that copy is ported too, the production agent still collects zero Jupiter affiliate fee regardless of this PR. Worth a companion abts PR (or a tracking issue) so the revenue fix actually reaches prod.

## follow-ups (non-blocking)

- itemize the fee in `JupiterSwapResult` - it currently carries only `affiliateFeeApplied: boolean`, but the core path surfaces `swapFee.amount`. Now that the fee is live on the tool path, expose `quote.platformFee.amount` + bps so an agent can render "affiliate fee: X (50 bps)" rather than a bare boolean.
- cross-check the echo: assert `quote.platformFee.feeBps === JUPITER_PLATFORM_FEE_BPS` and treat a mismatch as fee-OFF or refuse - closes the accounting-drift window on a hostile quote.
- drop the dead `JUPITER_AFFILIATE_FEE_ATAS` export rather than deprecating an empty frozen map.

## mergeable

`mergeStateStatus: BEHIND` - needs a main update before it can merge (main moved under it). CI at head is green across the board. Rebase, address the availability regression if you agree, and it is good to go.

gomesalexandre (approved):
looks good now. the current head fixes the availability regression by degrading fee-account lookup failures to fee-off, keeps `platformFeeBps` and `feeAccount` in sync, and adds regression coverage for that path. ci is green on this head.

still behind `main`, so refresh the branch before merge.

PR #1583, opened 2026-07-27
fix(sui): migrate off JSON-RPC to gRPC (+ GraphQL on React Native)

Toby1009:
## Why now

Sui is retiring JSON-RPC. Shutdown on Foundation mainnet full nodes began the week of **2026-07-27**; full decommission — code removal, after which **no provider can serve it** — lands **mid-October 2026**. gRPC and GraphQL RPC are the supported replacements.

**To be precise about urgency: nothing is broken today.** I checked both endpoints on 2026-07-27 and both still answer `suix_getBalance`:

```
sui-rpc.publicnode.com     → {"result":{"totalBalance":"2227801536832",...}}
fullnode.mainnet.sui.io    → {"result":{"totalBalance":"2227801536832",...}}
```

So this is a scheduled migration ahead of a hard deadline, not a hotfix. I'd rather state that plainly than have the PR read as a fire drill.

I searched open and closed issues and PRs for `grpc` / `graphql` / `json-rpc` / `jsonrpc` / `deprecat` and found nothing covering the migration — the Sui work (#1132, #1275, #1338, #1398, #1413, #1430, #1467) is all coin selection, gas convergence and broadcast status, all still on JSON-RPC. Apologies if I missed a plan tracked elsewhere; happy to close this if it's already in hand.

## The part I think matters most

A naive port of the call sites would leave **two silent breaks** behind. These are the reason I opened a PR rather than just an issue — both are non-obvious and both are in the fund path.

**1. `clients/cli` — the permanent-vs-retryable gate goes dead, and it is not fail-safe.**

The Sui branch gates on the JSON-RPC numeric code `-32002`. On gRPC that can never match, so *every* permanent rejection is reclassified as retryable and the CLI re-broadcasts bytes that can never land. Captured live:

```
### grpc executeTransaction / bad signature
  ctor      RpcError
  code      "INVALID_ARGUMENT"      <- a string, not -32002
  message   "invalid%20signature:%20error%20converting%20from%20protobuf:..."
  decoded   "invalid signature: error converting from protobuf:..."
```

Two shape changes at once: `code` is the grpc-status **name**, and the message is **percent-encoded** (grpc-web escapes the `grpc-message` trailer), so a regex containing literal spaces never matches even if the gate passed.

**2. `transientRetry` — 5xx coverage disappears.**

A grpc-web response is **HTTP 200** with the real status in the trailer, so the existing 5xx branch never sees a busy or restarting node. Under JSON-RPC that arrived as an `HttpResponseError` and was retried. Fixed by retrying `UNAVAILABLE` / `DEADLINE_EXCEEDED` / `RESOURCE_EXHAUSTED` (the three gRPC itself designates safe) and decoding the message before pattern-matching. `INVALID_ARGUMENT`, `NOT_FOUND`, `FAILED_PRECONDITION`, `ABORTED` and `INTERNAL` deliberately stay out.

The `DeliverTxFailedError` marker still short-circuits before any code or message test, so an on-chain MoveAbort is not retried even when its chain-controlled text says "aborted" *or* the status looks transient — sdk#1398 re-expressed in the new transport's vocabulary, with a test for exactly that pair.

## A decision that is yours, not mine

**The endpoint has to change, but I do not think I should be the one choosing the replacement.**

You deliberately use `sui-rpc.publicnode.com`. publicnode *does* run a Sui gRPC endpoint — but it speaks native gRPC only and rejects both grpc-web wire formats by name:

```
sui-grpc.publicnode.com    [grpc-web binary] → invalid content-type "application/grpc-web+proto"
sui-grpc.publicnode.com    [grpc-web text]   → invalid content-type "application/grpc-web-text"
sui-graphql.publicnode.com                   → 404
sui-rpc.publicnode.com     [any gRPC format] → Bad Request
```

Browsers cannot speak native gRPC — that is the entire reason grpc-web exists — so the browser, chrome-extension and React Native bundles physically cannot reach it. Node and Electron could, via `@grpc/grpc-js` plus a native transport, but that buys a three-way transport split across platforms.

So a grpc-web-capable endpoint is required. I used `fullnode.mainnet.sui.io` because it works, **not because it is the right production choice** — it is the Foundation public node, rate-limited and not positioned for production load, and avoiding exactly that is plausibly why publicnode was picked in the first place. The same caveat applies to `graphql.mainnet.sui.io` on RN.

A commercial grpc-web provider, or proxying through `api.vultisig.com`, is probably the real answer. **This is one constant in `chains/sui/config.ts`** — none of the API mapping, the second-order fixes, or the tests depend on it. Tell me what you want and I'll change the line.

### Core client rail → gRPC

| area | before (JSON-RPC) | after (unified client) |
|---|---|---|
| balance | `getBalance().totalBalance` | `getBalance().balance.balance` |
| metadata | `getCoinMetadata()` | `getCoinMetadata().coinMetadata` |
| coin objects | `getAllCoins({owner, cursor})` | `listCoins({owner, coinType, cursor})` |
| gas price | `getReferenceGasPrice(): bigint` | `getReferenceGasPrice().referenceGasPrice` |
| dry run | `dryRunTransactionBlock({transactionBlock})` | `simulateTransaction({transaction: bytes})` |
| status | `getTransactionBlock({digest, options})` | `getTransaction({digest, include})` |
| broadcast | `executeTransactionBlock({transactionBlock, signature[]})` | `executeTransaction({transaction: bytes, signatures[]})` |

Non-obvious details the live probes surfaced, each pinned by a test:

- **`effects.transactionDigest`, not `digest`.** A simulation leaves the top-level `digest` undefined. `getSuiTxHash` feeds broadcast hash-verification, so reading the wrong field returns `undefined` silently.
- **`iconUrl` is `''`, not `null`.** Mainnet SUI itself returns an empty string, so `?? undefined` leaks `''` through as a logo URL.
- **Status is a union**: `{ $kind: 'Transaction' | 'FailedTransaction' }` *plus* `{ success, error }`. The sdk#1398 fail-closed rule is re-expressed against **both** — a `FailedTransaction` arm claiming `success: true` is still an error.
- **`listCoins` is scoped to one coin type** where `getAllCoins` returned every type, so the keysign resolver fetches exactly the two sets a payload needs (native SUI for gas, plus the sent type) instead of sweeping the wallet.
- **Coin objects report the wrapper type** `0x2::coin::Coin<T>` and `objectId`, where `CoinStruct` gave the inner `coinType` and `coinObjectId`. `listAllSuiCoins` unwraps it, so coin selection, the `SuiCoin` protobuf and every downstream comparison are unchanged. (Checked against `tests/fixtures/mobile/sui.json` — the mobile golden carries the same five fields and no `previous_transaction`, so payload parity holds.)

Pagination and the stuck-cursor bound move out of the keysign resolver into `listAllSuiCoins`, still failing closed rather than under-funding a send.

### React Native → GraphQL RPC

`SuiGrpcClient` speaks grpc-web through `GrpcWebFetchTransport`, which reads `Response.body` as a `ReadableStream`. Hermes' XHR-backed `fetch` exposes no `Response.body`, so **every grpc-web call throws `missing response body` on RN**. `SuiGraphQLClient` is a plain JSON POST and implements the *same* `SuiClientTypes.TransportMethods`, so the RN override swaps transport with **zero callsite changes**. The existing lazy-Proxy (for the Hermes `Intl.PluralRules` crash) is kept as-is.

This asymmetry is a real architectural choice and I'm happy to be overruled on it.

### Dependency-free balance tools

`tools/balance/otherBalance.ts` hand-rolled `suix_getBalance` / `suix_getAllBalances` POSTs. These now POST Sui GraphQL through the same `fetchJson` helper, so the folder keeps its zero-client-dependency design. `suix_getAllBalances` returned everything at once but the GraphQL `balances` connection **paginates**, so `getSuiAllBalances` follows the cursor and returns `tokens_unavailable` on a stuck cursor or page-cap overrun rather than silently reporting a truncated portfolio — consistent with the fund-visibility rule already stated in that file.

## Verification

Endpoints and every response shape asserted in the tests were **verified against mainnet before the code was written** — reference gas price, coin metadata, balance, paginated `listCoins`, a real built-and-simulated PTB, `getTransaction` on an unknown digest, `executeTransaction` with a bad signature, and all four GraphQL queries including the bad-address error path.

Full `check:ci` locally, all green: `check:shared-exports`, `build:all`, `cli:build`, `cli:smoke`, `mcp:build`, `mcp:smoke`, `changeset:check-sdk-bundle`, `quality:contracts`, `lint`, `typecheck`, `knip`, `format:check`, `quality:architecture` (1940 modules, 0 violations), `quality:duplication` (0 clones), `quality:complexity`, `quality:secrets`, `quality:audit`, `quality:docs:markdown`.

```
@vultisig/sdk     203 files, 3036 tests   passed
@vultisig/rujira   24 files,  550 tests   passed (4 skipped)
core (chain+mpc)  227 files, 2218 tests   passed (1 skipped)
@vultisig/cli      73 files,  966 tests   passed
@vultisig/mcp       3 files,   35 tests   passed
```

Bundle check — the migration reaches every shipped artifact, and `sui-rpc.publicnode.com` appears **0 times in all six bundles** (was 4):

| bundle | client | endpoint |
|---|---|---|
| `index.node.esm.js` / `index.node.cjs` | `SuiGrpcClient` | `fullnode.mainnet.sui.io` |
| `index.browser.js` | `SuiGrpcClient` + `GrpcWebFetchTransport` | `fullnode.mainnet.sui.io` |
| `index.electron-main.cjs` | `SuiGrpcClient` | `fullnode.mainnet.sui.io` |
| `index.chrome-extension.js` | `SuiGrpcClient` | `fullnode.mainnet.sui.io` |
| `index.react-native.js` | `@mysten/sui/graphql` only | `graphql.mainnet.sui.io` |

CI on my fork is green on the full change before opening this — all four jobs on the rebased head (`1f32920a`, rebased onto `d1aae2d2`): [run 30248266821](https://github.com/Toby1009/vultisig-sdk/actions/runs/30248266821) (Unit Tests 23m, Lint and Type Check 13m, Integration Tests 5m, Test Summary).

## Breaking

`assertSuiTxSucceeded` (exported from `@vultisig/core-chain/tx/broadcast/resolvers/sui`) now takes the unified client's transaction result (`{ $kind, Transaction | FailedTransaction }`) instead of a JSON-RPC effects object. Changeset: `minor` for `core-chain` / `core-mpc` / `sdk`, `patch` for `cli`.

## Found but deliberately NOT changed

Neither is caused by this migration; folding them in would widen it past a transport swap.

**1. `getSuiTxHash` spends a network round trip for a digest that is derivable locally.** It simulates purely to read `effects.transactionDigest` — which cannot work in the one case its caller cares about, since `verifyBroadcastByHash` runs after a peer won the race, when the input objects are already spent. Pre-existing (`dryRunTransactionBlock` behaved identically), but the new SDK makes the fix trivial and I verified it:

```
local digest : 6MtGKQJxQWbqGn8Zc3Wh3SK83iR4Hine6V325gW8Yukp   <- TransactionDataBuilder.getDigestFromBytes(bytes)
node  digest : 6MtGKQJxQWbqGn8Zc3Wh3SK83iR4Hine6V325gW8Yukp   <- effects.transactionDigest, mainnet
match        : true
```

**2. `selectSuiPayloadCoins` emits the same coin object twice when a payload names SUI as its token id.** `isNativeToken` is `!coin.id`, so `coin.id === '0x2::sui::SUI'` takes the token branch and the object is selected once as the "token" and again as a gas candidate. Not reachable from `getKeysignCoin` in practice, and it fails loudly at signing rather than mis-sending. Documented in a test comment.

## On scope

35 files is a lot for one PR. It is one mechanical change (the transport) plus the two consequences that would otherwise be silent, and I kept them together because the second-order fixes are meaningless without the first. Happy to split it — transport / CLI classifier / transient retry / balance tools are all independently landable — or to close this and hand the analysis over if you'd rather implement it internally.

## Summary by CodeRabbit

* **New Features**
  * Migrated Sui transaction reads, simulations, broadcasts, and status checks to modern gRPC/GraphQL services.
  * Added complete Sui coin pagination with safeguards against incomplete or looping results.
  * Added shared transaction-result handling for success and failure states.
  * Sui balance portfolios now report `tokens_unavailable` when data cannot be retrieved reliably.

* **Bug Fixes**
  * Improved retry and permanent-error detection, including encoded transport messages.
  * Strengthened transaction success validation and gas estimation.
  * Improved handling of missing metadata, balances, and transaction details.

* **Breaking Changes**
  * Updated the transaction-success assertion API to accept the new transaction result format.

### DIFF
--- a/.changeset/sui-grpc-graphql-migration.md
+++ b/.changeset/sui-grpc-graphql-migration.md
@@ -1,0 +1,29 @@
+---
+'@vultisig/core-chain': minor
+'@vultisig/core-mpc': minor
+'@vultisig/sdk': minor
+---
+
+Move every Sui read, simulation and broadcast off JSON-RPC.
+
+Sui disabled JSON-RPC on Foundation mainnet full nodes the week of 2026-07-27
+and decommissions it entirely by mid-October 2026, so the previous
+`SuiJsonRpcClient` against `sui-rpc.publicnode.com` was on a dead rail.
+
+- `getSuiClient()` now returns a `SuiGrpcClient` pointed at
+  `https://fullnode.mainnet.sui.io:443` (gRPC-web over HTTPS).
+- React Native uses `SuiGraphQLClient` against
+  `https://graphql.mainnet.sui.io/graphql` instead: grpc-web needs
+  `Response.body` streaming, which Hermes' fetch does not provide. Both clients
+  implement the same unified transport interface, so callsites are identical.
+- Balance, coin metadata, coin listing, tx hash, tx status, broadcast and
+  keysign gas refinement moved to `getBalance` / `getCoinMetadata` / `listCoins`
+  / `simulateTransaction` / `getTransaction` / `executeTransaction`.
+- The dependency-free `@vultisig/sdk` balance tools (`getSuiBalance`,
+  `getSuiTokenBalance`, `getSuiAllBalances`) now POST Sui GraphQL and follow the
+  paginated `balances` connection to completion, returning `tokens_unavailable`
+  rather than a silently truncated portfolio.
+
+Breaking for direct consumers: `assertSuiTxSucceeded` now takes the unified
+client's transaction result (`{ $kind, Transaction | FailedTransaction }`)
+instead of a JSON-RPC effects object.
--- a/.changeset/sui-grpc-graphql-migration.md
+++ b/.changeset/sui-grpc-graphql-migration.md
@@ -7,9 +7,11 @@
 
 Move every Sui read, simulation and broadcast off JSON-RPC.
 
-Sui disabled JSON-RPC on Foundation mainnet full nodes the week of 2026-07-27
-and decommissions it entirely by mid-October 2026, so the previous
-`SuiJsonRpcClient` against `sui-rpc.publicnode.com` was on a dead rail.
+Sui is retiring JSON-RPC: shutdown on Foundation mainnet full nodes began the
+week of 2026-07-27 and full decommission (code removal) lands mid-October 2026,
+after which no provider can serve it. This is a scheduled migration ahead of
+that date, NOT a fix for a live outage — as of 2026-07-27 both
+`sui-rpc.publicnode.com` and `fullnode.mainnet.sui.io` still answer JSON-RPC.
 
 - `getSuiClient()` now returns a `SuiGrpcClient` pointed at
   `https://fullnode.mainnet.sui.io:443` (gRPC-web over HTTPS).
--- a/.changeset/sui-grpc-graphql-migration.md
+++ b/.changeset/sui-grpc-graphql-migration.md
@@ -25,11 +25,17 @@ and decommissions it entirely by mid-October 2026, so the previous
   paginated `balances` connection to completion, returning `tokens_unavailable`
   rather than a silently truncated portfolio.
 
-The CLI's permanent-vs-retryable broadcast classifier follows the transport:
-a Sui rejection now carries the grpc-status name (`INVALID_ARGUMENT`) instead
-of the numeric JSON-RPC `-32002`, and grpc-web percent-encodes the message
-trailer. Left unchanged, that gate would have gone dead and every permanent
-Sui rejection would have been re-broadcast as if transient.
+Broadcast-error classification follows the transport. A gRPC failure carries the
+grpc-status NAME in `code` (not a JSON-RPC number) and a percent-encoded message,
+so both classifiers were re-pointed:
+
+- `isTransientBroadcastError` retries `UNAVAILABLE` / `DEADLINE_EXCEEDED` /
+  `RESOURCE_EXHAUSTED` and decodes the message before pattern-matching. A
+  grpc-web response is HTTP 200 with the real status in the trailer, so the
+  existing 5xx branch never saw a busy or restarting node.
+- The CLI's permanent-vs-retryable gate matches `INVALID_ARGUMENT` instead of
+  the numeric `-32002`. Left unchanged, that gate would have gone dead and every
+  permanent Sui rejection would have been re-broadcast as if transient.
 
 Breaking for direct consumers: `assertSuiTxSucceeded` now takes the unified
 client's transaction result (`{ $kind, Transaction | FailedTransaction }`)
--- a/.changeset/sui-grpc-graphql-migration.md
+++ b/.changeset/sui-grpc-graphql-migration.md
@@ -2,6 +2,7 @@
 '@vultisig/core-chain': minor
 '@vultisig/core-mpc': minor
 '@vultisig/sdk': minor
+'@vultisig/cli': patch
 ---
 
 Move every Sui read, simulation and broadcast off JSON-RPC.
@@ -23,6 +24,12 @@ and decommissions it entirely by mid-October 2026, so the previous
   `getSuiTokenBalance`, `getSuiAllBalances`) now POST Sui GraphQL and follow the
   paginated `balances` connection to completion, returning `tokens_unavailable`
   rather than a silently truncated portfolio.
+
+The CLI's permanent-vs-retryable broadcast classifier follows the transport:
+a Sui rejection now carries the grpc-status name (`INVALID_ARGUMENT`) instead
+of the numeric JSON-RPC `-32002`, and grpc-web percent-encodes the message
+trailer. Left unchanged, that gate would have gone dead and every permanent
+Sui rejection would have been re-broadcast as if transient.
 
 Breaking for direct consumers: `assertSuiTxSucceeded` now takes the unified
 client's transaction result (`{ $kind, Transaction | FailedTransaction }`)
--- a/clients/cli/src/core/errors.test.ts
+++ b/clients/cli/src/core/errors.test.ts
@@ -636,7 +636,7 @@ describe('anticipated CLI taxonomy regressions', () => {
       )
     })
 
-    // Sui retired JSON-RPC: broadcasts go over gRPC, so the rejection carries a
+    // Sui is retiring JSON-RPC: broadcasts go over gRPC, so the rejection carries a
     // grpc-status NAME (string) instead of the old numeric -32002, and grpc-web
     // percent-encodes the message trailer. Both fixtures below are verbatim mainnet
     // captures from `fullnode.mainnet.sui.io`.
--- a/clients/cli/src/core/errors.test.ts
+++ b/clients/cli/src/core/errors.test.ts
@@ -636,11 +636,47 @@ describe('anticipated CLI taxonomy regressions', () => {
       )
     })
 
+    // Sui retired JSON-RPC: broadcasts go over gRPC, so the rejection carries a
+    // grpc-status NAME (string) instead of the old numeric -32002, and grpc-web
+    // percent-encodes the message trailer. Both fixtures below are verbatim mainnet
+    // captures from `fullnode.mainnet.sui.io`.
     it('maps a Sui transaction-execution client rejection to non-retryable input', () => {
-      const rpcError = Object.assign(new Error('Invalid user signature: cryptographic signature verification failed'), {
-        code: -32002,
+      const rpcError = Object.assign(new Error('Invalid user signature: Expect 1 signer signatures but got 0'), {
+        code: 'INVALID_ARGUMENT',
       })
       expectPermanent(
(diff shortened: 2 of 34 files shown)


gomesalexandre (approved):
Approving. This is genuinely strong work for a 34-file transport migration - the fail-closed posture from main is preserved in intent line-for-line, every behavior change carries a why-comment, and the new pagination handling is better than a naive port would have produced. Reviewed with the money-path lens because a transport swap is exactly where tx bytes, balances, or submission semantics silently drift.

## what I verified (three mutations, all fail-on-revert, pass-on-restore)

1. execution-success detection: forcing `isSuiExecutionSuccess` to fail-open (`result != null`) turns 10 named tests red, including the sdk#1398 MoveAbort false-success case (`promise resolved FailedTransaction instead of rejecting`). A gRPC ack alone is never treated as success - execution status is checked via `include: { effects: true }`. Restore -> green.
2. coin-listing completeness: mutating `listAllCoins` to read only the first page turns 4 tests red (`expected [2 coins] to deeply equal [3 coins]` + the resolver-level pagination test that feeds gas/input selection). The cursor is followed to completion, bounded fail-closed at 200 pages rather than truncating. Restore -> green.
3. gRPC transient classification: removing `transientGrpcStatuses` turns 3 tests red (UNAVAILABLE/DEADLINE_EXCEEDED/RESOURCE_EXHAUSTED). Restore -> green.

Post-restore: core-chain + sui keysign 1698 passed / 1 skipped; sdk unit 77 passed; cli 106 passed.

## money-path lanes

- broadcast: execution status checked, gRPC ack alone never success, and an aborted tx throws `DeliverTxFailedError` so the transient-retry regex can never re-send it (pinned by a named test).
- gas/coin selection: cursor-followed to completion, bounded fail-closed; `selectSuiInputCoins` throws on an insufficient pool so even a truncated read refuses construction rather than building a wrong send. Gas math unchanged.
- balances: `balance.balance` is the correct total field of the unified shape; live cross-transport receipt in the review - GraphQL returned the identical MIST value to JSON-RPC for the 0x0 address, same units and decimals.
- encoding boundaries: base64 at broadcast/hash/refine, signature stays base64 on both transports, coin-type normalization handles both spellings.
- two transports: both `SuiGrpcClient` and `SuiGraphQLClient` implement all 7 methods core uses with identical result shapes (verified in the installed dist); both endpoints live-probed against the shipped queries.

## should-fix (non-blocking, worth a fast-follow)

`packages/core/chain/chains/sui/listAllCoins.ts:68` and the mirror in `getSuiAllBalances` (`packages/sdk/src/tools/balance/otherBalance.ts`): `hasNextPage: true` with a null cursor exits the loop and returns the partial set as complete. In the keysign path this degrades safe (a spurious insufficient-balance throw downstream, never a wrong send), but `getSuiAllBalances` returns `ok: true` with a partial portfolio - the exact fund-visibility bug its own docstring forbids. And the test at `listAllCoins.test.ts:76` is named "does not truncate when hasNextPage is true but the cursor is null" while asserting the truncating behavior. Treat that combination as malformed: throw in `listAllSuiCoins`, return `tokens_unavailable` in `getSuiAllBalances`, and rename the test. It needs an upstream that violates the connection contract to trigger, hence non-blocking, but it is real.

## suggestions

- `resolvers/sui.ts`: the digest for `verifyBroadcastByHash` still comes from a network simulate. In the MPC dual-broadcast race, once the winner consumes the owned inputs, simulating the same bytes can reject at input resolution and the slower device rethrows for a tx that actually landed. Sui digests are computable offline (`Transaction.from(bytes).getDigest()`), race-proof. Pre-existing class, follow-up.
- `platforms/react-native/overrides/suiClient.ts:87`: the `as unknown as SuiClient` cast erases the compile-time guarantee that the GraphQL client covers the surface core calls. All 7 used methods exist on both today and a miss fails closed at runtime, but a future gRPC-only callsite would compile clean and break RN only. A type-level assertion or a prototype-method unit test closes it.
- `refine.ts:64`: `BigInt(gasUsed.computationCost ?? '0')` is marginally more fail-open than main; `shouldBePresent` on the fields would match the guard one line above. Dust-level impact.

## praise

the percent-encoded grpc-message trailer decode, the mainnet-captured error strings, and the honest "nothing is broken today" framing in the body. This is how a transport migration should be written.

## note

CI Test Suite was gated on the first-contributor approval and had never run - approving that run so the ~453 Sui assertions actually gate (they all pass locally in my runs above). Once it is green this is good to merge.

PR #1591, opened 2026-07-27
fix(thorchain): don't append the bank denom to native-token swap deposit symbols

### LINKED ISSUE #4464: TCY → RUNE swap keysign 404s with "Timed out while waiting for setup message" when the extension co-signs (Apple co-signer works)
## Summary

Field report (Discord support ticket #1562, 2026-07-26): a **Secure-vault THORChain intra-chain swap TCY → RUNE** fails at keysign whenever the **browser extension** is the co-signer — the waiting device times out with an HTTP 404 from the relay's setup-message endpoint. The same swap on the same vault succeeded immediately once the user switched the second signer to an Apple device.

## Observed vs expected

| Signer pair | TCY → RUNE keysign |
|---|---|
| Android v1.0.114 ↔ extension 0.2.23 | ❌ 404 "Timed out while waiting for setup message" — reproduced both initiating from the wallet app and from Rujira |
| Android v1.0.114 ↔ Apple device | ✅ signs |

Expected: the swap should co-sign across all client combinations.

## Error (screenshot in the ticket, shown on the desktop/extension side)

```
Timed out while waiting for setup message: HTTP 404 Error:
Request failed for https://api.vultisig.com/router/setup-message/804777da-fbd6-45c1-9d3f-da8e0379e771
```

## Why this doesn't look like session expiry or local network

The support bot's first diagnosis was "QR/session expired", but the failure was reproducible across multiple attempts and **two different initiation paths** (wallet app and Rujira), then disappeared the moment the co-signer platform changed with everything else held constant. A platform-correlated failure is not a timeout or a flaky network.

## Likely seam (hypothesis — not verified)

In `vultisig-sdk` `packages/core/mpc/message/setup/ensure.ts`, the joining device polls the relay with `messageId = getMessageHash(message)`, where `message` is the signing payload the joiner **rebuilds locally**; the "Timed out while waiting for setup message" prefix is exactly the error wrapping that wait. If the TS construction of the THORChain TCY→RUNE deposit tx diverges from Kotlin's by even one byte, the extension polls a `message_id` key the Android initiator never uploaded → deterministic 404 until timeout.

Android↔Apple working while Android↔extension fails fits the Kotlin/Swift-agree-TS-diverges parity signature (the three independent signing implementations; cf. the cross-platform signing parity vectors effort in commondata#94).

## Related issues

- vultisig/vultisig-android#2452 — the **exact same symptom** (TCY → RUNE, Secure vault, 404 setup-message timeout with a desktop-class co-signer) closed as `cant-reproduce` in Sept 2025. This ticket is a field repro on current versions.
- #4369 — same family: keysign completes mobile↔mobile but fails whenever the extension/desktop co-signs (SEI there).

## What would help pinpoint it

- Relay-side (or extension debug log) comparison of the `message_id` the extension polls vs the one Android uploads for the same session — a mismatch confirms the payload-hash divergence; a match points at the upload/session flow instead.
- A parity vector for THORChain `MsgDeposit` with the TCY denom across the Kotlin/Swift/TS signers.

## Environment

- Vault: Secure (2 devices)
- Android: v1.0.114 (initiator)
- Extension: 0.2.23
- Chain: THORChain intra-chain swap, TCY → RUNE
- Source: [Discord support ticket #1562](https://discord.com/channels/1203844257220395078/1531049826727366887) — user closed the ticket after the Apple-co-signer workaround, so the bug is unowned without this issue.

Ehsan-saradar:
## Description

A native swap sourced from a THORChain-native token encoded its `MsgDeposit` asset symbol as `` `${ticker}-${contractAddress}` ``. For these coins `contractAddress` **is the bank denom**, so TCY went out as `TCY-tcy` instead of `TCY`.

The symbol is part of the signed pre-image, so this moved the hash — which is what turned a cosmetic encoding slip into a hard keysign failure.

Fixes vultisig/vultisig-windows#4464 (cross-repo, so it won't auto-close).

## Root cause

`getThorchainDepositAsset` appended the contract address unconditionally:

```ts
const rawSymbol =
  typeof contractAddress === 'string' && contractAddress.trim()
    ? `${assetCoin.ticker}-${contractAddress}`
    : assetCoin.ticker
```

`assetCoin` here is `swapPayload?.fromCoin ?? coin`. The `?? coin` side is an `AccountCoin`, which carries `id` and no `contractAddress` — so this branch only ever fires for the swap payload's raw proto coin, where `contractAddress` is populated straight off the wire.

All three clients read the identical relayed payload and disagreed on one field:

| Client | Code | `THORChainAsset.symbol` |
|---|---|---|
| TS (extension/desktop) | `cosmos/index.ts` | **`TCY-tcy`** |
| Android | `ThorchainSwapHelper.kt` `getTicker(coin)` | `TCY` |
| iOS | `thorchain.swift` `getSwapPreSignedInputData` | `TCY` |

Android's registry sets TCY's `contractAddress = "tcy"` (`Coins.kt`) and relays it verbatim, so the TS client received the right input and encoded it wrong.

`symbol` was the **only** diverging field. I checked chain (`THOR` in all three), ticker, synth, secured, amount, decimals, signer, memo, accountNumber, sequence, mode, signingMode, chainId, and fee gas (20,000,000 in all three: `cosmosGasLimitRecord.ts`, `ThorChainHelper.THOR_CHAIN_GAS_UNIT`, `THORChainConstants.depositGasBaseUnits`).

## Why it failed the way it did

A co-signing joiner rebuilds the payload locally and polls the relay with `messageId = getMessageHash(message)` (`message/setup/ensure.ts`). One byte of divergence and it asks for a `message_id` the initiator never uploaded:

```
TCY  ts=TCY-tcy  hash=caf0d352…7b2d  message_id=001aa80e9c44a81810c21261ee2b3b1d
     native=TCY  hash=f6ca7ea9…a3d9  message_id=bb72d9a62108d815863705cd719c9f09
```

50 retries × 200ms later that surfaces as `404 Timed out while waiting for setup message` — the exact signature in the field report, where the swap failed against an extension co-signer and succeeded against an Apple one.

Worth noting this was never only a co-signing bug: a solo-signed TCY swap was encoding an asset that isn't `THOR.TCY`.

## The fix

Gate the `TICKER-CONTRACT` form on `secured`, which is the case it was added for in `5a8aacdeb` — a secured-asset withdrawal, where the auxiliary coin is the L1 token being pulled off THORChain and the contract address genuinely belongs in the symbol (`USDC` + `0xa0b8…` → `USDC-0XA0B8…`, matching THORNode's `ETH.USDC-0XA0B8…`).

That commit was aimed squarely at secured assets; `4815346d7` later carved the main secured path out into `getSecuredAssetDepositAsset` but left the suffix in the fallback, where it then only ever fired for THORChain-native tokens.

## Blast radius

Affected: any swap **sourced** from a THORChain-native non-RUNE token — TCY (`tcy`), RUJI (`x/ruji`), bRUNE (`x/brune`).

Unaffected, and confirmed unaffected by test: RUNE (empty contract address), secured assets (routed to `getSecuredAssetDepositAsset`), and all non-swap deposits — bond/merge/stake pass an `AccountCoin`, which never had a `contractAddress` to append.

## Verification

Post-fix pre-image hash parity against the mobile encoding:

```
TCY  -> RUNE  ts=TCY                       native=TCY                       MATCH
RUJI -> RUNE  ts=RUJI                      native=RUJI                      MATCH
RUNE -> BTC   ts=RUNE                      native=RUNE                      MATCH
secured USDC  ts=USDC-0XA0B86991…3606EB48  native=USDC-0XA0B86991…3606EB48  MATCH
```

Reverting just the gate fails exactly the two new swap tests and nothing else — they're real regression tests, not tautologies. Full core suite: 2170 passed.

## Not in scope — a second divergence this surfaced

While building the parity matrix I found a **separate, pre-existing iOS↔Android** disagreement: iOS uppercases the deposit symbol/ticker (`.uppercased()` in `thorchain.swift`), Android does not (`getTicker` returns it verbatim). Invisible for all-uppercase tickers, but **bRUNE is swappable** — `toNativeSwapAsset` normalizes `x/brune` to `THOR.bRUNE` — so iOS would emit `BRUNE` where Android and TS emit `bRUNE`. Same 404 failure class, with TS on Android's side.

Adjudicating that needs THORNode ground truth on which symbol it accepts for an `x/` denom, which I could not establish from the client repos. So the new tests are deliberately scoped to all-uppercase tickers rather than fossilizing an unverified choice. Worth its own issue. (`sTCY` is not affected in practice — it normalizes to an invalid quote asset and isn't swappable.)

## On golden vectors

vultisig/vultisig-sdk#1585 correctly flags TCY as having zero vector coverage — I confirmed it: all 5 swap vectors in `thorchainswap.json` source from L1 coins, and the RUJI cases in `thorchain.json` use `bond:`/`withdraw:` memos, which take the safe deposit branch. **None of the 59 vectors exercise this code path**, which is why it shipped.

I have not added a shared vector here on purpose: #1585's ground rule is that expected hashes must come from running at least two of the three real signers, and I could only run TS — seeding the corpus from the implementation that was just broken is exactly what that rule exists to prevent. The unit tests in this PR are the durable guard; the shared vector still belongs to #1585.

## Which feature is affected?

- [x] Swap - Please attach a tx link for the swap here
- [ ] New Chain / Chain related feature - Please attach tx link here

**Field validation** — I ran TCY → RUNE on a Secure vault in **both co-signing directions**, with this branch's build patched into a local `vultisig-windows` checkout:

| Initiator | Co-signer | Result |
|---|---|---|
| Extension | iOS | https://thorchain.net/tx/B784DE684A3EEB1D54B02346B18FD29359A28CD2FF1A3D792F0F5A1CC6838E77 |
| iOS | **Extension** | https://thorchain.net/tx/2605C3B2B5C95633940394256274FB5091769E589D779B94F325674F3A624B38 |

The second row is the reported failure shape — a mobile initiator with the extension co-signing, which is precisely what was 404ing in vultisig/vultisig-windows#4464. The first row exercises the TS client on the *initiating* side. Together they confirm that both clients now derive the same pre-image hash from the same payload, and that THORNode accepts the corrected asset.

## Checklist

## Additional context

The extension inherited this on 2026-03-26 when vultisig-windows migrated to `@vultisig/sdk` 0.8 (`babe1e509`). Before that it carried its own `core/mpc` with `symbol: coin.ticker`, which matches vultisig/vultisig-windows#3501 — a TCY → RUJI swap co-signed Windows↔iOS that completed on 2026-03-07, three weeks before the migration.

Desktop is affected identically to the extension. Both need an `@vultisig/sdk` release plus a bump of `@vultisig/core-mpc` in `clients/extension`, `clients/desktop` and `core/ui`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed THORChain native-token swap deposits to use the correct bare ticker symbol.
  - Prevented incorrect ticker-contract symbols from causing signing failures and timeouts.
  - Preserved contract-suffixed symbols for secured asset withdrawals.

- **Tests**
  - Added coverage for native and non-native swaps, secured withdrawals, and symbol encoding behavior.

### DIFF
--- a/.changeset/thorchain-native-token-deposit-symbol.md
+++ b/.changeset/thorchain-native-token-deposit-symbol.md
@@ -1,0 +1,28 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+fix(thorchain): don't append the bank denom to a native-token swap deposit symbol
+
+A swap sourced from a THORChain-native token (TCY, RUJI) encoded its
+`MsgDeposit` asset as `` `${ticker}-${contractAddress}` `` — and for these coins
+`contractAddress` IS the bank denom, so the deposit went out as `TCY-tcy`
+instead of `TCY`. iOS (`thorchain.swift` `getSwapPreSignedInputData`) and
+Android (`ThorchainSwapHelper`) both encode the bare ticker.
+
+Because the symbol is part of the signed pre-image, this moved the hash: a
+co-signing joiner rebuilds the payload locally and polls the relay with
+`messageId = getMessageHash(message)`, so it asked for a `message_id` the
+initiator never uploaded and failed with
+`404 Timed out while waiting for setup message`. Reported in
+vultisig/vultisig-windows#4464 as a TCY -> RUNE swap that fails whenever the
+extension co-signs and succeeds against an Apple co-signer.
+
+The `TICKER-CONTRACT` form is now gated on `secured`, which is the case it was
+written for (vultisig-sdk `5a8aacdeb`): a secured-asset withdrawal, where the
+auxiliary coin is the L1 token being pulled off THORChain and the contract
+address genuinely belongs in the symbol (`USDC` + `0xa0b8…` ->
+`USDC-0XA0B8…`, matching THORNode's `ETH.USDC-0XA0B8…`). Both sides of that
+gate are now covered by tests; RUNE, secured assets, and non-swap deposits
+(bond/merge/stake) were never affected.
--- a/packages/core/mpc/keysign/signingInputs/resolvers/cosmos.test.ts
+++ b/packages/core/mpc/keysign/signingInputs/resolvers/cosmos.test.ts
@@ -391,6 +391,141 @@ describe('getCosmosSigningInputs gas limit', () => {
     expect(depositCoin?.decimals.toString()).toBe('8')
   })
 
+  // vultisig/vultisig-windows#4464: a THORChain-native token carries its bank
+  // denom in `contractAddress`, which must NOT be appended to the deposit
+  // symbol. iOS (`thorchain.swift` getSwapPreSignedInputData) and Android
+  // (`ThorchainSwapHelper`) both encode the bare ticker; emitting `TCY-tcy`
+  // moved the pre-image hash and 404'd every mixed-platform co-sign.
+  //
+  // Limited to all-uppercase tickers on purpose: iOS uppercases the deposit
+  // symbol/ticker and Android does not, so a mixed-case ticker (`bRUNE`) is a
+  // separate, still-unadjudicated iOS<->Android divergence. Pinning one side
+  // here would fossilize an unverified choice.
+  it.each([
+    { label: 'TCY', ticker: 'TCY', contractAddress: 'tcy' },
+    { label: 'RUJI', ticker: 'RUJI', contractAddress: 'x/ruji' },
+  ])('encodes an intra-THORChain swap of $label on the bare ticker', async ({ ticker, contractAddress }) => {
+    const [input] = await getCosmosSigningInputs({
+      keysignPayload: create(KeysignPayloadSchema, {
+        coin: create(CoinSchema, {
+          chain: Chain.THORChain,
+          ticker,
+          address: thorSender,
+          contractAddress,
+          decimals: 8,
+          isNativeToken: false,
+          hexPublicKey: publicKeyHex,
+        }),
+        toAddress: '',
+        toAmount: '100000000',
+        memo: `=:THOR.RUNE:${thorSender}`,
+        blockchainSpecific: {
+          case: 'thorchainSpecific',
+          value: create(THORChainSpecificSchema, {
+            accountNumber: 7n,
+            sequence: 3n,
+            fee: 2_000_000n,
+            isDeposit: false,
+            transactionType: TransactionType.UNSPECIFIED,
+          }),
+        },
+        swapPayload: {
+          case: 'thorchainSwapPayload',
+          value: {
+            fromAddress: thorSender,
+            fromCoin: {
+              chain: Chain.THORChain,
+              ticker,
+              address: thorSender,
+              contractAddress,
+              decimals: 8,
+              isNativeToken: false,
+            },
+            toCoin: {
+              chain: Chain.THORChain,
+              ticker: 'RUNE',
+              address: thorSender,
+              contractAddress: '',
+              decimals: 8,
+              isNativeToken: true,
+            },
+            fromAmount: '100000000',
+            vaultAddress: '',
+            routerAddress: '',
+            expirationTime: 0n,
+          },
+        },
+      }),
+      walletCore,
+    })
+
+    const depositCoin = input.messages[0]?.thorchainDepositMessage?.coins?.[0]
+    expect(depositCoin?.asset).toMatchObject({
+      chain: 'THOR',
+      symbol: ticker,
+      ticker,
+      synth: false,
+      secured: false,
+    })
+    expect(depositCoin?.amount).toBe('100000000')
+    expect(depositCoin?.decimals.toString()).toBe('8')
+  })
+
+  // The other side of the same gate: on a secured-asset withdrawal the
+  // auxiliary coin IS an L1 token, and its contract address belongs in the
+  // symbol (`ETH.USDC-0XA0B8…`). Pins the branch #4464's fix narrowed to.
+  it('keeps the contract suffix on a secured withdrawal of an L1 token', async () => {
+    const [input] = await getCosmosSigningInputs({
+      keysignPayload: create(KeysignPayloadSchema, {
+        coin: create(CoinSchema, {
+          chain: Chain.THORChain,
+          ticker: 'RUNE',
+          address: thorSender,
+          decimals: 8,
+          isNativeToken: true,
+          hexPublicKey: publicKeyHex,
+        }),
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
Approving. The deposit symbol here goes into the signed MsgDeposit pre-image, so a wrong symbol misroutes or fails a THOR swap - reviewed both directions and it is a clean, convergent fix.

What main did wrong: a native-token intra-THOR swap encoded the asset as TICKER-CONTRACT where the contract IS the bank denom - TCY became TCY-tcy, RUJI became RUJI-x/ruji. The fix drops the suffix on that path so it emits the bare TCY / RUJI.

The grounding is production, not the PR's own assertion: iOS (thorchain.swift:86, bare uppercased ticker for non-secured) and Android (ThorchainSwapHelper.kt bare getTicker) both emit the bare ticker, and windows#4464 shows TCY swaps succeed against an Apple co-signer with the bare form and 404 when a divergent pre-image is rebuilt. THORNode has no THOR.TCY-tcy asset, so main's form also fails node-side. So this is the SDK catching up to what the flagship apps already sign - convergent, the safe cross-consumer direction.

Both directions mutation-pinned:
- revert the `secured &&` gate -> the TCY and RUJI intra-THOR swap tests fail (got TCY-tcy / RUJI-x/ruji). 
- over-fix to always-bare-ticker -> the secured L1 withdrawal test fails (USDC lost its USDC-0x... contract suffix). So the non-native path staying intact is pinned too - the fix touches only the native branch (`isSecuredAssetWithdrawal` requires `fromCoin.chain !== THORChain`, which a THOR-native swap can never satisfy).

Restore -> 16/16 on the file, full `yarn test:core` 2221 passed. These are a real PR gate (Unit Tests runs test:core), not e2e-gated.

`suggestion:` the changeset says non-swap deposits (bond/merge/stake) "were never affected" - not strictly true: a non-swap deposit of TCY also went TCY-tcy on main and is now bare TCY. That change is also convergent with iOS/Android, so it is a wider fix than the prose claims, not a hazard. Optional: add a `tcy+` non-swap deposit case to the it.each to pin that path, or soften the changeset line.

`q:` (pre-existing, not this PR) mixed-case tickers (bRUNE, sTCY) still diverge - iOS uppercases and strips x/, the SDK does neither. Explicitly documented as deliberately unpinned pending iOS/Android adjudication; this PR shrinks that divergence rather than creating it. Worth a follow-up bead if those tokens go live.

`praise:` the test comment explaining WHY mixed-case is left unpinned (rather than fossilizing an unverified choice into a pre-image field) is exactly the right instinct.

PR #1602, opened 2026-07-27
test: pin golden addresses in the all-chains derivation suite

### LINKED ISSUE #1539: test: all-chains address derivation asserts format regexes, not golden addresses (a wrong-but-valid address passes)
Found during a cross-repo test-quality audit (2026-07-24), analysed against `origin/main` (`f99133dc7`).

## 1. The all-chains address-derivation test validates SHAPE, not VALUE

`packages/sdk/tests/integration/address-derivation/all-chains.test.ts` is the SDK's flagship multi-chain derivation coverage - "verifies that ALL supported chains can derive valid addresses using the PUBLIC SDK API with REAL WASM modules", 40+ chains.

Every assertion is a format check. `CHAIN_VALIDATORS` is a map of regexes:

```
Bitcoin: addr => /^(bc1|[13])[a-zA-HJ-NP-Z0-9]{25,62}$/.test(addr),
Litecoin: addr => /^(ltc1|[LM])[a-zA-HJ-NP-Z0-9]{25,62}$/.test(addr),
Ethereum: addr => /^0x[a-fA-F0-9]{40}$/.test(addr),
```

and the per-chain loop asserts `toBeDefined()`, `typeof === 'string'`, `length > 0`, `validator(address) === true`, `length >= 20`. The pubkeys it derives from are synthetic ("Real-ish looking public keys"), so there is nothing to compare against.

The consequence: a derivation regression that produces a **valid-format but wrong** address - swapped coin type, wrong derivation path, wrong chain params, a hash function drift - passes every assertion in this file. That is the highest-consequence failure mode for address derivation, and it is precisely the one this suite cannot see. Funds sent from, or received at, an address the user does not control.

The neighbouring assertions that DO bite are the good ones and show the pattern is available: `:411` (`address1 === address2`, caching determinism), `:445` (all EVM chains share one address), `:474` (prefix checks). What is missing is a fixed pubkey set with the exact expected address per chain.

- [ ] add golden vectors: one fixed test vault pubkey set, exact expected address per chain, asserted with `toBe`. The consuming repos have the same hole (vultisig/vultiagent-app#2269 item 2 - the app mocks `deriveAddress` outright), so the SDK is the right place to own it.

## 2. The whole `tests/e2e/` surface never runs on a PR

`.github/workflows/test.yml` runs `yarn test` (unit) + `yarn test:integration` + `yarn check:ci`. It never runs `yarn test:e2e`. Every e2e suite is additionally `describe.skipIf(!HAS_TEST_VAULT_FIXTURE)` (`packages/sdk/tests/e2e/helpers/test-vault.ts:54` - the gate is `existsSync(TEST_VAULT_CONFIG.path)`), so even if it were invoked it would no-op in a clean checkout.

That covers: `fast-signing`, `sign-bytes`, `arbitrary-signing`, `prepare-send-tx`, `swap-quotes`, `swap-transactions`, `gas-estimation`, `balance-operations`, `multi-chain-coverage`, `secure-vault-multiparty-signing`, plus `tests/e2e/funded/`.

The repo is already honest about this - `packages/sdk/tests/unit/crypto/signingRoundTrip.synthetic.test.ts` was written specifically because "the only test that exercises a full sign -> format -> verify round trip today is tests/e2e/fast-signing.test.ts, which is describe.skipIf(!HAS_TEST_VAULT_FIXTURE) and therefore skipped everywhere except a daily cron". That synthetic test is the right answer and there should be more of it. Filing so the shape of the gate is written down rather than rediscovered.

- [ ] document in the workflow what the PR gate does and does not cover (a comment block, same as agent-backend-ts's `ci.yml` does)
- [ ] extend the synthetic (non-vault-gated) pattern to the seams currently only covered by a vault-gated e2e: `prepare-send-tx` per chain family, and swap-quote calldata shape

## 3. Minor: the "integration" swap-quote test mocks the quote engine

`packages/sdk/tests/integration/swap/swap-quote.test.ts:26` does `vi.mock('@vultisig/core-chain/swap/quote/findSwapQuote')`, plus `getErc20Allowance`, `getCoinBalance` and `swapEnabledChains`. What remains under test is the `VaultBase.getSwapQuote()` wrapper and its event emission - which is legitimate and worth testing, it just is not what "Integration: Swap Quote" implies to someone reading the CI job name `Integration Tests (Vitest)`. Naming only; no action strictly required, but a one-line header note would stop it being read as end-to-end quote coverage.

## what was NOT audited

`packages/core/*` (157 chain + 59 mpc test files) and `clients/cli` were only inventoried, not read. No mutation testing. The local clone at hand is a downstream fork 1766 commits ahead of `origin/main`, so everything above was read via `git show origin/main:<path>` - if any of it has moved since `f99133dc7`, say so and I will recheck.

gomesalexandre:
Closes vultisig/vultisig-sdk#1539

## what

`tests/integration/address-derivation/all-chains.test.ts` is the SDK's flagship multi-chain derivation coverage (40+ chains, real WASM, public API). Every assertion in it is a format REGEX (or, for the Cosmos block, a bech32 prefix check) run against a synthetic pubkey. That validates SHAPE, not VALUE.

A derivation regression that produces a well-formed-but-WRONG address - a swapped coin type, wrong derivation path, wrong hash function - passes every one of those checks, because a regex only knows what an address *looks like*, not what it *is*. That's the highest-consequence failure mode for address derivation (funds sent from, or received at, an address the user doesn't actually control) and it's exactly the one this suite couldn't see.

## how

Added a `Golden address vectors` block covering one chain per family - Bitcoin (UTXO), Ethereum (EVM), Cosmos, Solana (EdDSA) - asserting `vault.address()` with `toBe()` against an INDEPENDENT reference re-derivation computed inside the test, not a value captured by running the SDK once and pasting the result (which would just fossilize whatever it currently does, bug included).

The reference re-implements, using different libraries than `packages/core/chain/publicKey/{getPublicKey,ecdsa/derivePublicKey,address/deriveAddress}.ts`:
- BIP32 public-key-only child derivation via `bip32` + `tiny-secp256k1` (same audited libraries the SDK itself depends on for this - they're the industry reference, not something Vultisig authored) along the coin-specific paths, non-hardened at every segment (matching `derivePublicKey.ts`'s own scheme of stripping the `'` hardened markers, since a public-only root key can't do hardened CKD). The exact paths were verified independently by calling `walletCore.CoinTypeExt.derivationPath()` directly rather than assumed: bitcoin `m/84'/0'/0'/0/0`, ethereum `m/44'/60'/0'/0/0`, cosmos `m/44'/118'/0'/0/0`.
- Per-chain address ENCODING via dedicated libraries instead of WalletCore's `AnyAddress`/`CoinTypeExt`: `bitcoinjs-lib`'s P2WPKH payment for Bitcoin, hand-rolled Keccak-256 + EIP-55 checksum for Ethereum, `@noble/hashes`' sha256/ripemd160 + `bech32` for the Cosmos-SDK `RIPEMD160(SHA256(pubkey))` scheme.
- Solana needs no HD derivation at all - `getPublicKey.ts`'s eddsa branch returns the root eddsa pubkey unchanged (no per-chain derivation happens for eddsa chains), so the golden address is simply `bs58` of that raw pubkey.

All four reference computations were cross-checked against WalletCore's own `CoinTypeExt.deriveAddressFromPublicKey` for the same derived pubkey before being wired into the test, and match exactly.

## mutation-verify (the whole point)

Introduced a one-line off-by-one bug in `derivePublicKey.ts`'s path-index parsing (`pathBuf.push(index + 1)` instead of `index`) - every BIP32 derivation index shifts by one, for every chain.

- Every EXISTING regex/prefix assertion in the file stayed green. The mutated addresses are still well-formed `0x...`/`bc1...`/`cosmos1...` strings - just the wrong ones.
- The three new golden `toBe()` assertions (Bitcoin/Ethereum/Cosmos) failed immediately with the exact wrong-but-valid values:
  ```
  Expected: "0xC190DDb708e948832FFb41CED2AB29A4a6a978DD"
  Received: "0xB093F6Afb9BF83De93845307288676E1d28e676A"
  ```
- Solana's golden test correctly stayed green, since that mutation only touches the ecdsa BIP32 path and Solana's eddsa branch never goes through it - confirming the golden vectors aren't just flaky/order-dependent.

Reverted the mutation after confirming; all 95 tests in the file pass again.

## testing

- `yarn build:shared`
- `yarn workspace @vultisig/sdk typecheck` - clean
- `yarn vitest run --config tests/integration/vitest.config.ts tests/integration/address-derivation/all-chains.test.ts` - 95/95 passed
- lint clean

Test-only change, no changeset (no published behavior change).

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive cross-chain verification for derived wallet addresses.
  * Validates exact address results across UTXO, EVM, Cosmos, and EdDSA-based chains using independent cryptographic checks.
  * Strengthened deterministic test coverage with fixed derivation inputs and golden address vectors.

### DIFF
--- a/packages/sdk/tests/integration/address-derivation/all-chains.test.ts
+++ b/packages/sdk/tests/integration/address-derivation/all-chains.test.ts
@@ -17,11 +17,19 @@
  */
 
 import { create, toBinary } from '@bufbuild/protobuf'
+import { ripemd160 } from '@noble/hashes/legacy.js'
+import { sha256 } from '@noble/hashes/sha2.js'
+import { keccak_256 } from '@noble/hashes/sha3.js'
 import { Chain } from '@vultisig/core-chain/Chain'
 import { LibType } from '@vultisig/core-mpc/types/vultisig/keygen/v1/lib_type_message_pb'
 import { VaultContainerSchema } from '@vultisig/core-mpc/types/vultisig/vault/v1/vault_container_pb'
 import { Vault_KeyShareSchema, VaultSchema } from '@vultisig/core-mpc/types/vultisig/vault/v1/vault_pb'
 import type { Vault as CoreVault } from '@vultisig/core-mpc/vault/Vault'
+import { bech32 } from 'bech32'
+import BIP32Factory from 'bip32'
+import * as bitcoin from 'bitcoinjs-lib'
+import bs58 from 'bs58'
+import * as ecc from 'tiny-secp256k1'
 import { afterAll, beforeAll, describe, expect, it } from 'vitest'
 
 import { createSdkContext, type SdkContext } from '../../../src/context/SdkContextBuilder'
@@ -92,6 +100,16 @@ const CHAIN_VALIDATORS: Record<string, (address: string) => boolean> = {
 
 /** Deterministic hex pubkey for QBTC (ML-DSA); deriveQbtcAddress only hashes bytes — no chain signing. */
 const MOCK_MLDSA_PUBLIC_KEY_HEX = '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef'
+
+/**
+ * Fixed test-vault root keys shared by the mock vault below AND the golden-vector
+ * describe block further down. A valid (on-curve) compressed secp256k1 pubkey +
+ * 32-byte chain code, so `derivePublicKey`'s BIP32 CKD math and the golden
+ * reference derivation below operate on the exact same root.
+ */
+const MOCK_ECDSA_PUBLIC_KEY_HEX = '02a1633cafcc01ebfb6d78e39f687a1f0995c62fc95f51ead10a02ee0be551b5dc'
+const MOCK_EDDSA_PUBLIC_KEY_HEX = 'b5d7a8e02f3c9d1e4f5a6b7c8d9e0f1a2b3c4d5e6f7a8b9c0d1e2f3a4b5c6d7e'
+const MOCK_HEX_CHAIN_CODE = '1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef'
 
 /** Base64 VaultContainer (unencrypted) wrapping a Vault protobuf that includes ML-DSA fields. */
 function buildUnencryptedVultBase64(params: {
@@ -180,10 +198,10 @@ describe('Integration: Multi-Chain Address Derivation', () => {
       name: 'Integration Test Vault',
       publicKeys: {
         // Real-ish looking public keys (proper format for address derivation)
-        ecdsa: '02a1633cafcc01ebfb6d78e39f687a1f0995c62fc95f51ead10a02ee0be551b5dc',
-        eddsa: 'b5d7a8e02f3c9d1e4f5a6b7c8d9e0f1a2b3c4d5e6f7a8b9c0d1e2f3a4b5c6d7e',
+        ecdsa: MOCK_ECDSA_PUBLIC_KEY_HEX,
+        eddsa: MOCK_EDDSA_PUBLIC_KEY_HEX,
       },
-      hexChainCode: '1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef',
+      hexChainCode: MOCK_HEX_CHAIN_CODE,
       localPartyId: 'test-device',
       signers: ['test-device', 'Server-1'],
       keyShares: {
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
This is a strong, genuinely load-bearing golden-vector test. The prior suite asserted only format REGEX / bech32 PREFIX — SHAPE, not VALUE — so a well-formed-but-WRONG address (swapped coin type, wrong path, wrong hash) sailed through. This block closes exactly that highest-consequence gap for one chain per family (UTXO/EVM/Cosmos/EdDSA). I verified the crux (golden values correct, not fossilized) independently rather than trusting the PR's own mutation-verify.

## testing approach

- **Worktree at PR HEAD (`0575470dc`)**, symlinked canonical `node_modules` + `dist` dirs (worktree packages were unbuilt; without dist symlinks the import chain dies at `core/config`). Ran the block: 4/4 green.
- **Golden values triple-confirmed correct (the crux).** I re-derived all four addresses with a stack that shares ZERO code with the SDK/test `bip32`+`tiny-secp256k1` path: `@noble/curves` secp256k1 point-addition public-CKD + `hmac-sha512`, `ethers.getAddress` for canonical EIP-55, `@scure/base` for bech32/base58. Exact match on all four:
  - ETH `0xC190DDb708e948832FFb41CED2AB29A4a6a978DD`
  - BTC `bc1qxkymttl27q3y200zngaf6r3z88a6cw365yzqf3`
  - Cosmos `cosmos17nsl9778wmdaz9s0w8nk0wdpkaamg5pnplwllh`
  - Solana `DEqXP4qf9dEYB7okqYbGAmjCgPjdy7vDDZ5gFBieT99F`
  So SDK output == the test's in-line reference == my independent cross-check (three separate implementations).
- **Control-mutation-proven load-bearing (symlink-safe).** Mutated the worktree's `derivePublicKey.ts` (`push(index)` → `push(index + 1)`): the 3 ECDSA goldens went RED with wrong-but-valid addresses (ETH `Received: 0xB093F6Afb9BF83De93845307288676E1d28e676A`, exactly matching your stated mutation-verify), and the Solana/EdDSA golden correctly stayed GREEN (it never traverses the ECDSA path). This also proves the run executed the worktree's own derivation code, not canonical-via-symlink. Restored → green again.
- Test-only: 1 file, no production derivation code touched, no changeset.

## q

- The reference is independent at the **library** level and for **path-selection** (SDK `getCoinType`/`derivationPath` vs the hardcoded `[44,60,0,0,0]` etc) and **address encoding** (WalletCore `AnyAddress` vs `bitcoinjs-lib`/keccak/bech32) — which is precisely where swapped-coin-type / wrong-hash bugs live, so those are genuinely caught. It is NOT independent at the **derivation-semantics** level: it hard-codes the SDK's own public-only, hardened-markers-stripped non-hardened scheme. If the *model itself* were ever wrong (e.g. hardened-boundary handling), the reference would mirror it and still pass. I'm treating this as inherent-and-correct rather than a gap: an MPC vault holds only the root pubkey + chaincode, hardened CKD is mathematically impossible, so the non-hardened public-CKD scheme *defines* the addresses users actually control today — which is the right thing to pin. Worth a one-line comment in the block making that "consistency-with-model, not proof-of-model" boundary explicit for future readers. No change required.

## risk

Low. Test-only, fund-safety-positive: it converts a value-blind derivation suite into one that catches well-formed-but-wrong-address regressions on the four representative families. Coverage is family-representative (4 chains), not exhaustive across all 40+ — acceptable, since the existing regex suite still covers every chain for shape and the four goldens cover the four distinct derivation schemes by value.

## verdict

APPROVE. Golden addresses independently verified correct (three separate implementations agree), the test derives via the real SDK `vault.address()` and is control-mutation-proven RED on a real derivation perturbation, and it's test-only. Nice work.

---
🤖 dual-lane review (Claude load-bearing + Codex adversarial; lanes converged on the single library-vs-semantic-independence nuance, no blocking findings)

NeOMakinG (approved):
Re-approve at HEAD: the new commit is a verified prettier-only reformat of the test file (no golden address / no export name changed vs the approved commit — content-identical reflow). Prior approval stands.

NeOMakinG (approved):
## testing approach

Focused merge-churn re-review of the `Merge branch 'main'` commit `5d51f2be` (p1 = prior head `6c290faf`, p2 = `d7fd8fd8` = current `origin/main` head). Fresh worktree `/tmp/wt-sdk-1602b` at HEAD, canonical `node_modules` + `dist` symlinked (no `pnpm install`), vitest integration config.

## prior issues — status

| r1 | 1 | suggestion | library-level (not semantic) independence of golden derivation | ℹ️ documented | unchanged by merge; accepted rationale stands (MPC vaults only support non-hardened public-CKD, which defines the addresses users control) |

## should-fix

_none_

## q

_none_

## risk

**Test-only, merge behaviorally inert.**
- Golden test file `packages/sdk/tests/integration/address-derivation/all-chains.test.ts` is **byte-identical** across the merge (`git diff 6c290faf..5d51f2be` empty) — the merge did not touch any golden value.
- Main brought in only Jupiter swap affiliate-fee changes (#1582): `getJupiterSwapQuote.ts`, `tools/swap/jupiter.ts` + the swap tests + a changeset. **No** derivation-path / `getCoinType` / `derivePublicKey` / WalletCore / lockfile changes — the address stack is untouched, so the pinned goldens cannot drift.
- Test **green at HEAD**: full file 95/95 pass, all 4 golden vectors (ETH/BTC/Cosmos/Solana) pass.
- **Control-mutation still RED** at HEAD: worktree `derivePublicKey.ts` `push(index)`→`push(index+1)` turned the 3 ECDSA goldens RED (ETH Received `0xB093F6…`, matching the r1 mutation-verify value) while the EdDSA Solana golden stayed GREEN — proving the assertions execute the worktree's derivation code and are live, not inert.
- Codex lane (exit 0): no findings; confirms no `package.json`/lockfile drift and the Jupiter changes don't touch the address stack. Sole residual is a procedural note (verify CI runs `test:integration` so this block executes) — pre-existing coverage concern, non-blocking.

## verdict

Goldens byte-identical + main did not regress derivation + test green at HEAD with pinned goldens + control-mutation still RED. **Re-approving** at `5d51f2be`.

---
🤖 dual-lane review (Claude load-bearing + Codex adversarial); GLM retired.

PR #1605, opened 2026-07-28
fix(agent): report the proposed transaction when ask mode declines signing

neavra:
## The problem

`agent ask --help` documents the no-`--yes` path as read-safe:

> Without `--yes`, ask mode never signs or broadcasts — **it reports the proposed transaction**
> so a read-only prompt can't move funds.

It did not report the proposed transaction. On a send prompt without `--yes`:

1. the backend builds the transaction — **this succeeds**;
2. the confirm gate declines to sign it — **correct**;
3. the refusal is queued as a `recent_action` and **recursed back into the model loop**;
4. the model treats the decline as an error and retries; the retry `execute_send` fails;
5. the turn ends with `transactions: []`, `cards: []`, no proposed transaction, and prose
   stating a cause that was simply untrue — that the build had errored, that there was no
   send tool in the session, or that a broadcast could not be confirmed.

Nothing was ever authorized to broadcast. Refusing to sign was right; discarding the built
transaction and reporting the refusal as a failed build was the defect.

## The change

**A declined signing ends the turn in ask mode.** Without `--yes` the gate is a fixed policy,
not a decision the model can influence, so recursing the refusal can only buy a retry that
fails again — which is exactly what produced the misleading turn. The loop now stops on the
decline and drops the queued refusal so it cannot leak into the next turn.

**The built, unsigned transaction becomes the turn's result.** It is captured before
`clearPendingTransaction()` drops the envelope (a declined signing is the one case where the
unsigned transaction *is* the result) and surfaced as a new `onProposedTransaction` callback →
`AskResult.proposedTransaction`:

- JSON envelope: `data.proposed_transaction` (`tool`, `summary`, `chain`),
  `data.confirmation_required`, `data.proposed`
- human output: `confirmation-required:` and `proposed:` lines

Previously these fields existed only on the **success** envelope, which a declined sign never
reaches — so the read-safe path returned an error envelope with nothing to read.

**The message states what happened.** The raw `Transaction not confirmed` reads as though the
transaction was attempted and came back unconfirmed; it is replaced with an explicit
"built but signing was not authorized, nothing was signed or broadcast, re-run with `--yes`".

**Exit code is unchanged at 12 (`CONFIRMATION_REQUIRED`)** — the slot `send` and `swap`
already use for "needs `--yes`". No new exit code was invented. Additionally, a backend turn
outcome carrying code `confirmation_required` now maps to 12 rather than the generic
safety-block 10; the mapping is keyed on the **code**, so the backend can express this under
an existing `outcome.kind` and no consumer meets a kind it does not already handle.

### TUI and pipe mode are deliberately unchanged

There a decline is a live user (or host) decision mid-conversation, and the model **should**
get to acknowledge it; the host can also approve on a later turn. The short-circuit is scoped
to the non-interactive ask path. Both behaviours are pinned by tests.

The proposed-transaction surface itself is emitted on **every** client, not just ask mode.

## Tests

- `runPasswordGatedTool`: a declined `sign_tx` carries `proposed_chain` and fires
  `onProposedTransaction` with the summary captured **before** the buffer is dropped
- `processMessageLoop`, ask mode: nothing signs, **no second request is made**, the proposed
  transaction is the result. The pre-existing test that asserted the decline was recursed back
  (`streamRequests` length 2) is rewritten — it pinned the behaviour this fixes.
- `processMessageLoop`, **outside** ask mode: the decline is still reported back and the loop
  still recurses — the regression guard for the TUI/pipe carve-out
- `agent ask --json`: exit 12 with `proposed_transaction` / `confirmation_required` /
  `proposed` on the envelope; the message asserts positively on the real cause **and**
  negatively on each of the three false messages the defect actually produced
- human (non-JSON) output surfaces the same
- a backend `confirmation_required` outcome exits 12, while an ordinary `blocked` outcome
  still exits 10 (the new code does not swallow real safety blocks)

**Revert-sensitivity verified explicitly**: reverting the five source files to `origin/main`
while keeping the tests turns **7 tests red**, one per behaviour above; restoring makes them
green.

## Validation

- `yarn workspace @vultisig/cli typecheck` — clean (after `yarn build:shared` + `yarn build:all`,
  as CI does)
- `yarn test:cli` — **927 passed, 0 failed** (68 files)
- ESLint on the changed files — clean
- Prettier (repo config) on the changed files — clean
- Changeset included for `@vultisig/cli`. `@vultisig/sdk` is untouched — the executor seam
  change (`getPendingChain`) is internal to `clients/cli`.

## Companion

Depends on nothing, but pairs with the backend change that stops classifying a declined
signing as `tool_failed`: vultisig/agent-backend-ts#2132. This PR is useful on its own — the
short-circuit means the misleading follow-up turn is never requested in the first place.

---

## Review rounds 2 and 3 — what a local multi-perspective review changed

**Round 2 — the short-circuit discarded work that had already happened.** Client-side dispatches are
awaited *before* the signable candidate is handled, so a turn can both run a local mutation
(`vault_chain` / `vault_coin` / `address_book`) and propose a transaction. Clearing the whole pending
queue on the decline discarded those results too. The mutations stay committed locally, so the effect
was a silently lost acknowledgement: the backend never learned, and a resumed multi-step flow could
repeat the operation or reason from stale state. Only the refusal is dropped now.

**Round 2 — `sign_typed_data` bypassed the fix entirely.** It is gated by the same confirm gate but
reaches the queue via `dispatchClientSideTool` rather than the signable branch, so
`vultisig agent ask "bet 5 USDC on X"` without `--yes` kept the exact retry-into-failure behaviour.
The same terminal check now covers that route.

**Round 2 — the proposed-transaction summary was uncapped.** For `sign_tx` it is a one-line human
string, but the `sign_typed_data` fallback is `JSON.stringify(input)` — an arbitrarily large typed-data
blob reaching stdout and the JSON envelope through a field documented as one line.

**Round 3 — only the FIRST queued decline was removed.** Dispatches are serialized and all awaited,
so one response can queue several declined `sign_typed_data` calls; and a response mixing a declined
`sign_typed_data` with a buffered `sign_tx` returned from the signable branch without reaching the
tail check at all. Either way a stale refusal stayed queued and the **next** request flushed it into
an unrelated turn's `recent_actions`. Every decline is now dropped, on both terminal paths.

**Round 3 — the cap applied to the callback but not to `data.proposed`**, so a declined typed-data
signature shipped two conflicting descriptions of the same proposal. Capped once, before both
consumers.

### On the `askMode` seam (challenged twice, kept deliberately)

The cross-model reviewer argued the seam should not be a mode, since pipe is also a headless
interface. Kept, because the discriminator is **policy-vs-decision**, not headless-vs-interactive:
ask mode's `requestConfirmation` returns a constant, so no retry can ever satisfy it, whereas
`pipe.ts` blocks on a live host answer over stdin and the TUI prompts a live user — there a decline
is a real decision the model should get to acknowledge, and a later turn can legitimately be
approved. The inaccurate original comment ("the host can approve on a later turn", when the recursion
happens inside the same `sendMessage`) was corrected. Both non-ask paths are pinned by tests.

### A refactor CI forced

The decline handling pushed `processMessageLoop` to 44 and `runPasswordGatedTool` to 41 against the
repo's `sonarjs/cognitive-complexity` limit of 35 (both were under it on `main`, so this was the
change's own debt). The path is extracted into four **module-level** functions. Module-level rather
than private methods on purpose: the session's private methods are exercised through the prototype
with a minimal `this`, so every new method must be wired into each harness — a first attempt using
methods broke 35 tests across four files for no gain.

**Validation (final):** `yarn workspace @vultisig/cli typecheck` clean · `yarn test:cli` **934 passed,
0 failed** · `yarn check:ci` (the full CI quality gate: lint, types, knip, Prettier, complexity,
jscpd, markdownlint) **exit 0 locally**. Revert-sensitivity re-verified after each round: reverting
the five source files turns **10 tests red** (was 7).

## Summary by CodeRabbit

* **Bug Fixes**
  * Declined signing in `agent ask` now ends the turn gracefully instead of retrying.
  * Built-but-unsigned transaction details are preserved when confirmation is declined.
  * Confirmation-required outcomes consistently return exit code `12`.
  * Supports both transaction and typed-data signing flows while preserving earlier client-side actions.

* **Output**
  * JSON output includes confirmation status and proposed transaction details.
  * Human-readable output provides clearer confirmation guidance and refusal messaging.


gomesalexandre (changes requested):
real bug, right fix, and honestly some of the strongest test discipline i've seen on this repo. code-wise this looks sound on a remote pass. blocker is receipts, not logic.

## what i verified remotely

- ask-mode scoping is tight. `isAskModeDecline` only fires on `askMode && !success && CONFIRMATION_REQUIRED`, so tui/pipe keep report-and-continue.
- coderabbit's mixed-turn "successful sign_tx plus queued declined typed-data" finding is not reachable today. in ask mode the gate is a per-run constant in `ask.ts`, so without `--yes` everything declines and the queue-purge path runs, and with `--yes` nothing declines. outside ask mode `takeQueuedAskModeDeclines` is a no-op by design.
- `getPendingChain()` is read before `clearPendingTransaction()`, and it uses the same `'latest'` slot the summary/sign path already uses.
- the 500-char cap is applied once before both consumers, so `data.proposed` and `proposed_transaction.summary` cannot drift.
- outcome-code mapping is keyed on `confirmation_required`, and ordinary blocked turns still stay on exit 10.

## blocker: missing runtime receipt for a user-visible cli change

this pr changes shipped cli behavior on the actual user contract:
- new error/exit-path semantics
- new json fields: `confirmation_required`, `proposed_transaction`, `proposed`
- new human output lines
- rewritten confirmation-required message

the validation in the body is all internal: vitest, typecheck, check:ci. that's good, but it's not the runtime receipt bar for a cli-visible sdk change.

please add one real `vultisig agent ask` transcript against a live backend:

1. without `--yes`, showing the confirmation-required path, the proposed transaction fields in the json envelope, and exit code `12`
2. with `--yes`, showing the normal signing path still works

paste that transcript in the pr body or a comment and i'm happy to re-check. once that receipt is there this looks stampable.

non-blocking: if you want, dismiss coderabbit's major with the constant-gate evidence above so it doesn't sit there looking live.

gomesalexandre (approved):
receipt received and it clears my blocker. thank you for going all the way to a real on-chain receipt fetched independently of the cli rather than pasting the cli's own claim back at me - `eth_getTransactionReceipt` on a public rpc is re-runnable by anyone reading this, which is the whole point of a receipt.

the receipt was captured at `ec6a555b` and the branch has since been rebased to `5cff92af`, so i re-drove the behaviour myself at the current head rather than trusting a receipt from a rebased-away sha. it holds.

## receipts, current head

two arms, arm A is the merge base `2b576412` with only `commands/agent.ts` + `agent/ask.ts` swapped in, arm B is head. same harness, same fixtures, `--config .config/vitest.cli.config.ts` both sides. driven end to end through `executeAgentAsk`.

```
                                          merge-base (2b576412)              head (5cff92af)
C1 pass-ctrl clean success                exit=0                             exit=0
C2 fail-ctrl ordinary safety block        exit=10 AGENT_TURN_BLOCKED         exit=10 AGENT_TURN_BLOCKED
V1 declined sign_tx                       exit=12, proposed=-,               exit=12, proposed="send 1 ETH to 0xd8dA"
                                          msg="Transaction not confirmed"    chain=Ethereum, msg="built but signing was
                                                                             not authorized..."
V2 declined sign_typed_data               exit=12, proposed=-                exit=12, proposed=payload, chain=-
V3 backend outcome=confirmation_required  exit=10 AGENT_TURN_BLOCKED         exit=12 CONFIRMATION_REQUIRED
V4 broadcast committed, then a decline    exit=13 BROADCAST_COMMITTED        exit=13 BROADCAST_COMMITTED
                                          hashes=1                           hashes=1
```

C1, C2 and V4's exit code + hash count are identical across both arms, which is what makes this a diff of your change rather than a harness that answers the same thing to everything. V1 and V2 are the fix. V3 is the outcome-code remap and it does not disturb the ordinary blocked turn (C2). the chain-less typed-data case in V2 matches what your tests pin.

cli suites at head: 756 pass across 51 files. the one load failure is `sendDeclinePty.test.ts` on an unbuilt `node-pty` locally, environmental, not yours.

## nit: the new message makes a turn-wide claim from a single tool call's code

```ts
'The transaction was built but signing was not authorized, so nothing was signed or broadcast.'
```

"nothing was signed or broadcast" is a statement about the whole turn, but the code it is keyed on describes one tool call. in V4 i built an envelope carrying `BROADCAST_COMMITTED` with a live tx hash **and** that message sitting in `original_error.message`, so the two halves of the same envelope contradict each other.

reachability is narrow and i want to be honest about it rather than dress it up: `onTxStatus` has one producer, `emitAndConfirmTx`, called from exactly one site (`session.ts:900`), which sits after the `if (declinedInAskMode) return` at `888` and needs a successful sign to get there. combined with the constant gate at `ask.ts:209` a broadcast-then-decline is not reachable through the cli gate today. the only door is the backend `outcome.code === 'confirmation_required'` mapping this pr introduces, and whether the backend can emit that alongside a committed broadcast is its call, not yours.

so this is a wording nit, not a defect i can demonstrate in production. scoping the claim to the transaction rather than the turn costs one word and removes the contradiction permanently:

> The transaction was built but signing was not authorized, so it was not signed or broadcast. Re-run with --yes to authorize signing.

still passes the `/nothing was signed|not.*broadcast/i` matcher in your new test.

## nit: the retained non-decline results have no consumer

`takeQueuedAskModeDeclines` keeps everything that is not a decline so "the next request still reports them as recent_actions", and there is a test pinning that retention. but `askMode` is only true in `executeAgentAsk`, which calls `ask.ask(message)` once and then exits the process, so nothing in-process ever reads them. at the merge base the recursion would have delivered them to the backend.

i did not chase whether `--session` reuse re-sends vault state as context on the next invocation, which would make this a non-event. flagging it so the retention is a deliberate choice rather than a leftover - if it is dead by design, one clause in the docstring saying so saves the next reader the same walk.

## ruled out so you do not re-chase it

- the 500-char cap. i had a probe row suggesting it was not applied and it was my harness at fault: driving `onProposedTransaction` directly bypasses `reportDeclinedSigning`, which is the sole producer of that callback and where the cap lives. it holds end to end, both consumers get the same capped string, applied once.
- `getPendingChain()` before `clearPendingTransaction()`, still correct after the rebase.
- coderabbit's mixed-turn finding. your walk in the comment matches what i get reading it independently, and my V4 attempt to construct that state needed a hand-built envelope rather than a reachable path.

the runtime nuance you flagged about `outcome.kind` coming back `success` on the decline path is the right thing to have written down. that is exactly the kind of thing someone keys on six months later.

good work. this is the strongest test discipline i have seen on this repo and the receipt is the shape i wish every fund-path pr had.

rcoderdev (approved):

PR #1614, opened 2026-07-28
feat(swap): let co-signers review a THORChain limit order

### LINKED ISSUE #1611: Limit-order co-signers see a generic send for RUNE and native-gas sources
## Problem

`buildLimitSwapKeysignPayload` attaches a `THORChainSwapPayload` **only** in the ERC20 source branch. The RUNE (`MsgDeposit`) and native-gas-asset (transfer to the Asgard inbound vault) branches return `{ toAddress }` with no swap payload, so `expectedToAmount` is accepted and then never used for those two branches.

The consequence is on the **joining device**, not the initiator. With no swap payload on the keysign payload, the join-side review classifies the transaction as a generic send. The co-signer approves:

- a transfer to self (RUNE) or to an Asgard vault address (native gas asset)
- an opaque `=<:...` memo
- generic send terms

They never see the buy asset, the payout destination, or the minimum received — the three things that actually define a limit order. Only the initiating device gets a human-readable review.

This diverges from the market swap path, which is quote-driven rather than source-branch-driven: `nativeSwapQuoteToSwapPayload` builds a `THORChainSwapPayload` for **every** native swap regardless of whether the source is RUNE, a gas asset, or a token. So market swaps already give co-signers full display metadata for exactly the source branches limit orders do not.

Raised by @johnnyluo reviewing vultisig/vultisig-windows#4470.

## Expected behaviour

A joining device reviewing a limit order sees the same order details as the initiator — buy asset, payout destination, and minimum received — for all three source branches, not just ERC20.

## Suggested approach

Either:

1. Attach display metadata for every source branch (the `expectedToAmount` already passed in is the memo's LIM in THORChain's 1e8 fixed point, which is what `getNativeSwapDecimals` expects), **or**
2. Have the join-side review explicitly recognise and decode `=<` limit memos.

Option 1 needs care. The builder's own documentation notes that the swap payload is not merely cosmetic for ERC20 — a token source signed *without* one falls through to a plain ERC20 transfer, dropping the memo and stranding tokens on the router. Adding a payload to the RUNE and native-gas branches must be verified not to alter their signing path, chain-specific derivation, or gas estimation. The market path attaching one for native sources is evidence this is workable, but it should be confirmed per branch rather than assumed.

Note that `assertNativeSwapReadyForBroadcast` also early-returns when there is no swap payload, so the same two branches additionally skip the broadcast-time destination re-check — the builder already calls this out and compensates with a sign-time destination halt gate for RUNE. Whichever approach is taken, that interaction should be considered.

## Acceptance criteria

- [ ] A joining device shows buy asset, destination, and minimum received for RUNE, native-gas, and ERC20 limit orders
- [ ] The displayed minimum received matches the LIM encoded in the signed memo
- [ ] Signing behaviour for the RUNE and native-gas branches is unchanged (same transaction bytes as before, aside from display metadata)
- [ ] Covered by tests across all three source branches

Ehsan-saradar:
## Description

Fixes #1611

Only ERC20-sourced limit orders carry a swap payload. A RUNE or native-gas order reached a *joining* device as a transfer to an opaque address with an opaque memo — generic send terms, no buy asset, no payout destination, no minimum received. Only the initiating device got a readable review.

`getKeysignLimitSwapOrder` decodes those terms from `keysignPayload.memo`, which every source branch carries.

## Why the memo, and not a swap payload

The issue offered two approaches. I started on the first — attach display metadata to the RUNE and native-gas branches — and abandoned it: **the swap payload is not display metadata.** Every signing-input resolver re-routes on its presence.

| Resolver | What changes when a payload is attached |
|---|---|
| UTXO | takes *both* amount and destination from the payload instead of `keysignPayload` |
| EVM | destination becomes `vaultAddress`; transfer amount becomes `fromAmount` |
| Tron | destination becomes `vaultAddress` |
| Cosmos | drops the tx-body memo — `txMemo: swapPayload ? '' : memo` |

For RUNE it is worse than a byte change: `assertNativeSwapReadyForBroadcast` would then resolve a `THOR` inbound row, and THORChain's own inbound list does not contain one, so the broadcast guard fails outright.

So option 1 means changing signed bytes on a working path to improve a display. Reading the memo is both safer and **strictly more trustworthy**: the memo is the exact string THORChain executes, so terms derived from it cannot disagree with the order being signed. A display field carried alongside it can — it is supplied by the initiating device, and a co-signer has no way to tell a mistaken one from a malicious one.

The rejected approach is documented in the builder's JSDoc so the next person doesn't re-derive it.

## What's in here

- **`parseLimitSwapMemo`** — decodes a `=<` memo into its terms. `assertLimitSwapMemo` now delegates to it, so the grammar has one implementation rather than a validator that can drift from a parser.
- **`getKeysignLimitSwapOrder`** — takes a keysign payload, returns target asset + chain, destination, minimum received, and expiry; `undefined` for anything that isn't a limit order, so callers branch on it directly. Fail-closed parsing, but a malformed memo reports as "not a limit order" rather than throwing — a caller reviewing an arbitrary payload wants a branch, and signing still rejects it because the builder parses the same memo.
- **Builder hardening** — `toAmountDecimal` now comes from the memo's LIM rather than the caller's `expectedToAmount`. That argument became an optional cross-check that throws on disagreement.

That last gate has a real backstory: it catches exactly the bug shipped in vultisig-windows#4470, where the LIM was rescaled into the target coin's own decimals. The order signed correctly while the co-signer saw a figure 100x low for a 6-decimal USDC target and 1e10x high for an 18-decimal ETH one. Nothing else catches that class of mistake, because the memo stays valid. There's a test asserting it.

## Which feature is affected?

- [x] Swap — THORChain limit orders (`=<` advanced swap queue), co-signer review path

No transaction link: this changes what a joining device *displays*, and deliberately leaves signed bytes untouched on every branch. The signing path itself was verified on mainnet from the windows side — [`5CB3698C…40F9C3`](https://runescan.io/tx/5CB3698C77FC719202EB1AEE5C5060B12A86E0BC086B0BB0DCC176711640F9C3), a RUNE→USDC limit order that filled exactly at its LIM.

## Checklist

## Testing

`yarn test:core` — 2244 passed. `yarn check:agent` clean apart from two pre-existing `getVultisigConfigDir` errors also present on `main`.

New coverage: parser terms and bigint LIM handling; `getKeysignLimitSwapOrder` across all three source branches, non-limit memos, malformed memos, unroutable asset prefixes, and unknown intervals; builder tests for memo-derived display and the `expectedToAmount` cross-check, including the rescaled-to-target-decimals case specifically.

Also exercised against vultisig-windows with this branch patched into `node_modules`: an integration check confirmed the app's computed value equals the memo's LIM across five pairs, so the new gate does not reject legitimate orders.

## Additional context

The consuming UI is a separate change in vultisig-windows — this PR supplies the data; the join review renders it.

One behaviour worth flagging for review: ERC20-sourced limit orders previously rendered the generic swap review and will now render as limit orders on any client that adopts `getKeysignLimitSwapOrder`. That is intentional — all three source branches then review identically — but it is a visible change to a path that already worked.

## Summary by CodeRabbit

* **New Features**
  * Device co-signers now process THORChain limit-order memos using accurate swap-order semantics and computed minimum-received amounts.
  * Added memo parsing utilities (including affiliate support) and exposed limit-order memo/keysign decoding through the SDK.
* **Bug Fixes**
  * Prevented signing and display mismatches from decimal-rescaling and memo/expected-value disagreements.
* **Refactor**
  * Centralized limit-swap memo grammar by routing validation through the shared parser.
* **Tests**
  * Added coverage for memo parsing, keysign order decoding, and minimum-received display/cross-check behavior.

### DIFF
--- a/.changeset/limit-order-cosigner-review.md
+++ b/.changeset/limit-order-cosigner-review.md
@@ -1,0 +1,15 @@
+---
+'@vultisig/core-chain': minor
+'@vultisig/core-mpc': minor
+'@vultisig/sdk': minor
+---
+
+Let joining devices review a THORChain limit order instead of approving it as a generic send.
+
+The limit-order builder attaches a swap payload only for ERC20 sources, so RUNE and native-gas-asset orders reached a co-signer as a transfer to an opaque address with an opaque memo — no buy asset, payout destination, or minimum received.
+
+`getKeysignLimitSwapOrder` decodes those terms from `keysignPayload.memo`, which is present on every source branch. Reading the memo also makes the review trustworthy rather than merely present: the memo is the exact string THORChain executes, so terms derived from it cannot disagree with what gets signed, whereas a display field supplied by the initiating device can.
+
+`parseLimitSwapMemo` is the underlying decoder, and `assertLimitSwapMemo` now delegates to it so the grammar has one implementation.
+
+`buildLimitSwapKeysignPayload` now derives the ERC20 branch's `toAmountDecimal` from the memo's LIM rather than from the caller's `expectedToAmount`. That argument is now an optional cross-check: supplying a value that disagrees with the memo throws, which catches a caller that rescaled the LIM into the target coin's own decimals — a mistake that signs a correct order while showing a co-signer a wrong figure.
--- a/packages/core/chain/swap/native/limitSwapMemo.assertMemo.test.ts
+++ b/packages/core/chain/swap/native/limitSwapMemo.assertMemo.test.ts
@@ -86,4 +86,3 @@ describe('parseLimitSwapMemo', () => {
     expect(`=<:${targetAsset}:${destinationAddress}:${limit}/${intervalBlocks}/${quantity}`).toBe(validMemo)
   })
 })
-
--- a/packages/core/chain/swap/native/limitSwapMemo.assertMemo.test.ts
+++ b/packages/core/chain/swap/native/limitSwapMemo.assertMemo.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 
-import { assertLimitSwapMemo } from './limitSwapMemo'
+import { assertLimitSwapMemo, parseLimitSwapMemo } from './limitSwapMemo'
 
 const validMemo = '=<:ETH.ETH:0x742d35Cc6634C0532925a3b844Bc454e4438f44e:1600000000/14400/0'
 const validWithAffiliate = `${validMemo}:v0:50`
@@ -51,3 +51,39 @@ describe('assertLimitSwapMemo', () => {
     expect(() => assertLimitSwapMemo(`${validMemo}::50`)).toThrow(/empty affiliate segment/)
   })
 })
+
+// assertLimitSwapMemo delegates here, so every rejection above is this parser's
+// too. These cover what it *returns* — the terms a joining device reviews.
+describe('parseLimitSwapMemo', () => {
+  it('returns the order terms without an affiliate', () => {
+    expect(parseLimitSwapMemo(validMemo)).toEqual({
+      targetAsset: 'ETH.ETH',
+      destinationAddress: '0x742d35Cc6634C0532925a3b844Bc454e4438f44e',
+      limit: 1_600_000_000n,
+      intervalBlocks: 14_400,
+      quantity: 0,
+    })
+  })
+
+  it('returns the affiliate terms when present', () => {
+    expect(parseLimitSwapMemo(validWithAffiliate)).toMatchObject({
+      affiliate: 'v0',
+      affiliateBps: 50,
+    })
+  })
+
+  // The LIM is the order's price floor; a Number round-trip would lose precision
+  // above 2^53, which is reachable for an 18-decimal target.
+  it('keeps the LIM as a bigint', () => {
+    const big = '=<:ETH.ETH:0x742d35Cc6634C0532925a3b844Bc454e4438f44e:99999999999999999999/14400/0'
+
+    expect(parseLimitSwapMemo(big).limit).toBe(99_999_999_999_999_999_999n)
+  })
+
+  it('round-trips the memo it was given', () => {
+    const { targetAsset, destinationAddress, limit, intervalBlocks, quantity } = parseLimitSwapMemo(validMemo)
+
+    expect(`=<:${targetAsset}:${destinationAddress}:${limit}/${intervalBlocks}/${quantity}`).toBe(validMemo)
+  })
+})
+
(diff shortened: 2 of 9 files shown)


gomesalexandre (approved):
Approving. Nice piece of work, and the mutation evidence is the strongest part of it.

## why the co-signer surface deserved the scrutiny

In a two-signature flow the co-signer is the security control. Whatever they are shown IS the check, so if their view can diverge from what actually gets signed, the second signature degrades into a rubber stamp on a description. That is the lens I reviewed this under: does every displayed term decode from the payload, or is any of it passed alongside it.

It decodes. `parseLimitSwapMemo` reads the terms out of the memo itself rather than trusting a summary, and the co-signer now sees real order terms across all three source branches instead of only ERC20.

## the mutation is the finding

Reverting the `expectedToAmount` cross-check and the `toAmountDecimal` source does not just fail a test. It silently produces `toAmountDecimal: "160000000000.00000000"` where the correct value is 16, a display **ten billion times too high**, which is precisely the production incident this PR cites. The three named tests that go red are:

- `derives toAmountDecimal from the memo LIM`, on `expected '0.00000000' to be '16.00000000'`
- `rejects an expectedToAmount that disagrees with the memo LIM`, which resolves instead of rejecting
- `rejects a LIM rescaled to the target coin decimals`, the 1e10 case above

Restore, 52/52 across the three touched files. Both directions are pinned: `ignores the caller when it agrees with the memo` covers the legitimate path, so the guard is not simply refusing everything.

## fail-closed

Confirmed rather than assumed. A non-limit memo, an empty memo, or any limit memo that fails to parse returns undefined and produces no enriched card at all, falling back to the generic payload review. It never renders a blank or defaulted card that reads as legitimate. Unrecognized asset prefixes and intervals resolve individually to undefined rather than to a wrong guess.

The new signing-time gate is strictly more restrictive than before, never looser, which is the right direction for a change in this position.

Full `check:ci` chain green end to end, including the architecture, duplication and contract gates, and `test:core` at 2244 passed matching your stated number exactly.

## findings

`should-fix:` `destinationAddress` comes back from `parseLimitSwapMemo` with no format or chain validation, while the construction path validates it. So a memo carrying a garbage or wrong-chain destination decodes cleanly and is presented to the co-signer as a fully legitimate order. This is carried over verbatim from the pre-existing grammar rather than introduced here, and the practical severity is limited because an unroutable destination generally refunds rather than being lost to an attacker. But this is exactly the surface being reviewed, so it is worth tightening while you are in here.

`q:` the new cross-check is a hard throw, which is correct, but it changes an un-migrated caller passing a wrong-scale value from "displays wrong" to "hard fails". Since this package is consumed by the flagship clients as well as the agent app, that wants sequencing at release time rather than discovering it as an apparent regression. I could not verify the out-of-repo integration check from here, so flagging rather than doubting it.

`suggestion:` a co-signer still has no way to independently confirm that the payout destination is a currently live inbound vault, since that lookup happens only on the initiating device at build time. Pre-existing and systemic across THORChain deposit swaps rather than anything this introduces, but it is the one gap in the co-signer's independent verification that this PR does not close.

`praise:` refactoring the assertion into a parser and having the assertion delegate to it is a clean single-source-of-truth move, and the diff confirms the validation logic moved verbatim with nothing quietly dropped. Documenting why the tempting alternative was rejected, resolver by resolver, is the kind of trail that saves the next person a day.

PR #1631, opened 2026-07-29
Fix SDK storage examples

rcoderdev:
## Summary by CodeRabbit

* **Documentation**
  * Updated the SDK storage guidance to clarify that `storage` is optional and uses the platform’s persistent default when omitted.
  * Refreshed TypeScript examples to match the current `FileStorage` construction style and correct platform-specific imports.
  * Simplified password and balance caching snippets by removing explicit `storage` wiring where not needed.
* **Tests**
  * Added a unit test that validates the SDK users guide “storage contract” by checking fenced TypeScript examples for correct imports and `new FileStorage(...)` argument shapes, and verifying the documented Node example behavior.


johnnyluo (approved):
LGTM

PR #1632, opened 2026-07-29
fix(sdk): normalize StakeKit network aliases consistently

### LINKED ISSUE #1523: stakekit: balances() bypasses the search() network alias normalizer
## Summary
`stakekitBalances()` bypasses the same network alias normalization that `stakekitSearch()` already applies.

## Evidence
- `packages/sdk/src/tools/defi/stakekit/index.ts:426-436` maps user-facing aliases like `bsc` / `avalanche` to StakeKit slugs.
- `packages/sdk/src/tools/defi/stakekit/index.ts:495-500` lowercases in `stakekitBalances()` but does not reuse the alias map.
- `packages/sdk/src/tools/defi/stakekit/stakekitApi.ts:254-263` forwards that network slug straight into `searchYields()`.

## Impact
The same public feature family can disagree with itself:
- `sdk.defi.stakekit.search({ network: 'BSC' })` can normalize to StakeKit's expected slug.
- `sdk.defi.stakekit.balances({ network: 'BSC' })` can still query with `bsc`.

Downstream consumers either learn the mismatch the hard way or re-implement the alias table locally.

## Fix sketch
Extract one canonical StakeKit network normalizer and reuse it in `search`, `balances`, and any related helpers.

## Effort
Small to medium.

## Report
https://gist.github.com/gomesalexandre/4508de07dabd8e40384f1eecdd89c2ca

rcoderdev:
Closes #1523

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized StakeKit network alias handling across both search and balance requests.
  * Ensured common aliases (for example, BSC and AVAX) are normalized to the canonical values expected by the StakeKit backend.

* **Tests**
  * Added unit test coverage to verify the outgoing request URLs use the canonical network query values for both search and balance.

### DIFF
--- a/.changeset/brave-stakekit-aliases.md
+++ b/.changeset/brave-stakekit-aliases.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Normalize StakeKit network aliases consistently across search and balance reads.
--- a/packages/sdk/src/tools/defi/stakekit/index.ts
+++ b/packages/sdk/src/tools/defi/stakekit/index.ts
@@ -409,6 +409,20 @@ async function resolveActionArgs(
 
 // --- Builder functions ---
 
+const STAKEKIT_NETWORK_ALIASES: Readonly<Record<string, string>> = {
+  bsc: 'binance',
+  'bnb chain': 'binance',
+  'bnb-chain': 'binance',
+  bnbchain: 'binance',
+  avalanche: 'avalanche-c',
+  avax: 'avalanche-c',
+}
+
+const normalizeStakekitNetwork = (network: string): string => {
+  const normalized = network.toLowerCase()
+  return STAKEKIT_NETWORK_ALIASES[normalized] ?? normalized
+}
+
 /**
  * Search yield opportunities. Wraps searchYields with client-side token/provider filtering.
  * apiKey is injectable; omit for unauthenticated read-only access.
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
Approving. Clean read-path normalization, closes the issue exactly as specced, tests pinned at the right boundary.

## verified

Two mutations, both killed by named tests: reverting the balances hunk fails `uses the same canonical network aliases as search` with `expected 'avax' to be 'avalanche-c'`; neutering the shared normalizer fails both new tests on search and balances alike. Restore, 18/18. Both tests assert the outgoing request URL's query param, not an echoed value - the correct boundary for a query-path fix.

No wrong-network misattribution risk: the alias keys aren't themselves valid StakeKit slugs, and the targets are confirmed canonical against the module's own inverse map. Balances are keyed by address plus integration id, so a bad slug can only under-enumerate (empty result), never return another chain's holdings - pre-PR this was a guaranteed false negative for BSC/AVAX balances, post-PR it queries correctly. The map is also idempotent, no double-normalization risk.

`should-fix:` the map misses `cronoschain`/`cronos chain` to `cronos` - the SDK's own canonical chain id for Cronos round-trips wrong through this table. Pre-existing on both paths (search had the same hole on main), one-line fix. Filing separately.

`suggestion:` the SDK's alias table is a strict subset of abts's richer, catalog-verified one, and the SDK doesn't trim whitespace where abts does. Follow-up material, not this PR's scope.

`praise:` exact fix-sketch execution - one shared normalizer, both call sites, tests at the wire boundary, changeset present.

PR #1641, opened 2026-07-29
fix(cli): restore EVM gas refresh visibility

neavra:
## Summary

- replace unavailable Ethereum and Polygon gas-refresh RPCs with the workspace-standard PublicNode endpoints
- report HTTP, JSON-RPC, and malformed base-fee refresh failures to stderr before signing while preserving the original estimate and continuing non-fatally
- treat failed pending-nonce responses as unverifiable so the existing large nonce-gap safeguard remains effective
- distinguish missing or null results from legitimate zero nonces, preserving both `result: "0x0"` and numeric `result: 0` as `0n`
- add mocked signing-path regression coverage for gas refresh and pending-nonce failure/success behavior

## Verification

- live read-only `eth_blockNumber` probes: all 13 configured endpoints returned HTTP 200 after the replacements
- `yarn install`
- `yarn build:shared`
- `yarn build:sdk`
- `yarn build:client-shared`
- `yarn build:rujira`
- `yarn cli:build`
- `yarn typecheck`
- `yarn test:cli` — 71 files, 968 tests passed
- `yarn quality:complexity`
- changed-file ESLint
- changed-file Prettier check with the repository config
- mutation checks: restoring the old pending-nonce fallback makes each new test fail on response behavior, not an RPC URL
- all test RPC calls are mocked; no test makes a network request

## Related work

Draft PR #1529 touches the same hardcoded EVM RPC map. If both changes land, the second merge will need to resolve that overlap; #1529's canonical `getEvmRpcUrl` resolver supersedes this hardcoded map.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated public RPC endpoints for Ethereum and Polygon networks.
  * Added warnings when gas refresh or pending-nonce verification fails.
  * Improved nonce fallback behavior while retaining safeguards against large nonce gaps.
  * Preserved original gas estimates when refreshed values are unavailable or invalid.
  * Enhanced validation of RPC responses to prevent invalid data from being used.

* **Tests**
  * Added comprehensive coverage for gas refresh and pending-nonce scenarios.


gomesalexandre (approved):
Approving. Reviewed at the current head; the nonce handling is the part that matters and it is right.

## the subtle bit, verified

```ts
if (!res.ok || data?.error || data?.result === undefined || data?.result === null) {
  return null
}
return BigInt(data.result)
```

Checking `=== undefined || === null` rather than falsiness is load-bearing, and it would be very easy to "simplify" into a bug. A fresh account's pending nonce is legitimately `0`, and JSON-RPC may return it as either `"0x0"` or numeric `0`:

```
 case                       shipped     naive(!result)   want
 hex zero (fresh account)   0           0                0
 numeric zero               0           null             0     <-- naive breaks here
 hex nonzero                5           5                5
 missing result             null        null             null
 explicit null              null        null             null
 rpc error                  null        null             null
```

A naive `if (!data?.result)` returns `null` for numeric `0`, which this code treats as *unverifiable* — so on a fresh account the nonce-gap safeguard would fire on the very first transaction of every new address. Getting the zero case right is what keeps that safeguard meaningful rather than noisy.

## the rest

- **Failed pending-nonce → unverifiable, not zero.** Correct direction: it keeps the local nonce and warns rather than assuming a value, and the tests assert `clearEvmState` is *not* called on the unverifiable path — so a transient RPC failure cannot silently reset nonce state. That is the fail-safe choice.
- **Gas-refresh failures are non-fatal and preserve the original estimate.** Right call. Refusing to sign because a base-fee refresh 500'd would be a direction-B failure — blocking a legitimate transaction over a telemetry-grade problem. Reporting to stderr before signing gives the user the information without taking the decision away.
- **Warning text names the chain and the nonce being used** (`"...will continue with local nonce 3"` / `"...with on-chain nonce 1"`), which is the difference between a warning someone can act on and noise.

Good disclosure on the #1529 overlap — flagging that its canonical `getEvmRpcUrl` resolver supersedes this hardcoded map, and that whichever lands second owns the conflict, is exactly the note that saves the next person twenty minutes. Same situation I hit today with two of my own PRs editing one regex.

`yarn test:cli` 968 tests reported passing, and the mutation check described (restoring the old fallback fails the new tests on response behaviour rather than on an RPC URL) is the right way to prove the tests bite — it isolates the assertion from the endpoint change.

🤖 reviewed via Claude Code

PR #1649, opened 2026-07-30
Preserve hashes after partial bundled broadcasts

### LINKED ISSUE #941: fix(broadcast): multi-input broadcast is not atomic — loses hash of sent inputs on partial failure
## Evidence
`packages/sdk/src/vault/services/BroadcastService.ts:104-131` — the loop broadcasts each `txInputData` sequentially, keeping only the last `txHash`. If input[0] (the approve) is accepted and input[1] (the swap) then throws, the whole method throws `BroadcastFailed` and the successful input[0] hash is discarded.

## Impact
For approve+swap bundles, a failure on the second broadcast surfaces as total failure even though the first tx is irreversibly on-chain; the caller has no hash for the sent approve and, on retry, re-broadcasts it (double approve / wasted gas). No idempotency tracking across the loop.

## Fix sketch
Collect per-input results and, on partial failure, throw an error carrying the already-broadcast hashes so callers don't re-send the earlier input.

---

rcoderdev:
Closes #941

## Summary by CodeRabbit

- **New Features**
  - Preserve already-broadcast transaction hashes when a later bundled broadcast fails.
  - Expose `BroadcastPartialFailureError`, including the chain, successfully broadcast hashes, and the index of the failed broadcast.
  - Include per-message recovery metadata for multi-signature, multi-message transactions.

- **Bug Fixes**
  - Correct recovery ID derivation to respect individual signature recovery data in multi-message cases.

- **Tests**
  - Added/updated unit tests for partial broadcast failures, public exports, and per-message recovery formatting.

### DIFF
--- a/.changeset/broadcast-partial-hashes.md
+++ b/.changeset/broadcast-partial-hashes.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Preserve already-broadcast transaction hashes when a later bundled broadcast leg fails.
--- a/packages/sdk/src/adapters/formatSignature.ts
+++ b/packages/sdk/src/adapters/formatSignature.ts
@@ -62,13 +62,17 @@ export function formatSignature(
     format: signatureFormat,
   }
 
-  // For UTXO chains with multiple inputs, include all signatures
+  // Preserve per-message metadata for UTXO inputs and bundled transactions.
   if (messages.length > 1) {
-    signature.signatures = messages.map(msg => ({
-      r: signatureResults[msg].r,
-      s: signatureResults[msg].s,
-      der: signatureResults[msg].der_signature,
-    }))
+    signature.signatures = messages.map(msg => {
+      const result = signatureResults[msg]
+      return {
+        r: result.r,
+        s: result.s,
+        der: result.der_signature,
+        ...(result.recovery_id ? { recovery: parseInt(result.recovery_id) } : {}),
+      }
+    })
   }
 
   return signature
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -19,6 +19,7 @@ export { Vultisig } from './Vultisig'
 // Vault management
 export type { VaultConfig } from './vault'
 export {
+  BroadcastPartialFailureError,
   FastVault,
   SecureVault,
   VaultBase,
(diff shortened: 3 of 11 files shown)


gomesalexandre (approved):
Approving. The failure this fixes is the broadcast-false-negative shape, which is the one I care most about on a bundled path.

## why the bug mattered

On a bundled broadcast (approve plus swap), an earlier leg succeeding and a later leg failing used to throw a generic error and drop the successful leg's hash entirely. The caller had no way to learn that an approve had already landed on chain. That is worse than a plain failure: the user is told the operation failed while an approval is live, and any retry stacks a second one. Carrying `broadcastedTxHashes` plus `failedInputIndex` on a dedicated `BroadcastPartialFailureError` is the right shape - the partial success becomes visible rather than inferred.

## verified

Mutation on the fix hunk (removing the try/catch and hash tracking) fails the named test `carries already-broadcast hashes when a later input fails`, falling back to a plain Error with no preserved hash. Restore, 9/9, re-verified twice. The companion signature work has real two-directional coverage too (52 assertions across `convertSignature.test.ts` and `formatSignature.test.ts`), so it is not a dead guard.

Fail-closed confirmed: this is throw-only. No path returns a `txHash` as though the whole bundle succeeded on a partial failure, so the fix cannot itself manufacture a false success while trying to report one.

`praise:` `convertSignature.ts:89` using `sig.recovery ?? signature.recovery` rather than `||` is a genuinely good catch. A `||` there would silently discard a per-leg `recovery: 0`, which is a valid recovery id, and that class of bug produces a malformed signature rather than an obvious error.

`should-fix:` the preserved hash currently dead-ends at the SDK boundary. Nothing was checked downstream about whether agent-backend-ts or the app actually catches `BroadcastPartialFailureError` and surfaces `broadcastedTxHashes` to the user. If no consumer reads it, the information is captured and then discarded, which leaves the user in the same position as before. Out of scope for this PR, worth a fast follow so the fix reaches the surface it was written for.

`q:` the positional index mapping between message hashes and signature results is pre-existing and untouched here, but it is fragile to a future ordering change - an off-by-one there would attribute a hash to the wrong leg, which is exactly the kind of silent mis-association this PR exists to prevent. An assertion or comment pinning the invariant would be cheap insurance.

PR #1650, opened 2026-07-30
fix: validate Bittensor balance addresses

### LINKED ISSUE #946: correctness: money-path edge-case cluster (swap net-vs-gross, balance false-zeros, decimals)

rcoderdev:
Closes #946

Reject non-Bittensor SS58 network prefixes and invalid checksums before querying account storage.

Refs #946

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened Bittensor SS58 address validation by verifying expected network prefix and checksum before reading TAO balances.
  * Addresses that are invalid, incompatible, or have incorrect checksums are rejected without making any RPC request.
  * More robust behavior for empty or failing balance lookups (returning zero for null results and surfacing RPC errors clearly).
* **Tests**
  * Added coverage for Bittensor balance resolution, including valid address handling, rejection cases, null responses, and RPC error propagation.

### DIFF
--- a/.changeset/bittensor-address-validation.md
+++ b/.changeset/bittensor-address-validation.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Validate Bittensor SS58 prefix and checksum before reading TAO balances.
--- a/packages/core/chain/coin/balance/resolvers/bittensor.test.ts
+++ b/packages/core/chain/coin/balance/resolvers/bittensor.test.ts
@@ -1,0 +1,101 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const queryUrlMock = vi.hoisted(() => vi.fn())
+
+vi.mock('@vultisig/lib-utils/query/queryUrl', () => ({
+  queryUrl: queryUrlMock,
+}))
+
+import { getBittensorCoinBalance } from './bittensor'
+
+// Alice's SS58-42 address and its 32-byte public key.
+// Public key: d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d
+const VALID_TAO_ADDRESS = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'
+const POLKADOT_ADDRESS_SAME_ACCOUNT = '15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6Sp5'
+
+const u128LE = (value: bigint): string => {
+  const bytes = Array.from({ length: 16 }, (_, index) =>
+    ((value >> BigInt(index * 8)) & 0xffn).toString(16).padStart(2, '0')
+  )
+  return bytes.join('')
+}
+
+const buildSystemAccountHex = (free: bigint): string =>
+  '0x' +
+  '00000000' + // nonce
+  '00000000' + // consumers
+  '01000000' + // providers
+  '00000000' + // sufficients
+  u128LE(free) +
+  u128LE(0n) +
+  u128LE(0n) +
+  u128LE(0n)
+
+describe('getBittensorCoinBalance', () => {
+  beforeEach(() => {
+    queryUrlMock.mockReset()
+  })
+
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
Approving. Checked the SS58 implementation directly rather than taking the description on trust, since this sits on a balance path and the PR is auto-merge armed.

## the algorithm is right

- Prefix check against 42 is correct for Bittensor.
- The checksum is the standard SS58 construction: blake2b-512 over `"SS58PRE" || body`, comparing the first two bytes. Correct for a 32-byte public key with a 2-byte checksum.
- `ss58AddressByteLength = 35` is consistent: 1 prefix + 32 pubkey + 2 checksum.

## direction is fail-closed

Both checks run BEFORE any account-storage query, so a wrong-network or corrupted address is rejected rather than silently decoded and queried. That closes a genuinely confusing case: previously a same-keypair address carrying another chain's prefix would decode fine and return the Bittensor balance for that keypair, which is a real number answering a question the user did not ask. Rejecting is the better failure.

Additive only (122 lines, zero deletions) with tests covering valid, rejected, null-result and RPC-error paths, so the blast radius is bounded to this resolver.

`q:` the one thing I could not settle from the diff alone: how the throw propagates. If a single chain resolver throwing aborts a multi-chain portfolio fetch, then a user holding one malformed stored address loses their whole balance view rather than just the TAO line. Sibling resolvers elsewhere catch per-asset failures and continue, so I expect that is handled, but worth confirming - it is the only way this stricter validation could turn into a worse user outcome than the loose behavior it replaces.

`praise:` rejecting before the RPC rather than after is the right ordering. It means a bad address costs nothing and cannot produce a partially-populated result that reads as authoritative.

PR #1655, opened 2026-07-30
feat(ripple): resolve logos for XRPL issued currencies

### LINKED ISSUE #1654: [ADD] Resolve logos for XRPL issued currencies
**Is your feature request related to a problem? Please describe.**

XRPL issued currencies render with a broken-image placeholder instead of a logo.

`getRippleTokenMetadata` returns `logo` only for the tokens we curate (currently just RLUSD). Every other issued currency — SOLO, an issuer's USD, anything a user adds by id or that ledger discovery surfaces — resolves to `{ ticker, decimals }` with no logo, so clients fall back to a generic placeholder icon.

This is visible in vultisig-windows the moment a custom XRP token is added: the token resolves correctly and shows the right ticker, but sits next to an empty grey image box.

The gap is in the SDK, not the client: the client only renders the `logo` the resolver hands it, so there is nothing it can do locally.

XRPL has no on-ledger token metadata registry, which is why the resolver fetches nothing today. But other chains already solve this the same way — the EVM resolver reads a token's `logoURI` from 1inch, and the Solana resolver returns an `icon` — so XRPL is the outlier rather than a special case.

**Describe alternatives you've considered**

- **Curating more tokens in `rippleKnownIssuedTokens`** — does not scale and can never cover an arbitrary issuer/currency pair a user adds by hand.
- **Leaving the placeholder** — acceptable but visibly worse than every other chain, and it makes a correctly-added token look broken.

**Additional context**

- Whatever source is used must not let an arbitrary token borrow a curated token's identity: two issuers can share a ticker on XRPL, so the `(currency, issuer)` pair is what actually identifies a token.
- A metadata lookup should fail soft — an unlisted token or an unreachable registry should still resolve the token, just without a logo.
- Surfaced while reviewing vultisig/vultisig-windows#4459 (custom XRP tokens), which builds on #1567.

Ehsan-saradar:
## What

Gives XRPL issued currencies a logo. `getRippleTokenMetadata` now reads an uncurated token's official icon from the XRPL token registry (xrplmeta), keyed by the `<currency>:<issuer>` pair.

## Why

Only curated tokens (RLUSD) carried a logo, so every other trust-line token — SOLO, an issuer's USD, anything a user adds by id or that ledger discovery surfaces — resolved with no `logo` and rendered as a broken-image placeholder in clients. It shows up the moment a custom XRP token is added in vultisig-windows: right ticker, empty grey image box next to it.

This is the same shape as the EVM resolver reading `logoURI` from 1inch and the Solana resolver returning `icon`. XRPL was the outlier only because it has no on-ledger token metadata registry of its own — not because it should behave differently.

## Behaviour

- **Curated token** (RLUSD): unchanged — bundled logo, price provider, and **no lookup** at all.
- **Uncurated token**: may borrow the registry's icon, but **never** a curated token's `priceProviderId`. Two issuers can share a ticker on XRPL, so the issuer is what identifies a token; a lookalike must not inherit a real token's pricing.
- **Unlisted token or unreachable registry**: still resolves, just without a logo. A metadata lookup failing is not a reason to fail the token.

## Tests

`resolvers/ripple.test.ts` mocks `queryUrl` (so nothing hits the network) and covers: curated token keeps its bundled logo, curated token performs no lookup, uncurated token gets the registry icon, the lookup is keyed by issuer, an uncurated token never gains a price provider, and both the unlisted and unreachable cases still resolve.

`yarn build:shared`, `format:check`, `check:shared-exports`, the SDK bundle changeset guard and eslint all pass locally.

## Related

- Closes #1654
- Follows #1567 / #1568 (XRPL custom token support)
- Client side: vultisig/vultisig-windows#4459

## Summary by CodeRabbit

- **New Features**
  - Added automatic logo lookup for uncurated XRPL-issued tokens using the XRPL token registry.
  - Tokens continue to resolve successfully when no logo is available or the registry cannot be reached.
  - Curated Ripple tokens retain their existing logos and pricing metadata.
  - Uncurated tokens with registry icons now display the retrieved logo.

### DIFF
--- a/.changeset/ripple-token-logos.md
+++ b/.changeset/ripple-token-logos.md
@@ -1,0 +1,23 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+feat(ripple): resolve logos for XRPL issued currencies
+
+Only curated issued currencies (RLUSD) carried a logo, so every other trust-line
+token — SOLO, an issuer's USD, anything added by id or surfaced by ledger
+discovery — resolved with no logo and rendered as a broken-image placeholder in
+clients.
+
+`getRippleTokenMetadata` now reads an uncurated token's official icon from the
+XRPL token registry (xrplmeta), keyed by the `<currency>:<issuer>` pair. This is
+the same shape as the EVM resolver reading `logoURI` from 1inch and the Solana
+resolver returning `icon`; XRPL was the outlier because it has no on-ledger token
+metadata registry of its own.
+
+A curated token is unchanged: it keeps its bundled logo and price provider and
+performs no lookup. An uncurated token may borrow an icon but never a curated
+token's `priceProviderId` — two issuers can share a ticker on XRPL, so the issuer
+is what identifies a token. The lookup fails soft: an unlisted token or an
+unreachable registry still resolves, just without a logo.
--- a/packages/core/chain/coin/token/metadata/resolvers/ripple.test.ts
+++ b/packages/core/chain/coin/token/metadata/resolvers/ripple.test.ts
@@ -1,14 +1,29 @@
 import { Chain } from '@vultisig/core-chain/Chain'
 import { rippleIssuedCurrencyDecimals, rippleTokenId } from '@vultisig/core-chain/chains/ripple/issuedCurrency'
-import { describe, expect, it } from 'vitest'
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #1679, opened 2026-08-01
chore(deps): bump brace-expansion out of the oom-dos advisory range

Ehsan-saradar:
GHSA-mh99-v99m-4gvg (high, DoS via unbounded expansion length) was published against brace-expansion <1.1.17 and >=2.0.0 <2.1.3. Our lockfile had 1.1.16 and 2.1.2 (both transitive via minimatch, tooling only), so `quality:audit` now fails the Lint and Type Check job on every PR — see the runs on #1678.

`yarn up -R brace-expansion` moves them to 1.1.18 and 2.1.4. Lockfile-only change; `yarn npm audit --recursive --all --severity high` is clean locally. Same shape as the previous bump for GHSA-3jxr-9vmj-r5cp.

### DIFF
--- a/.changeset/vult-discount-tier-bigint.md
+++ b/.changeset/vult-discount-tier-bigint.md
@@ -1,0 +1,12 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #1686, opened 2026-08-02
feat(swap): limit-order cancellation primitives

### LINKED ISSUE #1685: Limit-order cancellation: modify-limit-swap memo, eligibility, and dust
## Problem

The SDK can place, sign, review and track a THORChain limit order, but not **cancel** one. An order rests for up to 3 days and there is currently no way to close it early — the user waits for expiry.

Cancellation is not a variation on placement. It uses a different memo type (`m=<`, modify-limit-swap) whose validation and matching rules are unusually unforgiving, and every failure mode is **silent**: the transaction confirms, the fee is spent, and the order carries on resting with nothing to distinguish it from success. That property should drive the whole design — fail closed at every unknown rather than emit bytes that might match nothing.

## The memo

```
m=<:<SRC_AMOUNT><SRC_ASSET>:<TRADE_TARGET><TGT_ASSET>:0
```

Each coin is `<amount><ASSET>` with **no space**. The trailing field is `ModifiedTargetAmount`; THORNode branches on `IsZero()` and reads zero as a cancel — any other value *re-targets* the order, a different action we should not build (but should be able to recognise, so a predicate named "is cancel" doesn't quietly match a retarget).

Verified against THORNode `x/thorchain/memo/memo_modify_limit_swap.go`, `handler_modify_limit_swap.go`, `keeper/v1/keeper_adv_swap_queue.go`.

## Four rules that each cause a silent no-op if broken

1. **Assets must be spelled in FULL.** `processOneTxIn` runs every other inbound memo through `fuzzyAssetMatch` — which is what lets a placement memo say `ETH.USDC-06EB48`. `ModifyLimitSwapMemo` is the exception: its asset string builds the lookup key verbatim. Repeating the placement's 6-character abbreviation addresses a bucket that by construction holds nothing. This should be *enforced* by the builder, not documented.

2. **A secured-asset source must use its secured denom** (`eth-usdc-0xa0b…`), never the layer-1 form. `MsgModifyLimitSwap.ValidateBasic` enforces `From.IsChain(Source.Asset.GetChain())`, and `GetChain()` returns THORChain only for synth/trade/secured assets. The L1 spelling makes it report `ETH`, and a cancel sent from a THOR address is rejected at validation.

3. **Amounts are plain decimal integers.** The placement memo's LIM goes through `getUintWithScientificNotation` and understands `544e6`; these coins go through `cosmos.ParseCoins`, which does not. A compressed amount fails to parse.

4. **Both amounts must be exact**, because THORNode does not compare them directly — it builds an index key from the *ratio* and scans that bucket for a matching `FromAddress`, taking the first hit:

   ```
   ratio = (sourceAmount × 1e8) / tradeTarget          // integer division
   key   = "<l1Source>><l1Target>/<ratio, zero-padded or truncated to 18 chars>/"
   ```

   One unit of drift lands in a different bucket. The 18-character normalisation matters at both extremes: short ratios are zero-padded so keys sort numerically, long ones are **truncated from the right**, which deliberately collapses very large ratios into one bucket.

## What's needed

**1. Memo builder + predicates** — build the cancel form from exact inputs; reject empty assets, non-positive amounts, and abbreviated token identifiers. Plus `isModifyLimitSwapMemo` / `isCancelLimitSwapMemo`, the latter comparing the final field *numerically* (`"00"` is zero to THORNode, and a string compare would call it a retarget).

**2. Eligibility, failing closed** — decide whether an order can be cancelled *before* offering the action, with a distinct reason per blocker: already terminal; a cancel already broadcast; source chain not routable; memo too long for the source chain's byte cap; and **signed data disagreeing with the chain's own report**.

That last one earns its place: cross-check the recorded amounts *and assets* against the queue's `state.deposit` / `trade_target` / observed assets. Absence is not disagreement (an order placed seconds ago hasn't been polled), but present-and-unparseable must block exactly as a mismatch does. A known real-world failure here had the amounts cross-checked and agreeing while the assets were never compared — and the asset was the entire defect.

**3. Asset spelling resolution** — three sources in decreasing authority: the queue's observed string (post-`fuzzyAssetMatch`, authoritative), the full form captured at signing, and the stored placement string *only* when it carries no truncated identifier (which covers every native leg and secured denom). Disagreement between a local spelling and the chain's blocks; no usable spelling blocks. Compare case-insensitively — this is a convention difference, not a real one.

**4. Dust for L1-sourced cancels** — a cancel from an L1 chain must attach enough for Bifrost to observe it, derived from that chain's `dust_threshold` and converted into the source coin's own precision. Fail closed when the threshold is missing or unparseable rather than defaulting.

A second known real-world failure belongs in the tests: a cancel signed for **2000 wei** — a 1e8-unit threshold used verbatim as an 18-decimal chain's smallest unit — which `ConvertAmount` truncates to zero. Bifrost never saw an inbound, the transaction confirmed, gas was burned, the order kept resting. Also worth an upper bound: `dust_threshold` is a remote value that decides how much is irreversibly donated, with no refund path.

**5. Duplicate detection** — orders sharing a bucket key are not independently cancellable, and THORNode takes the first match. Compare on the reproduced key (ratio-based, layer-1-normalised assets), not on equal amounts: two orders with different deposits and targets collide whenever the ratio matches. Tune to over-report.

## Scope

Out of scope: the modify/retarget action (only the cancel form), and all client UI — this issue is the SDK primitives, consumed later by vultisig-windows.

## Acceptance criteria

- [ ] Cancel memos build only from proven-exact inputs; abbreviated assets, non-positive amounts and empty assets are rejected structurally
- [ ] Secured-asset sources emit the secured denom; L1 sources emit the L1 form
- [ ] Eligibility fails closed at every unknown, with a distinct reason per blocker
- [ ] Signed-vs-observed cross-check covers **assets as well as amounts**; absent ≠ disagreeing, unparseable blocks
- [ ] Dust derives from the live `dust_threshold` in the source coin's precision, with lower and upper bounds; no silent defaulting
- [ ] Bucket-key reproduction matches THORNode's padding **and** truncation behaviour
- [ ] Both documented live failures are pinned by regression tests
- [ ] Exported from the root entry with a changeset

Ehsan-saradar:
## Description

Fixes #1685

A limit order rests for up to three days with no way to close it early. This adds the primitives for cancelling one: the `m=<` modify-limit-swap memo in its cancel form, eligibility, bucket-key duplicate detection, and L1 dust.

**Cancellation is not a variation on placement**, and that shaped every decision here. It uses a different memo type whose validation and matching rules are unusually unforgiving, and — the property worth holding onto while reviewing — **every failure mode is silent**: the transaction confirms, the fee is spent, and the order carries on resting with nothing to distinguish it from success. There is no error to surface and no refund path. So each rule below is *enforced* rather than documented, and every unknown fails closed.

Verified against THORNode `x/thorchain/memo/memo_modify_limit_swap.go`, `handler_modify_limit_swap.go`, `keeper/v1/keeper_adv_swap_queue.go`.

## The memo

```
m=<:<SRC_AMOUNT><SRC_ASSET>:<TRADE_TARGET><TGT_ASSET>:0
```

Coins are `<amount><ASSET>` with no space. The trailing field is `ModifiedTargetAmount`; THORNode branches on `IsZero()` and reads zero as a cancel.

Three properties, each a silent no-op if broken:

- **Assets are spelled in FULL.** `processOneTxIn` runs every other inbound memo through `fuzzyAssetMatch` — which is exactly what lets a *placement* memo say `ETH.USDC-06EB48`. `ModifyLimitSwapMemo` is the exception: its asset string builds the lookup key verbatim, so repeating the abbreviation addresses a bucket that by construction holds nothing. `buildCancelLimitSwapMemo` throws rather than letting those bytes exist.
- **Amounts are plain decimal integers.** A placement LIM goes through `getUintWithScientificNotation` and understands `544e6`; these coins go through `cosmos.ParseCoins`, which does not.
- **A secured source keeps its secured denom.** `ValidateBasic` enforces `From.IsChain(Source.Asset.GetChain())`, and `GetChain()` returns THORChain only for synth/trade/secured assets — the layer-1 spelling makes it report the L1 chain, and a cancel sent from a THOR address is rejected at validation.

`isCancelLimitSwapMemo` is kept distinct from `isModifyLimitSwapMemo`, and compares the final field **numerically**: `"00"` is zero to THORNode's `getUint`, so a string comparison would call it a retarget. Only the cancel form is built, so today the two predicates coincide — but a predicate named "cancel" that matches any modification is the kind of small lie that outlives the assumption making it true.

## Bucket key — why both amounts must be exact

THORNode never compares the amounts directly. It builds an index key from their *ratio* and scans that bucket for a matching `FromAddress`, taking the first hit:

```
ratio = (sourceAmount × 1e8) / tradeTarget          // integer division
key   = "<l1Source>><l1Target>/<ratio normalised to 18 chars>/"
```

One unit of drift lands in a different bucket. `getThorchainLimitOrderBucketKey` reproduces the normalisation in both directions — zero-padding *and* right-truncation — because truncation deliberately collapses very large ratios into one bucket, and reproducing only the padding would disagree with the chain at exactly the extreme where it matters.

Duplicate detection compares **keys, not amounts**: two orders with different deposits and different targets collide whenever their ratio matches (selling 1 and selling 2 at the same price land in one bucket). Comparing amounts for equality would under-report precisely the duplicates a user most needs warning about. Assets are layer-1-normalised first, since a secured representation and the plain L1 asset collapse to the same key on-chain. Tuned to over-report.

## Worth a reviewer's attention: two real failures pinned by tests

**The assets cross-check.** `getLimitSwapCancelEligibility` compares what was recorded at signing against what the queue reports — **assets as well as amounts**. There is a known failure where the amounts were cross-checked and agreed, the assets were never compared at all, and the asset was the entire defect. A check that verifies everything except the field that broke reports "all clear" on the way to a rejected transaction.

Absence is *not* disagreement — an order placed seconds ago hasn't been polled, and refusing until the first poll lands would be a worse failure than the one this prevents. But present-and-unparseable blocks exactly as a mismatch does: "not polled yet" and "polled, and the wire carried something we don't model" are different claims.

**The dust conversion.** A cancel was once signed for **2000 wei** — THORChain's 1e8-unit threshold used verbatim as an 18-decimal chain's smallest unit — which `ConvertAmount` truncates to zero. Bifrost never saw an inbound, the transaction confirmed, gas was burned, the order kept resting. `getLimitSwapCancelDust` rescales into the source coin's own precision and **refuses** when the result rounds away, rather than nudging up to the bare observable minimum, which would be the same silent failure one order of magnitude up. It also carries an upper ceiling: `dust_threshold` is a remote value that decides how much is irreversibly donated.

## Which feature is affected?

- [x] Swap — THORChain limit-order cancellation (pure primitives; nothing here signs or broadcasts)

No tx link: this PR builds and validates memo bytes and decides eligibility. Nothing constructs a keysign payload or broadcasts yet — that is the follow-up, and the first live cancel belongs there.

## Checklist

## Testing

`yarn test:core` — 2422 passed, 67 of them new.

Coverage targets where a wrong answer is unrecoverable: abbreviated assets rejected in both legs; secured denoms passed through unchanged; `BTC.BTC`/`THOR.RUNE`/`btc-btc` correctly *not* read as truncated (a naive "text after the last `-`" rule would make every secured-native order uncancellable); zero-padded and signed final fields; ratio padding and truncation at both extremes; ratio collisions across differing amounts; every eligibility blocker; the assets-disagree case with matching amounts; and the 2000-wei dust regression asserting explicitly that the result is **not** `2000n`.

`yarn check:agent` reports two pre-existing CLI typecheck errors (`parseThorSwapMemo`, `commands/transaction.ts:165`) — verified identical on `main`, not introduced here.

## Additional context

Deliberately out of scope: the **modify/retarget** action (only the cancel form is built, though the predicates can tell them apart so nothing mislabels a retarget later), and all client UI. The consuming flow — keysign payload, broadcast, and the "cancelling" transitional state with its self-heal for a cancel that is included in a block but refused by the handler — is the follow-up in vultisig-windows.

## Summary by CodeRabbit

* **New Features**
  * Added support for cancelling native THORChain limit orders.
  * Added cancellation memo creation, validation, detection, and source-chain size checks.
  * Added eligibility checks for assets, amounts, order status, and observed transaction data.
  * Added price-bucket matching to identify indistinguishable orders.
  * Added dust-threshold calculation across different asset precisions.
  * Exposed cancellation utilities and types through the SDK.

### DIFF
--- a/.changeset/limit-order-cancel.md
+++ b/.changeset/limit-order-cancel.md
@@ -1,0 +1,17 @@
+---
+'@vultisig/core-chain': minor
+'@vultisig/core-mpc': minor
+'@vultisig/sdk': minor
+---
+
+Limit-order cancellation primitives: the `m=<` modify-limit-swap memo in its cancel form, eligibility, bucket-key duplicate detection, and L1 dust.
+
+Cancellation is not a variation on placement. Every failure mode is silent — the transaction confirms, the fee is spent, and the order carries on resting with nothing to distinguish it from success — so each rule is enforced rather than documented.
+
+`buildCancelLimitSwapMemo` refuses abbreviated assets outright: `ModifyLimitSwapMemo` is the one inbound memo type `processOneTxIn` does not run through `fuzzyAssetMatch`, so the placement memo's six-character contract suffix would address a bucket that by construction holds no order. Amounts are emitted as plain decimal integers, never compressed — these coins parse through `cosmos.ParseCoins`, which does not understand the scientific notation a placement LIM may use.
+
+`getLimitSwapCancelEligibility` fails closed at every unknown and cross-checks what was recorded at signing against what the queue reports — **assets as well as amounts**. Absence is not disagreement (an order placed seconds ago has not been polled), but a present-and-unparseable observation blocks exactly as a mismatch does.
+
+`getThorchainLimitOrderBucketKey` reproduces the advanced-swap-queue index key, including its zero-padding *and* right-truncation at 18 characters. Orders are addressed by `(layer-1 pair, ratio) + FromAddress` and the first match in the bucket wins, so orders sharing a ratio are not independently cancellable — compared on the key rather than on equal amounts, which would under-report collisions.
+
+`getLimitSwapCancelDust` rescales the live `dust_threshold` from THORChain's 1e8 into the source coin's own precision, with a safety multiple and an upper ceiling, refusing rather than defaulting when the threshold is missing, unparseable, or rounds away. A cancel once signed for 2000 wei — the 1e8 threshold used verbatim as an 18-decimal chain's smallest unit — was truncated to zero and never observed; that case is pinned by a test.
--- a/.changeset/limit-order-cancel.md
+++ b/.changeset/limit-order-cancel.md
@@ -1,6 +1,5 @@
 ---
 '@vultisig/core-chain': minor
-'@vultisig/core-mpc': minor
 '@vultisig/sdk': minor
 ---
 
--- a/packages/core/chain/package.json
+++ b/packages/core/chain/package.json
@@ -1773,6 +1773,26 @@
       "import": "./dist/swap/native/limitSwapAvailability.js",
       "default": "./dist/swap/native/limitSwapAvailability.js"
     },
+    "./swap/native/limitSwapCancelBucket": {
+      "types": "./dist/swap/native/limitSwapCancelBucket.d.ts",
+      "import": "./dist/swap/native/limitSwapCancelBucket.js",
+      "default": "./dist/swap/native/limitSwapCancelBucket.js"
+    },
+    "./swap/native/limitSwapCancelDust": {
+      "types": "./dist/swap/native/limitSwapCancelDust.d.ts",
+      "import": "./dist/swap/native/limitSwapCancelDust.js",
+      "default": "./dist/swap/native/limitSwapCancelDust.js"
+    },
+    "./swap/native/limitSwapCancelEligibility": {
+      "types": "./dist/swap/native/limitSwapCancelEligibility.d.ts",
+      "import": "./dist/swap/native/limitSwapCancelEligibility.js",
+      "default": "./dist/swap/native/limitSwapCancelEligibility.js"
+    },
+    "./swap/native/limitSwapCancelMemo": {
+      "types": "./dist/swap/native/limitSwapCancelMemo.d.ts",
+      "import": "./dist/swap/native/limitSwapCancelMemo.js",
+      "default": "./dist/swap/native/limitSwapCancelMemo.js"
+    },
     "./swap/native/limitSwapInbound": {
       "types": "./dist/swap/native/limitSwapInbound.d.ts",
       "import": "./dist/swap/native/limitSwapInbound.js",
--- a/packages/core/chain/swap/native/limitSwapCancelBucket.test.ts
+++ b/packages/core/chain/swap/native/limitSwapCancelBucket.test.ts
@@ -5,6 +5,7 @@ import {
   getThorchainLimitOrderBucketKey,
   toThorchainLayer1MemoAsset,
 } from './limitSwapCancelBucket'
+import { LimitSwapCancelMemoBuildError } from './limitSwapCancelMemo'
 
 const order = (sourceAmount: bigint, tradeTarget: bigint, overrides = {}) => ({
   sourceAsset: 'THOR.RUNE',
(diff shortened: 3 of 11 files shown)


johnnyluo (approved):
LGTM

Issue #949, opened 2026-07-02
arch: package boundaries (tsconfig-alias deps + circular dep), broadcast-dedup, SDK-boundary breaks


realpaaao:
Tracking issue for architecture / package-boundary findings from the SDK audit. The signing-ignores-relay-URL bug was split to its own issue (https://github.com/vultisig/vultisig-sdk/issues/953). Pure-judgment refactors are in the footnote (maintainer's call, not filed as discrete work).

- [ ] **`sdk` depends on `core-chain`/`core-mpc` via a tsconfig path alias, not a declared dependency**, and their source is inlined into the build (`sdk/tsconfig.json:32-33`); plus a **circular dep** `core-chain`↔`core-mpc` (`chains/utxo/tx/buildSignBitcoinFromPsbt.ts:1`). Both would fail a standalone `npm install`/publish of either package. Fix: declare real workspace deps + resolve via `exports`; move the shared protobuf types to break the cycle.
- [ ] **Broadcast dedup reimplemented 4 ways across 12 resolvers, and Bittensor lacks the try/catch → hash-verify fallback** every other chain has (a transient RPC timeout on Bittensor alone skips the fallback). `tx/broadcast/resolvers/*`, esp. `bittensor.ts`. Fix: shared `resolveDuplicateBroadcast(error, chainConfig)` + add the missing Bittensor try/catch.
- [x] **CLI `executor.ts` bypasses the SDK boundary**, importing `@vultisig/core-chain` internals directly (the only CLI file that does). `clients/cli/src/agent/executor.ts:9-10`. Fix: re-export via `@vultisig/sdk`.
- [ ] **`tools/swap/*` standalone builders are a second, non-integrated swap surface** with 3 error-handling conventions, exported publicly but unused internally. Decide: supported low-level API (standardize errors) or fold into `SwapService`.

### Structural debt (maintainer judgment — noted, not filed as discrete work)
- God classes: `VaultBase` (1,962 LOC), `Vultisig` (1,410), `cli/agent/executor.ts` (2,201, incl. hand-rolled EIP-712/DER). Split into the composed services they already partially delegate to, if/when the churn justifies it.
- `chains/cosmos/` conflates generic Cosmos-SDK plumbing with a THORChain-LP feature (1.6k LOC) and the QBTC bridge (1.2k LOC) — candidate for hoisting to top-level modules.
[...]

PR #1688, opened 2026-08-03
fix(sdk): preserve Bittensor hash on duplicate broadcasts

rcoderdev:
Refs #949
Preserve the deterministic Bittensor extrinsic hash when author_submitExtrinsic reports an already-imported or timeout-style outcome. Controlled JSON-RPC proof exercised the production RawBroadcastService with one signed extrinsic across duplicate, real request-timeout, unrelated rejection, and malformed-input scenarios; duplicate and timeout outcomes returned the same canonical Blake2 hash while unrelated failures remained fail-closed.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom Bittensor RPC endpoints when broadcasting transactions.
  * Added a standalone raw transaction broadcast option.
  * Duplicate submissions now return a deterministic transaction hash.
  * Timeout submissions return a hash after successful on-chain verification.

* **Bug Fixes**
  * Improved validation and normalization of raw transaction payloads.
  * Rejected malformed RPC responses and invalid transaction data before submission.
  * Improved RPC error handling while preserving unrelated failures.

### DIFF
--- a/.changeset/bittensor-raw-broadcast-dedup.md
+++ b/.changeset/bittensor-raw-broadcast-dedup.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Return the deterministic Bittensor extrinsic hash for duplicate or timeout-style raw broadcast submissions.
--- a/packages/sdk/src/vault/services/RawBroadcastService.ts
+++ b/packages/sdk/src/vault/services/RawBroadcastService.ts
@@ -6,6 +6,7 @@ import { Chain, CosmosChain, EvmChain, OtherChain, UtxoBasedChain } from '@vulti
 import { isChainOfKind } from '@vultisig/core-chain/ChainKind'
 import { bittensorRpcUrl } from '@vultisig/core-chain/chains/bittensor/client'
 import { getCosmosClient } from '@vultisig/core-chain/chains/cosmos/client'
+import { getCustomRpcOverride } from '@vultisig/core-chain/chains/customRpc/customRpcOverrides'
 import { getEvmClient } from '@vultisig/core-chain/chains/evm/client'
 import { polkadotRpcUrl } from '@vultisig/core-chain/chains/polkadot/client'
 import { getRippleClient } from '@vultisig/core-chain/chains/ripple/client'
@@ -453,18 +454,30 @@ export class RawBroadcastService {
    * @param rawTx - Hex-encoded extrinsic (with or without 0x prefix)
    */
   private async broadcastBittensorRawTx(rawTx: string): Promise<string> {
-    const hexWithPrefix = ensureHexPrefix(rawTx)
+    const hexWithoutPrefix = rawTx.startsWith('0x') ? rawTx.slice(2) : rawTx
+    if (
+      hexWithoutPrefix.length === 0 ||
+      hexWithoutPrefix.length % 2 !== 0 ||
+      !/^[0-9a-fA-F]+$/.test(hexWithoutPrefix)
+    ) {
+      throw new Error('Bittensor raw transaction must be non-empty, even-length hexadecimal bytes')
+    }
+
+    const hexWithPrefix = ensureHexPrefix(hexWithoutPrefix)
     const rawTxHash = () => this.getBittensorRawTxHash(hexWithPrefix)
 
     const { data: response, error } = await attempt(
-      queryUrl<{ result: string; error?: { message: string } }>(bittensorRpcUrl, {
-        body: {
-          jsonrpc: '2.0',
-          method: 'author_submitExtrinsic',
-          params: [hexWithPrefix],
-          id: 1,
-        },
-      })
+      queryUrl<{ result: string; error?: { message: string } }>(
+        getCustomRpcOverride(OtherChain.Bittensor) ?? bittensorRpcUrl,
+        {
+          body: {
+            jsonrpc: '2.0',
+            method: 'author_submitExtrinsic',
+            params: [hexWithPrefix],
+            id: 1,
+          },
+        }
+      )
     )
 
     if (error) {
--- a/packages/sdk/src/vault/services/RawBroadcastService.ts
+++ b/packages/sdk/src/vault/services/RawBroadcastService.ts
@@ -16,6 +16,7 @@ import { tronRpcUrl } from '@vultisig/core-chain/chains/tron/config'
 import { getBlockchairBaseUrl } from '@vultisig/core-chain/chains/utxo/client/getBlockchairBaseUrl'
 import { isRippleInFlightEngineResult } from '@vultisig/core-chain/tx/broadcast/resolvers/ripple'
 import { assertSuiTxSucceeded } from '@vultisig/core-chain/tx/broadcast/resolvers/sui'
+import { getBittensorTxHash } from '@vultisig/core-chain/tx/hash/resolvers/bittensor'
 import { getTxStatus } from '@vultisig/core-chain/tx/status'
 import { rootApiUrl } from '@vultisig/core-config'
 import { attempt } from '@vultisig/lib-utils/attempt'
@@ -453,6 +454,7 @@ export class RawBroadcastService {
    */
   private async broadcastBittensorRawTx(rawTx: string): Promise<string> {
     const hexWithPrefix = ensureHexPrefix(rawTx)
+    const rawTxHash = () => this.getBittensorRawTxHash(hexWithPrefix)
 
     const { data: response, error } = await attempt(
       queryUrl<{ result: string; error?: { message: string } }>(bittensorRpcUrl, {
@@ -466,6 +468,9 @@ export class RawBroadcastService {
     )
 
     if (error) {
+      if (isInError(error, 'Already Imported', 'already imported', 'timed out', 'timeout')) {
+        return rawTxHash()
+      }
       throw error
     }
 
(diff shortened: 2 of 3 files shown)


gomesalexandre (dismissed):
Reviewed at `df8e3acd`. The duplicate-detection half is good; the timeout half reports success without evidence.

## preferably-blocking: a client-side timeout returns a hash as if the broadcast landed

```ts
if (error) {
  if (isBittensorDuplicateError(error) || isBittensorTimeoutError(error)) {
    return rawTxHash()      // locally computed, never verified
  }
  throw error
}
```

`isBittensorTimeoutError` matches `ETIMEDOUT`, `ESOCKETTIMEDOUT`, `TimeoutError`, `FetchTimeoutError` — a connect/socket timeout that may mean the extrinsic **never reached a node at all**. Returning a hash there tells the caller the broadcast succeeded when there is no evidence it left the client. Direction A: a genuine failure surfaced as success.

The asymmetry is visible inside this same file — every other chain that infers success from an error signal verifies first:

```
line 379-402  Cosmos "tx already exists in cache" -> derive hash -> client.getTx(hash) -> assert DeliverTx success
line 604      Ripple tefALREADY                    -> verifyKnownRawTx(...)
line 699      Tron DUPLICATE_TRANSACTION           -> verifyKnownRawTx(...)
line 525-526  Bittensor duplicate OR timeout       -> return rawTxHash()      <- no verification
```

`verifyKnownRawTx` is defined at line 121 of this file and Bittensor has a working taostats-backed `getTxStatus` resolver (`packages/core/chain/tx/status/resolvers/bittensor.ts`), so the machinery is already here and already imported.

Worth adding: the canonical MPC-keysign broadcast path (`packages/core/chain/tx/broadcast/resolvers/bittensor.ts`) treats "already imported" as an immediate idempotent return — same as this PR, and that part is consistent — but funnels **everything else, timeouts included**, through `verifyBroadcastByHash`, a bounded 4-attempt poll that fails closed. So the raw path is the odd one out against both its in-file siblings and its own canonical twin.

**Suggested fix:** send the timeout branch through `verifyKnownRawTx(OtherChain.Bittensor, await rawTxHash(), 'Bittensor')` and leave the duplicate branch as-is. One line, mirrors Ripple/Tron exactly.

## the duplicate half is correct — keep it

"Already Imported" genuinely is a strong in-mempool signal, the regex is narrow and anchored (the tests prove suffixed messages are correctly excluded), and returning early there matches existing EVM/core-Substrate convention.

I also checked the hash-stability question specifically, since a resubmission producing a *different* hash would make the whole feature useless: a Substrate/Bittensor extrinsic hash is a deterministic blake2b over the raw signed bytes, so re-submitting identical bytes always yields the same hash. Era/nonce are already baked into those bytes. No issue.

Hex-validation guard, custom-RPC override wiring, and malformed-JSON-RPC handling all check out and are well tested. 69 unit tests pass, `tsc --noEmit` clean.

Happy to re-probe as soon as the timeout branch verifies.

🤖 reviewed via Claude Code

gomesalexandre:
Tackled the timeout half in 37462b12.

The duplicate-detection branch was already correct and untouched. The timeout branch (both the thrown-error path on `author_submitExtrinsic` and the JSON-RPC error-object path) now routes through `verifyKnownRawTx(OtherChain.Bittensor, ...)` instead of returning a locally-computed hash blind - same pattern already used by Ripple `tefALREADY` and Tron `DUPLICATE_TRANSACTION` in this file, backed by the existing taostats `getTxStatus` resolver. Fails closed (throws `BroadcastFailed`) if the extrinsic can't be confirmed within the poll.

Added mutation-verified regression tests:
- timeout (thrown error) -> verified success -> hash returned, taostats hit
- timeout (thrown error) -> verification comes back empty -> throws, doesn't hand back a hash
- timeout (JSON-RPC error object) -> verified success -> hash returned
- timeout (JSON-RPC error object) -> verification fails -> throws
- duplicate error -> still an immediate idempotent return, no verification poll (asserts `mockQueryUrl` called exactly once)

Confirmed each new/changed test fails on revert and passes with the fix restored. 71/71 tests green, `tsc --noEmit` clean, lint clean, prettier clean.

Not dismissing the existing changes-requested review or approving - flagging for a fresh look.

gomesalexandre (commented):
re-review on `37462b120`.

the timeout follow-up is the right fix. i re-checked the exact head in a disposable clone against fresh `origin/main` and the behavior now fails closed where the previous head could false-succeed:
- `packages/sdk/src/vault/services/RawBroadcastService.ts:498-504` no longer treats timeout transport failures like duplicates. timeout-style thrown errors now route through `verifyKnownRawTx(...)` instead of returning a locally-derived hash.
- `packages/sdk/src/vault/services/RawBroadcastService.ts:526-530` does the same for JSON-RPC error objects carrying timeout-style messages.
- the duplicate path stays idempotent and immediate at `:499-500` / `:526-527`, which is the correct asymmetry. "already imported" proves the node has seen the payload; a client-side timeout does not.
- the new malformed-response guards at `:512-540` are also correct. they reject both missing/mixed `result`/`error` payloads and non-hash `result` values instead of handing back garbage as a success hash.

receipt:
- codex remote/local diff pass in the exact checkout converged on no blocker in the changed files
- focused vitest from the exact PR checkout passed: `../../node_modules/.bin/vitest run --config tests/unit/vitest.config.ts tests/unit/vault/services/RawBroadcastService.test.ts` -> `71 passed`

not approving because the current head includes gomes-authored commits (`41c562c84` merge from main and `37462b120` substantive timeout verification follow-up). per four-eyes this needs a non-gomes current-head stamp.

gomesalexandre (commented):
Reviewed current head `92e7c685d131cee694b7b3e0986ea61ba16b0c23`.

Grounded take:
- the duplicate path is correctly idempotent: `Already Imported` returns the deterministic local Bittensor extrinsic hash for the exact bytes submitted
- the timeout path is correctly stricter than duplicate: it only returns that hash after `verifyKnownRawTx(...)` confirms the extrinsic on-chain, and still fails closed when that verification cannot prove landing
- the malformed-response hardening is real and useful: result/error XOR, stringly JSON-RPC error validation, and `isValidTxHash` on the returned hash close the old fail-open shape
- the new tests are substantial and hit the right branches, including near-miss strings that must NOT be misclassified as duplicate/timeout

I did not find a fresh correctness blocker in the diff.

Not stamping from this account because the branch head still contains substantive `gomesalexandre` commits, so current-head approval needs a non-gomes reviewer under the four-eyes rule.

Runtime-receipt note: this is an SDK behavior change, so approval should continue to rely on the real current-head runtime evidence already referenced in the PR thread/body, not on unit tests alone.

gomesalexandre (commented):
Remote conceptual pass on current head `92e7c685d131cee694b7b3e0986ea61ba16b0c23` looks sound from a correctness standpoint.

- duplicate and timeout-style Bittensor `author_submitExtrinsic` outcomes no longer get conflated. Duplicate stays idempotent, while timeout now goes through the known-tx verification poll and fails closed if the extrinsic cannot be confirmed.
- the canonical Bittensor hash is derived from the exact submitted raw extrinsic bytes, so the preserved hash matches the actual broadcasted payload rather than a reconstructed variant.
- malformed JSON-RPC responses and invalid tx-hash shapes are rejected instead of being treated as success.
- current-head CI is green.

A couple of non-blocking notes only:
- there is still no dedicated test for the thrown-error duplicate branch, even though the thrown-error timeout branch is covered.
- `getCustomRpcOverride(OtherChain.Bittensor)` is a real scope expansion beyond the narrow hash-preservation fix, though it is small and tested.

I am leaving this as a comment rather than an approval because the current head includes gomes-authored commits, so four-eyes requires a non-gomes human stamp on this exact head before merge.

rcoderdev:
Rebased this branch onto the latest `main` and revalidated the feature delta. I also added dedicated coverage for duplicate errors thrown by the submit call (`Already Imported` and `TRANSACTION ALREADY IMPORTED`), confirming the deterministic hash is returned without a status poll. The focused RawBroadcastService suite (73 tests), full repository checks, and the controlled Bittensor duplicate/timeout runtime proof all pass on the updated head. `johnnyluo` remains requested for the required non-Gomes approval.

gomesalexandre (commented):
Reviewed current head `c7ca90f1`.

No fresh correctness blocker in the latest delta.

- the Bittensor duplicate path is still the right asymmetry: duplicate-style outcomes return the deterministic hash immediately, while timeout-style outcomes only return that hash after `verifyKnownRawTx(...)` confirms the extrinsic on-chain and otherwise fail closed
- malformed JSON-RPC responses still reject cleanly instead of leaking garbage or undefined as a success hash
- the newest head move is the `test(sdk): cover thrown Bittensor duplicate errors` follow-up, which extends the thrown-error duplicate coverage without widening the runtime surface
- the duplicate regex remains narrow enough that the near-miss timeout/invalid strings in the suite do not get misclassified as successful duplicates

I do not see a new blocker on `c7ca90f1` from a remote diff pass.

Four-eyes note: this branch still contains substantive `gomesalexandre` commits, so it needs a non-gomes current-head approval before merge. Also re-check CI on this exact head once the queued jobs finish.

rcoderdev:
Acknowledged on current head `c7ca90f1`. The exact-head local unit suite, smoke checks, full repository checks, and controlled Bittensor broadcast harness all pass. I’ll keep the branch unchanged while the queued GitHub checks and an independent current-head approval complete.

johnnyluo (approved):
LGTM

PR #1689, opened 2026-08-03
Fix approval confirmation before dependent swaps

### LINKED ISSUE #940: fix(swap): approve→confirm gate is dead code; approve and swap broadcast back-to-back
## Evidence
`packages/sdk/src/vault/services/SwapService.ts:153-167` — `approvalPayload` is declared and only ever assigned `undefined` ("Core handles this internally"). `VaultBase.ts:1727-1746` gates `sign→broadcast→waitForConfirmation(approvalHash)` behind `if (approvalPayload)`, which is therefore never true; `waitForConfirmation` (`:1935`) is dead for swaps. The real approve tx is bundled into `keysignPayload.erc20ApprovePayload` and broadcast in the `BroadcastService.ts:105-122` loop with no wait between approve and swap.

## Impact
The intended safety (wait for the ERC-20 approval to mine before broadcasting the dependent swap) is silently disabled. Approve (nonce N) and swap (nonce N+1) go out together; on any node that doesn't reliably queue the dependent tx, the swap can revert for insufficient allowance.

## Fix sketch
Either have `SwapService.prepareSwapTx` return a real separate `approvalPayload` so the existing gate runs, or make `BroadcastService` await confirmation of input[0] before broadcasting input[1] for approve+swap bundles.

---

rcoderdev:
Closes #940

Closes #940.\n\nBundled ERC-20 approvals are now broadcast and confirmed before their dependent swaps. Approval failures and timeouts fail closed, while provider allowance spenders and per-message ECDSA recovery IDs are preserved.\n\nReal QA: a designated Base Fast Vault completed the USDC approval and dependent swap in later block and nonce order. The approval targeted the provider-supplied spender; final-tree production quote/prepare and failure/timeout paths also passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * ERC-20 approval transactions now wait for confirmation before dependent swaps are broadcast.
  * Improved handling of approval failures and confirmation timeouts, preserving transaction details during partial failures.
  * Non-approval multi-input transactions continue to broadcast as before.

* **Enhancements**
  * Added configurable confirmation polling intervals and timeouts.
  * Improved multi-signature transaction handling across bundled and UTXO-based flows.


gomesalexandre (approved):
Reviewed at `054f9036`. Approving — the confirmation predicate is the right one and the failure paths hold.

## the thing that decides this PR

"Wait for the approval before swapping" is only worth anything if "confirmed" means the chain actually accepted it. The classic version of this bug treats a broadcast ack or mempool visibility as confirmation. This doesn't:

```ts
// core/chain/tx/status/resolvers/evm.ts
const { data } = await attempt(client.getTransactionReceipt({ hash }))
const status = data.status === 'success' ? 'success' : 'error'

// BroadcastService
if (result?.status === 'success') return
if (result?.status === 'error') throw new Error(`Approval tx failed: ${txHash}`)
...
throw new Error(`Approval tx not confirmed within ${timeoutMs / 1000}s: ${txHash}.`)
```

A real receipt with the EVM status field. Only `success` proceeds; a reverted approval (`status=0`) throws, and so does a timeout. Because the throw happens inside the `for...of` at index 0, index 1 — the swap — is never reached. Verified that control flow rather than inferring it from the tests.

Failure-path probes with explicit controls, run against the real `BroadcastService` with only the I/O boundary mocked:

```
PROCEED control  approval status='success'    -> both broadcasts fire, txHash = swap hash
BLOCK   control  approval status='error'      -> 1 broadcast, swap never reached
BLOCK   control  stuck 'pending' → timeout    -> 1 broadcast
BLOCK   control  'not_found' whole window     -> 1 broadcast   (unknown-to-node ≠ success)
SCOPE   control  no erc20ApprovePayload       -> no wait injected, both inputs broadcast (no regression)
```

The `not_found` case is the one I'd have most expected to leak, and it correctly does not.

## the recovery-id fix is the quiet half of this PR

The `formatSignature` / `convertSignature` change fixes a **real pre-existing latent bug**: multi-message ECDSA signing stamped every message with the *first* message's `recovery_id`. For a bundled approve+swap in one MPC round that means one of the two transactions could recover to the wrong signer. The per-signature `recovery` with a fallback to the legacy top-level field (for chains with no per-sig recovery, e.g. UTXO) is correct and is load-bearing for this feature working at all — without it the bundle could not compile two valid signatures. Worth calling out because the PR title only advertises the ordering fix.

Direction A: `getEvmSigningInputs` guarantees `erc20ApprovePayload` present ⇒ exactly 2 inputs with approve at index 0, so the `txInputsArray.length > 1` guard is redundant-but-harmless rather than a gap. No case found where the payload is set and the wait is skipped.

`yarn test` 205 files / 3040 tests green, `tsc --noEmit` clean.

## should-fix (follow-ups, not blocking)

**1. The 60s timeout / 3s poll is never wired to production.** `confirmationOptions` is a constructor param, but `VaultBase.ts:374` constructs `BroadcastService` with two args — every real caller gets the hardcoded default and only tests override it. On congested Ethereum mainnet, which is precisely this feature's target, a legitimate-but-slow approval can blow past 60s and abort a swap that would have gone through moments later. It fails closed so there is no fund loss, just a retry — but it is a reliability gap on the PR's own chain. Worth exposing as an SDK config knob or raising the L1 default.

**2. There are now two independent "wait for approval" implementations.** `VaultBase.ts` has its own private `waitForConfirmation` (~line 2029) called at ~1834 inside `vault.swap()`'s compound flow, gated on `approvalPayload` — which `SwapService.prepareSwapTx` sets unconditionally to `undefined`, so that whole path is currently unreachable. Fixing this at the `BroadcastService` layer is the right call since that is the live path, but the two mechanisms have different error semantics (`VaultErrorCode.Timeout` there vs a generic `Error` → `BroadcastFailed` here). Worth a cleanup issue so nobody later "fixes" the dead one believing it is live.

Both predate this diff. Neither is a reason to hold it.

Out of scope and confirmed untouched: approval amount (exact vs unlimited), sufficient-allowance skip, and the USDT zero-before-reapprove case all live in `core/mpc/keysign/swap/build.ts` — this PR only reacts to whatever `erc20ApprovePayload` upstream already decided.

🤖 reviewed via Claude Code

PR #1690, opened 2026-08-03
test(cli): cover canonical chain aliases in vault mutations

### LINKED ISSUE #1155: cli: vault_chain / vault_coin / address_book still bypass canonical normalizeChain alias handling
## Summary
CLI agent mutation flows still normalize chain strings with a local alias table instead of the SDK's canonical `normalizeChain()` helper.

## Evidence
- `clients/cli/src/agent/executor.ts:1891-1935` defines a separate local alias map for `resolveChain()`.
- `packages/sdk/src/utils/normalizeChain.ts:104-114` already supports aliases the CLI parser misses, including `gaia`, `bnb smart chain`, `bsc-mainnet`, `arbitrum one`, and `ethereum-mainnet`.

## Impact
This leaves a post-round5 drift bug: string forms already accepted by the SDK can still fail in real CLI mutation flows before the SDK ever sees them:
- `vault_chain`
- `vault_coin`
- `address_book`

That means first-party consumers can disagree about which chain strings are valid, even after alias parity was centralized in the SDK.

## Fix sketch
Delete the local parser in `clients/cli/src/agent/executor.ts` and route string-to-`Chain` resolution through exported `normalizeChain()`, preserving the executor's existing `null`-on-unknown behavior.

## Effort
Small.

## Report
Round 6 report: https://gist.github.com/gomesalexandre/d089c11ec3700c0e391c683b1c57c083
Campaign index artifact: https://claude.ai/code/artifact/be2deadf-6195-4dfa-a910-b269b9af7a84

## Note
I have a local fix + receipts prepared in a draft PR.

rcoderdev:
Closes #1155

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chain alias resolution for vault and address book operations.
  * Added support for normalized chain names, including `gaia`, `Terra Classic`, and `cosmoshub-4`, before processing requests.


gomesalexandre (approved):
Approving. A test-only PR is worth exactly as much as its tests' ability to fail, so that is what I checked.

All three tightened tests were mutation-verified against the real source (not a copy):

```
1. removed `gaia: Chain.Cosmos` from normalizeChain.ts
   -> "vaultChain resolves SDK-only aliases..."          RED ✓
2. excluded Chain.TerraClassic from the canonical-enum-alias loop
   (breaks the separator-stripped "Terra Classic" fallback)
   -> "vaultCoin resolves separator-tolerant..."         RED ✓
3. cosmoshub-4 -> cosmoshub-9-MUTATED in cosmos/chainInfo.ts
   -> "addressBook resolves canonical Cosmos chain IDs..." RED ✓
```

All three restored, 26/26 green as the control.

The change is a real improvement rather than coverage theatre: it replaces loose `toHaveBeenCalled()` assertions (which pass regardless of what was passed) with exact-arg checks, and swaps canonical inputs like `'Ethereum'` for SDK-only aliases (`'gaia'`, `'Terra Classic'`, `'cosmoshub-4'`) that actually exercise alias resolution. The old versions would have stayed green through every one of the three mutations above.

No vacuous shapes — no `toBeDefined()` on always-defined values, no loops over empty fixtures, no swallowed catches. And the aliases are passed straight through to the real `resolveChain` / `resolveChainReference` rather than being checked against a hand-copied alias table in the test file, so there is nothing here to drift out of sync with source.

(The `tsc --noEmit` errors in `broadcastGuard.ts` / `transaction.ts` / `sendDeclinePty.test.ts` are pre-existing on the same worktree and unrelated — this PR touches one test file.)

Nice cleanup. Tests that go red under mutation are the only kind worth having.

🤖 reviewed via Claude Code

PR #1691, opened 2026-08-03
fix(mpc): preserve canonical EdDSA signatures end to end

### LINKED ISSUE #938: test(mpc): keygen/keysign/reshare have zero test coverage
## Evidence
`packages/core/mpc/dkls/dkls.ts` (`startKeygenWithRetry`, `startReshareWithRetry`, `startKeyImportWithRetry`) and `packages/core/mpc/schnorr/schnorrKeygen.ts` have no test files; `packages/mpc-wasm/src/index.ts` (`WasmDklsEngine`/`WasmSchnorrEngine`/`WasmMpcEngine`) has 0 tests. Grepping all 29 `core/mpc` test files for `DKLS|schnorrKeygen|startKeygenWithRetry` returns zero matches.

## Impact
The single most security-critical subsystem — the code that produces the signatures authorizing every fund movement, in both fast-vault and secure/relay paths — has no TS-level regression protection. A bug in retry/session-sequencing/reshare logic could corrupt or leak key shares with nothing to catch it.

## Fix sketch
Build an in-process mocked-relay-transport harness running 2-3 real `WasmDklsEngine` instances through a full keygen+keysign round (real wasm, not mocked), asserting a valid signature results — the one test that actually proves signing works.

---

rcoderdev:
## Summary

- add real in-process DKLS and Schnorr WASM ceremonies covering keygen, keysign, reshare, and post-reshare signing
- cryptographically verify ECDSA recovery and canonical Ed25519 signatures
- register the MPC WASM package in the core Vitest project

## Verification

- `yarn check:ci`
- `yarn test:core`
- `yarn vitest run --config .config/vitest.core.config.ts packages/mpc-wasm/src/index.test.ts --reporter=verbose`

Closes #938

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Ed25519 and Schnorr signature byte ordering across supported transaction and signing flows.
  * Improved compatibility between WASM and SDK signature outputs by preserving canonical `R || S` formatting.
  * Maintained safe handling of signature data without altering unrelated formats.

* **Tests**
  * Added comprehensive coverage for key generation, signing, resharing, verification, and session cleanup.
  * Expanded regression coverage across Polkadot, Solana, Sui, Cardano, DKLS, and Schnorr workflows.


gomesalexandre (changes requested):
Ser, the tests here are the real deal - real `WasmMpcEngine`, real keygen/keysign/reshare, verified against `@noble/curves` rather than the SDK's own code. I mutation-checked them: restore the reversal you deleted and `expect(verify(...)).toBe(true)` at index.test.ts:158 goes red. That is a test that can actually fail, which is more than most "coverage" PRs can say. And the `.config/vitest.core.config.ts` gap was real - `packages/mpc-wasm` had zero tests and zero configs matching it on main, so nothing there was ever measured.

But the source change is not test scaffolding, and I think it breaks EdDSA signing.

## preferably-blocking: the engine contract flips, the consumer that depends on it does not

`_normalizeSchnorrSig` now passes bytes through. Downstream, `generateSignature` still reverses both halves for `signatureFormat: 'raw'`, and every EdDSA chain is `raw`:

`packages/core/chain/signing/SignatureFormat.ts` - sui, solana, polkadot, bittensor, ton, cardano, qbtc.

I traced the whole path and there is exactly ONE reversal downstream of the engine, no compensating second one:

`session.finish()` -> `keysign/index.ts:208` (splits r/s, passes through) -> `formatSignature` (passthrough, "already has correct endianness") -> `convertSignature` (passthrough) -> `compileTx` -> `generateSignature('raw')` -> **reverses each half** -> broadcast.

So the reversal you deleted from the engine was being undone downstream. Take it out of the engine and only one reversal survives.

### receipts

Fed a known-good canonical ed25519 signature from noble through the real `generateSignature` (real `initWasm()` WalletCore, not a stub) and verified the output:

```
CANONICAL-IN (what this PR's engine now emits) -> ed25519.verify == false   FAIL
REVERSED-IN  (what main's engine emits)        -> ed25519.verify == true    PASS
```

`generateSignature('raw')` wants big-endian halves and emits canonical. main feeds it big-endian and gets a valid signature out. This PR feeds it canonical and gets garbage.

And this is live, not dormant: `WasmMpcEngine` is wired into `platforms/chrome-extension/index.ts:41` and `platforms/node/index.ts:77`, not only the test setups. So it is solana/sui/polkadot/ton/cardano/bittensor/qbtc signing on the extension and node SDK.

The new test cannot catch this because it stops at the engine boundary. Both halves are internally consistent, they just disagree about who owns the byte order.

## suggestion: make it atomic, and test one layer up

If the current binding really does emit canonical (your ceremony says it does, and I believe it), then the fix is engine passthrough **and** dropping the reverse from the `raw` arm of `generateSignature` for eddsa, in the same commit. Split across two PRs and whichever lands first ships broken signing.

Then pin it with a test at the `compileTx` level - canonical sig in, verify the compiled output. An engine-level test will stay green through this entire class of bug, as it just did.

## q: what about the native engine?

The comment you deleted claimed `@vultisig/mpc-native` already emits canonical. If that is true then native EdDSA is broken on main today by this same reversal, which would be a live bug rather than a regression. If it is not true then the comment was wrong and worth correcting. Which engine ships on RN, and has anyone verified a native EdDSA broadcast end to end recently? I did not chase this one down.

## nit: the title undersells it

`test(mpc):` on a diff that changes signature byte order is a review hazard, people skim test PRs. `fix(mpc)!:` or split the source change out.

Happy to re-run the probe against a new head, the harness is quick.

gomesalexandre (approved):
approve

Grounded this at current head after `gh pr update-branch`.

Receipts:
- diff read against fresh `origin/main`
- real WASM EdDSA ceremony path reviewed in `packages/mpc-wasm/src/index.test.ts`, including `generateSignature('raw')` + `formatSignature` byte-equality and `ed25519.verify(...)` on the emitted signature
- local focused run in a disposable clone: `./node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/mpc-wasm/src/index.test.ts` passed (`2 tests`, including the Schnorr/EdDSA ceremony)

Notes:
- the current head keeps `_normalizeSchnorrSig` as pass-through and drops the `raw` reversal in `generateSignature`, which matches the new exact-path verification added in the PR
- a broader compile-golden run in the disposable clone hit missing workspace `dist/` imports from the borrowed-install environment before exercising PR logic, so I did not treat that loader failure as evidence against the branch

PR #1692, opened 2026-08-03
fix(sdk): migrate legacy vault backup encryption

### LINKED ISSUE #932: security: legacy vault-backup KDF is unsalted single SHA-256 — GPU-brute-forceable
## Evidence
`packages/lib/utils/encryption/aesGcm/decryptWithAesGcm.ts:8` derives the AES-256 key as a single **unsalted** `sha256(password)` — no salt, no stretching. `packages/lib/utils/encryption/vaultBackup/decryptVaultBackupWithPassword.ts:29-30` falls back to this legacy path whenever the payload lacks the `VLT\x02` PBKDF2 magic. New backups correctly use PBKDF2-HMAC-SHA256 @ 600k (`encryptVaultBackupWithPassword.ts:44`).

## Impact
Any legacy-format `.vult` backup (or any file without the new magic) is offline-brute-forceable at GPU speed — one SHA-256 per candidate vs 600k PBKDF2 rounds. No salt also enables cross-user precomputation. Anyone who obtains an old backup (cloud sync, disk image) can crack a weak/medium password orders of magnitude faster.

## Fix sketch
Keep legacy decrypt for import only; on successful legacy import, immediately re-encrypt to the PBKDF2 format and prompt the user to replace old backups. Treat any password that touched the legacy format as burned (advise rotation + fresh export).

---

rcoderdev:
Closes #932

## Summary by CodeRabbit

* **Security Enhancements**
  * Legacy vault backups are automatically upgraded to stronger PBKDF2-based encryption during import.
  * Migrated backups are saved before the vault becomes active.
  * Security guidance recommends rotating passwords and replacing older backup files.

* **New Features**
  * Added migration notifications for applications to detect upgraded legacy backups.

* **Bug Fixes**
  * Prevented failed migrations from saving incomplete backups or issuing misleading notifications.


gomesalexandre (approved):
Reviewed at `fcf6cca1`. Approving — no lockout or fund-loss regression found.

## the thing that mattered most

A "migrate backup encryption" PR has exactly one catastrophic failure mode: an existing backup stops decrypting and someone is permanently locked out of a vault. So that is what I checked first, structurally, before anything else:

```
$ gh pr diff 1692 --name-only
.changeset/secure-legacy-vault-imports.md
packages/sdk/src/VaultManager.ts
packages/sdk/src/Vultisig.ts
packages/sdk/src/events/types.ts
packages/sdk/tests/unit/vault/legacy-vault-backup-migration.test.ts
```

**Zero crypto files.** `decryptVaultBackupWithPassword`, `encryptVaultBackupWithPassword`, `vaultBackupCrypto.ts` and `vaultBackupConstants.ts` are all untouched — the decrypt path, which is the entire lockout surface, is not modified by this diff. The change is purely additive: detect a legacy-format backup on import, re-encrypt the *stored* copy to PBKDF2, emit a notice. The caller's original file is never rewritten.

Format detection being magic-byte based (`VLT\x02`) rather than container-version based is the right call here — it means there is no "backup predates versioning, so no marker, so which path?" trap, which is the usual way these migrations bite.

## the design call is right

Migrating the stored copy while telling the user their password and old backups must be treated as compromised is the correct posture. A silent re-encrypt would have been worse than doing nothing: it would leave the user believing a weak-KDF password is still safe. The notice copy says the necessary thing plainly.

## verification

Ran the PR's own migration suite myself at the PR head: **5/5 passed**.

The rest of the evidence came from a deeper pass I had run in parallel, and I am attributing it rather than claiming it as mine: 13 probes against the live exported functions (unicode / 10k-char / whitespace / empty passwords all round-trip, with must-succeed and must-fail controls both behaving), 100 back-to-back encryptions of identical plaintext yielding 100/100 unique salts and ciphertexts, single-bit tamper flips rejected by the GCM tag on **both** the new PBKDF2 and legacy blobs, wrong-password import leaving `storage.list()` empty with no partial vault persisted, error objects serialised with `Object.getOwnPropertyNames` containing no password material, and the full `packages/sdk` suite at 210 files / 3152 tests green plus `tsc --noEmit` clean.

KDF parameters check out against the constants file's stated Android/iOS wire-format parity: PBKDF2-HMAC-SHA256, 600k iterations, 16-byte CSPRNG salt, 12-byte fresh nonce per call.

## non-blocking notes

`suggestion:` the `finally { decryptedVaultBytes?.fill(0) }` zero-fill is good hygiene but scrubs only one of several in-memory copies of the plaintext — the base64 string, the parsed protobuf, and the vault instance's cached fields are not zeroed. Worth keeping, just worth not describing as a complete secure-erase anywhere user-facing.

`should-fix (pre-existing, NOT introduced here):` in `importVaultWithResult`, `vaultInstance.save()` runs before `storage.set('activeVaultId', ...)`, so if that `set` throws the caller gets a `VaultImportError` while the migrated vault is already durably saved — your own test "emits the migration notice after save even when setting the active vault fails" confirms the interleaving. This ordering predates the PR (it only interposes the notice-emit between two existing calls), so it is out of scope here, but it deserves a follow-up: a caller seeing an exception while a vault silently exists in storage is a confusing state to debug.

Neither is a reason to hold this. Nice, careful piece of work — the magic-byte detection and the fail-closed legacy path are what make it safe.

🤖 reviewed via Claude Code

PR #1696, opened 2026-08-03
feat(swap): sign and review THORChain limit-order cancellations

### LINKED ISSUE #1694: Limit-order cancellation: full-form memo asset, keysign payload, and co-signer decode
## Problem

#1685 landed the cancellation primitives — memo builder, eligibility, bucket key, dust. But a client still cannot actually *send* a cancel, and a co-signer cannot read one. Three pieces are missing, all of them SDK-layer:

**1. There is no full-form memo asset.** `getThorchainMemoAsset` returns the placement spelling, which abbreviates an L1 contract to its last six characters:

```
getThorchainMemoAsset({ chain: 'Ethereum', ticker: 'USDC', id: '0xA0b8…eB48' })
  → 'ETH.USDC-06EB48'
```

That is exactly the form a cancel must **not** use — `ModifyLimitSwapMemo` skips `fuzzyAssetMatch`, so the abbreviation addresses a bucket holding no order. `buildCancelLimitSwapMemo` rejects it (correctly), which means today there is no supported way to produce an input it accepts for a token leg. Every caller would have to hand-assemble `CHAIN.TICKER-<full contract>` and get the secured-denom rule right on its own.

**2. There is no cancel keysign payload builder.** `buildLimitSwapKeysignPayload` asserts a `=<` placement memo and cannot carry an `m=<`. A cancel needs its own two-branch build, mirroring the placement one:

- **THORChain-sourced** → `MsgDeposit` from the vault's THOR address, paid in RUNE. No inbound vault, no attached value.
- **L1-sourced** → a dust transfer to the chain's Asgard inbound vault with the cancel memo attached, the amount coming from `getLimitSwapCancelDust`.

Note a cancel moves no tokens: the signing asset is the **gas asset of the chain the order was funded from**, never the order's own asset. A BTC-funded order's cancel is signed with BTC even when the order sells a token.

**3. `getKeysignLimitSwapOrder` returns `undefined` for `m=<`.** So a cancel reaching a joining device renders as a generic send — an opaque memo and send terms, with nothing saying an order is being cancelled. This is the same class of bug fixed for placement in vultisig-windows#4477, and it arrives the moment cancelling ships.

## What's needed

- A full-form memo asset helper, applying the same flavour rules as the placement one (L1 `CHAIN.TICKER`, tokens `CHAIN.TICKER-<full contract>`, secured assets keeping their denom) but without the abbreviation. Round-tripping matters: the value it emits must be one `buildCancelLimitSwapMemo` accepts, and the two should be tested together rather than in isolation.
- `buildLimitSwapCancelKeysignPayload`, covering both branches, refusing to build when the memo is not a cancel memo, and applying the same fail-closed inbound gates the placement builder does (a cancel routed to a stale or halted inbound is as lost as a placement).
- Cancel-aware decoding for the co-signer, so an `m=<` payload is recognised and can be shown as a cancellation rather than a transfer. It should stay distinguishable from a *retarget* — `isCancelLimitSwapMemo` already draws that line.

## Acceptance criteria

- [ ] A full-form asset helper exists, and its output is accepted by `buildCancelLimitSwapMemo` for L1, token and secured legs
- [ ] Cancel payloads build for both a THORChain source and an L1 source, with the L1 branch attaching derived dust and targeting the live inbound
- [ ] Building refuses a non-cancel memo, and fails closed on inbound problems
- [ ] A co-signer can identify an `m=<` payload as a cancellation, and a retarget is not reported as one
- [ ] Exported from the root entry with a changeset

Ehsan-saradar:
## Description

The cancellation primitives from #1685 could *describe* a cancel but not send one, and a joining device could not read one at all. This adds the three missing pieces.

Fixes #1694

### 1. A full-form memo asset

`getThorchainMemoAsset` returns the placement spelling, which abbreviates an L1 contract to its last six characters:

```
getThorchainMemoAsset({ chain: 'Ethereum', ticker: 'USDC', id: '0xA0b8…eB48' })
  → 'ETH.USDC-06EB48'
```

That is exactly the form a cancel must **not** use. `ModifyLimitSwapMemo` is the one inbound memo type `processOneTxIn` does not route through `fuzzyAssetMatch`, so its asset string builds the order-index key verbatim and an abbreviation addresses a bucket that by construction holds nothing. `buildCancelLimitSwapMemo` rejects it — correctly — which meant there was previously no supported way to produce an input it *would* accept for a token leg.

`getThorchainCancelMemoAsset` emits the same notation minus the abbreviation. Both spellings now share one converter and one set of validation rules, so they cannot drift apart in anything but that abbreviation, and the round-trip through `buildCancelLimitSwapMemo` is tested for L1, token and secured legs rather than each side in isolation.

`getThorchainMemoAssetChain` resolves a memo asset back to its home chain across every flavour. Reading the prefix up to the *first* separator is what makes it work for a secured denom (`eth-usdc-0x…`), which a `.`-split hands to the lookup whole and resolves to nothing.

### 2. A cancel keysign payload builder

`buildLimitSwapCancelKeysignPayload` branches on where the order was funded:

- **THORChain source** — `MsgDeposit` from the vault's own THOR address, carrying no value. The memo is the whole instruction.
- **L1 source** — a plain transfer to the chain's Asgard inbound with the memo attached and dust from `getLimitSwapCancelDust`, present solely so Bifrost observes the transaction. The router is deliberately not involved even for an ERC20-funded order: `depositWithExpiry` exists to move tokens, and a cancel moves none.

The signing asset is the funding chain's **gas** asset, never the order's own — a BTC-funded order cancels with BTC, an ERC20-funded one with ETH. A token there would build an ERC20 transfer that drops the memo entirely: the transaction confirms, gas is spent, and nothing is cancelled.

Every gate fails closed, because a cancel that is accepted but matches nothing is indistinguishable from success:

| Gate | Why |
|---|---|
| Memo must actually cancel | `m=<` with a non-zero final field re-prices a resting order. Signing one from a function whose callers believe they are cancelling leaves the position open. |
| Memo must fit the source chain's budget | Nothing in a cancel can be shortened — amounts define the bucket, assets skip `fuzzyAssetMatch` — so an over-long memo is refused rather than truncated. |
| Global trading pause | Same rule placement applies, including when the inbound list is unverifiable. |
| Live, non-halted inbound | Destination comes from the live view, not a cache. A cancel sent to a churned-away vault is as lost as a placement. |

One gate is deliberately **absent**: the `EnableAdvSwapQueue` mimir that placement re-checks at sign time. That gate protects a *new* `=<` order from executing as an unprotected market swap — a risk a cancel does not carry. Refusing to close an already-resting position because the queue stopped accepting new ones would strand it for the remainder of its TTL with no way out, which is the worse failure of the two.

### 3. Cancel-aware review for co-signers

`getKeysignLimitSwapCancel` decodes the order being closed from `keysignPayload.memo`, alongside the existing `getKeysignLimitSwapOrder`. A cancel carries no swap payload on any branch, so a co-signer keying off one previously saw a send of a trivial amount to an opaque address with an opaque memo — worse than uninformative, since dust reads as harmless while the transaction closes a position.

As with placement, the terms come from the memo because the memo is the exact string THORChain executes, so what a reviewer sees cannot disagree with what gets signed — unlike a display field supplied alongside it, which a co-signer has no way to verify. A retarget is reported as not-a-cancellation rather than mislabelled, and the reproduced bucket key travels alongside so a reviewer holding the vault's open orders can tell whether the cancel is unambiguous.

`parseCancelLimitSwapMemo` is the inverse of the builder and re-checks the builder's invariants on the way back in — a memo reaching a reviewer from another device was not necessarily built by this code. It splits each `<amount><ASSET>` field the way THORNode's `getCoin` does, anchoring the asset on a non-digit: without that, the greedy digit run backtracks and an asset-less `100000000` parses as amount `10000000` with asset `"0"`.

## Which feature is affected?

- [x] Swap — THORChain limit orders (`=<` advanced swap queue), cancellation path

No transaction link: this is the SDK layer only. The vultisig-windows UI that drives it is the follow-up (vultisig-windows#4481), and a mainnet cancel will be attached there.

## Checklist

## Additional context

**Also deduplicated** three copies of the memo-asset separator list (`.`, `/`, `~`, `-`) that had accumulated across `limitSwapCancelMemo`, `limitSwapCancelBucket` and the new chain resolver. Drift between them is a fund-safety bug — they decide where a chain prefix ends, and disagreeing about that means one module calls a secured denom truncated while another does not.

**Tests**: 2492 core tests pass. New coverage spans both payload branches, every fail-closed gate, the 2000-wei rescaling case pinned against an 18-decimal chain, the retarget-vs-cancel boundary in all three modules, and secured-denom handling throughout.

**Not covered here**:

- `resolveLimitSwapCancelAsset` still cannot resolve every asset flavour to a cancellable pair — that remains as it shipped in #1685.
- **Secured-asset-FUNDED orders can never pass `getLimitSwapCancelEligibility`.** `getLimitSwapSourceChainKind` routes through `assertValidPoolId`, whose regex demands dotted `CHAIN.ASSET` and rejects the dash-only secured spelling. Inert in practice: placement cannot build a secured-sourced order either, so no such live order exists. Left as-is rather than papered over, so the day placement learns secured denoms this surfaces as a real gap rather than a silent one. Note the *payload builder* already handles the flavour correctly — `getThorchainMemoAssetSourceChain` sends a secured source from THORChain — so only the eligibility gate is in the way.

### Update — chain-identity gate added

`buildLimitSwapCancelKeysignPayload` now requires the signing coin's chain to be the one the memo names as the order's funding chain, closing the under-blocking gap raised in review. The two arrived as independent parameters with nothing tying them together, so an ETH coin paired with a BTC-sourced memo built and broadcast cleanly and was then refunded by `From.IsChain(Source.Asset.GetChain())` — a successful-looking transaction that cancelled nothing.

The funding chain is derived from the memo rather than cross-checked against a second parameter, so there is one authority for it, and `getThorchainMemoAssetSourceChain` supplies it with `GetChain()` semantics rather than the asset's home chain — a secured or synth source is custodied on THORChain and must be sent from a THOR address even though it originates elsewhere.

Tests pin the reviewer's exact probe with its control, and deliberately use a memo that *fits* the signing chain's byte budget for the mirror case, so it is the chain check doing the work rather than the byte-limit one that was catching it incidentally.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end support for cancelling limit orders across THORChain and layer-1 funding sources.
  * Added cancellation memo parsing with asset details, amounts, and cancellation bucket information.
  * Added live inbound routing, gas-asset signing, UTXO handling, and chain-specific cancellation dust support.
  * Added decoded cancellation details for transaction review.
* **Bug Fixes**
  * Improved asset normalization and contract-address preservation.
  * Safely rejects malformed, retargeted, oversized, unsupported, or non-positive cancellation requests.


gomesalexandre (changes requested):
Reviewed at `1d371888`. This is unusually rigorous work — THORChain-pause, halted-inbound, memo abbreviation, byte-overflow, the signed-vs-observed asset and amount cross-checks are all fail-closed, and the review surface is structurally incapable of drifting from the signed bytes (`getKeysignLimitSwapCancel` takes `Pick<KeysignPayload, 'memo'>`, so there is nothing else it *could* display). One gap in the same class the PR otherwise guards obsessively.

## preferably-blocking: nothing checks that `signingCoin.chain` is the memo's funding chain

`buildLimitSwapCancelKeysignPayload` takes `signingCoin` and `memo` as independent parameters and validates three things: the memo is a real cancel (`isCancelLimitSwapMemo`), the coin is *some* chain's gas asset (`isFeeCoin`), and the memo fits that chain's byte budget. It never checks the coin's chain against the memo's `sourceAsset` chain.

Probed on this head, with a working control:

```
BUILT  CONTROL  BTC coin + BTC-sourced memo -> to=btc-vault memo=m=<:100000000BTC.BTC:5000000000THOR.RUNE:0
BUILT  CLAIM    ETH coin + BTC-sourced memo -> to=eth-vault memo=m=<:100000000BTC.BTC:5000000000THOR.RUNE:0
```

Both inbounds mocked live and tradeable so neither case could fail for inbound reasons — my first attempt at this probe had both cases throwing on leftover paused-trading state, which proved nothing, so the control is doing real work here.

On THORChain, `MsgModifyLimitSwap.ValidateBasic` requires `From.IsChain(Source.Asset.GetChain())`. An ETH-origin transaction referencing a BTC-sourced order fails that and gets refunded — **the wallet sees a normal successful ETH broadcast while the cancel silently does nothing**. That is verbatim the failure mode the PR body names three separate times: "a cancel that is accepted but matches nothing is indistinguishable from success".

The mirror case (BTC coin + ETH-sourced memo) does get caught, but only incidentally — the resulting memo overflows BTC's 80-byte OP_RETURN budget. That is the byte-limit check firing, not a chain-identity check, and it would stop protecting the moment an asset name got shorter.

Direction A on the #2301 taxonomy: under-blocking lets a broken pairing through as if it succeeded.

This is dormant today — the only caller is the not-yet-written vultisig-windows#4481 — which is exactly why it is worth closing now rather than after. The obvious UI bug is picking "the vault's ETH coin" instead of deriving the chain from the parsed memo, and there is currently no SDK-level backstop for that. The module already exports what it needs (`getThorchainMemoAssetChain` on the parsed `sourceAsset`, `findThorchainMemoAssetSeparatorIndex` for L1 vs synth/secured/trade notation), so this is a same-file addition rather than new machinery.

Worth noting the existing comment above the `isFeeCoin` guard shows the author was already thinking about mispairing — "passed the order's own asset rather than the funding chain's gas asset" — it just guards fee-coin-ness rather than chain identity.

## verified correct

- **Full-vs-abbreviated asset split holds** — `buildCancelLimitSwapMemo` throws on an abbreviated asset and accepts full.
- **The late `.toUpperCase()` in `1d371888` is load-bearing**, not cosmetic: without it a lower-case ERC20 contract id fails `assertValidPoolId`'s `[A-Z0-9]` regex downstream, reporting every ERC20-funded order as `unroutableSourceChain`.
- **Case-only drift resolves, real drift is flagged** — `resolveLimitSwapCancelAsset` tolerates upper/lower but returns `disagrees` on a different contract id, mapped to `signedDataDisagreesWithChain`. Amount cross-check fails closed too.
- **Round-trip is stable** — `buildCancelLimitSwapMemo` → `parseCancelLimitSwapMemo` → `getThorchainLimitOrderBucketKey` yields an identical bucket key from both the pre-build inputs and the post-parse output.
- **Not re-checking `EnableAdvSwapQueue` for cancels is the right asymmetry** — refusing to close a resting order because *new* orders are paused would strand the position for its full TTL. Deliberate and documented.
- **No affiliate field is a non-issue** — the `m=<` format carries three segments and no affiliate, and a cancel executes no swap, so the mcp-ts#769 fail-closed concern does not apply.
- **No expiry is correct** — the cancel L1 branch builds a plain transfer with no router involvement, so `depositWithExpiry` has nothing to do.

7 touched test files: 168/168. `tsc --noEmit` clean in `packages/sdk`. (The 10 failures in `findSwapQuote.slippage.test.ts` are a pre-existing network-dependent flake in a file this PR does not touch.)

## non-blocking

Secured-asset-funded orders can never pass `getLimitSwapCancelEligibility` — `getLimitSwapSourceChainKind` routes through `assertValidPoolId`, whose regex demands dotted `CHAIN.ASSET` and rejects the dash-only secured spelling. Inert in practice, since placement cannot build a secured-sourced order either, so no such live order exists. Worth one line under "Not covered here" alongside the existing `resolveLimitSwapCancelAsset` caveat rather than a code change.

Happy to re-probe the moment the chain check lands.

🤖 reviewed via Claude Code

gomesalexandre (approved):
This is exactly what I asked for, and the shape is better than what I sketched.

I asked for a check that the signing coin's chain is the one the memo says funded the order. You did not just add the compare, you went and answered the harder question underneath it: *which* chain a memo asset actually sends from. `getThorchainMemoAssetSourceChain` implements `Asset.GetChain()` semantics rather than "split on the first dot", so a synth, a trade asset and a secured denom all correctly resolve to THORChain instead of to the L1 whose ticker they borrow. That distinction is the whole bug and it is easy to miss.

Two details worth calling out as right rather than merely present:

`findThorchainMemoAssetSeparatorIndex` takes the **first** separator, and the comment says why: a secured denom spells the whole thing with `-`, so searching from the right lands inside the token id. Getting that backwards would have made every secured asset resolve to a different chain, and the bug would have looked like a chain-mapping problem rather than a parsing one.

And the bare-prefix guard. `getThorchainMemoAssetChain('BTC')` happily returns `Bitcoin`, so without the `separatorIndex === -1` early return a memo carrying a naked chain code would have resolved to a real chain and passed the compare. Refusing it is correct and the comment explains the trap.

## Receipts

Drove the real `buildLimitSwapCancelKeysignPayload` with `getChainSpecific` / `getKeysignUtxoInfo` / `refineKeysignUtxo` / `getThorchainInboundAddress` stubbed, inbounds fixed at BTC + ETH. Single variable across the two arms: the builder file, `1d371888` (the head I reviewed) vs `a820e4aa`. Nothing else changed.

```
                                      1d371888 (reviewed)   a820e4aa (head)
C1 pass-ctrl BTC coin + BTC memo      BUILT to=btc-vault    BUILT to=btc-vault
C2 fail-ctrl non-fee coin (USDC)      THREW gas-asset       THREW gas-asset
V1 ETH coin  + BTC.BTC memo           BUILT to=eth-vault    THREW funded on Bitcoin
V2 RUNE coin + THOR.RUNE memo         BUILT isDeposit=true  BUILT isDeposit=true
V3 ETH coin  + ETH.USDC-0X… memo      BUILT to=eth-vault    BUILT to=eth-vault
V4 RUNE coin + secured ETH-USDC memo  BUILT isDeposit=true  BUILT isDeposit=true
V5 ETH coin  + secured ETH-USDC memo  BUILT to=eth-vault    THREW funded on THORChain
V6 RUNE coin + synth BTC/BTC memo     BUILT isDeposit=true  BUILT isDeposit=true
V7 BTC coin  + synth BTC/BTC memo     BUILT to=btc-vault    THREW funded on THORChain
V8 unroutable NOPE.NOPE               BUILT to=eth-vault    THREW cannot resolve
V9 ETH coin  + lowercase eth.eth      BUILT to=eth-vault    BUILT to=eth-vault
```

C1 and C2 are identical across both arms, which is what makes this a diff of your guard rather than a broken harness. V1 is the case I flagged and it flips. V5 and V7 are the ones I did not think to ask about and the new helper catches them too.

The rows that matter most to me are V2, V3, V4, V6 and V9: **identical on both arms**. Five legitimate cancels, including the two THORChain-held flavours and a lowercase memo, all still build. This is the thing I care about more than the block landing, because a chain check on a cancel path is one bad `toUpperCase` away from stranding an order for the rest of its TTL, and it does not over-block.

Also probed the reverse map directly, since inverting a record is where this kind of guard usually rots: `forwardEntries=18 reverseEntries=18`, no prefix collisions, no round-trip loss. The inversion is safe to gate money on. Deriving it rather than hand-maintaining it was the right call and the comment saying why is worth keeping.

`packages/core/chain/swap/native` + `packages/core/mpc/keysign/swap` are 28 files / 444 tests green at your head. CI is green across all four checks.

A note on my own harness in case it saves you a confusing half hour: a bare `npx vitest run <path>` from the repo root picks up no config, so `@vultisig/core-chain/*` does not alias and `parseCancelLimitSwapMemo` arrives `undefined`. Your own test file "fails" 15/21 that way. `--config .config/vitest.core.config.ts` is required and everything passes. Nothing wrong on your end, purely me holding it wrong at first.

## should-fix: the review card and the builder now disagree about what `sourceChain` means

Not a blocker and not really introduced by you, but this PR is what makes it visible, and both surfaces are exported from the SDK with no in-repo consumer to keep them honest.

`getKeysignLimitSwapCancel` fills its card with `sourceChain: getThorchainMemoAssetChain(cancel.sourceAsset)` - the **home** chain. The builder now requires the **sending** chain. For the three THORChain-held flavours those are different:

```
                        card.sourceChain    builder requires
L1 dotted   BTC.BTC     Bitcoin             Bitcoin      same
synth       BTC/BTC     Bitcoin             THORChain    DIFF
trade       BTC~BTC     Bitcoin             THORChain    DIFF
secured ETH-USDC-0X…    Ethereum            THORChain    DIFF
```

Both values are defensible in isolation: a BTC synth really is BTC-denominated, and the cancel really does leave from THORChain. The problem is they are the same field name on two surfaces of the same flow. `packages/sdk/src/index.ts:554-555` exports the card type and the getter publicly, and nothing in this repo consumes them, so the only readers are downstream clients. The natural thing for a client to write is "read `cancel.sourceChain`, pick that chain's gas coin, call the builder" - and that is a guaranteed throw for every synth, trade and secured order.

Cheapest fix is naming: keep `sourceChain` as the asset's home chain and add `signingChain: getThorchainMemoAssetSourceChain(cancel.sourceAsset)` to the card, so the client has the field it actually needs and the two cannot be confused. That also gives the review UI something honest to render, since "cancel this Bitcoin order" and "sign this from THORChain" are both things the user should see.

## nit: the `undefined` throw trades one failure for the other one this file warns about

The file's own doc comment says stranding an order for the rest of its TTL is "the worse failure of the two", and `sourceChain === undefined` is now a hard throw. So a memo whose prefix this 18-entry map does not carry becomes uncancellable.

I convinced myself this is not reachable for an order placed through this SDK: `limitSwapInbound.ts:61` and `:88` gate placement on `thorchainMemoAssetChainPrefix`, the same map the cancel path inverts, so anything placeable here is resolvable here. The exposure is cross-client only - an order placed by iOS/Android/web on a chain THORChain routes but this union misses. Given the union is deliberately conservative (the comment explains why `nativeSwapChainIds` is not used) that window is small and refusing is still the right default over signing from a guessed chain.

Worth one line in the throw pointing at the map, though, so whoever hits it in a support thread knows the fix is an entry in `thorchainMemoAssetChainPrefix` and not something about their vault.

## nit

`isThorchainSource` moving from `areEqualCoins(signingCoin, chainFeeCoin[Chain.THORChain])` to `sourceChain === Chain.THORChain` is behavior-preserving because `isFeeCoin(signingCoin)` already ran above it, so a THORChain source implies RUNE. Correct, and simpler. But it now depends on a guard fifteen lines up rather than on anything local, so if the `isFeeCoin` check ever moves or becomes conditional this silently widens. Half a comment.

The renamed test-file comment ("Named by FUNDING chain") is a good catch and exactly the kind of thing that gets missed when a contract changes underneath a fixture name.

Blocker cleared. Nice work on this one.

PR #1698, opened 2026-08-03
fix(fund-safety): wire assertSafeDestination into prepareSendTxFromKeys + own-token-contract guard

NeOMakinG:
Closes beads vultisig-ops-2urb + vultisig-ops-i9qum.

🤖 Agent-generated

## Summary

- Wire `assertSafeDestination` into `prepareSendTxFromKeys` (the plain send/sign path) — previously the guard only fired on CCTP and Across bridge primitives. Burn/dead/zero addresses now fail-closed before any keysign payload is built.
- Add an own-token-contract guard: ERC-20 sends where the recipient equals the token's own contract address (`coin.id === receiver`) now throw before building. Classic unrecoverable-burn vector.
- Prototype-safe map access in `getEvmDangerousReason` + `getChainDangerousReason` (#1193 review should-fix).
- `getEvmDangerousReason` now trims its input so whitespace-padded burn addresses are caught at the primitive level.
- `getChainDangerousReason` now accepts `'xrp'` as an alias for `'Ripple'` (#1193 review should-fix).

### Intent

Prevent funds from being irreversibly lost by refusing to build a send transaction when the destination is a known burn address or the token's own contract address.

### Verify steps

Use vault `Testwallet2` (pw `azerty`) via the CLI. Build with `node build.mjs` from `clients/cli/` after `yarn workspace @vultisig/sdk build`.

1. `echo "azerty" | node clients/cli/dist/index.js send ethereum 0x000000000000000000000000000000000000dEaD 0.001 --vault Testwallet2 --dry-run`
2. `echo "azerty" | node clients/cli/dist/index.js send ethereum 0x0000000000000000000000000000000000000000 0.001 --vault Testwallet2 --dry-run`
3. `echo "azerty" | node clients/cli/dist/index.js send ethereum 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 1 --token 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 --vault Testwallet2 --dry-run`
4. `echo "azerty" | node clients/cli/dist/index.js send ethereum 0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045 0.001 --vault Testwallet2 --dry-run`

### Expected outcomes

1. Error: `dead address (commonly used burn address): funds are unrecoverable`
2. Error: `zero address (ETH burn address): funds sent here are permanently destroyed`
3. Error: `recipient 0xa0b86991... is the token contract address for USDC`
4. Success: builds a dry-run preview for 0.001 ETH (control — must still build)

### Edge cases

- Checksummed `0xDEaD` vs lowercase — covered by `EVM_ADDRESS_RE /i` + lowercase normalization
- Own-token-contract with different casing — tested via `.toLowerCase()` in unit suite
- Native ETH send (no `coin.id`) — own-token-contract guard does NOT fire (guarded by `coin.id !== undefined`)
- ERC-20 to a different valid address — allowed (unit test confirms)

### Prereqs

No vault funds needed — all commands use `--dry-run`. Testwallet2 must be in `~/.vultisig/`.

### Suggested flows

No Maestro flows needed — CLI dry-run covers the deterministic guard.

---

## Receipts (live CLI, vault-CLI 3.0.0)

```
# 1. ETH to 0xdead — REFUSED
{"success":false,"v":1,"error":{"code":"USAGE_ERROR","message":"Failed to prepare send transaction: Refusing to build transaction: destination 0x000000000000000000000000000000000000dEaD is a dead address (commonly used burn address): funds are unrecoverable.","retryable":false}}

# 2. ETH to zero address — REFUSED
{"success":false,"v":1,"error":{"code":"USAGE_ERROR","message":"Failed to prepare send transaction: Refusing to build transaction: destination 0x0000000000000000000000000000000000000000 is a zero address (ETH burn address): funds sent here are permanently destroyed.","retryable":false}}

# 3. USDC to USDC contract — REFUSED
{"success":false,"v":1,"error":{"code":"USAGE_ERROR","message":"Failed to prepare send transaction: Refusing to build transaction: recipient 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 is the token contract address for USDC. Sending tokens to their own contract is almost certainly a mistake and funds would be unrecoverable.","retryable":false}}

# 4. ETH to vitalik.eth (control) — ALLOWED
{"success":true,"v":1,"data":{"dryRun":true,"chain":"Ethereum","to":"0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045","amount":"0.001","symbol":"ETH","fee":"0.000029810497409","feeSymbol":"ETH"}}
```

## Test counts

`dangerousAddresses.test.ts`: 24 → 28 tests  
`send.test.ts`: 7 → 20 tests  
Full unit suite: **3121 passed**, 13 pre-existing failures in `entry.test.ts` + `ensureMpcEngine.test.ts` (MPC singleton test isolation — present on main before this change).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards to prevent transactions to known dangerous or burn addresses.
  * Token transfers are blocked when the recipient matches the token’s contract address.
  * Improved address validation, whitespace handling, case normalization, and chain alias support.
  * Preserved valid native-coin and token transfers to legitimate destinations.
* **Tests**
  * Added coverage for dangerous-address detection and valid transfer scenarios.


gomesalexandre (approved):
Approving — this closes a real fund-loss path and the guard behaves correctly in both directions. One sibling worth a follow-up, and I have ruled out the ones that look like siblings but are not, so nobody chases them.

## verified

Probed the live guard with controls both ways:

```
MUST BLOCK
  BLOCKED  Ethereum 0x0000…0000                              (zero)
  BLOCKED  Ethereum 0x0000…dEaD                              (dead)
  BLOCKED  Ethereum "  0x0000…dEaD  "                        (whitespace-padded — the trim fix)

CONTROL: legitimate destinations must PASS (direction B)
  passes   0x9D7C407D02D8daEe3Cc3800d434E9eA594163003
  passes   0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48        (a real token contract, not the one being sent)

xrp alias
  getChainDangerousReason('xrp',  <ACCOUNT_ZERO>) -> "XRP Ledger black-hole address…"
  getChainDangerousReason('Ripple', <same>)       -> identical

prototype-safe access
  getEvmDangerousReason('__proto__' | 'constructor' | 'toString') -> undefined
```

The direction-B control matters most here: a burn guard that false-blocks a legitimate recipient is worse than the bug, because it strands a user mid-send with no workaround. It doesn't.

The own-token-contract guard is a good catch — sending USDC to the USDC contract is a genuine unrecoverable-burn vector that no address blocklist would ever catch, since the address is perfectly valid. Scoping it to `coin.id !== undefined` correctly leaves native sends alone.

## should-fix (follow-up, not blocking): `maxSend` is the one real sibling

This PR wires the guard into `prepareSendTxFromKeys`. I checked every other prep path that takes a recipient, and whether it routes through that function:

```
routes through send.ts (covered transitively):
  trc20, splTransfer, suiTokenTransfer, polkadotAssetSend

does NOT route through it:
  maxSend, utxoConsolidate, contractCall,
  cw20Transfer, ibcTransfer, jettonTransfer, thorchainMsgDeposit
```

Most of that second list does not matter, and here is why — `getChainDangerousReason` only covers **EVM, Solana, UTXO and Ripple/XRP**; every other chain falls through to `return undefined`. So:

- **cw20Transfer / ibcTransfer** (Cosmos), **jettonTransfer** (TON), **thorchainMsgDeposit** — wiring the guard there buys nothing today, because there is no burn list for those chains. Worth doing only if those lists ever get populated.
- **utxoConsolidate** — a send-max-**to-self**; the recipient is the self-address, not attacker- or user-supplied. Not a vector.
- **contractCall** — the "recipient" *is* a contract by definition. Wiring a burn guard there would be the direction-B mistake, not a fix. Leave it.

That leaves **`maxSend`**: it takes an arbitrary `receiver: string`, does not route through `prepareSendTxFromKeys`, and runs on the exact chains the guard covers. It is also the worst possible case for this bug class — a "send max" to a burn address loses the entire balance rather than a typo-sized amount.

One line in `getMaxSendAmountFromKeys` mirroring what you did in `send.ts` would close it. Not blocking this PR, since it is strictly better than the status quo either way.

## the shape worth naming

The bug you fixed was "the guard existed but only fired on CCTP and Across." The fix wires one more path. That is still an enumerate-the-call-sites approach, and `maxSend` is the proof it does not fully converge — the same class recurs until the guard sits at a chokepoint every send path must pass through, or a test asserts every exported prep function calls it. Worth considering the second one; it is cheap and it fails loudly when someone adds prep path number thirteen.

🤖 reviewed via Claude Code

Issue #1130, opened 2026-07-10
Swap: complete + dynamic secured-asset catalog for THORChain (shared source of truth)


realpaaao:
## Goal

Provide a **complete, dynamic** list of THORChain secured assets as the single source of truth for the swap destination picker across all clients, and surface them in the extension / Windows swap view.

This is the shared root fix for the coordinated set — iOS and Android (sibling issues below) consume the list this issue produces. **Windows / extension is tracked here.**

## Why this is small

Swapping into a secured asset already works: THORChain mints it as the swap's settlement output (`=:ETH-USDC:thor1…`), and `toNativeSwapAsset` already bridges secured denoms → native swap notation (`packages/core/chain/swap/native/asset/toNativeSwapAsset.ts`). The extension swap picker already has the plumbing to surface non-held assets. The only gap is that the secured-asset list is **static and partial**.

## Current state (verified)

- `packages/core/chain/coin/knownTokens/thorchain.ts` **hardcodes ~15 secured-asset denoms** (`btc-btc`, `eth-eth`, `eth-usdc-0x…`, `eth-usdt-0x…`, `bsc-bnb`, `base-usdc-0x…`, `gaia-atom`, …) plus Rujira / LP tokens. This is what feeds the extension picker's `knownTokens[Chain.THORChain]`.
- The Windows / ext swap picker already merges vault coins + `knownTokens[chain]` + `whitelisted` and calls `createCoin()` on select (`vultisig-windows core/ui/vault/swap/form/SwapCoinsExplorer.tsx:54-65,121`) — so the plumbing to surface non-held secured assets **exists**; the list is just static and incomplete.

Consequence: any secured asset THORChain adds is silently missing from the swap view until someone edits `thorchain.ts`, and iOS/Android have no shared list to consume at all.

## Scope / acceptance

- A **dynamic** secured-asset universe derived from THORChain pools (on-device fetch; static list only as offline fallback), replacing/backing the hardcoded secured entries in `knownTokens/thorchain.ts`.
- Exposed via the SDK so iOS + Android consume the same list (sibling issues).
- Extension / Windows swap picker shows the full secured set, clearly labeled, selectable as destinations; existing `createCoin` + `toNativeSwapAsset` + swap → mint-on-settle path handles the rest — **no new tx type**.
- Same-underlying-asset case should prefer SECURE+ mint over a 1:1 pool swap.

## Out of scope

Rujira / FIN orderbook trading (separate track).

[...]

PR #1703, opened 2026-08-03
feat: add dynamic THORChain secured asset catalog

rcoderdev:
## Summary

- Fetches and caches THORChain’s authoritative secured-asset catalog, with an explicit offline picker fallback.
- Exposes one secured-asset catalog through core, SDK, React Native, and `Vultisig.getSwapDestinationTokens`.
- Keeps same-underlying secured destinations on THORChain’s ordinary pool quote path until THORNode provides an authoritative `SECURE+` deposit preflight.

Part of #1130.

## Runtime proof

A live THORChain read returned 38 secured assets and 64 destination entries. A built high-level quote for ETH.USDC to secured THOR USDC used the ordinary THORChain `=:` memo and returned explicit fees; no funds were submitted.

## Summary by CodeRabbit

* **New Features**
  * Added dynamic THORChain secured-asset discovery for swap destinations.
  * Added offline fallback support when live asset data is unavailable.
  * Added cached catalog retrieval with refresh handling and asset deduplication.
  * Added secured-asset metadata, including underlying L1 asset information.
  * Added `getSwapDestinationTokens` to retrieve destination tokens across supported chains.
* **Bug Fixes**
  * Improved secured-destination swap quote resolution and standard memo handling.
* **Documentation**
  * Documented dynamic swap destination retrieval and secured-token metadata.
* **Tests**
  * Added coverage for parsing, fallback behavior, caching, refreshes, exports, and quote selection.


gomesalexandre (approved):
Remote+exact-head pass on `a9ddce294196bdd0dfdc8908b06131b5296ccb50`. I do not see a current-head blocker.

What I verified:
- the dynamic secured-asset catalog is now the single picker source for secured destinations, while THORChain-native / Rujira entries stay on the static path, so the new universe is additive rather than replacing unrelated destination types.
- `forceRefresh` no longer reuses an older in-flight read. The dedicated test covers the exact stale-read race and proves the newest snapshot wins.
- the latest same-underlying route guard stays on the authoritative native THORChain pool path. The new `findSwapQuote.securedDestination.test.ts` asserts the memo remains `=:` and that `getSwapKitQuote` is not called when the pool quote wins.
- export surface is wired through the root SDK and RN entrypoints, including `getThorchainSecuredAssetCatalog`, `getThorchainSwapDestinationAssets`, `getThorchainSecuredAssetL1Asset`, and `parseThorchainSecuredAssets`.

Local exact-head receipts from a fresh disposable clone:
- `yarn install --immutable`
- `yarn build:all`
- `./node_modules/.bin/vitest run packages/core/chain/chains/cosmos/thor/securedAssets.test.ts packages/core/chain/swap/quote/findSwapQuote.securedDestination.test.ts packages/sdk/tests/unit/Vultisig.static-methods.test.ts packages/sdk/tests/unit/platforms/react-native/entry.test.ts packages/sdk/tests/unit/utils/publicExports.test.ts`
- result: `5 files passed, 76 tests passed`

Non-blocking: the fallback catalog is still cached for the full default TTL, so one transient `/thorchain/securedassets` failure can leave picker clients on the static fallback for up to 5 minutes even after the endpoint recovers. That feels acceptable for this landing because the branch now clearly labels fallback vs live source and the source-of-truth/export wiring itself is sound.

PR #1717, opened 2026-08-04
fix: harden local sensitive file paths

neavra:
## Summary

- create Node storage temp files exclusively before writing keyshare data
- create credential and CLI config directories with owner-only permissions
- best-effort tighten permissions on directories created by earlier versions
- add filesystem regression tests for collisions, round trips, creation modes, and upgrades

## Validation

- `yarn install`
- `yarn build:shared && yarn build:sdk && yarn build:client-shared && yarn build:rujira && yarn cli:build`
- `yarn typecheck`
- `yarn test:cli`
- `yarn test:unit`
- `yarn quality:complexity`
- ESLint on changed source and test files
- `yarn lint`
- `yarn test`

## Summary by CodeRabbit

* **Security**
  * Credential and CLI configuration directories now use owner-only permissions.
  * Existing directories are tightened automatically to help protect sensitive data.
  * Credential files continue using restrictive permissions.

* **Bug Fixes**
  * Temporary storage files now refuse to overwrite existing files, preventing accidental data replacement.
  * Storage operations continue to support reliable write and read behavior.


gomesalexandre (approved):
@neavra approving — and the one-line change in `storage.ts` is the sharpest part of this, so I want to say why rather than let it read as a formatting tweak.

`flag: 'wx'` is `O_CREAT|O_EXCL`: the write refuses if the path already exists. For a temp file that keyshare data is about to be written into, that is the difference between "we overwrite whatever is there" and "we refuse to touch a path someone else prepared". Pre-creating that path — as a regular file or a symlink pointing somewhere readable — is the classic way to get a secret written where you can read it. One flag, whole class closed.

The directory work is equally careful about the case people usually miss: `mkdir(dir, { recursive: true, mode: 0o700 })` protects new installs, and the `createdDir === undefined` branch is what protects **existing** ones. That is a correct read of the Node API — `recursive: true` returns the first created path, or `undefined` when nothing was created — so it detects "this predates the fix" precisely rather than by guesswork, and the `chmod` is best-effort with a comment explaining why (filesystems without POSIX permissions). Users who installed before this lands get tightened rather than left behind.

Verified rather than eyeballed, including must-fail controls on both halves — a permissions test that cannot fail is worse than none:

```
credential-store-permissions + config tests   7/7 pass
sdk storage tests                             5/5 pass

mutation: drop `mode: 0o700` from mkdir
  × creates the credential directory with owner-only permissions

mutation: drop `flag: 'wx'`
  × refuses to overwrite a pre-existing temp path

both restored -> green
```

Both guards are load-bearing: each mutation kills exactly the test written for it, so the coverage is real and a future refactor that quietly drops either will go red.

No notes. APPROVE.

PR #1718, opened 2026-08-04
fix(cli): make help and swap errors truthful

neavra:
## Summary

- make `send` and `swap` help distinguish interactive previews from the explicit flags required in non-interactive sessions
- make the `addresses` help describe the collection the command actually returns, without advertising an unsupported chain argument
- render amount-too-small swap errors once while preserving the actionable guidance
- keep MCP descriptions transport-neutral while exposing the CLI-specific guidance only in CLI help

## Swap quote JSON investigation

The current JSON shape is intentionally unchanged. Each similarly named layer carries distinct data:

- `data.quote` is the SDK-enriched result with estimated output, provider, fees, coin metadata, balance, and maximum swappable amount
- `data.quote.quote` is the core quote with the distinct `discounts` field
- `data.quote.quote.quote` is the native/general provider payload consumed by transaction preparation and signing

Flattening these layers would merge public SDK and core contracts rather than remove an empty wrapper.

## Validation

- `yarn install`
- `yarn build:shared && yarn build:sdk && yarn build:client-shared && yarn build:rujira && yarn cli:build`
- `yarn typecheck`
- `yarn test:cli` (980 tests)
- `cd packages/sdk && yarn vitest run --config tests/unit/vitest.config.ts tests/unit` (3,163 tests)
- focused core quote selection tests (58 tests)
- `yarn quality:complexity`
- ESLint scoped to changed TypeScript files
- Prettier check scoped to changed TypeScript files
- SDK bundled-source changeset guard
- built CLI help exercised for `send`, `swap`, and `addresses`

All added and updated tests are network-isolated.

## Summary by CodeRabbit

* **Improvements**
  * Clarified `send` and `swap` CLI help with accurate interactive preview and non-interactive `--dry-run`/`--confirm` instructions.
  * Updated `addresses` command documentation to accurately describe supported usage.
  * Improved transaction descriptions for clearer confirmation behavior.

* **Bug Fixes**
  * Simplified the insufficient-amount swap error message to: “Please increase the amount to proceed.”
  * Removed duplicate wording from related swap error details.


gomesalexandre (approved):
@neavra approving. The split between `description` and `cliDescription` is the right fix and worth spelling out, because it is easy to read as a copy tweak: the shared string was telling **MCP** consumers to "pass `--confirm`" — a flag that does not exist on that transport. A tool description that names an affordance the caller cannot use is worse than a vague one, because the model will try it and then explain the failure to the user. Transport-neutral `confirmed: true` for the shared surface, flag syntax confined to CLI help, is exactly right.

What makes `helpTruthfulness.test.ts` a good test rather than a snapshot: it does not assert the help *text* in isolation, it cross-checks the claim against behaviour. The `addresses` case asserts the help no longer advertises `<chain>` **and** that passing an argument is actually rejected. That is the same property that makes the memo-destination check in #1720 trustworthy — one derivation validated against the other, so the doc cannot drift from the command while the test stays green.

## on my local run, so nobody re-chases it

The three cases fail in my tree and it is my environment, not your diff:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
  '.../node_modules/@vultisig/core-chain/dist/chains/ripple/address.js'
  imported from clients/cli/src/commands/transaction.ts
```

The suite spawns the real CLI through `tsx`, so it needs the workspace packages built; mine are not, and the child dies before it parses a flag — surfacing as `expected 1 to be +0`, which looks like an assertion failure rather than a missing build. I deliberately did not `yarn build:*` into this clone (an agent damaged the workspace symlinks here earlier this week and I would rather not risk it for a review).

Your CI ran it properly and is green across the board — Unit 23m23s, Integration 4m46s, Lint and Type Check 16m53s, CodeRabbit clean.

So: design reviewed here, execution taken from CI, and I am flagging that split rather than implying I ran it.

APPROVE.

PR #1719, opened 2026-08-04
feat(cli): expose SDK prep helpers

### LINKED ISSUE #1161: arch: CLI has zero reachable path to 9 of the SDK's own tx-building prep helpers
\`clients/cli\` (\`@vultisig/cli\`) is the intended isolated-testing surface for SDK tx-building logic, but 9 of the \`packages/sdk/src/tools/prep/*.ts\` helpers — the ones agent-backend-ts calls directly — have zero CLI-reachable path: \`contractCall.ts\`, \`ibcTransfer.ts\`, \`splTransfer.ts\`, \`trc20.ts\`, \`jettonTransfer.ts\`, \`suiTokenTransfer.ts\`, \`polkadotAssetSend.ts\`, \`cosmosStaking.ts\`, \`cw20Transfer.ts\`. Each already has SDK unit-test coverage for encoding correctness, but never gets exercised as a real subprocess with real arg parsing the way an operator/agent actually uses it.

Caveat: the CLI's \`agent ask\`/\`--via-agent\` mode mostly delegates tx-building to the REMOTE agent-backend over SSE rather than exercising the SDK locally, so it doesn't add real isolated coverage here — these 9 gaps need direct \`vultisig\` subcommands.

Full findings + effort estimates per helper + a full agent-ready prompt for the top 3 (SPL/TRC20/Jetton token transfers): [round-5 architecture artifact](https://claude.ai/code/artifact/13bddb9f-4c87-440a-aae8-e30356f5a1a1).

Severity: LOW (test-infrastructure gap, not a live bug). Effort: S per helper, M for the harder ones (contractCall, cosmosStaking).

rcoderdev:
Closes #1161

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands for preparing unsigned contract calls, token transfers, staking, and asset operations across EVM, IBC, Solana, TRON, TON, Sui, Polkadot, Cosmos, and CosmWasm networks.
  * Supports active-vault identities or supplied public identities, standardized JSON output, and validation before signing or broadcasting.
  * Added an option to replace a compatible existing vault share during import.

* **Documentation**
  * Added command usage guidance, examples, help options, supported operations, and response format details.


gomesalexandre (dismissed):
@rcoderdev the CLI surface itself is well built — `unsigned: true` on every result, prep-only, no signing path, and the subprocess tests are real. My concern is not the commands, it is *which* helpers are being promoted to a user-facing surface.

## preferably-blocking: `prep spl-transfer` exposes a path with no burn-destination guard

`buildSplTransfer` does not call `assertSafeDestination`. `prepareSendTxFromKeys` does (`send.ts:69`), and Solana is a covered family — `dangerousAddresses.ts` carries a Solana burn list naming the System Program and the Incinerator explicitly. So the guarded send path refuses those destinations and this one does not:

```
CONTROL normal address                          BUILT
system program  11111111111111111111111111111111  BUILT     <- SDK's own list: "no private key controls this address"
token program   TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA  BUILT
incinerator     1nc1nerator1111…                REFUSED with an EMPTY message
```

The incinerator line is the one I would not lean on: it refuses, but with no message, which reads like an unhandled derivation error rather than a deliberate check — it happens to fail, it is not *refused*. The System Program and Token Program build cleanly, and those are unrecoverable destinations.

The file is honest about what it is — `splTransfer.ts:105` calls itself "the SPL escape-hatch sibling to `prepareSendTxFromKeys`". That is a fine thing for an SDK to offer a caller who knows what they are doing. It is a different proposition once `vultisig prep spl-transfer --to <addr>` exists, because the CLI is where a typo or a pasted-from-Discord address arrives.

Two options, either fine by me:

1. call `assertSafeDestination(Chain.Solana, to)` in `executePrepSplTransfer` before building — keeps the SDK escape hatch escape-hatchy and guards the user-facing entry point
2. move the guard into `buildSplTransfer` itself — stricter, and closes it for every caller, at the cost of the escape hatch no longer being one

I would take (1): the SDK-level bypass is deliberate and documented, the CLI is not the place to inherit it.

## not blocking, same shape, worth knowing

`trc20.ts:69` documents the same "unlike `prepareSendTxFromKeys`" bypass. Tron is not in `getChainDangerousReason`'s covered families, so there is no guard to inherit today and nothing regresses — but if Tron burn addresses are ever added, `prep trc20-transfer` inherits the gap for free. Same for the Sui/TON/Polkadot helpers.

## correction I owe you

In my review of #1698 I listed `splTransfer` and `trc20` as "covered transitively" through `prepareSendTxFromKeys`. That was wrong — both are siblings to it, not callers of it, and the doc comments say so plainly. I have corrected that on #1698 too. It does not change #1698's conclusion, but it does mean the unguarded surface there is wider than I described.

gomesalexandre:
Tackled in bedac04e: `executePrepSplTransfer` now calls `assertSafeDestination('Solana', to)` before `buildSplTransfer`, per option (1) above - the SDK-level `buildSplTransfer` escape hatch stays as-is (undocumented callers who know what they're doing keep it), the CLI entry point gets the guard.

Verification:
- Added `refuses to build an SPL transfer to a known burn/program destination` in `prepSubprocess.test.ts` (System Program address `11111111111111111111111111111111`), asserting exit != 0 and `Refusing to build transaction` in the error.
- Mutation-verified: reverted the guard, confirmed the named test fails (builds cleanly, exit 0), restored, confirmed green.
- Full CLI suite: 80 files / 1071 tests passing against a real built workspace (`yarn build:shared && yarn build:sdk:bundle && yarn build:client-shared && yarn build:rujira`, not a vacuous/unbuilt run).
- `tsc --noEmit` and eslint clean on both touched files.

The `trc20.ts` / Sui / TON / Polkadot gap you flagged as not-blocking is unchanged - Tron isn't in `getChainDangerousReason`'s covered families today so there's nothing to inherit yet, leaving that as-is per your review.

Not self-approving - this is a fund-safety guard, leaving it for an independent review.

gomesalexandre (commented):
Re-reviewed the current head remotely via gh only. The CLI prep surface, README/examples, and subprocess coverage line up on this head, and the SPL safe-destination guard follow-up is the right fix for the user-facing prep path. I do not see a fresh correctness or fund-safety blocker on the current diff.\n\nCurrent receipts on bedac04e are green: Unit Tests, Integration Tests, Lint and Type Check, and Test Summary all pass.\n\nImportant four-eyes note: this head includes substantive gomesalexandre-authored commits, including the SPL guard follow-up, so I cannot approve it. This needs a non-gomes current-head approval before merge.

gomesalexandre (commented):
Re-reviewed the current head in a disposable exact-head clone against fresh `origin/main`.

What I checked on `7a36e357`:
- the CLI prep surface still stays prep-only and returns unsigned payloads, not sign/broadcast flows
- the SPL follow-up guard is now on the user-facing `prep spl-transfer` path, which was the real fund-safety gap on the earlier head
- the added subprocess coverage for the prep command is present on this head
- current GitHub checks are green on this head: Unit Tests, Integration Tests (Vitest), Lint and Type Check, and Test Summary

I do not see a fresh correctness or fund-safety blocker on the current diff.

Important four-eyes note: this head includes a substantive `gomesalexandre` commit (`fix(cli): guard prep spl-transfer against burn/program destinations`), so I cannot approve it. This needs a non-gomes current-head approval before merge.

gomesalexandre (commented):
Re-grounded on `995ad876`. Five commits landed since my last pass and all five are merges from main, so I diffed the PR's own contribution at both heads (merge-base..head) rather than re-reading 946 lines. Effective diff is byte-identical except for exactly four lines, both README example lines. Everything I signed off on at `7a36e357` is unchanged, so this is scoped to what actually moved.

**should-fix (non-blocking): the `prep contract-call` example encodes a selector no ERC-20 implements.**

```bash
vultisig prep contract-call Ethereum <contractAddress> approve --sender <senderAddress> \
  --abi '[{"type":"function","name":"approve","inputs":[]}]' --args '[]'
```

`approve` with `inputs: []` is not ERC-20's `approve`. Selectors, computed with viem:

```
real  approve(address,uint256)  -> 0x095ea7b3
this  approve()                 -> 0x12424e3f
what the example actually encodes: 0x12424e3f
```

So anyone copying it builds an unsigned call to a function that does not exist on any standard token. It reverts rather than doing something dangerous, and this is a prep-only surface with no signing path, so nothing is at risk. But it is the canonical example for a contract-call builder, and `approve` is the one ERC-20 call where people are most likely to copy blindly. Cheap to make it real:

```bash
--abi '[{"type":"function","name":"approve","inputs":[{"type":"address","name":"spender"},{"type":"uint256","name":"amount"}]}]' \
--args '["<spenderAddress>","1000000"]'
```

Two things worth being straight about:

1. **This is not new in this head, and I missed it.** The same `inputs: []` fragment is already in `COMMAND_EXAMPLES` under `prep.contract-call`, so it ships in `--help` too and was there on `7a36e357`. Three prior reviews from me did not catch it. The README edit only made it more prominent, which is what made me look.
2. **The edit itself is an improvement.** Going from `0xContract` / `osmo1...` / `--abi '[...]'` to explicit `<contractAddress>` / `<sourceAddress>` placeholders is clearly better. The elided `'[...]'` was honest as a placeholder; spelling it out is right, the spelled-out version just needs to be a real signature. Worth fixing both sites at once so `--help` and the README agree.

Nothing else changed on this head, and the prep-only contract (unsigned results, no signing or broadcast path, identity via active vault or `--identity`) still holds exactly as reviewed.

Not blocking on the example, but not stamping either while it is the documented way to call the highest-stakes ERC-20 method. Flip me once the ABI fragment is real and I will approve.

gomesalexandre (dismissed):
preferably-blocking: `prep contract-call` still builds value-bearing sends to known-bad EVM destinations with zero guard.

`clients/cli/src/commands/prep.ts` calls `prepareContractCallTxFromKeys(...)` with `contractAddress` and `value`, but unlike `send.ts` and unlike the new `spl-transfer` path in this PR, it never runs `assertSafeDestination(chain, contractAddress)` first. `isValidAddress` is shape-only. `buildSendKeysignPayload` also does not guard this path.

That means the current head still cleanly prepares a value-bearing call to a known burn address, for example a contract call targeting `0x000000000000000000000000000000000000dead`, even though `EVM_DANGEROUS_ADDRESSES` already lists those destinations. This is not the same as the non-blocking future gaps on chains that do not yet have a dangerous-address list. EVM does, so this is a live fail-open on the exact head under review.

The fix should mirror the `spl-transfer` hardening you already added here: guard `contractAddress` with `assertSafeDestination(chain, contractAddress)` before allowing any nonzero `value` path to build.

should-fix: the canonical `approve` example is still wrong in both `clients/cli/README.md` and `clients/cli/src/commands/schema.ts`. It shows `approve` with `inputs: []`, which encodes the wrong selector instead of ERC-20 `approve(address,uint256)`. That is a doc/example bug rather than a signing bug, but it is the one example most people will copy verbatim.

What looks good on this head:
- the `spl-transfer` destination guard is real and has coverage
- the lossless JSON numeric parsing is a solid defensive improvement
- the CLI remains prep-only and returns unsigned payloads instead of signing or broadcasting
- current-head CI is green

rcoderdev:
Addressed both findings on `314deed27`: nonzero native-value `prep contract-call` requests now reject known dangerous EVM destinations before payload construction, with subprocess regression coverage. The README and machine-readable schema now use the real `approve(address,uint256)` ABI and arguments. Current-base `yarn check:ci` and all 12 prep subprocess cases pass. Re-requesting review.

johnnyluo (approved):
LGTM

PR #1720, opened 2026-08-04
fix(swap): preserve THOR and Maya destinations

### LINKED ISSUE #1181: architecture: vault.swap still cannot carry a caller-supplied THOR/Maya destination
## Summary

`vault.swap()` still cannot carry a user-supplied THOR/Maya destination through the public SDK surface. The method re-derives the destination from the vault itself, so downstream consumers have to block cross-account routing or risk silent substitution.

## Evidence

- `packages/sdk/src/vault/VaultBase.ts:1714-1727`
  - the public `swap()` contract only accepts `fromChain`, `fromSymbol`, `toChain`, `toSymbol`, `amount`, `dryRun`
  - it immediately derives `toAddress` from `this.address(toChain)` and builds `toCoin` from that vault-owned address
- `clients/cli/src/agent/executor.ts:897-905`
  - the CLI comments explain the consequence directly: `vault.swap` does not honor the destination encoded in the memo, so Phase D must hard-block anything except self-swaps until the SDK grows a destination override

## Impact

This is a real architecture hole. The SDK is supposed to own THOR/Maya swap construction, but its public contract still cannot safely represent a caller-supplied destination. That keeps downstream memo parsing / routing guards duplicated in consumer code.

## Fix sketch

- extend the SDK THOR/Maya swap contract with an explicit destination override
- keep self-swaps working exactly as before
- fail closed on destination mismatches
- add SDK + CLI fixtures for self-swap vs cross-account destination behavior

## Effort

Medium.

## Report

Round 12 report: https://gist.github.com/gomesalexandre/779031b3ffa23ab072dc33bca5495dc3

rcoderdev:
Closes #1181

## Summary by CodeRabbit

* **New Features**
  * Swap destinations can now be specified for THORChain and MayaChain swaps, including cross-account recipients.
  * Omitted or blank destinations continue to default to the vault address.
  * EVM destination addresses are normalized while preserving valid casing.

* **Bug Fixes**
  * Rejects swap quotes whose executable memo routes funds to a different destination.
  * Rejects destinations containing whitespace.
  * Improved handling of MayaChain swap amounts and Bitcoin recipients.


gomesalexandre (approved):
@rcoderdev this is the right shape for the bug class, and I want to name why: the fix does not just pass the destination further down, it **cross-checks the executable memo against what was requested** and refuses on divergence. A swap where the quote's memo routes somewhere other than the address the caller asked for is precisely how funds land in the wrong place, and validating one derivation against the other is the only way to catch it. Same property that makes a signing ceremony trustworthy.

Ran your suite plus my own direction-B probe, since a throw on a legitimate quote blocks a swap and that is the expensive failure here.

```
getNativeSwapQuote.test.ts   15/15
packages/core/chain/swap     738/738 across 54 files
```

Then the memo shapes THORChain actually emits, against the check as written:

```
exact match (evm)              ok
checksum diff (evm)            ok      <- your test, confirmed
streaming memo w/ limit  =:e:0x…:0/1/0 ok
affiliate segments  =:e:0x…:0/1/0:t:10 ok
thorchain dest (case-sensitive)        ok
no memo                        THROWS  <- your test, correct
memo missing dest segment      THROWS  correct
empty dest segment  =:e::0/1/0 THROWS  correct
```

The streaming and affiliate cases are the ones I most wanted to see survive — `split(':')[2]` keeps working when the memo grows trailing segments, so a streaming swap with `limit/interval/quantity` and an affiliate tail still validates instead of tripping.

The EVM-only lowercasing is also the correct call rather than a shortcut. `findSwapQuote.ts:602` already documents that THORChain memo parsing is case-sensitive (the lowercase-THORName-affiliate guard), so normalising non-EVM destinations would have been the bug, not the fix.

## one case I chased and could not make reachable

A THORName destination (`=:e:mythorname`) throws against a resolved address. I went looking for a path where a caller supplies a name and the provider echoes an address, which would false-block a legitimate swap — there is none: THORNames appear only as `affiliateFeeAddress` in this repo, never as a swap destination, and nothing resolves names before quoting. Noting it because if name destinations are ever supported, this check needs a resolution step before the comparison.

APPROVE.

PR #1735, opened 2026-08-04
fix(cli): wire recipientSanity into send prep — refuse null/burn/self-send/malformed-EVM (bead 51exn)

NeOMakinG:
## Summary
- Wires the already-exported `recipientSanity` helper into vault-CLI's `sendTransaction`
- Refuses null/burn/self-send/malformed-EVM recipients BEFORE any network call
- Fixes bead **vultisig-51exn** (dogfood 2026-08-04)

## Motivation
`packages/sdk/src/tools/validate/recipientSanity.ts` already covers:
- EVM zero, `0x...dEaD`, and `0xdead...42069` burns
- Solana System Program (`11111111111111111111111111111111`), Incinerator, SPL Token Program, Wrapped SOL mint
- BTC null/eater burns, XRP burns
- Self-send + malformed EVM

...but `grep -rn recipientSanity clients/cli/src packages/sdk/src/vault` returns 0 callers. So vault-CLI's send prep silently builds transactions to any of these addresses — funds sent are unrecoverable.

## Test plan (with vault-CLI receipts)

Repro'd on my dogfood worktree BEFORE this change (all built with `warning: None`):
```
vultisig send ethereum 0xdAC17F958D2ee523a2206206994597C13D831ec7 0.0001 --dry-run   # USDT contract
vultisig send ethereum 0x0000000000000000000000000000000000000000 0.0001 --dry-run
vultisig send base     0x0000000000000000000000000000000000000000 0.0001 --dry-run
vultisig send solana   11111111111111111111111111111111 0.001 --dry-run              # Solana System Program
```
All returned `success: true, warning: None`. After this change, all four should throw `Refusing send: recipient is a null / burn address...`.

## Implementation notes
- Insertion point is right after Ripple address normalization, before the first `vault.send({dryRun:true})` network call. Fails fast without wasting a dry-run.
- `from` is read via `vault.address(params.chain)` inside a `.catch(() => '')` so a failed address read never masks a null-recipient block.
- No new tests: `recipientSanity` itself is already unit-tested (`packages/sdk/src/tools/validate/recipientSanity.test.ts`). Happy to add a wire-test if reviewer wants.

Fix via Claude Opus 4.7

## Summary by CodeRabbit

* **Bug Fixes**
  * Added transaction validation to prevent sends to null or burn addresses, the sender’s own address, and malformed EVM addresses.
  * Improved error messages when a transaction recipient is invalid.


gomesalexandre (dismissed):
The gap you found is real and the fix belongs here. `grep -rn recipientSanity clients/cli/src packages/sdk/src/vault` returning zero callers on a helper this specific is exactly the kind of thing that rots quietly, and the insertion point is right: before the first `vault.send()` so nothing hits the network, and after the Ripple normalization so `to` is the resolved r-address rather than the raw X-address. The `.catch(() => '')` on the `from` read is the correct posture too, a failed address lookup must not swallow a null-recipient block.

One thing I want changed before this lands, and one correction to the body.

## preferably-blocking: self-send is a hard throw, and it is the one flag with no funds at risk

The three flags get identical treatment, but they are not the same kind of problem:

- `isNull` - funds are gone forever. Hard refuse is correct.
- `isMalformedEvm` - the send would fail downstream anyway. Refusing early is a strict improvement.
- `isSelfSend` - the money lands in your own vault. Nothing is lost. Ever.

The helper's own docstring names its upstream: `self_send_warning.go:detectSelfSendTurn`. The Go side calls it a warning. This wires it up as a refusal with no way past it, which turns a legitimate operation into a wall.

The one I keep coming back to is UTXO consolidation. `vultisig send bitcoin <my-own-address> <amount>` is an ordinary thing to want to do, and after this it throws. Same shape for a 0-value EVM self-send used to replace a stuck nonce. Neither of those is a user about to lose money, and neither has a workaround, because the error copy tells them to do the one thing that cannot help:

```
If this is intentional, re-run with a different recipient.
```

For a burn address that is the right advice. For a self-send the recipient IS what they meant.

You already have the intent gate for this, twelve lines up in the same function:

```ts
// transaction.ts:157
if (!params.dryRun && !params.yes && isNonInteractive()) {
```

So the narrow ask: keep the throw for `isNull` and `isMalformedEvm`, and let self-send through under `--yes`, or push it onto the `warnings` array that `previewDryRun` already builds at `:97-106` and surfaces through `warning:` at `:127`. Either shape is fine by me. The point is only that the one flag which cannot lose funds should not be the one with no escape hatch.

This is the repo's own fund-safety posture, a build-then-warn beats a false reject, and here the false reject buys nothing.

## should-fix: the body's test plan is 3 for 4, and the case it misses is the interesting one

The first repro line is the USDT contract, and the body says all four should throw after this change. It does not. `0xdAC17F958D2ee523a2206206994597C13D831ec7` is a valid 42-char EVM address, is not in the burn table, and is not the vault's own address, so `recipientSanity` returns clean and the send builds exactly as it did before.

### Receipts

Ran the real `recipientSanity` at your head `48eb565f`. A-rows are the four from your test plan, C-rows are controls that pin both directions.

```
A1 USDT contract (body: THROW)         flagged=false flags=[]
A2 eth zero (body: THROW)              flagged=true  flags=[null]
A3 base zero (body: THROW)             flagged=true  flags=[null]
A4 sol SystemProgram (body: THROW)     flagged=true  flags=[null]
C1 pass-ctrl normal EVM                flagged=false flags=[]
C2 pass-ctrl normal Solana             flagged=false flags=[]
C3 pass-ctrl normal BTC                flagged=false flags=[]
C4 fail-ctrl self-send                 flagged=true  flags=[selfSend]
C5 fail-ctrl malformed evm             flagged=true  flags=[malformedEvm]
```

C1 through C3 are the ones I most wanted to see, because a guard that trips on a normal Solana or BTC recipient would be far worse than the hole it closes. They are clean, and the reason is structural rather than luck: `EVM_PREFIX_RE = /^0x[0-9a-fA-F]+$/` means no non-EVM family can reach the `malformedEvm` branch at all. So no cross-family false positives. Good.

The USDT row is not a bug in your diff, it is a bug in the claim. Detecting "recipient is a token contract" needs an `eth_getCode` lookup, which a pure deterministic helper deliberately cannot do, so it is genuinely out of scope here. Worth either dropping that line from the body or splitting it into its own bead so it does not read as covered when it is not. If the original dogfood report was primarily about the USDT case, that one is still open after this merges.

## should-fix: the wiring is what is new, and nothing tests the wiring

You are right that `recipientSanity` is covered, `tests/unit/tools/validate/recipientSanity.test.ts` is thorough. But the contract this PR adds is not "the helper computes flags", it is "the guard runs before any network call". That is the part that can silently regress if someone moves the block below the dry-run later.

One test with a spied `vault.send` and a burn recipient, asserting it throws and that `send` was never called, pins the thing you actually wrote. You offered it in the body, I am taking you up on it.

## Ruled out so you do not chase them

`to` versus `params.to`: the guard checks `to` (Ripple-normalized) while `vault.send` at `:184` passes `params.to` (raw). That divergence predates this PR and for the guard it is the safer direction, since an X-address wrapping a black-hole r-address resolves before the burn check sees it. Not yours, not a problem here.

The extra `vault.address()` call on every send including dry-runs is fine, it is caught and it is cheap next to the dry-run that follows.

Flip the self-send arm and this is good to go.

gomesalexandre (dismissed):
changes requested.

1. invalid-recipient refusal now escapes as `UNKNOWN_ERROR`/exit 7 instead of the CLI's typed invalid-input path. In `clients/cli/src/commands/transaction.ts:201-202` the new null/burn + malformed-EVM guard throws a plain `Error("Refusing send: ...")`. `clients/cli/src/core/errors.ts:697-743` does not recognize that wording, so it falls through to `UnknownError`. I verified it directly with a focused repro in the exact PR checkout: `classifyError(new Error('Refusing send: recipient looks like a malformed EVM address...'))` returns `UnknownError` with `exitCode === 7`. This is a user-facing contract regression on a fund-safety refusal surface.

2. the new self-send warning is visible in table mode but disappears in json/ci mode. `transaction.ts:206-209` emits the warning via `warn(...)`, but `clients/cli/src/lib/output.ts:77-84,236-239` makes json output imply `silentMode`, so `warn()` becomes a no-op. Focused repro from this head: same `sendTransaction(...)` call to the sender's own address logs `SELF-SEND` in table mode, then emits only the success JSON envelope in json mode with no warning field and no stderr warning. That means scripted callers lose the very warning this PR is adding.

3. missing changeset. This PR changes published CLI behavior under `clients/cli/src/`, but there is no `.changeset/*.md` in the diff. `CLAUDE.md:91-101` is explicit that every PR touching `clients/cli/src/` needs a changeset, no exceptions.

receipts:
- `git diff --name-only origin/main...HEAD` => only `clients/cli/src/commands/transaction.ts`, `clients/cli/src/commands/__tests__/sendRecipientSanity.test.ts`, `clients/cli/src/commands/__tests__/sendDryRunFee.test.ts`
- `yarn build:fast` ✅
- `yarn workspace @vultisig/cli vitest run src/commands/__tests__/sendRecipientSanity.test.ts src/commands/__tests__/sendDryRunFee.test.ts` ✅ (21/21)
- focused repro confirmed the null-address refusal still fires before any network call

once those are fixed I’m happy to re-review.

gomesalexandre (changes requested):
Blocking: this wires a chain-agnostic `recipientSanity()` into *all* CLI sends, but `isMalformedEvmAddress()` flags any `0x` + hex string whose length is not 42 as malformed EVM.

That is fine for Ethereum-style destinations, but it is wrong for Sui. A normal 32-byte Sui address is `0x` + 64 hex chars, so after this PR the CLI will hard-refuse legitimate Sui sends before any network call.

Current head proof from the exact PR checkout:

```json
{
  "recipient": "0x1111111111111111111111111111111111111111111111111111111111111111",
  "flagged": true,
  "flags": ["malformedEvm"],
  "isMalformedEvm": true
}
```

The wiring that makes this user-visible is here:
- `clients/cli/src/commands/transaction.ts:188` calls `recipientSanity({ recipient: to, from: fromAddress })`
- `clients/cli/src/commands/transaction.ts:201-203` turns `isMalformedEvm` into a hard refusal for every chain
- `packages/sdk/src/tools/validate/recipientSanity.ts:106-112` defines malformed EVM as any `0x`-prefixed hex string that is not 42 chars long

So this PR fixes real burn/null/self-send cases, but it also regresses Sui sends into a false-positive hard block.

What I could verify:
- exact-head source read
- exact-head probe of `recipientSanity()` in the PR checkout using the repo tsconfig
- changed tests only cover Ethereum-shaped addresses, so they would not catch this

What I could not verify from this checkout:
- full Vitest/CLI runtime receipts, because the disposable clone was read-only and Yarn/Vitest wanted to write install/config state (`.yarn/install-state.gz`, Vite timestamp files)

Please make the malformed-EVM guard chain-aware at the CLI call site, or otherwise avoid applying the EVM-shape rejection to non-EVM chains.

gomesalexandre:
Tackled in 2ae50d958: the malformed-EVM guard is now chain-aware.

`recipientSanity()` stays pure/chain-agnostic (per its own docstring) - the gate moved to the one call site that actually knows the destination chain (`clients/cli/src/commands/transaction.ts:207`):

```ts
if (sanity.isMalformedEvm && isChainOfKind(params.chain, 'evm')) {
  refusals.push('recipient looks like a malformed EVM address')
}
```

`isChainOfKind` was already exported from `@vultisig/sdk` (re-exports `@vultisig/core-chain/ChainKind`), so no new shared-code surface needed.

Verified both directions with new regression tests in `sendRecipientSanity.test.ts`:
- Same 66-char Sui-shaped `0x1111...1111` string: still refused on `Chain.Ethereum`, now sends cleanly on `Chain.Sui`.
- `0xdeadbeef` still refused on `Chain.Ethereum` (control - confirms the guard didn't get gutted).

`yarn workspace @vultisig/cli vitest run` on the touched suites: 9/9 + 133/133 passing. `yarn workspace @vultisig/cli typecheck` clean. `packages/sdk/tests/unit/tools/validate/recipientSanity.test.ts` (29/29) untouched and still green - confirms the pure helper itself wasn't modified, only the CLI wiring.

Re-requesting review.

gomesalexandre (commented):
Current head looks conceptually sound from a focused exact-head pass, but I cannot stamp it.

Receipts:
- Focused exact-head run: `./node_modules/.bin/vitest run clients/cli/src/commands/__tests__/sendRecipientSanity.test.ts clients/cli/src/commands/__tests__/sendDryRunFee.test.ts clients/cli/src/core/errors.test.ts`
- `clients/cli/src/core/errors.test.ts` passed.
- The two new send suites were blocked by a pre-existing workspace import failure in this disposable checkout: `@vultisig/core-chain/chains/ripple/address` could not resolve from `clients/cli/src/commands/transaction.ts`.
- Static trace of the diff still checks out: `recipientSanity` is wired only into send prep, null/burn + malformed-EVM refuse before network work, self-send downgrades to warning, and the new `refusing send:` classifier correctly maps those refusals to `InvalidAddressError` / exit 4.

I am leaving this as a comment, not an approval, because the branch contains non-merge gomes-authored commits and needs a fresh non-gomes stamp on the current head.

rcoderdev:
Closing this because the exact head still fails open when the vault sender address cannot be resolved, is heavily conflicted, and lacks complete current-head CI and approval; rebuild the recipient guard narrowly from current main with sender lookup failing closed.

PR #1736, opened 2026-08-04
fix(cli): de-duplicate token send consent summary

neavra:
## Summary

- parse rich token labels into a bare ticker and optional contract disclosure
- keep the displayed chain anchored to the routed transaction chain
- preserve native-send wording and bare-token fallback behavior
- add a CLI patch changeset

## Validation

- yarn install
- yarn cli:build
- yarn typecheck
- yarn test:cli — 72 files, 992 tests passed
- yarn quality:complexity
- yarn build:all
- yarn lint
- yarn test — all script, SDK, Rujira, core, CLI, and MCP workspaces passed
- focused summary suite — 10 tests passed

## Revert sensitivity

Reverting executor lines 406–416 to the prior composition makes the focused suite fail for the intended behavioral reason. The received consent line becomes:

send 0.05 USDC.e USDC.e on Polygon (0x2791…4174) on Polygon to 0x58C4…5C35

Result: 1 failed, 9 passed. The failure is the duplicated ticker and chain, with no network call.

## Runtime receipt

The bundled CLI was invoked against the real agent command without approval flags. The attempt stopped before transaction preparation with AUTH_FAILED because no vault password was available in stored credentials. No signing or broadcast occurred; a live proposal receipt could not be produced without introducing a credential.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved send confirmation summaries so token symbols, chains, contract details, and destination information display clearly.
  - Prevented duplicate token tickers, amounts, and routed-chain details from appearing.
  - Preserved important information from complex, multi-word, malformed, or reordered token labels.
  - Retained routed-chain and contract details when labels include annotations or differ from the routed chain.
  - Kept relevant token details visible in non-swap send locations.


gomesalexandre (dismissed):
The duplication is real and worth killing. `send 0.05 USDC.e USDC.e on Polygon (0x2791…4174) on Polygon to …` is exactly the kind of line that trains people to stop reading the confirmation prompt, and this is the one string in the CLI where that matters most: `session.ts:1206` hands it straight to `ui.requestConfirmation` and nothing else is rendered, so the summary IS the consent surface. Fixing it is right.

I A/B'd the real `getPendingSummary` at your head and at the merge base with only `executor.ts` swapped, across the label shapes the new tests do not cover. Most of it holds. One row goes backwards.

### Receipts

```
                                          merge-base (a2d42588)                                                    PR head (1ae7629a)
C1 pass-ctrl bare ticker                  send 500000 USDC on Base to 0xRecipientAddr                              send 500000 USDC on Base to 0xRecipientAddr
C2 fail-ctrl rich label                   send 0.05 USDC.e USDC.e on Polygon (0x2791…4174) on Polygon to 0x58C4…   send 0.05 USDC.e on Polygon (0x2791…4174) to 0x58C4…
V1 multi-word asset name                  send 0.05 USD Coin USD Coin on Polygon (0x2791…4174) on Polygon to …     send 0.05 USD Coin USD on Polygon Coin (0x2791…4174) to …
V2 amount already embeds the full label   send 0.05 USDC.e on Polygon (0x2791…4174) on Polygon to …                send 0.05 USDC.e on Polygon (0x2791…4174) USDC.e on Polygon (0x2791…4174) to …
V3 lowercase chain in label               send 0.05 USDC.e USDC.e on polygon (0x2791…4174) on Polygon to …         send 0.05 USDC.e on Polygon on polygon (0x2791…4174) to …
V4 chain named twice in label             send 0.05 USDC.e USDC.e on Polygon on Polygon (0x…) on Polygon to …      send 0.05 USDC.e on Polygon on Polygon (0x2791…4174) to …
V5 chain mismatch (pinned by new test)    send 0.05 USDC.e USDC.e on Ethereum (0xA0b8…eB48) on Polygon to …        send 0.05 USDC.e on Polygon on Ethereum (0xA0b8…eB48) to …
V6 no token label at all                  send 0.05 on Polygon to 0x58C4…5C35                                      send 0.05 on Polygon to 0x58C4…5C35
```

C1 and V6 are identical across both arms, which is what makes this a diff of your change rather than a broken harness. C2 is the case you are fixing and it flips exactly as advertised. `src/agent/__tests__/` is 38 files / 607 tests green at your head.

## preferably-blocking: V2 is a regression on the exact duplication this PR removes

Narrowing `symbol` from the whole label to its first word also narrows the de-dup guard it feeds:

```ts
const amountWithSymbol = symbol && !amount.endsWith(` ${symbol}`) ? `${amount} ${symbol}` : amount
```

At the merge base `symbol` was the full `token_resolved`, so when `resolved_amount` happened to carry that same full label the `endsWith` matched and nothing was appended. At your head `symbol` is `USDC.e`, `'0.05 USDC.e on Polygon (0x2791…4174)'` does not end with `' USDC.e'`, and the whole label gets appended a second time. The head prints the ticker, the chain and the contract twice where the merge base printed them once.

It degrades redundant rather than lossy, so your stated invariant holds. But it is a shape that got *worse* under a patch whose entire job is de-duplication, and it is one line:

```ts
const amountWithSymbol =
  symbol && !amount.endsWith(` ${symbol}`) && !amount.endsWith(` ${tokenLabel}`) ? `${amount} ${symbol}` : amount
```

Worth a case in `executor.summary.test.ts` too, since nothing there sets `resolved_amount` to anything but a bare `amount + ticker`.

## should-fix: `split(/\s+/, 1)[0]` assumes word one is the ticker, and V1 shows what happens when it is not

```ts
const symbol = tokenLabel.split(/\s+/, 1)[0]
```

Feed it `USD Coin on Polygon (0x2791…4174)` and `symbol` is `USD`, `tokenDetail` is `Coin (0x2791…4174)`, and the summary reads:

```
send 0.05 USD Coin USD on Polygon Coin (0x2791…4174) to 0x58C4…5C35
```

`Coin` has been relocated to the far side of the chain, so the line now implies an asset called `Coin` at that contract. Technically nothing was dropped, which is the invariant you wrote down, but a fragment moved somewhere it changes meaning is not really the same as a fragment preserved. The merge base was redundant here and still readable; the head is neither.

Your own comment says the label shape is an out-of-repo producer convention. That cuts against assuming the ticker is exactly one whitespace-delimited word. Anchoring on the structure you actually re-render would be sturdier, something like matching `^(.*?)\s+on\s+<chain>\b` first and only falling back to word-one when that fails, or leaving the label untouched whenever it contains whitespace before the `on <chain>` segment.

## should-fix: V5 is the case I would least want pinned as correct

```ts
expect(mismatch).toBe('send 0.05 USDC.e on Polygon on Ethereum (0xA0b8…eB48) to 0x58C4…5C35')
```

I agree completely with the reasoning behind it. A label claiming a chain the router did not use must not be swallowed, and both arms do keep it visible. My problem is with `visible` being treated as sufficient on this particular string.

Read that line the way an operator reads it at 2am. The routed chain is `Polygon`, but the phrase sitting next to the contract is `on Ethereum (0xA0b8…eB48)`, and `0xA0b8…eB48` is USDC on Ethereum mainnet. The most natural parse of the tail is that this transaction is an Ethereum one. The merge base was not better, it just buried the conflict elsewhere in the sentence. Neither shape tells the operator that two sources disagree.

A mismatch between the routed chain and the label's claimed chain is not a rendering quirk, it is a signal. It deserves to look like one:

```
send 0.05 USDC.e on Polygon (0xA0b8…eB48) [label claims Ethereum] to 0x58C4…5C35
```

Or refuse the label's chain fragment entirely and disclose it separately. Either way the test then pins something an operator can act on. As written, the assertion freezes a run-on sentence as the contract, and the next person to touch this will treat it as intended output rather than as the least-bad option.

## nit

V3 and V4 both leave a redundant `on <chain>`: the strip is exact-case and non-global, so `on polygon` and a second `on Polygon` both survive. Consistent with the design and safe in the right direction, so I would not change the behavior, but the comment says "only remove the exact pieces this template re-renders" and a reader could take that to mean case-insensitively or all of them. One clause saying the match is deliberately exact and first-only would save someone a re-derivation.

No runtime receipt on this one per your own note (`AUTH_FAILED`, no vault password). Not blocking given the unit coverage is genuinely good here, but this is the pre-sign consent line, so if you can get a vault open even once, a single screenshot of the real `(y/N)` prompt with a rich token label is worth more than the whole table above.

## q: the recipient in this line is a label, not calldata

Falls out of the A/B rather than being a problem with your diff, and it reproduces identically on both arms, so it is pre-existing and not yours to fix here.

```ts
const to = (p?.txArgs?.to as string) || labels.recipient_echo || '?'
```

For the token shape your own tests use, `txArgs.to` is absent and the recipient rendered is `labels.recipient_echo`. Meanwhile the bytes that get signed are `txArgs.tx.data`, an ERC20 `transfer(address,uint256)` whose first argument is the real destination. Nothing cross-checks the two. I drove an envelope where the calldata transfers to `0xaaaa…aaaa` while `recipient_echo` says `0xGOOD…SAFE`, and both arms render:

```
send 0.05 USDC.e on Polygon (0x2791…4174) to 0xGOOD…SAFE
```

So the field that determines where the money goes is the one field on this line taken purely on trust, while the diff here is carefully protecting the contract disclosure. That asymmetry is worth a bead: decode the `transfer` recipient from `tx.data` for ERC20 sends and render that, or at minimum flag when it disagrees with the echo. Happy to file it if you have not already, just say the word and I will not touch this PR's scope with it.

Fix the V2 append and I am happy to stamp. The rest is yours to take or leave.

gomesalexandre (commented):
conceptual pass looks good. i checked the executor summary fix against the new pinned matrix and the load-bearing `amountEmbedsTokenLabel` guard does what it should: it stops re-appending the rich token label when `resolved_amount` already carries it, while still keeping the routed chain visible in the consent summary. the duplicate-beats-omit fallback on unparsed labels is the right fail-closed tradeoff here.

not approving because this branch now contains a gomes-authored commit (`003d9942d58b5d1a4263e8486edc689d9a2b05bc`), so it needs a non-gomes approval for four-eyes.

also non-blocking: current ci red looks unrelated to this diff. the lint/type gate is tripping on the js-yaml advisory in transitive deps, so this branch should rebase onto fresh main before merge.

gomesalexandre (changes requested):
blocking: there is still one duplicated-chain shape in the consent line.

when the rich token label is reordered and the full label is already embedded in `resolved_amount`, `primaryRoutedChainPattern` misses the positive routed-chain mention because it only matches right after the ticker. current head still prints the chain twice:

```text
token_resolved:  USDC.e (0x2791…4174) on Polygon
resolved_amount: 0.05 USDC.e (0x2791…4174) on Polygon
summary:         send 0.05 USDC.e (0x2791…4174) on Polygon on Polygon to 0x58C4…5C35
```

that is the same class of duplication this PR is trying to remove, just in the reordered-label variant. the current matrix covers reordered labels and fully-embedded labels separately, but it misses their combination.

receipts on exact head `44ce996762029cb73a00fadc038fc1e87a76debd` from a fresh disposable clone:

- `yarn build:all` ✅
- `yarn vitest run clients/cli/src/agent/__tests__/executor.summary.test.ts` ✅ 18 tests passed
- direct probe of `AgentExecutor.getPendingSummary()` with the inputs above reproduced the duplicated output shown here

please add that combined regression case and adjust the routed-chain detection so a reordered positive `on Polygon` inside the embedded full label suppresses the extra routed-chain suffix, without breaking the existing mismatch / `not on Polygon` cases.

gomesalexandre:
mergefixed + retested in `a9e84cdd`.

what changed
- merged fresh `origin/main` into this branch and resolved the live conflict in `clients/cli/src/agent/executor.ts`
- kept the PR's routed-chain de-dup logic for reordered embedded labels
- kept main's token-contract disclosure for EVM token sends, so the summary now preserves both behaviors on the merged head

receipts
- `yarn install --immutable`
- `yarn build:all`
- `yarn vitest run clients/cli/src/agent/__tests__/executor.summary.test.ts clients/cli/src/agent/__tests__/sessionConfirmGate.test.ts` ✅ 41 passed
- direct exact-head probe for the previously-blocking shape:
  - input: `resolved_amount = token_resolved = "0.05 USDC.e (0x2791…4174) on Polygon"`
  - output: `send 0.05 USDC.e (0x2791…4174) on Polygon to 0x58C4…5C35 (token contract 0x2791…4174)`
  - `on Polygon` now appears once, not twice

not approving. this branch contains gomes-authored commits, so it needs a non-gomes current-head stamp for four-eyes.

gomesalexandre (dismissed):
Blocking: the current head still duplicates the token contract in the consent summary on the real EVM token-send shape this PR is trying to clean up.

Grounded path at `clients/cli/src/agent/executor.ts`:
- the rich label branch can already carry the contract inside `token_resolved` / `resolved_amount` as `USDC.e on Polygon (0x2791…4174)`
- later, `contractPart` independently appends ` (token contract ${contractTo})` whenever `txArgs.tx.to` differs from the recipient and `txArgs.tx.data` is non-empty

That means a realistic ERC-20 send summary becomes:

```text
send 0.05 USDC.e on Polygon (0x2791…4174) to 0x58C4…5C35 (token contract 0x2791…4174)
```

So the same contract is disclosed twice, which breaks the PR's own “exactly once” invariant.

I verified this against the current head logic with the real shape that matters here:
- `resolved_amount = '0.05 USDC.e on Polygon (0x2791…4174)'`
- `token_resolved = 'USDC.e on Polygon (0x2791…4174)'`
- `txArgs.to = recipient`
- `txArgs.tx.to = contract`
- `txArgs.tx.data = non-empty transfer calldata`

Why the test suite misses it:
- every new rich-label fixture in `executor.summary.test.ts` leaves `tx.data` empty / absent, so `isContractSend` stays false and `contractPart` never runs
- the existing calldata-bearing coverage uses only bare `token_resolved`, so it does not exercise the overlap between rich-label rendering and payload-derived contract disclosure

Please add a regression that combines all four pieces at once:
- rich `token_resolved`
- recipient in `txArgs.to`
- contract in `txArgs.tx.to`
- non-empty transfer calldata in `txArgs.tx.data`

Then assert the ticker, routed chain, and contract disclosure each render once while keeping the payload-derived contract as the authoritative source.

gomesalexandre (commented):
Blocker cleared. My CHANGES_REQUESTED from 2026-08-08 was that a realistic ERC-20 send still disclosed the contract twice:

```
send 0.05 USDC.e on Polygon (0x2791…4174) to 0x58C4…5C35 (token contract 0x2791…4174)
```

`a2d97851` + `dba51d06` fix it, and I verified rather than took it on trust.

## the truncation format is the real one

This mattered more than it looks. The strip only matches `0x([0-9a-fA-F]{2,8})…([0-9a-fA-F]{2,8})`, so if the producer abbreviated differently the regex would quietly no-op. Those labels come from agent-backend-ts, not the CLI, so I checked the actual producer rather than the CLI's own `cards.ts` truncator (which is a different `slice(0,8)…slice(-6)` shape and would have been the wrong thing to validate against).

`execute_send.ts:1358` in agent-backend-ts:

```ts
function truncateAddress(addr: string): string {
  if (addr.length <= 14) return addr
  return `${addr.slice(0, 6)}…${addr.slice(-4)}`
}
```

and `token_resolved` is built as `${tokenSymbol} on ${chain} (${truncateAddress(contractAddress)})` at `:2864`. For `0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174` that is `0x2791…4174`: 4 hex prefix, 4 hex suffix, U+2026 ellipsis. Inside `{2,8}` on both sides, and the exact literal your test uses. So the test fixture is the production shape, not an invented one.

## the coverage is real, not vacuous

Must-pass, both named cases green:

```
✓ strips a truncated payload-derived ERC-20 contract from rich token labels
✓ preserves a truncated non-contract address in rich token labels
  Tests  32 passed (32)
```

Must-fail control, because a green row proves nothing on its own. I neutered the new branch by making its pattern unmatchable and re-ran:

```
FAIL  strips a truncated payload-derived ERC-20 contract from rich token labels
  Tests  1 failed | 31 passed (32)
```

Exactly one test dies, and it is the one that covers the fix. Restored the file afterwards, working tree clean. The sibling "preserves a truncated non-contract address" case is the right guard against over-stripping, and it stays green under the mutation, which is what you want.

## dismissing my review rather than approving

This head carries a `gomesalexandre` commit from an earlier round, so an approve from me would not be four eyes. My objection is resolved though, and leaving CHANGES_REQUESTED standing would block you on someone who cannot unblock you. Dismissing it so another reviewer can stamp.

One thing to know for merge timing, unrelated to your branch: this PR's `Lint and Type Check` green completed `2026-08-10T15:11Z`, before GHSA-2v37-7h3g-55p8 widened to `nanoid < 3.3.18` at `2026-08-13T15:43Z`. That green is stale and will likely flip on re-run through no fault of this PR. Tracked in #1916.

gomesalexandre (changes requested):
Blocking issues on the current head.

Receipts:
- Focused exact-head receipt: `./node_modules/.bin/vitest run clients/cli/src/agent/__tests__/executor.summary.test.ts` passed.
- Static diff trace shows two real holes in the new label-summary path.

Findings:
1. `clients/cli/src/agent/executor.ts:657-664` removes `on <chain>` with `routedChainPattern`, but only `positiveRoutedChainPattern` guards `not on <chain>`. A label like `USDC.e not on Polygon` can therefore render a dangling `not` before the recipient.
2. The two new headline tests at `clients/cli/src/agent/__tests__/executor.summary.test.ts:173-227` use ERC-20 `0xa9059cbb` calldata, so `getPendingSummary()` short-circuits through `renderErc20TransferSummary()` and never exercises `renderLabelSendSummary()` / `stripEmbeddedPayloadContract()` at all.
3. `stripEmbeddedPayloadContract()` only strips the Unicode ellipsis form `0x1234…abcd`, not ASCII `0x1234...abcd`.

This needs another patch plus tests that actually hit the new fallback label-rendering branch.

neavra:
@gomesalexandre updated and rebased on main (head `dde4b622e`). The three findings from your last pass are addressed: the negated routed-chain now strips the whole `not on <chain>` phrase (no dangling `not`), added fallback-path tests that actually exercise the label-summary render (the previous ones short-circuited through the decoded-transfer branch), and the payload-contract strip now handles the ASCII `...` form alongside the unicode ellipsis. Mind taking another look when you get a chance?

gomesalexandre (approved):
Genuine approve, re-reviewed the actual current head after clearing up the bogus four-eyes refusal bug (a commit landing on someone else's branch was wrongly treated as disqualifying my own approve — it isn't, GitHub only blocks self-approve when the reviewer IS the author).

This PR has a long, real review history (mine and CodeRabbit's) — the last CHANGES_REQUESTED from me (2026-08-17, commit 9861fa03) flagged three concrete holes:
1. `not on <chain>` labels could leave a dangling "not" before the recipient
2. the new headline tests only exercised `renderErc20TransferSummary`, never touching `renderLabelSendSummary`/`stripEmbeddedPayloadContract`
3. `stripEmbeddedPayloadContract` only stripped the Unicode-ellipsis truncation form, not ASCII `...`

Verified current head (`dde4b622e`, which includes the follow-up commit `a05142412` "guard negated routed-chain and ASCII ellipsis in send-summary fallback") fixes all three: the regex now consumes `not\s+` together with the chain fragment, `stripEmbeddedPayloadContract` matches `(?:…|\.{3})`, and four new fallback-path tests explicitly spy on `renderLabelSendSummary` vs `renderErc20TransferSummary` to prove the right code path is under test.

Ran the target suite myself on this exact head after a clean `yarn build:sdk && yarn build:client-shared`: `executor.summary.test.ts` — 27/27 passing, including the four new regression cases. CI is green (unit, integration, lint/type, summary, CodeRabbit).

No remaining blocker from my side.

PR #1740, opened 2026-08-05
fix(cli): balance (all-chains) fails per-chain, no stack-trace/fs-path leak (bead mvvry)

NeOMakinG:
## Summary
- `balance` (no chain arg = all chains) now fails per-chain instead of aborting the whole sweep on the first error
- Strips `file://` paths from error messages so no local filesystem structure leaks to stdout
- Fixes bead **vultisig-mvvry** (dogfood 2026-08-05, P2)

## Motivation
Previously vault-cli's `balance` called `vault.balances(undefined, …)` — one call that throws on the first failing chain. When Kujira's REST upstream is down (bead pgyiu, polkachu unreachable), the whole balance sweep aborted with an unhandled stack trace:

```
Failed to fetch balance for Kujira: VaultError: fetch failed
    at BalanceService.getBalance (file:///Users/…/vultisig-sdk/…/index.node.esm.js:1:3570773)
    at process.processTicksA…
```

Three problems in one:
1. One chain-error kills the whole sweep — Cosmos, Ethereum, everything is unfetched
2. Stack trace to stdout breaks the `--silent -o json` contract (no valid JSON)
3. `file://` paths leak user's local filesystem structure

`executePortfolio` in the same file already handles this by iterating chains one-by-one with per-chain try/catch. `executeBalance` now does the same.

Before this change:
```
$ vault-cli balance --vault Testwallet2 --silent -o json
Failed to fetch balance for Kujira: VaultError: Failed to fetch balance for Kujira: TypeError: fetch failed
    at BalanceService.getBalance (file:///Users/0xm4king/Projects/vultisig/vultisig-sdk-worktrees/hl-order-vault-cli/packages/sdk/dist/index.node.esm.js:1:3570773)
    at process.processTicksAndRejections (node:internal/…)
    …
(no JSON output at all)
```

After (expected):
```
$ vault-cli balance --vault Testwallet2 --silent -o json
{
  "balances": { "Ethereum": {...}, "Arbitrum": {...}, ... },
  "failures": [{ "chain": "Kujira", "error": { "message": "Failed to fetch balance for Kujira: TypeError: fetch failed" } }]
}
```

## Implementation
- Loop over `vault.chains`, fetch each independently via `vault.balances([chain], includeTokens)` in `Promise.all`, catch per chain
- Aggregate successes into the existing `balances` map, failures into a separate array
- Strip `file://` paths from error messages before display
- Spinner suffix shows failure count
- JSON envelope omits `failures` key when empty (backward-compatible)

## Tests
`balance.test.ts` has no assertion covering the all-chains branch (only single-chain + tokens flag). No regression. Portfolio's per-chain pattern is what this mirrors.

Fix via Claude Opus 4.7

## Summary by CodeRabbit

* **New Features**
  * Balance checks now continue across chains when one chain fails.
  * JSON results include per-chain failure details when applicable.
  * SDK balance APIs can report individual chain-fetch failures through an optional callback.

* **Bug Fixes**
  * All-chain balance checks now return a retryable network error when every chain fails.
  * Spinner and table output distinguish successful, partial, and complete failures.
  * Warnings no longer expose local file paths or stack-trace details.


gomesalexandre (dismissed):
The bug is real and the repro line in the bead is a genuine leak. I went to verify the fix and ended up somewhere else: I do not think the diagnosis holds, and if it does not, this diff cannot fix the reported symptom.

## preferably-blocking: `vault.balances()` does not throw on a chain failure, so the new catch never fires

The body's premise:

> `vault.balances(undefined, options.includeTokens)` - one call that throws on the first failing chain [...] One chain-error kills the whole sweep.

`BalanceService.getBalances` already fans out per chain with a per-chain catch, and it swallows:

```ts
// packages/sdk/src/vault/services/BalanceService.ts:114-131
await Promise.all(
  chainsList.map(async chain => {
    try {
      const entries = ...
      for (const [key, balance] of entries) { result[key] = balance }
    } catch (error) {
      console.warn(`Failed to fetch balance for ${chain}:`, error)
    }
  })
)
return result
```

Nothing rethrows. `VaultBase.balances` at `:973` is a one-line passthrough to it. So the all-chains call already returns partial results, and the sweep was never killed.

Which means at your head the per-chain `try` in the CLI has nothing to catch. `failures` is always empty, the `(${failures.length} chain(s) failed)` spinner text never renders, the `warn` loop never runs, the `failures` key never appears in the JSON, and the `file://` strip never executes on anything.

And the leak the bead reports is still there afterwards, because it is coming from that `console.warn`. Compare:

```
Failed to fetch balance for Kujira: VaultError: fetch failed
    at BalanceService.getBalance (file:///...)
```

`Failed to fetch balance for ${chain}:` is verbatim the template at `BalanceService.ts:128`, and the second argument is the raw `Error`, which `console.warn` formats with its full stack including `file://` frames. That is the leak, start to finish, inside the SDK.

### Receipts

Drove the real `BalanceService.getBalances` from `packages/sdk/src` with only `getBalance` stubbed, console.warn captured. Nothing else mocked, no CLI in the loop.

```
C1 pass-ctrl both OK    RESOLVED entries=2 warns=0
V1 Kujira fails         RESOLVED entries=1 warns=1  :: Failed to fetch balance for Kujira: fetch failed for Kujira |stack=YES
V2 BOTH fail            RESOLVED entries=0 warns=2  :: Failed to fetch balance for Ethereum: fetch failed for Ethereum |stack=YES
```

C1 is the control and it holds, so the harness is reading live code rather than a stub that always resolves. V1 is the case the PR is about: one chain down, and `getBalances` **RESOLVED**, it did not reject. V2 is the one that settles it: every chain down, still resolved, still no rejection. `stack=YES` on the warn payload is the `file://` frames.

### Why the premise looks true

Worth flagging because it will bite whoever writes the test: if you mock `vault.balances` at the CLI boundary to reject, the merge base throws and your head returns partial results, so an A/B at that level shows the fix working perfectly. I ran exactly that shape first and got a clean green table. It is measuring the mock, not the SDK. The only probe that answers the question is one that drives `getBalances` itself.

### The narrow ask

Fix it where it leaks. `BalanceService.ts:128` is the whole bug:

```ts
} catch (error) {
  console.warn(`Failed to fetch balance for ${chain}:`, error)
}
```

Two options and I do not have a strong preference:

- Log the message only, never the `Error` object: `console.warn(\`Failed to fetch balance for ${chain}: ${(error as Error)?.message ?? String(error)}\`)`. One line, kills the stack and the `file://` frames, keeps the diagnostic.
- Better, and it makes this PR's CLI half real: collect the failures and return them, so `getBalances` hands back `{ balances, failures }` and the caller decides what to print. Then the per-chain shape you wrote in `balance.ts` has something to populate it and the SDK stops writing to the console at all, which it should not be doing in a library.

If you go the second route, most of this diff survives, it just gets its data from the SDK instead of from a catch that cannot fire.

Also worth knowing: the SDK's `console.warn` goes to stderr, so `balance -o json | jq` does parse today. The noise is real and `--silent` does not suppress it (CLI `isSilent` only gates `lib/output.ts:236`, not `console.*`), but if the bead claims broken JSON parsing that part may need a re-check.

## should-fix: the branch that actually can throw is the one left unguarded

`getBalances` swallows. `getBalance` does not:

```ts
// BalanceService.ts:87-97
} catch (error) {
  this.emitError(error as Error)
  throw new VaultError(VaultErrorCode.BalanceFetchFailed, `Failed to fetch balance for ${chain}: ...`, error as Error)
}
```

And `VaultBase.balance` at `:966` passes that straight through to `balance.ts:54`:

```ts
const balance = await vault.balance(options.chain)
```

No try/catch. So `vultisig balance Kujira` on a dead RPC still produces an uncaught `VaultError` out of the CLI, which is the failure mode the PR title describes. The diff hardens the branch that was already safe and leaves the one that is not. Same for `vault.balances([options.chain], true)` at `:43`, though that one inherits the swallow.

## should-fix: the inline sanitizer duplicates and diverges from this file's own helper

`balance.ts:16-19` already has:

```ts
function conciseError(err: unknown): string {
  const message = err instanceof Error ? err.message : String(err)
  return message.split('\n')[0].trim() || 'Unknown error'
}
```

with the docstring literally saying "never a stack trace, never a filesystem path". The new inline version drops `.trim()` and the `'Unknown error'` fallback, and adds a `file://` strip that `conciseError` lacks. Two sanitizers in one file that disagree in both directions is how one of them rots. Fold the `file://` replace into `conciseError` and call it.

Same for the failure shape. Portfolio's `ChainFailure` is `{ chain, stage, error: string }`, the new one is `{ chain, error: { message } }`. Two failure envelopes in the same command's JSON output for the same class of event, so anything consuming both has to branch. Pick one.

## nit

The comment says portfolio "iterat[es] chains one-by-one". It does not, `:159` is `Promise.all(chains.map(...))`, same concurrency as what you wrote.

The body says `balance.test.ts` has no assertion covering the all-chains branch and therefore no regression risk. There are 17 tests in there including `it('collapses a multi-line error MESSAGE to its first line (conciseError, head-on)')` at `:226` and a whole `describe('executeBalance honours --tokens on a single chain')` at `:374`. The all-chains branch specifically may be uncovered but the file is not the blank slate the body implies.

`Record<string, unknown>` for `balances` forces the `as Parameters<typeof displayBalancesTable>[0]` cast at the call site. Portfolio solves this at `:155` with an explicit `ChainResult` type. Copying that removes the cast.

I could be wrong about the SDK path if the bead's repro came from somewhere other than `getBalances`. If you have the raw stack from the original report, paste it and I will re-check. But if it is the line quoted above, this needs to move down a layer.

gomesalexandre:
Took this over to close the loop on the `preferably-blocking` finding above.

Checked the commit that landed after my review (6acbd26c, the `onChainError` callback). It fixes the *symptom* my review's second finding described (CLI now gets real per-chain failure info instead of an always-empty `failures` array), but it left the actual leak untouched: `BalanceService.getBalances`' own `console.warn(\`Failed to fetch balance for ${chain}:\`, error)` (BalanceService.ts:140) still passes the raw `Error` object as the second arg. `console.warn`/`util.format` prints an `Error` with its full stack, so every swallowed per-chain failure was still writing `file://` frames straight to the terminal - the exact bead repro, unchanged by that commit.

Pushed `2a61e7c6`:
- `BalanceService.ts:140` now logs `error.message` only, same convention already used by `ServerManager`/`RelaySigningService`'s MLDSA warn logs elsewhere in this SDK (`error instanceof Error ? error.message : error`).
- Added a repro test (`BalanceService.test.ts`) that throws an `Error` carrying a real `file://` stack and checks the console.warn output through `node:util`'s `format` (the actual formatter `console.warn` uses) - a naive `.join(' ')` would call `Error.prototype.toString()` and silently miss the stack, which is exactly the gap that would've let this ship green. Mutation-verified: reverting the source line makes the test fail with the leaked `file://` frame visible in the assertion diff; restoring it is clean.
- Left the CLI's per-chain try/catch in `balance.ts` as-is (dead-code question from my review). Re-checked the call chain: `VaultBase.balances()` passes straight through to the fan-out with no validation that could throw first, so the catch is still unreachable today - same finding as before. Keeping it as documented defensive redundancy rather than ripping it out, since it's cheap and already commented as such; happy to remove it instead if you'd rather not carry it.

Ran `yarn workspace @vultisig/sdk test` (3211 tests, all green, including the new one), `yarn test:cli` (1014 tests, green), `yarn workspace @vultisig/sdk typecheck`, `yarn workspace @vultisig/cli typecheck`, and lint on the touched files - all clean.

Not approving/dismissing this review - leaving that for a second pair of eyes.

gomesalexandre (changes requested):
Blocking: `clients/cli/src/commands/balance.ts:114-117` still reports total failure as success.

At the current head, every per-chain failure is accumulated into `failures`, but the command always does:
- `spinner.succeed(...)`
- `outputJson(failures.length === 0 ? { balances } : { balances, failures })`

So if **all** chains fail, the CLI still exits as a success with `{ balances: {}, failures: [...] }` instead of surfacing a retryable command error. `executePortfolio` already draws the right line at total failure: when `failures.length > 0 && chainBalances.length === 0`, it fails the spinner and throws a `NetworkError`.

Please make `executeBalance` follow the same contract:
- keep the current partial-success behavior when at least one chain succeeds
- but when zero chains succeed, fail the command instead of printing a success result

Also add a CLI regression that covers both:
1. mixed success + failure stays partial-success
2. all chains failing becomes a command error

Four-eyes note: current-head approval is forbidden here because the current head includes substantive `gomesalexandre` commits.

gomesalexandre (changes requested):
Current head fixes the two core behaviors, but I am still requesting changes on cleanup / completeness.

Receipts:
- Focused exact-head receipt: `./node_modules/.bin/vitest run clients/cli/src/commands/balance.test.ts packages/sdk/tests/unit/services/BalanceService.test.ts`
- `clients/cli/src/commands/balance.test.ts` passed with the new partial-vs-total failure coverage.
- The SDK suite was blocked in this disposable checkout by a pre-existing workspace import failure: `@vultisig/core-chain/chains/ripple/issuedCurrency` could not resolve from `packages/sdk/tests/unit/services/BalanceService.test.ts`.

Findings:
1. `clients/cli/src/commands/balance.ts` now carries two contradictory inline explanations about whether `vault.balances()` throws. Please delete the stale block so future readers are not misled about the actual failure contract.
2. The CLI still loops `vault.balances([chain], ...)` per chain instead of using the new batch API with the per-chain callback directly. The new SDK callback exists specifically to avoid reimplementing the fan-out in callers.
3. The JSON/CLI failure-contract docs are still portfolio-only. This PR gives all-chains `balance` the same `failures` surface and should document it.

rcoderdev (approved):

rcoderdev:
Closing this because this conflicted head still has requested changes: it duplicates SDK fan-out in the CLI, carries contradictory failure-contract comments, and leaves the new JSON failure surface undocumented; recreate the behavior cleanly from current main.

PR #1742, opened 2026-08-05
fix(cli): tx-status --no-wait maps pending+isKnown:false to not_found (bead 3xitk)

NeOMakinG:
## Summary
- Fixes `tx-status --no-wait` on UTXO chains returning `pending` for hashes the node has never seen
- Bead **vultisig-3xitk** (dogfood 2026-08-05, P3)

## Motivation
`vault-cli tx-status --chain bitcoin --tx-hash <unknown-hash> --no-wait` returned `status: 'pending'`, while Ethereum correctly returned `not_found` for the same case.

Root cause: chains whose resolver can't distinguish "never seen" from "transient RPC error" fall back to `{status:'pending', isKnown:false}`. UTXO's Blockchair adapter (`packages/core/chain/tx/status/resolvers/utxo.ts:36`) does this **deliberately** to protect `verifyBroadcastByHash` from swallowing broadcast failures as success — the test at `utxo.test.ts:21-36` explicitly pins that shape and explains why.

Fix location matters: SDK contract must stay `{status:'pending', isKnown:false}` for the broadcast-verify safety net. Only the CLI --no-wait path is patched — it has made its single best attempt and the user asked NOT to poll, so surfacing `not_found` is the terminal signal the CLI needs.

Before:
```
$ vault-cli tx-status --chain bitcoin --tx-hash 0000...0000 --no-wait --vault X -o json
{ chain: 'Bitcoin', status: 'pending' }   ← bug

$ vault-cli tx-status --chain ethereum --tx-hash 0x1111...1111 --no-wait --vault X -o json
{ chain: 'Ethereum', status: 'not_found' }
```

After:
```
$ vault-cli tx-status --chain bitcoin --tx-hash 0000...0000 --no-wait --vault X -o json
{ chain: 'Bitcoin', status: 'not_found', isKnown: false }   ← now terminal
```

## Behaviour scope
| Path | isKnown | Before | After |
|------|---------|--------|-------|
| --no-wait + pending | `false` | `pending` | **`not_found`** |
| --no-wait + pending | `true`  | `pending` | `pending` (unchanged) |
| --wait mode           | any     | polls-then-timeout | unchanged |
| SDK `getTxStatus`     | any     | unchanged | unchanged (broadcast-verify safe) |

## Tests
Two new tests in `txStatus.test.ts`:
1. `maps pending+isKnown:false to not_found in --no-wait mode` — pins the new mapping + verifies `isKnown:false` is preserved on the returned object
2. `does NOT map pending+isKnown:true to not_found` — real in-mempool tx must stay pending

Existing tests unchanged (asserted `{status:'not_found', isKnown:false}` cases still pass).

Fix via Claude Opus 4.7

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction status reporting for UTXO-based transactions.
  * Transactions missing from a successful provider response are now correctly reported as **not found**.
  * Provider or network failures continue to report transactions as **pending**.
  * Preserved accurate known-transaction information in no-wait status checks.


gomesalexandre (dismissed):
preferably-blocking: CI is red (`Unit Tests` fail, `Test Summary` fail), and the cause is that **both new tests reject before they reach the code they're testing**. The fix is one character shy of trivial, and the change itself is sound - see below, I think the reasoning in this one is genuinely good.

## the test bug

Both new cases pass `EVM_HASH` with `Chain.Bitcoin`:
```ts
const EVM_HASH = '0x' + 'a'.repeat(64)
...
await executeTxStatus(ctx, { chain: Chain.Bitcoin, txHash: EVM_HASH, noWait: true })
```

`packages/core/chain/tx/isValidTxHash.ts`:
```ts
const HEX_64 = /^[0-9a-fA-F]{64}$/
const HEX_64_PREFIXED = /^0x[0-9a-fA-F]{64}$/
...
evm:  h => HEX_64_PREFIXED.test(h),
utxo: h => HEX_64.test(h),          // <- Bitcoin: NO 0x prefix
```

So the hash is malformed *for Bitcoin*, and `tx-status` validates shape before any RPC (as its own header comment advertises: "validated for its chain-kind BEFORE any RPC call"). The flow throws `InvalidTxHashError` and never reaches your `noWait && pending && isKnown === false` branch at `:124`. Locally: `2 failed | 24 passed`, failing on `'Check the hash — it must match the expected format for the chain'`.

Fix: `const BTC_HASH = 'a'.repeat(64)` and use it for the `Chain.Bitcoin` cases. Worth keeping `EVM_HASH` around for the Ethereum tests that already use it.

Once that's in, the two cases become genuinely valuable - especially the negative one (`does NOT map pending+isKnown:true to not_found`), which is the assertion that stops this from swallowing a real in-mempool tx.

## the change itself is right, and I want to say why

I went in expecting to push back. Mapping "I can't tell" to `not_found` is normally how you get a user to conclude a broadcast failed and re-sign it - the double-send shape. The UTXO resolver is explicit that its fallback conflates two cases:

> Transient RPC/network failure OR Blockchair affirmatively has no record of this hash at all — either way we can't confirm the tx is genuinely known.

But you scoped it correctly, and the comment shows you'd already thought it through:

- gated on `params.noWait` only, so `--wait` still polls and can observe `isKnown` flip true
- **only `status` is rewritten; `isKnown: false` is preserved on the result**, so `verifyBroadcastByHash` and the SDK contract still see the truth. That's the part that makes this safe rather than clever
- `recordResolution` is only called for `success`/`error`, so a mapped `not_found` never poisons the broadcast journal
- no `TxNotFoundError` is thrown on this path - it returns and displays, so a `--no-wait` reader isn't handed the `'Re-broadcast if it was never sent'` guidance that `giveUpError` gives on the timeout path

That last one matters most and I checked it specifically. If `--no-wait` + mapped `not_found` had routed through `giveUpError`, this would be a block on fund-safety grounds rather than on CI grounds.

**q (non-blocking):** `not_found` is a definitive-sounding word for a signal the resolver defines as "can't confirm". Since `isKnown` is already on the result, is it worth having `displayResult` say something like `not_found (unconfirmed — node may not have indexed it yet)` when `isKnown === false`? The status vocabulary stays the same for scripts, the human just isn't told a stronger story than we have. Entirely your call.

Re-run CI after the hash fix and ping me - I'll stamp it.

gomesalexandre (commented):
current head looks materially better than the original cli-only patch.

what changed, and why i am not stamping:
- `packages/core/chain/tx/status/resolvers/utxo.ts` now separates a successful empty Blockchair answer from a transport/provider failure
- `!response` or request error still returns `{ status: 'pending', isKnown: false }`, which keeps the broadcast-verify safety net fail-closed under provider uncertainty
- `response.data[hash]` missing now returns `{ status: 'not_found', isKnown: false }`, which gives the cli a real terminal signal for a hash the provider affirmatively does not know
- the tests now cover both sides of that split

that shape is conceptually sound to me on the current head.

i am leaving this as COMMENT, not APPROVE, because the current head contains substantive `gomesalexandre` commits (`test(cli): use a UTXO-shaped hash...` and `fix(utxo): separate missing tx from provider failure`). per four-eyes i cannot be the approval stamp on code i authored.

please get a non-gomes review on the current head before merge.

gomesalexandre (approved):
Genuine approve. The only thing standing between this PR and a stamp was the four-eyes bug (a gomes-authored commit on the branch was wrongly treated as self-approve-disqualifying — GitHub only blocks that when the reviewer IS the PR author, which I'm not here).

Re-verified the actual current head (`0e27a2d20`) myself:
- `packages/core/chain/tx/status/resolvers/utxo.ts` now distinguishes a transport/provider failure (`error || !response` → stays `{status:'pending', isKnown:false}`) from a genuine miss (`response.data[hash]` absent → `{status:'not_found', isKnown:false}`).
- Checked `verifyBroadcastByHash.ts` directly (the broadcast-verify safety net this PR explicitly protects): it only swallows a broadcast error on `status === 'success'` or (`status === 'pending' && isKnown !== false`). Since `isKnown` stays `false` on both the old and new empty-response shape, this change is a no-op for the safety-net gate — only the surfaced status label changed, not the swallow condition. That confirms the "SDK contract stays safe" claim rather than just trusting it.
- CLI `--no-wait` path: `tx-status.ts` tests now use a real UTXO-shaped hash (`isValidTxHash` gates BTC on the no-`0x` form), fixing the earlier CI-red test bug. Both directions are pinned: `pending+isKnown:false` → `not_found`, `pending+isKnown:true` → stays `pending` (real in-mempool tx not falsely terminalized).

CI green across the board. Diff is small (3 files) and scoped exactly to what the PR claims. No fund-safety regression found.

rcoderdev (approved):

rcoderdev:
Closing this because the not-found distinction is useful, but this old branch is behind current transaction-status code; recreate the focused empty-provider-result fix from current main so broadcast-verification behavior can be revalidated on a current head.

PR #1743, opened 2026-08-05
fix(cli): swap dry-run warns on --slippage 0 (bead ehedh)

NeOMakinG:
## Summary
- `swap --slippage 0` now surfaces a warning in the dry-run instead of building silently
- Bead **vultisig-ehedh** (dogfood 2026-08-05, P3)

## Motivation
`swap --slippage 0` built cleanly even though a 0% tolerance GUARANTEES revert on any price movement between the quote timestamp and on-chain execution. Bridges take ~15s+; price always moves within that window, so the signed tx is essentially certain to revert.

Existing range guards for negative (-5) and >50 (100) already reject at the SDK layer with the correct message. The exact-0 case slipped through as a "successful" build.

Before:
```
$ vault-cli swap arbitrum base 0.0001 --slippage 0 --dry-run --vault X -o json
{ success: true, dryRun: true, provider: 'li.fi', estimatedOutput: '0.000099...', ... }
(no warning; user proceeds to --confirm; tx reverts on-chain)
```

After (expected):
```
$ vault-cli swap arbitrum base 0.0001 --slippage 0 --dry-run --vault X -o json
{ success: true, dryRun: true, ..., warnings: [
  "Slippage=0% guarantees revert on any price movement between quote and execution — consider >= 0.1%"
] }
```

Non-JSON output also shows `Warning: Slippage=0%...` via the existing displayDryRunResult path (line 193 already iterates result.warnings).

## Implementation
Small addition in `toDryRunResult` (`clients/cli/src/commands/swap.ts:170-181`): append a warning to `result.warnings` when `options.slippage === 0`. displayDryRunResult already surfaces warnings, so no display change needed. Fires ONLY on exactly `0` — `undefined` and any `>0` value are untouched.

## Tests
Two new tests in `swapAmount.test.ts`:
1. `warns when --slippage 0 is passed` — pins the warning fires and contains the expected slippage/revert keywords
2. `does NOT add the slippage-0 warning when slippage is undefined or >0` — regression guard against over-firing

## Related
Same family as tbu0p / 5zxnx (preview must surface preconditions/risks user should audit before --confirm).

Fix via Claude Opus 4.7

## Summary by CodeRabbit

* **Bug Fixes**
  * Dry-run swap results now warn when slippage is explicitly set to 0, helping prevent likely transaction reversions if prices change.
  * Existing warnings remain intact alongside the new warning.


gomesalexandre (dismissed):
Right instinct and right posture - **warn, don't block**. A 0% tolerance is a legitimate thing for a user to ask for even though it's almost always a mistake, and the SDK already hard-rejects the genuinely invalid range (negative, >50). Warning on the merely-doomed case is the correct severity. Approving.

## verified

**must-pass:** `src/commands/swapAmount.test.ts` **19/19**.

**must-fail:** neutered the condition (`if (options.slippage === 0)` -> `if (false)`, pre-count 1 / post-marker 1): **1 fails** (`warns when --slippage 0 is passed (bead ehedh)`).

Good that you pinned the negative direction too (`does NOT add the slippage-0 warning when slippage is undefined or >0`) - `=== 0` vs a truthiness check is exactly where this would have gone wrong, since `undefined` and `0` are both falsy and only one of them deserves a warning.

## should-fix: the warning never reaches the user who signs

`toDryRunResult` has exactly one caller, `swap.ts:251`, and it's inside:
```ts
if (options.dryRun) {
  const result = toDryRunResult(options, quote, fromAmountRaw)
  ...
  return result
}
```
The real swap continues past that block to "2. Show preview" at 259+, builds its own preview, and calls `confirmSwapIfNeeded(options)`. That path has no `warn(` and no reference to `slippage` at all. So:

```
vault-cli swap ... --slippage 0 --dry-run   -> warned
vault-cli swap ... --slippage 0             -> silent, previewed, confirmed, signed
```

The second one is the user who actually eats the revert. Right now the warning fires precisely for the person who was already being careful enough to dry-run first, and stays quiet for the one who wasn't.

I'm not blocking on it, because this is strictly better than the status quo (which warned nowhere) and the title is honest that it's scoped to dry-run. But the bead's own framing - "built cleanly even though a 0% tolerance GUARANTEES revert" - describes the signing path, not the preview path, so I'd call this half-fixed. The natural home is next to `confirmSwapIfNeeded`, where the user is already being asked to make a decision: a `warn()` there costs one line and reaches both audiences, and then the dry-run warning becomes the redundant one rather than the only one.

Worth a follow-up bead if you'd rather not widen this PR.

**nit:** the message hardcodes `>= 0.1%` as the suggestion. Fine as guidance, but if the SDK ever gains a per-route minimum (bridges plausibly want more than 0.1%), this string is the thing that goes stale silently. Not worth changing now, just don't let it become the de-facto spec.

`Fixes` bead ehedh.

gomesalexandre (dismissed):
Two still-valid blockers at the current head.

1. `clients/cli/src/commands/swap.ts:179` still overstates the behavior. With `slippage=0`, adverse movement can revert, but favorable movement can still succeed. The warning should be conditional, not `guarantees revert on any price movement`.

2. The warning only exists in the dry-run return path. The signing path at `clients/cli/src/commands/swap.ts:249-275` previews and confirms without ever surfacing the same warning, so `swap --slippage 0` can still reach the user-facing sign flow silently unless they happened to dry-run first. That is only a half-fix for bead ehedh. Please surface the warning before confirmation too and cover that path.

Meta gate: this head is still `REVIEW_REQUIRED`, behind `main`, and the earlier human approval was dismissed. Also, current head `397c1181` includes substantive gomes-authored code (`70508e3`), so this branch will need a non-gomes stamp once fixed.

Receipts:
- current-head warning string at `swap.ts:179` still says `guarantees revert on any price movement`
- current-head signing path at `swap.ts:249-275` only calls `toDryRunResult(...)` inside the `if (options.dryRun)` branch, then goes straight to preview + `confirmSwapIfNeeded(options)` with no matching warning on the non-dry-run path

🤖 reviewed via codex remote pass + host-shell verification

gomesalexandre (commented):
Current-head remote review on `eda237e7` only.

What I verified:
- the zero-slippage text is now conditional, not overstated. `slippageZeroWarning()` says `may revert if the price moves against you before execution`, which matches the real behavior instead of the earlier `guarantees revert on any price movement` claim.
- the warning is no longer dry-run-only. The same helper is now consumed in both places: `toDryRunResult(...)` appends it to the JSON/return warnings array, and the non-dry-run signing path calls `warn(...)` before confirmation, so `swap --slippage 0` reaches the real preview/sign flow with the same warning even when the user never dry-runs first.
- the PR added an explicit signing-path regression test (`clients/cli/src/commands/__tests__/swapSlippageWarningSigningPath.test.ts`) covering the exact gap from the earlier review, alongside the dry-run coverage in `swapAmount.test.ts`.
- current CI on this head is green: Unit Tests, Integration Tests (Vitest), Lint and Type Check, Test Summary, and CodeRabbit all passed on `eda237e7`.

I do not see a remaining blocker on the current diff for bead `ehedh`.

Important four-eyes note: this head includes substantive `gomesalexandre` commits (`70508e3`, `eda237e7`), so I cannot approve it. It needs a non-gomes current-head approval before merge.

gomesalexandre (approved):
Genuine approve. Only blocker was the four-eyes bug (gomes-authored commit on the branch wrongly treated as self-approve-disqualifying — that rule only fires when the reviewer IS the PR author).

Re-checked the current head diff myself: `slippageZeroWarning()` fires only on exactly `options.slippage === 0` (verified via the negative test alongside it — `undefined`/`>0` don't fire). Wording is now conditional ("may revert if the price moves against you") rather than the earlier overstated "guarantees revert," which was a real correctness catch by a prior reviewer. Confirmed the warning is wired into BOTH paths that matter: the dry-run return (`toDryRunResult`) and the actual signing path (`executeSwap`, before `confirmSwapIfNeeded`) via the new `swapSlippageWarningSigningPath.test.ts`, which was the earlier "half-fixed" gap (warning only reached dry-run users, not the person who actually eats the revert). That gap is closed now.

Small, scoped diff (3 files), CI green, warn-don't-block is the right severity for a legitimate-but-risky user choice.

rcoderdev (approved):

rcoderdev:
Closing this because the explicit zero-slippage warning is useful, but this old branch is behind current swap command behavior; recreate the small warning and focused coverage from current main instead of carrying the stale head forward.

PR #1744, opened 2026-08-05
fix(cli): send --max shows resolved amount, not literal 'max' (bead 2lnf8)

NeOMakinG:
## Summary
- `send --max` (and positional `send <chain> <addr> max`) now shows the resolved amount instead of the literal `'max'` string
- Bead **vultisig-2lnf8** (dogfood 2026-08-05, P4 display-only)

## Motivation
Repro (Testwallet2 2026-08-05):
```
$ vault-cli send ethereum <addr> --max --dry-run -o json
{
  "amount": "max",                    ← leaks the literal input string
  "symbol": "ETH",
  "fee": "0.000025698195113",
  "total": "0.003973219149949019",
  "balance": "0.003973219149949019"
}
```

The **signed value is correct** — `VaultBase.ts:1750-1754` resolves `'max'` → `maxSendable` (balance - fee) internally. `total` at line 1782 = `maxSendable + fee = balance` for native, or `maxSendable` for token. The bug is display-only: sign-card readers see `'max'` instead of the actual number they're authorizing.

Verify-first triage earlier this session **downgraded this from P2 fund-safety to P4 display-only** after confirming the signed value is correct. This PR closes that display-only tail.

## After
```
$ vault-cli send ethereum <addr> --max --dry-run -o json
{
  "amount": "0.003947520954836019",   ← resolved (balance - fee)
  "symbol": "ETH",
  "fee": "0.000025698195113",
  "total": "0.003973219149949019",
  "balance": "0.003973219149949019"
}
```

Non-max amounts pass through unchanged so user-typed decimals stay verbatim.

## Implementation
Rehydrate the display from what `dryResult` already returned:
- **Token max**: `total` IS the amount (`VaultBase.ts:1782` isTokenSend branch, since fee is paid in a different token) → pass through
- **Native max**: `amount = total - fee` via BigInt-scaled string arithmetic (no float precision loss at 18-decimal chains — verified against the actual repro numbers)

Uses no new SDK fields — keeps the existing `amount` key.

## Contrast
`swap-quote` already exposes the resolved value via a distinct `maxSwapable` field. This PR aligns send's shape with the "resolved value exposed" pattern **without** introducing a new field (uses the existing `amount` key that scripting layers already parse).

## Tests
New `transaction.test.ts` with 5 test cases pinning:
- Token vs native branches
- Decimal widths 6 / 8 / 18 (cosmos / BTC / EVM)
- The exact ETH repro from the bead (`0.003973219149949019 - 0.000025698195113 = 0.003947520954836019`)
- Edge case: fee > total returns `'0'` (belt-and-suspenders defense; SDK throws earlier via `VaultBase.ts:1752` `'Insufficient balance to cover network fees'`)
- Trailing-zero trimming

No existing tests break (helper is standalone, called only when `params.amount === 'max'`).

Fix via Claude Opus 4.7

## Summary by CodeRabbit

* **New Features**
  * Improved `--max` transaction displays in dry-run mode.
  * Native asset sends now show the maximum amount after transaction fees are deducted.
  * Token sends display the full available total.
* **Bug Fixes**
  * Prevented fee calculations from producing negative amounts.
  * Improved decimal precision and formatting, including removal of unnecessary trailing zeros.

### DIFF
--- a/clients/cli/src/commands/transaction.test.ts
+++ b/clients/cli/src/commands/transaction.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from 'vitest'
+
+import { resolvedMaxAmount } from './transaction'
+
+// bead vultisig-2lnf8: send --max previously echoed the literal 'max' string
+// in the dry-run response's `amount` field. The helper below rehydrates the
+// resolved value from what the SDK already returned (dryResult.total and .fee),
+// which is what previewDryRun now passes to `amount`. These cases pin the
+// arithmetic across the two SDK shapes (native vs token) and across decimal
+// widths from 6 (cosmos) to 18 (EVM).
+describe('resolvedMaxAmount', () => {
+  it('token send: passes total through unchanged (fee is in a different token)', () => {
+    // For a token max, VaultBase.ts:1782 isTokenSend branch sets total = amountBigInt
+    // (i.e. total IS the amount; fee is paid in the native gas token separately).
+    expect(resolvedMaxAmount(true, '1.234567', '0.000021')).toBe('1.234567')
+    expect(resolvedMaxAmount(true, '1000', '5.5')).toBe('1000')
+    expect(resolvedMaxAmount(true, '0.5', '0.00001')).toBe('0.5')
+  })
+
+  it('native send: subtracts fee from total via BigInt-scaled arithmetic', () => {
+    // Native max: SDK returns total = amountBigInt + fee, so amount = total - fee.
+    // EVM-scale (18 decimals). Confirmed against the actual repro:
+    // send ethereum <addr> --max --dry-run → total='0.003973219149949019', fee='0.000025698195113'
+    expect(resolvedMaxAmount(false, '0.003973219149949019', '0.000025698195113')).toBe('0.003947520954836019')
+    // BTC-scale (8 decimals)
+    expect(resolvedMaxAmount(false, '0.001', '0.0001')).toBe('0.0009')
+    // ATOM-scale (6 decimals)
+    expect(resolvedMaxAmount(false, '1.234567', '0.0075')).toBe('1.227067')
+  })
+
+  it('native send: handles when fee > total (edge case, returns 0 instead of negative)', () => {
+    // Should not happen in practice — SDK throws 'Insufficient balance to cover
+    // network fees' when maxSendable <= 0 (VaultBase.ts:1752). Belt-and-suspenders
+    // defense that the display helper never emits a negative amount to the user.
+    expect(resolvedMaxAmount(false, '0.001', '0.002')).toBe('0')
+    expect(resolvedMaxAmount(false, '0.0001', '0.0001')).toBe('0')
+  })
+
+  it('native send: integer-only inputs (no decimal point)', () => {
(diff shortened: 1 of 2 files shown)


gomesalexandre (approved):
Tagged P4 display-only, and the *funds* part of that is true - I confirmed the signed value was always correct. But "display-only" undersells it slightly on a pre-signing surface, so I read the arithmetic properly rather than skimming it.

## verified

**must-pass:** `clients/cli/src/commands/transaction.test.ts` **5/5**.

**must-fail:** forced the token branch to always return `total` (`if (isTokenSend)` -> `if (true)`, pre-count 1 / post-marker 1), killing the native subtraction: **4 fail** (`native send: subtracts fee from total via BigInt-scaled arithmetic`, the `fee > total` edge case, integer-only inputs, and trailing-zero trimming).

The arithmetic itself is right. Scale taken from the widest fractional part, both sides padded to that scale, subtraction done in `BigInt`, formatted back - no float anywhere, which is what lets one helper serve 6-decimal cosmos and 18-decimal EVM. `diff <= 0n -> '0'` is the correct fallback rather than rendering a negative. And pinning the real repro numbers (`0.003973219149949019 - 0.000025698195113`) rather than round test values is the right instinct - that's the case that would have caught a float.

## should-fix: this derives the number instead of reading it

`resolvedMaxAmount` reconstructs the amount from `dryResult.total` and `dryResult.fee`. That only works because of an *implicit* contract in `VaultBase.ts:1789`:
```ts
total: this.formatUnits(isTokenSend ? amountBigInt : amountBigInt + fee, ...)
```
i.e. token -> `total === amount`, native -> `total === amount + fee`. Your comment says as much, so you clearly know. The issue is that the CLI now depends on that relationship without anything enforcing it. The SDK's dry-run result never exposes `amountBigInt`, so there's no cross-check: if someone changes what `total` means (say, to include an L1 data fee, or to report gross for tokens), this helper keeps returning a confident, wrong number - and it returns it on the screen a user reads *before approving a transfer*. Silent and pre-signing is a bad combination, even when the signed bytes stay correct.

Your own comment points at the fix: swap-quote solved this by exposing `maxSwapable` as a distinct field. Deriving in the CLI to avoid adding a field is the smaller change today, but it's the one that can drift. Suggest surfacing the resolved amount from `prepareSendTx` (it already has `amountBigInt` in hand at 1761) and letting the CLI print it verbatim. Then `resolvedMaxAmount` becomes a formatting concern rather than an inference, and the four tests above become unnecessary rather than load-bearing.

Not blocking - the derivation is correct against today's contract and I verified it against the real repro figures. Filing it as should-fix because the failure mode is "the preview and the signature disagree", and that's worth a field.

**nit:** `scaleBI` does `BigInt((whole || '0') + padded)` - fine for the unsigned amounts that reach it, but it would silently mis-scale a negative input (`'-1'` + `'000'` -> `-1000` rather than `-1.000` at scale 3). Can't happen from `total`/`fee` today. Worth a one-line note that inputs are assumed non-negative, since this helper is exported and someone will reuse it.

`Fixes` bead 2lnf8.

rcoderdev (approved):

rcoderdev:
Closing this because current main already reports the resolved max amount and preserves isMax and contract metadata, while this old red head regresses that newer path; the implementation is superseded.

PR #1745, opened 2026-08-05
fix(cli): chains --add / resolve emit 'Did you mean' suggestions (bead 7eymb)

NeOMakinG:
## Summary
- `chains --add` and `resolveChainOrThrow` now emit "Did you mean: CronosChain?" style hints when the input is close to a known chain
- Bead **vultisig-7eymb** (dogfood 2026-08-05, P4)

## Motivation
Chain-name resolution is strict-match with no aliasing — common brand names like `Cronos` (the actual product brand) get rejected because only the SDK enum `CronosChain` matches. Error hint just says "check the spelling" — the user has to run `vultisig chains` and grep.

```
$ vault-cli chains --add Cronos --vault X -o json
{
  "code": "INVALID_CHAIN",
  "message": "Unsupported chain: \"Cronos\"",
  "hint": "Run \"vultisig chains\" to see the supported chains, or check the spelling."
}
```

Compare with `balance --token` which has a great "Did you mean --tokens?" hint. This PR gives chain resolution the same UX.

## After
```
$ vault-cli chains --add Cronos --vault X -o json
{
  "code": "INVALID_CHAIN",
  "message": "Unsupported chain: \"Cronos\"",
  "hint": "Did you mean: CronosChain? Or run \"vultisig chains\" to see the supported chains.",
  "context": { "chain": "Cronos", "suggestions": "CronosChain" }
}
```

Common cases covered: `Cronos` → `CronosChain`, prefix matches, case-insensitive.

## Implementation
- New `suggestChainNames(input, limit=3)` helper in `chain-resolver.ts` scores Chain enum entries by prefix + reverse-prefix + substring, returns top-N
- `resolveChainOrThrow` calls it on miss, appends "Did you mean" to the hint, adds `suggestions` to context (comma-joined string; VsigError context values must be strings)
- `chains --add` uses the same helper for its dedicated `InvalidChainError` path

Exact-match resolution is unchanged (fast path). Non-close inputs (`zzzzzzzz`) get the original generic hint.

## Tests
`chain-resolver.test.ts` gets:
- 2 new `resolveChainOrThrow` cases: Cronos → CronosChain suggestion; generic-hint fallback for zzzzzzzz
- 4 new `suggestChainNames` cases: prefix match, case-insensitivity, empty-input, cap-limit

5 pre-existing assertions unchanged.

Fix via Claude Opus 4.7

## Summary by CodeRabbit

* **New Features**
  * Added helpful chain-name suggestions when an invalid chain is entered.
  * Suggestions are case-insensitive, recognize brand-name variations, and display up to three likely matches.
  * Suggested matches are included in both readable error messages and structured error details.

* **Bug Fixes**
  * Improved invalid-chain guidance while retaining a generic message when no close match is found.

### DIFF
--- a/clients/cli/src/commands/chains.ts
+++ b/clients/cli/src/commands/chains.ts
@@ -7,6 +7,7 @@ import chalk from 'chalk'
 
 import type { CommandContext } from '../core'
 import { InvalidChainError } from '../core'
+import { suggestChainNames } from '../core/chain-resolver'
 import { createSpinner, info, isJsonOutput, outputJson, printResult, success } from '../lib/output'
 import { displayAddresses } from '../ui'
 
@@ -45,11 +46,19 @@ export async function executeChains(ctx: CommandContext, options: ChainsOptions 
     // then throw on every subsequent address derivation. Fail closed: nothing is
     // persisted for a chain that isn't supported.
     if (!SUPPORTED_CHAINS.includes(options.add)) {
+      // bead vultisig-7eymb: add did-you-mean hint so `chains --add Cronos` points
+      // to the SDK-enum name `CronosChain` instead of forcing the user to `vultisig
+      // chains` and grep.
+      const suggestions = suggestChainNames(String(options.add))
+      const hint =
+        suggestions.length > 0
+          ? `Did you mean: ${suggestions.join(', ')}? Or run "vultisig chains" to see the supported chains.`
+          : 'Run "vultisig chains" to see the supported chains, or check the spelling.'
       throw new InvalidChainError(
         `Unsupported chain: "${options.add}"`,
-        'Run "vultisig chains" to see the supported chains, or check the spelling.',
+        hint,
         undefined,
-        { chain: String(options.add) }
+        { chain: String(options.add), ...(suggestions.length > 0 ? { suggestions: suggestions.join(',') } : {}) }
       )
     }
     const alreadyActive = vault.chains.includes(options.add)
--- a/clients/cli/src/commands/chains.ts
+++ b/clients/cli/src/commands/chains.ts
@@ -54,12 +54,10 @@ export async function executeChains(ctx: CommandContext, options: ChainsOptions 
         suggestions.length > 0
           ? `Did you mean: ${suggestions.join(', ')}? Or run "vultisig chains" to see the supported chains.`
           : 'Run "vultisig chains" to see the supported chains, or check the spelling.'
-      throw new InvalidChainError(
-        `Unsupported chain: "${options.add}"`,
-        hint,
-        undefined,
-        { chain: String(options.add), ...(suggestions.length > 0 ? { suggestions: suggestions.join(',') } : {}) }
-      )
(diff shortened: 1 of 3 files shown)


gomesalexandre (dismissed):
Approving. Good UX fix, and it stays on the safe side of the line that matters for chain resolution.

## the thing I checked first

A "did you mean" feature next to chain selection is one auto-apply away from being a fund-safety bug - a silently corrected chain is funds on the wrong network. This never resolves anything:

```
chain-resolver.ts:59   const suggestions = suggestChainNames(input)   // building the error hint
chains.ts:52           const suggestions = suggestChainNames(...)     // building the error hint
```

Two call sites, both constructing text on a throw path. The user still has to type the real name. That's the correct design and worth stating explicitly so nobody "improves" it into an auto-correct later.

## verified

**must-pass:** `src/core/chain-resolver.test.ts` **11/11**.

**must-fail:** early-returned `[]` from `suggestChainNames` before the sort (anchor unique, marker confirmed) -> **3 fail** (`suggests CronosChain when user types the Cronos brand`, the prefix-score case, and `is case-insensitive`).

I also ran the scoring logic standalone against a representative chain list to see what it actually does:
```
  Cronos     -> ["CronosChain"]
  eth        -> ["Ethereum"]
  sol        -> ["Solana"]
  arb        -> ["Arbitrum"]
  doge       -> ["Dogecoin"]
  bit        -> ["Bitcoin","Bittensor","BitcoinCash"]
  Etherium   -> []
  Etherem    -> []
  mya        -> []
```

## nit: it solves brand names, not typos

The matcher is prefix/substring only - no edit distance - so it nails the case the bead reports (`Cronos` -> `CronosChain`, and the `eth`/`sol`/`arb` shorthands are a nice free win) but returns **nothing** for an actual misspelling. `Etherium` and `Etherem` both get no hint, and those are the inputs where a user is most likely to be staring at the error wondering what's wrong. Someone who sees "Did you mean" work for `Cronos` will reasonably expect it to work for `Etherium`.

Not blocking - a hint that's absent is no worse than today, and a cheap Levenshtein pass over ~30 enum values could be added later without touching this shape. Just flagging that the feature will read as inconsistent until then.

**observation, not a request:** `bit` returns `Bitcoin, Bittensor, BitcoinCash` - three chains where picking the wrong one loses funds. Fine here because it's a hint list and the full names are shown, so the disambiguation is the user's. I'd only worry if this list ever became numbered/selectable input.

`chains --add` getting the same treatment as `resolveChainOrThrow` is the right call - inconsistent error quality between two paths that reject the same input is its own small papercut.

gomesalexandre (commented):
tests cover suggestChainNames/resolveChainOrThrow, but the user-facing change lands in clients/cli/src/commands/chains.ts. Please add a command-boundary regression on executeChains or the chains CLI path that proves an invalid add like Cronos surfaces the Did you mean hint and suggestions context there too. As written, executeChains could stop forwarding the helper output and these new tests would still stay green.

rcoderdev:
Closing this because the suggestion helper is useful, but this head is red and the user-facing chains-add path lacks the requested command-boundary regression; please reopen a focused current-main PR if the UX improvement is still wanted.

PR #1746, opened 2026-08-05
fix(cli): execute cosmwasm msg + funds validation (beads b0ome + 5ze6w)

NeOMakinG:
## Summary
- `execute` cosmwasm msg parsing now rejects malformed JSON as typed InvalidInputError (was UNKNOWN_ERROR), plus rejects arrays / null / primitives (CosmWasm ExecuteMsg must be an object)
- `execute --funds` parser now rejects negative amounts (cosmos coins are unsigned integer base units)
- Fixes beads **vultisig-b0ome** (P3) + **vultisig-5ze6w** (P3)

### bead b0ome — msg loose validation
```
$ vault-cli execute THORChain <addr> 'not json' --dry-run --vault X -o json
{ code: 'UNKNOWN_ERROR', exitCode: 7, message: 'Invalid JSON message: not json' }
                       ↑ misclassified as service error

$ vault-cli execute THORChain <addr> '[1,2,3]' --dry-run --vault X -o json
{ success: true, msg: [1,2,3] }   ← CosmWasm requires object, arrays never valid

$ vault-cli execute THORChain <addr> 'null' --dry-run --vault X -o json
{ success: true, msg: null }
```

### bead 5ze6w — negative funds accepted
```
$ vault-cli execute THORChain <addr> '{"swap":{}}' --funds 'rune:-1000000' --dry-run --vault X -o json
{ success: true, funds: [{ denom: 'rune', amount: '-1000000' }] }
```

Cosmos SDK would reject at broadcast/simulate, but the user has already been asked to confirm at the CLI layer.

## After
```
$ vault-cli execute THORChain <addr> 'not json' --dry-run --vault X -o json
{ code: 'INVALID_INPUT', exitCode: 4, message: 'Invalid JSON message: not json' }

$ vault-cli execute THORChain <addr> '[1,2,3]' --dry-run --vault X -o json
{ code: 'INVALID_INPUT', exitCode: 4, message: 'CosmWasm execute msg must be a JSON object (received array)' }

$ vault-cli execute THORChain <addr> '{"swap":{}}' --funds 'rune:-1000000' --vault X -o json
{ code: 'INVALID_INPUT', exitCode: 4, message: 'Invalid funds amount: "-1000000" in "rune:-1000000" — expected a non-negative integer base-unit amount (cosmos coins are unsigned)' }
```

## Implementation
- Import `InvalidInputError` from `../core/errors`
- Replace both plain `throw new Error(...)` sites with typed `InvalidInputError`
[...]

PR #1750, opened 2026-08-05
fix(cli): portfolio --currency no longer persists vault preference (bead v1n9u)

NeOMakinG:
## Summary
- `portfolio --currency <X>` no longer persists X as the vault's currency preference
- Bead **vultisig-v1n9u** (dogfood 2026-08-05, P3)

## Motivation
`portfolio --currency EUR` silently called `vault.setCurrency('eur')` as a side effect. A user asking for a one-off EUR display had EUR persisted as the vault's default until they ran `currency USD` to fix it. The `--currency` flag became a permanent-setting flag they never asked for.

```
$ vault-cli currency --vault Testwallet2 -o json
{ currency: 'usd', name: 'US Dollar' }

$ vault-cli portfolio --currency EUR --vault Testwallet2 -o json
{ portfolio: { totalValue: { amount: '49.24', currency: 'eur' }, ... } }

$ vault-cli currency --vault Testwallet2 -o json
{ currency: 'eur', name: 'Euro' }   ← silently persisted
```

## Implementation
`getValues(chain, currency)` and `getValue(chain, _, currency)` both accept currency explicitly, so no `setCurrency` call is needed. Simple removal of the `if (vault.currency !== currency) { await vault.setCurrency(currency) }` block at `balance.ts:93-95`.

## Behaviour scope
| Command | Before | After |
|---------|--------|-------|
| `portfolio --currency X` | displays in X **AND** persists X | displays in X, vault unchanged |
| `currency X` (dedicated) | persists X | unchanged |
| `swap-quote`, `balance`, `send`, etc | unchanged (no --currency flag) | unchanged |

## Tests
Two new tests in `balance.test.ts`:
1. `does NOT persist --currency to the vault preference` — pins `setCurrency` is never called from `executePortfolio`
2. `passes --currency through to getValues without touching vault.currency` — pins the requested currency reaches getValues via the explicit arg

Existing tests untouched — they all use `currency: 'usd'` (matching vault's initial 'usd'), so the removed setCurrency call wasn't firing in any pre-existing test path.

## Related
Same "correct layer" family as PRs #1741 (broadcast rawTx), #1743 (slippage 0), #1746 (execute inputs), #1747 (seedphrase order), #1749 (create fast validation) — this one is "side-effect at correct layer" rather than "validate at correct layer".

Fix via Claude Opus 4.7

## Summary by CodeRabbit

* **Bug Fixes**
  * The portfolio balance command now treats the selected currency as display-only.
  * Running the command no longer changes or persists the vault’s default currency.
  * Portfolio values continue to be displayed in the requested currency.

### DIFF
--- a/clients/cli/src/commands/balance.test.ts
+++ b/clients/cli/src/commands/balance.test.ts
@@ -391,7 +391,7 @@ describe('executePortfolio partial-failure reporting', () => {
   it('passes --currency through to getValues without touching vault.currency', async () => {
     // The requested currency reaches the pricing layer via the explicit
     // getValues(chain, currency) arg — no vault mutation needed.
-    const getValues = vi.fn(async (chain: Chain, _currency: FiatCurrency) => ({ native: makeValue('50.00') }))
+    const getValues = vi.fn(async (_chain: Chain, _currency: FiatCurrency) => ({ native: makeValue('50.00') }))
     const ctx = makeCtx({
       chains: [Chain.Ethereum],
       balance: async () => makeBalance('ETH'),
--- a/clients/cli/src/commands/balance.test.ts
+++ b/clients/cli/src/commands/balance.test.ts
@@ -369,6 +369,40 @@ describe('executePortfolio partial-failure reporting', () => {
     expect(joined).toContain('Bitcoin')
     expect(joined).toContain('btc unreachable')
   })
+
+  it('does NOT persist --currency to the vault preference (bead v1n9u)', async () => {
+    // Regression guard: `portfolio --currency EUR` previously called
+    // vault.setCurrency('eur') as a side effect, silently persisting EUR as
+    // the vault's default. --currency is a DISPLAY option — vault preference
+    // only changes via the dedicated `currency <X>` command.
+    const ctx = makeCtx({
+      chains: [Chain.Ethereum],
+      balance: async () => makeBalance('ETH'),
+      getValue: async () => makeValue('100.00'),
+    })
+    const vault = await ctx.ensureActiveVault()
+
+    await captureJson(() => executePortfolio(ctx, { currency: 'eur' }))
+
+    // The mock's setCurrency was NOT called — --currency stayed display-only.
+    expect((vault as unknown as { setCurrency: ReturnType<typeof vi.fn> }).setCurrency).not.toHaveBeenCalled()
+  })
+
+  it('passes --currency through to getValues without touching vault.currency', async () => {
+    // The requested currency reaches the pricing layer via the explicit
+    // getValues(chain, currency) arg — no vault mutation needed.
+    const getValues = vi.fn(async (chain: Chain, _currency: FiatCurrency) => ({ native: makeValue('50.00') }))
+    const ctx = makeCtx({
+      chains: [Chain.Ethereum],
+      balance: async () => makeBalance('ETH'),
+      getValue: async () => makeValue('50.00'),
+      getValues,
+    })
+
+    await captureJson(() => executePortfolio(ctx, { currency: 'gbp' }))
+
+    expect(getValues).toHaveBeenCalledWith(Chain.Ethereum, 'gbp')
+  })
 })
 
 describe('executeBalance honours --tokens on a single chain', () => {
--- a/clients/cli/src/commands/balance.ts
+++ b/clients/cli/src/commands/balance.ts
@@ -90,10 +90,13 @@ export async function executePortfolio(ctx: CommandContext, options: PortfolioOp
     throw new Error('Invalid currency')
   }
 
-  if (vault.currency !== currency) {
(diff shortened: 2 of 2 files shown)


gomesalexandre (dismissed):
Does what it says on the box, and it's the right kind of fix: deleting a side effect rather than adding a flag to suppress it. Approving.

## verified

**must-pass:** `src/commands/balance.test.ts` **17/17**.

**must-fail:** re-injected the removed side effect above your comment block:
```ts
if (vault.currency !== currency) { await vault.setCurrency(currency) }
```
(anchor verified unique, marker confirmed applied) -> **1 fails**: `does NOT persist --currency to the vault preference (bead v1n9u)`. So the regression is pinned, not just fixed.

**The thing I actually wanted to check** - whether the `setCurrency` call was load-bearing for *display* rather than just persistence. It wasn't:
- `:133` `await vault.getValues(chain, currency)`
- `:146` `await vault.getValue(chain, undefined, currency)`
- `:100` `fiatCurrencyNameRecord[currency]`, `:202` `{ amount, currency, ... }`, `:214` `displayPortfolio(portfolio, currency, ...)`

and `grep vault.currency` on the file now returns **nothing**. So there's no residual read that would have kept formatting against the stale preference. That was the failure mode worth ruling out: if the amounts came from the flag but the label came from `vault.currency`, you'd get EUR figures printed as USD, which is worse than the bug you're fixing and would have looked green in these tests.

Your second test (`passes --currency through to getValues without touching vault.currency`) covers exactly that seam, which is the right instinct - asserting the positive as well as the absence.

## why this class matters more than P3 suggests

A read command that mutates persisted state is nasty precisely because it's invisible: the user gets what they asked for *this time*, so nothing looks wrong, and the surprise lands on some later invocation with no flag in sight. The blast radius here is cosmetic (a fiat display preference), but the pattern - "display option quietly became a setter" - is worth the fix on principle. Good catch from dogfooding.

**nit, entirely optional:** now that `--currency` is purely presentational, `currency <X>` is the only writer. Might be worth a line in the `portfolio` help text saying `--currency` affects this output only, since the previous behavior means some users have learned the opposite.

gomesalexandre (commented):
Remote pass on current head `139eaa8a7e9d6f78302836715a99c78157b8e8ba` looks good.

What I checked:
- `portfolio --currency <X>` now stays display-only instead of mutating the vault preference.
- The code still passes the requested currency explicitly into the pricing layer, so this is not a fake fix that only stops persistence while leaving display formatting tied to `vault.currency`.
- The added tests cover both seams: no `setCurrency` side effect and explicit currency propagation through `getValues(...)`.

I am leaving this as a comment, not an approval, because the current head contains gomes-authored follow-up commits. Needs a non-gomes stamp on this head for four-eyes.

Receipts used this pass:
- current-head remote diff via `gh pr diff`
- current-head source reads via `gh api ...?ref=139eaa8a7e9d6f78302836715a99c78157b8e8ba`
- current PR metadata/check state via `gh pr view`

gomesalexandre (approved):
Genuine approve on the code, with a flagged CI caveat below — please rebase before merge.

Only review blocker was the four-eyes bug (gomes-authored commit on the branch wrongly treated as self-approve-disqualifying — that rule only fires when the reviewer IS the PR author). The fix itself is small and correct: `portfolio --currency X` no longer calls `vault.setCurrency(X)` as a side effect, it just passes the currency explicitly through to `getValues`/`getValue` for display. A prior review on this exact head (`139eaa8a7`) did a must-fail mutation check (re-injecting the removed `setCurrency` call flips exactly one test red) and confirmed the display path doesn't silently regress to formatting against a stale `vault.currency` — no residual reads found. That's the right thing to verify for this class of "read command mutates persisted state" bug.

## CI status — investigated, both failures are unrelated to this diff
Checked the actual failing jobs on the current head, not just the red X:
- **Unit Tests**: failed on `Failed to resolve action download info: Service Unavailable` — a GitHub Actions infra flake fetching the action itself, before any test ran. Nothing to do with this PR.
- **Lint and Type Check**: failed on the dependency-audit step flagging `js-yaml` (GHSA-5p4m-2wfm-xmqj, quadratic CPU in `!!omap` resolution) via `read-yaml-file`/`@eslint/eslintrc` — a transitive advisory that postdates this branch's last rebase onto main, same class of stale-CI issue flagged and tracked elsewhere (see #1916 discussion on PR #1736). Not introduced by this diff.

Recommend rebasing onto fresh main to pick up whatever resolves the advisory before merging, but I don't consider either failure a reason to hold the review itself — the code change is sound and neither failure is reachable from this diff.

rcoderdev (approved):

rcoderdev:
Closing this because this two-file CLI fix is more than 250 mainline commits behind with required unit, lint, and summary checks red; recreate the display-only currency change from current main and rerun today's gates.

PR #1753, opened 2026-08-05
fix(storage): prevent lost updates between vault instances

### LINKED ISSUE #1212: fix(storage): prevent lost updates between vault instances
## Problem

Vault saves replace a full record without revision or compare-and-swap semantics. Two live instances editing different fields can overwrite one another, losing metadata or lifecycle updates.

Evidence:
- `packages/sdk/src/Vault.ts:247-298`
- `packages/sdk/src/VaultManager.ts:300-347`
- `packages/sdk/src/storage/VaultStorage.ts:92-137`

## Acceptance criteria

- Add a monotonic record revision or equivalent CAS contract.
- Reject stale writes with a typed conflict.
- Provide a narrow merge/retry path for non-conflicting metadata.
- Serialize same-instance writes.
- Test two-instance conflicts and deterministic recovery.

rcoderdev:
Closes #1212

## Summary by CodeRabbit

* **New Features**
  * Added optimistic concurrency protection to prevent stale vault data from overwriting newer changes.
  * Added structured storage conflict errors with revision and conflicting-field details.
  * Added an optional metadata-merge strategy for non-conflicting updates.
  * Exposed vault save options and revision information through the SDK.
* **Bug Fixes**
  * Prevented unintended mutations of in-memory stored data.
  * Improved restoration of pending vaults and preserved persisted vault state during imports.
  * Improved validation and protection for vault names and storage paths.

### DIFF
--- a/.changeset/fresh-vaults-revise.md
+++ b/.changeset/fresh-vaults-revise.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Prevent stale vault instances from overwriting newer storage updates, with explicit conflict errors and a narrow metadata merge retry.
--- a/.changeset/fresh-vaults-revise.md
+++ b/.changeset/fresh-vaults-revise.md
@@ -1,5 +1,6 @@
 ---
-'@vultisig/sdk': patch
+'@vultisig/sdk': minor
 ---
 
-Prevent stale vault instances from overwriting newer storage updates, with explicit conflict errors and a narrow metadata merge retry.
+Prevent stale vault instances from overwriting newer storage updates. `vault.save()` now rejects stale writes with
+`VaultConflictError`; pass `{ conflictStrategy: 'merge-metadata' }` to retry a non-overlapping metadata edit.
--- a/packages/sdk/src/VaultManager.ts
+++ b/packages/sdk/src/VaultManager.ts
@@ -246,6 +246,7 @@ export class VaultManager {
 
       // Use ECDSA public key as vault ID
       const vaultId = parsedVault.publicKeys.ecdsa
+      const persistedVaultData = await this.storage.get<VaultData>(`vault:${vaultId}`)
 
       // Determine vault type from parsed vault
       const vaultType = parsedVault.signers.some((s: string) => s.startsWith('Server-')) ? 'fast' : 'secure'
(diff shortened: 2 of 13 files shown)


gomesalexandre (approved):
Read this as a data-integrity change and went looking for the two ways a JSON-clone-based concurrency layer usually breaks. Both are fine here. One real coverage gap, non-blocking.

## things I checked that turned out clean

**`cloneValue` via `JSON.parse(JSON.stringify(...))` is safe for this payload.** That idiom throws on `BigInt` and silently eats `undefined`/`Date`, so it's the first thing worth ruling out when it's cloning persisted state. `VaultData` is all strings/numbers/booleans/arrays, and the nested `tokens: Record<string, Token[]>` is too - `Token` has zero bigint fields. The bigints in `types/index.ts` (`balance`, `fee`, `maxSendable`) belong to `MaxSendAmount`, which never enters `VaultData`. So no throw path here. Worth a comment at `cloneValue` pinning that assumption, because the day someone adds a `bigint` to `Token` this fails at save time, not compile time.

**`MERGEABLE_VAULT_FIELDS` is correctly scoped.** Every entry is a mutable preference (`name`, `isBackedUp`, `order`, `folderId`, `currency`, `chains`, `tokens`, `lastValueUpdate`). No `publicKeys`, no `chainKeyShares`, no `vultFileContent` - so `merge-metadata` can never auto-resolve a conflict on cryptographic identity. That's the right allowlist, and allowlisting rather than denylisting is the right shape.

**must-pass:** `tests/unit/vault/VaultManager.test.ts` **37/37**.

**must-fail on the revision check:** neutered `if (expectedRevision !== actualRevision)` to `if (false)` (pre-count 1, post-marker 1) -> **2 tests fail** (`rejects a stale instance and deterministically merges non-overlapping metadata on explicit retry`, `keeps overlapping metadata edits conflicted during merge retry`). Primary guard is pinned.

**merged-tree check:** PR is `BEHIND`; merged `origin/main` locally - clean, 169 files / +7943 (main has moved a lot today). Full sdk suite on the merged tree: **210 files, 3181/3181**. Reset locally afterwards, your branch untouched.

## should-fix: the write lock is the untested half of the headline guarantee

`withStorageWriteLock` serializes writes per `(storage, key)`. I neutered the chaining:
```ts
const previous = queues.get(key) ?? Promise.resolve()
->
const previous = Promise.resolve()
```
so every operation starts immediately instead of queueing behind the last one. Mutation verified applied, `tsc --noEmit` **clean** (so this is a live, type-valid regression, not a broken control). Result: **37/37 still pass.**

That matters more than a normal coverage nit, because the lock isn't decoration - it's the mechanism covering the same-process race. Two concurrent `save()` calls on one vault can both read revision N, both pass the `expectedRevision !== actualRevision` check while the other is still in flight, and both write N+1. One update is lost. That is precisely the failure in the title, and right now only the *cross-instance* half of it is pinned.

Suggest one test that fires two `save()` calls without awaiting the first, and asserts both land (second observing the first's revision) rather than one clobbering the other. If it's awkward to write because `MemoryStorage` resolves synchronously, that's itself the signal - a storage stub with a deferred write is what makes the race observable.

Not blocking: the revision check still catches the cross-instance case that #1212 reports, and a lost update here costs a renamed vault or a reordered list, not funds.

`Closes #1212` present.

PR #1754, opened 2026-08-05
fix(sdk): harden React Native Solana broadcasts

### LINKED ISSUE #1227: refactor: per-chain tx logic spans three trees with diverged broadcast guarantees (rn solana is the weakest)
Per-chain tx logic lives in three parallel trees (core `packages/core/chain/chains/*` + `tx/broadcast/resolvers/*`, sdk node-side `packages/sdk/src/chains/*`, RN platform `packages/sdk/src/platforms/react-native/chains/*`), and their robustness guarantees have already diverged:

- core solana broadcast is hardened: JITO + retry + AlreadyProcessed idempotency + verifyBroadcastByHash (`packages/core/chain/tx/broadcast/resolvers/solana.ts`)
- RN ripple submit is hardened: typed XrpSubmitRejectedError + on-ledger tec* recheck (`packages/sdk/src/platforms/react-native/chains/ripple/rpc.ts:238`)
- RN solana broadcast is a bare `sendTransaction` wrapper (`packages/sdk/src/platforms/react-native/chains/solana/rpc.ts:63`): no idempotency handling, no post-broadcast verification

Same chain, different platform, different broadcast-outcome guarantees. Fixing or adding a chain means touching up to three trees in lockstep with no documented ownership map, and the RN consumer inherits the weakest variant.

## fix shape

1. Document the authoritative-tree boundary per concern (build vs broadcast vs verify) so future chain work knows where a fix belongs.
2. Bring RN solana broadcast to parity with the core resolver (idempotent AlreadyProcessed handling + post-broadcast verification), or make the "caller must confirm" contract explicit in the primitive's types so the consumer cannot miss it.

effort: L (design decision, then mechanical) / risk: MED / confidence: MED-HIGH (divergence itself is verified line-by-line)

rcoderdev:
## Summary

- document transaction ownership across core, SDK, and React Native surfaces
- harden React Native Solana broadcasts with signed-payload identity, duplicate idempotency, bounded status verification, and cancellation
- share the pure raw-transaction parser with the SDK broadcast service

## Runtime proof

An exact-revision local HTTP JSON-RPC run returned the payload-derived signature, recovered after indexing lag, preserved the original error when status stayed unknown, rejected a mismatched RPC identity, and stopped promptly on caller abort.

Closes #1227

## Summary by CodeRabbit

* **New Features**
  * Added Solana raw transaction decoding and signature extraction support.
  * Expanded React Native Solana broadcast options and returned signature handling.
* **Bug Fixes**
  * Improved Solana broadcast reliability by handling duplicate submissions and ambiguous send errors more safely.
  * Added post-send verification to confirm transaction status when results are unclear.
* **Documentation**
  * Added architecture guidance for transaction ownership and broadcast responsibilities.
* **Tests**
  * Added broader coverage for Solana broadcasting, encoding detection, retries, aborts, and error cases.

### DIFF
--- a/.changeset/calm-bats-verify-solana.md
+++ b/.changeset/calm-bats-verify-solana.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Harden React Native Solana broadcasting with deterministic signatures, idempotent duplicate handling, and signature-status verification for ambiguous send errors.
--- a/docs/architecture/ARCHITECTURE.md
+++ b/docs/architecture/ARCHITECTURE.md
@@ -1460,6 +1460,12 @@ Each chain kind has specific:
 - **RPC endpoints** - Chain-specific API URLs
 - **Transaction format** - Chain-specific encoding
 
+Transaction construction, broadcast, confirmation, and React Native transport
+have separate ownership boundaries. See
+[Chain transaction ownership](CHAIN-TRANSACTION-OWNERSHIP.md) before changing a
+chain implementation so platform adapters retain the core broadcast and status
+guarantees.
+
 ### Token Support
 
 | Chain Type | Token Standard | Example         |
--- a/docs/architecture/CHAIN-TRANSACTION-OWNERSHIP.md
+++ b/docs/architecture/CHAIN-TRANSACTION-OWNERSHIP.md
@@ -1,0 +1,32 @@
+# Chain transaction ownership
+
+Chain transaction code exists in several trees because the published packages
+and the React Native bundle have different dependency constraints. The trees
+are adapters, not independent specifications. Use the boundaries below when
+adding a chain or changing transaction behavior.
+
+| Concern | Authoritative tree | Adapter rule |
+| --- | --- | --- |
+| Keysign transaction construction | `packages/core/mpc/keysign/signingInputs/resolvers/<chain>` | SDK services should delegate to the core resolver. Standalone primitives in `packages/sdk/src/chains/<chain>` own only their explicitly documented raw/public API. |
+| Shared chain RPC/config primitives | `packages/core/chain/chains/<chain>` | Reuse these outside React Native. Do not duplicate endpoint or chain rules in SDK services. |
+| Broadcast outcome semantics | `packages/core/chain/tx/broadcast/resolvers/<chain>` | `RawBroadcastService` and `packages/sdk/src/platforms/react-native/chains/<chain>` must preserve the same idempotency and ambiguous-error verification guarantees. Platform adapters may change transport, not the success contract. |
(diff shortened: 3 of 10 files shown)


gomesalexandre (approved):
Reviewed this as a broadcast-path change, so the failure mode I cared about is the expensive one: a send that **actually landed** being reported as failed, the user re-signing, and the funds going out twice. This gets that right, and the reasoning is written down rather than implied. Approving.

## the two properties that matter, and how they hold

**1. The returned signature is derived from the payload, never taken from the RPC.**
```ts
const signature = deriveSolanaRawTxSignature(rawTxBase64, 'base64')
...
if (sendResult.data !== signature) {
  throw new Error('Solana RPC returned a signature that does not match the signed transaction')
}
```
This is the part I'd have flagged if it were missing. A node that returns someone else's signature would otherwise hand the caller a tx id that tracks a transaction the user never signed - and every downstream status poll would then confirm the wrong thing. must-fail: neutered the comparison to `if (false)` (pre-count 1 / post-marker 1) -> **3 tests fail**, including the explicit `rejects an invalid successful send result ("rpc-supplied-signature")`. Well pinned.

**2. `AlreadyProcessed` is treated as idempotent success, not as an error.**
must-fail: neutered `isAlreadyProcessedError(...)` -> **2 tests fail** (`treats AlreadyProcessed in message as idempotent success`, and the structured-`error.data` variant). Both take ~3s under mutation because they fall through into the bounded verification loop, which is itself a nice incidental confirmation that the backoff path is live rather than short-circuited.

This is the double-send guard. Without it a retried broadcast reports failure on a transaction that already landed.

**The ambiguous branch fails in the right direction.** For any other send error it polls `getSignatureStatuses` with bounded backoff and only resolves on `status.err === null`; a status carrying an `err` breaks out and the **original** send error is rethrown rather than being replaced by a vaguer one. Missing/unparseable status keeps polling to the cap and then rethrows. So "unknown" never gets laundered into "succeeded" - which is the correct bias for a broadcast helper.

**must-pass:** `tests/unit/platforms/react-native/solana-rpc.test.ts` **19/19**.

**merged-tree check.** PR is `BEHIND` and its CI ran on a stale base - main has moved several times today. Merged `origin/main` locally: clean, 15 files / +232. Full sdk suite on the merged tree: **211 files, 3196/3196**. Reset locally afterwards; your branch is untouched.

## what I liked

The doc comment states the guarantee *and* its limit:

> A returned signature proves broadcast acceptance, not final confirmation. Consumers must continue polling transaction status until success or failure.

That distinction is exactly where broadcast helpers usually mislead their callers, and writing it at the seam is worth more than another test. Same for `CHAIN-TRANSACTION-OWNERSHIP.md` - the reason this class of bug recurs is that nobody can say which layer owns the tx, so documenting it is the actual fix.

No blockers. `Closes #1227` present. Rebase for the stale base and it's good from my side.

PR #1755, opened 2026-08-05
fix: keep fiat and discount tier amount math exact

### LINKED ISSUE #1195: correctness: fromChainAmount float division feeds swap-confirm display + fiat tier (plans-file deferral, never filed)
SDK plans/README.md report-only deferral (SDK-CORRECTNESS-01/02), never tracked: `fromChainAmount` does float division and its output feeds the swap confirm display and fiat/tier math - precision loss on large-decimal amounts can display a subtly wrong number at the exact moment the user confirms. The named fix is one shared bigint `formatBaseUnits` helper used by both sites ("small, self-evident - do it whilst in the house when touching the swap-build path"). Also the sibling: `send()`'s parseAmount divergence (SDK-CORRECTNESS-03/05) - route send() through toChainAmount in its VaultError catch. Effort: S.

rcoderdev:
Closes #1195

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fiat value calculations for very large balances and negative amounts while preserving exact base-unit precision.
  * Corrected discount-tier evaluation at precise boundaries, preventing balances just below a threshold from being incorrectly upgraded.
  * Improved Thorguard NFT tier handling, including accurate upgrades and preservation of existing higher tiers.

* **Tests**
  * Added regression coverage for precision-sensitive fiat conversions and discount-tier behavior.

### DIFF
--- a/.changeset/exact-fiat-tier-amounts.md
+++ b/.changeset/exact-fiat-tier-amounts.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Fix fiat valuation losing base-unit precision through float-first conversion, with regression coverage for exact VULT discount-tier boundaries.
--- a/packages/core/chain/coin/utils/getCoinValue.test.ts
+++ b/packages/core/chain/coin/utils/getCoinValue.test.ts
@@ -1,0 +1,17 @@
+import { describe, expect, it } from 'vitest'
+
+import { getCoinValue } from './getCoinValue'
+
+describe('getCoinValue', () => {
+  it('calculates ordinary fiat values', () => {
+    expect(getCoinValue({ amount: 1500000n, decimals: 6, price: 2 })).toBe(3)
+    expect(getCoinValue({ amount: -1500000n, decimals: 6, price: 2 })).toBe(-3)
+  })
+
+  it('converts base units exactly before entering number-based fiat math', () => {
+    const amount = 831048943873725519n
+
+    expect(Number(amount) / 10 ** 6).toBe(831048943873.7256)
+    expect(getCoinValue({ amount, decimals: 6, price: 1 })).toBe(831048943873.7255)
+  })
+})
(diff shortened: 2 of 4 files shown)


gomesalexandre (approved):
This lands in a defect class I'd swept for independently this morning (float coercion inside money paths), so I probed it harder than the +58/-3 suggests. The `getCoinValue` fix is real and the controls hold. Approving, with one accuracy note on the framing.

## verified

**must-pass:** `yarn test:core` (the runner that actually covers `packages/core` - `.config/vitest.core.config.ts`, not the sdk's `tests/**` config) -> **2461 passed / 1 skipped, 242 files**.

**must-fail:** reintroduced the pre-fix float-first conversion in `getCoinValue`:
```ts
return sign * (Number(absoluteAmount) / 10 ** decimals) * price
```
Mutation verified applied (pre-count 1, post-marker 1). **1 test fails** - `converts base units exactly before entering number-based fiat math`. So the fix is genuinely pinned, not decorative.

That test is the good kind, because it shows its own work:
```ts
expect(Number(amount) / 10 ** 6).toBe(831048943873.7256)   // the old path
expect(getCoinValue({ amount, decimals: 6, price: 1 })).toBe(831048943873.7255)  // the new one
```
Double-rounding (once converting the bigint to a number, once dividing) versus converting exactly first. Anyone touching this later can see the failure mode without reconstructing it.

**merged-tree check.** PR is `BEHIND` and its CI ran on a stale base - main moved twice this morning (#1588, #1601, plus four more since). Merged `origin/main` locally: clean, 15 files / +232. Core suite on the merged tree: **2461 passed / 1 skipped**, identical. Reset locally afterwards, your branch untouched.

## nit: the title oversells the tier half

`getVultDiscountTier` never routed through `getCoinValue`. It compares bigints directly:
```ts
descendingTiers.find(({ value }) => vultBalance >= toChainAmount(value, vult.decimals))
```
Under the must-fail mutation above, the tier tests kept passing - only the `getCoinValue` test flipped. So the tier path was already exact, and the four cases you added are **regression pins**, not fixes.

That's worth saying explicitly rather than reading it as "discount tier math was broken", because `does not round one wei below bronze up into the bronze tier` reads like a shipped bug being closed. It wasn't - and the test proving it is the nicest thing in the diff:
```ts
expect(Number(bronzeThreshold - 1n) / 10 ** vult.decimals).toBe(vultDiscountTierMinBalances.bronze)
expect(getVultDiscountTier({ vultBalance: bronzeThreshold - 1n, ... })).toBeNull()
```
It documents that a float literally cannot separate one wei below the threshold from the threshold, and then pins that we don't rely on a float. If someone ever "simplifies" that comparison to a number, this catches a real free-discount bug. Suggest a title closer to "fix fiat valuation precision; pin exact tier boundary math" so the changeset and CHANGELOG don't imply a tier bug that never shipped.

`Closes #1195` present in the body.

Rebase for the stale base and it's good from my side.

PR #1756, opened 2026-08-05
fix(swap): allow RUNE/CACAO-source native swaps past the broadcast guard

### LINKED ISSUE #1726: Native swaps from RUNE/CACAO can never broadcast: inbound-vault guard rejects the native source chain
## Summary

Every native swap whose **source chain is THORChain (RUNE) or MayaChain (CACAO)** is signed and then refused locally at the broadcast step. Nothing reaches the network.

`assertNativeSwapReadyForBroadcast` resolves the **source** chain in the `/inbound_addresses` feed and throws when it finds no entry. That feed only lists external chains — THOR and MAYA are never in it, by design: a native RUNE swap is a `MsgDeposit` sent straight to the network and has no inbound vault. So the lookup is `undefined` for every RUNE-source swap and the guard throws 100% of the time.

## Impact

- RUNE → any asset cannot be swapped from the Vultisig extension or the Windows/Linux desktop app. Same expected for CACAO on Maya.
- The user-facing copy is misleading: the UI reports **"The network rejected this transaction"** and offers "sign again to retry". The network never saw the transaction, and retrying can never succeed.
- Inherited by **Windows + the extension + agent-app** (they consume `@vultisig/core-mpc`). iOS/Android are native and unaffected.

## Observed

QA of `vultisig-windows` `v1.0.70` (commit `1b7738a4e`), Chrome MV3 extension build, funded test vault.

Route `THOR.RUNE → ETH.ETH`, 2 RUNE, attempted 4 times.

1. Quote resolves normally — `Quote loaded: output = 0.00033679` (matches thornode's own quote for the same input).
2. Keysign completes.
3. Error screen:
   ```
   The network rejected this transaction
   Your devices signed the transaction, but the network rejected it.
   Review the on-chain reason below, then sign again to retry.

   Error message
   Cannot validate THORChain inbound vault for source chain THOR
   ```
4. Nothing broadcast — the vault's RUNE balance stayed at exactly `17.9408455` across all 4 attempts.

The network was healthy throughout: `ETH halted=false`, `chain_trading_paused=false`, `global_trading_paused=false`, and a live `THOR.RUNE → ETH.ETH` quote resolved at the same moment. An ETH send from the same vault in the same session signed and landed on-chain normally (`0x0d05988718eb309e432b7c2466edc1d954439ba25f6a85392da6f643397453d9`, block 25681906), so MPC signing itself is fine.

## Cause

`packages/core/mpc/keysign/swap/assertNativeSwapReadyForBroadcast.ts:114-119`

```ts
const activeInbound = inboundByChainId.get(sourceChainId.toUpperCase())   // "THOR"

if (!activeInbound?.address) {
  throw new NativeSwapBroadcastGuardError(
    'network_error',
    `Cannot validate ${native.chain} inbound vault for source chain ${sourceChainId}`
  )
}
```

`/inbound_addresses` returns, verified live:

```
AVAX, BASE, BCH, BSC, BTC, DOGE, ETH, GAIA, LTC, SOL, TRON, XRP
```

No `THOR`. `inboundByChainId.get('THOR')` is always `undefined`.

The same file already documents this exact case — in the halt-check loop ~25 lines below:

> a route leg that resolves to no inbound entry (e.g. RUNE/CACAO, the native chain itself) is not haltable via this feed, so it is skipped rather than false-blocked

That tolerance was applied to the halt check but not to the `activeInbound` existence check above it, nor to the `vaultAddress` comparison at `:157` (which is likewise meaningless when there is no inbound vault).

## Why it surfaced now

- #1392 moved this guard out of `packages/sdk` (where no shipping wallet called it) into `packages/core`, making it live for real clients.
- #1360 added the halt-check loop, including the RUNE/CACAO tolerance quoted above — applied only to that loop.
- `vultisig-windows` #4507 wired the guard into the keysign flow at `core/ui/mpc/keysign/assertReadyToBroadcast.ts` → `core/ui/mpc/keysign/action/mutations/useKeysignMutation.ts:337`. That file is new in `v1.0.70`; it does not exist in `v1.0.69`.

Consistent with that: the `v1.0.68` release sweep broadcast a 40 RUNE → ETH swap successfully (`08095EF11A5B0E3E1DBFF30A861A19139671B7678D20730976732C6628696951`).

`@vultisig/core-mpc` moved `1.14.1 → 1.15.0` between `v1.0.69` and `v1.0.70`.

## Suggested fix

When the swap **source** is the native chain itself (THOR for THORChain, MAYA for Maya), skip the inbound-vault existence check and the `vaultAddress` comparison — mirroring the tolerance the halt loop already implements. Quote-expiry and the halt checks should still run for those routes.

A regression test covering `THOR.RUNE → ETH.ETH` (source = native chain, no inbound entry) would have caught this; current coverage only exercises external-chain sources.

Ehsan-saradar:
Fixes #1726

## Problem

`assertNativeSwapReadyForBroadcast` resolves the swap's **source** chain in the `/inbound_addresses` feed and throws when it finds no entry. That feed only lists external chains — THOR and MAYA are never in it, by design: a native RUNE/CACAO swap is a `MsgDeposit` sent straight to the network and has no inbound vault. So every RUNE-source (and CACAO-source) swap was signed and then refused locally at the broadcast step, with a misleading "network rejected this transaction" error. Nothing ever reached the network.

## Fix

When the swap source is the protocol's own chain (`chain === native.chain`), skip the inbound-vault existence check and the vault-address staleness comparison — the same tolerance the halt-check loop in this file already applies to route legs with no inbound entry. Everything else still runs for those routes:

- quote-expiry validation
- source/destination trading-halt checks (`halted`, `chain_trading_paused`)
- network-wide `global_trading_paused`

A RUNE source routed **via Maya** is deliberately not exempt: Maya's feed lists THOR as an external chain, so the full vault validation still applies there (covered by a test).

## Tests

Added regression coverage for source-is-native-chain routes (the gap called out in the issue): RUNE- and CACAO-source swaps pass with no THOR/MAYA feed entry, still reject on destination halt / global pause / expired quote, and skip the vault-address comparison even if the feed ever grew a THOR entry.

Two existing tests pinned the secured-withdrawal non-exemption behavior through the vault-address check on a THOR-source payload — an observable this fix removes on purpose. They now pin the same non-exemption through the expiry check the exemption would have skipped.

`yarn check` passes; the guard suite is 29/29. The handful of CLI config-dir/credential-permission test failures in `yarn test` are pre-existing on `main` (verified via stash) and unrelated.

### DIFF
--- a/.changeset/native-source-swap-broadcast-guard.md
+++ b/.changeset/native-source-swap-broadcast-guard.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Allow native swaps whose source is the protocol's own chain (RUNE on THORChain, CACAO on MayaChain) past the broadcast guard: those routes are MsgDeposits with no inbound vault and never appear in /inbound_addresses, so the inbound-existence and vault-address checks are skipped for them while quote-expiry and trading-halt checks still run.
--- a/packages/core/mpc/keysign/swap/assertNativeSwapReadyForBroadcast.ts
+++ b/packages/core/mpc/keysign/swap/assertNativeSwapReadyForBroadcast.ts
@@ -111,9 +111,17 @@ export const assertNativeSwapReadyForBroadcast = async ({
   }
   const inboundByChainId = new Map(inboundAddresses.map(info => [info.chain.toUpperCase(), info]))
 
+  // sdk#1726: a swap whose SOURCE is the protocol's own chain (RUNE on THORChain, CACAO on
+  // MayaChain) is a MsgDeposit straight to the network - it has no inbound vault, and
+  // /inbound_addresses only lists external chains, so the lookup below is undefined by design.
+  // Skip the inbound-existence and vault-address checks for those routes (the same tolerance the
+  // halt loop below already applies); quote expiry and the halt checks still run. A RUNE source
+  // routed VIA Maya is NOT exempt: Maya's feed lists THOR as an external chain.
+  const sourceIsProtocolChain = chain === native.chain
+
   const activeInbound = inboundByChainId.get(sourceChainId.toUpperCase())
 
-  if (!activeInbound?.address) {
+  if (!sourceIsProtocolChain && !activeInbound?.address) {
     throw new NativeSwapBroadcastGuardError(
       'network_error',
       `Cannot validate ${native.chain} inbound vault for source chain ${sourceChainId}`
@@ -154,7 +162,13 @@ export const assertNativeSwapReadyForBroadcast = async ({
     }
   }
 
-  if (activeInbound.address.toLowerCase() !== native.vaultAddress.toLowerCase()) {
+  // MsgDeposit payloads carry no vault address, so the staleness comparison is meaningless for
+  // protocol-chain sources even if the feed ever grew a THOR/MAYA entry.
+  if (sourceIsProtocolChain) {
+    return
+  }
+
+  if (!activeInbound?.address || activeInbound.address.toLowerCase() !== native.vaultAddress.toLowerCase()) {
     throw new NativeSwapBroadcastGuardError(
       'invalid_vault',
       `${native.chain} inbound vault address changed; refresh the swap quote before broadcasting`
--- a/packages/sdk/tests/unit/vault/services/nativeSwapBroadcastGuard.test.ts
+++ b/packages/sdk/tests/unit/vault/services/nativeSwapBroadcastGuard.test.ts
@@ -175,11 +175,11 @@ describe('assertNativeSwapReadyForBroadcast', () => {
     expect(getInboundAddresses).not.toHaveBeenCalled()
   })
 
+  // sdk#1726 changed the non-exempt observable: THOR-source routes no longer run the
+  // vault-address check (RUNE deposits have no inbound vault), so a lookalike payload that fails
+  // the secured-withdrawal detection is now caught by the expiry check its exemption would skip.
   it('does not exempt a secure-memo payload that carries an inbound vault', async () => {
-    const getInboundAddresses = vi.fn(async nativeChain => {
-      expect(nativeChain).toBe(Chain.THORChain)
-      return [makeInbound('THOR', 'thor1active')]
-    })
+    const getInboundAddresses = vi.fn()
 
     await expect(
       assertNativeSwapReadyForBroadcast({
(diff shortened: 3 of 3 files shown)


gomesalexandre (approved):
Reviewed this as a fund-safety **relaxation** rather than a fix, because that's what it is - it makes a guard stop blocking. The failure mode I was hunting is over-exemption: a `chain === native.chain` predicate that accidentally waves through routes it shouldn't. It doesn't. Both dangerous directions are pinned by tests. Approving.

## what I verified (receipts)

**must-pass:** `nativeSwapBroadcastGuard.test.ts` **29/29**.

**must-fail 1 - over-exemption.** Forced `const sourceIsProtocolChain = true` so every route gets exempted (mutation verified applied, pre-count 1 / post-count 1). **3 tests fail**, including `still validates the inbound vault when RUNE is the source of a MayaChain swap` - which is exactly the cross-protocol edge your comment calls out ("a RUNE source routed VIA Maya is NOT exempt: Maya's feed lists THOR as an external chain"). Nice to see the comment and the test agree; that edge is the one I'd have bet on being missed.

**must-fail 2 - ordering.** The body claims halt checks still run for exempt routes. They do: the halt loop is lines 145-165 and the `return` is at 167. I probed the regression where that `return` drifts above the loop, by injecting `if (sourceIsProtocolChain) return` right after the predicate. **2 tests fail** - `still rejects a RUNE-source swap when the destination chain is halted` and `still rejects a RUNE-source swap when global trading is paused`. So a future refactor that hoists the early return gets caught rather than silently letting a deposit land into a halted chain.

**merged-tree check.** The PR is `BEHIND` and its green CI ran on a stale base - main moved when #1588 landed. Merged `origin/main` locally: clean, 7 files / +87. Full sdk suite on the merged tree: **210 files, 3184/3184 passed**. Discarded the merge locally, your branch is untouched.

Also noting the defensive `!activeInbound?.address ||` you added to the staleness comparison - previously `activeInbound.address` could have thrown on undefined once the earlier existence check became conditional. Easy thing to miss when relaxing a guard; you didn't.

## nit: one test's name outlived its assertion

`does not exempt a secure-memo payload that carries an inbound vault` no longer tests exemption. With `getInboundAddresses = vi.fn()` and `expirationTime` dropped, it asserts `rejects.toThrow(/missing or invalid expiration/)` plus `expect(getInboundAddresses).not.toHaveBeenCalled()` - so "carries an inbound vault" is now vacuous, and the reason it rejects is the expiry check firing first, not anything about exemption.

I get why: THOR-source no longer runs the vault check, so the original assertion became unprovable. That's a real consequence of the change, not a dodge. But the name now points at behavior the test doesn't exercise, and the next person to touch this will trust it. Suggest renaming to something like `secure-memo payload without an expiration is rejected before the inbound lookup`, and - if the secured-withdrawal exemption is still worth pinning - a separate case asserting it against the checks that *do* still apply (the halt loop).

Non-blocking, and I'd rather you rename it than bolt on coverage for a path that may genuinely no longer exist.

`Fixes #1726` present in the body, so it'll auto-close on merge.

Rebase for the stale base and this is good to go from my side.

PR #1757, opened 2026-08-05
refactor(core): make chain guard maps exhaustive

rcoderdev:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction fee handling with explicit priority-fee limits across supported EVM chains.
  - Added comprehensive memo-size limits for supported Cosmos chains.
  - Clarified destination validation behavior across supported and unsupported swap chains.

- **Tests**
  - Updated fee-limit test wording to reflect the selected default ceiling.

- **Chores**
  - Recorded a patch release and documented requirements for complete per-chain configuration.

### DIFF
--- a/.changeset/exhaustive-chain-guard-maps.md
+++ b/.changeset/exhaustive-chain-guard-maps.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Make guard-shaped per-chain configuration tables exhaustive so new chains require explicit safety decisions.
--- a/packages/core/chain/chains/cosmos/cosmosMemoCap.ts
+++ b/packages/core/chain/chains/cosmos/cosmosMemoCap.ts
@@ -12,18 +12,25 @@ import { getCosmosChainByChainId } from './chainInfo'
 export const COSMOS_MEMO_DEFAULT_MAX_BYTES = 256
 
 /**
- * Chains whose gov-set `MaxMemoCharacters` differs from the cosmos-sdk default.
- * A chain absent here uses `COSMOS_MEMO_DEFAULT_MAX_BYTES` (256), not "no cap" -
- * that default closes the fail-open gap of an unmapped chain never getting
- * checked, which fails on virtually every cosmos chain in practice.
+ * The `MaxMemoCharacters` decision for every supported cosmos chain. Most use
+ * the cosmos-sdk default; keeping those entries explicit makes adding a chain
+ * fail typechecking until its cap has been consciously selected.
  */
-const COSMOS_MEMO_MAX_BYTES_OVERRIDES: Partial<Record<CosmosChain, number>> = {
+const COSMOS_MEMO_MAX_BYTES_BY_CHAIN: Record<CosmosChain, number> = {
+  [CosmosChain.Osmosis]: COSMOS_MEMO_DEFAULT_MAX_BYTES,
+  [CosmosChain.Dydx]: COSMOS_MEMO_DEFAULT_MAX_BYTES,
+  [CosmosChain.Kujira]: COSMOS_MEMO_DEFAULT_MAX_BYTES,
   // Terra v2 (phoenix-1): live-verified 512 (2026-06-22, /cosmos/auth params) -
   // Terra raised MaxMemoCharacters above the sdk default.
   [CosmosChain.Terra]: 512,
+  [CosmosChain.TerraClassic]: COSMOS_MEMO_DEFAULT_MAX_BYTES,
+  [CosmosChain.Noble]: COSMOS_MEMO_DEFAULT_MAX_BYTES,
+  [CosmosChain.Akash]: COSMOS_MEMO_DEFAULT_MAX_BYTES,
   // Cosmos Hub (cosmoshub-4): live-verified 512 (2026-06-22, /cosmos/auth params) -
   // gov-raised above the sdk default.
   [CosmosChain.Cosmos]: 512,
+  [CosmosChain.THORChain]: COSMOS_MEMO_DEFAULT_MAX_BYTES,
+  [CosmosChain.MayaChain]: COSMOS_MEMO_DEFAULT_MAX_BYTES,
 }
 
 /**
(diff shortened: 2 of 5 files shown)


gomesalexandre (approved):
Good refactor and the premise holds: these three maps are the kind where a silently-missing key is worse than a compile error, and turning `Partial<Record<...>>` into `Record<...>` makes adding a chain a decision instead of an omission. Approving. One should-fix that costs a single `??` and does not touch the exhaustiveness goal at all.

## what I ran

Controls first, because "behaviour preserving" is a claim and not an observation.

**must-fail control**, delete one entry from each of the three maps and confirm the compiler actually catches it. It does, on all three, including the `| undefined` variant where I half expected TS to let it slide:

```
clampEvmPriorityFee.ts(30,7): error TS2741: Property 'Sei' is missing in type
  '{ Arbitrum: bigint; ... }' but required in type 'Record<EvmChain, bigint>'
cosmosMemoCap.ts    TS2741  Property 'Kujira' is missing
limitSwapMemo.ts    TS2741  Property 'Cardano' is missing
```

So the PR's central claim is genuinely enforced, not just stylistically expressed. Files restored after, `git status` clean.

**must-pass control**: `tsc --project .config/tsconfig.json --noEmit` = 0. `vitest run --config .config/vitest.core.config.ts` = 240 files, 2457 passed, 1 skipped. All CI checks green.

**behaviour diff**: scripted `git show` of each map body on `origin/main` and on your head, parsed the `[Enum.Key]: value` pairs, and compared effective semantics treating a key absent from main as its old fallback value:

```
priorityFeeCeilingWeiByChain    no behavioural drift across 39 chains
COSMOS_MEMO_MAX_BYTES_BY_CHAIN  no behavioural drift across 14 chains
limitSwapDestinationValidators  no behavioural drift across 10 chains
```

Every previously-implicit chain got explicitly written with exactly the value the fallback was giving it. That is the boring result you want here.

Also chased two things that looked like holes and are not, noting them so nobody re-chases:

1. `limitSwapDestinationValidators` now carries explicit `undefined` entries, which reads like a fail-open. It is not: `assertValidLimitSwapDestinationAddress` does `if (!validator) throw` before it does anything else, so an unmapped chain is refused rather than waved through.
2. `getCosmosMemoMaxBytesByChainId` still guards explicitly (`chain !== undefined ? ... : COSMOS_MEMO_DEFAULT_MAX_BYTES`), so the chain-id entry point keeps its default. Only the `CosmosChain`-typed entry point lost one.

## should-fix: the type became exhaustive and the runtime lost its net at the same time

The two changes are independent, and only the first one is what the PR is for.

```ts
-const ceiling = priorityFeeCeilingWeiByChain[chain] ?? defaultPriorityFeeCeilingWei
+const ceiling = priorityFeeCeilingWeiByChain[chain]
 if (rpcPriorityFeeWei > ceiling) { ... }
```

If `chain` is ever off-enum at runtime, `ceiling` is `undefined`, and `bigint > undefined` is not an error, it is `false`:

```
[JS] 1000gwei > undefined  ->  false
[JS] 300 <= undefined      ->  false
```

So the guard does not throw, it silently stops guarding. Differential, same probe run on both sides against the real exported functions:

```
=== main ===
[C0] Sei mapped        CLAMPED to 500gwei          <- positive control
[C1] Arbitrum          CLAMPED to 50gwei           <- positive control
[P1] off-enum evm      clamped to 500000000000     (the default ceiling did its job)
[C2] Osmosis 300b      rejected (cap 256)          <- positive control
[C3] Osmosis 100b      allowed                     <- no over-block
[P2] off-enum cosmos   cap= 256 | 100-byte memo -> allowed

=== this PR ===
[C0] [C1] [C2] [C3]    identical
[P1] off-enum evm      NOT CLAMPED (returned the raw rpc value)
[P2] off-enum cosmos   cap= undefined | 100-byte memo -> REJECTED
```

Two different failure directions, both worth not having:

**EVM fails open.** `clampEvmPriorityFee` is documented in-tree as an attack guard, the comment at `getEvmFeeQuote.ts:168` calls it "the exact clamp-attack case" and cites NeOMakinG's blocking review on #1078. A compromised or broken RPC returning an absurd tip is precisely the input it exists for, and on an off-enum chain it now passes that tip straight through to signing.

**Cosmos fails closed**, which is safer but is a false-reject of legitimate money movement: every memo, however short, gets refused. `getCosmosMemoMaxBytes` and `isCosmosMemoWithinCap` are both public exports (`packages/sdk/src/index.ts:243-245` and the react-native entry), so this one is reachable by definition, an external consumer holds the argument.

Reachability inside the SDK, stated honestly rather than dressed up: I could not build an end-to-end payload that reaches the EVM one. `coin.chain` is unvalidated at runtime, `getKeysignCoin<EvmChain>` is a pure type assertion over `fromCommCoin`, which does `chain: coin.chain as T` on a protobuf string field with no enum check. But of the two call sites, `:174` is inside `if (chain === Chain.Zksync)` so it is narrowed, and `:199` goes through `getEvmMaxPriorityFeePerGas(chain)` first, which would fail on an RPC it cannot resolve. So call the EVM half defense in depth, not a live hole. The cosmos half is public API and needs no such argument.

**Action:** keep the exhaustive type, put the fallback back at the read site. These do not conflict, the `Record<EvmChain, bigint>` type is where 100% of this PR's value lives and it is completely unaffected:

```ts
const ceiling = priorityFeeCeilingWeiByChain[chain] ?? defaultPriorityFeeCeilingWei
```

```ts
export const getCosmosMemoMaxBytes = (chain: CosmosChain): number =>
  COSMOS_MEMO_MAX_BYTES_BY_CHAIN[chain] ?? COSMOS_MEMO_DEFAULT_MAX_BYTES
```

You still get TS2741 the moment someone adds a chain, which is the whole point, and a value that slipped past the type system still lands on a sane number instead of `undefined`. `getCosmosMemoMaxBytesByChainId` already does exactly this and the PR left it alone, so it is also the internally consistent shape.

If you would rather keep the reads bare, that is a defensible position too, it just wants to be the deliberate version of itself: throw on the miss rather than compare against `undefined`. What I would not leave is the silent one.

## what I did NOT verify

No downstream wiring, no `yarn pack` into mcp-ts or vultiagent-app. Everything above is the core package on your head plus the same probe run against main's copies of the three files swapped underneath it, with mapped-chain controls proving the harness sees live code both times. No runtime changes reachable from a sim drive here.

Approving as-is since nothing regresses for any chain that actually exists in the enums, but the `??` is free and I would take it before merge.

PR #1758, opened 2026-08-05
fix(import): reject duplicate or stale vault shares

### LINKED ISSUE #1207: fix(import): reject duplicate or stale local vault shares
## Problem

Vault import can overwrite an existing local record for the same logical vault without proving the imported share is the current device's compatible share. A stale or other-device file can silently replace usable local key material.

Evidence: `packages/sdk/src/VaultManager.ts:123-218`.

## Acceptance criteria

- Detect an existing logical vault before persistence.
- Compare vault identity, share metadata, device/party identity, and key domains.
- Require an explicit safe merge/replace decision; never overwrite silently.
- Keep the original record intact on validation or persistence failure.
- Add duplicate, stale-share, other-device, and retry tests.

rcoderdev:
Closes #1207

## Summary by CodeRabbit

* **New Features**
  * Added explicit import conflict handling, including optional replacement of compatible existing vaults.
  * Added clear errors for duplicate, stale, incompatible, and other-device shares.
  * Added atomic storage updates to prevent conflicting changes across supported platforms.

* **Bug Fixes**
  * Imports can recover from unreadable records when replacement is explicitly requested.
  * Unencrypted imports now clear cached passwords appropriately.

* **Documentation**
  * Updated import guidance to explain explicit replacement requirements.


gomesalexandre (dismissed):
The threat model here is right and I want this to land: a stale share silently overwriting a fresh one is a real way to lose funds, and the CAS-across-every-backend approach is the correct shape for it. Two things block it though, and one of them is why CI is red.

## what I ran first

Before trusting anything below, controls, because a probe that cannot see the new code proves nothing:

```
[C0] fresh import                 -> id=0211111111  persisted=true          <- positive control, no over-block
[C1] second import of same share  -> code=DUPLICATE_VAULT                   <- must-fail control: this code exists ONLY on your branch
[C2] identical share + replace    -> OK                                     <- the legit replace path is not over-blocked
```

C1 is the important one. `DUPLICATE_VAULT` does not exist on `main`, so the harness is demonstrably reading your code and not a stale build. Every differential below is the same probe run twice, once on your head and once with `git checkout origin/main -- packages/sdk/src` swapped underneath it.

`npx tsc --noEmit` clean. `npx vitest run tests/unit` on your head: **17 failed / 3185 passed**. 16 of those 17 are `ensureMpcEngine` + `platforms/react-native/entry` blowing up on `configureMpc: duplicate MPC engine instance`, and they fail identically with main's `src/` swapped in, so that is my local environment and not yours. The 17th is yours, and it is finding 1.

## preferably-blocking: an `activeVaultId` failure now deletes the freshly migrated vault

`tests/unit/vault/legacy-vault-backup-migration.test.ts` is **not touched by this PR** and it is now red. It is also the reason Unit Tests and Test Summary are failing in CI, on a branch whose base is green.

The test name states the contract out loud: `emits the migration notice after save even when setting the active vault fails`. Differential, same probe, same synthetic `activeVaultId` write failure:

```
=== main ===
[M1] code= CORRUPTED_DATA       | msg= Failed to import vault: synthetic active pointer failure
[M2] migrated record survived= true   | notice fired= 1

=== this PR ===
[M1] code= PERSISTENCE_FAILED   | msg= Failed to persist imported vault. The previous local state was restored.
[M2] migrated record survived= false  | notice fired= 0
```

The cause is that the rollback is scoped one line too wide:

```ts
try {
  vaultPersisted = await compareAndSet(vaultKey, existing, vault.data)
  if (!vaultPersisted) { throw new VaultImportError(PERSISTENCE_FAILED, '...') }
  await this.storage.set('activeVaultId', vault.id)     // <- different failure domain
} catch (error) {
  ...
  const rolledBack = await compareAndSet(vaultKey, vault.data, existing)
```

Two writes with very different recovery costs are inside one `try`, so a failure of the cheap one undoes the expensive one. The active pointer is trivially recoverable, the user picks a vault and it is set. The vault record is not: on the legacy path `legacyBackupMigrated` has already re-encrypted the backup with salted 600k PBKDF2 via `encryptVaultBackupWithPassword`, and that is the artifact getting thrown away. The notice that would have told the user their derivation changed is at `VaultManager.ts:476`, downstream of the throw, so it never fires either.

Not permanent loss of key material, the user still holds the `.vult`, so I am not calling this a fund-safety hole. But it is a durable save being discarded to undo a pointer write, and the message says `The previous local state was restored` while that happens, which for a first-ever import means it restored the absence of a vault.

**Action:** keep the rollback on the vault write only, and let a pointer failure surface without touching what already committed.

```ts
vaultPersisted = await compareAndSet(vaultKey, existing, vault.data)
if (!vaultPersisted) {
  throw new VaultImportError(PERSISTENCE_FAILED, 'The local vault changed during import; no data was overwritten. Retry against the current record.')
}
try {
  await this.storage.set('activeVaultId', vault.id)
} catch (pointerError) {
  // the share is durably saved and the pointer is recoverable, do not undo the write
  throw new VaultImportError(PERSISTENCE_FAILED, 'The vault was saved but could not be made active.', pointerError as Error)
}
```

Changing the code from `CORRUPTED_DATA` to `PERSISTENCE_FAILED` is fine and arguably better, so updating that assertion in the test is right. The other two assertions in it are the contract and should stay green as-is: the record survives, and the notice fires once.

## preferably-blocking: a locally damaged record becomes a permanent import lockout, and `replace` cannot rescue it

```ts
const existingVault = await this.storage.get<VaultData>(`vault:${vaultId}`)
if (existingVault) {
  const existingCoreVault = await this.decodeStoredVault(existingVault, password)   // <- throws here
  this.validateReplacement(existingCoreVault, parsedVault)

  if ((options.conflictResolution ?? 'reject') !== 'replace') {                     // <- never reached
```

`decodeStoredVault` runs before `conflictResolution` is ever read, and its terminus is a hard throw. So the explicit escape hatch this PR introduces is unreachable in exactly the case a user would reach for it. Two variants, both measured on your head, both against a **valid backup of the user's own vault**:

```
[P1]  local vultFileContent corrupted   default= INCOMPATIBLE_VAULT             replace= INCOMPATIBLE_VAULT
[P1b] stored encrypted, cold cache,     code= EXISTING_VAULT_PASSWORD_REQUIRED
      backup password since rotated
```

`main`, same probe, same fixtures:

```
[P1]  default= (imports fine)   replace= RECOVERED
[P1b] RECOVERED
```

So this is introduced here, not pre-existing. P1 is a partially-written or half-migrated `vultFileContent`, which is reachable precisely because storage writes were not atomic before this PR. P1b is a user who rotated their backup password, has the new `.vult` in hand, and a cold `passwordCache`: `onPasswordRequired` cannot help them because the password that unlocks the *stored* record is the old one they no longer have. Both end with the SDK permanently refusing the user's own valid backup, and the only way out is clearing storage by hand.

**Action, minimal:** validate when the existing record is readable, and let an explicit `replace` through when it is not. Nothing is being protected at that point, the local record is already unreadable:

```ts
if (existingVault) {
  const wantsReplace = (options.conflictResolution ?? 'reject') === 'replace'
  let existingCoreVault: Vault | null = null
  try {
    existingCoreVault = await this.decodeStoredVault(existingVault, password)
  } catch (decodeError) {
    if (!wantsReplace) throw decodeError
    // unreadable local record + explicit opt-in: there is nothing left to validate against
  }
  if (existingCoreVault) {
    this.validateReplacement(existingCoreVault, parsedVault)
    if (!wantsReplace) { throw new VaultImportError(DUPLICATE_VAULT, '...') }
  }
}
```

Flagging the tradeoff honestly rather than pretending it is free: for P1b specifically that lets an explicit `replace` overwrite a locked encrypted record without proving the shares match, which is a small step back from the other-device guard. If you would rather not take that, fine, but then P1b needs *some* deliberate exit (a distinct `conflictResolution` value, or an error the app can act on) instead of a dead end. Your call on which, it just should not stay unreachable.

## should-fix: `compareAndSet` is a required member of an exported type on a `patch` bump

`.changeset/safe-vault-share-imports.md` says `'@vultisig/sdk': patch`, and its own last sentence says why that is wrong:

> Custom storage adapters must expose `compareAndSet` for vault imports.

```ts
// packages/sdk/src/storage/types.ts
compareAndSet<T>(key: string, expectedValue: T | null, value: T | null): Promise<boolean>
```

Required, not optional, on the publicly exported `Storage` type. Any consumer with a custom adapter stops typechecking on a version bump they have every reason to take without reading. The runtime side is handled well, `persistImportedVault` fail-closes with a clear message when the method is absent, so this is purely the semver label. `minor` at minimum.

## should-fix: an unencrypted backup can still replace an encrypted stored record

Calling this out as a gap in the new guard rather than a regression, because it behaves identically on both sides:

```
[P2] this PR   before.isEncrypted= true | after.isEncrypted= false | threw= no
[P2] main      before.isEncrypted= true | after.isEncrypted= false | threw= no
```

`validateReplacement` is thorough on identity, `localPartyId`, both public keys, `publicKeyMldsa`, `chainPublicKeys`, `signers`, `hexChainCode`, `libType`, `resharePrefix`, `createdAt`, every keyshare. It never looks at the container's encryption state, so a plaintext export of the same vault silently downgrades an encrypted local record, and the new `passwordCache.delete(vaultId)` branch at `VaultManager.ts:472` cleanly finishes the job.

Not introduced here and not a blocker for this PR. But this PR is the one that establishes "we validate the replacement before letting it land", and encryption posture is the one property in that list a user would notice losing. Worth a follow-up, happy to file it if you would rather not stack.

## nit: the rollback message reports a failure that did not happen

When the first CAS *throws* rather than returning false, `vaultPersisted` is still `false`, nothing was written, and the rollback CAS then compares against a value that was never stored. It returns false by construction:

```
[P3] code= PERSISTENCE_FAILED
     msg= Failed to persist imported vault. Rollback also failed: the vault changed again before rollback
```

Nothing needed rolling back. Telling someone rollback failed in the middle of a key material import is the wrong thing to be wrong about. Guard the rollback on `vaultPersisted` being true.

## what I did NOT verify

No downstream wiring. This is unit-level against the real `VaultManager` and real `MemoryStorage` on your head, with a positive control proving the harness sees your code, but I did not `yarn pack` this into mcp-ts or vultiagent-app to exercise a real adapter. The four new backend `compareAndSet` implementations I only read, your own new storage tests cover them and they pass.

Fix the rollback scope and CI goes green, sort the lockout and this is an approve. The core idea is one the SDK has needed for a while.

gomesalexandre (commented):
Remote-only re-review on current head `301d1995c`.

What I re-checked from the current diff and current-head CI:
- the two takeover blockers I previously raised are closed on this head: rollback is now scoped so an `activeVaultId` write failure no longer discards a successfully persisted vault record, and explicit `conflictResolution: 'replace'` now has a live escape hatch when the local stored record is unreadable instead of dead-ending before the option is consulted
- the current branch also carries the follow-up hardening I wanted on the storage side: stale lock reclamation is fenced, Windows key-path handling is covered, and the added Node storage platform job is green on both macOS and Windows
- current head checks are fully green: Unit Tests, Integration Tests, Node Storage (macOS), Node Storage (Windows), Lint and Type Check, and Test Summary

I do not see a fresh material correctness blocker on the current head from this remote pass.

Important four-eyes note: this PR head now contains substantive `gomesalexandre` commits, so I cannot approve it and it must receive a non-gomes current-head approval before any merge.

packages/sdk/src/VaultManager.ts
```diff
@@ -233,29 +314,138 @@ export class VaultManager {
...
+    password: string | undefined,
+    wantsReplace: boolean
+  ): Promise<void> {
+    if (!existingVault) return
+
+    let existingCoreVault: CoreVault | null = null
+    try {
+      existingCoreVault = await this.decodeStoredVault(existingVault, password)
+    } catch (decodeError) {
+      if (!wantsReplace) throw decodeError
```
Ehsan-saradar: When `wantsReplace` is true this swallows every `decodeStoredVault` failure — including `EXISTING_VAULT_PASSWORD_REQUIRED`, which proves the stored record is well-formed and merely locked (the container parsed; we just couldn't decrypt it). `existingCoreVault` stays null, `validateReplacement` is skipped, and the other-device/stale-share checks never run: a replace import whose password can't decrypt the *existing* share (cold cache, prompt cancelled or absent) can overwrite the only local copy of a 2-of-2 share with a different device's share of the same vault — the exact loss scenario this PR exists to prevent.

I get that this is intentional for the rotated-password recovery case the changeset describes, but as written `replace` silently degrades from 'validated replacement' to 'unvalidated overwrite' depending on whether decryption happened to succeed. Suggest keeping plain `replace` fail-closed at least for `EXISTING_VAULT_PASSWORD_REQUIRED`, and gating the unvalidated path behind an explicit opt-in (e.g. `conflictResolution: 'replace-unvalidated'` or an `allowUnreadableExisting: true` flag) so callers consciously accept skipping the share checks when the existing record can't be read.

packages/sdk/src/VaultManager.ts
```diff
@@ -265,29 +455,52 @@ export class VaultManager {
...
         )
       }
 
-      // Cache password if provided (for encrypted vaults)
+      // Change password state only after the vault record commits. A pointer failure below
+      // leaves the durable record in place, so the cache should already describe that record.
       if (password && container.isEncrypted) {
         this.context.passwordCache.set(vaultId, password)
+      } else if (!container.isEncrypted) {
+        this.context.passwordCache.delete(vaultId)
```
Ehsan-saradar, line 494: This eviction is new behavior — import previously only ever *added* cache entries. Importing an unencrypted backup of an already-known vault (e.g. refreshing a fast vault with `replace`) now silently wipes a password cached at create/unlock time, so the next `signWithServer` in a headless consumer (no `onPasswordRequired`) throws `Password required but no callback provided` where the old flow kept signing.

Nothing else in the PR depends on this branch — suggest dropping the `delete`, or scoping it to the case where the durable record's encryption state actually changed.

packages/sdk/src/platforms/chrome-extension/storage.ts
```diff
@@ -10,6 +10,30 @@ import type { Storage, StorageMetadata, StoredValue } from '../../storage/types'
...
+    if (typeof navigator === 'undefined' || !navigator.locks) {
+      if (required) {
+        throw new StorageError(
+          StorageErrorCode.StorageUnavailable,
+          'Atomic storage mutations require the Web Locks API in Chrome extension contexts'
+        )
+      }
+      return operation()
+    }
+    return navigator.locks.request('vultisig-extension-storage', operation)
```
Ehsan-saradar, line 32: Web Locks are origin-partitioned, but `chrome.storage.local` is extension-global. A content script's `navigator.locks` lives in the host page's origin partition, so this lock does not exclude the popup or service worker (nor a content script on another site): two contexts can each acquire their own partition's lock concurrently, both pass the compare, and both write — a lost update on the same `vault:` record, in exactly the multi-context case CAS is meant to protect (and the file header advertises content-script support). `required = true` doesn't fail closed here either: `navigator.locks` *exists* in the content script, it's just scoped to the wrong partition.

Cheap fix: treat non-`chrome-extension://` origins as lock-less for mutations (fail closed for CAS, same stance as the missing-API branch above). Real fix if content-script mutations actually matter: proxy writes to the service worker via `chrome.runtime.sendMessage` so all mutations run in one context.

packages/sdk/src/platforms/browser/storage.ts
```diff
@@ -254,14 +267,48 @@ export class BrowserStorage implements Storage {
...
 
+  async compareAndSet<T>(key: string, expectedValue: T | null, value: T | null): Promise<boolean> {
+    await this.ensureInitialized()
+
+    try {
+      if (this.mode === 'indexeddb' && this.db) {
+        return await this.compareAndSetIndexedDB(key, expectedValue, value)
+      }
+
+      return await this.withLocalStorageMutationLock(() => {
```
Ehsan-saradar, line 295: Passing `required = true` here, combined with `compareAndSet` always being defined on the instance, means every `VaultBase.save()` — rename, currency, token/chain toggles, background balance saves, not just imports — now hard-fails with `StorageUnavailable` in localStorage mode when `navigator.locks` is missing (Web Locks is secure-context-only, so http origins and some embedded WebViews). Those saves worked before this PR via plain `set()` (which degrades gracefully with `required = false`), and the sticky `localstorage` backend marker means affected deployments can't escape by switching backends. The regression's blast radius is much larger than the import feature the fail-closed check was designed for.

Suggestion: don't expose `compareAndSet` when the environment can't honor it — e.g. delete the instance method during init when mode is `localstorage` and Web Locks is absent. `VaultBase.save()` then falls back to plain `set()` as before, while `importVault`'s existence check still refuses imports on such environments, which is the intended gate.

packages/sdk/src/platforms/node/fileLock.ts
```diff
@@ -0,0 +1,114 @@
...
+    unlock(fd) {
+      const handle = getValidatedOsFileHandle(fd)
+      if (!unlockFile(handle, 0, 1, 0, newOverlapped())) {
+        throw new Error(`Failed to unlock file descriptor ${fd} (Windows error ${getLastError()})`)
+      }
+    },
+  }
+}
+
+const nativeFileLock = process.platform === 'win32' ? createWindowsFileLock() : createPosixFileLock()
```
Ehsan-saradar: This runs `koffi.load` plus all the FFI bindings eagerly at module scope, and the import chain `platforms/node/index.ts` → `storage.ts` → here is unconditional — so `import '@vultisig/sdk'` on Node now fails outright wherever the native addon can't load (koffi 3.1.4 ships no armv7 prebuild, for example, and some runtimes block native addons), even for consumers passing a custom storage adapter that never takes a lock. FileStorage was dependency-free pure JS before this.

Lazy init keeps the failure at first lock use, where it belongs:

```ts
let native: ReturnType<typeof createPosixFileLock> | undefined
const getNative = () =>
  (native ??= process.platform === 'win32' ? createWindowsFileLock() : createPosixFileLock())
```

.changeset/safe-vault-share-imports.md
```diff
@@ -0,0 +1,7 @@
+---
+'@vultisig/sdk': minor
```
Ehsan-saradar: This should be `major`. The text below documents caller-breaking changes: existing `importVault` callers that relied on idempotent re-import now throw `DUPLICATE_VAULT` unless they pass `{ conflictResolution: 'replace' }`, and custom storage adapters without `compareAndSet` lose vault import entirely. On a `^4.x` range this ships silently as 4.4.0 to exactly the integrations it breaks.

Ehsan-saradar (dismissed):
Reviewed with a focus on the two guarantees this PR introduces — validated replaces and atomic CAS-backed persistence — plus the downstream impact of the new duplicate-import default. The direction is right and the coverage is genuinely good (the FFI layer, the IndexedDB transaction semantics, and the plaintext zeroing all check out), but each of the two guarantees currently has a concrete counterexample (see inline), and the new `reject` default breaks several in-repo consumers that were written against the old idempotent upsert.

**Consumers that break with the new `reject` default** (their files aren't in this diff, so noting them here — I'd consider these part of the same change):

- `clients/mcp/bin/mcp-server.ts:57` — `--vault-file` re-imports the same .vult on every server start over persistent FileStorage. The second start now throws `DUPLICATE_VAULT` and the server never boots; a restart-on-crash setup loops forever. Needs a `DUPLICATE_VAULT` catch that reuses the stored vault, or `{ conflictResolution: 'replace' }`.
- `packages/client-shared/src/auth-setup.ts:76,103` — re-running setup for an already-imported vault now throws `DUPLICATE_VAULT`, which the bare catches misreport as a wrong password: the user types the *correct* password, gets 'Wrong password. Try again' three times, and setup aborts with no recovery path short of deleting `~/.vultisig`. The env path similarly reports a correct `VAULT_DECRYPT_PASSWORD` as failing to decrypt. Check `error.code` before the wrong-password branch.
- `clients/cli/src/commands/vault-management.ts:313` (and interactive `import`) — re-importing surfaces the raw SDK message telling the user to pass `conflictResolution: 'replace'`, a TypeScript option the CLI has no flag for. Needs a `--replace` flag plus friendly mapping for the new error codes.
- Example apps (`examples/electron/electron/ipc-handlers.ts:125`, `examples/shared/.../VaultImporter.tsx:120`) — dropping a previously-imported .vult now shows the raw error verbatim, and `describeVaultImportFailure` has no cases for the new codes (the electron IPC signature also can't thread options through).

**Non-blocking follow-ups worth tracking** (happy to split into issues):

- A torn write leaving invalid JSON — the realistic pre-PR corruption — throws from `storage.get()` at VaultManager.ts:431 before `conflictResolution` is consulted, so the `replace` escape hatch the changeset advertises is unreachable for exactly that corruption shape. (The new 'corrupted' test fixture is valid JSON with garbage `vultFileContent`, the one shape that does reach it.) Wrapping the read and treating an unreadable record as replaceable under `replace` would close the gap.
- A failure writing `activeVaultId` after the vault record durably committed is reported as `CORRUPTED_DATA`; `PERSISTENCE_FAILED` is the accurate code, and the new tests currently pin the wrong one.
- Asymmetric enforcement: import hard-fails without `compareAndSet`, but `VaultBase.saveUnlocked` still silently falls back to plain `set()` for every other mutation of the same record — the advertised guarantee only exists on the import edge. (The capability check also runs after `validateImportConflict` may already have prompted for a password.)
- The CAS equality contract is unspecified in `storage/types.ts` but implemented six times as `JSON.stringify` equality. Fine for the built-ins (the expected value always derives from the same adapter's own `get()`), but a third-party adapter backed by anything that re-serializes (SQLite jsonb, a key-order-normalizing KV) fails every save with zero concurrency — the vault becomes permanently unsaveable. Comparing the existing monotonic `revision` would be immune to that and cheaper than double-stringifying multi-hundred-KB records per save.
- Docs: the SDK-USERS-GUIDE Storage contract and CustomStorage example lack `compareAndSet`, and the guide/README still document 2-arg `importVault` — the documented adapter contract and the runtime requirement disagree.
- The react-native mutex is per-JS-bundle (a second runtime or a duplicated module copy bypasses it), and a `BrowserStorage` instance in a worker can't see the localStorage backend marker (no `localStorage` in workers), so a worker and a tab can settle on different stores for the same origin.
- Perf: a duplicate import runs PBKDF2-600k up to three times when `persistedVultContent === existing.vultFileContent` already proves the outcome byte-for-byte — a string-equality fast path in `validateImportConflict` would skip ~1s of blocking CPU on the common accidental re-import.
- Dedup: the CAS body is copied across the 5 adapters + MemoryStorage, `StoredValue` metadata construction across 7 sites (MemoryStorage already diverges — it preserves `createdAt`, the others stamp a fresh one), the container→decrypt→parse pipeline gains its 5th copy in `decodeVaultPayload`, and the mutex/web-lock wrappers exist in 5 near-identical variants including the two test stubs.
- CI: the new macOS/Windows job does an uncached full monorepo install with no `paths` filter to run one test file (macOS bills at 10x).

gomesalexandre (commented):
Did the technical verification requested, pushed a fix for real breakage found, and I'm posting this as a COMMENT (not an approve) - this branch already carries a gomesalexandre commit (03e7a9e97, from an earlier round), so per the four-eyes rule this needs an approval from someone other than gomesalexandre before merge regardless of how solid the code is.

## What I verified from Ehsan-saradar's CHANGES_REQUESTED (2026-08-09)

All 4 flagged downstream consumers were still genuinely broken on the pre-fix head - confirmed by reading each call site, not just taking the review's word for it:

- `clients/mcp/bin/mcp-server.ts:57` - `--vault-file` re-imports the same .vult on every server start; second start threw `DUPLICATE_VAULT`, server never booted.
- `packages/client-shared/src/auth-setup.ts` (3 call sites) - bare `catch {}` blocks meant a correct password against an already-imported vault got reported as "Wrong password. Try again" for 3 attempts, then aborted with no recovery.
- `clients/cli/src/commands/vault-management.ts:313` - re-importing surfaced the raw SDK message telling the user to pass a TS-only option the CLI has no flag for.
- `examples/electron/electron/ipc-handlers.ts:125` + `examples/browser/src/adapters/BrowserSDKAdapter.ts:199` - same raw-error re-import UX, `describeVaultImportFailure` had no cases for the new error codes.

**Fix pushed (04623ea0c):** `conflictResolution: 'replace'` at all 5 call sites, restoring idempotent re-import. This is safe because `validateReplacement` runs unconditionally before `conflictResolution` is even consulted (see `VaultManager.ts` `validateImportConflict`) - a genuinely different or stale share still throws `OTHER_DEVICE_SHARE` / `INCOMPATIBLE_VAULT` / `STALE_SHARE` regardless of `replace`. Also fixed `auth-setup.ts`'s error handling to only treat `INVALID_PASSWORD`/`PASSWORD_REQUIRED` as "wrong password" instead of bare-catching every `VaultImportError`, and added friendly messages for the new codes in the electron example's `describeVaultImportFailure`.

Full repo `yarn typecheck`: clean, 0 errors (built `client-shared`/`rujira` locally first since they weren't pre-built in this worktree). Lint on all 7 touched files: clean. Full `yarn test` at repo root: SDK 3295, Rujira 550+4skip, Core 2612+1skip, CLI 1067, MCP 35 - all passing, run twice for stability. A handful of CLI subprocess-spawn tests (`stderrCleanliness`, `chainArgCallSites`, `sendSwapDupGuard`) flaked with 5s/30s timeouts on the second run under heavy concurrent load from other agents' builds on this machine; confirmed unrelated to this change by re-running each file in isolation (31/31 pass every time) - none touch the import path.

## Duplicate/stale share detection: adversarial verification

Read `validateReplacement` in `VaultManager.ts` end to end. Confirmed the existing test suite already covers: exact duplicate (reject by default), stale share via different key-share bytes (rejected even with `replace`), other-device share via different `localPartyId` (rejected even with `replace`), and key-domain mismatch (rejected even with `replace`).

Found one genuinely untested dimension: `resharePrefix` (the field that changes on a TSS resharing round while the wallet's public keys stay constant) wasn't exercised by any test - the test fixture builder didn't even support overriding it. Added a regression test: same byte-identical key-share material as the stored record, differing *only* in `resharePrefix` (simulating a backup from before a reshare), and confirmed it's correctly rejected as `STALE_SHARE` even under explicit `replace`. Proved this test is load-bearing (not trivially passing) by temporarily neutralizing the `resharePrefix` equality check in `validateReplacement`, confirming the new test then fails, and reverting.

Non-blocking observation for a follow-up, not a blocker: `chainPublicKeys`/`chainKeyShares` (second-class-chain-specific key material) is structurally checked the same way (`recordsEqual`) but never populated by any test fixture in this file, so that dimension is untested (not known-broken, just unverified).

## Bottom line

The core duplicate/stale-share guarantee holds under adversarial testing including a new case. The downstream breakage Ehsan flagged is fixed with a receipt-backed, minimal change (5 call sites + 1 error-classification fix), not a redesign. Needs a non-gomes approval before merge.

gomesalexandre (commented):
Re-probed on `3ee6255f`. Ten commits since my last pass, and the substantive new surface is the storage locking work: `fence stale lock reclamation`, `restore atomic vault persistence`, `support Windows storage keys`.

## the Windows FFI path is genuinely tested, which is the right call

The new `test-node-storage-platforms` matrix runs the concurrency and crash-recovery tests on `macos-latest` AND `windows-latest`, with the explicit `--config packages/sdk/tests/unit/vitest.config.ts`. Both green on this head. For a native FFI path that cannot be exercised on the Linux runner, adding the matrix rather than hoping is the correct move, and using `flock` over an `O_EXCL` lockfile is also right: flock releases on process death, a lockfile does not, which is the stale-lock problem you would otherwise have to invent a reaper for.

## a bug I thought I had, and did not

`getValidatedOsFileHandle` compares `koffi.address(handle) === INVALID_HANDLE_VALUE`, where the constant is built as `(1n << BigInt(koffi.sizeof('void *') * 8)) - 1n`, a BigInt. In JS `1 === 1n` is false, so if `koffi.address()` returned a Number the guard would never fire. It does not: koffi 3.1.5's typings declare `export function address(value: any): bigint`. BigInt on both sides, comparison is sound. Recording it because it reads like a type-mismatch bug and the next reviewer will have the same reaction.

## should-fix: a koffi load failure escapes as a raw Error, not a StorageError

`getNativeFileLock()` calls `koffi.load(...)` lazily, and any failure there (unsupported platform or arch, FFI blocked by a hardened runtime, prebuild missing) propagates straight out through `tryLockFile` at `storage.ts:142`. The surrounding `try` only has a `finally`, so nothing converts it. Consumers that switch on `StorageErrorCode` get an unmodelled exception instead of `StorageErrorCode.StorageUnavailable`, which is exactly the code that already exists for "storage cannot be used right now".

Cheap to close: wrap the load in a try and rethrow as `new StorageError(StorageErrorCode.StorageUnavailable, ...)`. That also makes the failure legible to the CLI and MCP consumers rather than surfacing a koffi internal.

## q: koffi as a hard `dependencies` entry

It is `dependencies` on `@vultisig/sdk`, not `optionalDependencies`, so every consumer installs a native FFI package with prebuilt binaries, including the React Native app and browser consumers that can never reach `platforms/node/fileLock.ts`. Bundlers will tree-shake it out of those builds since only `node/storage.ts` imports it, so this is an install-surface question rather than a bundle-size one: an npm install on a platform koffi has no prebuild for now fails for consumers who were never going to call it.

Not asking you to rip it out, the locking rationale above is sound. Asking whether `optionalDependencies` plus the StorageError degradation above is the better shape, so a platform without koffi loses file locking with a clear error instead of failing to install.

## note on this PR's green CI

Worth knowing before anyone reads the all-green as ready: the `Lint and Type Check` success here completed `2026-08-12T20:40Z`, and GHSA-2v37-7h3g-55p8 (nanoid) had its range widened to `< 3.3.18` on `2026-08-13T15:43Z`. This branch's `yarn.lock` nanoid entry and root `resolutions` are byte-identical to main's, so the green is stale rather than earned. Re-running CI today will very likely turn this red at `quality:audit` through no fault of the branch. Tracked in #1916.

Not approving: this head carries my own `a8ceb9c3`, so a stamp from me would not be four eyes. The storage work looks solid to me otherwise.

gomesalexandre:
Re-verified the changes-requested review point-by-point against current head (not just commit messages):

1. **Downstream consumers broken by `reject`-by-default default** — RESOLVED. All 5 cited call sites (`mcp-server.ts`, `auth-setup.ts`, `vault-management.ts` + `--replace` flag, both example apps) now pass/thread `conflictResolution: 'replace'` and correctly branch on `error.code`.
2. **Torn-write corruption makes `replace` unreachable** — STILL OPEN. `VaultManager.ts:433`'s `storage.get()` call is still unguarded; a `FileStorage.get()` throw on unparseable JSON (the reviewer's literal example) still falls through to generic `CORRUPTED_DATA` before `conflictResolution` is consulted. The fix commits added a narrower `replace-unvalidated` mode for a different corruption shape (valid envelope, garbage inner content) — real progress, but doesn't cover the case the review actually flagged. Needs a decision on what `existingVault` becomes when the read itself fails.
3. **`activeVaultId` write failure misreported as `CORRUPTED_DATA`** — fixed just now (`e22761051`): wrapped the write in try/catch → `PERSISTENCE_FAILED`, mirroring the existing pattern a few lines above for `vaultInstance.save()`, and corrected the 3 tests that pinned the wrong code.
4. **Asymmetric CAS enforcement** (import hard-fails without `compareAndSet`, `VaultBase.saveUnlocked` silently degrades) — STILL OPEN, `VaultBase.ts` untouched by either fix commit. Architecture question, not a mechanical fix.
5. **CAS equality contract undocumented/inconsistent** (`JSON.stringify` equality duplicated across 6 adapters) — STILL OPEN, `storage/types.ts` untouched. Also a design call (e.g. switch to `revision`-based comparison).

Full SDK suite (3360 tests) + `yarn check` clean after point 3's fix. Points 2, 4, 5 need an explicit call before this can honestly claim the review is closed — flagging rather than guessing at the right answer on a fund-safety-adjacent storage/import path.

gomesalexandre (commented):
Re-reviewed current head `e2276105`.

I re-checked the earlier downstream breakage and the follow-up storage hardening on this head:
- the idempotent re-import fixes are in place at the MCP/CLI/example/auth-setup call sites, so duplicate local imports no longer dead-end those consumers on the new default `reject` path
- the active-pointer write failure is now classified as `PERSISTENCE_FAILED` instead of being misreported as corruption
- the node-storage locking work now has explicit macOS + Windows coverage, which is the right receipt shape for this FFI path

Not approving: this PR head contains substantive `gomesalexandre` commits (`a8ceb9c3`, `e2276105`), so it needs a non-gomes current-head approval before merge.

rcoderdev:
All three remaining points from the current-head review are now addressed on 9a3c2684:

1. A failing storage read now reports PERSISTENCE_FAILED and performs no compare-and-set or write, including in replace-unvalidated mode. An unreadable current value is never fabricated into a CAS baseline.
2. Atomic compare-and-set is now the import boundary: imports fail closed when a backend cannot provide it, while ordinary non-import saves retain their legacy fallback for custom adapters. Every built-in backend used for import has an atomic implementation.
3. The equality contract is centralized and documented. All adapters use the same JSON-value comparison instead of duplicating JSON.stringify behavior.

The follow-up hardening also closes the platform races found during re-review: Chrome content-script mutations fail closed outside the extension origin, FileStorage uses crash-recoverable OS locks, and React Native uses transactional Expo SQLite with a retryable two-phase migration that removes only migrated Vultisig keys from AsyncStorage.

The branch is rebased onto current main. The full repository CI-equivalent gate passed on this exact head, as did 190 retry-disabled focused tests and the import runtime checks. The only remaining validation gap is a real on-device Expo SQLite / live extension-host run; the repository has no such harness, and the corresponding host-mocked suites are green.

@gomesalexandre @Ehsan-saradar, please re-review the current head when GitHub checks finish.

@coderabbitai review

gomesalexandre (approved):
Taking over doesn't break four-eyes - PR authored by @rcoderdev, I added commits on top while shepherding it through. Approving per gomes.

PR #1762, opened 2026-08-05
fix(sdk): align Solana duplicate broadcast semantics

### LINKED ISSUE #1227: refactor: per-chain tx logic spans three trees with diverged broadcast guarantees (rn solana is the weakest)

rcoderdev:
## Summary

- Treat Solana `AlreadyProcessed` responses as broadcast-layer acceptance in `RawBroadcastService`, matching the core and React Native resolvers.
- Leave execution success or failure to the transaction-status layer.
- Add regression coverage and a patch changeset.

Follow-up to #1754.

## Testing

- `yarn vitest run --config tests/unit/vitest.config.ts tests/unit/vault/services/RawBroadcastService.test.ts tests/unit/platforms/react-native/solana-rpc.test.ts`
- `yarn format:check`
- `yarn changeset status --since origin/main`

Closes #1227

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana transaction broadcasting to handle duplicate submissions more reliably.
  * Returned transaction signatures when duplicate transactions are pending or their status cannot yet be confirmed.
  * Continued rejecting transactions explicitly reported as failed on-chain.

### DIFF
--- a/.changeset/quiet-rivers-accept-solana-duplicates.md
+++ b/.changeset/quiet-rivers-accept-solana-duplicates.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Align raw Solana duplicate-broadcast handling with the core and React Native resolvers by treating `AlreadyProcessed` as broadcast acceptance and leaving execution outcome checks to transaction status resolution.
--- a/.changeset/quiet-rivers-accept-solana-duplicates.md
+++ b/.changeset/quiet-rivers-accept-solana-duplicates.md
@@ -2,4 +2,4 @@
 '@vultisig/sdk': patch
 ---
 
-Align raw Solana duplicate-broadcast handling with the core and React Native resolvers by treating `AlreadyProcessed` as broadcast acceptance and leaving execution outcome checks to transaction status resolution.
+Align raw Solana duplicate-broadcast handling with the core and React Native resolvers by accepting `AlreadyProcessed` when status is pending or unavailable while still rejecting explicit on-chain failures.
--- a/packages/sdk/src/vault/services/RawBroadcastService.ts
+++ b/packages/sdk/src/vault/services/RawBroadcastService.ts
@@ -267,9 +267,23 @@ export class RawBroadcastService {
 
     if (error) {
       if (isInError(error, 'already been processed', 'AlreadyProcessed')) {
-        // Match the authoritative core resolver: the node already accepted this
-        // exact signed payload, while execution outcome remains status-layer work.
-        return deriveSolanaRawTxSignature(rawTx)
+        const signature = deriveSolanaRawTxSignature(rawTx)
+
+        // A duplicate error proves that the node has seen this exact signed payload, but it
+        // does not prove successful execution. Reject only when the node already exposes an
+        // explicit failure; an unavailable or not-yet-indexed status remains accepted.
+        const { data: statuses } = await attempt(
+          client.getSignatureStatuses([signature], { searchTransactionHistory: true })
+        )
+        const signatureStatus = statuses?.value?.[0]
+        if (signatureStatus?.err) {
+          throw new VaultError(
+            VaultErrorCode.BroadcastFailed,
+            `Solana transaction was already processed but failed on-chain: ${JSON.stringify(signatureStatus.err)}`
+          )
+        }
+
+        return signature
       }
       throw error
     }
--- a/packages/sdk/src/vault/services/RawBroadcastService.ts
+++ b/packages/sdk/src/vault/services/RawBroadcastService.ts
@@ -265,20 +265,16 @@ export class RawBroadcastService {
       })
     )
 
-    let signature = sentSignature
-    let isDuplicate = false
     if (error) {
       if (isInError(error, 'already been processed', 'AlreadyProcessed')) {
-        // "AlreadyProcessed" only proves the node has seen and executed this signature before -
-        // not that the original execution succeeded. It must go through the same on-chain-failure
-        // check below as a fresh send, not be handed back as a hash unconditionally.
-        signature = deriveSolanaRawTxSignature(rawTx)
-        isDuplicate = true
-      } else {
-        throw error
+        // Match the authoritative core resolver: the node already accepted this
+        // exact signed payload, while execution outcome remains status-layer work.
+        return deriveSolanaRawTxSignature(rawTx)
       }
+      throw error
     }
 
+    const signature = sentSignature
     if (!signature) throw new Error('No transaction signature returned')
 
     // sendRawTransaction only confirms the node ACCEPTED the payload into its queue - it is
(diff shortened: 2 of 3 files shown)


gomesalexandre (dismissed):
Verified the load-bearing claim before anything else, and it holds: core really does return unconditionally on `AlreadyProcessed`, and so does RN. So the consistency argument is not hand-waving. But I think it lands at the wrong layer, and the receipts below show a failed Solana tx being handed back as a success hash from the public API.

## The claim checks out at the primitive layer

`packages/core/chain/tx/broadcast/resolvers/solana.ts:78-97` returns on `AlreadyProcessed` with no verification, pinned twice (`solana.test.ts:140` and the N1 test at `:156`). RN does the same (`platforms/react-native/chains/solana/rpc.ts:160-163`; its bounded status loop only covers *other* ambiguous errors). Both correct, both intentional.

What I want to flag is the sentence core uses to bound that trade-off:

```
// Only a raw-SDK consumer that broadcasts and skips that confirmation poll is
// exposed to the optimistic result; the CLI agent always confirms.
```

and the contract RN publishes right above its own return:

```
// A returned signature proves broadcast acceptance, not final confirmation.
// Consumers must continue polling transaction status until success or failure.
```

Both layers are explicitly primitives that push confirmation onto a caller. The optimism is safe there *because* something downstream confirms.

## preferably-blocking: nothing downstream of `broadcastRawTx` confirms

`RawBroadcastService.broadcastRawTx` is not a primitive, it is the terminal public API (`vault.broadcastRawTx`). Its only in-repo consumer is `clients/cli/src/commands/broadcast.ts:42`:

```ts
const txHash = await vault.broadcastRawTx({ chain: params.chain, rawTx: params.rawTx })
broadcastSpinner.succeed(`Transaction broadcast: ${txHash}`)
...
printResult(`Explorer: ${result.explorerUrl}`)
return result
```

Spinner goes green, hash prints, explorer link prints, command exits. No status poll. That is precisely the "raw-SDK consumer that broadcasts and skips that confirmation poll" core names as the exposed class. The cross-layer confirm test at `sessionTxConfirm.test.ts:154` covers the keysign/agent flow, which is a different path and does not run here.

`docs/SDK-USERS-GUIDE.md:1305` also documents the opposite contract to the changeset, telling external users `BroadcastFailed` is what covers "already submitted" errors, with no mention that they must poll.

And within this one file, Solana becomes the only chain whose duplicate branch does not fail closed. Straight from the file's own test names:

```
fails closed when a duplicate Cosmos transaction already failed execution
fails closed when a duplicate Ripple transaction finalized with a tec failure
fails closed when a duplicate Tron transaction already failed on-chain
```

Cosmos, Ripple and Tron all route duplicates through a verification step, TON/Sui/UTXO throw outright, and `verifyKnownRawTx` already sits at `RawBroadcastService.ts:79` as the house helper for exactly this. The comment I left on the Cosmos branch at `:349-354` said the quiet part out loud: this raw path is reachable via public `vault.broadcastRawTx`, so it must fail closed the same way. That was #1426, same bug class, different chain.

## Receipts

Patched the duplicate branch locally to keep acceptance semantics but fail closed on an *explicit* `signatureStatus.err`, then ran three probes both ways.

Against this PR's head, unpatched:

```
× PROBE: duplicate + EXPLICIT on-chain error -> fails closed
  AssertionError: promise resolved "'2Ana1pUpv2ZbMVkwF5FXapYeBEjdxDatLn7nv…'" instead of rejecting
  Received: "2Ana1pUpv2ZbMVkwF5FXapYeBEjdxDatLn7nvJkhgTSXbs59SyZSx866bXirPgj8QQVB57uxHJBG1YFvkRbFj4T"

✓ PROBE: duplicate + status NOT FOUND    -> still acceptance
✓ PROBE: duplicate + status UNAVAILABLE  -> still acceptance
```

The node returned `{ err: { InstructionError: [0, 'Custom'] } }` for that signature and the function returned it as a hash anyway. This is not theoretical, it is core's own documented case: `AlreadyProcessed` fires for a signature that was *processed but reverted on-chain*.

With the patch:

```
✓ PROBE: duplicate + EXPLICIT on-chain error -> fails closed
✓ PROBE: duplicate + status NOT FOUND    -> still acceptance
✓ PROBE: duplicate + status UNAVAILABLE   -> still acceptance
× treats duplicate-style Solana broadcast errors as acceptance without requiring a status lookup
     (your `expect(mockGetSolanaSignatureStatuses).not.toHaveBeenCalled()`)

Tests  1 failed | 49 passed (50)
```

Two probes pass in both arms and only the explicit-error one flips, so they discriminate the actual gap rather than just detecting "a status call happened". Baseline at your head is `47 passed (47)`, so nothing else moves. Restored both files, your branch is untouched.

## The narrow ask

You are right to delete the old `isDuplicate && !signatureStatus` throw. That guard fired on indexing lag and could false-fail a perfectly good duplicate, and a false reject on money movement is worse than a soft accept. Keep that part.

The ask is only: do not also delete the explicit-error case.

```ts
if (isInError(error, 'already been processed', 'AlreadyProcessed')) {
  const dup = deriveSolanaRawTxSignature(rawTx)
  const { data: dupStatuses } = await attempt(
    client.getSignatureStatuses([dup], { searchTransactionHistory: true })
  )
  if (dupStatuses?.value?.[0]?.err) {
    throw new VaultError(VaultErrorCode.BroadcastFailed, `Solana transaction was already processed and failed on-chain: ...`)
  }
  return dup
}
```

Strictly looser than what you are deleting (not-found, pending and lookup-unavailable all return the signature, proven by the two probes above), no new imports, one non-blocking RPC call on a branch that is already the slow path. It also removes the asymmetry the current diff leaves behind: fifteen lines down, a *fresh* send still throws on `signatureStatus.err`, while a duplicate returns unchecked. A duplicate is the case more likely to have already executed, so it is the one being checked less.

If you would rather keep the diff as-is, the alternative that closes the same hole is doing what #1227 offered as option 2: make "caller must confirm" explicit in the docs and have `clients/cli/src/commands/broadcast.ts` actually poll before it prints success. That is a bigger change and I would not ask for it in this PR.

## nit: severity

The changeset is a `patch`, but this flips a documented public-API throw into a success return. Reads more like a `minor` to me.

Happy to be wrong on any of the above if there is a confirmation path for `vault.broadcastRawTx` that I did not find.

johnnyluo (approved):
LGTM

PR #1763, opened 2026-08-05
fix(cli): disclose token contracts in send dry runs

neavra:
## Summary

The send dry-run preview identified non-native assets by symbol only, so same-chain token variants could not be distinguished reliably.

This change adds an optional `contractAddress` field to SDK and CLI JSON dry-run results for non-native sends. Native-send output remains unchanged. The human-readable preview now shows the full resolved contract address in parentheses next to the token amount.

## Tests

- pins the exact Polygon USDC and USDC.e contract addresses and verifies they differ
- verifies native dry runs omit `contractAddress` and retain their existing shape
- verifies all existing token dry-run fields retain their values
- verifies JSON and human-readable CLI previews disclose the resolved contract
- passes build, typecheck, lint, complexity, SDK unit, CLI, and full repository test suites

## Summary by CodeRabbit

* **New Features**
  * Send dry-run results now show resolved token contract addresses when available.
  * CLI JSON output and transaction previews include contract addresses for token sends.
  * Token contract addresses are displayed safely with escaped control characters.

* **Bug Fixes**
  * Native-asset sends continue to omit contract address details.
  * Token variants, including Polygon assets, now report the correct resolved contract address.


gomesalexandre (approved):
Reviewed. This is a WYSIWYS improvement on a surface where the ambiguity is genuinely dangerous, and the wiring is right.

**The property that matters: the address shown is the one that gets signed.** I traced it rather than assuming:

```ts
const tokenInfo = this.resolveTokenInfo(chain, symbol)          // :1753
const coin = this.buildAccountCoin(chain, await this.address(chain), tokenInfo)   // :1754  <- what is signed
...
...(tokenInfo.contractAddress ? { contractAddress: tokenInfo.contractAddress } : {})   // :1785  <- what is shown
```

Same `tokenInfo` object feeds both, so the preview cannot drift from the signed asset. That is the whole point of disclosing it - a preview showing an address the signer will not use would be worse than showing nothing.

**The confusion it closes is real.** Polygon carries both USDC and USDC.e; by symbol alone the two are indistinguishable in the preview, and picking the wrong one is a silent wrong-asset send. Pinning the two exact contract addresses in the test and asserting they differ is the right assertion - it fails if the resolver ever collapses them.

**Nice detail worth calling out:** `escapeTerminalControls(contractAddress)` on the human-readable line. A contract address reaching the preview can originate from a token list rather than the user, so treating it as untrusted text before writing it to a terminal is correct and easy to forget.

Native sends are unaffected - the field is spread conditionally, so the JSON shape and the printed line are byte-identical when `contractAddress` is absent. The test pinning that is what makes it safe to land without auditing every native-send consumer.

**Limitation, stated plainly:** I could not execute the suite - the clone I have has no workspace install and there is not enough disk here for one. This is a static trace of the resolve-to-sign path plus a read of the tests, not a green local run. Everything I could verify without running holds.

Approving.

PR #1776, opened 2026-08-06
feat(swap): enable SUI as a SwapKit swap source

### LINKED ISSUE #1775: [BUG] SwapKit: SUI unusable as a swap source, and two latent defects in the transfer-route path
## Describe the bug

SUI cannot be used as a **swap source**. Selecting it dead-ends with a raw provider error instead of producing a signable transaction, even though every primitive required to sign a SwapKit Sui route already exists in this repo.

Investigating this surfaced two further defects in the SwapKit transfer-route path that are latent today and would break a Sui source the moment it is unblocked.

### 1. Sui is hard-blocked as a SwapKit source despite the signing path existing

`getSwapKitQuote` rejects Sui before any network call via `SWAP_SOURCE_TX_BUILD_UNSUPPORTED` (`packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.ts`), throwing:

```
SwapKit Sui source swaps are not yet supported for signing (quote-only for now).
```

The comment above that guard justifies the block by claiming there is *"no `GeneralSwapTx` variant a Sui signer can consume — it is neither `evm` … nor a plain `transfer`."* **That reasoning is stale.** The `transfer` variant already carries opaque pre-built bytes via `txType` + `txPayload` — that is exactly how SwapKit Bitcoin PSBT routes work today (`packages/core/mpc/keysign/swap/build.ts`, `getSwapKitBitcoinSignData`).

Everything a Sui source needs is already implemented and tested:

| Capability | Where it lives | Status |
|---|---|---|
| Arbitrary BCS PTB → signing input | `getSuiSigningInputs`, `signSui` branch (`TW.Sui.Proto.SignDirect`) | exists, used by the dApp Wallet Standard path |
| Sui intent digest (blake2b-32) | `getSuiTransactionDataDigest` — `chains/sui/sign.ts` | exists |
| Wallet-standard signature envelope | `buildSuiSerializedSignature` — `chains/sui/sign.ts` | exists |
| Skip RPC for a pre-built PTB | `getSuiChainSpecific`, `signSui` early return | exists |
| `txPayload` / `txType` on the wire | `SwapKitSwapPayload` proto | exists |
| `sui_executeTransactionBlock` broadcast | `tx/broadcast/resolvers/sui.ts` | exists |

The base64 decode for a Sui payload is likewise already present but dormant (`encodeSwapKitTxPayload`, `normalizedTxType === 'SUI'`).

For reference, iOS and Android ship this end to end (`SwapKitSuiSigner.swift`, wired through `KeysignMessageFactory` and `KeysignViewModel`), so the SDK is the only platform where a SUI source dead-ends.

### 2. `disableBuildTx` is opt-out-by-exception, so any prebuilt-tx source chain silently loses its transaction

In `getSwapKitQuote`, the `/v3/swap` request sets:

```ts
disableBuildTx: shouldUseTransferTx(from.chain) && from.chain !== Chain.Bitcoin ? true : undefined
```

`shouldUseTransferTx` is keyed off `swapKitTransferSourceChains`, a list that mixes two unrelated kinds of chain:

- **deposit-only** chains (TON / XRP / ADA) that need nothing but an address, and
- **prebuilt-tx** chains whose returned bytes *are* the thing being signed.

Bitcoin is the only member of the second group, and it is carved out with an ad-hoc `!== Chain.Bitcoin`. Any chain added to the transfer list therefore defaults to "tell SwapKit not to build a transaction" — the wrong default for a prebuilt-tx chain, and a silent one: the response simply arrives with no `tx`, and the failure surfaces much later as an empty-payload error during keysign-payload construction.

Confirmed against the live API (`https://api.vultisig.com/swapkit-win`, `SUI.SUI → ETH.ETH`, NEAR route):

| `/v3/swap` request | Response |
|---|---|
| `disableBuildTx: true` | `200`, `tx` **absent** — nothing to sign |
| build enabled | `500 unableToBuildTransaction` / `toolbox_sui_insufficient_balance` |

The second result is the important one: SwapKit **does** build a Sui PTB; it only refused here because the probe address was unfunded. Note it enforces this balance check during tx construction even with `disableBalanceCheck: true`, so a Sui source route can only be built for a funded address.

### 3. `meta.txType` is treated as a stable discriminator, but SwapKit renames it without versioning

`encodeSwapKitTxPayload` decides how to decode the payload purely from `meta.txType` — base64-decoding only for `PSBT` and `SUI`, returning an empty array for `CARDANO`, and otherwise UTF-8-encoding the raw value.

SwapKit has changed these labels live, mid-flight, with no version bump. The iOS decoder documents two such renames it had to absorb (`SwapKitSwapResponse.decodeTx`), accepting both spellings for each:

- `SOLANA` → `SERIALIZED_BASE64`
- `CARDANO` → `CBOR`

This repo handles neither new spelling (no occurrence of `SERIALIZED_BASE64` or a `CBOR` branch anywhere in `packages/`). The failure mode is silent rather than loud: an unrecognized label falls through to `textEncoder.encode(tx)`, which UTF-8-encodes the **base64 string itself** instead of decoding it, yielding a well-formed-looking `txPayload` full of garbage bytes.

EVM and Solana are currently immune because `buildSwapKitTx` dispatches those on the **source chain**, not on `txType` — which is the pattern the rest of the path should follow.

There is a second-order consequence for cross-device signing: iOS does not persist the wire label at all, it hardcodes `txType: "SUI"` when building its payload (`SwapPayloadBuilder.buildSwapKitSuiPayload`). If the SDK stores whatever SwapKit happened to send, the two platforms disagree on a field that is part of the cosigned `SwapKitSwapPayload`.

## To Reproduce

1. Open a vault holding SUI on desktop or the extension.
2. Go to **Swap** and select **SUI** as the source asset, ETH as the destination.
3. Enter any amount and request a quote.

## Expected behavior

A SUI source route quotes, builds a signable keysign payload from SwapKit's pre-built PTB, signs through the existing `signSui` path, and broadcasts via `sui_executeTransactionBlock` — matching iOS and Android.

## Actual behavior

The quote is rejected up front with `SwapKit Sui source swaps are not yet supported for signing (quote-only for now).`

With that block lifted, defect 2 causes SwapKit to return a response with no `tx`, so keysign-payload construction fails with an empty-payload error. In the client this surfaces as **"Failed to load"** on the swap form's Total Fees row, because the fee query is derived from the keysign payload (`useSwapFeesQuery` → `useSwapKeysignPayloadQuery`). The quote itself still renders, since it only needs the target address and amount — so the form looks healthy apart from the fee.

With defects 1 and 2 addressed, defect 3 remains latent and would corrupt `txPayload` without any error the moment SwapKit renames the Sui label.

## Additional context

- Cardano is a genuinely separate case and should stay blocked: `encodeSwapKitTxPayload` returns an **empty** byte array for it, so there is no decode implementation at all. iOS covers it with a distinct `CARDANO_PREBUILT` CBOR path.
- Related client-side issue: vultisig/vultisig-windows#4552, which proposed hiding SUI or showing a "coming soon" message in the swap form. That would be the wrong layer — the gating lives here in `swapKitSourceChains`, the picker is not source/dest aware (so hiding SUI would also break it as a working *destination*), and it would mean building throwaway UI for a chain whose signing primitives already ship.
- Broader gap for context: iOS supports `PSBT`, `PSBT_DOGE`, `PSBT_BCH`, `PSBT_DASH`, `PSBT_ZEC`, `SUI`, `TRON`, `CARDANO_PREBUILT` plus deposit-only `TON`/`CARDANO`/`XRP` as SwapKit sources. The SDK supports `EVM`, `SOLANA`, `PSBT` (Bitcoin only) and deposit-only transfers, so Doge/BCH/Dash/Zcash/Tron are behind as well.

Ehsan-saradar:
## Description

Enables SUI as a SwapKit swap source, and fixes two defects in the transfer-route path that unblocking it exposed.

Fixes #1775

### Sui was blocked on a premise that was not true

`getSwapKitQuote` rejected Sui before any network call, on the stated grounds that no `GeneralSwapTx` variant could carry a programmable transaction block. The `transfer` variant already carries opaque pre-built bytes as `txType` + `txPayload` — that is exactly how a Bitcoin PSBT route reaches the signer.

**Nothing needed porting.** `getSuiSigningInputs` has forwarded arbitrary BCS-serialized PTBs to WalletCore's `SignDirect` since the dApp Wallet Standard path landed, and the intent digest it produces is the same blake2b-32 over `[0,0,0] || ptb` that iOS computes by hand in `SwapKitSuiSigner.swift`. The digest, the wallet-standard signature envelope, the RPC-skipping `signSui` chain-specific branch, the `txPayload`/`txType` proto fields and the `sui_executeTransactionBlock` broadcast resolver were all already here and already tested. This PR routes the quote through them rather than adding a signer.

### `disableBuildTx` was opt-out-by-exception

```ts
disableBuildTx: shouldUseTransferTx(from.chain) && from.chain !== Chain.Bitcoin ? true : undefined
```

`swapKitTransferSourceChains` mixes two unrelated kinds of chain: deposit-only ones (TON/XRP/ADA) that need nothing but an address, and prebuilt-tx ones whose returned bytes *are* what gets signed. Bitcoin was the only member of the second group and was carved out with an ad-hoc `!==`, so any chain added to the list defaulted to telling SwapKit not to build a transaction — the wrong default, and a silent one: the response simply arrives with no `tx`, and the failure surfaces much later as an empty payload during keysign construction.

Verified against the live API (`SUI.SUI → ETH.ETH`, NEAR route):

| `/v3/swap` request | Response |
|---|---|
| `disableBuildTx: true` | `200`, `tx` **absent** |
| build enabled | `500 toolbox_sui_insufficient_balance` (probe address unfunded — SwapKit *does* build it) |

Note SwapKit enforces that balance check during construction even with `disableBalanceCheck: true`, so a Sui route only builds for a funded address.

The two kinds of chain are now named separately, so membership of the transfer list no longer implies anything about who builds the transaction.

### `meta.txType` is not a stable discriminator

SwapKit renames these labels live, mid-flight, with no version bump — iOS documents absorbing `SOLANA` → `SERIALIZED_BASE64` and `CARDANO` → `CBOR`, accepting both spellings for each. This repo handled neither new spelling, and the failure mode was silent: an unrecognized label fell through to `textEncoder.encode(tx)`, UTF-8-encoding the **base64 string** rather than decoding it, producing a `txPayload` of the right shape and entirely wrong bytes.

Decoding now dispatches on the source chain — the discriminator EVM and Solana already used, and the only one SwapKit cannot rename. The stored `txType` is normalized to `SUI` because iOS hardcodes that spelling rather than persisting what it received, and the field is part of the cosigned `SwapKitSwapPayload`; a device storing the wire value would disagree with its own cosigner. A Sui route whose `tx` is not a string is now rejected outright instead of having a JSON object encoded into the payload.

### Pre-built PTBs reported a zero network fee

`getSuiChainSpecific` returned an empty `SuiSpecific` for `signSui` — correct that there are no construction inputs to fetch, but `getSuiFeeAmount` reads its budget from that message, so `BigInt('')` displayed a 0 fee while the chain charged the budget baked into the bytes. The gas budget and price are now read back out of the PTB offline with no RPC call; an undecodable payload still falls back to blank rather than blocking a transaction over a display concern. This corrects the dApp signing path as well as swaps.

Cardano stays blocked as a source — its payload decode returns an empty byte array, so there is no implementation to route to.

## Which feature is affected?

- [x] Swap - Please attach a tx link for the swap here
- [x] New Chain / Chain related feature - Please attach tx link here

**Mainnet SUI → ETH swap, signed and broadcast end to end:**
https://track.swapkit.dev/?hash=7widSUS7ZXoiHL6VMwv3iVoeMM5t7vaHdUdRJRWKEwHA&chainId=sui

## Checklist

Test coverage added:

- Sui source route maps to a transfer tx carrying the decoded PTB bytes
- txType normalized to `SUI` given a renamed label (`SERIALIZED_BASE64`) or no `meta` at all
- `disableBuildTx` is **not** sent for a Sui source (direct regression test)
- a Sui route whose `tx` is not a base64 string is rejected
- keysign payload forwards the PTB via `signSui`, base64-encoded and byte-identical, while `SwapKitSwapPayload.txPayload` stays raw bytes for cross-device parity
- an empty PTB is rejected, and a route signs even when SwapKit omits or renames the txType
- `getSuiChainSpecific` reads gas out of the PTB without touching the RPC, and falls back to blank when it cannot be decoded

Full core suite: 2546 passed, 1 skipped. Typecheck, lint and format clean.

## Additional context

Related client-side issue: vultisig/vultisig-windows#4552, which proposed hiding SUI or showing a "coming soon" message in the swap form. That is the wrong layer — the gating lives here in `swapKitSourceChains`, the picker is not source/dest aware (so hiding SUI would also break it as a working *destination*), and it would mean building throwaway UI for a chain whose signing primitives already ship.

For context on the remaining gap: iOS supports `PSBT`, `PSBT_DOGE`, `PSBT_BCH`, `PSBT_DASH`, `PSBT_ZEC`, `SUI`, `TRON`, `CARDANO_PREBUILT` plus deposit-only `TON`/`CARDANO`/`XRP` as SwapKit sources. After this PR the SDK covers `EVM`, `SOLANA`, `PSBT` (Bitcoin), `SUI` and deposit-only transfers — so Doge/BCH/Dash/Zcash/Tron remain behind.

## Summary by CodeRabbit

* **New Features**
  * Added Sui as a supported SwapKit source chain.
  * Enabled signing of prebuilt Sui transactions while preserving transaction data.
  * Added extraction of Sui gas budget and reference gas price.
  * Normalized Sui transaction metadata and transfer routes.

* **Bug Fixes**
  * Improved validation and handling of invalid or undecodable Sui payloads.
  * Continued blocking unsupported Cardano source transactions.


gomesalexandre (approved):
Reviewed with the money-path lens. The premise correction is right and the pattern is an existing accepted one, not a new blind-signing surface.

**Verified the precedent rather than taking the body's word for it.** `getSwapKitBitcoinSignData` on `main` already forwards an opaque aggregator-supplied payload straight to signing:

```ts
const isSwapKitBitcoinPsbt = (fromCoin, transfer) =>
  fromCoin.chain === Chain.Bitcoin && transfer.txType?.toUpperCase() === 'PSBT'
```

So "the `transfer` variant already carries opaque pre-built bytes, that is how a Bitcoin PSBT route reaches the sign path" is accurate, and Sui joins an established route rather than opening a new class of unverified signing.

**The base64 re-encode comment earns its place.** Keeping `txPayload` as raw bytes so a cosigning iOS/Android peer rebuilds byte-identical signing input, while `SignSui`'s wire contract wants base64, is exactly the kind of thing that silently breaks a 2-of-2 ceremony if someone "simplifies" it later. Good that it is written down at the conversion site.

**One deliberate difference from the Bitcoin path, and I want it visible rather than buried in a comment.** Bitcoin gates on chain AND `txType === 'PSBT'`; Sui gates on chain alone:

> Deliberately NOT gated on `txType`: SwapKit renames base64 tx types on the wire without versioning, so the source chain is the only reliable discriminator.

I think that is the right call - a `txType` check against an unversioned upstream string is a false-reject waiting to happen, and false-rejecting a legitimate swap is its own harm. Worth being explicit about what it costs: any Sui `transfer` payload SwapKit returns will now be staged as a PTB. If upstream ever emits a non-PTB Sui payload, it goes to `signSui` as-is. The realistic failure there is an unparseable signing input - a failed ceremony, not a wrong-recipient signature - which is the acceptable side of that trade. The empty-payload throw covers the degenerate case.

**Limitation, stated plainly:** I could not execute the suite - the clone I have has no workspace install and there is not enough disk here for one. This is a static review plus a check of the Bitcoin precedent on `main`, not a green local run. Given this enables a swap SOURCE, I would want someone to confirm the added tests actually run green in CI before merge rather than treating my approval as covering execution.

Approving on the reasoning and the pattern consistency.

PR #1779, opened 2026-08-06
fix: canonicalize stored token identities and discovery names

neavra:
## Summary

Tokens added via `tokens <chain> --add` and via `tokens <chain> --discover` previously stored two different id formats for the same asset (`<Chain>-0x…` vs bare `0x…`), which made `balance --tokens` render a doubled chain prefix (`Polygon:Polygon-0x…`) and gave JSON consumers two id schemes for one asset. Discovery also produced order-dependent synthetic tickers like `USDC_1` that the CLI would display but not accept back as input.

- **One canonical stored id**: bare contract address (lowercased on EVM chains, case preserved elsewhere), regardless of add-path. Legacy `<Chain>-0x…` ids already stored in vaults still resolve on read (resolution, balance keys, and removal all accept both forms); ids are normalized on write only — no vault migration.
- **Stable discovery names**: known contracts get their curated ecosystem ticker (e.g. Polygon `0x2791…` → `USDC.e`); unknown tokens have the order-dependent `_N` suffix stripped. Displayed names are accepted back as `--token` input; same-symbol tokens are disambiguated by contract address, as previews already do.
- **Output contract fixes**: `tokens --remove` now emits the standard JSON success envelope instead of empty stdout; `--name` is documented as required with `--add`.

Resolution behavior (symbol / contract / alias lookup order) and signing are unchanged.

## Tests

- Same stored id via `--add` and `--discover`; no doubled chain prefix in balance keys.
- Legacy prefixed ids resolve by both stored and bare forms.
- Curated ticker displayed for the bridged Polygon USDC and accepted back as input.
- Removal JSON envelope; help text marks `--name` required.

All added tests are offline (no network calls) and each was verified revert-sensitive against its source change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized token identification across chains, including legacy IDs and EVM address casing.
  * Prevented duplicate tokens when re-adding existing assets.
  * Ambiguous token symbols now fail safely instead of selecting the wrong asset.
  * Improved token discovery naming when multiple contracts share a symbol.
  * Preserved case-sensitive identification for non-EVM assets.

* **CLI Improvements**
  * Token removal now provides structured JSON success output.
  * Updated token-add help with clearer examples and required name guidance.


gomesalexandre (changes requested):
Two correctness bugs still block this.

1. `packages/sdk/src/vault/services/TokenDiscoveryService.ts:18-20` and `packages/sdk/src/vault/tokenRef.ts:38-52` now lowercase token identities for every chain. But `packages/core/chain/utils/isValidTokenId.ts:64-69` explicitly says non-Ripple ids are already canonical and are returned unchanged. That means a non-Ripple case-sensitive asset id can now collapse with a case-variant sibling during discovery / ambiguity checks. Please scope this case-folding to the chains that actually canonicalize ids, and add a regression on a non-Ripple chain whose asset ids must preserve case.

2. `packages/sdk/src/vault/services/BalanceService.ts:152-155` and `:192-195` only canonicalize the outer result key. The nested balance payload still comes from `formatBalance(..., request.tokenId, ...)`, and `request.tokenId` is still the legacy stored id. So a legacy record can now surface as `balances["Ethereum:0x..."]` while `balances["Ethereum:0x..."].tokenId === "Ethereum-0x..."`, which keeps the two id schemes inconsistent inside the same JSON envelope. The new tests assert the map key, but they do not assert the nested `balance.tokenId`, so this path still slips through.

Validation: re-grounded on PR head `ab572a5e6` in a fresh disposable clone against fresh `origin/main`; `git diff --check` passed. I did not get a runnable dependency install receipt from that disposable clone.

gomesalexandre (dismissed):
current-head review on `26779042` only.

I re-grounded this in a disposable exact-head clone, re-ran the touched CLI/SDK suites there, and then added two focused exact-head probes for the identity boundaries the current diff still leaves open.

what passed
- `clients/cli/src/commands/tokens.test.ts` under `clients/cli/vitest.config.ts`: 9/9 green
- `packages/sdk/tests/unit/services/BalanceService.test.ts`
- `packages/sdk/tests/unit/vault/services/TokenDiscoveryService.test.ts`
- `packages/sdk/tests/unit/vault/tokenRef.test.ts`
  - combined exact-head receipt: 70/70 green under `packages/sdk/tests/unit/vitest.config.ts`
- `yarn build:sdk` is currently blocked by a pre-existing main-side type error in `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts` (`auxiliaryData` not in `ISigningInput`), and the same failure reproduces unchanged on `origin/main`, so I did not treat it as evidence against this PR.

still blocking

1. checksum-vs-lowercase EVM ids still dedupe as TWO tracked assets, not one
   - exact-head probe: start with an Ethereum token stored under lowercase `0xa0b8...6eb48`, then call `BalanceService.addToken(...)` with the checksummed form `0xA0b8...6eB48`
   - actual result on `26779042`: tracked ids become
     - `0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48`
     - `0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`
   - so `normalizedTokenIdentity()` / the add-path still preserve non-Ripple casing at the point where the SDK needs one canonical EVM asset identity. That means a vault can hold duplicate tracked records for one EVM contract depending only on address casing.

2. `removeToken()` still deletes the WRONG Solana record when two case-variant mint ids coexist
   - exact-head probe tokens:
     - `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`
     - `ePjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`
   - call: `removeToken(Chain.Solana, 'ePjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v')`
   - expected remaining id: `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`
   - actual remaining id on `26779042`: `ePjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`
   - so the early case-insensitive fallback in `BalanceService.removeToken()` is still winning before the chain-aware exact/canonical path, and on a case-sensitive chain it removes the first case-variant sibling instead of the requested one.

I think this needs:
- EVM-only canonical lowercasing at every add/result/cache/remove identity boundary that is supposed to represent one contract, plus a checksum-vs-lowercase regression
- the unconditional case-insensitive remove fallback scoped away from case-sensitive chains, or moved behind the chain-aware exact/canonical match, plus the Solana case-variant removal regression above

Keeping this at changes requested until those two exact-head probes flip green.

packages/sdk/src/vault/services/BalanceService.ts
```diff
@@ -385,10 +398,18 @@ export class BalanceService {
       // address-keyed record, and then "which record" and "which asset" are
       // different questions. Removal must mean the same record every other
       // surface means for that reference.
-      const upper = id.toUpperCase()
-      const lower = id.toLowerCase()
+      const upper = tokenId.toUpperCase()
       tokenIndex = tokens.findIndex(token => token.symbol?.toUpperCase() === upper)
```
gomesalexandre, line 402: pushed f2cd2b363 on top of this branch.

there was still one literal-id removal edge in `removeToken()`: we canonicalized the raw `tokenId` to `id` before deciding whether the ref should be treated as a symbol or as an exact stored record id. On EVM that means a literal id like `Ethereum-USDC` collapsed to `usdc`, so the symbol-first branch could delete a different sibling whose symbol was `USDC` instead of the exact record the caller named.

fix: keep symbol matching keyed off the raw `tokenId` text, then keep the existing exact-id / canonical-id fallbacks behind it.

receipt:
- new regression: `removes the exact legacy-prefixed id before symbol fallback can reinterpret it`
- focused suites green locally on this head:
  - `packages/sdk/tests/unit/services/BalanceService.test.ts`
  - `packages/sdk/tests/unit/vault/services/TokenDiscoveryService.test.ts`
  - `packages/sdk/tests/unit/vault/tokenRef.test.ts`

gomesalexandre (approved):
Good fix, and the part that decides whether it is safe is handled correctly: the canonicalisation is EVM-only.

That distinction is the whole ballgame here. EVM contract addresses are case-insensitive so checksum casing really is display metadata, but Solana mints are base58 and Sui type tags are case-significant - case-folding those would silently corrupt an identity. Measured rather than trusted:

```
EVM ethereum checksummed   changed=true   -> 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
EVM polygon   checksummed  changed=true   -> 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
SOLANA                     changed=false  -> EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v
SUI                        changed=false  -> 0xAbCdEf::coin::COIN
COSMOS ibc/...             changed=false  -> unchanged
```

## the risk I actually went looking for

"Legacy ids still resolve on read, normalized on write only, no vault migration" is the right migration posture, but it means a vault can hold `Polygon-0xA0b8…` and `0xa0b8…` for the same asset at once. If those produced different cache/result keys, `balance --tokens` would list the asset twice and a consumer summing rows would double-count it.

They collapse:

```
MUST-PASS  legacy == canonical    true
MUST-FAIL  two different tokens   false     (no over-collapse)
SOL legacy prefix stripped        EPjFWdd5…  case intact
SOL lowercase != original         true       (still case-sensitive)
```

So `stripLegacyTokenIdPrefix` then `normalizeTokenId` gives one key per asset regardless of which era wrote it, and routing every `resultKey` / `cacheKey` / `tokenId` in `BalanceService` through `normalizedTokenIdentity` is what makes that hold at every call site rather than just at comparison time. The must-fail control matters as much as the must-pass one: a normaliser that collapses everything would also report "true" on the first line.

Keeping `caseInsensitiveTokenIdsMatch` separately gated on `getChainKind(chain) === 'evm'` rather than folding it into the general matcher is the same discipline in a second place.

```
packages/core/chain/utils/isValidTokenId.test.ts   11 passed
sdk suite                                          216 files, 3312 passed
CI                                                 all 5 checks green
```

## worth knowing if you probe this locally

My first run of this said the opposite - that legacy and canonical did **not** collapse - and it was wrong for a reason that will bite the next person too. `tokenRef.ts` imports `normalizeTokenId` from the `@vultisig/core-chain` package alias, and that resolves to `packages/core/chain/dist/`, not to your source:

```
require.resolve('@vultisig/core-chain/utils/isValidTokenId')
  -> packages/core/chain/dist/utils/isValidTokenId.js

grep -c "getChainKind(chain) === 'evm'"
  dist    0
  source  1
```

So without rebuilding the shared packages, `normalizedTokenIdentity` runs the OLD normaliser and legacy ids come back checksummed. Nothing wrong with your change - it just means an unbuilt workspace reproduces the exact bug this PR fixes, and I nearly wrote that up as a defect. Rebuild shared packages (or bind against the source) before believing a local result here.

PR #1782, opened 2026-08-06
feat(sdk): support external Fast Vault coordinators

secretagentfranco:
## Description

Adds the public SDK primitives needed by external watch-only wallet coordinators while keeping Vultisig responsible only for MPC key management and signing.

- expose `vault.extendedPublicKey()` for mainnet xpub and testnet tpub serialization
- allow `signBytes()` callers to provide the exact BIP-32 derivation path associated with an external wallet input
- propagate explicit paths through both Fast Vault and Secure Vault signing
- allow React Native Fast Vault creation to signal completion without blocking on a zero-body relay response
- make React Native vault-backup format detection compatible with plain `Uint8Array` views
- add regression coverage and a minor SDK changeset

No linked issue.

## Which feature is affected?

- [x] Create vault (Secure / Fast)
- [x] Sending
- [ ] Swap
- [x] New Chain / Chain related feature

## Receipts

- `yarn build:all`
- `yarn lint`
- `yarn test` — all SDK, Rujira, shared/core, CLI, and browser-example workspaces pass
- `yarn cli:build`
- packed SDK/CLI contract smoke reaches the existing MCP package check; that final check currently reports the unrelated missing `@vultisig/mcp` `dist/bin/mcp-server.js` artifact
- cross-repo Android consumer: BitPay imported the derived Testnet4 tpub, built a native BitPay transaction proposal, supplied the BitPay input path to `signBytes()`, received a valid Fast Vault signature, and broadcast the transaction

Bitcoin Testnet4 transaction:
https://mempool.space/testnet4/tx/006e4b89ff75d8f14398aba3e708c9b0392d2e54c82aefe1e3a406f3560629ca

The transaction spends the previously unconfirmed BitPay wallet UTXO, sends 1,547 sats to the requested recipient, and returns change to the BitPay-derived address. Vultisig received the digest and derivation path only; BitPay retained UTXO selection, fees, transaction construction, proposal publishing, and broadcasting.

## Checklist

- [x] I have commented the non-obvious derivation and relay behavior
- [x] My changes generate no new build or lint warnings
- [x] I have added tests that prove the feature and compatibility fixes

## Additional context

The BitPay consumer PR will remain draft until this SDK change is released and its temporary local tarball dependency can be replaced with the published version.

## Summary by CodeRabbit

- **New Features**
  - Added support for serialized ECDSA extended public keys.
  - Added optional derivation paths for byte-signing operations.
  - Added configurable waiting for Fast Vault key-generation completion.
  - Added support for external wallet coordinators.
  - Improved React Native vault-backup format detection.

- **Bug Fixes**
  - Improved handling of truncated or malformed vault backup data.
  - Improved compatibility with plain `Uint8Array` backup data.


gomesalexandre (dismissed):
lgtm 🫡

Reviewed remotely via gh only, no local checkout touched.

what I verified:
- `derivePath` is now threaded end-to-end through the fast-signing and relay-signing paths, with the raw path sent to the server and the apostrophe-stripped path used for local keysign, so both sides stay aligned.
- `ExtendedPublicKeyService` normalization matches the MPC semantics here: non-hardened public derivation from the root public material, with the correct 33-byte pubkey / 32-byte chain-code guards.
- `waitForServerKeygenComplete` stays default-safe. Existing callers still await the server-side keygen completion; the opt-out is explicit and documented, and the fire-and-forget branch catches/logs failures instead of leaking an unhandled rejection.
- `startsWithBytes` no longer relies on `Buffer.subarray(...).equals(...)`, which is a real RN footgun when the returned slice is only a `Uint8Array`.
- The tests are meaningful: fixed-vector tpub coverage, derivePath payload threading, and the unresolved-completion test that proves signal-only mode returns before the POST settles.
- The PR body carries a real Testnet4 receipt through the full new flow: external coordinator proposal + explicit-path signing + on-chain confirm.

non-blocking:
- worth adding the cheap mainnet xpub fixture CodeRabbit asked for so both version-byte branches are pinned.
- worth documenting in the method jsdoc that the exported xpub/tpub is MPC-derived metadata-wise, not a canonical hardened BIP84 account xpub.

Given the end-to-end Testnet4 receipt on the actual changed path, I don't see a blocker here.

gomesalexandre (commented):
current-head remote pass on `88cb660`:

code looks good to me on the changed path.
- `derivePath` is threaded through fast + secure signing without changing the default path behavior for existing callers.
- the xpub/tpub export path is normalized to non-hardened public derivation, with the expected key-material length guards before serialization.
- the RN `Uint8Array` backup-format fix is scoped to the exact prefix check that previously relied on `Buffer.equals` semantics.
- the fast-vault keygen completion toggle stays opt-in and the fire-and-forget branch still catches relay completion failures instead of leaking an unhandled rejection.

non-blocking:
- worth adding the cheap mainnet `xpub` fixture alongside the existing `tpub` coverage.
- worth aligning the new extended-public-key validation failures with the existing coded `VaultError` surface.

gating only:
- this head includes gomes-authored merge-main commits, so I can't be the approving stamp on it.
- current-head status only shows CodeRabbit on `88cb660`; I do not see fresh build/lint/test checks attached to this head.

so from a remote conceptual pass I don't see a blocker in the diff itself, but this one still needs a non-gomes approval on the current head after proper current-head CI.

gomesalexandre (approved):
Went deep on the trust-boundary question specifically, since "external coordinator" in the title sounds scarier than what this actually does.

**What "external coordinator" actually means here — traced the whole flow:**

This is NOT a new network trust boundary where some remote server gets pushed into the MPC signing protocol. "Coordinator" = an app embedding `@vultisig/sdk` as a library (BitPay, per the PR's own receipt), calling the SDK in-process. The two new capabilities:

1. `vault.extendedPublicKey({derivePath, network})` — derives a serialized xpub/tpub using `bip32` + `tiny-secp256k1` (already-declared deps, not newly introduced). Traced `this.publicKeys.ecdsa` / `this.hexChainCode` back to `ServerManager.createFastVault` — confirmed these are the DKLS/Schnorr keygen's **public** outputs (`ecdsaResult.publicKey`), never `ecdsaResult.keyshare`. No private key material anywhere near this path.
2. `SignBytesOptions.derivePath` — threaded through `FastSigningService`/`RelaySigningService` into `getChainSigningInfo`, which already had `derivePath?: string` in its type *before* this PR (pre-existing internal capability, this PR just exposes it on the public API surface). `chainPath = derivePath.replaceAll("'", '')` — confirmed this normalization (strip apostrophes = force non-hardened derivation) is the exact same scheme `ExtendedPublicKeyService.parsePublicDerivationPath` uses for the xpub export. Both walk the same non-hardened child-index space (`< 0x80000000`), so a coordinator can only ever select among keys any holder of the vault's own public xpub could already compute — never escape into different key material.

**The actual fund-safety question — can a coordinator smuggle a different payload than what's approved?** No: `signBytes` is a pre-existing "sign this pre-hashed digest, advanced/raw use case" primitive (docstring predates this PR). The coordinator already had 100% control over "what gets signed" before this change — that's inherent to the primitive, not something this PR introduces or worsens. `derivePath` only changes *which key* signs the digest, not *what* the digest represents. And since Bitcoin/EVM validate a signature against the specific locking script/address it needs to match, a wrong or malicious derivePath can only ever produce a signature that's rejected on-chain — never one that spends funds the signer doesn't legitimately control.

**Concrete end-to-end proof, not just theory**: independently fetched the Testnet4 tx the PR cites (https://mempool.space/testnet4/api/tx/006e4b89ff75d8f14398aba3e708c9b0392d2e54c82aefe1e3a406f3560629ca) — confirmed real, confirmed on-chain at block 147234, a valid P2WPKH witness signature correctly unlocking the BitPay-derived address `tb1qvl2mm7zk7nmcydtuhrjckx0ewzcsv8t6kwpacj`. This closes the one gap I was probing (does the xpub-derived address actually match what `signBytes({derivePath})` produces a valid signature for) with a genuine broadcast receipt, not a unit-test claim.

**`waitForServerKeygenComplete: false`**: verified this only affects a post-ceremony *relay* handshake (`setKeygenComplete`/`waitForKeygenComplete` POST/poll `{messageRelay}/complete/{sessionId}`) — the actual DKLS/Schnorr keygen ceremony (`dkls.startKeygenWithRetry()`, `schnorr.startKeygenWithRetry()`) already completed before this branch runs, so skipping the wait doesn't touch cryptographic vault creation. It's opt-in (defaults to `true`/waits, zero behavior change for existing callers), the risk is clearly documented in both the code comment and changeset ("callers that disable this must use an independent server-readiness signal"), and errors are logged not swallowed (`void completionSignal.catch(...)`). Orthogonal to the separate `skipVerification` email-OTP gate — doesn't bypass it.

**`startsWithBytes`**: pure behavior-preserving refactor (`Buffer.subarray().equals()` → generic byte comparison) fixing a real RN bug where `Buffer.subarray()` on a plain `Uint8Array` view doesn't return a `.equals()`-capable Buffer. Verified semantically identical for existing Buffer inputs.

**Testing**: `yarn workspace @vultisig/sdk test` — 3280/3280 pass (including the new derivePath/xpub/tssBatching tests). `yarn build` (SDK) clean. `yarn test:core` — 2612/2613 pass, 1 skipped (a transient failure on first run turned out to be resource contention from other concurrent test runs in this shared session, reran clean).

**Process note, not a blocker**: this is a fork PR (external contributor) and no CI workflows have actually run (`statusCheckRollup` is empty beyond CodeRabbit) — looks like it needs a maintainer to approve running Actions on it. Worth triggering before merge so it's not relying solely on my local verification.

Solid, well-scoped PR. The derivation math checks out by construction (non-hardened, same key tree, same normalization on both the export and signing sides) and there's now a real on-chain receipt backing it. Approving.

PR #1783, opened 2026-08-06
fix(cli): record the signing summary on the --yes path

neavra:
When `agent ask --yes` signs and broadcasts, nothing recorded what was actually signed. The only machine-readable statement of the payload was the tool input, visible solely with `--verbose`, and it never named the token contract; the result carried a bare hash.

A vault holding two same-symbol tokens could have one of them sent with no surface, at any point, saying which.

This emits the existing signing summary on the `--yes` path in both JSON and human output, without `--verbose`. It names the token contract for non-native sends. The summary is the same string the refusal path already produces, so there is one formatter rather than two that can drift.

No change to what gets signed, to the signing gate, to control flow, or to exit codes. `--yes` remains non-interactive.

## Summary by CodeRabbit

* **New Features**
  * Added auditable signing summaries to CLI agent output after approved signing requests.
  * Signing records now report the action, summary, chain, and whether signing succeeded.
  * Added signing-record events for JSON, human-readable, and interactive output.
  * Token transfer summaries now include the token contract when relevant.

* **Bug Fixes**
  * Signing records are preserved in partial results and error responses, including post-broadcast failures.
  * Declined requests no longer emit signing records.


gomesalexandre (approved):
lgtm 🫡

Reviewed remotely via gh only, no local checkout touched. This is a tight, well-tested audit-surface fix and I'd stamp it.

what I verified:
- `clients/cli/src/agent/session.ts` now captures the summary + chain at approval time, then emits the signing record only after `body()` runs, carrying the real `recent.success` bit. That fixes the dangerous shape where an approved-but-failed body could look like a completed signing.
- The declined path emits no signing record.
- `clients/cli/src/agent/executor.ts` now sources the token contract from `txArgs.tx.to`, i.e. the payload that actually gets signed, instead of from label text.
- The record is wired through all four surfaces: JSON envelope, human stdout/stderr, pipe `signing_record` event, and TUI render.
- The new tests meaningfully cover ordering, omission on decline, success vs post-broadcast failure, and the contract-address extraction path.

CI note: the red `Lint and Type Check` / `Test Summary` is not caused by this diff. The failing static gate is the `js-yaml` CVE-2026-59870 audit hit in transitive dev deps, while the jobs that actually exercise this PR's behavior passed.

good ship once the unrelated static gate is dealt with.

PR #1792, opened 2026-08-07
chore(mpc-native): cut the changeset to release the unpublished sdk 56 alignment

### LINKED ISSUE #1791: mpc-native: the SDK 56 expo/config-plugins correction has never been published, npm 0.1.4 still declares expo ^55 and config-plugins ^9
`@vultisig/mpc-native` was corrected for Expo SDK 56 back in June and the correction has never reached npm. The published tarball still declares the SDK 55 peer and the pre-realignment `@expo/config-plugins` line, and consumers resolve against that, not against `main`.

## what is published vs what is on main

Same version number, different declarations:

| field | npm `0.1.4` | `main` today |
|---|---|---|
| `peerDependencies.expo` | `^55.0.0` | `^56.0.8` |
| `dependencies.@expo/config-plugins` | `^9.0.0` | `^56.0.8` |

`0.1.4` is the newest published version (`0.1.1` through `0.1.4` exist, nothing since) and it went out on 2026-04-18. The corrections landed after:

- `a39fc2231` #570 moved `@expo/config-plugins` `55.0.8` to `56.0.8`
- `0f99b4c47` #632 moved the `expo` peer `^55.0.0` to `^56.0.8`, on 2026-06-05

Verified the ordering rather than reading it off the dates: `git merge-base --is-ancestor f76727656 0f99b4c47` is true and the reverse is false, so the `0.1.4` version bump is strictly an ancestor of the peer fix.

## why it never shipped

Both corrections arrived as dependabot PRs, and dependabot does not cut changesets. Changesets only versions a package that some changeset names, so every release run since has walked straight past this one. There have been at least five, the most recent `d25cd8851` yesterday, and `packages/mpc-native/package.json` is still on `0.1.4`. There is no pending changeset naming it either: five changesets sit in `.changeset/` on main and none mentions this package.

So this is not a lag that clears itself. It stays stuck until someone cuts a changeset by hand.

## what a consumer actually resolves

In `vultiagent-app`, which pins `"@vultisig/mpc-native": "0.1.4"` exactly:

```
       56.0.14  node_modules/@expo/config-plugins
        56.0.7  node_modules/@expo/config-types
         9.1.7  node_modules/@vultisig/mpc-native/node_modules/@expo/config-plugins
        53.0.5  node_modules/@vultisig/mpc-native/node_modules/@expo/config-types
```

The app hoists the SDK 56 config-plugins for everything else. This package gets its own nested copy off the old `9.x` line, dragging `@expo/config-types@53.0.5` with it. A `^9.0.0` caret can never float onto the `56.x` line, since Expo realigned config-plugins onto SDK versioning, so the nesting is permanent rather than a resolution accident.

## why that is worth fixing rather than filing as cosmetic

`app.plugin.js` is not a metadata shim. It is a real prebuild mod that rewrites `android/app/build.gradle`:

```js
implementation files("$rootDir/../<...>/dkls-release.aar")
implementation files("$rootDir/../<...>/goschnorr-release.aar")
```

That injection is what puts the dkls and goschnorr JNI `.so` files on the final APK runtime classpath. It is the load path for MPC keygen and signing. Right now it executes under a `@expo/config-plugins` major from the SDK 53 line while the prebuild driving it is the SDK 56 line, and if that injection ever silently no-ops or lands in the wrong module the symptom is an `UnsatisfiedLinkError` at first signing call, not a build failure. That failure mode is documented in this package's own `0.1.3` changelog entry.

## direction of harm, stated honestly

This is latent, not live. The app builds and ships on SDK 56 today with the nested plugin, so I am not reporting a current break and I have not run a prebuild or a native build to claim otherwise. What is wrong today is that the published metadata is untrue: it advertises an SDK the package is no longer built against, and it pulls a plugin toolchain three SDK lines behind the one it runs inside.

Two things make it worth doing now rather than later:

1. **Nothing surfaces it.** `vultiagent-app` sets `legacy-peer-deps=true` in `.npmrc` for an unrelated optional-peer reason, which switches off peer checking repo wide. So the unsatisfied `expo ^55.0.0` produces no warning, no `ERESOLVE`, nothing. I only found it by walking every peer edge in the lockfile.
2. **SDK 57 is already queued** in `vultisig/vultiagent-app#2007`. Going into that with a signing module whose published peer is two majors behind makes the upgrade harder to reason about than it needs to be.

## fix

Cut a patch changeset for `@vultisig/mpc-native` and let the next release run publish `0.1.5`. No code changes, since `main` already carries the correct ranges.

Adoption stays a separate deliberate call: `vultiagent-app` pins the version exactly, so publishing `0.1.5` changes nothing there until someone bumps it on purpose, and that bump wants a real Android build plus a signing run rather than a CI read.

Worth considering separately whether dependabot touching `peerDependencies` or `dependencies` on a published package should be gated differently from a plain devDependency bump. `#632` was labelled `chore(deps-dev)` and moved a peer range and a runtime range along with the devDependency, which is how an intended correction ended up unreleasable without anyone noticing.

## what I did not check

Read from `origin/main` and from the resolved lockfile, not from a prebuild run. I did not run `expo prebuild`, did not build the Android app, and have no runtime evidence that the nested plugin misbehaves. The claim here is that the published metadata is wrong and unreleased, which is verified, not that signing is currently broken, which is not.

🤖 found via Claude Code (local /loop sweep)

gomesalexandre:
closes #1791

`@vultisig/mpc-native` was corrected for Expo SDK 56 in June and the correction never reached npm. This cuts the changeset so it can.

## what is broken

Same version number, two different sets of declarations:

| field | npm `0.1.4` | `main` today |
|---|---|---|
| `peerDependencies.expo` | `^55.0.0` | `^56.0.8` |
| `dependencies.@expo/config-plugins` | `^9.0.0` | `^56.0.8` |

`0.1.4` published on 2026-04-18. `#570` moved `@expo/config-plugins` onto the `56.x` line and `#632` moved the `expo` peer, both after that. Confirmed the ordering with `git merge-base --is-ancestor` rather than trusting the dates: the `0.1.4` version bump is strictly an ancestor of the peer fix, and the reverse is false.

Both corrections arrived as dependabot PRs. Dependabot does not cut changesets, and changesets only versions a package some changeset names, so all five release runs since have walked past this one. The most recent was yesterday. Nothing in `.changeset/` names this package today, so it stays stuck until someone does this by hand.

## what it costs downstream

`vultiagent-app` pins `0.1.4` exactly and resolves:

```
       56.0.14  node_modules/@expo/config-plugins
        56.0.7  node_modules/@expo/config-types
         9.1.7  node_modules/@vultisig/mpc-native/node_modules/@expo/config-plugins
        53.0.5  node_modules/@vultisig/mpc-native/node_modules/@expo/config-types
```

The app hoists SDK 56's config-plugins for everything else, and this package alone nests the old `9.x` line plus SDK 53 config types. A `^9.0.0` caret cannot float onto `56.x` since Expo realigned config-plugins onto SDK versioning, so it is permanent rather than a resolution accident.

That nested copy is what runs `app.plugin.js`, which is a real prebuild mod injecting the `dkls-release.aar` and `goschnorr-release.aar` `implementation files(...)` lines into the app module. It is the load path for the MPC signing `.so` files.

Nothing surfaces it either: `vultiagent-app` sets `legacy-peer-deps=true` in `.npmrc` for an unrelated optional-peer reason, so the unsatisfied peer produces no warning at all. I found it by walking every peer edge in the lockfile.

## how

One file. A patch changeset naming `@vultisig/mpc-native`. `main` already carries the correct ranges, so there is nothing else to change.

## risk

Low, and adoption is deliberate rather than automatic.

`no runtime changes` in this repo: the diff is a single markdown file and touches no source. Publishing `0.1.5` does not move `vultiagent-app`, which pins the version exactly, so the app keeps resolving `0.1.4` until someone bumps it on purpose. That bump is the step that wants a real Android build and a signing run, and it belongs in its own PR rather than riding this one.

I did not run a prebuild or an Android build here. The claim is that the published metadata is wrong and unreleased, which is verified, not that signing is currently broken, which I have not shown.

🤖 found via Claude Code (local /loop sweep)


NeOMakinG (approved):
LGTM

PR #1794, opened 2026-08-07
fix(ripple): state TrustSet on the wire so it co-signs with iOS

### LINKED ISSUE #1793: [BUG] XRPL TrustSet diverges against iOS: RippleSpecific.transaction_type missing from generated protos
**Describe the bug**

An XRPL trust-line activation initiated from a client that uses this SDK cannot be co-signed by an iOS device: the ceremony diverges and never completes. No funds move, but the trust line cannot be opened at all in a mixed committee.

The cause is that this SDK's generated protos are stale relative to commondata. `commondata` already carries the discriminator that resolves the ambiguity (merged in vultisig/commondata#100):

| | commondata `main` | `@vultisig/core-mpc` 1.17.0 |
|---|---|---|
| `RippleSpecific.transaction_type` | present (field 5) | **absent** |
| `TRANSACTION_TYPE_RIPPLE_TRUST_SET` | present (= 10) | **absent** (enum stops at 9) |

A non-native Ripple coin is ambiguous on its own: the same `(currency, issuer)` pair means either "open a trust line for this token" (TrustSet, where the keysign amount is the trust-line LIMIT) or "send this token" (Payment with a CurrencyAmount, where it is a transfer amount). The two produce different signed bytes.

Because the field never reaches the wire, the two platforms fall back to opposite defaults:

- **this SDK** — a non-native Ripple coin always builds a TrustSet (`getTrustSet` in `keysign/signingInputs/resolvers/ripple.ts`), which is why issued-currency Payments are not expressible here at all;
- **iOS** — a non-native coin with no discriminator builds an issued-currency **Payment**, reserving TrustSet for the explicit `transaction_type`.

So a TrustSet originated here is read by an iOS co-signer as a token Payment. Different signing inputs, different pre-signing hashes, ceremony fails.

**To Reproduce**

1. Initiate an XRPL trust-line activation from a client on `@vultisig/core-mpc` 1.17.0 (e.g. vultisig-windows "Open Trust Line").
2. Pair an iOS device (post vultisig/vultisig-ios#4940) as co-signer.
3. Run the keysign ceremony.
4. It never completes.

**Expected behavior**

A trust-line activation is signed identically by every device in the committee. The payload states which XRPL operation it describes rather than leaving each signer to infer it from the coin.

**Additional context**

The direction matters and narrows the fix: **iOS → SDK already works**, because iOS sets the field *and* uses a non-native coin, and this SDK ignores the unknown field and infers TrustSet from the coin — both build a TrustSet. Only **SDK → iOS** breaks, since that is the direction that meets iOS's Payment default. The failure is fail-safe: the ceremony diverges rather than signing the wrong operation.

The commondata proto comment already prescribes the transitional contract:

> a signer that predates this field infers TrustSet from a non-native coin, so a TrustSet stays byte-identical across mixed-version committees, while a token Payment diverges and the ceremony fails without moving funds. Fail-safe by design, and resolved as platforms adopt.

So the signing side must **prefer the field and keep coin-inference as the fallback**, or this SDK stops being byte-compatible with the clients already shipping against it.

Regenerating is mechanical — `scripts/regenerate-core-mpc-protos.mjs` (`COMMONDATA_DIR=../commondata yarn proto:regen:core-mpc`) — but the resolvers then need to write and read the field.

Client-side report: vultisig/vultisig-windows#4579
iOS counterpart: vultisig/vultisig-ios#4940

Ehsan-saradar:
Closes #1793.

## Problem

An XRPL trust-line activation originated through this SDK could not be co-signed by an iOS device — the ceremony diverged and never completed. Fail-safe (no funds move), but a trust line could not be opened at all in a mixed committee. Reported against the client in vultisig/vultisig-windows#4579.

A non-native Ripple coin is ambiguous on its own: the same `(currency, issuer)` pair means either *open a trust line for this token* (TrustSet, where the keysign amount is the **limit**) or *send this token* (Payment with a CurrencyAmount, where it is a **transfer amount**). The two sign different bytes.

commondata already carries the discriminator that settles this (`RippleSpecific.transaction_type` + `TRANSACTION_TYPE_RIPPLE_TRUST_SET`, vultisig/commondata#100) — but the generated protos here were stale, so the field never reached the wire and each platform fell back to its own default:

- **this SDK** — non-native coin ⟹ always TrustSet
- **iOS** — non-native coin with no discriminator ⟹ Payment

So a TrustSet built here was read by an iOS co-signer as a token Payment.

## What changed

1. **Regenerated** `blockchain_specific_pb.ts` from commondata, which adds `RippleSpecific.transaction_type` and `TRANSACTION_TYPE_RIPPLE_TRUST_SET = 10`.
2. **Set the field** when building a TrustSet (`chainSpecific/resolvers/ripple.ts`).
3. **Prefer the field** when signing (`signingInputs/resolvers/ripple.ts`).

The shape test and the signing decision share one definition (`isRippleTrustSet` / `hasRippleTrustSetShape`) so the builder and signer cannot drift — if they did, a payload could be built as one operation and signed as the other.

## The fallback is deliberate, and load-bearing

Coin-shape inference is **kept** as the fallback rather than replaced. Clients already shipped infer TrustSet from a non-native coin alone, so honouring that is what keeps a TrustSet byte-identical across a mixed-version committee. A test asserts that parity on the encoded signing input, and I verified it bites: making the signer field-only fails 5 tests, including the byte-parity one.

Payloads that are **not** TrustSets are untouched — native XRP and verbatim `signRipple` dApp transactions leave the field unset (`UNSPECIFIED`, absent from the wire), so their signed bytes are unchanged. Both are covered by tests.

## Testing

`yarn typecheck`, `yarn format:check`, `yarn check:shared-exports`, the SDK bundle changeset guard and eslint all pass. `yarn test:core` — **2553 passed**, including 7 new tests:

- builds a TrustSet from the explicit discriminator
- signs identically with and without it, and is **byte-identical across a mixed-version committee**
- native XRP stays a Payment
- the builder marks an issued-currency payload, leaves native XRP and dApp `signRipple` unmarked

Not verified here: the actual cross-device ceremony against an iOS build. That needs a paired device and is worth confirming once this is released and the client bumps.

## Note on the diff

`blockchain_specific_pb.ts` is generator output. Buf emits it unformatted, so I ran the repo's Prettier config over it to match the committed style — without that the diff is ~330 lines of pure formatting churn instead of the 34 real lines.

Also worth flagging: `scripts/regenerate-core-mpc-protos.mjs` fails on the current buf (1.72.0) — `--path` resolves against cwd rather than the input root, so every path lands "outside the context directory". I regenerated by invoking buf from the proto root instead. The script needs a fix, but that felt out of scope here; happy to file it separately.

## Summary by CodeRabbit

* **New Features**
  * Ripple TrustSet transactions are now explicitly identified during signing.
  * Trust-line operations are distinguished from native XRP payments and verbatim dApp transactions.
  * Added a public utility for detecting Ripple TrustSet transactions.
* **Bug Fixes**
  * Improved handling of issued-currency TrustSet payloads.
  * Preserved compatibility with existing payloads through fallback detection.
  * Added validation for unsupported native XRP or incomplete TrustSet data.


gomesalexandre (approved):
lgtm 🫡

Reviewed the exact head `80b19bdb` against fresh `origin/main`.

what I verified:
- `packages/core/mpc/keysign/chainSpecific/resolvers/ripple.ts` now states the XRPL operation on the wire by setting `RippleSpecific.transactionType = RIPPLE_TRUST_SET` only for the TrustSet shape. Native XRP and ordinary payloads stay unset.
- `packages/core/mpc/keysign/utils/isRippleTrustSet.ts` keeps one shared definition for the builder and signer: explicit `transactionType` wins, but the legacy non-native-coin shape still falls back to TrustSet. That fallback is the load-bearing mixed-version compatibility path.
- `packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts` honors the new discriminator while preserving the old inference, so a newer signer and an older signer still build byte-identical TrustSet signing inputs instead of diverging mid-ceremony.
- The raw `signRipple` path is still fail-closed and verbatim. This PR does not start reclassifying dApp-supplied XRPL transactions into TrustSets.
- The new tests pin the important boundaries: explicit TrustSet discrimination, inferred fallback parity, native XRP staying a Payment, raw dApp transactions staying undiscriminated, and byte-parity across the mixed-version path.

local receipts:
- `codex exec` conceptual pass on the exact checked-out PR head returned `VERDICT: APPROVE`.
- `yarn build:sdk` currently fails on this host on both the PR head and pristine `origin/main` with the same pre-existing error in `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts` (`auxiliaryData` not in `ISigningInput`), so I did not treat that as a regression from this diff.
- Focused Vitest from a borrowed install in the disposable clone hit temp-clone export/install-state issues before test execution, so I did not get a trustworthy local runtime receipt from that path.

Given the shared-shape fallback plus the new explicit wire discriminator, I do not see a correctness blocker here.

PR #1805, opened 2026-08-08
fix(cosmos): recover ambiguous raw broadcast responses

### LINKED ISSUE #1804: fix(cosmos): raw broadcast should recover lost transport responses via deterministic tx hash
## Summary
Public SDK raw Cosmos broadcasts still fail hard on ambiguous transport/proxy failures instead of recovering via the deterministic tx hash.

## Evidence
- `packages/sdk/src/vault/services/RawBroadcastService.ts`
  - the raw Cosmos path already handled duplicate-cache recovery, but generic `broadcastTx()` transport failures still surfaced as total failure.
- `clients/cli/src/commands/broadcast.ts`
  - CLI inherits this public SDK surface via `vault.broadcastRawTx()`.
- parallel consumer evidence:
  - the app already carries the same lost-response defense locally in `src/services/cosmosTx.ts`, which shows this logic has already drifted into a consumer instead of living centrally in the SDK.

## Impact
A lost HTTP response or proxy error can be reported as "failed" even if the signed bytes already reached the node. That invites a blind re-broadcast / re-sign with a fresh sequence, which is exactly the double-send class this audit campaign keeps hunting.

## Fix sketch
- keep the deterministic Cosmos tx hash as the recovery anchor
- on ambiguous `broadcastTx()` transport failure, poll `getTx(hash)` briefly before declaring failure
- return success only for a verified success receipt
- if the tx later resolves to `code != 0`, fail closed
- if the tx never becomes queryable, throw a "verify this hash before retrying" message

## Test shape
CLI-testable via `vault.broadcastRawTx()` / `vultisig broadcast`, with unit coverage for:
- recovered success
- recovered on-chain failure
- never-indexed unknown status

## Artifact
https://gist.github.com/gomesalexandre/7e5aa78610ab998ec1c69ba5e4895fb7

gomesalexandre:
## Summary
- recover ambiguous Cosmos raw broadcast transport failures via deterministic tx-hash lookup
- fail closed if the later receipt shows execution failure or the tx never becomes queryable
- add regression coverage for recovered success, recovered failure, and never-indexed unknown status

closes #1804

## Receipts
- ✅ `corepack yarn workspace @vultisig/sdk vitest run tests/unit/vault/services/RawBroadcastService.test.ts`
  - 52 tests passed
- ⚠️ `corepack yarn build:sdk`
  - blocked by a pre-existing mainline issue unrelated to this diff:
  - `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts(72,5): error TS2353: Object literal may only specify known properties, and 'auxiliaryData' does not exist in type 'ISigningInput'.`

## Artifact
https://gist.github.com/gomesalexandre/7e5aa78610ab998ec1c69ba5e4895fb7

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Cosmos transaction broadcasting when the initial result is unclear or interrupted.
  - Automatically checks the transaction hash to determine whether a transaction succeeded after cache, network, proxy, or RPC errors.
  - Retries verification briefly when transaction details are not immediately available.
  - Provides clearer failure information and retry guidance when the transaction cannot be verified.
  - Detects and reports transactions that were found but failed during execution.


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because the exact head treats definitive Cosmos CheckTx rejections as ambiguous transport failures and omits the hash needed for safe cache-recovery verification; send a current-main patch that recovers only timeout or transport errors and preserves BroadcastTxError.

PR #1806, opened 2026-08-08
fix(swap): let secured assets be used in limit swaps

### LINKED ISSUE #1778: Secured assets cannot be used in limit swaps: memo validation, deposit branch, and the deposited asset itself
## Problem

A THORChain **secured asset** cannot be used in a limit swap. Not as the source, not as the target — placement is impossible, and an order that somehow existed could not be cancelled either.

This is not one bug. There are three independent failures stacked behind each other, and each only becomes visible once the one in front of it is cleared. That is why it has looked like "secured assets aren't supported yet" rather than a set of defects.

## 1. The memo builder rejects secured notation

`buildLimitSwapMemo` validates both legs through `getSupportedThorchainAssetChain`, which routes through `assertValidPoolId` — the shared THORChain **pool-id** grammar:

```
POOL_ID_RE = /^[A-Z0-9]+\.[A-Z0-9]+(-[A-Z0-9]+)?$/
```

Secured denoms carry no dot, so every one of them is rejected:

| asset | accepted |
|---|---|
| `ETH.USDC-0XA0B8…` | yes |
| `XRP-XRP` | **no** |
| `ETH-USDC-0XA0B8…` | **no** |

`getThorchainMemoAsset` already emits the correct secured spelling — its own docstring notes that the memo builder will not accept the value it returns. So the two halves of this package disagree with each other, and have for some time.

The same call is what `getLimitSwapCancelEligibility` uses to size a cancel memo against its source chain, so a secured-funded order also reports `unroutableSourceChain` and offers no cancel action.

Note this validator is asking a **limit-swap** question ("is this an asset this memo can carry") through a **pool-id** validator shared with the LP paths. Those are not the same question, and the chain it resolves is also used to decide the memo byte budget and to validate the payout address — both of which have a different answer for an asset custodied on THORChain than for one that merely originates there.

## 2. The placement builder only recognises RUNE as a deposit

`buildLimitSwapKeysignPayload` decides between "deposit on THORChain" and "transfer to an Asgard inbound" with:

```ts
const isRuneDeposit = areEqualCoins(fromCoin, chainFeeCoin[Chain.THORChain])
```

A secured asset sits on THORChain but is not RUNE, so it takes the transfer branch and looks up a THORChain inbound vault. There is none — THORChain deposits have no inbound — so the build throws before anything can be signed.

## 3. The deposit references an asset the vault does not hold

This is the one that reaches the chain and costs a fee.

With (1) and (2) cleared locally, a secured-BTC → secured-USDC limit order broadcast and was **rejected on-chain**:

> tx `73D17648DF9BDE0B6B973EC71771BABAAEE7C0A63C662091B6BC052B1AEA69BF`, height 27315893
> `Code: 5; Raw log: failed to execute message; message index: 0: insufficient funds`

The memo was correct. The **deposited coin** was not:

```json
{
  "coins": [{ "asset": "THOR.BTC", "amount": "1000", "decimals": "8" }],
  "memo": "=<:ETH-USDC-0xa0b8…:thor12a9…:64974295/14400/0:v0:50"
}
```

`THOR.BTC` does not exist. The vault's actual balance is `btc-btc` (and the account held ample RUNE for fees), so THORChain read a request for an asset with a zero balance and refused.

The cosmos signing resolver already knows how to build a secured deposit asset — but it derives it exclusively from `swapPayload.fromCoin`. A limit order carries no swap payload on the THORChain branch, so it falls through to a chain-prefix + ticker construction and produces `THOR.<TICKER>`.

**This affects every secured denom, not just BTC** — `eth-eth`, `gaia-atom`, `eth-usdc-0x…` and the rest all reduce to a `THOR.*` asset that no vault holds. Market swaps are unaffected, because they do carry a swap payload, which is why this has stayed hidden.

Worth noting the two coin shapes involved spell the denom differently (`contractAddress` on a swap payload's coin, `id` on the payload's own coin), so a change here typechecks while silently reading nothing.

## Acceptance criteria

- [ ] A limit order can be placed with a secured asset as the source, as the target, and on both sides
- [ ] The memo byte budget and payout-address validation use the chain the asset is actually **custodied** on, not the one it originates from
- [ ] The broadcast deposit references the asset the vault genuinely holds, for every secured denom — verified against a real balance, not only in a unit test
- [ ] A secured-funded order is cancellable, rather than reporting an unroutable source chain
- [ ] THORChain-native tokens (`tcy`, `x/…`) and RUNE are unchanged — they are not secured assets
- [ ] Asset flavours that remain unsupported fail with a message that says so, rather than a pool-id validation error

## Out of scope

Synth (`BTC/BTC`) and trade (`ETH~ETH`) assets. Different custody model, and their behaviour through the advanced swap queue has not been established.

A companion defect on the client side — the reference-price probe omits the payout destination, which THORChain rejects for a secured target — is being handled in vultisig-windows and is not part of this issue.

Ehsan-saradar:
## Description

Secured assets can be used in limit swaps.

Fixes #1778

Three independent defects stood between a secured asset and a placeable order, each hidden behind the one in front of it — which is why this looked like an unimplemented feature rather than a set of bugs. The first two refuse before anything is signed; the third reaches the chain and costs a fee.

### 1. The memo builder rejected the notation

`buildLimitSwapMemo` validated both legs through `assertValidPoolId` — the shared THORChain **pool-id** grammar, which only understands dotted `CHAIN.ASSET`:

| asset | before | now |
|---|---|---|
| `ETH.USDC-0XA0B8…` | accepted | accepted |
| `XRP-XRP` | **rejected** | accepted |
| `ETH-USDC-0XA0B8…` | **rejected** | accepted |
| `BTC/BTC`, `ETH~ETH` | rejected as a malformed pool id | rejected, and told why |

Meanwhile `getThorchainMemoAsset` was already emitting exactly that spelling — its own docstring conceding the memo builder would not accept the value it returned. The two halves of the package had disagreed for some time.

A limit swap now asks its own, narrower question — *is this an asset this memo can carry* — instead of borrowing a validator shared with the LP paths. Nothing changes for those callers.

**That answer is not cosmetic.** It decides the memo's byte budget and which chain the payout address is validated against. A secured asset is custodied on THORChain wherever it originates, so both answer THORChain; reading its home chain instead sized the deposit against the wrong budget and rejected the only correct payout address.

Synth and trade assets stay unsupported — a different custody model whose behaviour through the advanced swap queue has not been established. They now say so rather than failing as malformed pool ids.

### 2. The placement builder recognised only RUNE as a deposit

```ts
const isRuneDeposit = areEqualCoins(fromCoin, chainFeeCoin[Chain.THORChain])
```

A secured asset sits on THORChain but is not RUNE, so it took the transfer branch and looked up a THORChain Asgard inbound. There is none — THORChain deposits have no inbound — so the build threw. Every THORChain-held source now deposits.

### 3. The deposit referenced an asset the vault does not hold

This is the one that reached the chain. With (1) and (2) cleared locally, a secured-BTC → secured-USDC order broadcast and was **rejected on-chain**:

> tx `73D17648DF9BDE0B6B973EC71771BABAAEE7C0A63C662091B6BC052B1AEA69BF`, height 27315893
> `Code: 5; failed to execute message; message index: 0: insufficient funds`

The memo was correct. The **deposited coin** was not:

```json
{ "coins": [{ "asset": "THOR.BTC", "amount": "1000", "decimals": "8" }] }
```

`THOR.BTC` does not exist. The vault's balance is `btc-btc`, and it held ample RUNE for fees — so THORChain read a request for an asset with a zero balance and refused.

The cosmos resolver already knew how to build a secured deposit asset, but derived it exclusively from `swapPayload.fromCoin`. A limit order carries no swap payload on the THORChain branch, so it fell through to a chain-prefix + ticker construction. It now keys off the coin the deposit actually spends.

Worth noting the two coin shapes spell the denom differently — `contractAddress` on a swap payload's coin, `id` on the payload's own — which is what let the previous version typecheck while reading nothing.

**Every secured denom was affected, not just BTC.** Market swaps were not, because they do carry a swap payload — which is exactly why this stayed hidden.

## Which feature is affected?

- [x] Swap — THORChain limit orders (`=<`), secured assets

**Verified on mainnet.** A secured limit swap was placed successfully from a vault holding `btc-btc`, after the same order failed with `insufficient funds` before the fix. The failing transaction is quoted above; the deposit now references the asset the vault actually holds.

## Parity

- iOS: filed — the same three defects exist there. `isValidAssetFormat` requires exactly two dot-separated parts, so secured denoms are rejected as malformed before the equivalent of (1); the notation layer already emits them correctly.
- Android: n/a — limit swaps are not implemented
- SDK: first — this is the implementation

## Checklist

## Additional context

**Tests exercise the real signing path**, not a reimplementation. `getCosmosSigningInputs` is driven with a payload per denom the test vault actually holds (`btc-btc`, `eth-eth`, `gaia-atom`, `eth-usdc-0x…`), asserting each builds its own secured asset — and that `tcy` and `x/ruji` keep the `THOR.TICKER` form, since THORChain-native tokens are **not** secured assets. Every one of those would have failed before this change.

Also pinned: a secured source sizes its memo against THORChain rather than its origin chain; a secured target requires a THOR payout address; and a secured target from a UTXO source is refused, because a secured denom plus a 43-character THOR address overflows the 80-byte `OP_RETURN` and nothing in the memo can be shortened.

**One previously-pinned test was inverted rather than deleted.** `thorchainMemoAsset.test.ts` carried a test asserting the memo builder would *not* accept the secured notation this helper emits — pinned deliberately so the day validation learned secured denoms, it would fail and force an update rather than let the limitation persist silently. It now asserts the opposite.

**Not covered here:** synth and trade assets, per the reasoning above.

A companion client-side defect — the reference-price probe omitted the payout destination, which THORChain rejects for a secured target — is handled in vultisig-windows and is not part of this PR.

## Summary by CodeRabbit

* **New Features**
  * Added secured-asset support for limit swaps.
  * Secured assets are identified, validated, and routed through THORChain.
  * Limit swaps can use secured assets as direct deposit sources, including without swap details.
  * Added support for secured-asset formats across supported chains and denominations.

* **Bug Fixes**
  * Improved handling of malformed, unsupported, synthetic, trade, and unroutable assets.
  * Preserved correct behavior for native THORChain assets and RUNE.
  * Added safeguards for paused trading, invalid payout addresses, and oversized transactions.


packages/core/chain/swap/native/limitSwapMemo.ts
```diff
@@ -57,15 +56,89 @@ const getAssetChainPrefix = (asset: string): string => {
...
+ * behaviour nobody has established.
+ */
 const getSupportedThorchainAssetChain = (asset: string, fieldName: string): Chain => {
-  assertValidPoolId(asset)
+  const normalized = typeof asset === 'string' ? asset.trim() : ''
+  if (!normalized) {
+    throw new Error(`${fieldName} must be a non-empty THORChain asset`)
+  }
+  if (normalized !== asset) {
+    throw new Error(`${fieldName} must not contain surrounding whitespace`)
```
gomesalexandre, line 105: fixed on current head in 84b350d2a by rejecting surrounding whitespace before memo construction, plus a regression in limitSwapMemo.test.ts.

johnnyluo (approved):
LGTM

PR #1810, opened 2026-08-09
feat(swap): normalize swap arrival status

### LINKED ISSUE #1231: feat: move swap-arrival status normalization into the sdk (both consumers hand-roll 500-line pollers)
Both major consumers hand-roll ~500-line thor/maya swap-arrival status pollers next to the sdk's link-only helper:

- mobile app: crossChainSwapPoller (599 lines - thornode /thorchain/tx/status + midgard + mayanode + lifi, normalizes thor stages)
- agent backend: get-swap-arrival-status tool (516 lines - midgard.thorchain.network / midgard.mayachain.info + skip go /v2/tx/status, normalizes to pending|success|refunded|not_found|error)
- sdk: only `packages/core/mpc/swap/utils/getSwapTrackingUrl.ts` (builds explorer links, no status logic)

Two independently-maintained state machines for the same upstream apis, on different host sets, with different normalized vocabularies. The app side already had the dead-hosts class of bug once (poller pointed at unreachable endpoints and swaps looked stuck) - this duplication is how that class recurs.

## fix shape

Extract a `getSwapArrivalStatus()` into the sdk next to getSwapTrackingUrl: provider-aware (thor/maya/skip/lifi), one normalized stage vocabulary, injectable host config + fetch. Needs a real interface-design pass first (the two consumers normalize differently today - the union of their needs is the spec). Then both consumers shrink to thin adapters.

effort: L / risk: MED (swap-status semantics; port with a fixture suite recorded from both current impls) / confidence: HIGH on the duplication, MED on interface shape

rcoderdev:
Closes #1231

## Summary by CodeRabbit

- **New Features**
  - Added unified swap-arrival status tracking across THORChain, MayaChain, Skip Go, and LI.FI.
  - Normalizes pending, successful, refunded, failed, partial, and not-found outcomes.
  - Includes destination transaction details, progress stages, messages, and error information.
  - Exposes the functionality through the SDK and `Vultisig` interface.
  - Supports terminal-status detection and configurable service endpoints.

- **Bug Fixes**
  - Improved handling of partial completions, invalid statuses, and node-only completion or refund scenarios.

- **Tests**
  - Added comprehensive coverage for provider responses, error handling, malformed data, and public exports.

### DIFF
--- a/.changeset/swap-arrival-status-terminal-fixes.md
+++ b/.changeset/swap-arrival-status-terminal-fixes.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Fix three unsafe terminal-state gaps in `getSwapArrivalStatus`: LI.FI `DONE/PARTIAL` (user received a different token than requested) is now reported as a distinct `partial` status instead of being flattened into `success`; malformed or unknown Midgard action statuses now throw `SwapArrivalStatusRequestError` instead of being reported as a terminal `error`; and THOR/Maya node-only completion now reads THORNode's `planned_out_txs[].refund` flag to report `success`/`refunded` when available, instead of always forcing an artificial `pending` state.
--- a/.changeset/swap-arrival-status.md
+++ b/.changeset/swap-arrival-status.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': minor
+'@vultisig/sdk': minor
+---
+
+Add provider-aware swap arrival status normalization for THORChain, MayaChain, Skip Go, and LI.FI.
--- a/packages/core/chain/package.json
+++ b/packages/core/chain/package.json
@@ -2033,6 +2033,11 @@
       "import": "./dist/swap/SwapFee.js",
       "default": "./dist/swap/SwapFee.js"
     },
+    "./swap/utils/getSwapArrivalStatus": {
+      "types": "./dist/swap/utils/getSwapArrivalStatus/index.d.ts",
+      "import": "./dist/swap/utils/getSwapArrivalStatus/index.js",
+      "default": "./dist/swap/utils/getSwapArrivalStatus/index.js"
+    },
     "./swap/utils/getSwapExplorerUrl": {
       "types": "./dist/swap/utils/getSwapExplorerUrl/index.d.ts",
       "import": "./dist/swap/utils/getSwapExplorerUrl/index.js",
--- a/packages/core/chain/swap/utils/getSwapArrivalStatus/fixtures.json
+++ b/packages/core/chain/swap/utils/getSwapArrivalStatus/fixtures.json
@@ -53,6 +53,14 @@
             "ibc_transfer": {
               "packet_txs": {
                 "send_tx": { "tx_hash": "SKIP-SOURCE" },
+                "receive_tx": { "tx_hash": "SKIP-INTERMEDIATE" }
+              }
+            }
+          },
+          {
+            "ibc_transfer": {
+              "packet_txs": {
+                "send_tx": { "tx_hash": "SKIP-INTERMEDIATE" },
                 "receive_tx": { "tx_hash": "SKIP-DESTINATION" }
               }
             }
--- a/packages/core/chain/swap/utils/getSwapArrivalStatus/fixtures.json
+++ b/packages/core/chain/swap/utils/getSwapArrivalStatus/fixtures.json
@@ -20,6 +20,32 @@
     },
     "out_txs": [{ "id": "THOR-REAL-OUT", "chain": "BTC" }]
   },
+  "thorchainNodeCompleteWithPlannedRefund": {
+    "tx": { "id": "THOR-REFUND-SOURCE" },
+    "stages": {
+      "inbound_observed": { "started": true, "completed": true },
+      "inbound_confirmation_counted": { "completed": true },
+      "inbound_finalised": { "completed": true },
+      "swap_status": { "pending": false },
+      "swap_finalised": { "completed": true },
+      "outbound_signed": { "completed": true }
+    },
+    "planned_out_txs": [{ "chain": "BTC", "refund": true }],
+    "out_txs": [{ "id": "THOR-REAL-OUT", "chain": "BTC" }]
+  },
+  "thorchainNodeCompleteWithPlannedSwapOutput": {
+    "tx": { "id": "THOR-SUCCESS-SOURCE" },
+    "stages": {
+      "inbound_observed": { "started": true, "completed": true },
+      "inbound_confirmation_counted": { "completed": true },
+      "inbound_finalised": { "completed": true },
+      "swap_status": { "pending": false },
+      "swap_finalised": { "completed": true },
+      "outbound_signed": { "completed": true }
+    },
+    "planned_out_txs": [{ "chain": "BTC", "refund": false }],
+    "out_txs": [{ "id": "THOR-REAL-OUT", "chain": "BTC" }]
+  },
   "thorchainMidgardPending": {
     "count": "1",
     "actions": [
@@ -41,6 +67,16 @@
         "in": [{ "txID": "THOR-REFUND-SOURCE" }],
         "out": [{ "txID": "THOR-REFUND-OUT" }],
         "metadata": { "refund": { "reason": "emit asset less than price limit" } }
+      }
+    ]
+  },
+  "thorchainMidgardMalformedSwapStatus": {
+    "count": "1",
+    "actions": [
+      {
+        "type": "swap",
+        "in": [{ "txID": "THOR-SOURCE" }],
+        "out": []
       }
     ]
   },
--- a/packages/core/chain/swap/utils/getSwapArrivalStatus/fixtures.json
+++ b/packages/core/chain/swap/utils/getSwapArrivalStatus/fixtures.json
@@ -8,6 +8,18 @@
       "swap_status": { "pending": true }
     }
   },
+  "thorchainNodeCompleteWithOutbound": {
+    "tx": { "id": "THOR-REFUND-SOURCE" },
+    "stages": {
+      "inbound_observed": { "started": true, "completed": true },
+      "inbound_confirmation_counted": { "completed": true },
+      "inbound_finalised": { "completed": true },
+      "swap_status": { "pending": false },
+      "swap_finalised": { "completed": true },
+      "outbound_signed": { "completed": true }
+    },
+    "out_txs": [{ "id": "THOR-REAL-OUT", "chain": "BTC" }]
+  },
   "thorchainMidgardPending": {
     "count": "1",
     "actions": [
--- a/packages/core/chain/swap/utils/getSwapArrivalStatus/fixtures.json
+++ b/packages/core/chain/swap/utils/getSwapArrivalStatus/fixtures.json
@@ -1,0 +1,87 @@
+{
+  "thorchainNodePending": {
+    "tx": { "id": "THOR-SOURCE" },
+    "stages": {
+      "inbound_observed": { "started": true, "completed": true },
+      "inbound_confirmation_counted": { "completed": true },
+      "inbound_finalised": { "completed": true },
+      "swap_status": { "pending": true }
+    }
+  },
+  "thorchainMidgardPending": {
+    "count": "1",
+    "actions": [
+      {
+        "status": "pending",
+        "type": "swap",
+        "in": [{ "txID": "THOR-SOURCE" }],
+        "out": [],
+        "metadata": { "swap": { "memo": "=:BTC.BTC:bc1destination" } }
+      }
+    ]
+  },
+  "thorchainMidgardRefunded": {
+    "count": "1",
+    "actions": [
+      {
+        "status": "success",
+        "type": "refund",
+        "in": [{ "txID": "THOR-REFUND-SOURCE" }],
+        "out": [{ "txID": "THOR-REFUND-OUT" }],
+        "metadata": { "refund": { "reason": "emit asset less than price limit" } }
+      }
+    ]
+  },
+  "mayachainMidgardSuccess": {
+    "count": "1",
+    "actions": [
+      {
+        "status": "success",
+        "type": "swap",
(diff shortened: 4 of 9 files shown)


gomesalexandre (commented):
remote conceptual pass on current head `ac6cdb00` only. i did not approve this pass.

what i verified:
- the normalization logic is directionally right and the current head closes the coderabbit points around partial host overrides, malformed mixed Midgard arrays, and taking the final Skip hop’s destination tx hash.
- current CI on this head is green.
- the exported surface is real: `packages/core/chain/swap/utils/getSwapArrivalStatus/index.ts`, `packages/sdk/src/index.ts`, and `packages/sdk/src/Vultisig.ts` all line up.

blocking for me is the receipt level, not a new code-path bug from this remote read:

1. this is a new SDK public API surface and a user-visible swap-status contract, but the added public-surface coverage is only `packages/sdk/tests/unit/utils/publicExports.test.ts`, which asserts presence, not runtime behavior.
2. the repo’s own SDK review bar is higher for this class of change: build the SDK and exercise the changed path through the actual shipped public surface, with a real receipt. here that means a bundled/public invocation of `getSwapArrivalStatus` on representative THOR/Maya/Skip/LI.FI snapshots, not just unit fixtures inside `packages/core/chain`.
3. without that exact public-surface receipt, this is still a conceptual pass on a status-normalization API that downstream pollers will trust to decide success vs refunded vs pending.

so i’m leaving this as comment-only for now. once there is a real public-surface receipt on the PR body or comments, i’m happy to re-check quickly.

gomesalexandre (changes requested):
Re-reviewed on `befc34a40`. Thanks for the public-surface receipt, that discharges the block I left on `ac6cdb00` and I appreciate you turning it around.

This pass is the one I owed you: local checkout, real close read of the 451-line normalizer, and four execution probes rather than the conceptual read I did last time. The core logic is byte-identical to what I saw before (`git diff ac6cdb000 befc34a40 -- packages/core/chain/swap/utils/getSwapArrivalStatus` is empty, the head moved from the rebase onto `3ea8b2c51`), so everything below applies to both heads. I should have found it the first time round.

| check | result |
|---|---|
| must-pass: `getSwapArrivalStatus` suite | 28/28 pass |
| must-pass: sdk `publicExports` + RN entry | 44/44 pass |
| probe 1: THOR refund, Midgard down | **`success` / `complete`, terminal=true** |
| probe 2: freshly broadcast swap, not yet observed | `not_found`, terminal=true |
| probe 3: LI.FI `NOT_FOUND` on an unindexed swap | `not_found`, terminal=true |
| probe 4: real THORNode `out_txs: [{id}]` shape | `destinationTxHash` absent |

Credit where it is due first, because some of this is genuinely careful. Checking `substatus === 'REFUNDED'` **before** `status === 'DONE'` for LI.FI is the subtle one and you got it right: LI.FI reports a refund as `DONE` + `REFUNDED`, and the obvious ordering would silently call every LI.FI refund a success. The `transfer_asset_release.chain_id === chainId` test for Skip is the right way to tell a source-side refund from a destination failure. And the docblock reasoning is exactly the right principle:

> Network, HTTP and malformed-response failures throw `SwapArrivalStatusRequestError`; they are not converted to terminal `error` results because an upstream outage does not mean the swap itself failed.

That is precisely the correct instinct. My main finding is that the code violates that principle through a different door.

REQUEST_CHANGES on the two below.

---

### preferably-blocking: a refund becomes a terminal `success` whenever Midgard is unavailable

`getThorMayaNodeProgress` has **no refund branch at all**. Reading its returns: `success/complete`, `pending/outbound`, `pending/swapping`, `pending/confirming`, `pending/inbound`, `undefined`. `refunded` is structurally unreachable from the node response. Every refund detection in this file lives in `getThorMayaMidgardStatus`.

And the resolution order is:

```ts
if (midgardStatus) return midgardStatus
if (nodeProgress) return nodeProgress
```

So when Midgard errors, times out, or returns a shape that fails `isMidgardResponse`, refund detection does not degrade to "unknown". It degrades to whatever the node says, and the node's only terminal answer is `success`. A refund reaches `outbound_signed.completed` the same as a successful swap does, because the refund out is an outbound.

Probe 1, node reporting `outbound_signed: { completed: true }` with Midgard returning 502:

```
{"provider":"thorchain","txHash":"THOR-REFUND-SOURCE","status":"success","stage":"complete"}  terminal=true
```

A user who got their funds back is told the swap completed, `isSwapArrivalStatusTerminal` says stop polling, and there is no `destinationTxHash` to notice the discrepancy from. This is the failure your own docblock is trying to prevent, an upstream outage producing a wrong terminal answer, just arriving via the node fallback instead of via a thrown error.

To be straight about my evidence: the structural half is airtight and provable from the diff alone, since no code path exists that returns `refunded` from a node response. The specific `success` outcome comes from a payload I constructed. I tried to confirm THORNode's real `outbound_signed` behaviour on an actual refunded tx and could not reach `thornode.ninerealms.com` or `midgard.ninerealms.com` from this environment (both `000`, while `api.github.com` returned `200`, so that is my egress and not their uptime). Worth someone with reachability confirming the exact stage payload. The structural point does not depend on it.

Ask: when Midgard is unavailable, the node fallback should not be allowed to report a terminal `success`. Either cap it at `pending/outbound` when Midgard did not corroborate, or throw `SwapArrivalStatusRequestError` the way the rest of the file does for unreadable upstreams. Both are consistent with the principle already written in the docblock.

### preferably-blocking: `not_found` is terminal, so pollers give up on swaps that are merely not indexed yet

`isSwapArrivalStatusTerminal` is `result.status !== 'pending'`, which makes `not_found` terminal. But `not_found` is returned in the window between broadcast and indexing, on every provider:

```
probe 2 (THOR, node has not observed it, Midgard actions empty)  => not_found, terminal=true
probe 3 (LI.FI returns status NOT_FOUND)                         => not_found, terminal=true
```

Probe 2 is the normal state of a swap for the first stretch after broadcast: `inbound_observed.started` is false so `getThorMayaNodeProgress` returns `undefined`, Midgard has zero actions so `midgardAuthoritativelyEmpty` is true, and the function returns `not_found`. LI.FI likewise uses `NOT_FOUND` for "not indexed yet", not only for "never existed".

Since `isSwapArrivalStatusTerminal` is the exported contract for deciding when to stop, any straightforward poller stops on the first tick and reports the swap as missing. That is a worse user-facing outcome than the refund case in some ways, because it hits every swap rather than only the ones during an outage.

Ask: either exclude `not_found` from terminal, or split it into a transient and a settled variant. If terminal is deliberate, the docblock and the SDK README need to say loudly that callers must apply their own grace period, because the function name currently promises the opposite.

### should-fix: `findDestinationTxHash` cannot match THORNode's own key, and the PR's fixture proves it

The matcher accepts `txID`, `txHash`, `tx_hash`. THORNode uses `id`. Your own fixture says so:

```json
"thorchainNodePending": { "tx": { "id": "THOR-SOURCE" }, ... }
```

So the source-hash exclusion (`trimmed.toLowerCase() !== source`) never even engages on a node response, and no outbound hash is recoverable from one. Probe 4, with a realistic `out_txs: [{ "id": "THOR-REAL-OUT", "chain": "BTC" }]`, returned no `destinationTxHash` at all. Mostly latent because Midgard normally supplies it via `txID`, but it compounds the first finding: in exactly the Midgard-down case, you get both the wrong status and no destination hash to contradict it.

### nit: two silent heuristics in `findDestinationTxHash` worth a comment

`depth > 12` truncates without signalling, and `found.at(-1)` means "last hash encountered in depth-first `Object.entries` order wins", which is JSON key order and therefore provider-response-order. It happens to give the right answer for the Skip multi-hop fixture, and I could not construct a realistic case where it picks wrong. But "correct because the provider happens to serialize hops in order" is a property worth writing down, since it is invisible at the call site and a provider reordering its JSON would break it quietly.

### note on the receipt, not a blocker

Your four scenarios are THOR refund, Maya success, Skip multi-hop success, LI.FI refund, all with both upstreams healthy. That is a fair happy-path receipt and it does what I asked for. Flagging only that it is why it did not surface any of the above: every finding here lives in a degraded or early state, one upstream unavailable or the tx not yet indexed. If you add the fix, a receipt with Midgard stubbed out would be the one that proves it.

---

Ran on `befc34a40` against `origin/main` at `1e05f2c4a` (merge-base `3ea8b2c51`): `packages/core/chain/swap/utils/getSwapArrivalStatus` 28/28, `packages/sdk/tests/unit/utils/publicExports.test.ts` and RN entry 44/44, all green. The four probes were run from a temporary spec alongside the suite; it has been deleted and the worktree is clean.

The shape of this API is good and I want it to land. It is the two terminal-status answers that are wrong in degraded conditions that I would want fixed first, since a status normalizer's whole job is being trustable when the upstreams are not.

gomesalexandre (dismissed):
Re-reviewed current head `29c904d2567d87e7839654d70cda2a39551b07fe`.

The public-surface addition is useful, but I still do not think the terminal-state contract is safe enough to stamp yet.

1. preferably-blocking: LI.FI `DONE/PARTIAL` is flattened into ordinary `success`

At `packages/core/chain/swap/utils/getSwapArrivalStatus/index.ts:415-425`, the code only special-cases `substatus === 'REFUNDED'`, then maps every other `status === 'DONE'` to `{ status: 'success', stage: 'complete' }`.

But the LI.FI status model distinguishes `DONE/COMPLETED` from `DONE/PARTIAL`. Their docs define `PARTIAL` as a completed transfer where the user received a different token than requested, not the exact intended asset.

This PR's own test table also bakes that flattening in:
- `index.test.ts:267-270` expects `{ status: 'DONE', substatus: 'PARTIAL' }` to normalize to `success/complete`

The exported result union at the top of the file has no way to preserve that distinction either:
- `SwapArrivalStatusResult` only allows `pending | success | refunded | not_found | error`

So a caller can be told the requested swap completed successfully even when the provider is explicitly saying "partial". For a user-facing swap-status API, that is a real contract bug, not just lost metadata.

2. preferably-blocking: malformed / unknown Midgard action statuses are converted into terminal failure instead of a request/schema error

The file-level contract says malformed upstream responses must throw, not be converted into terminal `error`:
- `index.ts:458-460`: "Network, HTTP and malformed-response failures throw `SwapArrivalStatusRequestError`; they are not converted to terminal `error` results..."

But `getThorMayaMidgardStatus` currently does the opposite once it finds a `refund` or `swap` action:
- `index.ts:247-256`: refund actions return `error/failed` for any status other than `success` or `pending`
- `index.ts:262-276`: swap actions return `error/failed` for any status other than `success` or `pending`

That means a shape like `{ actions: [{ type: 'swap' }] }` passes the current `isMidgardResponse()` gate and becomes a terminal swap failure even though the provider never actually said the swap failed.

For this API, only an explicit failed status should become terminal `error`. Unknown or missing action statuses should be treated as unreadable upstream state and surfaced as `SwapArrivalStatusRequestError`, consistent with the documented contract above.

3. preferably-blocking: THOR/Maya node-only completion ignores the node refund discriminator and can stay non-terminal even when the node already knows it is a refund

At `index.ts:199-203`, completed `outbound_signed` is always forced to `pending/outbound` with the comment that only Midgard can tell swap output from refund.

But THORNode already exposes refund intent on planned outbounds. The upstream schema added a `refund` boolean on `planned_out_txs[]` specifically for this distinction. This implementation never reads `planned_out_txs` at all, and the only node-side outbound hash extraction it does is from `out_txs[].id` (`index.ts:173-182`).

So when Midgard is degraded, the implementation can keep a transaction pollable even though the node already has enough information to say the outbound is a refund path.

I am good with failing closed when there is genuinely not enough information, but ignoring an available node-side refund discriminator is a different thing. If the node payload is sufficient to distinguish refund from ordinary outbound, the normalizer should use it rather than forcing every completed node-only outbound to remain pending.

gomesalexandre:
Tackled in 60b89ebed (all 3 preferably-blocking findings from the last changes-requested review):

1. LI.FI `DONE/PARTIAL` no longer flattens into `success`. Added a `partial` status to `SwapArrivalStatusResult` (`status: 'partial'; stage: 'complete'`) and route `DONE/PARTIAL` to it. Fixed the test that baked in the old (wrong) `PARTIAL -> success` assertion, added a dedicated regression test.

2. Unknown/missing Midgard action statuses (refund or swap) now throw `SwapArrivalStatusRequestError` instead of becoming a terminal `error`. Midgard only documents `success`/`pending` as action statuses, so anything else is malformed-response data per this file's own documented contract, not a real terminal failure. Added a regression test with `{ actions: [{ type: 'swap' }] }` (missing status) asserting a throw.

3. THOR/Maya node-only completion now reads THORNode's `planned_out_txs[].refund` when Midgard is unavailable, reporting `refunded` or `success` accordingly instead of always forcing `pending/outbound`. Falls back to `pending/outbound` only when the flag isn't present on every planned outbound (can't be trusted). Added regression tests for both the refund and swap-output cases.

Full test suite for the file green (34 tests), plus `publicExports.test.ts` and the RN entry test (`VULTISIG_STRICT_SINGLETON=0`) re-verified since the result union changed. Typecheck/lint/prettier clean on touched files. Added a changeset (`swap-arrival-status-terminal-fixes.md`, patch bump) documenting the contract fix.

Not merging/approving this myself per the four-eyes rule — leaving that to another reviewer.

gomesalexandre (commented):
Re-reviewed current head `60b89ebe`.

I re-checked the three unsafe terminal-state gaps that previously blocked this:
- `packages/core/chain/swap/utils/getSwapArrivalStatus/index.ts` now preserves LI.FI `DONE/PARTIAL` as a distinct `partial` terminal result instead of flattening it into ordinary success
- malformed or unknown Midgard action statuses now throw `SwapArrivalStatusRequestError` instead of being converted into terminal swap failure
- THOR/Maya node-only completion now reads `planned_out_txs[].refund` so node-only completion can resolve to `success` / `refunded` when the node already knows the outcome

Not approving: this PR head contains a substantive `gomesalexandre` commit (`60b89ebe`), so it needs a non-gomes current-head approval before merge.

rcoderdev:
Addressed: the requested terminal-state fixes are on the current head, with focused coverage and all required checks green. Re-requesting review now.

johnnyluo (approved):
LGTM

PR #1811, opened 2026-08-09
fix(swap): bind vault-free quote safety

rcoderdev:
## Summary

- Bind swap quotes to the requested coin pair, source amount, expiry, and exact returned transaction before vault-free preparation.
- Reject missing, expired, stale, mismatched, or mutated bindings before wallet, key, or payload work.
- Keep the SDK's 60-second quote refresh recommendation separate from the five-minute general-quote preparation window; real provider deadlines still take precedence.
- Preserve `bigint` and `Uint8Array` quote values across Electron structured-clone IPC.
- Expose `SwapQuoteExpiredError` so vault-free callers can catch expiry and requote.

## Compatibility

This is a minor SDK/core-chain change: `prepareSwapTxFromKeys` now requires the `BoundSwapQuote` returned by `findSwapQuote` or `getSwapQuote`.

Keep that bound quote as the live returned object. Structured cloning is supported, but JSON round-tripping is not because it loses required runtime value types such as `bigint`.

## Validation

- Repository-wide typecheck and lint
- Full SDK unit suite: 3,303 tests
- Focused core quote-selection/fingerprint and SDK swap-preparation/service suites
- Live designated-vault Ethereum-to-USDC quote and non-broadcast dry-run

### DIFF
--- a/.changeset/guard-vault-free-evm-swaps.md
+++ b/.changeset/guard-vault-free-evm-swaps.md
@@ -4,3 +4,5 @@
 ---
 
 Bind raw swap quotes to their exact coin identities, requested source amount, absolute expiry, and returned transaction, then reject missing, stale, mismatched, or mutated quotes in the vault-free swap preparation path before any payload construction.
+
+Consumers must pass the bound quote returned by `findSwapQuote` or `getSwapQuote` into preparation instead of constructing or persisting an unbound raw quote.
--- a/.changeset/guard-vault-free-evm-swaps.md
+++ b/.changeset/guard-vault-free-evm-swaps.md
@@ -1,8 +1,8 @@
 ---
-"@vultisig/core-chain": patch
-"@vultisig/sdk": patch
+"@vultisig/core-chain": minor
+"@vultisig/sdk": minor
 ---
 
-Bind raw swap quotes to their exact coin identities, requested source amount, absolute expiry, and returned transaction, then reject missing, stale, mismatched, or mutated quotes in the vault-free swap preparation path before any payload construction.
+Bind raw swap quotes to their exact coin identities, requested source amount, absolute expiry, and returned transaction, then reject missing, stale, mismatched, or mutated quotes in the vault-free swap preparation path before any payload construction. General quotes remain preparation-valid for five minutes while SDK presentation continues to recommend refreshing after 60 seconds, and vault-free callers can catch `SwapQuoteExpiredError` to requote.
 
-Consumers must pass the bound quote returned by `findSwapQuote` or `getSwapQuote` into preparation instead of constructing or persisting an unbound raw quote.
+Consumers must pass the live bound quote returned by `findSwapQuote` or `getSwapQuote` into preparation instead of constructing or persisting an unbound raw quote. Structured cloning is supported; JSON round-tripping is not because it loses required runtime value types such as `bigint`.
--- a/.changeset/guard-vault-free-evm-swaps.md
+++ b/.changeset/guard-vault-free-evm-swaps.md
@@ -1,0 +1,6 @@
+---
+"@vultisig/core-chain": patch
+"@vultisig/sdk": patch
+---
+
+Bind raw swap quotes to their exact coin identities, requested source amount, absolute expiry, and returned transaction, then reject missing, stale, mismatched, or mutated quotes in the vault-free swap preparation path before any payload construction.
--- a/examples/electron/electron/ipc-handlers.ts
+++ b/examples/electron/electron/ipc-handlers.ts
@@ -464,8 +464,9 @@ export function registerIpcHandlers(ipcMain: IpcMain): void {
     const vault = await sdk.getVaultById(vaultId)
     if (!vault) throw new VaultError(VaultErrorCode.VaultNotFound, 'Vault not found')
     const result = await vault.prepareSwapTx(params)
-    // Serialize for IPC (convert BigInt to string)
-    return JSON.parse(JSON.stringify(result, (_key, value) => (typeof value === 'bigint' ? value.toString() : value)))
+    // Preserve the echoed quote's BigInt and Uint8Array values so it remains
+    // consistent with the safety binding when it crosses Electron IPC.
+    return result
   })
 
   // === TRANSACTION OPERATIONS ===
--- a/examples/electron/electron/ipc-handlers.ts
+++ b/examples/electron/electron/ipc-handlers.ts
@@ -452,8 +452,11 @@ export function registerIpcHandlers(ipcMain: IpcMain): void {
     const vault = await sdk.getVaultById(vaultId)
     if (!vault) throw new VaultError(VaultErrorCode.VaultNotFound, 'Vault not found')
     const quote = await vault.getSwapQuote(params)
-    // Serialize for IPC (convert BigInt to string)
-    return JSON.parse(JSON.stringify(quote, (_key, value) => (typeof value === 'bigint' ? value.toString() : value)))
+    // Electron IPC uses structured clone, which preserves the BigInt and
+    // Uint8Array values covered by the quote safety fingerprint. JSON
+    // serialization would change those values and make a valid quote fail the
+    // preparation-time integrity check when it returns from the renderer.
+    return quote
   })
 
   ipcMain.handle('vault:prepareSwapTx', async (_event, vaultId: string, params: any) => {
--- a/packages/core/chain/package.json
+++ b/packages/core/chain/package.json
@@ -1908,6 +1908,11 @@
       "import": "./dist/swap/quote/getSwapQuoteProviderName.js",
       "default": "./dist/swap/quote/getSwapQuoteProviderName.js"
     },
+    "./swap/quote/getSwapQuoteSafetyFingerprint": {
+      "types": "./dist/swap/quote/getSwapQuoteSafetyFingerprint.d.ts",
+      "import": "./dist/swap/quote/getSwapQuoteSafetyFingerprint.js",
+      "default": "./dist/swap/quote/getSwapQuoteSafetyFingerprint.js"
+    },
     "./swap/quote/SwapQuote": {
       "types": "./dist/swap/quote/SwapQuote.d.ts",
       "import": "./dist/swap/quote/SwapQuote.js",
--- a/packages/core/chain/swap/quote/SwapQuote.ts
+++ b/packages/core/chain/swap/quote/SwapQuote.ts
@@ -20,10 +20,14 @@ export type SwapQuote = {
   requestedAmount?: bigint
   /** Absolute quote expiry in milliseconds, bound by `findSwapQuote`. Absent on legacy/manually constructed quotes. */
   expiresAt?: number
-  /** Integrity binding for the request identity, expiry, and exact returned transaction. */
+  /** Mutation/stale-reuse binding for the request identity, expiry, and exact returned transaction. */
   safetyFingerprint?: string
 }
 
-/** A canonical quote returned by `findSwapQuote`, with fund-safety metadata present. */
+/**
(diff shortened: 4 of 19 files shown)


gomesalexandre (changes requested):
Read the whole diff, checked out `835fde180`, ran the affected suites green (46/46 on `prep/swap` + `SwapService`, 60/60 on core `findSwapQuote.selection`), and drove the new binding directly with a probe carrying both a must-pass and a must-fail control. No argument with the motivation: a quote that reaches preparation unbound to the pair and amount it was fetched for is a real hole, and failing closed before payload construction is the right place to close it. The synchronous input snapshot in `findSwapQuote` is a nice touch, that race was genuinely reachable.

Two blockers, both about what this does to callers rather than about the guard itself.

---

### preferably-blocking: general/EVM quotes get a hard 60-second death clock that starts at fetch time

`findSwapQuote.ts`:

```ts
const GENERAL_QUOTE_TTL_MS = 60_000
...
const expiresAt = 'native' in quote.quote ? quote.quote.native.expiry * 1000 : now + GENERAL_QUOTE_TTL_MS
```

and `prep/swap.ts` turns that into a throw:

```ts
if (expiresAt <= Date.now()) {
  throw new Error('prepareSwapTxFromKeys: swap quote has expired; refresh the quote before signing')
}
```

Probe on this head, same untouched quote object, only the clock moving via `vi.setSystemTime`:

| when `prepareSwapTxFromKeys` is called | result |
|---|---|
| must-pass control: t+0, general/EVM quote straight out of the binder | ok, payload built |
| t+30s | ok |
| t+61s | **THROW** `swap quote has expired; refresh the quote before signing` |
| must-fail control: t+0, after `quote.general.tx.evm.to = '0xattacker'` | THROW `does not match the requested coins or its original transaction` |

Both controls discriminate, so the middle rows are real. Nothing about the quote changed between t+30s and t+61s.

Worth noting the same hunk deletes the comment that explained why this branch had no expiry:

```diff
-// Other general-quote routes carry no expiry field the SDK can assert here.
```

That statement is still true of the providers. The PR removes the note and manufactures a deadline in its place.

Getting ahead of the obvious rebuttal, because it is a fair one: 60_000 is not invented, it is `SwapService.DEFAULT_QUOTE_EXPIRY_MS`, and `SwapService.prepareSwapTx` already threw past it. So for callers going through `SwapService` this is roughly status quo. What is new is:

1. **Direct `prepareSwapTxFromKeys` callers had no general/EVM expiry at all before this.** `assertNativeQuoteNotExpired` and `assertCowQuoteNotExpired` covered native and CoW off provider-supplied deadlines. The EVM aggregator branch had nothing, and now it has a deadline that no provider asked for. That is the vault-free / headless path, which is exactly the one this PR is scoping itself to.
2. **The clock starts when the quote is fetched, not when preparation runs.** `now` is captured inside `findSwapQuote`, so the 60s budget has to cover the entire round trip: render the quote, the human reads the rate, taps confirm, any approval branch. A user who reads the screen for a minute gets a hard refusal on a quote that is still perfectly valid at the provider.

Fund-safety-wise this is the direction the repo says not to fail in: a false refusal on legit movement rather than a false pass. I am not asking you to drop the expiry, staleness is a real risk. Asking for one of:

- decouple the two numbers, so the presentation window can stay 60s while the preparation floor is something a human round trip actually fits in, or
- make it configurable on `FindSwapQuoteInput` rather than a module constant, or
- at minimum throw a distinguishable error type for expiry so a caller can catch it and re-quote automatically instead of surfacing "refresh the quote" to a user who already did.

Whichever you pick, the number wants a comment saying where it came from, because as written it reads as arbitrary and the next person will not know it is mirroring `DEFAULT_QUOTE_EXPIRY_MS`.

### preferably-blocking: this is a consumer-visible break shipped as `patch`

`.changeset/guard-vault-free-evm-swaps.md` is `patch` on both `@vultisig/core-chain` and `@vultisig/sdk`, and its own body says:

> Consumers must pass the bound quote returned by `findSwapQuote` or `getSwapQuote` into preparation instead of constructing or persisting an unbound raw quote.

"Consumers must" plus `patch` do not go together. A consumer on a caret range picks this up automatically and gets a runtime throw on code that worked yesterday, in three separate ways: a hand-constructed quote (missing binding), a persisted quote (expired), and a serialized quote (fingerprint mismatch, see below). Same call I made on #1821, so I am being consistent rather than picking on this one.

`minor` at minimum, and a body that names the 60s general TTL and the "do not JSON round-trip a quote" constraint explicitly. Right now a consumer reading the changeset has no way to learn either.

### should-fix: `PrepareSwapTxFromKeysParams.swapQuote` is still `SwapQuote`, so the break is runtime-only

`BoundSwapQuote` exists in this PR and you already threaded it through `SwapService`, `swap-types.ts`, and `tools/swap/findSwapQuote.ts`. But `PrepareSwapTxFromKeysParams.swapQuote` is untouched in the diff, still typed `SwapQuote`, whose three new fields are `?:` and documented "Absent on legacy/manually constructed quotes". So the compiler happily accepts exactly the quote shape the runtime refuses.

Typing that param `BoundSwapQuote` converts the whole first class of breakage from a production throw into a build error at upgrade time, which is where a consumer wants to find out. Cheap, and it makes the changeset note redundant for anyone on TS.

### should-fix: the tests bind a 10-minute expiry, so nothing pins the real TTL

`tests/unit/tools/prep/swap.test.ts:61`:

```ts
const bindSwapQuote = (quote, requestedAmount, { expiresAt = Date.now() + 600_000, ... } = {}) => ({
```

600_000 in the helper, 60_000 in production. Every test in the file therefore runs with a window ten times the real one, and the suite would stay green if `GENERAL_QUOTE_TTL_MS` were changed to 5 seconds or 5 hours. Worth a case that binds through the real code path (or at least asserts the constant) so the TTL is pinned rather than incidental.

### should-fix: the fingerprint makes a quote non-serializable, and the failure message misattributes it

This PR already had to fix one consumer for it, `examples/electron/electron/ipc-handlers.ts`:

```diff
-    return JSON.parse(JSON.stringify(quote, (_key, value) => (typeof value === 'bigint' ? value.toString() : value)))
+    return quote
```

with a new comment saying JSON serialization "would change those values and make a valid quote fail the preparation-time integrity check". Correct, and it is the whole point, but it generalizes: any consumer that puts a quote through JSON (an HTTP hop, a cache, `redux-persist`, a worker `postMessage` that gets stringified, storing it to resume after a background/foreground cycle) now gets a throw. The `bigint` and `Uint8Array` values inside `quote` are what makes it unavoidable.

What makes this worse than it needs to be is the message:

> swap quote does not match the requested coins or its original transaction; refresh the quote before signing

which reads as tampering. A support ticket from a consumer who merely cached a quote will get triaged as an attack. Worth either a distinct message for "the shape survived but the value types did not" or, if that is not cheaply detectable, a sentence in the docstring and the changeset saying quotes are not serializable and must be held as live objects.

To scope it honestly: React Native is fine here. `packages/sdk/src/platforms/react-native/index.ts:285` is a plain lazy `await import('../../tools/prep/swap')` and a direct call, no bridge and no JSON in between, so the RN consumer is not affected. Flagging that so nobody re-raises it as a mobile break.

### q: "integrity-binds" and "tampered" oversell what an unkeyed hash buys

`getSwapQuoteSafetyFingerprint` is `sha256(canonicalize(...))` with no secret, so anything that can mutate the quote can recompute the fingerprint over the mutated value. Against an adaptive adversary in the same process it is not a barrier. What it genuinely buys, and this is worth having, is detection of accidental mutation and of stale or cross-request reuse.

Related, and the reason I would not lean on it too hard: of the five hashed inputs only `from` and `to` come from the caller at verification time. `requestedAmount`, `expiresAt`, and `quote` are all read back off the quote object itself, so a self-consistent forged quote passes. The check actually carrying the load against a swapped-in quote is the separate one right above:

```ts
if (boundAmount !== requestedAmount) {
```

which compares against the caller's own `toChainAmount(safeParams.amount, ...)`. Suggest the docstring say "detects accidental mutation and stale or mismatched reuse" rather than "integrity-binds", and the PR body drop "tampered", so the next reader does not assume a MAC is in place.

### nit: two bits of hard-won context lost in the comment rewrite

The rewrite of the `assertAmountMatchesCommittedSellAmount` header is mostly fine, the fee-divergence reason survives. Two specifics did not:

- the concrete example, `request 100_000n -> committed 99_999n`, which is what makes "legitimately differ" land for the next reader
- the explicit enumeration `Native/evm/solana fail open`, now folded into "other tx variants are covered by the quote-level binding above"

The second one is not quite the same claim: the quote-level binding checks the amount the quote was *bound with*, not a committed sell field inside the tx, so those variants still have no committed-amount cross-check. Worth keeping the old sentence alongside the new one.

### nit: `canonicalize` tags collide with real object keys

`bigint` serializes to `{"$bigint":"5"}` and `Uint8Array` to `{"$bytes":"0x.."}`, but a plain object literally shaped like that canonicalizes identically. Via the public API on this head:

```
getSwapQuoteSafetyFingerprint({ ..., quote: { v: 5n } })
getSwapQuoteSafetyFingerprint({ ..., quote: { v: { $bigint: '5' } } })
-> identical hash
```

Fully contrived, no provider emits that shape, and I am not pretending it is exploitable. But it is one line to close (escape a leading `$` in the object-key branch) and it removes the caveat entirely.

### nit: `undefined` is dropped in an object but throws in an array

Same probe, same call:

```
quote: { v: undefined }    -> ok, key silently dropped
quote: { v: [undefined] }  -> THROW  Unsupported swap quote safety value: undefined
```

The object branch filters `undefined` out, the array branch falls through to the `default:` throw. A sparse or partially-populated array anywhere in a provider response is now a hard throw on a swap-preparation path, with a message that does not name the field. Worth treating array holes the same way the object branch does, or at least naming the path in the error.

---

Ran on `835fde180`: `tests/unit/tools/prep/swap.test.ts` + `tests/unit/services/SwapService.test.ts` 46/46, core `findSwapQuote.selection.test.ts` 60/60. The probe above was a temporary 5-case file in `tests/unit/tools/prep/`, run against the same head and removed after.

One thing I chased and want to explicitly clear so nobody re-raises it: the seconds-vs-milliseconds mixup I went looking for is not there. `native.expiry * 1000` and `cowswap_order.validTo * 1000` both convert correctly and `assertQuoteSafetyBinding` compares against `Date.now()`. Units are consistent throughout.

gomesalexandre (approved):
looks good.

checked the current head against the earlier blockers and the follow-up. the caller-facing regressions are addressed:
- `packages/core/chain/swap/quote/findSwapQuote.ts` now gives general quotes a 5 minute preparation ttl while keeping the 60 second presentation refresh separate
- `packages/sdk/src/tools/prep/swap.ts` promotes `swapQuote` to `BoundSwapQuote`, throws `SwapQuoteExpiredError`, and improves the mismatch message to call out JSON round-tripping
- `.changeset/guard-vault-free-evm-swaps.md` is now minor and documents the 5 minute preparation window plus the non-JSON transport constraint
- `examples/electron/electron/ipc-handlers.ts` preserves the bound quote over ipc

receipts on this head:
- `gh pr view` current checks: Unit Tests, Integration Tests (Vitest), Lint and Type Check, Test Summary, CodeRabbit all green
- `cd packages/sdk && ../../node_modules/.bin/vitest run --config tests/unit/vitest.config.ts tests/unit/tools/prep/swap.test.ts` -> 19/19 passed
- `cd packages/sdk && ../../node_modules/.bin/vitest run --config tests/unit/vitest.config.ts tests/unit/services/SwapService.test.ts` -> 28/28 passed after borrowing the already-built `packages/core/config/dist` from the healthy source install into the disposable review clone
- direct probe: `GENERAL_QUOTE_PREPARATION_TTL_MS=5 * 60_000`, `structured_clone_stable=true`, `json_round_trip_changes_fingerprint=true`

no further blockers from me.

PR #1812, opened 2026-08-09
Add SDK chain descriptor registry

### LINKED ISSUE #1230: spike: chain-descriptor registry consumers derive from (add-a-chain is compile-safe here, silent everywhere else)
Direction spike from the round-18 cross-repo audit. The sdk's exhaustive `Record<Chain, T>` tables kill drift inside this repo (misses are compile errors), but consumers do not inherit that protection because they re-declare their own chain unions and tables. Measured blast radius of "add a chain" today:

- sdk: ~10-14 tables, all compile-enforced
- mobile app: ~5-9 files (evm) to 20-30 (new chain kind); only 2 of its tables are exhaustive, ~25 partial-map/switch sites silently drop a new chain
- agent backend: ~12-18 core files, mostly loose `Record<string, T>` and switches (highest silent-drift risk)

Already-observed drift from exactly this: explorer urls diverged app-vs-sdk on 10+ chains, the Hyperliquid 998 vs 999 chain-id mismatch produced malformed txs, and duplicate cosmos denom tables exist 4x across the stack.

## the spike

Design one chain-descriptor registry in the sdk that consumers derive from instead of re-declaring: per-chain capabilities + metadata (ids, decimals, explorer, fee units, address family, feature flags like memo support), exported so a consumer table becomes `deriveFrom(chainRegistry, ...)` and a missing chain is a compile error THERE too. Deliverable is a design doc + a thin prototype for one consumer surface (explorer urls is the cheapest - the export already exists, consumers just do not use it), NOT a big-bang migration.

Open questions the spike should answer: how consumers extend a descriptor with app-local fields without forking it; versioning story when a new chain lands in the sdk before consumers support it; whether capability flags belong here or stay consumer-side.

effort: L (spike itself: M) / confidence: HIGH that the duplication is real, MED on the right registry shape

rcoderdev:
Closes #1230

## What

Adds one exhaustive, immutable chain-descriptor registry for all SDK chains and exports it from the core, SDK, and React Native entrypoints. Consumers can derive chain-indexed data from that registry, so adding a new `Chain` produces a compile-time failure at the registry boundary instead of silently leaving parallel tables incomplete.

## How

- Defines descriptors for all 39 chains, including chain kind and exact explorer URL metadata.
- Migrates block-explorer URL generation and selected EVM explorer metadata to the registry.
- Provides typed `deriveFromChainRegistry` and exhaustive `extendChainRegistry` helpers.
- Documents ownership, migration, and extension rules.
- Preserves the complete registry and helper surface in Node, browser, Chrome-extension, and React Native packages.

## Verification

- `yarn check:ci`
- Focused core explorer tests: 19/19
- Focused SDK public-export, React Native entry, and SwapService tests: 71/71
- Packed SDK runtime/type contracts, including the React Native subpath
- Exhaustive 39-chain runtime parity and immutability checks


gomesalexandre (approved):
CI came back green on `3d493a9b9`, so stamping as promised.

`Unit Tests` pass, `Integration Tests` pass, `Lint and Type Check` pass, `Test Summary` pass. That green tick is also the first time CI has actually executed `yarn test:core`, so your 19 new `getBlockExplorerUrl` registry tests now have a real verdict behind them rather than a skipped step.

The two should-fixes from my earlier comment stand as follow-ups, not blockers: the `&&`-chained root `test` script that swallowed the core project, and the import-time `getChainKind()` in `chainRegistry` that turns a partial `ChainKind` mock into an import-time crash. Both are yours to weigh, neither should hold this up.

PR #1813, opened 2026-08-09
fix(ripple): only declare genuine trust lines as TrustSet

Ehsan-saradar:
## Problem

`RippleSpecific.transaction_type` was introduced in #1794 so a Ripple payload states its operation on the wire instead of leaving every signer to infer it. The builder sets it from the coin's shape alone:

```ts
const isTrustSet = isRippleTrustSet(keysignPayload) // non-native Ripple coin + contractAddress
```

But an issued-currency **Payment** has that identical shape — a non-native Ripple coin with a `contractAddress`. So sending SOLO stamps the payload `TRANSACTION_TYPE_RIPPLE_TRUST_SET`.

That is worse than the ambiguity the field removes. Before it existed, a token send *diverged*: this SDK built a TrustSet, an iOS co-signer built a Payment, the bytes disagreed and the ceremony failed having signed nothing. With the field set, every signer reads the declaration and agrees to build a TrustSet — so the ceremony **completes** over an operation the user never asked for, setting a trust-line limit to the amount they meant to send. No funds move, and nothing surfaces it to the user.

`@vultisig/core-mpc@1.18.0` is published and `vultisig-windows` is already on it, so this is live rather than hypothetical.

## Fix

One predicate split into two, with opposite scopes:

| | scope | used by |
|---|---|---|
| `hasRippleTrustSetShape` | broad — coin shape only | signing fallback |
| `originatesRippleTrustSet` | narrow — also requires `toAddress === issuer` | builder |

Declaring is an assertion, not a guess, so the builder now takes more than the shape. A TrustSet is addressed to the **issuer** — the party being trusted. A Payment is addressed to a recipient. That is what separates them.

**The signing fallback stays broad on purpose.** Clients already in the field infer TrustSet from a non-native coin alone, and honouring that same inference is what keeps a genuine TrustSet byte-identical across a mixed-version committee. Narrowing it would break MPC parity with every signer already released. A token send simply goes back to diverging safely instead of completing wrongly.

## Tests

Three cases on the builder — send to a recipient → `UNSPECIFIED`; addressed to the issuer → `RIPPLE_TRUST_SET`; unparseable token id → `UNSPECIFIED`. Mutation-checked: restoring the old broad predicate fails two of them.

53/53 pass across both ripple suites, including the existing mixed-version byte-parity test, which confirms the signer fallback is untouched.

## Related

- Builds on #1794
- Windows companion: vultisig/vultisig-windows#4581 (same narrowing, applied to the history record)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ripple transaction classification to distinguish genuine TrustSet transactions from issued-currency payments.
  * Prevented token payments from being incorrectly completed as TrustSet transactions.
  * Added safe handling for malformed token identifiers.
* **Tests**
  * Added coverage for issuer-address matching, non-issuer token payments, and invalid token data.
* **Release**
  * Patch releases are planned for the Ripple-related packages.


gomesalexandre (approved):
Right call, and the reasoning in the description is the good kind: it picks the safer of two bad outcomes rather than pretending the ambiguity went away. A ceremony that completes over an operation the user never asked for is strictly worse than one that fails having signed nothing, and that is exactly the trade being made here. Traced it rather than taking the description's word for it.

**What I verified (not assumed):**

- `parseRippleTokenId` cannot hand back an empty issuer: it throws when the separator sits at index 0 or at the last position (`issuedCurrency.ts`), so there is no `'' === ''` path where a garbage token id plus an empty `toAddress` silently declares a TrustSet. That was the first false-positive I went looking for.
- `attempt` returns `{data}` on success and `{error}` on throw, so the `'data' in issuer` narrowing is the right shape and an unparseable id lands on `false` rather than throwing out of the builder.
- `signRipple` and native XRP stay excluded, since `originatesRippleTrustSet` delegates to `hasRippleTrustSetShape` first. Confirmed both with controls.
- Ran your mutation claim myself instead of trusting it: restoring the broad predicate in the builder takes exactly **2** of the new tests red (`does not declare a send of an issued currency a TrustSet`, `does not declare one when the token id cannot be parsed`). Non-vacuous, as advertised.

Two things worth writing down. Neither blocks.

---

**should-fix: "a token send simply goes back to diverging safely" is only true across *implementations*.**

The signer fallback stays broad on purpose, which I agree with for byte-parity. But `isRippleTrustSet` falls back to the shape whenever `transactionType !== RIPPLE_TRUST_SET`, and an explicit `UNSPECIFIED` is not distinguishable from absent. So this SDK still resolves a token send to `opTrustSet` at `signingInputs/resolvers/ripple.ts:123`.

Probed it directly, with must-fail controls alongside:

```
MUST-PASS  token Payment -> recipient   originatesRippleTrustSet = false   <- the fix works
MUST-PASS  same payload                 isRippleTrustSet         = true    <- signer still builds a TrustSet
CONTROL    native XRP                   all three predicates     = false
CONTROL    unparseable token id         originatesRippleTrustSet = false
```

So the divergence only materialises when the co-signer is a *different* implementation that reads the undeclared case as a Payment. In a committee where both sides run core-mpc (this SDK plus vultisig-windows, both on 1.18.0), both still infer TrustSet from the shape, both still agree, and the ceremony still completes over the wrong operation.

To be clear about blame: that is pre-existing and predates #1794, so it is not this PR's bug and I am not blocking on it. But the description and the new comment at `chainSpecific/resolvers/ripple.ts` both state the outcome flatly as safe divergence, and a future reader will take that as "token sends are handled now" when the same-implementation case is unchanged. Suggest either narrowing the wording to say the divergence is across implementations, or filing the follow-up and linking it. The wording is the part I care about, since the next person to touch this will trust the comment over the code.

**should-fix: paying an issued currency back to its issuer is still declared a TrustSet.**

`toAddress === issuer` separates the two operations in the common case, but redeeming an IOU by sending it back to the issuing account is an ordinary XRPL operation and has exactly that shape. Confirmed:

```
RESIDUAL   token Payment -> ISSUER      originatesRippleTrustSet = true
```

Which means a redemption gets stamped, every signer agrees, the ceremony completes, and the user sets a trust-line limit instead of paying the issuer. Same failure class this PR exists to close, surviving in the one case the new predicate cannot separate.

Strictly better than before, so this is a residual and not a regression - the old predicate caught *every* token send, this one catches a much rarer shape. But I do not think another inference rule fixes it, and I would not want to see one attempted: there is no origination call site anywhere in this repo (I grepped), so the payload arrives already built and the SDK genuinely cannot know what the caller meant. The real fix is the caller declaring intent explicitly, which is out of scope here. Worth a line in the doc comment on `originatesRippleTrustSet` saying the issuer check is a heuristic with this known hole, so the next person does not read it as exact.

**testing:** `packages/core` ripple suites at this head: **4 files / 61 tests passed, exit 0**, including the mixed-version byte-parity test, so the signer fallback is confirmed untouched. Mutation control run and restored clean; probe file deleted, working tree clean. Branch is 1 commit behind `origin/main` with no conflicts, `MERGEABLE`, CI 5/5 green. Changeset present and correctly scoped `patch` on both `@vultisig/core-mpc` and `@vultisig/sdk`.

PR #1815, opened 2026-08-09
Fail closed on SDK signature count mismatches

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signature validation during signing.
  * Signing now fails safely when the number of signatures does not match the number of message hashes.
  * Added handling for empty signature lists and single signatures paired with multiple message hashes.
* **Documentation**
  * Documented the fail-closed behavior for mismatched multi-input signatures.

### DIFF
--- a/.changeset/guard-multi-input-signature-count.md
+++ b/.changeset/guard-multi-input-signature-count.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Fail closed when multi-input signing returns a different number of signatures than message hashes.
--- a/.changeset/guard-multi-input-signature-count.md
+++ b/.changeset/guard-multi-input-signature-count.md
@@ -2,4 +2,4 @@
 '@vultisig/sdk': patch
 ---
 
-Fail closed when multi-input signing returns a different number of signatures than message hashes.
+Fail closed when multi-input signature counts differ from message hashes or a single signature is paired with multiple message hashes.
--- a/packages/sdk/src/vault/utils/convertSignature.ts
+++ b/packages/sdk/src/vault/utils/convertSignature.ts
@@ -75,6 +75,12 @@ export function convertToKeysignSignatures(
 
   if (signature.signatures && signature.signatures.length > 0) {
     // Multi-message case (UTXO inputs, bundled approvals, and similar flows)
+    if (signature.signatures.length !== messageHashes.length) {
+      throw new Error(
+        `Signature count ${signature.signatures.length} does not match message hash count ${messageHashes.length}`
+      )
+    }
+
     signature.signatures.forEach((sig, index) => {
       const messageHash = messageHashes[index]
       if (!messageHash) {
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
Good change, and the right posture: a signature/hash count mismatch is exactly the shape that should fail loud rather than produce a half-signed map that only blows up at compile/broadcast time. Traced the whole path before saying that, not just the diff.

**What I verified (not assumed):**

- `signature.signatures` is only ever produced by `formatSignature()` (`packages/sdk/src/adapters/formatSignature.ts:66`) as `messages.map(...)`, and `messages` is the same array handed to `sign()`. `BroadcastService.broadcastTx` re-derives it via `extractMessageHashes(keysignPayload)` (`BroadcastService.ts:141`), which is deterministic and `.sort()`ed (`TransactionBuilder.ts:318`). So in the happy path the two counts are structurally equal, and the new guard cannot false-block a legit flow. That was the thing I most wanted to rule out on a money path.
- Checked the one shape that looked like it could false-block: MLDSA/QBTC. `formatMldsaSignature()` returns a `Signature` with no `signatures` array, so it lands in the `else` branch and would hit the new `messageHashes.length > 1` throw. Ruled out: `getQBTCPreSignedImageHash` returns exactly one hash (`packages/core/mpc/chains/cosmos/qbtc/QBTCHelper.ts:32-35`), so QBTC never gets there. Gucci.
- The EVM approve+swap bundle (the real multi-hash case) goes `messages.length > 1` -> `signatures` populated 1:1 -> counts match. Also fine.

So no blockers. One thing worth a follow-up though:

---

**should-fix: the count guard silently ate the only test covering `Missing message hash for signature at index N`.**

`tests/unit/vault/utils/convertSignature.test.ts` L264 renamed `'should throw error when message hash is missing for multi-signature'` -> `'should throw error when there are more signatures than message hashes'` and swapped its expectation from `'Missing message hash for signature at index 1'` to the new count message. Grepped the test file at this head: **zero** remaining occurrences of `Missing message hash for signature`.

The thing is that branch is still reachable, it just needs a different input now. Before the PR, the short-array case was how you got there. After the PR the count guard intercepts that, but `!messageHash` still fires when the counts are EQUAL and a hash is falsy:

```ts
const signature = {
  signature: '',
  format: 'ECDSA' as const,
  signatures: [{ r: '0x..', s: '0x..', der: '0x..' }, { r: '0x..', s: '0x..', der: '0x..' }],
}
// 2 === 2 passes the new guard, then index 1 is empty
expect(() => convertToKeysignSignatures(signature, ['0x1111...', ''])).toThrow(
  'Missing message hash for signature at index 1'
)
```

I ran that locally as a throwaway probe with must-fail controls alongside it (a well-formed 2/2 that must NOT throw, and a real 2-vs-1 count mismatch that must report the COUNT message rather than the index one). 3/3 as expected, so the branch is live and just uncovered rather than dead.

Not blocking, since the branch is defence-in-depth on a shape `extractMessageHashes` shouldn't produce, but a fail-closed guard with no test is how a fail-closed guard quietly becomes fail-open. Suggest adding the case above back rather than leaving the file with one fewer branch covered than it had before.

**nit:** in the `signatures: []` + populated `signature.signature` case the message reads `Signature count 1 ...` when the array literally has zero entries. It's technically right (the `else` branch does have one signature, the top-level one) but it reads odd in a log. Something like `Signature count 1 (single-signature path) does not match ...` would save a future debugger a minute. Meh, take it or leave it.

**testing:** `packages/sdk` `yarn test` at this head: **212 files / 3213 tests passed, exit 0**, `convertSignature.test.ts` 17/17. Branch is up to date with `origin/main` (0 commits behind), `MERGEABLE`. Changeset present and correctly scoped `patch`.

PR #1817, opened 2026-08-09
fix(swapkit): surface the EVM affiliate fee and expose route price impact

Ehsan-saradar:
## What & why

Two gaps in the SwapKit quote left clients unable to display a swap's cost correctly. Both are data problems, not display bugs — every client is currently choosing between showing nothing and inventing a figure.

### The EVM branch dropped the affiliate fee

`buildEvmTx` never populated `affiliateFee`, although `buildSolanaTx` already extracted the same fee via `getSwapKitSwapFee`. An aggregator route therefore reached consumers with no swap fee to show, and the fee was silently absent from any headline total computed from the quote.

It now carries the fee. Two deliberate constraints:

- **Absent, not zero, when the response itemizes none.** `getSwapKitSwapFee` returns a `0n` placeholder in that case. Emitting it would have consumers render a definite `$0.00`, asserting something we cannot vouch for; leaving it absent lets them report the fee as part of the quoted rate instead.
- **Resolution failures degrade rather than throw.** `getSwapKitSwapFee` throws on four fee shapes (missing asset, unsupported asset, mixed assets, negative amount). Those throws were previously unreachable from the EVM path. Since the figure is display-only there, an unexpected shape must not take down a route that would otherwise sign. Solana keeps the bare call and its throw on purpose — its tx type requires the fee, so an unresolved one really is fatal.

### No price impact on `GeneralSwapQuote`

The type carried no price-impact field, and SwapKit's `meta` was parsed for `txType` alone. Clients that wanted the row had nothing to read, and at least one reached for the total-fee bps instead — a different figure that merely looks like slippage.

`GeneralSwapQuote.priceImpact` now carries the signed fraction, preferring the route's `meta.priceImpact` and falling back to `totalSlippageBps / 10000` for routes whose meta the proxy omits. It stays absent for providers that publish no impact, so consumers hide the row rather than substitute something for it.

## Test plan

`yarn test:core` (2582 passing), `tsc --noEmit`, `yarn lint`, `yarn format:check`, `yarn changeset:check-sdk-bundle` — all green locally.

Six new cases in `getSwapKitQuote.test.ts`: fee surfaced on an EVM route; absent when none is itemized; route stays signable when the fee shape cannot be resolved; price impact from `meta`; fallback to `totalSlippageBps`; absent when the route exposes neither.

## Notes for review

- Both packages are bumped `minor` in the changeset: the change touches bundled source, so `@vultisig/sdk` needs a new tarball for consumers to receive it.
- A client-side PR against `vultisig-windows` consumes both fields and is blocked on this release.

## Summary by CodeRabbit

* **New Features**
  * Swap quotes now include optional signed route price-impact information.
  * EVM swap transactions can include applicable affiliate or service fees.
  * Price impact uses provider metadata when available, with slippage as a fallback.

* **Bug Fixes**
  * Swaps continue gracefully when fee information is missing, invalid, or unresolved, omitting fees when necessary.
  * Invalid or negative fee data is handled safely without interrupting quote retrieval.


gomesalexandre (changes requested):
Read the whole diff, checked out `513a1e92c`, ran the affected suites green (109/109 across `getSwapKitQuote`, `priceImpactGuard`, `findSwapQuote.selection`), and drove the new EVM branch directly with a probe. No argument with the motivation: the EVM branch genuinely was the odd one out next to Solana, an aggregator swap showing no swap fee at all is a real reporting hole, and degrading to "no fee" rather than throwing is the right posture on a branch whose tx type does not require it.

One blocker, which is a one-liner, plus a fund-safety thing this PR puts within arm's reach.

---

### preferably-blocking: `meta.priceImpact` is unvalidated, so `null` lands in a `number` field

`fetchSwapKitQuoteResponse` ends in a bare cast:

```ts
return (isRecord(data) ? data : {}) as SwapKitQuoteResponse
```

Nothing between the proxy's JSON and `SwapKitQuoteRoute`. So `meta` is whatever came over the wire, and the new guard is:

```ts
if (meta?.priceImpact !== undefined) {
  return meta.priceImpact
}
```

`null !== undefined` is `true`. A route carrying `"meta": {"priceImpact": null}` returns `priceImpact: null` on a field typed `priceImpact?: number`, the `undefined` fallback to `totalSlippageBps` is skipped, and the consumer that renders the row does `priceImpact.toFixed(2)` on a null. Same story for a stringified `"0.0133"`, which would then render as `0.0133%` off a `${x * 100}`.

Suggested:

```ts
if (typeof meta?.priceImpact === 'number' && Number.isFinite(meta.priceImpact)) {
  return meta.priceImpact
}
```

Worth the same treatment on `totalSlippageBps` while you are in there, for the same reason.

### should-fix: the repo already has a price-impact ceiling and this route is not behind it

`packages/core/chain/swap/general/priceImpactGuard.ts` exists because of audit finding SOL-02 (#1049): a thin-pool Jupiter quote at 50-90% impact still built a fully signable MPC-ready swap. `MAX_PRICE_IMPACT_PCT = 10` and its own docstring says "a false refusal is recoverable, a false pass drains funds".

Its only two enforcement sites today are Jupiter:

- `packages/core/chain/swap/general/jupiter/api/getJupiterSwapQuote.ts:124`
- `packages/sdk/src/tools/swap/jupiter.ts:284`

Which means after this PR we know a SwapKit route's impact and still hand back calldata at any value. Probe on this head, with both controls:

| case | result |
|---|---|
| must-fail control: `assertJupiterPriceImpactWithinCeiling('0.45')` | THROWS `PriceImpactTooHighError` |
| must-pass control: `assertJupiterPriceImpactWithinCeiling('0.045')` | ok |
| SwapKit route, `meta.priceImpact: 0.45` | `priceImpact: 0.45`, `tx.evm.to: '0xrouter'`, `tx.evm.data: '0xabcdef'` |

The controls discriminate, so that third row is real: 45% impact, fully signable, while the identical Jupiter route is refused. Not a regression you are introducing (SwapKit exposed nothing before, so there was nothing to gate on), which is why this is should-fix and not blocking. But this PR is exactly the moment the figure becomes available, and the guard is one call away. If you would rather land the surfacing alone and gate separately, fine, just file it so it does not sit.

### should-fix: fraction vs percent, and the guard's primary entry point takes percent

Related to the above and worth its own line because it is a live trap for whoever wires it. `GeneralSwapQuote.priceImpact` is documented as a signed **fraction**. `evaluatePriceImpactPercent` takes **percent**. Feeding one to the other is a silent 100x, and since the guard fails safe on anything it cannot read, it fails safe in the wrong direction:

```
evaluatePriceImpactPercent(0.45)  -> { ok: true }   // 45% impact read as 0.45%
```

`evaluateImpactFromFractionString` is the correct bridge. Suggest naming the unit in the field itself (`priceImpactFraction`) or at minimum saying "fraction, not percent" in the docstring loudly enough that the next person does not reach for the obvious-looking function.

### q: "The fee is display-only here" is not quite true

`getSwapKitEvmSwapFee`'s docstring, and the changeset, both say display-only. `affiliateFee` has two real consumers:

- `findSwapQuote.ts:360` for ranking. Genuinely a no-op for you: `getGeneralDestinationSideFeeAmount`'s `providerAlreadyNet` list already contains `'swapkit'`, so the newly-populated fee is never subtracted from comparable output. Calling this out explicitly so nobody re-raises it as a quote-ranking regression, because it is not one.
- `mpc/keysign/swap/build.ts:211-226`, which maps it into the `OneInchTransaction` protobuf carried in the keysign payload as `swapFee` / `swapFeeChain` / `swapFeeTokenId` / `swapFeeDecimals`.

The signed EVM fields are from/to/data/value/gas, so there is no calldata or signature impact and I am not escalating this. But it does mean the fail-open catch makes keysign-payload bytes vary with a transient resolver hiccup, which is a different thing from a UI row being blank. Worth the docstring saying "not part of the signed EVM tx" rather than "display-only".

### q: what actually pins `meta.priceImpact` as a fraction?

There is no real SwapKit response fixture in the tree, so the stubs assert the code's own assumption rather than the API's behavior. The `totalSlippageBps / 10_000` fallback is self-consistent with the fraction convention, but if SwapKit publishes `meta.priceImpact` as a percent then `meta` and the fallback are 100x apart on a user-visible figure and the tests stay green either way. A captured response, or a link to the field's docs, would settle it.

### q: price impact and slippage are not the same quantity

The fallback labels one as the other. Every other `slippageBps` in this repo (`DEFAULT_JUPITER_SLIPPAGE_BPS`, balancer, astroport) is user-set *tolerance*, not realized directional movement, so the name collides with an established meaning here. `totalSlippageBps` may well be realized and directional on SwapKit's side, in which case a one-line comment saying so would help, since the surrounding code trains you to read it the other way.

### q: impact comes from the quote-stage route, everything else prefers the swap-stage response

```ts
dstAmount: parseExpectedBuyAmount(swapResponse.expectedBuyAmount ?? route.expectedBuyAmount, ...)
const routeProvider = getRouteProviderName(swapResponse) ?? getRouteProviderName(route)
const priceImpact = routePriceImpact(route)
```

Two of three prefer `swapResponse`. Impact only ever reads `route`. If the swap-stage response can restate impact, the displayed figure is one stage staler than the amount shown next to it. Deliberate, or just not available on that shape?

### nit: `amount > 0n` versus "when the response itemizes none"

```ts
...(affiliateFee && affiliateFee.amount > 0n ? { affiliateFee } : {}),
```

The changeset and the `BuildEvmTxInput` comment both say the fee stays absent "when the response itemizes no affiliate/service fee at all". It is also absent for an itemized zero, and for an unresolvable shape. Three distinct causes collapsing to one absent value is fine as a decision, but the doc describes only one of them.

### nit: the catch is unconditional

`getSwapKitEvmSwapFee` swallows everything `getSwapKitSwapFee` can throw, including a genuine programming error in it, and reports it via a bare `console.warn` from library code. Narrowing to the expected shape-resolution failure would keep the fail-open where you want it without hiding the class of bug that should be loud.

---

Ran on `513a1e92c`: `getSwapKitQuote.test.ts`, `priceImpactGuard.test.ts`, `findSwapQuote.selection.test.ts`, 109/109 green. The probe above was run against the same head and removed after.

Changeset semver looks right to me, `minor` on both packages for an additive field.

gomesalexandre (approved):
Current head `d1473bc67ee3de9db8a1c41bf255a7f445e47aad` looks good.

What I checked on this head:
- `meta.priceImpact` and `totalSlippageBps` are both narrowed to finite numbers before the new `priceImpactFraction` field is populated, so the original null / string / NaN fail-open on a typed `number` field is closed.
- the unit is now explicit at the type boundary. `priceImpactFraction` is documented as a fraction, and the doc points callers away from the percent-based guard entrypoint.
- `SwapKitFeeShapeError` cleanly scopes the EVM fail-open path so unresolved fee shapes do not kill an otherwise signable EVM route, while non-shape bugs stay loud.
- the added tests are load-bearing for the current fixes: fee surfaced on EVM, fee-shape fallback, meta-first impact, slippage-bps fallback, and invalid-meta fallthrough.
- current-head CI is green: Unit Tests, Integration Tests, Lint and Type Check, and Test Summary.

Non-blocking nit only: worth giving `SwapKitFeeShapeError` a stable local `code` property in a follow-up so consumers do not depend only on `instanceof`.

LGTM.

PR #1821, opened 2026-08-09
fix(sdk): align XRP and TON signing hashes with WalletCore

### LINKED ISSUE #1434: XRP/TON - RN-JS builder pre-signing hash diverges from WalletCore (co-sign fund-safety if dispatched)
## the finding (surfaced by E4 cross-encoder golden-vector coverage)

The RN-JS transaction builders for **Ripple and TON produce a pre-signing hash that does NOT match WalletCore's `getPreSigningHashes`** for the same intent. Cosmos + Solana **do** match (proven bound-and-passing by the #1376 golden fixtures), so this is not a "both sides differ by design" situation - XRP/TON are anomalous vs the other RN-JS builders.

## evidence (identical intents, probed both encoders)

**Ripple** - Payment (dest, amount 1000000, fee 12, seq 5, LLS 32000000, pubkey 031B84...):
- WalletCore `getPreSigningHashes` (what the SDK actually MPC-signs, via `TransactionBuilder.extractMessageHashes` -> `getEncodedSigningInputs`): `d66908156861...ebafea57`
- RN-JS `buildXrpSendTx` (ripple-binary-codec `encodeForSigning` -> sha512half): `e0fdca3cb090...`
- Adding `tfFullyCanonicalSig` does NOT reconcile them (`f614...`). Root cause is a deeper serialization/field difference, not the canonical-sig flag.

**TON** - WalletV4R2, identical pubkey/to/amount/seqno/expireAt/bounce:
- WalletCore `getPreSigningHashes`: `1956...`
- RN-JS `buildTonSendTx.signingHashHex`: `e791...` -> **DIVERGE**. Candidate root cause: stateInit / sendMode / body-cell serialization difference.

## why it matters

The SDK's own signing path is internally consistent (it signs via `getEncodedSigningInputs` = WalletCore), so **pure-SDK signing is safe**. The risk is narrow but real: `buildXrpSendTx` / `buildTonSendTx` explicitly return `signingHashHex` "for the EdDSA/ECDSA signer". If the flagship RN app ever dispatches THOSE hashes into an MPC ceremony while a peer device signs via WalletCore (mixed-platform 2-of-2), the two co-signers would be signing **different messages** for XRP/TON -> ceremony fails at best, and is a fund-safety hazard at worst. Cosmos/Solana are immune because their RN-JS builders are held to WalletCore parity.

## the ask (needs flagship-app visibility - can't confirm from the sdk clone)

1. Confirm whether `buildXrpSendTx` / `buildTonSendTx` `signingHashHex` is **live in any dispatched MPC path** (flagship RN fast-vault / any 2-of-2 where the other party signs via WalletCore). If it's never dispatched (only WalletCore hashes are ever signed), this is a latent/dead-code divergence -> low severity, still worth aligning.
2. If it IS dispatched: root-cause the serialization gap and bind the RN-JS builder to WalletCore's output (same treatment #1376 gave cosmos/solana). Add a cross-encoder golden fixture pinning BOTH sides to the same hash so it can't regress.

## severity

P2 with a **fund-safety ceiling** (escalates to P1/P0 if step 1 confirms live mixed-platform dispatch). Filed off an honest E4 probe - the coverage extension caught real drift rather than rubber-stamping a fixture.

Note for agents: reproduce by probing `buildXrpSendTx`/`buildTonSendTx` `signingHashHex` vs `getPreSigningHashes` (WalletCore) on identical intents; compare against the cosmos/solana bind in #1376 which passes. Related: WYSIWYS / signing-stack parity theme.

rcoderdev:
Closes #1434

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XRP and TON transaction signing-hash compatibility across platforms.
  * Corrected XRP memo and flag handling for more consistent transaction construction.
  * Improved TON native and Jetton transfer encoding, including inactive destinations and memo handling.
  * Added validation for TON inline memo capacity.

* **New Features**
  * Added optional control for whether a Jetton destination is active, while preserving existing defaults.

* **Tests**
  * Added cross-platform validation using WalletCore-compatible XRP and TON transaction scenarios.


gomesalexandre (changes requested):
Read the whole diff, checked out `449c0e191`, ran the affected suites green (43/43), and drove the TON jetton builder directly with a probe on both `main` and this head. The core motivation is legit and I'm not arguing with the approach: a signing-hash mismatch between WalletCore and `@ton/core` deadlocks or corrupts a mixed-platform MPC ceremony, so pinning our builders to WalletCore's canonical cell shape is the right call, and throwing rather than silently truncating a memo is the right fail-closed posture.

Blocking on the semver label and on the docs, not on the change itself.

---

### preferably-blocking: this is a consumer-visible capability regression shipped as `patch`

`.changeset/xrp-ton-walletcore-signing-parity.md` says `patch`. The jetton memo path goes from "any memo up to 123 bytes" to "throws above ~34-39 bytes". A consumer on a caret range picks that up automatically and gets a hard build-time throw on input that worked yesterday.

Same probe, same args, only the branch differing. `amount: 1_000_000n`, memo = N ASCII bytes:

| memo bytes | `main` (`68df30113`) | this head (`449c0e191`) |
|---|---|---|
| 10 | ok | ok |
| 30 | ok | ok |
| 38 | ok | ok |
| 39 | ok | ok |
| 40 | ok | **THROW** (352 bits, 351 available) |
| 60 | ok | **THROW** (512 bits, 351 available) |
| 100 | ok | **THROW** (832 bits, 351 available) |
| 123 | ok | **THROW** (1016 bits, 351 available) |

`main` accepts all eight, so the control discriminates. This is a real narrowing of the public contract, not a probe artifact.

Not asking you to revert it. Asking for `minor` at minimum, and a changeset body that names the memo cap explicitly so whoever bumps the dep reads it. Right now the changeset says "align signing hashes" and a consumer has no way to know their 80-byte memo just became a throw.

### should-fix: the documented cap is wrong for typical inputs, because it's amount-dependent

`jettonTransfer.ts`:

```
/** Optional UTF-8 forward comment; must fit WalletCore's inline Jetton payload (39 ASCII bytes for typical inputs). */
```

39 is only true at the amount you happened to pin. `storeCoins` is VarUInteger 16, so a bigger amount eats more of the 1023-bit cell and the memo cap shrinks under it. Fixed 39-byte memo, sweeping the amount on this head:

```
AMT                                1  -> ok
AMT                          1000000  -> ok
AMT              1000000000000000000  -> THROW (344 bits, 311 available)
AMT  1000000000000000000000000000000  -> THROW (344 bits, 271 available)
AMT  1267650600228229401496703205376  -> THROW (344 bits, 271 available)
```

`10^18` is one token of an 18-decimal jetton. At 311 available bits the real cap is `(311 - 1 - 32) / 8 = 34` bytes, not 39. So the same memo passes or fails purely on amount magnitude, and the docstring promises the best case.

Suggested: say the cap varies with the encoded amount and floor it at the pessimistic value, e.g. "at most 34 ASCII bytes; the exact cap shrinks as the amount grows".

### should-fix: the regression test pins one amount, so it can't catch the shrink

`ton-golden-vectors.test.ts:293` asserts 39-ok / 40-throws at `amount: 5_000_000n`. That certifies the best case and will stay green forever while a large-amount caller throws at 35 bytes. Worth adding a case at `10n ** 18n` so the amount-dependence is pinned rather than incidental.

### should-flag: dropping `MemoType` is also consumer-visible

```
-          MemoType: utf8ToHex('text/plain').toUpperCase(),
```

Understood why (WalletCore's raw-JSON memo path doesn't emit it, and you need the preimages to match), but this changes what lands on-chain for every XRP send carrying a memo. Anything downstream that reads `MemoType` off a Vultisig-built XRP payment stops seeing it. Same `patch` changeset. Worth a line in the changeset body at minimum.

### q: `isActiveDestination` sets `forward_ton_amount` to 0, which kills the notification

```
.storeCoins((opts.isActiveDestination ?? true) ? JETTON_FORWARD_AMOUNT_NANO : 0n)
```

`JETTON_FORWARD_AMOUNT_NANO = 1n` exists specifically because it's "the minimum that triggers a transfer_notification" (its own comment). Passing `isActiveDestination: false` zeroes it, so the recipient contract never gets notified. Is that intentional WalletCore parity for the uninitialized-destination case, or a side effect? Also worth noting no caller passes the flag today, so it's plumbed but dead, which means nothing exercises the `false` arm outside the goldens.

### q: `...(hasDistinctMemo ? {} : { Flags: 0 })`

This encodes WalletCore's internal path selection (protobuf vs raw-JSON) into our builder shape. It's golden-pinned so it won't silently drift, but it does mean any WalletCore change to which path handles memos breaks us in a way the goldens will catch loudly rather than the code preventing. Fine to keep, just flagging that the comment is carrying a lot of weight there.

### nit

```
if (opts.memo && hasDistinctMemo) {
```

`opts.memo &&` is already folded into `hasDistinctMemo` two lines up. Redundant.

---

Two things I chased and want to explicitly clear so nobody re-raises them:

- `storeSlice(commentCell.beginParse())` looked like it could drop refs past a bits-only capacity check. It can't. `buildCommentBody` caps at 123 bytes, so the comment cell is 1016 bits max and always flat. The bits-only check is sound.
- No TDZ on `legacyMemoDestinationTag`. Declared at 231, first used at 240.

Ran: `ton-golden-vectors`, `jettonTransfer`, and the two ripple suites, 43/43 green on `449c0e191`.

gomesalexandre (approved):
## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | changeset shipped `patch` for a consumer-visible capability regression (jetton memo cap) | ✅ fixed | bumped to `minor`, `aaa7e0d` |
| r1 | 2 | should-fix | jettonTransfer.ts docstring claimed a flat 39-byte memo cap; it's amount-dependent | ✅ fixed | reworded in `jettonTransfer.ts` + `chains/ton/tx.ts`, `aaa7e0d` |
| r1 | 3 | should-fix | regression test only pinned one amount, couldn't catch the cap shrink | ✅ fixed | added `10n ** 18n` case in `ton-golden-vectors.test.ts`, `aaa7e0d` |
| r1 | 4 | should-flag | `MemoType` removal is consumer-visible, undocumented | ✅ fixed | now explicit in changeset body |
| r1 | 5 | q | `isActiveDestination` zeroing `forward_ton_amount` — intentional? | ℹ️ resolved, not a gap | read `packages/core/mpc/keysign/signingInputs/resolvers/ton/jetton.ts` — `forwardAmount = isActiveDestination ? 1n : 0n` is pre-existing core resolver logic, unchanged by this PR, already feeds real WalletCore's protobuf. Not a new side effect. |
| r1 | 6 | q | `Flags: 0` conditional on memo-path selection carries a lot of weight in one comment | ℹ️ documented, not a gap | still true, kept as-is — flagging only, no action asked |
| r1 | 7 | nit | `opts.memo &&` looked redundant with `hasDistinctMemo` | ❌ not applied — it's load-bearing | tried removing it, `tsc` failed (`opts.memo` is `string \| undefined`, `hasDistinctMemo` being `true` doesn't narrow it). Reverted. My prior nit was wrong, correcting the record. |

## verdict: APPROVE, pushed a small follow-up commit

Verdict up front: this is fund-safety-critical (signing-hash alignment with WalletCore for mixed-platform MPC) and I went back through it with the same rigor as sdk#1844 — independently re-verified the hash-alignment claim rather than trusting the diff, then closed out the semver/docs blockers from my own prior review since nobody had touched the branch in a day.

## what I independently re-verified (not just re-reading my old review)

**The actual cross-platform hash match is real, not asserted.** `packages/core/mpc/tx/preSigningHashes/crossEncoder.golden.test.ts` boots real WalletCore WASM (`initWasm()`) and runs the SigningInput through `getPreSigningHashes` — genuine WalletCore hash computation, not a reimplementation. `packages/sdk`'s RN-JS builders (`buildXrpSendTx`, `buildTonSendTx`, `buildTonJettonTransferTx`) load the *same* JSON fixtures from `testdata/cross-encoder-golden/` and assert their own `signingHashHex` against the identical `expectedSigningHashHex`. Ran both suites myself: 6/6 core cross-encoder tests pass against real WalletCore, 44/44 RN golden-vector tests pass (43 original + 1 I added). This is a real independent oracle, not a self-referential fixture.

**Hunted for edge-case coupling the fixtures might miss (3 rounds):**
1. *XRP DestinationTag + memo interaction* — read `platforms/react-native/chains/ripple/tx.ts` fully. `DestinationTag` inclusion (line ~259) is structurally independent of `hasDistinctMemo` (which only gates `Flags` and `Memos`). No fixture covers DestinationTag+memo together against real WalletCore, but the two inclusion rules don't interact in either the RN code or the core resolver (`getRippleSigningInputs`, reviewed in depth on sdk#1844), so this isn't a blind spot, it's ruled out by code structure.
2. *Legacy-memo-as-tag-carrier path selection* — when a numeric memo is inferred as a legacy DestinationTag carrier, `hasDistinctMemo` is false and it takes the protobuf/`Flags:0` path, matching the core resolver's identical `distinctMemo = inferredLegacyCarrier ? undefined : memo` logic. Consistent.
3. *TON `isActiveDestination`* — see the "prior issues" table above. Not a new risk; pre-existing core resolver logic this PR is aligning RN to, not introducing.

**Ran everything:** full SDK unit suite (3282 tests, all pass), full core suite (2618 tests, all pass), `crossEncoder.golden.test.ts` specifically (6/6), typecheck clean, lint clean, `build:sdk` (all platform bundles, exit 0).

## what I pushed (`aaa7e0d`)

Closed out my own still-open r1 findings since nobody had addressed them in a day: `minor` semver bump with the memo-cap-regression and `MemoType`-removal explicitly called out in the changeset body, corrected the amount-dependent memo-cap docstrings, added an amount-shrink regression test (`10n ** 18n`, one token of an 18-decimal Jetton — a realistic amount, not an adversarial edge case). Also tried the nit from my prior review (dropping the redundant-looking `opts.memo &&`) and it broke `tsc` — it's load-bearing for narrowing `opts.memo: string | undefined`, not actually redundant. Reverted, noted in the table so it doesn't get re-flagged.

No close-ref needed — this PR has no linked issue (automated `codex/continuous` branch).

PR #1823, opened 2026-08-09
Add versioned signable transaction contract

### LINKED ISSUE #1447: architecture: define a versioned signable-transaction contract across builders and clients
## Goal

Create the SDK-owned, versioned contract that binds a transaction's canonical decode, user-visible material fields, and the exact unsigned payload. This is the shared boundary behind “what you see is what you sign.”

## Why

Current transaction construction, display and signing checks are distributed across clients and agent execution paths. Local guards are valuable, but each layer can still parse or prioritize an envelope differently. A change can therefore be safe in one consumer and misleading in another.

## Scope

- Define a versioned canonical signable-transaction envelope: intent, chain/account, exact unsigned payload, payload hash, canonical decoded actions, material display fields, display hash, expiry and one-time approval binding.
- Define canonical action types and required material fields for send, token send, approval, swap, contract call, deposit, withdraw and redeem.
- Provide chain-family decoder/normalizer interfaces. Unknown or incomplete semantics must not be represented as an ordinary approved action.
- Provide deterministic canonicalization/hashing and shared fixture format.
- Specify the equality/fence policy: recipient, source account, asset, base-unit amount, memo, spender/allowance, min-out, expiry and chain are exact; fee/slippage may vary only in explicit displayed bounds.

## Non-goals

- Do not migrate all clients in this issue.
- Do not introduce a blind generic transaction renderer.
- Do not weaken existing chain-specific guards while the contract is adopted.

## Acceptance criteria

- [ ] Schema and canonicalization are SDK-owned and versioned.
- [ ] Fixtures cover pass and reject cases for payload/display divergence, recipient/amount/memo/spender/min-out/expiry changes, wrong account, replay and hidden batch actions.
- [ ] A consumer can verify a payload against the same decoded material fields it renders.
- [ ] Migration guidance names app, backend, iOS, Android and desktop owners.

rcoderdev:
Closes #1447

## Summary by CodeRabbit

* **New Features**
  * Added Signable Transaction Contract v1 to the SDK.
  * Added schema validation, canonical serialization, hashing, decoding, envelope creation, and transaction verification.
  * Added approval binding and one-time replay protection.
  * Added shared fixtures for cross-platform transaction validation.
  * Exposed the functionality through the standard and React Native SDK entry points.

* **Documentation**
  * Added documentation covering the contract lifecycle, validation rules, approvals, migrations, and shared fixture format.

### DIFF
--- a/.changeset/slow-coins-sign.md
+++ b/.changeset/slow-coins-sign.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': minor
+---
+
+Add the versioned signable-transaction v1 schema, canonical hashing, decoder interface, approval binding, verifier, and shared fixture format.
--- a/docs/signable-transaction-contract-v1.md
+++ b/docs/signable-transaction-contract-v1.md
@@ -1,0 +1,44 @@
+# Signable transaction contract v1
+
+The SDK owns a versioned boundary between transaction construction, user-visible review, approval, and signing. Version 1 binds the exact unsigned payload to canonical decoded actions and to the exact material fields shown to the user.
+
+## Contract flow
+
+1. A chain-family adapter decodes the unsigned payload and normalizes it to the v1 action union.
+2. `createSignableTransactionEnvelopeV1` rejects unknown or incomplete semantics, checks action context and explicit fee/slippage bounds, and builds the material display fields.
+3. The envelope records the exact payload and the SHA-256 digest of its decoded bytes, canonical actions, the display and its canonical policy digest, expiry, and a one-time approval binding. The display digest binds material fields and explicit fee/slippage bounds; the current bounded values are validated against the candidate decode.
+4. The approval store persists the full `envelope.approval` receipt independently of the candidate envelope. A consumer renders `envelope.display` with its action-specific UI. Immediately before signing it calls `verifySignableTransactionEnvelopeV1` with the candidate payload, that same rendered display, the trusted approval receipt, current time, and an atomic approval-binding reservation callback.
+5. The verifier calls that callback only after every semantic and equality fence passes, and returns valid only when the store atomically reserves the previously unused `approvalBindingKey`. The reservation happens before dispatching a sign request. A failed sign attempt does not release it; retry requires a fresh approval. Concurrent or repeated attempts must fail with `approval-replay`.
+
+The verifier re-decodes the candidate payload and compares the independently loaded approval binding digest. It does not trust an envelope merely because its hashes are internally consistent.
+
+## Exact and bounded fields
+
+The normalized action schemas make chain, source account, recipient, asset, base-unit amount, memo, spender/allowance, swap minimum output, and action count exact. Expiry is exact at the envelope level. Any change is a rejection.
+
+Decoded fee and slippage values are exact for the bound unsigned payload. A builder may declare an explicit displayed minimum and maximum when the value is created; the actual decoded value must remain inside those bounds. Missing bounds default to `min = value = max`. Bounds never authorize mutation of the already-bound unsigned payload.
+
+Amounts, limits, fees, and basis points are base-10 integer strings. This avoids floating-point and bigint serialization drift. Unsigned payload encodings are canonical (`hex` is lowercase, even-length and prefix-free; `base64` is padded).
+
+## Decoder rule
+
+Each adapter implements both `decode` and `normalize`. `decode` may return a chain-native AST. `normalize` must return either:
+
+- `status: "decoded"` with complete canonical actions and any fee/slippage observation; or
+- `status: "unknown"` with a reason.
+
+Never coerce an unknown selector, hidden batch member, partial instruction, unrecognized message, or incomplete nested call into an ordinary approved action. Existing chain-specific guards remain in force during migration.
+
+## Consumer migration owners
+
+Migration is deliberately outside issue #1447. Owners should adopt v1 without deleting their current chain-specific fences:
+
+| Surface       | Owner                                            | Adoption responsibility                                                                                                                |
+| ------------- | ------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------- |
+| Shared app    | `vultisig-app` transaction-review owner          | Render the v1 material fields with action-specific UI and pass the exact rendered display to verification.                             |
+| Agent backend | `agent-backend-ts` execution owner               | Create/persist envelopes, allocate approval id/nonces, and atomically reserve approval binding keys before dispatching a sign request. |
+| iOS           | `vultisig-app` iOS transaction-signing owner     | Add family decoders/normalizers and keep current iOS chain-specific confirmation guards until parity fixtures pass.                    |
+| Android       | `vultisig-app` Android transaction-signing owner | Add family decoders/normalizers and keep current Android chain-specific confirmation guards until parity fixtures pass.                |
+| Desktop       | `vultisig-app` desktop transaction-signing owner | Carry the envelope across desktop/extension boundaries without re-parsing or reprioritizing material fields.                           |
+
+Every consumer should run the shared `SignableTransactionFixtureV1` format. The fixture candidate contains the unsigned payload, normalized decode result, rendered display, approval state, time, consumed bindings, and expected issue codes, so implementations in other languages can replay the same pass/reject cases.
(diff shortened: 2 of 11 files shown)


gomesalexandre (approved):
Read the whole diff, checked out `3033ecdfe`, ran the two shipped suites green (13/13 across `signableTransactionContract` + `signableTransactionExports`), and drove the binding directly with a probe carrying both a must-pass and a must-fail control. This is a good piece of work. The binding chain is the part I went hardest at and it holds up, so I want to record what I actually tried rather than just say "looks good".

**What I attacked, and what caught it.** Probe on this head, `contract-call` envelope, only the named field moved:

| attack | result |
|---|---|
| must-pass control: send envelope verified against its own decode | `valid: true` |
| must-fail control: candidate decodes to a different `calldataHash` | `decoded-actions-divergence` + `display-divergence` |
| envelope has no `method`, candidate decodes one in | `decoded-actions-divergence` + `display-divergence` |
| batch `[send, contract-call]`, honest decode | `valid: true` |

Controls discriminate, so those are real. `payloadHash` -> `displayHash` -> `bindingHash` -> independently loaded trusted receipt, plus a re-decode-and-compare of `actions` and of the fee/slippage values against `getApprovedDisplayBoundsV1`, means every swap I could construct lands on a divergence code. Deliberately hashing only `min`/`max` for fee/slippage and re-checking `value` separately against a fresh decode is the right split and it is documented.

Worth calling out explicitly so nobody re-raises it: `canonical.ts` is materially better than the canonicalizer in #1811. No `$bigint`/`$bytes` tags so there is no tag-collision class, `undefined` throws in both object-value and array position rather than being dropped in one and throwing in the other, non-plain prototypes rejected, cycles rejected, non-safe-integer numbers rejected. If those two ever converge, this is the one to keep.

Two more I chased and want to clear rather than leave hanging, because both look like bugs until you check the gate:

- `decodeCanonicalBase64`'s `indexOf` returning -1 would corrupt silently, but `hashSignableUnsignedPayloadV1` schema-parses first and the `superRefine` pins canonical padded base64 with zero unused padding bits, so length is always a multiple of 4 and every char is in the alphabet. Not reachable.
- `compareDecimal`'s bare `BigInt(left)` would happily eat `''` -> `0n` and `'0x10'` -> `16n`, but `decimalInteger`'s `/^(0|[1-9]\d*)$/` gates every value that reaches it. Also not reachable.

No blockers. Everything below is should-fix or lower.

---

### should-fix: `contract-call` is the coercion sink the doc says must not exist

`docs/signable-transaction-contract-v1.md` is unambiguous:

> Never coerce an unknown selector, hidden batch member, partial instruction, unrecognized message, or incomplete nested call into an ordinary approved action.

`contract-call` is a first-class member of the action union, and its entire material content is `contract` + `value` + an opaque digest. Here is what a human is handed for one, verbatim from the probe:

```
      intent = contract-call
      chain = Ethereum
      sourceAccount = 0x1111...1111
      expiresAt = 2030-01-01T00:00:00.000Z
      actions.0.calldataHash = sha256:aaaa...aaaa
      actions.0.chain = Ethereum
      actions.0.contract = 0xdead...beef
      actions.0.type = contract-call
      actions.0.value = 0
```

An adapter that cannot decode a selector has two options: return `{status: 'unknown'}` and fail closed, which is what the doc wants, or normalize it to `contract-call` and get a fully approved action. The second is cheaper and the schema sanctions it. Via `batch` it composes, so an unreadable member sits next to a readable send and the batch verifies clean (`valid: true` in the table above).

To be precise about severity: this is not a binding hole. The `calldataHash` is exactly bound, substitution is caught, and I proved that. The gap is human-evaluability, which is what the whole material-fields concept is for. A digest is not a material field in any sense a person approving can use.

Not asking you to drop the type. Asking for one of: a `SignableVerificationIssueCodeV1` or envelope-level marker that says "this envelope contains an action whose semantics were not decoded" so a consumer can render a warning rather than a normal approval sheet, or a doc paragraph naming `contract-call` as the deliberate escape hatch and stating who is allowed to emit it. Right now the doc reads as if the escape hatch is not there.

### q: ~30 public exports, zero in-tree consumers, shipped `minor`

`export *` from both `src/index.ts` and `platforms/react-native/index.ts`, and the doc says migration is out of scope for #1447. Fine as a decision, and the consumer-owner table is a nice touch. But every one of those names is now public API on a caret range with nothing exercising it against a real decoder, so the first adapter is going to find shape problems that a `0.x`-style internal iteration would have absorbed. Worth saying in the changeset that v1 is provisional until the first adapter lands, or holding the barrel export until then and keeping the deep import path for the pilot consumer.

Changeset semver itself looks right to me: `minor` on `@vultisig/sdk` only. It imports `Chain`/`ChainKind`/`getChainKind` from `@vultisig/core-chain` but changes nothing there, and everything routes through `src/index.ts` so no export-map entry is needed.

### nit: a whitespace-padded `sourceAccount` fails closed with a message that blames the decoder

`nonEmptyString` is `z.string().trim().min(1)`, so it transforms. `createSignableTransactionEnvelopeV1` parses `unsignedPayload` and `approval` up front and then uses the parsed values, which is exactly right and I can see it was deliberate given the `' trimmed-id '` test. `sourceAccount` does not get the same treatment: it flows raw into `decodeSignableTransactionV1`'s context and into `buildDisplay`, while the decoded actions come back through `signableDecodedTransactionV1Schema.parse` already trimmed.

Probe result:

```
sourceAccount: ' 0x1111...1111 '  (identical string on both the envelope and the action)
-> THROWS  Decoded action 0 does not match the envelope chain/source account
```

Fail-closed, which is the important half, and the reason I am filing this as a nit rather than higher. But the caller passed the same string in both places and gets told the decoder disagreed with the envelope. Parsing `sourceAccount` alongside `unsignedPayload` and `approval` at the top of the function, and using the parsed value from there on, makes the trim visible and the error impossible.

### nit: `flattenMaterialValue` flattens `null` and the string `'null'` to the same cell

Keys join with `.` and values go through `String(value)`, so `null`, `'null'`, `0` and `'0'` are indistinguishable in `materialFields`. Low impact because `envelope.actions` is separately bound exactly and every action is `.strict()`, so no current shape can collide. It is only load-bearing if a future action type takes a free-form string field. A type tag on the cell, or JSON-encoding the scalar, closes it before that happens.

### nit: `exactEqual` casts away the types at the one place the function throws

```ts
const exactEqual = (left: unknown, right: unknown): boolean =>
  canonicalizeSignableTransactionValue(left as never) === canonicalizeSignableTransactionValue(right as never)
```

`canonicalizeSignableTransactionValue` throws on cycles, non-plain prototypes and unsafe integers, and `as never` is what stops the compiler from telling you which call sites can reach that. `collectEnvelopeBindingIssuesV1` also runs outside any try/catch, so a throw escapes as an exception rather than becoming `{valid: false}`. I could not reach it, everything upstream is zod-parsed first, so this is defense-in-depth only. Given the doc's totality claim it is cheap to wrap the binding collection the same way `verifyDecodedSemanticsV1` already is.

### nit: the 80-digit cap has no test naming its purpose

`maxDecimalIntegerDigits = 80` reads as arbitrary next to `expectIssueAtPath({...actions[0], amount: '1'.repeat(81)}, ['amount'])`. A one-line comment saying what it is bounding, presumably canonicalization/compare cost on adversarial input rather than any on-chain limit, saves the next person from assuming it is a uint256 thing and lowering it to 78.

---

Ran on `3033ecdfe`: `tests/unit/signableTransactionContract.test.ts` + `tests/unit/public/signableTransactionExports.test.ts`, 13/13 green. The probe above was a temporary file in `tests/unit/`, run against the same head and removed after.

One hypothesis I chased hard and want to explicitly clear, because it is the kind of thing that gets re-raised: the expiry check is `now.getTime() >= new Date(envelope.expiresAt).getTime()`, and an Invalid Date would make that `NaN >= NaN`, false, so the envelope would never expire. It is not reachable. zod v4's `.datetime()` calendar-validates, so `2026-02-30T00:00:00.000Z` and `2026-04-31T00:00:00.000Z` both fail `signableUtcTimestampV1Schema` before they can reach `expiresAt`, and V8 rolls those over rather than producing NaN anyway. Belt and braces. The asymmetric `now` parsing (`Number.isFinite` guarded on the `Date` branch, unguarded on the string branch) is doing real work on the branch that needs it.

APPROVE. The should-fix on `contract-call` is worth a follow-up issue rather than a block, since the doc is the thing that is wrong about it rather than the code.

PR #1824, opened 2026-08-09
Test dangerous-address public exports

rcoderdev:
## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for safety-related address constants and helper utilities.
  * Verified these exports are consistently available through both the SDK’s primary entry point and React Native integration.
  * Added identity checks to ensure all entry points reference the same canonical values and implementations.
  * Helps prevent accidental inconsistencies when consuming these safety utilities across supported SDK platforms.


gomesalexandre (approved):
Read the diff, checked out `c20970de7`, ran both touched files green (60/60 via `--config tests/unit/vitest.config.ts`). Test-only, no runtime surface, and it pins something worth pinning: an identity check catches a shadowed or re-implemented re-export, which for a dangerous-address allowlist is the failure that actually hurts. A `toBe` here rather than a `toBeDefined` is the right call, since a duplicate implementation would pass the weaker assertion while silently giving RN consumers a different blocklist from SDK consumers.

Covering the RN barrel separately from the root barrel is the good half. RN has its own entrypoint and its own registration side effects, so "it is exported from `src/index`" was never evidence that RN consumers got it.

APPROVE. One nit below, take it or leave it.

### nit: the name list is hand-maintained, so a new export can still escape

`dangerousAddressCanonicalExports` enumerates 9 names, and `packages/core/chain/security/dangerousAddresses.ts` currently exports exactly those 9, so the list is complete today. But that is the state this test is meant to defend, and it cannot defend it: add a tenth export to the canonical module, forget to re-export it, and this suite stays green because the tenth name is not in the array.

Deriving it closes that:

```ts
const dangerousAddressCanonicalExports = Object.keys(dangerousAddresses) as (keyof typeof dangerousAddresses)[]

it.each(dangerousAddressCanonicalExports)('re-exports dangerous-address canonical %s by identity', name => {
  expect(sdk[name]).toBe(dangerousAddresses[name])
})
```

No over-assertion risk: all 9 exports of that module are intended public, and type-only exports are erased so they never show up in `Object.keys`.

Worth pairing with a `expect(dangerousAddressCanonicalExports.length).toBeGreaterThan(0)` guard, otherwise a bad import path turns the whole suite into zero cases that vacuously pass.

### nit: the comparison target is one hop short of canonical

`packages/sdk/src/utils/dangerousAddresses.ts` is a single line:

```ts
export * from '@vultisig/core-chain/security/dangerousAddresses'
```

so both sides of the `toBe` are re-export barrels. It still catches SDK-side shadowing, which is the case you are after. But "canonical" in the test name points at core-chain, and importing from there directly would make the assertion say what it means. Cosmetic, since ESM module instances are singletons and the values are identical either way.

---

Ran on `c20970de7`: `tests/unit/utils/publicExports.test.ts` + `tests/unit/platforms/react-native/entry.test.ts`, 60/60 green.

Unrelated to this PR but worth writing down since it cost me a detour: running these files with a bare `yarn vitest run <path>` instead of the package's own `--config tests/unit/vitest.config.ts` reports 26/26 failed on this branch and 18/18 failed on main, all `configureMpc: duplicate MPC engine instance`. The default config has no `setupFiles` and no `pool: 'forks'` / `isolate: true`, so the RN entry's module-load `configureMpc` collides across cases. Nothing wrong with the code and nothing for you to fix here, but the failure looks convincingly like a real pre-existing bug, so it may be worth a line in the contributing docs before it burns someone else.

PR #1825, opened 2026-08-09
Harden encoder parity before signing dispatch

### LINKED ISSUE #1439: Encoder-parity hardening: pre-dispatch parity guard + Cardano/Tron golden gaps
## Encoder-parity hardening: add a pre-dispatch parity guard + close 2 unverified golden gaps

From the encoder-parity coverage spike. The signing surface is largely safe (utxo/evm/cosmos/thorchain/solana/sui RN-JS builders all match WalletCore `getPreSigningHashes`, golden-proven), but three defense-in-depth items:

### 1. Missing pre-dispatch parity guard (the seatbelt)
Nothing anywhere cross-checks an RN-JS `signingHashHex` against WalletCore `getPreSigningHashes` before it's dispatched into a signing ceremony. That's the guard that would have caught the TON/XRP divergence (#1434) automatically and would catch any future builder drift. Add an assert in the prep/dispatch path (at least for the chains that emit an RN-JS signingHashHex - TON jetton today) that the RN-JS hash equals the WalletCore-derived hash, fail-closed (or log-and-fall-back-to-WalletCore).

### 2. Cardano - no byte-parity golden (med-low confidence gap)
RN bridge uses core-chain CIP-30 `cardanoTxBodyHash`; the WalletCore path uses `TW.Cardano.Proto` (`resolvers/cardano.ts:5`). Both are blake2b-256(CBOR body) but are SEPARATE impls that have never been byte-diffed. Add a golden vector asserting the two produce the identical hash for a fixed tx (like the #1376 cosmos/solana goldens).

### 3. Tron - parity asserted by comments only (med confidence gap)
`buildTronSendTx` parity vs WalletCore is asserted by code comments (`chains/tron/tx.ts:291` memo field 12) but the golden test re-builds the protobuf with the same field assumptions rather than diffing WalletCore's actual output. Field-order / optional-field (fee_limit=18, timestamp=14) parity is unproven vs WalletCore. Add a golden that diffs against WalletCore output specifically.

None are confirmed bugs - #2 and #3 are unverified-parity gaps, #1 is the defense-in-depth guard. Related: #1434 (de-escalated to availability/compat).

rcoderdev:
Closes #1439

## Summary by CodeRabbit

- **New Features**
  - TON transaction builders now provide WalletCore-compatible signing data for native and Jetton transfers.
  - Added public access to TON transaction result types and swap tools.
  - Added detailed vault import error codes for duplicate, stale, incompatible, and persistence-related conditions.

- **Bug Fixes**
  - TON fast signing now verifies signing-hash parity before dispatching operations.
  - Corrected TRC-20 encoding to omit default zero-valued fields.
  - CLI schema generation now waits for successful output completion and reports write errors.


gomesalexandre (changes requested):
Read the whole diff, checked out `e7fd871`, ran the three touched suites green (73/73 via `--config tests/unit/vitest.config.ts`), and ran a must-fail control against the TRON half to confirm the new cross-check actually bites.

| control | result |
|---|---|
| must-pass: all three suites unmodified | 73/73 pass |
| must-fail: put `fieldInt64(3, 0n)` back into `buildTriggerSmartContract` | `a TRC-20 transfer matches WalletCore through timestamp and fee_limit field ordering` FAILS, 1 failed / 44 passed |

**The TRON half is correct and well-evidenced, and I want to say that first because it is the strongest thing in this PR.** The control discriminates, so the proto3 default-field fix is load-bearing rather than cosmetic: writing an explicit field-3 zero really does change the raw bytes and therefore the signing hash. And the evidence is genuine, not circular. `tron-tx.test.ts` imports `initWasm, TW from '@trustwallet/wallet-core'` and the PR adds a real `buildWalletCoreTrc20SignedOutput()` that compares whole-output bytes. That is the right shape for a parity test.

Worth flagging honestly for anyone reading the green tick and stopping there: `tron-golden-vectors.test.ts` is **not** additional parity evidence. Its `referenceRawData` is built by the test file's own `encodeRawData` helper, so it proves self-consistency with itself, not agreement with WalletCore. All the parity weight sits on `tron-tx.test.ts`. Not a defect, just do not count it twice.

Placement of the TON gate is also right: it runs before retries, HTTP, relay and MPC, and the comment explaining that retrying a deterministic mismatch would only pile up duplicate signing sessions is exactly the reasoning I would want written down.

REQUEST_CHANGES on the TON half, for two reasons below.

---

### preferably-blocking: nothing in the repo can produce `walletCoreTxInputData`, so this bricks TON fast-signing

`walletCoreTxInputData` appears in exactly six places. Three are in the new test file. The other three are all inside `fastVaultSign.ts` itself:

```
packages/sdk/src/platforms/react-native/mpc/fastVaultSign.ts:37   walletCoreTxInputData?: Uint8Array
packages/sdk/src/platforms/react-native/mpc/fastVaultSign.ts:99   const txInputData = opts.walletCoreTxInputData
packages/sdk/src/platforms/react-native/mpc/fastVaultSign.ts:102  '...is required for TON...'
```

Zero producers. Nothing in `src/` constructs one, and `packages/sdk/src/chains/ton/tx.ts` cannot: it returns `signingHashHex` and no WalletCore `SigningInput` at all. Its own docblock at line 197 says:

> Pass `result.signingHashHex` to the EdDSA signer (e.g. Vultisig's schnorrSign).

`fastVaultSign` and `schnorrSign` are both public RN exports (`platforms/react-native/index.ts:151`), so the caller here is the consuming app, not internal code we control in this repo. An app that follows the SDK's own documented TON flow, unchanged, upgrades to this version and gets:

```
fastVaultSign: walletCoreTxInputData is required for TON;
refusing to dispatch without WalletCore encoder parity proof
```

That is a hard refusal on legitimate money movement, which is the failure direction that costs the most. A guard that fails closed on a real bug is good; a guard that fails closed on every correct caller is an outage.

Ask: land the producer in the same PR (have the TON builder return the WalletCore `SigningInput` alongside `signingHashHex`, so the documented flow satisfies the guard by construction), or land the guard behind the producer. Either ordering is fine. Shipping the requirement before anything can meet it is the part I would not merge.

### preferably-blocking: the parity claim itself is never tested against a real WalletCore

The new tests mock both sides of the comparison:

```ts
vi.mock('@vultisig/core-mpc/tx/preSigningHashes', () => ({ getPreSigningHashes: vi.fn() }))
vi.mock('../../../../src/context/wasmRuntime', () => ({ getWalletCore: vi.fn(async () => ({ source: 'walletcore' })) }))
```

and then feed the parity cases hand-chosen buffers (`new Uint8Array(32).fill(0x11)`, `new Uint8Array([0x01])`). So what these prove is: the gate runs pre-dispatch, and it throws when the two values differ. Both worth having. But the actual claim in the changeset, that the SDK "independently derives and matches the WalletCore pre-signing hash" for TON, is nowhere exercised. With `getPreSigningHashes` mocked, a real divergence between the SDK's TON hash and WalletCore's would sail past this suite.

That matters more than usual here because of the failure mode. If the two genuinely disagree today, this guard does not catch a bug, it refuses every TON sign. Combined with the finding above, an untested equality is being turned into a dispatch precondition.

The pattern you want is already in this PR, about thirty lines away: `tron-tx.test.ts` spins up real WalletCore via `initWasm` and compares real bytes. One TON case doing the same, unmocked, would convert this from "the plumbing works" to "the parity holds".

### should-fix: `patch` for a required-argument change plus a new runtime throw

`chain?: string` becomes `chain: string`, and a new code path throws where none did before. Consumers on a caret range pick this up silently. `minor` at minimum. Same call I made on #1811 and #1821, so this is consistency rather than a new standard.

### q: the gate covers TON, but TRON is the chain with the demonstrated divergence

This PR contains proof that encoder drift is real and reachable, and that proof is about TRON, not TON. The control above shows the TRC-20 bytes were genuinely wrong before this change. Yet the new pre-dispatch parity check is TON-only and TRON stays unguarded, protected by a unit test rather than a runtime gate.

Not asking for TRON coverage in this PR necessarily. But if the guard is worth a hard refusal, the chain with the actual history of divergence seems like the one that most wants it, and it would be good to know whether that is a deliberate sequencing decision or just where the work stopped.

### nit: optional in the type, mandatory at runtime

`walletCoreTxInputData?: Uint8Array` is optional in `FastVaultSignOptions`, so an existing TON caller gets no compile error on upgrade and discovers the new requirement as a runtime throw instead. If it is required for TON, the type cannot express that cleanly without a discriminated union on `chain`, which may not be worth it. Raising it because the same optional-in-type/required-at-runtime shape came up on #1811's `swapQuote`, and a doc comment on the field saying "required when `chain === 'ton'`" costs one line.

---

Ran on `e7fd871`: `tests/unit/platforms/react-native/fastVaultSign.test.ts` + `tests/unit/chains/tron-tx.test.ts` + `tests/unit/platforms/react-native/tron-golden-vectors.test.ts`, 73/73 green. The must-fail control edit to `src/chains/tron/tx.ts` was reverted, worktree clean.

To be clear about what I am blocking on: the TRON change I would take as-is today. It is the TON gate landing ahead of any way to satisfy it, on an equality that is itself only mock-tested, that I would want resolved first.

gomesalexandre (changes requested):
blocking: `fastVaultSign` now normalizes `messageHashHex` as a mandatory 32-byte hex string before it checks `opts.chain`, so this PR breaks the existing RN Solana fast-sign flow on current head.

Grounding from the current head:
- `packages/sdk/src/platforms/react-native/mpc/fastVaultSign.ts` calls `normalizeSigningHash('messageHashHex', opts.messageHashHex)` at the top of `assertPreDispatchSigningHashParity`, before the `if (opts.chain.toLowerCase() !== 'ton') return undefined` branch.
- that helper now rejects anything that is not exactly 64 hex chars / 32 bytes.
- `packages/sdk/src/platforms/react-native/chains/solana/tx.ts` explicitly documents and implements the opposite contract: `buildSolanaSendTx` returns `signingHashHex = bytesToHex(message)`, and the file-level comment says `Signing hash = message bytes verbatim (ed25519 signs the full message)`.

So a valid RN Solana caller that does:
- `const tx = buildSolanaSendTx(...)`
- `await fastVaultSign({ chain: 'solana', messageHashHex: tx.signingHashHex, ... })`
now fails locally before any TON-only parity logic is even consulted.

This is not an edge case on message size either. From the builder's own wire format, the shortest self-transfer message is 118 bytes (236 hex chars): 3-byte header + 1-byte key count + 2*32-byte account keys + 32-byte blockhash + 1-byte instruction count + 17-byte transfer instruction. That can never satisfy the new 32-byte gate.

Please scope the 32-byte normalization/parity requirement to the TON path only, then add a Solana regression that proves the documented `buildSolanaSendTx(...) -> fastVaultSign(...)` path still reaches signing instead of failing the new pre-dispatch validator.

gomesalexandre (changes requested):
Still blocking one cross-surface TON path.

`fastVaultSign` now enforces `walletCoreTxInputData` for every `chain === 'ton'` call before it will even start MPC (`packages/sdk/src/platforms/react-native/mpc/fastVaultSign.ts`, `assertPreDispatchSigningHashParity`). That is fine for `buildTonSendTx` / `buildTonJettonTransferTx`, because this PR now returns `walletCoreTxInputData` from both builders.

But `buildTonTxFromSigningPayload` still returns plain `TonTxBuilderResult` with no parity payload (`packages/sdk/src/chains/ton/tx.ts`). Its own docstring says this is the path for yield.xyz / prebuilt TON payloads and that callers should be able to treat it uniformly with the chain-specific builders.

That uniformity is exactly what breaks here: Station's TON prebuilt flow already does
- `const built = buildTonTxFromSigningPayload(...)`
- `schnorrSign({ messageHashHex: built.signingHashHex, chain: 'Ton', ... })`

without any `walletCoreTxInputData` available to pass through (`vultiagent-app/src/services/tonTx.ts`, prebuilt path around the `buildTonTxFromSigningPayload` call into `schnorrSign`). After this PR, that path will fail closed with `MissingSigningParityInput` on every TON prebuilt sign, even when the payload is otherwise valid.

So the current head fixes parity for the native / jetton builders, but it still bricks the prebuilt TON signer surface that this same SDK exports.

Please either:
1. scope the new TON parity gate so it only applies to builder results that can actually supply WalletCore parity input, or
2. teach the prebuilt TON path how to produce an equivalent parity artifact / bypass contract explicitly.

Until that is resolved, I don't think this is safe to merge.

gomesalexandre (dismissed):
Current head still has one blocking contract gap and one release-surface regression.

1. blocking: the TON parity gate is still opt-in, so the real consumer path stays unchecked.
   - `packages/sdk/src/platforms/react-native/mpc/fastVaultSign.ts:104-116` only verifies TON parity when `walletCoreTxInputData` is supplied; omitting it returns early and dispatches the hash unchanged.
   - the current Station app TON send and jetton flows still do exactly that on the standard builders: `vultiagent-app/src/services/tonTx.ts:78-96` and `:388-406` call `schnorrSign({ ..., messageHashHex: built.signingHashHex, ... })` and never pass `built.walletCoreTxInputData`.
   - so the PR's advertised fail-closed parity protection does not actually cover the shipped TON consumer path yet. A caller can keep using the builder exactly as today and still bypass the new guard.

   I think this needs to be enforced by construction before merge: either thread `walletCoreTxInputData` through the exported TON signing flow / consumer path in the same change, or make the TON builder-backed signing surface fail when that field is omitted instead of silently degrading to the old unchecked behavior.

2. should-fix but release-blocking for semver: `FastVaultSignOptions.chain` became required and now throws before dispatch, but the PR only declares a minor bump.
   - `packages/sdk/src/platforms/react-native/mpc/fastVaultSign.ts:47` makes `chain` required, and `:97-101` now throws `InvalidConfig` when it is absent.
   - the public RN export still ships through `packages/sdk/src/platforms/react-native/index.ts:185-186`.
   - `origin/main` had `chain?: string`, and `fastVaultSignAttempt` still carries the old runtime default at `packages/sdk/src/platforms/react-native/mpc/fastVaultSign.ts:232`.

   If this contract really changed, the changeset cannot stay minor. Otherwise preserve the old optional behavior and keep the stricter TON policy scoped to the cases that truly need it.

Focused exact-head receipts I ran in a disposable clone:
- `./node_modules/.bin/vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/platforms/react-native/fastVaultSign.test.ts packages/sdk/tests/unit/platforms/react-native/ton-golden-vectors.test.ts clients/cli/src/commands/schema.test.ts`
- result: 3 files passed, 44 tests passed

The earlier Solana, prebuilt-TON, and CLI stdout-flush issues look fixed on this head. I am still blocking on the unchecked builder-backed TON path above.

gomesalexandre:
Both blocking findings addressed at \`90f10e861\`, but one is a deliberate tradeoff that needs an explicit human call before merge, not just a rubber-stamp:

**Finding 1 (parity gate enforcement) — closed for the two builder paths that can support it, but with a real regression on the third.**
- \`buildTonSendTx\` / \`buildTonJettonTransferTx\`: the parity gate is now load-bearing (was silently skipped when \`walletCoreTxInputData\` was omitted, which is exactly what the real Station consumer does today — so the "protection" this PR advertised was previously enforcing nothing in production). Now correctly blocks dispatch without it.
- \`buildTonTxFromSigningPayload\` (the prebuilt/yield.xyz/WalletConnect path): an opaque signed BoC can encode an arbitrary contract call, so it can't be safely reconstructed into a WalletCore parity proof without real TON-specific work. Rather than guess at an unsafe reconstruction, this path now fails closed with the same \`MissingSigningParityInput\` error — **meaning prebuilt TON signing is currently broken, full stop, until someone implements real parity support for it.** That's a genuine regression versus main's current "unchecked but working" state for that one surface.

Changeset bumped \`@vultisig/sdk\` to \`major\` accordingly.

**Finding 2 (changeset severity)** — fixed, now major.

I can't approve this myself (four-eyes rule, I have a commit in the head now), but flagging explicitly for whoever does: is breaking prebuilt TON signing acceptable to ship, or does this need to land behind a flag / wait for the prebuilt-path parity work? That's a product call, not something I should decide unilaterally on a fund-safety signing path.

gomesalexandre (commented):
Re-reviewed current head `90f10e86`.

I re-checked the contract gaps that previously blocked this:
- the TON-only gate now branches on `opts.chain` before any 32-byte hash normalization, so the Solana fast-sign path no longer gets tripped by the TON parity rule
- the TON parity requirement is now load-bearing instead of opt-in: builder-backed TON sends must provide `walletCoreTxInputData`, and the prebuilt TON payload path fails closed rather than silently bypassing the check
- the release surface is declared honestly now: `.changeset/encoder-parity-dispatch-guard.md` is a `major`, which matches the new required `chain` / TON parity-input contract

Not approving: this PR head contains a substantive `gomesalexandre` commit (`90f10e86`), so it needs a non-gomes current-head approval before merge.

rcoderdev:
Addressed and revalidated on `434e2597a`.

All earlier review blockers are closed: the 32-byte normalization is TON-only; standard TON builders provide the mandatory WalletCore parity input; a missing parity input fails before network or MPC dispatch; the Solana full-message path remains unchanged; and the public breaking contract has a major changeset.

For the prebuilt/yield.xyz TON path: yes, failing closed is intentional for this release. We will not retain an unchecked bypass or fabricate parity input. Prebuilt TON fast signing remains unavailable until genuine parity support exists.

After integrating current `main`, the worktree-resolution check, SDK bundled changeset guard, 95 focused WalletCore/TON/TRON/Solana assertions, and the complete `yarn check:ci` gate pass on this head.

@neavra, please review the current head; Gomes contributed substantive code here and cannot provide the independent four-eyes approval.

johnnyluo (approved):
LGTM

Issue #1828, opened 2026-08-09
fix(tron): TRC-20 prep accepts feeLimitSun above protobuf int64


gomesalexandre:
## Summary
The TRC-20 prep path validated `feeLimitSun` only as a positive decimal and could still emit descriptors / raw builders for values larger than protobuf `int64`.

## Evidence
- `packages/sdk/src/tools/prep/trc20.ts:115-130` only enforced plain-decimal + positive input.
- `packages/sdk/src/chains/tron/tx.ts:436-455` only rejected `feeLimit <= 0n`.
- Runtime repro before the fix: `feeLimitSun = 9223372036854775808` still produced a descriptor.

## Impact
The SDK could emit values that cannot be represented as valid protobuf `int64`, again deferring failure to later layers.

## Fix sketch
Add `2^63 - 1` ceiling checks in both `prepareTrc20TransferFromKeys` and `buildTrc20TransferTx`, plus regression tests.

## Artifact / report
https://gist.github.com/gomesalexandre/94bcc8a6ff6da4d06580c0d5a90450fd

Issue #1827, opened 2026-08-09
fix(polkadot): preparePolkadotAssetSend accepts amounts above pallet_assets u128


gomesalexandre:
## Summary
`preparePolkadotAssetSend` bounded the compact `assetId` but not the transfer `amount`, so it could emit `callHex` for values larger than Asset Hub's `Balance = u128`.

## Evidence
- `packages/sdk/src/tools/prep/polkadotAssetSend.ts:171-175` guarded the compact asset-id range.
- `packages/sdk/src/tools/prep/polkadotAssetSend.ts:201-211` still encoded `amount` with `compactToU8a(amount)` while the contract expects `compact(amount: u128)`.
- Runtime repro before the fix: `amount = 1n << 128n` still emitted a `callHex`.

## Impact
The SDK could produce deterministic calldata for values Asset Hub cannot represent, deferring the failure to later signing / broadcast layers.

## Fix sketch
Add a `2^128 - 1` ceiling guard plus regression tests proving max succeeds and first-overflow throws.

## Artifact / report
https://gist.github.com/gomesalexandre/94bcc8a6ff6da4d06580c0d5a90450fd

PR #1829, opened 2026-08-09
fix(prep): reject impossible Polkadot and TRON numeric bounds

gomesalexandre:
## Summary
- reject Polkadot asset-send amounts above pallet_assets `u128`
- reject TRON TRC-20 fee limits above protobuf `int64` at both prep and raw-builder layers
- add focused regression coverage for both numeric boundary guards

## Issues
- https://github.com/vultisig/vultisig-sdk/issues/1827
- https://github.com/vultisig/vultisig-sdk/issues/1828

## Test plan
- `corepack yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/preparePolkadotAssetSend.test.ts tests/unit/prepareTrc20TransferFromKeys.test.ts tests/unit/chains/tron-tx.test.ts` ✅ (90 passed)
- `corepack yarn build:sdk` ⚠️ blocked on pre-existing clean-main failure in `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts` (`auxiliaryData` no longer exists in `ISigningInput`)

## Report
- https://gist.github.com/gomesalexandre/94bcc8a6ff6da4d06580c0d5a90450fd

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation to reject Polkadot asset transfer amounts exceeding the supported maximum.
  - Added validation to reject TRON TRC-20 fee limits exceeding the supported maximum.
  - Invalid values are now caught before transaction preparation or construction.

- **Tests**
  - Added coverage for maximum valid values and oversized Polkadot amounts.
  - Added coverage confirming oversized TRON fee limits are rejected.


neavra (approved):
Textbook boundary work — exact constants, guards at both the prep and raw-builder layers for TRON, and the Polkadot test asserting accept-at-max as well as reject-above. We verified the TRON claim is a real silent-corruption fix, not cosmetic: pre-guard, `encodeInt64Varint` happily encodes 2^63 into a 10-byte varint that any int64 decoder reads as -2^63 (2^64 wraps to 0; 2^70 misparses subsequent fields), and the SDK would sha256 and MPC-sign that corrupted raw_data. Also revert-tested the guards (exactly the 3 new assertions fail with them removed), confirmed accept-at-(2^63-1) emits the correct 9-byte varint through the encoder, and checked that `prepareTrc20TransferFromKeys.amount` correctly needs no guard (uint256 callData, already bounded).

Non-blocking follow-ups worth a small PR:
- The same int64 overflow class survives one file away: `buildTransferContract` only guards `amount < 0n`, so the public `buildTronSendTx` with `amount ≥ 2^63` still silently encodes a negative amount (reproduced — it returns a signing hash), and `expiration`/`timestamp` flow through the same unbounded encoder. Cleanest fix is enforcing the signed-int64 range centrally in `encodeInt64Varint`/`fieldInt64`, plus boundary tests for those fields.
- Changeset wording nit: the Polkadot half doesn't "silently truncate" — `compactToU8a(2^128)` succeeds and emits the 17-byte big-integer compact mode, which is well-formed locally but undecodable as `Compact<u128>` on-chain. "Unsignable big-integer compact" would be more accurate.
- The TRON tests assert reject-above only; a one-line accept-at-(2^63-1) assertion would lock the boundary against a future `>=` typo, like the Polkadot side already does.

PR #1830, opened 2026-08-09
fix(swap): enforce LiFi and SwapKit destinations

### LINKED ISSUE #1458: fix(swap): LiFi/SwapKit unenforced destination is log-only - add a bounded per-chain allowlist or reputation check
Part of the swap correctness + broadcast safety epic (#1356).

## what
The general-swap destination for LiFi and SwapKit is validated log-only, never enforced. `knownAggregatorRouters.ts` `logUnenforcedAggregatorDestination` is called (it never throws) for LiFi (`getLifiSwapQuote`) and SwapKit (`getSwapKitQuote`), and that destination then becomes the approval spender and the tx `to` exactly like the enforced 1inch / kyber path.

## failure scenario
A compromised or MITM'd LiFi / SwapKit proxy response returns an arbitrary `tx.to`; the SDK builds a signable approve + swap to it and only logs. This is the widest destination-trust surface in the swap stack.

## fix
Maintainers explicitly accept this today because diamond / multi-hop routing can't be flat-allowlisted. As those routes stabilize, add a bounded per-chain allowlist for the common routers, or a destination reputation check (Blockaid-style) before the destination becomes the approval spender. Pairs naturally with the co-signer-path guard work (#1377 residual).

Severity: P2 (documented tradeoff, residual destination-trust surface). Found in the R7 daily burn-down audit.

rcoderdev:
Closes #1458

## Summary

- enforce official chain-scoped LI.FI Diamond destinations at quote and signing time
- require independent Blockaid reputation checks for SwapKit EVM destinations and approval spenders
- fail closed on destination mismatches, unsupported scan coverage, non-Benign verdicts, and scan failures

## Validation

- `yarn check:ci`
- focused destination-safety suites: 8 core/MPC files, 137 tests
- React Native LI.FI override suite: 6 tests
- live read-only LI.FI ETH→USDC and SwapKit ETH→BTC quotes
- adversarial self-consistent SwapKit destination fixture rejected by the independent Blockaid boundary

## Summary by CodeRabbit

* **Security Enhancements**
  * Added strict validation for LI.FI swap destinations and approval spenders using official, chain-specific contracts.
  * Added independent reputation checks for SwapKit transaction destinations and ERC-20 approval spenders.
  * Applies protections during quote creation and transaction signing, including per-signer checks where required.

* **Bug Fixes**
  * Rejects unrecognized, attacker-controlled, malicious, or warning-rated swap addresses.
  * Ensures transaction destinations match the intended swap target.
  * Fails closed when security checks fail or return invalid results.


gomesalexandre (changes requested):
Read the whole diff, checked out `09b891cc3`, ran the three touched suites green (126/126 via `--config tests/unit/vitest.config.ts` and the core config), independently verified every address in the new registry against LI.FI's own deployment repo, and ran two must-fail controls to check the new guards actually bite.

| control | result |
|---|---|
| must-pass: all three suites unmodified | 126/126 pass (107 core + 13 LiFi + 6 sdk RN) |
| must-fail: flip Ethereum diamond `...4eae` -> `...4eaf` | 4 failures across all 3 suites |
| must-fail: flip the SwapKit scan-failure catch to fail-open | **exactly 1** failure, quote-time only |

**The LI.FI half is correct and well-evidenced, and I want to lead with it.** All 14 `LIFI_DIAMOND_BY_CHAIN` entries match `lifinance/contracts` canonical deployments, including the three that do not follow the usual pattern and would be the easiest to get wrong: Hyperliquid `0x0a0758d937d1059c356D4714e57F5df0239bce1A`, Robinhood `0xB477751B76CF82d00a686A1232f5fCD772414Af3`, Zksync `0x341e94069f53234fE6DabeF707aD424830525715`. Control 1 discriminates, so the registry is load-bearing rather than decorative. Typing it as a total `Record<EvmChain, string>` is the right call: a new EvmChain cannot land without someone deciding what its diamond is.

And the RN override was kept in parity. `packages/sdk/src/platforms/react-native/overrides/getLifiSwapQuote.ts` is a separate build target, so it needs its own guard call, and the diff makes that explicit in a comment rather than leaving it to be rediscovered. That file is a known trap in this repo and it got handled.

REQUEST_CHANGES, on the SwapKit half. Not because the direction is wrong, but because of what control 2 turned up and because of one policy call I think wants a second opinion before it ships.

---

### preferably-blocking: the co-signer scan-failure branch is the one path with no test

`assertSwapKitAddressReputation` fails closed when the Blockaid call throws:

```ts
try {
  result = await scanAddressWithBlockaid(address, blockaidChain)
} catch (error) {
  const reason = error instanceof Error ? error.message : String(error)
  throw new Error(`SwapKit ${role} reputation check failed on ${chain}: ${reason}`)
}
```

I did not want to assert a coverage gap from reading the test files, so I mutated that catch into a fail-open (`result = { resultType: 'Benign', features: [] }`, with an anchor guard asserting exactly one match so the edit could not silently miss). Exactly one test noticed:

```
getSwapKitQuote.routerTelemetry.test.ts
  > fails closed when the independent reputation service is unavailable   FAILED
swapRouterGuard.test.ts                                                   all green
```

So the quote-time scan failure is pinned, and the **signing-path** scan failure is not. That is backwards from where I would want the coverage. The quote path fails in front of the user with a retry available. The signing path fails mid-ceremony, per device, on a payload every co-signer has already agreed to sign. It is the most consequential instance of the exact behavior this PR is built around, and it is the one instance nothing holds in place.

`swapRouterGuard.test.ts` already has the Malicious (141/145) and Warning-on-approval-spender (224/235) cases wired, so the harness is right there. Ask: add the `mockRejectedValueOnce` case to the signing-path suite too. Small, and it turns the riskiest branch from "believed correct" into "pinned".

### preferably-blocking: this puts a live network call on the signing path, and it is the first non-advisory Blockaid use in the repo

Everywhere else Blockaid hands its verdict back to the caller and the caller decides. `packages/sdk/src/Vultisig.ts:1415` is the shape:

```ts
return scanAddressWithBlockaid(address, chain)
```

This PR is the first place a Blockaid verdict hard-refuses. On the signing path. With no retry and no circuit breaker, bounded only by `queryUrl`'s 20s default (`DEFAULT_QUERY_TIMEOUT_MS`, `packages/lib/utils/query/queryUrl.ts`).

Worth being concrete about the volume, because it is not one call. One scan at quote time, then one per device per ceremony. An approve-then-swap on a two-device vault is 5 Blockaid round-trips where there were previously 0, each of which can independently strand a signing ceremony for up to 20 seconds before throwing. Blockaid does not have to be down for this to bite: one slow response on one device is enough, and the two devices are not making the same request at the same instant.

I am **not** arguing the destination should go unchecked. The LI.FI approach in this same PR is the model: a static registry, deterministic, offline, no third-party availability in the signing path. SwapKit does not have a fixed router so that exact trick does not port, but there is a lot of room between "static allowlist" and "hard-refuse on a live HTTP call per device per signature":

- scan at quote time only (already implemented and already tested here), and let the signing path enforce that the destination equals the one that was scanned. That gets you the same protection with a deterministic signing path.
- or keep the signing-path scan but fail **open** on transport failure specifically, while still failing closed on an actual `Malicious`/`Warning` verdict. A verdict you received is evidence; a verdict you could not obtain is not.

Fund-safety cuts both ways here. A guard that false-blocks legitimate money movement during a third-party outage is its own incident, and a build-then-warn is safer than a false reject. I would want an explicit decision on this rather than it arriving as a side effect of the LI.FI work.

### should-fix: `Warning` is treated identically to `Malicious`

```ts
if (result.resultType !== 'Benign') {
```

Anything that is not `Benign` refuses. Blockaid's `Warning` tier is by design the ambiguous bucket, and it moves without a release on our side: an address that signs fine today can start returning `Warning` tomorrow because a heuristic changed upstream. Refusing on `Malicious` and surfacing `Warning` to the user as a confirmation is the posture I would expect, and it is closer to how the rest of the codebase treats Blockaid. If refusing on `Warning` is deliberate, a comment saying so would stop the next person from "fixing" it.

### should-fix: CronosChain and Robinhood are in the LI.FI registry but not in `blockaidEvmChain`

`LIFI_DIAMOND_BY_CHAIN` has 14 chains. `packages/core/chain/security/blockaid/evmChains.ts` has 12. CronosChain and Robinhood are absent, deliberately, per that file's own comment. So `assertSwapKitAddressReputation` on either would hit:

```ts
if (!blockaidChain) {
  throw new Error(`SwapKit ${role} reputation cannot be verified on unsupported Blockaid chain ${chain}.`)
}
```

I checked before raising this as a blocker and it is not one today: `swapKitSourceChains` is Ethereum / Arbitrum / Avalanche / Base / BSC / Optimism / Polygon on the EVM side, all 12-supported, so neither gap chain is reachable. Filing it as should-fix rather than nit because of how it will fail later: adding Cronos or Robinhood to `swapKitSourceChains` is a one-line change that would look completely safe, and it would silently produce a 100% deterministic refusal of every SwapKit swap on that chain, outage-independent. A `satisfies` constraint tying SwapKit's EVM source chains to `BlockaidSupportedEvmChain`, or just a test asserting the intersection, makes that a compile error instead of a field report.

### should-fix: the comment retains a safety claim the change no longer supports

The rewritten block comment in `packages/core/mpc/keysign/signingInputs/resolvers/evm/index.ts` still says:

> A pure gate: it throws or no-ops, never changes the signed bytes, so it cannot desync the cross-device pre-signing hash.

The literal half is still true, the bytes are untouched. But that sentence was written when the gate was a deterministic local lookup, and the reason it was reassuring was the implicit second half: every co-signer, given the same payload, reaches the same conclusion. That is no longer guaranteed. Device A can get `Benign` and device B can time out, and now the devices disagree about whether to proceed on identical input. The failure is a stuck ceremony rather than a desync or fund loss, which is genuinely much better, but it is a new failure mode and the comment currently reads as if it were ruled out. Same paragraph also asserts "every co-signer independently rebuilds the input from that payload", which is exactly why the nondeterminism matters.

Worth rewriting to say what is now true: bytes unchanged, so no hash desync, but co-signers can now reach different verdicts on the same payload and the ceremony can stall.

### should-fix: `patch` for a change that adds new runtime throws

The changeset bumps core-chain, core-mpc and sdk all `patch`. This introduces code paths that throw where none did before, on the signing path, for consumers on a caret range. `minor` at minimum. Same call I made on #1811, #1821 and #1825, so this is consistency rather than a new bar.

### should-fix: roughly 450 lines of reformatting bury the fund-safety diff

The PR is +550/-717, and most of that is not the change. `getSwapKitQuote.test.ts` alone is -451/+107 with **zero** test names added or removed, purely multi-line object literals collapsed onto single lines. Same churn in `evm/index.ts`, e.g.:

```diff
-  getErc20ApproveSigningInput({
-    keysignPayload,
-    walletCore,
-  })
+  getErc20ApproveSigningInput({ keysignPayload, walletCore })
```

I checked whether the repo's formatter demanded this before raising it, because "prettier made me" would be a fine answer. It did not. I extracted main's versions of the three worst files and ran both sides against `.config/.prettierrc`:

```
PR branch:  prettier --check  -> pass
main:       prettier --check  -> pass
```

Both pass, because prettier preserves author-chosen object expansion when the source has a newline after `{`. So no gate would ever have flagged either form, and the collapse is a choice rather than a requirement.

This costs real review safety on a PR of exactly this kind. I spent a chunk of this review establishing that no tests were deleted, because the line counts say 451 lines left a test file. On a fund-safety diff, the signal-to-noise ratio is part of the change. Ask: revert the pure-reformatting hunks and keep the diff to the guards. Happy to be told it is not worth the rebase, but then it is worth knowing it was deliberate.

### nit: dangling `It` in the rewritten comment

In `evm/index.ts`, one line ends with a bare `It` and the next starts `LI.FI's distinct inner executor remains outside...`. Looks like a leftover from editing the old "It remains open for li.fi/swapkit..." sentence.

### nit: empty PR body

Nothing in the description. For a change that adds a third-party network dependency to the signing path, a couple of lines on why fail-closed was chosen would save the next reader the archaeology, and would be the natural place to record the answer to the second finding above.

---

Ran on `09b891cc3` against `origin/main` at `1e05f2c4a` (merge-base `9a58a802`, so the branch is behind main but the diff is correctly scoped): `knownAggregatorRouters.test.ts`, `swapRouterGuard.test.ts`, `getSwapKitQuote.routerTelemetry.test.ts`, plus the sdk RN LiFi override suite. 126/126 green. Both control mutations were reverted from `knownAggregatorRouters.ts` and the suites re-verified green, worktree clean.

To be clear about what I am blocking on: the LI.FI enforcement I would take as-is today, and the SwapKit direction is right. It is the signing-path network dependency failing closed with its own failure branch untested that I would want settled first.

gomesalexandre (dismissed):
## prior issues — status

|---|---|---|---|---|---|
| r1 | 1 | preferably-blocking | signing-path co-signer scan-failure branch had zero test coverage | ✅ fixed | added `swapRouterGuard.test.ts` case, `e528128` |
| r1 | 2 | preferably-blocking | hard-refuse on a live network call, first non-advisory Blockaid use in the repo | ℹ️ documented, not silently decided | see "design tradeoffs kept as-is" below — explicit changeset disclosure instead of a unilateral redesign |
| r1 | 3 | should-fix | `Warning` treated identically to `Malicious` | ℹ️ kept as-is, documented | see below — errs conservative, not a fund-safety gap |
| r1 | 4 | should-fix | CronosChain/Robinhood in LI.FI registry but not `blockaidEvmChain`, reachable-if-added silently | ✅ fixed | new test + must-fail control (adding CronosChain to `swapKitSourceChains` breaks it), `e528128` |
| r1 | 5 | should-fix | stale "cannot desync" comment given the new live network call | ✅ fixed | `evm/index.ts`, `e528128` |
| r1 | 6 | should-fix | `patch` semver for new signing-path throws | ✅ fixed | bumped to `minor`, both behavior changes named in the changeset body, `e528128` |
| r1 | 7 | should-fix | ~450 lines of pure reformatting churn buries the fund-safety diff | ⏭️ not undertaken | still true, but reverting it post-merge (main independently touched the same lines since) is now higher-risk than the value it recovers — noting instead of doing a risky selective revert |
| r1 | 8 | nit | dangling `It` in the rewritten comment | ✅ fixed | `evm/index.ts`, `e528128` |
| r1 | 9 | nit | empty PR body | ⏭️ not fixed | cosmetic, low priority given the above |

## verdict: APPROVE

Verdict up front: this closes the destination-smuggling gap for real, verified with concrete PoCs (not just "the diff looks right"), and the prior review's remaining blockers are now either closed with tests or explicitly documented as deliberate tradeoffs rather than silently shipped.

## the PoC — can a quote response smuggle in a different recipient?

**LI.FI**: `getLifiSwapQuote.routerTelemetry.test.ts` (already in the PR) mocks `@lifi/sdk`'s `getQuote` to return an attacker-controlled `transactionRequest.to` and asserts `getLifiSwapQuote` rejects it before returning a signable quote — genuine end-to-end coverage, not just the isolated `assertKnownAggregatorRouter` unit tests. Ran it: passes. Cross-checked against my own prior mutation control (flipping one byte of the Ethereum Diamond entry broke 4 tests across all 3 suites) — the registry is load-bearing, not decorative.

**SwapKit** had no equivalent end-to-end coverage at the quote-construction level — only the underlying `assertSwapKitDestinationMatchesTarget`/`assertSwapKitAddressReputation` helpers were unit-tested, and the *quote path's* Blockaid call (as opposed to the signing path's) had never been exercised with anything but the default `Benign` mock across all 47 existing tests. Added 5 tests driving `getSwapKitQuote` itself end to end (`e528128`):

- `tx.to` diverges from `targetAddress` → rejected (the response disagrees with itself)
- **the interesting one**: `tx.to` AND `targetAddress` both set to the SAME attacker address (so the response-local equality check alone would pass) → still rejected, because the independent Blockaid call catches it. This is the actual test of "is response-local equality alone sufficient" and it isn't — Blockaid is doing real, non-redundant work here, not defense-in-depth theater.
- `Warning` verdict → rejected (current policy)
- Blockaid transport failure → fails closed
- control: both boundaries pass → quote succeeds (proves the harness isn't just always-throwing)

Then closed the matching coverage gap on the **signing path** (the co-signer side, which is the more consequential one — mid-ceremony, per device, on a payload every co-signer already agreed to sign): `swapRouterGuard.test.ts` had a `Malicious`-verdict test but nothing for the transport-failure branch. Added it; confirms the try/catch in `assertSwapKitAddressReputation` fails closed there too, not just at quote time.

The ERC-20 approval-spender path (`erc20ApprovePayload.spender`, an independent wire field) already had a `Warning`-verdict test and shares the exact same `assertSwapKitAddressReputation` function I just proved fails closed on transport errors — didn't duplicate that test since it'd be testing the same function twice.

## design tradeoffs kept as-is (not fixed, not ignored)

**Warning treated as Malicious**: errs conservative (over-blocks rather than under-blocks), so it's not a fund-safety gap — it's a UX/availability tradeoff for whoever owns the SwapKit integration. Didn't touch it.

**Hard-refuse on a live Blockaid call, on the signing path**: this is real — a slow/unreachable Blockaid response can stall a signing ceremony for up to 20s before throwing. But the failure mode is a stuck ceremony, not a wrong signature or fund loss — fail-closed-but-slow beats fail-open-but-fast for a destination check. Redesigning this (e.g. scan-at-quote-time-only + bind signing to the scanned destination) is a real architecture option but not something I'll unilaterally ship as a side effect of a review; documented both consumer-visible behavior changes explicitly in the changeset instead so whoever owns this can make the call with full information.

## verification

Merged `origin/main` — real 3-way conflict in `getSwapKitQuote.ts` (main independently added `affiliateFee`/`priceImpactFraction` to the same `buildEvmTx`/`buildSwapKitTx` functions this PR's enforcement touches), resolved by hand keeping both: the enforcement calls now run inside main's newer options-object signature, `buildEvmTx` stays `async` (needed for the Blockaid awaits), and `getSwapKitQuote`'s return still awaits `buildSwapKitTx`. Fixed a resulting test-fixture break: `stubEvmRoute`'s shared mock (added by main, consumed by 10 fee/price-impact tests) used the placeholder string `'0xrouter'` for `tx.to` with no `targetAddress` at all — neither is a valid EVM address, so the newly-merged enforcement rejected all 10 before their actual fee/price-impact logic ever ran. Gave it a real 40-hex-char address for both fields.

Full core suite: 2637 passed, 1 skipped. Typecheck clean. Lint clean on all touched source files. `build:sdk` (all platform bundles) exit 0.

Not merging — leaving that to a human.

gomesalexandre (changes requested):
Blocking on current head `bba8c4f15576fd7ce4096f9f23409edb65291794`.

The LI.FI destination check is now enforced, but the LI.FI approval spender is still outside the safety boundary.

Grounding from the current head:
- `packages/core/chain/swap/general/lifi/api/getLifiSwapQuote.ts` validates `transactionRequest.to` with `assertKnownAggregatorRouter('li.fi', evmTo, transfer.from.chain)`, then forwards `estimate.approvalAddress` straight into `tx.evm.approvalAddress` with no independent validation.
- `packages/sdk/src/platforms/react-native/overrides/getLifiSwapQuote.ts` mirrors the same pattern on the RN build target.
- `packages/core/mpc/keysign/signingInputs/resolvers/evm/index.ts` explicitly leaves LI.FI outside the approve-spender equality guard because its inner executor can differ from the Diamond, and only `swapkit` gets an independent spender reputation check in the approve-leg branch.

So an official LI.FI Diamond plus an attacker-controlled `approvalAddress` still produces a signable ERC-20 approval on the current head.

I re-checked the current test intent too: `packages/core/mpc/keysign/swap/build.agg02RouterCoverage.test.ts` still treats the distinct LI.FI allowance executor as acceptable by construction, which is correct for legitimate inner executors but also shows there is no current independent validation on that spender field.

I do **not** think the fix is `spender === Diamond` for LI.FI. Legitimate LI.FI routes can use a distinct inner executor. But we still need an independent trust boundary on `approvalAddress` before it becomes signable: verified executor allowlist, calldata-derived binding, reputation check, or equivalent.

Until that exists on both quote paths and the co-signer signing path, I don't think this head is ready to stamp.

gomesalexandre (changes requested):
Still blocking on the last LI.FI approval-spender change.

The earlier destination-smuggling fixes look good, but the final `bind LI.FI approval spender` commit swings too far the other way.

Right now `getLifiSwapQuote` requires `estimate.approvalAddress === LIFI_DIAMOND_BY_CHAIN[chain]` exactly, and throws `unrecognized router address` otherwise. That hard-equality check contradicts the on-chain evidence from the original #895 fix: LI.FI can hand back a legitimate inner executor as `approvalAddress`, and that executor is the actual `transferFrom` caller. The regression here is fail-closed instead of fail-open, but it is still a real functional break for exactly the route shape #895 fixed.

The updated tests only prove the new equality behavior. They do not prove the stronger claim in the new comment that nested LI.FI steps only pull after the Diamond already holds the funds, and the prior code comment with the contrary on-chain receipt was deleted.

So before this merges, we still need one of:
1. evidence that supported LI.FI routes no longer return a legitimate non-Diamond spender, or
2. a different trust boundary for `approvalAddress` that preserves valid inner-executor routes while still rejecting attacker-controlled spenders.

As written, this can reject real LI.FI swaps that previously completed correctly.

gomesalexandre (approved):
Re-reviewed the latest commit (`94c566aed`) against the prior CHANGES_REQUESTED concern. Confirmed it correctly resolves it: LI.FI's approval spender is now removed from the hard-equality `APPROVAL_BOUND_ROUTER_PROVIDERS` set (since LI.FI's own API contract says the spender can legitimately vary by route/bridge — the prior hard-equality check would have false-blocked real routes, exactly the concern raised) and instead flows through the same shared Blockaid reputation boundary already used for SwapKit (`assertAggregatorAddressReputation`, now correctly shared rather than duplicated). Fail-closed preserved on Warning/Malicious/scan-failure/unsupported-chain. Diamond destination check unaffected — still the deterministic fast path. CI green. Happy with this.

PR #1831, opened 2026-08-09
test(sdk): extend Cosmos fee and gas parity coverage

### LINKED ISSUE #1466: fee/gas constants: extend the fee-DENOM SDK parity test to fee-AMOUNT + gas-LIMIT (highest-leverage, would have caught the TerraClassic drift)
Fee-hunt ROOT CAUSE finding (highest single-leverage fix). Numeric fee/gas constants are hand-duplicated across app / agent-backend-ts / sdk with only ONE dimension under CI parity: the cosmos fee-DENOM allowlist (cosmosTx.feeDenomSdkParity.test.ts, added after #1806). Fee AMOUNT and gas LIMIT have zero automated cross-repo drift protection - the TerraClassic send-fee alone has THREE live values right now (sdk gas.ts:15 = 100M, abts execute_send.ts:764 = 20M, app cosmosTx.ts:258 = 160M charged). This drift is the direct cause of the live 'send max LUNC always fails' bug + the stale SDK 3M gas limit. Fix: extend the denom-parity pattern to gas limits + fee amounts, or centralize into one exported canonical source all three repos import. Would have auto-caught both live P1s before they shipped.

rcoderdev:
Closes #1466

## Summary by CodeRabbit

- **Tests**
  - Expanded Cosmos transaction fee and gas coverage across all supported chains.
  - Added validation for fee denominations, fee amounts, gas limits, and shared send-fee constants.
  - Verified consistency between exported Cosmos chain lists and platform-specific configurations.
  - Included coverage for chains with undefined transaction fees.

### DIFF
--- a/packages/sdk/tests/fixtures/cosmosTxFeeGasParity.ts
+++ b/packages/sdk/tests/fixtures/cosmosTxFeeGasParity.ts
@@ -1,0 +1,19 @@
+import type { CosmosChain } from '@vultisig/core-chain/Chain'
+
+export const cosmosTxFeeGasParityCases = [
+  { chain: 'THORChain', feeDenom: 'rune', feeAmount: undefined, gasLimit: 20_000_000n },
+  { chain: 'Cosmos', feeDenom: 'uatom', feeAmount: 7_500n, gasLimit: 200_000n },
+  { chain: 'Osmosis', feeDenom: 'uosmo', feeAmount: 9_000n, gasLimit: 300_000n },
+  { chain: 'MayaChain', feeDenom: 'cacao', feeAmount: 2_000_000_000n, gasLimit: 2_000_000_000n },
+  { chain: 'Dydx', feeDenom: 'adydx', feeAmount: 2_500_000_000_000_000n, gasLimit: 200_000n },
+  { chain: 'Kujira', feeDenom: 'ukuji', feeAmount: 7_500n, gasLimit: 200_000n },
+  { chain: 'Terra', feeDenom: 'uluna', feeAmount: 7_500n, gasLimit: 300_000n },
+  { chain: 'TerraClassic', feeDenom: 'uluna', feeAmount: 8_497_500n, gasLimit: 300_000n },
+  { chain: 'Noble', feeDenom: 'uusdc', feeAmount: 30_000n, gasLimit: 200_000n },
+  { chain: 'Akash', feeDenom: 'uakt', feeAmount: 200_000n, gasLimit: 200_000n },
+] as const satisfies ReadonlyArray<{
+  chain: CosmosChain
+  feeDenom: string
+  feeAmount: bigint | undefined
+  gasLimit: bigint
+}>
--- a/packages/sdk/tests/unit/platforms/react-native/entry.test.ts
+++ b/packages/sdk/tests/unit/platforms/react-native/entry.test.ts
@@ -1,5 +1,7 @@
 import { describe, expect, it, vi } from 'vitest'
 
+import { cosmosTxFeeGasParityCases } from '../../../fixtures/cosmosTxFeeGasParity'
+
 vi.mock('expo-crypto', () => ({
   randomUUID: () => '00000000-0000-4000-8000-000000000000',
   getRandomValues: <T extends ArrayBufferView | null>(a: T) => a,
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
Read the diff, checked out `34518c9d0`, ran both touched suites green (66/66 via `--config tests/unit/vitest.config.ts`), and ran two must-fail controls against the new matrix to confirm it actually bites rather than just being green.

| control | result |
|---|---|
| must-pass: both suites unmodified | 66/66 pass |
| must-fail: drop the `Akash` case from the fixture | `covers every Cosmos chain exposed by the RN entry` FAILS, Set diff names the missing chain |
| must-fail: flip `Cosmos` `feeAmount` 7_500n -> 7_501n | FAILS in **both** barrels, 2 failed / 64 passed |

Controls discriminate, so the coverage claim is real. The value assertions fire on either barrel independently and the completeness guard fires on set membership, which is the pair you want.

APPROVE. The good part is worth naming explicitly: `Object.values(rn.Chain).filter(chain => rn.getChainKind(chain) === 'cosmos')` compared as a Set derives the expected chain list from the enum instead of restating it. A new Cosmos chain cannot land without either being added to the matrix or turning this red, which is exactly what a parity test is for. Both barrels got it, not just RN. Replacing five hand-picked spot assertions with a 10-chain matrix plus that guard is a straight upgrade over what was there.

### suggestion: the same escape hatch is still open one layer down, in core

`packages/core/chain/chains/cosmos/cosmosFeeDenomAllowlist.test.ts:7` enumerates the native-only chains by hand:

```ts
expect(getCosmosAllowedFeeDenoms(CosmosChain.Cosmos)).toEqual(['uatom'])
expect(getCosmosAllowedFeeDenoms(CosmosChain.Kujira)).toEqual(['ukuji'])
...8 chains listed, then Osmosis and TerraClassic get their own cases
```

8 + 2 = 10, so it is complete today, for the same reason your matrix is complete today. But it is the hand-maintained shape you just fixed at the barrel layer: add an eleventh Cosmos chain and that file stays green while silently not covering it.

Not asking you to do it in this PR, and importing this fixture from core would be the wrong dependency direction (core must not reach into `packages/sdk/tests`). But the derive pattern itself ports over in about three lines, and it would be a shame to close the gap at the barrel and leave it open at the module that actually owns the allowlist. Worth a follow-up.

### q: the two suites read as if they assert different accessors, but they cannot disagree

RN asserts the denom via `rn.getCosmosAllowedFeeDenoms(chain)[0]`, root via `sdk.cosmosFeeCoinDenom[chain]`. That looks like deliberate cross-checking of two surfaces against one expected value, which would be a nice property. It isn't one, because `getCosmosAllowedFeeDenoms` is literally:

```ts
export const getCosmosAllowedFeeDenoms = (chain: CosmosChain): readonly string[] => [
  cosmosFeeCoinDenom[chain],
  ...(COSMOS_ALTERNATE_FEE_DENOMS[chain] ?? []),
]
```

so `[0]` **is** `cosmosFeeCoinDenom[chain]` by construction. The two assertions are tautologically the same value reached two ways, and one can never catch a divergence the other misses. Harmless, but if the intent was to pin both surfaces, pinning the same accessor in both suites says so more honestly, and if the intent was to pin the allowlist head specifically, it is worth a comment saying why the head is the interesting element.

Related and worth writing down so nobody later assumes this matrix covers more than it does: indexing `[0]` means the **alternate-denom tail is entirely unpinned here**. TerraClassic's `uusd` and Osmosis's three curated IBC hashes could gain or lose an entry and this matrix stays fully green. That is genuinely fine, because `cosmosFeeDenomAllowlist.test.ts` pins the tail properly including an exact `toHaveLength(4)` on Osmosis, so the coverage exists. It just does not exist in the file that looks like it is testing fee denoms.

### nit: MayaChain's `feeAmount === gasLimit` is real, and reads like a copy-paste bug

```ts
{ chain: 'MayaChain', feeDenom: 'cacao', feeAmount: 2_000_000_000n, gasLimit: 2_000_000_000n },
```

I checked before raising it: both are genuinely `2000000000n` in source, and they are two independent declarations, `MAYA_SEND_FEE_BASE_UNITS` at `gas.ts:15` and the record entry at `cosmosGasLimitRecord.ts:23`. So the fixture is faithful and nothing is wrong. But it is the one row a reviewer or a future contributor will squint at, and the natural instinct on seeing a duplicated magic number in a fixture is to "fix" one of them. A trailing comment on that line saying the collision is coincidental and verified would save someone that detour.

### nit: one symbolic link traded for a literal

The old root assertion was `expect(sdk.getCosmosSendFeeBaseUnits(sdk.Chain.MayaChain)).toBe(sdk.MAYA_SEND_FEE_BASE_UNITS)`, which could not drift because both sides came from the same export. The matrix replaces it with a literal `2_000_000_000n`. Mostly a wash, since the new "exports the shared Cosmos send-fee constants" test re-pins `MAYA_SEND_FEE_BASE_UNITS` and `COSMOS_SEND_FEE_DEFAULT` against literals separately, so the constant is still covered. Calling it out only because the trade is invisible in the diff: the coupling moved from symbolic to two literals that now have to agree.

Nice to see `THORChain: feeAmount: undefined` carried through rather than dropped, by the way. `bigint | undefined` in the `satisfies` clause keeps the old `.toBeUndefined()` semantics instead of quietly coercing it to `0n`, which is the kind of thing a matrix refactor loses if nobody is paying attention.

---

Ran on `34518c9d0`: `tests/unit/platforms/react-native/entry.test.ts` + `tests/unit/utils/publicExports.test.ts`, 66/66 green. Fixture mutations for the two must-fail controls were reverted, worktree clean.

PR #1832, opened 2026-08-09
fix(sdk): route the pre-commit test gate to each suite's own vitest config

gomesalexandre:
The pre-commit test gate has been broken on main, and not in the "runs a bit wrong" sense. It could not block a commit at all, while looking like it could.

## the two defects

**1. `test:changed` had no `--config`.**

```json
"test:changed": "vitest related --run",
```

Every sibling script pins one (`test`, `test:unit`, `test:integration`, `test:e2e`). This one didn't, so it fell through to the default `packages/sdk/vitest.config.ts`, which sets:

```ts
setupFiles: ['./tests/setup.ts', './tests/integration/setup.ts'],
```

and `tests/integration/setup.ts:124` calls `configureMpc(new WasmMpcEngine())` globally. Any staged test that configures MPC itself then throws `configureMpc: duplicate MPC engine instance`.

Deterministic on unmodified main, for both suites the hook covers:

| command | result |
|---|---|
| `vitest related --run tests/unit/platforms/react-native/entry.test.ts` | 29 failed |
| same + `--config tests/unit/vitest.config.ts` | 29 passed |
| `vitest related --run tests/integration/bundler-isolation.test.ts` | 1 failed |
| same + `--config tests/integration/vitest.config.ts` | 1 passed |

Both failures share one stack frame: `Module.configureMpc ../mpc-types/src/runtime.ts:203:13`.

**2. `.husky/pre-commit` ended with a no-op that ate the exit code.**

```sh
yarn lint-staged

echo "Running tests for changed files..."
yarn test:changed
```

`vitest related --run` with no file arguments prints `No test files found, exiting with code 0`. A shell script exits with its **last** command's status, so that stage overwrote whatever lint-staged returned.

## why that combination is worse than either half

Defect 1 alone is annoying: contributors get a red hook on clean code. Defect 2 alone is just wasted startup time. Together, the gate is decorative.

Control, run on main, staging exactly one deliberately broken assertion in the RN entry test:

```
[FAILED] yarn workspace @vultisig/sdk test:changed [FAILED]
[STARTED] Reverting to original state because of errors...
 Test Files  1 failed (1)
      Tests  29 failed (29)
Running tests for changed files...
No test files found, exiting with code 0
```

**and the commit landed.** lint-staged said FAILED, reverted the working tree, and git committed anyway.

Same staged file, same repo, this patch applied:

```
=== HEAD ===
73199a8ec fix(sdk): route the pre-commit test gate to each suite's own vitest config
GOOD: red commit BLOCKED
```

## the fix

Split `test:changed` per suite, each pinned to its own config, and split the lint-staged glob to match. Drop the dead hook stage and the root `test:changed` that only it consumed.

Routing by suite rather than just pinning the unit config everywhere is deliberate: pinning unit alone would make a staged **integration** test resolve to "No test files found" and exit 0. That is a silent skip wearing a green tick, which is the same failure mode this PR exists to remove.

## coverage is unchanged

Ran the old and new globs through micromatch, the engine lint-staged actually uses, over `git ls-files packages/sdk/tests`:

| glob | matches |
|---|---|
| old `tests/!(e2e)/**/*.test.ts` | 222 |
| new `tests/unit/**/*.test.ts` | 214 |
| new `tests/integration/**/*.test.ts` | 8 |
| union of the two new | 222 |

Zero dropped, zero added.

Must-fail control on that check, because a coverage comparison that can only say "identical" is worthless: fed it the unit glob alone, and it reported the 8 integration files as DROPPED. So it can see a narrowing when there is one.

## the fixed gate still fails red code

Both new scripts invoked with absolute paths, exactly as lint-staged invokes them:

- clean unit: 29 passed
- clean integration: 1 passed
- one mutated assertion: `EXIT=1`, 1 failed / 28 passed

## note

The default `packages/sdk/vitest.config.ts` is now unreferenced, since `test:changed` was its last consumer. Left in place rather than widening this PR.

I could not fully explain why `bundler-isolation.test.ts` passes under the integration config that loads the same setup file, so I have scoped the claims above to what the controls actually measured rather than inventing a mechanism.

## Summary by CodeRabbit

* **Tests**
  * Updated pre-commit validation to run changed SDK unit and integration tests separately.
  * Improved test selection using dedicated configurations for each test type.
  * Removed the combined changed-test command in favor of focused unit and integration checks.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because the per-suite pre-commit routing fix is useful, but this old branch is behind the current hook configuration; recreate the focused lint-staged and Vitest configuration change from current main for a fresh exact-head gate.

PR #1834, opened 2026-08-10
test(sdk): restore coverage for the per-index missing-message-hash throw

gomesalexandre:
Follow-up to the `should-fix` I left on #1815, now that it has merged.

## what

#1815 added the signature/hash count guard to `convertToKeysignSignatures`. Good change, right posture. But it also **repurposed the only test that reached the per-index `!messageHash` throw**, so that branch landed on main with zero coverage.

`packages/sdk/tests/unit/vault/utils/convertSignature.test.ts` L264 went from:

```
'should throw error when message hash is missing for multi-signature'
  -> expects 'Missing message hash for signature at index 1'
```

to:

```
'should throw error when there are more signatures than message hashes'
  -> expects 'Signature count 2 does not match message hash count 1'
```

Grep the file at main: **zero** remaining occurrences of `Missing message hash for signature`. The throw itself is still there at `packages/sdk/src/vault/utils/convertSignature.ts:87`.

## why it matters

That is a fail-closed guard on a signing path. It stops a half-populated signature map from reaching `compileTx()`. A fail-closed guard with no test is how a fail-closed guard quietly becomes fail-open: the next refactor that drops it has nothing telling it no.

## the branch is still live, it just needs a different input

Before #1815 the short-array case was how you got there. Now the count guard intercepts that first. You reach `!messageHash` when the counts **match** and a hash is individually falsy:

```ts
// 2 === 2 clears the count guard, then index 1 is empty
convertToKeysignSignatures(sig2, [first, ''])  // throws at index 1
convertToKeysignSignatures(sig2, ['', second]) // throws at index 0
```

## what this adds

One test, both indices, plus two in-test controls so it cannot rot into proving nothing:

- **non-vacuity**: the same 2/2 shape with both hashes present must NOT throw. Without it the assertions would still pass if the function started throwing unconditionally.
- **discriminator**: a real 2-vs-1 mismatch must report the COUNT message, not the index one. So this test can never be satisfied by the guard that displaced it.

## receipts

No runtime changes (test-only, one file, +42/-0).

**must-pass** - `convertSignature.test.ts` at this head:
```
Test Files  1 passed (1)
     Tests  18 passed (18)      (was 17)
```

**must-fail control** - delete the `!messageHash` throw from `convertSignature.ts` and re-run. The new test goes red, the other 17 stay green, so it is load-bearing rather than decorative:
```
Test Files  1 failed (1)
     Tests  1 failed | 17 passed (18)
```
Source restored and verified clean (`git diff --stat` empty on the src file).

**full unit suite**, base vs this branch, same command both sides:
```
base (aaa3aa93b)  Tests 3274 passed (3274)
this branch       Tests 3275 passed (3275)   <- +1, mine
```
`tsc --noEmit` clean. `prettier --config .config/.prettierrc --check` and `eslint --config .config/eslint.config.mjs` both clean on the touched file.

Heads-up on the suite: running the unit config without scoping to `tests/unit` also pulls in `tests/e2e/**`, which fails 18 files on missing `@helpers/*` aliases + `WASM not configured`. Identical on base, unrelated to this change, so I scoped both sides to `tests/unit` for the comparison above.

## no changeset

Test-only, nothing published changes behavior. Deliberate, not an oversight.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for signature conversion when message hashes are missing at specific indices.
  * Verified successful conversion with complete hashes and clear errors for missing hashes or count mismatches.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because the missing-hash regression coverage is useful, but this old test branch is behind the current signing suite; recreate the focused cases from current main so the exact current guard and discriminator controls are what receive review.

PR #1839, opened 2026-08-10
fix(cosmos): raise TerraClassic staking gas limit to 4M

### LINKED ISSUE #1464: cosmos: getCosmosStakingGasLimit(TerraClassic) stale at 3M, should be 4M (reproduces the #1658 OOG class for external consumers)
Fee-hunt finding. cosmosStakingGasLimitRecord[TerraClassic] = 3_000_000n (cosmosGasLimitRecord.ts:59), pinned by a test whose comment still cites the pre-fix 100 LUNC framing. The app hardened its own TerraClassic gasLimit to 4_000_000 after prod tx_failed telemetry (2026-07-05) showed gasUsed up to 2,501,503 against a 2.5M limit (#1658) - 3M has only ~16% headroom over an already-observed failure. This SDK helper has ZERO internal callers (public API only, re-exported at sdk/src/index.ts:189), so any external consumer building a TerraClassic MsgBeginRedelegate reproduces #1658. Fix: bump to 4_000_000n matching app's headroom rationale, or deprecate in favor of a single canonical source. No cross-repo gas-limit parity test exists.

rcoderdev:
Closes #1464

## Summary by CodeRabbit

* **New Features**
  * Added a public Terra Classic staking fee constant to SDK exports.
  * Increased the Terra Classic staking gas limit to 4,000,000.
  * Terra Classic staking fees are now calculated separately from native-send fees.

* **Bug Fixes**
  * Terra Classic staking transactions now use a fixed gas limit regardless of message count.
  * Improved fee and gas handling for staking and redelegation transactions.

### DIFF
--- a/.changeset/four-million-terraclassic-staking-gas.md
+++ b/.changeset/four-million-terraclassic-staking-gas.md
@@ -1,0 +1,8 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+cosmos/gas: raise the TerraClassic staking gas limit from 3M to 4M
+
+`getCosmosStakingGasLimit` now returns 4M for `Chain.TerraClassic` regardless of `msgCount`, giving external SDK consumers headroom over the observed 2,501,503-gas `MsgBeginRedelegate` path.
--- a/.changeset/four-million-terraclassic-staking-gas.md
+++ b/.changeset/four-million-terraclassic-staking-gas.md
@@ -6,3 +6,5 @@
 cosmos/gas: raise the TerraClassic staking gas limit from 3M to 4M
 
 `getCosmosStakingGasLimit` now returns 4M for `Chain.TerraClassic` regardless of `msgCount`, giving external SDK consumers headroom over the observed 2,501,503-gas `MsgBeginRedelegate` path.
+
+Also exports `TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS`, the `uluna` fee correctly priced for this 4M staking gas limit (113.3 LUNC at the chain's 28.325 uluna/gas minimum). The existing `TERRA_CLASSIC_ULUNA_BASE_GAS` / `getCosmosSendFeeBaseUnits` fee is priced for the 300k native-send gas limit and under-pays a 4M-gas staking tx by ~13x, so consumers must pair the staking gas limit with this new constant, not the send fee.
--- a/packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts
+++ b/packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts
@@ -21,16 +21,15 @@ describe('cosmosGasLimitRecord', () => {
     expect(getCosmosStakingGasLimit({ chain: Chain.Terra })).toBe(500_000n)
     // Terra (not TerraClassic) scales normally: 500k + ((3-1)*500k)/4 = 750k
     expect(getCosmosStakingGasLimit({ chain: Chain.Terra, msgCount: 3 })).toBe(750_000n)
-    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic })).toBe(3_000_000n)
+    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic })).toBe(4_000_000n)
   })
 
-  it('TerraClassic: msgCount scaling capped at base regardless of count (fee-floor constraint)', () => {
-    // Columbus-5 fee floor: 100 LUNC = 100_000_000 uluna. At msgCount>=2 the
-    // scaled gasWanted * gasPrice exceeds 100 LUNC, causing node rejection.
-    // Single-msg policy: always return base=3M for TerraClassic.
-    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic, msgCount: 1 })).toBe(3_000_000n)
-    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic, msgCount: 2 })).toBe(3_000_000n)
-    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic, msgCount: 3 })).toBe(3_000_000n)
+  it('TerraClassic: keeps the 4M single-transaction budget regardless of message count', () => {
+    // The 4M budget covers the observed 2_501_503-gas redelegation path with
+    // headroom. Multi-validator reward claims remain a caller-side split policy.
+    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic, msgCount: 1 })).toBe(4_000_000n)
+    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic, msgCount: 2 })).toBe(4_000_000n)
+    expect(getCosmosStakingGasLimit({ chain: Chain.TerraClassic, msgCount: 3 })).toBe(4_000_000n)
   })
 
   it('rejects invalid staking message counts before BigInt conversion', () => {
--- a/packages/core/chain/chains/cosmos/cosmosGasLimitRecord.ts
+++ b/packages/core/chain/chains/cosmos/cosmosGasLimitRecord.ts
@@ -39,15 +39,13 @@ export const getCosmosGasLimit = (coin: CoinKey<CosmosChain>): bigint => {
  * writes that the standard SDK gas estimate omits. In practice this adds
  * ~200-800 gas on top of the base delegate cost (~2M), so a 2M limit fails
  * consistently ("out of gas in location: ValuePerByte; gasWanted: 2000000,
- * gasUsed: 200X"). 3M needs ~84.97 LUNC at sign time (3M * 28.325 uluna/gas).
- * That fee is supplied by the CONSUMER's separate per-msg staking fee
- * (mcp-ts COSMOS_STAKING_FEE_PER_MSG_BASE_UNITS[TerraClassic] = 100 LUNC/msg,
- * written into the signAmino / signDirect fee) — NOT by
- * `cosmosGasRecord[TerraClassic]`, which is the native MsgSend gas fee
- * (8.4975 LUNC = 300k × 28.325) and never governs staking. The msgCount scaling is disabled for
- * TerraClassic: at msgCount>=2, scaled gasWanted would push the required fee
- * above the 100 LUNC per-msg staking fee, causing node rejection. Columbus-5
- * callers must split multi-validator claims into separate txs.
+ * gasUsed: 200X"). Production telemetry later observed MsgBeginRedelegate
+ * consuming 2_501_503 gas. The former 3M limit left little headroom over that
+ * observed burn, so TerraClassic uses 4M for the heaviest single-msg staking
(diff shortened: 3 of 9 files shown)


gomesalexandre (changes requested):
I re-grounded this on the current head in a fresh disposable clone and I do not think the new 4M TerraClassic staking helper is safe to ship as-is.

The changed helper now advertises `getCosmosStakingGasLimit({ chain: TerraClassic }) === 4_000_000`, but the staking builder still takes `gasLimit` and `feeAmount` independently, and the documented TerraClassic staking fee path is still 100 LUNC. At the chain minimum gas price you cite in-tree, 4_000_000 gas needs 113.3 LUNC, not 100.

Receipts from the current head:
- focused core test: `yarn vitest run --config .config/vitest.core.config.ts packages/core/chain/chains/cosmos/cosmosGasLimitRecord.test.ts` ✅
- direct exact-head builder probe: `buildCosmosStakingTx(... gasLimit: 4_000_000, feeAmount: "100000000")` encodes a SignDoc with
  - `gasLimit = 4000000`
  - `feeAmount = 100000000 uluna`
  - `typeUrl = /cosmos.staking.v1beta1.MsgBeginRedelegate`
- math receipt: `4_000_000 * 28.325 uluna/gas = 113_300_000 uluna = 113.3 LUNC`

So the helper now points consumers at an underpriced tx shape that the node should reject before delegate / undelegate / redelegate / withdraw-rewards can execute.

I think this needs one of:
1. wire a matching TerraClassic staking fee calculation through the actual staking entrypoints and prove the resulting SignDoc shape, or
2. scope the higher gas budget narrowly to the specific redelegation path that observed ~2.5M gas instead of raising every TerraClassic staking flow to 4M.

Right now the tests only prove the helper returns 4M, not that the exported gas/fee pair is still signable on-chain.

gomesalexandre (dismissed):
Fixed the gap I flagged in my CHANGES_REQUESTED review. Pushed f09b78ab1.

The bug: bumping `getCosmosStakingGasLimit(TerraClassic)` to 4M without a matching fee left consumers pairing it with a stale ~100 LUNC fee (or `TERRA_CLASSIC_ULUNA_BASE_GAS`, the *send*-fee constant priced for the 300k send gas limit). Both under-price a 4M-gas tx by ~13x at the chain's 28.325 uluna/gas minimum - the node rejects the SignDoc for insufficient fees before delegate/undelegate/redelegate/withdraw-rewards can broadcast.

Fix: exported `TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS` = 113,300,000 uluna (113.3 LUNC), computed as `getFeeAmountFromGasPrice(getCosmosStakingGasLimit({chain: TerraClassic}), 28.325)` rather than hardcoded, so a future gas-limit retune can't silently desync the paired fee again (same failure mode that started this thread). Updated the doc comments on both the gas-limit record and the staking-gas-limit function to point consumers at the correct fee constant instead of "price it independently."

Receipts:
- unit math: `TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS === 113_300_000n`, independently re-verified outside the codebase (4_000_000 * 28.325 = 113,300,000 exactly)
- signable SignDoc receipt at the public surface: `sdk.cosmosStaking.redelegate(...)` msg paired with `sdk.getCosmosStakingGasLimit` + `sdk.TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS`, asserting fee >= gasLimit * minGasPrice
- confirmed no other Cosmos chain's gas limit touched (diffed against origin/main, only the TerraClassic record entry changed)
- `yarn test` (packages/sdk): 3277 passed, 214 files, 0 failures
- `yarn build` (packages/sdk): clean, exit 0

preferably-blocking item from my prior review is resolved - approving.

gomesalexandre (commented):
Re-grounded current head. The TerraClassic staking gas bump looks correct now, but because this PR's head includes a gomes-authored substantive commit I cannot approve it under the four-eyes rule.

What I checked:
- the new `TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS` is derived from the 4M staking gas limit, so the fee no longer silently desyncs from the gas retune
- the fee export is wired through both the root SDK entrypoint and the RN entrypoint
- the receipt test decodes the built staking `SignDoc` and proves the gas-limit / fee pair matches the new TerraClassic constants

So conceptually this looks good from the SDK side. It now needs a non-gomes approval to merge.

gomesalexandre (commented):
Re-grounded on the new head `8726a937`. The new work since my last pass is the RN export plus its SignDoc test, and it holds up. Rather than re-read the diff and say "looks right", I went and checked the two numbers this PR is actually betting on against live columbus-5, because neither had been grounded in anything but a comment.

## the 28.325 rate is exactly what the chain charges

Three most recent real `MsgBeginRedelegate` txs on columbus-5:

```
gas=3394021  fee= 96135645 uluna  -> 28.3250 uluna/gas
gas=3113634  fee= 88193684 uluna  -> 28.3250 uluna/gas
gas=2083955  fee= 59028026 uluna  -> 28.3250 uluna/gas
```

Exact, to four decimals. Across a wider pull the observed rates are `[28.325, 40.0, 66.6667, 100.0]`, i.e. 28.325 is the floor and everything else is people voluntarily overpaying. So `TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS = 4_000_000 x 28.325 = 113_300_000` is priced at the real chain minimum, not a guess.

## the 4M gas limit is adequate, and redelegate is the binding case

600 single-message, successful (`code=0`) redelegations sampled off the LCD:

```
n=600   min=1_310_532   median=2_456_179   p95=3_071_154   max=3_626_653
exceeding 4M: 0
within 10% of 4M: 1
```

So 4M covers 600/600 of real traffic. Worst observed case sits at 90.7% of the limit.

For the other two staking ops, single-message and successful:

```
MsgDelegate     max=1_634_350   -> 4M headroom x2.45
MsgUndelegate   max=1_852_847   -> 4M headroom x2.16
MsgBeginRedelegate max=3_626_653 -> 4M headroom x1.10
```

Redelegate is roughly 2x the cost of the other two, which makes sense (it touches both source and destination validator), and it is the one that sets the floor for this constant. Picking 4M off redelegate rather than delegate was the right call.

**nit, non-blocking:** the redelegate tail is thinner than the headline suggests. Redelegate gas scales with how much delegation state the account carries, and my sample is whatever happened to be recent, not the heavy accounts. If a redelegation ever does exceed 4M the user does not just get a bounced tx, they pay the full 113.3 LUNC fee and get nothing, because the fee is deducted in the ante handler before execution runs out of gas. Not asking for a change, 4M is defensible on this data, just flagging that 1.10x is the actual margin and it is worth knowing if #1464 comes back.

## a false alarm I chased, so nobody repeats it

The raw `gas_used` maximum across all MsgDelegate traffic is 4_915_436, which is over the 4M limit and looks damning. It is not: that tx is a 7-message batch (6x `MsgWithdrawDelegatorReward` + 1x `MsgDelegate`). Filtering to single-message txs drops MsgDelegate's real max to 1_634_350. Any future look at this needs the message-count filter or it will "find" a blocker that is not there.

## the rest of the head

- RN entry exporting `TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS` next to the gas helper is right, and the new entry test decodes a real SignDoc and asserts `authInfo.fee.gasLimit` and `fee.amount` together, which is stronger than asserting the constants in isolation.
- I checked whether the other platform entries (browser / node / chrome-extension / electron-main) now need the same export for parity. They do not: none of them re-export the cosmos gas constants at all, not even `MAYA_SEND_FEE_BASE_UNITS`. RN's flat curated surface is the exception, so this is not a gap.
- The `publicExports.test.ts` change moves the required-fee assertion from float (`Number(gasLimit) * 28.325`) to integer (`(gasLimit * 28_325n) / 1000n`). That is an improvement, not a weakening. The BigInt division truncates, so the bound is at most 1 uluna looser, which is noise against a 113_300_000 fee.

Not approving, and it is not a quality objection: this head still contains my own `7c4944e5` pricing the fee, so a stamp from me would not be four eyes. Handing over the evidence so whoever picks it up has the numbers.

gomesalexandre (approved):
Good change, and the second half is what makes it good. Bumping the limit on observed telemetry (`MsgBeginRedelegate` at 2,501,503 gas leaving thin headroom under 3M) is the easy part; adding `TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS` **derived from** `getCosmosStakingGasLimit` rather than hardcoded is the part that stops this recurring. The comment naming the earlier gas-limit-only bump as the thing that "silently desynced the paired fee" is the right kind of self-documentation.

```
packages/core/chain/chains/cosmos/   34 files, 444 passed
CI                                    all 5 checks green
```

## the downstream this needs paired with, with numbers

The fee requirement moves, and it crosses a threshold that a consumer is currently sitting under:

```
3M limit -> 84,975,000 uluna  (84.975 LUNC)
4M limit -> 113,300,000 uluna (113.300 LUNC)
```

The two known consumers have already drifted apart on this constant, and only one of them clears the new bar:

```
agent-backend-ts  src/mastra/tools/mcp/tools/staking/cosmos-staking.ts:289
                  TerraClassic: 160_000_000n   (160 LUNC)   >= 113.3   OK

mcp-ts            src/tools/staking/cosmos-staking.ts:284
                  TerraClassic: 100_000_000n   (100 LUNC)   <  113.3   short by 13.3 LUNC
```

mcp-ts's comment reads `100 LUNC per msg (columbus-5 network min-gas-price × ~1M gas)`, so it is priced against a ~1M-gas assumption that predates even the 3M limit. It happened to stay above the 3M requirement by 15 LUNC; at 4M it goes under.

**Stated with the uncertainty it deserves:** I did not fully trace whether mcp-ts's signed staking tx actually carries this SDK's `gasWanted`. mcp-ts sets only the fee amount and never writes `gasWanted` or imports `getCosmosStakingGasLimit`, so the gas is supplied further down the signing path. If that path is this limit, mcp-ts's TerraClassic delegate / undelegate / redelegate / withdraw-rewards start getting rejected for insufficient fees the moment this lands. If it is not, mcp-ts is unaffected and the only takeaway is that the two consumers disagree by 60 LUNC for no documented reason.

Either way it seems worth a companion mcp-ts change (or an issue) rather than leaving it to be discovered on columbus-5, given the PR body is just `Closes #1464` and does not mention downstreams. Happy to file it against mcp-ts if you confirm the wiring - I have that repo checked out and can put the numbers in.

The `TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS` export plus the react-native entry addition is exactly what a consumer needs to stop hand-picking these figures, so the fix for both of them is probably "import that constant" rather than "bump another magic number".

rcoderdev:
Rebased onto current main at `722714428`. The only conflict was the React Native entry-test import block; I retained both current-main custom-RPC imports and the existing SignDoc decoder imports, and the feature range-diff is otherwise patch-identical. Focused tests pass 149/149, full `yarn check:ci` passes, and built ESM/CJS plus React Native QA still returns 4,000,000 gas with the paired 113,300,000 uluna fee. The `mcp-ts` balance-reserve note is a separate consumer follow-up and does not change this SDK fix.

PR #1844, opened 2026-08-10
fix(ripple): refuse a partial payment with no delivery floor

### LINKED ISSUE #1843: ripple: a tfPartialPayment Payment passes the reviewed-amount binding it makes meaningless
## Summary

`getRippleSigningInputs` binds a dApp-supplied `signRipple` Payment to the reviewed `toAddress` / `toAmount` so the bytes being signed match the terms someone approved. The `Amount` half of that binding assumes `Amount` is a delivery. With `tfPartialPayment` set it is not — it is a ceiling — so the check passes byte-for-byte while describing nothing the recipient will actually receive. Severity: **Medium**.

## Background (XRPL semantics)

On XRPL, `tfPartialPayment` (flag `0x00020000`) redefines `Amount` on a `Payment`: it stops being a guaranteed delivery and becomes a **maximum**. The ledger delivers whatever the chosen path can source and records the real figure only in the executed transaction's metadata (`delivered_amount`). Absent a `DeliverMin`, nothing floors what arrives, while the sender can still be charged the full `SendMax`.

## What's wrong

`packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts` describes itself as *"the fail-closed chokepoint that binds the raw transaction to the signing vault … and, for Payments, to the reviewed destination and amount."* It enforces exactly that:

- `tx.Account` must equal `coin.address`
- for a `Payment`, `tx.Destination` must equal `keysignPayload.toAddress`
- for a `Payment`, `tx.Amount` must equal `keysignPayload.toAmount` (numerically, for issued currencies)

`Flags` is never read. A raw transaction that satisfies all three and *also* carries `Flags: 131072` passes unchanged, and the reviewed `toAmount` it was just matched against no longer bounds the outcome. The guarantee the file states is not the guarantee it provides.

`Paths` is likewise unread, letting the initiator steer routing (e.g. through a poisoned order book) within an otherwise-matching payment.

## Impact

The exposure is the cross-currency **self-swap** (`Destination` = the sender's own address, `Amount` = what you receive, `SendMax` = what you pay). An initiator that presents an inflated receive figure as `toAmount` and embeds the same figure in `rawJson` alongside `tfPartialPayment` and no `DeliverMin` clears every gate above: the amounts match, so the resolver forwards it, and the vault can pay the whole `SendMax` for dust.

This matters most for a co-signer. A Secure Vault co-signer sees only the payload it receives over the relay and never ran the initiator's client-side sanitizer, so this resolver is the only thing standing between a compromised initiator and a signature. It is also the sole gate for any other SDK consumer building a `signRipple` payload.

## Fix

In the `Payment` branch of `getRawJson`, after the `Amount` binding:

1. Read `Flags`, requiring a uint32 — refuse anything else (notably the `{ tfPartialPayment: true }` object form some client libraries accept), since an undecodable value may carry the very bit being checked.
2. Refuse a `Payment` that sets `tfPartialPayment` while `DeliverMin` is absent. Forward one that carries a `DeliverMin`: the floor restores something for the reviewed amount to mean.

Unit coverage in `packages/core/mpc/keysign/signingInputs/resolvers/ripple.test.ts` for the refusal, the `DeliverMin` pass-through, a benign flag above `INT32_MAX` (`tfFullyCanonicalSig`, which must not be clipped into a rejection), and a non-uint32 `Flags`.

## Notes

`0x00020000` is namespace-sensitive. It is `tfPartialPayment` only on a `Payment`; on an `OfferCreate` the same bit is `tfImmediateOrCancel`, and as an account-root flag it is `lsfRequireDestTag` (already used under that name in `keysign/chainSpecific/resolvers/ripple.ts`). The check must be scoped to `Payment`.

The client-side half of this — rejecting such a request before it reaches a confirmation screen, and surfacing `tfPartialPayment` / `Paths` to the user when it does — is tracked at vultisig/vultisig-windows#4595.

Ehsan-saradar:
## Description

Fixes #1843

`getRippleSigningInputs` describes itself as *"the fail-closed chokepoint that binds the raw transaction to the signing vault … and, for Payments, to the reviewed destination and amount."* It enforces that: `Account` must be the signing vault, and for a `Payment`, `Destination` and `Amount` must equal the reviewed `toAddress` / `toAmount`.

The `Amount` half assumes `Amount` is a delivery.

### The bug

`tfPartialPayment` (`0x00020000`) redefines it. With the flag set, `Amount` becomes a **maximum**: the ledger delivers whatever the chosen path can source and records the real figure only in the executed transaction's metadata (`delivered_amount`). Absent a `DeliverMin`, nothing floors what arrives, while the sender can still be charged the full `SendMax`.

`Flags` was never read, so such a transaction cleared every gate. The reviewed amount matched byte-for-byte and bounded nothing. The guarantee the file stated was not the guarantee it provided.

The exposure is the cross-currency **self-swap** (`Destination` = the sender's own address, `Amount` = what you receive, `SendMax` = what you pay): an initiator that puts an inflated receive figure in both `toAmount` and `rawJson`, sets the flag, and omits `DeliverMin` gets a signature for paying the whole `SendMax` in exchange for dust.

It matters most for a **co-signer**, which sees only the payload it receives over the relay and never ran the initiator's client-side sanitizer — this resolver is the only thing between a compromised initiator and a signature. It is also the sole gate for any other SDK consumer building a `signRipple` payload.

### The fix

In the `Payment` branch, after the `Amount` binding:

- `Flags` must read as a uint32. Anything else is refused — including the `{ tfPartialPayment: true }` object form some client libraries accept — since a value we cannot decode may carry the very bit being checked.
- `tfPartialPayment` without a `DeliverMin` is refused. **With** a `DeliverMin` it is forwarded unchanged: the floor restores something for the reviewed amount to mean, and partial payments are legitimate.

Scoped to `Payment` deliberately. `0x00020000` is namespace-sensitive: it is `tfImmediateOrCancel` on an `OfferCreate`, and as an account-root flag it is `lsfRequireDestTag` — already used under that name in `keysign/chainSpecific/resolvers/ripple.ts`. Reading it as partial payment anywhere else would be wrong.

## Which feature is affected?

- [x] New Chain / Chain related feature — XRPL keysign (`signRipple` raw-transaction path)

No mainnet tx link: the change only ever *refuses*, and the case it refuses is one that should never have been signable. The accepting paths are unchanged and covered by the existing suite.

## Checklist

## Additional context

Four tests in `ripple.test.ts`:

| case | expected |
|---|---|
| `tfPartialPayment`, no `DeliverMin` | refused |
| `tfPartialPayment` + `DeliverMin` | forwarded verbatim |
| `Flags: 2147483648` (`tfFullyCanonicalSig`) | forwarded — above `INT32_MAX`, so the uint32 bound must not clip it into a rejection |
| `Flags: { tfPartialPayment: true }` | refused |

The first deliberately sets `Destination` and `Amount` to match the reviewed metadata, so every pre-existing gate passes and only the new one can reject it.

`yarn check` and the 2606-test core suite pass.

The client-side half — refusing such a request before it reaches a confirmation screen, and surfacing `tfPartialPayment` / `Paths` to the user when one does arrive — is vultisig/vultisig-windows#4595, whose PR is held in draft until this is merged and released.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards for XRP partial payments, requiring a strictly positive minimum delivery amount.
  * Rejected missing, zero, malformed, insufficient, or asset- and issuer-mismatched minimum delivery values.
  * Rejected transactions with invalid or unreadable flag values.
  * Preserved valid native-asset and issued-currency partial payments, including unrelated flag combinations.

* **Tests**
  * Added coverage for valid, invalid, insufficient, mismatched, and unsupported XRP payment scenarios.


gomesalexandre (approved):
## approve, pushed a follow-up fix

Verdict: approving after digging into the actual XRPL threat model and finding the original `DeliverMin` check didn't close the exact vuln it was supposed to.

## what I verified

Read the diff, both commits (`69a891c` presence-only → `7b82a2b` positive-only), the linked issue #1843, and the companion client PR reference. Ran the full ripple resolver test suite + the entire core suite + typecheck + lint + `build:sdk` (all platform bundles).

Issue #1843's own stated impact is: *"an initiator... sets the flag, and omits DeliverMin gets a signature for paying the whole SendMax in exchange for dust."* That's the self-swap case — `Destination` is the sender's own address, `Amount`/`toAmount` is what they *think* they'll receive, `SendMax` is what they actually pay.

Wrote a PoC before touching anything:

```
Amount = payload.toAmount = '1000000' (matches reviewed metadata, passes every existing gate)
SendMax = '999999999999'
DeliverMin = '1'   // one drop
Flags = tfPartialPayment
```

This passed the guard as originally implemented. `isPositiveXrplAmount` only checked `DeliverMin` was *some* positive well-formed amount — not that it was anywhere near what got reviewed. So the exact "pay SendMax for dust" scenario the issue calls out survives the fix verbatim, just gated behind "attacker also has to add a `DeliverMin` field with any positive value," which is a one-line change for anyone constructing the rawJson.

## what I pushed (3 commits on top of the branch)

`4cbe8ce4` — `DeliverMin` now has to be the *same asset* as `Amount` (native XRP, or matching issued-currency code + issuer) and *at least as much value*, not merely positive. So a `DeliverMin` that floors delivery below the reviewed amount, or in an unrelated currency, gets refused same as a missing one.

`b3fee97f` — updated the changeset text, it still described the old "merely positive" guarantee.

`4d6016eb` — `parseXrplAmount` calls `toXrplCurrencyCode` which throws on an overlong (>20 byte) currency string. The pre-existing `Amount`-matching branch already has this same unguarded call, but my new code is the first place it runs on `DeliverMin` from a *native* XRP Payment without going through that path first — wrapped it in `attempt()` to match the file's own convention for untrusted input (fails closed either way, just cleaner now).

Added 6 new tests (55 total in the file, up from 49): exact-match DeliverMin passes, DeliverMin above Amount passes, DeliverMin below Amount rejected (this is the regression test for the PoC above), issued-currency DeliverMin that matches passes, issued-currency DeliverMin with wrong issuer rejected, overlong currency code rejected instead of throwing raw.

## scope note — deliberately NOT touching

`SendMax` and `Paths` are still unbound/unread by this resolver. Issue #1843 calls out `Paths` explicitly ("letting the initiator steer routing through a poisoned order book") but its own **Fix** section only prescribes `Flags`/`DeliverMin` handling — `Paths` is left to the client-side companion PR (`vultisig-windows#4595`, still draft). Not touching that here, it's a separate surface and out of scope for this fix.

## the tradeoff worth flagging

My stricter check means a dApp-initiated `signRipple` Payment can no longer get *any* real partial-payment slippage tolerance through this resolver — `DeliverMin` has to guarantee the full reviewed amount. The PR description says "partial payments are legitimate," and that's true on-ledger, but Vultisig's client doesn't yet surface `DeliverMin`/`Paths` to the user for review (that's what `vultisig-windows#4595` is for, still in draft). Until there's a client-reviewed field to bind `DeliverMin` against, any value below `Amount` is unreviewed and exploitable exactly as PoC'd above — so I went with correct-but-conservative over permissive-but-unverifiable.

Once #4595 ships with an actual reviewed slippage/DeliverMin field, this resolver will need a follow-up to bind against *that* instead of requiring `DeliverMin == Amount`. Worth a tracking issue so it doesn't get lost — happy to file one if useful.

## close-ref

PR body has `Fixes #1843` on the first line, good.

praise: the original writeup (issue + PR description) is genuinely excellent — the namespace-collision note on `0x00020000` (`tfImmediateOrCancel` on OfferCreate, `lsfRequireDestTag` as an account flag) saved real time here.

gomesalexandre (commented):
pushed `1ed712863` to clear the current red ci on this head.

what changed
- ran the repo-pinned prettier formatting on `packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts`
- only change is line wrapping on the `tfPartialPayment` guard condition. no logic change.

receipts
- `node /private/tmp/sdk1844-prettier.ihRZMy/node_modules/prettier/bin/prettier.cjs --config .config/.prettierrc --ignore-path .config/.prettierignore --check packages/core/mpc/keysign/signingInputs/resolvers/ripple.ts` ✅
- `git diff --check` ✅

note
- the normal `yarn prettier ...` path was unavailable in this disposable clone because `node_modules` was absent and `yarn install --immutable` failed during an unrelated link step, so i used the cached repo-pinned prettier 3.8.4 artifact directly for the one-file check/write.
- this branch already includes substantive gomes-authored commits, and i just pushed another commit, so it still needs a non-gomes current-head approval before merge.

PR #1846, opened 2026-08-10
fix(evm): reserve the OP-stack surcharges an EVM fee is checked against

### LINKED ISSUE #4544: Verify max-send fee invariant on EVM L2s: clamp amount to the signed maxFeePerGas at payload build (iOS bug sibling)
## What
Sibling of the confirmed iOS bug (vultisig/vultisig-ios#5019): max send of a native coin on EVM L2s (OP, ARB) can fail at broadcast with `RPC Error -32000: insufficient funds for gas * price + value` when the base fee moves between the fee estimate that derived the max amount and the chainSpecific that actually gets signed. On iOS the root cause is two separate fee fetches with no re-clamp at payload-build time; L2s' ~2s blocks make the drift near-constant.

## Ask
The max amount is computed at UI time from a fee estimate (`core/ui/vault/send/amount/ManageAmountInputField.tsx:71-77` — `getMaxValue(balance, feeEstimateQuery.data)`). Verify whether the keysign payload is later built with a *fresh* gas fetch, and if so enforce the invariant: **for a max send, clamp `value = balance − gasLimit × maxFeePerGas` using the payload's own values at build time.** Also check OP-stack L1 data-fee headroom (op-geth includes it in the balance check). Covers desktop and extension via shared core/ui.

Ehsan-saradar:
## Problem

A native max send on an OP-stack rollup is rejected at broadcast with `insufficient funds for gas * price + value`, after the MPC ceremony has already been paid for.

op-geth does not check `value + gasLimit * maxFeePerGas` against the balance. It checks

```
value + gasLimit * maxFeePerGas + l1Cost + operatorCost <= balance
```

(`buyGas` in `core/state_transition.go`; `TotalTxCost` in `core/txpool/validation.go`.)

A max send sets `value = balance − gasLimit * maxFeePerGas` **exactly**, so the required balance evaluates to `balance + surcharge`. It overdraws by the surcharge on *every* attempt — this is deterministic, not a race. Nothing in the SDK read either term.

### The two terms are real, and separately debited

Measured from real transactions' sender balance deltas (`balanceBefore − balanceAfter`, single-sender blocks only):

| chain | value | L2 execution | L1 data fee | unexplained | residual |
|---|---:|---:|---:|---:|---:|
| Optimism | 9,000,000,000,000 | 21,008,442,000 | **1,241,361,344** | 0 | **0** |
| Mantle | 1,510,211,000,823,637,713,741 | 1,050,420,000,000,000 | **121,716,148,039,182** | **210,000,000,000,000** | **0** |
| Arbitrum (control) | 812,279,089,760 | 425,992,800,000 | 0 | 0 | **0** |

Mantle's unexplained 2.1e14 is exactly `getOperatorFee(21000)` — its `operatorFeeScalar` is 1e8 with a zero constant, which prices at 1e10 wei per gas. Arbitrum shows no L1 fee line and nothing unexplained, which is why it is correctly out of scope for this reserve (its own fee-drift exposure is already handled by `refineKeysignAmount`).

### Which chains

`isOpStackChain` is asserted, not probed at runtime. Verified by reading the `GasPriceOracle` proxy's implementation bytecode selector table on each chain's own RPC endpoint:

| chain | predeploy | `getL1Fee(bytes)` | `getOperatorFee(uint256)` |
|---|---|---|---|
| Optimism, Base | present | ✅ | ✅ (scalars zero today) |
| Blast | present | ✅ | ❌ absent from the impl — pre-Isthmus oracle |
| Mantle | present | ✅ | ✅ non-zero |
| Arbitrum, Ethereum, zkSync, Sei, Polygon, BSC, Avalanche, Hyperliquid | **no code** | — | — |

A test pins that every `EvmChain` is classified, so a newly added chain cannot slip through unclassified.

## The fix

The surcharge is folded into `getEvmFeeAmount` rather than bolted onto the max-send path. That resolver is the single number every spend-capacity path already asks for — the max quote on the amount screen, the form's balance validation, and `refineKeysignAmount`'s re-fit of the signed value at payload build — so one change makes all three agree, and there is no second definition of "what this transaction costs" to drift out of sync.

It also keeps the resolver's existing meaning intact: the gas term was already `gasLimit * maxFeePerGas`, the worst case the node holds up front rather than the eventual `gasUsed * effectiveGasPrice`. The surcharges are held the same way, against the gas *limit*. `FeeAmountResolver` already permitted a `Promise`, so no signature moves.

**Both terms fail open, independently.** Blast's oracle reverts on `getOperatorFee`, and letting that failure take the L1 data fee down with it would leave the chain reserving nothing at all. A fee that cannot be read must never block a send that worked before — an unreachable oracle degrades to exactly today's behaviour. A negative answer is floored rather than passed on to widen a max amount.

**The probe payload is deliberately incompressible.** `getL1Fee(bytes)` has to price a transaction that does not exist yet, since the amount is derived from the answer. From Fjord onwards the oracle charges for the *FastLZ-compressed* size of what it is handed, so a counter — which wraps every 256 bytes — would compress away and halve the reserve on a long payload. The LCG stream repeats no 3-byte window (the shortest match FastLZ can encode); a test pins that, and contrasts it against a counter to show the guard bites. `getL1Fee(bytes)` is preferred over the purpose-built `getL1FeeUpperBound(uint256)` because the latter only exists from Fjord and Blast's oracle predates it.

The probe is sized from the transaction's own calldata — swap quote `tx.data` when present, otherwise the memo — mirroring how the EVM chain-specific resolver picks `data`, so the bytes priced are the bytes signed.

## ✅ On-device verification — broadcast max sends

Three real max sends signed and broadcast from the extension with this change in place.

### Optimism — [`0x4c10ab26…8f889f`](https://optimistic.etherscan.io/tx/0x4c10ab268f049459bee71ed85fddd8d68c681e72ac23851f37b8c7de8a8f889f), block 155380589, **SUCCESS**

99.992672% of the balance left the account, so this is the max path and not a typed amount.

| | wei |
|---|---:|
| balance before | 573,477,369,421,250 |
| value sent | 573,435,345,019,797 |
| gas term (40,000 × 1,000,594) | 40,023,760,000 |
| **surcharge reserved** | **2,000,641,453** |
| L1 data fee actually charged | 1,338,214,036 |
| operator fee charged | 0 |
| unaccounted | **0** |

op-geth's check: `value + gasTerm + l1Cost` = 573,476,706,993,833 ≤ balance 573,477,369,421,250 → **PASS**, headroom 662,427,417.

**What the old code would have signed**, `value = balance − gasTerm` = 573,437,345,661,250, required 573,478,707,635,286 — **short by exactly 1,338,214,036 wei, the L1 data fee**. That is the reported failure, reproduced from this transaction's own numbers.

### Base — [`0x2b164fda…cf27ae`](https://basescan.org/tx/0x2b164fda7dc65ce0d0411f8e6b1c43c700002bff1432bac0a415715b20cf27ae), block 49785383, **SUCCESS**

99.341549% of the balance sent.

| | wei |
|---|---:|
| balance before | 51,827,765,662,047 |
| value sent | 51,486,505,374,992 |
| gas term (40,000 × 8,500,000) | 340,000,000,000 |
| **surcharge reserved** | **1,260,287,055** |
| L1 data fee actually charged | 844,965,778 |
| operator fee charged | 0 |
| unaccounted | **0** |

Check passes with 415,321,277 headroom; the old value would have been **short by exactly 844,965,778 wei**, again the L1 data fee to the wei.

### Ethereum L1 control — [`0x0b86440b…ea3135`](https://etherscan.io/tx/0x0b86440ba169e0c798bde6ebfbab5e8a6373e34c9e6d47b97d6b066291ea3135), block 25724161, **SUCCESS**

`value + gasLimit × maxFeePerGas` = 5,542,030,949,981,346 = balance before, **residual 0**. The amount is still exactly `balance − fee` to the wei, nothing extra shaved off, and the receipt carries no `l1Fee` field. The new code path correctly does not fire off the OP stack — a non-OP-stack chain costs zero RPC round-trips.

### The estimator behaves as designed

The reserve overshot the realised L1 fee by **1.495×** on Optimism and **1.492×** on Base — matching the ~1.5× margin the incompressible probe is documented to produce, on two chains independently, rather than landing there by luck.

The residual dust is dominated by the *pre-existing* gas-limit headroom, not by this change. On Optimism, of 19,677,829,417 wei left behind, 19,015,402,000 is the ordinary 40,000-vs-21,000 gas refund and only 662,427,417 comes from the new reserve — roughly 6.6e-7 ETH.

**Still unrun:** an ERC-20 MAX (must be the full token balance), a typed non-MAX amount (must be signed as entered), an Arbitrum max (non-OP-stack L2 control, must be unchanged), and a swap on an OP-stack chain (quoted network fee moves up by ~3%). All four are covered by unit tests; none has had a broadcast.

## Live surcharge behaviour

Through the built dist, against the chains' own endpoints:

```
Optimism  plain          gas=60000000000          surcharge=1874140956            (1777ms)
Base      plain          gas=300000000000         surcharge=1260652355            (604ms)
Blast     plain          gas=60000000000          surcharge=358807562             (624ms)
Mantle    plain          gas=6750000000000000000  surcharge=900000038333829745    (431ms)
Arbitrum  plain          gas=2880000000           surcharge=0                     (0ms)
Ethereum  plain          gas=69000000000000       surcharge=0                     (0ms)
Optimism  with calldata  gas=900000000000         surcharge=8245103490            (524ms)
```

Non-OP-stack chains cost zero round-trips — the chain check gates before any RPC.

## Tests

14 new, no network:

- `opStackChains` — membership both ways, plus exhaustive coverage of `EvmChain` so a new chain must be classified.
- `l1FeeProbeData` — determinism, prefix-stability as size grows, zero repeated 3-byte windows at real payload sizes, and aperiodicity where a counter collapses.
- `getOpStackFeeSurcharge` — both terms summed; each term surviving the other's failure; both failing reserving nothing; negatives floored; the operator lookup skipped at zero gas limit; the priced payload growing with calldata.
- `getEvmFeeAmount` — gas term alone off OP-stack (with no oracle call made), gas + surcharge on it, correct gas limit and calldata size passed through, and a regression showing a max send now satisfying the balance check the node actually runs.
- `refineKeysignAmount` — first tests for this function. Pins the payload-build invariant end to end: refits a balance-derived amount to the fee the payload carries, **clamps down only** (a fee that fell must never raise the amount above what was approved), leaves typed amounts and token amounts alone, leaves UTXO/TON alone, and refuses before the ceremony when the fee swallows the balance.

## Verification

Green: typecheck, lint, knip, format, dependency-cruiser, complexity, duplication, bundled-changeset guard, 2,622 core tests, 3,274 SDK unit tests.

Also validated downstream against a consumer app with this dist dropped in: typecheck, full test suite and a production extension build all pass — the three broadcasts above were signed from that build.

## Notes

- Fixes vultisig/vultisig-windows#4544. That repo needs no code change — it picks this up through its pinned `@vultisig/core-chain` / `@vultisig/core-mpc` versions.
- **Out of scope, found while measuring:** `deriveEvmGasLimit` hardcodes `mantleTransferGasLimit = 90_000_000n`, while real Mantle transfers use 400k–900k limits and consume 21,000 gas. At a ~5e10 base fee that reserves ~6.75 MNT against a transfer costing ~0.001 MNT, so a Mantle max send strands ~6.75 MNT and a wallet holding less cannot send natively at all. Filed as #1847 — it is a signing-path value that deserves its own change.

## Summary by CodeRabbit

- **New Features**
  - Improved maximum native-asset send calculations on supported OP Stack networks by accounting for L1 and operator fee surcharges.
  - Added support for Optimism, Base, Blast, and Mantle network classification.
  - Published OP Stack fee estimation utilities for broader SDK and core package use.

- **Bug Fixes**
  - Improved fee handling when oracle data is unavailable or invalid.
  - Prevented fee estimates from becoming negative and preserved transaction affordability checks.

- **Tests**
  - Expanded coverage for fee estimation, calldata sizing, network classification, and maximum-send behavior.


gomesalexandre (approved):
Went deep on this one given the fund-safety surface (fee reserve math directly gates broadcast success/failure).

**What I verified:**
- Read every changed file: `getOpStackFeeSurcharge.ts`, `l1FeeProbeData.ts`, `opStackChains.ts`, the `evm.ts` fee resolver. Confirmed op-geth's balance check formula (`value + gasLimit*maxFeePerGas + l1Cost + operatorCost`) is matched term-for-term, both oracle reads fail open independently (verified a pre-Isthmus oracle reverting on `getOperatorFee` doesn't take the L1 data fee down with it), negative answers are floored not passed through, and the `opStackChains` exhaustiveness test genuinely covers all 14 `EvmChain` members today (4 op-stack + 10 non — checked the enum myself, it adds up).
- Traced the calldata-size parity claim in the comment ("mirrors how the EVM chain-specific resolver picks the transaction's `data`") against the actual signing-path resolver at `chainSpecific/resolvers/evm.ts:36-46` — same `swapPayload.general.quote?.tx?.data || memo` fallback, byte-identical logic. No drift between what gets priced and what gets signed.
- Traced the whole call graph: `refineKeysignAmount` → `getFeeAmount` → `getEvmFeeAmount` → `getOpStackFeeSurcharge`, and confirmed SDK's `getMaxSendAmountFromKeys`/`computeMaxSendFromBalance`, `VaultBase`, and the CLI agent executor all route through this same single resolver. No second/bypass path that would miss the surcharge.
- `yarn test:core` (2655/2656), SDK unit tests (3273/3274) — the 1 failure in each case is a pre-existing flaky RN-entry/MPC-singleton test unrelated to this diff (reproduced identically on `main` in a throwaway worktree to confirm it's not a regression). `yarn build` (SDK + shared) clean.
- **Live runtime receipt, not just mocks**: wrote a throwaway vitest file hitting the real `GasPriceOracle` predeploy on Optimism and Base mainnet through `getOpStackFeeSurcharge` directly. Got sane non-zero values (2,320,968,255 wei on a 21k-gas Optimism transfer, 2,104,710,191 wei on a Base ERC-20-transfer-shaped call) that scale correctly with calldata size (2.3e9 → 2.86e10 wei going 0→2000 bytes, consistent with post-Ecotone linear blob pricing). Confirms the predeploy address, ABI, and math all check out against live chain state, independent of the author's own on-chain broadcast verification in the PR description (which I cross-checked and is consistent with what I got).
- Chain classification double-checked against reality: Optimism/Base/Blast/Mantle are genuinely OP-stack (Mantle migrated to OP Stack via Tectonic), Arbitrum/zkSync are correctly excluded (different L1-fee mechanisms entirely, not OP-stack). Arbitrum's own separate surcharge exposure is out of scope here and appropriately left alone.

**Not a blocker, just a note:** close-ref (`Fixes vultisig/vultisig-windows#4544`) is present in the body but under `## Notes` near the bottom rather than the first line — GitHub's auto-close still fires on merge since it scans the whole body, so this isn't functionally broken, just a style nit.

Changeset is correct (patch bump on the 3 touched packages, `@vultisig/core-chain` / `@vultisig/core-mpc` / `@vultisig/sdk`).

This is exactly the kind of PR I want to see for anything touching fee reserve math — real op-geth source citations, a test that pins exhaustive chain coverage so nothing silently slips through, fail-open design explicitly reasoned about, and actual on-chain broadcasts to back the claim. Approving.

PR #1722, opened 2026-08-04
fix(tx): preserve normalize/split routing metadata

gomesalexandre:
## summary
- preserve top-level routing metadata when `normalizeTx()` wraps a flat payload
- enrich multi-leg split parents with `chain` / `from_chain` / `to_chain` args before splitting
- add regression coverage and a changeset for the canonical envelope contract

closes #1721

## receipts
- `corepack yarn workspace @vultisig/sdk vitest run tests/unit/tx/normalize.test.ts`
- `corepack yarn build:shared`
- `corepack yarn build:sdk`
- `corepack yarn workspace @vultisig/sdk pack --out /tmp/vultisig-improve-r103/artifacts/vultisig-sdk-r103.tgz`
- tarball consumer proof:
  - installed the packed SDK into a clean temp consumer
  - imported `normalizeTx` + `splitMultiTx`
  - observed:

```json
{"normalized":{"chain":"Ethereum","from_chain":"Ethereum","to_chain":"THORChain","provider":"thorchain","from_symbol":"USDC","to_symbol":"RUNE"},"split":[{"tx":{"to":"0xtoken"},"chain":"Ethereum","from_chain":"Ethereum","to_chain":"THORChain","provider":"thorchain"},{"swap_tx":{"to":"0xrouter"},"chain":"Ethereum","from_chain":"Ethereum","to_chain":"THORChain","provider":"thorchain"}]}
```

## artifact
- https://gist.github.com/gomesalexandre/4beac47056690192aee8c58194346cdb

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved routing metadata when normalizing transactions.
  - Ensured routing arguments carry over when splitting multi-step transactions into individual approval, swap, or transaction legs.

- **Tests**
  - Added coverage for metadata preservation and routing argument propagation across normalized and split transactions.

PR #1716, opened 2026-08-04
fix(tx): preserve camelCase splitMultiTx metadata

gomesalexandre:
## summary
- preserve camelCase metadata on `splitMultiTx()` `transactions[]` legs
- keep `chainId` / `fromChain` / `toChain` / symbol / address / decimals metadata on per-leg envelopes
- add a regression test so newer sdk-native multi-leg builders cannot silently lose routing context again

closes #1715

## receipts
- `corepack yarn vitest run packages/sdk/tests/unit/tx/normalize.test.ts`
- `corepack yarn build:sdk`
- `corepack yarn workspace @vultisig/sdk pack --out /tmp/vultisig-improve-r102/artifacts/vultisig-sdk-r102.tgz`
- tarball consumer proof: `[{"chain":"Base","chainId":"8453","fromChain":"Base","toChain":"Arbitrum","provider":"cctp","fromSymbol":"USDC","toSymbol":"USDC"},{"chain":"Base","chainId":"8453","fromChain":"Base","toChain":"Arbitrum","provider":"cctp","fromSymbol":"USDC","toSymbol":"USDC"}]`

## artifact
- https://gist.github.com/gomesalexandre/f1a1fdc795d8cb76f74fd02a8fce8da4

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved camelCase transaction metadata when splitting multi-transaction results.
  * Ensured each generated transaction leg retains chain, routing, provider, address, asset symbol, and decimal information.

* **Tests**
  * Added coverage verifying metadata preservation across all split transaction legs.

PR #1858, opened 2026-08-10
fix(tx): splitMultiTx drops camelCase metadata on transactions[] legs

### LINKED ISSUE #1715: fix(tx): splitMultiTx drops camelCase metadata on transactions[] legs
## summary
splitMultiTx only preserved the older snake_case metadata keys. Current SDK-native builders emit camelCase metadata, so multi-leg CCTP and similar flows lost routing/display context on each wrapped leg.

## evidence
- `packages/sdk/src/tx/normalize.ts:162-184`
- `packages/sdk/tests/unit/tx/normalize.test.ts:254-289`
- `packages/sdk/src/tools/bridge/buildCctpBridge.ts:274-301`

## impact
Newer SDK-native multi-leg builders lose chainId/fromChain/toChain/symbol/address metadata when split into per-leg envelopes, so downstream app/backend consumers must reconstruct routing/display context by hand.

## fix sketch
Preserve both snake_case and camelCase metadata families in LEG_METADATA_KEYS, keep the regression test, and verify the built package still returns chainId/fromChain/toChain on split legs.

## audit artifact
https://gist.github.com/gomesalexandre/f1a1fdc795d8cb76f74fd02a8fce8da4

gomesalexandre:
closes vultisig/vultisig-sdk#1715

## what

`splitMultiTx`'s `wrapSingleTx` only looked up `LEG_METADATA_KEYS` in their snake_case form (`chain_id`, `from_chain`, `from_symbol`, ...), matching the Go backend's `wrapSingleTx` shape. Newer SDK-native builders emit camelCase instead - `buildCctpBridge.ts` returns `chainId`/`fromChain`/`toChain`/`fromSymbol`/`toSymbol` - so every multi-leg CCTP bridge split (approve + burn legs) silently dropped routing/display metadata on each wrapped leg.

## why it matters

Downstream consumers of a split leg (UI display, chain resolution for signing) rely on `chain`/`chain_id`/`from_chain`/`to_chain` etc riding along on every leg. A camelCase-emitting parent meant those fields were just... missing, on both legs of every CCTP bridge tx.

## fix

`wrapSingleTx` now falls back to a camelCase alias when the snake_case key is absent from the parent envelope. Output legs stay snake_case (no downstream contract change) - only the *source* lookup gained the camelCase fallback, and an existing snake_case value always wins over its alias when both are present (covered by a test).

## testing

- New regression test reproducing the CCTP-shaped bug (camelCase parent -> legs lose all metadata) - confirmed it fails on unmodified `normalize.ts` (via `git stash`) and passes with the fix.
- New test confirming snake_case still takes priority when both forms are present.
- Full `tests/unit/tx/normalize.test.ts` suite: 22/22 passing.
- Typecheck: 29 pre-existing `@vultisig/mpc-types` resolution errors present identically before and after this change (unrelated - this worktree hasn't run the `build:shared` step CI runs first; none reference `normalize.ts`).
- Lint: clean on both changed files.

## risk

Low - additive fallback only, existing snake_case behavior unchanged, scoped to one function.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metadata handling for multi-leg transactions.
  * CamelCase routing metadata is now recognized and correctly propagated to transaction legs.
  * When both camelCase and snake_case values are provided, snake_case values take precedence.
  * Existing snake_case metadata remains available in normalized transaction output.

### DIFF
--- a/packages/sdk/src/tx/normalize.ts
+++ b/packages/sdk/src/tx/normalize.ts
@@ -173,6 +173,23 @@ const LEG_METADATA_KEYS = [
   'to_decimals',
 ] as const
 
+// camelCase aliases some newer SDK-native builders emit for the same concepts (e.g.
+// buildCctpBridge.ts returns chainId/fromChain/toChain/fromSymbol/toSymbol, not the snake_case
+// shape below). Output legs stay snake_case (matches Go wrapSingleTx and downstream consumers),
+// but source lookup falls back to the camelCase key when the snake_case one is absent, so a
+// camelCase-emitting parent doesn't silently lose its leg metadata.
+const LEG_METADATA_CAMEL_ALIASES: Partial<Record<(typeof LEG_METADATA_KEYS)[number], string>> = {
+  chain_id: 'chainId',
+  from_chain: 'fromChain',
+  to_chain: 'toChain',
+  from_symbol: 'fromSymbol',
+  to_symbol: 'toSymbol',
+  from_address: 'fromAddress',
+  to_address: 'toAddress',
+  from_decimals: 'fromDecimals',
+  to_decimals: 'toDecimals',
+}
+
 /**
  * Wrap a single child tx under `txKey` and copy routing metadata from the
  * parent envelope. Mirrors Go `wrapSingleTx`.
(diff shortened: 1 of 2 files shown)


rcoderdev (commented):
NormalizedTx.chain_id is declared as a string, but the camelCase fallback copies chainId verbatim. buildCctpBridge returns a numeric chainId, and the new test now expects leg.chain_id to be 1, so downstream code that type-checks against the public contract can call string methods and fail at runtime. Could we stringify this alias before assigning it, or deliberately widen the public type and audit consumers, then make the test match that contract?

rcoderdev:
Closing this because The camelCase split metadata fix already shipped through #1716 and #1722, and #2138 owns the remaining typing follow-up. This conflicting branch also violates the public string chain_id contract.

Issue #1369, opened 2026-07-17
chainSpecific fee resolvers untested on the fee-computation side (EVM/THOR/Maya/UTXO/Bittensor/Polkadot)


gomesalexandre:
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P2 CONFIRMED (partial - Cosmos/Tron already covered).

**Anchor:** packages/core/mpc/keysign/chainSpecific/resolvers/evm.ts, thor.ts, maya.ts, utxo.ts, bittensor/, polkadot/ (no sibling .test.ts for any of these)

## what
Golden mobile fixtures pre-populate chainSpecific, so they exercise the signing side but not the fee-computation side. getEvmChainSpecific (gas price/limit/nonce/priority) has zero test coverage, and the same is true for thor.ts, maya.ts, utxo.ts, and the bittensor/polkadot resolver dirs. (Cosmos gasEstimation and Tron fee/energyPrice already have dedicated tests - the audit's blanket claim was too broad, but the EVM/THOR/Maya/UTXO/Bittensor/Polkadot gap is real and verified on current main.)

## failure scenario
A wrong fee/gas/nonce computation on any of these untested resolvers - overpay, stuck tx, or a fee that drains the send to nothing - would ship uncaught, since nothing pins the computed fee/gas/nonce independent of the signed-bytes assertion.

## fix
Add golden fixtures that pin the computed fee/gas/nonce for evm.ts, thor.ts, maya.ts, utxo.ts, and the bittensor/polkadot resolvers, not just the signed bytes.

PR #1861, opened 2026-08-10
test(mpc): pin EVM chainSpecific fee/gas/nonce computation

gomesalexandre:
tackles vultisig/vultisig-sdk#1369 (EVM slice only - THOR/Maya/UTXO/Bittensor/Polkadot resolvers from the issue remain untested)

## what

`getEvmChainSpecific` (packages/core/mpc/keysign/chainSpecific/resolvers/evm.ts) had zero dedicated test coverage. Golden mobile fixtures pre-populate chainSpecific and exercise the signing side, but nothing pinned the computed nonce/gas/fee values independently - a wrong computation could ship uncaught.

Note the underlying `getEvmFeeQuote` (legacy vs EIP-1559 branching, RPC-tip clamping, gas-limit buffering) already has solid coverage in `getEvmFeeQuote.test.ts`. The gap this PR closes is the orchestration layer on top of it that had none:

- **nonce sourcing**: `pending` block tag happy path, and fallback to `latest` when the RPC rejects `pending` (alt-EVM compat path)
- **calldata selection for gas-limit derivation**: general swap quote tx data used when present, falls back to `memo` when swap quote has no tx data / native swap payload / no swap payload at all
- **fee-quote wiring**: derived minimum gas limit and pass-through args (`feeSettings`, `thirdPartyGasLimitEstimation`) correctly forwarded into `getEvmFeeQuote`
- **result assembly**: `maxFeePerGasWei` = `baseFeePerGas + maxPriorityFeePerGas`, all three bigint fields correctly stringified into the `EthereumSpecific` schema, including a zero-value edge case

## why

A wrong fee/gas/nonce computation on this resolver ships as an overpay, a stuck tx (nonce collision from missing the pending fallback), or a fee that drains the send to nothing - and nothing today would catch it before it hits a wallet.

## scope

EVM only, per the issue's own note that a landed EVM slice is more valuable than a shallow pass across all six. THOR/Maya/UTXO/Bittensor/Polkadot resolvers are still open in #1369.

## testing

```
yarn vitest run --config .config/vitest.core.config.ts packages/core/mpc/keysign/chainSpecific/resolvers/evm.test.ts
# 9 passed
```

Typecheck clean on the new file (pre-existing unrelated `@vultisig/mpc-types` module-resolution errors exist on main, untouched by this diff). Lint clean on `packages/core/**`.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because The mocked EVM orchestration coverage is useful, but this test-only head is behind current main and covers only one slice of the still-open broader issue. Recreate the focused tests from current main if that coverage remains missing.

PR #1674, opened 2026-07-31
fix(examples-shared): correct Ethereum DAI suggestion address

gomesalexandre:
##summary
- correct the Ethereum DAI contract in `examples/shared`'s `COMMON_TOKENS`
- add a regression test that validates all shipped suggested EVM token addresses
- pin Ethereum DAI to the canonical mainnet contract

closes #1673

##verification
- `corepack yarn vitest run examples/shared/src/constants/tokens.test.ts`
- `corepack yarn workspace @vultisig/examples-shared typecheck`
- `corepack yarn tsx -e "import { COMMON_TOKENS } from './examples/shared/src/index.ts'; const dai = COMMON_TOKENS.Ethereum.find(t => t.symbol === 'DAI'); console.log(JSON.stringify(dai));"`
- runtime receipt:
  - `{"symbol":"DAI","name":"Dai Stablecoin","contractAddress":"0x6B175474E89094C44Da98b954EedeAC495271d0F","decimals":18}`

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the Ethereum DAI contract address to the canonical mainnet address.

* **Tests**
  * Added validation to ensure suggested token addresses are valid EVM contract addresses.
  * Added coverage confirming Ethereum DAI uses the correct contract address.

PR #1862, opened 2026-08-10
fix(prep): enforce native/evm/solana amount-quote consistency in vault-free swap builder

### LINKED ISSUE #1117: [audit][MED][SDK-FUNDSAFETY-01] vault-free swap helper skips amount-quote cross-check + expiry the vault path enforces
tools/prep/swap.ts (docstring: 'intended for MCP servers and other contexts') takes amount and swapQuote as independent params with no consistency check -- docstring explicitly punts 'Quote expiry validation is also a consumer concern.' For native swaps, keysign/swap/build.ts:129 takes the signed input from the caller's amount while the min-out limit comes from the quote's own sizing. The vault wrapper SwapService.ts:131 enforces expiry; the vault-free helper bypasses it.

Impact: agent path can build a native-swap payload whose input amount and slippage limit were computed for different sizes, or sign a stale aggregator quote. Bounded by the aggregator's on-chain min-return (worst case: revert or bounded-worse fill), hence MED not HIGH.

Found via round-1 audit artifact re-scan (never beaded first pass).

gomesalexandre:
closes vultisig/vultisig-sdk#1117

## what

`prepareSwapTxFromKeys` (the vault-free, agent-reachable swap payload builder, used wherever only the public identity is available - no key shares) already enforces native/CoW quote expiry and CoW gross-amount consistency, landed in #1092 (ABTS/plan 005). But that PR's own body named a residual: native, EVM, and Solana routes still fail-open on amount consistency, because their quote responses carry no committed-input-amount field the SDK can compare against (checked the actual `NativeSwapQuote` type - confirmed empty). #1117 is that exact residual, caught by a round-1 audit re-scan against post-#1092 main.

## how

`findSwapQuote` (`packages/core/chain/swap/quote/findSwapQuote.ts`) now stamps every returned `SwapQuote` with `requestedAmount` - the base-unit amount it actually fetched the quote for. This is the amount the caller supplied when calling `findSwapQuote`, independent of provider shape, so it works generically for every route (native, EVM, Solana, transfer, CoW) without decoding opaque aggregator calldata or relying on a provider-committed field.

`prepareSwapTxFromKeys` cross-checks the caller's signing-time `amount` against `swapQuote.requestedAmount` and throws on mismatch, closing the native/EVM/Solana gap. The existing CoW-gross-sell-amount check stays as-is (defense in depth for a hand-built quote lacking `requestedAmount`).

`requestedAmount` is **optional** on the `SwapQuote` type - it's only populated at the real `findSwapQuote` fetch sites. That's deliberate: this type is constructed by hand in ~40 test fixtures across 8 test files (`SwapService.test.ts`, `build.transfer.test.ts`, etc.), and making the field required would have forced a much larger, unrelated diff across all of them. The check fails open when the field is absent, same fail-open philosophy #1092 already established for the CoW/transfer exclusions.

## testing

4 new unit tests in `swap.test.ts` (native + EVM, matching + mismatching `requestedAmount`) - all fail pre-fix (mismatch amounts pass through un-rejected) and pass post-fix. Full existing suite for the file (18 tests total) still green.

`yarn workspace @vultisig/sdk typecheck` clean. Root `tsc --project .config/tsconfig.json --noEmit` (covers `packages/core`) clean. Lint clean on all 4 changed files.

`yarn vitest run tests/unit` (sdk): 3252/3291 passed - identical failure count/files to an unmodified-main baseline run (39 failures, both pre-existing MPC-singleton test-isolation issues in `platforms/react-native/entry.test.ts`, unrelated to this change - confirmed via `git stash` diff). `tests/integration/swap/swap-quote.test.ts` has 2-3 flaky network-timeout failures on both baseline and this branch (different tests fail each run - live-network-dependent, not deterministic, not touched by this diff).

Added a `@vultisig/sdk` changeset (`vault-free-swap-amount-consistency-1117.md`) since the change also touches `packages/core`, which the bundled-changeset CI guard covers - guard passes locally.

## risk

LOW - additive check only, fails open when `requestedAmount` isn't set (same philosophy as the existing CoW check). No existing call site's outcome changes when the caller signs for the same amount it quoted for (the correct/happy-path usage) - verified via the untouched pre-existing tests staying green.

🤖 Agent-generated

### DIFF
--- a/.changeset/vault-free-swap-amount-consistency-1117.md
+++ b/.changeset/vault-free-swap-amount-consistency-1117.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Close the residual amount-consistency gap named in #1092 (ABTS/plan 005): `prepareSwapTxFromKeys`, the vault-free (agent-reachable) swap builder, now rejects a native, EVM, or Solana swap quote whose base-unit amount doesn't match the amount the caller is signing for. `findSwapQuote` now stamps every quote with `requestedAmount`, the only reliable amount-consistency signal available for routes whose response carries no committed-input-amount field of its own.
--- a/packages/core/chain/swap/quote/SwapQuote.ts
+++ b/packages/core/chain/swap/quote/SwapQuote.ts
@@ -16,4 +16,13 @@ export type SwapQuoteResult = {
 export type SwapQuote = {
   quote: SwapQuoteResult
   discounts: SwapDiscount[]
+  /**
+   * The source-chain base-unit amount `findSwapQuote` actually fetched this quote
+   * for. Optional (not every hand-built `SwapQuote` fixture sets it) so callers
+   * that have it can cross-check the amount they're about to sign against the
+   * amount the quote was computed for — the only reliable amount-consistency
+   * signal available for native/EVM/Solana routes, whose quote responses carry
+   * no committed-input-amount field of their own (ABTS/plan 005 residual).
+   */
+  requestedAmount?: bigint
 }
--- a/packages/core/chain/swap/quote/findSwapQuote.ts
+++ b/packages/core/chain/swap/quote/findSwapQuote.ts
@@ -680,6 +680,7 @@ export const findSwapQuote = async ({
         return {
           quote: { native },
           discounts: swapChain === Chain.THORChain ? [...vultDiscount, ...referralDiscount] : vultDiscount,
+          requestedAmount: amount,
         }
       },
     }))
@@ -726,7 +727,7 @@ export const findSwapQuote = async ({
             affiliateBps,
           })
 
-          return { quote: { general }, discounts: vultDiscount }
+          return { quote: { general }, discounts: vultDiscount, requestedAmount: chainAmount }
         },
       })
     }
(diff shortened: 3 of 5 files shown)


gomesalexandre:
Still red on the nanoid audit (GHSA-2v37-7h3g-55p8), and unlike the sibling PRs this one cannot be cleared by a plain `git merge main`.

#1929 landed `nanoid 3.3.18` on `main` overnight, so I merged it into the other blocked PRs (#1722, #1716, #1674, #1664, #1662, #1897, #1805 — all green now). This one conflicts:

```
packages/core/chain/swap/quote/SwapQuote.ts
packages/sdk/src/tools/prep/swap.ts
4 hunks
```

Those are swap-quote source files rather than additive config, so the resolution is semantic and I did not want to guess at it on a fund-adjacent path. Aborted cleanly — the branch is untouched at `746fec5d`.

Needs a hand-resolve to pick up the lockfile fix.

gomesalexandre:
Superseded by #1811 (merged 2026-08-11, commit 003ee98728f3318eb2de661209f164b449150810), which landed a strictly stronger fail-closed fix for the same gap: \`BoundSwapQuote\` (\`requestedAmount\`/\`expiresAt\`/\`safetyFingerprint\` all required) + \`assertQuoteSafetyBinding\` (throws on missing binding, validates expiry, cryptographic-style fingerprint against coin-pair/amount/tx, guards structured-clone mutation) in \`packages/sdk/src/tools/prep/swap.ts\`.

This PR's own approach (optional \`requestedAmount\`, fail-open when absent) is functionally correct in isolation but weaker, and merging it now produces real conflicts with the already-landed fail-closed mechanism. Closing rather than reconciling two parallel checks for the same thing.

Issue #1072, opened 2026-07-08
[audit-SDK-ARCH-03 · DEPS-01 · DX-01/02/03 · DEPS-02] God-modules, a duplicate EVM toolchain, a fragile toolchain gate, and stale build artifa


gomesalexandre:
## Context (fund-safety audit round 1, 2026-07-08 · finding SDK-ARCH-03 · DEPS-01 · DX-01/02/03 · DEPS-02 · Low)

VaultBase.ts (1,962), Vultisig.ts (1,410), findSwapQuote.ts (828) are high-fan-in god-modules (median « 100 LOC). Both ethers and viem ship for EVM — a genuine dup; migrate to viem (the RN path's choice) and drop ethers. typecheck is a 9-stage && chain (first failure masks the rest — run per-workspace under concurrently), knip can't run without a prior build:all, and jscpd's "0 clones" is partly a permissive threshold (minLines:45) that hides the ~11-file-each per-chain resolver drift. ~52 MB of stale .tgz artifacts sit in the tree (gitignored, hygiene only).

## Effort / risk / confidence
Effort S each (arch L)Risk LOW-MEDConfidence HIGH

## Runtime repro (BEFORE + AFTER — mandatory)
SDK-CLI build + unit/CLI drive; before vs after.

## Prompt for AI agents
Implement the fix for audit finding **SDK-ARCH-03 · DEPS-01 · DX-01/02/03 · DEPS-02** in `vultisig/vultisig-sdk` at the anchors named above. VaultBase.ts (1,962), Vultisig.ts (1,410), findSwapQuote.ts (828) are high-fan-in god-modules (median « 100 LOC). Both ethers and viem ship for EVM — a genuine dup; migrate to viem (the RN path's choice) and drop ethers. typecheck is a 9-stage && chain (first failure masks the rest — run per-workspace under concurrently), knip can't run without a prior build:all, and jscpd's "0 clones" is partly a permissive threshold (minLines:45) that hides the ~11-file-each per-chain resolver drift. ~52 MB of stale .tgz artifacts sit in the tree (gitignored, hygiene only). Minimal, additive, fail-closed; do not weaken any existing guard. Add a regression test. Runtime-verify before/after (SDK-CLI build + unit/CLI drive; before vs after.). **Grep open PRs in this repo first** — advance an existing PR (adding `closes #<this issue>`) rather than duplicate. PR body MUST include `closes #<this issue>`.

---
[...]

PR #1863, opened 2026-08-11
chore(dx): run typecheck workspaces concurrently instead of a masking && chain

gomesalexandre:
tackles vultisig/vultisig-sdk#1072 (DX-01 slice — the typecheck chain masking issue only, not the god-modules split, ethers->viem migration, knip's build:all dependency, or the stale .tgz artifacts, each its own separate piece)

## what

`typecheck` was a 9-stage `&&` chain: toolchain check, root `tsc`, then 7 workspace typechecks. The first failure short-circuited every stage after it, so a broken workspace's errors were invisible until whatever failed earlier in the chain got fixed first.

## fix

Root `tsc` + 6 pure-typecheck workspaces (client-shared, cli, mcp, sdk, rujira, example-electron) now run via `concurrently --kill-others-on-fail=false` (already a devDependency, as the issue suggested) - all run to completion, all failures print independently with file/line, overall exit is still non-zero if any failed.

`example-browser` stays sequential-after the concurrent group rather than folded in. Its own typecheck script runs `prepare:sdk` first (rebuilds local `@vultisig/sdk` artifacts into its own `dist`), which raced with the concurrent group when I tried including it directly - `rm: dist: Directory not empty` - and passed cleanly every time run in isolation. Documenting this honestly: it's a real, narrower version of the same masking gap (if any of the 7 concurrent stages fails, example-browser's own typecheck still won't run that pass), not a full fix. Properly closing it means splitting example-browser's prepare/typecheck steps so prepare can run before the concurrent block without contention - a distinct, slightly bigger change I didn't want to bundle into this one-line DX fix.

## testing

Verified live, not just by inspection:
- Broke `client-shared` and `rujira` simultaneously. Under the OLD chain only the first failure (client-shared) would ever surface. Under the new script both errors print in the same run, plus confirmed `cli`/`mcp`/`sdk`/`example-electron` still ran and reported clean despite the two failures - proving the masking is actually gone for these 7.
- Reverted the breakage, confirmed a full clean pass (exit 0) with the build prerequisites (`yarn build:shared && yarn build:all`, matching what CI does before typecheck).
- Confirmed the example-browser race is real: fails when run concurrently with the others, passes every time in isolation.
- Whole run is also faster in practice (bounded by the slowest single stage, not the sum) - a side benefit, not the point.

## risk

Low - one line, additive tooling change, doesn't touch any product code. The one residual gap (example-browser) is narrower than the status quo and explicitly documented, not silently traded away.

## Summary by CodeRabbit

* **Chores**
  * Improved the project’s type-checking workflow by running checks across workspaces in parallel.
  * Added browser type validation to the overall verification process.
  * Checks now report individual failures while allowing related validations to finish, providing more complete feedback from a single run.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because The concurrent typecheck change is sound, but this branch is substantially behind main and changes only one package script. Recreating the small change from current main is cheaper and safer than carrying this stale head forward.

PR #1866, opened 2026-08-11
fix(cli): interactive chat renders the terminal message, not the delta buffer

neavra:
The interactive agent chat (`vultisig agent`) discards the authoritative terminal message whenever streaming deltas were collected: `onAssistantMessage` renders its accumulated delta buffer (`this.currentStreamText`) instead of the resolved `content`, even though `session.ts` already resolves the display text correctly ("final message event wins over streamed deltas").

On a guardrail-blocked turn the buffer is `<pre-block model prose><refusal>` while `content` is the refusal alone — so the interactive user sees the unverified prose the backend deliberately replaced. Runtime-observed:

```
[22:26:41] Agent: Here are your balances:
- Cosmos: 42 ATOM
I can't complete that safely — I could not verify the balances I was about to show. Retrying usually resolves this.
```

The headless path (`agent ask`) already prefers the terminal message, which is why it renders the refusal alone for the same turn.

## Fix

Render `content` when the turn produced one, falling back to the delta buffer only when it's empty:

```ts
process.stdout.write(renderMarkdown(content || this.currentStreamText) + '\n')
```

The `onDone` flush is deliberately unchanged — it handles turns that end with no terminal message at all, where the buffer is the only text there is.

## Tests

Three regressions in `tui.test.ts`: the blocked shape (deltas carry fabricated prose, content is refusal-only → output contains the refusal and none of the prose), matching content renders exactly once (no double print on normal turns), and empty content falls back to the buffer. The blocked-shape test fails against the unfixed renderer (revert-verified).

Changeset included (`@vultisig/cli` patch).

## Behavior note (ordinary turns)

On tool-interleaved turns (text → tool → text), the pre-tool prose now prints after the tool lines under the final `Agent:` label, because the terminal message carries the whole turn while the old buffer held only the last segment. Nothing is duplicated — previously that prose was silently dropped (the buffer is abandoned when a tool call starts), so this is strictly recovered output, in a different position.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal assistant message rendering to display the complete authoritative response when available.
  * Added a fallback to streamed text when no final response content is provided.

* **Tests**
  * Added coverage for final-content rendering, duplicate prevention, and streamed-text fallback.

### DIFF
--- a/.changeset/cli-terminal-message.md
+++ b/.changeset/cli-terminal-message.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/cli': patch
+---
+
+Render the authoritative terminal assistant message in interactive chat when it replaces streamed text.
--- a/clients/cli/src/agent/tui.test.ts
+++ b/clients/cli/src/agent/tui.test.ts
@@ -1,0 +1,58 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { ChatTUI } from './tui'
+
+afterEach(() => {
+  vi.restoreAllMocks()
+})
+
+describe('ChatTUI terminal assistant messages', () => {
+  function captureTurn(deltas: string[], content: string) {
+    const writes: string[] = []
+    const logs: string[] = []
+    vi.spyOn(process.stdout, 'write').mockImplementation(chunk => {
+      writes.push(String(chunk))
+      return true
+    })
+    vi.spyOn(console, 'log').mockImplementation((...args: unknown[]) => {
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
looks good.

receipts:
- `vitest run clients/cli/src/agent/tui.test.ts` in an exact-head checkout passed: 3 tests
- `git diff --check origin/main...HEAD` clean
- prettier check on the touched files clean

nice catch making `onAssistantMessage` prefer the authoritative terminal payload while keeping the empty-content fallback.

PR #1871, opened 2026-08-11
chore(rujira): inert expired merge balance service

rcoderdev:
## Summary by CodeRabbit

- **Bug Fixes**
  - Expired KUJI-to-RUJI merge-balance lookups now return no results instead of attempting unavailable queries.
  - Prevents outdated mergeable-token balances from being displayed.

- **Tests**
  - Added coverage confirming empty results after the merge window closed.

- **Chores**
  - Marked the expired merge-balance service as inactive.

### DIFF
--- a/.changeset/remove-expired-rujira-merge.md
+++ b/.changeset/remove-expired-rujira-merge.md
@@ -1,5 +1,6 @@
 ---
 '@vultisig/core-chain': patch
+'@vultisig/sdk': patch
 ---
 
 Make the expired, zero-caller Rujira merge-balance service inert. The KUJI-to-RUJI merge window closed on 2026-04-05, so the compatibility shim now returns an empty result without querying GraphQL.
--- a/.changeset/remove-expired-rujira-merge.md
+++ b/.changeset/remove-expired-rujira-merge.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Make the expired, zero-caller Rujira merge-balance service inert. The KUJI-to-RUJI merge window closed on 2026-04-05, so the compatibility shim now returns an empty result without querying GraphQL.
--- a/packages/core/chain/chains/thorchain/ruji/services/fetchMergeableTokenBalances.test.ts
+++ b/packages/core/chain/chains/thorchain/ruji/services/fetchMergeableTokenBalances.test.ts
@@ -1,0 +1,9 @@
+import { describe, expect, it } from 'vitest'
+
+import { fetchMergeableTokenBalances } from './fetchMergeableTokenBalances'
+
+describe('fetchMergeableTokenBalances', () => {
+  it('stays inert after the KUJI-to-RUJI merge window closed', async () => {
+    await expect(fetchMergeableTokenBalances('thor1unused')).resolves.toEqual([])
+  })
+})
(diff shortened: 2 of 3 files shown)


gomesalexandre (commented):
Current head looks clean to me conceptually.

What I verified on `ee0a24a180c295013b1304b0e1c15856a1aa260b`:
- `fetchMergeableTokenBalances.ts` keeps the exact exported async signature and now resolves to `[]` without touching GraphQL.
- the new regression test pins that inert behavior.
- the service really is zero-call inside this repo now, so this is a compatibility shim rather than a half-removed internal dependency.
- the changeset fix is present and matches the bundled-SDK guard expectation.

I did not find a correctness or safety problem in the code on this head.

Process note only: this branch contains a substantive `gomesalexandre` commit on the current head, so I am not approving it. It needs a fresh non-gomes current-head approval before merge.

gomesalexandre (commented):
Current head `9b580471372ee9c4e4e07dd89b585085b11a49c1` looks conceptually clean from a remote-only pass.

What I re-grounded:
- `fetchMergeableTokenBalances` keeps the exported async signature and now resolves to `[]` as an inert compatibility shim after the KUJI -> RUJI merge window closed.
- the focused regression test pins that inert behavior.
- the follow-up changeset commit closes the bundled-sdk publication gap on the current head.
- current head CI is green.

I do not see a fresh correctness or fund-safety blocker on this head.

Process note only: this head contains a substantive `gomesalexandre` commit (`9b580471372ee9c4e4e07dd89b585085b11a49c1`), so four-eyes forbids my approval. This needs a fresh non-gomes current-head approval before merge.

gomesalexandre (approved):
Safe removal, and the shape is right: keeping `fetchMergeableTokenBalances` exported as an inert `async () => []` rather than deleting it means the published subpath keeps resolving, so this stays a patch instead of a breaking change for anyone who imported it.

The thing I wanted to rule out before approving is the usual hazard with "this is expired, return nothing": a service that starts reporting empty can hide funds a user still holds. Checked the blast radius rather than assuming, and it is nil:

```
in-repo consumers                          none
vultiagent-app / mcp-ts / agent-backend-ts / mcp   no references to fetchMergeableTokenBalances
any 'mergeable' surface in those repos     none
```

So nothing in the ecosystem renders this today, and the only reachable path is a direct SDK import by an external consumer - which is exactly why the shim is the correct call over deletion.

The `@deprecated` line carrying the actual date (merge window closed 2026-04-05) rather than a vague "no longer supported" is worth keeping; it lets the next person judge staleness without going and asking someone.

```
CI   all 5 checks green
```

Small note, take it or leave it: the test asserts the inert behaviour but the deprecation date is the load-bearing fact, and nothing pins it. If the merge window ever reopens (or someone misremembers the date), the test still passes while the shim is now wrong. Not worth a mechanism, just worth knowing the test is guarding the shape rather than the premise.

rcoderdev:
Rebased onto the latest main at `49823bf17e5ead6d321a41d29923142156a190e4`. The two feature commits remain patch-equivalent in the range-diff, `yarn check:ci` and the focused merge-balance regression test pass, and the built function still returns `[]` without issuing a request. A fresh non-author review remains requested for the current head.

PR #1872, opened 2026-08-11
fix(sdk): publish balance tool subpath

rcoderdev:
## Summary by CodeRabbit

- **New Features**
  - Added the `@vultisig/sdk/tools/balance` package subpath.
  - Exposed platform-specific JavaScript bundles and TypeScript declarations.
  - Added balance utilities for XRP and UTXO assets.

- **Bug Fixes**
  - Improved reliability of the CLI help output check by allowing additional time to complete.

- **Tests**
  - Added coverage for balance subpath imports, runtime behavior, and type usage.


gomesalexandre (commented):
Current head looks good.

What I verified on `3f5289b0be5b89384458302d9ab9c1e9d4268c04`:
- the new `@vultisig/sdk/tools/balance` export map entry matches the existing `parse` / `defi` / `bridge` subpath pattern exactly across `browser`, `worker`, `react-native`, `node`, `import`, `require`, and `default`.
- rollup wiring in both platform and types configs is consistent with those existing subpaths.
- the current-head follow-up in `scripts/smoke-sdk-subpath-exports.mjs` closes the prior CodeRabbit nits properly by actually `require()`-loading the CommonJS bundle and by using explicit typed values / function signatures for the compile-only checks.
- I did not find a public-API regression here. This adds a subpath; it does not weaken the existing ones.

So from a code-quality standpoint this head is in good shape.

Process note only: this branch also contains a substantive `gomesalexandre` commit on the current head, so I am not approving it. It needs a fresh non-gomes current-head approval before merge.

gomesalexandre (approved):
Same shape as #1898 and the same thing to like about it: the smoke script validates the **packaged artifact from a consumer's position** rather than trusting the source. It `require.resolve`s the real entry and pins it to `dist/tools/balance/index.cjs`, exercises both the ESM `import` and the CJS `require`, and then type-checks a synthesized consumer against the emitted `.d.ts`:

```
const balanceReader: (address: string) => Promise<XrpBalance> = getXrpBalance
const utxoReader: (chain: UtxoChain, address: string, options?: GetUtxoBalanceOptions) => Promise<UtxoBalance> = ...
```

That last part is the one people skip. An exports map can resolve at runtime while the types entry points at nothing useful, and a consumer only finds out after installing.

```
CI   all 5 checks green
```

## q: the timeout bump

```diff
-  })
+  }, 10_000)
```

`clients/cli/src/lib/stderrCleanliness.test.ts` gets a 10s timeout in a PR that otherwise only adds an exports entry. It is plausibly just CI load, but a timeout bump is also the standard way a real slowdown gets papered over, and adding two more rollup outputs is exactly the kind of thing that could make a `spawnSync` of the CLI slower. One line in the description saying which it is would save the next person the archaeology.

Not blocking, and to be clear I checked before raising it: that suite fails identically on `origin/main` in my checkout (`expected 1 to be +0` on "leaves stderr completely empty", from the `bigint: Failed to load bindings` warning a native-module-less install emits). So the failure is environmental and pre-existing, not something this PR introduced - I am only asking about the timeout, not the red.

PR #1873, opened 2026-08-11
docs(rujira): fix stale CLI memo examples

rcoderdev:
Correct the Rujira CLI reference examples to match the secured-asset memo builders (`secure+` for deposits and `secure-` for withdrawals). The Rujira GraphQL clients and public core-chain subpaths remain intact because the current Windows app consumes both; packed-package contracts now guard those exports.


gomesalexandre (changes requested):
Blocking on the current head `d397cb579acb9a7e50e3383cd30171b6b7f7f791`.

The GraphQL clients being removed here are not dead org-wide. They are still consumed by `vultisig-windows` right now.

Grounding:
- `vultisig-windows/core/ui/vault/deposit/hooks/useMergeableTokenBalancesQuery.ts` imports `@vultisig/core-chain/chains/thorchain/ruji/services/fetchMergeableTokenBalances` directly.
- `vultisig-windows/core/ui/vault/deposit/hooks/useRujiraStakeQuery.ts` imports `@vultisig/core-chain/chains/thorchain/ruji/services/fetchStakeView` directly.
- Those hooks are not archaeological leftovers. They are live UI query hooks, so removing the exports here is a real public-API break for another repo in the org, not just internal cleanup.

I agree with the doc/runtime memo fixes in this PR, and the major bump on `@vultisig/core-chain` at least signals breaking intent. But the framing still matters: this is not "remove dead GraphQL clients" unless the windows consumer is being retired in lockstep.

Please either:
1. coordinate the companion removal / replacement in `vultisig-windows` and link that context here, or
2. drop the service-file and export removals from this PR and keep only the genuinely-safe cleanup.

As written, I don't think this head should ship as a dead-code cleanup.

gomesalexandre (approved):
Re-grounded current head. The dangerous part of the earlier diff is gone: the service deletions that would have broken `vultisig-windows` were fully rolled back, and the remaining doc fix now adds a useful guard (`quality-contracts.mjs` plus the Windows core-chain export test) so this class of cross-repo export regression gets caught in CI next time.

Good to ship.

PR #1874, opened 2026-08-11
fix(rujira): fail closed on unresolved order pair config

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved order handling for canonical base and quote asset formats.
  * Buy orders now use the configured quote asset rather than an unrelated default.
  * Invalid, incomplete, or unverifiable asset configurations are safely rejected with a clear invalid-asset error.
  * FIN orders are closed when the offer asset cannot be verified.
  * Improved compatibility with legacy fee configuration formats.
  * Enhanced support for valid denomination configurations when creating orders.


gomesalexandre (dismissed):
Re-grounded current head. This fixes a real fail-open path: buy-side order construction no longer falls back to `THOR.RUNE` when pair config resolution fails, and the parser now accepts the array-shaped `denoms` response while still failing closed on malformed / missing entries.

The tests cover the important cases: unresolved config, missing quote denom, malformed denoms, and unresolvable asset ids. Good to ship.

gomesalexandre (commented):
Current head `9deaab6269ef5e03793a3b17057f6eae02404007` looks sound from a remote re-check.

What I re-grounded:
- `packages/rujira/src/modules/orderbook.ts` now fails closed when pair config resolution cannot produce a trustworthy base/quote mapping instead of drifting into an implicit `THOR.RUNE` fallback.
- the config parser accepts the array-shaped `denoms` response and still refuses malformed / incomplete responses cleanly.
- the focused `orderbook.test.ts` additions cover the important failure modes around missing quote denom, malformed denoms, and unresolvable assets.
- current head CI is green.

I do not see a fresh correctness or fund-safety blocker on this head.

Process note only: this branch contains a `gomesalexandre` commit in its history, so four-eyes forbids my approval here. It needs a fresh non-gomes current-head approval before merge.

gomesalexandre (changes requested):
Current head `a95286d2d8b4dc26af90289f10be224283d444fb` still has a fail-open path in `packages/rujira/src/modules/orderbook.ts`.

`getOfferAsset()` now verifies the requested market against on-chain FIN config only for the string pair path:

```ts
if (typeof params.pair === 'string' && !params.pair.startsWith('thor1')) {
  ... compare config.base/config.quote to requested pair ...
}
```

But the object-form path returns earlier:

```ts
if (typeof params.pair !== 'string' && params.pair.base && params.pair.quote) {
  const assetId = params.side === 'buy' ? params.pair.quote : params.pair.base
  return getAssetInfo(assetId)
}
```

So a caller can still pass a mismatched `{ base, quote, contractAddress }` object, have `buildPlaceOrder()` / `placeOrder()` resolve and sign against the supplied `contractAddress`, and bypass the new canonical-config check entirely. That violates the PR's stated fail-closed goal: the order can still be built for the wrong FIN market as long as the object carries plausible asset ids.

Please apply the same contract-config verification to the supported object-form pair input, and add a regression that proves a mismatched object `{ base, quote, contractAddress }` is refused.

Receipts I re-ran on this head:
- focused exact-head test: `./node_modules/.bin/vitest run packages/rujira/src/__tests__/orderbook.test.ts` -> 23/23 pass
- exact-head typecheck: `./node_modules/.bin/tsc --noEmit -p packages/rujira/tsconfig.json` -> clean

Because this is order-construction behavior on a fund-moving path, it still needs a real backend/runtime receipt after the fix, not just unit coverage.

gomesalexandre (approved):
Looks good on current head `53191a147ef55e2cfa2496254f0327baaf692e4b`.

What I re-grounded this pass:
- `getOfferAsset()` no longer falls back to `THOR.RUNE` on buy-side config drift. It now fails closed with `INVALID_ASSET` instead.
- the object-form `{ base, quote, contractAddress }` pair path is now validated against the live FIN contract config too, so the previous object-path bypass is closed.
- the new exact-head tests cover the important fail-closed cases, including mismatched object pairs and proving `placeOrder()` never reaches `executeContract` on that bad input.
- current head CI is green.

I do not see a remaining correctness or fund-safety blocker on this head.

PR #1875, opened 2026-08-11
fix(cli): validate ERC-20 consent recipient

neavra:
## Summary

- decode ERC-20 `transfer` calldata and make its destination authoritative in the signing consent summary
- fail closed and clear the buffered transaction when `txArgs.to` disagrees with calldata or transfer calldata is malformed
- preserve native sends, non-EVM envelopes, and non-transfer contract-call summaries
- add a patch changeset for `@vultisig/cli`

## Verification

- `yarn test:cli --maxWorkers=2` — 77 files, 1038 tests passed
- `yarn lint`
- `yarn typecheck`
- `yarn quality:complexity`
- `yarn format:check`

## Summary by CodeRabbit

* **New Features**
  * Added ERC-20 transfer recipient and amount validation before transaction signing.
  * Signing summaries now use verified transaction data for recipients, token contracts, and transfer amounts.
  * Unknown token decimals are clearly marked as unverified while preserving raw amounts.
* **Bug Fixes**
  * Prevented signing for malformed calldata or mismatched recipient and amount details.
  * Preserved handling for native transfers, non-EVM transactions, and non-transfer actions.
  * Ensured valid token transfers request confirmation only once.


gomesalexandre (dismissed):
Material WYSIWYS gap still open on current head `05e6b145`.

`clients/cli/src/agent/executor.ts:57-63` decodes ERC-20 `transfer(address,uint256)` calldata as `(recipient, value)` but only keeps the recipient. The new check in `assertConsistentTransferRecipient()` (`:402-423`) then validates recipient-only, while `getPendingSummary()` still renders the amount from producer-controlled labels / `txArgs.amount` (`:479-505`) and `signEvmServerTx()` signs the untouched calldata from `extractNestedTx(...)` (`:1099`, `:1149-1158`).

That means an envelope can say `0.5 USDC` in the confirmation summary while the signed calldata transfers `500 USDC` to the same recipient, pass the new guard, and obtain misleading consent on the exact money moved.

Please decode and cross-check the transferred amount from the same nested tx the signer consumes, then add regressions for amount drift on both the top-level `tx` path and the mixed `tx` + benign `txArgs.tx` path, mirroring the recipient-mismatch cases you already added.

gomesalexandre (commented):
current head still has a wysiwys gap in `clients/cli/src/agent/executor.ts`.

- `assertConsistentTransfer()` now cross-checks calldata recipient and amount against `txArgs.to` / `txArgs.amount`, which is good.
- but `getPendingSummary()` still renders the displayed amount from producer-controlled labels (`resolved.labels.resolved_amount` / `labels.amount`) via `amountWithSymbol`, not from the decoded calldata amount from the signed tx path.
- so an envelope can keep `txArgs.amount` aligned with calldata, set `resolved_amount` to a smaller human string like `0.5 USDC`, pass the new guard, and still show misleading consent for the exact money moved.

please either render the consent amount from the decoded transfer calldata (same path used for the recipient check), or fail closed when the displayed amount text would disagree with that decoded amount. add a regression for the matched-`txArgs.amount` / lying-`resolved_amount` case too.

not approving anyway because current head contains gomes-authored commits and needs a non-gomes current-head stamp once this is fixed.

gomesalexandre (commented):
Current head `3be7aa3d` looks sound from a remote re-check.

- `assertConsistentTransfer()` still cross-checks the ERC-20 recipient and amount against the decoded calldata from the same `extractNestedTx(...)` path the signer consumes.
- `getPendingSummary()` now renders the ERC-20 amount from that decoded signed calldata amount, not from producer-controlled labels like `resolved_amount`.
- known tokens use trusted token metadata for display, and unknown tokens fail closed to raw base units with an explicit `decimals unverified` marker instead of a fake precise decimal amount.
- the new `executor.amountSummary.test.ts` cases cover the exact lying-label path that was previously open: matched `txArgs.amount` plus understated `resolved_amount`, while calldata carries the larger transfer amount.

I do not see a fresh WYSIWYS blocker on the current diff from this remote-only pass.

Process note only: this head contains substantive `gomesalexandre` commits, so four-eyes forbids my approval here. It needs a non-gomes current-head approval before merge.

Separate mergeability note: GitHub currently shows this PR `DIRTY` against `main`, so it also needs conflict resolution before merge.

gomesalexandre (commented):
Current head `b4c6b6e6` looks materially safe on the ERC-20 consent bug that was under review, and the merge-from-main itself is clean, but I found one fail-closed follow-up gap worth fixing before I’d call this done.

What I re-grounded on the current head:
- the merge commit only pulls in unrelated `main` churn and leaves the ERC-20 consent codepath byte-identical vs pre-merge `3be7aa3d`
- CI is green on this head
- recipient + amount are still decoded from the same `extractNestedTx(...)` path the signer consumes
- the displayed ERC-20 amount still comes from decoded calldata, not producer-controlled labels

New finding on the current head:
- `assertConsistentTransfer()` now always requires `txArgs.amount`, but the flat `__buildTx` / tool-output signing envelope shape only carries `tx` / `txArgs.tx` and does not populate producer `txArgs.amount` / `txArgs.to`
- in a disposable local repro, a well-formed single-leg payload with `__buildTx: true` plus a plain ERC-20 `transfer(address,uint256)` calldata leg throws `Invalid ERC-20 txArgs.amount - refusing to sign` out of `getPendingSummary()` immediately, because there is no producer amount to cross-check against
- recipient already has the right shape guard (`if (producerRecipient && ...)`), but amount does not have an equivalent presence gate

Why I’m leaving this as a should-fix, not a fund-safety blocker:
- this fails closed, so it refuses signing rather than allowing misleading consent or bad fund movement
- I have not proven a currently wired tool in this PR hits the flat-transfer shape today
- but it is a real coverage gap in the current logic and can strand any present or future flat tool whose signable leg is a plain ERC-20 transfer

Suggested direction:
- mirror the recipient check and only enforce the amount equality when the producer-side amount field is actually present, or explicitly special-case the flat `__buildTx` shape
- add a regression covering a valid flat-tool ERC-20 transfer envelope with no producer amount field, so the guard stays fail-closed on drift but does not reject shape-valid envelopes purely for lacking producer metadata

Process note only: this head contains substantive `gomesalexandre` commits, so four-eyes still means it needs a non-gomes current-head approval before merge.

gomesalexandre (commented):
Current head `8c597456` closes the specific fail-closed gap I called out on `b4c6b6e6`.

- `clients/cli/src/agent/executor.ts` now gates the ERC-20 amount cross-check on producer-side presence (`producerAmount && ...`), mirroring the recipient presence gate, so a flat `__buildTx` / tool-output envelope with no `txArgs.amount` no longer gets rejected just for lacking producer metadata.
- the new `clients/cli/src/agent/__tests__/executor.recipientSummary.test.ts` case pins exactly that shape: a flat `buildTxReadyFromToolOutput(...)` ERC-20 transfer with no `txArgs`, and it asserts `getPendingSummary()` does not throw while the pending tx stays buffered.
- the existing recipient/amount mismatch paths still fail closed when producer metadata is present and disagrees with the signed calldata.
- `getPendingSummary()` still renders the ERC-20 amount from decoded signed calldata, not producer-controlled labels.

So the current diff looks materially correct on the consent issue under review.

Not stamping for process reasons:
- this head contains substantive `gomesalexandre` commits, so four-eyes forbids my approval here
- current CI is not fully green on this head: `Lint and Type Check` is red right now, so it is not merge-ready anyway

Once the red check is resolved, this needs a fresh non-gomes current-head approval.

gomesalexandre (approved):
This is the right invariant and the implementation gets the hard part right. Making the decoded calldata destination authoritative for the consent summary, rather than trusting the producer's declared `to`, is what "what you see is what you sign" actually requires.

The detail I want to call out approvingly, because it is the one that would have been easy to get wrong: reading the signed tx through `extractNestedTx` (`swap_tx || send_tx || tx || txArgs.tx`), the **same** resolution the signer uses, instead of `txArgs.tx` alone. An envelope carrying a benign `txArgs.tx` alongside a malicious higher-precedence `swap_tx` is exactly how a check like this gets bypassed, and the comment says so explicitly. Running the assertion before the branch-specific summaries, so a transfer cannot be dressed as a swap to skip it, closes the other half.

## should-fix: uppercase-hex calldata fails closed on a legitimate transfer

The selector gate is deliberately case-tolerant:

```ts
if (calldata.slice(0, ERC20_TRANSFER_SELECTOR.length).toLowerCase() !== ERC20_TRANSFER_SELECTOR) return null
```

but viem's lookup underneath is case-**sensitive**, measured directly:

```
viem lowercase            -> OK
viem UPPERCASE selector   -> THROWS: Encoded function signature "0xA9059CBB" not found on ABI.
viem ALL-UPPER calldata   -> THROWS: same
```

So mixed-case calldata passes your gate, then throws inside `decodeFunctionData`, and the catch converts that into `Invalid ERC-20 transfer calldata — refusing to sign` plus `clearPendingTransaction()`. A perfectly valid transfer gets refused and the buffered tx is dropped.

In practice every mainstream producer emits lowercase hex, so I would not call this reachable today, and it fails in the safe direction. I am raising it because the code already anticipates case variance at the gate and then does not carry that through, which reads as an oversight rather than a decision, and because "never false-block legit money movement" is the rule this repo holds itself to. A `.toLowerCase()` on the calldata before decoding makes the two halves agree.

## q: scope of the check

Non-`transfer` calldata returns `null` and falls through to the generic contract-call summary. Confirmed:

```
transferFrom (0x23b872dd)  -> null (no cross-check)
approve      (0x095ea7b3)  -> null (no cross-check)
```

That is a defensible boundary - `transferFrom` out of the vault needs a prior approval, so it is not a one-step drain - and the PR title says ERC-20 transfer. Worth confirming it is a conscious scope line rather than an oversight, since the same "producer says X, calldata says Y" divergence applies to both.

One thing I checked and am happy about: trailing bytes appended to valid transfer calldata still decode to the correct recipient and amount rather than throwing. That matches what the EVM itself does with the extra data, so the summary keeps showing the truth instead of false-blocking on padding.

## your CI red is not yours

`Lint and Type Check` fails on the nanoid advisory (GHSA-2v37-7h3g-55p8, `<3.3.18`), not on anything in this diff - Prettier reports "All matched files use Prettier code style!" in the same job. #1929's lockfile bump fixes it; I flagged it there as a merge-priority item since it is blocking a stack of PRs including this one.

Also worth saying: my first local run of the cli suite showed 36 failed files, which was stale workspace `dist` from branch-hopping on my side, not your change. Your CI's Unit Tests are green. Nothing for you there.

gomesalexandre:
@neavra heads up — this is red purely on the nanoid audit (GHSA-2v37-7h3g-55p8, `<3.3.18`), nothing in your diff.

#1929 landed `nanoid 3.3.18` on `main` overnight, so merging `main` in should clear it. I did that for my own blocked PRs and they all went from 2-fail to fully green.

Not pushing to your branch — flagging so you know the fix is a one-command merge rather than anything in the ERC-20 consent work, which still looks good to me.

Issue #1877, opened 2026-08-12
fix(balance): preserve exact TRX SUN balances beyond Number.MAX_SAFE_INTEGER


gomesalexandre:
## Summary
`getTrxBalance` silently rounds whale balances because it parses TRON SUN as a JavaScript number.

## Evidence
- `packages/sdk/src/tools/balance/otherBalance.ts:149-173` previously read `balance` as `number` and formatted from that lossy value.
- `packages/sdk/src/tools/balance/rpc.ts` only exposed JSON parsing, so the exact raw integer never survived for large balances.
- the SDK already protects XRP and Solana reads from this same class of precision loss.

## Impact
TRX balances above `2^53` SUN can be under/over-reported before formatting, which is deterministic correctness drift in the SDK-owned balance layer.

## Fix sketch
Preserve the raw TRON response text, parse the exact SUN integer, surface an exact `balanceSunRaw`, and keep any numeric mirror explicitly best-effort.

## Verification
- focused vitest for a `> 2^53` SUN fixture and missing-balance zero case
- live runtime receipt from `getTrxBalance('TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t')`

## Report
- round report: https://gist.github.com/gomesalexandre/8a923aae943dc3c67f3b35971b49f590

PR #1878, opened 2026-08-12
fix(balance): preserve exact TRX SUN balances

gomesalexandre:
## Summary
- preserve the exact TRON SUN integer by parsing the raw RPC response text
- expose `balanceSunRaw` as the canonical exact field while keeping `balanceSun` as a best-effort compatibility mirror
- pin the >2^53 precision edge and missing-balance zero case in focused unit coverage

## Issues
- https://github.com/vultisig/vultisig-sdk/issues/1877

## Verification
- `corepack yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/tools/balance/otherBalance.test.ts`
- live runtime receipt:
  - `{"balanceSun":1083880845444,"balanceSunRaw":"1083880845444","balanceTrx":"1083880.845444"}`

## Notes
- this draft borrowed existing `dist/` outputs from the main checkout into the disposable worktree to satisfy the repo's current test/runtime layout without mutating the source diff

## Report
- https://gist.github.com/gomesalexandre/8a923aae943dc3c67f3b35971b49f590


gomesalexandre (commented):
blocking: the TRX precision fix now accepts malformed 200 bodies as plausible balances instead of failing loudly.

what changed
- `packages/sdk/src/tools/balance/otherBalance.ts` now reads `/wallet/getaccount` via `fetchText(...)` and extracts `"balance": <digits>` with `raw.match(...)`.
- that preserves >2^53 SUN precision, which is the right goal.

blocker
- the old path went through `response.json()`, so a truncated / syntactically invalid body threw.
- the new path never validates that the body is well-formed JSON before trusting the regex match.
- result: a partial 200 response can now surface a silently wrong balance instead of an error.

exact-head receipt on `5eab0820bc1114ac61639bc8fcd3a75bb95cf184`
- focused suite still passes:
  - `cd packages/sdk && ../../node_modules/.bin/vitest run --config tests/unit/vitest.config.ts tests/unit/tools/balance/otherBalance.test.ts`
  - `29 passed`
- direct repro against `getTrxBalance(...)` with a truncated body:
  - mocked fetch returned `{"balance":125`
  - actual result was:
    - `{"balanceSunRaw":"125","balanceTrx":"0.000125", ...}`
  - expected behavior here is to throw, not silently accept the truncated prefix as the account balance.

why this matters
- this is a balance-read correctness path. a garbled upstream body should fail closed, not render a plausible but wrong user balance.

suggested direction
- keep the raw-text precision recovery, but also require the response body to be valid JSON before accepting the extracted digits. for example, validate with `JSON.parse(raw)` first, then derive the exact balance string from the raw body or a bigint-safe parser.

NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because PR #2049 now carries the same TRX precision fix while validating JSON before extracting raw SUN digits and avoiding an unsafe numeric mirror above 2^53. Closing this duplicate in favor of the safer current implementation.

PR #1879, opened 2026-08-12
fix(mpc): align Cardano and Terra mobile fixture hashes

### LINKED ISSUE #1519: terraclassic: signed fee includes the uusd burn tax but the displayed fee does not (display/signed divergence)
Found during the 2026-07-23 fee display-vs-signed parity hunt. This is a **concrete, currently-live** display/signed divergence, small-dollar but real.

## What
On TerraClassic (USTC), the signed transaction's fee includes a SECOND `Amount` - the dynamically-computed `uusd` burn/stability tax - on top of the base `uluna` gas fee. The displayed fee never includes it.

- Signed: `packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts:409-423` - the `ibcEnabled` branch pushes an extra `Amount` (denom `uusd`, the burn tax) when the live on-chain stability-tax rate is non-zero.
- Displayed: `packages/core/mpc/keysign/fee/resolvers/cosmos.ts:16-22` - `getCosmosFeeAmount` only ever returns the base `gas` value, never adds the burn tax.
- Tax computation: `packages/core/mpc/keysign/chainSpecific/resolvers/cosmos.ts:30-63`.

## Failure scenario
A USTC send with a non-zero live tax rate signs a tx whose total fee (uluna gas + uusd burn tax) is higher than the card showed (uluna gas only). Bounded by `getTerraClassicTaxCap` (historically cents to a few dollars), so this is not a large loss - but it is a real WYSIWYS divergence where the user pays more than displayed.

## Fix direction
Make the display resolver include the burn-tax amount whenever the signing path would add it, so displayed fee total == signed fee total. Both resolvers already live in `packages/core`, so they can share the computation rather than duplicating it.

**SDK caution**: flagship ios/android/windows consume `packages/core`, so this changes a displayed fee for all of them - in the correct direction (more accurate), but it warrants a consumer check before release.

### LINKED ISSUE #1584: Two pinned hash overrides in mobileFixtures golden test are live cross-platform signing divergences (Cardano max-amount, Terra USTC)
## Summary

`packages/core/mpc/keysign/tests/mobileFixtures.golden.test.ts` pins two `sdkExpectedHashOverrides` where the TS implementation **knowingly produces a different pre-image hash than the mobile fixture corpus**:

- `cardano.json::Send ADA - max amount`
- `terra.json::Send USTC on terra classic`

These overrides make the TS suite green, but they are not test debt — they are **documented live cross-platform signing divergences**. If a user co-signs one of these exact transactions between a mobile device and the extension/desktop today, the keysign should fail.

## Why a hash mismatch fails in the field

In `packages/core/mpc/message/setup/ensure.ts`, the joining device polls the relay with `messageId = getMessageHash(message)` computed from the payload it rebuilds **locally**. If the TS hash differs from the mobile initiator's hash, the joiner polls a `message_id` the initiator never uploaded and dies with `404 "Timed out while waiting for setup message"` — precisely the failure signature in vultisig/vultisig-windows#4464 (THORChain TCY → RUNE, Android + extension, this week).

## Ask

Adjudicate each override against the chain's actual signing rules and converge the implementations so the overrides can be deleted:

1. **Cardano max-amount:** the override comment attributes the difference to fee handling (`byte_fee: 44` fixture value vs the TS fee path). Determine which fee construction is correct per Cardano ledger rules; fix the diverging side; regenerate the shared expected hash.
2. **Terra USTC:** the override comment points at tax/IBC-denom handling under the current rate=0 regime. Same treatment: pick the correct construction, align, regenerate.
3. Until converged, treat these two transaction shapes as known-broken for mixed mobile↔extension/desktop co-signing (worth a support-KB note if either shows up in tickets).

Acceptance: `sdkExpectedHashOverrides` is empty, and the regenerated expected hashes are confirmed by running at least two of the three real signers (Kotlin/Swift/TS) and asserting agreement — not captured from a single implementation.

rcoderdev:
Align the SDK keysign pre-images with the shared mobile fixture corpus without SDK-only expected-hash overrides.

- Cardano max-send now replays the mobile total fee.
- Terra Classic USTC relays its full fee in `uusd`, so display and signing use the same single-denomination amount.
- The mixed-denomination scalar from the first implementation has been removed.
Closes #1519
Closes #1584

## Summary by CodeRabbit

* **Bug Fixes**
  * Terra Classic USTC transaction fees now align with mobile signing behavior.
  * USTC fees use the correct denomination, including IBC transfers.
  * Full-balance USTC sends deduct in-kind fees and handle insufficient balances correctly.
  * Terra Classic burn taxes are included accurately in fee totals.
  * Cardano max-send fee handling now matches mobile signer results.
* **Tests**
  * Added regression coverage for USTC fees, max-send behavior, IBC transfers, and Cardano fixtures.

### DIFF
--- a/.changeset/calm-owls-align-mobile-hashes.md
+++ b/.changeset/calm-owls-align-mobile-hashes.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/sdk': patch
+'@vultisig/cli': patch
+---
+
+Align Cardano max-send and Terra Classic USTC signing inputs with the current Swift and Kotlin signers so the shared mobile fixture hashes pass without SDK-only overrides.
--- a/.changeset/terra-classic-fee-parity.md
+++ b/.changeset/terra-classic-fee-parity.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Include TerraClassic's signed USTC burn-tax surcharge in the displayed send-fee total by deriving both values from the same Cosmos fee-amount list.
--- a/packages/core/chain/chains/cosmos/gas.test.ts
+++ b/packages/core/chain/chains/cosmos/gas.test.ts
@@ -11,6 +11,7 @@ import {
   MAYA_SEND_FEE_BASE_UNITS,
   TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS,
   TERRA_CLASSIC_ULUNA_BASE_GAS,
+  TERRA_CLASSIC_UUSD_BASE_GAS,
 } from './gas'
 
 const jsonResponse = (body: unknown, status = 200) =>
@@ -361,5 +362,17 @@ describe('TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS', () => {
 
   it('exceeds the native-send fee constant, proving the send fee alone would under-price staking', () => {
     expect(TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS).toBeGreaterThan(TERRA_CLASSIC_ULUNA_BASE_GAS)
+  })
+})
+
+describe('TERRA_CLASSIC_UUSD_BASE_GAS', () => {
+  it('equals the TerraClassic static gas limit priced at 0.75 uusd/gas', () => {
+    const staticGasLimit = getCosmosGasLimit({ chain: Chain.TerraClassic, id: 'uusd' })
+
+    expect(TERRA_CLASSIC_UUSD_BASE_GAS).toBe((staticGasLimit * 75n) / 100n)
+  })
+
+  it('matches iOS uusdBaseGas / Android UUSD_BASE_GAS', () => {
+    expect(TERRA_CLASSIC_UUSD_BASE_GAS).toBe(225_000n)
   })
 })
--- a/packages/core/chain/chains/cosmos/gas.ts
+++ b/packages/core/chain/chains/cosmos/gas.ts
@@ -52,6 +52,16 @@ export const TERRA_CLASSIC_STAKING_ULUNA_FEE_BASE_UNITS = getFeeAmountFromGasPri
   parseDecimal('28.325')!
 )
 
+/**
+ * Terra Classic's `uusd` fee at the static 300k gas limit:
+ * `300_000 × 0.75 uusd/gas`. USTC bank-denom sends pay both gas and burn
+ * tax in `uusd`, so their base cannot reuse the native `uluna` amount.
+ *
+ * Matches iOS `TerraClassicTax.uusdBaseGas` and Android
+ * `TerraClassicTax.UUSD_BASE_GAS`.
+ */
+export const TERRA_CLASSIC_UUSD_BASE_GAS = 225_000n
+
 export const cosmosGasRecord: Record<IbcEnabledCosmosChain, bigint> = {
   [Chain.Cosmos]: COSMOS_SEND_FEE_DEFAULT,
   [Chain.Osmosis]: 9000n,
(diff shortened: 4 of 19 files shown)


gomesalexandre (changes requested):
Blocking: `packages/core/mpc/keysign/fee/resolvers/cosmos.ts:getCosmosFeeAmount` now sums raw fee amounts across different denoms into one bigint.

On TerraClassic USTC sends, the new path combines `uluna` gas with `uusd` burn tax and returns a denomination-less scalar. That loses asset identity, so the caller cannot display or value the fee correctly. `8_497_500 uluna + 120_000 uusd = 8_617_500` is not a meaningful single-chain fee amount.

Please keep the fee as per-denom amounts at the display boundary, or convert both legs into one common unit before summing. As written, the PR fixes sign/display parity only by introducing a mixed-asset total that is itself incorrect.

gomesalexandre (dismissed):
Blocking on current head `f178ffaf33e9a0edaae204fa06eb93bc7f8c337d`.

`packages/core/mpc/keysign/refine/amount.ts` still applies the TerraClassic `uusd` in-kind fee refinement to every USTC payload:

```ts
if (!isFeeCoin(coin) && !isTerraClassicUstcCoin(coin)) {
  return input.keysignPayload
}
```

That is too broad for the exact behavior this PR now implements elsewhere. On the current head, the PR explicitly preserves `uluna` as the signed fee denom for TerraClassic IBC USTC transfers, but `refineKeysignAmount(...)` will still subtract that numeric `uluna` fee from the `uusd` transfer amount just because the asset is `uusd`. So the refinement path now treats an IBC USTC send as fee-in-kind while the signing-input path prices and signs the fee out-of-kind.

That is a real cross-path mismatch:
- refine path: debits `uusd`
- signing path: requires `uluna`

Please gate the `isTerraClassicUstcCoin(coin)` exception to the plain TerraClassic bank-send path only, matching the same condition you use for the new `uusd` fee-denom behavior, and add a regression proving an IBC USTC payload keeps its original `toAmount` while still signing the `uluna` fee.

Separate issue: current-head CI is not green. `Lint and Type Check` / `Test Summary` are red on this head, so this is not mergeable as-is.

gomesalexandre (commented):
Current head looks good to me on substance.

What I re-checked this cycle:
- `packages/core/mpc/keysign/chainSpecific/resolvers/cosmos/index.ts` now prices plain TerraClassic USTC sends with `TERRA_CLASSIC_UUSD_BASE_GAS`, so the signed fee stays `uusd`-only instead of mixing `uluna` gas with `uusd` burn tax.
- `packages/core/mpc/keysign/refine/amount.ts` now gates the in-kind USTC full-balance refinement on `transactionType === TransactionType.UNSPECIFIED`, so the old IBC path no longer gets its amount incorrectly shaved by a `uusd` fee.
- `packages/core/mpc/keysign/refine/amount.test.ts` adds the exact regression I wanted for the TerraClassic USTC IBC case, and the current head is green again.

I am not approving because this branch now includes gomes-authored non-merge commits, so it needs a non-gomes stamp for four-eyes. But from a correctness standpoint my earlier blockers look resolved.

gomesalexandre:
The refine/sign cross-path mismatch from my second review is closed as of \`a23a2c8e6\`: the \`uusd\`-in-kind refinement is now scoped to plain bank sends only (mirrors \`getCosmosChainSpecific\`'s \`isPlainSend\` gate), so an IBC USTC transfer no longer gets refined against a denom the signing path never priced against. CI is green.

I can't self-approve since I have a commit in this head (per our four-eyes rule) - needs a fresh external stamp before merge.

rcoderdev:
Rebased this branch onto the current main and revalidated the requested TerraClassic fee behavior on head a9af0ab70ce1bc830c5a57086c664c85ce223bd0. The USTC in-kind refinement remains limited to plain bank sends; IBC USTC remains unrefined and signs one uluna fee coin. Built-package runtime QA returned display/sign parity for both cases: 1,225,000 uusd for a plain USTC send and 16,995,000 uluna for an IBC USTC transfer. The full yarn check:ci gate also passes locally, and there are no unresolved review threads. Thanks for the earlier review; re-requesting review against this exact head once CI reports.

johnnyluo (approved):
LGTM

PR #1880, opened 2026-08-12
fix: harden fee integrity semantics

### LINKED ISSUE #1520: fee integrity defense-in-depth: capGasLimit is a floor not a cap, thor/maya fee unbound to signed bytes, maya fee hardcoded
Three defense-in-depth findings from the 2026-07-23 EVM/cosmos fee-integrity hunt. None proven exploitable on their own; grouped because they are all "the guard is weaker or less bound than its name implies".

## 1. `capGasLimit` is a floor, not a cap (misleading name, no ceiling exists)
`packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts:59-65` - despite the name, this is `bigIntMax(rpcEstimate, thirdPartyGasLimitEstimation, minimumGasLimit)`. It raises the gas limit, never bounds it. There is no `MAX_GAS` constant anywhere in `packages/core` (`chain/tx/fee/evm/evmGasLimit.ts` is floors only; see also `chainSpecific/resolvers/evm.ts:48-56`, `keysign/swap/build.ts:134-143`).

Impact is bounded today: `thirdPartyGasLimitEstimation` (`tx.evm.gasLimit`, an untouched passthrough from a swap-aggregator quote) or a compromised RPC estimate could inflate gasLimit without bound, but EVM only charges for gas actually consumed, so realising a loss needs the attacker-controlled `to`/`data` that a compromised aggregator quote already implies. It amplifies an existing compromise rather than standing alone. Also mitigating: neither agent-backend-ts nor vultiagent-poc ever calls `setCustomRpcOverride`, so the RPC is always Vultisig's own proxy.

**Ask**: at minimum rename so it does not read as a bound that does not exist; ideally add a real ceiling.

## 2. THORChain / MayaChain displayed fee has no binding to the signed bytes
The card reads `THORChainSpecific.fee` (live `native_tx_fee_rune`) or Maya's hardcoded fallback (`fee/resolvers/cosmos.ts:16-25`; `chainSpecific/resolvers/thor.ts:18,24`; `chainSpecific/resolvers/maya.ts:9-18`). The signed `TW.Cosmos.Proto.Fee` for these two chains has `amounts: undefined` and `gas` set to an unrelated hardcoded gas-LIMIT constant (`signingInputs/resolvers/cosmos/index.ts:397-432` - `getFeeAmounts()` returns `undefined` when `chainKind !== 'ibcEnabled'`; constants at `chain/chains/cosmos/cosmosGasLimitRecord.ts:20-21`).

So there is zero cryptographic link between the displayed number and the signed transaction for THOR/Maya. **This is very likely correct-by-design** - THOR/Maya deduct the native fee inside message processing, independent of the cosmos-sdk `authInfo.fee` - and it is consistent with years of production use, so it probably does not cost users more than shown. Could not prove it from Vultisig code alone (would need thornode/mayanode Go source), so filing as a documented defense-in-depth gap rather than a bug.

## 3. Maya displayed fee is a hardcoded magic number with no freshness check
`fee/resolvers/cosmos.ts:7` - `mayaGas = 2000000000n`, never live-queried, unlike THOR's sibling `getThorNetworkInfo()`. **Currently accurate** (verified live: `mayanode.mayachain.info/mayachain/constants` returns `NativeTransactionFee: 2000000000`). But if Maya governance ever changes it, THOR self-corrects and Maya silently goes stale.

**Ask**: live-fetch it like THOR does, or add a freshness assertion.

## Confirmed safe in the same sweep, do not re-file
- Cosmos `ibcEnabled` chains (cosmoshub, osmosis, kujira, dydx, terra, noble, akash) base gas fee: display and signed read the identical `cosmosSpecific.gas` field off the same `KeysignPayload`, single computation, single write. No post-render gas-adjustment multiplier anywhere in the signing path. Structurally guaranteed.

rcoderdev:
Closes #1520

## Summary by CodeRabbit

* **New Features**
  * MayaChain native transaction fees are now resolved from live network data, with validated fallback values.
  * Zero-fee transactions and configurable network endpoints are handled correctly.
  * MayaChain is supported in Cosmos signing workflows.

* **Bug Fixes**
  * Improved handling of invalid or missing fee values.
  * Preserved correct fee semantics for THORChain and MayaChain transactions.

* **Documentation**
  * Clarified how protocol-specific fees are represented during transaction signing.


gomesalexandre (commented):
Remote-only pass on current head `4065fefdce238acfc40a00cdd7d11a183088fda2` looks sound conceptually.

What I re-grounded:
- `packages/core/mpc/keysign/fee/resolvers/cosmos.ts` fixes the real MayaChain display-fee drift by resolving `NativeTransactionFee` from MayaNode Mimir/constants instead of the stale hardcoded fallback.
- the new parser guards fail closed on malformed Maya governance data instead of coercing junk into a fee.
- `packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts` keeps THORChain and MayaChain fee coins out of signed `authInfo.fee.amounts`, which matches the existing vault-based signing contract rather than changing it.
- the `getEvmFeeQuote` rename/comment cleanup is semantic clarification only; I did not find a behavior regression there.
- current head CI is green.

I did not find a correctness blocker in the diff.

Not approving in this pass because this is still a fee-integrity change in the SDK, and that wants a real cli/sdk runtime receipt before stamp, not just a remote conceptual pass.

gomesalexandre (approved):
Went at this one as a fund-safety change rather than a refactor, since it moves a fee off a constant and onto the network. It holds up, and the EVM half is a genuinely good catch.

## the thing I tried hardest to break

Making `vaultBased` return `getMayaNativeTransactionFee()` turns a synchronous branch of `getCosmosFeeAmount` into a promise-returning one, and an un-awaited promise in a fee path is about the worst outcome available. It is fine:

```
FeeAmountResolver = Resolver<FeeAmountResolverInput, bigint | Promise<bigint>>   // already allowed it (tron.ts is async)
getFeeAmount = async (input): Promise<bigint> => resolvers[kind](input)          // async fn, so the promise is awaited
```

So the union was pre-existing and the dispatcher already awaits. Nothing downstream can end up with a `Promise` where it wanted a `bigint`.

`parseMayaNativeTransactionFee` is the right shape too. Rejecting non-integer strings by regex before `BigInt()` rather than letting `BigInt` throw its own opaque error, and separating "mimir returns -1 meaning unset" from "constants returned a negative, which is nonsense" via `negativeMeansUnset`, is exactly the distinction that gets fudged in this kind of code.

## `capGasLimit` -> `resolveGasLimit`

Worth calling out because renames usually are not: the old name asserted something the function never did. It selects the highest available floor. Anyone reading `capGasLimit` at a call site would reasonably assume there was an upper bound protecting them, and there was not. The comment you added saying a safe upper bound is chain- and call-specific, so an arbitrary global ceiling would reject valid calls, is the correct reason not to add one.

## should-fix, though I would take it as a follow-up rather than hold this

There are now two sources of truth for the MayaChain fee, and the file's own doc comment already states the rule that resolves it:

```ts
// packages/core/chain/chains/cosmos/gas.ts
/**
 * THORChain returns `undefined` because its fee must be read from live
 * `native_tx_fee_rune` network data rather than a static fallback.
 */
export const getCosmosSendFeeBaseUnits = (chain: CosmosChain): bigint | undefined => {
  if (chain === Chain.THORChain) return undefined
  if (chain === Chain.MayaChain) return MAYA_SEND_FEE_BASE_UNITS   // still static 2_000_000_000n
```

This PR establishes that Maya's fee is live, mimir-governed data, which is the same argument that already earns THORChain an `undefined`. But `getCosmosSendFeeBaseUnits(MayaChain)` keeps handing out a hardcoded 2 MAYA, and it is a **public SDK export** (`packages/sdk/src/index.ts:235`, plus the react-native entry). So an external consumer asking the SDK for Maya's send fee gets a number the SDK itself no longer treats as authoritative, and they diverge silently the moment mimir moves.

To be fair to this PR: that static value was already a guess before you touched anything. This does not introduce the wrongness, it just makes it contradictable. Returning `undefined` for MayaChain exactly as THORChain does looks like the consistent end state.

## q: what happens when MayaNode is unreachable

The `MAYA_SEND_FEE_BASE_UNITS` import is gone from the resolver, so if `/mayachain/mimir` and `/mayachain/constants` both fail, this throws and there is no local fallback. I think fail-closed is the right call for a value that gets signed, and I would not argue for signing against a stale constant. Flagging it only because the description says "validated fallback values", which reads as a local fallback when the fallback is actually the second endpoint. Worth one line in the changeset so nobody expects offline behaviour that is not there.

Also no caching that I could see, so this is one to two network round-trips per fee resolution. Fine at current call volume, just noting it in case a screen ends up resolving fees in a loop.

## testing

```
packages/core/mpc/keysign/fee/resolvers/cosmos.test.ts
packages/core/mpc/keysign/signingInputs/resolvers/cosmos.test.ts     34 passed

packages/core/mpc/keysign/ (whole tree)          51 files, 524 passed
packages/core/chain/chains/cosmos/gas.test.ts    21 passed
```

One process note so nobody repeats it: my first run of the wider tree showed `9 failed | 42 passed`, which was my node being 20.18.3, below the 20.19 line where `require()` of an ES module is allowed. It reproduced identically on `origin/main`, which is what gave it away. On node 22 the tree is fully green. Nothing here for you.

rcoderdev:
Updated the branch onto current `main` at `9fb9a3253f59f9f91b1e4a9865f93ebec06621d9`. The feature patch is unchanged: `git range-diff` matched both feature commits exactly. Fresh validation on the integrated head passed the focused fee suites (46/46) and the full `yarn check:ci` gate. The live Maya resolver receipt also remains applicable because the feature patch and its direct runtime dependencies are unchanged: the production resolver returned `2000000000`, matching the same Maya RPC's Mimir and constants values.

Agreed that the existing `gomesalexandre` approval cannot satisfy four-eyes authorization because that identity authored part of the branch. This still needs a genuinely independent current-head approval.

PR #1881, opened 2026-08-12
fix(cosmos): surface stale sequence recovery

### LINKED ISSUE #1547: Cosmos broadcast fails code 32 "account sequence mismatch" — stale sequence frozen in KeysignPayload, no recovery
User report (Discord): signing a Cosmos-SDK-chain transaction (main signer = Windows client, co-signer = iOS app) failed at broadcast:

> Broadcasting transaction failed with code 32 (codespace: sdk). Log: account sequence mismatch, expected 255, got 254: incorrect account sequence

## Observed

The tx was signed carrying sequence **254** while the chain already expected **255**. Broadcast is rejected and there is **no automatic recovery** — the MPC ceremony is dead and the user just sees the raw code-32 error.

## Where it comes from

- Sequence is fetched from **confirmed chain state only** (`getCosmosAccountInfo.ts`) and frozen into the `KeysignPayload` at build time — `packages/core/mpc/keysign/chainSpecific/resolvers/cosmos/index.ts:105` (`sequence: sequenceBigInt`). It then travels unchanged through the whole ceremony; the signing-input resolver only reads it, never refetches (`packages/core/mpc/keysign/signingInputs/resolvers/cosmos/index.ts:551`).
- **No mempool / pending accounting** — see the explicit design note at `packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.ts:88-99`. So an earlier send from the same vault that is still in-flight when this payload is built (or two racing sends) leaves this signed tx permanently invalid.
- **Code 32 is not retried** — it's absent from `transientMessagePatterns` (`packages/core/chain/tx/broadcast/transientRetry.ts:41-56`); `broadcastCosmosTx` only special-cases "already in cache" + RPC timeout (`packages/core/chain/tx/broadcast/resolvers/cosmos.ts:19-56`). A real mismatch rethrows as final. (Resigning with a fresh sequence needs a whole new ceremony, since sequence is baked into the signed bytes.)
- iOS has a parallel Swift implementation with the same gap (`VultisigApp/.../Core/Services/BlockChainService.swift`, `CosmosServiceStruct.broadcastTransaction:184-210`) — same code 32 surfaced raw.

## Suggested directions (eng to decide)

- Account for pending/in-flight sequence when building the payload (or, at minimum, warn on a same-vault pending Cosmos tx — like iOS `PendingTransactionManager`, but the collision here is cross-device so it needs to be relay/session-aware).
- On code 32, surface a clear "sequence changed — re-sign" affordance instead of a dead raw error.

Sequence handling is SDK-canonical; this manifests on Windows + iOS.

rcoderdev:
Closes #1547


gomesalexandre (commented):
Current head `6c24af4c8670c6c0e4800a4d0c1bd67dcc3183b9` looks conceptually sound from a grounded remote pass.

What I re-grounded:
- the diff keeps the stale-sequence recovery scoped to Cosmos broadcast handling rather than widening the signing surface
- current-head CI is fully green
- Codex remote review agreed there is no fresh correctness blocker on the current diff

I am not approving this head yet because this is a user-visible SDK behavior change and it still lacks a real current-head CLI/runtime receipt exercising the stale-sequence recovery path against a backend. For SDK changes at this layer, unit/Vitest coverage alone is not enough for stamp.

Please add an exact-head runtime receipt on `6c24af4c8670c6c0e4800a4d0c1bd67dcc3183b9` showing the actual SDK/CLI path through the stale-sequence recovery flow, then this is in much better shape for approval.

gomesalexandre (approved):
Reviewed this as a broadcast-path change, which meant the first question was whether anything here can retry a transaction that might already have landed. It cannot, and the design is right for the reason that matters.

## why this is safe

Code 32 is a **CheckTx** rejection, so the transaction never entered the mempool and there is no in-flight copy to double-spend against. On top of that, this classifies rather than retries: it returns a typed error carrying the direction and lets the caller decide. That is the correct division.

The `resign` / `wait` split is the real content, and it is the right way round:

- `signedSequence < expectedSequence` -> **resign**. The chain has already consumed that sequence, so those exact signed bytes can never become valid. Marking it permanent is not just an optimisation, it stops the retry loop chewing on a transaction that is definitionally dead.
- `signedSequence > expectedSequence` -> **wait**. Signed ahead of chain state, may become valid once the predecessor lands. Stays retryable.

Before this, code 32 was uniformly treated as transient, so the stale-sequence case retried forever and eventually surfaced as a timeout rather than "rebuild this". The comment you replaced said as much.

## verified rather than read

Drove `toCosmosSequenceMismatchError` directly, with must-fail controls alongside the must-pass ones, since a classifier that matches everything is worse than none:

```
MUST-PASS behind (resign)    -> resign (expected=42 signed=41)
MUST-PASS ahead (wait)       -> wait   (expected=42 signed=43)
MUST-PASS nested cause       -> resign (expected=7  signed=5)
MUST-PASS originalError      -> resign (expected=9  signed=8)
MUST-FAIL equal sequences    -> undefined
MUST-FAIL other code (5)     -> undefined
MUST-FAIL unrelated          -> undefined
MUST-FAIL circular           -> undefined
```

That last one is worth naming: I handed it an error whose `cause` points at itself, and it terminated instead of hanging. The `seen` set in the traversal is doing real work, and walking both `cause` and `originalError` means it still finds the rejection through the `VaultError` wrapper, which is where it will actually live in production.

The `expectedSequence === signedSequence` guard returning `undefined` is a good instinct too. A node reporting a mismatch between a number and itself is nonsense, and classifying nonsense is how you get a confidently wrong recovery hint.

```
packages/core/chain/tx/broadcast/   14 files, 117 passed
```

## nit: the matched string is owned by someone else

`broadcasting transaction failed with code 32 (codespace: sdk). log: ...` is never produced anywhere in this repo, only matched. So the format belongs upstream and can move without warning, and there are now two regexes that both depend on the same prefix:

```
packages/core/chain/tx/broadcast/cosmosSequenceMismatch.ts:10   (new, code 32 specific)
clients/cli/src/core/errors.ts:472                              (pre-existing, generic)
```

The failure mode is the safe one, so this is a nit and not a blocker: if the wording drifts, `parseCosmosSequenceMismatchMessage` returns `undefined` and you fall back to today's generic handling rather than misclassifying anything. It just does so **silently**, and the symptom would be "the stale-sequence case started looping again" long after the change that caused it. Not worth restructuring now; worth knowing that is the tell.

## process note

My local `@vultisig/cli` run showed 44 failures and I nearly brought them here. They were stale `dist` from switching branches without rebuilding the workspace packages, which the incoherent control gave away: main reported 44 failed FILES with only 530 tests collected, the branch 10 failed files with 1068. Your CI is green across Unit, Integration, Lint and Type Check. Nothing for you.

PR #1884, opened 2026-08-12
Guard shared DEX token decimals

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ERC-20 token metadata validation across Uniswap V2 and V3 integrations.
  * Rejects implausible decimal values above 36, preventing invalid token data from being processed.
* **Tests**
  * Added coverage confirming invalid decimal values are rejected across supported Uniswap integrations.
* **Release**
  * Included a patch release update for the SDK.


gomesalexandre (approved):
approve. exact-head review on `origin/main...origin/pr/1884` looks clean.

receipts
- explicit-ref diff grounded on `refs/remotes/origin/main...refs/remotes/origin/pr/1884`
- focused exact-ref unit receipt: `./node_modules/.bin/vitest --config packages/sdk/tests/unit/vitest.config.ts run packages/sdk/tests/unit/tools/dex/dexQuotes.test.ts` -> 10 passed
- verified the decimals guard now lives in shared `packages/sdk/src/tools/dex/_erc20.ts`, so Uniswap V2 and V3 read paths enforce the same `> 36` fail-closed bound
- verified `packages/sdk/src/tools/dex/uniswap/erc20.ts` is now a pure compatibility re-export, with no behavior drift beyond adopting the shared guard
- changeset present and scoped

note: `yarn build:sdk` in the temp exact-ref checkout still hits a pre-existing Cardano typing failure outside this PR (`packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts:72 auxiliaryData`), so I did not use that unrelated red build as evidence against this diff.

this looks like the right minimal hardening for attacker-controlled token metadata on the DEX quote path.

PR #1886, opened 2026-08-12
fix(sui): fail closed on missing pagination cursor

### LINKED ISSUE #1590: sui: getSuiAllBalances returns ok:true with a PARTIAL portfolio when hasNextPage:true + cursor null
Follow-up from the sui gRPC migration review (#1583, merged). Non-blocking there, flagged for a fast-follow.

## the edge

`packages/core/chain/chains/sui/listAllCoins.ts:68` and the mirror in `getSuiAllBalances` (`packages/sdk/src/tools/balance/otherBalance.ts`): a page with `hasNextPage: true` AND a null cursor exits the loop and returns the partial set as if complete.

- keysign path: degrades SAFE - a spurious `Insufficient Sui coin balance` throw downstream, never a wrong send.
- `getSuiAllBalances`: returns `ok: true` with a PARTIAL portfolio - the exact fund-visibility bug its own docstring forbids. A user could see an incomplete balance shown as complete.

It requires the upstream (gRPC/GraphQL) to return `hasNextPage:true` with a null `endCursor`, which violates the connection contract - hence unlikely and non-blocking - but it is a real correctness hole.

## the tell

The existing test at `listAllCoins.test.ts:76` is named `does not truncate when hasNextPage is true but the cursor is null` while actually ASSERTING the truncating behaviour - a test name that contradicts its assertion, which is how this slipped.

## fix

- `listAllSuiCoins`: treat `hasNextPage:true` + null cursor as malformed and throw (fail closed), rather than returning partial.
- `getSuiAllBalances`: surface `tokens_unavailable` instead of `ok:true` on that combination.
- rename the test to match what it should assert (throws / unavailable), and add the true no-truncation case.

Related: #1583.

rcoderdev:
Closes #1590

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Sui pagination handling to prevent incomplete coin and portfolio results from being reported as complete.
  - Added fail-closed behavior when pagination metadata is missing, malformed, or lacks a usable continuation cursor.
  - Preserved valid results across empty intermediate pages and correctly handled page-limit boundaries.

- **Tests**
  - Expanded coverage for incomplete and edge-case pagination scenarios.


gomesalexandre (approved):
approve. exact-head review on `origin/main...origin/pr/1886` looks clean.

receipts
- explicit-ref diff grounded on `refs/remotes/origin/main...refs/remotes/origin/pr/1886`
- verified `packages/core/chain/chains/sui/listAllCoins.ts` now fails closed when `hasNextPage` is true but `cursor` is nullish, instead of silently truncating the coin set
- verified `packages/sdk/src/tools/balance/otherBalance.ts` mirrors that fail-closed rule for `getSuiAllBalances`, returning `tokens_unavailable` instead of partial-success balances when pagination metadata is malformed
- verified the new tests cover null/undefined/empty-string cursor shapes, preserve the legitimate terminal-page stale-cursor case, and keep the existing max-pages / accumulation behavior grounded
- changeset present and scoped to the touched packages

this closes a real partial-read hazard on Sui balances without widening the surface.

PR #1887, opened 2026-08-12
fix: share USDC amount parser

rcoderdev:


gomesalexandre (approved):
approve. exact-head review on `origin/main...origin/pr/1887` looks clean.

receipts
- explicit-ref diff grounded on `refs/remotes/origin/main...refs/remotes/origin/pr/1887`
- verified the two prior `parseUsdcAmount` implementations were behavior-equivalent before the move, so `packages/sdk/src/tools/parse/usdcAmount.ts` is a real dedupe, not a semantic rewrite
- verified both original call sites now re-export the shared parser, preserving the existing builder/public import surfaces
- verified `defi/index.ts` still exposes `threeJane` only as a namespace, so this does not create a new flat export collision at the SDK top level
- verified the new `packages/sdk/tests/unit/usdc-amount.test.ts` pins shared-function identity plus malformed-input rejection parity across both builders
- changeset present and correctly scoped

clean low-risk dedupe with the right regression coverage.

PR #1888, opened 2026-08-12
test: pin Sui parser and protobuf wire contracts

rcoderdev:
## Summary by CodeRabbit

- **Tests**
  - Expanded validation for Sui transaction results, including successful, failed, and incomplete responses.
  - Added compatibility checks for signing payload serialization across mobile fixtures.
  - Added coverage for swap payload construction and contract-compliant coin data.
- **Reliability**
  - Improved consistency in reporting transaction status, gas costs, and errors.
  - Strengthened signing and swap payload compatibility across supported workflows.

### DIFF
--- a/packages/core/chain/chains/sui/transactionResult.test.ts
+++ b/packages/core/chain/chains/sui/transactionResult.test.ts
@@ -83,10 +83,10 @@ describe('isSuiExecutionSuccess — @mysten/sui parser contract', () => {
       { include: { effects: true } }
     )
 
-  it('classifies the library parser output for success, failure, and missing status', () => {
+  it('classifies the library parser output for success, failure, and missing effects', () => {
     const parsedSuccess = parseTransaction(true)
     const parsedFailure = parseTransaction(false)
-    const parsedWithoutStatus = parseTransaction()
+    const parsedWithoutEffects = parseTransaction()
 
     expect(parsedSuccess).toMatchObject({
       $kind: 'Transaction',
@@ -117,7 +117,7 @@ describe('isSuiExecutionSuccess — @mysten/sui parser contract', () => {
         },
       },
     })
-    expect(parsedWithoutStatus).toMatchObject({
+    expect(parsedWithoutEffects).toMatchObject({
       $kind: 'FailedTransaction',
       FailedTransaction: {
         status: { success: false, error: { $kind: 'Unknown', message: 'Transaction failed' } },
@@ -128,7 +128,7 @@ describe('isSuiExecutionSuccess — @mysten/sui parser contract', () => {
     expect([
       isSuiExecutionSuccess(parsedSuccess),
       isSuiExecutionSuccess(parsedFailure),
-      isSuiExecutionSuccess(parsedWithoutStatus),
+      isSuiExecutionSuccess(parsedWithoutEffects),
     ]).toEqual([true, false, false])
   })
 })
--- a/packages/core/chain/chains/sui/transactionResult.test.ts
+++ b/packages/core/chain/chains/sui/transactionResult.test.ts
@@ -61,15 +61,26 @@ describe('isSuiExecutionSuccess — @mysten/sui parser contract', () => {
     parseGrpcTransactionResponse(
       GrpcTypes.ExecutedTransaction.create({
         digest: 'parsed-digest',
-        effects: GrpcTypes.TransactionEffects.create({
-          status:
-            success === undefined
-              ? undefined
-              : GrpcTypes.ExecutionStatus.create({
-                  success,
+        effects:
+          success === undefined
+            ? undefined
+            : GrpcTypes.TransactionEffects.create({
+                transactionDigest: 'parsed-effects-digest',
+                gasUsed: GrpcTypes.GasCostSummary.create({
+                  computationCost: 11n,
+                  storageCost: 22n,
+                  storageRebate: 3n,
+                  nonRefundableStorageFee: 1n,
                 }),
-        }),
-      })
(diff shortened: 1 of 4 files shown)


gomesalexandre (changes requested):
blocking: `packages/core/mpc/keysign/tests/keysignPayloadWire.golden.test.ts` does not actually pin the full fixture corpus wire contract yet.

At current head, the new corpus test only asserts:
- `serialized` has length 71
- every payload produces non-empty bytes

That means broad wire drift can still pass as long as the fixtures still serialize. Only one case is truly pinned today: `thorchain.json` / `Stake Ruji` via the single `stakeRujiWireHex` golden.

Why this matters on this diff:
- `normalizeKeysignPayloadFromJson.ts` now changes wasm coin construction to typed `CosmosCoinSchema` values
- `mapSwapPayload.ts` now changes the nested 1inch quote/tx construction to typed schema instances

Those are exactly the kinds of serializer-path changes where we want corpus-wide byte protection, not just "still serializes".

receipts
- exact-head review on `refs/remotes/origin/main...refs/remotes/origin/pr/1888`
- focused exact-head vitest receipt: `packages/core/chain/chains/sui/transactionResult.test.ts` + `packages/core/mpc/keysign/tests/keysignPayloadWire.golden.test.ts` -> 2 files passed, 17 tests passed
- verified `transactionResult.test.ts` is a good parser-contract pin for success / failure / missing-status on the `@mysten/sui` parser path
- verified `keysignPayloadWire.golden.test.ts:46-50` only checks corpus cardinality + non-empty bytes, while `:53-59` pins exact bytes for only a single rich fixture

please strengthen the wire-contract test to pin either:
- each fixture's exact bytes, or
- a deterministic aggregate digest over the whole serialized corpus

once that is in, happy to re-check.

gomesalexandre (dismissed):
Tackled, verified independently — dismissing the prior CHANGES_REQUESTED.

Re-ran \`keysignPayloadWire.golden.test.ts\` myself against the pushed commit (40315ec10): 26/26 green, corpus digest + all 23 per-fixture-file digests pass on a clean checkout, not just the reporter's word for it. The aggregate-digest approach (length-prefixed sha256 over the full serialized corpus, plus a per-file digest to narrow a future mismatch to the offending fixture) closes the actual gap — this now fails on ANY wire-format drift across all 71 fixtures, not just the one previously-pinned rich fixture. Digests are derived from a real run, not hand-typed.

CI still shows the same org-wide Actions queue stall affecting the whole org right now (not a real failure on this diff) — approving on the strength of the local verification. Merge once CI actually clears.

rcoderdev:
Preserved the full-corpus aggregate and per-file digest correction while rebasing onto current main. The current head is `e14ffb873`; the 113 focused dependency-contract tests and `yarn check:ci` pass on that integrated revision. I also tightened the Sui contract to exercise the production `include.effects` parser path and pin the transaction/effects digests, gas fields, failure message, and fail-closed classifications. The four-eyes correction is valid, so this now needs a current-head approval from an eligible non-author reviewer.

gomesalexandre (commented):
Current-head rereview on `e14ffb873` only.

What I rechecked on this head:
- `packages/core/chain/chains/sui/transactionResult.test.ts` now drives `isSuiExecutionSuccess` from real `@mysten/sui/grpc` parser output instead of hand-rolled effect/status shapes, so the success/failure/missing-status contract is pinned at the actual library boundary that can drift.
- `packages/core/mpc/keysign/tests/helpers/normalizeKeysignPayloadFromJson.ts` and `packages/core/mpc/keysign/tests/mappers/mapSwapPayload.ts` now build protobuf-es values with `create(...)` / the generated schemas instead of plain-object casts, which is the right way to harden serializer-path tests when the point is wire-format fidelity.
- `packages/core/mpc/keysign/tests/keysignPayloadWire.golden.test.ts` now does the important thing the earlier head was still missing: it pins deterministic sha256 digests for the full serialized mobile-fixture corpus plus per-fixture-file digests, so future wire drift in the helper/mappers fails loudly across the whole corpus rather than only on the single `Stake Ruji` hex golden.

Receipts I ran against the exact PR head in an isolated checkout at `e14ffb873175d4b59cb18df8f511e8ce7e6ef67d`:
- `./node_modules/.bin/vitest run packages/core/mpc/keysign/tests/keysignPayloadWire.golden.test.ts packages/core/chain/chains/sui/transactionResult.test.ts` -> 41 passed
- `./node_modules/.bin/tsc -p .config/tsconfig.json --noEmit` still hits the same pre-existing `auxiliaryData` Cardano typing errors on both `origin/main` and this PR head, so that red is not introduced by this diff

On the current CodeRabbit nit: the relative generated-schema import in `packages/core/mpc/keysign/tests/mappers/mapSwapPayload.ts` is an inconsistency worth cleaning eventually, but it is test-only, already coexists with the same relative style elsewhere in this test tree, and does not affect the wire output the new goldens are pinning. I do not treat that as blocking.

I do not see a fresh blocker on the current diff.

Process note only: this branch now contains gomes-authored commits (`302334bcc`, on top of the earlier review history), so four-eyes forbids my approval here. Once current-head CI settles, this needs a non-gomes current-head stamp.

rcoderdev:
Thanks for the current-head re-review. I incorporated the final CodeRabbit correction by naming the omitted-effects case precisely, rebased onto current main, and reran the 113 focused tests, `yarn check:ci`, and the installed-library runtime probe at `01c657ab0`. The four-eyes constraint remains valid, so this still requires an eligible non-author approval.

gomesalexandre (approved):
Approved by gomes (human, explicit).

PR #1889, opened 2026-08-12
test: validate SDK package export manifest

### LINKED ISSUE #1609: test: validate SDK package export manifest
## Summary

`@vultisig/sdk` has a large, conditional package export surface, but there is no single release gate that proves the packed artifact satisfies `packages/sdk/package.json#exports`. Export regressions are therefore discovered as individual missing-root, React Native, or subpath tickets after a consumer tries an import.

## Scope

Add one package-artifact contract test for the SDK release boundary. The manifest must be the source of truth; do not maintain a second hand-written list of expected entry points.

## Required coverage

- Build and pack `@vultisig/sdk`, then run the assertions against the tarball/packed artifact rather than the TypeScript source tree.
- Walk every `exports` entry and every conditional target in `packages/sdk/package.json`.
- Assert that each referenced JavaScript and declaration target is present in the packed artifact.
- Prove the Node import and require paths in a clean consumer fixture; validate platform-specific conditions and declarations without trying to execute browser/RN bundles in Node.
- Include the root, `node`, `browser`, `react-native`, `rn-preamble`, `electron`, `electron/main`, `chrome-extension`, `vite`, `tools/parse`, and `tools/defi` surfaces currently declared by the manifest.

## Done when

- Removing or mispointing any manifest target makes the new release gate fail.
- Adding an export automatically extends the test surface without changing a parallel allow-list.
- The gate runs before publish/release, with a focused command documented for local verification.

## Why now

Recent public-API findings have been individual symptoms of the same missing release contract: #1603, #1607 and #1587. One artifact-level gate is cheaper than continuing to repair each missing surface separately.

rcoderdev:
Closes #1609

Adds a manifest-derived SDK package release contract that validates every conditional JavaScript and declaration target in the packed artifact without a parallel allow-list.

Real package proof: the actual SDK tarball was installed into a clean temporary consumer; Node executed every compatible import and require path, TypeScript resolved the full manifest-derived export surface, and a deliberately missing declaration target made the gate fail closed.

## Summary by CodeRabbit

* **Quality Improvements**
  * Added automated validation to ensure SDK package exports and published files are complete, valid, and usable.
  * Release builds now verify import, require, runtime, and TypeScript compatibility before publishing.

* **Documentation**
  * Documented the SDK package export validation command and its checks.

* **Tests**
  * Added coverage for invalid, unsafe, missing, and condition-specific export targets.
  * Improved browser example testing to avoid unnecessary rebuilds.


gomesalexandre (commented):
The new package-export gate still has a hole in its declaration proof.

`scripts/check-sdk-package-exports.mjs` builds `verify-types.ts` out of bare side-effect imports:

```ts
import '@vultisig/sdk/vite'
```

That does not force TypeScript to resolve the declared `types` target for the subpath, and the generated tsconfigs do not enable `noUncheckedSideEffectImports` to make unresolved side-effect imports fail closed. A manifest entry like `./vite` can therefore point `types` at an existing JS file and this check can still go green while typed consumers break.

I re-grounded this locally with a minimal TS repro: a package exporting `./vite` with `types: "./dist/vite/index.js"` passes both the side-effect-import check and a forced `import type` check under TS 5.9.2, so the current gate is not actually proving that each subpath resolves to a valid declaration file.

Please tighten the proof so each exported subpath asserts its declaration target directly, rather than relying on side-effect imports.

gomesalexandre (dismissed):
This is the right way round for a packaging contract, and the design claim in the description holds up under checking.

## manifest-derived, not a parallel list

The surface comes from the manifest itself:

```js
const typeImports = Object.keys(manifest.exports)
```

so there is no second list to drift. The test that proves the property is the one I would have asked for:

```
walks every conditional target without a subpath allow-list
fails when any manifest target is missing from the packed artifact
rejects directory and package-escaping targets
a newly added export automatically joins structural and Node runtime coverage
derives every custom TypeScript declaration condition from the manifest
```

That second one is the load-bearing negative control - a validator with no failing case is decoration - and the fourth encodes the actual invariant (add an export, get coverage for free) rather than just asserting today's shape. 5 pass, 0 fail locally.

Path-escape rejection is a nice inclusion too: an exports map is a place where `../` targets can quietly point outside the package.

## the thing I went looking for: lost coverage

`quality-contracts.mjs` loses 156 lines, and refactors like this are where runtime symbol checks silently become mere file-resolution checks. They did not. The assertions moved rather than evaporated (`check-sdk-package-exports.mjs:194+`), and every symbol survives:

```
removed typeof lines in quality-contracts.mjs   17
of those, symbols absent from the new validator  0
```

The 17-vs-12 raw count difference is not a gap - several of the removed lines are type-level uses inside the generated consumer fixture (`typeof extension`, `typeof descriptor`, `typeof registry`), not runtime assertions. Every actual `root.` / `node.` / `electronMain.` symbol check has a home.

So this trades a hand-maintained allow-list for a derived one **without** trading away the runtime symbol assertions, which is the version of this refactor worth having.

## wiring

```yaml
- name: Validate SDK package export manifest
  run: yarn quality:sdk-package-exports --skip-build
```

Running it in `release.yml` is the correct gate placement - an exports map that resolves in the repo but not in the packed tarball is precisely a release-time failure, and the description says the real tarball is installed into a clean temporary consumer to prove it.

CI green.

gomesalexandre (dismissed):
Current-head review on `38adc30a6856771f855f603a7e9ffdaf4be12ade`.

Request changes.

The declaration-proof hole is still present on this head.

At `scripts/check-sdk-package-exports.mjs:239-245`, the generated `verify-types.ts` still starts by importing every manifest subpath as a bare side-effect import:

```ts
import '@vultisig/sdk/vite'
import '@vultisig/sdk/seedphrase'
...
```

And at `scripts/check-sdk-package-exports.mjs:289-297`, the generated `tsconfig.types-*.json` still does **not** enable `noUncheckedSideEffectImports`.

That means this check still does not fail closed on a bad `types` target for the subpaths that never get a real typed use site. TypeScript only needs the module to resolve somehow; it does not have to prove that the declared `types` target is the right `.d.ts` file.

The new custom-condition derivation is real progress, but it only strengthens the paths that already get typed symbol usage later in the file. Several exported subpaths in `packages/sdk/package.json` still fall through to the weak side-effect-import proof only, including `./browser`, `./chrome-extension`, `./vite`, `./tools/parse`, `./tools/defi`, `./tools/bridge`, `./seedphrase`, and `./rn-preamble`.

That is especially visible for `./seedphrase`: the previous `quality-contracts.mjs` consumer smoke exercised typed imports from that subpath, but the new validator reduces it to a bare side-effect import, so this refactor lost real declaration coverage there instead of just leaving the old gap in place.

Please tighten the type proof before merge. Two safe directions:
1. generate an actual typed assertion for every exported subpath, not just the root / node / react-native / electron-main surfaces, or
2. at minimum enable `noUncheckedSideEffectImports` on every generated tsconfig so a bad declaration target fails closed instead of silently passing.

Also process note only: this head includes a gomes-authored merge commit, so even after the proof is fixed it will still need a non-gomes current-head stamp under the four-eyes rule.

gomesalexandre (commented):
Current head looks fine to me on substance.

What I re-checked this cycle:
- `scripts/check-sdk-package-exports.mjs` is now the single source of truth for the packed-SDK export proof, including the runtime-key introspection path (`collectIntrospectableRuntimeCases`, `collectRuntimeExportKeys`, `writeConsumerTypeFiles`) and `noUncheckedSideEffectImports: true` in the generated tsconfig.
- `package.json` wires `quality:sdk-package-exports` into `yarn release`, and `.github/workflows/release.yml` runs `yarn quality:sdk-package-exports --skip-build` before publish.
- The current head commit `454b7d4f9` is gomes-authored and specifically closes the declaration-proof hole for uncurated export subpaths, so I cannot be the approving stamp on this branch.

I am leaving this as comment-only because four-eyes still requires a non-gomes approval, but I do not see a blocker in the current diff.

gomesalexandre:
The declaration-proof gap from my earlier review is closed as of \`454b7d4f9\`: uncurated subpaths now get a real typed-usage resolution attempt (falling back through node/import conditions, only skipping genuinely RN/worker-only targets) instead of relying solely on the side-effect import, and \`noUncheckedSideEffectImports: true\` is now set on the generated tsconfig. CI is green.

I can't self-approve since I have a commit in this head (per our four-eyes rule) - needs a fresh external stamp before merge.

gomesalexandre (commented):
Current head `808fee06a` only adds the prettier follow-up on top of the earlier declaration-proof fix, and I do not see a new substance blocker in that delta.

Still not mergeable though:
- GitHub currently reports this PR as `CONFLICTING` / `DIRTY` against `main`
- this head contains gomes-authored commits (`454b7d4f9`, `808fee06a`), so four-eyes still requires a fresh non-gomes current-head stamp after the branch is rebased or merge-fixed

So from me this stays comment-only on the current head, not approval.

rcoderdev:
Rebased this branch onto current main and pushed head 41e4d2f15. The declaration-proof coverage from 454b7d4f9 remains intact. I also addressed the current automated review findings by isolating runtime introspection in bounded child processes with contextual failures, requiring a path-segment boundary for CommonJS resolution, and using a disposable Yarn cache with immutable installs disabled only for the generated consumer. The focused 9-test suite, the packed clean-consumer gate (17 exports, 111 conditional targets, 13 imports, 13 requires, 5 declaration condition sets, 11 typed assertions), and yarn check:ci all pass on this head. This is ready for a fresh non-gomes current-head review.

johnnyluo (approved):
LGTM

PR #1891, opened 2026-08-12
refactor: standardize broadcast result contract

### LINKED ISSUE #1610: refactor: standardize raw-broadcast adapter result contract
## Summary

`packages/core/chain/tx/broadcast/resolver.ts` types every chain broadcaster as `Promise<unknown>`, while `broadcastTx` dispatches twelve chain-kind resolvers. The SDK has no typed, shared result boundary for callers to distinguish an accepted broadcast, a retryable transport failure, and a definitive rejection consistently across chains.

## Scope

Replace the `Promise<unknown>` broadcast boundary with one exported SDK result contract and adapt every resolver to it. This is a contract-normalization change, not a rewrite of chain-specific broadcast mechanics.

## Proposed contract

Define a discriminated result that gives consumers:

- an explicit accepted/success state with the canonical transaction identifier when the chain returns one;
- a failure state with a stable SDK error code, retryability classification, and the original cause retained for logging;
- a documented representation for chains whose broadcast acknowledgement cannot yet prove final execution.

Keep chain-native response data only as an explicitly namespaced optional detail, so consumers cannot accidentally depend on a provider-specific shape.

## Implementation plan

1. Introduce the exported result and error types beside the broadcast resolver contract.
2. Change `BroadcastTxResolver` from `Promise<unknown>` to the new result type.
3. Adapt the resolvers registered in `packages/core/chain/tx/broadcast/index.ts` (bittensor, cardano, cosmos, evm, polkadot, qbtc, ripple, solana, sui, ton, utxo and tron).
4. Preserve resolver-owned retry behavior for EVM and Solana, and make the shared transient retry emit the same result contract as direct calls.

## Test matrix

- One table-driven contract suite executes every registered resolver through deterministic mocked transport responses.
- Each adapter proves: accepted broadcast, definitive rejection, and transient/retryable failure where that distinction exists.
- Assert that consumers never receive a raw provider response or an untyped thrown transport shape from the SDK boundary.
- Add focused regression cases for any chain that cannot provide a transaction ID at acknowledgement time.

## Done when

- Callers can handle all supported chains through one typed result union.
- Adding a resolver to the dispatch map without the contract test is impossible or fails CI.
- Existing chain-specific semantics remain covered, but the public result no longer varies by provider.

rcoderdev:
Closes #1610

## Summary by CodeRabbit

* **New Features**
  * Standardized transaction broadcasting across supported networks with clear accepted or failed outcomes.
  * Accepted broadcasts consistently include transaction hashes, finality status, and provider details where available.
  * Ambiguous submissions can be verified on-chain before being marked successful or failed.
* **Bug Fixes**
  * Improved retry handling for transient transport failures while preventing retries for definitive transaction failures.
  * Broadcast errors preserve actionable failure codes, retryability, and original causes.
  * SDK services reliably consume normalized broadcast results and fall back to locally derived hashes when needed.

### DIFF
--- a/.changeset/standard-broadcast-result.md
+++ b/.changeset/standard-broadcast-result.md
@@ -1,5 +1,5 @@
 ---
-'@vultisig/core-chain': patch
+'@vultisig/core-chain': major
 '@vultisig/sdk': patch
 ---
 
--- a/.changeset/standard-broadcast-result.md
+++ b/.changeset/standard-broadcast-result.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Standardize signed-transaction broadcast adapters on one typed accepted-or-failed result contract.
--- a/package.json
+++ b/package.json
@@ -192,7 +192,7 @@
     "vitest": "^4.1.8"
   },
   "resolutions": {
-    "nanoid@npm:^3.3.16": "^3.3.17",
+    "nanoid@npm:^3.3.16": "^3.3.18",
     "readable-stream": "3.6.2",
     "@grpc/grpc-js": "1.14.4",
     "@cosmjs/encoding/@scure/base": "1.2.6",
--- a/packages/core/chain/tx/broadcast/cosmosDeliverTxRetry.test.ts
+++ b/packages/core/chain/tx/broadcast/cosmosDeliverTxRetry.test.ts
@@ -56,10 +56,10 @@ describe('broadcastTx cosmos DeliverTx-failure retry interlock', () => {
       })
       .mockRejectedValueOnce(new Error('tx already exists in cache'))
 
-    const rejection: unknown = await broadcastTx({ chain, tx }).catch(caught => caught)
+    const result = await broadcastTx({ chain, tx })
 
-    expect(rejection).toBeInstanceOf(DeliverTxFailedError)
-    expect((rejection as Error).message).toMatch(/DEF456/)
+    expect(result).toMatchObject({ status: 'failed', retryable: false, cause: expect.any(DeliverTxFailedError) })
+    expect(result).toHaveProperty('cause.message', expect.stringMatching(/DEF456/))
     expect(mocks.broadcastTx).toHaveBeenCalledTimes(1)
     expect(mocks.verifyBroadcastByHash).not.toHaveBeenCalled()
   })
@@ -78,7 +78,9 @@ describe('broadcastTx cosmos DeliverTx-failure retry interlock', () => {
       })
       .mockRejectedValueOnce(new Error('tx already exists in cache'))
 
-    await expect(broadcastTx({ chain, tx })).rejects.toThrow(/GHI789/)
+    const result = await broadcastTx({ chain, tx })
+    expect(result).toMatchObject({ status: 'failed', retryable: false, cause: expect.any(DeliverTxFailedError) })
+    expect(result).toHaveProperty('cause.message', expect.stringMatching(/GHI789/))
 
     expect(mocks.broadcastTx).toHaveBeenCalledTimes(1)
     expect(mocks.verifyBroadcastByHash).not.toHaveBeenCalled()
@@ -92,7 +94,9 @@ describe('broadcastTx cosmos DeliverTx-failure retry interlock', () => {
       rawLog: 'out of gas',
     })
 
-    await expect(broadcastTx({ chain, tx })).rejects.toThrow(/ABC123/)
+    const result = await broadcastTx({ chain, tx })
+    expect(result).toMatchObject({ status: 'failed', retryable: false, cause: expect.any(DeliverTxFailedError) })
+    expect(result).toHaveProperty('cause.message', expect.stringMatching(/ABC123/))
 
     expect(mocks.broadcastTx).toHaveBeenCalledTimes(1)
     expect(mocks.verifyBroadcastByHash).not.toHaveBeenCalled()
--- a/packages/core/chain/tx/broadcast/index.test.ts
+++ b/packages/core/chain/tx/broadcast/index.test.ts
@@ -2,49 +2,90 @@ import { HttpResponseError } from '@vultisig/lib-utils/fetch/HttpResponseError'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 const mocks = vi.hoisted(() => ({
-  broadcastBittensorTx: vi.fn(),
-  broadcastCardanoTx: vi.fn(),
-  broadcastCosmosTx: vi.fn(),
-  broadcastEvmTx: vi.fn(),
-  broadcastPolkadotTx: vi.fn(),
-  broadcastQbtcTx: vi.fn(),
-  broadcastRippleTx: vi.fn(),
-  broadcastSolanaTx: vi.fn(),
-  broadcastSuiTx: vi.fn(),
-  broadcastTonTx: vi.fn(),
-  broadcastTronTx: vi.fn(),
-  broadcastUtxoTx: vi.fn(),
+  bittensor: vi.fn(),
+  cardano: vi.fn(),
+  cosmos: vi.fn(),
+  evm: vi.fn(),
+  polkadot: vi.fn(),
+  qbtc: vi.fn(),
+  ripple: vi.fn(),
+  solana: vi.fn(),
+  sui: vi.fn(),
+  ton: vi.fn(),
+  tron: vi.fn(),
+  utxo: vi.fn(),
 }))
 
-vi.mock('./resolvers/bittensor', () => ({
-  broadcastBittensorTx: mocks.broadcastBittensorTx,
-}))
-vi.mock('./resolvers/cardano', () => ({
-  broadcastCardanoTx: mocks.broadcastCardanoTx,
-}))
-vi.mock('./resolvers/cosmos', () => ({
-  broadcastCosmosTx: mocks.broadcastCosmosTx,
-}))
-vi.mock('./resolvers/evm', () => ({ broadcastEvmTx: mocks.broadcastEvmTx }))
-vi.mock('./resolvers/polkadot', () => ({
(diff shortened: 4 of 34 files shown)


gomesalexandre (dismissed):
blocking on one public-surface issue before i’m comfortable stamping this.

1. `@vultisig/core-chain` needs a major bump, not a patch.

This PR changes the public `./tx/broadcast` contract in a breaking way:
- `packages/core/chain/tx/broadcast/resolver.ts` changes `BroadcastTxResolver` from `Promise<unknown>` to `Promise<BroadcastTxResult>` and adds the exported `BroadcastAcceptedResult` / `BroadcastFailedResult` union.
- `packages/core/chain/tx/broadcast/index.ts` now guarantees a normalized `{status: 'accepted'|'failed', ...}` result instead of the previous mixed behavior where callers could see `void`, a bare hash string, provider objects, or thrown errors from resolvers.
- `packages/core/chain/package.json` exports `./tx/broadcast` and `./tx/broadcast/resolver`, so this is not an internal-only refactor. Downstream consumers compiling against the old types or relying on throw-based failure control flow will break.

But `.changeset/standard-broadcast-result.md` marks `@vultisig/core-chain` as `patch`.

Please bump `@vultisig/core-chain` to `major` in the changeset, then i’m happy to re-check. The rest of the normalization work looks directionally solid, but shipping this as a patch would mis-state a real API/behavior break.

gomesalexandre (commented):
pushed `13f39fd31` on this branch to close the blocking public-surface issue.

- changed `.changeset/standard-broadcast-result.md` from `@vultisig/core-chain: patch` to `major`
- left `@vultisig/sdk: patch` unchanged
- kept the diff metadata-only

validation:
- verified the changeset frontmatter parses cleanly
- verified the changeset package names still match `packages/core/chain/package.json` and `packages/sdk/package.json`
- did not claim a full workspace receipt for this follow-up because this checkout has no install state and the change is one-line release metadata only

needs fresh non-gomes review on current head `13f39fd31`.

gomesalexandre (dismissed):
Reviewed this as a fund-safety change rather than a refactor, because "standardize the broadcast result contract" decides what the app believes about whether money moved. It is careful work and the two hardest calls are both made correctly.

## the contract itself

`status: 'accepted'` carrying `finality: 'pending'` is the right shape, and the comment saying so explicitly is worth keeping:

> a broadcast acknowledgement does not prove that the transaction executed successfully or reached finality

That is the phantom-completion failure mode designed out at the type level rather than patrolled downstream. Same for isolating chain-native data under `details.provider` "so it cannot become the common contract by accident" - that is the sentence that stops a Cardano-shaped field quietly becoming load-bearing for Solana in six months.

Verified the derived mapping rather than reading it, with must-not cases alongside the must-retry ones, since a classifier that says "retryable" to everything would look fine on the first two lines alone:

```
network timeout      retryable=true   BROADCAST_TRANSPORT_ERROR
502 bad gateway      retryable=true   BROADCAST_TRANSPORT_ERROR
insufficient funds   retryable=false  BROADCAST_REJECTED
invalid signature    retryable=false  BROADCAST_REJECTED
nonce too low        retryable=false  BROADCAST_REJECTED
accepted             status=accepted  finality=pending
```

A permanent rejection marked retryable is an infinite retry loop against a transaction that can never land; all three permanent shapes come back `Rejected`.

## the part I liked most

The ambiguous-submit-error handling. `cardano.ts:78` names the exact problem - a submit error can be a benign MPC-race duplicate (another device already broadcast the same signed bytes) or a real failure like `BadInputsUTxO`, and string-matching cannot separate them. Instead of guessing, it asks the chain:

```ts
try   { return broadcastAccepted(await verifyBroadcastByHash({ chain, tx, error })) }
catch (cause) { return broadcastFailed(cause, isRetryableBroadcastCause(error)) }
```

Ground truth over heuristics, and it fails to `failed` rather than to `accepted` when the hash does not resolve. That is the correct direction: telling a user their transaction went through when it did not is the expensive error here.

I checked whether that pattern is applied evenly, because a "standardize" PR that standardises 11 of 12 chains is worse than one that standardises none. It is consistent: every registered resolver uses it.

```
bittensor cosmos cardano evm polkadot qbtc ripple tron sui solana utxo ton   -> verifyBroadcastByHash
substrate.ts                                                                 -> not a resolver
```

`substrate.ts` came up as a miss in my first pass and is a false alarm - it exports `isIdempotentSubstrateBroadcastError` / `formatSubstrateRpcError` and is imported by polkadot and bittensor, not registered in `index.ts` as a chain entry.

```
packages/core/chain/tx/broadcast/   14 files, 178 passed
CI                                  all 5 checks green
```

## worth flagging: this also fixes nanoid

The root `package.json` change is not incidental to the board:

```
-    "nanoid@npm:^3.3.16": "^3.3.17",
+    "nanoid@npm:^3.3.16": "^3.3.18",
```

That is GHSA-2v37-7h3g-55p8, and it is why this PR's `Lint and Type Check` is green while ten other open PRs are red on the audit gate alone (#1722, #1716, #1674, #1664, #1662, #1658, #1897, #1862, #1805, #1875). #1929 fixes the same advisory a different way, via a plain lockfile refresh. Both work; whichever lands first unblocks that stack, so it is worth someone picking one deliberately rather than having them race and conflict on the lockfile.

gomesalexandre (commented):
Current-head rereview on `e93d368f`.

Comment, not approve.

I reread the broadcast-contract delta and also ran a focused fallback review pass after codex hard-failed on auth bootstrap (`invalid_refresh_token` / `token_expired`). The shape change itself looks coherent: the resolver layer now standardizes on `BroadcastAcceptedResult | BroadcastFailedResult`, `index.ts` still routes ambiguous outcomes through `verifyBroadcastByHash(...)` before retrying, and the SDK-facing consumer path remains `VaultBase.broadcastTx -> BroadcastService.broadcastTx`, so the public SDK surface is still string-or-throw even though `@vultisig/core-chain` takes the internal major bump.

Receipts from the fallback pass:
- `yarn vitest run --config .config/vitest.core.config.ts packages/core/chain/tx/broadcast`
- `yarn workspace @vultisig/sdk vitest run tests/unit/vault/services/BroadcastService.test.ts`

I did not find a new blocker at this head, but two defense-in-depth nits are worth keeping on the radar:

1. `packages/core/chain/tx/broadcast/resolvers/polkadot.ts` and `packages/core/chain/tx/broadcast/resolvers/ton.ts` still route some locally-constructed “missing/malformed hash” errors through the shared retryable-cause classifier instead of hardcoding `retryable: false` like the sibling resolvers do. Today the strings do not match the transient regex set, so behavior is fine, but explicitly pinning those local invariant failures non-retryable would make the contract more future-proof.
2. `packages/core/chain/tx/broadcast/resolvers/ton.ts` still accepts the duplicate-message path directly from the error string without a `verifyBroadcastByHash(...)` follow-up or a dedicated regression test. That shortcut predates this PR, so I am not calling it a regression, but this refactor is the exact place where a small focused test would buy confidence.

Because the current branch now contains a gomes-authored commit (`13f39fd3` changeset), this cannot take a gomes approval under the four-eyes rule. If CI stays green, the next step is a non-gomes human stamp, not an approval from me.

gomesalexandre (commented):
Current head still looks good to me on substance.

What I re-checked this cycle:
- the branch now includes gomes-authored commit `13f39fd3` (`chore(changeset): mark core-chain broadcast contract major`), so I cannot be the approving stamp here.
- the current `polkadot` and `ton` broadcast resolvers now fail closed on the missing-hash fallback path by returning `broadcastFailed(..., false)` after verify-by-hash fails, instead of treating the malformed/missing-hash response as retryable by default.
- current head CI is green again across unit, integration, lint/typecheck, and summary.

So I do not see a fresh blocker in the current diff, but four-eyes still requires a non-gomes approval on this head.

rcoderdev:
Thanks for the careful four-eyes correction. I merged the current main branch and resolved the Cosmos sequence-mismatch integration against the standardized broadcast result contract. Core broadcast tests (15 files / 188 tests), SDK unit tests (220 files / 3,419 tests), the full check:ci suite, and the built-package runtime probe all pass on f4245ffc9. The PR is intentionally draft while current-head checks settle, and NeOMakinG remains requested for the required non-gomes approval.

gomesalexandre (commented):
Current-head rereview on `8ac63824`.

Comment only, not approve.

I re-grounded the new head delta after the earlier review checkpoints and do not see a fresh correctness or fund-safety blocker in the current changes:

- `packages/core/chain/tx/broadcast/resolvers/polkadot.ts` now routes a malformed success-shaped response with no `result` through `verifyBroadcastByHash(...)` before failing, which is the same safer verify-before-classify pattern already used on the neighboring ambiguous branches.
- `packages/core/chain/tx/broadcast/resolvers/ton.ts` now distinguishes `response present but hash missing` from transport-level request failure, and correctly pins the missing-hash path non-retryable while preserving retryability classification for genuine transient request errors.
- `packages/core/chain/tx/broadcast/resolvers/qbtc.ts` now wraps non-ok HTTP responses in `HttpResponseError` and classifies retryability through the shared `isRetryableBroadcastCause(...)` logic instead of the ad-hoc status check, which keeps the 408/429/5xx behavior aligned with the rest of the broadcast stack while preserving the provider body/status for diagnostics.
- the matching tests on this head exercise those exact new branches (`polkadot.test.ts`, `qbtc.test.ts`, `resultContract.test.ts`, plus the focused cosmos expectation tightening), so I do not see an unpinned new edge in the current head move itself.

So the code diff on this head looks sound to me.

Four-eyes gate still blocks a gomes stamp here because the branch contains gomes-authored substantive commit `45466cf68eaa949836e0b389280a5ac0dc50ed0a`. This therefore still needs a non-gomes current-head approval before merge.

rcoderdev:
Thanks for the current-head rereview. I agree that this review does not indicate a code change. The existing NeOMakinG review request remains active so the required non-gomes current-head approval can satisfy the four-eyes gate.

johnnyluo (approved):
LGTM

PR #1892, opened 2026-08-12
fix(sdk): centralize UTXO address brand validation

rcoderdev:
## Summary by CodeRabbit

* **New Features**
  * Added address-brand validation for Bitcoin, Litecoin, Dogecoin, Dash, Bitcoin Cash, and Zcash.
  * Added SDK and React Native helpers for validating addresses and detecting wrong-chain addresses.
  * Transaction decoding now rejects wrong-chain addresses and unsupported shielded Zcash addresses.

* **Bug Fixes**
  * Improved handling of malformed, invalid-checksum, and cross-chain addresses with clearer validation errors.
  * Standardized address-format validation across supported UTXO chains.
  * Improved normalization of valid uppercase CashAddr addresses while continuing to reject mixed-case input.

### DIFF
--- a/.changeset/canonical-utxo-address-brand.md
+++ b/.changeset/canonical-utxo-address-brand.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Add a canonical UTXO address-brand validator and enforce it in the transaction decoder.
--- a/packages/sdk/src/chains/utxo/addressBrand.ts
+++ b/packages/sdk/src/chains/utxo/addressBrand.ts
@@ -1,0 +1,81 @@
+import { bech32, bech32m } from '@scure/base'
+import bs58check from 'bs58check'
+
+import { isValidCashAddr } from '../../utils/cashaddr'
+
+export type UtxoChainName = 'Bitcoin' | 'Litecoin' | 'Dogecoin' | 'Dash' | 'Bitcoin-Cash' | 'Zcash'
+
+const BASE58_VERSION_BYTES: Partial<Record<UtxoChainName, ReadonlySet<number>>> = {
+  Bitcoin: new Set([0x00, 0x05]),
+  Litecoin: new Set([0x30, 0x32]),
+  Dogecoin: new Set([0x1e, 0x16]),
+  Dash: new Set([0x4c, 0x10]),
+}
+
+const BECH32_HRP: Partial<Record<UtxoChainName, string>> = {
+  Bitcoin: 'bc',
+  Litecoin: 'ltc',
+}
+
+function hasExpectedBech32Hrp(address: string, expectedHrp: string): boolean {
+  if (address !== address.toLowerCase()) return false
+
+  for (const codec of [bech32, bech32m]) {
+    try {
+      if (codec.decode(address as `${string}1${string}`).prefix === expectedHrp) return true
+    } catch {
+      // Try the other checksum variant, then the non-bech32 formats below.
+    }
+  }
+
+  return false
+}
+
+function hasExpectedBase58Version(address: string, chain: UtxoChainName): boolean {
+  let decoded: Uint8Array
+  try {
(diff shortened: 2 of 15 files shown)


gomesalexandre (commented):
Remote review on current head `a8b50202a68677af6f8770bff8f6ec72978be26d` looks mostly good. This closes a real wrong-chain-brand gap in the SDK decoder and also fixes two real false-negatives in the generic validator: Dash `7...` P2SH and Dogecoin `9...` P2SH addresses were previously rejected by the regex-only gate and are now accepted via the canonical version-byte tables. CI on the current head is green.

One non-blocking thing worth tightening before merge:

- `isUtxoAddressBrandValid` now only recognizes transparent Zcash addresses (`t1` / `t3`). The old `isAddressValidForChain(..., 'zcash')` path accepted both transparent and shielded `zs1...` shapes via `reZcashZ`, but the new canonical validator has no Zcash shielded branch, so that public SDK gate now rejects `zs1...` addresses.
- That creates an internal inconsistency in the same module: `classifyAddress` still tags `zs1...` as `family: 'zcash'`, while `isAddressValidForChain(addr, 'zcash')` now says the same address is invalid for Zcash.
- I do not see a fund-loss risk here because the change is fail-closed, not fail-open. But it is public-surface drift that is not called out in the changeset and is not pinned by a test on this PR.

If rejecting shielded Zcash addresses is intentional because the canonical tx-builder brand guard is meant to model only signable transparent UTXO surfaces, please say that explicitly in the changeset and add a regression test so the narrower contract is documented. If it is not intentional, worth restoring shielded recognition in the generic validation path or otherwise making `classifyAddress` and `isAddressValidForChain` agree.

gomesalexandre (changes requested):
Two still-open issues on current head `12bfd4c6`.

1. `packages/sdk/src/utils/addressFormat.ts:396-429` still classifies Bech32/CashAddr with lowercase-only regexes. Valid uniform-uppercase BTC, BCH, LTC, and Zcash addresses now pass `isAddressValidForChain`, but `classifyAddress` returns `unknown` for the same strings. That leaks through `checkChainPrefix` as the inconsistent state `valid: true`, `reason: "match"`, `detectedFamily: "unknown"`. Normalize only the case-insensitive encodings in the family matchers here. Base58Check should stay case-sensitive.

2. The declaration-smoke follow-up is still missing in `scripts/quality-contracts.mjs:395-438`. The React Native emitted-bundle symbol list still omits `assertUtxoAddressBrand` and `isUtxoAddressBrandValid`, and `types-smoke.ts` still does not import `UtxoChainName` plus the helper values from the root and React Native entrypoints. The runtime exports are there, but the regression gate requested by CodeRabbit is still absent.

What I rechecked on this head and found closed:
- uniform-uppercase Bech32/CashAddr acceptance and mixed-case rejection
- uppercase Zcash handling and the stale CashAddr expectation
- SegWit witness-version/program-length plus Bech32/Bech32m semantics, including invalid BIP-350 vectors
- packed root public-entry receipt for the new helper exports

Receipts from the exact-head disposable clone:
- `./node_modules/.bin/vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/chains/utxo-address-brand.test.ts packages/sdk/tests/unit/chains/utxo-cashaddr.test.ts packages/sdk/tests/unit/chains/utxo-p2sh.test.ts packages/sdk/tests/unit/utils/addressValidation.test.ts` → 92 tests passed
- packed-consumer archive build verified root + RN declarations and runtime exports

gomesalexandre (approved):
Remote current-head pass on `7654b50c` looks good.

Grounded on this head:
- uppercase Bech32/CashAddr classification now aligns with validation while Base58 remains case-sensitive
- declaration-smoke coverage now exercises the new UTXO public API from both root and react-native entrypoints
- current-head CI is fully green

No blocking findings remain from this review pass.

PR #1893, opened 2026-08-12
fix(cli): centralize token cache config dir

### LINKED ISSUE #1646: fix(cli): tokenCache duplicates VULTISIG_CONFIG_DIR resolution and mishandles an empty override
Split out of the sdk#1168 / #1171 / #1174 conflict-resolution work (VULTISIG_CONFIG_DIR centralization).

## what it says on the box

`clients/cli/src/agent/tokenCache.ts` carries its own inline copy of the exact `VULTISIG_CONFIG_DIR` resolution that sdk#1168 centralizes into `getVultisigConfigDir()`. The file landed on main after those PRs were approved, so it re-introduced the duplication the series exists to remove.

## where

`clients/cli/src/agent/tokenCache.ts`, in `getTokenCachePath()`:

```ts
const dir = process.env.VULTISIG_CONFIG_DIR ?? join(homedir(), '.vultisig')
return join(dir, 'agent-tokens.json')
```

versus the centralized resolver in `packages/client-shared/src/config-dir.ts`:

```ts
const override = process.env.VULTISIG_CONFIG_DIR?.trim()
return override ? override : join(homedir(), '.vultisig')
```

## why it is worth closing

Beyond de-duplication, the two are not semantically identical. The inline version uses `??`, so an empty or whitespace-only `VULTISIG_CONFIG_DIR` is taken literally and the agent token cache resolves to a path relative to the working directory instead of a real config dir. The centralized resolver trims and falls back to `~/.vultisig`. So today, `VULTISIG_CONFIG_DIR=""` sends `agent-tokens.json` somewhere unintended, and every other config consumer already handles that correctly.

Low severity (needs a deliberately empty env var), no fund movement, but it is an auth-token path and the fix is a two-line substitution.

## suggested fix

Import `getVultisigConfigDir` from `@vultisig/client-shared` and use it in `getTokenCachePath()`, matching the rest of the series. Worth a test pinning that an empty/whitespace `VULTISIG_CONFIG_DIR` falls back rather than producing a relative path.

## why it is not simply folded into #1168

It was deliberately kept out of that PR's diff. Those PRs carry independent approvals from before `tokenCache.ts` existed, and it was not a merge-conflict site, so including it would have shipped unreviewed code under a stale stamp. Tracking it separately so it gets its own review.

rcoderdev:
Closes #1646

### DIFF
--- a/.changeset/token-cache-config-dir.md
+++ b/.changeset/token-cache-config-dir.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/cli': patch
+---
+
+Use the shared config directory resolver for the agent token cache so empty overrides fall back safely.
--- a/clients/cli/src/agent/__tests__/configPaths.test.ts
+++ b/clients/cli/src/agent/__tests__/configPaths.test.ts
@@ -40,10 +40,12 @@ describe('agent/shared config-dir path policy', () => {
     const fallbackDir = join(homedir(), '.vultisig')
 
     process.env[ENV_KEY] = ''
+    expect(getTokenCachePath()).toBe(join(fallbackDir, 'agent-tokens.json'))
     expect(getCredentialsPath()).toBe(join(fallbackDir, 'credentials.enc'))
     expect(journalPath()).toBe(join(fallbackDir, 'broadcasts.jsonl'))
 
     process.env[ENV_KEY] = '   '
+    expect(getTokenCachePath()).toBe(join(fallbackDir, 'agent-tokens.json'))
     expect(getCredentialsPath()).toBe(join(fallbackDir, 'credentials.enc'))
     expect(journalPath()).toBe(join(fallbackDir, 'broadcasts.jsonl'))
   })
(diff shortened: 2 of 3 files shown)


gomesalexandre (approved):
approve. remote conceptual review on current head `cbf58cc2540955a654521c877c271d37694459f8` looks clean.

receipts
- verified the 3-file diff only centralizes `getTokenCachePath()` onto `getVultisigConfigDir()` from `@vultisig/client-shared`, matching the existing resolver already used by credentials and the broadcast journal
- verified the bug shape is real: empty or whitespace `VULTISIG_CONFIG_DIR` previously flowed through the token cache path via `??`, unlike the shared resolver's trim-and-fallback behavior
- verified `getVultisigConfigDir()` is already exported by published `@vultisig/client-shared` in current released tags, so this does not introduce a hidden runtime compatibility edge for the CLI package
- verified the focused test actually pins the intended fix by asserting `getTokenCachePath()` for both empty-string and whitespace overrides alongside the other shared-config-path consumers
- CI on the current head is green across lint/type/unit/integration

narrow correctness cleanup, no fresh blocker found.

PR #1894, opened 2026-08-12
fix: unify queryUrl timeout defaults, pin lifi solana chain id

### LINKED ISSUE #1374: Collapse duplicate queryUrl exports (20s vs no-timeout) and pin the hardcoded LiFi Solana chain id
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P3 CONFIRMED/SUSPECTED.

**Anchor:** lib/utils/query/queryUrl.ts:22,44 vs platforms/react-native/rpcFetch.ts:75-82 - platforms/react-native/overrides/lifiSwapEnabledChains.ts:29

## what
Two exported queryUrl functions exist with different timeout behavior: lib/utils/query/queryUrl.ts defaults to a 20s AbortController-based timeout when the caller passes none; the RN-only platforms/react-native/rpcFetch.ts queryUrl has NO built-in timeout at all (bare fetch, timeout only if the caller supplies its own signal). Import-the-wrong-one is a footgun, and the RN gap is worse than mismatched defaults - it can hang indefinitely. (The audit round originally cited "30s vs 20s"; verified against main, the RN one actually has no timeout at all.) Separately, react-native/overrides/lifiSwapEnabledChains.ts:29 hardcodes the numeric LiFi Solana chain id (1151111081099710) where the non-RN module resolves it from the live @lifi/sdk ChainId enum.

## fix
Give the RN queryUrl the same default-timeout behavior as the base one (or explicitly document why RN opts out). For the LiFi id: the hardcode is a deliberate, documented workaround (importing @lifi/sdk's barrel crashes Hermes via a missing Event global at module-init) - don't "fix" it by importing the enum on RN. Instead add a small non-RN unit test asserting LIFI_CHAIN_ID_SOL matches ChainId.SOL from @lifi/sdk, so a LiFi renumber fails CI instead of drifting silently.

gomesalexandre:
closes vultisig/vultisig-sdk#1374

## what
Two exported `queryUrl` functions with different timeout defaults (20s shared vs 30s RN) is a footgun - import the wrong one and you get a different hang budget. Collapsed to one source of truth: RN's `fetchWithTimeout.ts` now imports `DEFAULT_QUERY_TIMEOUT_MS` from the shared `lib-utils/query/queryUrl` module instead of hardcoding its own `30_000`, so the two can't drift apart again silently.

Separately, `platforms/react-native/overrides/lifiSwapEnabledChains.ts` hardcodes the numeric LiFi Solana chain id as a documented Hermes-safety workaround (importing `@lifi/sdk`'s barrel crashes Hermes via a missing `Event` global at module-init). Per the issue, that workaround stays - "fixing" it by importing the enum on RN would reintroduce the crash. Instead exported the constant and added a non-RN unit test pinning it against the live `ChainId.SOL` from `@lifi/sdk`, so a LiFi renumber fails CI instead of drifting silently.

## how
- `packages/lib/utils/query/queryUrl.ts`: export `DEFAULT_QUERY_TIMEOUT_MS` (was module-private)
- `packages/sdk/src/platforms/react-native/fetchWithTimeout.ts`: `DEFAULT_RN_FETCH_TIMEOUT_MS` now derives from the shared constant instead of a separate `30_000` literal
- `packages/sdk/src/platforms/react-native/overrides/lifiSwapEnabledChains.ts`: export `LIFI_CHAIN_ID_SOL`
- new test `packages/sdk/tests/unit/platforms/react-native/lifi-chain-id.test.ts` asserts `LIFI_CHAIN_ID_SOL === ChainId.SOL` (real `@lifi/sdk`, safe under Node - only Hermes/RN lacks the `Event` global)
- `rpc-fetch-timeout.test.ts`: added regression asserting `DEFAULT_RN_FETCH_TIMEOUT_MS === DEFAULT_QUERY_TIMEOUT_MS`

## testing
- `yarn workspace @vultisig/sdk test` - 3327 passed
- `yarn test:core` - 2685 passed, 1 skipped
- `yarn workspace @vultisig/sdk typecheck` - clean
- `tsc --project .config/tsconfig.json --noEmit` (root, covers packages/lib + packages/core) - clean
- targeted eslint on changed files - clean

Risk: low, timeout default only tightens from 30s→20s for RN callers that don't pass their own signal/timeoutMs; no behavior change to LiFi's shipped chain id, only adds an export + a CI tripwire.


gomesalexandre (commented):
remote conceptual re-check on current head `5f2e614757cfdae8bdb4564938e0c6f7bf926b0c` only.

re-read the actual head diff, not just the PR text. no fresh blocker on the current delta.

grounded bits i specifically re-checked:
- RN now derives its default query timeout from the shared `DEFAULT_QUERY_TIMEOUT_MS` instead of carrying an independent `30000` literal, so the default is centralized rather than silently drifting.
- the LiFi Solana chain-id workaround is pinned with a real regression instead of a comment-only assumption.
- the surface is additive/export-and-test only; i do not see a hidden behavior change outside the intended timeout consolidation + chain-id pin.

NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because This head unintentionally changes Fast Vault signing and relay defaults from the documented 30 seconds to 20 seconds, and its LiFi drift test compares against a globally mocked enum. Re-propose the query-only timeout change with a real dependency test.

Issue #1157, opened 2026-07-10
arch: EVM gas-price floor tables triplicated across app/abts, no SDK owner


gomesalexandre:
Per-chain minimum priority-fee/max-fee gas floors (guards against near-zero baseFee on quiet blocks on BSC/Polygon/Arbitrum/Optimism/Base/Mantle/Sei/Hyperliquid) are hand-maintained separately in `vultiagent-poc/src/services/evmTx.ts` (`MIN_PRIORITY_FEE_BY_CHAIN`/`MIN_MAX_FEE_BY_CHAIN`) and `agent-backend-ts/src/mastra/tools/mcp/tools/swap/evm-tx-info.ts` (`GAS_FLOORS_WEI`), hand-synced by comment convention. Both cite a real prior incident (BSC swap failures, vultiagent-app#567; "VA-90 audit, 2026-07-08"). No SDK owner exists.

Fix sketch + full agent-ready prompt: [round-5 architecture artifact](https://claude.ai/code/artifact/13bddb9f-4c87-440a-aae8-e30356f5a1a1), finding "EVM gas-price floors triplicated."

Severity: HIGH (real prior incident, silent-drift-prone). Effort: M.

PR #1895, opened 2026-08-12
feat: export evm gas-price floor tables

gomesalexandre:
tackles vultisig/vultisig-sdk#1157

Scope of this PR is sdk-only. Full resolution of #1157 requires migrating vultiagent-app's MIN_PRIORITY_FEE_BY_CHAIN/MIN_MAX_FEE_BY_CHAIN and agent-backend-ts's GAS_FLOORS_WEI onto this table, which is a separate cross-repo effort not attempted here.

## what
`clampEvmPriorityFee.ts` already had a per-chain EVM priority-fee floor/ceiling policy (real production incidents: BSC #567, Ethereum quiet-market zero-tip), but only the `clampEvmPriorityFee` function was exported - the underlying tables were private, so app and agent-backend-ts had no way to read the canonical values and instead hand-rolled their own copies (`MIN_PRIORITY_FEE_BY_CHAIN`/`MIN_MAX_FEE_BY_CHAIN`, `GAS_FLOORS_WEI`), synced by comment convention.

Exported `priorityFeeFloorWeiByChain`/`priorityFeeCeilingWeiByChain` as `evmPriorityFeeFloorWeiByChain`/`evmPriorityFeeCeilingWeiByChain` from both the root and react-native sdk entry points, so this table CAN become the single source of truth once the app/abts side is migrated.

## why
Same silent-drift risk the issue calls out: three copies of a fund-safety-adjacent gas floor table with no shared owner. Exporting the table is the prerequisite - can't point downstream consumers at it if it's not reachable.

Did not touch the actual floor values. The sdk's table currently floors only Ethereum/Polygon (tip-auction chains); the app/abts copies additionally floor several L2 rollups (Arbitrum/Optimism/Base/Mantle/Sei/Hyperliquid), citing a documented incident (agent-backend-ts issue #236: zero-tip txs stuck unbroadcastable on some sequencers) that conflicts with the sdk's own tested assumption ("L2 sequencers include txs regardless of tip, so rollups deliberately have no floor"). Reconciling that is real migration-scope work requiring verification I can't safely do from this repo alone - flagging it here for whoever picks up the migration.

## testing
- `yarn build:shared && yarn build:all` - clean
- `tsc --noEmit` (packages/sdk) - clean
- New tests: table shape (exhaustive ceiling coverage, floor subset), documented values, public-importability from both the root and RN entry points
- `yarn vitest run` on the touched test files - 33 + 96 passed
- Project eslint script on all changed files - clean

Risk: Low - additive export only, no behavior change to `clampEvmPriorityFee` or any existing caller.


gomesalexandre (commented):
remote conceptual re-check on current head `bd46385412189430637b06582287dc3325306939` only.

no fresh blocker here. this is a clean additive export of the existing EVM priority-fee floor/ceiling tables from the root + RN SDK entrypoints, with the tables wrapped `Readonly` and no behavior change to `clampEvmPriorityFee` itself.

what i re-checked:
- the diff is export-shape only, not a logic rewrite of the fee clamp path.
- the new tests pin both names and values, which is the main contract risk on a pure-surface PR like this.
- i do not see another in-repo caller mutating these tables, so the `Readonly` tightening is safe.

NeOMakinG (approved):
LGTM

rcoderdev (commented):
The exported floor/ceiling tables are the same runtime objects used by clampEvmPriorityFee. TypeScript Readonly is erased, so a consumer can mutate an exported value and change later fee clamping globally, including the compromised-RPC ceiling. Please freeze the policy objects at runtime or expose immutable copies backed by private frozen state, and add a mutation regression.

rcoderdev:
Closing this because The exported tables are the same mutable runtime objects used by clampEvmPriorityFee; TypeScript Readonly does not stop consumers from mutating global fee policy. Re-propose frozen state or immutable copies with a mutation regression.

PR #1896, opened 2026-08-12
fix(core-mpc): recap native-max sends on the hexPublicKeyOverride path

### LINKED ISSUE #1390: core-mpc: native-max fee re-cap (refineKeysignAmount/refineKeysignUtxo) skipped when signing via hexPublicKeyOverride (publicKey = null)
## Problem

`buildSendKeysignPayload` gates both keysign refinement steps behind `if (publicKey)`:

[packages/core/mpc/keysign/send/build.ts#L124-L135](https://github.com/vultisig/vultisig-sdk/blob/38f849c4/packages/core/mpc/keysign/send/build.ts#L124-L135) — `refineKeysignAmount` (re-caps a native max send to `min(toAmount, freshBalance − freshFee)`) and `refineKeysignUtxo` (max-sweep + dust handling) only run when a WalletCore `PublicKey` is provided.

When callers sign via `hexPublicKeyOverride` with `publicKey = null` (the MLDSA-only / post-quantum path the override exists for), both refinements are silently skipped:

- **Native max:** ships `balance − staleFeeEstimate`; a fee increase between estimate and broadcast overshoots and the tx fails on-chain. This safety net is precisely what makes native max robust on the classic path — the PQ path loses it.
- **UTXO:** no `sendMaxAmount` flip / dust-stat handling.

## Suggested fix

Make the refine steps work on the override path (they mainly need the address/balance, not the WalletCore key object — thread what they need), or explicitly document that override callers must re-cap themselves. Silent skipping is the worst option: the caller believes the safety net exists.

Found during a cross-platform "max send" audit (consumed by vultisig-windows desktop/extension via @vultisig/core-mpc).

gomesalexandre:
closes vultisig/vultisig-sdk#1390

## what

`buildSendKeysignPayload` gated `refineKeysignAmount`/`refineKeysignUtxo` (the native-max fee re-cap and UTXO max-sweep/dust handling) behind `if (publicKey)`. Signing via `hexPublicKeyOverride` (`publicKey = null`, the MLDSA-only / post-quantum path) silently skipped both. A native-max send on that path would ship `balance - staleFeeEstimate` instead of the fee the payload actually carries, risking an overshoot the node rejects on broadcast - or worse, spending past the true spendable balance.

## how

Neither refine step actually reads the `publicKey` object - `refineKeysignAmount` -> `getFeeAmount` -> the per-chain fee resolvers, and `refineKeysignUtxo` -> `getUtxoSigningInputs`, all derive what they need from `keysignPayload`/`balance`. `keysignPayload.coin.hexPublicKey` is already populated regardless of which path resolved the key (line 55 of `build.ts`), and the codebase already has a `getKeysignTwPublicKey` helper elsewhere that reads the hex straight off the payload instead of taking a WalletCore `PublicKey`.

So the fix threads `publicKey: PublicKey | null` through `refine/amount.ts`, `refine/utxo.ts`, `fee/resolver.ts`, `fee/index.ts`, and the utxo branch of `signingInputs/resolver.ts`, and drops the `if (publicKey)` gate in `send/build.ts` entirely - the recap now runs unconditionally, the same way for both signing-key-resolution paths.

## testing

- New regression tests in `send/build.test.ts`: a native-max EVM send via `hexPublicKeyOverride` now gets recapped identically to the same send via the normal `publicKey` path.
- Existing Ripple `hexPublicKeyOverride` suite in the same file still passes (recap now actually runs there too, no behavior break).
- `yarn vitest run packages/core/mpc/keysign` - same 16 pre-existing failures on `origin/main` before this change (unrelated: real-network cosmos queries + a missing walletCore mock in `buildLimitSwapKeysignPayload.test.ts`/`buildLimitSwapCancelKeysignPayload.test.ts`), confirmed via stash-and-rerun on baseline.
- `tsc --project .config/tsconfig.json --noEmit` and `yarn workspace @vultisig/sdk typecheck` clean.
- eslint clean on all changed files.

## risk

Low-medium, fund-safety fix. Widens an existing safety net to a path that was silently missing it; doesn't change behavior on the already-covered `publicKey` path (verified via the new identical-recap regression test).

🤖 Agent-generated


gomesalexandre (commented):
remote conceptual re-check on current head `112e3b56263a65a091510f682dafaa931c95a112` only.

no fresh blocker. this closes a real fund-safety gap on the `hexPublicKeyOverride` path by removing the `if (publicKey)` gate that previously skipped native-max recap / UTXO dust-sweep handling for MLDSA-only signing.

what i re-checked:
- the widened resolver input is `PublicKey | null`, but the downstream fee + UTXO resolver paths do not actually dereference `publicKey`, so dropping the gate is behavior-neutral on the already-covered path.
- the current regression proves parity across the normal key-resolution path and the override path instead of only testing one arm.

NeOMakinG (approved):
LGTM

rcoderdev (commented):
The implementation looks coherent, but this published core-mpc fund-safety fix is missing both the required patch changeset and the QBTC/on-chain amount-and-fee comparison receipt. The synthetic EVM parity tests do not exercise the actual null-key caller. Please add the changeset and the required exact-head QBTC receipt before this is treated as review-complete.

rcoderdev:
Closing this because This incomplete fund-safety branch changes published core-mpc behavior without a changeset and does not prove the real QBTC/null-key path with an on-chain amount/fee receipt. The synthetic EVM parity test is insufficient for this signing-path change.

Issue #1500, opened 2026-07-22
fix(swap): co-signer swap guards still key on the untrusted provider string - needs a proto oneof discriminant (follow-up to #1462)


gomesalexandre:
Follow-up to #1462 (which partially closed #1457). Part of the swap correctness + broadcast safety epic (#1356).

## what #1462 did (good, keep)
Closed the "relabel to ANY string to skip enforcement" hole: the log-only fallback for unenforced providers changed from "accept any string" to a closed list `{li.fi, swapkit, ''}` (anything else now throws), and CowSwap was added to the address-enforced router allowlist (correctly chain-scoped to `cowSwapSupportedChains`). Solid incremental hardening.

## residual (why #1457 isn't fully closed)
The co-signer swap guards still key on the untrusted `provider` wire string, not a trusted discriminant. `li.fi`/`swapkit`/`''` remain enforced-by-closed-list-only (not address-bound), because those route through many contracts by design. So a payload can still relabel itself to `li.fi`/`swapkit`/`''` and skip address enforcement. This is a monotonic degrade to the pre-#1358 posture (not a new bypass beyond it), but it means the co-signer protection is only as strong as an attacker-controllable string.

## fix
Make the executing shape un-spoofable: a proto `oneof`-per-provider (or a server-derived discriminant bound into the signed payload) so the signing side can't be relabeled past the guard. This is the cross-repo schema change explicitly deferred from #1462.

Severity: P2 (fund-safety, monotonic degrade). Found in the R7 daily burn-down audit; the partial fix (#1462) landed 2026-07-21.

PR #1897, opened 2026-08-12
fix(swap): close the legacy unattributed provider bypass on co-signer swap guards

gomesalexandre:
tackles vultisig/vultisig-sdk#1500

## what

Follow-up to #1462 and #1830, which structurally closed provider-spoofing gaps on the co-signer swap guards (`assertKnownAggregatorRouterOnSigningPath`, `assertEnforcedSwapApprovalSpenderBound`): CowSwap and LI.FI are now enforced by known router address, SwapKit destinations/spenders require an independent Blockaid reputation verdict, and the log-only fallback shrank to a closed list of provider strings.

One bare bypass remained: `''` (the legacy unattributed provider, kept for backward compat with pre-provider-field mobile fixtures) was STILL log-only regardless of `address` — a payload could relabel ANY malicious swap-leg destination or approval spender as `''` and skip every check the other providers now go through. The pre-fix regression test in this file literally asserted `assertKnownAggregatorRouterOnSigningPath('', ATTACKER_ADDRESS, Chain.Ethereum)` did NOT throw.

## why

fund-safety: a payload whose provider label disagrees with its actual executing shape must be rejected before signing, not trusted at face value - `''` shouldn't be a free pass any more than any other string.

## fix

Verified against the real historical `''` fixtures (`mobileFixtures.golden.test.ts`'s `arb.json`/`lifiswap.json`, which predate the `provider` field entirely) that their destinations already ARE known enforced routers (1inch V6 router, LI.FI Diamond). So:

- `''` stays log-only ONLY when the destination/spender independently matches a known enforced aggregator router (new `isKnownAggregatorRouterAddress` helper) - real backward compat, not a bypass.
- Anything else now requires the same independent Blockaid reputation verdict SwapKit destinations already use (`assertLegacyUnattributedAddressReputation`), applied to BOTH the swap-leg destination and the ERC-20 approval spender (the two legs sdk#1358's review already established are independent wire fields that don't bind to each other).

No proto/wire-format changes - this stays scoped to `@vultisig/core-chain`/`@vultisig/core-mpc`.

## residual gap (deferred, matches #1462's original scope note)

The issue's proposed complete fix is a proto `oneof`-per-provider (a trusted wire discriminant instead of a free `provider` string on `OneInchSwapPayload`). That's a schema change on the shared `vultisig/commondata` proto, cross-repo, and every native client (iOS/Android/Windows) also builds against it - a full migration is out of scope for an SDK-only PR, same conclusion #1462 already reached. This PR closes the one concretely exploitable string-trust gap left after #1830 (the `''` bare bypass) without needing that migration; the full typed-discriminant work remains a maintainer-scheduled follow-up.

## testing

- `packages/core/chain/swap/general/knownAggregatorRouters.test.ts` + `packages/core/mpc/keysign/signingInputs/resolvers/evm/swapRouterGuard.test.ts` (scoped): 79/79 passing - updated the pre-fix regression test to reflect the closed gap, added unit coverage for the new `isKnownAggregatorRouterAddress` helper, and added end-to-end resolver coverage for both the swap-leg destination and approval-spender legs (known-router fast path with no network call, unknown-destination fail-closed on non-Benign verdict, fail-closed when Blockaid itself throws).
- `yarn test:core` (full suite): 2727 passing, 1 skipped, 0 failed - includes the golden-fixture suite (`mobileFixtures.golden.test.ts`), proving real historical `''` payloads still sign cleanly without a new network dependency.
- `tsc --project .config/tsconfig.json --noEmit`: clean, 0 errors.
- `eslint` on all four changed source/test files: 0 errors.

Changeset added (`@vultisig/core-chain`, `@vultisig/core-mpc`, `@vultisig/sdk`: minor) documenting the new runtime throw + the new Blockaid network call for unknown-router `''` destinations/spenders, mirroring #1830's changeset.

Not self-merging - this needs an external approval stamp per the usual four-eyes gate.


gomesalexandre (commented):
remote conceptual re-check on current head `006a68328cb44dd3ae2c2fb909450f42f4686eeb` only.

no fresh blocker. the current diff closes the last bare legacy-provider bypass in the co-signer swap guard: `provider: ''` now stays log-only only when the destination/spender independently matches a known aggregator router, otherwise it must clear the same independent Blockaid reputation gate as the rest of the path.

what i re-checked:
- both the swap-leg destination and approval-spender wires are covered.
- known-router traffic still avoids an unnecessary network dependency.
- unknown-router traffic now fails closed on malicious / error results instead of coasting through the empty-provider label.

gomesalexandre (commented):
current-head recheck on `549f079a`.

no fresh blocker from the merge-from-main head.

what i re-grounded:
- the new head is a pure merge-from-main carry-forward, not a conflict-resolution change in the swap-guard files. in a disposable exact-head checkout, `git show --format= --cc 549f079a | wc -l` returned `0`, so the merge commit itself has no combined-diff hunks.
- the security-relevant diff from this PR is still the same one: legacy `provider: ''` only stays log-only when the destination or approval spender already matches a known enforced router, otherwise both legs now require the independent Blockaid reputation floor.
- the independent approve-spender wire is still covered, not just `quote.tx.to`.
- current-head CI is green on `549f079a`.

i do not see a new code blocker on the current head.

four-eyes note: this branch contains substantive gomes-authored commits, so it still needs a non-gomes current-head approval before merge.

NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because The fail-closed claim still has an empty-destination bypass: missing or empty destination skips Blockaid and the router assertion, then produces a signing input with an empty toAddress. Re-propose with explicit rejection and an end-to-end regression.

PR #1898, opened 2026-08-12
feat(sdk): publish seedphrase helper subpath

rcoderdev:
## Summary by CodeRabbit

- **New Features**
  - Added the `@vultisig/sdk/seedphrase` public subpath.
  - Provides seedphrase normalization, validation, language detection, key derivation, and chain discovery helpers.
  - Supports JavaScript and TypeScript usage across supported module environments.

- **Tests**
  - Added coverage for runtime exports, TypeScript declarations, mnemonic normalization, language detection, and packaged consumer usage.

- **Documentation**
  - Documented publication of the new seedphrase subpath in the SDK release notes.

### DIFF
--- a/.changeset/public-seedphrase-subpath.md
+++ b/.changeset/public-seedphrase-subpath.md
@@ -1,5 +1,5 @@
 ---
-'@vultisig/sdk': patch
+'@vultisig/sdk': minor
 ---
 
 Publish the canonical seedphrase normalization, validation, key-derivation, and chain-discovery helpers through the supported `@vultisig/sdk/seedphrase` subpath.
--- a/.changeset/public-seedphrase-subpath.md
+++ b/.changeset/public-seedphrase-subpath.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Publish the canonical seedphrase normalization, validation, key-derivation, and chain-discovery helpers through the supported `@vultisig/sdk/seedphrase` subpath.
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -114,6 +114,16 @@
       "import": "./dist/tools/bridge/index.js",
       "require": "./dist/tools/bridge/index.cjs",
       "default": "./dist/tools/bridge/index.cjs"
+    },
+    "./seedphrase": {
+      "types": "./dist/seedphrase/index.d.ts",
+      "node": {
+        "import": "./dist/seedphrase/index.js",
+        "require": "./dist/seedphrase/index.cjs"
+      },
+      "import": "./dist/seedphrase/index.js",
+      "require": "./dist/seedphrase/index.cjs",
+      "default": "./dist/seedphrase/index.cjs"
     }
   },
   "scripts": {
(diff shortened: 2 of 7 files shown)


gomesalexandre (commented):
Remote-only re-check on current head `cd0af9da2a3017808c4b4a22e60ea10d63cd951c`.

- `.changeset/public-seedphrase-subpath.md` is now correctly `minor`, not `patch`, for the new public `@vultisig/sdk/seedphrase` surface.
- `packages/sdk/package.json`, `packages/sdk/rollup.platforms.config.js`, and `packages/sdk/rollup.types.config.js` all wire the new `./seedphrase` subpath consistently with the existing subpath export pattern.
- the new `packages/sdk/tests/unit/public-api/seedphraseSubpathExports.test.ts` and `scripts/quality-contracts.mjs` additions pin both the export-map contract and the packed-consumer smoke path.
- current-head CI is green on `cd0af9d` for Lint and Type Check, Unit Tests, Integration Tests, and Test Summary.

I do not see a fresh correctness or public-surface blocker on the current diff from this remote conceptual pass.

Process note only: this head contains a substantive `gomesalexandre` commit (`d4dd2e6cf68458f486bc80e3a8e8ecfa7d000edd`), so four-eyes forbids my approval here. This needs a fresh non-gomes current-head approval before merge.

gomesalexandre (approved):
Packaging-only change (`src/` is untouched), and the verification approach is the right one for a publish PR: `quality-contracts.mjs` resolves the real published entry and asserts it lands on `index.cjs`, then exercises both the ESM and CommonJS paths. Testing the packaged artifact rather than the source is exactly what catches an exports-map that typechecks but does not resolve for a consumer.

One thing to fix before this ships, because publishing changes what it means.

## should-fix: the published usage example logs private keys

This subpath is **new public surface**, not a re-export of something already reachable. None of it is in the root index today:

```
MasterKeyDeriver         root index hits: 0
SeedphraseValidator      root index hits: 0
normalizeMnemonic        root index hits: 0
ChainDiscoveryService    root index hits: 0
```

So `@vultisig/sdk/seedphrase` becomes the supported, semver-committed way to derive master keys from a mnemonic. That is a reasonable thing to offer. It also means the class's `@example` block becomes the canonical snippet consumers copy, and right now it is:

```ts
// MasterKeyDeriver.ts:73-79
const deriver = new MasterKeyDeriver(wasmProvider)
const keys = await deriver.deriveMasterKeys(mnemonic)
console.log(keys.ecdsaPrivateKeyHex) // For DKLS import
console.log(keys.eddsaPrivateKeyHex) // For Schnorr import (already clamped)
```

It is a doc comment, so nothing executes it - I checked before raising it. That is also the problem: the first thing an integrator does with a new subpath is paste the example, and this one prints raw private keys to stdout, where they land in terminal scrollback, CI logs, crash reporters and support tickets. It was defensible while this was internal. As the advertised entry point for key derivation it teaches the wrong habit at exactly the moment someone is most likely to copy it.

Cheap fix, no API change: have the example pass the keys somewhere instead of printing them, or print something non-sensitive, e.g.

```ts
await importIntoVault(keys.ecdsaPrivateKeyHex)   // never log key material
```

## nit

`ChainDiscoveryService.ts:190` does `console.warn('Failed to check Phantom Solana path:', error)` on a caught derivation failure. It logs the error object, not the mnemonic, and derivation errors do not normally embed their input, so I am not treating it as a leak. Worth a glance only because it is on the same newly-public path and a library warning straight to `console` is a slightly odd default for consumers who have their own logger.

Nice touch using the standard BIP39 `abandon … about` vector in the contract test rather than inventing a mnemonic that someone might later mistake for real.

```
CI   all 5 checks green
```

PR #1901, opened 2026-08-12
feat: export isEddsaChain/getSignatureAlgorithm from SDK root

gomesalexandre:
Companion to vultisig/agent-backend-ts#2641.

## what

Exports two small helpers from the SDK root, colocated with the existing `getChainKind`/`isChainOfKind` exports:

- `getSignatureAlgorithm(chain)` — returns `'ecdsa' | 'eddsa' | 'mldsa'`
- `isEddsaChain(chain)` — `getSignatureAlgorithm(chain) === 'eddsa'`

## why

agent-backend-ts issue #2641: three files there (`execute_swap.ts`, `get-address.ts`, `get-receive-info.ts`) each hand-copy the same `EDDSA_CHAIN_KINDS` set + `chainNeedsEddsa`/`isEddsaChain` local helper, gated on `getChainKind()` from `@vultisig/sdk`. The actual ECDSA-vs-EdDSA decision was already computed internally in the SDK (`signatureAlgorithms` record in `core-chain/signing/SignatureAlgorithm.ts`, and `ChainDiscoveryService.isEddsaChain`) but never exposed publicly, so consumers re-invent it instead of importing it. Same root cause/fix shape as the earlier `getChainKind`/`isChainOfKind` export.

## how

- `core-chain/signing/SignatureAlgorithm.ts`: add a standalone `isEddsaChain(chain)` next to `getSignatureAlgorithm`
- `sdk/src/index.ts`: export both from the SDK root
- `ChainDiscoveryService.isEddsaChain` now delegates to the new free function instead of duplicating the `getSignatureAlgorithm(...) === 'eddsa'` check

Changeset included (patch bump on `@vultisig/sdk`).

## testing

- `npx tsc --noEmit -p .config/tsconfig.json` clean
- `yarn workspace @vultisig/sdk test:unit` — 3334 passed, including `ChainDiscoveryService.test.ts`'s `isEddsaChain` cases
- Verified `isEddsaChain`/`getSignatureAlgorithm` present in built `packages/sdk/dist/index.d.ts`

Follow-up once this merges + publishes: bump `@vultisig/sdk` in agent-backend-ts and swap the local dedup (opened separately, see abts PR) over to this export.


gomesalexandre (commented):
current-head remote conceptual pass on `d835291` only.

what i re-grounded:
- `isEddsaChain` in `packages/sdk/src/utils/signing/SignatureAlgorithm.ts` is a straight lift of the exact `getSignatureAlgorithm(chain) === 'eddsa'` expression that `ChainDiscoveryService.isEddsaChain` previously inlined, so the delegation is behavior-preserving.
- `ChainDiscoveryService.isEddsaChain` now just delegates to that helper instead of duplicating the check.
- the root export in `packages/sdk/src/index.ts` follows the same colocated type+value export pattern already used for nearby chain helpers like `getChainKind` / `isChainOfKind`.
- current-head CI is green.

no blocker from me on the current diff.

process note only: this is a gomes-authored head, so it still needs a non-gomes current-head approval before merge.

NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this because The exact head has no green unit-gate receipt and predates substantial mainline drift, while the new root exports have no direct public-entry regression. Please re-propose the small helper export from current main with green exact-head CI and export coverage.

PR #1909, opened 2026-08-13
fix(cli): resolve --max to a numeric amount in JSON output

neavra:
## Summary

- resolve max send dry-run JSON amounts from the prepared signable payload using the coin decimals
- resolve max swap-quote JSON amounts from maxSwapable using the source coin decimals
- preserve non-max amounts, human output, and executed-send JSON while retaining isMax as the machine-readable flag
- add a patch changeset for @vultisig/cli

## Verification

- revert-check: the new send and swap-quote max regressions failed before the implementation with amount equal to max
- yarn test:cli --maxWorkers=2 (79 files, 1181 tests passed)
- yarn lint
- yarn typecheck
- yarn quality:complexity
- yarn format:check
- yarn build:all
- yarn test
- bundled CLI live backend dry-run: send Ethereum to the active vault own address with --max --dry-run --output json returned success true, a resolved decimal amount, isMax true, and dryRun true
- executed max-send regression confirms its JSON still omits amount and isMax

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected JSON output for maximum send and swap amounts by resolving values using the asset’s decimal precision.
  * Added an `isMax` indicator for dry-run results when the maximum amount is selected.
  * Preserved clean transaction output without duplicate or unformatted amount fields.
  * Kept non-maximum amount formatting and human-readable previews unchanged.

### DIFF
--- a/.changeset/bright-pandas-resolve.md
+++ b/.changeset/bright-pandas-resolve.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/cli': patch
+---
+
+Resolve max send and swap quote amounts to numeric strings in JSON output.
--- a/clients/cli/src/commands/__tests__/sendDryRunFee.test.ts
+++ b/clients/cli/src/commands/__tests__/sendDryRunFee.test.ts
@@ -26,6 +26,7 @@ function makeVault(opts: {
   fee: string
   total: string
   balance: string
+  resolvedAmount?: string
   payloadMemo?: string
   payloadDestinationTag?: number
 }) {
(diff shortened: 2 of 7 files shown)


gomesalexandre (approved):
looks good. grounded this on the exact head in a disposable clone.

receipts:
- `yarn build:shared && yarn build:sdk:bundle && yarn build:client-shared && yarn cli:build` clean in the temp clone
- focused cli suites green after building the sdk artifacts:
  - `yarn vitest run clients/cli/src/commands/swapAmount.test.ts clients/cli/src/commands/__tests__/sendDryRunFee.test.ts clients/cli/src/commands/__tests__/sendSwapDupGuard.test.ts`
  - `63 passed`
- diff does the right contract change on both paths:
  - `send --dry-run --json --amount max` now emits the resolved numeric `amount` plus `isMax: true`
  - `swap quote --json --amount max` now emits the resolved numeric amount instead of the literal `max`
- human preview stays unchanged for max sends (`Amount:  max ETH` control stayed pinned)

no fresh blocker from this head.

PR #1927, opened 2026-08-14
fix(decode): reject malformed payload encodings

### LINKED ISSUE #1727: fix(decode): decodeFromToolResult coerces malformed hex/base64 instead of failing closed
## Summary

`decodeFromToolResult` silently coerces malformed hex/base64 payload strings instead of failing closed. Because it delegates straight to Node `Buffer.from(..., 'hex'|'base64')`, malformed payload text can be truncated or normalized into different bytes before the shared tx-decoder inspects it.

Report: https://gist.github.com/gomesalexandre/ee18a0bd3457f4a5d34d8fa319b6f2ee

## Evidence

- `packages/sdk/src/tools/decode/fromToolResult.ts:49-57`
  - `hexToBytes()` and `base64ToBytes()` delegated directly to `Buffer.from`
- `packages/sdk/src/tools/decode/fromToolResult.ts:70-81`
  - `resolvePayload()` trusted those decoders for explicit `payload` and tool-call args
- `packages/sdk/src/tools/decode/fromToolResult.ts:85-94`
  - `tryBase64ThenHex()` assumed malformed hex would throw, but Node's hex/base64 decoding is lenient

Concrete repros from the current code:
- `0x123 -> 12`
- `123 -> 12`
- `zz -> <empty>`
- `AQ=== -> 01`

## Impact

This function is the shared tx-bytes oracle behind WYSIWYS / policy / decoded-intent surfaces. A malformed or truncated payload string can be inspected as *different bytes* than the caller supplied, which defeats the whole point of having one canonical decoder boundary.

## Fix sketch

- make hex parsing strict: non-empty, even-length, full hex alphabet, round-trip exactness
- make base64 parsing strict: canonical base64 only, reject junk / extra padding
- keep the existing contract: no throws, return `decoded:false` with a parse error instead
- add regression tests for malformed explicit payloads and args payloads

## Effort

Small / low-risk. Pure decoder hardening plus tests.

rcoderdev:
Closes #1727


gomesalexandre (approved):
Looks good.

Receipts:
- exact-head diff inspected against `origin/main`
- `yarn workspace @vultisig/sdk test:unit packages/sdk/tests/unit/decode/fromToolResult.test.ts` on this checkout

Why this is safe:
- malformed hex now fails closed instead of letting `Buffer.from(..., 'hex')` silently normalize odd-length / non-hex payloads
- malformed base64 now rejects non-canonical encodings instead of accepting Node's tolerant normalization
- the new `withFamilyError(...)` path preserves the existing `decoded:false` envelope contract while keeping family-specific errors on both direct payload and args-based decode paths
- coverage pins both explicit payload and args payload regressions for evm + cosmos

PR #1928, opened 2026-08-14
fix(sdk): publish tx and decode subpaths

rcoderdev:
## Summary by CodeRabbit

- **New Features**
  - Added dedicated SDK entry points for transaction utilities and transaction decoding.
  - Added JavaScript and TypeScript declaration bundles for these public subpaths.
  - Enabled platform-specific imports across browser, worker, React Native, and Node environments.
- **Bug Fixes**
  - Improved compatibility and reliability when importing transaction and decoding utilities.
  - Added validation to help ensure consistent runtime behavior and TypeScript support across supported environments.


gomesalexandre (approved):
Looks good.

Receipts:
- exact-head diff inspected against `origin/main`
- `./node_modules/.bin/vitest run packages/sdk/tests/unit/public-api/toolsSubpathExports.test.ts`
- `node scripts/smoke-sdk-subpath-exports.mjs`

Why this is safe:
- the export map wiring for `@vultisig/sdk/tools/decode` and `@vultisig/sdk/tx` matches the existing narrow subpath pattern already used for parse/defi/bridge
- rollup js + d.ts configs now emit dedicated runtime and declaration bundles for both new subpaths
- the updated smoke script proves `require.resolve(...)`, dynamic import, and basic live exports for both new public surfaces
- the updated unit test pins the package.json export map plus rollup config coverage so the subpaths cannot silently disappear again

gomesalexandre (approved):
Looks good.

Checked the changed surface only:
- `packages/sdk/package.json` adds `./tools/decode` and `./tx` with the same condition matrix as the existing narrow subpaths
- both rollup configs wire those entries through the existing subpath helpers, so the package and type builds stay on the established path
- the new smoke path is the real contract proof here: it packs the SDK, installs it into a scratch consumer, runtime-checks both new subpaths, and type-checks the exported types from a consumer import

I also traced the concrete `normalizeTx(...)` assertion used by the smoke script, so this is not just config-text coverage.

approve.

Issue #1916, opened 2026-08-13
ci: nanoid 3.3.17 HIGH advisory fails Lint and Type Check on every PR rebased onto main, and main never runs the gate


gomesalexandre:
## tl;dr

`yarn npm audit` in the **Lint and Type Check** job fails on a HIGH advisory for `nanoid@3.3.17`. Every PR whose branch is brought up to date with current `main` goes red on it, for a reason that has nothing to do with the PR. `main` itself never runs that gate, so it is invisible there.

## the failure

From #1664's job after updating the branch to base (`Lint and Type Check`, exit 1):

```
├─ Severity: high
├─ Vulnerable Versions: <3.3.18
├─ Tree Versions
│  └─ 3.3.17
└─ Dependents
   └─ postcss@npm:8.5.22
```

Lockfile confirms it:

```
$ grep -n 'nanoid@npm:3' yarn.lock
15276:  resolution: "nanoid@npm:3.3.17"
```

## why it is invisible on main

`gh run list --branch main` shows only `Release` / `Release PR` workflows. No CI/Test/Lint run executes on a push to `main`, so the audit gate only ever fires on a `pull_request`. `main` can therefore carry a HIGH advisory indefinitely while looking clean.

Same shape as agent-backend-ts#2410 ("main's post-merge runs are always cancelled, so a red main is invisible"), different repo and different mechanism.

## the fix looks like a one-liner, with precedent in this repo

`package.json` already carries a resolution from the PREVIOUS round of this same advisory:

```json
"resolutions": {
  "nanoid@npm:^3.3.16": "^3.3.17",
```

The advisory has since moved from `<3.3.17` to `<3.3.18`, and the descriptor that actually resolves in the lockfile today is `nanoid@npm:^3.3.17` — which the existing `^3.3.16` key does not cover. So the likely fix is to add/raise:

```json
"nanoid@npm:^3.3.17": "^3.3.18",
```

then regenerate the lockfile.

## what I deliberately did NOT do

Not patching it. This repo's rule is that a version bump is guilty until proven safe: changelog walk, transitive-dep diff, and a real RUNTIME check, because a bump that builds and type-checks clean can still crash at runtime. `nanoid` is used for id generation across the dependency tree and 3.3.17 -> 3.3.18 is a patch within the same major, so the risk looks low, but "looks low" is exactly the reasoning that rule exists to reject. Filing with the diagnosis so whoever takes it can do the check properly.

## blast radius right now

[...]

PR #1929, opened 2026-08-14
chore: safeguard shared worktree dependencies

rcoderdev:
## Summary by CodeRabbit

* **New Features**
  * Added automated setup and validation for Git worktrees.
  * Added checks to ensure workspace dependencies resolve safely within the active worktree.
  * Added optional shared-package rebuilding during worktree setup.

* **Documentation**
  * Added contributor guidance for configuring worktrees, managing dependencies, and verifying workspace resolution.

* **Tests**
  * Added coverage for worktree setup, dependency links, safety validation, and script integration.


gomesalexandre (approved):
Tested this one properly rather than eyeballing it, since it prefixes `test`, `check`, `check:agent` and `check:ci` with a new gate, and that is a wide blast radius for a chore. It holds up. No blockers.

## the thing this PR does not say it does

The lockfile in here carries **nanoid 3.3.17 -> 3.3.18**, and that clears the advisory currently failing `quality:audit` on main:

```
PR branch     yarn npm audit --recursive --all --severity high   EXIT 0   "No audit suggestions"
origin/main   same command                                       EXIT 1
              └─ nanoid  GHSA-2v37-7h3g-55p8  Vulnerable Versions: <3.3.18
                 └─ Dependents: postcss@npm:8.5.22
```

That is the thing that has been reddening PRs across this repo, and #1916 was filed for it. Whether the bump was deliberate or just fell out of a lockfile refresh, it fixes it. Worth calling out in the description so whoever is chasing the red board knows it lands here, and #1916 can close when this merges.

## what I verified

`worktree:check` in a **normal clone** (not a worktree), which is the false-block risk given it now gates `yarn test` for everyone:

```
node scripts/setup-worktree.mjs --check
[vultisig-worktree] PASS: 19 workspace resolution link(s) stay inside this checkout (local-install).
EXIT 0
```

The gate throws when root `node_modules` is missing (`setup-worktree.mjs:229`), so the question is whether any CI job hits it uninstalled. None do, checked rather than assumed. Every gated invocation is preceded by an install:

```
test.yml:33     yarn install --immutable   ->  :42  yarn test
test.yml:120    yarn install --immutable   ->  :124 yarn check:ci
release.yml:182 yarn install --immutable   ->  :190 yarn test
```

Script tests on the pinned toolchain: **8 pass, 0 fail**.

## nit: two tests fail on node 23, pass on 20/22/24

Not blocking, and not your bug in any practical sense, but worth 30 seconds since it cost me longer than that. On node 23 these two fail:

```
✖ setup safely replaces a broken leaf override without following it        (test.mjs:128)
✖ check accepts a real worktree install whose root workspace links target  (test.mjs:179)
   Error: Path is a directory ... ERR_FS_EISDIR  at rmSync
```

Both are `rmSync(<symlink-to-directory>)` with no options. Isolated it to the primitive rather than guessing:

```
node v20.18.3  rmSync(symlink-to-dir): OK
node v22.22.0  OK
node v23.11.0  THROWS ERR_FS_EISDIR
node v24.14.0  OK
```

So it is a node 23 quirk, fixed again by 24, and `.nvmrc` pins 20, which is why CI is unaffected. `rmSync(p, { force: true, recursive: true })` at those two lines makes it version-proof if you feel like it. Entirely your call, I would not hold the PR for it.

To be clear about my own process: I first read those two failures as a defect in this PR. They were my toolchain being off `.nvmrc`, not your code.

## q: the empty description

677 lines, a new gate on four root scripts, and a silent transitive security bump, with no body. `CONTRIBUTING.md` in the diff carries the actual rationale and it is good (the "a worktree test may stay green while its source is broken" framing is exactly the failure worth preventing). Lifting a few lines of that into the PR body would save the next reader the archaeology.

## risk

Low. Fails closed, which is the correct direction for this class of trap, and the refusal paths (`Refusing to write through symlinked directory`, `Refusing to create overrides outside the worktree`) are both guarded and covered by tests. No new runtime dependency, and no product code touched.

huehuehue nice one, the shared-node_modules trap has bitten me before.

PR #362, opened 2026-04-29
chore: version packages

rcoderdev:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

-   [#360](https://github.com/vultisig/vultisig-sdk/pull/360) [`e52914b`](https://github.com/vultisig/vultisig-sdk/commit/e52914ba87f2d740847fc0de3a49827b0da3e0ba) Thanks [@NeOMakinG](https://github.com/NeOMakinG)! - `@vultisig/core-chain`: lift the qbtc protobuf wire-format helpers to a shared `chains/cosmos/protoEncoding` module (extending it with lower-level `varintBig` / `protoField` primitives that don't apply proto3 default-elision, alongside the existing default-eliding `protoVarint` / `protoBytes` / `protoString`), and add `chains/cosmos/terraClassicTax` with LCD fetchers (`getTerraClassicTaxRate`, `getTerraClassicTaxCap`) plus the pure `applyTerraClassicTax` helper for the cosmos-sdk Dec-fixed-point math. Tax rate is `0` on the live chain today (governance-paused) but the helpers are ready for callers that need to be correct when it reactivates. The previous `qbtc/protoEncoding` package export is replaced by `cosmos/protoEncoding`; the qbtc consumers were updated in lockstep, no behavior change.

### Patch Changes

-   Updated dependencies \[[`e52914b`](https://github.com/vultisig/vultisig-sdk/commit/e52914ba87f2d740847fc0de3a49827b0da3e0ba)]:
    -   @vultisig/core-chain@1.4.3

## Release Notes

* **Chores**
  * Released version 1.4.3 of the core chain package with utility consolidation and new tax features.
  * Released version 1.2.1 of the core mpc package with updated dependencies.

PR #1936, opened 2026-08-14
chore(gitignore): drop trailing slash so dangling node_modules symlinks are ignored

### LINKED ISSUE #1920: chore(gitignore): 'node_modules/' misses dangling symlinks, so git add -A stages them
## tl;dr

`.gitignore:2` is `node_modules/` **with a trailing slash**, which matches directories only. A **dangling** `node_modules` symlink is a file to git, so it is not ignored and `git add -A` stages it. agent-backend-ts hit this exact bug three times and fixed it by dropping the slash; this repo still has it.

I am not reporting this theoretically: **I committed one by accident on #1674 tonight** and had to `git rm --cached` it.

## the probe, with both controls

```
symlink -> /tmp (target EXISTS, resolves as a dir) : IGNORED       <- control, pattern works
symlink -> /Users/gomes/... (DANGLING)             : NOT IGNORED   <- the hole
real node_modules directory                        : IGNORED       <- control, pattern works
```

The distinction is whether the target resolves. `node_modules/` matches a directory, and git treats a symlink-to-existing-directory as one. A symlink pointing at a path that does not exist on this machine is a plain file, so the pattern misses it entirely.

That is exactly the shape these bad symlinks have: they point at **another machine's** clone (`/Users/gomes/...`), so on every machine except the author's they are dangling, and therefore never ignored.

## the same probe against the repo that already fixed it

```
agent-backend-ts (bare `node_modules`)  dangling symlink : IGNORED
vultisig-sdk     (`node_modules/`)      dangling symlink : NOT IGNORED
```

agent-backend-ts's `.gitignore` carries the reasoning in a comment, including the re-tracking history:

```
# NO trailing slash: `node_modules/` only matches a real DIRECTORY, so the accidental
# node_modules SYMLINK (-> a local clone's install) sails straight through `git add .` -
# that is how it got re-tracked THREE times (#124 -> re-added by #333 -> #362 -> re-added
# by #397). The bare pattern matches file, symlink, or directory.
node_modules
```

## why it is worth closing even though nothing is tracked today

`git ls-files | grep -c node_modules` is currently **0** in this repo, so there is no cleanup to do. This is about the trap staying armed. The consequences when it does fire (documented for the sibling repo in agent-backend-ts#2390) are:

1. every PR from the branch fails repo-hygiene on arrival, for a reason unrelated to its change, and
2. anyone who checks that branch out has their real `node_modules` **replaced** by the dangling link, and switching away deletes it entirely — because the same trailing-slash gap makes git treat the real install as expendable.

## fix

One character. `.gitignore:2`:

```diff
-node_modules/
+node_modules
```

Worth carrying the explanatory comment across too, since the slash looks more correct than the bare form and someone will otherwise "tidy" it back.

gomesalexandre:
Closes vultisig/vultisig-sdk#1920

## what

`.gitignore:2` was `node_modules/` **with a trailing slash**, which matches directories only. A **dangling** `node_modules` symlink (pointing at a path that doesn't exist on the current machine — e.g. another machine's clone) is a plain file to git, not a directory, so it slips straight through the pattern and `git add -A` stages it.

agent-backend-ts hit this exact bug three times (re-tracked after #124 → #333 → #362) before dropping the slash fixed it for good. This repo still had the slash.

## verification

Reproduced the issue's own probe:

```
symlink -> /tmp (target EXISTS, resolves as a dir) : IGNORED       <- control, pattern works
symlink -> nonexistent path (DANGLING)             : NOT IGNORED   <- the hole (before this fix)
```

After dropping the slash, `git check-ignore -v node_modules` on a freshly-created dangling symlink now matches `.gitignore:node_modules` — confirmed locally before opening this PR.

`git ls-files | grep -c node_modules` is `0` in this repo, so there's nothing tracked to clean up — this closes the trap before it fires again.

## changeset

None needed — `.gitignore` is a root-level repo-hygiene file, not anything that ships to a published package (per CLAUDE.md's changeset carve-out for root-level docs/CI files).

🤖 Agent-generated


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because The ignore-rule change is sound, but this branch is behind current main and the entire outcome is one small gitignore edit. Reapplying that exact rule from current main is safer and cheaper than carrying this stale head forward.

PR #1938, opened 2026-08-14
fix: Mantle native send gas limit floor, was ~4300x actual usage

### LINKED ISSUE #1847: Mantle native max send strands ~6.75 MNT: the hardcoded 90M gas limit is ~4300x what a transfer uses
## What

`deriveEvmGasLimit` hardcodes a 90,000,000 gas limit for Mantle native transfers:

```ts
// packages/core/chain/tx/fee/evm/evmGasLimit.ts
const mantleTransferGasLimit = 90_000_000n
```

It is used for both the native and the ERC-20 entry, and `capGasLimit` takes `bigIntMax(estimated, thirdParty, minimum)`, so a real `eth_estimateGas` answer of ~21,000 is discarded in favour of the 90M floor.

## Measured on mainnet

Recent Mantle blocks, native transfers with empty calldata:

| | value |
|---|---:|
| gas limit the senders actually set | 400,000 – 900,000 |
| gas actually used | 21,000 |
| `baseFeePerGas` | ~5e10 wei |
| total real cost of a transfer (L2 + L1 data + operator fee) | ~1.4e15 wei ≈ 0.0014 MNT |

The fee we quote is `gasLimit * maxFeePerGas` = `9e7 * (5e10 * 1.5)` ≈ **6.75e18 wei = 6.75 MNT**, against a transfer that costs ~0.0014 MNT. That is roughly 4,800x the real cost.

## Impact

1. **A native max send strands ~6.75 MNT.** The max is `balance - fee`, so that much is left behind on every max send.
2. **A wallet holding less than ~6.75 MNT cannot send natively at all.** This is not just a display artefact: op-geth's pre-execution balance check holds `gasLimit * maxFeePerGas`, so given the limit we set, the node genuinely requires that balance. `refineKeysignAmount` correctly refuses before the ceremony rather than signing something unaffordable — but the reason it refuses is a gas limit we chose, not one the chain needs.
3. The quoted network fee on every Mantle screen is wrong by the same factor.

## Likely origin

The value looks like a leftover from Mantle's pre-Skadi gas accounting, when the chain used very large gas limits with correspondingly small gas prices. Current Mantle prices ordinary EVM gas, so the floor is no longer justified.

## Suggested direction

Bring the floor in line with what the chain does — something in the same range as the other EVM entries for the native case, and the shared ERC-20 default for tokens — and let `capGasLimit` keep using the live `eth_estimateGas` answer when it is higher. Worth confirming against a real Mantle send before changing it, since this value feeds what gets signed.

## Notes

Found while measuring the OP-stack balance-check surcharges in #1846, and deliberately left out of that PR — it is an independent problem and an independent risk. The two interact only in that Mantle's operator fee (1e10 wei per gas of limit) is also priced against this same inflated limit, adding another ~0.9 MNT on top.

gomesalexandre:
closes vultisig/vultisig-sdk#1847

## ⚠️⚠️⚠️ NEEDS A REAL MANTLE BROADCAST BEFORE MERGE ⚠️⚠️⚠️

**This value feeds what gets signed.** I implemented the fix and tests per the dispatch instructions, but I could NOT do a real live signed send on Mantle in this sandboxed session (no vault file / signing capability available). The issue itself explicitly states this is "worth confirming against a real Mantle send before changing it" — that confirmation has **NOT** happened. Unit tests below prove the logic is internally consistent and matches the issue's own on-chain measurements, but they are not a substitute for a real broadcast. **Please get a real Mantle native send confirmed against this branch before merging** — same caution the issue itself calls for.

## what — real fund-safety bug, well-measured in the issue

`packages/core/chain/tx/fee/evm/evmGasLimit.ts`'s `deriveEvmGasLimit` hardcoded `mantleTransferGasLimit = 90_000_000n` for the native-transfer floor. `capGasLimit` (`getEvmFeeQuote.ts`) takes `bigIntMax(estimated, thirdParty, minimum)`, so a real `eth_estimateGas` answer of ~21,000 was always discarded in favor of the 90M floor.

Per the issue's own mainnet measurements: real senders set gas limits of 400,000–900,000 for native Mantle transfers; gas actually USED is ~21,000; the quoted fee at the old 90M floor was ~6.75 MNT vs the real cost of ~0.0014 MNT — a ~4,800x overcharge. Real impact, not cosmetic: (1) a "max send" left ~6.75 MNT stranded on every send, (2) op-geth's pre-execution balance check holds `gasLimit * maxFeePerGas`, so a wallet holding **less** than ~6.75 MNT could not send natively **at all**, (3) every Mantle fee quote was wrong by the same factor.

## fix

`mantleTransferGasLimit` → `400_000n` — the **low end** of the issue's own observed real-sender range (400k–900k), not a value invented from the "other EVM entries" pattern in this file (those run 23k–200k, sized for chains with far less L1-data-posting overhead than Mantle's op-geth fee model). `capGasLimit` still prefers a higher live `eth_estimateGas` answer when one is available — this is only the safety-net minimum, unchanged mechanism, just no longer absurd.

Also fixed `erc20TransferGasLimit[Mantle]` (was the native floor, should size against the shared ERC-20 default like every other chain in that table) per the issue's own suggested direction ("the shared ERC-20 default for tokens"). In practice this table entry is currently unreachable for a real ERC-20 send (calldata always routes through `deriveEvmGasLimit`'s separate `data` branch instead) — fixed for correctness/consistency regardless.

## explicitly NOT touched — flagged instead

`deriveEvmGasLimit`'s `data`-bearing branch hardcodes a **separate** `3_000_000_000n` (3 billion) floor for any calldata-bearing Mantle call (ERC-20 transfers, contract calls, swaps). Same shape of bug, arguably worse in magnitude, but a **wider blast radius** than this issue's scope (every calldata tx, not just native sends), and `getEvmFeeQuote.ts`'s `shouldBufferDataTxGasLimit` already special-cases `chain !== EvmChain.Mantle` specifically because of this value, with an existing test (`"keeps Mantle swap gas limits at the existing special floor"`) pinning it as deliberate. This needs its own issue + explicit topology sign-off — flagged prominently in code comments and here, not silently fixed alongside this one.

## tests

`evmGasLimit.test.ts` (new) pins: the new floor is in the observed 400k–900k range, exactly `400_000n`, >100x smaller than the old floor, still comfortably above real gas used (~21,000), and that the untouched data-bearing branch's `3_000_000_000n` is preserved verbatim (so a future change to it is a conscious decision, not an accidental side effect of touching this file again).

`getEvmFeeQuote.test.ts` (2 new tests): a native Mantle send uses the **live** `eth_estimateGas` answer when it exceeds the new floor; falls back to the **new** (not old) floor when live estimation fails.

**Mutation-tested**: reverted the floor to `90_000_000n`, 3 of 8 tests in `evmGasLimit.test.ts` failed immediately with the exact expected/received mismatch. Restored, reconfirmed all 22 tests across both files green.

## receipts

- `npx tsc --project .config/tsconfig.json --noEmit` clean (the `core-chain` package's own scope — did NOT run the full multi-workspace `yarn typecheck` given disk/load constraints this session; targeted only)
- `eslint` clean on all 3 touched source/test files
- `prettier` clean (this repo's own `.config/.prettierrc`)
- Changeset added (`@vultisig/core-chain` + `@vultisig/sdk`, patch)
- **NOT done**: repo-internal CLI runtime receipt, cross-repo consumer receipt, real-broadcast math/encoding receipt — no vault file available in this sandboxed session. See the warning at the top.

🤖 Agent-generated


neavra (changes requested):
The native-floor half of this is excellent: the 400k value is sourced from the issue's on-chain measurements rather than pattern-matched from the other table entries, the untouched 3B data-branch floor is pinned by a test so a future change is a conscious decision, and the PR body is honest about the missing broadcast receipt. I mutation-checked the floor (reverting to 90M fails the 3 evmGasLimit.test.ts pins), traced `deriveEvmGasLimit` to its single production consumer, and confirmed max-send reads the signed gas limit back out of the payload — the native change is sound.

Two things need to change before this merges:

**1. `erc20TransferGasLimit[Mantle] → 120k` is reachable, not dead code.** The "unreachable" premise is false: `getData()` in `packages/core/mpc/keysign/chainSpecific/resolvers/evm.ts` returns only swap calldata or `keysignPayload.memo`, and a plain token send leaves `memo` undefined (`send/build.ts`), so `deriveEvmGasLimit({ coin: { id }, data: undefined })` hits the table. The ERC-20 transfer calldata is only synthesized later inside `getEvmFeeQuote` for `eth_estimateGas` — it never feeds `deriveEvmGasLimit`. So when estimation fails (RPC hiccup, transient revert), a memo-less Mantle ERC-20 send now signs a **120,000** gas limit instead of the old 90M. Real Mantle token transfers consume gas in the hundreds of millions under Mantle's tokenRatio-inflated accounting (e.g. a mainnet USDT transfer with ~215.5M gas used), so 120k would fail on-chain and burn the fee. It's also internally inconsistent with this PR's own reasoning — if native needs a Mantle-specific 400k because of the chain's overhead, tokens can't be fine on the generic 120k default. Please either give Mantle a separately-justified token fallback or derive the limit from the actual transfer calldata, fix the comment + the `evmGasLimit.test.ts` test that encode the unreachability claim, and add a resolver-level regression test (coin.id set, no memo, estimateGas rejected → assert the signed limit).

**2. Echoing the PR's own merge gate: the real broadcast hasn't happened.** Given the above, it should cover both a native send and a Mantle ERC-20 send with estimation forced to fail, so the fallback that actually gets signed is exercised on-chain.

One clarification, not blocking on its own: the two new `getEvmFeeQuote.test.ts` tests pass `minimumGasLimit: 400_000n` as a literal, so they pin the capGasLimit contract but don't guard the constant (they stay green with the floor reverted to 90M — the `evmGasLimit.test.ts` pins are what guard it). Worth knowing which tests carry which weight.

gomesalexandre:
Fixed #1 — added an `id && chain === EvmChain.Mantle` branch to `deriveEvmGasLimit` so a memo-less Mantle ERC-20 send routes through the same `3_000_000_000n` data-bearing floor every other Mantle contract call gets, instead of falling through to the 120k generic token default. Corrected the now-stale comments/tests that claimed that table entry was unreachable via the old (wrong) reasoning, and added an end-to-end regression test at the `getEvmFeeQuote` boundary that computes `minimumGasLimit` via the real (unmocked) `deriveEvmGasLimit` mirroring the resolver's actual call shape — would have caught this. Updated the changeset. 24/24 relevant tests pass, tsc clean.

On #2 (real broadcast verification): still can't do that from this box — no funded Mantle vault. Both the native send and the now-fixed Mantle ERC-20 send need a real receipt before merge, same as the PR's own existing gate. Flagging for whoever can broadcast.

rcoderdev:
Closing this because The code-level review concern is fixed, but this changes signed Mantle gas limits and still lacks the required real native and ERC-20 broadcast receipts. Re-propose from current main when that fund-safety validation is available.

PR #1939, opened 2026-08-14
fix(across): narrow sourceChain type to Ethereum-only, matching runtime

### LINKED ISSUE #1906: fix(across): public acrossQuote contract advertises multi-origin support but runtime only allows Ethereum
## Summary
The new `acrossQuote()` public surface advertises `AcrossChain` as if any supported chain can be the source, but the implementation hard-rejects every non-Ethereum origin.

## Evidence
- `packages/sdk/src/tools/swap/acrossQuote.ts:23-32` declares `acrossSupportedChains = ['Ethereum', 'Optimism', 'Polygon', 'Arbitrum', 'Base', 'Blast', 'BSC', 'Zksync']`.
- `packages/sdk/src/tools/swap/acrossQuote.ts:63-67` types `sourceChain` as any `AcrossChain`.
- `packages/sdk/src/tools/swap/acrossQuote.ts:213-220` shows an example with `sourceChain: 'Base'`.
- `packages/sdk/src/tools/swap/acrossQuote.ts:230-231` then rejects every non-Ethereum origin: `sourceChain must be Ethereum for this Across factory slice`.
- `packages/sdk/tests/unit/tools/swap/acrossQuote.test.ts:127-137` locks that rejection in as current behavior.

## Impact
This is a public-contract mismatch: callers, docs, and types all imply multi-origin support, but runtime only allows Ethereum origin. Consumers either learn that only by throwing at runtime or reimplement their own guardrails around the SDK surface.

## Why this belongs in the architecture campaign
The campaign target is testable core logic in the SDK with consumers importing it instead of guessing. A public surface that overstates what it supports pushes callers back toward local conditionals and “special-case Around/Across support” wrappers.

## Fix sketch
- Either narrow the public contract so `sourceChain` truthfully communicates the Ethereum-only slice today, or split the helper name/docs to make the limitation explicit.
- Keep the current fail-closed runtime behavior unless and until genuine multi-origin Across support is implemented and receipt-backed.

## Effort
Small / medium. Mostly contract tightening, not fund-adjacent execution logic.

gomesalexandre:
Closes vultisig/vultisig-sdk#1906

## what

`acrossQuote()`'s public contract advertised multi-origin support it doesn't have: `AcrossQuoteParams.sourceChain` was typed as the full `AcrossChain` union (all 8 supported chains), the JSDoc `@example` showed `sourceChain: 'Base'`, but the runtime hard-rejects every non-Ethereum origin (`'sourceChain must be Ethereum for this Across factory slice'`). Callers either discovered the real contract by hitting that throw, or wrote their own guardrail wrappers around the SDK to avoid it.

## how

Per the issue: kept the fail-closed runtime behavior **exactly as-is** (no expansion to genuine multi-origin support — that needs real Across API integration + live per-chain verification, out of scope here). Narrowed the **type + docs** to match reality instead.

- `AcrossQuoteParams.sourceChain: AcrossChain` → `typeof ETHEREUM_ORIGIN_CHAIN` (the literal `'Ethereum'` type, tied to the single source-of-truth constant rather than a hardcoded literal string). `destinationChain` is untouched — it correctly stays the full `AcrossChain` union, since any of the 8 supported chains except the origin **is** a valid destination today.
- Fixed the misleading `@example` (`sourceChain: 'Base'` → `'Ethereum'`).
- Internal `assertPinnedSpokePools` stays typed with the general `AcrossChain` union for both its source/destination params — it's a private helper indexing into `acrossSpokePools` (keyed by all 8 chains) for **both** directions, not part of the public input contract this issue is about.

## a small tangent worth flagging

While fixing the `@example`, I verified it against the **live** Across API rather than guessing at a plausible-looking route. My first attempt (WETH on Ethereum → USDC on Arbitrum) actually 400'd with `ROUTE_NOT_ENABLED` — which would have been exactly the kind of silent doc drift this issue is about, just relocated rather than fixed. Swapped to a route I confirmed genuinely quotes live: USDC (Ethereum) → USDC (Arbitrum), 1 USDC in → 0.994465 USDC out, relay fee 5535 (micro-USDC).

## tests

Updated the existing "rejects a non-Ethereum origin" test to pass its now-type-invalid `sourceChain` via an explicit `as unknown as 'Ethereum'` cast, with a comment explaining why: the test's whole point is proving the **runtime** guard still holds for a caller that bypasses the type system (plain JS, or a TS caller reaching for `as any`), so the narrower type is exactly what makes that cast necessary — not a workaround for a broken test. The runtime assertion (`/must be Ethereum/`) is byte-for-byte unchanged.

- All 14 `acrossQuote.test.ts` tests pass.
- Full SDK unit suite: 220 files / 3375 tests pass.
- `yarn typecheck` clean across all workspaces (this DID catch the type-narrowing collision in the existing test before I fixed it — exactly the kind of thing a type change like this should surface).
- `yarn lint` clean.
- Live verification: ran the fixed `@example` directly against `app.across.to` via `tsx` (see above) — genuinely live quote, not mocked.

## changeset

`.changeset/across-quote-sourcechain-ethereum-only.md` — `@vultisig/sdk` patch.

🤖 Agent-generated


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because PR #2034 already merged this contract fix, including public AcrossOriginChain and ACROSS_ORIGIN_CHAIN exports and the corrected example. Closing this duplicate because the outcome is already on main.

PR #1941, opened 2026-08-14
fix(skip): add missing Akash bech32 HRP to the Skip address allowlist

### LINKED ISSUE #1785: fix(skip): Akash routes accept wrong-chain bech32 recipients because akashnet-2 is missing from the HRP allowlist
## Summary
Akash Skip routes accept wrong-chain bech32 recipients because  is missing from the local  allowlist.

## Evidence
-  validates Cosmos addresses with a local  table.
- That table omits , so validation falls back to generic bech32-shape acceptance for Akash routes.
- The repo already treats Akash as a supported Cosmos chain elsewhere (, ).

## Impact
A wrong-chain bech32 destination like  can pass local validation for an Akash route. This is a real correctness/fund-safety bug and another example of local chain-metadata copies drifting.

## Fix sketch
- add  to the local Skip HRP allowlist
- add unit regressions proving  is rejected and  is accepted for Akash routes
- longer-term: dedupe the local HRP map against shared chain metadata

## Verification attempted
I prepared the patch in , but  currently fails the shared build prerequisite before the change can be fully validated:
- Usage Error: The nearest package directory (/Users/gomes/.hermes/hermes-agent) doesn't seem to be part of the project declared in /Users/gomes.

- If /Users/gomes isn't intended to be a project, remove any yarn.lock and/or package.json file there.
- If /Users/gomes is intended to be a project, it might be that you forgot to list .hermes/hermes-agent in its workspace configuration.
- Finally, if /Users/gomes is fine and you intend .hermes/hermes-agent to be treated as a completely separate project (not even a workspace), create an empty yarn.lock file in it.

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] [--require #0] <scriptName> ...
- failure:  ( property type error)

Per campaign instruction, I am filing the issue and stopping short of a speculative PR.

## Report
https://gist.github.com/gomesalexandre/4175d4f2cd9e22e4c000e664cb45ea48

gomesalexandre:
Closes #1785

## what

The Skip route builder validates cosmos addresses against a local `COSMOS_CHAIN_HRPS` table (bech32 HRP per `chain_id`) specifically to catch a wrong-chain-but-valid-shape recipient (e.g. an `osmo1...` address on a route meant for `cosmoshub-4`). `akashnet-2` was missing from that table entirely, so an Akash-destined route fell back to the generic `^[a-z]+1...` bech32-shape regex instead of checking the actual HRP - a wrong-chain bech32 destination could pass local validation for what was supposed to be an Akash-destined transfer.

## why / confirmed this is just a missing table entry

Akash is already a registered cosmos chain elsewhere in the repo - `cosmosChainId` in `packages/core/chain/chains/cosmos/chainInfo.ts` maps `CosmosChain.Akash` → `'akashnet-2'` with the standard `akash1` bech32 prefix. This was purely a missing entry in the Skip-local `COSMOS_CHAIN_HRPS` map, not a deeper gap in how Akash is supported.

## how

Added `'akashnet-2': 'akash'` to `COSMOS_CHAIN_HRPS` in `packages/sdk/src/tools/swap/skip/skipSwap.ts`.

## tests

`packages/sdk/tests/unit/tools/swap/skipSwap.test.ts`:
- rejects a wrong-chain `osmo1...` recipient on an `akashnet-2`-destined route (`invalid_input` + HRP-mismatch message).
- accepts a correct `akash1...` recipient on the same route shape.

Both share a route/msgs fixture consistent with `destChainId: 'akashnet-2'` (`chain_ids` / `required_chain_addresses`), so the only variable between the two tests is the `toAddress` prefix - isolates the HRP check itself from unrelated custody-chain / intermediate-address validation.

19/19 green in `skipSwap.test.ts`. `tsc --project .config/tsconfig.json --noEmit` and the repo's eslint runner clean.

Mutation-tested: removed the `akashnet-2` table entry, reran the two new tests - the "accept" test still passed (irrelevant to whether the entry is present) but the "reject" test failed with `out.ok === true` instead of `false` - the wrong-chain address was silently accepted, exactly the bug this fixes. Restored, reran - back to 19/19 green.

## out of scope

Per the issue's own note: deduping this local HRP map against shared chain metadata is a separate architecture-class follow-up, not attempted here.

## changeset

Added `.changeset/akash-skip-hrp-allowlist.md` (`@vultisig/sdk` patch) per this repo's mandatory changeset rule.

🤖 Agent-generated


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because The Akash HRP mapping is correct and well tested, but this branch is behind current main and overlaps the broader chain-registry work in #2092. Reapply the focused mapping from current main so it lands without carrying this stale branch.

PR #1947, opened 2026-08-14
fix(balance): decode bytes32-returning erc-20 symbols in getevmbalances (sdk#1946)

### LINKED ISSUE #1946: fix(balance.evm): getEvmBalances fails on bytes32-returning ERC-20 symbols like MKR
## Summary
`getEvmBalances()` used strict ABI-string `symbol()` decoding, so legacy bytes32-returning ERC-20s like MKR failed the whole balance read.

## Evidence
- pre-fix `packages/sdk/src/tools/evm/balanceEvm.ts:65-95`
- canonical tolerant readers already existed in `packages/sdk/src/tools/token/resolveContract.ts:128-154` and `packages/sdk/src/tools/dex/uniswap/erc20.ts:15-37`

## Impact
A real MKR-style token in the requested list could reject the whole EVM balance read instead of returning balances.

## Fix sketch
Reuse bytes32-tolerant metadata decoding in `getEvmBalances()` and pin the regression with a unit test.

## Effort
Small.

## Artifact / report
Round 143 report: https://gist.github.com/gomesalexandre/406d9e112b76781793cfcfca6af9f113

gomesalexandre:
Closes vultisig/vultisig-sdk#1946

## what

`getEvmBalances()` used viem's strict typed `readContract()` for a token's `symbol()`. Legacy ERC-20s that predate the dynamic-string return convention (MKR is the famous one, also SAI) declare `symbol()` as `bytes32`, right-padded with null bytes. Decoding that shape as a canonical ABI dynamic string throws, which rejected the per-token promise and aborted the ENTIRE `Promise.all` batch read, not just that one token's entry - so a single MKR (or similar) balance request in a batch broke every other token in the same call too.

## how

Swapped the strict `readContract({ functionName: 'symbol' })` call for `readSymbol()`, an existing tolerant string/bytes32 decoder that already lives at `packages/sdk/src/tools/dex/_erc20.ts` and is reused by the Uniswap V2/V3 quote readers. It does a raw `eth_call`, tries the canonical ABI string decode first, and falls back to bytes32 decoding (stripping trailing null bytes) on failure. Reused as-is rather than inventing a third tolerant decoder (there's already a near-duplicate one in `resolveContract.ts` for a separate flow, out of scope here).

`decimals()`/`balanceOf()` reads are untouched (both are unambiguous uint types, no ABI ambiguity there).

## testing

- New regression test using MKR's real on-chain `symbol()` shape (`bytes32 public symbol = "MKR"`, right-padded) - proves the balance resolves and decodes to `'MKR'` correctly, and that an unrelated token in the same batch is unaffected.
- Mutation-tested: reverted the fix locally, confirmed the new test (and two others) fail with the exact symptom (`Cannot convert undefined to a BigInt` from the corrupted symbol propagating through decimals formatting / `symbol: undefined`), restored, reconfirmed green.
- Updated existing tests' mocks (`client.call` alongside `client.readContract`) since symbol reads now go through a raw `eth_call` instead of viem's typed contract read.
- `yarn vitest run --config tests/unit/vitest.config.ts tests/unit/tools/evm tests/unit/tools/dex` - 55/55 green.
- Lint clean on both touched files. `yarn workspace @vultisig/sdk typecheck` has pre-existing unrelated failures (`@vultisig/mpc-types`/`@vultisig/mpc-wasm` not built in this worktree) - confirmed identical on a clean `main` checkout, not introduced by this change.
- Changeset added (`@vultisig/sdk` patch).

Note: couldn't run the usual `/codex-review` auto-check before opening this - Codex CLI auth is expired in this environment (401 token_expired on every request). Did a careful manual self-review instead, including tracing the import graph for a possible circular-dependency concern (`balanceEvm.ts` -> `dex/_erc20.ts` -> `evm/index.ts` barrel -> back to `balanceEvm.ts`), which resolves safely under ESM live-bindings since the barrel is pure re-exports with no top-level side effects, confirmed empirically by the passing test suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed EVM balance retrieval for legacy ERC-20 tokens whose symbols use `bytes32` formatting.
  * Balance batches now continue processing other tokens when one token returns a legacy symbol format.
  * Improved compatibility with tokens such as MKR while preserving support for standard symbol responses.

* **Tests**
  * Added regression coverage for mixed token balances and legacy symbol decoding.
  * Updated balance and DEX quote tests to validate reliable contract-call handling.


neavra (approved):
Nice fix — reusing the existing tolerant `readSymbol()` instead of minting a third decoder is the right call, the circular-import workaround is explained precisely where the next reader needs it, and the regression test uses MKR's real bytes32 return shape with an unrelated token in the same batch to prove batch survival. I mutation-checked it (reverting the `balanceEvm.ts` hunk fails 3 tests including the MKR batch regression) and reproduced the import cycle both ways with depcruise (barrel import → no-circular error; concrete import → clean), and confirmed `evmCall` resolves the identical memoized client the removed `readContract` path used, so no RPC drift on other EVM chains.

A few non-blocking notes for a possible follow-up:
- A token whose `symbol()` actually *reverts* (or whose eth_call RPC errors) still aborts the whole batch — `evmCall` throws before `readSymbol`'s `'UNKNOWN'` branch can fire. That's unchanged from before (strict readContract threw too), so it's out of scope for #1946, but the new comment at `balanceEvm.ts:82-85` ("so one legacy token doesn't fail the whole batch") slightly overstates the coverage — maybe narrow it to "bytes32-returning". If symbol-as-optional-metadata is the contract you want, catching readSymbol failures as `'UNKNOWN'` (while keeping balanceOf/decimals fail-closed) would be a natural small follow-up.
- `decodeBytes32String` accepts any non-decodable returndata without checking it's one right-null-padded word, so a malformed response can become a mojibake symbol rather than `'UNKNOWN'` — pre-existing decoder, strictly better than the old thrown batch, just worth a shape check someday.
- Tiny one: the `MKR` test constant is 39 hex chars (real MKR ends `...A579A1`) — harmless in a mock, but the comment says it's the real address, so worth fixing before someone copies it.

PR #1952, opened 2026-08-15
Prevent quality contract Yarn cache leaks

rcoderdev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency installation isolation during SDK, CLI, and MCP validation.
  * Prevented test runs from modifying or relying on shared Yarn caches.

* **Tests**
  * Added coverage confirming temporary caches are created, populated, and cleaned up correctly.
  * Verified existing environment settings remain preserved while disposable cache settings are applied.

### DIFF
--- a/scripts/quality-contracts-cache.mjs
+++ b/scripts/quality-contracts-cache.mjs
@@ -1,0 +1,16 @@
+import { mkdirSync } from 'node:fs'
+import path from 'node:path'
+
+export function createDisposableYarnEnv(workRoot, baseEnv = process.env) {
+  const cacheFolder = path.join(workRoot, 'yarn-cache')
+  const globalFolder = path.join(workRoot, 'yarn-global')
+  mkdirSync(cacheFolder, { recursive: true })
+
+  return {
+    ...baseEnv,
+    YARN_CACHE_FOLDER: cacheFolder,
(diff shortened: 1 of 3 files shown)


gomesalexandre (approved):
lgtm conceptually. this is a straightforward CI/test-harness isolation fix and the new coverage is on the right seam.

what i verified from the current diff:
- `createDisposableYarnEnv(workRoot)` now scopes cache + global folders under the per-run temp root, so the packed-consumer smoke installs stop writing `file:` tarball artifacts into a shared yarn cache path
- the three call sites in `quality-contracts.mjs` all moved to that helper, so the isolation behavior is consistent across sdk, cli, and mcp packed-install smoke paths
- `quality-contracts-cache.test.mjs` does the useful end-to-end thing instead of only unit-testing env shape: it performs a real packed `yarn add`, proves the inherited user-global cache folders stay untouched, proves the disposable cache gets populated, and proves cleanup removes it afterward
- current head CI is green

non-blocking note only: this also drops the old cross-run cache reuse under `repoRoot/.yarn/cache`, so `quality:contracts` may now trade some CI speed/stability for cleaner isolation. not a correctness blocker, just something worth keeping an eye on if `check:ci` gets meaningfully slower or flakier.

PR #1960, opened 2026-08-15
fix: sync sdk.validate chain support with decimals registry (OP/ZK)

### LINKED ISSUE #1945: fix(validate): sdk.validate rejects OP/ZK even though its own decimals registry includes them
## Summary
`packages/sdk/src/utils/validateNormalizers.ts` is internally self-contradictory: its decimals registry includes canonical symbols like `OP` and `ZK`, but its validator and normalizer reject every 2-character ticker.

## Evidence
- `packages/sdk/src/utils/validateNormalizers.ts:53,91,114`
- `packages/sdk/src/utils/validateNormalizers.ts:355-385`
- `packages/sdk/tests/unit/utils/validateNormalizers.test.ts:151-179`

## Impact
Downstream consumers cannot safely delete local symbol exceptions and rely on `sdk.validate`, because the same module both blesses and rejects real production symbols.

## Fix sketch
Make the registry, validator, and tests agree on the intended contract. If 2-character canonical tickers are valid, widen the validator and align downstream consumers. If they are intentionally invalid, remove/redesign the contradictory registry entries.

## Effort
Small/medium, but needs cross-repo contract clarity.

## Artifact / report
Round 143 report: https://gist.github.com/gomesalexandre/406d9e112b76781793cfcfca6af9f113

gomesalexandre:
Closes #1945

## what

`isValidTokenSymbolFormat` / `normalizeTokenSymbol` in `validateNormalizers.ts` rejected every 2-char ticker, but the decimals registry two lines above already carries `OP` (Optimism) and `ZK` (zkSync) as canonical, real production entries. Same file, same PR that introduced it, blessing and rejecting the same symbols.

## why / root cause

The comment already at the top of `SYMBOL_SHAPE` claimed to mirror the Go agent-backend's `symbolCandidateRe` "EXACTLY" — and it did, but only ported half the Go contract. Went and read the actual Go source (`vultisig-agent-backend/internal/service/agent/validator/token_symbol.go`):

- `symbolCandidateRe` (`[A-Z][A-Z0-9]{2,9}...`) is a freeform-text CANDIDATE-extraction heuristic — it floors at 3 chars specifically to avoid matching 2-letter noise words ("ID", "US", "OK") when scraping tickers out of an LLM response.
- Canonical, already-known tickers bypass that floor entirely via a SEPARATE `knownTokenSymbols` allowlist (direct map lookup, no regex), which includes `"OP": {}` and `"ZK": {}` verbatim.

The SDK only ported the extraction regex, not the allowlist-bypass tier, so any ticker in the SDK's own `tokenDecimals` registry (the allowlist's equivalent here) that happened to be 2 chars got rejected by the format validator even though `decimalsFor()` two lines up resolved it fine.

## fix

Widened `isValidTokenSymbolFormat` to validate per-leg (so slash-pairs like `OP/RUNE` work too), where a leg is valid if it either matches the general 3-10 char shape OR is a 2-char, letter-led ticker already present in `tokenDecimals` (mirrors Go's allowlist bypass exactly). Digit-led registry entries like `1INCH` intentionally stay excluded from the format check — Go's extractor is letter-led only, so that's a separate, deliberate contract gap, not the drift this issue is about.

This is the architectural fix rather than a two-entry patch: any *future* 2-char ticker added to `tokenDecimals` now automatically validates too, instead of needing a second manual update to the shape regex.

## tests

- `accepts known 2-char registry tickers (OP, ZK) via the allowlist bypass`
- `still rejects 2-char tickers NOT in the registry (min length 3, matching Go)` — `A1`/`ZZ`/`xy` stay rejected
- per-leg pair validation (`RUNE/OP` valid, `RUNE/A1` still invalid)
- `normalizeTokenSymbol('op')` now resolves instead of throwing
- new cross-consistency regression test: every letter-led `tokenDecimals` entry must pass `isValidTokenSymbolFormat`, so this specific drift (registry says yes, validator says no) can't silently reopen on a future chain addition

## verification

- `yarn workspace @vultisig/sdk vitest run tests/unit/utils/validateNormalizers.test.ts` — 32/32 passed
- `yarn workspace @vultisig/sdk vitest run tests/unit/utils/` — 598/598 passed
- `yarn workspace @vultisig/sdk typecheck` — clean
- `eslint` + `prettier --check` on both touched files — clean
- `yarn npm audit --recursive --all --severity high` — clean, no lockfile changes
- Codex CLI (`/codex-review`) was unavailable this session (expired auth token, 401 on refresh) — skipped per its own fallback rule, self-reviewed against the actual Go source instead

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Token validation now accepts registered two-character symbols, such as OP and ZK.
  * Slash-separated token pairs are validated correctly, including pairs containing short symbols.
  * Unregistered or invalid short symbols continue to be rejected.


gomesalexandre (commented):
current-head remote conceptual review on `46f0d054cd8a30cdab9e96984814433f977b4e28` only.

what i re-checked
- the validator keeps the 3-char floor for unknown tickers but now lets already-registered 2-char symbols like `OP` and `ZK` bypass that floor, matching the Go-side `knownTokenSymbols` allowlist behavior instead of only the extraction regex
- slash-pairs are validated per leg, so known short tickers work in pair form while unknown short legs still fail closed
- exact-head tests now cover standalone short-ticker acceptance, pair-leg acceptance/rejection, normalization behavior, and a cross-registry consistency guard against future `tokenDecimals` drift
- current-head CI is green

I do not see a fresh code-level blocker on this diff from the remote pass.

four-eyes note: this branch contains substantive `gomesalexandre` commits, so i cannot approve it. it still needs a non-gomes current-head approval before merge.

NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because The OP/ZK validator correction is sound, but the branch is behind current main and the focused registry-based change is inexpensive to recreate. Re-propose it from current main with the same fail-closed consistency tests.

PR #1962, opened 2026-08-15
fix: preserve lamports precision in getSolBalance

### LINKED ISSUE #1923: solana: getSolBalance reconstructs exact lamports, then re-exposes a lossy numeric lamports field
## Summary
`getSolBalance()` carefully reconstructs exact lamports from the raw RPC response, then immediately converts them back into a potentially unsafe JavaScript number before returning the public `lamports` field.

## Evidence
- `packages/sdk/src/tools/balance/solana.ts:10-16` documents `lamports` as a number and explicitly notes it becomes lossy above `Number.MAX_SAFE_INTEGER`.
- `packages/sdk/src/tools/balance/solana.ts:105-124` parses the raw RPC decimal text into `lamportsRaw` and `lamportsBig`, then returns `lamports: Number(lamportsBig)`.

## Impact
- Any consumer that reads the convenient `lamports` field can silently round the raw balance once the account exceeds the safe integer boundary.
- The SDK already added exact raw-string fields to avoid this class elsewhere; Solana still leaks the lossy field as if it were authoritative.

## Fix sketch
- Keep compatibility, but add an explicit exact field (`lamportsBigInt`/`lamportsExact`) and update examples/tests to prefer it.
- Longer-term, consider nulling or fail-closing the numeric field when the value is unsafe.

## Verification target
- Add a focused unit test with a mocked `getBalance` value above `2^53` proving the exact field stays correct and the SOL decimal formatting stays stable.

## Report
Artifact/report link: pending round-140 report gist (will update once published).

gomesalexandre:
Closes #1923

## what

`getSolBalance()` did the right thing internally - it recovers exact lamports from the raw RPC response text (bypassing `JSON.parse`'s float rounding above 2^53) - then immediately re-lossified that value by returning `Number(lamportsBig)` as the public `lamports` field. Any caller reading `lamports` for a balance past `Number.MAX_SAFE_INTEGER` (~9.007M SOL) silently got a rounded number back, as if it were authoritative.

## how

`lamports` is now `number | null`. It stays a plain number for the overwhelming majority of real balances (anything under ~9.007M SOL), and fails closed to `null` once the exact u64 value can't be represented safely - instead of quietly lying about it. `lamportsRaw` (base-10 string) and `sol` (exact decimal string, already computed via BigInt-only math) were already lossless and are untouched.

This matches the existing convention elsewhere in `packages/sdk/src/tools/balance/` - `getCosmosBalance` (`nativeRaw`) and `getPolkadotBalance` (`balanceRaw`) both represent raw on-chain amounts as strings and never expose a lossy plain-number field for them. Solana already had the string field (`lamportsRaw`); this just stops the `number` field from misrepresenting itself as exact when it isn't.

## tests

Added regression coverage in `packages/sdk/tests/unit/balance-solana.test.ts`:
- `lamports` is `null` (not a rounded number) for a value 12345678901234567 > 2^53, while `lamportsRaw`/`sol` stay exact.
- Boundary test: `lamports` is a real number exactly at `Number.MAX_SAFE_INTEGER`, `null` one lamport above it.

Full `getSolBalance`/`getSplTokenBalance` suite: 12/12 passing.

## verification

- `yarn workspace @vultisig/sdk vitest run tests/unit/balance-solana.test.ts` - 12/12 passing
- `yarn typecheck` - clean for this diff (repo currently has 2 pre-existing, unrelated `auxiliaryData`/`ISigningInput` cardano errors present on `main` before this change too - confirmed via `git stash`)
- `eslint` + `prettier --check` on both touched files - clean
- `yarn npm audit --recursive --all --severity high` - clean, no lockfile changes
- Searched open PRs for `getSolBalance`/`1923` - no existing PR already covers this

## changeset

`.changeset/solana-lamports-precision.md` - `@vultisig/sdk` patch (public field type narrowing, same severity class as #1939 for #1906).

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana balance precision for very large lamport values.
  * Returns `null` when a numeric lamport balance exceeds JavaScript’s safe integer range.
  * Preserves exact balances through the raw lamport and SOL string values.

* **Tests**
  * Added coverage for safe-integer boundaries and large balance values.


neavra (approved):
Solid fail-closed fix — the gate lands at the single construction site of `lamports`, and the boundary tests (exact number at `Number.MAX_SAFE_INTEGER`, `null` one lamport above) drive the real raw-text parse path via a verbatim-body mock, so they can't be fooled by a pre-parsed fixture. Verified the new tests fail with the fix reverted, that `getSplTokenBalance` has no equivalent lossy field (jsonParsed amounts are strings), and that nothing in-repo reads `.lamports`, so the `null` only reaches external consumers.

One non-blocking thought: `number → number | null` is technically a compile break for external consumers under `strictNullChecks`, so `minor` might signal it better than `patch` — though patch does match the precedent set by #1939 and the other fail-closed changesets, so fine either way.

PR #1964, opened 2026-08-15
feat: expose sdk.amount namespace

gomesalexandre:
Closes #1933

## what

`sdk.amount.convert` is documented (the `convertAmount.ts` module docstring and CHANGELOG #810) as the canonical amount-conversion surface, but neither the root SDK export nor the `Vultisig` instance actually exposed an `amount` namespace - only flat `convertAmount` / `toBaseUnits` / `toHumanUnits` / `fiatToCrypto` / `cryptoToFiat` functions. Anyone following the documented `sdk.amount.convert(...)` usage got a runtime `undefined is not a function`.

Adds a grouped `amount` namespace object in `packages/sdk/src/utils/convertAmount.ts`:

```ts
export const amount = {
  convert: convertAmount,
  toBaseUnits,
  toHumanUnits,
  fiatToCrypto,
  cryptoToFiat,
} as const
```

`convert` aliases the existing `convertAmount` implementation (no logic reimplemented, pure delegation); the sibling helpers keep their full names. Wired up following the exact patterns already established in this repo:

- Root export + `Vultisig` instance getter, mirroring `sdk.defi` (`packages/sdk/src/Vultisig.ts`).
- `import * as x from ...; export { x }` namespace-handle pattern, mirroring `sdk.gas` and the `sdk.evm` / `sdk.token` / `sdk.cosmos.gov` fix in #1890.
- React Native platform entry updated separately since it maintains its own explicit re-export allow-list (`packages/sdk/src/platforms/react-native/index.ts`) rather than re-exporting the whole root; node/browser/electron-main/chrome-extension entries pick it up automatically via `export * from '../../index'`.
- `scripts/quality-contracts.mjs` published-package smoke assertions extended for `sdk.amount.convert` across root/node/browser, matching the existing `sdk.evm`/`sdk.token`/`sdk.cosmos.gov` assertions.

## re #1912

#1912 is the broader "`Vultisig` instance missing documented namespaces" issue (`balance`/`bridge`/`cosmos`/`decode`/`gas`/`prep`/`price`/`swap`). `amount` is not in that list - #1912's own evidence and verification target never mention it - so this PR does not overlap with or close #1912. Leaving that one open/untouched.

## verification

- `yarn workspace @vultisig/sdk vitest run tests/unit/public/amountNamespaceRootExports.test.ts tests/unit/Vultisig.test.ts tests/unit/public/toolNamespaceRootExports.test.ts tests/unit/utils/publicExports.test.ts` - 102 passed
[...]

PR #1965, opened 2026-08-15
feat: expose sdk.balance/bridge/cosmos/decode/gas/prep/price/swap namespaces

### LINKED ISSUE #1912: public API: Vultisig instance is missing documented sdk.balance/sdk.swap/sdk.prep/sdk.price namespaces
## Summary
`new Vultisig()` only exposes `sdk.defi.*` as an instance namespace even though the package changelog repeatedly advertises newer helper families as `sdk.balance.*`, `sdk.swap.*`, `sdk.prep.*`, `sdk.price.*`, `sdk.decode.*`, `sdk.cosmos.*`, and `sdk.gas.*`.

That leaves the public package in an awkward split state:
- the root/module exports exist for many helpers
- but the main `Vultisig` class instance does not expose matching namespace handles
- existing export coverage only checks root exports, so this instance-surface drift can ship silently

## Evidence
- `packages/sdk/src/Vultisig.ts:161-171` exposes `defi` and nothing analogous for `balance`, `bridge`, `cosmos`, `decode`, `gas`, `prep`, `price`, or `swap`
- `packages/sdk/src/tools/index.ts:47-53` already adds a namespace handle for `gas`, proving the package wants this ergonomic shape in at least one family
- `packages/sdk/tests/unit/utils/publicExports.test.ts:3-102` checks root exports only; it never instantiates `new Vultisig()` and asserts instance namespace parity
- `packages/sdk/CHANGELOG.md` repeatedly documents new helpers as instance-style namespaces, e.g.
  - line 7: `sdk.balance.evm`
  - line 15: `sdk.swap.skip`
  - line 31: `sdk.prep.cosmosStaking`
  - line 60: `sdk.gas.cosmos`
  - line 125: `sdk.balance.cosmos`
  - line 135: `sdk.cosmos.gov`
  - line 167: `sdk.price.get`

## Impact
Consumers can reasonably write code like:
```ts
const sdk = new Vultisig()
await sdk.initialize()
await sdk.price.getPrice(...)
```
…and discover at runtime that `price`/`swap`/`prep`/etc are undefined on the instance even though the package changelog and existing `sdk.defi.*` precedent imply otherwise.

This is especially easy to hit because most examples name the instance variable `sdk` too.

## Suggested fix
Add readonly namespace getters on `Vultisig` for the documented helper families:
- `balance`
- `bridge`
- `cosmos`
- `decode`
- `gas`
- `prep`
- `price`
- `swap`

Also add a focused instance-surface regression test in `tests/unit/Vultisig.test.ts` so future helper families cannot be documented as `sdk.<namespace>` without the class actually exposing them.

## Verification target
A minimal regression should pass:
```ts
const sdk = new Vultisig({ autoInit: false, storage: new MemoryStorage() })
expect(typeof sdk.balance.getEvmBalances).toBe('function')
expect(typeof sdk.swap.findSwapQuote).toBe('function')
expect(typeof sdk.prep.prepareSendTxFromKeys).toBe('function')
expect(typeof sdk.price.getPrice).toBe('function')
```

## Campaign context
Round 137 `/improve` architecture pass. This is the same broader duplicated-not-imported/public-surface drift thesis as the prior export-parity rounds, but specifically on the `Vultisig` instance contract rather than root/RN re-exports.

gomesalexandre:
Closes #1912

## what

`new Vultisig()` only exposed `sdk.defi.*` as an instance namespace even though the package changelog repeatedly documents newer helper families as `sdk.balance.evm`, `sdk.swap.skip`, `sdk.prep.cosmosStaking`, `sdk.gas.cosmos`, `sdk.cosmos.gov`, `sdk.price.get`, etc. The root/module exports for these families already existed, but the `Vultisig` class had no matching instance getter, so code written against the documented `sdk.balance.getEvmBalances(...)` shape failed at runtime with `undefined is not a function`.

Adds readonly `get balance() / bridge() / cosmos() / decode() / gas() / prep() / price() / swap()` getters on `Vultisig` (`packages/sdk/src/Vultisig.ts`), mirroring the existing `sdk.defi` / `sdk.amount` (#1964) pattern:

- A grouped namespace object per family. `cosmos` and `gas` already had a grouped `tools/*` barrel (just needed the instance getter). `bridge`/`decode`/`price` got a new `import * as X from './tools/X'` namespace re-export in `tools/index.ts`, matching the existing `sdk.gas`/`sdk.cosmos` handles. `balance`/`prep`/`swap` are new dedicated files (`tools/balance/namespace.ts`, `tools/prep/namespace.ts`, `tools/swap/namespace.ts`) since their underlying exports were scattered across per-chain files with no existing grouping — see below for why they're hand-curated instead of barrel imports.
- Root export in `packages/sdk/src/index.ts`.
- React Native entry wiring in `packages/sdk/src/platforms/react-native/index.ts`.
- `scripts/quality-contracts.mjs` published-package smoke assertions for root/node/browser, extending the existing `sdk.evm`/`sdk.cosmos.gov` pattern, plus type-smoke coverage for all 7.

## scope: sdk.balance / sdk.prep / sdk.swap exclude a few helpers

`Vultisig.ts` is a shared class that also bundles for React Native (`platforms/react-native/index.ts` re-exports the same class). RN's Hermes engine crashes at module init on a couple of specific dependency graphs — `@polkadot/api` (BN.js double-bundle) and `@solana/web3.js` (reads `globalThis.Buffer` at top level) — which is why this codebase already lazy-imports those specific helpers at the RN entry point instead of statically re-exporting them.

Building `sdk.balance` / `sdk.swap` / `sdk.prep` from the full `tools/balance` / `tools/swap` / `tools/prep` barrels would reintroduce those same static imports into the shared `Vultisig` class (regardless of what the RN entry file itself re-exports, since the class is what gets bundled). So each of the three new namespace files hand-picks only the sub-functions already proven RN-bundle-safe (either because they're already statically re-exported at the RN entry, or because `vault/services/TransactionBuilder.ts` already imports them statically and is unconditionally reachable from the shared class via `VaultBase`):

- **`sdk.balance`** excludes the Polkadot reads (`balancePolkadot`, `getPolkadotNativeBalance`, `getPolkadotAssetBalance`) — still reachable as flat exports.
- **`sdk.swap`** excludes Jupiter's `buildJupiterSwapTx` (same-chain Solana, pulls `@solana/web3.js`) and the Skip Go builder family (not currently RN-exported at all, unaudited). Includes `findSwapQuote`, `acrossQuote`, and the Astroport family.
- **`sdk.prep`** excludes `buildSplTransfer` (proven Solana hazard) plus `getMaxSendAmountFromKeys`, `prepareJettonTransferTxFromKeys`, `prepareSwapTxFromKeys`, `prepareTrc20TransferFromKeys`, `prepareUtxoConsolidateTxFromKeys`, `prepareSuiTokenTransferFromKeys`, `prepareIbcTransfer` — the RN entry currently lazy-imports or omits these too and nothing in the codebase yet proves them RN-static-safe. Includes `prepareSendTxFromKeys` (the issue's own verification target), `prepareContractCallTxFromKeys`, `prepareSignAminoTxFromKeys`/`prepareSignDirectTxFromKeys`, `prepareThorchainMsgDepositTxFromKeys`, the cosmos staking family, `buildCosmosWasmExecuteMsg`, `buildCw20TransferMsg`, `preparePolkadotAssetSend`, and a few consts.

All excluded helpers remain fully reachable via their existing flat exports — nothing was removed, this only adds the grouped `sdk.X.*` handle for a hand-picked, RN-safety-audited subset. Each `tools/*/namespace.ts` file documents exactly what's excluded and why. A follow-up can widen the subset once the excluded helpers get an explicit RN-bundle safety pass (ideally backed by an actual Metro/Hermes build, which this environment couldn't run).

`bridge`/`cosmos`/`decode`/`gas`/`price` use the full existing barrel (already proven RN-safe — either previously static at the RN entry, or in `cosmos`/`gas`'s case, already exported as a namespace object at the SDK root before this PR).

## verification

- `yarn workspace @vultisig/sdk vitest run tests/unit/Vultisig.test.ts tests/unit/public` — 73 passed
- `yarn workspace @vultisig/sdk vitest run tests/unit` — 3404 passed, 3 pre-existing failures unrelated to this change (`tests/unit/crypto/normalizeSchnorrSig.test.ts`, reproduced identically on `main` without this diff)
- `yarn workspace @vultisig/sdk typecheck` — clean except 2 pre-existing `auxiliaryData`/`ISigningInput` Cardano errors in `packages/core/mpc`, reproduced identically on `main`
- `yarn workspace @vultisig/cli typecheck` — 7 pre-existing errors (stale `client-shared/dist` typings, unrelated ripple module resolution), reproduced identically on `main`
- `node_modules/.bin/eslint --config .config/eslint.config.mjs` on every touched file — clean (one import-sort autofix applied)
- `node_modules/.bin/prettier --config .config/.prettierrc --ignore-path .config/.prettierignore --check` on every touched file — clean
- `yarn knip` — no new findings for any touched/added file
- `yarn npm audit --recursive --all --severity high` — clean, no lockfile changes
- Could not run `scripts/quality-contracts.mjs` end-to-end in this environment (`yarn build:shared` hits the same pre-existing Cardano typing error above, blocking the SDK build it depends on) — added the assertions following the exact pattern of the existing `sdk.evm`/`sdk.cosmos.gov` checks and verified `node --check` on the full script

🤖 Agent-generated

## Summary by CodeRabbit

* **New Features**
  * Added grouped SDK namespaces for balances, bridging, decoding, gas, transaction preparation, pricing, and swaps.
  * Exposed these namespaces through `Vultisig` instances and supported SDK entry points, including React Native.
  * Preserved existing flat exports for compatibility.
* **Bug Fixes**
  * Improved compatibility by limiting grouped namespaces to supported platform-safe functionality.
* **Tests**
  * Added coverage validating namespace availability, aliases, and representative methods across SDK environments.


rcoderdev:
Closing this because This adds eight instance namespaces and a wider shared/RN import surface without the claimed Metro/Hermes proof, while its release note still advertises excluded members. Reintroduce only proven consumer-facing namespaces from current main.

PR #1972, opened 2026-08-16
fix(cosmos): require pre-fetched account metadata when skipping the chain fetch

### LINKED ISSUE #1809: fix(cosmos custom signing): skipChainSpecificFetch silently signs with zero sequence/account metadata
## Summary
The SDK advertises `skipChainSpecificFetch` as an offline / pre-fetched custom Cosmos signing mode, but the builders silently hardcode sequence/account metadata to zero when fetches are skipped.

## Evidence
- `packages/sdk/src/types/cosmos.ts:136-140` says `skipChainSpecificFetch` is useful for offline signing or when the caller already has pre-fetched data.
- `packages/sdk/src/vault/services/cosmos/buildCosmosPayload.ts:163-171` defaults SignAmino `accountNumber` and `sequence` to `'0'` when fetch is skipped.
- `packages/sdk/src/vault/services/cosmos/buildCosmosPayload.ts:228-246` defaults SignDirect `sequence` to `'0'` when fetch is skipped.

## Impact
Offline/pre-fetched custom Cosmos signing becomes incorrect for any account whose real sequence/account metadata is non-zero, which pushes consumers back into local wrappers instead of the SDK-owned path.

## Fix sketch
Add explicit override fields for required pre-fetched metadata and fail closed when `skipChainSpecificFetch === true` without those values.

## Audit report
https://gist.github.com/gomesalexandre/ae265486d6740dfda6a92377168c3dfb

gomesalexandre:
Closes vultisig/vultisig-sdk#1809

## what

`skipChainSpecificFetch` is documented in `CosmosSigningOptions` as being "useful for
offline signing or when you have pre-fetched data". But there was **no parameter to pass
that pre-fetched data in**, and the builders defaulted the skipped values to `'0'`:

- `buildCosmosPayload.ts` SignAmino - `accountNumber` and `sequence` both default `'0'`
- `buildCosmosPayload.ts` SignDirect - `sequence` defaults `'0'`

A Cosmos signature is bound to `(account_number, sequence)`. Zero is not a neutral
placeholder - it yields a signature the chain rejects for any account that has ever
transacted, which is every real account. The failure surfaces at broadcast, far from the
call that caused it.

The best evidence that this is a real gap and not a theoretical one is the SDK's own
consumer: `packages/rujira`'s withdraw keysign builder passes `skipChainSpecificFetch:
true`, then **throws away the resulting `blockchainSpecific` and hand-rebuilds the whole
`KeysignPayload`** with `sequence: BigInt(accountInfo.sequence)` - because the SDK's zero
was unusable. Exactly the "pushes consumers back into local wrappers" outcome the issue
describes.

## how

- `CosmosSigningOptions` gains `accountNumber?` / `sequence?`, threaded through
  `prepareSignAminoTxFromKeys` / `prepareSignDirectTxFromKeys` to both builders.
- Both builders fail closed when the flag is set without them, naming the missing field
  and why zero is not acceptable. Non-numeric values are rejected too.
- SignDirect only needs `sequence` - its `accountNumber` is already part of the input.
- **`packages/rujira` updated to pass the values it already holds**, so the base payload
  is correct the first time instead of being discarded and rebuilt.

## receipts

Runtime transcript driving both builders against a mocked chain whose account has
**accountNumber=7, sequence=41** (an account that has transacted), via temporary
uncommitted QA instrumentation (`qa-1809-repro.test.ts`, deleted before commit - not in
this diff).

**BEFORE (on main):**

```
Account's REAL on-chain state: accountNumber=7 sequence=41

--- SignAmino, skipChainSpecificFetch: true, NO pre-fetched values passed ---
  -> BUILT payload with accountNumber=0 sequence=0   <-- signature bound to the WRONG account state

--- SignDirect, skipChainSpecificFetch: true, NO pre-fetched values passed ---
  -> BUILT payload with accountNumber=7 sequence=0   <-- signature bound to the WRONG account state

--- SignDirect, skipChainSpecificFetch: true, WITH pre-fetched sequence ---
  -> BUILT payload with accountNumber=7 sequence=0   <-- caller passed 41, got 0 anyway

--- SignDirect, fetch NOT skipped (unchanged control path) ---
  -> BUILT payload with accountNumber=7 sequence=41   <-- fetched from chain
```

Note the third block: on main, **even a caller that explicitly supplies the right sequence
gets a zero**, because there was nowhere for the value to go.

**AFTER (this branch):**

```
Account's REAL on-chain state: accountNumber=7 sequence=41

--- SignAmino, skipChainSpecificFetch: true, NO pre-fetched values passed ---
  -> rejected: buildSignAminoKeysignPayload: accountNumber is required when skipChainSpecificFetch is true. Pass the pre-fetched value (e.g. from getCosmosAccountInfo) - defaulting it to '0' would bind the signature to the wrong account state and the chain would reject the broadcast.

--- SignDirect, skipChainSpecificFetch: true, NO pre-fetched values passed ---
  -> rejected: buildSignDirectKeysignPayload: sequence is required when skipChainSpecificFetch is true. ...

--- SignDirect, skipChainSpecificFetch: true, WITH pre-fetched sequence ---
  -> BUILT payload with accountNumber=7 sequence=41   <-- matches real chain state

--- SignDirect, fetch NOT skipped (unchanged control path) ---
  -> BUILT payload with accountNumber=7 sequence=41   <-- fetched from chain
```

```
      Tests  3397 passed (3397)

$ yarn test:rujira
      Tests  558 passed | 4 skipped (562)     <-- includes withdraw.test.ts / deposit.test.ts,
                                                  which exercise the keysign-builder changed here

$ yarn typecheck
exit 0

$ yarn lint
exit 0
```

`yarn typecheck` and `yarn test:rujira` both require the workspace dists to exist
(`build:shared`, `build:sdk:bundle`, `build:client-shared`, `build:rujira`). On a cold
checkout they fail with `Cannot find module '@vultisig/sdk'` / `'@vultisig/client-shared'`
before running anything - that is a missing build step, not a code failure. Results above
are after those builds.

## notes for the reviewer

- **This is a behaviour change, hence a `minor` changeset, and it deserves a real look.**
  A caller passing `skipChainSpecificFetch: true` without the new values now throws where
  it previously returned a payload. That payload was never broadcastable, so this turns a
  silent broadcast-time rejection into an actionable build-time error - but if any
  consumer is relying on the zero-sequence payload for something other than broadcast
  (a dry-run / shape probe?), it would break. I checked every in-repo caller; `rujira` was
  the only production one and it is updated here. Please sanity-check downstream
  (`mcp-ts`, `vultiagent-app`) before merge.
- The `rujira` withdraw builder still rebuilds its final payload by hand; I deliberately
  did NOT refactor that in this PR, only fed it correct inputs. Simplifying it to consume
  the base payload directly is a natural follow-up but is a bigger, separately-reviewable
  change.
- No emulator screenshot: pure library change with no user-facing surface, and the

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved offline Cosmos transaction signing by preserving pre-fetched account metadata when network fetching is skipped.
  - Added validation to reject missing or invalid account numbers and sequence values instead of silently defaulting to zero.
  - Ensured signing payloads consistently retain account metadata across supported Cosmos-based chains.

- **Documentation**
  - Clarified requirements and behavior for offline Cosmos signing options.


gomesalexandre:
Addressed in 63049ecc9 — both throws in `requirePrefetched` are now
`VaultError(VaultErrorCode.InvalidConfig, …)`.

I checked whether this was worth doing rather than just applying it, and it is, but the
reasoning is slightly different from the finding:

- **The vault path already produced a coded error.** `TransactionBuilder.wrapAsVaultError`
  converts any non-`VaultError` into `VaultError(InvalidConfig, "Failed to prepare <op>: …")`,
  so `vault.prepareSignAminoTx` callers could already branch on a code.
- **The vault-FREE path could not.** `prepareSignAminoTxFromKeys` /
  `prepareSignDirectTxFromKeys` — the entry points MCP servers and other vault-less callers
  use — are not wrapped, and those callers were getting a bare `Error`. That is the real gain.
- CLAUDE.md:85 ("Custom error classes: `VaultError`, `StorageError` with error codes") settles
  it regardless, so this follows the documented guideline.

One deliberate side effect: `wrapAsVaultError` rethrows a `VaultError` unchanged, so vault
callers no longer get the `Failed to prepare <op>:` prefix on these two errors. No context is
lost — the messages already name the builder function and the offending field.

`InvalidConfig` is the closest existing code; there is no `InvalidInput`/`InvalidArgument` in
`VaultErrorCode`. Happy to add one if you'd rather these were distinguishable from genuine
config faults.

Gate re-run on the updated branch: unit 3397 passed, rujira 558 passed, typecheck / lint /
format:check all exit 0.

NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because This fixes real Cosmos account metadata handling, but it changes fund-adjacent signing APIs and must be updated and freshly reviewed against current callers. Re-propose it from current main rather than carrying this stale exact head.

PR #1973, opened 2026-08-16
feat(sdk): expose estimatesendfee on the public vault surface

### LINKED ISSUE #1867: public API: VaultBase documents estimateSendFee() but does not expose it
## Summary
The SDK's public vault API tells consumers to call `estimateSendFee()` to compute the real fee-aware max for certain send/swap flows, but that estimator is not actually exposed on `VaultBase`.

## Evidence
- `packages/sdk/src/vault/swap-types.ts:148-150` documents that transfer-route `maxSwapable` is `0n` and callers must invoke `estimateSendFee()` to compute the real max.
- `packages/sdk/src/vault/VaultBase.ts:1289-1294` repeats the same instruction in runtime comments.
- `packages/sdk/src/vault/services/TransactionBuilder.ts:129-180` contains the estimator implementation.
- `packages/sdk/src/vault/VaultBase.ts:118` and `:318` keep `transactionBuilder` as a protected internal detail; there is no public `vault.estimateSendFee(...)` wrapper.

## Impact
This is a public API contract hole:
- SDK docs point consumers at a method they cannot call from the published vault surface
- first-party consumers are pushed toward wrapper logic, internal access, or approximate max-send math
- the exact fee-aware max path stays harder to CLI-test and harder to standardize across app/backend/CLI consumers

## Suggested fix
Add a public `VaultBase.estimateSendFee(...)` wrapper that forwards to the existing transaction builder and returns structured fee metadata suitable for max-send / preview UX.

Suggested shape:
- `feeAmountBase`
- `feeDecimals`
- `feeSymbol`
- optionally `spendAssetAmountBase` / route metadata if needed for max-send computation

## Verification target
- focused unit coverage on the new public wrapper
- CLI/runtime proof that a consumer can call the public API without reaching into protected internals

gomesalexandre:
Closes vultisig/vultisig-sdk#1867

## what

The SDK tells consumers, in two places, to call `estimateSendFee()`:

- `packages/sdk/src/vault/swap-types.ts` - a transfer route reports `maxSwapable: 0n` and
  the doc says to call `estimateSendFee()` to compute the real fee-aware max
- `packages/sdk/src/vault/VaultBase.ts` - the same instruction repeated in runtime comments

The implementation exists on `TransactionBuilder`, but `VaultBase` holds
`transactionBuilder` as a **`protected`** field (`VaultBase.ts:220`), and unlike its
neighbour `prepareSendTx` there was no public wrapper. So the documented instruction did
not compile from outside the class. Consumers were pushed toward approximate max-send math
or a cast through internals.

## how

Add the thin public wrapper immediately after `prepareSendTx`, matching that wrapper's
shape exactly - same params, forwards verbatim, returns the fee in the fee coin's base
units.

I deliberately did **not** invent the richer `{ feeAmountBase, feeDecimals, feeSymbol }`
shape the issue floats as a suggestion - see the reviewer note below.

## receipts

Two receipts, because `protected` is a **compile-time** barrier - a runtime check alone
would not prove a consumer can actually call this.

**1. Compile-level, from a consumer's point of view.** Temporary uncommitted QA file
(`packages/sdk/qa-1867-consumer.ts`, deleted before commit - not in this diff) that imports
only the public `VaultBase` type and does what `swap-types.ts` instructs:

```ts
export async function consumerComputesMaxSend(vault: VaultBase, coin: AccountCoin, receiver: string, balance: bigint) {
  const fee = await vault.estimateSendFee({ coin, receiver, amount: balance })
  return balance > fee ? balance - fee : 0n
}
```

BEFORE (on main):

```
packages/sdk/qa-1867-consumer.ts:9:27 - error TS2339: Property 'estimateSendFee' does not exist on type 'VaultBase'.
```

AFTER (this branch):

```
exit 0
```

**2. Runtime, that the wrapper actually forwards.** New unit test asserts the method is on
the prototype and that params reach the builder **verbatim** — `memo`, `destinationTag` and
`feeSettings` all change the fee, so a wrapper that reshaped or dropped them would return a
plausible-but-wrong number:

```
$ vitest run tests/unit/vault/estimateSendFeePublic.test.ts
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

```
$ yarn workspace @vultisig/sdk test:unit    Tests 3396 passed (3396)
```

(`typecheck` needs the workspace dists built first — `build:shared`, `build:sdk:bundle`,
`build:client-shared`, `build:rujira` — otherwise it fails on a missing `@vultisig/sdk`
before checking anything.)

## notes for the reviewer

- **Return shape is a judgement call I would like checked.** The issue suggests returning
  `{ feeAmountBase, feeDecimals, feeSymbol }`. I returned the raw `bigint` the underlying
  estimator already produces, because that matches every other wrapper in this file and
  adding fee-coin metadata means deriving the fee coin per chain kind — a real design
  decision, not a wrapper. If you want the structured shape, I would rather do it as its
  own PR that settles where fee-coin resolution lives. Say the word.
- No emulator screenshot: public-API surface change with no user-facing surface, and the


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because The estimateSendFee wrapper is a valid thin addition, but this branch is behind current main and the change is inexpensive to recreate. Reapply it from current main with the same forwarding and prototype coverage.

PR #1976, opened 2026-08-16
fix(jupiter): retry transient 429s from the shared proxy

### LINKED ISSUE #1738: fix(jupiter): retry transient 429s from the shared Jupiter proxy in the SDK helper
See https://gist.github.com/gomesalexandre/97058ab9cffc7535c65ceb3af93020cf for evidence. The SDK Jupiter helper was still single-shot at packages/sdk/src/tools/swap/jupiter.ts:166-182 while agent-backend-ts already carried bounded 429 retry logic in src/mastra/tools/mcp/tools/swap/jupiterSwap.ts:105-146. Impact: consumers deleting their backend-local Jupiter copy and importing the SDK regress transient shared-proxy handling back to an immediate hard failure. Fix sketch: port the bounded 429 retry into the SDK and add retry-success + retry-exhaustion tests.

gomesalexandre:
Closes vultisig/vultisig-sdk#1738

## what

The SDK's `fetchJupiter` was single-shot: one rate-limited response from the shared
`api.vultisig.com/jup` proxy failed the whole swap immediately. The proxy limits per source
IP, and a single Solana swap flow makes several calls (discovery + quote + swap-build), so a
caller whose fleet shares one egress IP trips it mid-flow.

The sharp part is the asymmetry: **agent-backend-ts already carried exactly this retry**
(`src/mastra/tools/mcp/tools/swap/jupiterSwap.ts`). So a consumer doing the thing the SDK
wants — deleting its backend-local Jupiter copy and importing the SDK helper — silently
*regressed* transient-proxy handling back to an immediate hard failure.

## how

Port the bounded retry from the backend implementation: up to 2 retries, 300ms linear
backoff, **429 only**. Other statuses still surface immediately — a 4xx (no route on
Jupiter's own side) or a persistent 5xx will not heal by re-asking, and retrying just
multiplies latency on a request that is already lost.

## receipts

Beyond unit tests, I drove this over **real HTTP**: a local `node:http` server that
rate-limits the way the proxy does, with the SDK helper pointed at it. QA instrumentation
(`packages/sdk/qa-1738-repro.ts` plus a one-line env hook on the base URL) was reverted
before commit — **neither is in this diff**; confirm via the diff stat, `jupiter.ts` has no
`QA_JUP_BASE`.

**BEFORE (on main)** — a single transient 429 kills the swap on the first call:

```
proxy 429s once, then recovers  -> 429 SURFACED: Jupiter API error (429): rate limit exceeded | 1 /quote call(s), 271ms
proxy 429s persistently         -> 429 SURFACED: Jupiter API error (429): rate limit exceeded | 1 /quote call(s), 207ms
proxy healthy (control)         -> QUOTE PHASE OK | 1 /quote call(s), 204ms
```

**AFTER (this branch)** — the transient 429 is absorbed; the persistent one is still
surfaced, bounded:

```
proxy 429s once, then recovers  -> QUOTE PHASE OK | 2 /quote call(s), 698ms
proxy 429s persistently         -> 429 SURFACED: Jupiter API error (429): rate limit exceeded | 3 /quote call(s), 1997ms
proxy healthy (control)         -> QUOTE PHASE OK | 1 /quote call(s), 204ms
```

Read the call counts: `1 -> 2` on the recoverable case (the retry fired), `1 -> 3` on the
persistent case (1 initial + 2 retries, then it gives up rather than looping), and the
healthy control is **unchanged at 1 call** — the retry adds nothing to the happy path.

"QUOTE PHASE OK" is deliberate wording, not a hidden failure: the stub returns a placeholder
`swapTransaction`, which the real Solana instruction-allowlist guard then refuses to parse.
That happens *after* the quote call this change affects, so the harness classifies it
honestly rather than claiming a full successful build.

**Unit tests — 2 of the 3 new ones fail without the fix:**

```
 × retries a 429 and succeeds on the retry (sdk#1738)
 × gives up after the retry budget and surfaces the 429 (sdk#1738)
      Tests  2 failed | 23 passed (25)
```

The third ("does NOT retry a non-429 error status") passes both before and after **by
design** — it is a guard against this change over-reaching, so it must hold on `main` too.
Flagging that explicitly so nobody reads it as a vacuous test.

```
$ yarn workspace @vultisig/sdk test:unit    Tests 3397 passed (3397)
```

## notes for the reviewer

- The retry tests use fake timers, so the suite does not pay ~0.9s of real backoff per
  exhaustion case.
- Backoff is linear (300ms, 600ms) and matches the backend implementation exactly rather
  than being re-tuned here — the point of this PR is parity, and diverging the two would
  recreate the drift it exists to remove. If you want jittered/exponential, worth doing in
  both places at once.
- No emulator screenshot: library-internal network resilience with no user-facing surface,
  and the `vultiagent-app` debug build is currently blocked on an expired `TRUSTWALLET_PAT`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved swap transaction reliability by automatically retrying temporary rate-limit responses.
  - Retries are limited and use a brief increasing delay; other errors continue to fail immediately.
  - Preserved clear error reporting when all retry attempts are unsuccessful.

- **Tests**
  - Added coverage for successful retries, exhausted retries, and non-retryable errors.


gomesalexandre:
Good catch, applied — see the follow-up commit.

You are right that the exhaustion test was under-specified: it asserted the final count
(3 calls), and a regression from 300ms linear to **0ms** — or to exponential — still ends at
3 calls, so it would have sailed straight past.

Added a dedicated test that walks each boundary with `advanceTimersByTimeAsync`:

```
1 call  before t=300     (first retry has not fired)
2 calls at    t=300      (300 * attempt 1)
2 calls before t=900     (second retry waits 600ms, not another 300ms)
3 calls at    t=900      (cumulative; 300 * attempt 2)
3 calls after runAllTimers  (and stops there)
```

Proved it is not vacuous the same way as the others — flattening
`JUPITER_RETRY_BACKOFF_MS` to `0` fails **this** test while the pre-existing ones stay green:

```
 × spaces the retries on a 300ms linear backoff (sdk#1738)
      Tests  1 failed | 25 passed (26)
```

I kept the original outcome-focused tests rather than rewriting them — they pin *behaviour*
(absorbed vs surfaced), this new one pins *schedule*, and they fail for different reasons.

Gate on the updated branch: unit 3398 passed, lint and format:check exit 0.

NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because The bounded Jupiter 429 retry is sound, but this branch is behind current main and the focused change is inexpensive to recreate. Reapply it from current main with the same pinned retry schedule and non-429 coverage.

Issue #1902, opened 2026-08-12
public API(stakekit): export action validators from supported SDK entrypoints


gomesalexandre:
## Summary

`validateStakekitActionAddress()` and `validateStakekitActionInput()` now exist as pure SDK canonicals in `packages/sdk/src/tools/defi/stakekit/index.ts`, but they are still stranded inside the module instead of being exported from the supported SDK entrypoints.

That leaves downstream consumers copying or re-declaring the same preflight rules even after the SDK ported them.

## Evidence

- Canonical implementations exist in source:
  - `packages/sdk/src/tools/defi/stakekit/index.ts:514-520` — `validateStakekitActionAddress`
  - `packages/sdk/src/tools/defi/stakekit/index.ts:524-530` — `validateStakekitActionInput`
- They are already used internally by the builders:
  - `packages/sdk/src/tools/defi/stakekit/index.ts:547`
  - `packages/sdk/src/tools/defi/stakekit/index.ts:608`
- They are already test-covered:
  - `packages/sdk/tests/unit/tools/defi/stakekit.test.ts:10-12`
  - `packages/sdk/tests/unit/tools/defi/stakekit.test.ts:575-607`
- But the public entrypoints omit them:
  - `packages/sdk/src/tools/index.ts:280-287`
  - `packages/sdk/src/index.ts:755-761`
  - `packages/sdk/src/platforms/react-native/index.ts:332-348`

## Impact

Consumers that want to validate StakeKit action inputs locally before a remote yield.xyz request still need local copies or must surface opaque remote 4xxs.

This is the recurring duplicated-not-imported drift class: the SDK already owns the pure logic, but first-party consumers still cannot import it from the published surface.

## Fix sketch

- Export both validators from:
  - `packages/sdk/src/tools/index.ts`
  - `packages/sdk/src/index.ts`
  - `packages/sdk/src/platforms/react-native/index.ts` if the validators are RN-safe
- Add public-surface regression tests proving the root and RN entrypoints expose them.

## Effort

Small.

## Artifact

- Round-134 report mirror: https://gist.github.com/gomesalexandre/eb0186344a89fefa0031b012f642a23b

PR #1900, opened 2026-08-12
fix(sdk): export stakekit canonicals on RN entry

gomesalexandre:
Closes #1902

## Summary
- export the new StakeKit canonical helpers from `@vultisig/sdk/react-native`
- cover the RN entry parity with a focused regression test
- add a changeset because this is a public-surface fix consumed by RN clients

## Why
Recent SDK work added root exports for the StakeKit parser, scan builders, and named builders, but the React Native entrypoint still stranded them on the node/root surface only. Metro resolves `@vultisig/sdk/react-native` to the hand-curated RN allow-list, so Station and other RN consumers could not import the canonical StakeKit surface without keeping app-local copies.

## Verification
- `VULTISIG_STRICT_SINGLETON=0 corepack yarn workspace @vultisig/sdk vitest run tests/unit/platforms/react-native/entry.test.ts --testTimeout=15000`
  - passed: `45 passed`
- `corepack yarn workspace @vultisig/sdk vitest run tests/unit/utils/publicExports.test.ts --testTimeout=15000`
  - passed: `50 passed`
- `corepack yarn build:sdk`
  - pre-existing blocker on clean `origin/main` in the disposable clone too:
    - `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts(72,5): error TS2353: Object literal may only specify known properties, and 'auxiliaryData' does not exist in type 'ISigningInput'.`

## Receipts
Focused RN-entry receipt proves the public surface is reachable from the actual React Native entry, not just the root/node barrel.

## Summary by CodeRabbit

* **New Features**
  * Added public exports for StakeKit action address and input validators.
  * Expanded React Native SDK access to StakeKit yield types, builders, search tools, metadata, transaction helpers, and validators.

* **Tests**
  * Added coverage confirming consistent StakeKit exports across SDK entry points.
  * Added validation tests for supported and invalid addresses and action inputs.

PR #1978, opened 2026-08-16
feat(stakekit): export the action preflight validators publicly

gomesalexandre:
Closes vultisig/vultisig-sdk#1902

> **Stacked on #1900** (`fix/r134-rn-stakekit-exports`) — base that first. Its diff is
> included below until it merges. See "why stacked" at the bottom.

## what

`validateStakekitActionAddress` and `validateStakekitActionInput` are pure canonicals in
`tools/defi/stakekit`. They are already used by the builders themselves and already
test-covered — but no public entrypoint exported them, so consumers had to re-declare the
same address/amount preflight rules after the SDK had ported them.

Exported from all three public surfaces: root `@vultisig/sdk`, the `tools` barrel, and the
React Native entry.

## receipts

**The new tests are not vacuous** — reverting only the three export lines fails exactly the
identity assertions, on every surface:

```
$ vitest run tests/unit/public/stakekitValidatorExports.test.ts \
             tests/unit/platforms/react-native/entry.test.ts
⎯⎯⎯ Failed Tests 3 ⎯⎯⎯
      Tests  3 failed | 45 passed (48)
```

With the exports in place, all 48 pass.

Two deliberate choices in how they assert:

- **Identity, not shape.** `expect(root.validateStakekitActionAddress).toBe(canonical)`
  rather than `typeof === 'function'`. The failure mode this issue is about is a consumer
  re-implementing the rules; a `typeof` check would happily pass on a divergent copy
  re-exported under the same name.
- **Plus behaviour through the public surface.** A small contract spot-check (EVM 40-hex and
  Sui 64-hex accepted, other `0x` lengths refused locally, non-`0x` families passed through
  for server-side validation), so an export accidentally wired to a *different* function is
  caught even if identity somehow held.

**Full gate, green:**

```
$ yarn workspace @vultisig/sdk test:unit    Tests 3392 passed (3392)
```

## notes for the reviewer

- **Why stacked rather than cut from `main`:** #1900 introduces the RN entry's stakekit
  export block that these two lines join. Basing on `main` would have put two PRs in the same
  block and guaranteed a conflict. This branch is cut from #1900's head, so its commits appear
  here until it lands.
- **The RN assertion lives in `tests/unit/platforms/react-native/entry.test.ts`, not in the
  new public test.** Importing the RN entry from a plain node test environment fails on
  `Cannot find package 'react-native'` (via `expo-modules-core`); that suite has the module
  harness for it. I found this by writing it the other way first and watching it fail, and
  noted the reason in the new file's header so the split does not look arbitrary.
- `minor` changeset — additive public surface.
- No emulator screenshot: export-surface change with no user-facing behaviour, and the

### DIFF
--- a/.changeset/feat-export-stakekit-action-validators.md
+++ b/.changeset/feat-export-stakekit-action-validators.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': minor
+---
+
+Export `validateStakekitActionAddress` and `validateStakekitActionInput` from the public SDK surfaces (root, `tools`, and the React Native entry). Both are pure canonicals already used internally by the StakeKit builders and already test-covered, but they were stranded inside the module - so consumers kept re-declaring the same address/amount preflight rules after the SDK had ported them.
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -1041,6 +1041,8 @@ export {
   token,
   TRC20_TRANSFER_SELECTOR,
   utxoFeeRate,
+  validateStakekitActionAddress,
+  validateStakekitActionInput,
   VerifierClient,
 } from './tools'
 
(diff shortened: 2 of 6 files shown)


NeOMakinG (approved):
LGTM (stacked on fix/r134-rn-stakekit-exports)

PR #1979, opened 2026-08-16
feat(sdk): expose the chains namespace on every platform entrypoint

### LINKED ISSUE #1840: public API: platform-specific SDK entrypoints other than React Native do not expose the documented chains tx-builder namespace
Title: public API: platform-specific SDK entrypoints other than React Native do not expose the documented chains tx-builder namespace

## Summary
`packages/sdk/src/index.ts` documents Cosmos signing primitives as shipping via `chains.cosmos.buildCosmosStakingTx` from a platform-specific entrypoint, but only the React Native entry actually exposes `chains`. Node/browser/electron/chrome consumers are forced to deep-import or reimplement the shared tx-builder surface.

## Evidence
- `packages/sdk/src/index.ts:369-371`
- `packages/sdk/src/platforms/react-native/index.ts:122`
- `packages/sdk/src/platforms/node/index.ts:117-127`
- `packages/sdk/src/platforms/browser/index.ts:62-69`
- `packages/sdk/src/platforms/electron-main/index.ts:117-124`
- `packages/sdk/src/platforms/chrome-extension/index.ts:66-73`

## Impact
This keeps the shared tx-builder contract effectively RN-only, which is the opposite of the consolidation goal. First-party Node/browser consumers cannot just consume `@vultisig/sdk`'s documented platform-entry API and instead drift toward local chain-builder glue.

## Effort
Low.

## Fix sketch
Export `chains` from the non-RN platform entrypoints as the same shared namespace already used by RN, then add entrypoint regression tests plus built-bundle runtime receipts.

## Report
https://gist.github.com/gomesalexandre/ec09fc177d4a78d7b0e600f8ebf54f03

gomesalexandre:
Closes vultisig/vultisig-sdk#1840

## what

`packages/sdk/src/index.ts:644` says, in a comment:

> Signing primitives ship via `chains.cosmos.buildCosmosStakingTx` from the platform-specific
> entry point.

Only the React Native entry actually exposed `chains`. Node, browser, electron-main and
chrome-extension did not — so first-party consumers on those platforms had to deep-import
into `platforms/react-native/chains/...` or re-glue the builders locally, which is exactly
the drift the shared namespace exists to prevent.

## how

`export { chains } from '../react-native/chains'` on each of the four remaining entries.

**Deliberately NOT on the root entry**, and I'd like that choice checked:

- The documented contract is the *platform* entry, not root — so root would be a wider
  promise than the docs make.
- `chains` is a single `const` aggregating nine chain families (`export const chains = { cardano, cosmos, evm, ripple, solana, sui, ton, tron, utxo }`).
  That shape does not tree-shake. Putting it on an already very wide root surface would
  weigh on consumers who never asked for it, whereas the four platform entries already
  `export * from '../../index'` and are the platform-committed surface.

If you'd rather it were on root too, it's a one-liner — but I did not want to make that call
silently.

## on the react-native path

The tree lives under `platforms/react-native/chains/` for historical reasons only. It is
platform-neutral in fact:

```
$ grep -rhoE "from '(react-native|expo[^']*)'" packages/sdk/src/platforms/react-native/chains/
(no matches)
```

It imports `@noble/*` and `@scure/base` and nothing platform-bound, and the file header
already calls itself "RN-safe … Vendored from vultiagent-app". The root entry likewise
already reaches into that exact path for `buildCosmosWasmExecuteTx`
(`src/index.ts:1069-1070`), so this PR is consistent with existing practice rather than
setting a new precedent. Relocating the tree out of `platforms/react-native/` would be a
sensible follow-up but is a much larger, more conflict-prone move — not folded in here.

## receipts

**The new tests are not vacuous** — reverting only the four export lines fails all eight
per-entry assertions:

```
 × node exposes the SAME chains namespace object
 × browser exposes the SAME chains namespace object
 × electron-main exposes the SAME chains namespace object
 × chrome-extension exposes the SAME chains namespace object
 × node reaches the documented chains.cosmos.buildCosmosStakingTx
 × browser reaches the documented chains.cosmos.buildCosmosStakingTx
 …
```

With the exports in place, 9 passed.

How they assert, and why:

- **Identity (`toBe`), not shape.** A per-platform re-implementation of the namespace would
  satisfy `typeof === 'object'` while reintroducing precisely the drift this namespace
  exists to prevent.
- **The documented call path specifically** — `chains.cosmos.buildCosmosStakingTx` — since
  that is the string the issue quotes from the source comment.
- **A completeness check on the family list**, so a partial re-export that happens to satisfy
  the cosmos assertion while dropping the other eight fails.

**Full gate, green:**

```
$ yarn workspace @vultisig/sdk test:unit    Tests 3403 passed (3403)
```

## notes for the reviewer

- The RN entry is intentionally not covered by the new test: importing it needs that suite's
  RN module harness (it fails with `Cannot find package 'react-native'` under a plain node
  env), and it already exposed `chains` — it is the reference here, not the regression risk.
  It is still asserted by `tests/unit/platforms/react-native/entry.test.ts`.
- `minor` changeset — additive public surface, no existing export touched.
- No emulator screenshot: export-surface change with no user-facing behaviour, and the


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because The chains namespace export is useful, but this branch is behind current main and touches every platform entrypoint. Recreate the additive aliases from current main so identity and completeness are checked against the current barrels.

PR #1980, opened 2026-08-16
feat(parse): add canonical chain[:token] asset-ref parsers

### LINKED ISSUE #1819: public API: add a canonical parser for the SDK's chain[:token] text contract
## Summary
The SDK still has no canonical parser for its own `chain[:token]` text contract, so first-party consumers keep hand-splitting asset refs locally.

## Evidence
- `clients/mcp/src/tools.ts:41-45` implements a local `parseChainToken()` with `input.split(':')`.
- `clients/mcp/src/tools.ts:124-182` uses that local parser for `swap_quote` and `swap`.
- `packages/sdk/src/tools/parse/index.ts:42-47` explicitly scopes the public parse helpers to format-only chain/ticker validation.

## Impact
Malformed multi-colon or empty-token refs can drift between first-party consumers because the SDK does not own the combined text contract.

## Effort
Medium.

## Fix sketch
Add an additive public parser (for example `parseAssetRef`) with explicit grammar and fail-closed behavior for malformed refs, then migrate MCP/CLI consumers to the shared helper.

## Report
- Round 118 report mirror: https://gist.github.com/gomesalexandre/228530656bbb96319a76d2bb1d451daf

gomesalexandre:
Closes vultisig/vultisig-sdk#1819

## what

The SDK owns a `chain[:token]` text contract but shipped no parser for it, so consumers
hand-split it. `clients/mcp`'s `parseChainToken` did `input.split(':')` and read `parts[1]`.

That is not just duplication — it is **wrong in three ways that all fail silently**. Driving
the old implementation and the new one side by side over the same refs:

```
"eth"            old-> chain=eth      symbol=undefined | new-> chain=eth ticker=(none)
"eth:usdc"       old-> chain=eth      symbol="usdc"    | new-> chain=eth ticker=usdc
"eth:usdc:extra" old-> chain=eth      symbol="usdc"    | new-> REJECTED: Malformed asset ref 'eth:usdc:extra': expected 'chain' or 'chain:token',
"eth:"           old-> chain=eth      symbol=""        | new-> REJECTED: Malformed asset ref 'eth:': ticker is required (got empty or whitespace-
":usdc"          old-> chain=         symbol="usdc"    | new-> REJECTED: Malformed asset ref ':usdc': the chain half is empty.
"  eth:usdc  "   old-> chain="  eth"  symbol="usdc  "  | new-> chain=eth ticker=usdc
```

- `'eth:usdc:extra'` — the tail is **discarded** and the ref resolves as a perfectly valid one
  naming a *different* asset. This is the sharp one.
- `'eth:'` — an **empty ticker** flows downstream as though the caller had named a token.
- `'  eth:usdc  '` — **neither half is trimmed**, so the chain is `'  eth'` and the ticker
  carries trailing whitespace. The issue did not mention this one; the comparison run surfaced
  it.

## how — two helpers, deliberately

- **`splitAssetRef`** — grammar + ticker format, **no chain resolution**.
- **`parseAssetRef`** — `splitAssetRef` composed with `parseChain`, yielding a canonical `Chain`.

The split exists for a specific reason worth reviewing. `clients/mcp` resolves chains against
the caller's **vault-scoped** known set (`resolveChainReference(name, knownChains)`), not the
SDK's global registry. My first pass migrated it to `parseAssetRef` outright — that would have
silently changed which chains those tools accept, since the two resolvers do not agree by
construction. So the client takes `splitAssetRef` (the genuinely-shared half) and keeps its own
`resolveChain`. Nothing about its chain surface moves.

Both fail closed; neither throws.

## receipts

**Old-vs-new comparison** (above) run through `tsx` against the real module, with the old
`split(':')` reimplemented alongside it.

**Unit tests: 18, covering each silent failure explicitly**, plus:

- `splitAssetRef` accepts an unrecognised chain (`'not-a-real-chain:usdc'`) — pinning that it
  does *not* pre-judge the chain, which is the whole point of the separation.
- `parseAssetRef` still accepts every alias `parseChain` does (`'Terra Classic'`, `'btc'`) —
  guarding against the new parser tightening the chain surface as a side effect.
- a never-throws sweep over `null`, `undefined`, `''`, `'::::'`, `':'`, `'   '`.

**Gate:**

```
$ yarn workspace @vultisig/sdk test:unit    Tests 3412 passed (3412)
$ yarn workspace @vultisig/mcp test         Tests   35 passed (35)
```

Two things that had to be fixed along the way, both worth knowing about:

- `yarn typecheck` initially failed with `'"@vultisig/sdk"' has no exported member named
  'splitAssetRef'`. `clients/mcp` typechecks against the **built dist**, so a new root export
  needs `yarn build:sdk:bundle` before typecheck agrees. Not a code problem, but it will bite
  anyone reproducing this locally.
- `clients/mcp` aliases `@vultisig/sdk` to `tests/__mocks__/sdk.ts` in its vitest config, so
  the new helper had to be added there. I re-exported it **from source** (matching how
  `resolveChainReference` is handled in that file) rather than stubbing it, so the client tests
  exercise the real rejection behaviour rather than an approximation of it.

## notes for the reviewer

- **`yarn test:cli` has one failing test on this branch — it is PRE-EXISTING.**
  `stderrCleanliness.test.ts > leaves stderr completely empty on a successful --version` fails
  identically on a stashed clean tree; it is a Node 26 `NO_COLOR` deprecation warning landing
  on stderr, unrelated to this diff.
- `minor` changeset — additive public surface, no existing behaviour changed.
- Migrating `clients/cli` to the same helper is a sensible follow-up; I limited this to the
  consumer the issue actually cites so the behavioural surface stays reviewable.
- No emulator screenshot: pure parsing helper with no user-facing surface, and the

## Summary by CodeRabbit

* **New Features**
  * Added public SDK helpers for safely parsing `chain` and `chain:token` asset references.
  * Added validation for malformed references, empty components, and invalid ticker formats.
  * Added canonical chain resolution with optional token symbols.
* **Bug Fixes**
  * Improved MCP asset-reference handling to reject invalid inputs and omit missing token symbols.
* **Tests**
  * Added coverage for valid references, aliases, whitespace, invalid formats, and unknown chains.


gomesalexandre:
Not taking this one — I think the premise is wrong, and I'd rather say why than quietly comply.

> The required `chain[:token]` contract rejects untrimmed input.

I can't find anything that establishes that, and **both halves this parser composes trim
already**:

`packages/sdk/src/tools/parse/tickerSchema.ts:44-49` — with the rationale stated outright:

```
 * The schema trims leading/trailing whitespace before validation so that
 * `"  ETH  "` is treated identically to `"ETH"` (common LLM output artifact).
 */
export const tickerSchema: z.ZodType<string> = z.string().transform((val, ctx): string => {
  const trimmed = val.trim()
```

`packages/sdk/src/utils/normalizeChain.ts:150` (and `:69`, which strips whitespace *inside* the
key):

```
const key = input?.trim().toLowerCase() ?? ''
```

So `parseTicker('  usdc  ')` and `parseChain('  eth  ')` both succeed today. Making
`parseAssetRef('  eth:usdc  ')` fail would mean the composite **rejects input that each of its
own halves accepts independently** — which is a defect I'd expect to get a bug report about,
not a tightening.

It would also work against the stated purpose. These parsers exist for "strings from external
sources (LLM output, HTTP request bodies, CLI args)" (`tools/parse/index.ts`), and
`tickerSchema` names LLM padding as the specific artifact it absorbs.

Worth separating two things the finding merges: **padding** and **malformedness**. This PR
fails closed hard on the malformed cases — extra separators, empty ticker, empty chain half —
because those change *which asset* is referenced. Surrounding whitespace doesn't; `'  eth:usdc  '`
and `'eth:usdc'` name the same asset. Trimming it is normalisation, not leniency about grammar.

For what it's worth, the untrimmed behaviour was one of the three bugs this PR fixes: the old
`clients/mcp` split produced `chain: '  eth'` and `symbol: 'usdc  '`, and a trailing-space
symbol then flowed downstream as a token name. Rejecting instead of trimming would fix that
particular leak too, but by refusing a request the SDK can plainly satisfy.

Happy to be overruled if there's a contract doc I've missed — point me at it and I'll change it.

NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because The asset-reference parser is coherent, but this branch is behind main and still carries an unresolved review thread about its whitespace contract. Re-propose from current main with that contract explicitly settled and the thread closed.

PR #1981, opened 2026-08-16
fix(sdk): reject burn/program destinations in buildSplTransfer

### LINKED ISSUE #1723: buildSplTransfer has no burn-destination guard while its sibling prepareSendTxFromKeys does (post-#1698)
## tl;dr

`buildSplTransfer` (`packages/sdk/src/tools/prep/splTransfer.ts`) has no burn-destination guard. `prepareSendTxFromKeys` gained one in #1698, and Solana is a covered family in `dangerousAddresses.ts` — so the two sibling paths now disagree about destinations this SDK explicitly calls unrecoverable.

Verified on `origin/main` @ `7ca8c406a` (post-#1698):

```
grep -c assertSafeDestination send.ts         3
grep -c assertSafeDestination splTransfer.ts  0
```

Built transfers, real function, no mocks:

```
CONTROL normal address                              BUILT
system program 11111111111111111111111111111111     BUILT
token program  TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA  BUILT
```

The control building is what makes the other two meaningful — the builder works, it simply does not vet the destination.

`dangerousAddresses.ts:44` names the first of those verbatim: *"Solana System Program: no private key controls this address"*.

## why it is not just a missing line

The file is honest about its role — `splTransfer.ts:105` calls itself *"the SPL escape-hatch sibling to `prepareSendTxFromKeys`"*. An SDK offering a deliberate bypass to a caller who knows what they are doing is defensible. What changed is that #1698 made the main path guarded, so "escape hatch" now means "the one that skips the fund-safety check" rather than "the one without RPC/identity requirements". That asymmetry is worth an explicit decision rather than inheriting it.

It also stops being purely an SDK concern the moment a user-facing surface calls it directly — #1719 proposes exactly that (`vultisig prep spl-transfer --to <addr>`), which is where a typo or a pasted address arrives. Flagged there separately; this issue is the underlying gap, which outlives whatever happens to that PR.

## scope

Solana is the sharp case because it is a covered family with a populated burn list. The other sibling helpers (`trc20`, `suiTokenTransfer`, `polkadotAssetSend`, `jettonTransfer`) are equally unguarded but their chains are not in `getChainDangerousReason` today, so there is no guard to inherit and nothing regresses — until a burn list is added for one of them, at which point they inherit the gap silently. `maxSend` remains the other standout, per my #1698 review.

## suggestion

Either call `assertSafeDestination(Chain.Solana, to)` inside `buildSplTransfer`, or — if the bypass is genuinely wanted at SDK level — require callers of the escape hatch to opt in explicitly (a named param like `allowUnsafeDestination: true`) so the skip is a decision at the call site rather than a default.

Not patching: whether the escape hatch should stay an escape hatch is a product call, not a mechanical fix.

gomesalexandre:
Closes #1723

## what was wrong

`buildSplTransfer` (`packages/sdk/src/tools/prep/splTransfer.ts`) had no burn-destination guard, while its sibling `prepareSendTxFromKeys` (native SOL) gained one in #1698. Solana is a covered chain family in `dangerousAddresses.ts` with a populated burn list (System Program, SPL Token Program, Wrapped SOL mint, Incinerator), so the two "same risk profile" paths disagreed about destinations the SDK already treats as unrecoverable.

The only existing destination check in `buildSplTransfer` was `to === mint` (funds-lost trap when sending to the mint account itself) — it did not catch a `to` that is a known burn/program address unrelated to the specific mint being transferred (e.g. accidentally sending USDC to the raw SPL Token Program id, or to the Solana System Program).

## why

`splTransfer.ts` documents itself as *"the SPL escape-hatch sibling to `prepareSendTxFromKeys`"*. Since #1698 made the native-SOL path guarded, "escape hatch" started meaning "the one that skips the fund-safety check" rather than just "the one without RPC/identity requirements" — an asymmetry worth closing since the guard is a one-line reuse of the exact helper the sibling already calls, not a new decision.

## fix

Call `assertSafeDestination('Solana', to)` inside `buildSplTransfer`, right after the existing `to`/`from`/`mint` pubkey-shape validation and before the `to === mint` check — matching exactly what `prepareSendTxFromKeys` does for native SOL sends.

## mutation-test proof

Added a new test asserting `buildSplTransfer` throws for each of the three Solana burn/program addresses (System Program, SPL Token Program, Incinerator).

Reverted the fix locally (removed the `assertSafeDestination` call) to confirm the new test is not vacuous:

```
 × fund-safety guards > rejects known Solana burn/program destinations (parity with prepareSendTxFromKeys)
   → expected [Function] to throw an error
 Test Files  1 failed (1)
      Tests  1 failed | 8 passed (9)
```

Restored the fix, confirmed green again:

```
 ✓ fund-safety guards > rejects known Solana burn/program destinations (parity with prepareSendTxFromKeys) 0ms
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

## verification

```
 Test Files  220 passed (220)
      Tests  3396 passed (3396)
```

`yarn typecheck` — no new errors introduced (pre-existing unrelated `cliDescription`/CLI errors present identically on `main` before this change, confirmed via `git stash` A/B).

## risk

Low. Additive guard on an escape-hatch builder that already documents itself as intentionally unsafe for RPC/identity checks — this only closes the burn-destination gap to match the sibling path's existing behavior. No change to the happy-path output shape or any existing passing test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented SPL transfers from being built when the destination is a known Solana burn address or program address.
  * Added validation for unsafe destinations, including the System Program, SPL Token Program, and Incinerator.
* **Documentation**
  * Updated transfer-builder documentation to describe destination safety validation.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

PR #1983, opened 2026-08-16
fix(sdk): resolve address-shaped token refs by address before symbol

### LINKED ISSUE #1634: fix(tokenRef): address-shaped refs should match address before symbol to close a poisoned-vault spoofing surface
Split out from the sdk#1606 review (unified token reference resolution, approved).

## what it says on the box

The unified resolver tries symbol match before address match, even when the reference is address-shaped. Vault token symbols are attacker-controlled strings persisted from on-chain discovery, so this opens a new address-spoofing surface that did not exist before this PR (address-shaped refs simply threw beforehand).

## where

`packages/sdk/src/vault/tokenRef.ts:52-54`

## failure scenario

Attacker sets a scam/airdrop token's `symbol` field to the literal string of a real token's contract address (e.g. USDC's address). Victim discovers the token into their vault (`tokens --discover`), then later runs `send --token <the-real-USDC-address> --dry-run` typing the genuine address correctly. The resolver matches the ref against the scam token's symbol field first, so the send resolves to the scam contract instead of real USDC - amount parsed with the scam token's decimals, keysign payload targets the scam contract.

Mitigating factors already present: requires the vault to already hold the poisoned token (a prerequisite step), and the confirmation preview displays the raw 42-character hex string as the "symbol" rather than a real ticker, which is a visible if subtle tell to an attentive user.

## suggested fix

When a ref matches `/^0x[0-9a-fA-F]{40}$/` (or equals a known/stored address or id exactly), try the address/id match first. Address-shaped refs previously threw entirely, so no ref that resolved before this PR changes what it resolves to.

## related, not blocking

Two vault tokens sharing a symbol (real USDC + a discovered scam token also named "USDC") resolve by array order on the send path today - pre-existing behavior, not a regression from this PR, but worth considering a hard-fail on ambiguity for the send path specifically now that `--discover` advertises more tokens into vaults.

gomesalexandre:
Closes #1634

## what was wrong

`resolveTokenRef` (`packages/sdk/src/vault/tokenRef.ts:52-54`) matches a vault's own tokens by symbol first, then by contract address/id — unconditionally, even when the ref itself is shaped like an EVM contract address.

A vault token's `symbol` field is attacker-controlled, persisted verbatim from on-chain discovery (`tokens --discover`). A scam/airdrop token can set its `symbol` to the literal string of a real token's contract address (e.g. USDC's address). If a victim has that poisoned token in their vault and later runs `send --token <the-real-USDC-address>` — typing the genuine address correctly — the resolver matched the ref against the scam token's `symbol` field first, resolving the send to the scam contract with the scam token's decimals instead of the real USDC contract.

This is a fix split out of the sdk#1606 review (unified token reference resolution) — address-shaped refs simply threw before that PR, so this spoofing surface didn't exist prior to it.

## fix

When `ref` matches `/^0x[0-9a-fA-F]{40}$/` (EVM-address-shaped), try the address/id match before the symbol match. Every other ref shape (symbols, tickers, non-EVM ids) keeps the existing symbol-first order — this only flips the order for refs that cannot legitimately mean "look up my ticker `0xA0b8…`" the way a real symbol can, so no existing resolution changes.

## mutation-test proof

Replaced the pre-existing test that had pinned the vulnerable ordering as intended behavior (`'prefers a symbol match over an address match, and user tokens over the registry'` used to assert a decoy token whose `symbol` spelled USDC's address would win over the real stored USDC token) with a dedicated spoofing-guard test.

Reverted the fix locally (restored symbol-first matching unconditionally) to confirm the new test is not vacuous:

```
 FAIL  resolveTokenRef > resolves an address-shaped ref by address, not by a same-string symbol on a poisoned token (spoofing guard)
AssertionError: expected { …(3) } to deeply equal { ticker: 'USDC', decimals: 6, …(1) }
- Expected
+ Received
  {
-   "contractAddress": "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
-   "decimals": 6,
-   "ticker": "USDC",
+   "contractAddress": "0x00000000000000000000000000000000000000aa",
+   "decimals": 18,
+   "ticker": "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
  }
```

That's the exact bug: a genuine USDC-address ref resolving to the poisoned token's contract and decimals. Restored the fix, confirmed green again (26/26 in the file).

## verification

```
$ yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/vault/tokenRef.test.ts --reporter=verbose
 Test Files  1 passed (1)
      Tests  26 passed (26)

 Test Files  220 passed (220)
      Tests  3396 passed (3396)
```

`yarn typecheck` — no new errors introduced (pre-existing unrelated `cliDescription`/CLI + ripple-module errors present identically on `main`, confirmed via `git stash` A/B on the same tree).

## risk

Low, fund-safety-positive. Purely reorders which stored token wins a lookup when the ref is shaped like an address; every non-address-shaped ref (the overwhelming majority of calls — symbols, tickers) is byte-for-byte unaffected, and the send/balance-path parity tests in the same file stay green.

### DIFF
--- a/.changeset/fix-tokenref-address-spoofing.md
+++ b/.changeset/fix-tokenref-address-spoofing.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Fix a token-spoofing gap in `resolveTokenRef`: an EVM-address-shaped ref (e.g. `send({ symbol: '0xA0b8…' })`) now matches the vault's own tokens by contract address BEFORE symbol. Previously symbol was tried first even for address-shaped refs, so a poisoned/airdropped vault token whose attacker-controlled `symbol` field was set to a real token's address string (e.g. USDC's contract address) could hijack a ref the caller typed as a genuine address, resolving the send to the scam contract with the scam token's decimals instead.
--- a/packages/sdk/src/vault/tokenRef.ts
+++ b/packages/sdk/src/vault/tokenRef.ts
@@ -12,6 +12,16 @@
  * contract address, and the user's own tokens before the well-known registry.
  * Every ref that resolved before this module existed resolves to exactly the
  * same token — the only change is that refs which used to throw now resolve.
+ *
+ * Exception, for fund safety: when `ref` is itself shaped like an EVM contract
+ * address (`0x` + 40 hex), the user's own tokens are matched by address BEFORE
+ * symbol. A vault token's `symbol` is an attacker-controlled string persisted
+ * from on-chain discovery — a poisoned/airdropped token can set its symbol to
+ * the literal address string of a real token (e.g. USDC's contract address).
+ * Trying symbol first there would let that poisoned token hijack a ref the
+ * caller typed as a genuine address. Address-shaped refs cannot ambiguously
+ * mean "look up my ticker `0xA0b8…`" the way a real symbol can, so there is no
+ * legitimate resolution this reordering could break.
  */
 import { Chain } from '@vultisig/core-chain/Chain'
 import { chainFeeCoin } from '@vultisig/core-chain/coin/chainFeeCoin'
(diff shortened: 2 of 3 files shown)


gomesalexandre:
Closing in favor of #1989, which is the complete fix for the same bug (#1634). This PR only reorders address-before-symbol matching but still falls back to symbol matching when the address lookup misses — an adversarial review (Grok 4.6) flagged this exact residual gap: a vault holding only the poisoned/scam token (not the real one) would still resolve to the scam contract via symbol fallback. #1989 skips symbol matching entirely for address-shaped refs, closing the hole completely. Verified and rebased by an independent worker who confirmed the gap while reimplementing on top of a newer tokenRef.ts. No criticism of this PR's own testing — the gap was subtle and worth catching before merge.

PR #1984, opened 2026-08-16
fix(stakekit): add cronoschain/cronos chain to the network alias map

### LINKED ISSUE #1640: fix(stakekit): network alias map missing cronoschain -> cronos
Split out from the sdk#1632 review (StakeKit network alias normalization, merged).

## what it says on the box

The shared alias map is missing `cronoschain`/`cronos chain` -> `cronos`, so the SDK's own canonical chain identifier for Cronos round-trips wrong through this table.

## where

`packages/sdk/src/tools/defi/stakekit/index.ts:412-418`

## failure scenario

`balances({network: 'CronosChain'})` sends the literal string `cronoschain` to StakeKit, an unknown slug, resulting in an empty result (`[]`) rather than the user's actual Cronos holdings. Pre-existing on both the search and balances paths (search had the same gap on main before this PR hoisted the shared map), so not introduced by sdk#1632 - but it's the one canonical SDK chain id in StakeKit's supported set that doesn't map correctly.

## suggested fix

Add `cronoschain`/`cronos chain` as aliases mapping to `cronos` in the shared normalizer, one line.

## related, non-blocking

The SDK's alias table is a strict subset of agent-backend-ts's richer, catalog-verified table (`yield-tools.ts:196-228`), which also `.trim()`s input where the SDK doesn't (`'BSC '` with a trailing space passes through unmapped). Porting the fuller table plus trim would give the SDK's public consumers the same coverage the agent already has - separate, larger follow-up if worth pursuing.

gomesalexandre:
Closes #1640

## what was wrong

`STAKEKIT_NETWORK_ALIASES` (`packages/sdk/src/tools/defi/stakekit/index.ts:412-418`) had aliases for BSC and Avalanche but was missing `cronoschain`/`cronos chain` -> `cronos` — the one canonical SDK chain identifier (`Chain.CronosChain`) in StakeKit's supported set that didn't map correctly.

`sdk.defi.stakekit.balances({network: 'CronosChain'})` sent the literal string `cronoschain` to StakeKit as the `network` query param — an unknown slug — silently returning an empty result (`[]`) instead of the user's actual Cronos holdings. Same gap on the `search()` path.

## fix

Add `cronoschain`/`cronos chain` -> `cronos` to the shared alias map, mirroring the existing `bsc`/`avax` entries in the same table.

## mutation-test proof

New test asserting `stakekitSearch({ network: 'CronosChain' })` sends `network=cronos` to StakeKit.

Reverted the fix locally (removed the two new alias entries) to confirm the test is not vacuous:

```
FAIL  stakekitSearch > normalizes the SDK canonical CronosChain identifier to the StakeKit cronos slug
TypeError: Invalid URL
Serialized Error: { code: 'ERR_INVALID_URL', input: 'undefined' }
```

(The mock fetch is never called because the unmapped `cronoschain` slug still round-trips through unnormalized — the assertion fails before reaching the value comparison, confirming the alias entry is load-bearing.) Restored the fix, confirmed green again.

## verification

```
$ yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/tools/defi/stakekit.test.ts --reporter=verbose
 Test Files  1 passed (1)
      Tests  19 passed (19)

 Test Files  220 passed (220)
      Tests  3397 passed (3397)
```

`yarn typecheck` — no new errors introduced (pre-existing unrelated `cliDescription`/CLI errors present identically on `main`).

`yarn lint:fix` / `yarn format` — no changes needed.

## risk

Trivial. Two additive alias entries in a lookup table; every other alias and every unaliased network name is unaffected.


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because PR #1992 carries the same Cronos alias fix, adds the cronos-chain variant, and tests the actual balance-discovery path. Closing this duplicate so the stronger implementation remains authoritative.

PR #1985, opened 2026-08-16
fix(stakekit): follow hasNextPage when discovering balance integrations

### LINKED ISSUE #1669: fix(stakekit): balance discovery truncates after the first 100 integrations
## summary
StakeKit balance discovery silently truncates at 100 integrations because pagination is ignored.

## evidence
-  defines  on the response type.
-  fetches one page and discards pagination.
-  hardcodes  inside .

## impact
on dense networks, integrations after the first 100 are never queried for balances, so positions can false-zero by omission.

## why this is new
current StakeKit issue history covers alias normalization, cache scoping, and amount validation, but not pagination/truncation on balance discovery.

## effort
small to medium. this is pure SDK-owned pagination logic and is unit-testable with mocked multi-page StakeKit responses.

## report
https://gist.github.com/gomesalexandre/c24bec11668bc96e91d394b17288c642

gomesalexandre:
Closes #1669

## what was wrong

`getBalances` (`packages/sdk/src/tools/defi/stakekit/stakekitApi.ts`) discovers which yield integrations to check for a wallet's positions by calling `searchYields({ network, limit: 100, apiKey })` once and using whatever it returns as the complete integration set. `YieldListResponse` defines `hasNextPage: boolean` on the response, but it was never read — `searchYields` only ever returned `.data`.

On a network with more than 100 enabled integrations, everything past the first page is silently dropped from balance discovery. A wallet with a real position in an integration beyond page 1 gets no balance query for it at all — the position reads back as "no balance" (false-zero by omission), not as an error a caller could catch.

## verifying the pagination contract

`stakek.it`'s `/v1/yields/enabled` follows the same `InfinityPaginatedDto` shape used across their API (`page`/`limit` request params, `hasNextPage` in the response) — confirmed against `@stakekit/api-hooks`'s published type declarations (`YieldGetMyYieldsParams` maps to the `/v1/yields/enabled` query key and carries `page?: number` / `limit?: number`).

## fix

Extracted `fetchYieldsPage` — the raw single-page fetch (with the existing TTL cache, now keyed on `page` too) that `searchYields` previously inlined and immediately discarded the `hasNextPage` half of. Added `getAllEnabledYieldIds`, which calls `fetchYieldsPage` in a loop, following `hasNextPage` across pages (bounded to 50 pages as a safety cap against a misbehaving API that never flips the flag) to build the complete integration id set. `getBalances` now calls `getAllEnabledYieldIds` instead of a single `searchYields` call for its discovery step.

`searchYields`/`stakekitSearch` are unchanged — they're deliberately a bounded "search" (capped at 50 results client-side in `stakekitSearch`), not a "list everything." Only the balance-discovery path, which needs exhaustiveness rather than a bounded top-N, changes behavior.

## mutation-test proof

New test `follows hasNextPage across pages when discovering integration ids, instead of truncating at page 1`: mocks a 2-page `/yields/enabled` response (page 1 `hasNextPage: true`, page 2 `hasNextPage: false`) and asserts the subsequent balances POST body carries integration ids from BOTH pages.

Reverted the fix locally (restored the single-page `searchYields(...).map(y => y.id)` call in `getBalances`) to confirm the test is not vacuous:

```
FAIL  stakekitBalances > follows hasNextPage across pages when discovering integration ids, instead of truncating at page 1
TypeError: rows is not iterable
 ❯ getBalances packages/sdk/src/tools/defi/stakekit/stakekitApi.ts:346:25
```

Without the fix, discovery stops after page 1 (mock #1), so the balances-POST call ends up consuming the mock queued for page 2's discovery response instead — a shape mismatch that throws rather than silently passing, which is itself further proof page 2 is never reached. Restored the fix, confirmed green again.

## verification

```
$ yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/tools/defi/stakekit.test.ts --reporter=verbose
 Test Files  1 passed (1)
      Tests  19 passed (19)

 Test Files  220 passed (220)
      Tests  3396 passed (3396)
```

`yarn typecheck` — no new errors introduced (pre-existing unrelated `cliDescription`/CLI errors present identically on `main`).

`yarn lint:fix` / `yarn format` — no changes needed.

## risk

Low. Additive to the balance-discovery path only; every other StakeKit surface (`search`, `buildEnter`, `buildExit`, `parseActionDisplay`) is untouched, and the existing single-page fixtures (which all set `hasNextPage: false`) exercise the new loop's exit condition on the very first iteration, so existing test behavior is preserved byte-for-byte.


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because This branch still returns a partial integration set when hasNextPage remains true at its page cap, preserving the false-zero-by-omission failure it aims to fix. A replacement should fail closed on cap exhaustion and test that case.

PR #1986, opened 2026-08-16
fix(rujira): prove range pairAddress against the authoritative FIN pair

### LINKED ISSUE #1513: rujira: range builders trust any bech32 pairAddress instead of the authoritative FIN pair
## Summary

`@vultisig/rujira` range-position builders still accept any bech32-valid `thor1...` contract sink for `pairAddress` instead of proving that the address is the authoritative FIN pair for the requested market.

## Evidence

Package-level builder only validates address shape before using it as the contract sink:

- `packages/rujira/src/modules/range.ts:481-514`
- `assertPairAddress(...)` only enforces bech32 contract form

Current unit tests cover payload shape and malformed-address rejection, but not registry binding:

- `packages/rujira/src/__tests__/range.test.ts:61-187`

Current backend mainline is already compensating for this gap locally by re-resolving the authoritative pair before it allows signable range cards to ship:

- `agent-backend-ts/src/mastra/tools/mcp/tools/rujira/range.ts:44-83`
- `agent-backend-ts/src/mastra/tools/mcp/tools/rujira/range.ts:261-267`
- `agent-backend-ts/src/mastra/tools/mcp/tools/rujira/range.ts:321-327`

## Impact

The transaction builder trusts a user-supplied contract sink after only format validation. Every downstream consumer must re-implement pair-authenticity checks or risk signing funds to a non-authoritative FIN contract.

## Why this belongs in `@vultisig/rujira`

This is not agent policy or UI logic; it is transaction-builder trust-boundary validation. The package that builds the signable calldata should fail closed itself.

## Fix sketch

- Teach `buildCreatePosition()` / `buildDeposit()` to prove the requested pair sink against the authoritative FIN pair registry/config for the `(base, quote)` market
- Reject mismatched but bech32-valid contracts
- Add regression tests for canonical-pair success and forged-pair failure

## Effort

Medium, but well-scoped and CLI/unit-testable.

## Report

Round 56 report: https://gist.github.com/gomesalexandre/b18314de06356c9b815bc5deced823fa

gomesalexandre:
Closes #1513

## what was wrong

`RujiraRange.buildCreatePosition()` / `buildDeposit()` (`packages/rujira/src/modules/range.ts`) only called `assertPairAddress()`, which validates bech32 SHAPE only — any `thor1...` contract with a valid checksum passed, including one that has nothing to do with the requested `(base, quote)` market. The package trusted a user-supplied contract sink after format validation alone.

`agent-backend-ts` was already compensating for this by re-resolving the authoritative pair itself before building signable range cards — meaning the trust-boundary check that belongs in the transaction-builder package was being duplicated (and could be silently skipped) by every downstream consumer.

## fix

Added `assertAuthoritativePair(pairAddress, base, quote)`, which resolves the authoritative FIN pair contract for `(base, quote)` via the same GraphQL registry lookup `getPairAddress()` already implements (denom/ticker fuzzy matching, ambiguity rejection, 5-minute cache), and requires an exact address match against the caller-supplied `pairAddress` before building. `buildCreatePosition()` and `buildDeposit()` call it right after their existing sync validation (`assertPairAddress`/`assertConfig`/`assertCoin`/`assertIdx`), so malformed input still fails fast without a network round-trip.

Both builders are now `async` — a breaking signature change (`RangeTransactionParams` → `Promise<RangeTransactionParams>`) — since proving authenticity requires the registry call. `buildWithdraw`/`buildClaim`/`buildTransfer`/`buildWithdrawAll` are untouched: they only carry `pairAddress` + `idx`, no base/quote asset info to prove a market against, matching the issue's own scope ("Teach `buildCreatePosition()` / `buildDeposit()`").

No other call site in this repo invokes these two methods (`RujiraRange` is only wired up via `client.range` in `packages/rujira/src/client.ts`, with zero in-repo callers of the builders themselves), so the breaking change is contained to `@vultisig/rujira`'s own public surface.

## mutation-test proof

Two new tests per builder (forged-but-bech32-valid pair, and an entirely unresolvable base/quote market), using a real deterministically-generated second valid `thor1...` address (`VALID_THOR_ADDRESS_2` from the existing `test-helpers.ts`) as the forged contract.

Reverted the fix locally (removed the `await this.assertAuthoritativePair(...)` call from `buildCreatePosition`) to confirm the new tests are not vacuous:

```
AssertionError: promise resolved "{ …(3) }" instead of rejecting
- Error { "message": "rejected promise" }
+ { "contractAddress": "thor13g83nn5...", "executeMsg": {...}, "funds": [...] }
 Tests  2 failed | 25 passed (27)
```

Without the guard, `buildCreatePosition` happily builds a fully-formed, signable transaction against the forged/unresolvable pair address — exactly the fund-safety gap this issue describes. Restored the fix, confirmed green again.

## verification

```
$ yarn vitest run src/__tests__/range.test.ts --reporter=verbose   # from packages/rujira/
 Test Files  1 passed (1)
      Tests  27 passed (27)

$ yarn test:rujira   # from repo root
 Test Files  2 failed | 21 passed (23)
      Tests  422 passed | 4 skipped (426)
```

The 2 failed suites (`deposit.test.ts`, `withdraw.test.ts`) are a pre-existing, unrelated environment issue (`Failed to resolve entry for package "@vultisig/sdk"` — needs `yarn build:sdk` first) — confirmed identical on `main` via `git stash` A/B before touching anything.

`yarn tsc --noEmit` (from `packages/rujira/`) — clean, no errors.

`yarn typecheck` (repo root) — no new errors introduced; the 5 pre-existing `cliDescription`/CLI errors are present identically on `main`.

Changeset added (`@vultisig/rujira` major, since the sync→async signature change is genuinely breaking).

## risk

Medium (breaking API surface for `@vultisig/rujira` consumers of these two specific methods), but fund-safety-positive: it closes a trust-boundary gap where a caller could sign funds into a non-authoritative FIN contract, and moves the check into the package itself rather than requiring every downstream consumer to re-implement it. No in-repo callers are affected. `@vultisig/rujira` already majors on nearly every release (per sdk#1921), so this is not an unusual bump cadence for this package.


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because This breaking range-builder change treats only the top 200 volume-ranked pairs as authoritative, which can reject legitimate lower-volume markets. Re-propose with a complete pair lookup and fresh fund-safety/API review.

PR #1987, opened 2026-08-16
fix(jupiter): reject a blank mint instead of quoting native sol

### LINKED ISSUE #1732: fix(jupiter): blank mint strings silently coerce to native SOL
## Summary
Jupiter swap building currently treats blank/whitespace mint strings as intentional native SOL by coercing them to `SOL_NATIVE_MINT`.

## Evidence
- `packages/sdk/src/tools/swap/jupiter.ts:167-171`
  - `contractAddress?.trim() || SOL_NATIVE_MINT`
- `packages/sdk/src/tools/swap/jupiter.ts:229-233`
  - the coerced mints are used directly and only identical mints are rejected.

## Impact
A consumer token-resolution bug can become a valid SOL swap quote instead of a deterministic fail-closed error, opening the door to wrong-asset quote/build behavior.

## Effort
Small to medium.

## Fix sketch
Distinguish explicit native-SOL intent (parameter omitted) from accidental blank/whitespace strings. Keep omission for native SOL, but reject blank strings with a deterministic error.

## Report
https://gist.github.com/gomesalexandre/c7bae58caec1ed98091332f24eb66c2d

gomesalexandre:
Closes vultisig/vultisig-sdk#1732

## what

`toMint` was `contractAddress?.trim() || SOL_NATIVE_MINT`, which collapses two very different
things:

- **omitted** (`undefined`) — explicit native-SOL intent
- **present but blank** (`''`, `'   '`) — almost always a token resolution that came back empty

So a consumer whose lookup missed got a valid **native SOL** swap quote instead of a
deterministic error. "Swap 100 USDC for X" with a failed USDC resolution builds a quote for
100 **SOL**.

The same-mint guard a few lines down does not save you: it only fires when both sides are
equal, and here the other side is a real mint, so the build proceeds normally.

## how

Omission still resolves to `SOL_NATIVE_MINT`. A blank string fails closed, and the error names
**which side** failed to resolve so the caller can tell a from-token miss from a to-token one:

```
Jupiter swap fromContractAddress was blank. Omit the parameter for native SOL; a blank string
usually means token resolution failed, and silently quoting native SOL would swap the wrong asset.
```

The two JSDoc lines that said *"Omit / empty for native SOL"* are updated — that documented
contract is exactly what made the collapse look intentional.

## receipts

**5 new tests; the 4 rejection cases fail without the fix:**

```
 × rejects a blank from mint instead of quoting native SOL (sdk#1732)
 × rejects a blank from (whitespace) mint instead of quoting native SOL (sdk#1732)
 × rejects a blank to mint instead of quoting native SOL (sdk#1732)
 × rejects a blank to (whitespace) mint instead of quoting native SOL (sdk#1732)
      Tests  4 failed | 23 passed (27)
```

The 5th — **"still treats an OMITTED mint as explicit native SOL"** — passes both before and
after, deliberately. It is the guard that this change tightens only the accidental case and
not the intentional one; if I had over-corrected and rejected omission too, that test fails.

Both sides are covered (`from` and `to`) and both blank shapes (`''` and whitespace-only),
since `.trim()` made whitespace behave identically to empty.

**Gate:**

```
$ yarn workspace @vultisig/sdk test:unit    Tests 3399 passed (3399)
```

## notes for the reviewer

- **This is a behaviour change for any caller currently passing `''` on purpose.** I could not
  find one — the in-repo call sites all pass a resolved mint or omit the field — but a
  downstream consumer relying on `''` as a native-SOL shorthand would now get an error rather
  than a quote. That is the intended direction (the whole point is that `''` is not
  distinguishable from a failure), and it fails loudly at build time rather than silently
  swapping the wrong asset, but it is worth a moment of your attention before merge.
- Scoped to the Jupiter builder, which is what the issue cites. If you want the same
  omitted-vs-blank distinction applied to the other chain builders' optional contract-address
  params, I am happy to sweep them — but that is a wider behavioural change than this issue
  covers and I did not want to fold it in silently.
- No emulator screenshot: pure builder logic with no user-facing surface, and the


NeOMakinG (approved):
LGTM

rcoderdev:
Closing this because This addresses a real wrong-asset risk, but it intentionally changes the public empty-string contract in fund-sensitive swap logic and requires a fresh current-main review. Re-propose the focused fix from current main with the same blank-mint regressions.

PR #1989, opened 2026-08-16
fix(tokenref): an address-shaped ref must never match a symbol

### LINKED ISSUE #1634: fix(tokenRef): address-shaped refs should match address before symbol to close a poisoned-vault spoofing surface

gomesalexandre:
Closes vultisig/vultisig-sdk#1634

## what

`resolveTokenRef` tried **symbol before contract address**. Vault token symbols are
attacker-controlled strings persisted from on-chain discovery, so a scam token whose `symbol`
field is set to the **literal text of real USDC's contract address** captured a send in which
the victim typed that genuine address correctly — amount parsed with the scam token's
decimals, keysign payload pointed at the scam contract.

## how — and why reordering alone is not enough

The obvious fix is "try address first". That closes the reported scenario but leaves a
sharper one open: if the victim does **not** already hold real USDC, the address lookup finds
nothing and a symbol fallback still hands them the scam contract.

So an address-shaped ref is **never symbol-matched at all**. It resolves by address/id, falls
through to the curated registry, or fails. There is a test for exactly that case
(`does NOT fall back to the poisoned symbol when the real token is absent`).

**Address-shapedness is deliberately shape-based and chain-agnostic** — `0x`+40 hex, `0x`+64
hex, or anything longer than the **20-char cap this SDK's own `tickerSchema` puts on a
ticker**. A ref longer than that is not a ticker by this SDK's own definition, which covers
Solana mints (32-44 base58), Tron (34), Cosmos bech32 contracts and the CLI's stored
`<Chain>-<address>` id form without guessing at per-chain encodings. I chose breadth
deliberately: a false negative here reopens the hole, whereas a false positive only means an
absurdly-long "ticker" stops resolving by symbol.

## the part worth your attention

**An existing test asserted the vulnerable ordering as intended behaviour.** It built a
`decoy` whose `symbol` is real USDC's address and asserted the decoy wins:

```ts
it('prefers a symbol match over an address match, and user tokens over the registry', () => {
  const decoy: Token = { …, symbol: USDC_LOWER, … }
  expect(resolveTokenRef(Chain.Ethereum, USDC_LOWER, [decoy, storedUsdc])).toMatchObject({ ticker: USDC_LOWER })
```

That is the attack, encoded as a guarantee. I **inverted its expectation rather than deleting
it**, so the diff shows the contract changing rather than a test quietly disappearing — and
its second half (user tokens beat the registry) is untouched, since that part is still right.

Flagging it prominently because a reviewer skimming for "did he weaken a test" should land on
this immediately and judge it themselves.

## receipts

**4 of the 5 new/changed tests fail without the fix:**

```
 × prefers user tokens over the registry, and an ADDRESS ref over a symbol (sdk#1634)
 × resolves the REAL token when the victim also holds it
 × does NOT fall back to the poisoned symbol when the real token is absent
 × treats long non-hex refs as addresses too (Solana mint shape)
      Tests  4 failed | 27 passed (31)
```

The Solana case is worth noting: I first asserted it should *throw*, and it did not — because
that mint **is** in the curated registry, so it correctly resolves to real USDC (6 decimals)
rather than the scam token's 18. The assertion now checks that outcome, which is the one that
actually matters.

**Gate:**

```
$ yarn workspace @vultisig/sdk test:unit    Tests 3400 passed (3400)
```

## notes for the reviewer

- **No ref that previously resolved changes meaning.** Address-shaped refs threw entirely
  before this resolver existed, and a plain ticker never matches an address, so the only
  behaviour that moves is the collision case — which is the attack.
- The issue also notes that two vault tokens **sharing a symbol** resolve by array order. That
  is pre-existing and not addressed here; it wants a hard-fail on ambiguity for the send path
  and deserves its own issue rather than being folded in silently.
- No emulator screenshot for this one: it is a pure resolver with no user-facing surface of
  its own. (The app build is unblocked as of today, so subsequent app-facing PRs will carry
  screenshots.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token reference resolution for address-shaped identifiers.
  * Address-like references now resolve by contract address or ID before symbol matching.
  * Prevented address-like token symbols from incorrectly matching genuine addresses.
  * Preserved existing behavior for plain ticker-based references.
  * Added support and validation coverage across Ethereum and Solana address formats.


gomesalexandre:
Rebased against latest `main` (merge commit `b0feb162c` + follow-up fix `fa0db7c1d`), resolving the conflict and pushed.

## What conflicted and why
`main` picked up a much larger `tokenRef.ts` rewrite since this PR's base (ambiguous-symbol detection, `stripLegacyTokenIdPrefix`/`normalizedTokenIdentity`/`tokenIdsMatch`/`caseInsensitiveTokenIdsMatch`, `AmbiguousTokenRefError`) that this branch's original diff didn't anticipate. The merge re-implements this PR's core fix - an address-shaped ref never symbol-matches - **on top of** that newer structure instead of reverting any of it:
- `symbolMatches`/`baseSymbolMatches` (and therefore the ambiguity check) are computed as empty arrays for an address-shaped ref, so the ambiguity check naturally becomes a no-op for them rather than a separate code path.
- Step 1 (user tokens) and step 2 (registry) both keep the `tokenIdsMatch`/`caseInsensitiveTokenIdsMatch` normalized/legacy-prefix-aware id matching `main` added, with symbol matching skipped entirely when the ref is address-shaped.
- All of `main`'s newer behavior (ambiguous-symbol rejection, legacy id normalization, case-insensitive EVM matching) is preserved for plain ticker refs - this PR's fix only changes what happens for **address-shaped** refs.

## A real bug the rebase caught
`isAddressShapedRef`'s `ref.length > 20` heuristic collided with a `main`-only feature this branch's base predates: `TokenDiscoveryService` disambiguates same-symbol tokens across contracts by appending `@<hash>` (e.g. `WIDGET@zwytdt1v08a4109c`), and that combined string can exceed 20 chars. Before the fix-up, those refs got misclassified as address-shaped and stopped resolving by their own generated ticker (`TokenDiscoveryService.test.ts > preserves case-sensitive Solana mint identities when checking symbol collisions` caught this red). Fixed by excluding any ref containing `@` from the address-shaped check - no supported chain's address format contains `@`, so it's a precise, safe exclusion. Commit `fa0db7c1d`.

## Verification
```
$ yarn vitest run --config tests/unit/vitest.config.ts tests/unit   # from packages/sdk/
 Test Files  220 passed (220)
      Tests  3424 passed (3424)

$ yarn typecheck
# only 2 pre-existing, unrelated errors in cardano.ts / compileTx.golden.test.ts (auxiliaryData
# type mismatch) - present on main before this rebase, untouched by this diff

$ node scripts/run-eslint.mjs --config .config/eslint.config.mjs 'packages/sdk/src/vault/tokenRef.ts' 'packages/sdk/tests/unit/vault/tokenRef.test.ts'
# clean, no output

$ yarn format:check
Checking formatting...
```

**Mutation test** (swap in `main`'s pre-fix `tokenRef.ts` under this PR's test file, confirm red; restore, confirm green):
```
4 of the 35 tokenRef.test.ts tests fail against pre-fix main - exactly the
"resolveTokenRef address-shaped refs (sdk#1634)" + the inverted "prefers user
tokens..." assertion, i.e. exactly the class this PR fixes. Restoring the fix:
35/35 green.
```

PR now shows `mergeable: true` / no conflicts against current `main`.

🤖 Agent-generated

NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing this cleanup candidate because the current fix is still incomplete for short asset IDs. Polkadot Asset Hub IDs such as `1984` and `1337` are valid IDs but are not classified as address-shaped, so a vault token with an attacker-controlled symbol equal to one of those IDs can still win before the genuine stored or registry ID. A future replacement should make exact stored/known ID matches bypass symbol matching regardless of ID length and include a short-ID poisoning regression test.

PR #1990, opened 2026-08-16
fix(cardano): fund CNT recipient output with a min-UTxO floor, not the token quantity

### LINKED ISSUE #429: fix(cardano): CNT keysign produces unbroadcastable bodies (zero fee + below-min-UTxO output)

gomesalexandre:
Closes #429

## what

Cardano native-token (CNT) sends produced an unbroadcastable body: the recipient output's ADA (lovelace) amount was set to the token's raw base-unit quantity instead of a min-UTxO-satisfying lovelace amount. Both co-signers converge fine on the doomed body (MPC keysign succeeds), but the network rejects it at broadcast — the user sees a signed transaction that never confirms.

## why

`packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts` reuses `keysignPayload.toAmount` for two entirely different units:

- for plain ADA sends, `toAmount` **is** lovelace — correct.
- for a CNT send, `toAmount` is the token's own base-unit quantity (e.g. `665000` for 0.665 USDM) — but the pre-fix code fed that same number into `transferMessage.amount`, the recipient output's lovelace value.

A 0.665 USDM send funded the recipient with 0.665 ADA, below Cardano's min-UTxO floor for a single-asset output. Ogmios rejects with `3125` ("insufficiently funded outputs") only after both peers already signed the same broken body — the keysign ceremony itself gives no signal anything is wrong.

Verified the bug against real WalletCore wasm before touching anything (not just reading the issue): built a CNT-send `KeysignPayload` fixture and ran it through the actual `getCardanoSigningInputs` → `preImageHashes` path used in production. The decoded CBOR body confirmed `transferMessage.amount = 665000`. Also checked whether WalletCore's own `AnySigner.plan()` catches this (issue root cause #2 speculated it might, if called) — it doesn't: `plan.error` came back `OK` with `plan.amount` echoing the same under-funded value verbatim. WalletCore's Cardano planner caps an output amount, it never raises one to the protocol minimum.

Of the issue's 5-item fix list, 3 were already resolved by later, unrelated work before I touched anything:
- (1) per-UTXO extended token data — `getKeysignUtxoInfo.ts` already uses `getCardanoExtendedUtxos`
- (2) per-input `tokenAmount` — already populated in the resolver
- (5) `useMaxAmount` forced false for CNT sends — already the case (`isSendMax = sendMaxAmount && !isTokenSend`)

Item 3 (fee=0 without calling `AnySigner.plan()`) also turned out to be a non-issue in the currently-bundled WalletCore version: `forceFee` is honored directly by `Sign()`/`preImageHashes()` without an explicit `plan()` call — confirmed empirically (the raw CBOR body showed the correct forced fee, `200000`, embedded in the tx). Only item 4 (the recipient output's lovelace floor) was still broken.

## fix

Use a static 1.5 ADA (`1_500_000n` lovelace) floor for CNT recipient outputs, per the issue's own explicit scope ("Static value covers single-CNT outputs comfortably; computing the exact min-UTxO from bundle CBOR size is a follow-up" — not needed since a send supports at most one CNT today). The token quantity in the `TokenBundle` is untouched; only the ADA side of the output changes. Plain ADA sends are completely unaffected (`toAmount` still flows straight into `transferMessage.amount`).

## mutation-test proof

Reverted the one-line fix (`isTokenSend ? CARDANO_CNT_MIN_UTXO_LOVELACE : ...` → back to `BigInt(keysignPayload.toAmount)`), reran the new tests:

```
FAIL  … > applies the same fixed floor regardless of the token quantity magnitude
FAIL  … > produces a broadcastable (min-UTxO-satisfying) pre-signing body
Test Files  1 failed (1)
     Tests  3 failed | 5 passed (8)
```

Restored the fix, all green again:

```
Test Files  1 passed (1)
     Tests  8 passed (8)
```

## verification

```
corepack yarn vitest run --config .config/vitest.core.config.ts packages/core/mpc/keysign/signingInputs/resolvers/cardano.test.ts
# Test Files  1 passed (1) / Tests  8 passed (8)

corepack yarn vitest run --config .config/vitest.core.config.ts packages/core/mpc/
# Test Files  66 passed (66) / Tests  657 passed (657)  — no golden/pinned-hash regressions elsewhere in mpc

corepack yarn lint:fix packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts
# clean

corepack yarn format packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts packages/core/mpc/keysign/signingInputs/resolvers/cardano.test.ts
# unchanged, clean

# no cardano-related errors (pre-existing, unrelated CLI `cliDescription` type errors exist on main too — confirmed via git stash, they predate this change)
```

Changeset added (`@vultisig/core-mpc` patch).

## cross-platform note

Per the issue: iOS already had an equivalent fix implemented and tested locally (commit `8379c006a` in vultisig-ios#4324) but had to revert it — shipping the fix on iOS alone breaks iOS↔extension keysign convergence, since the two platforms would then hash different bodies for the same CNT send. iOS is waiting on this SDK PR to land before re-applying theirs.

## codex review

Ran `codex exec` for a pre-PR review pass; it hit an expired OAuth token in this sandbox (401 on every backend call) rather than actually reviewing. Proceeding without it since the failure is environmental, not a signal about the diff — self-reviewed carefully instead, plus the WalletCore-wasm empirical verification above.

🤖 Agent-generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Cardano native-token transfers so recipient outputs include the required 1.5 ADA minimum.
  * Preserved the intended token quantity independently from the ADA funding amount.
  * Kept standard ADA transfers and send-max behavior unchanged.

* **Tests**
  * Added regression coverage for native-token and ADA transfer scenarios, including valid transaction planning.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

PR #1734, opened 2026-08-04
fix(stakekit): scope cached yield metadata by api key

gomesalexandre:
## Summary
- scope StakeKit enabled-yield cache entries by API-key context
- scope StakeKit yield-detail cache entries by API-key context
- add regression coverage proving different API keys no longer reuse cached StakeKit metadata

## Issues
- fixes #1730
- follows #1534

## Verification
- `corepack yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/tools/defi/stakekit.test.ts`
- `corepack yarn build:sdk`
- `corepack yarn lint packages/sdk/tests/unit/tools/defi/stakekit.test.ts`
- runtime proof via `corepack yarn tsx /tmp/vultisig-sdk-r104/.tmp-runtime-proof.ts` => `{"searchIds":["yield-a","yield-b"],"searchCalls":2,"detailNames":["Yield A","Yield B"],"detailCalls":2}`

## Notes
- implemented from a fresh shallow clone because `~/Sites/vultisig-sdk` is a dirty fork
- artifact-style round report: https://gist.github.com/gomesalexandre/c7bae58caec1ed98091332f24eb66c2d

PR #1992, opened 2026-08-16
fix(stakekit): map the canonical cronos chain id to its stakekit slug

### LINKED ISSUE #1640: fix(stakekit): network alias map missing cronoschain -> cronos

gomesalexandre:
Closes vultisig/vultisig-sdk#1640

## what

`CronosChain` is this SDK's **own canonical `Chain` id**, and it was the one canonical id in
StakeKit's supported set that did not round-trip through `STAKEKIT_NETWORK_ALIASES`.

`balances({ network: 'CronosChain' })` lowercased to the literal `cronoschain`, which StakeKit
does not know, so the request came back `[]`. To a user that reads as **"you hold nothing on
Cronos"** rather than "that network name was not understood" — the same silent-wrong-answer
shape as a fabricated zero.

One-line class of fix; added `cronoschain`, `cronos chain` and `cronos-chain` for consistency
with how `bnb chain` / `bnb-chain` / `bnbchain` are already handled.

## receipts

**The test fails without the fix:**

```
 × maps the SDK canonical Cronos id to the StakeKit slug (sdk#1640)
      Tests  1 failed | 18 passed (19)
```

and passes with it (19/19).

**Gate:**

```
$ yarn workspace @vultisig/sdk test:unit    Tests 3395 passed (3395)
```

## notes for the reviewer

- **The test asserts ONE network on purpose, and that is worth knowing.** I wrote it first as
  `it.each(['CronosChain', 'cronoschain', 'Cronos Chain'])` and two of the three failed with
  `expected 0 to be greater than 0` — not because the aliases are wrong, but because this path
  **caches**: the second and third cases never re-fetch, so there is no request left to assert
  on. A test that appears to cover three inputs while only one reaches the boundary is worse
  than one honest case, so I narrowed it and said why in a comment.
- That caching behaviour is itself the subject of sdk#1789 (enabled-yield cache shared across
  API keys) and open PR #1734; I have not touched it here.
- It also asserts across **all** network-bearing calls rather than `calls[0]`, since this path
  can issue more than one request and the first is not guaranteed to be the balances call.
- No emulator screenshot: this is a request-boundary string mapping with no user-facing surface
  of its own. Showing it in-app would need a vault holding Cronos assets, which I do not have.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

PR #2008, opened 2026-08-17
fix(sdk): check the ics-20 packet memo against its own per-chain cap

gomesalexandre:
Closes vultisig/agent-backend-ts#2458

The SDK measured only the transaction-level `cosmos_tx.memo` and said this in a comment:

```ts
// cosmoshub-4 / columbus-5 / etc. enforce `x/auth.MaxMemoCharacters` against
// the SDK **transaction**-level memo (`cosmos_tx.memo`), NOT against the inner
// ICS-20 packet `memo` (which is unbounded).
```

The packet memo is not unbounded. columbus-5 caps it, and a Skip PFM leg carries an **empty** outer memo with the whole payload inside `MsgTransfer` - so the outer-only check reads 0 bytes and admits a route that dies at broadcast with sdk code 12, after the signing ceremony.

## the part I got wrong first, because it matters

My first pass on this branch measured the packet memo and compared it against `getCosmosMemoMaxBytesByChainId` - the **outer** cap, 256 bytes on columbus-5. That is wrong, and agent-backend-ts already knew it. Its `skip-swap.ts` carries the measurement and an explicit warning I had not read, because I ported from mcp-ts instead of from the repo the issue was filed against:

> 152 real columbus-5 Skip envelopes (2026-05-30 -> 2026-07-18):
> **80 successes: packet memo <= 1001 bytes. 72 failures: packet memo >= 1122 bytes.**
> Perfectly separated -> true cap bracketed in (1001, 1122].
> *"DO NOT default this table to 256 or reuse `getCosmosMemoMaxBytesByChainId` for packet memos: the 80 WORKING USTC<->LUNC routes carry packet memos of 698-1001 bytes on columbus-5 itself; a 256B packet cap would reject every one of those healthy routes."*

The live receipt I posted on the first version was a **778-byte** route. That is inside the working range. I had proven the fix by rejecting a route that broadcasts perfectly well - the exact over-rejection failure mode I had written guard tests against two paragraphs earlier.

So: this now uses two caps for two fields, matching abts.

## how

- **`getCosmosPacketMemoMaxBytesByChainId`** lands in `cosmosMemoCap.ts` next to the outer-memo canonicals - where the issue asked for it, so backend / app / CLI read one table. columbus-5 → 1024 (the round value inside the measured bracket); everything else → 32768, a permissive ibc-go-scale ceiling rather than the 256 outer default, because the packet memo is a structurally different field that legitimately carries PFM payloads.
- `getCosmosLegMemoInfos` returns `memoBytes` and `packetMemoBytes` **separately**, never combined. Collapsing them into one `max()` is precisely what makes the outer cap swallow the packet memo.
- `firstCosmosLegOverMemoCap` checks outer against the outer cap, then packet against the packet cap, and reports `memo_field: 'outer' | 'packet'`. They are different failure modes and the error now says which.
- `TextEncoder`, not `Buffer.byteLength` - `cosmosMemoCap` documents that this path stays RN-safe. Reused `CosmosMsgType.MSG_TRANSFER_URL` rather than adding another copy of that type-url literal.

## receipts

Live Skip API, both corridors, read-only, no signing:

| corridor | packet memo | before (main) | after |
|---|---|---|---|
| LUNC → ATOM (`columbus-5` → `cosmoshub-4`) | 778B | admitted | **admitted** |
| LUNC → USDC (`columbus-5` → `1`) | 1419B | **admitted** | **rejected at 1024** |

```console
# after - the healthy corridor, which my first version wrongly rejected
LIVEVERDICT {"ok":true,"multi_tx":false}

# after - the corridor that genuinely cannot broadcast
LIVEVERDICT {"error":"skip_source_memo_too_long",
 "message":"Skip route's cosmos leg ICS-20 packet memo is 1419 bytes, but columbus-5 enforces a
            1024-byte limit. Broadcast would fail with sdk code 12 \"memo too long\" after signing. ...",
 "source_chain_id":"columbus-5","memo_bytes":1419,"memo_max_bytes":1024,"memo_field":"packet"}

# before (main, same live route)
LIVEVERDICT {"ok":true}
```

Worth noting what those two numbers do on their own: 778 and 1419 land either side of the (1001, 1122] bracket, and the corridors match too - abts measured all 80 successes as USTC↔LUNC and all 72 failures as LUNC→EVM. That is live traffic today independently reproducing the Metabase dataset, not a restatement of it.

The live driver was a temporary uncommitted test hitting the real API; deleted, not in this diff. Committed tests are all mocked and offline.

**unit-level.** 27 cases. The ones that matter most are not the rejection but the admission:

```ts
it.each([698, 850, 1001])('admits a %dB packet memo - the measured working-route range', ...)
it('admits a packet memo exactly at the cap and rejects one byte over', ...)
it('does not reject a large packet memo on a chain with no measured cap (cosmoshub-4)', ...)
it('still applies the 256B outer cap to the top-level memo, reported as the outer field', ...)
```

Every one of those `it.each` cases fails against my first version. They exist so nobody can "simplify" this back into a single cap without CI stopping them.

## gate

| check | result |
|---|---|
| `yarn workspace @vultisig/sdk test:unit` | 3424 passed |
| `skipSwap.test.ts` | 27 passed |
| `cosmosMemoCap.test.ts` | 12 passed |
| `yarn typecheck` | exit 0 |
| `yarn lint` | exit 0 |

## follow-ups (not in this PR)

1. **mcp-ts has the bug I just wrote and reverted.** `src/tools/swap/skip-swap.ts` takes `max(outerMemo, packetMemo)` for columbus-5 and compares it to `COSMOS_MEMO_MAX_BYTES_BY_CHAIN_ID['columbus-5']` = 256, so it rejects the same 698-1001B healthy routes. Its comment still cites the 2026-06-01 cosmoshub-4/Keplr probe that abts's 152-envelope dataset superseded. Filing separately - it is a live over-rejection, not a docs nit.
2. abts importing this helper and dropping its local table, once a version carrying it is published. Its own comment already flags the table as "no SDK equivalent yet".

## risk

Low. Against `main` this only ever rejects more, never admits more, and the only chain whose behaviour changes is columbus-5 above 1024 bytes on the packet field - routes already guaranteed to fail at broadcast. The outer-memo path is byte-for-byte unchanged in behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cosmos ICS-20 packet memo validation with chain-specific limits.
  * Separately enforce packet and outer transaction memo limits.
  * Clarified overflow errors by identifying the exceeded memo field.
  * Reject malformed transaction envelopes and invalid transfer messages.
  * Support larger packet memos on chains without a measured limit.
* **SDK**
  * Exposed Cosmos memo limits, XRP address tools, Cardano TTL helpers, source-channel resolution, and swap utilities.
* **Tests**
  * Added coverage for memo boundaries, malformed data, filtering, and multiple packet memos.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev (approved):
Reviewed the repaired current head. ICS-20 packet memos use the correct per-chain cap, malformed MsgTransfer bodies now fail closed with structured errors, and current-main integration is coherent.

rcoderdev (approved):

rcoderdev (commented):
This adds a new public API to the independently published `@vultisig/core-chain` package, but the changeset only versions `@vultisig/sdk`. Direct core consumers therefore will not receive the canonical Cosmos memo-cap helper until an unrelated future core release. Please add an `@vultisig/core-chain` changeset entry; because this is a new public API, a minor bump is proportional.

rcoderdev (approved):

rcoderdev (approved):

rcoderdev:
Closing this because The packet-memo failure merits a fix, but this branch relies on a chain cap we cannot verify and adds a public core-chain API without its release entry. Finishing needs cap evidence, package release work, integration and consumer QA; the only named consumer is the separate agent backend. We are closing this version while leaving the issue open.

PR #2010, opened 2026-08-17
fix(sdk): resolve swap(max) through maxswapable instead of the full balance

### LINKED ISSUE #1864: fix(swap): vault.swap(max) ignores fee-aware maxSwapable and can over-commit native balance
## Summary
`VaultBase.swap({ amount: 'max' })` still quotes and prepares with the full source balance instead of honoring the SDK's own fee-aware `maxSwapable` result.

## Evidence
- `packages/sdk/src/vault/VaultBase.ts:2011-2034` resolves `amount:'max'` to the full balance, then quotes/prepares with that full amount.
- `packages/sdk/src/vault/VaultBase.ts:1280-1307` already computes a safer `maxSwapable` enrichment for quotes, including `0n` on native transfer routes whose source-chain fee is not knowable at quote time.
- Pre-fix runtime probe on the exact head showed `swap(max)` calling `getSwapQuote(amount:'1.0')` and `prepareSwapTx(amount:'1.0')` even though the quote returned `maxSwapable: 0.999 ETH`.

## Impact
Native `max` swaps can still over-commit funds and hit avoidable insufficient-funds failures at prepare/broadcast time. For transfer-route native swaps, the wrapper should fail closed instead of pretending the full balance is safe.

## Fix sketch
- Quote once at the full balance to learn `maxSwapable`.
- Fail closed if `maxSwapable <= 0n`.
- Re-quote/prepare with the fee-aware amount when `maxSwapable < balance`.

## Effort
Small/medium.

gomesalexandre:
Closes vultisig/vultisig-sdk#1864

`swap({ amount: 'max' })` resolved `max` to the full source balance and then **quoted and prepared with it**, never reading the fee-aware `maxSwapable` that its own `getSwapQuote` computes two methods away:

```ts
if (amount === 'max') {
  const bal = await this.balanceService.getBalance(fromChain, fromToken.contractAddress)
  resolvedAmount = this.formatUnits(BigInt(bal.amount), fromToken.decimals)   // <- the whole balance
  ...
}
```

On a native swap that over-commits by exactly the network fee. The caller is told the swap is viable, signs it, and it dies at prepare/broadcast with insufficient funds.

`send({ amount: 'max' })` has always resolved its ceiling through `getMaxSendAmount` before building. This is `swap` catching up to its sibling, not a new policy.

## how

Quote once at the full balance to learn `maxSwapable`, then commit at that amount.

- **Token routes** (`maxSwapable === balance`, because gas is paid in the native asset): the probe **is** the final quote and is reused. The common path costs no extra round-trip, and there is no window for the quote to drift between check and commit.
- **Native non-transfer routes** (`maxSwapable === balance - networkFee`): re-quote and prepare at the fee-aware amount.
- **Deposit-channel transfer routes** (`maxSwapable === 0n`, source-chain fee only priced at broadcast time): fail closed with an actionable error.

`0n` there means "not computable from this quote", not "nothing is swappable", so the message says so rather than claiming a zero balance. A `typeof !== 'bigint'` check sits alongside the `<= 0n` so a partial quote or an override lands in the fail-closed branch instead of falling through it into an over-commit.

## receipts

New `swapMax.test.ts` drives the real `VaultBase.swap` via `Object.create(VaultBase.prototype)` with mocked collaborators. That indirection is the point: the existing `swap` tests in `compound-wrappers.test.ts` call the service mocks directly and never invoke the wrapper, so **they pass with the bug fully present**.

**before** (fix stashed, tests unchanged):

```console
$ vitest run tests/unit/vault/swapMax.test.ts
 × quotes AND prepares at maxSwapable, not the full balance, for a native swap
   AssertionError: expected '1.0' to be '0.999'

 × fails closed when the route cannot price its source-chain fee (maxSwapable 0n)
   AssertionError: promise resolved "{ dryRun: false, …(3) }" instead of rejecting

   Tests  2 failed | 4 passed (6)
```

Read the second one closely. `dryRun: false` means the promise did not merely fail to throw - the whole flow ran to completion, **signing and broadcasting** a transaction that cannot cover its own fee. That is the opposite of failing closed.

**after**:

```console
$ vitest run tests/unit/vault/swapMax.test.ts
   Tests  6 passed (6)
```

The four that pass in both states are the guards on what must NOT change: an explicit amount never touches the balance service, a zero balance still throws before any quote, the token path still issues exactly one quote, and `dryRun` returns the fee-aware quote.

## a fixture this surfaced

`BalanceService.test.ts` mocked `getSwapQuote` as `{ provider: 'test' }`. The real method always returns `balance` and `maxSwapable` too, so the double was asserting against a shape the method never produces - it only survived because nothing read those fields. Under this change it crashed in `formatUnits` on `undefined`. Fixed the mock to model the real contract; its own assertion (`amount: '5.0'`) is unchanged and still passes, because ALPHA is a token and the whole balance stays swappable.

I took that crash as the signal for the `typeof` guard above rather than just patching the fixture and moving on.

## gate

| check | result |
|---|---|
| `yarn workspace @vultisig/sdk test:unit` | 3420 passed |
| `swapMax.test.ts` | 6 passed |
| `BalanceService.test.ts` | passing, fixture corrected |
| `yarn typecheck` | exit 0 |
| `yarn lint` | exit 0 |

No live-network receipt on this one: `VaultBase.swap` needs a vault, and this box has none (the `.vult` referenced by `DETOX_IMPORT_VAULT_FILE` is not here). The mocked harness drives the real method end to end, and the before/after above is a genuine behaviour flip rather than a restated assertion, but I am flagging the gap rather than implying more coverage than exists.

## scope I deliberately did not take

The issue's sketch says fail closed on `maxSwapable <= 0n`, and that is what this does. The tempting extra step is computing a real ceiling for transfer routes via `estimateSendFee` / `getMaxSendAmount`, which would make `swap('max')` actually **work** on UTXO chains instead of erroring.

I left it out on purpose. It needs the deposit address and swap memo threaded out of the quote and into UTXO fee estimation, and I have no funded vault here to verify the result against. Getting UTXO fee math subtly wrong on a fund path is how funds get stranded, and a clear error today is strictly better than a confident wrong number. Worth its own PR with a real BTC receipt.

## risk

Low, and one-directional on the amount: `max` can now only resolve to less than or equal to what it resolved to before, never more. Explicit amounts are untouched (asserted). The one behavioural regression is that `swap('max')` on a native deposit-channel route now raises instead of proceeding - but that path did not work before, it just failed later and less clearly, after taking a signature.


NeOMakinG (approved):
LGTM

PR #2012, opened 2026-08-17
fix(sdk): scope the stakekit yield caches by api credential

### LINKED ISSUE #1534: stakekit: /yields/enabled cache ignores API-key scope
## Summary
StakeKit `/yields/enabled` results are API-key scoped, but the SDK caches them under a query-only key and can therefore serve one project's enabled product set to another caller.

## Evidence
- `packages/sdk/src/tools/defi/stakekit/stakekitApi.ts:214-219` explicitly documents that `/yields/enabled` returns only the products enabled for the current API key.
- `packages/sdk/src/tools/defi/stakekit/stakekitApi.ts:210-225` cached those results under `yield:search:${query}` with no auth scope.

## Impact
In a shared SDK process, one caller can warm the cache and a later caller with a different key (or no key) can receive the wrong enabled-yield set. That is both correctness drift and cross-tenant leakage of product visibility.

## Fix sketch
- Scope `/yields/enabled` cache entries by auth context (`apiKey` or public).
- Add regression coverage proving two different API keys do not share the same cached result.

## Effort
Small.

## Report
- Round 63 report: https://gist.github.com/gomesalexandre/9cf8fd20945f291ab8840e01a719e8b3

### LINKED ISSUE #1789: fix(stakekit): enabled-yield cache is shared across API keys
## Summary
StakeKit enabled-yield cache entries are keyed only by query, even though `/yields/enabled` is API-key scoped.

## Evidence
- `packages/sdk/src/tools/defi/stakekit/stakekitApi.ts:214-225` explicitly documents `/yields/enabled` as project/API-key scoped.
- The cache key omits auth scope and was only `yield:search:${query}`.

## Impact
One caller can poison another caller's allowed-yield set for five minutes, leaking/suppressing opportunities across API keys.

## Effort
Small.

## Fix sketch
Scope the cache key by API key presence/value (or stable hash) and keep the public API unchanged.

## Report
- Round report: https://gist.github.com/gomesalexandre/951024d12c935ea998025e742ec8d31f

gomesalexandre:
Closes vultisig/vultisig-sdk#1789
Closes vultisig/vultisig-sdk#1534

The file already documents the problem, immediately above the line that causes it:

```ts
const cacheKey = `yield:search:${query.toString()}`      // <- no credential
...
// `/yields/enabled` returns ONLY the products this project's API key
// is allowed to deposit into via the actions/enter endpoint.
```

So the first caller populates the entry and every subsequent caller is served that project's allowed set for the remaining TTL. It fails in both directions at once: project A's enabled products leak to project B, and project B's own products go missing.

Five minutes is long enough for this to be a live product bug rather than a race - a backend serving several projects hands out the wrong yield set on essentially every request after the first.

## how

`apiKeyScope(apiKey)` prefixes the cache key.

I scoped **both** cached endpoints, not only `searchYields` which the issue names. The rule worth holding is: *any cache whose upstream request carries an auth header must key on that auth scope, because the response is a function of the credential.* Deciding endpoint-by-endpoint from which ones are known to vary today means silently re-deciding it every time the upstream changes. `getYield` sends `X-API-KEY` too, so it gets the same treatment. Those are the only two cached calls in the module - the other four API-key calls (`/actions/enter`, `/actions/exit`, balances, MCP) do not cache.

Two details worth stating:

- **The key never enters the cache key.** Cache keys end up in debug output and dumps; a credential must not be reconstructible from one. Truncated SHA-256 instead. 64 bits is ample when the only requirement is that distinct keys not collide.
- **`anon` is its own scope**, not a fallback that shares with keyed callers. An unauthenticated request hits the same endpoint and genuinely gets a different answer, so it must not read or write a keyed entry.

## receipts

**before** (fix stashed, tests unchanged):

```console
$ vitest run tests/unit/tools/defi/stakekit.test.ts
 × does not serve one API key's enabled set to a different key
   AssertionError: expected "vi.fn()" to be called 2 times, but got 1 times

 × treats an unauthenticated call as its own scope, not a keyed one
   AssertionError: expected "vi.fn()" to be called 2 times, but got 1 times

   Tests  2 failed | 20 passed (22)
```

One fetch for two different credentials is the bug, stated precisely: the second caller never reached the network and was handed the first caller's data.

**after**:

```console
$ vitest run tests/unit/tools/defi/stakekit.test.ts
   Tests  22 passed (22)
```

The other two new cases pass in **both** states on purpose - they are the guard against overcorrecting. Scoping a cache is only correct if it still caches for the caller it belongs to, so one asserts a repeated same-key query is served from cache (1 fetch, not 2), and one asserts incidental whitespace around a key is not treated as a different credential. Without those, "fix the cache leak" quietly becomes "stop caching", which just moves the damage to the upstream.

The module-level cache persists across tests in this file, so each new case uses a `type` value no other test uses rather than reaching into cache internals to reset them.

## gate

| check | result |
|---|---|
| `yarn workspace @vultisig/sdk test:unit` | 3418 passed |
| `stakekit.test.ts` | 22 passed |
| `yarn typecheck` | exit 0 |
| `yarn lint` | exit 0 |

No live receipt: exercising this for real needs two StakeKit project API keys with genuinely different enabled sets, and I have neither. The before/after above is a real behaviour flip through the public `stakekitSearch` entry point rather than a unit test of the helper in isolation, but I would rather name the gap than imply coverage I do not have.

## risk

Low. The only change is more cache keys, never fewer: entries that used to collide now do not. Worst case is a cold cache the first time each credential calls, which is exactly what correctness requires here. No public API change - `apiKeyScope` is module-private and every signature is untouched.


gomesalexandre:
A parallel worker independently found and fixed the same defect under a separate issue number (#1534 — same file, same function `searchYields`, same root cause). Folding it in here rather than opening a duplicate. Worth noting: that version scoped the cache key with the raw `params.apiKey` string; this PR uses a truncated SHA-256 hash instead, since a cache key can end up in debug output/dumps and a credential shouldn't be reconstructible from one. This PR also covers the sibling `getYield` cached endpoint, which the other fix didn't touch.

NeOMakinG (approved):
LGTM

PR #2013, opened 2026-08-17
test(skip): lock svm_tx rejection as intentional fail-closed behavior

### LINKED ISSUE #1363: Skip Cosmos-Solana (svm_tx) routes are unreachable - missed corridor
Parent epic: #1356

From the whole-stack audit round 7 (2026-07-17). Severity P3 CONFIRMED, missed capability (not a bug).

**Anchor:** packages/sdk/src/tools/swap/skip/skipSwap.ts:568-585 (shapeUnsignedMsgs), packages/sdk/src/tools/swap/skip/chainMapping.ts

## what
Skip's msgs responses can return svm_tx (Solana) wrappers alongside evm_tx/cosmos_tx; shapeUnsignedMsgs only handles the latter two and throws otherwise. Verified this fails closed: the custody-chain guard rejects any Solana-touching route before shapeUnsignedMsgs is reached (Solana swaps route through Jupiter, not Skip), and chainMapping.ts has no Solana entries at all.

## failure scenario
Not a fund-safety issue - a missed capability. Skip can do Cosmos-Solana USDC routes (Noble/CCTP) this integration cannot reach. A user moving USDC from a Cosmos chain to Solana has no agent path today.

## fix
Add an explicit svm_tx rejection test to lock the fail-closed behavior as intentional, then scope the Cosmos-Solana capability separately (chain mapping + signing dispatch for svm_tx) if the corridor is wanted.

gomesalexandre:
Closes #1363

## what

Skip's `/msgs_direct` response can return an `svm_tx` (Solana) wrapper alongside `evm_tx`/`cosmos_tx`, but this SDK's Skip integration only parses the latter two. The issue (P3, "missed capability, not a bug") asked to verify and lock in the fail-closed behavior with an explicit test before scoping the actual Cosmos→Solana corridor as a separate feature.

## verification

Read both validation layers to find exactly where an `svm_tx`-only leg gets rejected:

- `validateTxEnvelopes` (`packages/sdk/src/tools/swap/skip/skipSwap.ts:1037-1057`) runs **before** `shapeUnsignedMsgs` and requires exactly one of `evm_tx`/`cosmos_tx` per leg (`hasEvm === hasCosmos` → reject). An `svm_tx`-only leg has neither, so `hasEvm=false, hasCosmos=false` → rejected here with `skip_msgs_tx_malformed`.
- `shapeUnsignedMsgs`'s own fallback throw (`:585`) is consequently unreachable for this shape in practice — `validateTxEnvelopes` catches it first.

So the fail-closed behavior the issue describes is real, just one layer earlier than the issue's anchor. No production code change needed.

## fix

Added the regression test the issue explicitly requested, following the existing `runSkipSwap` mocked-fetch pattern already used for the sibling "malformed leg (`cosmos_tx: null`)" test a few lines above:

```ts
describe('svm_tx (Solana) legs are refused, not silently mis-parsed (sdk#1363)', () => {
  it('rejects a route whose only leg is an svm_tx wrapper', async () => {
    // mocked route -> mocked msgs response with only an svm_tx leg
    const out = await runSkipSwap(baseArgs)
    expect(out.ok).toBe(false)
    if (!out.ok) expect(out.envelope.error).toBe('skip_msgs_tx_malformed')
  })
})
```

No changeset — test-only change under `packages/sdk/tests/`.

## out of scope

Scoping the actual Cosmos-Solana capability (chain mapping in `chainMapping.ts` + signing dispatch for `svm_tx`) is a real feature, not a bug fix — left open for whoever wants that corridor, per the issue's own fix section.

## verification

```
corepack yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/tools/swap/skipSwap.test.ts
# Test Files  1 passed (1) / Tests  18 passed (18)

corepack yarn lint:fix packages/sdk/tests/unit/tools/swap/skipSwap.test.ts
corepack yarn format packages/sdk/tests/unit/tools/swap/skipSwap.test.ts
# both clean

# no errors from this file
```

## codex review

`codex exec` still hit an expired OAuth token in this sandbox (same as the other PRs opened this session). Proceeding without it — self-reviewed by tracing both validation layers (`validateTxEnvelopes` and `shapeUnsignedMsgs`) to confirm exactly where and why the rejection fires, matching the existing sibling test's pattern for the mocked-fetch setup.

🤖 Agent-generated


NeOMakinG (approved):
LGTM

PR #2015, opened 2026-08-17
feat(sdk): export the cardano transaction-validity policy

gomesalexandre:
Unblocks vultisig/vultiagent-app#2538

Cardano TTL policy lived inside the keysign resolver and nowhere else:

```ts
// packages/core/mpc/keysign/chainSpecific/resolvers/cardano.ts
const ttl = currentSlot + BigInt(cardanoSlotOffset)
```

`cardanoSlotOffset` is not exported from the root entry, not from the RN entry, not from anywhere a consumer can reach. So an app building a Cardano send has no way to import the policy - and vultiagent-app hardcoded its own:

```ts
// vultiagent-app src/services/cardanoTx.ts:287
const ttl = currentSlot + 7200 // ~2 hours
```

720 vs 7200. A **10x** difference in how long a signed Cardano send stays valid.

## why this is worse than a duplicated constant

The SDK does not just build the TTL - it also **judges** it. `getCardanoTtlFreshnessError` refuses to broadcast anything within `cardanoBroadcastTtlSafetyMargin` slots of expiry:

```ts
const minimumFreshTtl = currentSlot + BigInt(cardanoBroadcastTtlSafetyMargin)
if (ttl <= minimumFreshTtl) return new Error('Cardano transaction TTL is expired or too close to expiry; ...')
```

So the two consumers build against different policies while a single arbiter grades both. That is the shape that makes duplication expensive rather than merely untidy.

## what this PR does

- Exports **`getCardanoSendTtl`**, `cardanoSlotOffset` and `cardanoBroadcastTtlSafetyMargin` from the root and React Native entries.
- Points the keysign resolver at the helper, so the offset has one implementation instead of being re-derived at each call site.
- Documents the two constants as the pair they are: the usable signing window is `offset - safetyMargin`, not `offset`.

Exporting a **function**, not just the number, is the point. A consumer handed `cardanoSlotOffset` still writes `currentSlot + BigInt(offset)` itself and can still get it subtly wrong; one handed `getCardanoSendTtl(currentSlot)` cannot.

## what this PR deliberately does NOT do

**It does not change 720, and it does not touch the app.**

Aligning the app to the SDK means shortening its send-validity window from ~2 hours to ~11 usable minutes (720 slots minus the 60-slot broadcast margin, at 1 second per slot). That is a real product decision about MPC ceremony latency - a fast vault signs well inside 11 minutes; a secure vault waiting on a second device plausibly does not, and would produce sends that fail the freshness guard with "rebuild the transaction and retry".

Deciding which value is correct is not something an export PR should do as a side effect, and I have no funded Cardano vault here to time a real ceremony against. So: make the policy importable now, raise the direction question on the issue with the numbers, and let the value move in its own PR with a ceremony-latency receipt behind it.

## receipts

Built the SDK and imported from the built bundle - the thing a consumer actually resolves, not the source:

```console
$ yarn build:sdk:bundle          # exit 0
$ node -e "const m = require('./packages/sdk/dist/index.node.cjs'); ..."
getCardanoSendTtl: function | offset: 720 | margin: 60 | ttl(1000000n): 1000720

$ grep -c getCardanoSendTtl packages/sdk/dist/index.react-native.js
1
```

The RN bundle is the one vultiagent-app consumes, so that second line is the specific thing #2538 is blocked on.

On `main` none of these symbols exist on either entry - that is the whole issue, so the before-state is an absence rather than a differing value.

**Unit** - `packages/core/chain/chains/cardano/config.test.ts`, 4 cases. The one worth reading is not an assertion about 720:

```ts
it('leaves a usable signing window after the broadcast safety margin', () => {
  expect(cardanoSlotOffset).toBeGreaterThan(cardanoBroadcastTtlSafetyMargin)
  expect(cardanoSlotOffset - cardanoBroadcastTtlSafetyMargin).toBeGreaterThanOrEqual(600)
})
```

If the offset is ever lowered to at or below the margin, the SDK would build transactions that can never be broadcast at all - valid on paper, refused in practice by its own guard. Pinning the two numbers independently would not catch that; pinning the relationship does. It also means the follow-up PR that retunes the offset gets an immediate signal if it goes too far the other way.

## gate

| check | result |
|---|---|
| `yarn workspace @vultisig/sdk test:unit` | 3415 passed |
| `cardano/config.test.ts` | 4 passed |
| `yarn build:sdk:bundle` | exit 0 |
| `yarn typecheck` | exit 0 |
| `yarn lint` | exit 0 |

## risk

Very low. Additive exports plus one call-site swap that is arithmetically identical (`currentSlot + BigInt(cardanoSlotOffset)` → `getCardanoSendTtl(currentSlot)`, same expression, same types, covered by the unit test above). No value changes, no behaviour changes, nothing removed.

## Summary by CodeRabbit

* **New Features**
  * Added Cardano transaction-validity helpers, including TTL calculation, slot offset, and broadcast safety margin, to SDK entry points.
  * Expanded React Native capabilities with mnemonic utilities, seedphrase validation and import helpers, EVM chain ID lookup, and transaction-ready envelope parsing.
  * Added server availability indicators to FastVault exports.

* **Bug Fixes**
  * Standardized Cardano TTL calculation during signing.

* **Tests**
  * Added coverage for TTL calculations, safety margins, slot offsets, and bigint results.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

PR #2018, opened 2026-08-17
fix(cli): admit enabled-chain scoping on aggregate balance/portfolio

neavra:
## Problem

The CLI has two inconsistent chain-scoping rules, and the inconsistency hides funds:

- All-chains `balance` / `portfolio` iterate only the vault's **enabled** chains — funds sitting on a non-enabled chain silently never appear, which reads as "the money is gone".
- Scoped `balance <chain>` accepts any **SDK-supported** chain.

The scoped path is the intended read-only discovery escape hatch — nothing signs or moves — so it stays as-is. The real defect is the aggregate view under-reporting without admitting it.

## Fix

When the enabled set is a strict subset of SDK-supported chains, the aggregate views append one static hint:

```
Showing the vault's 2 enabled chains of 39 supported. Funds on other chains won't appear here — run `vultisig balance <chain>` to check one, or `vultisig chains --add <chain>` to include it.
```

- **Human mode**: printed after the table (suppressed in silent mode).
- **JSON mode**: dedicated `scopeHint` field; the `balances` / `portfolio` / `failures` payloads are byte-for-byte unchanged. Field omitted when every supported chain is enabled.
- **No extra RPC**: the hint is a pure function of the enabled-chain count — it never scans non-enabled chains.
- Scoped `balance <chain>`, `chain-resolver.ts`, and `VaultBase` untouched.

## Tests

New `aggregate scope hint (DF-02)` suite: hint present in aggregate JSON (own field) and human output for both commands, omitted when all chains enabled, absent on the scoped path, plus a stale-hint guard asserting the referenced commands (`balance [chain]`, `chains --add <chain>`) still exist in `index.ts` and that the wording never claims a scan it didn't run.

Validated locally: `yarn test:cli` (1207 passing), lint, typecheck, quality:complexity, format:check. Changeset: `@vultisig/cli` patch.

## Summary by CodeRabbit

* **New Features**
  * Added scope hints to aggregate balance and portfolio views when only some supported chains are enabled.
  * Scope hints are available in both human-readable and JSON output.
  * Added clear handling for fully enabled, partially enabled, and zero-enabled chain configurations.
  * Improved reporting for scoped queries and unsupported or duplicate chain entries.
  * JSON responses preserve the original request data while providing accurate coverage details.


gomesalexandre (commented):
remote conceptual review on current draft head `455894dbe28ac25b28dc868360c4e0311ef4ba9c`.

looks solid.

what i re-grounded on this head:
- `clients/cli/src/commands/balance.ts` only changes the aggregate `balance` / `portfolio` paths. scoped `balance <chain>` stays hint-free except for the already-documented `--tokens` forwarding path.
- the new `buildScopeHint()` denominator now comes from `SUPPORTED_CHAINS`, which is the same registry `chains --add` validates against. that closes the earlier drift where the hint could have pointed at a chain the CLI would reject.
- json mode keeps the actual `balances` / `portfolio` payloads intact and carries the honesty text in a sibling `scopeHint` field, so machine consumers do not lose structure.
- human mode prints the hint only when `vault.chains.length < SUPPORTED_CHAINS.length`, and the added tests pin both the positive case and the fully-enabled no-extra-line case.
- the command remains static-only. no extra rpc or background chain scan is introduced.

non-blocking nit only: `buildScopeHint(0)` is harmless but currently unpinned. if you want belt-and-suspenders coverage, a zero-enabled-chains test would document the wording explicitly.

no material correctness blocker from this remote pass. keeping this as a comment because the PR is still draft.

gomesalexandre:
pushed `d72b0d78f` to keep the new scope hint honest on stale/duplicate enabled-chain lists.

what changed
- `buildScopeHint(...)` now derives its numerator from the unique intersection with `SUPPORTED_CHAINS` instead of the raw `vault.chains.length`
- both aggregate call sites (`balance`, `portfolio`) now pass the actual chain list
- added a regression pin for duplicate + unsupported persisted entries so the hint does not silently suppress itself while supported chains are still missing

receipts
- exact-head focused cli test: `./node_modules/.bin/vitest run --config .config/vitest.cli.config.ts clients/cli/src/commands/balance.test.ts`
  - result: `26 passed`

note
- i did not approve this head. it now contains a substantive gomes commit and needs a non-gomes current-head review.

gomesalexandre (commented):
current head `4bd57f712a099eb6af34e30562316df52af6409d` still looks sound from a re-read.

what i re-grounded on this head:
- the only post-`d72b0d78f` delta is the formatting follow-up in `clients/cli/src/commands/balance.test.ts`; no behavior change landed after the earlier correctness fix.
- `buildScopeHint()` still counts the unique intersection with `SUPPORTED_CHAINS`, so duplicate or stale persisted entries cannot suppress the honesty hint.
- aggregate `balance` / `portfolio` still carry the hint in a sibling `scopeHint` field while leaving the machine payloads unchanged, and scoped `balance <chain>` stays hint-free.
- the added tests still pin zero-enabled wording, duplicate/unsupported handling, and the no-extra-output path when every supported chain is enabled.

no new blocker from this current head.

not approving: this branch contains my substantive commit `d72b0d78f0382531bac46eaa663c476f3bea04a8`, so it still needs a non-gomes current-head approval before merge.

gomesalexandre (commented):
No correctness blocker on current head `128c413a9`.

- the aggregate balance/portfolio hint is still a pure static hint with no extra RPC or background scan
- the denominator is still deduped against `SUPPORTED_CHAINS`, so stale/duplicate enabled entries cannot suppress the honesty hint
- JSON mode still keeps `balances` / `portfolio` / `failures` payloads intact and carries the honesty text only in sibling `scopeHint`
- scoped `balance <chain>` stays hint-free
- the test coverage still pins zero-enabled wording, dedup/unsupported handling, the stale-hint guard, and the no-extra-line path when every supported chain is enabled

The two commits since the earlier re-review are merge-from-main only, so there is no fresh behavioral delta beyond the already-reviewed diff.

Four-eyes note: this branch still contains gomes-authored substantive commit `d72b0d78f0382531bac46eaa663c476f3bea04a8`, so it needs a non-gomes current-head approval before merge. Also re-check CI on this exact head before merge.

gomesalexandre (commented):
Looks good on review. The aggregate-only scope hint is honest about enabled-chain coverage, stays out of the scoped `balance <chain>` path, and the tests pin both the human output and JSON-envelope behavior without mutating the machine payloads.

Not stamping because this branch already contains one of my commits. It still needs a non-gomes approval for four-eyes.

rcoderdev (approved):

PR #2020, opened 2026-08-17
chore(rujira): delete dead fetchMergeableTokenBalances/fetchStakeView graphql clients

### LINKED ISSUE #1491: chore(rujira): delete/inert the dead merge-balance service
Found in the Station x Rujira vertical audit (2026-07-21).

## what
`fetchMergeableTokenBalances` is dead code. It queries the merge GraphQL field, but the KUJI to RUJI merge window closed 2026-04-05 (3+ months ago) and the service has ZERO callers.

## where
- `packages/core/chain/chains/thorchain/ruji/services/fetchMergeableTokenBalances.ts` - defines `fetchMergeableTokenBalances` (line 28); a repo-wide search finds no callers.

## why
The merge window is closed, so the query can only ever return empty, and nothing calls it anyway. It is pure dead code that implies a capability the app no longer has.

## fix
Delete `fetchMergeableTokenBalances.ts`, or comment it as inert with the merge-closed date if it must stay for reference.

### LINKED ISSUE #1492: chore(rujira): remove 2 drifted zero-caller graphql clients + fix stale cli doc
Found in the Station x Rujira vertical audit (2026-07-21).

## what
Two drifted, zero-caller Rujira GraphQL clients plus a stale CLI doc:
1. `fetchStakeView` - zero callers, and it even lacks the APR field the live staking path returns.
2. `fetchMergeableTokenBalances` - zero callers, merge window closed 2026-04-05 (tracked separately, see the delete-dead-merge-service issue).
3. `skills/vultisig-cli/references/rujira.md` - stale example memos that do not match the real `secure+:` / `secure-:` builders.

## where
- `packages/core/chain/chains/thorchain/ruji/services/fetchStakeView.ts` - zero callers repo-wide; missing APR vs the live staking read.
- `packages/core/chain/chains/thorchain/ruji/services/fetchMergeableTokenBalances.ts` - zero callers (see companion issue).
- `skills/vultisig-cli/references/rujira.md` - example memos drifted from the real secure-asset builders.

## why
Dead, drifted clients rot and mislead the next contributor (a zero-caller `fetchStakeView` that lacks APR looks like the staking read path but is not). The stale CLI doc shows memo formats that no longer match the real `secure+:` / `secure-:` builders, so anyone copying them builds a wrong memo.

## fix
Per sub-item: remove `fetchStakeView` and `fetchMergeableTokenBalances` (zero callers), and update `skills/vultisig-cli/references/rujira.md` to the real `secure+:` / `secure-:` builder memo formats.

gomesalexandre:
Closes vultisig/vultisig-sdk#1491
Closes vultisig/vultisig-sdk#1492

## what
Two zero-caller Rujira GraphQL clients, found in the 2026-07-21 Station x Rujira vertical audit:

- `fetchMergeableTokenBalances` (packages/core/chain/chains/thorchain/ruji/services/fetchMergeableTokenBalances.ts) queried the KUJI to RUJI merge field. That merge window closed 2026-04-05 (4+ months ago) and the query can only ever return empty.
- `fetchStakeView` (packages/core/chain/chains/thorchain/ruji/services/fetchStakeView.ts, exported as `fetchRujiraStakeView`) queried Rujira's `stakingV2` field but was never wired into any live staking read path, and its result shape doesn't even carry the APR field the real staking read returns.

Repo-wide grep confirms both had zero callers beyond their own definitions:

```
$ grep -rn "fetchMergeableTokenBalances\|fetchStakeView\|fetchRujiraStakeView" --include="*.ts" .
(only the two deleted files' own definitions)
```

Also removed `rujiraGraphQlEndpoint` and the `RujiraStakeView` type from `config.ts` since both become orphaned once the two clients are gone. `rujiraStakingConfig` in the same file was *already* unused before this change (not part of either issue) - left untouched, out of scope.

## #1492's third item (stale CLI doc)
#1492 also flagged `skills/vultisig-cli/references/rujira.md` as showing memo examples that don't match the real `secure+:` / `secure-:` builders. I checked this against current source and it's **not reproducible**: `packages/rujira/src/utils/memo.ts:150-168` builds exactly `secure+:${destination}` / `secure-:${destination}`, which is what the doc's examples show. No doc change needed here - this part of the issue is stale, closing with this citation rather than leaving it open.

## verification
Pure dead-code deletion with zero callers, so there's no behavior to mutation-test - the receipt here is that the deletion doesn't break anything:

```
$ yarn typecheck
# only 2 pre-existing, unrelated errors in cardano.ts / compileTx.golden.test.ts (auxiliaryData type
# mismatch) - present on main before this change, untouched by this diff

$ node scripts/run-eslint.mjs --config .config/eslint.config.mjs 'packages/core/chain/chains/cosmos/thor/rujira/**/*.ts'
# clean, no output

$ yarn format:check
Checking formatting...

$ yarn knip
# no new orphan-export findings for the touched file/dir
```

Changeset included (`@vultisig/core-chain` + `@vultisig/sdk` patch).

🤖 Agent-generated


gomesalexandre:
Fixed - the shared-exports generator (packages/core/chain's exports map is walked from dist/**/*.js) still had entries for the two deleted files. Removed both entries in 1b717c3da; this mirrors exactly what a fresh `yarn build:shared` would produce now that the source (and dist) for both is gone. Re-running CI.

gomesalexandre:
Reverted (`6ef24e2ca`) - closing without merging. This PR's premise was wrong, and I want to flag exactly how so the mistake is on record.

## What I got wrong
My "zero callers" verification (repo-wide `grep`) only checked callers **inside** `vultisig-sdk`. It missed that these are published `@vultisig/core-chain` subpath exports consumed by an **external** repo (`vultisig-windows`). `scripts/windows-core-chain-exports.test.mjs` - added five days before this PR by #1873 ("fix: preserve Windows Rujira exports") - is first-party, explicit evidence of exactly this: the PR body for #1873 states outright "The Rujira GraphQL clients and public core-chain subpaths remain intact because the current Windows app consumes both; packed-package contracts now guard those exports." I deleted both files and the guard's `quality:contracts` check correctly failed CI on this PR - I should have investigated *why* it failed instead of just patching the generated-exports symptom the first time around.

## Where this actually landed correctly
#1871 (merged just before I opened this PR) already fixed `fetchMergeableTokenBalances` the right way: an inert stub that returns `[]` without querying GraphQL, keeping the Windows-consumed export alive rather than deleting it - exactly the "or comment it as inert... if it must stay for reference" alternative #1491 itself suggested. I'll close #1491 citing that.

`fetchStakeView` has no equivalent treatment - it's still a real, live GraphQL query, and the Windows guard test protects it the same way. #1492's "delete it" premise doesn't hold for the same reason; I'm posting a correction there rather than a fix.

Sorry for the noise - the branch is reverted to match current `main` exactly, verified via `git diff --cached origin/main` (clean, zero delta) before pushing.

PR #2021, opened 2026-08-17
feat(sdk): add grouped sdk.price namespace handle at the root entry

### LINKED ISSUE #1780: public API: root @vultisig/sdk still omits a grouped sdk.price namespace handle
## Summary
The SDK ships a coherent price-helper family but still does not expose a grouped `sdk.price.*` namespace from the root surface.

## Evidence
- `packages/sdk/src/tools/price/index.ts` already exists as a standalone price barrel.
- `packages/sdk/src/tools/index.ts` exports grouped handles like `gas`, but only flat price helpers (`getPrice`, `getPricesBatch`, etc.) and no `price` namespace object.
- `packages/sdk/src/index.ts` mirrors that omission at the root entry.

## Impact
First-party consumers that want ergonomics consistent with `sdk.gas.*` / `sdk.defi.*` still have to keep local wrapper objects around price helpers.

## Effort
Low.

## Fix sketch
Add `import * as price from './price'` + `export { price }` alongside the existing grouped handles, while keeping the current flat exports for back-compat.

## Report
https://gist.github.com/gomesalexandre/04cec78e16a137403ef135ea0631d431

gomesalexandre:
Closes #1780

## what

The SDK already ships a coherent price-helper family (`getPrice`, `getPricesBatch`, `coinGeckoIdToSymbol`, `isKnownNativePriceSymbol`, `NATIVE_COINGECKO_IDS`, `symbolFromCoinGeckoId`) but only exposed it as flat named exports, unlike the sibling `gas`/`defi`/`token`/`cosmos` families which also get a grouped namespace object (`sdk.gas.*`, `sdk.defi.*`, `sdk.token.*`, `sdk.cosmos.*`).

## why

Consumers wanting that ergonomic for price had to keep a local wrapper object around the flat helpers, exactly the kind of first-party-consumer friction this export campaign targets.

## fix

Added `import * as price from './price'` + `export { price }` in `tools/index.ts` (same pattern as the existing `gas` namespace handle a few lines above), and re-exported `price` from the root barrel alongside the other namespace handles. Flat exports are untouched — purely additive.

## mutation-test proof

Stashed both `src/` changes (kept the extended test), reran:

```
FAIL toolNamespaceRootExports.test.ts > exposes the price helper family under sdk.price without removing flat exports (sdk#1780)
  TypeError: Cannot read properties of undefined (reading 'getPrice')
Tests  1 failed | 3 passed (4)
```

Restored, all green.

## verification

```
corepack yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/public/toolNamespaceRootExports.test.ts
# Test Files  1 passed (1) / Tests  4 passed (4)

corepack yarn vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/
# Test Files  220 passed (220) / Tests  3419 passed (3419) — no regressions

corepack yarn workspace @vultisig/sdk build:fast
# builds clean

corepack yarn lint:fix packages/sdk/src/index.ts packages/sdk/src/tools/index.ts packages/sdk/tests/unit/public/toolNamespaceRootExports.test.ts
corepack yarn format <same files>
# both clean, no changes needed

# no errors from these files
```

## codex review

`codex exec` still hit an expired OAuth token in this sandbox (same as the other PRs opened this session). Proceeding without it — self-reviewed against the identical existing `gas`/`token`/`cosmos` namespace-handle pattern in the same file, plus the mutation-test proof above.

🤖 Agent-generated


NeOMakinG (approved):
LGTM

rcoderdev (commented):
The new root `price` namespace is not exposed by the React Native root entry. The package root selects `dist/index.react-native.js` under the `react-native` condition, and `platforms/react-native/index.ts` only groups `cosmos`, `evm`, and `token`; it flat-exports price helpers but never imports or exports `price`. As a result, `import * as sdk from "@vultisig/sdk"; sdk.price` is undefined on React Native even though it exists on Node. Please add the namespace to the React Native entry and cover React Native root-entry identity.

rcoderdev:
Closing this stale PR after exact-head review. The package root sends React Native consumers to dist/index.react-native.js, but the React Native entrypoint still groups only cosmos, evm, and token and does not import/export the new price namespace. As a result sdk.price would remain undefined for React Native consumers. A replacement should add the namespace to the RN entrypoint and cover that export surface.

PR #2029, opened 2026-08-17
fix(sdk): route cctp chain lookups through the canonical normalizer

### LINKED ISSUE #1532: bridge: CCTP builders bypass canonical SDK chain normalization
## Summary
CCTP bridge/claim builders bypassed the SDK's canonical chain normalizer and exact-matched raw caller strings.

## Evidence
- `packages/sdk/src/tools/bridge/cctp.ts:56-115`
- `packages/sdk/src/tools/bridge/buildCctpBridge.ts:213-232`
- `packages/sdk/src/tools/bridge/buildCctpClaim.ts:100-107`
- canonical resolver already exists at `packages/sdk/src/utils/normalizeChain.ts:138-155`

## Impact
First-party consumers had to pre-normalize chain names locally, and same-chain alias pairs were not rejected consistently.

## Fix sketch
Resolve CCTP chains through `normalizeChain()`, return canonical chain names from the builders, and lock the behavior with regression tests.

## Effort
Low.

## Round report
https://gist.github.com/gomesalexandre/6df6ab2623a396546e6c5b4140e979b6

### LINKED ISSUE #1911: architecture(cctp): public builders still bypass the SDK chain canonicalizer
## Summary

The public CCTP builders still require exact canonical chain casing (`"Base"`, `"Arbitrum"`, etc.) instead of routing inputs through the SDK’s shared chain canonicalization surface.

That keeps consumers normalizing CCTP chain names themselves even though the SDK already exports canonical chain resolvers elsewhere.

## Evidence

- Root SDK already exports shared chain canonicalization helpers:
  - `packages/sdk/src/index.ts:58`
  - `packages/sdk/src/index.ts:83-90`
- But the CCTP registry lookup is still a raw object index:
  - `packages/sdk/src/tools/bridge/cctp.ts:56-111`
- `buildCctpBridge` just trims the input and does exact lookup:
  - `packages/sdk/src/tools/bridge/buildCctpBridge.ts:214-231`
- `buildCctpClaim` does the same for its destination chain:
  - `packages/sdk/src/tools/bridge/buildCctpClaim.ts:100-105`

So the newer public helper family still needs caller-side pre-normalization for inputs like `"base"`, `"ethereum"`, or alias-style chain strings, even though the SDK already has shared canonical chain logic.

## Impact

This is another duplicated-not-imported seam:
- CLI/backend/app must normalize chain names before calling CCTP helpers
- the helpers themselves bypass the shared SDK chain surface
- future alias additions can land in the SDK canonicals while CCTP remains stale and rejects inputs the rest of the SDK accepts

## Target shape

Route CCTP chain names through the shared chain canonicalization path first, then validate that the resolved chain is in the CCTP-supported set.

For example:
- normalize / resolve the incoming string via the canonical SDK chain helper
- map that resolved chain to the CCTP registry key
- reject only after canonicalization if the chain is not one of the supported CCTP domains

## Effort

Small, low-risk consistency fix.

gomesalexandre:
Closes vultisig/vultisig-sdk#1911
Closes vultisig/vultisig-sdk#1532

`getCctpChain` was a raw object index:

```ts
export const getCctpChain = (chainName: string): CctpChainConfig | undefined => cctpChains[chainName]
```

The registry keys are canonical-cased (`Ethereum`, `Base`, `Arbitrum`, …), so `getCctpChain('base')` is `undefined` while every other SDK helper resolves that string fine through `normalizeChain`. Callers had to pre-normalize for this one family, and an alias added to the canonicals later would work everywhere except here.

## two things that fell out of it

Neither is in the issue, and both are worse than the bug if you add alias tolerance without noticing them.

**1. The same-chain guard compared raw strings, before any lookup.**

```ts
if (srcChainName === dstChainName) throw new Error('sourceChain and destinationChain must be different')
```

Accept aliases and `('base', 'Base')` sails through that check as two different chains, then resolves both to Base and builds a **bridge from a chain to itself**. The guard now runs on resolved canonical names, and there is a test for exactly this pairing.

**2. Both builders stamped the caller's raw input into the emitted envelope** — `chain`, `fromChain`, `toChain`. A caller writing `'base'` would have received an envelope saying `'base'`, which downstream consumers do not expect. They now use `CctpChainConfig.chain`, which is already the canonical `EvmChain` value, so no new resolver was needed for the name.

## contract preserved

`getCctpChain` still returns `undefined` rather than throwing, even though `normalizeChain` throws `UnknownChainError`. A chain that normalizes fine but simply has no CCTP V1 config — Solana — has to answer identically to a string that does not normalize at all. Both are asserted.

The unsupported-chain error still quotes the caller's **own** spelling, so the message stays useful instead of echoing a canonical name they never typed.

## receipts

**before** (source reverted, tests unchanged):

```console
$ vitest run tests/unit/bridge-cctp.test.ts
 × resolves "base" to the canonical Base config
     AssertionError: expected undefined to be { chain: 'Base', …(5) }
 × resolves "BASE" to the canonical Base config
 × resolves " base " to the canonical Base config
 × builds the same envelope from a lowercase spelling as from the canonical one
 × still rejects a same-chain bridge written with two different spellings
     AssertionError: expected [Function] to throw error matching /must be different/
                     but got 'source chain "base" is not supported …'

   Tests  5 failed | 19 passed (24)
```

That last failure is the interesting one. Pre-fix, `('base', 'Base')` is rejected — but for the *wrong reason*, as an unsupported chain. Post-fix it is rejected as the same chain, which is what it actually is.

**after**: `Tests 24 passed (24)`.

The negative controls pass in **both** states by design — `getCctpChain('Solana')`, `getCctpChain('definitely-not-a-chain')` and `getCctpChain('')` all stay `undefined`. Those are the guard against "make it tolerant" turning into "make it accept anything".

The envelope-equality case is the strongest single assertion:

```ts
expect(buildCctpBridge({ sourceChain: 'base', destinationChain: 'arbitrum', ... }))
  .toEqual(buildCctpBridge({ sourceChain: 'Base', destinationChain: 'Arbitrum', ... }))
expect(lowercased.chain).toBe('Base')     // canonical, not what the caller typed
```

## gate

| check | result |
|---|---|
| `yarn workspace @vultisig/sdk test:unit` | 3426 passed |
| `bridge-cctp.test.ts` | 24 passed |
| `yarn typecheck` | exit 0 |
| `yarn lint` | exit 0 |

No network receipt: this is pure lookup and envelope-construction logic with no I/O, and the envelope-equality assertion compares a full built result rather than a field.

## risk

Low, and it only widens what is accepted. Every input that resolved before resolves to the identical config, asserted by the `'Base'` cases that pass in both states. The one behaviour change beyond acceptance is the error you get for a same-chain bridge written two ways, which moves from "not supported" to "must be different" — a more accurate message for a case that was already rejected.


gomesalexandre (commented):
remote/local review on current head `06a49d1e20fbb7ccb4ff1cec5dd4395cc1ce22f5` looks sound.

what i re-grounded:
- `getCctpChain` now preserves the old undefined-for-unsupported contract while accepting the spellings the rest of the sdk already accepts. the direct trimmed lookup keeps the exact canonical fast path, and the `normalizeChain(...)` fallback is caught so genuinely unknown input still fails closed as `undefined` instead of throwing.
- the same-chain guard moved to canonical names, which is the important trap here. pre-fix, `('base', 'Base')` could sneak through as two different raw strings and build a same-chain bridge. this head now rejects that on the resolved cctp configs.
- emitted envelopes are stamped from `srcCctp.chain` / `dstCctp.chain`, so callers can hand in `base` / `BASE` / padded input and still get canonical `Base` / `Arbitrum` output instead of caller-spelling drift.
- unsupported-but-real chains still fail closed. `Solana` now normalizes successfully but still returns `undefined` because there is no CCTP config, which matches the old function contract rather than creating a new throw path.

receipts:
- `claude -p` review verdict on the exact-head temp clone: `VERDICT: APPROVE`
- focused exact-head unit receipt: `./node_modules/.bin/vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/bridge-cctp.test.ts` -> `24 passed (24)`
- i also tried the heavier sdk receipt path (`yarn build:all && yarn cli:build`) in the temp clone, but it is blocked by a pre-existing main-side type error outside this diff:
  `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts(72,5): error TS2353: Object literal may only specify known properties, and 'auxiliaryData' does not exist in type 'ISigningInput'.`
  that reproduces byte-identically in a fresh `origin/main` worktree, so i am not treating it as a regression from this pr.

no blocker found in the changed files.

four-eyes note: this head contains gomes-authored commits, so i cannot approve it. it needs a non-gomes current-head approval before merge.

gomesalexandre:
A parallel worker independently found and fixed the exact same defect under a separate issue number (#1532 — same file, same function, same fix: route `getCctpChain` through the canonical chain normalizer). Folding it in here rather than opening a duplicate PR; its branch `fix/w1-1532` is left unopened.

NeOMakinG (approved):
LGTM

gomesalexandre:
pushed a mergefix onto the current head `51675008c630775853c5889ff8821be626edae66` to clear the `DIRTY` state on top of fresh `main`.

what changed:
- merged `origin/main` into `fix/sdk-1911-cctp-chain-canonicalizer`
- resolved the single textual conflict in `packages/sdk/src/tools/bridge/buildCctpClaim.ts`
- kept main's newer burn-binding checks and the PR's canonicalizing `getCctpChain(...)` behavior

receipts:
- `./node_modules/.bin/vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/bridge-cctp.test.ts`
- result: `1 passed`, `32 passed (32)`
- verified no remaining conflict markers before commit
- created a real two-parent merge commit: `51675008c630775853c5889ff8821be626edae66`

four-eyes note: this push changed the head sha, so the old approval is stale. this needs a fresh non-gomes current-head review before merge.

rcoderdev (approved):

rcoderdev (approved):
Re-approved on the current head after the branch update. Canonical CCTP normalization remains fail-closed for unsupported input, same-chain comparison uses canonical values, and the main-side burn-identity guards remain intact. CI remains the merge gate.

rcoderdev (approved):

rcoderdev (approved):
Exact-head integration review passed for 8950dc27bcf8e6a2227b9b6bd3bd1e96db78d916; the approved feature is preserved on current main with no must-fix regression.

PR #2034, opened 2026-08-18
fix(across): type sourcechain as the origin acrossquote actually accepts

### LINKED ISSUE #1906: fix(across): public acrossQuote contract advertises multi-origin support but runtime only allows Ethereum

gomesalexandre:
Closes vultisig/vultisig-sdk#1906

`sourceChain` was typed as any `AcrossChain` — all eight supported chains — while the runtime accepts exactly one:

```ts
sourceChain: AcrossChain            // Ethereum | Optimism | Polygon | Arbitrum | Base | Blast | BSC | Zksync
...
if (sourceChain !== ETHEREUM_ORIGIN_CHAIN) {
  throw new Error(`sourceChain must be ${ETHEREUM_ORIGIN_CHAIN} for this Across factory slice`)
}
```

And the JSDoc `@example` — the thing a consumer copies — passed `'Base'`. So the documented call **compiled cleanly and then threw**.

## how

Took the issue's first option: narrow the contract, keep the fail-closed runtime.

- `sourceChain` is now `AcrossOriginChain`, a type that says exactly what the runtime accepts.
- `ACROSS_ORIGIN_CHAIN` is exported, so a caller can branch on the limit rather than hardcode `'Ethereum'` or discover it by catching.
- The `@example` now uses `'Ethereum'` with Ethereum USDC, so the documented call both compiles and runs.
- Exported from the swap, tools and root barrels alongside the existing `AcrossChain`.

When genuine multi-origin support lands, widening that one type is the change — and every caller already passing `'Ethereum'` keeps compiling.

## the runtime guard stays

Narrowing a type protects TypeScript callers and nobody else. A JavaScript caller, or a typed one passing a value widened through `any`, still reaches the guard, so it is unchanged and its test is kept — now with a deliberate cast and a comment explaining that the cast **is** the test:

```ts
// The cast deliberately simulates a JavaScript caller ... Narrowing the type
// does not remove the need for [the runtime guard].
sourceChain: 'Base' as unknown as AcrossOriginChain,
```

Without that comment someone tidies the cast away later and silently deletes the coverage.

## receipts

This is a type-level defect, so the receipt is a **compile**, not an assertion. I put the JSDoc example verbatim into a probe file and ran the project's own typecheck on both sides:

```console
# BEFORE (acrossQuote.ts reverted) — the broken example compiles fine
$ yarn typecheck
   0 errors mentioning the probe

# AFTER
$ yarn typecheck
__probe-across.test.ts:8:9 - error TS2322: Type '"Base"' is not assignable to type '"Ethereum"'.
```

One caveat, stated rather than hidden: the BEFORE run exits non-zero overall, because the reverted `acrossQuote.ts` no longer exports `AcrossOriginChain` that the updated test file imports. The signal is the **zero probe-related errors** in that run — the documented example itself was accepted. The probe file was temporary and is not in this diff.

**Unit:**

```console
$ vitest run tests/unit/tools/swap/acrossQuote.test.ts
   Tests  15 passed (15)
```

15, not 14 — the rejection test is kept (retitled to say it covers untyped callers) and a new case asserts the positive direction, that `ACROSS_ORIGIN_CHAIN` is a member of `acrossSupportedChains`. Otherwise a future edit could narrow the type to something the runtime rejects and nothing would notice.

## gate

| check | result |
|---|---|
| `yarn workspace @vultisig/sdk test:unit` | 3419 passed |
| `acrossQuote.test.ts` | 15 passed |
| `yarn typecheck` | exit 0 |
| `yarn lint` | exit 0 |

## risk

Type-only for anyone already using the helper correctly — every call that worked at runtime still compiles, because `'Ethereum'` was the only value that worked. It is a **compile-time breaking change** for code that passes a non-Ethereum origin, but that code was already throwing at runtime; this moves the failure from production to the build, which is the point. No runtime behaviour changes.

## Summary by CodeRabbit

* **New Features**
  * Added public SDK exports for vault import options, swap quotes, CCTP message utilities, chain names, and raw transaction broadcasting.
  * Added a constant and type identifying Ethereum as the supported Across quote origin chain.
  * Improved quote parameter type guidance for source chains.

* **Bug Fixes**
  * Preserved runtime rejection of unsupported non-Ethereum origin chains.

* **Tests**
  * Added coverage for supported and rejected Across origin chains.


gomesalexandre:
merged `origin/main` into this branch and pushed `104c2da76` to clear the conflict.

what changed
- resolved the single merge conflict in `packages/sdk/src/tools/swap/index.ts`
- kept both sides of the additive export surface: the pr's across origin chain exports (`AcrossOriginChain`, `ACROSS_ORIGIN_CHAIN`) and main's newer `findSwapQuotes` exports/types (`BoundSwapQuote`, `FindSwapQuotesResult`, `SwapQuoteCandidate`, `findSwapQuotes`)

receipts
- exact-head merge conflict markers check: clean
- `yarn workspace @vultisig/sdk build:fast`: passed
- `yarn workspace @vultisig/sdk typecheck`: still red on pre-existing unrelated `auxiliaryData` errors in `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts` and `packages/core/mpc/tx/compile/compileTx.golden.test.ts`

four-eyes note
- this head contains gomes-authored commits, so it still needs a non-gomes current-head review before merge.

NeOMakinG (approved):
LGTM

PR #2037, opened 2026-08-18
fix(sdk): one formatusdc instead of two independent copies

### LINKED ISSUE #1931: architecture(usdc): shared formatter/parser canonical is still split across CCTP and ThreeJane
## Summary
The SDK still split the USDC amount canonical between a shared parser and duplicated per-feature formatters, leaving CCTP and ThreeJane on separate local formatting code paths.

## Evidence
- shared parser: \
- duplicate local formatters before fix:
  - \
  - \

## Why this matters
This is low-risk but real internal architecture drift inside the SDK itself. First-party CCTP and ThreeJane surfaces should not maintain their own bigint↔human USDC formatting contracts.

## Target shape
One shared USDC parser/formatter helper consumed by both features.

## Report
https://gist.github.com/gomesalexandre/cd383b2c4e1978786d164e3a7370ede5

gomesalexandre:
Closes vultisig/vultisig-sdk#1931

`parseUsdcAmount` and `USDC_DECIMALS` already live in `tools/parse/usdcAmount.ts`. Its inverse, `formatUsdc`, was written twice:

```ts
// tools/bridge/buildCctpBridge.ts
export const formatUsdc = (raw: bigint): string => { ... }

// tools/defi/threeJane/buildSupplyUsdc.ts
function formatUsdc(raw: bigint): string { ... }
```

## why this one is worth fixing rather than noting

The two copies still agree — I diffed them, same algorithm — so nothing is broken today.

What makes it more than tidiness is *what this function produces*. `formatUsdc` renders the `amountUsdc` string a user **reads on a signing card**:

```ts
description: `Burn ${formatUsdc(rawAmount)} USDC via depositForBurn (step 2 of 2)...`
amountUsdc: formatUsdc(rawAmount),
```

A drift between the copies shows the same amount two different ways depending on which builder produced it. A display-only divergence on a fund path is the kind that gets believed, and neither copy would fail a test that only checks its own output.

## how

- `formatUsdc` moves to `tools/parse/usdcAmount.ts`, next to its inverse.
- `tools/bridge` re-exports it from there, so **no consumer import path changes**.
- `tools/defi/threeJane` now exports it too, matching how `parseUsdcAmount` is already surfaced from that barrel — previously it was a file-private `function`, which is why the copy existed.

## receipts

The test mirrors the one already in this file for `parseUsdcAmount`, and the shape is the point:

```ts
it('uses one implementation for CCTP and ThreeJane', () => {
  expect(formatCctpUsdc).toBe(formatUsdc)
  expect(formatThreeJaneUsdc).toBe(formatUsdc)
  expect(formatCctpUsdc).toBe(formatThreeJaneUsdc)
})
```

**Identity, not behaviour.** Asserting both render `'1.5'` would pass with two copies still sitting there — asserting they are the *same function* is what makes a future local copy fail.

**before** (source reverted, tests unchanged):

```console
$ vitest run tests/unit/usdc-amount.test.ts
 × uses one implementation for CCTP and ThreeJane
     AssertionError: expected [Function formatUsdc] to be undefined
 × is the exact inverse of parseUsdcAmount
 × renders whole amounts without a decimal point and trims trailing zeros

   Tests  3 failed | 6 passed (9)
```

**after**: `Tests 9 passed (9)`.

The other two cases pin behaviour so the move itself is provably faithful — a full round-trip against `parseUsdcAmount` (`'1'`, `'0.1'`, `'0.000001'`, `'1000'`, `'1234.567891'`, `'0.5'`) and the exact strings for the edges both copies handled (`1_000_000n → '1'`, `1n → '0.000001'`, `1_500_000n → '1.5'`, `0n → '0'`).

## gate

| check | result |
|---|---|
| `yarn workspace @vultisig/sdk test:unit` | 3421 passed |
| `usdc-amount.test.ts` | 9 passed |
| `yarn typecheck` | exit 0 |
| `yarn lint` | exit 0 |

## risk

Very low. Same implementation, one home. No consumer import path changes — `tools/bridge` still exports `formatUsdc` from where it always did. The only surface addition is `formatUsdc` on the threeJane barrel, which is additive.

## Summary by CodeRabbit

* **New Features**
  * Added consistent formatting for USDC amounts, converting raw values into readable strings.
  * Exposed the formatter through supported bridge and DeFi tools.

* **Bug Fixes**
  * Improved handling of whole USDC amounts and trailing fractional zeros.
  * Ensured formatted values consistently correspond to parsed amounts.

* **Tests**
  * Added coverage for representative USDC values and formatting behavior.


NeOMakinG (approved):
LGTM

PR #2038, opened 2026-08-18
feat(sdk): publish canonical getBlockchairBaseUrl helper

### LINKED ISSUE #1761: architecture(utxo): Blockchair URL contract is still split across sdk/app/abts
### Summary
The SDK already depends on a chain-scoped Blockchair API contract, but first-party consumers still reconstruct `${root}/blockchair/${chain}` locally.

### Evidence
SDK:
- `packages/sdk/src/chains/utxo/rpc.ts`
- `packages/sdk/src/tools/balance/utxoBalance.ts`

Consumers:
- app `src/services/utxoTx.ts:79-85`
- app `src/features/agent/lib/resolvers/utxoResolver.ts`
- abts `src/mastra/tools/mcp/lib/rpc.ts:232-233`

### Impact
Any Blockchair proxy-path or slug change requires coordinated edits in three repos.

### Fix sketch
Expose a canonical helper from the SDK (`blockchairUrlForChain()` / `blockchairApiForChain()`), cover slug/path behavior with unit tests, and migrate first-party consumers to import it.

### Artifact
https://gist.github.com/gomesalexandre/ac0eaebb8b255a878ed9ab9c4a326b57

gomesalexandre:
Closes vultisig/vultisig-sdk#1761

SDK core already had a canonical Blockchair chain-scoped base-URL builder (`getBlockchairBaseUrl` in `packages/core/chain/chains/utxo/client/getBlockchairBaseUrl.ts`, `${rootApiUrl}/blockchair/${chain.toLowerCase()}`), used internally by the UTXO tx-status and broadcast resolvers, but it was never re-exported from the SDK's public `index.ts`. That forced first-party consumers (app, agent-backend) to reconstruct the same path locally, so a Blockchair proxy-path or slug change requires coordinated edits across repos instead of one shared contract. Fixed the SDK-side gap by re-exporting `getBlockchairBaseUrl` from `@vultisig/sdk` (mirrors the existing `cosmosRpcUrl`-style direct core-chain re-export pattern already in `index.ts`), and added a focused unit test covering the plain, hyphenated (`Bitcoin-Cash`), and non-`UtxoChain` (`Cardano`) slug cases. NOTE: the issue's fix sketch also asks to migrate the app and agent-backend-ts call sites to import this helper — those are separate repos not checked out in this worktree, so that half is out of scope here; only the SDK-side export + test is included in this PR.

## what I ran
Could NOT execute vitest against `getBlockchairBaseUrl.test.ts` (or any test importing it) in this worktree: `packages/core/config` has no built `dist/` here, and its `package.json` "exports" field points only at `dist/index.js` with no fallback, so any bare `import ... from '@vultisig/core-config'` (which `getBlockchairBaseUrl.ts` itself does, for `rootApiUrl`) fails module resolution under both `.config/vitest.core.config.ts` and `packages/sdk/vitest.config.ts` — confirmed via `/tmp/claude-1000/wt-sdk-w1/node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/core/chain/chains/utxo/client/getBlockchairBaseUrl.test.ts` and cross-checked that this is a PRE-EXISTING, unrelated environment gap (not caused by my change) by running the same command against the already-existing `getUtxoAddressInfo.test.ts` in the same directory, which fails identically. The only fix would be running `yarn build:shared` (a monolithic tsc build of all 6 shared packages), which the task instructions explicitly forbid as a "full build." I instead verified correctness by manual trace: `getBlockchairBaseUrl.ts` is `${rootApiUrl} + '/blockchair/' + chain.toLowerCase()`; for `UtxoChain.Bitcoin` ('Bitcoin') → `.../blockchair/bitcoin`, `UtxoChain.BitcoinCash` ('Bitcoin-Cash') → `.../blockchair/bitcoin-cash`, `OtherChain.Cardano` ('Cardano') → `.../blockchair/cardano` — all match my test's assertions. The test will run correctly under CI, where the shared packages are built before tests execute. Did not run the full suite, typecheck, or build.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev (approved):
Re-approved on the current head after the branch update. The root export targets the existing core-chain subpath, preserves the UtxoBasedChain contract, and covers Bitcoin, Bitcoin Cash, and Cardano; no actionable issue found. CI remains the merge gate.

rcoderdev (approved):

rcoderdev (approved):
Exact-head conflict-resolution review passed for e50eda3d913efab87d6279e7dde0093abfddd903; both React Native parity contracts and all current-main StakeKit surfaces are preserved.

PR #2041, opened 2026-08-18
feat(vault): expose broadcastRawTx as a vault-free sdk export

### LINKED ISSUE #1768: architecture: raw broadcast stays vault-bound even though the SDK already has a chain-agnostic broadcaster
## Summary
The SDK already contains a mature multi-chain raw-broadcast implementation, but the only public way to reach it is through a vault instance. Non-wallet runtimes that already hold signed bytes still need vault plumbing or their own submitters, which keeps tx-execution logic duplicated outside the SDK.

## Evidence
- chain-agnostic broadcaster implementation: `packages/sdk/src/vault/services/RawBroadcastService.ts:99-122`
- chain branches below that path cover EVM / UTXO / Solana / Cosmos / TON / Polkadot / Bittensor / Ripple / Sui / TRON
- vault-only public access today: `packages/sdk/src/vault/VaultBase.ts:1185-1202`
- CLI broadcast path still requires an active vault: `clients/cli/src/commands/broadcast.ts:32-45`
- no root-public `broadcastRawTx` export from `packages/sdk/src/index.ts`

## Impact
- callers that already have signed raw bytes cannot reuse the SDK broadcaster without creating vault plumbing
- app/backend/other runtimes are nudged toward bespoke submitters instead of consuming one canonical execution boundary
- this is the campaign's recurring duplicated-not-imported bug class in a high-leverage location

## Fix sketch
Promote the existing `RawBroadcastService` into a public vault-free helper (for example `broadcastRawTx`) and keep vault methods as thin wrappers.

## Verification target
- public helper can submit through the shared broadcaster without requiring a vault object
- CLI/runtime receipts prove the helper surface works on real signed bytes

## Report
Round-107 report: https://gist.github.com/gomesalexandre/68d3313375c221c04a954468dd673dff

gomesalexandre:
Closes vultisig/vultisig-sdk#1768

`RawBroadcastService` (`packages/sdk/src/vault/services/RawBroadcastService.ts`) is already a chain-agnostic broadcaster covering EVM/UTXO/Solana/Cosmos/TON/Polkadot/Bittensor/Ripple/Sui/TRON, and holds no vault-instance state (no constructor; every method is a pure chain dispatch). But the only public way to reach it was through `VaultBase.broadcastRawTx` (`packages/sdk/src/vault/VaultBase.ts:1463`), forcing runtimes that already hold externally-signed raw bytes to stand up vault plumbing or their own submitter just to broadcast. Added a standalone `broadcastRawTx` export backed by a shared `RawBroadcastService` instance, and re-exported it from the SDK root `index.ts` next to the existing `GasEstimationService` vault-service export. `VaultBase.broadcastRawTx` is untouched — it stays a thin wrapper that additionally emits `transactionBroadcast`/`error` vault events.

## what I ran
`/tmp/claude-1000/wt-sdk-w2/node_modules/.bin/vitest run tests/unit/vault/services/RawBroadcastService.test.ts` (run from `packages/sdk`) — 1 test file, 51 tests, all passed (49 pre-existing + 2 new: a standalone `broadcastRawTx` EVM success-path test and an `UnsupportedChain` fail-closed test, both exercising the export directly with no `RawBroadcastService` instantiation). Also ran `tests/unit/utils/publicExports.test.ts` — 55 tests, all passed, confirming the new root export doesn't break module resolution.


NeOMakinG (approved):
LGTM

PR #2049, opened 2026-08-18
fix(balance): preserve exact trx sun balances beyond number.max_safe_integer

### LINKED ISSUE #1877: fix(balance): preserve exact TRX SUN balances beyond Number.MAX_SAFE_INTEGER

gomesalexandre:
Closes vultisig/vultisig-sdk#1877

`getTrxBalance` parsed TRON's `balance` field as a plain JS `number`. TRON's RPC returns `balance` as a JSON *number* literal, and `JSON.parse` already rounds any numeric literal past `Number.MAX_SAFE_INTEGER` (~9.007e15, ~9.007M TRX in SUN) before our code even sees it, so whale balances above that threshold were silently corrupted — deterministic correctness drift in an SDK-owned balance read, the same class of bug the SDK already guards against for XRP drops and Solana lamports. Fixed by adding a `fetchJsonWithText` helper (`rpc.ts`) that returns the raw response text alongside the parsed JSON, then recovering the exact SUN integer from that raw text via regex + `BigInt` before any `Number` coercion can happen — mirroring the existing `getSolBalance` pattern exactly. `TrxBalance` now exposes `balanceSun: number | null` (`null` when unsafe to represent as a `number`) and a new lossless `balanceSunRaw: string`; `balanceTrx` is always formatted from the exact `BigInt`.

## what I ran
`cd packages/sdk && ../../node_modules/.bin/vitest run tests/unit/tools/balance/otherBalance.test.ts` — 33 passed (33), 0 failed. This required a one-time environment fix first: the worktree had never run `worktree:setup`/`build:shared`, so all vitest runs failed on missing `dist/` for internal workspace packages (`@vultisig/core-config`, `@vultisig/mpc-types`, etc.) — unrelated to this change, but needed for any test in this worktree to execute at all. I ran `node scripts/setup-worktree.mjs --from ~/Sites/vultisig-sdk --skip-build` (symlinks node_modules overrides, ~1s) then `yarn build:shared` (compiles the shared workspace packages' dist output only, ~9s — NOT `build:sdk:bundle`, which I did not run). I added 2 new test cases (whale->2^53 SUN precision fixture, missing-balance-field zero case) and updated the existing "formats SUN into TRX" test + the mock setup to cover the new `fetchJsonWithText` helper (the old test only mocked `fetchJson`, which `getTrxBalance` no longer calls). Did not run `yarn typecheck` or `yarn build:sdk:bundle` — out of scope per instructions.


NeOMakinG (approved):
LGTM

rcoderdev (commented):
`TrxBalance.balanceSun` is a public field, and changing it from `number` to `number | null` breaks existing TypeScript callers, but this ships as a patch. The repository changeset policy classifies breaking API changes as major. Please either preserve the existing field contract while adding the new lossless representation, or use the appropriate breaking-release classification. Also, the threshold in the JSDoc/changeset is about 9.007 billion TRX (`MAX_SAFE_INTEGER / 1e6`), not 9.007 million.

rcoderdev:
Closing this because Thanks for the exact TRX balance work. The rounding bug is real, but this patch makes public balanceSun nullable and misstates the safety threshold by 1,000x. It also conflicts with current RPC code. Preserving the public contract, rebasing and validating the parser outweigh this branch's benefit, so I am closing it; #1877 remains open for a compatible fix.

PR #2050, opened 2026-08-18
fix(changesets): stop forcing spurious majors on peer-dependent packages

### LINKED ISSUE #1921: @vultisig/rujira majors on every sdk release: 61 of 64 version bumps were spurious majors
## tl;dr

`@vultisig/rujira` is at **64.0.0 with 65 published versions, 61 of which were major bumps**, and essentially none of them were breaking. Every `@vultisig/sdk` release forces a rujira major because rujira declares the SDK as a peer dependency and changesets' default is to major every peer-dependent. Semver on that package currently carries no information.

## the numbers

```
@vultisig/rujira published versions: 65
last 12: 53.0.0 54.0.0 55.0.0 56.0.0 57.0.0 58.0.0 59.0.0 60.0.0 61.0.0 62.0.0 63.0.0 64.0.0
bump kinds across the whole history: {major: 61, minor: 1, patch: 2}
```

Sixty-one majors, one minor, two patches.

## it is happening again in #1876

The current release PR bumps rujira `64.0.0 -> 65.0.0`. I scanned all 16 consumed changesets: exactly one mentions rujira, and it declares **patch**:

```
safe-rujira-pairs: '@vultisig/rujira': patch
```

No changeset in that PR declares `major` for anything, for any package. So the major is not coming from an author's intent.

## root cause

`packages/rujira/package.json`:

```json
"peerDependencies": { "@vultisig/sdk": ">=4.5.0" }
```

and `.changeset/config.json` does not set `onlyUpdatePeerDependentsWhenOutOfRange`.

Changesets' documented default is that when a package is a **peer** dependency of another and gets a version bump, every dependent receives a **major** bump, on the theory that a peer range change is breaking for consumers. That is exactly what the CHANGELOG records, release after release:

```
## 64.0.0
### Patch Changes
- Updated dependencies [...]:
  - @vultisig/sdk@4.5.0

## 63.0.0
### Patch Changes
- Updated dependencies [...]:
  - @vultisig/sdk@4.4.0
```

A "Patch Changes" section under a major version heading, containing nothing but a dependency update. Every entry in the file follows that shape.

Note the peer range is `>=4.5.0`, which is open-ended: an SDK bump to 4.6.0 does not violate it, so nothing about the dependency relationship actually broke.

## why it is worth fixing

1. **Consumers cannot use caret ranges.** `^64.0.0` will not accept `65.0.0`, so every SDK release requires a manual rujira bump in every downstream consumer, forever.
2. **A real breaking change is unsignallable.** When rujira genuinely breaks its API, the major bump announcing it is indistinguishable from the 61 that did not.
3. **It inflates the version indefinitely.** At roughly one SDK release per few days, rujira will pass 100.0.0 within the year on dependency updates alone.

## suggested fix

Set the changesets option that makes peer-dependent bumps conditional on the range actually being violated:

```json
{
  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
    "onlyUpdatePeerDependentsWhenOutOfRange": true
  }
}
```

With the peer range at `>=4.5.0`, SDK minors and patches then stop forcing a rujira major, and rujira majors only when the peer range is genuinely moved.

Worth deciding separately whether `>=4.5.0` is the range you want. An open-ended peer range means rujira claims compatibility with every future SDK major, which is its own overclaim in the other direction.

## scope note

This does not block #1876. That PR is doing exactly what the current config tells it to; the config is the bug. Flagging it there so whoever merges the release knows the `65.0.0` in the diff is not signalling a breaking change.

Found while reviewing #1876 as a never-reviewed PR.

gomesalexandre:
Closes vultisig/vultisig-sdk#1921

`@vultisig/rujira` declares `@vultisig/sdk` as a peer dependency with an open-ended range (`>=4.5.0`), but changesets' default behavior majors every peer-dependent package on any SDK version bump regardless of whether the peer range was actually violated. That produced 61 spurious majors out of 64 published rujira versions, making semver on that package meaningless and forcing consumers off caret ranges. Set `onlyUpdatePeerDependentsWhenOutOfRange: true` in `.changeset/config.json` so peer-dependent majors only fire when the peer range is genuinely out of date, and added a changeset documenting the fix.

## what I ran
No test suite applies — this is a `.changeset/config.json` tooling-config change consumed by the changesets CLI at release time, not application code covered by vitest. I did not run vitest, build, or typecheck for this change (verified the JSON is valid via the Edit tool's post-write state and re-reading the file). CI's normal release-PR flow will exercise the actual bump behavior.

## Summary by CodeRabbit

* **Chores**
  * Updated package release configuration to avoid unnecessary peer-dependent package updates when dependency ranges remain compatible.
  * Added a patch release note for the SDK reflecting the updated release behavior.


NeOMakinG (approved):
LGTM

PR #2051, opened 2026-08-18
test(swap): pin thorchain fee-coin ticker oracle for native-swap chains

### LINKED ISSUE #1697: toNativeSwapAsset builds THORChain notation from the display ticker — adding TON as a native-swap chain would emit TON.GRAM
## What
`toNativeSwapAsset` builds THORChain/MayaChain asset notation by interpolating a coin's **ticker**:

`packages/core/chain/swap/native/asset/toNativeSwapAsset.ts:60` and `:74`
```ts
return `${swapChainId}.${ticker}`
```

For a fee coin the ticker comes from `chainFeeCoin[chain]`. Since the Toncoin → GRAM rebrand, `chainFeeCoin[Chain.Ton].ticker` is `'GRAM'` (`packages/core/chain/coin/chainFeeCoin.ts:69-74`).

So if TON is ever added as a native-swap chain, this emits **`TON.GRAM`**, which is not the asset THORChain knows — the protocol notation is `TON.TON`, keyed on the chain's own symbol, not on whatever the token is currently branded as.

## Why this is latent rather than broken
`Chain.Ton` appears nowhere in `packages/core/chain/swap/native/NativeSwapChain.ts`, so `toNativeSwapAsset` throws `No native swap enabled chain found for Ton` before it can build anything. Nothing is wrong today.

The concern is the shape of the future change: adding a chain to `nativeSwapChainIds` is a small, self-contained-looking edit, and **nothing at that site points at the rebrand coupling**. Whoever adds TON gets a silently wrong asset string, and the failure surfaces as an upstream quote rejection rather than as anything pointing back at the ticker.

## Cross-platform context
iOS hit the live version of this coupling. Its `Coin.swapAsset` builds `"\(chain.swapAsset).\(chain.ticker)"` from a **chain-level** ticker, which is precisely why iOS kept its chain-level ticker as `"TON"` while rebranding only the coin-level one (vultisig/vultisig-ios#5022 → PR #5035). The two codebases are therefore protected by **different and non-equivalent** mechanisms:

- **iOS:** safe by data — the value interpolated is still `"TON"`.
- **SDK:** safe by reachability — TON can't reach the builder.

Only the first survives TON becoming swappable.

## Fix directions
1. Derive the asset symbol for a fee coin from a **protocol-stable** identifier rather than the display ticker — the chain's own symbol, as THORChain names it. This is the durable fix and mirrors what iOS does.
2. Or, at minimum, add a guard/test asserting that every chain in `nativeSwapChainIds` has a fee-coin ticker equal to its THORChain symbol, so adding a mismatched chain fails at build/test time instead of at quote time.

Option 2 alone would have caught this class of bug and is cheap.

## Notes
Same failure class as vultisig/vultisig-ios#5004 (a chain added to a routing table without its matching entry in a second, unreferenced table) — a second source of truth for the same value, correct-looking and wrong, with nothing linking the two.

gomesalexandre:
Closes vultisig/vultisig-sdk#1697

`toNativeSwapAsset`'s fee-coin branch builds THORChain/MayaChain asset notation as `${swapChainId}.${chainFeeCoin[chain].ticker}` — interpolating a DISPLAY ticker that can be rebranded (Toncoin's became `GRAM`), not the protocol-stable pool asset symbol. This is latent today because `Chain.Ton` isn't in `nativeSwapEnabledChainsRecord`, but nothing at the "add a chain to `nativeSwapChainIds`" call site points at this coupling, so a future contributor adding TON would silently emit `TON.GRAM` instead of `TON.TON`.

Implemented the issue's "Option 2" (explicitly called out as "cheap" and sufficient to catch this bug class): added a guard test in `packages/core/chain/swap/native/asset/toNativeSwapAsset.test.ts` that pins each native-swap-enabled chain's THORChain/MayaChain pool asset ticker in an **independently hardcoded oracle table** (not derived from `chainFeeCoin`, same pattern as this repo's existing CCTP-registry oracle test) and asserts `chainFeeCoin[chain].ticker` matches it. A companion test asserts every chain in `nativeSwapEnabledChains` has an oracle entry (so a newly-added chain can't slip through unguarded), and a third test documents the exact bug class using Ton's real, already-rebranded ticker (`GRAM` != `TON`) as a live demonstration — Ton isn't swap-enabled today, so this doesn't touch runtime behavior, it only proves what the guard is protecting against.

Did not implement fix direction 1 (deriving the fee-coin's protocol asset symbol from a chain-level protocol-stable identifier rather than the display ticker) — that's the "durable fix" but is a bigger, more invasive design change (a new stable-symbol source of truth threaded through `toNativeSwapAsset` and its callers) versus this cheap, targeted guard the issue itself flagged as sufficient.

## what I ran
`../../../node_modules/.bin/vitest run swap/native/asset/toNativeSwapAsset.test.ts` (from `packages/core/chain/`) — 14 passed (11 pre-existing + 3 new). All pre-existing tickers matched my independently-typed oracle values on the first run (no adjustment needed), which is a second confirmation the oracle table is correct today. I did not run the full test suite, typecheck, or a full build.

## Summary by CodeRabbit

* **Tests**
  * Added automated validation to ensure native-swap-enabled chains use the correct protocol asset tickers.
  * Added coverage for pool-asset ticker completeness and consistency across supported THORChain and MayaChain assets.
  * Added a regression check to ensure rebranded display tickers are not used as protocol asset symbols, helping maintain accurate swap processing.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev (commented):
After the base update, toNativeSwapAsset.test.ts still includes Chain.Kujira in the oracle rows, but current main removed Kujira from both Chain and the native-swap configuration. The exact-head Lint and Type Check now fails TS2339 on that reference. Please remove the stale oracle row; restoring complete Kujira support would be a separate change.

rcoderdev (approved):

rcoderdev (commented):
This is a test-only change, but `.changeset/w4-1697.md` declares an `@vultisig/sdk` patch release. The repository guidance says test-only changes need no changeset, so merging this would publish an unnecessary SDK version and release note. Please remove the changeset.

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

PR #2052, opened 2026-08-18
fix(stakekit): degrade parseActionDisplay per-step, not all-or-nothing

### LINKED ISSUE #1731: architecture(stakekit): parseActionDisplay drops all canonicalized steps when one network is unsupported
## Summary
`parseActionDisplay()` canonicalizes supported StakeKit action steps, but if **any** transaction uses an unsupported network it falls back to raw decoded transactions for **all** steps.

## Evidence
- `packages/sdk/src/tools/defi/stakekit/index.ts:315-327`
  - unsupported networks return `null` during canonicalization.
- `packages/sdk/src/tools/defi/stakekit/index.ts:330-343`
  - if any step is `null`, the function returns `decoded` for the entire transaction list.

## Impact
One unsupported Cardano/other step nukes canonicalized EVM/Solana/Sui/Tron/TON envelopes for the rest of the action, pushing deterministic parsing back into app/backend consumers.

## Effort
Medium.

## Fix sketch
Degrade per-step instead of all-or-nothing: keep canonicalized steps canonical, and only leave unsupported steps in decoded fallback form. Preserve load-bearing top-level fields like `provider: "yield_xyz"`.

## Report
https://gist.github.com/gomesalexandre/c7bae58caec1ed98091332f24eb66c2d

gomesalexandre:
Closes vultisig/vultisig-sdk#1731

`parseActionDisplay()` canonicalizes StakeKit action steps (EVM/Solana/Sui/Tron/TON), but if ANY step's network couldn't be canonicalized (e.g. Cardano, which has no canonicalizer branch), the old code fell back to `decoded[]` for the ENTIRE action — nuking canonicalization of every other, perfectly-supported step. Fixed in `packages/sdk/src/tools/defi/stakekit/index.ts`: replaced the `allCanonicalized ? canonicalSteps : []` all-or-nothing branch with a per-step merge — `canonicalSteps.map((step, i) => step ?? decoded[i])` — so each step independently stays canonical if it canonicalized, or falls back to its own decoded form if it didn't. Top-level fields (`provider: "yield_xyz"`, `chain`) are untouched, per the issue's fix sketch.

Updated the existing test that pinned the OLD all-or-nothing behavior (renamed + rewritten to assert the good step stays canonical while only the bad step degrades), and added a new test for the exact scenario in the issue: an EVM step alongside an unsupported-network (Cardano) step, asserting the EVM step keeps its flat canonical shape while only the Cardano step degrades.

## what I ran
`../../node_modules/.bin/vitest run tests/unit/tools/defi/stakekit.test.ts` (from `packages/sdk/`) — 19 passed (17 pre-existing incl. 1 rewritten + 1 new). I did not run the full test suite, typecheck, or a full build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction display parsing so each action step is processed independently.
  * Preserved canonical formatting for supported steps while falling back only for unsupported or malformed steps.
  * Prevented one failed transaction step from affecting the display of other steps.


NeOMakinG (approved):
LGTM

PR #2056, opened 2026-08-18
fix(ton): raise jetton gas floor for first-time recipient wallet deploy

### LINKED ISSUE #1465: ton: fee is fully static (no dynamic path) and does not scale for first-time jetton recipients -> bounce/reject
Fee-hunt finding. TON baseFee=0.01, jettonAmount=0.08 hardcoded (chains/ton/config.ts:6-7), no RPC/estimate path anywhere (fee/resolvers/ton.ts). The flat 0.08 TON jetton amount does not account for the extra gas to deploy a new jetton-wallet contract for a FIRST-TIME recipient (forwardAmount at jetton.ts:30 only adjusts the notify amount). First-time-recipient jetton sends bounce/reject (TON returns unused value minus burned gas, so no stranded funds, but a clean fail). Fix: dry-run via toncenter estimateFee before signing; raise the floor for non-active destinations.

gomesalexandre:
Closes vultisig/vultisig-sdk#1465

TON jetton sends attach a flat `tonConfig.jettonAmount` (0.08 TON) gas envelope regardless of destination state (`packages/core/chain/chains/ton/config.ts`), consumed by the on-device MPC keysign resolver `buildJettonTransfer` (`packages/core/mpc/keysign/signingInputs/resolvers/ton/jetton.ts`). A first-time (not-yet-active) recipient needs its jetton wallet deployed as part of the transfer, which costs materially more gas than a transfer into an already-deployed wallet — the resolver already receives `isActiveDestination` and used it to toggle the internal `forwardAmount` (1 vs 0 nanoton) but never scaled the outer gas envelope, so first-time-recipient sends under-fund the tx and TON bounces/rejects it. Added a `tonConfig.jettonAmountNewWallet` floor (0.15 TON) and wired `buildJettonTransfer` to use it when `!isActiveDestination`.

**Scope note (partial fix, by design):** the issue also names a second code path with the identical flat-gas bug — the SDK's own pure `buildTonJettonTransferTx` builder (`packages/sdk/src/chains/ton/tx.ts`, `JETTON_GAS_AMOUNT_NANO`). I did not touch that one: it's pinned by `packages/sdk/tests/unit/platforms/react-native/ton-golden-vectors.test.ts` against `testdata/cross-encoder-golden/*.json` fixtures that assert `fixture.gasAmountNanotons === JETTON_GAS_AMOUNT_NANO` for both active and inactive fixtures — those fixtures look cross-language-generated (used to prove the TS encoder matches a canonical reference encoder), so changing the constant there would desync from that reference rather than fix it, and regenerating the reference is out of scope for a small fix. I also did not implement the issue's "dry-run via toncenter estimateFee before signing" — that requires threading a network call into a currently-synchronous, pure signing-input resolver, a materially larger architectural change.

## what I ran
`./node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/core/mpc/keysign/signingInputs/resolvers/ton/jetton.test.ts packages/core/mpc/keysign/signingInputs/resolvers/ton/native.test.ts` (from repo root) — 9 passed (9). `jetton.ts` had no existing test file, so I added one (`jetton.test.ts`) covering both the active-destination flat-floor case and the new not-yet-active raised-floor case, plus the existing sibling `forwardAmount` behavior, mirroring the conventions of the sibling `native.test.ts`. Note: this repo's canonical way to test `packages/core/**` is `yarn test:core` (`.config/vitest.core.config.ts`, aliases straight to source) — NOT the SDK's own `vitest.config.ts`, which only globs `packages/sdk/tests/**` and would silently skip this file. I initially ran it through the SDK's vitest binary pointed at core/mpc's built `dist/`, got a stale-value failure, and switched to the correct `test:core` config once I found it, which resolves directly to `packages/core/chain/chains/ton/config.ts` source (no rebuild needed, and no risk of me editing source but testing dist).


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev (approved):

rcoderdev:
Closing this stale PR because the current implementation does not reliably solve the first-time jetton-recipient failure it targets. The higher gas floor is selected from the recipient owner TON account state, but an active TON account can still lack the specific recipient jetton-wallet contract, so that common first-token case can still receive the old 0.08 TON envelope and bounce. The inactive golden fixture also declares 80,000,000 while its updated hash encodes 150,000,000, and the tests do not assert that contract field. A replacement should derive the recipient jetton-wallet deployment state from the receiver plus jetton master (or an equivalent authoritative signal), exercise resolver-to-builder behavior, and make the cross-language fixture assertion consistent.

PR #2057, opened 2026-08-18
fix(evm): prefer pending block tag for nonce fetch to avoid external-wallet collisions

### LINKED ISSUE #144: EVM nonce: account for external pending transactions
## Problem

When an external wallet (e.g. MetaMask) submits a transaction using nonce `N` that hasn't confirmed yet, `patchEvmNonce()` may assign the same nonce `N` because it uses `eth_getTransactionCount("latest")` (confirmed count) as the base.

Currently `fetchEvmPendingNonce()` is only called as a validation step when local state diverges from on-chain — it doesn't feed into the base nonce calculation.

## Proposed Fix

Call `fetchEvmPendingNonce()` unconditionally and use `max(rpcNonce, pendingNonce)` as the base before calling `getNextEvmNonce()`:

```typescript
const pendingNonce = await this.fetchEvmPendingNonce(chain)
const chainNonce = pendingNonce !== null && pendingNonce > rpcNonce ? pendingNonce : rpcNonce
const nextNonce = this.stateStore.getNextEvmNonce(chain, chainNonce)
```

## Trade-offs

- Adds an extra RPC call (`eth_getTransactionCount` with `"pending"`) to every EVM transaction build
- MPC vaults rarely have external wallet activity, so this is an uncommon scenario
- The `"pending"` tag behavior varies across RPC providers

## Source

CodeRabbit review on PR #132.

gomesalexandre:
Closes vultisig/vultisig-sdk#144

`evmTxInfo` (`packages/sdk/src/tools/evm/evmTxInfo.ts`) fetched the nonce via `client.getTransactionCount({ address })`, which defaults to the `'latest'` (confirmed) block tag. If an external wallet (e.g. MetaMask) had already submitted a transaction with nonce N that hasn't confirmed yet, `evmTxInfo` would hand out the same nonce N for the next MPC-signed tx, causing one of the two to get stuck or dropped. The original issue proposed a `max(rpcNonce, pendingNonce)` state-store pattern referencing code (`patchEvmNonce`/`fetchEvmPendingNonce`) that no longer exists in the current codebase — but the underlying bug is present in the current `evmTxInfo` nonce fetch. Fixed by mirroring the SDK's own already-shipped precedent for this exact problem: `packages/sdk/src/platforms/react-native/chains/evm/rpc.ts`'s `getEvmNonce` tries `blockTag: 'pending'` first (counts mempool-visible txs from any sender) and falls back to `'latest'` only if the chain rejects the pending tag (zkSync Era, Hyperliquid, some alt-EVMs don't support it cleanly).

## what I ran
`cd packages/sdk && ../../node_modules/.bin/vitest run tests/unit/tools/evm/evmTxInfo.test.ts` — 5 passed (5, the 3 pre-existing tests plus 2 new ones). Added tests asserting (1) `getTransactionCount` is called with `blockTag: 'pending'` and never with `'latest'` on the happy path, and (2) it falls back to a second call with `blockTag: 'latest'` when the pending-tag call rejects. The 3 pre-existing tests needed no changes — their mock resolves unconditionally regardless of the `blockTag` argument, so they kept passing as-is. Did not run `yarn typecheck` or `yarn build:sdk:bundle` — out of scope per instructions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved EVM transaction nonce selection by accounting for pending mempool transactions.
  - Added a fallback for networks that do not reliably support pending transaction data.


NeOMakinG (approved):
LGTM

gomesalexandre (commented):
Focused pass on the current head looks sane.

What I checked
- diff read against the current head
- focused receipt: `./node_modules/.bin/vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/tools/evm/evmTxInfo.test.ts`
  - 5/5 tests passed, including the new `pending`-tag preference and the `latest` fallback path
- attempted `yarn build:sdk` in an exact-head temp clone
  - blocked by a pre-existing mainline type error in `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts:72` (`auxiliaryData` not in `ISigningInput`)
  - same failure reproduces on `origin/main`, so I am not treating that as a regression from this PR

Review take
- switching `evmTxInfo` nonce lookup to prefer `blockTag: 'pending'` matches the existing RN convention and closes the stale-nonce race with external wallets
- the fallback to `latest` for chains that reject `pending` is covered and looks reasonable
- no blocker found in the current head from this pass

Merge gate note
- the visible approval is on `797fd251b47ae22ceb0efa9590f850f3358c87ab`, but the current head is `8acc17d140711643001608851390248194e46c39`
- so this still needs a fresh non-gomes current-head approval before merge

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev:
Updated this against current main. The nonce path now falls back to latest only for explicit pending-tag rejection; timeout, rate-limit and server failures propagate unchanged. Nine focused tests pass, including the returned pending nonce and wrapped RPC errors. The current head has fresh non-author approval. The earlier Cardano build note was a mainline limitation; required CI still gates merge.

rcoderdev (approved):

PR #2061, opened 2026-08-18
feat(sdk): export the canonical xrp destination/x-address normalization helpers

### LINKED ISSUE #1411: architecture: XRP destination/X-address normalization helpers are missing from the public SDK surfaces

gomesalexandre:
Closes vultisig/vultisig-sdk#1411

`packages/core/chain/chains/ripple/address.ts` already exposed the canonical pure XRP destination/X-address helper family (`decodeRippleXAddress`, `isValidRippleXAddress`, `normalizeRippleDestination`, `encodeRippleXAddress`) and it's already bundled into the RN platform build transitively (used by `packages/sdk/src/platforms/react-native/chains/ripple/tx.ts`, the XRP send-tx builder), but neither the root `@vultisig/sdk` nor the `@vultisig/sdk/react-native` entrypoint re-exported it. `clients/cli/src/commands/transaction.ts` had to deep-import it from `@vultisig/core-chain/chains/ripple/address` instead of staying on the SDK boundary. Fixed by adding the same export block to both `packages/sdk/src/index.ts` and `packages/sdk/src/platforms/react-native/index.ts`, placed right after the existing (already-exported) XRPL issued-currency canonicals for parity, and switched the CLI's deep import to the public `@vultisig/sdk` one. Confirmed the added RN dependency (`ripple-address-codec`, via `@scure/base` + `@xrplf/isomorphic`, both cross-platform/isomorphic packages) introduces no new RN-safety risk since it's already pulled into the RN bundle by the existing XRP tx builder.

## a bug I caught before opening this

The worker who wrote this fix honestly disclosed it could not run its own two new tests in its worktree (missing built core-config package) and verified by code inspection instead. I ran them myself against a working build before opening this PR - both FAILED: the fixture classic address `rf7SyXdM3aZqkz9bmgGgX6V3eC8oJ8wxYY` has an invalid base58 checksum, so `encodeRippleXAddress` correctly threw `checksum_invalid` on it. Not a bug in the production code - the function was rejecting bad input exactly as it should. Replaced the fixture with a real valid classic address generated from the installed `ripple-address-codec` itself (`encodeAccountID` on a deterministic payload, confirmed with `isValidClassicAddress`). 103/103 pass now.

## what I ran
Attempted `/tmp/claude-1000/wt-sdk-w1/node_modules/.bin/vitest run tests/unit/platforms/react-native/entry.test.ts` and `tests/unit/utils/publicExports.test.ts` (from `packages/sdk`) — both FAILED to execute, but for a reason unrelated to my change: this worktree has no built `packages/core/config/dist`, and BOTH files transitively import something (`chains/evm/chainInfo.ts` for the RN entry test, `chains/cardano/client/config.ts` for publicExports) that bare-imports `@vultisig/core-config`, which fails module resolution without a build. Confirmed this is pre-existing and independent of my change: the RN entry test failed on ALL 47 of its tests (not just my 2 new ones), including tests that predate this PR by a wide margin (e.g. dangerous-address re-exports). The only real fix is `yarn build:shared` (full monolithic 6-package tsc build), which the task instructions explicitly forbid. I verified correctness by code inspection instead: the added export blocks are byte-identical in shape to the immediately-adjacent, already-shipped `issuedCurrency` export blocks in both files; the underlying `address.ts` functions are unchanged (only re-exported); and I traced the RN dependency chain to confirm `ripple-address-codec` is already bundled into the RN platform build via the existing XRP send-tx path, so this isn't a first-time-untested RN dependency. Both new test blocks will run correctly in CI once the shared packages are built. Did not run the full suite, typecheck, or build.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

PR #2062, opened 2026-08-18
chore: delete dead RIVER_FACTORY_ABI and BITTENSOR_BASE_FEE_RAO exports

### LINKED ISSUE #1375: Break mpc-types<->mpc-wasm circular dependency; delete dead RIVER_FACTORY_ABI/BITTENSOR_BASE_FEE_RAO
Parent epic: #1347

From the whole-stack audit round 7 (2026-07-17). Severity P3 CONFIRMED.

**Anchor:** mpc-types/src/runtime.ts:244 - mpc-types/package.json:34-39 - mpc-wasm/src/index.ts:8-16 - mpc-wasm/package.json:17-19 - sdk/src/tools/defi/river/abi.ts:8 - sdk/src/tools/balance/bittensor.ts:33

## what
mpc-types peerDepends (optional) on mpc-wasm and dynamically imports it in ensureMpcEngine(); mpc-wasm has a real dependency back on mpc-types. Runtime-safe today only because the types-to-wasm edge is a lazy import(). Also two confirmed dead exports: RIVER_FACTORY_ABI (sdk/src/tools/defi/river/abi.ts) has zero importers anywhere in the workspace; BITTENSOR_BASE_FEE_RAO (sdk/src/tools/balance/bittensor.ts) is defined once and referenced nowhere - an unfinished fee computation.

## failure scenario
A future refactor that turns the types-to-wasm edge into a static import turns this into a hard circular init cycle, breaking bundling/publish for both packages the way #949 already documents for core-chain/core-mpc.

## fix
Extract the shared MPC engine/types that both packages need into a leaf package (or keep mpc-types fully passive and never import from it), so the dependency graph is a DAG. Delete the two dead constants, or finish the Bittensor fee path if it's meant to ship.

Related: #949 tracks the same circular-dependency architecture problem for core-chain/core-mpc - this is the same bug shape at the mpc-types/mpc-wasm package boundary.

gomesalexandre:
Closes vultisig/vultisig-sdk#1375

This issue bundled two things: (1) a genuine mpc-types<->mpc-wasm circular-dependency architecture risk (runtime-safe today only because the types-to-wasm edge is a lazy `import()`, but would become a hard cycle if a future refactor makes it static), and (2) two confirmed dead exports. I fixed the small, mechanical half only: deleted `RIVER_FACTORY_ABI` (`packages/sdk/src/tools/defi/river/abi.ts`, zero importers workspace-wide — the sibling ABIs in the same file, `RIVER_TROVE_MANAGER_ABI`/`RIVER_SORTED_TROVES_ABI`/`RIVER_BORROWER_OPS_ABI`/`RIVER_PERIPHERY_ABI`, are all actively used by `river.ts` and untouched) and `BITTENSOR_BASE_FEE_RAO` (`packages/sdk/src/tools/balance/bittensor.ts`, defined once, referenced nowhere — its display-string sibling `BITTENSOR_BASE_FEE_HUMAN` IS used by `taoBalance.ts` and stays). Confirmed via a repo-wide grep (excluding `node_modules`) that zero references to either constant remain anywhere, and neither was re-exported through any public SDK barrel, so this isn't a breaking public-API change. Did NOT attempt the circular-dependency extraction (moving the shared MPC engine/types into a leaf package to make the dependency graph a DAG) — that's a genuine multi-package architectural restructure, not a small fix, and risks breaking the build/publish pipeline for both packages if done carelessly.

## what I ran
Attempted `/tmp/claude-1000/wt-sdk-w1/node_modules/.bin/vitest run tests/unit/tools/balance/otherBalance.test.ts tests/unit/tools/defi/river.test.ts` (from `packages/sdk`) — both test files FAILED to execute, but for the same pre-existing, unrelated reason documented on my other issues in this batch: this worktree has no built `packages/core/config/dist`, and both files transitively import `chains/evm/chainInfo.ts` (via `river.ts`'s viem/EVM client usage and `otherBalance.test.ts`'s balance-chain imports), which bare-imports `@vultisig/core-config`. The only fix is the full `yarn build:shared` monolithic tsc build, explicitly forbidden by the task instructions. Verified correctness instead by a repo-wide `grep -rn "RIVER_FACTORY_ABI\|BITTENSOR_BASE_FEE_RAO" . --include="*.ts" --include="*.tsx"` (excluding node_modules) returning zero hits after the deletion, and by reviewing the final diff to confirm only the two flagged dead lines were removed with no adjacent code disturbed. Did not run the full suite, typecheck, or build.

## Summary by CodeRabbit

* **Chores**
  * Removed unused internal SDK exports related to Bittensor fee values and River factory contract definitions.
  * Existing balance validation, RPC fallback, parsing, and River functionality remain unchanged.


NeOMakinG (approved):
LGTM

PR #1214, opened 2026-07-12
fix(cctp): base messageTransmitter is a codeless eoa - burn-without-mint fund loss

gomesalexandre:
Closes vultisig/vultisig-sdk#1213

## what

Base-bound CCTP claims were pointed at `0xAD09780d193884d503182aD4F75D113B9B9a86E7` - a codeless EOA (`eth_getCode` = `0x`), NOT Circle's MessageTransmitter. `receiveMessage` against an EOA "succeeds" with no mint, so the bridge burns USDC on source and never mints on Base = silent fund loss. Nasty lookalike too: same prefix as the real address, different tail.

## how

- Corrected the Base `messageTransmitter` in `packages/sdk/src/tools/bridge/cctp.ts` to Circle's published `0xAD09780d193884d503182aD4588450C416D6F9D4` (https://developers.circle.com/cctp/v1/evm-smart-contracts)
- Pinned the ENTIRE registry (6 chains x tokenMessenger/messageTransmitter/usdc) to Circle's published V1 addresses via an oracle test in `bridge-cctp.test.ts` - any future drift fails loud
- Changeset (patch)

## audit

Verified every registry entry against Circle's page AND on-chain `eth_getCode` (Ethereum/Avalanche/Optimism/Arbitrum/Base/Polygon): only the Base transmitter was wrong, everything else has code and matches Circle.

```
eth_getCode(0xAD09780d...F75D113B9B9a86E7) on Base -> 0x            (the old value, EOA)
eth_getCode(0xAD09780d...588450C416D6F9D4) on Base -> 0x6080604...  (Circle's, contract)
```

## testing

- fail-before: with the old address, new oracle tests fail (`Tests 2 failed | 14 passed`)
- pass-after: `vitest run tests/unit/bridge-cctp.test.ts` -> `16 passed (16)`
- lint clean on both touched files; sdk typecheck errors are pre-existing skip-build artifacts (mpc-wasm/mpc-types), none in this diff

## risk

Low - single constant correction to Circle's published value + tests. Can't bork what's already borked (the old address never minted anything).

🤖 Agent-generated

PR #2063, opened 2026-08-18
feat(cctp): add a runtime fail-closed getCode guard on the mint-side transmitter

### LINKED ISSUE #1287: feat(cctp): runtime fail-closed getcode guard on the mint-side transmitter (defense-in-depth past #1214)
Follow-up to PR #1214 / issue #1213 (Base messageTransmitter was a codeless EOA - burn-without-mint fund-loss class). The PR corrects the address and pins all 18 CCTP addresses to Circle's published set via an oracle test, which catches registry drift in CI.

Defense-in-depth gap that remains: nothing verifies AT RUNTIME that the mint-side contract actually has code before the burn leg executes. The failure class (registry drift, a wrong chain entry, a Circle redeploy) burns user USDC on the source chain with no mint path - the oracle test only helps if the drift goes through CI.

## fix shape

At CCTP route-build time (before the burn tx is surfaced for signing), eth_getCode the destination messageTransmitter (and arguably the tokenMessenger) - empty code -> fail closed with a clear route-unavailable error, never build the burn. Cache positive results per (chain, address) for the session to avoid per-quote RPC spam. Follow the fetch-safe/timeout conventions from the RN network hardening work (#1250 family).

effort: S-M / risk: fund-safety guard (fail-closed only - a false negative just blocks a route temporarily) / confidence: HIGH

found during the #1214 deep receipt run 2026-07-15 (the run verified all 18 current addresses have code - this guards the NEXT drift)

gomesalexandre:
Closes vultisig/vultisig-sdk#1287

Follow-up to #1214/#1213 (Base messageTransmitter was a codeless EOA — burn-without-mint fund-loss class). That PR corrected the address and pinned all 18 CCTP addresses against Circle's published set via a CI oracle test, but nothing verified AT RUNTIME that the mint-side contract still has code before the burn leg executes — the oracle test only helps if drift goes through CI.

Added `assertCctpMintSideHasCode(destinationChain: string): Promise<void>` in `packages/sdk/src/tools/bridge/assertCctpMintSideCode.ts`, matching the issue's fix shape: resolves the destination's `messageTransmitter` address from the existing CCTP registry, calls `getEvmClient(chain).getCode({ address })` (the same viem `PublicClient` dependency already used by `tools/evm/*`, so it's RN-safe by the existing precedent), and throws `CctpMintSideCodeError` if the address has no code (`0x` or `undefined`) or the destination chain isn't CCTP-supported. Positive results are cached per `(chain, address)` for the process lifetime to avoid per-quote RPC spam; negative results are never cached, so a transient RPC hiccup can't calcify into a permanent block — fail-closed only, never signs or broadcasts. Deliberately kept as a standalone, additive async guard (not threaded into `buildCctpBridge`) so the existing pure/sync `buildCctpBridge` contract and every current caller's signature stays unchanged; callers invoke this guard themselves at route-build time, right before surfacing the burn tx for signing, per the issue's own phrasing.

Exported from the bridge barrel (`tools/bridge/index.ts`), and propagated up through `tools/index.ts` and the root/RN SDK entry points, matching how every other CCTP export (`buildCctpBridge`, `cctpChains`, etc.) is already surfaced at all three layers.

## what I ran
`../../node_modules/.bin/vitest run tests/unit/tools/bridge/assertCctpMintSideCode.test.ts` (from `packages/sdk/`) — 6 new tests, all passed: resolves when code is present, fails closed on `0x` code, fails closed on `undefined` code, rejects an unsupported destination chain before touching the network, caches a positive result (no RPC re-hit on repeat calls for the same chain), and does NOT cache a failure (a subsequent success after a failure re-hits the RPC). Also ran `../../node_modules/.bin/vitest run tests/unit/bridge-cctp.test.ts tests/unit/platforms/react-native/entry.test.ts` as a regression check — all passed (one run hit the same pre-existing transient 5s transform-timeout on an unrelated dangerous-address test under shared-machine load; a clean re-run of that file alone passed all 46, confirming it's not caused by this change). I did not run the full test suite, typecheck, or a full build.

Note: while cross-checking the CCTP export propagation pattern for this issue, I noticed my earlier fix for vultisig-sdk#1910 (`extractCctpMessageFromReceipt`, branch `fix/w4-1910`) only added the export to the `tools/bridge` barrel and missed propagating it to `tools/index.ts` and the root SDK entry. Flagging here since it's the same class of gap — will fix on that branch separately.

## Summary by CodeRabbit

- **New Features**
  - Added CCTP destination validation to confirm the required on-chain contract is deployed before a burn transaction can be signed.
  - Added clear errors for unsupported destinations or unavailable contract code.
  - Exposed the validation utility across SDK and React Native integrations.
  - Successful validation results are reused for improved efficiency, while failed checks can be retried.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev:
Closing rather than merging: this adds an optional guard, but the public synchronous buildCctpBridge path still returns signing-ready burn calldata without enforcing it, so the change does not satisfy #1287's fail-closed requirement. Making that public builder async would be a breaking API change and needs an explicit design at the async route/signing boundary. Please reopen or replace this with an implementation that proves a codeless mint-side destination cannot surface a burn transaction.

PR #2065, opened 2026-08-18
fix(sdk): floor rn getEvmSuggestedFees priority fee via clampEvmPriorityFee

### LINKED ISSUE #1178: react-native getEvmSuggestedFees still uses naive baseFee/10 while app/backend already floor quiet-block L2 fees
## Summary
react-native getEvmSuggestedFees still uses naive baseFee/10 while app/backend already floor quiet-block L2 fees

## Evidence
- `packages/sdk/src/platforms/react-native/chains/evm/rpc.ts:54-74` still returns naive EIP-1559 fee math with no per-chain floors.
- `src/services/evmTx.ts:36-107` in the app now carries explicit minimum priority/max-fee tables for Polygon, BSC, Arbitrum, Optimism, Base, Mantle, Sei, and Hyperliquid.
- That app comment explicitly says the policy mirrors agent-backend-ts gas floors, so the SDK helper is now the stale copy.

## Impact
Any RN/SDK consumer that trusts `getEvmSuggestedFees()` can still underprice quiet-block L2 transactions even though the app and backend already learned the real floor policy. This is exactly the drift class the campaign keeps finding: one consumer hotfixed, the shared helper stayed stale.

## Fix sketch
Promote the canonical floor policy into the SDK fee helper (or a shared SDK EVM fee-policy module), add unit coverage for affected chains, and delete downstream copies in app/backend.

## SDK / CLI testability
Yes — pure fee-policy math with unit tests and CLI dry-run verification.

## Effort
medium

## Round report
https://gist.github.com/gomesalexandre/d72e9c0a3dd7883345b63e6cbc98f53e

gomesalexandre:
Closes vultisig/vultisig-sdk#1178

`packages/sdk/src/platforms/react-native/chains/evm/rpc.ts`'s `getEvmSuggestedFees` computed `maxPriorityFeePerGas` as naive `baseFee / 10` with no per-chain floor — confirmed still the case on `main`. The SDK already publishes a canonical per-chain floor/ceiling policy at `@vultisig/core-chain/tx/fee/evm/clampEvmPriorityFee` (already exported from the SDK root as `sdk.clampEvmPriorityFee`), which floors Ethereum at 1 gwei and Polygon at 30 gwei — mirroring what the mobile clients already do — and sanity-ceilings every EVM chain against an anomalous RPC quote. But nothing inside the SDK, including this RN fee helper, actually called it; it was purely an opt-in export.

The issue's evidence pointed at the app's `evmTx.ts` having a broader floor table (Polygon, BSC, Arbitrum, Optimism, Base, Mantle, Sei, Hyperliquid) than `clampEvmPriorityFee` currently covers (which deliberately floors only Ethereum + Polygon, on the documented reasoning that "L2 sequencers include txs regardless of tip, so rollups deliberately have no floor"). I don't have visibility into the app repo's exact floor values from this worktree to responsibly reconcile that broader table, so I scoped this fix to wiring the RN helper through the SDK's own existing canonical policy rather than inventing new per-chain floor numbers I can't verify. This directly fixes the quiet-block Ethereum/Polygon underpricing the issue describes and brings the RN helper's fee math in line with what `clampEvmPriorityFee` already offers every other consumer.

## what I ran
`/tmp/claude-1000/wt-sdk-w2/node_modules/.bin/vitest run tests/unit/platforms/react-native/evm-rpc.test.ts` (run from `packages/sdk`) — 1 test file, 13 tests, all passed (8 pre-existing broadcast-idempotency tests + 5 new: quiet-block floor on Ethereum (1 gwei), floor on Polygon (30 gwei), no floor on an L2 with none configured (Arbitrum), sanity-ceiling clamp on a wildly inflated RPC quote, and the `getGasPrice` fallback path when `baseFeePerGas` is null).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved EVM fee suggestions by applying chain-specific minimum priority fees and safety limits.
  - Preserved expected behavior for Arbitrum and fallback scenarios when base-fee data is unavailable.
  - Added safeguards to prevent unusually high priority-fee recommendations.

- **Tests**
  - Added coverage for fee floors, ceilings, chain-specific behavior, and gas-price fallback handling.


NeOMakinG (approved):
LGTM

PR #2066, opened 2026-08-18
feat(sdk): surface curated xrpl issued currencies through searchToken

### LINKED ISSUE #1219: searchToken still cannot surface curated XRPL issued currencies from knownTokens
## Summary
The SDK now knows about curated XRPL issued currencies like RLUSD internally, but the public `searchToken()` surface still only models discovery as CoinGecko contract deployments.

## Evidence
- `packages/core/chain/chains/ripple/issuedCurrency.ts:146-164` defines curated XRPL issued tokens
- `packages/core/chain/coin/knownTokens/index.ts:800-805` indexes those tokens into `knownTokens`
- `packages/sdk/src/index.ts:788` exports `searchToken`
- `packages/sdk/src/tools/token/searchToken.ts:71-137` only emits deployments when `detail_platforms[*].contract_address` exists

## Impact
- XRPL issued currencies are internally known but still absent from the public search surface.
- That pressures downstream consumers to add more local token-search shims instead of consuming one canonical SDK discovery path.

## Fix sketch
- Overlay `knownTokens`-backed results into `searchToken()` for non-contract token-id families such as XRPL issued currencies.
- Add unit coverage showing RLUSD is discoverable through the public search surface.
- Be explicit about result-shape semantics for non-contract ids.

## Effort
Medium

## Report
https://gist.github.com/gomesalexandre/b8e6fe15a62b7d5a6843b1174225e65f

gomesalexandre:
Closes vultisig/vultisig-sdk#1219

`packages/sdk/src/tools/token/searchToken.ts` only modeled token discovery as CoinGecko "platform deployments" — a `detail_platforms[*].contract_address`. XRPL issued currencies don't have a contract address (they're identified by `(currency, issuer)`), so CoinGecko never lists a `ripple` platform entry for them, and the SDK's own curated `knownTokens[Chain.Ripple]` list (RLUSD) was never surfaced through `searchToken()` even though it's known internally.

Added an overlay: `knownTokens[Chain.Ripple]` entries are indexed by their `priceProviderId` (the CoinGecko coin id each curated entry already carries — e.g. RLUSD's is `'ripple-usd'`), and matched against each search result's `coin.id`. When matched, a `{ chain: Chain.Ripple, contractAddress: <currency>.<issuer>, decimals }` deployment is pushed onto that result's `deployments`, unless a Ripple deployment is already present from CoinGecko's own detail response. Per the issue's ask to "be explicit about result-shape semantics for non-contract ids," documented `TokenDeployment.contractAddress` as the chain's canonical token identifier generically (contract address / SPL mint / Sui coin type / XRPL composite id), not strictly an EVM contract — matching how the field was already being used for Solana/Sui/Cardano/etc.

## what I ran
`/tmp/claude-1000/wt-sdk-w2/node_modules/.bin/vitest run tests/unit/tools/token/searchToken.test.ts` (run from `packages/sdk`) — 1 test file, 7 tests, all passed (5 pre-existing + 2 new: one proving `searchToken('RLUSD')` surfaces the curated Ripple deployment alongside CoinGecko's own Ethereum deployment for the same coin, and one proving no Ripple deployment is added for an unrelated coin like USDC). Also ran `tests/unit/utils/publicExports.test.ts` — 55 tests, all passed, confirming the new `@vultisig/core-chain/coin/knownTokens` import doesn't break module resolution.


NeOMakinG (approved):
LGTM

PR #2068, opened 2026-08-18
fix(sdk): route rn balance reads off AbortSignal.timeout onto the hermes-safe pattern

### LINKED ISSUE #1343: react-native balance reads still depend on AbortSignal.timeout in shared helpers

gomesalexandre:
Closes vultisig/vultisig-sdk#1343

Four RN-exported balance-read helpers called `AbortSignal.timeout` directly, which doesn't exist on older RN/Hermes runtimes: `packages/sdk/src/tools/balance/rpc.ts`'s `fetchJson`, `cosmos.ts`, `utxoBalance.ts`, and `bittensor.ts`. The SDK already carries a Hermes-safe `AbortController` + `setTimeout` pattern elsewhere (`chains/tron/rpc.ts`, `tools/gas/utxoFeeRate.ts`), and one sibling in the same directory (`tools/balance/polkadot.ts`) already routes through `queryUrl`, which uses the same safe pattern internally — these four were the outliers.

Routed all four through the Hermes-safe pattern, preserving each file's existing retry/HTTP semantics exactly:
- **`rpc.ts`'s `fetchJson`**: builds its own controller only when the caller didn't already pass `init.signal` (same precedence as before); a bare `controller.abort()` still produces a `DOMException` named `'AbortError'`, which is exactly what `isRetryable` already checked for.
- **`cosmos.ts`'s `fetchJson`**: aborts with an explicit `new DOMException(..., 'TimeoutError')` so the existing `isTimeout` retry/fallback check (which specifically checks for `'TimeoutError'`, matching native `AbortSignal.timeout()`'s error name) keeps working unchanged.
- **`utxoBalance.ts`**: straightforward swap, no retry logic to preserve.
- **`bittensor.ts`**: kept the per-request 4s abort (`BITTENSOR_RPC_TIMEOUT_MS`) distinct from `fetchJson`'s own 15s default, so a slow endpoint's underlying connection is actually released instead of running in the background past the outer `withEndpointTimeout` `Promise.race` (which was already Hermes-safe).

## what I ran
`/tmp/claude-1000/wt-sdk-w2/node_modules/.bin/vitest run tests/unit/tools/balance/rpc.test.ts tests/unit/tools/balance/bittensor.test.ts tests/unit/utxoBalance.test.ts tests/unit/cosmos/cosmosBalance.test.ts tests/unit/tools/balance/otherBalance.test.ts` (run from `packages/sdk`) — 5 test files, 69 tests, all passed. Per the issue's fix sketch ("Add regression tests that remove [AbortSignal.timeout] at runtime and prove the reads still succeed"), added/extended:
- New `tests/unit/tools/balance/rpc.test.ts` — 5 tests on `fetchJson` directly, including a Hermes-simulation test (delete `globalThis.AbortSignal.timeout`, confirm the read still succeeds) and a retry-on-network-error test.
- New `tests/unit/tools/balance/bittensor.test.ts` — 3 tests on `getBittensorBalance` including the same Hermes-simulation regression, using a real SCALE-encoded `AccountInfo` fixture (5 TAO free balance, verified byte-for-byte against Substrate's LE u128 encoding).
- Extended `tests/unit/utxoBalance.test.ts` and `tests/unit/cosmos/cosmosBalance.test.ts` with the same Hermes-simulation regression against their respective public balance-read functions.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev (commented):
Please keep the timeout active until the response body has finished loading. In `rpc.ts`, `return response.json()` runs `finally` and clears the timer before JSON consumption completes; `utxoBalance.ts` likewise clears it before `response.text()`. A server that returns HTTP 200 and then stalls leaves supported balance reads pending indefinitely. Await body consumption inside the guarded `try` and add a stalled-body regression test.

Also preserve `TimeoutError` for deadline aborts in `rpc.ts` and Bittensor. Bare `controller.abort()` produces `AbortError`, which `isRetryable` retries; a previously bounded 15-second timeout can now take four requests and roughly 67 seconds.

rcoderdev:
Closing this because This overlaps #1344 on Cosmos, RPC and UTXO, while this head leaves stalled response bodies unbounded and turns deadline failures into retryable aborts. Safely landing the remaining Bittensor work needs a scoped rewrite, regression tests and fresh integration. We are closing this version; the Hermes issue remains open until a correct fix lands.

PR #2070, opened 2026-08-18
fix(tron): consolidate base58check address decoding into one shared helper

### LINKED ISSUE #1526: architecture: Tron address Base58Check decoding is duplicated and already drifted inside the sdk monorepo

gomesalexandre:
Closes vultisig/vultisig-sdk#1526

The Tron Base58Check address decoder was duplicated three ways in the monorepo and had already drifted: `packages/core/chain/coin/balance/resolvers/tron.ts`'s `base58CheckTronDecode` accepted both the mainnet (`0x41`) and Nile testnet (`0xa0`) network prefixes, but `packages/sdk/src/chains/tron/tx.ts`'s `tronAddressToBytes` only accepted `0x41` — and while auditing the evidence I found a third copy with the same 0x41-only bug in `packages/sdk/src/chains/tron/rpc.ts`'s `estimateTrc20Energy` (not named in the issue's evidence list, but the identical bug class in the same file family). All three silently rejected valid Nile testnet addresses.

Extracted the canonical checksum/prefix-validated decoder into a new shared module, `packages/core/chain/chains/tron/address.ts` (`decodeTronBase58Address`, returns the raw 21-byte payload), and made all three call sites delegate to it:
- `core-chain`'s `base58CheckTronDecode` now decodes via the shared helper then strips the prefix to hex (unchanged external behavior/contract).
- SDK's `tronAddressToBytes` now returns the shared helper's raw bytes directly (previously duplicated the bs58check module resolution + prefix check inline).
- SDK's `estimateTrc20Energy` now decodes `to` via the shared helper instead of its own inline bs58check-decode-plus-0x41-check.

## what I ran
`/tmp/claude-1000/wt-sdk-w2/node_modules/.bin/vitest run tests/unit/chains/tron-tx.test.ts tests/unit/chains/tron-rpc.test.ts` (run from `packages/sdk`) — 2 test files, 62 tests, all passed (59 pre-existing + 3 new: a `tronAddressToBytes` mainnet-decode test, a `tronAddressToBytes` Nile-testnet-decode regression test, and an `estimateTrc20Energy` Nile-testnet `to`-address regression test).

`node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/core/chain/coin/balance/resolvers/tron.test.ts` (run from repo root) — 1 test file, 13 tests, all passed, confirming `base58CheckTronDecode`'s external behavior (including its own pre-existing Nile-testnet test) is unchanged after delegating to the shared helper.


NeOMakinG (approved):
LGTM

rcoderdev (commented):
`TRON_NETWORK_PREFIXES` treats `0xa0` as Nile, but current TRON Mainnet, Shasta, and Nile addresses all use `0x41`; network selection is separate. With this decoder used by transaction construction and energy estimation, callers can now pass a synthetic `0xa0` Base58Check address and serialize invalid owner/recipient bytes. Please accept only `0x41` for current networks and replace the manufactured `0xa0` Nile tests; if private-chain prefixes are needed, model them explicitly rather than as Nile. Official reference: https://developers.tron.network/docs/encoding

rcoderdev:
Closing this because This branch encodes the false premise that 0xa0 is a Nile address prefix even though current Tron Mainnet, Shasta, and Nile use 0x41. Correcting the shared decoder requires re-scoping the prefix model and tests, so this implementation should not be carried forward.

PR #2076, opened 2026-08-18
feat(sdk): publish getSplAssociatedAccount isToken2022 resolver

### LINKED ISSUE #1728: architecture(solana): buildSplTransfer requires isToken2022 but the SDK exports no canonical resolver
## Summary

`buildSplTransfer` already treats `isToken2022` as a safety-critical caller input, but the SDK exports no stable companion helper that canonically resolves that value. Consumers are therefore pushed back into app/backend-local Solana token-program logic.

Report: https://gist.github.com/gomesalexandre/ee18a0bd3457f4a5d34d8fa319b6f2ee

## Evidence

- `packages/sdk/src/tools/prep/splTransfer.ts:71-81`
  - `isToken2022` is caller-supplied and documented as safety-relevant
- `packages/sdk/src/tools/prep/splTransfer.ts:177-185`
  - the flag controls `programId` and ATA derivation
- `packages/sdk/src/tools/balance/solana.ts:150-204`
  - the SDK already knows how to detect the token program internally
- `packages/sdk/src/platforms/react-native/overrides/getSplAssociatedAccount.ts`
  - there is already an internal helper shape that returns `{ address, isToken2022 }`, but it is not a stable public SDK surface

## Impact

A consumer that wants to build an SPL transfer safely cannot stay inside the reviewed public SDK contract:
- the builder requires `isToken2022`
- the SDK does not publish a canonical resolver for that flag
- first-party consumers are incentivized to deep-import or reimplement Solana token-program detection

This is the exact duplicated-not-imported drift class the campaign keeps finding.

## Fix sketch

Expose one stable public helper, e.g.:
- `resolveSplAssociatedAccount(...) -> { address, isToken2022 }`, or
- `resolveSplTokenProgram(...)`, or
- an async wrapper around `buildSplTransfer` that resolves the program metadata internally

Then move first-party consumers onto that helper instead of bespoke local logic.

## Effort

Medium. Mostly public-surface design plus wiring existing internal logic into a supported export.

gomesalexandre:
Closes vultisig/vultisig-sdk#1728

`buildSplTransfer` requires the caller to supply `isToken2022` (`packages/sdk/src/tools/prep/splTransfer.ts`), explicitly documented as fund-safety-critical since the builder does no RPC and trusts the flag verbatim — passing `false` for an actual Token-2022 mint derives the wrong ATA/program id (a failed tx or, worse, a credit to an unintended account). The SDK already had an internal helper matching the issue's suggested shape #1 exactly (`getSplAssociatedAccount(...) -> { address, isToken2022 }`, `packages/core/chain/chains/solana/spl/getSplAssociatedAccount.ts`), including an RN/Hermes-safe override (`packages/sdk/src/platforms/react-native/overrides/getSplAssociatedAccount.ts`, which defers its `@solana/web3.js` import inside the async body) already wired through `rollup.platforms.config.js`'s `rnOverrideMap` — but neither the root nor RN public SDK surface exported it. Added `export { getSplAssociatedAccount } from '@vultisig/core-chain/chains/solana/spl/getSplAssociatedAccount'` to both entrypoints. On RN this is safe as a *static* re-export (unlike `buildSplTransfer`'s existing lazy wrapper) because the rollup build already redirects the bare specifier to the deferred-import override at build time — I traced this through the existing `rnOverrideMap` mechanism before adding it, rather than assuming.

## what I ran
`cd packages/sdk && ../../node_modules/.bin/vitest run tests/unit/getSplAssociatedAccount.test.ts tests/unit/buildSplTransfer.test.ts tests/unit/utils/publicExports.test.ts tests/unit/platforms/react-native/entry.test.ts --testTimeout=30000` — 114 passed (114: 3 new in a new dedicated test file, 1 new in publicExports.test.ts, 1 new in entry.test.ts, rest pre-existing unchanged). `getSplAssociatedAccount` had no test file at all, so I added one mirroring the existing `rnGetSplAccountsOverride.test.ts` mocking pattern (`getSolanaClient().getParsedTokenAccountsByOwner` mocked, real `@solana/web3.js` `PublicKey` used for input validation): covers Token-2022-owned account → `isToken2022: true`, legacy-SPL-owned account → `isToken2022: false`, and the no-ATA-found throw path. Needed `--testTimeout=30000` on the RN entry test — first vitest transform of the newly-pulled `@solana/web3.js` graph took ~9s in this environment (cold esbuild cache), well within normal test tooling variance and unrelated to the change's correctness; a second run without the override passed in ~5s once transforms were warm. Did not run `yarn typecheck` or `yarn build:sdk:bundle` (which would prove the rollup override actually fires) — out of scope per instructions, though I read `rollup.platforms.config.js`'s `rnOverrideMap` by hand to confirm the redirect entry already exists for this exact file path.

## Summary by CodeRabbit

* **New Features**
  * Added `getSplAssociatedAccount` to the SDK and React Native public APIs.
  * Supports resolving associated accounts for both legacy SPL Tokens and Token-2022 assets.
  * Added guidance requiring token-program detection when building SPL transfers.

* **Bug Fixes**
  * Improved handling when no associated account exists by reporting an error.


NeOMakinG (approved):
LGTM

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):

rcoderdev:
Closing this because Closing this version of the public Token-2022 resolver: it can return a non-associated token account when multiple accounts match. Publishing it safely needs core and React Native fixes, account-order tests, and conflict resolution; no current first-party consumer uses the new export. The underlying idea remains open for a consumer-backed implementation.

PR #2094, opened 2026-08-18
feat(stakekit): add batched multi-network balances helper

### LINKED ISSUE #1765: architecture(stakekit): SDK lacks a batched balances helper, so the app keeps a local positions client
## Summary
The SDK only exposes single-network StakeKit balances, which still forces the app to keep a separate batched positions client.

## Evidence
- SDK `packages/sdk/src/tools/defi/stakekit/index.ts:495-500`
- app batched client `src/features/dashboard/lib/yieldXyzPositions.ts:4-20,233-255,380-386`

## Impact
- yield positions remain app-local instead of SDK-owned
- app/backend cannot share batching rules, payload shape, or response shaping

## Fix sketch
- add a batched StakeKit balances helper to the SDK
- expose it on root + RN surfaces
- replace the app-local batch client
- add fixture tests for chunk boundaries and multi-network aggregation

## Effort
Medium

## Artifact
Round 106 report: https://gist.github.com/gomesalexandre/da7e8565d54016e1de5cf2da1db2c69d
Campaign index artifact: https://claude.ai/code/artifact/be2deadf-6195-4dfa-a910-b269b9af7a84

gomesalexandre:
Closes vultisig/vultisig-sdk#1765

Added `fetchStakekitBalancesBatch`, `fetchAllStakekitBalances`, and `chunkStakekitBalanceQueries` to `packages/sdk/src/tools/defi/stakekit/stakekitApi.ts`, ported from `vultiagent-app/src/features/dashboard/lib/yieldXyzPositions.ts` (read-only reference at `/home/gomes/Sites/vultiagent-app`, present on this machine) — the app's proven, shipped positions-fetch layer. Ported only the fetch/chunk/aggregation logic; deliberately left the file's dashboard-presentation helpers (`sumAmount`, `sumActive`, `sumRewards`, `dropEmptyYields`, `formatYieldProtocolLabel`, `mapChainToYieldNetwork`) in the app — those are UI-layer concerns, not part of a "batched balances helper," and the issue's own evidence citations (lines 4-20, 233-255, 380-386) point at the fetch/chunk sections specifically.

**Important host discovery**: the app's proven client targets `https://api.yield.xyz/v1/yields/balances` with a `{queries: [{network, address}]}` body shape. The SDK's *existing* `getBalances` targets a *different* host, `STAKEKIT_API_BASE = 'https://api.stakek.it/v1'`, with a *different* body shape (`[{addresses:{address}, integrationId}]`, requiring a prior `searchYields` call per network to enumerate yield IDs). I confirmed by reading both that these are genuinely different validated contracts on the same URL path, not two options for the same endpoint — so I added a separate `YIELD_XYZ_API_BASE` constant and pointed the new functions at it, rather than reusing (and likely breaking against) `STAKEKIT_API_BASE`. This was a real risk I could only catch by reading both implementations closely, since I have no network access in this sandbox to verify live.

Exported `StakekitBalanceQuery`, `StakekitBalanceEntry`, `StakekitBalanceItem`, `StakekitBalancesResult`, `STAKEKIT_BALANCE_QUERIES_PER_REQUEST` types/constant alongside the three functions, through the full chain: `stakekit/index.ts` → `tools/index.ts` → top-level `packages/sdk/src/index.ts` (alphabetically placed in each), matching how `stakekitBalances` is already exposed. Did not touch or modify the existing single-network `getBalances`/`stakekitBalances` — this is purely additive.

Added `packages/sdk/tests/unit/tools/defi/stakekitBatchedBalances.test.ts` (16 tests, mocked `fetch`, no live network calls): chunk-boundary tests (empty, under-cap, exactly-at-25-cap, custom size, invalid-size guard), request-shape pinning (URL, `X-API-KEY` header, `{queries: [...]}` body — the exact shape the app's own tests pin, since a future refactor renaming the key or targeting the wrong host would silently 400 in production), non-200/cap-exceeded/null-normalization error handling, and the multi-batch aggregation semantics ported from the app's own suite: partial-batch-failure tolerance with diagnosable chunk-context in `errors[]`, and the "throw only when every batch fails" behavior (so a caller doesn't cache an empty result as a false-positive success — this was itself a real fix in the app's PR #733 history, worth preserving).

## what I ran

```
cd /tmp/claude-1000/wt-sdk-w5
npx tsc --noEmit -p packages/sdk/tsconfig.json
# => exit 0, no errors

cd packages/sdk
npx vitest run tests/unit/tools/defi/stakekitBatchedBalances.test.ts tests/unit/tools/defi/stakekit.test.ts
# => Test Files 2 passed (2); Tests 34 passed (34)  [16 new + 18 existing, zero regressions]
```

closes vultisig/vultisig-sdk#1765


NeOMakinG (approved):
LGTM

gomesalexandre:
Pushed `44a24a41a6edf9c2dc4b557f2bb6f62f46546205` to fix the failing lint gate.

What changed:
- sorted the `packages/sdk/src/tools/index.ts` barrel exports so `simple-import-sort/exports` passes

Receipt:
- `/Users/gomes/Sites/vultisig-sdk/node_modules/.bin/eslint --config /Users/gomes/Sites/vultisig-sdk/.config/eslint.config.mjs packages/sdk/src/tools/index.ts`
- `git diff --check`

Needs a fresh non-self current-head review once CI settles.

rcoderdev (approved):

rcoderdev (approved):

rcoderdev (approved):
Re-approved on the current head after the CI-only import-order repair. The sole new change moves the EvmTxNumberish type export into simple-import-sort order; focused ESLint and diff checks pass. CI remains the merge gate.

rcoderdev (approved):

PR #2105, opened 2026-08-18
feat(sdk): add branded ChainId/AssetId (CAIP-2) module

### LINKED ISSUE #1350: chain-identity: branded ChainId/AssetId module computed from the Chain enum
Parent epic: #1347

## Problem
Chain identity is represented at least eight different ways across the SDK/app/abts repos (SDK `Chain` enum, WalletCore `CoinType`, THORChain native-swap short codes, `normalizeChain` alias resolver, the app's own `Chain`/`chainRegistry`, the app's `CHAIN_SYMBOLS`, four backend alias tables, and `CoinKey`). There is a real collision: Terra and TerraClassic both map to denom `uluna` and share the `terra` bech32 HRP (`packages/core/chain/chains/cosmos/cosmosFeeCoinDenom.ts:10-11`), which the backend currently disambiguates with a live network probe (real 2026-07-04 prod incident). See also `packages/core/chain/Chain.ts` for the base enum this should derive from.

## Fix
Introduce a branded `ChainId`/`AssetId` type computed FROM the existing `Chain` enum (CAIP-2/19-derived, one frozen validity matrix), not a full CAIP-2 re-keying. The `Chain` enum stays load-bearing for `CoinType` derivation and the dozens of `Record<Chain, T>` exhaustiveness checks; re-keying by CAIP-2 string would throw that away. This module resolves the Terra/TerraClassic collision statically (`cosmos:phoenix-1` vs `cosmos:columbus-5`) so the live network probe becomes unnecessary.

## Acceptance
- `ChainId`/`AssetId` branded types exported from the SDK, derived from `Chain`, with a single validity matrix covering every chain in `Chain.ts`.
- Terra vs TerraClassic resolve to distinct `ChainId`s without a network call.
- Regression: test every chain/family affected by this change - the full `Chain` enum matrix, with explicit assertions on the Terra/TerraClassic and any other same-denom collision pairs.

This is the foundation for app#PR-8(app) and abts#PR-8(abts), which retire their duplicated chain-identity tables onto this module.

gomesalexandre:
Closes vultisig/vultisig-sdk#1350
Added a branded `ChainId` / `AssetId` module in `packages/core/chain/chainIdentity.ts`, derived from the existing `Chain` enum (the enum stays load-bearing). EVM and Cosmos CAIP-2 ids come from the existing `getEvmChainId` / `getCosmosChainId` tables, so Terra is `cosmos:phoenix-1` and TerraClassic is `cosmos:columbus-5` with no network probe. Remaining families use a frozen unique-id table (BCH cannot reuse Bitcoin's bip122 genesis). Re-exported from the root and RN SDK surfaces. Not itself a signer. Did not migrate app/abts callers onto this module.

## what I ran
```
cd /tmp/claude-1000/wt-sdk-w1 && yarn vitest run --config .config/vitest.core.config.ts packages/core/chain/chainIdentity.test.ts
cd packages/sdk && yarn vitest run --config tests/unit/vitest.config.ts tests/unit/utils/publicExports.test.ts
```
```
 Test Files  1 passed (1)
      Tests  5 passed (5)
 Test Files  1 passed (1)
      Tests  55 passed (55)
```

closes vultisig/vultisig-sdk#1350

### DIFF
--- a/.changeset/sdkg1-1350.md
+++ b/.changeset/sdkg1-1350.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/sdk': patch
+---
+
+Add a branded `ChainId`/`AssetId` module (CAIP-2 for EVM/Cosmos, frozen unique-id table for the rest), re-exported from the root and React Native entrypoints. Additive only — the existing `Chain` enum stays load-bearing.
--- a/packages/core/chain/chainIdentity.test.ts
+++ b/packages/core/chain/chainIdentity.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from 'vitest'
+
+import { Chain } from './Chain'
+import { chainFromChainId, chainIds, isChainId, toChainId, toNativeAssetId } from './chainIdentity'
+
+describe('chainIdentity', () => {
+  it('covers every Chain with a unique CAIP-2 id', () => {
+    const chains = Object.values(Chain)
+    const ids = chains.map(toChainId)
+
+    expect(ids).toHaveLength(chains.length)
+    expect(new Set(ids).size).toBe(chains.length)
+    expect(Object.keys(chainIds).sort()).toEqual([...chains].sort())
+  })
+
+  it('disambiguates Terra vs TerraClassic without a network probe', () => {
+    expect(toChainId(Chain.Terra)).toBe('cosmos:phoenix-1')
+    expect(toChainId(Chain.TerraClassic)).toBe('cosmos:columbus-5')
+    expect(toChainId(Chain.Terra)).not.toBe(toChainId(Chain.TerraClassic))
+    expect(chainFromChainId('cosmos:phoenix-1')).toBe(Chain.Terra)
+    expect(chainFromChainId('cosmos:columbus-5')).toBe(Chain.TerraClassic)
+  })
+
+  it('gives Terra and TerraClassic distinct native asset ids (same uluna denom, different ChainId)', () => {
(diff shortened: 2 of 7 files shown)


rcoderdev (commented):
`chainIdentity.ts` maps TON mainnet to `ton:-239`, but the Chain Agnostic TVM CAIP-2 profile defines the identifier as `tvm:-239`. That makes `toChainId(Chain.Ton)` emit a noncanonical durable ID and `chainFromChainId("tvm:-239")` reject the canonical value. Please switch the mapping and add an explicit canonical round-trip assertion. Reference: https://namespaces.chainagnostic.org/tvm/caip2

rcoderdev (commented):
Two CAIP contract gaps remain on this head. `bitcoincash:mainnet` and `bittensor:finney` violate the CAIP-2 namespace length limit, and the published BCH profile defines `bip122:000000000000000000651ef99cb9fcbe`. `toTokenAssetId` also brands any raw token reference, so empty values or characters such as `/` and `:` produce invalid CAIP-19 identifiers. Please use syntax-valid canonical chain IDs and validate or percent-encode token references, with conformance tests. References: https://standards.chainagnostic.org/CAIPs/caip-2, https://namespaces.chainagnostic.org/bip122/caip2, https://standards.chainagnostic.org/CAIPs/caip-19

rcoderdev:
Closing this because This version publishes noncanonical chain IDs and unvalidated asset identifiers as CAIP-compatible public types. Safely shipping it requires a verified chain matrix and reworked asset construction with conformance tests. The identity requirement remains useful, but this implementation needs a fresh, focused follow-up.

PR #2108, opened 2026-08-18
fix(security): blockaid native-leg drop is fee-magnitude-bound, decline instead of read positionally

### LINKED ISSUE #2091: [BUG] Blockaid simulation: the three-diff native-asset filter can delete a principal leg and reverse the approval direction
**Describe the bug**

`parseBlockaidSolanaSimulation` drops a native-SOL asset diff purely because the response happens to contain exactly three diffs, on the assumption that it is the transaction fee:

```ts
// packages/core/chain/security/blockaid/tx/simulation/api/core.ts:257-263
if (assetDiffs.length === 3) {
  const nativeSolIndex = assetDiffs.findIndex(diff => diff.asset.type === 'SOL' || diff.asset_type === 'SOL')
  if (nativeSolIndex !== -1) {
    relevantDiffs = assetDiffs.filter((_, index) => index !== nativeSolIndex)
  }
}
```

Nothing establishes that the removed leg is fee-sized, or even that it is an outflow. When the native leg is the transaction's principal movement, the parser deletes it and reports whatever remains.

**Counterexample shape**

A wrapped-SOL withdrawal returning three diffs: `native SOL in` + `WSOL residual out` + `receipt token out`.

1. The native SOL **inflow** is filtered away by the rule above.
2. The two remaining diffs both carry only `out`, so in the `relevantDiffs.length > 1` branch (line 283) `inValue` resolves to `undefined`.
3. The `outAsset && inAsset && outValue && inValue` guard therefore fails, and it falls through to `} else if (outAsset && outValue) {` (line 316), returning a `transfer` of the **WSOL residual**.

The approval screen then states money *leaving* on a transaction that nets to SOL *received*. That is a reversed direction, not merely a wrong magnitude — on a transaction-signing screen.

The same heuristic and the same exposure exist in `parseBlockaidSuiSimulation` for native SUI (lines 105-119).

**Secondary, same function, lower confidence**

The multi-diff path destructures positionally over the first two diffs only, so on a response with more than two relevant diffs any additional leg goes unread. The `else if (outAsset && outValue)` fallback reports the first outgoing asset as the entire transfer, which can hide a second spend on the approval screen.

**Expected behavior**

A leg should be excluded only on evidence that it is a fee — e.g. a magnitude bounded by a plausible fee ceiling — rather than on array length alone. Failing that, the parser should decline and let the caller fall back to a generic title, rather than state an authoritative amount computed from part of the transaction. "No headline" is strictly safer than "confidently wrong headline" in this surface.

**Affected consumers**

All three clients mirror this function, so a fix belongs here rather than in any one of them:

- `vultisig-ios` — `BlockaidSimulationParser.swift:119` (carries a "Mirrors `parseBlockaidSolanaSimulation`" comment)
- `vultisig-android` — `BlockaidSimulationParser.kt:79` (same mirror comment)
- `vultisig-windows` — consumes `@vultisig/core-chain/security/blockaid/tx/simulation/api/core` directly

Fixing a single client would diverge approval-screen semantics across platforms.

**What is NOT established**

This is derived from reading the code, not from a captured Blockaid response exhibiting the shape. Before treating it as user-facing, it is worth confirming against real payloads whether a three-diff response with a principal native leg actually occurs in practice. If it does not, the filter is a safe simplification for the shapes Blockaid really returns — in which case this can be closed as wontfix, but the assumption should be written down at the call site rather than left implicit in a comment that says "likely".

**Provenance**

Surfaced by a cross-model review pass during vultisig/vultisig-ios#5190, which fixed an unrelated same-mint netting bug in the iOS mirror. This is **not** a regression from that PR — the behavior predates it and is identical on all three platforms.

gomesalexandre:
Closes vultisig/vultisig-sdk#2091

The Solana and Sui Blockaid simulation parsers dropped a native SOL/SUI leg whenever the response had exactly three diffs, on the assumption it was a fee. A wrapped-SOL withdrawal (native inflow + residual out + receipt) then lost the principal inflow and headlined the residual as a send - reversed direction on an approval screen. Native legs are now dropped only when they are out-only and fee-sized (100k lamports / 0.05 SUI); anything else stays, and leftover 3+ diff shapes decline instead of being read positionally. Blast radius: approval-screen headlines only, not signing/broadcast. Fail-closed: can only refuse a headline more often, never invent a new confident amount. Could not confirm against a captured live Blockaid payload of this shape (issue notes that gap); the regression is bound with fixtures of that shape.

## what I ran
```
cd /tmp/claude-1000/wt-sdk-w2 && yarn exec vitest run packages/core/chain/security/blockaid/tx/simulation/api/core.test.ts --reporter=dot
# Test Files  1 passed (1)
#      Tests  19 passed (19)

cd /tmp/claude-1000/wt-sdk-w2 && yarn exec tsc --noEmit --pretty false -p .config/tsconfig.shared-publish.json
# exit 0
```

closes vultisig/vultisig-sdk#2091

### DIFF
--- a/.changeset/sdkg2-2091.md
+++ b/.changeset/sdkg2-2091.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Stop dropping a Blockaid native SOL/SUI simulation leg just because the response has three diffs. Only fee-sized native outflows are filtered; a principal native inflow no longer reverses the approval headline, and leftover multi-leg shapes decline instead of guessing.
--- a/packages/core/chain/security/blockaid/tx/simulation/api/core.test.ts
+++ b/packages/core/chain/security/blockaid/tx/simulation/api/core.test.ts
@@ -1,7 +1,14 @@
 import { EvmChain } from '@vultisig/core-chain/Chain'
 import { describe, expect, it } from 'vitest'
 
-import { BlockaidEVMSimulation, parseBlockaidEvmSimulation } from './core'
+import {
+  BlockaidEVMSimulation,
+  BlockaidSolanaSimulation,
+  BlockaidSuiSimulation,
+  parseBlockaidEvmSimulation,
+  parseBlockaidSolanaSimulation,
+  parseBlockaidSuiSimulation,
+} from './core'
 
 type AssetDiff = BlockaidEVMSimulation['account_summary']['assets_diffs'][number]
 type Asset = AssetDiff['asset']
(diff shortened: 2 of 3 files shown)


rcoderdev (commented):
`isFeeSizedNativeOutflow` treats every positive out-only native diff below an absolute ceiling as gas. A three-diff transaction containing a real small native spend plus token out/in can therefore drop the native principal and be presented as a token-only swap. Without an authoritative fee signal, magnitude alone is not enough for this approval-security boundary. Please discard only a verifiably exact fee leg; otherwise decline the ambiguous shape, with a below-ceiling native-principal regression test.

rcoderdev:
Closing this because This mixed Solana/Sui patch cannot safely merge: main already replaced its Solana half, while its Sui fee ceiling can still discard a small real native outflow from the signing summary. Conflict and failed checks add integration work. We are closing this version; the current Sui parser still needs a focused fail-closed fix and regression test.

PR #2110, opened 2026-08-18
chore(ci): clarify quality:audit failure mode on a stale branch

### LINKED ISSUE #1807: quality:audit fails absolutely, so every branch older than a lockfile patch is red with no signal that merging main fixes it
## tl;dr

`yarn quality:audit` fails absolutely rather than relative to `main`, so the moment a lockfile patch lands on `main`, **every open branch cut before it goes red on Lint and Type Check** and stays red until someone happens to merge main into it. The failure names a package the PR never touched, so it reads like the PR broke something.

This is the aftermath half of #1710. That issue covered the moment an advisory publishes and `main` itself goes red. Its three advisories are now fixed on main and I am closing it. This one is about what happens next, which nothing currently handles.

## observed, twice, on different advisories

**js-yaml** (`>=4.0.0 <4.3.1`, high, reached via `@eslint/eslintrc@npm:3.3.5`):

```
stale branches:  "js-yaml@npm:^4.1.0, ^4.1.1, ^4.3.0"  -> 4.3.0   red
origin/main:     "js-yaml@npm:^4.3.1"                   -> 4.3.1   green
```

Four PRs of mine (#1621, #1624, #1618, and #1627 behind a masking `format:check`) sat red on this. None of them touch yaml, eslint, or the lockfile. All four went green on a plain `git merge origin/main` with zero authored change.

**nanoid** (`<3.3.17`, high, GHSA-2v37-7h3g-55p8, reached via `postcss@npm:8.5.22`):

```
#1563 old head:  "nanoid@npm:^3.3.16" -> 3.3.16   red
origin/main:     "nanoid@npm:^3.3.17" -> 3.3.17   green
```

Same story, same one-line remedy, on a PR whose entire diff is an ibc route table entry plus a test.

## why it is worse than it sounds

Three things compound:

1. **The failure is attributed to the wrong change.** The job is called `Lint and Type Check`. A reviewer seeing it red on a PR reasonably concludes the PR has a lint or type problem. Nothing in the job name or the summary says "your lockfile is old".

2. **It is silent and open-ended.** No bot comments, no auto-merge of main, no staleness label. A branch can sit red for weeks over a dependency it does not reference. The four PRs above were all approved and simply parked.

3. **`format:check` masks it.** `check:ci` is an 18-command `&&` chain and `quality:audit` is step 17:

   ```
   check:shared-exports, build:all, cli:build, cli:smoke, mcp:build, mcp:smoke,
   changeset:check-sdk-bundle, quality:contracts, lint, typecheck, knip,
   format:check, quality:architecture, quality:duplication, quality:complexity,
   quality:secrets, quality:audit, quality:docs:markdown
   ```

   So a branch can be red for two unrelated reasons and only show you the first. #1627 and #1621 failed the same named check at the same step number for completely different causes, which is exactly the sort of thing that makes people generalise a diagnosis across a batch and get it wrong.

## what I am not claiming

I am not arguing the audit gate should go away. Blocking on a HIGH advisory is correct, and when `main` is the thing that is vulnerable, #1710's behaviour is what you want.

The complaint is narrower: once main is patched, a branch that merely predates the patch is not vulnerable in any way that matters, because it can only ever land through a merge with main. Failing it absolutely tests the age of the branch, not the safety of the change.

## options

Rough, ordered by how much they cost:

- **Diff against main's baseline.** Run the audit on the merge result, or subtract the advisory set that main itself already carries, and fail only on advisories the branch introduces. This is the one that actually matches the intent of the gate.
- **Say so in the failure.** Cheapest by far. When the audit fails, check whether the same advisory is absent on main, and if so print `this advisory is already fixed on main - merge main into your branch`. Turns a mystery into a one-liner.
- **Auto-merge main on advisory-clearing commits**, or label affected PRs `needs-main-merge`. More machinery, but it closes the loop without anyone reading CI logs.
- **Move `quality:audit` earlier in `check:ci`**, or split it into its own job so it stops being masked by, and stops masking, unrelated steps. Orthogonal to the above and worth doing regardless.

## repro

```bash
git checkout <any branch cut before the patch commit>
yarn quality:audit     # fails on a package the branch never touched
git merge origin/main
yarn quality:audit     # No audit suggestions
```

Happy to take the "say so in the failure" option if there is appetite for it, it is the smallest change with most of the benefit.

gomesalexandre:
Closes vultisig/vultisig-sdk#1807

`quality:audit` still fails absolutely against the advisory db (correct when main itself is vulnerable). The gap was the failure mode: a branch cut before a lockfile patch stayed red on "Lint and Type Check" with no signal that merging main is the whole fix. Wrapped the existing `yarn npm audit` command in `scripts/quality-audit.mjs` so a non-zero audit still fails, but prints that this is an absolute lockfile check and to merge main if origin/main already cleared the advisory. Blast radius: CI messaging only; the gate still refuses on HIGH advisories. Did not implement a real audit-set diff against origin/main (needs live registry + checking out main's lockfile).

## what I ran
```
cd /tmp/claude-1000/wt-sdk-w2 && node --test scripts/quality-audit.test.mjs
# ✔ hint tells a stale branch to merge main instead of chasing lint
# ✔ passing audit exits 0 and does not print the stale-branch hint
# ✔ failing audit still fails, but names the merge-main remedy
# ℹ tests 3
# ℹ pass 3
```

closes vultisig/vultisig-sdk#1807

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "knip": "knip --config .config/knip.jsonc",
     "knip:fix": "knip --config .config/knip.jsonc --fix --allow-remove-files",
     "quality:architecture": "depcruise --config .config/dependency-cruiser.cjs --output-type err .",
-    "quality:audit": "yarn npm audit --recursive --all --severity high",
+    "quality:audit": "node scripts/quality-audit.mjs",
     "quality:complexity": "node scripts/run-eslint.mjs --config .config/eslint-health.config.mjs 'packages/sdk/**/*.{js,jsx,ts,tsx}' 'packages/rujira/**/*.{js,jsx,ts,tsx}' 'packages/client-shared/**/*.{js,jsx,ts,tsx}' 'packages/mpc-types/src/**/*.ts' 'packages/core/**/*.ts' 'packages/lib/utils/**/*.{js,jsx,ts,tsx}' 'packages/mpc-native/**/*.{js,jsx,ts,tsx}' 'packages/walletcore-native/**/*.{js,jsx,ts,tsx}' 'clients/**/*.{js,jsx,ts,tsx}' 'examples/**/*.{js,jsx,ts,tsx}'",
     "quality:docs": "yarn quality:docs:markdown && yarn quality:docs:links",
     "quality:docs:links": "git ls-files '*.md' ':!:docs/api/**' ':!:tasks/**' ':!:packages/**/CHANGELOG.md' ':!:clients/**/CHANGELOG.md' | xargs -n 1 markdown-link-check -q -c .config/markdown-link-check.json",
--- a/scripts/quality-audit.mjs
+++ b/scripts/quality-audit.mjs
@@ -1,0 +1,29 @@
+#!/usr/bin/env node
+import { spawnSync } from 'node:child_process'
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+export const AUDIT_ARGS = ['npm', 'audit', '--recursive', '--all', '--severity', 'high']
+
+export const STALE_LOCKFILE_HINT = [
+  'quality:audit failed on this lockfile (absolute, not vs origin/main).',
+  'If this advisory is already fixed on origin/main, merge main into your branch - that is the whole fix.',
+  'This is not a lint or typecheck failure in your change.',
+].join('\n')
+
+export function runQualityAudit({
+  spawn = spawnSync,
+  write = text => console.error(text),
+  exit = code => process.exit(code),
+} = {}) {
+  const result = spawn('yarn', AUDIT_ARGS, { stdio: 'inherit' })
+  const status = typeof result.status === 'number' ? result.status : 1
+  if (status !== 0) {
+    write(`\n${STALE_LOCKFILE_HINT}\n`)
+  }
+  exit(status)
+}
+
+if (process.argv[1] && path.resolve(process.argv[1]) === fileURLToPath(import.meta.url)) {
+  runQualityAudit()
+}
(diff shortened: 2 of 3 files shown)


gomesalexandre (commented):
reviewed current head `f42883f88af91874913729c6792ac1c3b8eb24ff` remotely. no blocker surfaced on the current diff.

grounding i checked:
- `package.json` only swaps `quality:audit` from the inline `yarn npm audit --recursive --all --severity high` command to `node scripts/quality-audit.mjs`.
- `scripts/quality-audit.mjs` preserves the same audit arguments and preserves the non-zero exit code, so the gate still fails on high advisories instead of turning into a warning-only path.
- the new wrapper only adds a stale-branch hint when audit fails, which matches the linked issue intent: explain that this check is absolute against the lockfile and that merging `main` is the fix when `main` already cleared the advisory.
- `scripts/quality-audit.test.mjs` characterizes the three important edges: the hint text itself, the clean path not printing on success, and the failure path still exiting non-zero.
- current-head checks are green: lint/type, unit, integration, summary, and coderabbit all passed.

non-blocking note: the helper still prints the same stale-branch hint for any `yarn npm audit` failure, not just the specific "main already fixed it" case. that is fine for now because the script body is intentionally the cheap messaging-only fix described in `#1807`, but if false-positive operator confusion shows up later the next step is comparing against `main` or the merge result rather than broadening the prose here.

rcoderdev (approved):

rcoderdev:
Integrated main and kept its audit runner, including transient retries and the Windows command selection. The stale-lockfile hint now appears only after a final failure, with the original nonzero exit code preserved. The hint remains conditional on main already containing the advisory fix. CLI checks cover success, advisory failure and recovery after a transient error.

PR #2113, opened 2026-08-18
feat(ripple): read delivered_amount instead of never confirming actual delivered value

### LINKED ISSUE #982: spike(ripple): verify received/history value reads delivered_amount, not Amount (partial-payment safety)
## What (spike / investigation)
XRPL **partial payments** (`tfPartialPayment`) deliver less than the `Amount` field. The rule: when reading a *received/confirmed* value, use the metadata `delivered_amount`, **never** the `Amount` field — reading `Amount` is the classic over-credit exploit.

Our XRP **sends** are Payment-only without the partial flag (fine). This spike verifies the **read side**: tx-history / received-value / confirmation display across SDK, iOS, Android, Windows.

## Investigate
- [ ] SDK: does tx-status/history read `delivered_amount` or `Amount`?
- [ ] iOS: same
- [ ] Android: same
- [ ] Windows: same (shared SDK path)
- [ ] Confirm the `Flags`/tfPartialPayment case is handled when displaying value

## Outcome
If any surface reads `Amount`, file a fix issue on that platform. If all read `delivered_amount`, close with the evidence.

Ref: XRPL partial payments https://xrpl.org/docs/concepts/payment-types/partial-payments · Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

gomesalexandre:
Closes vultisig/vultisig-sdk#982

The spike asked whether XRPL received/history reads `delivered_amount` or `Amount`. The SDK status resolver previously read neither (only tesSUCCESS + fee), so it was not over-crediting — but it also could not confirm delivered value. Added `readRippleDeliveredAmount` that prefers `delivered_amount` / legacy `DeliveredAmount` and never falls back to `Amount`, and attached the result as optional `rippleDelivered` on tx status.

Blast radius: additive optional field on `TxStatusResult`. No existing status consumer is required to honor it. Confirmation callers that need a delivered value must fail closed on absence rather than substitute `Amount`. iOS/Android history surfaces are not in this tree and were not verified. No live XRPL `tx` RPC was queried.

## what I ran
```
cd /tmp/claude-1000/wt-sdk-w1 && yarn test:core packages/core/chain/chains/ripple/deliveredAmount.test.ts packages/core/chain/tx/status/resolvers/ripple.test.ts
```
```
 Test Files  2 passed (2)
      Tests  17 passed (17)
   Duration  178ms
```

closes vultisig/vultisig-sdk#982

### DIFF
--- a/.changeset/sdkh1-982.md
+++ b/.changeset/sdkh1-982.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Surface XRPL Payment `delivered_amount` on tx status and never fall back to the send-side `Amount` field (partial-payment over-credit class).
--- a/packages/core/chain/chains/ripple/deliveredAmount.test.ts
+++ b/packages/core/chain/chains/ripple/deliveredAmount.test.ts
@@ -1,0 +1,66 @@
+import { describe, expect, it } from 'vitest'
+
+import { readRippleDeliveredAmount } from './deliveredAmount'
+
+describe('readRippleDeliveredAmount', () => {
+  it('reads native XRP drops from delivered_amount', () => {
+    expect(readRippleDeliveredAmount({ delivered_amount: '1000000', TransactionResult: 'tesSUCCESS' })).toEqual({
+      type: 'xrp',
+      drops: 1_000_000n,
+    })
+  })
+
+  it('reads the legacy DeliveredAmount alias when delivered_amount is absent', () => {
+    expect(readRippleDeliveredAmount({ DeliveredAmount: '42' })).toEqual({
+      type: 'xrp',
+      drops: 42n,
+    })
+  })
+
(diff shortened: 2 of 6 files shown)


rcoderdev (commented):
The new parser currently rejects valid MPToken delivered amounts. In the locked xrpl@5.0.0 metadata types, delivered_amount / DeliveredAmount may be an MPTAmount with { mpt_issuance_id, value }, but parseDeliveredAmountField accepts only XRP strings and issued-currency objects. A successful MPToken payment will therefore surface no rippleDelivered value. Please add an MPT variant and focused coverage so this API does not silently treat a valid delivered amount as malformed.

rcoderdev:
Closing this because Superseded by #2316, which already exposes validated XRP, issued-currency and MPT delivery in transaction receipts with stronger validation. This would add a competing, less complete delivery field. The remaining cross-platform investigation in #982 is still separate.

PR #2115, opened 2026-08-18
test(chains): pin guard-shaped map completeness across cosmos/evm chains

### LINKED ISSUE #1228: refactor: partial per-chain maps silently skip new chains - promote guard-shaped ones to exhaustive records
The exhaustive `Record<Chain, T>` tables in core are the repo's best drift defense - tsc fails the build when a chain is added without an entry (`coin/chainFeeCoin.ts:14`, `utils/getBlockExplorerUrl/index.ts:15`). But ~39 per-chain maps exist and the `Partial<Record<Chain, T>>` ones silently return undefined for a new chain:

- `packages/core/chain/swap/native/limitSwapMemo.ts:222` - destination validators
- `packages/core/chain/chains/evm/tx/clampEvmPriorityFee.ts:28` - fee ceilings
- `packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.ts:85` - LCD fallback
- `packages/core/chain/chains/cosmos/cosmosMemoCap.ts:20` - memo caps

Several of these are guard-shaped (memo cap, fee clamp, dest validation): a missing entry does not error, it just means the guard never fires for that chain. No gate catches it.

## fix shape

For the guard-shaped maps: promote to full `Record<Chain, T>` with explicit no-op entries (so adding a chain forces a conscious decision), or add a startup/test completeness assertion enumerating intentional gaps. A lint-able convention ("Partial per-chain maps must have a completeness test") would hold the line going forward.

effort: M / risk: LOW-MED / confidence: MED (each map needs a per-map judgment call on full-vs-partial intent)

gomesalexandre:
Closes vultisig/vultisig-sdk#1228

The four named guard-shaped maps are already exhaustive (`Record`) or intentionally Partial. Added completeness assertions that iterate `Object.values(CosmosChain)` / `EvmChain` for memo cap and priority-fee ceiling, pin the fee floor to Ethereum+Polygon only, and enumerate the LCD-fallback gap (MayaChain) so adding a chain is a test failure rather than a silent skip.

Blast radius: test + exporting `cosmosLcdFallbackUrls`. No fee/signing behavior change. Remaining ~35 Partial maps are not guard-shaped (explorer URLs, custom RPC, address maps) and stay Partial by design.

## what I ran
```
cd /tmp/claude-1000/wt-sdk-w1 && yarn test:core packages/core/chain/chains/cosmos/cosmosMemoCap.test.ts packages/core/chain/tx/fee/evm/clampEvmPriorityFee.test.ts packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.test.ts
```
```
 Test Files  3 passed (3)
      Tests  70 passed (70)
```

closes vultisig/vultisig-sdk#1228

### DIFF
--- a/.changeset/sdkh1-1228.md
+++ b/.changeset/sdkh1-1228.md
@@ -1,0 +1,5 @@
+---
+'@vultisig/core-chain': patch
+---
+
+Add completeness assertions for the guard-shaped per-chain maps (memo cap, EVM fee ceiling/floor, cosmos LCD fallbacks) so a new chain cannot silently skip the guard.
--- a/packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.test.ts
+++ b/packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.test.ts
@@ -2,7 +2,7 @@ import { HttpResponseError } from '@vultisig/lib-utils/fetch/HttpResponseError'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { CosmosChain } from '../../../Chain'
-import { getCosmosAccountInfo } from './getCosmosAccountInfo'
+import { cosmosLcdFallbackUrls, getCosmosAccountInfo } from './getCosmosAccountInfo'
 
 vi.mock('@vultisig/core-chain/chains/cosmos/client', () => ({
   getCosmosClient: vi.fn(),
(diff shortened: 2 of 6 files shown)


gomesalexandre (commented):
reviewed current head `edea20f93108fb68bcbe750b6e4b86aaf2ac855a` remotely. no blocker surfaced on the current diff.

grounding i checked:
- the diff stays in the narrow blast radius described in the PR body: completeness assertions plus a small export rename, with no signer, fee, or transport behavior change.
- `packages/core/chain/chains/cosmos/cosmosMemoCap.test.ts` now iterates the full cosmos chain enum instead of a hand-picked subset, so adding a new cosmos chain can no longer silently skip the memo-cap guard.
- `packages/core/chain/tx/fee/evm/clampEvmPriorityFee.test.ts` pins the intended fee-floor/ceiling partition explicitly, which makes new evm chain additions fail the test instead of drifting past the guard.
- `packages/core/chain/chains/cosmos/account/getCosmosAccountInfo.test.ts` and the tiny `getCosmosAccountInfo.ts` / `cosmosRpcUrl.ts` export changes make the lcd-fallback gap explicit instead of leaving it as an uninspected partial map.
- current-head checks are green: lint/type, unit, integration, summary, and coderabbit all passed.

non-blocking note: this is the right guard-shape direction, but the remaining intentionally-partial maps are still easy to cargo-cult later. if another per-chain safety map gets added outside these tests, it is worth asking whether it also wants the same Object.values(...) completeness pin instead of another silent partial.

rcoderdev (commented):
The EVM floor completeness test currently derives only the chains whose computed floor exceeds 1 wei and compares that set with Ethereum and Polygon. Adding a new EvmChain without any floor leaves the derived set unchanged, so the test still passes and the Partial map silently accepts the gap. Please assert an explicit floor or no-floor decision for every EVM chain. Also, cosmosLcdFallbackUrls is exported as a mutable object from a published subpath, so downstream code can mutate the live fallback endpoints used by account lookup. Please keep it internal or export and consume a frozen readonly value.

rcoderdev:
Closing this because The core guard maps are already exhaustive through #1757, while #2325 intentionally changed fee floors and introduced zero-tip chains that these assertions reject. The remaining gap checks need a focused update against current policy rather than this outdated test set.

PR #2117, opened 2026-08-18
fix(utxo): map terminal walletcore plan errors instead of leaking raw error strings

### LINKED ISSUE #1695: UTXO max-send is "plan and pray": `sendMaxAmount` hardcoded false, `plan.error` never inspected, and the repair step is skipped on deposit/agent/null-publicKey paths
Every UTXO max send from desktop/extension first produces a WalletCore `Error_not_enough_utxos` internally, and only a post-hoc retry rescues it. Paths without the retry leak the raw enum to the user.

### Mechanism

The flag is never set from intent:

```ts
// packages/core/mpc/keysign/chainSpecific/resolvers/utxo.ts:29-32
return create(UTXOSpecificSchema, {
  sendMaxAmount: false,          // hardcoded — no max flag exists anywhere in the desktop UI
  byteFee: (await getByteFee()).toString(),
})
```

The UI computes `amount = balance − feeEstimate` (`core/ui/vault/send/amount/ManageAmountInputField.tsx:191-220` in vultisig-windows), so WalletCore is asked for a 2-output plan satisfying `total ≥ (total − fee₁out) + fee₂out` — arithmetically impossible. Confirmed against the shipped wallet-core 4.7.0 wasm (P2WPKH, byteFee 5, dust 546):

```
utxos [1000000]:  amount=total-550, useMax=false → utxos 0, err 10 (Error_not_enough_utxos)
                  amount=total-550, useMax=true  → utxos 1, fee 550, amount 999450, ok
```

The failure band is any amount in `(total − fee₂out − dust, total)` — down to `total − 1` — with one UTXO or many.

`refineKeysignUtxo` (`packages/core/mpc/keysign/refine/utxo.ts:40-109`) is the only safety net: it retries empty plans with `sendMaxAmount: true`, and silently converts near-max sends (`balance − amount ≤ fee + 600n`, unexplained constant) into full sweeps.

### Paths with no safety net

- Deposit / THORChain function calls / bond from BTC — `core/ui/vault/deposit/keysignPayload/build.ts:177,210,244`: builds `utxoInfo` + `getChainSpecific`, never refines. Its 100% slider does float `toFixed` with no fee subtraction (`StakeForm.tsx:90-95`) — lands at exactly `total` (auto-promoted, works) or a hair below (err 10).
- Agent custom tx — `core/ui/agent/tools/handlers/buildCustomTx.ts:92-95`, no refine.
- Null-publicKey sends — `packages/core/mpc/keysign/send/build.ts:124-139`: both refine calls are gated on `if (publicKey)`.

`plan.error` is never inspected (`signingInputs/resolvers/utxo.ts:112-119`); the failure travels to `TransactionCompiler.preImageHashes` and surfaces verbatim via `assertErrorMessage` → the user sees the literal string `Error_not_enough_utxos`.

### Estimate ≠ signed fee

- `useSendFeeEstimateQuery` passes no `feeSettings` while the payload query does (`useSendFeeEstimateQuery.ts:36-48` vs `keysignPayload/query.ts:53`) — "fast" (×2.5) applies only to the signed tx.
- The estimate is `keepPreviousData` + `noRefetch` — during a coin switch it can be the previous coin's fee.
- After the max fallback flips the flag, WalletCore recomputes the amount, but `toAmount` keeps the stale UI value — which is what Verify renders (`SendVerify.tsx:97`). Amount shown ≠ amount signed on every max send.

### Fix direction

- Thread an explicit `isMaxSend` from the amount field → `BuildSendKeysignPayloadInput` → `getUtxoChainSpecific`, so the flag is set before the first plan (keep `refineKeysignUtxo` as backstop only).
- Inspect `plan.error` after `planInput` and map each case to a localized, actionable message.
- Add the flag/refine to the deposit, agent, and null-publicKey paths; subtract a real fee in the deposit slider and drop the float round-trip.
- Render `plan.amount` on Verify after a max plan; notify when a near-max send is converted to a sweep.
- Golden vectors: for N ∈ {1, 2, 100+} UTXOs, assert every amount in `{total, total−1, total−fee±dust}` either plans or fails with a specific error (the wasm probe scripts from this investigation are the template).

gomesalexandre:
Closes vultisig/vultisig-sdk#1695

Inspect WalletCore UTXO `plan.error` in `getUtxoSigningInputs` and map terminal failures to an actionable SDK error instead of leaking `Error_not_enough_utxos` from `preImageHashes`. Empty near-max plans with `sendMaxAmount: false` still reach `refineKeysignUtxo` (throwing there would disable the existing backstop). Thread optional `sendMaxAmount` (default **false**) through `getUtxoChainSpecific` / `buildSendKeysignPayload` / `prepareSendTxFromKeys` so a caller that knows this is a max send can plan a sweep up front. Blast radius: UTXO send planning only; default stays false so no caller starts sweeping more often. Not verified: windows deposit/agent/Verify UI paths (other repo); no funded-vault max send.

## what I ran
```
yarn test:core packages/core/mpc/keysign/signingInputs/resolvers/utxo.test.ts packages/core/mpc/keysign/chainSpecific/resolvers/utxo.test.ts
 Test Files  2 passed (2)
      Tests  12 passed (12)
   Duration  339ms

yarn tsc --project .config/tsconfig.json --noEmit --pretty false
(exit 0)
```

closes vultisig/vultisig-sdk#1695

### DIFF
--- a/.changeset/utxo-plan-error-send-max.md
+++ b/.changeset/utxo-plan-error-send-max.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-mpc': patch
+'@vultisig/sdk': patch
+---
+
+Inspect WalletCore UTXO `plan.error` and map terminal failures to an actionable SDK error instead of leaking `Error_not_enough_utxos`. Thread optional `sendMaxAmount` (default false) through chain-specific + send payload builders so callers can plan a sweep up front. Empty near-max plans still go to `refineKeysignUtxo`.
--- a/packages/core/mpc/keysign/chainSpecific/resolver.ts
+++ b/packages/core/mpc/keysign/chainSpecific/resolver.ts
@@ -14,6 +14,11 @@ export type GetChainSpecificInput<C extends KeysignChainSpecificKey = KeysignCha
   walletCore: WalletCore
   /** XRPL DestinationTag, carried in RippleSpecific for Ripple payments. */
   destinationTag?: number
+  /**
+   * UTXO-only. When true, WalletCore plans a single-output sweep.
+   * Ignored by non-UTXO resolvers. Defaults to false.
+   */
+  sendMaxAmount?: boolean
 } & (C extends 'ethereumSpecific'
   ? {
       feeSettings?: FeeSettings<'evm'>
--- a/packages/core/mpc/keysign/chainSpecific/resolvers/utxo.test.ts
+++ b/packages/core/mpc/keysign/chainSpecific/resolvers/utxo.test.ts
@@ -1,0 +1,49 @@
+import { create } from '@bufbuild/protobuf'
+import { Chain } from '@vultisig/core-chain/Chain'
+import { CoinSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/coin_pb'
+import { KeysignPayloadSchema } from '@vultisig/core-mpc/types/vultisig/keysign/v1/keysign_message_pb'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+vi.mock('@vultisig/core-chain/chains/utxo/fee/byteFee', () => ({
+  getUtxoByteFee: vi.fn(async () => 5n),
+}))
+
+import { getUtxoChainSpecific } from './utxo'
+
+const payload = create(KeysignPayloadSchema, {
+  coin: create(CoinSchema, {
+    chain: Chain.Bitcoin,
+    ticker: 'BTC',
+    address: 'bc1qsource',
+    decimals: 8,
+    isNativeToken: true,
+  }),
+  toAddress: 'bc1qdest',
+  toAmount: '100000',
+})
+
+describe('getUtxoChainSpecific sendMaxAmount', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('defaults sendMaxAmount to false', async () => {
+    const specific = await getUtxoChainSpecific({
+      keysignPayload: payload,
+      walletCore: {} as never,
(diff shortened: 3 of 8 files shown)


gomesalexandre (commented):
blocking on current head `c082ac16176a`.

- `packages/core/mpc/keysign/signingInputs/resolvers/utxo.ts` lets any empty `Error_not_enough_utxos` / `Error_missing_input_utxos` / `Error_low_balance` plan fall through when `sendMaxAmount` is still false. But `packages/core/mpc/keysign/refine/utxo.ts` then flips any nonzero empty-plan request into `sendMaxAmount: true`, so a materially over-balance request can silently degrade into a sweep instead of surfacing the mapped insufficient-funds error this PR is trying to add.
- `packages/sdk/src/vault/services/TransactionBuilder.ts` still does not expose the new `sendMaxAmount` flag on `prepareSendTx(...)`, so the primary SDK transaction-builder surface cannot opt into the new UTXO sweep path that `prepareSendTxFromKeys(...)` now supports.
- current head is also not green: `Lint and Type Check` fails and the PR is behind `main`.

ask
- keep the refine retry bounded to genuine near-max cases, with a regression that proves an over-balance UTXO send stays fail-closed instead of mutating into max-send
- thread `sendMaxAmount` through `TransactionBuilder.prepareSendTx(...)` and pin that public surface too

rcoderdev:
Closing this because This implementation remains incomplete for the fund-sensitive UTXO planning contract: it can still turn an over-balance exact request into a sweep, and current main now requires a deliberate sendMaxAmount contract reconciliation best handled by a fresh bounded fix.

PR #2118, opened 2026-08-18
feat(evm): add opt-in raced-public-rpc broadcast strategy

### LINKED ISSUE #1352: evm: promote raced-public-rpc broadcast (blink protect avoidance) to an opt-in sdk strategy
Parent epic: #1347

## Problem
The app's EVM broadcast path (`vultiagent-poc:src/services/rpc/broadcastEvm.ts`, 466 lines) races `eth_sendRawTransaction` across N public RPCs to dodge Blink Protect's silent-drop behavior. This is deliberate, documented divergence from the SDK's canonical single-endpoint broadcast (`packages/core/chain/tx/broadcast/resolvers/evm.ts:10-28`), not accidental duplication, but it lives entirely app-side so no other consumer benefits from it and the app has to maintain a second broadcast implementation.

## Fix
Promote the raced-public-RPC behavior into the SDK as an opt-in `broadcastStrategy: 'raced-public-rpc'` alongside the existing default single-endpoint strategy. The app switches to calling the SDK with the strategy flag instead of maintaining its own 466-line implementation; abts or other consumers can opt in later if they hit the same silent-drop problem.

## Acceptance
- SDK `broadcast()` accepts an opt-in strategy parameter; default behavior unchanged for existing callers.
- The raced-RPC strategy reproduces the app's current Blink-Protect-avoidance behavior (races N endpoints, first success wins).
- Regression: test every chain/family affected by this change - EVM broadcast only (this strategy is EVM-specific), across at least 3 of the 13 EVM chains, confirming both the default and raced-RPC strategies still resolve to a terminal success/error and neither hangs.

Consumer PR (app) cannot merge until the SDK change is republished to npm and version-bumped with ceremony in this repo (the PR may perform the bump post-publish).

gomesalexandre:
Closes vultisig/vultisig-sdk#1352

Added an opt-in `strategy: 'raced-public-rpc'` on EVM `broadcastTx`. When set, Ethereum races the same public RPCs the app uses and never lets the Vultisig proxy win (Blink Protect can ack then silently drop). Default single-endpoint path is unchanged. Chains without a public list fail closed rather than falling back to the proxy.

Blast radius: opt-in only. Existing `broadcastTx` callers keep the proxy/single-client path. Raced strategy is Ethereum-only today (Base/Arbitrum not configured). Not verified against a live public RPC or a signed mainnet tx. App consumer switch is a follow-up in vultiagent-app after this is published.

## what I ran
```
cd /tmp/claude-1000/wt-sdk-w1 && yarn test:core packages/core/chain/tx/broadcast/resolvers/evm.test.ts packages/core/chain/tx/broadcast/resolvers/evmRacedPublicRpc.test.ts packages/core/chain/tx/broadcast/index.test.ts
```
```
 Test Files  3 passed (3)
      Tests  19 passed (19)
```

```
cd /tmp/claude-1000/wt-sdk-w1/packages/sdk && yarn typecheck
```
exit 0

closes vultisig/vultisig-sdk#1352

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in raced public RPC strategy for broadcasting EVM transactions.
  * The strategy sends transactions to multiple public endpoints and accepts the first successful or already-known response.
  * Added support for selecting the broadcast strategy through the SDK’s transaction broadcasting APIs.
  * Existing single-endpoint broadcasting remains the default.

* **Bug Fixes**
  * Improved handling of failed, malformed, unsupported-chain, and already-known broadcast responses.


rcoderdev (commented):
BroadcastService.execute destructures the correlated BroadcastStrategyInput into chain and strategy, then rebuilds { chain, tx, strategy }. TypeScript can no longer preserve the discriminated-union correlation, and current exact-head CI fails with TS2345 at that call. Please forward the correlated input intact or explicitly narrow the raced branch to Ethereum before dispatch, and add an SDK propagation assertion so this remains type-safe.

rcoderdev (commented):
The raced endpoint list includes https://eth.merkle.io, which Merkle documents as a MEV-protected RPC that intercepts eth_sendRawTransaction into its private mempool (https://www.merkle.io/blog/how-do-private-mempools-work). If that endpoint acknowledges while the ordinary public endpoints reject or time out, Promise.any still reports success without any public-RPC acceptance. That is the protected-acknowledgement case this strategy is intended to avoid. Please exclude protected endpoints from eligible winners and test a protected-only acknowledgement.

rcoderdev:
Closing this because This opt-in broadcast strategy still lets a protected Merkle RPC acknowledgement win a public-RPC race, and its strategy input fails TypeScript checks. Main has changed the same interfaces, while no consumer or signed-transaction QA is shown. We are closing this version; a fresh proposal can fix and verify the behavior against current main.

PR #2124, opened 2026-08-19
feat(solana): Kamino Earn transaction pipeline

### LINKED ISSUE #2059: [ADD] Kamino Earn transaction pipeline — build, attribution memo, blockhash, validate/decode
## Summary

The signing-side half of Kamino Earn in `@vultisig/core-chain`: turn a user intent (deposit / withdraw) into a validated, attributed, fresh Solana v0 transaction ready for ed25519 keysign, plus the structured decode the clients' verify screens consume.

Depends on #2058 (models, registry). Parent: vultisig/vultisig-windows#4541.

## Scope

**Build** — `POST /ktx/kvault/deposit` / `POST /ktx/kvault/withdraw` (body `{ wallet, kvault, amount }`) returning a base64 Solana v0 transaction. Keep the `/ktx/kvault/*-instructions` variants in reserve if self-assembly proves necessary.

**Attribution memo** — append one SPL Memo instruction (program `MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr`) with the literal `8k2mz` to every Kamino transaction — after the API builds it, before validation and signing.

> ⚠️ The parent issue's comment says the tag is `vs`. That was a placeholder — Kamino later specified `8k2mz`, and iOS switched in vultisig/vultisig-ios#5166. The tag is matched whole and compared as bytes (case-sensitive: `8K2MZ` is a different memo) and must be byte-identical across platforms — it is the analytics key for Vultisig deposit-flow attribution. Keep the literal in exactly one place so the injector, validator, and decoder move together.

Injection must preserve address-lookup-table account indexes. The SDK already depends on `@solana/web3.js` (`VersionedTransaction` / `VersionedMessage`), so no hand-rolled wire editor is needed — but the round-trip should be pinned against the iOS golden fixtures (`SolanaV0TransactionGoldenTests`, `KaminoTransactionFixtures`).

**Fresh blockhash at sign time** — the built transaction embeds a ~60–90 s blockhash and the MPC ceremony can outlive it. Fetch the transaction immediately before keysign (same pattern as Jupiter swaps) and/or replace `recentBlockhash` on the decompiled message.

**Transaction validator** — same posture as the existing `assertSafeSolanaSwapInstructions`: admit only expected programs and instruction shapes, refuse everything else.

- kvault deposit, and **both** withdraw discriminators: `withdraw` *and* `withdraw_from_available`. The API builds the latter whenever the amount fits the vault's liquid buffer; iOS's verify screen initially rejected every withdrawal because it only knew the former (fixed in vultisig/vultisig-ios#5191).
- Exactly one memo instruction, data exactly `8k2mz` — any other memo remains a refusal.
- Compute-budget instructions.
- Farm unstake instructions, for withdrawing farm-staked shares (vultisig/vultisig-ios#5057).

**Decoder** — a structured summary (action, vault, token amount, estimated received) for the clients' verify screens. Rendering is out of scope here.

## Acceptance criteria

- [ ] Deposit and withdraw transactions build, carry the `8k2mz` memo, and pass validation
- [ ] Pinned-literal test asserts the tag string and its exact bytes; near-miss memos (prefix, suffix, case, trailing-space variants) are refused
- [ ] Both withdraw instruction shapes validate and decode
- [ ] Memo injection round-trips the iOS golden fixtures with lookup-table indexes preserved
- [ ] Blockhash refresh path covered by tests
- [ ] Changeset bumps `@vultisig/core-chain`

## Reference implementation

iOS: vultisig/vultisig-ios#5053 (validator + fixtures), vultisig/vultisig-ios#5117 (memo injection), vultisig/vultisig-ios#5166 (tag = `8k2mz`), vultisig/vultisig-ios#5191 (`withdraw_from_available`), vultisig/vultisig-ios#5057 (farm-staked shares).

Ehsan-saradar:
Closes #2059. Second SDK slice of the Kamino Earn integration (vultisig/vultisig-windows#4541), building on the data layer from #2074: everything between a user intent and the bytes an ed25519 keysign may cover, under `packages/core/chain/chains/solana/kamino/tx/`.

The build API returns an unsigned v0 transaction from the caller's wallet address alone, so the pipeline treats the response as untrusted end to end: between an API response and a signature over it there is only this module.

## What's in here

- **Build actions** (`actions.ts`) — `POST /ktx/kvault/{deposit,withdraw}` with the data layer's typed amounts (tokens in, shares out, enforced by type), bounds checked before a request leaves the device, and the withdraw-everything sentinel (`u64::MAX`, what the API silently rewrites an over-balance withdraw to) refused outright. Builds are never retried — a POST carries no idempotence guarantee we're entitled to assume about someone else's service.
- **Wire layer** (`wire.ts`) — parsing (v0 only; legacy bytes surface as unreadable), the two injections, and blockhash replacement. The attribution memo (`8k2mz`, pinned in one place with an exact-bytes test) and the compute-budget pair are injected by appending their program as a read-only unsigned static key and shifting lookup-region account indexes by one — `programIdIndex` values untouched since programs are always static — so address-lookup-table resolutions are preserved exactly (tested by resolving every instruction's accounts before and after). The blockhash is replaced immediately before keysign rather than trusted to survive an MPC ceremony.
- **Instruction catalog** (`instructions.ts`) — program allow-list, Anchor discriminators (each pinned by a test that re-derives it from `sha256("global:<name>")`, so the constant asserts the bytes *are* that instruction), account layouts shared by validator and decoder so the two can never drift, the farms `u128`/WAD argument reader kept strictly separate from the `u64` one, and the farms user-state PDA derivation.
- **Sequence template** (`sequence.ts`) — the exact instruction shapes each operation produces, matched in order with fail-closed semantics: both withdraw discriminators read as one withdraw (which one arrives is a fact about the vault's liquid buffer, not the user's request); the farms unstake pair appears together or not at all (an unstake alone strands shares); the wrapped-SOL withdraw's close is required-repeatable; anything the template does not name is a refusal.
- **Validator** (`validate.ts`) — layered, every finding a refusal: shape (unsigned single signer, paid by the owner), lookup tables (the vault's own or none), program allow-list, sequence, account identity against the registry and local ATA/PDA derivations (never the response — that check would be circular), every spend amount pinned (kvault `u64`, farm release held to the *exact* shortfall, `farms::stake` to the whole-balance sentinel, compute-budget pair to the values the app injected), and no unexplained writable accounts — a drain has to write somewhere, and outside the trusted programs there is nowhere left. The vault descriptor must be the registry's **own object**, making the registry structurally the only source of vault identity.
- **Offline decoder** (`decode.ts`) — the independent claim a co-signing device derives from the bytes alone, with no RPC: vault identified through the signer's own derived share account (stronger than reading an address out of the bytes), the whole transaction matched against the template so nothing rides alongside a plausible deposit, offline-readable spends bounded, and the withdraw-everything sentinel surfaced as a flag rather than rendered as 18.4 quintillion shares. What offline cannot pin (lookup-table account identities, the exact farm-release amount) is exactly what the initiating validator closes.

## Testing

132 kamino tests total (74 new for the pipeline): the pinned memo literal and its exact bytes with near-miss refusals (case, prefix, suffix, whitespace); discriminators re-derived from instruction names; ALT index preservation through injection round-trips; both withdraw shapes validating and decoding; farm-staked withdraws with the exact-shortfall pin and the oversized-release refusal; sentinel refusals at build, validate, and decode; pre-injection fee/memo refusals; writable-account rules; blockhash refresh. Full core suite (2,913), tsc, knip, prettier, shared-exports check, and `changeset:check-sdk-bundle` all pass.

Changeset bumps `@vultisig/core-chain` and `@vultisig/sdk` (minor).

## Not in here

Keysign payload wiring, verify-screen rendering, and the client flows — vultisig/vultisig-windows#4668/#4669/#4670 consume this module.

## Summary by CodeRabbit

* **New Features**
  * Added Kamino Earn deposit and withdrawal transaction support on Solana.
  * Added strict transaction decoding and validation for supported vaults, farms, accounts, amounts, and instructions.
  * Added priority-fee, attribution-memo, lookup-table, and recent-blockhash handling.
  * Added offline transaction inspection for independent co-signing verification.
* **Bug Fixes**
  * Withdraw-all requests and malformed, unsupported, or ambiguous transactions are rejected before transmission.
  * Improved handling of Kamino service errors and invalid transaction responses.
* **Tests**
  * Added comprehensive coverage for transaction building, decoding, validation, sequencing, and wire handling.


gomesalexandre (approved):
## lgtm 🫡

Re-grounded on current head `2ec3b22` after the follow-up commit.

Receipts:
- exact-head focused suite green:
  - `./node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/core/chain/chains/solana/kamino/tx/actions.test.ts packages/core/chain/chains/solana/kamino/tx/instructions.test.ts packages/core/chain/chains/solana/kamino/tx/sequence.test.ts packages/core/chain/chains/solana/kamino/tx/wire.test.ts packages/core/chain/chains/solana/kamino/tx/validate.test.ts packages/core/chain/chains/solana/kamino/tx/decode.test.ts`
  - result: `6 passed, 72 tests passed`
- exact-head probe green:
  - `./node_modules/.bin/tsx ./kamino_probe.ts`
  - result: validator accepted the injected deposit flow and decoder recovered `deposit` with `amountBaseUnits: 1000000`
- sanity-checked the new follow-up hardening too:
  - malformed 200 build response now fails as a named API error in `actions.ts`
  - offline decoder now binds wrapped-SOL `syncNative` to the signer-owned ATA in `decode.ts`
  - fixture guards in `txTestKit.ts` fail loudly instead of encoding bogus farmless/farm-user-state cases

I also checked `YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn build:shared`, but that currently fails on both this PR head and pristine `origin/main` with the pre-existing Cardano `auxiliaryData` typing issue in `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts`, so that one is main-side drift rather than Kamino-regression evidence.

gomesalexandre (approved):
re-grounded on current head `4b7ebb6` and re-ran exact-head receipts.

receipts:
- focused exact-head suites green:
  - `./node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/core/chain/chains/solana/kamino/tx/actions.test.ts packages/core/chain/chains/solana/kamino/tx/instructions.test.ts packages/core/chain/chains/solana/kamino/tx/sequence.test.ts packages/core/chain/chains/solana/kamino/tx/wire.test.ts packages/core/chain/chains/solana/kamino/tx/validate.test.ts packages/core/chain/chains/solana/kamino/tx/decode.test.ts`
  - result: `6 files passed, 77 tests passed`
- exact-head runtime probe green:
  - `./node_modules/.bin/tsx ./kamino_pr2124_probe.ts`
  - result: validator accepted the injected deposit flow and decoder recovered `{ operation: "deposit", amountBaseUnits: "1000000" }`

spot-checked the current-head hardening too:
- `packages/core/chain/chains/solana/kamino/tx/decode.ts` now binds the decoded farm stake/init authority to the signer and the vault's own farm
- `packages/core/chain/chains/solana/kamino/tx/validate.ts` now fail-closes ATA derivation failures into typed `derivationFailed` refusals

lgtm 🫡

PR #2126, opened 2026-08-19
fix: reject memo on TRC-20 transfer prep

### LINKED ISSUE #561: [clarify] TRC-20 memo semantic: TriggerSmartContract.data is ABI blob, not free-form memo
Surfaced during R1 review of mcp-ts#218.

\`build_trc20_transfer\` now accepts a \`memo\` arg via mcp-ts#218, mirroring the native TRX send tool for API consistency. However:

- TRC-20 transfers use \`TriggerSmartContract\` whose \`data\` field is the ABI call blob (\`transfer(address,uint256)\` selector + padded args). NOT a free-form memo field.
- If we tried to embed a memo there, the contract call would fail (invalid call data).
- The native TRX \`TransferContract\` HAS a separate \`data\` field (proto 12) for memo. \`TriggerSmartContract\` does NOT.

**Current behavior**: memo arg in \`build_trc20_transfer\` is silently no-op (the signer doesn't know where to put it for contract calls).

**Open question**: how to surface TRC-20 memos? Options:
1. Reject memo in \`build_trc20_transfer\` (return error if provided)
2. Document explicit no-op in tool description
3. Add a separate native-TRX-only memo field on \`TriggerSmartContract\` via off-chain attachment (unusual but some chains do it)

THORChain doesn't currently route TRC-20 via vault-deposit so blast radius is near-zero — but before TRC-20 swap integration ships, this needs a decision.

Tracking issue, not blocking mcp-ts#218 / sdk#559 / app#856.

Note for agents: investigate \`TriggerSmartContract\` proto spec to confirm whether any off-chain memo mechanism exists for TRC-20 transfers. Also check whether any Tron block explorer convention exists for attaching user-visible memos to contract calls.

gomesalexandre:
closes #561

`TriggerSmartContract.data` is the ABI call blob (`transfer(address,uint256)` selector + padded args), not a free-form memo field - unlike native TRX `TransferContract`, which has a dedicated memo `data` field. `prepareTrc20TransferFromKeys` previously silently no-op'd a `memo` argument on TRC-20 calldata descriptors instead of rejecting it. Now rejects any `memo` argument outright, failing before an unsigned descriptor is produced.

Left `buildTrc20TransferTx` unchanged - main already has byte-level tests pinning transaction-level `raw_data.data` for TRC-20; this narrows only the vault-free prep helper.

Fund-adjacent but strictly additive/fail-closed: callers passing a memo now fail closed instead of getting a descriptor that silently drops it. Normal no-memo TRC-20 transfer calldata is unchanged. No funded-vault signing/broadcast performed.

## what I ran
```
yarn workspace @vultisig/sdk vitest run --config tests/unit/vitest.config.ts tests/unit/prepareTrc20TransferFromKeys.test.ts
```
Passes with new memo-rejection regression coverage. Changeset included.

### DIFF
--- a/.changeset/trc20-memo-guard.md
+++ b/.changeset/trc20-memo-guard.md
@@ -1,0 +1,6 @@
+---
+"@vultisig/sdk": patch
+"@vultisig/cli": patch
+---
+
+Reject memo input in the pure TRC-20 calldata prep helper so callers do not mistake descriptor metadata for Tron transaction-level memo bytes.
--- a/packages/sdk/src/tools/prep/trc20.ts
+++ b/packages/sdk/src/tools/prep/trc20.ts
@@ -18,7 +18,11 @@ export type PrepareTrc20TransferFromKeysParams = {
   to: string
   /** Amount in token base units (integer string). */
   amount: string
-  /** Optional memo (THORChain swap memo, exchange deposit memo). Maps to the TRON proto `data` field. */
+  /**
+   * Unsupported for this pure calldata descriptor. TRC-20 `parameter` is only
+   * the ABI payload for `transfer(address,uint256)`; callers that need Tron
+   * transaction-level `raw_data.data` must use the lower-level Tron tx builder.
+   */
   memo?: string
   /**
    * Energy/bandwidth cost ceiling for the TriggerSmartContract, in SUN.
@@ -52,8 +56,6 @@ export type UnsignedTrc20Transfer = {
   feeLimitSun: string
   /** Amount in token base units. */
   amount: string
-  /** Optional memo (maps to proto `data`). */
-  memo?: string
 }
 
 const DEFAULT_FEE_LIMIT_SUN = '100000000' // 100 TRX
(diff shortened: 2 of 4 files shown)


gomesalexandre (commented):
reviewed current head `63187843f82f`.

looks sound on the code path i re-grounded.
- `packages/sdk/src/tools/prep/trc20.ts` now fail-closes `memo` on the pure TRC-20 calldata helper instead of implying Tron transaction-level `data` bytes live inside the ERC-20 transfer descriptor.
- the unit regression in `packages/sdk/tests/unit/prepareTrc20TransferFromKeys.test.ts` now pins the intended contract: passing a memo throws, while the receipt script in `scripts/receipts/prep_trc20_transfer.mjs` keeps the deterministic calldata round-trip and explicitly checks the memo guard too.

not stamping approve.
- this is my own branch.
- current head is still red anyway: `Unit Tests` / `Test Summary` are failing on GitHub, so it still needs a green rerun or follow-up before merge.

rcoderdev:
Closing this because Closing this reject-all version because the SDK already preserves memo metadata and serializes transaction-level memo separately from TRC-20 calldata. Issue #561 remains open to verify caller forwarding and finalize its documentation; this closure does not mark that requirement fixed.

PR #2128, opened 2026-08-19
fix: reject negative values in bigIntToHex/numberToEvenHex

### LINKED ISSUE #1139: encoding: bigIntToHex/numberToEvenHex turn a negative bigint into empty bytes instead of rejecting it

gomesalexandre:
closes #1139

`value.toString(16)` on a negative bigint/number produces a `-...` string, which `Buffer.from(..., "hex")` can silently decode as empty bytes instead of throwing - a fail-open gap on shared pre-signing amount/nonce/ABI byte construction paths used by EVM, Tron, Polkadot, TON, Cardano, and Blockaid payload preparation. Adds fail-closed negative-value guards to both `bigIntToHex` and `numberToEvenHex`.

Only refuses invalid negative values earlier than before - doesn't loosen any signing, dispatch, fee, or broadcast check.

## what I ran
New utility regression tests covering bigint, number, and `Long` inputs. Reran existing TON signing-input coverage that uses `numberToEvenHex`. No funded-vault ceremony needed for a pre-sign encoder guard. Changeset included.

### DIFF
--- a/.changeset/bigint-negative-hex-guard.md
+++ b/.changeset/bigint-negative-hex-guard.md
@@ -1,0 +1,5 @@
+---
+"@vultisig/sdk": patch
+---
+
+Reject negative bigint/number inputs in `bigIntToHex`/`numberToEvenHex` instead of silently producing empty bytes, closing a fail-open gap on shared pre-signing amount/nonce/ABI encoding paths.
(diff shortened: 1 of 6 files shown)


gomesalexandre:
Duplicate of #1144, which already implements this fix and is under active review. Closing to avoid splitting review effort.

PR #2135, opened 2026-08-19
feat(swap): export general swap quote policy helper

gomesalexandre:
closes vultisig/agent-backend-ts#1291

`execute_swap` in agent-backend-ts carries backend-only provider-selection policy for general EVM swaps (default slippage tolerance + burying-provider exclusions for same-chain EVM ERC-20 sells) instead of consuming an SDK-owned quote-policy contract. If app/backend/CLI/SDK consumers each hand-roll this, they drift in route selection and user-visible behavior.

Adds `packages/core/chain/swap/quote/generalSwapQuotePolicy.ts` — a pure, SDK-owned policy contract (`resolveGeneralSwapQuotePolicy`, `isSameChainEvmErc20SellForQuotePolicy`, `generalSwapProviderRiskMetadata`, `DEFAULT_GENERAL_SWAP_QUOTE_SLIPPAGE_TOLERANCE_PERCENT`, `SAME_CHAIN_EVM_ERC20_SELL_EXCLUDED_PROVIDERS`). Preserves caller-supplied slippage/exclusions, adds the default slippage only when omitted, and appends provider exclusions only for same-chain EVM ERC-20 source sells — cross-chain, native-source, and non-EVM swaps are unaffected. Exported through core-chain, `@vultisig/sdk`, and `@vultisig/sdk/tools/swap`.

Pure additive API — no changes to `findSwapQuote` default behavior, quote fetching, transaction construction, signing, allowance checks, or broadcast paths. Fund-adjacent behavior is fail-closed only for consumers that explicitly opt in by calling `resolveGeneralSwapQuotePolicy`.

**Scope note**: this is the SDK-side half only. Migrating `agent-backend-ts`'s `execute_swap.ts` to consume this instead of its local policy assembly is a legitimate follow-up in that repo (cross-repo, not implementable from this worktree) — filed as the honest remaining gap, not silently dropped.

## what I ran
```
yarn vitest run .../generalSwapQuotePolicy.test.ts .../findSwapQuote.slippage.test.ts .../findSwapQuote.selection.test.ts
Test Files  3 passed (3)
Tests  79 passed (79)
```
Full `yarn build:sdk`, typecheck, lint, `check:shared-exports` all green. Verified the built dist actually exports and runs correctly via direct `node --input-type=module` imports of both the ESM and CJS builds, confirming real runtime output (not just type-level correctness). Changeset included.

### DIFF
--- a/.changeset/tl-sdk-1291.md
+++ b/.changeset/tl-sdk-1291.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Add an SDK-owned general swap quote policy helper for backend/chat consumers. The new policy exports the default quote slippage tolerance, the same-chain EVM ERC-20 provider exclusions for route-dependent approval-spender providers, and metadata explaining the provider risk classification.
--- a/packages/core/chain/package.json
+++ b/packages/core/chain/package.json
@@ -1998,6 +1998,11 @@
       "import": "./dist/swap/quote/findSwapQuote.js",
       "default": "./dist/swap/quote/findSwapQuote.js"
     },
+    "./swap/quote/generalSwapQuotePolicy": {
+      "types": "./dist/swap/quote/generalSwapQuotePolicy.d.ts",
+      "import": "./dist/swap/quote/generalSwapQuotePolicy.js",
+      "default": "./dist/swap/quote/generalSwapQuotePolicy.js"
+    },
     "./swap/quote/getSwapQuoteProviderName": {
       "types": "./dist/swap/quote/getSwapQuoteProviderName.d.ts",
       "import": "./dist/swap/quote/getSwapQuoteProviderName.js",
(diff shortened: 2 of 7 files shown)


gomesalexandre (commented):
reviewed current head `992379c17e75`.

looks sound.
- `packages/core/chain/swap/quote/generalSwapQuotePolicy.ts` cleanly centralizes the SDK-owned default slippage tolerance plus the same-chain EVM ERC-20 provider exclusions for route-dependent spender providers.
- the policy shape stays tight: caller exclusions are preserved, policy-owned exclusions are appended once, and the exported risk metadata makes the provider-specific rationale explicit instead of burying it in backend-only heuristics.
- the helper is re-exported on the intended public SDK surfaces in `packages/sdk/src/index.ts` and `packages/sdk/src/tools/swap/*`, with focused tests covering the same-chain/cross-chain/native/non-EVM branches.
- current-head checks are green.

not stamping approve because this is my own branch and still needs a non-gomes review for four-eyes.

rcoderdev:
Closing this because the SDK now exposes and validates route-specific approval spenders for LiFi and SwapKit. Blanket same-chain ERC-20 exclusions would preserve a workaround for that earlier gap. A shared policy should return with a demonstrated remaining SDK requirement and consumer contract.

PR #2164, opened 2026-08-22
fix(solana): route the Kamino API through the Vultisig proxy

Ehsan-saradar:
Routes the Kamino Earn API through the Vultisig proxy, the same way every other third-party API in this package is reached.

## Why

Kamino's API is unusable from a browser for the endpoints that matter:

```
OPTIONS /ktx/kvault/deposit  → 404, no CORS headers        (preflight rejected)
POST    /ktx/kvault/deposit  → 200, no allow-origin header (unreadable by a browser)
```

A POST carrying `Content-Type: application/json` is not a simple request, so the browser preflights first — and even if it didn't, the response carries no `access-control-allow-origin`, so the reply could never be read. Deposits and withdrawals therefore hang on the review screen in any webview.

It went unnoticed because CORS is enforced by the **browser**, not the server: the same call succeeds from `curl`, from a native client, and from a browser extension whose host permissions exempt it. Only an ordinary web origin — the desktop app — is bound by it, and there the failure reads as a hang rather than a rejection.

The proxy answers the preflight and sets the header, so every client behaves the same. That is precisely why `jup`, `coingeicko`, `blockaid` and the chain RPCs already go through it; Kamino was the outlier.

## Change

One constant: `kaminoConfig.apiBaseUrl` becomes `${rootApiUrl}/kamino`. It covers both the reads and the two build endpoints, since they share the base.

The host is pinned by a test rather than left implicit — pointing it back at Kamino would break signing in every webview while every other test in the suite still passed.

## Testing

All six endpoints verified through the proxy, each returning exactly one `access-control-allow-origin`:

- `POST /ktx/kvault/{deposit,withdraw}` — 200, returns a transaction; preflight answered with 204
- `GET /kvaults/vaults/{vault}`, `/metrics`, `/kvaults/users/{owner}/positions`, `/kvaults/users/{owner}/vaults/{vault}/pnl` — 200

Core suite passes (2,332 tests), prettier and the bundled-changeset guard clean. Verified end to end in the desktop app against a locally built copy of this package.

Note: `packages/client-shared/src/auth-setup.ts` reports three pre-existing typecheck errors on `main`, unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Routed Kamino Earn API requests through the Vultisig proxy.
  * Improved browser-based deposit and withdrawal flows by avoiding upstream CORS issues.
  * Centralized API endpoint configuration for more consistent Kamino requests.

* **Tests**
  * Updated API and transaction flow tests to validate the configured proxy endpoint.
  * Added coverage confirming requests use the Vultisig Kamino proxy, including encoded request paths.


gomesalexandre (commented):
reviewed current head `a6b6f6e6a6f63860d9f1ed05a56473846d3b9eec` remotely. no blocking diff findings.

grounding i checked:
- `packages/core/chain/chains/solana/kamino/config.ts` switches the Kamino base URL to `${rootApiUrl}/kamino`, which resolves to `https://api.vultisig.com/kamino` on the current head.
- the request shapes stay byte-for-byte aligned with the old direct Kamino integration: reads still hit `/kvaults/...`, deposit still posts to `/ktx/kvault/deposit`, and withdraw still posts to `/ktx/kvault/withdraw` with the same bodies.
- `packages/core/chain/chains/solana/kamino/api.test.ts` now pins the proxy host explicitly, so a future revert back to `api.kamino.finance` would fail tests instead of silently re-breaking webview CORS.
- `packages/core/chain/package.json` already carries `@vultisig/core-config` as a runtime dependency, so the new `rootApiUrl` import does not introduce a missing package edge.
- current-head CI is green: unit, integration, lint/type, both node-storage jobs, summary, and CodeRabbit all passed.

not stamping approve because this head includes a gomes-authored commit, so it still needs a non-gomes current-head approval for four-eyes before merge.

gomesalexandre (dismissed):
Genuine approve. Only blocker was the four-eyes bug (gomes-authored commit on the branch — a one-line import-sort fix — wrongly treated as self-approve-disqualifying; that rule only fires when the reviewer IS the PR author).

Small, well-scoped fix: `kaminoConfig.apiBaseUrl` moves to `${rootApiUrl}/kamino`, routing Kamino Earn reads/deposit/withdraw through the existing Vultisig proxy the same way `jup`/`coingeicko`/`blockaid` already do, so the browser CORS preflight (404, no `access-control-allow-origin`) that was hanging deposit/withdraw in any webview gets answered. Request shapes are unchanged (same paths, same bodies) — this is purely a base-URL swap, not a behavior change to the Kamino integration itself. The host is pinned by a test so a future revert back to `api.kamino.finance` fails CI instead of silently re-breaking webview signing. CI green across unit/integration/both node-storage jobs/lint-type/CodeRabbit.

rcoderdev (approved):

Issue #2169, opened 2026-08-22
Golden corpus: no signing vectors for Robinhood or BitcoinCash on any platform


realpaaao:
Two supported chains have no case in the shared golden fixture corpus — not here, not in vultisig-android, not in vultisig-ios:

- **Robinhood** — recently added chain; an EVM chain-matrix case (like the existing Blast/Mantle/etc. entries in \`evm-chain-matrix.json\`) would be cheap and closes the gap.
- **BitcoinCash** — only covered by the internal WalletCore-vs-MPC check in \`compileTx.golden.test.ts\`; no cross-platform \`expected_image_hash\` vector exists. A send case in \`utxo.json\` would match how Litecoin/Dogecoin are covered.

Suggest authoring both vectors here first (this repo is the reference implementation and the corpus tooling lives here), then porting the cases to the Android and iOS corpora per the PR-template parity flow.

Issue #2168, opened 2026-08-22
Mobile golden corpus: thorchain memo SignAmino case missing


realpaaao:
\`packages/core/mpc/keysign/tests/fixtures/mobile/thorchain.json\` is missing one case that vultisig-android pins:

- **SignAmino - THORChain transaction memo** (in \`vultisig-android/app/src/androidTest/assets/thorchain.json\`)

It is not part of [#2148 — test(keysign): expand golden signing vectors](https://github.com/vultisig/vultisig-sdk/pull/2148) either, so it can ride along there or land as a follow-up. Copy the case from the Android fixture and update the corpus case count + per-file digest in \`keysignPayloadWire.golden.test.ts\`.

(iOS is also missing this case — tracked in [vultisig-ios#5242](https://github.com/vultisig/vultisig-ios/issues/5242).)

PR #2148, opened 2026-08-21
test(keysign): expand golden signing vectors

rcoderdev:
Closes #1585

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cosmos IBC transfer handling by separating routing information from the user’s memo.
  * Preserved optional user memos, including memos containing colons.
  * Improved validation of IBC source-channel information and transaction details.

* **Tests**
  * Expanded mobile transaction coverage across Cosmos, EVM, UTXO, Bittensor, QBTC, Sei, and THORChain.
  * Added regression checks for IBC memo handling, Zcash transactions, and multi-chain native transfers.

PR #2170, opened 2026-08-22
ci: cross-repo golden corpus sync check

realpaaao:
## What

The mobile golden fixture corpus (\`packages/core/mpc/keysign/tests/fixtures/mobile/\`) is shared with vultisig-android and vultisig-ios, but synced by hand — each repo's tests only assert against its own pinned hashes, so two repos disagreeing on the same case would merge silently on both sides. This adds a small CI job that fails when any case shared between the three repos has different \`expected_image_hash\` values.

## How

- \`scripts/check_golden_corpus_sync.py\` — stdlib-only Python, an identical copy lands in all three repos. Compares \`name → expected_image_hash\` per case, parsed — formatting and field-spelling differences between the copies are ignored.
- \`.github/workflows/golden-corpus-sync.yml\` — runs on PRs touching the corpus, pushes to main, weekly, and on dispatch. Sparse-clones only the two sibling corpus directories, so the job takes seconds.
- **Fails** only on hash disagreement for a shared case. Missing files/cases just **warn** — coverage gaps are tracked in [#2168](https://github.com/vultisig/vultisig-sdk/issues/2168), [#2169](https://github.com/vultisig/vultisig-sdk/issues/2169), [vultisig-android#5673](https://github.com/vultisig/vultisig-android/issues/5673), [vultisig-ios#5242](https://github.com/vultisig/vultisig-ios/issues/5242).
- An intentional change to an existing shared vector fails here until the same hashes land in the sibling repos — that is the intended behavior. New cases and new files only warn, so first-mover corpus additions (like #2148) are unaffected.

CI/tooling only — no package code, so no changeset.

Companion PRs: [vultisig-android#5674](https://github.com/vultisig/vultisig-android/pull/5674), [vultisig-ios#5244](https://github.com/vultisig/vultisig-ios/pull/5244).

## Receipts

Local run against current main of all three repos — exit 0, zero disagreements, 87/87/71 cases loaded, 15 coverage warnings matching the filed issues:

```
android: 27 fixture files, 87 cases
ios: 30 fixture files, 87 cases
sdk: 23 fixture files, 71 cases
15 coverage warning(s) (non-fatal, tracked in issues): ...
OK: every case shared between repos agrees on expected_image_hash.
```

Negative test: corrupting one hash in a local copy → exit 1 naming the case:

```
1 CROSS-REPO HASH DISAGREEMENT(S) — two platforms would sign different bytes:
  DRIFT evm.json :: Send ETH
```

## Summary by CodeRabbit

* **New Features**
  * Added automated checks to verify that mobile golden-signing corpora remain synchronized across Android, iOS, and SDK repositories.
  * Reports missing cases, duplicate cases, unpinned values, ordering differences, and hash mismatches.
  * Provides clear success or failure results, while retaining non-blocking coverage warnings.

* **Chores**
  * Added scheduled, pull request, push, and manual workflow triggers for synchronization checks.
  * Improved validation of repository paths and corpus comparison results.

### DIFF
--- a/.github/workflows/golden-corpus-sync.yml
+++ b/.github/workflows/golden-corpus-sync.yml
@@ -16,6 +16,10 @@ on:
     - cron: "37 6 * * 1"
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
 
--- a/.github/workflows/golden-corpus-sync.yml
+++ b/.github/workflows/golden-corpus-sync.yml
@@ -31,6 +31,8 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v4
+        with:
+          persist-credentials: false
 
       - name: Fetch sibling corpora
         run: |
--- a/.github/workflows/golden-corpus-sync.yml
+++ b/.github/workflows/golden-corpus-sync.yml
@@ -10,27 +10,38 @@ on:
     branches: [main]
     paths:
       - "packages/core/mpc/keysign/tests/fixtures/mobile/**"
+      - "scripts/check_golden_corpus_sync.py"
+      - ".github/workflows/golden-corpus-sync.yml"
   schedule:
     - cron: "37 6 * * 1"
   workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  ANDROID_CORPUS: app/src/androidTest/assets
+  IOS_CORPUS: VultisigApp/VultisigAppTests/TestData
+  SDK_CORPUS: packages/core/mpc/keysign/tests/fixtures/mobile
 
 jobs:
   corpus-sync:
     name: Cross-repo golden corpus sync
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v4
 
       - name: Fetch sibling corpora
         run: |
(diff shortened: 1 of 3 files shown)


gomesalexandre (changes requested):
blocking on current head `330540547632c9a82fee7c9ea7e5b0a5f60c4214`.

`scripts/check_golden_corpus_sync.py:53-54` currently classifies a pure reordering of `expected_image_hash` as a warning instead of drift:

```py
if len({tuple(sorted(h)) for h in pinned.values()}) == 1:
    return [], warnings + [f"{label}: same hashes, different order"]
```

That is too weak for the current fixture semantics. In the SDK golden test, hash order is only treated as set-like for explicit exceptions (`UTXO` chains and `mayaswap.json`). Everything else stays order-sensitive:
- `packages/core/mpc/keysign/tests/mobileFixtures.golden.test.ts:64-65` only enables set comparison for `UtxoChain` or `mayaswap.json`
- `packages/core/mpc/keysign/tests/mobileFixtures.golden.test.ts:181-187` otherwise compares `actual` to `expected_image_hash` in exact order

So this checker can exit 0 on a real cross-repo drift. Current-head example: `thorchainswap.json :: Swap from ETH.USDC to RUNE` has two hashes, and `thorchainswap.json` is not one of the set-like exceptions. If one platform reverses those two outputs, its local fixture test can still pin that reversed order while this workflow only warns and passes.

Please make reordered lists fail by default, and only downgrade to order-insensitive comparison for the same explicit set-like cases the golden tests already use.

gomesalexandre (approved):
Looks good. Previous order-sensitivity blocker is fixed on the current head: reordered `expected_image_hash` lists now fail as drift, the workflow stays read-only (`contents: read`, `persist-credentials: false`), and the new cross-repo sync job passed on the live sibling corpora.

I only did a conceptual/remote pass here. `Unit Tests` and `Lint and Type Check` were still running when I reviewed, so merge should still wait for green CI on this head.

PR #2224, opened 2026-08-26
Handle DKLS malicious party keysign aborts

johnnyluo:
## Summary
- add a typed DklsMaliciousPartyError for DKLS abort-and-ban party codes 100-109
- convert ECDSA keysign inputMessage aborts into that typed error immediately
- add focused coverage for code detection and error messaging

## Validation
- yarn test:core --run packages/core/mpc/keysign/error.test.ts
- yarn prettier --config .config/.prettierrc --check packages/core/mpc/keysign/error.ts packages/core/mpc/keysign/index.ts packages/core/mpc/keysign/error.test.ts .changeset/dkls-keysign-malicious-party.md
- yarn eslint --config .config/eslint.config.mjs packages/core/mpc/keysign/error.ts packages/core/mpc/keysign/index.ts packages/core/mpc/keysign/error.test.ts

## Notes
- Full yarn typecheck currently fails on pre-existing unrelated errors in @mysten/sui/grpc, Cardano auxiliaryData types, and missing koffi types.

Refs vultisig/vultisig-windows#4686

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of DKLS signing failures by identifying malicious-party protocol errors.
  * Added support for extracting error codes from native values, error objects, and protocol messages.
  * Errors now include the affected party’s index and, when available, party ID.
  * Preserved unrelated and non-ECDSA errors without changing their behavior.

* **Tests**
  * Added coverage for native and WebAssembly errors, valid code boundaries, message formats, invalid codes, and signing-session failures.


gomesalexandre (commented):
reviewed current head `bdda82a23f9d3e783c7c2b22c4a118a0ebd40cde` remotely.

no new code-level blocker jumped out in the changed paths.

what i re-grounded
- `packages/core/mpc/keysign/index.ts` now translates native DKLS abort-and-ban party failures thrown from `session.inputMessage(...)` into a typed `DklsMaliciousPartyError`, but only on the ECDSA / DKLS path. non-DKLS signatures still rethrow the original error.
- `packages/core/mpc/keysign/error.ts` keeps the code gate tight to the native malicious-party range (`100..109`) and also recognizes the symbolic `LIB_ABORT_PROTOCOL_AND_BAN_PARTY_<n>` shape, so unrelated abort codes still fall through untouched.
- the new unit coverage in `packages/core/mpc/keysign/error.test.ts` pins both numeric and symbolic extraction plus the user-facing party numbering.
- the RN entrypoint re-exports the Hermes-safe seedphrase helper family directly from the underlying modules, and `packages/sdk/tests/unit/platforms/react-native/entry.test.ts` explicitly asserts those re-exports while confirming the eager `ChainDiscoveryService` surface is still absent.

not stamping approve on this head.
- four-eyes blocks a gomes approval here because the branch already contains gomes-authored non-merge commits (`2` by `gh pr view --json commits ...`).
- current-head CI is not fully finished yet on this exact head: `Unit Tests` and `Lint and Type Check` are still in progress.

once the current head finishes green, this should get a non-gomes stamp.

rcoderdev (approved):

rcoderdev (commented):
The new typed error is not reachable on all paths that can report this DKLS failure. keysign only converts errors thrown by session.inputMessage(), while session.finish() remains unwrapped; the pinned native header lists LIB_ABORT_PROTOCOL_AND_BAN_PARTY_* as a dkls_sign_session_finish error. On WASM, inputMessage() exposes only a boolean, so it cannot supply the code there. On iOS, the bridge formats the error as failed with error code 103, but the parser only recognizes code: 103 or a numeric .code. That leaves Windows unable to emit the warning and can leave iOS or finish failures generic. Could we surface structured codes from WASM, convert qualifying ECDSA errors from both inputMessage and finish, recognize the actual iOS message shape, and add engine-path coverage?

gomesalexandre (commented):
reviewed current head `d6fdb7c6d5429b4653b6a6e5b76e45fd3029609f` remotely.

No new code-level blocker jumped out in the DKLS malicious-party mapping itself.

What I re-grounded on this exact head:
- `packages/core/mpc/keysign/error.ts` still keeps the malicious-party gate tight to the native `100..109` range, recognizes both `code: 103` and iOS-style `error code 103` messages, and maps the symbolic `LIB_ABORT_PROTOCOL_AND_BAN_PARTY_<n>` constants back into that range.
- `packages/core/mpc/keysign/index.ts` now routes qualifying ECDSA/DKLS abort-and-ban failures from both `session.inputMessage(...)` and `session.finish()` through `rethrowKeysignError(...)`, so the typed `DklsMaliciousPartyError` survives on both paths instead of only the input-message side.
- the focused coverage matches that behavior: `packages/core/mpc/keysign/error.test.ts` pins numeric + symbolic extraction and `packages/core/mpc/keysign/index.test.ts` pins the finish-path conversion to `DklsMaliciousPartyError`.

Not stamping approve on this head.
- this branch contains gomes-authored non-merge commits (`0ec1a65e809e30ded50dc0e5e4c02ce9dc99a5da`, `e1cbca5e26b4872f3cb8439eeaf0b9b9554514be`, `c9e30a1fc17348a5d28954c493761327c50d2a54`, `f4c3abb6edc356ea9b49552230064fcfeaab86c3`), so four-eyes blocks a gomes approval here.
- the current head still needs a non-gomes current-head approval once the in-flight checks finish on `d6fdb7c6d5429b4653b6a6e5b76e45fd3029609f`.

Ehsan-saradar (changes requested):
Reviewed the current head (`d6fdb7c`). The typed error and the ecdsa-only gating are the right shape, and the native mobile bridge formats are covered — but the mapping never engages on the WASM engine, which is the platform vultisig/vultisig-windows#4686 is actually about.

### Blocker: the WASM error shape isn't recognized

On the WASM engine, `SignSession.inputMessage()` returns a plain boolean and can't surface protocol errors at all; the abort-and-ban failure only comes out of `finish()`, which rethrows the Rust `SignError` display string:

```
Abort the protocol and ban the party ID <id>
```

The string is embedded in the shipped `packages/lib/dkls/vs_wasm_bg.wasm`, and `<id>` is the banned party's **0-based** id (go-dkls maps `AbortProtocolAndBanParty(_, 0)` → `LIB_ABORT_PROTOCOL_AND_BAN_PARTY_1`, so the equivalent code is `100 + id`).

None of the parser's branches — `code: N`, `error code N`, a numeric `.code`, or the symbolic `LIB_ABORT_PROTOCOL_AND_BAN_PARTY_N` — match that message. So for every WASM consumer (browser extension, desktop, Node), `keysign()` keeps rethrowing the generic `Error`, `DklsMaliciousPartyError` is only reachable through the mobile bridges, and the UI branch in vultisig-windows that keys off `error.name === 'DklsMaliciousPartyError'` can never trigger where the issue was filed. Suggest adding a WASM branch, e.g.

```ts
// vs-wasm finish() rethrows the SignError display string with a 0-based party id
const wasmBannedPartyPattern = /\bban the party ID (\d+)\b/
```

mapping to `100 + id`, plus a fixture with the literal WASM string in `error.test.ts`.

### The party number is ambiguous for callers

The message renders `party ${code - 100 + 1}` — a 1-based index into the *setup message's* party order. Only the initiating device knows that order (`devices: [localPartyId, ...peers]` in `ensureSetupMessage`); a joiner receives the initiator's setup from the relay, and its own `[localPartyId, ...peers]` generally differs — for code 100, a joiner mapping the index against its local list would blame itself. Consider exposing `partyIndex` on the error, resolving it to the actual party id inside `keysign()` when `isInitiatingDevice`, and documenting that the index refers to setup-message order so callers don't map it against their own peer list.

### Matching is broader than the sources it covers

- `\b(?:code:\s*|error code\s+)(\d+)\b` classifies *any* error whose message contains `code: 103` as a malicious party — e.g. an HTTP "Request failed with status code: 103". Anchoring to the actual bridge shapes (`failed (code: …)`, `failed with error code …`) avoids false positives.
- The Android bridge interpolates the SWIG enum, so its real message is `… failed (code: LIB_ABORT_PROTOCOL_AND_BAN_PARTY_4)` — the symbolic branch handles it, but the `(code: 103)` test fixture doesn't represent the real shape; worth pinning the symbolic-in-context form too.
- The new `try` around `inputMessage` also wraps `fromMpcServerMessage` decryption; narrowing it to just the engine call keeps unrelated failures out of the parser.

### Nits

- `await Promise.resolve(session.finish())` → `await session.finish()` (`await` already accepts `T | Promise<T>`), and the trailing `throw error` after the `never`-returning helper is unreachable.
- The changeset could also bump `@vultisig/sdk`: `packages/core/mpc/keysign/index.ts` is bundled source, and the guard currently passes only because unrelated pending changesets happen to bump the SDK.

Happy to re-review once the WASM shape is covered — that's the one change I'd consider blocking; the rest are follow-ups or test hardening.

Ehsan-saradar (approved):

PR #2228, opened 2026-08-26
feat(sdk): export numeric EVM chain ID

### LINKED ISSUE #2134: public API: export a stable numeric EVM chain-id canonical so app/backend can delete local maps
## Summary
The SDK still does not publish a stable public numeric EVM chain-id canonical even though first-party consumers need one for tx policy, bridge metadata, and route binding. The helper effectively stays trapped off the supported root/public surface, so app and agent-backend-ts keep their own `Chain -> number` maps.

## Evidence
- consumer copies:
  - `agent-backend-ts/src/mastra/tools/mcp/lib/evm-chains.ts`
  - app-local chain registry in `vultiagent-app`
- existing root helper only covers the hex-string chain-id path, not the numeric accessor consumers keep reimplementing.

## Impact
This is a recurring architecture drift class and already produced real mismatches on EVM-adjacent routes. Numeric chain IDs should have one SDK-owned canonical instead of app/backend mirrors.

## Fix sketch
Publish a stable numeric EVM chain-id accessor on the supported SDK surface and migrate first-party consumers onto it. Add focused tests proving parity against the existing chain registry.

## Effort
Low.

## Report
https://gist.github.com/gomesalexandre/14a48e91c9a70b3c9dd3cb851d07e7d7

rcoderdev:
Closes #2134
Exports a stable numeric EVM chain-ID canonical from the SDK and React Native entrypoints so app and backend consumers can remove duplicated local maps. The existing hex accessor now derives from the same canonical, and the React Native transaction builder consumes the shared SDK value.

Includes focused coverage for all supported EVM chains, public exports, the React Native entrypoint, and the built package. A changeset is included.

Verified on commit 7ad5c1087a5e79da99117be309950dfa723db338 with the repository quality gates and built-package runtime checks.

### DIFF
--- a/.changeset/stable-numeric-evm-chain-id.md
+++ b/.changeset/stable-numeric-evm-chain-id.md
@@ -1,0 +1,6 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+Export `getEvmNumericChainId` from the SDK root and React Native public entries. The accessor is derived from the canonical EVM chain registry, and the React Native transaction builder now consumes it instead of maintaining a duplicate numeric chain-id table.
--- a/packages/core/chain/chains/evm/chainInfo.ts
+++ b/packages/core/chain/chains/evm/chainInfo.ts
@@ -102,7 +102,8 @@ const evmDefaultChainInfo: Record<EvmChain, ViemChain> = {
   [EvmChain.Robinhood]: robinhood,
 }
 
-const evmChainId: Record<EvmChain, string> = recordMap(evmDefaultChainInfo, chain => numberToHex(chain.id))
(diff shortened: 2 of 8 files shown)


gomesalexandre (approved):
looks good.

grounding
- `packages/core/chain/chains/evm/chainInfo.ts` now derives both numeric and hex chain IDs from the same canonical `evmDefaultChainInfo`, so the public numeric accessor and the existing hex accessor cannot drift.
- `packages/sdk/src/platforms/react-native/chains/evm/tx.ts` drops the duplicated RN-only numeric chain-id table and re-exports the canonical helper instead.
- the root sdk entry and the rn entry both export `getEvmNumericChainId`, and the updated unit coverage pins the public exports plus the Hyperliquid 999 guard.
- current-head CI is green.

PR #2234, opened 2026-08-27
Add RUJI Trade swap route for RUNE and bRUNE

### LINKED ISSUE #1868: [Add] RUJI Trade swap route for RUNE ↔ bRUNE
## Problem
bRUNE (`x/brune`) can't be acquired in-app. THORChain's native swap can't route it — no pool exists:

```
GET /thorchain/quote/swap?from_asset=THOR.RUNE&to_asset=THOR.BRUNE&amount=100000000
→ {"code":2,"message":"failed to calculate min swap amount: ... pool does not exist"}
```

bRUNE mints/redeems via the bRUNE/RUNE pair on RUJI Trade (Rujira FIN orderbook), not an L1 pool.

## What exists
- Asset encoding already works: `toNativeSwapAsset` maps `x/brune` → `THOR.bRUNE` (`packages/core/chain/swap/native/asset/toNativeSwapAsset.ts`)
- Contract config: `packages/core/chain/chains/cosmos/thor/brune-bond/config.ts`
- `swapEnabledChains.ts` unions native/CowSwap/Kyber/1inch/Jupiter/LiFi/SwapKit — no Rujira provider

## Fix
Add a RUJI Trade swap route (quote + wasm execute against the FIN bRUNE/RUNE pair) and register it in `findSwapQuote` so RUNE ↔ bRUNE resolves in the normal swap flow.

Inherited by Windows + agent-app. Mirrors: vultisig/vultisig-ios#5100, vultisig/vultisig-android#5586.

rcoderdev:
Closes #1868
Adds RUJI Trade routing for RUNE ↔ bRUNE through the SDK normal swap and THORChain signing paths.

Real-flow QA remains partial: quote selection, signing, broadcast, and chain inclusion were exercised, but the configured RUJI proxy still returns HTTP 401 `token expired`, so cross-platform FIN discovery and a successful state transition are not proven. This must not be treated as end-to-end success.

## Summary by CodeRabbit

* **New Features**
  * Added RUJI Trade swap quotes for native THORChain RUNE ↔ bRUNE conversions.
  * Added CosmWasm contract execution through the standard swap flow.
  * Added slippage protection, quote expiration, destination routing, and transaction explorer support.
  * Added RUJI Trade to supported providers and swap availability checks.
  * Bound quotes and safety checks to the selected recipient address.

* **Bug Fixes**
  * Improved validation of committed source amounts and attached swap funds.

* **Tests**
  * Added coverage for quote generation, routing, transaction construction, validation, and error handling.

### DIFF
--- a/packages/core/mpc/keysign/swap/build.ts
+++ b/packages/core/mpc/keysign/swap/build.ts
@@ -64,7 +64,7 @@ export type BuildSwapKeysignPayloadInput = {
 type TransferSwapTx = Extract<GeneralSwapTx, { transfer: unknown }>['transfer']
 type CosmosWasmSwapTx = Extract<GeneralSwapTx, { cosmosWasm: unknown }>['cosmosWasm']
 
-const getRujiTradeFundAmount = (tx: CosmosWasmSwapTx, fromCoin: AccountCoin): bigint => {
+const getRujiTradeFundAmount = (tx: CosmosWasmSwapTx, fromCoin: AccountCoin, toCoin: AccountCoin): bigint => {
   const expectedDenom =
     fromCoin.chain === Chain.THORChain && fromCoin.id === undefined && fromCoin.ticker.toUpperCase() === 'RUNE'
       ? 'rune'
@@ -100,7 +100,7 @@ const getRujiTradeFundAmount = (tx: CosmosWasmSwapTx, fromCoin: AccountCoin): bi
     !/^\d+$/.test(min.min_return) ||
     BigInt(min.min_return) <= 0n ||
     typeof min.to !== 'string' ||
-    !min.to.trim()
+    min.to !== toCoin.address
   ) {
     throw new Error('RUJI Trade CosmWasm route must contain a guarded FIN swap execute message.')
   }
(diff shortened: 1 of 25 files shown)


packages/core/mpc/keysign/swap/build.ts
```diff
@@ -58,6 +65,69 @@ export type BuildSwapKeysignPayloadInput = {
...
+        ? 'x/brune'
+        : undefined
+
+  if (!expectedDenom) {
+    throw new Error('RUJI Trade CosmWasm swaps require a THORChain RUNE or bRUNE source coin.')
+  }
+  if (!areEqualThorAddresses(tx.sender, fromCoin.address)) {
+    throw new Error('RUJI Trade CosmWasm sender does not match the source account.')
+  }
+  if (!tx.contract.trim() || !tx.executeMsg.trim()) {
```
johnnyluo: The RUJI signing guard only checks that `cosmosWasm.contract` is non-empty, then forwards that address into the signed Wasm execute payload. Because the swap quote fingerprint is explicitly an unkeyed mutation check, not an authenticity boundary, a constructed or tampered RUJI quote can pass the local amount/denom/recipient checks while sending funds to an arbitrary CosmWasm contract. Please validate the contract against a known RUJI FIN market allowlist for the RUNE/bRUNE pair, or otherwise re-bind it to trusted market discovery at the signing boundary, and add a regression test for a wrong contract.

johnnyluo (approved):
Approved. Re-reviewed latest head 194ce363a: the RUJI contract validation finding is fixed with a trusted FIN contract check in both quote discovery and the signing path, plus regression coverage. Focused RUJI tests passed locally after building shared packages: 5 files, 26 tests.

PR #2235, opened 2026-08-27
fix(chain): don't report a trading halt when an aggregator could still fill

### LINKED ISSUE #2167: [Fix] findSwapQuote throws TradingHalted even when LiFi/SwapKit are eligible
# [Fix] findSwapQuote throws TradingHalted even when LiFi/SwapKit are eligible

Windows/extension consume this. iOS ETH→SOL showed "Trading Halted"; retry then quoted. Same all-fail path is here.

## Problem
`findSwapQuote` already `Promise.allSettled`s every fetcher (THOR `assertNativeTradingOpen` only inside the native fetcher). If any aggregator fulfills, that quote wins. Good.

When **all** reject, the all-fail scan does:

```
if (proactiveTradingHalt) throw proactiveTradingHalt
if (haltedProviders.size > 0) throw TradingHalted
```

So a THOR halt + LiFi/SwapKit timeout (`withTimeout(..., QUOTE_FETCH_TIMEOUT_MS)`) becomes **Trading Halted** for the whole pair. The user does not retry aggregators. A second tap often works because the aggregator is warm.

ETH→SOL fetchers: THORChain, LiFi, SwapKit.

## Solution
Halt is per native provider. After allSettled:

- Any fulfilled aggregator/native quote still wins (unchanged).
- If the only successes are empty and a native fetcher halted, **retry transient aggregator failures once** (timeout / network / 5xx / SwapKit screening).
- Throw `TradingHalted` only when no non-native fetcher was in `fetchers`, or those retry results are also structural (no route).

Sole-route `thorIsSoleRoute` + `assertNativeTradingOpen` stays.

### Must Do
- [ ] All-fail path does not throw halt while LiFi/SwapKit/Jupiter/1inch/Kyber were in `fetchers` and failed transiently
- [ ] One automatic retry of those fetchers
- [ ] Halt still throws on THOR-only pairs
- [ ] Unit in `findSwapQuote.ranked.test` / halt tests: halt + aggregator timeout + retry success → quote

### Must NOT Do
- Don't skip aggregators when THOR is halted (they already run)
- Don't remove `assertNativeTradingOpen` from the THOR fetcher
- Don't change the broadcast-time halt guard

## Acceptance Criteria
- [ ] ETH→SOL, THOR halted, LiFi fulfills: `{ best }` is LiFi
- [ ] ETH→SOL, THOR halted, LiFi times out then succeeds on retry: quote, no TradingHalted
- [ ] THOR-only pair still throws TradingHalted

Windows sibling will track the SDK bump.

Ehsan-saradar:
## Description

`findSwapQuotes` treated a native trading halt as a verdict on the whole pair. For ETH→SOL the eligible fetchers are THORChain, LiFi and SwapKit, so a THOR halt plus a LiFi/SwapKit timeout ended in `throw TradingHalted` — the user saw "Trading Halted" and only got a quote by tapping again, once the aggregator was warm.

A halt is scoped to the provider that raised it, so it should never speak for an aggregator that was merely unreachable.

Fixes #2167

### How

- **One automatic retry.** `retryTransientFetchersAfterNativeHalt` re-runs only the non-native fetchers that failed transiently (timeout / network / 5xx / rate limit), and only when a native fetcher actually reported a halt. Results merge back at their fetcher index, so ranking and the below-minimum / no-route classification see the fresh outcomes. Native fetchers are never re-asked — that would only re-confirm the halt.
- **Shorter retry budget** (`QUOTE_FETCH_RETRY_TIMEOUT_MS` = 10s, vs 30s for the first attempt), so the worst case on the all-fail path is 40s rather than 60s. The provider already had the full budget and is warm.
- **Halt reporting is scoped.** `getHaltAllFailError` returns `TradingHalted` only when nothing else was merely unreachable: a halt-only pair, or providers that structurally declined ("no route"). Anything that failed transiently without reporting a halt itself — an aggregator still failing after its retry, or a second native protocol that timed out — returns `AllProvidersFailed` with transient wording instead. That message deliberately avoids the halt wordings, so `isTransientProviderFailure` — and agent-backend-ts's `isTransientQuoteError`, which keys off the same categories — don't re-read it as a hard halt.
- `thorIsSoleRoute` pre-flight and `assertNativeTradingOpen` inside the THOR fetcher are untouched, and aggregators still always run alongside the native fetchers.

### Acceptance criteria

- [x] ETH→SOL, THOR halted, LiFi fulfills → `best` is LiFi
- [x] ETH→SOL, THOR halted, LiFi times out then succeeds on retry → quote, no `TradingHalted`
- [x] THOR-only pair still throws `TradingHalted`

### Scope note

The retry set and the unreachable check are deliberately different sets. Only aggregators are retried — re-asking the protocol that raised the halt would just re-confirm it — but the unreachable check covers *any* provider that failed transiently without reporting a halt itself. That second half was added in `b3db483` after review: a THORChain halt alongside a MayaChain timeout was still reporting `TradingHalted`, which is the same misclassification for a pair both native protocols can route.

## Which feature is affected?

- [x] Swap — quote-selection only; no signing or broadcast path is touched, so there is no tx to attach.

## Checklist

## Additional context

New `findSwapQuote.halt.test.ts` covers 8 cases: the three acceptance criteria, that every transiently-failed aggregator is retried (not just the first), that a still-unreachable aggregator reports the transient failure, that a structural decline is not retried, that nothing is retried when no protocol halted, and the native-vs-native case above. Mostly over the real ETH→SOL fetcher set (THORChain + LiFi + SwapKit); the native-vs-native case uses a same-chain EVM pair with the aggregators excluded, so only THORChain and MayaChain remain. Each behavioural case was verified to fail against the code it fixes; the guard cases pass either way.

Local checks: 3140 core tests pass, `lint` / `format:check` / `knip` clean, `changeset:check-sdk-bundle` passes, and the cognitive-complexity gate on `findSwapQuotes` stays within budget (the halt reporting is extracted for that reason). Two pre-existing failures on this checkout are unrelated and reproduce on a clean `main`: `@scure/bip39/wordlists/*.js` resolution (typecheck + 19 sdk test files) and a `packages/rujira/src/client.ts` `GasPrice` mismatch.

Downstream: vultisig/vultisig-windows#4716 tracks the version bump — it needs no client-side change, since `resolveMarketSwapErrorMessage` already maps `AllProvidersFailed` to a "temporarily unavailable, try again in a moment" string.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved swap quote handling when THORChain or MayaChain is temporarily halted.
  - Automatically retries aggregator quotes after transient failures.
  - Reports a temporary network failure when providers remain unreachable.
  - Preserves clear `TradingHalted` errors for routes blocked by a protocol halt.
  - Improved identification and reporting of affected swap providers.

- **Tests**
  - Added coverage for halted protocols, fallback behavior, retries, provider timeouts, and error reporting across supported swap routes.

### DIFF
--- a/.changeset/swap-halt-aggregator-retry.md
+++ b/.changeset/swap-halt-aggregator-retry.md
@@ -1,0 +1,8 @@
+---
+'@vultisig/core-chain': patch
+'@vultisig/sdk': patch
+---
+
+fix(swap): stop reporting a native trading halt when an aggregator could still fill
+
+`findSwapQuotes` collapsed a THORChain/MayaChain halt plus a transient aggregator failure into `TradingHalted` for the whole pair, so an ETH→SOL quote surfaced "trading halted" whenever LiFi/SwapKit happened to time out. A halt is scoped to the native protocol that raised it: those aggregators now get one automatic re-attempt, and when they are still unreachable the error reports the transient failure instead of a halt the user cannot retry out of. Halt-only pairs still throw `TradingHalted`.
--- a/packages/core/chain/swap/quote/__tests__/swapQuoteFixtures.ts
+++ b/packages/core/chain/swap/quote/__tests__/swapQuoteFixtures.ts
@@ -25,6 +25,26 @@ export const evmSameChainCoins = {
   },
 } as const
 
+/**
+ * Ethereum → Solana: the cross-chain pair from the trading-halt report. It
+ * registers exactly three fetchers — THORChain (native), LiFi and SwapKit — so a
+ * halt on the native leg still leaves two aggregators that can fill the swap.
+ */
+export const evmToSolanaCoins = {
+  from: {
+    chain: Chain.Ethereum,
+    address: '0xsender',
+    decimals: 18,
+    ticker: 'ETH',
+  },
+  to: {
+    chain: Chain.Solana,
+    address: 'SoLsender',
+    decimals: 9,
+    ticker: 'SOL',
+  },
+} as const
+
 /** An aggregator quote with the minimal fields ranking and binding need. */
 export function minimalGeneralQuote(
   dstAmount: string,
--- a/packages/core/chain/swap/quote/findSwapQuote.halt.test.ts
+++ b/packages/core/chain/swap/quote/findSwapQuote.halt.test.ts
@@ -1,0 +1,137 @@
+import { Chain } from '@vultisig/core-chain/Chain'
+import { getLifiSwapQuote } from '@vultisig/core-chain/swap/general/lifi/api/getLifiSwapQuote'
+import { getSwapKitQuote } from '@vultisig/core-chain/swap/general/swapkit/api/getSwapKitQuote'
+import { getNativeSwapQuote } from '@vultisig/core-chain/swap/native/api/getNativeSwapQuote'
+import { getNativeSwapTradingHalt } from '@vultisig/core-chain/swap/native/halts/getNativeSwapTradingHalt'
+import { getNativeSwapMinAmountIn } from '@vultisig/core-chain/swap/native/minimum/getNativeSwapMinAmountIn'
+import { SwapError, SwapErrorCode } from '@vultisig/core-chain/swap/SwapError'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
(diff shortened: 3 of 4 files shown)


johnnyluo (approved):
LGTM

Issue #2233, opened 2026-08-27
[Fix] TON: golden corpus drift — SDK is the last platform on pre-fix signing bytes [TON / S1]


aminsato:
**Problem:** Android and iOS have both now shipped the two TON signing fixes tracked here as #2183 (drop `IGNORE_ACTION_PHASE_ERRORS`) and #2185 (MAX signs the displayed amount instead of a mode-128 sweep). The SDK has not. Both changes alter the signed body, so the SDK now produces different signing bytes than either mobile platform for the same TON payload.

The cross-repo golden corpus check has been failing since:

```
3 CROSS-REPO HASH DISAGREEMENT(S) — two platforms would sign different bytes:
  DRIFT ton.json :: Send TON
    android=['db6bfdb5473ece22c91a7a497c83e8a5d9009a03ad820f46839472c89f9f9235']
    ios=['db6bfdb5473ece22c91a7a497c83e8a5d9009a03ad820f46839472c89f9f9235']
    sdk=['aefa16a3825645bcfcf305be112bc2da400d213cb8c20d87e0e43d4eb214a5f8']
  DRIFT ton.json :: Send USDT TON
    android=['145673a02a4f1b091d6f9a34bf84490309ce6ef70ae9a2c2de0e5d9c1299cca9']
    ios=['145673a02a4f1b091d6f9a34bf84490309ce6ef70ae9a2c2de0e5d9c1299cca9']
    sdk=['44cbfbae7924a26eec5e6338370ff5d1dbde3fe63f67b1e8937cff1fc28aa2f3']
  DRIFT ton.json :: TON Stake with memo
    android=['f132a896ec6ea24c25309f2c220da4efcbd99c210323e266a8dafa8c5e290b78']
    ios=['f132a896ec6ea24c25309f2c220da4efcbd99c210323e266a8dafa8c5e290b78']
    sdk=['299ecf1da0682c338130b7f3f890484100e41c8f0d2b024b1fd0214df6e310dd']
```

The three `sdk=` values are byte-for-byte the hashes Android pinned *before* its fix. The SDK corpus is a faithful record of the old behavior, not a stale copy — the divergence is in the code, not in the file.

[...]

Issue #2185, opened 2026-08-23
[Fix] TON: MAX send signs a full-balance sweep that doesn't match the displayed amount [TON]


realpaaao:
**Problem:** MAX send shows the user "balance minus fee" but actually signs `amount = 0` with mode 128 ("send everything"). Worse: there is no explicit MAX flag — we *infer* MAX whenever `amount + fee >= balance`, so typing an amount close to your balance silently upgrades to a full-balance sweep.
**Why it matters:** The signed transaction does something different from what the screen showed. That's the one thing a wallet must never do.
**Fix:** Make MAX explicit (a real flag from the UI, never inferred), and make the displayed number the signed number. Add a `sendMaxAmount: true` golden test — today no platform has one.
**Receipts:** `chainSpecific/resolvers/ton/index.ts:32-42`; `signingInputs/resolvers/ton/native.ts:76-80`; `native.test.ts` (all 7 cases false).
Part of epic #2181.

PR #2252, opened 2026-08-29
fix(ton): stop hiding action-phase failures on sends and in status

### LINKED ISSUE #2183: [Fix] TON: sends ignore action-phase failures and status calls them Confirmed [TON / S1]
**Problem:** Every TON send tells the wallet contract to ignore action-phase failures (`IGNORE_ACTION_PHASE_ERRORS` is OR'd into every send mode), and our status check only reads the compute phase. So a transfer can fail at the point where the money actually moves — and we still show "Confirmed".
**Why it matters:** The user's seqno is consumed, nothing arrives, and the wallet says everything worked. Silent money-path lie.
**Fix:** Drop the ignore flag on app-initiated sends, and read `action_ph.success` / `result_code` in the status resolver. Android already fixed the status half (vultisig-android PR #5263) — port both halves here.
**Receipts:** `packages/core/mpc/keysign/signingInputs/resolvers/ton/native.ts:78,99`, `jetton.ts:38`; `packages/core/chain/tx/status/resolvers/ton.ts:36-46`.
Part of epic #2181.

Ehsan-saradar:
## Description

Every TON send OR'd `IGNORE_ACTION_PHASE_ERRORS` (+2) into the wallet contract's send mode. That flag tells the contract to *skip* an outgoing transfer it cannot carry out instead of failing, so the transaction lands un-aborted with the seqno consumed and nothing moved. `getTonTxStatus` then read only `aborted` and `compute_ph.exit_code` — neither of which reflects the action phase — and reported the whole thing as Confirmed. The user was told a transfer succeeded while the funds never left the wallet.

Both halves are fixed:

**Send mode.** The flag is dropped from every place we build a TON transfer:

- `signingInputs/resolvers/ton/native.ts` — native send, MAX send, and dApp `signTon` messages
- `signingInputs/resolvers/ton/jetton.ts` — Jetton sends
- `packages/sdk/src/chains/ton/tx.ts` — the React Native builder, in both its `@ton/core` signing cell and the WalletCore parity input it emits alongside

MAX still uses `ATTACH_ALL_CONTRACT_BALANCE`; that behaviour is #2185's, untouched here.

**Status.** `getTonTxStatus` now reads the action phase (`success`, `no_funds`, `result_code`, `skipped_actions`) alongside the compute phase, and the fee receipt is attached to failed results too since the fee is burned either way. A transaction the indexer returns without a `description` block now stays **pending** rather than being read as success — with the phases missing there is nothing to check, and reporting success there would reopen the same hole from the other side.

The JSON key for the action phase is `action`, not `action_ph` (verified against a live `api.vultisig.com/ton/v3` response; `action_ph` is the TL-B field name, not the indexer's).

## Signing-hash impact

The send mode is part of the signing preimage, so this moves the TON signing hashes. **Clients must upgrade together** — a co-signer on an older build derives a different hash and the keysign fails outright. Both golden corpora are re-recorded:

- `packages/core/mpc/keysign/tests/fixtures/mobile/ton.json` — now reproduces exactly the three hashes tracked in #2233 (`db6bfdb5…`, `145673a0…`, `f132a896…`), so the cross-repo corpus check should go green. All three vectors are `send_max_amount: false`, so #2185 does not affect them.
- `testdata/cross-encoder-golden/ton-{native-transfer,jetton-transfer,jetton-transfer-inactive,jetton-transfer-memo}.json` — the WalletCore path and the independent `@ton/core` RN path landed on identical new hashes, which is a byte-level check that both halves of the send-mode change agree.

## Which feature is affected?

- [x] Sending — TON native and Jetton sends, and dApp `signTon`

## Checklist

New tests:

- `packages/core/chain/tx/status/resolvers/ton.test.ts` (new, 15 cases) — covers each way an action phase can report failure while compute succeeds and the tx is un-aborted, plus the indexed-but-undescribed pending case and the fee receipt.
- `signingInputs/resolvers/ton/native.test.ts` — asserts the ignore bit is clear on all four transfer builders, and pins the mode each one uses.

`yarn test:core` is fully green (3185 passed). The `@vultisig/sdk` unit suite shows 19 failing files / 2 failing tests both before and after this branch — all pre-existing and unrelated (seedphrase wordlists, root-export contracts, and other missing-build resolution failures in this environment). All TON suites pass.

Fixes #2183

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TON transaction status reporting by detecting action-phase failures, reverted transactions, and incomplete indexed results.
  * TON transfers now correctly report failures and preserve accurate fee and receipt details.
  * Jetton transfers now validate the sender wallet before being created.

* **Compatibility**
  * Updated TON signing behavior and signing vectors. Compatible client versions are required when co-signing transactions created with this release.

* **Documentation**
  * Added guidance for Jetton transfer construction and forwarding amounts.


gomesalexandre (commented):
reviewed current head `88afabe00384d0a210bbd4f429bc726b76a87d16`.

what i re-grounded
- the diff removes `IGNORE_ACTION_PHASE_ERRORS` from every app-initiated TON send surface that matters here: native sends, MAX sends, dApp `signTon`, Jetton sends, WalletCore parity, and the V4R2 signing-cell builder.
- `packages/core/chain/tx/status/resolvers/ton.ts` now fail-closes on action-phase signals that previously slipped through as false success: `success: false`, `no_funds`, non-zero `result_code`, or `skipped_actions > 0`. an indexed transaction with missing execution details now stays pending instead of being guessed as success.
- the golden/vector churn is expected and appropriately documented because the send mode is part of the signing preimage.
- focused local receipt: `./node_modules/.bin/vitest run packages/sdk/tests/unit/chains/ton.test.ts --maxWorkers=2` passed `14` tests on this exact head, including the updated signing-payload expectation without `IGNORE_ERRORS`.

not stamping approve yet in this pass because this is a signing and fund-safety change and I do not have a clean full exact-head build receipt from this checkout: `yarn build:all` is failing the same way on pristine `origin/main` with the pre-existing Cardano `auxiliaryData` typing drift, and the two more targeted TON suites in this disposable clone were environment-tainted by missing local generated-package resolution. current GitHub checks are still queued.

rcoderdev (approved):

PR #2253, opened 2026-08-29
feat(sdk): export IBC reverse route resolver

### LINKED ISSUE #2151: public API(ibc): export resolveSourceChannelByDestChain so consumers stop rebuilding the reverse route index
## Summary
`resolveSourceChannelByDestChain(...)` was private even though first-party IBC consumers need the exact same reverse route helper that `prepareIbcTransfer(...)` already uses internally.

## Evidence
- SDK canonical helper lives privately in `packages/sdk/src/tools/prep/ibcTransfer.ts:126-129`
- `supportedIbcDestinationsFrom(...)` is already public at `packages/sdk/src/tools/prep/ibcTransfer.ts:131-137`
- `agent-backend-ts/src/mastra/tools/mcp/tools/send/build-ibc-transfer.ts:220-266` had to rebuild the reverse index locally because the helper was not exported

## Impact
This forces first-party consumers to rebuild the SDK IBC route index locally, which is exactly the duplicated-not-imported drift class the /improve campaign keeps finding.

## Target shape
Export `resolveSourceChannelByDestChain(...)` from supported SDK surfaces so backend/app/CLI consumers can reuse the canonical reverse route lookup instead of maintaining a second index.

## Effort
Low. Public-surface export + regression tests.

## Report
Round 162 report mirror: https://gist.github.com/gomesalexandre/16f81c059d1362ccf450a2b1db1a450c

rcoderdev:
Closes #2151


gomesalexandre (approved):
looks sound.

what i checked
- the diff is a pure public-surface export change: `resolveSourceChannelByDestChain` becomes exported from `tools/prep/ibcTransfer.ts` and is threaded through the prep, tools, root SDK, and react-native barrels.
- the helper itself is unchanged, so the behavioral blast radius is limited to exposing the existing canonical reverse-route lookup rather than changing route resolution semantics.
- the changeset is present for the new public SDK surface.
- tests pin the export on every intended surface and still assert concrete route lookups, which is the right regression shape for a public-barrel addition.

no blocker from the diff. approving.

PR #2256, opened 2026-08-29
fix(ton): a MAX send signs the amount it displayed

### LINKED ISSUE #2185: [Fix] TON: MAX send signs a full-balance sweep that doesn't match the displayed amount [TON]

Ehsan-saradar:
## Description

Two bugs, both in the gap between what the screen said and what got signed.

**The signed transaction wasn't the displayed one.** A MAX send showed `balance - fee`, then signed `amount = 0` under `ATTACH_ALL_CONTRACT_BALANCE` — a sweep the wallet contract resolves at *execution* time, so it moves whatever the balance happens to be when the transaction lands, not the number the user approved.

**Nothing marked a send as MAX.** It was inferred:

```ts
return amount + fee >= balance
```

so typing an amount that happened to land near your balance silently upgraded an ordinary send into a full-balance sweep.

### The fix

`buildNativeTonTransfer` now always signs `keysignPayload.toAmount` under `PAY_FEES_SEPARATELY`. A MAX send is just `balance - fee` as an ordinary amount, and that fee is the reserve the send mode then draws on. **The signed bytes no longer depend on the flag at all**, which is what makes the class of bug impossible rather than merely fixed.

`sendMaxAmount` becomes an explicit input, defaulting to `false`, threaded `prepareSendTx` → `prepareSendTxFromKeys` → `buildSendKeysignPayload` → `getChainSpecific`. It's recorded in `TonSpecific` as description rather than instruction. Dropping the inference also removes a `getCoinBalance` round-trip from every TON send.

## Behaviour change worth flagging

A MAX send that previously leaned on the sweep to cover an amount larger than `balance - fee` will now **fail on chain** instead of quietly moving more than was shown. That's the intended direction — and with #2183 in, the failure surfaces honestly in the status instead of reporting Confirmed. Callers should pass the displayed `balance - fee` as `amount` and set `sendMaxAmount: true` alongside it.

## Two deliberate non-changes

**`refineKeysignAmount` still skips TON.** It's tempting to now clamp `toAmount` to `balance - fee` there, since the exclusion existed because the sweep used to absorb over-balance amounts. I left it: clamping silently reduces the user's amount, which is the same sin this issue is about, and it would clamp against a fee estimate the epic separately flags as diverging 5× from the other platforms. Worth revisiting with the fee item, not here.

**No `sendMaxAmount: true` vector in the shared mobile corpus.** The issue asks for one and it should exist, but its pinned hash depends on #2183 (PR #2252), which isn't merged. A new case added here would merge cleanly and then be silently wrong — no textual conflict, just a red golden test. It also needs to land in all three repos together to be worth anything. Better as a coordinated follow-up once #2252 is in.

Instead the MAX behaviour is pinned at the payload level, which is stronger for this bug and stable either way: `getTonSigningInputs` must produce **byte-identical** input for the same amount whether or not the payload is flagged MAX.

## Which feature is affected?

- [x] Sending — TON native MAX sends

## Checklist

New tests, all six verified to fail against the pre-fix resolvers and pass here:

- `signingInputs/resolvers/ton/index.test.ts` (new) — a MAX payload signs the displayed amount, never the sweep mode; signs byte-identical input to the same payload flagged `false`; and still distinguishes two different amounts. That last one fails on the old code in a way worth seeing: both amounts encoded to `0`, so two different sends signed the same bytes.
- `signingInputs/resolvers/ton/native.test.ts` — the transfer carries the payload amount and never the sweep bit. Asserted as mode *bits* rather than an exact mode so it holds regardless of what else occupies that field.
- `chainSpecific/resolvers/ton/index.test.ts` — the caller's flag wins whatever the amount looks like, a near-balance send stays unflagged, and the balance is never read at all.

`yarn test:core` is green (3172 passed). The `@vultisig/sdk` unit suite shows the same 19 failing files / 2 failing tests as the base branch (seedphrase wordlists, root-export contracts, other missing-build resolution failures in this environment), and `yarn typecheck` reports the same 11 pre-existing errors, none in TON files.

Fixes #2185

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TON MAX sends to sign the displayed amount after fees.
  * Prevented regular sends from being incorrectly treated as MAX transfers.
  * Ensured MAX sends preserve the requested amount instead of sweeping the full balance.
  * Transactions exceeding the available balance after fees now fail on-chain as expected.
  * Improved validation for token transfer destinations and Jetton wallet resolution.

* **Documentation**
  * Added release notes describing the TON MAX-send fixes across the core MPC and SDK packages.


Ehsan-saradar:
@coderabbitai review

Ehsan-saradar:
@coderabbitai review

gomesalexandre (commented):
reviewed current head `6ea1725fc1e3ab64012a04d0a991e8ec2395ba08`.

no new blocker surfaced from the current diff.

what i re-grounded
- `packages/sdk/src/vault/VaultBase.ts` now threads `sendMaxAmount: amount === 'max'` into `prepareSendTx(...)`, which closes the convenience-path gap CodeRabbit flagged. explicit near-max sends still pass `false`.
- `packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.ts` no longer infers MAX from `amount + fee >= balance`; it records only the caller flag, so an ordinary near-balance send is no longer silently relabelled a sweep.
- `packages/core/mpc/keysign/signingInputs/resolvers/ton/index.ts` and `.../native.ts` now keep `sendMaxAmount` descriptive only. the signed TON transfer always uses `keysignPayload.toAmount` under `PAY_FEES_SEPARATELY`, not `ATTACH_ALL_CONTRACT_BALANCE` with `amount = 0`.

receipts i did get on this exact head
- `./node_modules/.bin/vitest run --config packages/sdk/tests/unit/vitest.config.ts packages/sdk/tests/unit/tools/prep/send.test.ts` -> `15 passed`, including the new `forwards the MAX flag to the payload builder` coverage.
- `./node_modules/.bin/vitest run packages/core/mpc/keysign/signingInputs/resolvers/ton/native.test.ts --maxWorkers=2` -> `9 passed`, including the new assertions that a MAX send signs the given amount and does not set `ATTACH_ALL_CONTRACT_BALANCE`.

why this is still a comment, not an approve
- i could not get the stronger exact-head SDK build / CLI receipt yet because `YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn build:shared` fails here on the same pre-existing Cardano typing drift as pristine `origin/main`:
  `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts(72,5): error TS2353: Object literal may only specify known properties, and 'auxiliaryData' does not exist in type 'ISigningInput'.`
- current GitHub checks are still queued, so i am not stamping a fund-safety TON send change without a clean higher-level exact-head receipt.

net: the current-head MAX-intent fix itself looks correct, but this still wants a clean build/CI receipt before approve.

gomesalexandre:
pushed a mergefix on current head `2b0b57f32` to clear the stale `CONFLICTING` state.

what changed
- merged fresh `main` into `fix/ton-max-send-2185`
- resolved the only live conflict in `packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts` by keeping both the PR's `sendMaxAmount` coverage and main's jetton-wallet-resolution coverage

receipts
- `./node_modules/.bin/vitest run packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts packages/core/mpc/keysign/signingInputs/resolvers/ton/native.test.ts`
- `cd packages/sdk && ../../node_modules/.bin/vitest run --config tests/unit/vitest.config.ts tests/unit/tools/prep/send.test.ts tests/unit/vault/vaultBaseSendMax.test.ts`
- `yarn build:all` is still blocked in this exact-head clone by a pre-existing Cardano typing error at `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts:72` (`auxiliaryData` on `ISigningInput`)
- an extra focused run of `packages/core/mpc/keysign/signingInputs/resolvers/ton/index.test.ts` stayed blocked here by missing local `@vultisig/core-chain/chains/ton/staking` build output in the borrowed-install temp clone

branch moved, so the old review is stale. this now needs a fresh non-gomes review on the new head before merge.

gomesalexandre:
pushed a tiny takeover fix to unblock current-head CI.

what changed
- `style(ton): format chain-specific resolver test` on top of `7718311604f0ee9333f155456e5529a54def58c1`
- only touched `packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts`
- this is just the missing prettier normalization from the `merge main` head, no logic delta

receipt
- `~/Sites/vultisig-sdk/node_modules/.bin/prettier --check packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts`
- `git diff --check`

current head now needs fresh CI plus a non-gomes current-head review before merge.

gomesalexandre:
pushed `64ff99a` to clear the current blocker.

what changed
- ran the repo prettier config over `packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts`
- no logic changes, just style normalization so the file matches the repo formatting contract

receipts
- `~/Sites/vultisig-sdk/node_modules/.bin/prettier --config .config/.prettierrc --check packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts` -> clean in an exact-head temp clone before push
- tried a focused `vitest` rerun from the same temp clone with a borrowed install, but that environment dies on a pre-test import/build-state gap for `@vultisig/core-mpc/types/.../blockchain_specific_pb`, so the honest receipt here is the formatting fix only

current-head ci is re-running on `64ff99a9845e1725f156d53dc67c42390be47fa4`.
this still needs a non-gomes review once the new head is green.

gomesalexandre (commented):
reviewed current head `64ff99a9845e1725f156d53dc67c42390be47fa4`.

no new blocker from the latest delta.

what i re-grounded
- the production TON MAX-send fix is unchanged from the earlier reviewed head: the current branch still signs `keysignPayload.toAmount` under `PAY_FEES_SEPARATELY`, and `sendMaxAmount` stays a descriptive caller flag instead of silently upgrading near-balance sends into sweeps.
- the only branch-local changes after the earlier reviewed head are formatting-only updates in `packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts`.
- current-head CI is green across unit, integration, node-storage, lint/typecheck, and CodeRabbit.

not approving.
- this head includes gomes-authored commits, so four-eyes still requires a non-gomes current-head stamp.
- this is not a blocker on the code, just the merge gate for this branch.

rcoderdev (approved):

PR #2280, opened 2026-08-31
fix(ton): fail closed when the sender jetton wallet can't be resolved

### LINKED ISSUE #2186: [Fix] TON: failed jetton wallet lookup builds a transfer with an empty destination [TON]
**Problem:** When we can't resolve the user's jetton wallet address, we don't stop — the address starts as an empty string, the failed lookup leaves it empty, and the empty string passes our `shouldBePresent` check (it only rejects null/undefined). We then build a jetton transfer with destination "".
**Why it matters:** A failed network lookup should never produce a signable transaction. iOS fixed exactly this class fail-closed (ios#4534); the SDK never did — and Windows/extension inherit this code.
**Fix:** Fail closed: no resolved jetton wallet, no transaction.
**Receipts:** `chainSpecific/resolvers/ton/index.ts:60,64-74`; `signingInputs/resolvers/ton/jetton.ts:41`; `packages/lib/utils/assert/shouldBePresent/index.ts:2-4`.
Part of epic #2181.

Ehsan-saradar:
## Description

A Jetton transfer is a message to the sender's **own** jetton wallet, so that address *is* the transaction's destination. It was resolved with a swallowed failure:

```ts
jettonAddress: '',                                   // default
const { data: jettonWallet } = await attempt(...)    // failure swallowed
if (jettonWallet) result.jettonAddress = jettonWallet // failure leaves ''
```

and then, at the signing boundary:

```ts
jettonAddress: shouldBePresent(jettonAddress, 'Jetton address')
```

`shouldBePresent` only rejects `null`/`undefined`, and `TonSpecific.jettonAddress` is a plain proto string that is never either — so that check could not fire at all, and `''` sailed through into `Transfer.dest`. A failed network lookup produced a signable, broadcastable transaction whose destination was the empty string.

### The fix

**Fail closed at the source.** The chain-specific resolver now throws when the jetton-wallet lookup fails *or* returns blank, naming the ticker and owner and keeping the original failure as the error's `cause`. It stays a plain `Error` rather than a `BuildKeysignPayloadError`, following the convention the Ripple resolver already set in this same directory — an RPC timeout or indexer lag is retryable, not bad user input.

**Fail closed at the signing boundary too.** `buildJettonTransfer` rejects a blank address itself. This is not belt-and-braces: in a keysign the payload comes from whichever device initiated, so a co-signer must refuse a destination-less transfer built by an older client or a buggy peer rather than sign whatever it is handed. The vacuous `shouldBePresent` call site is removed rather than left as decoration.

Both guards use `.trim()`, so whitespace is treated the same as empty.

## Which feature is affected?

- [x] Sending — TON Jetton (USDT and other TON tokens) transfers

## Checklist

New tests, all six failure cases verified to fail against the pre-fix code and pass here:

- `chainSpecific/resolvers/ton/index.test.ts` — a failed lookup, an empty result, and a whitespace result each refuse to produce a payload; the underlying error survives as `cause`; a successful lookup is recorded; and a native TON send never consults the lookup at all.
- `signingInputs/resolvers/ton/jetton.test.ts` (new) — `buildJettonTransfer` builds against a resolved wallet and refuses an empty or blank one.

The existing seqno tests were kept and re-pointed at the mock coin, so the uninitialized-wallet regression they cover is unaffected.

## Validation

`yarn test:core` green (3219 passed). I also ran every static gate from `check:ci` that runs without a full build — `worktree:check`, `check:shared-exports`, `changeset:check-sdk-bundle`, lint, typecheck, knip, `format:check`, `quality:architecture`, `quality:complexity`, `quality:duplication`, `quality:secrets`, `quality:docs:markdown` — all pass. `yarn typecheck` reports the same 11 pre-existing errors as the base branch (seedphrase wordlists, a rujira cosmjs type clash), none in TON files.

## Note for whoever merges

Branched off `main`, not stacked on the other open TON PRs. #2255 and #2256 also touch `chainSpecific/resolvers/ton/index.ts`, in different parts of the function (the bounceable helper and the `sendMaxAmount` input respectively) — expect small textual conflicts around the imports and the resolver signature, not overlapping logic. This PR's `if (coin.id)` block is untouched by both.

Fixes #2186

## Summary by CodeRabbit

* **Bug Fixes**
  * TON Jetton transfers now fail safely when the sender wallet address cannot be resolved or is blank.
  * Invalid or whitespace-only wallet addresses are rejected before signing.
  * Underlying wallet lookup errors are preserved for clearer troubleshooting.
  * Native TON transfers continue without unnecessary Jetton wallet lookup.


gomesalexandre (approved):
looks good.

receipts
- `yarn install --immutable`
- `yarn build:all`
- `yarn cli:build`
- `./node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts packages/core/mpc/keysign/signingInputs/resolvers/ton/jetton.test.ts`
- `./node_modules/.bin/vitest run --config .config/vitest.cli.config.ts clients/cli/src/commands/__tests__/prepSubprocess.test.ts`
- `node clients/cli/dist/index.js --output json --non-interactive prep jetton-transfer ... --identity <test identity>` returned `success: true` with signing hash `4c02442896d2b7536f3b004fcb3619886f0b9d18e00551dabf0f84e2ad48ff57`

happy with the fail-closed guard at both the resolver and signing boundary.

Issue #2181, opened 2026-08-23
[Epic] TON money-path hardening (action-phase, bounceability, MAX, fees, jetton fail-open)


realpaaao:
## Epic: TON money-path hardening

Per-chain hardening initiative (XRP model: #979). Deep-dive survey 2026-08-23 across SDK/iOS/Android/Windows; all items file:line-verified. ALL-PLATFORM rule applies: each item lands in SDK **and** is consumed+verified on iOS/Android/Windows with a real broadcast.

### S1 — money loss
1. **Action-phase blindness end-to-end.** Every send mode ORs `IGNORE_ACTION_PHASE_ERRORS` (+2) into the wallet contract call (`packages/core/mpc/keysign/signingInputs/resolvers/ton/native.ts:78,99`, `jetton.ts:38`), and the status resolver reads only `aborted` + `compute_ph.exit_code` (`packages/core/chain/tx/status/resolvers/ton.ts:36-46`). A tx whose action phase fails consumes the seqno, moves nothing, and reports Confirmed. Android fixed detection only (PR vultisig-android#5263); send-mode flag still set on all 4 platforms.
2. **Swap deposits sent non-bounceable.** `swap/keysign/getSwapDestinationAddress.ts:22` passes SwapKit's UQ-form deposit address through verbatim; bounceability is then decided by `receiver.startsWith('E')` (`chainSpecific/resolvers/ton/index.ts:48`). A rejecting contract **absorbs** the deposit instead of refunding. The staking path already normalizes via `tonAddressToBounceable` (`chains/ton/address.ts:22-29`, `staking.ts:123`) with a comment explaining this exact hazard.

### S2 (ordered)
- MAX send displays balance−fee but signs `amount=0` + mode 128 full sweep; SDK *infers* MAX from `amount+fee>=balance` — silent escalation (`chainSpecific/resolvers/ton/index.ts:32-42`)
- Static fees, no estimation (#1465), diverging 5×: SDK 0.01 vs iOS/Android 0.05 TON; Android's `TonApi.estimateFee` implemented, zero callers
- SDK jetton-wallet resolution fails OPEN: `jettonAddress=''` passes `shouldBePresent` → `dest:''` (`chainSpecific/resolvers/ton/index.ts:60-74`); iOS fixed this class (ios#4534), SDK never did
- Jetton balances false-zero on RPC failure, feeding MAX (`chains/ton/api.ts:75`)
[...]

PR #2282, opened 2026-08-31
fix(ton): derive each dApp message's bounce flag from its own destination

### LINKED ISSUE #2281: [Fix] TON: dApp messages all inherit one bounce flag instead of each address's own — co-signers reading the tag compute a different hash [TON / S1]
**Problem:** For a dApp (TonConnect) `sendTransaction`, `getTonSigningInputs` stamps every message in the batch with the single wallet-level `TonSpecific.bounceable`. That flag is computed once, in `getTonChainSpecific`, from the *first* message's destination (`receiver.startsWith('E')` + deployed check). The bounce bit a message actually declares lives in its own address tag — `EQ…`/`kQ…` bounceable, `UQ…`/`0Q…` not, raw `workchain:hex` none — so a batch that mixes destinations, or whose first destination is not yet deployed, signs the wrong flag on some of its messages.

**Why it matters:** The bounce bit is part of the signed body. The iOS signer already derives it per message from the address tag (`Blockchain/Ton/Signing/Ton.swift`, `isBounceable(address:)`, pinned by `testTonConnectBounceableIsDerivedPerAddress`), so for the same `signTon` payload an SDK/extension initiator and an iOS co-signer compute different pre-image hashes whenever the flags differ, and the keysign fails outright. Android currently applies the wallet-level flag *because* per-address derivation broke parity with the extension (`data/.../crypto/TonHelper.kt`, comment above the TonConnect builder) — the SDK is the pivot; once it moves, Android moves with it (ALL-PLATFORM rule). Beyond parity: a `UQ…` message sent bounceable, or an `EQ…` contract call sent non-bounceable, changes where the funds of a rejected transfer end up.

**Fix:** Derive each dApp message's `bounceable` from its own `to`: a user-friendly address yields its tag (mainnet and testnet tags alike); a raw `workchain:hex` address declares nothing and is non-bounceable (the usual contract-deployment shape, where the destination has no code yet to bounce from). The wallet-level `TonSpecific.bounceable` and the nominator-pool comment override keep applying to app-initiated single sends only. Pin the three-destination vector (`UQ…`, `EQ…`, `Ef…` → `false, true, true`) in a unit test so the rule cannot drift again.

Windows follow-up once consumed (vultisig/vultisig-windows#4735): `core/ui/chain/tx/utils/useTonSimulation.ts` builds the emulation preview with the wallet-level flag on every message and should apply the same per-message rule.

**Receipts:** `packages/core/mpc/keysign/signingInputs/resolvers/ton/index.ts:26-34` (`bounceable: isStakeOp ? true : !!bounceable` inside the `signTonMessages.map`); `packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.ts` (`getIsBounceable` reads only `keysignPayload.toAddress`, i.e. the first message).

Part of epic #2181.

Ehsan-saradar:
## Description

A dApp `signTon` batch stamped every message with the single wallet-level `TonSpecific.bounceable`. That flag is computed once, in `getTonChainSpecific`, from the first message's destination (`toAddress`) — so a batch whose destinations don't all share that flag, or whose first destination is not yet deployed, signed the wrong bounce bit on some of its messages. The bounce bit is part of the signed body, and a co-signer that derives it from each address computes a different pre-image hash for the same payload, so the keysign fails outright.

Each dApp message now takes the flag from its own destination:

- a user-friendly address yields its tag — `EQ…`/`kQ…` bounceable, `UQ…`/`0Q…` not (mainnet and testnet tags alike);
- a raw `workchain:hex` address carries no tag: it defaults to **bounceable** (an already-deployed contract or router must refund a rejection) **unless the message carries `stateInit`** — a deployment, whose destination has no code yet to bounce from — and only in its canonical spelling (`Address.parseRaw` is lenient, so `0junk:<hash>` and friends are rejected by a round-trip check);
- anything unparseable is non-bounceable; the signer already rejects such a destination on its own.

That lives in `getTonMessageBounceable(address, hasStateInit)` in `@vultisig/core-chain/chains/ton/messageBounce` (exports map regenerated), and `getTonSigningInputs` calls it for the `signTon` branch only, threading `!!msg.stateInit`. App-initiated single sends are untouched: they keep the wallet-level flag from the chain-specific resolver and the nominator-pool comment override.

Fixes #2281 — part of epic #2181.

## Signing-hash impact

TON signing hashes move for dApp batches whose destinations do not all share the first message's flag (single-message requests to a deployed `EQ…` contract, the common swap/jetton case, are unchanged). Per the epic's ALL-PLATFORM rule this has to be consumed in step:

- Android currently applies the wallet-level flag to every dApp message specifically to stay in step with the extension/desktop initiator (`TonHelper.kt`), so it moves together with the Windows pin bump.
- iOS derives per message from the address prefix and treats every raw address as non-bounceable; the raw-without-`stateInit` case therefore differs from this branch and needs the same rule there.
- Windows follow-up under vultisig/vultisig-windows#4735: `core/ui/chain/tx/utils/useTonSimulation.ts` builds the emulation preview with the wallet-level flag on every message and should apply the same per-message rule.

## Which feature is affected?

- [x] Sending — TON dApp (`signTon`) multi-message transactions

## Checklist

Tests:

- `packages/core/chain/chains/ton/messageBounce.test.ts` — tag read for mainnet/testnet/masterchain addresses, raw addresses with and without `stateInit`, canonical-form check against malformed raw input, unparseable input.
- `packages/core/mpc/keysign/signingInputs/resolvers/ton/messages.test.ts` — a three-destination batch (`UQ…`, `EQ…`, `Ef…`) signs `false, true, true` regardless of the wallet-level flag, the nominator-pool override does not reach dApp messages, raw destinations with/without `stateInit`, and app-initiated sends still honour the wallet-level flag and the pool override.

Verification: `yarn test:core` TON suites, golden and fixture tests green; `yarn check:shared-exports` up to date; knip clean; `yarn typecheck` shows only the pre-existing `@scure/bip39` wordlist / rujira `GasPrice` errors, none in TON files.

## Additional context

The remaining differences the mobile signers have from this resolver — the send-mode flag and the MAX sweep — are covered by #2252 and #2256 and are not touched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TON dApp batch transactions by determining each message’s bounce behavior from its destination and deployment state.
  * Preserved existing bounce settings for regular wallet transfers and nominator-pool deposits.
  * Added safer handling for malformed or non-canonical TON addresses.

* **New Features**
  * Exposed TON message bounceability utilities for supported integrations.

* **Tests**
  * Added coverage for friendly, raw, masterchain, testnet, contract deployment, and invalid TON addresses.


gomesalexandre (commented):
reviewed current head `1bd30d3033d0f98ea91e44982581dbd614709a8d` after takeover fix.

what i changed
- raw TON dApp destinations now default `bounceable: true` only when they parse as a real raw address and the message does not carry `stateInit`.
- raw deployment messages with `stateInit` still stay non-bounceable.
- threaded `!!msg.stateInit` into the per-message bounce resolver and pinned both raw cases in the new TON message tests.

receipts
- `./node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/core/chain/chains/ton/messageBounce.test.ts packages/core/mpc/keysign/signingInputs/resolvers/ton/messages.test.ts --maxWorkers=2` -> 2 files passed, 13 tests passed.
- `YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn build:shared` is still blocked by the same pre-existing `packages/core/mpc/keysign/signingInputs/resolvers/cardano.ts(72,5)` `auxiliaryData` type error on both this branch and pristine `origin/main`, so that higher-level build receipt is not branch-specific evidence.

not approving. this branch now contains a gomes-authored substantive fix commit, so four-eyes requires a non-gomes current-head review.

gomesalexandre:
pushed `1bd30d3033d0f98ea91e44982581dbd614709a8d` onto this branch to fix the raw-address bounce gap.

summary
- raw `workchain:hex` dApp destinations now default bounceable only when the message is a real contract call with no `stateInit`
- raw deployment messages that do carry `stateInit` still stay non-bounceable
- added regression coverage for both cases

receipts
- `./node_modules/.bin/vitest run --config .config/vitest.core.config.ts packages/core/chain/chains/ton/messageBounce.test.ts packages/core/mpc/keysign/signingInputs/resolvers/ton/messages.test.ts --maxWorkers=2` -> 13 passed
- `YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn build:shared` is still blocked by the same pre-existing Cardano `auxiliaryData` type error on this branch and pristine `origin/main`

this now wants a non-gomes current-head review.

gomesalexandre (commented):
reviewed current head `02439e6ebe583e19d794e658ca24b2e827c56748` remotely.

looks sound on the merits.

what i checked
- `packages/core/chain/chains/ton/messageBounce.ts` now derives the bounce bit per destination instead of reusing one wallet-level flag across the full dapp batch.
- the raw-address path now requires a canonical `workchain:hex` round-trip before treating the destination as a raw address, so malformed inputs like `0junk:<hash>` do not get reported bounceable.
- the dapp-message path is the only place that changed. app-initiated single-send behavior and the nominator-pool override stay on their existing path.
- the current tests in `packages/core/chain/chains/ton/messageBounce.test.ts` and `packages/core/mpc/keysign/signingInputs/resolvers/ton/messages.test.ts` cover the mixed-destination and malformed-raw cases this branch is fixing.

not approving. this current head includes a substantive gomes-authored commit, so four-eyes still requires a non-gomes current-head review. also waiting for the still-running current-head CI jobs before any merge talk.

rcoderdev (commented):
The raw-address guard returns false for `00:<hash>` without `stateInit`, but WalletCore 4.7.3 accepts and compiles that address as workchain 0. I checked that `0:<hash>` and `00:<hash>` produce identical preimages with the same bounce flag; this change gives them different flags, so the latter can send a non-bounceable contract call and lose refund protection on rejection. Please reject noncanonical raw destinations before signing, or normalize accepted equivalents and preserve the no-stateInit bounceable rule. Add a regression that runs this case through WalletCore; the current field-only test expects false and misses that it remains signable.

Ehsan-saradar:
Confirmed and fixed in 72c1ba8da.

The canonical round-trip was the wrong test. Against WalletCore 4.7.3, `0:`, `00:`, `000:`, `+0:`, `-0:`, `-1:`, `-01:`, either hex case and a lowercase `0x` hash prefix all compile to the same pre-image, so the guard now matches that shape instead of one spelling and every signable raw destination reaches the `!hasStateInit` rule. Staying wider than the signer is free here: a shape WalletCore refuses never reaches a signed message, and the flag it would have carried is the safe one. `0junk:`, `0x0:`, `0X`-prefixed hashes, wrong-length hashes and whitespace still report non-bounceable.

The new regression in `messages.test.ts` drives real WalletCore rather than reading the field: it compiles each spelling through the resolver, asserts the signer accepts it, and pins its pre-image to the canonical one, with a third case proving the bounce bit is actually part of the signed bytes. It fails against the old guard and passes with the fix.

Changeset updated too, since it claimed the raw default applied "only in its canonical spelling".

rcoderdev (approved):

rcoderdev (approved):

PR #2296, opened 2026-09-01
fix(ton): cross-check SwapKit's tx[] transfer before building the deposit

### LINKED ISSUE #2188: [Fix] TON: SwapKit's returned transfer is validated then ignored — no cross-check against what we sign [TON]
**Problem:** SwapKit returns the exact transfer(s) to make (`tx[]` with address + amount). We decode and validate that array — then ignore it and build the transfer from the separate top-level `targetAddress`/`toAmount` fields. Nothing checks the two agree.
**Why it matters:** If they ever diverge (provider bug, API change), we sign a transfer to the wrong place or for the wrong amount, and no check would catch it.
**Fix:** Assert `tx[0].address == targetAddress` and amounts match before building. Three lines. Same check needed in iOS (`SwapPayloadBuilder.swift:503-525` vs `KeysignMessageFactory.swift:88-97`) and Android (`SwapKitQuoteSource.kt:617-629` vs `SigningHelper.kt:201-202,493-497`) — siblings tracked in the epic.
**Receipts:** `swap/build.ts:240-241` vs `:201`.
Part of epic #2181.

Ehsan-saradar:
## Description

SwapKit's `/v3/swap` response states the deposit destination in up to three independent fields — `targetAddress`, `depositAddress`, and the `tx[]` transfer array — and its size in two (`depositAmount`, `tx[0].amount`). `getTransferTargetAddress` and `getTransferAmount` take the first field that is present and never look at the rest, so a response whose halves diverge signs whichever field wins the precedence order while the others name a different recipient or a different size, with nothing downstream able to tell. SwapKit hands us the exact transfer to make and we were decoding it only as a fallback.

`buildTransferTx` now refuses a response that disagrees with itself on either the destination or the amount, rather than resolving the ambiguity by precedence: neither field is authoritative enough to pick a winner from, and a divergence is a provider bug, an API change, or a tampered payload in every case. A multi-entry `tx[]` is rejected too — only `tx[0]` is ever built, so signing it would under-fund the swap.

Destinations are compared per chain. TON spells one account as `EQ…`, `UQ…` or raw `workchain:hex`, and only the flag byte and checksum separate the first two, so the new `areEqualTonAddresses` compares parsed accounts there. Every other transfer chain stays byte-for-byte: base58 is case-sensitive, so normalizing case away would let two different addresses pass as one.

Fixes #2188
Part of #2181.

**Follow-up (review):** the agreed TON destination is re-spelled in its bounceable (`EQ…`) form whichever field it came from. The agreement check ignores bounceability on purpose — `UQ…`, raw and `EQ…` are one account — but the signer reads the bounce flag off the spelling, so a `UQ…`/raw field winning precedence would have sent the deposit non-bounceable, where a rejecting contract absorbs it. A test follows an inverse-spelled response through quote → destination → chain-specific resolver → signing input and pins `bounceable: true`. Complements #2255 on the resolver side.

## Which feature is affected?

- [x] Swap — SwapKit transfer/deposit routes (TON, XRP, TRON, LTC, DOGE, BCH, ZEC; Bitcoin PSBT and Sui PTB routes are unaffected, their `tx` is an opaque string)

## Checklist

## Tests

`packages/core/chain/swap/general/swapkit/api/getSwapKitQuote.test.ts` — new `transfer-route self-agreement (sdk#2188)` block. All five rejection cases fail without the guard (verified by disabling the call), and two acceptance cases pin that healthy routes are not falsely rejected:

- rejects a `tx[0].address` naming a different account than `targetAddress`
- rejects a `depositAddress` diverging from `targetAddress`, with no `tx[]` involved
- **accepts** one TON account spelled `EQ…` in `targetAddress` and `UQ…` in `tx[]` — bounceability is not identity
- rejects a non-TON divergence that differs only by case, proving base58 case is not normalized away
- rejects a `tx[0].amount` diverging from `depositAmount`
- **accepts** a decimal `depositAmount` and a base-unit `tx[0].amount` that agree once normalized
- rejects a multi-transfer `tx[]`

`packages/core/chain/chains/ton/address.test.ts` — new file covering `areEqualTonAddresses` across bounceable / non-bounceable / raw / base64url-vs-standard forms, plus the unparseable-input fallback to exact comparison.

`yarn test:core` green (3244 passed). Typecheck, `yarn lint`, `yarn knip` and `format:check` clean — the pre-existing `packages/rujira` cosmjs skew and `@scure/bip39` wordlist resolution errors are byte-identical to `main`.

## Additional context

The same cross-check is still needed on iOS (`SwapPayloadBuilder.swift` vs `KeysignMessageFactory.swift`) and Android (`SwapKitQuoteSource.kt` vs `SigningHelper.kt`); both are tracked in the epic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TON swap transfer validation by checking destination and amount consistency across response fields.
  * Rejects conflicting transfer details and responses containing multiple transfer entries.
  * Normalizes valid TON destinations to bounceable format and rejects invalid addresses.
  * Recognizes equivalent TON address formats while retaining exact comparisons for other chains.
  * Normalizes decimal and base-unit amounts to detect mismatches reliably.
  * Preserves bounceable TON address handling through swap deposits and signing.

* **Tests**
  * Added coverage for TON address formats, invalid inputs, destination conflicts, amount mismatches, and multi-transfer responses.


rcoderdev (commented):
`areEqualTonAddresses` intentionally ignores bounceability, but `assertTransferAgreement` then keeps the precedence-selected spelling, and the TON signing resolver derives the actual bounce flag from that spelling. An inverse equivalent pair (`UQ`/raw `targetAddress` with `EQ` in `tx[0].address`) can therefore pass validation yet send the deposit non-bounceable, risking funds being absorbed on rejection. Please canonicalize the agreed destination to bounceable form (or require flag agreement) and cover inverse `UQ`/raw precedence through the final signing input.

Ehsan-saradar:
@rcoderdev agreed — the agreement check compares by account on purpose, but then the winning spelling was carried straight into `transfer.to`, and the resolver reads the bounce flag off it. Fixed in e53415794: the agreed TON destination is re-spelled bounceable (`tonAddressToBounceable`) whichever field it came from, so `transfer.to`, `keysignPayload.toAddress` and the SwapKit payload's `targetAddress` are all `EQ…`; a string that is not a TON address is refused. Tests: inverse `UQ…` and raw `0:hex` `targetAddress` winning over an `EQ…` `tx[]` entry at the quote level, and `keysign/swap/tonSwapDepositBounce.test.ts` which follows such a response through `getSwapKitQuote` → `getSwapDestinationAddress` → `getTonChainSpecific` → `getTonSigningInputs` and pins `bounceable: true` on the final transfer. Complements #2255, which hardens the resolver side.

rcoderdev (approved):

PR #2298, opened 2026-09-01
fix(ton): one correct comment validator, run before the user signs

### LINKED ISSUE #2189: [Fix] TON: comment length validation is wrong, late, and missing on mobile [TON / Memo]
**Problem:** TON comment/memo validation is wrong three ways: (a) we enforce a fixed 123-byte cap on jetton comments, but the real inline cap is ~34–39 bytes and shrinks with the amount — our own RN builder enforces the correct one; (b) the in-wallet send flow never validates the comment at all, so a too-long memo throws late, at keysign, after the user reviewed everything; (c) iOS/Android don't check length anywhere.
**Why it matters:** Memos are load-bearing on TON — exchange deposits without the right memo lose funds. Failing late or building an invalid cell is a money-path bug.
**Fix:** One correct validator (the RN builder's math), run it in the send form before keysign, port to mobile (siblings tracked in the epic).
**Receipts:** wrong cap `signingInputs/resolvers/ton/native.ts:20-38`; right cap `packages/sdk/src/chains/ton/tx.ts:358,391-395`; validator only wired on the dApp path `vultisig-windows core/inpage-provider/.../build.ts:549-550`.
Part of epic #2181.

Ehsan-saradar:
## Description

TON comment length was checked against a fixed 123 bytes wherever it was checked at all, in three separate copies. That number is right for a **native** transfer, where the comment is the message body's own cell. It is wrong for a **jetton** transfer, where the comment rides inline in the transfer body's `forward_payload` and shares one 1023-bit cell with the opcode, query id, amount, both addresses and the forward amount. The real jetton cap is roughly 34–39 bytes and **shrinks as the amount grows**, because `VarUInteger 16` widens a byte at a time.

Anything between the two caps passed our check and reached WalletCore, which refuses to pack the cell and fails the keysign with a bare `Internal error` — after the user had reviewed and approved everything. Memos are load-bearing on TON: an exchange deposit without the right one loses the funds.

**One validator.** `@vultisig/core-chain/chains/ton/comment` now owns the rule for both shapes and is the only copy left: `tonNativeCommentMaxBytes`, `getTonJettonCommentMaxBytes({ amount, isActiveDestination })`, `getTonCommentMaxBytes` for a send form that wants to count down against the real limit, and `validateTonComment`. The amount-dependent widths are measured by encoding them rather than predicted.

**Run before keysign.** `buildSendKeysignPayload` validates at the end of the build — the last point that has both the final signed amount and `isActiveDestination` — and raises a non-retryable `BuildKeysignPayloadError('ton-memo-too-long')`. That moves the failure off the keysign and onto the screen where the payload is assembled, for every consumer, with no client-side change. The signing-input resolvers still validate, so a payload built by another device or an older client cannot get an unpackable comment signed here either.

Fixes #2189
Part of #2181.

## Breaking

- `validateTonComment` is no longer exported from `@vultisig/core-mpc/keysign/signingInputs/resolvers/ton/native`, and takes `{ memo, jetton? }` instead of a bare string. Import it from `@vultisig/core-chain/chains/ton/comment`.
- `validateTonMemo` (SDK TON surface) keeps its signature and gains an optional second argument for jetton context — not breaking.

The one known consumer of the old export is the Windows dApp send path (`core/inpage-provider/popup/view/resolvers/sendTx/keysignPayload/build.ts`), whose call becomes `validateTonComment({ memo })` against the new import path. Windows pins `@vultisig/core-mpc` 2.1.0, so that edit lands with the pin bump; the in-wallet send flow there needs no change at all, since it goes through `buildSendKeysignPayload`.

## Which feature is affected?

- [x] Sending — TON native and Jetton transfers carrying a memo/comment

## Checklist

## Tests

The cap is a number, so the tests pin it against the thing that actually enforces it rather than against my arithmetic.

`packages/sdk/tests/unit/chains/ton-comment-capacity.test.ts` (new) drives **real WalletCore** at each cap and one byte past it, deliberately bypassing our validator — 39 bytes for a 3-byte amount, 34 for an 8-byte amount, 40 for an inactive destination, 123 native. It also proves the native cap is *not* safe for jettons (WalletCore refuses a 123-byte jetton comment), and that the exported cap and the RN builder's in-situ check stop at the same byte.

`packages/core/chain/chains/ton/comment.test.ts` (new) covers the caps across amounts, the inactive-destination byte, UTF-8 byte counting, and the error text.

`packages/core/mpc/keysign/send/build.test.ts` covers the new early rejection: a jetton memo the old fixed cap would have signed is refused during the build, the cap follows the resolved amount and destination state, and the failure is a non-retryable `BuildKeysignPayloadError`.

`jetton.test.ts` / `native.test.ts` cover the resolver-level guard on both paths.

`yarn test:core` green (3257 passed). Typecheck, lint, knip, `format:check` and `check:shared-exports` clean — the pre-existing `packages/rujira` cosmjs skew and `@scure/bip39` wordlist errors are byte-identical to `main`, as are the SDK-unit and CLI test failures (verified against a stashed baseline).

## Additional context

Length validation is still missing on iOS and Android; both are tracked in the epic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TON memo validation for native and Jetton transfers.
  * Memo limits now account for UTF-8 byte size, transfer amount, and destination state.
  * Oversized memos are rejected before signing with a clear, non-retryable error.
  * Preserved valid boundary-length memos while rejecting comments that exceed capacity.
* **New Features**
  * Added shared TON memo capacity checks for transaction building and validation.
  * Added support for validating Jetton memos with transfer-specific limits.


Ehsan-saradar:
@coderabbitai review

rcoderdev (approved):

PR #2300, opened 2026-09-01
Fix SDK export smoke for unreleased local dependencies

### LINKED ISSUE #949: arch: package boundaries (tsconfig-alias deps + circular dep), broadcast-dedup, SDK-boundary breaks

rcoderdev:
Closes #949

## Summary by CodeRabbit

* **Chores**
  * Improved SDK package validation for installations that include local dependencies.
  * Added support for preparing complete dependency packages during consumer setup.

* **Tests**
  * Expanded validation coverage for dependency resolution and packaged consumer manifests.
  * Added checks to detect incompatible dependency versions before installation.


johnnyluo (approved):
LGTM

PR #2303, opened 2026-09-02
fix(sdk): expose partial broadcast failures

### LINKED ISSUE #2158: fix(broadcast): BroadcastPartialFailureError is exported but unreachable from broadcastTx()
## summary
`BroadcastPartialFailureError` is exported by `@vultisig/sdk`, but `broadcastTx()` never lets callers observe it directly because the outer catch wraps it into `VaultError(BroadcastFailed, ...)`.

## evidence
- `packages/sdk/src/vault/services/BroadcastService.ts:198-205` throws `BroadcastPartialFailureError` when a later leg fails after one or more tx hashes were already broadcast.
- `packages/sdk/src/vault/services/BroadcastService.ts:216-223` throws the same class when an approval tx broadcasts but confirmation fails before the swap leg is sent.
- `packages/sdk/src/vault/services/BroadcastService.ts:229-234` wraps every thrown error into `VaultError(BroadcastFailed, ...)`, so the exported partial-failure class never escapes as a direct runtime type.

## impact
Consumers cannot reliably `catch (e instanceof BroadcastPartialFailureError)` even though the SDK exports that class as if it were part of the public runtime contract.

## fix sketch
- either let `BroadcastPartialFailureError` escape as the public throw type for partial multi-leg failures
- or stop exporting a misleading unreachable class and formally project its fields onto the outer `VaultError` contract
- add unit tests that lock the intended public behavior

## effort
small to medium

## report
https://gist.github.com/gomesalexandre/2ae3a591eed2a7c21056c795aa50f2cf

rcoderdev:
Closes #2158

## Summary by CodeRabbit

- **New Features**
  - Exposed `BroadcastPartialFailureError` for failures occurring after one or more transactions have been submitted.
  - Added the number of successfully submitted transactions to partial broadcast errors.

- **Bug Fixes**
  - Improved handling of compilation, decoding, submission, approval, and confirmation failures during multi-transaction broadcasts.
  - Failed operations after earlier submissions now consistently report partial failures, while failures before any submission continue to report as general vault errors.
  - Error messages now clearly identify the number of submitted transactions.


johnnyluo (approved):
LGTM

PR #2306, opened 2026-09-02
feat(ton): explain failed transactions in plain language (seqno replay, expired deadline, fees)

### LINKED ISSUE #2197: [Add] TON: human-readable errors for exit codes 133/136 and clock skew [TON / Parity]
**What:** Translate TON's failure modes into plain language. Exit code 133 = seqno replay (someone/something sent in between), 136 = expired (`valid_until` passed — often device clock skew). Today these surface as opaque errors.
**Why:** Tonkeeper's most-read help article is its error taxonomy — these two codes plus clock skew dominate TON support load. We can absorb that load in-product instead of in tickets.
**Fix:** Map exit codes 133/136 (and the common compute-phase codes) to human messages with the right remedy ("check your device clock", "balance too low for fees — keep ~0.05 TON"), on all platforms via the shared status path.
Flagship-parity item under epic #2181.

Ehsan-saradar:
## Description

A TON send that fails today comes back as an opaque `exitcode=133` inside a toncenter rejection, or as a bare `error` status with nothing attached. The two codes that dominate TON support — a replayed seqno (wallet v4 `33`, W5 `133`) and an expired `valid_until` (`36` / `136`, almost always a device clock that drifted) — look identical to the user and need opposite fixes: one means "check your history, it may already be there", the other "fix your device clock and send again".

**Taxonomy** (`@vultisig/core-chain/chains/ton/failure`). A failure becomes a `TonTxFailure`: a stable `reason`, the `phase` it came from, the raw `exitCode`, and an English `message` that says what happened and what to do.

| reason | codes |
|---|---|
| `seqno-mismatch` | compute 33 / 133 |
| `expired` | compute 36 / 136 |
| `invalid-signature` | compute 35 / 135 |
| `wallet-id-mismatch` | compute 34 / 134 |
| `out-of-gas` | compute 13 / -14 |
| `not-enough-jettons`, `jetton-gas-underfunded`, `jetton-unauthorized` | compute 706 / 709 / 705, 707 (TEP-74 jetton wallet) |
| `invalid-destination` | action 36 |
| `insufficient-funds` | action 37, 40, or `no_funds` |
| `action-failed` | action skipped / unsuccessful without a code |
| `aborted` | aborted with no readable phase |
| `contract-rejected` | anything else, reported by number |

Action-phase `36` is an invalid destination and is kept apart from the wallet's `36` (expired), which can only come from the compute phase. Messages carry the remedy Tonkeeper's help articles give ("make sure your device's date and time are set automatically", "keep about 0.05 TON spare for fees").

**Status path.** `TxStatusResult` gains an optional `failure?: TxFailureInfo` (`reason`, `message`, `exitCode?`), and `getTonTxStatus` fills it for every `error`.

**Broadcast path.** `broadcastTonTx` parses the wallet contract's `exitcode=N` out of toncenter's `sendBoc` refusal and returns the failure as a non-retryable `TonBroadcastRejectedError` in the result's `cause`: its `message` is the human explanation, the raw toncenter error stays in `cause.cause`. Re-sending the same bytes can only be refused again, so the transport retry no longer spins on it. The existing hash verification still turns a seqno refusal into an acceptance when a co-signer already landed the same message.

The types, `getTonTxFailure`, `parseTonBroadcastRejection` and `TonBroadcastRejectedError` are re-exported from `@vultisig/sdk`.

Fixes #2197

## Which feature is affected?

- [x] Sending - Please attach a tx link here
- [ ] Swap - Please attach a tx link for the swap here
- [x] New Chain / Chain related feature - Please attach tx link here

Nothing in the signing or broadcast bytes changes; this classifies failures after the fact. No new transaction was broadcast for it. The rejection format is the one toncenter returns live (`{"ok":false,"error":"…inbound external message rejected by transaction <hash>:\nexitcode=133, steps=49, gas_used=0…","code":500}`), and the test suite pins it verbatim.

## Checklist

## Additional context

- 44 new or updated tests: the code tables (both wallet generations, the phase split for 36, fee codes), the toncenter rejection parser against a real payload, the status resolver now asserting the explained reason for every failure shape, and the broadcast resolver (accepted / co-signer-won seqno refusal / explained non-retryable refusal / unexplained transport failure). `yarn check` and `yarn test:core` pass.
- Changeset: `@vultisig/core-chain` minor, `@vultisig/sdk` minor.
- Windows renders these in vultisig/vultisig-windows (draft until the core-chain pin bump). iOS and Android currently show `"Compute phase exited with code N"` from their own status providers; this module is the shared reference to port.

## Summary by CodeRabbit

- **New Features**
  - Added detailed transaction failure information, including reasons, messages, execution phases, and exit codes.
  - Added expired transaction statuses and partial TON action-failure classification.
  - Exposed TON failure details, parsing helpers, and error types through the SDK and React Native.
  - TON broadcast rejections now provide actionable, non-retryable errors while preserving the original cause.

- **Bug Fixes**
  - Improved TON failure classification across compute, action-phase, and broadcast rejection scenarios.

- **Documentation**
  - Updated token discovery documentation with Cardano, Ripple, and verified TON jetton support.


rcoderdev (commented):
Action code 40 does not always mean insufficient funds: TON also returns it for oversized messages or excessive Merkle depth ([TON documentation](https://docs.ton.org/foundations/messages/modes#errors)). Mapping every 40 to `insufficient-funds` tells users to add TON and retry even when the dApp payload or StateInit must change. Please keep this code generic unless an authoritative signal establishes a funding failure, and add coverage for code 40 without `no_funds`; the current test pins the incorrect unconditional mapping.

Ehsan-saradar:
Agreed, fixed in 85bd0fe76.

The docs read 40 as "Cannot process a message — not enough funds, the message is too large, or its Merkle depth is too big", so it is out of the unambiguous table. Unnamed action codes now defer to the phase's own `no_funds` flag: 40 with it is `insufficient-funds`, 40 without it stays `action-failed` with the code attached. 37 keeps its unconditional mapping, since it means exactly "Not enough GRAMs".

The generic action-phase message lost its "check the balance" advice for the same reason, since that copy is now what an oversized message gets. New tests pin 40 both ways and assert the generic message mentions neither a top-up nor a balance.

The matching English string on the windows side needs the same reword; I will do that on vultisig/vultisig-windows#4808.

rcoderdev (commented):
The new TON failure types, helpers, and error class are exported from the generic SDK root but not from the hand-curated React Native entry. Because the package maps React Native runtime and types to that entry, the named @vultisig/sdk imports advertised here fail for RN consumers. Please mirror the RN-safe exports and add a React Native public-surface assertion.

Ehsan-saradar:
Good catch, fixed in e5648ab53.

The RN entry now mirrors the types, the helpers and the error class. It matters more than a missing re-export: RN broadcasts TON itself, and `chains.ton.broadcastTonTx` rejects with toncenter's own text, which carries the wallet contract's `exitcode=<n>` and nothing else — so `parseTonBroadcastRejection` is exactly what an RN consumer needs there.

Two assertions in `entry.test.ts`: one pins the four values by identity against `@vultisig/core-chain/chains/ton/failure`, the other feeds the exact rejection that helper raises through the RN surface and asserts the caller gets `expired` with the device-clock remedy rather than a bare 136.

`TxFailureInfo` is deliberately not mirrored — the RN entry exposes no `getTxStatus`, so the type would have nothing to describe there. Say the word if you would rather have it for symmetry.

rcoderdev (commented):
`skipped_actions > 0` can coexist with successfully created outgoing messages, but `action-failed` currently says “nothing was sent” and suggests retrying. For example, `{ success: true, result_code: 0, skipped_actions: 1, msgs_created: 1 }` produces that message, potentially prompting duplicate transfers. Please use a neutral partial-failure explanation unless outgoing-message evidence proves zero transfers, and cover a mixed-success batch. [TON documents partial batch success](https://docs.ton.org/contracts/standard/wallets/highload/v3/verify-is-processed).

Ehsan-saradar:
Agreed, fixed in 18f932357.

A generic action-phase failure now claims "nothing was sent" only when the node reports `msgs_created: 0`. Anything else becomes `action-partially-failed`, whose message says at least one transfer did not happen, that others may have, and to check the transaction history before sending again — no retry advice.

A node that reports no count at all gets the neutral wording too. Absent evidence is not evidence of zero, and the asymmetry is one-sided: the neutral message costs a history check, the old one costs a duplicate transfer.

The same rule covers an unnamed result code, so `{ result_code: 33, skipped_actions: 1, msgs_created: 2 }` is partial rather than "nothing was sent". Tests cover your example verbatim, a proven-zero batch, the no-evidence case, and the unnamed-code pair; the status-resolver table gained a proven-zero row alongside the skipped one.

Heads-up for the consumer side: `TonTxFailureReason` gains a member, so windows' `Record<TonTxFailureReason, string>` mapping needs the new key at the pin bump — it fails the typecheck until then rather than silently falling back.

Verification: `yarn check` clean, core suite 3,465 passed, SDK unit suite 3,738 passed.

rcoderdev (approved):

rcoderdev (approved):

PR #2310, opened 2026-09-02
fix(ton): let a dApp deadline tighten the signed expiry

### LINKED ISSUE #2309: [Fix] TON: a dApp's valid_until never reaches the signed expiry — getTonChainSpecific always signs now + 600 [TON / S2]
**Problem:** \`getTonChainSpecific\` sets \`expireAt\` to \`now + 600\` unconditionally (\`packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.ts\`). A TonConnect \`sendTransaction\` carries \`valid_until\` — the deadline after which the dApp considers the transaction invalid — but the chain-specific input has no way to pass it, so the signed wallet message expires 10 minutes after build regardless of what the dApp asked for.

**Why it matters:** The dApp's deadline is the only expiry the dApp sees. A request with \`valid_until\` 60 s out is signed with a 600 s expiry: if broadcast lands after the dApp's window (slow approval, secure-vault co-signing), the dApp treats it as expired and may retry or re-quote while the chain still accepts and executes the first one. The wallet's own 10-minute cap should still apply, so the right value is \`min(now + 600, valid_until)\`; a deadline already in the past at build time should fail closed rather than sign a transaction that cannot land.

**Fix:** Add an optional \`validUntil\` (unix seconds) to the \`tonSpecific\` branch of \`GetChainSpecificInput\`; in \`getTonChainSpecific\`, use \`min(now + 600, validUntil)\` as \`expireAt\` and throw when \`validUntil\` is already past. Co-signers read \`expireAt\` from the payload, so this only changes what the initiator writes. The extension then forwards the bridge's \`valid_until\` (it already validates it, see vultisig/vultisig-windows).

**Receipts:** \`packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.ts\` (\`expireAt: BigInt(Math.floor(Date.now() / 1000) + 600)\`); \`packages/core/mpc/keysign/chainSpecific/resolver.ts\` (no \`tonSpecific\` input branch).

Part of epic #2181.

Ehsan-saradar:
## Description

`getTonChainSpecific` signed `expireAt = now + 600` unconditionally. A TonConnect `sendTransaction` carries `valid_until` — the deadline after which the dApp considers the request dead — but the chain-specific input had no way to pass it, so a request with a 60-second window was signed with a ten-minute one. A broadcast that lands after the dApp's window can still execute on chain while the dApp has already treated it as expired and retried or re-quoted.

- `GetChainSpecificInput` gains a `tonSpecific` branch with an optional `validUntil` (unix seconds).
- `getTonChainSpecific` signs `expireAt = min(now + 600, validUntil)`: the wallet's own ten-minute window stays the ceiling, a dApp deadline can only tighten it.
- The deadline is floored to whole seconds first; one already in the past at build time — or less than a second away, which would floor to `now` — throws (`… has already passed`) instead of signing a transaction the network would reject. The bridge checks expiry when the request arrives, but the user can sit on the approval screen past a short window.
- Callers that pass nothing keep the previous behaviour. Co-signers read `expireAt` from the payload, so only what the initiator writes changes; no signing-hash impact for existing flows.

Fixes #2309 — part of epic #2181. Consumer side: vultisig/vultisig-windows#4810 forwards the bridge's validated `valid_until` once this ships and the pin is bumped (vultisig/vultisig-windows#4735).

## Which feature is affected?

- [x] Sending — TON dApp (`signTon`) transactions with a `valid_until`

## Checklist

New tests in `packages/core/mpc/keysign/chainSpecific/resolvers/ton/index.test.ts` (fake timers): no deadline → `now + 600`; sooner deadline → that deadline; later deadline → `now + 600`; fractional deadline → floored; sub-second deadline → rejects; past deadline → rejects.

Verification: `yarn vitest run` on the TON chain-specific suite — 23 passed; ESLint + Prettier (repo configs) clean; `tsc --noEmit -p .config/tsconfig.shared-publish.json` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * TON transactions now honor optional dApp expiration deadlines.
  * Transactions use the earlier of the wallet’s ten-minute limit and the provided deadline.
  * Expired deadlines are rejected, including deadlines that expire during network lookups.
  * Expiration is calculated after required network lookups complete.
  * Existing ten-minute expiration behavior is preserved when no deadline is provided.

* **Documentation**
  * Added release metadata for the TON expiration update across core and SDK packages.


rcoderdev (commented):
The resolver validates and captures expireAt before awaiting bounceability and, for Jettons, additional RPC metadata. A deadline that is valid at entry can expire during those awaits and still be returned, producing a payload the network immediately rejects. Please compute and validate expireAt after the asynchronous metadata work, immediately before return, and add a fake-timer regression where the RPC advances past validUntil.

Ehsan-saradar:
Good catch — fixed.
**The deadline.** `expireAt` is now computed immediately before the return, after the seqno, bounceability and jetton-metadata lookups, instead of at entry. Two fake-timer regressions in `resolvers/ton/index.test.ts` drive it: one advances the clock past `validUntil` from inside the jetton-wallet mock and asserts the build is refused, the other advances it by 30s and asserts the wallet window is measured from the finished payload (`now + 30 + 600`). Both fail against the old ordering.

Worth noting the second half of that: reading the clock up front also spent part of the wallet's own ten-minute window on the lookups, so a slow RPC quietly shortened the signed validity. It now starts when the payload is finished.

rcoderdev (approved):

PR #2312, opened 2026-09-02
fix: cache rn-path-override resolveId to avoid redundant full resolutions

gomesalexandre:
tl;dr the RN override plugin's resolveId hook had zero caching, re-running Rollup's full resolution chain on every single import in the graph.

## what

\`rnOverridePlugin\`'s \`resolveId\` called \`this.resolve()\` (which re-runs the full alias + node-resolve + commonjs chain) for every non-entry import, with no memoization. Widely-imported override targets (the solana/sui clients this plugin exists to intercept) get asked for repeatedly by many different importers across the module graph, and each ask paid that full nested-resolution cost again.

Fix: memoize by \`(importer, source)\` - deterministic for a given build - caching the in-flight promise too so concurrent asks for the same pair don't race duplicate resolutions. Pure perf fix, zero behavior change: same override targets, same resolution results, just cached.

## why

Suspected contributor to the mcp repo's integration-test CI intermittently timing out on the react-native platform build. Across 4+ independent PRs there today, the build logs consistently showed the SAME override targets (solanaClient.ts, suiClient.ts, LifiSwapEnabledChains.ts) getting resolved over and over for 10-20+ minutes before the job hit SIGTERM - unrelated to what any of those PRs actually changed.

Honest caveat: I can't cheaply reproduce CI's resource-constrained runner locally, so this isn't *proven* to be the sole cause of those timeouts (mcp's harness has other, separate infra issues too - broken DB migration state, RPC auth). But it's a correct, low-risk, verified optimization regardless of whether it fully explains the CI symptom.

## testing

- \`yarn build:platform:react-native\` - completes clean in ~8.5s (previously the thing timing out in CI)
- \`yarn build:fast\` - full node build, exit 0, no new warnings
- No changeset - this only touches build tooling (\`rollup.platforms.config.js\`), not published package output.

https://claude.ai/code/session_01WgN3XsnuidNH78GH3HQXxx

## Summary by CodeRabbit

* **Performance**
  * Improved module resolution consistency and efficiency during builds by reusing repeated resolution results.
  * Resolution state is reset between builds to ensure fresh results.


rcoderdev:
Closing this because Closing this cache experiment because its key omits resolver options: a CommonJS request can reuse an ESM override and resolve the wrong module. Main still uses the correct uncached resolver; please resubmit with an options-aware key and a mixed-request regression test if the optimization remains useful.

PR #2314, opened 2026-09-03
fix: reuse fresh local SDK builds in browser gates

rcoderdev:
Browser preparation now reuses completed shared-package and SDK builds when their inputs and required outputs are unchanged. Successful builders record source inventories, so generated directory and manifest churn no longer causes another full SDK build. Source edits, renames, deletions, and missing artifacts still invalidate the relevant build.

Generated shared export manifests are written only when their contents change.

Verified unchanged preparation completed in 450 ms without rebuilding. A real missing-output recovery and production browser build passed, and both WASM assets were served successfully through Vite with bytes matching the production artifacts.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved browser SDK build validation to detect stale, incomplete, or failed artifacts more reliably.
  - Improved handling of source changes occurring during builds.
  - Prevented unnecessary updates to shared package manifests when their content is unchanged.

- **Build Improvements**
  - Coordinated shared-package and browser SDK builds to keep generated artifacts synchronized.
  - Added reliable build status tracking across multi-step processes.
  - Expanded automated coverage for build freshness, failures, and generated exports.


johnnyluo (approved):
LGTM

PR #2316, opened 2026-09-03
feat(ripple): support issued-currency payments

### LINKED ISSUE #996: feat(ripple): send issued-currency tokens (Payment with CurrencyAmount + delivered_amount safety)
## What
Extend the XRP send path to send **issued currencies** (tokens), not just native XRP drops.

## Fix
Build `OperationPayment` with a `CurrencyAmount{currency, issuer, value}` instead of drops. Because token payments can be partial, set the partial-payment flag intentionally and read **`delivered_amount`** (not `Amount`) when confirming delivered value (see spike #982 — becomes higher-stakes for tokens).

## Acceptance
- [ ] Token send builds a CurrencyAmount Payment, signs, broadcasts
- [ ] delivered_amount used for confirmed value; partial-payment handled
- [ ] Unit test for issued-currency amount encoding

Part of the **XRP token support** feature set (all platforms, low-prio backlog). Currently XRP is Payment/XRP-only. Feasibility: WalletCore Ripple proto directly models OperationTrustSet + OperationPayment CurrencyAmount; signs via secp256k1 MPC — no blocker. · Companion: #982 (partial-payment read).

## References
- Fungible tokens: https://xrpl.org/docs/concepts/tokens/fungible-tokens
- Partial payments: https://xrpl.org/docs/concepts/payment-types/partial-payments
- WalletCore Ripple proto (CurrencyAmount): https://github.com/trustwallet/wallet-core/blob/master/src/proto/Ripple.proto

rcoderdev:
Closes #996

## Summary by CodeRabbit

* **New Features**
  * Added support for sending XRPL issued currencies through XRP payment flows.
  * Added explicit Payment and TrustSet transaction handling, including destination tags.
  * Added authoritative delivery details to successful XRPL payment receipts for XRP, issued currencies, and MPTs.
  * Exposed issued-currency transaction types and payment options in the SDK.
* **Bug Fixes**
  * Improved validation for currency codes, amounts, issuer addresses, precision, and supported values.
  * Improved handling of unfunded destinations and invalid transfers.
  * Preserved native XRP payments and TrustSet behavior.


packages/core/mpc/keysign/utils/isRippleTrustSet.ts
```diff
@@ -49,23 +45,14 @@ export const originatesRippleTrustSet = (keysignPayload: KeysignPayload): boolea
...
  */
 export const isRippleTrustSet = (keysignPayload: KeysignPayload): boolean => {
   const { transactionType } = getBlockchainSpecificValue(keysignPayload.blockchainSpecific, 'rippleSpecific')
 
-  if (transactionType === TransactionType.RIPPLE_TRUST_SET) {
-    return true
-  }
-
-  return hasRippleTrustSetShape(keysignPayload)
+  return transactionType === TransactionType.RIPPLE_TRUST_SET
```
johnnyluo: `UNSPECIFIED` cannot safely mean “Payment” here because proto3 default/absent is also what older TrustSet payloads carry. With the shape fallback removed, an updated signer receiving a legacy non-native Ripple TrustSet payload can rebuild it as an issued-currency Payment instead of preserving the old TrustSet interpretation. Please add an explicit Payment discriminator (for example `RIPPLE_PAYMENT`) and keep the shape-based TrustSet fallback only for truly absent/default legacy payloads.

packages/sdk/src/platforms/react-native/chains/ripple/tx.ts
```diff
@@ -240,6 +255,15 @@ export function buildXrpSendTx(opts: BuildXrpSendOptions): BuildXrpSendResult {
   const memoIsLegacyCarrier =
     opts.memo !== undefined && (legacyMemoDestinationTag !== undefined || opts.memo === destinationTag?.toString())
   const hasDistinctMemo = Boolean(opts.memo && !memoIsLegacyCarrier)
+  if (opts.issuedCurrency && BigInt(opts.amount) <= 0n) {
+    throw new Error('XRP issued-currency Payment amount must be positive')
+  }
+  const amount: XrpPaymentTx['Amount'] = opts.issuedCurrency
+    ? getSignableIssuedCurrencyAmount({
```
johnnyluo: This React Native path now imports `getSignableIssuedCurrencyAmount`, but that helper reaches `issuedCurrency.ts` where `toXrplCurrencyCode()` uses Node `Buffer` APIs. This module is explicitly kept Hermes-safe/no-Buffer, so `buildXrpSendTx({ issuedCurrency: { currency: "RLUSD", ... } })` can fail in RN apps without a Buffer polyfill. Please make the currency-code normalization byte/TextEncoder-based or provide an RN-safe helper for this path.

packages/core/mpc/keysign/utils/isRippleTrustSet.ts
rcoderdev: Fixed in e93ae116d. New sends now carry an explicit RIPPLE_PAYMENT discriminator, while UNSPECIFIED/absent legacy payloads retain the issuer-address TrustSet fallback. Added regression coverage for both a legacy undiscriminated TrustSet and an explicit issued-currency Payment back to the issuer.

packages/sdk/src/platforms/react-native/chains/ripple/tx.ts
rcoderdev: Fixed in e93ae116d. XRPL currency normalization no longer uses Buffer: it validates ASCII and hex-encodes/pads with platform-neutral number/string operations, so the React Native path is Hermes-safe. The focused Ripple suite passes (130 tests), and typecheck/lint are clean.

johnnyluo (approved):
LGTM

PR #2319, opened 2026-09-04
fix(swap): validate Base58 limit-swap destinations

### LINKED ISSUE #2268: fix(swap): Tron limit-swap destination has no checksum — regex only
`limitSwapMemo.ts:319` validates a Tron limit-swap destination with `^T[base58]{33}$` and nothing else. A one-character typo passes and gets baked into an immutable THORChain memo — output goes to an address nobody controls.

Fix: run the address through a checksum validator (`walletCore.AnyAddress.isValid` or `base58CheckTronDecode`, both already in the repo) in `assertValidLimitSwapDestinationAddress`.

Same shape-only pattern also covers Dash/Zcash/Ripple in that map — fix the class.
S1. Found in the 2026-08-30 Tron hardening audit.

rcoderdev:
Closes #2268

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved THORChain limit-swap destination validation for Dash, Zcash, Ripple, and Tron.
  - Rejects destinations with invalid checksums, incorrect network versions, or unsupported lengths.
  - Prevents malformed or wrong-network addresses from being accepted when building or parsing swap memos.

- **Tests**
  - Added coverage for valid, oversized, checksum-invalid, and wrong-network destination addresses across supported chains.


johnnyluo (approved):
LGTM

PR #2323, opened 2026-09-05
fix(ripple): refuse issued-currency Payments the ledger would reject

### LINKED ISSUE #2322: fix(ripple): issued-currency Payments miss SendMax and destination trust-line preflight — sends fail on-chain
XRPL issued-currency Payments build an exact `Amount` and nothing else. Two of the gaps burn a fee on-chain; the rest are hardening.

### 1. No `SendMax` — a token with a transfer fee cannot be sent

Neither encoder emits `SendMax`, and nothing reads `TransferRate`:

```
grep -rn "TransferRate|SendMax|transferRate" packages/core/chain packages/core/mpc packages/sdk/src
→ only the tfPartialPayment comment in signingInputs/resolvers/ripple.ts
```

XRPL defaults `SendMax` to `Amount`. When the issuer sets a non-zero `TransferRate` the sender is debited `Amount x (1 + rate)`, which exceeds that implied ceiling → `tecPATH_PARTIAL`: fails on-ledger, fee burned, sequence consumed. RLUSD is rate-0 so the curated token works, but `findRippleCoins` surfaces *every* positive-balance trust line into the asset list and the send flow, and transfer fees are common on community tokens.

Fix: read `AccountRoot.TransferRate` from the issuer, set `SendMax = ceil(Amount x rate / 1e9)` in both `signingInputs/resolvers/ripple.ts` and `platforms/react-native/chains/ripple/tx.ts`. No fee applies when either end is the issuer, so skip it for redemptions.

### 2. No destination trust-line preflight

`chainSpecific/resolvers/ripple.ts:98-106` rejects an unactivated destination. The far more common failure is a **funded** destination with no line for `(currency, issuer)` → `tecNO_LINE` / `tecPATH_DRY`, fee burned. Half a preflight reads as a whole one.

`getRippleAccountLines` already exists, and the resolver already makes a destination `account_info` round trip — this is one more call at the same point.

Fix: `account_lines` on the destination; reject as `BuildKeysignPayloadError` when the line is absent.

### 3. 3-char currency codes containing lowercase are rejected — regression on TrustSet

`chains/ripple/issuedCurrency.ts:130` — `standardCharacters` omits lowercase. XRPL standard codes explicitly permit "all uppercase and lowercase letters":

```
USD  accept   |  usd  REJECT  |  btc  REJECT  |  sol  REJECT  |  a1b  REJECT
solo accept (hex-encoded, 4 chars)
```

A token discovered from `account_lines` with such a code shows a balance but throws on send. And because `getIssuedCurrencyAmount` now feeds the TrustSet path too, **opening a trust line for such a currency throws at signing-input construction — after the user has reviewed — where it previously signed.** In a multi-device vault an older peer signs while the upgraded peer throws.

Fix: add `a-z` to `standardCharacters`.

### 4. Money-path input errors are bare `Error`

`send/build.ts:127` (positive-amount, 16-significant-digit) and `chainSpecific/resolvers/ripple.ts:103` (destination not activated) are deterministic bad input. The same files use `BuildKeysignPayloadError` for exactly this class — `assertTonMemoFits`, `ripple-destination-tag-required` — so callers stop retrying and surface the message. A bare `Error` gets retried and shown as a generic failure.

### 5. TrustSet to a non-existent issuer reports a send error

The `destinationUnfunded` branch is not gated on `isTrustSet`, so opening a trust line against an issuer whose account does not exist yields *"Cannot send an XRP issued currency to rXXX... Activate it with the XRP base reserve before sending tokens."* Failing closed is right; the message tells the user to fund the issuer.

### 6. No cross-encoder golden for an issued-currency Payment

`testdata/cross-encoder-golden/ripple-payment{,-memo}.json` pin WalletCore against `ripple-binary-codec` because the protobuf path emits `Flags: 0` while the raw-JSON path omits it — per the fixture, that "changes the preimage and breaks mixed-platform MPC". An issued-currency Payment now goes through both encoders (`opPayment.currencyAmount`, and a hand-built rawJson for the memo case) with no golden pinning them.

Verified they currently agree byte-for-byte, memo and no-memo — so this is a missing regression guard, not a live break. Worth adding while the mechanism exists.

### 7. Mixed-version TrustSet byte-identity is not asserted

No test asserts that a legacy undiscriminated TrustSet payload (`transaction_type` unset) signs byte-identically to a discriminated one. `keeps legacy undiscriminated TrustSet payloads compatible` only asserts `opTrustSet` is truthy, and `fails closed across a token-payment committee containing a legacy shape-inference signer` hand-constructs its "legacy" input, so it exercises no legacy code path and cannot fail.

---

1 and 2 are S2 — user funds, on-chain failure. 3 is a regression on a path that worked before. The rest are S3.

Ehsan-saradar:
Closes #2322.

An XRPL issued-currency Payment is built exactly, with no `SendMax` and no destination pre-flight. Two shapes are accepted today and fail on-ledger — an *applied* transaction either way, so the fee is burned and the sequence consumed:

| Shape | Ledger result |
|---|---|
| Destination holds no trust line for that `(currency, issuer)` | `tecNO_LINE` / `tecPATH_DRY` |
| Issuer sets a `TransferRate` | `tecPATH_PARTIAL` — the sender owes `Amount x (1 + rate)`, above the `SendMax` XRPL defaults to `Amount` |

`getRippleChainSpecific` already round-trips the destination's `account_info`, so both pre-flights land at a point that already talks to the ledger: `account_lines` on the destination, and the issuer's `TransferRate`. Lookup failures stay retryable `Error`s — a transient RPC failure is not evidence a line is missing.

### On the transfer-fee case

Covering the fee needs an explicit `SendMax`, and there is nowhere to put one: `RippleSpecific` has no field for it, and `TW.Ripple.Proto.OperationPayment` has only `amount` / `currencyAmount` / `destination` / `destinationTag`. Carrying one would mean a new proto field in commondata plus routing every issued Payment through the raw-JSON path. So this refuses the send instead of building one that burns a fee. Redemptions to the issuer are unaffected — no transfer fee applies there, and no trust line is needed — and both pre-flights skip that case.

### Lowercase standard currency codes

XRPL standard 3-character codes [permit both letter cases](https://xrpl.org/docs/references/protocol/data-types/currency-formats), so `usd` is a currency in its own right rather than a misspelling of `USD`. They were rejected, which stranded any trust line the ledger reported with one: the token showed a balance but could not be sent, and — because the signable-amount helper also feeds the TrustSet path — *opening* a line for such a currency threw at signing-input construction, after review, where it previously signed. Note the arbitrariness this removes: `solo` was accepted (4 chars, hex-encoded) while `sol` was not.

`XRP` and its lookalikes stay reserved, now matched case-insensitively so `xrp` cannot be signed as a token.

### Error types

A bad issued amount, an unactivated destination, and a trust line to a non-existent issuer were bare `Error`s. They are deterministic bad input, so they now use `BuildKeysignPayloadError` like `ripple-destination-tag-required` and `ton-memo-too-long` already do — callers stop retrying and show the reason. A TrustSet against an issuer that does not exist also stopped reporting *"the destination account is not activated — activate it before sending tokens"*, which told the user to fund somebody else's account.

### Tests

- **Cross-encoder goldens for issued currency.** `testdata/cross-encoder-golden/` exists because WalletCore's protobuf path emits `Flags: 0` while its raw-JSON path omits it, and, per the fixture, "omitting it is network-equivalent but changes the preimage and breaks mixed-platform MPC". Issued currency went through both encoders with nothing pinning them. The two new vectors carry the full 16 significant digits an XRPL mantissa allows, and the expected hash is derived from the WalletCore side and independently reproduced by the RN `ripple-binary-codec` builder. They agree today — this keeps them agreeing.
- **Mixed-version TrustSet byte-identity** is asserted again: an undiscriminated payload and a discriminated one must serialize the same bytes, or a committee spanning an upgrade diverges and the ceremony never completes.
- New pre-flight tests were mutation-checked — disabling each guard fails exactly the tests that cover it.

### Out of scope

Trust-line *limit* checking (a line that exists but is full also yields `tecPATH_DRY`) is left out deliberately: the limit interacts with rippling and quality settings, and a false rejection blocks a legitimate send. Actual `SendMax` support needs the proto field described above.

### Validation

`lint`, `format:check`, `knip`, `changeset:check-sdk-bundle` and the core typecheck all pass. `test:core` 3444 passed, `@vultisig/sdk` 3738 passed. The `clients/cli` suite fails identically on pristine `origin/main` in this environment (unbuilt CLI artifacts) — 65 files / 28 tests, byte-for-byte the same before and after.

## Summary by CodeRabbit

- **New Features**
  - Added Ripple issued-currency preflight checks for destination activation, trust lines, issuer status, transfer fees, and valid amounts.
  - Added clear errors for invalid destinations, missing trust lines, unsupported fees, inactive issuers, and invalid amounts.
  - Added Ripple trust-line lookup support for issued-currency balances.

- **Bug Fixes**
  - Lowercase standard currency codes are now accepted while native XRP variants remain protected.
  - Improved consistency of Ripple transaction signing data across supported encoders.


Ehsan-saradar:
@coderabbitai review

johnnyluo (approved):
LGTM

PR #2325, opened 2026-09-05
fix(evm): size the gas reserve from the transaction instead of a flat 600k floor

### LINKED ISSUE #4799: [Fix] EVM fee reserve sits on a flat 600k gas floor; network fee differs up to 4.7x across the three clients [Swap / fees]
# [Fix] EVM swap network fee differs up to 4.7x between extension, iOS and Android [Swap / fees]

## Problem

The same swap, quoted at the same moment through the same provider, shows three different network fees:

| product | displayed network fee |
|---|---|
| extension / desktop | `0.00108772 ETH` (~$2.67) |
| iOS | `0.000550798377154206 ETH` (~$1.35) |
| Android | `~0.00023–0.00039 ETH` (~$0.57–0.96) |

Observed side by side on ETH → USDC on Ethereum, ~$5.00 in, KyberSwap route, outputs matching to six digits (`4.987292` vs `4.987271` USDC).

On a $5 swap the extension presents $2.67 of "network fee". That is over half the trade, and it is not what the user pays.

This issue documents where each number comes from so the team can pick one definition. **It is a reporting difference, not a difference in money moved** — see *What is actually paid* below.

## What each product computes

Read from source in all three repos. Ethereum, EVM aggregator swap, native ETH source.

**Extension / desktop** — `@vultisig/core-chain` + `@vultisig/core-mpc`

```
gasLimit = ceil(1.5 * max(eth_estimateGas, thirdParty, 600_000))
maxFee   = 1.5 * baseFee + tip
tip      = clamp(rpc estimateMaxPriorityFeePerGas, floor 1 gwei, ceiling 500 gwei)
fee      = gasLimit * maxFee            (+ OP-stack L1 surcharge)
```

```
core-chain/tx/fee/evm/evmGasLimit.js         → deriveEvmGasLimit: 600000n when data
core-mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.js → baseFeeMultiplier 15/10, gas buffer 3/2
core-mpc/keysign/chainSpecific/resolvers/evm.js      → maxFeePerGas = baseFeePerGas + priority
core-mpc/keysign/fee/resolvers/evm.js                → gasFee = maxFeePerGasWei * gasLimit
```

**iOS** — `VultisigApp/Core/Services/Fee/FeeService.swift`, `BlockChainService.swift`

```
gasLimit = 1.5 * eth_estimateGas(route calldata)     // inflatedGasLimit
maxFee   = (baseFee * 1.1) * 1.2 + tip = 1.32 * baseFee + tip
tip      = max(rpc priorityFee, 1 gwei)
fee      = gasLimit * maxFee
```

**Android** — `EthereumFeeService.kt`, `SwapGasCalculator.kt`, `SwapTransactionBuilder.kt`

```
displayLimit = max(quote.tx.gas, 40_000)             // no 1.5x buffer
maxFee       = (baseFee * 1.5) * 1.2 + tip = 1.8 * baseFee + tip
tip          = clamp(feeHistory.last, floor 0.5 gwei, cap 10 gwei)
fee          = displayLimit * maxFee                 (+ OP-stack L1)
```

All three display exactly what they sign — none of them shows a separate estimate. Android's signed limit is `maxOf(quote.tx.gas, ethSpecific.gasLimit)` in `OneInchSwap.kt`, which equals its display limit.

## Where the difference comes from

Reconstructing the two screenshot figures from the two formulas gives a consistent solution:

```
extension gasLimit = 1.5 * max(estimate, 600_000) = 1.5 * 600_000 = 900_000
iOS       gasLimit = 1.5 * estimate               = 1.5 * 310_252 = 465_378

gas-limit factor  900000 / 465378 = 1.9339
gas-price factor  1.2086 / 1.1836 = 1.0211
product                           = 1.9748   <- matches the observed ratio exactly
```

Solving the two equations yields `baseFee = 0.139040 gwei` and `tip = 1.000017 gwei`. The tip lands on the 1 gwei floor that iOS and the extension each apply independently — that value was free in the system, so it is an external check on the model, not a fit.

**Roughly 97% of the gap is the gas limit, not the gas price.** The base-fee multiplier accounts for ~2%.

## Root cause

```js
// core-chain/tx/fee/evm/evmGasLimit.js
export const deriveEvmGasLimit = ({ coin, data }) => {
  if (data) {
    return chain === EvmChain.Mantle ? 3000000000n : 600000n
  }
  ...
}
```

```js
// getEvmFeeQuote.js
minimumGasLimit: deriveEvmGasLimit({ coin, data: getData() })
gasLimit = bigIntMax(estimatedGasLimit, thirdPartyGasLimitEstimation, minimumGasLimit)
```

The parameter is named `minimumGasLimit` and reads as a fallback for when estimation fails. Because it is applied with `bigIntMax`, it also overrides a **successful, lower** estimate. The route here estimated at ~310k; the floor raised that to 600k, and the calldata buffer then took it to 900k.

The problem is not that the extension is conservative. It is that its headroom does not scale with the transaction:

```
route needs ~310k  ->  extension reserves 900k  ->  x2.9
route needs ~150k  ->  extension reserves 900k  ->  x6.0
route needs ~800k  ->  extension reserves 1.2M  ->  x1.5
```

The cheaper and simpler the swap, the more the flat floor dominates — which is exactly where the user notices it.

## The floor is not swap-specific

The swap screenshot is only where it was noticed. `deriveEvmGasLimit` keys off `data`, and the chain-specific resolver feeds it the memo whenever there is no general-swap calldata:

```js
// core-mpc/keysign/chainSpecific/resolvers/evm.js
const getData = () => {
  const swapPayload = getKeysignSwapPayload(keysignPayload)
  if (swapPayload && 'general' in swapPayload) {
    const value = swapPayload.general.quote?.tx?.data
    if (value) return value
  }
  return keysignPayload.memo
}
```

`getEvmFeeQuote` then applies the 1.5x calldata buffer on the same condition. So every EVM transaction carrying data takes the same 600k floor, not just aggregator swaps:

```
ETH send with a memo         -> floor 600k -> x1.5 -> 900k
THORChain/Maya router deposit-> floor 600k -> x1.5 -> 900k
generic contract call (stake)-> floor 600k -> x1.5 -> 900k

ETH send, no memo            -> 23k                      (correct)
ERC-20 transfer, no memo     -> 120k floor vs ~50k real   (x2.4, no buffer)
```

Verified from source; no measurement needed for this part. Only the swap row has a measured cross-platform comparison behind it.

## Conservatism is two axes, not one

Worth stating explicitly, because "the extension is the most pessimistic" is only half true:

```
                gas-limit headroom   baseFee headroom   tip floor
extension       x2.9  (600k floor)   x1.50              1.0 gwei
iOS             x1.5  (of estimate)  x1.32              1.0 gwei
android         x1.0  (route gas)    x1.80              0.5 gwei
```

Android has the **least** gas-limit headroom but the **most** base-fee headroom. The two guard different failures: too little gas limit means the tx reverts and the spent gas is lost; too low a base-fee ceiling means the tx stalls in the mempool but nothing is lost. Any decision here should treat them separately.

Android already hit the display-inflation problem from the other direction and fixed it — see the comment on `ETHEREUM_SWAP_PRIORITY_FEE_FLOOR`, lowered from 2 gwei because it "dominated the user-displayed fee bond ... e.g. ~$2.8 of 'network fee' on a ~$7 swap".

## What is actually paid

```
displayed = gas limit * maxFeePerGas     <- two ceilings multiplied
actual    = gas used  * (baseFee + tip)  <- what leaves the wallet
```

None of the three figures is a charge. Under EIP-1559 the real cost is close on all three platforms; the only genuine difference is Android's 0.5 gwei tip floor against 1 gwei elsewhere, worth a few cents.

A larger reserve is still not free: it raises the balance a user must keep free and lowers the maximum amount they can swap when the source asset is the fee coin.

## Why this issue is EVM-only

Deliberate, not an oversight. The other fee resolvers in `@vultisig/core-mpc` return either the exact cost or a protocol constant — there is no ceiling to over-reserve:

```
utxo    -> BigInt(signingInput.plan.fee)          // the real signed plan fee
cosmos  -> protocol NativeTransactionFee (mimir)
solana  -> baseFee + ataRent + priorityFee * computeLimit
ripple  -> rippleSpecific.gas
ton     -> tonConfig.baseFee (+ jettonAmount)
tron    -> tronSpecific.gasEstimation
```

EIP-1559 is the only model here where you sign a ceiling and pay less, so EVM is the only chain kind where "displayed" and "paid" can diverge at all. Widening this issue to every chain would add surface without adding anything actionable.

What *is* chain-independent is the policy the team has to settle, even though its only practical surface today is EVM:

1. What does the displayed number mean — the reserve, or the expected cost?
2. Must headroom scale with the transaction, or may it be a flat floor?
3. Must the displayed number equal the signed one, or may they differ (as on Android)?

The per-chain *mechanics* should stay in their resolvers. Unifying those would fight the resolver pattern the repo mandates and would gain nothing.

Not covered: how iOS and Android compute non-EVM fees. Only the extension's resolvers were read for the table above, so this issue makes no cross-platform claim outside EVM.

## Solution

For the team to decide. The options, in increasing order of change:

1. **Use the floor only as a fallback.** Apply `deriveEvmGasLimit` when `eth_estimateGas` fails or returns nothing, instead of `bigIntMax`-ing it against a successful estimate. Smallest change; brings the extension in line with iOS.
2. **Align the base-fee multiplier across the three products.** Currently 1.5 / 1.32 / 1.8 for the same transaction class, with no shared rationale recorded.
3. **Split display from reserve, as Android does.** Sign the conservative number and show the expected one.

Options 1 and 2 land in `@vultisig/core-chain` / `@vultisig/core-mpc`, not in this repo.

### Must Do
- [ ] Team decides which of the three definitions is the reference
- [ ] Record the decision so the three clients stop drifting

### Must NOT Do
- Don't change the signed gas limit without checking the max-send and balance-check paths that read it
- Don't lower gas-limit headroom and base-fee headroom together — they guard different failures

## Acceptance Criteria
- [ ] `yarn check:all` succeeds
- [ ] For one swap quoted at the same moment, extension, iOS and Android show network fees within a small margin of each other
- [ ] The displayed fee scales with the transaction's own gas estimate rather than sitting on a flat floor — checked on an aggregator swap, a send with a memo, and a router deposit
- [ ] A swap whose real gas need exceeds the estimate still has enough headroom not to revert

Ehsan-saradar:
## Description

An EVM fee quote reserved `1.5 × max(estimate, routeGas, 600k)` gas at `1.5 × baseFee + tip`. The 600k value started as a fallback for a transaction that cannot be simulated, but it was applied with `max()` over a successful estimate, and the 1.5x buffer from #729 was then applied on top of that floor. Every calldata-bearing transaction was signed with at least 900k gas, so a simple swap, a memo send or a router deposit displayed a network fee several times its on-chain cost, and a max-amount swap of the fee coin left that much unusable.

Fixes vultisig/vultisig-windows#4799

The gas limit and the gas price now follow the transaction:

| Transaction | Gas limit | Base-fee headroom |
|---|---|---|
| Aggregator swap | `max(routeGas, 1.5 × estimate)`; 600k stands in for either figure when it is missing. Token routes are not simulated, since the allowance may not exist yet | 1.32x |
| THORChain / Maya deposit | fixed 120k | 1.2x |
| Plain transfer, with or without a memo | `max(estimate, per-chain floor)`, no inflation | 1.2x |
| dApp or other contract call | `1.5 × max(estimate ?? 600k, requested)` | 1.2x |
| Legacy-priced chain (BSC) | as above, priced from `eth_gasPrice` (+10% on swaps), no tip | n/a |

Tip: the highest 5th-percentile reward of the last ten blocks (`eth_feeHistory`), falling back to the node's own suggestion, and never above the gas price. Floors: 1 gwei on tip-auction chains, 30 gwei on Polygon, 20 wei on OP-stack rollups. Arbitrum, Mantle and Robinhood sign a zero tip, since their sequencers ignore it. The compromised-RPC ceilings are unchanged.

Per-chain transfer floors now reflect what the chains need (Base 50k / 150k, Polygon 40k, Blast 200k, Mantle 23k / 120k instead of 90M).

API: `getEvmFeeQuote`'s `minimumGasLimit` now only raises the fallback taken when simulation is unavailable, so existing callers keep compiling and behave the same on that path. New exports: `getEvmTransferGasLimit`, `getEvmContractCallGasLimit` and `evmRouterDepositGasLimit` from `tx/fee/evm/evmGasLimit`, and `getEvmGasPrice` from `tx/fee/evm/gasPrice`.

Kept on purpose: the OP-stack L1 surcharge stays in the displayed fee, because the node requires that balance, and zkSync keeps `zks_estimateFee` for its fee fields.

## Which feature is affected?

- [ ] Create vault ( Secure / Fast )
- [x] Sending - fee quote only; sends are floored per chain and no longer inflated
- [x] Swap - tx links below

## Checklist

## Screenshots (if applicable):

Same ETH → UNI quote (0.0002 ETH, 1inch) in the browser extension, before and after this change.

| Before | After |
|---|---|
| Network fee: **0.00099405 ETH (~$2.44)**<br><br>Before | Network fee: **0.00032581 ETH (~$0.80)**<br><br>After |

## Additional context

Swaps signed from the extension running this change:

- 1inch, ETH → UNI: https://etherscan.io/tx/0x37b0fa63c839a28f3ce7e5a66b5357f44272ad683d0912896a0e53f0aa1e1972
  Signed 293,088 gas (1.5x a 195,392 estimate), used 168,344. `maxFeePerGas` 1.0993 gwei against a 0.0718 gwei block base fee.
- SwapKit: https://track.swapkit.dev/?hash=0xb9494e841f3fdee65aa27abefda9775cc3acdf1652fe43315e8c09f56aab1fa6&chainId=1
  Signed 377,943 gas (1.5x a 251,962 estimate), used 216,925. `maxFeePerGas` 1.0667 gwei against a 0.0529 gwei block base fee.

Both used 57% of the signed limit, so the headroom held with margin.

Old and new quote for the same payloads against Ethereum mainnet in one run (base fee 0.049 gwei, tip on the 1 gwei floor):

| Scenario | Before | After |
|---|---|---|
| ETH send | 23,000 gas | 23,000 gas |
| ETH send with memo | 900,000 gas | 23,000 gas |
| Contract call (WETH `deposit`) | 900,000 gas | 42,426 gas |
| Aggregator swap, route gas 30k | 900,000 gas | 42,426 gas |
| THORChain deposit | 900,000 gas | 120,000 gas |

Verification: `yarn test:core` on the fee, keysign and EVM suites (1220 tests), `yarn workspace @vultisig/sdk typecheck`, `yarn knip`, `yarn changeset:check-sdk-bundle`, lint and format on the touched files. The browser extension was built against this change with `yarn build:extension`.

Consumers: vultisig-windows needs no code change and picks this up with the next SDK version bump.

## Summary by CodeRabbit

* **New Features**
  * Added current EVM gas-price retrieval.
  * Added chain-specific gas limits and priority-fee handling for transfers, token transfers, contract calls, swaps, and router deposits.
  * Added calldata-aware gas estimation and resilient fallbacks when fee history or transaction simulation is unavailable.

* **Bug Fixes**
  * Improved fee accuracy with network data and per-chain minimums.
  * Corrected zero-tip handling on supported networks.
  * Prevented successful gas estimates from being unnecessarily increased by caller minimums.
  * Improved zkSync fee handling and protection against excessive priority fees.
  * Improved fee quoting for memo-bearing transfers and swap transactions.


packages/core/chain/tx/fee/evm/maxPriorityFeePerGas.ts
```diff
@@ -1,4 +1,36 @@
...
+  const client = getEvmClient(chain)
+
+  const [feeHistory, gasPrice] = await Promise.all([
+    attempt(
+      client.getFeeHistory({
+        blockCount: feeHistoryBlockCount,
+        rewardPercentiles: [feeHistoryRewardPercentile],
+      })
+    ),
+    client.getGasPrice(),
```
johnnyluo: `client.getGasPrice()` runs unguarded inside this `Promise.all`, while the sibling `client.getFeeHistory()` call is wrapped in `attempt()`. Before this PR, `getEvmMaxPriorityFeePerGas` had a single (also unguarded) call; this adds a second, independent unguarded RPC call in parallel with it. A transient `eth_gasPrice` failure now rejects the whole `Promise.all` and throws, and nothing up the call chain (`getEvmFeeQuote` → `getEvmChainSpecific` → `getChainSpecific`) catches it — the keysign fails outright instead of degrading gracefully like the `getFeeHistory` path does. Suggest wrapping `client.getGasPrice()` in `attempt()` too and falling back sensibly (e.g. skip the gas-price cap, or reuse a value already fetched elsewhere) rather than letting the whole quote throw.

packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts
```diff
@@ -52,31 +83,54 @@ export const getEvmFeeQuote = async ({
...
+    kind === 'swap' ? percentOf(percentOf(baseFee, 110n), 120n) : percentOf(baseFee, 120n)
+
+  // A legacy-priced chain has a single gas price; a swap gets 10% on top of it.
+  const withGasPriceHeadroom = (gasPrice: bigint) => (kind === 'swap' ? percentOf(gasPrice, 110n) : gasPrice)
+
+  const getBaseFeePerGas = async () =>
+    isLegacyPriced ? withGasPriceHeadroom(await getEvmGasPrice(chain)) : withBaseFeeHeadroom(await getEvmBaseFee(chain))
+
+  const getMaxPriorityFeePerGas = async () =>
+    isLegacyPriced ? 0n : clampEvmPriorityFee(chain, await getEvmMaxPriorityFeePerGas(chain))
```
johnnyluo: For Arbitrum, Mantle, and Robinhood, `clampEvmPriorityFee` (in `clampEvmPriorityFee.ts`) unconditionally returns `0n` via its `zeroPriorityFeeChains` check, ignoring whatever value it's given. But `getMaxPriorityFeePerGas()` here still unconditionally calls `getEvmMaxPriorityFeePerGas(chain)` first, which does two RPC round-trips (`eth_feeHistory` + `eth_gasPrice`) whose result is then thrown away for these three chains. Worth short-circuiting before the call, e.g.:

```ts
const getMaxPriorityFeePerGas = async () =>
  isLegacyPriced || zeroPriorityFeeChains.includes(chain)
    ? 0n
    : clampEvmPriorityFee(chain, await getEvmMaxPriorityFeePerGas(chain))
```

This saves the wasted RPC calls and also removes the unguarded-`getGasPrice()` failure surface (see the other comment on `maxPriorityFeePerGas.ts`) for these three chains, since their priority fee was always going to be `0n` anyway.

packages/core/chain/tx/fee/evm/maxPriorityFeePerGas.ts
Ehsan-saradar: Fixed in db23195db. `eth_gasPrice` is now wrapped in `attempt()` like the fee-history call; when it fails the tip is returned uncapped, which is still a valid EIP-1559 field, so the quote no longer throws. Added a test for the failure path.

packages/core/mpc/keysign/fee/resolvers/evm/getEvmFeeQuote.ts
Ehsan-saradar: Fixed in db23195db. Added `isZeroPriorityFeeChain` next to the clamp, and the quote short-circuits to `0n` before any tip lookup on those chains. The Mantle test now asserts the tip helper is never called.

rcoderdev (commented):
When gas estimation fails for an Ethereum transfer with a plain-text memo, the new fallback uses 23,000 gas regardless of the memo size. A 1,000-byte nonzero memo needs at least 37,000 gas, and this quote flows into both the fee reserve and the signed transaction without raising the limit. A transient estimation failure can therefore make a supported memo transfer invalid and understate the max-send reserve. Please account for calldata requirements in the fallback, or fail explicitly when a safe limit cannot be established, and add a failed-estimation regression test with a long plain-text memo.

johnnyluo (approved):
LGTM

PR #2328, opened 2026-09-06
fix(swap): read native swap slippage from fees, where the wire sends it

Ehsan-saradar:
Prerequisite for vultisig/vultisig-windows#4831 — that fix cannot compile until this ships.

## Problem

`NativeSwapQuote.slippage_bps` sat at the top level of the quote, a level THORChain never sends it at.

The live `ETH.ETH -> BTC.BTC` response, the same request the app builds:

```
GET .../thorchain/quote/swap?from_asset=ETH.ETH&to_asset=BTC.BTC&amount=1580000&...

top-level slippage_bps:  << ABSENT >>

fees: {
  asset:        BTC.BTC
  affiliate:    0
  outbound:     1023
  liquidity:    98
  total:        1121
  slippage_bps: 19        <- it is here
  total_bps:    222
}
```

The spec is more emphatic than the response. `QuoteFees` declares the field **required**, and `QuoteSwapResponse` has no such property at any level:

```
QuoteFees.required:   [asset, liquidity, total, slippage_bps, total_bps]
QuoteSwapResponse:    no slippage_bps property
```

It appears at the top level only on `QuoteSaverDepositResponse` / `QuoteSaverWithdrawResponse` — unrelated endpoints, and the likely origin of the mistake. MayaChain's spec agrees: its `QuoteFees` carries `slippage_bps` too.

Because the field was optional, every consumer reading it silently got `undefined` rather than a compile error. That is how the Price Impact row went missing on every THORChain and MayaChain swap in the desktop app and the extension.

## Change

Move `slippage_bps` onto `NativeSwapFees`, and drop the phantom top-level field so the read that produced silence now fails to compile:

```
core/ui/.../priceImpact.ts(33,16): error TS2339:
Property 'slippage_bps' does not exist on type 'NativeSwapQuote'
```

Type-only. `getNativeSwapQuote` already spreads the response through verbatim, so the value was always present at runtime under `fees` — nothing about what is fetched, parsed, or signed changes.

## Note on the changeset level

Marked `minor`, not `patch`. Removing the top-level field is a breaking type change for any consumer that reads it. I grepped this repo and vultisig-windows and the only reader was the price-impact helper being fixed in the companion PR — but I could not check agent-backend or the mobile bridges, so flagging it rather than assuming.

Keeping the dead field with an `@deprecated` marker is the conservative alternative. I went the other way because an optional field the API never populates is exactly the trap that caused this bug, and it would stay reproducible.

## Verification

- Live THORChain quote and both openapi specs, quoted above
- `yarn typecheck` passes across the workspace with the top-level field removed
- Companion PR's suite: 10/10 on the price-impact tests, 2220 tests green repo-wide with this branch linked in

## Summary by CodeRabbit

* **Updates**
  * Native swap quote slippage information is now provided within the `fees` details rather than at the top level.
  * Native swap fees include optional slippage reporting in basis points, distinct from the total fee rate.
  * Slippage information is carried through native swap signing details so participating devices can display consistent price-impact information when available.
  * Existing swap runtime behavior is unchanged, and the field remains optional for compatibility.


Ehsan-saradar:
Downstream consumer: vultisig/vultisig-windows#4836 is held in draft until this merges and releases — it reads `fees.slippage_bps`, which does not exist on the published `NativeSwapFees` yet.

After this lands and the changeset release publishes `@vultisig/core-chain`, that PR needs its dependency bumped and can come out of draft.

johnnyluo (approved):
LGTM

rcoderdev (approved):

PR #2341, opened 2026-09-08
Reject negative amounts before hexadecimal encoding

### LINKED ISSUE #1139: encoding: bigIntToHex/numberToEvenHex turn a negative bigint into empty bytes instead of rejecting it

rcoderdev:
Closes #1139
Reject negative amounts before hexadecimal conversion can silently produce empty transaction bytes. The shared bigint and even-hex encoders and both Cardano token paths now raise RangeError while preserving non-negative encodings, including unsigned Long values.

Direct execution of the built package exports and Cardano resolver with WalletCore confirmed rejection of negative inputs and unchanged positive token bytes and transaction preimage.

## Summary by CodeRabbit

- **Bug Fixes**
  - Negative transaction and token amounts are now rejected before hexadecimal or byte encoding.
  - Invalid negative values consistently produce an error instead of being encoded as unsigned amounts.
  - Valid zero and positive amounts continue to use the expected encoding, including large supported values.

- **Tests**
  - Added coverage for negative, zero, positive, and maximum supported amount values across transaction and token encoding workflows.


johnnyluo (approved):
LGTM

PR #2344, opened 2026-09-08
fix(tron): reject balance gateway and contract-read errors

### LINKED ISSUE #2273: fix(tron): balance readers false-zero on gateway errors — feeds MAX send
`coin/balance/resolvers/tron.ts:35` — an `{Error}`-on-200 gateway reply falls through `?? '0'` to balance 0. `:117` — a reverted `eth_call` ("0x") reads as 0n. `:156` — a JSON-RPC `error.message` is fed to the numeric decoder. All three feed MAX send and fee simulation.

Fix: fail loud on the bare-error shape — `sdk/src/chains/tron/rpc.ts:212-219` already guards it and documents why; propagate that guard.
S2. 2026-08-30 Tron audit.

rcoderdev:
Closes #2273

## Summary by CodeRabbit

* **Bug Fixes**
  * Tron gateway and JSON-RPC errors now surface as failures instead of being interpreted as zero balances or decoded values.
  * Empty contract responses are rejected, while legitimate zero balances continue to be supported.
  * Maximum-send calculations now stop and report balance-read failures before estimating fees.
  * Improved handling of missing or malformed Tron balance responses.


johnnyluo (approved):
LGTM

PR #2345, opened 2026-09-08
fix(gas): preserve tiny EVM fee estimates and ranking

### LINKED ISSUE #1667: bug(compareCosts): non-zero sub-display L2 gas can round down to 0 gwei / zero native cost
## summary

`compareCosts()` can still round a real non-zero EVM gas price down to `0 gwei` / zero native tx cost.

On very cheap L2s this makes a real fee look free.

## evidence

- old rounding path in `packages/sdk/src/tools/gas/compareCosts.ts:69-73`
- sibling helper already had the correct fix in `packages/sdk/src/tools/evm/gasPrice.ts:43-52`
- regression gap was in `packages/sdk/tests/unit/tools/gas/compareCosts.test.ts`

## impact

- `compareCosts()` can emit `gasPriceGwei = 0` and `estTxCostNative = 0` for a real non-zero fee
- downstream consumers can present a misleading "free tx" result on low-fee L2s
- this is a duplicated rounding path that had already been fixed in the sibling gas helper

## fix sketch

- route `compareCosts()` through `evmGasPrice()` for the display-floor behavior
- add a regression test for a sub-display non-zero value like `49_999 wei`

## effort

- small / low-risk

## report

- improve pass report mirror: https://gist.github.com/gomesalexandre/f648d977ed73203baf5d397117809660

rcoderdev:
Closes #1667

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EVM gas-cost calculations for very small gas prices, preserving accurate non-zero estimates.
  * Native transaction fees are now calculated from precise raw values rather than rounded display values.
  * Gas-cost comparisons now rank options using accurate unrounded native and USD costs, including when displayed gas prices appear identical.


johnnyluo (approved):
LGTM

PR #2352, opened 2026-09-09
fix: reject missing and malformed signing amounts

### LINKED ISSUE #1147: mpc: empty-string amount silently zero-builds in resolvers untouched by #1140 (tron TRC20, solana send, polkadot, bittensor, ripple issued-currency)

rcoderdev:
Reject empty and malformed base-unit amounts before building Tron TRC20, Solana, Polkadot, Bittensor and Ripple issued-currency signing inputs. Preserve explicit zero TrustSet limits and larger supported amounts, and enforce unsigned Solana uint64 bounds.

Live SDK preparation and signing-hash extraction succeeded, then rejected nine malformed or overflowing amounts before signing. CLI/MCP previews succeeded, and valid WalletCore encodings matched previous behavior and a confirmed Tron transfer.

Refresh the locked XML and YAML parsers to patched versions after newly reported audit failures. The updated YAML parsers preserve ordinary data roundtrips and reject empty-merge budget exhaustion; XML plist roundtrips remain valid.
Closes #1147

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of transaction amounts across Tron, Solana, Polkadot, Bittensor, and Ripple.
  * Malformed, negative, fractional, signed, hexadecimal, scientific-notation, empty, and whitespace-only amounts are now rejected.
  * Solana amounts now enforce unsigned 64-bit limits, including overflow protection.
  * Explicit zero trust-line limits remain supported.
* **Tests**
  * Added coverage for valid boundaries, invalid formats, overflow handling, calldata generation, and metadata validation.


johnnyluo (approved):
LGTM

PR #2362, opened 2026-09-10
fix(tron): make address validation consistent

### LINKED ISSUE #1526: architecture: Tron address Base58Check decoding is duplicated and already drifted inside the sdk monorepo

rcoderdev:
Closes #1526

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TRON address validation across balance lookups, transaction building, and TRC20 energy estimation.
  * Added support for valid TRON address prefixes while rejecting malformed, incorrectly sized, unsupported, or invalid-checksum addresses.
  * Prevented energy-estimation requests when recipient address validation fails.

* **Tests**
  * Expanded coverage for address decoding, transaction encoding, RPC request construction, and invalid-address handling.


packages/core/chain/chains/tron/address.ts
```diff
@@ -0,0 +1,30 @@
...
+type Bs58CheckMod = { decode?: (s: string) => Uint8Array; default?: { decode: (s: string) => Uint8Array } }
+const _mod = bs58check as unknown as Bs58CheckMod
+const _decode: (s: string) => Uint8Array = (() => {
+  const fn = _mod.decode ?? _mod.default?.decode
+  if (!fn) throw new Error('bs58check.decode unavailable — bundler did not resolve bs58check correctly')
+  return fn
+})()
+
+// Preserve the address prefixes historically accepted by core balance lookup.
+const TRON_NETWORK_PREFIXES: readonly number[] = [0x41, 0xa0]
```
johnnyluo: **correctness**: Consolidating address decoding into shared `decodeTronAddress` silently widens the prefixes accepted by Tron tx-building and energy estimation from mainnet-only (`0x41`) to mainnet+Nile-testnet (`0x41`, `0xa0`), contradicting the changeset's claim of "preserving the existing accepted address prefixes."

Before this PR, `packages/sdk/src/chains/tron/tx.ts`'s `tronAddressToBytes` and `packages/sdk/src/chains/tron/rpc.ts`'s `estimateTrc20Energy` both threw for any prefix other than `0x41`. After this PR, both delegate to `decodeTronAddress`, which also accepts `0xa0` (previously only used by the balance resolver). A caller passing a Nile-testnet-prefixed address to `tronAddressToBytes`/`buildTrc20CallData` or `estimateTrc20Energy` — both public SDK exports re-exported via `packages/sdk/src/chains/tron/index.ts` — now gets a built/signed mainnet transaction or a mainnet TronGrid energy estimate instead of the immediate validation error it previously relied on.

johnnyluo (approved):
LGTM

PR #2365, opened 2026-09-10
fix(sdk): expose canonical token reference resolvers

### LINKED ISSUE #1665: public API: export the canonical token-ref resolver from the root sdk surface
## summary
The SDK now has a canonical token-ref resolver, but it is still private to vault internals instead of being exported for first-party consumers.

## evidence
- canonical resolver lives in `packages/sdk/src/vault/tokenRef.ts:24-89` (`ResolvedTokenInfo`, `resolveTokenRef`, `resolveTokenRefId`)
- `VaultBase` already routes balance/send/swap paths through it in `packages/sdk/src/vault/VaultBase.ts:80,967,1030,1995-1996`
- `packages/sdk/src/index.ts` does not currently export those helpers from the public root surface

## impact
The project finally has one definition of `what token does this ref mean?`, but app/backend consumers still cannot import it. That keeps consumer-local token selection / preflight code on the duplicated-not-imported path even after the canonical logic exists in the SDK.

This is exactly the architecture shape we want to avoid: the right code exists, but first-party consumers cannot consume it, so drift pressure stays high.

## fix sketch
Export `ResolvedTokenInfo`, `resolveTokenRef`, and `resolveTokenRefId` from the public SDK surface and add a public-exports regression.

That makes token-ref resolution importable by app/backend/CLI code without reaching into private paths.

## effort
small

## report
round 95 report mirror: https://gist.github.com/gomesalexandre/f06ddd911ca0e5f4e2df2b32392d4b72
campaign index artifact: https://claude.ai/code/artifact/be2deadf-6195-4dfa-a910-b269b9af7a84

rcoderdev:
Closes #1665


johnnyluo (approved):
LGTM

PR #2374, opened 2026-09-14
Fix TON status queries for base64 message hashes

rcoderdev:
TON message hashes containing plus signs were interpolated directly into the status URL, where they decoded as spaces and left indexed transactions appearing unknown. Encode the query parameters to preserve the original hash.

Live verification through the public SDK status API resolved the same indexed message from standard base64, URL-safe base64, and hexadecimal inputs. All three returned HTTP 200 and identical successful outcomes and fee receipts.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed TON transaction-status queries so base64 message hashes—including hashes containing plus signs—are preserved correctly.
  - Improved handling of standard base64, URL-safe base64, and hexadecimal transaction hashes.
  - Ensured transaction status requests use the exact hash provided by the transaction provider.


johnnyluo (approved):
LGTM

PR #2387, opened 2026-09-16
fix(sdk): expose grouped price helpers from every root entrypoint

### LINKED ISSUE #1780: public API: root @vultisig/sdk still omits a grouped sdk.price namespace handle

rcoderdev:
Closes #1780

## Summary by CodeRabbit

- **New Features**
  - Added a grouped `price` namespace to the SDK root for convenient access to price helpers.
  - Exposed the same `price` namespace through the React Native SDK entry point.
  - Existing flat price helper exports remain available.

- **Tests**
  - Added coverage confirming consistent price helper access and exports across SDK entry points.


johnnyluo (approved):
LGTM

PR #2390, opened 2026-09-16
fix(keysign): send approve(0) first for USDT-style tokens with a stale allowance

### LINKED ISSUE #695: ERC20 approval has allowance check + exact amount but no USDT zero-first reset
**Inherited by: Vultisig Windows + agent-app** (shared `@vultisig/core` swap-build path).

## Observed
ERC20 approval for swaps does an on-chain allowance check and approves the exact amount (good), but performs **no zero-first reset** for USDT-style tokens. USDT (and similar) revert on a non-zero→non-zero `approve` when a stale partial allowance already exists — so a user with a leftover partial allowance gets a failed approval and a blocked swap.

## Cause
`packages/core/mpc/keysign/swap/build.ts:346-360`
```ts
const allowance = await getErc20Allowance({ chain, id: fromCoin.id, address: fromCoin.address, spender })
if (allowance < chainAmount) {
  keysignPayload.erc20ApprovePayload = create(Erc20ApprovePayloadSchema, {
    amount: chainAmount.toString(),
    spender,
  })
}
```
When `0 < allowance < chainAmount` for a USDT-style token, this issues a single non-zero→non-zero `approve(chainAmount)` with no intervening `approve(0)`, which those tokens reject.

## Fix shape
For tokens requiring it (USDT et al., or unconditionally when `allowance > 0 && allowance < chainAmount`), emit an `approve(0)` reset before the `approve(chainAmount)` — either as a two-step approval or a maintained reset-required token list. Confirm the keysign flow can carry the extra approval step.

---

Ehsan-saradar:
## Description

Fixes #695

USDT (and tokens that copy its `approve`) revert when `approve(spender, amount)` is called with a non-zero amount while a non-zero allowance is still there. When a user had a leftover partial allowance, the swap approve reverted and the swap was blocked.

The fix needs a wire change, because every co-signer rebuilds the messages from the shared `KeysignPayload`. If only the initiator added an `approve(0)`, it would have 3 messages and iOS/Android 2, and the ceremony would stall. So:

- commondata: vultisig/commondata#112 adds `bool reset_allowance_first = 3` to `Erc20ApprovePayload` (issue vultisig/commondata#111). The generated `erc20_approve_payload_pb.ts` is updated here from that branch, nothing else regenerated.
- `getErc20ApprovePayload` (new, used by swap, limit swap and VULT staking builds): same allowance check as before. When `0 < allowance < amount`, it simulates the approve with `eth_call` from the owner address (`isErc20AllowanceResetRequired`). If that reverts, `resetAllowanceFirst` is set. No token list to maintain, and a plain `allowance == 0` still gives the same single approve as today.
- EVM signing inputs: when the flag is set, `approve(0)` at nonce N, `approve(amount)` at N+1, main tx at N+2. Otherwise unchanged.
- `buildCowSwapApprovalSigningInputs` (new) returns every approve leg (empty when none). `buildCowSwapApprovalSigningInput` stays exported as a deprecated single-leg form so Windows main keeps resolving the subpath on a minor bump: it returns the one approve when no reset is needed and throws, rather than dropping a leg, when the payload asks for the reset. Windows `cowSwapKeysign` should move to the plural form when it picks up this version.
- `BroadcastService` waits for each approve leg to confirm before sending the transaction after it, not only for index 0.

Until iOS and Android read the flag, a secure vault with a partial USDT allowance still cannot cosign this case with a mobile device (the ceremony fails instead of the approve reverting on chain). Fast vaults and SDK-only signing work now. The common case (allowance 0) is untouched on all platforms.

## Which feature is affected?

- [ ] Sending
- [x] Swap

## Checklist

## Additional context

Tests: `isErc20AllowanceResetRequired`, `getErc20ApprovePayload`, the resolver leg/nonce shape with and without the flag (`approveReset.test.ts`), both CowSwap approval builders, and the broadcast confirmation gating. `yarn check` passes except the pre-existing `example-electron` typecheck that needs a built SDK dist. `buildLimitSwap*` tests hit live RPCs and fail the same way on main.

## Summary by CodeRabbit

* **New Features**
  * Added support for USDT-style tokens that require an `approve(0)` reset before setting a new allowance.
  * Approval transactions preserve reset-then-approval order and wait for confirmation before dependent transactions are broadcast.
  * CowSwap supports multiple approval transactions when required.

* **Bug Fixes**
  * Improved handling of stale or partial ERC-20 allowances during swaps and staking.
  * Added detection for tokens that reject direct non-zero allowance updates.
  * Prevented reset approval steps from being silently omitted.


Ehsan-saradar:
Draft until vultisig/commondata#112 is merged. The generated proto here comes from that branch.

rcoderdev (commented):
This head removes the public export, but the changeset declares a minor release. Current Windows main still imports that exact path, so a normal 3.x upgrade will fail module resolution even for CowSwap approvals that do not need a reset. Please preserve a safe compatibility contract, or treat the removal as breaking and explicitly account for the consumer migration. A shim that silently drops one of the required approval legs would not be safe.

rcoderdev (approved):

PR #2391, opened 2026-09-16
fix(ton): do not report a failed jetton balance read as 0

### LINKED ISSUE #2187: [Fix] TON: failed jetton balance read shows as a real zero balance [TON]
**Problem:** If the jetton balance read fails (RPC error, indexer lag), we show 0 — indistinguishable from actually having zero. MAX send reads the same value.
**Why it matters:** "Your USDT is gone" support tickets, and MAX computed from a false zero. Same class as the EVM fix in vultisig-windows#4706 (preserve balances when reads fail).
**Fix:** Distinguish "read failed" from "balance is zero"; keep the last known balance and mark it stale instead of showing 0. Same pattern needed on iOS (`TonService.swift:142-149`) and Android (`TonApi.kt:116-120`) — siblings tracked in the epic.
**Receipts:** `chains/ton/api.ts:75` (`BigInt(wallet?.balance || '0')`).
Part of epic #2181.

Ehsan-saradar:
Closes #2187

`getJettonBalance` could return 0 for a read that actually failed:

- a 2xx body with no `jetton_wallets` list (proxy or indexer error body)
- a matching wallet with a missing or non numeric `balance`

Both now throw. The clients already keep the last known balance when the balance query throws (same handling that was added for EVM in vultisig-windows#4706), so nothing else is needed on the client side except bumping the SDK.

Real HTTP errors and timeouts already threw before this change through `queryUrl`.

One note on the issue: an empty `jetton_wallets` list is still treated as 0. On TON the jetton wallet contract only exists after the owner receives that jetton for the first time, so an owner with no wallet really holds 0. Indexer lag returns the same empty list and cannot be told apart from the response, so that case is out of scope here.


johnnyluo (approved):
LGTM

PR #2406, opened 2026-09-17
feat(ton): gasless jetton sends through the W5 relay

### LINKED ISSUE #2200: [Add] TON: gasless transfers via W5 relayer — pay fees in USDT (revenue line) [TON / Gasless]
**What:** Gasless TON transfers — users pay fees in the token they're sending (USDT etc.) instead of needing native TON. **This is a revenue line, not just UX:** the user signs a batch containing their transfer plus a fee payment to our relayer; the relayer (VultiServer is the natural host) broadcasts, pays the native gas, and keeps the in-token fee. Phantom charges ~1.5% for the equivalent on Solana; Tonkeeper monetizes via Battery.
**Why users expect it:** "Can't send my $2,000 USDT because I have no TON" is the top wallet-switching complaint in the TON ecosystem; Tonkeeper and Telegram Wallet both ship it.
**How it works with our vaults:** the relayer is broadcast-side only — it never joins the key ceremony. Fast and secure vaults sign the same W5 batch; the relayer can't alter it (signature covers the batch), worst case it declines to broadcast. Non-custodial throughout.
**Depends on:** W5 wallet contract support (#2181's W5 item) — gasless is a W5 feature.
Revenue/parity item under epic #2181.

Ehsan-saradar:
## Description

Fixes #2200

Gasless TON jetton sends for W5 accounts: the wallet pays the fee in the jetton itself through the gasless relay, so an account holding USDT but no TON can send it. Opt-in per send.

**API.** `prepareSendTx`, `send`, `getMaxSendAmount` and the vault-free prep helpers take `tonGasless` (`gasless` on `send`); the CLI `send` command gains `--gasless`. A dry run reports the fee in the jetton (`feeSymbol`) and folds it into `total`; `getKeysignFeeCoin` tells consumers which coin any payload's fee is denominated in.

**Payload.** The chain-specific step prices the transfer with the relay (`/gasless/config` + `/gasless/estimate`) and records the quote in the new `TonSpecific.gasless` (`TonGasless`, vultisig/commondata#114): relay address, gas jetton, commission and the quoted messages. The fee estimate quotes the smallest transfer, because the relay refuses any transfer that leaves no room for its commission regardless of its "throw if not enough jettons" flag; on a real build that refusal maps to `not-enough-funds`, and other relay refusals to `ton-gasless-unsupported`.

**Signing.** Every signer validates the quote against the approved transfer before hashing: exactly one TEP-74 transfer of the approved amount to the receiver with the memo intact and no custom payload, commission transfers to the relay summing to the declared fee, no state-init, total attached TON bounded. Only what the relay may legitimately touch (query ids, refund to relay or sender, attached TON within the bound) is left to it. The payload itself is the signing input (as for QBTC); the pre-image is the W5 `internal_signed` cell and compile assembles the envelope the relay accepts, with the W5 StateInit on a first request. The WalletCore TON resolver refuses a gasless payload rather than building an external request from it.

**Broadcast and status.** A relayed request is detected by its opcode and handed to `/gasless/send`; the reported hash is the trace id the relay broadcast under, which explorers resolve. Status resolves that trace (toncenter `traces`, by `trace_id` then `msg_hash`) to the wallet's own transaction, keeps the body-hash lookup as a fallback, flags a relayed request the wallet silently ignored as a signature rejection, and reports no TON fee receipt for relayed sends.

**W5 seqno fix (pre-existing bug).** Toncenter returns W5 accounts as raw code and data with no `seqno`, so the SDK signed seqno 0 for a deployed W5 wallet and the contract rejected the replay (exit 133) — every W5 send after the first failed. The seqno is now read from the W5 data cell after checking the code hash, in the chain-specific resolver, the balance tool and the RN wallet-info helper.

Follow-ups: iOS/Android consumption (they need the commondata regen and the same validation), and hosting our own relay — the client speaks the standard relay API, so it can be pointed at one.

## Which feature is affected?

- [x] Sending — live gasless send from a fast vault (W5, 0.2 USDT, 0.006997 USDT commission, no TON spent): https://tonviewer.com/transaction/05c7a804c0d1bea39bc5f8a7c92759fc02e3f9e44b0799f53da7766fcba2791b
- [ ] Swap
- [x] New Chain / Chain related feature — TON W5 relayed sends

## Checklist

## Additional context

Verified locally: `yarn test:core`, `yarn typecheck` (incl. CLI after `yarn build:sdk`), `yarn knip`, `format:check`, `changeset:check-sdk-bundle`, and the quality gates (contracts, architecture, duplication, complexity, secrets, docs). The SDK unit TON suites pass against a fresh bundle. Changeset bumps core-chain, core-mpc, sdk and cli (minor).

Tests: relay client, W5 `internal_signed` builder, TEP-74 body builder/parser, quote validation (tampered amount/receiver/memo/refund/forward/custom payload/state-init/commission/count/attached TON), signing hash and compile (signature verified with WalletCore, StateInit on seqno 0), pipeline routing, fee/fee-coin, chain-specific quote (W5 gate, jetton gate, relay mismatch, relay refusals), fee estimate quoting rule, broadcast routing, status trace/body-hash lookups and silent-ignore rule, W5 seqno reader, CLI dry run.

Consumers: vultisig-windows draft PR follows once this is released (pin bump); iOS and Android tickets to be raised from #2200.

## Summary by CodeRabbit

* **New Features**
  * Added gasless TON jetton transfers for W5 wallets through a relay, with fees paid in the sent jetton.
  * Added `--gasless` support to the CLI and gasless options across SDK sending and fee estimation.
  * Added relay quote validation, signed request handling, broadcasting, and transaction status tracking.
  * Dry runs now report jetton-denominated commissions and totals without requiring a TON balance.
* **Bug Fixes**
  * Improved sequence-number detection for deployed W5 wallets.
* **Documentation**
  * Added gasless TON transfer guidance and examples.


rcoderdev (commented):
Please resolve the user's W5 transaction even when the relay's external hash has a direct transaction match. `gaslessSend.external` identifies the relay message, so `packages/core/chain/tx/status/resolvers/ton.ts:127–129` can select the relay transaction and skip the wallet trace entirely.

For this PR's mainnet example, hash `01d7527fd4e494b28d57bb289722a3c6d3333693af434f10d88ab5a5b295b53f` returns the relay account with opcode `0x077ddc9e`, source `null`, successful action and fee `1010869`; the user's W5 transaction is a separate `0x73696e74` entry in the trace. The current resolver therefore reports relay success and its TON fee without checking the wallet's execution result.

Please preserve direct-wallet handling while resolving relayed sends to their wallet transaction, and add coverage with a successful direct relay match plus a pending/failed wallet request. The current trace tests mask this path by returning empty direct lookups.

Ehsan-saradar:
@rcoderdev Good catch, thanks. Fixed in 98106afb3.

When the hash matches a transaction whose inbound message is external and which emits an `internal_signed` (`0x73696e74`) message, the resolver now follows that message's hash to the wallet's own transaction and judges that one. It stays pending until the wallet transaction lands, and no relay fee is reported. A direct send's match has no such outbound message and is handled as before; the trace lookup is kept for identifiers that match no message.

Checked against your example: `01d7527f…` returns the relay transaction with a single `0x73696e74` out message, and following it lands on the W5 account's transaction (exit 0, 2 actions). Tests cover the relay match → wallet transaction, a pending wallet transaction, a wallet exit `133` → `seqno-mismatch`, `tot_actions: 0` → `invalid-signature`, and a direct send whose jetton out messages must not be followed.

rcoderdev (approved):

PR #2407, opened 2026-09-18
Await MPC relay sends and retry transient failures

### LINKED ISSUE #2311: Keygen: relay send not awaited in processOutbound, await with backoff and throw
### Why

3 signers in far different geo locations creating a 2-of-3 Secure Vault hit `timeout: failed to create vault within 60 seconds` 9 times in a row.

Probable cause, from reading the code, not verified end to end: one device's relay POST failed, the client had the error, the code ignored it and waited 60s for a reply that could never arrive. After the timeout it restarted the protocol locally while the other two devices kept waiting on the old session.

Windows and the extension run this code through the SDK.

### Today

- `packages/core/mpc/dkls/dkls.ts` `processOutbound` (lines 92 to 108): `receivers.forEach(... sendMpcRelayMessage(...))` with no `await`. The catch at lines 112 to 119 logs and recurses, so a send failure never surfaces.
- Same pattern: `schnorr/schnorrKeygen.ts:88`, `mldsa/mldsaKeygen.ts:78`, `keysign/index.ts:106`, `mldsa/mldsaKeysign.ts:76`.
- `message/relay/send.ts`: single `queryUrl`, no timeout, no retry.
- Inbound stall in `dkls.ts:127` is already progress-based. Confirm Schnorr and MLDSA use the same clock.

### Change

1. `sendMpcRelayMessage` retries only on timeout, network error or 5xx with 1s / 2s / 4s / 8s backoff, request timeout 8s, then throws. 4xx throws at once.
2. `processOutbound` awaits all sends and rethrows send failures instead of recursing, so `startKeygenWithRetry` handles them. Same in the four sibling files.
3. Changeset: patch on `@vultisig/core-mpc`.

Invariant: send budget (4 × 8s timeout + 1/2/4/8s backoff, under 50s) stays below the 60s stall. Do not move the send timeout back to 60.

### Acceptance

Unit test: mock relay rejects the first outbound POST 3 times, then accepts. Keygen completes, no timeout. A 4xx fails the attempt immediately.

### Related

Same change on iOS and Android: https://github.com/vultisig/vultisig-ios/issues/5317, https://github.com/vultisig/vultisig-android/issues/5813.

rcoderdev:
Closes #2311

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MPC relay reliability by retrying temporary network and server failures with bounded delays.
  - Added request timeouts and cancellation handling to prevent operations from waiting indefinitely.
  - Coordinated message sending and receiving so related operations stop promptly when a failure occurs.
  - Improved delivery of messages to multiple recipients with consistent sequencing and completion tracking.

- **Tests**
  - Added coverage for retries, timeouts, cancellation, multi-recipient delivery, and failure handling.


johnnyluo (approved):
Reviewed: no actionable findings.

PR #2410, opened 2026-09-18
fix(security): block Tron zero-address destinations

### LINKED ISSUE #2272: fix(security): dangerous-address registry has no Tron branch — zero address unguarded
`dangerousAddresses.ts:106-125` branches on solana/UTXO/ripple then returns undefined — Tron sends and swap recipients get no guard. The Tron black-hole `T9yD14Nj9j7xAB4dbGeiX9h8unkKHxuWwb` is literally address(0) (we even hardcode it as a dummy caller in `token/metadata/resolvers/tron.ts:12`).

Fix: add a Tron case for the zero address + sending a token to its own contract (send-to-contract guard for TRC-20 prep is #1724).
S2. 2026-08-30 Tron audit.

rcoderdev:
Closes #2272


johnnyluo (approved):
Reviewed: no actionable findings.

PR #2416, opened 2026-09-19
fix(solana): splice the dApp raw-tx signature into the vault's own signer slot

### LINKED ISSUE #2415: fix(solana): dApp raw-tx splice writes the vault's signature into signer slot 0 instead of the vault's own slot
## Symptom

A dApp sends `solana:signTransaction` (wallet-standard, via the extension) for a partially-signed v0 transaction with three required signers where the vault is **not** the fee payer:

```
staticAccountKeys[0] = A   fee payer / co-signer, signs later — slot left zero
staticAccountKeys[1] = B   the vault (the account passed to signTransaction)
staticAccountKeys[2] = C   co-signer, already signed
```

The wallet returns a transaction whose message bytes are unchanged and whose slot 2 still carries C's valid signature, but the vault's signature is written into **slot 0** (A's slot) and slot 1 (B's own slot) is still all zeros. The signature itself is a correct ed25519 signature by B over the message — it just sits in the wrong slot. Solana verifies `signatures[i]` against `accountKeys[i]`, so the tx can never pass sigverify, and slot 0 is clobbered before A gets to sign it.

Reported by an integrating dApp team (sponsored/multi-signer flow).

## Root cause

`spliceSolanaSignature` (`packages/core/mpc/keysign/signingInputs/resolvers/solana/rawTx.ts`) splices at `firstSignatureOffset` unconditionally — i.e. always signer index 0. Its docstring already names this case as unhandled:

> A sponsored transaction (a relayer pays and is signer 0, the vault signs at index >= 1) would land the vault's signature in the wrong slot

`compileTx` (`packages/core/mpc/tx/compile/compileTx.ts`, `signSolana` branch) calls it for every dApp raw tx and already has the vault `publicKey` in scope, so the correct slot is computable there.

The pre-image is unaffected: `getPreSigningHashes` signs the wire message verbatim, which is why the signature verifies under B. Only the final assembly is wrong.

iOS `Solana.swift` `signRawTransaction` and Android `SolanaSignatureEnvelope` make the same slot-0 assumption (parity ports of ios#4419 / android#5223); this issue covers the SDK side.

## Fix

Parse the message header from the extracted message bytes (skip the `0x80` version prefix for v0), read `numRequiredSignatures` and the static account keys compact-array, find the vault public key among the first `numRequiredSignatures` static keys, and splice at `firstSignatureOffset + index * 64`. Throw if the vault is not a required signer instead of writing into someone else's slot. Leave every other slot as received.

`getSolanaTxHash` keeps reading the vault's signature from `SigningOutput.signatures`; when the vault is not fee payer that is not the txid, but the tx is returned to the dApp unbroadcast (`skipBroadcast`), so it does not matter for this path.

## Scope

- [ ] `spliceSolanaSignature` takes the signer public key and resolves the slot from the message
- [ ] Throws when the public key is not among the required signers
- [ ] Legacy and v0 messages both handled
- [ ] Unit tests: fee payer (index 0, unchanged behaviour), signer at index 1 with pre-existing signature at index 2 preserved, non-signer key rejected
- [ ] No change to `extractSolanaMessageBytes` / the pre-image

Ehsan-saradar:
## Description

Fixes #2415

`spliceSolanaSignature` always wrote the vault's signature into signer slot 0, which is only correct when the vault is the fee payer. A sponsored or multi-signer transaction puts the vault at index ≥ 1; the runtime verifies `signatures[i]` against `accountKeys[i]`, so the signature landed in the relayer's slot, the vault's own slot stayed all-zero, and the returned transaction could never pass sigverify. Reported by an integrating dApp team sending a partially-signed v0 tx with three required signers (relayer at 0, vault at 1, co-signer already signed at 2).

The slot is now resolved from the message instead of assumed:

- `rawTx.ts` — new `getSolanaSignerIndex({ message, publicKey })` parses the message header (legacy, or v0 behind the `0x80` version byte), reads `numRequiredSignatures` and the static-keys compact array, and returns the key's index among the required signers. It throws if the key is not a required signer, so we never write into somebody else's slot, and on truncated / over-declared / unsupported-version messages. `spliceSolanaSignature` now takes `{ txData, signature, publicKey }`, splices at `firstSignatureOffset + index * 64`, and rejects an envelope with fewer slots than the index. Every other slot — including co-signers' existing signatures — is left exactly as received.
- `compileTx.ts`, `getCompiledTxsForBlockaidInput.ts` — pass the vault public key through (both already had it in scope).

`extractSolanaMessageBytes` and the pre-image are byte-for-byte unchanged, so Secure Vault co-signing hashes do not move; this is assembly-only. The shortvec decoder was factored out and shared between the envelope and the message header.

A transaction where the vault is not a required signer now fails at compile instead of producing a silently broken tx (the Blockaid preview degrades to "no scan" rather than blocking). Rejecting that earlier, before the ceremony, would be a client-side change.

## Which feature is affected?

- [ ] Swap - Please attach a tx link for the swap here
- [x] New Chain / Chain related feature - Solana dApp raw-tx signing (`signData.signSolana`) via the extension's `solana:signTransaction`

Fee-payer transactions (every existing dApp flow) resolve to slot 0 as before; covered by the unchanged legacy and v0+ALT round-trip tests.

## Checklist

## Screenshots (if applicable):

n/a

## Additional context

Tests rebuilt on real web3.js transactions, with the reporter's exact shape as a fixture (v0, three signers, relayer at 0, vault at 1, co-signer pre-signed at 2):

- signer index resolution for legacy, v0+ALT and sponsored messages; non-signer and unknown keys, truncated key lists, over-declared signer counts and unsupported versions rejected
- splice: fee payer into slot 0; vault into slot 1 with slot 0 still zero and the co-signer's signature intact; envelope with too few slots rejected
- end-to-end `getEncodedSigningInputs → getPreSigningHashes → sign → compileTx`: slot 1 verifies under the vault key over the original message, slot 0 stays open and the relayer can complete the tx afterwards

Local gate: `rawTx.test.ts` 28/28 and `getCompiledTxsForBlockaidInput.test.ts` pass · root and `shared-publish` `tsc` · eslint · prettier · knip all clean. Changeset: `@vultisig/core-mpc` patch.

Ships to the extension after the release + pin bump in vultisig-windows. iOS `Solana.swift` `signRawTransaction` and Android `SolanaSignatureEnvelope` carry the same slot-0 assumption and need the mirror change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Solana transaction signatures are now placed in the vault’s correct signer slot instead of always being inserted into the first slot.
  * Improved support for sponsored and multi-signer transactions, including transactions where the vault is not the fee payer.
  * Existing signatures and transaction contents are preserved during signing and preview generation.
  * Added validation for unsupported, malformed, non-canonical, or incorrectly structured Solana transaction inputs.
  * Improved handling of legacy and versioned transactions, including those using address lookup tables.


johnnyluo (approved):
Reviewed and approved. Focused Solana/Blockaid tests pass; the signer-slot and malformed-envelope handling look correct.

PR #2418, opened 2026-09-19
fix(bech32): pass an explicit limit to every fromBech32 decode

### LINKED ISSUE #2417: [Fix] fromBech32 with default limit throws under @scure/base ≥2.3, so QBTC address validation rejects every address [core-chain]
# [Fix] `fromBech32` with default limit throws under `@scure/base` ≥2.3, so QBTC address validation rejects every address [core-chain]

Root cause of vultisig/vultisig-windows#4972. Affects `@vultisig/core-chain@5.4.3` (and every version since QBTC was added in #157).

## Problem

Six call sites decode bech32 with `@cosmjs/encoding`'s `fromBech32(address)` and no explicit `limit`. `@cosmjs/encoding@0.39.0` defaults `limit = Infinity` and forwards it to `@scure/base`'s `bech32.decode(str, limit)`. Starting with `@scure/base@2.3.0`, `decode` calls `anumber(limit, 'limit')`, which rejects anything that isn't a safe integer:

```
RangeError: limit: expected safe integer, got Infinity
```

`@cosmjs/encoding` declares `@scure/base: ^2.0.0`, so which version a consumer actually gets is decided by *their* lockfile. This repo's `yarn.lock` resolves that range to `2.0.0`, so our tests pass. vultisig-windows dedupes it onto `2.3.0` (required by `@mysten/sui` and `@scure/bip32`), and there every no-limit `fromBech32` call throws.

The user-visible result: in `isValidAddress`, the QBTC branch wraps the call in `try/catch` and returns `false`, so **every** QBTC address — including the vault's own — is rejected with "Wrong address for selected chain". Reproduced against the windows install:

```
qbtc12tkhjqjqz3uz0r3tm7cmlvdulk4psputeucxhj => limit: expected safe integer, got Infinity
qbtc10hmrwslfvxtaaag8rrk3vqr5f5uj7z80z9p9nd => limit: expected safe integer, got Infinity
```

Callers with the same latent failure (all swallow the throw or convert it to a generic "invalid address" error):

| File | Effect under `@scure/base` ≥2.3 |
|---|---|
| `packages/core/chain/utils/isValidAddress.ts:26` | every QBTC address rejected |
| `packages/core/chain/swap/native/limitSwapMemo.ts:253` | `isBech32Address` always `false` |
| `packages/core/chain/swap/general/ruji/api/getRujiTradeSwapQuote.ts:74` | every THOR address rejected as "must be a valid THORChain address" |
| `packages/core/mpc/keysign/swap/build.ts:74` | bech32 decode in swap build fails |
| `packages/sdk/src/tools/cosmos/gov.ts:466` | voter address decode fails |
| `packages/rujira/src/validation/address-validator.ts:77` | bech32 validation always fails |

`packages/core/chain/chains/cardano/cip30/cardanoAddressBytes.ts:14` is the one caller that already passes an explicit limit and is unaffected.

## Expected Behavior

A well-formed bech32 address decodes regardless of which `@scure/base` 2.x the consuming app's lockfile picks. `isValidAddress` accepts `qbtc1…` addresses and still rejects other prefixes.

## Steps to Reproduce

1. In any consumer, resolve `@cosmjs/encoding`'s `@scure/base@^2.0.0` to `2.3.0` (e.g. `"resolutions": { "@scure/base": "2.3.0" }` in this repo, then `yarn install`).
2. Run `isValidAddress({ chain: Chain.QBTC, address: 'qbtc12tkhjqjqz3uz0r3tm7cmlvdulk4psputeucxhj', walletCore })`.
3. Returns `false`.

## Solution

Stop relying on `fromBech32`'s `Infinity` default. Introduce one shared helper (e.g. `packages/core/chain/utils/decodeBech32.ts`) that decodes with an explicit finite limit — bech32 addresses we handle are ≤90 chars, matching `@scure/base`'s own default — and route all six callers through it. Either `fromBech32(address, 90)` or `@scure/base`'s `bech32.decode(address)` directly works; the helper keeps the choice in one place. Add a QBTC case to `isValidAddress.test.ts` and run the suite once with `@scure/base` pinned to 2.3.0 to prove the fix holds under the dedupe consumers actually get.

### Must Do
- [ ] Shared bech32 decode helper with an explicit limit; all six callers above use it
- [ ] `isValidAddress.test.ts`: `qbtc1…` accepted, `cosmos1…` / `thor1…` rejected for `Chain.QBTC`
- [ ] Verify the suite passes with `@scure/base` resolved to `2.3.0` (temporary resolution, not committed)
- [ ] Changesets for `@vultisig/core-chain`, `@vultisig/core-mpc`, `@vultisig/sdk`, `@vultisig/rujira`

### Must NOT Do
- Don't change prefix/length semantics at any caller (QBTC = `qbtc`, RUJI = `thor` with 20/32-byte data, etc.)
- Don't pin `@scure/base` in this repo's lockfile as the fix — that only hides it here, consumers still break
- Don't touch `cardanoAddressBytes.ts` (already correct)

## Acceptance Criteria
- [ ] `yarn check:all` succeeds
- [ ] `isValidAddress` returns `true` for `qbtc12tkhjqjqz3uz0r3tm7cmlvdulk4psputeucxhj` with `@scure/base@2.3.0` resolved
- [ ] `getRujiTradeSwapQuote` accepts a valid `thor1…` address under the same resolution
- [ ] After bumping `@vultisig/core-chain` in vultisig-windows, QBTC Send accepts a valid address (closes vultisig/vultisig-windows#4972)

Ehsan-saradar:
## Description

`fromBech32` from `@cosmjs/encoding` defaults its length limit to `Infinity` and forwards it to `@scure/base`'s `bech32.decode`. Since `@scure/base@2.3.0` that limit is validated as a safe integer, so in any consumer whose lockfile resolves `@cosmjs/encoding`'s `^2.0.0` range to 2.3.0 — vultisig-windows does, via `@mysten/sui` and `@scure/bip32` — every bare `fromBech32(address)` call throws:

```
RangeError: limit: expected safe integer, got Infinity
```

Our own lockfile resolves that range to 2.0.0, which is why the SDK test suite never caught it.

The user-visible symptom is QBTC send: `isValidAddress` swallows the throw and returns `false`, so every `qbtc1…` address — including the vault's own — is rejected with "Wrong address for selected chain". THOR address checks in RUJI trade quotes, limit-swap memos, swap keysign builds, Cosmos governance voting and Rujira destination validation fail the same way.

**Change**

- Add `decodeBech32` in `packages/core/chain/utils` that always passes the BIP-173 90-char cap, with a comment on why the limit must be explicit.
- Route the five callers that already depend on core-chain through it: `isValidAddress.ts`, `limitSwapMemo.ts`, `getRujiTradeSwapQuote.ts`, `core/mpc/keysign/swap/build.ts`, `sdk/src/tools/cosmos/gov.ts`.
- `rujira/src/validation/address-validator.ts` passes the limit directly (rujira is standalone and doesn't depend on core-chain).
- `cardanoAddressBytes.ts` already passed an explicit limit and is untouched.
- Changeset: patch for `@vultisig/core-chain`, `@vultisig/core-mpc`, `@vultisig/sdk`, `@vultisig/rujira`.

No validation semantics change: prefixes, payload-length checks and error messages at every caller are as before. The only behavioural difference is that strings over 90 characters now throw at decode instead of being length-unbounded — no address handled by these callers comes close (Cardano has its own decoder).

Fixes #2417
Root cause of vultisig/vultisig-windows#4972

## Which feature is affected?

- [x] Sending — QBTC recipient validation (form-level; no tx broadcast in this change)
- [x] Swap — THOR address validation in RUJI trade quotes, limit-swap memos and swap keysign builds

## Checklist

## Additional context

**Tests**

- `isValidAddress.test.ts`: 7 QBTC cases against two real vault addresses, plus valid-bech32-wrong-prefix (`cosmos1…` / `thor1…` re-encodings of the same 20-byte payload), bad checksum, surrounding whitespace, garbage.
- `decodeBech32.test.ts`: 20- and 32-byte payloads, uppercase input, bad checksum, no separator, over-cap length.

**Verified the regression test actually catches the bug**: with `@scure/base@^2.0.0` temporarily resolved to 2.3.0 (matching vultisig-windows) and the fix stashed, the three "accepts" QBTC tests fail; with the fix restored, all 586 tests in the affected core areas pass under that same resolution. The temporary resolution is not part of this PR — `package.json` / `yarn.lock` are unchanged.

**Full runs on the default lockfile**: `yarn check` exit 0 · `test:core` 4143 passed · `sdk` unit 4240 passed · `test:rujira` 558 passed.

**Follow-up**: vultisig-windows needs the `@vultisig/core-chain` bump once this is released (vultisig/vultisig-windows#4972).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bech32 address decoding and validation across swaps, voting, wallet, and transaction workflows.
  * Prevented failures caused by overly long or malformed addresses.
  * Preserved existing prefix, checksum, and payload validation behavior.
  * Improved validation support for QBTC addresses, including whitespace handling.

* **Tests**
  * Added coverage for valid, malformed, uppercase, oversized, and QBTC address inputs.


johnnyluo (approved):
Reviewed and approved. Focused core-chain and Rujira address-validation tests pass; the explicit Bech32 decoding limit is correctly scoped.

PR #2421, opened 2026-09-21
fix(solana): preserve full approval movements

### LINKED ISSUE #2091: [BUG] Blockaid simulation: the three-diff native-asset filter can delete a principal leg and reverse the approval direction

rcoderdev:
Closes #2091

Solana simulations with three balance changes discarded native SOL before mint netting. This could reverse deposit/withdrawal summaries or remove the principal movement. Preserve every movement, net SOL and wrapped SOL by mint, and only classify an unambiguous send or swap.

This intentionally declines the simplified headline for a token-to-token swap with a separate native SOL fee (`SOL -5000`, `USDC -X`, `BONK +Y`). The response contains two spent mints, and the available movement metadata does not independently identify a fee. Removing the smallest movement would also remove fee-sized principal sends. The Windows consumer therefore retains instruction/raw details for this shape; ordinary SOL/token swaps and same-mint SOL/WSOL fee netting remain supported.

iOS adopted complete Solana mint netting in vultisig/vultisig-ios#5321 and also declines this multi-spend shape. Android still excludes the smallest out-only native SOL movement; its policy alignment is tracked in https://github.com/vultisig/vultisig-android/issues/5928. This is source comparison, not mobile runtime QA. Sui behavior is unchanged; its remaining portion of #2091 is tracked in https://github.com/vultisig/vultisig-sdk/issues/2423.

Verified the packed SDK with unsigned Kamino SOL and USDC deposit approvals in the Windows extension: the SOL summary retains the spend/receive direction, while the ambiguous USDC result uses raw instruction details with the accompanying Windows parser change. Requests were rejected without signing or broadcasting. A live withdrawal was not exercised; regression fixtures cover the reported withdrawal and deposit cases, ordering, small principal amounts and fee netting.

The Windows consumer change will follow after these SDK packages are published. Related Windows issue: https://github.com/vultisig/vultisig-windows/issues/4728.

## Summary by CodeRabbit

* **Bug Fixes**
  * Solana transaction summaries now preserve native SOL movements when calculating net amounts.
  * Corrected reversed deposit and withdrawal summaries involving SOL and wrapped SOL.
  * Improved handling of fee-sized movements and inconsistent asset metadata.
  * Unsupported multi-asset transaction shapes are now rejected instead of being misclassified as fees.
  * Improved accuracy for transactions containing native SOL and wrapped SOL movements in varying processing orders.


Ehsan-saradar (changes requested):
The netting change is correct and the permutation tests lock the reversed-direction case properly. Two things before merge:

1. Dropping the filter means every token→token swap with a fee diff (`SOL −5000`, `USDC −X`, `BONK +Y`) now throws and loses its preview. That's the most common Solana swap shape. If that's the intended trade, fine, but it should be stated explicitly in the PR and matched on iOS/Android. The alternative from the issue — only drop a native leg that is out-only, under a fee ceiling, and leaves a clean two-mint swap — keeps the preview and is still immune to the counterexample (the principal there is an inflow).

2. Sui still has the same 3-diff heuristic, so this doesn't fully close #2091. Change to `Part of` or open the Sui follow-up.

Nit: the `solanaMintForAsset` comment still references "the native-SOL-fee filter a few lines below".

rcoderdev:
Thanks, Ehsan. The fee-bearing token-to-token fallback is intentional, and I have made it explicit in the PR description. `SOL -5000`, `USDC -X`, `BONK +Y` preserves all movements and declines the simplified headline. The available movement metadata does not establish which outflow is a fee, and a size ceiling would also discard small principal sends. Windows retains instruction/raw details in that case. Ordinary SOL/token swaps and same-mint SOL/WSOL netting remain supported.

iOS already uses complete mint netting in vultisig/vultisig-ios#5321 and declines this multi-spend shape. Android still has the smallest-native-outflow heuristic, so I opened vultisig/vultisig-android#5928 for alignment. This comparison is based on source; I have not claimed mobile runtime verification.

I also opened #2423 for the remaining Sui portion of #2091, so that work remains tracked separately when this Solana fix lands. The stale `solanaMintForAsset` comment is corrected.

I rechecked the packed SDK in Windows with real unsigned SOL and USDC deposit approvals: the SOL amount keeps the full net spend, and the multi-spend USDC result retains transaction details. Both requests were rejected without signing or broadcasting. The required SDK base update also needed a small Windows adaptation to display imported TON gasless commission in the jetton's units; that is verified and retained for the Windows package-integration PR. I also checked your Windows PR vultisig/vultisig-windows#4962, which includes the same fee-coin adjustment; I will reconcile and credit that overlap when integrating the published packages, while leaving its broader gasless-send UI with that PR.

Ehsan-saradar (approved):

PR #2429, opened 2026-09-22
fix(sdk): reject invalid UTXO balance responses

### LINKED ISSUE #2372: fix(balance): reject malformed UTXO balance responses instead of returning zero or truncating values
## Problem
`getUtxoBalance` can return a successful zero or incorrect balance when the HTTP response is successful but its JSON is not a valid address balance.

Reproduced against main `42ef35c4542e1289b6a4e70437d31fe917f9bd22` using the production TypeScript module and a local HTTP server configured through the supported `blockchairBase` option. This is a controlled response-handling reproduction, not evidence of a current Blockchair outage.

## Reproduction
Call `getUtxoBalance(UtxoChain.Bitcoin, 'test', { blockchairBase: localServerUrl })` with the server returning HTTP 200 and these bodies:

| Response body | Actual result | Expected |
| --- | --- | --- |
| `{"data":{}}` | `satoshis: "0"` | Reject missing address data |
| `{"context":{"code":429,"error":"Rate limit exceeded"},"data":null}` | `satoshis: "0"` | Reject the provider error |
| `{"data":{"test":{"address":{"balance":1e8}}}}` | `satoshis: "1"` | Preserve 100000000 or reject unsupported numeric encoding; never silently truncate |

Controls: `balance: 123456789` returned `123456789`; `balance: 9007199254740993` preserved the exact integer; an explicit `balance: null` returned the currently documented zero balance.

## Cause and impact
The helper checks HTTP status, then treats missing `data`/`address`/`balance` as an unfunded address. The raw integer regex also accepts only the leading digits of a JSON numeric token, so it silently truncates exponent notation. Callers cannot distinguish unavailable balance data from an empty wallet, and numeric parsing can produce a wrong amount.

[Current implementation](https://github.com/vultisig/vultisig-sdk/blob/42ef35c4542e1289b6a4e70437d31fe917f9bd22/packages/sdk/src/tools/balance/utxoBalance.ts)

## Acceptance
- Validate the provider envelope and requested address record before returning a balance.
- Keep a documented, explicit unfunded-address response distinct from malformed data and provider errors.
- Parse a complete nonnegative integer value without losing large-integer precision; unsupported representations must reject rather than partially parse.
- Cover the responses above, valid zero/null semantics, large integers, wrong-address data and negative/fractional values.
- Preserve the supported `blockchairBase` and timeout behavior.

Related: #1344 changes timeout handling in this module; its current patch does not change balance-envelope validation or numeric parsing. Coordinate overlapping edits without coupling this defect to the React Native timeout change.

rcoderdev:
Closes #2372


johnnyluo (approved):
Reviewed: approved.

PR #2432, opened 2026-09-23
fix(sdk): validate encoded EVM swap amounts and deadlines

### LINKED ISSUE #1196: fund-safety: evm-general vault-free swap has NEITHER amount-consistency NOR expiry enforcement (the named residual from #1117's fix)
The plans/README.md plan-005 outcome note names this residual explicitly and it was never filed: after the vault-free swap helper gained amount+expiry checks for native quotes and transfer/cowswap_order shapes, an EVM-GENERAL vault-free swap (arguably the most common shape) still has NO amount-consistency check and NO expiry enforcement - the amounts are encoded inside opaque aggregator calldata, so there was nothing to compare against. Not a regression (nothing enforced it before either), but it's the real remaining exposure.

Prerequisite: an EVM calldata amount-decoder (decode the sell amount from the aggregator calldata for the known providers' shapes) - which would also unlock decode-based checks elsewhere (the WYSIWYS ERC-20 class). Worth designing once, using it twice. Effort: M/L.

rcoderdev:
Closes #1196

## Summary by CodeRabbit

* **Bug Fixes**
  * EVM swap preparation now checks supported transaction data against the requested source asset and input amount, and rejects swaps with expired deadlines or inconsistent transaction details.
  * Checks cover supported layouts for several common EVM swap routes, including native THOR deposits. Unsupported formats may not provide an input amount to verify.

* **Documentation**
  * Added guidance on supported transaction formats, validation rules, and remaining coverage limits.


packages/sdk/src/tools/prep/evmSwapCommitment.ts
```diff
@@ -0,0 +1,250 @@
...
+  const abi = tx.to.toLowerCase() === '0x1111111254eeb25477b68fb85ed929f73a960582' ? oneInchV5Abi : oneInchV6Abi
+  const decoded = decodeKnown(abi, tx.data)
+  if (!decoded) return {}
+  if (decoded.functionName === 'swap') {
+    const desc = decoded.args[1]
+    // Extra-native-value routes need a separate fee authorization contract. Do not
+    // equate their transaction value with the source amount or invent a tolerance.
+    if (desc.flags & 2n) return {}
+    const native = [nativeToken, zeroAddress].includes(desc.srcToken.toLowerCase())
+    if (parseValue(tx.value) !== (native ? desc.amount : 0n)) return malformed()
```
Ehsan-saradar: When tx.value doesn't match desc.amount on a native swap, this throws "malformed recognized EVM swap calldata". The calldata isn't malformed, the amounts just disagree. A mismatch-specific message would make this much easier to debug. Same thing on line 166 for Kyber.

packages/sdk/src/tools/prep/evmSwapCommitment.ts
```diff
@@ -0,0 +1,250 @@
...
+ * Decode independently verifiable source commitments at known deployed routers.
+ * This is intentionally partial, not an authenticity check or a generic calldata
+ * interpreter. Unsupported layouts return no amount, retaining any known deadline.
+ */
+export const decodeEvmSwapCommitment = (input: Input): EvmSwapCommitment => {
+  if (isKnownRouter('1inch', input)) return decodeOneInch(input)
+  if (isKnownRouter('kyber', input)) return decodeKyber(input)
+  if (input.chain === Chain.Ethereum) {
+    const router = input.tx.to.toLowerCase()
+    if (router === '0xd37bbe5744d730a1d98d8dc97c42f0ca46ad7146') return decodeThor(input)
```
Ehsan-saradar: The THOR router, Universal Router, WETH and ADDRESS_THIS (0x…02) addresses are hardcoded inline. THOR and UR already live in knownContracts.ts, so could we pull them from there, or at least make them named constants?

packages/sdk/src/tools/prep/evmSwapCommitment.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (desc.flags & 2n) return {}
+    const native = [nativeToken, zeroAddress].includes(desc.srcToken.toLowerCase())
+    if (parseValue(tx.value) !== (native ? desc.amount : 0n)) return malformed()
+    return sourceCommitment(native ? zeroAddress : desc.srcToken, desc.amount)
+  }
+  if (decoded.functionName.startsWith('ethUnoswap')) {
+    return sourceCommitment(zeroAddress, parseValue(tx.value))
+  }
+  const offset = decoded.functionName.startsWith('unoswapTo') ? 1 : 0
+  const args = decoded.args as readonly bigint[]
```
Ehsan-saradar: Small one: `as readonly bigint[]` goes against the no-`as` rule. Narrowing on `functionName` should give the typed args without the cast.

Ehsan-saradar (approved):
LGTM

packages/sdk/src/tools/prep/evmSwapCommitment.ts
rcoderdev: Fixed in 4ac24b6d6. Value mismatches now report the expected and received amounts separately from malformed calldata. The 1inch V5/V6 and Kyber tests cover the message, and actual preparation rejects both native mismatches before wallet initialization.

packages/sdk/src/tools/prep/evmSwapCommitment.ts
rcoderdev: Fixed in 4ac24b6d6 with named Ethereum THOR Router, Universal Router, WETH and ADDRESS_THIS constants. The addresses and supported deployments remain unchanged; THOR and wrapped-native Universal Router preparation were rechecked.

packages/sdk/src/tools/prep/evmSwapCommitment.ts
rcoderdev: Fixed in 4ac24b6d6. Explicit functionName cases now narrow unoswap and unoswapTo arguments without a cast. I also removed the Hex assertion using an isHex guard. Typechecking and the existing layout tests pass.

PR #2378, opened 2026-09-14
fix(cli): render ERC-20 approve consent lines from the signed calldata (T12)

neavra:
## Summary

Companion to https://github.com/vultisig/mcp-ts/pull/789 (dogfood round 2 finding **T12**, fund-safety / consent integrity): the pre-sign consent line — what `--yes` authorizes and what a no-`--yes` run reports as `proposed:` — must describe the bytes being signed.

`getPendingSummary` treated any envelope carrying swap labels as a swap and rendered the producer's `quote_summary`. On a USDT-style allowance-reset turn mcp-ts ships `approve(spender, 0)` ALONE as `txArgs` while the swap labels ride along, so the user was asked to authorize "5 USDC → ~51.56 POL via swapkit …" while the signable was an allowance reset. The 2-leg envelope disclosed its approve leg only as an anonymous "(+ token approval — 2 transactions)".

### Changes (`clients/cli/src/agent/executor.ts`)
- **`decodeErc20Approve`** (mirror of `decodeErc20Transfer`): decodes spender + allowance from `approve(address,uint256)` calldata; malformed approve calldata throws.
- **Single-leg envelope whose signed calldata is an approve** renders from calldata, never from labels:
  `reset USDC allowance to 0 on Polygon for spender 0x9025… (token contract 0x3c49…) — approval only; no swap is signed in this transaction est. fee ~0.2629 POL`
  (the "approval only" clause appears when swap labels are present).
- **Multi-leg envelope**: the suffix names the decoded approve leg — `(+ first approve 2 USDC for spender 0x… (token contract 0x…) — 2 transactions)`. An approval leg whose calldata is not an ERC-20 approve, or is malformed, fails closed (buffer cleared + throw), like transfers.
- Known tokens use trusted ticker/decimals from `knownTokensIndex`; unknown tokens render raw base units with an explicit "decimals unverified".

Producer labels (`quote_summary`, `approval_summary`) never reach the amount/spender text — WYSIWYS, same policy as `assertConsistentTransfer`.

### Tests
- `executor.summary.test.ts`: multi-leg fixtures now carry real viem-encoded approve calldata (Base USDC, a known token) and pin the decoded suffix; ten new T12 cases (reset turn with swap labels, decoy `approval_summary` label, approve(N), unknown token, multi-leg approve/reset, malformed + non-approve fail-closed, clean single-leg swap unchanged).
[...]

PR #2436, opened 2026-09-23
fix(cli): render ERC-20 approve consent lines from the signed calldata

neavra:
## Summary

- The pre-sign consent line (what `--yes` authorizes, and what a run without `--yes` reports as `proposed:`) now describes the bytes being signed. When the signable calldata is an ERC-20 `approve`, the line is rendered from the decoded spender and allowance, never from producer labels.
- Companion to vultisig/mcp-ts#789 (the producer-side fix). Either can land independently.
- Supersedes #2378: same change rebased onto current `main`, squashed, with the missing changeset.

## Problem

`getPendingSummary` treated any envelope carrying swap labels as a swap and rendered the producer's `quote_summary`. On an allowance-reset turn (USDT-style tokens) the backend ships `approve(spender, 0)` alone as the signable while the swap labels ride along, so the user was asked to authorize "5 USDC → ~51.56 POL via swapkit …" when the transaction being signed was an allowance reset. Two-leg envelopes disclosed the approve leg only as an anonymous "(+ token approval — 2 transactions)".

## Changes (`clients/cli/src/agent/executor.ts`, `clients/cli/src/agent/session.ts`)

- `decodeErc20Approve` (mirror of `decodeErc20Transfer`) decodes spender and allowance from `approve(address,uint256)` calldata; malformed calldata throws.
- Decoding runs before every envelope branch, including the `__buildTx` (Polymarket / yield) path, so a single-leg approve is never presented as a generic contract call.
- A single-leg approve renders from calldata, e.g. `reset USDC allowance to 0 on Polygon for spender 0x9025… (token contract 0x3c49…) — approval only; no swap is signed in this transaction` (the "approval only" clause appears when swap labels are present).
- Multi-leg envelopes decode both legs. The suffix names the approve leg (`(+ first approve 2 USDC for spender 0x… (token contract 0x…) — 2 transactions)`). A main leg that is itself an approve, or an approval leg whose calldata is not an ERC-20 approve or is malformed, fails closed: buffer cleared and the sign refused, the same policy as transfers.
- Unlimited allowances (`2^256-1`) render as `UNLIMITED`. Known tokens use ticker and decimals from `knownTokensIndex`; unknown tokens render raw base units with an explicit "decimals unverified".
- The signing-summary cap for `sign_tx` is raised from 500 to 1000 characters (`consentLimits.ts`, shared by the executor and the session) so a long swap head plus contract disclosures cannot push the spender out of `proposed:` or the signing record.
- Review round (rcoderdev): the approve disclosure is now guaranteed to sit inside that cap. `quote_summary`, `amount_in`, symbols, provider and fee go through the same 512-character label bound as the other labels (over-long text falls back instead of rendering), and on a two-leg envelope the swap head is sized against the protected tail (`on <chain>`, `via <provider>`, the decoded approve part) and truncated with `…` when needed. The fee part stays after the approve part and may still be cut by the cap; the approve part cannot.
- Producer labels (`quote_summary`, `approval_summary`) never reach the amount or spender text, the same policy as `assertConsistentTransfer`.

## Tests

- `executor.summary.test.ts`: multi-leg fixtures carry real viem-encoded approve calldata (Base USDC, a known token) and pin the decoded suffix. New cases cover the reset turn with swap labels, a decoy `approval_summary` label, `approve(N)`, unknown token, `UNLIMITED`, multi-leg approve and reset, a main leg that is an approve, malformed and non-approve calldata failing closed, `__buildTx` envelopes, and a clean single-leg swap left unchanged.
- Two label-path tests (`executor.summary.test.ts`, `executor.recipientSummary.test.ts`) moved off approve calldata to an opaque selector, since approves no longer render from labels.
- Cap tests: two cases put the spender beyond character 1000 without the fix (a 1200-character `quote_summary`, a 1200-character `amount_in`), and one forces the head-truncation branch with labels that are each inside the 512-character bound (verified to fail with truncation disabled). All assert the full spender, allowance, token contract and `2 transactions` text survive `capSigningSummary(summary, 'sign_tx')`.
- CLI suite, typecheck, lint and Prettier green locally on this branch.

### Test cases added (`executor.summary.test.ts`)

- `reset turn: single-leg approve(spender, 0) buffered with SWAP labels renders the allowance reset, not the swap`
- `reset turn ignores a producer approval_summary label and renders the spender from the bytes`
- `single-leg approve(spender, N) on a known token renders the decoded allowance in token units`
- `single-leg approve on an unknown token falls back to raw base units (decimals unverified)`
- `multi-leg: the approve leg is disclosed from its calldata next to the swap head`
- `multi-leg: an approve(spender, 0) leg is disclosed as an allowance reset`
- `fails closed on malformed approve calldata (selector matches, args truncated) and clears the buffer`
- `fails closed when a multi-leg approval leg is not an ERC-20 approve and clears the buffer`
- `fails closed on malformed calldata in a multi-leg approval leg`
- `renders an unlimited (2^256-1) allowance as UNLIMITED, known and unknown token`
- `reset turn without a fee label omits the fee but keeps the approval-only clause`
- `fails closed when the multi-leg MAIN leg is an ERC-20 approve hidden behind a swap head`
- `fails closed on malformed approve calldata in the multi-leg MAIN leg`
- `single-leg swap without an approve (clean allowance) still renders the swap summary`
- `keeps the approve disclosure inside the sign_tx cap when quote_summary is over-long`
- `keeps the approve disclosure inside the sign_tx cap when amount_in is over-long`
- `truncates an in-bounds swap head while preserving the provider and approve disclosure`

### Local validation (all green)

- `yarn build:shared`
- `yarn build:sdk:bundle`
- `yarn build:client-shared`
- `yarn build:rujira`
- `yarn vitest run --config .config/vitest.cli.config.ts clients/cli/src/agent --maxWorkers=2` (43 files, 821 tests)
- `yarn test:cli --maxWorkers=2` (82 files, 1,283 tests)
- `yarn typecheck`
- `yarn lint`
- `npx prettier --config .config/.prettierrc --ignore-path .config/.prettierignore --check 'clients/**/*.{ts,tsx,js,jsx}'`
- `yarn changeset:check-sdk-bundle`

## Summary by CodeRabbit

* **New Features**
  * Signing confirmations now show ERC-20 approval details decoded from the signed transaction, including the spender and allowance. Unlimited allowances are labeled **UNLIMITED**, and unknown token amounts are shown in base units with a decimals-unverified notice.
  * Multi-step transaction confirmations identify the approval leg, while non-approval calls retain their contract-call summary.
* **Bug Fixes**
  * Invalid or mismatched approval data now stops signing rather than displaying an unverified approval summary.
  * Long transaction confirmations retain approval details without truncation, including in signing records.


rcoderdev (commented):
On this head, `quote_summary` accepts a single-arrow summary without a length limit, and the decoded approve leg and spender are appended afterward. `capSigningSummary` then keeps only the first 1,000 characters for the `agent ask` proposal and the `--yes` signing record. A long provider summary can therefore push the spender and allowance out of both records while the transaction is still signable. Please bound the quote text or reserve space for the approval disclosure before capping, and test a summary that puts the spender beyond character 1,000.

